PR #19, opened 2012-12-17
SEC-2083: Created a class that allows filtering of immutable collections...

matsev:
This is a suggestion on how the suggestion in https://jira.springsource.org/browse/SEC-2083 can be implemented. I was hesitating whether or not to extend the DefaultMethodSecurityExpressionHandler for this use case, or refactor it to form an AbstractMethodSecurityExpressionHandler that could be extended. Personally, I prefer the latter, but with the provided solution no existing code is changed.

### DIFF
--- a/core/src/main/java/org/springframework/security/access/expression/method/ImmutableMethodSecurityExpressionHandler.java
+++ b/core/src/main/java/org/springframework/security/access/expression/method/ImmutableMethodSecurityExpressionHandler.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2002-2012 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.access.expression.method;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.springframework.expression.EvaluationContext;
+import org.springframework.expression.Expression;
+import org.springframework.security.access.PermissionCacheOptimizer;
+import org.springframework.security.access.expression.ExpressionUtils;
+
+import java.util.*;
+
+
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@matsev Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal)!

pivotal-issuemaster:
@matsev Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=19) to manually synchronize the status of this Pull Request.

core/src/main/java/org/springframework/security/access/expression/method/ImmutableMethodSecurityExpressionHandler.java
```diff
@@ -0,0 +1,105 @@
...
+    @Override
+    public void setPermissionCacheOptimizer(PermissionCacheOptimizer permissionCacheOptimizer) {
+        this.permissionCacheOptimizer = permissionCacheOptimizer;
+    }
+
+    private List filterCollection(Collection filterTarget, Expression filterExpression, EvaluationContext ctx) {
+        final MethodSecurityExpressionOperations rootObject = (MethodSecurityExpressionOperations) ctx.getRootObject().getValue();
+        final boolean debug = logger.isDebugEnabled();
+        List retainList = new ArrayList(filterTarget.size());
+        if (debug) {
```
vborcea: this tow ifs could be combined:
if(debug){
 +            logger.debug("Filtering with expression: " + filterExpression.getExpressionString());
 +            logger.debug("Filtering collection with " + filterTarget.size() + " elements");
 +        }

pivotal-issuemaster:
@matsev Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=19)!

eleftherias:
@matsev I'm moving the discussion from gh-2316 to this thread for better visibility.

> Thanks for the submission. I think I would prefer a strategy style pattern rather than inheritance. This would allow handling of custom types. One issue I see w/ the pull request is it does not seem to consider that some users may use specific types of collections. For example, if the method signature were a LinkedList this solution does not look like it would work since LinkedList is a List and the solution would use an ArrayList. A PR that would be merged would certainly need some tests to demonstrate these use cases working.

Let us know if you are interested in modifying the PR according to the comment above.

matsev:
@eleftherias I am sorry, but I have no intention of updating the PR.

After reading back, I recall the the problem I had back then, but I changed projects a long time ago. Supporting immutable collections still seems like good idea though, feel free to modify the PR in any way you find suitable.

PR #40, opened 2013-08-08
SEC-977: Add support for CAS gateway feature

miremond:
The opportunity and the implementation details of this new feature were discussed in Jira SEC-977.

The new filter TriggerCasGatewayAuthenticationFilter has been added to
call the CasAuthenticationEntryPoint when we want try a silent CAS
authentication (typically on a public page). The trigger criteria is
done with a requestMatcher instance.
The method unsuccessfulAuthentication has been overridden in
CasAuthenticationFilter in order to redirect to the saved url if there
was no SSO session (no service ticket sent from CAS).
To avoid infinite loop, we use the DefaultGatewayResolverImpl from Jasig
Cas Client.

### DIFF
--- a/cas/src/main/java/org/springframework/security/cas/authentication/TriggerCasGatewayException.java
+++ b/cas/src/main/java/org/springframework/security/cas/authentication/TriggerCasGatewayException.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2013-2013 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.cas.authentication;
+
+import org.springframework.security.cas.web.CasAuthenticationEntryPoint;
+import org.springframework.security.core.AuthenticationException;
+
+/**
+ * Exception used to indicate the {@link CasAuthenticationEntryPoint} to make a CAS gateway authentication request.
+ *
+ * @author Michael Remond
+ *
+ */
+public class TriggerCasGatewayException extends
+        AuthenticationException {
+
+    //~ Constructors ===================================================================================================
+
+    /**
+     * Constructs an {@code InitiateCasGatewayAuthenticationException} with the specified message and no root cause.
+     *
+     * @param msg the detail message
+     */
+    public TriggerCasGatewayException(String msg) {
+        super(msg);
+    }
+
+    /**
+     * Constructs an {@code InitiateCasGatewayAuthenticationException} with the specified message and root cause.
+     *
+     * @param msg the detail message
+     * @param t the root cause
+     */
+    public TriggerCasGatewayException(String msg, Throwable t) {
+        super(msg, t);
+    }
+
+}
--- a/cas/src/main/java/org/springframework/security/cas/web/CasAuthenticationEntryPoint.java
+++ b/cas/src/main/java/org/springframework/security/cas/web/CasAuthenticationEntryPoint.java
@@ -1,4 +1,4 @@
-/* Copyright 2004, 2005, 2006 Acegi Technology Pty Limited
+/* Copyright 2004, 2005, 2006, 2013 Acegi Technology Pty Limited
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,10 +22,11 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 import org.jasig.cas.client.util.CommonUtils;
+import org.springframework.beans.factory.InitializingBean;
 import org.springframework.security.cas.ServiceProperties;
+import org.springframework.security.cas.authentication.TriggerCasGatewayException;
 import org.springframework.security.core.AuthenticationException;
 import org.springframework.security.web.AuthenticationEntryPoint;
-import org.springframework.beans.factory.InitializingBean;
 import org.springframework.util.Assert;
 
 
@@ -38,6 +39,9 @@ import org.springframework.util.Assert;
  * redirect to the page indicated by the <code>service</code> property. The <code>service</code> is a HTTP URL
  * belonging to the current application. The <code>service</code> URL is monitored by the {@link CasAuthenticationFilter},
  * which will validate the CAS login was successful.
+ * <p>
+ * If the raised exception is {@link TriggerCasGatewayException}, the parameter <code>gateway</code> is set
+ * to true on the redirect url.
  *
  * @author Ben Alex
  * @author Scott Battaglia
@@ -72,7 +76,7 @@ public class CasAuthenticationEntryPoint implements AuthenticationEntryPoint, In
             final AuthenticationException authenticationException) throws IOException, ServletException {
 
         final String urlEncodedService = createServiceUrl(servletRequest, response);
-        final String redirectUrl = createRedirectUrl(urlEncodedService);
+        final String redirectUrl = createRedirectUrl(servletRequest, response, authenticationException, urlEncodedService);
 
         preCommence(servletRequest, response);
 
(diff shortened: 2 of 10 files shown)


jgribonvald:
+1 interest

Otherwise i followed this approch to redirect after authentication on specific page : http://jlorenzen.blogspot.fr/2013/07/remember-target-url-with-spring.html

julicrack:
I'm afraid this solution does not fit perfectly to my circumstances.

In my opinion there are two pitfalls:
* The first one is that redirecting promiscuously on every single request is a performance problem.
* On other hand this may affect to browsing history behaviour.
* And at last but not least from a SEO point of view may be a real problem robots index a 302 as response to a home page, for instance.

I have implemented a easy solution based on javascript in order to trigger a gateway aware request if and only if CAS query (CASTGC) is present. Let me know if you need any further information.

jgribonvald:
@julicrack I'm interested, because i used my specified approach to authenticate on a REST api for a web single page done with angularJS. And I found the CAS approach wasn't really easy/obvious (for this case) when we compare to other auth.

julicrack:
My approach consists on including an iframe within client page pointing to CAS domain.

This iframe would include a javascript in order to dinamically obtain CAS ticket granting cookie and if 'valid' reload page on top window using gateway mode.

The aim is to avoid as much as possible gateway requests.

This is the iframe content:
```html
<html>
	<head>
		<script src="/js/gateway.js"></script>
	</head>
	<body onload="CAS.triggerGatewayAuth()">
		<form id="reloadParent" name="reloadParent" method="GET" target="_top">
			<input type="hidden" name="gateway" value="true" />
		</form>
	</body id="body">
</html>
```
And this is the included javascript:
```javascript

var CAS = {
	// Regexp that matches an URL
	urlRegexp : /^(?:([A-Za-z]+):)?(\/{0,3})([0-9.\-A-Za-z]+)(?::(\d+))?(?:\/([^?#]*))?(?:\?([^#]*))?(?:#(.*))?$/,

	// Regexp that matches the Query String Part (QSP)
	qspRegexp : /([^?=&]+)(=([^&]*))?/g,

	// Triggers a CAS gateway authentication if Ticket Granting Cookie (TGC) is present and has not already being
	// processed (TGC == CTR)
	triggerGatewayAuth : function() {
		/*
		 * Parse iframe location to obtain a gateway aware redirect location
		 * 
		 * i.e.: <iframe src="http://my.cas.com/gateway.html#http://mysite.com?foo=bar"></iframe>
		 */
		var triggerCasGatewayUrl = decodeURI(CAS.urlRegexp.exec(location.href)[7]); // i.e.: http://mysite.com?foo=bar
		var triggerCasGatewayUrlParts = CAS.urlRegexp.exec(triggerCasGatewayUrl);
		var triggerCasGatewayHst = triggerCasGatewayUrlParts[3]; // i.e.: mysite.com
		var triggerCasGatewayQsp = triggerCasGatewayUrlParts[6]; // i.e.: ?foo=bar

		var casCtrCookieName = 'CASCTR.' + triggerCasGatewayHst;

		var casTgc = W3S.getCookie('CASTGC'); // CAS ticket granting cookie
		var casCtr = W3S.getCookie(casCtrCookieName); // Last tested ticket granting cookie

		// If ticket granting cookie exist, has not been tested and
		if ((typeof casTgc != 'undefined') && (typeof triggerCasGatewayUrl != 'undefined')) {

			if (casCtr != casTgc) {

				var triggerCasGatewayQsp = CAS.urlRegexp.exec(triggerCasGatewayUrl)[6]; // i.e.: ?foo=bar

				var triggerCasGatewayQspMap = {}; // https://remysharp.com/2008/06/24/query-string-to-object-via-regex
				triggerCasGatewayQsp.replace(CAS.qspRegexp, function($0, $1, $2, $3) {
					triggerCasGatewayQspMap[$1] = $3;
				});

				// Avoid infinite loop
				W3S.setCookie(casCtrCookieName, casTgc);

				// Trigger a gateway aware authentication request
				CAS.decorateForm(document.forms[0], triggerCasGatewayUrl, triggerCasGatewayQspMap).submit();
			}
		}
	},

	// Decorated the form that triggers the authentication request
	decorateForm : function(formElement, action, parameterMap) {

		formElement.action = action;

		for ( var key in parameterMap) {
			var inputElement = document.createElement("input");

			inputElement.setAttribute('type', 'hidden');
			inputElement.setAttribute('name', key);
			inputElement.setAttribute('value', parameterMap[key]);

			formElement.appendChild(inputElement);
		}

		return formElement;
	}
};

/*
 * http://www.w3schools.com/js/js_cookies.asp
 */
var W3S = {
	setCookie : function(cname, cvalue, exdays) {
		var d = new Date();
		d.setTime(d.getTime() + (exdays * 24 * 60 * 60 * 1000));
		var expires = "expires=" + d.toUTCString();
		document.cookie = cname + "=" + cvalue + "; " + expires;
	},

	getCookie : function(cname) {
		var name = cname + "=";
		var ca = document.cookie.split(';');
		for (var i = 0; i < ca.length; i++) {
			var c = ca[i];
			while (c.charAt(0) == ' ')
				c = c.substring(1);
			if (c.indexOf(name) == 0)
				return c.substring(name.length, c.length);
		}
		// return "";
	}
}
```

Iframe src should be something like this:
https://cas.domain/html/gateway.html#current_page_url_with_query_string

Gateway mode is enabled if and only if 'gateway' parameter is included and true.

Do note external javascript frameworks are intentionally avoided.

Any feedback would be really apprecitated.

miremond:
hello
@julicrack From what I can see here, you implemented the gateway feature almost entirely in javascript; I think this does not fit to the spring security philosophy. We should ask @rwinch. 
I agree with you that SEO could be an issue with the gateway mechanism. I think one can easily fix that with a simple cookie indicating if we should attempt gateway. Then, if you consider my PR, you can simply override the method:
    protected boolean performGatewayAuthentication(HttpServletRequest request) {
          return true;
     }

But again, there is something missing to fulfill this PR; I will soon submit an update.

julicrack:
Umm, I'm afraid that Google does (or will) take into account JavaScript and/or cookie during indexing.

That is the reason why script above triggers a gateway authentication if and only if a cookie *from other domain* is present. Google will never keep track of this cookie (CASTGC) and gateway mode is avoided during crawling procceses.

On other hand, the unique mission of this script is just to realize when is the more accurate moment to trigger a gateway request. In such a case the negotiation is done according with this branch.

Further reading:
http://www.tomanthony.co.uk/blog/googlebot-accepting-cookies/

rwinch:
Thanks for carrying this forward guys. As @miremond guessed, I'd prefer to allow this feature to work without JavaScript. I don't think there is necessarily a problem with using JavaScript in individual applications for customizations, I just don't thinks something like CAS Gateway should require JavaScript. With that in mind, I prefer the cookie approach.

julicrack:
A cookie trigerring based solution is easier... and 90% of times easier means better.

Supossing Google won't follow cookies this would be a valid approach from SEO point of view.

However asumming every single anonymous request will carry out a 302 is too much IMHO.

Aproach above not only reduce the number of hits to CAS server but also simplifies browsing and backward and forward button from cliente webapp point of view.

Thanks indeed for your support.

dave0100:
Hi all, I just wanted to share a server side implementation based on the feedback, patches and discussion here - in case others don't want to manually build spring security or modify cas. We of course want to use this patch once it landed, but this approch works for us so far:

- map a filter before the spring security filter chain in your web.xml to your landing page, the one you want both anonymous access on but autologin in users which already have a CAS ticket:

```
    public void doFilter(javax.servlet.ServletRequest req, javax.servlet.ServletResponse res, javax.servlet.FilterChain chain) throws java.io.IOException, javax.servlet.ServletException {

        if(req instanceof HttpServletRequest) {
            HttpServletRequest request = ((HttpServletRequest)req);

            LOG.trace("CASAuthFilter called");

            if(toggles.is("cas.gateway.enabled")) {

                LOG.debug("CASAuthFilter recognized request for main page, checking login conditions");

                if(WebUtils.getSessionAttribute(request, "ssoCheckDone") == null) {

                    LOG.debug("SSO check not done yet");

                    //don't use this feature for requests made via redirect.html (yet, may enable at will)
                    if(WebUtils.getSessionAttribute(request, "startImmediately") == null) {

                        WebUtils.setSessionAttribute(request, "ssoCheckDone", true);

                        HttpServletResponse response = ((HttpServletResponse) res);

                        String url = casURL + "/login?gateway=true&service=" + service + "/j_spring_cas_security_check";

                        LOG.debug("Redirecting to gateway URL: "+url);

                        response.sendRedirect(url);
                        response.flushBuffer();
                        response.getOutputStream().close();

                        return;
                    }
                    else {
                        LOG.debug("SSO check skipped because startImmediatly is set");
                    }
                }
                else {
                    LOG.debug("SSO check has already been done");
                }
            }
        }

        chain.doFilter(req, res);
    }
```

The HTTP session attribute ssoCheckDone takes care of just using the CAS gateway feature once for each session.

- Map an error handler in your web.xml an let him handle the Exception that is thrown if /j_spring_cas_security_check (the gateway return URL) is called without a valid ticket. This redirects back to the landingpage:

```
	@RequestMapping("error")
	public String handelError(final HttpServletRequest request, HttpServletResponse response, final Model model) {

		//ONSE-9988 Fail gracefully here, as it can be normal behaviour after using the gateway feature
		if(toggles != null && toggles.is("cas.gateway.enabled") && handleGracefulSSO(request, response)) {
			return null;
		}
         return "error/error";
    }
```

```
	private boolean handleGracefulSSO(final HttpServletRequest request, HttpServletResponse response) {
		String fwdURL = (String) request.getAttribute(RequestDispatcher.FORWARD_REQUEST_URI);

		LOG.warn(fwdURL);

		if(fwdURL != null && fwdURL.endsWith("/j_spring_cas_security_check")) {
			LOG.warn("CAS security check failed, redirecting to /app gracefully");

			try {
				response.sendRedirect("/app");
				response.flushBuffer();
				response.getOutputStream().close();

				return true;
			}
			catch (Throwable t) {
				LOG.error("Response failed", t);
			}
		}

		return false;
	}
```

pivotal-issuemaster:
@miremond Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal)!

pivotal-issuemaster:
@miremond Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=40) to manually synchronize the status of this Pull Request.

rwinch:
Closing this given there hasn't been the updates (i.e. use cookies to prevent redirecting for every request) have not been made

PR #69, opened 2014-01-28
SEC-1869 additions: SuccessHandler and FailureHandler for pre-auth filter implementation

sryll:
This is a solution I want to suggest to address the issue described in https://jira.springsource.org/browse/SEC-1869

The point is that the documentation states that AuthenticationSuccessHandlers and AuthenticationFailureHandlers can be used to react to (un)successful authentication events. As these handlers receive request and response objects via their signature, they are more powerful than reacting to AuthenticationEvents.

<quote from documentation, PDF section 9.4>
The destination following a successful authentication or an authentication failure is controlled by the AuthenticationSuccessHandler and  AuthenticationFailureHandler strategy interfaces, respectively.
</quote>

This pull request adds support for both handlers types to pre-authenticated filter implementations by amending AbstractPreAuthenticatedProcessingFilter.

If that patch needs some more work to be ready for inclusion, pls let me know.

I have submitted the CLA in 2011 already, in the context of SEC-1793.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/preauth/AbstractPreAuthenticatedProcessingFilter.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/preauth/AbstractPreAuthenticatedProcessingFilter.java
@@ -19,7 +19,9 @@ import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.security.web.WebAttributes;
+import org.springframework.security.web.authentication.AuthenticationFailureHandler;
 import org.springframework.security.web.authentication.AuthenticationSuccessHandler;
+import org.springframework.security.web.authentication.SimpleUrlAuthenticationFailureHandler;
 import org.springframework.security.web.authentication.WebAuthenticationDetailsSource;
 import org.springframework.util.Assert;
 import org.springframework.web.filter.GenericFilterBean;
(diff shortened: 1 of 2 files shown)


sryll:
Just checking back after a while, and wondering what the "Waiting for triage" label means for me? Is there any expectation attached and I missed something?

rwinch:
@sryll Thanks for the bump. Waiting for triage means that we haven't gotten a good understanding of the issue yet.

I took a look at the issue and I'm not certain I understand the reasoning for wanting the AuthenticationSuccessHandler and AuthenticationFailureHandler in a pre authenticated scenario. Can you describe to me what you are doing in those scenarios?

The reason I am asking is because typically a AuthenticationSuccessHandler and AuthenticationFailureHandler will commit the response. This makes sense for form based login where we must then trigger the original request (authentication success) or send to a page that a user can re-enter credentials (authentication failure). However, a PreAuthenticationFilter will continue invoking the FilterChain which will likely cause failures.

sryll:
@rwinch The use case I have is to have a hook to initialize the HttpSession after a successful authentication. I want to load some user data, preferences etc. from a database upon login and then store it in the session.
I don't have a similar story for the fail-handler, but it looked sensible to do it consistently.

Is there maybe some better way to achieve this session init?

pivotal-issuemaster:
@sryll Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

pivotal-issuemaster:
@sryll Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal)!

pivotal-issuemaster:
@sryll Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=69)!

sryll:
Equivalent changes have been merged in the meantime with the fix for issue #3389

PR #103, opened 2014-07-02
Trivial logging fix in saveContext method in HttpSessionSecurityContextRepository

bcecchinato:

I've added this log as when you have some session management issues (I had the case with Jetty), so you can check that sessions are the same between requests.


rwinch:

PR #106, opened 2014-07-11
DefaultSpringSecurityContextSource does not work with spring-ldap-2.0.0

knapsu:
Spring LDAP as of version 2.0.0 made userDn field private which DefaultSpringSecurityContextSource used directly. Because of this webapps that had spring-security-3.2.x and spring-ldap-2.0.0 failed to deploy/initialize beans. This does not happen with spring-ldap-1.3.2 which has this field not private.

spring-ldap-2.0.2 created a workaround for this bug andmade userDn field protected and marked it as deprecated  with suggestion to use a getter method getUserDn().


rwinch:
Thanks for the report and PR! We want it to work with Spring LDAP 2.x and 1.x so we cannot use the getter. Instead update to LDAP 2.0.1 See https://jira.spring.io/browse/LDAP-294

PR #108, opened 2014-07-12
SEC-2189: Modifed the AbstractAuthorizeTag#authorizeUsingUrlCheck

kazuki43zoo:
When make an authorization decision based on the URL and HTTP method attributes,
if multiple WebInvocationPrivilegeEvaluator are defined,
was modified as call the isAllowed() of all WebInvocationPrivilegeEvaluator.

Please review because i was fix this issue.
If there are not problem, please merge to master branch.


rwinch:
Thanks for the PR! I don't think this actually resolves the issue though. This will return false if any of the WebInvocationEvaluators deny access, but they may have deny by default. For example, consider what happens for the URL of /public:

```xml
<http pattern="/admin/**" ...>
    ...
    <intercept-url pattern="/**" access="denyAll"/>
</http>
<http ...>
    <intercept-url pattern="/**" access="permitAll"/>
</http>
```

kazuki43zoo:
Thanks for comment. I consider again !

PR #129, opened 2014-09-09
SEC-2071: Allow to provide a root DN different from the domain

henri-tremblay:
This is an improvement to the current implementation to provide flexibility when the root DN doesn't match the domain and/or the userPrincipalName.


rwinch:
Thanks for the PR! This is now merged into master (see fbb902c9b52f9b42b11028ada357f7bb2401be95) with one slight change. I moved the optional parameter to be the last argument (see c0fbd30e654c804be6c1717a20bd02996ecd982f). This is a convention that we typically use (i.e. developers can remove arguments from left to right to use more defaults).

PR #136, opened 2014-09-26
Update JdbcTokenRepositoryImpl.java

BigMichi1:
param should be used else the param is useless


rwinch:
Thanks for submitting this and for the PR!

I have merged this into master (see 8c0b16820b9f1e6b827c3d1dfa960821f96bfb23) and 3.2.x. I also created an issue to track this https://jira.spring.io/browse/SEC-2879, finally I added a test (see 6fd45df1e4e54d4d0ff654d2d7a176d6c3411696) to ensure it does not break again. Thanks again!

PR #145, opened 2014-11-19
HeaderWriterFilter rejects empty header writers list 

apechinsky:
HeaderWriterFilter rejects empty header writers list with message 'headerWriters cannot be null'.
So the check doesn't match the message.


rwinch:
This is intentional. If you do not want any headers added, do not add the Filter. I'm closing this.

If you have a good reason that you need to support empty writers please provide an explanation & reopen.

apechinsky:
First. Current implementatation is inconsistent because method Assert.notEmpty() checks if **collection is empty** but message reports that **collection is null**. This is confusing. 

To fix change method or message:
  - Assert.**notNull**(headerWriters, "headerWriters cannot be null");
  - Assert.notEmpty(headerWriters, "headerWriters cannot be **empty**");

Second. I think method should check for null but not for empty collection. 
This is because header writer list may be created dynamically at runtime. For example I have several spring configuration profiles some of them doesn't define header writers. When such configuration is activated I get IllegaleStateException due Assert check.

PR #155, opened 2015-01-05
Upgrade to PowerMock 1.6.1

wilkinsona:
The Platform would like to move to JUnit 4.12 but cannot do so at the
moment as Spring Security uses a version of PowerMock which is
incompatible with JUnit 4.12. This commit updates Spring Security to use
PowerMock 1.6.1 with is compatible with JUnit 4.12.


rwinch:
Thanks for spotting this has been released @wilkinsona! This is now merged into master.

PR #156, opened 2015-01-07
add setter for using a custom name for the rememberMeParameter

BigMichi1:


rwinch:
Thanks this has been merged into master via c8b79289c9fbebc08a4e0bb40490d59f28b4b6d5 and is tracked via https://jira.spring.io/browse/SEC-2874

PR #157, opened 2015-01-11
SEC-1915: Custom ActiveDirectory search filter

MateuszRasinski:
Currently the search filter used when retrieving user details is hard coded.

New property in ActiveDirectoryLdapAuthenticationProvider:
- searchFilter - the LDAP search filter to use when searching for authorities,
default to search using 'userPrincipalName' (current) OR 'sAMAccountName'

Performed also a refactoring of the ActiveDirectoryLdapAuthenticationProvider.

### DIFF
--- a/ldap/src/main/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProvider.java
+++ b/ldap/src/main/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2012 the original author or authors.
+ * Copyright 2002-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
  * the License. You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


MateuszRasinski:

MateuszRasinski:
Similar and partially based on: https://github.com/spring-projects/spring-security/pull/18

donovanmuller:
Is there any idea when this could be merged?
Facing the same problem and seems there has been no movement on this issue (as well as SEC-2224)

igler:
In my case the following link helped: http://stackoverflow.com/questions/23617096/spring-mvc-application-authenticating-to-active-directory

donovanmuller:
@igler I have done the same until this is fixed. :(

rwinch:
Thanks for the PR! I have merged this into master (see c54346b69090062e1695d0ffdc196c0dd3a95611) with a little bit of polish (see cd352f665bce2bbf8963f6bd4cc40cb6d145291d). The main point of the polish is that we do not want to change the default search filter to remain passive. I have also pushed the changes into 3.2.x so this will be available in 4.0.0.RC2+ and 3.2.6+

PR #162, opened 2015-01-18
SEC-2819

paskos:
SEC-2819
SimpDestinationMessageMatcher(String pattern, SimpMessageType type) never passes type param to this() constructor


rwinch:
Thanks for the PR @paskos! This is now merged into master

PR #163, opened 2015-01-19
SEC-2820

paskos:
I work on a project with complex dynamic security access rules. We
implemented our own SecurityExpressionHandler and we want to instantiate
MessageSecurityExpressionRoot but we can't because it's package
protected.


rwinch:
Thanks for the PR This is merged into master and will be in 4.0 GA

PR #176, opened 2015-03-01
SEC-2888

paskos:
SEC-2888 AntPathRequestMatcher ignores variables in pattern when pattern finishes with /**


rwinch:
This is now merged into master. Thanks again for the PR!

PR #177, opened 2015-03-04
Adds "challenge" property to BasicAuthenticationEntryPoint

dsyer:
This provides useful additional flexibility if you want a
WWW-Authenticate header but don't want the "Basic" challenge
(e.g. to prevent a browser from popping up a dialog).

See https://jira.spring.io/browse/SEC-2803


dsyer:
That second commit is just an idea (the general version with a replacement for the realm in the deafult case). If you don't like it please just discard.

rwinch:
@dsyer Thanks for the PR!

I wonder if it makes more sense to put this in another AuthenticationEntryPoint all together. The reason is that BasicAuthenticationEntryPoint seems to imply that it is performing basic authentication. As soon as the challenge is removed, this is no longer the case.

What are your thoughts on creating an HttpStatusAuthenticationEntryPoint that allows setting the HttpStatus and (optionally) any headers the client wishes to send? I'm guessing in most cases a client doesn't even need the headers, so perhaps we should leave that out for now.

Thoughts?

dsyer:
I'm ok with it either way. Thanks.

rwinch:
Closing this in favor of e776a1fd35dd

PR #179, opened 2015-03-06
Spring 4 should be allowed in bundle dependencies

soleger:
According to the JIRA release notes for Spring Security 3.2, it should have compatibility with Spring Framework 4.0. However, the springRange variable limits OSGi bundle compatibility to Spring Framework 3.2. The variable should be changed to allow compatibility with Spring Framework 4.0.

See:
https://jira.spring.io/browse/SEC-2350
https://jira.spring.io/browse/SEC-2385
https://jira.spring.io/browse/SEC-2434

I cannot find documentation about whether Spring Framework 4.1 is also considered to be compatible with Spring Security 3.2 but this is also a possibility.

### DIFF
--- a/gradle/javaprojects.gradle
+++ b/gradle/javaprojects.gradle
@@ -72,7 +72,7 @@ ext.apachedsDependencies = [
 ext.bundlorProperties = [
     version: version,
     secRange: "[$version, 3.3.0)",
-    springRange: "[$springVersion, 3.3.0)",
+    springRange: "[$springVersion, 4.1.0)",
     aspectjRange: '[1.6.0, 1.8.0)',
     casRange: '[3.1.1, 3.2.0)',
     cloggingRange: '[1.0.4, 2.0.0)',


rwinch:
@soleger AFAIK Spring 4 no longer does OSGi bundling so I'm not sure how much this would help. Can you provide any insight into as what your process in obtaining Spring 4 OSGi versions is?

soleger:
OSGi bundles for Spring Framework are maintained by Apache ServiceMix and are available on Maven's Central repository, for example:

http://repo1.maven.org/maven2/org/apache/servicemix/bundles/org.apache.servicemix.bundles.spring-core/

These bundles are used in the Apache Karaf OSGi container ecosystem. Spring Framework 'features' (collections of OSGi bundles) are included as part of the base Karaf container and they use the ServiceMix bundles in these features.

Because the OSGi metadata for Spring Security 3.2 limits the Spring Framework bundle version to [3.2,3.3), it is not possible to load Spring Security 3.2 in Apache Karaf as part of a project that depends on Spring Framework 4.0.

rwinch:
Thanks for the clarification. I created https://jira.spring.io/browse/SEC-2899

soleger:
Thanks! I look forward to seeing this compatibility update in a future release.

rwinch:
Thanks again for the report and the PR. I have updated to include 4.x

PR #175, opened 2015-02-25
SEC-2880: Add a setter method to override the cookie name of remember-me

kazuki43zoo:
I resolved the [SEC-2880](https://jira.spring.io/browse/SEC-2880).
I agreed the CLA.

Please review.

PR #184, opened 2015-03-27
SEC-2743: Add rememberCookieName() to RememberMeConfigurer

knalli:
This one allows

```java
http
   .rememberMe()
       .rememberCookieName("customName")
```

This reduce the requirement of a custom rememberMeService only for changing the default cookie name.

This is my first PR.. did I made it right?


knalli:
Ah.. looks like being a dup already. I have not seen #175

rwinch:
Thanks for putting this together. You are correct this is a duplicate. Resolving this as a duplicate of #175

PR #186, opened 2015-04-06
Deduplication of the inner SaveOnAccessCsrfToken classes

panchenko:
The SaveOnAccessCsrfToken classes nested into CsrfAuthenticationStrategy and CsrfFilter are identical, it makes to have a single class definition for it.


rwinch:
Thanks for the PR! I'd prefer to keep this class package scope or lower visibility. 

We were able to achieve that and removing the duplicate code via #3790

PR #187, opened 2015-04-07
"static" for inner classes

panchenko:
So these inner classes don't have an automatically generated reference to the outer class.


rwinch:
Thanks for the PR! This is now resolved via 0a118336d4ce64b93fb6a352227d7849d797ee5d and tracked with [SEC-2955|https://jira.spring.io/browse/SEC-2955]

PR #191, opened 2015-05-02
SEC-2957: add missing dependencies

shaehnel:
I created SEC-2957
I added dependencies which appeared to be missing to some of the sample projects


rwinch:
This is fixed in master. See the comments of https://jira.spring.io/browse/SEC-2957 for details

PR #198, opened 2015-06-09
Add new option for AES encryption with GCM

dsyer:
The Galois Counter Mode (GCM) is held to be superior than the current
default CBC. This change adds an extra parameter to the constructor
of AesBytesEncryptor and a new convenience method in Encryptors.

Fixes SEC-3002


rwinch:
Thanks for the PR! This is now merged into master a48cc1885 and 3.2.x ae77229

PR #200, opened 2015-06-15
SEC-2757: Removed assertion of 'sn' when creating LdapUserDetails

ckarawani:


rwinch:

PR #202, opened 2015-06-16
Add missing jstl dependency

ractive:
register.jsp uses jstl, but no jstl dependencies are imported.


rwinch:
Thanks for the PR! This is now resolved via be27ede0e9b4281963ca40ed08a05603243f5098

PR #203, opened 2015-07-02
SEC-2678 Disable HSTS by default

dvdckl:
See SEC-2678 for justification.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
@@ -235,7 +235,6 @@ public final class HttpSecurity extends
 	 *                 .contentTypeOptions();
 	 *                 .xssProtection()
 	 *                 .cacheControl()
-	 *                 .httpStrictTransportSecurity()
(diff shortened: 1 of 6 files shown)


rwinch:
Thank you for the PR. As per the discussion on [SEC-2678](https://jira.spring.io/browse/SEC-2678) we are closing the issue as "Works as Designed". From the issue:

> I'm closing this as "Works as Designed". The defaults of a security framework should be as secure as possible by default (without breaking passivity).
>
> Spring Security 3.2 ensured that existing users experienced the same behavior they had in previous versions. To this end there is no reason in my mind to change in regards to passivity.
>
> Spring Security's newly added Java configuration had different defaults to the XML configuration. Admittedly, this is somewhat confusing. However, the defaults ensured that users with Java Configuration had more secure defaults and a smoother upgrade experience to 4.x all while remaining passive.
>
> One of the major goals of the changes made in Spring Security 4.x was to ensure a more "secure by default" implementation was provided. This can be seen in multiple different changes throughout the framework (i.e. CSRF enabled by default, HTTP Response Headers included by default, etc).
>
> If there are suggestions on how to better document the defaults, then please create separate tickets so we can address those separately.

PR #204, opened 2015-07-06
Improve error message when using the wrong xsd schema.

SnakeSVx:
When using the wrong xsd schema < 4.0 a message was shown that the schema needed to be version 3.2.

In reality this schema had to be version 4.0.


rwinch:
Thanks this is merged via aed288d

PR #206, opened 2015-07-08
Use the latest version of the Spring IO Plugin

wilkinsona:
This change is necessary to allow Spring IO Platform 2.0 to remove the platform-versions properties file that was deprecated in 1.1. To allow Spring Security's Platform compliance to be verified as part of Spring IO Platform 2.0's build and release process, I'd like this to be included in Spring Security 4.0.x which is the version that's currently in Spring IO Platform 2.0. If you have any questions, please let me know.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ buildscript {
 	}
 	dependencies {
 		classpath("org.springframework.build.gradle:propdeps-plugin:0.0.7")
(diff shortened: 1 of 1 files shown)


rwinch:
Thanks for the PR! This is merged in as 350b48e3fd5d54bb7610b7d564ad17ccc4b02c31

PR #208, opened 2015-07-13
Upgrade to a version of Powermock that’s compatible with JUnit 4.12

wilkinsona:
The PR upgrades to Powermock 1.6.2 which is compatible with JUnit 4.12. It also upgrades Mockito as Powermock 1.6.2 is incompatible with the version of Mockito that was being used. 1.10.19 is the version used by Powermock 1.6.2.


rwinch:
Thanks for the PR! This is resolved via 92c23fbbdcc

PR #210, opened 2015-07-14
SEC-2666 Improve Field Visibility in DefaultMethodSecurityExpressionHandler

kevinconaway:
https://jira.spring.io/browse/SEC-2666

We are extending `DefaultMethodSecurityExpressionHandler` in order provide a custom `SecurityExpressionRoot`.

The base `SecurityExpressionRoot` requires an `AuthenticationTrustResolver` in order for all of the expressions to properly work.

 `DefaultMethodSecurityExpressionHandler` creates an `AuthenticationTrustResolver` but it is marked private and is unavailable to subclasses.  This requires subclasses to duplicate the initialization logic from the parent class.

This PR improves the visibility of the fields in `DefaultMethodSecurityExpressionHandler` from private to protected in order to better support subclasses.

### DIFF
--- a/core/src/main/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandler.java
+++ b/core/src/main/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandler.java
@@ -37,10 +37,10 @@ public class DefaultMethodSecurityExpressionHandler extends
 
 	protected final Log logger = LogFactory.getLog(getClass());
 
-	protected AuthenticationTrustResolver trustResolver = new AuthenticationTrustResolverImpl();
-	protected ParameterNameDiscoverer parameterNameDiscoverer = new DefaultSecurityParameterNameDiscoverer();
-	protected PermissionCacheOptimizer permissionCacheOptimizer = null;
-	protected String defaultRolePrefix = "ROLE_";
+	private AuthenticationTrustResolver trustResolver = new AuthenticationTrustResolverImpl();
+	private ParameterNameDiscoverer parameterNameDiscoverer = new DefaultSecurityParameterNameDiscoverer();
+	private PermissionCacheOptimizer permissionCacheOptimizer = null;
+	private String defaultRolePrefix = "ROLE_";
 
 	public DefaultMethodSecurityExpressionHandler() {
 	}
@@ -51,7 +51,7 @@ public class DefaultMethodSecurityExpressionHandler extends
 	 */
 	public StandardEvaluationContext createEvaluationContextInternal(Authentication auth,
 			MethodInvocation mi) {
-		return new MethodSecurityEvaluationContext(auth, mi, parameterNameDiscoverer);
+		return new MethodSecurityEvaluationContext(auth, mi, getParameterNameDiscoverer());
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@kevinconaway Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

kevinconaway:
Signed

pivotal-issuemaster:
@kevinconaway Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal)!

pivotal-issuemaster:
@kevinconaway Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

jgrandja:
Thanks for your input @kevinconaway. FYI, this design is intended. Class fields tend to be private for the most part but we would consider opening it up to protected access for subclass use. However, another design option we may choose instead is to provide a protected getter/setter while preserving the private member field.

There is a public setter available for your use case:

`public void setTrustResolver(AuthenticationTrustResolver trustResolver)`

So what you can do for your implementation is the following:

```
public class CustomMethodSecurityExpressionHandler extends DefaultMethodSecurityExpressionHandler {

	public CustomMethodSecurityExpressionHandler(AuthenticationTrustResolver trustResolver) {
		setTrustResolver(trustResolver);
	}
}

```

This solution should work for you. I'm going to close this PR. However, if this doesn't address your scenario please let me know and we can figure it out.

kevinconaway:
> However, if this doesn't address your scenario please let me know and we can figure it out.

Unfortunately not.  If I remember the scenario correctly (I've since switched jobs since last year), we were adding a `CustomMethodSecurityExpressionHandler` in order to add new expressions to the default ones like `hasRole`.

This is done by creating a custom `MethodSecurityExpressionRoot` and adding custom methods on that class, similar to [hasAuthority()](https://github.com/spring-projects/spring-security/blob/ce20cfcb9803a6175f6befc9008cadcc4e3b6d72/core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java#L66) or [hasRole()](https://github.com/spring-projects/spring-security/blob/ce20cfcb9803a6175f6befc9008cadcc4e3b6d72/core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java#L74).

To instantiate a custom `MethodSecurityExpressionRoot`, you need access to the `trustResolver` and `defaultRolePrefix`: https://github.com/spring-projects/spring-security/blob/ce20cfcb9803a6175f6befc9008cadcc4e3b6d72/core/src/main/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandler.java#L77

both of which are private.

What would work is providing getters for those fields or making them protected, not private.

jgrandja:
Gotchya.

My preference would be to expose protected accessors for subclasses rather than expose the members to protected.

Given that you switched jobs a while ago, do you still need a solution for this?

jgrandja:
I've re-opened the PR if you would like to update it as discussed with the protected accessors.
Thanks!

kevinconaway:
@jgrandja committed fixes per our discussion

jgrandja:
Thanks for the PR @kevinconaway. This is now merged to master.

PR #212, opened 2015-07-30
SEC-3056 - Fix JavaDoc errors.

thomasdarimont:
Fixed JavaDoc errors accross multiple modules in order to make javadoc happy with Java 8.

Unsupported close-tags like `</p>` were removed or replaced with `<p>` where appropriate.
Correct parameter references in `@param` doclets.
Fixed unresolvable `@link`doclets by either using the right target if possible or generating a raw html link via `<a href="...">` tags.
Correctly encoded html entities in tags.
Fixed incorrectly closed tags.
Removed old documentation parts that weren't in sync with the actual code / API.

With this change the spring-security project can be build with Java 7 and Java 8 without any errors.


rwinch:
Thanks! This is now merged into master!

PR #214, opened 2015-08-07
First attempt at fixing https://jira.spring.io/browse/SEC-3066

Stexxen:
I have added a new method to AclService. Not ideal, buts its just to start the ball rolling.

PR #215, opened 2015-08-07
Resolves SEC-3066 by handling exception.

Stexxen:
After thinking about it some more, realised that this is a much better solution.
By the time the Exception has been raised in AclService, the PermissionCacheOptimiser has done its job and already loaded the ACLs present.

So all thats required is to handle the Exception by a log entry and nothing else.

#3267

### DIFF
--- a/acl/src/main/java/org/springframework/security/acls/AclPermissionCacheOptimizer.java
+++ b/acl/src/main/java/org/springframework/security/acls/AclPermissionCacheOptimizer.java
@@ -10,6 +10,7 @@ import org.springframework.security.access.PermissionCacheOptimizer;
 import org.springframework.security.acls.domain.ObjectIdentityRetrievalStrategyImpl;
 import org.springframework.security.acls.domain.SidRetrievalStrategyImpl;
 import org.springframework.security.acls.model.AclService;
+import org.springframework.security.acls.model.NotFoundException;
 import org.springframework.security.acls.model.ObjectIdentity;
 import org.springframework.security.acls.model.ObjectIdentityRetrievalStrategy;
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@Stexxen Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

pivotal-issuemaster:
@Stexxen Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal)!

pivotal-issuemaster:
@Stexxen Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=215)!

rwinch:
I'm closing this for a few reasons:

* We cannot accept code that does a try / catch and does nothing to handle it (other than a debug statement). This will certainly cause problems for other users. This is what I view as the deal breaker
* We cannot accept a pull request without a test. This is not ideal, but in many circumstances I could add a test for you.

Stexxen:
I had described why it was done this way on the original issue (last comment) #3267

The exception occurs on a valid configuration, so no way past it without changing the interface as per pull #214, which is not ideal.

rwinch:
@Stexxen Thanks for the response.

The problem is that we cannot just swallow the exception either. Most systems are not configured to log debug messages in production (because it would not perform well), so try catch log is the same thing as swallowing the exception.

Stexxen:
no probs.
I suppose another way forward is to not raise the exception, but unfortunately is a valid exception in most circumstances just not this one. 
oh well, at least the bug still stands, but I have no viable fix for it.

PR #218, opened 2015-08-29
Fix NullPointerException

quangqiu:
- Need to compile against servlet-api 3.1


quangqiu:
also  for  null byte[] / chars[],  the content will be printed out as "null", thus the content length is 4, not 0

rwinch:
Thanks for the PR! The PR has additional changes that were not for this specific issue (newly added methods) and no tests. I'm going to close this in favor of #3823

PR #219, opened 2015-08-30
why websocket makes 

azizkhani:
i have an application that using spring sec version 4.0.2.RELEASE in weblogic
when i add websocket to my project and then connect to endpoint in jsp 

var socket = new WebSocket('ws://game.example.com:12010/updates');
socket.onopen = function () {
};
this 2 problem occurs.
Important note:
this problems occurs just when client connect to endpoint .without connection Everything OK
problems:
1)when user logout. session destroyed after session timeout but before add websocket session destroy on logout.
2)session time out not work and user remain in sessions 

i don't know why websocket makes this problem .


rwinch:
This is a question not a PR. Please post questions to StackOverflow with the spring-security tag http://stackoverflow.com/questions/tagged/spring-security  If you believe you have found a bug, please log them to Spring Security's JIRA at https://jira.spring.io/browse/SEC/

PR #221, opened 2015-09-07
Fix reference to Spring Security version in the manual (3.0->4.0)

codingdiscer:


rwinch:
Thanks for the PR! This is merged into master

PR #225, opened 2015-09-19
snasphot -> snapshot

zhanhb:


rwinch:
Thanks for the PR. The pom file is generated so we should also fix the source of the problem in https://github.com/spring-projects/spring-security/blob/master/gradle/maven-deployment.gradle I merged this PR and also fixed the gradle build.

Thanks again!

PR #226, opened 2015-09-24
Fix reference to Null Object pattern in the manual

nivenoct:


rwinch:
Thanks for your PR! This is now merged into master :)

PR #227, opened 2015-09-29
Fixed incorrect dn.

zshift:
Format of the dn was incorrect. Looks like a bad copy/paste.


rwinch:
Thanks this is now merged

PR #232, opened 2015-10-30
CachingUserDetailsService constructor visbility

vorburger:
CachingUserDetailsService constructor's visbility probably should not be package only, but probably public, or at least protected?


jgrandja:
Thanks for the feedback!

The package access constructor is intentional. It is used internally by one of the implementations of `AbstractUserDetailsServiceBeanDefinitionParser` which is in the same package as `CachingUserDetailsService`.

When the `BeanDefinitionParser` encounters a `cache-ref` attribute, it will automatically register a `CachingUserDetailsService`and set the delegate to the implementation of the provided/configured`UserDetailsService`.

I'm going to close this PR. If you need further clarification feel free to reach out.
Thanks.

PR #234, opened 2015-11-10
Fix sessionAuthenticationStrategy setter

didiez:
JIRA issue: https://jira.spring.io/browse/SEC-3143
SessionManagementConfigurer does not allow to inject a custom SessionAuthenticationStrategy due to a typo in the assignment to the member field.


jgrandja:
Thanks for the PR and that catch! This is now merged.

didiez:
Is it in line with the comment Rob Winch did on https://github.com/spring-projects/spring-security/issues/3341#issuecomment-180607631?
I finally workaround my problem avoiding the use of .maximumSessions(1)

jgrandja:
Thanks for pointing this out...Rob's comment.
Applying updates shortly.

PR #241, opened 2015-11-24
Ability to specify custom login form authentication processing filter

mkopylec:


rwinch:
@mkopylec Thank you for taking the time to submit a PR! I'm wondering if it makes sense to use the following instead:

```java
http
    .addFilter(usernamePasswordAuthenticationFilter)
```

Thoughts?

mkopylec:
The change is mainly for convinient and more readable and intuitive `HttpSecurity` configuration.
With the change you can do:
```java
http.formLogin(usernamePasswordAuthenticationFilter).loginPage("/login").usernameParameter("user").defaultSuccessUrl("/success")...
```
Pros:
- adding custom login filter to existing `HttpSecurity` configuration could not be easier
- there is no need to set all the filter dependencies manually (using setters etc.)
Cons:
- I don't see any :)

rwinch:
Thank you for your feedback.

I'm not sure that the tradeoff of the the additional complexity in the codebase vs value gained is worth it. There is already a mechanism that works for adding a custom filter that is a single line of code. Configuring the custom filter can already be done with the setter methods which is a one to one with the DSL methods.

My fear is that providing too many options, the DSL will start to get complicated. If you have done any PERL programming you will know what I mean when I say too much flexibility makes things complicated.

For these reasons, I'm going to close this as won't fix.

Of course if you still disagree, I'm open to discussion about it. Thank you for reaching out with this proposal and I hope to see additional ideas you might be able to bring to the table!

mkopylec:
Maybe you are right.
Basically I would like to notice that the whole security DSL lacks an ability to be extend. Maybe this is intended, but it would be really nice if all configurers (all classes that extends `AbstractHttpConfigurer`) were not final. Also `AbstractHttpConfigurer` could be public. With those changes no additional complexity is provided straight forward and the DSL gains an ability to be extend. 

What do you think about my proposition?

rwinch:
I'd consider making `AbstractHttpConfigurer` public, but the other classes are final intentionally as this helps to ensure I can refactor without breaking others. If there is good reason to make a specific configurer non-final I would also consider it.

In general, I would recommend using composition. For example, you can do something like done in [boot-examples/x-auth-security](https://github.com/joshlong/boot-examples/blob/master/x-auth-security/src/main/java/demo/xauth/XAuthTokenConfigurer.java). For example:

```java
public class XAuthTokenConfigurer extends SecurityConfigurerAdapter<DefaultSecurityFilterChain, HttpSecurity> {

	private UserDetailsService detailsService;

	public XAuthTokenConfigurer(UserDetailsService detailsService) {
		this.detailsService = detailsService;
	}
	
	public void init(HttpSecurity http) throws Exception {
		XAuthTokenFilter customFilter = new XAuthTokenFilter(detailsService);
		http.addFilterBefore(customFilter, UsernamePasswordAuthenticationFilter.class);
	}

        public static XAuthTokenConfigurer xauth(UserDetailsService uds) {
             return new XAuthTokenConfigurer(uds);
        }
}

@EnableWebSecurity
class WebSecurityConfig extends WebSecurityConfigurerAdapter {
    @Autowired
    UserDetailsService userDetailsService;

    @Override
        http
            .apply(xauth(userDetailsService));
    }
}
```

mraible:
@rwinch Is the code above the recommended way to configure a custom `UsernamePasswordAuthenticationFilter`? Is it possible to replace the filter entirely, or only to add one before? I found this example that shows how to replace it entirely, but not sure if it's recommended.

http://stackoverflow.com/a/7728249/65681

rwinch:
@mraible Thanks for reaching out. 

If your Filter is an `instanceof` one of the known Spring Security Filters (i.e. `UsernamePasswordAuthenticationFilter`) you can use the following and it will be correctly inserted in the proper order:

```
http
   .addFilter(customFilter)
```

In terms of if you need the custom `SecurityConfigurerAdapter` that is really up to you. If you have the same code being applied in many places and want a custom DSL to apply it, then this makes sense. If you are just doing this once and want to get things done, then it is overkill.

Let me know if that answers your question or not.

PR #245, opened 2015-12-14
SEC-3170: Possible pitfall when using RememberMeServices

nikosk:
Added a null check in LogoutConfigurer.addLogoutHandler() method to ensure that a logout handler is always provided..


rwinch:
Thanks for the PR! I merged it into master via b28c62a6fe9b7f12a6cd152d190ec083412e1d6a

I also applied a bit of polish via 337f1885eacdbc9319e9b1fe4d319fadef761e78

PR #247, opened 2015-12-20
Fixed formatting

juzerali:
The code ticks was broken.

It looked like this
 `Authentication"request" object is built and then presented to the `AuthenticationManager`.

Now fixed to 
`Authentication` "request" object is built and then presented to the `AuthenticationManager`.


rwinch:
Thanks for the report and the PR! This is now fixed in master via 1f32e96d31c6a149e480ebdb5ce0274889877995

PR #250, opened 2016-01-05
fix typo in doc

ziedzaiem:
target class is org.springframework.security.test.context.support.WithSecurityContextTestExecutionListener


rwinch:

PR #251, opened 2016-01-08
Fix logout url in doc

wimdeblauwe:
The default for logout is to redirect to `/login?logout` as is correctly written below (in note <3> )


jgrandja:
Thanks for that catch! This is now merged.

PR #252, opened 2016-01-11
SEC-3190: Add support for colons in remember-me token values

jaffadog:
https://jira.spring.io/browse/SEC-3190

We have an issue where remeber-me cookie token strings that contain a colon break the existing decoding strategy, which tokenizes on colons.  Note that this happens as a matter of course when using spring-social, which adopts a username convention of <provider>:<userId> (e.g. "twitter:123245").  Username is one of the token values stored in the remeber-me cookie.

This change urlencodes the individual tokens when creating the cookie string; and urldecodes them decoding the cookie and extracting the tokens; such that colons are now supported in the token values.  This also eliminates the need for some existing code in decodeCookie() that deals with openid tokens which contain urls, and thus colons.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/rememberme/AbstractRememberMeServices.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/rememberme/AbstractRememberMeServices.java
@@ -1,6 +1,10 @@
 package org.springframework.security.web.authentication.rememberme;
 
+import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Method;
+import java.net.URLDecoder;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 
 import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@jaffadog Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

pivotal-issuemaster:
@jaffadog Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal)!

jgrandja:
Thanks for submitting this PR @jaffadog.
I applied some very minor polish and merged these changes to master.

jgrandja:
I'm re-opening this PR as this change is actually non-passive. For example, all existing remember-me tokens will not work on next auto-login since the cookie tokens were not originally url encoded as this change expects that.

We can introduce this change in the 5.0 release but need to avoid this in 4.2 release. For now, we'll target 5.0 for this but if you can come up with a non-passive change then we can introduce it before.
Thanks.

rwinch:
Thanks for the PR @jaffadog! This is now merged into master via 832f5c39c1edcda739f9d7196625858d9b7c3d88

PR #253, opened 2016-01-12
SEC-3191 - Add support method to AuthenticationProvider

ractive:
Allow an AuthenticationProvider to decide if it supports a given
Authentication instance, e.g. by deciding on the username if it should
try authentication.


rwinch:
@ractive Thanks for the report & PR! Unfortunately we cannot add a new method to an interface as this breaks passivity.

Fortunately, you can have the AuthenticationProvider return null if it does not support the Authentication object.

PR #254, opened 2016-01-13
4.0.x

liuduosan:


rwinch:
Given the lack of context on this and the number of commits this looks like an accident.

PR #255, opened 2016-01-18
Enforce constant time comparison

avri-schneider:
To prevent side-channel leakage of actual password length, enforce constant time even when expectedLength != actualLength.

### DIFF
--- a/core/src/main/java/org/springframework/security/authentication/encoding/PasswordEncoderUtils.java
+++ b/core/src/main/java/org/springframework/security/authentication/encoding/PasswordEncoderUtils.java
@@ -32,7 +32,7 @@ class PasswordEncoderUtils {
 		}
 			
 		for (int i = 0; i < actualLength; i++) {
-			result |= expectedBytes[i % expectedLength] ^ actualBytes[i % actualLength];
+			result |= expectedBytes[i % (expectedLength!=0?expectedLength:1)] ^ actualBytes[i % actualLength];
 		}
 		return result == 0;
 	}
--- a/core/src/main/java/org/springframework/security/authentication/encoding/PasswordEncoderUtils.java
+++ b/core/src/main/java/org/springframework/security/authentication/encoding/PasswordEncoderUtils.java
@@ -22,13 +22,10 @@ class PasswordEncoderUtils {
 		byte[] actualBytes = bytesUtf8(actual);
 		int expectedLength = expectedBytes == null ? -1 : expectedBytes.length;
 		int actualLength = actualBytes == null ? -1 : actualBytes.length;
-		if (expectedLength != actualLength) {
-			return false;
-		}
 
-		int result = 0;
-		for (int i = 0; i < expectedLength; i++) {
-			result |= expectedBytes[i] ^ actualBytes[i];
+		int result = expectedLength != actualLength ? 1 : 0;
+		for (int i = 0; i < actualLength; i++) {
+			result |= expectedBytes[i % expectedLength] ^ actualBytes[i % actualLength];
 		}
 		return result == 0;
 	}
--- a/core/src/main/java/org/springframework/security/authentication/encoding/PasswordEncoderUtils.java
+++ b/core/src/main/java/org/springframework/security/authentication/encoding/PasswordEncoderUtils.java
@@ -21,10 +21,16 @@ class PasswordEncoderUtils {
 		byte[] expectedBytes = bytesUtf8(expected);
 		byte[] actualBytes = bytesUtf8(actual);
 		int expectedLength = expectedBytes == null ? 0 : expectedBytes.length;
-		int actualLength = actualBytes == null ? -1 : actualBytes.length;
+		int actualLength = actualBytes == null ? 0 : actualBytes.length;
 
 		int result = expectedLength != actualLength ? 1 : 0;
 		result |= ((expectedBytes == null && actualBytes != null) || (expectedBytes != null && actualBytes == null)) ? 1 : 0;
+		
+		if (expectedBytes == null) {
+			expectedBytes = new byte[1];
+			expectedBytes[0] = (byte) 0xFF; // value is ignored, just initializing.
+		}
+			
 		for (int i = 0; i < actualLength; i++) {
 			result |= expectedBytes[i % expectedLength] ^ actualBytes[i % actualLength];
 		}
--- a/core/src/main/java/org/springframework/security/authentication/encoding/PasswordEncoderUtils.java
+++ b/core/src/main/java/org/springframework/security/authentication/encoding/PasswordEncoderUtils.java
@@ -20,10 +20,11 @@ class PasswordEncoderUtils {
 	static boolean equals(String expected, String actual) {
 		byte[] expectedBytes = bytesUtf8(expected);
 		byte[] actualBytes = bytesUtf8(actual);
-		int expectedLength = expectedBytes == null ? -1 : expectedBytes.length;
+		int expectedLength = expectedBytes == null ? 0 : expectedBytes.length;
 		int actualLength = actualBytes == null ? -1 : actualBytes.length;
 
 		int result = expectedLength != actualLength ? 1 : 0;
+		result |= ((expectedBytes == null && actualBytes != null) || (expectedBytes != null && actualBytes == null)) ? 1 : 0;
 		for (int i = 0; i < actualLength; i++) {
 			result |= expectedBytes[i % expectedLength] ^ actualBytes[i % actualLength];
 		}
--- a/core/src/main/java/org/springframework/security/authentication/encoding/PasswordEncoderUtils.java
+++ b/core/src/main/java/org/springframework/security/authentication/encoding/PasswordEncoderUtils.java
@@ -22,15 +22,14 @@ class PasswordEncoderUtils {
 		byte[] actualBytes = bytesUtf8(actual);
 		int expectedLength = expectedBytes == null ? 0 : expectedBytes.length;
 		int actualLength = actualBytes == null ? 0 : actualBytes.length;
-
-		int result = expectedLength != actualLength ? 1 : 0;
+		byte[] tmpBytes = new byte[1];
+		int result = (expectedLength != actualLength) ? 1 : 0;
+		
+		tmpBytes[0] = (byte) 0xFF; // value is ignored, just initializing.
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@avri-schneider Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

pivotal-issuemaster:
@avri-schneider Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal)!

pivotal-issuemaster:
@avri-schneider Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

rwinch:
@avri-schneider Thanks for the PR!

I do see that the length is leaked. However, I'm finding it difficult to make a change to the code given cryptographers have recommended the existing code. You can see that [MessageDigest.isEqual(String,String)](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/java/security/MessageDigest.java#MessageDigest.isEqual%28byte%5B%5D%2Cbyte%5B%5D%29) also terminates early if the length does not match. This code is also reviewed by experts.

Do you have any evidence that this solution is recommended and what the practical implications are (other than leaking the length) if this isn't changed?

avri-schneider:
`public interface PasswordEncoder`
...

> /* This will generally be a one-way message digest such as MD5 or SHA, but may also be
	 * a plaintext variant which does no encoding at all, **but rather returns the same
	 * password it was fed. The latter is useful to plug in when the original password
	 * must be stored as-is**. */

In case you don't use a message digest - leaking the length of the password could be useful for an attacker.

rwinch:
@jtmelton Can you provide your thoughts on the idea and the proposed code changes?

jtmelton:
So I took a quick look. It appears the existing solution is directly from ( https://codahale.com/a-lesson-in-timing-attacks/ ), which is great. However, in the reference used in that article ( https://rdist.root.org/2009/05/28/timing-attack-in-google-keyczar-library/ ), the assumption is made that there are 2 HMACs being compared (the result of the operation). In this specific case the comparison _could_ be on what is effectively the raw input. If the plaintext password is input and compared, you could be leaking the length. A key point is made in the keyczar article that "This function is data-independent, except for revealing the total length of the correctMac string. Since this is not considered important to security, it is acceptable. Of course, this might not be true for another use of this same code, so it cannot be blindly used in other applications."

Having said all that, I may be missing some context. From what I can see, the existing code would allow an attacker to try different password lengths until the right length is determined (assuming the plaintext password encoder is being used). Does that seem accurate? If so, that's obviously an undesirable characteristic.

Lastly, I wanted to comment on the code of the proposed change. Pieces of that code appear too complex to me to be considered for acceptance. Given the crypto usage of this code, I would prefer that it is better documented, and that each line is performing a single step. In particular, the for loop should be altered for clarity. In addition, I would want some tests covering this change.

I am also tagging in @kwwall for this one as he's a better crypto guy than I am.

kwwall:
@avri-schneider @jtmelton Fortunately, Oracle fixed MessageDigest.isEquals() so it is no longer susceptible to timing side-channel attacks. It was fixed as of release JDK 1.6.0_17. So unless you intend on supporting really old versions of Java prior to that, I would suggest using MessageDigest.isEquals() for the comparison. If you need to support ancient end-of-life JDKs, you could do it the way that ESAPI did in the arrayCompare() static method in https://github.com/ESAPI/esapi-java-legacy/blob/2.1/src/main/java/org/owasp/esapi/crypto/CryptoHelper.java. However in the next release of ESAPI, I am either going to replace that code with a call to MessageDigest.isEquals() or just deprecate it and use MessageDigest.isEquals() internally. (My feeling on that is if you are still using a EOL JDK, you should at least be using the latest patched version. And if you are still using an unsupported EOL JDK, I think you have bigger troubles than subtle timing side-channel attacks against your crypto. Just sayin'.)

Also important to note that a timing side-channel attack of this specific type has a very small window of opportunity for an attack. The timing difference is so small that it falls into the noise range of general network jitter. Tim Morgan (tim.morgan@owasp.org) presented an excellent presentation last summer to our local OWASP group on this. He showed that while one could detect this subtle timing side-channel from the local host on which the software is running, it wasn't possible over a WAN spanning the Internet. (I don't recall if it were possible from a local LAN; it it was, it took lots of repetitions.) Now all this is NOT to say that the timing side-channel should not be fixed. Clearly it should. When dealing with frameworks especially, framework developers should address all known security vulnerabilities because one cannot possibly be aware of all your client's threat models.

-kevin

jtmelton:
Thanks @kwwall . Yeah, the Watson Ladd post I referenced around keyczar implies a timing attack is possible over the network, but it would take a lot of requests, and it would be less trustworthy than either local or LAN attacks.

ecbftw:
Hi, I'm Tim, the guy @kwwall mentioned.  The talk he is referring to that my brother and I gave can also be found here:
https://www.youtube.com/watch?v=KirTCSAvt9M&list=PLH15HpR5qRsXF78lrpWP2JKpPJs_AFnD7&index=27

My understanding of the research (not my research, but timing attack researchers who came before me) is that in most modern platforms, short-circuit string comparisons can leak timing information, but usually the comparison is performed in 32bit chunks, which makes it very hard to attack.  Also, the timing differences per character are sub-nanosecond IIRC.  Very very small.  So unless you're on the local network with a vulnerable string comparison running on a slow 8-bit processor (think IoT), then you're probably not vulnerable.  But yeah, as Kevin says, just fix it anyway, but don't stress about it too much.

You might also find these bits of research helpful:
  https://www.youtube.com/watch?v=MCSRJxhh0s0&list=PLStO1VqVBvmGyghpuCaosPGtrwVy5Fbqi&index=41
  https://www.youtube.com/watch?v=wZnzmRYDfRY&list=PLwibn_3po6c8nIzgDyWdLpohPWeZ7dIz5&index=95

cheers,
tim

rwinch:
Thank you for the feedback everyone.

It seems like I am getting a mixed understanding. It appears that we should be concerned with leaking the length of the password. A few suggestions were made to fix this including:

* MessageDigest.isEqual which also leaks the length by terminating early. See http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/java/security/MessageDigest.java#443
* CrytoHelper.arrayCompare also leaks the length. See https://github.com/ESAPI/esapi-java-legacy/blob/esapi-2.1.0.1/src/main/java/org/owasp/esapi/crypto/CryptoHelper.java#L333

So it would seem if we want to prevent the length from leaking, then we need to cleanup the existing PR.

ecbftw:
I have no idea how/where this function is used, so forgive my ignorance, but this may help reframe the discussion:

Shouldn't you be hashing all passwords anyway with something like bcrypt or scrypt?  So you shouldn't really ever be comparing a stored plaintext password to one provided by a client.  If you're just comparing password hashes with the implementations being discussed, then leaking the length is no big deal.  If you are comparing passwords directly, I think you have bigger problems. =\

rwinch:
@ecbftw Thanks for the reply:

> Shouldn't you be hashing all passwords anyway with something like bcrypt or scrypt? So you shouldn't really ever be comparing a stored plaintext password to one provided by a client.

Correct. However, Spring Security is a framework and we cannot control what users are doing. We really need to help protect them from themselves as much as possible. This goes into what @kwwall was [saying](https://github.com/spring-projects/spring-security/pull/255#issuecomment-255534548):

> When dealing with frameworks especially, framework developers should address all known security vulnerabilities because one cannot possibly be aware of all your client's threat models.

rwinch:
I've applied polish to simplify the code a little. See [d3685d8](https://github.com/spring-projects/spring-security/blob/d3685d89c5139e4411667f2e371ddb721ca566d7/core/src/main/java/org/springframework/security/authentication/encoding/PasswordEncoderUtils.java#L33-L46) for the end result.

jtmelton:
@rwinch do we have tests covering this? It seems after a quick glance that if expected were an empty string you'd get a divide by zero issue. Apart from that, the code looks cleaner and easier to understand.

rwinch:
@jtmelton Thanks for the feedback. I fixed this issue and added another test in e62596f36ddb31703eb0184216c3da8a6cf8cf11

PS: There were tests already in [PasswordEncoderUtilsTests](https://github.com/spring-projects/spring-security/blob/e62596f36ddb31703eb0184216c3da8a6cf8cf11/core/src/test/java/org/springframework/security/authentication/encoding/PasswordEncoderUtilsTests.java#L25), but they did not cover when expected was an empty string and actual was both non-null and non-empty.

avri-schneider:
A service may run in the cloud, on shared hosting.
In shared hosting, one can't guarantee that an attacker does not have an instance on the same LAN or even on the same host as the victim.

avri-schneider:
@kwwall <https://github.com/kwwall> Could you please refer to what @jmelton
wrote: "From what I can see, the existing code would allow an attacker to
try different password lengths until the right length is determined
(assuming the plaintext password encoder is being used). Does that seem
accurate? If so, that's obviously an undesirable characteristic.".

Victim running his service on cloud shared environment (no dedicated
server) - what prevents attacker from running his attacking code on the
same host?

jtmelton:
Awesome!

rwinch:
@avri-schneider I don't think anyone disagreed with you that this needed fixed. Admitedly I was hesitant on the implementation.

This is fixed in master now with a few changes from me. Please see [PasswordEncoderUtils](https://github.com/spring-projects/spring-security/blob/e62596f36ddb31703eb0184216c3da8a6cf8cf11/core/src/main/java/org/springframework/security/authentication/encoding/PasswordEncoderUtils.java) for the end result

kwwall:
These won't leak password length if used to compare *hashes*, which is what
you _should_ be doing.

-kevin
--
Blog: http://off-the-wall-security.blogspot.com/.   | Twitter: @KevinWWall
NSA: All your crypto bit are belong to us.

jmelton:
Mind of I get off this reply train? You want @jtmelton not jmelton.

rwinch:
Sorry jmelton. You can use the unsubscribe button so that you no longer receive notifications.

image

kwwall:
Tim Morgan's research, I think, concluded that this type of timing side
channel attack wasn't even feasible over a WAN. Maybe Tim can refer you to
his notes. Of course, it all depends on your threat model. Regardless, it
should be fixed.

-kevin
--
Blog: http://off-the-wall-security.blogspot.com/.   | Twitter: @KevinWWall
NSA: All your crypto bit are belong to us.

PR #256, opened 2016-01-25
SEC-3204 NullPointerException fix. UserAttributes which are null will be ignored, now.

g17:
Fix for SEC-3204.

The InMemoryUserDetailsManager might throw a NullPointerException in case a corrupt properties file containing the user details is loaded. It might be the case that the UserAttributes.getValue() method is invoked on a null reference. For more details please refer to https://jira.spring.io/browse/SEC-3204


jgrandja:
Thanks for the PR.

You are correct that a NPE will be thrown in the case of a corrupt properties file. However, if we apply the null check in `InMemoryUserDetailsManager` it will ultimately hide the real issue - the corrupt file.

I believe it would be beneficial to the user to be made aware that the properties file is corrupt by a thrown NPE. That way they can restore the file back to the expected format/state.

I'm going to close this PR. Please re-open if you feel there is more discussion around this.
Thank you.

g17:
IMHO the problem is not the corrupt input file but the input validation. In any case I would never expect an application to throw a ```NullPointerException```. I would consider that as a bug (or as bad practice). The pull request gracefully goes over a non-readable user detail. But of course a warning could be added in addition, there.

Issue #3697, opened 2016-02-11
SwitchUserFilter.ROLE_PREVIOUS_ADMINISTRATOR cannot be overridden


andrei-ivanov:
Using the switch user functionality, I'm trying to see if the current user has the PREVIOUS_ADMINISTRATOR role using a different prefix than the standard one.
This fails since the name is hardcoded with the prefix: `org.springframework.security.web.authentication.switchuser.SwitchUserFilter.ROLE_PREVIOUS_ADMINISTRATOR = "ROLE_PREVIOUS_ADMINISTRATOR"`

rwinch:
@andrei-ivanov Thanks for the submission! If this is something you are interested in, can you submit a PR to have a member variable for the previous administrator's role that defaults to  `SwitchUserFilter.ROLE_PREVIOUS_ADMINISTRATOR`. You would then provide a setter to allow overriding the default.

andrei-ivanov:
PR created.

andrei-ivanov:
BTW, why give up on Jira?
I find it much nicer than the GH issues tracker.

rwinch:
@andrei-ivanov Thank you for creating a PR! I have responded to the PR!

> BTW, why give up on Jira?
> I find it much nicer than the GH issues tracker.

Thank you for your feedback. I agree that JIRA has some advantages over GitHub issues. However, we have gotten lots of feedback that users prefer GitHub issues as they already have accounts, integrates better with their tooling, etc. I have also found that the projects using GitHub issues have a lot more community involvement which is something that we are looking for. I hope this makes sense :)

andrei-ivanov:
BTW, I think some mechanism would be nice to set the role prefix in a single place.
Currently I have this in 4 places:
```xml
<bean id="defaultMethodSecurityExpressionHandler" class="org.springframework.security.access.expression.method.DefaultMethodSecurityExpressionHandler">
	<property name="defaultRolePrefix" value="ROLE:" />
</bean>

<bean id="defaultWebSecurityExpressionHandler" class="org.springframework.security.web.access.expression.DefaultWebSecurityExpressionHandler">
	<property name="defaultRolePrefix" value="ROLE:" />
</bean>

<bean id="securityContextHolderAwareRequestFilter" class="org.springframework.security.web.servletapi.SecurityContextHolderAwareRequestFilter">
	<property name="rolePrefix" value="ROLE:" />
</bean>

[...]

PR #3700, opened 2016-02-12
Fixes gh-3697

andrei-ivanov:
Fixes gh-3697.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/switchuser/SwitchUserFilter.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/switchuser/SwitchUserFilter.java
@@ -123,6 +123,7 @@ public class SwitchUserFilter extends GenericFilterBean implements
 	private String targetUrl;
 	private String switchFailureUrl;
 	private String usernameParameter = SPRING_SECURITY_SWITCH_USERNAME_KEY;
+	private String switchAuthorityRole = ROLE_PREVIOUS_ADMINISTRATOR;
 	private SwitchUserAuthorityChanger switchUserAuthorityChanger;
 	private UserDetailsService userDetailsService;
 	private UserDetailsChecker userDetailsChecker = new AccountStatusUserDetailsChecker();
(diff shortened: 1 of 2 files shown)


web/src/test/java/org/springframework/security/web/authentication/switchuser/SwitchUserFilterTests.java
```diff
@@ -396,12 +403,14 @@ public void modificationOfAuthoritiesWorks() {
 	// SEC-1763
 	@Test
 	public void nestedSwitchesAreNotAllowed() throws Exception {
```
rwinch: Let's leave this test alone and add a new test for the new functionality.

web/src/main/java/org/springframework/security/web/authentication/switchuser/SwitchUserFilter.java
```diff
@@ -528,6 +529,15 @@ public void setUsernameParameter(String usernameParameter) {
 	}
 
 	/**
+	 * Allows the role of the switchAuthority to be customized.
+	 * 
+	 * @param switchAuthorityRole the role name. Defaults to {@link #ROLE_PREVIOUS_ADMINISTRATOR}
+	 */
+	public void setSwitchAuthorityRole(String switchAuthorityRole) {
```
rwinch: Can you add an Assert.notNull check here? Please also ensure to add a test for this.

rwinch:
Thanks for the fast PR! I have added feedback to the code.

web/src/main/java/org/springframework/security/web/authentication/switchuser/SwitchUserFilter.java
andrei-ivanov: Done. I was just following the example of the other setters.

andrei-ivanov:
I've updated the PR.

rwinch:
Thanks for the PR! this is now merged into master via 9008a7a

web/src/main/java/org/springframework/security/web/authentication/switchuser/SwitchUserFilter.java
rwinch: Not a problem.

As an FYI...We try to have an assertion for any setter on a property that has a default value. There may be some older setters that don't follow this convention. Also if the default value of the property is null, then there is no need for the assertion.

PR #3705, opened 2016-02-18
Log AuthenticationException on unsuccessful authentication

liweinan0423:
I suggest we log the exception object for more effective debugging. 

I am currently troubleshooting an SAML authentication issue in my project, I was not able to see the detailed stack trace and root exception in console log but I had to debug through the spring source code, which was not very convenient.


jgrandja:
Thanks for the PR! This is now merged to master.

Issue #2571, opened 2013-09-25
SEC-2342: DefaultPermissionGrantingStrategy directly compares permission mask


spring-issuemaster:
[Igor Artamonov](https://jira.spring.io/secure/ViewProfile.jspa?name=iartamonov) (Migrated from [SEC-2342](https://jira.spring.io/browse/SEC-2342?redirect=false)) said:

`DefaultPermissionGrantingStrategy.isGranted` uses direct comparison for the permission `mask`:

```
//line :68
if ((ace.getPermission().getMask() == p.getMask()) && ace.getSid().equals(sid)) {
```

So, for:
- composite mask "_RW_" (`ace` mask is 3) 
- when we require "_R_" (`p` mask is 1) permission 

`isGranted` will return `false`.

I believe there should be `&` used, instead of `==`

ndouba:
+1

rwinch:
@ndouba Thanks for the input and submitting the PR!

This is related if not a duplicate of https://github.com/spring-projects/spring-security/issues/1388#issuecomment-180595870

I'm going to need some time to look into some of the history behind the issue. Also, (assuming we accept the change) I'm a bit hesitant to update existing code in a non-passive way. We may need to add a flag to the class to enable the functionality or create another class that performs the operation with the updates.

ndouba:
Maybe a better alternative to my PR is to create a new interface (PermissionComparisonStrategy) which allows people to inject their own permission comparison into the DefaultPermissionGrantingStrategy bean? I can also see how some people may want to implement their own version of it (i.e. mask1 > mask2, or if mask1 == ADMINISTER || (mask1 & mask2) != 0, etc.).

kortatu:
I've done the same as @ndouba , copy the code in my PermissionGrantingStrategy but executing a bitwise and operation and comparing with 0.

eleftherias:
Fixed by gh-5908, which allows overriding `isGranted` method in `DefaultPermissionGrantingStrategy` for adding custom permission comparison.

PR #3708, opened 2016-02-19
Use bitwise AND to check for perm mask intersect

ndouba:
Otherwise, for example, an ACE that has `RWD` will not pass for a permissions check of `R`. Using the strategy proposed you can now consolidate multiple permissions into one mask instead of having multiple permissions per permissions bit.


rwinch:
I'm closing this PR for the reasons described in https://github.com/spring-projects/spring-security/issues/2571#issuecomment-186021807

Issue #3702, opened 2016-02-16
ScryptPasswordEncoder


rwinch:
An idea is https://github.com/wg/scrypt

NOTE: There is other password storage advice here https://paragonie.com/blog/2016/02/how-safely-store-password-in-2016

rwinch:
PR submitted at #3717

PR #3717, opened 2016-02-26
Add SCryptPasswordEncoder

shazin:

### DIFF
--- a/crypto/crypto.gradle
+++ b/crypto/crypto.gradle
@@ -12,5 +12,6 @@ configure(project.tasks.withType(Test)) {
 }
 
 dependencies {
-    compile 'org.bouncycastle:bcprov-jdk15on:1.54'
+    optional 'org.bouncycastle:bcprov-jdk15on:1.54'
+    testCompile 'org.assertj:assertj-core:3.3.0'
 }
--- a/crypto/crypto.gradle
+++ b/crypto/crypto.gradle
@@ -10,3 +10,7 @@ configure(project.tasks.withType(Test)) {
 		exclude '**/EncryptorsTests.class'
 	}
 }
+
+dependencies {
+    compile 'org.bouncycastle:bcprov-jdk15on:1.54'
+}
--- a/crypto/src/main/java/org/springframework/security/crypto/scrypt/SCryptPasswordEncoder.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/scrypt/SCryptPasswordEncoder.java
@@ -15,13 +15,18 @@
  */
 package org.springframework.security.crypto.scrypt;
 
+import java.util.Base64;
+import java.util.Base64.Decoder;
+import java.util.Base64.Encoder;
+
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import org.springframework.security.crypto.codec.Hex;
 import org.springframework.security.crypto.codec.Utf8;
+import org.springframework.security.crypto.keygen.BytesKeyGenerator;
 import org.springframework.security.crypto.keygen.KeyGenerators;
 import org.springframework.security.crypto.password.PasswordEncoder;
 import org.bouncycastle.crypto.generators.SCrypt;
+
 
 
 /**
@@ -39,24 +44,24 @@ public class SCryptPasswordEncoder implements PasswordEncoder {
     
     private final int memoryCost;
     
-    private final int parallelization;
+    private final int parallelization;  
     
-    private final byte[] saltBytes;
+    private final int keyLength;
+    
+    private final BytesKeyGenerator saltGenerator;
     
     public SCryptPasswordEncoder() {
-        this(new String(KeyGenerators.secureRandom().generateKey()), 16384, 8, 1);
+        this(16384, 8, 1, 32, 64);
     }
     
     /**
-     * @param salt value for the algorithm
      * @param cpu cost of the algorithm. must be power of 2 greater than 1
      * @param memory cost of the algorithm
      * @param parallelization of the algorithm
+     * @param salt length
+     * @param key length for the algorithm
      */
-    public SCryptPasswordEncoder(CharSequence salt, int cpuCost, int memoryCost, int parallelization) {
-        if(salt == null || salt.length() == 0) {
-            throw new IllegalArgumentException("Secret is required to get salt value");
-        }
+    public SCryptPasswordEncoder(int cpuCost, int memoryCost, int parallelization, int keyLength, int saltLength) {
         if (cpuCost <= 1) {
             throw new IllegalArgumentException("Cpu cost parameter must be > 1.");
         }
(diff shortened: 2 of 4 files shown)


rwinch:
Thanks for the PR!

rwinch:
@shazin Looking at the referenced library with a bit more scrutiny, I'm a bit hesitant to use it. Among other things, the library has not had a commit in 3 years which is a bit scary for me.

It might be nice to explore alternatives. For example, [Bouncy Castle's SCrypt](http://bouncy-castle.1462172.n4.nabble.com/Java-Bouncy-Castle-scrypt-implementation-td4656832.html)

rwinch:
This resolves #3702

shazin:
Thanks for the review @rwinch. Will make the changes and push.

rwinch:
@shazin Thanks for the fast response. Another thing to point out is that the dependencies should go into the crypto.gradle (Maven poms are generated) and should be optional.

shazin:
Changes with Bouncy Castle Implementation of SCrypt pushed.

rwinch:
@shazin Thanks for the updates!

crypto/crypto.gradle
```diff
@@ -10,3 +10,7 @@ configure(project.tasks.withType(Test)) {
 		exclude '**/EncryptorsTests.class'
 	}
 }
+
+dependencies {
+    compile 'org.bouncycastle:bcprov-jdk15on:1.54'
```
rwinch: This should be optional. Also don't forget to regenerate the pom.xml using the following from this directory

../gradlew generatePom

rwinch:
Please make sure you use tabs not spaces (there are some spaces in the files)

crypto/src/test/java/org/springframework/security/crypto/scrypt/SCryptPasswordEncoderTests.java
```diff
@@ -0,0 +1,101 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.crypto.scrypt;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
```
rwinch: Please use assertj for your assertions. 

See http://joel-costigliola.github.io/assertj/

crypto/src/main/java/org/springframework/security/crypto/scrypt/SCryptPasswordEncoder.java
```diff
@@ -0,0 +1,104 @@
...
+    private final int cpuCost;
+    
+    private final int memoryCost;
+    
+    private final int parallelization;
+    
+    private final byte[] saltBytes;
+    
+    public SCryptPasswordEncoder() {
+        this(new String(KeyGenerators.secureRandom().generateKey()), 16384, 8, 1);
```
rwinch: Where did you get your default values? What do they imply (i.e. how much memory is used, how much time does it take on your CPU, etc)?

crypto/src/main/java/org/springframework/security/crypto/scrypt/SCryptPasswordEncoder.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+    
+    /**
+     * @param salt value for the algorithm
+     * @param cpu cost of the algorithm. must be power of 2 greater than 1
+     * @param memory cost of the algorithm
+     * @param parallelization of the algorithm
+     */
+    public SCryptPasswordEncoder(CharSequence salt, int cpuCost, int memoryCost, int parallelization) {
```
rwinch: Salt should be secure random generated for each individual password. That means we should not pass it into the constructor. Instead we should allow for a salt length to be injected and the salt should be added as a prefix of the password. Something like:

<salt><password>

we can then verify the password by extracting the salt using the number of bytes to extract out the salt that was generated.

crypto/src/main/java/org/springframework/security/crypto/scrypt/SCryptPasswordEncoder.java
```diff
@@ -0,0 +1,104 @@
...
+        
+        this.saltBytes = Utf8.encode(salt);
+        this.cpuCost = cpuCost;
+        this.memoryCost = memoryCost;
+        this.parallelization = parallelization;
+    }
+
+	@Override
+	public String encode(CharSequence rawPassword) {
+        return new String(Hex.encode(SCrypt.generate(Utf8.encode(rawPassword), saltBytes, cpuCost, memoryCost, parallelization, saltBytes.length)));
```
rwinch: Ideally we should be encoding with the parameters in the encoded value (along with the salt as mentioned above). This allows for increasing the strength as computing resources get better.

crypto/src/test/java/org/springframework/security/crypto/scrypt/SCryptPasswordEncoderTests.java
```diff
@@ -0,0 +1,101 @@
...
+    @Test(expected = IllegalArgumentException.class)
+    public void invalidMemoryCostParameter() {
+        new SCryptPasswordEncoder(UUID.randomUUID().toString(), 2, Integer.MAX_VALUE, 2);     
+    }
+    
+    @Test(expected = IllegalArgumentException.class)
+    public void invalidParallelizationParameter() {
+        new SCryptPasswordEncoder(UUID.randomUUID().toString(), 2, 8, Integer.MAX_VALUE);     
+    }
+
```
rwinch: Add a test that verifies update of password strength. Pseudocode below

```
SCryptPasswordEncoder oldEncoder = new SCryptPasswordEncoder(... some arguments ...);
SCryptPasswordEncoder newEncoder = new SCryptPasswordEncoder(... all different arguments ...);

String password = "secret";
String oldEncodedPassword = oldEncoder.encode(password);
assertThat(newEncoder.matches(password, oldEncodedPassword)).isTrue();
```

rwinch: Add a test that encoding the same password twice results in different results (due to a random salt being used).

```java
SCryptPasswordEncoder encoder = new SCryptPasswordEncoder();
String password = "secret";
assertThat(encoder.encode(password)).isNotEqualTo(encoder.encode(secret));
```

shazin:
Pushed the required changes. All Except;

Default parameter values were taken based on the following Scrypt document. 16384, 8, 1, 64

https://www.tarsnap.com/scrypt/scrypt.pdf

As per the guidelines Spaces must be used not tabs. Correct me if I am wrong.

https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.md

rwinch:
@shazin Thanks for the response!

> Default parameter values were taken based on the following Scrypt document. 16384, 8, 1, 64

I reviewed the default parameters and they seem to make sense. They use 16mb+ of memory so they should be considered secure.

> As per the guidelines Spaces must be used not tabs. Correct me if I am wrong.

You are right, but the document is wrong! We have switched to tabs :) I updated the document

I went ahead and merged your changes 7d02e259df9e5ac367c237599503e965fd0b4eb6 and applied a bit of polish on it via fc75a679d9e8a8bef4e1d029659637b62b420ce0

Thanks again for your contribution and fast responses!

rwinch:
@shazin Btw...In the release notes I'd like to mention that you added this feature. Do you have a preferred link to use (i.e. link to GitHub profile, Twitter profile, site, etc)

shazin:
Thanks for the final polish and merge. My twitter is https://twitter.com/shazinsadakath

Issue #3722, opened 2016-03-01
Javadoc on ProviderManager.authenticate clarification


rwinch:
The javadoc on `ProviderManger` should make it clear that an `AuthenticationException` from an `AuthenticationProvider` will cause additional `AuthenticationProvider`s to be tried.

PR #3724, opened 2016-03-02
Fix Javadoc on ProviderManager.authenticate

hmolsen:
https://github.com/spring-projects/spring-security/issues/3722


rwinch:
Thanks for the PR! This is merged into master via  b248eae

PR #3744, opened 2016-03-13
User Random instead of SecureRandom

zhanhb:


rwinch:
Can you explain why you think this change is good? I don't think we want to make the API prone to errors (i.e. using an instance of Random vs SecureRandom).

zhanhb:
The design should be as top of the class as it can be. Maybe someone want to use TheadLocalRandom.current() instead of SecurityRandom.

rwinch:
@zhanhb Thank you for the response. 

We do not want to allow use of `ThreadLocalRandom` as this would not be cryptographically secure. From the [ThreadLocalRandom javadoc](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ThreadLocalRandom.html):

> Instances of `ThreadLocalRandom` are **not cryptographically secure**.  Consider instead using `java.security.SecureRandom` in security-sensitive applications. 

For this reason, I'm closing this as won't fix.

PR #3763, opened 2016-03-17
Add Content Security Policy Support

pivotal-joe-grandja:
- Adds CSP support to the Java DSL of HeadersConfigurer
- Implementation is pretty straight forward but the documentation is important here so please review and provide feedback on whether it's sufficient

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
@@ -27,11 +27,7 @@ import org.springframework.security.config.annotation.web.builders.HttpSecurity;
 import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
 import org.springframework.security.web.header.HeaderWriter;
 import org.springframework.security.web.header.HeaderWriterFilter;
-import org.springframework.security.web.header.writers.CacheControlHeadersWriter;
-import org.springframework.security.web.header.writers.HpkpHeaderWriter;
-import org.springframework.security.web.header.writers.HstsHeaderWriter;
-import org.springframework.security.web.header.writers.XContentTypeOptionsHeaderWriter;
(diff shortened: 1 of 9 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
```diff
@@ -658,6 +656,79 @@ private HpkpConfig enable() {
...
+		 * @param policyDirectives the security policy directive(s) to write in the response.
+		 *
+		 * @return the {@link ContentSecurityPolicyConfig} for additional configuration
+		 */
+		public ContentSecurityPolicyConfig directives(String policyDirectives) {
+			if (writer == null) {
+				this.writer = new ContentSecurityPolicyHeaderWriter(policyDirectives);
+			} else {
+				// TODO Allow for over-riding previously set directives? Is this really needed?
+				this.writer.setPolicyDirectives(policyDirectives);
```
rwinch: I think this should be removed. User's can always invoke `contentSecurityPolicy` again.

rwinch:
@pivotal-joe-grandja Thanks for the PR! 

I would try to add a link to one or more of the more popular CSP guides (http://www.html5rocks.com/en/tutorials/security/content-security-policy/ or https://developer.mozilla.org/en-US/docs/Web/Security/CSP are nice)

We should also update the [headers section](https://github.com/spring-projects/spring-security/blob/41c6a797c3303b78a6f1ac6d82efeae3773f680c/docs/manual/src/docs/asciidoc/index.adoc#headers) of the reference. 

Also, we will likely merge all of the CSP stuff at once so feel free to update this PR and squash commits as you get the XML support and documentation updated.

pivotal-joe-grandja:
Ok thanks Rob.
I'll apply those updates.

rwinch:
This will fix gh-3763

rwinch:
@pivotal-joe-grandja Thanks for the updates! This is now merged into master via 2f7f2ff5896cbc1e8fbc4e6577c32c3176d6d0c4

PR #3764, opened 2016-03-17
Authentication Success and Failure Handlers in AbstractPreAuthenticatedProcessingFilter

### LINKED ISSUE #3389: SEC-3183: Add AuthenticationSuccessHandler support to AbstractPreAuthenticatedProcessingFilter
[Tomoyuki Ikeya](https://jira.spring.io/secure/ViewProfile.jspa?name=ikeyat) (Migrated from [SEC-3183](https://jira.spring.io/browse/SEC-3183?redirect=false)) said:

`AbstractPreAuthenticatedProcessingFilter` is useful to support SSO authentication that add principals in HTTP header.
It has authentication success handler method `successfulAuthentication()` and triggers `InteractiveAuthenticationSuccessEvent`. But exiting `AuthenticationSuccessHandler` interface is more subclass-friendly, accessible and natural for Spring Security users. It should support `AuthenticationSuccessHandler` and `AuthenticationFailureHandler` for its extensibility.

shazin:
Fixes #3389

### DIFF
--- a/config/src/test/groovy/org/springframework/security/config/annotation/web/configurers/NamespaceHttpX509Tests.groovy
+++ b/config/src/test/groovy/org/springframework/security/config/annotation/web/configurers/NamespaceHttpX509Tests.groovy
@@ -63,7 +63,7 @@ public class NamespaceHttpX509Tests extends BaseSpringSpec {
 			request.setAttribute("javax.servlet.request.X509Certificate", [certificate] as X509Certificate[] )
 			springSecurityFilterChain.doFilter(request, response, chain);
 		then:
-			response.status == 302
+			response.status == 200
 			authentication().name == 'rod'
 	}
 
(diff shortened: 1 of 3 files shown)


rwinch:
Thanks for the PR!

I'm curious what your scenario is for needing authentication success / failure handlers in pre authentication? Pre authenticated scenarios means that authentication has already occurred (i.e. using container based authentication, a proxy server like SiteMinder, etc). This means if there is success or failure in authentication it would happen outside of Spring Security (just as authentication did).

Another concern I have is that we cannot make non-passive changes to our default values. The changes in existing tests tells me that we need to restore the default behavior.

shazin:
One of the Reasons I worked on this issue after seeing this is that I actually had to override 	`unsuccessfulAuthentication `of `AbstractPreAuthenticatedProcessingFilter` in my previous job. 

The use case we have was indeed a Pre Authenticated Header Token but the Header Token was mapped to a user in the DB. And a user must be activated after registration before he is allowed to login into the system. In that use case we have to manually do this. 

Similar use cases may be applied for this change.

rwinch:
I am ok with adding the handlers, but the default behavior would need to be the same.

shazin:
The default behavior is restored with the ability have an Optional Authentication Success and Failure Handler.

rwinch:
Thanks for the PR! This is now merged into master via bd520099ee26c361b7f02ce43b519f82ee2b9d1e

Issue #3772, opened 2016-03-28
Document https://report-uri.io/


rwinch:
We should include https://report-uri.io/ in CSP documentation

**UPDATE** We should add this to the public key pinning docs too

PR #3773, opened 2016-03-30
Includes reference to https://report-uri.io/ in CSP and HPKP docs

jgrandja:
Includes a reference to the https://report-uri.io/ service in the CSP and HPKP documentation.

Fixes gh-3772


rwinch:
Thanks for the PR! This is now merged :)

Issue #3282, opened 2015-09-08
SEC-3103: Logout Success Content Negotiation


spring-issuemaster:
[Rob Winch](https://jira.spring.io/secure/ViewProfile.jspa?name=rwinch) (Migrated from [SEC-3103](https://jira.spring.io/browse/SEC-3103?redirect=false)) said:

Add default content negotiation for Logout Success (similar to AthenticationEntryPoint). This would mean if I logged out with a REST and/or AJAX call I would get an HTTP status of no content. If I logged out using the standard browser I would get a redirect.

PR #3777, opened 2016-03-31
Content Negotiating Logout Success Handler #3282

shazin:


rwinch:
@shazin Thanks for the PR!

I think I'd prefer to make something a little more flexible. Perhaps a RequestMatcherLogoutHandler would make more sense. The implementation would have a `LinkedHashMap` of `RequestMatcher` to `LogoutHandler` that is injected into it. The first `RequestMatcher` that is matched on would have its corresponding `LogoutHandler` invoked. There would also be a default `LogoutHandler` if no `RequestMatcher` were matched

shazin:
The requested changes are pushed.

rwinch:
@shazin Thanks for the fast response! A few more things:

* We should include some documentation updates
* Why the constructor? I think it is unlikely to be correct since users will probably want to customize the logout success URL.
* I think it is best to have a default `LogoutSuccessHandler` so that users know that something was executed.
* Sorry. I think I misguided you on the name of the class. To be consistent with the current architecture (i.e. `DelegatingAccessDeniedHandler`, `DelegatingAuthenticationEntryPoint`, etc) we should use the name `DelegatingLogoutSuccessHandler`

rwinch:
Thank you for the PR! I have included these changes along with the necessary polish in gh-3828

Closing in favor of gh-3828

PR #3779, opened 2016-04-01
BouncyCastle implementation of "AES/CBC/PKCS5Padding"

### LINKED ISSUE #2917: SEC-2778: Support Encryptors without needing to download JCE extensions
[Dave Syer](https://jira.spring.io/secure/ViewProfile.jspa?name=david_syer) (Migrated from [SEC-2778](https://jira.spring.io/browse/SEC-2778?redirect=false)) said:

Using the crypto module (AesBytesEncryptor in particular) requires the user to download JCE extension policy jars and install them in the JRE, which is a bit of a barrier to entry for some people. It would be good if there was a bouncy castle alternative, or a system property we can set or something (maybe already there)?

william-tran:
Fixes #2917 

I ran a very cursory JMH benchmark on encrypt/decrypt of a 1 MB byte array of random bytes and this implementation is 3x slower than JCE, and for a 1 kB byte array this is 4.7x slower.

### DIFF
--- a/crypto/src/main/java/org/springframework/security/crypto/encrypt/BouncyCastleAesBytesEncryptor.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/encrypt/BouncyCastleAesBytesEncryptor.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2011-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.crypto.encrypt;
+
+import java.io.ByteArrayOutputStream;
+
+import org.bouncycastle.crypto.PBEParametersGenerator;
+import org.bouncycastle.crypto.generators.PKCS5S2ParametersGenerator;
+import org.bouncycastle.crypto.io.CipherOutputStream;
+import org.bouncycastle.crypto.params.KeyParameter;
+import org.springframework.security.crypto.codec.Hex;
+import org.springframework.security.crypto.keygen.BytesKeyGenerator;
+import org.springframework.security.crypto.keygen.KeyGenerators;
+
+/**
+ * Base class for AES-256 encryption using Bouncy Castle.
+ *
+ * @author William Tran
+ *
+ */
+abstract class BouncyCastleAesBytesEncryptor implements BytesEncryptor {
+
+	final KeyParameter secretKey;
+	final BytesKeyGenerator ivGenerator;
+
+	BouncyCastleAesBytesEncryptor(String password, CharSequence salt) {
+		this(password, salt, KeyGenerators.secureRandom(16));
+	}
+
+	BouncyCastleAesBytesEncryptor(String password, CharSequence salt,
+			BytesKeyGenerator ivGenerator) {
+		if (ivGenerator.getKeyLength() != 16) {
+			throw new IllegalArgumentException("ivGenerator key length != block size 16");
+		}
+		this.ivGenerator = ivGenerator;
+		PBEParametersGenerator keyGenerator = new PKCS5S2ParametersGenerator();
+		byte[] pkcs12PasswordBytes = PBEParametersGenerator
+				.PKCS5PasswordToUTF8Bytes(password.toCharArray());
+		keyGenerator.init(pkcs12PasswordBytes, Hex.decode(salt), 1024);
+		this.secretKey = (KeyParameter) keyGenerator.generateDerivedParameters(256);
+	}
+
+	byte[] process(CipherOutputStream cipherOutputStream,
+			ByteArrayOutputStream byteArrayOutputStream, byte[] bytes) {
+		try {
+			cipherOutputStream.write(bytes);
+			// close() invokes the  doFinal method of the encapsulated cipher object
+			// and flushes to the underlying outputStream. It must be called before
+			// we get the output.
+			cipherOutputStream.close();
+			return byteArrayOutputStream.toByteArray();
+		}
+		catch (Throwable e) {
+			try {
+				// attempt release of resources
+				cipherOutputStream.close();
+			}
+			catch (Throwable e1) {
+			}
+			throw new IllegalStateException("unable to encrypt/decrypt", e);
+		}
+	}
+
+}
--- a/crypto/src/main/java/org/springframework/security/crypto/encrypt/BouncyCastleAesCbcBytesEncryptor.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/encrypt/BouncyCastleAesCbcBytesEncryptor.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2011-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.crypto.encrypt;
+
+import static org.springframework.security.crypto.util.EncodingUtils.concatenate;
+import static org.springframework.security.crypto.util.EncodingUtils.subArray;
+
+import java.io.ByteArrayOutputStream;
+
+import org.bouncycastle.crypto.engines.AESFastEngine;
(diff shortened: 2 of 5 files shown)


crypto/src/main/java/org/springframework/security/crypto/encrypt/BouncyCastleAesBytesEncryptor.java
```diff
@@ -0,0 +1,108 @@
...
+		PBEParametersGenerator keyGenerator = new PKCS5S2ParametersGenerator();
+		final byte[] pkcs12PasswordBytes = PBEParametersGenerator
+				.PKCS5PasswordToUTF8Bytes(password.toCharArray());
+		keyGenerator.init(pkcs12PasswordBytes, Hex.decode(salt), 1024);
+		secretKey = keyGenerator.generateDerivedParameters(256);
+		this.ivGenerator = ivGenerator;
+	}
+
+	public BouncyCastleAesBytesEncryptor(String password, CharSequence salt) {
+		this(password, salt, null);
```
rwinch: We should not use a null IV as this is not safe. We always need to create IV for AES/CBC.

crypto/src/main/java/org/springframework/security/crypto/encrypt/BouncyCastleAesBytesEncryptor.java
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	public byte[] encrypt(byte[] bytes) {
+		synchronized (this.blockCipher) {
```
rwinch: I wonder if this is part of the bottle neck you were encountering? Does Bouncy Castle recommend creating a new instance every time or to use synchronization?

crypto/src/main/java/org/springframework/security/crypto/encrypt/BouncyCastleAesBytesEncryptor.java
william-tran: This matches the behaviour of the NULL_IV_GENERATOR in AesBytesEncryptor, which is shown in BouncyCastleAesBytesEncryptorEquivalencyTest.bcJceWithoutIvEquivalent(), and Encryptors.queryableText produces this sort of encryptor.

rwinch: Yes that is also incorrect (but we cannot change).

crypto/src/main/java/org/springframework/security/crypto/encrypt/BouncyCastleAesBytesEncryptor.java
william-tran: The performance differences in BC vs JCE are known http://bouncy-castle.1462172.n4.nabble.com/Cipher-performance-comparision-td4656983.html and the benchmarks seem to agree... though I was only running a single thread for those benchmarks. 

What's interesting is the difference in performance in creating a new instance vs reusing one (1 thread, 10 sec warmup, 10 sec sample):

using a `new PaddedBufferedBlockCipher(new CBCBlockCipher(new AESFastEngine()), new PKCS7Padding());` per invocation: 
```
Benchmark                        Mode  Samples       Score      Error  Units
o.s.MyBenchmark.bcEncryptor     thrpt       10   26211.948 ±  958.095  ops/s
o.s.MyBenchmark.jceEncryptor    thrpt       10  124323.434 ± 2743.512  ops/s

```

reusing and synchronizing
```
Benchmark                        Mode  Samples       Score      Error  Units
o.s.MyBenchmark.bcEncryptor     thrpt       10   26593.366 ±  635.466  ops/s
o.s.MyBenchmark.jceEncryptor    thrpt       10  124363.429 ± 2640.243  ops/s
```

At that rate, I think I'll modify it and create new ones, as I imagine users use singletons of these encryptors.

crypto/src/main/java/org/springframework/security/crypto/encrypt/BouncyCastleAesBytesEncryptor.java
william-tran: Ok so I'll remove that constructor and expect an IV generator in the code. Should I also make it package scope and final and add a method to Encryptors?

rwinch: Let's leave it public. I think the static methods in Encryptors need to go eventually. The problem with them is that what is "standard", "stronger", etc changes with time. However we cannot modify these methods and stay passive.

william-tran:
I ran some more benchmarks comparing using a new instance of BlockCipher per invocation vs synchronized reuse of the BlockCipher. The test uses 4 threads sharing the same BouncyCastleAesBytesEncryptor, with 10 sec warmup, and 10 sec sampling.

synchronized reuse:
```
Benchmark                        Mode  Samples      Score      Error  Units
o.s.MyBenchmark.bcEncryptor     thrpt       10  21407.710 ±  855.481  ops/s
o.s.MyBenchmark.jceEncryptor    thrpt       10  92248.023 ± 2361.111  ops/s
```

new
```
Benchmark                        Mode  Samples      Score      Error  Units
o.s.MyBenchmark.bcEncryptor     thrpt       10  44368.225 ± 2895.580  ops/s
o.s.MyBenchmark.jceEncryptor    thrpt       10  92362.021 ± 5653.288  ops/s
```

That's 2x throughput than synchronized. I ran it off my dual core macbook air so the numbers make sense given 4 threads.

Then I tested the GC differences with a single thread running 1 million test cycles, against `-Xms1g -Xmx1g -XX:+UseParallelGC`. jvisualvm reports these numbers:
BC synchronized:
image

BC new:
image

JCE (synchronized):
image

All three show very similar minor GC counts (23, 25, 24), and actually JCE shows higher GC cpu time (50, 57, 81) ms. 

Given all this, I think using a new instance of the cipher per invocation is the way to go. So I think that's it, should I squash the commits?

rwinch:
A few additional comments:

* I think we should provide a constructor with a default IV generator that makes it easy for users
* We should likely provide a GCM mode since this is the preferred mechanism today

william-tran:
Github was having issues yesterday and seems to not be kicking off Travis... lemme see if I can make a few force pushed to get that going again

rwinch:
@william-tran Thanks for the updates! 

Looking at the code it appears my suggestion to use a base class might have caused more code/complexity than intended. I'm wondering if the base class makes sense. Do you think we could try to:

* Remove the need for OutputStreams
* Remove passing in a boolean for ENCRYPT or DECRYPT. To me this is confusing even if the BC APIs define it this way. I'd much rather have some duplicate code.

I think the best way we could do this is to create the two classes totally independent of one another (without thinking about the other code). If there are similarities those can be extracted into a package private superclass. If not, then they live in their own classes.

william-tran:
The whole need for OutputStreams is so that method can return two objects, a properly sized ByteArrayOutputStream and a properly initialized CipherOutputStream, then common code can process those. Both those objects depend on a BlockCipher that doesn't share a common subclass. (though have the same method names). The process method is a bit ugly due to the need to try/catch/finally/close streams so I didn't want to repeat that. While at first glance what you're asking for would seem clearer, it'll result in the duplication of a lot of complexity.

rwinch:
@william-tran Thanks for the response. I'm much prefer the duplication of the code to needing to add classes like `OutputStreams` which adds its own variation of complexity.

william-tran:
I prefer the previous revision, but what matters the most to me is that it works. We can entirely eliminate the common class by duplicating key generation and OutputStream processing, I only went half way.

crypto/src/main/java/org/springframework/security/crypto/encrypt/BouncyCastleAesBytesEncryptor.java
```diff
@@ -0,0 +1,77 @@
...
+		byte[] pkcs12PasswordBytes = PBEParametersGenerator
+				.PKCS5PasswordToUTF8Bytes(password.toCharArray());
+		keyGenerator.init(pkcs12PasswordBytes, Hex.decode(salt), 1024);
+		this.secretKey = (KeyParameter) keyGenerator.generateDerivedParameters(256);
+	}
+
+	byte[] process(CipherOutputStream cipherOutputStream, ByteArrayOutputStream byteArrayOutputStream, byte[] bytes) {
+		try {
+			cipherOutputStream.write(bytes);
+			cipherOutputStream.close();
```
rwinch: `cipherOutputStream` is closed in the finally block on line 69. Is it intentional to close `cipherOutputStream` twice?

william-tran: The second call could go in the catch instead of a finally. close() does a doFinal and releases resources, in case the write() throws exception we should try to release those resources

william-tran: moved the second close() to the catch block and added comments to clarify things.

crypto/src/main/java/org/springframework/security/crypto/encrypt/BouncyCastleAesBytesEncryptor.java
```diff
@@ -0,0 +1,78 @@
...
+			ByteArrayOutputStream byteArrayOutputStream, byte[] bytes) {
+		try {
+			cipherOutputStream.write(bytes);
+			// close() invokes the  doFinal method of the encapsulated cipher object
+			// and flushes to the underlying outputStream. It must be called before
+			// we get the output.
+			cipherOutputStream.close();
+			return byteArrayOutputStream.toByteArray();
+		}
+		catch (IOException e) {
```
william-tran: In case this isn't an IOException, we won't attempt to release resources and wrap in IllegalStateException. What about just catching any Exception?

rwinch:
Thanks for the hard work @william-tran! This is now merged into master

william-tran:
Woohoo! Thanks for guiding this to landing.

Issue #3781, opened 2016-04-01
Fix missing # in 23.2.2 Path Variables in Web Security Expressions


ksokol:
Spring Expression Language (SpEL) - [8.5.11 Variables](http://docs.spring.io/spring/docs/4.1.9.RELEASE/spring-framework-reference/htmlsingle/#expressions-ref-variables):

> Variables can be referenced in the expression using the syntax `#variableName`

Leading `#` has been left out in code example in [23.2.2 Path Variables in Web Security Expressions. ](http://docs.spring.io/spring-security/site/docs/4.1.0.BUILD-SNAPSHOT/reference/htmlsingle/#el-access-web-path-variables)

rwinch:
Thanks for the report and the PR! This is now fixed in master via 9c3db557ddd3b2927da071d196a3ba5e67879e54

Souchong:
Hello,
Thank you for this update. I think there is another correction : 
".authorizeUrls()" shouldn't it be replace with ".authorizeRequests()" in code example in 23.2.2 Path Variables in Web Security Expressions ?

I add the code of this example in my application but I have a problem. 
When "checkUserId(Authentication authentication, String userid)" is called, userid is always null.  
Here is my code : 
Access control : 
`        @Override
    protected void configure(HttpSecurity http) throws Exception {
[...]
http
        	.authorizeRequests()
.antMatchers("/{userid}/homes").access("@webSecurity.checkUserId(authentication,#userid)")
[...]
}
[...]
    @Bean
    public WebSecurity webSecurity() {
        return new WebSecurity();
    }  
`

`public class WebSecurity {
	public boolean checkUserId(Authentication authentication, String userid) {
		String uname = authentication.getName();
		return uname.equals(userid);
	}
}`
Any idea ? I pass one day to debug in spring core and spring security...but I do not find where is the problem.

Any help is welcome

rwinch:
@Souchong Thanks for the feedback!

> ".authorizeUrls()" shouldn't it be replace with ".authorizeRequests()" in code example in 23.2.2 Path Variables in Web Security Expressions ?

Good catch! I created #3875 to track this

> I add the code of this example in my application but I have a problem. 
> When "checkUserId(Authentication authentication, String userid)" is called, userid is always null.

[...]

PR #3782, opened 2016-04-01
SpEL variables can be referenced in the expression using the syntax #…

ksokol:
…variableName.

Leading # has been left out in code example in 23.2.2 Path Variables in Web Security Expressions.

Fixes gh-3781

I have signed and agree to the terms of the Spring Individual Contributor License Agreement.

### DIFF
--- a/docs/manual/src/docs/asciidoc/index.adoc
+++ b/docs/manual/src/docs/asciidoc/index.adoc
@@ -4851,7 +4851,7 @@ You could refer to the method using:
 ----
 <http>
 	<intercept-url pattern="/user/{userId}/**"
-		access="@webSecurity.checkUserId(authentication,userId)"/>
+		access="@webSecurity.checkUserId(authentication,#userId)"/>
 	...
 </http>
 ----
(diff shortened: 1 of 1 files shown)


ksokol:
Sorry, for the noise.

rwinch:
@ksokol Not sure I follow. At a quick glance this looks like a good PR. Any reason you felt it was noise and closed it?

ksokol:
@rwinch Sorry! I was a little bit confused right now. Everything looks good know.

rwinch:
@ksokol Not a problem :) Thanks for reporting the issue and submitting the PR! I'll wait for the build to pass and then plan on merging it.

Thanks again!!

ksokol:
I have to thank you for your great job on Spring Security!

rwinch:
Thanks again for the PR! This is now merged via 9c3db557ddd3b2927da071d196a3ba5e67879e54

Vaelyr:
Very useful feature for me and tried to use it. 

Discovered that camel-case identifiers seem not to be picked up by the expression handler. So the `userId` in the example would return null, assuming it is `Long`, while just `id` or `userid` is handled properly, or anything using only lower case. Maybe worth mentioning in the document or better yet to investigate?

rwinch:
@Vaelyr Thanks for the feedback! I think you have found a bug. I created #3786 to address this

PR #3785, opened 2016-04-05
Fix typo in xsd

### LINKED ISSUE #3229: SEC-3018: Fix typo in xsd
[Florian Waltersdorfer](https://jira.spring.io/secure/ViewProfile.jspa?name=florian.waltersdorfer) (Migrated from [SEC-3018](https://jira.spring.io/browse/SEC-3018?redirect=false)) said:

in the documentation section of the attribute "create-session"
"This differs from the use of "never" which {color:red}_mans_{color} that Spring Security will not create a session, but will make use of one if the application does."

below are plain (git diff) patch files that change the type, 3 for each seperate file, 1 for 3.1+3.2 and 1 for all together

Qkyrie:
…xsd, spring-security-4.0.xsd

Fixes #3229

[x] I have signed the CLA

### DIFF
--- a/config/src/main/resources/org/springframework/security/config/spring-security-3.1.rnc
+++ b/config/src/main/resources/org/springframework/security/config/spring-security-3.1.rnc
@@ -297,7 +297,7 @@ http.attlist &=
 http.attlist &=
     use-expressions?
 http.attlist &=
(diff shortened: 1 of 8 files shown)


rwinch:

Issue #3752, opened 2016-03-14
Restructure Samples


rwinch:
Structure the Samples

We should restructure the samples so they are easier to navigate. For example, we might do something like:

* Boot
  * Authentication
  * web
  * method
* Java
  * Authentication
  * web
  * method
* XML
  * Authentication
  * web
  * method

This is just a suggestion we can work out the details later

rwinch:
Thanks for the PR @jgrandja! This is now merged into master

PR #3799, opened 2016-04-08
Organizes the sample projects under sub-directories: javaconfig and xml

jgrandja:
Organizes the sample projects under sub-directories: javaconfig and xml

Fixes gh-3752


rwinch:
Thanks for the PR! Overall this looks good.

We need to be sure that [samples/xml/cas](https://github.com/jgrandja/spring-security/tree/master/samples/xml/cas), [samples/xml](https://github.com/jgrandja/spring-security/tree/master/samples/xml), and [samples/javaconfig](https://github.com/jgrandja/spring-security/tree/master/samples/javaconfig) are not projects. I noticed that each of these have their own pom.xml and will be imported into the IDE. This is not ideal because I think this will produce artifacts that are deployed and we do not really want to see them in the IDE.

jgrandja:
Ok I'll make the change to exclude those 3 mentioned projects.

jgrandja:
Updates have been applied to exclude those projects.

PR #3803, opened 2016-04-12
Fix typo in setMessageExpessionHandler

McNetic:
I noticed there is a small typo. However, I'm not sure the fix is correct like this.


rwinch:

Issue #3784, opened 2016-04-05
Add addFilterAt option


rwinch:
We should provide an option for inserting Spring Security Filter at a specific location. For example:

```
http
    .addFilterAt(customFilter, UsernamePasswordAuthenticationFilter.class)
```

PR #3809, opened 2016-04-13
Add HttpSecurity.addFilterAt

rwinch:
Fixes gh-3784

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/builders/FilterComparator.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/builders/FilterComparator.java
@@ -150,6 +150,23 @@ final class FilterComparator implements Comparator<Filter>, Serializable {
 	}
 
 	/**
+	 * Registers a {@link Filter} to exist at a particular {@link Filter} position
+	 * @param filter the {@link Filter} to register
+	 * @param atFilter the {@link Filter} that is already registered and that
+	 * {@code filter} should be placed at.
+	 */
+	public void registerAt(Class<? extends Filter> filter,
+			Class<? extends Filter> atFilter) {
+		Integer position = getOrder(atFilter);
+		if (position == null) {
+			throw new IllegalArgumentException(
+					"Cannot register after unregistered Filter " + atFilter);
+		}
+
+		put(filter, position);
+	}
+
+	/**
 	 * Registers a {@link Filter} to exist before a particular {@link Filter} that is
 	 * already registered.
 	 * @param filter the {@link Filter} to register
--- a/config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
@@ -1026,6 +1026,25 @@ public final class HttpSecurity extends
 	}
 
 	/**
+	 * Adds the Filter at the location of the specified Filter class. For example, if you
+	 * want the filter CustomFilter to be registered in the same position as
+	 * {@link UsernamePasswordAuthenticationFilter}, you can invoke:
+	 *
+	 * <pre>
+	 * addFilterAt(new CustomFilter(), UsernamePasswordAuthenticationFilter.class)
+	 * </pre>
+	 *
+	 * @param filter the Filter to register
+	 * @param atFilter the location of another {@link Filter} that is already registered
+	 * (i.e. known) with Spring Security.
+	 * @return the {@link HttpSecurity} for further customizations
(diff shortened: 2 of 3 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/builders/FilterComparator.java
```diff
@@ -150,6 +150,23 @@ public void registerAfter(Class<? extends Filter> filter,
 	}
 
 	/**
+	 * Registers a {@link Filter} to exist at a particular {@link Filter} position
+	 * @param filter the {@link Filter} to register
+	 * @param atFilter the {@link Filter} that is already registered and that
+	 * {@code filter} should be placed at.
+	 */
+	public void registerAt(Class<? extends Filter> filter,
```
jgrandja: The new registerAt method has the same implementation as the existing registerAfter method. Possible case for reuse.

config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
```diff
@@ -1026,6 +1026,26 @@ public HttpSecurity addFilter(Filter filter) {
...
+	 * </pre>
+	 *
+	 * @param filter the Filter to register
+	 * @param atFilter the location of another {@link Filter} that is already registered
+	 * (i.e. known) with Spring Security.
+	 * @return the {@link HttpSecurity} for further customizations
+	 */
+	public HttpSecurity addFilterAt(Filter filter, Class<? extends Filter> atFilter) {
+		this.comparitor.registerAt(filter.getClass(), atFilter);
+		this.filters.add(filter);
```
jgrandja: Both addFilterBefore() and addFilterAfter() call the addFilter() method to add the filter which also does a pre-check before adding. However, the new addFilterAt() method directly puts the filter in the filters member variable avoiding the extra check in addFilter(). Although the first call to comparitor.registerAt() (if it succeeds) would ensure that the addFilter check will succeed too. So the current implementation would work the same way. But it's not consistent with the other 2 methods of adding. Is there a reasoning behind this that I'm not seeing?

config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
```diff
@@ -1026,6 +1026,26 @@ public HttpSecurity addFilter(Filter filter) {
...
+	 * addFilterAt(new CustomFilter(), UsernamePasswordAuthenticationFilter.class)
+	 * </pre>
+	 *
+	 * @param filter the Filter to register
+	 * @param atFilter the location of another {@link Filter} that is already registered
+	 * (i.e. known) with Spring Security.
+	 * @return the {@link HttpSecurity} for further customizations
+	 */
+	public HttpSecurity addFilterAt(Filter filter, Class<? extends Filter> atFilter) {
+		this.comparitor.registerAt(filter.getClass(), atFilter);
```
jgrandja: Typo on comparitor....should be comparator

config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
rwinch: The check probaly isn't necessary, but I think it reads better and is consistent. I will update this

config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
rwinch: This is a separate issue as comparitor already existed

config/src/main/java/org/springframework/security/config/annotation/web/builders/FilterComparator.java
rwinch: This was a copy / paste error. The index should be `position` not `position + 1`. This is now fixed

jgrandja: Give this logic, if existing FilterA is at position 101
then the call registerAt(FilterB, FilterA)
would result in both filters at position 101
can you see any possible side effects of them being equal?

rwinch: Yes, but this is up to the users discretion. This is no different than a user stating to put a key / value pair in a Map if the key already exists

jgrandja: Ok sounds good just wanted to point that out.

Issue #3786, opened 2016-04-05
Camel case URI variables do not work properly


rwinch:
If a user leverages:

```
http
    .authorizeUrls()
        .antMatchers("/user/{userId}/").access("@check.access(#userId)")
```

It will not work. I believe this is due to the fact that we are ignoring case by default.

PR #3814, opened 2016-04-14
Support Camel case URI variables

rwinch:
Perviously there were issues with case insenstive patterns and URI
variables that contained upper case characters. For example, the pattern
"/user/{userId}" could not resolve the variable #userId Instead it was
forced to lowercase and #userid was used.

Now if the pattern is case insensitive then so is the variable. This means
that #userId will work as will #userid.

Fixes gh-3786

### DIFF
--- a/config/src/test/groovy/org/springframework/security/config/http/InterceptUrlConfigTests.groovy
+++ b/config/src/test/groovy/org/springframework/security/config/http/InterceptUrlConfigTests.groovy
@@ -137,6 +137,39 @@ class InterceptUrlConfigTests extends AbstractHttpConfigTests {
 		response.status == HttpServletResponse.SC_FORBIDDEN
 	}
 
+	def "gh-3786 intercept-url supports cammel case path variables"() {
+		setup:
+		MockHttpServletRequest request = new MockHttpServletRequest(method:'GET')
+		MockHttpServletResponse response = new MockHttpServletResponse()
+		MockFilterChain chain = new MockFilterChain()
+		xml.http('use-expressions':true) {
+			'http-basic'()
+			'intercept-url'(pattern: '/user/{userName}/**', access: "#userName == authentication.name")
+			'intercept-url'(pattern: '/**', access: "denyAll")
+		}
+		createAppContext()
+		login(request, 'user', 'password')
+		when: 'user can access'
+		request.servletPath = '/user/user/abc'
+		springSecurityFilterChain.doFilter(request,response,chain)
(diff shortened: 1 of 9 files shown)


config/src/test/groovy/org/springframework/security/config/http/InterceptUrlConfigTests.groovy
```diff
@@ -137,6 +137,39 @@ class InterceptUrlConfigTests extends AbstractHttpConfigTests {
...
 	}
 
+	def "gh-3786 intercept-url supports cammel case path variables"() {
+		setup:
+		MockHttpServletRequest request = new MockHttpServletRequest(method:'GET')
+		MockHttpServletResponse response = new MockHttpServletResponse()
+		MockFilterChain chain = new MockFilterChain()
+		xml.http('use-expressions':true) {
+			'http-basic'()
+			'intercept-url'(pattern: '/user/{userName}/**', access: "#userName == authentication.name")
```
jgrandja: Should we add another test where the uri pattern variable (userName) is lowercase referenced in the access check (access: #username)

config/src/test/java/org/springframework/security/config/annotation/web/configurers/AuthorizeRequestsTests.java
```diff
@@ -157,6 +157,47 @@ protected void configure(AuthenticationManagerBuilder auth) throws Exception {
 		}
 	}
 
+	// gh-3786
+	@Test
+	public void antMatchersPathVariablesCaseInsensitiveCamelCaseVariables() throws Exception {
+		loadConfig(AntPatchersPathVariablesCammelCaseVariables.class);
```
jgrandja: Typo: should be Ant**M**atchersPathVariablesCammelCaseVariables

config/src/test/groovy/org/springframework/security/config/http/InterceptUrlConfigTests.groovy
rwinch: This is more of a integration test. The unit tesst cover the edge cases

config/src/test/java/org/springframework/security/config/annotation/web/configurers/AuthorizeRequestsTests.java
rwinch: Thanks! I pushed a fix

jgrandja:
The for the PR @rwinch!

Issue #3409, opened 2016-02-02
SEC-3211: SecurityContextRequestPostProcessorSupport dirties context and causes side effects


spring-issuemaster:
[Martin Bonato](https://jira.spring.io/secure/ViewProfile.jspa?name=martin.bonato@prime-sign.com) (Migrated from [SEC-3211](https://jira.spring.io/browse/SEC-3211?redirect=false)) said:

Using SecurityMockMvcRequestPostProcessors like `authentication()`, `securityContext()`, `testSecurityContext()`, `user()`, etc. in unit tests might cause side effects on other tests in the same application context.

`SecurityMockMvcRequestPostProcessors.SecurityContextRequestPostProcessorSupport.save(SecurityContext, HttpServletRequest)` calls `WebTestUtils.setSecurityContextRepository(HttpServletRequest, SecurityContextRepository)` to set a `TestSecurityContextRepository` in the `SecurityContextPersistenceFilter` registered in the application context. This dirties the context, because a registered bean (`SecurityContextPersistenceFilter`) is modified such that it causes side effects.

Since the side effects are rather rare and seem to show up randomly, it took me a while to track down the issue.

Here is an example of two unit tests where the first unit test influences the second:
``` 
	@Test
	@WithMockUser
	public void test1() throws Exception {
		mvc.perform(get("/"))//
				.andExpect(authenticated().withRoles("USER"));
	}
	
	// fails if executed after test1 in the same context
	@Test
	public void test2() throws Exception {
		mvc.perform(get("/"))//
			.andExpect(unauthenticated());
	}
``` 

If test2() is executed in its own application context it succeeds, since the default `HttpSessionSecurityContexRepository` used by the `SecurityContextPersistenceFilter` does not store anonymous authentication tokens. However, if test2() is executed after test1() in the same context it fails, because @WithMocUser caused the `TestSecurityContextRepository` to be registered which stores anonymous authentication tokens.

rwinch:
@mbonato Thanks for reporting [SEC-3211](https://jira.spring.io/browse/SEC-3211?redirect=false). We migrated all our issues on GitHub.

I'm running this with the latest version of Spring Security and unable to reproduce an issue. Can you expand on the problem you are encountering?

mbonato:
@rwinch I've created a test case for the issue: https://github.com/primesign/spring-security-issue3409. The test case still fails with the current release candidate 4.1.0.RC1.

rwinch:
[...]

PR #3818, opened 2016-04-14
Fix unauthenitcated() and AnonymousAuthenticationToken

rwinch:
Previously if unauthenticated() experienced an AnonymousAuthenticationToken
it would not match.

This commit ensures that if the user is anonymous (not just null)
unauthenticated() works.

Fixes gh-3409

### DIFF
--- a/test/src/test/java/org/springframework/security/test/web/servlet/response/Gh3409Tests.java
+++ b/test/src/test/java/org/springframework/security/test/web/servlet/response/Gh3409Tests.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2012-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.test.web.servlet.response;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.config.annotation.web.builders.HttpSecurity;
+import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
+import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
+import org.springframework.security.core.context.SecurityContextImpl;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+import org.springframework.test.context.web.WebAppConfiguration;
+import org.springframework.test.web.servlet.MockMvc;
+import org.springframework.test.web.servlet.setup.MockMvcBuilders;
+import org.springframework.web.context.WebApplicationContext;
+import org.springframework.web.servlet.config.annotation.EnableWebMvc;
+
+import static org.springframework.security.test.web.servlet.request.SecurityMockMvcRequestPostProcessors.securityContext;
+import static org.springframework.security.test.web.servlet.response.SecurityMockMvcResultMatchers.unauthenticated;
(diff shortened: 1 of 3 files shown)


test/src/test/java/org/springframework/security/test/web/servlet/response/Gh3409Tests.java
```diff
@@ -0,0 +1,130 @@
...
+			.with(securityContext(new SecurityContextImpl())));
+
+		this.mockMvc
+			.perform(get("/public/"))
+			.andExpect(unauthenticated());
+		// @formatter:on
+	}
+
+	@Test
+	public void unauthenitcatedNullAuthenitcation() throws Exception {
```
jgrandja: Typo on method name

test/src/test/java/org/springframework/security/test/web/servlet/response/Gh3409Tests.java
```diff
@@ -0,0 +1,130 @@
...
+		this.mockMvc = MockMvcBuilders
+			.webAppContextSetup(this.context)
+			.apply(springSecurity())
+			.build();
+		// @formatter:on
+	}
+
+	// gh-3409
+	@Test
+	public void unauthenitcatedAnonymousUser() throws Exception {
```
jgrandja: Typo on method name

test/src/test/java/org/springframework/security/test/web/servlet/response/Gh3409Tests.java
rwinch: Thanks! I pushed a fix

test/src/test/java/org/springframework/security/test/web/servlet/response/Gh3409Tests.java
rwinch: Thanks! I pushed a fix

test/src/test/java/org/springframework/security/test/web/servlet/response/Gh3409Tests.java
```diff
@@ -0,0 +1,130 @@
...
+					.and()
+				.formLogin().and()
+				.httpBasic();
+			// @formatter:on
+
+		}
+
+		// @formatter:off
+		@Autowired
+		public void configureGlobal(AuthenticationManagerBuilder auth) throws Exception {
```
jgrandja: The authManager configuration is redundant?

test/src/test/java/org/springframework/security/test/web/servlet/response/Gh3409Tests.java
```diff
@@ -0,0 +1,130 @@
...
+			auth
+				.inMemoryAuthentication()
+					.withUser("user").roles("USER","SELLER").password("password");
+		}
+		// @formatter:on
+
+		@RestController
+		static class Controller {
```
jgrandja: The Controller is redundant?

test/src/test/java/org/springframework/security/test/web/servlet/response/Gh3409Tests.java
rwinch: It is redundant to what?

test/src/test/java/org/springframework/security/test/web/servlet/response/Gh3409Tests.java
rwinch: It is redundant to what?

test/src/test/java/org/springframework/security/test/web/servlet/response/Gh3409Tests.java
jgrandja: I meant redundant in the sense that the code is not needed to pass the test

test/src/test/java/org/springframework/security/test/web/servlet/response/Gh3409Tests.java
jgrandja: I meant redundant in the sense that the Controller is not needed to pass the test

test/src/test/java/org/springframework/security/test/web/servlet/response/Gh3409Tests.java
rwinch: Thanks for the clarification. I pushed an update

test/src/test/java/org/springframework/security/test/web/servlet/response/Gh3409Tests.java
rwinch: Thanks for the clarification. I pushed an update

jgrandja:
Thanks @rwinch! It is now merged.

PR #3394, opened 2016-04-15
SEC-3189: By default use any RoleHierarchy Bean for SecurityExpressionHandler

rwinch:
Migrated from [SEC-3189](https://jira.spring.io/browse/SEC-3189?redirect=false)

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
@@ -20,10 +20,12 @@ import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.List;
 
+import org.springframework.context.ApplicationContext;
 import org.springframework.security.access.AccessDecisionVoter;
 import org.springframework.security.access.ConfigAttribute;
 import org.springframework.security.access.SecurityConfig;
 import org.springframework.security.access.expression.SecurityExpressionHandler;
+import org.springframework.security.access.hierarchicalroles.RoleHierarchy;
 import org.springframework.security.authentication.AuthenticationTrustResolver;
 import org.springframework.security.config.annotation.ObjectPostProcessor;
 import org.springframework.security.config.annotation.web.HttpSecurityBuilder;
@@ -188,6 +190,14 @@ public final class ExpressionUrlAuthorizationConfigurer<H extends HttpSecurityBu
 			if (trustResolver != null) {
 				defaultHandler.setTrustResolver(trustResolver);
 			}
+			ApplicationContext context = http.getSharedObject(ApplicationContext.class);
+			if(context != null) {
+				String[] roleHiearchyBeanNames = context.getBeanNamesForType(RoleHierarchy.class);
+				if(roleHiearchyBeanNames.length == 1) {
+					defaultHandler.setRoleHierarchy(context.getBean(roleHiearchyBeanNames[0], RoleHierarchy.class));
+				}
+			}
+
 			expressionHandler = postProcess(defaultHandler);
 		}
 
--- a/config/src/test/groovy/org/springframework/security/config/annotation/method/configuration/GlobalMethodSecurityConfigurationTests.groovy
+++ b/config/src/test/groovy/org/springframework/security/config/annotation/method/configuration/GlobalMethodSecurityConfigurationTests.groovy
@@ -15,6 +15,10 @@
  */
 package org.springframework.security.config.annotation.method.configuration
 
+
+import org.springframework.security.access.hierarchicalroles.RoleHierarchy;
+import org.springframework.security.access.hierarchicalroles.RoleHierarchyImpl;
+
 import java.lang.reflect.Proxy;
 
 import org.junit.After;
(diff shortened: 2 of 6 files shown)


rwinch:
@jgrandja Can you review this please?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
```diff
@@ -188,6 +190,14 @@ final ExpressionBasedFilterInvocationSecurityMetadataSource createMetadataSource
 			if (trustResolver != null) {
 				defaultHandler.setTrustResolver(trustResolver);
 			}
+			ApplicationContext context = http.getSharedObject(ApplicationContext.class);
+			if(context != null) {
```
jgrandja: I'm wondering if we need to do this null check? The context is auto-wired into WebSecurityConfigurerAdapter and shared when getHttp() returns. If it's not set as a shared object then we would want a NPE to be thrown so we can detect the programming error. What do you think about removing the null check?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
```diff
@@ -188,6 +190,14 @@ final ExpressionBasedFilterInvocationSecurityMetadataSource createMetadataSource
...
+				String[] roleHiearchyBeanNames = context.getBeanNamesForType(RoleHierarchy.class);
+				if(roleHiearchyBeanNames.length == 1) {
```
jgrandja: What if more than 1 RoleHierarchy bean is in the context? Should we default to the first one? Or should we throw a configuration exception? What makes more sense?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
rwinch: There is no guarantee the context has been set (it is optional). This feature is only available if we have access to the ApplicationContext.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
rwinch: It makes sense to do it only for the scenario where there is one bean. Multiple beans means the order is not deterministic. This would cause very strange errors if in dev they get once bean and then deploy to production and get a different Bean.

If users need multiple beans they can always explicitly wire up the instance they wish to use.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
jgrandja: Maybe I'm missing something? I see that the context is required for `WebSecurityConfigurerAdapter` as the default for `Autowired.required` is true:

```
	@Autowired
	public void setApplicationContext(ApplicationContext context) {
		this.context = context;
	}

```

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
jgrandja: Yes makes sense if the order is not deterministic.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
rwinch: `HttpSecurity` can be initialized by users outside of `WebSecurityConfigurerAdapter `

config/src/test/groovy/org/springframework/security/config/annotation/method/configuration/GlobalMethodSecurityConfigurationTests.groovy
```diff
@@ -470,4 +474,26 @@ public class GlobalMethodSecurityConfigurationTests extends BaseSpringSpec {
...
+			service.preAuthorizeAdmin()
+		then:
+			noExceptionThrown()
+	}
+
+	@EnableGlobalMethodSecurity(prePostEnabled = true)
+	@Configuration
+	public static class RoleHierarchyConfig extends BaseMethodConfig {
+		@Bean
+		RoleHierarchy roleHiearch() {
```
jgrandja: Typo? I think you meant roleHierarchy

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
jgrandja: Yes it can. Sounds good.

config/src/test/groovy/org/springframework/security/config/annotation/method/configuration/GlobalMethodSecurityConfigurationTests.groovy
rwinch: Good catch. I pushed a fix

jgrandja:
Thanks for the PR @rwinch! This is now merged.

PR #3822, opened 2016-04-15
Only use methods present in Bouncy Castle 1.47.

william-tran:
- [] I have signed the CLA

This forces us to avoid using CipherOutputStream, and instead use the
BlockCiphers directly. As an extra measure for correctness, test the
equivalence of the BC implementations against data sizes from 1 to 2048
bytes.

Fixes gh-2917


rwinch:

Issue #3823, opened 2016-04-18
NullPointerException OnCommittedResponseWrapper trackContentLength


rwinch:
`OnCommittedResponseWrapper` can cause a `NullPointerException` in `trackContentLength` 

Replaces #218

PR #3824, opened 2016-04-18
Fix NPE in OnCommittedResponseWrapper trackContentLength

rwinch:
OnCommittedResponseWrapper trackContentLength will throw a
NullPointerException when the content length passed in is null.

This commit properly tracks the null value as a length of 4.

Fixes gh-3823


jgrandja:
Thanks for this fix @rwinch! It's now merged.

Issue #3371, opened 2015-11-25
SEC-3161: InvalidSessionStrategy Not Configurable From Security Config


spring-issuemaster:
[Alper Kanat](https://jira.spring.io/secure/ViewProfile.jspa?name=alperkanat) (Migrated from [SEC-3161](https://jira.spring.io/browse/SEC-3161?redirect=false)) said:

Hi,

I have a security configuration defined like below. It seems like InvalidSessionStrategy isn't configurable by using the configure(final HttpSecurity http) method in WebSecurityConfigurerAdapter.

I tried manually defining a bean for SessionManagementFilter and setting an InvalidSessionStrategy but this caused problems during initialization of Spring Boot since SecurityContextRepository doesn't seem to exist at that point.

I tried moving filter creation into configure method but this time I get org.springframework.security.config.annotation.AlreadyBuiltException: This object has already been built exception.

Adding a setter method to SessionManagementConfigurer would solve this problem. What I need is to return a custom response instead of redirection.

``` 
@Configuration
@EnableWebSecurity
public class SecurityConfig extends WebSecurityConfigurerAdapter {
    @Autowired
    private Environment env;

    @Override
    protected void configure(final HttpSecurity http) throws Exception {
        final SessionManagementFilter sessionManagementFilter = new SessionManagementFilter(
            http.getSharedObject(SecurityContextRepository.class)
        );

        sessionManagementFilter.setInvalidSessionStrategy(invalidSessionStrategy);

        http
            .csrf().disable()
            .addFilterBefore(sessionManagementFilter, UsernamePasswordAuthenticationFilter.class)
            .exceptionHandling()
                .authenticationEntryPoint(authenticationEntryPoint)
                .accessDeniedHandler(accessDeniedHandler)
                .and()
            ...
    }
}
```

PR #3827, opened 2016-04-19
Add Java Configuration InvalidSessionStrategy

rwinch:
Allow configuring the InvalidSessionStrategy in Java Configuration.

Fixes gh-3371


jgrandja:
Thanks @rwinch!

PR #3836, opened 2016-04-21
fix(testing): change the way to detect the springSecurityFilterChain …

jgoldhammer:
- [X] I have signed the CLA

…in WebTestUtils

When enabling debug for spring security, the FilterChainProxy will be wrapped by the DebugFilter.
This debugfilter will be registered as bean springSecurityFilterChain. The WebTestUtils will now search for the bean by name instead of the hardcoded class.
In this case we have to cast to a Java ServletFilter to support both filter...

No testcases so far...

fixes the bug discussed on gitter
https://gitter.im/spring-projects/spring-security?at=5718e2d3b129b59c56da3a80

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: java
 
 jdk:
@@ -16,5 +17,6 @@ cache:
   directories:
     - $HOME/.gradle/caches/
     - $HOME/.gradle/wrapper/
+    - $HOME/.m2
 
 script: ./gradlew build
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: false
 language: java
 
 jdk:
@@ -17,6 +16,5 @@ cache:
   directories:
     - $HOME/.gradle/caches/
     - $HOME/.gradle/wrapper/
-    - $HOME/.m2
 
 script: ./gradlew build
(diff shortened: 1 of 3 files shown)


rwinch:
Thanks for the PR! We will definitely want a test case for this before we merge. You can create a new test similar to https://github.com/spring-projects/spring-security/blob/master/test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestPostProcessorsCsrfTests.java Since the test only supports a single context, you would need to create a new class `SecurityMockMvcRequestPostProcessorsCsrfDebugFilterTests`

jgoldhammer:
@rwinch , please check my test. I don´t know how to test it really because the new TestCsrfTokenRepository steps in and I have no chance (besides reflection) to test it really...
Or I am wrong here? Any advice?

Thanks,
Jens

jgoldhammer:
I have changed the test and hope that this fulfills your expectations...

rwinch:
@jgoldhammer Thanks for the update! Unfortunately the tests appear to be failing (likely spaces instead of tabs being used).

NOTE: If you import into Eclipse with the checkstyle plugin installed you will get the errors immediately within the IDE.

jgrandja:
@jgoldhammer I applied the fix for the checkstyle to pass (replaced spaces with tabs) and a little polish for the test. This is now merged. Thanks!

jgoldhammer:
Thanks for that.  Little time at the moment,  so I really appreciate that...

PR #3839, opened 2016-04-22
Clarifies sessionAuthenticationStrategy setter

jgrandja:
Issue gh-234


rwinch:
Thanks for the PR! this is now merged into master

Issue #3835, opened 2016-04-21
HttpOnly Flag for CSRF Cookie created by CookieCsrfTokenRepository


jgoldhammer:
### Summary

The new CookieCsrfTokenRepository creates a new cookie in org.springframework.security.web.csrf.CookieCsrfTokenRepository.saveToken(CsrfToken, HttpServletRequest, HttpServletResponse).
This cookie should be httpOnly by default- the setting should be configurable.

### Actual Behavior

At the moment the cookie has no httpOnly flag.

### Expected Behavior

The cookie should have the httpOnly flag by default and should be configurable when creating the CookieCsrfTokenRepository

### Configuration

```
CookieCsrfTokenRepository csrfTokenRepository = new CookieCsrfTokenRepository();
csrfTokenRepository.setCookieName(CsfSecurityConfiguration.csrfTokenCookieName);
// proposal to configure the httpOnly flag...
csrfTokenRepository.setCookieHttpOnly(true);

http.csrf().csrfTokenRepository(csrfTokenRepository)
```
 
### Version

4.1RC2

### Sample

-

rwinch:
@jgoldhammer Thanks for the report!

jgoldhammer:
Hi Rob (@rwinch) and Joe (@jgrandja),
I was able to dive deeper into the topic and my created issue and proposed solution causes a bug in our application. Why?
HttpOnly cookies are important in general for web security for protecting sensitive values in browsers which should not be stolen by javascript injection like xsss, but httponly cookies cannot be read from javascript, so our application was not able to read the csrf token from the cookie.

How does our scenario looks like?
- Server generates a csrf token and place it in the http response as http cookie
- browser reads the cookie and place the csrf token as http header in each request (this is done by angularjs)
- server validates the csrf token from the http header 

So the csrf cookie cannot be protected in this case which makes are application less secure...
I think, the general mechanism for CSRF and single page applications (only using Spring MVC for building up the rest service) should be adressed in future releases for Spring Security. What do you think? 

Thanks,
Jens

rwinch:
@jgoldhammer Thanks for the feedback.

I had given this some thought already. You are right that JavaScript cannot access the cookie. This might be desirable in the case where an application populates the token manually and simply wants to store the CSRF in a cookie rather than session.

If an application needs JavaScript to read the cookie, then they can set httpOnly to false explicitly. This would ensure that JavaScript can still read the CSRF token.

[...]

PR #3841, opened 2016-04-22
CookieCsrfTokenRepository supports HttpOnly

jgrandja:
Issue gh-3835

### DIFF
--- a/web/src/main/java/org/springframework/security/web/csrf/CookieCsrfTokenRepository.java
+++ b/web/src/main/java/org/springframework/security/web/csrf/CookieCsrfTokenRepository.java
@@ -49,12 +49,15 @@ public final class CookieCsrfTokenRepository implements CsrfTokenRepository {
 
 	private String cookieName = DEFAULT_CSRF_COOKIE_NAME;
 
-	private Method setHttpOnlyMethod;
+	private final Method setHttpOnlyMethod;
 
-	private boolean cookieHttpOnly = true;
+	private boolean cookieHttpOnly;
 
 	public CookieCsrfTokenRepository() {
 		this.setHttpOnlyMethod = ReflectionUtils.findMethod(Cookie.class, "setHttpOnly", boolean.class);
+		if (this.setHttpOnlyMethod != null) {
+			this.cookieHttpOnly = true;
+		}
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


web/src/main/java/org/springframework/security/web/csrf/CookieCsrfTokenRepository.java
```diff
@@ -47,6 +49,14 @@
 
 	private String cookieName = DEFAULT_CSRF_COOKIE_NAME;
 
+	private Method setHttpOnlyMethod;
+
+	private boolean cookieHttpOnly = true;
+
+	public CookieCsrfTokenRepository() {
+		this.setHttpOnlyMethod = ReflectionUtils.findMethod(Cookie.class, "setHttpOnly", boolean.class);
```
jgoldhammer: Just for understanding: Any reason why using Reflection here? Support of older servlet specs?

jgrandja: Yes, you are correct. HttpOnly attribute was introduced in Servlet 3.0. Using this reflective pattern allows us to support older versions of the servlet spec.

web/src/test/java/org/springframework/security/web/csrf/CookieCsrfTokenRepositoryTests.java
```diff
@@ -115,6 +115,30 @@ public void saveTokenNull() {
...
+		this.repository.setCookieHttpOnly(false);
+		CsrfToken token = this.repository.generateToken(this.request);
+		this.repository.saveToken(token, this.request, this.response);
+
+		Cookie tokenCookie = this.response
+				.getCookie(CookieCsrfTokenRepository.DEFAULT_CSRF_COOKIE_NAME);
+
+		assertThat(tokenCookie.isHttpOnly()).isFalse();
+	}
+
```
rwinch: We need a test for Servlet 3.0. The test can be written by using PowerMock and mocking the result of `ReflectionUtils.findMethod`. You can find an example in [AbstractRememberMeServicesTests](https://github.com/spring-projects/spring-security/blob/4.1.0.RC2/web/src/test/java/org/springframework/security/web/authentication/rememberme/AbstractRememberMeServicesTests.java)

web/src/main/java/org/springframework/security/web/csrf/CookieCsrfTokenRepository.java
```diff
@@ -115,6 +129,15 @@ public void setCookieName(String cookieName) {
 		this.cookieName = cookieName;
 	}
 
+	/**
+	 * Sets the HttpOnly attribute on the cookie containing the CSRF token. Defaults to true.
+	 *
+	 * @param cookieHttpOnly true sets the HttpOnly attribute, false does not set it.
+	 */
+	public void setCookieHttpOnly(boolean cookieHttpOnly) {
```
rwinch: We should check to see a user explicitly sets `true` and `setHttpOnlyMethod ` is null. If so, we should throw an `IllegalArgumentException` since we will not be setting the cookie as HttpOnly

web/src/main/java/org/springframework/security/web/csrf/CookieCsrfTokenRepository.java
```diff
@@ -47,6 +49,14 @@
 
 	private String cookieName = DEFAULT_CSRF_COOKIE_NAME;
 
+	private Method setHttpOnlyMethod;
```
rwinch: This should likely be final

jgoldhammer:
At least, this should be in the docs that httponly is not supported before servlet 3.0. Or what do you think?

jgrandja:
@jgoldhammer yes agreed, I will document the javadoc mentioning this is supported for Servlet >= 3.0

rwinch:
Thanks @jgrandja! This is merged into master

PR #3846, opened 2016-04-24
Provide editor config for spring-security

### LINKED ISSUE #3845: Provide editorconfig for spring-security
Currently, there is no editorconfig for spring-security.

see http://editorconfig.org/ for more details.

jgoldhammer:
This commit provides following editor config:
- Tabs, not spaces
- Unix (LF), not dos (CRLF) line endings
- Eliminate all trailing whitespace
- Latin-1 (ISO-8859-1) encoding for Java sources; use native2ascii to convert if necessary

Following convention cannot be set by editorconfig (only vim and emacs support maxline length of editorconfig):
- Aim to wrap code at 120 characters, but favor readability over wrapping

fixes #3845

- [X] I have signed the CLA

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+# EditorConfig for Spring Security
+# see https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.md#mind-the-whitespace
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = latin1
+
+# 4 tabs indentation
+indent_style = tab
+tab_width = 4
+
+trim_trailing_whitespace = true


jgoldhammer:
Any comments here?  Thanks Jens

jgrandja:
Hi @jgoldhammer. Thanks for your input here!

@rwinch and I have been discussing on a common strategy for ensuring a consistent coding style across development platforms (IntelliJ and Eclipse). We definitely plan on coming up with a strategy soon.
I haven't used EditorConfig before but at first glance the supported properties are fairly light. However, we will look into this further as we evaluate the available options.

We'll keep you in the loop!

rwinch:
I think given this is a step in the right direction, could be changed without being considered a non-passive change, and won't hurt users who don't care about it we should merge. If you agree @jgrandja please merge.

jgrandja:
@rwinch Agreed that this is a step in the right direction. If we had to change the code-style strategy at a later point in time, it would be fairly easy to switch. I'll go ahead with this merge.

jgrandja:
Thanks for adding this @jgoldhammer! This is now merged to master.

Issue #3852, opened 2016-04-28
Cannot build root project on Windows


volyx:
### Summary

Command `gradle build` failed with ` String index out of range: -1` in `settings.gradle' line: 64`
### Actual Behavior

```
:buildSrc:compileJava UP-TO-DATE
:buildSrc:compileGroovy
:buildSrc:processResources UP-TO-DATE
:buildSrc:classes
:buildSrc:jar UP-TO-DATE
:buildSrc:assemble UP-TO-DATE
:buildSrc:compileTestJava UP-TO-DATE
:buildSrc:compileTestGroovy UP-TO-DATE
:buildSrc:processTestResources UP-TO-DATE
:buildSrc:testClasses UP-TO-DATE
:buildSrc:test UP-TO-DATE
:buildSrc:check UP-TO-DATE
:buildSrc:build UP-TO-DATE

FAILURE: Build failed with an exception.

* Where:
Settings file 'D:\private\github\spring-security\settings.gradle' line: 64

* What went wrong:
A problem occurred evaluating settings 'spring-security'.
> String index out of range: -1

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED
```

### Expected Behavior

Project build succesfully
### Configuration

Windows 7, Gradle 2.10

### Version

Current master branch

jgrandja:
Thanks for catching this and the PR @volyx! This is now merged.

PR #3853, opened 2016-04-28
Fix File separator in settings.gradle script

volyx:
Parent spring-security project failing when run `gradle build` on Windows. The reason is that procedure `includeSamples(String samplesDir)` in settings.gradle file(line 59) use hardcoded file separator.

Constant `File.separator` fixes the problem.

#3852 

- [X] I have signed the CLA

Fixes gh-3852


jgrandja:
This has been merged. Thanks for the PR!

PR #3854, opened 2016-04-28
argument order for passwordencoder

githubusernameforme:
passwordencoder declares the arguments as encodedpassword and rawpassword, but here it seems to be getting called in reverse order

- [X] I have signed the CLA


rwinch:
Thanks for the PR!

The code is fine as is. Perhaps the confusion is that the old [PasswordEncoder](https://github.com/spring-projects/spring-security/blob/4.1.0.RC2/core/src/main/java/org/springframework/security/authentication/encoding/PasswordEncoder.java) API is used?

The fact that the tests on this PR fail reinforces that things are working as expected. See https://travis-ci.org/spring-projects/spring-security/builds/126336526

Issue #3859, opened 2016-05-03
Add @AuthenticationPrincipal expression


rwinch:
Sometimes it may be necessary to transform the principal in some way. For example, if `CustomUser` needed to be final it could not be extended. In this situation the `UserDetailsService` might returns an `Object` that implements `UserDetails` and provides a method named `getCustomUser` to access `CustomUser`. For example, it might look like:

```java
public class CustomUserUserDetails extends User {
		// ...
		public CustomUser getCustomUser() {
				return customUser;
		}
}
```

We could then access the `CustomUser` using a [https://docs.spring.io/spring/docs/current/spring-framework-reference/html/expressions.html](SpEL expression) that uses `Authentication.getPrincipal()` as the root object:

```java
import org.springframework.security.core.annotation.AuthenticationPrincipal;

// ...

@RequestMapping("/messages/inbox")
public ModelAndView findMessagesForUser(@AuthenticationPrincipal(expression = "customUser") CustomUser customUser) {

	// .. find messags for this user and return them ...
}
```

PR #3860, opened 2016-05-03
Add @AuthenticationPrincipal expression

rwinch:
It is now possible to provide a SpEL expression for
@AuthenticationPrincipal. This allows invoking custom logic including
methods on the principal object.

Fixes gh-3859

### DIFF
--- a/core/src/main/java/org/springframework/security/core/annotation/AuthenticationPrincipal.java
+++ b/core/src/main/java/org/springframework/security/core/annotation/AuthenticationPrincipal.java
@@ -30,9 +30,9 @@ import org.springframework.security.core.Authentication;
  * @author Rob Winch
  * @since 4.0
  *
- *  See: <a href="{@docRoot}/org/springframework/security/web/method/annotation/AuthenticationPrincipalArgumentResolver.html">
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/springframework/security/core/annotation/AuthenticationPrincipal.java
```diff
@@ -46,4 +46,33 @@
...
+	 * If specified will use the provided SpEL expression to resolve the principal. This
+	 * is convenient if users need to transform the result.
+	 *
+	 * <p>
+	 * For example, perhaps the user wants to resolve a CustomUser object that is final
+	 * and is leveraging a UserDetailsService. This can be handled by returning an object
+	 * that looks like:
+	 * </p>
+	 *
+	 * <code>
```
jgrandja: Javadoc is not rendering properly...
screen shot 2016-05-03 at 5 37 14 pm

rwinch: Fixed

jgrandja:
Thanks for this new feature @rwinch! This is now merged to master.

Issue #3865, opened 2016-05-05
Update CookieCsrfTokenRepository documentation to set cookiHttpOnly=false


rwinch:
Currently [CookieCsrfTokenRepository](http://docs.spring.io/spring-security/site/docs/4.1.0.RELEASE/reference/htmlsingle/#csrf-cookie) does not specify that the httpOnly flag needs set to false. We should update the reference to include this setting (and a comment about it) since it states that the settings will work with AngularJS.

PR #3868, opened 2016-05-05
Update CookieCsrfTokenRepository docs to cookieHttpOnly=false

rwinch:
Currently CookieCsrfTokenRepository does not specify that the httpOnly
flag needs set to false. We should update the reference to include this
setting (and a comment about it) since it states that the settings will
work with AngularJS.

This commit updates the documentation and provides a convenience factory
method to create a CookieCsrfTokenRepository with cookiHttpOnly=false

Fixes gh-3865

### DIFF
--- a/web/src/main/java/org/springframework/security/web/csrf/CookieCsrfTokenRepository.java
+++ b/web/src/main/java/org/springframework/security/web/csrf/CookieCsrfTokenRepository.java
@@ -31,7 +31,7 @@ import org.springframework.web.util.WebUtils;
 /**
  * A {@link CsrfTokenRepository} that persist the CSRF token in a cookie named
  * "XSRF-TOKEN" and reads from the header "X-XSRF-TOKEN" following the conventions of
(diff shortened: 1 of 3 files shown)


web/src/main/java/org/springframework/security/web/csrf/CookieCsrfTokenRepository.java
```diff
@@ -153,6 +153,19 @@ private String getCookiePath(HttpServletRequest request) {
 		return contextPath.length() > 0 ? contextPath : "/";
 	}
 
+	/**
+	 * Factory method to conveniently creaate an instance that has
```
jgrandja: typo on cre**aa**te

rwinch: Thanks! This is now fixed

jgrandja:
Thanks @rwinch! This is now merged to master.

Issue #3866, opened 2016-05-05
Add Spring Boot Helloworld guide


jgrandja:
Add Spring Boot Helloworld guide

PR #3877, opened 2016-05-09
Add Spring Boot Helloworld guide

jgrandja:
* Add Spring Boot Helloworld guide
* Add Spring Boot Insecure sample
* Add reference to Samples/Guides to Manual

Fixes gh-3866


rwinch:
Thanks! This is now merged into master.

Issue #2900, opened 2014-07-09
SEC-2680: Fire an event when logout has finished


spring-issuemaster:
[Marten Deinum](https://jira.spring.io/secure/ViewProfile.jspa?name=mdeinum) (Migrated from [SEC-2680](https://jira.spring.io/browse/SEC-2680?redirect=false)) said:

During login several events can be fired like for instance `AuthenticationSuccessEvent` and `AuthenticationFailureExpiredEvent`. It would be nice that on a logout (successful, session limit exceeded etc.) also an event would be fired.

This would enable to create an `ApplicationListener` which logs the start/end of a user session.

ptahchiev:
+1 on this one.

kazuki43zoo:
This issue is same with #3307.

okohub:
@rwinch hello, is this issue still active? are there any plans to integrate this feature?

okohub:
@rwinch Hello Rob, there is an active another issue about this #3307. Should I open a new PR for this enhancement? What do you think?

rwinch:
Thanks for pointing out the duplicate. If you are able, we'd love for a PR to be submitted.

okohub:
@rwinch hello rob, the PR is open. Do you have a time to check? May be this can be included to 5.2.0 line. thanks! :) 

https://github.com/spring-projects/spring-security/pull/7306

PR #3882, opened 2016-05-13
Fire a LogoutSuccessEvent in case of logout

mdeinum:
This commit introduces a LogoutSuccessEvent which is
fired upon logout. Both the LogoutFilter and ConcurrentSessionFilter
fire the event to indicate a session ended forcefully or not.

Closes: gh-2900

- [X] I have signed the CLA


rwinch:
Thanks for the PR @mdeinum!

Perhaps this makes more sense implemented as a LogoutHandler? This would allow the logic to be reused. Thoughts?

mdeinum:
I was thinking about that as well. However I wanted to make sure that the event was fired even if someone would register his own `LogoutHandler`s. As long as we can make sure that this one is always part of the full collection of `LogoutHandler`s I don't see the problem with that.

rwinch:
I think that we should use LogoutHandler as the preferred approach. 

What might be good to do is also provide a `CompositeLogoutHandler` implementation which iterates over a bunch of `LogoutHandler` instances. Then anything using a collection of `LogoutHandler`'s could create an instance of `CompositeLogoutHandler`. I created a distinct issue for this #3895

kazuki43zoo:
Hi @rwinch, 
I think `LogoutHandler` has role for customizing a web layer processing (clear http session, clear any cookies, etc... ). On the other hand, `LogoutSuccessEvent` has role for customizing a domain layer processing(not depends on the web layer).
I prefer to add a new event to notify logout success to any spring bean. 
What do you think ?

Note: I've submit a same PR on #3883.

rwinch:
@kazuki43zoo I'm not sure I understand your comment or perhaps you didn't understand mine.

My suggestion was that LogoutFilter and ConcurrentSessionFilter should not fire the event directly. Instead, the LogoutSuccessEvent will be fired by a LogoutHandler. This allows the logic for firing of the event to be reused.

I think we agree that LogoutSuccessEvent should not depend on the web layer. The only difference is where the event should be fired from.

kazuki43zoo:
@rwinch Thanks for your comment!!

I understood your opinion!!
I agree with you.

mdeinum:
@rwinch What you describe was actually more or less my initial attempt :) (with a `CompositeLogoutHandler`. I can refactor the code. Would then also refactor the array/collection of `LogoutHandler`s in the `LogoutFilter` and `ConcurrentSessionFilter` (and maybe other places) to use this `CompositeLogoutHandler`?

I can have a look at that issue and also refactor this into using a `LogoutHandler` instead of directly firing events (I might even still have that in my local history).

Issue #3881, opened 2016-05-12
Regression: 4.1.0 broke with(csrf())


theHacker:
### Summary

Since version 4.1.0 tests using `with(csrf())` are failing because the `CsrfFilter` does not get the token anymore and therefore an unexpected `AccessDeniedException` is thrown.

### Actual Behavior

Tests using `with(csrf())` are getting a status code 403.

I dug a little deeper. The problems seems to be the introduction of `TestCsrfTokenRepository`.
https://github.com/spring-projects/spring-security/blame/d3a9cc6eae56fb9826d86f021e71faa482675abe/test/src/main/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestPostProcessors.java#L323

`SecurityMockMvcRequestPostProcessors.postProcessRequest()` wrapps the `repository` with a `TestCsrfTokenRepository`.
When `saveToken()` is called the `TestCsrfTokenRepository` only attaches the token to the request as an attribute. But it does not delegate the call to the `HttpSessionCsrfTokenRepository` to save the token onto the session.
`CsrfFilter`'s `tokenRepository` is the `HttpSessionCsrfTokenRepository`. When the request is filtered, a lookup is done in the session, but there is no token, so `AccessDeniedException` is thrown and the test ends up with status code 403.

### Expected Behavior

`CsrfFilter` gets the token correctly, when test uses `with(csrf())`.

### Configuration

Nothing special. Spring Security enables CSRF protection by default.

### Version

- 4.0.2, 4.0.3, 4.0.4: not affected
- 4.1.0.RC1, 4.1.0.RELEASE: affected

### Sample

    @Test
    public void testMe() throws Exception {
        MyController myController = new MyController();

        MockMvc mockMvc = MockMvcBuilders
                .standaloneSetup(myController)
                .addFilters(springSecurityFilterChain)
                .build();

        mockMvc.perform(post("/foo.html")     // Spring Security default requires a CSRF token for POST
                .with(csrf()))                // Yes, we have a valid token
                .andExpect(status().is(302)); // expect redirection since we are not logged in
    }

I created a full sample project: https://github.com/theHacker/spring-security-4.1.0-csrf-bug
In `pom.xml` switch the version of Spring Security to enable/disable the bug.

rwinch:
Thank you for such a detailed write up!

rwinch:
[...]

PR #3884, opened 2016-05-13
Fix test .standaloneSetup

rwinch:
Previously, Spring Security's test support did not work well with the
standalone setup. This was because the springSecurityFilterChain was not
found by the WebTestUtils.

This commit ensures that the springSecurityFilterChain is added as a
servlet attribute if it is explicitly defined. WebTestUtils can then
find the springSecurityFilterChain in the ServletContext.

Fixes gh-3881


rwinch:
@jgrandja For your review

NOTE: There are two commits to keep formatting changes in a distinct commit. We should rebase and then merge into master vs using GitHub UI as we do not want to squash the commits.

jgrandja:
Thanks for the PR @rwinch. This has now been merged to master.

Issue #3873, opened 2016-05-06
Update documentation to include need for spring-security-test dependency


rwinch:
Update documentation to include need for spring-security-test dependency

Related #3861

PR #3885, opened 2016-05-13
Document spring-security-test dependency

rwinch:
Fixes gh-3873


jgrandja:
Thanks for the PR @rwinch. This is now merged to master.

PR #3897, opened 2016-05-21
Updated FAQ to clarify how the authorize tag works

geditdk:
I've added a section in the FAQ about using the `<sec:authorize url="">` JSP tag with method security on the controllers.

It clarifies that it doesn't and gives a description why.

- [X] I have signed the CLA


rwinch:
Thanks for the PR @geditdk! This is now merged into master via 503828c9949e3adac953595b8823e9e565d37fab

Issue #3895, opened 2016-05-20
Create CompositeLogoutHandler


rwinch:
We should create a `CompositeLogoutHandler` which allows for a `List<LogoutHandler>` to be injected into it. The implementation would simply iterate over all the `LogoutHandler` instances. Then we can leverage the `CompositeLogoutHandler` in places like `LogoutFilter, `HttpServlet3RequestFactory`, `SecurityContextHolderAwareRequestFilter`, and `ConcurrentSessionFilter` 

**UPDATE** 

It would look something like:

```java
public final class CompositeLogoutHandler implements LogoutHandler {
     private final List<LogoutHandler> logoutHandlers;

     public CompositeLogoutHandler(LogoutHandler... logoutHandlers) {
          this.logoutHandlers = Arrays.asList(logoutHandlers);
    }

    public void logout(HttpServletRequest request, HttpServletResponse response, Authentication authentication ) { 
        for(LogoutHandler delegate : delegates) {
            delegate.logout(request, response, authentication);
        }
   }
}
```

mdeinum:
Do we want the current API to be backwards compatible? I.e deprecate the respective constructors and setter methods and create a `CompositeLogoutHandler` in there?

rwinch:
Yes. But probably don't even need to deprecate. Just instantiate the
CompositeLogoutHandler in the constructor. Since it accepts var args it's
convenient to keep around.

PR #3898, opened 2016-05-21
Add CompositeLogoutHandler

eddumelendez:
- [X] I have signed the CLA

Fixes gh-3895


rwinch:
@eddumelendez Thanks for the PR! This isn't quite what I had in mind though.

I imagined that `CompositeLogoutHandler` would implement `LogoutHandler` and perform the logic that is within `LogoutFilter` (i.e. iterate over each implementation and perform logout). See the updated description which provides an example. Of course we would want some tests for it too.

I'm going to close this. Looking forward to getting an update :)

PR #3933, opened 2016-06-16
Feature for using different bind user in active directory LDAP authentication provider

MichaelF25:
This adds a feature in the authentication provider for active directory to use a different user for the bind operation as the user which credentials are to be authenticated.

In some company AD servers it's not allowed for ordinary users to connect via LDAP. Microsoft recommends to use a dedicated machine to machine (M2M) user account for these operations.


jgrandja:
Thanks for submitting this PR @MichaelF25 

We are currently re-thinking our approach to an Active Directory `AuthenticationProvider` implementation. Our direction is to deprecate `ActiveDirectoryLdapAuthenticationProvider` and leverage the existing `LdapAuthenticationProvider` and provide a custom strategy of `LdapAuthenticator` for Active Directory.

The work has already started in #4064 if you would like to track the progress and provide any input/feedback.

I'm going to close this PR as we will address your input/feedback in #4064.
Thank you.

Issue #3910, opened 2016-06-01
HTTP response splitting attack prevention


tiwatsuka:
### Summary

When a user set http header value containing CR/LF,  the http response is split.
https://www.owasp.org/index.php/HTTP_Response_Splitting

Of course, users should validate the value to avoid this kind of problems.
However, it's worth to provide measures from framework for careless users.

I think HttpFirewall is suitable place to remove CR/LF in header.

### Actual Behavior

HttpHeaders headers = new HttpHeaders();
headers.add("abc", "abc¥r¥nContent-Length:100¥r¥n¥r¥n <http>...");

abc: abc
Content-Length: 100

<http>
...

### Expected Behavior

HttpHeaders headers = new HttpHeaders();
headers.add("abc", "abc¥r¥nContent-Length:100¥r¥n¥r¥n <http>...");

abc: abc Content-Length: 100  <http> ...

rwinch:
Thanks for the report. Can you expand on this issue?

tiwatsuka:
I'm sorry. I accidentally submitted before writing description.

According to my investigation, some application servers remove CR/LF by default. But not all. 
I confirmed that Undertow does not do it.

rwinch:
@tiwatsuka And how are you able to perform response splitting? Is your application allowing user provided input as the value of the HTTP Response headers? Are you finding Spring Security is doing this? Please elaborate.

tiwatsuka:
I assume that some careless developers use user input as value of the headers without validation.
I think that any CR/LF in the value of the HTTP response headers have to be trimmed off. Isn't it nice if Spring Security do it by default?

At  `org.springframework.security.web.firewall.FirewalledResponse#sendRedirect`, the argument is checked whether it does not contain CR/LF.
Similarly, the value of header can be checked at `addHeader` and `setHeader`.

rwinch:
@tiwatsuka Thanks for the response. Thanks for explaining the context (I was trying to determine if this was a new feature or a bug).

This would be nice. Would you be interested in providing a pull request?

eddumelendez:
@rwinch method `addHeader` is part of `OnCommittedResponseWrapper` but `setHeader` is part of `MessageBuilder` which is in `spring-messaging` artifact. Should `MessageBuilder` manage the same behavior?

rwinch:
@eddumelendez As always thanks for responding! I commented on the PR. Let's continue the discussion there

glavoie:
[...]

PR #3938, opened 2016-06-19
Prevent HTTP response splitting

eddumelendez:
Evaluate if http header value contains CR/LF.

Reference: https://www.owasp.org/index.php/HTTP_Response_Splitting

Fixes gh-3910

### DIFF
--- a/web/src/main/java/org/springframework/security/web/firewall/FirewalledResponse.java
+++ b/web/src/main/java/org/springframework/security/web/firewall/FirewalledResponse.java
@@ -22,8 +22,10 @@ import java.util.regex.Pattern;
 
 /**
  * @author Luke Taylor
+ * @author Eddú Meléndez
  */
 class FirewalledResponse extends HttpServletResponseWrapper {
+
 	private static final Pattern CR_OR_LF = Pattern.compile("\\r|\\n");
 
 	public FirewalledResponse(HttpServletResponse response) {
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@eddumelendez Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=3938)!

web/src/main/java/org/springframework/security/web/util/OnCommittedResponseWrapper.java
```diff
@@ -29,9 +29,12 @@
  *
  * @since 4.0.2
  * @author Rob Winch
+ * @author Eddú Meléndez
  */
 public abstract class OnCommittedResponseWrapper extends HttpServletResponseWrapper {
 
+	private static final String CR_OR_LF = "\r\n";
```
rwinch: Is there a reason you made the change here? I think the best place for the change is in `FirewalledResponse`. You would override both setHeader and addHeader methods of `HttpServletResponseWrapper`

eddumelendez:
@rwinch PR updated

rwinch:
@eddumelendez Thanks for the fast response. I had something like this in mind:


> You would override both setHeader and addHeader methods of HttpServletResponseWrapper

```java

@Override
public void setHeader(String name, String value) {
	if (CR_OR_LF.matcher(value).find()) {
		throw new IllegalArgumentException(
				"Invalid characters (CR/LF) in redirect location");
	}
	super.setHeader(name, value);
}

@Override
public void addHeader(String name, String value) {
	if (CR_OR_LF.matcher(value).find()) {
		throw new IllegalArgumentException(
				"Invalid characters (CR/LF) in redirect location");
	}
	super.addHeader(name, value);
}
```

eddumelendez:
@rwinch  Thanks for the hint. PR updated.

web/src/main/java/org/springframework/security/web/firewall/FirewalledResponse.java
```diff
@@ -34,10 +37,27 @@ public FirewalledResponse(HttpServletResponse response) {
...
+	@Override
+	public void addHeader(String name, String value) {
+		validateCRLF(value);
+		super.addHeader(name, value);
+	}
+
+	private void validateCRLF(String value) {
+		if (CR_OR_LF.matcher(value).find()) {
 			throw new IllegalArgumentException(
 					"Invalid characters (CR/LF) in redirect location");
```
rwinch: We need a different error message since these are not redirect locations. You could probably just change this to header since redirect location is a header.

eddumelendez:
@rwinch PR updated. I am also printing the header in the exception message.

rwinch:
Thanks for the PR! This is merged via 26fa4a4bf01433a4a6c93eeb03b432792df6696b

PR #3945, opened 2016-06-24
LogoutConfigurer: only allow suitable http methods

Exordian:
If the CSRF filter is disabled, any HTTP method is (and should be) allowed, according to the documentation.

We just stepped into the problem that using spring-security with CSRF does not reset the session cookie. The problem is that the first request is an `OPTIONS` request, which already triggered the logout. Any HTTP method really means any, so `HEAD`, `OPTIONS` or `TRACE` are also considered as valid. This PR would restrict it to any suitable method IMO (although `GET` and `POST` may already be enough).

This would improve spring-security when using single page application, that are not using CSRF yet.
I also don't think that anyone really wants to logout with a `HEAD`, `OPTIONS` or `TRACE` request.


pivotal-issuemaster:
@Exordian Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=3945)!

pivotal-issuemaster:
@Exordian Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=3945)!

rwinch:
@Exordian Thanks for the PR! Any chance you could update the tests?

Exordian:
delivered :)

rwinch:
Thanks for the PR! This is fixed via b17870ee07c18d48f753fa266c5b5cb98872d887

PR #3946, opened 2016-06-25
Gradle: upgrade to version 2.14

Exordian:
https://docs.gradle.org/2.14/release-notes#faster-gradle-builds
https://docs.gradle.org/2.13/release-notes#faster-gradle-builds
https://docs.gradle.org/2.12/release-notes#faster-build-script-compilation

Gradle is currently focusing on speeding up the build. Almost every version speeds up the build a little.


Exordian:
I just saw there was a heap spacd limit of 1GB and an incresed permagen space configuration. 

Permagen space was removed/replaced with Java8. Since travis is only checking Java8, I think this option can be removed safetly.

I don't see any benefits in limiting the JVM heap space limit to 1G, but if desired, I'd add it

jgrandja:
Thanks for the PR @Exordian!

I added back the JVM options for max memory size and max PermGen size.

FYI, the integration tests need to run under JDK 7 for Spring IO compatibility.

Also, if the memory limit is left out then the default size will take effect which may cause an OutOfMemory issue, as we've noticed that there are memory leaks with the Tomcat gradle plugin.

PR #3955, opened 2016-07-02
Fix test assertions

eddumelendez:
Previous tests have not been matching the assert and they are always running ok. Now, it has been fixed.


rwinch:
Thanks for the PR! This is merged via 13b0ddb7e65416003d44121a8e08cb25d2126e4e

Issue #3701, opened 2016-02-15
Create single point of defining the default role prefix


andrei-ivanov:
If one wants to change the default role prefix this must be done in more than one place and this can lead to errors if one of them is forgotten.
Currently I set it in 4 places:
```xml
<bean id="defaultMethodSecurityExpressionHandler" class="org.springframework.security.access.expression.method.DefaultMethodSecurityExpressionHandler">
	<property name="defaultRolePrefix" value="ROLE:" />
</bean>

<bean id="defaultWebSecurityExpressionHandler" class="org.springframework.security.web.access.expression.DefaultWebSecurityExpressionHandler">
	<property name="defaultRolePrefix" value="ROLE:" />
</bean>

<bean id="securityContextHolderAwareRequestFilter" class="org.springframework.security.web.servletapi.SecurityContextHolderAwareRequestFilter">
	<property name="rolePrefix" value="ROLE:" />
</bean>

<bean id="daoUserDetailsService" class="custom.DaoUserDetailsServiceImpl">
	<property name="rolePrefix" value="ROLE:" />
</bean>
```
Some strategy should be created to set the default role in a single place and the other components should use it.

rwinch:
@andrei-ivanov Thanks for creating this! I don't suppose you are interested in creating a PR too?

andrei-ivanov:
I can try. Any suggestions of component? :)

rwinch:
This would be something easy to change as we think more about it. However, perhaps something like `org.springframework.security.config.core.GrantedAuthorityDefaults` makes sense?

andrei-ivanov:
Uf, I'm afraid I got caught with a new project that started and I don't think I'll be able to implement this myself.
I hope you or someone else can :)

eddumelendez:
@rwinch so the expected configuration would be something like this?

```xml
<bean id="defaultMethodSecurityExpressionHandler" class="org.springframework.security.access.expression.method.DefaultMethodSecurityExpressionHandler">
    <property name="defaultRolePrefix" ref="grantedAuthorityDefaults" />
</bean>

<bean id="defaultWebSecurityExpressionHandler" class="org.springframework.security.web.access.expression.DefaultWebSecurityExpressionHandler">
    <property name="defaultRolePrefix" ref="grantedAuthorityDefaults" />
</bean>

<bean id="securityContextHolderAwareRequestFilter" class="org.springframework.security.web.servletapi.SecurityContextHolderAwareRequestFilter">
[...]

PR #3956, opened 2016-07-02
Make single definition of `defaultRolePrefix` and `rolePrefix`

eddumelendez:
Previous to this commit, role prefix had to be set in every class
causing repetition. Now, bean `GrantedAuthorityDefaults` can be used to
define the role prefix in a single point.

Fixes gh-3701

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/authentication/configurers/ldap/LdapAuthenticationProviderConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/authentication/configurers/ldap/LdapAuthenticationProviderConfigurer.java
@@ -15,11 +15,15 @@
  */
 package org.springframework.security.config.annotation.authentication.configurers.ldap;
 
+import java.io.IOException;
+import java.net.ServerSocket;
+
 import org.springframework.ldap.core.support.BaseLdapPathContextSource;
 import org.springframework.security.authentication.AuthenticationManager;
 import org.springframework.security.authentication.AuthenticationProvider;
 import org.springframework.security.authentication.encoding.PasswordEncoder;
 import org.springframework.security.authentication.encoding.PlaintextPasswordEncoder;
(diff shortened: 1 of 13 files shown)


eddumelendez:
@rwinch is it ok to replace all classes involved in this commit? Early feedback is welcome. Thanks in advance.

rwinch:
@eddumelendez Thanks for the PR! I think I'd prefer that any of the domain object avoid adding `ApplicationContextAware`. Instead, the configuration specific classes would try to detect the default role prefix and then set that on the domain object. This decouples the way the defaults are provided from the domain object and keeps the logic in the configuration.

eddumelendez:
@rwinch Thanks for feedback.

> I think I'd prefer that any of the domain object avoid adding ApplicationContextAware

Just to clarify, you meant `RoleVoter`, right?

rwinch:
@eddumelendez Thanks for the fast response. 

I mean unless a class is in the `spring-security-config` module this PR should not add 'implements ApplicationContextAware` to it. Some examples of classes that should *NOT* implement `ApplicationContextAware` are: `Jsr250MethodSecurityMetadataSource`, `SecurityExpressionRoot `, `DefaultMethodSecurityExpressionHandler`, `RunAsManagerImpl`, `RoleVoter`, `JdbcDaoImpl`, etc.

For now we should not add `GrantedAuthorityDefaults` to any of these classes either (instead just invoke the setter for default role prefix on each object).

eddumelendez:
@rwinch PR has been updated

rwinch:
Thanks for the PR! This is merged via eabeaf35d6087bf48329bf91320148d2e5177649 I applied some Polish in b443baef04dbca1e48bd69b77944af86771a7dca for your review. The main goal was to continue with the suggestion from https://github.com/spring-projects/spring-security/pull/3956#issuecomment-243468545

PR #3965, opened 2016-07-08
Document schema changes in CONTRIBUTING.md

michael-simons:
Direct changes to XSD schemas will be overwritten by the build, it is necessary that the developer updates the RELAX NG schema instead.

See discussion on commit e297706e8bfda38117a99fd06dace74c5a8cc5f1.


rwinch:

Issue #2969, opened 2014-10-21
SEC-2744: BasicAuthenticationFilter decodes the Passwords with UTF-8


spring-issuemaster:
[Rob Winch](https://jira.spring.io/secure/ViewProfile.jspa?name=robwinch) (Migrated from [SEC-2744](https://jira.spring.io/browse/SEC-2744?redirect=false)) said:

All browser encode the Password in BasicAuthentification in ISO-8859-1.

See: Stackoverflow Discussion http://stackoverflow.com/questions/7242316/what-encoding-should-i-use-for-http-basic-authentication

The [BasisAuthentificationFilter](https://github.com/spring-projects/spring-security/blob/master/web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilter.java) explicitly decodes using UTF-8.

This broke Authentification with Umlauts in the password (example: passwordöäü) for me.

Further Analysis:
Chrome encodes the AuthentificationInfo with UTF-8
Firefox and IE use ISO8859-1

private String[] extractAndDecodeHeader(String header, HttpServletRequest request) throws IOException {

    byte[] base64Token = header.substring(6).getBytes("UTF-8");

Martin-Wegner:
The bug is still present.
A password with umlauts works only in Chrome but not in Firefox or Internet Explorer.

But the problem is not `byte[] base64Token = header.substring(6).getBytes("UTF-8");` because a Base64 token only contains encoding independent characters (you can use every encoding you want). 

The problem is another and results in a draft:
https://tools.ietf.org/id/draft-reschke-basicauth-enc-00.html

So the solution should be:

> On the other hand, the strategy below may already improve the user-visible behavior today:
> 
> In the first authentication request, choose the character encoding based on the user's credentials: if they do not need any characters outside the ISO-8859-1 character set, default to ISO-8859-1, otherwise use UTF-8.
> If the first attempt failed and the encoding used was ISO-8859-1, retry once with UTF-8 encoding instead.

I created a PR:
https://github.com/spring-projects/spring-security/pull/3966

Martin-Wegner:
@jgrandja can you close this issue please?

jgrandja:
@Martin-Wegner For sure.

jgrandja:
A solution to this Issue can be found [here](https://github.com/spring-projects/spring-security/pull/3966#issuecomment-242123838)

PR #3966, opened 2016-07-08
fix issue #2969: SEC-2744: BasicAuthenticationFilter decodes the Passwords with UTF-8

Martin-Wegner:
fix issue #2969: SEC-2744: BasicAuthenticationFilter decodes the Passwords with UTF-8

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilter.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilter.java
@@ -196,7 +196,7 @@ public class BasicAuthenticationFilter extends OncePerRequestFilter {
 			/*
 			 * https://tools.ietf.org/id/draft-reschke-basicauth-enc-00.html
 			 * > On the other hand, the strategy below may already improve the user-visible behavior today:
-			 * > 
+			 * >
 			 * > In the first authentication request, choose the character encoding based on the user's credentials: if they
 			 * > do not need any characters outside the ISO-8859-1 character set, default to ISO-8859-1, otherwise use UTF-8.
 			 * > If the first attempt failed and the encoding used was ISO-8859-1, retry once with UTF-8 encoding instead.
--- a/web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilter.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilter.java
@@ -192,6 +192,27 @@ public class BasicAuthenticationFilter extends OncePerRequestFilter {
 
 		}
 		catch (AuthenticationException failed) {
+
+			/*
+			 * https://tools.ietf.org/id/draft-reschke-basicauth-enc-00.html
+			 * > On the other hand, the strategy below may already improve the user-visible behavior today:
+			 * > 
+			 * > In the first authentication request, choose the character encoding based on the user's credentials: if they
+			 * > do not need any characters outside the ISO-8859-1 character set, default to ISO-8859-1, otherwise use UTF-8.
+			 * > If the first attempt failed and the encoding used was ISO-8859-1, retry once with UTF-8 encoding instead.
+			 */
+			if (failed instanceof BadCredentialsException && "UTF-8".equals(getCredentialsCharset(request))) {
+
+				try {
+					setCredentialsCharset("ISO-8859-1");
+					doFilterInternal(request, response, chain);
+					return;
+				} finally {
+					setCredentialsCharset("UTF-8");
+				}
+
+			}
+
 			SecurityContextHolder.clearContext();
 
 			if (debug) {
--- a/web/src/test/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilterTests.java
+++ b/web/src/test/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilterTests.java
@@ -20,6 +20,8 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.AdditionalMatchers.not;
 import static org.mockito.Mockito.*;
 
+import java.nio.charset.Charset;
+
 import javax.servlet.FilterChain;
 import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
@@ -59,9 +61,9 @@ public class BasicAuthenticationFilterTests {
 	public void setUp() throws Exception {
 		SecurityContextHolder.clearContext();
 		UsernamePasswordAuthenticationToken rodRequest = new UsernamePasswordAuthenticationToken(
-				"rod", "koala");
+				"rod", "koal\u00e4");
 		rodRequest.setDetails(new WebAuthenticationDetails(new MockHttpServletRequest()));
-		Authentication rod = new UsernamePasswordAuthenticationToken("rod", "koala",
+		Authentication rod = new UsernamePasswordAuthenticationToken("rod", "koal\u00e4",
 				AuthorityUtils.createAuthorityList("ROLE_1"));
 
 		manager = mock(AuthenticationManager.class);
@@ -139,7 +141,7 @@ public class BasicAuthenticationFilterTests {
 
 	@Test
 	public void testNormalOperation() throws Exception {
-		String token = "rod:koala";
+		String token = "rod:koal\u00e4";
 		MockHttpServletRequest request = new MockHttpServletRequest();
 		request.addHeader("Authorization",
 				"Basic " + new String(Base64.encodeBase64(token.getBytes())));
@@ -182,7 +184,7 @@ public class BasicAuthenticationFilterTests {
 	@Test
 	public void testSuccessLoginThenFailureLoginResultsInSessionLosingToken()
 			throws Exception {
-		String token = "rod:koala";
+		String token = "rod:koal\u00e4";
 		MockHttpServletRequest request = new MockHttpServletRequest();
 		request.addHeader("Authorization",
 				"Basic " + new String(Base64.encodeBase64(token.getBytes())));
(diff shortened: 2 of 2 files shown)


pivotal-issuemaster:
@Martin-Wegner Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=3966)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=3966) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@Martin-Wegner Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=3966)!

jgrandja:
Thanks for the PR @Martin-Wegner.

The link to the draft you referenced in the PR [Encoding Parameter for HTTP Basic Authentication](https://tools.ietf.org/id/draft-reschke-basicauth-enc-00.html) is outdated and actually expired.

I reviewed your PR and the basic authentication is retried once more using ISO-8859-1 encoding only if the first attempt UTF-8 was used to decode. ISO-8859-1 is a legacy charset so I'm feeling reluctant applying this. Can you please describe the specific scenario where this is causing an issue for you? Are you able to ensure the user-agent uses UTF-8 encoding for the authorization header?

Martin-Wegner:
My intention was: Keep the change as small as possible and don't break existing applications. UTF-8 is the default encoding in the current implementation. If the user does not change the encoding, then we can try ISO-8859-1 after UTF-8. Please remember: Firefox and Internet Explorer use ISO-8859-1 and only Chrome uses UTF-8. The current implementation does only work with Chrome but not with Firefox or Internet Explorer when the credentials contain umlauts. My PR repairs this until the user decides to choose another encoding as UTF-8.

jgrandja:
Thanks for the explanation on your use case. Very interesting and unique case re: umlaut characters.

I took a look again at the draft doc you referenced and in [Section 3](https://tools.ietf.org/id/draft-reschke-basicauth-enc-00.html#encoding) it states the following:

> Servers MAY use the "encoding" authentication parameter to express the character encoding they expect the user agent to use. [case-sens] [also-cred]
> 
> **The only allowed value is "UTF-8"**, to be matched case-insensitively (see [RFC2978], Section 2.3), indicating that the server expects the UTF-8 character encoding to be used ([RFC3629]).

Although this draft hasn't been accepted, it also references that the only allowed value is UTF-8.

Regardless of that, we would like to address your scenario. Quick question, do you have to use Basic Auth for your scenario? Can you use custom form login?

I looked at your proposed changes and I'm not sure if the recursive call to `doFilterInternal()` is the way to go. The typical approach is to configure Web Security using the `security namespace` or `HttpSecurity` via `WebSecurityConfigurerAdapter`.

For example, how about this?

```
@EnableWebSecurity
public class SecurityConfig extends WebSecurityConfigurerAdapter {

	@Override
		http
			.authorizeRequests()
				.anyRequest().authenticated()
			.and()
			.httpBasic().supportedCredentialsCharset("UTF-8", " ISO-8859-1")
			...
			...

```

Martin-Wegner:
I don't know if Firefox and Internet Explorer set the encoding. If they do, you could fill the existing `getCredentialsCharset(HttpServletRequest httpRequest)` method with live because the current implementation is without any logic:

```
protected String getCredentialsCharset(HttpServletRequest httpRequest) {
	return this.credentialsCharset;
}
```

The encoding could be extracted from the HTTP request.

Yes, I need the native basic authentication.

The reported issue is nearly 2 years old and no one was interested in it. Therefore, my approach was very pragmatic ;) Your solution looks very nice :)

jgrandja:
I'm not sure if we can depend on the user-agent sending the encoding.

I spoke to @rwinch and he made a very good point regarding my proposed solution. Exposing a configuration option:

`.httpBasic().supportedCredentialsCharset("UTF-8", " ISO-8859-1")`

may be very confusing for most users. The fact is this is a very specialized case and therefore a more advanced configuration. We do want to ensure the Java Config and XML namespace is as easy and intuitive as possible. Therefore, the proposed solution is actually not the best option.

@rwinch proposed leveraging a **Strategy** for this solution. One possible Strategy that we can re-use is the [Convert SPI](http://docs.spring.io/spring/docs/current/spring-framework-reference/htmlsingle/#core-convert).

For example, let's say we have the following implementation:

```
	public class BasicAuthenticationCredentialsConverter implements Converter<HttpServletRequest, Authentication> {

		@Override
		public Authentication convert(HttpServletRequest request) {
			.....
		}
	}

```

and we expose a setter on `BasicAuthenticationFilter`, for example

```
	public void setCredentialsConverter(Converter<HttpServletRequest, Authentication> credentialsConverter) {
           .....
	}

```

The default implementation of the `credentialsConverter` would cover what's happening today which is mostly in the `extractAndDecodeHeader()` method.

And for your use case, you can provide another implementation of `Converter<HttpServletRequest, Authentication>`  that uses ISO-8859-1 to decode the auth creds. Then you can plug in this implementation via `BasicAuthenticationFilter.setCredentialsConverter()` overriding the default.

Given that this is a special use case and more of an advanced configuration, you would be required to configure the `BasicAuthenticationFilter` directly and then add it to `HttpSecurity`, for example

`httpSecurity.addFilter(basicAuthenticationFilter)`

What are your thoughts on this?

Martin-Wegner:
Umlauts in credentials are a normal use case. Firefox and Internet Explorer use the ISO-8859-1 encoding and are widespread browsers. Chrome uses the UTF-8 encoding and is also a widespread browser. So the use case we are talking about is not a special case. Therefore, I do not like to force the user to implement such a complex converter to solve daily business.

Martin-Wegner:
Any plans or ideas for next steps?

jgrandja:
I've been tied up with other priorities the past couple weeks and SpringOne is next week.

Our preference is to leverage the Strategy pattern as part of this solution. It is a widely used pattern in the framework already and provides the flexibility for future implementations that can be plugged in (or configured) without changing the core filter processing code.

I understand your viewpoint on keeping things simple for the user and totally agree as this is always the goal. What do you think about this solution:

Instead of exposing a setter on `BasicAuthenticationFilter` that allows for configuring 1 converter, we allow a `List` of converters to be supplied, as per below:

```
public void setCredentialsConverters(List<Converter<HttpServletRequest, Authentication>> credentialsConverters) {
           .....
    }
```

The default implementation of  `BasicAuthenticationFilter` will have 2 converters supplied by default. The first one will use UTF-8 to decode the auth creds (what's happening today in the extractAndDecodeHeader method). The second one will use ISO-8859-1 to decode the auth creds.

So `BasicAuthenticationFilter` will iterate through it's list of `credentialsConverters` delegating to each converter until one of them is able to successfully convert the creds.

This default implementation will avoid the user from configuring the `BasicAuthenticationFilter`, however, if they need another strategy to decode using another charset, they have the flexibility to provide a new implementation and plug it in to the list of converters.

Thoughts? Suggestions?

Martin-Wegner:
Sounds good to me and I hoped that you can provide a default implementation. Would it be possible for the user to remove the both default converters?

jgrandja:
The user would be able to remove any of the default converters, however, there will be an assertion that at least 1 converter is required.

FYI, I'll circle back to this after SpringOne.

Martin-Wegner:
I hope you had a great SpringOne convention :) Any news on this issue?

jgrandja:
@Martin-Wegner Sorry not yet but plan on getting back to it shortly.

jgrandja:
Hi @Martin-Wegner I've circled back to this issue and have managed to come up with a solution for you (with the help of @rwinch) without adding any new code to the base. This solution is a custom configuration that sets up 2x `BasicAuthenticationFilter` in the chain. The first filter decodes using `ISO-8859-1` and the next one uses `UTF-8`. I've tested this configuration with IE, Firefox, Safari and Chrome using umlaut characters in the credentials and it works. Give this a try and let me know if you have any questions.

Here is the complete configuration:

```
package org.springframework.security.samples.config;

import org.springframework.beans.factory.annotation.Autowired;
import org.springframework.security.config.annotation.ObjectPostProcessor;
import org.springframework.security.config.annotation.authentication.builders.AuthenticationManagerBuilder;
import org.springframework.security.config.annotation.web.builders.HttpSecurity;
import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
import org.springframework.security.web.authentication.www.BasicAuthenticationEntryPoint;
import org.springframework.security.web.authentication.www.BasicAuthenticationFilter;

import java.nio.charset.Charset;

@EnableWebSecurity
public class SecurityConfig extends WebSecurityConfigurerAdapter {

	@Autowired
	private ObjectPostProcessor<Object> objectPostProcessor;

	// @formatter:off
	@Override
		http
				.authorizeRequests()
					.antMatchers("/public").permitAll()
					.antMatchers("/private").hasRole("USER");

		configureHttpBasic(http, true);
	}
	// @formatter:on

	// @formatter:off
	@Autowired
	public void configureGlobal(AuthenticationManagerBuilder auth) throws Exception {
		auth
			.inMemoryAuthentication()
				.withUser("user1").password("password").roles("USER")
				.and()
				.withUser("user2").password("passwordöäü").roles("USER");
	}
	// @formatter:on

	private void configureHttpBasic(HttpSecurity http, boolean configureDefault) throws Exception {
		if (configureDefault) {
			http.httpBasic();		// Configure the default http basic filter (decodes credentials using UTF-8)

			// Create a custom http basic filter that decodes credentials using ISO-8859-1
			BasicAuthenticationFilter customBasicAuthFilter = new BasicAuthenticationFilter(authenticationManager());
			customBasicAuthFilter.setCredentialsCharset(Charset.forName("ISO-8859-1").name());
			if (objectPostProcessor != null) {
				customBasicAuthFilter = objectPostProcessor.postProcess(customBasicAuthFilter);
			}

			// Add the custom filter before the default filter.
			http.addFilterBefore(customBasicAuthFilter, BasicAuthenticationFilter.class);
			// It's important that we add the custom filter before the default filter because
			// the custom filter will allow the request to 'proceed' to the next filter in the chain
			// which is the default http basic filter. This configuration is automatically set based on
			// the constructor used for the custom filter.
			//
			// For example, calling BasicAuthenticationFilter(AuthenticationManager authenticationManager)
			// will set ignoreFailure = true. This is what we want so the next filter in the chain
			// will attempt to authenticate using UTF-8 for decoding the credentials.

		} else {
			// Create a custom http basic filter that decodes credentials using ISO-8859-1.
			//
			// NOTE:
			// This will configure only 1 BasicAuthenticationFilter in the chain. For umlaut characters
			// or anything in the Latin1 code range between 128-255 this will work in Firefox, Safari and IE.
			// However, this configuration won't work in Chrome as the encoding using by Chrome is UTF-8.
			// If you want this to work for Chrome as well then you need to go with the configuration above
			// that have both BasicAuthenticationFilter's in the chain.

			BasicAuthenticationEntryPoint basicAuthEntryPoint = new BasicAuthenticationEntryPoint();
			basicAuthEntryPoint.setRealmName("Default Realm");
			BasicAuthenticationFilter customBasicAuthFilter = new BasicAuthenticationFilter(authenticationManager(), basicAuthEntryPoint);
			customBasicAuthFilter.setCredentialsCharset(Charset.forName("ISO-8859-1").name());
			if (objectPostProcessor != null) {
				customBasicAuthFilter = objectPostProcessor.postProcess(customBasicAuthFilter);
			}
			http.exceptionHandling().authenticationEntryPoint(basicAuthEntryPoint);
			http.addFilter(customBasicAuthFilter);
		}
	}
}

```

Martin-Wegner:
Oh very nice, it works :)

So it only needs 4 more lines of code in the existing configuration code:

```
// default BasicAuthenticationFilter uses UTF-8 for credentials encoding but we also want ISO-8859-1
// see https://github.com/spring-projects/spring-security/pull/3966#issuecomment-242123838
BasicAuthenticationFilter iso88591BasicAuthenticationFilter = new BasicAuthenticationFilter(
    authenticationManager());
iso88591BasicAuthenticationFilter.setCredentialsCharset(StandardCharsets.ISO_8859_1.name());
iso88591BasicAuthenticationFilter = objectPostProcessor.postProcess(iso88591BasicAuthenticationFilter);
http.addFilterBefore(iso88591BasicAuthenticationFilter, BasicAuthenticationFilter.class);
```

jgrandja:
@Martin-Wegner I'm happy we were able to get this resolved for you!

PR #3986, opened 2016-07-21
Fix <headers defaults-disabled=true> with no children

frickjack:
The online docs
     http://docs.spring.io/spring-security/site/docs/current/reference/html/headers.html
suggest that

```xml
     <headers defaults-disabled="true"/>
```

ought to disable the default security headers, but the HeaderWriterFilter has an assertion that its header-writer list is non empty.  This patch just changes Assert.notEmpty to Assert.notNull

### DIFF
--- a/web/src/main/java/org/springframework/security/web/header/HeaderWriterFilter.java
+++ b/web/src/main/java/org/springframework/security/web/header/HeaderWriterFilter.java
@@ -51,7 +51,7 @@ public class HeaderWriterFilter extends OncePerRequestFilter {
 	 * {@link HttpServletResponse}.
 	 */
 	public HeaderWriterFilter(List<HeaderWriter> headerWriters) {
-		Assert.notEmpty(headerWriters, "headerWriters cannot be null or empty");
+		Assert.notNull(headerWriters, "headerWriters cannot be null");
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@frickjack Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=3986)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=3986) to manually synchronize the status of this Pull Request.

frickjack:
Ugh - the XML was busted in the comment:
    headers defaults-disabled="true"
XML element should disable headers, but the filter is unhappy ...

pivotal-issuemaster:
@frickjack Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=3986)!

rwinch:
Given the lack of response for submitting an updates PR, I went ahead and fixed as I discussed in my previous comment.

Thanks again for submitting the PR and the issue. It is only through community members like you (that take the time to report issues) that we are able to make Spring better!

Issue #3997, opened 2016-08-01
Default Logout Handler with HTTP Basic and XMLHttpRequest should be 204


rwinch:
The default behavior for logout when the header `X-Requested-With: XMLHttpRequest` and HTTP Basic is enabled should be a 201. This will improve AngularJS experience which sends an accept header which includes `text/html`

PR #3999, opened 2016-08-01
Logout is 201 for XMLHttpRequest

rwinch:
Fixes gh-3997


rwinch:
Closing since should be based off of 4.1.x

Issue #4020, opened 2016-08-09
Exposing Beans for defaultMethodExpressionHandler can prevent Method Security


bitsofinfo:
### Updated Summary

If a `@Configuration` provides a `@Bean` that is used to default `GlobalMethodSecurityConfiguration's 
defaultMethodExpressionHandler`s `defaultMethodExpressionHandler` it will prevent any `@Bean` that is `@Autowired` into the same `@Configuration` from having method security enabled. For example:

```java
@Configuration
@EnableGlobalMethodSecurity(prePostEnabled = true)
public class SecurityConfig {
	// any one of the following @Bean will prevent DenyAllService from being
	// secured since DenyAllService is also Autowired into this same Configuration
	@Bean
	PermissionEvaluator permissionEvaluator() {
		return mock(PermissionEvaluator.class);
	}

	@Bean
	RoleHierarchy RoleHierarchy() {
		return mock(RoleHierarchy.class);
	}

	@Bean
	AuthenticationTrustResolver trustResolver() {
		return mock(AuthenticationTrustResolver.class);
	}

	@Autowired
	DenyAllService denyAll;
}

@Configuration
public class ServiceConfig {
	@Bean
	DenyAllService denyAllService() {
		return new DenyAllService();
	}
}

@PreAuthorize("denyAll")
public class DenyAllService {
	void denyAll() {
	}
}
```

### Summary

spring-data-rest `@PreAuthorize` annotated methods on a `@RepositoryRestResource ` annotated `PagingAndSortingRepository` interface fail to be evaluated on invocation if the resulting repository bean instance is `@Autowired` into a `@Configuration` annotated class.

### Actual Behavior

See this project for a runnable example: 
https://github.com/bitsofinfo/spring-boot-pre-authorize-issue-01

### Expected Behavior

@PreAuthorize expressions should be evaluated on requests that hit the repository

### Configuration

See this project for a runnable example: 
https://github.com/bitsofinfo/spring-boot-pre-authorize-issue-01

### Version

All latest spring-boot components

See this project for a runnable example: 
https://github.com/bitsofinfo/spring-boot-pre-authorize-issue-01

### Sample

https://github.com/bitsofinfo/spring-boot-pre-authorize-issue-01

rwinch:
@bitsofinfo Thanks for the report and the sample!  I am unable to reproduce (what I understand as) the problem.

[...]

PR #4021, opened 2016-08-11
Fix defaultMethodExpressionHandler autowiring

rwinch:
Previously if a Bean for GlobalMethodSecurityConfiguration's
defaultMethodExpressionHandler was found on a Configuration that also
@Autowired a Bean that enabled method security, the Bean that was
@Autowired would not have security enabled.

This fixes the issue by delaying the lookup of Beans populated on
GlobalMethodSecurityConfiguration's defaultMethodExpressionHandler.

Fixes gh-4020


jgrandja:
Thanks for the PR @rwinch!

PR #4026, opened 2016-08-16
Fix typo in autowiring warning

### LINKED ISSUE #3402: SEC-3203: GlobalMethodSecurityConfiguration typo in autowiring log message
[Robert Novotny](https://jira.spring.io/secure/ViewProfile.jspa?name=novotnyr) (Migrated from [SEC-3203](https://jira.spring.io/browse/SEC-3203?redirect=false)) said:

GlobalMethodSecurityConfiguration has two typos:

* method `setMethodSecurityExpressionHandler` logs a message "Not autwiring PermissionEvaluator since size != 1". However, there are no PermissionEvaluators involved. Instead, method handles `MethodSecurityExpressionHandler`, which may lead to confusing log message. Furthermore, there is a typo in the message
* method `setPermissionEvaluator` logs a message with a typo "Not autwiring PermissionEvaluator since size != 1".

novotnyr:
Fixes a misleading message that warns about PermissionEvaluator when MethodSecurityExpressionHandler should be mentioned instead.

Fixes #3402


rwinch:
Thanks for the report and the fix @novotnyr! This is now merged into master

Issue #4027, opened 2016-08-17
MvcMatchers: UnsupportedOperationException: Section 4.4 of the Servlet 3.0 specification does not permit this method to be called from a ServletContextListener that was not defined in web.xml, a web-fragment.xml file nor annotated with @WebListener


bjornharvold:
### Summary
Upgraded to v4.1.2 of Spring Security. Application no longer starts siting an issue with the mvcMatchers.

### Actual Behavior
This code has worked great since I started using it but fails in 4.1.2.

### Expected Behavior
Start up my Spring web app and have Spring Security secure my urls using mvcMatchers

### Configuration
```
private static final String API_ADMINISTRATION_URL = "/api/administration/**";
private static final String API_MANAGER_URL = "/api/manager/**";
private static final String API_USER_URL = "/api/secured/**";
private static final String API_AUTHENTICATED_URL = "/api/authenticated/**";

@Override
    protected void configure(HttpSecurity http) throws Exception {
        log.info("Configuring springSecurityFilterChain...");

        // header details
        http
                .cors()
                .and()
                .headers()
                .frameOptions()
                .sameOrigin()
                .cacheControl()
                .and().xssProtection()
                .and().contentTypeOptions()
                .and().httpStrictTransportSecurity()
        ;

        http
                .sessionManagement().sessionCreationPolicy(SessionCreationPolicy.STATELESS);

        // the url patterns to secure
        http
                .authorizeRequests()
                .mvcMatchers(API_AUTHENTICATED_URL).hasAnyRole("ADMIN", "MANAGER", "USER")
                .mvcMatchers(API_ADMINISTRATION_URL).hasRole("ADMIN")
                .mvcMatchers(API_MANAGER_URL).hasRole("MANAGER")
                .mvcMatchers(API_USER_URL).hasRole("USER")
                .mvcMatchers(API_LOGIN_URL).anonymous()
                .mvcMatchers(API_PUBLIC_URL).permitAll()
        ;

        // new logout handler from spring 4.0.2 - maybe nice to have. Has not been tested yet.
//        HttpStatusReturningLogoutSuccessHandler logoutHandler = new HttpStatusReturningLogoutSuccessHandler();
//        http
//                .logout().logoutSuccessHandler(logoutHandler);

        // filter details
        http
                .addFilterAfter(authFilter(), ApiRequestHeaderAuthenticationFilter.class)
                .addFilter(preAuthFilter());
        http
                .csrf().disable();

        http
[...]

PR #4031, opened 2016-08-19
Remove MvcRequestMatcher.afterPropertiesSet()

rwinch:
The validation does not work due to restrictions within the servlet
container. Specifically we cannot access the servlets that are registered.

This commit reverts the validation logic for MvcRequestMatcher to determine
if servletPath is required.

Fixes gh-4027


jgrandja:
Looks like we can also remove [this method](https://github.com/spring-projects/spring-security/pull/4031/commits/9dc3242db318a301ae0fd30ef58ae0978136c332#diff-a38bd1d0c99766ff91afbcf25f7dbb7dR126)

rwinch:
@jgrandja Thanks...that is somewhat unrelated to this change, so we should probably do that separately.

jgrandja:
Thanks for the PR @rwinch. This is now merged to 4.1.x

PR #4032, opened 2016-08-19
Remove MvcRequestMatcher.afterPropertiesSet()

rwinch:
The validation does not work due to restrictions within the servlet
container. Specifically we cannot access the servlets that are registered.

This commit reverts the validation logic for MvcRequestMatcher to determine
if servletPath is required.

Fixes gh-4027


jgrandja:
Thanks for the PR @rwinch. This is now merged to master.

Issue #4001, opened 2016-08-02
CSRF tokens are vulnerable to a BREACH attack


johnraycp:
### Summary

The CSRF tokens generated by Spring are vulnerable to a BREACH attack. More details at http://breachattack.com/

I'll create a pull request with a proposed code change to fix this.
### Actual Behavior

Spring always returns the same CSRF token to the browser.
### Expected Behavior

The simplest way to mitigate this would be to return a token which is composed of a random per secret request XORed with an internal CSRF token. This effectively means that the browser receives a new CSRF token with each request.
### Configuration

This only occurs when you turn on CSRF protection in Spring and also have HTTP compression enabled somewhere in your web server stack.
### Version

Currently exists in latest version in Git.
### Sample

N/A

jgrandja:
The proposed solution will be to introduce a default method to `CsrfToken`, as follows:

```
public interface CsrfToken {
    …
    default boolean matches(CsrfToken token) {
        return getToken().equals(token.getToken());
    }
}
```

Then override the default method in `DefaultCsrfToken` to handle the XOR logic.

Targeted for Spring Security 5.0 which will support Java 8.

Wadeck:
@jgrandja care about potential timing-attack with your proposal. Using `MessageDigest.isEqual` could be better at that point.

rh-id:
Hi @rwinch @jgrandja @eddumelendez do you mind if i take over this?

rwinch:
That would be great by me, but let's wait and get confirmation from @eddumelendez 

A few notes for whomever implements this https://github.com/spring-projects/spring-security/pull/7221#issuecomment-536724408

eddumelendez:
@rh-id sure, please go ahead! 🙂

sebtacke:
This issue popped up in a whitesource scan for me and I must say the proposed solution feels like building a high security gateway on an open field.
The CSRF protection requires session management, so the pages that have it will also have a session id cookie. If you enable HTTP compression for such a page an attacker can bruteforce the session id with a BREACH attack. Once the attacker has the session id, a randomized CSRF token won't do anything, because you can just call the page with the session id cookie and have a valid csrf token.
The only working protection against BREACH is not enabling HTTP compression for any page/URL that returns user supplied input. The safest solution is limiting compression to static resources or disabling it completely.

NicoleG25:
Hey @jzheaux sorry to randomly pick you but you were the last one to commit to the repo 😄 
[...]

PR #4042, opened 2016-08-26
Fix Issue 4001: CSRF tokens are vulnerable to a BREACH attack

johnraycp:
This is a change to the CsrfFilter class to add a random value that the CSRF token is XORed with. I've isolated the change to just the CsrfFilter. The code maybe isn't as clean as my prior pull request but it's simpler and doesn't require changing any interfaces.

NOTE: On a server using SSL and gzip compression this does not protect sensitive data which can be retrieved via a GET request to a page. However, it does stop an attacker from submitting forms or viewing sensitive data returned from a form submit.

Related #4001 and #4042

### DIFF
--- a/web/src/main/java/org/springframework/security/web/csrf/CsrfFilter.java
+++ b/web/src/main/java/org/springframework/security/web/csrf/CsrfFilter.java
@@ -16,6 +16,7 @@
 package org.springframework.security.web.csrf;
 
 import java.io.IOException;
+import java.security.SecureRandom;
 import java.util.Arrays;
 import java.util.HashSet;
 
@@ -28,6 +29,7 @@ import javax.servlet.http.HttpSession;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
+import org.springframework.security.crypto.codec.Base64;
 import org.springframework.security.web.access.AccessDeniedHandler;
 import org.springframework.security.web.access.AccessDeniedHandlerImpl;
 import org.springframework.security.web.util.UrlUtils;
@@ -63,6 +65,8 @@ public final class CsrfFilter extends OncePerRequestFilter {
 	 */
 	public static final RequestMatcher DEFAULT_CSRF_MATCHER = new DefaultRequiresCsrfMatcher();
 
+	private static final SecureRandom secureRandom = new SecureRandom();
+
 	private final Log logger = LogFactory.getLog(getClass());
 	private final CsrfTokenRepository tokenRepository;
 	private RequestMatcher requireCsrfProtectionMatcher = DEFAULT_CSRF_MATCHER;
@@ -93,8 +97,9 @@ public final class CsrfFilter extends OncePerRequestFilter {
 			csrfToken = this.tokenRepository.generateToken(request);
 			this.tokenRepository.saveToken(csrfToken, request, response);
 		}
-		request.setAttribute(CsrfToken.class.getName(), csrfToken);
-		request.setAttribute(csrfToken.getParameterName(), csrfToken);
+		XorEncodedToken xorEncodedToken = new XorEncodedToken(csrfToken);
+		request.setAttribute(CsrfToken.class.getName(), xorEncodedToken);
+		request.setAttribute(xorEncodedToken.getParameterName(), xorEncodedToken);
 
 		if (!this.requireCsrfProtectionMatcher.matches(request)) {
 			filterChain.doFilter(request, response);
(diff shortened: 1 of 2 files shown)


jgrandja:
Hi @johnraycp. Thanks for taking the time in submitting this PR.

Although the CSRF token may be exposed to BREACH when returned in a response body, for example, a hidden parameter in a form, it still does not generally solve BREACH overall. As @rwinch noted in the [comment](https://github.com/spring-projects/spring-security/pull/4002#issuecomment-239827460), there may be other sensitive data exposed in a form, such as bank information or SIN #.

The ideal solution IMO would allow the "masking" of all sensitive data on the outgoing response, not just the CSRF token. This feature would obviously require more of an effort but may very well provide a full-feature for solving or at least mitigating BREACH.

Disabling HTTP Compression altogether or [Referrer Check Mitigation or HTTP Chunked Encoding Mitigation](https://blog.qualys.com/ssllabs/2013/08/07/defending-against-the-breach-attack) are other possible solutions but this would require changes at the Web Server level and are less-intrusive.

I did look at your proposed changes and wanted to point out that there are a few failing tests as a result of the changes.

You mentioned in your [comment](https://github.com/spring-projects/spring-security/pull/4002#issuecomment-238385862) that other frameworks have implemented the "masking" of secrets. Can you please provide me more information on which frameworks have implemented this as I'd be curious to look at their implementation.

johnraycp:
If you're willing to accept this change let me know and I'll fix the tests.

Other examples I've seen using this technique are:
- Rails - https://github.com/rails/rails/pull/16570
- Facebook - https://www.facebook.com/notes/protect-the-graph/preventing-a-breach-attack/1455331811373632/
- PHP add-on - https://defuse.ca/mitigating-breach-tls-attack-in-php.htm

The PHP add-on also has an obfuscate function that randomly adds the Unicode Character 'ZERO WIDTH SPACE' (U+200B) before and after each character being encoded. So the single character 'a' ends up like

`
&#x200b;&#8203;&#x200b;&#8203;&#8203;a&#8203;&#8203;&#8203;&#8203;&#x200b;
`

The obfuscate idea could work for displaying text to a user on the page but won't work for the CSRF token. Perhaps bank information or SIN # should be fixed separately as a new ThymeLeaf obfuscate tag?

jgrandja:
Thanks for sending the info @johnraycp.

I read the following re: Facebook:

> Facebook's defense in depth keeps the CSRF token under wraps even in the most hostile circumstances. Many other proposed solutions, such as adding random padding to payloads, would merely delay an exploit rather than prevent it. Turning off compression would slow Facebook down dramatically, especially for mobile clients. Some websites could get away with watching the referrer and only enabling compression for trusted referrers. Facebook has a significant presence on third-party websites, so this would again hurt browsing speed on a huge number of websites where Facebook components are embedded.
> 
> **Unfortunately, there is currently no silver bullet that can prevent this attack everywhere. As a result, each website owner—especially those managing user accounts—should evaluate their defense against BREACH and CSRF attacks to make sure they are taking sufficient precautions.**
> 

As the bold highlight indicates, the method of defence goes a lot further than what any framework can help solve or mitigate as this needs to go deeper into the IT infrastructure setup, network layer security and even constant monitoring of these types of attacks to be really tight on security.

Nevertheless, as @rwinch indicated and I agree as well, fixing the CSRF support would be a step in the right direction.

Looking at the changes further, my preference would be to contain these changes in an implementation of  `CsrfTokenRepository` instead of the `CsrfFilter` itself. It would be a similar implementation as `HttpSessionCsrfTokenRepository` that stores the `CsrfToken` in the session and would also contain the XOR logic when the token is generated and compared with. This new implementation of `CsrfTokenRepository` can then be set on the `CsrfFilter` without any changes to the filter itself.
What are your thoughts on this strategy?

johnraycp:
Unfortunately the CsrfFilter does a String equals on the tokens. Since the whole point of this enhancement is to return a different token value each time the comparison always fails. So at least a minor change to the CsrfFilter is required to call a function to see if the token String in a request matches the token. I had submitted a request with a compare() method add to the CsrfToken interface but it was rejected because it modified an interface.

jgrandja:
@rwinch and I discussed on what the best possible strategy to go with that would be completely passive and the default method strategy in Java 8 looks like the way to go.

For example, we could add the following default method to `CsrfToken`
```
public interface CsrfToken {
    …
    default boolean matches(CsrfToken token) {
        return getToken().equals(token.getToken());
    }
}
```

We can then override the default method in our implementation of `CsrfToken` to handle the XOR logic.

Spring Security 5.0 will support Java 8 so we will have to wait until that release to implement this. So for now I will close this PR but leave the issue open so we keep track of this feature for 5.0 release.

We appreciate your input and welcome more contributions.
Thank you.

Issue #4043, opened 2016-08-27
Improve exception message on Hex#decode


kazuki43zoo:
### Summary

Currently, if detect a non-hex character, it throw exception with message including input value as follow:

```java
throw new IllegalArgumentException("Non-hex character in input: " + s);
```

I think that input value should not be include in exception message because this method is used from class that deal with confidential information (e.g. `PasswordEncoder` or some `Encryptor`).
I think this implementation has a potential risk for disclose a confidential information via application log file.

### Version

All versions on GA.

### Possible Solutions

Change to notify a position of non-hex character as follow:

```java
throw new IllegalArgumentException(
    "Detected a Non-hex character at " + (i + 1) + " or " + (i + 2) + " position");
```

What do you think ?
I will submit a PR at later.

PR #4045, opened 2016-08-29
Improve exception message on Hex#decode

kazuki43zoo:
I've backported gh-4043 to 4.1.x line.


rwinch:
Thanks for the PR! We are not doing any additional releases on 4.0.x as an update to 4.1.x contains only passive changes. We are only applying critical fixes to 4.1.x at this time, so the changes will only be applied to master (4.2.x)

kazuki43zoo:
thanks for comment. I agree.

PR #4062, opened 2016-09-16
Add cookiePath to CookieCsrfTokenRepository

juliovalcarcel:
When using Spring Security to secure a REST API and a JS frontend I run the REST API using Tomcat and the frontend using NPM. Both tomcat and NPM are exposed via an NGINX reverse proxy which forwards / to NPM and /api to tomcat.

I ran into an issue implementing CSRF protection with an AngularJS app in which making a request to tomcat at /api the CSRF cookie's path would be set to /api. In order for Angular to be able to see the cookie the path needs to be set to /. This pull request would allow for the CSRF cookie's path to be set explicitly instead of being derived from the request context and would only set the path if the developer explicitly wanted to otherwise it will default to using the request context.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/csrf/CookieCsrfTokenRepository.java
+++ b/web/src/main/java/org/springframework/security/web/csrf/CookieCsrfTokenRepository.java
@@ -53,7 +53,7 @@ public final class CookieCsrfTokenRepository implements CsrfTokenRepository {
 
 	private boolean cookieHttpOnly;
 
-  private String cookiePath;
+	private String cookiePath;
 
 	public CookieCsrfTokenRepository() {
 		this.setHttpOnlyMethod = ReflectionUtils.findMethod(Cookie.class, "setHttpOnly", boolean.class);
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@juliovalcarcel Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=4062)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=4062) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@juliovalcarcel Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=4062)!

rwinch:
Thanks for the PR! Can you please add some tests too?

juliovalcarcel:
Added tests and also updated the JavaDoc for the setCookiePath method to note that if that value is set it will override the default functionality.

Issue #4065, opened 2016-09-20
HTTP Response Splitting prevention causes NPE


rwinch:
See https://github.com/spring-projects/spring-security/issues/3910#issuecomment-248301126

glavoie:
Rob, after looking a bit at the issue, I found multiple other cases where HTTP Response Splitting is possible, with header name and cookie attributes. I've added multiple tests to cover those cases in my PR. I can open another ticket if you want to cover this change.

rwinch:
@glavoie Thanks. I'd love to cover those changes. I'll create the ticket as we are just about to do our 4.2 M1 release and this will ensure your changes get in.

Thanks again!

rwinch:
Closing in favor of additional commits on #3910 since this has not been released and additional fixes were found

glavoie:
Should I create a new PR with the edited commit comment referring to #3910 instead?

rwinch:
@glavoie Thanks for the offer. I took the liberty of modifying your commit message (leaving you as the author). You can find your changes in 4a1f00b90f540429a924a14c5d64dd2b006e8d2a I also applied some polish of my own via 2e6656e9d3fe99167333a8f6eebb480cac42ca70 Let me know if you have any questions comments.

glavoie:
@rwinch thanks! I added a comment directly in https://github.com/spring-projects/spring-security/commit/2e6656e9d3fe99167333a8f6eebb480cac42ca70 regarding the creation of the test objects. Everything else looks fine to me.

PR #4067, opened 2016-09-21
Issue #4065: HTTP header value check should be null safe.

glavoie:
- Adding multiple test.
- HTTP response splitting should be validated too on cookie attributes and header name.


rwinch:
Thanks for the PR! This was merged via 4a1f00b90f540429a924a14c5d64dd2b006e8d2a with polish from my side in 2e6656e9d3fe99167333a8f6eebb480cac42ca70

Issue #3352, opened 2015-12-22
SEC-3182: Inspect redirected URLs through the filter chain


spring-projects-issues:
[Takuya Iwatsuka](https://jira.spring.io/secure/ViewProfile.jspa?name=iwatsuka) (Migrated from [SEC-3182](https://jira.spring.io/browse/SEC-3182?redirect=false)) said:

We have a request for a new feature.
Briefly, it is that whenever redirected by applications, the target URLs are checked through the filter chain.

We use a custom AuthenticationHandler to achieve the requirements below.
- specify target URLs of redirect after authentication success using request parameter  
- restrict the URLs to be context relative for security reasons

Generally speaking, this kind of check is necessary not only after authentication success, but also for any redirect.
And we are sure that this feature should be provided by Spring Security.
Our proposal is that the filter chain inspects the target URLs.
Though there may be several possible means to do so, using HMAC signing is the best way as far as we know.
( cf. [http://bnbsec.blogspot.jp/2015/03/fixing-open-redirect.html] )

We show our custom AuthenticationHandler just for reference.
But we would like to have more general function that covers other than authentication.
( [https://github.com/terasolunaorg/terasoluna-gfw/blob/master/terasoluna-gfw-security-web/src/main/java/org/terasoluna/gfw/security/web/redirect/RedirectAuthenticationHandler.java])

PR #4074, opened 2016-09-28
Add open redirect prevention

tiwatsuka:
To prevent open redirect by checking signature of the target URL of redirects, this PR add some classes described below.
- RedirectCheckFilter : A servlet filter to wrap responses with SignedRedirectHttpServletResponse
- SignedRedirectHttpServletResponse : A wrapper of servlet response to validate redirect target URL by using signature
- SignCalculator : An interface which executes actual calculation of the signature
- HmacSignCalculator : An implementation of SignCalculator
- RedirectorTag : A JSP tag to generate signature in the page

Issue #3352

### DIFF
--- a/taglibs/src/main/java/org/springframework/security/taglibs/redirect/RedirectorTag.java
+++ b/taglibs/src/main/java/org/springframework/security/taglibs/redirect/RedirectorTag.java
@@ -1,0 +1,124 @@
+/*
+ * Copyright 2002-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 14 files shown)


sjohnr:
Hi, @tiwatsuka. Thank you for your interest in the project!

This PR has been open quite a long time, and we have not observed much activity on the issue or PR, which would typically be a sign that the community may not require this feature (at least not directly in the framework). However, I think the idea has value! To that end, I would encourage you (if you haven't already) to consider creating a small stand-alone library with this filter, and see if there's any interest.

At this time however, I think it's best to close this PR and the corresponding issue. If you or anyone else in the community has additional thoughts, please don't hesitate to let us know.

Issue #4073, opened 2016-09-26
Updating to Jackson 2.7+ Fails


rwinch:
Using Jackson 2.7+ with Spring Security's Jackson Support fails when trying to deserialize authorities. For example, after updating to Jackson 2.8.3 and adding the following added to `SimpleGratnedAuthorityMixinTests`

```java
@Test
@SuppressWarnings("unchecked")
public void deserializeGrantedAuthoritiesTest() throws IOException {
	Set<GrantedAuthority> authorities = mapper.readValue(AUTHORITIES_SET_JSON, Set.class);
	assertThat(authorities).isNotNull();
	assertThat(authorities).extracting("authority").containsOnly("ROLE_USER");
}
```

fails with the following:

```
java.lang.IllegalStateException: No default constructor for [collection type; class java.util.Collections$UnmodifiableSet, contains [simple type, class java.lang.Object]]
	at com.fasterxml.jackson.databind.deser.std.StdValueInstantiator.createUsingDefault(StdValueInstantiator.java:240)
	at com.fasterxml.jackson.databind.deser.std.CollectionDeserializer.deserialize(CollectionDeserializer.java:249)
	at com.fasterxml.jackson.databind.deser.std.CollectionDeserializer.deserialize(CollectionDeserializer.java:26)
	at com.fasterxml.jackson.databind.jsontype.impl.AsArrayTypeDeserializer._deserialize(AsArrayTypeDeserializer.java:110)
	at com.fasterxml.jackson.databind.jsontype.impl.AsArrayTypeDeserializer.deserializeTypedFromArray(AsArrayTypeDeserializer.java:50)
	at com.fasterxml.jackson.databind.deser.std.CollectionDeserializer.deserializeWithType(CollectionDeserializer.java:310)
	at com.fasterxml.jackson.databind.deser.impl.TypeWrappedDeserializer.deserialize(TypeWrappedDeserializer.java:42)
	at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:3788)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:2779)
	at org.springframework.security.jackson2.SimpleGrantedAuthorityMixinTests.deserializeGrantedAuthoritiesTest(SimpleGrantedAuthorityMixinTests.java:68)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #4078, opened 2016-10-03
UnmodifiableSet Deserializer added to support Jackson 2.7+ 

jeetmp3:
Hi @rwinch when updating `jackson-databind` to 2.7+ somehow Jackson is not picking our `UnmodifiableSetMixin` at the time of deserialization.  I've added CustomDeserializer for the same and tested it w/ 2.7+ and older versions.   #4073


rwinch:
Thanks for the PR! This is merged via 48ff518a41bc1174fac7f5deaa5e5a9c60b37f82

Issue #4077, opened 2016-09-29
WebSecurityConfiguration should autowire PermissionEvaluator the same way GlobalMethodSecurityConfiguration does


candrews:
`GlobalMethodSecurityConfiguration` autowires `PermissionEvaluator` from the context:
https://github.com/spring-projects/spring-security/blob/4.1.3.RELEASE/config/src/main/java/org/springframework/security/config/annotation/method/configuration/GlobalMethodSecurityConfiguration.java#L154

`WebSecurityConfiguration` should do the same thing.

Currently, it's surprising that when a `PermissionEvaluator` is set up, it just works (with no configuration other than declaring the `PermissionEvaluator` bean) when used from Java annotations but the same expression always returns denied (as that's what the default configuration does) when used from in a web context (such as in a JSP sec: expression).

rwinch:
Thanks for creating the issue! Would you be interested in submitting a PR?

candrews:
@rwinch PR submitted. Thanks in advance for working with me on this issue.

candrews:
I've updated the PR to include testing.

Note that I've followed the style/approach of #4020

rwinch:
@candrews Thanks!

rwinch:
Thanks for your work on this! This has been merged via 3bf6bf10de959810c33ecbeaf68ded4f1dde53d0  I put polish on it in 8a66d0c78d9f51e2294229ff3c4038dfe5008c73 so that the actual object is not changed but the configuration is updated. This ensures the logic of looking up beans is kept within Configuration and the Bean is kept just a simple Bean.

rwinch:
@candrews This issue is closed. Can you please create a new ticket with details of the issue you are experiencing?

candrews:
> @candrews This issue is closed. Can you please create a new ticket with details of the issue you are experiencing?

I was thinking this issue should be reopened, as the reported issue was not fixed. :-)

rwinch:
Thanks for creating the new ticket. I can understand why you might have thought that reopening this was what needed to be done. To clarify...Once a ticket is associated with a release and that release is performed we do not reopen the issue because the change log is considered immutable at that point. Instead we create a new ticket. Thanks again!

PR #4115, opened 2016-11-01
Configure permissionEvaluator and roleHierarchy by default

candrews:
Configure WebSecurity's default WebSecurityExpressionHandler to use the ApplicationContext's PermissionEvaluator and RoleHierarchy by default, the same way that GlobalMethodSecurityConfiguration does for its default MethodSecurityExpressionHandler.

Issue gh-4077

### DIFF
--- a/core/src/main/java/org/springframework/security/access/expression/AbstractSecurityExpressionHandler.java
+++ b/core/src/main/java/org/springframework/security/access/expression/AbstractSecurityExpressionHandler.java
@@ -40,8 +40,12 @@ public abstract class AbstractSecurityExpressionHandler<T> implements
 		SecurityExpressionHandler<T>, ApplicationContextAware {
 	private ExpressionParser expressionParser = new SpelExpressionParser();
 	private BeanResolver br;
+	private ApplicationContext context;
 	private RoleHierarchy roleHierarchy;
 	private PermissionEvaluator permissionEvaluator = new DenyAllPermissionEvaluator();
+	private boolean roleHierarchySet = false;
+	private boolean permissionEvaluatorSet = false;
+
 
 	public final ExpressionParser getExpressionParser() {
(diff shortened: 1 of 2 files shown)


rwinch:
Thanks for the PR @candrews! Any chance you can add some tests for this?

candrews:
@rwinch I've added tests. If there's anything else I can do, please let me know.

Thanks!

rwinch:
Thanks that is all I need. It will be merged into our next feature release (5.0.0.M1) once we create a new branch for it (next few weeks).

core/src/main/java/org/springframework/security/access/expression/AbstractSecurityExpressionHandler.java
```diff
@@ -102,22 +106,47 @@ protected abstract SecurityExpressionOperations createSecurityExpressionRoot(
 			Authentication authentication, T invocation);
 
 	protected RoleHierarchy getRoleHierarchy() {
+		if(! roleHierarchySet && context != null) {
```
vborcea: maybe extract this if sentence to a conditional method so that the code would be easier to read
e.g. private boolean roleHerarchyNotSetForValidContext(){
! roleHierarchySet && context != null
}

candrews: I don't know how much of an improvement it is, but I've made the change you requested and updated this PR accordingly.

vborcea: It's making the code more easily to read

candrews:
@rwinch Looks like 5.0 is in progress in master... can this be merged?

Thanks!

rwinch:
Thanks for the PR @candrews! This is merged into master

Issue #4110, opened 2016-10-27
Add Referrer-Policy Header support


rwinch:
We should add support for [Referrer-Policy](https://www.w3.org/TR/referrer-policy/) via a header. This would be off by default to remain passive. Examples might be:

``` xml
<http>
    <headers>
        <!-- defaults to using "no-referrer" -->
        <referrer-policy />
    </headers>
</http>
```

``` xml
<http>
    <headers>
        <referrer-policy policy="same-origin"/>
    </headers>
</http>
```

``` java
    http
        .headers()
            // defaults to using "no-referrer"
            .referrerPolicy().and()
```

``` java
    http
        .headers()
            // defaults to using "no-referrer"
            .referrerPolicy(ReferrerPolicy.SAME_ORIGIN).and()
```

The possible values come from https://www.w3.org/TR/referrer-policy/#referrer-policies

Changes should
- Update the XML and Java Config and include tests
- Update the the rnc files used to create the schema the schema (can be generated from the config directory using `../gradlew rncToXsd`
- Update the documentation (both headers section of the docs and the namespace appendix)

NOTE: A good example of the touch points can be found in #3763

eddumelendez:
I'm working on it

PR #4127, opened 2016-11-10
Add useful configuration methods for Referrer Policy header

kazuki43zoo:
I've added some configuration method like the `FrameOptionsConfig` (`deny()`, `sameOrigin()`).
This changes allow the following configuration.

```java
@Override
    http.headers()
            .referrerPolicy().sameOrigin();
}
```

What do you think ?

## Related issues

* gh-4110


rwinch:
Thanks for the PR @kazuki43zoo! I'm going to close this as won't fix. The reasoning is that we are really duplicating the code with very little gain.

kazuki43zoo:
@rwinch Thanks for replying. I agree with your opinion.

Issue #4130, opened 2016-11-14
Add HttpSessionRequestCache sessionAttrName property


rwinch:
We should add a property `sessionAttrName` to `HttpSessionRequestCache` which is the session attribute that the saved request is persited. The default value would be `HttpSessionRequestCache.SAVED_REQUEST`

Related #4111

PR #4131, opened 2016-11-15
Add HttpSessionRequestCache sessionAttrName property

eddumelendez:
This commit allows to customize the session attribute name. Default is
SPRING_SECURITY_SAVED_REQUEST.

Fixes gh-4130

### DIFF
--- a/web/src/main/java/org/springframework/security/web/savedrequest/HttpSessionRequestCache.java
+++ b/web/src/main/java/org/springframework/security/web/savedrequest/HttpSessionRequestCache.java
@@ -32,6 +32,7 @@ import org.springframework.security.web.util.matcher.RequestMatcher;
  * The {@link DefaultSavedRequest} class is used as the implementation.
  *
  * @author Luke Taylor
+ * @author Eddú Meléndez
  * @since 3.0
  */
 public class HttpSessionRequestCache implements RequestCache {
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/springframework/security/web/savedrequest/HttpSessionRequestCache.java
```diff
@@ -129,4 +131,15 @@ public void setCreateSessionAllowed(boolean createSessionAllowed) {
 	public void setPortResolver(PortResolver portResolver) {
 		this.portResolver = portResolver;
 	}
+
+	/**
+	 * If the {@code sessionAttrName} property is set, the request is stored in
+	 * the session using this attribute name. Default is
+	 * "SPRING_SECURITY_SAVED_REQUEST".
```
stonio: I suggest to use `Defaut is {@value #SAVED_REQUEST}.`

Reference: https://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html#CHDDCDHH

rwinch: @stonio the static constant is package scope, so we will stick with the String in the Javadoc

rwinch:
Thanks! This is merged into master

PR #4148, opened 2016-12-06
Prevent BCrypt NPE

### LINKED ISSUE #4147: BCrypt NPE when salt not provided
### Summary

BCrypt.checkpw(String plaintext, String hashed) should not throw NPE.

### Actual Behavior

The BCrypt.checkpw(String plaintext, String hashed) throws a NPE when there is a null-value provided as the "hashed"-parameter. This is because the underlying hashpw(String password, String salt) expects the salt-paramter to be present as it needs to figure out its length. 

### Expected Behavior

As hashpw(String password, String salt) already throws IllegalArgumentExceptions if an e.g. malformed salt is passed, I would expect it to also throw an IllegalArgumentException if the salt is null.

### Version

4.2.0.RELEASE

### Sample

```
java.lang.NullPointerException
    at org.springframework.security.crypto.bcrypt.BCrypt.hashpw(BCrypt.java:544) ~[spring-security-crypto-4.2.0.RELEASE.jar:4.2.0.RELEASE]
    at org.springframework.security.crypto.bcrypt.BCrypt.checkpw(BCrypt.java:654) ~[spring-security-crypto-4.2.0.RELEASE.jar:4.2.0.RELEASE]
```

jbspeakr:
This PR introduces a simple null check within BCrypt.hashpw(String password, String salt) to properly handle null-value salts in a consistent way and prevent NPEs. 

Fixes #4147

### DIFF
--- a/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCrypt.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCrypt.java
@@ -532,14 +532,19 @@ public class BCrypt {
 	 * @param password the password to hash
 	 * @param salt the salt to hash with (perhaps generated using BCrypt.gensalt)
 	 * @return the hashed password
+	 * @throws IllegalArgumentException if invalid salt is passed
 	 */
-	public static String hashpw(String password, String salt) {
+	public static String hashpw(String password, String salt) throws IllegalArgumentException {
 		BCrypt B;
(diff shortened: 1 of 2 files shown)


rwinch:
Thanks for the PR! This is now merged into master. I also polished the error message to indicate why the salt was bad (i.e. that it could not be null)

jbspeakr:
this was fast, thanks!

PR #4166, opened 2016-12-21
Xenoterracide nullsafe equals

xenoterracide:
provide a nullsafe implimenation of equals and hashcode. Although these are never null, when subclassing and testing equals some 3rd party testing utilities such as equalsverifier may not be aware of this (see https://github.com/jqno/equalsverifier/issues/164)

side note, I'm using 4.1.x can backport. Also I created the branch using githubs online editor, so happy to clean up patches as necessary, if the change is otherwise acceptable.

### DIFF
--- a/core/src/main/java/org/springframework/security/core/userdetails/User.java
+++ b/core/src/main/java/org/springframework/security/core/userdetails/User.java
@@ -200,7 +200,7 @@ public class User implements UserDetails, CredentialsContainer {
 	@Override
 	public boolean equals(Object rhs) {
 		if (rhs instanceof User) {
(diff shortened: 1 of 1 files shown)


rwinch:
Thanks for the PR! The checks do not need to be in place because the constructor throws an exception if the username is null or empty. This means that the username cannot be null.

I'm closing this as invalid

xenoterracide:
I'm aware of the check in the constructor, but since the lack of using a null safe equals causes problems with a code quality checking tool (which is arguably a bug in the tool, which was also reported) and changing the code doesn't hurt anything... anyways not surprised.

PR #4177, opened 2016-12-28
principals and sessionIds should be set using constructor so that can…

codecracker2014:
… be shared across node in cluster

As principals and sessionIds are set in class itself so one can't share session count of user across nodes(Cluster). By setting principals and sessionIds we can pass Cache map to constructor which can enable common session count in cluster otherwise user would be allowed to logged in with multiple sessions. There is no point keeping  principals and sessionIds completely internal.

### DIFF
--- a/core/src/main/java/org/springframework/security/core/session/SessionRegistryImpl.java
+++ b/core/src/main/java/org/springframework/security/core/session/SessionRegistryImpl.java
@@ -48,10 +48,22 @@ public class SessionRegistryImpl implements SessionRegistry,
 	protected final Log logger = LogFactory.getLog(SessionRegistryImpl.class);
 
 	/** <principal:Object,SessionIdSet> */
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@codecracker2014 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=4177)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=4177) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@codecracker2014 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=4177)!

PR #4181, opened 2016-12-29
Enabling spring security to handle single session per user in clustered environment.

codecracker2014:
As of now spring security allows single session per user which is working out of box for single node application. This feature won't work in clustered environment as each node has it's separate session count in SessionRegistry. In order to overcome this limitation we can store principals, and sessionIds map in a data grid(synchronized Cache).
As principals and sessionIds are set in class itself so one can't share user session count across nodes(Cluster). Using constructor for setting principals and sessionIds we can pass Cache map to constructor which can enable common session count in cluster otherwise user would be allowed to logged in with multiple sessions.


rwinch:
Thanks for the PR @codecracker2014! This is now in master. As an FYI Spring Session also supports this https://docs.spring.io/spring-session/docs/1.3.1.RELEASE/reference/html5/#spring-security-concurrent-sessions

PR #4192, opened 2017-01-20
Update SecurityContextHolder.java - Use StringUtils#hasText(String)

stonio:
Refactor class `SecurityContextHolder` to use utility method `StringUtils#hasText(String)` to handle blank text.


rwinch:
Thanks for the Pull Request! This is now merged into master :)

PR #4194, opened 2017-01-31
prevent NPE in case of null BeanDefinition bean class name

### LINKED ISSUE #4112: NullPointerException in WebSocketMessageBrokerSecurityBeanDefinitionParser at bootup
### Summary

Using namespace config, added <websocket-message-broker>, NullPointerException is thrown by WebSocketMessageBrokerSecurityBeanDefinitionParser.java:258 at application startup
### Actual Behavior

Been used Spring Security namespace to secure a Spring MVC application, no problem. 
now on to config Spring Security for WebScocket, did the following
1. Added spring-security-messaging.jar
2. Added the following tag:

``` xml
    <websocket-message-broker>
        <intercept-message type="CONNECT" access="permitAll" />
        <intercept-message type="UNSUBSCRIBE" access="permitAll" />
        <intercept-message type="DISCONNECT" access="permitAll" />
        <intercept-message pattern="/**" access="hasRole('USER')" />
    </websocket-message-broker>
```
### Expected Behavior

Application boot up normally and WebSocket is secured
### Configuration

Entire Spring application is configured via namespace, using spring core, spring mvc, spring jpa  and spring security modules.
### Version

Spring version: 4.3.2.Release
Spring Security version: 4.2.0.RC1 and 4.1.0.Release
### Observation:

The code responsible for the NullPointerException is:

``` java
    String[] beanNames = registry.getBeanDefinitionNames();
    for (String beanName : beanNames) {
        BeanDefinition bd = registry.getBeanDefinition(beanName);
        String beanClassName = bd.getBeanClassName();
        if (beanClassName.equals(SimpAnnotationMethodMessageHandler.class
                        .getName()) || beanClassName.equals(WEB_SOCKET_AMMH_CLASS_NAME)) {
...}
```

to be exact, when beanClassName is null, the NullPointerException is thrown.

Now I've observed two occasions where beanClassName is null:
1. when configuring datasource for the app, I use bean definition inheritance for DIY purposes, it goes like:

``` xml
<bean id="parentDataSource" class="com.mchange.v2.c3p0.ComboPooledDataSource"
        destroy-method="close">

<bean id="readDataSource" parent="parentDataSource"></bean>
<bean id="writeDataSource" parent="parentDataSource"></bean>
```

Now somehow readDataSource and writeDataSource don't have their own class, thus they don't have a beanClassName, and hence the NullPointerException. 
I managed to fix this by assigning class to their bean definition like this:

``` xml
<bean id="readDataSource" parent="parentDataSource" class="com.mchange.v2.c3p0.ComboPooledDataSource"></bean>
```

However:
2. the NullPointerException is still thrown, this time, by a bean named "org.springframework.security.config.http.FilterInvocationSecurityMetadataSourceParser$DefaultWebSecurityExpressionHandlerBeanFactory#0$created#0"
I don't know where it's from and how to assign a beanClass to it.

The easiest fix to the problems is apparent:
Either add a null check guard before, or using the reversed equal check: eg:

fix1:

``` java
if (beanClassName != null && (beanClassName.equals(SimpAnnotationMethodMessageHandler.class
                        .getName()) || beanClassName.equals(WEB_SOCKET_AMMH_CLASS_NAME)))
```

fix2:

``` java
if (SimpAnnotationMethodMessageHandler.class
                        .getName().equals(beanClassName) || WEB_SOCKET_AMMH_CLASS_NAME.equals(beanClassName))
```

But i dunno if there's more bug related to beanClassName being null deep down.

jkuipers:
This fixes #4112, which can occur for a number of reasons. It's completely trivial, it just turns the `beanClassName`, which can be `null`, into the operand of the `equals()` invocations to make them null-safe.

I was surprised to see that this hadn't been fixed yet, so I'm assuming the issue was simply overlooked. We ran into this as well, and AFAICT it's simply valid for a bean definition's bean class name to be `null`.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/websocket/WebSocketMessageBrokerSecurityBeanDefinitionParser.java
+++ b/config/src/main/java/org/springframework/security/config/websocket/WebSocketMessageBrokerSecurityBeanDefinitionParser.java
@@ -255,8 +255,8 @@ public final class WebSocketMessageBrokerSecurityBeanDefinitionParser implements
 			for (String beanName : beanNames) {
 				BeanDefinition bd = registry.getBeanDefinition(beanName);
 				String beanClassName = bd.getBeanClassName();
-				if (beanClassName.equals(SimpAnnotationMethodMessageHandler.class
-						.getName()) || beanClassName.equals(WEB_SOCKET_AMMH_CLASS_NAME)) {
+				if (SimpAnnotationMethodMessageHandler.class.getName().equals(beanClassName) ||
+						WEB_SOCKET_AMMH_CLASS_NAME.equals(beanClassName)) {
 					PropertyValue current = bd.getPropertyValues().getPropertyValue(
 							CUSTOM_ARG_RESOLVERS_PROP);
 					ManagedList<Object> argResolvers = new ManagedList<Object>();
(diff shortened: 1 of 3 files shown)


rwinch:
@jkuipers Thanks for the PR! Any chance you can create tests to ensure this doesn't break again?

jkuipers:
Sure: I've added a simple test which doesn't assert anything but simply ensures that there'll not be any NPEs when calling the updated code with a registry containing a bean definition without a class. I've also patched an existing test, as it was failing on Windows b/o hardcoded line ending assumptions.

rwinch:
Thanks for the PR! This is now fixed in master. I removed the windows fix since this was resolved in another ticket and squashed the commits

PR #4206, opened 2017-02-09
Renaming Bcrypt methods to follow JavaNamingStandards

vborcea:
* removing unused methods from test


rwinch:
Thanks for the PR @vborcea!

This is a copy of BCrypt from mindrot, so we prefer to keep it exactly as is to ensure merging is easy. Also, we cannot rename public methods. For these reasons we are going to leave this as is.

Issue #4220, opened 2017-02-18
A default Realm is different with XML and Java Config


kazuki43zoo:
### Summary

A default Realm name of Basic Authentication is different with XML and Java Config.

XML Namespace : `"Spring Security Application"`
Java Config : `"Realm"`

### Expected Behavior

I will expect the `"Realm"`( same value using Java Config).

### Version

* 4.2.1.RELEASE

PR #4221, opened 2017-02-18
Change a default realm name to 'Realm'

kazuki43zoo:
Change a default realm name of Basic Authentication for XML namespace to 'Realm'.

Fixes gh-4220


rwinch:
Thanks for the PR @kazuki43zoo! This is now in master

PR #4229, opened 2017-02-24
Fix warning from spring-boot plugin

eddumelendez:


rwinch:
Thanks for the  PR! This is now merged into master and will be part of 4.2.2

Issue #2742, opened 2014-03-19
SEC-2528: Support PBKDF2 SHA256 for JDK8+


spring-issuemaster:
Migrated from [SEC-2528](https://jira.spring.io/browse/SEC-2528?redirect=false)

Issue #2383, opened 2013-04-12
SEC-2158: PasswordEncoder for PBKDF2WithHmacSHA1


spring-projects-issues:
Migrated from [SEC-2158](https://jira.spring.io/browse/SEC-2158?redirect=false)

PR #4232, opened 2017-02-28
Fully customizable PBKDF2 password encoder

gwallet:
Hi,

I had once to implement a new org.springframework.security.crypto.password.PasswordEncoder to fulfill the following security requirements:
- Hash algorithm: PBKDF2
- PRF: HMAC-SHA256
- Hash length: 256 bits
- Salt length: 64 bits
- Iteration count: 100,000

Keeping an eye on the project to migrate on an existing one, I find out the new one org.springframework.security.crypto.password.Pbkdf2PasswordEncoder was close to fit but suffer from a lack of customization regarding to the above requirements.

Here comes a suggestion to enable the use of PBKDF2 as password encoder algorithm (like mentioned in #2383) with some settings improvement.
The new PBKDF2 password encoder :
- makes PBKDF2 algorithm a first class citizen (like BCrypt)
- uses one of the PBKDF2 flavor supported by the platform (like HMAC-SHA256 and HMAC-SHA512 [starting from JDK 8](http://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html#SecretKeyFactory))
- uses the given salt/hash length, iterations count
- uses the given codec in order to represent the hash as Base64 or Hexadecimal, ... or whatever

Because it's not fully backward compatible with the previous one (no secret key), I choose to  deprecate it in and create a new dedicated package.

I tried to follow coding guideline, to test it and to document as much as possible, any review's feedback will be welcome.

### DIFF
--- a/crypto/src/main/java/org/springframework/security/crypto/codec/Codec.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/codec/Codec.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2011-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.crypto.codec;
+
+/**
+ * A {@code Codec} is responsible for translating byte array into {@link String} and a {@link String} back into byte array in a compatible manner.
+ * <p>
+ * Compatible manner means:
+ * <dl>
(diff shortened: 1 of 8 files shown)


pivotal-issuemaster:
@gwallet Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=4232)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=4232) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@gwallet Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=4232)!

gwallet:
It also fixes #2742

rwinch:
Thanks for the PR @gwallet! While we want to support customization PBKDF2, we don't want to deprecate our existing implementation if we do not need to (I don't think we do).  We also would like to avoid duplication since we already have utilities for [Base64 encoding](https://github.com/spring-projects/spring-security/blob/5.0.0.M5/crypto/src/main/java/org/springframework/security/crypto/codec/Base64.java) (which are deprecated since in JDK 8 provides encoding utilities already).

Ultimately, I agree with the the goal of this PR (ability to customize PBKDF2), but I would like to try to do this in a less invasive way.

PR #4235, opened 2017-03-02
Fixes #4234 - Minor improvements to UrlUtils

### LINKED ISSUE #4234: UrlUtils. isAbsoluteUrl should reuse the Pattern rather than recompiling it
This pattern should be a constant in the class. There is no need to recompile it for each invocation of this method. 

https://github.com/spring-projects/spring-security/commit/017e9834bd5e0ad4b59f2a251ed25c492ab88645#commitcomment-21117632

martin-g:
Fix for #4234 and minor improvements in the usage of StringBuilder

### DIFF
--- a/web/src/main/java/org/springframework/security/web/util/UrlUtils.java
+++ b/web/src/main/java/org/springframework/security/web/util/UrlUtils.java
@@ -29,6 +29,10 @@ import javax.servlet.http.HttpServletRequest;
  * @author Ben Alex
  */
 public final class UrlUtils {
+	// ~ Constants
+	private static final Pattern ABSOLUTE_URL = Pattern.compile("\\A[a-z0-9.+-]+://.*",
+	                                                            Pattern.CASE_INSENSITIVE);
+
 	// ~ Methods
 	// ========================================================================================================
 
@@ -61,7 +65,7 @@ public final class UrlUtils {
 		}
 		else if ("https".equals(scheme)) {
 			if (serverPort != 443) {
-				url.append(":").append(serverPort);
+				url.append(':').append(serverPort);
 			}
 		}
 
@@ -70,7 +74,7 @@ public final class UrlUtils {
 		url.append(requestURI);
 
 		if (queryString != null) {
-			url.append("?").append(queryString);
+			url.append('?').append(queryString);
 		}
 
 		return url.toString();
@@ -119,7 +123,7 @@ public final class UrlUtils {
 		}
 
 		if (queryString != null) {
-			url.append("?").append(queryString);
+			url.append('?').append(queryString);
 		}
 
 		return url.toString();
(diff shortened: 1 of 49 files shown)


pivotal-issuemaster:
@martin-g Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=4235)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=4235) to manually synchronize the status of this Pull Request.

martin-g:
I've singed the CLA.

Why `./gradlew build` produces diffs like:
```diff
diff --git i/acl/pom.xml w/acl/pom.xml
index 6ecd84a..a75abab 100644
--- i/acl/pom.xml
+++ w/acl/pom.xml
@@ -3,7 +3,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.springframework.security</groupId>
   <artifactId>spring-security-acl</artifactId>
-  <version>4.2.2.RELEASE</version>
+  <version>4.2.3.BUILD-SNAPSHOT</version>
   <name>spring-security-acl</name>
   <description>spring-security-acl</description>
   <url>http://spring.io/spring-security</url>
@@ -51,7 +51,7 @@
     <dependency>
       <groupId>org.springframework.security</groupId>
       <artifactId>spring-security-core</artifactId>
-      <version>4.2.2.RELEASE</version>
+      <version>4.2.3.BUILD-SNAPSHOT</version>
       <scope>compile</scope>
     </dependency>
     <dependency>
@@ -151,6 +151,12 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+  <repositories>
+    <repository>
+      <id>spring-snapshot</id>
+      <url>https://repo.spring.io/snapshot</url>
+    </repository>
+  </repositories>
   <build>
     <plugins>
       <plugin>
```
Did I use the wrong branch ?

rwinch:
Thanks for the PR @martin-g!

> Why ./gradlew build produces diffs like:

This was because we had just done a release and the poms had not been regenerated since. Please pull in the latest from master and the pom will not be changed.

pivotal-issuemaster:
@martin-g Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=4235)!

web/src/main/java/org/springframework/security/web/util/UrlUtils.java
```diff
@@ -61,7 +65,7 @@ public static String buildFullRequestUrl(String scheme, String serverName,
 		}
 		else if ("https".equals(scheme)) {
 			if (serverPort != 443) {
-				url.append(":").append(serverPort);
```
martin-g: See `java.lang.AbstractStringBuilder#append(java.lang.String)` vs `java.lang.AbstractStringBuilder#append(char)`.
The latter just does `value[count++] = c;` while the String version calls `java.lang.String#getChars(int, int, char[], int)` which always does `System.arraycopy(value, srcBegin, dst, dstBegin, srcEnd - srcBegin);`.
So the `char` version is much cheaper. But the gain is noticeable only when there are many concatenations.

rwinch: I don't think this is worth the inconsistency in the code base. If it were, then I'm sure the `append(String)` method could call `append(char)` if it were a single character.

Can you please revert the changes to the append? At that point I we can merge the fix for `Pattern`

PR #4253, opened 2017-03-17
Improved Jackson support for Spring Session

### LINKED ISSUE #4252: Spring Session integration only supports User as the principal
### Summary

In a Spring Boot application using both Spring Security and Spring Session (with Redis and JSON serialization), a  `UsernamePasswordAuthenticationToken` with a custom class as the principal can be serialized, but not deserialized.

### Actual Behavior

The deserialization produces the following error:

> Could not read JSON: Unexpected token (END_OBJECT), expected FIELD_NAME: missing property '@class' that is to contain type id  (for class java.lang.Object)\n at [Source: N/A; line: -1, column: -1] (through reference chain: org.springframework.security.core.context.SecurityContextImpl["authentication"]); nested exception is com.fasterxml.jackson.databind.JsonMappingException: Unexpected token (END_OBJECT), expected FIELD_NAME: missing property '@class' that is to contain type id  (for class java.lang.Object)
> at [Source: N/A; line: -1, column: -1] (through reference chain: org.springframework.security.core.context.SecurityContextImpl["authentication"])

### Expected Behavior

Not getting an error when the session is deserialized from Redis.

### Configuration

N/A

### Version

Spring Security 4.2.2
Spring Session 1.3.0

### Sample

N/A

fpavageau:
The pull-request addresses 2 issues:
- Incorrect use of the Jackson API, by using `JsonNode.toString()` to serialize a sub-tree before deserializing it using an `ObjectMapper`.
- Forced deserialization of the principal as `User`.

This fixes #4252.

- [x] I have signed the CLA

### DIFF
--- a/core/src/main/java/org/springframework/security/jackson2/UnmodifiableSetDeserializer.java
+++ b/core/src/main/java/org/springframework/security/jackson2/UnmodifiableSetDeserializer.java
@@ -50,10 +50,10 @@ class UnmodifiableSetDeserializer extends JsonDeserializer<Set> {
 				Iterator<JsonNode> nodeIterator = arrayNode.iterator();
 				while (nodeIterator.hasNext()) {
 					JsonNode elementNode = nodeIterator.next();
-					resultSet.add(mapper.readValue(elementNode.toString(), Object.class));
+					resultSet.add(mapper.readValue(elementNode.traverse(mapper), Object.class));
 				}
 			} else {
-				resultSet.add(mapper.readValue(node.toString(), Object.class));
+				resultSet.add(mapper.readValue(node.traverse(mapper), Object.class));
 			}
 		}
 		return Collections.unmodifiableSet(resultSet);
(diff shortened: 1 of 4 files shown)


fpavageau:
By the way, I'm really interested in having the fix integrated in 4.2, I can rebase my branch on `4.2.x` if necessary or helpful.

fpavageau:
The branch has been rebased on master and adapted to the changes in #4370.

rwinch:
Thanks for the PR @fpavageau! This is merged into master. If you would like to submit a PR for 4.2.x, I'd be glad to merge it

fpavageau:
Hi @rwinch, thanks for merging.

I've submitted #4759 for `4.2.x`.

core/src/main/java/org/springframework/security/jackson2/UnmodifiableSetDeserializer.java
```diff
@@ -50,10 +50,10 @@ public Set deserialize(JsonParser jp, DeserializationContext ctxt) throws IOExce
 				Iterator<JsonNode> nodeIterator = arrayNode.iterator();
 				while (nodeIterator.hasNext()) {
 					JsonNode elementNode = nodeIterator.next();
-					resultSet.add(mapper.readValue(elementNode.toString(), Object.class));
+					resultSet.add(mapper.readValue(elementNode.traverse(mapper), Object.class));
```
cowtowncoder: ... just by chance noticed this. May have been changed since then, but method `mapper.treeToValue()` would perhaps be best method to use here. It does translate to very similar call so it's not a big deal either way, and code above works fine.

Issue #2251, opened 2012-08-07
SEC-2026: MethodSecurityExpressionRoot should be a public class


spring-projects-issues:
[Clark Duplichien](https://jira.spring.io/secure/ViewProfile.jspa?name=cduplichien) (Migrated from [SEC-2026](https://jira.spring.io/browse/SEC-2026?redirect=false)) said:

Similar to SEC-1691, except don't need to specify as part of config... just need it to be public.
Currently, you can extend DefaultMethodSecurityExpressionHandler and provide the reference to the expression-handler element of the global-method-security config, but even if you override the createSecurityExpressionRoot method, you can't provide an instance that extends MethodSecurityExpressionRoot, because it's protected.
This is also inconsistent with MethodSecurityExpressionRoot, which is already public.

OrangeDog:
It really should be easier to add to the expression context than having to completely re-implement this class.

rwinch:
What exactly do you want to expand? Have you tried using a Bean. For example, you can do something like:

```
@beanName.methodThatReturnsBoolean(some args...)
```

OrangeDog:
It's a bit much just to add a couple more string constants to the existing `read`, `write`, etc.

rwinch:
I'm not sure why it is a bit much. You can accomplish this goal by creating a single Bean. One option is to do something like:

```java
@Component
public class Authz {
    public boolean hasReadAccess(MyUser myUser, Object object) {
        // ...
       return result;
    }
    public boolean hasWriteAccess(MyUser myUser, Object object) {
        // ...
       return result;
    }
}
```

Then you can leverage the bean using:

```java
@PostAuthorize("@authz.hasReadAccess(principal, returnObject)")
```

This has the additional benefit of not being coupled to the Spring Security APIs since the Bean you have defined is relying only on your domain objects (rather than having Spring Security's Authentication object in the method signature).

Alternatively, you can create a bean with the values on it and refer to it that way. For example:

```java
@Component
public class Permissions
    public final String read = "read";
    public final String write = "write";
}
```

Then you can leverage it using:

```java
@PostAuthorize("hasPermission(returnObject, @permissions.read")
```

If you are wanting to create a custom `MethodSecurityExpressionRoot`, you would need to create at least one Bean.

[...]

PR #4266, opened 2017-04-05
SEC-2026: MethodSecurityExpressionRoot should be a public class

simplicii:
changed class accessor from package-private to public

#2251 SEC-2026: MethodSecurityExpressionRoot should be a public class
Changed class accessor to public as requested. It is anoying to have to copy the code just because its package-private.
#SEC-2026:


eleftherias:
Thanks for the PR @simplicii.
As mentioned in gh-2251, we are not looking to change `MethodSecurityExpressionRoot` to a public class.
The same goal can be accomplished by using a bean, as demonstrated in [this comment](https://github.com/spring-projects/spring-security/issues/2251#issuecomment-248404307).
If there are any further concerns, feel free to add them to the associated issue.

PR #4268, opened 2017-04-07
Provide AuthenticatedPrincipal abstraction

jgrandja:
The goal for the `AuthenticatedPrincipal` abstraction is to decouple `Principal` type of abstractions from    `AbstractAuthenticationToken`, specifically `AbstractAuthenticationToken.getName()`.

As of now we are referencing `UserDetails` in `AbstractAuthenticationToken.getName()` and it would be nice to abstract this away to a generic interface that would provide the name of the `Principal` for any representation of a `Principal`, for example, `UserDetails`, _OAuth2 User_, _OpenID Connect User_, etc.

### DIFF
--- a/core/src/main/java/org/springframework/security/core/AuthenticatedPrincipal.java
+++ b/core/src/main/java/org/springframework/security/core/AuthenticatedPrincipal.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2012-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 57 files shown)


core/src/main/java/org/springframework/security/core/AuthenticatedPrincipal.java
```diff
@@ -0,0 +1,53 @@
...
+ * @see org.springframework.security.core.userdetails.UserDetails
+ */
+public interface AuthenticatedPrincipal {
+
+	/**
+	 * Returns the identifier for the authenticated <code>Principal</code>. Never <code>null</code>.
+	 *
+	 * @return the identifier for the authenticated <code>Principal</code>
+	 */
+	String getId();
```
rwinch: We don't need an id, so we shouldn't have it

rwinch:
Thanks...that's all the feedback I had. Given we discussed the id and you agreed go ahead and make that change and then you can merge :)

core/src/main/java/org/springframework/security/core/AuthenticatedPrincipal.java
jgrandja: getId() has been removed

PR #4274, opened 2017-04-13
Polish AssertJ assertions

stonio:
Revamp AssertJ assertions with cool methods such as `hasSize(...)` `isZero()` `startsWith(...)` `containsExactly(...)`.


rwinch:
Awesome! Thank you very much @stonio :) This is now merged into master

Issue #4276, opened 2017-04-13
add ABAC to spring security (Attribute Based Access Control)


simplicii:
### Summary

ABAC is a standard published by NIST (US National Institute od Standarts and Technology) in cooperation with industry and universities.
ABAC grants access based on user attributes and policies.

Wikipedia: https://en.wikipedia.org/wiki/Attribute-Based_Access_Control
NIST: http://csrc.nist.gov/projects/abac/

### Actual Behavior

Actual we only have RBAC (Role based Access Control) and ACL (Access Controll List)

### Expected Behavior

Implement a spring-abac module to support ABAC policies 

### Sample

```
[
  {
    "name": "Liquor shop for adults only",
    "type": "",
    "description": "grant access to liquor store if user is older than 21",
    "applicable": "resource.type=='LIQUOR_STORE'",
    "condition": "subject.principal.age>21"
  }
]
```

```
@RequestMapping(path = "shops")
    @PreAuthorize("hasRole('USER') and checkPolicy()")
    public List<Shop> listShops(){
        ....
    }
```

grunberg:
Where did ABAC for Spring security go?

PR #4277, opened 2017-04-13
ABAC support for spring security

simplicii:
ABAC is a standard published by NIST (US National Institute od Standarts and Technology) in cooperation with industry and universities.
ABAC grants access based on user attributes and policies.

Updated spring manual: https://github.com/simplicii/spring-abac-sample/blob/master/spring-security-reference.pdf

Wikipedia: https://en.wikipedia.org/wiki/Attribute-Based_Access_Control
NIST: http://csrc.nist.gov/projects/abac/

This is the implemenation for #4276


rwinch:
@simplicii Thank you for the PR! 

Can you elaborate how this will work with large data sets? I'm curious what your real world experience is when using ABAC is? Historically I have never been able to use authorization which happens after the database query because it will not perform. Obviously Spring Security ACL behaves the same way, but it is too late to change that.

It may be worth mentioning that rather than extending the expression root, I typically recommend users to [refer to another bean](https://docs.spring.io/spring-security/site/docs/current/reference/htmlsingle/#el-access-web-beans) in the expression using `@beanName.methodThatReturnsBoolean(...)`.

simplicii:
As you said, acl behaves unfortunately the same way. Thats why I did it this way too.
For @Pre operation it should be no problem and the PostFilter will have the same issue than the ACL.
Fortunately I only used ACL on MickeyMouse projects so far. 
I think I would anyway use Native Sql queries to perform ACL on bigger data sets to let DB optimize it.
Honestly, your proposal to use [refer to another bean](https://docs.spring.io/spring-security/site/docs/current/reference/htmlsingle/#el-access-web-beans) . Might be more flexible anyway.  Possible that this is obsolete from a different perspective.

The Idea to write spring-abac was to have something inbetween ROLES and ACL.

rwinch:
Thank you for your response.

I'm somewhat hesitant to add support for something that would not work on large data sets. I realize that Spring ACL has the same limitations, but that decision was made before I had a vote.

> Honestly, your proposal to use refer to another bean might be more flexible anyway. Possible that this is obsolete from a different perspective.

This is probably the best option. Perhaps we need to make this more obvious in the documentation that it is available.

simplicii:
Makes sense if you could update the documentation to make it more obvious.
I will delete my PR for now and see how that goes.

ugultopu:
What is the end decision here? Will Spring Security have ABAC? Or should we simulate ABAC using SpEL? Will Spring Security have XACML support?

rwinch:
@ugultopu For now users should leverage SpEL expressions. If someone can describe a way to use ABAC in a large scale system, we can reconsider adding ABAC support.

Kiddinglife:
[](https://www.openpolicyagent.org/)

Issue #4311, opened 2017-04-27
JndiDnsResolver doesn't support resolving all hostnames for a service


hansvanbeneden:
Recently I posted a [question on stackoverflow](http://stackoverflow.com/questions/43604856/how-to-resolve-all-active-directory-server-urls-with-jndidnsresolver) about resolving the hostnames for a service with JndiDnsResolver. I didn't got any response so I decided to try and fix it myself. 

The actual behavior is that the JndiDnsResolver.resolveServiceEntry only returns one hostname (with the highest priority and weight). I want to add a method that returns a list of hostnames ordered by priority and weight.

Example:
If the DNS server provides the following the service records:
```
_ldap._tcp.springsource.com IN SRV 10 40 88 ldap2.springsource.com.
_ldap._tcp.springsource.com IN SRV 10 70 88 ldap1.springsource.com.
_ldap._tcp.springsource.com IN SRV 20 50 88 ldap3.springsource.com.

```
Then I would expect a method that returns all the hostnames ordered by priority and weight like this:
`[ldap1.springsource.com, ldap2.springsource.com, ldap3.springsource.com]`

eleftherias:
@hansvanbeneden I do not think a method that retrieves all hostnames belongs in the DnsResolver class. The DnsResolver should return a single hostname based on priority and weight, and the user should not need to know about the internal implementation.  
I have left a comment on your PR as well with more details.

spring-issuemaster:
If you would like us to look at this issue, please provide the requested information. If the information is not provided within the next 7 days this issue will be closed.

spring-issuemaster:
Closing due to lack of requested feedback. If you would like us to look at this issue, please provide the requested information and we will re-open the issue.

PR #4314, opened 2017-04-28
added support for resolving all hostnames for a service to JndiDnsResolver

hansvanbeneden:
I'm using the ActiveDirectoryLdapAuthenticationProvider to authenticate against Active Directory. The ActiveDirectoryLdapAuthenticationProvider constructor takes multiple ldap urls. I need to resolve these ldap urls based on the ldap domain, I found the JndiDnsResolver which does exactly that. But JndiDnsResolver.resolveServiceEntry returns only 1 hostname (with the highest priority). I want a method that returns all hostnames (ordered on priority and weight), since some of the servers could be down. I added two methods: JndiDnsResolver.resolveAllServiceEntries() and JndiDnsResolver.resolveAllServiceIpAddresses(). I didn't remove the existing JndiDnsResolver to obtain backward compatibility.

This is the implementation for #4311

### DIFF
--- a/remoting/src/main/java/org/springframework/security/remoting/dns/DnsResolver.java
+++ b/remoting/src/main/java/org/springframework/security/remoting/dns/DnsResolver.java
@@ -16,6 +16,8 @@
 
 package org.springframework.security.remoting.dns;
 
+import java.util.List;
+
 /**
  * Helper class for DNS operations.
  *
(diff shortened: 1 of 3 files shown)


eleftherias:
Thanks for the PR @hansvanbeneden.  
However, I do not think a method that retrieves all hostnames belongs in the `DnsResolver` class. The `DnsResolver` should return a single hostname based on priority and weight, and the user should not need to know about the internal implementation.  
In order to proceed, I recommend creating a new class/interface that can return a list of hostnames. Then, the `DnsResolver` could use that class for the lookup.  
 Alternatively, you can create a custom implementation of `DnsResolver` with any additional methods you need.

eleftherias:
Closing due to inactivity and lack of feedback.
A workaround is provided in the previous comment.
Please let us know if the issue persists and we will re-open it.

PR #4331, opened 2017-05-11
Extending AuthorityUtils with String list to GrantedAuthority list converter

lpalonek:
AuthorityUtil suffer from tool that enables conversion from list of strings to GrantedAuthority. Right now it enables only conversion from array of strings which forces user that works with List to use some odd conversions like this

`AuthorityUtils.createAuthorityList(listOfAuthorities.toArray(new String[0]))`

### DIFF
--- a/core/src/main/java/org/springframework/security/core/authority/AuthorityUtils.java
+++ b/core/src/main/java/org/springframework/security/core/authority/AuthorityUtils.java
@@ -21,6 +21,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


eleftherias:
@lpalonek thank you for the PR.
Do you have a scenario in mind where the authorities have to be in list?
Throughout the Spring Security codebase the authorities are always passed in as one or multiple String parameters. 
Additionally, it seems that it would be fairly simple to convert a list of `String`s to a list of `GrantedAuthority`s by applying a map function to the `SimpleGrantedAuthority` constructor, and therefore using `AuthorityUtils` wouldn't add much value.

eleftherias:
Closing due to inactivity and lack of feedback.

PR #4365, opened 2017-06-02
Fix for UserInfoTypeConverter auto-registration

### LINKED ISSUE #4364: UserInfoTypeConverter is registered only if client is registered
### Summary

OAuth2LoginAutoConfiguration is registering all UserInfoConverter objects configured in META-INF/oauth2-clients-defaults.yml even for unused oauth2 clients. This will throw exception for clients like github, facebook which don't have user-info-name-attribute-key configuration key configured

### Actual Behavior

OAuth2LoginAutoConfiguration is registering all UserInfoConverter objects configured in META-INF/oauth2-clients-defaults.yml even for unused oauth2 clients.

### Expected Behavior

OAuth2LoginAutoConfiguration should register only UserInfoConverter objects for configured  OAuth2 clients (with client-id key)

### Version

Spring Security 5-SNAPSHOT

tha2015:
Fixes #4364


jgrandja:
Thank you for that update @tha2015. This is now in master.

tha2015:
Thanks Joe.

PR #4391, opened 2017-06-13
Okta: Remove Add Authorization Server section since it's not needed

mraible:
Update docs to remove **Add Authorization Server** section since it's not needed.


jgrandja:
Thanks @mraible !

PR #4424, opened 2017-06-28
Allow use of non-numeric (e.g. UUID) values for ObjectIdentity.getIdentifier()

### LINKED ISSUE #1224: SEC-972: Allow use of non-numeric (i.e. string) values for ObjectIdentity.getIdentifier() with the default JDBC implementation
["Zachary Roadhouse":https://jira.spring.io/secure/ViewProfile.jspa?name=zroadhouse](Migrated from ["SEC-972":https://jira.spring.io/browse/SEC-972?redirect=false]) said:

I am integrating the Spring Security ACL module into our application which uses a mixture of UUID and numeric primary keys as our domain object identifiers.  Based on reading the interface documentation fo ObjectIdentity, I had hoped that our UUID identifiers would work with the JdbcMutableAclService implementation without any customizations to the source.  This is not currently true as the JdbcMutableAclService and related classes all assume that the identifier value can be converted to a Long.  This assumption is limited to a small number of points in the implementation.

This is a request to change the implementation to allow the choice of Long vs. String to be made into a configuration setting (or alternate implementation).

pwheel:
This PR fixes #1224.

I discussed the general idea for this change over a year ago with @rwinch on gitter and shared https://github.com/pwheel/spring-security/commit/f3882817ae97c787258e833aa77a0b8933903296#comments with him.

This took a while to get ready for a PR as when I got round to writing the tests properly, there was a problem with how HSQLDB was handling converting UUID to varchar so I had to submit a fix for that which came out in HSQLDB 2.4.0 earlier this year.

The change allows for the fully qualified class name of the ObjectIdentity identifier type to be specified in the acl_class table and a ConversionService can be provided to BasicLookupStrategy to convert from String to the actual identifier type.

There are the following other changes:

1. BasicLookupStrategy has a new property, aclClassIdSupported, which
 is used to retrieve the new column from the database. This preserves
 backwards-compatibility, as it is false by default.

2. JdbcMutableAclService has the same property, aclClassIdSupported,
 which is needed to modify the insert statement to write to the
 new column. Defaults to false for backwards-compatibility.

3. Tests have been updated to verify both the existing functionality
 for backwards-compatibility and the new functionality.

Other things to call out:

1. I haven't updated the documentation; I can do that but wanted feedback first as it may result in changes.

2. I've changed the default schemas to use varchar for acl_object_identity.object_id_identity column, but I haven't added the new acl_class.class_id_type column by default. That seems a bit inconsistent in retrospect but again appreciate your feedback here.

3. Master isn't building atm (fails with sample.HelloWebfluxApplicationTests on my machine at least) but the tests for the acl module pass. Apologies for submitting in this state, I wanted to kick-start the process.

Cheers,
Paul


graeme-dougal:
Any idea if / when this will be implemented

pwheel:
Merged changes from master that fixed the broken tests, all tests pass locally now

Andy2003:
I tested this PR and it works for me

sterowney:
:+1:

truongvanluan:
Hello @pwheel , there is still one class needs to be modified, it is JdbcAclService. If you look into its findChildren method (which is used by JdbcMutableAclService), you should see this
String javaType = rs.getString("class");
Long identifier = new Long(rs.getLong("obj_id"));

And this is the obj_id per DEFAULT_SELECT_ACL_WITH_PARENT_SQL
select obj.object_id_identity as obj_id

So because we are using Uuid this cast will cause a problem.
 Can we have this fixed as well ? (or correct me if this is not an issue). Thanks anyway for this PR, really appreciate it!

pwheel:
Thanks @truongvanluan good catch. I updated some of the tests to cover this & confirm was a problem. I've done a little bit of refactoring as needed to use some of the logic in to classes now, so created a utility class AclClassIdUtils to handle it.

I had to merge in master as had compile time errors due to Spring 5 snapshot dependencies, & had to do some updates for checkstyle & Apache licence too.

Commit history isn't so clean now… I can squash / rebase again but wanted to push this up for feedback first. Hope that's alright!

rwinch:
Thanks for the PR @pwheel! This is merged into master via 6decf1c8ef8e31b0d9de9a2f2b364ce682d8b166

pwheel:
That's awesome, thanks @rwinch!

graeme-dougal:
Hi thanks for this, been waiting a while for this to be implemented so been keen to try it out...

I've added `setAclClassIdSupported(true)` to both my `LookupStrategy ` and `JdbcMutableAclService`

However, I'm still getting an error in `AclClassIdUtils` as it fails because there is no `ConversionService` set so it therefore attempts to convert to Long when I have a String

The offending method in `AclClassIdUtils` is
`
private <T> boolean canConvertFromStringTo(Class<T> targetType) {
	return hasConversionService() && conversionService.canConvert(String.class, targetType);
}
`
So my questions is - do you need to set the conversion service and if so how ?  Otherwise, is there something else I am missing ?

Thanks
Graeme

pwheel:
Hi @graeme-dougal, yes you do. If you take a look at the test setup here that should hopefully answer your question - https://github.com/spring-projects/spring-security/blob/63e2db72eabc647d3a750189dc6549d7d060dd63/acl/src/test/resources/jdbcMutableAclServiceTestsWithAclClass-context.xml#L9

Cheers,
Paul

graeme-dougal:
Hi Paul

Thanks for response.  I'm using java config rather than xml config.  My issue with wiring in `AclClassIdUtils` is that the class is not public so cannot be accessed.

I'd like to do the following but this won't compile due to the class not being public.
`
@Bean
    public AclClassIdUtils aclClassIdUtils() {

        AclClassIdUtils classIdUtils = new AclClassIdUtils();
        classIdUtils.setConversionService(new DefaultConversionService());
        
        return classIdUtils
    }
`

Thanks
Graeme

pwheel:
Ah, yeah, got it, doh. I refactored this a couple of times and originally thought this class would be internal.

It should be public now, for sure. Raised #4814 will fix this week

PR #4429, opened 2017-06-29
Removed Unicode Character from Parameter Name

### LINKED ISSUE #4428: Compilation Error on Windows due to Illegal Character
Building Spring Security on Windows fails due to a compilation error stemming from an 'illegal character' contained within [SessionInformationExpiredStrategy](https://github.com/spring-projects/spring-security/blob/master/web/src/main/java/org/springframework/security/web/session/SessionInformationExpiredStrategy.java#L32)

```java
void onExpiredSessionDetected(SessionInformationExpiredEvent eventØ)
			throws IOException, ServletException;
```

The parameter name `eventØ` contains unicode character `\u02dc` which is the culprit.

There may be others but I haven't found an effective method for searching for these characters.

andersonkyle:
Removed Unicode character from a parameter name that caused compilation errors on Windows.

Fixes #4428

### DIFF
--- a/web/src/main/java/org/springframework/security/web/session/SessionInformationExpiredStrategy.java
+++ b/web/src/main/java/org/springframework/security/web/session/SessionInformationExpiredStrategy.java
@@ -29,6 +29,6 @@ import javax.servlet.ServletException;
  */
 public interface SessionInformationExpiredStrategy {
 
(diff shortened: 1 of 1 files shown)


rwinch:
Thanks for the PR! This is now fixed in master

PR #4436, opened 2017-07-06
Fix the JavaDoc for Pbkdf2PasswordEncoder so that it uses the actual …

jorispz:
This small pull request updates the (outdated?) JavaDoc in Pbkdf2PasswordEncoder, so that it uses the actual values for the default number of iterations and hash width

### DIFF
--- a/crypto/src/main/java/org/springframework/security/crypto/password/Pbkdf2PasswordEncoder.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/password/Pbkdf2PasswordEncoder.java
@@ -53,7 +53,7 @@ public class Pbkdf2PasswordEncoder implements PasswordEncoder {
 
 	/**
 	 * Constructs a PBKDF2 password encoder with no additional secret value. There will be
-	 * 360000 iterations and a hash width of 160. The default is based upon aiming for .5
+	 * {@value DEFAULT_ITERATIONS} iterations and a hash width of {@value DEFAULT_HASH_WIDTH}. The default is based upon aiming for .5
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@jorispz Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=4436)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=4436) to manually synchronize the status of this Pull Request.

jorispz:
Signed the CLA!

pivotal-issuemaster:
@jorispz Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=4436)!

rwinch:
@jorispz Thanks for the PR!

I don't think we want to do that because the constants are private at the moment. Perhaps we should update the constants to be public at which time these changes would make sense. Could you update your PR?

jorispz:
@rwinch you're welcome!

I'm not sure I understand your comment. The @value tag in JavaDoc inserts the *value* of the referenced variable, not its name, regardless of whether it is public or private. If you look at the generated JavaDoc, you will see the actual values. In essence, it's a little truck to make the JavaDoc always be in sync with the actual values used as default.

rwinch:
Thanks. You are right. This is now merged in via de9fe3e3b1496e1e873a1e3e98f56b043b33e43c

Issue #4434, opened 2017-07-05
Add support for JWT/JWS


jgrandja:
We need to support JWT as the Id Token is a JWT.

We also need the capability of verifying the signature of a JWS (JSON Web Signature).

Issue #4410, opened 2017-06-21
Leverage id_token for user authentication


jgrandja:
For OpenID Connect providers, we should leverage the `id_token` to authenticate the user and therefore not needing to call the UserInfo endpoint for obtaining user information.

jgrandja:
Related #4312

Issue #4409, opened 2017-06-21
Expose configuration for authorities mapping


jgrandja:
We should expose a configuration for setting a custom `GrantedAuthoritiesMapper` so the user is able to map OIDC claims contained in the `IdToken` and `UserInfo` to one or more `GrantedAuthority`(s).

For standard OAuth2 providers, the user would map the user attributes fetched from the _UserInfo Endpoint_ to one or more `GrantedAuthority`(s).

jgrandja:
Related #4373

Issue #4312, opened 2017-04-27
UserInfo vs IdToken


jgrandja:
The default implementation of `UserInfo` is `DefaultUserInfo` which uses `StandardClaimName`.

`StandardClaimName` are [standard claims](http://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) defined by the _OpenID Connect_ specification. These claims may be used in a _UserInfo Response_ or an _ID Token_. We don't have a construct for _ID Token_ as of yet but that will be coming soon. So the question...is `UserInfo` the same as `IdToken` (future class)? Or does `IdToken` have more use cases around it than `UserInfo`?

jgrandja:
Closed via #4410

PR #4439, opened 2017-07-07
Support for JWT/JWS and id_token

jgrandja:


jgrandja:
Related #4434 #4410 #4409 #4312

samples/boot/oauth2login/src/main/resources/application.yml
```diff
@@ -21,14 +21,15 @@ security:
       github:
         client-id: your-app-client-id
         client-secret: your-app-client-secret
-        user-info-name-attribute-key: "name"
+        user-name-attribute-name: "name"
```
rwinch: While `user-name-attribute-name` is an option, we shouldn't list it for each of these because they should be defaulted for us. We know a reasonable default value for github, facebook, etc.

Issue #2448, opened 2013-07-23
SEC-2224: ActiveDirectoryLdapAuthenticationProvider throws BadCredentialsException if userPrincipalName not equal to sAMAccountName + @domain


spring-issuemaster:
[Michael Solano](https://jira.spring.io/secure/ViewProfile.jspa?name=mike.solano.developer@gmail.com) (Migrated from [SEC-2224](https://jira.spring.io/browse/SEC-2224?redirect=false)) said:

When using the sAMAccountName for authentication via ActiveDirectoryLdapAuthenticationProvider, a BadCredentialsException will be thrown if the userPrincipalName is not the sAMAccountName with @domain post-fixed.

For example, if the sAMAccountName is "bwayne" but the userPrincipalName is "bruce.wayne@batcave.net", authentication will fail. The createBindPrincipal method assumes the userPrincipalName will be "bwayne@batcave.net" and not "bruce.wayne@batcave.net". 

The code below shows the details of that method:

```
    String createBindPrincipal(String username) {
        if (domain == null || username.toLowerCase().endsWith(domain)) {
            return username;
        }

        return username + "@" + domain;
    }
```

gkibilov:
Still an issue using Spring 4.0.3, injecting sAMAccountName filter has no effect as searchForUser still uses bindPrincipal when calling searchForSingleEntryInternal!

taasjord:
This is still an issue. In my case the userPrincipalName has a different suffix than domain, ie. while bindPrincipal is "bwayne@batcave.net", the userPrincipalName is "bwayne@gotham.com" and sAMAccountName is "bwayne".

So I need to pass username to match sAMAccountName instead. My solution was just to copy the entire source of ActiveDirectoryLdapAuthenticationProvider (along with a package private exception) and change bindPrincipal to username.

adamjhamer:
Hi,

I've been looking at the issues of sAMAccountName in search filters and the history of requests and changes to ActiveDirectoryLdapAuthenticationProvider, and I believe this issue can now be closed in favour of using the new rootDn and leaving the domain blank.

[...]

PR #4459, opened 2017-07-18
Pass username as second parameter for search filter.

taasjord:
Allows the username only (without domain) to be used in custom search filter like "sAMAccountName={1}",
in eg. situations where the userPrincipalName has a different suffix than domain.

Thanks to contributors in issue.

fixes gh-2448


rwinch:
Thanks for the PR @taasjord! This is now merged into master via 8d717c62afd5d98b0aba467035389d3011434b51

PR #4497, opened 2017-08-16
Make `AuthenticatedPrincipal` serializable

vpavic:
As a root of principal hierarchy, the `AuthenticatedPrincipal` should extend `Serializable`.


rwinch:
Thanks for the PR!

It should be up to implementations if they want to implement `Serializable` and not enforced by the interface. For example, you don't see `List` extending `Serializable` but you do see `ArrayList` implementing it.

I understand that Spring Security has some interfaces that do this, but I don't think we want to continue this going forward.

PR #4504, opened 2017-08-21
Add localization to error messages from ExceptionTranslationFilter

britter:
In by simple Spring Boot application with Basic Authentication, I can add localization and custom messages if somebody tries to access a resource with wrong credentials. This is implemented in `AbstractUserDetailsAuthenticationProvider`.

However when somebody forgets to add authentication information the `ExceptionTranslationFilter` will throw an `InsufficientAuthenticationException`. Since the message is hard coded, there is no easy way to customize or localize the error message.

This PR adds localization using message bundles and a `MessageSourceAccessor` the same way as it is implemented in `AbstractUserDetailsAuthenticationProvider`.


britter:
@jgrandja any feedback?

web/src/main/java/org/springframework/security/web/access/ExceptionTranslationFilter.java
```diff
@@ -83,6 +86,8 @@
 
 	private RequestCache requestCache = new HttpSessionRequestCache();
 
+	private MessageSourceAccessor messages = SpringSecurityMessageSource.getAccessor();
```
jgrandja: Add final modifier

web/src/main/resources/org/springframework/security/web/messages_en.properties
```diff
@@ -0,0 +1 @@
+ExceptionTranslationFilter.insufficientAuthentication=Full authentication is required to access this resource
```
jgrandja: Add this key to messages.properties (default bundle) as well

britter: Done!

britter:
I need some time to work on a test, since I have to understand how the current tests work and how to make the test reliable on different system locales. Thank you!

britter:
@jgrandja all done. Let's see what Travis thinks.

rwinch:
Thanks for the PR @britter! This is now merged into master via fffd781

PR #4515, opened 2017-08-31
Prefer `sub` claim as OIDC principal name

vpavic:
At present `DefaultOidcUser` uses `sub` claim as principal name by default, however if `name` claim is provided by UserInfo Endpoint it will be used instead.

This is IMO not ideal since `name` claim can neither be considered as unique nor non-reassignable identified of the user.

From the [OIDC Core spec](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) on the `name` claim:

> End-User's full name in displayable form including all name parts, possibly including titles and suffixes, ordered according to the End-User's locale and preferences.

### DIFF
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/oidc/core/user/DefaultOidcUser.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/oidc/core/user/DefaultOidcUser.java
@@ -13,21 +13,19 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package org.springframework.security.oauth2.oidc.core.user;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
 
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.oauth2.core.user.DefaultOAuth2User;
 import org.springframework.security.oauth2.oidc.core.IdToken;
 import org.springframework.security.oauth2.oidc.core.IdTokenClaim;
-import org.springframework.security.oauth2.oidc.core.StandardClaim;
 import org.springframework.security.oauth2.oidc.core.UserInfo;
-
-import java.util.Map;
-import java.util.Set;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
-import static org.springframework.security.oauth2.oidc.core.StandardClaim.NAME;
+import org.springframework.util.Assert;
 
 /**
  * The default implementation of an {@link OidcUser}.
@@ -35,10 +33,10 @@ import static org.springframework.security.oauth2.oidc.core.StandardClaim.NAME;
  * <p>
  * The claim used for accessing the &quot;name&quot; of the
  * user <code>Principal</code> via {@link #getClaims()}
- * is {@link StandardClaim#NAME} or if not available
- * will default to {@link IdTokenClaim#SUB}.
+ * is {@link IdTokenClaim#SUB}.
  *
  * @author Joe Grandja
+ * @author Vedran Pavic
  * @since 5.0
  * @see OidcUser
  * @see DefaultOAuth2User
@@ -50,20 +48,22 @@ public class DefaultOidcUser extends DefaultOAuth2User implements OidcUser {
 	private final UserInfo userInfo;
 
 	public DefaultOidcUser(Set<GrantedAuthority> authorities, IdToken idToken) {
-		this(authorities, idToken, null);
+		this(authorities, idToken, null, IdTokenClaim.SUB);
+	}
+
+	public DefaultOidcUser(Set<GrantedAuthority> authorities, IdToken idToken, String nameAttributeKey) {
+		this(authorities, idToken, null, nameAttributeKey);
 	}
 
 	public DefaultOidcUser(Set<GrantedAuthority> authorities, IdToken idToken, UserInfo userInfo) {
-		super(authorities, idToken.getClaims(), IdTokenClaim.SUB);
+		this(authorities, idToken, userInfo, IdTokenClaim.SUB);
+	}
+
+	public DefaultOidcUser(Set<GrantedAuthority> authorities, IdToken idToken, UserInfo userInfo,
+			String nameAttributeKey) {
+		super(authorities, resolveAttributes(idToken, userInfo), nameAttributeKey);
 		this.idToken = idToken;
 		this.userInfo = userInfo;
-		if (userInfo != null) {
-			this.setAttributes(
-				Stream.of(this.getAttributes(), userInfo.getClaims())
-					.flatMap(m -> m.entrySet().stream())
-					.collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (k1, k2) -> k1))
-			);
-		}
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


jgrandja:
@vpavic If you look at the javadoc for `AuthenticatedPrincipal.getName()` it states:

> Returns the name of the authenticated `Principal`

The `name` claim aligns with this. The `sub` claim is not a name rather it's a unique id assigned by the provider. From what I've seen so far from some of the provider's I've tested against, it usually is a `Number`.

`Authentication.getName()` is used quite a bit, usually to display the authenticated user's name in some UI. Having the `sub` claim displayed in a UI is not ideal.

Also, as far as uniqueness goes, `sub` is not sufficient in this case because in those rare cases, there may be 2 `sub`'s that are the same but have different `iss` claims. So an `OidcUser` is uniquely identified by the combination of `iss` and `sub`.

vpavic:
@jgrandja Thanks for the feedback!

My problem with the current situation is that the value of `name` claim depends on a lot of mutable user attributes as well as locale and can easily return a different value for a successive requests (as in ID Token requests).

Traditionally, `Authentication#getName` in Spring Security would return an unique identifier of the user in subject, which can then be either used to retrieve any locally store user data, or do audit logging (Boot's `AuditEvent` relies on that as well).

vpavic:
I agree that the original proposal of relying on `sub` claim isn't ideal, but I'm not sure the `email` claim is either. Obviously, that works just fine with Google as an ID provider but with other providers IMO it won't since often times email is a mutable attribute.

I'm thinking we should have the preferred principal claim declared via constructor like in `DefaultOAuth2User`, with `sub` claim remaining present as a fallback. WDYT @jgrandja?

Once we agree on the approach, I'll rework the PR accordingly.

jgrandja:
@vpavic After I proposed `email` claim I realized that this won't work because it won't always be available in `OidcUser` because it depends on the requested scopes and/or if the `UserInfo Endpoint` is eagerly fetched during the authentication flow or not. Regardless, `email` is not a good option.

I was just chatting with @rwinch and we came up with another option. How about `DefaultOidcUser.getName()` returns `sub`@`iss` and the constructor remains as-is with `sub` as default?

To provide more flexibility, as you suggested, we can also provide an additional constructor param for principal claim name like in `DefaultOAuth2User`.

vpavic:
> How about DefaultOidcUser.getName() returns sub@iss and the constructor remains as-is with sub as default?

I think that could be used as a fallback/default but IMO you have to provide users with option to configure the principal according to the specific OpenID Provider. For Google, that would mean setting `email` claim, for Github it's `login`, for some other provider it is something else so constructor parameter feels as a necessity.

jgrandja:
@vpavic The `sub` and  `iss` claims are required attributes for the _ID Token_ and therefore this uniquely and globally identifies the principal across all OpenID Providers. So I feel this should be the default. For OAuth2 Providers, like GitHub, we need to provide this option to allow configuring the principal name, which is what is provided today. It's not currently provided for `OidcUser` but it wouldn't hurt to provide this extra configuration.

I feel the `getName()` by default should return a combination of `iss` and `sub`, following this format:

https://{sub}@{iss}, for example, https://499926453@google.accounts.com

vpavic:
OK, my previous comment was a bit off since GitHub obviously isn't a OpennID Provider, but generally speaking the described situation applies when considering different providers.

Back to your proposal, how actually would you implement the `sub`@`iss` as a default? The way I see it that basically describes the default principal name as a `Function`. If you implement the `sub`@`iss` logic in the `DefaultOidcUser#getName` then the `nameAttributeKey` supplied to `DefaultOAuth2User` in `DefaultOidcUser` becomes useless and you'd need to introduce another `nameAttributeKey` as a member of `DefaultOidcUser`.

Other than that, while I see what you're aiming at with `sub`@`iss` proposal, IMO that also has some limitations in a real world app. Obviously such combination indeed is unique across different OpenID Providers but if your app does indeed provide an option to log in using multiple providers I'd expect it to have some locally stored user data which will in turn be linked to identities at different providers. So that also makes identifying user a bit problematic if they use different Providers across logins - think audit logging again or resolving user preferences (which is local user data basically) as an examples. This somewhat implies that replacing the `nameAttributeKey` with a `Function` might be a good idea.

Thoughts?

jgrandja:
@vpavic Great discussions here :)

Again, you have a good point as far as `getName()` being implemented as `sub@iss`...this would invalidate the `nameAttributeKey` attribute. This wouldn't be good from a design standpoint.

The best option and the most flexible is the one you suggested in this [comment](https://github.com/spring-projects/spring-security/pull/4515#issuecomment-327526220).

So let's remove the `DefaultOidcUser.getName()` override and leave the default `nameAttributeKey` as `sub`. As well, let's provide another constructor that will allow customization of the `nameAttributeKey` as follows:

`public DefaultOidcUser(Set<GrantedAuthority> authorities, IdToken idToken, String nameAttributeKey)`

Sound good?

vpavic:
> Great discussions here :)

Agreed! :+1:

> Sound good?

Yes, with addition of one more constructor:

`DefaultOidcUser#DefaultOidcUser(Set<GrantedAuthority>, IdToken, UserInfo, String)`

I've just pushed the update with those changes, and I thought this PR might also be a good opportunity to provide some test for `DefaultOAuth2User` and `DefaultOidcUser`. Doing so uncovered a couple of problems:

- there's no validation of `IdToken` passed to `DefaultOidcUser` which can result in NPE
- using claim like `StandardClaim.EMAIL` as `nameAttributeKey` isn't possible since the current logic applies `UserInfo` claims after calling super constructor, i.e. such scenario will result in `IllegalArgumentException` due to `attributes` not containing provided `nameAttributeKey`
- `idToken` and `userInfo` members aren't accessed after being used to populate attributes during initialization

Let me know what you think.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/oidc/core/user/DefaultOidcUser.java
```diff
@@ -13,67 +13,68 @@
...
  * @author Joe Grandja
+ * @author Vedran Pavic
  * @since 5.0
  * @see OidcUser
  * @see DefaultOAuth2User
  * @see IdToken
  * @see UserInfo
  */
 public class DefaultOidcUser extends DefaultOAuth2User implements OidcUser {
-	private final IdToken idToken;
```
jgrandja: I don't feel we should get rid of these fields. Even though the claims from both `IdToken` and `UserInfo` are copied to `attributes` we may want to provide getters for these 2 constructs.

oauth2/oauth2-core/src/test/java/org/springframework/security/oauth2/core/user/DefaultOAuth2UserTests.java
```diff
@@ -0,0 +1,110 @@
...
+	private static final String TEST_USERNAME = "test";
+
+	private static final Map<String, Object> TEST_ATTRIBUTES = Collections.singletonMap(
+			TEST_ATTRIBUTE_NAME_KEY, TEST_USERNAME);
+
+	@Rule
+	public ExpectedException thrown = ExpectedException.none();
+
+	@Test
+	public void constructValid() {
```
jgrandja: Our naming conventions for test methods follow this format - > methodNameWhen&lt;condition&gt;Then&lt;Expectation&gt;

So for this test the method name would be something like -> constructorWhenParametersAreValidThenIsCreated

jgrandja:
Thanks for adding the tests and catching some issues!

> using claim like StandardClaim.EMAIL as nameAttributeKey isn't possible since the current logic applies UserInfo claims after calling super constructor, i.e. such scenario will result in IllegalArgumentException due to attributes not containing provided nameAttributeKey

How about we do this:

```
super(authorities, resolveAttributes(idToken, userInfo), IdTokenClaim.SUB);
this.setNameAttributeKey(nameAttributeKey)
```

`setNameAttributeKey` would be in `DefaultOAuth2User` and would provide the assertion

```
			throw new IllegalArgumentException("Invalid nameAttributeKey: " + nameAttributeKey);
		}
```

vpavic:
> Looks like you got around this with `resolveAttributes`.

Correct.

I've updated the PR according to your feedback.

jgrandja:
Thanks for the PR @vpavic! I added the getters for `IdToken` and `UserInfo` and applied some minor polish. This is now merged to master.

Issue #4482, opened 2017-08-02
Spring Security Documentation for Section 21.3 Concurrency Control


onlysolace:
The documentation provides this sample code in 21.3

```
<beans:bean id="concurrencyFilter"
class="org.springframework.security.web.session.ConcurrentSessionFilter">
<beans:property name="sessionRegistry" ref="sessionRegistry" />
<beans:property name="expiredUrl" value="/session-expired.htm" />
</beans:bean>
```

But when I run this with Jetty, I get an exception telling me that the concurrencyFilter bean cannot be instantiated because there was no default constructor found. I believe Spring Security expects us to inject the properties into the bean like so:

```
<beans:bean id="concurrencyFilter" class="org.springframework.security.web.session.ConcurrentSessionFilter">
       <beans:constructor-arg ref="sessionRegistry" /> 
       <beans:constructor-arg value="/login" /> 
    </beans:bean>
```

This is working for me.

PR #4622, opened 2017-10-12
gh-4482 - Updating docs for `ConcurrentSessionFilter` configuration e…

tristanchanson:
Updating docs to not use deprecated example and to fix beans tags to set constructor-args instead of properties

Fixes gh-4482


rwinch:
Thanks for the PR @tristanchanson! This is now merged into master

Issue #4574, opened 2017-09-26
Add Validator for ClientRegistration


jgrandja:
We should provide a `Validator` for a `ClientRegistration`.

The `Validator` can be applied before it's composed within a `ClientRegistrationRepository` to ensure data integrity. For example, `authorizationUri`, `tokenUri` and `userInfoUri` should be valid (parseable) `URI`'s.

jgrandja:
At this time, I don't feel there is a need for a `Validator` of `ClientRegistration` given the validation logic is encapsulated in `ClientRegistration.Builder` which is sufficient. Closing this issue.

PR #4631, opened 2017-10-14
Add Validator for ClientRegistration #4574

shazin:
Add Validator for ClientRegistration #4574


jgrandja:
Thanks for the PR @shazin.

I was thinking of leveraging `org.springframework.validation.Validator` and providing an implementation for `ClientRegistration`. I don't feel like we need to introduce the new interface `ClientRegistrationValidator`.

Also, you may have noticed that the validation checks are in:

`ClientRegistration.Builder.validateAuthorizationCodeGrantType()`
`ClientRegistration.Builder.validateImplicitGrantType()`

I logged this ticket a couple weeks back thinking it might be a good idea to externalize the validation checks, however, I'm not convinced that it is necessary.

Either way, we are going to hold off on this issue until after we release 5.0 and than consider if it's beneficial to externalize this.

Thanks for your efforts!

Issue #4666, opened 2017-10-20
Add DelegatingPasswordEncoder


rwinch:
### Summary

Create a DelegatingPasswordEncoder which supports delegating to multiple different PasswordEncoder implementations. 

```java
String idForEncode = "bcrypt";
Map<String,PasswordEncoder> encoders = new HashMap<>();
encoders.put(idForEncode, new BCryptPasswordEncoder());
encoders.put("noop", NoOpPasswordEncoder.getInstance());
encoders.put("pbkdf2", new Pbkdf2PasswordEncoder());
encoders.put("scrypt", new SCryptPasswordEncoder());
encoders.put("sha256", new StandardPasswordEncoder());

PasswordEncoder passwordEncoder = new DelegatingPasswordEncoder(idForEncode, encoders);
```

The following values would be stored:

```
{bcrypt}$2a$10$dXJ3SW6G7P50lGmMkkmwe.20cQQubK3.HZWzG3YB1tlRy.fqvM/BG
{noop}password
{pbkdf2}5d923b44a6d129f3ddf3e3c8d29412723dcbde72445e8ef6bf3b508fbf17fa4ed4d6b99ca763d8dc
{scrypt}$e0801$8bWJaSu2IKSn9Z9kM+TPXfOc/9bdYSrN1oD9qfVThWEwdRTnO7re7Ei+fUZRJ68k9lTyuTeUp4of4g24hHnazw==$OAOec05+bXxvuu/1qZ6NUR+xQYvYv7BeL1QxwRpY5Pc=
{sha256}97cde38028ad898ebc02e690819fa220e88c62e0699403e94fff291cfffaf8410849f27605abcbc0
```
 
We would then use the following logic for matching:

* The first password would have a `PasswordEncoder` id of `bcrypt` and
encodedPassword of `$2a$10$dXJ3SW6G7P50lGmMkkmwe.20cQQubK3.HZWzG3YB1tlRy.fqvM/BG`.
When matching it would delegate to
`BCryptPasswordEncoder`
* The second password would have a `PasswordEncoder` id of `noop` and
encodedPassword of `password`. When matching it would delegate to `NoOpPasswordEncoder`
* The third password would have a `PasswordEncoder` id of `pbkdf2` and
encodedPassword of
`5d923b44a6d129f3ddf3e3c8d29412723dcbde72445e8ef6bf3b508fbf17fa4ed4d6b99ca763d8dc`.
[...]

PR #4669, opened 2017-10-22
Fix Javadoc of PasswordEncoderFactories#createDelegatingPasswordEncoder

kazuki43zoo:
See gh-4666

Probably, Javadoc is wrong. In the current implementation, the `BCryptPasswordEncoder` is used for encoding a password. 
Is wrong implementation?


rwinch:
Thanks for the PR @kazuki43zoo! This is now merged into master

PR #4670, opened 2017-10-22
Add new factory method for creating DelegatingPasswordEncoder

kazuki43zoo:
See gh-4666

I will propose to add new factory method that can specify the id of an encoder for encoding password.
What do you think?


rwinch:
@kazuki43zoo This solution poses a problem because the user does not have a reference to existing id's so they would have to read the code to enable it. I'm not even sure if this is a good approach overall. If users want more flexibility, they can create an instances using `DelegatingPasswordEncoder` directly.

We could discuss some alternative builder possibilities, but at the moment I'm not convinced we should add them since using  `DelegatingPasswordEncoder` directly is an option.

kazuki43zoo:
@rwinch Thanks for your comment. I agree with your opinion.
Thanks again.

PR #4759, opened 2017-10-30
Replace improper JsonNode.toString() in Jackson support

### LINKED ISSUE #4252: Spring Session integration only supports User as the principal

fpavageau:
This is #4253 reported on `4.2.x` to fix #4252 on that branch as well, for people that won't be able to upgrade to `5.0` right away.

### DIFF
--- a/core/src/main/java/org/springframework/security/jackson2/UnmodifiableSetDeserializer.java
+++ b/core/src/main/java/org/springframework/security/jackson2/UnmodifiableSetDeserializer.java
@@ -50,10 +50,10 @@ class UnmodifiableSetDeserializer extends JsonDeserializer<Set> {
 				Iterator<JsonNode> nodeIterator = arrayNode.iterator();
 				while (nodeIterator.hasNext()) {
 					JsonNode elementNode = nodeIterator.next();
-					resultSet.add(mapper.readValue(elementNode.toString(), Object.class));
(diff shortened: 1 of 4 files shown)


rwinch:
Thanks for the PR @fpavageau! This is now in 4.2.x

PR #4761, opened 2017-10-31
Fix UsernamePasswordAuthenticationTokenMixin to handle null credentials/details

### LINKED ISSUE #4698: SecurityJackson2Modules deserializes "credentials" and "details" wrong
* When a null for "credentials" is serialized out as JSON, it gets deserialized back as a String value of "null".
* When a null for "details" is serialized out as JSON, it gets deserialized back as a JSON NullNode.

The point is that both of these fields should round trip back as nulls to allow for a perfect match.

For an example, see test cases in https://github.com/spring-projects/spring-session-data-mongodb/pull/18

gregturn:
Resolves #4698


rwinch:
Thanks for the PR @gregturn! This is now merged into master

PR #4819, opened 2017-11-13
AclClassIdUtils should be public

### LINKED ISSUE #4814: AclClassIdUtils should be public
### Summary

In #4424 I added `AclClassIdUtils` to handle some common logic for dealing with acl classes with types other than Long. I thought this class would be internal, then decided it was better to have it explicitly defined.

The class has package level visibility, but should be public.

### Actual Behavior

Unable to define a bean of type AclClassIdUtils using Java Config as the class is not public.

### Expected Behavior

Should be able to define a bean of type AclClassIdUtils using Java Config.

### Configuration

`@Bean
public AclClassIdUtils aclClassIdUtils() {

    AclClassIdUtils classIdUtils = new AclClassIdUtils();
    classIdUtils.setConversionService(new DefaultConversionService());
    
    return classIdUtils
}
`

pwheel:
AclClassIdUtils was added in #4424 but should be made public so can be instantiated via Java Config.

Fixes #4814

### DIFF
--- a/acl/src/main/java/org/springframework/security/acls/jdbc/AclClassIdUtils.java
+++ b/acl/src/main/java/org/springframework/security/acls/jdbc/AclClassIdUtils.java
@@ -29,13 +29,17 @@ import org.springframework.security.acls.model.ObjectIdentity;
  * the correct Java type as specified by <code>acl_class.class_id_type</code>.
  * @author paulwheeler
  */
-public class AclClassIdUtils {
+class AclClassIdUtils {
 	private static final String DEFAULT_CLASS_ID_TYPE_COLUMN_NAME = "class_id_type";
 	private static final Log log = LogFactory.getLog(AclClassIdUtils.class);
 
 	private ConversionService conversionService;
 
 	public AclClassIdUtils() {
+	}
+
+	public AclClassIdUtils(ConversionService conversionService) {
+		this.conversionService = conversionService;
 	}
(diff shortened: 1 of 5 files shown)


pwheel:
Hi @rwinch can you take a look at this please? Minor fix to enable Java Config to be used.

Thanks,
Paul

rwinch:
Thanks for the PR and the ping.

I think perhaps a better way of handling this is to allow injecting the `ConversionService` into classes using `AclClassIdUtils` and then internally the setter method creates a `AclClassIdUtils` instance with the `ConversionService`.

pwheel:
Yep, that sounds good @rwinch, thanks. I was never quite happy with this bit, I like your suggestion.

I've made the changes.
Cheers.

pwheel:
@rwinch ping :-)

pwheel:
Hi @rwinch are you able to give some feedback on this? I've made some changes in keeping with your suggestions I hope.

Thanks,
Paul

rwinch:
Thanks for the PR & sorry for the delays @pwheel! This is now merged into master

pwheel:
No worries @rwinch I should have chased you harder! 😄 That's awesome, thanks a lot

rwinch:
It was all me. I need to be better about being responsive (I'm focusing on that going forward) Thanks again for the PR!

PR #4823, opened 2017-11-15
Remove duplicate eruby option in asciidoctor task

izeye:
This PR removes duplicate `eruby` option in `asciidoctor` task.


rwinch:

Issue #4478, opened 2017-07-28
Remove HttpClientConfig in favor of ResourceRetriever Bean


rwinch:
### Summary

Rather than having `HttpClientConfig` replace it with allowing the user to provide a `ResourceRetriever` Bean which would override the default. Users wishing to change the timeout could then provide a `DefaultResourceRetriever` Bean with a different timeout.

This removes configuration specific code and provides more flexibility by allowing the `ResourceRetriever` that is used to be replaced with a different implementation.

Related #4477

jgrandja:
`ResourceRetriever` is used by `RemoteJWKSet` in `NimbusJwtDecoderJwkSupport` when fetching the JWK Set Uri. However, `ResourceRetriever` is **not used** for the `TokenRequest` in `NimbusAuthorizationCodeTokenExchanger` nor is it used when fetching the `UserInfoRequest` in `NimbusOAuth2UserService`.

This is the reason I decided to create `HttpClientConfig` so it can be used consistently across these 3 use cases and future use cases for outbound requests. Does this make sense or do you have another suggestion?

rwinch:
I don't like having configuration directly in our service layer. Configuration belongs in the configuration. Perhaps we need to introduce an abstraction for retrieving resources / HTTPRequest that can unify the settings. That implementation can delegate to the original methods with the proper settings

jgrandja:
Ok let me see what I can come up with.

jgrandja:
@rwinch I removed `HttpClientConfig` and set the default read and connection timeouts to 30 secs for all 3 client calls in `NimbusAuthorizationCodeTokenExchanger`, `NimbusOAuth2UserService` and `NimbusJwtDecoderJwkSupport`.

Let's revisit the client abstraction for retrieving resources after we go GA.

I'll leave this issue open.

rwinch:
Closing as this is now invalid

PR #4830, opened 2017-11-15
Allow configuration of timeout on NimbusAuthorizationCodeTokenResponseClient

kazuki43zoo:
In a current implementation, following values are hard coding.

* connectTimeout : 30000 (30 sec)
* readTimeout : 30000 (30 sec)

I think better that those values can be customized.

WDYT?


jgrandja:
@kazuki43zoo Thanks for the PR.
We are planning on introducing a Client API abstraction that will allow for configuring these values and others. We didn't have the time to design this in time for 5.0 release so it will come after at some point.
At this point, I'm not sure if we'll expose this configuration via `OAuth2AccessTokenResponseClient`. Likely not though. It will most likely be a Client API that will be composed within an implementation of `OAuth2AccessTokenResponseClient`.
Given that, I'm going to close this PR and revisit this at a later point. The default values will be more than sufficient for all use cases.
Thanks again.

jgrandja:
Related #4478

kazuki43zoo:
thanks for your comment! I understood it.

PR #4725, opened 2017-10-28
Ensure Copyright header reads 2002-2017

unlimitedsola:
fixes gh-4655

Signed-off-by: Sola <dev@sola.love>

Issue #4655, opened 2017-10-18
Ensure Copyright header reads 2002-2017


jgrandja:
Some of the Copyright header in the OAuth 2.0 modules read 2012-2017. It should be 2002-2017.

PR #4847, opened 2017-11-17
Fix copyright start years

izeye:
This PR fixes wrong copyright start years.

See gh-4655
See gh-4725


rwinch:

PR #4848, opened 2017-11-17
Create OAuth2AuthorizationResponse lazily

izeye:
This PR creates `OAuth2AuthorizationResponse` as lazily as possible to prevent the creation when `authorizationRequest` is `null`.

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2LoginAuthenticationFilter.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2LoginAuthenticationFilter.java
@@ -108,7 +108,6 @@ public class OAuth2LoginAuthenticationFilter extends AbstractAuthenticationProce
 			OAuth2Error oauth2Error = new OAuth2Error(OAuth2ErrorCodes.INVALID_REQUEST);
 			throw new OAuth2AuthenticationException(oauth2Error, oauth2Error.toString());
 		}
-		OAuth2AuthorizationResponse authorizationResponse = this.convert(request);
 
 		OAuth2AuthorizationRequest authorizationRequest = this.authorizationRequestRepository.loadAuthorizationRequest(request);
 		if (authorizationRequest == null) {
@@ -119,6 +118,8 @@ public class OAuth2LoginAuthenticationFilter extends AbstractAuthenticationProce
 
 		String registrationId = (String) authorizationRequest.getAdditionalParameters().get(OAuth2ParameterNames.REGISTRATION_ID);
 		ClientRegistration clientRegistration = this.clientRegistrationRepository.findByRegistrationId(registrationId);
+
+		OAuth2AuthorizationResponse authorizationResponse = this.convert(request);
 
 		OAuth2LoginAuthenticationToken authenticationRequest = new OAuth2LoginAuthenticationToken(
 				clientRegistration, new OAuth2AuthorizationExchange(authorizationRequest, authorizationResponse));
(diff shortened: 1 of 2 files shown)


rwinch:
@izeye Would you consider adding a test for this?

izeye:
@rwinch Sure. I added one in 60259873090cf489550b9bb7d959b1b79981277a.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2LoginAuthenticationFilter.java
```diff
@@ -147,7 +148,7 @@ public final void setAuthorizationRequestRepository(AuthorizationRequestReposito
 		this.authorizationRequestRepository = authorizationRequestRepository;
 	}
 
-	private OAuth2AuthorizationResponse convert(HttpServletRequest request) {
```
jgrandja: This needs to remain `private`.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/OAuth2LoginAuthenticationFilterTests.java
```diff
@@ -263,6 +267,23 @@ public void doFilterWhenCustomFilterProcessesUrlThenFilterProcesses() throws Exc
 		verify(this.filter).attemptAuthentication(any(HttpServletRequest.class), any(HttpServletResponse.class));
 	}
 
+	@Test
+	public void attemptAuthenticationWhenOAuth2AuthorizationRequestIsNullDoesNotCreateOAuth2AuthorizationResponse() throws IOException, ServletException {
```
jgrandja: This test is pretty much the same as `doFilterWhenAuthorizationResponseAuthorizationRequestNotFoundThenAuthorizationRequestNotFoundError`.
It doesn't actually test the change you added.

izeye: @jgrandja This is testing the invocation of `convert()` which creates an `OAuth2AuthorizationResponse`. If you revert the change, this will fail. So I think this is effectively testing the change.

I couldn't see any better way to test this. Any suggestion?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2LoginAuthenticationFilter.java
izeye: @jgrandja I couldn't verify `convert()` invocation easily without this change and I thought it's a good compromise to be package-private to make it easy to test. If there's any way not to do this, please let me know.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/OAuth2LoginAuthenticationFilterTests.java
jgrandja: @izeye You can mock private methods using [Powermock](https://github.com/powermock/powermock/wiki/Mockito#a-full-example-of-partial-mocking-of-a-private-method) and verify the zero interaction on the invocation.

Take a look at [OidcAuthorizationCodeAuthenticationProviderTests](https://github.com/spring-projects/spring-security/blob/master/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/oidc/authentication/OidcAuthorizationCodeAuthenticationProviderTests.java#L412) which is mocking a private method.

izeye:
@jgrandja Thanks for the pointer!

As you suggested, I changed the test to use `PowerMockito` and reverted the access modifier change in 53757b2.

jgrandja:
Thanks for the PR @izeye. I applied some minor polish to the test and merged to master.

PR #4850, opened 2017-11-17
Editing pass for OAuth 2.0 Login ref doc

Buzzardo:
At Joe's request, I edited the changes he made to account for OAuth2.


jgrandja:
Thanks for the PR @Buzzardo ! This is now merged to master after I applied some more polish.

PR #4851, opened 2017-11-18
Remove explicit super() calls

izeye:
This PR removes explicit `super()` calls.


rwinch:

PR #4853, opened 2017-11-19
Use diamond type

izeye:
This PR replaces type arguments with diamond type.


rwinch:
Thanks for the PR @izeye! This is now merged into master via 57353d18e

PR #4860, opened 2017-11-24
Remove initialization of ServerHttpSecurity.headers

izeye:
There is `null` handing for `ServerHttpSecurity.headers` but it can't be `null` unless something similar `CsrfSpec.disable()` should be added.

This PR simply drops its initialization.


rwinch:
This is intentional as we want the headers to be on even if no one invokes the headers method (they would need to explicitly disable the headers).

izeye:
@rwinch You mean `HeaderSpec.disable()` needs to be added?

PR #4879, opened 2017-11-28
Format security filters enums for readability

fhanik:
One filter per line makes it a bit easier to read when comparing order.


rwinch:
Thanks for the PR @fhanik! This is now in master

PR #4893, opened 2017-12-02
Change default scope to 'read:user' for GitHub

kazuki43zoo:
I think the `read:user`(read-only scope) is a more proper scope as OAuth 2.0 Login.
WDYT?


jgrandja:
Thanks for the PR @kazuki43zoo. This is now in master.

PR #4895, opened 2017-12-03
Remove address and phone from default scope for Google

kazuki43zoo:
I think the `address` and `phone` are ignored(=not supported) on the userinfo endpoint of Google.
About supported scopes for OpenID Connect refer to [here](https://developers.google.com/identity/protocols/googlescopes#openid_connect).

WDYT?


jgrandja:
Thanks for the PR @kazuki43zoo. This is now in master.

PR #4899, opened 2017-12-05
Remove @Nullable annotations in UserDetailsMapFactoryBean

izeye:
It looks impossible to be `null`, so this PR removes the `@Nullable` annotations.


rwinch:
Thanks for the PR @izeye! This is now merged into master via 316fd0572

Issue #4878, opened 2017-11-28
spring-security-bom provides dependency management for itself


wilkinsona:
### Summary

In 5.0.0.RELEASE, `spring-security-bom` has started providing dependency management for itself.

### Actual Behavior

`spring-security-bom` provides dependency management for an `org.springframework.security:spring-security-bom` jar artifact that does not exist.

### Expected Behavior

`spring-security-bom` does not provide dependency management for a `org.springframework.security:spring-security-bom` jar artifact.

### Configuration

N/A

### Version

5.0.0.RELEASE

### Sample

N/A

wilkinsona:
I can work around it, but this is breaking the IO Platform's build as it verifies that every managed dependency can be downloaded. `spring-security-bom-5.0.0.RELEASE.jar` does not exist so this verification fails.

PR #4902, opened 2017-12-06
Remove spring-security-bom in dependencyManagement

eddumelendez:
spring-security-bom it provides dependency management for itself. This
commit avoids adding it to the dependency management section.

Fixes gh-4878


rwinch:
Thanks for the PR @eddumelendez! This is now merged into master via 9b84204e10f17acbf5ba97c4aa4897931d6677dc

Issue #4915, opened 2017-12-18
BCrypt encode calls increase startup times


philwebb:
### Summary
Whilst investigating Spring Boot [startup time regressions](https://github.com/spring-projects/spring-boot/issues/11226#issuecomment-352589988) it became apparent that `BCrypt.encode` is a hot method.

### Actual Behavior
`DaoAuthenticationProvider` currently generates an encoded password in [`doAfterPropertiesSet`](https://github.com/spring-projects/spring-security/blob/9f6af4f3b8e79a8a45369d99862cd3d96f4083ce/core/src/main/java/org/springframework/security/authentication/dao/DaoAuthenticationProvider.java#L97). This encoded password is decoded when a `UsernameNotFoundException` is throw in order to prevent timing attacks (a user that isn't found takes about the same time to process as one that is).

### Expected Behavior
It would be nice to remove the `encode` call from the application startup critical-path so that application startup times are not increased. One possible option might be to make `userNotFoundEncodedPassword` late binding and to calculate it on the first call to `retrieveUser`.

I believe (although I'm by no means an expert here) that this should allow faster startup without being susceptible to timing attacks. The first call to `retrieveUser` would be a little slower, but this will be constant regardless of if a user is ultimately found or not. Subsequent call will have the same behavior as the current implementation.

### Configuration
Spring Boot 2.0.M7 running `spring-boot-sample-actuator-ui`.

### Version
5.0.0.RELEASE

### Sample
https://github.com/spring-projects/spring-boot/tree/master/spring-boot-samples/spring-boot-sample-actuator-ui

philwebb:
With PR #4927 applied Spring Boot startup improves from 

```
Started SampleActuatorUiApplication in 3.373 seconds (JVM running for 3.781)
```

to

```
Started SampleActuatorUiApplication in 3.142 seconds (JVM running for 3.545)
```

philwebb:
Sweeeeeet!! Thanks.

PR #4927, opened 2017-12-22
Lazily initialize userNotFoundEncodedPassword

philwebb:
Update `DaoAuthenticationProvider` so that `userNotFoundEncodedPassword`
is lazily initialized on the first call to `retrieveUser`, rather than
in `doAfterPropertiesSet`.

Since some `PasswordEncoder` implementations can be slow, this change
can help to improve application startup times and the expense of some
delay with the first login.

Note that `userNotFoundEncodedPassword` creation occurs on the first
user retrieval, regardless of whether the user is ultimately found. This
ensures consistent processing times, regardless of the outcome.

First Call:
	Found      = encode(userNotFound) + decode(supplied)
	Not-Found  = encode(userNotFound) + decode(userNotFound)

Subsequent Call:
	Found      = decode(supplied)
	Not-Found  = decode(userNotFound)

Fixes gh-4915


rwinch:
Thanks for the PR @philwebb! I merged this into master with a bit of polish 
6ba225b

PR #4934, opened 2017-12-29
Remove unused imports and fixes some compile time errors for some tests

adolfoweloy:
- Remove invalid imports referencing `org.springframework.security.config.annotation.authentication.AuthenticationManagerBuilder`.

- Change the way to use `contains` method from `Asssertions` class from Assertj. When checking
for a concrete class against a generic collection, Eclipse and IntelliJ were presenting compiling time errors. There is an issue regarding the usage of generic types on `contains` method from `Assertions` class as follows:

https://github.com/joel-costigliola/assertj-core/issues/772

- Change `DelegatingOAuth2UserServiceTests` to start using concrete types to create the `ArrayList` of services that are needed to create an instance of `DelegatingOAuth2UserService`.

### DIFF
--- a/config/src/test/groovy/org/springframework/security/config/annotation/provisioning/UserDetailsManagerConfigurerTests.groovy
+++ b/config/src/test/groovy/org/springframework/security/config/annotation/provisioning/UserDetailsManagerConfigurerTests.groovy
@@ -15,6 +15,8 @@
  */
 package org.springframework.security.config.annotation.provisioning
 
+import org.springframework.security.config.annotation.authentication.builders.AuthenticationManagerBuilder
+import org.springframework.security.config.annotation.authentication.configurers.provisioning.InMemoryUserDetailsManagerConfigurer
 import org.springframework.security.config.annotation.authentication.configurers.provisioning.UserDetailsManagerConfigurer;
 import org.springframework.security.core.authority.SimpleGrantedAuthority
 import org.springframework.security.core.userdetails.UserDetails
@@ -30,10 +32,8 @@ import spock.lang.Specification
 class UserDetailsManagerConfigurerTests extends Specification {
 
 	def "all attributes supported"() {
-		setup:
-			InMemoryUserDetailsManager userDetailsManager = new InMemoryUserDetailsManager([])
 		when:
-			UserDetails userDetails = new UserDetailsManagerConfigurer<InMemoryUserDetailsManager,UserDetailsManagerConfigurer<InMemoryUserDetailsManager>>(userDetailsManager)
+			UserDetails userDetails = new InMemoryUserDetailsManagerConfigurer<AuthenticationManagerBuilder>()
 				.withUser("user")
 					.password("password")
 					.roles("USER")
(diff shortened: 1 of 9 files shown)


config/src/test/groovy/org/springframework/security/config/annotation/provisioning/UserDetailsManagerConfigurerTests.groovy
```diff
@@ -30,10 +32,8 @@ import spock.lang.Specification
 class UserDetailsManagerConfigurerTests extends Specification {
 
 	def "all attributes supported"() {
-		setup:
-			InMemoryUserDetailsManager userDetailsManager = new InMemoryUserDetailsManager([])
 		when:
-			UserDetails userDetails = new UserDetailsManagerConfigurer<InMemoryUserDetailsManager,UserDetailsManagerConfigurer<InMemoryUserDetailsManager>>(userDetailsManager)
```
rwinch: These tests are suppose to be validating `UserDetailsManagerConfigurer` not `InMemoryUserDetailsManagerConfigurer`. Is there a reason you switched this?

adolfoweloy: In fact I did a mistake here. But rather than parameterizing `UserDetailsManagerConfigurer` with `InMemoryUserDetailsManager` I think it should be better to use `AuthenticationManagerBuilder` as follows:

```java
UserDetails userDetails = new UserDetailsManagerConfigurer<AuthenticationManagerBuilder,
    UserDetailsManagerConfigurer<InMemoryUserDetailsManager>>(userDetailsManager)
```

That's because as the first generic parameter type for `UserDetailsManagerConfigurer` which is `B` extends `ProviderManagerBuilder`,  I realized that it should make sense to use `AuthenticationManagerBuilder`.

If I keep using `InMemoryUserDetailsManager` for the first generic parameter type, I got the following error on STS:
```
Groovy:The type InMemoryUserDetailsManager is not a valid substitute 
for the bounded parameter 
<B extends org.springframework.security.config.annotation.authentication.ProviderManagerBuilder<B>>
```

I made the changes to use `AuthenticationManagerBuilder` as proposed but did not pushed yet just to preserve our conversation here. If you think it makes sense to be applied I can submit the changes.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/userinfo/DelegatingOAuth2UserServiceTests.java
```diff
@@ -44,11 +44,10 @@ public void constructorWhenUserServicesIsEmptyThenThrowIllegalArgumentException(
 	}
 
 	@Test(expected = IllegalArgumentException.class)
-	@SuppressWarnings("unchecked")
 	public void loadUserWhenUserRequestIsNullThenThrowIllegalArgumentException() {
 		DelegatingOAuth2UserService<OAuth2UserRequest, OAuth2User> delegatingUserService =
 			new DelegatingOAuth2UserService<>(
-				Arrays.asList(mock(OAuth2UserService.class), mock(OAuth2UserService.class)));
+				Arrays.asList(mock(DefaultOAuth2UserService.class), mock(DefaultOAuth2UserService.class)));
```
jgrandja: The purpose of this test is to ensure that the `OAuth2UserRequest` passed into `loadUser` is **not** null else throw `IllegalArgumentException`. Changing the `List` of `OAuth2UserService` mocks to `DefaultOAuth2UserService` doesn't really apply to what is being tested here. Please revert this. Thank you.

rwinch:
Closing in favor of #4949

Issue #4936, opened 2017-12-30
DelegatingPasswordEncoder does not work with Maps throwing a NPE


michael-simons:
### Summary

Using the `DelegatingPasswordEncoder` with a Map that throws a `NullPointerException` when using a null key does not work. The contract of Javas Map says

> Throws: NullPointerException - if the specified key is null and this map does not permit null keys (optional)

https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#get-java.lang.Object-

This is especially true when using Java 9s immutable collections (java.util.ImmutableCollections)

### Actual Behavior

I am using a Java 9 map

```
final Map<String, PasswordEncoder> encoders = Map.of(
            idForEncode, defaultEncoder,
            "bcrypt", new BCryptPasswordEncoder(),
            "scrypt", new SCryptPasswordEncoder()
        );

final DelegatingPasswordEncoder passwordEncoder = new DelegatingPasswordEncoder(idForEncode, encoders);

passwordEncoder.matches(rawPassword, encodedPasswordNoPrefix);
```

and get:

```
java.lang.NullPointerException: null
	at java.base/java.util.ImmutableCollections$MapN.probe(ImmutableCollections.java:779) ~[na:na]
	at java.base/java.util.ImmutableCollections$MapN.get(ImmutableCollections.java:721) ~[na:na]
	at org.springframework.security.crypto.password.DelegatingPasswordEncoder.matches(DelegatingPasswordEncoder.java:193) ~[spring-security-core-5.0.0.RELEASE.jar:5.0.0.RELEASE]
```

Regardless wether `setDefaultPasswordEncoderForMatches` is used or not.

### Expected Behavior

I want to see the `IllegalArgumentException` like this:

```
Exception in thread "main" java.lang.IllegalArgumentException: There is no PasswordEncoder mapped for the id "null"
	at org.springframework.security.crypto.password.DelegatingPasswordEncoder$UnmappedIdPasswordEncoder.matches(DelegatingPasswordEncoder.java:236)
```

as long as I don't set `setDefaultPasswordEncoderForMatches`. 

### Version

Spring Security 5.0, *Java 8* with Maps that throw a `NullPointerException` for a null key or using *Java 9* map-builder.

### Sample

See my PR.

PR #4937, opened 2017-12-30
Catch possible NullPointerException

michael-simons:
Some maps may throw a NullPointerException when get is called with null. This commit catches the exceptions and just leaves the delegate null.

Fixes gh-4936


michael-simons:
Checking the Id for null would be another solution, but some maps (i.e. `HashMap`) allow null keys.

rwinch:
Thanks for the report and the PR! This was merged via 2b66793535eff78fba14fee0a75d0091c5636f56 I then applied some polish via 22737dce7e5021e1a6682bc9313fbae3a14b4f77 that copies the provided Map implementation into `HashMap`. This cleans up the code (no need for `try/catch`) and it ensures that references to the provided `Map` cannot change the state of `DelegatingPasswordEncoder`.

michael-simons:
Even better (clear visible in the test :) ).

Issue #4939, opened 2018-01-02
Migrate Groovy to Java


rwinch:
### Summary

The tests that are written in Groovy are very convenient and easy to read, but place an unnecessary burden on contributors unfamiliar with Groovy. We should migrate the Groovy tests to Java.

For the samples using Geb, take a look at https://github.com/spring-projects/spring-security/tree/b9152701a65df5b3fa78fe4bd2a946018f9ca352/samples/javaconfig/webflux-form/src/integration-test/java/sample

The following commits in Spring Session might be useful as well

https://github.com/spring-projects/spring-session/commit/1a318b89d907d01d80c9ba1b7c6ac99ca3cc3753

and

https://github.com/spring-projects/spring-session/commit/8e7c736a0aef27eda21371eaf7a237498098dcad

Anyone taking on this task I'd encourage them to break it up into multiple PRs that migrate smaller portions at a time (i.e. a sample or a package) and include

```
Migrate <what is migrated> groovy->java

(optional additional details)

Issue: gh-4939
```

I'd also encourage contributors to comment on this issue to state which piece they are going to start working on migrating. This will allow multiple people to help tackle this large issue and avoid duplicating efforts.

michael-simons:
Next: xml-jaas

michael-simons:
Next: javaconfig-form

adolfoweloy:
Migrate `UserDetailsManagerConfigurerTests` test groovy->java from security-config #4949

michael-simons:
Next: javaconfig-jdbc

michael-simons:
I will have the rest of the samples ready by the end of the weekend.
One question though: There are more Groovy tests, also dedicated to tickets and stuff. Do you want to migrate them as well?

rwinch:
@michael-simons We would like to migrate all Groovy within Spring Security to Java. Does that help?

michael-simons:
Yep, that was the question. Thanks.

rwinch:
@michael-simons Awesome. I think you were too quick for my edits so as an FYI I added a little more information above. Thanks again for your help!

rwinch:
I'm starting on config/src/test/groovy/org/springframework/security/config/annotation/authentication

jgrandja:
I'm starting on `/config/src/test/groovy/org/springframework/security/config/annotation/web`

rwinch:
I'm starting on config/src/test/groovy/org/springframework/security/config/annotation/authentication

[...]

PR #4941, opened 2018-01-02
Remove duplicate dependency

michael-simons:
Just a small polish, noticed while looking at #4939


rwinch:
Thanks for the PR @michael-simons! This is merged into master

PR #4942, opened 2018-01-03
Gh 4939 helloworld

michael-simons:
This is a suggestion how I would migrate the (nice) GebTests. While lucking through other tests, I noticed that the naming conventions differ throughout various samples. I opted for the same package and class names as in the old ones. 

If this is useful, I can work my way throughout the others while my time permits.

See #4939


rwinch:
Thanks for the PR @michael-simons! I have merged the changes into master

michael-simons:
Thanks for the feedback!

I just can comment on the original ticket when I start another.

rwinch:
PS: I updated the original ticket to request people comment on it. I probably should have done that in the first place. Thanks again for your contributions :)

PR #4946, opened 2018-01-04
Remove hardcoded Spring versions from namespaces

### LINKED ISSUE #4945: XML based samples should not use hardcoded Spring versions
### Summary

While working on migrating the Groovy based integration tests for samples I noticed that most XML-config files use hardcoded Spring versions. That is an antipattern (see for example [this answer](https://stackoverflow.com/a/20900801)).

Although XML config is a bit dated by now, it's probably useful having them as reference when working on legacy applications or integration scenarios.

michael-simons:
This fixes #4945.


rwinch:
Thanks for the PR @michael-simons! This is merged into master

PR #4947, opened 2018-01-04
Migrate xml-jaas groovy->java

michael-simons:
See #4939.


rwinch:
Thanks for the PR @michael-simons! You are really rocking this migration which helps a lot! Keep them coming :)

PR #4948, opened 2018-01-04
Migrate javaconfig-form groovy->java

michael-simons:
See #4939


rwinch:
Thanks for the PR @michael-simons! This is now merged into master

PR #4953, opened 2018-01-08
Rename userDetailsRepository to userDetailsService

izeye:
This PR renames `userDetailsRepository` to `userDetailsService` to align with its type.


rwinch:

PR #4964, opened 2018-01-15
Migrate remaining Gh 4939 samples

michael-simons:
The only sample that's missing is the CAS sample which I couldn't get to run.

See #4939


rwinch:
Thanks for the PR @michael-simons! This is now merged into master

michael-simons:
You're welcome. I wish I could have done more the last week, but project has been interesting to say.

PR #4987, opened 2018-02-01
Allow spring-security-core:compileJava to be cached

lptr:
Instead of copying classes to the compile output, we now add them directly to the JAR. This allows JavaCompile to be cached, since there are no overlapping outputs anymore.


rwinch:
@lptr Thanks for the PR! This is now merged into master

PR #5023, opened 2018-02-16
Allow httpBasic to override SecurityContextRepository

gregturn:


rwinch:
@gregturn This PR does not appear to be correct. Reactive security allows for a different security context to be used for saving the context vs reading it. This is intentional so that HTTP Basic and form log in can both be used at the same time with form log in creating a session at time of authentication and http basic not creating a session

rwinch:
I'm closing this as it is changing the intended behavior

Issue #2598, opened 2013-10-23
SEC-2377: Handle EnableWebSecurity in parent and child context


spring-projects-issues:
[Rob Winch](https://jira.spring.io/secure/ViewProfile.jspa?name=rwinch) (Migrated from [SEC-2377](https://jira.spring.io/browse/SEC-2377?redirect=false)) said:

When @EnableWebSecurity s used in both a parent and a child ApplicationContext the springSecurityFilterChain gets built twice causing an error similar to:

```
2013-10-23 23:40:00,862 [main] ERROR: Context initialization failed
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'springSecurityFilterChain' defined in class org.springframework.security.config.annotation.web.configuration.WebSecurityConfiguration: Instantiation of bean failed; nested exception is org.springframework.beans.factory.BeanDefinitionStoreException: Factory method [public javax.servlet.Filter org.springframework.security.config.annotation.web.configuration.WebSecurityConfiguration.springSecurityFilterChain() throws java.lang.Exception] threw exception; nested exception is java.lang.IllegalStateException: This object has already been built
    at org.springframework.beans.factory.support.ConstructorResolver.instantiateUsingFactoryMethod(ConstructorResolver.java:581)
    at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.instantiateUsingFactoryMethod(AbstractAutowireCapableBeanFactory.java:1025)
    at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:921)
    at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:487)
    at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:458)
    at org.springframework.beans.factory.support.AbstractBeanFactory$1.getObject(AbstractBeanFactory.java:295)
    at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:223)
    at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:292)
    at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:194)
[...]

PR #5033, opened 2018-02-21
Migrate sec2377 groovy->java

jzheaux:
Migrate org.springframework.security.config.annotation.web.configuration.sec2377 groovy->java

Issue: gh-4939


pivotal-issuemaster:
@jzheaux Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=5033)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=5033) to manually synchronize the status of this Pull Request.

config/src/test/java/org/springframework/security/config/annotation/web/configuration/sec2377/Sec2377Tests.java
```diff
@@ -0,0 +1,52 @@
...
+
+		AnnotationConfigWebApplicationContext child =
+			new AnnotationConfigWebApplicationContext();
+		child.register(Sec2377BConfig.class);
+		child.setParent(parent);
+		child.refresh();
+	}
+
+	@EnableWebSecurity
+	static class Sec2377AConfig extends WebSecurityConfigurerAdapter {
```
jgrandja: `Sec2377AConfig` should be in a different package from `Sec2377BConfig`. See the original (groovy) package structure. See original [Issue](https://github.com/spring-projects/spring-security/issues/2598)

config/src/test/java/org/springframework/security/config/annotation/web/configuration/sec2377/Sec2377Tests.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * @author Rob Winch
+ * @author Josh Cummings
+ */
+public class Sec2377Tests {
+
+	@Test
+	public void contextRefreshWhenParentAndChildRegisteredThenNoException() {
+		AnnotationConfigWebApplicationContext parent =
```
jgrandja: We should use `SpringTestRule` for all tests. For example: 

```
	@Rule
	public final SpringTestRule spring = new SpringTestRule();

this.spring.register(Sec2377AConfig.class).autowire();
```

Search for examples of `SpringTestRule`

config/src/test/java/org/springframework/security/config/annotation/web/configuration/sec2377/Sec2377Tests.java
jzheaux: Okay, no problem. Was the original issue only reproduceable when parent and child configs were in different repos? Otherwise, what is the motive for preserving the package structure?

config/src/test/java/org/springframework/security/config/annotation/web/configuration/sec2377/Sec2377Tests.java
jzheaux: To represent the scenario correctly, I need two contexts, correct? If I've got that right, and if I'm reading `SpringTestRule` correctly, then I'll need two instances of the rule as well as add a `parent` method to tie one context to the other through `SpringTestRule`, e.g.:

```
@Rule
public final SpringTestRule parent = new SpringTestRule();

@Rule
public final SpringTestRule child = new SpringTestRule();

@Test
public void test() {
  // something like this
  parent.register(Sec2377AConfig.class).autowire();
  child.register(Sec2377BConfig.class).parent(parent).autowire();
}
```

Is this the kind of thing you are expecting?

config/src/test/java/org/springframework/security/config/annotation/web/configuration/sec2377/Sec2377Tests.java
jgrandja: I didn't work on the original issue. Maybe preserving the package structure is not necessary? But the Configs were put in a separate package when this test was originally written so there must be a reason for it?

config/src/test/java/org/springframework/security/config/annotation/web/configuration/sec2377/Sec2377Tests.java
jgrandja: That looks right with one correction

`this.child.register(Sec2377BConfig.class).getContext().setParent(parent.getContext()).autowire();`

pivotal-issuemaster:
@jzheaux Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=5033)!

jgrandja:
Thanks for the PR @jzheaux! This has been merged to master.

PR #5037, opened 2018-02-22
Migrate AnonymousConfigurerTests groovy->java

jzheaux:
See #4939


config/src/test/java/org/springframework/security/config/annotation/web/configurers/AnonymousConfigurerTests.java
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	@EnableWebSecurity
+	static class InvokeTwiceDoesNotOverride extends WebSecurityConfigurerAdapter {
+
+		@Override
+		protected void configure(HttpSecurity http) throws Exception {
+			http
+				.anonymous()
+				.principal("custom")
```
jgrandja: The groovy test was configuring `key` rather than `principal`

config/src/test/java/org/springframework/security/config/annotation/web/configurers/AnonymousConfigurerTests.java
```diff
@@ -0,0 +1,77 @@
...
+
+		assertThat(result.getResponse().getContentAsString()).isEqualTo("custom");
+	}
+
+	@EnableWebSecurity
+	static class InvokeTwiceDoesNotOverride extends WebSecurityConfigurerAdapter {
+
+		@Override
+		protected void configure(HttpSecurity http) throws Exception {
+			http
```
jgrandja: Please format the `HttpSecurity` configuration with indentations. For example:

```
			http
				.anonymous()
					.key("custom")
					.and()
				.anonymous();
```

config/src/test/java/org/springframework/security/config/annotation/web/configurers/AnonymousConfigurerTests.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+public class AnonymousConfigurerTests {
+	@Rule
+	public final SpringTestRule spring = new SpringTestRule();
+
+	@Autowired
+	private MockMvc mockMvc;
+
+	@Test
+	public void requestWhenAnonymousTwiceInvokedThenDoesntOverride() throws Exception {
```
jgrandja: Rename to `requestWhenAnonymousTwiceInvokedThenDoesNotOverride`...I think it reads better?

config/src/test/java/org/springframework/security/config/annotation/web/configurers/AnonymousConfigurerTests.java
jzheaux: okay, will do.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/AnonymousConfigurerTests.java
jzheaux: Sure thing, yes, agreed.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/AnonymousConfigurerTests.java
jzheaux: Checking the key has the downside that it gets hashed by the `AnonymousAuthenticationToken` constructor, meaning that to confirm it behaviorally, I'd need to exercise the same hashing logic in the test, e.g.:

```assertThat(result.getResponse().getContentAsString()).isEqualTo("custom".hashCode());```
which is why I changed it to principal. This wasn't a problem in the original test since it just verified the structure of `AnonymousAuthenticationFilter`.

I can test the key for sure; I'm just duplicating some logic that is internal to `AnonymousAuthenticationToken`. And since both key and principal are configurable values on `AnonymousAuthenticationFilter`, it seemed like a fair trade-off for confirming that configs aren't overridden.

Are there other things I should consider here? (Other than the realization that I should have put this explanation in my original commit to explain why I departed from the original test :) )

Comment (jgrandja) [reply config/src/test/java/org/springframework/security/config/annotation/web/configurers/AnonymousConfigurerTests.java]: If you pass in the same `key` to `AnonymousAuthenticationToken` as is configured via `.anonymous().key("custom")` than the test should pass as the hash check will succeed in the `AnonymousAuthenticationProvider`. So try this in your test and it should work:

```
AnonymousAuthenticationToken authRequest = new AnonymousAuthenticationToken("custom", "anonymous", AuthorityUtils.createAuthorityList("ROLE_ANONYMOUS"));
		
MvcResult result = this.mockMvc.perform(get("/").with(authentication(authRequest))).andReturn();
```

File: config/src/test/java/org/springframework/security/config/annotation/web/configurers/AnonymousConfigurerTests.java [hunk 4]
```diff
@@ -0,0 +1,77 @@
...
+		@Override
+		protected void configure(HttpSecurity http) throws Exception {
+			http
+				.anonymous()
+					.principal("custom")
+					.and()
+				.anonymous();
+		}
+
+		@Controller
```
rwinch: Use `@RestController` and remove `@ResponseBody`

config/src/test/java/org/springframework/security/config/annotation/web/configurers/AnonymousConfigurerTests.java
```diff
@@ -0,0 +1,77 @@
...
+		static class PrincipalController {
+			@RequestMapping("/")
```
rwinch: Prefer `@GetMapping` to general `@RequestMapping`

config/src/test/java/org/springframework/security/config/annotation/web/configurers/AnonymousConfigurerTests.java
```diff
@@ -0,0 +1,77 @@
...
+			@ResponseBody String principal() {
+				Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
```
rwinch: Make `Authentication` an argument to the method rather than resolving through `SecurityContextHolder` which requires a static method to work

rwinch:
When you have completed your changes, can you also squash your commits please?

config/src/test/java/org/springframework/security/config/annotation/web/configurers/AnonymousConfigurerTests.java
jzheaux: I originally attempted this, as I feel the same way; however `Authentication` doesn't appear to resolve, e.g. the method param is always `null`. I also attempted with `@AuthenticationPrincipal String principal` as well as simply `SecurityContext`. Happy to dig into this further, but do you happen to see another configuration piece I am missing?

config/src/test/java/org/springframework/security/config/annotation/web/configurers/AnonymousConfigurerTests.java
jzheaux: will do

config/src/test/java/org/springframework/security/config/annotation/web/configurers/AnonymousConfigurerTests.java
jzheaux: will do

config/src/test/java/org/springframework/security/config/annotation/web/configurers/AnonymousConfigurerTests.java
jzheaux: Looks like if I add `@EnableWebMvc` to my configuration, then `AuthenticationPrincipalArgumentResolver` gets wired, and now I can use `@AuthenticationPrincipal`, which does clean up the code nicely. Does that sound as you would expect?

jgrandja:
Thanks for the PR @jzheaux. This is now in master.

PR #5051, opened 2018-02-27
Migrate config-debug groovy->java

jzheaux:
All tests in `org.springframework.security.config.debug` are migrated.

Note that `SecurityDebugBeanFactoryPostProceessorTest` preserves the original structure-verifying strategy used in the Groovy test. Verifying debug behavior turns out to be fairly tricky since being behaviorally invisible is in its nature.

Issue: gh-4939


rwinch:

PR #5066, opened 2018-03-03
NamespaceGlobalMethodSecurity groovy->java

jzheaux:
Note that the `WhenUsingAspectJ` tests are still simply verifying structure instead of behavior. This is because the project appearsto be misconfigured in some way such that AspectJ advice isn't getting woven in at runtime. The original Groovy tests also only verified structure and they may be that way for a similar reason.

Either way, I will open up a ticket so we can review why that is the case and if there is a good fix.

Issue: gh-4939


config/src/test/java/org/springframework/security/config/annotation/method/configuration/MethodSecurityServiceConfig.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+package org.springframework.security.config.annotation.method.configuration;
+
+import org.springframework.context.annotation.Bean;
+
+/**
+ * @author Josh Cummings
+ */
+public interface MethodSecurityServiceConfig {
```
rwinch: If you are going to extract this out, I'd do so as a class for consistency with the other tests. Then in the tests just register multiple configurations.

config/src/test/java/org/springframework/security/config/annotation/method/configuration/NamespaceGlobalMethodSecurityExpressionHandlerTests.java
```diff
@@ -0,0 +1,101 @@
...
+			.doesNotThrowAnyException();
+
+		assertThatThrownBy(() -> this.service.postHasPermission("denied"))
+			.isInstanceOf(AccessDeniedException.class);
+	}
+
+	@EnableGlobalMethodSecurity(prePostEnabled = true)
+	public static class CustomAccessDecisionManagerConfig extends GlobalMethodSecurityConfiguration {
+		@Bean
+		public MethodSecurityServiceImpl service() {
```
rwinch: Could this be removed an instead register MethodSecurityServiceConfig?

config/src/test/java/org/springframework/security/config/annotation/method/configuration/MethodSecurityServiceConfig.java
jzheaux: great, will do

config/src/test/java/org/springframework/security/config/annotation/method/configuration/NamespaceGlobalMethodSecurityExpressionHandlerTests.java
jzheaux: Yes, just missed it.

rwinch:

PR #5068, opened 2018-03-03
Update ExpressionBasedFilterInvocationSecurityMetadataSource.java ,make it easy to use

json20080301:
ExpressionBasedFilterInvocationSecurityMetadataSource(Map<String, String> urlMap ) make ExpressionBasedFilterInvocationSecurityMetadataSource easy to use.

this change  can load  SecurityMetadataSource  from propertie  config ; then  you can load config from other store like file ,db or other store. 

this change can make security config  like shiro config ;[http://shiro.apache.org/spring.html]
````
<bean id="shiroFilter" class="org.apache.shiro.spring.web.ShiroFilterFactoryBean">
    <property name="securityManager" ref="securityManager"/>
    <!-- override these for application-specific URLs if you like:
    <property name="loginUrl" value="/login.jsp"/>
    <property name="successUrl" value="/home.jsp"/>
    <property name="unauthorizedUrl" value="/unauthorized.jsp"/> -->
    <!-- The 'filters' property is not necessary since any declared javax.servlet.Filter bean  -->
    <!-- defined will be automatically acquired and available via its beanName in chain        -->
    <!-- definitions, but you can perform instance overrides or name aliases here if you like: -->
    <!-- <property name="filters">
        <util:map>
            <entry key="anAlias" value-ref="someFilter"/>
        </util:map>
    </property> -->
    <property name="filterChainDefinitions">
        <value>
            # some example chain definitions:
            /admin/** = authc, roles[admin]
            /docs/** = authc, perms[document:read]
            /** = authc
            # more URL-to-FilterChain definitions here
        </value>
    </property>
</bean>
 
````

### DIFF
--- a/web/src/main/java/org/springframework/security/web/access/expression/ExpressionBasedFilterInvocationSecurityMetadataSource.java
+++ b/web/src/main/java/org/springframework/security/web/access/expression/ExpressionBasedFilterInvocationSecurityMetadataSource.java
@@ -32,31 +32,7 @@ public final class ExpressionBasedFilterInvocationSecurityMetadataSource extends
         Assert.notNull(expressionHandler, "A non-null SecurityExpressionHandler is required");
     }
     
-     /**
-     * @param urlMap store url<--> expression
-     */
-    public ExpressionBasedFilterInvocationSecurityMetadataSource(Map<String, String> urlMap ) {
-        loadSecurityMetadataSourceFromUrlMap(urlMap);
-    }
 
-    /**
-     * @param urlMap store url<--> expression
-     */
-    public synchronized void loadSecurityMetadataSourceFromUrlMap(Map<String, String> urlMap) {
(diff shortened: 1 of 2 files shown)


rwinch:
Thank you for the PR!

I think perhaps a better approach would be to create a factory bean that creates a `ExpressionBasedFilterInvocationSecurityMetadataSource` This would would provide a number of advantages:

* Users could customize the type of `RequestMatcher` created from the String (i.e. MVC, Ant, RegEx, etc)
* Users could easily configure the `WebSecurityExpressionHandler`. The current solution injects `DefaultWebSecurityExpressionHandler` but it provides no way for it to be configured with things like a `BeanResolver` which limits its functionality
* There would be no need to delegate to all the methods

Before this would get merged we also need some tests added. Is this something you would be able to do?

rwinch:
I'm closing this due to lack of activity.

Issue #2984, opened 2014-10-30
SEC-2758: Consistency with ROLE_


spring-projects-issues:
[Greg Turnquist](https://jira.spring.io/secure/ViewProfile.jspa?name=gregturn) (Migrated from [SEC-2758](https://jira.spring.io/browse/SEC-2758?redirect=false)) said:
# Updated Description

Spring Security does not treat roles consistently. For example, the Java Configuration allows you to perform something like:

```
.withUser("user").password("password").roles("USER")
```

```
http
   .authorizeRequests()
       .anyRequest().hasRole("USER")
```

Internally, these is automatically prefixed with "ROLE_". This means when using any of the following:

```
// or any other SpEL annotation
@PreAuthorize("hasRole('ROLE_USER')")
```

```
// JSR 250
@RolesAllowed("ROLE_USER")
```

```
<http use-expressions="true">
   <intercept-url pattern="/**" access="hasRole('ROLE_USER')"/>
</http>
```

we must restate the ROLE_ even though it has already been expressed as a role.

h1. How should we improve this?

Anytime we see role elsewhere, it should automatically add the prefix of "ROLE_" if it is not already there (this helps things remain passive). So that means we should be able to do the following (which is the equivalent of the above example):

```
// or any other SpEL annotation
@PreAuthorize("hasRole('USER')")
```

```
// JSR 250
@RolesAllowed("USER")
```

```
<http use-expressions="true">
   <intercept-url pattern="/**" access="hasRole('USER')"/>
</http>
```

h2. Implementation Details

h3. Expressions

To support expressions we will update the `SecurityExpressionRoot` to automatically prefix hasRole and hasAnyRole with "ROLE_" if it does not exist. For example:

```
public abstract class SecurityExpressionRoot implements SecurityExpressionOperations {
    ...
    private String defaultRolePrefix = "ROLE_";

    ...

    /**
     * <p>
     * Sets the default prefix to be added to {@link #hasAnyRole(String...)} or
     * {@link #hasRole(String)}. For example, if hasRole("ADMIN") or hasRole("ROLE_ADMIN") is passed in,
     * then the role ROLE_ADMIN will be used when the defaultRolePrefix is "ROLE_" (default).
     * </p>
     *
     * <p>
     * If null or empty, then no default role prefix is used.
     * </p>
     *
     * @param defaultRolePrefix the default prefix to add to roles. Default "ROLE_".
     */
[...]

PR #5079, opened 2018-03-07
Sec2758Tests groovy->java

jzheaux:
Note that the old groovy test used a configuration of

```
http
    .authorizeRequests()
        .anyRequest().hasAnyAuthority("USER")
```

However, as I read the issue, gh-2984, the problem this issue
identifies is the non-passive change of defaulting to prefix
ROLE_ with all role-based configuration methods. So, the test now
does the following:

```
http
    .authorizeRequests()
        .anyRequest().access("hasAnyRole('USER')")
```

which demonstrates, given the configuration in this test, that
ROLE_ is correctly not prefixed in this expression, even though
it is a role-based configuration.

Issue: gh-4939


rwinch:

PR #5084, opened 2018-03-07
Add support for authorization_code grant

### LINKED ISSUE #4928: Provide support for authorization_code grant
The OAuth 2.0 Login feature is implemented using the `authorization_code` grant with the addition of fetching the UserInfo resource after the flow is complete.

We should also provide support for the `authorization_code` grant as defined by OAuth 2.0 so clients are able to obtain access token(s) without having to use OAuth 2.0 Login.

jgrandja:
Resolves #4928

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,14 +14,6 @@
  * limitations under the License.
  */
 package org.springframework.security.config.annotation.web.builders;
-
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-
-import javax.servlet.Filter;
-import javax.servlet.http.HttpServletRequest;
 
 import org.springframework.context.ApplicationContext;
 import org.springframework.http.HttpMethod;
@@ -56,12 +48,13 @@ import org.springframework.security.config.annotation.web.configurers.SecurityCo
 import org.springframework.security.config.annotation.web.configurers.ServletApiConfigurer;
 import org.springframework.security.config.annotation.web.configurers.SessionManagementConfigurer;
 import org.springframework.security.config.annotation.web.configurers.X509Configurer;
+import org.springframework.security.config.annotation.web.configurers.oauth2.OAuth2Configurer;
+import org.springframework.security.config.annotation.web.configurers.oauth2.client.OAuth2LoginConfigurer;
 import org.springframework.security.config.annotation.web.configurers.openid.OpenIDLoginConfigurer;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.context.SecurityContext;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.security.core.userdetails.UserDetailsService;
-import org.springframework.security.config.annotation.web.configurers.oauth2.client.OAuth2LoginConfigurer;
 import org.springframework.security.web.DefaultSecurityFilterChain;
 import org.springframework.security.web.PortMapper;
 import org.springframework.security.web.PortMapperImpl;
@@ -78,6 +71,13 @@ import org.springframework.util.Assert;
 import org.springframework.web.cors.CorsConfiguration;
 import org.springframework.web.filter.CorsFilter;
 import org.springframework.web.servlet.handler.HandlerMappingIntrospector;
+
+import javax.servlet.Filter;
+import javax.servlet.http.HttpServletRequest;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
 
 /**
  * A {@link HttpSecurity} is similar to Spring Security's XML &lt;http&gt; element in the
@@ -989,6 +989,18 @@ public final class HttpSecurity extends
 	 */
 	public OAuth2LoginConfigurer<HttpSecurity> oauth2Login() throws Exception {
 		return getOrApply(new OAuth2LoginConfigurer<>());
+	}
+
+	/**
+	 * Configures support for the <a target="_blank" href="https://tools.ietf.org/html/rfc6749">OAuth 2.0 Authorization Framework</a>.
+	 *
+	 * @author Joe Grandja
+	 * @since 5.1
+	 * @return the {@link OAuth2Configurer} for further customizations
+	 * @throws Exception
+	 */
+	public OAuth2Configurer oauth2() throws Exception {
+		return getOrApply(new OAuth2Configurer());
 	}
 
 	/**
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2ClientConfigurerUtils.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2ClientConfigurerUtils.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2002-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.config.annotation.web.configurers.oauth2.client;
+
+import org.springframework.beans.factory.BeanFactoryUtils;
+import org.springframework.beans.factory.NoUniqueBeanDefinitionException;
+import org.springframework.context.ApplicationContext;
+import org.springframework.security.config.annotation.web.HttpSecurityBuilder;
+import org.springframework.security.config.annotation.web.configurers.AbstractHttpConfigurer;
+import org.springframework.security.oauth2.client.InMemoryOAuth2AuthorizedClientService;
+import org.springframework.security.oauth2.client.OAuth2AuthorizedClientService;
+import org.springframework.security.oauth2.client.registration.ClientRegistrationRepository;
+
+import java.util.Map;
+
+/**
+ * Utility methods for the OAuth 2.0 Client {@link AbstractHttpConfigurer}'s.
+ *
+ * @author Joe Grandja
+ * @since 5.1
+ */
+final class OAuth2ClientConfigurerUtils {
+
+	private OAuth2ClientConfigurerUtils() {
+	}
+
+	static <B extends HttpSecurityBuilder<B>> ClientRegistrationRepository getClientRegistrationRepository(B builder) {
+		ClientRegistrationRepository clientRegistrationRepository = builder.getSharedObject(ClientRegistrationRepository.class);
+		if (clientRegistrationRepository == null) {
+			clientRegistrationRepository = getClientRegistrationRepositoryBean(builder);
+			builder.setSharedObject(ClientRegistrationRepository.class, clientRegistrationRepository);
+		}
+		return clientRegistrationRepository;
+	}
+
+	private static <B extends HttpSecurityBuilder<B>> ClientRegistrationRepository getClientRegistrationRepositoryBean(B builder) {
+		return builder.getSharedObject(ApplicationContext.class).getBean(ClientRegistrationRepository.class);
+	}
+
+	static <B extends HttpSecurityBuilder<B>> OAuth2AuthorizedClientService getAuthorizedClientService(B builder) {
+		OAuth2AuthorizedClientService authorizedClientService = builder.getSharedObject(OAuth2AuthorizedClientService.class);
+		if (authorizedClientService == null) {
+			authorizedClientService = getAuthorizedClientServiceBean(builder);
+			if (authorizedClientService == null) {
+				authorizedClientService = new InMemoryOAuth2AuthorizedClientService(getClientRegistrationRepository(builder));
+			}
+			builder.setSharedObject(OAuth2AuthorizedClientService.class, authorizedClientService);
+		}
+		return authorizedClientService;
+	}
+
+	private static <B extends HttpSecurityBuilder<B>> OAuth2AuthorizedClientService getAuthorizedClientServiceBean(B builder) {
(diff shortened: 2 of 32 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
```diff
@@ -991,6 +992,10 @@ private ApplicationContext getContext() {
 		return getOrApply(new OAuth2LoginConfigurer<>());
 	}
 
+	public OAuth2Configurer<HttpSecurity> oauth2() throws Exception {
```
rwinch: Add @since 5.1

rwinch:
We want to ensure that we update OAuth2AuthorizationCodeGrantFilter to only process specific URLs. We also want to update OAuth2AuthorizationCodeGrantFilter to use a saved request for returning to the original URL.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2ClientConfigurerUtils.java
```diff
@@ -0,0 +1,69 @@
...
+			}
+			builder.setSharedObject(OAuth2AuthorizedClientService.class, authorizedClientService);
+		}
+		return authorizedClientService;
+	}
+
+	private static <B extends HttpSecurityBuilder<B>> OAuth2AuthorizedClientService getAuthorizedClientServiceBean(B builder) {
+		Map<String, OAuth2AuthorizedClientService> authorizedClientServiceMap = BeanFactoryUtils.beansOfTypeIncludingAncestors(
+				builder.getSharedObject(ApplicationContext.class), OAuth2AuthorizedClientService.class);
+		return (!authorizedClientServiceMap.isEmpty() ? authorizedClientServiceMap.values().iterator().next() : null);
```
jzheaux: Without a lot of context, my questions here are more for my education. How do you know that you are picking the "right" instance here? If there were indeed more than one in the map, could this cause a misconfiguration?

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/AuthorizationCodeGrantConfigurerTests.java
```diff
@@ -0,0 +1,165 @@
...
+		when(accessTokenResponse.getAccessToken()).thenReturn(mock(OAuth2AccessToken.class));
+		accessTokenResponseClient = mock(OAuth2AccessTokenResponseClient.class);
+		when(accessTokenResponseClient.getTokenResponse(any(OAuth2AuthorizationCodeGrantRequest.class))).thenReturn(accessTokenResponse);
+	}
+
+	@Test
+	public void configureWhenAuthorizationRequestThenRedirectForAuthorization() throws Exception {
+		this.spring.register(AuthorizationCodeGrantConfig.class).autowire();
+
+		MvcResult mvcResult = this.mockMvc.perform(get("/oauth2/authorization/registration-1"))
```
jzheaux: @rwinch showed me a cool reduction for this in `PortMapperConfigurerTests`:

```
this.mockMvc.perform(get("/oauth/authorization/registration-1"))
    .andExpect(redirectedUrl("https://provider.com/oauth2/authorize\\?..."));
```

Not sure if this `redirectedUrl` returns a `ResultMatcher` that supports the regex you are doing, but it might be a valuable simplification.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/AuthorizationCodeGrantConfigurerTests.java
```diff
@@ -0,0 +1,165 @@
...
+
+	@EnableWebSecurity
+	static class AuthorizationCodeGrantConfig extends WebSecurityConfigurerAdapter {
+		@Override
+		protected void configure(HttpSecurity http) throws Exception {
+			http
+				.authorizeRequests()
+					.anyRequest().authenticated();
+
+			this.authorizationCodeGrant(http)
```
jzheaux: Again, just for my education, why is the preferred way to configure `authorizationCodeGrant` by `apply`ing it to `http`? Is that because the fluent configurer method just hasn't been added yet to `HttpSecurity` or is there a different reason?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/authentication/OAuth2AuthorizationCodeAuthenticationToken.java
```diff
@@ -0,0 +1,114 @@
...
+	private OAuth2AccessToken accessToken;
+
+	/**
+	 * This constructor should be used when the Authorization Request/Response is complete.
+	 *
+	 * @param clientRegistration the client registration
+	 * @param authorizationExchange the authorization exchange
+	 */
+	public OAuth2AuthorizationCodeAuthenticationToken(ClientRegistration clientRegistration,
+														OAuth2AuthorizationExchange authorizationExchange) {
```
jzheaux: Is this the preferred whitespacing or is Github (or IntelliJ) just being weird?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/authentication/OAuth2AuthorizationExchangeValidator.java
```diff
@@ -0,0 +1,54 @@
...
+import org.springframework.security.oauth2.core.endpoint.OAuth2AuthorizationResponse;
+
+/**
+ * A validator for an &quot;exchange&quot; of an OAuth 2.0 Authorization Request and Response.
+ *
+ * @author Joe Grandja
+ * @since 5.1
+ * @see OAuth2AuthorizationExchange
+ */
+final class OAuth2AuthorizationExchangeValidator {
```
jzheaux: I'm wondering about this being a final class with static methods. Is your reasoning that since we are following a spec, we don't want folks to be able to extend and supply their own validation rules? (Also, I know I'm late to the party--you and @rwinch may have already had these convos earlier on.)

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationCodeGrantFilter.java
```diff
@@ -0,0 +1,192 @@
...
+ * @see OAuth2AuthorizationCodeAuthenticationToken
+ * @see OAuth2AuthorizationCodeAuthenticationProvider
+ * @see OAuth2AuthorizationRequest
+ * @see OAuth2AuthorizationResponse
+ * @see AuthorizationRequestRepository
+ * @see OAuth2AuthorizationRequestRedirectFilter
+ * @see ClientRegistrationRepository
+ * @see OAuth2AuthorizedClient
+ * @see OAuth2AuthorizedClientService
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc6749#section-4.1">Section 4.1 Authorization Code Grant</a>
```
jzheaux: I like this.

samples/boot/oauth2/authcodegrant/src/main/resources/application.yml
```diff
@@ -0,0 +1,24 @@
+server:
+  port: 8080
+
+logging:
+  level:
+    root: INFO
+    org.springframework.web: INFO
+    org.springframework.security: INFO
+#    org.springframework.boot.autoconfigure: DEBUG
```
jzheaux: Is this something that we keep in our samples to make it easy for folks to toggle it on and off while they are playing around with the sample?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2ClientConfigurerUtils.java
jgrandja: The main reason for the use of `BeanFactoryUtils.beansOfTypeIncludingAncestors()` is that it does not throw a `BeanException` if it's not registered in the context, unlike `ApplicationContext.getBean()`. If the application is not a Spring Boot application (`OAuth2AuthorizedClientService` is auto-configured by Boot) and the application has not registered an `OAuth2AuthorizedClientService` than this will return empty. However, if there is more than one instance than this is a misconfiguration. So good catch, we should handle this scenario. Possibly checking for `@Primary` or fast-failing. Would you mind logging an issue for this?

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/AuthorizationCodeGrantConfigurerTests.java
jgrandja: Yeah I've used the `redirectedUrl()` `ResultMatcher` but from what I recall it does not support regex. Hence why I'm using `assertThat().matches()`

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/AuthorizationCodeGrantConfigurerTests.java
jgrandja: I added `HttpSecurity.oauth2()` after this test. I need to update this test.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/authentication/OAuth2AuthorizationCodeAuthenticationToken.java
jgrandja: Not sure I understand? Can you clarify?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/authentication/OAuth2AuthorizationExchangeValidator.java
jgrandja: It's much easier to change things passively but as soon as it's a public API than it's obviously very difficult to change and time to deprecate. We can always expose things later if it's requested by the community. Moreover, these specific validation checks need to happen and should not be easily overridden or it will introduce a vulnerability

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationCodeGrantFilter.java
jgrandja: Yeah Nimbus does the same thing in their javadoc. I think it's very useful to the user to know which spec and which part of the spec the feature is implementing

samples/boot/oauth2/authcodegrant/src/main/resources/application.yml
jgrandja: yes that's the motivation why I put it in there

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2ClientConfigurerUtils.java
jgrandja: @jzheaux Actually, don't worry about logging this issue. I will handle it in this PR.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/authentication/OAuth2AuthorizationCodeAuthenticationToken.java
jzheaux: On my screen, the tabbing looks funny. The `OAuth2AuthorizationExchange` line is several tabs to the right of the `ClientRegistration` line.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/authentication/OAuth2AuthorizationExchangeValidator.java
jzheaux: Okay. Good points.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/authentication/OAuth2AuthorizationCodeAuthenticationToken.java
jgrandja: I see. If you view the full file (in GitHub) than the tabbing is normal.

samples/boot/oauth2/authcodegrant/src/main/java/sample/config/SecurityConfig.java
```diff
@@ -0,0 +1,56 @@
...
+			.formLogin()
+				.and()
+			.oauth2()
+				.client()
+					.authorizationCodeGrant();
+	}
+
+	@Bean
+	public UserDetailsService userDetailsService() {
+		UserDetails userDetails = User.builder()
```
rwinch: Please switch this to `withDefaultPasswordEncoder()` which is slightly better than storing passwords using `{noop}` since the password in memory is hashed.

jgrandja: ok will do

jgrandja: Just noticed that `User.withDefaultPasswordEncoder()` is deprecated so I'll update it to:

`.passwordEncoder(PasswordEncoderFactories.createDelegatingPasswordEncoder()::encode)`

Ok?

rwinch: Please use withDefaultPasswordEncoder. If you read the documentation...It is only deprecated to draw attention to the fact that you shouldn't use it in production environments.

jgrandja: Gotchya

jgrandja: Update applied and pushed

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationCodeGrantFilter.java
```diff
@@ -0,0 +1,205 @@
...
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc6749#section-4.1.2">Section 4.1.2 Authorization Response</a>
+ */
+public class OAuth2AuthorizationCodeGrantFilter extends OncePerRequestFilter {
```
rwinch: If I enable oauth2Login() and oauth2().authorizationCode(). We can fix this by ensuring that the log in processes /login/oauth2/code/{registratationId} and this filter only processes /authorize/oauth2/code/{registrationId}. We can update the default template to be something like: {baseUrl}/{action}/oauth2/code/{registrationId} such that for login action = login and for authorizationCode action = authorize

jgrandja:
@rwinch I've applied all the updates as per your comments. Good to go for next review.

jgrandja:
Merged via 982fc36

Issue #5073, opened 2018-03-06
Inconsistent behaviour of TestingAuthenticationToken constructors


e-g-hategan:
### Summary

The 2 constructors are inconsistent in setting the authenticated flag and creates confusion. 
I've switched from String authorities to List of GrantedAuthority and then spent hours debugging.

The excerpt below from TestingAuthenticationToken talks by itself
```
	public TestingAuthenticationToken(Object principal, Object credentials,
			String... authorities) {
		this(principal, credentials, AuthorityUtils.createAuthorityList(authorities));
		setAuthenticated(true); // !!! this is authenticated
	}

	public TestingAuthenticationToken(Object principal, Object credentials,
			List<GrantedAuthority> authorities) {
		super(authorities);
		this.principal = principal;
		this.credentials = credentials;
                // !!! this is NOT authenticated
	}
```

### Actual Behavior

If the TestingAuthenticationToken is constructed with String authorities, it comes out authenticated. 
If the TestingAuthenticationToken is constructed with List of GrantedAuthority it comes out unauthenticated. 

### Expected Behavior

All constructors where authorities are specified should return an authenticated token. 

### Configuration

Not necessary.

### Version

4.2.3

### Sample

See above

e-g-hategan:
You're right, I've updated the description.

PR #5085, opened 2018-03-07
Authorities authenticate TestingAuthenticationToken

jzheaux:
In other extensions of `AbstractAuthenticationToken`, the constructors
that include `authorities` call `setAuthenticated(true)`. This includes
`PreAuthenticated`-, `UsernamePassword`-, and
`RememberMeAuthenticationToken`.

This change brings `TestingAuthenticationToken` in line with that
convention.

Note that this was done once already to one of the constructors
(ee13be4) in `TestingAuthenticationToken` that takes an arity of
`authorities`. It was not propagated to the constructor that takes a
collection, which is what this commit remedies.

Fixes: gh-5073

### DIFF
--- a/cas/src/test/java/org/springframework/security/cas/web/CasAuthenticationFilterTests.java
+++ b/cas/src/test/java/org/springframework/security/cas/web/CasAuthenticationFilterTests.java
@@ -16,11 +16,6 @@
 
 package org.springframework.security.cas.web;
 
-import static org.assertj.core.api.Assertions.*;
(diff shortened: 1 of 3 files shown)


rwinch:
Thanks for the PR @jzheaux! Can you please add a test that fails before your changes and verifies the correct behavior?

jzheaux:
Added a test. Let me know what else is needed.

rwinch:

Issue #5099, opened 2018-03-09
Custom Base64EncodingTextEncryptor not possible for Encryptors.queryableText()


ksdev-pl:
Hi

When the user wants to store an encrypted text in the database (in text, not binary format) using the available `TextEncryptor` (for example through `Encryptors.delux()`), the resultant String is hex-encoded (through `HexEncodingTextEncryptor`).

Hex-encoding is very inefficient when it comes to size. So the user writes his own `Base64EncodingTextEncryptor` to use it instead of `HexEncodingTextEncryptor`, as the Base64 is much more efficient.

It's easy to use such a `Base64EncodingTextEncryptor` to wrap the default `Encryptors` methods, like the `Encryptor.stronger()`. But it's not easily possible (AFAIK) to use the Base64 encoding to wrap a `Encryptors.queryableText()` method (because it already returns a hex-encoded string, unless we decode the hex-encoded string to encode it again in Base64, which is not the best option).

The problem:

- there is no possibility to use a `TextEncryptor` other than `HexEncodingTextEncryptor` for the "queryable" texts (that is, texts encrypted with empty iv).
- "queryable" text functionality justifies the empty iv (when it's really needed).

Possible solutions:

- public `AesBytesEncryptor` (so we can wrap it with custom `TextEncryptor` - when using empty iv)
- something like `noIvText` method in `Encryptors` (so we can wrap an `Encryptors` empty iv method with custom `TextEncryptor`)

Or is there another possibility?

ksdev-pl:
Another simple solution:

```java
public static BytesEncryptor queryableText(CharSequence password, CharSequence salt) {
	return new AesBytesEncryptor(password.toString(), salt);
}
```

Having an additional `queryableText` method which returns a `BytesEncryptor`, the user can wrap it in any custom `TextEncryptor` he wants - he's no longer limited to `HexEncodingTextEncryptor`.

PR #5102, opened 2018-03-10
Add a queryable method to Encryptors

ksdev-pl:
A proposition related to: #5099

### DIFF
--- a/crypto/src/main/java/org/springframework/security/crypto/encrypt/Encryptors.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/encrypt/Encryptors.java
@@ -91,6 +91,21 @@ public class Encryptors {
 	}
 
 	/**
+	 * Creates an encryptor for queryable bytes that uses standard password-based
+	 * encryption. Uses a 16-byte all-zero initialization vector so encrypting the same
+	 * data results in the same encryption result. This is done to allow encrypted data to
+	 * be queried against.
+	 *
+	 * @param password the password used to generate the encryptor's secret key; should
+	 * not be shared
+	 * @param salt a hex-encoded, random, site-global salt value to use to generate the
+	 * secret key
+	 */
+	public static BytesEncryptor queryable(CharSequence password, CharSequence salt) {
+		return new AesBytesEncryptor(password.toString(), salt);
+	}
+
+	/**
 	 * Creates an encryptor for queryable text strings that uses standard password-based
 	 * encryption. Uses a 16-byte all-zero initialization vector so encrypting the same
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@ksdev-pl Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=5102)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=5102) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ksdev-pl Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=5102)!

crypto/src/main/java/org/springframework/security/crypto/encrypt/Encryptors.java
```diff
@@ -90,6 +90,21 @@ public static TextEncryptor text(CharSequence password, CharSequence salt) {
...
+	 * encryption. Uses a 16-byte all-zero initialization vector so encrypting the same
+	 * data results in the same encryption result. This is done to allow encrypted data to
+	 * be queried against.
+	 *
+	 * @param password the password used to generate the encryptor's secret key; should
+	 * not be shared
+	 * @param salt a hex-encoded, random, site-global salt value to use to generate the
+	 * secret key
+	 */
+	public static BytesEncryptor queryable(CharSequence password, CharSequence salt) {
```
SangameshUppin: Why is the method called 'queryable'?, Should this class allow encryptions specific to make query.

ksdev-pl: I'm not sure if it's the best name. But currently we have:
- (Encryptor) which is "stronger"
- (Encryptor) which is "standard"
- (Encryptor) which is "delux"
- so an Encryptor which is "queryable" seemed not so bad.

Currently we have a method named "queryableText", which returns a `TextEncryptor`. Maybe the same method which returns a `BytesEncryptor` should be called "queryableBytes"? But we don't have "strongerBytes" or "standardBytes", just "stronger" and "standard", so maybe the "queryable" is enough?

rwinch:
I don't want to continue adding methods to Encryptors. The class was a mistake because it assumes that security is static. What was stronger when the class was written is no longer stronger. Instead we should open up the constructors of the specific implementations and deprecate algorithms that are not secure anymore. If you would, like you can submit a new PR to make AesByesEncryptor public which should achieve the same end result for you.

PR #5111, opened 2018-03-13
HttpSessionOAuth2AuthorizationRequestRepository handle multiple OAuth2AuthorizationRequest per session

### LINKED ISSUE #5110: HttpSessionOAuth2AuthorizationRequestRepository handle multiple OAuth2AuthorizationRequest per session
The current implementation of `HttpSessionOAuth2AuthorizationRequestRepository` should be capable of saving multiple `OAuth2AuthorizationRequest`'s for a particular `HttpSession`.

There may be a scenario where a user triggers an Authorization Request for more than one client at a time (in subsequent steps).

jgrandja:
Resolves #5110

### DIFF
--- a/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationRequestRedirectFilterTests.java
+++ b/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationRequestRedirectFilterTests.java
@@ -17,6 +17,7 @@ package org.springframework.security.oauth2.client.web;
 
 import org.junit.Before;
 import org.junit.Test;
+import org.mockito.ArgumentCaptor;
 import org.springframework.http.HttpStatus;
 import org.springframework.mock.web.MockHttpServletRequest;
 import org.springframework.mock.web.MockHttpServletResponse;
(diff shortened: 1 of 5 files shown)


oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationRequestRedirectFilterTests.java
```diff
@@ -17,6 +17,10 @@
 
 import org.junit.Before;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.ArgumentCaptor;
+import org.powermock.core.classloader.annotations.PrepareForTest;
```
rwinch: If possible let's avoid powermock

jgrandja: Any particular reason to avoid using powermock? There are a few classes in `oauth2-core` that are marked as `final` and so I've used powermock quite a bit already in the oauth2 specific tests.

rwinch: You can mock the interfaces instead. First by mocking the interface, you isolate the test as a unit test which provides a better unit test. Second, powermock is very powerful but historically is not as stable as I'd like. This is due to the fact that it works using internal APIs. Ideally we would get rid of it, but practically it provides a lot of value when working with things like static methods. The alternative is to wrap every static method in an interface declaration which isn't very realistic in my opinion.

jgrandja: Gotchya. Yes I can mock `OAuth2AuthorizationRequestRepository`

jgrandja:
@rwinch Feedback updates have been applied.

rwinch:
This is now merged via 59cef7d339de9d3a825a5ec35643d50e3f03cd4e

PR #5122, opened 2018-03-16
Provide Bearer Token resolving strategy

### LINKED ISSUE #5121: Provide capability for resolving Bearer Token
This feature implements [The OAuth 2.0 Authorization Framework: Bearer Token Usage](https://tools.ietf.org/html/rfc6750) and will be leveraged by Resource Server and Authorization Server.

Provide a `BearerTokenResolver` interface with the default implementation resolving from the Authorization Request Header field.

vpavic:
This PR creates `oauth2-resource-server module` and provides a strategy for resolving Bearer Token from HTTP request together with default implementation that aligns with RFC 6750.

The default implementation, `DefaultBearerTokenResolver`, by default uses Authorization Request Header Field for token resolution. It also optionally supports resolution using Form-Encoded Body Parameter and URI Query Parameter. This behavior closely follows [Section 2 of RFC 6750](https://tools.ietf.org/html/rfc6750#section-2) which states:

- Authorization Request Header Field:
> Clients SHOULD make authenticated requests with a bearer token using the "Authorization" request header field with the "Bearer" HTTP authorization scheme.  Resource servers MUST support this method.
- Form-Encoded Body Parameter:
> The "application/x-www-form-urlencoded" method SHOULD NOT be used except in application contexts where participating browsers do not have access to the "Authorization" request header field. Resource servers MAY support this method.
- URI Query Parameter:
> Because of the security weaknesses associated with the URI method (see Section 5), including the high likelihood that the URL containing the access token will be logged, it SHOULD NOT be used unless it is impossible to transport the access token in the "Authorization" request header field or the HTTP request entity-body. Resource servers MAY support this method.

Additionally, in presence of multiple methods of transferring token, an exception will be thrown, as per:

> Clients MUST NOT use more than one method to transmit the token in each request.

> The request is missing a required parameter, includes an unsupported parameter or parameter value, repeats the same parameter, uses more than one method for including an access token, or is otherwise malformed. The resource server SHOULD respond with the HTTP 400 (Bad Request) status code.

The PR also includes `BearerTokenAuthenticationException` and `BearerTokenError` which align with [Section 3 of RFC 6750](https://tools.ietf.org/html/rfc6750#section-3) as well as `BearerTokenErrorCodes` to enumerate standard Bearer Token error codes ([Section 3.1 of RFC 6750](https://tools.ietf.org/html/rfc6750#section-3.1)).

Closes #5121

### DIFF
--- a/oauth2/oauth2-resource-server/spring-security-oauth2-resource-server.gradle
+++ b/oauth2/oauth2-resource-server/spring-security-oauth2-resource-server.gradle
@@ -1,0 +1,13 @@
+apply plugin: 'io.spring.convention.spring-module'
+
+dependencies {
+	compile project(':spring-security-core')
+	compile project(':spring-security-oauth2-core')
+	compile project(':spring-security-web')
+	compile springCoreDependency
+	compile 'com.nimbusds:oauth2-oidc-sdk'
+
+	optional project(':spring-security-oauth2-jose')
+
+	provided 'javax.servlet:javax.servlet-api'
+}
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/BearerTokenAuthenticationException.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/BearerTokenAuthenticationException.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2002-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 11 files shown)


jgrandja:
@vpavic Thanks so much for submitting this PR! 
Can you please submit this PR to [spring-security-oauth2-resource-server](https://github.com/jzheaux/spring-security-oauth2-resource-server) repo instead.
This will be the repo where all the Resource Server features will be added until it's ready to be submitted as a PR to the Spring Security repo.
Thanks!

vpavic:
I've resubmitted this as jzheaux/spring-security-oauth2-resource-server#3. Can we close this one as duplicate?

rwinch:
Yes. Let's close it here

PR #5123, opened 2018-03-16
ReactorContextTestExecutionListener should obtain SecurityContext on TestExecutionListenerThread

OlegDokuka:
# The Problem 

During some playing around testing of an application, especially during verification of WebSocket API, I have found that `@WithMockUser` annotation does not work as it expected.

# What I have found

Since we do not have something similar to `TestWebClient` but for verification of the WebSocket API, we need to use `WebSocketClient` (for now). It means that we need to run spring boot app with `@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT)` configuration. In turn, it means that execution of the method annotated with, for instance, JUnit's annotation `@Test` will be in a slightly different thread than the execution of `WebSocketHandler`, which will be, for example for Netty, in `reactor-http-nio-blah-blah-blah`.

# It results in

Since `TestSecurityContextHolder` is based on thread local, it means that during execution of `org.springframework.security.test.context.support.ReactorContextTestExecutionListener.DelegateTestExecutionListener.SecuritySubContext#currentContext` mocked `SecurityContext` will not be available in `TestSecurityContextHolder`, since execution of `org.springframework.security.test.context.support.WithSecurityContextTestExecutionListener#beforeTestMethod` (or simply setting `SecurityContext` to `TestSecurityContextHolder`) is on the same thread as execution of test method and in another toward handling of web-socket connection.

### DIFF
--- a/test/src/main/java/org/springframework/security/test/context/support/ReactorContextTestExecutionListener.java
+++ b/test/src/main/java/org/springframework/security/test/context/support/ReactorContextTestExecutionListener.java
@@ -19,6 +19,7 @@ package org.springframework.security.test.context.support;
 import org.reactivestreams.Subscription;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.context.ReactiveSecurityContextHolder;
+import org.springframework.security.core.context.SecurityContext;
 import org.springframework.security.test.context.TestSecurityContextHolder;
 import org.springframework.test.context.TestContext;
 import org.springframework.test.context.TestExecutionListener;
@@ -54,7 +55,8 @@ public class ReactorContextTestExecutionListener
 	private static class DelegateTestExecutionListener extends AbstractTestExecutionListener {
 		@Override
 		public void beforeTestMethod(TestContext testContext) throws Exception {
-			Hooks.onLastOperator(Operators.lift((s, sub) -> new SecuritySubContext<>(sub)));
+			SecurityContext securityContext = TestSecurityContextHolder.getContext();
+			Hooks.onLastOperator(Operators.lift((s, sub) -> new SecuritySubContext<>(sub, securityContext)));
 		}
 
 		@Override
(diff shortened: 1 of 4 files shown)


OlegDokuka:
Ping :)

test/src/main/java/org/springframework/security/test/context/support/ReactorContextTestExecutionListener.java
```diff
@@ -54,7 +54,8 @@ private static TestExecutionListener createDelegate() {
 	private static class DelegateTestExecutionListener extends AbstractTestExecutionListener {
 		@Override
 		public void beforeTestMethod(TestContext testContext) throws Exception {
-			Hooks.onLastOperator(Operators.lift((s, sub) -> new SecuritySubContext<>(sub)));
+			Authentication authentication = TestSecurityContextHolder.getContext().getAuthentication();
```
rwinch: Instead of obtaining the Authentication, please use the SecurityContext so that the original SecurityContext is used (this ensures that if a custom SecurityContext was established it is used later on as well)

rwinch: Please also add a test for custom SecurityContext being established

OlegDokuka: Sure, give me an hour for that

OlegDokuka: Regarding the second comment, do you mean the case when `beforeTestMethod` is called on another thread?

OlegDokuka: @rwinch

OlegDokuka: Or you mean the end to end test case? Because I have provided a case when `beforeTestMethod` is executed on main thread and verification is being executed on ForJoinPool thread

OlegDokuka:
@rwinch, I have replaced `Authentication` final field with `SecurityContext` and provided a few additional integration tests with `TestWebClient` call from the different thread with the mutated mock user. I hope that I got your point correctly :)

OlegDokuka:
ping :)

rwinch:
Thanks for the updated PR! This is now merged into master

rwinch:
This is also back ported to 5.0.x

OlegDokuka:
wooohooo :) Thanks :)

PR #5152, opened 2018-03-22
Close quoted expected path in log when matching

markhobson:
When a request fails to match `AntPathRequestMatcher.matches` logs the actual and expected requests. The quoted expected request is missing the trailing quote. This PR fixes that.


rwinch:
Thanks for the Pull Request! This is now merged into master :)

PR #5156, opened 2018-03-26
XsdDocumentedTests groovy->java

jzheaux:
Groovy has more extensive support for Xml parsing via XmlSlurper.
To replace it, this conversion also introduces a SAX wrapper,
NicerXmlParser, and a companion Node wrapper, NicerNode, that
allowed for less modification of the converted tests.

Issue: gh-4939


config/src/test/java/org/springframework/security/config/doc/NicerNode.java
```diff
@@ -0,0 +1,54 @@
...
+package org.springframework.security.config.doc;
+
+import org.w3c.dom.Node;
+
+import java.util.stream.Stream;
+
+/**
+ * @author Josh Cummings
+ */
+public interface NicerNode {
```
rwinch: I'd avoid interfaces unless we need multiple implementations or if the API is made accessible to users. For those reasons, `NicerNode` (along with the Nicer* interfaces) should be removed

jzheaux: I thought you might say that. :) No problem, will do.

rwinch:
Thanks for the PR! This is merged via 0c0abea3ada05ee4ccea89c8d85ddf7c79685538 I also applied a little polish via 234c20eb3070b26dda36fbf598520417f959d627

jzheaux:
Looks good @rwinch

PR #5197, opened 2018-04-03
Fix @since for MockEventListener

izeye:
`MockEventListener` looks added since 5.0.2.RELEASE although it can be removed as it's a class for internal tests.

See 49e5b15ce28440fc8b9135217ba0c435a25ff543


rwinch:

Issue #4332, opened 2017-05-11
ProviderManager dead loop with AuthenticationManagerBuilder、AuthenticationManagerDelegator


yamingd:
### Summary

trying spring cloud security oauth2, but found dead loop and finally caused stack overflow;

```
@Override
    @Bean
    public AuthenticationManager authenticationManagerBean() throws Exception {
        return super.authenticationManagerBean();
    }

```

```
@Configuration
@EnableAuthorizationServer
public class OAuthServerConfiguration extends AuthorizationServerConfigurerAdapter {

    @Autowired
    @Qualifier("authenticationManagerBean")
    private AuthenticationManager authenticationManager;

    @Autowired
    private DataSource dataSource;

    @Autowired
    private BCryptPasswordEncoder passwordEncoder;


    /**
     * The OAuth2 tokens are defined in the datasource defined in the
     * <code>authenticationManager-server.yml</code> file stored in the Spring Cloud config
     * github repository.
     *
     * @return
     */
    @Bean
    public JdbcTokenStore tokenStore() {
        return new JdbcTokenStore(dataSource);
    }

    @Bean
    protected AuthorizationCodeServices authorizationCodeServices() {
        return new JdbcAuthorizationCodeServices(dataSource);
    }

    @Override
    public void configure(AuthorizationServerSecurityConfigurer oauthServer)
            throws Exception {

        oauthServer.passwordEncoder(passwordEncoder)
                    .tokenKeyAccess("permitAll()")
                    .checkTokenAccess("isAuthenticated()")
                    .allowFormAuthenticationForClients();

    }

    /**
     * We set our authorization storage feature specifying that we would use the
     * JDBC store for token and authorization code storage.<br>
     * <br>
     *
     * We also attach the {@link AuthenticationManager} so that password grants
     * can be processed.
     */
    @Override
    public void configure(AuthorizationServerEndpointsConfigurer endpoints)
            throws Exception {

        endpoints.authorizationCodeServices(authorizationCodeServices())
                .authenticationManager(authenticationManager)
                .tokenStore(tokenStore())
                .approvalStoreDisabled()
                .allowedTokenEndpointRequestMethods(HttpMethod.GET, HttpMethod.POST);
    }

    /**
     * Setup the client application which attempts to get access to user's
     * account after user permission.
     */
    @Override
    public void configure(ClientDetailsServiceConfigurer clients)
            throws Exception {

        clients.jdbc(dataSource)
[...]

PR #5218, opened 2018-04-07
Delete the providermanager.authenticate method, when the result is the parent validation of the air conditioner.

Maijh97:
When ProviderManager.authenticate () is validated, if result is empty, the call parent is tried, if the instance WebSecurityConfigurerAdapter.authenticationManagerBean situation is the result of the validation into the dead cycle and the abnormal java.lang.StackOverflowError:null.


eleftherias:
@Maijh97 Can you please provide a test or sample that reproduces this issue?  
I have found a similar issue that may be related to this. Please take a look at gh-4332 and let me know if it helps.

Issue #5219, opened 2018-04-08
Genereate csrf token error in concurrent environment


someoneYYYY:
### Summary
We use spring-security csrf token mechanism to generate csrf token , but occurs some error in concurrent environment , we get csrf token by this way
```
 CsrfToken token = (CsrfToken)request.getAttribute(CsrfToken.class.getName());
 String sessionId = RequestContextHolder.currentRequestAttributes().getSessionId();
 CsrfTokenDTO csrfTokenDTO = new CsrfTokenDTO();
 csrfTokenDTO.setSessionId(sessionId);
 csrfTokenDTO.setCsrfToken(token.getToken());
```
If we send two requests at same time ,we will get two csrf token response from server, but the first return  csrf token  is invalid, in other word, one session have two csrf token, one of them is invalid. just like below

```json
{"csrfToken":"7e5fc5e2-084e-4042-bed4-d133fe910d26","sessionId":"341AB97535817BF0D04507E65FC9CCA7","description":"Please add Header 'X-CSRF-TOKEN', and cookie '' for all Post request","sessionName":""}
```
```json
{"csrfToken":"7e5fc5e2-7552e-4042-bed4-d133fe910d26","sessionId":"341AB97535817BF0D04507E65FC9CCA7","description":"Please add Header 'X-CSRF-TOKEN', and cookie '' for all Post request","sessionName":""}
```
we thought this behavior cause by this reason , there is no have `synchronized `in `CsrfFilter`

```
@Override
	protected void doFilterInternal(HttpServletRequest request,
			HttpServletResponse response, FilterChain filterChain)
					throws ServletException, IOException {
		request.setAttribute(HttpServletResponse.class.getName(), response);
               // need add synchronized avoid generate two csrf token
		CsrfToken csrfToken = this.tokenRepository.loadToken(request);
		final boolean missingToken = csrfToken == null;
		if (missingToken) {
			csrfToken = this.tokenRepository.generateToken(request);
			this.tokenRepository.saveToken(csrfToken, request, response);
		}
		request.setAttribute(CsrfToken.class.getName(), csrfToken);
		request.setAttribute(csrfToken.getParameterName(), csrfToken);

		if (!this.requireCsrfProtectionMatcher.matches(request)) {
			filterChain.doFilter(request, response);
			return;
		}
}
```
### Actual Behavior
A session corresponds to two csrf token

### Expected Behavior
A session corresponds to a csrf token

### Configuration

we use csrf mechanism  by this way
```
[...]

PR #5221, opened 2018-04-08
Fix genereate csrf token error in concurrent environment

someoneYYYY:
Generate invalid csrf token in concurrent environment ,can't pass csrf token verify

Add synchronized avoid two thread generate csrf-token at same time in CsrfFilter

Fixes: gh-5219


rwinch:
Unfortunately you cannot use synchronization. Please seem my comments at https://github.com/spring-projects/spring-security/issues/5219

Issue #5083, opened 2018-03-07
Add CookieServerCsrfTokenRepository


edeandrea:
A cookie implementation of `ServerCsrfTokenRepository` (like `CookieCsrfTokenRepository`) is missing. In this implementation it would be nice to allow the setting of the domain as well.

edeandrea:
I added [PR 5245](https://github.com/spring-projects/spring-security/pull/5245) for this.

rwinch:
Thanks for creating this ticket and the PR! This is now merged into master. Please see the referenced commits for the polish that I applied

PR #5245, opened 2018-04-18
Add CookieServerCsrfTokenRepository

edeandrea:
A cookie implementation of ServerCsrfTokenRepository (like CookieCsrfTokenRepository)
is missing. In this implementation it would be nice to allow the setting of the domain as well.

Fixes gh-5083


rwinch:
Thanks for the PR! This is now merged into master with some polish. See #5083 for the additional commits

Issue #5242, opened 2018-04-17
There is no example.


blusky10:
Doc page : 
https://docs.spring.io/spring-security/site/docs/5.0.4.RELEASE/reference/htmlsingle/#jc-httpsecurity

Section : 5.2 HttpSecurity
......
Additionally, it is not quite so obvious we are using Spring Security which helps to prevent information leaks. For example:
...

The sentence ends with For Example. But there is no example.

PR #5247, opened 2018-04-19
Remove Default URLs and Parameters Commentary

jzheaux:
This paragraph was originally there to explain why Spring Security was
moving from the old default URLs and parameters to new ones.

Now that the defaults are the same in XML and well as Java Config, the
explanation is now no longer relevant.

Fixes: gh-5242


jzheaux:
Here is the commit where the paragraph was removed: https://github.com/spring-projects/spring-security/commit/c67ff42b8abe124b7956896c78e9aac896fd79d9

PR #5269, opened 2018-04-29
Fix 'attributes' deprecation warning in spring-security-docs-guides

izeye:
Build this project shows the following warning:

```
Configure project :spring-security-docs-guides 
Attributes found in options. Existing attributes will be replaced due to assignment. Please use 'attributes' method instead as current behaviour will be removed in future
```

This PR fixes the warning by applying the suggested way in the warning message.

See https://github.com/asciidoctor/asciidoctor-gradle-plugin#options-attributes


rwinch:
Thanks for the PR! This is now merged into master and backported into 5.0.x via #5308

PR #5298, opened 2018-05-04
Fix incorrect explanation for customizing query on JdbcDaoImpl

kazuki43zoo:
I've fixed an incorrect explanation on the `JdbcDaoImpl`.
Probably, we should be mapped back to same column **positions** rather than same column names as in the default query when a custom query will be use.


rwinch:
Thanks for the PR! This is now merged into master and back ported to 5.0.x and 4.2.x

Issue #5064, opened 2018-03-02
OAuth2AuthorizedClientService must handle anonymous Principal


jgrandja:
At the completion of the  `authorization_code` grant flow, the `OAuth2AuthorizedClientService.saveAuthorizedClient()` is called to save the `OAuth2AuthorizedClient` to the backing store. One of the parameters to `saveAuthorizedClient()` is `Authentication` representing the current `Principal`. There may be a case where the current `Principal` is anonymous, which would result in `OAuth2AuthorizedClient`'s being over-written by a subsequent anonymous user that completed an  `authorization_code` grant flow.

Therefore, the `OAuth2AuthorizedClientService` must be capable of identifying if the current `Principal` is anonymous, and if true, than it should save the `OAuth2AuthorizedClient` to the `HttpSession` regardless of the backing store implementation.

Currently, the only implementation of `OAuth2AuthorizedClientService` is `InMemoryOAuth2AuthorizedClientService`.

PR #5307, opened 2018-05-04
InMemoryOAuth2AuthorizedClientService handles anonymous Principal

jgrandja:
Fixes gh-5064


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryOAuth2AuthorizedClientService.java
```diff
@@ -38,6 +40,7 @@
 public final class InMemoryOAuth2AuthorizedClientService implements OAuth2AuthorizedClientService {
 	private final Map<String, OAuth2AuthorizedClient> authorizedClients = new ConcurrentHashMap<>();
 	private final ClientRegistrationRepository clientRegistrationRepository;
+	private final OAuth2AuthorizedClientService httpSessionAuthorizedClientService;
```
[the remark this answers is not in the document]
jgrandja: Yeah you have a point. Let me give this some thought. Maybe a delegation-based strategy would work here.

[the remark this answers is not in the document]
rwinch: I'm thinking perhaps a new level of abstraction used by the filter that also has the session as an argument to avoid the thread local too.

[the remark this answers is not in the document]
jgrandja: I wasn't that much in favour of using `RequestContextHolder` so I'll look your suggested alternative.

jgrandja:
@rwinch Regarding your comment:

> I'm thinking perhaps a new level of abstraction used by the filter that also has the session as an argument to avoid the thread local too

I decided to stick with an `HttpSession` based implementation of `OAuth2AuthorizedClientService` which requires the use of `RequestContextHolder`.

I didn't feel adding a new abstraction that is similar to `OAuth2AuthorizedClientService` and accepts either `HttpServletRequest` or `HttpSession` was necessary.

Maybe I'm not seeing the concerns you have with the use of the `ThreadLocal`. Can you provide me more details around your concerns?

Either way, take a look at the updates in this PR and let me know your thoughts.

rwinch:
I think that the Filter should only have one repository API rather than toggling based upon if the user is logged in or not. Putting the logic for authenticated or not in the Filter is not ideal because it puts too many responsibilities in the Filter which means it is less flexible.

Instead we should introduce the new API and it performs the delegation.

Another concern is this relies on the RequestContextFilter to be setup. This is not automatically done unless you are in a Spring Boot application, so it will prove to be more setup instructions for users that are not using Boot. It also gets complicated because we need to worry about the RequestContextFilter being in the correct order (i.e. before Spring Security's Filters).

jgrandja:
> Instead we should introduce the new API and it performs the delegation.

I'm not sure I completely understand but I think this is what you are saying...

Introduce a new API that looks like `OAuth2AuthorizedClientService` but adds the additional parameter `HttpSession` or `HttpServletRequest` to each operation.

Then the implementation of this new API performs the delegation to either the `OAuth2AuthorizedClientService` for authenticated access OR if access is unauthenticated than loads/save from `HttpSession`.

Is my understanding correct?

> Another concern is this relies on the RequestContextFilter to be setup. This is not automatically done unless you are in a Spring Boot application

Oh yeah that's right. And agreed that would complicate the setup which is not good.

rwinch:
Yes that sounds accurate. We should also support the response for writing it to cookies if the user decides that is what they want to do.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizedClientRepository.java
```diff
@@ -0,0 +1,86 @@
...
+
+	@Override
+	public void removeAuthorizedClient(String clientRegistrationId, String principalName,
+										HttpServletRequest request, HttpServletResponse response) {
+		Assert.hasText(clientRegistrationId, "clientRegistrationId cannot be empty");
+		Assert.notNull(request, "request cannot be null");
+		Map<String, OAuth2AuthorizedClient> authorizedClients = this.getAuthorizedClients(request);
+		authorizedClients.remove(clientRegistrationId);
+		if (!authorizedClients.isEmpty()) {
+			request.getSession().setAttribute(this.sessionAttributeName, authorizedClients);
```
rwinch: If there was no HttpSession prior to invoking removeAuthorizedClient, we should not create a new session for removal

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationCodeGrantFilter.java
```diff
@@ -106,17 +104,17 @@
 	 * Constructs an {@code OAuth2AuthorizationCodeGrantFilter} using the provided parameters.
 	 *
 	 * @param clientRegistrationRepository the repository of client registrations
-	 * @param authorizedClientService the authorized client service
+	 * @param authorizedClientRepository the authorized client repository
 	 * @param authenticationManager the authentication manager
 	 */
 	public OAuth2AuthorizationCodeGrantFilter(ClientRegistrationRepository clientRegistrationRepository,
-												OAuth2AuthorizedClientService authorizedClientService,
```
rwinch: This is a breaking change. We should add a new constructor. The old constructor can be deprecated and just create an instance of AuthenticatedPrincipalOAuth2AuthorizedClientRepository with the provided OAuth2AuthorizedClientService

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/method/annotation/OAuth2AuthorizedClientArgumentResolver.java
```diff
@@ -54,16 +56,16 @@
...
 	/**
 	 * Constructs an {@code OAuth2AuthorizedClientArgumentResolver} using the provided parameters.
 	 *
-	 * @param authorizedClientService the authorized client service
+	 * @param authorizedClientRepository the authorized client repository
 	 */
-	public OAuth2AuthorizedClientArgumentResolver(OAuth2AuthorizedClientService authorizedClientService) {
-		Assert.notNull(authorizedClientService, "authorizedClientService cannot be null");
-		this.authorizedClientService = authorizedClientService;
+	public OAuth2AuthorizedClientArgumentResolver(OAuth2AuthorizedClientRepository authorizedClientRepository) {
```
rwinch: This is a breaking change. We should add a new constructor. The old constructor can be deprecated and just create an instance of AuthenticatedPrincipalOAuth2AuthorizedClientRepository with the provided OAuth2AuthorizedClientService

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/AuthenticatedPrincipalOAuth2AuthorizedClientRepositoryTests.java
```diff
@@ -0,0 +1,126 @@
...
+	@Before
+	public void setup() {
+		this.authorizedClientService = mock(OAuth2AuthorizedClientService.class);
+		this.httpSessionAuthorizedClientRepository = mock(OAuth2AuthorizedClientRepository.class);
+		this.authorizedClientRepository = new AuthenticatedPrincipalOAuth2AuthorizedClientRepository(this.authorizedClientService);
+		this.authorizedClientRepository.httpSessionAuthorizedClientRepository = this.httpSessionAuthorizedClientRepository;
+		this.request = new MockHttpServletRequest();
+		this.response = new MockHttpServletResponse();
+	}
+
```
rwinch: An `@After` that invokes `SecurityContextHolder.clearContext()` needs to be added. If you don't then tests will randomly fail depending on ordering due to the fact that it leaks memory. It can also cause memory leaks due to classloader being leaked on the underlying SecurityContext class stored in the threadlocal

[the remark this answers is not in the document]
jgrandja: Yes, I missed that.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/AuthenticatedPrincipalOAuth2AuthorizedClientRepository.java
```diff
@@ -0,0 +1,93 @@
...
+	 * @param authorizedClientService the authorized client service
+	 */
+	public AuthenticatedPrincipalOAuth2AuthorizedClientRepository(OAuth2AuthorizedClientService authorizedClientService) {
+		Assert.notNull(authorizedClientService, "authorizedClientService cannot be null");
+		this.authorizedClientService = authorizedClientService;
+	}
+
+	@Override
+	public <T extends OAuth2AuthorizedClient> T loadAuthorizedClient(String clientRegistrationId, String principalName,
+																		HttpServletRequest request) {
```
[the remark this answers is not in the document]
jgrandja: The reason I'm not asserting on the method arguments in `AuthenticatedPrincipalOAuth2AuthorizedClientRepository` is because the pre-checks are performed in the delegated backing implementation. As well, `loadAuthorizedClient` should never be called with a `null` `principalName` as this would be an error by the caller. I think the current logic makes sense so I'm going to leave this one.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/AuthenticatedPrincipalOAuth2AuthorizedClientRepository.java
```diff
@@ -0,0 +1,93 @@
...
+		if (this.isCurrentPrincipalAuthenticated()) {
+			this.authorizedClientService.saveAuthorizedClient(authorizedClient, principal);
+		} else {
+			this.httpSessionAuthorizedClientRepository.saveAuthorizedClient(authorizedClient, principal, request, response);
+		}
+	}
+
+	@Override
+	public void removeAuthorizedClient(String clientRegistrationId, String principalName,
+										HttpServletRequest request, HttpServletResponse response) {
```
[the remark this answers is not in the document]
jgrandja: `Authentication`  is not passed into `removeAuthorizedClient`, only `principalName`

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationCodeGrantFilter.java
jgrandja: This feature is targeted for 5.1 so we're good with the change.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/method/annotation/OAuth2AuthorizedClientArgumentResolver.java
jgrandja: This feature is targeted for 5.1 so we're good with the change.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationCodeGrantFilter.java
rwinch: Thanks for pointing that out. That was a miss on my side that the code has never reached GA.

jgrandja:
Thanks for the feedback @rwinch. I've applied all the updates and is ready for review.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizedClientRepository.java
```diff
@@ -0,0 +1,89 @@
...
+		request.getSession().setAttribute(this.sessionAttributeName, authorizedClients);
+	}
+
+	@Override
+	public void removeAuthorizedClient(String clientRegistrationId, Authentication principal,
+										HttpServletRequest request, HttpServletResponse response) {
+		Assert.hasText(clientRegistrationId, "clientRegistrationId cannot be empty");
+		Assert.notNull(request, "request cannot be null");
+		Map<String, OAuth2AuthorizedClient> authorizedClients = this.getAuthorizedClients(request);
+		if (!authorizedClients.isEmpty()) {
```
rwinch: Please add a test to ensure that HttpSession is not created if one did not exist

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizedClientRepository.java
```diff
@@ -0,0 +1,89 @@
...
+			if (authorizedClients.remove(clientRegistrationId) != null) {
+				if (!authorizedClients.isEmpty()) {
+					request.getSession().setAttribute(this.sessionAttributeName, authorizedClients);
```
rwinch: Please add a test for this use case

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/AuthenticatedPrincipalOAuth2AuthorizedClientRepositoryTests.java
```diff
@@ -0,0 +1,142 @@
...
+	@Test
+	public void removeAuthorizedClientWhenAnonymousPrincipalThenRemoveFromAnonymousRepository() {
+		this.setupAnonymousPrincipal();
+		Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+		this.authorizedClientRepository.removeAuthorizedClient(this.registrationId, authentication, this.request, this.response);
+		verify(this.anonymousAuthorizedClientRepository).removeAuthorizedClient(this.registrationId, authentication, this.request, this.response);
+	}
+
+	private void setupAuthenticatedPrincipal() {
```
rwinch: We should not set the SecurityContextHolder. Instead return the Authentication and use it directly.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/AuthenticatedPrincipalOAuth2AuthorizedClientRepositoryTests.java
```diff
@@ -0,0 +1,142 @@
...
+		TestingAuthenticationToken authentication = new TestingAuthenticationToken(this.principalName, "password");
+		authentication.setAuthenticated(true);
+		SecurityContext securityContext = SecurityContextHolder.createEmptyContext();
+		securityContext.setAuthentication(authentication);
+		SecurityContextHolder.setContext(securityContext);
+	}
+
+	private void setupAnonymousPrincipal() {
```
rwinch: We should not set the SecurityContextHolder. Instead return the Authentication and use it directly.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/AuthenticatedPrincipalOAuth2AuthorizedClientRepositoryTests.java
```diff
@@ -0,0 +1,142 @@
...
+		this.authorizedClientService = mock(OAuth2AuthorizedClientService.class);
+		this.anonymousAuthorizedClientRepository = mock(OAuth2AuthorizedClientRepository.class);
+		this.authorizedClientRepository = new AuthenticatedPrincipalOAuth2AuthorizedClientRepository(this.authorizedClientService);
+		this.authorizedClientRepository.setAnonymousAuthorizedClientRepository(this.anonymousAuthorizedClientRepository);
+		this.request = new MockHttpServletRequest();
+		this.response = new MockHttpServletResponse();
+	}
+
+	@After
+	public void cleanup() {
```
rwinch: After cleaning up the setup*Principal methods below, this can be removed

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/method/annotation/OAuth2AuthorizedClientArgumentResolver.java
```diff
@@ -101,12 +103,12 @@ public Object resolveArgument(MethodParameter parameter,
 		if (principal == null) {
```
rwinch: If anonymous Authentication is disabled, this might be null so we shouldn't throw an Exception here.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/result/method/annotation/OAuth2AuthorizedClientArgumentResolver.java
```diff
@@ -92,7 +92,7 @@ public boolean supportsParameter(MethodParameter parameter) {
 							clientRegistrationId.flatMap(id -> Mono.error(new IllegalStateException(
 									"Unable to resolve the Authorized Client with registration identifier \""
 											+ id
-											+ "\". An \"authenticated\" or \"unauthenticated\" session is required. To allow for unauthenticated access, ensure ServerHttpSecurity.anonymous() is configured."))))
+											+ "\". An \"authenticated\" or \"anonymous\" request is required. To allow for anonymous access, ensure ServerHttpSecurity.anonymous() is configured."))))
```
rwinch: I don't think this should be changed since on the reactive side we don't support anonymous users.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationCodeGrantFilterTests.java
```diff
@@ -285,6 +291,47 @@ public void doFilterWhenAuthorizationResponseSuccessHasSavedRequestThenRedirecte
 		assertThat(response.getRedirectedUrl()).isEqualTo("http://localhost/saved-request");
 	}
 
+	@Test
+	public void doFilterWhenAuthorizationResponseSuccessAndAnonymousAccessThenAuthorizedClientSavedToHttpSession() throws Exception {
+		AnonymousAuthenticationToken anonymousPrincipal =
+				new AnonymousAuthenticationToken("key-1234", "anonymousUser", AuthorityUtils.createAuthorityList("ROLE_ANONYMOUS"));
+		SecurityContext securityContext = SecurityContextHolder.createEmptyContext();
+		securityContext.setAuthentication(anonymousPrincipal);
+		SecurityContextHolder.setContext(securityContext);
```
rwinch: This class is missing `@After SecurityContextHolder.clearContext()`

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/method/annotation/OAuth2AuthorizedClientArgumentResolverTests.java
```diff
@@ -104,7 +110,7 @@ public void resolveArgumentWhenRegistrationIdEmptyAndOAuth2AuthenticationThenRes
 		securityContext.setAuthentication(authentication);
 		SecurityContextHolder.setContext(securityContext);
 		MethodParameter methodParameter = this.getMethodParameter("registrationIdEmpty", OAuth2AuthorizedClient.class);
-		this.argumentResolver.resolveArgument(methodParameter, null, null, null);
+		this.argumentResolver.resolveArgument(methodParameter, null, new ServletWebRequest(this.request), null);
 	}
```
rwinch: This class is missing `@After SecurityContextHolder.clearContext()`

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/result/method/annotation/OAuth2AuthorizedClientArgumentResolverTests.java
```diff
@@ -113,8 +113,8 @@ public void resolveArgumentWhenParameterTypeOAuth2AuthorizedClientAndCurrentAuth
 		assertThatThrownBy(() -> resolveArgument(methodParameter))
 				.isInstanceOf(IllegalStateException.class)
 				.hasMessage("Unable to resolve the Authorized Client with registration identifier \"client1\". " +
-						"An \"authenticated\" or \"unauthenticated\" session is required. " +
```
rwinch: I don't think this should be changed since it does not yet support anonymous

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/method/annotation/OAuth2AuthorizedClientArgumentResolver.java
jgrandja: If the user is using `RegisteredOAuth2AuthorizedClient` and has configured `HttpSecurity.anonymous().disabled()` than IMO this is a mis-configuration and the `IllegalStateException` signals this to the user. Does this make sense?

rwinch: Why is this a misconfiguration? Is there any reason this should not work if the principal is null?

jgrandja: `OAuth2AuthorizedClientService` requires `Principal.getName()` in order to load/save an `OAuth2AuthorizedClient` so if `HttpSecurity.anonymous().disabled()` is configured than `Authentication` will be `null` and `RegisteredOAuth2AuthorizedClient` can either return null (which you are suggesting) or throw the exception as current behaviour. 

If we don't throw the exception to signal to the user (and return `null` instead) than they might think there is a bug because it's not resolving the `OAuth2AuthorizedClient`. At least with the exception and detailed message, the user will than know to either `HttpSecurity.anonymous().enable()` or remove the use of `@RegisteredOAuth2AuthorizedClient`.

PR #5312, opened 2018-05-05
Suppress deprecation warnings in spring-security-crypto

izeye:
This PR suppresses deprecation warnings in `spring-security-crypto`.


rwinch:

Issue #5315, opened 2018-05-07
CookieServerCsrfTokenRepository fails with IllegalArgumentException when cookie value is null or empty


edeandrea:
The `CookieServerCsrfTokenRepository` fails with an `IllegalArgumentException` when a cookie is present but the value is `null` or empty.

rwinch:
Thanks for creating the ticket and the PR! This is now merged into master

PR #5316, opened 2018-05-07
CookieServerCsrfTokenRepository fails when cookie is null/empty

edeandrea:
The CookieServerCsrfTokenRepository fails with an IllegalArgumentException
 when a cookie is present but the value is null or empty.

Fixes gh-5315


rwinch:
Thanks for the PR @edeandrea! This is now merged into master

PR #4997, opened 2018-02-04
Update FilterSecurityInterceptor.java

json20080301:
if  observeOncePerRequest==false ,DO NOT 
	"fi.getRequest().setAttribute(FILTER_APPLIED, Boolean.TRUE);"	

// first time this request being called, so perform security checking
			if (fi.getRequest() != null && observeOncePerRequest) {
				fi.getRequest().setAttribute(FILTER_APPLIED, Boolean.TRUE);
			}

PR #5366, opened 2018-05-21
FilterSecurityInterceptor once per request set attr

json20080301:
Only set the attribute if once per request is true


jzheaux:
Thanks, @json20080301! Will you please also add a test like this one here: https://github.com/spring-projects/spring-security/commit/fcf967687bc1132747336fba98c227bad6c0a401

You might just be able to cherry-pick that commit.

eleftherias:
Given that gh-4997 ensures that this change is already part of the `master` branch and the need to remain passive on `4.2.x` versions, we do not need to merge this into the `4.2.x` branch.

Issue #5337, opened 2018-05-11
Automatically add CsrfServerLogoutHandler if csrf enabled


rwinch:
### Summary

The configuration DSL should automatically add CsrfServerLogoutHandler if csrf is enabled

See #4840

rwinch:
> Should we build a DelegatingServerLogoutHandler that can delegate to multiple ServerLogoutHandlers?

@edeandrea Thanks for reaching out! Yes I think that would be the ideal way of handling it.

edeandrea:
Seems the `DelegatingServerLogoutHandler` ticket is already raised in #4839

edeandrea:
I've added #5379 for this.

PR #5379, opened 2018-05-24
Automatically add CsrfServerLogoutHandler if csrf enabled

edeandrea:
The configuration DSL should automatically add `CsrfServerLogoutHandler` if csrf is enabled

Fixes gh-5337

### DIFF
--- a/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
@@ -92,6 +92,7 @@ import org.springframework.security.web.server.authentication.ServerAuthenticati
 import org.springframework.security.web.server.authentication.ServerAuthenticationSuccessHandler;
 import org.springframework.security.web.server.authentication.ServerFormLoginAuthenticationConverter;
 import org.springframework.security.web.server.authentication.ServerHttpBasicAuthenticationConverter;
+import org.springframework.security.web.server.authentication.logout.DelegatingServerLogoutHandler;
 import org.springframework.security.web.server.authentication.logout.LogoutWebFilter;
 import org.springframework.security.web.server.authentication.logout.ServerLogoutHandler;
 import org.springframework.security.web.server.authentication.logout.ServerLogoutSuccessHandler;
@@ -2360,7 +2361,20 @@ public class ServerHttpSecurity {
 		 * @see #logoutHandler(ServerLogoutHandler)
 		 */
 		public LogoutSpec additionalLogoutHandler(ServerLogoutHandler logoutHandler) {
-			this.logoutWebFilter.addLogoutHandler(logoutHandler);
+			ServerLogoutHandler currentLogoutHandler = this.logoutWebFilter.getLogoutHandler();
+
+			if (currentLogoutHandler != null) {
+				if (currentLogoutHandler instanceof DelegatingServerLogoutHandler) {
+					((DelegatingServerLogoutHandler) currentLogoutHandler).addDelegate(logoutHandler);
+				}
+				else {
+					logoutHandler(new DelegatingServerLogoutHandler(currentLogoutHandler, logoutHandler));
+				}
+			}
+			else {
+				logoutHandler(logoutHandler);
+			}
+
 			return this;
 		}
 
--- a/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
@@ -106,6 +106,7 @@ import org.springframework.security.web.server.context.ReactorContextWebFilter;
 import org.springframework.security.web.server.context.SecurityContextServerWebExchangeWebFilter;
 import org.springframework.security.web.server.context.ServerSecurityContextRepository;
 import org.springframework.security.web.server.context.WebSessionServerSecurityContextRepository;
+import org.springframework.security.web.server.csrf.CsrfServerLogoutHandler;
 import org.springframework.security.web.server.csrf.CsrfWebFilter;
 import org.springframework.security.web.server.csrf.ServerCsrfTokenRepository;
 import org.springframework.security.web.server.header.CacheControlServerHttpHeadersWriter;
@@ -1600,6 +1601,7 @@ public class ServerHttpSecurity {
 		}
 
 		protected void configure(ServerHttpSecurity http) {
+			http.logout().additionalLogoutHandler(new CsrfServerLogoutHandler(this.filter.getCsrfTokenRepository()));
 			http.addFilterAt(this.filter, SecurityWebFiltersOrder.CSRF);
 		}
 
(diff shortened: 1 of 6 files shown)


rwinch:
In general we do not want the Filter or any components other than configuration to know about multiple logout handlers. Can you rework this request to account for this?

edeandrea:
Hi @rwinch are you saying you would prefer to not have the concept of `additionalLogoutHandler` outside of the `LogoutWebFilter` class?

If so I will take a look and see what I can do to refactor a bit.

edeandrea:
Hi @rwinch I did some cleanup. Let me know if this looks good.

edeandrea:
Hi @rwinch just wanted to see if the re-work I had done on 6/19 was more what you were looking for.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -1508,12 +1511,38 @@ private HeaderSpec() {
...
 		}
 
+		/**
+		 * Configutes an additional logout handler.
+		 * @param logoutHandler The additional handler to use
+		 * @return This instance
+		 * @since 5.1
+		 * @see #logoutHandler(ServerLogoutHandler)
+		 */
+		public LogoutSpec additionalLogoutHandler(ServerLogoutHandler logoutHandler) {
```
rwinch: Let's keep this private for now please.

web/src/main/java/org/springframework/security/web/server/authentication/logout/DelegatingServerLogoutHandler.java
```diff
@@ -35,21 +37,34 @@
...
-		this.delegates = new ArrayList<>(delegates);
+		this.delegates.addAll(delegates);
+	}
+
+	/**
+	 * Adds a delegate {@link ServerLogoutHandler}
+	 *
+	 * @param delegate The delegate to add
+	 */
+	public void addDelegate(ServerLogoutHandler delegate) {
```
rwinch: We want this to be immutable, so this method needs to be removed

web/src/main/java/org/springframework/security/web/server/authentication/logout/LogoutWebFilter.java
```diff
@@ -85,11 +85,24 @@ public void setLogoutSuccessHandler(
...
+	/**
+	 * Sets the {@link ServerLogoutHandler}. The default is {@link SecurityContextServerLogoutHandler}.
+	 * @param logoutHandler The handler to use
+	 */
 	public void setLogoutHandler(ServerLogoutHandler logoutHandler) {
 		Assert.notNull(logoutHandler, "logoutHandler must not be null");
 		this.logoutHandler = logoutHandler;
 	}
 
+	/**
```
rwinch: We should not add a getter as it breaks encapsulation

web/src/main/java/org/springframework/security/web/server/csrf/CsrfWebFilter.java
```diff
@@ -75,6 +76,15 @@ public void setCsrfTokenRepository(
 		this.csrfTokenRepository = csrfTokenRepository;
 	}
 
+	/**
```
rwinch: We should not add a getter as it breaks encapsulation

edeandrea:
Thanks @rwinch I will take a look.

Since there has been some time between when this was submitted and the current _master_ branch, what is the recommended approach for pulling my local branch from the remote upstream.

What I've done in the past is to first sync my local fork's _master_ with the upstream _master_ using the command

```
git fetch upstream && git checkout master && git merge upstream/master && git push
```

I then would merge the changes from my local _master_ into my local branch. Last time I went through this though I ran into lots of issues where all the individual commits were added to the PR.

I think I essentially want to squash all the intermediate commits into my local branch. Whats the best way to do this?

Currently my local git repo has not been synced from the upstream _master_ in a while.

web/src/main/java/org/springframework/security/web/server/authentication/logout/LogoutWebFilter.java
edeandrea: The initial review back in May said that we wanted to keep the idea of `DelegatingServerLogoutHandler` to the configuration DSL. In order to do that the configuration DSL needs to be able to reach into the current state of things (i.e. the `LogoutWebFilter`'s current `ServerLogoutHandler` as well as the currently configured `CsrfWebFilter`'s `CsrfTokenRepository`. I'm unsure of how to accomplish that without exposing a getter for that.

web/src/main/java/org/springframework/security/web/server/authentication/logout/DelegatingServerLogoutHandler.java
edeandrea: If we truly want this to be immutable, then the configuration DSL could potentially be setting a `DelegatingServerLogoutHandler` who has in its collection of delegates other `DelegatingServerLogoutHandler`s (which could, in turn, contain other `DelegatingServerLogoutHandler`s). This method's purpose was to allow the configuration DSL to guard against that.

If you're ok with that situation then sure we can remove this method.

edeandrea:
Thanks @vpavic so the steps would be

1. Sync upstream master to my local master
1. Rebase my PR branch onto local master, resolving any conflicts
1. Push local master onto my origin master
1. Merge local master into local PR branch

Would that be the correct sequence?

vpavic:
Step 4 should be rebase, not merge. So something like this:

```
$ git fetch upstream
$ git checkout master && git merge upstream/master && git push
$ git checkout pr-branch && git rebase master
```

At that point, you resolve the conflicts (if any) and push force to the remote PR branch.

edeandrea:
Thanks @vpavic that's exactly what I needed

edeandrea:
Hi @rwinch just wanted to see if you had a chance to look at my comments above.

web/src/main/java/org/springframework/security/web/server/authentication/logout/DelegatingServerLogoutHandler.java
rwinch: I think the DSL would just track the `ServerLogoutHandler`s If there were multiple it would create a `DelegatingServerLogoutHandler` with all the added `ServerLogoutHandler`s. Otherwise it would just use the single `ServerLogoutHandler`

web/src/main/java/org/springframework/security/web/server/authentication/logout/LogoutWebFilter.java
rwinch: See https://github.com/spring-projects/spring-security/pull/5379#discussion_r219199535

edeandrea:
Hi @rwinch I committed a few changes. Please have a look.

rwinch:
Thanks for the polish. This is now in master b060ec050a5b961cb68945f29995faef55be8f58 . I squashed the commits and applied some polish 45a9c0fd54bc48a5dbbfa2362bec286b32dc68f3.

Issue #4911, opened 2017-12-13
Support Custom Query Parameters within OAuth2AuthorizationRequestUriBuilder


andersonkyle:
### Summary

When using a 3rd party OAuth provider such as Auth0 or Okta it would be very useful to allow developers to hook into the Authorization Request URI [build process](https://github.com/spring-projects/spring-security/blob/9f6af4f3b8e79a8a45369d99862cd3d96f4083ce/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationRequestUriBuilder.java#L38-L52) in order to add custom query parameters.  This is because the providers mentioned above have added additional parameters on top of the specification to allow for more functionality.

### Examples

---

[Auth0](https://auth0.com/docs/api/authentication#authorization-code-grant)

`audience: The unique identifier of the target API you want to access.`

This parameter impacts the format of the access token

---

[Okta](https://developer.okta.com/docs/api/resources/oauth2.html#obtain-an-authorization-grant-from-a-user)

`idp: The Identity provider used to do the authentication.`

This parameter instructs Okta to either use itself as the Identity Provider or to use another identity provider connected to the Okta Authorization Server.

---

A simple hook into the building of the URI should provide sufficient flexibility for these custom parameters and any others that are introduced in the future.

jgrandja:
@andersonkyle Makes sense. We'll aim to add this for 5.1

jgrandja:
@andersonkyle I'm finally circling back to address this requirement.

The simplest solution to allow for custom authorization request parameters would be to allow the user to provide the request parameters via the `ClientRegistration` configuration.

For example, if you would like to provide the `audience` and `idp` in the Authorization Request, you would configure it as follows:

```
spring:
  security:
    oauth2:
      client:
        registration:
          okta:
            client-id: client-id
            client-secret: secret
	    ...
	    ...
        provider:
          okta:
            authorization-uri: https://tenant1.okta.com/oauth2/v2/auth?audience=resource1&idp=https://provider2.com
	   ...
	   ...
```

We would ensure that any request parameters supplied in the `authorization-uri` are preserved when building the Authorization Request.

Would this strategy meet your requirement?

andersonkyle:
Yeah, that should do the trick.  Thanks!

The same flexibility is needed for the token request:

```
{
	"grant_type":"http://auth0.com/oauth/grant-type/password-realm",
	"username": "jdoe@example.com",
	"password": "password",
[...]

PR #5444, opened 2018-06-14
Add support for custom authorization request parameters

jgrandja:
This PR adds support for appending additional request parameters on top of the standard [OAuth 2.0 Authorization Request parameters](https://tools.ietf.org/html/rfc6749#section-4.1.1).

OpenID Connect 1.0 defines [Authentication Request parameters](http://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) on top of the OAuth 2.0 Authorization Request parameters that this feature will be able to support.

Providers may also define non-standard parameters which this feature will be able to support as well.

This PR fixes gh-4911


rwinch:
@jgrandja For record keeping purposes. As discussed offline, we can resolve this by placing query parameters in the authorization URL. I'm marking this as waiting-for-changes as I don't think we can safely close it and that is the closest status we have for pending closure

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationRequestRedirectFilter.java
```diff
@@ -112,14 +100,23 @@ public OAuth2AuthorizationRequestRedirectFilter(ClientRegistrationRepository cli
...
+				clientRegistrationRepository, authorizationRequestBaseUri);
+	}
+
+	/**
+	 * Sets the resolver used for resolving {@link OAuth2AuthorizationRequest}'s.
+	 *
+	 * @since 5.1
+	 * @param authorizationRequestResolver the resolver used for resolving authorization requests
+	 */
+	public final void setAuthorizationRequestResolver(OAuth2AuthorizationRequestResolver authorizationRequestResolver) {
```
rwinch: This should be a separate constructor. The reason is that with a custom OAuth2AuthorizationRequestResolver implementation, the ClientRegistrationRepository should not be required for this class. However, it is required to construct the class at the moment.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java
```diff
@@ -50,6 +56,7 @@
 	private Set<String> scopes;
 	private String state;
 	private Map<String, Object> additionalParameters;
+	private URI authorizationRequestUri;
```
rwinch: Change to String for consistency.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java
```diff
@@ -144,6 +161,64 @@ public static Builder implicit() {
...
+	}
+
+	/**
+	 * Returns a {@code URI} representation for the provided {@code authorizationRequest}.
+	 *
+	 * @since 5.1
+	 * @param authorizationRequest the authorization request
+	 * @return a {@code URI} representation of the authorization request
+	 */
+	public static URI asUri(OAuth2AuthorizationRequest authorizationRequest) {
```
rwinch: This method does not make sense to me. What is it constructing? The OAuth2AuthorizationRequest has a authorizationRequestUri property, so it is illogical to pass in an argument that contains the value that you are producing. I think this logic belongs encapsulated within the resolver and we can open it up later if necessary.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationRequestRedirectFilter.java
jgrandja: Good catch. Makes sense.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java
jgrandja: If we store this as a `String` than is it the encoded or decoded version? The `OAuth2AuthorizationRequestRedirectFilter` and `OAuth2AuthorizationRequestRedirectWebFilter` will need to know this before it initiates the redirect.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java
jgrandja: > The OAuth2AuthorizationRequest has a authorizationRequestUri property, so it is illogical to pass in an argument that contains the value that you are producing

Yeah good point

> I think this logic belongs encapsulated within the resolver

Why not keep this logic in `OAuth2AuthorizationRequest` encapsulated. It provides the default `authorizationRequestUri` in `build()` if the user has not supplied it.

```
			authorizationRequest.authorizationRequestUri =
					this.authorizationRequestUri != null ?
						this.authorizationRequestUri : asUri(authorizationRequest);

```

jgrandja:
@rwinch I've applied all the updates from your recent feedback. However, I decided to leave the logic for building the `authorizationRequestUri` in `OAuth2AuthorizationRequest` (encapsulated) instead of the `DefaultOAuth2AuthorizationRequestResolver`. See me previous comment for my reasoning.

Also, it almost slipped by me as we need to add the ability to configure a custom `OAuth2AuthorizationRequestResolver` in `OAuth2LoginConfigurer` and `OAuth2ClientConfigurer`. Would you prefer a new ticket logged or bring it into this PR?

rwinch:
This is now merged in via 779597af2a6ed777707f08ae8106818e0b8e299e

Issue #5457, opened 2018-06-25
Submitting alg=none should not discuss class


jzheaux:
### Summary

`NimbusJwtDecoderJwkSupport` takes whatever exception message Nimbus sends, wraps it in `JwtException` and rethrows to the caller.

One of Nimbus's messages, regarding when a plain (unsigned) JWT is submitted is:

```bash
 Unsecured (plain) JWTs are rejected, extend class to handle
```

This message reveals too much about the underlying implementation, and Spring Security should have its own message that doesn't refer to the class.

### Actual Behavior

When a plain JWT is presented, the exception message mentions classes.

### Expected Behavior

When a plain JWT is presented, the exception message should simply state that plain JWTs are not supported.

### Version

5.1.0.M2
5.0.7

PR #5459, opened 2018-06-25
Close Nimbus Information Leak

jzheaux:
This commit captures and remaps the exception that Nimbus throws
when a PlainJWT is presented to it.

While the surrounding classes are likely only used today by the
oauth2Login flow, since they are public, we'll patch them at this
point for anyone who may be using them directly.

Fixes: gh-5457


rwinch:

jzheaux:
Thanks, @rwinch

Issue #5237, opened 2018-04-16
Resource Server supports hasAuthority


jzheaux:
### Summary

Under the current plan, Resource Server SpEL expressions will only support oauth2-specific expressions at the root. For example:

```java
@PreAuthorize("hasScope('message.read')")
```

will work, but

```java
@PreAuthorize("hasAuthority('message.read')")
```

won't. This will have two unfortunate consequences for end users.

First, it may put oauth2-specific SpEL expressions into generic interfaces, e.g.:

```java
public interface MessageRepository implements CrudRepository {
    @PreAuthorize("hasScope('message.read')")
    Message findById(String id);
}
```

when we'd really like:

```java
public interface MessageRepository implements CrudRepository {
    @PreAuthorize("hasAuthority('message.read')")
    Message findById(String id);
}
```

Second, it may be impossible for a service that needs to support more than just OAuth2 to do so. For example:

```java
public interface MessageRepository implements CrudRepository {
    @PreAuthorize("hasAuthority('message.read') or hasScope('message.read')")
    Message findById(String id);
}
```

may be impossible.

The purpose of this issue is to provide support for `hasAuthority` in Security SpEL expressions when the server is configured as an OAuth2 Resource Server.

Each individual scope in the `scope` attribute in OAuth 2 appears to be a good mapping for the intent of `hasAuthority(authority)`.

jzheaux:
Relates to #4887

Issue #5130, opened 2018-03-16
Resource Server verifies JWT using JWK


jgrandja:
Resource Server will initially provide support for verifying the signature ([JWS](https://tools.ietf.org/html/rfc7515)) of a JWT using a [JWK](https://tools.ietf.org/html/rfc7517) obtained from the JWK Set URI.

dfcoffin:
When will Resource Server support for simple Bearer tokens be available?

jgrandja:
Opaque token support will likely come after 5.1 is released.

Issue #5128, opened 2018-03-16
Resource Server supports JWT


jgrandja:
Resource Server will initially provide support for JWT Access Token Type.

jzheaux:
PR - https://github.com/spring-projects/spring-security/pull/5476/files

Issue #5125, opened 2018-03-16
Resource Server handles missing/invalid Bearer Token


jgrandja:
This feature implements  [Section 3 of RFC6750](https://tools.ietf.org/html/rfc6750#section-3) _The WWW-Authenticate Response Header Field_ .

`AuthenticationEntryPoint` can be leveraged for the implementation.

jgrandja:
Related #5121

PR #5476, opened 2018-07-02
Resource Server JWK support

jzheaux:
Introduce support for OAuth 2.0 Bearer Token Authentication with JWT-encoded bearer tokens. This includes support for remote JWK set signature verification.

#5121 
#5125 
#5128 
#5130 
#5237


config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
```diff
@@ -0,0 +1,896 @@
...
+	public void getWhenUsingDefaultsWithUnsignedBearerTokenThenInvalidToken()
+			throws Exception {
+
+		this.spring.register(DefaultConfig.class).autowire();
+		String token = this.token("Unsigned");
+
+		this.mvc.perform(get("/").with(bearerToken(token)))
+				.andExpect(status().isUnauthorized())
+				.andExpect(invalidTokenHeader("An error occurred while attempting to decode the Jwt: " +
+						"Unsecured (plain) JWTs are rejected, extend class to handle"));
```
jzheaux: Separate pull request out to override this error message.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderJwkSupport.java
```diff
@@ -119,8 +121,18 @@ public Jwt decode(String token) throws JwtException {
 
 			jwt = new Jwt(token, issuedAt, expiresAt, headers, jwtClaimsSet.getClaims());
 
+		} catch ( RemoteKeySourceException jwkFailure ) {
+			if ( jwkFailure.getCause() instanceof ParseException ) {
+				throw new JwtException("An error occurred while attempting to decode the Jwt: Malformed Jwk set");
```
jzheaux: This exception message reveals the underlying jwk set or the contents of whatever page it is pointed at, so this catches and overrides it with a less-revealing message.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderJwkSupport.java
```diff
@@ -119,8 +121,18 @@ public Jwt decode(String token) throws JwtException {
...
+			} else {
+				throw new JwtException("An error occurred while attempting to decode the Jwt: " + jwkFailure.getMessage(), jwkFailure);
+			}
 		} catch (Exception ex) {
-			throw new JwtException("An error occurred while attempting to decode the Jwt: " + ex.getMessage(), ex);
+			if ( ex.getCause() instanceof ParseException ) {
+				throw new JwtException("An error occurred while attempting to decode the Jwt: Malformed payload");
```
jzheaux: This Nimbus exception message reveals the underlying jwt contents.

samples/boot/oauth2resourceserver/README.adoc
```diff
@@ -0,0 +1,161 @@
...
+```
+
+Then, using that access token:
+
+```bash
+curl -H  "Authorization: Bearer eyJhbGciOiJSUzI1NiJ9.eyJzY29wZSI6Im1lc3NhZ2U6cmVhZCIsImV4cCI6MzEwNjMyODEzMSwianRpIjoiOGY5ZjFiYzItOWVlMi00NTJkLThhMGEtODg3YmE4YmViYjYzIn0.CM_KulSsIrNXW1x6NFeN5VwKQiIW-LIAScJzakRFDox8Ql7o4WOb0ubY3CjWYnglwqYzBvH9McCFqVrUtzdfODY5tyEEJSxWndIGExOi2osrwRPsY3AGzNa23GMfC9I03BFP1IFCq4ZfL-L6yVcIjLke-rA40UG-r-oA7r-N_zsLc5poO7Azf29IQgQF0GSRp4AKQprYHF5Q-Nz9XkILMDz9CwPQ9cbdLCC9smvaGmEAjMUr-C1QgM-_ulb42gWtRDLorW_eArg8g-fmIP0_w82eNWCBjLTy-WaDMACnDVrrUVsUMCqx6jS6h8_uejKly2NFuhyueIHZTTySqCZoTA" \
+  localhost:8081
+```
+
+I get:
```
rwinch: Avoid the use of "I" in documentation as it is not inclusive of the reader. Generally, I find it good to avoid personal pronouns, but if you do use them leverage ones that are inclusive of the reader (i.e. we)

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -0,0 +1,237 @@
...
+			}
+		}
+	}
+
+	public class JwtConfigurer {
+		private String jwkSetUri;
+
+		JwtConfigurer() {}
+
+		public OAuth2ResourceServerConfigurer<H> jwkSetUri(String uri) {
```
rwinch: Change this method to set a JwkDecoder vs keeping track of the uri. For example:

```java
this.jwtDecoder = new NimbusJwtDecoderJwkSupport(uri);
```

This is more flexible for the future without the need to keep track of as many member variables.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -0,0 +1,237 @@
...
+
+		BearerTokenAuthenticationFilter filter =
+				new BearerTokenAuthenticationFilter(manager);
+		filter.setBearerTokenResolver(resolver);
+		filter = postProcess(filter);
+
+		http.addFilterBefore(filter, BasicAuthenticationFilter.class);
+
+		if (this.jwtConfigurer != null) {
+			JwtDecoder decoder = getJwtDecoder(http);
```
rwinch: Place the configuration logic inside a protected method of `JwtConfigurer` rather than managing it externally.

samples/boot/oauth2resourceserver/src/integration-test/java/sample/OAuth2ResourceServerApplicationTests.java
```diff
@@ -0,0 +1,139 @@
...
+	public void performWhenValidBearerTokenThenAllows()
+		throws Exception {
+
+		this.mvc.perform(get("/").with(bearerToken(this.noScopesToken)))
+				.andExpect(status().isOk())
+				.andExpect(content().string(containsString("Hello")));
+	}
+
+	@Test
+	public void performWhenValidBearerTokenThenNoSessionCreated()
```
rwinch: This test seems like it should not be in a sample. I believe that most users would just expect this to work and not want it shown off as a feature in a sample.

samples/boot/oauth2resourceserver/src/integration-test/java/sample/OAuth2ResourceServerApplicationTests.java
```diff
@@ -0,0 +1,139 @@
...
+
+		MvcResult result =
+				this.mvc.perform(get("/").with(bearerToken(this.noScopesToken)))
+						.andReturn();
+
+		assertThat(result.getRequest().getSession(false)).isNull();
+	}
+
+	@Test
+	public void performWhenMalformedBearerTokenThenUnauthorized()
```
rwinch: This test seems like it should not be in a sample. I believe that most users would just expect this to work and not want it shown off as a feature in a sample.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -0,0 +1,237 @@
...
+	public void init(H http) throws Exception {
+		registerDefaultAccessDeniedHandler(http);
+		registerDefaultEntryPoint(http);
+		registerDefaultCsrfOverride(http);
+	}
+
+	@Override
+	public void configure(H http) throws Exception {
+		BearerTokenResolver resolver = getBearerTokenResolver(http);
+		resolver = postProcess(resolver);
```
rwinch: We want to avoid invoking postProcess on beans a user may have injected since this would cause issues (i.e. their lifecycle methods would be invoked twice).

The current code is a bit fragile in this respect because it is likely that in the future a resolver will be able to be registered by a user explicitly. If this happens, we would likely forget to only postProcess the default value.

We should only invoke postProcess right next to a new statement within a configurer.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/BearerTokenAuthenticationFilter.java
```diff
@@ -0,0 +1,163 @@
...
+		final boolean debug = this.logger.isDebugEnabled();
+
+		String token;
+
+		try {
+			token = this.bearerTokenResolver.resolve(request);
+		} catch ( OAuth2AuthenticationException invalid ) {
+			this.authenticationEntryPoint.commence(request, response, invalid);
+			return;
+		} catch ( Exception unexpected ) {
```
rwinch: Why do we continue processing the filter chain here? I'd expect this fails. If we are expecting this to be a normal flow we should not log it as an error.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/BearerTokenAuthenticationFilter.java
```diff
@@ -0,0 +1,163 @@
...
+			this.logger.error("An unexpected error occured when trying to parse the request for the bearer token." +
```
rwinch: A typo: instead of occured it is occurred (two r's)

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -0,0 +1,237 @@
...
+		this.matcher.setResolver(resolver);
+
+		AuthenticationManager manager = http.getSharedObject(AuthenticationManager.class);
+
+		BearerTokenAuthenticationFilter filter =
+				new BearerTokenAuthenticationFilter(manager);
+		filter.setBearerTokenResolver(resolver);
+		filter = postProcess(filter);
+
+		http.addFilterBefore(filter, BasicAuthenticationFilter.class);
```
rwinch: Please create a ticket to add BearerTokenAuthenticationFilter to SecurityFilters and ensure it is added to the reference

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -0,0 +1,237 @@
...
+
+			} else {
+				throw new BeanCreationException("Jwt is the only supported format for bearer tokens " +
+						"in Spring Security and no instance of JwtDecoder could be found. Make sure to specify " +
+						"a jwk set uri by doing http.oauth2().resourceServer().jwt().jwkSetUri(uri)");
+			}
+		}
+	}
+
+	public class JwtConfigurer {
```
rwinch: Please create a ticket for adding XML support.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/CsrfConfigurer.java
```diff
@@ -150,6 +150,35 @@ public CsrfConfigurer(ApplicationContext context) {
...
+	 * http
+	 *     .csrf()
+	 *         .ignoringRequestMatchers(request -> request.getHeader("Authorization") != null)
+	 *         .and()
+	 *     ...
+	 * </pre>
+	 *
+	 * @since 5.1
+	 */
+	public CsrfConfigurer<H> ignoringRequestMatchers(RequestMatcher... requestMatchers) {
```
rwinch: Please create this as separate ticket and link to that ticket. Users will be interested to know about this separately. The separate commit can remain in this PR, but please make it a separate ticket and commit.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
```diff
@@ -104,6 +107,25 @@ public ExceptionHandlingConfigurer() {
...
+	 * default {@link AccessDeniedHandler}. If multiple default
+	 * {@link AccessDeniedHandler} instances are configured, then a
+	 * {@link RequestMatcherDelegatingAccessDeniedHandler} will be used.
+	 *
+	 * @param deniedHandler the {@link AccessDeniedHandler} to use
+	 * @param preferredMatcher the {@link RequestMatcher} for this default
+	 * {@link AccessDeniedHandler}
+	 * @return the {@link ExceptionHandlingConfigurer} for further customizations
+	 */
+	public ExceptionHandlingConfigurer<H> defaultAccessDeniedHandlerFor(
```
rwinch: Please create this as separate ticket and link to that ticket. Users will be interested to know about this separately. The separate commit can remain in this PR, but please make it a separate ticket and commit.

config/src/test/groovy/org/springframework/security/config/annotation/web/configurers/CsrfConfigurerTests.groovy
```diff
@@ -507,6 +509,108 @@ class CsrfConfigurerTests extends BaseSpringSpec {
 		}
 	}
 
+	def 'ignoring request matchers augment configured request matcher'() {
```
rwinch: Please do not create new Groovy tests. Instead you can create another Java based test with the new tests in it. Sometimes I will rename the old Groovy tests to end in Spec to avoid duplicate classes with the same name.

This comment should be applied generally to all code going into Spring Security (i.e. make this change for all .groovy files that were edited in this PR)

core/src/main/java/org/springframework/security/core/TransientAuthentication.java
```diff
@@ -0,0 +1,34 @@
...
+ * A marker for {@link Authentication}s that should never be stored across requests, for example
+ * a bearer token authentication
+ *
+ * @author Josh Cummings
+ * @since 5.1
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface TransientAuthentication {
+}
```
rwinch: Please create this as separate ticket and link to that ticket. Users will be interested to know about this separately. The separate commit can remain in this PR, but please make it a separate ticket and commit.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationProvider.java
```diff
@@ -0,0 +1,151 @@
...
+		return new BearerTokenError(
+				BearerTokenErrorCodes.INVALID_TOKEN,
+				HttpStatus.UNAUTHORIZED,
+				message,
+				"https://tools.ietf.org/html/rfc6750#section-3.1");
+	}
+
+	private static Collection<String> getScopes(Jwt jwt) {
+		for ( String attributeName : WELL_KNOWN_SCOPE_ATTRIBUTE_NAMES ) {
+			Object scopes = jwt.getClaims().get(attributeName);
```
rwinch: Consider using jwt.getClaimAsStringList instead.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolver.java
```diff
@@ -0,0 +1,110 @@
...
+
+	private boolean allowUriQueryParameter = false;
+
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public String resolve(HttpServletRequest request) {
+		String authorizationHeaderToken = resolveFromAuthorizationHeader(request);
+		String parameterToken = request.getParameter("access_token");
```
rwinch: We should also fail if the parameter is included twice. This can be checked using request.getParameterValues("access_token")

web/src/test/java/org/springframework/security/web/context/HttpSessionSecurityContextRepositoryTests.java
```diff
@@ -633,4 +637,39 @@ public void failsWithStandardResponse() {
...
+		context.setAuthentication(authentication);
+
+		repo.saveContext(context, holder.getRequest(), holder.getResponse());
+
+		MockHttpSession session = (MockHttpSession) request.getSession(false);
+		assertThat(session).isNull();
+	}
+
+	@TransientAuthentication
+	private static class SomeTransientAuthentication extends AbstractAuthenticationToken {
```
rwinch: Add tests for meta annotations and super class that is annotated with TransientAuthentication

samples/boot/oauth2resourceserver/src/main/resources/application.yml
```diff
@@ -0,0 +1,4 @@
+server:
+  port: 8081
```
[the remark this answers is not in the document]
jzheaux: Was actually just trying to stay out of the way--several authorization servers use 8080 by default (e.g. uaa, keycloak).

Of course, this is very easy for the user to edit. Do you prefer to remove it in favor of having a more minimal example?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
jzheaux: Okay, great. In hindsight, I had intended to ask you when postProcess is useful. Your explanation makes it clear.

config/src/test/groovy/org/springframework/security/config/annotation/web/configurers/CsrfConfigurerTests.groovy
jzheaux: Ah, okay, sure. I can move these easily to their Java equivalent.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/BearerTokenAuthenticationFilter.java
jzheaux: @rwinch, No, it is not a normal flow.

My reasoning is: If the bearer token resolver fails in an unanticipated way, then we cannot know at that point that it is definitely a bearer token request--it might be another type of authentication. In that circumstance, it seemed like this filter would play nicer with others if it yielded.

The alternative, as I see it, is to fail fast, e.g. not have the extra `Exception` block. The trade-off is that if the bearer token resolver has some kind of bug that throws an unexpected exception, that might limit this filter's ability to coexist with downstream filters (which may have been able to handle the request correctly).

Certainly, bearer token resolution is simple enough that such a circumstance is pretty unlikely, so, I think a less polite posture would also be fine if you see it causing more headaches than it is worth. What is your experience regarding past bugs in other similar filters?

samples/boot/oauth2resourceserver/README.adoc
```diff
@@ -0,0 +1,161 @@
...
+
+To run as a stand-alone application, do:
+
+```bash
+./gradlew bootRun
+```
+
+Or import the project into your IDE and run `OAuth2ResourceServerApplication` from there.
+
+Once it is up, you can retreive a valid JWT token from the authorization server, and then hit the endpoint:
```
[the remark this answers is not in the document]
jzheaux: Ooh, I like that. Yes, will do.

samples/boot/oauth2resourceserver/src/integration-test/java/sample/OAuth2ResourceServerApplicationTests.java
jzheaux: Whoops, my bad. Intended to remove these after I added OAuth2ResourceServerConfigurerTests.

samples/boot/oauth2resourceserver/src/main/java/sample/provider/MockProvider.java
```diff
@@ -0,0 +1,134 @@
...
+	);
+
+	public MockProvider() throws IOException {
+		Dispatcher dispatcher = new Dispatcher() {
+			@Override
+			public MockResponse dispatch(RecordedRequest request) throws InterruptedException {
+				if ("/.well-known/jwks.json".equals(request.getPath())) {
+					return JWKS_RESPONSE;
+				}
+
```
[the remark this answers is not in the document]
jzheaux: The other two responses are so that the reader can follow all the instructions in the readme regardless of the authorization server she is using (mock or otherwise). Specifically, she can hit the `/token` endpoint for either a token that has no scopes or a token that has the `message:read` scope from the command-line.

While your comment about hardcoding the token in the instructions obviates the technical need for the mock's `/token` endpoint, the endpoint is still instructive for folks drawing the conceptual bridge from this sample to using their real authorization server.

For example, a user could write a simple bash script that retrieves a token, sets it to an environment variable, and then requests the endpoint using that token. Now, the user can point that script at either the mock authorization server or their own authorization server.

Does this step out of the bounds of the purpose of the sample?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
jzheaux: fyi - #5478

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
jzheaux: fyi - https://github.com/spring-projects/spring-security/issues/5479

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
jzheaux: How detailed should a task for adding XML support be?

As I imagine that such a task will be somewhat on-going, i.e. as each task get completed, there is additional XML support to add.

config/src/test/groovy/org/springframework/security/config/annotation/web/configurers/CsrfConfigurerTests.groovy
jzheaux: I see that there is a class called `CsrfConfigurerNoWebMvcTests`. Was going to create `CsrfConfigurerJavaTests` (to differentiate from the groovy `CsrfConfigurerTests`) before I saw this.

What convention should I be shooting for? For example, is naming it `CsrfConfigurerRequestMatcherTests` more appropriate?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
jzheaux: fyi - #5480

core/src/main/java/org/springframework/security/core/TransientAuthentication.java
jzheaux: fyi - #5481

samples/boot/oauth2resourceserver/src/main/java/sample/provider/MockProvider.java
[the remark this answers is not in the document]
rwinch: I think for the sake of simplicity we should just give the user the token rather than pretend the mock server is a valid authorization server. The MockProvider is not a valid provider, nor do we want it to be. It does not validate the client id / secret, it does not validate the request, etc. Instead, leave this for a step 2 in the using a real Authorization Server steps. This helps the user build their knowledge gradually vs trying to force them to learn too much at once. It also avoids confusing them since the MockProvider is not a valid authorization server.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/BearerTokenAuthenticationFilter.java
jzheaux: Can you clarify what you meant here: 

> Allowing it to pass through makes it confusing as a user would then get some sort of other error that isn't related to the bearer token.

This statement makes sense if our premise is that this request is a bearer token request, and we just messed up in determining that.

My premise, though, is that if an unanticipated exception is thrown in the resolver, then we don't know that it is a bearer token request since the code that determines whether or not this filter wants to participate has a bug itself.

If I don't know that it is a bearer token request, then why should I be worried about errors from downstream filters (indeed those errors may be the correct ones since it may be a request of another type)?

Again, just looking for clarification on that specific point. Fail-fast is a good principle, and it would certainly help us prioritize that (hypothetical) bug instead of it lingering in the codebase.

rwinch: If you continue the FilterChain vs letting the Exception be thrown, it draws attention away from the actual problem. If there is a bug in code, let it have a bug and let the developer know where it is by allowing the Exception to be thrown.

Consider a situation where a client is trying to use a Bearer token, there is a bug, and we continue processing the request after an Exception is thrown. The developer now has much more code to sift through to figure out what is happening since we continue the FilterChain and there is more logging to look through.

samples/boot/oauth2resourceserver/src/integration-test/java/sample/OAuth2ResourceServerApplicationTests.java
```diff
@@ -0,0 +1,139 @@
...
+		this.mvc.perform(get("/denyAll").with(bearerToken(this.messageReadToken)))
+				.andExpect(status().isForbidden())
+				.andExpect(header().string(HttpHeaders.WWW_AUTHENTICATE,
+						containsString("Bearer error=\"insufficient_scope\", " +
+								"error_description=\"The token provided has insufficient scope [message:read] for this request\", " +
+								"error_uri=\"https://tools.ietf.org/html/rfc6750#section-3.1\", " +
+								"scope=\"message:read\"")));
+	}
+
+	private static class BearerTokenRequestPostProcessor implements RequestPostProcessor {
```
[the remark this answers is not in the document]
jzheaux: fyi - #5485

samples/boot/oauth2resourceserver/src/main/java/sample/provider/MockProvider.java
```diff
@@ -0,0 +1,134 @@
...
+ * - The token is signed by a RS256 private key whose public key counterpart is served from the JWK endpoint of the AS.
+ *
+ * There is also a test that verifies insufficient scope. In that case, the token should have the following characteristics:
+ *
+ * - The token is missing the "message:read" scope
+ * - The token is signed by a RS256 private key whose public key counterpart is served from the JWK endpoint of the AS.
+ *
+ * @author Josh Cummings
+ */
+public class MockProvider implements EnvironmentPostProcessor {
```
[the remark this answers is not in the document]
jzheaux: fyi - https://github.com/spring-projects/spring-security/issues/5486

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationProvider.java
jzheaux: Actually, it appears that this method only returns the list if it is already a list, otherwise, it returns null. I'm inclined to leave this for now and submit and possibly submit an enhancement for getClaimAsStringList, maybe and overloaded method that takes a delimiter as a parameter?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
jzheaux: fyi - #5480

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/BearerTokenAuthenticationEntryPoint.java
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * @author Vedran Pavic
+ * @since 5.1
+ * @see BearerTokenError
+ * @see <a href="https://tools.ietf.org/html/rfc6750#section-3" target="_blank">RFC 6750 Section 3: The WWW-Authenticate
+ * Response Header Field</a>
+ */
+public final class BearerTokenAuthenticationEntryPoint implements AuthenticationEntryPoint {
+
+	private String defaultRealmName;
```
rwinch: Is there a reason this is named defaultRealmName vs realmName? I think that this should be renamed to realmName since it is the realm name and not just a default. The property realmName is also more consistent with BasicAuthenticationEntryPoint.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/access/BearerTokenAccessDeniedHandler.java
```diff
@@ -0,0 +1,134 @@
...
+	public final void setRealmName(String realmName) {
+		this.realmName = realmName;
+	}
+
+	private static String getScope(AbstractOAuth2TokenAuthenticationToken token) {
+
+		Map<String, Object> attributes = token.getTokenAttributes();
+
+		for (String attributeName : WELL_KNOWN_SCOPE_ATTRIBUTE_NAMES) {
+			Object scopes = attributes.get(attributeName);
```
jgrandja: `scopes` could be `null` so should do `null` check first

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/BearerTokenAuthenticationFilter.java
```diff
@@ -0,0 +1,158 @@
...
+		Assert.notNull(authenticationManager, "authenticationManager cannot be null");
+		this.authenticationManager = authenticationManager;
+	}
+
+	/**
+	 * Construct a {@code BearerTokenAuthenticationFilter} using the provided parameter(s)
+	 * @param authenticationManager
+	 * @param authenticationEntryPoint
+	 */
+	public BearerTokenAuthenticationFilter(
```
jgrandja: Remove this constructor as it's only being used in a test.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -0,0 +1,228 @@
...
+ * for a custom one.
+ *
+ * @author Josh Cummings
+ * @since 5.1
+ * @see BearerTokenAuthenticationFilter
+ * @see JwtAuthenticationProvider
+ * @see NimbusJwtDecoderJwkSupport
+ * @see AbstractHttpConfigurer
+ */
+public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<H>> extends
```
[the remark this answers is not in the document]
rwinch: Yes, but let's create a separate ticket.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -0,0 +1,228 @@
...
+			JwtDecoder decoder = this.jwtConfigurer.getJwtDecoder(http);
+
+			if (decoder != null) {
+				JwtAuthenticationProvider provider =
+						new JwtAuthenticationProvider(decoder);
+				provider = postProcess(provider);
+
+				http.authenticationProvider(provider);
+			} else {
+				throw new BeanCreationException("Jwt is the only supported format for bearer tokens " +
```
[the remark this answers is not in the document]
jzheaux: That is not true. It can happen if the user does:

```java
http.oauth2().resourceServer().jwt();
```

instead of:

```java
http.oauth2().resourceServer().jwt().jwkSetUri(uri)
```

There is a test that confirms this: https://github.com/spring-projects/spring-security/pull/5476/files#diff-0e46655e74089007a737c658d46b43b6R588

What did I miss?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -0,0 +1,228 @@
...
+		AuthenticationManager manager = http.getSharedObject(AuthenticationManager.class);
+
+		BearerTokenAuthenticationFilter filter =
+				new BearerTokenAuthenticationFilter(manager);
+		filter.setBearerTokenResolver(resolver);
+		filter = postProcess(filter);
+
+		http.addFilterBefore(filter, BasicAuthenticationFilter.class);
+
+		if (this.jwtConfigurer != null) {
```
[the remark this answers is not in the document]
jzheaux:
```java
http.oauth2().resourceServer()
```

This is a valid configuration that allows folks to get bearer token support only. I would want bearer token support in the case that I want to, for example, wire my own authentication provider instead of the provided one. There is an explanation in the class java doc in this class.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -0,0 +1,228 @@
...
+		private JwtDecoder decoder;
+
+		JwtConfigurer() {}
+
+		public OAuth2ResourceServerConfigurer<H> jwkSetUri(String uri) {
+			this.decoder = new NimbusJwtDecoderJwkSupport(uri);
+			return OAuth2ResourceServerConfigurer.this;
+		}
+
+		protected JwtDecoder getJwtDecoder(H http) {
```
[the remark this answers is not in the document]
jzheaux: Impressive catch! I actually intentionally left this because I see this in other places in the code, which I figured was meant to signal to others that this variable is an allowable mechanism for making the implementation decision.

Such is the case, for example, in ExceptionHandlingConfigurer already: https://github.com/spring-projects/spring-security/blob/master/config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java#L194

Am I reading too much into that (note that I followed it too in ExceptionHandlingConfigurer when I created createDefaultDeniedHandler)?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -0,0 +1,228 @@
...
+		}
+	}
+
+	public class JwtConfigurer {
+		private JwtDecoder decoder;
+
+		JwtConfigurer() {}
+
+		public OAuth2ResourceServerConfigurer<H> jwkSetUri(String uri) {
+			this.decoder = new NimbusJwtDecoderJwkSupport(uri);
```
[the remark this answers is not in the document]
jzheaux: I talked with @rwinch about this, and he recommended moving getJwtDecoder into this class for better management. As such, the implementation choice was made here.

However, I see your point about deferring until the configure phase, in which case I could place it in `getJwtDecoder`. It just means keeping track of another variable in the class (`uri`).

What do you think of placing it in `getJwtDecoder` (which is called from `configure`)?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -0,0 +1,228 @@
...
+			this.jwtConfigurer = new JwtConfigurer();
+		}
+
+		return this.jwtConfigurer;
+	}
+
+	@Override
+	public void setBuilder(H http) {
+		super.setBuilder(http);
+		initSessionCreationPolicy(http);
```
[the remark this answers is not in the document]
jzheaux: Actually, no, it can't. I can give you a brief explanation here, and ping me if it still isn't clear why.

`SessionManagementConfigurer` makes its decisions about which `SecurityContextRespository` and `RequestCache` to use inside its `init` method, and it does so based on what the user has set for `sessionCreationPolicy`. By the time this configurer sees this decision, it cannot tell whether the user made that decision or `SessionManagementConfigurer` by default.

This is one that @rwinch and I chatted about for a bit early on. I'm open to suggestions if there is a better way (make sure to see the changes I made to `SessionManagementConfigurer` and how those changes affect its `init` method, not its `configure` method.).

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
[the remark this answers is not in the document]
jzheaux: > Yes, but let's create a separate ticket

@rwinch To clarify, for this PR or no? As I see it, these are not necessary to support JWK sets nor JWT support, so they are not part of this PR, at least as I understand https://github.com/jzheaux/spring-security/issues/18

And if they are for this PR, could you explain why as this doesn't seem to jive in my head with "take everything out that isn't directly needed for X feature set"? The reason for my question is not to debate the point, but just to make sure that I have more context for the future.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/CsrfConfigurerIgnoringRequestMatchersTests.java
```diff
@@ -0,0 +1,134 @@
...
+	@Autowired
+	MockMvc mvc;
+
+	@Rule
+	public final SpringTestRule spring = new SpringTestRule();
+
+	@Test
+	public void requestWhenIgnoringRequestMatchersThenAugmentedByConfiguredRequestMatcher()
+			throws Exception {
+		this.spring.register(IgnoringRequestMatchers.class, BasicController.class).autowire();
```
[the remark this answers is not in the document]
jzheaux: I don't think I follow. The reason I'm wiring `BasicController` is because I need those endpoints (in `BasicController`) to confirm the functionality.

If you are saying that `@EnableWebMvc` will autodiscover them, are you suggesting doing that since all the tests require `BasicController` anyway? My concern with that thinking is that as more tests get added, they may not need `BasicController`, and it may not be readily apparent that they are getting an instance of it (e.g. I think that asking directly in the setup makes it clearer what this test is actually using and what it isn't).

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
```diff
@@ -0,0 +1,927 @@
...
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.content;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.header;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+import static org.springframework.web.bind.annotation.RequestMethod.GET;
+import static org.springframework.web.bind.annotation.RequestMethod.POST;
+
+/**
+ * @author Josh Cummings
+ */
+public class OAuth2ResourceServerConfigurerTests {
```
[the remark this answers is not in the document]
jzheaux: Sure can. I've seen that in the codebase in a couple of places, though it appeared to be more tactical,--e.g. a place to put Java tests while we wait for the corresponding Groovy test class to be migrated--than strategic.

I think something like this is better-suited for a separate ticket. Thoughts?

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderJwkSupport.java
```diff
@@ -135,8 +138,18 @@ private Jwt createJwt(String token, JWT parsedJwt) {
 
 			jwt = new Jwt(token, issuedAt, expiresAt, headers, jwtClaimsSet.getClaims());
 
+		} catch (RemoteKeySourceException ex) {
+			if (ex.getCause() instanceof ParseException) {
+				throw new JwtException("An error occurred while attempting to decode the Jwt: Malformed Jwk set");
+			} else {
+				throw new JwtException("An error occurred while attempting to decode the Jwt: " + ex.getMessage(), ex);
```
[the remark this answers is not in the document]
jzheaux: Yes. It occurs, for example, when the remote url is unresponsive. We do have this one: https://github.com/spring-projects/spring-security/pull/5476/files#diff-0e46655e74089007a737c658d46b43b6R148 but, yes, we should add one to this class's unit tests.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/BearerTokenAuthenticationToken.java
```diff
@@ -0,0 +1,81 @@
...
+	public BearerTokenAuthenticationToken(String token) {
+		super(Collections.emptyList());
+
+		Assert.hasText(token, "token cannot be empty");
+
+		this.token = token;
+	}
+
+	/**
+	 * Get the <a href="https://tools.ietf.org/html/rfc6750#section-1.2" target="_blank">Bearer Token</a>
```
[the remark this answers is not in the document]
jzheaux: Are we removing it because it is copy-paste? If I am reading a Javadoc, I want to know what a method does, and I won't know if it isn't described.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/BearerTokenAuthenticationToken.java
```diff
@@ -0,0 +1,81 @@
...
+		return this.getToken();
+	}
+
+	/**
+	 * Get the <a href="https://tools.ietf.org/html/rfc6750#section-1.2" target="_blank">Bearer Token</a>
+	 * @return
+	 */
+	@Override
+	public Object getPrincipal() {
+		return this.getToken();
```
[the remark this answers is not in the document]
jzheaux: I believe this comes down to whether or not the token is a principal at all. Note that `UsernamePasswordAuthenticationToken`, when used as an authentication request, does take the value from the constructor, set it as the principal, and returns it from getPrincipal. The difference is that `username` is more clearly a principal than `token`.

@rwinch and I were just discussing this one a few days ago. Another option is to return `null`. Really for me, the best `Authentication` would be one that is principal-less since Bearer Tokens are primarily about authority and incidentally about identity (say, when there is a `sub`).

Returning `null` clearly states that there is no "entity" that is part of this request, returning `null` just feels like a smell.

Thoughts, @rwinch ?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/access/BearerTokenAccessDeniedHandler.java
jzheaux: I believe this is what `instanceof` does for me, no? e.g. A `NullPointerException` is not possible in the ensuing block of code.

Are you suggesting that this makes it more readable? I tend to see unnecessary `null` checks as extra verbosity, of which Java already has plenty. :)

web/src/main/java/org/springframework/security/web/access/RequestMatcherDelegatingAccessDeniedHandler.java
```diff
@@ -0,0 +1,76 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.springframework.security.access.AccessDeniedException;
+import org.springframework.security.web.util.matcher.RequestMatcher;
+import org.springframework.util.Assert;
+
+/**
+ * An {@link AccessDeniedHandler} that delegates to other {@link AccessDeniedHandler}
+ * instances based upon the type of {@link AccessDeniedException} passed into
```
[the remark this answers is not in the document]
jzheaux: Hmm.. those got lost in a copy at some point. These were up to date... either way, good catch.

web/src/test/java/org/springframework/security/web/context/HttpSessionSecurityContextRepositoryTests.java
```diff
@@ -633,4 +619,102 @@ public void failsWithStandardResponse() {
...
+	}
+
+	private static class SomeTransientAuthenticationSubclass extends SomeTransientAuthentication {
+
+	}
+
+	@Target(ElementType.TYPE)
+	@Retention(RetentionPolicy.RUNTIME)
+	@TransientAuthentication
+	public @interface TestTransientAnnotation {
```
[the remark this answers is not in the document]
jzheaux: > Rename to `TestTransientAuthentication`?

Yep, will do.

samples/boot/oauth2resourceserver/src/main/java/sample/provider/MockProvider.java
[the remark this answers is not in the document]
jgrandja: Interesting approach on mocking the Jwk Set endpoint. However, I'm wondering if this will confuse the user as the `EnvironmentPostProcessor` is not intended for starting up a mock server. I'm wondering if there is another approach, for example, an `ApplicationContextInitializer` via `SpringApplication.addInitializers()`.

@rwinch What are your thoughts?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/SessionManagementConfigurer.java
```diff
@@ -18,7 +18,6 @@
 import java.util.ArrayList;
```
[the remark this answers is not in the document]
rwinch: Yes

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderJwkSupport.java
```diff
@@ -135,8 +138,18 @@ private Jwt createJwt(String token, JWT parsedJwt) {
 
 			jwt = new Jwt(token, issuedAt, expiresAt, headers, jwtClaimsSet.getClaims());
 
+		} catch (RemoteKeySourceException ex) {
```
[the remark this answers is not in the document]
rwinch: I think we should leave it as is. 

We don't want to require services to be started in a specific order which would be the case if we checked at startup time. Just because it worked at startup doesn't mean we won't get the error later either.

It is quite possible they keys that are already loaded will work for many requests, so failing immediately for this request is quite reasonable as long as a previous key that was already loaded continues to work.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/BearerTokenAuthenticationToken.java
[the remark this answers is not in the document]
rwinch: I still struggle with this myself, but I think at least or now we should leave as is. Perhaps we should create a ticket to consider this separately.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/SessionManagementConfigurerSessionCreationPolicyTests.java
```diff
@@ -0,0 +1,93 @@
...
+	public void doFilterWhenUserSessionCreationPolicyConfigurationThenOverrides()
+			throws Exception {
+
+		this.spring.register(StatelessCreateSessionUserConfig.class).autowire();
+
+		MvcResult result = this.mvc.perform(get("/")).andReturn();
+
+		assertThat(result.getRequest().getSession(false)).isNull();
+	}
+
```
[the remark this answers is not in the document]
jzheaux: Added `getWhenDefaultsThenLoginChallengeCreatesSession`. This fulfills @jgrandja's requirement because it is circumscribed by `http` defaults. So, this test simply confirms that the defaults behave as expected.

web/src/main/java/org/springframework/security/web/access/RequestMatcherDelegatingAccessDeniedHandler.java
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright 2002-2015 the original author or authors.
```
jgrandja: Update header to 2018

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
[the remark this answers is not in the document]
jgrandja: I spoke to @rwinch about this use case and he also agrees that this should be left out of this PR. Please log a new ticket for this.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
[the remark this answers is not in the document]
jgrandja: I'm fine with another ticket. After looking at it again, it's not too long but I feel it will get long sooner than later so having a ticket to re-org it at a later point is fine with me.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/BearerTokenAuthenticationToken.java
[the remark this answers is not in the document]
jgrandja: I personally feel `getCredentials()` should return the `token` and `getPrincipal()` should return `""`.
Logging a ticket to revisit is fine with me too.

rwinch:
Merged in via https://github.com/spring-projects/spring-security/commits/d5950988231d03c7669c5c9433efed0d1af0b406

Issue #5492, opened 2018-07-10
Enhance OAuth2AccessToken to be Serializable


mhyeon-lee:
### Summary

Change the TokenType to Serializable so that the OAuth2AccessToken can be serialized.\
(org.springframework.security.oauth2.core.OAuth2AccessToken.TokenType)

### Actual Behavior

An exception is thrown when serializing to store OAuth2AccessToken in the external repository.

### Expected Behavior

The OAuth2AccessToken value can be serialized.

### Version

5.0.6

### Sample

Below test will throw exception.
``` java
@Test
public void oAuth2AccessTokenSerialize() throws IOException {
	OAuth2AccessToken accessToken = new OAuth2AccessToken(
				TOKEN_TYPE, TOKEN_VALUE, ISSUED_AT, EXPIRES_AT, SCOPES);

	// serialize OAuth2AccessToken
	byte[] serialized;
	try (ByteArrayOutputStream baos = new ByteArrayOutputStream();
		ObjectOutputStream oos = new ObjectOutputStream(baos)) {
		oos.writeObject(accessToken);
		serialized = baos.toByteArray();
	}
}
```

```
java.io.NotSerializableException: org.springframework.security.oauth2.core.OAuth2AccessToken$TokenType

	at java.base/java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1185)
	at java.base/java.io.ObjectOutputStream.defaultWriteFields(ObjectOutputStream.java:1553)
	at java.base/java.io.ObjectOutputStream.writeSerialData(ObjectOutputStream.java:1510)
	at java.base/java.io.ObjectOutputStream.writeOrdinaryObject(ObjectOutputStream.java:1433)
	at java.base/java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1179)
	at java.base/java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:349)
	at org.springframework.security.oauth2.core.OAuth2AccessTokenTests.oAuth2AccessTokenSerializeAndDeserialize(OAuth2AccessTokenTests.java:91)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #5493, opened 2018-07-10
Enhance OAuth2AccessToken to be serializable.

mhyeon-lee:
Change the TokenType to Serializable
so that the OAuth2AccessToken can be serialized.
(org.springframework.security.oauth2.core.OAuth2AccessToken.TokenType)

Fixes gh-5492


jgrandja:
Thanks for the PR @mhyeon-lee! This is now in master.

Issue #5494, opened 2018-07-10
OAuth2 ClientRegistration scope can be null / empty


mhyeon-lee:
### Summary

Allows scope of OAuth2 ClientRegistration to be null.

- The scope setting in the RFC document is defined as Optional.
https://tools.ietf.org/html/rfc6749#section-4.1.1

    > scope:  OPTIONAL.  The scope of the access request as described by Section 3.3.

- When the client omits the scope parameter, validation is determined by the authorization server.
https://tools.ietf.org/html/rfc6749#section-3.3

    > If the client omits the scope parameter when requesting
   authorization, the authorization server MUST either process the
   request using a pre-defined default value or fail the request
   indicating an invalid scope.  The authorization server SHOULD
   document its scope requirements and default value (if defined).

### Actual Behavior

Config ClientRegistration scope is null.

``` java
ClientRegistration.withRegistrationId(REGISTRATION_ID)
			.clientId(CLIENT_ID)
			.clientSecret(CLIENT_SECRET)
			.clientAuthenticationMethod(ClientAuthenticationMethod.BASIC)
			.authorizationGrantType(AuthorizationGrantType.AUTHORIZATION_CODE)
			.redirectUriTemplate(REDIRECT_URI)
			.scope((String[]) null)
			.authorizationUri(AUTHORIZATION_URI)
			.tokenUri(TOKEN_URI)
			.clientName(CLIENT_NAME)
			.build();
```

### Expected Behavior

Build ClientRegistration  success.

### Version

5.0.6

jgrandja:
@mhyeon-lee Yes, the `scope` is OPTIONAL as per spec. However, it is required based on the **OAuth 2.0 Login** feature. The user needs to specify either `openid` (required) if using an OpenID Connect provider (and any other scopes they may require). Also, if the provider is not OpenID Connect compliant, than the user needs to specify the `scope` required to access the `UserInfo` endpoint.

mhyeon-lee:
@jgrandja 
The following OpenID Connect provider(OidcAuthorizationCodeAuthenticationProvider) checks if `openid` exist in scopes.
So even if it allow nulls for scopes configuration, it will need an "openid" scope when using OpenID Connect.
https://github.com/spring-projects/spring-security/blob/master/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcAuthorizationCodeAuthenticationProvider.java#L108

---

[...]

PR #5495, opened 2018-07-10
Fix OAuth2 ClientRegistration scope can be null.

mhyeon-lee:
Allows scope of OAuth2 ClientRegistration to be null.

- The scope setting in the RFC document is defined as Optional.
https://tools.ietf.org/html/rfc6749#section-4.1.1

    > scope:  OPTIONAL.
    > The scope of the access request as described by Section 3.3.

- When the client omits the scope parameter,
validation is determined by the authorization server.
https://tools.ietf.org/html/rfc6749#section-3.3

    > If the client omits the scope parameter when requesting
   authorization, the authorization server MUST either process the
   request using a pre-defined default value or fail the request
   indicating an invalid scope.  The authorization server SHOULD
   document its scope requirements and default value (if defined).

 Fixed gh-5494


jgrandja:
Thank you for the PR @mhyeon-lee! This is now in master.

Issue #5520, opened 2018-07-16
DefaultOAuth2AuthorizationRequestResolver should exclude queryParams in baseUrl


mhyeon-lee:
### Summary

To create redirect_uri in `DefaultOAuth2AuthorizationRequestResolver`, queryParam is included in the current request-based baseUrl.
So when binding to the redirectUriTemplate, the wrong type of redirect_uri may be created.

### Actual Behavior

redirectUriTempate: "{baseUrl}/{action}/oauth2/code/{registrationId}"

request: http://localhost/oauth2/authorization/registration-1?foo=bar
redirect_uri: http://localhost?foo=bar/login/oauth2/code/registration-1

### Expected Behavior

redirect_uri: http://localhost/login/oauth2/code/registration-1

### Configuration

- The following tests fail.

``` java
@Test
public void resolveWhenAuthorizationRequestRedirectUriTemplatedThenRedirectUriExpandedExcludesQueryString() {
	ClientRegistration clientRegistration = this.registration2;
	String requestUri = this.authorizationRequestBaseUri + "/" + clientRegistration.getRegistrationId();
	MockHttpServletRequest request = new MockHttpServletRequest("GET", requestUri);
	request.setServletPath(requestUri);
	request.setQueryString("foo=bar");

	OAuth2AuthorizationRequest authorizationRequest = this.resolver.resolve(request);
	assertThat(authorizationRequest.getRedirectUri()).isEqualTo(
			"http://localhost/login/oauth2/code/" + clientRegistration.getRegistrationId());
}
```

### Version

- commit: 779597af2a6ed777707f08ae8106818e0b8e299e
- related: #4911

PR #5522, opened 2018-07-16
Fix DefaultOAuth2AuthorizationRequestResolver baseUrl to exclude queryString

mhyeon-lee:
To create redirect_uri in DefaultOAuth2AuthorizationRequestResolver,
queryParam is included in the current request-based baseUrl.
So when binding to the redirectUriTemplate,
the wrong type of redirect_uri may be created.

redirectUriTempate: "{baseUrl}/{action}/oauth2/code/{registrationId}"
request: http://localhost/oauth2/authorization/registration-1?foo=bar

redirect_uri: http://localhost?foo=bar/login/oauth2/code/registration-1
-> http://localhost/login/oauth2/code/registration-1

Fixed: gh-5520


jgrandja:
Thanks for the fix @mhyeon-lee! This is now in master.

Issue #5479, opened 2018-07-03
Add BearerTokenAuthenticationFilter to SecurityFilters


jzheaux:
Also, add it to the reference documentation.

Issue #5480, opened 2018-07-03
XML support for Resource Server Jwt Support


jzheaux:
Add XML Resource Server Jwt support corresponding to the existing Java config.

rwinch:
Thank you for the submission. I'm going to close this because XML Namespace support is effectively in maintenance mode. This means that while we will fix bugs, we do not expect to enhance it going forward.

PR #5524, opened 2018-07-16
Add Bearer Token filter to Security Filters

jzheaux:
This introduces BearerTokenAuthenticationFilter to SecurityFilters so
that it can be used in the various addFilter methods and with the
custom-filter xml tag.

Fixes: gh-5479


config/src/main/resources/org/springframework/security/config/spring-security-5.1.xsd
```diff
@@ -2721,6 +2721,7 @@
          <xs:enumeration value="LOGIN_PAGE_FILTER"/>
          <xs:enumeration value="LOGOUT_PAGE_FILTER"/>
          <xs:enumeration value="DIGEST_AUTH_FILTER"/>
+         <xs:enumeration value="BEARER_TOKEN_AUTH_FILTER"/>
```
jzheaux: Is this something that we should wait on until https://github.com/spring-projects/spring-security/issues/5480

rwinch:
Thanks for the PR @jzheaux! It is now merged into master

config/src/main/resources/org/springframework/security/config/spring-security-5.1.xsd
jgrandja: @jzheaux I'm not sure if the `xsd` or `rnc` file should have been changed? I never did update these for `oauth2-client` filters given that xml support is not there yet. Also, when I do a local build one of these files is changed which never happened before this commit. I believe it's the `xsd` file that get's generated. 

@rwinch Is this change necessary?

Issue #5519, opened 2018-07-16
Support user-provided JwtDecoder


jgrandja:
We need to allow the capability for the user to configure a `JwtDecoder` via `oauth2().resourceServer().jwt().decoder(jwtDecoder)` AND support `@Bean` discovery.

See `ClientRegistrationRepository` in `OAuth2ClientConfigurer`...same logic should apply

PR #5531, opened 2018-07-17
User-Specified JwtDecoder

jzheaux:
This exposes JwtConfigurer#decoder as well as makes the configurer
look in the application context for a bean of type JwtDecoder.

Fixes: gh-5519

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
@@ -18,8 +18,8 @@ package org.springframework.security.config.annotation.web.configurers.oauth2.se
 
 import javax.servlet.http.HttpServletRequest;
 
+import org.springframework.context.ApplicationContext;
 import org.springframework.security.authentication.AuthenticationManager;
-import org.springframework.security.authentication.AuthenticationProvider;
 import org.springframework.security.config.annotation.web.HttpSecurityBuilder;
 import org.springframework.security.config.annotation.web.configurers.AbstractHttpConfigurer;
 import org.springframework.security.config.annotation.web.configurers.CsrfConfigurer;
@@ -51,7 +51,19 @@ import org.springframework.util.Assert;
  * </ul>
  *
  * <p>
- * When using {@link #jwt()}, a Jwk Set Uri must be supplied via {@link JwtConfigurer#jwkSetUri}
+ * When using {@link #jwt()}, either
+ *
+ * <ul>
+ * <li>
+ * supply a Jwk Set Uri via {@link JwtConfigurer#jwkSetUri}, or
+ * </li>
+ * <li>
+ * supply a {@link JwtDecoder} instance via {@link JwtConfigurer#decoder}, or
+ * </li>
+ * <li>
+ * expose a {@link JwtDecoder} bean
+ * </li>
+ * </ul>
  *
  * <h2>Security Filters</h2>
  *
@@ -77,10 +89,6 @@ import org.springframework.util.Assert;
  * <li>{@link AuthenticationManager}</li>
  * </ul>
  *
- * If {@link #jwt()} isn't supplied, then the {@link BearerTokenAuthenticationFilter} is still added, but without
- * any OAuth 2.0 {@link AuthenticationProvider}s. This is useful if needing to switch out Spring Security's Jwt support
- * for a custom one.
- *
  * @author Josh Cummings
  * @since 5.1
  * @see BearerTokenAuthenticationFilter
@@ -100,9 +108,14 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 	private BearerTokenAccessDeniedHandler accessDeniedHandler
 			= new BearerTokenAccessDeniedHandler();
 
-	private JwtConfigurer jwtConfigurer = new JwtConfigurer();
+	private JwtConfigurer jwtConfigurer;
 
 	public JwtConfigurer jwt() {
+		if ( this.jwtConfigurer == null ) {
+			ApplicationContext context = this.getBuilder().getSharedObject(ApplicationContext.class);
+			this.jwtConfigurer = new JwtConfigurer(context);
+		}
+
 		return this.jwtConfigurer;
 	}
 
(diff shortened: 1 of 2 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -154,12 +166,26 @@ public void configure(H http) throws Exception {
...
 		public OAuth2ResourceServerConfigurer<H> jwkSetUri(String uri) {
 			this.decoder = new NimbusJwtDecoderJwkSupport(uri);
 			return OAuth2ResourceServerConfigurer.this;
 		}
 
-		private JwtDecoder getJwtDecoder() {
+		private JwtDecoder getJwtDecoder(H http) {
+			if ( this.decoder == null ) {
+				ApplicationContext context = http.getSharedObject(ApplicationContext.class);
+				Map<String, JwtDecoder> decoders =
```
rwinch: It is surprising if there are multiple `JwtDecoder` beans that we just use the first one. This can make troubleshooting issues pretty tricky. I think we should just rely on `context.getBean(JwtDecoder.class)` if the user did not explicitly configure the `JwtDecoder` as this will give a reasonable error message that we need a `JwtDecoder`.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
```diff
@@ -689,6 +747,98 @@ protected void configure(HttpSecurity http) throws Exception {
...
+			// @formatter:on
+		}
+
+		@Bean
+		public JwtDecoder decoder() {
+			return new CustomJwtDecoder();
+		}
+	}
+
+	static class CustomJwtDecoder implements JwtDecoder {
```
rwinch: I'd prefer we just use a mock instead of a custom implementation. This is why Mockito exists and it helps reduces boilerplate code.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
```diff
@@ -689,6 +747,98 @@ protected void configure(HttpSecurity http) throws Exception {
...
+
+		@Bean
+		public JwtDecoder decoder() {
+			return new NimbusJwtDecoderJwkSupport("https://example.org");
+		}
+	}
+
+	@EnableWebSecurity
+	static class CustomJwtDecoderAndJwkSetUri extends WebSecurityConfigurerAdapter {
+		@Value("${mock.jwk-set-uri:https://example.org}") String uri;
```
rwinch: I think we should be able to just hardcode this uri. This will make the test easier to follow.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
rwinch: We should add a tests for the cases where

* there are two `JwtDecoder` beans and it is explicitly configured via the DSL (expected to work with the configured value)
* there are two `JwtDecoder` beans and it is not explicitly configured via the DSL (should produce an error since it is ambiguous which bean to use)

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
jzheaux: The trouble I ran into was that since it is using MockWebServer, then I don't know the uri at this point. However, I can see about instead auto-wiring MockWebServer into this class.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
jzheaux: That is certainly simpler, and yes I'd prefer leaving that resolution logic to someone else.

I'm curious because I see this pattern of calling BeanFactoryUtils in other configurers (that is, call this same method, and take the first value from the map), and I wonder if there is good reason for using this pattern elsewhere?

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
rwinch: Can we just fail to connect to the URL? I don't think we should be connecting to a server in tests.  Starting a server seems more like integration tests and should be done pretty sparingly rather than for edge cases.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
rwinch: It depends on if it needs to be lazy or not. For example AuthenticationConfiguration uses the `BeanFactoryUtils` to create a lazy implementation of an object because looking the bean up at that point in time is likely to cause a cycle. I noticed that OAuth2LoginConfigurer is using BeanFactoryUtils...this appears to be a mistake from what I can tell. Can you create a ticket for it?

jgrandja: @jzheaux Have you seen `OAuth2ClientConfigurerUtils`? It handles the case where there is more than 1 `@Bean` registered and throws a descriptive exception. The same logic would apply.

@rwinch Can you expand on the issue using `BeanFactoryUtils.beansOfTypeIncludingAncestors` in `OAuth2ClientConfigurerUtils`?

rwinch: @jzheaux I'm not sure what advantage this brings to just `context.getBean(clazz)`. It also throws an informative error message.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
```diff
@@ -506,6 +514,56 @@ public void requestWhenSessionManagementConfiguredThenUserConfigurationOverrides
...
+		this.spring.register(CustomJwtDecoderConflicting.class, BasicController.class).autowire();
+
+		this.mvc.perform(get("/authenticated")
+				.with(bearerToken("token")))
+				.andExpect(status().isOk())
+				.andExpect(content().string("custom-test-subject"));
+	}
+
+	@Test
+	public void requestWhenJwtDecoderAndJwkSetUriConfiguredThenJwkSetUriTakesPrecedence()
```
jgrandja: I'm wondering if this makes sense? re: `jwkSetUri` takes precedence over `JwtDecoder`. If the user supplied both of these configurations than we should at least log a warning message or something alerting the user. The user may wonder why their custom `JwtDecoder` is not wired?

rwinch: This is always the case with Spring Security. If you provide an explicit configuration via the DSL the bean is ignored. This makes sense since the bean is a default value.

jzheaux: I asked @rwinch a similar question a while back about conflicting configurations (different scenario, though), so I'd like to hear his thoughts here.

At least, I don't think we should error, but I can see value in a logs that explain why we made this or that wiring decision. I know for my part, the "--debug" setting in Spring Boot has often been helpful to me when debugging why a given @Conditional-wiring passed or failed.

jgrandja: @rwinch The test (and user) configures `jwkSetUri` via the DSL and registers a `JwtDecoder` `@Bean`. As a user, if my `JwtDecoder` `@Bean` is not leveraged than I'm left wondering why? This scenario should at least alert the user up-front about the configuration?

rwinch: I disagree. Should we warn a user if they invoke the same method twice? Should we warn a user if the configure a dsl and then disable it afterwards? This leads to too much information.

This style of configuration has worked very successfully for Spring Security.

jzheaux: I think a message like this would be valuable at a debug level, similar to what Spring Boot does with the --debug flag, indicating why a given wiring decision was made.

That said, if we don't already have an established pattern for alerting folks to wiring choices (or that our pattern is that we don't alert them), then an isolated message like this would kind of come from nowhere. Probably better to consider the value of the pattern as a whole first (and at a later time).

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
jzheaux: fyi - https://github.com/spring-projects/spring-security/issues/5538

rwinch:
Thanks for the PR @jzheaux! This is merged into master

Issue #5535, opened 2018-07-18
Support OAuth2 redirect_uri for Forwarded Header


mhyeon-lee:
### Summary

When generating the `redirect_uri` from the DefaultOAuth2AuthorizationRequestResolver, apply the `Forwarded` Header to `{baseUrl}`.

Also,  shouldProcessAuthorizationResponse of OAuth2AuthorizationCodeGrantFilter should be compared to redirect_uri based on Forwarded Header application.

### Actual Behavior

- request: http://localhost/oauth2/authorization/google
- header: Forwarded host=192.168.0.1
- redirect_uri: http://localhost/login/oauth2/code/google

### Expected Behavior

- redirect_uri: http://192.168.0.1/login/oauth2/code/google

### Sample

The following test fails.

``` java
@Test
public void resolveWhenAuthorizationRequestRedirectUriTemplatedThenRedirectUriForwardedIPv4HostHeader() {
	ClientRegistration clientRegistration = this.registration2;
	String requestUri = this.authorizationRequestBaseUri + "/" + clientRegistration.getRegistrationId();
	MockHttpServletRequest request = new MockHttpServletRequest("GET", requestUri);
	request.setServletPath(requestUri);
	request.setScheme("http");
	request.setServerName("localhost");
	request.setServerPort(80);
	request.addHeader("Forwarded", "host=192.168.0.1");

	OAuth2AuthorizationRequest authorizationRequest = this.resolver.resolve(request);
	assertThat(authorizationRequest.getRedirectUri()).isNotEqualTo(
			clientRegistration.getRedirectUriTemplate());
	assertThat(authorizationRequest.getRedirectUri()).isEqualTo(
			"http://192.168.0.1/login/oauth2/code/" + clientRegistration.getRegistrationId());
}
```

jgrandja:
@mhyeon-lee Thank you very much for your recent contributions!

As an FYI, the recommended approach to handle `X-Forwarded-*` headers is to use the [ForwardedHeaderFilter](https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/web/filter/ForwardedHeaderFilter.java).

Please see these resources for more detail.

[Forwarded headers](https://github.com/spring-projects/spring-framework/wiki/Upgrading-to-Spring-Framework-5.x#forwarded-headers)
[Centralize handling of "Forwarded" headers to ForwardedHeaderFilter](https://jira.spring.io/browse/SPR-16668)

[...]

PR #5536, opened 2018-07-18
Support oauth2 redirecturi apply forwarded header

mhyeon-lee:
When generating the redirect_uri from the DefaultOAuth2AuthorizationRequestResolver, apply the Forwarded Header to {baseUrl}.

Also, shouldProcessAuthorizationResponse of OAuth2AuthorizationCodeGrantFilter should be compared to redirect_uri based on Forwarded Header application.

#5535


jgrandja:
Closing as per [comment](https://github.com/spring-projects/spring-security/issues/5535#issuecomment-406001453)

PR #5550, opened 2018-07-20
Bearer Token Resolver Configuration

jzheaux:
This introduces #bearerTokenResolver(BearerTokenResolver) to the
Resource Server DSL, allowing users to configure the resolver to allow
the access token as part of the request body or a query parameter. It
also allows the user to replace the resolver with a completely custom
one.

This also introduces the same ability by exposing a bean of type
BearerTokenResolver

Fixes: gh-5479

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/OAuth2Configurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/OAuth2Configurer.java
@@ -16,6 +16,7 @@
 package org.springframework.security.config.annotation.web.configurers.oauth2;
 
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.ApplicationContext;
 import org.springframework.security.config.annotation.ObjectPostProcessor;
 import org.springframework.security.config.annotation.web.HttpSecurityBuilder;
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
@@ -96,7 +97,8 @@ public final class OAuth2Configurer<B extends HttpSecurityBuilder<B>>
 	}
 
 	private void initResourceServerConfigurer() {
-		this.resourceServerConfigurer = new OAuth2ResourceServerConfigurer<>();
+		ApplicationContext context = getBuilder().getSharedObject(ApplicationContext.class);
+		this.resourceServerConfigurer = new OAuth2ResourceServerConfigurer<>(context);
 		this.resourceServerConfigurer.setBuilder(this.getBuilder());
 		this.resourceServerConfigurer.addObjectPostProcessor(this.objectPostProcessor);
 	}
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
@@ -47,6 +47,7 @@ import org.springframework.util.Assert;
  * The following configuration options are available:
  *
  * <ul>
+ * <li>{@link #bearerTokenResolver(BearerTokenResolver)} - customizes how to resolve a bearer token from the request</li>
  * <li>{@link #jwt()} - enables Jwt-encoded bearer token support</li>
  * </ul>
  *
@@ -99,7 +100,11 @@ import org.springframework.util.Assert;
 public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<H>> extends
 		AbstractHttpConfigurer<OAuth2ResourceServerConfigurer<H>, H> {
 
-	private BearerTokenResolver bearerTokenResolver = new DefaultBearerTokenResolver();
+	private final ApplicationContext context;
+
+	private BearerTokenResolver bearerTokenResolver;
+	private JwtConfigurer jwtConfigurer;
+
 	private BearerTokenRequestMatcher requestMatcher = new BearerTokenRequestMatcher();
 
 	private BearerTokenAuthenticationEntryPoint authenticationEntryPoint
@@ -108,12 +113,18 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 	private BearerTokenAccessDeniedHandler accessDeniedHandler
 			= new BearerTokenAccessDeniedHandler();
 
-	private JwtConfigurer jwtConfigurer;
+	public OAuth2ResourceServerConfigurer(ApplicationContext context) {
+		this.context = context;
+	}
+
+	public OAuth2ResourceServerConfigurer<H> bearerTokenResolver(BearerTokenResolver bearerTokenResolver) {
+		this.bearerTokenResolver = bearerTokenResolver;
+		return this;
+	}
 
 	public JwtConfigurer jwt() {
 		if ( this.jwtConfigurer == null ) {
-			ApplicationContext context = this.getBuilder().getSharedObject(ApplicationContext.class);
-			this.jwtConfigurer = new JwtConfigurer(context);
+			this.jwtConfigurer = new JwtConfigurer(this.context);
 		}
 
 		return this.jwtConfigurer;
@@ -231,13 +242,20 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 		csrf.ignoringRequestMatchers(this.requestMatcher);
 	}
 
-	private BearerTokenResolver getBearerTokenResolver() {
+	BearerTokenResolver getBearerTokenResolver() {
+		if ( this.bearerTokenResolver == null ) {
+			if ( this.context.getBeanNamesForType(BearerTokenResolver.class).length > 0 ) {
+				this.bearerTokenResolver = this.context.getBean(BearerTokenResolver.class);
+			} else {
+				this.bearerTokenResolver = new DefaultBearerTokenResolver();
+			}
+		}
+
 		return this.bearerTokenResolver;
 	}
 
 	private static final class BearerTokenRequestMatcher implements RequestMatcher {
-		private BearerTokenResolver bearerTokenResolver
-				= new DefaultBearerTokenResolver();
+		private BearerTokenResolver bearerTokenResolver;
 
 		@Override
 		public boolean matches(HttpServletRequest request) {
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
@@ -25,6 +25,7 @@ import org.springframework.security.config.annotation.web.configurers.AbstractHt
 import org.springframework.security.config.annotation.web.configurers.CsrfConfigurer;
 import org.springframework.security.config.annotation.web.configurers.ExceptionHandlingConfigurer;
 import org.springframework.security.config.http.SessionCreationPolicy;
+import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
 import org.springframework.security.oauth2.jwt.JwtDecoder;
 import org.springframework.security.oauth2.jwt.NimbusJwtDecoderJwkSupport;
 import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationProvider;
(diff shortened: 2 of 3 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/OAuth2Configurer.java
```diff
@@ -96,7 +97,8 @@ private void initClientConfigurer() {
 	}
 
 	private void initResourceServerConfigurer() {
-		this.resourceServerConfigurer = new OAuth2ResourceServerConfigurer<>();
+		ApplicationContext context = getBuilder().getSharedObject(ApplicationContext.class);
+		this.resourceServerConfigurer = new OAuth2ResourceServerConfigurer<>(context);
```
jzheaux: @jgrandja you and I were talking about trying to keep the ApplicationContext resolution approach consistent across OAuth2 configurers; however with this set of tests for bearer token resolver, it still seemed the cleanest approach. Let me know if you see one you'd prefer more.

jgrandja: For `oauth2-client` we use `OAuth2ClientConfigurerUtils` to get `@Bean`'s from the context - passing in the builder (http). This centralizes the logic for `@Bean` lookup so it can be reused across configurer's since there is more than one for `oauth2-client`. I'm not sure if there will be more than one for resource server. Either way, I think it's fine leaving as is.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
```diff
@@ -564,7 +667,7 @@ public void requestWhenCustomJwtDecoderExposedAsBeanThenUsed()
 	@Test
 	public void getJwtDecoderWhenConfiguredWithDecoderAndJwkSetUriThenLastOneWins() {
 		OAuth2ResourceServerConfigurer.JwtConfigurer jwtConfigurer =
-				new OAuth2ResourceServerConfigurer().new JwtConfigurer(null);
+				new OAuth2ResourceServerConfigurer(null).jwt();
```
jgrandja: We should not pass `null` context.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
```diff
@@ -574,7 +677,7 @@ public void getJwtDecoderWhenConfiguredWithDecoderAndJwkSetUriThenLastOneWins()
 		assertThat(jwtConfigurer.getJwtDecoder()).isEqualTo(decoder);
 
 		jwtConfigurer =
-				new OAuth2ResourceServerConfigurer().new JwtConfigurer(null);
+				new OAuth2ResourceServerConfigurer(null).jwt();
```
jgrandja: We should not pass `null` context.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
```diff
@@ -526,6 +528,107 @@ public void requestWhenSessionManagementConfiguredThenUserConfigurationOverrides
...
+		GenericWebApplicationContext context = new GenericWebApplicationContext();
+		context.registerBean("resolverOne", BearerTokenResolver.class, () -> resolverBean);
+		context.registerBean("resolverTwo", BearerTokenResolver.class, () -> resolverBean);
+		this.spring.context(context).autowire();
+
+		OAuth2ResourceServerConfigurer oauth2 = new OAuth2ResourceServerConfigurer(context);
+
+		oauth2.bearerTokenResolver(resolver);
+
+		assertThat(oauth2.getBearerTokenResolver()).isEqualTo(resolver);
```
jgrandja: I see why you made `getBearerTokenResolver()` package private. However, I'd rather avoid opening up methods so it can work for testing.

This can be tested by issuing a request and then `verify()` or `verifyZeroInteractions()` to assert which resolver is in effect.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
```diff
@@ -526,6 +528,107 @@ public void requestWhenSessionManagementConfiguredThenUserConfigurationOverrides
 		assertThat(result.getRequest().getSession(false)).isNotNull();
 	}
 
+	// -- custom bearer token resolver
+
+	@Test
+	public void requestWhenBearerTokenResolverAllowsRequestBodyThenEitherHeaderOrRequestBodyIsAccepted()
```
jgrandja: Please add a test where the token is in the header and body AND another test where the token is in header and query param

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
```diff
@@ -526,6 +528,107 @@ public void requestWhenSessionManagementConfiguredThenUserConfigurationOverrides
...
+		BearerTokenResolver resolverBean = mock(BearerTokenResolver.class);
+
+		GenericWebApplicationContext context = new GenericWebApplicationContext();
+		context.registerBean("resolverOne", BearerTokenResolver.class, () -> resolverBean);
+		context.registerBean("resolverTwo", BearerTokenResolver.class, () -> resolverBean);
+		this.spring.context(context).autowire();
+
+		OAuth2ResourceServerConfigurer oauth2 = new OAuth2ResourceServerConfigurer(context);
+
+		assertThatCode(() -> oauth2.getBearerTokenResolver())
```
jgrandja: Related to previous comment on making `getBearerTokenResolver()` private...this can be tested by initializing the context which will initialize the configurer and the exception and you can `assertThatThrownBy()`

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
```diff
@@ -526,6 +528,107 @@ public void requestWhenSessionManagementConfiguredThenUserConfigurationOverrides
...
+		BearerTokenResolver resolverBean = mock(BearerTokenResolver.class);
+
+		GenericWebApplicationContext context = new GenericWebApplicationContext();
+		context.registerBean(BearerTokenResolver.class, () -> resolverBean);
+		this.spring.context(context).autowire();
+
+		OAuth2ResourceServerConfigurer oauth2 = new OAuth2ResourceServerConfigurer(context);
+		oauth2.bearerTokenResolver(resolver);
+
+		assertThat(oauth2.getBearerTokenResolver()).isEqualTo(resolver);
```
jgrandja: This can be tested by issuing a request and then `verify()` or `verifyZeroInteractions()` to assert which resolver is in effect.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
```diff
@@ -526,6 +528,107 @@ public void requestWhenSessionManagementConfiguredThenUserConfigurationOverrides
...
+	}
+
+	@Test
+	public void getBearerTokenResolverWhenNoResolverSpecifiedThenTheDefaultIsUsed() {
+		ApplicationContext context =
+				this.spring.context(new GenericWebApplicationContext()).getContext();
+
+		OAuth2ResourceServerConfigurer oauth2 = new OAuth2ResourceServerConfigurer(context);
+
+		assertThat(oauth2.getBearerTokenResolver()).isInstanceOf(DefaultBearerTokenResolver.class);
```
jgrandja: This can be tested by issuing a request and then ensuring the request works and therefore the default resolver is set or else an error would happen.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -231,13 +242,20 @@ private void registerDefaultCsrfOverride(H http) {
 		csrf.ignoringRequestMatchers(this.requestMatcher);
 	}
 
-	private BearerTokenResolver getBearerTokenResolver() {
+	BearerTokenResolver getBearerTokenResolver() {
```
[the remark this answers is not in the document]
jzheaux: This needs to be default so that unit tests can access it. What this comes down to is how we prefer to do unit tests for configurers. In the PR for JwtDecoder, the same tradeoff was made in order to allow for corner cases to be confirmed (as unit tests) instead of as performing a full MockMvc roundtrip.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
jzheaux: Can you elaborate on how this confirms the intended behavior? That is, we want to confirm that the correct implementation is picked. We won't actually invoke the mock without performing some kind of mock mvc roundtrip, which, this being a unit test, might be a bit odd.

We did the same thing with `getJwtDecoder` in `JwtConfigurer` to achieve what appears to me to be following the same reasoning, e.g. corner case should be a unit test, so we make the method package-private so the unit test can confirm. I agree that bumping up the access level of a method for a unit test is worth thinking twice about, but I don't see another way to do it other than reflection or turning it into an integration test.

Now, we could look at this becoming an integration test; however, that seems to deviate from the original thrust of this set of tests. (Again, for context, take a look at the tests for JwtConfigurer and see that some are integration tests and some are unit tests.)

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
jzheaux: We have unit tests for these. For my education, how do you gauge what use cases need both an integration test as well as a unit test?

Where my mind goes right now is these integration tests confirm that things got wired correctly and behave accordingly, and unit tests confirm that, given XYZ arbitrary wirings, the contract is correctly upheld.

I don't have a problem adding more tests, I just want to learn from you how you prioritize your own efforts with things like this.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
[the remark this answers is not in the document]
jgrandja: What's the disadvantage of doing MockMvc roundtrip? I see `SecurityConfigurers` more like integration testing instead of unit testing.

> What this comes down to is how we prefer to do unit tests for configurers

Can you clarify what you mean by "we prefer"? Did I miss a convo you and @rwinch had?

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
jzheaux: Also, this specific recommendation actually uncovered a bug (awesome!) which the unit test would not have uncovered, which is why I'm interested in your thinking around it.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
jzheaux: Thank you for catching that! Intended clean up that I missed.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
jgrandja: This class mixes unit tests and integration tests which we typically separate in different classes. Also, `SecurityConfigurer`s are integration testing. Are there other configuers that do unit testing with the exception of `JwtConfigurer`?

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
jgrandja: Sorry I missed that - the unit tests cover the scenarios I suggested. Those tests should be in the units tests.

> Where my mind goes right now is these integration tests confirm that things got wired correctly and behave accordingly

Agreed, the integration tests should confirm things got wired correctly and behave as expected.

Issue #5496, opened 2018-07-10
Make Bearer Token Resolution Configurable through DSL


jzheaux:
Adding bearer token resolver configuration to the DSL would make it easy for users to configure Resource Server to read access tokens from places other than the header, e.g. 

```java
protected void configure(HttpSecurity http) {
    DefaultBearerTokenResolver tokenResolver = new DefaultBearerTokenResolver();
    tokenResolver.setAllowFormEncodedBodyParameter(true);

    http
        .oauth2()
            .resourceServer()
                .bearerTokenResolver(tokenResolver);
}
```
Also, users get the added benefit of being able to fully customize the way the token is specified, using a custom token resolver implementation.

PR #5561, opened 2018-07-23
 Bearer Token Resolver Configuration

jzheaux:
This introduces #bearerTokenResolver(BearerTokenResolver) to the
Resource Server DSL, allowing users to configure the resolver to allow
the access token as part of the request body or a query parameter. It
also allows the user to replace the resolver with a completely custom
one.
    
This also introduces the same ability by exposing a bean of type
BearerTokenResolver
    
Fixes: gh-5496


jgrandja:
Thanks for the PR @jzheaux. This is now in master via 6a45ecd

Issue #5341, opened 2018-05-11
Validate @EnableGlobalMethodSecurity usage


jzheaux:
### Summary

It is possible and even easy to accidentally misconfigure @EnableGlobalMethodSecurity to actually do nothing, e.g.

```java
@EnableGlobalMethodSecurity
```
The above will not secure any methods because none of `prePostEnabled`, `securedEnabled`, or `jsr250Enabled` is set to `true`, e.g.

```java
@EnableGlobalMethodSecurity(prePostEnabled=true)
```
Because Spring Security supports multiple @EnableGlobalMethodSecurity annotations, these need to continue to default to false.

But we could add validation at runtime that would warn that in the composition of all global method configuration, no annotation support was actually activated.

PR #5563, opened 2018-07-24
Validate @EnableGlobalMethodSecurity usage

ayudovin:
adding warn that in the composition of all global method configuration, no annotation support was actually activated.
this [improvement](https://github.com/spring-projects/spring-security/issues/5341)


jzheaux:
@ayudovin Thanks for the PR! 

Instead of logging a warning, let's throw an `IllegalStateException`. While it is nice that a warning makes this a passive update for folks, misconfiguration like this is a security vulnerability, and we want to make sure that users fix it.

Also, in that case, it should be possible to create a test around it. Would you mind adding one?

ayudovin:
@jzheaux , I have added changes by suggestion.

jzheaux:
Thanks @ayudovin for the PR! This has been merged to master with 1e864ad7645fcf6d66edd714cb4faf32f76c0716.

Also, I did just a bit of polish with 73c1abbba03e1a9e799a25bb7a6528a19eb6dd36.

ayudovin:
@jzheaux, thank you!

PR #5580, opened 2018-07-26
Fix typo

izeye:
This PR fixes typos in Javadoc as it looks a copy and paste error from the below.


jgrandja:
Thanks for that correction @izeye. This is now in master.

PR #5627, opened 2018-08-01
Add support for client_credentials grant

jgrandja:
Fixes gh-4982

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultClientCredentialsTokenResponseClient.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultClientCredentialsTokenResponseClient.java
@@ -93,42 +93,14 @@ public class DefaultClientCredentialsTokenResponseClient implements OAuth2Access
 
 		Assert.notNull(clientCredentialsGrantRequest, "clientCredentialsGrantRequest cannot be null");
 
-		ClientRegistration clientRegistration = clientCredentialsGrantRequest.getClientRegistration();
-
-		// Headers
-		HttpHeaders headers = new HttpHeaders();
-		headers.setAccept(Collections.singletonList(MediaType.APPLICATION_JSON));
-		headers.setContentType(MediaType.APPLICATION_FORM_URLENCODED);
-		if (ClientAuthenticationMethod.BASIC.equals(clientRegistration.getClientAuthenticationMethod())) {
-			headers.setBasicAuth(clientRegistration.getClientId(), clientRegistration.getClientSecret());
-		}
-
-		// Form parameters
-		MultiValueMap<String, String> formParameters = new LinkedMultiValueMap<>();
-		formParameters.add(OAuth2ParameterNames.GRANT_TYPE, clientCredentialsGrantRequest.getGrantType().getValue());
-		if (!CollectionUtils.isEmpty(clientRegistration.getScopes())) {
-			formParameters.add(OAuth2ParameterNames.SCOPE,
-					StringUtils.collectionToDelimitedString(clientRegistration.getScopes(), " "));
-		}
-		if (ClientAuthenticationMethod.POST.equals(clientRegistration.getClientAuthenticationMethod())) {
-			formParameters.add(OAuth2ParameterNames.CLIENT_ID, clientRegistration.getClientId());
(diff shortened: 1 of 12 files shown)


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultClientCredentialsTokenResponseClient.java
```diff
@@ -0,0 +1,258 @@
...
+
+	@Override
+	public OAuth2AccessTokenResponse getTokenResponse(OAuth2ClientCredentialsGrantRequest clientCredentialsGrantRequest)
+			throws OAuth2AuthenticationException {
+
+		Assert.notNull(clientCredentialsGrantRequest, "clientCredentialsGrantRequest cannot be null");
+
+		ClientRegistration clientRegistration = clientCredentialsGrantRequest.getClientRegistration();
+
+		// Headers
```
rwinch: This is a pretty large method. I think each of these block with a comment above them make sense to be in a separate private method

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/method/annotation/OAuth2AuthorizedClientArgumentResolver.java
```diff
@@ -56,15 +64,22 @@
...
 	 * Constructs an {@code OAuth2AuthorizedClientArgumentResolver} using the provided parameters.
 	 *
-	 * @param authorizedClientRepository the authorized client repository
+	 * @param clientRegistrationRepository the repository of client registrations
+	 * @param authorizedClientRepository the repository of authorized clients
 	 */
-	public OAuth2AuthorizedClientArgumentResolver(OAuth2AuthorizedClientRepository authorizedClientRepository) {
+	public OAuth2AuthorizedClientArgumentResolver(ClientRegistrationRepository clientRegistrationRepository,
+													OAuth2AuthorizedClientRepository authorizedClientRepository) {
+		Assert.notNull(clientRegistrationRepository, "clientRegistrationRepository cannot be null");
```
rwinch: Is it true that users will always need a ClientRegistrationRepository or is that only necessary for Client Credentials? If it is not always required, I prefer to overload the constructor with optional parameters last

rwinch: I'm even wondering if makes sense to have a separate resolver for the client credentials? If nothing else, the resolveArgument should probably be broken up into more than one method (and likely by the grant type)

jgrandja: `ClientRegistrationRepository` is required as I need to determine the `clientRegistration.getAuthorizationGrantType()` before I can delegate to the correct logic.

jgrandja: > I'm even wondering if makes sense to have a separate resolver for the client credentials? If nothing else, the resolveArgument should probably be broken up into more than one method (and likely by the grant type)

The logic for `authorization_code` is this:

```

		if (AuthorizationGrantType.AUTHORIZATION_CODE.equals(clientRegistration.getAuthorizationGrantType())) {
			throw new ClientAuthorizationRequiredException(clientRegistrationId);
		}
```

I don't see  much value in breaking this up into different methods? Do you find this unreadable?

rwinch: I do find it unreadable. As soon as you start adding block comments that should be an indicator that the method is too large. Instead change the block into a method that reads like the block comment.

jgrandja: I think you meant this comment for the `getTokenResponse()` method? I am breaking that up.

rwinch:
This has been rebased and merged in via 952743269d9abeb25daa1b8aea16ef1a1d382d28

Issue #5523, opened 2018-07-16
Support auto-configuration of JwkSet Uri


jgrandja:
We should provide support for auto-configuration of the JwkSet Uri via the supplied `issuer` claim that can be used to discover [Provider Metadata](http://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata) via OpenID Connect Discovery. The Provider Metadata contains the `jwks_uri` attribute which would be used to supply the `JwtDecoder`.

There is existing logic for OpenID Connect Discovery in `OidcConfigurationProvider`. Please re-use the provider metadata discovery logic for Resource Server.

jgrandja:
After this feature has been merged, we need to log a ticket with the Boot team to support the `issuer` configuration property.

jgrandja:
@jzheaux As far as approach on implementing this feature, @rwinch and I spoke and agree that it should be implemented similar to how `OidcConfigurationProvider.issuer()` is implemented for `ClientRegistration` metadata discovery.

For example: 
`JwtDecoder OidcConfigurationProvider.issuer()`

The returned `JwtDecoder` would be configured with the `jwks_uri` from the discovered Provider Configuration and than supplied to `resourceServer().jwt().decoder(jwtDecoder)`.

**NOTE:** We may have to re-factor the existing `OidcConfigurationProvider` to support both configurations of `ClientRegistration` and `JwtDecoder`. However, we're also debating on keeping these in separate classes - client and resource server specific configs. It's difficult to tell at this point until the work get's started so any suggestions you have let's start that dialogue in a PR.

rwinch:
>  and than supplied to resourceServer().jwt().decoder(jwtDecoder).

Spring Security would not configure on the DSL. User's (and Boot) could create a JwtDecoder Bean using the factory and then it would be used by the DSL by default. Alternatively, user's would be able to explicitly inject it via the decoder(..) method

jgrandja:
Related #5584

Issue #5603, opened 2018-07-30
Allow configurable RestOperations for NimbusJwtDecoderJwkSupport


jgrandja:
`NimbusJwtDecoderJwkSupport` uses Nimbus's HTTP Client internally to retrieve the `JwkSet` resource.

We should re-factor to use `RestOperations` instead and additionally expose a `NimbusJwtDecoderJwkSupport.setRestOperations(restOperations)` to allow user's the ability to supply a pre-configured `RestOperations` if needed.

Related #5601

mnovoseltsev:
Hi @jgrandja , I wanted to set custom rest operations for `NimbusJwtDecoderJwkSupport`, but it appears that `NimbusJwtDecoderJwkSupport` instance is created  inside of the private method: `OidcAuthorizationCodeAuthenticationProvider#getJwtDecoder`. So it is impossible to use `NimbusJwtDecoderJwkSupport#setRestOperations` method.

For more details - https://github.com/spring-projects/spring-security/issues/6255

kumarimanjari:
Hi, I have a problem in this area with Spring boot starter oAuth2 client version 3.2.0 which internally uses Spring Security 6.2.0, I have below SecurityConfig file - 

I am getting below error - 
`[invalid_id_token] An error occurred while attempting to decode the Jwt: Couldn't retrieve remote JWK set: org.springframework.web.client.ResourceAccessException: I/O error on GET request for "JWK SET URI": issuer-uri`
Can someone assist on this issue ?

`required imports

@Configuration
@EnableWebSecurity
@RequiredArgsConstructor
public class SecurityConfig {

	@NonNull
	private PropertyUtil propertyUtil;

	@Bean
	public SecurityFilterChain clientSecurityFilterChain(HttpSecurity http) throws Exception {
		http
				.authorizeHttpRequests(requests -> requests.requestMatchers("/", "/api/**").authenticated()
						.anyRequest().permitAll())
				.oauth2Login(Customizer.withDefaults());
		return http.build();
	}

	@Bean
	public OAuth2AccessTokenResponseClient<OAuth2AuthorizationCodeGrantRequest> authorizationCodeAccessTokenResponseClient() {
		DefaultAuthorizationCodeTokenResponseClient accessTokenResponseClient = new DefaultAuthorizationCodeTokenResponseClient();
		accessTokenResponseClient.setRestOperations(restTemplate());

		return accessTokenResponseClient;
	}

	@Bean
[...]

Issue #5584, opened 2018-07-26
Consider renaming OidcConfigurationProvider


jgrandja:
We should consider renaming `OidcConfigurationProvider` to `OidcProviderConfiguration` to align more closely to the spec -> [OpenID Provider Configuration](http://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig)

rwinch:
or even something that includes ClientRegistration in it too

jgrandja:
Resolved via 742716f

PR #5628, opened 2018-08-01
Auto-configure JwtDecoder via OpenId Configuration

jzheaux:
Adding OpenIdConfigurationJwtDecoderProvider which takes an issuer
and derives from it the jwk set uri via a call to
.well-known/openid-configuration

Fixes: gh-5523

### DIFF
--- a/oauth2/oauth2-jose/spring-security-oauth2-jose.gradle
+++ b/oauth2/oauth2-jose/spring-security-oauth2-jose.gradle
@@ -6,6 +6,7 @@ dependencies {
 	compile springCoreDependency
 	compile 'com.nimbusds:nimbus-jose-jwt'
 
+	optional 'com.nimbusds:oauth2-oidc-sdk'
 	optional 'io.projectreactor:reactor-core'
 	optional 'org.springframework:spring-webflux'
 
--- a/config/src/main/resources/org/springframework/security/config/spring-security-5.1.xsd
+++ b/config/src/main/resources/org/springframework/security/config/spring-security-5.1.xsd
@@ -2467,17 +2467,19 @@
   </xs:attributeGroup>
   <xs:element name="feature-policy">
       <xs:annotation>
-          <xs:documentation>Adds support for Feature Policy</xs:documentation>
+         <xs:documentation>Adds support for Feature Policy
+                </xs:documentation>
       </xs:annotation>
       <xs:complexType>
-          <xs:attributeGroup ref="security:feature-options.attlist"/>
+         <xs:attributeGroup ref="security:feature-options.attlist"/>
       </xs:complexType>
-  </xs:element>
+   </xs:element>
   <xs:attributeGroup name="feature-options.attlist">
       <xs:attribute name="policy-directives" type="xs:token">
-          <xs:annotation>
-              <xs:documentation>The security policy directive(s) for the Feature-Policy header.</xs:documentation>
-          </xs:annotation>
+         <xs:annotation>
+            <xs:documentation>The security policy directive(s) for the Feature-Policy header.
+                </xs:documentation>
+         </xs:annotation>
       </xs:attribute>
   </xs:attributeGroup>
   <xs:element name="cache-control">
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoders.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoders.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2002-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.jwt;
+
+import com.nimbusds.oauth2.sdk.ParseException;
+import com.nimbusds.openid.connect.sdk.op.OIDCProviderMetadata;
+
+import org.springframework.security.oauth2.core.OAuth2TokenValidator;
+import org.springframework.web.client.RestTemplate;
+
+/**
+ * Allows creating a {@link JwtDecoder} from an
+ * <a href="https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig">OpenID Provider Configuration</a>.
+ *
+ * @author Josh Cummings
+ * @since 5.1
+ */
+public final class JwtDecoders {
+
+	/**
+	 * Creates a {@link JwtDecoder} using the provided
+	 * <a href="http://openid.net/specs/openid-connect-core-1_0.html#IssuerIdentifier">Issuer</a> by making an
+	 * <a href="https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationRequest">OpenID Provider
+	 * Configuration Request</a> and using the values in the
+	 * <a href="https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationResponse">OpenID
+	 * Provider Configuration Response</a> to initialize the {@link JwtDecoder}.
+	 *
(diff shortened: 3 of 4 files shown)


oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/OpenIdConfigurationJwtDecoderProvider.java
```diff
@@ -0,0 +1,75 @@
...
+import org.springframework.web.client.RestTemplate;
+
+/**
+ * Allows creating a {@link JwtDecoder} from an
+ * <a href="https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig">OpenID Provider Configuration</a>.
+ *
+ * @author Josh Cummings
+ * @since 5.1
+ */
+public final class OpenIdConfigurationJwtDecoderProvider {
```
jgrandja: This class needs to be in the `config` module just like `OidcConfigurationProvider` is.

Also, the name of this class restricts it to `JwtDecoder` configuration only. I think we need to rename it to be more generic so it makes sense when we need other provider configuration and associated static factory methods.

The common naming convention is to prefix with `Oidc` not `OpenId`

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/OpenIdConfigurationJwtDecoderProvider.java
```diff
@@ -0,0 +1,75 @@
...
+	 * Configuration Request</a> and using the values in the
+	 * <a href="https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationResponse">OpenID
+	 * Provider Configuration Response</a> to initialize the {@link JwtDecoder}.
+	 *
+	 * @param issuer the <a href="http://openid.net/specs/openid-connect-core-1_0.html#IssuerIdentifier">Issuer</a>
+	 * @return a {@link JwtDecoder} that was initialized by the OpenID Provider Configuration.
+	 */
+	public static JwtDecoder issuer(String issuer) {
+		String openidConfiguration = getOpenidConfiguration(issuer);
+		OIDCProviderMetadata metadata = parse(openidConfiguration);
```
jgrandja: The code in here is duplicated from `OidcConfigurationProvider`, the `getOpenidConfiguration` and `parse` methods. We should try to reuse.

oauth2/oauth2-jose/spring-security-oauth2-jose.gradle
```diff
@@ -6,6 +6,7 @@ dependencies {
 	compile springCoreDependency
 	compile 'com.nimbusds:nimbus-jose-jwt'
 
+	optional 'com.nimbusds:oauth2-oidc-sdk'
```
jgrandja: This dependency does not make sense. However, this will be removed when moving `OpenIdConfigurationJwtDecoderProvider` to the config module

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderJwkSupport.java
```diff
@@ -111,6 +111,10 @@ public Jwt decode(String token) throws JwtException {
 		throw new JwtException("Unsupported algorithm of " + jwt.getHeader().getAlgorithm());
 	}
 
+	public String getJwkSetUrl() {
```
jgrandja: This is not used anywhere. We should remove it.

jgrandja:
@jzheaux Also, please consider this comment when re-working this PR

> NOTE: We may have to re-factor the existing OidcConfigurationProvider to support both configurations of ClientRegistration and JwtDecoder. However, we're also debating on keeping these in separate classes - client and resource server specific configs. It's difficult to tell at this point until the work get's started so any suggestions you have let's start that dialogue in a PR.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderJwkSupport.java
jzheaux: @jgrandja please consult one of the unit tests, which confirms that the configuration took effect. The two alternatives are to use reflection to confirm this setting or to perform some heavier configuration to mock this url endpoint in addition to the openid-configuration endpoint already mocked.

I'm okay with the first option, though not precisely ideal, though the second option I believe increases the scope of the test too much.

What would you do in this situation?

jzheaux: Also, I believe this situation would be simplified if we could expose RestOperations as configurable here as well as in NimbusJwtDecoderJwkSupport. Then, this test could provide a mock RestOperations and alleviate the need to mock an underlying server.

That being the case, I'm also okay with leaving the unit test in a suboptimal state (calling reflectively a private method) in anticipation of the class under test being able to more easily configure mock endpoints.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/OpenIdConfigurationJwtDecoderProvider.java
jgrandja: > Can you clarify why the existing `OidcConfigurationProvider` is in the config module

This is considered Java Config, just like `OAuth2ResourceServerConfigurer` and `OAuth2ClientConfigurer`, hence it belonging in `config` module

> the only generic name that makes sense to me is `OidcConfigurationProvider`

Yeah same for me. Although we have this [ticket](https://github.com/spring-projects/spring-security/issues/5584) to rename to `OidcProviderConfiguration`

>  What are your thoughts on two classes of the same name, in different packages

I was thinking the same and initially it makes sense but at the same time having users (or Boot) to fully qualify the class name when both are used (Client and Resource Server) would be a pain. I'm not sure what other name we can use at the moment.

> Or what rules do we have for moving the existing OidcConfigurationProvider into a more generic package like o.s.s.config.oauth2.oidc.OidcConfigurationProvider which could then house both methods?

Housing both methods in `OidcConfigurationProvider` would require both dependencies - client and resource server - which is not good especially if the user is only using client for example. 

However, I believe @rwinch did suggest this strategy (housing in one config class) when we spoke? Maybe he can clarify this further as I'm not sure if I understood him correctly.

> I we can combine the two, then I might like something named OidcDiscoveryConfigurationProvider

I think we should use the name `OidcProviderConfiguration` to align with the spec. Or at least have that in the class name.

I do like your implementation approach on caching the `OIDCProviderMetadata` as per

> public static OidcConfigurationProvider issuer(String issuer)

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderJwkSupport.java
jgrandja: I must be missing something? I don't see any references to `getJwkSetUrl()` after searching for usages within IDE.

> I believe this situation would be simplified if we could expose RestOperations as configurable here as well as in NimbusJwtDecoderJwkSupport

This is coming in [#5603](https://github.com/spring-projects/spring-security/issues/5603)

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/OpenIdConfigurationJwtDecoderProvider.java
jzheaux: > I think we should use the name OidcProviderConfiguration to align with the spec. Or at least have that in the class name.

My bad, I had misread the spec when I saw "Discovery" in the title [1], forgetting that we are just focused on the configuration section.

[1] - https://openid.net/specs/openid-connect-discovery-1_0.html#IssuerDiscovery

jgrandja:
@jzheaux Looks like you included the commit **Jwt Claim Validation** in this PR. Can you please remove that and re-submit.

jzheaux:
Regarding including the Jwt Claim Validation commit, that is only temporary. It's a little messy, but since this is dependent on that, I rebased off of it so that the PR would compile. Once we merge the validation PR, I'll remove that commit here.

oauth2/oauth2-jose/spring-security-oauth2-jose.gradle
jzheaux: For the record, @jgrandja and I discussed, and this will stay here in `oauth2-jose` since `JwtDecoder`, the security component being instantiated by `JwtDecoders`, is in `oauth2-jose`.

jzheaux:
@rwinch This is ready for your review.

rwinch:
Thanks this is merged into master

PR #5632, opened 2018-08-02
Fix typo

### LINKED ISSUE #5579: there' sa small typo `funciton` in the `@deprecated` tag in `LdapShaPasswordEncoder.java`
* @deprecated Digest based password encoding is not considered secure. Instead use an
 * adaptive one way funciton like BCryptPasswordEncoder, Pbkdf2PasswordEncoder, or
 * SCryptPasswordEncoder. Even better use {@link DelegatingPasswordEncoder} which supports
 * password upgrades. There are no plans to remove this support. It is deprecated to indicate
 * that this is a legacy implementation and using it is considered insecure.

izeye:
This PR fixes typos.

Closes #5579


rwinch:
Thanks for the PR @izeye! This is merged into master, 5.0.x, and 4.2.x

PR #5634, opened 2018-08-03
Add @Deprecation on removeAuthorizationRequest()

izeye:
This PR adds `@Deprecation` on `AuthorizationRequestRepository.removeAuthorizationRequest() ` as `@deprecated` tag was added in its Javadoc but `@Deprecation` annotation looks missed in 2c1c2c78c363978fce4297b5de8e929bb8344e52.


jgrandja:
Thanks for the PR @izeye. This is now in master.

Issue #5602, opened 2018-07-30
Allow configurable RestOperations for CustomUserTypesOAuth2UserService


jgrandja:
`CustomUserTypesOAuth2UserService` uses Nimbus's HTTP Client internally to retrieve the `UserInfo` resource.

We should re-factor to use `RestOperations` instead and additionally expose a `CustomUserTypesOAuth2UserService.setRestOperations(restOperations)` to allow user's the ability to supply a pre-configured `RestOperations` that are capable of adding custom parameters and/or headers.

Related #5601

Issue #5600, opened 2018-07-30
Allow configurable RestOperations for DefaultOAuth2UserService


jgrandja:
`DefaultOAuth2UserService` uses Nimbus's HTTP Client internally to retrieve the `UserInfo` resource.

We should re-factor to use `RestOperations` instead and additionally expose a `DefaultOAuth2UserService.setRestOperations(restOperations)` to allow user's the ability to supply a pre-configured `RestOperations` that are capable of adding custom parameters and/or headers.

Related #5601

PR #5641, opened 2018-08-04
Expose RestOperations in DefaultOAuth2UserService and CustomUserTypesOAuth2UserService

jgrandja:
Fixes gh-5600, Fixes gh-5602


oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/userinfo/OAuth2UserRequestEntityConverterTests.java
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * @author Joe Grandja
+ */
+public class OAuth2UserRequestEntityConverterTests {
+	private OAuth2UserRequestEntityConverter converter = new OAuth2UserRequestEntityConverter();
+	private OAuth2UserRequest userRequest;
+
+	@Before
+	public void setup() {
+		ClientRegistration clientRegistration = ClientRegistration.withRegistrationId("registration-1")
```
rwinch: Could use leverage TestClientRegistrations here?

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/userinfo/OAuth2UserRequestEntityConverterTests.java
```diff
@@ -0,0 +1,125 @@
...
+		assertThat(headers.getAccept()).contains(MediaType.APPLICATION_JSON_UTF8);
+		assertThat(headers.getContentType()).isEqualTo(
+				MediaType.valueOf(APPLICATION_FORM_URLENCODED_VALUE + ";charset=UTF-8"));
+
+		MultiValueMap<String, String> formParameters = (MultiValueMap<String, String>) requestEntity.getBody();
+		assertThat(formParameters.getFirst(OAuth2ParameterNames.ACCESS_TOKEN)).isEqualTo(
+				this.userRequest.getAccessToken().getTokenValue());
+	}
+
+	private ClientRegistration.Builder from(ClientRegistration registration) {
```
rwinch: You could remove this method by creating a ClientRegistration.Builder as a member variable and then you can modify it in the tests before building. Alternatively, you could also directly use TestClientRegistrations and modify it. I'd probably prefer the layer of indirection with the member variable so refactoring would be easier.

jgrandja:
@rwinch I applied the updates based on your feedback.

PR #5643, opened 2018-08-06
Replace isEqualTo(null) with isNull()

izeye:
This PR replaces `isEqualTo(null)` with `isNull()`.


jzheaux:
Thanks for the contribution, @izeye!

Issue #5629, opened 2018-08-01
Expose strategy for converting Jwt into Authentication


jzheaux:
Let's expose a strategy for converting a Jwt into an Authentication.

The flow from bearer token to Authentication goes something like this:

String token -> Jwt jwt
Jwt jwt -> Authentication authentication

The first step can be configured by wiring `JwtAuthenticationProvider` with an instance of `JwtDecoder`.

We could do the same thing with the second step.

Spring Core has a `Converter` interface that could work well.

Let's include also an abstract implementation that makes it easier to only participate in the conversion of the Jwt into a set of GrantedAuthorities.

gburboz:
Is this for converting jwt based access token to authentication without having to invoke user-info service? 

**Suggestion**
* it will also be great if default jwt converter can perform conversion based off [OIDC Standard Claims](http://openid.net/specs/openid-connect-core-1_0.html#Claims) and/or provide ability to configure mapping of claims on token to expected claim names.
* need to consider which claim to be used for principal name which should default to claim expression `<sub>@<iss>`

jzheaux:
@gburboz The initial target for this strategy is Resource Server, where the story is more about inferring a set of authorities from a bearer token and less about who is the subject.  To that end, the converter for the time being only exposes an override for inferring a `Collection<GrantedAuthority>` from the jwt.

That said, yes, the idea is that broader conversions will be possible as well. At this point, the only place we have concrete plans to expose this strategy is in the Resource Server DSL.

I like your suggestions for some common conversion use cases and have taken note. Would recommend that you open a ticket and ping @jgrandja about it.

gburboz:
Thanks @jzheaux  , I have raised ticket #5659

PR #5645, opened 2018-08-07
Jwt -> Authentication Conversion

jzheaux:
Exposes ability to specify strategy for converting Jwt into an
Authentication, specifically in JwtAuthenticationProvider.

Fixes: gh-5629

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
@@ -19,6 +19,8 @@ package org.springframework.security.config.annotation.web.configurers.oauth2.se
 import javax.servlet.http.HttpServletRequest;
 
 import org.springframework.context.ApplicationContext;
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.security.authentication.AbstractAuthenticationToken;
 import org.springframework.security.authentication.AuthenticationManager;
 import org.springframework.security.config.annotation.web.HttpSecurityBuilder;
 import org.springframework.security.config.annotation.web.configurers.AbstractHttpConfigurer;
@@ -26,8 +28,10 @@ import org.springframework.security.config.annotation.web.configurers.CsrfConfig
 import org.springframework.security.config.annotation.web.configurers.ExceptionHandlingConfigurer;
 import org.springframework.security.config.http.SessionCreationPolicy;
 import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
+import org.springframework.security.oauth2.jwt.Jwt;
 import org.springframework.security.oauth2.jwt.JwtDecoder;
 import org.springframework.security.oauth2.jwt.NimbusJwtDecoderJwkSupport;
+import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationConverter;
 import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationProvider;
 import org.springframework.security.oauth2.server.resource.web.BearerTokenAuthenticationEntryPoint;
 import org.springframework.security.oauth2.server.resource.web.BearerTokenAuthenticationFilter;
@@ -68,6 +72,15 @@ import org.springframework.util.Assert;
  * </li>
  * <li>
  * expose a {@link JwtDecoder} bean
+ * </li>
+ * </ul>
+ *
+ * Also with {@link #jwt()} consider
+ *
+ * <ul>
+ * <li>
+ * customizing the conversion from a {@link Jwt} to an {@link org.springframework.security.core.Authentication} with
+ * {@link JwtConfigurer#jwtAuthenticationConverter(Converter)}
  * </li>
  * </ul>
  *
@@ -182,9 +195,12 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 		}
 
 		JwtDecoder decoder = this.jwtConfigurer.getJwtDecoder();
+		Converter<Jwt, ? extends AbstractAuthenticationToken> jwtAuthenticationConverter =
+				this.jwtConfigurer.getJwtAuthenticationConverter();
 
 		JwtAuthenticationProvider provider =
 				new JwtAuthenticationProvider(decoder);
+		provider.setJwtAuthenticationConverter(jwtAuthenticationConverter);
 		provider = postProcess(provider);
 
 		http.authenticationProvider(provider);
@@ -194,6 +210,9 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 		private final ApplicationContext context;
 
 		private JwtDecoder decoder;
+
+		private Converter<Jwt, ? extends AbstractAuthenticationToken> jwtAuthenticationConverter =
+				new JwtAuthenticationConverter();
 
 		JwtConfigurer(ApplicationContext context) {
 			this.context = context;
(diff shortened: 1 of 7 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -209,10 +221,27 @@ public JwtConfigurer jwkSetUri(String uri) {
 			return this;
 		}
 
+		public JwtConfigurer jwtConverter(Converter<Jwt, ? extends AbstractAuthenticationToken> jwtConverter) {
```
jgrandja: This name does not align with the current `decoder(JwtDecoder decoder)` method.

However, I'm not sure if `converter(Converter<Jwt, ? extends AbstractAuthenticationToken> jwtConverter)` makes sense either?

I'm wondering if `authenticationConverter(Converter<Jwt, ? extends AbstractAuthenticationToken> converter)` works?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationProvider.java
```diff
@@ -107,6 +110,11 @@ public boolean supports(Class<?> authentication) {
 		return BearerTokenAuthenticationToken.class.isAssignableFrom(authentication);
 	}
 
+	public void setJwtConverter(Converter<Jwt, ? extends AbstractAuthenticationToken> jwtConverter) {
```
jgrandja: Same as comment in DSL, we should consider maybe changing this to `setAuthenticationConverter()`

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtConverter.java
```diff
@@ -16,36 +16,42 @@
...
+import org.springframework.security.oauth2.jwt.Jwt;
+import org.springframework.util.StringUtils;
+
 /**
  * @author Rob Winch
+ * @author Josh Cummings
  * @since 5.1
  */
-class JwtConverter {
+public class JwtConverter implements Converter<Jwt, AbstractAuthenticationToken> {
```
jgrandja: We should rename this to `JwtAuthenticationConverter` as the result is a `JwtAuthenticationToken`

jgrandja: Change `AbstractAuthenticationToken` -> `JwtAuthenticationToken`

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
jzheaux: Yeah, let's circle back with @rwinch on this. He mentioned that `jwtDecoder` would be more correct than `decoder`. I'm not sure I've captured his reasoning yet, but, yes, if we are going to stick with `decoder`, then this should be `converter` (or `authenticationConverter` as you suggested).

Btw, it appears that we are introducing the notion of `AuthenticationConverter` in another PR, which converts from a `ServerWebExchange` to a `Mono<Authentication>`. Just something to keep in mind as we consider the name here.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtConverter.java
jzheaux: I actually ran into difficulty with this. When I'm trying to pick this up as a bean, I use `ResolvableType`, e.g. `getBeanNamesForType(ResolvableType)` so that I can differentiate between `Converter`s of other types from the ones that I want. I figured this:

```
context.getBeanNamesForType(ResolvableType.forClassWithGenerics(Converter.class, Jwt.class, AbstractAuthenticationToken.class))
```

would catch both those extending `Conveter<Jwt, AbstractAuthenticationToken>` and `Converter<Jwt, JwtAuthenticationToken>`, but that didn't appear to be the case. It appears to only pick up those Converters with those exact types.

Are you aware of another way I could achieve this?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
jzheaux: @rwinch would `ServerAuthenticationConverter`'s contract imply that the hypothetical `AuthenticationConverter` would be `HttpServletRequest -> Authentication`, then?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtConverter.java
jzheaux: Nevermind, I see your comment below about not performing bean resolution.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
rwinch: I was mostly clarifying the name. I don't think it really doesn't matter too much from this conversations perspective.

> I think that JwtConverter or JwtAuthenticationConverter would be acceptable with a slight preference to JwtAuthenticationConverter.

I agree. It is converting from a Jwt to an authentication token so I find JwtAuthenticationConverter useful.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -209,10 +229,27 @@ public JwtConfigurer jwkSetUri(String uri) {
...
+			this.jwtConverter = jwtConverter;
+			return this;
+		}
+
 		public OAuth2ResourceServerConfigurer<H> and() {
 			return OAuth2ResourceServerConfigurer.this;
 		}
 
+		Converter<Jwt, ? extends AbstractAuthenticationToken> getJwtConverter() {
```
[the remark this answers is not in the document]
jzheaux: It seems reasonable that boot would have a property like `spring.security.oauth2.resourceserver.scopeAttributeName`, in which case boot may publish an instance of a `Converter` as a bean for us to pick up.

How would we introduce something like that passively?

The reason I mention is because, while I agree that while your reasoning effectively works around the resolution issue I detailed above, it could be reintroduced later on, at which point we won't be able to change the type in the DSL to something more targetted.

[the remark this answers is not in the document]
rwinch: I think we can revisit that if/when the time comes. If we do need that support the converter can be automatically converted to a new interface type passively. So I'm not super concerned there.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
jzheaux: For historical purposes, I'll mention that I made a comment earlier about not wanting to go with `AuthenticationConverter` for `Jwt` -> `AbstractAuthenticationToken` conversions since `ServerAuthetnicationConverter` is semantically deviant from that. Then, I posited that `JwtAuthenticationConverter` might be a better name.

*However* I had misread the conversation, and once I realized it didn't apply, I deleted my comment, but not before Rob speedily replied. :)

Sorry for any confusion for those trying to figure out who Rob was quoting.

jzheaux:
@rwinch This is ready for your review.

Issue #5241, opened 2018-04-17
Follow ClaimsAccessor pattern for scope attribute


jzheaux:
### Summary

`ClaimsAccessor` is a powerful interface that makes it easy to project strongly-typed claim semantics over the top of a "weakly-typed" `Map`:

```java
Map<String, Object> attributes = ...;
ClaimsAccessor claims = () -> attributes;
```

Let's add support for the `scope` attribute in `ClaimsAccessor` or the appropriate sub-type.

jzheaux:
Relates to #4887

jzheaux:
Closing in favor of https://github.com/spring-projects/spring-security/issues/6014

PR #5646, opened 2018-08-07
ScopeClaimAccessor

jzheaux:
Adds a mechanism for retrieving the scope attribute as a collection of
individual strings.

Fixes: gh-5241

### DIFF
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ScopeClaimAccessor.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ScopeClaimAccessor.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2002-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.core;
+
+import java.util.Collection;
+
+/**
+ * @author Josh Cummings
+ * @since 5.1
+ */
+public interface ScopeClaimAccessor extends ClaimAccessor {
+	/**
+	 * Retrieve the scope attribute by {@code scopeClaimName} as a collection of individual scopes
+	 *
+	 * @param scopeClaimName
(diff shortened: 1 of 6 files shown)


oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ScopeClaimAccessor.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package org.springframework.security.oauth2.core;
+
+import java.util.Collection;
+
+/**
+ * @author Josh Cummings
+ * @since 5.1
+ */
+public interface ScopeClaimAccessor extends ClaimAccessor {
```
jgrandja: I don't think we need this. We can use `ClaimAccessor.getClaimAsStringCollection(scope, " ")` directly

jzheaux: We can, but wouldn't you agree that `getIssuer` is easier to comprehend than `getClaimAsURL(JwtClaimNames.ISS)`?

Also, consider what the documentation to a user will look like:

> To get the list of scopes, simply call:
    `jwt.getClaimAsStringCollection("scope", " ")`

vs

> To get the list of scopes, simply call:
    `ScopeClaimAccessor accessor = () -> jwt.getClaims();
    jwt.getScope("scope")`

I realize this also isn't ideal because `Jwt` can't extend `ScopeClaimAccessor`, but I think the second makes the code more readable.

jgrandja: I feel both options are not ideal as far as readability goes. I just think adding another public API that only retrieves `scope` is a bit overkill. Let's see what @rwinch thinks about this.

> I realize this also isn't ideal because Jwt can't extend ScopeClaimAccessor

`Jwt` can implement `ScopeClaimAccessor`

jzheaux: Whoops, yes, implement. :)

And for clarity, I (and I believe Joe) were both saying that Jwt _could_ implement ScopeClaimAccessor, but that it shouldn't since scopes aren't in the Jwt spec.

jzheaux: @rwinch do you think this situation merits a separate interface?

Alternatives are possibly to roll into an existing accessor interface or to just invoke `getClaimAsStringCollection` instead of having a `getScope` convenience method.

rwinch: As Joe mentioned, I don't think we should add a new API for a single use case.

I'm thinking eventually we add support for converting claims into specific types and add a <T> T ClaimsAccessor.getClaim(String) method

jzheaux:
Closing in favor of https://github.com/spring-projects/spring-security/issues/5676

Issue #5131, opened 2018-03-16
Allow configuration of RSA Public Key for Resource Server


jgrandja:
If the Authorization Server does not support/expose a [JWK](https://tools.ietf.org/html/rfc7517) Set URI than we need an alternative method for configuring the Resource Server with the Public Key(s) used for verifying the signature of the JWT.

The initial support will be for RSA keys only.

The RSA key(s) may be discoverable via the `ApplicationContext` as `@Bean` or well-known properties in the `Environment`.

Issue #5648, opened 2018-08-07
Nimbus Jwt Decoder Configurability


jzheaux:
It would be nice if `NimbusJwtDecoderJwkSupport` were a bit easier to test. Also, `NimbusJwtDecoderJwkSupport` is very opinionated in the way that Nimbus gets configured.

We could alleviate both of these by introducing a constructor or separate `JwtDecoder` implementation entirely that takes a Nimbus `JwtProcessor`:

```java
public NimbusJwtDecoderJwkSupport(JwtProcessor<SecurityContext> jwtProcessor) {
    this.jwtProcessor = jwtProcessor;
}
```

or

```java
public class NimbusJwtDecoder(JwtProcessor<SecurityContext> jwtProcessor) {
    this.jwtProcessor = this.jwtProcessor;
}

// and

private NimbusJwtDecoder delegate;

public class NimbusJwtDecoderJwkSupport(String jwkSetUrl) {
    JwtProcessor<SecurityContext> jwtProcessor = // .... configure
    this.delegate = new NimbusJwtDecoder(jwtProcessor);
}
```

Either of these would make it possible to:

1. Write tests that do not depend on a legitimate static Jwt and Jwk Set to be provided, simplifying unit tests
2. Provide for easier configuration for users who want to configure Nimbus themselves outside of Spring Security's guidance.

Flaw101:
This would help greatly in areas where IDP isn't providing the Key's via the jwks url, which is poor but painful to workaround in 5.1. Currently having to copy the the decoder to be able to set a custom keyset. 

```
 X509Certificate certificate = (X509Certificate) CertificateFactory.getInstance("X509") //
                        .generateCertificate(
                                new URL(https://public-signing-key);

                RSAKey jwk = new RSAKey.Builder((RSAPublicKey) certificate.getPublicKey()).keyID("signing-cert")
                        .build();
                final JWKSet jwkSet = new JWKSet(jwk);
                JWKSource keySource = new ImmutableJWKSet(jwkSet);
                ConfigurableJWTProcessor jwtProcessor = new DefaultJWTProcessor();
                JWSAlgorithm expectedJWSAlg = JWSAlgorithm.RS384;

                JWSKeySelector keySelector = new JWSVerificationKeySelector(expectedJWSAlg, keySource);
                jwtProcessor.setJWSKeySelector(keySelector1);

                final JWT jwt = JWTParser.parse(s);
[...]

PR #5655, opened 2018-08-08
Jwt Decoder Local Key Configuration

jzheaux:
Adds support for configuring Resource Server DSL with a local public
key.

Fixes: gh-5131


oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderJwkSupport.java
```diff
@@ -99,13 +105,57 @@ public NimbusJwtDecoderJwkSupport(String jwkSetUrl, String jwsAlgorithm) {
...
+	 *
+	 * Note that this decoder will first verify that the "alg" claim in the JWT matches
+	 * the configured algorithm before verifying the signature.
+	 *
+	 * @param key - the public key to use when verifying the signature
+	 * @param jwsAlgorithm - the expected algorithm
+	 *
+	 * @since 5.1
+	 */
+	public NimbusJwtDecoderJwkSupport(PublicKey key, String jwsAlgorithm) {
```
jzheaux: In case this seems odd, I'll clarify here that this constructor is in this class because it correlates to a constructor in NimbusReactiveJwtDecoder, which supports both remote and local key configuration. The name NimbusJwtDecoderJwkSupport evokes only remote support, which may cause confusion.

I'm actually partial to keeping the two use cases separate, mostly because part of their state is mutually exclusive. For example, the remote construction of this class needs a RestOperations, whereas the local construction does not. If this is instantiated with a public key, then setRestOperations becomes a no-op which may be confusing to users. It makes the class feel less cohesive when it supports both uses cases.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -213,6 +218,11 @@ public JwtConfigurer jwkSetUri(String uri) {
 			return OAuth2ResourceServerConfigurer.this;
 		}
 
+		public JwtConfigurer publicKey(RSAPublicKey key) {
```
jgrandja: I'm wondering if `verifyKey` make more sense. We need to consider when we support JWE and what that method name will be. The current name `publicKey` will collide between verification and decryption.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderJwkSupport.java
jgrandja: > the remote construction of this class needs a RestOperations, whereas the local construction does not. If this is instantiated with a public key, then setRestOperations becomes a no-op which may be confusing to users

I totally agree. We should have a separate implementation of `JwtDecoder` for RSA key configuration.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
jzheaux: Actually, I think that publicKey would never collide so long as we always take public keys (private keys for decryption).

But, you are right because of symmetric keys, which I wasn't thinking about until you mentionoed this. Eventually, I would suppose that this method will take just `javax.security.Key`, which would actually then not only collide, but it would also be semantically untrue since I'm certainly *not* giving it a public key.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderJwkSupport.java
jzheaux: Let's see what @rwinch thinks as this would likely have reactive implications.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
jgrandja: You are right, it wouldn't collide since `PrivateKey` is used for decryption.

I still feel `verifyKey` is more intuitive than `publicKey`.

I'll put my proposal here and we can gather feedback from @rwinch.

```
verifyKey(java.security.interfaces.RSAPublicKey key)		// Verify signature

decryptKey(java.security.interfaces.RSAPrivateKey key)	// Decrypt content

sharedKey(javax.crypto.SecretKey key)	// Encrypt/Decrypt content
```

jzheaux: To me, invoking `decryptKey` indicates that you intend to perform (or allow) decryption in this app.

If I am using a shared key and only intend to verify signatures, how would I do this in your proposal? e.g. I don't want to "enable" decryption just because I'm using a shared key.

jzheaux:
@rwinch There are a couple of comments here awaiting your feedback.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderJwkSupport.java
rwinch: I really think that we should just expose this as the same API with different builders as static methods on the class. The logic is the same, but how you build the implementation just differs.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
rwinch: At least for now, I would remove this all together. This is not a very common case, so users can just either expose a JwtDecoder bean or inject the JwtDecoder into the DSL.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderJwkSupport.java
rwinch: I will add that the name NimbusJwtDecoderJwkSupport is misleading. I find this quite unfortunate. Perhaps we should create a NimbusJwtDecoder that is more generic and supports the builder. The  NimbusJwtDecoderJwkSupport would be deprecated and just delegate to the NimbusJwtDecoder

jzheaux: I think my concerns would be allayed if there weren't state hanging around that was sometimes used in one circumstance but not another, e.g. rest operations (or the jwkSetUrl before that).

For example, if this class had only an immutable instance of the Nimbus JWTProcessor, I think keeping everything in the class makes more sense since how the key is obtained shouldn't matter to this class.

If we do something like this (https://github.com/spring-projects/spring-security/issues/5648), for example, say down the road, then I don't think we need two separate classes. (Not saying we need to specifically do what that tasks says here, it just an example of what I'm thinking about.)

rwinch: Let's take this in stride then. Please first create a ticket for NimbusJwtDecoder which follows this approach. We can add builders for supporting the various use cases. Then NimbusJwtDecoderJwkSupport will internally delegate to NimbusJwtDecoder

jzheaux:
This PR is based on another PR, and so has a couple of extra commits. When reviewing, please feel free to just take a look at [the most recent commit](https://github.com/spring-projects/spring-security/pull/5655/commits/685277af7a46d49da52fb767a6c41b1e204eeda2).

jzheaux:
@rwinch @jgrandja I've updated this to use the new JwtProcessors class instead.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtProcessors.java
```diff
@@ -159,4 +174,39 @@ public Resource retrieveResource(URL url) throws IOException {
 			}
 		}
 	}
+
+	/**
+	 * A builder for creating Nimbus {@link JWTProcessor} instances based on a
+	 * public key.
+	 */
+	public static final class PublicKeyJwtProcessorBuilder {
+		private RSAPublicKey key;
```
rwinch: Does it perhaps make sense to change this type to RSAKey right away since we don't need the RSAPublicKey type except for creating the RSAKey?

config/src/test/resources/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests-WrongAlgorithm.token
```diff
@@ -0,0 +1 @@
+eyJhbGciOiJSUzUxMiJ9.eyJleHAiOjQ2ODczNDQ2NzN9.hvVUW_xwUXd7nGm27E5tLTZ21x64YjP0o-TMW6t_bOkfG1Vp1AMEX8fXvSqeG0vK8TWiB2_keOGtH-eFmAGBEYXq1o1zj1BgMHeaZAVio9n-77DkTzQ7CiOF5M1M7B_Ng4K8ra4DpieZZXVjHTWsuOiU1hWoI1tIna8VucAxZln-oh7PkrYmgwFTlsL2Z9aZZYN_X7ECyRQDf3lRrLwr4Go_XpJ5i9F-GT5LvUYa42uggGjvq_frfb0t5wcmPgjtqiE6l2mnrYFjjKTq1nQRYrJ5wFWOHUTRxNsGS8PwrNxzh6JW1ZZTS0n_JIOvSh__w0WAB241QLoKBx4AETMLQA
```
rwinch: I've also been rethinking our tokens being in external files. I think it might be better to just have a static constant in the test file. This would certainly reduce IO for tests and speed them up. It would also be less jumping around to figure out what is being done.

If we decide to do this, it would probably make sense to keep things as is in this PR and create a separate ticket to change it.

Thoughts?

jzheaux: I think the two competing motives are first have the tests be quick and accurate and second to show typical usage to individuals inspecting the tests for tips. The later motive is why the xml config tests have their configs in separate xml files.

I don't see the second benefit with the tokens being in separate files, so yes, I think it would clean up the tests here to not have the separate files for the tokens, but instead use static constants.

jzheaux:
@rwinch I put some additional thought into whether or not we need to ask for the algorithm, and, in the end, we do need this. The JWS algorithms imply a key algorithm as well as a hashing algorithm, like RS256 means "an RSA key and a SHA-256 signature". We can't infer the signature from the key, so it needs to be specified.

Actually, this ends up aligning nicely with the other builder.

The builder now checks the algorithm to make sure that it is compatible with the key and throws an exception that contains some guidance.

config/src/test/resources/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests-WrongAlgorithm.token
rwinch: Thanks can you create a ticket @jzheaux ?

rwinch:
Thanks @jzheaux This is now merged into master

Issue #5466, opened 2018-06-27
Support Custom Parameters within OAuth Token Request


andersonkyle:
### Summary

When using a 3rd party OAuth provider such as Auth0 or Okta it would be very useful to allow developers to hook into the Token Request build process in order to add custom parameters. This is because the providers mentioned above have added additional parameters on top of the specification to allow for more functionality.

### Examples

---

[Auth0](https://auth0.com/docs/api/authentication#resource-owner-password)

`audience: The unique identifier of the target API you want to access.`

This parameter impacts the format of the access token

`realm: String value of the realm the user belongs.`

This parameter instructs Auth0 to use a particular Identity Provider.

---

A simple hook into the building of the request should provide sufficient flexibility for these custom parameters and any others that are introduced in the future.

**NOTE**: In addition to the custom parameters mentioned above, some providers use custom headers to facilitate a token request.  Auth0 uses an `auth0-forwarded-for` header for brute-force protection.  This is mentioned in the documentation link above.  

This issue is very similar to #4911 which focuses on the Authorization Request.

jgrandja:
@mhyeon-lee It looks like the `auth_type` parameter is supported for the Authorization Request. However, this ticket is addressing the support for custom parameters in the Token Request.

Am I misunderstanding your comment?

mhyeon-lee:
@jgrandja 
You are right. I misunderstood. Thank you.

jgrandja:
Related #5547

jgrandja:
@andersonkyle I just submitted #5656 which should resolve this issue. When you have a moment please take a look at the PR and let me know if this will fulfill your use case.

The PR provides `DefaultAuthorizationCodeTokenResponseClient` and `AbstractOAuth2AccessTokenResponseClient`, which you can extend from and override the necessary methods. You have the ability to customize headers and/or request parameters and even modify the request completely. Also, you don't need to extend/override and can instead provide a new instance of `RestOperations` via `AbstractOAuth2AccessTokenResponseClient.setRestOperations()` and modify the outgoing request however you like using a `ClientHttpRequestInterceptor`.

I think this will give you the flexibility you need?

andersonkyle:
@jgrandja Wonderful.  This fits the bill nicely.  Thanks again!

jgrandja:
[...]

PR #5656, opened 2018-08-10
Add support for customizing the Access Token Request

### LINKED ISSUE #5547: Add RestOperations implementation of OAuth2AccessTokenResponseClient
The current (and default) implementation of `OAuth2AccessTokenResponseClient` for the `authorization_code` grant is `NimbusAuthorizationCodeTokenResponseClient`, which leverages the Nimbus HTTP Client internally.

We should provide a new implementation of `OAuth2AccessTokenResponseClient` for the `authorization_code` grant named `DefaultAuthorizationCodeTokenResponseClient` that uses `RestOperations` for requesting to the Token Endpoint.

Additionally, `DefaultAuthorizationCodeTokenResponseClient` should expose `setRestOperations(restOperations)` method to allow user's the ability to supply a pre-configured `RestOperations`, which can provide the user the capability to add custom parameters and/or headers if required.

`DefaultAuthorizationCodeTokenResponseClient` should be the new default in `OAuth2LoginConfigurer` and `OAuth2ClientConfigurer`.

Related #5601

jgrandja:
This PR resolves #5547 and #5466 

The 1st commit pushes most of the code from `DefaultClientCredentialsTokenResponseClient` to an abstract implementation `AbstractOAuth2AccessTokenResponseClient` to allow for reuse in the next commit.

The 2nd commit adds a new implementation of `OAuth2AccessTokenResponseClient` for the `authorization_code` grant `DefaultAuthorizationCodeTokenResponseClient` - extending `AbstractOAuth2AccessTokenResponseClient`.

The 3rd commit makes `DefaultAuthorizationCodeTokenResponseClient` the default client replacing the previous default `NimbusAuthorizationCodeTokenResponseClient`.

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultAuthorizationCodeTokenResponseClient.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultAuthorizationCodeTokenResponseClient.java
@@ -1,0 +1,125 @@
+/*
+ * Copyright 2002-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.client.endpoint;
+
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.http.RequestEntity;
+import org.springframework.http.ResponseEntity;
+import org.springframework.http.converter.FormHttpMessageConverter;
+import org.springframework.http.converter.HttpMessageConverter;
+import org.springframework.security.oauth2.client.http.OAuth2ErrorResponseErrorHandler;
+import org.springframework.security.oauth2.client.http.converter.OAuth2AuthorizationCodeGrantRequestEntityConverter;
+import org.springframework.security.oauth2.core.AuthorizationGrantType;
+import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
+import org.springframework.security.oauth2.core.OAuth2Error;
+import org.springframework.security.oauth2.core.endpoint.OAuth2AccessTokenResponse;
+import org.springframework.security.oauth2.core.http.converter.OAuth2AccessTokenResponseHttpMessageConverter;
+import org.springframework.util.Assert;
+import org.springframework.util.CollectionUtils;
+import org.springframework.web.client.ResponseErrorHandler;
+import org.springframework.web.client.RestClientException;
+import org.springframework.web.client.RestOperations;
+import org.springframework.web.client.RestTemplate;
+
+import java.util.Arrays;
+
+/**
+ * The default implementation of an {@link OAuth2AccessTokenResponseClient}
+ * for the {@link AuthorizationGrantType#AUTHORIZATION_CODE authorization_code} grant.
+ * This implementation uses a {@link RestOperations} when requesting
+ * an access token credential at the Authorization Server's Token Endpoint.
+ *
+ * @author Joe Grandja
+ * @since 5.1
+ * @see OAuth2AccessTokenResponseClient
+ * @see OAuth2AuthorizationCodeGrantRequest
+ * @see OAuth2AccessTokenResponse
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc6749#section-4.1.3">Section 4.1.3 Access Token Request (Authorization Code Grant)</a>
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc6749#section-4.1.4">Section 4.1.4 Access Token Response (Authorization Code Grant)</a>
+ */
+public final class DefaultAuthorizationCodeTokenResponseClient implements OAuth2AccessTokenResponseClient<OAuth2AuthorizationCodeGrantRequest> {
+	private static final String INVALID_TOKEN_RESPONSE_ERROR_CODE = "invalid_token_response";
+
+	private Converter<OAuth2AuthorizationCodeGrantRequest, RequestEntity<?>> requestEntityConverter =
+			new OAuth2AuthorizationCodeGrantRequestEntityConverter();
+
+	private RestOperations restOperations;
+
+	public DefaultAuthorizationCodeTokenResponseClient() {
+		RestTemplate restTemplate = new RestTemplate(Arrays.asList(
+				new FormHttpMessageConverter(), new OAuth2AccessTokenResponseHttpMessageConverter()));
+		restTemplate.setErrorHandler(new OAuth2ErrorResponseErrorHandler());
+		this.restOperations = restTemplate;
+	}
+
+	@Override
+	public OAuth2AccessTokenResponse getTokenResponse(OAuth2AuthorizationCodeGrantRequest authorizationCodeGrantRequest) throws OAuth2AuthenticationException {
+		Assert.notNull(authorizationCodeGrantRequest, "authorizationCodeGrantRequest cannot be null");
+
+		RequestEntity<?> request = this.requestEntityConverter.convert(authorizationCodeGrantRequest);
+
+		ResponseEntity<OAuth2AccessTokenResponse> response;
+		try {
+			response = this.restOperations.exchange(request, OAuth2AccessTokenResponse.class);
+		} catch (RestClientException ex) {
+			OAuth2Error oauth2Error = new OAuth2Error(INVALID_TOKEN_RESPONSE_ERROR_CODE,
+					"An error occurred while attempting to retrieve the OAuth 2.0 Access Token Response: " + ex.getMessage(), null);
+			throw new OAuth2AuthenticationException(oauth2Error, oauth2Error.toString(), ex);
+		}
+
+		OAuth2AccessTokenResponse tokenResponse = response.getBody();
+
+		if (CollectionUtils.isEmpty(tokenResponse.getAccessToken().getScopes())) {
+			// As per spec, in Section 5.1 Successful Access Token Response
+			// https://tools.ietf.org/html/rfc6749#section-5.1
+			// If AccessTokenResponse.scope is empty, then default to the scope
+			// originally requested by the client in the Token Request
+			tokenResponse = OAuth2AccessTokenResponse.withResponse(tokenResponse)
+					.scopes(authorizationCodeGrantRequest.getClientRegistration().getScopes())
+					.build();
+		}
+
+		return tokenResponse;
+	}
+
+	/**
+	 * Sets the {@link Converter} used for converting the {@link OAuth2AuthorizationCodeGrantRequest}
+	 * to a {@link RequestEntity} representation of the OAuth 2.0 Access Token Request.
+	 *
+	 * @param requestEntityConverter the {@link Converter} used for converting to a {@link RequestEntity} representation of the Access Token Request
+	 */
+	public void setRequestEntityConverter(Converter<OAuth2AuthorizationCodeGrantRequest, RequestEntity<?>> requestEntityConverter) {
+		Assert.notNull(requestEntityConverter, "requestEntityConverter cannot be null");
+		this.requestEntityConverter = requestEntityConverter;
+	}
+
+	/**
+	 * Sets the {@link RestOperations} used when requesting the OAuth 2.0 Access Token Response.
+	 *
+	 * <p>
+	 * <b>NOTE:</b> At a minimum, the supplied {@code restOperations} must be configured with the following:
+	 * <ol>
+	 *  <li>{@link HttpMessageConverter}'s - {@link FormHttpMessageConverter} and {@link OAuth2AccessTokenResponseHttpMessageConverter}</li>
+	 *  <li>{@link ResponseErrorHandler} - {@link OAuth2ErrorResponseErrorHandler}</li>
+	 * </ol>
+	 *
+	 * @param restOperations the {@link RestOperations} used when requesting the Access Token Response
+	 */
+	public void setRestOperations(RestOperations restOperations) {
+		Assert.notNull(restOperations, "restOperations cannot be null");
+		this.restOperations = restOperations;
+	}
+}
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultAuthorizationCodeTokenResponseClient.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultAuthorizationCodeTokenResponseClient.java
@@ -21,7 +21,6 @@ import org.springframework.http.ResponseEntity;
 import org.springframework.http.converter.FormHttpMessageConverter;
 import org.springframework.http.converter.HttpMessageConverter;
 import org.springframework.security.oauth2.client.http.OAuth2ErrorResponseErrorHandler;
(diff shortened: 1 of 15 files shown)


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/AbstractOAuth2AccessTokenResponseClient.java
```diff
@@ -0,0 +1,290 @@
...
+
+	protected AbstractOAuth2AccessTokenResponseClient() {
+		RestTemplate restTemplate = new RestTemplate();
+		// Disable the ResponseErrorHandler as errors are handled directly within this class
+		restTemplate.setErrorHandler(new NoOpResponseErrorHandler());
+		this.restOperations = restTemplate;
+	}
+
+	@Override
+	public OAuth2AccessTokenResponse getTokenResponse(T authorizationGrantRequest) throws OAuth2AuthenticationException {
```
gburboz: Rename parameter `authorizationGrantRequest` to `accessTokenRequest` and mark it `final`

rwinch:
At a high level, I don't really like the idea of the `AbstractOAuth2AccessTokenResponseClient`. It seems to be introducing additional APIs that are already available in RestTemplate. For example, there is a `buildRequestMethod` which is already supported in RestTemplate using a MessageConverter. The conversion of the error response is already possible using a `ResponseErrorHandler`. The parsing of the response is possible using a MessageConverter as well.

I propose the code be refactored to be exposed using the existing APIs within RestTemplate. The DefaultClientCredentialsTokenResponseClient and DefaultAuthorizationCodeTokenResponseClient can then use a RestTemplate with the appropriate API implementations registered by default. User's would then be required to provide a RestOperations implementation that can convert from a request/response to the OAuth2AccessTokenResponseClient. Note that proper setup of the RestTempalte is already required since the AbstractOAuth2AccessTokenResponseClient relies on a NoOpResponseErrorHandler being set to handle the errors.

jgrandja:
@rwinch Ok this makes sense as far as leveraging `HttpMessageConverter` for reading `OAuth2AccessTokenResponse` on a success response and reading `OAuth2Error` using `ResponseErrorHandler` on error response.

Can you clarify:

> there is a buildRequestMethod which is already supported in RestTemplate using a MessageConverter.

I can't seem to find `buildRequestMethod`? Did you mean to leverage `RequestCallback`?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/AbstractOAuth2AuthorizationGrantRequest.java
```diff
@@ -26,19 +27,34 @@
...
+	}
+
+	/**
+	 * Sub-class constructor.
+	 *
+	 * @since 5.1
+	 * @param authorizationGrantType the authorization grant type
+	 * @param clientRegistration the client registration
+	 */
+	protected AbstractOAuth2AuthorizationGrantRequest(
```
rwinch: Is this change necessary for this PR?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultAuthorizationCodeTokenResponseClient.java
```diff
@@ -0,0 +1,133 @@
...
+ */
+package org.springframework.security.oauth2.client.endpoint;
+
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.RequestEntity;
+import org.springframework.http.ResponseEntity;
+import org.springframework.http.client.ClientHttpResponse;
+import org.springframework.http.converter.FormHttpMessageConverter;
+import org.springframework.security.oauth2.client.endpoint.converter.OAuth2AuthorizationCodeGrantRequestEntityConverter;
```
rwinch: This is a package tangle because a parent package is accessing a subpackage of itself.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultAuthorizationCodeTokenResponseClient.java
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * @author Joe Grandja
+ * @since 5.1
+ * @see OAuth2AccessTokenResponseClient
+ * @see OAuth2AuthorizationCodeGrantRequest
+ * @see OAuth2AccessTokenResponse
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc6749#section-4.1.3">Section 4.1.3 Access Token Request (Authorization Code Grant)</a>
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc6749#section-4.1.4">Section 4.1.4 Access Token Response (Authorization Code Grant)</a>
+ */
+public class DefaultAuthorizationCodeTokenResponseClient implements OAuth2AccessTokenResponseClient<OAuth2AuthorizationCodeGrantRequest> {
```
rwinch: Most of the methods are already final, let's make the class final

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultClientCredentialsTokenResponseClient.java
```diff
@@ -151,120 +97,37 @@ public OAuth2AccessTokenResponse getTokenResponse(OAuth2ClientCredentialsGrantRe
...
 	}
 
 	/**
-	 * Sets the {@link RestOperations} used when requesting the access token response.
+	 * Sets the {@link RestOperations} used when requesting the OAuth 2.0 Access Token Response.
 	 *
-	 * @param restOperations the {@link RestOperations} used when requesting the access token response
+	 * @param restOperations the {@link RestOperations} used when requesting the Access Token Response
 	 */
 	public final void setRestOperations(RestOperations restOperations) {
```
rwinch: How can someone inject the RestOperations and still get valid error handling? I think we may need to make OAuth2ErrorResponseErrorHandler public.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/converter/OAuth2AuthorizationCodeGrantRequestEntityConverter.java
```diff
@@ -0,0 +1,110 @@
...
+		return new RequestEntity<>(formParameters, headers, HttpMethod.POST, uri);
+	}
+
+	/**
+	 * Returns the {@link HttpHeaders} used for the Access Token Request.
+	 *
+	 * @param authorizationCodeGrantRequest the authorization code grant request
+	 * @return the {@link HttpHeaders} used for the Access Token Request
+	 */
+	protected HttpHeaders buildHeaders(OAuth2AuthorizationCodeGrantRequest authorizationCodeGrantRequest) {
```
rwinch: I'd like this to be private. User's that need to customize the headers can do so by delegation to this class and constructing a modified RequestEntity.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/converter/OAuth2AuthorizationCodeGrantRequestEntityConverter.java
```diff
@@ -0,0 +1,110 @@
...
+		return headers;
+	}
+
+	/**
+	 * Returns a {@link MultiValueMap} of the form parameters used for the Access Token Request body.
+	 *
+	 * @param authorizationCodeGrantRequest the authorization code grant request
+	 * @return a {@link MultiValueMap} of the form parameters used for the Access Token Request body
+	 */
+	protected MultiValueMap<String, String> buildFormParameters(OAuth2AuthorizationCodeGrantRequest authorizationCodeGrantRequest) {
```
rwinch: I'd like this to be private. User's that need to customize the headers can do so by delegation to this class and constructing a modified RequestEntity.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/converter/OAuth2ClientCredentialsGrantRequestEntityConverter.java
```diff
@@ -0,0 +1,112 @@
...
+		return new RequestEntity<>(formParameters, headers, HttpMethod.POST, uri);
+	}
+
+	/**
+	 * Returns the {@link HttpHeaders} used for the Access Token Request.
+	 *
+	 * @param clientCredentialsGrantRequest the client credentials grant request
+	 * @return the {@link HttpHeaders} used for the Access Token Request
+	 */
+	protected HttpHeaders buildHeaders(OAuth2ClientCredentialsGrantRequest clientCredentialsGrantRequest) {
```
rwinch: I'd like this to be private. User's that need to customize the headers can do so by delegation to this class and constructing a modified RequestEntity.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/converter/OAuth2ClientCredentialsGrantRequestEntityConverter.java
```diff
@@ -0,0 +1,112 @@
...
+		return headers;
+	}
+
+	/**
+	 * Returns a {@link MultiValueMap} of the form parameters used for the Access Token Request body.
+	 *
+	 * @param clientCredentialsGrantRequest the client credentials grant request
+	 * @return a {@link MultiValueMap} of the form parameters used for the Access Token Request body
+	 */
+	protected MultiValueMap<String, String> buildFormParameters(OAuth2ClientCredentialsGrantRequest clientCredentialsGrantRequest) {
```
rwinch: I'd like this to be private. User's that need to customize the headers can do so by delegation to this class and constructing a modified RequestEntity.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultClientCredentialsTokenResponseClient.java
rwinch: We should document what the RestOperations required (i.e. the types of converters it would need, error handling, etc)

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultAuthorizationCodeTokenResponseClient.java
```diff
@@ -0,0 +1,133 @@
...
+		Assert.notNull(requestEntityConverter, "requestEntityConverter cannot be null");
+		this.requestEntityConverter = requestEntityConverter;
+	}
+
+	/**
+	 * Sets the {@link RestOperations} used when requesting the OAuth 2.0 Access Token Response.
+	 *
+	 * @param restOperations the {@link RestOperations} used when requesting the Access Token Response
+	 */
+	public final void setRestOperations(RestOperations restOperations) {
```
rwinch: Same comments as the other setRestOperations

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultAuthorizationCodeTokenResponseClient.java
rwinch: i'm wondering if to fix the tangle if we could move `OAuth2AuthorizationCodeGrantRequestEntityConverter` and `OAuth2ClientCredentialsGrantRequestEntityConverter` to `org.springframework.security.oauth2.core.http.converter`

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/AbstractOAuth2AuthorizationGrantRequest.java
jgrandja: Agreed this change can be removed

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultAuthorizationCodeTokenResponseClient.java
jgrandja: `OAuth2AuthorizationCodeGrantRequest` and `OAuth2ClientCredentialsGrantRequest` reside in the `client.endpoint` package so we should keep it in the `client` module.

The 2 options I can think of right now are:
`org.springframework.security.oauth2.client.endpoint` (keep in same package)
`org.springframework.security.oauth2.client.converter`

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultClientCredentialsTokenResponseClient.java
jgrandja: I was thinking of exposing `OAuth2ErrorResponseErrorHandler` but the implementation was pretty easy so decided to hide it.

Either way, I'm fine with exposing it.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultAuthorizationCodeTokenResponseClient.java
rwinch: Good point, I think given it is associated with RequestEntity, we should consider a variation of what you proposed. What do you think of `org.springframework.security.oauth2.client.http.converter`

jgrandja: Agreed that makes sense....let's go with `org.springframework.security.oauth2.client.http.converter`

jgrandja:
@rwinch I've applied the updates as discussed. I've also rebased/squashed to 1 commit and excluded the changes to `DefaultClientCredentialsTokenResponseClient`. After this is merged, I'll resolve #5735

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultAuthorizationCodeTokenResponseClient.java
```diff
@@ -0,0 +1,125 @@
...
+ */
+package org.springframework.security.oauth2.client.endpoint;
+
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.http.RequestEntity;
+import org.springframework.http.ResponseEntity;
+import org.springframework.http.converter.FormHttpMessageConverter;
+import org.springframework.http.converter.HttpMessageConverter;
+import org.springframework.security.oauth2.client.http.OAuth2ErrorResponseErrorHandler;
+import org.springframework.security.oauth2.client.http.converter.OAuth2AuthorizationCodeGrantRequestEntityConverter;
```
rwinch: Unfortunately this is another package tangle. The `client.endpoint` package refers to `http.converter` above and `http.converter` refers to `client.endpoint`

jgrandja:
`DefaultAuthorizationCodeTokenResponseClient` and `OAuth2AuthorizationCodeGrantRequest` reside in the `client.endpoint` so I moved `OAuth2AuthorizationCodeGrantRequestEntityConverter` to the same package as this made the most sense rather than it's own `http.converter` package.

`DefaultAuthorizationCodeTokenResponseClient` **uses** `OAuth2AuthorizationCodeGrantRequestEntityConverter` which **uses** `OAuth2AuthorizationCodeGrantRequest`.

rwinch:
Thanks please squash and rebase off master as there have been some recent checkstyle changes. Once that is done and the build passes, please feel free to merge.

jgrandja:
Thanks @rwinch. This is now in master.

Issue #5238, opened 2018-04-16
Support WebAuthn


rwinch:
### Summary
https://www.w3.org/TR/webauthn

ynojima:
Hello @rwinch,
As I've commented on https://github.com/spring-projects/spring-security/issues/2603#issuecomment-377922919, I'm implementing WebAuthn for spring-security (https://github.com/ynojima/spring-security-webauthn).
 It is still a proof of concept, but in future, I'd like to send pull request. I'll be happy if you look into it.

rwinch:
@ynojima Thank you for reaching out (again) and pointing me to your comment! Sorry I had missed your comment. Let's move all of the discussion to here going forward.

I'd love for you to send a pull request. I will review your proof of concept and pull request this week.

Thanks again!

ynojima:
Thank you for your reply, and here is a separate pull request to make a foundation for multi-factor authentication in spring security: https://github.com/spring-projects/spring-security/pull/5196
My spring-security-webauthn is built on the top of it, and other kind multi-factor(step) authentication provider may utilize it to indicate a user who passed the first authentication.

rwinch:
Thanks! I'm taking a look at your sample now.

rwinch:
Thank you for putting this together. I have tried the sample in both FireFox and Chrome.

In Chrome 65.0.3325.181 with Web Authentication API I attempt to register my U2F FIDO key on the register page and it gives me the following error in my Boot application:

```
2018-04-19 17:00:22.195 ERROR 9086 --- [nio-8080-exec-8] o.a.c.c.C.[.[.[/].[dispatcherServlet]    : Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed; nested exception is com.webauthn4j.exception.NotImplementedException] with root cause

com.webauthn4j.exception.NotImplementedException: null
        at com.webauthn4j.context.validator.attestation.FIDOU2FAttestationStatementValidator.validateTrustworthiness(FIDOU2FAttestationStatementValidator.java:58)
        at com.webauthn4j.context.validator.attestation.AbstractAttestationStatementValidator.validate(AbstractAttestationStatementValidator.java:15)
        at com.webauthn4j.context.validator.WebAuthnRegistrationContextValidator.validateAttestationStatement(WebAuthnRegistrationContextValidator.java:117)
[...]

PR #5665, opened 2018-08-14
Make a foundation for multi-factor(step) authentication including WebAuthn

ynojima:
This patch is to make a foundation for multi-factor(step) authentication including WebAuthn (#5238).

## Changes

\* no interface is changed

* Add `MultifactorAuthenticationToken` to represent a user in the middle of multi factor(step) authentication process
* Add `MFATokenEvaluator`/`MFATokenEvaluatorImpl` for `Authentication` type check
* Make `ExceptionTranslationFilter`, `AuthenticationTrustResolverImpl`, and `HttpSessionSecurityContextRepository` use `MFATokenEvaluator` to support multi-factor authentication
* Add `MultiFactorAuthenticationProvider`, which authenticates a user by delegating to another `AuthenticationProvider` and generates `MultifactorAuthenticationToken`

[spring-security-webauthn](https://github.com/ynojima/spring-security-webauthn) is a concrete user facing code using this patch.
If possible, I'd like to send whole spring-security-webauthn as a pull-request to spring-security now, but 
I understand the spring's backward compatibility policy, WebAuthn specification is still updated after the declaration of Candidate Recomendation phase, and browser implementation is still on going. It is not the time.

Meanwhile, spring-security-webauthn requires user feedbacks, and patching to spring-security core is a big roadblocks for users to try spring-security-webauthn.
I extracted minimal changes to spring-security core into this patch. and I suppose future changes to WebAuthn specification will not affect this.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/authentication/configurers/mfa/MultiFactorAuthenticationProviderConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/authentication/configurers/mfa/MultiFactorAuthenticationProviderConfigurer.java
@@ -26,35 +26,35 @@ import org.springframework.security.config.annotation.authentication.ProviderMan
  * @param <B> the type of the {@link SecurityBuilder}
  */
 public class MultiFactorAuthenticationProviderConfigurer<B extends ProviderManagerBuilder<B>>
-        extends SecurityConfigurerAdapter<AuthenticationManager, B> {
+		extends SecurityConfigurerAdapter<AuthenticationManager, B> {
 
-    //~ Instance fields
-    // ================================================================================================
-    private AuthenticationProvider authenticationProvider;
-    private MFATokenEvaluator mfaTokenEvaluator = new MFATokenEvaluatorImpl();
+	//~ Instance fields
+	// ================================================================================================
+	private AuthenticationProvider authenticationProvider;
+	private MFATokenEvaluator mfaTokenEvaluator = new MFATokenEvaluatorImpl();
 
-    /**
-     * Constructor
-     * @param authenticationProvider {@link AuthenticationProvider} to be delegated
-     */
-    public MultiFactorAuthenticationProviderConfigurer(AuthenticationProvider authenticationProvider) {
-        this.authenticationProvider = authenticationProvider;
-    }
+	/**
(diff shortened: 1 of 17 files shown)


ynojima:
Hi @rwinch, just a gentle ping to see if this PR needs any more polish or anything.
Are you busy on releasing Spring Security 5.1?

rwinch:
@ynojima Thanks for the nudge. Among other things, Spring Security 5.1 is taking a lot of my time. I have added reviewing this to [my queue](https://github.com/orgs/spring-projects/projects/1) and will get back to you.


NOTE: 

I have not looked over the code, but one thing that comes to mind is that I really look for to ensure that we provide a real feature to end users vs just infrastructure changes. This is especially true with large changes where it is hard to anticipate what the infrastructure change needs to be.

So if we have something that is providing a foundation for multifactor authentication, I want code that also can use that foundation in a real scenario ready to merge too. In these cases there will be more than one commit, but I will very rarely merge something that is large without seeing all the moving parts together. The reason is that we really want to ensure that all code is going to provide value to our end users.

ynojima:
[spring-security-webauthn](https://github.com/ynojima/spring-security-webauthn) is a consumer of the foundation classes in a real scenario. It also has a [sample application](https://ynojima.github.io/spring-security-webauthn/en/#sample-application), but sad to say, it is not ready to be merged into spring-security as [browser (ex. Chrome) WebAuthn  implementation is still on going](https://bugs.chromium.org/p/chromium/issues/list?q=component:Blink%3EWebAuthentication).
There is a room I need to change the behavior of spring-security-webauthn to fit for browser implementation, but I assume the foundation classes in this pull-request will not be affected.
These foundation classes are only used for FIDO-U2F token support, and FIDO-U2F token is already supported by WebAuthn implementation of Chrome, Firefox, and Edge, I've confirmed the foundation classes works with these browsers.

I don't mind if you don't merge this patch at this moment (I'll be more happy if you merge it :) ),
I really appriciate if you review the the direction of this patch and [spring-security-webauthn](https://github.com/ynojima/spring-security-webauthn) again.

Thanks!

ynojima:
As I opened a new pull-request with wider scopes (concrete use-case), I close this pull-request.

PR #5670, opened 2018-08-15
Add support for Feature-Policy security header

vpavic:
See the following resources:

- https://wicg.github.io/feature-policy/
- https://scotthelme.co.uk/a-new-security-header-feature-policy/
- https://developers.google.com/web/updates/2018/06/feature-policy

Although the Feature Policy RFC is not finalized yet, there is already [browser adoption](https://caniuse.com/#search=feature%20policy) and the configuration of `Feature-Policy` header is (similarly to `Content-Security-Policy`) fairly simple and based on a set of directives that we configure as a `String`.

Also note that https://securityheaders.com/ already supports this header in their report.


rwinch:
@vpavic Thanks for the PR! Any chance you would be able to provide a PR for WebFlux support as well? See https://github.com/spring-projects/spring-security/issues/5672

Issue #2312, opened 2012-12-29
SEC-2109: Custom socket factory in ActiveDirectoryLdapAuthenticationProvider


spring-issuemaster:
[Davide Angelocola](https://jira.spring.io/secure/ViewProfile.jspa?name=dfa) (Migrated from [SEC-2109](https://jira.spring.io/browse/SEC-2109?redirect=false)) said:

Actually there is no way to specify a custom socket factory in _ActiveDirectoryLdapAuthenticationProvider_ (property `java.naming.ldap.factory.socket` of _DirContext_).

It could be useful for a number of reasons as described by
http://docs.oracle.com/javase/jndi/tutorial/ldap/security/ssl.html (section "Using Custom Sockets").

sksjava:
Not sure why the issue is still not progressed. This is a valid issue and causes inconviniece when developing multiple softwares running on same JVM and are connecting do different URLs over SSL.

The current implementation is not allowing the software to process LDAPS without having to override the JVM truststore parameters. This causes issues if there are multiple application installed on the same server having different truststores.

nE0sIghT:
This is very important feature that will allow to properly validate custom enterprise certificates.
We can introduce some method to set ```SSLSocketFactory``` name and later use it in ```bindAsUser()```.

rwinch:
@nE0sIghT  Thanks for the feedback.

I think what we should do is allow leverage a `ContextSource.getContext(String principal, String credentials)` implementation to resolve the DirContext for `ActiveDirectoryLdapAuthenticationProvider.bindAsUser`. Users wishing to customize the environment could inject a custom ContextSource. 

If you are interested in submitting a Pull Request, please do so!

rwinch:
Closing in favor of #5674 since it is a PR

PR #5674, opened 2018-08-16
ActiveDirectoryLdapAuthenticationProvider should allow custom environment

nE0sIghT:
This change allows to inject custom factory for directory context initialization so additional environment properties may be provided for InitialLdapContext.

Closes: https://github.com/spring-projects/spring-security/issues/2312

### DIFF
--- a/ldap/src/main/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProvider.java
+++ b/ldap/src/main/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProvider.java
@@ -107,6 +107,7 @@ public final class ActiveDirectoryLdapAuthenticationProvider extends
 	private final String url;
 	private boolean convertSubErrorCodesToExceptions;
 	private String searchFilter = "(&(objectClass=user)(userPrincipalName={0}))";
+	private Hashtable<String, Object> contextEnvironmentProperties = null;
 
 	// Only used to allow tests to substitute a mock LdapContext
 	ContextFactory contextFactory = new ContextFactory();
@@ -190,7 +191,7 @@ public final class ActiveDirectoryLdapAuthenticationProvider extends
 		// TODO. add DNS lookup based on domain
 		final String bindUrl = url;
 
-		Hashtable<String, String> env = new Hashtable<>();
+		Hashtable<String, Object> env = new Hashtable<>();
 		env.put(Context.SECURITY_AUTHENTICATION, "simple");
 		String bindPrincipal = createBindPrincipal(username);
 		env.put(Context.SECURITY_PRINCIPAL, bindPrincipal);
@@ -198,6 +199,10 @@ public final class ActiveDirectoryLdapAuthenticationProvider extends
 		env.put(Context.SECURITY_CREDENTIALS, password);
 		env.put(Context.INITIAL_CONTEXT_FACTORY, "com.sun.jndi.ldap.LdapCtxFactory");
 		env.put(Context.OBJECT_FACTORIES, DefaultDirObjectFactory.class.getName());
+
+		if(contextEnvironmentProperties != null) {
+			env.putAll(contextEnvironmentProperties);
+		}
 
 		try {
 			return contextFactory.createContext(env);
@@ -396,6 +401,16 @@ public final class ActiveDirectoryLdapAuthenticationProvider extends
 	public void setSearchFilter(String searchFilter) {
 		Assert.hasText(searchFilter, "searchFilter must have text");
 		this.searchFilter = searchFilter;
+	}
+
+	/**
+	 * Allows a custom environment properties to be used to create initial LDAP context.
+	 *
+	 * @param contextFactory
+	 */
+	public void setContextEnvironmentProperties(Hashtable<String, Object> environment) {
+		Assert.notEmpty(environment, "environment must not be empty");
+		this.contextEnvironmentProperties = new Hashtable<>(environment);
 	}
 
 	static class ContextFactory {
--- a/ldap/src/test/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProviderTests.java
+++ b/ldap/src/test/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProviderTests.java
@@ -393,6 +393,31 @@ public class ActiveDirectoryLdapAuthenticationProviderTests {
 
 	}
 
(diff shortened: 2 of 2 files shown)


pivotal-issuemaster:
@nE0sIghT Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=5674)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=5674) to manually synchronize the status of this Pull Request.

ldap/src/main/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProvider.java
```diff
@@ -398,8 +408,16 @@ public void setSearchFilter(String searchFilter) {
 		this.searchFilter = searchFilter;
 	}
 
-	static class ContextFactory {
-		DirContext createContext(Hashtable<?, ?> env) throws NamingException {
+	/**
+	 * Factory for directory context
+	 */
+	public static class ContextFactory {
```
rwinch: Rather than exposing a custom type, I think we should change this type to `ContextSource`.

ldap/src/main/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProvider.java
```diff
@@ -214,6 +214,16 @@ private DirContext bindAsUser(String username, String password) {
 		}
 	}
 
+	/**
+	 * Allows a custom context factory to be used to create initial LDAP context.
+	 * This may be used to inject additional environment properties to LDAP context.
+	 *
+	 * @param contextFactory
+	 */
+	public void setContextFactory(ContextFactory contextFactory) {
```
rwinch: We should add Assert.notNull

ldap/src/main/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProvider.java
nE0sIghT: What do you think about using (and exposing via ```setContextFactory```) less generic ```LdapContextSource```?
It is perfectly suite AD provider needs as I think.

rwinch: We should not create a new public API unless we really need it.

nE0sIghT:
@rwinch 
Thanks for review. I pushed reworked commit with ```ContextFactory``` changed to ```LdapContextSource``` (still used internally only) and introduced new public method ```setContextEnvironmentProperties``` which allows to set custom context properties.

ldap/src/main/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProvider.java
```diff
@@ -187,29 +186,28 @@ protected DirContextOperations doAuthentication(
...
 		String bindPrincipal = createBindPrincipal(username);
-		env.put(Context.SECURITY_PRINCIPAL, bindPrincipal);
-		env.put(Context.PROVIDER_URL, bindUrl);
-		env.put(Context.SECURITY_CREDENTIALS, password);
-		env.put(Context.INITIAL_CONTEXT_FACTORY, "com.sun.jndi.ldap.LdapCtxFactory");
-		env.put(Context.OBJECT_FACTORIES, DefaultDirObjectFactory.class.getName());
+
+		// TODO. add DNS lookup based on domain
+		contextSource.setUrl(url);
+		contextSource.setUserDn(bindPrincipal);
```
rwinch: This is a race condition because we are setting a stack variable onto a member variable.

rwinch:
@nE0sIghT Sorry I don't think I was clear. Sorry for the confusion.

I was suggesting that users would need to provide their own ContextSource as apposed to providing a way of setting the properties. This gives more flexibility.

In hindsight, I do prefer the option you chose with allowing the user to set any property because it is more passive. However, we cannot make changes to the ContextFactory then.

nE0sIghT:
@rwinch 
If we will allow to provide ```ContextSource``` implementation is it ok to use ```LdapContextSource``` for that and not ```ContextSource```? Otherwise I don't think that it will be better for me as framework user.
```
public void setContextSource(LdapContextSource)
```

rwinch:
>If we will allow to provide ContextSource implementation is it ok to use LdapContextSource for that and not ContextSource? Otherwise I don't think that it will be better for me as framework user.

@nE0sIghT Why? `LdapContextSource` implements `ContextSource` so if it is an instance of `LdapContextSource`, then it can be injected into the `setContextSource(ContextSource)` method. Allowing an interface to be injected is much better because it does not limit user's to leveraging the `LdapContextSource` implementation.

nE0sIghT:
@rwinch 
Because ```LdapContextSource``` requires some initialization.
Let's assume that user will provide instance of ```LdapContextSource``` to ```setContextSource(ContextSource)``` method:

1. ```ActiveDirectoryLdapAuthenticationProvider``` will not know about ```LdapContextSource``` and we should not check if ```contextSource``` is instance of ```LdapContextSource``` otherwise what is the reason of abstraction?
2. ```ActiveDirectoryLdapAuthenticationProvider``` will not work if user provides just basic ```LdapContextSource``` implementation because ```afterPropertiesSet``` must be called before ```getContext```. Also ```setUrl```, ```setUserDn```, ```setPassword``` must be called outside of ```ActiveDirectoryLdapAuthenticationProvider``` and there is also will not be real usage for ```url``` argument of constructor.

In my case (providing just custom SSLSocketFactory) I also will have to provide some custom implementation of ContextSource.

rwinch:
I think we have gotten off track due to the way I worded things initially. Sorry about that.

When I stated:

> In hindsight, I do prefer the option you chose with allowing the user to set any property because it is more passive. However, we cannot make changes to the ContextFactory then.

I was trying to say:

1) Please use setContextEnvironmentProperties as you have in the Pull Request
2) Use the the existing env properties + the contextEnvironmentProperties to pass into the old ContextFactory
3) Please remove all other changes

These changes are preferred because it is more passive than introducing the ContextSource

Does that make more sense?

nE0sIghT:
@rwinch 

> Does that make more sense?

Sure! :-) I pushed changes.

rwinch:
Thanks for the update PR @nE0sIghT! This is merged into master

PR #5694, opened 2018-08-19
Polish Reference Documentation

donhuvy:
- Fix Broken Link
- Update docs to state Spring Security 5 instead of Spring Security 4


rwinch:
Thanks for the PR @donhuvy! This is now merged into master via 1d8db10

Issue #5540, opened 2018-07-18
Add support for getting end_session_endpoint from OIDC Configuration


mraible:
### Summary

Okta has an `end_session_endpoint` property that can be used to construct a global logout that happens on the client. Example code [https://developer.okta.com/blog/2018/06/18/spring-boot-angular-auth-code-flow#logout-with-okta](here). 

It would be nice if Spring Security's OIDC support provided a way to access this `end_session_endpoint`. This property is present in [`./well-known/openid-configuration`](https://dev-669532.oktapreview.com/oauth2/default/.well-known/openid-configuration).

### Version

5.1.0.BUILD-SNAPSHOT

### Sample

https://github.com/oktadeveloper/okta-spring-boot-react-crud-example/compare/spring-security-5.1

jgrandja:
Related #5350

jgrandja:
@mraible Thanks for the report!

There currently is OpenID Connect Discovery code in `OidcConfigurationProvider`. One of our tasks is to extract that bit of code so it can be re-used in other places within Spring Security.

We'll get this in for 5.1

jgrandja:
@mraible This is now available via 057587e.

To access `end_session_endpoint` do this:

`ClientRegistration.getProviderDetails().getConfigurationMetadata().get("end_session_endpoint")`

mraible:
@jgrandja Is it possible to inject the `ClientRegistration` into a RestController? It doesn't seem like it.

rwinch:
@mraible You can pass in the `ClientRegistrationRepository` and look it up. 

```java
@Autowired
public void setClientRegistrationRepository(ClientRegistrationRepository registrations) {
    this.registration = registrations.findByRegistrationId("okta");
}
```

Alternatively you could use `SpeL` to get the `ClientRegistration` itself. 

```java
public void setClientRegistration(
    @Value("#{@clientRegistrationRepository.findByRegistrationId('okta')}") ClientRegistration clientRegistration) {
    this.registration = clientRegistration;
}
```

PR #5729, opened 2018-08-23
ClientRegistration contains Provider Configuration Metadata

jgrandja:
Fixes gh-5540

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
@@ -19,7 +19,6 @@ import org.springframework.security.oauth2.core.AuthenticationMethod;
 import org.springframework.security.oauth2.core.AuthorizationGrantType;
 import org.springframework.security.oauth2.core.ClientAuthenticationMethod;
 import org.springframework.util.Assert;
-import org.springframework.util.CollectionUtils;
 import org.springframework.util.StringUtils;
 
 import java.util.Arrays;
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
@@ -453,7 +453,7 @@ public final class ClientRegistration {
 		 * @return the {@link Builder}
 		 */
 		public Builder providerConfigurationMetadata(Map<String, Object> configurationMetadata) {
-			if (!CollectionUtils.isEmpty(configurationMetadata)) {
+			if (configurationMetadata != null) {
 				this.configurationMetadata = new LinkedHashMap<>(configurationMetadata);
 			}
 			return this;
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
@@ -19,12 +19,14 @@ import org.springframework.security.oauth2.core.AuthenticationMethod;
 import org.springframework.security.oauth2.core.AuthorizationGrantType;
 import org.springframework.security.oauth2.core.ClientAuthenticationMethod;
 import org.springframework.util.Assert;
+import org.springframework.util.CollectionUtils;
 import org.springframework.util.StringUtils;
 
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashSet;
+import java.util.Map;
 import java.util.Set;
 
 /**
@@ -153,6 +155,7 @@ public final class ClientRegistration {
 		private String tokenUri;
 		private UserInfoEndpoint userInfoEndpoint = new UserInfoEndpoint();
 		private String jwkSetUri;
+		private Map<String, Object> configurationMetadata;
 
 		private ProviderDetails() {
 		}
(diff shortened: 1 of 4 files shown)


oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/registration/ClientRegistrationTests.java
```diff
@@ -42,6 +44,13 @@
 	private static final String TOKEN_URI = "https://provider.com/oauth2/token";
 	private static final String JWK_SET_URI = "https://provider.com/oauth2/keys";
 	private static final String CLIENT_NAME = "Client 1";
+	private static final Map<String, Object> PROVIDER_CONFIGURATION_METADATA = new LinkedHashMap<>();
+
+	static {
```
rwinch: Please extract initialization into a method.

rwinch: This isn't truly final because Map is mutable. Please change the Map to be immutable or (preferably) remove the static modifier.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
```diff
@@ -153,6 +155,7 @@ public String toString() {
 		private String tokenUri;
 		private UserInfoEndpoint userInfoEndpoint = new UserInfoEndpoint();
 		private String jwkSetUri;
+		private Map<String, Object> configurationMetadata;
```
rwinch: I think it would be good to make this by default empty so user's don't need to worry about NPE

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
```diff
@@ -262,6 +275,7 @@ public static Builder withRegistrationId(String registrationId) {
 		private AuthenticationMethod userInfoAuthenticationMethod = AuthenticationMethod.HEADER;
 		private String userNameAttributeName;
 		private String jwkSetUri;
+		private Map<String, Object> configurationMetadata;
```
rwinch: I think it would be good to make this by default empty so user's don't need to worry about NPE

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
```diff
@@ -430,6 +444,18 @@ public Builder jwkSetUri(String jwkSetUri) {
...
 
+		/**
+		 * Sets the metadata describing the provider's configuration.
+		 *
+		 * @since 5.1
+		 * @param configurationMetadata the metadata describing the provider's configuration
+		 * @return the {@link Builder}
+		 */
+		public Builder providerConfigurationMetadata(Map<String, Object> configurationMetadata) {
+			this.configurationMetadata = configurationMetadata;
```
rwinch: In combination with making `configurationMetadata` empty, we should also prevent this from be null

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
jgrandja: NPE would never happen with `configurationMetadata` - see where it gets [init](https://github.com/spring-projects/spring-security/pull/5729/files#diff-804db4206dece0e20061c06b32546443R505)

jgrandja:
@rwinch I applied the requested updates to `ClientRegistration`.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
```diff
@@ -430,6 +445,20 @@ public Builder jwkSetUri(String jwkSetUri) {
...
 
+		/**
+		 * Sets the metadata describing the provider's configuration.
+		 *
+		 * @since 5.1
+		 * @param configurationMetadata the metadata describing the provider's configuration
+		 * @return the {@link Builder}
+		 */
+		public Builder providerConfigurationMetadata(Map<String, Object> configurationMetadata) {
+			if (!CollectionUtils.isEmpty(configurationMetadata)) {
```
rwinch: This is not actually setting the value (as the Javadoc states) if it is empty. We should avoid setting the value to null, but we should not avoid setting it to an empty collection if the input is empty.

jgrandja: I applied the update @rwinch

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
```diff
@@ -453,7 +453,7 @@ public Builder jwkSetUri(String jwkSetUri) {
 		 * @return the {@link Builder}
 		 */
 		public Builder providerConfigurationMetadata(Map<String, Object> configurationMetadata) {
-			if (!CollectionUtils.isEmpty(configurationMetadata)) {
+			if (configurationMetadata != null) {
```
rwinch: Can we add a test since this was a problem?

jgrandja: @rwinch I added a couple of tests

rwinch:
@jgrandja The changes look good. Once the Travis build passes, please go ahead and squash and merge.

jgrandja:
Merged via 057587e

PR #5730, opened 2018-08-23
Reactive Resource Server Csrf Bypass

jzheaux:
This makes requests identified as bearer token requests skip the csrf
filter.

Fixes: gh-5710

### DIFF
--- a/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
@@ -43,6 +43,7 @@ import org.springframework.security.authorization.AuthenticatedReactiveAuthoriza
 import org.springframework.security.authorization.AuthorityReactiveAuthorizationManager;
 import org.springframework.security.authorization.AuthorizationDecision;
 import org.springframework.security.authorization.ReactiveAuthorizationManager;
+import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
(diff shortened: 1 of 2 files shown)


config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -1209,6 +1258,7 @@ public CsrfSpec csrfTokenRepository(
 		public CsrfSpec requireCsrfProtectionMatcher(
 			ServerWebExchangeMatcher requireCsrfProtectionMatcher) {
 			this.filter.setRequireCsrfProtectionMatcher(requireCsrfProtectionMatcher);
+			this.specifiedRequireCsrfProtectionMatcher = true;
```
jzheaux: I'm not a big fan of this as I'd prefer it be atomic, but I don't really see another way to tell whether or not the user called this method or not.

I tried introducing a new field to temporarily hold the matcher, and then defer the filter configuration to the `configure` method, but this presents the problem that once `JwtSpec` wants to configure `csrf`, `CsrfSpec#configure` has already been invoked.

This approach at least has the elegance that other Specs can continue to configure the filter after `CsrfSpec#configure` has been invoked.

Open to other ideas, of course.

rwinch: I think this approach is fine. There really isn't a way to avoid adding some sort of state to determine if the default is being used. The state is private, so we can always refactor later too.

rwinch:
Thanks for the PR @jzheaux! This is now merged into master

jzheaux:
Thanks, @rwinch

PR #5736, opened 2018-08-26
Polish ActiveDirectoryLdapAuthenticationProviderTests

izeye:
This PR polishes `ActiveDirectoryLdapAuthenticationProviderTests.contextEnvironmentPropertiesUsed()` by:

- Adding `fail()` to prevent from going through an unexpected path.
- Asserting that the root cause is an instance of `ClassNotFoundException` as the current code doesn't seem to right unless I'm missing something.


rwinch:

Issue #5717, opened 2018-08-21
Allow configurable JwtDecoder for IdToken verification


jgrandja:
We should expose a configuration in the `oauth2Login()` DSL that will allow users to configure a custom `JwtDecoder` to be used for verifying the signature of the `OidcIdToken` during an OpenID Connect flow.

The `OidcAuthorizationCodeAuthenticationProvider` contains the property `private final Map<String, JwtDecoder> jwtDecoders`, which maps `clientRegistrationId` to `JwtDecoder`. This property is currently hidden so there is no way to configure it. We should leverage the same property and existing logic internally, however, the DSL should allow for the configuration of this property.

For example:

```java
oauth2Login().tokenEndpoint().oidcIdTokenDecoder(String clientRegistrationId, JwtDecoder decoder)
```

OR

```java
oauth2Login().tokenEndpoint().oidcIdTokenDecoders(Map<String, JwtDecoder> jwtDecoders)
```

jgrandja:
@rwinch 

> At this time the user does not easily have the information necessary (i.e. the keys) to create the JwtDecoder

My assumption is if the user provides a custom implementation of `JwtDecoder` than they have all the info they need to implement the decoding and signature verification in their implementation. They can than configure the mapping using the `clientRegistrationId` and their `JwtDecoder` impl.

If there is no mapping configured for a `clientRegistrationId` than the existing logic will be applied where we default to `NimbusJwtDecoderJwkSupport`.

I do agree though that a factory method that accepts a `ClientRegistration` and returns a `JwtDecoder` provides greater flexibility. However, using the `Converter` doesn't make sense to me as it's not converting. I prefer `Function` over `Converter`, as is currently implemented in `OidcAuthorizationCodeReactiveAuthenticationManager.DefaultDecoderFactory`. However, is `Function` too generic? I'm not sure we want to introduce a new API at this point either?

rwinch:
> If there is no mapping configured for a clientRegistrationId than the existing logic will be applied where we default to NimbusJwtDecoderJwkSupport.

My point is that if there is 1M client registrations, then with the current approach the developer must lookup 1M client registrations at startup time to perform the mappings. This is not practical. One might say 1M client registrations is unrealistic, but I think very large numbers are possible (especially for a multi tenancy application).

> I do agree though that a factory method that accepts a ClientRegistration and returns a JwtDecoder provides greater flexibility.

Ok seems like we are on the same page here.

[...]

PR #5751, opened 2018-08-28
Allow configurable JwtDecoder for ID Token verification

jgrandja:
This PR resolves gh-5717

@rwinch There is an issue with the changes in this PR. 

Given the following that allows for setting a custom `JwtDecoder` factory:

```java
http
  .oauth2Login()
    .tokenEndpoint()
      .jwtDecoderFactory(Function<ClientRegistration, JwtDecoder> jwtDecoderFactory)
      ...         
```

This introduces the `oauth2-jose` dependency for `oauth2Login` which is not ideal.
I'm not sure what we can do at this point to not include `oauth2-jose` as a required dependency?
Maybe you have an idea?


jgrandja:
@rwinch 

> Perhaps the better approach is allowing the user to configure the AuthenticationProvider instead

Sure this is another option to consider. However, given where we are with the release, this won't make it into 5.1. We'll come back to this after 5.1 is out.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcAuthorizationCodeAuthenticationProvider.java
```diff
@@ -183,29 +196,34 @@ public boolean supports(Class<?> authentication) {
...
-						null
-				);
-				throw new OAuth2AuthenticationException(oauth2Error, oauth2Error.toString());
+	private static class DefaultJwtDecoderFactory implements Function<ClientRegistration, JwtDecoder> {
+		private final Map<String, JwtDecoder> jwtDecoders = new ConcurrentHashMap<>();
+
+		@Override
+		public JwtDecoder apply(ClientRegistration clientRegistration) {
+			JwtDecoder jwtDecoder = this.jwtDecoders.get(clientRegistration.getRegistrationId());
+			if (jwtDecoder == null) {
```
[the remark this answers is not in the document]
jgrandja: Thanks for that catch @jzheaux! Yes, we should use `computeIfAbsent`. I'll apply that update.

rwinch:
I think you are right we need a custom interface. However, I think we should consider a different name. Perhaps `JwtDecoderFactory`?

jgrandja:
I feel `JwtDecoderFactory` is too general given this factory is really meant for client only.  The name `JwtDecoderFactory` can be applied to client and resource server whereas `IdTokenDecoderFactory` is meant for client as far as how a user may interpret the name.

The reason I'm thinking `IdToken` should be in the name is that it's only meant for ID Token verification. I'm not anticipating the use of a `JwtDecoder` for client in any other areas - at least at this moment.

I think `IdTokenDecoderFactory` is fine but I also like `IdTokenDecoders` as well.

jgrandja:
> Do we anticipate needing something similar for the resource server?

Very likely, probably with the multi-tenancy feature we still need to get to

> we make the input generic and stick with something like JwtDecoderFactory

I like this option and provides us the flexibility. I'll go with this

jgrandja:
@rwinch I added the new `JwtDecoderFactory` and `ReactiveJwtDecoderFactory`.

oauth2/oauth2-core/src/test/java/org/springframework/security/oauth2/core/endpoint/TestOAuth2AuthorizationRequests.java
```diff
@@ -32,6 +32,7 @@
 		return OAuth2AuthorizationRequest.authorizationCode()
 				.authorizationUri("https://example.com/login/oauth/authorize")
 				.clientId(clientId)
+				.scope("openid")
```
rwinch: Did you verify this isn't going to break any existing test assumptions? We probably shouldn't modify this as we don't know what impact this is having on existing tests. It may invalidate a test that has different expectations of the scopes.

jgrandja: I did validate that this does not trigger any failed tests. However, it's probably safer to avoid this change and keep the changes isolated to the test that requires the `openid` scope. I'll make this change.

jgrandja: @rwinch I pushed the update for this

Issue #5750, opened 2018-08-28
OAuth2ResourceServerSpec should allow its authentication manager to be configurable


jzheaux:
If a user wanted to supply a custom authentication manager for a Reactive OAuth2 Resource Server, then she would need to create the entire `AuthenticationWebFilter` from scratch.

Since on the Reactive side of Spring Security, authentication managers are typically separated by Authentication type, it makes sense to expose OAuth2 Resource Server's authentication manager to be configurable:

```java
http
    .oauth2ResourceServer()
        .jwt()
            .authenticationManager(customAuthenticationManager())
```

PR #5752, opened 2018-08-28
AuthenticationManager for OAuth2ResourceServerSpec

jzheaux:
This makes the AuthenticationManager used by the OAuth2 Resource
Server configurable, focusing at this point on the Jwt use case.

Fixes: gh-5750

### DIFF
--- a/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
@@ -714,10 +714,22 @@ public class ServerHttpSecurity {
 		 * Configures JWT Resource Server Support
 		 */
 		public class JwtSpec {
+			private ReactiveAuthenticationManager authenticationManager;
 			private ReactiveJwtDecoder jwtDecoder;
 
 			private BearerTokenServerWebExchangeMatcher bearerTokenServerWebExchangeMatcher =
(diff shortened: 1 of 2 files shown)


config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -769,6 +780,17 @@ protected void configure(ServerHttpSecurity http) {
 					.addFilterAt(oauth2, SecurityWebFiltersOrder.AUTHENTICATION);
 			}
 
+			protected ReactiveAuthenticationManager getAuthenticationManager() {
```
rwinch: Can this be private?

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -769,6 +780,17 @@ protected void configure(ServerHttpSecurity http) {
...
+				if (this.authenticationManager == null) {
```
rwinch: It is easier to read if there is less code indented. Instead of a relatively large if block either extract this into a create method or use `if (this.authenticationManager != null) return this.authenticationManager`

rwinch:
Thanks for the PR @jzheaux! This is merged into master

Issue #5223, opened 2018-04-09
JWT Claims Mapping


jzheaux:
### Summary

When a JWT-encoded claim set is handed to Nimbus, the claim set is decoded and then coerced into a chosen Java type for each claim. Specifically:

- `nbf`, `iat`, and `exp` are mapped to `java.util.Date`
- `aud` is mapped to `java.util.List`

Additionally, Nimbus allowed the temporal claims to be null, since the JWT spec allows for this.

Spring Security maps the above fields in the following way:

- `nbf`, `iat` and `exp` are mapped to `Long`
- `iat` and `exp` are given a *default value*
- `aud` is mapped to `java.util.List`

With both libraries, all other claims are left as-is, including custom claims.

For now, the work to transform the Nimbus representation into the Spring Security representation of the claims set is embedded into `NimbusJwtDecoderJwkSupport`, though at least a couple of concerns are likely to surface again and again with other implementations of `JwtDecoder`:

1. The need to map a third-party library's representation of the temporal fields to `Long`

This issue affirms that both Nimbus and Auth0's implementation of the JWT spec coerce the temporal fields into `java.util.Date`. There are likely more libraries that do the same thing, which will require devs to duplicate code that transforms those `Date`s into `Long`s.

2. The need to give `exp` and `iat` default values

`NimbusJwtDecoderJwkSupport` today wants to give `exp` and `iat` default values. If Spring Security intends to enforce this opinion in other implementations (for consistency), then the code needs to not be embedded in the Nimbus-specific implementation.

### Actual Behavior

Certain shared logic is currently embedded in `NimbusJwtDecoderJwkSupport`.

### Expected Behavior

Said logic should be in a more generic location.

### Version

5.1.0.BUILD-SNAPSHOT

PR #5753, opened 2018-08-28
Jwt Claim Mapping

jzheaux:
This introduces a hook for users to customize standard Jwt Claim
values in cases where the JWT issuer isn't spec compliant.

Fixes: gh-5223

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
@@ -1,0 +1,241 @@
+/*
+ * Copyright 2002-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.oauth2.jwt;
+
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URL;
+import java.time.Instant;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.util.Assert;
+
+/**
+ * Converts a JWT claim set, claim by claim. Can be configured with custom converters
+ * by claim name.
+ *
+ * @author Josh Cummings
+ * @since 5.1
+ */
+public final class MappedJwtClaimSetConverter
+		implements Converter<Map<String, Object>, Map<String, Object>> {
+
+	private static final Converter<Object, Collection<String>> AUDIENCE_CONVERTER = new AudienceConverter();
+	private static final Converter<Object, URL> ISSUER_CONVERTER = new IssuerConverter();
+	private static final Converter<Object, String> STRING_CONVERTER = new StringConverter();
+	private static final Converter<Object, Instant> TEMPORAL_CONVERTER = new InstantConverter();
+
+	private final Map<String, Converter<Object, ?>> claimConverters;
+
+	/**
+	 * Constructs a {@link MappedJwtClaimSetConverter} with the provided arguments
+	 *
+	 * This will completely replace any set of default converters.
+	 *
+	 * @param claimConverters The {@link Map} of converters to use
+	 */
+	public MappedJwtClaimSetConverter(Map<String, Converter<Object, ?>> claimConverters) {
+		Assert.notNull(claimConverters, "claimConverters cannot be null");
+		this.claimConverters = new HashMap<>(claimConverters);
+	}
+
+	/**
+	 * Construct a {@link MappedJwtClaimSetConverter}, overriding individual claim
+	 * converters with the provided {@link Map} of {@link Converter}s.
+	 *
+	 * For example, the following would give an instance that is configured with only the default
+	 * claim converters:
+	 *
+	 * <pre>
+	 * 	MappedJwtClaimSetConverter.withDefaults(Collections.emptyMap());
+	 * </pre>
+	 *
+	 * Or, the following would supply a custom converter for the subject, leaving the other defaults
+	 * in place:
+	 *
+	 * <pre>
+	 * 	MappedJwtClaimsSetConverter.withDefaults(
+	 * 		Collections.singletonMap(JwtClaimNames.SUB, new UserDetailsServiceJwtSubjectConverter()));
+	 * </pre>
+	 *
+	 * To completely replace the underlying {@link Map} of converters, {@see MappedJwtClaimSetConverter(Map)}.
+	 *
+	 * @param claimConverters
+	 * @return An instance of {@link MappedJwtClaimSetConverter} that contains the converters provided,
+	 *   plus any defaults that were not overridden.
+	 */
+	public static MappedJwtClaimSetConverter withDefaults
+			(Map<String, Converter<Object, ?>> claimConverters) {
+		Assert.notNull(claimConverters, "claimConverters cannot be null");
+
+		Map<String, Converter<Object, ?>> claimNameToConverter = new HashMap<>();
+		claimNameToConverter.put(JwtClaimNames.AUD, AUDIENCE_CONVERTER);
+		claimNameToConverter.put(JwtClaimNames.EXP, TEMPORAL_CONVERTER);
+		claimNameToConverter.put(JwtClaimNames.IAT, TEMPORAL_CONVERTER);
+		claimNameToConverter.put(JwtClaimNames.ISS, ISSUER_CONVERTER);
+		claimNameToConverter.put(JwtClaimNames.JTI, STRING_CONVERTER);
+		claimNameToConverter.put(JwtClaimNames.NBF, TEMPORAL_CONVERTER);
+		claimNameToConverter.put(JwtClaimNames.SUB, STRING_CONVERTER);
+		claimNameToConverter.putAll(claimConverters);
+
+		return new MappedJwtClaimSetConverter(claimNameToConverter);
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public Map<String, Object> convert(Map<String, Object> claims) {
+		Assert.notNull(claims, "claims cannot be null");
+
+		Map<String, Object> mappedClaims = new HashMap<>(claims);
+
+		for (Map.Entry<String, Converter<Object, ?>> entry : this.claimConverters.entrySet()) {
+			String claimName = entry.getKey();
+			Converter<Object, ?> converter = entry.getValue();
+			if (converter != null) {
+				Object claim = claims.get(claimName);
+				Object mappedClaim = converter.convert(claim);
+				mappedClaims.compute(claimName, (key, value) -> mappedClaim);
+			}
+		}
+
+		Instant issuedAt = (Instant) mappedClaims.get(JwtClaimNames.IAT);
+		Instant expiresAt = (Instant) mappedClaims.get(JwtClaimNames.EXP);
+		if (issuedAt == null && expiresAt != null) {
+			mappedClaims.put(JwtClaimNames.IAT, expiresAt.minusSeconds(1));
+		}
+
+		return mappedClaims;
+	}
+
+	/**
+	 * Coerces an <a target="_blank" href="https://tools.ietf.org/html/rfc7519#section-4.1.3">Audience</a> claim
+	 * into a {@link Collection<String>}, ignoring null values, and throwing an error if its coercion efforts fail.
+	 */
+	private static class AudienceConverter implements Converter<Object, Collection<String>> {
+
+		@Override
+		public Collection<String> convert(Object source) {
+			if (source == null) {
+				return null;
+			}
+
+			if (source instanceof Collection) {
+				return ((Collection<?>) source).stream()
+						.filter(Objects::nonNull)
+						.map(Objects::toString)
+						.collect(Collectors.toList());
+			}
+
+			return Arrays.asList(source.toString());
+		}
+	}
+
+	/**
+	 * Coerces an <a target="_blank" href="https://tools.ietf.org/html/rfc7519#section-4.1.1">Issuer</a> claim
+	 * into a {@link URL}, ignoring null values, and throwing an error if its coercion efforts fail.
+	 */
+	private static class IssuerConverter implements Converter<Object, URL> {
+
+		@Override
+		public URL convert(Object source) {
+			if (source == null) {
+				return null;
+			}
+
+			if (source instanceof URL) {
+				return (URL) source;
+			}
+
+			if (source instanceof URI) {
+				return toUrl((URI) source);
+			}
+
+			return toUrl(source.toString());
+		}
+
+		private URL toUrl(URI source) {
+			try {
+				return source.toURL();
+			} catch (MalformedURLException e) {
+				throw new IllegalStateException("Could not coerce " + source + " into a URL", e);
+			}
+		}
+
+		private URL toUrl(String source) {
+			try {
+				return new URL(source);
+			} catch (MalformedURLException e) {
+				throw new IllegalStateException("Could not coerce " + source + " into a URL", e);
+			}
+		}
+	}
+
+	/**
+	 * Coerces a claim into an {@link Instant}, ignoring null values, and throwing an error
+	 * if its coercion efforts fail.
+	 */
+	private static class InstantConverter implements Converter<Object, Instant> {
+		@Override
+		public Instant convert(Object source) {
+			if (source == null) {
+				return null;
+			}
+
+			if (source instanceof Instant) {
+				return (Instant) source;
+			}
+
+			if (source instanceof Date) {
+				return ((Date) source).toInstant();
+			}
+
+			if (source instanceof Number) {
+				return Instant.ofEpochSecond(((Number) source).longValue());
+			}
+
+			try {
+				return Instant.ofEpochSecond(Long.parseLong(source.toString()));
+			} catch (Exception e) {
+				throw new IllegalStateException("Could not coerce " + source + " into an Instant", e);
+			}
+		}
+	}
+
+	/**
+	 * Coerces a claim into a {@link String}, ignoring null values, and throwing an error if its
+	 * coercion efforts fail.
+	 */
+	private static class StringConverter implements Converter<Object, String> {
+		@Override
+		public String convert(Object source) {
+			if (source == null) {
+				return null;
+			}
+
+			return source.toString();
+		}
+	}
+}
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderJwkSupport.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderJwkSupport.java
@@ -40,6 +40,7 @@ import com.nimbusds.jwt.SignedJWT;
 import com.nimbusds.jwt.proc.ConfigurableJWTProcessor;
 import com.nimbusds.jwt.proc.DefaultJWTProcessor;
 
+import org.springframework.core.convert.converter.Converter;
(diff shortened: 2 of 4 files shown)


jzheaux:
@rwinch What do you think of this as an API for doing claims mapping? I like it because it feels similar to `JwtAuthenticationConverter`.

A little unsure about the name, but I didn't like `JwtBuilder` since it doesn't really follow the builder pattern (just like `JwtAuthenticationConverter` doesn't).

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtCreator.java
```diff
@@ -0,0 +1,91 @@
...
+import java.time.Instant;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * @author Josh Cummings
+ * @since 5.1
+ */
+public class JwtCreator {
```
rwinch: I"m not very keen on using a class for conversion because the whole point is to allow the user to customize the conversion which means they will likely want a new implementation. For that reason, our main hook should be an interface.

I don't like the idea of introducing a custom interface just for this either.

Perhaps the basis could be Converter<Jtw, Jwt> such that we take in the default Jwt but then they can convert it into one that has converted types on headers or claims that they want.

jzheaux:
@rwinch I've implemented two different approaches here, just continuing the conversation we were having on slack.

Here are the tradeoffs I see with the mapped approach (MappedJwtClaimSetConverter):

- Pro: It is flexible enough to handle any claim, including custom ones.
- Con: We commit to a constructor that presumes a mapping approach. I don't really know another way around this other than to expose a setter method for each claim.
- Con: We lose the ability to communicate a type contract for each of the claims, meaning if they performed an incompatible conversion, they would find out at runtime instead of compile time.

And with the sub-class approach (JwtClaimSetConverter):

- Pro: The contract is clear.
- Con: We expose a larger API. However, we can expose these passively on an as-needed basis.
- Con: It doesn't handle custom claims, though I would imagine that a user could easily use a delegate pattern to chain converters, the first being the standard one and the second being their custom claims (and custom implementation of Converter).

Personally, I lean slightly towards JwtClaimSetConverter because I like the type safety, and I also think that the API is simple enough that I'm not worried about exposing a method per standard claim. I also find the mapped converter a bit abstract for my taste.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
```diff
@@ -0,0 +1,83 @@
...
+			this.attributeConverters.put(entry.key, entry.converter);
+		}
+	}
+
+	@Override
+	public Map<String, Object> convert(Map<String, Object> source) {
+		Map<String, Object> mappedClaims = new HashMap<>(source);
+		JwtClaimAccessor claimAccessor = () -> mappedClaims;
+
+		for (Map.Entry<String, Converter<JwtClaimAccessor, ?>> entry : this.attributeConverters.entrySet()) {
```
jzheaux: Iterating through the converters instead of the claims because this allows converters to add and remove values as well as convert them.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
```diff
@@ -0,0 +1,83 @@
...
+			Converter<JwtClaimAccessor, ?> converter = this.attributeConverters.get(entry.getKey());
+			if ( converter != null ) {
+				put(mappedClaims, entry.getKey(), converter.convert(claimAccessor));
+			}
+		}
+
+		return mappedClaims;
+	}
+
+	public static class ConverterEntry {
```
jzheaux: I liked this because it made the constructor very simple. I also find working with var args much simpler than having to build a map and then pass it in.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
```diff
@@ -0,0 +1,83 @@
...
+		public ConverterEntry(String key, Converter<JwtClaimAccessor, ?> converter) {
+			this.key = key;
+			this.converter = converter;
+		}
+	}
+
+	private class IssuedAtConverter implements Converter<JwtClaimAccessor, Instant> {
+
+		@Override
+		public Instant convert(JwtClaimAccessor claimAccessor) {
```
jzheaux: This is necessary so that we continue to convert issuedAt in the same way we have currently released (e.g. falling back to expiresAt).

jzheaux:
@rwinch This is ready for your review

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
```diff
@@ -907,7 +909,7 @@ public void requestWhenClockSkewSetButJwtStillTooLateThenReportsExpired()
 				.andExpect(invalidTokenHeader("Jwt expired at"));
 	}
 
-	// -- converter
+	// -- jwt authentication converter
```
rwinch: These tests shouldn't be added because they are much slower than unit tests and they do not have anything to do with Java Configuration. We should only have tests added to NimbusJwtDecoderJwkSupport with a Mock converter and tests in `MappedJwtClaimSetConverterTests`

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
```diff
@@ -0,0 +1,288 @@
...
+	 * </pre>
+	 *
+	 * To completely replace the underlying {@link Map} of converters, {@see MappedJwtClaimSetConverter(Map)}.
+	 *
+	 * @param attributeConverters
+	 * @return An instance of {@link MappedJwtClaimSetConverter} that contains the converters provided,
+	 *   plus any defaults that were not overridden.
+	 */
+	public static MappedJwtClaimSetConverter withDefaults
+			(Map<String, Converter<Map<String, Object>, ?>> attributeConverters) {
```
rwinch: I prefer to name a map based upon the key and value. For example, `claimIdToConverter`

rwinch: I had imagined the converters to operate on an individual key value pair. By doing this, we could have a standard Instant converter, List<String> converter, etc. Is there a reason/benefit to having it operate on the entire `Map`?

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
```diff
@@ -0,0 +1,288 @@
...
+		private URL toUrl(URI issuer) {
+			try {
+				return issuer.toURL();
+			} catch (MalformedURLException e) {
+				throw new IllegalStateException("Could not coerce iss value to URL");
+			}
+		}
+	}
+
+	private static class IssuedAtConverter extends TemporalConverter {
```
rwinch: Favor composition to inheritance since Java only lets you inherit from a single class. Instead the temporal converter could accept an actual value (instead of the entire Map).

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
```diff
@@ -0,0 +1,288 @@
...
+
+			Instant expiresAt = getExpiresAt(source);
+			if (expiresAt == null) {
+				return issuedAt;
+			}
+
+			return expiresAt.minusSeconds(1);
+		}
+
+		public void setAttributeConverters
```
rwinch: This should likely be a constructor argument since is it is required. A few alternatives to consider is injecting only the required converter or perhaps the converted value.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
```diff
@@ -0,0 +1,288 @@
...
+		}
+
+		@Override
+		public Instant convert(Map<String, Object> source) {
+			Instant issuedAt = super.convert(source);
+			if (issuedAt != null) {
+				return issuedAt;
+			}
+
+			Instant expiresAt = getExpiresAt(source);
```
rwinch: I really view this as something distinct from conversion of the attributes. It is really defaulting the value. I think it might be best to place this in `MappedJwtClaimSetConverter` directly. This will also allow simplification of the API as suggested above

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
jzheaux: Just curious, would these be appropriate here if they were fast and didn't use a mock web server? A number of the tests in this class are there for user demonstration and also verification of a documentable feature.

Adding or changing a claim do not seem like corner cases to me, and so where would we place a test that verifies this configuration combination (or why would we not want to verify it)?

Would it be appropriate to create a ticket to add these tests after we have a NimbusJwtDecoder implementation that can easily mock the Nimbus JwtProcessor (thus making these tests fast and not require a mock web server)?

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
jzheaux: The reason that they take a Map (the claim set) is so that the implementer can access all attributes in making their conversion decision. Taking a Map doesn't prevent having a standard Instant converter, etc., unless I misunderstand your thinking. The PR has TemporalConverter and StringConverter already, for example. Are these less than ideal, and if so, why?

To connect this to another comment that you made (apologies for any noise, but I believe they go together), I think that the decision to have issuedAt fall back to one second before expiresAt is implementation-specific and likely to be overridden. As a user, I would personally prefer to override this to simply accept and convert whatever issuedAt is, and so it's isolated to the conversion of that property.

Given that, an example of where having the entire map available is with issuedAt where we fall back to expiresAt, and thus need to look up expiresAt to correctly derive issuedAt. (See my later comments for more discussion on this specific matter. I only include it here, too, to hopefully make it easier to tie the two points together.)

I agree that converters that simply take and return a single attribute would be simpler. What are your thoughts on future passivity if converters need to look at more than one attribute to make their conversion decision? (Presuming hypothetically that we can find a more elegant way to allow users to override our defaulting of issuedAt, removing the immediate need for this requirement).

As a side note, these originally looked something like this:

```java
withDefaults(Map<String, Converter<JwtClaimAccessor, ?>> overrides) {
  // ...
  withDefaults.put(JwtClaimNames.SUB, accessor -> accessor.getSubject());
  withDefaults.put(JwtClaimNames.JTI, accessor -> accessor.getId());
  // ...
}
```
which may be something still considering, though I think we'd still want the conversion logic to reside here in the converter, even if we use JwtClaimAccessor as the conversion source.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
jzheaux: I agree that this is more about defaults than conversion. I also think that we should focus just on conversion and not concern ourselves with defaulting properties in the first place. My motive, then, for placing this logic in the converter is to make it easy for users to override this defaulting.

Since this has the potential to change the contract (as discussed above), I'd like to hear your thoughts on other ways that users could easily override this defaulting of issuedAt. My concern is that I cannot do the fallback before conversion since it includes date arithmetic. And so, if it is after the set of conversions, then the user needs to completely reimplement the converter (or the JwtDecoder) in order to remove that defaulting.

jzheaux:
@rwinch I simplified the contract to focus on individual attributes. If a user needs the entire map to make a decision, they can just wire an additional converter that delegates to this one and then does whatever other work it needs to do. Let me know what you think.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
```diff
@@ -0,0 +1,232 @@
...
+		for (Map.Entry<String, Converter<Object, ?>> entry : this.attributeConverters.entrySet()) {
+			Converter<Object, ?> converter = this.attributeConverters.get(entry.getKey());
+			if (converter != null) {
+				Object claim = claims.get(entry.getKey());
+				Object mappedClaim = converter.convert(claim);
+				put(mappedClaims, entry.getKey(), mappedClaim);
+			}
+		}
+
+		Instant issuedAt = (Instant) mappedClaims.get(JwtClaimNames.IAT);
```
jzheaux: Separate ticket maybe, but I like the idea of also having IssuedAtDefaultingClaimSetConverter which delegates to MappedJwtSetConverter and performs these steps on the map, e.g.:

```java
public class IssuedAtDefaultingClaimSetConverter implements Converter<...> {
    private final Conveter<...> delegate = MappedJwtClaimSetConverter.withDefaults(...);

    public Map<String, Object> convert(Map<String, Object> claims) {
        Map<String, Object> mappedClaims = this.delegate.convert(claims);

        Instant issuedAt = (Instant) mappedClaims.get(JwtClaimNames.IAT);
        Instant expiresAt = (Instant) mappedClaims.get(JwtClaimNames.EXP);
        if (issuedAt == null && expiresAt != null) {
            mappedClaims.put(JwtClaimNames.IAT, expiresAt.minusSeconds(1));
        }

        return mappedClaims;
}
```

This way, users can easily do

```java
Converter<...> converter = MappedJwtClaimSetConverter.withDefaults(...);
NimbusJwtDecoder#setClaimSetConverter(converter);
```

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
```diff
@@ -0,0 +1,232 @@
...
+
+			try {
+				return Instant.ofEpochSecond(Long.parseLong(temporal.toString()));
+			} catch (Exception e) {
+				throw new IllegalStateException("Could not coerce this value into Instant");
+			}
+		}
+	}
+
+	private static class StringConverter implements Converter<Object, String> {
```
rwinch: Aren't the inputs almost always a String anyways? Is this necessary?

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
```diff
@@ -0,0 +1,232 @@
...
+			} catch (MalformedURLException e) {
+				throw new IllegalStateException("Could not coerce iss value to URL");
+			}
+		}
+
+		private URL toUrl(URI issuer) {
+			try {
+				return issuer.toURL();
+			} catch (MalformedURLException e) {
+				throw new IllegalStateException("Could not coerce iss value to URL");
```
rwinch: Please include the value in the exception message

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
```diff
@@ -0,0 +1,232 @@
...
+			}
+
+			return toUrl(issuer.toString());
+		}
+
+		private URL toUrl(String issuer) {
+			try {
+				return new URL(issuer);
+			} catch (MalformedURLException e) {
+				throw new IllegalStateException("Could not coerce iss value to URL");
```
rwinch: Please include the value in the error message

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
```diff
@@ -0,0 +1,232 @@
...
+			}
+
+			if (temporal instanceof Number) {
+				return Instant.ofEpochSecond(((Number) temporal).longValue());
+			}
+
+			try {
+				return Instant.ofEpochSecond(Long.parseLong(temporal.toString()));
+			} catch (Exception e) {
+				throw new IllegalStateException("Could not coerce this value into Instant");
```
rwinch: Please include the value and the caused by

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
```diff
@@ -0,0 +1,232 @@
...
+	/**
+	 * Constructs a {@link MappedJwtClaimSetConverter} with the provided arguments
+	 *
+	 * This will completely replace any set of default converters.
+	 *
+	 * @param attributeConverters The {@link Map} of converters to use
+	 */
+	public MappedJwtClaimSetConverter(Map<String, Converter<Object, ?>> attributeConverters) {
+		Assert.notNull(attributeConverters, "attributeConverters cannot be null");
+		this.attributeConverters.putAll(attributeConverters);
```
rwinch: I'd prefer to make this more explicit that the attributeConverters does not contain any values already and keep the initialization logic in one place. Please change this to `new HashMap<>(attributeConverters)`

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
```diff
@@ -0,0 +1,232 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.util.Assert;
+
+/**
+ * Converts a JWT claim set attribute by attribute. Can be configured with custom converters
```
rwinch: > by attribute

to

> by attribute name.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
```diff
@@ -0,0 +1,232 @@
...
+	 * {@inheritDoc}
+	 */
+	@Override
+	public Map<String, Object> convert(Map<String, Object> claims) {
+		Assert.notNull(claims, "claims cannot be null");
+
+		Map<String, Object> mappedClaims = new HashMap<>(claims);
+
+		for (Map.Entry<String, Converter<Object, ?>> entry : this.attributeConverters.entrySet()) {
+			Converter<Object, ?> converter = this.attributeConverters.get(entry.getKey());
```
rwinch: You can get the converter using `entry.getValue()` which will be slightly better performance.

rwinch: entry.getKey() is used in multiple places and is not very readable. Please assign to something like attributeName

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
```diff
@@ -0,0 +1,232 @@
...
+		public String convert(Object value) {
+			if (value == null) {
+				return null;
+			}
+
+			return String.valueOf(value);
+		}
+	}
+
+	private void put(Map<String, Object> map, String key, Object value) {
```
rwinch: remove this and use Map.compute directly

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
```diff
@@ -0,0 +1,232 @@
...
+		private URL toUrl(URI issuer) {
+			try {
+				return issuer.toURL();
+			} catch (MalformedURLException e) {
+				throw new IllegalStateException("Could not coerce iss value to URL");
+			}
+		}
+	}
+
+	private static class TemporalConverter implements Converter<Object, Instant> {
```
rwinch: Is there a reason this is called TemporalConverter? It converts to an Instant which is what the other converters are named after.

jzheaux:
Awesome, @rwinch, thanks for the feedback. I've responded with code changes.

Issue #5749, opened 2018-08-28
Reactive HTTPS redirect support


jzheaux:
We should add reactive support similar to what Spring Security already has on the imperative side for upgrading HTTP to HTTPS.

A feature to consider is exchange matching:

```java
http
    .requiresHttps()
        .serverWebExchangeMatchers(matcher1, matcher2)
```

PR #5755, opened 2018-08-30
Reactive Redirect to Https

jzheaux:
This introduces the capability to configure Reactive Spring Security
to upgrade requests to HTTPS

Fixes: gh-5749


web/src/main/java/org/springframework/security/web/server/transport/HttpsRedirectWebFilter.java
```diff
@@ -0,0 +1,112 @@
...
+				new AndServerWebExchangeMatcher(
+						this.defaultNotHttpsMatcher,
+						requiresHttpsRedirectMatcher);
+	}
+
+	private URI calculateRedirectUri(ServerWebExchange exchange) {
+		int port = exchange.getRequest().getURI().getPort();
+
+		UriComponentsBuilder builder =
+				UriComponentsBuilder.fromHttpRequest(exchange.getRequest());
```
rwinch: Do not use fromHttpRequest because it adapts from the HTTP headers. In general, Spring Security requires a ForwardedHeaderFilter or proper proxy server configuration to impact how the request is treated. This is important because we don't want users to not realize that we are looking at headers since this can lead to security exploits. Instead, users must explicitly opt into supporting X-Forwarded using the ForwardedHeaderFilter or proper proxy server configuration

web/src/test/java/org/springframework/security/web/server/transport/HttpsRedirectWebFilterTests.java
```diff
@@ -0,0 +1,138 @@
...
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * Tests for {@link HttpsRedirectWebFilter}
+ *
+ * @author Josh Cummings
+ */
+@RunWith(MockitoJUnitRunner.class)
+public class HttpsRedirectWebFilterTests {
```
rwinch: Please add a few tests that include a path too

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -280,6 +284,11 @@ public ServerHttpSecurity securityContextRepository(ServerSecurityContextReposit
 		return this;
 	}
 
+	public HttpsRedirectSpec redirectToHttps() {
```
rwinch: Please add some javadoc. An example of how you might look for a header only in production so that redirects always happen in production might be useful.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -1210,6 +1222,27 @@ public AuthorizeExchangeSpec access(ReactiveAuthorizationManager<AuthorizationCo
...
 
+	public class HttpsRedirectSpec {
+		private ServerWebExchangeMatcher serverWebExchangeMatcher;
+
+		public HttpsRedirectSpec when(ServerWebExchangeMatcher... matchers) {
+			this.serverWebExchangeMatcher = new OrServerWebExchangeMatcher(matchers);
+			return this;
+		}
+
+		public void configure(ServerHttpSecurity http) {
```
rwinch: this should not be public because it exposes an unnecessary method to the users

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -280,6 +284,41 @@ public ServerHttpSecurity securityContextRepository(ServerSecurityContextReposit
...
+	 * 	    http
+	 * 	        // ...
+	 * 	        .redirectToHttps();
+	 * 	    return http.build();
+	 * 	}
+	 * </pre>
+	 *
+	 * Then all non-HTTPS requests will be redirected to HTTPS.
+	 *
+	 * Typically, all requests should be HTTPS; however, the focus for redirection can also be narrowed:
```
rwinch: Please change this example to be something that we might recommend doing. For example, if a production header is in the request then redirect to HTTPS. This makes it simple to enable HTTPS for production but not other environments that don't have HTTPS support.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -1275,6 +1318,46 @@ public AuthorizeExchangeSpec access(ReactiveAuthorizationManager<AuthorizationCo
...
 	}
 
+	/**
+	 * Configures HTTPS redirection rules
+	 *
+	 * @author Josh Cummings
+	 * @since 5.1
+	 * @see #redirectToHttps()
+	 */
+	public class HttpsRedirectSpec {
```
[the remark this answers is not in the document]
jzheaux: @rwinch Would you recommend this at the top level since likely other filters will want to use it too? e.g. `http.portMapper(PortMapper)` vs `http.redirectToHttps().portMapper(PortMapper)`

PR #5830, opened 2018-09-11
Fix Code Format in Reference

genie-youn:
Please check it. is right?


rwinch:
Thanks for the PR @genie-youn! This is now merged into master

Issue #5861, opened 2018-09-18
Documentation for Secured is truncated


johnlinp:
### Summary

The [documentation for Secured](https://docs.spring.io/spring-security/site/docs/4.2.4.RELEASE/apidocs/org/springframework/security/access/annotation/Secured.html) explains the argument `value` with a truncated sentence.

### Actual Behavior

The documentation says:

```
Returns the list of security configuration attributes (e.g.
```

### Expected Behavior

The document should be:

```
Returns the list of security configuration attributes (e.g. ROLE_USER, ROLE_ADMIN).
```

### Version

4.2.4

PR #5862, opened 2018-09-18
Fix truncated javadoc for Secured

johnlinp:
Fixes: gh-5861


rwinch:
Thanks for the PR @johnlinp! This is now merged into master

PR #5900, opened 2018-09-23
Add "TestSecurityContextHolder#setContext(Authentication)"

ttddyy:
Hi,

While writing tests, I see repeat of followings for programmatic population of `SecurityContext`:

```
  SecurityContext context = SecurityContextHolder.createEmptyContext();
  context.setAuthentication(authentication);
  TestSecurityContextHolder.setContext(context);
```

Most of the time, default SecurityContext(`SecurityContextImpl`) is enough for use.

If `setContext()` method takes `Authentication` object, this can be down to one liner:

```
  TestSecurityContextHolder.setContext(authentication);
```

Existing `setContext(SecurityContext)` allows to pass a custom `SecurityContext` if needed.

So, sending a PR for this.

Thanks,


rwinch:
Thanks for the feedback and the PR @ttddyy! I'm not sure I want to add the additional complexity to the Object for internal usage. Do you find that you are setting the context directly? What are the use cases that you need to do this vs using the built in support?

If we do add an additional method, I think we want to find a different method name. While it is describing what it does, to me it is a bit confusing to call it `setContext` with `Authentication` as an argument. I'm trying to think if there are other examples in Spring or other popular Java projects where this is done. Perhaps it would make more sense to call it `setAutentication`?

ttddyy:
Hi @rwinch

I should've put the background first.
I've been writing internal libraries for other teams/projects.
We have a custom auth mechanism and writing integration for it.
So the intention of this PR is not for internal usage of spring-security itself, but for the actual usage.

For declarative use case, I have a custom `@With...` annotation.

For programmatic usage, people use `TestSecurityContextHolder`.
When populating security context for tests, the implementation of `SecurityContext` is not essential whereas auth token is more on focus. So, having a way of directly setting `Authentication` to `TestSecurityContextHolder` would reduce the repetition of this code.

```
SecurityContext context = SecurityContextHolder.createEmptyContext();
context.setAuthentication(authentication);
TestSecurityContextHolder.setContext(context);
```


I agree with the method naming. I updated the PR with `setAuthentication`.
Also updated this issue title accordingly.


Thanks,

PR #5904, opened 2018-09-26
Fix method name in the Contact sample code

johnlinp:
The method name for handling `/secure/del.htm` was `handleRequest()`, but it should be `delContact()` (similar to `addContact()`).


rwinch:
Thanks for the PR @johnlinp! This is now merged into master

Issue #1388, opened 2009-04-17
SEC-1140: ace masks are not being compared as bitmasks.


spring-projects-issues:
["Brian Ewins":https://jira.spring.io/secure/ViewProfile.jspa?name=bazzargh](Migrated from ["SEC-1140":https://jira.spring.io/browse/SEC-1140?redirect=false]) said:

Granting permissions with bitmasks with more than one bit set does not have the expected result, eg:

```
    MutableAcl acl = (MutableAcl) mutableAclService.readAclById(objectIdentity);
    Permission perm = BasePermission.buildFromMask(BasePermission.CREATE.getMask() | BasePermission.DELETE.getMask());
    acl.insertAce(0, perm, sid, true);
    assert(acl.isGranted(new Permission[]{BasePermission.CREATE}, new Sid[] {sid}, false));
```

As well as the forum link above, this gets mentioned other times in the forum:
http://forum.springsource.org/showthread.php?p=235305
http://forum.springsource.org/showthread.php?t=68655
and as one of the many issues in SEC-479. I'm raising this to split this one issue into a separate bug.

each time its come up, the outcome is that people are reimplementing AclImpl, BasicLookupStrategy etc. In our case we're migrating data from a different acl system, and this results in us having to split up every individual permission in the database, (since otherwise the CumulativePermission loaded never matches). It does seem to defeat the purpose of these things being bitmasks, as well as meaning we need to store 32x as many rows.

spring-issuemaster:
This issue supersedes #740

PR #5908, opened 2018-09-27
extract permission mask comparison for subclasses to override

Metavirulent:
**SEC-2342**
Extracts the mask comparison in DefaultPermissionGrantingStrategy into a protected method.
This allows overriding the default behavior in a subclass without having to copy the whole isGranted() method.
As discussed in the above issue (and others like [SEC-1140](https://github.com/spring-projects/spring-security/issues/1388)), users can choose to implement a different strategy, e.g. do a bitwise mask comparison instead of the integer comparison.

### DIFF
--- a/acl/src/main/java/org/springframework/security/acls/domain/DefaultPermissionGrantingStrategy.java
+++ b/acl/src/main/java/org/springframework/security/acls/domain/DefaultPermissionGrantingStrategy.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


acl/src/main/java/org/springframework/security/acls/domain/DefaultPermissionGrantingStrategy.java
```diff
@@ -142,4 +142,25 @@ public boolean isGranted(Acl acl, List<Permission> permission, List<Sid> sids,
...
+	 * } else {
+	 *    return ace.getPermission().getMask() == p.getMask();
+	 * }
+	 * }</pre>
+	 *
+	 * @param ace the ACE from the Acl holding the mask.
+	 * @param p the Permission we are checking against.
+	 * @return true, if the respective masks are considered to be equal.
+	 */
+	protected boolean comparePermissionMasks(AccessControlEntry ace, Permission p) {
```
rwinch: This method is very implementation specific and would not make sense if a user decides to change the strategy. Please rename to isGranted.

rwinch: Please add a test to demonstrate overriding this method works

rwinch:
Moved to 5.2.x since it is an enhancement

eleftherias:
Thanks for the PR @Metavirulent! This is now merged into master.

Issue #3392, opened 2016-01-03
SEC-3187: LdapUserDetailsManager password change with LDAP operation (RFC 3062)


spring-projects-issues:
[Mark Janssen](https://jira.spring.io/secure/ViewProfile.jspa?name=praseodym) (Migrated from [SEC-3187](https://jira.spring.io/browse/SEC-3187?redirect=false)) said:

Currently the [LdapUserDetailsManager changePassword method](https://github.com/spring-projects/spring-security/blob/1f32e96d31c6a149e480ebdb5ce0274889877995/ldap/src/main/java/org/springframework/security/ldap/userdetails/LdapUserDetailsManager.java#L169-L212) modifies the password attribute directly. It would be better to (optionally) use the [LDAP Password Modify Extended Operation as described in RFC 3062](https://www.ietf.org/rfc/rfc3062.txt). This way, any associated attributes (e.g. Samba NTLM hashed passwords) will also be updated by the LDAP server.

quanah:
This is a security issue, not just an improvement.  By doing a direct modify on the userPassword attribute, the security configuration for userPassword is bypassed.

kopax:
I have clear password stored in my LDAP because of that. Any update on this?

carlspring:
Hi,

Is there any update on this?

fuss86:
+1

sbespalov:
+1

carlspring:
@monowai,
Why the downvotes?

jacovt:
+1

jacovt:
This is raised when putting code through vulnerability analysis tools like Snyk:

https://snyk.io/vuln/SNYK-JAVA-ORGSPRINGFRAMEWORK-31644

carlspring:
Yeah, we're getting it in our snyk.io reports as well!

Could we have an update on this, please?

acooley:
+1.  This needs to be fixed.

steve-todorov:
Are there any updates on this? @tekul, @rwinch, @jgrandja ping?

carlspring:
Is there any update on this? It's been around for a while now and it's causing our snyk.io checks to fail, (which is, of course the least of our concerns, given the seriousness of the issue). Could we get some sort of update, please?

carlspring:
Thanks for fixing this @jzheaux ! In which version do you think this will be released and any idea when that will be? :)

ddillard:
Is there any plan to get a CVE for this issue?

[...]

PR #5922, opened 2018-10-04
Modify Password Extended Operation Support

jzheaux:
Issue: gh-3392


ldap/spring-security-ldap.gradle
```diff
@@ -20,12 +20,15 @@ dependencies {
...
+	testCompile project(':spring-security-test')
 	testCompile 'org.slf4j:jcl-over-slf4j'
 	testCompile 'org.slf4j:slf4j-api'
 }
 
 integrationTest {
-	include('**/ApacheDSServerIntegrationTests.class', '**/ApacheDSEmbeddedLdifTests.class')
+	include('**/ApacheDSServerIntegrationTests.class',
+		'**/ApacheDSEmbeddedLdifTests.class',
+		'**/LdapUserDetailsManagerModifyPasswordTests')
```
rwinch: Is this actually necessary since we have a separate source folder?

ldap/spring-security-ldap.gradle
```diff
@@ -20,12 +20,15 @@ dependencies {
...
 	}
 
+	testCompile project(':spring-security-test')
 	testCompile 'org.slf4j:jcl-over-slf4j'
 	testCompile 'org.slf4j:slf4j-api'
 }
 
 integrationTest {
-	include('**/ApacheDSServerIntegrationTests.class', '**/ApacheDSEmbeddedLdifTests.class')
+	include('**/ApacheDSServerIntegrationTests.class',
```
rwinch: Is this actually necessary since we have a separate source folder?

ldap/src/main/java/org/springframework/security/ldap/userdetails/LdapUserDetailsManager.java
```diff
@@ -414,4 +409,174 @@ public void setGroupMemberAttributeName(String groupMemberAttributeName) {
...
+	 *
+	 * If set to {@code false}, then {@link LdapUserDetailsManager#changePassword} will modify
+	 * the user's password by directly modifying attributes on the corresponding entry.
+	 *
+	 * Before using this setting, ensure that the corresponding LDAP server supports this extended operation.
+	 *
+	 * By default, {@code usePasswordModifyExtensionOperation} is false.
+	 *
+	 * @param usePasswordModifyExtensionOperation
+	 * @since 5.1.1
```
rwinch: We should backport this to 5.0.x and 4.2.x too so we need to change this to 4.2.9

ldap/spring-security-ldap.gradle
jzheaux: The reason, as far as I can tell, is because `ApacheDSServerIntegrationTests` uses the `@Suite` annotation to invoke other dependent integration tests (in the same integration-test directory). If the tests are run independent from this orchestrating test, then they fail because they are dependent on the server that `ApacheDSServerIntegrationTests` starts up.

The password-modify tests are not written in this way, but need to be included in the list since we are already using the `include` directive.

jzheaux: I'm not really a fan of it either. If you think it valuable, I'd recommend a ticket to rewrite the ldap tests to not depend on a parent integration test. Then we could remove this `include`.

rwinch: Ahh yes thanks for the reminder about that. Yes let's do that as a separate ticket.

jzheaux: Done - https://github.com/spring-projects/spring-security/issues/5942

rwinch:
@jzheaux This is ready to merge. Can you please merge and backport it to 4.2.x and 5.0.x?

Issue #5092, opened 2018-03-09
StrictHttpFirewall allows "//path" "//path//path" patterns


rwinch:
### Summary

Backport of #5044

PR #5924, opened 2018-10-05
Reactive Jwt Authentication Converter Support

jzheaux:
Fixes: gh-5092

### DIFF
--- a/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
@@ -38,8 +38,10 @@ import org.springframework.context.ApplicationContext;
 import org.springframework.core.Ordered;
 import org.springframework.core.ResolvableType;
 import org.springframework.core.annotation.AnnotationAwareOrderComparator;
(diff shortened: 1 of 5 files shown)


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/ReactiveJwtAuthenticationConverter.java
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * A reactive {@link Converter} for converting a {@link Jwt} into an {@link AbstractAuthenticationToken}.
+ *
+ * This class is particularly useful in reactive resource servers.
+ *
+ * @author Josh Cummings
+ * @since 5.1.1
+ */
+public class ReactiveJwtAuthenticationConverter
```
rwinch: Since we know that JwtAuthenticationConverter is non blocking we don't need to reimplement this. We could just delegate to JwtAuthenticationConverter but instead perhaps an adapter from imperative to reactive is more valuable since it would support other imperative implementations that don't block

PR #5925, opened 2018-10-06
Reuse SpringSecurityLdapTemplate instance in FilterBasedLdapUserSearch

izeye:
This PR changes to reuse `SpringSecurityLdapTemplate` instance in `FilterBasedLdapUserSearch`.

Along the way, this PR also polishes a bit around it.

### DIFF
--- a/ldap/src/main/java/org/springframework/security/ldap/search/FilterBasedLdapUserSearch.java
+++ b/ldap/src/main/java/org/springframework/security/ldap/search/FilterBasedLdapUserSearch.java
@@ -26,7 +26,6 @@ import org.springframework.dao.IncorrectResultSizeDataAccessException;
 
 import org.springframework.util.Assert;
 
-import org.springframework.ldap.core.ContextSource;
 import org.springframework.ldap.core.DirContextOperations;
 import org.springframework.ldap.core.support.BaseLdapPathContextSource;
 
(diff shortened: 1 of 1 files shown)


ldap/src/main/java/org/springframework/security/ldap/search/FilterBasedLdapUserSearch.java
```diff
@@ -118,9 +117,6 @@ public DirContextOperations searchForUser(String username) {
 					+ this);
 		}
 
-		SpringSecurityLdapTemplate template = new SpringSecurityLdapTemplate(
-				contextSource);
-
 		template.setSearchControls(searchControls);
```
rwinch: We should not be setting a value on a member variable because it introduces a race condition. We either need to to keep this at a method level or we need to restructure the code so that updates to searchControls via the setter methods update the template.

izeye: @rwinch Thanks for the feedback! I overlooked that part and thought it's immutable somehow. Sorry for the noise. I'm closing this for now.

izeye: @rwinch BTW based on the your comment, it looks there's visibility problem with `searchControls` in the current implementation. Am I missing something?

rwinch: I think you are right. It should have been structured differently, but it is likely due to historical reasons.

izeye: @rwinch I see. I'd like to try but I don't have much knowledge on the class to restructure it 😅

PR #5928, opened 2018-10-08
Allow override AbstractPreAuthenticatedProcessingFilter#requiresAuthentication

ttddyy:
Hi,

I'd like to add more precheck logic on `AbstractPreAuthenticatedProcessingFilter#requiresAuthentication`. (Currently, it is a private method)
Similar class `AbstractAuthenticationProcessingFilter` has `requiresAuthentication` as a protected method.

Making the method protected allows sub-classes to extend the precondition check.

Thanks,

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/preauth/AbstractPreAuthenticatedProcessingFilter.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/preauth/AbstractPreAuthenticatedProcessingFilter.java
@@ -34,6 +34,7 @@ import org.springframework.security.core.AuthenticationException;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.security.web.WebAttributes;
 import org.springframework.security.web.authentication.*;
+import org.springframework.security.web.util.matcher.RequestMatcher;
 import org.springframework.util.Assert;
 import org.springframework.web.filter.GenericFilterBean;
 
@@ -73,6 +74,7 @@ import org.springframework.web.filter.GenericFilterBean;
  * @author Luke Taylor
  * @author Ruud Senden
  * @author Rob Winch
+ * @author Tadaya Tsuyukubo
  * @since 2.0
  */
 public abstract class AbstractPreAuthenticatedProcessingFilter extends GenericFilterBean
@@ -86,6 +88,7 @@ public abstract class AbstractPreAuthenticatedProcessingFilter extends GenericFi
 	private boolean invalidateSessionOnPrincipalChange = true;
 	private AuthenticationSuccessHandler authenticationSuccessHandler = null;
 	private AuthenticationFailureHandler authenticationFailureHandler = null;
+	private RequestMatcher requiresAuthenticationRequestMatcher = new PreAuthenticatedProcessingRequestMatcher();
 
 	/**
 	 * Check whether all required properties have been set.
@@ -114,7 +117,7 @@ public abstract class AbstractPreAuthenticatedProcessingFilter extends GenericFi
 					+ SecurityContextHolder.getContext().getAuthentication());
 		}
 
-		if (requiresAuthentication((HttpServletRequest) request)) {
+		if (requiresAuthenticationRequestMatcher.matches((HttpServletRequest) request)) {
 			doAuthenticate((HttpServletRequest) request, (HttpServletResponse) response);
 		}
 
@@ -191,39 +194,6 @@ public abstract class AbstractPreAuthenticatedProcessingFilter extends GenericFi
 				throw failed;
 			}
 		}
-	}
-
-	private boolean requiresAuthentication(HttpServletRequest request) {
-		Authentication currentUser = SecurityContextHolder.getContext()
-				.getAuthentication();
-
-		if (currentUser == null) {
-			return true;
-		}
-
-		if (!checkForPrincipalChanges) {
-			return false;
-		}
-
-		if (!principalChanged(request, currentUser)) {
-			return false;
-		}
-
-		logger.debug("Pre-authenticated principal has changed and will be reauthenticated");
-
-		if (invalidateSessionOnPrincipalChange) {
-			SecurityContextHolder.clearContext();
-
-			HttpSession session = request.getSession(false);
-
-			if (session != null) {
-				logger.debug("Invalidating existing session");
-				session.invalidate();
-				request.getSession();
-			}
-		}
-
-		return true;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


rwinch:
Thanks for the suggestion. Please change this to use a RequestMatcher with the default implementation being private and using the same existing logic. This is preferable since Spring prefers composition to inheritance.

ttddyy:
@rwinch 

Updated the PR to use `RequestMatcher` in `AbstractPreAuthenticatedProcessingFilter`
Thanks,

web/src/main/java/org/springframework/security/web/authentication/preauth/AbstractPreAuthenticatedProcessingFilter.java
```diff
@@ -194,6 +203,11 @@ private void doAuthenticate(HttpServletRequest request, HttpServletResponse resp
 	}
 
 	private boolean requiresAuthentication(HttpServletRequest request) {
+
+		if (!authenticationRequestMatcher.matches(request)) {
```
rwinch: I'm interested in discussing this approach where it augments the existing logic for when authentication is required vs being the entire logic (i.e. the default implementation of RequestMatcher copies the logic from this method). What are your thoughts on each approach?

My concern about augmenting the existing logic is that there is no way to change the existing logic.

ttddyy: You mean performing request matcher check in `requiresAuthentication()` method vs performing the check in `doFilter()`, so that entire logic of this filter may be skipped if request doesn't match?

```java
public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
    throws IOException, ServletException {
    
    if(!requiresAuthenticationRequestMatcher.matches(request)){
      return;
    }
    
    // existing logic from here
    
    if (!requiresAuthentication(request, response)) {
      return;
    }
    ...
}
```


Currently, `requiresAuthentication()` method on this class `AbstractPreAuthenticatedProcessingFilter` (APAPF) is `private`, so where to put the request-matcher check doesn't really affect any other existing sub classes. So, behavior-wise it is same.

The similar class `AbstractAuthenticationProcessingFilter`(AAPF), the request-matcher check is in `requiresAuthentication()` which is a `protected` method.

So, to align the implementation with AAPF, then request-matcher check is in `requiresAuthentication()`.
If we consider request-matcher check is the first-check even before `requiresAuthentication()`, then the check can go into `doFilter()`(shown in above snippet), and maybe update the AAPF as well (updating AAPF will be a behavioral change).

My preference is put the request-matcher check in `requiresAuthentication()` as in current PR which aligns with AAPF, then also make `requiresAuthentication()` protected, so that sub-class can modify the behavior, even it is inheritance. (Yes, I know spring is for composition, but since this is an abstract class which is meant to be inherited and may be overridden.)


> My concern about augmenting the existing logic is that there is no way to change the existing logic.

Currently the method is `private`, adding all-match request-matcher should have no impact to existing logic.

rwinch: Yes. Since the default implementation is internal it would still be able to refer to the member variables.

ttddyy: OK, updated the PR.
Moved the existing auth check logic into the new request matcher.

web/src/main/java/org/springframework/security/web/authentication/preauth/AbstractPreAuthenticatedProcessingFilter.java
```diff
@@ -193,37 +200,8 @@ private void doAuthenticate(HttpServletRequest request, HttpServletResponse resp
...
-
-			if (session != null) {
-				logger.debug("Invalidating existing session");
-				session.invalidate();
-				request.getSession();
-			}
-		}
-
-		return true;
+	protected boolean requiresAuthentication(HttpServletRequest request) {
```
rwinch: There is no need for this to be protected since the authenticationRequestMatcher contains all the logic. In fact, we can remove this method all together since the logic is entirely encapsulated into the matcher.

web/src/main/java/org/springframework/security/web/authentication/preauth/AbstractPreAuthenticatedProcessingFilter.java
```diff
@@ -359,4 +345,46 @@ public void setAuthenticationFailureHandler(AuthenticationFailureHandler authent
 	 * return a dummy value.
 	 */
 	protected abstract Object getPreAuthenticatedCredentials(HttpServletRequest request);
+
+	/**
+	 * Request matcher for default auth check logic
+	 */
+	protected class PreAuthenticatedProcessingRequestMatcher implements RequestMatcher {
```
rwinch: Please make this private

web/src/main/java/org/springframework/security/web/authentication/preauth/AbstractPreAuthenticatedProcessingFilter.java
```diff
@@ -348,6 +326,14 @@ public void setAuthenticationFailureHandler(AuthenticationFailureHandler authent
 		this.authenticationFailureHandler = authenticationFailureHandler;
 	}
 
+	/**
+	 * Sets the request matcher to check whether to proceed the request further.
+	 */
+	public void setAuthenticationRequestMatcher(RequestMatcher authenticationRequestMatcher) {
```
rwinch: rename to setRequiresAuthenticationRequestMatcher to align with AbstractAuthenticationProcessingFilter

web/src/main/java/org/springframework/security/web/authentication/preauth/AbstractPreAuthenticatedProcessingFilter.java
```diff
@@ -86,6 +92,7 @@
 	private boolean invalidateSessionOnPrincipalChange = true;
 	private AuthenticationSuccessHandler authenticationSuccessHandler = null;
 	private AuthenticationFailureHandler authenticationFailureHandler = null;
+	private RequestMatcher authenticationRequestMatcher = new PreAuthenticatedProcessingRequestMatcher();
```
rwinch: rename to requiresAuthenticationRequestMatcher to align with AbstractAuthenticationProcessingFilter

rwinch:
Thanks for the updates. I have provided additional feedback.

ttddyy:
Updated the PR and reflected the feedbacks

rwinch:
Thanks for the fast turnaround! Looks good to me. I'm going to wait for the build to complete before merging into master

eleftherias:
Thanks for the PR @ttddyy! This is now merged into master.

PR #5936, opened 2018-10-10
Add NimbusJwtDecoder

jzheaux:
Introduces a JwtDecoder which takes a raw Nimbus JWTProcessor
configuration.

Fixes: gh-5648

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
@@ -1,0 +1,150 @@
+/*
+ * Copyright 2002-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.oauth2.jwt;
+
+import java.text.ParseException;
+import java.time.Instant;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import com.nimbusds.jose.RemoteKeySourceException;
+import com.nimbusds.jose.proc.SecurityContext;
+import com.nimbusds.jwt.JWT;
+import com.nimbusds.jwt.JWTClaimsSet;
+import com.nimbusds.jwt.JWTParser;
+import com.nimbusds.jwt.SignedJWT;
+import com.nimbusds.jwt.proc.JWTProcessor;
+
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.security.oauth2.core.OAuth2TokenValidator;
+import org.springframework.security.oauth2.core.OAuth2TokenValidatorResult;
+import org.springframework.util.Assert;
+
+/**
+ * A low-level Nimbus implementation of {@link JwtDecoder} which takes a raw Nimbus configuration.
+ *
+ * @author Josh Cummings
+ * @since 5.2
+ */
+public final class NimbusJwtDecoder implements JwtDecoder {
+	private static final String DECODING_ERROR_MESSAGE_TEMPLATE =
+			"An error occurred while attempting to decode the Jwt: %s";
+
+	private final JWTProcessor<SecurityContext> jwtProcessor;
+
+	private Converter<Map<String, Object>, Map<String, Object>> claimSetConverter =
+			MappedJwtClaimSetConverter.withDefaults(Collections.emptyMap());
+	private OAuth2TokenValidator<Jwt> jwtValidator = JwtValidators.createDefault();
+
+	/**
+	 * Configures a {@link NimbusJwtDecoder} with the given parameters
+	 *
+	 * @param jwtProcessor - the {@link JWTProcessor} to use
+	 */
+	public NimbusJwtDecoder(JWTProcessor<SecurityContext> jwtProcessor) {
+		Assert.notNull(jwtProcessor, "jwtProcessor cannot be null");
+		this.jwtProcessor = jwtProcessor;
+	}
+
+	/**
+	 * Use this {@link Jwt} Validator
+	 *
+	 * @param jwtValidator - the Jwt Validator to use
+	 */
+	public void setJwtValidator(OAuth2TokenValidator<Jwt> jwtValidator) {
+		Assert.notNull(jwtValidator, "jwtValidator cannot be null");
+		this.jwtValidator = jwtValidator;
+	}
+
+	/**
+	 * Use the following {@link Converter} for manipulating the JWT's claim set
+	 *
+	 * @param claimSetConverter the {@link Converter} to use
+	 */
+	public void setClaimSetConverter(Converter<Map<String, Object>, Map<String, Object>> claimSetConverter) {
+		Assert.notNull(claimSetConverter, "claimSetConverter cannot be null");
+		this.claimSetConverter = claimSetConverter;
+	}
+
+	/**
+	 * Decode and validate the JWT from its compact claims representation format
+	 *
+	 * @param token the JWT value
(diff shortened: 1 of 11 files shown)


oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -0,0 +1,150 @@
...
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.security.oauth2.core.OAuth2TokenValidator;
+import org.springframework.security.oauth2.core.OAuth2TokenValidatorResult;
+import org.springframework.util.Assert;
+
+/**
+ * A low-level Nimbus implementation of {@link JwtDecoder} which takes a raw Nimbus configuration.
+ *
+ * @author Josh Cummings
+ * @since 5.1.1
```
jgrandja: This cannot be introduced in a patch version. See [Semantic Versioning](https://semver.org/spec/v2.0.0.html) and [Apache Versioning](https://apr.apache.org/versioning.html)

jzheaux:
@rwinch That sounds reasonable. Regarding the builder, there is [a separate ticket](https://github.com/spring-projects/spring-security/issues/6010) for providing this capability in a separate class (not directly on NimbusJwtDecoder). Would it be better to add that builder functionality onto NimbusJwtDecoder and in this PR?

rwinch:
@jzheaux Yes. Let's merge with this issue.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -0,0 +1,280 @@
...
+		return jwt;
+	}
+
+	/**
+	 * A builder for configuring a {@link JwtDecoder}
+	 *
+	 * @author Josh Cummings
+	 * @since 5.2
+	 */
+	public static class Builder {
```
rwinch: Let's rename this to be more precise. When we add support for building with a static key, we want to ensure that it isn't confusing that support isn't on this builder. Perhaps a name of `JwkSetUriBuilder`?

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -0,0 +1,280 @@
...
+		private Builder() {}
+
+		/**
+		 * Use the given
+		 * <a href="https://tools.ietf.org/html/rfc7517#section-5">JWK Set</a> uri.
+		 *
+		 * @param jwkSetUri the JWK Set uri to use
+		 * @return a {@link Builder} for further configurations
+		 */
+		public static Builder withJwkSetUri(String jwkSetUri) {
```
rwinch: Let's move this method outside of the builder to the class level.

rwinch: I think we should add validation on jwkSetUri. At minimum it should ensure that the String is not null or empty.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
rwinch: It is a bit surprising to me that the `Builder` only provides the properties to build the `JWTProcessor` for the constructor to `NimbusJwtDecoder`. After using the `Builder` to create `NimbusJwtDecoder`, I'm not sure I would think to look for properties on `NimbusJwtDecoder` (i.e. `setClaimSetConverter`) for additional customization.

I wonder if the `Builder` should either
* Have all the properties on it
* Build a `JWTProcessor` instead

Other options are welcome

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderJwkSupport.java
```diff
@@ -62,27 +53,23 @@
...
  * @author Josh Cummings
  * @since 5.0
  * @see JwtDecoder
+ * @see NimbusJwtDecoder
  * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7519">JSON Web Token (JWT)</a>
  * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7515">JSON Web Signature (JWS)</a>
  * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7517">JSON Web Key (JWK)</a>
  * @see <a target="_blank" href="https://connect2id.com/products/nimbus-jose-jwt">Nimbus JOSE + JWT SDK</a>
  */
 public final class NimbusJwtDecoderJwkSupport implements JwtDecoder {
```
rwinch: I think if we have done things well in `NimbusJwtDecoder` then we should be able to delete more of this class. For example, I don't think we should need `RestOperationsResourceRetriever`. A way to minimize this would be to have a member of the `Builder` and a member of the `NimbusJwtDecoder` If a setter is invoked, it updates the builder and then builds a new delegate `NimbusJwtDecoder`

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
jzheaux: I like the idea of introducing additional properties, and in fact, I would prefer that when constructing these myself. I prefer treating my objects as immutable after using a builder.

I think that building a `JWTProcessor` could work, but it also might encourage a very Nimbus-like API. While this is indeed a Nimbus-specific class, I'm not sure I want to position this as an authoritative way to configure Nimbus primitives.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -0,0 +1,280 @@
...
+			ConfigurableJWTProcessor<SecurityContext> jwtProcessor = new DefaultJWTProcessor<>();
+			jwtProcessor.setJWSKeySelector(jwsKeySelector);
+
+			// Spring Security validates the claim set independent from Nimbus
+			jwtProcessor.setJWTClaimsSetVerifier((claims, context) -> { });
+
+			return new NimbusJwtDecoder(jwtProcessor);
+		}
+
+		private static URL toURL(Object url) {
```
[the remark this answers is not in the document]
jzheaux: I can't remember now. It was needed in an early draft, but it appears I forgot to clean it up.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -0,0 +1,280 @@
...
+			this.restOperations = restOperations;
+			return this;
+		}
+
+		/**
+		 * Build the configured {@link JwtDecoder}.
+		 *
+		 * @return the configured {@link JwtDecoder}
+		 */
+		public NimbusJwtDecoder build() {
```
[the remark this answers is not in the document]
jzheaux: Note, @jgrandja in the latest PR that the builder now returns `JWTProcessor` instead.

Though, I was debating the same thing as you at the time.

jzheaux:
@rwinch, @jgrandja Thank you both for the valuable feedback. I've updated based on it.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -0,0 +1,280 @@
...
+
+		try {
+			// Verify the signature
+			JWTClaimsSet jwtClaimsSet = this.jwtProcessor.process(parsedJwt, null);
+
+			Map<String, Object> headers = new LinkedHashMap<>(parsedJwt.getHeader().toJSONObject());
+			Map<String, Object> claims = this.claimSetConverter.convert(jwtClaimsSet.getClaims());
+
+			Instant expiresAt = (Instant) claims.get(JwtClaimNames.EXP);
+			Instant issuedAt = (Instant) claims.get(JwtClaimNames.IAT);
```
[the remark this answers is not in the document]
jzheaux: Added https://github.com/spring-projects/spring-security/pull/5936/files#diff-bf492c57db9d7f365bc14fb30572ed16R102

rwinch:
Thanks for the PR @jzheaux! This is now merged into master

PR #5941, opened 2018-10-11
Use more specific GrantedAuthority for mapped roles

### LINKED ISSUE #5929: Use more specific LdapAuthority in DefaultLdapAuthoritiesPopulator
### Summary
When using `DefaultLdapAuthoritiesPopulator`, the set of `GrantedAuthorities` contains instances of `SimpleGrantedAuthority` while I would expect instances of `LdapAuthority`; as it is now, I cannot discriminate LDAP authorities from others.

`NestedLdapAuthoritiesPopulator` already uses the more specific `LdapAuthority`.

### Actual Behavior

At [DefaultLdapAuthoritiesPopulator.java#L255](https://github.com/spring-projects/spring-security/blob/2307b01a7a710caf1fe5619bd9db5ac28caebabe/ldap/src/main/java/org/springframework/security/ldap/userdetails/DefaultLdapAuthoritiesPopulator.java#L255) the authority set is populated wiht:

```java
authorities.add(new SimpleGrantedAuthority(this.rolePrefix + role));
```

### Expected Behavior

I would expect the authority set to be populated with:

```java
authorities.add(new LdapAuthority(this.rolePrefix + role, roleDn));
```

### Version

Spring Security 5.1.0

heruan:
As discussed in #5929 this adds an implementation of `LdapAuthoritiesPopulator` which uses `LdapAuthority` instead or the generic `SimpleGrantedAuthority`.

The new implementation shares the same code as the original so that the old one can be deprecated and then removed.

Keep in mind that `NestedLdapAuthoritiesProvider` still extends `DefaultLdapAuthoritiesProvider`.

## Alternative
Change `DefaultLdapAuthoritiesPopulator` implementation to use `LdapAuthority` instead of `SimpleGrantedAuthority`: this would be a "breaking" change if and only if an existing logic is based on the fact that authorities from the populator are instances of `SimpleGrantedAuthority` and I can't figure out any case where this would make sense. Of course this will wait a next major, but would not introduce an additional implementation that might create confusion and introduce duplicated code.

Fixes #5929

### DIFF
--- a/ldap/src/integration-test/java/org/springframework/security/ldap/userdetails/DefaultLdapAuthoritiesPopulatorTests.java
+++ b/ldap/src/integration-test/java/org/springframework/security/ldap/userdetails/DefaultLdapAuthoritiesPopulatorTests.java
@@ -18,9 +18,12 @@ package org.springframework.security.ldap.userdetails;
 
 import static org.assertj.core.api.Assertions.*;
 
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+
 import org.junit.*;
 import org.junit.runner.RunWith;
-
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.ldap.core.ContextSource;
 import org.springframework.ldap.core.DirContextAdapter;
@@ -29,10 +32,9 @@ import org.springframework.ldap.core.DistinguishedName;
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.authority.AuthorityUtils;
 import org.springframework.security.ldap.ApacheDsContainerConfig;
+import org.springframework.security.ldap.SpringSecurityLdapTemplate;
 import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.junit4.SpringRunner;
-
-import java.util.*;
 
 /**
  *
@@ -184,5 +186,26 @@ public class DefaultLdapAuthoritiesPopulatorTests {
 
 		assertThat(authorities).as("Should have 1 role").hasSize(1);
 		assertThat(authorities.contains("ROLE_MANAGER")).isTrue();
+	}
+
+	@Test
+	public void customAuthoritiesMappingFunction() {
+		populator.setAuthorityMapper(record -> {
+			String dn = record.get(SpringSecurityLdapTemplate.DN_KEY).get(0);
+			String role = record.get(populator.getGroupRoleAttribute()).get(0);
+			return new LdapAuthority(role, dn);
+		});
+
+		DirContextAdapter ctx = new DirContextAdapter(new DistinguishedName(
+				"cn=mouse\\, jerry,ou=people,dc=springframework,dc=org"));
+
+		Collection<GrantedAuthority> authorities = populator.getGrantedAuthorities(ctx, "notused");
+
+		assertThat(authorities).allMatch(LdapAuthority.class::isInstance);
+	}
+
+	@Test(expected = IllegalArgumentException.class)
+	public void customAuthoritiesMappingFunctionThrowsIfNull() {
+		populator.setAuthorityMapper(null);
 	}
 }
--- a/ldap/src/main/java/org/springframework/security/ldap/userdetails/DefaultLdapAuthoritiesPopulator.java
+++ b/ldap/src/main/java/org/springframework/security/ldap/userdetails/DefaultLdapAuthoritiesPopulator.java
@@ -20,13 +20,14 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import java.util.function.Function;
 
 import javax.naming.directory.SearchControls;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-
 import org.springframework.ldap.core.ContextSource;
 import org.springframework.ldap.core.DirContextOperations;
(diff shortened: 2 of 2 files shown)


rwinch:
Thanks for the PR! 

I'm sorry but after looking at this PR I think I have steered you in the wrong direction. There is a lot of code duplication and it doesn't solve the problem for NestedLdapAuthoritiesProvider which leads me to think we need a better solution.

I'm wondering if the existing DefaultLdapAuthoritiesProvider could have a ContextMapper (or something similar) injected into it that determines the type that is extracted out. I haven't been able to look into this extensively, but I feel like that might be the best (and most flexible) approach.

Thoughts?

heruan:
Yeah, I agree we need a better solution. One might be introducing a mapper as you said in `DefaultLdapAuthoritiesProvider`, which by default builds a `SimpleGrantedAuthority` (non-breaking) but can be customized to build the more specific `LdapAuthority`. E.g.

```java
var populator = new DefaultLdapAuthoritiesPopulator();
populator.setAuthorityMapper(record -> {
	var roleDn = record.get(SpringSecurityLdapTemplate.DN_KEY).get(0);
	var role = record.get(this.groupRoleAttribute).get(0);
	return new LdapAuthority(role, roleDn);
});
```

In either case `NestedLdapAuthoritiesMapper` wouldn't be affected as it already returns instances of `LdapAuthority`.

When time comes for a breaking change then, the default mapper could be changed to return the more correct `LdapAuthority`.

heruan:
I don't think I've introduced formatting changes (I just added a couple of missing blank lines). About tests, the existing test already covers the default mapping function, or am I missing something?

rwinch:
> I don't think I've introduced formatting changes (I just added a couple of missing blank lines).

My mistake on that. There was code changes there too. Please ignore.

> About tests, the existing test already covers the default mapping function, or am I missing something?

Yes they do. However, it would be good to add a test to ensure that if a custom authority mapper is set, then the behavior changes (i.e. add a test for your specific use case of returning an LdapAuthority). Also add a test for setting a null mapper to ensure an exception is thrown.

heruan:
Test added!

rwinch:
Thanks for the test! It appears the checkstyle is failing. Could you please run `./gradlew checkstyleMain checkstyleTest` and fix the checkstyle errors?

heruan:
Fixed! It was a couple of unused imports, not sure how they ended up in a file I didn't modify in the latest push (and the previous passed the checks)!

heruan:
@rwinch not sure if anything is pending here from my side!

eleftherias:
@heruan thanks for the PR! 
It appears that the tests you added are failing.  

The one test `customAuthoritiesMappingFunctionThrowsIfNull` is expecting a `NullPointerException`, but really it should expect an `IllegalArgumentException`.  

The other test `customAuthoritiesMappingFunction` is checking that a list of `String`s  are instances of `LdapAuthority`, which will not be true. If you remove the transformation `AuthorityUtils.authorityListToSet`, then you should get the expected result.  

Let me know if you can update the tests and we can get this PR merged.

ldap/src/integration-test/java/org/springframework/security/ldap/userdetails/DefaultLdapAuthoritiesPopulatorTests.java
```diff
@@ -185,4 +187,26 @@ public void userDnWithEscapedCharacterParameterReturnsExpectedRoles() {
...
+		DirContextAdapter ctx = new DirContextAdapter(new DistinguishedName(
+				"cn=mouse\\, jerry,ou=people,dc=springframework,dc=org"));
+
+		Set<String> authorities = AuthorityUtils.authorityListToSet(populator
+				.getGrantedAuthorities(ctx, "notused"));
+
+		assertThat(authorities).allMatch(LdapAuthority.class::isInstance);
+	}
+
+	@Test(expected = NullPointerException.class)
```
eleftherias: This should be an `IllegalArgumentException`.

ldap/src/integration-test/java/org/springframework/security/ldap/userdetails/DefaultLdapAuthoritiesPopulatorTests.java
```diff
@@ -185,4 +187,26 @@ public void userDnWithEscapedCharacterParameterReturnsExpectedRoles() {
...
+		populator.setAuthorityMapper(record -> {
+			String dn = record.get(SpringSecurityLdapTemplate.DN_KEY).get(0);
+			String role = record.get(populator.getGroupRoleAttribute()).get(0);
+			return new LdapAuthority(role, dn);
+		});
+
+		DirContextAdapter ctx = new DirContextAdapter(new DistinguishedName(
+				"cn=mouse\\, jerry,ou=people,dc=springframework,dc=org"));
+
+		Set<String> authorities = AuthorityUtils.authorityListToSet(populator
```
eleftherias: Right now the authorities are transformed to `String`s so they will not be instances of `LdapAuthority`. If you remove `AuthorityUtils.authorityListToSet` then the test should work correctly.

heruan:
Thank you @eleftherias for the hints, I'll fix the tests: let's see if they pass!

eleftherias:
@heruan The tests look good now, thank you. One last request, could you please squash your 2 commits into one. That will make the commit history clearer when this is merged.

heruan:
Done! I was confident they were going to be squashed on merge 🙂

eleftherias:
Thanks for the PR @heruan! This is now merged into master.

PR #5962, opened 2018-10-15
Add postProcess call to HttpSecurity.oauth2Login()

bdemers:
This matches the pattern oauth2Client() and oauth2ResourceServer() use

NOTE: I'm not sure this is approprate or not for this method.  I went looking for a way to customize the `OAuth2LoginConfigurer` from an auto configuration and realized this didn't follow the pattern of the other two.

A PR seemed like the easiest way to discuss it, feel free to close otherwise.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
@@ -989,7 +989,9 @@ public final class HttpSecurity extends
 	 * @throws Exception
 	 */
 	public OAuth2LoginConfigurer<HttpSecurity> oauth2Login() throws Exception {
-		return getOrApply(new OAuth2LoginConfigurer<>());
+		OAuth2LoginConfigurer configurer = getOrApply(new OAuth2LoginConfigurer<>());
+		this.postProcess(configurer);
+		return configurer;
 	}
 
 	/**


pivotal-issuemaster:
@bdemers Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=5962)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=5962) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@bdemers Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=5962)!

rwinch:
Thanks for the PR!

Can you elaborate more on what you are trying to do? You could easily just invoke oauth2Client() in your own configuration to achieve the same thing.

bdemers:
Hey @rwinch 

I was trying to figure out how to _best_ update the configuration for down stream consumers (i.e. my special `super-cool-oauth2` lib adds an auto config that decorates or modifies the configuration of the OAuth configurer.

@jgrandja pointed me to the [new docs](https://docs.spring.io/spring-security/site/docs/5.1.1.RELEASE/reference/htmlsingle/#oauth2login-override-boot-autoconfig). So that helps!

If seems like I should be able to change the default oauth config outside of a `WebSecurityConfigurerAdapter`.  Possibly through an ObjectPostProcessor or supplying a different instance to be loaded when something like `HttpSecurity.getOrApply(new OAuth2LoginConfigurer<>())` is called.


A more simple use case might be to swap out the JwtDecoder instance by simply supplying another jar on the classpath.

rwinch:
> If seems like I should be able to change the default oauth config outside of a WebSecurityConfigurerAdapter. Possibly through an ObjectPostProcessor or supplying a different instance to be loaded when something like HttpSecurity.getOrApply(new OAuth2LoginConfigurer<>()) is called.

See if this helps https://docs.spring.io/spring-security/site/docs/5.0.x/reference/htmlsingle/#jc-custom-dsls

> A more simple use case might be to swap out the JwtDecoder instance by simply supplying another jar on the classpath.

You can have create a JwtDecoder exposed as a bean and that would be picked up

bdemers:
Thanks again @rwinch!

I was trying to customize the `OAuth2UserService` and decorate the user with extra authorities based on the claims from the current token. I'm starting to think the custom dsl might be the way to go, as chances are end user app will have a `WebSecurityConfigurerAdapter` anyway. 🤔 

I just realized the bit about the JwtDecoder bean (after posting this), I'm really digging the `Oauth2TokenValidator` and `DelegatingOAuth2TokenValidator`, it should be cake to add custom validation!

jgrandja:
@bdemers 

> I was trying to customize the OAuth2UserService and decorate the user with extra authorities based on the claims from the current token

You might want to check out [Delegation-based strategy with OAuth2UserService](https://docs.spring.io/spring-security/site/docs/5.1.1.RELEASE/reference/htmlsingle/#oauth2login-advanced-map-authorities-oauth2userservice)

Issue #4841, opened 2017-11-16
Support Mono<Boolean> for Method Security SpEL expressions


rwinch:
### Summary

Reactive method security requires the SpEL expression to return Boolean which does not work of the logic to determine access is blocking. We should allow the result to be `Mono<Boolean>`

edeandrea:
Based on [your comment on this issue](https://github.com/spring-projects/spring-security/issues/5046#issuecomment-427045308) the fact that this ticket remains open does that mean it is not possible to do this? Do the methods referenced within the expression in the `@PreAuthorize` annotations still need to return `boolean` over `Mono<Boolean>`, therefore meaning you can't perform any blocking operations within the method?

```java
@Component
public class SomeBean {
  public Mono<Boolean> someMethod() {
    return Mono.fromRunnable(() -> Mono.just(true));
  }
}

@RestController
public class SomeController {
  @GetMapping("/somePath")
  @PreAuthorize("@someBean.someMethod()")
  public SomePojo getSomething() {
    return new SomePojo();
  }
}
```

edeandrea:
I believe I've answered my own question and the answer is that I can not do what I posted above.

Seems what is missing in Spring Security are reactive versions of `ExpressionBasedPreInvocationAdvice` & `ExpressionBasedPostInvocationAdvice`.

Am I correct in that assessment?

rwinch:
That and we need a reactive equivalent of `PrePostAdviceReactiveMethodInterceptor`

edeandrea:
Isn't `PrePostAdviceReactiveMethodInterceptor` already reactive? Seems like its reactive in the sense that its looking in the `ReactiveSecurityContextHolder` for the `Authentication`, but still calling the non-reactive `PreInvocationAuthorizationAdvice`/`PostInvocationAuthorizationAdvice`. The work `Reactive` in the class name is misleading :)

rwinch:
> Isn't PrePostAdviceReactiveMethodInterceptor already reactive? 

You are right but it would need to accept the reactive advice.

> Would it be better to create a new version of PrePostAdviceReactiveMethodInterceptor or to modify the existing one to support the reactive pre/post advice classes?

Modify the existing one, deprecate the existing constructor (modify it to just adapt to use the reactive APIs to simplify the code), add a new constructor w/ reactive equivalents

> Would you be open to a PR for this or is it something you are already actively working on?

I'd love a PR for this.

edeandrea:
[...]

PR #5980, opened 2018-10-16
Support reactive method security with reactive types

edeandrea:
From #4841 - copying in last comment from that issue into this PR, while adding some comments. What is in this PR is probably not the final version - there will probably be some re-work, but I want to settle on a few design decisions before re-working a bit:

   > We don't want to remove tests that already exist (i.e. EnableReactiveMethodSecurityTests) This runs the risk of breaking backwards compatibility.

   > Focus on as few changes as possible. For example, ReactiveMessageService has methods removed from it.

In my opinion - the tests that were removed don't really make sense for reactive types. Things like

```java
@PostAuthorize("returnObject?.contains(authentication?.name)")
```

when the return type is a `Mono` / `Flux` / `Publisher` don't make a whole lot of sense. Spring-el doesn't yet support lambda functions within expressions so calling methods on those types within an expression doesn't make much sense to me.

I also added a ton of new functional/integration (see [ReactiveMethodSecurityConfigurationTests](https://github.com/spring-projects/spring-security/pull/5980/files#diff-1b6bcc9a28eb0c2969367060081ef646R51)) tests to cover lots of the scenarios.

   > We cannot [remove a constructor](https://github.com/edeandrea/spring-security/commit/00d8b1a7a48e467da8b77434f9755c7a9f07668c#diff-f31832b1006ea62f7afe660bbeb60f72L61) because it is a breaking change. The old constructor should be adapted to work with the new reactive APIs and a new constructor should be added.

I'm unsure of how to "adapt" the old constructor and its arguments (`PreInvocationAuthorizationAdvice` and `PostInvocationAuthorizationAdvice`) to work with the new reactive api (`ReactivePreInvocationAuthorizationAdvice` and `ReactivePostInvocationAuthorizationAdvice`).
- Their method signatures are completely different (blocking types vs reactive types)
- Functionally their implementations are very different (blocking (using `ExpressionUtils`) vs non-blocking (using `ReactiveExpressionUtils`))
- It isn't enough to change these arguments - the body of the `invoke` method needs to change as well in order to work with the reactive streams

What I ended up doing was leaving `PrePostAdviceReactiveMethodInterceptor` alone & marking the class itself as deprecated and introducing a new `ReactivePrePostAdviceMethodInterceptor` class which uses the new reactive pre/post advice implementations.

If you have a clearer picture of how to adapt the existing pre/post advice interfaces into the new reactive ones then I'm all ears - I just don't see what they have in common to be able to programmatically adapt the old ones to the new ones.

   > We do not want to make SimpleMethodInvocation mutable. This adds concerns around concurrency

I made that change because of the way the `MethodInvocation` itself works. In the pre-advice scenario we need to be able to alter the arguments that are "fed" into `MethodInvocation.proceed()`. In the blocking scenario, `DefaultMethodSecurityExpressionHandler.filter` actually mutates the arguments _in place_, assuming the argument is a `Collection`. In the reactive world we don't have that luxury - the reactive types are all immutable and can't be modified _in place_. So in order to get around that we need to be able to supply a different set of arguments completely (i.e. a `Mono` / `Flux` with a `filterWhen` attached to it).

The only solution I was able to come up with was in [`ExpressionBasedReactivePreInvocationAuthorizationAdvice.setArguments`](https://github.com/spring-projects/spring-security/commit/2d75b19b0d972a443fc209cc2bac5ac8efa1b433#diff-14e767a0e12b05e6d98b272bb0eb2bf9R121). I thought it best to be able to support all the various implementations of `MethodInvocation` as possible (not just `ProxyMethodInvocation`). That's why I added the `setArguments` method to `SimpleMethodInvocation`.

Not to mention all the tests I wrote for testing all this functionality use a sub-class of `SimpleMethodInvocation`, so without it none of my tests will work either.

I can refactor a bit and take `SimpleMethodInvocation` out of the picture but then the whole pre-advice will only work correctly if the `MethodInvocation` is a `ProxyMethodInvocation`. I don't know enough about the rest of the codebase to know whether that is acceptable or not.

The only other solution I thought about but I feel would be a somewhat larger undertaking would be to change the implementation of the `ReactivePrePostAdviceMethodInterceptor` altogether and use the AspectJ classes (`ProceedingJoinPoint`) instead of the Spring AOP `MethodInvocation`. From there instead of only having a single `MethodInvocation.proceed()` method, there would be a `ProceedingJoinPoint.proceed(Object[])` method that we could use to supply different arguments. This would also involve some re-structuring of the way the pre-advice class works as well, since right now it only has a single `before` method that returns `Mono<Boolean>`. I think we would instead need 2 methods - 1 for `beforeAuthorize` and one for `beforeFilter`.

I didn't go that route originally though because I didn't see AspectJ being used anywhere and I wanted to keep the implementation on the reactive side as close as possible to the non-reactive side.

I'm open to ideas though - just wanted to put some context into why I made some of the decisions that I did.

### DIFF
--- a/config/src/test/java/org/springframework/security/config/annotation/method/configuration/DelegatingReactiveMessageService.java
+++ b/config/src/test/java/org/springframework/security/config/annotation/method/configuration/DelegatingReactiveMessageService.java
@@ -16,10 +16,12 @@
 package org.springframework.security.config.annotation.method.configuration;
 
 import org.reactivestreams.Publisher;
-import org.springframework.security.access.prepost.PostAuthorize;
-import org.springframework.security.access.prepost.PreAuthorize;
+
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
+
+import org.springframework.security.access.prepost.PostAuthorize;
+import org.springframework.security.access.prepost.PreAuthorize;
 
 public class DelegatingReactiveMessageService implements ReactiveMessageService {
 	private final ReactiveMessageService delegate;
@@ -47,7 +49,7 @@ public class DelegatingReactiveMessageService implements ReactiveMessageService 
 	}
 
 	@Override
-	@PostAuthorize("returnObject?.contains(authentication?.name)")
+	@PostAuthorize("@authz.containsAuthenticationName(returnObject, authentication?.name)")
 	public Mono<String> monoPostAuthorizeFindById(
 			long id) {
 		return delegate.monoPostAuthorizeFindById(id);
@@ -80,7 +82,7 @@ public class DelegatingReactiveMessageService implements ReactiveMessageService 
 	}
 
 	@Override
-	@PostAuthorize("returnObject?.contains(authentication?.name)")
+	@PostAuthorize("@authz.containsAuthenticationName(returnObject, authentication?.name)")
 	public Flux<String> fluxPostAuthorizeFindById(
 			long id) {
 		return delegate.fluxPostAuthorizeFindById(id);
@@ -113,7 +115,7 @@ public class DelegatingReactiveMessageService implements ReactiveMessageService 
 	}
 
 	@Override
-	@PostAuthorize("returnObject?.contains(authentication?.name)")
+	@PostAuthorize("@authz.containsAuthenticationName(returnObject, authentication?.name)")
 	public Publisher<String> publisherPostAuthorizeFindById(
 			long id) {
 		return delegate.publisherPostAuthorizeFindById(id);
--- a/config/src/test/java/org/springframework/security/config/annotation/method/configuration/EnableReactiveMethodSecurityTests.java
+++ b/config/src/test/java/org/springframework/security/config/annotation/method/configuration/EnableReactiveMethodSecurityTests.java
@@ -16,11 +16,21 @@
 
 package org.springframework.security.config.annotation.method.configuration;
 
+import static org.assertj.core.api.Assertions.*;
+import static org.mockito.Mockito.*;
+
 import org.assertj.core.api.AssertionsForClassTypes;
 import org.junit.After;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.reactivestreams.Publisher;
+
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+import reactor.test.StepVerifier;
+import reactor.test.publisher.TestPublisher;
+import reactor.util.context.Context;
+
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Bean;
 import org.springframework.security.access.AccessDeniedException;
@@ -28,14 +38,6 @@ import org.springframework.security.authentication.TestingAuthenticationToken;
 import org.springframework.security.core.context.ReactiveSecurityContextHolder;
 import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.junit4.SpringRunner;
-import reactor.core.publisher.Flux;
-import reactor.core.publisher.Mono;
-import reactor.test.StepVerifier;
-import reactor.test.publisher.TestPublisher;
-import reactor.util.context.Context;
-
-import static org.assertj.core.api.Assertions.assertThatThrownBy;
-import static org.mockito.Mockito.*;
 
 /**
  * @author Rob Winch
@@ -66,7 +68,7 @@ public class EnableReactiveMethodSecurityTests {
 		assertThatThrownBy(() -> this.messageService.notPublisherPreAuthorizeFindById(1L))
 			.isInstanceOf(IllegalStateException.class)
 			.extracting(Throwable::getMessage)
-			.isEqualTo("The returnType class java.lang.String on public abstract java.lang.String org.springframework.security.config.annotation.method.configuration.ReactiveMessageService.notPublisherPreAuthorizeFindById(long) must return an instance of org.reactivestreams.Publisher (i.e. Mono / Flux) in order to support Reactor Context");
+			.isEqualTo("The return type java.lang.String on method public abstract java.lang.String org.springframework.security.config.annotation.method.configuration.ReactiveMessageService.notPublisherPreAuthorizeFindById(long) must return an instance of org.reactivestreams.Publisher (i.e. Mono / Flux) in order to support Reactor Context");
 	}
 
 	@Test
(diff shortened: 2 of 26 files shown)


edeandrea:
I pushed a new commit (& rebased previous one) to bring back the tests that were removed. I changed the expression

```java
@PostAuthorize("returnObject?.contains(authentication?.name)")
```

to be

```java
@PostAuthorize("T(org.springframework.security.config.annotation.method.configuration.ReactiveAuthz).containsAuthenticationName(returnObject, authentication?.name)")
```

and added some methods in the new `ReactiveAuthz` class to handle the calls.

config/src/test/java/org/springframework/security/config/annotation/method/configuration/DelegatingReactiveMessageService.java
```diff
@@ -47,7 +49,7 @@ public String notPublisherPreAuthorizeFindById(long id) {
 	}
 
 	@Override
-	@PostAuthorize("returnObject?.contains(authentication?.name)")
```
rwinch: Changing the tests means we are potentially breaking backwards compatibility. We need to leave tests alone and ensure that any changes that we make don't break the existing tests.

config/src/test/java/org/springframework/security/config/annotation/method/configuration/EnableReactiveMethodSecurityTests.java
```diff
@@ -589,8 +591,8 @@ public DelegatingReactiveMessageService defaultMessageService() {
 		}
 
 		@Bean
-		public Authz authz() {
```
rwinch: Again please leave existing tests as they are. Instead create only new tests that demonstrate the new functionality.

core/src/main/java/org/springframework/security/access/expression/method/DefaultReactiveMethodSecurityExpressionHandler.java
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	private <T extends Publisher<?>> T applyFilter(T filterTarget, Expression filterExpression, EvaluationContext ctx) {
+		MethodSecurityExpressionOperations rootObject = (MethodSecurityExpressionOperations) ctx
+				.getRootObject().getValue();
+
+		Class<?> returnType = filterTarget.getClass();
+
+		Function<Object, ? extends Publisher<Boolean>> filter = item -> {
+			rootObject.setFilterObject(item);
```
rwinch: This looks like a race condition because each subscription to the publisher (which can happen on any thread) is going to modify the same rootObject

core/src/main/java/org/springframework/security/access/expression/method/ExpressionBasedAnnotationAttributeFactory.java
```diff
@@ -34,13 +37,24 @@
...
 			MethodSecurityExpressionHandler handler) {
 		this.handler = handler;
 	}
 
+	/**
+	 * Support for reactive method secutiry
+	 * @param handler The {@link ReactiveMethodSecurityExpressionHandler}
+	 * @since 5.1.2
+	 */
+	public ExpressionBasedAnnotationAttributeFactory(
```
rwinch: We cannot add a reactive interface to an implementation that is used in imperative world because we will get NoClassDefFoundErrors

edeandrea:
Thank you @rwinch for the feedback. I'm sure you are busy so your feedback is much appreciated.

I think there is a disconnect between my understanding of how the expressions are supposed to behave vs the current implementation on the reactive-side as well as the current implementation on the non-reactive-side, specifically around the use of `returnObject`. I'll use the example above here

```java
@PostAuthorize("returnObject?.contains(authentication?.name)")
public Flux<String> fluxPostAuthorizeFindById(long id)
```

As things work today before any changes I made, in the above example `returnObject` is actually a reference to _each item in the returned `Flux`_, not the `Flux` itself (same if the return is a `Mono`). My understanding of how method security expressions work is that `returnObject` should be a reference to the `Flux` (or `Mono`) itself.

All the work I did in this PR assumes that `returnObject` is a reference to the `Flux` / `Mono` itself, and not the elements within it.

If we were to use the same example on the blocking side and write the expression

```java
@PostAuthorize("returnObject?.contains(authentication?.name)")
public List<String> listPostAuthorizeFindById(long id)
```

then in that case `returnObject` references the `List` itself, not the elements in the `List` (I built a test in `GlobalMethodSecurityConfigurationTests` which can confirm this - putting a breakpoint in `ExpressionBasedPostInvocationAdvice.after`, the `returnedObject` passed in is indeed the `List` itself).

Based on that, plus the way the [documentation is written](https://docs.spring.io/spring-security/site/docs/current/reference/htmlsingle/#access-control-using-preauthorize-and-postauthorize)

   > Less commonly, you may wish to perform an access-control check after the method has been invoked. This can be achieved using the `@PostAuthorize` annotation. To access the return value from a method, use the built-in name `returnObject` in the expression.

I would say that the `returnObject` that currently is used for each item in the `Flux` / `Mono` is actually a bug because functionally it does not work the same as it does on the blocking side. If `returnObject` is indeed intended to be a reference to the `Mono` / `Flux` itself, then the expression `returnObject?.contains(authentication?.name)` is invalid and will never compile/execute because there is no `contains` method on a `Mono` / `Flux` object.

If the decision is that the behavior of `returnObject` should indeed be the items inside the `Flux` / `Mono` rather than the `Flux` / `Mono` itself, then yes _adapting_ the old interface to the new reactive one becomes much simpler. In that scenario though I think the documentation would need to be updated to reflect that there is an implementation difference in what `returnObject` actually means in the reactive world vs the blocking world.

In my opinion I think functionally it should behave the same in both blocking & reactive, which would mean the change being made as part of this PR would not be functionally backwards-compatible _in some use cases. Ultimately its your call and I'll adapt to whichever way you want to go with it.

rwinch:
If returnObject refers to the publisher type (i.e. `Mono` or `Flux`) how does the user leverage that in their SpEL expressions?

sjohnr:
Hello @edeandrea. This PR has been open quite some time without activity, and is currently in conflict. I know you worked quite hard on what is here, but reviewing comments, it doesn't seem like a conclusion was quite reached and quite a bit of additional work would be required it seems. Given all of the above, I'm going to close this PR in a few days, but wanted to give you a heads up first just in case.

edeandrea:
Thank you @sjohnr you are correct it has been quite some time. I do not have plans to resume working on it.

sjohnr:
Thanks for your quick reply. I will go ahead and close this for now.

Issue #3320, opened 2015-10-07
SEC-3121: Support for bcrypt revision $2b$


spring-issuemaster:
[Peter De Wachter](https://jira.spring.io/secure/ViewProfile.jspa?name=pdewacht) (Migrated from [SEC-3121](https://jira.spring.io/browse/SEC-3121?redirect=false)) said:

In February 2014 a new revision of the bcrypt algorithm was published, which Spring Security does not yet implement. This is starting to cause interoperability problems, as various tools now generate bcrypt hashes that are not supported by Spring Security.

jadekler:
This is kind of interesting. What's the general approach for this? Support both through config + defaults, switch to supporting the new one, etc?

pdewacht:
The reference URL got lost in the migration. This message explains the change:
https://marc.info/?l=openbsd-misc&m=139320023202696

icey502:
You guys are welcome to my version of the bcrypt code, it is essentially the same except I replaced the hard-coded version string usage with something that I could modify to comply with the PHP stack I was up against at the time.  Trivial workaround once I realized what was happening.  I could imagine a "real" solution involving either a configuration setting or some kind of apriori inspection of the revision, possibly both.  To give an idea (there are several areas that I had to touch to make it work, but the approach is the same):

private static final char SALT_REVISION_HACK_PHP = 'y';
...
in gensalt(...):
....
            //rs.append("$2a$");  <-- original
            rs.append("$2"+SALT_REVISION_HACK_PHP+"$");  <-- yucky but working change

Gross, but operational :-)  Alas I am not a Bcrypt expert by any stretch, so alternate validation is probably a good idea.

neuhaus:
The [webpage](http://www.openwall.com/crypt/) for the openwall crypt_blowfish library which also implements bcrypt has some information and pointers regarding $2b$ and $2y$.

semiosis:
This [SO answer](https://stackoverflow.com/a/36225192) suggests that the versions `a`, `b`, `x`, and `y`, are all interoperable and only serve a purpose in PHP or OpenBSD implementations.  Since Spring uses it's own implementation of BCrypt, it could accept any of those versions.

[...]

PR #5992, opened 2018-10-18
fix issue#3320 Support for bcrypt revision

lin199231:
From issue #3320
I find that jBCrypt are already support
https://code.google.com/archive/p/jbcrypt/issues/9
So I try to add bcrypt revision support for BCryptPasswordEncoder, and compatible with old version.
Please merge my request, thanks.

### DIFF
--- a/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCrypt.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCrypt.java
@@ -13,252 +13,363 @@
 // OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 package org.springframework.security.crypto.bcrypt;
 
-import java.io.ByteArrayOutputStream;
 import java.io.UnsupportedEncodingException;
-
+import java.util.Arrays;
 import java.security.SecureRandom;
 
 /**
- * BCrypt implements OpenBSD-style Blowfish password hashing using the scheme described in
- * "A Future-Adaptable Password Scheme" by Niels Provos and David Mazieres.
+ * BCrypt implements OpenBSD-style Blowfish password hashing using
+ * the scheme described in "A Future-Adaptable Password Scheme" by
+ * Niels Provos and David Mazieres.
  * <p>
- * This password hashing system tries to thwart off-line password cracking using a
- * computationally-intensive hashing algorithm, based on Bruce Schneier's Blowfish cipher.
- * The work factor of the algorithm is parameterised, so it can be increased as computers
- * get faster.
+ * This password hashing system tries to thwart off-line password
+ * cracking using a computationally-intensive hashing algorithm,
+ * based on Bruce Schneier's Blowfish cipher. The work factor of
+ * the algorithm is parameterised, so it can be increased as
+ * computers get faster.
  * <p>
- * Usage is really simple. To hash a password for the first time, call the hashpw method
- * with a random salt, like this:
- * <p>
- * <code>
- * String pw_hash = BCrypt.hashpw(plain_password, BCrypt.gensalt()); <br>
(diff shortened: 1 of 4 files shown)


crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.java
```diff
@@ -51,32 +53,77 @@ public BCryptPasswordEncoder(int strength) {
 		this(strength, null);
 	}
 
+	/**
+	 * @param version the version of bcrypt, can be 2a,2b,2y
+	 */
+	public BCryptPasswordEncoder(String version) {
```
rwinch: Let's change version inputs from the user to be a inner nested enum named BCryptVersion so that an invalid version cannot be provided.

crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.java
```diff
@@ -51,32 +53,77 @@ public BCryptPasswordEncoder(int strength) {
...
 		String salt;
-		if (strength > 0) {
-			if (random != null) {
-				salt = BCrypt.gensalt(strength, random);
+		if(version != null) {
+			if (strength > 0) {
+				if (random != null) {
+					salt = BCrypt.gensalt(version, strength, random);
+				} else {
+					System.out.println(version);
```
rwinch: Remove the println statement

crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.java
```diff
@@ -51,32 +53,77 @@ public BCryptPasswordEncoder(int strength) {
...
 		this.strength = strength;
 		this.random = random;
 	}
 
 	public String encode(CharSequence rawPassword) {
 		String salt;
-		if (strength > 0) {
-			if (random != null) {
-				salt = BCrypt.gensalt(strength, random);
+		if(version != null) {
```
rwinch: Rather than having this check and two different code paths, let's make it so that version is final and always non-null with a default of $2y

crypto/src/test/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoderTests.java
```diff
@@ -33,6 +33,14 @@ public void matches() {
 		String result = encoder.encode("password");
```
rwinch: Generally with these tests please split up these tests so that there is a different test for each method

crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCrypt.java
```diff
@@ -13,252 +13,363 @@
 // OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 package org.springframework.security.crypto.bcrypt;
 
-import java.io.ByteArrayOutputStream;
```
rwinch: Please remove unnecessary formatting changes in here so it is easier to follow what has changed.

lin199231:
Hi,

    Thanks for the reply, that's very useful advice. I will change it and
update later.

BR
Lin

Rob Winch <notifications@github.com> 于2018年10月18日周四 下午10:39写道：

lin199231:
I have a question, if use $2y as default version. The old version of Spring Security's default version is $2a. Is this OK to change BCryptPasswordEncoder 's default version from $2a to $2y?

rwinch:
@lin199231 Thanks for pointing that out. Please use `$2a` as the default version.

lin199231:
OK，the default version is $2a now.

PR #5994, opened 2018-10-18
Allow in-memory authorized client services to be constructed with a map

vpavic:
This is the equivalent of #5918 only this time for `InMemoryOAuth2AuthorizedClientService` ~and its reactive counterpart~.

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryOAuth2AuthorizedClientService.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryOAuth2AuthorizedClientService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,7 +20,6 @@ import org.springframework.security.oauth2.client.registration.ClientRegistratio
 import org.springframework.security.oauth2.client.registration.ClientRegistrationRepository;
 import org.springframework.util.Assert;
 
-import java.util.Base64;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
@@ -29,6 +28,7 @@ import java.util.concurrent.ConcurrentHashMap;
  * {@link OAuth2AuthorizedClient Authorized Client(s)} in-memory.
  *
  * @author Joe Grandja
+ * @author Vedran Pavic
  * @since 5.0
  * @see OAuth2AuthorizedClientService
  * @see OAuth2AuthorizedClient
@@ -36,7 +36,7 @@ import java.util.concurrent.ConcurrentHashMap;
  * @see Authentication
  */
 public final class InMemoryOAuth2AuthorizedClientService implements OAuth2AuthorizedClientService {
-	private final Map<String, OAuth2AuthorizedClient> authorizedClients = new ConcurrentHashMap<>();
+	private final Map<OAuth2AuthorizedClientId, OAuth2AuthorizedClient> authorizedClients = new ConcurrentHashMap<>();
 	private final ClientRegistrationRepository clientRegistrationRepository;
 
 	/**
@@ -50,6 +50,7 @@ public final class InMemoryOAuth2AuthorizedClientService implements OAuth2Author
 	}
 
 	@Override
+	@SuppressWarnings("unchecked")
 	public <T extends OAuth2AuthorizedClient> T loadAuthorizedClient(String clientRegistrationId, String principalName) {
 		Assert.hasText(clientRegistrationId, "clientRegistrationId cannot be empty");
 		Assert.hasText(principalName, "principalName cannot be empty");
@@ -57,15 +58,15 @@ public final class InMemoryOAuth2AuthorizedClientService implements OAuth2Author
 		if (registration == null) {
 			return null;
 		}
-		return (T) this.authorizedClients.get(this.getIdentifier(registration, principalName));
+		return (T) this.authorizedClients.get(OAuth2AuthorizedClientId.create(registration, principalName));
 	}
 
 	@Override
 	public void saveAuthorizedClient(OAuth2AuthorizedClient authorizedClient, Authentication principal) {
 		Assert.notNull(authorizedClient, "authorizedClient cannot be null");
 		Assert.notNull(principal, "principal cannot be null");
-		this.authorizedClients.put(this.getIdentifier(
-			authorizedClient.getClientRegistration(), principal.getName()), authorizedClient);
+		this.authorizedClients.put(OAuth2AuthorizedClientId.create(authorizedClient.getClientRegistration(),
+				principal.getName()), authorizedClient);
 	}
 
 	@Override
@@ -74,12 +75,8 @@ public final class InMemoryOAuth2AuthorizedClientService implements OAuth2Author
 		Assert.hasText(principalName, "principalName cannot be empty");
 		ClientRegistration registration = this.clientRegistrationRepository.findByRegistrationId(clientRegistrationId);
 		if (registration != null) {
-			this.authorizedClients.remove(this.getIdentifier(registration, principalName));
+			this.authorizedClients.remove(OAuth2AuthorizedClientId.create(registration, principalName));
 		}
 	}
 
-	private String getIdentifier(ClientRegistration registration, String principalName) {
-		String identifier = "[" + registration.getRegistrationId() + "][" + principalName + "]";
-		return Base64.getEncoder().encodeToString(identifier.getBytes());
-	}
 }
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryOAuth2AuthorizedClientService.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryOAuth2AuthorizedClientService.java
@@ -36,8 +36,8 @@ import java.util.concurrent.ConcurrentHashMap;
  * @see Authentication
  */
 public final class InMemoryOAuth2AuthorizedClientService implements OAuth2AuthorizedClientService {
-	private final Map<OAuth2AuthorizedClientId, OAuth2AuthorizedClient> authorizedClients = new ConcurrentHashMap<>();
 	private final ClientRegistrationRepository clientRegistrationRepository;
+	private Map<OAuth2AuthorizedClientId, OAuth2AuthorizedClient> authorizedClients = new ConcurrentHashMap<>();
 
 	/**
 	 * Constructs an {@code InMemoryOAuth2AuthorizedClientService} using the provided parameters.
@@ -47,6 +47,15 @@ public final class InMemoryOAuth2AuthorizedClientService implements OAuth2Author
 	public InMemoryOAuth2AuthorizedClientService(ClientRegistrationRepository clientRegistrationRepository) {
 		Assert.notNull(clientRegistrationRepository, "clientRegistrationRepository cannot be null");
 		this.clientRegistrationRepository = clientRegistrationRepository;
+	}
+
+	/**
+	 * Sets the map of authorized clients to use.
+	 * @param authorizedClients the map of authorized clients
+	 */
+	public void setAuthorizedClients(Map<OAuth2AuthorizedClientId, OAuth2AuthorizedClient> authorizedClients) {
+		Assert.notNull(authorizedClients, "authorizedClients cannot be null");
+		this.authorizedClients = authorizedClients;
 	}
 
 	@Override
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryReactiveOAuth2AuthorizedClientService.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryReactiveOAuth2AuthorizedClientService.java
@@ -15,7 +15,6 @@
  */
 package org.springframework.security.oauth2.client;
 
-import java.util.Base64;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
@@ -31,6 +30,7 @@ import reactor.core.publisher.Mono;
  * {@link OAuth2AuthorizedClient Authorized Client(s)} in-memory.
  *
  * @author Rob Winch
+ * @author Vedran Pavic
  * @since 5.1
  * @see OAuth2AuthorizedClientService
  * @see OAuth2AuthorizedClient
@@ -38,7 +38,7 @@ import reactor.core.publisher.Mono;
  * @see Authentication
  */
 public final class InMemoryReactiveOAuth2AuthorizedClientService implements ReactiveOAuth2AuthorizedClientService {
-	private final Map<String, OAuth2AuthorizedClient> authorizedClients = new ConcurrentHashMap<>();
+	private final Map<OAuth2AuthorizedClientId, OAuth2AuthorizedClient> authorizedClients = new ConcurrentHashMap<>();;
 	private final ReactiveClientRegistrationRepository clientRegistrationRepository;
 
 	/**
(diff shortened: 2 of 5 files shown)


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryOAuth2AuthorizedClientService.java
```diff
@@ -29,26 +29,40 @@
...
 	private final ClientRegistrationRepository clientRegistrationRepository;
 
 	/**
 	 * Constructs an {@code InMemoryOAuth2AuthorizedClientService} using the provided parameters.
 	 *
+	 * @param authorizedClients the map of authorized clients
 	 * @param clientRegistrationRepository the repository of client registrations
 	 */
-	public InMemoryOAuth2AuthorizedClientService(ClientRegistrationRepository clientRegistrationRepository) {
+	public InMemoryOAuth2AuthorizedClientService(Map<String, OAuth2AuthorizedClient> authorizedClients,
```
jgrandja: If this constructor is called with a non-empty `Map<String, OAuth2AuthorizedClient> authorizedClients` and the `key(s)` are not correctly built as per `InMemoryOAuth2AuthorizedClientService.getIdentifier()` than a subsequent call to `loadAuthorizedClient()` would return `null` which would be in error. How can we ensure the `key(s)` in the provided `Map` are built correctly?

vpavic: Two options come to mind - either validate keys in constructor, or define a dedicated type for key.

jgrandja: Validate keys in constructor would be easy. Can you expand on

> dedicated type for key

Do you mean `Map<SomeNewType, OAuth2AuthorizedClient>` OR an interface that generates the key, for example `KeyGenerator`?

vpavic: I meant the former - having something like `OAuth2AuthorizedClientId` feels like the optimal solution from the perspective of API quality IMO. It encapsulates what in reality we have here - a composite id. At first sight, ensuring backwards compatibility should be doable with `default` methods. I can take a closer look at that as a part of this PR, if you agree.

jgrandja: @rwinch Can you expand on this?

> I'd be in favor of creating a static key class that accepts the ClientRegistration as a constructor argument and provides the implementation of the hashCode and equals methods for the key

Sounds like you are in favour of a key class as @vpavic has suggested, for example `OAuth2AuthorizedClientId`? ...that implements `equals` and `hashCode`

rwinch: Yes.

jgrandja: > provides the implementation of the hashCode and equals methods for the key vs exposing implementation details of how the key is constructed.

This makes sense to me as well - hiding the implementation details on how the key is constructed using `equals` and `hashCode`. @vpavic Would you like to update this PR based on the discussed approach?

vpavic: Thanks, both - I'll try to update the PR some time during this week.

vpavic:
I've updated this PR with a commit that that adds `OAuth2AuthorizedClientId`. I've put that commit as first, and updated the original commit to leverage the newly introduced identifier.

Please take a quick look before I get any further with this. The thing is that the only consumer of `OAuth2AuthorizedClientService` (and its reactive counterpart) is `AuthenticatedPrincipalOAuth2AuthorizedClientRepository` (and its reactive counterpart). I believe the APIs of those should also be updated to use `OAuth2AuthorizedClientId` in order to have a real benefit from the newly introduced identifier.

rwinch:
@vpavic While I'm open to considering updating the APIs to use `OAuth2AuthorizedClientId`, I don't think we should make this a part of this PR as I don't think there is a need for it now. Instead, this should be  limited to the Map based implementations. We can then consider updating the `OAuth2AuthorizedClientService` (and reactive equivalent) separately. 

Thoughts?

vpavic:
I agree it's better to handle that separately. I'm going to update this PR with that in mind, and open a new one to consider API changes once this (and `OAuth2AuthorizedClientId`) is merged.

vpavic:
I've updated the PR.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryReactiveOAuth2AuthorizedClientService.java
```diff
@@ -31,31 +30,49 @@
...
  * @author Rob Winch
+ * @author Vedran Pavic
  * @since 5.1
  * @see OAuth2AuthorizedClientService
  * @see OAuth2AuthorizedClient
  * @see ClientRegistration
  * @see Authentication
  */
 public final class InMemoryReactiveOAuth2AuthorizedClientService implements ReactiveOAuth2AuthorizedClientService {
-	private final Map<String, OAuth2AuthorizedClient> authorizedClients = new ConcurrentHashMap<>();
```
rwinch: Is it really reasonable to do this to the reactive equivalent? Afterall injecting a distributed map would be blocking and thus not work.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryOAuth2AuthorizedClientService.java
```diff
@@ -29,43 +28,59 @@
...
 
 	/**
 	 * Constructs an {@code InMemoryOAuth2AuthorizedClientService} using the provided parameters.
 	 *
+	 * @param authorizedClients the map of authorized clients
 	 * @param clientRegistrationRepository the repository of client registrations
 	 */
-	public InMemoryOAuth2AuthorizedClientService(ClientRegistrationRepository clientRegistrationRepository) {
+	public InMemoryOAuth2AuthorizedClientService(
+			Map<OAuth2AuthorizedClientId, OAuth2AuthorizedClient> authorizedClients,
```
rwinch: Given Map is an optional parameter, I'd prefer it to be exposed via a setter. I understand that means it is mutable object, but this is the typical pattern Spring uses.

jgrandja: I agree with @rwinch that we should expose via setter instead - `Map` should be optional. Also, this object is mutable anyway given `saveAuthorizedClient` and `removeAuthorizedClient`.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryReactiveOAuth2AuthorizedClientService.java
jgrandja: Yes this is an important point.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2AuthorizedClientId.java
```diff
@@ -0,0 +1,78 @@
...
+		OAuth2AuthorizedClientId that = (OAuth2AuthorizedClientId) obj;
+		return Objects.equals(this.clientRegistrationId, that.clientRegistrationId)
+				&& Objects.equals(this.principalName, that.principalName);
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hash(this.clientRegistrationId, this.principalName);
+	}
+
```
jgrandja: It might be useful to implement `toString` as well?

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/OAuth2AuthorizedClientIdTests.java
```diff
@@ -0,0 +1,94 @@
...
+	public void equalsWhenSameRegistrationIdAndPrincipalThenShouldReturnTrue() {
+		OAuth2AuthorizedClientId id1 = OAuth2AuthorizedClientId.create(testClientRegistration("test-client"),
+				"test-principal");
+		OAuth2AuthorizedClientId id2 = OAuth2AuthorizedClientId.create(testClientRegistration("test-client"),
+				"test-principal");
+		assertThat(id1.equals(id2)).isTrue();
+	}
+
+	@Test
+	public void equalsWhenDifferentRegistrationIdAndSamePrincipalThenShouldReturnTrue() {
```
jgrandja: This method name should be changed to equalsWhenDifferentRegistrationIdAndSamePrincipalThenShouldReturn**False**

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/OAuth2AuthorizedClientIdTests.java
```diff
@@ -0,0 +1,94 @@
...
+		OAuth2AuthorizedClientId id1 = OAuth2AuthorizedClientId.create(testClientRegistration("test-client1"),
+				"test-principal");
+		OAuth2AuthorizedClientId id2 = OAuth2AuthorizedClientId.create(testClientRegistration("test-client2"),
+				"test-principal");
+		assertThat(id1.equals(id2)).isFalse();
+	}
+
+	@Test
+	public void equalsWhenSameRegistrationIdAndDifferentPrincipalThenShouldReturnTrue() {
```
jgrandja: This method name should be changed to equalsWhenSameRegistrationIdAndDifferentPrincipalThenShouldReturn**False**

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/OAuth2AuthorizedClientIdTests.java
```diff
@@ -0,0 +1,94 @@
...
+	public void hashCodeWhenSameRegistrationIdAndPrincipalThenShouldReturnSame() {
+		OAuth2AuthorizedClientId id1 = OAuth2AuthorizedClientId.create(testClientRegistration("test-client"),
+				"test-principal");
+		OAuth2AuthorizedClientId id2 = OAuth2AuthorizedClientId.create(testClientRegistration("test-client"),
+				"test-principal");
+		assertThat(id1.hashCode()).isEqualTo(id2.hashCode());
+	}
+
+	@Test
+	public void hashCodeWhenDifferentRegistrationIdAndSamePrincipalThenShouldReturnTrue() {
```
jgrandja: This method name should be changed to hashCodeWhenDifferentRegistrationIdAndSamePrincipalThen**NotEqual**

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/OAuth2AuthorizedClientIdTests.java
```diff
@@ -0,0 +1,94 @@
...
+		OAuth2AuthorizedClientId id1 = OAuth2AuthorizedClientId.create(testClientRegistration("test-client1"),
+				"test-principal");
+		OAuth2AuthorizedClientId id2 = OAuth2AuthorizedClientId.create(testClientRegistration("test-client2"),
+				"test-principal");
+		assertThat(id1.hashCode()).isNotEqualTo(id2.hashCode());
+	}
+
+	@Test
+	public void hashCodeWhenSameRegistrationIdAndDifferentPrincipalThenShouldReturnTrue() {
```
jgrandja: This method name should be changed to hashCodeWhenSameRegistrationIdAndDifferentPrincipalThen**NotEqual**

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2AuthorizedClientId.java
```diff
@@ -0,0 +1,78 @@
...
+	private OAuth2AuthorizedClientId(String clientRegistrationId, String principalName) {
+		Assert.notNull(clientRegistrationId, "clientRegistrationId cannot be null");
+		Assert.notNull(principalName, "principalName cannot be null");
+		this.clientRegistrationId = clientRegistrationId;
+		this.principalName = principalName;
+	}
+
+	/**
+	 * Factory method for creating new {@link OAuth2AuthorizedClientId} using
+	 * {@link ClientRegistration} and {@link Authentication}.
```
jgrandja: `{@link Authentication}` should be -> the principal name

vpavic:
Since RC1 is around the corner, heads-up that reactive bits of #5918 need to get reverted. Or maybe all of it, if this PR doesn't make the cut.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryOAuth2AuthorizedClientService.java
jgrandja: @vpavic Apologies for back-tracking my previous comment but what are your thoughts on removing the setter and adding a 2nd constructor for both `ClientRegistrationRepository` and the `Map`?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2AuthorizedClientId.java
```diff
@@ -0,0 +1,78 @@
...
+	 * @param clientRegistration the client registration
+	 * @param principalName the principal name
+	 * @return the new authorized client id
+	 */
+	public static OAuth2AuthorizedClientId create(ClientRegistration clientRegistration,
+			String principalName) {
+		return new OAuth2AuthorizedClientId(clientRegistration.getRegistrationId(),
```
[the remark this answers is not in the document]
jgrandja: I think the parameter `ClientRegistration clientRegistration` should be changed to `String clientRegistrationId`, which aligns with `OAuth2AuthorizedClientService.loadAuthorizedClient(String clientRegistrationId, String principalName)`.

I'm also curious on why this factory method is needed? Alternatively, we can make the constructor `public`. What are your throughts?

jgrandja:
Thanks for the heads up @vpavic. I left a couple of comments. Also, would you be interested in submitting a PR that reverts the reactive bits of #5918?

jgrandja:
@vpavic Assuming we merge this PR and revert the reactive bits in #5918, this will allow a `Map` to be set by the caller for both `InMemoryClientRegistrationRepository` and `InMemoryOAuth2AuthorizedClientService`. However, the reactive counterparts `InMemoryReactiveClientRegistrationRepository` and `InMemoryReactiveOAuth2AuthorizedClientService` **will not allow** a `Map` to be set by the caller since this could introduce a blocking operation as @rwinch [mentioned](https://github.com/spring-projects/spring-security/pull/5994#discussion_r233919098).

What are your thoughts on this? This will ultimately introduce inconsistent implementations between the two which I'm not sure at this point is the right way to go. Given that you raised this issue to allow for a distributed `Map` to be set, you will not be able to achieve this for a Reactive implementation/application. Do you have any concerns with this?

vpavic:
I'm afraid I won't have time to revisit this and the #5918 partial revert as I'm quite occupied by stuff for next Spring Session release so feel free to do what you feel is the best.

Generally I'm interested primarily in Servlet side of the things (and have included reactive bits for completeness sake) so that inconsistency is fine with me. I also don't think inconsistency should be the reason for not providing something that's supporting a perfectly valid (and IMO needed) use case for Servlet based implementation.

Also I'll express a concern that I've already express in other places, which I believe is important and is the motivation for this and other related PRs - all the store-like interfaces in new Spring Security OAuth 2.0 are only supported by in-memory implementations, and that's IMO something that you typically don't want to be running in production.

jgrandja:
@vpavic No worries if you don't have the time...totally understand. I'll take care of the remaining work.

> I'm interested primarily in Servlet side of the things

Ok I don't see any issues on merging the changes for Servlet.

> ...only supported by in-memory implementations, and that's IMO something that you typically don't want to be running in production

Totally agreed and yes the in-memory implementations are meant for development only. It's up to the user to define `@Bean` of either `ClientRegistrationRepository` and/or `OAuth2AuthorizedClientService` for a production setup. We very well may provide other implementations in a later release but we've had other priorities to focus on as of late.

Thanks again for the reminder on this ticket and #5918...it almost slipped through the cracks.

jgrandja:
@vpavic I just merged this along with a polish commit 23d61d4. I also did a partial revert of #5918 in this commit e554547.
Let me know if you see any issues with the updates.
Thanks.

PR #5999, opened 2018-10-19
Remove unnecessary concatenation of sql in JdbcUserDetailsManager

drumonii:


rwinch:
Thanks for the PR @drumonii! This is merged into master

PR #6026, opened 2018-10-28
#3743 adds PostFilter support for streams

### LINKED ISSUE #3743: Add support for filtering Java 8 Streams with @PostFilter
It would be very useful to be able to `@PostFilter` methods that return a `java.util.stream.Stream` object. It should be relatively simple to add something like this to [DefaultMethodSecurityExpressionHandler](../blob/master/core/src/main/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandler.java)`.filter()`

```
if (filterTarget instanceof Stream) {
    Stream<?> filtered = (Stream<?>) filterTarget;
    filtered = filtered.filter(filterObject -> {
        rootObject.setFilterObject(filterObject);
        return ExpressionUtils.evaluateAsBoolean(filterExpression, ctx);
    });
    return filtered;
}
```

kagof:
Resolves #3743 - adding support for `Stream`s in `@PostFilter` expressions by adding to `DefaultMethodSecurityExpressionHandler::filter`.

### DIFF
--- a/core/src/main/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandler.java
+++ b/core/src/main/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandler.java
@@ -175,21 +175,12 @@ public class DefaultMethodSecurityExpressionHandler extends
 
 		if (filterTarget instanceof Stream) {
 			final Stream<?> original = (Stream<?>) filterTarget;
-			if (debug) {
-				logger.debug("Filtering stream with " + original.count() + " elements");
-			}
 
-			Stream<?> filtered = original.filter(filterObject -> {
+			return original.filter(filterObject -> {
 				rootObject.setFilterObject(filterObject);
 				return ExpressionUtils.evaluateAsBoolean(filterExpression, ctx);
 			})
 					.onClose(original::close);
-
-			if (debug) {
-				logger.debug("Retaining elements: " + filtered.collect(Collectors.toList()));
-			}
-
-			return filtered;
 		}
 
 		throw new IllegalArgumentException(
--- a/core/src/main/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandler.java
+++ b/core/src/main/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandler.java
```diff
@@ -172,8 +173,27 @@ public Object filter(Object filterTarget, Expression filterExpression,
 			return filtered;
 		}
 
+		if (filterTarget instanceof Stream) {
+			final Stream<?> original = (Stream<?>) filterTarget;
+			if (debug) {
+				logger.debug("Filtering stream with " + original.count() + " elements");
```
jzheaux: While I appreciate the consistency that this affords to the other two data types this class supports, this won't work since you cannot operate on a stream after it's been terminated.

e.g. after running line 179, then line 182 would fail with `stream has already been operated on or closed`.

This would be the same with line 189 and whatever the user wanted to do downstream.

That being the case, let's remove these debug if blocks.

core/src/test/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandlerTests.java
```diff
@@ -68,4 +72,22 @@ public void createEvaluationContextCustomTrustResolver() {
...
+		final Stream<String> stream = Stream.of("1", "2", "3");
+
+		Expression expression = handler.getExpressionParser().parseExpression("filterObject ne '2'");
+
+		EvaluationContext context = handler.createEvaluationContext(authentication,
+				methodInvocation);
+
+		Object filtered = handler.filter(stream, expression, context);
+
+		Assert.assertTrue("response was wrong type", Stream.class.isAssignableFrom(filtered.getClass()));
```
jzheaux: It would be nice if these used a static import for `Assert`.

core/src/test/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandlerTests.java
```diff
@@ -68,4 +72,22 @@ public void createEvaluationContextCustomTrustResolver() {
 
 		verify(trustResolver).isAnonymous(authentication);
 	}
+
+	@Test
+	@SuppressWarnings("unchecked")
+	public void testFilteringStream() {
```
jzheaux: Do you have time to add just one more test? Because @nightswimmings brought up the fact that we need to preserve the closeability of the original stream, it would be good to prove that that works in a unit test. Something like:

```suggestion
	@Test
	public void filterWhenClosedThenUpstreamGetsClosed() {
		Stream<String> upstream = mock(Stream.class);

		// use expression handler to get filtered stream

		filtered.close();
		verify(upstream).close();
	}
```

kagof:
Hi @jzheaux, thanks for the feedback! I've made the changes you requested in a new commit.

jzheaux:
Thanks, again @kagof for the PR!

This has been merged into master via 92e68a589acdf4fb1390a8fb7452cad5b3237349 and 50d26c9d28e04edf9ec7434ca9002b18bebaa00a.

Also, I did a quick polish on a method name: 7d3302f52b876357bcb063f92daf58849cfc6822

Issue #5499, opened 2018-07-11
Security-related HTTP headers not written if response is committed during INCLUDE dispatch


chschu:
### Summary

```HeaderWriterFilter``` does not write HTTP headers if the response is committed inside a ```DispatcherType.INCLUDE``` request dispatch.

This means that critical headers might be missing in the response, introducing a potential security risk.

### Actual Behavior

According to the servlet specification, response HTTP headers cannot be set during an INCLUDE dispatch, e.g. in an included JSP:

> Section 9.3 - The Include method
[...] It cannot set headers or call any method that affects the headers of the response, with the exception of the HttpServletRequest.getSession() and HttpServletRequest.getSession(boolean) methods. Any attempt to set the headers must be ignored, [...]

So if the response is committed during a include that produced a lot of output to the response body, the ```HeaderWriterFilter``` calls the response methods that would normally set the HTTP headers, but those calls are ignored by the servlet container.

At least Tomcat 8.5.x adheres to the servlet specification an does not write the headers to the response.

### Expected Behavior

The headers are always written to the response, even if the response is committed during an include.

### Configuration

- Tomcat 8.5.28
- Spring-based web application with a JSP that includes other JSP producing more than 8 KB of output (default buffer size)

### Version

5.0.6.RELEASE

### Sample

[spring-security-gh5499.zip](https://github.com/spring-projects/spring-security/files/2184683/spring-security-gh5499.zip)

Attached is a Zip file containing a simple Spring Boot application. Simply run the `de.chschu.spring.security.gh5499.Application` main class.

The response of http://localhost:8080/positive will have the headers, while http://localhost:8080/negative (which simply includes the other one) will not.

haasc:
I'm have the same problem, but I've diagnosed it a little differently.  

In my case, by the time the header writing is attempted during the processing of the include file, the HttpResponse is wrapped with org.apache.catalina.core.ApplicationHttpResponse with the "included" attribute set to true.  With the included field set to true, all of the headers that get emitted are suppressed.  However, as expected, the response is committed.

So, when the processing gets back to the header filter, the response has been committed, so again headers are suppressed.

The net result is that no headers get emitted.

rwinch:
I think we can resolve this using an OnComittedRequestWrapper which triggers the headers to be written before an attempt to include content happens.

PR #6030, opened 2018-10-30
Write Security Headers Before Servlet Include

jzheaux:
HeaderWriterFilter wraps request dispatcher so it can write security
headers before the include occurs.

Fixes: gh-5499


rwinch:
Thanks for the PR @jzheaux! Can you backport this to 5.1.x and 5.0.x please? Also a PR to Spring Session's on commit hook would be good https://github.com/spring-projects/spring-session/blob/2.1.1.RELEASE/spring-session-core/src/main/java/org/springframework/session/web/http/OnCommittedResponseWrapper.java

PR #4136, opened 2016-11-21
Changes related to issue #3912

shazin:
Fixes #3912

PR #6035, opened 2018-11-01
#3912 lazyBean method respects @Primary annotation

### LINKED ISSUE #3912: AuthenticationConfiguration.lazyBean(AuthenticationManager.class) should honor @Primary if multiple candidates are available
Using Spring Boot 1.3.4, Spring Security 4.0.4:

I encounter the following Exception for some (MockMvc) Tests that I want to run:

```
Caused by: java.lang.IllegalArgumentException: Expecting to only find a single bean for type interface org.springframework.security.authentication.AuthenticationManager, but found [docAuthenticationManager, webAuthenticationManager, globalAuthenticationManager]
    at org.springframework.util.Assert.isTrue(Assert.java:68)
    at org.springframework.security.config.annotation.authentication.configuration.AuthenticationConfiguration.lazyBean(AuthenticationConfiguration.java:116)
    at org.springframework.security.config.annotation.authentication.configuration.AuthenticationConfiguration.getAuthenticationMangerBean(AuthenticationConfiguration.java:128)
    at org.springframework.security.config.annotation.authentication.configuration.AuthenticationConfiguration.getAuthenticationManager(AuthenticationConfiguration.java:84)
    at org.springframework.security.config.annotation.method.configuration.GlobalMethodSecurityConfiguration.authenticationManager(GlobalMethodSecurityConfiguration.java:257)
    at org.springframework.security.config.annotation.method.configuration.GlobalMethodSecurityConfiguration.methodSecurityInterceptor(GlobalMethodSecurityConfiguration.java:123)
    at ch.belimo.cloud.server.services.security.SecurityConfiguration$MethodSecurityConfiguration$$EnhancerBySpringCGLIB$$8864b9ed.CGLIB$methodSecurityInterceptor$8(<generated>)
    at ch.belimo.cloud.server.services.security.SecurityConfiguration$MethodSecurityConfiguration$$EnhancerBySpringCGLIB$$8864b9ed$$FastClassBySpringCGLIB$$b67a6873.invoke(<generated>)
    at org.springframework.cglib.proxy.MethodProxy.invokeSuper(MethodProxy.java:228)
    at org.springframework.context.annotation.ConfigurationClassEnhancer$BeanMethodInterceptor.intercept(ConfigurationClassEnhancer.java:356)
    at com.mystuff.SecurityConfiguration$MethodSecurityConfiguration$$EnhancerBySpringCGLIB$$8864b9ed.methodSecurityInterceptor(<generated
...
```

I tried to fix the issue by defining an `@Primary` annotated AuthenticationManager instance for my Test, but this has no effect. The reason being that the `AuthenticationConfiguration.lazyBean()` Method does not honor `@Primary`, if multiple candidates are found:

```
String[] beanNamesForType = BeanFactoryUtils.beanNamesForTypeIncludingAncestors(
            applicationContext, interfaceName);
    if (beanNamesForType.length == 0) {
        return null;
    }
    Assert.isTrue(beanNamesForType.length == 1,
            "Expecting to only find a single bean for type " + interfaceName
                    + ", but found " + Arrays.asList(beanNamesForType));
```

kagof:
Resolves #3912 . 

I know that @shazin created #4136 as well, however it is now nearly two years old and full of conflicts.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/AuthenticationConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/AuthenticationConfiguration.java
@@ -15,21 +15,15 @@
  */
 package org.springframework.security.config.annotation.authentication.configuration;
 
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.concurrent.atomic.AtomicBoolean;
-
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-
 import org.springframework.aop.framework.ProxyFactoryBean;
 import org.springframework.aop.target.LazyInitTargetSource;
 import org.springframework.beans.factory.BeanFactoryUtils;
 import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.ApplicationContext;
+import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
@@ -48,6 +42,13 @@ import org.springframework.security.core.userdetails.UserDetailsService;
 import org.springframework.security.crypto.factory.PasswordEncoderFactories;
 import org.springframework.security.crypto.password.PasswordEncoder;
 import org.springframework.util.Assert;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
 
 /**
  * Exports the authentication {@link Configuration}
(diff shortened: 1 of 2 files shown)


config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/AuthenticationConfiguration.java
```diff
@@ -151,10 +152,23 @@ public void setObjectPostProcessor(ObjectPostProcessor<Object> objectPostProcess
 		if (beanNamesForType.length == 0) {
 			return null;
 		}
-		Assert.isTrue(beanNamesForType.length == 1,
-				() -> "Expecting to only find a single bean for type " + interfaceName
-						+ ", but found " + Arrays.asList(beanNamesForType));
-		lazyTargetSource.setTargetBeanName(beanNamesForType[0]);
+		String beanName;
+		if (beanNamesForType.length > 1) {
+			List<String> primaryBeanNames = Arrays.stream(beanNamesForType)
```
rwinch: This should be able to be simplified to something like:

```java
applicationContext.getAutowireCapableBeanFactory().resolveNamedBean(Foo.class);
```

kagof: Hi @rwinch, would that replace whole `BeanFactoryUtils.beanNamesForTypeIncludingAncestors(applicationContext, interfaceName);` part? If I understand correctly, since the implementation of `resolveNamedBean` handles primary, this method could just become something along the lines of

```java
LazyInitTargetSource lazyTargetSource = new LazyInitTargetSource();
String beanName;
try {
	beanName = applicationContext.getAutowireCapableBeanFactory().resolveNamedBean(interfaceName).getBeanName();
} catch (NoSuchBeanDefinitionException ex) {
	return null;
}
lazyTargetSource.setTargetBeanName(beanName);
lazyTargetSource.setBeanFactory(applicationContext);
ProxyFactoryBean proxyFactory = new ProxyFactoryBean();
proxyFactory = objectPostProcessor.postProcess(proxyFactory);
proxyFactory.setTargetSource(lazyTargetSource);
return (T) proxyFactory.getObject();
```

However, is this method still lazy if we do that? It looks to me like `resolveNamedBean` will init an instance of the bean.

config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/AuthenticationConfiguration.java
```diff
@@ -151,10 +152,23 @@ public void setObjectPostProcessor(ObjectPostProcessor<Object> objectPostProcess
...
+				.filter(i -> applicationContext instanceof ConfigurableApplicationContext)
+				.filter(n -> ((ConfigurableApplicationContext) applicationContext).getBeanFactory().getBeanDefinition(n).isPrimary())
+				.collect(Collectors.toList());
+
+			Assert.isTrue(primaryBeanNames.size() != 0, () -> "Found " + beanNamesForType.length
+					+ " beans for type " + interfaceName + ", but none marked as primary");
+			Assert.isTrue(primaryBeanNames.size() == 1, () -> "Found " + primaryBeanNames.size()
+					+ " beans for type " + interfaceName + " marked as primary");
+			beanName = primaryBeanNames.get(0);
+		} else {
```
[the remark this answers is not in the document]
kagof: I believe that is checked at line 152 - if there are 0 bean names, we short circuit and return `null`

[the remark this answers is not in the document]
rwinch: Thanks you are correct.

config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/AuthenticationConfiguration.java
rwinch: Thanks for pointing that out. I missed that it did create the bean too.

PR #6038, opened 2018-11-02
Fix csrf:token-repository-ref XSD documentation

### LINKED ISSUE #6037: XSD documentation for csrf element's token-repository-ref attribute is misleading
### Summary

The XML schema definition has an minor issue when it comes to describing the default for the `token-repository-ref` attribute of the `<csrf>` element. This has been the case for many versions now.

### Actual

The documentation reads: 

> The CsrfTokenRepository to use. The default is HttpSessionCsrfTokenRepository

### Expected

It should say: 

> The CsrfTokenRepository to use. The default is HttpSessionCsrfTokenRepository wrapped by LazyCsrfTokenRepository.

You can see the `HttpSessionCsrfTokenRepository` is wrapped here: https://github.com/spring-projects/spring-security/blob/8a475e39be67d34b463b62322959bef71645a126/config/src/main/java/org/springframework/security/config/http/CsrfBeanDefinitionParser.java#L87

The documentation is misleading because overriding the attribute with a reference to a plain `HttpSessionCsrfTokenRepository` will lead to eager token generation, while leaving it default would lead to lazy generation. 

_One scenario in which eager CSRF token generation might be necessary is when Thymeleaf 3 commits a part of the response before accessing the CSRF token. See https://github.com/spring-projects/spring-security/issues/3906#issuecomment-222729442._

### Version

4.x, 5.x

----

I'd be happy to submit a PR, if I'd know where I'd need to fix this and for which versions. I think the xsd is generated from this file for version 5.1:

https://github.com/spring-projects/spring-security/blob/8a475e39be67d34b463b62322959bef71645a126/config/src/main/resources/org/springframework/security/config/spring-security-5.1.rnc#L741

evpaassen:
The documentation of the `token-repository-ref attribute` of the `csrf``  element in the schema has been updated to make clear the default repository is lazy.

Fixes #6037.

I wasn't sure which versions should be targeted, so I chose 4.2, 5.0 and 5.1. If this is incorrect, I'll update this PR.

### DIFF
--- a/config/src/main/resources/org/springframework/security/config/spring-security-4.2.rnc
+++ b/config/src/main/resources/org/springframework/security/config/spring-security-4.2.rnc
@@ -748,7 +748,7 @@ csrf-options.attlist &=
 	## The RequestMatcher instance to be used to determine if CSRF should be applied. Default is any HTTP method except "GET", "TRACE", "HEAD", "OPTIONS"
 	attribute request-matcher-ref { xsd:token }?
 csrf-options.attlist &=
-	## The CsrfTokenRepository to use. The default is HttpSessionCsrfTokenRepository
+	## The CsrfTokenRepository to use. The default is HttpSessionCsrfTokenRepository wrapped by LazyCsrfTokenRepository.
 	attribute token-repository-ref { xsd:token }?
 
 headers =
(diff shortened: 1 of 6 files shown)


rwinch:
Thanks for the PR @evpaassen! This is now merged into 4.2.x, 5.0.x, 5.1.x, and master

PR #6039, opened 2018-11-02
InMemoryUserDetailsManager.updatePassword case-insenstive

dperezcabrera:
When username is not in lowercase It throws a null pointer exception.

### DIFF
--- a/core/src/main/java/org/springframework/security/provisioning/InMemoryUserDetailsManager.java
+++ b/core/src/main/java/org/springframework/security/provisioning/InMemoryUserDetailsManager.java
@@ -143,7 +143,7 @@ public class InMemoryUserDetailsManager implements UserDetailsManager,
 	@Override
 	public UserDetails updatePassword(UserDetails user, String newPassword) {
 		String username = user.getUsername();
-		MutableUserDetails mutableUser = this.users.get(username);
+		MutableUserDetails mutableUser = this.users.get(username.toLowerCase());
 		mutableUser.setPassword(newPassword);
 		return mutableUser;
 	}
--- a/core/src/test/java/org/springframework/security/provisioning/InMemoryUserDetailsManagerTests.java
+++ b/core/src/test/java/org/springframework/security/provisioning/InMemoryUserDetailsManagerTests.java
@@ -18,6 +18,7 @@ package org.springframework.security.provisioning;
 
 import org.junit.Test;
 import org.springframework.security.core.userdetails.PasswordEncodedUser;
+import org.springframework.security.core.userdetails.User;
 import org.springframework.security.core.userdetails.UserDetails;
 
 import static org.assertj.core.api.Assertions.*;
@@ -36,5 +37,16 @@ public class InMemoryUserDetailsManagerTests {
 		String newPassword = "newPassword";
 		this.manager.updatePassword(this.user, newPassword);
 		assertThat(this.manager.loadUserByUsername(this.user.getUsername()).getPassword()).isEqualTo(newPassword);
+	}
+
+	@Test
+	public void changePasswordWhenUsernameIsNotInLowercase() {
+		UserDetails userNotLowerCase = User.withUserDetails(PasswordEncodedUser.user())
+				.username("User")
+				.build();
+
+		String newPassword = "newPassword";
+		this.manager.updatePassword(userNotLowerCase, newPassword);
+		assertThat(this.manager.loadUserByUsername(userNotLowerCase.getUsername()).getPassword()).isEqualTo(newPassword);
 	}
 }


pivotal-issuemaster:
@dperezcabrera Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=6039)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=6039) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@dperezcabrera Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=6039)!

core/src/main/java/org/springframework/security/provisioning/InMemoryUserDetailsManager.java
```diff
@@ -143,7 +143,7 @@ public void changePassword(String oldPassword, String newPassword) {
 	@Override
 	public UserDetails updatePassword(UserDetails user, String newPassword) {
 		String username = user.getUsername();
-		MutableUserDetails mutableUser = this.users.get(username);
+		MutableUserDetails mutableUser = this.users.get(username.toLowerCase());
```
gburboz: Is `username` expected to be case insensitive?

gburboz: If `username` is itself `null`, `username.toLowerCase()` will cause `NullPointerException`

dperezcabrera: Yes, It is. You can see the full context: [https://github.com/spring-projects/spring-security/blob/master/core/src/main/java/org/springframework/security/provisioning/InMemoryUserDetailsManager.java
](link)

And this is an example of case insensitive:

```
        ...
	public void createUser(UserDetails user) {
		Assert.isTrue(!userExists(user.getUsername()), "user should not exist");

		users.put(user.getUsername().toLowerCase(), new MutableUser(user));
	}

	public void deleteUser(String username) {
		users.remove(username.toLowerCase());
	}

	public void updateUser(UserDetails user) {
		Assert.isTrue(userExists(user.getUsername()), "user should exist");

		users.put(user.getUsername().toLowerCase(), new MutableUser(user));
	}

	public boolean userExists(String username) {
		return users.containsKey(username.toLowerCase());
	}
       ...
```

dperezcabrera: `username` can't be `null`.

gburboz: Agreed.

core/src/main/java/org/springframework/security/provisioning/InMemoryUserDetailsManager.java
```diff
@@ -143,7 +143,7 @@ public void changePassword(String oldPassword, String newPassword) {
 	@Override
 	public UserDetails updatePassword(UserDetails user, String newPassword) {
 		String username = user.getUsername();
```
gburboz: This is missing check `Assert.isTrue(userExists(user.getUsername()), "user should exist");`

core/src/main/java/org/springframework/security/provisioning/InMemoryUserDetailsManager.java
gburboz: Considering overall theme of this class, `null` check is not required. I guess when `null` is passed, `NullPointerException` is expected outcome.

rwinch:
Thanks for the PR @dperezcabrera! Could you please add a test for this?

dperezcabrera:
I would like to add this change to milestone 5.1.2 if possible. Should I change the origin and destination branch to 5.1.x?

core/src/test/java/org/springframework/security/core/userdetails/PasswordEncodedUser.java
```diff
@@ -23,7 +23,7 @@
  * @since 5.0
  */
 public class PasswordEncodedUser {
-	private static final UserDetails USER = withUsername("user").password("password").roles("USER").build();
+	private static final UserDetails USER = withUsername("User").password("password").roles("USER").build();
```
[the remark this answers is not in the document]
dperezcabrera: I understand what you are trying to say, however this does not change other test assumptions because this variable is only used as a variable returned from mocks.
IMHO It's not a good idea to increase the unit test batteries for trivial cases like this one (It doesn't event change the cyclomatic complexity) since they will have to be maintained without adding more value to the code. I thought that even if now I don't impact the tests globally, it's interesting to make the change here to keep in mind that usernames may not be in lower case always.
Perhaps I've made too many assumptions and didn't explain them. 
Anyway, I have no objection to following your suggestion. Thank you very much.

core/src/test/java/org/springframework/security/core/userdetails/PasswordEncodedUser.java
```diff
@@ -25,12 +25,18 @@
 public class PasswordEncodedUser {
 	private static final UserDetails USER = withUsername("user").password("password").roles("USER").build();
 
+        private static final UserDetails USER_NOT_LOWERCASE = withUsername("User").password("password").roles("USER").build();
```
rwinch: Sorry. I think you misunderstood what I was asking. Rather than creating a new public user in here. Just create one in the test itself.

core/src/test/java/org/springframework/security/provisioning/InMemoryUserDetailsManagerTests.java
```diff
@@ -37,4 +38,11 @@ public void changePassword() {
 		this.manager.updatePassword(this.user, newPassword);
 		assertThat(this.manager.loadUserByUsername(this.user.getUsername()).getPassword()).isEqualTo(newPassword);
 	}
+        
+        @Test
+	public void changePasswordWhenUsernameIsNotInLowercase() {
```
rwinch: Instead of any changes in PasswordEncodedUser, use something like this to create the user in the test method (not a member variable because it is only used once):

```java
UserDetails userNotLowerCase = User.withUserDetails(PasswordEncodedUser.user())
    .username("User")
    .build();
```

core/src/test/java/org/springframework/security/core/userdetails/PasswordEncodedUser.java
rwinch: This will be removed, but in general please pay attention to whites pace. This is incorrectly indented with spaces rather than tabs which means the checkstyle tests will fail. Please use tabs.

dperezcabrera: Ok, Thanks.

core/src/test/java/org/springframework/security/provisioning/InMemoryUserDetailsManagerTests.java
dperezcabrera: Ok, It's done, thank you.

rwinch:
Thanks for the PR! This is now merged into master and backported to 5.1.x

Issue #5508, opened 2018-07-13
Spring Security ACL: No operator matches the given name and argument type


vjykumar:
When trying to use Spring Security ACL, I am facing error: org.postgresql.util.PSQLException: ERROR: operator does not exist: bigint = character varying

Below is the error stack:

> org.springframework.jdbc.BadSqlGrammarException: PreparedStatementCallback; bad SQL grammar [select acl_object_identity.object_id_identity, acl_entry.ace_order,  acl_object_identity.id as acl_id, acl_object_identity.parent_object, acl_object_identity.entries_inheriting, acl_entry.id as ace_id, acl_entry.mask,  acl_entry.granting,  acl_entry.audit_success, acl_entry.audit_failure,  acl_sid.principal as ace_principal, acl_sid.sid as ace_sid,  acli_sid.principal as acl_principal, acli_sid.sid as acl_sid, acl_class.class from acl_object_identity left join acl_sid acli_sid on acli_sid.id = acl_object_identity.owner_sid left join acl_class on acl_class.id = acl_object_identity.object_id_class   left join acl_entry on acl_object_identity.id = acl_entry.acl_object_identity left join acl_sid on acl_entry.sid = acl_sid.id  where ( (acl_object_identity.object_id_identity = ? and acl_class.class = ?)) order by acl_object_identity.object_id_identity asc, acl_entry.ace_order asc]; nested exception is org.postgresql.util.PSQLException: ERROR: operator does not exist: bigint = character varying
>   Hint: No operator matches the given name and argument type(s). You might need to add explicit type casts.
>   Position: 781
> 	at org.springframework.jdbc.support.SQLStateSQLExceptionTranslator.doTranslate(SQLStateSQLExceptionTranslator.java:101)
> 	at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:72)
> 	at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:81)
> 	at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:81)
> 	at org.springframework.jdbc.core.JdbcTemplate.translateException(JdbcTemplate.java:1402)
> 	at org.springframework.jdbc.core.JdbcTemplate.execute(JdbcTemplate.java:620)
[...]

PR #6050, opened 2018-11-06
Fix issue with PostgreSQL: org.postgresql.util.PSQLException

nenaraab:
When trying to use Spring Security ACL, I am facing the error: org.postgresql.util.PSQLException: ERROR: operator does not exist: bigint = character varying.

As recommended in the [Spring.io documentation](https://docs.spring.io/spring-security/site/docs/current/reference/html5/#postgresql) i've setup the acl tables in my PostgreSQL database, with column `acl_object_identity.object_id_identity` of type `varchar(36)`.

IMHO the `object_id_identity` in the `jdbcTemplate` method calls MUST always be passed as a character sequence.

This relates to issue:
#5508

### DIFF
--- a/acl/src/main/java/org/springframework/security/acls/jdbc/JdbcAclService.java
+++ b/acl/src/main/java/org/springframework/security/acls/jdbc/JdbcAclService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2004, 2005, 2006, 2017 Acegi Technology Pty Limited
+ * Copyright 2004, 2005, 2006, 2017, 2018 Acegi Technology Pty Limited
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -27,6 +27,7 @@ import javax.sql.DataSource;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.springframework.core.convert.ConversionService;
+import org.springframework.jdbc.core.JdbcOperations;
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.jdbc.core.RowMapper;
 import org.springframework.security.acls.domain.ObjectIdentityImpl;
@@ -67,7 +68,7 @@ public class JdbcAclService implements AclService {
 	// ~ Instance fields
 	// ================================================================================================
 
-	protected final JdbcTemplate jdbcTemplate;
+	protected final JdbcOperations jdbcOperations;
 	private final LookupStrategy lookupStrategy;
 	private boolean aclClassIdSupported;
 	private String findChildrenSql = DEFAULT_SELECT_ACL_WITH_PARENT_SQL;
@@ -77,9 +78,13 @@ public class JdbcAclService implements AclService {
 	// ===================================================================================================
 
 	public JdbcAclService(DataSource dataSource, LookupStrategy lookupStrategy) {
-		Assert.notNull(dataSource, "DataSource required");
+		this(new JdbcTemplate(dataSource), lookupStrategy);
+	}
+
+	public JdbcAclService(JdbcOperations jdbcOperations, LookupStrategy lookupStrategy) {
+		Assert.notNull(jdbcOperations, "JdbcOperations required");
 		Assert.notNull(lookupStrategy, "LookupStrategy required");
-		this.jdbcTemplate = new JdbcTemplate(dataSource);
+		this.jdbcOperations = jdbcOperations;
 		this.lookupStrategy = lookupStrategy;
 		this.aclClassIdUtils = new AclClassIdUtils();
 	}
@@ -88,15 +93,14 @@ public class JdbcAclService implements AclService {
 	// ========================================================================================================
 
 	public List<ObjectIdentity> findChildren(ObjectIdentity parentIdentity) {
-		Object[] args = { parentIdentity.getIdentifier(), parentIdentity.getType() };
-		List<ObjectIdentity> objects = jdbcTemplate.query(findChildrenSql, args,
+		Object[] args = { parentIdentity.getIdentifier().toString(), parentIdentity.getType() };
+		List<ObjectIdentity> objects = jdbcOperations.query(findChildrenSql, args,
 				new RowMapper<ObjectIdentity>() {
 					public ObjectIdentity mapRow(ResultSet rs, int rowNum)
 							throws SQLException {
 						String javaType = rs.getString("class");
 						Serializable identifier = (Serializable) rs.getObject("obj_id");
 						identifier = aclClassIdUtils.identifierFrom(identifier, rs);
-
 						return new ObjectIdentityImpl(javaType, identifier);
 					}
 				});
--- a/acl/src/test/java/org/springframework/security/acls/jdbc/JdbcAclServiceTests.java
+++ b/acl/src/test/java/org/springframework/security/acls/jdbc/JdbcAclServiceTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


pivotal-issuemaster:
@nenaraab Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=6050)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=6050) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@nenaraab Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=6050)!

acl/src/main/java/org/springframework/security/acls/jdbc/JdbcAclService.java
```diff
@@ -88,7 +88,7 @@ public JdbcAclService(DataSource dataSource, LookupStrategy lookupStrategy) {
 	// ========================================================================================================
 
 	public List<ObjectIdentity> findChildren(ObjectIdentity parentIdentity) {
-		Object[] args = { parentIdentity.getIdentifier(), parentIdentity.getType() };
+		Object[] args = { "" + parentIdentity.getIdentifier(), parentIdentity.getType() };
```
mangei: What about calling `toString()` instead of doing string concatination? Or may the objects be `null`?

```suggestion
		Object[] args = { parentIdentity.getIdentifier().toString(), parentIdentity.getType() };
```

rwinch:
Thank you for the PR @nenaraab! 

1) Could you please rebase off master?
2) Could you please use `String.valueOf` as suggested by @mangei?

If you need any help with the rebase, please let me know and I'd be happy to help.

acl/src/main/java/org/springframework/security/acls/jdbc/JdbcAclService.java
nenaraab: @mangei Thanks! 
No the identifier could never be null, I've adapted the code accordingly!

nenaraab:
@rwinch 
Thanks for cooperation! I've updated the PR. This will fix all my current PostgreSQL issues...

acl/src/main/java/org/springframework/security/acls/jdbc/JdbcAclService.java
```diff
@@ -88,7 +88,7 @@ public JdbcAclService(DataSource dataSource, LookupStrategy lookupStrategy) {
 	// ========================================================================================================
 
 	public List<ObjectIdentity> findChildren(ObjectIdentity parentIdentity) {
-		Object[] args = { parentIdentity.getIdentifier(), parentIdentity.getType() };
+		Object[] args = { parentIdentity.getIdentifier().toString(), parentIdentity.getType() };
```
rwinch: Can you please account for the fact that this `getIdentifier()` might be `null`? Initially I thought this would be best with `String.valueOf`, but that converts `null` to `"null"` which is not ideal. Instead I think we should do a `null` check and if null use `null`.

NOTE: Rather than commenting in each spot, can you please apply this change throughout

nenaraab:
@rwinch
no problem! Please confirm... i will add some checks (using `Assert`) at the right places to make it bulletproof
- check that identifier is not NULL (if not already checked in the calling public method)
- check that `Spring.valueOf` does not return null

rwinch:
Thanks for the fast reply @nenaraab!

Sorry for the confusion. I was not very clear. What I meant was to do something like this:

```java
String id = parentIdentity.getIdentifier() == null ? 
    null : parentIdentity.getIdentifier().toString()
```

As for the testing I meant to add some tests that break before your changes and are fixed after the changes. Also add a test that validates that a null value is passed through if `getIdentifier()` is passed in. I'd probably add them to [JdbcAclServiceTests](https://github.com/spring-projects/spring-security/blob/master/acl/src/test/java/org/springframework/security/acls/jdbc/JdbcAclServiceTests.java#L54).

To make testing more realistic, I'd:
* [Change JdbcTemplate](https://github.com/spring-projects/spring-security/blob/5.1.2.RELEASE/acl/src/main/java/org/springframework/security/acls/jdbc/JdbcAclService.java#L69) to be `JdbcOperations`
* Add a constructor that allows injecting the `JdbcOperations` and `LookupStrategy`

Then the tests could inject a Mock `JdbcOperations` for the tests mentioned above.

Thanks again for your help with this ticket! If you have any questions, don't hesitate to reach out.

nenaraab:
@rwinch 
It's not that i dislike to implement a unittest... and thanks for the guidance here.

But as `ObjectIdentity.getIdentifier()` never returns `null` (see [documentation](https://github.com/spring-projects/spring-security/blob/5.1.2.RELEASE/acl/src/main/java/org/springframework/security/acls/model/ObjectIdentity.java#L57), see [all constructor implementations](https://github.com/spring-projects/spring-security/blob/5.1.2.RELEASE/acl/src/main/java/org/springframework/security/acls/domain/ObjectIdentityImpl.java#L45)) i don't get the point of introducing a null check at all.

Furthermore I've checked the ObjectIdentityImplTests thoroughly. All not null assertions according to the identifier are covered by tests.

acl/src/test/java/org/springframework/security/acls/jdbc/JdbcAclServiceTests.java
```diff
@@ -57,8 +57,8 @@ public void setUp() {
 	public void readAclByIdMissingAcl() {
 		Map<ObjectIdentity, Acl> result = new HashMap<>();
 		when(
-				lookupStrategy.readAclsById(anyListOf(ObjectIdentity.class),
```
rwinch: If we modify existing tests we run the risk of breaking other functionality later down the road. Please leave existing tests as is and create a new test that validates the new beahvior.

nenaraab: Sorry, my fault, I was not clear on the change of the test case.

I've created a new issue / pull request to discuss and fix that issue:
https://github.com/spring-projects/spring-security/issues/6212

And I've have reverted the change on the test case :-)

rwinch: No problem. Thanks for reverting that and submitting the other PR.

Can you still add [some tests](https://github.com/spring-projects/spring-security/pull/6050#issuecomment-443025708) please? No need to add one for the null value (which as you pointed out cannot happen), but we should add one for the issue that we are fixing (i.e. ensure that non String types are converted).

Thanks!

nenaraab:
@rwinch 
great, added two test cases to test the code change.

Furthermore I've added integration tests (using hsql in-memory database) to test also a proper conversion of the fetched entries, and the find-children sql queries as well.

acl/src/main/java/org/springframework/security/acls/jdbc/JdbcAclService.java
```diff
@@ -79,7 +80,16 @@
 	public JdbcAclService(DataSource dataSource, LookupStrategy lookupStrategy) {
 		Assert.notNull(dataSource, "DataSource required");
 		Assert.notNull(lookupStrategy, "LookupStrategy required");
-		this.jdbcTemplate = new JdbcTemplate(dataSource);
+		this.jdbcOperations = new JdbcTemplate(dataSource);
+		this.lookupStrategy = lookupStrategy;
+		this.aclClassIdUtils = new AclClassIdUtils();
+	}
+
+	public JdbcAclService(DataSource dataSource, LookupStrategy lookupStrategy, JdbcOperations jdbcOperations) {
```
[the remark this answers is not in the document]
nenaraab: Great, like your suggestion! I've kept the `Assert.notNull(dataSource, "DataSource required");` in the "old" constructor for compatibility reason...

acl/src/main/java/org/springframework/security/acls/jdbc/JdbcAclService.java
```diff
@@ -77,9 +78,14 @@
 	// ===================================================================================================
 
 	public JdbcAclService(DataSource dataSource, LookupStrategy lookupStrategy) {
+		this(new JdbcTemplate(dataSource), lookupStrategy);
 		Assert.notNull(dataSource, "DataSource required");
```
rwinch: this assertion is not necessary because JdbcTemplate validates that the dataSource is not null.

nenaraab: I'm so sorry! 
[this @Nullable`](https://github.com/spring-projects/spring-framework/blob/master/spring-jdbc/src/main/java/org/springframework/jdbc/support/JdbcAccessor.java#L45) was missleading...
but you're right: the method `afterPropertiesSet` raises an exception, so I've removed the assertion. Sorry for inconvenience...

rwinch: No problem. It is a bit confusing and due the fact there is also a default constructor.

Thanks for the fast turnaround!

juzerali:
In which version is this fix available?

jzheaux:
You can see by looking at the commit hash - https://github.com/spring-projects/spring-security/commit/d1a754fcf25ef0e59b0c46f80f34b0921ecd1c13. In the comment at the top, the versions are listed. Let me know if I can help further.

juzerali:
Got it, so it is available since v5.2. Thanks @jzheaux

Issue #6052, opened 2018-11-06
Reactive OAuthResourceServerSpec should allow custom error handling


jzheaux:
Servlet-based Spring Security allows for custom error handling via the DSL:

```java
http
    .oauth2ResourceServer()
        .accessDeniedHandler(...)
        .authenticationEntryPoint(...)
        ....
```

It would be nice if Spring Security WebFlux Resource Server could do the same.

sepanniemi:
Should it be possible to customize the error handling via ExceptionHandlingSpec in ServerHttpSecurity? 

It seems that currently defining access denied handler like this has an effect

```
serverHttpSecurity
            .exceptionHandling()
                .accessDeniedHandler(accessDeniedHandler)
```

but defining authentication entry point in a similar way does not have an effect.
``` 
serverHttpSecurity
            .exceptionHandling()
                .authenticationEntryPoint(authEntryPoint)
```

and that's because AuthWebFilter is getting a hardcoded authentication entry point in the configuration?

https://github.com/spring-projects/spring-security/blob/master/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java#L988

PR #6057, opened 2018-11-07
Add Reactive Custom Bearer Token Error Handling Support

jzheaux:
Users can specify a custom access denied handler and authentication
entry point for reactive resource servers.

Fixes: gh-6052


jzheaux:
Thanks, @rwinch

Issue #6065, opened 2018-11-09
Docs: Fix Maven Property example `spring-security.version`


ghillert:
### Summary

The closing tag for the Spring Security Maven property example in the docs is incorrect:

```xml
<spring-security.version>{spring-security-version}</spring.security.version>
```
### Version

The affects version `MASTER`.

PR #6066, opened 2018-11-09
Docs: Fix Maven Property example `spring-security.version`

ghillert:
resolves https://github.com/spring-projects/spring-security/issues/6065

Should this possibly be back-ported?


rwinch:
Thanks for the PR @ghillert! I backported this to 5.1.x via #6081

PR #6070, opened 2018-11-10
Make InMemory*ClientRegistrationRepository Consistent

dperezcabrera:
Fix inconsistencies for class constuctors:
- InMemoryReactiveClientRegistrationRepository
- InMemoryClientRegistrationRepository

The ConcurrentHashMap is not needed because of [this](https://stackoverflow.com/a/5951203/4796021).

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/InMemoryClientRegistrationRepository.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/InMemoryClientRegistrationRepository.java
@@ -18,16 +18,13 @@ package org.springframework.security.oauth2.client.registration;
 import org.springframework.util.Assert;
 
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ConcurrentMap;
 import java.util.function.Function;
-import java.util.stream.Collector;
-
-import static java.util.stream.Collectors.collectingAndThen;
-import static java.util.stream.Collectors.toConcurrentMap;
+import java.util.stream.Collectors;
 
 /**
  * A {@link ClientRegistrationRepository} that stores {@link ClientRegistration}(s) in-memory.
@@ -39,6 +36,7 @@ import static java.util.stream.Collectors.toConcurrentMap;
  * @see ClientRegistration
  */
 public final class InMemoryClientRegistrationRepository implements ClientRegistrationRepository, Iterable<ClientRegistration> {
+
 	private final Map<String, ClientRegistration> registrations;
 
 	/**
@@ -47,7 +45,8 @@ public final class InMemoryClientRegistrationRepository implements ClientRegistr
 	 * @param registrations the client registration(s)
 	 */
 	public InMemoryClientRegistrationRepository(ClientRegistration... registrations) {
-		this(Arrays.asList(registrations));
+		Assert.notEmpty(registrations, "registrations cannot be empty");
+		this.registrations = createClientRegistrationIdToClientRegistration(Arrays.asList(registrations));
 	}
 
 	/**
@@ -57,10 +56,13 @@ public final class InMemoryClientRegistrationRepository implements ClientRegistr
 	 */
 	public InMemoryClientRegistrationRepository(List<ClientRegistration> registrations) {
 		Assert.notEmpty(registrations, "registrations cannot be empty");
-		Collector<ClientRegistration, ?, ConcurrentMap<String, ClientRegistration>> collector =
-			toConcurrentMap(ClientRegistration::getRegistrationId, Function.identity());
-		this.registrations = registrations.stream()
-			.collect(collectingAndThen(collector, Collections::unmodifiableMap));
+		this.registrations = createClientRegistrationIdToClientRegistration(registrations);
+	}
+
+	private static Map<String, ClientRegistration> createClientRegistrationIdToClientRegistration(Collection<ClientRegistration> registrations) {
+		return Collections.unmodifiableMap(registrations.stream()
+				.peek(registration -> Assert.notNull(registration, "registrations cannot contain null values"))
+				.collect(Collectors.toMap(ClientRegistration::getRegistrationId, Function.identity())));
 	}
 
 	@Override
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/InMemoryReactiveClientRegistrationRepository.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/InMemoryReactiveClientRegistrationRepository.java
@@ -17,12 +17,6 @@ package org.springframework.security.oauth2.client.registration;
 
 import java.util.Iterator;
 import java.util.List;
-import java.util.Map;
-import java.util.function.Function;
-import java.util.stream.Collectors;
-
-import org.springframework.util.Assert;
-import org.springframework.util.ConcurrentReferenceHashMap;
 
 import reactor.core.publisher.Mono;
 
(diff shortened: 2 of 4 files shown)


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/InMemoryClientRegistrationRepository.java
```diff
@@ -47,7 +44,7 @@
 	 * @param registrations the client registration(s)
 	 */
 	public InMemoryClientRegistrationRepository(ClientRegistration... registrations) {
-		this(Arrays.asList(registrations));
+		this.registrations = toUnmodifiableMapByRegistrationId(Stream.of(checkEmptyRegistrations(registrations, Assert::notEmpty)));
```
rwinch: Is there a reason to change this vs keep it as is?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrationRepositoryUtils.java
```diff
@@ -0,0 +1,51 @@
...
+ * @since 5.1
+ * @see InMemoryClientRegistrationRepository
+ * @see InMemoryReactiveClientRegistrationRepository
+ */
+public final class ClientRegistrationRepositoryUtils {
+
+	private ClientRegistrationRepositoryUtils() {
+	}
+
+	static Map<String, ClientRegistration> toUnmodifiableMapByRegistrationId(Stream<ClientRegistration> registrations) {
```
rwinch: I'd prefer to keep the input as a Collection as it is one less method invocation for the consumer of this API

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrationRepositoryUtils.java
```diff
@@ -0,0 +1,51 @@
...
+		return Collections.unmodifiableMap(registrations
+				.peek(registration -> Assert.notNull(registration, "registrations cannot contain null values"))
+				.collect(Collectors.toMap(ClientRegistration::getRegistrationId, Function.identity())));
+	}
+
+	static <T> T checkEmptyRegistrations(T registrations, BiConsumer<? super T, String> checker){
```
rwinch: Rather than adding a separate method, I think it would be good to validate in the conversion to a Map

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrationRepositoryUtils.java
rwinch: I'd like to see this renamed to `createClientRegistrationIdToClientRegistration`. This aligns more with Spring Security's naming conventions.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/InMemoryReactiveClientRegistrationRepository.java
```diff
@@ -45,12 +45,7 @@
 	 * @param registrations the client registration(s)
 	 */
 	public InMemoryReactiveClientRegistrationRepository(ClientRegistration... registrations) {
-		Assert.notEmpty(registrations, "registrations cannot be empty");
-		this.clientIdToClientRegistration = new ConcurrentReferenceHashMap<>();
-		for (ClientRegistration registration : registrations) {
-			Assert.notNull(registration, "registrations cannot contain null values");
-			this.clientIdToClientRegistration.put(registration.getRegistrationId(), registration);
-		}
+		this.clientIdToClientRegistration = toUnmodifiableMapByRegistrationId(Stream.of(checkEmptyRegistrations(registrations, Assert::notEmpty)));
```
rwinch: Could this be changed to `this(Arrays.asList(registrations))`

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/InMemoryClientRegistrationRepository.java
dperezcabrera: registrations can be null

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrationRepositoryUtils.java
dperezcabrera: ok

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/InMemoryReactiveClientRegistrationRepository.java
dperezcabrera: registrations can be null. should It throw a NPE or an illegal argument exception?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrationRepositoryUtils.java
dperezcabrera: Thank you very much for the comments, I find them very instructive. If you run the PR unit tests without the main sources, you'll see that the pairs of builders are not fully consistent. Do you think it's better now?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/InMemoryClientRegistrationRepository.java
rwinch: I think that is fine. If it is null we should throw an exception.

rwinch:
Looking at this again, I think we could clean it up a bit more. Rather than externalizing the creation of the `Map` in  `ClientRegistrationRepositoryUtils` we could change `InMemoryReactiveClientRegistrationRepository` to delegate to `InMemoryClientRegistrationRepository` and keep the creation logic private to `InMemoryClientRegistrationRepository`

dperezcabrera:
This is a good idea.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/InMemoryReactiveClientRegistrationRepository.java
```diff
@@ -46,11 +45,7 @@
 	 */
 	public InMemoryReactiveClientRegistrationRepository(ClientRegistration... registrations) {
 		Assert.notEmpty(registrations, "registrations cannot be empty");
-		this.clientIdToClientRegistration = new ConcurrentReferenceHashMap<>();
-		for (ClientRegistration registration : registrations) {
-			Assert.notNull(registration, "registrations cannot contain null values");
-			this.clientIdToClientRegistration.put(registration.getRegistrationId(), registration);
-		}
+		this.clientIdToClientRegistration = createClientRegistrationIdToClientRegistration(Arrays.asList(registrations));
```
rwinch: Sorry it wasn't clear. I was suggesting to keep `createClientRegistrationIdToClientRegistration` a private static method. Then `InMemoryReactiveClientRegistrationRepository` would no longer use a `Map`. Instead, it would use an instance of `InMemoryClientRegistrationRepository`. For example:

```java
this.delegate = new InMemoryClientRegistrationRepository(registrations);
```

```java
@Override
public Mono<ClientRegistration> findByRegistrationId(String registrationId) {
	return Mono.justOrEmpty(this.delegate.findByRegistrationId(registrationId));
}
```

dperezcabrera: Okay, what about now?

rwinch: Thanks! The last thing is ...Can you please clean up the commit message to follow the guidelines in [Contributing](https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.md#format-commit-messages)? I'm not super strict about the 50 character subject, but I do prefer it to be as concise as possible. Something like:

```
Make InMemory*ClientRegistrationRepository Consistent

The previous builders with the list argument were inconsistent with their 
respective builders of var args.
```

dperezcabrera: It's done, thank you very much!

rwinch:
Thanks for the PR @dperezcabrera! This is now merged into 5.1.x and master via #6113

Issue #6015, opened 2018-10-23
Reactive Nimbus Jwt Decoder should convert claims


jzheaux:
Add the reactive equivalent for converting claims on read, like the servlet-based version does with `MappedJwtClaimSetConverter`.

jzheaux:
Related to #5223

PR #6080, opened 2018-11-13
Reactive Jwt Claim Set Converter Support

jzheaux:
Exposes setClaimSetConverter on NimbusReactiveJwtDecoder, lining it up
with the same support on NimbusJwtDecoder.

Fixes: gh-6015


rwinch:
Thanks @jzheaux This is now merged into master

Issue #6088, opened 2018-11-14
Update webflux-form sample to use Thymeleaf Built in CSRF Support


rwinch:
### Summary

Backport of #6061

NOTE: This will be resolved by merging changes from the original issue above, so there is no need for any extra work for this ticket (no PR necessary).

raphaelDL:
@rwinch  Can I take this?

rwinch:
@raphaelDL Sorry for not being clear, this will be merged from the original issue and so no extra work will be necessary. I updated the ticket to clarify this. Sorry again and thanks for contributing!

raphaelDL:
Sorry language barrier. I'm picking up the terms. Thank you anyway :)

rwinch:
@raphaelDL Not a problem at all. I wasn't clear, so it is really my fault. We are really looking to grow contributions. If you are looking for issues to work on please let me know. I did provide a few at https://github.com/spring-projects/spring-security/pull/6093#issuecomment-438812117 As mentioned there just comment on the ticket. Thanks again for your willingness to contribute!

rwinch:
Fixed via #8655caa2de7dee6b9d86ca4db06190d4394c36c1

Issue #6087, opened 2018-11-14
Jackson exception when deserialising token response


hfgbarrigas:
### Summary
When dealing with additional information in the token response (<String, Object>) jackson throws an exception. Looking at OAuth2AccessTokenResponseBody class to extract the OAuth2AccessTokenResponse:

`
@Override
public Mono<OAuth2AccessTokenResponse> extract(ReactiveHttpInputMessage inputMessage,
			Context context) {
		ParameterizedTypeReference<Map<String, String>> type = new ParameterizedTypeReference<Map<String, String>>() {};
		BodyExtractor<Mono<Map<String, String>>, ReactiveHttpInputMessage> delegate = BodyExtractors.toMono(type);
		return delegate.extract(inputMessage, context)
				.map(json -> parse(json))
				.flatMap(OAuth2AccessTokenResponseBodyExtractor::oauth2AccessTokenResponse)
				.map(OAuth2AccessTokenResponseBodyExtractor::oauth2AccessTokenResponse);
}
`
Looks like Map<String, String> is forced and values other than strings cannot be parsed. Down below on _parse (line 66)_ new JSONObject() can deal with Map<String, ?>.

### Actual Behavior
Exception thrown on jackson.

> Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize instance of java.lang.String out of START_ARRAY token
at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: java.util.LinkedHashMap["client_authorities"])

### Expected Behavior

Token should be parsed correctly and _OAuth2AccessTokenResponse_ correctly constructed.

### Version

5.1.3.BUILD-SNAPSHOT

### Sample

https://github.com/hfgbarrigas/oauth2client-webflux-error-sample

hfgbarrigas:
Great. Is there a snapshot to test?

Issue #6075, opened 2018-11-13
ReactorContextTestExecutionListener should use named hooks


bsideup:
NOTE: This is marked for first time contributors only. If you want the issue, please comment to claim it. If you need any help with this issue, just ask on the ticket and we will be glad to help. 
### Summary

Currently, `ReactorContextTestExecutionListener` resets all previously set `onLastOperator` hooks:
https://github.com/spring-projects/spring-security/blob/dca3645850e082d845e17a6572af84aa44b97f12/test/src/main/java/org/springframework/security/test/context/support/ReactorContextTestExecutionListener.java#L66-L68
This might affect users' code, and it is recommended to use named hooks instead.

### Actual Behavior
`ReactorContextTestExecutionListener` resets all pre-configured `onLastOperator` hooks after test method.

### Expected Behavior
ReactorContextTestExecutionListener resets only the hooks it sets.

### Version
5.1.1.RELEASE

rwinch:
This would be fixed by [registering](https://github.com/spring-projects/spring-security/blob/dca3645850e082d845e17a6572af84aa44b97f12/test/src/main/java/org/springframework/security/test/context/support/ReactorContextTestExecutionListener.java#L62) it via 

```java
private static final String CONTEXT_OPERATOR_KEY = SecurityContext.class.getName();
...
Hooks.onLastOperator(CONTEXT_OPERATOR_KEY, Operators.lift((s, sub) -> new SecuritySubContext<>(sub, securityContext)));
```

Then it would get removed via

```java
Hooks.resetOnLastOperator(CONTEXT_OPERATOR_KEY); 
```

We then want to ensure we [add a test](https://github.com/spring-projects/spring-security/blob/dca3645850e082d845e17a6572af84aa44b97f12/test/src/test/java/org/springframework/security/test/context/support/ReactorContextTestExecutionListenerTests.java#L176) that validates if a different Hook was registered, then `.afterTestMethod` will not remove that different Hook.

ankitthakur:
I think, it is working correctly as is expected, unless we would like to create some singleton api for test environment. Otherwise this approach might be useless on different environments with different db configurations, Like Linux with Oracle test cases could be different than With SqlServer. Whereas it is common across both platform for business logics.

daejii:
Mine 😁🥳

rwinch:
[...]

Issue #6061, opened 2018-11-08
Update webflux-form sample to use Thymeleaf Built in CSRF Support


rwinch:
### Summary

Thymeleaf provides [automatic integration with Spring Security's CSRF](https://github.com/spring-projects/spring-security/issues/5867#issuecomment-423327852) support. We should update the [webflux-form](https://github.com/spring-projects/spring-security/tree/5.1.1.RELEASE/samples/boot/webflux-form) sample to demonstrate it.

* Update the dependencies to include both org.springframework.boot:spring-boot-starter-thymeleaf org.thymeleaf.extras:thymeleaf-extras-springsecurity5
* Remove the [CsrfControllerAdvice](https://github.com/spring-projects/spring-security/blob/5.1.1.RELEASE/samples/boot/webflux-form/src/main/java/sample/CsrfControllerAdvice.java)
* Ensure the tests still pass by running `../../../gradlew check`

Daniel69:
I'd like to take this one 😁

rwinch:
Thanks @Daniel69 The issue is all yours! If you need any help, please don't hesitate to reach out to me

Daniel69:
Thank @rwinch, I'll start by reading https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.md

Daniel69:
@rwinch I removed the CsrfControllerAdvice and added a dependency to org.thymeleaf.extras:thymeleaf-extras-springsecurity5:3.0.4.RELEASE because  'org.thymeleaf.extras:thymeleaf-extras-springsecurity' alone does not resolve, but two tests fail once I remove the Controller Advice (loginWhenInvalidUsernameThenError and loginAndLogout).
When I run the example alone after these changes and attempt to login, the only message that I can see is : 'CSRF Token has been associated to this client',  no matter I use 'user' or 'invalid' as username.

rwinch:
@Daniel69 Sorry for the delay getting back to you. The problem is that we aren't using Spring Boot 2.1.0.RELEASE yet. I created #6082 which will resolve this. Once it is updated to Boot 2.1.0.RELEASE you should be able to remove the version from `org.thymeleaf.extras:thymeleaf-extras-springsecurity5` too

rwinch:
Thanks again for your patience. Now that #6082 has been resolved, you should be able to rebase off of master and remove the version from thymeleaf-extras-springsecurity5

PR #6093, opened 2018-11-14
CSRF mitigation is now handled using Thymeleaf

raphaelDL:
From   #6088


rwinch:
Thanks for the PR @raphaelDL!

Sorry I wasn't clear on #6088, but this will get fixed just by merging the results of #6061 into the 5.1.x branch. I'm going to close this in favor of that work since someone already claimed that ticket.

If you are looking for another ticket, perhaps you would be interested in trying #6075 (I think this will be pretty simple) or #6087? If so, please just claim one of them by commenting on the ticket. Of course if you have any questions, I'd be more than happy to help.

PR #6096, opened 2018-11-14
OAuth2AccessTokenResponseBodyExtractor now handles additional attribu…

raphaelDL:
From #6087 

@rwinch help please,  I took this one but I wonder if this is the right approach for this. What do you think?


rwinch:
@raphaelDL Thank you so much for the PR! The code is exactly what I had imagined. I have a few  minor suggestions that help ensure project consistency. 

1) Can you please clean up the commit message to follow Spring's [commit message format](https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.md#format-commit-messages)? In particular, I'd like to see the subject line as concise as possible. I realize this can be difficult to keep under 50, but we really like to keep it concise. The second aspect is we like the commit to end in a line that states "Fixes: gh-6087" This means that when the Pull Request is merged, it automatically closes the ticket and links the commit to the ticket for tracking purposes. An example of what the cleaned up commit might look like is:

```
OAuth2AccessTokenResponseBodyExtractor supports Object values

This commit ensures the token response is parsed correctly if the values are not a String.

Fixes: gh-6087
```

2) In the test, can you please add a comment that refers to the github issue that it is testing for? This helps us for tractability later on. For example, if we run into that test breaking when we do some refactoring later, we can easily see the issue that it is testing for and get more context on why it was necessary. The comment would look like:

```java
@Test
// gh-6087
public void oauth2AccessTokenResponseWhenMultipleAttributeTypesThenCreated() throws Exception {

``` 

You can make these changes locally, amend the commit, and force push to your branch and this PR will be updated. As always, if you have any questions, feedback, or need any help, please don't hesitate to reach out.

Thanks again for your help!

raphaelDL:
Working on it :)

oauth2/oauth2-core/src/test/java/org/springframework/security/oauth2/core/web/reactive/function/OAuth2BodyExtractorsTests.java
```diff
@@ -120,4 +120,33 @@ public void oauth2AccessTokenResponseWhenValidThenCreated() throws Exception {
 		assertThat(result.getRefreshToken().getTokenValue()).isEqualTo("tGzv3JOkF0XG5Qx2TlKWIA");
 		assertThat(result.getAdditionalParameters()).containsEntry("example_parameter", "example_value");
 	}
+
+
+	@Test
+	// Fixes: gh-6087
```
rwinch: While the convention is that the commit message says "Fixes: ", the comment should just be gh-6087. Can you please change this and then I think we are ready to merge.

raphaelDL: yup

Issue #6089, opened 2018-11-14
WebClientReactiveClientCredentialsTokenResponseClient should fail if not 2xx status


rwinch:
### Summary

If anything but a 2xx status is returned (i.e. a redirect) when trying to get the token, we should fail

See #6071

raphaelDL:
maybe something like this? 🤔

```java
return this.webClient.post()
					.uri(tokenUri)
					.accept(MediaType.APPLICATION_JSON)
					.headers(headers(clientRegistration))
					.body(body)
					.exchange()
					.filterWhen(response -> Mono.just(!response.statusCode().is2xxSuccessful()))
					.flatMap(response -> response.body(oauth2AccessTokenResponse()))
					.map(response -> {
						if (response.getAccessToken().getScopes().isEmpty()) {
							response = OAuth2AccessTokenResponse.withResponse(response)
								.scopes(authorizationGrantRequest.getClientRegistration().getScopes())
								.build();
						}
						return response;
					});
		});
´´´

rwinch:
Thanks for the comment!

* You can simplify it using .filter which does not require a `Mono`
* We should `switchIfEmpty` with a `Mono.error` that contains a message stating that the attempt to get the token returned an unsuccessful status along with the status code.

Would you like to take this one @raphaelDL?

raphaelDL:
yup

rwinch:
Thanks @raphaelDL! As always let me know if you have any questions!

raphaelDL:
Not completely sure how to approach this, if I filter the response just after I call exchange, I won't be able to recover the status at the end when calling swtchIfEmpty. Should I replay the mono values ?

I tried this:

´´´java
return this.webClient.post()
					.uri(tokenUri)
					.accept(MediaType.APPLICATION_JSON)
					.headers(headers(clientRegistration))
					.body(body)
					.exchange()
					.map(response ->{
						if(!response.statusCode().is2xxSuccessful()){
							throw
									WebClientResponseException.create(response.rawStatusCode(),
											"Cannot get token, expected 2xx HTTP Status code",
											null,
											null,
											null
									);
						}; return response;
					})
					.flatMap(response -> response.body(oauth2AccessTokenResponse()))
					.map(response -> {
						if (response.getAccessToken().getScopes().isEmpty()) {
							response = OAuth2AccessTokenResponse.withResponse(response)
								.scopes(authorizationGrantRequest.getClientRegistration().getScopes())
								.build();
						}
[...]

PR #6105, opened 2018-11-16
WebClientReactiveClientCredentialsTokenResponseClient.getTokenRespons…

raphaelDL:
WebClientReactiveClientCredentialsTokenResponseClient.getTokenResponse expects 2xx http status code

This ensures that token response is only extracted when ClientResponse has a successful status

Fixes: gh-6089


rwinch:
Thanks for the PR @raphaelDL! You are crushing it :1st_place_medal: 

This is now merged into master and 5.1.x via #6112

PR #6109, opened 2018-11-17
OAuth2 client cannot work well with reactor redis session

wangzw:
Fix Issue #6215

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/WebSessionOAuth2ServerAuthorizationRequestRepository.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/WebSessionOAuth2ServerAuthorizationRequestRepository.java
@@ -53,7 +53,7 @@ public final class WebSessionOAuth2ServerAuthorizationRequestRepository
 		if (state == null) {
 			return Mono.empty();
 		}
-		return getStateToAuthorizationRequest(exchange, false)
+		return getStateToAuthorizationRequest(exchange)
 				.filter(stateToAuthorizationRequest -> stateToAuthorizationRequest.containsKey(state))
 				.map(stateToAuthorizationRequest -> stateToAuthorizationRequest.get(state));
 	}
@@ -62,9 +62,8 @@ public final class WebSessionOAuth2ServerAuthorizationRequestRepository
 	public Mono<Void> saveAuthorizationRequest(
 			OAuth2AuthorizationRequest authorizationRequest, ServerWebExchange exchange) {
 		Assert.notNull(authorizationRequest, "authorizationRequest cannot be null");
-		return getStateToAuthorizationRequest(exchange, true)
-				.doOnNext(stateToAuthorizationRequest -> stateToAuthorizationRequest.put(authorizationRequest.getState(), authorizationRequest))
-				.then();
+		return saveStateToAuthorizationRequest(exchange).doOnNext(stateToAuthorizationRequest ->
+				stateToAuthorizationRequest.put(authorizationRequest.getState(), authorizationRequest)).then();
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


rwinch:
Thanks for the PR @wangzw! Before we merge, I need to understand a bit better what is happening. We will also want a test. Let's move our discussion to #6108 to iron out what is happening first.

wangzw:
@rwinch Any comments?

cschockaert:
This fix is working, thanks !!

symptom is:

when you create a session and add a request authorization for the first time to it, it is correctly setting into redis.

if you add another request authorization (make another call to /oauth2 login endpoint) with a JESSIONID pointing to the previous created redis backuped session, then this newly request authorization object is not saved into the existing http redis backuped session.

wangzw:
@rwinch Finally got time to add test case. Please review.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/WebSessionOAuth2ServerAuthorizationRequestRepository.java
```diff
@@ -53,7 +53,7 @@
 		if (state == null) {
 			return Mono.empty();
 		}
-		return getStateToAuthorizationRequest(exchange, false)
```
rwinch: This looks like it breaks functionality. It should not add an empty Map if we are loading a value and nothing exists.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/server/WebSessionOAuth2ServerAuthorizationRequestRepositoryTests.java
```diff
@@ -99,8 +103,175 @@ public void loadAuthorizationRequestWhenSavedThenAuthorizationRequest() {
 				.verifyComplete();
 	}
 
+	/**
+	 * Redis and Hazelcast distributed session implementation only update the change to underlay storage.
+	 * This is a simulation of such distributed session implementation.
+	 */
+	private static class DeltaOptimizedWebSession extends MockWebSession {
```
rwinch: I think we should be able to use a Mock object here instead of extending MockWebSession

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/server/WebSessionOAuth2ServerAuthorizationRequestRepositoryTests.java
```diff
@@ -99,8 +103,175 @@ public void loadAuthorizationRequestWhenSavedThenAuthorizationRequest() {
...
+			}
+
+			return Mono.empty();
+		}
+	}
+
+	/**
+	 * A simulation to capture the session data change.
+	 */
+	private static class DeltaOptimizedMap implements Map<String, Object> {
```
rwinch: This should not be needed if we use a mock object instead

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/server/WebSessionOAuth2ServerAuthorizationRequestRepositoryTests.java
```diff
@@ -99,8 +103,175 @@ public void loadAuthorizationRequestWhenSavedThenAuthorizationRequest() {
...
+		}
+
+		@Override
+		public Set<Entry<String, Object>> entrySet() {
+			return cached.entrySet();
+		}
+	}
+
 	@Test
-	public void loadAuthorizationRequestWhenMultipleSavedThenAuthorizationRequest() {
```
rwinch: Please do not modifiy any of the existing tests as this means that we might have broken something. Please instead add a new test that does not change existing tests

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/WebSessionOAuth2ServerAuthorizationRequestRepository.java
wangzw: The behaviour here is kept the same as before and I only split `getStateToAuthorizationRequest(ServerWebExchange exchange, boolean create)` into `getStateToAuthorizationRequest(ServerWebExchange exchange)` and `saveStateToAuthorizationRequest(ServerWebExchange exchange)` to make it clear.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/server/WebSessionOAuth2ServerAuthorizationRequestRepositoryTests.java
wangzw: Let me see how to use a Mock object here.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/WebSessionOAuth2ServerAuthorizationRequestRepository.java
wangzw: And the fix is in `saveStateToAuthorizationRequest`

wangzw:
Rewrite test case with mock object and rebase against master.

rwinch:
Thanks for the PR! This is merged via a60fd43534db4dd515f4aea1ba5aef73a3913172 I also polished the formatting via 752733e8de867c74d613715ee2591c907b9f5d8d

Issue #4469, opened 2017-07-25
WebSocket matchers ignore parameters


karayv:
### Summary
I want to authorize topic subscription by topic name. `.simpSubscribeDestMatchers("/topic/list/{location}/**")                   .access("@webSecurity.checkLocation(authentication,#location)")`

### Actual Behavior
My `location` parameter is not passed to `webSecurity.checkLocation()`. The method is called, but the parameter is `null`.

### Expected Behavior
Correct, non-null, `location` parameter passed to `webSecurity.checkLocation()` method. According to the documentation this is possible for antMatchers
https://docs.spring.io/spring-security/site/docs/current/reference/htmlsingle/#el-access-web-path-variables

Please suggest workarounds if exist.

### Configuration
```java
@Configuration
public class WebSocketSecurityConfig extends AbstractSecurityWebSocketMessageBrokerConfigurer {

    @Override
    protected boolean sameOriginDisabled() {
        return true;
    }

    @Override
    protected void configureInbound(MessageSecurityMetadataSourceRegistry messages) {
        messages //
                .nullDestMatcher().authenticated() //
                .simpDestMatchers("/app/**").authenticated() //
                .simpSubscribeDestMatchers("/topic/list/{location}/**")
                      .access("@webSecurity.checkLocation(authentication,#location)") //
                .anyMessage().denyAll();
    }
}
```

### Version
   `compile group: 'org.springframework.security.oauth', name: 'spring-security-oauth2', version: '2.1.1.RELEASE'`

   `compile group: 'org.springframework.security', name: 'spring-security-messaging', version: '4.2.3.RELEASE'`

karayv:
I found a workaround.

You can use implicit variable `message`:
`.simpSubscribeDestMatchers("/topic/list/*")
                .access("@webSocketSecurity.checkLocationByMsg(authentication,message)")`

And you need extra work to be done in the service:
```java
@Component
public class WebSocketSecurity {

    @Autowired
    private WebSecurity webSecurity;

    public boolean checkLocationByMsg(Authentication authentication, Message<?> message) {

        StompHeaderAccessor sha = StompHeaderAccessor.wrap(message);
        String topic = sha.getDestination();
        String location = topic.replace(MyConstants.listTopicPrefix, "");

        return webSecurity. checkLocation(authentication, location);
    }
```

neuhalje:
Here is my (even more hackish) approach to extract the path variables:

```java

@Configuration
[...]

PR #6110, opened 2018-11-18
Support Path Variables in Message Expressions

dbuos:
Extract path variables expressed in SimpDestinationMessageMatcher's pattern.

Fixes: gh-4469

### DIFF
--- a/messaging/src/main/java/org/springframework/security/messaging/access/expression/MessageExpressionConfigAttribute.java
+++ b/messaging/src/main/java/org/springframework/security/messaging/access/expression/MessageExpressionConfigAttribute.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


messaging/src/main/java/org/springframework/security/messaging/access/expression/MessageExpressionConfigAttribute.java
```diff
@@ -15,30 +15,48 @@
...
 		this.authorizeExpression = authorizeExpression;
+		this.matcher = matcher;
+	}
+
+	/**
+	 * Creates a new instance
+	 *
+	 * @param authorizeExpression the {@link Expression} to use. Cannot be null
+	 */
+	public MessageExpressionConfigAttribute(Expression authorizeExpression) {
```
rwinch: Since this is a package private class we don't need to have this additional constructor. We should remove it because other wise there is the chance that the matcher is null which means we need to do null checks within the code. If we validate up front this is non-null then we no longer need null checks

messaging/src/main/java/org/springframework/security/messaging/access/expression/MessageExpressionConfigAttribute.java
```diff
@@ -53,4 +71,15 @@ public String getAttribute() {
 	public String toString() {
 		return authorizeExpression.getExpressionString();
 	}
+
+	@Override
+	public EvaluationContext postProcess(EvaluationContext ctx, Message<?> message) {
+		if (matcher instanceof SimpDestinationMessageMatcher) {
```
rwinch: We either need to validate when this is constructed that matcher is not null or perform a null check here.

Daniel69: @rwinch, already validated in the constructor as the previous comment says, anyway the "null instanceof SomeClass" expression evaluates to false.

luc-clement:
Is this PR ready to be merged or are they still changes awaited ?

rwinch:
Thanks for the PR @dbuos! This is now merged into master and will be part of Spring Security 5.2 via f97ac4daa6403d9909b1009ebb71b64ba5decc77 I applied a bit of polish via fcb4def4ac0d1c9ad6230a4b38db8b2bddb9a66c

Issue #6106, opened 2018-11-16
Update to Gradle 4.10.2


raphaelDL:
With current version of Gradle 4.10.1 when I run a task with

`./gradlew ...`

I get these warnings:

`> Configure project :spring-security-aspects
The Task.deleteAllActions() method has been deprecated. This is scheduled to be removed in Gradle 5.0.`

`> Configure project :spring-security-samples-javaconfig-x509
Found slf4j-api dependency but no providers were found.  Did you mean to add slf4j-simple? See https://www.slf4j.org/codes.html#noProviders .`

What if we do the next?

1) Upgrade Gradle
`./gradlew wrapper --gradle-version 4.10.2 `

2) Update `buildSrc/src/main/groovy/aspectj/AspectJPlugin.groovy`

from `javaCompileTask.deleteAllActions()`
to
`javaCompileTask.setActions new ArrayList(0)` or `javaCompileTask.setActions Arrays.asList()`

3) Update `/samples/javaconfig/x509/spring-security-samples-javaconfig-x509.gradle`
Remove ` compile 'org.slf4j:slf4j-api'`

rwinch:
@raphaelDL I am not sure I understand this issue. Did you try to update to Gradle 4.10.2 and it failed? I know there are warnings, but that seems distinct from the subject of `Update to Gradle 4.10.2`. Can you help me understand please?

rwinch:
Thanks. Yes. We'd love to update to a newer version of Gradle and (in separate PRs as you mentioned) remove the deprecated APIs.

PR #6114, opened 2018-11-20
Update to Gradle 4.10.2

raphaelDL:
Fixes gh-6106


rwinch:
Thanks for the PR @raphaelDL! This is now merged into master and backported to 5.1.x via #6124

Issue #6128, opened 2018-11-23
Replace deprecated Gradle Task.deleteAllActions() method in AspectJPlugin.groovy


raphaelDL:
`buildSrc/src/main/groovy/aspectj/AspectJPlugin.groovy`
uses
`javaCompileTask.deleteAllActions()`
 but this method is deprecated, this gives us a warning message when executing gradlew tasks.
 We should now replace it so it won't give us future issues.

One approach can be:
`javaCompileTask.setActions Arrays.asList()`

If we want to delete all actions, we can use the setActions method and pass an empty list

PR #6129, opened 2018-11-23
Replace deprecated Gradle Task method in AspectJPlugin.groovy

raphaelDL:
This commit ensures that the method Task.deleteAllActions is not used

Fixes: gh-6128


jgrandja:
Thanks for the PR @raphaelDL. This is now in master.

Issue #6130, opened 2018-11-23
Remove unused compile dependency in javaconfig x509 sample


raphaelDL:
When building the project a warning is thrown stating that `org.slf4j:slf4j-api` is defined in `samples/javaconfig/x509/spring-security-samples-javaconfig-x509.gradle` but no implementation is given.

If we remove it we get rid of this warning and have a cleaner building process

Should we remove it?

jgrandja:
@raphaelDL `org.slf4j:slf4j-api` is actually used as it's brought in as a transitive dependency for `org.slf4j:jcl-over-slf4j`, `org.slf4j:jul-to-slf4j`, and `org.slf4j:log4j-over-slf4j`.

What's the exact warning message you're getting when you run the build?

raphaelDL:
oh god, that’s what I thought... 

‘> Configure project :spring-security-samples-javaconfig-x509 Found slf4j-api dependency but no providers were found. Did you mean to add slf4j-simple? See https://www.slf4j.org/codes.html#noProviders .’

jgrandja:
@raphaelDL I see `logback.xml` in the project but no `logback-classic.jar` dependency declared. I think you can add this dependency and remove the other 3x dependencies. See [Binding with a logging framework at deployment time](https://www.slf4j.org/manual.html#swapping).

Give that a try and I think it should resolve the warning.

raphaelDL:
You're all right, it worked.... I pushed a change @jgrandja. Thank you very much for your suggestion

raphaelDL:
I dropped
```
	compile 'org.slf4j:jcl-over-slf4j'
 	compile 'org.slf4j:jul-to-slf4j'
 	compile 'org.slf4j:log4j-over-slf4j'
 	compile 'org.slf4j:slf4j-api'
```

and put
```
compile 'ch.qos.logback:logback-classic'
```
instead

PR #6131, opened 2018-11-23
Remove unused dependency slf4j-api in javaconfig x509 sample application

raphaelDL:
This commit ensures that unused dependencies are not imported, that will prevent warning messages being thrown when building the project.

Fixes: gh-6130


jgrandja:
Thanks for the PR @raphaelDL. This is now in master.

Issue #6092, opened 2018-11-14
SonarQube jobs are not reporting code coverage


jzheaux:
We haven't seen code coverage on our [SonarQube analyses](https://sonar.spring.io/dashboard?id=org.springframework.security%3Aspring-security) for a while.

raphaelDL:
I think I may have parts of the solution, will submit al PR tomorrow, midnight already but I needed to know what was going on with this
![Local SonarQube](https://pbs.twimg.com/media/DsgUgjRV4AY1ywJ.jpg)

raphaelDL:
Turns out the step a) is not needed, I just ran:

`./gradlew clean jacocoTestReport sonarqube -PexcludeProjects='**/samples/**' -Dsonar.jacoco.reportPaths='**/build/jacoco/*.exec'`

and got the coverage report shown again in a fresh new sonar container. Maybe what we were missing was telling sonar where the reports are located.

I think we just need to edit the command in the Sonar step.

As stated in `build.gradle` we need to explicitly call `jacocoTestReport` to get the report, otherwise its will not be generated.

My question is: what do we use wen calling the Sonar routine, Jenkinsfile or travis.yml?

Knowing this I think I can submit a PR

jzheaux:
@raphaelDL wow, thanks for your extra research here! You are the man.

We use `Jenkinsfile` for the sonar routine.

Are you saying that when you run with `./gradlew clean sonarqube -PexcludeProjects='**/samples/**'` you *don't* see code coverage in the Sonar report, but when you do `./gradlew clean jacocoTestReport sonarqube -PexcludeProjects='**/samples/**' -Dsonar.jacoco.reportPaths='**/build/jacoco/*.exec'` then you do see code coverage in the report?

Makes sense, just want to make sure we are on the same page.

If so, then I think that is enough to submit a PR, yes.

raphaelDL:
Absolutely ¡, let's try it

raphaelDL:
@jzheaux @rwinch  

I'm posting a comment here instead of the gradle-plugins repo, 'cause I think is more related with what we have in this project.

Here's what I've found

when the Jacoco plugin is applied coverage reports should be generated when calling `check` and `test` tasks.

We are preventing the reports being generated here: https://github.com/spring-projects/spring-security/blob/7618d236c4562491fb2cead8f866c04dabdca21f/build.gradle#L24

[...]

PR #6135, opened 2018-11-23
Enable Code Coverage Reports in Sonar

raphaelDL:
This commit ensures that jacocoTestReport task is called when running the Sonar stage.
Additionally, a variable is passed instructing Sonar where to find the test result data.

Fixes: gh-6092


jzheaux:
Thanks for the PR @raphaelDL! We have code coverage again in our pipeline. You totally rock!

(Note, this was also backported to [5.1.2](https://github.com/spring-projects/spring-security/issues/6139) and [5.0.10](https://github.com/spring-projects/spring-security/issues/6140))

PR #6141, opened 2018-11-26
Add GenericConversionService with support for UUID and Strings

nenaraab:
### Motivation
As of now only `acl_object_identity.object_id_identity` of type `java.util.Long` are supported. In case you use identifiers of type `java.lang.String` or `java.util.UUID` the application needs to provide a ConversionService. This commit comes with two default Converters to support Strings and UUIDs.

### related commit
 https://github.com/pwheel/spring-security/commit/8efec0243f0f7580c9100d0ac034ac881907cc0a

### related issue(s)
- #4814
- #4424

### DIFF
--- a/acl/src/main/java/org/springframework/security/acls/jdbc/AclClassIdUtils.java
+++ b/acl/src/main/java/org/springframework/security/acls/jdbc/AclClassIdUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,11 +18,17 @@ package org.springframework.security.acls.jdbc;
 import java.io.Serializable;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.util.UUID;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.springframework.core.convert.ConversionFailedException;
 import org.springframework.core.convert.ConversionService;
+import org.springframework.core.convert.TypeDescriptor;
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.core.convert.support.GenericConversionService;
 import org.springframework.security.acls.model.ObjectIdentity;
+import org.springframework.util.Assert;
 
 /**
  * Utility class for helping convert database representations of {@link ObjectIdentity#getIdentifier()} into
@@ -36,9 +42,14 @@ class AclClassIdUtils {
 	private ConversionService conversionService;
 
 	public AclClassIdUtils() {
+		GenericConversionService genericConversionService = new GenericConversionService();
+		genericConversionService.addConverter(String.class, Long.class, new StringToLongConverter());
+		genericConversionService.addConverter(String.class, UUID.class, new StringToUUIDConverter());
+		this.conversionService = genericConversionService;
 	}
 
 	public AclClassIdUtils(ConversionService conversionService) {
+		Assert.notNull(conversionService, "conversionService must not be null");
 		this.conversionService = conversionService;
 	}
 
@@ -90,15 +101,11 @@ class AclClassIdUtils {
 	}
 
 	private <T> boolean canConvertFromStringTo(Class<T> targetType) {
-		return hasConversionService() && conversionService.canConvert(String.class, targetType);
+		return conversionService.canConvert(String.class, targetType);
 	}
 
 	private <T extends Serializable> T convertFromStringTo(String identifier, Class<T> targetType) {
 		return conversionService.convert(identifier, targetType);
-	}
-
-	private boolean hasConversionService() {
-		return conversionService != null;
 	}
 
 	/**
@@ -111,7 +118,7 @@ class AclClassIdUtils {
 	 */
 	private Long convertToLong(Serializable identifier) {
 		Long idAsLong;
-		if (hasConversionService()) {
+		if (canConvertFromStringTo(Long.class)) {
 			idAsLong = conversionService.convert(identifier, Long.class);
 		} else {
 			idAsLong = Long.valueOf(identifier.toString());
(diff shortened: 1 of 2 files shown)


acl/src/main/java/org/springframework/security/acls/jdbc/AclClassIdUtils.java
```diff
@@ -36,10 +41,18 @@
 	private ConversionService conversionService;
 
 	public AclClassIdUtils() {
+		this(null);
```
jzheaux: This means now that the conversion service can never be `null`. That means that all the checks for `hasConversionService` are no longer necessary.

Would you mind creating a ticket to improve that (remove the `hasConversionService` checks) so that we remember for the future?

acl/src/test/java/org/springframework/security/acls/jdbc/AclClassIdUtilsTest.java
```diff
@@ -120,7 +118,9 @@ public void shouldReturnLongIfTypeClassNotFound() throws SQLException {
 	public void shouldReturnLongIfTypeClassCannotBeConverted() throws SQLException {
 		// given
 		given(resultSet.getString("class_id_type")).willReturn("java.lang.Long");
+		given(conversionService.convert(DEFAULT_IDENTIFIER_AS_STRING, Long.class)).willReturn(new Long(DEFAULT_IDENTIFIER_AS_STRING));
```
jzheaux: Really, it feels super smelly that we are mocking this service so that it a) converts Strings to Longs and b) states that it cannot convert Strings to Longs.

We should probably have a separate ticket to clean this up (something like `AclClassIdUtils#convertToLong` should not assume `conversionService` can convert to `Long`). Would you mind adding that?

acl/src/test/java/org/springframework/security/acls/jdbc/AclClassIdUtilsTest.java
```diff
@@ -157,4 +154,15 @@ public void shouldReturnUUIDWhenUUIDClassIdType() throws SQLException {
 		assertThat(newIdentifier).isEqualTo(identifier);
 	}
 
+	@Test
+	public void shouldReturnStringWhenStringClassIdType() throws SQLException {
+		// given
+		given(resultSet.getString("class_id_type")).willReturn("java.lang.String");
+
+		// when
+		Serializable newIdentifier = aclClassIdUtils.identifierFrom(IDENTIFIER_AS_STRING, resultSet);
```
jzheaux: This extra variable seems needless. Could you please use `DEFAULT_IDENTIFIER_AS_STRING` instead?

nenaraab:
Sorry, i was not sure how to correct a pull request best... If i would have squashed it, would it then look prettier, i.e. better readable for review?

acl/src/main/java/org/springframework/security/acls/jdbc/AclClassIdUtils.java
```diff
@@ -18,11 +18,17 @@
 import java.io.Serializable;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.util.UUID;
```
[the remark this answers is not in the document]
nenaraab: Done 👍

acl/src/test/java/org/springframework/security/acls/jdbc/AclClassIdUtilsTest.java
```diff
@@ -157,4 +138,28 @@ public void shouldReturnUUIDWhenUUIDClassIdType() throws SQLException {
 		assertThat(newIdentifier).isEqualTo(identifier);
 	}
 
+	@Test
+	public void shouldReturnStringWhenStringClassIdType() throws SQLException {
+		// given
+		String identifier = "MY_STRING_IDENTIFIER";
```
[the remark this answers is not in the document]
nenaraab: No, this was by intention. For me it is relevant, to make sure that the String2String converter works as expected. But i've moved the variable into the test case.

acl/src/main/java/org/springframework/security/acls/jdbc/AclClassIdUtils.java
```diff
@@ -124,6 +131,31 @@ private boolean isString(Serializable object) {
...
+
+	private static class StringToUUIDConverter implements Converter<String, UUID> {
+		@Override
+		public UUID convert(String identifierAsString) {
+			if (identifierAsString == null) {
+				throw new ConversionFailedException(TypeDescriptor.valueOf(String.class),
+						TypeDescriptor.valueOf(UUID.class), null, null);
+
+			}
+			return UUID.fromString(identifierAsString);
```
nenaraab: @jzheaux 

This is only called in case the custom ConversionService passed in via `SetConversionService` does not support StringToLong conversion => should this case be supported?

@see test case `shouldReturnLongEvenIfCustomConversionServiceDoesNotSupportLongConversion`

jzheaux:
@nenaraab Squashing helps with backpatches mostly, but yes it also helps a bit with the review process.

nenaraab:
@jzheaux 
In regard to your question with the  `DEFAULT_IDENTIFIER_AS_STRING`...

Before I'v started with the code change, i created the test to simulate the _issues_ i've seen within my application when not specifying any ConversionService.

Before of this pull request: this test case would have been green.
```java
   @Test
   public void shouldReturnStringWhenStringClassIdType() throws SQLException {
      // given
      String identifier = DEFAULT_IDENTIFIER_AS_STRING;
      // when
      Serializable newIdentifier = aclClassIdUtils.identifierFrom(identifier, resultSet);
   ...
   }
```
...as “999” can be transformed to Long 999L. But the one below not:

```java
   @Test
   public void shouldReturnStringWhenStringClassIdType() throws SQLException {
      // given
      String identifier = "MY_STRING_IDENTIFIER";
      // when
      Serializable newIdentifier = aclClassIdUtils.identifierFrom(identifier, resultSet);
      ....
   }
```
It gives me an `java.lang.NumberFormatException`: For input string: "MY_STRING_IDENTIFIER" because “MY_STRING_IDENTIFIER” can NOT be transformed to Long.

That's why i've decided to use a real "String" Identifier ...

nenaraab:
So, the commits are squashed... anything else to change :-)

jzheaux:
Thanks again, @nenaraab for the PR! This is now merged into `master` via 918a4cd3230c7dcd273a47ada5288c535ffcb030 where I [edited the commit message for clarity](https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.md#format-commit-messages).

PR #6151, opened 2018-11-28
AccountStatusUserDetailsChecker implements MessageSourceAware

### LINKED ISSUE #3980: AccountStatusUserDetailsChecker should implement MessageSourceAware
### Summary

We're using Spring Security on Google App Engine. The class `java.util.ResourceBundle$Control` is a restricted class on GAE, which makes our application fail when using Spring Security's `AbstractRememberMeServices`, which internally uses the `AccountStatusUserDetailsChecker`.
The `AccountStatusUserDetailsChecker` uses a `SpringSecurityMessageSource`, whose javadoc states that `all such classes [using SpringSecurityMessageSource] will also implement <code>MessageSourceAware</code>`.
### Actual Behavior

AccountStatusUserDetailsChecker uses SpringSecurityMessageSource, but doesn't implement MessageSourceAware.

Consequently, we cannot inject our own MessageSource into the default UserDetailsChecker, so that we would be forced to sub-class or implement a UserDetailsChecker with a GAE-compatible MessageSource.
### Expected Behavior

AccountStatusUserDetailsChecker should implement MessageSourceAware.

Alternatively it could help to configure the SpringSecurityMessageSource to use a custom MessageSource?
### Version

We're using Spring Security 4.0.3.RELEASE. The current master seems to have the same issue, though.

weebl2000:
This should fix #3980

### DIFF
--- a/core/src/main/java/org/springframework/security/authentication/AccountStatusUserDetailsChecker.java
+++ b/core/src/main/java/org/springframework/security/authentication/AccountStatusUserDetailsChecker.java
@@ -15,17 +15,20 @@
  */
 package org.springframework.security.authentication;
 
+import org.springframework.context.MessageSource;
+import org.springframework.context.MessageSourceAware;
 import org.springframework.security.core.SpringSecurityMessageSource;
 import org.springframework.security.core.userdetails.UserDetails;
 import org.springframework.security.core.userdetails.UserDetailsChecker;
 import org.springframework.context.support.MessageSourceAccessor;
+import org.springframework.util.Assert;
 
 /**
  * @author Luke Taylor
  */
-public class AccountStatusUserDetailsChecker implements UserDetailsChecker {
+public class AccountStatusUserDetailsChecker implements UserDetailsChecker, MessageSourceAware {
 
-	protected final MessageSourceAccessor messages = SpringSecurityMessageSource
+	protected MessageSourceAccessor messages = SpringSecurityMessageSource
 			.getAccessor();
 
 	public void check(UserDetails user) {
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@weebl2000 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=6151)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=6151) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@weebl2000 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=6151)!

core/src/main/java/org/springframework/security/authentication/AccountStatusUserDetailsChecker.java
```diff
@@ -51,4 +53,9 @@ public void check(UserDetails user) {
 					"User credentials have expired"));
 		}
 	}
+
+	@Override
```
rwinch: Please add Javadoc of `@since 5.2`

core/src/main/java/org/springframework/security/authentication/AccountStatusUserDetailsChecker.java
```diff
@@ -51,4 +53,9 @@ public void check(UserDetails user) {
...
+	public void setMessageSource(MessageSource messageSource) {
```
rwinch: Add `Assert.notNull(messageSource, "messageSource cannot be null");`

weebl2000:
Should be fixed

rwinch:
Thanks @weebl2000! 

Can you please update the commit message to align with [the Spring Security conventions](https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.md#format-commit-messages)? It is missing `Fixes: gh-3980` so something like...

```
Make AccountStatusUserDetailsChecker implement MessageSourceAware

Fixes: gh-3980
```

weebl2000:
Amended commit message

Issue #6187, opened 2018-11-29
Allow ISO Date encoded timestamp fields in JWT tokens.


greyfairer:
### Summary

Both Auth0 and OneLogin identity providers use ISO Date strings for the 'updated_at' claim in JWT Tokens, even if the spec says these should be numeric (unix timestamp). https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

If this happens, the StandardClaimAccessor#getUpdatedAt throws exceptions.

### Actual Behavior

If you use an OpenID Connector that writes the 'updated_at' claim in a JWT token in ISO Date format instead of a numerical unix timestamp, the getters on OidcUserInfo throw errors, so you cannot use e.g. jackson to serialize this user info.

### Expected Behavior

In general, getters should not throw exceptions. If the field is invalid, the constructor or the setter should throw an exception, or the field should be null. 

In this case, it would be nice to accept ISO Date formatted timestamps as valid values, since apparently this is used by at least two major OIDC vendors.

### Version

5.1.1.RELEASE

### Sample

```
@Controller
public class MyCollaboratorInfoController {
    @RequestMapping(value = "/myCollaboratorInfo.json")
    @ResponseBody
    public Principal currentUser(Principal principal) {
        return principal; //throws java.lang.IllegalArgumentException: Unable to convert claim 'updated_at' of type 'class java.lang.String' to Instant.
    }

    @RequestMapping(value = "/myCollaboratorInfo2.json")
    @ResponseBody
    public Map<String, Object> currentUser2(Principal principal) {
        return ((OAuth2AuthenticationToken)principal).getPrincipal().getAttributes(); //workaround that avoids the ClaimAccessor issue.
    }
```

greyfairer:
See https://github.com/spring-projects/spring-security/issues/5250#issuecomment-396439310
in #5250 by @sirianni

jgrandja:
@greyfairer Agreed...

> getters should not throw exceptions

I'll apply some improvements here.

jgrandja:
@greyfairer After giving this some further thought, we need to apply some more involved changes to `ClaimAccessor` to ensure it's more robust going forward. I have logged #6245 detailing the required changes. We're going to target this improvement for the 5.2.0 release.

Given this, I'm going to close this issue and associated PR as `ClaimAccessor` needs a bigger improvement compared to what is in the PR. I apologize for the back-and-forth and really appreciate your efforts and bringing up this issue so we can address it for the next release.

PR #6189, opened 2018-11-29
Add ClaimAccessor Support for ISO Date formatted timestamps

greyfairer:
…y Auth0 and OneLogin)

#6187

### DIFF
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
@@ -20,6 +20,8 @@ import org.springframework.util.Assert;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.time.Instant;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
 import java.util.Date;
@@ -101,15 +103,21 @@ public interface ClaimAccessor {
 		}
 		if (String.class.isAssignableFrom(claimValue.getClass())) {
 			try {
-				return Instant.parse((String) claimValue);
+				return Instant.from(DateTimeFormatter.ISO_INSTANT.parse((String)claimValue));
 			}
 			catch(DateTimeParseException e){
-				throw new IllegalArgumentException("Unable to convert claim '" + claim +
-														   "' from string '" + claimValue + "' to Instant.");
+				try {
+					return ZonedDateTime.from(DateTimeFormatter.ISO_ZONED_DATE_TIME.parse((String)claimValue))
+							.toInstant();
+				}
+				catch(DateTimeParseException e2){
+					throw new IllegalArgumentException("Unable to convert claim '" + claim +
+							"' from string '" + claimValue + "' to Instant.");
+				}
 			}
 		}
 		throw new IllegalArgumentException("Unable to convert claim '" + claim +
-												   "' of type '" + claimValue.getClass() + "' to Instant.");
+				"' of type '" + claimValue.getClass() + "' to Instant.");
 
 	}
 
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
@@ -20,6 +20,7 @@ import org.springframework.util.Assert;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.time.Instant;
+import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
@@ -98,8 +99,18 @@ public interface ClaimAccessor {
 		if (Instant.class.isAssignableFrom(claimValue.getClass())) {
 			return (Instant) claimValue;
 		}
+		if (String.class.isAssignableFrom(claimValue.getClass())) {
+			try {
+				return Instant.parse((String) claimValue);
+			}
+			catch(DateTimeParseException e){
+				throw new IllegalArgumentException("Unable to convert claim '" + claim +
+														   "' from string '" + claimValue + "' to Instant.");
+			}
+		}
 		throw new IllegalArgumentException("Unable to convert claim '" + claim +
-				"' of type '" + claimValue.getClass() + "' to Instant.");
+												   "' of type '" + claimValue.getClass() + "' to Instant.");
+
 	}
 
 	/**
--- a/oauth2/oauth2-core/src/test/java/org/springframework/security/oauth2/core/ClaimAccessorTests.java
+++ b/oauth2/oauth2-core/src/test/java/org/springframework/security/oauth2/core/ClaimAccessorTests.java
@@ -93,6 +93,17 @@ public class ClaimAccessorTests {
 				expectedClaimValue.minusSeconds(1), expectedClaimValue.plusSeconds(1));
 	}
 
+	// gh-6187
+	@Test
+	public void getClaimAsInstantWhenISODateStringThenReturnInstant() {
+		Instant expectedClaimValue = Instant.now();
+		String claimName = "isoDate";
+		this.claims.put(claimName, expectedClaimValue.toString());
+
+		assertThat(this.claimAccessor.getClaimAsInstant(claimName)).isBetween(
+				expectedClaimValue.minusSeconds(1), expectedClaimValue.plusSeconds(1));
+	}
+
 	// gh-5608
 	@Test
 	public void getClaimAsStringWhenValueIsNullThenReturnNull() {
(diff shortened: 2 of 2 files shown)


pivotal-issuemaster:
@greyfairer Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=6189)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=6189) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@greyfairer Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=6189)!

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
```diff
@@ -98,8 +99,18 @@ default Instant getClaimAsInstant(String claim) {
 		if (Instant.class.isAssignableFrom(claimValue.getClass())) {
 			return (Instant) claimValue;
 		}
+		if (String.class.isAssignableFrom(claimValue.getClass())) {
+			try {
+				return Instant.parse((String) claimValue);
+			}
+			catch(DateTimeParseException e){
+				throw new IllegalArgumentException("Unable to convert claim '" + claim +
```
greyfairer: We could return null, here. Or do a try-catch in the calling code.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
```diff
@@ -98,8 +99,18 @@ default Instant getClaimAsInstant(String claim) {
...
+														   "' from string '" + claimValue + "' to Instant.");
+			}
+		}
 		throw new IllegalArgumentException("Unable to convert claim '" + claim +
```
greyfairer: Same here: return null.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
```diff
@@ -98,8 +99,18 @@ default Instant getClaimAsInstant(String claim) {
 		if (Instant.class.isAssignableFrom(claimValue.getClass())) {
 			return (Instant) claimValue;
 		}
+		if (String.class.isAssignableFrom(claimValue.getClass())) {
+			try {
+				return Instant.parse((String) claimValue);
```
jgrandja: Instead of `Instant.parse()` let's attempt both of these:

```
ZonedDateTime.parse(claimValue, DateTimeFormatter.ISO_INSTANT)
ZonedDateTime.parse(claimValue, DateTimeFormatter.ISO_ZONED_DATE_TIME)
```

If both of these do not parse along with the previous (existing) parse attempts than we'll keep with throwing `IllegalArgumentException` as is the current behaviour. I know I said I agreed to not throw for getters, but this is a special case IMO. If the claim exists, which is the case if the parse attempt proceed than we cannot just return `null` as we are basically silently ignoring the error and the _existing_ claim. This will just cause bigger issues for the user as tracking the root issue will make things more difficult.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
jgrandja: Let's keep the existing behaviour of throwing `IllegalArgumentException` as mentioned in previous comment.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
```diff
@@ -98,8 +99,18 @@ default Instant getClaimAsInstant(String claim) {
...
+			}
+			catch(DateTimeParseException e){
```
jgrandja: Please ensure the code formatting is consistent with the current codebase.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
jgrandja: Same as comment above.

oauth2/oauth2-core/src/test/java/org/springframework/security/oauth2/core/ClaimAccessorTests.java
```diff
@@ -93,6 +93,17 @@ public void getClaimAsInstantWhenDoubleTypeSecondsThenReturnInstant() {
 				expectedClaimValue.minusSeconds(1), expectedClaimValue.plusSeconds(1));
 	}
 
+	// gh-6187
+	@Test
+	public void getClaimAsInstantWhenISODateStringThenReturnInstant() {
+		Instant expectedClaimValue = Instant.now();
```
jgrandja: Let's explicitly provide the `String` value for the test, for example, `2011-12-03T10:15:30+01:00`

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
greyfairer: ZonedDateTime.parse(claimValue, DateTimeFormatter.ISO_INSTANT) doesn't work, but  Instant.from(DateTimeFormatter.ISO_INSTANT.parse((String)claimValue)) does.

The second one: DateTimeFormatter.ISO_ZONED_DATE_TIME works for UTC timestamps too (i.e. ending with Z), so I'm not sure if doing the ISO_INSTANT first is necessary.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
```diff
@@ -98,8 +101,24 @@ default Instant getClaimAsInstant(String claim) {
 		if (Instant.class.isAssignableFrom(claimValue.getClass())) {
 			return (Instant) claimValue;
 		}
+		if (String.class.isAssignableFrom(claimValue.getClass())) {
+			try {
+				return Instant.from(DateTimeFormatter.ISO_INSTANT.parse((String)claimValue));
```
greyfairer: The second one: DateTimeFormatter.ISO_ZONED_DATE_TIME works for UTC timestamps too (i.e. ending with Z), so I'm not sure if doing the ISO_INSTANT first is necessary, but I'll keep it as per your request.

jgrandja: ok let's remove `DateTimeFormatter.ISO_INSTANT` and let `DateTimeFormatter.ISO_ZONED_DATE_TIME` handle both cases. Can you ensure we have 2 tests for:
- 2011-12-03T10:15:30Z
- 2011-12-03T10:15:30+01:00

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
```diff
@@ -98,8 +101,24 @@ default Instant getClaimAsInstant(String claim) {
...
+			}
+			catch(DateTimeParseException e){
+				try {
+					return ZonedDateTime.from(DateTimeFormatter.ISO_ZONED_DATE_TIME.parse((String)claimValue))
+							.toInstant();
+				}
+				catch(DateTimeParseException e2){
```
jgrandja: Can you ensure the code formatting is consistent with the class for try/catch - see `getClaimAsURL`

oauth2/oauth2-core/src/test/java/org/springframework/security/oauth2/core/ClaimAccessorTests.java
```diff
@@ -93,6 +93,31 @@ public void getClaimAsInstantWhenDoubleTypeSecondsThenReturnInstant() {
...
+		this.claims.put(claimName, "2011-12-03T10:15:30+01:00");
+		assertThat(this.claimAccessor.getClaimAsInstant(claimName)).isBetween(
+				expectedClaimValue.minusSeconds(1), expectedClaimValue.plusSeconds(1));
+		this.claims.put(claimName, "2011-12-03T10:15:30+01:00[Europe/Paris]");
+		assertThat(this.claimAccessor.getClaimAsInstant(claimName)).isBetween(
+				expectedClaimValue.minusSeconds(1), expectedClaimValue.plusSeconds(1));
+	}
+
+	// gh-6187
+	@Test(expected = IllegalArgumentException.class)
```
jgrandja: Please remove `expected = IllegalArgumentException.class` and use `assertThat(this.claimAccessor...` to be consistent with other tests in this class. Also, `assertThat(..)` is the preferred pattern for testing.

jgrandja:
Closing this PR. Please see [comment](https://github.com/spring-projects/spring-security/issues/6187#issuecomment-445012711)

Issue #6186, opened 2018-11-29
OAuth2ResourceServerSpec should allow its ServerBearerTokenAuthenticationConverter to be configurable


edeandrea:
Currently in `ServerHttpSecurity.OAuth2ResourceServerSpec.JwtSpec.configure(ServerHttpSecurity)` the `ServerBearerTokenAuthenticationConverter` is hard-coded in the `configure` method. The non-reactive side (`OAuth2ResourceServerConfigurer`) allows for the user to supply a `BearerTokenResolver`. The reactive side should allow for the same.

I propose adding similar functionality to `ServerHttpSecurity.OAuth2ResourceServerSpec` for this.

I'm happy to supply a PR for this if you would like.

jgrandja:
@edeandrea That would be great. Please do at your convenience.

edeandrea:
I'm almost wrapped up with it & wanted to ask a question regarding an implementation detail.

I'd like to change `ServerBearerTokenAuthenticationConverter` from a class to an interface and introduce a `DefaultServerBearerTokenAuthenticationConverter` class, mirroring what is there on the blocking side (`BearerTokenResolver` with a `DefaultBearerTokenResolver`). 

It's implementation would look like

```java
@FunctionalInterface
public interface ServerBearerTokenAuthenticationConverter extends ServerAuthenticationConverter {
	/**
	 * Resolve any <a href="https://tools.ietf.org/html/rfc6750#section-1.2" target="_blank">Bearer Token</a>
	 * value from the request
	 *
	 * @param request The {@link ServerHttpRequest}
	 * @return The token found in the request or {@code null} if none found
	 * @throws org.springframework.security.oauth2.core.OAuth2AuthenticationException If the found token is invalid
	 * @since 5.2
	 */
	@Nullable
	String resolveToken(ServerHttpRequest request);

	@Override
	default Mono<Authentication> convert(ServerWebExchange exchange) {
		return Mono.justOrEmpty(resolveToken(exchange.getRequest()))
			.map(BearerTokenAuthenticationToken::new);
	}
}
```

That would constitute a breaking change though. My thought on that though is that currently the `ServerBearerTokenAuthenticationConverter` is not "settable" anywhere in the configuration - so is a breaking change acceptable?

[...]

PR #6190, opened 2018-11-29
OAuth2ResourceServerSpec should allow its ServerAuthenticationConverter to be configurable

edeandrea:
Fixes gh-6186

### DIFF
--- a/config/src/test/java/org/springframework/security/config/web/server/OAuth2ResourceServerSpecTests.java
+++ b/config/src/test/java/org/springframework/security/config/web/server/OAuth2ResourceServerSpecTests.java
@@ -15,6 +15,15 @@
  */
 package org.springframework.security.config.web.server;
 
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.hamcrest.core.StringStartsWith.startsWith;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
 import java.io.IOException;
 import java.math.BigInteger;
 import java.security.KeyFactory;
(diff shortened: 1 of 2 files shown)


config/src/test/java/org/springframework/security/config/web/server/OAuth2ResourceServerSpecTests.java
```diff
@@ -429,6 +447,38 @@ ReactiveAuthenticationManager authenticationManager() {
...
 
+	@EnableWebFlux
+	@EnableWebFluxSecurity
+	static class CustomBearerTokenServerAuthenticationConverter {
+		@Bean
+		SecurityWebFilterChain springSecurity(ServerHttpSecurity http) throws Exception {
+			// @formatter:off
+			http
+				.authorizeExchange()
+					.anyExchange().access(authorizationManager())
```
jzheaux: Is there a reason you need this extra instrumentation, as opposed to something like 

```suggestion
					.anyExchange().access("SCOPE_message:read")
```

If we can, it's nice to be able to reduce the number of moving parts in a test.

edeandrea: You are correct I can switch this test. I will issue a new commit that changes this.

edeandrea:
@jzheaux I pushed a new commit and rebased so the commit message is better. Sorry about that!

edeandrea:
Also - any chance of getting this in pre 5.2?

jgrandja:
@edeandrea This can only get into 5.2 since [bearerTokenConverter(...)](https://github.com/spring-projects/spring-security/pull/6190/files#diff-f6e82f2a5486ce7fe3f6438343fc05f0R927) is new

edeandrea:
@jgrandja [bearerTokenConverter(...)](
https://github.com/spring-projects/spring-security/pull/6190/files#diff-f6e82f2a5486ce7fe3f6438343fc05f0R927) is what I added in the PR. There aren't any new classes or anything added in this PR - just a change to the configuration DSL.

jzheaux:
Thanks for the PR, @edeandrea! This is now merged into `master`.

PR #6191, opened 2018-11-29
Extract bearer token from arbitrary header

elefeint:
The new `HeaderBearerTokenResolver` extracts the bearer token (most likely JWT) from the application-defined HTTP request header.

This allows Spring Security Resource Server to understand, for example, [Google Cloud IAP proxy authentication](https://cloud.google.com/iap/docs/signed-headers-howto), which places JWT tokens in a `x-goog-iap-jwt-assertion` header.

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/HeaderBearerTokenResolver.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/HeaderBearerTokenResolver.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


rwinch:
I'm open to discussion on this, but despite it being simple I think this may be common enough that it would be valuable for us to provide an implementation.

elefeint:
@jzheaux Yep, the lambda is what I ended up doing for [GCP IAP authentication](https://github.com/spring-cloud/spring-cloud-gcp/blob/f9581d3a773b0375ab187a5e6b9a01d5a21fae5a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java#L93), but I was hoping for something more semantically precise, like this:
```	
public BearerTokenResolver iatTokenResolver(IapAuthenticationProperties properties) {
	return new HeaderBearerTokenResolver(properties.getHeader());
}
```

I understand the reasoning for not adding a trivial class, though.

Thanks for looking into this!

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/HeaderBearerTokenResolver.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2002-2018 the original author or authors.
```
jzheaux: Would you please update these to `2019`?

jzheaux:
Thanks so much, @elefeint! Last thing, will you please squash your commits?

elefeint:
@jzheaux Squashed, thank you!

jzheaux:
Thanks for your patience on this one @elefeint! This is now merged into `master`.

PR #6200, opened 2018-11-30
Restored Jacoco default task dependence

raphaelDL:
This commit ensures that the jacoco plugin is applied when calling
check and test tasks.
Also remoed the clean task that prevented sonarqube using coverage data

Fixes: gh-6199


rwinch:
Thanks this is now merged into master and backported via #6207 and #6208

Issue #6195, opened 2018-11-30
Bearer should be case-insensitive in ServerBearerTokenAuthenticationConverter


jgrandja:
Related #6150

nlebas:
Thank you @jgrandja for catching this, I'll need it. I'll submit a PR with the same fix as #6150

PR #6210, opened 2018-12-01
Accept a case-insensitive "Bearer" keyword

nlebas:
The Authorization header was matched for OAuth2
against the "Bearer" keyword in a case sensitive
fashion.
According to RFC 2617, it should be case insensitive
and some oauth clients (including some earlier
versions of spring-security) expect it so.

This is the reactive counterpart to commit
63f2b6094f59cc9ded6a83ac3def4a1726890a8b .

Fixes gh-6195


jgrandja:
Thanks for the PR @nlebas. This is now in master.

Issue #6212, opened 2018-12-03
Spring ACL uses deprecated anyListOf method


nenaraab:
### Summary
[Mockito Matchers](https://static.javadoc.io/org.mockito/mockito-core/2.23.4/org/mockito/Matchers.html) are deprecated and should be replaced with [Mockito ArgumentMatcher](https://static.javadoc.io/org.mockito/mockito-core/2.23.4/org/mockito/ArgumentMatchers.html).

Example:
[`public static <T> List<T> anyListOf(Class<T> clazz)`](https://static.javadoc.io/org.mockito/mockito-core/2.23.4/org/mockito/ArgumentMatchers.html#anyListOf-java.lang.Class-) is deprecated. With Java 8 this method will be removed in Mockito 3.0. This method is only used for generic friendliness to avoid casting, this is not anymore needed in Java 8.

Deprecated because Java 8 compiler can infer the type now. 

### Actual Implementation, e.g.

```
import static org.mockito.Matchers.anyListOf;
...

lookupStrategy.readAclsById(anyListOf(ObjectIdentity.class),
						anyListOf(Sid.class))).thenReturn(result);
```

### Expected Implementation, e.g.

```
import static org.mockito.ArgumentMatchers.anyList;
...

lookupStrategy.readAclsById(anyList(),
						anyList())).thenReturn(result);
```

### Version
LATEST

### Further details
Files affected: JdbcAclServiceTests and AclPermissionEvaluatorTests

### Further References
- https://stackoverflow.com/questions/10512374/mockito-list-matchers-with-generics

PR #6213, opened 2018-12-03
Spring ACL uses deprecated Mockito methods

nenaraab:
Replaces deprecated Mockito methods. Find further details in the following issue:
https://github.com/spring-projects/spring-security/issues/6212


nenaraab:
@rwinch 

I've updated the commit accordingly, thanks for the howto-contribute-link!

PR #6221, opened 2018-12-04
gh-5834: Use computeIfAbsent

### LINKED ISSUE #5834: SessionRegistryImpl should use computeIfAbsent
Java 8 introduced `computeIfAbsent` on `ConcurrentHashMap`:

> If the specified key is not already associated with a value,
> attempts to compute its value using the given mapping function
> and enters it into this map unless {@code null}.  The entire
> method invocation is performed atomically, so the function is
> applied at most once per key.

Since `SessionRegistryImpl` uses `ConcurrentHashMap` and since it uses it as `<String, List<?>>`, it seems like an ideal circumstance to make the update to `principals` atomic.

Instead of:

```java
Set<String> sessionsUsedByPrincipal = principals.get(principal);

if (sessionsUsedByPrincipal == null) {
	sessionsUsedByPrincipal = new CopyOnWriteArraySet<>();
	Set<String> prevSessionsUsedByPrincipal = principals.putIfAbsent(principal,
			sessionsUsedByPrincipal);
	if (prevSessionsUsedByPrincipal != null) {
		sessionsUsedByPrincipal = prevSessionsUsedByPrincipal;
	}
}

sessionsUsedByPrincipal.add(sessionId);
```

It could do:

```java
principals.computeIfAbsent(principal, key -> new CopyOnWriteArraySet<>())
	.add(sessionId);
```

rmartinus:
Fixes #5834 
Improve code by using Java 8's computeIfAbsent()


rwinch:
Thanks for the PR @rmartinus!

Can you please update the commit message to align with [the Spring Security conventions](https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.md#format-commit-messages)?

Something like:

```
SessionRegistryImpl uses computeIfAbsent

Fixes: gh-5834
```

rwinch:
Thanks @rmartinus! This is now merged into master and backported to 5.1.x and 5.0.x via #6232 #6233

Issue #6219, opened 2018-12-03
Integrate UserDetailsChecker into ReactiveAuthenticationManager & OAuth2 resource server implementations


edeandrea:
I propose adding a reactive version of `UserDetailsChecker` along with an implementing class that mirrors the `AccountStatusUserDetailsChecker`.

This is frequently used when combined with a `UserDetailsService` for checking the status of a user's account and we need this on the reactive side as well.

It also needs to be tied into the `ReactiveAuthenticationManager` implementations (i.e. `UserDetailsRepositoryReactiveAuthenticationManager`). I would say we open a separate ticket for that integration and we keep this ticket just for the introduction of the `ReactiveUserDetailsChecker`.

I'm happy to supply a PR for this as I already have it built and could submit it pretty quickly.

rwinch:
Thanks for the ticket @edeandrea!

I'm curious if we need a reactive version of `UserDetailsChecker` since the `UserDetails` is provided as an argument. Instead we could potentially use something like:

```java
Mono<UserDetails> ud = ...;

return ud.doOnSucces(ud -> userDetailsChecker.check(ud));
```

rwinch:
Sounds good. I'd stick with `.doOnNext` which is doing something with the `UserDetails` without needing to change its value.

edeandrea:
I'll rename this ticket then to something more appropriate for integrating `UserDetailsChecker` into the `ReactiveAuthenticationManager` implementations.

I'm happy to supply a PR for it as well.

edeandrea:
I opened #6229 for this.

PR #6229, opened 2018-12-04
Add Reactive Support for UserDetailsChecker

edeandrea:
Integrate UserDetailsChecker into ReactiveAuthenticationManager and
OAuth2 resource server authentication converters.

Fixes gh-6219

### DIFF
--- a/core/src/main/java/org/springframework/security/authentication/UserDetailsRepositoryReactiveAuthenticationManager.java
+++ b/core/src/main/java/org/springframework/security/authentication/UserDetailsRepositoryReactiveAuthenticationManager.java
@@ -16,16 +16,17 @@
 
 package org.springframework.security.authentication;
 
-import org.springframework.security.core.Authentication;
-
-import org.springframework.security.core.userdetails.ReactiveUserDetailsPasswordService;
-import org.springframework.security.core.userdetails.ReactiveUserDetailsService;
-import org.springframework.security.crypto.factory.PasswordEncoderFactories;
-import org.springframework.security.crypto.password.PasswordEncoder;
-import org.springframework.util.Assert;
 import reactor.core.publisher.Mono;
 import reactor.core.scheduler.Scheduler;
 import reactor.core.scheduler.Schedulers;
+
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.userdetails.ReactiveUserDetailsPasswordService;
+import org.springframework.security.core.userdetails.ReactiveUserDetailsService;
+import org.springframework.security.core.userdetails.UserDetailsChecker;
+import org.springframework.security.crypto.factory.PasswordEncoderFactories;
+import org.springframework.security.crypto.password.PasswordEncoder;
+import org.springframework.util.Assert;
 
 /**
  * A {@link ReactiveAuthenticationManager} that uses a {@link ReactiveUserDetailsService} to validate the provided
@@ -42,6 +43,8 @@ public class UserDetailsRepositoryReactiveAuthenticationManager implements React
 	private ReactiveUserDetailsPasswordService userDetailsPasswordService;
 
 	private Scheduler scheduler = Schedulers.parallel();
+
+	private UserDetailsChecker postAuthenticationChecks = userDetails -> {};
 
 	public UserDetailsRepositoryReactiveAuthenticationManager(ReactiveUserDetailsService userDetailsService) {
 		Assert.notNull(userDetailsService, "userDetailsService cannot be null");
@@ -65,6 +68,7 @@ public class UserDetailsRepositoryReactiveAuthenticationManager implements React
 					}
 					return Mono.just(u);
 				})
+				.doOnNext(this.postAuthenticationChecks::check)
 				.map(u -> new UsernamePasswordAuthenticationToken(u, u.getPassword(), u.getAuthorities()) );
 	}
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/springframework/security/authentication/UserDetailsRepositoryReactiveAuthenticationManager.java
```diff
@@ -43,6 +44,8 @@
 
 	private Scheduler scheduler = Schedulers.parallel();
 
+	private UserDetailsChecker userDetailsChecker;
```
rwinch: We should default this to just do nothing, so we don't need to worry about null values.

core/src/main/java/org/springframework/security/authentication/UserDetailsRepositoryReactiveAuthenticationManager.java
```diff
@@ -102,4 +110,16 @@ public void setUserDetailsPasswordService(
...
 	}
+
+	/**
+	 * Sets the strategy which will be used to validate the loaded <tt>UserDetails</tt>
+	 * object for the user.
+	 *
+	 * @param userDetailsChecker The {@link UserDetailsChecker}
+	 * @since 5.2
+	 */
+	public void setUserDetailsChecker(UserDetailsChecker userDetailsChecker) {
```
rwinch: We should expose this as `setPostAuthenticationChecks` to align with `AbstractUserDetailsAuthenticationProvider`. We should also update the member variable to `postAuthenticationChecks`

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtUserDetailsCheckingAuthenticationConverter.java
```diff
@@ -0,0 +1,94 @@
+/*
```
rwinch: Please remove this from the PR as it is something that should be considered separately.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/ReactiveJwtUserDetailsCheckingAuthenticationConverter.java
```diff
@@ -0,0 +1,95 @@
+/*
```
rwinch: Please remove this from the PR as it is something that should be considered separately.

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/authentication/JwtUserDetailsCheckingAuthenticationConverterTests.java
```diff
@@ -0,0 +1,238 @@
+/*
```
rwinch: Please remove this from the PR as it is something that should be considered separately.

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/authentication/ReactiveJwtUserDetailsCheckingAuthenticationConverterTests.java
```diff
@@ -0,0 +1,268 @@
+/*
```
rwinch: Please remove this from the PR as it is something that should be considered separately.

core/src/main/java/org/springframework/security/authentication/UserDetailsRepositoryReactiveAuthenticationManager.java
edeandrea: Do you mean build some kind of `NoOpUserDetailsChecker` and default to that? Or do something like this inside `UserDetailsRepositoryReactiveAuthenticationManager`

```java
private UserDetailsChecker userDetailsChecker = userDetails -> {};
```

rwinch: Just use a lambda

edeandrea:
I made the requested changes & rebased the commits into a single commit. I'll open a separate ticket for the `UserDetailsCheckingAuthenticationConverter`s.

edeandrea:
I opened #6237 for the other part of what I removed from this PR.

core/src/test/java/org/springframework/security/authentication/UserDetailsRepositoryReactiveAuthenticationManagerTests.java
```diff
@@ -56,6 +55,9 @@
 	@Mock
 	private Scheduler scheduler;
 
+	@Mock
+	private UserDetailsChecker userDetailsChecker;
```
rwinch: Can you please rename this to `postAuthenticationChecks` to be consistent?

core/src/test/java/org/springframework/security/authentication/UserDetailsRepositoryReactiveAuthenticationManagerTests.java
```diff
@@ -140,4 +142,19 @@ public void authenticateWhenPasswordServiceAndUpgradeFalseThenNotUpdated() {
...
+		this.manager.setPasswordEncoder(this.encoder);
+		this.manager.setPostAuthenticationChecks(this.userDetailsChecker);
+
+		assertThatExceptionOfType(LockedException.class)
+				.isThrownBy(() -> this.manager.authenticate(new UsernamePasswordAuthenticationToken(this.user, this.user.getPassword())).block())
+				.withMessage("account is locked");
+
+		verify(this.userDetailsChecker).check(eq(this.user));
+	}
 }
```
rwinch: Can you add another test that verifies that `postAuthenticationChecks` is not invoked if the `UserDetails` is not found?

core/src/test/java/org/springframework/security/authentication/UserDetailsRepositoryReactiveAuthenticationManagerTests.java
edeandrea: Done!

core/src/test/java/org/springframework/security/authentication/UserDetailsRepositoryReactiveAuthenticationManagerTests.java
edeandrea: Done!

edeandrea:
Added a new commit with requested changes.

edeandrea:
Added a new commit with requested changes.

edeandrea:
Done!

Issue #6224, opened 2018-12-04
MethodBasedEvaluationContext Should Delegate to MethodBasedEvaluationContext


rwinch:
### Summary

Spring Security's `MethodSecurityEvaluationContext` should delegate to Spring Framework's `MethodBasedEvaluationContext`

dbuos:
I'd like to take it ;)

rwinch:
Thanks @dbuos the issue is yours!

dbuos:
@rwinch I made a PR that solves this issue https://github.com/spring-projects/spring-security/pull/6249, I'll be happy to make any adjustments that may be necessary

PR #6249, opened 2018-12-07
Make MethodSecurityEvaluationContext Delegates to MethodBasedEvaluationContext

dbuos:
Spring Security's MethodSecurityEvaluationContext should delegate to Spring Framework's
MethodBasedEvaluationContext

Fixes: gh-6224

### DIFF
--- a/core/src/main/java/org/springframework/security/access/expression/method/MethodSecurityEvaluationContext.java
+++ b/core/src/main/java/org/springframework/security/access/expression/method/MethodSecurityEvaluationContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/springframework/security/access/expression/method/MethodSecurityEvaluationContext.java
```diff
@@ -15,105 +15,38 @@
...
-	private boolean argumentsAdded;
-
-	/**
-	 * Intended for testing. Don't use in practice as it creates a new parameter resolver
-	 * for each instance. Use the constructor which takes the resolver, as an argument
-	 * thus allowing for caching.
-	 */
-	public MethodSecurityEvaluationContext(Authentication user, MethodInvocation mi) {
-		this(user, mi, new DefaultSecurityParameterNameDiscoverer());
-	}
```
eleftherias: Is it necessary to remove this constructor? Although it is only intended for testing, removing it would not be backward compatible

dbuos: Hi @eleftherias I'll proceed to check the comment and rebase, thanks for your comment.

eleftherias: @dbuos Have you had the chance to review the requested changes?

eleftherias:
Merged via 150b66824df6d514c5430c45dc09edd91e6b2ee5
Polished via 924b9e95a1a2c9d5bd05528d6986e29ee921c930

PR #6258, opened 2018-12-07
setting paramName only when it is not null

shabari004:
#checking whether paramNames[i] is null, if it is not null then only we call StandardEvaluationContext.setVariable method


pivotal-issuemaster:
@shabari004 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=6258)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=6258) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@shabari004 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=6258)!

shabari004:
Hi @rwinch 
-  I updated the commit message and submit again (sorry about that).

- For creating test class, I created a test class namedMethodSecurityAddArgumentsTests. I added the file below, I'm not sure how much help i can ask, if you dont mind can you check it please
[MethodSecurityAddArgumentsTests.txt](https://github.com/spring-projects/spring-security/files/2692367/MethodSecurityAddArgumentsTests.txt)

rwinch:
Thanks for the feedback. It is probably best to just include the files in the PR. We can modify them as needed. That makes it easier to review and provide feedback.

For the tests:

* I'd start by creating a standard test named `MethodSecurityEvaluationContextTests`.
* Then I would create a static inner class named `NotNullVariableMethodSecurityEvaluationContext` that behaves exactly as you have defined `MethodSecurityAddArgumentsTests.setVariable`
* Then create a test that creates a `NotNullVariableMethodSecurityEvaluationContext` instance and injects a mock `ParameterNameDiscoverer` that returns a `new String[] { null }`. 
* Next invoke `NotNullVariableMethodSecurityEvaluationContext.lookupVariable` on the instance created in the previous step. Before your change this would then invoke the `setVariable` method with a null variable name. Since this test overrides the setVariable method to forbid a null value, if there is no Exception at this point, we know the code is fixed. You can validate that the test is real by temporarily removing the null check and see that it will fail.

shabari004:
I created the test cases as you mentioned. can you review it please. Thank you.

core/src/test/java/org/springframework/security/access/expression/method/MethodSecurityEvaluationContextTests.java
```diff
@@ -0,0 +1,81 @@
...
+		NotNullVariableMethodSecurityEvaluationContext NNVMSEC= new NotNullVariableMethodSecurityEvaluationContext(mock(Authentication.class), mock(MethodInvocation.class), paramNameDiscoverer);
+		try {
+			NNVMSEC.lookupVariable("TESTVALUE");
+		}
+		catch (IllegalArgumentException unexpected) {
+			fail("name  should not be null");
+
+		}
+	}
+
```
rwinch: Remove these extra new lines

rwinch:
Thanks for the updates. It appears we now have a few problems:

* There are test failures. You can run `./gradlew check` to find the errors
* The branch needs to be [rebased](https://help.github.com/articles/about-git-rebase/) off of master due to the conflicts
* The commits need to be [squashed](https://blog.github.com/2016-04-01-squash-your-commits/)

shabari004:
@rwinch i think i lost it, is there anyway somebody can work on this bug. I think i dont have enough knowledge, Thanks for guidance till this point.

ankurpathak:
> @rwinch i think i lost it, is there anyway somebody can work on this bug. I think i dont have enough knowledge, Thanks for guidance till this point.

@rwinch @shabari004 I would like to take it forward.

ankurpathak:
Here is new pull request for the issue:
https://github.com/spring-projects/spring-security/pull/6332

rwinch:
@shabari004 Thanks for following up. I will proceed with #6332

PR #6272, opened 2018-12-11
Minor Improvement In UsernamePasswordAuthenticationFilter

ankurpathak:
Improve UsernamePasswordAuthenticationFilter to extract
username and password from request by using Java 8
Optional.


rwinch:
Thanks for the PR @ankurpathak! However, I'm going to reject this as we generally prefer to avoid creating new objects (i.e. an `Optional`) unless they are necessary.

Issue #6270, opened 2018-12-11
Simplify Configuring Realm Name for WebFlux Applications


tensberg:
### Summary

I'd like to be able to change the HTTP BASIC Realm configured by the WebFlux `ServerHttpSecurity http.httpBasic()`. It looks to me like there is no way to get at the underlying `HttpBasicSpec.entryPoint` to change the default.

### Actual Behavior

`ServerHttpSecurity http.httpBasic()` always uses the default realm *Realm*.

### Expected Behavior

Add API `ServerHttpSecurity http.httpBasic().realmName("myRealm")` similar to `HttpSecurity` to change the realm.

### Configuration

```
@Bean
    public SecurityWebFilterChain springSecurityFilterChain(ServerHttpSecurity http) {
        http
            .authorizeExchange()
                .anyExchange().authenticated()
                .and()
            .httpBasic()
        return http.build();
    }
```

### Version

Spring Boot 2.1.1.RELEASE
Spring Security 5.1.2.RELEASE

rwinch:
@tensberg Thanks for the report! Would you be interested in submitting a PR for this?

ankurpathak:
@rwinch i would like to take this one.

ankurpathak:
@rwinch I was free and wanted something to code. So I completed this issue and providing a pull 
request. I won't mind even if dosen't get accepted and chance get passed to @tensberg .
https://github.com/spring-projects/spring-security/pull/6279

PR #6279, opened 2018-12-13
Allow setting realm for Http Basic

ankurpathak:
1. Added method realmName in ServerHttpSecurity to allow
specifying realm name.
2. Added test in ServerHttpSecurityTests to check if
specified realm name is returned.

Pull request for issue: https://github.com/spring-projects/spring-security/issues/6270

### DIFF
--- a/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
@@ -1879,6 +1879,19 @@ public class ServerHttpSecurity {
 		}
 
 		/**
+		 * Allows easily setting the entry point.
+		 * @param authenticationEntryPoint the {@link ServerAuthenticationEntryPoint} to use
+		 * @return {@link HttpBasicSpec} for additional customization
+		 * @since 5.2.0
+		 * @author Ankur Pathak
+		 */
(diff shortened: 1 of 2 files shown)


config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -1878,6 +1878,19 @@ public HttpBasicSpec securityContextRepository(ServerSecurityContextRepository s
...
 		}
 
+		/**
+		 * Allows easily changing the realm.
+		 * @param realmName the HTTP Basic realm to use
+		 * @return {@link HttpBasicSpec} for additional customization
+		 * @since 5.2.0
+		 * @author Ankur Pathak
+		 */
+		public HttpBasicSpec realmName(String realmName){
```
rwinch: I think that rather than specifying realmName we should allow injection of the entryPoint instead. This gives the user more flexibility and we don't need this additional logic which does an instance of check.

ankurpathak: I do agree, allowing setting entryPoint will give user more flexibility. So method to set realmName
removed in its favour. The new method authenticationEntryPoint addedin its place.

rwinch:
Thanks for the PR @ankurpathak! This is now merged into master

Issue #6256, opened 2018-12-07
Validate ClientRegistration.scopes


jgrandja:
We should validate `ClientRegistration.scopes` to ensure it contains valid characters as per [spec](https://tools.ietf.org/html/rfc6749#section-3.3).

Related [spring-boot#15398](https://github.com/spring-projects/spring-boot/issues/15398)

kabronkline:
With this fix I'm unable to supply resource as a scope parameter to ADFS and Azure STS. 

https://learn.microsoft.com/en-us/windows-server/identity/ad-fs/overview/ad-fs-openid-connect-oauth-flows-scenarios

scope = [resource url]//[scope values e.g., openid]

With this fix the '//' is rejected for having invalid characters.

https://learn.microsoft.com/en-us/azure/active-directory/develop/scopes-oidc

Notice scope "https://graph.microsoft.com/.default" which has invalid characters.

PR #6285, opened 2018-12-14
Validate Scopes in ClientRegistration.Builder

jzheaux:
Fixes: gh-6256


jgrandja:
Thanks for the PR @jzheaux. This is now in master.

Issue #6265, opened 2018-12-07
Remove Servlet 3.0 Support in CacheControlHeadersWriter


jzheaux:
Related to #6220 

`CacheControlHeadersWriter` checks first to see [if `HttpServletRequest#getHeader` is available](https://github.com/spring-projects/spring-security/blob/master/web/src/main/java/org/springframework/security/web/header/writers/CacheControlHeadersWriter.java#L56).

Since Spring Framework 5.x has a baseline of Servlet Spec 3.1, this check is no longer necessary.

shawnbiesan:
Hello, I submitted a PR for this. I think I followed the contributor rules correctly!

PR #6290, opened 2018-12-16
Remove check for method HttpServletRequest#getHeader and related test

ghost:
Fixes: gh-6265


shawnbiesan:
Not sure I understand the travis job. It says it's pending but when I look at details it seems to be in error? Can't find a message though

jzheaux:
@shawnbiesan Thank you! I think we just need one change, which is to amend the commit message. Ideally, the commit title is under 50 characters, but if you want to be more descriptive than that, then you can add extra information to the commit body.

This will also bump the Travis build, which I'm looking into now.

shawnbiesan:
OK, length is lessened (I assume force push is the way to go to amend to message? Haven't done a PR/fork on here before?). Looks like travis passed now too

PR #6293, opened 2018-12-17
Add web-flux jackson module and add check for javax.servlet package in classpath in web jackson module

### LINKED ISSUE #5623: Spring WebFlux doesn't mix when Jackson mixins are applied
Origin: https://github.com/spring-projects/spring-session-data-mongodb/issues/33

Problem: Spring WebFlux + Spring Security + Spring Session MongoDB + `JacksonMongoSessionConverter`.

`WebJackson2Module` registers multiple mixins including `CookieMixin`. This fails because `Cookie` is not on the classpath.

Possible solution: Conditionalize some of these registrations so they are only done when Spring MVC on the classpath.

```java
    public void setupModule(SetupContext context) {
        SecurityJackson2Modules.enableDefaultTyping((ObjectMapper)context.getOwner());
        context.setMixInAnnotations(Cookie.class, CookieMixin.class);
        context.setMixInAnnotations(SavedCookie.class, SavedCookieMixin.class);
        context.setMixInAnnotations(DefaultCsrfToken.class, DefaultCsrfTokenMixin.class);
        context.setMixInAnnotations(DefaultSavedRequest.class, DefaultSavedRequestMixin.class);
        context.setMixInAnnotations(WebAuthenticationDetails.class, WebAuthenticationDetailsMixin.class);
        context.setMixInAnnotations(PreAuthenticatedAuthenticationToken.class, PreAuthenticatedAuthenticationTokenMixin.class);
    }
```

finke-ba:
1. Add new module WebServerJackson2Module for spring-web-flux.
I made it because DefaultCsrfServerToken exists in two packages at the moment(web and web.server) but only one mixin is configured(for web module).
I don't know what what to add to WebServerJackson2Module except DefaultCsrfServerTokenMixin, but in my feb-flux  project that's enough.
2. Add check for javax.servlet package in classpath in WebJackson2Module. This also applies to web-flux, since there is no javax.servlet in web-flux module.

Fixes #5623

### DIFF
--- a/web/src/main/java/org/springframework/security/web/jackson2/WebJackson2Module.java
+++ b/web/src/main/java/org/springframework/security/web/jackson2/WebJackson2Module.java
@@ -26,14 +26,15 @@ import org.springframework.security.web.savedrequest.DefaultSavedRequest;
 import org.springframework.security.web.savedrequest.SavedCookie;
 
 import com.fasterxml.jackson.core.Version;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.module.SimpleModule;
+import org.springframework.util.ClassUtils;
 
 /**
  * Jackson module for spring-security-web. This module register {@link CookieMixin},
  * {@link DefaultCsrfTokenMixin}, {@link DefaultSavedRequestMixin} and {@link WebAuthenticationDetailsMixin}. If no
  * default typing enabled by default then it'll enable it because typing info is needed to properly serialize/deserialize objects.
  * In order to use this module just add this module into your ObjectMapper configuration.
+ * If there is no javax.servlet package in classpath all mixins that depends on it will be skipped.
  *
  * <pre>
  *     ObjectMapper mapper = new ObjectMapper();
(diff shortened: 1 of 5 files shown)


web/src/main/java/org/springframework/security/web/jackson2/WebJackson2Module.java
```diff
@@ -53,12 +54,14 @@ public WebJackson2Module() {
 
 	@Override
 	public void setupModule(SetupContext context) {
-		SecurityJackson2Modules.enableDefaultTyping((ObjectMapper) context.getOwner());
-		context.setMixInAnnotations(Cookie.class, CookieMixin.class);
-		context.setMixInAnnotations(SavedCookie.class, SavedCookieMixin.class);
+		SecurityJackson2Modules.enableDefaultTyping(context.getOwner());
+		if (ClassUtils.isPresent("javax.servlet.http.Cookie", this.getClass().getClassLoader())) {
+			context.setMixInAnnotations(Cookie.class, CookieMixin.class);
```
rwinch: Depending on the JDK this can cause issues. The reason is that many JDK's will load all the classes necessary for the entire class to run at once even if the code is not necessarily going to be executed. Instead, this code needs to be isolated in a separate module and then conditionally loaded.

finke-ba:
@rwinch thank you for quick feedback! 
Of cource we could split it. Should I create two new tickets and link them with #5623 or you will do it?
And after that should I create two pull requests for that tickets based on corresponding commits or what is the best option?

web/src/main/java/org/springframework/security/web/jackson2/WebJackson2Module.java
finke-ba: Hm, that's interesting, I didn't know that..
Is that means that some JDK(if it's fast could you give me example of that specific JDK or link to specification?) will try to load all the classes that presents in that class in initialization step of the that class? And that 'Cookie.class' will be try to load without looking at the condition?
But if I put everything in brackets into a separate class(new module) that could be load but will not instantiate, base on condition, that will work? 
And that loading of all necessary classes will done only if I will try to create instance of that new module?

finke-ba:
Created two issues - #6302, #6303

Issue #5760, opened 2018-09-02
adding query parameter to authorization_uri creates malformed url


mlevkovsky:
### Summary

When creating the authorization uri to login with google, there is the option to add a query parameter in order to get back the refresh token.
However, when the authorization_uri is set to: 

>https://accounts.google.com/o/oauth2/v2/auth?access_type=offline

The uri that I get redirect to is:
>https://accounts.google.com/o/oauth2/v2/auth?access_type=offline?response_type=code&client_id=[my client id]&scope=[scopes]&state=[state]&redirect_uri=[redirect uri]

Note the **?access_type=offlince?response_type...** 
This url is malformed and google complains saying response_type and basic query params are not passed in.

### Actual Behavior

1.  User goes to /login
2. User sees an error from Google due to malformed URL

### Expected Behavior

1. User goes to /login
2. User sees the google login page and the following URL in the address bar:
`https://accounts.google.com/o/oauth2/v2/auth?access_type=offline&response_type=code&client_id=[my client id]&scope=[scopes]&state=[state]&redirect_uri=[redirect uri]`
The access_type query parameter is after the ? and following query parameters should have an & between them.
The order of the query params does not matter.

### Configuration
My application.yaml
```yaml
spring:
  security:
    oauth2:
      client:
        registration:
          google:
            client-id: xxxxx
            client-secret: yyyyy
            scope: profile,email,https://www.googleapis.com/auth/analytics
        provider:
          google:
            authorization-uri: https://accounts.google.com/o/oauth2/v2/auth?access_type=offline
```

My WebSecurityConfigurationAdapter
```java
@Configuration
public class WebSecurityConfiguration extends WebSecurityConfigurerAdapter {

    @Override
    protected void configure(HttpSecurity http) throws Exception {
        http.authorizeRequests()
                .antMatchers("/login").permitAll()
                .anyRequest().authenticated()
                .and()
                .oauth2Login()
                    //.loginPage("/login")
                        .defaultSuccessUrl("/dashboard")
                        .failureUrl("/loginFailure")
                    .authorizationEndpoint()
                        .authorizationRequestRepository(authorizationRequestRepository())
                    .and()
                        .tokenEndpoint().accessTokenResponseClient(accessTokenResponseClient());
    }

    @Bean
    public AuthorizationRequestRepository<OAuth2AuthorizationRequest> authorizationRequestRepository() {
[...]

PR #6299, opened 2018-12-17
authorization_uri Supports Query Parameters

jzheaux:
Fixes: gh-5760

### DIFF
--- a/config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2ClientConfigurerTests.java
+++ b/config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2ClientConfigurerTests.java
@@ -135,7 +135,10 @@ public class OAuth2ClientConfigurerTests {
 		MvcResult mvcResult = this.mockMvc.perform(get("/oauth2/authorization/registration-1"))
 			.andExpect(status().is3xxRedirection())
 			.andReturn();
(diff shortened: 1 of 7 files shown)


jzheaux:
@jgrandja note that while `UriComponentsBuilder` can encode query parameters on its own, it does not encode the `redirect_uri` in the same way as `URLEncoder.encode`.

For that reason, the code still manually encodes the query parameters. However, let me know if it's okay to change the existing way `redirect_uri` is encoded. Specifically `:` and `/` are not encoded by `UriComponentsBuilder`.

rwinch:
@jzheaux Do we know why there is a difference? Is this something we should take up with Spring Framework team?

jzheaux:
@rwinch It appears to be related to [RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.4) (emphasis mine):

> The characters slash ("/") and question mark ("?") may represent data
within the query component.  Beware that some older, erroneous
implementations may not handle such data correctly when it is used as
the base URI for relative references (Section 5.1), apparently
because they fail to distinguish query data from path data when
looking for hierarchical separators.  However, as query components
are often used to carry identifying information in the form of
"key=value" pairs **and one frequently used value is a reference to
another URI, it is sometimes better for usability to avoid percent-
encoding those characters**.

jzheaux:
Given that `UriComponentsBuilder` likely has reasonable defaults and given this recommendation from the spec, I'd prefer to just rely on `UriComponentsBuilder` to do the encoding. Are there reasons the code needs to continue percent-encoding `\` and `:`?

jzheaux:
@rwinch @jgrandja I've updated this to rely on `UriComponentsBuilder` for encoding as well.

jzheaux:
@jgrandja yes, I considered that, but I prefer the explicit statement to use `UTF-8`, as you may have gathered.

Issue #6274, opened 2018-12-12
Give SecurityFilterChain a name for tracing!


zidom:
Give SecurityFilterChain a name for debug tracing!

zidom:
When there are multipleSecurityFilterChain, eachSecurityFilterChain has multiple filters, and these filters are the same or similar in most cases, during the development period, it is difficult to determine which SecurityFilterChain the request is, it is easy to confuse! Therefore, whether you can give the SecurityFilterChain a name, in the debug log with the corresponding name, I think this is very helpful for debugging and monitoring!

rwinch:
This sounds like a reasonable idea. I think we can use the beanName which is injected using `BeanNameAware` interface. Would you be interested in submitting a PR?

ankurpathak:
> This sounds like a reasonable idea. I think we can use the beanName which is injected using `BeanNameAware` interface. Would you be interested in submitting a PR?

I have implemented BeanNameAware in DefaultSecurityFilterChain. but it won't be helping much
because I think Spring doesn't register instances of it as Bean in BeanFactory. I debugged it after 
using artifacts with these changes and was getting the Bean Name as null in DefaultSecurityFilterChain.

Here is pull reuest with above changes:
https://github.com/spring-projects/spring-security/pull/6288

ankurpathak:
![screenshot from 2018-12-18 07-52-14](https://user-images.githubusercontent.com/6482301/50128213-2b870980-029a-11e9-9992-92bfb1fac47d.png)
@rwinch Please take a look at this scrrenshot its using Spring Boot 1.5 with Old Spring OAuth 2 Project.
It has four Spring Security Chains. The First two has Request Macher which has some path with them
so can be used in logging and Second two(created For OAuth) have Request Matcher have no path with them so can not be used in logging. So I thing name for them will help a lot.

rwinch:
I think we could add a id property to the request matcher that is then reflected in the toString. This ensures that we don't need to do instanceof checks on the DefaultSecurityFilterChain

ankurpathak:
@rwinch I think its a good idea as it will work with individual flters using RequestMatcher or any thing
else using Request Matcher. I have done changes in below pull request to accomodated it.
https://github.com/spring-projects/spring-security/pull/6308
Let me know if it requires any modifications or changes.

LivingLikeKrillin:
Hi, I'd like to work on this if it's still available.
[...]

PR #6308, opened 2018-12-19
Better logging support by id in RequestMatcher

ankurpathak:
1. Added default method getId in RequestMatcher and
marked it as FunctionalInterface
2. Created Class AbstractRequestMatcher with default
support for getId logging things
3. Implemented build in matchers using AbstractRequestMatcher
4. Changed toString methods of build in matchers to reflect
id.

Fixes: gh-6274

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/PermitAllSupport.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/PermitAllSupport.java
@@ -20,6 +20,7 @@ import javax.servlet.http.HttpServletRequest;
 import org.springframework.security.access.SecurityConfig;
 import org.springframework.security.config.annotation.web.HttpSecurityBuilder;
 import org.springframework.security.config.annotation.web.configurers.AbstractConfigAttributeRequestMatcherRegistry.UrlMapping;
+import org.springframework.security.web.util.matcher.AbstractRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher;
 
 /**
@@ -64,7 +65,7 @@ final class PermitAllSupport {
 		}
 	}
 
-	private final static class ExactUrlRequestMatcher implements RequestMatcher {
+	private final static class ExactUrlRequestMatcher extends AbstractRequestMatcher {
 		private String processUrl;
 
 		private ExactUrlRequestMatcher(String processUrl) {
@@ -89,7 +90,7 @@ final class PermitAllSupport {
 		@Override
 		public String toString() {
 			StringBuilder sb = new StringBuilder();
-			sb.append("ExactUrl [processUrl='").append(processUrl).append("']");
+			sb.append("ExactUrl [processUrl='").append(processUrl).append("', id=").append(getId()).append("]");
 			return sb.toString();
 		}
 	}
--- a/web/src/main/java/org/springframework/security/web/DefaultSecurityFilterChain.java
+++ b/web/src/main/java/org/springframework/security/web/DefaultSecurityFilterChain.java
@@ -15,13 +15,14 @@
  */
 package org.springframework.security.web;
 
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
-import org.springframework.security.web.util.matcher.RequestMatcher;
-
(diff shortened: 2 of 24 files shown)


rwinch:
Thanks for the PR @ankurpathak!

After some more thought I think we need to take a more holistic approach to logging improvements. I created gh-6311 for us to improve logging within Spring Security as a whole. We will keep this ticket open and potentially merge these changes depending on the findings of that ticket

config/src/main/java/org/springframework/security/config/annotation/web/configurers/PermitAllSupport.java
```diff
@@ -89,11 +90,11 @@ public boolean matches(HttpServletRequest request) {
 		@Override
 		public String toString() {
 			StringBuilder sb = new StringBuilder();
-			sb.append("ExactUrl [processUrl='").append(processUrl).append("']");
+			sb.append("ExactUrl [processUrl='").append(processUrl).append("', id=").append(getId()).append("]");
```
dkapil: I think we can live away with String builder now, by-default string concatenations internally invoke string builder only. Plus it will result in cleaner code

config/src/main/java/org/springframework/security/config/annotation/web/configurers/PermitAllSupport.java
```diff
@@ -89,11 +90,11 @@ public boolean matches(HttpServletRequest request) {
...
 			return sb.toString();
 		}
 	}
 
 	private PermitAllSupport() {
 	}
-}
\ No newline at end of file
+}
```
dkapil: let's use the existing formatter so this does not occur

web/src/main/java/org/springframework/security/web/DefaultSecurityFilterChain.java
```diff
@@ -15,25 +15,27 @@
...
+
 /**
  * Standard implementation of {@code SecurityFilterChain}.
  *
  * @author Luke Taylor
  *
  * @since 3.1
  */
-public final class DefaultSecurityFilterChain implements SecurityFilterChain {
+public final class DefaultSecurityFilterChain implements SecurityFilterChain{
```
dkapil: let's use the existing formatter so this does not occur

web/src/main/java/org/springframework/security/web/DefaultSecurityFilterChain.java
```diff
@@ -61,4 +63,5 @@ public boolean matches(HttpServletRequest request) {
 	public String toString() {
 		return "[ " + requestMatcher + ", " + filters + "]";
 	}
+
```
dkapil: let's use the existing formatter so this does not occur

web/src/main/java/org/springframework/security/web/util/matcher/AndRequestMatcher.java
```diff
@@ -76,6 +75,6 @@ public boolean matches(HttpServletRequest request) {
 
 	@Override
 	public String toString() {
-		return "AndRequestMatcher [requestMatchers=" + requestMatchers + "]";
+		return "AndRequestMatcher [requestMatchers=" + requestMatchers + ", id=" + getId()+"]";
```
dkapil: is code formatted correctly?

web/src/main/java/org/springframework/security/web/DefaultSecurityFilterChain.java
ankurpathak: What do you mean by existing formatter??

dkapil: I meant, these whitespaces means that the code formatter we have used for this PR does not match the one used originally. Formatter can also mean any settings in your IDE that indents/formats/structures code. Please see the section on whitespaces in contributing guidelines

jgrandja:
@ankurpathak I'm going to close this PR as it's quite old and would need to be rebased off main. As well, there are a few logging improvement commits you'll notice linked in gh-6311.

If you feel the updates provided in this PR are still required, please submit a new PR rebased off main and align with the logging improvements already merged and referenced in gh-6311.

Issue #4399, opened 2017-06-16
Make JdbcUserDetailsManager to be able to handle UserDetails'es: nonLocked, nonExpired, credentialsNonExpired


paul-ovchinnikov:
### Summary
```JdbcUserDetailsManager``` has excellent native implementation CRUD and utility methods to handle User state (groups, authorities, roles and so on)

```org.springframework.security.core.userdetails.User``` and ```org.springframework.security.core.userdetails.UserDetails``` also determines additional states:
* accountNonExpired
* accountNonLocked
* credentialsNonExpired

### Actual Behavior

At the moment ```org.springframework.security.provisioning.JdbcUserDetailsManager``` and ```org.springframework.security.core.userdetails.jdbc.JdbcDaoImpl``` only can handle *enabled* state of the ```User```

On the other hand oauth2 Spring implementation is able to handle this states already.
But security framework returns default hardcoded values for *accountNonExpired, accountNonLocked, credentialsNonExpired (true, true, true)*

### Expected Behavior
Please extend ```org.springframework.security.provisioning.JdbcUserDetailsManager``` and ```org.springframework.security.core.userdetails.jdbc.JdbcDaoImpl``` to support all ```UserDetails``` and ```User``` states.
And make private methods to be public:
* private void insertUserAuthorities(UserDetails user)
* private void deleteUserAuthorities(String username)

### Version
Spring Framework:  4.3.8.RELEASE
Spring Security Framework:  4.2.3.RELEASE

avaud:
We run into this issue at work, it's problematic for us and it seems quite simple to fix, is there any reason it was never addressed ?
Would a pull request fixing it be accepted ?

Thanks !

123raoul123:
Same issue here. I went around it by extending the JdbcUserDetailsManager. However then I run into issues with private validation functions: private void validateUserDetails(UserDetails user) and private void validateAuthorities(Collection<? extends GrantedAuthority> authorities). I would rather see them protected.

rwinch:
Thanks for the ping @123raoul123 If someone wants to send a PR (the changes must be passive) for this, we'd gladly accept it.

pvliss:
Hello @rwinch, I would like to take this one

rwinch:
Thank you for volunteering @pvliss! The issue is all yours. Please let us know if you have any questions!

pvliss:
Hello @rwinch,

[...]

PR #6309, opened 2018-12-19
JdbcUserDetailsManager handles extra UserDetails attributes

pvliss:
Check `ResutSetMetaData` to see if extra columns are present in order to also handle the `UserDetails` attributes: `accountNonExpired`, `accountNonLocked` and `credentialsNonExpired`.

Fixes gh-4399

### DIFF
--- a/core/src/main/java/org/springframework/security/provisioning/JdbcUserDetailsManager.java
+++ b/core/src/main/java/org/springframework/security/provisioning/JdbcUserDetailsManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/springframework/security/provisioning/JdbcUserDetailsManager.java
```diff
@@ -179,7 +230,7 @@ public void setValues(PreparedStatement ps) throws SQLException {
 		userCache.removeUserFromCache(user.getUsername());
 	}
 
-	private void insertUserAuthorities(UserDetails user) {
+	public void insertUserAuthorities(UserDetails user) {
```
rwinch: Can you help me understand why you changed the scope of this method (along with the other methods listed in the PR)? It does not seem to align with the goal of the ticket to me.

rwinch: Thanks for bringing that to my attention. My understanding was that they wanted to work around the limitation by extending those methods, but with these changes they should not need to.

In any event, let's keep the methods private for now. Determining if we should change the scope can be done as a separate ticket. Thanks again for all your help!

pvliss:
Hello @rwinch.

API changes have now been removed.

rwinch:
@pvliss Thank you for all your work on this! I'm very pleased with how this turned out :) This is now merged into master.

We hope to see more contributions from you! If you are looking for something to contribute a good place to look is the [Help Wanted](https://github.com/spring-projects/spring-security/labels/help%20wanted) issues which have been marked because they are typically a bit easier. If you want one of the issues, please comment on the issue to claim it so as not to duplicate efforts of others. Of course, if you find any issue you want to work on, don't hesitate to ping us via the comments. In any case, we are always happy to help you with a contribution. Thank you again for your contribution to the community!

pvliss:
@rwinch Thanks for all your guidance through my first PR!!!

Will certainly try to help as much as I can.

Issue #6281, opened 2018-12-13
AuthenticationFailureBadCredentialsEvent published twice


mptardy:
### Summary

`AuthenticationFailureBadCredentialsEvent` gets published twice with due the fix of #6009, WebSecurityConfigurerAdapter.java:203.

### Actual Behavior

If you create a `ApplicationListener<AuthenticationFailureBadCredentialsEvent>` and listen to `AuthenticationFailureBadCredentialsEvent`, you get notified twice when the users provides wrong credentials.

### Expected Behavior

Same as `AuthenticationSuccessEvent`, the `AuthenticationFailureBadCredentialsEvent` should get published only once.

### Configuration

Can be reproduced if you use spring-boot-samples/spring-boot-sample-web-secure-custom and add an `ApplicationListener<AuthenticationFailureBadCredentialsEvent>`.

### Version

Spring Security 5.1.2.RELEASE

### Sample

Take spring-boot-samples/spring-boot-sample-web-secure-custom and add an `ApplicationListener<AuthenticationFailureBadCredentialsEvent>`.

```
@Component
protected static class LoginAttemptAuthenticationFailureEventListener implements ApplicationListener<AuthenticationFailureBadCredentialsEvent> {
	@Override
	public void onApplicationEvent(AuthenticationFailureBadCredentialsEvent event) {
		System.out.println(event.toString());
	}
}
```

[spring-boot-sample-web-secure-custom.zip](https://github.com/spring-projects/spring-security/files/2676027/spring-boot-sample-web-secure-custom.zip)

clevertension:
we should discussion how to publish the failure event between Parent and Child ProviderManager authentication

Parent throw AuthenticationFailureCredentialsExpiredEvent, Child throw AuthenticationFailureBadCredentialsEvent, do we need to publish both the two failure exception? i think we should to do so

so i think it is reasonable to publish twice AuthenticationFailureBadCredentialsEvent here, because they are authenticated from two **AuthenticationManager**, it is just a coincidence that the two AuthenticationManager throw the same Failure Exception

clevertension:
![](https://farm5.staticflickr.com/4887/32493729628_9a14637a09_o.png)

in the above scenario, we have 4 provider manger which is linked as parent and child, and it will publish both success event and failure event, but how to define its behavior of event publish?

clevertension:
[...]

PR #6313, opened 2018-12-20
fix AuthenticationFailureBadCredentialsEvent published twice

clevertension:
#6281 
this pr fix the issue of AuthenticationFailureBadCredentialsEvent published twice

### DIFF
--- a/core/src/main/java/org/springframework/security/authentication/ProviderManager.java
+++ b/core/src/main/java/org/springframework/security/authentication/ProviderManager.java
@@ -15,9 +15,6 @@
  */
 package org.springframework.security.authentication;
 
-import java.util.Collections;
-import java.util.List;
-
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.springframework.beans.factory.InitializingBean;
@@ -29,6 +26,12 @@ import org.springframework.security.core.AuthenticationException;
 import org.springframework.security.core.CredentialsContainer;
 import org.springframework.security.core.SpringSecurityMessageSource;
 import org.springframework.util.Assert;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
 
 /**
  * Iterates an {@link Authentication} request through a list of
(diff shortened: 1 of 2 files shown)


jgrandja:
@clevertension Thanks for the PR. However, I already have a [fix](https://github.com/jgrandja/spring-security/commit/22af1b742c88cdef7e8020ddf007702ad9d6ae7b) for this and pushed it to my fork.

I'm holding off on merging this until I get a response from you from my [comment](https://github.com/spring-projects/spring-security/issues/6281#issuecomment-448711363).

As an FYI, it's best to see if an issue is assigned before you start work on it so you don't waste any of your time. This issue was already assigned to me. I would have been more than happy to have you work on it if you initially requested it. However, we still need to have the discussion first to ensure we are on the same page. Can you please respond to my comment so I can better understand your scenario. Thanks.

clevertension:
@jgrandja , 
1. thanks for your remind, i will apply for the issue before i plan to work on it at next time
2. after i review and test your code carefully, i think you may fail to pass the **authenticationExceptionFromParentOverridesPreviousOnes** test case in **ProviderManagerTests**, so i think it is not so easy to fix this issue

3. yes,discussion is very important, i will give a demo project ASAP,

jgrandja:
@clevertension I reviewed your PR and your changes **do not** allow for duplicate failure events to be published. Although this is the behaviour we want in order to resolve #6281, this is not what triggered your initial [comment](https://github.com/spring-projects/spring-security/issues/6281#issuecomment-448154352):

> so i think it is reasonable to publish twice AuthenticationFailureBadCredentialsEvent here, because they are authenticated from two AuthenticationManager, it is just a coincidence that the two AuthenticationManager throw the same Failure Exception

So I guess I'm confused at this point as you originally argued that the `AuthenticationFailureBadCredentialsEvent` resulting from `BadCredentialsException` in the child and parent `AuthenticationManager` should both get published. However, the changes in your PR does not allow for duplicated events to be published.

As already mentioned in this [comment](https://github.com/spring-projects/spring-security/pull/6313#issuecomment-449032331), I have a fix for #6281 so I'm going to close this PR.

It is still not clear to me if you are facing a specific issue? If you are, please provide a sample that reproduces the issue so I can better understand your use case. Also, please log a new ticket providing the details and link to this issue. Thanks.

clevertension:
yes, i meet with this issue #6324 , we can talk to there

Issue #6325, opened 2018-12-22
Typo in Spring Security spring-security-x.y.rnc Files


ankurpathak:
Spring Security spring-security-x.y.rnc Files has typo 'amount' written as 'ammount'. This issue
is created to provide fix to that typo. This type also affects corresponding also affects corresponding xsd files.
![screenshot from 2018-12-22 18-50-48](https://user-images.githubusercontent.com/6482301/50374909-0fc58f80-061b-11e9-85d1-b2c09cc1c1db.png)

ankurpathak:
Pull request for issue:
https://github.com/spring-projects/spring-security/pull/6326

PR #6326, opened 2018-12-22
Fixes typo in x,rnc files

ankurpathak:
1. Fixes type ammount to amount in *.rnc files
2. Regenerates *.xsd files from *.rnc files

Fixes: gh-6325


jzheaux:
Thanks, again @ankurpathak!

Issue #6223, opened 2018-12-04
MethodSecurityEvaluationContext Should Check Parameter Names Are Not Null


rwinch:
### Summary

Spring Security's [MethodSecurityEvaluationContext](https://github.com/spring-projects/spring-security/blob/5.1.2.RELEASE/core/src/main/java/org/springframework/security/access/expression/method/MethodSecurityEvaluationContext.java#L115) should ignore parameter names that are `null`.

Related: https://jira.spring.io/browse/SPR-17565

shraiysh:
Hi @rwinch, I would like to take this one.

rwinch:
Thanks for the offer. I'd like to keep this for a first time contributor for the moment. If you are looking for more issues, please check out the help wanted and exclude the first timers only label. Here is the query if you need it https://github.com/spring-projects/spring-security/issues?utf8=%E2%9C%93&q=is%3Aopen+label%3A%22help+wanted%22+sort%3Aupdated-desc++-label%3Afirst-timers-only

shabari004:
Hi can i work on this

rwinch:
Thank you for volunteering to work on this @shabari004! The issue is yours. Please let me know if you need any help.

shabari004:
@rwinch i dont see any test cases for this class, do we need any test cases for this

shraiysh:
Thanks @rwinch.

rwinch:
> @rwinch i dont see any test cases for this class, do we need any test cases for this

@shabari004 There are not currently unit tests as this is tested in other places. However, I think it would make sense for this scenario to create unit tests.

shabari004:
@rwinch sorry to bother you, can i get some help in creating the test cases please.

rwinch:
@shabari004 You may. What help do you need? Can you describe the problem you are having? Can you post a branch with the code you have?

shabari004:
@rwinch Hi i created a pull request with my code. Thanks

PR #6332, opened 2018-12-24
fixes setting paramName only when it is not null

ankurpathak:
Fixes: gh-6223

### DIFF
--- a/core/src/main/java/org/springframework/security/access/expression/method/MethodSecurityEvaluationContext.java
+++ b/core/src/main/java/org/springframework/security/access/expression/method/MethodSecurityEvaluationContext.java
@@ -112,7 +112,9 @@ class MethodSecurityEvaluationContext extends StandardEvaluationContext {
 		}
 
 		for (int i = 0; i < args.length; i++) {
-			super.setVariable(paramNames[i], args[i]);
+			if (paramNames[i] != null) {
+				setVariable(paramNames[i], args[i]);
+			}
 		}
 	}
 
--- a/core/src/test/java/org/springframework/security/access/expression/method/MethodSecurityEvaluationContextTests.java
+++ b/core/src/test/java/org/springframework/security/access/expression/method/MethodSecurityEvaluationContextTests.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2002-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.access.expression.method;
+
+import java.lang.reflect.Method;
+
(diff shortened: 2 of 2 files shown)


core/src/test/java/org/springframework/security/access/expression/method/MethodSecurityEvaluationContextTests.java
```diff
@@ -0,0 +1,78 @@
...
+			extends MethodSecurityEvaluationContext {
+
+		public NotNullVariableMethodSecurityEvaluationContext(Authentication auth, MethodInvocation mi,
+				ParameterNameDiscoverer parameterNameDiscoverer) {
+			super(auth, mi, parameterNameDiscoverer);
+		}
+
+		@Override
+		public void setVariable(String name, @Nullable Object value) {
+			if (name == null)
```
rwinch: Please add `{` to follow Spring Security's code conventions

core/src/main/java/org/springframework/security/access/expression/method/MethodSecurityEvaluationContext.java
```diff
@@ -112,7 +112,9 @@ private void addArgumentsAsVariables() {
 		}
 
 		for (int i = 0; i < args.length; i++) {
-			super.setVariable(paramNames[i], args[i]);
+			if (paramNames[i] != null) {
+				super.setVariable(paramNames[i], args[i]);
```
rwinch: For the test below to work, you need to remove `super.`

core/src/test/java/org/springframework/security/access/expression/method/MethodSecurityEvaluationContextTests.java
```diff
@@ -0,0 +1,78 @@
...
+	private ParameterNameDiscoverer paramNameDiscoverer;
+	@Mock
+	private Authentication authentication;
+	@Mock
+	private MethodInvocation methodInvocation;
+	@Mock
+	private AuthenticationTrustResolver trustResolver;
+
+	@Test
+	public void setVariableTest() {
```
rwinch: please rename to `lookupVariableWhenParameterNameNullThenNotSet`

core/src/test/java/org/springframework/security/access/expression/method/MethodSecurityEvaluationContextTests.java
ankurpathak: Added { to if part.

ankurpathak: Do we need to add it for else part as well??

core/src/main/java/org/springframework/security/access/expression/method/MethodSecurityEvaluationContext.java
ankurpathak: Removed super.

core/src/test/java/org/springframework/security/access/expression/method/MethodSecurityEvaluationContextTests.java
ankurpathak: renamed test to lookupVariableWhenParameterNameNullThenNotSet

core/src/test/java/org/springframework/security/access/expression/method/MethodSecurityEvaluationContextTests.java
rwinch: Yes. We do not use implied `{` for one line statements

ankurpathak:
changed commit message to reflect Fixes: gh-6223.

rwinch:
Thanks @ankurpathak! This is now merged into master

PR #6336, opened 2018-12-27
X509 Reactive Support 

alek-sys:


web/src/main/java/org/springframework/security/web/server/authentication/X509AuthenticationWebFilter.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.web.server.authentication;
+
+import org.springframework.security.authentication.ReactiveAuthenticationManager;
+import org.springframework.security.web.authentication.preauth.x509.X509PrincipalExtractor;
+
+public class X509AuthenticationWebFilter extends AuthenticationWebFilter {
```
rwinch: Let's remove this as we don't extend AuthenticationWebFilter unless it is necessary (we prefer composition to inheritance)

alek-sys: Thanks for your feedback, @rwinch ! I've removed the class now, although testing that filter was added is a bit tricky now and uses `TestReflectionUtils` to check `authenticationConverter` field. It is not ideal, but with a combination of integration test and convertes tests should provide good enough level of coverage.

samples/boot/webflux-x509/src/main/java/sample/WebfluxX509Application.java
```diff
@@ -0,0 +1,42 @@
...
+import org.springframework.security.config.web.server.ServerHttpSecurity;
+import org.springframework.security.web.server.SecurityWebFilterChain;
+import reactor.core.publisher.Mono;
+
+@SpringBootApplication
+public class WebfluxX509Application {
+
+	@Bean
+	public SecurityWebFilterChain securityWebFilterChain(ServerHttpSecurity http) {
+		return http
```
rwinch: Please fix this [indentation](https://spring.io/blog/2013/07/11/spring-security-java-config-preview-readability/)

```java
// @formatter:off
http
    .x509()
        .authenticationManager(Mono::just)
	    .and()
	.authorizeExchange()
        .anyExchange().permitAll();
// @formatter:on
return http.build();
```

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -572,6 +577,68 @@ public FormLoginSpec formLogin() {
...
+	 * @return the {@link X509Spec} to customize
+	 */
+	public X509Spec x509() {
+		if (this.x509 == null) {
+			this.x509 = new X509Spec();
+		}
+
+		return this.x509;
+	}
+
```
rwinch: Please add `@since 5.2` and `@author`

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -572,6 +577,68 @@ public FormLoginSpec formLogin() {
...
+	 *              .principalExtractor(principalExtractor);
+	 *      return http.build();
+	 *  }
+	 * </pre>
+	 *
+	 * Note that if extractor is not specified, {@link SubjectDnX509PrincipalExtractor} will be used.
+	 * If authenticationManager is not specified, the default one from {@code http.authenticationManager} will be used.
+	 *
+	 * @return the {@link X509Spec} to customize
+	 */
```
rwinch: Please add `@since 5.2` and `@author`

samples/boot/webflux-x509/src/main/java/sample/MeController.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.springframework.security.core.context.ReactiveSecurityContextHolder;
+import org.springframework.security.core.context.SecurityContext;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+import reactor.core.publisher.Mono;
+
+@RestController
+@RequestMapping("/me")
```
rwinch: Please add `@since 5.2` and `@author`

samples/boot/webflux-x509/src/main/java/sample/WebfluxX509Application.java
```diff
@@ -0,0 +1,42 @@
...
+package sample;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.context.annotation.Bean;
+import org.springframework.security.config.web.server.ServerHttpSecurity;
+import org.springframework.security.web.server.SecurityWebFilterChain;
+import reactor.core.publisher.Mono;
+
+@SpringBootApplication
```
rwinch: Please add `@since 5.2` and `@author`

web/src/main/java/org/springframework/security/web/server/authentication/ServerX509AuthenticationConverter.java
```diff
@@ -0,0 +1,72 @@
...
+import org.springframework.security.web.authentication.preauth.x509.X509PrincipalExtractor;
+import org.springframework.web.server.ServerWebExchange;
+import reactor.core.publisher.Mono;
+
+import java.security.cert.X509Certificate;
+
+/**
+ * Converts from a {@link SslInfo} provided by a request to an {@link PreAuthenticatedAuthenticationToken} that can be authenticated.
+ *
+ * @author Alexey Nesterov
```
rwinch: Please add `@since 5.2`

alek-sys:
Hey @rwinch, sorry for not coming back on this, I'm looking into this and hope to get it updated today or tomorrow.

config/src/test/java/org/springframework/security/config/web/server/ServerHttpSecurityTests.java
```diff
@@ -279,6 +283,33 @@ public void basicWithCustomRealmName() {
 		assertThat(result.getResponseCookies().getFirst("SESSION")).isNull();
 	}
```
rwinch: I'd like one more test that verifies that everything works with just `http.x509()` (the `principalExtractor` and `authenticationManager` should be optional). If any code changes are necessary for the test to pass make those too.

docs/manual/src/docs/asciidoc/_includes/reactive/x509.adoc
```diff
@@ -0,0 +1,24 @@
+[[reactive-x509]]
+= Reactive X.509 Authentication
+
+Similar to <<x509,Servlet X.509 authentication>>, reactive x509 authentication filter allows to extract an authentication token from a certificate provided by a client.
+
```
rwinch: I'd like a minimal example configuration shown first with just `.x509()` then document all the properties (i.e. adding `authenticationManager` and `principalExtractor` in a separate configuration). This allows users to not only see the minimum requirements but then see all the options.

alek-sys:
I bet you knew that! 👍 

Yes, there is an issue with using default authentication manager, even more of that, I've discovered that default authentication manager won't work, because `PreAuthenticatedAuthenticationToken` created by the converted will not be authorized. Also, to get feature parity with Servlet X509 I'd like to add `PreAuthenticatedAuthenticationProvider` support as well. So I keep working on this one.

docs/manual/src/docs/asciidoc/_includes/reactive/x509.adoc
alek-sys: I've added another example of x509 configuration to demonstrate how to change `authenticationManager` and `principalExtractor` and described how defaults will work.

config/src/test/java/org/springframework/security/config/web/server/ServerHttpSecurityTests.java
alek-sys: That was indeed a very good point. Making default configuration to work forced me to think about the whole solution a little bit more. So to get feature parity with Servlet x509 I've added `ReactivePreAuthenticatedAuthenticationManager` which fetches user details using `ReactiveUserDetailsService` and checks account status. Account status check is delegated to  `AccountStatusUserDetailsChecker` using `ReactiveDelegatingUserDetailsChecker` to bridge reactive world with blocking. 

I've also added `ReactiveUserDetailsChecker`interface so developers can implement a proper non-blocking account check if they want to.

core/src/main/java/org/springframework/security/core/userdetails/ReactiveUserDetailsChecker.java
```diff
@@ -0,0 +1,29 @@
...
+import reactor.core.publisher.Mono;
+
+/**
+ * This is a reactive counterpart of {@link UserDetailsChecker}. To use existing an checker,
+ * refer to {@link org.springframework.security.authentication.ReactiveDelegatingUserDetailsChecker}
+ *
+ * @author Alexey Nesterov
+ * @since 5.2
+ */
+public interface ReactiveUserDetailsChecker {
```
rwinch: I don't know that we need a reactive equivalent since the UserDetails has already been looked up and thus everything is non blocking. This does presume that nothing additional is needed to be looked up, but I think this is a pretty safe assumption given the API is for checking a UserDetails. If additional details are necessary it can operate on a custom implementation of UserDetails which contains the additional properties that need to be checked.

alek-sys: It is gone now. I wasn't really sure about this one, my confusion was related to the fact that current checker throws exceptions, rather than returning a value. However more I look into that, more I understand that shifting from exception is much bigger task.

alek-sys:
I've also updated the integration test to use `WebTestClient`, it makes more readable.

alek-sys:
Hey @rwinch, I'm not planning any more updates here, so feel free to review this PR whever it is a good time for you. Thanks for your help!

alek-sys:
@rwinch sorry for a nudge, do you think this PR can make it to 5.2?

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -576,6 +583,93 @@ public FormLoginSpec formLogin() {
...
+		public X509Spec authenticationManager(ReactiveAuthenticationManager authenticationManager) {
+			this.authenticationManager = authenticationManager;
+			return this;
+		}
+
+		public ServerHttpSecurity and() {
+			return ServerHttpSecurity.this;
+		}
+
+		public void configure(ServerHttpSecurity http) {
```
rwinch: Please make this protected to avoid exposing methods that consumers of the API do not care about.

alek-sys:
@rwinch no worries, I've just changed the method visibility. Also, rebase on top of lastest changes in master. Thanks for your guidance and help on that!

Issue #6338, opened 2018-12-28
DelegatingRequestMatcherHeaderWriter with Tiles not work as intended


yoshikawaa:
### Summary
Related #5945 

`DelegatingRequestMatcherHeaderWriter` can not determine the request path exactly as intended.
`DelegatingRequestMatcherHeaderWriter` determines the request path when committing or including the response, but in that case the request path may have been changed to the JSP path.

My environments as follows.
* Application
  * Spring Boot Dependencies 2.1.1.RELEASE (Spring MVC + Spring Security)
  * Tiles JSP 3.0.8
* Application Server
  * Tomcat 9.0.10 (cargo-maven2-plugin), Tomcat 8.5.9 (Pivotal TC Server 3.2.4)

Please see my sample.
* https://github.com/yoshikawaa/spring-security-gh6338

### Actual Behavior

* `DelegatingRequestMatcherHeaderWriter#requestMatcher#pattern` : `/delegating/**`.
* The Request Path : `/delegating`
* The Actual Matching Path : `/WEB-INF/views/layout/template.jsp`

The request path should be matched and the header should be output.
But in fact, at the timing of `DelegatingRequestMatcherHeaderWriter#writeHeaders` the request path has already been changed  and it will not match.

The problem is that even if this is a problem in implementing an Application server or JSP, it is a possible problem with major servers.

### Expected Behavior

It is necessary to be able to match against the actual request path without being influenced by unintended change of the request.

I think that the only correct way to solve the problem is to determine the request path at the timing of `HeaderWriterFilter#doFilterInternal`.

jzheaux:
@yoshikawaa Thanks for you post! 

There are a couple of ways that I think this could be addressed.

One option you have is that, in a forward, the original request uri is saved in the request as an attribute, which you could obtain like so:

```java
AntPathMatcher pathMatcher = new AntPathMatcher();
RequestMatcher requestMatcher = request -> {
    String uri = (String) request.getAttribute(WebUtils.FORWARD_REQUEST_URI_ATTRIBUTE);
    return pathMatcher.matches("/delegating/**", uri);
}
return new DelegatingRequestMatcherHeaderWriter(requestMatcher, headerWriter);
```

A slightly less powerful way is to configure so the headers are written at the beginning of the request: https://github.com/spring-projects/spring-security/issues/6501 This can cause problems with the cache control header, so it's not ideal, but might be worth investigating.

[...]

PR #6339, opened 2018-12-28
improve matching request path at DelegatingRequestMatcherHeaderWriter Fixes: gh-6338

yoshikawaa:
Fixes: gh-6338


rwinch:
Thanks for the PR @yoshikawaa! However, I'm going to close this so we can better understand what you are trying to achieve. We can discuss on the original gh-6338

jzheaux:
Following up with @rwinch's comment that we close this PR while discussion ensues on gh-6338.

PR #6348, opened 2019-01-03
Use X-Forwarded headers during OAuth2 callback

### LINKED ISSUE #6347: OAuth2 reactive callback invalid_redirect_uri_parameter behind reverse proxy
### Summary

When using OAuth2/OIDC in reactive mode behind a reverse proxy, callback uri fails with error code invalid_redirect_uri_parameter

### Actual Behavior

Before redirecting user to OAuth2 server, redirect URI is computed using `org.springframework.web.util.UriComponentsBuilder#fromHttpRequest` which uses `org.springframework.web.util.UriComponentsBuilder#adaptFromForwardedHeaders`.
When user comes back from OAuth2 server, current URI is compared with stored one, but URI is computed using `org.springframework.web.util.UriComponentsBuilder#fromUri` which does not use `X-Forwarded-*` headers

### Expected Behavior

Compute URI using same mechanism.

### Version

5.1.1.RELEASE

I will submit a pull request

trecloux:
X-Forwarded headers are used to compute redirect_uri before redirecting
user to OAuth2 server (DefaultServerOAuth2AuthorizationRequestResolver),
but not when comparing current URI with stored URI.

Fixes #6347

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/ServerOAuth2AuthorizationCodeAuthenticationTokenConverter.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/ServerOAuth2AuthorizationCodeAuthenticationTokenConverter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -105,7 +105,7 @@ public class ServerOAuth2AuthorizationCodeAuthenticationTokenConverter
 	private static OAuth2AuthorizationResponse convertResponse(ServerWebExchange exchange) {
 		MultiValueMap<String, String> queryParams = exchange.getRequest()
 				.getQueryParams();
-		String redirectUri = UriComponentsBuilder.fromUri(exchange.getRequest().getURI())
+		String redirectUri = UriComponentsBuilder.fromHttpRequest(exchange.getRequest())
 				.query(null)
 				.build()
 				.toUriString();


rwinch:
Thanks for the PR! However, we recommend that users leverage [ForwardedHeaderTransformer](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/server/adapter/ForwardedHeaderTransformer.html) or  [ForwardedHeaderFilter](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/filter/reactive/ForwardedHeaderFilter.html) if you are still on Spring 5.0.x.

This is preferred for a few reasons:

* This allows users to opt in to using the X-Forwarded headers
* Opt in vs on by default is important since malicious users can use X-Forwarded headers to perform exploits if a developer is unaware that X-Forwarded headers are being used
* This is a cross cutting concern. Rather than changing code everywhere it should be centralized
* There are [other](https://tools.ietf.org/html/rfc7239) [ways](http://www.haproxy.org/download/1.8/doc/proxy-protocol.txt) to specify a proxy header and we want users to be able to swap this out easily.

trecloux:
Hi @rwinch ,

Thanks for your detailed feedback, however I think that it would be consistent to use the same URI computation source when we first compute `redirect_uri` in `DefaultServerOAuth2AuthorizationRequestResolver` class.
Current situation leads to errors comparing current URI with stored redirect_uri.

Would you want me to submit a pull request to normalize URI computation in the opposite way ?

rwinch:
@trecloux Yes please!

springfan:
> Thanks for the PR! However, we recommend that users leverage [ForwardedHeaderTransformer](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/server/adapter/ForwardedHeaderTransformer.html) or [ForwardedHeaderFilter](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/filter/reactive/ForwardedHeaderFilter.html) if you are still on Spring 5.0.x.

Does this has the same effect as using:
```
server:
  use-forward-headers: true
  tomcat:
    remote-ip-header: x-forwarded-for
    protocol-header: x-forwarded-proto
```
I'm using a reverse proxy struggling with `[invalid_redirect_uri_parameter]` while the redirect urls on the client side/on the browser are correct. The message appears after the oauth provider redirects the client back to the Spring app.

Reverse proxy url: [host.tld]/app
Original url: [host2.tld]

PR #6352, opened 2019-01-04
Add Support for Opaque OAuth2 Tokens to Resource Server

jzheaux:

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,6 +22,7 @@ import org.springframework.context.ApplicationContext;
 import org.springframework.core.convert.converter.Converter;
 import org.springframework.security.authentication.AbstractAuthenticationToken;
 import org.springframework.security.authentication.AuthenticationManager;
+import org.springframework.security.authentication.AuthenticationProvider;
 import org.springframework.security.config.annotation.web.HttpSecurityBuilder;
 import org.springframework.security.config.annotation.web.configurers.AbstractHttpConfigurer;
 import org.springframework.security.config.annotation.web.configurers.CsrfConfigurer;
@@ -31,6 +32,7 @@ import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
 import org.springframework.security.oauth2.jwt.Jwt;
 import org.springframework.security.oauth2.jwt.JwtDecoder;
 import org.springframework.security.oauth2.jwt.NimbusJwtDecoder;
+import org.springframework.security.oauth2.server.resource.authentication.OAuth2IntrospectionAuthenticationProvider;
 import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationConverter;
 import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationProvider;
 import org.springframework.security.oauth2.server.resource.web.BearerTokenAuthenticationEntryPoint;
@@ -86,6 +88,10 @@ import static org.springframework.security.oauth2.jwt.JwtProcessors.withJwkSetUr
  * </li>
  * </ul>
  *
+ * <p>
+ * When using {@link #opaque()}, supply an introspection endpoint and its authentication configuration
+ * </p>
+ *
  * <h2>Security Filters</h2>
  *
  * The following {@code Filter}s are populated when {@link #jwt()} is configured:
@@ -123,7 +129,9 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 	private final ApplicationContext context;
 
 	private BearerTokenResolver bearerTokenResolver;
+
 	private JwtConfigurer jwtConfigurer;
+	private OpaqueTokenConfigurer opaqueTokenConfigurer;
 
 	private AccessDeniedHandler accessDeniedHandler = new BearerTokenAccessDeniedHandler();
 	private AuthenticationEntryPoint authenticationEntryPoint = new BearerTokenAuthenticationEntryPoint();
@@ -160,6 +168,14 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 		return this.jwtConfigurer;
 	}
 
+	public OpaqueTokenConfigurer opaqueToken() {
+		if (this.opaqueTokenConfigurer == null) {
+			this.opaqueTokenConfigurer = new OpaqueTokenConfigurer();
+		}
+
+		return this.opaqueTokenConfigurer;
+	}
+
 	@Override
 	public void init(H http) throws Exception {
 		registerDefaultAccessDeniedHandler(http);
@@ -182,24 +198,34 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 
 		http.addFilter(filter);
 
-		if ( this.jwtConfigurer == null ) {
-			throw new IllegalStateException("Jwt is the only supported format for bearer tokens " +
-					"in Spring Security and no Jwt configuration was found. Make sure to specify " +
-					"a jwk set uri by doing http.oauth2ResourceServer().jwt().jwkSetUri(uri), or wire a " +
-					"JwtDecoder instance by doing http.oauth2ResourceServer().jwt().decoder(decoder), or " +
-					"expose a JwtDecoder instance as a bean and do http.oauth2ResourceServer().jwt().");
+		if (this.jwtConfigurer != null && this.opaqueTokenConfigurer != null) {
+			throw new IllegalStateException("Spring Security only supports JWTs or Opaque Tokens, not both at the " +
+					"same time");
 		}
 
-		JwtDecoder decoder = this.jwtConfigurer.getJwtDecoder();
-		Converter<Jwt, ? extends AbstractAuthenticationToken> jwtAuthenticationConverter =
-				this.jwtConfigurer.getJwtAuthenticationConverter();
+		if (this.jwtConfigurer == null && this.opaqueTokenConfigurer == null) {
+			throw new IllegalStateException("Jwt and Opaque Token are the only supported formats for bearer tokens " +
+					"in Spring Security and neither was found. Make sure to configure JWT " +
+					"via http.oauth2ResourceServer().jwt() or Opaque Tokens via " +
+					"http.oauth2ResourceServer().opaque().");
+		}
 
-		JwtAuthenticationProvider provider =
-				new JwtAuthenticationProvider(decoder);
-		provider.setJwtAuthenticationConverter(jwtAuthenticationConverter);
-		provider = postProcess(provider);
+		if (this.jwtConfigurer != null) {
+			JwtDecoder decoder = this.jwtConfigurer.getJwtDecoder();
+			Converter<Jwt, ? extends AbstractAuthenticationToken> jwtAuthenticationConverter =
+					this.jwtConfigurer.getJwtAuthenticationConverter();
 
-		http.authenticationProvider(provider);
+			JwtAuthenticationProvider provider =
+					new JwtAuthenticationProvider(decoder);
+			provider.setJwtAuthenticationConverter(jwtAuthenticationConverter);
+			provider = postProcess(provider);
+
+			http.authenticationProvider(provider);
+		}
+
+		if (this.opaqueTokenConfigurer != null) {
+			http.authenticationProvider(this.opaqueTokenConfigurer.getProvider());
+		}
 	}
 
 	public class JwtConfigurer {
@@ -245,6 +271,31 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 			}
 
 			return this.decoder;
+		}
+	}
+
+	public class OpaqueTokenConfigurer {
+		private String introspectionUri;
+		private String introspectionClientId;
+		private String introspectionClientSecret;
+
+		public OpaqueTokenConfigurer introspectionUri(String introspectionUri) {
+			Assert.notNull(introspectionUri, "introspectionUri cannot be null");
+			this.introspectionUri = introspectionUri;
+			return this;
+		}
+
+		public OpaqueTokenConfigurer introspectionClientCredentials(String clientId, String clientSecret) {
+			Assert.notNull(clientId, "clientId cannot be null");
+			Assert.notNull(clientSecret, "clientSecret cannot be null");
+			this.introspectionClientId = clientId;
+			this.introspectionClientSecret = clientSecret;
+			return this;
+		}
+
+		AuthenticationProvider getProvider() {
+			return new OAuth2IntrospectionAuthenticationProvider(this.introspectionUri,
+					this.introspectionClientId, this.introspectionClientSecret);
 		}
 	}
 
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OAuth2IntrospectionAuthenticationProvider.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OAuth2IntrospectionAuthenticationProvider.java
@@ -1,0 +1,283 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.server.resource.authentication;
+
+import java.net.URI;
+import java.net.URL;
+import java.time.Instant;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import com.nimbusds.oauth2.sdk.TokenIntrospectionResponse;
+import com.nimbusds.oauth2.sdk.TokenIntrospectionSuccessResponse;
+import com.nimbusds.oauth2.sdk.http.HTTPResponse;
+import com.nimbusds.oauth2.sdk.id.Audience;
+
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
+import org.springframework.http.RequestEntity;
+import org.springframework.http.ResponseEntity;
+import org.springframework.http.client.support.BasicAuthenticationInterceptor;
+import org.springframework.security.authentication.AbstractAuthenticationToken;
+import org.springframework.security.authentication.AuthenticationProvider;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.core.GrantedAuthority;
+import org.springframework.security.core.authority.SimpleGrantedAuthority;
+import org.springframework.security.oauth2.core.OAuth2AccessToken;
+import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
+import org.springframework.security.oauth2.server.resource.BearerTokenAuthenticationToken;
+import org.springframework.security.oauth2.server.resource.BearerTokenError;
+import org.springframework.util.Assert;
+import org.springframework.util.LinkedMultiValueMap;
+import org.springframework.util.MultiValueMap;
+import org.springframework.web.client.RestOperations;
+import org.springframework.web.client.RestTemplate;
+
+import static org.springframework.security.oauth2.server.resource.authentication.OAuth2IntrospectionClaimNames.AUDIENCE;
+import static org.springframework.security.oauth2.server.resource.authentication.OAuth2IntrospectionClaimNames.CLIENT_ID;
+import static org.springframework.security.oauth2.server.resource.authentication.OAuth2IntrospectionClaimNames.EXPIRES_AT;
+import static org.springframework.security.oauth2.server.resource.authentication.OAuth2IntrospectionClaimNames.ISSUED_AT;
+import static org.springframework.security.oauth2.server.resource.authentication.OAuth2IntrospectionClaimNames.ISSUER;
+import static org.springframework.security.oauth2.server.resource.authentication.OAuth2IntrospectionClaimNames.NOT_BEFORE;
+import static org.springframework.security.oauth2.server.resource.authentication.OAuth2IntrospectionClaimNames.SCOPE;
+
+/**
+ * An {@link AuthenticationProvider} implementation for opaque
+ * <a href="https://tools.ietf.org/html/rfc6750#section-1.2" target="_blank">Bearer Token</a>s,
+ * using an
+ * <a href="https://tools.ietf.org/html/rfc7662" target="_blank">OAuth 2.0 Introspection Endpoint</a>
+ * to check the token's validity and reveal its attributes.
+ * <p>
+ * This {@link AuthenticationProvider} is responsible for introspecting and verifying an opaque access token,
+ * returning its attributes set as part of the {@see Authentication} statement.
+ * <p>
+ * Scopes are translated into {@link GrantedAuthority}s according to the following algorithm:
+ * <ol>
+ * <li>
+ * If there is a "scope" attribute, then convert to a {@link Collection} of {@link String}s.
+ * <li>
+ * Take the resulting {@link Collection} and prepend the "SCOPE_" keyword to each element, adding as {@link GrantedAuthority}s.
+ * </ol>
+ *
+ * @author Josh Cummings
+ * @since 5.2
+ * @see AuthenticationProvider
+ */
+public final class OAuth2IntrospectionAuthenticationProvider implements AuthenticationProvider {
+	private URI introspectionUri;
+	private RestOperations restOperations;
+
+	/**
+	 * Creates a {@code OAuth2IntrospectionAuthenticationProvider} with the provided parameters
+	 *
+	 * @param introspectionUri The introspection endpoint uri
+	 * @param clientId The client id authorized to introspect
+	 * @param clientSecret The client secret for the authorized client
+	 */
+	public OAuth2IntrospectionAuthenticationProvider(String introspectionUri, String clientId, String clientSecret) {
+		Assert.notNull(introspectionUri, "introspectionUri cannot be null");
+		Assert.notNull(clientId, "clientId cannot be null");
+		Assert.notNull(clientSecret, "clientSecret cannot be null");
+
+		this.introspectionUri = URI.create(introspectionUri);
+		RestTemplate restTemplate = new RestTemplate();
+		restTemplate.getInterceptors().add(new BasicAuthenticationInterceptor(clientId, clientSecret));
+		this.restOperations = restTemplate;
+	}
+
+	/**
+	 * Creates a {@code OAuth2IntrospectionAuthenticationProvider} with the provided parameters
+	 *
+	 * @param introspectionUri The introspection endpoint uri
+	 * @param restOperations The client for performing the introspection request
+	 */
+	public OAuth2IntrospectionAuthenticationProvider(String introspectionUri, RestOperations restOperations) {
+		Assert.notNull(introspectionUri, "introspectionUri cannot be null");
+		Assert.notNull(restOperations, "restOperations cannot be null");
+
+		this.introspectionUri = URI.create(introspectionUri);
+		this.restOperations = restOperations;
+	}
+
+	/**
+	 * Introspect and validate the opaque
+	 * <a href="https://tools.ietf.org/html/rfc6750#section-1.2" target="_blank">Bearer Token</a>.
+	 *
+	 * @param authentication the authentication request object.
+	 *
+	 * @return A successful authentication
+	 * @throws AuthenticationException if authentication failed for some reason
+	 */
+	@Override
+	public Authentication authenticate(Authentication authentication) throws AuthenticationException {
+		if (!(authentication instanceof BearerTokenAuthenticationToken)) {
+			return null;
+		}
+
+		// introspect
+		BearerTokenAuthenticationToken bearer = (BearerTokenAuthenticationToken) authentication;
+		TokenIntrospectionSuccessResponse response = introspect(bearer.getToken());
+		Map<String, Object> claims = convertClaimsSet(response);
+		Instant iat = (Instant) claims.get(ISSUED_AT);
+		Instant exp = (Instant) claims.get(EXPIRES_AT);
+
+		// construct token
+		OAuth2AccessToken token  = new OAuth2AccessToken(OAuth2AccessToken.TokenType.BEARER,
+				bearer.getToken(), iat, exp);
+		Collection<GrantedAuthority> authorities = extractAuthorities(claims);
+		AbstractAuthenticationToken result =
+				new OAuth2IntrospectionAuthenticationToken(token, claims, authorities);
+		result.setDetails(bearer.getDetails());
+		return result;
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public boolean supports(Class<?> authentication) {
+		return BearerTokenAuthenticationToken.class.isAssignableFrom(authentication);
+	}
+
+	private TokenIntrospectionSuccessResponse introspect(String token) {
+		return Optional.of(token)
+				.map(this::buildRequest)
+				.map(this::makeRequest)
+				.map(this::adaptToNimbusResponse)
+				.map(this::parseNimbusResponse)
+				.map(this::castToNimbusSuccess)
+				// relying solely on the authorization server to validate this token (not checking 'exp', for example)
+				.filter(TokenIntrospectionSuccessResponse::isActive)
+				.orElseThrow(() -> new OAuth2AuthenticationException(
+						invalidToken("Provided token [" + token + "] isn't active")));
+	}
+
+	private RequestEntity<MultiValueMap<String, String>> buildRequest(String token) {
+		HttpHeaders headers = requestHeaders();
+		MultiValueMap<String, String> body = requestBody(token);
+		return new RequestEntity<>(body, headers, HttpMethod.POST, this.introspectionUri);
+	}
+
+	private HttpHeaders requestHeaders() {
+		HttpHeaders headers = new HttpHeaders();
+		headers.setAccept(Collections.singletonList(MediaType.APPLICATION_JSON_UTF8));
+		return headers;
+	}
+
+	private MultiValueMap<String, String> requestBody(String token) {
+		MultiValueMap<String, String> body = new LinkedMultiValueMap<>();
+		body.add("token", token);
+		return body;
+	}
+
+	private ResponseEntity<String> makeRequest(RequestEntity<?> requestEntity) {
+		try {
+			return this.restOperations.exchange(requestEntity, String.class);
+		} catch (Exception ex) {
+			throw new OAuth2AuthenticationException(
+					invalidToken(ex.getMessage()), ex);
+		}
+	}
+
+	private HTTPResponse adaptToNimbusResponse(ResponseEntity<String> responseEntity) {
+		HTTPResponse response = new HTTPResponse(responseEntity.getStatusCodeValue());
+		response.setHeader(HttpHeaders.CONTENT_TYPE, responseEntity.getHeaders().getContentType().toString());
+		response.setContent(responseEntity.getBody());
+
+		if (response.getStatusCode() != HTTPResponse.SC_OK) {
+			throw new OAuth2AuthenticationException(
+					invalidToken("Introspection endpoint responded with " + response.getStatusCode()));
+		}
+		return response;
+	}
+
+	private TokenIntrospectionResponse parseNimbusResponse(HTTPResponse response) {
+		try {
+			return TokenIntrospectionResponse.parse(response);
+		} catch (Exception ex) {
+			throw new OAuth2AuthenticationException(
+					invalidToken(ex.getMessage()), ex);
+		}
+	}
+
+	private TokenIntrospectionSuccessResponse castToNimbusSuccess(TokenIntrospectionResponse introspectionResponse) {
+		if (!introspectionResponse.indicatesSuccess()) {
+			throw new OAuth2AuthenticationException(invalidToken("Token introspection failed"));
+		}
+		return (TokenIntrospectionSuccessResponse) introspectionResponse;
+	}
+
+	private Map<String, Object> convertClaimsSet(TokenIntrospectionSuccessResponse response) {
+		Map<String, Object> claims = response.toJSONObject();
+		if (response.getAudience() != null) {
+			List<String> audience = response.getAudience().stream()
+					.map(Audience::getValue).collect(Collectors.toList());
+			claims.put(AUDIENCE, Collections.unmodifiableList(audience));
+		}
+		if (response.getClientID() != null) {
+			claims.put(CLIENT_ID, response.getClientID().getValue());
+		}
+		if (response.getExpirationTime() != null) {
+			Instant exp = response.getExpirationTime().toInstant();
+			claims.put(EXPIRES_AT, exp);
+		}
+		if (response.getIssueTime() != null) {
+			Instant iat = response.getIssueTime().toInstant();
+			claims.put(ISSUED_AT, iat);
+		}
+		if (response.getIssuer() != null) {
+			claims.put(ISSUER, issuer(response.getIssuer().getValue()));
+		}
+		if (response.getNotBeforeTime() != null) {
+			claims.put(NOT_BEFORE, response.getNotBeforeTime().toInstant());
+		}
+		if (response.getScope() != null) {
+			claims.put(SCOPE, Collections.unmodifiableList(response.getScope().toStringList()));
+		}
+
+		return claims;
+	}
+
+	private Collection<GrantedAuthority> extractAuthorities(Map<String, Object> claims) {
+		Collection<String> scopes = (Collection<String>) claims.get(SCOPE);
+		return Optional.ofNullable(scopes).orElse(Collections.emptyList())
+				.stream()
+				.map(authority -> new SimpleGrantedAuthority("SCOPE_" + authority))
+				.collect(Collectors.toList());
+	}
+
+	private URL issuer(String uri) {
+		try {
+			return new URL(uri);
+		} catch (Exception ex) {
+			throw new OAuth2AuthenticationException(
+					invalidToken("Invalid " + ISSUER + " value: " + uri), ex);
+		}
+	}
+
+	private static BearerTokenError invalidToken(String message) {
+		return new BearerTokenError("invalid_token",
+				HttpStatus.UNAUTHORIZED, message,
+				"https://tools.ietf.org/html/rfc7662#section-2.2");
+	}
+}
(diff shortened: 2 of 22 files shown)


oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenWithClaims.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.springframework.util.Assert;
+
+/**
+ * A composite that contains both an OAuth 2.0 token and any verified claims
+ *
+ * @author Josh Cummings
+ * @since 5.2
+ */
+public class OAuth2TokenWithClaims extends AbstractOAuth2Token implements ClaimAccessor {
```
rwinch: I'm struggling a bit with the addition of this class. I think a lot of this is that the name feels a bit artificial to me. It should be something that is a part of the spec rather than having a specific property on it. I don't think we should have classes that are named based upon properties they contain.

I haven't looked around yet for a more appropriate existing class, but it seems like there should be one. If there isn't one, I believe we should rethink the name to align with something within the spec.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/IntrospectedOAuth2TokenAuthenticationToken.java
```diff
@@ -0,0 +1,57 @@
...
+
+	public IntrospectedOAuth2TokenAuthenticationToken(OAuth2TokenWithClaims token,
+			Collection<? extends GrantedAuthority> authorities) {
+
+		super(token, authorities);
+		setAuthenticated(true);
+	}
+
+	@Override
+	public Map<String, Object> getTokenAttributes() {
```
rwinch: It is strange that the token attributes are getting overridden to be the claims.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/IntrospectedOAuth2TokenAuthenticationToken.java
```diff
@@ -0,0 +1,57 @@
...
+		return getToken().getClaims();
+	}
+
+	@Override
+	public String getName() {
+		return (String) getTokenAttributes().get("username");
```
rwinch: Again it is strange this is overridden vs using the principal toString

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
```diff
@@ -0,0 +1,244 @@
...
+
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public boolean supports(Class<?> authentication) {
+		return BearerTokenAuthenticationToken.class.isAssignableFrom(authentication);
+	}
+
+	private HTTPRequest introspectionRequest(String token) {
```
rwinch: I'd prefer us to use RestTemplate to retrieve and then convert it using Nimbus. This allows for easily modifying timeouts, proxy settings, SSL certificates, etc. Can you please update this to use RestTemplate and allow a user to inject RestOperations?

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenWithClaims.java
jzheaux: I agree with this sentiment. One name I've considered is `OAuth2TokenDetails`. My reasoning is that `UserDetails` contains a user and details about the user, and an `OAuth2TokenDetails` contains a token and details about the token. It also aligns with the fact that both of these would be the result of `getPrincipal`, in their associated contexts.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/IntrospectedOAuth2TokenAuthenticationToken.java
jzheaux: Do you see it as inconsistent with the same behavior in [`JwtAuthenticationToken`](https://github.com/spring-projects/spring-security/blob/master/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationToken.java#L64)? That is, they appear to be similar situations to me.

Is there a semantic difference between attributes and claims, e.g. attributes == verified claims? If so, then it would seem that `JwtAuthenticationToken` might be taking the wrong approach as well.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/IntrospectedOAuth2TokenAuthenticationToken.java
jzheaux: Good point, I'll think about this a bit more. I'll just mention here that this is similar to [`JwtAuthenticationToken`](https://github.com/spring-projects/spring-security/blob/master/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationToken.java#L72), which was the main driver for doing it this way.

jzheaux: It would be odd, I think, to call `getPrincipal().toString()` here because `getPrincipal()` is an `AbstractOAuth2Token`.

It would mean that either `AbstractOAuth2Token` (or its derivative for this class) would need to pick a claim from itself to be the principal's "name", and that feels strange since an OAuth token is not inherently a "named" principal.

Perhaps it would be better to add `String name` as a constructor parameter to this authentication token.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenWithClaims.java
jzheaux: Another option I considered was `IntrospectedToken`. This certainly refers to a spec term, which is good, but it fails at a couple of points:

- First, it's odd that the name indicates how it was validated. `UserDetails`, `Jwt`, and `username` do not indicate how the principal was validated, though the `Authentication` type might, e.g. `UsernamePasswordAuthenticationToken`.
- Second, it pushes this information into web controllers:

```java
@GetMapping
public String request(@AuthenticationPrincipal IntrospectedToken token) {
    // ...
}
```
Which isn't as flexible.

For example, a class like this is handy because the code could now do:

```java
public Jwt extends OAuth2TokenWithClaims
```

And controllers could now do

```java
@GetMapping
public String request(@AuthenticationPrincipal OAuth2TokenWithClaims token)
```

Making the user's code more flexible.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/IntrospectedOAuth2TokenAuthenticationToken.java
rwinch: I'm just wondering if the claims should be translated into attributes and injected as such.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
```diff
@@ -0,0 +1,256 @@
...
+
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public boolean supports(Class<?> authentication) {
+		return BearerTokenAuthenticationToken.class.isAssignableFrom(authentication);
+	}
+
+	private TokenIntrospectionSuccessResponse introspect(String token) {
```
jgrandja: As per spec:

> The introspection endpoint MAY accept other OPTIONAL parameters to
   provide further context to the query.  For instance, an authorization
   server may desire to know the IP address of the client accessing the
   protected resource to determine if the correct client is likely to be
   presenting the token.

Given that optional parameters may be accepted by a provider, the user will not be able to customize the request given the current design. Please take a look at `DefaultAuthorizationCodeTokenResponseClient.setRequestEntityConverter()` to see how this can be supported.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
```diff
@@ -0,0 +1,256 @@
...
+		HttpHeaders headers = new HttpHeaders();
+		headers.setAccept(Collections.singletonList(MediaType.APPLICATION_JSON_UTF8));
+		RequestEntity<String> requestEntity =
+				new RequestEntity<>("token=" + token, headers, HttpMethod.POST, this.introspectionUri);
+		TokenIntrospectionSuccessResponse response = asSuccess(parse(adapt(request(requestEntity))));
```
jgrandja: I don't see any benefit in using Nimbus' API's to extract the response. The response returned is really a `Map` that can easily be handled by `RestTemplate`. Or another option and the one I prefer is to provide a `HttpMessageConverter`. See `OAuth2AccessTokenResponseHttpMessageConverter` which is used by `DefaultAuthorizationCodeTokenResponseClient`.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
```diff
@@ -0,0 +1,256 @@
...
+	private RestOperations restOperations;
+
+	/**
+	 * Creates a {@code OpaqueTokenAuthenticationProvider} with the provided parameters
+	 *
+	 * @param introspectionUri The introspection endpoint uri
+	 * @param clientId The client id authorized to introspect
+	 * @param clientSecret The client secret for the authorized client
+	 */
+	public OpaqueTokenAuthenticationProvider(String introspectionUri, String clientId, String clientSecret) {
```
jgrandja: As per spec:

> the endpoint MUST also require
   some form of authorization to access this endpoint, such as client
   authentication as described in OAuth 2.0 [RFC6749] or a separate
   OAuth 2.0 access token such as the bearer token described in OAuth
   2.0 Bearer Token Usage [RFC6750].  The methods of managing and
   validating these authentication credentials are out of scope of this
   specification.

The Authorization Server may support other methods for client authentication, not just HTTP Basic. So this constructor will limit us. We need to consider a more flexible approach for customizing client credentials. The approach of `DefaultAuthorizationCodeTokenResponseClient.setRequestEntityConverter` might do the trick.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
```diff
@@ -0,0 +1,256 @@
...
+	private TokenIntrospectionSuccessResponse asSuccess(TokenIntrospectionResponse introspectionResponse) {
+		if (!introspectionResponse.indicatesSuccess()) {
+			throw new OAuth2AuthenticationException(invalidToken("Token introspection failed"));
+		}
+		return (TokenIntrospectionSuccessResponse) introspectionResponse;
+	}
+
+	private Map<String, Object> extractClaimSet(TokenIntrospectionSuccessResponse response) {
+		Map<String, Object> claims = response.toJSONObject();
+		if (response.getAudience() != null) {
```
jgrandja: Why do we need to explicitly extract each claim? There may be custom (non-standard) claims in the response that we won't know and will be excluded in the claims set. I also see that `jti` and `sub` are not extracted here. Either way, I don't think we should extract claims individually. I feel all claims returned from the response should be the claim set that is composed in the `Authentication`

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
```diff
@@ -0,0 +1,256 @@
...
+	 * <a href="https://tools.ietf.org/html/rfc6750#section-1.2" target="_blank">Bearer Token</a>.
+	 *
+	 * @param authentication the authentication request object.
+	 *
+	 * @return A successful authentication
+	 * @throws AuthenticationException if authentication failed for some reason
+	 */
+	@Override
+	public Authentication authenticate(Authentication authentication) throws AuthenticationException {
+		if (!(authentication instanceof BearerTokenAuthenticationToken)) {
```
jgrandja: This check is not needed given that `supports` is implemented.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
```diff
@@ -0,0 +1,256 @@
...
+			return null;
+		}
+
+		// introspect
+		BearerTokenAuthenticationToken bearer = (BearerTokenAuthenticationToken) authentication;
+		TokenIntrospectionSuccessResponse response = introspect(bearer.getToken());
+		Map<String, Object> claims = extractClaimSet(response);
+		Instant iat = (Instant) claims.get("iat");
```
jgrandja: We should use constants wherever possible - > `JwtClaimNames.IAT`

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenWithClaims.java
jgrandja: My vote is to name it `IntrospectedOAuth2AccessToken` and extend `OAuth2AccessToken`. I think the name also aligns with `IntrospectedOAuth2TokenAuthenticationToken`. However, to make it more aligned we may consider renaming to `IntrospectedOAuth2AccessTokenAuthenticationToken` - bit of a long one though.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/IntrospectedOAuth2TokenAuthenticationToken.java
jgrandja: Something else to consider is that `username` is an optional attribute. So is `sub` but that would be a more logical choice I think. Additionally, depending on the provider, the subject identifier may be a non-standard attribute. This may need to be customized.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
jzheaux: This is a good point - I prefer it this way so that this method does not require another method to be called in order to be correct. That said, some `AuthenticationProvider`s throw an `IllegalArgumentException` while others return `null`.

@rwinch what is the reason that other `AuthenticationProvider`s do this? It appears to be a common pattern to check the cast before proceeding in `authenticate`.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
jzheaux: It would surprise me to see `JwtClaimNames` here since this isn't a JWT. Also, does this create a package dependency that we don't want?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
jzheaux: We need to ensure that each claim is of the correct type, e.g. we are coercing here on write instead of requiring coercion on read. Or did I misunderstand and you are asking for something else?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
jzheaux: We can add other constructors later, so I've created a ticket that allows us to address this:  https://github.com/spring-projects/spring-security/issues/6422

Note that `NimbusReactiveJwtDecoder` began with taking a `URL` in one constructor an `RSAPublicKey` in another; eventually, it will likely take something more generic, too. Do you think that that principle does not apply here?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
jzheaux: The benefit is that we don't need to write our own `HttpMessageConverter`. Why do you think we should write this ourselves?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
jzheaux: This is something we can support later, so I've created a ticket to make sure that we address this: https://github.com/spring-projects/spring-security/issues/6424

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenWithClaims.java
jzheaux: @jgrandja, I believe `IntrospectedOAuth2AccessToken` is not ideal for the same reasons that `IntropsectedToken` is not ideal.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/IntrospectedOAuth2TokenAuthenticationToken.java
jzheaux: From the spec, `sub` is:

> Usually a machine-readable identifier of the resource owner who authorized this token.

And `username` is:

> Human-readable identifier for the resource owner who authorized this token.

There isn't any indication in `javax.security.Principal` as to whether it is intended to be human or machine-readable, but it seems to me that if `getPrincipal().toString()` is usually the clearer choice, then I'd expect it to be the human-readable one.

What makes you say that `sub` is the more logical choice?

jzheaux:
@rwinch I've updated to use `RestTemplate`. Please feel free to comment there or anywhere else you see room for improvement.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenWithClaims.java
jzheaux: I've updated this class to be named `OAuth2TokenAccessor` since it extends `AbstractOAuth2Token` and implements `ClaimAccessor`. I like this on hierarchical clarity, but still feel that `OAuth2TokenDetails` carries more semantic weight.

I believe there is value in this becoming the parent to `Jwt`, allowing the application to consider the OAuth 2.0 principal without needing to know the manner in which the token was validated.

This is why I'm not keen on `OAuth2AccessTokenDetails` since its parent would be `OAuth2AccessToken`, meaning that I still need to know which validation method was used in order to correctly obtain the principal, e.g. if I'm using `opaque()` then get an `OAuth2AccessTokenDetails`, but if I'm using `jwt()` then get a `Jwt`. This is disappointing since my access patterns at that level will largely be identical.

I like `OAuth2TokenDetails` and `OAuth2TokenAccessor` because they bear out the semantics that this particular token contains unstructured information--it doesn't follow the JWT structure or the JWS structure, though `Jwt`, its subclass, adds the JWT structure to it, like headers and a specialized `ClaimsAccessor`.

I've also considered two other names, which, based on my explanations above, will hopefully clearly be sub-optimal, but I'll include them for completeness:

`StructuredOAuth2Token` - A plain OAuth 2.0 token doesn't have structure, but one that has been introspected at least has the implied structure of attributes and values. Subclasses impose additional structure, in the same way as I already explained.
`AccessibleOAuth2Token` - An adjective version of `OAuth2TokenAccessor`, simply to maintain the concrete class at the end of the name

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
jzheaux: After chatting with @jgrandja offline about this, it appears that I made a poor choice in method name here. I'll update it to clarify its role.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
jzheaux: In the latest PR, I've added `IntrospectionClaimNames` to house the constants described in the [Introspection RFC](https://tools.ietf.org/html/rfc7662#section-2.2).

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenWithClaims.java
jgrandja: I'm not really feeling `OAuth2TokenAccessor`. Domain objects implicitly have accessors so I feel that it's redundant in the name. As well, the existing classes that implement `ClaimAccessor` do not have `Accessor` in the name either.

I prefer `OAuth2TokenDetails` over `OAuth2TokenAccessor` but I still feel very strongly that we should stick with a name that is referred to in the spec. The obvious is to have `Introspection` or `Introspected` in the name.

After reviewing the spec again, it came to my occurrence that an `IntrospectedOAuth2Token`  should not live outside of an `Authentication` or an Introspection Request/Response flow. I think the following design proposal may address some of your previous concerns @jzheaux as far as naming.

```java
public class IntrospectedOAuth2TokenAuthenticationToken
		extends AbstractOAuth2TokenAuthenticationToken<IntrospectedOAuth2TokenAuthenticationToken.IntrospectedOAuth2Token> {

	public IntrospectedOAuth2TokenAuthenticationToken(IntrospectedOAuth2Token introspectedToken) {
		super(introspectedToken);
	}

	@Override
	public Map<String, Object> getTokenAttributes() {
		return getToken().getAttributes();
	}

	public static class IntrospectedOAuth2Token extends AbstractOAuth2Token {
		private final Map<String, Object> attributes;

		public IntrospectedOAuth2Token(String tokenValue, Instant issuedAt,
										Instant expiresAt, Map<String, Object> attributes) {
			super(tokenValue, issuedAt, expiresAt);
			this.attributes = Collections.unmodifiableMap(new LinkedHashMap<>(attributes));
		}

		public Map<String, Object> getAttributes() {
			return this.attributes;
		}
	}
}
```

Thoughts?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/IntrospectedOAuth2TokenAuthenticationToken.java
jgrandja: > What makes you say that sub is the more logical choice?

`sub` should be a GUID when combined with `iss`. However, it's possible that `username` may not be unique since 2 people can have the same name/username across systems (`iss`).

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
jgrandja: I think this should be part of this PR as it's minimal effort and we have an established pattern in place. See `DefaultAuthorizationCodeTokenResponseClient`, `DefaultClientCredentialsTokenResponseClient` and `DefaultOAuth2UserService`,

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
jgrandja: It just seems inconsistent to use `RestTemplate` to initiate the request and Nimbus API to parse the response. `HttpMessageConverter` is part of `RestTemplate`'s API so we should leverage it for response parsing.

As well, the main benefit is that the user will be able to customize the response processing by either extending our implementation of `HttpMessageConverter` for the Introspection Response or providing their own. Keep in mind that custom parameters may come back from the Introspection Response so users will need a way to access them.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenWithClaims.java
rwinch: I don't like the fact that the name has Token in it twice. This seems to be a problem in more than one place.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -160,6 +168,14 @@ public JwtConfigurer jwt() {
 		return this.jwtConfigurer;
 	}
 
+	public OpaqueConfigurer opaque() {
```
rwinch: I think we want to name this something like opaqueToken()

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -248,6 +274,35 @@ JwtDecoder getJwtDecoder() {
 		}
 	}
 
+	public class OpaqueConfigurer {
```
rwinch: I think we want something like OpaqueTokenConfigurer here. I realize we have JwtConfigurer, but Jwt has Token in the acronym.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenAccessor.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.springframework.util.Assert;
+
+/**
+ * An OAuth 2.0 token that implements {@link ClaimAccessor}
+ *
+ * @author Josh Cummings
+ * @since 5.2
+ */
+public class OAuth2TokenAccessor extends AbstractOAuth2Token implements ClaimAccessor {
```
rwinch: Some ideas for a name `OAuth2TokenAttributes` `OAuth2TokenClaims`

samples/boot/oauth2resourceserver-opaque/src/main/java/sample/OAuth2ResourceServerSecurityConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+	@Value("${spring.security.oauth2.resourceserver.opaque.introspection-uri}") String introspectionUri;
+	@Value("${spring.security.oauth2.resourceserver.opaque.introspection-client-id}") String clientId;
+	@Value("${spring.security.oauth2.resourceserver.opaque.introspection-client-secret}") String clientSecret;
+
+	@Override
+	protected void configure(HttpSecurity http) throws Exception {
+		// @formatter:off
+		http
+			.authorizeRequests()
+				.antMatchers("/message/**").hasAuthority("SCOPE_message:read")
```
rwinch: Use mvcMatchers

samples/boot/oauth2resourceserver-opaque/src/main/resources/application.yml
```diff
@@ -0,0 +1,8 @@
+spring:
+  security:
+    oauth2:
+      resourceserver:
+        opaque:
+          introspection-uri: ${mockwebserver.url}/introspect
```
rwinch: Just want to ensure we think about if/how this aligns with OAuth2 Log In. Do we support discovery? If an application is supporting oauth2 log in and resource server do we want them to have to configure the client id/secret twice?

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenWithClaims.java
jgrandja: @rwinch 

> This seems to be a problem in more than one place.

Where is the other place you are referring to?

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenAccessor.java
jgrandja: I like `OAuth2TokenClaims` which may also be the parent for `Jwt` and we can also considerate it for `OidcIdToken`. However, what do we name the `Authentication` class? If we follow the same naming pattern for other `AbstractAuthenticationToken` implementations than we have `OAuth2TokenClaimsAuthenticationToken`. Same would apply to `OAuth2TokenAttributesAuthenticationToken`.

@rwinch This brings us back to the same naming issue  you pointed out [here](https://github.com/spring-projects/spring-security/pull/6352#discussion_r250299115)

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/IntrospectionClaimNames.java
```diff
@@ -0,0 +1,86 @@
...
+package org.springframework.security.oauth2.server.resource.authentication;
+
+/**
+ * The names of the &quot;Introspection Claims&quot; defined by an
+ * <a target="_blank" href="https://tools.ietf.org/html/rfc7662#section-2.2">Introspection Response</a>.
+ *
+ * @author Josh Cummings
+ * @since 5.2
+ */
+public interface IntrospectionClaimNames {
```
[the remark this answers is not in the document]
jgrandja: > Instead perhaps it makes more sense to have concrete methods

I agree and was more envisioning adopting the same pattern as `JwtClaimAccessor`, `IdTokenClaimAccessor`, and `StandardClaimAccessor`. Defining the constant only is the first step. We still need something like `IntrospectionClaimAccessor`

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenWithClaims.java
rwinch: Both `AbstractOAuth2TokenAuthenticationToken` and `IntrospectionOAuth2TokenAuthenticationToken` have token in them twice

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/IntrospectionClaimNames.java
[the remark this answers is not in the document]
jzheaux: What is the use case for end users actually using this class? As a user, I don't care about accessing "introspected" attributes, per se. JWT is a structure, so it makes sense for me to have a `JwtClaimAccessor`, e.g. "I want to access claims this through the JWT-defined structure". "Introspection" is not a structure.

I think it's sensible to have a common place for attributes for classes that need to commonly refer to introspection-defined properties, but it isn't really the end user that needs them.

I don't yet see the need for an Introspection-specific claim accessor, but I think we could add this later if the need becomes apparent.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenAccessor.java
jgrandja: @jzheaux One thing I want to point out when you do the comparison between `UserDetails` and the potential/proposed name `OAuth2TokenDetails`

> A UserDetails contains the user and details about the user. "OAuth2TokenClaims contains the token and claims about the token" doesn't feel as clear.

It's not always the case that a user is associated to an access token. For the `client_credentials` grant, there never is a user association. So `OAuth2TokenDetails` would not make sense here as a comparison to `UserDetails`.

jzheaux: My comparison is not based on the merit of it being a "User". Indeed, there isn't really a user in many requests with Resource Server. It is based on the semantics of the word "Details".

Because `UserDetails` implies that the details can include the `User`, too, then `OAuth2TokenDetails` implies that the details can include the `OAuth2Token`. `Claims` doesn't bear this same weight in the codebase. I don't believe users would infer that `OAuth2TokenClaims` would contain anything more than the claims themselves.

jzheaux: Another idea is `DetailedOAuth2Token`. It indicates that this is a token with additional information. `StructuredOAuth2Token` contains spec language, e.g. the introspection endpoint returns the token in a JSON structure, the JWT spec describes the token in the JWT structure, and I like it for that reason.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/IntrospectionClaimNames.java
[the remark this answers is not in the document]
jzheaux: The user shouldn't have to care that the token was introspected. They care that the claims are available and trusted, both of which are implied by their controller method being invoked (and Spring Security being correctly configured). That they were obtained via an introspection endpoint is immaterial to the user.

For example, if I want to get the `aud` attribute in my controller, it's more flexible to do one of these:

```java
token.getAttributes("aud")
// or
token.getAttributes(CommonAttributeNames.AUDIENCE)
// or
token.getAttributes(JwtClaimNames.AUDIENCE) // see my reasoning below
```

Or, so long as the object hierarchy isn't bound to the way it was validated:

```java 
token.getAudience()
```

An approach that makes sense to me is to write my code this way:

```java
@GetMapping("/protected")
public String get(StructuredOAuth2Token token) {
    JwtClaimsAccessor accessor = () -> token.getAttributes();
    List<String> audience = accessor.getAudience();
    // ...
}
```

This is sensible because I'm accessing the JWT-like attributes, which is the same thing that the spec does (emphasis mine):

> aud
      OPTIONAL.  Service-specific string identifier or list of string
      identifiers representing the intended audience for this token, **as
      defined in JWT**

And to be clear, I'm not saying that the user shouldn't ever care. Like you said, Introspection has attributes that are specific to its defined structure, and should I care to retrieve attributes that are Introspection-specific, it's reasonable that the API would indiciate that I've crossed into Introspection-specific territory.

For example, if the user wanted to do `token.getClientId()`, the only way we'd be able to do that is through an Introspection-specific hierarchy. I don't see this as a common attribute to retrieve, but I agree that it's valuable for the sake of completeness, so I've created a ticket to add `IntrospectionClaimAccessor` that we can prioritize separately.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenWithClaims.java
jzheaux: This is resolved by pushing this down to a discussion about what is the principal in the authentication token, for which I have a ticket.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
jzheaux: Based on https://github.com/spring-projects/spring-security/issues/6498, I've made the claim names class package-private, so that the provider, the token, and their tests can all use the same constants without exposing this too early.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenWithClaims.java
jzheaux: This conversation is resolved with https://github.com/spring-projects/spring-security/issues/6498

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/IntrospectedOAuth2TokenAuthenticationToken.java
jzheaux: For consistency with other flows, let's keep with `sub`.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenAccessor.java
jzheaux: This is resolved via https://github.com/spring-projects/spring-security/issues/6498

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/IntrospectionClaimNames.java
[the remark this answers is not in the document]
jzheaux: This conversation is resolved via https://github.com/spring-projects/spring-security/issues/6498

samples/boot/oauth2resourceserver-opaque/src/main/resources/application.yml
jzheaux: Yes, I think discovery makes sense through the `issuer-uri` once it supports the `/.well-known/oauth-authorization-server` endpoint, which defines `introspection_endpoint` as an attribute.

As for the client, ideally, the user wouldn't need to specify the same client twice as I imagine that the 95% case is that they are the same. If there is only one client in client registration, it seems reasonable to infer that client is the same for the introspection endpoint.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
jgrandja: @rwinch When you get a chance can you provide feedback here.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
jgrandja: @rwinch When you get a chance can you provide feedback here.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
jgrandja: @rwinch When you get a chance can you provide feedback here.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
jgrandja: Spoke to @rwinch and we're good to leave this out now. Reason being is that there really is no dynamic parameters that we need since `BearerTokenAuthenticationToken` is the input and there isn't any context info we can extract other than the raw token at this point. 
**NOTE:** The `introspectionUri` constructor arg should support static request parameters.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
jgrandja: I spoke to @rwinch about this one and he also felt that being able to customize the response can be left out for now. As well, using Nimbus  API for response processing is consistent with what `ClientRegistrations` does today.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
jgrandja: Marking this as resolved and will address in #6422

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OAuth2IntrospectionAuthenticationProvider.java
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	private TokenIntrospectionSuccessResponse asSuccess(TokenIntrospectionResponse introspectionResponse) {
+		if (!introspectionResponse.indicatesSuccess()) {
+			throw new OAuth2AuthenticationException(invalidToken("Token introspection failed"));
+		}
+		return (TokenIntrospectionSuccessResponse) introspectionResponse;
+	}
+
+	private Map<String, Object> convertClaimsSet(TokenIntrospectionSuccessResponse response) {
```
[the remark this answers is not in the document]
jzheaux: They will not be `null` since this method simply updates the existing `Map`.

Nevertheless, agreed that it's better to have a test that confirms that all claims are correctly exposed as attributes, and I'll update the tests accordingly.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OAuth2IntrospectionAuthenticationProvider.java
```diff
@@ -0,0 +1,265 @@
...
+	@Override
+	public boolean supports(Class<?> authentication) {
+		return BearerTokenAuthenticationToken.class.isAssignableFrom(authentication);
+	}
+
+	private TokenIntrospectionSuccessResponse introspect(String token) {
+		HttpHeaders headers = new HttpHeaders();
+		headers.setAccept(Collections.singletonList(MediaType.APPLICATION_JSON_UTF8));
+		RequestEntity<String> requestEntity =
+				new RequestEntity<>("token=" + token, headers, HttpMethod.POST, this.introspectionUri);
```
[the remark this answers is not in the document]
jzheaux: For my own education, why is that better?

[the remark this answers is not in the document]
jzheaux: I asked @jgrandja offline, and he explained to me that `MultiValueMap` has the advantage that we aren't formatting the body here. 

So, instead of `"token=" + token`, which imposes a format to the body, we use an instance of `MultiValueMap` and allow `RestTemplate` to deal with that concern.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OpaqueTokenAuthenticationProvider.java
rwinch: It is being defensive because there is no guarantee someone called the supports method.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/OAuth2IntrospectionAuthenticationProvider.java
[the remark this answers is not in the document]
jgrandja: > They will not be `null` since this method simply updates the existing `Map`

Yes, somehow I missed this...`Map<String, Object> claims = response.toJSONObject()`

jgrandja:
Thanks @jzheaux. Good to go on my end.

As an FYI, it's best to not force push while in the review process as it's difficult to keep track of changes between review steps. I would just keep adding commits as you update based on feedback.

jzheaux:

PR #6363, opened 2019-01-08
Allow a entry to grant more than one bit of permission

Atry:


jzheaux:
@Atry, is this something that you could provide more description on and a unit test to confirm its behavior?

eleftherias:
@Atry could you add more details explaining what you are trying to accomplish, and add a unit test to confirm it is behaving as expected?

eleftherias:
Closing because the permission mask comparison can be customized as of a449d6c3.

Issue #6359, opened 2019-01-07
AbstractRequestMatcherRegistry should throw an exception when matchers are supplied after anyRequest


jzheaux:
The following configuration:

```java
http
    .authorizeRequests()
        .anyRequest().authenticated()
        .mvcMatchers("/myendpoint").permitAll();
```

is invalid since matchers are processed in the order they are declared. When a request comes in, regardless of its contents, `anyRequest` will always match it before the filter checks the `/myendpoint` matcher.

Having the matchers in this order is a bit like:

```java
if (true) { 
    // do this
} else if (request == '/myendpoint') {
    // do that
}
```

This is clearly erroneous and the code should fail fast so that the user can correct the misconfiguration.

If `anyRequest()` is invoked then no further calls to `antMatchers`, `mvcMatchers`, `regexMatchers` or `requestMatchers` should be allowed. For any further calls, we should throw an exception with a helpful error message.

ankurpathak:
Here is a pull request:
https://github.com/spring-projects/spring-security/pull/6365

PR #6365, opened 2019-01-08
No RequestMatcher After AnyRequest

ankurpathak:
Don't allow any type of RequestMatchers
after any request by throwing IllegalStateException

Fixes: gh-6359

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,6 +25,7 @@ import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
 import org.springframework.security.web.util.matcher.AnyRequestMatcher;
 import org.springframework.security.web.util.matcher.RegexRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher;
+import org.springframework.util.Assert;
 import org.springframework.web.servlet.handler.HandlerMappingIntrospector;
 
 import java.util.ArrayList;
@@ -39,6 +40,7 @@ import java.util.List;
  * @param <C> The object that is returned or Chained after creating the RequestMatcher
  *
  * @author Rob Winch
+ * @author Ankur Pathak
  * @since 3.2
  */
 public abstract class AbstractRequestMatcherRegistry<C> {
@@ -47,6 +49,8 @@ public abstract class AbstractRequestMatcherRegistry<C> {
 	private static final RequestMatcher ANY_REQUEST = AnyRequestMatcher.INSTANCE;
 
 	private ApplicationContext context;
+
+	private boolean anyRequestConfigured = false;
 
 	protected final void setApplicationContext(ApplicationContext context) {
 		this.context = context;
@@ -67,7 +71,10 @@ public abstract class AbstractRequestMatcherRegistry<C> {
 	 * @return the object that is chained after creating the {@link RequestMatcher}
 	 */
 	public C anyRequest() {
-		return requestMatchers(ANY_REQUEST);
+		Assert.state(!this.anyRequestConfigured, "Can't configure anyRequest after itself");
+		C configurer = requestMatchers(ANY_REQUEST);
+		this.anyRequestConfigured = true;
+		return configurer;
 	}
 
 	/**
@@ -97,6 +104,7 @@ public abstract class AbstractRequestMatcherRegistry<C> {
 	 * @return the object that is chained after creating the {@link RequestMatcher}
 	 */
 	public C antMatchers(HttpMethod method, String... antPatterns) {
+		Assert.state(!this.anyRequestConfigured, "Can't configure antMatchers after anyRequest");
 		return chainRequestMatchers(RequestMatchers.antMatchers(method, antPatterns));
 	}
 
@@ -111,6 +119,7 @@ public abstract class AbstractRequestMatcherRegistry<C> {
 	 * @return the object that is chained after creating the {@link RequestMatcher}
 	 */
 	public C antMatchers(String... antPatterns) {
+		Assert.state(!this.anyRequestConfigured, "Can't configure antMatchers after anyRequest");
 		return chainRequestMatchers(RequestMatchers.antMatchers(antPatterns));
 	}
 
@@ -160,7 +169,7 @@ public abstract class AbstractRequestMatcherRegistry<C> {
 	 */
 	protected final List<MvcRequestMatcher> createMvcMatchers(HttpMethod method,
 			String... mvcPatterns) {
-
+		Assert.state(!this.anyRequestConfigured, "Can't configure mvcMatchers after anyRequest");
 		ObjectPostProcessor<Object> opp = this.context.getBean(ObjectPostProcessor.class);
 		if (!this.context.containsBean(HANDLER_MAPPING_INTROSPECTOR_BEAN_NAME)) {
 			throw new NoSuchBeanDefinitionException("A Bean named " + HANDLER_MAPPING_INTROSPECTOR_BEAN_NAME +" of type " + HandlerMappingIntrospector.class.getName()
@@ -195,6 +204,7 @@ public abstract class AbstractRequestMatcherRegistry<C> {
 	 * @return the object that is chained after creating the {@link RequestMatcher}
 	 */
 	public C regexMatchers(HttpMethod method, String... regexPatterns) {
+		Assert.state(!this.anyRequestConfigured, "Can't configure regexMatchers after anyRequest");
 		return chainRequestMatchers(RequestMatchers.regexMatchers(method, regexPatterns));
 	}
 
@@ -209,6 +219,7 @@ public abstract class AbstractRequestMatcherRegistry<C> {
 	 * @return the object that is chained after creating the {@link RequestMatcher}
 	 */
 	public C regexMatchers(String... regexPatterns) {
+		Assert.state(!this.anyRequestConfigured, "Can't configure regexMatchers after anyRequest");
 		return chainRequestMatchers(RequestMatchers.regexMatchers(regexPatterns));
 	}
 
@@ -221,6 +232,7 @@ public abstract class AbstractRequestMatcherRegistry<C> {
 	 * @return the object that is chained after creating the {@link RequestMatcher}
 	 */
 	public C requestMatchers(RequestMatcher... requestMatchers) {
+		Assert.state(!this.anyRequestConfigured, "Can't configure requestMatchers after anyRequest");
 		return chainRequestMatchers(Arrays.asList(requestMatchers));
 	}
 
--- a/config/src/test/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistryAnyMatcherTests.java
+++ b/config/src/test/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistryAnyMatcherTests.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.config.annotation.web;
+
+import org.junit.Test;
+
+import org.springframework.beans.factory.BeanCreationException;
+import org.springframework.mock.web.MockServletContext;
+import org.springframework.security.config.annotation.web.builders.HttpSecurity;
+import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
+import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
+import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
+import org.springframework.web.context.support.AnnotationConfigWebApplicationContext;
+
+/**
+ * Tests for {@link AbstractRequestMatcherRegistry}.
+ *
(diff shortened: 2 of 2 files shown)


akuma8:
Thanks for this PR. You can also consider some cases like this:
```
http
    .authorizeRequests()
        .antMatchers("/demo/**").permitAll()
        .antMatchers("/demo/test").authenticated();
```
I don't know the level of difficulty to treat all cases.

rwinch:
@akuma8 Thanks for the suggestion, but this is too complicated to detect.

config/src/test/groovy/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationsTests.groovy
```diff
@@ -451,6 +452,94 @@ public class ExpressionUrlAuthorizationConfigurerTests extends BaseSpringSpec {
 		}
 	}
 
+	@EnableWebSecurity
```
jzheaux: Would you mind placing these new tests in a Java class, like `ExpressionUrlAuthorizationConfigurerAnyMatcherTests`? We're in the process of migrating all Groovy tests to Java.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
```diff
@@ -108,6 +109,57 @@ public ExpressionInterceptUrlRegistry getRegistry() {
 			extends
 			ExpressionUrlAuthorizationConfigurer<H>.AbstractInterceptUrlRegistry<ExpressionInterceptUrlRegistry, AuthorizedUrl> {
 
+		private boolean isAnyRequestConfigured = false;
```
jzheaux: Is there a reason to add this to `ExpressionUrlAUthorizationConfigurer` instead of `AbstractRequestMatcherRegistry`? I think it would be nice if all subclasses of `AbstractRequestMatcherRegistry` were able to inherit this behavior. What did I miss?

jzheaux: If it's due to `mvcMatchers`, I think that you could add the check to `AbstractRequestMatcherRegistry#createMvcMatchers`.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
```diff
@@ -108,6 +109,57 @@ public ExpressionInterceptUrlRegistry getRegistry() {
...
+
+		@Override
+		public AuthorizedUrl anyRequest() {
+			if (isAnyRequestConfigured) {
```
jzheaux: Also, would you please prefix member variables with `this.`? That way it's easy to differentiate between member and local variables.

jzheaux: Let's use `Assert.state(!this.anyRequestConfigured, "Can't configure anyRequest after itself")` sinces its a bit more compact

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
jzheaux: Also, let's leave `is` off the front and call it `anyRequestConfigured` instead.

jzheaux:
Thanks for the PR, @ankurpathak! I've left some comments inline.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
ankurpathak: > Is there a reason to add this to `ExpressionUrlAUthorizationConfigurer` instead of `AbstractRequestMatcherRegistry`? I think it would be nice if all subclasses of `AbstractRequestMatcherRegistry` were able to inherit this behavior. What did I miss?

Fixed.

ankurpathak: > If it's due to `mvcMatchers`, I think that you could add the check to `AbstractRequestMatcherRegistry#createMvcMatchers`.

Fixed.

ankurpathak: > Also, let's leave `is` off the front and call it `anyRequestConfigured` instead.

Fixed

config/src/test/groovy/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationsTests.groovy
ankurpathak: > Would you mind placing these new tests in a Java class, like `ExpressionUrlAuthorizationConfigurerAnyMatcherTests`? We're in the process of migrating all Groovy tests to Java.

Fixed

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
ankurpathak: > Let's use `Assert.state(!this.anyRequestConfigured, "Can't configure anyRequest after itself")` sinces its a bit more compact

Replaced IllegalStateException use by Assert.

ankurpathak: > Also, would you please prefix member variables with `this.`? That way it's easy to differentiate between member and local variables.

Migrated to follow this convention.

config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java
```diff
@@ -67,6 +72,10 @@ protected final ApplicationContext getApplicationContext() {
 	 * @return the object that is chained after creating the {@link RequestMatcher}
 	 */
 	public C anyRequest() {
+		if (this.anyRequestConfigured) {
```
jzheaux: Sorry, I didn't explain this well. `Assert.state` will do the if check for you as well, so we can remove that, too.

config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java
```diff
@@ -25,11 +25,13 @@
 import org.springframework.security.web.util.matcher.AnyRequestMatcher;
 import org.springframework.security.web.util.matcher.RegexRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher;
+import org.springframework.util.Assert;
```
jzheaux: Can we change the copyright date to any class you modify or create in this PR to `2019` instead of `2018`?

config/src/test/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistryAnyMatcherTests.java
```diff
@@ -0,0 +1,120 @@
...
+ * Tests for {@link AbstractRequestMatcherRegistry}.
+ *
+ * @author Ankur Pathak
+ */
+public class AbstractRequestMatcherRegistryAnyMatcherTests{
+
+	@EnableWebSecurity
+	static class AntMatchersAfterAnyRequestConfig extends WebSecurityConfigurerAdapter {
+		protected void configure(HttpSecurity http) throws Exception {
+			http
```
jzheaux: Will you please format these like this:

```java
http
    .authorizeRequests()
        .anyRequest().authenticated()
        .antMatchers("/demo/**").permitAll()
```

jzheaux: Generally, I mean
```java
http
    .authorizeRequests()
        .someMatcher
        .anotherMatcher
```
for any of the others in the test. Thanks!

config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java
ankurpathak: > Sorry, I didn't explain this well. `Assert.state` will do the if check for you as well, so we can remove that, too.

Fixed.

config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java
ankurpathak: > Can we change the copyright date to any class you modify or create in this PR to `2019` instead of `2018`?

Fixed.

config/src/test/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistryAnyMatcherTests.java
ankurpathak: > Generally, I mean
> 
> ```java
> http
>     .authorizeRequests()
>         .someMatcher
>         .anotherMatcher
> ```
> 
> for any of the others in the test. Thanks!

I did this also. But please confirm if its ok or what??

config/src/test/groovy/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationsTests.groovy
```diff
@@ -43,6 +38,12 @@ import org.springframework.security.config.annotation.web.configurers.Expression
 import org.springframework.security.core.Authentication
 import org.springframework.security.core.authority.AuthorityUtils
 import org.springframework.security.web.access.intercept.FilterSecurityInterceptor
+import org.springframework.security.authentication.UsernamePasswordAuthenticationToken
```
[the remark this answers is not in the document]
ankurpathak: Are you sure on this? Beacuse its present even befor my PR. I have not added it.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
```diff
@@ -87,7 +87,6 @@
 	private static final String authenticated = "authenticated";
 	private static final String fullyAuthenticated = "fullyAuthenticated";
 	private static final String rememberMe = "rememberMe";
-
```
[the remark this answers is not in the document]
ankurpathak: Are you sure on this? Its a public API and how we can remove it like this. Its getting used in many places. Beacuse its present even before my PR. I have not added it.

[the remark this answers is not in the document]
ankurpathak: screenshot from 2019-01-20 02-19-40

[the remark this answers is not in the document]
ankurpathak: Sorry  I didn't understand first. But I hv removed this file from commit.

config/src/test/groovy/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationsTests.groovy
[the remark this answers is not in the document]
ankurpathak: Sorry  I didn't understand first. But I hv removed this file from commit.

config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java
```diff
@@ -221,6 +232,9 @@ public C regexMatchers(String... regexPatterns) {
 	 * @return the object that is chained after creating the {@link RequestMatcher}
 	 */
 	public C requestMatchers(RequestMatcher... requestMatchers) {
+		if (!Objects.equals(requestMatchers[0], AnyRequestMatcher.INSTANCE)){
```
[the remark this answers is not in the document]
ankurpathak: I have added that Objets.equals if condition and Assert.state in anyRequest to consider case of
```anyRequest().anyRequest()
```
But you are right this will break many legimate cases.
So lets leave this case and I am revoving those both ifs.
Cheers.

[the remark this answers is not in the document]
ankurpathak:
```
anyRequest().anyRequest()
```
This case is also working by Assert in .requestMatchers().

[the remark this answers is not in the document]
jzheaux: @ankurpathak, this approach won't work because it changes the inheritance behavior. For example, if I had overridden `requestMatchers`, I may be expecting `anyMatcher` to call it (since it originally was). `anyMatcher` needs to continue to call `requestMatchers` in order to remain backwards compatible.

ankurpathak:
@jzheaux Backward compatability of any method is only related to its Signature and Behaviour not its Implementation. Implementation of method can change with time for various improvements, Will the method becomes Backward Incompatible with improvements??

jzheaux:
@ankurpathak Does this not work:

```java
public C anyRequest() {
    Assert.state(!this.anyRequestConfigured, "Can't configure anyRequest after itself");
    C configurer = requestMatchers(ANY_REQUEST);
    this.anyRequestConfigured = true;
    return configurer;
}
```

ankurpathak:
> @ankurpathak This is what I suggested earlier:
> 
> ```java
> public C anyRequest() {
>     Assert.state(!this.anyRequestConfigured, "Can't configure anyRequest after itself");
>     C configurer = requestMatchers(ANY_REQUEST);
>     this.anyRequestConfigured = true;
>     return configurer;
> }
> ```
> 
> instead of calling `chainRequestMatchers` directly. I'm sorry that wasn't clear in my inline feedback. Does this implementation not work for some reason?
> 
> > I hardly see any point in call of anyRequest to acutal method going through requestMatcher if the call can go directly.
> 
> The point is that because `requestMatchers` is overridable, then the methods that call it should continue to do so. Of course, if we can't, then we can't and we move on.
> 
> But, I believe my suggestion above provides the same feature while at the same time maintaining passivity. If a user is overriding `requestMatchers`, then calling `anyMatcher` will continue to work for them as they have designed. Do you disagree that this is a valuable thing?

I think it will work. Becoz we are setting flag after calling requestmatcher, I am sorry I was mising this line. I am working with your version. I will get back to you soon once I am done.

ankurpathak:
@jzheaux Now it is reflecting your version of anyRequest.

jzheaux:
Nice work, @ankurpathak! This has now been merged into `master`.

Issue #6341, opened 2018-12-29
Webflux Oauth2 .oauth2Client() doesn't redirect back to the original request after authenticating in the auth server


rozagerardo:
### Summary
When configuring a WebFlux application as an OAuth2 Client using an authentication_code grant type and without implementing the OAuth2Login feature, the application redirects to the "/" path after authenticating in the Authentication Server instead of redirecting back to the original request

### Actual Behavior

1-Calling an endpoint in an OAuth2 Client application (without using OAuth2 login) using Auth code Grant type
2- Authenticate in the Authorization Server
3- Get redirected to the /authorize/oauth2/code/[myclient] endpoint
4- Get redirected to the root ("/") URL

### Expected Behavior

1-Calling an endpoint in an OAuth2 Client application (without using OAuth2 login) using Auth code Grant type
2- Authenticate in the Authorization Server
3- Get redirected to the /authorize/oauth2/code/[myclient] endpoint
**4- Get redirected to the endpoint we called in the first place**

### Configuration

1- Set up an application with the following using the following ServerHttpSecurity configuration:
```
@Bean
public SecurityWebFilterChain springSecurityFilterChain(ServerHttpSecurity http) {
    http.authorizeExchange()
        .anyExchange()
        .permitAll()
        .and()
        .oauth2Client();
    return http.build();
}
```
2- A client registration:
```
spring.security.oauth2.client.registration.myclient.client-name=myclient
spring.security.oauth2.client.registration.myclient.client-id=myclient-client-id
spring.security.oauth2.client.registration.myclient.client-secret=myclient-secret
spring.security.oauth2.client.registration.myclient.authorization-grant-type=authorization_code
spring.security.oauth2.client.registration.myclient.redirect-uri-template=http://localhost:8080/authorize/oauth2/code/myclient

spring.security.oauth2.client.provider.myclient.token-uri=http://localhost:8085/oauth/token
spring.security.oauth2.client.provider.myclient.authorization-uri=http://localhost:8085/oauth/authorize
```
Note: In my case, I set up the Client registration using Spring Boot 2.x. The client is registered in a custom Authentication Provider that I configured using Spring Security Oauth, but the issue should be present for well-known providers as well.

3- Configure the WebClient:
```
@Bean
    WebClient webClient(ReactiveClientRegistrationRepository clientRegistrations, ServerOAuth2AuthorizedClientRepository authorizedClients) {
[...]

PR #6418, opened 2019-01-13
Save original request on oauth2Client filter

rozagerardo:
When we used the oauth2Client directive and requested an endpoint that
required client authorization on the authorization server, the
SPRING_SECURITY_SAVED_REQUEST was not persisted, and therefore after
creating the authorized client we were redirected to the root page ("/").

Now we are storing the session attribute and getting redirected back to
the original URI as expected.

Note that the attribute is stored only when a
ClientAuthorizationRequiredException is thrown in the chain, we don't
want to store it as a response to the
/oauth2/authorization/{registrationId} endpoint, since we would end
up in an infinite loop

Fixes related Issue: #6341

### DIFF
--- a/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/server/OAuth2AuthorizationRequestRedirectWebFilterTests.java
+++ b/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/server/OAuth2AuthorizationRequestRedirectWebFilterTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,6 +26,7 @@ import org.springframework.security.oauth2.client.registration.ClientRegistratio
 import org.springframework.security.oauth2.client.registration.ReactiveClientRegistrationRepository;
 import org.springframework.security.oauth2.client.registration.TestClientRegistrations;
 import org.springframework.security.oauth2.core.endpoint.OAuth2AuthorizationRequest;
+import org.springframework.security.web.server.savedrequest.ServerRequestCache;
 import org.springframework.test.web.reactive.server.FluxExchangeResult;
 import org.springframework.test.web.reactive.server.WebTestClient;
 import org.springframework.web.server.handler.FilteringWebHandler;
@@ -52,6 +53,9 @@ public class OAuth2AuthorizationRequestRedirectWebFilterTests {
 
 	@Mock
 	private ServerAuthorizationRequestRepository<OAuth2AuthorizationRequest> authzRequestRepository;
+
+	@Mock
+	private ServerRequestCache requestCache;
 
 	private ClientRegistration registration = TestClientRegistrations.clientRegistration().build();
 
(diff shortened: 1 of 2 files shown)


rozagerardo:
Hey Rob thanks for taking the time to review the PR and for the comment
You mean a test not just mocking and verifying that the `ServerRequestCache.saveRequest` method is called, but to actually allow it save the request in the session, and validate that it was correctly stored?
I now added a new test to to that (`filterWhenHtmlRequestWithExceptionThenRedirectedWithSavedRequestSessionAttribute`), feel free to have a look, and let me know if it's ok :)
@rwinch

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/server/OAuth2AuthorizationRequestRedirectWebFilterTests.java
```diff
@@ -53,6 +60,15 @@
 	@Mock
 	private ServerAuthorizationRequestRepository<OAuth2AuthorizationRequest> authzRequestRepository;
 
+	@Mock
+	private ServerRequestCache requestCache;
+
+	@Spy
+	private WebSessionServerRequestCache spiedRequestCache;
```
rwinch: Let's change this to a `ServerRequestCache` and mock out the methods. This ensures the test isn't relying on `WebSessionServerRequestCache` behavior

rozagerardo: Sorry for the comings and goings here @rwinch , but I'd like to make sure we're on the same page.
Wouldn't that be what I first did on the [filterWhenExceptionThenRedirected](https://github.com/spring-projects/spring-security/pull/6418/files#diff-6eb7b3bb0614ba3228ea67da35a04b9aL132) test?
I first chose to modify that test instead of creating a new one, since that will be the actual scenario that we want to test, i.e. that when we receive a `ClientAuthorizationRequiredException`, we redirect and save the original request in the process.
I can create a new test hat verifies that the `ServerRequestCache` is used, but it will be pretty much the same test that the one I mentioned above, wouldn't it? Please let me know if I'm missing something here :). I'll clean everything once I understand your perspective correctly, sounds good?

rwinch: Yes it is quite similar. However, please use a Mock instead of a Spy. With the Spy we are relying on behavior from `WebSessionServerRequestCache` to see if things are working. For example, the assertion is checking to see if a session attribute has been set or not. This means that this test requires `WebSessionServerRequestCache` sets that attribute. All we do care is that the save method is invoked with the correct arguments.

rozagerardo: Of course, I agree that we don't want an integration test here.
Let me try to explain better, I now removed the integration test and added a *new* test. I believe that's the test that you had in mind, correct? 
My question is, does this test bring any value to the table? It's exactly the same scenario that for the [filterWhenExceptionThenRedirected](https://github.com/spring-projects/spring-security/pull/6418/files#diff-6eb7b3bb0614ba3228ea67da35a04b9aL132) isn't it?
That's why I considered that we don't need a new test, but just adding the validation to the pre existing test would be enough, but please correct me if I'm wrong, I'm probably missing something here :)

rwinch:
Thanks again for the PR @rozagerardo! This is now merged into master

Issue #6417, opened 2019-01-12
Invalid html in default login page


bestwmm:
### Summary
Invalid html in spring-security/web/src/main/java/org/springframework/security/web/authentication/defaultLoginPageGeneratingFilter.java::generateLoginPageHtml.

### Actual Behavior
div class="container" at line 237 is not closed.  Closing div required around line 291 before "</body>".

Page does not validate, but Chrome gracioiusly display the page correctly.

### Expected Behavior
Pass validation test.

### Configuration
n/a

### Version
org.spring.framework.boot,l artifactID spring-boot-starter-security version 2.1.0.RELEASE

rwinch:
Thanks for the report @bestwmm! Would you be interested in submitting a PR for this?

bestwmm:
What’s a PR?  (Pull Request + submission with fix?)

rwinch:
Yes. If you are wanting to learn more, GitHub has more information on what a [Pull Request (PR)](https://help.github.com/articles/about-pull-requests/) is. If you are unable to help, that is fine, but your help is always welcome :)

ankurpathak:
@rwinch I would like to take this one.

rwinch:
Thanks @ankurpathak! The issue is yours.

@bestwmm If you are wanting to help out with another issue, please use the Help Wanted tag to find issues we feel are good for contributions.

bestwmm:
> Here is pull request for issue:
> #6448

Thank you.

PR #6448, opened 2019-01-17
HTML markup fixed in DefaultLoginPageGeneratingFilter

ankurpathak:
Ending div moved  out of condition.

Fies: gh-6417


rwinch:
Thanks for the quick fix @ankurpathak! This is now merged into master and backported via gh-6468

PR #6479, opened 2019-01-24
Missing spring: prefix on jwk-set-uri example

xyloman:
Debugging an issue today because our jwk properties where not working and discovered that the doc [here](https://docs.spring.io/spring-security/site/docs/current/reference/htmlsingle/#oauth2resourceserver-jwkseturi) should be `spring.security.oauth2.resourceserver.jwt.jwk-set-uri` to align with the spring boot autoconfiguration [here](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/resource/reactive/ReactiveOAuth2ResourceServerJwkConfiguration.java#L46)


rwinch:
Thanks for the PR @xyloman! This is now merged into master and backported via gh-6481

Issue #6446, opened 2019-01-15
Support PKCE for Client


jgrandja:
We should add client support for [PKCE](https://tools.ietf.org/html/rfc7636).

Related #4943

sdoxsee:
Hi @jgrandja. Added a PR but it's definitely not "ready-to-go". I've added some questions to the PR that can be discussed. Thanks.

jgrandja:
Thanks @sdoxsee ! I will get to this within the next 2 days. Just need to address a couple other issues first.

Issue #5652, opened 2018-08-07
Client Secret should not be mandatory


gburboz:
### Summary
As per OIDC spec [9. ClientAuthentication](https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication) one of the client authentication method is `none` hence we should not have client registration's `client-secret` as mandatory

### Actual Behavior

Current code validates client secret at following two locations and causes exception 

*OAuth2ClientProperties.validateRegistration()*
```
Caused by: java.lang.IllegalStateException: Client secret must not be empty.
	at org.springframework.boot.autoconfigure.security.oauth2.client.OAuth2ClientProperties.validateRegistration(OAuth2ClientProperties.java:65) ~[spring-boot-autoconfigure-2.0.4.RELEASE.jar:2.0.4.RELEASE]
```

*ClientRegistration.validateAuthorizationCodeGrantType()*
```
Caused by: java.lang.IllegalArgumentException: clientSecret cannot be empty
	at org.springframework.util.Assert.hasText(Assert.java:276) ~[spring-core-5.0.8.RELEASE.jar:5.0.8.RELEASE]
	at org.springframework.security.oauth2.client.registration.ClientRegistration$Builder.validateAuthorizationCodeGrantType(ClientRegistration.java:437) ~[spring-security-oauth2-client-5.0.7.RELEASE.jar:5.0.7.RELEASE]
```

### Expected Behavior
Above validations must be removed and `clientId` can be populated as query parameter without any client secret (when `clientAuthenticationMethod` is `none`). 

### Version
Spring Security 5

jgrandja:
@gburboz Client secret is manadatory for **Confidential Clients**, which is the case for clients configured for the `authorization_code` grant. It's not mandatory for clients configured for the `implicit` grant.

> As per OIDC spec 9. ClientAuthentication one of the client authentication method is none hence we should not have client registration's client-secret as mandatory

As you can see, the `none` client authentication method is for **Public Clients**:

> none
The Client does not authenticate itself at the Token Endpoint, either because it uses only the **Implicit Flow** (and so does not use the Token Endpoint) or because it is a **Public Client** with no Client Secret or other authentication mechanism.

I'm going to close this issue as the implementation is correct.
Hope this explains things?

gburboz:
@jgrandja , from specification, my interpretation of `none`, implies either one of the below and not just first one

1.  `Implicit Flow`
2.  OR `Public Client with no Client Secret`
[...]

Issue #5940, opened 2018-10-11
OAuth2AuthorizationRequest.additionalParameters should not contain registration_id


jgrandja:
When `DefaultOAuth2AuthorizationRequestResolver` creates an `OAuth2AuthorizationRequest` it saves the `clientRegistration.getRegistrationId()` as follows:

```java
		Map<String, Object> additionalParameters = new HashMap<>();
		additionalParameters.put(OAuth2ParameterNames.REGISTRATION_ID, clientRegistration.getRegistrationId());
```

Then `OAuth2LoginAuthenticationFilter` retrieves the `registrationId` via `OAuth2AuthorizationRequest.additionalParameters()` in order to lookup the `ClientRegistration` using `ClientRegistrationRepository` during Authorization Response processing.

We should avoid storing the `registrationId` in `OAuth2AuthorizationRequest.additionalParameters()` as this may cause issues when a user configures their own `OAuth2AuthorizationRequestResolver` and overrides `additionalParameters` with their custom request parameters and does not include the `registrationId`.

rwinch:
NOTE: To whom fixes this issue please check the reactive equivalent when this issue is resolved

jgrandja:
We should consider introducing `OAuth2AuthorizationRequest.attributes` (of type `Map<String, Object>`), which would serve the similar purpose as `ServletRequest.getAttribute()` and `RequestAttributes`. The `registrationId` would than be stored in `OAuth2AuthorizationRequest.attributes`.

This makes a lot of sense as I'm anticipating we'll need a place to store contextual information associated to an Authorization Request while it's in flight during a grant flow. And we should not mix request parameters with request context attributes.

PR #6485, opened 2019-01-28
Add client support for PKCE

sdoxsee:
This is a PR for #6446 that involves simple extensions to existing spring security that would probably be used by spring desktop apps (swing, javafx, etc.). It is a WIP but wanted to get some early feedback. I made some tweaks to the sample app oauth2webclient just to illustrate how this can be used with spring boot. I omitted the `client-secret` and added `client-authentication-method: none`. Thanks!

Took the discussion in #5652 into account.

Here's a brief overview of the use-cases for client-side PKCE as it pertains to spring security (ref #6446):

1. **Native/Mobile**

    Of course, only JVM-based Mobile OS's (Android). Android uses a "Custom URI scheme" that is registered with the OS to capture the `code`. I'd imagine that you'd probably use a library like https://github.com/openid/AppAuth-Android rather than spring security? It appears there was an effort at one time to connect Spring and Android (https://github.com/spring-projects/spring-android) but the project now seems quite stale (dead?).  In theory, you could probably have a webserver embedded in your native app that, if it used spring security, could handle the code callback but that seems heavy to me. 

    **Bottom line**: I don't think this is the target use-case for spring security client PKCE support but could be used in theory.

2. **Desktop**

    JVM desktop apps with spring exist (we have one!). We'd need some sort of loopback webserver to capture the authorization `code`. This could be a very simple one that spins up, on demand, serves a certain port and then shuts down--but that requires custom code (and could be done at a later time). If I simply make our non-web app a web one (by commenting out `main.web-application-type: none`), we'd get a tomcat and spring security auto configuration just like other web applications. The difference here is that the app needs to be delivered to the users own machines--thereby making the client_secret unviable and PKCE a good fit. Turning non-web apps into web ones feels a bit heavy but simple.

    I think using refresh tokens in spring security clients is safe because it's in memory rather than disk?

    **Bottom line**: Simpler is better, desktops have the resources for a local webserver. Why not leverage spring-boot's embedded server? Any other lighter-weight suggestions?

3. **Client-side browser-based JavaScript clients (SPA)**

    These authorization request are typically all javascript-initiated. The authorization `code` comes back to client and client makes the token call itself. I figure that spring security is necessary. Question: Why are implicit calls created in spring security (e.g. OAuth2AuthorizationRequest.implicit())?

    **Bottom** line: Why is spring supporting this category at all? We probably don't need PKCE support.

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -500,6 +500,11 @@ public final class ClientRegistration implements Serializable {
 			clientRegistration.clientId = this.clientId;
 			clientRegistration.clientSecret = StringUtils.hasText(this.clientSecret) ? this.clientSecret : "";
 			clientRegistration.clientAuthenticationMethod = this.clientAuthenticationMethod;
+			if (AuthorizationGrantType.AUTHORIZATION_CODE.equals(this.authorizationGrantType) &&
+					!StringUtils.hasText(this.clientSecret)) {
+				clientRegistration.clientAuthenticationMethod = ClientAuthenticationMethod.NONE;
+			}
+
 			clientRegistration.authorizationGrantType = this.authorizationGrantType;
 			clientRegistration.redirectUriTemplate = this.redirectUriTemplate;
 			clientRegistration.scopes = this.scopes;
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolver.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolver.java
@@ -20,14 +20,19 @@ import org.springframework.security.crypto.keygen.StringKeyGenerator;
 import org.springframework.security.oauth2.client.registration.ClientRegistration;
 import org.springframework.security.oauth2.client.registration.ClientRegistrationRepository;
 import org.springframework.security.oauth2.core.AuthorizationGrantType;
+import org.springframework.security.oauth2.core.ClientAuthenticationMethod;
 import org.springframework.security.oauth2.core.endpoint.OAuth2AuthorizationRequest;
 import org.springframework.security.oauth2.core.endpoint.OAuth2ParameterNames;
+import org.springframework.security.oauth2.core.endpoint.PkceParameterNames;
 import org.springframework.security.web.util.UrlUtils;
 import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
 import org.springframework.util.Assert;
 import org.springframework.web.util.UriComponentsBuilder;
 
 import javax.servlet.http.HttpServletRequest;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.Base64;
 import java.util.HashMap;
 import java.util.Map;
@@ -52,6 +57,7 @@ public final class DefaultOAuth2AuthorizationRequestResolver implements OAuth2Au
 	private final ClientRegistrationRepository clientRegistrationRepository;
 	private final AntPathRequestMatcher authorizationRequestMatcher;
 	private final StringKeyGenerator stateGenerator = new Base64StringKeyGenerator(Base64.getUrlEncoder());
+	private final StringKeyGenerator codeVerifierGenerator = new Base64StringKeyGenerator(Base64.getUrlEncoder().withoutPadding(), 96);
 
 	/**
 	 * Constructs a {@code DefaultOAuth2AuthorizationRequestResolver} using the provided parameters.
@@ -102,9 +108,17 @@ public final class DefaultOAuth2AuthorizationRequestResolver implements OAuth2Au
 			throw new IllegalArgumentException("Invalid Client Registration with Id: " + registrationId);
 		}
 
+		Map<String, Object> attributes = new HashMap<>();
+		attributes.put(OAuth2ParameterNames.REGISTRATION_ID, clientRegistration.getRegistrationId());
+
 		OAuth2AuthorizationRequest.Builder builder;
 		if (AuthorizationGrantType.AUTHORIZATION_CODE.equals(clientRegistration.getAuthorizationGrantType())) {
 			builder = OAuth2AuthorizationRequest.authorizationCode();
+			if (ClientAuthenticationMethod.NONE.equals(clientRegistration.getClientAuthenticationMethod())) {
+				Map<String, Object> additionalParameters = new HashMap<>();
+				addPkceParameters(attributes, additionalParameters);
+				builder.additionalParameters(additionalParameters);
+			}
 		} else if (AuthorizationGrantType.IMPLICIT.equals(clientRegistration.getAuthorizationGrantType())) {
 			builder = OAuth2AuthorizationRequest.implicit();
 		} else {
@@ -114,9 +128,6 @@ public final class DefaultOAuth2AuthorizationRequestResolver implements OAuth2Au
 		}
 
 		String redirectUriStr = this.expandRedirectUri(request, clientRegistration, redirectUriAction);
-
-		Map<String, Object> attributes = new HashMap<>();
-		attributes.put(OAuth2ParameterNames.REGISTRATION_ID, clientRegistration.getRegistrationId());
 
 		OAuth2AuthorizationRequest authorizationRequest = builder
 				.clientId(clientRegistration.getClientId())
@@ -155,5 +166,35 @@ public final class DefaultOAuth2AuthorizationRequestResolver implements OAuth2Au
 		return UriComponentsBuilder.fromUriString(clientRegistration.getRedirectUriTemplate())
 				.buildAndExpand(uriVariables)
 				.toUriString();
+	}
+
+	/**
+	 * Creates and adds additional PKCE parameters for use in the OAuth 2.0 Authorization and Access Token Requests
+	 *
+	 * @param attributes where {@link PkceParameterNames#CODE_VERIFIER} is stored for the token request
+	 * @param additionalParameters where {@link PkceParameterNames#CODE_CHALLENGE} and, usually,
+	 * {@link PkceParameterNames#CODE_CHALLENGE_METHOD} are added to be used in the authorization request.
+	 *
+	 * @since 5.2
+	 * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7636#section-1.1">1.1.  Protocol Flow</a>
+	 * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7636#section-4.1">4.1.  Client Creates a Code Verifier</a>
+	 * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7636#section-4.2">4.2.  Client Creates the Code Challenge</a>
+	 */
+	private void addPkceParameters(Map<String, Object> attributes, Map<String, Object> additionalParameters) {
+		String codeVerifier = codeVerifierGenerator.generateKey();
+		attributes.put(PkceParameterNames.CODE_VERIFIER, codeVerifier);
+		try {
+			String codeChallenge = createCodeChallenge(codeVerifier);
+			additionalParameters.put(PkceParameterNames.CODE_CHALLENGE, codeChallenge);
+			additionalParameters.put(PkceParameterNames.CODE_CHALLENGE_METHOD, "S256");
+		} catch (NoSuchAlgorithmException e) {
+			additionalParameters.put(PkceParameterNames.CODE_CHALLENGE, codeVerifier);
+		}
+	}
+
+	private String createCodeChallenge(String codeVerifier) throws NoSuchAlgorithmException {
+		MessageDigest md = MessageDigest.getInstance("SHA-256");
+		byte[] digest = md.digest(codeVerifier.getBytes(StandardCharsets.US_ASCII));
+		return Base64.getUrlEncoder().withoutPadding().encodeToString(digest);
 	}
 }
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/DefaultServerOAuth2AuthorizationRequestResolver.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/DefaultServerOAuth2AuthorizationRequestResolver.java
@@ -24,8 +24,10 @@ import org.springframework.security.crypto.keygen.StringKeyGenerator;
 import org.springframework.security.oauth2.client.registration.ClientRegistration;
 import org.springframework.security.oauth2.client.registration.ReactiveClientRegistrationRepository;
 import org.springframework.security.oauth2.core.AuthorizationGrantType;
+import org.springframework.security.oauth2.core.ClientAuthenticationMethod;
 import org.springframework.security.oauth2.core.endpoint.OAuth2AuthorizationRequest;
 import org.springframework.security.oauth2.core.endpoint.OAuth2ParameterNames;
+import org.springframework.security.oauth2.core.endpoint.PkceParameterNames;
 import org.springframework.security.web.server.util.matcher.PathPatternParserServerWebExchangeMatcher;
 import org.springframework.security.web.server.util.matcher.ServerWebExchangeMatcher;
 import org.springframework.util.Assert;
@@ -34,6 +36,9 @@ import org.springframework.web.server.ServerWebExchange;
 import org.springframework.web.util.UriComponentsBuilder;
 import reactor.core.publisher.Mono;
 
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.Base64;
 import java.util.HashMap;
 import java.util.Map;
@@ -67,6 +72,8 @@ public class DefaultServerOAuth2AuthorizationRequestResolver
 	private final ReactiveClientRegistrationRepository clientRegistrationRepository;
 
 	private final StringKeyGenerator stateGenerator = new Base64StringKeyGenerator(Base64.getUrlEncoder());
+
+	private final StringKeyGenerator codeVerifierGenerator = new Base64StringKeyGenerator(Base64.getUrlEncoder().withoutPadding(), 96);
 
 	/**
 	 * Creates a new instance
@@ -124,6 +131,11 @@ public class DefaultServerOAuth2AuthorizationRequestResolver
 		OAuth2AuthorizationRequest.Builder builder;
 		if (AuthorizationGrantType.AUTHORIZATION_CODE.equals(clientRegistration.getAuthorizationGrantType())) {
 			builder = OAuth2AuthorizationRequest.authorizationCode();
+			if (ClientAuthenticationMethod.NONE.equals(clientRegistration.getClientAuthenticationMethod())) {
+				Map<String, Object> additionalParameters = new HashMap<>();
+				addPkceParameters(attributes, additionalParameters);
+				builder.additionalParameters(additionalParameters);
+			}
 		}
 		else if (AuthorizationGrantType.IMPLICIT.equals(clientRegistration.getAuthorizationGrantType())) {
 			builder = OAuth2AuthorizationRequest.implicit();
@@ -163,5 +175,35 @@ public class DefaultServerOAuth2AuthorizationRequestResolver
 		return UriComponentsBuilder.fromUriString(clientRegistration.getRedirectUriTemplate())
 				.buildAndExpand(uriVariables)
 				.toUriString();
+	}
+
+	/**
+	 * Creates and adds additional PKCE parameters for use in the OAuth 2.0 Authorization and Access Token Requests
+	 *
+	 * @param attributes where {@link PkceParameterNames#CODE_VERIFIER} is stored for the token request
+	 * @param additionalParameters where {@link PkceParameterNames#CODE_CHALLENGE} and, usually,
+	 * {@link PkceParameterNames#CODE_CHALLENGE_METHOD} are added to be used in the authorization request.
+	 *
+	 * @since 5.2
+	 * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7636#section-1.1">1.1.  Protocol Flow</a>
+	 * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7636#section-4.1">4.1.  Client Creates a Code Verifier</a>
+	 * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7636#section-4.2">4.2.  Client Creates the Code Challenge</a>
+	 */
+	private void addPkceParameters(Map<String, Object> attributes, Map<String, Object> additionalParameters) {
+		String codeVerifier = codeVerifierGenerator.generateKey();
+		attributes.put(PkceParameterNames.CODE_VERIFIER, codeVerifier);
+		try {
+			String codeChallenge = createCodeChallenge(codeVerifier);
+			additionalParameters.put(PkceParameterNames.CODE_CHALLENGE, codeChallenge);
+			additionalParameters.put(PkceParameterNames.CODE_CHALLENGE_METHOD, "S256");
+		} catch (NoSuchAlgorithmException e) {
+			additionalParameters.put(PkceParameterNames.CODE_CHALLENGE, codeVerifier);
+		}
+	}
+
+	private String createCodeChallenge(String codeVerifier) throws NoSuchAlgorithmException {
+		MessageDigest md = MessageDigest.getInstance("SHA-256");
+		byte[] digest = md.digest(codeVerifier.getBytes(StandardCharsets.US_ASCII));
+		return Base64.getUrlEncoder().withoutPadding().encodeToString(digest);
 	}
 }
--- a/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/endpoint/WebClientReactiveAuthorizationCodeTokenResponseClientTests.java
+++ b/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/endpoint/WebClientReactiveAuthorizationCodeTokenResponseClientTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -32,9 +32,12 @@ import org.springframework.security.oauth2.core.endpoint.OAuth2AccessTokenRespon
 import org.springframework.security.oauth2.core.endpoint.OAuth2AuthorizationExchange;
 import org.springframework.security.oauth2.core.endpoint.OAuth2AuthorizationRequest;
 import org.springframework.security.oauth2.core.endpoint.OAuth2AuthorizationResponse;
+import org.springframework.security.oauth2.core.endpoint.PkceParameterNames;
 import org.springframework.web.reactive.function.client.WebClient;
 
 import java.time.Instant;
+import java.util.HashMap;
+import java.util.Map;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
@@ -84,6 +87,9 @@ public class WebClientReactiveAuthorizationCodeTokenResponseClientTests {
 		Instant expiresAtBefore = Instant.now().plusSeconds(3600);
 
 		OAuth2AccessTokenResponse accessTokenResponse = this.tokenResponseClient.getTokenResponse(authorizationCodeGrantRequest()).block();
+		String body = this.server.takeRequest().getBody().readUtf8();
+
+		assertThat(body).isEqualTo("grant_type=authorization_code&code=code&redirect_uri=%7BbaseUrl%7D%2F%7Baction%7D%2Foauth2%2Fcode%2F%7BregistrationId%7D");
 
 		Instant expiresAtAfter = Instant.now().plusSeconds(3600);
 
@@ -287,5 +293,52 @@ public class WebClientReactiveAuthorizationCodeTokenResponseClientTests {
 		OAuth2AccessTokenResponse response = this.tokenResponseClient.getTokenResponse(authorizationCodeGrantRequest()).block();
 
 		verify(customClient, atLeastOnce()).post();
+	}
+
+	@Test
+	public void getTokenResponseWhenOAuth2AuthorizationRequestContainsPkceParametersThenTokenRequestBodyShouldContainCodeVerifier() throws Exception {
+		String accessTokenSuccessResponse = "{\n" +
+				"	\"access_token\": \"access-token-1234\",\n" +
+				"   \"token_type\": \"bearer\",\n" +
+				"   \"expires_in\": \"3600\"\n" +
+				"}\n";
+		this.server.enqueue(jsonResponse(accessTokenSuccessResponse));
+
+		this.tokenResponseClient.getTokenResponse(pkceAuthorizationCodeGrantRequest()).block();
+		String body = this.server.takeRequest().getBody().readUtf8();
+
+		assertThat(body).isEqualTo("grant_type=authorization_code&code=code&redirect_uri=%7BbaseUrl%7D%2F%7Baction%7D%2Foauth2%2Fcode%2F%7BregistrationId%7D&client_id=client-id&code_verifier=code-verifier-1234");
+	}
+
+	private OAuth2AuthorizationCodeGrantRequest pkceAuthorizationCodeGrantRequest() {
+		ClientRegistration registration = this.clientRegistration
+				.clientSecret(null)
+				.build();
+
+		Map<String, Object> attributes = new HashMap<>();
+		attributes.put(PkceParameterNames.CODE_VERIFIER, "code-verifier-1234");
+
+		Map<String, Object> additionalParameters = new HashMap<>();
+		additionalParameters.put(PkceParameterNames.CODE_CHALLENGE, "code-challenge-1234");
+		additionalParameters.put(PkceParameterNames.CODE_CHALLENGE_METHOD, "S256");
+
+		OAuth2AuthorizationRequest authorizationRequest = OAuth2AuthorizationRequest
+				.authorizationCode()
+				.clientId(registration.getClientId())
+				.state("state")
+				.authorizationUri(registration.getProviderDetails().getAuthorizationUri())
+				.redirectUri(registration.getRedirectUriTemplate())
+				.scopes(registration.getScopes())
+				.attributes(attributes)
+				.additionalParameters(additionalParameters)
+				.build();
+		OAuth2AuthorizationResponse authorizationResponse = OAuth2AuthorizationResponse
+				.success("code")
+				.state("state")
+				.redirectUri(registration.getRedirectUriTemplate())
+				.build();
+		OAuth2AuthorizationExchange authorizationExchange = new OAuth2AuthorizationExchange(authorizationRequest,
+				authorizationResponse);
+		return new OAuth2AuthorizationCodeGrantRequest(registration, authorizationExchange);
 	}
 }
--- a/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/registration/ClientRegistrationTests.java
+++ b/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/registration/ClientRegistrationTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -172,6 +172,41 @@ public class ClientRegistrationTests {
 				.clientName(CLIENT_NAME)
 				.build();
 		assertThat(clientRegistration.getClientAuthenticationMethod()).isEqualTo(ClientAuthenticationMethod.BASIC);
+	}
+
+	@Test
+	public void buildWhenAuthorizationCodeGrantClientAuthenticationMethodNotProvidedAndClientSecretNullThenDefaultToNone() {
+		ClientRegistration clientRegistration = ClientRegistration.withRegistrationId(REGISTRATION_ID)
+				.clientId(CLIENT_ID)
+				.clientSecret(null)
+				.authorizationGrantType(AuthorizationGrantType.AUTHORIZATION_CODE)
+				.redirectUriTemplate(REDIRECT_URI)
+				.scope(SCOPES.toArray(new String[0]))
+				.authorizationUri(AUTHORIZATION_URI)
+				.tokenUri(TOKEN_URI)
+				.userInfoAuthenticationMethod(AuthenticationMethod.FORM)
+				.jwkSetUri(JWK_SET_URI)
+				.clientName(CLIENT_NAME)
+				.build();
+		assertThat(clientRegistration.getClientAuthenticationMethod()).isEqualTo(ClientAuthenticationMethod.NONE);
+	}
+
+	@Test
+	public void buildWhenAuthorizationCodeGrantClientAuthenticationMethodNotProvidedAndClientSecretBlankThenDefaultToNone() {
+		ClientRegistration clientRegistration = ClientRegistration.withRegistrationId(REGISTRATION_ID)
+				.clientId(CLIENT_ID)
+				.clientSecret(" ")
+				.authorizationGrantType(AuthorizationGrantType.AUTHORIZATION_CODE)
+				.redirectUriTemplate(REDIRECT_URI)
+				.scope(SCOPES.toArray(new String[0]))
+				.authorizationUri(AUTHORIZATION_URI)
+				.tokenUri(TOKEN_URI)
+				.userInfoAuthenticationMethod(AuthenticationMethod.FORM)
+				.jwkSetUri(JWK_SET_URI)
+				.clientName(CLIENT_NAME)
+				.build();
+		assertThat(clientRegistration.getClientAuthenticationMethod()).isEqualTo(ClientAuthenticationMethod.NONE);
+		assertThat(clientRegistration.getClientSecret()).isEqualTo("");
 	}
 
 	@Test(expected = IllegalArgumentException.class)
--- a/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/registration/ClientRegistrationsTest.java
+++ b/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/registration/ClientRegistrationsTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -92,7 +92,8 @@ public class ClientRegistrationsTest {
 			+ "    \"token_endpoint\": \"https://example.com/oauth2/v4/token\", \n"
 			+ "    \"token_endpoint_auth_methods_supported\": [\n"
 			+ "        \"client_secret_post\", \n"
-			+ "        \"client_secret_basic\"\n"
+			+ "        \"client_secret_basic\", \n"
+			+ "        \"none\"\n"
 			+ "    ], \n"
 			+ "    \"userinfo_endpoint\": \"https://example.com/oauth2/v3/userinfo\"\n"
 			+ "}";
@@ -119,7 +120,7 @@ public class ClientRegistrationsTest {
 
 	@Test
 	public void issuerWhenAllInformationThenSuccess() throws Exception {
-		ClientRegistration registration = registration("");
+		ClientRegistration registration = registration("").build();
 		ClientRegistration.ProviderDetails provider = registration.getProviderDetails();
 
 		assertThat(registration.getClientAuthenticationMethod()).isEqualTo(ClientAuthenticationMethod.BASIC);
@@ -154,7 +155,7 @@ public class ClientRegistrationsTest {
 	public void issuerWhenScopesNullThenScopesDefaulted() throws Exception {
 		this.response.remove("scopes_supported");
 
-		ClientRegistration registration = registration("");
+		ClientRegistration registration = registration("").build();
 
 		assertThat(registration.getScopes()).containsOnly("openid");
 	}
@@ -163,7 +164,7 @@ public class ClientRegistrationsTest {
 	public void issuerWhenGrantTypesSupportedNullThenDefaulted() throws Exception {
 		this.response.remove("grant_types_supported");
 
-		ClientRegistration registration = registration("");
+		ClientRegistration registration = registration("").build();
 
 		assertThat(registration.getAuthorizationGrantType()).isEqualTo(AuthorizationGrantType.AUTHORIZATION_CODE);
 	}
@@ -184,7 +185,7 @@ public class ClientRegistrationsTest {
 	public void issuerWhenTokenEndpointAuthMethodsNullThenDefaulted() throws Exception {
 		this.response.remove("token_endpoint_auth_methods_supported");
 
-		ClientRegistration registration = registration("");
+		ClientRegistration registration = registration("").build();
 
 		assertThat(registration.getClientAuthenticationMethod()).isEqualTo(ClientAuthenticationMethod.BASIC);
 	}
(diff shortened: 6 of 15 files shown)


oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java
```diff
@@ -362,5 +382,17 @@ private String buildAuthorizationRequestUri() {
 					.build()
 					.toUriString();
 		}
+
+		private String getCodeChallenge(String codeVerifier) {
```
rwinch: Just as the OAuth2AuthorizationRequest has no logic in generating the state or code verifier, I'd prefer to avoid logic on generating a code challenge. More concretely, I think we should externalize how the code challenge is created rather than embedding the logic in here. This gives us more flexibility for supporting other types of challenges.

crypto/src/main/java/org/springframework/security/crypto/keygen/PkceCodeVerifierStringKeyGenerator.java
```diff
@@ -0,0 +1,18 @@
+package org.springframework.security.crypto.keygen;
+
+import java.security.SecureRandom;
+import java.util.Base64;
+
+/**
```
rwinch: Please add a little more Javadoc.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolver.java
```diff
@@ -105,6 +109,9 @@ private OAuth2AuthorizationRequest resolve(HttpServletRequest request, String re
 		OAuth2AuthorizationRequest.Builder builder;
 		if (AuthorizationGrantType.AUTHORIZATION_CODE.equals(clientRegistration.getAuthorizationGrantType())) {
 			builder = OAuth2AuthorizationRequest.authorizationCode();
+			if (isEmpty(clientRegistration.getClientSecret())) {
```
rwinch: Is this the best way to determine if PKCE should be used? This seems implicit rather than explicit which can lead to problems later on (if there are other reasons the client secret can be empty). Additionally implicit options in security tends to lead to security vulnerabilities.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/DefaultServerOAuth2AuthorizationRequestResolver.java
```diff
@@ -125,6 +130,9 @@ private OAuth2AuthorizationRequest authorizationRequest(ServerWebExchange exchan
 		OAuth2AuthorizationRequest.Builder builder;
 		if (AuthorizationGrantType.AUTHORIZATION_CODE.equals(clientRegistration.getAuthorizationGrantType())) {
 			builder = OAuth2AuthorizationRequest.authorizationCode();
+			if (isEmpty(clientRegistration.getClientSecret())) {
```
rwinch: Again is the best way to determine if PKCE should be used?

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java
```diff
@@ -44,6 +43,7 @@
  * @see OAuth2AuthorizationResponseType
  * @see <a target="_blank" href="https://tools.ietf.org/html/rfc6749#section-4.1.1">Section 4.1.1 Authorization Code Grant Request</a>
  * @see <a target="_blank" href="https://tools.ietf.org/html/rfc6749#section-4.2.1">Section 4.2.1 Implicit Grant Request</a>
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7636#section-4.3">Section 4.3 Client Sends the Code Challenge with the Authorization Request</a>
  */
 public final class OAuth2AuthorizationRequest implements Serializable {
```
rwinch: I think it makes sense to allow this to be extended so that there is a specific implementation that is used for supporting pkce. This makes sense because we don't want to have a request that supports all extensions making it confusing for users who only need a simple request.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2ParameterNames.java
```diff
@@ -60,6 +60,12 @@
 	 */
 	String STATE = "state";
 
+	String CODE_CHALLENGE = "code_challenge";
```
rwinch: I think it might make sense to extend the class and put the extension parameters in a new interface.

sdoxsee:
I've force-pushed a new commit that tries to take @rwinch's suggestions into consideration. 

As for explicit detection of PKCE rather that detecting the absence of a `client_secret`, would ClientAuthenticationMethod.NONE be appropriate?

I've extended the OAuth2AuthorizationRequest class to handle PKCE specifics. The OAuth2AuthorizationRequest class needed tweaks to allow it to be subclassed as well as its Builder.

I didn't subclass OAuth2AuthorizationCodeGrantRequestEntityConverter so I used an "instanceof" to be sure I could access the codeVerifier on the OAuth2AuthorizationRequest :/

I plan on writing tests once the design is a little more settled. Thanks for the feedback!

crypto/src/main/java/org/springframework/security/crypto/keygen/PkceCodeVerifierStringKeyGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+ * Creates a high-entropy cryptographic random string to use as the PKCE Code Verifier. By default,
+ * this uses 32 octets to produce a 43 character string--the minimum for a PKCE Code Verifier.
+ *
+ * <a target="_blank" href="https://tools.ietf.org/html/rfc7636#section-4.1">Section 4.1 Client Creates a Code Verifier</a>
+ * @since 5.2
+ * @author Stephen Doxsee
+ * @author Kevin Bolduc
+ */
+public class PkceCodeVerifierStringKeyGenerator implements StringKeyGenerator {
```
jgrandja: Please remove this and use `Base64StringKeyGenerator` by supplying it a `Base64.Encoder` and `keyLength`. Also note that it uses `SecureRandom` internally.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClientAuthenticationMethod.java
```diff
@@ -31,6 +31,12 @@
 	private static final long serialVersionUID = SpringSecurityCoreVersion.SERIAL_VERSION_UID;
 	public static final ClientAuthenticationMethod BASIC = new ClientAuthenticationMethod("basic");
 	public static final ClientAuthenticationMethod POST = new ClientAuthenticationMethod("post");
+
+	/**
```
jgrandja: Add javadoc

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java
```diff
@@ -45,7 +45,7 @@
  * @see <a target="_blank" href="https://tools.ietf.org/html/rfc6749#section-4.1.1">Section 4.1.1 Authorization Code Grant Request</a>
  * @see <a target="_blank" href="https://tools.ietf.org/html/rfc6749#section-4.2.1">Section 4.2.1 Implicit Grant Request</a>
  */
-public final class OAuth2AuthorizationRequest implements Serializable {
+public class OAuth2AuthorizationRequest implements Serializable {
```
jgrandja: `OAuth2AuthorizationRequest` does not require any changes. One of the primary use cases for the `OAuth2AuthorizationRequestResolver` is to customize the **Authorization Request** by adding additional parameters. This will be needed for OpenID Connect flows and also can be leveraged for PKCE. Please see the [reference doc](https://docs.spring.io/spring-security/site/docs/5.1.3.RELEASE/reference/htmlsingle/#oauth2Client-authorization-request-resolver) on how to do this.

For the changes required in this PR, I would put the logic in `DefaultOAuth2AuthorizationRequestResolver` and place the PKCE parameters in `OAuth2AuthorizationRequest.additionalParameters`

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/pkce/endpoint/CodeChallengeTransformation.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Interface for the transformation of Code Verifier to a Code Challenge.
+ *
+ * @author Stephen Doxsee
+ * @author Kevin Bolduc
+ * @since 5.2
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7636#section-4.2">4.2 Client Creates the Code Challenge</a>
+ */
+public interface CodeChallengeTransformation {
```
jgrandja: Please remove this and encapsulate the `code_challenge` computation in `DefaultOAuth2AuthorizationRequestResolver`. We can expose later if need be.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/pkce/endpoint/PkceOAuth2AuthorizationRequest.java
```diff
@@ -0,0 +1,93 @@
...
+ * A representation of an OAuth 2.0 Authorization Request
+ * for the authorization code grant type with Proof Key
+ * for Code Exchange by OAuth Public Clients.
+ *
+ * @author Stephen Doxsee
+ * @author Kevin Bolduc
+ * @since 5.2
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7636#section-4.3">4.3 Client Sends the Code Challenge with the Authorization Request</a>
+ */
+public final class PkceOAuth2AuthorizationRequest extends OAuth2AuthorizationRequest {
```
jgrandja: Please remove as per previous comment - use `OAuth2AuthorizationRequest.additionalParameters`

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/pkce/endpoint/PkceParameterNames.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Standard parameter names defined in the OAuth Parameters Registry
+ * and used by the authorization endpoint and token endpoint.
+ *
+ * @author Stephen Doxsee
+ * @author Kevin Bolduc
+ * @since 5.2
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7636#section-6.1">6.1 OAuth Parameters Registry</a>
+ */
+public interface PkceParameterNames {
```
jgrandja: Please remove this and put the PKCE parameters in `OAuth2ParameterNames`

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/pkce/endpoint/Sha256CodeChallengeTransformation.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * A @{@link CodeChallengeTransformation} for the transformation of Code Verifier to
+ * a Code Challenge using SHA-256.
+ *
+ * @author Stephen Doxsee
+ * @author Kevin Bolduc
+ * @since 5.2
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7636#section-4.2">4.2 Client Creates the Code Challenge</a>
+ */
+public class Sha256CodeChallengeTransformation implements CodeChallengeTransformation {
```
jgrandja: Please remove this and encapsulate the `code_challenge` computation in `DefaultOAuth2AuthorizationRequestResolver`. We can expose later if need be.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/pkce/endpoint/PkceParameterNames.java
```diff
@@ -0,0 +1,43 @@
...
+
+	/**
+	 * {@code code_challenge} - used in Authorization Request.
```
jgrandja: Please javadoc all parameters - copy as-is from spec

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java
jgrandja: Agreed. See my comment on putting this logic in `DefaultOAuth2AuthorizationRequestResolver` instead

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolver.java
jgrandja: I think this makes sense. If `AuthorizationGrantType.AUTHORIZATION_CODE` and `ClientAuthenticationMethod.NONE` (public client) than PKCE parameters can be applied.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java
jgrandja: I agree that it makes sense to allow `OAuth2AuthorizationRequest` to be extended but don't feel that it's required at this moment. PKCE supplies 2 parameters and is quite simple. I think this logic can be default logic in `DefaultOAuth2AuthorizationRequestResolver`. See [comment](https://github.com/spring-projects/spring-security/pull/6485#discussion_r253268516)

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java
sdoxsee: I'm open to change on this but I feel strongly that the Code Challenge shouldn't be calculated outside of the OAuth2AuthorizationRequest. The Code Challenge is a function of the Code Verifier such that if you know the Code Verifier and the Code Challenge Method (i.e. "f(x)"--where "f()" is always SHA-256 unless the client can't calculate it and "x" is the Code Verifier) then you can determine the Code Challenge. It feels wrong (to me) to store both the Code Verifier AND the Code Challenge on the OAuth2AuthorizationRequest. By setting the Code Challenge externally (i.e. in the DefaultOAuth2AuthorizationRequestResolver), then the users of the OAuth2AuthorizationRequest.Builder could set incompatible Code Verifier and Code Challenge values. That would seems like a bad builder api to me. I'm ok breaking out the actual transformation/function to another class with the ability to set a future new transformation but even that seems unnecessary at this point with the RFC being what it is (i.e. S256 mandatory unless client can't->PLAIN). Again, I'm open to change my mind but can you see where I'm coming from? I understand wanting to keep logic out of what's essentially an stateful entity but I don't see it so much as "logic"--rather a straightforward mathematical function of the Code Verifier. Please convince me otherwise. Thanks!

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolver.java
sdoxsee: @jgrandja I like the `AuthorizationGrantType.AUTHORIZATION_CODE` and `ClientAuthenticationMethod.NONE` combination to determine PKCE usage. @rwinch I agree the absence of a client secret is not great. Are you ok with `AuthorizationGrantType.AUTHORIZATION_CODE` + `ClientAuthenticationMethod.NONE`?

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/pkce/endpoint/PkceParameterNames.java
sdoxsee: I'll re-copy as-is from spec. Thanks!

crypto/src/main/java/org/springframework/security/crypto/keygen/PkceCodeVerifierStringKeyGenerator.java
sdoxsee: I'll try that out. Thanks

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2ParameterNames.java
sdoxsee: I suppose that if we subclassed OAuth2AuthorizationRequest with PkceOAuthAuthorizationRequest then it would make sense to separate the PKCE-specific params into PkceParameterNames. However, if we want to back PKCE support directly into OAuth2AuthorizationRequest then OAuth2ParameterNames works? The thing I like about subclassing OAuth2AuthorizationRequest and keeping PkceParameterNames is that it might make the testing of the classes a little cleaner IMO. @rwinch I'll wait for you to chime back in on this. Add PKCE support directly to OAuth2AuthorizationRequest or subclass?

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java
jgrandja: The `OAuth2AuthorizationRequest` is a domain class, pretty much a POJO but OAuth2 specific.  In my view, it stores attributes of an Authorization Request, e.g. `client-id`, `redirect-uri`, `response_type`, etc. As far as PKCE support goes, it also stores attributes such as `code_challenge` and `code_challenge_method`. These attributes serve as standard and additional (`OAuth2AuthorizationRequest.additionalParameters`) request parameters to the Authorization Endpoint. 

The `DefaultOAuth2AuthorizationRequestResolver` is responsible for creating an `OAuth2AuthorizationRequest` and setting it's attributes/parameters. It makes sense to me that the `code_challenge` and `code_challenge_method` would also be set in `OAuth2AuthorizationRequest.additionalParameters` by `DefaultOAuth2AuthorizationRequestResolver`. However, before it can compute the `code_challenge`, it needs to create the `code_verifier` and associate it to `OAuth2AuthorizationRequest` so it can be sent later to the Token Endpoint. It doesn't make sense to store the `code_verifier` in `additionalParameters()` because it's not a parameter to the Authorization Endpoint. So I'm going to propose that the `code_verifier` is stored in `OAuth2AuthorizationRequest.attributes` (see related #5940). Th `attributes` member serves the purpose of storing contextual information of an Authorization Request, which the `code_verifier` is exactly that as it lives during the duration of the Authorization Request and subsequent callback of the Authorization Response.

What are your thoughts @sdoxsee @rwinch ?

sdoxsee: Hey @jgrandja. I agree with your thinking on this. The named parameters on the OAuth2AuthorizationRequest should really be about the authorization endpoint call and not the token endpoint which would make something like `attributes` good for the `code_verifier`--to be used for the token endpoint call. Once the PKCE parameters are moved into additionalParameters and attributes to keep the PKCE specifics out entirely, I can buy the proposition of doing the `code_challenge` logic in the DefaultOAuth2AuthorizationRequestResolver. I like it because it's extensible but keeps things simple. I don't like subclassing if I can avoid it.

Several more questions:

1. Can I use PkceParameterNames as your solution means that OAuth2AuthorizationRequest is agnostic of PKCE and I don't like dropping them into OAuth2ParameterNames?
2. Shall I mix issues and do #5940 as well?
3. @rwinch are you ok with this before I code it up?

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClientAuthenticationMethod.java
sdoxsee: BASIC and POST don't have anything. Should I just differentiate this by saying something like "@Since 5.2, For public clients that don't require client authentication."?

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java
sdoxsee: Thanks @jgrandja. If you're able to wrap #5940 up quickly then that might be best for efficiency--I'll rebase when you let me know. I'm ok with OAuth2ParameterNames or PkceParameterNames. I suppose the same could be argued for the existing OidcParameterNames though as well? I'll wait for @rwinch before coding anyway.

jgrandja: Ok I'll take care of #5940 and let you know when it's in. Good point re: `OidcParameterNames` - let's see what @rwinch has to say.

jgrandja: @sdoxsee #5940 has been resolved via 594a169

sdoxsee:
@jgrandja I rebased on #5940 so I had to force push, sorry! Working on tests now but if you or @rwinch can provide feedback then that would be great. Thanks!

sdoxsee:
I've added some tests and fixed the bad checkstyle from the previous commit. When you get a chance, please let me know how to proceed. Thanks!

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClientAuthenticationMethod.java
```diff
@@ -31,6 +31,12 @@
...
+	 * @since 5.2
+	 */
+	public static final ClientAuthenticationMethod NONE = new ClientAuthenticationMethod("none");
```
jgrandja: Please add test

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/pkce/endpoint/PkceParameterNames.java
jgrandja: On second thought I'm ok with leaving this.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolver.java
```diff
@@ -52,6 +57,7 @@
 	private final ClientRegistrationRepository clientRegistrationRepository;
 	private final AntPathRequestMatcher authorizationRequestMatcher;
 	private final StringKeyGenerator stateGenerator = new Base64StringKeyGenerator(Base64.getUrlEncoder());
+	private final StringKeyGenerator codeVerifierGenerator = new Base64StringKeyGenerator(Base64.getUrlEncoder().withoutPadding());
```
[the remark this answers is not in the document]
sdoxsee: @jgrandja we should use `withoutPadding()` to ensure that we generate a code_verifier that doesn't contain any `=` padding characters at the end of the String since the spec only allows `[A-Z] / [a-z] / [0-9] / "-" / "." / "_" / "~"`.

As for the default length, I'll set the key length to 96 (bytes) rather than the default 32 (bytes) because it translates to the maximum base64 url-encoded string length of 128.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolver.java
```diff
@@ -156,4 +167,29 @@ private String expandRedirectUri(HttpServletRequest request, ClientRegistration
...
+			additionalParameters.put(PkceParameterNames.CODE_CHALLENGE, codeVerifier);
+		}
+	}
+
+	private String transformationS256(String codeVerifier) throws UnsupportedEncodingException, NoSuchAlgorithmException {
+		byte[] bytes = codeVerifier.getBytes("US-ASCII");
+		MessageDigest md = MessageDigest.getInstance("SHA-256");
+		md.update(bytes, 0, bytes.length);
+		byte[] digest = md.digest();
+		return Base64.getUrlEncoder().withoutPadding().encodeToString(digest);
```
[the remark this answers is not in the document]
sdoxsee: I don't think that's correct. That interprets the bytes as ASCII but we want to interpret them as Base64 URL-encoded characters. 
Examples: 
- https://github.com/networknt/light-4j/blob/1.5.28/utility/src/main/java/com/networknt/utility/CodeVerifierUtil.java#L119
- https://auth0.com/docs/api-auth/tutorials/authorization-code-grant-pkce (see Java implementations)

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolver.java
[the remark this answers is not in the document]
jgrandja: > ensure that we generate a code_verifier that doesn't contain any `=` padding characters

Gotchya, makes sense.

> As for the default length, I'll set the key length to 96 (bytes)

Great

sdoxsee:
@jgrandja Just pushed another commit to address your latest suggestions. Tests pass locally--hopefully Travis will agree. Thanks!

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolver.java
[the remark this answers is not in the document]
jgrandja: Yes, sorry I missed that the `codeVerifier` was already Base64 encoded.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolver.java
sdoxsee: Hi @jgrandja and @rwinch, I received an email from someone watching this PR that raises a good point. Attaching below:

> Hi Stephen,
> 
> Recently I bumped into your pullrequest for PKCE.
> 
> For our project we even have the requirement of using PKCE for normal web applications as an extra hardening.
> 
> Okta has put a demo of the PKCE in their playground:
> 
> https://oauth.com/playground/authorization-code-with-pkce.html
> 
> You will see that in the final step, to get the token, they pass both the client_secret (client authentication client_secret_basic in this case) and code_verifier.
> 
> 
> In the pull request I saw some discussion on how to determine PKCE:
> 
> https://github.com/spring-projects/spring-security/pull/6485#discussion_r253276486
> 
> In our case or the above mentioned example on Okta this would not be fine. I believe we need a special parameter where we can force PKCE.
> 
> What do you think?

I think he makes a good point but it would mean some kind of further flag(?) on ClientRegistration like "withPkce" (true/false) that would only apply for authorization_code grant type. This way PKCE could be used with or without a client secret.

jgrandja: @sdoxsee I'm not sure I'm following the thread in the email you received.

PKCE is meant to be used for Public Clients leveraging the Authorization Code grant. Public Clients are not assigned full credentials (no client secret). So I guess I'm confused with what the user is saying.

Looking at the Okta developer guide for [Authorization Code Flow with PKCE](https://developer.okta.com/authentication-guide/implementing-authentication/auth-code-pkce), you will see that the client authentication is not required (so no `client_secret`).

> Important: Unlike the regular Authorization Code Flow, this call does not require the Authorization header with the client ID and secret. This is why this version of the Authorization Code flow is appropriate for native apps.

And for the standard [Authorization Code Flow](https://developer.okta.com/authentication-guide/implementing-authentication/auth-code), you will see that the client authentication is required as this flow is targeted for confidential clients.

> Important: The call to the /token endpoint requires authentication. In this case, it is a Basic Auth digest of the client ID and secret. You can find the client ID and secret in your application’s General tab. This requirement is why this call is only appropriate for applications that can guarantee the secrecy of the client secret. For more on Basic Auth, please see Token Authentication Methods.

sdoxsee: @jgrandja. I agree that it is not typical to compose authorization_code grant (with `client_secret`) AND PKCE but, despite the PKCE spec being "Proof Key for Code Exchange by OAuth **Public Clients**", the spec doesn't seem to prohibit the use of `client_secret`. For instance, the [token endpoint request for PKCE](https://tools.ietf.org/html/rfc7636#section-4.5) just builds on the generic [OAuth2 authorization_code token request](https://tools.ietf.org/html/rfc6749#section-4.1.3). 

You're right, for public clients, client authentication is not _required_ but it's not prohibited. In any case, this is not an important customization for my use case. I've contacted the person who emailed to chime in if it is important for them.

jgrandja: @sdoxsee In reference to...

> For instance, the [token endpoint request for PKCE](https://tools.ietf.org/html/rfc7636#section-4.5) just builds on the generic [OAuth2 authorization_code token request](https://tools.ietf.org/html/rfc6749#section-4.1.3).

If you dig deeper into section, 4.1.3.  Access Token Request

> **If the client type is confidential** or the client was issued client
   credentials (or assigned other authentication requirements), the
   client MUST authenticate with the authorization server as described
   in Section 3.2.1.

So for a public client (PKCE), which is never issued a `client_secret`, since it cannot securely store it (e.g. a native app) it could not authenticate anyway since it does not have a `client_secret` assigned during registration. See [Client Types](https://tools.ietf.org/html/rfc6749#section-2.1)

sdoxsee: Thanks @jgrandja. Yes, I drilled down to that definition as well. What this other person is saying is that they have a **confidential** client (likely a traditional oauth2 authorization code flow web app?) but want to add PKCE also--as extra security. In such a case, you WOULD have a PKCE client that HAS a `client_secret`

Why? I'm not sure but I can only imagine that it's to protect against the scenario where the "code" comes back via the front channel and they want to prevent some malicious agent, that intercepts the "code" and somehow has not only the client_id but also the client_secret, from getting tokens. If their `client_secret` is compromised, I believe they probably have bigger problems. However, the person who brought this up is saying that they have this requirement to add PKCE to a confidential client and I can't see why they should not be able to do it according to the spec.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/registration/ClientRegistrationsTest.java
```diff
@@ -92,7 +92,8 @@
 			+ "    \"token_endpoint\": \"https://example.com/oauth2/v4/token\", \n"
 			+ "    \"token_endpoint_auth_methods_supported\": [\n"
 			+ "        \"client_secret_post\", \n"
-			+ "        \"client_secret_basic\"\n"
```
jgrandja: Please update copyright header to 2019

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/registration/ClientRegistrationsTest.java
```diff
@@ -193,11 +194,22 @@ public void issuerWhenTokenEndpointAuthMethodsNullThenDefaulted() throws Excepti
...
 
 		assertThat(registration.getClientAuthenticationMethod()).isEqualTo(ClientAuthenticationMethod.POST);
 	}
 
+	@Test
+	public void issuerWhenTokenEndpointAuthMethodsNoneThenMethodIsNone() throws Exception {
+		this.response.put("token_endpoint_auth_methods_supported", Arrays.asList("none"));
+
+		ClientRegistration registration = registration("")
+				.clientAuthenticationMethod(ClientAuthenticationMethod.NONE)
```
jgrandja: This line should be removed to ensure `clientAuthenticationMethod` is `ClientAuthenticationMethod.NONE`

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClientAuthenticationMethod.java
jgrandja: Please update copyright header to 2019

oauth2/oauth2-core/src/test/java/org/springframework/security/oauth2/core/ClientAuthenticationMethodTests.java
```diff
@@ -40,4 +40,9 @@ public void getValueWhenAuthenticationMethodBasicThenReturnBasic() {
 	public void getValueWhenAuthenticationMethodPostThenReturnPost() {
 		assertThat(ClientAuthenticationMethod.POST.getValue()).isEqualTo("post");
 	}
+
```
jgrandja: Please update copyright header to 2019

oauth2/oauth2-core/src/test/java/org/springframework/security/oauth2/core/ClientAuthenticationMethodTests.java
```diff
@@ -40,4 +40,9 @@ public void getValueWhenAuthenticationMethodBasicThenReturnBasic() {
...
+	@Test
+	public void getValueWhenAuthenticationMethodNoneThenReturnPost() {
```
jgrandja: Rename `getValueWhenAuthenticationMethodNoneThenReturnPost` -> `getValueWhenAuthenticationMethodNoneThenReturnNone`

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolver.java
jgrandja: Thanks for the explanation @sdoxsee. Now I understand the use case. However, this seems more of an edge case to me. I'd prefer that we leave this out of this PR so we can get this merged. Can you please add a new ticket so we can gather more information.

> the person who brought this up is saying that they have this requirement to add PKCE to a confidential client

Can you also please have the user respond to the new ticket with more detail and specifically which Authorization Server supports PKCE for a confidential client (Okta?). A link to server docs would be even better.

vanwobe: Hi,
The current best practices for OAuth security specifies that PKCE should also be used for webapps:

> Note: although PKCE so far was recommended as a mechanism to protect
   native apps, this advice applies to all kinds of OAuth clients,
   **_including web applications_**.

see: https://tools.ietf.org/html/draft-ietf-oauth-security-topics-11#section-2.1.1

Now that you are implementing PKCE you better foresee that PKCE can also be used for web applications.

jgrandja: @vanwobe Thanks for the reference. Just below the reference it continues with...

> Authorization servers MUST bind authorization codes to a certain
   client and authenticate it using an appropriate mechanism (e.g.
   client credentials **OR** PKCE).

The statement below is a bit misleading from the previous as it states **client credentials OR PKCE** (not both).

Either way, I'm all for making things more secure. I'm very curious though which OAuth 2.0 Provider's have implemented this recommended best practice? Can you point me to a provider that supports PKCE for a confidential client (in a web app)? A link to the reference documentation would be helpful.

Our primary goal when building out new features is to be spec compliant first and foremost and to provide out-of-the-box support for extension features supported by **popular** OAuth 2.0 Providers. As well, to provide extension points for users to implement for edge-case scenarios.

vanwobe: We have it implemented in our custom provider but that aside, did you see the example from oauth.com: https://oauth.com/playground/authorization-code-with-pkce.html
This exactly demonstrates the use of PKCE in a web application.

jgrandja: @vanwobe @sdoxsee I have confirmed that Okta does indeed support PKCE for confidential clients. I have logged #6548 to track this feature enhancement. I'd rather get this PR merged based on the spec and than look at adding that support as a feature enhancement after.

sdoxsee:
> Thanks for the updates @sdoxsee. We're almost ready to get this merged! See my comments for minor updates.
> 
> Also, we need to update `WebClientReactiveAuthorizationCodeTokenResponseClient` to support a PKCE token request.
> 
> Thanks!

Hi @jgrandja. I've pushed the changes you recommended. I added support for PKCE token request to WebClientReactiveAuthorizationCodeTokenResponseClient. As I did that, I found some differences in logic compared to OAuth2AuthorizationCodeGrantRequestEntityConverter that was adding REDIRECT_URI regardless of whether it was null or not (and it is optional for authorization_code). I now only add it if it has a non-null value. I didn't add any tests for WebClientReactiveAuthorizationCodeTokenResponseClient since the test class was only looking at responses so far. I can certainly add tests to inspect the request before it is sent but that was starting to balloon a bit. Please let me know and I'll add some tests (and for the things I mention in the next paragraph).

Also in WebClientReactiveAuthorizationCodeTokenResponseClient and OAuth2AuthorizationCodeGrantRequestEntityConverter, I add the `client_id` for any ClientAuthenticationMethod other than BASIC because the [spec](https://tools.ietf.org/html/rfc6749#section-3.2.1) says:
>    In the "authorization_code" "grant_type" request to the token endpoint, an
>    **unauthenticated client** MUST send its "client_id" to prevent itself
>    from inadvertently accepting a code intended for a client with a
>    different "client_id".
where I assume "unauthenticated client** must mean a public client (ie. NONE)...but `client_id` must also appear for POST because it isn't coming in the `Authorization: Basic` header. You could argue that we should always send it since we always have it and it IS optional or that I should explicitly list off NONE and POST as the condition where `client_id` should be included. Anyway, sorry to open another can of worms but glad you pushed off #6548 so we can get this one out soon--hopefully :)

dfcoffin:
Hi @sdoxsee.  If you look at the first paragraph of the spec <https://tools.ietf.org/html/rfc6749#section-3.2.1>  you quoted, it says: 

 

Confidential clients or other clients issued client credentials (e.g., ClientID and secret) MUST

Authenticate with the authorization server as described in Section 2.3 <https://tools.ietf.org/html/rfc6749#section-2.3>  when making requests
to the token endpoint.

 

Section 2.3 Client Authentication states:

 

If the client type is confidential, the client and authorization

server establish a client authentication method suitable for the

security requirements of the authorization server.  The authorization

server MAY accept any form of client authentication meeting its

security requirements.

 

Confidential clients are typically issued (or establish) a set of

client credentials used for authenticating with the authorization

server (e.g., password, public/private key pair).

 

The authorization server MAY establish a client authentication method

with public clients.  However, the authorization server MUST NOT rely

on public client authentication for the purpose of identifying the

client.

 

The client MUST NOT use more than one authentication method in each

request.

 

I believe, based on the last paragraph of Section 2.3, inclusion of the ClientID in both the Authorization

Header and in the body is a violation of the standard.

 

Just my 2 cents.

 

 

Best regards,

Don

Donald F. Coffin

Founder/CTO

 

REMI Networks

2335 Dunwoody Xing #E

Dunwoody, GA 30338-8221

 

Phone:      (949) 636-8571

Email:       donald.coffin@reminetworks.com <mailto:donald.coffin@reminetworks.com>

jgrandja:
@dfcoffin Thanks for pointing out the following statement in [2.3.  Client Authentication](https://tools.ietf.org/html/rfc6749#section-3.2.1)

> The client MUST NOT use more than one authentication method in each
   request.

Okta does support client authentication via credentials as well as `code_verifier` to verify the `code_challenge`, so in essence, it is authenticating the client twice. Will keep this in mind while we consider #6548

sdoxsee:
Thanks @jgrandja and @dfcoffin. Specs are fun eh? ...sigh. Haha. Thanks for hashing out the details!

@jgrandja I see your point regarding `code_verifier` potentially being considered as another authentication mechanism but perhaps it's not so much _authenticating_ as it is making sure the token request is being made by the same party that made authorization request? The [spec](https://tools.ietf.org/html/rfc7636#section-1) calls it a "mitigation" or "proof of possession of the 'code verifier'". I think there's at least sufficient "gray" in the spec to allow for it.

@dfcoffin. Regarding where the spec says
> "The client MUST NOT use more than one authentication method in each request."

you say,
> I believe, based on the last paragraph of Section 2.3, inclusion of the ClientID in both the Authorization Header and in the body is a violation of the standard.

I believe that including the `client_id` in the body is fine all the time (even though the [oauth2 authorization_code example](https://tools.ietf.org/html/rfc6749#section-4.1.3) doesn't include it) because it not for authentication unless it also includes the `client_secret`. We only include the `client_secret` in the body for ClientAuthenticationMethod.POST--in which case no Authorization Header is sent...therefore, no violation? 

What do you think? Thanks!

dfcoffin:
@sdoxsee  The reason I pointed out that the last paragraph in Section 2.3 will be violated if the `client_id` is always provided, unless it is omitted for `confidential clients` or any client provided with a secret. The [Section 3.2.1 Client Authentication](https://tools.ietf.org/html/rfc6749#section-3.2.1) opening paragraph clearly states any client issued a secret MUST supply a `client_id` and `secret` in the HTTP Authorization header element using Base64 encoding.

Therefore, if `client_id` is always included in the `Authorization Code` request body, it will violate Section 2.3's final paragraph.

sdoxsee:
Thanks @dfcoffin 
> The [Section 3.2.1 Client Authentication](https://tools.ietf.org/html/rfc6749#section-3.2.1) opening paragraph clearly states any client issued a secret MUST supply a `client_id` and `secret` in the HTTP Authorization header element using Base64 encoding.

I'm not able to find anywhere in either 3.2.1 or 2.3 where it says that "any client issued a secret MUST supply a `client_id` and `secret` in the HTTP Authorization header". In fact it says in 2.3.1
> Clients in possession of a client password **MAY** use the HTTP Basic
   authentication scheme as defined in [RFC2617] to authenticate with
   the authorization server.

and then goes on to describe other methods of authenticating in 2.3.2. No?

Regards,
Stephen

dfcoffin:
@sdoxsee @jgrandja I agree [Section 2.3.1 Client Password's ](https://tools.ietf.org/html/rfc6749#section-2.3.1) title is a bit misleading, but it describes how clients assigned a `client password (a.k.a. secret)` MAY use the HTTP Basic authentication scheme as defined in [[RFC2617]](https://tools.ietf.org/html/rfc2617) to authenticate with the authorization server. It provides the following example:

`   For example (with extra line breaks for display purposes only):

     Authorization: Basic czZCaGRSa3F0Mzo3RmpmcDBaQnIxS3REUmJuZlZkbUl3

   Alternatively, the authorization server MAY support including the
   client credentials in the request-body using the following
   parameters:

   client_id
         REQUIRED.  The client identifier issued to the client during
         the registration process described by Section 2.2.

   client_secret
         REQUIRED.  The client secret.  The client MAY omit the
         parameter if the client secret is an empty string.
`
It then states:
`   Including the client credentials in the request-body using the two
   parameters is NOT RECOMMENDED and SHOULD be limited to clients unable
   to directly utilize the HTTP Basic authentication scheme (or other
   password-based HTTP authentication schemes).  The parameters can only
   be transmitted in the request-body and MUST NOT be included in the
   request URI.
`
I agree this section is rather vague, but given the example of using `client_id` and `secret` as `userID:password` as defined by [RFC2617], I believe was what the authors of the standard intended as the method an authorization server MUST use when authenticating `confidential_clients`, which the standard defines in [Section 2.1 Client Types](https://tools.ietf.org/html/rfc6749#section-2.1).

jgrandja:
@sdoxsee Can you please add a test for `WebClientReactiveAuthorizationCodeTokenResponseClientTests`. Alternatively, you can update  `getTokenResponseWhenSuccessResponseThenReturnAccessTokenResponse` to check the request similar to how `DefaultAuthorizationCodeTokenResponseClientTests.getTokenResponseWhenSuccessResponseThenReturnAccessTokenResponse()`, e.g.  `RecordedRequest recordedRequest = this.server.takeRequest();`

Also, one more change I think makes sense. In `ClientRegistration.create()`, if `authorization_code` and `client_secret` `empty` than default to `ClientAuthenticationMethod.NONE`.

sdoxsee:
Thanks @jgrandja. I'll probably add 1 test to `WebClientReactiveAuthorizationCodeTokenResponseClientTests` testing for PCKE parameters as well as enhance the existing getTokenResponseWhenSuccessResponseThenReturnAccessTokenResponse to assert on the request string (i.e. especially that it doesn't have PKCE params)

As for
> Also, one more change I think makes sense. In `ClientRegistration.create()`, if `authorization_code` and `client_secret` `empty` than default to `ClientAuthenticationMethod.NONE`.

Do you think that contradicts @rwinch's earlier [comment](https://github.com/spring-projects/spring-security/pull/6485#discussion_r252038492) by making this detection "implicit" rather than "explicit"? However, I agree with what you say since #6548 will mean that PKCE can be used *with* a secret as well--only that the `ClientAuthenticationMethod` would be `BASIC` rather than `NONE` in that case.

Also, what should be done with the defaulted `BASIC` for `clientAuthenticationMethod` on ClientRegistration? Should I add `BASIC` as the fallback in `create()` if nothing else is set on the builder? I think defaulting it at `create()` is better than setting it at initialization time because otherwise I can't easily detect if someone set it to something explicit on the builder before calling build (i.e. we'd never get a `null` without it being explicitly set to `null`)

Thanks,
Stephen

jgrandja:
@sdoxsee 

> what should be done with the defaulted `BASIC` for `clientAuthenticationMethod` on ClientRegistration?...


I think this is all the logic we need in `create()`

```java
// this.clientAuthenticationMethod is initialized to ClientAuthenticationMethod.BASIC
clientRegistration.clientAuthenticationMethod = this.clientAuthenticationMethod;

// Override for this case
if (AuthorizationGrantType.AUTHORIZATION_CODE.equals(this.authorizationGrantType) &&
		StringUtils.isEmpty(this.clientSecret)) {

	clientRegistration.clientAuthenticationMethod = ClientAuthenticationMethod.NONE;
}
```

Makes sense?

sdoxsee:
Yes @jgrandja. Thanks, almost wrote the exact same but with `!StringUtils.hasText(this.clientSecret)` to be consistent with the logic a couple lines up from there (e.g. `"<whitespace>"` should be considered as `""` for `clientSecret`). Going to go with `!StringUtils.hasText(this.clientSecret)` if that's ok and I'll add your comments in the code as well.

sdoxsee:
@jgrandja pushed but not sure what's wrong with the build :/

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/endpoint/WebClientReactiveAuthorizationCodeTokenResponseClientTests.java
```diff
@@ -288,4 +294,52 @@ public void setCustomWebClientThenCustomWebClientIsUsed() {
 
 		verify(customClient, atLeastOnce()).post();
 	}
+
+	private OAuth2AuthorizationCodeGrantRequest pkceAuthorizationCodeGrantRequest() {
```
jgrandja: Please move this method below the associated test.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
```diff
@@ -500,6 +500,11 @@ private ClientRegistration create() {
 			clientRegistration.clientId = this.clientId;
```
jgrandja: Update copyright header.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/registration/ClientRegistrationTests.java
```diff
@@ -174,6 +174,41 @@ public void buildWhenAuthorizationCodeGrantClientAuthenticationMethodNotProvided
 		assertThat(clientRegistration.getClientAuthenticationMethod()).isEqualTo(ClientAuthenticationMethod.BASIC);
```
jgrandja: Update copyright header.

sdoxsee:
Rebased, squashed and passing. Happy to contribute to #6548 when the time comes. 
Thanks to @kbolduc for pairing with me on this and to others who provided insights.
Thanks to both @jgrandja and @rwinch for all of _your_ help and patience :)

jgrandja:
Thank you @sdoxsee @kbolduc for your great work! This is now in master!

ExpDev07:
Any documentation on how to use this in a Spring app? How would I setup the configuration?

heyarny:
I'd like to see an example on how to use this as well, especially with password grant.
I don't want to see the refresh token within the SPA.
Any documentation out there about this new feature?

rwinch:
@ExpDev07 @heyarny There is no documentation for this feature yet. Would one of you create a ticket?

@sdoxsee Would you be interested in submitting a PR for adding documentation for PKCE?

PR #6495, opened 2019-01-31
Support symmetric key for JwtDecoder

### LINKED ISSUE #5465: Provide support for symmetric key verification via JwtDecoder
JWT signature verification for `id_token` fails in `OidcAuthorizationCodeAuthenticationProvider`  for OIDC providers like Ping Federate that have alternate signature verification rules.

From [Ping Identity Developers Guide](https://www.pingidentity.com/developer/en/resources/openid-connect-developers-guide.html#id_token):
>Note: Signature validation is only required for tokens not received directly from the token endpoint (i.e. for the Implicit Client Profile). In other cases where the id_token is received directly by the client from the token endpoint over HTTPS, transport layer security should be sufficient to vouch for the integrity of the token.

And
image

These rules are valid according to [the OpenID Connect specification](http://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation):
>If the ID Token is received via direct communication between the Client and the Token Endpoint (which it is in this flow), the TLS server validation MAY be used to validate the issuer in place of checking the token signature. The Client MUST validate the signature of all other ID Tokens according to JWS [JWS] using the algorithm specified in the JWT alg Header Parameter. The Client MUST use the keys provided by the Issuer.

There is a method within `OidcAuthorizationCodeAuthenticationProvider` to dynamically lookup a `JwtDecoder` for a given `ClientRegistration`:
```
private JwtDecoder getJwtDecoder(ClientRegistration clientRegistration) {
```
A few challenges here:
1. That method is `private`
2. There is no way to plug in sublcasses or alternate mappings in the `.oauth2Login()` configuration DSL

As such, there need to be changes to the core classes to enable such pluggability.

jgrandja:
Resolves #5465

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/ReactiveOidcIdTokenDecoderFactory.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/ReactiveOidcIdTokenDecoderFactory.java
@@ -20,6 +20,9 @@ import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
 import org.springframework.security.oauth2.core.OAuth2Error;
 import org.springframework.security.oauth2.core.OAuth2TokenValidator;
 import org.springframework.security.oauth2.core.oidc.OidcIdToken;
+import org.springframework.security.oauth2.jose.jws.JwsAlgorithm;
+import org.springframework.security.oauth2.jose.jws.MacAlgorithm;
+import org.springframework.security.oauth2.jose.jws.SignatureAlgorithm;
 import org.springframework.security.oauth2.jwt.Jwt;
 import org.springframework.security.oauth2.jwt.NimbusReactiveJwtDecoder;
 import org.springframework.security.oauth2.jwt.ReactiveJwtDecoder;
@@ -27,9 +30,15 @@ import org.springframework.security.oauth2.jwt.ReactiveJwtDecoderFactory;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
+import javax.crypto.spec.SecretKeySpec;
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
+
+import static org.springframework.security.oauth2.jwt.NimbusReactiveJwtDecoder.withJwkSetUri;
+import static org.springframework.security.oauth2.jwt.NimbusReactiveJwtDecoder.withSecretKey;
 
 /**
  * A {@link ReactiveJwtDecoderFactory factory} that provides a {@link ReactiveJwtDecoder}
@@ -45,14 +54,45 @@ import java.util.function.Function;
  */
 public final class ReactiveOidcIdTokenDecoderFactory implements ReactiveJwtDecoderFactory<ClientRegistration> {
 	private static final String MISSING_SIGNATURE_VERIFIER_ERROR_CODE = "missing_signature_verifier";
+	private static Map<JwsAlgorithm, String> jcaAlgorithmMappings = new HashMap<JwsAlgorithm, String>() {
+		{
+			put(MacAlgorithm.HS256, "HmacSHA256");
+			put(MacAlgorithm.HS384, "HmacSHA384");
+			put(MacAlgorithm.HS512, "HmacSHA512");
+		}
+	};
 	private final Map<String, ReactiveJwtDecoder> jwtDecoders = new ConcurrentHashMap<>();
 	private Function<ClientRegistration, OAuth2TokenValidator<Jwt>> jwtValidatorFactory = OidcIdTokenValidator::new;
+	private Function<ClientRegistration, JwsAlgorithm> jwsAlgorithmResolver = clientRegistration -> SignatureAlgorithm.RS256;
 
 	@Override
 	public ReactiveJwtDecoder createDecoder(ClientRegistration clientRegistration) {
 		Assert.notNull(clientRegistration, "clientRegistration cannot be null");
 		return this.jwtDecoders.computeIfAbsent(clientRegistration.getRegistrationId(), key -> {
-			if (!StringUtils.hasText(clientRegistration.getProviderDetails().getJwkSetUri())) {
+			NimbusReactiveJwtDecoder jwtDecoder = buildDecoder(clientRegistration);
+			OAuth2TokenValidator<Jwt> jwtValidator = this.jwtValidatorFactory.apply(clientRegistration);
+			jwtDecoder.setJwtValidator(jwtValidator);
+			return jwtDecoder;
+		});
+	}
+
+	private NimbusReactiveJwtDecoder buildDecoder(ClientRegistration clientRegistration) {
+		JwsAlgorithm jwsAlgorithm = this.jwsAlgorithmResolver.apply(clientRegistration);
+		if (jwsAlgorithm != null && SignatureAlgorithm.class.isAssignableFrom(jwsAlgorithm.getClass())) {
+			// https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation
+			//
+			// 6. If the ID Token is received via direct communication between the Client
+			// and the Token Endpoint (which it is in this flow),
+			// the TLS server validation MAY be used to validate the issuer in place of checking the token signature.
+			// The Client MUST validate the signature of all other ID Tokens according to JWS [JWS]
+			// using the algorithm specified in the JWT alg Header Parameter.
+			// The Client MUST use the keys provided by the Issuer.
+			//
+			// 7. The alg value SHOULD be the default of RS256 or the algorithm sent by the Client
+			// in the id_token_signed_response_alg parameter during Registration.
+
+			String jwkSetUri = clientRegistration.getProviderDetails().getJwkSetUri();
+			if (!StringUtils.hasText(jwkSetUri)) {
 				OAuth2Error oauth2Error = new OAuth2Error(
 						MISSING_SIGNATURE_VERIFIER_ERROR_CODE,
 						"Failed to find a Signature Verifier for Client Registration: '" +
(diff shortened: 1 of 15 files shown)


oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusReactiveJwtDecoderTests.java
```diff
@@ -205,4 +219,70 @@ public void setClaimSetConverterWhenNullThrowsIllegalArgumentException() {
...
+		this.decoder = new NimbusReactiveJwtDecoder(secretKey, JwsAlgorithms.HS512);
+
+		assertThatThrownBy(() -> this.decoder.decode(signedJWT.serialize()).block())
+				.isInstanceOf(JwtException.class);
+	}
+
+	private SecretKey secretKey() throws NoSuchAlgorithmException {
+		KeyGenerator keyGenerator = KeyGenerator.getInstance("AES");
+		keyGenerator.init(256);
+		return keyGenerator.generateKey();
```
jzheaux: It'd be nice to just use a static key here and use `SecretKeySpec`. It would make the test faster.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/ReactiveOidcIdTokenDecoderFactory.java
```diff
@@ -80,4 +138,28 @@ public final void setJwtValidatorFactory(Function<ClientRegistration, OAuth2Toke
...
+
+	private static String getJcaAlgorithm(String jwsAlgorithm) {
+		if (JwsAlgorithms.HS256.equals(jwsAlgorithm)) {
+			return "HmacSHA256";
+		} else if (JwsAlgorithms.HS384.equals(jwsAlgorithm)) {
+			return "HmacSHA384";
+		} else if (JwsAlgorithms.HS512.equals(jwsAlgorithm)) {
+			return "HmacSHA512";
+		}
+		return "";
```
jzheaux: Seems like we'd want to error here since this would indicate an unexpected state (you are already checking `isHMAC` earlier so to get to this line would indicate an error).

jgrandja:
Thanks for the feedback @rwinch @jzheaux. I've resolved all comments and pushed updates.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtProcessors.java
```diff
@@ -235,4 +247,54 @@ public PublicKeyJwtProcessorBuilder jwsAlgorithm(String jwsAlgorithm) {
...
+		 * when generating the MAC.
+		 *
+		 * The value should be one of
+		 * <a href="https://tools.ietf.org/html/rfc7518#section-3.2" target="_blank">HS256, HS384 or HS512</a>.
+		 *
+		 * @param jwsAlgorithm the JWS algorithm to use
+		 * @return a {@link SecretKeyJwtProcessorBuilder} for further configurations
+		 */
+		public SecretKeyJwtProcessorBuilder jwsAlgorithm(JwsAlgorithm jwsAlgorithm) {
+			Assert.notNull(jwsAlgorithm, "jwsAlgorithm cannot be null");
```
rwinch: I'd still like to see if we can have type safety here vs checking a boolean.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusReactiveJwtDecoder.java
```diff
@@ -92,6 +96,32 @@ public NimbusReactiveJwtDecoder(RSAPublicKey publicKey) {
...
+	 * Constructs a {@code NimbusReactiveJwtDecoder} using the provided {@code SecretKey}.
+	 *
+	 * @since 5.2
+	 * @param secretKey the {@code SecretKey}
+	 * @param jwsAlgorithm the {@link JwsAlgorithm JWS algorithm}
+	 */
+	public NimbusReactiveJwtDecoder(SecretKey secretKey, JwsAlgorithm jwsAlgorithm) {
+		Assert.notNull(secretKey, "secretKey cannot be null");
+		Assert.notNull(jwsAlgorithm, "jwsAlgorithm cannot be null");
+		Assert.isTrue(jwsAlgorithm.isHMAC(),
```
rwinch: Another example of where it would be nice to have type safety.

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jose/TestKeys.java
```diff
@@ -0,0 +1,42 @@
...
+	static {
+		try {
+			DEFAULT_SECRET_KEY = secretKey();
+		} catch (NoSuchAlgorithmException ex) {
+			throw new RuntimeException(ex);
+		}
+	}
+
+	public static SecretKey secretKey() throws NoSuchAlgorithmException {
+		KeyGenerator keyGenerator = KeyGenerator.getInstance("AES");
```
jzheaux: Is there any reason to have this be generated? It would be nice if it were a static value, e.g.

```suggestion
		return new SecretKeySpec(Base64.getDecoder().decode("Ky8xmu1fg/OqVlUr9PRKhutauHvuj0rLHExRk+5XkSU="), "AES")
```

jgrandja: There is no particular reason. I feel `secretKey()` can be useful as some tests may require their own secret. Is there a benefit to having a static value other than the one-time optimization at class load?

jzheaux: Having a "pre-generated" one is faster. To be clear, I think `secretKey()`, should stay, it should just use load a pre-generated key instead of a live-generated key.

Also, taking a look at a few other tests, it is fairly common to use a pre-generated value, for example, `NimbusReactiveJwtDecoderTests#decodeWhenRSAPublicKeyThenSuccess`, and the resource server tests.

jgrandja: > Having a "pre-generated" one is faster

True, but the generated key is done once at class load so I don't think this will make the tests run any much faster.

jgrandja:
@rwinch I've implemented dedicated types for the JWS algorithms.

```java
package org.springframework.security.oauth2.jose.jws;

public enum MacAlgorithm implements JwaAlgorithm {
   ...

public enum SignatureAlgorithm implements JwaAlgorithm {
   ...

```

I've also considered this design for when we implement support for JWE.
I'm thinking we would introduce the following:

```java
package org.springframework.security.oauth2.jose.jwe;

public enum SecreKeyEncryptionAlgorithm implements JwaAlgorithm {
   ...

public enum PublicKeyEncryptionAlgorithm implements JwaAlgorithm {
   ...

```

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtProcessors.java
```diff
@@ -235,4 +249,52 @@ public PublicKeyJwtProcessorBuilder jwsAlgorithm(String jwsAlgorithm) {
...
+		 * <a href="https://tools.ietf.org/html/rfc7515#section-4.1.1" target="_blank">algorithm</a>
+		 * when generating the MAC.
+		 *
+		 * The value should be one of
+		 * <a href="https://tools.ietf.org/html/rfc7518#section-3.2" target="_blank">HS256, HS384 or HS512</a>.
+		 *
+		 * @param jwsAlgorithm the JWS algorithm to use
+		 * @return a {@link SecretKeyJwtProcessorBuilder} for further configurations
+		 */
+		public SecretKeyJwtProcessorBuilder jwsAlgorithm(MacAlgorithm jwsAlgorithm) {
```
rwinch: Does it make sense to rename this method and argument now that it is of type MacAlgorithm?

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jose/TestKeys.java
rwinch: I'd prefer to have this be a static value too. This is not only faster, but it ensures the tests are consistent and predictable.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusReactiveJwtDecoder.java
```diff
@@ -92,6 +96,30 @@ public NimbusReactiveJwtDecoder(RSAPublicKey publicKey) {
...
 	}
 
+	/**
+	 * Constructs a {@code NimbusReactiveJwtDecoder} using the provided {@code SecretKey}.
+	 *
+	 * @since 5.2
+	 * @param secretKey the {@code SecretKey}
+	 * @param jwsAlgorithm the {@link MacAlgorithm JWS algorithm}
+	 */
+	public NimbusReactiveJwtDecoder(SecretKey secretKey, MacAlgorithm jwsAlgorithm) {
```
rwinch: Should we rename jwsAlgorithm to macAlgorithm?

jgrandja:
@rwinch @jzheaux I've addressed all the feedback so please go ahead with the review.

If we can get this into M2, that would be ideal.

jgrandja:
@rwinch @jzheaux There were quite a few changes to `NimbusJwtDecoder` and `NimbusReactiveJwtDecoder` since I initially submitted this PR, e.g. `JwtProcessors` was removed in 9478abe which my PR was using.

This resulted in quite a few conflicts and was very difficult to merge. So I had no choice but to start from master and apply my updates manually and force push. 

I know timing is tight to get this into M2 but it would be ideal and this one has been around for quite some time.

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jose/TestKeys.java
jzheaux: Looking through the tests, I don't see this method being used. Let's remove it, adding it back in when there is a clear need for a test to have a dynamically-generated secret key.

jgrandja:
Merged via bed3371

PR #6496, opened 2019-02-01
AuthenticationSupplier feature for Web Security

sbespalov:
This PR targets to make Authentication Filters implementation to be more *"SOLID"*.
Currently there are set of different Atuhenticaion Filters like `BasicAuthenticationFilter`, `DigestAuthenticationFilter` etc. and most of such filters logic is almost the same:
1. expose `Authentication` from request (supply);
2. proceed filter chain, if there is no supported `Authentication` provided
3. authenticate the `Authentication` object with `AuthenticationManager`, if we have supported `Authentication`

This PR contains following main changes:
1. Added `AuthenticationSupplier` interface;
2. Added `GenericAuthenticationFilter` class;
3. Basic Authentication support implemented with `BasicAuthenticationSupplier` and `GenericAuthenticationFilter`;

If this concept will be accepted in general then additional changes can be provided.

### DIFF
--- a/cas/src/main/java/org/springframework/security/cas/web/CasAuthenticationFilter.java
+++ b/cas/src/main/java/org/springframework/security/cas/web/CasAuthenticationFilter.java
@@ -242,7 +242,7 @@ public class CasAuthenticationFilter extends AbstractAuthenticationProcessingFil
 	@Override
 	public Authentication attemptAuthentication(final HttpServletRequest request,
 			final HttpServletResponse response) throws AuthenticationException,
-			IOException {
+			IOException, ServletException {
 		// if the request is a proxy request process it and return null to indicate the
 		// request has been processed
 		if (proxyReceptorRequest(request)) {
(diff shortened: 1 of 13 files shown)


pivotal-issuemaster:
@sbespalov Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=6496)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=6496) to manually synchronize the status of this Pull Request.

rwinch:
Thanks for the PR @sbespalov!

In general, there are too many changes in this one PR. For example, there are polish actions that add ServletException where it is not necessary.

In regards to the Supplier, I think it may be a good idea, but let's not combine AuthenticationSupplier and AuthenticationEntryPoint. If you look at the reactive bits it uses a ServerAuthenticationConverter which is closer to what I would have in mind.

I'm going to close this PR. If you are still wanting to see this support, please create a ticket for us to discuss a bit so we can be on the same page before writing more code. This will ensure you don't spend a lot of time coding to find out that we are not yet on the same page.

sbespalov:
@rwinch thanks for your response.

[6506](https://github.com/spring-projects/spring-security/issues/6506)

Issue #6494, opened 2019-01-31
Add RSA Key Converters


jzheaux:
Spring Security supports supplying an RSA public key in its OAuth 2.0 APIs.

Local keys, though, are often distributed in files or as a PEM-encoded configuration property. It would be nice to have some support for converting an encoded RSA key into the appropriate JCA instance, e.g. `RSAPublicKey`.

One way to do this would be by using the `Converter` interface:

```java
public class RSAPublicKeyConverter implements Converter<String, RSAPublicKey> {
    public RSAPublicKey converter(String value) {
        // ...
    }
}
```

Which then users could use directly or could be registered with a `ConversionService`.

PR #6505, opened 2019-02-04
Introduce Support for Reading RSA Keys

jzheaux:
Fixes: gh-6494


crypto/src/main/java/org/springframework/security/crypto/factory/RsaKeyConverters.java
```diff
@@ -0,0 +1,119 @@
...
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.util.Assert;
+
+/**
+ * Used for creating {@link java.security.Key} converter instances
+ *
+ * @author Josh Cummings
+ * @since 5.2
+ */
+public class RsaKeyConverters {
```
rwinch: Perhaps we should consider changing the input to be an `InputStream`? This would allow for more flexibility in terms of how the key is read.

jzheaux:
@rwinch, I've added to more commits to this PR which address integrating with the application context. 

I'm partial to actually having that in a separate PR, but if you feel like the changes are largely obvious/non-controversial, then I'm fine reviewing it all here.

rwinch:
As discussed offline I will give you some time to polish this (i.e. move addConverters to be private within the post processor)

jzheaux:
@rwinch, I've applied some polish. First, I moved the converter registration into the post-processor. Also, the post-processor has been simplified a bit, to implement `BeanFactoryPostProcessor` instead of `BeanDefinitionRegistryPostProcessor`.

Second, I consolidated and simplified the ApplicationContext-integration tests.

I think this looks much better now, and I'm okay with having everything reviewed together.

config/src/main/java/org/springframework/security/config/annotation/web/configuration/ConversionServicePostProcessor.java
```diff
@@ -0,0 +1,88 @@
...
+import org.springframework.core.io.ResourceLoader;
+import org.springframework.security.crypto.factory.RsaKeyConverters;
+
+/**
+ * Adds Spring Security converters to the configured {@link ConversionService}
+ *
+ * @author Josh Cummings
+ * @since 5.2
+ */
+public class ConversionServicePostProcessor implements BeanFactoryPostProcessor {
```
rwinch: I wonder if there is a reason this is so generic given it adds RSA converts. I also wonder if it makes sense in the current package since RSA is applicable to non web apps and to reactive applications.

config/src/main/java/org/springframework/security/config/annotation/web/configuration/ConversionServicePostProcessor.java
```diff
@@ -0,0 +1,88 @@
...
+		Converter<String, InputStream> pemInputStreamConverter = pemInputStreamConverter();
+		converterRegistry.addConverter(String.class, RSAPrivateKey.class,
+				pair(pemInputStreamConverter, RsaKeyConverters.pkcs8()));
+		converterRegistry.addConverter(String.class, RSAPublicKey.class,
+				pair(pemInputStreamConverter, RsaKeyConverters.x509()));
+	}
+
+	private Converter<String, InputStream> pemInputStreamConverter() {
+		ResourceLoader resourceLoader = new DefaultResourceLoader();
+		return source -> source.startsWith("-----") ?
```
rwinch: Can you help me understand what the conditional logic is for? It looks like this might be to handle things like `classpath:foo.pem`. However, my concern is that it should not interfere with existing String to `InputStream` conversions.

config/src/main/java/org/springframework/security/config/annotation/web/configuration/ConversionServicePostProcessor.java
```diff
@@ -0,0 +1,88 @@
...
+	private void addRsaConverters(ConverterRegistry converterRegistry) {
+		Converter<String, InputStream> pemInputStreamConverter = pemInputStreamConverter();
+		converterRegistry.addConverter(String.class, RSAPrivateKey.class,
+				pair(pemInputStreamConverter, RsaKeyConverters.pkcs8()));
+		converterRegistry.addConverter(String.class, RSAPublicKey.class,
+				pair(pemInputStreamConverter, RsaKeyConverters.x509()));
+	}
+
+	private Converter<String, InputStream> pemInputStreamConverter() {
+		ResourceLoader resourceLoader = new DefaultResourceLoader();
```
rwinch: I think we should make the ResourceLoader a member variable and allow it to be injected.

config/src/main/java/org/springframework/security/config/annotation/web/configuration/ConversionServicePostProcessor.java
rwinch: Does this work if a variable (i.e. `${some.path}` is used)? I cannot recall if anything special is necessary, but it has bitten me before.

config/src/main/java/org/springframework/security/config/annotation/web/configuration/ConversionServicePostProcessor.java
```diff
@@ -0,0 +1,88 @@
...
+ * @author Josh Cummings
+ * @since 5.2
+ */
+public class ConversionServicePostProcessor implements BeanFactoryPostProcessor {
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory) throws BeansException {
+		ConversionService service = beanFactory.getConversionService();
```
rwinch: Does Spring always register a `ConversionService` (XML, Java Config, and Boot)?

config/src/main/java/org/springframework/security/config/annotation/web/configuration/ConversionServicePostProcessor.java
jzheaux: Regarding your first question, the actual registered converter is `String -> RSAPublicKey` so it will only be invoked when the target type is `RSAPublicKey`. Yes, the conditional logic is to differentiate between a raw key and one referred to via `classpath:foo.pem`.

I didn't test your second case, but I will double-check this.

config/src/main/java/org/springframework/security/config/annotation/web/configuration/ConversionServicePostProcessor.java
jzheaux: No, XML and Java Config do not register a `ConversionService`. If the user (or Boot) does not supply one then Spring will fall back to the PropertyEditor conversion style.

I didn't look into property editors - would you recommend that in order to provide more in-depth support?

config/src/main/java/org/springframework/security/config/annotation/web/configuration/ConversionServicePostProcessor.java
jzheaux: A quick search across the code base shows a handful of classes that extend `BeanFactoryPostProcessor`. For example, there is one for `ldap`, one for `debug`, and one for `websocket`, among others.

Each of these is in `spring-security-config` and follows the pattern `org.springframework.security.config.{module}`.

So, it seems natural to rename this as `org.springframework.security.config.crypto.RsaKeyConversionServicePostProcessor` since it is post processing the `ConversionService` to support RSA keys.

config/src/main/java/org/springframework/security/config/annotation/web/configuration/ConversionServicePostProcessor.java
jzheaux: I changed the sample to use `${some.property.name}` to confirm that works.

jzheaux:
@rwinch, I've responded to your comments. Let me know what you think.

crypto/src/main/java/org/springframework/security/crypto/factory/RsaKeyConverters.java
```diff
@@ -0,0 +1,123 @@
...
+	 * If the key file begins with "-----BEGIN RSA PRIVATE KEY-----", then it is PKCS#1. If it is
+	 * PKCS#8 formatted, then it begins with "-----BEGIN PRIVATE KEY-----".
+	 *
+	 * @return A {@link Converter} that can read a PEM-encoded PKCS#8 RSA Private Key and return a
+	 * {@link RSAPrivateKey}.
+	 */
+	public static Converter<InputStream, RSAPrivateKey> pkcs8() {
+		KeyFactory keyFactory = rsaFactory();
+		return source -> {
+			BufferedReader reader = new BufferedReader(new InputStreamReader(source));
```
rwinch: Perhaps I'm missing something, but shouldn't we close the `InputStream` after reading it? If not, we should make it clear we don't close the `InputStream` in the Javadoc and explain why.

crypto/src/main/java/org/springframework/security/crypto/factory/RsaKeyConverters.java
```diff
@@ -0,0 +1,123 @@
...
+	 * Construct a {@link Converter} for converting a PEM-encoded X.509 RSA Public Key
+	 * into a {@link RSAPublicKey}.
+	 *
+	 * @return A {@link Converter} that can read a PEM-encoded X.509 RSA Public Key and return a
+	 * {@link RSAPublicKey}.
+	 */
+	public static Converter<InputStream, RSAPublicKey> x509() {
+		KeyFactory keyFactory = rsaFactory();
+		return source -> {
+			BufferedReader reader = new BufferedReader(new InputStreamReader(source));
```
rwinch: Perhaps I'm missing something, but shouldn't we close the `InputStream` after reading it? If not, we should make it clear we don't close the `InputStream` in the Javadoc and explain why.

crypto/src/main/java/org/springframework/security/crypto/factory/RsaKeyConverters.java
```diff
@@ -0,0 +1,123 @@
...
+			List<String> lines = reader.lines().collect(Collectors.toList());
+			Assert.isTrue(!lines.isEmpty() && lines.get(0).startsWith(X509_PEM_HEADER),
+					"Key is not in PEM-encoded X.509 format");
```
rwinch: Perhaps add details about the X509_PEM_HEADER is missing to the error message?

crypto/src/main/java/org/springframework/security/crypto/factory/RsaKeyConverters.java
```diff
@@ -0,0 +1,123 @@
...
+	 * @return A {@link Converter} that can read a PEM-encoded PKCS#8 RSA Private Key and return a
+	 * {@link RSAPrivateKey}.
+	 */
+	public static Converter<InputStream, RSAPrivateKey> pkcs8() {
+		KeyFactory keyFactory = rsaFactory();
+		return source -> {
+			BufferedReader reader = new BufferedReader(new InputStreamReader(source));
+			List<String> lines = reader.lines().collect(Collectors.toList());
+			Assert.isTrue(!lines.isEmpty() && lines.get(0).startsWith(PKCS8_PEM_HEADER),
+					"Key is not in PEM-encoded PKCS#8 format");
```
rwinch: Perhaps add details about PKCS8_PEM_HEADER missing to the error message?

crypto/src/main/java/org/springframework/security/crypto/factory/RsaKeyConverters.java
jzheaux: Good catch. My intent was to not make an assumption about the nature of the stream being passed in. So, I'll add a Javadoc to both methods.

That said, I think it makes sense to close it in the post processor since it's the post processor that opens it, so I'll change that, too.

crypto/spring-security-crypto.gradle
```diff
@@ -1,6 +1,7 @@
 apply plugin: 'io.spring.convention.spring-module'
 
 dependencies {
+	optional 'org.springframework:spring-core'
```
rwinch: Sorry for not realizing this till late in the game...I'm realizing we have a tangle here. The core includes crypto and now we have optional dependency on core. I think we need to move the crypo classes that use core into core.

jzheaux:
@rwinch I moved `RsaKeyConverters` and `RsaKeyConvertersTest` over to `core`. Otherwise, I simply cleaned up the commits a bit so now we just have two: One for the new feature and one for the change in one of the resource server samples.

jzheaux:

PR #6508, opened 2019-02-06
Introduce OAuth2AuthorizationRequest.attributes

### LINKED ISSUE #5940: OAuth2AuthorizationRequest.additionalParameters should not contain registration_id

jgrandja:
Fixes #5940

### DIFF
--- a/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolverTests.java
+++ b/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolverTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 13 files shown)


oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolverTests.java
```diff
@@ -105,7 +105,7 @@ public void resolveWhenAuthorizationRequestWithValidClientThenResolves() {
 				.isEqualTo("http://localhost/login/oauth2/code/" + clientRegistration.getRegistrationId());
 		assertThat(authorizationRequest.getScopes()).isEqualTo(clientRegistration.getScopes());
 		assertThat(authorizationRequest.getState()).isNotNull();
-		assertThat(authorizationRequest.getAdditionalParameters())
```
rwinch: Since this is to fix gh-5940 we should assert that the additional parameters do not contain the registration id (just because it is in the attributes -- the next assertion -- doesn't mean it isn't also in the parameters)

jgrandja: Makes sense. Applied and pushed.

jgrandja:
Thanks @rwinch. I tried building locally and it passed.

jgrandja:
Merged via 594a169

Issue #6501, opened 2019-02-01
Add support for HeaderWriterFilter to write headers at the beginning of the request


jzheaux:
Related to https://github.com/spring-projects/spring-security/issues/6414

Let's add some support to `HeaderWriterFilter` for writing headers at the beginning of the request:

```java
public void setShouldWriteHeadersEagerly(boolean shouldWriteHeadersEagerly) {
    this.shouldWriteHeadersEagerly = shouldWriteHeadersEagerly;
}
```

Which would then allow folks to post process the `HeaderWriterFilter` and configure it:

```java
http
    .headers()
        .withObjectPostProcessor(new ObjectPostProcessor<HeaderWriterFilter>() {
            public HeaderWriterFilter postProcess(HeaderWriterFilter filter) {
                filter.setShouldWriteHeadersEagerly(true);
                return filter;
            }
        }  
```

`HeaderWriterFilter` writes headers at the end of the request. However, headers cannot be written after an include. 

If the user uses `servletContext.getRequestDispatcher(path).include(request, response)`, for example, then the security headers will not get written.

Spring Security is able to wrap `HttpServletRequest` so that `request.getRequestDispatcher(path).include(request, response)` is captured, but it is not possible to do the same with `ServletContext`.

ankurpathak:
I would like to take this one.

jzheaux:
It's yours, @ankurpathak!

ankurpathak:
Pull request for the issue:
https://github.com/spring-projects/spring-security/pull/6509

PR #6509, opened 2019-02-06
HeaderWriterFilter writes headers at beginning

ankurpathak:
Add support for HeaderWriterFilter to write headers at the beginning of the request

Fixes: gh-6501

### DIFF
--- a/web/src/main/java/org/springframework/security/web/header/HeaderWriterFilter.java
+++ b/web/src/main/java/org/springframework/security/web/header/HeaderWriterFilter.java
@@ -52,6 +52,11 @@ public class HeaderWriterFilter extends OncePerRequestFilter {
 	private final HeaderWriter headerWriter;
 
 	/**
+	 * Indicates whether to write the headers at the beginning of the request.
+	 */
+	private boolean shouldWriteHeadersEagerly = false;
+
+	/**
 	 * Creates a new instance.
 	 *
 	 * @param headerWriters the {@link HeaderWriter} instances to write out headers to the
@@ -67,16 +72,39 @@ public class HeaderWriterFilter extends OncePerRequestFilter {
 			HttpServletResponse response, FilterChain filterChain)
 			throws ServletException, IOException {
 
+		if (this.shouldWriteHeadersEagerly) {
+			doHeadersBefore(request, response, filterChain);
+		} else {
+			doHeadersAfter(request, response, filterChain);
+		}
+	}
+
+	private void doHeadersBefore(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain) throws IOException, ServletException {
+		this.headerWriter.writeHeaders(request, response);
+		filterChain.doFilter(request, response);
+	}
+
+	private void doHeadersAfter(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain) throws IOException, ServletException {
 		HeaderWriterResponse headerWriterResponse = new HeaderWriterResponse(request,
 				response, this.headerWriter);
 		HeaderWriterRequest headerWriterRequest = new HeaderWriterRequest(request,
 				headerWriterResponse);
-
 		try {
 			filterChain.doFilter(headerWriterRequest, headerWriterResponse);
 		} finally {
 			headerWriterResponse.writeHeaders();
 		}
+	}
+
+	/**
+	 * Allow writing headers at the beginning of the request.
+	 *
+	 * @param shouldWriteHeadersEagerly boolean to allow writing headers at the beginning of the request.
+	 * @author Ankur Pathak
+	 * @since 5.2
+	 */
+	public void setShouldWriteHeadersEagerly(boolean shouldWriteHeadersEagerly) {
+		this.shouldWriteHeadersEagerly = shouldWriteHeadersEagerly;
 	}
 
 	static class HeaderWriterResponse extends OnCommittedResponseWrapper {
--- a/config/src/test/java/org/springframework/security/config/annotation/authentication/configurers/HeadersConfigurerJavaTests.java
+++ b/config/src/test/java/org/springframework/security/config/annotation/authentication/configurers/HeadersConfigurerJavaTests.java
@@ -1,0 +1,131 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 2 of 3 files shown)


ankurpathak:
@jzheaux I am not getting any Idea for tests. Please give some idea how I can write test.

web/src/main/java/org/springframework/security/web/header/HeaderWriterFilter.java
```diff
@@ -73,12 +78,28 @@ protected void doFilterInternal(HttpServletRequest request,
...
 	}
 
+	/**
+	 * Allow writing headers at the beginning of the request.
+	 *
+	 * @param shouldWriteHeadersEagerly boolean to allow writing headers at the beginning of the request.
+	 * @author Ankur Pathak
+	 * @since 5.2
+	 */
+	public void setShouldWriteHeadersEagerly(boolean shouldWriteHeadersEagerly) {
```
jzheaux: Let's go ahead and add a couple of tests that confirm that this functionality works. 

One test in `HeaderWriterFilterTests` that confirms that when using the setter, then the headers are written early.

And one in `HeadersConfigurerTests` that confirms that the setting can be made via an `ObjectPostProcessor`. (If you haven't done that before, you can see an example in the ticket description.) In this case, I think you can just confirm that the headers still get written.

ankurpathak:
```
http
    .addObjectPostProcessor(new ObjectPostProcessor<HeaderWriterFilter>() {
        public HeaderW riterFilter postProcess(HeaderWriterFilter filter) {
            filter.setShouldWriteHeadersEagerly(true);
            return filter;
        }
    }
```
I can not find any addObjectPostProcessor method in HttpSecurity. Please check this url:
https://docs.spring.io/spring-security/site/docs/5.2.0.BUILD-SNAPSHOT/api/org/springframework/security/config/annotation/web/builders/HttpSecurity.html

ankurpathak:
```@EnableWebSecurity
	static class HeadersAtTheBeginningOfRequestConfig extends WebSecurityConfigurerAdapter {
        @Override
            http
                    .headers()
                    .addObjectPostProcessor(new ObjectPostProcessor<HeaderWriterFilter>() {
						@Override
						HeaderWriterFilter postProcess(HeaderWriterFilter filter) {
							filter.shouldWriteHeadersEagerly = true
							return filter
						}
					})
        }
    }
```
This inside groovy is giving me errors below:
screenshot from 2019-02-08 09-11-54

ankurpathak: I am done with changes.

web/src/main/java/org/springframework/security/web/header/HeaderWriterFilter.java
```diff
@@ -73,12 +78,28 @@ protected void doFilterInternal(HttpServletRequest request,
 				headerWriterResponse);
 
 		try {
+			if (shouldWriteHeadersEagerly) {
+				headerWriterResponse.writeHeaders();
```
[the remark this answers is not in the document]
ankurpathak: Done.

web/src/main/java/org/springframework/security/web/header/HeaderWriterFilter.java
ankurpathak: > Let's go ahead and add a couple of tests that confirm that this functionality works.
> 
> One test in `HeaderWriterFilterTests` that confirms that when using the setter, then the headers are written early.
> 
> And one in `HeadersConfigurerTests` that confirms that the setting can be made via an `ObjectPostProcessor`. (If you haven't done that before, you can see an example in the ticket description.) In this case, I think you can just confirm that the headers still get written.

Done.

jzheaux: Looks like you found it, yes, the correct way is:

```java
http
    .headers()
        .withObjectPostProcessor(...
```

ankurpathak: @jzheaux You can review it. Its complete.

config/src/test/groovy/org/springframework/security/config/annotation/web/configurers/HeadersConfigurerTests.groovy
```diff
@@ -565,4 +567,20 @@ class HeadersConfigurerTests extends BaseSpringSpec {
 		    responseHeaders == ['Strict-Transport-Security': 'max-age=31536000 ; includeSubDomains ; preload']
 	}
 
+    def "headers written at beginning of request"() {
```
jzheaux: Please make this a Java test. We are in the process of coverting the Groovy tests over to Java, so we don't want to make more work for ourselves.

config/src/test/java/org/springframework/security/config/annotation/authentication/configurers/HeadersConfigurerTestsConfigs.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Tests config for {@link HeadersConfigurerTests}.
+ *
+ * @author Ankur Pathak
+ */
+public class HeadersConfigurerTestsConfigs {
+
+	@EnableWebSecurity
+	public static class HeadersAtTheBeginningOfRequestConfig extends WebSecurityConfigurerAdapter {
```
jzheaux: Usually, we place configs alongside the tests, as inner classes. Could we do that here, too, please?

config/src/test/groovy/org/springframework/security/config/annotation/web/configurers/HeadersConfigurerTests.groovy
ankurpathak: Done.

config/src/test/java/org/springframework/security/config/annotation/authentication/configurers/HeadersConfigurerTestsConfigs.java
ankurpathak: Done.

jzheaux:
@ankurpathak Thanks for the PR! This is now merged into `master` via ac13b55ecd93b9eaf40866aabb043b14dd4f3350. 

Also, I added a polish commit of 2b960b074b8518c81a789214e542954d82deedb8 for the Java Config test - please take a look at the commit message for future reference.

PR #6510, opened 2019-02-07
Properties should reference scope not scopes

sdoxsee:
OAuth2ClientProperties.Registration has a member `scope` but not `scopes`. This is the easy fix for the sample but i think it should be `scopes`

### DIFF
--- a/docs/manual/src/docs/asciidoc/_includes/reactive/oauth2/access-token.adoc
+++ b/docs/manual/src/docs/asciidoc/_includes/reactive/oauth2/access-token.adoc
@@ -14,7 +14,7 @@ spring:
           github:
             client-id: replace-with-client-id
             client-secret: replace-with-client-secret
-            scopes: read:user,public_repo
+            scope: read:user,public_repo
 ----
 
 You will need to replace the `client-id` and `client-secret` with values registered with GitHub.
--- a/samples/boot/oauth2webclient-webflux/README.adoc
+++ b/samples/boot/oauth2webclient-webflux/README.adoc
@@ -40,7 +40,7 @@ spring:
             client-id: replace-with-client-id
             client-secret: replace-with-client-secret
             provider: github
-            scopes: read:user,public_repo
+            scope: read:user,public_repo
 ----
 +
 .OAuth Client properties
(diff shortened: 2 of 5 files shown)


jgrandja:
Thanks for the PR @sdoxsee. Would you be able to update the same in `oauth2webclient-webflux`?

sdoxsee:
done. Thanks!

rwinch:
Thanks @sdoxsee It looks like there are few additional places. Can you update the PR to get those as well? Here are all the places I found with a quick search:

```bash
$ rg "scopes: "
samples/boot/oauth2webclient-webflux/README.adoc
43:            scopes: read:user,public_repo

samples/boot/oauth2webclient/README.adoc
43:            scopes: read:user,public_repo

samples/boot/oauth2webclient/src/main/resources/application.yml
19:            scopes: read:user,public_repo

samples/boot/oauth2webclient-webflux/src/main/resources/application.yml
19:            scopes: read:user,public_repo

docs/manual/src/docs/asciidoc/_includes/reactive/oauth2/access-token.adoc
17:            scopes: read:user,public_repo
```

sdoxsee:
Thanks @rwinch. I've got those now as well. Would it make more sense to change the `OAuth2ClientProperties.Registration.scope` member to  `scopes` rather than change other places to use `scope`?

jgrandja:
@sdoxsee In regards to...

> Would it make more sense to change the `OAuth2ClientProperties.Registration.scope` member to `scopes`

The name `scope` is intentional to align with the [spec](https://tools.ietf.org/html/rfc6749#section-3.3). You than may be wondering why the plural name `Set<String> ClientRegistration.getScopes()`? It's because it returns a `Set`. This is the naming convention we adopted a ways back.

The changes look great in this PR and it's ready to get merged. Can you please squash to 1 commit and ensure the commit message follows this [format](https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.md#format-commit-messages) and includes `Fixes gh-*` at the end.
Thanks!

jgrandja:
Thanks for the PR @sdoxsee! This is now in master.

sdoxsee:
Awesome. Thanks @jgrandja

Issue #6513, opened 2019-02-07
Add Reactive Opaque Token Support to Resource Server


jzheaux:
The reactive equivalent for #5200

PR #6519, opened 2019-02-08
Reactive Opaque Token Support

jzheaux:
Fixes: gh-6513

### DIFF
--- a/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
@@ -16,6 +16,24 @@
 
 package org.springframework.security.config.web.server;
 
+import java.io.IOException;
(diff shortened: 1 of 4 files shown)


config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -1128,17 +1214,10 @@ public OAuth2ResourceServerSpec and() {
 			}
 
 			protected void configure(ServerHttpSecurity http) {
-				this.bearerTokenServerWebExchangeMatcher.setBearerTokenConverter(bearerTokenConverter);
```
jzheaux: Note that this code was extracted from `JwtSpec` so that it could be executed for `OpaqueTokenSpec` as well.

rwinch:
Thanks for the PR @jzheaux! I have merged this into master

Issue #6483, opened 2019-01-25
ServletOAuth2AuthorizedClientExchangeFilterFunction Does Not Work For Chained Reactive Methods


rwinch:
### Summary

The defaults for ServletOAuth2AuthorizedClientExchangeFilterFunction are performed on the main thread so that ThreadLocal's like SecurityContextHolder and RequestContextHolder can be accessed. The problem is that when chaining occurs, the thread has also chained. This means these defaults cannot be accessed.

An example would be something like this:

```kotlin
override fun getPermissionForDefaultLogin(repositoryRef: RepositoryRef): Mono<Permission> {
    return defaultGitHubLogin()
            // getPermissionsForLogin is deferred to another thread which means defaults cannot be found on the ThreadLocal objects
            .flatMap { login -> getPermissionForLogin(repositoryRef, login) }
}

private fun getPermissionForLogin(repositoryRef: RepositoryRef, login: String): Mono<Permission> {
    return webClient.get()
            .uri("$baseGitHubUrl/repos/${repositoryRef.fullName}/collaborators/$login/permission")
            .attributes(clientRegistrationId("github"))
            .retrieve()
            .bodyToMono<Map<String,Object>>()
            .map { body -> body.get("permission")?.toString()!! }
            .map { p -> Permission(login, p) }
}

private fun defaultGitHubLogin(): Mono<String> {
    return webClient.get()
            .uri("$baseGitHubUrl/user")
            .retrieve()
            .bodyToMono<Map<String,Object>>()
            .map { body -> body.get("login")?.toString() }
}
```

We should find a way to ensure that the defaults are propagated to chained methods (likely using Reactor's context)

PR #6526, opened 2019-02-12
ServletOAuth2AuthorizedClientExchangeFilterFunction supports chaining

jgrandja:
Fixes gh-6483

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
@@ -16,6 +16,7 @@
 
 package org.springframework.security.oauth2.client.web.reactive.function.client;
 
+import org.reactivestreams.Subscription;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.MediaType;
@@ -44,8 +45,12 @@ import org.springframework.web.reactive.function.client.ClientResponse;
 import org.springframework.web.reactive.function.client.ExchangeFilterFunction;
 import org.springframework.web.reactive.function.client.ExchangeFunction;
 import org.springframework.web.reactive.function.client.WebClient;
+import reactor.core.CoreSubscriber;
+import reactor.core.publisher.Hooks;
 import reactor.core.publisher.Mono;
+import reactor.core.publisher.Operators;
 import reactor.core.scheduler.Schedulers;
+import reactor.util.context.Context;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -108,6 +113,8 @@ public final class ServletOAuth2AuthorizedClientExchangeFilterFunction implement
 	private static final String HTTP_SERVLET_REQUEST_ATTR_NAME = HttpServletRequest.class.getName();
 	private static final String HTTP_SERVLET_RESPONSE_ATTR_NAME = HttpServletResponse.class.getName();
 
+	private static final String REQUEST_CONTEXT_OPERATOR_KEY = RequestContextSubscriber.class.getName();
+
 	private Clock clock = Clock.systemUTC();
 
 	private Duration accessTokenExpiresSkew = Duration.ofMinutes(1);
@@ -123,11 +130,14 @@ public final class ServletOAuth2AuthorizedClientExchangeFilterFunction implement
 
 	private String defaultClientRegistrationId;
 
-	public ServletOAuth2AuthorizedClientExchangeFilterFunction() {}
+	public ServletOAuth2AuthorizedClientExchangeFilterFunction() {
+		Hooks.onLastOperator(REQUEST_CONTEXT_OPERATOR_KEY, Operators.lift((s, sub) -> createRequestContextSubscriber(sub)));
+	}
 
 	public ServletOAuth2AuthorizedClientExchangeFilterFunction(
 			ClientRegistrationRepository clientRegistrationRepository,
 			OAuth2AuthorizedClientRepository authorizedClientRepository) {
+		this();
 		this.clientRegistrationRepository = clientRegistrationRepository;
 		this.authorizedClientRepository = authorizedClientRepository;
 	}
@@ -266,13 +276,34 @@ public final class ServletOAuth2AuthorizedClientExchangeFilterFunction implement
 
 	@Override
 	public Mono<ClientResponse> filter(ClientRequest request, ExchangeFunction next) {
-		Optional<OAuth2AuthorizedClient> attribute = request.attribute(OAUTH2_AUTHORIZED_CLIENT_ATTR_NAME)
-				.map(OAuth2AuthorizedClient.class::cast);
-		return Mono.justOrEmpty(attribute)
-				.flatMap(authorizedClient -> authorizedClient(request, next, authorizedClient))
+		return Mono.just(request)
+				.filter(req -> req.attribute(OAUTH2_AUTHORIZED_CLIENT_ATTR_NAME).isPresent())
+				.switchIfEmpty(mergeRequestAttributesFromContext(request))
+				.filter(req -> req.attribute(OAUTH2_AUTHORIZED_CLIENT_ATTR_NAME).isPresent())
+				.flatMap(req -> authorizedClient(req, next, getOAuth2AuthorizedClient(req.attributes())))
 				.map(authorizedClient -> bearer(request, authorizedClient))
 				.flatMap(next::exchange)
 				.switchIfEmpty(next.exchange(request));
+	}
+
+	private Mono<ClientRequest> mergeRequestAttributesFromContext(ClientRequest request) {
+		return Mono.just(ClientRequest.from(request))
+				.flatMap(builder -> Mono.subscriberContext()
+						.map(ctx -> builder.attributes(attrs -> populateRequestAttributes(attrs, ctx))))
+				.map(ClientRequest.Builder::build);
+	}
+
+	private void populateRequestAttributes(Map<String, Object> attrs, Context ctx) {
+		if (ctx.hasKey(HTTP_SERVLET_REQUEST_ATTR_NAME)) {
+			attrs.putIfAbsent(HTTP_SERVLET_REQUEST_ATTR_NAME, ctx.get(HTTP_SERVLET_REQUEST_ATTR_NAME));
+		}
+		if (ctx.hasKey(HTTP_SERVLET_RESPONSE_ATTR_NAME)) {
+			attrs.putIfAbsent(HTTP_SERVLET_RESPONSE_ATTR_NAME, ctx.get(HTTP_SERVLET_RESPONSE_ATTR_NAME));
+		}
+		if (ctx.hasKey(AUTHENTICATION_ATTR_NAME)) {
+			attrs.putIfAbsent(AUTHENTICATION_ATTR_NAME, ctx.get(AUTHENTICATION_ATTR_NAME));
+		}
+		populateDefaultOAuth2AuthorizedClient(attrs);
 	}
 
 	private void populateDefaultRequestResponse(Map<String, Object> attrs) {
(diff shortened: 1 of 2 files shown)


jgrandja:
@rwinch A couple of things for you to consider during the review.

- Do we need to add `Hooks.resetOnLastOperator()`
- I noticed that `RequestContextSubscriber` is instantiated a few times along with the call to `currentContext()`...I'm wondering if this behaviour is expected? Please double check this on your end.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -268,11 +289,41 @@ public void setAccessTokenExpiresSkew(Duration accessTokenExpiresSkew) {
 	public Mono<ClientResponse> filter(ClientRequest request, ExchangeFunction next) {
 		Optional<OAuth2AuthorizedClient> attribute = request.attribute(OAUTH2_AUTHORIZED_CLIENT_ATTR_NAME)
 				.map(OAuth2AuthorizedClient.class::cast);
-		return Mono.justOrEmpty(attribute)
-				.flatMap(authorizedClient -> authorizedClient(request, next, authorizedClient))
-				.map(authorizedClient -> bearer(request, authorizedClient))
-				.flatMap(next::exchange)
-				.switchIfEmpty(next.exchange(request));
+		if (!attribute.isPresent()) {
```
rwinch: This code is the same as the else block. It should be able to be restructured like: try and get the authorized client, if not there default it, process (same process in either case).

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -268,11 +289,41 @@ public void setAccessTokenExpiresSkew(Duration accessTokenExpiresSkew) {
...
+	}
+
+	private Mono<ClientRequest> mergeRequestAttributesFromContext(ClientRequest request) {
+		return Mono.just(ClientRequest.from(request))
+				.flatMap(builder -> Mono.subscriberContext()
+						.map(ctx -> builder.attributes(attrs -> populateRequestAttributes(attrs, ctx))))
+				.map(ClientRequest.Builder::build);
+	}
+
+	private void populateRequestAttributes(Map<String, Object> attrs, Context ctx) {
```
rwinch: Do we want `Context` to override potentially existing `attrs`?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -169,9 +176,23 @@ public void setDefaultClientRegistrationId(String clientRegistrationId) {
 	 * @return the {@link Consumer} to configure the builder
 	 */
 	public Consumer<WebClient.Builder> oauth2Configuration() {
+		Hooks.onLastOperator(REQUEST_CONTEXT_OPERATOR_KEY, Operators.lift((s, sub) -> createRequestContextSubscriber(sub)));
```
rwinch: We can make this implement DisposableBean to clean up the hooks

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
jgrandja: I original had `attrs.putIfAbsent()` but didn't see any reason for it. Do you anticipate a potential issue with overriding? The attributes in the `Context` will always be what the filter needs so I don't see any issue with this. What do you think?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
jgrandja: Not sure I understand? The `ExchangeFilterFunction` would not really be a `@Bean`. The `WebClient` would be but that's not always the case either. What were you thinking exactly?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
jgrandja: Applied in latest commit

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
jgrandja: Applied `putIfAbsent()`

jgrandja:
@rwinch Test has ben added and applied all feedback comments except for this [comment](https://github.com/spring-projects/spring-security/pull/6526#discussion_r256170728). I'll apply that after you reply with more detail.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
rwinch: My concern was if someone had a custom `Authentication` they were working with, it might be explicitly set as an attribute. Perhaps this is so that a global access token is used for client_credentials access tokens.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
rwinch: We need a way to clean up the Hooks and a DisposableBean makes a lot of sense for that reason. Why wouldn't/couldn't it be a Bean? How would you clean it up?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
jgrandja: I see. Ok so `putIfAbsent()` is needed...and it was updated in latest commit

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
jgrandja: I agree `DisposableBean` is an option and yes we need to cleanup/release the Hook. I just don't understand who would be responsible for registering the `ExchangeFilterFunction` as a `@Bean`? Would this be required by the user to register? What if they don't register it? Am I missing something from what you are suggesting?

rwinch: The user is already in charge of creating the `ExchangeFilterFunction` and registering it with the `WebClient`. This just means they should create it as a `Bean`.

If they forget, then it is the same problem as any other `DisposableBean` that they didn't register as a Bean. Things end up in an inconsistent state. This scenario is really more important when the JVM process runs longer than the ApplicationContext (i.e. tests) I doubt it will cause a lot of issues, but it is not semantically correct to leave the Hook registered.

If you don't think DisposableBean is a good idea, what do you think we should do to ensure it gets cleaned up?

jgrandja: > This just means they should create it as a `Bean`.

As a user I would question why do I need to register it as a `@Bean`?

> then it is the same problem as any other `DisposableBean` that they didn't register as a Bean

Yes but if the user is registering a user-defined `@Bean` and they have implemented a cleanup/release of resources than they know why they need to register it as a `@Bean`. With the `ExchangeFilterFunction`, the framework is doing the cleanup so it's not as obvious to them that they need to register it. It will have to be well documented so it's not missed.

> it is not semantically correct to leave the Hook registered.

Agreed, or else the `Context` will contain the request attributes for every sequence created which is not good.

We need to do something for sure. I'm thinking `PostConstruct` instead of `DisposableBean` since the framework recommends this approach as well. What do you think?

jgrandja:
Merged via 0c27f64

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -498,4 +542,55 @@ private UnsupportedOperationException unsupported() {
...
+			this.authentication = authentication;
+		}
+
+		@Override
+		public Context currentContext() {
+			Context context = this.delegate.currentContext();
+			if (context.hasKey(CONTEXT_DEFAULTED_ATTR_NAME)) {
+				return context;
+			}
+			return Context.of(
```
robotmrv: @jgrandja @rwinch   
This subscriber applies for each reactive chain by Hooks.onLastOperator() globally  
Context does not alllow null values (see https://github.com/reactor/reactor-core/pull/1800)  
what if Mono.subscribe()/block() is executed out of security or web context (rabbitmq listener or just some scheduled task...)?

robotmrv: Context javadoc states 
```
 * Note that contexts are optimized for low cardinality key/value storage, and a user
 * might want to associate a dedicated mutable structure to a single key to represent his
 * own context instead of using multiple {@link #put}, which could be more costly.
 * Past five user key/value pair, the {@link Context} will use a copy-on-write
 * implementation backed by a new {@link java.util.Map} on each {@link #put}.
```
https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/util/context/Context.java#L36-L44

but this subscriber creates 4 keys (occupies almost all "optimmized" implementations of Context) 
and as far as I understand `CONTEXT_DEFAULTED_ATTR_NAME` key is just a marker  

maybe it would be better to put all this attrs into some holder so you can reduce number of keys from 4 to 1. And it allows remove extra `CONTEXT_DEFAULTED_ATTR_NAME`

rwinch: @robotmrv Thanks for the report. Can you create a ticket or a a PR for this?

robotmrv: @rwinch 
I've created new issue 
please see https://github.com/spring-projects/spring-security/issues/7228

robotmrv: and here is PR https://github.com/spring-projects/spring-security/pull/7235

PR #6542, opened 2019-02-15
Replace strange hashCode() implementations

bagyoni:
XORing with 31 in these methods seems suspicious as it doesn't affect collision rate, it just flips the last 5 bits. And using bitwise XOR here just feels weird in general.
These are all unlikely to cause collisions but I would replace them with generated ones just in case.
All additions in this commit are generated by Eclipse.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/util/matcher/AntPathRequestMatcher.java
+++ b/web/src/main/java/org/springframework/security/web/util/matcher/AntPathRequestMatcher.java
@@ -220,11 +220,10 @@ public final class AntPathRequestMatcher
 
 	@Override
 	public int hashCode() {
-		int code = 31 ^ this.pattern.hashCode();
-		if (this.httpMethod != null) {
-			code ^= this.httpMethod.hashCode();
-		}
-		return code;
+		int result = this.pattern != null ? this.pattern.hashCode() : 0;
+		result = 31 * result + (this.httpMethod != null ? this.httpMethod.hashCode() : 0);
+		result = 31 * result + (this.caseSensitive ? 1231 : 1237);
+		return result;
 	}
 
 	@Override
--- a/acl/src/main/java/org/springframework/security/acls/domain/ObjectIdentityImpl.java
+++ b/acl/src/main/java/org/springframework/security/acls/domain/ObjectIdentityImpl.java
@@ -152,11 +152,9 @@ public class ObjectIdentityImpl implements ObjectIdentity {
 	 */
 	@Override
 	public int hashCode() {
-		int code = 31;
-		code ^= this.type.hashCode();
-		code ^= this.identifier.hashCode();
-
-		return code;
+		int result = this.type.hashCode();
+		result = 31 * result + this.identifier.hashCode();
+		return result;
 	}
 
 	@Override
--- a/core/src/main/java/org/springframework/security/authentication/jaas/JaasGrantedAuthority.java
+++ b/core/src/main/java/org/springframework/security/authentication/jaas/JaasGrantedAuthority.java
@@ -55,7 +55,9 @@ public final class JaasGrantedAuthority implements GrantedAuthority {
 
 	@Override
 	public int hashCode() {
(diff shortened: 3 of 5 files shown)


pivotal-issuemaster:
@ba-sz-at Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=6542)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=6542) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ba-sz-at Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=6542)!

nealeu:
@bagyoni What led you to investigate these?  Is there a memory footprint or hash collision issue that you've found in use?
I would be reluctant to introduce changes that add cpu cycles without any benefit.

In particular there are null checks that have been added to hashCode() where those values are already known to be non-null.

bagyoni:
@nealeu I was debugging an unrelated issue and happened to notice this. If you are concerned about performance, I would either remove those unnecessary null checks or keep the current hashCodes - though I would still at least remove the 31s, given they don't do anything (and are confusing).

jzheaux:
To the points in this conversation (all good ones!):

* Beginning with 31 does seem a little odd. XORing also has demonstrable inefficiencies as a hash code. You'll notice that newer classes favor the pattern that this PR advocates.
* While these could have a performance impact, it should be negligible. While fixing something that is broken is a stronger argument, I believe the argument for consistency with other code is strong enough in this case.
* Agreed that removing unnecessary null checks would be wise.

And actually, please take a look at newer Spring Security classes like [`AbstractOAuth2Token`](https://github.com/spring-projects/spring-security/blob/master/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/AbstractOAuth2Token.java#L112) to see how to make these proposed implementations more consistent with existing code.

jzheaux:
@bagyoni, can you make the requested changes? Specifically:

* Don't check for null on items known to be non-null
* Follow the pattern found in [`AbstractOAuth2Token`](https://github.com/spring-projects/spring-security/blob/master/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/AbstractOAuth2Token.java#L112)

bagyoni:
@jzheaux Done. Some additional notes:
* Some classes do not have null checks in their equality checks, so e. g. `new JaasGrantedAuthority(null, null).equals(something)` will throw NullPointerException.
* In AntPathRequestMatcher, both the original and the generated hashCodes were inconsistent with equals.

web/src/main/java/org/springframework/security/web/util/matcher/AntPathRequestMatcher.java
```diff
@@ -78,7 +78,7 @@ public AntPathRequestMatcher(String pattern) {
 	/**
 	 * Creates a matcher with the supplied pattern and HTTP method in a case insensitive
 	 * manner.
-	 *
+	 *matcher
```
bagyoni: Inserted by error, removing.

bagyoni:
@nealeu, @jzheaux Gentle reminder that all requested changes have been implemented.

jzheaux:
Thank you for the ping, @bagyoni. And actually, another ticket added [null checks into the constructors](https://github.com/spring-projects/spring-security/issues/6892) that you mentioned. 

Would you mind, then, simplifying the hashcode implementations for `SwitchUserGrantedAuthority`, `JaasGrantedAuthority`, and `RequestKey` accordingly?

bagyoni:
@jzheaux Fair point, I didn't notice that that ticket had been closed. Changed hashcodes accordingly.

jzheaux:
Thanks again, @bagyoni! This is now merged into `master`.

Issue #6546, opened 2019-02-20
Introduce @CurrentSecurityContext for method arguments


jzheaux:
A user can easily obtain the current user principal using the `@AuthenticationPrincipal` annotation, say in a Spring MVC application:

```java
@GetMapping("/path")
public String endpoint(@AuthenticatedPrincipal UserDetails user) {
    // ....
}
```

It would be nice to be able to do something more fundamental, like obtain the current security context:

```java
@GetMapping("/path")
public String endpoint(@CurrentSecurityContext SecurityContext context) {
    // ....
}
```
and thus:
```java
@GetMapping("/path")
public String endpoint(@CurrentSecurityContext(expression="authentication") Authentication authentication) {
    // ....
}
```

This ought to work for both Servlet and WebFlux applications, so we'd need to have two argument resolvers, one that retrieved the security context from `SecurityContextHolder` and another that got it from `ReactiveSecurityContextHolder`. Likely, they could be modeled after `AuthenticationPrincipalArgumentResolver`.

We'd also want to register these respectively in `WebMvcSecurityConfiguration` and `ServerHttpSecurityConfiguration`.

clevertension:
@jzheaux are you working on this issue, if not, i think i can provide a PR :)

jzheaux:
It's yours, @clevertension!

clevertension:
do we need to support the property dig with Authentication?

```java
@GetMapping("/path")
public String endpoint(@CurrentSecurityContext(expression="authentication.details") Object details) {
    // ....
}
```

jzheaux:
@rwinch Is there a compelling need for `errorOnInvalidType`?

jgrandja:
Closed via #6562

PR #6562, opened 2019-02-26
Introduce @CurrentSecurityContext for method arguments

clevertension:
it can both support with spring web mvc and reactive web, user can retrieve `SecurityContext` in controller easily, such as

```java
import org.springframework.security.core.Authentication;
import org.springframework.security.core.annotation.CurrentSecurityContext;
import org.springframework.security.core.context.SecurityContext;
import org.springframework.stereotype.Controller;
import org.springframework.web.bind.annotation.RequestMapping;

@Controller
public class MySecurityController {
	@RequestMapping("/security-context")
	public String securityContext(@CurrentSecurityContext SecurityContext context) {
		Authentication auth = context.getAuthentication();
		//do with security context and related authentication
		return "security-context";
	}
}
```

#6546

### DIFF
--- a/core/src/main/java/org/springframework/security/core/annotation/CurrentSecurityContext.java
+++ b/core/src/main/java/org/springframework/security/core/annotation/CurrentSecurityContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2013 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,7 +26,7 @@ import java.lang.annotation.Target;
  * argument.
  *
  * @author Dan Zheng
- * @since 5.2.x
+ * @since 5.2
  *
  * See: <a href=
  * "{@docRoot}/org/springframework/security/web/bind/support/CurrentSecurityContextArgumentResolver.html"
@@ -43,29 +43,21 @@ public @interface CurrentSecurityContext {
 	 * @return
 	 */
 	boolean errorOnInvalidType() default false;
+
 	/**
 	 * If specified will use the provided SpEL expression to resolve the security context. This
 	 * is convenient if users need to transform the result.
 	 *
+	 * <pre>
+	 * &#64;CurrentSecurityContext(expression = "authentication") Authentication authentication
+	 * </pre>
+	 *
 	 * <p>
-	 * For example, perhaps the user wants to resolve a CustomUser object that is final
-	 * and is leveraging a UserDetailsService. This can be handled by returning an object
-	 * that looks like:
+	 *    if you want to retrieve more object from the authentcation, you can see the following the expression
 	 * </p>
 	 *
 	 * <pre>
-	 * public class CustomUserUserDetails extends User {
-	 *     // ...
-	 *     public CustomUser getCustomUser() {
-	 *         return customUser;
-	 *     }
-	 * }
-	 * </pre>
-	 *
-	 * Then the user can specify an annotation that looks like:
-	 *
-	 * <pre>
-	 * &#64;CurrentSecurityContext(expression = "authentication")
+	 * &#64;CurrentSecurityContext(expression = "authentication.principal") Object principal
 	 * </pre>
 	 *
 	 * @return the expression to use.
--- a/core/src/main/java/org/springframework/security/core/annotation/CurrentSecurityContext.java
+++ b/core/src/main/java/org/springframework/security/core/annotation/CurrentSecurityContext.java
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *      http://www.apache.org/licenses/LICENSE-2.0
+ *      https://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
@@ -28,11 +28,19 @@ import java.lang.annotation.Target;
  * @author Dan Zheng
  * @since 5.2
  *
+ * <p>
  * See: <a href=
  * "{@docRoot}/org/springframework/security/web/bind/support/CurrentSecurityContextArgumentResolver.html"
- * > CurrentSecurityContextArgumentResolver </a>
+ * > CurrentSecurityContextArgumentResolver</a> For Servlet
+ * </p>
+ *
+ * <p>
+ * See: <a href=
+ * "{@docRoot}/org/springframework/security/web/reactive/result/method/annotation/CurrentSecurityContextArgumentResolver.html"
+ * > CurrentSecurityContextArgumentResolver</a> For WebFlux
+ * </p>
  */
-@Target({ ElementType.PARAMETER })
+@Target({ ElementType.PARAMETER, ElementType.ANNOTATION_TYPE })
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
 public @interface CurrentSecurityContext {
--- a/core/src/main/java/org/springframework/security/core/annotation/CurrentSecurityContext.java
+++ b/core/src/main/java/org/springframework/security/core/annotation/CurrentSecurityContext.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2002-2013 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.core.annotation;
+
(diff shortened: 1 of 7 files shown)


web/src/test/java/org/springframework/security/web/bind/support/CurrentSecurityContextArgumentResolverTests.java
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	@Test
+	public void resolveArgumentSecurityContextErrorOnInvalidTypeTrue() throws Exception {
+		String principal = "invalid_type_true";
+		setAuthenticationPrincipal(principal);
+		try {
+			resolver.resolveArgument(showSecurityContextErrorOnInvalidTypeTrue(), null, null, null);
+			fail("should not reach here");
+		} catch(ClassCastException ex) {}
```
jzheaux: Can we use AssertJ's `assertThatCode` here instead of the try-catch?

web/src/main/java/org/springframework/security/web/reactive/result/method/annotation/CurrentSecurityContextArgumentResolver.java
```diff
@@ -0,0 +1,184 @@
...
+
+	/**
+	 * Obtains the specified {@link Annotation} on the specified {@link MethodParameter}.
+	 *
+	 * @param annotationClass the class of the {@link Annotation} to find on the
+	 * {@link MethodParameter}
+	 * @param parameter the {@link MethodParameter} to search for an {@link Annotation}
+	 * @return the {@link Annotation} that was found or null.
+	 */
+	private <T extends Annotation> T findMethodAnnotation(Class<T> annotationClass,
```
jzheaux: @rwinch, it may be a good time to consider whether this ought to be moved elsewhere since it's in four places now. What do you think?

web/src/test/java/org/springframework/security/web/bind/support/CurrentSecurityContextArgumentResolverTests.java
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	private MethodParameter getMethodParameter(String methodName, Class<?>... paramTypes) {
+		Method method = ReflectionUtils.findMethod(TestController.class, methodName,
+				paramTypes);
+		return new MethodParameter(method, 0);
+	}
+
+	public static class TestController {
+		public void showSecurityContextNoAnnotation(String user) {
```
jzheaux: Would be nice if one of the methods used a custom implementation for `SecurityContext` so it proves the type matching works.

core/src/main/java/org/springframework/security/core/annotation/CurrentSecurityContext.java
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright 2002-2013 the original author or authors.
```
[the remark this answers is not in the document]
clevertension: ok

core/src/main/java/org/springframework/security/core/annotation/CurrentSecurityContext.java
```diff
@@ -0,0 +1,74 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation that is used to resolve {@link SecurityContext#getAuthentication()} to a method
+ * argument.
+ *
+ * @author Dan Zheng
+ * @since 5.2.x
```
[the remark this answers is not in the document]
clevertension: ok

core/src/main/java/org/springframework/security/core/annotation/CurrentSecurityContext.java
```diff
@@ -0,0 +1,74 @@
...
+@Documented
+public @interface CurrentSecurityContext {
+	/**
+	 * True if a {@link ClassCastException} should be thrown when the current
+	 * {@link SecurityContext} is the incorrect type. Default is false.
+	 *
+	 * @return
+	 */
+	boolean errorOnInvalidType() default false;
+	/**
```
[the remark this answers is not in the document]
clevertension: ok

web/src/test/java/org/springframework/security/web/bind/support/CurrentSecurityContextArgumentResolverTests.java
clevertension: i think it is better to use 

```java
assertThatExceptionOfType
```
just like
```java
assertThatExceptionOfType(ClassCastException.class).isThrownBy(() -> resolver.resolveArgument(showSecurityContextErrorOnInvalidTypeTrue(), null,
				null, null));
```

web/src/main/java/org/springframework/security/web/reactive/result/method/annotation/CurrentSecurityContextArgumentResolver.java
clevertension: yes, i think it is good to move to `AnnotationUtils`, :blush:

web/src/test/java/org/springframework/security/web/bind/support/CurrentSecurityContextArgumentResolverTests.java
clevertension: ok, i will add more cases

clevertension:
@jzheaux  thanks very much for your review feedback, i have already fix all the issues that you mentioned, so should i add extra 1 commit, or still keep only one commit by force push here?

jzheaux:
@clevertension Either way. Before we merge, you'll need to force push down to one commit. Some folks like to keep commits separate while the review is going on. For small PRs like this, it doesn't make a big difference.

clevertension:
:thumbsup:,:thumbsup:,:thumbsup:

clevertension:
@rwinch , can you spare time to review this PR? thank you very much

web/src/main/java/org/springframework/security/web/bind/support/CurrentSecurityContextArgumentResolver.java
```diff
@@ -0,0 +1,172 @@
...
+			}
+		}
+		return securityContextResult;
+	}
+	/**
+	 * Sets the {@link BeanResolver} to be used on the expressions
+	 * @param beanResolver the {@link BeanResolver} to use
+	 */
+	public void setBeanResolver(BeanResolver beanResolver) {
+		this.beanResolver = beanResolver;
```
jzheaux: Please check for `null` here using something like:

```java
Assert.notNull(beanResolver, "beanResolver cannot be null");
```

web/src/main/java/org/springframework/security/web/reactive/result/method/annotation/CurrentSecurityContextArgumentResolver.java
```diff
@@ -0,0 +1,184 @@
...
+	public CurrentSecurityContextArgumentResolver(ReactiveAdapterRegistry adapterRegistry) {
+		super(adapterRegistry);
+	}
+
+	/**
+	 * Sets the {@link BeanResolver} to be used on the expressions
+	 * @param beanResolver the {@link BeanResolver} to use
+	 */
+	public void setBeanResolver(BeanResolver beanResolver) {
+		this.beanResolver = beanResolver;
```
jzheaux: Please check for `null` here as well.

web/src/test/java/org/springframework/security/web/bind/support/CurrentSecurityContextArgumentResolverTests.java
```diff
@@ -0,0 +1,287 @@
...
+		return getMethodParameter("showSecurityContextErrorOnInvalidTypeTrue", String.class);
+	}
+
+	private MethodParameter getMethodParameter(String methodName, Class<?>... paramTypes) {
+		Method method = ReflectionUtils.findMethod(TestController.class, methodName,
+				paramTypes);
+		return new MethodParameter(method, 0);
+	}
+
+	public static class TestController {
```
jzheaux: I think a test for meta-annotations is missing. Since this is a more complex class, it's better to make sure we are well covered.

web/src/main/java/org/springframework/security/web/bind/support/CurrentSecurityContextArgumentResolver.java
clevertension: fixed

web/src/main/java/org/springframework/security/web/reactive/result/method/annotation/CurrentSecurityContextArgumentResolver.java
clevertension: fixed

web/src/test/java/org/springframework/security/web/bind/support/CurrentSecurityContextArgumentResolverTests.java
clevertension: can you give me some examples? or `The Existing Annotation for reference`,  thanks

jzheaux: For example, if you do:

```java
@Retention(RUNTIME)
@CurrentSecurityContext(expression="authentication")
public @interface CurrentAuthentication
```

And then

```java
public String myMethod(@CurrentAuthentication Authentication auth) {
}
```

Then, the resolution should still work.

clevertension: fixed

jzheaux:
Thanks, @clevertension! This is now merged into `master`.

Issue #5385, opened 2018-05-25
Add resource server support for multiple trusted JWT access token issuers


dyroberts:
### Summary

Add resource server support for multiple trusted JWT access token issuers

### Actual Behavior

Presently we can configure a resource server's trusted jwt token issuer like:

`security.oauth2.resource.jwt.keyUri=...`
or
`security.oauth2.resource.jwk.keySetUri=...`

### Expected Behavior

I'd like to suggest supporting something like (switching to yml):

```
security:
  oauth2:
    resource:
      issuers:
        issuer1:
          keyUri: ...
          issuerClaim: ...
        issuer2:
          keySetUri: ...
        issuer3:
          keyValue: ...
```
Note the optional issuerClaim for verification against the 'iss' claim, optionally.

jgrandja:
The key question to ask the UAA team is the `kid` unique across identity zones?

jgrandja:
@jzheaux Has this been solved via #5351?

jzheaux:
@jgrandja good question - it's certainly possible via #5351 but it would not be very efficient. The resulting `AuthenticationManagerResolver` would need to parse the JWT and select the appropriate `JwtAuthenticationManager` instance accordingly, at which point it would be parsed again by the underlying `JwtDecoder`.

It seems to me that a `JwtDecoder` implementation may be better suited for this use case. 

The `AuthenticationManagerResolver` is suitable for aspects of the request that are immediately obtainable, like a path, a header, or a subdomain.

jzheaux:
Closed via de87675

cadethacker:
Sorry to comment on an older issue, but one point of confusion.  The OP asked about defining multiple authorization services via the yaml/config.  I spent a good bit of time coding up support [by using the docs](https://docs.spring.io/spring-security/site/docs/current/reference/html5/#oauth2resourceserver-multitenancy) (very good BTW) and it works, but...

I still didn't see a clear answer in this issue on if the yaml/config support multiple authorization server by issuer yet as per the OP question. I know I can code it, but can I config it :D

jzheaux:
@cadethacker you are correct, YAML config is not supported for multiple issuers.

cadethacker:
huge thank you @jzheaux for the clarification. Is there a correct way to request this as a feature?  I think this would be a great addition to Spring Security especially in the awesome work for the entire 5 series.   I'm happy to hack on it, but no clue where to start.  If you can give me a compass heading and a shove in the right direction.

[...]

PR #6563, opened 2019-02-26
Multi tenancy for Resource Server

jzheaux:

### DIFF
--- a/samples/boot/oauth2resourceserver-multitenancy/src/main/java/sample/OAuth2ResourceServerSecurityConfiguration.java
+++ b/samples/boot/oauth2resourceserver-multitenancy/src/main/java/sample/OAuth2ResourceServerSecurityConfiguration.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package sample;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import javax.servlet.http.HttpServletRequest;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Bean;
+import org.springframework.security.authentication.AuthenticationManager;
+import org.springframework.security.authentication.AuthenticationManagerResolver;
+import org.springframework.security.config.annotation.web.builders.HttpSecurity;
+import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
+import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
+import org.springframework.security.oauth2.jwt.JwtDecoder;
+import org.springframework.security.oauth2.jwt.JwtProcessors;
+import org.springframework.security.oauth2.jwt.NimbusJwtDecoder;
+import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationProvider;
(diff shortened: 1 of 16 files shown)


jzheaux:
@rwinch I'm using only `AuthenticationManager` to support multi-tenancy. The reason is because it's possible for `Authentication` to contain all the necessary information to resolve tenant differences.

Originally, I tried out an `AuthenticationManagerResolver.resolve(HttpServletRequest)`, though I'm of the mind that there's value in this interface not being tied to the request. In my experimentation with alternatives, it was simpler (no new interfaces) to simply allow the user to supply a custom `AuthenticationManager` and a custom `AuthenticationDetailsSource`.

This has a nice cognate on the reactive side, injecting a custom `ServerAuthenticationConverter` to get an `Authentication`, and then send that `Authentication` to a custom `ReactiveAuthenticationManager`. 

I'm open to discussion here, of course.

jgrandja:
Related #5385

samples/boot/oauth2resourceserver-multitenancy/src/main/java/sample/OAuth2ResourceServerSecurityConfiguration.java
```diff
@@ -0,0 +1,83 @@
...
+		JwtDecoder jwtDecoder = new NimbusJwtDecoder(JwtProcessors.withJwkSetUri(this.jwkSetUri).build());
+		return new JwtAuthenticationProvider(jwtDecoder);
+	}
+
+	OAuth2IntrospectionAuthenticationProvider opaque() {
+		return new OAuth2IntrospectionAuthenticationProvider(this.introspectionUri, "client", "secret");
+	}
+
+	AuthenticationDetailsSource<HttpServletRequest, ?> authenticationDetailsSource() {
+		return request -> request.getPathInfo().split("/")[1];
```
jgrandja: @jzheaux I don't think it makes sense to use `AuthenticationDetailsSource` for extracting the tenant info. In this example, `AuthenticationManager` indirectly relies on `HttpServletRequest` which is not the intended design for `AuthenticationManager` as it's not meant to be tied to the web-tier - it should work on a desktop application for example.

Also, the tenant info may come from a request parameter so I'm not sure how this would work with the current setup.

Either way, I feel we need to extract the tenant info in the web-tier and populate it with a new impl of  `Authentication` that is passed into `AuthenticationManager`. We may even consider a new type that is populated in a `AuthenticationDetailsSource`? I just think we should leave the extracting of the tenant info in the web-tier - `Filter` or something similar.

jzheaux: > In this example, `AuthenticationManager` indirectly relies on `HttpServletRequest`

What do you mean that it is indirectly reliant? By that same logic, isn't it also indirectly reliant on `HttpServletRequest` because of the bearer token resolver? No web-ness makes it down into the authentication manager. We just have a tenantId that so happens to be the path.

This also happens on the reactive side. The `AuthenticationWebFilter` calls a `ServerAuthenticationConverter`, which takes a `request` and returns an `Authentication`, which is then passed to a manager.

> We may even consider a new type that is populated in a `AuthenticationDetailsSource

A stronger type may be in order, as you said, though. Calling `setDetails(Object)` doesn't offer any type safety, nor any direction on what ought to go there. However, I don't think that I'd want to enforce it at the setter level. For example, I wouldn't want to do this:

```java
public void setAuthenticationDetailsSource
    (AuthenticationDetailsSource<HttpServletRequest, Tenant>)
```

Since there are certainly other scenarios where the user would like to provide details in the `Authentication` object.

> I just think we should leave the extracting of the tenant info in the web-tier - `Filter` or something similar.

Just to be clear, this is already technically the case since the filter calls `source.buildDetails` before invoking the authentication manager. But, I'm thinking you are calling out something more nuanced than that.

jgrandja: @jzheaux Regarding my comment

> `AuthenticationManager` indirectly relies on `HttpServletRequest` which is not the intended design...

Please ignore as I misinterpreted the implementation on my initial look. All seems fine. Actually, I like the fact that you're leveraging `Authentication.details` to store the tenant info. This allows for an easy integration.

I'm wondering if it makes sense to leverage (or extend) `WebAuthenticationDetailsSource` and store the tenant info in `WebAuthenticationDetails` or subclass of it?

PR #6599, opened 2019-03-08
wrap ServerWebExchange to support `.getPrincipal()`

### LINKED ISSUE #6598: In Test @AuthenticationPrincipal is null because ServerWebExchange is not wrapped
### Summary

Using `@WithMockUser` or `org.springframework.security.test.web.reactive.server.SecurityMockServerConfigurers#mockAuthentication` directly in combination with `@AuthenticationPrincipal` in a WebFlux controller leaves the principal empty while things like `@PreAuthorize('isAuthenticated()')` work as expected.

### Actual Behavior

Mocking the principal in a test does not work with `@AuthenticaitonPrincipal`. 

Because my SecurityFilterChain does not match the request of the web test client (see below, only matches Requests that contain an Authorization: Basic header), the `org.springframework.security.web.server.context.SecurityContextServerWebExchangeWebFilter` gets never called (which is part of the SecurityFilterChain), meaning the request is never wrapped with a `org.springframework.security.web.server.context.SecurityContextServerWebExchange`.

`org.springframework.security.web.reactive.result.method.annotation.AuthenticationPrincipalArgumentResolver#resolveArgument` uses `exchange.getPrincipal()` to resolve the principal, which returns an empty Mono because the exchange is still the `DefaultServerWebExchange` (not security!). 

Now the principal set in `org.springframework.security.test.web.reactive.server.SecurityMockServerConfigurers.MutatorFilter` is not considered.

The `org.springframework.security.access.prepost.PrePostAdviceReactiveMethodInterceptor` uses `org.springframework.security.core.context.ReactiveSecurityContextHolder#getContext` which would work for the resolver as well.

### Expected Behavior

The mocked principal should be resolved correctly for the `@AuthenticationPrincipal` argument.

The exchange should be wrapped correctly.

### Configuration

WebTestClient (minimal)
```java
        this.webTestClient = WebTestClient.bindToApplicationContext(context)
            .apply(springSecurity())
            .configureClient()
            .mutateWith(mockUser(someUser))
```

Security (minimal):

```java
    @Configuration
    public static class BasicAuthSecurity {
        @Bean
        public SecurityWebFilterChain basicAuthSecurityWebFilterChain(ServerHttpSecurity http) {
            return http
                .securityMatcher(new BasicAuthWebExchangeMatcher())
                .authorizeExchange().anyExchange().permitAll().and()
                .headers().frameOptions().disable().and()
                .csrf().disable()
                .logout().disable()
                .httpBasic().securityContextRepository(NoOpServerSecurityContextRepository.getInstance()).and()
                .exceptionHandling()
                    .accessDeniedHandler(new HttpStatusServerAccessDeniedHandler(HttpStatus.UNAUTHORIZED))
                    .authenticationEntryPoint(new HttpStatusServerEntryPoint(HttpStatus.UNAUTHORIZED)).and()
                .build();
        }
    }
```

```java
public class BasicAuthWebExchangeMatcher implements ServerWebExchangeMatcher {
    @Override
    public Mono<MatchResult> matches(ServerWebExchange exchange) {
        var header = exchange.getRequest().getHeaders().getFirst("Authorization");
        return StringUtils.startsWithIgnoreCase(header, "Basic")
            ? MatchResult.match() : MatchResult.notMatch();
    }
}
```

The important part here is the security matcher, which makes it so my Security chain does *not* match, nor does any other!

### Version

Spring Security 5.1.4

### Sample

Company code.

Dav1dde:
Fixes #6598

Properly wraps the `ServerWebExchange` so that in a mocked environment without active `SecurityFilterChain` the principal can still be received.

### DIFF
--- a/test/src/main/java/org/springframework/security/test/web/reactive/server/SecurityMockServerConfigurers.java
+++ b/test/src/main/java/org/springframework/security/test/web/reactive/server/SecurityMockServerConfigurers.java
@@ -26,6 +26,7 @@ import org.springframework.security.core.context.SecurityContext;
 import org.springframework.security.core.context.SecurityContextImpl;
 import org.springframework.security.core.userdetails.User;
 import org.springframework.security.core.userdetails.UserDetails;
+import org.springframework.security.web.server.context.SecurityContextServerWebExchange;
 import org.springframework.security.web.server.csrf.CsrfWebFilter;
 import org.springframework.security.web.server.util.matcher.ServerWebExchangeMatcher;
 import org.springframework.test.web.reactive.server.MockServerConfigurer;
@@ -285,11 +286,12 @@ public class SecurityMockServerConfigurers {
 
 		@Override
 		public Mono<Void> filter(ServerWebExchange exchange, WebFilterChain webFilterChain) {
-			Supplier<Mono<SecurityContext>> context = exchange.getAttribute(ATTRIBUTE_NAME);
-			if (context != null) {
+			Supplier<Mono<SecurityContext>> contextSupplier = exchange.getAttribute(ATTRIBUTE_NAME);
+			if (contextSupplier != null) {
+				Mono<SecurityContext> context = contextSupplier.get();
 				exchange.getAttributes().remove(ATTRIBUTE_NAME);
-				return webFilterChain.filter(exchange)
-					.subscriberContext(ReactiveSecurityContextHolder.withSecurityContext(context.get()));
+				return webFilterChain.filter(new SecurityContextServerWebExchange(exchange, context))
+					.subscriberContext(ReactiveSecurityContextHolder.withSecurityContext(context));
 			}
 			return webFilterChain.filter(exchange);
 		}


eleftherias:
@Dav1dde please see my comment on the GitHub issue gh-6598.  
We do not want to wrap the context in a `SecurityContextServerWebExchange`, since that does not mimic what is happening when the application is running.  
Instead, I suggest creating an `HttpBasicMutator`, which will include the HTTP basic `Authorization` header in the request.

eleftherias:
@Dav1dde Can you confirm that this issue was fixed in gh-7687 and if so please close this PR?

Dav1dde:
@eleftherias my current issue is resolved with that, but now it's more of a question of how the spring security context interacts with `exchange.getPrincipal()`. But from my understanding and as we discussed in gh-6598, you (the spring-security team) think the request has to go through a security-filter  in order to "get" a principal,  which is fair and makes sense - this makes the PR obselete.

PR #6604, opened 2019-03-10
Throw exception that was created but not thrown

### LINKED ISSUE #5462: Created exception is not throwing
### Summary
Exception object was created but not throwing (and not used later).

### Actual Behavior
It some sort of *programmer's typo*: create exception object and forget it

### Expected Behavior
An exception must propagate to caller in order DigestAuthenticationFilter to work properly

### Configuration
No configuration required -- it's obvious code-level mistake.

### Version
Found in latest 5.1.0.M1. Still exists in earlier versions and in master for the moment issue was created.

### Sample
[link to line in class](https://github.com/spring-projects/spring-security/blob/a1c93b1cce804072f34bc54f7c9aef5b736a1793/web/src/main/java/org/springframework/security/web/authentication/www/DigestAuthenticationFilter.java#L433)

ThomasVitale:
Fixes #5462 

In the `DigestAuthenticationFilter`, a `BadCredentialsException` was created when the nonce signature is invalid, but never thrown.

This case is already covered by a test (`testNonceWithIncorrectSignatureForNumericFieldReturnsForbidden` in the `DigestAuthenticationFilterTests` class), but it was green since an exception was thrown anyway at a later point of the filter, resulting in a 401 status (line 217).
https://github.com/spring-projects/spring-security/blob/ad9dc49d5559ff0b093ee425da57b77d46d69e07/web/src/main/java/org/springframework/security/web/authentication/www/DigestAuthenticationFilter.java#L217)

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/www/DigestAuthenticationFilter.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/www/DigestAuthenticationFilter.java
@@ -431,7 +431,7 @@ public class DigestAuthenticationFilter extends GenericFilterBean
 					.md5Hex(this.nonceExpiryTime + ":" + entryPointKey);
 
 			if (!expectedNonceSignature.equals(nonceTokens[1])) {
-				new BadCredentialsException(DigestAuthenticationFilter.this.messages
+				throw new BadCredentialsException(DigestAuthenticationFilter.this.messages
 						.getMessage("DigestAuthenticationFilter.nonceCompromised",
 								new Object[] { nonceAsPlainText },
 								"Nonce token compromised {0}"));


jzheaux:
@ThomasVitale good catch! I see your comment about an existing test. 

Is there a test that you could add which would correctly fail before your change, but would pass afterward? It'd be nice to add that as part of this PR.

ThomasVitale:
I thought about that, but I couldn't find a reasonable way to do it. The filter implementation can throw a `BadCredentialsException` for different reasons, but from the outside we are not aware of that since the filter catches the exception internally and returns a 401 response. And that is what is tested in `DigestAuthenticationFilterTests`. Do you have any suggestion? I would really much like to improve this PR with a test.

jzheaux:
Yeah, this is a bit of a tricky one. And digest auth is pretty complex on its own.

Thinking out loud here a bit, what we'd need is for this check:

```java
String expectedNonceSignature = 
		DigestAuthUtils.md5Hex(this.nonceExpiryTime + ":" + entryPointKey);

if (!expectedNonceSignature.equals(nonceTokens[1])) {
```

to fail, but not any other check.

So, if the test's nonce were generated with something other than the configured `entryPointKey`, then I think that check would fail while the others would pass. Your test, for example, could call `generateNonce`, but not use `KEY`:

```java
@Test
public void test() throws Exception {
	String badNonce = generateNonce(60, "badkey");
	String responseDigest = DigestAuthUtils.generateDigest(false, USERNAME, REALM,
			PASSWORD, "GET", REQUEST_URI, QOP, badNonce, NC, CNONCE);

	request.addHeader("Authorization", createAuthorizationHeader(USERNAME, REALM,
			badNonce, REQUEST_URI, responseDigest, QOP, NC, CNONCE));

	// ...
}
```

jzheaux:
Thanks again, @ThomasVitale, for the contribution! This is now merged into `master` via d88c2c19f0ada887fda8d3365271efc6089cf411. I also added a unit test via 20a7bc47854e89987f7419d93437d798e6fe89d0.

Issue #6548, opened 2019-02-21
Consider enabling PKCE for confidential clients


jgrandja:
The goal of [PKCE](https://tools.ietf.org/html/rfc7636) is to provide an added level of security for **OAuth 2.0 public clients** (utilizing the Authorization Code Grant) from an authorization code interception attack.

However, based on OAuth 2.0 Security Best Current Practice, in [section 2.1.1.  Authorization Code Grant](https://tools.ietf.org/html/draft-ietf-oauth-security-topics-11#section-2.1.1):

> Note: although PKCE so far was recommended as a mechanism **to protect
   native apps**, this advice applies to all kinds of OAuth clients,
   **including web applications**.

It can also be leveraged for **confidential clients** for an added layer of security.

Given this, we should consider adding this support. From initial research, Okta does support this scenario where a confidential client is used to authenticate with the Token Endpoint and the `code_verifier` is sent as a parameter.

We should conduct further research to see which other providers support this client configuration/registration.

See this [comment](https://github.com/spring-projects/spring-security/pull/6485#discussion_r258058756) for further details.

Related #6446

dfcoffin:
IETF RFC 8252 "OAuth 2.0 for Native Apps" modifies RFC 6749 "The OAuth 2.0 Authorization Framework" and states in Section 

[Appendix A.  Server Support Checklist

   OAuth servers that support native apps must:

   1.  Support private-use URI scheme redirect URIs.  This is required
       to support mobile operating systems.  See Section 7.1.

   2.  Support "https" scheme redirect URIs for use with public native
       app clients.  This is used by apps on advanced mobile operating
       systems that allow app-claimed "https" scheme URIs.  See
       Section 7.2.

   3.  Support loopback IP redirect URIs.  This is required to support
       desktop operating systems.  See Section 7.3.

   4.  Not assume that native app clients can keep a secret.  If secrets
       are distributed to multiple installs of the same native app, they
       should not be treated as confidential.  See Section 8.5.

   5.  Support PKCE [RFC7636].  Required to protect authorization code
       grants sent to public clients over inter-app communication
       channels.  See Section 8.1](https://tools.ietf.org/html/rfc8252#section-4.1)

jgrandja:
@sdoxsee Thanks for the offer on taking on this ticket. It would be appreciated. But before we do any work here...

> We should conduct further research to see which other providers support this client configuration/registration.

[...]

PR #6608, opened 2019-03-13
Enable to expand authorization URI from base url.

wangzw:
We use spring security with spring-cloud-gateway and we expect the oauth2 authorization URI which will be redirected to the user can be expand from incoming request's base url. So that we can determine the host and protocol dynamically and add gateway's context path to authorization URI. This patch add such feature and keep the backward compatibility.

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolver.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolver.java
@@ -98,6 +98,22 @@ public final class DefaultOAuth2AuthorizationRequestResolver implements OAuth2Au
 		return action;
 	}
 
+	private String expandAuthorizationUri(HttpServletRequest request, String authorizationUri) {
+		// Supported URI variables -> baseUrl
+		Map<String, String> uriVariables = new HashMap<>();
+
+		String baseUrl = UriComponentsBuilder.fromHttpUrl(UrlUtils.buildFullRequestUrl(request))
+				.replaceQuery(null)
+				.replacePath(request.getContextPath())
+				.build()
+				.toUriString();
+		uriVariables.put("baseUrl", baseUrl);
+
+		return UriComponentsBuilder.fromUriString(authorizationUri)
+				.buildAndExpand(uriVariables)
+				.toUriString();
+	}
+
 	private OAuth2AuthorizationRequest resolve(HttpServletRequest request, String registrationId, String redirectUriAction) {
 		if (registrationId == null) {
 			return null;
@@ -131,7 +147,8 @@ public final class DefaultOAuth2AuthorizationRequestResolver implements OAuth2Au
 
 		OAuth2AuthorizationRequest authorizationRequest = builder
 				.clientId(clientRegistration.getClientId())
-				.authorizationUri(clientRegistration.getProviderDetails().getAuthorizationUri())
+				.authorizationUri(expandAuthorizationUri(request,
+						clientRegistration.getProviderDetails().getAuthorizationUri()))
 				.redirectUri(redirectUriStr)
 				.scopes(clientRegistration.getScopes())
 				.state(this.stateGenerator.generateKey())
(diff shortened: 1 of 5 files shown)


wangzw:
Hi @jgrandja , thanks for your reply. It is exactly what we do currently.  But 1) it is not an elegant solution for this requirement. In this PR it only use few lines to fix. But for a custom `OAuth2AuthorizationRequestResolver` it cost a wrapper class and it has to construct a new `OAuth2AuthorizationRequest` based on the return value of `DefaultOAuth2AuthorizationRequestResolver` since there is no way to modify `OAuth2AuthorizationRequest`. 2) It may be a general requirement to expand authorization URI from base url, we already did this for `redirectUri` isn't it? And this patch can save others time.

jgrandja:
@wangzw The delegating `OAuth2AuthorizationRequestResolver` approach is quite powerful as it allows you to completely modify the `OAuth2AuthorizationRequest` as you see fit. This was the motivation of bringing this feature in. I'm happy to see that you are using it and it's working for you.

As an FYI, the expanding of the `redirect-uri` is different than expanding of the `authorization-uri`. We needed to expand/resolve the  `redirect-uri` based on certain functionality. I haven't seen any need from the community to expand the `authorization-uri` until now. The reason I'm reluctant to merge this patch is we can also apply the same reasoning to `tokenUri`, `userInfoEndpoint.uri`,  `jwkSetUri`, etc. All of this can be achieved via a custom `OAuth2AuthorizationRequestResolver` for the specific use case in hand - as you have solved via your use case and Gateway.

Please take a look at this [comment](https://github.com/spring-projects/spring-security/issues/6548#issuecomment-475761846) as we're looking at adding a new feature in `DefaultOAuth2AuthorizationRequestResolver` that may help your use case.

wangzw:
The requests `redirect-uri` and `authorization-uri` are issued from client browser and not like the requests `tokenUri` and `userInfoEndpoint.uri` which are issued from `spring-security`(api gateway in my case). That is why `redirect-uri` and `authorization-uri` should expand from base url but`tokenUri` and `userInfoEndpoint.uri` not.

So `redirect-uri` and `authorization-uri` should be expand in similar way. Currently only `redirect-uri` get expanded.

I did not say the delegating OAuth2AuthorizationRequestResolver approach is not powerful, what I mean is that `DefaultOAuth2AuthorizationRequestResolver` is not feature complete.

I think it is reasonable to add missing feature into `DefaultOAuth2AuthorizationRequestResolver` instead of a delegated object.

wangzw:
I can add more background to my use case. Our system is microservices architecture and api gateway (spring-cloud-gateway) is the boundary of internal services and external users.

For internal service such as authentication/authorization can be reached by internal unique dns name or service name.  This is why `tokenUri` and `userInfoEndpoint.uri` can be hardcoded, it need not to expose to external users (only accessed by gateway).

`redirect-uri` and `authorization-uri` should be reachable from external users (proxied by gateway) as defined in oauth2 protocol. They were hardcoded before and worked well since our service has only one domain name. Until we have to add more domain name for different customer and partners, we have to expand `redirect-uri` and `authorization-uri` dynamically.

jgrandja:
@wangzw

> The requests `redirect-uri` and `authorization-uri` are issued from client browser and not like the requests `tokenUri` and `userInfoEndpoint.uri` which are issued from `spring-security`(api gateway in my case). That is why `redirect-uri` and `authorization-uri` should expand from base url but`tokenUri` and `userInfoEndpoint.uri` not.

I don't agree that `authorization-uri` should be expandable and I would not compare it to the `redirect-uri` either. Please review the spec in section [3.  Protocol Endpoints](https://tools.ietf.org/html/rfc6749#section-3) - the Authorization Endpoint and Redirection Endpoint. Based on this [specific change](https://github.com/spring-projects/spring-security/pull/6608/files#diff-41fd5e9667bf9988c15c550bebecfb78R110) in the PR, the assumption is that the `authorization-uri` lives in the same application space as the client application. **This is not recommended**. The Authorization Server should be a completely separate deployment from the client application. I feel like you might need to re-visit your architecture and/or deployment model and adjust as necessary.

We're not going to be able to accept this PR because the changes here imply that the Authorization Endpoint lives in the sample application space as the client application, which we are not recommending as a best practice for a deployment topology.

Issue #6632, opened 2019-03-20
Opaque Token Intermediate Type


jzheaux:
One clear benefit of `JwtDecoder` is it acts as an intermediate type for the purposes of configuration. It makes auto-configuration with Spring Boot very clean, and presents opportunities where the user can easily customize Bearer Token authentication strategy without needing to wire up a custom authentication provider.

It would be nice if the opaque token support had the same flexibility.

Issue #6544, opened 2019-02-15
Opaque Token Support with Dynamic Credentials


rwinch:
### Summary

I think it may make sense to provide first class support (or since we haven't hit GA change our current support) for opaque token support with dynamic credentials. Currently, there is no first class API for supporting rotating the client credentials.

I wonder if it makes sense to use the ClientRegistrationRepository which could be used for looking up both the client key/secret but also the endpoint that is validated against. This would also make it simple for configuration via the issuer uri.

jzheaux:
Would propose that this introduce an optional dependency on `oauth2-client` - it would at least be optional based on configuring for opaque tokens.

I'm imagining that this would replace the existing constructors - is there value in keeping the (clientId, clientSecret) constructor, though?

jzheaux:
It's not yet clear the means by which dynamic credentials ought to be provided.

An obvious first candidate is using `ClientRegistration` and `ClientRegistrationRepository` since these hold client ids and secrets already, and these are in `oauth2-client`. `ClientRegistration` turns out to not be such a great fit, though, since it is focused on OAuth 2.0 grant flows. A grant flow is just one of numerous ways a client can identify itself to an introspection endpoint, and most times it wouldn't make sense to specify a grant type for client authentication.

There is an additional problem with selecting `ClientRegistrationRepository` and that is that, when a `ClientRegistrationRepository` is exposed by the application, Spring Boot will wire the application with filters for negotiating the authorization code grant flow. Exposing a `ClientRegistrationRepository` means that the application is identifying itself as an OAuth 2.0 Client Application, so this again is not what a resource server is when it's making introspection requests.

Another candidate is a new contract. For example, we could introduce a contract called  `ClientCredentials`. This immediately begs the question, though, about the numerous other modes for client authentication and how will those be represented. The challenge with this is that, being so early-on with introspection, it isn't clear how that contract will evolve.

It's also not yet clear where something like this would live since it should not go into `oauth2-client` insofar as that dependency is focused on identifying that application as an OAuth 2.0 Client Application. There is, of course, a natural hesitancy to place it in `oauth2-core` as it may not be genuinely cross-cutting.

The user can already add dynamic support themselves, too, since `OAuth2TokenIntrospectionClient` accepts a `RestOperations` as a constructor parameter. Quite simply, an application can do:

```java
@Bean 
public OAuth2TokenIntrospectionClient introspectionClient() {
    RestTemplate rest = new RestTemplate();
[...]

PR #6633, opened 2019-03-20
Opaque Token Intermediate Type

jzheaux:
Introducing OAuth2TokenIntrospectionClient and also
ReactiveOAuth2TokenIntrospectionClient as configuration points.

The DSL looks in the application context for these types in the same
way it looks for JwtDecoder and ReactiveJwtDecoder, and exposes
similar configuration methods.

Fixes: gh-6632

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
@@ -16,6 +16,7 @@
 
 package org.springframework.security.config.annotation.web.configurers.oauth2.server.resource;
 
+import java.util.function.Supplier;
 import javax.servlet.http.HttpServletRequest;
 
 import org.springframework.context.ApplicationContext;
@@ -23,7 +24,6 @@ import org.springframework.core.convert.converter.Converter;
 import org.springframework.security.authentication.AbstractAuthenticationToken;
 import org.springframework.security.authentication.AuthenticationManager;
 import org.springframework.security.authentication.AuthenticationManagerResolver;
-import org.springframework.security.authentication.AuthenticationProvider;
 import org.springframework.security.config.annotation.web.HttpSecurityBuilder;
 import org.springframework.security.config.annotation.web.configurers.AbstractHttpConfigurer;
 import org.springframework.security.config.annotation.web.configurers.CsrfConfigurer;
@@ -36,6 +36,8 @@ import org.springframework.security.oauth2.jwt.NimbusJwtDecoder;
 import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationConverter;
 import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationProvider;
 import org.springframework.security.oauth2.server.resource.authentication.OAuth2IntrospectionAuthenticationProvider;
+import org.springframework.security.oauth2.server.resource.introspection.NimbusOAuth2TokenIntrospectionClient;
+import org.springframework.security.oauth2.server.resource.introspection.OAuth2TokenIntrospectionClient;
 import org.springframework.security.oauth2.server.resource.web.BearerTokenAuthenticationEntryPoint;
 import org.springframework.security.oauth2.server.resource.web.BearerTokenAuthenticationFilter;
 import org.springframework.security.oauth2.server.resource.web.BearerTokenResolver;
@@ -179,7 +181,7 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 
 	public OpaqueTokenConfigurer opaqueToken() {
 		if (this.opaqueTokenConfigurer == null) {
-			this.opaqueTokenConfigurer = new OpaqueTokenConfigurer();
+			this.opaqueTokenConfigurer = new OpaqueTokenConfigurer(this.context);
 		}
 
 		return this.opaqueTokenConfigurer;
@@ -237,7 +239,10 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 		}
 
 		if (this.opaqueTokenConfigurer != null) {
-			http.authenticationProvider(this.opaqueTokenConfigurer.getProvider());
+			OAuth2TokenIntrospectionClient introspectionClient = this.opaqueTokenConfigurer.getIntrospectionClient();
+			OAuth2IntrospectionAuthenticationProvider provider =
+					new OAuth2IntrospectionAuthenticationProvider(introspectionClient);
+			http.authenticationProvider(provider);
 		}
 	}
 
@@ -288,27 +293,46 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 	}
 
 	public class OpaqueTokenConfigurer {
+		private final ApplicationContext context;
+
 		private String introspectionUri;
-		private String introspectionClientId;
-		private String introspectionClientSecret;
+		private String clientId;
+		private String clientSecret;
+		private Supplier<OAuth2TokenIntrospectionClient> introspectionClient;
+
+		OpaqueTokenConfigurer(ApplicationContext context) {
+			this.context = context;
+		}
 
 		public OpaqueTokenConfigurer introspectionUri(String introspectionUri) {
 			Assert.notNull(introspectionUri, "introspectionUri cannot be null");
 			this.introspectionUri = introspectionUri;
+			this.introspectionClient = () ->
+					new NimbusOAuth2TokenIntrospectionClient(this.introspectionUri, this.clientId, this.clientSecret);
 			return this;
 		}
 
 		public OpaqueTokenConfigurer introspectionClientCredentials(String clientId, String clientSecret) {
 			Assert.notNull(clientId, "clientId cannot be null");
 			Assert.notNull(clientSecret, "clientSecret cannot be null");
-			this.introspectionClientId = clientId;
-			this.introspectionClientSecret = clientSecret;
+			this.clientId = clientId;
+			this.clientSecret = clientSecret;
+			this.introspectionClient = () ->
+					new NimbusOAuth2TokenIntrospectionClient(this.introspectionUri, this.clientId, this.clientSecret);
 			return this;
 		}
 
-		AuthenticationProvider getProvider() {
-			return new OAuth2IntrospectionAuthenticationProvider(this.introspectionUri,
-					this.introspectionClientId, this.introspectionClientSecret);
+		public OpaqueTokenConfigurer introspectionClient(OAuth2TokenIntrospectionClient introspectionClient) {
+			Assert.notNull(introspectionClient, "introspectionClient cannot be null");
+			this.introspectionClient = () -> introspectionClient;
+			return this;
+		}
+
+		OAuth2TokenIntrospectionClient getIntrospectionClient() {
+			if (this.introspectionClient != null) {
+				return this.introspectionClient.get();
+			}
+			return this.context.getBean(OAuth2TokenIntrospectionClient.class);
 		}
 	}
 
--- a/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
@@ -30,9 +30,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
-import org.springframework.security.oauth2.client.endpoint.OAuth2AuthorizationCodeGrantRequest;
-import org.springframework.security.oauth2.client.endpoint.ReactiveOAuth2AccessTokenResponseClient;
 import reactor.core.publisher.Mono;
 import reactor.util.context.Context;
 
@@ -61,6 +60,8 @@ import org.springframework.security.oauth2.client.InMemoryReactiveOAuth2Authoriz
 import org.springframework.security.oauth2.client.ReactiveOAuth2AuthorizedClientService;
 import org.springframework.security.oauth2.client.authentication.OAuth2AuthorizationCodeReactiveAuthenticationManager;
 import org.springframework.security.oauth2.client.authentication.OAuth2LoginReactiveAuthenticationManager;
+import org.springframework.security.oauth2.client.endpoint.OAuth2AuthorizationCodeGrantRequest;
+import org.springframework.security.oauth2.client.endpoint.ReactiveOAuth2AccessTokenResponseClient;
 import org.springframework.security.oauth2.client.endpoint.WebClientReactiveAuthorizationCodeTokenResponseClient;
 import org.springframework.security.oauth2.client.oidc.authentication.OidcAuthorizationCodeReactiveAuthenticationManager;
 import org.springframework.security.oauth2.client.oidc.userinfo.OidcReactiveOAuth2UserService;
@@ -88,6 +89,8 @@ import org.springframework.security.oauth2.server.resource.authentication.JwtAut
 import org.springframework.security.oauth2.server.resource.authentication.JwtReactiveAuthenticationManager;
 import org.springframework.security.oauth2.server.resource.authentication.OAuth2IntrospectionReactiveAuthenticationManager;
 import org.springframework.security.oauth2.server.resource.authentication.ReactiveJwtAuthenticationConverterAdapter;
+import org.springframework.security.oauth2.server.resource.introspection.NimbusReactiveOAuth2TokenIntrospectionClient;
+import org.springframework.security.oauth2.server.resource.introspection.ReactiveOAuth2TokenIntrospectionClient;
 import org.springframework.security.oauth2.server.resource.web.access.server.BearerTokenServerAccessDeniedHandler;
 import org.springframework.security.oauth2.server.resource.web.server.BearerTokenServerAuthenticationEntryPoint;
 import org.springframework.security.oauth2.server.resource.web.server.ServerBearerTokenAuthenticationConverter;
@@ -1364,8 +1367,9 @@ public class ServerHttpSecurity {
 		 */
 		public class OpaqueTokenSpec {
 			private String introspectionUri;
-			private String introspectionClientId;
-			private String introspectionClientSecret;
+			private String clientId;
+			private String clientSecret;
+			private Supplier<ReactiveOAuth2TokenIntrospectionClient> introspectionClient;
 
 			/**
 			 * Configures the URI of the Introspection endpoint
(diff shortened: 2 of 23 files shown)


oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/introspection/TestOAuth2TokenIntrospectorResponses.java
```diff
@@ -0,0 +1,59 @@
...
+import static org.springframework.security.oauth2.introspection.OAuth2IntrospectionClaimNames.AUDIENCE;
+import static org.springframework.security.oauth2.introspection.OAuth2IntrospectionClaimNames.CLIENT_ID;
+import static org.springframework.security.oauth2.introspection.OAuth2IntrospectionClaimNames.EXPIRES_AT;
+import static org.springframework.security.oauth2.introspection.OAuth2IntrospectionClaimNames.ISSUER;
+import static org.springframework.security.oauth2.introspection.OAuth2IntrospectionClaimNames.NOT_BEFORE;
+import static org.springframework.security.oauth2.introspection.OAuth2IntrospectionClaimNames.SCOPE;
+import static org.springframework.security.oauth2.introspection.OAuth2IntrospectionClaimNames.SUBJECT;
+import static org.springframework.security.oauth2.introspection.OAuth2IntrospectionClaimNames.USERNAME;
+
+public class TestOAuth2TokenIntrospectorResponses {
```
jzheaux: Since introspection does not have a domain object, I'm struggling with the right name for this class. `TestMaps` isn't right.

I'd like the name to stay focused on the domain, if possible, so that this class can be used for both the imperative and reactive introspection tests. Thoughts?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -288,14 +301,26 @@ public OpaqueTokenConfigurer introspectionUri(String introspectionUri) {
...
 			Assert.notNull(clientSecret, "clientSecret cannot be null");
-			this.introspectionClientId = clientId;
-			this.introspectionClientSecret = clientSecret;
+			RestTemplate restTemplate = new RestTemplate();
+			restTemplate.getInterceptors().add(new BasicAuthenticationInterceptor(clientId, clientSecret));
+			this.restOperations = restTemplate;
+			return this;
+		}
+
+		public OpaqueTokenConfigurer introspector(OAuth2TokenIntrospector introspector) {
```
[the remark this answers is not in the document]
jzheaux: Note that since this comment, I've changed the name to `OpaqueTokenIntrospectionClient`.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -288,14 +301,26 @@ public OpaqueTokenConfigurer introspectionUri(String introspectionUri) {
...
+			Assert.notNull(introspector, "introspector cannot be null");
+			this.introspector = introspector;
 			return this;
 		}
 
-		AuthenticationProvider getProvider() {
-			return new OAuth2IntrospectionAuthenticationProvider(this.introspectionUri,
-					this.introspectionClientId, this.introspectionClientSecret);
+		OAuth2TokenIntrospector getIntrospector() {
+			if (this.introspector != null) {
```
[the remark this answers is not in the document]
jzheaux: Note that since this comment, I've changed the name to `OpaqueTokenIntrospectionClient`.

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/introspection/TestOAuth2TokenIntrospectorResponses.java
jzheaux: Note that since this comment, I've renamed this to `TestOAuth2TokenIntrospectionClientResponses`. The question still stands, though, as this rename was simply to align with the change to `OAuth2TokenIntrospectionClient`.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/introspection/OAuth2TokenIntrospectionClient.java
```diff
@@ -0,0 +1,41 @@
...
+ * the contents of an OAuth 2.0 Token.
+ *
+ * A typical implementation would contact an
+ * <a href="https://tools.ietf.org/html/rfc7662" target="_blank">OAuth 2.0 Introspection Endpoint</a>
+ * hosted on an OAuth 2.0 Authorization Server.
+ *
+ * @author Josh Cummings
+ * @since 5.2
+ */
+public interface OAuth2TokenIntrospectionClient {
```
rwinch: I like the name change, but can you update the commit comment and PR description to reflect the name change?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/introspection/OAuth2TokenIntrospectionClient.java
```diff
@@ -0,0 +1,41 @@
...
+
+	/**
+	 * Introspect the given token's attributes
```
rwinch: I like the class level Javadoc, but just looking at this method the description is a bit abstract to me. Perhaps you can provide more details or a link to the class level Javadoc? 

Whatever you do here, should also be done on the reactive bits.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -1293,8 +1300,15 @@ public OpaqueTokenSpec introspectionUri(String introspectionUri) {
 			public OpaqueTokenSpec introspectionClientCredentials(String clientId, String clientSecret) {
 				Assert.hasText(clientId, "clientId cannot be empty");
 				Assert.notNull(clientSecret, "clientSecret cannot be null");
-				this.introspectionClientId = clientId;
-				this.introspectionClientSecret = clientSecret;
+				this.webClient = WebClient.builder()
+						.defaultHeader(HttpHeaders.AUTHORIZATION, basicHeaderValue(clientId, clientSecret))
```
rwinch: I think it would be easier to use this instead:

```
.defaultHeaders(h -> h.setBasicAuth(clientId, clientSecret))
```

You could then remove `basicHeaderValue`

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/introspection/NimbusReactiveOAuth2TokenIntrospectionClient.java
```diff
@@ -0,0 +1,200 @@
...
+	 * @param clientSecret The client secret for the authorized client
+	 */
+	public NimbusReactiveOAuth2TokenIntrospectionClient(String introspectionUri, String clientId, String clientSecret) {
+		Assert.hasText(introspectionUri, "introspectionUri cannot be empty");
+		Assert.hasText(clientId, "clientId cannot be empty");
+		Assert.notNull(clientSecret, "clientSecret cannot be null");
+
+		this.introspectionUri = URI.create(introspectionUri);
+		this.webClient = WebClient.builder()
+				.defaultHeader(HttpHeaders.AUTHORIZATION, basicHeaderValue(clientId, clientSecret))
```
rwinch: I think it would be easier to use this instead:

```
.defaultHeaders(h -> h.setBasicAuth(clientId, clientSecret))
```

You could then remove `basicHeaderValue`

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/introspection/OAuth2TokenIntrospectionClient.java
```diff
@@ -0,0 +1,41 @@
...
+	 *
+	 * @param token the token to introspect
+	 * @return the token's attributes, including whether or not the token is active
+	 */
+	Map<String, Object> introspect(String token);
```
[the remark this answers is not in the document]
jgrandja: Keep in mind that we will likely need to support static and dynamic custom parameters. Dynamic parameters may come from the `HttpServletRequest` or other source?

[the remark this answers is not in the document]
jzheaux: I like the idea of returning a response domain object as it provides some greater API flexibility than returning a `Map`. 

I'm not certain that it's worth it, though, since it seems to me that it would just be a `Map`-holder. In that case, I think we'll get more mileage out of having a dedicated token + attributes class like `OAuth2TokenAttributes` when the attributes get exposed in controllers. What benefits do you see us getting from a response object?

I don't yet see the benefit of a request domain object. The reason is that `token_type_hint` only defines two values, `access_token` and `refresh_token` and it would be pretty odd, I think, for folks to be presenting a resource server with a refresh token as credentials. (I'm not predicting that folks won't want to do it - I just think that's strange enough to keep it from being a reason for us to introduce a domain object.)

If a user's authorization server takes custom `token_type_hint` values or custom request parameters, it would seem they are already in custom territory and would need to pull stuff out of the upstream `Authentication` object anyway, which means a custom authentication provider.

@rwinch, feel free to chime in as well if you have thoughts about this contract.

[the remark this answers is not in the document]
jgrandja: The main concern I have with this API `Map<String, Object> introspect(String token)` is that it does not provide us the flexibility to build on it if we need to support custom request parameters at a later point. If we had a request object than adding another attribute will not break the core contract. Same applies to the response although a `Map` will provide more flexibility.

Have you conducted a search of well-known providers to see what their introspection endpoint supports for request parameters? I did a quick one and here is what I found.

[Okta](https://developer.okta.com/docs/api/resources/oidc/#introspect) - supports a few request parameters.

[IBM](https://www.ibm.com/support/knowledgecenter/en/SSPREK_9.0.5/com.ibm.isam.doc/config/concept/con_oauth_introspection.html) Here is what they have to say about the optional `token_type_hint`  parameter:

> The introspection endpoint supports use of the 'token_type_hint' as per section 2.1 (https://tools.ietf.org/html/rfc7662#section-2.1). This allows an optimization in the time of the token lookup. This does not limit the breadth of the search for the token in the token cache. Any token type will still be found even when its type is not the same as the hint.

[Keycloak](https://www.keycloak.org/docs/2.5/authorization_services/topics/service/protection/token-introspection.html) - They support the `token_type_hint` parameter as well but the value is `requesting_party_token` - which is used by a RP to introspect an access token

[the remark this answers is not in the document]
jgrandja: Also, based on Okta's and Keycloak's API - the `token_type_hint` may be either `access_token`, `refresh_token`, `requesting_party_token`, maybe even `id_token` - tells me that this client might belong in `oauth2-client`. Something that we should consider at least.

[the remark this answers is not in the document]
jzheaux: The only thing that Resource Server receives in an HTTP request is the bearer token itself, so there isn't a way for the request to indicate additional parameters to the introspection client. If custom parameters were necessary, they would likely get configured at the application level.

The majority of the parameters you indicated are related to client authentication, which can be done as a `RestTemplate` interceptor, though I think this might change to something more like what other Spring Security clients offer via `RequestEntity` `Converter`s. In that case, these would be the way that the request would be customized.

As I see it, the reason for creating a dedicated object is to provide additional request-level parameters, which Resource Server doesn't have.

That said, there are likely some legacy stories out there where clients send resource servers more than just the bearer token. Perhaps in these same stories those extra request details get transmuted to an introspection invocation. In circumstances like these, there may be an argument for a request object. Is this the kind of scenario you are referring to?

> this client might belong in `oauth2-client`

This is an interesting point because if the client goes there, then things other than Resource Server may invoke it. And in that case, I think a dedicated request object makes a lot of sense.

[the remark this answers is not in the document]
jzheaux: I think we should wait until we have a clear use case other than Bearer Token Authentication. Or, at least, we shouldn't make the change in this PR since we are still speaking in the context of how the Resource Server would use this. There's a possibility that #6544 will shed light on other use cases. Or, I've added #6798 to make sure that we add support for custom parameters, and we may discover things there, too.

In the meantime, I think the package name is wrong. Currently, it is `org.springframework.security.oauth2.introspection`, which implies more general usage of this client, but clearly that's not the intent. We should move it to `org.springframework.security.oauth2.server.resource.introspection` until it's clear that we want to expose the client more generally.

> That said, there are likely some legacy stories out there where clients send more than just the bearer token to a resource server.

I'd like to add a bit more color to this comment I made earlier.

The `AuthenticationManager` API is abstracted away from the HTTP stack altogether. So, for an application to add more request detail to an authentication mechanism (read: a special header or request parameter -- this is like adding a special authentication header to BASIC authentication, not unheard of, but a little odd), the application needs to do two separate things: first create a custom `AuthenticationDetailsSource` and second create a custom `AuthenticationProvider` that knows how to handle their custom details.

Using that notion as a basis, we could hypothetically create an introspection request object, as you indicated, that has an `Object details` field and the `OAuth2IntrospectionAuthenticationProvider` would ferry the details from `Authentication#getDetails` to `OAuth2TokenIntrospectionRequest#setDetails` or similar. But, then the user would also need to provide a custom way to turn those details into their actual custom request.

In those cases, as a developer, I'd want to invoke the "let me talk to your manager" clause and create a custom filter that address the endpoint directly:

```java
public class MyFilter extends OncePerRequestFilter {
    protected void doFilterInternal(...) {
        String token = this.bearerTokenResolver.resolve(request);
        // ... get other things from request
        Map<String, Object> response = restTemplate.exchange(requestEntity, Map.class);
        Optional.ofNullable(response.get("active")).filter(... check if active ...)
                .map(active -> toOAuth2IntrospectionAuthenticationToken(response))
                .orElse(this.bearerTokenEntryPoint.commence(...));
    }
}
```

Or, if I didn't mind tying myself to the web tier, then I'd just use `RequestContextHolder` to get what extra details I needed.

Again, to restate, this is just what I'm thinking about for custom request-level attributes. For stuff that isn't part of the request (e.g. the resource server's IP address, the client credentials, other custom, but static, attributes), these can be added via a custom `RestOperations` or perhaps better, a custom `RequestEntity` `Converter`.

jzheaux:
I force-pushed this last change, but it only changes package changes, moving the client from `org.springframework.security.oauth2.introspection` to `org.springframework.security.oauth2.server.resource.introspection`.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/introspection/OAuth2TokenIntrospectionClient.java
jzheaux: @rwinch I've updated the method-level Javadoc - https://github.com/spring-projects/spring-security/pull/6633/commits/dd2f331f5d435b5bbc51b2c8beb6dac2583f387b

Let me know if that's more in line with what you were thinking.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -288,9 +295,15 @@ JwtDecoder getJwtDecoder() {
...
 	public class OpaqueTokenConfigurer {
+		private final ApplicationContext context;
+
 		private String introspectionUri;
-		private String introspectionClientId;
-		private String introspectionClientSecret;
+		private RestOperations restOperations;
+		private OAuth2TokenIntrospectionClient introspectionClient;
+
+		public OpaqueTokenConfigurer(ApplicationContext context) {
```
jgrandja: Please remove the `public` modifier

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -301,14 +314,26 @@ public OpaqueTokenConfigurer introspectionUri(String introspectionUri) {
...
+			Assert.notNull(introspectionClient, "introspectionClient cannot be null");
+			this.introspectionClient = introspectionClient;
+			return this;
+		}
+
+		OAuth2TokenIntrospectionClient getIntrospectionClient() {
+			if (this.introspectionClient != null) {
+				return this.introspectionClient;
+			}
+			if (this.introspectionUri != null || this.restOperations != null) {
```
jgrandja: The operator should be `&&` not `||`

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -1364,8 +1367,8 @@ private ReactiveAuthenticationManager getAuthenticationManager() {
 		 */
 		public class OpaqueTokenSpec {
```
jgrandja: `OpaqueTokenSpec` should have a private constructor

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -1401,8 +1411,17 @@ public OAuth2ResourceServerSpec and() {
...
-				return new OAuth2IntrospectionReactiveAuthenticationManager(
-						this.introspectionUri, this.introspectionClientId, this.introspectionClientSecret);
+				return new OAuth2IntrospectionReactiveAuthenticationManager(getIntrospectionClient());
+			}
+
+			protected ReactiveOAuth2TokenIntrospectionClient getIntrospectionClient() {
+				if (this.introspectionClient != null) {
+					return this.introspectionClient;
+				}
+				if (this.introspectionUri != null || this.webClient != null) {
```
jgrandja: The operator should be `&&`

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
jzheaux: This is by design to guard against partial configuration. Changing to `&&` could lead partial configuration to prioritize the wrong thing (instead of just failing as it should).

For example:

```java
http
    .oauth2ResourceServer()
        .opaqueToken()
            .introspectionUri(uri)
```

This is only partially configured and is erroneous. However, if a `@Bean` were also published, then using `&&` would end up publishing the bean instead of failing on partial configuration.

That said, I think this logic reads in a confusing way, so I will express it differently. I think this is also an opportunity to make sure `.introspectionClient(client)` and `introspectionUri(uri).introspectionClientDetails(id, secret)` get equal priority, like `jwtDecoder` and `jwkSetUri` do.

PR #6640, opened 2019-03-25
Oauth2login adfs sample

selvinsource:
Added an example on how to integrate with Microsoft ADFS using the out of the box spring boot auto-configuration for Oauth2 Client.

The reason for this is that I spent several days on trying to figure out if ADFS worked with Spring oauth2-client and almost went down the path of implementing a custom AdfsUserInfoTokenServices as explained here https://stackoverflow.com/questions/43148552/how-to-configure-spring-boot-security-oauth2-for-adfs/52590323 underpinning the security of the flow.

In the end I found (after many attempts) the right combination of properties that makes ADFS works out of the box and I thought it would be useful to share.

If you accept the pull request I am going to answer the stackoverflow question above providing this sample application.

FYI, I signed the ICLA.

### DIFF
--- a/samples/boot/oauth2login/README.adoc
+++ b/samples/boot/oauth2login/README.adoc
@@ -324,6 +324,7 @@ This section shows how to configure the sample application using Microsoft AD FS
 To use Microsoft AD FS authentication system for login, you must add an Application Group in AD FS Management console.
 
 The following steps describe how to configure the application group in AD FS:
+
 . In AD FS Management, right-click on Application Groups and select Add Application Group. 
 . On the Application Group Wizard, for the name enter ADFSSSO and under Client-Server applications select Web browser accessing a web application template. Click Next.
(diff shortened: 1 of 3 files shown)


jgrandja:
@selvinsource Thanks for the PR. The stackoverflow question is related to the [Spring Security OAuth](https://spring.io/projects/spring-security-oauth) project, however, this PR is for the new OAuth support we've been building since [Spring Security](https://spring.io/projects/spring-security) 5.0.

We appreciate the PR but after reviewing the changes the only new addition is the instructions on how to setup the OAuth client in Microsoft AD FS 4.0 (the new `*.client.registration.adfs` in `application.yml` is the same as the other example client registrations). Given that those instructions may change by Microsoft, I'm not sure we want to add to the list of instructions that we currently have for the 4 existing providers. If the oauth2 client configuration required some customizations than we may have considered the PR. Either way, we appreciate you putting this together and hope to see other contributions from you in the future.

PR #6717, opened 2019-03-27
Make UnAuthenticatedServerOAuth2AuthorizedClientRepository threadsafe

philsttr:
Previously UnAuthenticatedServerOAuth2AuthorizedClientRepository used a HashMap for storing OAuth2AuthorizedClients.
UnAuthenticatedServerOAuth2AuthorizedClientRepository and its HashMap are potentially accessed by multiple threads without any synchronization.
Since HashMap is not threadsafe itself, this makes UnAuthenticatedServerOAuth2AuthorizedClientRepository not threadsafe.

Now UnAuthenticatedServerOAuth2AuthorizedClientRepository uses a ConcurrentHashMap for storing OAuth2AuthorizedClients.
Since ConcurrentHashMap is threadsafe, UnAuthenticatedServerOAuth2AuthorizedClientRepository will now be threadsafe as well.

Please backport this change to 5.1.x if possible


jgrandja:
Thanks you @philsttr for these recent improvements! I also back-ported this to 5.1.x #6730

PR #6719, opened 2019-03-27
Defer downstream filter execution if no OAuth2AuthorizedClient is found

philsttr:
Prior to this change, ServerOAuth2AuthorizedClientExchangeFilterFunction would invoke next.exchange:
- first at assembly time inside the .switchIfEmpty call.
- second at execution time inside .flatMap when a OAuth2AuthorizedClient is found.

While this double-call should not technically cause any functional problems, since the Mono returned by the first call will not be subscribed if a OAuth2AuthorizedClient is found,
it does result in a lot of unnecessary execution and object creation.  There is no technical need to invoke the downstream filters twice.

This change defers the call inside .switchIfEmpty, so that it will only execute at execution time if an OAuth2AuthorizedClient is not found.

After this change, ServerOAuth2AuthorizedClientExchangeFilterFunction will not invoke next.exchange at assembly time, and will only execute next.exchange once per subscription at execution time.

Please backport this change to 5.1.x if possible


jgrandja:
Thanks for the PR @philsttr. I also back-ported to 5.1.x #6729.

PR #6740, opened 2019-04-05
Eliminates redundant method calls

intzerodev:
Hello, 
Thank you in advance for your review!
I made some changes to avoid redundant method calls.

1. ServerOAuth2AuthorizedClientExchangeFilterFunction class

(1) changed **authorizeWithClientCredentials()**
- authorizeWithClientCredentials()
: I removed the redundant code that calls 'this.authorizedClientRepository.saveAuthorizedClient()', 
because the method is already executed after the 'this.authorizedClientResolver.clientCredentials()' is called in OAuth2AuthorizedClientResolver class.
`clientCredentials() -> getTokenResponse() -> clientCredentialsResponse() -> 'authorizedClientRepository.saveAuthorizedClient()'`

(2) changed **refreshIfNecessary()** and **shouldRefreshToken()**
- refreshIfNecessary() method execution flow
> hasTokenExpired()?
-> Yes, shouldRefreshToken()?
&ensp;&ensp;&ensp; -> Yes - authorizeWithRefreshToken()
&ensp;&ensp;&ensp; -> No - authorizeWithClientCredentials()
-> No - End
- shouldRefreshToken()
: Just checks the grant types and whether the refresh token exists. Because the refresh token would not be used in client credentials grant type, I added the grant type check code here.
Due to the changes of refreshIfNecessary() method, the shouldRefreshToken() no longer needs to call the hasTokenExpired().
This change removes redundant call of 'hasTokenExpired()' when the refresh token is used.

(3) Removed **isClientCredentialsGrantType()**
- isClientCredentialsGrantType()
: This method was used to check the type to avoid using the refresh token in client credentials grant type. Because the grant type will be checked in the updated shouldRefreshToken() method, I removed this one.

2. ServerOAuth2AuthorizedClientExchangeFilterFunctionTests class
- filterWhenClientCredentialsTokenExpiredThenGetNewToken()
: I have also modified the test class according to above changes.

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
@@ -251,27 +251,20 @@ public final class ServerOAuth2AuthorizedClientExchangeFilterFunction implements
 
 	private Mono<OAuth2AuthorizedClient> refreshIfNecessary(ClientRequest request, ExchangeFunction next, OAuth2AuthorizedClient authorizedClient) {
 		ClientRegistration clientRegistration = authorizedClient.getClientRegistration();
-		if (isClientCredentialsGrantType(clientRegistration) && hasTokenExpired(authorizedClient)) {
+		if (hasTokenExpired(authorizedClient)) {
 			return createRequest(request)
-					.flatMap(r -> authorizeWithClientCredentials(clientRegistration, r));
-		} else if (shouldRefreshToken(authorizedClient)) {
-			return createRequest(request)
-				.flatMap(r -> authorizeWithRefreshToken(next, authorizedClient, r));
+					.flatMap(r -> {
+						if (shouldRefreshToken(authorizedClient)) {
+							return authorizeWithRefreshToken(next, authorizedClient, r);
+						}
+						return authorizeWithClientCredentials(clientRegistration, r);
+					});
 		}
 		return Mono.just(authorizedClient);
 	}
 
-	private boolean isClientCredentialsGrantType(ClientRegistration clientRegistration) {
-		return AuthorizationGrantType.CLIENT_CREDENTIALS.equals(clientRegistration.getAuthorizationGrantType());
-	}
-
 	private Mono<OAuth2AuthorizedClient> authorizeWithClientCredentials(ClientRegistration clientRegistration, OAuth2AuthorizedClientResolver.Request request) {
-		Authentication authentication = request.getAuthentication();
-		ServerWebExchange exchange = request.getExchange();
-
-		return this.authorizedClientResolver.clientCredentials(clientRegistration, authentication, exchange).
-				flatMap(result -> this.authorizedClientRepository.saveAuthorizedClient(result, authentication, exchange)
-						.thenReturn(result));
+		return this.authorizedClientResolver.clientCredentials(clientRegistration, request.getAuthentication(), request.getExchange());
 	}
 
 	private Mono<OAuth2AuthorizedClient> authorizeWithRefreshToken(ExchangeFunction next,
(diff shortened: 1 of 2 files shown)


nealeu:
That's quite some area to be getting into as your first ever activity on Github.  Are you working for the NSA? ;-)

nealeu:
One thing I've spotted is that I so far see no contract that says that a side-effect of any authorizedClientResolver.clientCredentials() implementation is that authorizedClientRepository.saveAuthorizedClient() has been called.

Perhaps @rwinch has an idea as to how that may not always the case?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -251,27 +251,20 @@ public void setAccessTokenExpiresSkew(Duration accessTokenExpiresSkew) {
 
 	private Mono<OAuth2AuthorizedClient> refreshIfNecessary(ClientRequest request, ExchangeFunction next, OAuth2AuthorizedClient authorizedClient) {
 		ClientRegistration clientRegistration = authorizedClient.getClientRegistration();
-		if (isClientCredentialsGrantType(clientRegistration) && hasTokenExpired(authorizedClient)) {
```
jgrandja: The logic for a `client_credentials` client is different compared to a client authorized via `authorization_code`. A `client_credentials` client will never have a refresh token but an `authorization_code` client may have a refresh token. Please review the [spec](https://tools.ietf.org/html/rfc6749#section-4.4.3):

> A refresh token SHOULD NOT be included

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -251,27 +251,20 @@ public void setAccessTokenExpiresSkew(Duration accessTokenExpiresSkew) {
...
-	private boolean isClientCredentialsGrantType(ClientRegistration clientRegistration) {
-		return AuthorizationGrantType.CLIENT_CREDENTIALS.equals(clientRegistration.getAuthorizationGrantType());
-	}
-
 	private Mono<OAuth2AuthorizedClient> authorizeWithClientCredentials(ClientRegistration clientRegistration, OAuth2AuthorizedClientResolver.Request request) {
-		Authentication authentication = request.getAuthentication();
-		ServerWebExchange exchange = request.getExchange();
-
-		return this.authorizedClientResolver.clientCredentials(clientRegistration, authentication, exchange).
-				flatMap(result -> this.authorizedClientRepository.saveAuthorizedClient(result, authentication, exchange)
```
jgrandja: `this.authorizedClientRepository.saveAuthorizedClient()` should not have been removed

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -303,11 +296,14 @@ private boolean shouldRefreshToken(OAuth2AuthorizedClient authorizedClient) {
 		if (this.authorizedClientRepository == null) {
 			return false;
 		}
+		if (authorizedClient.getClientRegistration().getAuthorizationGrantType().equals(AuthorizationGrantType.CLIENT_CREDENTIALS)) {
```
jgrandja: Same comments as above - a `client_credentials` client would never have a refresh token and this method is logic specific to refresh token.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -303,11 +296,14 @@ private boolean shouldRefreshToken(OAuth2AuthorizedClient authorizedClient) {
...
+			return false;
+		}
 		OAuth2RefreshToken refreshToken = authorizedClient.getRefreshToken();
 		if (refreshToken == null) {
 			return false;
 		}
-		return hasTokenExpired(authorizedClient);
```
jgrandja: This should not have been removed

Issue #6624, opened 2019-03-18
Make it possible to use Spring Security with functional bean registration


sdeleuze:
We would like to start adding Spring Security support in Spring Fu (via an external contributor) but we are currently blocked because unlike Spring Boot configurations, it is not possible to inject programmatically required beans in Spring Security configuration classes.

Our goal is to provide a functional bean registration equivalent of `@EnableWebFluxSecurity` which imports following configurations:
 - `ServerHttpSecurityConfiguration`
 - `WebFluxSecurityConfiguration`
 - `ReactiveOAuth2ClientImportSelector$OAuth2ClientWebFluxSecurityConfiguration`

We would like to use these configuration classes like we use Spring Boot configuration classes in https://github.com/spring-projects/spring-fu/tree/master/autoconfigure-adapter, but autowired private field make that impossible.

To solve this, autowired fields should be injectable via protected/package private setters or constructor parameters. That has also the benefit to make these classes easily testable.

Any chance you could make these changes to allow us to move forward on functional configuration for Spring Security?

rwinch:
@sdeleuze Thanks for the report! We definitely want to enable this to progress. Are there reasons that the DSL only makes sense for functional beans and not allowing it to be used in standard Spring Security setups? I'm asking because we have had this sort of request before.

sdeleuze:
I have also that goal in mind, the Kotlin DSL that will arise on Spring Fu incubating side could be usable in regular Javaconfig mode as well, like we do on WebFlux router and bean registration side. My plan is to provide a PR for that at some point, when that will be mature enough.

jgrandja:
@sdeleuze I'm not sure it makes sense to expose setter/constructor params in the `@Configuration` classes. **NOTE:** `ServerHttpSecurityConfiguration`, `WebFluxSecurityConfiguration` and `ReactiveOAuth2ClientImportSelector` are package-private as well so these would need to be exposed as well, which again we're not really keen on.

We definitely want to help move this forward for you, I'm just not seeing how reusing `@Configuration` makes sense for functional `@Bean` registration. Maybe I'm missing something and you can help me better understand?

Would you consider using `ServerHttpSecurity` directly to configure it functionally and than register it as a `@Bean` via the `ApplicationContextInitializer` adaptor module? Let me know if this strategy would work for you.

sdeleuze:
[...]

PR #6761, opened 2019-04-09
Expose bean setters in @Configuration used by @EnableWebFluxSecurity

jgrandja:
Fixes gh-6624

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/reactive/WebFluxSecurityConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/reactive/WebFluxSecurityConfiguration.java
@@ -51,11 +51,15 @@ class WebFluxSecurityConfiguration {
 	private static final boolean isOAuth2Present = ClassUtils.isPresent(
 			REACTIVE_CLIENT_REGISTRATION_REPOSITORY_CLASSNAME, WebFluxSecurityConfiguration.class.getClassLoader());
 
-	@Autowired(required = false)
 	private List<SecurityWebFilterChain> securityWebFilterChains;
 
 	@Autowired
 	ApplicationContext context;
+
+	@Autowired(required = false)
+	void setSecurityWebFilterChains(List<SecurityWebFilterChain> securityWebFilterChains) {
+		this.securityWebFilterChains = securityWebFilterChains;
+	}
 
 	@Bean(SPRING_SECURITY_WEBFILTERCHAINFILTER_BEAN_NAME)
 	@Order(value = WEB_FILTER_CHAIN_FILTER_ORDER)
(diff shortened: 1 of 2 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/reactive/WebFluxSecurityConfiguration.java
```diff
@@ -49,12 +49,20 @@
...
-	@Autowired
 	ApplicationContext context;
 
+	@Autowired(required = false)
+	void securityWebFilterChains(List<SecurityWebFilterChain> securityWebFilterChains) {
+		this.securityWebFilterChains = securityWebFilterChains;
+	}
+
+	@Autowired
+	void context(ApplicationContext context) {
```
sdeleuze: After having a look to the implementation, I am wondering if we could not avoid this `ApplicationContext` and just leverage regular bean autowiring which would be conceptually cleaner IMO.

jgrandja: Good point - I reverted that change. I applied the same for  `ServerHttpSecurityConfiguration.beanFactory` as well so reverted that as well.

FYI, I did a force-push to start fresh.

jgrandja:
I also feel that it can be confusing to a user looking at this code and wondering why is a `@Configuration` class is being used by a functional API in it's implementation.

I agree with @rwinch, it might be the right direction to take as far as

> If it needs the logic in `ServerHttpSecurityConfiguration`, perhaps we should extract out another class that can be used by both `@Configuration` and the functional code?

sdeleuze:
@rwinch In Spring Fu case, the annotations are indeed not used, but I would like to leverage these classes to create the same infra via function bean registration like I do for Spring Boot, Spring MVC or WebFlux configuration. If the principle is not clear, please have a look to https://github.com/spring-projects/spring-fu/tree/master/autoconfigure-adapter. These initializers are leveraged by Kofu and Jafu DSLs. From my understanding, these configuration classes are used in a functional way and are needed to make it possible to have a `ServerHttpSecurity` instance. If my understanding is wrong, please let me know.

jgrandja:
@sdeleuze I spoke with @rwinch and we came up with the following approach:

```java
ServerHttpSecurityConfiguration configuration = new Configurer()
		.adapterRegistry(adapterRegistry)
		.authenticationManager(authenticationManager)
		.reactiveUserDetailsService(reactiveUserDetailsService)
		.passwordEncoder(passwordEncoder)
		.userDetailsPasswordService(userDetailsPasswordService)
		.configuration();
```

Does this work for you?

jgrandja:
@sdeleuze As per our discussion, you don't require a fluent configuration API so I've provided you standard setters with package-private access and moved `@Autowired` from field to these new setters.

sdeleuze:
That's exactly what I need for programmatic use, and I like how it improves configuration testability, so :+1: from me.

Issue #6500, opened 2019-02-01
JwtDecoders and ClientRegistrations should support RFC 8414


jzheaux:
Related to https://github.com/spring-projects/spring-security/issues/5543

`JwtDecoders` and `ClientRegistrations` can already retrieve configuration via an OIDC Provider Configuration Endpoint:

```java
JwtDecoder decoder = JwtDecoders.withOidcIssuerLocation(oidcIssuerLocation);
ClientRegistration.Builder builder = ClientRegistrations.withOidcIssuerLocation(oidcIssuerLocation);
```

[RFC 8414](https://tools.ietf.org/html/rfc8414) defines a metadata endpoint not tied directly to OIDC, meaning that we should not use the OIDC-specific method here but instead introduce a new one.

One possibility is:

```java
JwtDecoder decoder = JwtDecoders.withIssuer(issuer);
ClientRegistration.Builder builder = ClientRegistrations.withIssuer(issuer);
```

This more generic name is important since it may still perform the OIDC call as a [compatibility measure](https://tools.ietf.org/html/rfc8414#section-5) in addition to the OAuth 2.0 discovery endpoint.

rhamedy:
Hi @jzheaux 
Can I work on this feature? I don't know what is the complexity but, seems like doable. I have some knowledge of `JWT` and `OAuth2` and might require some guidance on what to keep in mind.

jzheaux:
Certainly, that would be awesome!

Please double check[ the compatibility section](https://tools.ietf.org/html/rfc8414#section-5) of the spec as we'll want to follow it carefully.

Let me know where I can help out.

rhamedy:
Hi @jzheaux , 

I am looking into this issue right now. I read some of `RFC 8414` specification as well as looked inside `JWTDecoder` and `ClientRegistration.Builder` classes. In regards to the possibility you mentioned above

> JwtDecoder decoder = JwtDecoders.withIssuer(issuer);
ClientRegistration.Builder builder = ClientRegistrations.withIssuer(issuer);

We might have to provide a mechanism to accept both issuer along with configuration i.e. `openid-configuration OR oauth-authorization-server OR etc`. Any thoughts? 

The `RFC 8414` metadata lists a ton of metadata fields. Are all those fields available already and where? if not what and what not should be added? Some pointers would be helpful. 

The `RFC 8414` also talks about some endpoints that are related `token, registration, revocation, etc.` and where can I look for these? 

[...]

PR #6765, opened 2019-04-11
Added Support for RFC 8414 OAuth 2.0 Authorization Server Metadata

rhamedy:
Fixes: gh-6500

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
@@ -31,6 +31,7 @@ import org.springframework.web.util.UriComponentsBuilder;
 
 import java.net.URI;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -38,8 +39,8 @@ import java.util.Map;
 /**
  * Allows creating a {@link ClientRegistration.Builder} from an
  * <a href="https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig">OpenID Provider Configuration</a>
- * and
- * <a href="https://tools.ietf.org/html/rfc8414#section-3">Obtaining Authorization Server Metadata</a>.
+ * or <a href="https://tools.ietf.org/html/rfc8414#section-3">Authorization Server Metadata</a> based on
+ * provided issuer.
  *
  * @author Rob Winch
  * @author Josh Cummings
@@ -49,6 +50,10 @@ import java.util.Map;
 public final class ClientRegistrations {
 	private static final String OIDC_METADATA_PATH = "/.well-known/openid-configuration";
 	private static final String OAUTH2_METADATA_PATH = "/.well-known/oauth-authorization-server";
+
+	enum ProviderType {
+		OIDCV1, OIDC, OAUTH2;
+	}
 
 	/**
 	 * Creates a {@link ClientRegistration.Builder}  using the provided
@@ -83,23 +88,42 @@ public final class ClientRegistrations {
 	 * @return a {@link ClientRegistration.Builder} that was initialized by the OpenID Provider Configuration.
 	 */
 	public static ClientRegistration.Builder fromOidcIssuerLocation(String issuer) {
-		String configuration = getIssuerConfiguration(issuer, OIDC_METADATA_PATH);
-		OIDCProviderMetadata metadata = parse(configuration, OIDCProviderMetadata::parse);
+		Map<ProviderType, String> configuration = getIssuerConfiguration(issuer, OIDC_METADATA_PATH);
+		OIDCProviderMetadata metadata = parse(configuration.get(ProviderType.OIDCV1), OIDCProviderMetadata::parse);
 		return withProviderConfiguration(metadata, issuer)
 				.userInfoUri(metadata.getUserInfoEndpointURI().toASCIIString());
 	}
 
 	/**
-	 * Creates a {@link ClientRegistration.Builder} using the provided issuer by making an
-	 * <a href="https://tools.ietf.org/html/rfc8414#section-3.1">Authorization Server Metadata Request</a> and using the
-	 * values in the <a href="https://tools.ietf.org/html/rfc8414#section-3.2">Authorization Server Metadata Response</a>
-	 * to initialize the {@link ClientRegistration.Builder}.
+	 * Unlike <strong>fromOidcIssuerLocation</strong> the <strong>fromIssuerLocation</strong> queries three different endpoints and uses the
+	 * returned response from whichever that returns successfully. When <strong>fromIssuerLocation</strong> is invoked with an issuer
+	 * the following sequence of actions take place
 	 *
-	 * <p>
-	 * For example, if the issuer provided is "https://example.com", then an "Authorization Server Metadata Request" will
-	 * be made to "https://example.com/.well-known/oauth-authorization-server". The result is expected to be an "Authorization
-	 * Server Metadata Response".
-	 * </p>
+	 * <ol>
+	 * 	<li>
+	 *     The first request is made against <i>{host}/.well-known/openid-configuration/issuer1</i> where issuer is equal to
+	 *     <strong>issuer1</strong>. See <a href="https://tools.ietf.org/html/rfc8414#section-5">Compatibility Notes</a> of RFC 8414
+	 *     specification for more details.
+	 *  </li>
+	 *  <li>
+	 *  	If the first attempt request returned non-Success (i.e. 200 status code) response then based on <strong>Compatibility Notes</strong> of
+	 *  <strong>RFC 8414</strong> a fallback <a href="https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationRequest">
+	 *  OpenID Provider Configuration Request</a> is made to <i>{host}/issuer1/.well-known/openid-configuration</i>
+	 *  </li>
+	 *  <li>
+	 *  	If the second attempted request returns a non-Success (i.e. 200 status code) response then based a final
+	 *  <a href="https://tools.ietf.org/html/rfc8414#section-3.1">Authorization Server Metadata Request</a> is being made to
+	 *  <i>{host}/.well-known/oauth-authorization-server/issuer1</i>.
+	 *  </li>
+	 * </ol>
+	 *
+	 *
+	 * As explained above, <strong>fromIssuerLocation</strong> would behave the exact same way as <strong>fromOidcIssuerLocation</strong> and that is
+	 * because <strong>fromIssuerLocation</strong> does the exact same processing as <strong>fromOidcIssuerLocation</strong> behind the scene. Use of
+	 * <strong>fromIssuerLocation</strong> is encouraged due to the fact that it is well-aligned with RFC 8414 specification and more specifically
+	 * it queries latest OIDC metadata endpoint with a fallback to legacy OIDC v1 discovery endpoint.
+	 *
+	 * The <strong>fromIssuerLocation</strong> is based on <a href="https://tools.ietf.org/html/rfc8414">RFC 8414</a> specification.
 	 *
 	 * <p>
 	 * Example usage:
@@ -110,13 +134,24 @@ public final class ClientRegistrations {
 	 *     .clientSecret("client-secret")
 	 *     .build();
 	 * </pre>
+	 *
 	 * @param issuer
 	 * @return a {@link ClientRegistration.Builder} that was initialized by the Authorization Sever Metadata Provider
 	 */
 	public static ClientRegistration.Builder fromIssuerLocation(String issuer) {
-		String configuration = getIssuerConfiguration(issuer, OIDC_METADATA_PATH, OAUTH2_METADATA_PATH);
-		AuthorizationServerMetadata metadata = parse(configuration, AuthorizationServerMetadata::parse);
-		return withProviderConfiguration(metadata, issuer);
+		Map<ProviderType, String> configuration = getIssuerConfiguration(issuer, OIDC_METADATA_PATH, OAUTH2_METADATA_PATH);
+
+		if (configuration.containsKey(ProviderType.OAUTH2)) {
+			AuthorizationServerMetadata metadata = parse(configuration.get(ProviderType.OAUTH2), AuthorizationServerMetadata::parse);
+			ClientRegistration.Builder builder = withProviderConfiguration(metadata, issuer);
+			return builder;
+		} else {
+			String response = configuration.getOrDefault(ProviderType.OIDC, configuration.get(ProviderType.OIDCV1));
+			OIDCProviderMetadata metadata = parse(response, OIDCProviderMetadata::parse);
+			ClientRegistration.Builder builder = withProviderConfiguration(metadata, issuer)
+					.userInfoUri(metadata.getUserInfoEndpointURI().toASCIIString());
+			return builder;
+		}
 	}
 
 	private static ClientRegistration.Builder withProviderConfiguration(AuthorizationServerMetadata metadata, String issuer) {
@@ -152,34 +187,47 @@ public final class ClientRegistrations {
 
 	/**
 	 * When the length of paths is equal to one (1) then it's a request for OpenId v1 discovery endpoint
-	 * hence a request to "/issuer1/.well-known/openid-configuration" is being made. Otherwise, all
-	 * three (3) discovery endpoint are queried one after another depending on result of previous query
-	 * as shown below in the following order
-	 *
-	 * 1) Request "/.well-known/openid-configuration/issuer1"
-	 *
-	 * 2) If (1) is not resolved then request "/issuer1/.well-known/openid-configuration"
-	 *
-	 * 3) If (2) is not resolved then request "/.well-known/oauth-authorization-server/issuer1"
-	 *
-	 * If none of the above is resolved then thrown an error indicating that issuer could not be
-	 * resolved.
+	 * hence the request is made to <strong>{host}/issuer1/.well-known/openid-configuration</strong>.
+	 * Otherwise, all three (3) metadata endpoints are queried one after another.
 	 *
 	 * @param issuer
 	 * @param paths
-	 * @return String Configuration Metadata
+	 * @throws IllegalArgumentException if the paths is null or empty or if none of the providers
+	 * responded to given issuer and paths requests
+	 * @return Map<String, Object> - Configuration Metadata from the given issuer
 	 */
-	private static String getIssuerConfiguration(String issuer, String... paths) {
+	private static Map<ProviderType, String> getIssuerConfiguration(String issuer, String... paths) {
 		Assert.notEmpty(paths, "paths cannot be empty or null.");
 
-		String[] urls = buildIssuerConfigurationURLs(issuer, paths);
-		for (String url: urls) {
-			String response = makeIssuerRequest(url);
-			if (response != null) {
-				return response;
-			}
+		Map<ProviderType, String> providersUrl = buildIssuerConfigurationUrls(issuer, paths);
+		Map<ProviderType, String> providerResponse = new HashMap<>();
+
+		if (providersUrl.containsKey(ProviderType.OIDC)) {
+			providerResponse = mapResponse(providersUrl, ProviderType.OIDC);
 		}
-		throw new IllegalArgumentException("Unable to resolve Configuration with the provided Issuer of \"" + issuer + "\"");
+
+		// Fallback to OpenId v1 Discovery Endpoint based on RFC 8414 Compatibility Notes
+		if (providerResponse.isEmpty() && providersUrl.containsKey(ProviderType.OIDCV1)) {
+			providerResponse = mapResponse(providersUrl, ProviderType.OIDCV1);
+		}
+
+		if (providerResponse.isEmpty() && providersUrl.containsKey(ProviderType.OAUTH2)) {
+			providerResponse = mapResponse(providersUrl, ProviderType.OAUTH2);
+		}
+
+		if (providerResponse.isEmpty()) {
+			throw new IllegalArgumentException("Unable to resolve Configuration with the provided Issuer of \"" + issuer + "\"");
+		}
+		return providerResponse;
+	}
+
+	private static Map<ProviderType, String> mapResponse(Map<ProviderType, String> providersUrl, ProviderType providerType) {
+		Map<ProviderType, String> providerResponse = new HashMap<>();
+		String response = makeIssuerRequest(providersUrl.get(providerType));
+		if (response != null) {
+			providerResponse.put(providerType, response);
+		}
+		return providerResponse;
 	}
 
 	private static String makeIssuerRequest(String uri) {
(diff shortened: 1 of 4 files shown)


rhamedy:
@jzheaux 
As per our conversations. Please see the first version of my changes. A few things 

- I have repurposed the usage of `ClientType` enum and I am open to alternative suggestions (if any)
- Refactored the existing code to come up with `reusable` utility methods 
- The existing tests are not updated and runs ✅ ensuring that existing functionality is not broken
- Added a matching unit test for `OAuth2` (matching with OpenID) 
- Added tests using `RequestDispatcher` to simulate the compatibility section of RFC 8414 

As you might have noticed, the `refactored/utility` methods in `JwtDecoders.java` and `ClientRegistrations.java` are very similar. Any suggestion on how we could avoid this duplication? Would be nice to port some of the common methods where it's accessible by both classes. 

Lastly, this is not the final change and the javadocs/comments require some more work. Once we agree on a `optimal` solution then I will proceed with addressing the comments and anything else that's necessary 👍

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
```diff
@@ -142,6 +214,74 @@ private static OIDCProviderMetadata parse(String body) {
...
+		String completePath = "";
+
+		/**
+		 * As per <a href="https://tools.ietf.org/html/rfc8414#section-3">Section 3</a> of RFC8414 the following
+		 * Creates "/.well-known/openid-configuration" OR "/.well-known/oauth-server-authorization"
+		 */
+		String wellknownPath = WELL_KNOWN_PATH + type.getIssuerIdentifier();
+
+		URI uri = URI.create(issuer);
+		String issuerPath = uri.getPath();
```
jzheaux: I think a lot of this URL work can be cleaned up with `UriComponentsBuilder`.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
```diff
@@ -100,16 +164,23 @@
...
 
-	private static String getOpenidConfiguration(String issuer) {
+	private static String getConfiguration(String issuer, ClientType type) {
 		RestTemplate rest = new RestTemplate();
 		try {
-			return rest.getForObject(issuer + "/.well-known/openid-configuration", String.class);
+			return rest.getForObject(composeServerMetadataUrl(issuer, type, true), String.class);
+		} catch(HttpClientErrorException ex) {
+			 if(ex.getStatusCode() == HttpStatus.NOT_FOUND) {
+				 return rest.getForObject(composeServerMetadataUrl(issuer, type, false), String.class);
```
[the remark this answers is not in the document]
rhamedy: Thanks for spotting this! I overlooked this. I should have at least had a check for `clientType`. Anyway, with my recent changes I have fixed this as `open id` and `oauth2` have their own dedicated methods.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
```diff
@@ -69,19 +102,50 @@
...
+				.jwkSetUri(metadata.getJWKSetURI().toASCIIString())
+				.providerConfigurationMetadata(configurationMetadata)
+				.userInfoUri(metadata.getUserInfoEndpointURI().toASCIIString())
+				.tokenUri(metadata.getTokenEndpointURI().toASCIIString())
+				.clientName(issuer);
+	}
+
+	public static ClientRegistration.Builder fromOAuth2IssuerLocation(String issuer) {
+		String configuration = getConfiguration(issuer, ClientType.OAUTH2);
+		OIDCProviderMetadata metadata = parse(configuration);
```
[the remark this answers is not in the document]
rhamedy: Make sense. I did have to remove the `getUserInfoEndpoint` though as its not available in `AuthorizationServerMetadata` 👍

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
```diff
@@ -142,6 +214,74 @@ private static OIDCProviderMetadata parse(String body) {
...
+				return issuer + wellknownPath;
+			}
+		} else if (type.equals(ClientType.OAUTH2)) {
+			/**
+			 * For OAuth always form the server metadata URL as specified in
+			 * <a href="https://tools.ietf.org/html/rfc8414#section-3">Section 3</a> of RFC8414 that would look as follow
+			 * "https://example.com/.well-known/oauth-authorization-server/issuer1/" and in absence of path in supplied
+			 * issuer the string URl would look like "https://example.com/.well-known/oauth-authorization-server"
+			 */
+			completePath = (wellknownPath + issuerPath == null || issuerPath.equals("/") ? "" : issuerPath);
```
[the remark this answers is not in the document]
rhamedy: The answer is `never`, I have refactored this too 👍

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
rhamedy: Thanks for suggesting `UriComponentBuilder` I have used that, let me know what you think of recent changes.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
```diff
@@ -142,6 +214,74 @@ private static OIDCProviderMetadata parse(String body) {
...
+	private static void validateIssuer(OIDCProviderMetadata metadata, String issuer, ClientType type) {
+		String metadataIssuer = metadata.getIssuer().getValue();
+		if (!issuer.equals(metadataIssuer)) {
+			throw new IllegalStateException("The Issuer \"" + metadataIssuer + "\" provided in the " + getTypeName(type) +
+					" did not match the requested issuer \"" + issuer + "\"");
+		}
+	}
+
+	private static String getTypeName(ClientType type) {
+		return type.equals(ClientType.OIDC) ? "OpenID Configuration" : "OAuth Configuration";
```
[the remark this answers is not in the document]
rhamedy: I agree 👍 I have refactored this. It does not exist.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
```diff
@@ -142,6 +214,74 @@ private static OIDCProviderMetadata parse(String body) {
...
+		URI uri = URI.create(issuer);
+		String issuerPath = uri.getPath();
+
+		if(type.equals(ClientType.OIDC)) {
+			/**
+			 * Construct a URL of form "https://example.com/.well-known/openid-configuration/issuer1/" where issuer1 is
+			 * the value of path retrieved from supplied issuer. In it's absence the string URL would look as follow
+			 * "https://example.com/.well-known/openid-configuration"
+			 */
+			if(firstAttempt) {
```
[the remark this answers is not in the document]
rhamedy: I threw away the idea of having a `compose url` method and instead decided to have to separate `getOpenIdConfiguration` and `getOAuth2Configuration` and that way the logic is now contained in each method.

rhamedy:
Hey @jzheaux 

Thanks for the review. The branching logic did feel a little messy. I have made some changes (please see the recent commit)

- Removed the `validateIssuer` and `validateIssuer` methods 
- Removed the `ClientType` enum
- Created `getOpenIdConfiguration` and `getOAuth2Configuration` methods that also contain logic of how to construct the url
- I was thinking of refactoring the `parse` method to something as shown at the bottom. For now, created dedicated methods. 
- Refactored the `getScopes` method to make it reusable for both `OpenId` and `OAuth` by passing the `Scope` instead of `Metadata` 

```
private static <T> T parse(String body, Class<T> clazz) {
		Assert.isTrue((clazz.equals(OIDCProviderMetadata.class) || clazz.equals(AuthorizationServerMetadata.class)),
				"clazz allowed types are OIDCProviderMetadata and AuthorizationServerMetadata.");
		try {
			return clazz.equals(OIDCProviderMetadata.class)
					? clazz.cast(OIDCProviderMetadata.parse(body))
					: clazz.cast(AuthorizationServerMetadata.parse(body));
		}
		catch (ParseException e) {
			throw new RuntimeException(e);
		}
	}
```

Note: Once we are 👍 with the implementation details, then I will do the same for `JwtDecoders`

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
```diff
@@ -142,6 +233,15 @@ private static OIDCProviderMetadata parse(String body) {
 		}
 	}
 
+	private static AuthorizationServerMetadata parseAuthorizationServerMetadata(String body) {
```
jzheaux: Yeah, these checked exceptions make the code a little noisy. You might consider:

```java
private interface ThrowingFunction<S, T, E extends Throwable> {
    T apply(S src) throws E;
}
```

And then:

```java
private static <T> T parse(String body, ThrowingFunction<String, T, ParseException> parser) {
    try {
        return parser.apply(body);
    } catch (ParseException e) {
        throw new RuntimeException(e);
    }
}
```

It's probably sixes, though it could help clean up the two public methods:

```java
public static ClientRegistration.Builder withOidcIssuerLocation(String issuer) {
    String configuration = getOpenIdConfiguration(issuer);
    OIDCProviderMetadata metadata = parse(configuration, OIDCProviderMetadata::parse);
    return withProviderConfiguration(metadata)
            .userInfoUri(metadata.getUserInfoEndpointURI().toASCIIString());
}

public static ClientRegistration.Builder withOAuth2IssuerLocation(String issuer) {
    String configuration = getOAuth2Configuration(issuer);
    AuthorizationServerMetadata metadata = parse(configuration, AuthorizationServerMetadata::parse);
    return withProviderConfiguration(metadata);
}

private static ClientRegistration.Builder withProviderConfiguration(AuthorizationServerMetadata metadata) {
    // ... everything except for user info uri
}
```

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
```diff
@@ -100,16 +113,94 @@
...
+				  * As per the <a href="https://tools.ietf.org/html/rfc8414#section-5">Section 5</a> when the first attempt for
+				  * https://example.com/.well-known/openid-configuration/issuer1 failed then for backward compatibility
+				  * check https://example.com/issuer1/.well-known/openid-configuration for Open ID only.
+				  *
+				  * Results in /issuer1/.well-known/openid-configuration where issuer is https://example.com/issuer1
+				  */
+				 String url = UriComponentsBuilder.fromUri(uri).replacePath(uri.getPath() + wellKnownPath).toUriString();
+				 return rest.getForObject(url, String.class);
+			 } else {
+				 throw new IllegalArgumentException(invalidIssuerMessage);
```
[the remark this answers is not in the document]
rhamedy: Good idea 👍

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
```diff
@@ -120,11 +211,11 @@ private static ClientAuthenticationMethod getClientAuthenticationMethod(String i
...
 			return ClientAuthenticationMethod.NONE;
 		}
-		throw new IllegalArgumentException("Only ClientAuthenticationMethod.BASIC, ClientAuthenticationMethod.POST and ClientAuthenticationMethod.NONE are supported. The issuer \"" + issuer + "\" returned a configuration of " + metadataAuthMethods);
+		throw new IllegalArgumentException("Only ClientAuthenticationMethod.BASIC, ClientAuthenticationMethod.POST and "
+				+ "ClientAuthenticationMethod.NONE are supported. The issuer \"" + issuer + "\" returned a configuration of " + metadataAuthMethods);
 	}
 
-	private static List<String> getScopes(OIDCProviderMetadata metadata) {
-		Scope scope = metadata.getScopes();
+	private static List<String> getScopesAsStringList(Scope scope) {
```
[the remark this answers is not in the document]
rhamedy: I cannot believe how I missed this 😞 I checked the `AuthorizationServerMetadata` to see if it implements and interface but, I overlooked that `OIDCProviderMetadata` subclassing `AuthorizationServerMetadata`. Thanks for pointing out 👍

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
rhamedy: That definitely simplifies the public methods. 
We might have to pass the `issuer` as well to `withProviderConfiguration`. I also had to update the error message for `issuer` mismatch to as follow 

```
throw new IllegalStateException("The Issuer \"" + metadataIssuer + "\" provided in the configuration metadata did "
					+ "not match the requested issuer \"" + issuer + "\"");
```

rhamedy:
@jzheaux pushed the recent changes to both `ClientRegistrations` and `JwtDecoders`. I have not touched the tests, please let me know if you have any ideas to make them more effective/useful. I also added some `javadoc` and once again, feel free to leave some comments.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
```diff
@@ -70,7 +78,7 @@
 	 * Example usage:
 	 * </p>
 	 * <pre>
-	 * ClientRegistration registration = ClientRegistrations.fromOidcIssuerLocation("https://example.com")
+	 * ClientRegistration registration = ClientRegistrations.withOidcIssuerLocation("https://example.com")
```
rhamedy: Ups! Gonna change them back to `fromOidcIssuerLocation` as well as below `fromOAuth2IssuerLocation` in my final changes :+1:

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/JwtDecodersTests.java
```diff
@@ -134,18 +180,69 @@ public void issuerWhenRequestedIssuerIsUnresponsiveThenThrowsIllegalArgumentExce
...
+
+	private void prepareConfigurationResponse() {
 		String body = String.format(DEFAULT_RESPONSE_TEMPLATE, this.issuer, this.issuer);
-		prepareOpenIdConfigurationResponse(body);
+		prepareConfigurationResponse(body);
 	}
 
-	private void prepareOpenIdConfigurationResponse(String body) {
+	private void prepareConfigurationResponse(String body) {
+		//this.issuer = createIssuerFromServer();
```
jzheaux: Make sure to clean this up in your final changes

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoders.java
```diff
@@ -15,26 +15,33 @@
...
 import org.springframework.web.util.UriComponentsBuilder;
 
-import static org.springframework.security.oauth2.jwt.NimbusJwtDecoder.withJwkSetUri;
-
 /**
  * Allows creating a {@link JwtDecoder} from an
  * <a href="https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig">OpenID Provider Configuration</a>.
  *
  * @author Josh Cummings
+ * @author Rafiullah Hamedy
```
jzheaux: There is also `ReactiveJwtDecoders` to take care of. 

In that case, you might consider moving `getOpenIdConfiguration` and `getOAuth2Configuration` to a package-private class, since `JwtDecoders` and `ReactiveJwtDecoders` are in the same package, though abstracting this out is something that I think can wait for another PR.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
jzheaux: That's correct. `withOAuth2IssuerLocation` was just a typo on my part.

jzheaux:
We should be careful about changing error messages, but these changes appear reasonable to me.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/registration/ClientRegistrationsTest.java
```diff
@@ -118,11 +123,45 @@ public void cleanup() throws Exception {
...
+	 *
+	 * GET /.well-known/openid-configuration/issuer1 HTTP/1.1 and responded: HTTP/1.1 404 Client Error
+	 *
+	 * and subsequent attempt
+	 *
+	 * GET /issuer1/.well-known/openid-configuration HTTP/1.1 and responded: HTTP/1.1 200 OK
+	 * @throws Exception
+	 */
+	@Test
+	public void issuerWithPathForOIDCCompatibilityAllInformaitonThenSuccess() throws Exception {
```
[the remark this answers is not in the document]
rhamedy: Thanks for pointing out. I updated the test names and replace OAuth occurrences with `Oauth2` in test names 👍

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoders.java
rhamedy: After looking at the `ReactiveJwtDecoders`, I agree it does make sense to refactor and reuse. I would be happy to work on a subsequent PR for it whenever we decide to create the issue for it 👍

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/JwtDecodersTests.java
rhamedy: Good catch! I took care of it 👍

rhamedy:
Hi @jzheaux, appreciate the feedback. There is always something new to learn. 

I have pushed my recent changes. Please let me know if you there is more room for improvement. I have added some javadoc and once feel free to point out areas that are confusing and need further clarification 👍

jzheaux:
@rwinch @jgrandja Because we initially discussed `fromOidcIssuerLocation` together to establish its name, and we are introducing an additional method of a similar nature, I'd like to get your feedback on this contract before proceeding further.

The contract this PR proposes is:

* `fromOidcIssuerLocation` - no contract changes here, but it will now include the [compatibility measures recommended in RFC 8414](https://tools.ietf.org/html/rfc8414#section-5)
* `fromOAuth2IssuerLocation` - a new method that will query [the standard endpoint outlined in RFC 8414](https://tools.ietf.org/html/rfc8414#section-3.1) to obtain the same information

rhamedy:
@jzheaux based on our recent conversation on #6765, I have updated the `ClientRegistrations.java` and `ClientRegistrationsTest.java` for now. Please let me know if you wish any changes to be made before I go ahead and update the `JWTDecoder` its unit test and update the javadocs.

rhamedy:
Hi @jzheaux, I wanted to pick your brain with regards to unit testing. Now that we have decided to make the added contract `fromIssuerLocation` query all the endpoints (i.e. original oidc path, recommended oidc path by RFC 8414 and Oauth2, that's 3) then it means that for each existing test we need to add another 3 tests to test all the paths? Does that make sense? 

For example, the following test was there from before 

```
@Test
	public void issuerWhenResponseIsTypicalThenReturnedDecoderValidatesIssuer() {
		prepareConfigurationResponse();
		this.server.enqueue(new MockResponse().setBody(JWK_SET));

		JwtDecoder decoder = JwtDecoders.fromOidcIssuerLocation(this.issuer);

		assertThatCode(() -> decoder.decode(ISSUER_MISMATCH))
				.isInstanceOf(JwtValidationException.class)
				.hasMessageContaining("This iss claim is not equal to the configured issuer");
	}
```
So we keep the existing test `as is` since it's enforcing the existing `fromOidcIssuerLocation` contract. For the newly introduced contract `fromIssuerLocation` we need the following tests

1. Test the original oidc v1 (i.e. same as above test) ensuring that `fromIssuerLocation` can be used in place of `fromOidcIssuerLocation`
2. Test the recommended oidc by RFC 8414 i.e. `/.well-known/openid-configuration/issuer` 
3. Test the `oauth2` by RFC 8414 

The 3 test cases would look as follow (refactored assertions) and using separate `mock responses` to simulate each case 

```
        @Test
	public void issuerWhenOidcV1ResponseIsTypicalThenReturnedDecoderValidatesIssuer() {
		prepareConfigurationResponseForOidcV1("issuer1");
		JwtDecoder decoder = JwtDecoders.fromIssuerLocation(this.issuer);
		assertResponseIsTypicalThenReturnedDecoderValidatesIssuer(decoder);
	}

	@Test
	public void issuerWhenOidcResponseIsTypicalThenReturnedDecoderValidatesIssuer() {
		prepareConfigurationResponseForOidc("issuer1");
		JwtDecoder decoder = JwtDecoders.fromIssuerLocation(this.issuer);
		assertResponseIsTypicalThenReturnedDecoderValidatesIssuer(decoder);
	}

	@Test
	public void issuerWhenOauth2ResponseIsTypicalThenReturnedDecoderValidatesIssuer() {
		prepareConfigurationResponseForOauth2("issuer1");
		JwtDecoder decoder = JwtDecoders.fromIssuerLocation(this.issuer);
		assertResponseIsTypicalThenReturnedDecoderValidatesIssuer(decoder);
	}

	private void assertResponseIsTypicalThenReturnedDecoderValidatesIssuer(JwtDecoder decoder) {
		assertThatCode(() -> decoder.decode(ISSUER_MISMATCH))
		.isInstanceOf(JwtValidationException.class)
		.hasMessageContaining("This iss claim is not equal to the configured issuer");
	}
```
If you notice the first line of each of the above 3 test is different and that's because we would like the server to `respond` with `200` depending on which `case` we are testing otherwise `respond` with `404`. Here is a sample 

```
private void prepareConfigurationResponseForOauth2(String path) {
		this.issuer = this.server.url(path).toString();
		String body = String.format(DEFAULT_RESPONSE_TEMPLATE, this.issuer, this.issuer);

		final Dispatcher dispatcher = new Dispatcher() {
			@Override
			public MockResponse dispatch(RecordedRequest request) throws InterruptedException {
				switch(request.getPath()) {
					case "/.well-known/oauth-authorization-server/issuer1":
						return buildSuccessMockResponse(body);
					case "/issuer1/.well-known/jwks.json":
						return buildSuccessMockResponse(JWK_SET);
				}
				return new MockResponse().setResponseCode(404);
			}
		};
		this.server.setDispatcher(dispatcher);
	}
```
The switch case for `prepareConfigurationResponseForOidc` and `prepareConfigurationResponseForOidcV1` are different. 

We do the same for all of the existing tests for both `JwtDecodersTest` and `ClientRegistrationsTest`. Does this make sense? It kinda feels like a huge volume of test and I wanted to run it by you in case I have overlooked something and we do not have to do all that?

jzheaux:
@rhamedy, it's a good question - addressing all permutations can get pretty hard to keep up with for really not that much benefit.

I think the critical logic to verify is 

1. Does the endpoint resolution logic work?, e.g. does it correctly fall back, and
2. Does it correctly parse the response into a `ClientRegistrations.Builder`.

I think these can be tested independently for the most part - I wouldn't imagine there being 3x tests, it'd probably be one (or some) tests to verify the fallback logic and another set of tests confirming the parsing logic against the oauth2 endpoint.

rhamedy:
Hi @jzheaux 

Please see the recent changes. I have refactored the `JwtDecoers`, added `unit tests` as well as `javadoc`. I am sure there will be changes to be made, please feel free to let me know. 

As far as `ClientRegistrations` is concerned, since we changed the `scope` of `fromIssuerLocation` to query all 3 endpoint, before we invoke `withProviderConfiguration` we need to know whether this is `Oidc` or `Oauth2` and that is because we will have to include `.userInfoUri(metadata.getUserInfoEndpointURI().toASCIIString());` and that means we have to either parse `AuthorizationServerMetadata` or `OidcProviderMetadata` depending on which `endpoint` responded successfully. 
I have implemented a solution for now (my recent changes) but, feel free to purpose a different one comes in mind. 

I have also added `javadoc` please feel free to point out any that need to be updated. 
Thanks for these reviews/feedback btw.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
```diff
@@ -95,21 +181,118 @@
...
+	 * Otherwise, all three (3) metadata endpoints are queried one after another.
+	 *
+	 * @param issuer
+	 * @param paths
+	 * @throws IllegalArgumentException if the paths is null or empty or if none of the providers
+	 * responded to given issuer and paths requests
+	 * @return Map<String, Object> - Configuration Metadata from the given issuer
+	 */
+	private static Map<ProviderType, String> getIssuerConfiguration(String issuer, String... paths) {
+		Assert.notEmpty(paths, "paths cannot be empty or null.");
```
rhamedy: Might have to change that message to `paths cannot be null or empty.`

rhamedy:
The use of `Map<ProviderType, String>` adds quite a bit of `branching` logic that might make the code not easy to read, would be open to suggestions on ways if we could `avoid` the `if-checks` and still be able to make `fromIssuerLocation` query all three endpoints.

jzheaux:
@rhamedy, this is now merged into `master` via f5b7706942350476bcb61f0b2b960e706543bab7. Thanks for all your efforts on this, I think it turned out well.

I squashed your commits (which is why you see a different commit hash) and then simplifed some of the branching logic that you called out - see the polish commit of 1739ef8d3c75871367e8bf9a0daf46547abf6d45 for details.

rhamedy:
Whoa! :open_mouth: That's a massive `polish`! Thanks for that. This PR definitely went through some ups and downs and change of requirements but, at the end of the day like you said turned out pretty good and thanks to your polish commit, things looks much clearer.

Issue #6778, opened 2019-04-14
Resource Server - Multi-Tenant Jwt Decoder by Issuer


gburboz:
### Summary
This is related to Issue #5351 but takes different approach to support multi-tenant Jwt Decoders by issuer

### Actual Behavior

Currently Resource Server with jwt is configured as shown below which is then configured with underlying `NimbusJwtDecoder` to decode tokens.

```yaml
security.oauth2.resourceserver:
  jwt:
    issuer-uri: https://idp.example.com
```
OR
```yaml
security.oauth2.resourceserver:
  jwt:
    jwks_url: https://idp.example.com
```

### Expected Behavior
Proposal is to add `MultiTenantDelegatingJwtDecoder` which is composed of multiple `NimbusJwtDecoder` indexed by a mandatory `issuer-uri` (`Map<URL, NimbusJwtDecoder>`) . 

Use following configuration:
```yaml
security.oauth2.resourceserver:
  multi-tenant-jwt:
    -
        jwt:
          issuer-uri: https://idp.example.com
          jwks_url: https://idp.example.com
    -
        jwt:
          issuer-uri: https://idp-other.example.com
          jwks_url: https://idp-other.example.com
```

The multi tenant decoder does initial parsing (`JWT jwt = parse(token);`) and lookup the issuer claim from parsed JWT. 

Based off issuer claim, it will look up underlying `NimbusJwtDecoder` and delegates further processing to it.

`NimbusJwtDecoder` can optionally be modified so that it will have additional `Jwt decode(JWT token)` to avoid double parsing (`JWT jwt = parse(token);`).

### Version
Spring Security 5.1.x

### Sample

If `NimbusJwtDecoder` is not to be modified, `MultiTenantDelegatingJwtDecoder ` can be implemented with pull request #6779

jzheaux:
Related to https://github.com/spring-projects/spring-security/issues/5385

gburboz:
@jzheaux and @jgrandja , can you please review pull request #6817

jzheaux:
@gburboz Thanks for reaching out! It's good to see you again, @gburboz. 

This ticket is one that we've been discussing a bit internally. We'll get back to you as soon as possible.

jgrandja:
@gburboz I think a `JwtDecoder` implementation that delegates to one of it's composed `JwtDecoder`(s) based on the `iss` claim makes a lot of sense to provide further support for multi-tenancy. This was my initial thinking when #5385 was logged.

[...]

PR #6779, opened 2019-04-15
Resource Server - Multi-Tenant Jwt Decoder by Issuer

gburboz:
Resolves issue #6778 : Resource Server - Multi-Tenant Jwt Decoder by Issuer

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MultiTenantDelegatingJwtDecoder.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MultiTenantDelegatingJwtDecoder.java
@@ -1,0 +1,76 @@
+package org.springframework.security.oauth2.jwt;
+
+import com.nimbusds.jose.JOSEObject;
+import com.nimbusds.jose.util.Base64URL;
+import com.nimbusds.jose.util.JSONObjectUtils;
+import net.minidev.json.JSONObject;
+import org.springframework.util.Assert;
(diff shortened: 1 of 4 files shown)


samples/boot/oauth2resourceserver-multitenancy/src/main/resources/application.yml
```diff
@@ -3,6 +3,14 @@ spring:
     oauth2:
       resourceserver:
         jwt:
+          issuer-uri: ${mockwebserver.url}
           jwk-set-uri: ${mockwebserver.url}/.well-known/jwks.json
+        multi-tenant-jwt:
```
gburboz: I did not have mock server that supports JWTs with iss claims hence have put these as sample

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MultiTenantDelegatingJwtDecoder.java
```diff
@@ -0,0 +1,76 @@
...
+		}
+		if (jwtDecoder == null && decoderDefault != null) {
+			jwtDecoder = decoderDefault;
+		} else {
+			throw new JwtException(String.format("Unable to determine JwtDecoder"));
+		}
+		return jwtDecoder.decode(token);
+	}
+
+	private String parseAndFindIssuer(String token) {
```
gburboz: Unfortunately this parsing occurs twice, once here and also when NimbusJwtDecoder is invoked

Issue #5907, opened 2018-09-27
Excessive security metadata being stored - possible memory optimisations


nealeu:
Version 4.2.8

### The puzzling bits 

I've been looking at memory usage in YourKit as we're looking to reduce the memory footprint of our application, and I've come across a few puzzles which may be an opportunity to improve things in Spring Security.

`DelegatingMethodSecurityMetadataSource.attributeCache` has over 26000 objects.

1. Within this there are references to `Object.notify` and `Object.wait`, as well as many internal Spring classes, which then map to empty values.

_Could these entries be reclaimed such that `attributeCache.get(key)` returns `null` instead of returning an empty list?_

2. In terms of our security requirements, 99% of those 26000 entries will be junk.  We should only have a relatively small number of `@Controller` and `@Service` methods which require security. 

_Could Spring Security emit a warning at end of scanning if their configuration is causing too much to be considered?_

3. Lastly, if I look at poor hash distribution, I find that `DelegatingMethodSecurityMetadataSource.DefaultCacheKey` is highlighted here.

Some of the hash collisions (I have up to 27 nodes for a table entry against an average of 1.3) are due to `Method.hashCode()` working with the method name and not the signature.  I'm not sure that there's much that can be done here that would benefit as we cannot get at `Method.signature` which would give us a unique String being hashed.

Other hash collisions are more puzzling in that Java's `Class.hashCode()` seems to have hash collisions (scratching my head on this one).

### My further analysis

I've edited this, as I've had a time to look further.

Regarding 1, it looks like emptyList() is used to indicate that the metadata is cached, so a non-existent entry is not possible.  

It may be possible to globally blacklist (or more likely whitelist) all the methods from Object that are effectively just a side effect of all objects being derived from Object.  I suspect that half of the entries in the attribute cache are from java.lang.Object.

rwinch:
Thanks for the feedback. Can you highlight what the actual problem you are trying to solve is? For example, is Spring Security's method security performing slowly? If so how do you reproduce?

nealeu:
Hi Rob.

We're trying to manage hosting costs by avoiding having to increase memory, so that's what had me look at what large objects are being retained.  Anything that can be done to cut this down would be good.

I think that there is potential to improve memory, startup time and runtime performance by improving this area.

What do you think to having an option to whitelist the methods on Object by returning emptyList() if targetClass is Object.class?  

[...]

PR #6794, opened 2019-04-16
Extract default metatdata sources from methodSecurityMetadataSource()

nealeu:
Allows a custom replacement for DelgatingSecurityMetdataSource to be easily configured while giving easy re-use of the default sources that implement Spring Security annotations. 

This also allows customMethodSecurityMetadataSource to be deprecated as this new method can be overridden to add one or more custom sources.  Less code. Yay!

See #5907 for reasons.

I think this is also a candidate for cherry-pick to maintenance branches.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/method/configuration/GlobalMethodSecurityConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/method/configuration/GlobalMethodSecurityConfiguration.java
@@ -357,6 +357,17 @@ public class GlobalMethodSecurityConfiguration
 	 */
 	@Bean
 	public MethodSecurityMetadataSource methodSecurityMetadataSource() {
+		List<MethodSecurityMetadataSource> sources = defaultMethodSecurityMetadataSources();
+		return new DelegatingMethodSecurityMetadataSource(sources);
+	}
+
+	/**
+	 * Creates a list of {@link MethodSecurityMetadataSource} for supported Spring Security annotations. This method
+	 * can be overridden to add one or more additional metadata sources.
+	 *
+	 * @return list of {@link MethodSecurityMetadataSource}
+	 */
+	protected List<MethodSecurityMetadataSource> defaultMethodSecurityMetadataSources() {
 		List<MethodSecurityMetadataSource> sources = new ArrayList<>();
 		ExpressionBasedAnnotationAttributeFactory attributeFactory = new ExpressionBasedAnnotationAttributeFactory(
 				getExpressionHandler());
@@ -391,7 +402,7 @@ public class GlobalMethodSecurityConfiguration
 			}
 			sources.add(jsr250MethodSecurityMetadataSource);
 		}
-		return new DelegatingMethodSecurityMetadataSource(sources);
+		return sources;
 	}
 
 	/**


nealeu:
@rwinch Will this make the 5.2 release?  Also cherry-pick to maintenance branches?

jzheaux:
While you are awaiting feedback from Rob, I have a question, @nealeu.

I guess I'm not clear on how overriding the default sources solves the caching issue since `DelegatingMethodSecurityMetadataSource` is the thing doing the caching. How does this help address #5907? I would have expected something that allows you to turn off or customize the caching aspect of `DelegatingMethodSecurityMetadataSource`.

Also, it seemed like there was some agreement that https://github.com/spring-projects/spring-security/issues/5907#issuecomment-428665327 was a reasonable plan forward, but this PR doesn't reflect that. Do you see this PR as replacing that plan or are you just pointing out a related discussion?

nealeu:
Hi Josh,

There is configuration to be done once 5907 is implemented, and that would need access to the default sources.

With this method extracted, it saves having to duplicate that.

I don't see this as replacing the plan, but it does fit with the time I can put towards this before 5.2 goes RC.

rwinch:
@nealeu Thanks for the PR!

Can you explain why this is necessary vs a user just overriding `methodSecurityMetadataSource()`?

nealeu:
Hi,

They'd have to copy/paste the contents for configuring the defaults rather than being able to call the new method for the sources.

They would then be able to override

	public MethodSecurityMetadataSource methodSecurityMetadataSource() {
		List<MethodSecurityMetadataSource> sources = defaultMethodSecurityMetadataSources();
		return new DelegatingMethodSecurityMetadataSource(sources);
	}

with something like:
	public MethodSecurityMetadataSource methodSecurityMetadataSource() {
		List<MethodSecurityMetadataSource> sources = defaultMethodSecurityMetadataSources();
                Predicate<Method> isCandidateMethod = (Method method) -> { ... }
		return new SelectivelyCachingDelgatingSecurityMetadataSource(sources, filter);
	}

As implied by your comment in https://github.com/spring-projects/spring-security/issues/5907#issuecomment-428665327.

rwinch:
Thanks for the response. I don't think we need to extract out an additional method. Wouldn't this work?

```java
public MethodSecurityMetadataSource methodSecurityMetadataSource() {
	MethodSecurityMetadataSource source = super.methodSecurityMetadataSource();
    // if your API needs a list do the following
    // List<MethodSecurityMetadataSource> sources = Arrays.asList(source);
    Predicate<Method> isCandidateMethod = (Method method) -> { ... }
    return new SelectivelyCachingDelgatingSecurityMetadataSource(source, filter);
}

jgrandja:
@nealeu Closing this PR as we haven't received a response from you in this [comment](https://github.com/spring-projects/spring-security/pull/6794#issuecomment-493506312). I do agree with @rwinch that there is no need to extract out an additional method as the sample code provided should work.

If you disagree, we can reopen and discuss further.

nealeu:
Hi @jgrandja,
Yes. I agree.  Sorry not to have replied to @rwinch's comments.
Hope you have a good festive season :christmas_tree:

Issue #6245, opened 2018-12-06
Improve ClaimAccessor and externalize coercion


jgrandja:
The methods in `ClaimAccessor` attempt to coerce a mapped value of `Object` to a specific type, for example:

- `getClaimAsBoolean`
- `getClaimAsInstant`
- `getClaimAsURL`

The coercion happens on each `get` which is not ideal. It would be more efficient if the coercion happened once and before the `Map<String, Object>` of claims is associated to the specific type of `ClaimAccessor` implementation at construction time. This change would apply to `OidcIdToken`, `OidcUserInfo`, and `Jwt`.

This capability is already implemented by `MappedJwtClaimSetConverter`, which is used by `NimbusJwtDecoder` to _normalize_ the claims before it's associated to the `Jwt`.

We should reuse/share the logic in `MappedJwtClaimSetConverter` as part of this task. For example, the `Converter` implementations in `MappedJwtClaimSetConverter` should be reused.

jgrandja:
Related #6187 #5250

PR #6799, opened 2019-04-18
Externalize coercion in ClaimAccessor

jgrandja:
Fixes gh-6245

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/ReactiveOidcIdTokenDecoderFactory.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/ReactiveOidcIdTokenDecoderFactory.java
@@ -15,11 +15,17 @@
  */
 package org.springframework.security.oauth2.client.oidc.authentication;
 
+import org.springframework.core.convert.TypeDescriptor;
+import org.springframework.core.convert.converter.Converter;
 import org.springframework.security.oauth2.client.registration.ClientRegistration;
 import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
 import org.springframework.security.oauth2.core.OAuth2Error;
 import org.springframework.security.oauth2.core.OAuth2TokenValidator;
+import org.springframework.security.oauth2.core.converter.ClaimConversionService;
+import org.springframework.security.oauth2.core.converter.ClaimTypeConverter;
+import org.springframework.security.oauth2.core.oidc.IdTokenClaimNames;
 import org.springframework.security.oauth2.core.oidc.OidcIdToken;
+import org.springframework.security.oauth2.core.oidc.StandardClaimNames;
 import org.springframework.security.oauth2.jose.jws.JwsAlgorithm;
 import org.springframework.security.oauth2.jose.jws.MacAlgorithm;
 import org.springframework.security.oauth2.jose.jws.SignatureAlgorithm;
@@ -31,7 +37,10 @@ import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
 import javax.crypto.spec.SecretKeySpec;
+import java.net.URL;
 import java.nio.charset.StandardCharsets;
+import java.time.Instant;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
@@ -61,17 +70,55 @@ public final class ReactiveOidcIdTokenDecoderFactory implements ReactiveJwtDecod
 			put(MacAlgorithm.HS512, "HmacSHA512");
 		}
 	};
+	private static final Converter<Map<String, Object>, Map<String, Object>> DEFAULT_CLAIM_TYPE_CONVERTER =
+			new ClaimTypeConverter(createDefaultClaimTypeConverters());
 	private final Map<String, ReactiveJwtDecoder> jwtDecoders = new ConcurrentHashMap<>();
 	private Function<ClientRegistration, OAuth2TokenValidator<Jwt>> jwtValidatorFactory = OidcIdTokenValidator::new;
 	private Function<ClientRegistration, JwsAlgorithm> jwsAlgorithmResolver = clientRegistration -> SignatureAlgorithm.RS256;
+	private Function<ClientRegistration, Converter<Map<String, Object>, Map<String, Object>>> claimTypeConverterFactory =
+			clientRegistration -> DEFAULT_CLAIM_TYPE_CONVERTER;
+
+	/**
+	 * Returns the default {@link Converter}'s used for type conversion of claim values for an {@link OidcIdToken}.
+	 *
+	 * @return a {@link Map} of {@link Converter}'s keyed by {@link IdTokenClaimNames claim name}
+	 */
+	public static Map<String, Converter<Object, ?>> createDefaultClaimTypeConverters() {
+		Converter<Object, ?> booleanConverter = getConverter(TypeDescriptor.valueOf(Boolean.class));
+		Converter<Object, ?> instantConverter = getConverter(TypeDescriptor.valueOf(Instant.class));
+		Converter<Object, ?> urlConverter = getConverter(TypeDescriptor.valueOf(URL.class));
+		Converter<Object, ?> collectionStringConverter = getConverter(
+				TypeDescriptor.collection(Collection.class, TypeDescriptor.valueOf(String.class)));
+
+		Map<String, Converter<Object, ?>> claimTypeConverters = new HashMap<>();
+		claimTypeConverters.put(IdTokenClaimNames.ISS, urlConverter);
+		claimTypeConverters.put(IdTokenClaimNames.AUD, collectionStringConverter);
+		claimTypeConverters.put(IdTokenClaimNames.EXP, instantConverter);
+		claimTypeConverters.put(IdTokenClaimNames.IAT, instantConverter);
+		claimTypeConverters.put(IdTokenClaimNames.AUTH_TIME, instantConverter);
+		claimTypeConverters.put(IdTokenClaimNames.AMR, collectionStringConverter);
+		claimTypeConverters.put(StandardClaimNames.EMAIL_VERIFIED, booleanConverter);
+		claimTypeConverters.put(StandardClaimNames.PHONE_NUMBER_VERIFIED, booleanConverter);
+		claimTypeConverters.put(StandardClaimNames.UPDATED_AT, instantConverter);
+		return claimTypeConverters;
+	}
+
+	private static Converter<Object, ?> getConverter(TypeDescriptor targetDescriptor) {
+		final TypeDescriptor sourceDescriptor = TypeDescriptor.valueOf(Object.class);
+		return source -> ClaimConversionService.getSharedInstance().convert(source, sourceDescriptor, targetDescriptor);
+	}
 
 	@Override
 	public ReactiveJwtDecoder createDecoder(ClientRegistration clientRegistration) {
 		Assert.notNull(clientRegistration, "clientRegistration cannot be null");
 		return this.jwtDecoders.computeIfAbsent(clientRegistration.getRegistrationId(), key -> {
 			NimbusReactiveJwtDecoder jwtDecoder = buildDecoder(clientRegistration);
-			OAuth2TokenValidator<Jwt> jwtValidator = this.jwtValidatorFactory.apply(clientRegistration);
-			jwtDecoder.setJwtValidator(jwtValidator);
+			jwtDecoder.setJwtValidator(this.jwtValidatorFactory.apply(clientRegistration));
+			Converter<Map<String, Object>, Map<String, Object>> claimTypeConverter =
+					this.claimTypeConverterFactory.apply(clientRegistration);
+			if (claimTypeConverter != null) {
+				jwtDecoder.setClaimSetConverter(claimTypeConverter);
+			}
 			return jwtDecoder;
 		});
 	}
@@ -162,5 +209,17 @@ public final class ReactiveOidcIdTokenDecoderFactory implements ReactiveJwtDecod
 	public final void setJwsAlgorithmResolver(Function<ClientRegistration, JwsAlgorithm> jwsAlgorithmResolver) {
 		Assert.notNull(jwsAlgorithmResolver, "jwsAlgorithmResolver cannot be null");
 		this.jwsAlgorithmResolver = jwsAlgorithmResolver;
+	}
+
+	/**
+	 * Sets the factory that provides a {@link Converter} used for type conversion of claim values for an {@link OidcIdToken}.
+	 * The default is {@link ClaimTypeConverter} for all {@link ClientRegistration clients}.
+	 *
+	 * @param claimTypeConverterFactory the factory that provides a {@link Converter} used for type conversion
+	 *                                  of claim values for a specific {@link ClientRegistration client}
+	 */
+	public final void setClaimTypeConverterFactory(Function<ClientRegistration, Converter<Map<String, Object>, Map<String, Object>>> claimTypeConverterFactory) {
+		Assert.notNull(claimTypeConverterFactory, "claimTypeConverterFactory cannot be null");
+		this.claimTypeConverterFactory = claimTypeConverterFactory;
 	}
 }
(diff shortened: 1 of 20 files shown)


oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverterTests.java
```diff
@@ -43,19 +42,18 @@
...
-
-		MappedJwtClaimSetConverter converter = MappedJwtClaimSetConverter
-				.withDefaults(Collections.singletonMap(JwtClaimNames.EXP, expiresAtConverter));
+	public void convertWhenIssuedAtNullThenDerivedFromExpiresAt() {
+		MappedJwtClaimSetConverter converter =
+				MappedJwtClaimSetConverter.withDefaults(Collections.emptyMap());
 
+		Instant exp = Instant.now();
 		Map<String, Object> source = new HashMap<>();
+		source.put(JwtClaimNames.EXP, exp);
```
jgrandja: @jzheaux The original test did not pass in any claims but the resulting claim set added `iss`. This does not seem right. If no claims are passed in than I would expect nothing to happen.

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverterTests.java
```diff
@@ -140,30 +138,6 @@ public void convertWhenUsingCustomConverterThenAllOtherDefaultsAreStillUsed() th
 		assertThat(target.get(JwtClaimNames.SUB)).isEqualTo("1234");
 	}
 
-	@Test
-	public void convertWhenConverterReturnsNullThenClaimIsRemoved() {
```
jgrandja: @jzheaux I removed this test as described in #6800

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverterTests.java
```diff
@@ -140,30 +138,6 @@ public void convertWhenUsingCustomConverterThenAllOtherDefaultsAreStillUsed() th
...
-				.withDefaults(Collections.emptyMap());
-
-		Map<String, Object> source = Collections.singletonMap(JwtClaimNames.ISS, null);
-		Map<String, Object> target = converter.convert(source);
-
-		assertThat(target).doesNotContainKey(JwtClaimNames.ISS);
-	}
-
-	@Test
-	public void convertWhenConverterReturnsValueWhenEntryIsMissingThenEntryIsAdded() {
```
jgrandja: @jzheaux I removed this test as described in #6800

core/src/main/java/org/springframework/security/converter/ObjectConverters.java
```diff
@@ -0,0 +1,203 @@
...
+
+	private ObjectConverters() {
+	}
+
+	/**
+	 * Returns an {@code Object} to {@code String} {@link Converter}.
+	 *
+	 * @return an {@code Object} to {@code String} {@link Converter}
+	 */
+	public static Converter<Object, String> stringConverter() {
```
jzheaux: What are your thoughts around

```java
public static final Converter<Object, String> STRING_CONVERTER = ObjectConverters::asString
```

It seems like a constant lookup would perform faster than a method invocation.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
```diff
@@ -49,7 +46,7 @@
 	 */
 	default Boolean containsClaim(String claim) {
```
jzheaux: Now that these are converted up front, what are your thoughts on adding a method:

```java
default <T> T getClaim(String claim)
```

It'd be really nice for users to be able to do something like:

```java
List<String> scopes = jwt.getClaim("scope");
```

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
```diff
@@ -143,11 +137,15 @@ default URL getClaimAsURL(String claim) {
...
 	default List<String> getClaimAsStringList(String claim) {
-		if (!this.containsClaim(claim) || !List.class.isAssignableFrom(this.getClaims().get(claim).getClass())) {
+		if (!containsClaim(claim)) {
 			return null;
 		}
-		List<String> claimValues = new ArrayList<>();
-		((List<?>) this.getClaims().get(claim)).forEach(e -> claimValues.add(e.toString()));
-		return claimValues;
+		Object claimValue = getClaims().get(claim);
+		List<String> convertedValue = ObjectConverters.listStringConverter().convert(claimValue);
```
jzheaux: I think Optional is just a bit more readable in situations like these:

```java
return Optional.ofNullable(ObjectConverters.listStringConverter().convert(claimValue))
        .orElseThrow(() -> new IllegalArgumentException...
```

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverterTests.java
jzheaux: Thanks for pointing this out. Did you mean the `iat` claim? This was to remain backward compatible with `NimbusJwtDecoderJwkSupport`, though there are perhaps other ways to achieve this.

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverterTests.java
jzheaux: I'm not keen on removing these test until we come to an agreement on #6800. This PR isn't about changing how `MappedJwtClaimSetConverter` behaves.

docs/manual/src/docs/asciidoc/_includes/servlet/preface/java-configuration.adoc
```diff
@@ -819,56 +817,6 @@ JwtDecoder jwtDecoder() {
 ```
 This will keep all the defaults, except it will override the default claim converter for `sub`.
 
-[[oauth2resourceserver-claimsetmapping-add]]
```
jzheaux: It doesn't seem to me that this PR is about changing how `MappedClaimSetConverter` works. We should defer changing its documentation to maybe #6800.

core/src/main/java/org/springframework/security/converter/ObjectConverters.java
jgrandja: > It seems like a constant lookup would perform faster than a method invocation

Not sure about this? Do you think the static method approach would be too slow?

Also keep in mind that this follows the same pattern as `RsaKeyConverters.pkcs8()` and `RsaKeyConverters.x509` and being consistent is key.

jzheaux: The reason in `RsaKeyConverters` is so that each gets its own instance of `KeyFactory`, but I didn't see any state in the `ObjectConverters`.

core/src/main/java/org/springframework/security/converter/ObjectConverters.java
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * A collection of {@link Converter} implementations
+ * that are capable of converting from {@code Object} to another type.
+ *
+ * @author Joe Grandja
+ * @since 5.2
+ * @see Converter
+ */
+public final class ObjectConverters {
```
rwinch: Have we looked into using built in Spring types (i.e. ConversionService)? It seems we should probably try to use that over writing the conversion ourselves.

core/src/main/java/org/springframework/security/converter/ObjectConverters.java
rwinch: There is a bit of a tradeoff. Holding static values improves performance a bit as it won't need to be garbage collected. However, it cannot be garbage collected when there continues to be a reference to it, so that means there is additional memory overhead.

core/src/main/java/org/springframework/security/converter/ObjectConverters.java
jgrandja: Yes, I did originally check `Converter` impl's in framework when I first wrote `ClaimAccessor` but did not find anything that met the needs. I will check once again to make sure.

core/src/main/java/org/springframework/security/converter/ObjectConverters.java
jgrandja: @rwinch It's not clear to me whether you are requesting changes here? If you are, can you please provide more detail? With the current design, garbage collection won't happen after the class is loaded. For example, using `stringConverter()` as an example, which is `static` and it references the `static` method `asString`. Are you suggesting that `stringConverter()` be changed to non-static, more like a factory method?

core/src/main/java/org/springframework/security/converter/ObjectConverters.java
jgrandja: @rwinch I checked again and the reason I did not reuse `Converter`(s) in framework is because most (if not all) `Converter` implementations in `org.springframework.core.convert.support` are package-private.

Also, I don't feel it's necessary to reuse `ConversionService` (or `GenericConversionService`) as it provides additional API's that I won't use so I'd like to keep things as simple as possible.

rwinch: I'm worry about adding these converters as we will likely need to support more complex things like timestamps that already have support for conversion. What's more is this is rather general, so it means that as we need more types of conversions we need to support them directly vs relying on ConversionService. Perhaps we could still leverage ConversionService, but the API itself accept a Converter that was adapted from ConversionService

core/src/main/java/org/springframework/security/converter/ObjectConverters.java
rwinch: I didn't really suggest one way or the other, but pointed out the tradeoffs. 

> With the current design, garbage collection won't happen after the class is loaded.

It will happen every time `stringConverter()` is used and the reference is no longer held. For example, here GC could happen as soon as `foo()` completed:

```java
private void foo() {
    Converter<Object, String> toString = stringConverter();
    ...
}
```

If `foo()` is invoked a lot then there is going to be a lot of GC that can happen.

On the flip side `ObjectConverters` creates static members and returns the same instance, there would not be GC happening and multiple invocations of `stringConverter()` would return the same object. This is a plus. However, the static values will never be GC'd which is a tradeoff if we anticipate the values only being used at startup.

Given this is likely to be used outside of startup, I would suggest using the static members as @jzheaux suggested.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
jgrandja: I'm not sure how we would implement it with the latest implementation that uses a `ClaimConversionService`. We would need to supply the `targetType` using `ClaimConversionService.getSharedInstance().convert(claimValue, targetType)`, which must be captured via the return type of caller. Off the top of my head, I'm not aware how we can capture the generic return type? If it's even possible?

Either way, this would be a feature enhancement as it's not required for this PR. Feel free to log a ticket if you would like this enhancement.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
jgrandja: I think this is a matter of preference. The current implementation is easily readable as well and it has the benefit of not creating  an extra object (`Optional`) as part of the check.

jgrandja:
@rwinch @jzheaux I applied all changes and it's ready for next review. FYI, I force pushed as I had to remove code related to #6800 that should be excluded. @rwinch I also made use of `ConversionService` via the new `ClaimConversionService` and removed `ObjectConverters`. Thanks for persisting on that as this is a much better solution.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
rwinch: @jgrandja I think this can be a separate issue. However, I agree that this would be a nice feature.

Correct me if I'm wrong @jzheaux but I think the idea would be that it presumes it is already converted to the type. No conversion is done when `getClaim` is invoked.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
rwinch: I agree this is a matter of preference. Personally, I find `Optional` harder to read. For me this is especially the case when doing more than one thing on a single line. It is unfortunate, but this often happens because the way IDEs try to format the code with method chaining.

Note that using `Optional` throughout the code base adds quite a bit of object creation which means it can impact GC.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/ReactiveOidcIdTokenDecoderFactory.java
```diff
@@ -61,17 +70,55 @@
 			put(MacAlgorithm.HS512, "HmacSHA512");
 		}
 	};
+	private static final Converter<Map<String, Object>, Map<String, Object>> DEFAULT_CLAIM_TYPE_CONVERTER =
```
rwinch: Thinking out loud a bit, but this API doesn't account for the fact that conversion may be a blocking operation. For example, if we have an `InputStream` that needs converted to an Object (i.e. reading in JSON) that is blocking and the API does not return a reactive type.

I'm wondering if in practice this is a concern in this situation since we have likely already done all the IO (it probably is converting from a String not an `InputStream`). What do we think the likelihood is that we will need to do a blocking operation during conversion? Are there other reasons we might need to block during conversion?

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
jzheaux: Agreed that it can be a separate issue. Yes, @rwinch this would simply do:

```java
default <T> T getClaim(String claim) {
    return (T) this.getClaims().get(claim);
}
```

jgrandja: I now understand what you are suggesting @jzheaux. As @rwinch mentioned

> the idea would be that it presumes it is already converted to the type

I agree that this would be a nice improvement as long as the user is aware that the claim has already been converted otherwise a `ClassCastException` will occur.

I'll add a new ticket for this.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/ReactiveOidcIdTokenDecoderFactory.java
jgrandja: @rwinch 

> I'm wondering if in practice this is a concern in this situation since we have likely already done all the IO 

The `source` of claims passed into the `Converter` should have already been parsed from it's JSON representation. For the ID Token use case, Nimbus's JSON parser would have read/parsed all claims in the raw ID Token (including custom claims) into an instance of `JWTClaimsSet`. Even for cases where there is a custom claim and it's a JSON object, Nimbus would parse it into a `Map`.

> Are there other reasons we might need to block during conversion?

At this point, I do not see  a use case where a blocking operation could happen with the default `NimbusReactiveJwtDecoder`.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
jgrandja: See #6947

jgrandja:
Thanks @jzheaux. I applied your feedback.

PR #6801, opened 2019-04-19
Set "rolePrefix" in ReactiveMethodSecurityConfiguration

ttddyy:
Currently, `GrantedAuthorityDefaults` is not considered in
`ReactiveMethodSecurityConfiguration`.
This commit updates the configuration to be aware of
`GrantedAuthorityDefaults` and update `rolePrefix` when the bean is
available.

Also, use the same instance of `DefaultMethodSecurityExpressionHandler`
when constructing `ExpressionBasedAnnotationAttributeFactory`.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/method/configuration/ReactiveMethodSecurityConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/method/configuration/ReactiveMethodSecurityConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 
 package org.springframework.security.config.annotation.method.configuration;
 
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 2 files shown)


jzheaux:
Thanks, @ttddyy! Would you also be able to write some tests that confirm that this behavior works as expected? I think none have been added to this point, but you should be able to do something similar to [`GlobalMethodSecurityConfigurationTests`](https://github.com/spring-projects/spring-security/blob/master/config/src/test/java/org/springframework/security/config/annotation/method/configuration/GlobalMethodSecurityConfigurationTests.java).

And then, is there any reason that you chose to have the  `@Bean` method depend on `DefaultMethodSecurityExpressionHandler` instead of `MethodSecurityExpressionHandler`?

ttddyy:
Hi @jzheaux 

Updated the PR for
- Added test - `ReactiveMethodSecurityConfigurationTests`
- Changed to take `MethodSecurityExpressionHandler` on `methodMetadataSource` bean

Thanks,

config/src/main/java/org/springframework/security/config/annotation/method/configuration/ReactiveMethodSecurityConfiguration.java
```diff
@@ -28,17 +29,22 @@
...
  * @author Rob Winch
+ * @author Tadaya Tsuyukubo
  * @since 5.0
  */
 @Configuration
 class ReactiveMethodSecurityConfiguration implements ImportAware {
 	private int advisorOrder;
 
+	@Autowired(required = false)
+	private GrantedAuthorityDefaults grantedAuthorityDefaults;
```
rwinch: Please add a package private setter and move the `@Autowired(required = false)` on to the setter to align with gh-6624

ttddyy: @rwinch Updated the PR to add a package private setter for Spring Fu support.

Issue #6803, opened 2019-04-20
NPE in ExpressionBasedPreInvocationAdvice


Maxvgrad:
### Summary

Get NPE while using @PreFilter before method invocation, if the method accepts more then one argument and annotation attribute _filterTarget_ is not set.

### Actual Behavior

```
java.lang.NullPointerException: null
	at org.springframework.security.access.expression.method.ExpressionBasedPreInvocationAdvice.findFilterTarget(ExpressionBasedPreInvocationAdvice.java:86) ~[spring-security-core-5.0.9.RELEASE.jar:5.0.9.RELEASE]
	at org.springframework.security.access.expression.method.ExpressionBasedPreInvocationAdvice.before(ExpressionBasedPreInvocationAdvice.java:50) ~[spring-security-core-5.0.9.RELEASE.jar:5.0.9.RELEASE]
	at org.springframework.security.access.prepost.PreInvocationAuthorizationAdviceVoter.vote(PreInvocationAuthorizationAdviceVoter.java:72) ~[spring-security-core-5.0.9.RELEASE.jar:5.0.9.RELEASE]
```

### Expected Behavior

Provide a related exception message.

### Version

spring-security-core-5.0.9.RELEASE

### Sample

```
    @PutMapping(value = "/test")
    @PreFilter("true")
    public void update(@RequestBody List<String> listForFilter, @RequestParam String param) {
        //some logic
    }
```

jzheaux:
Fixed via 59acda0

PR #6805, opened 2019-04-21
Fix NPE ExpressionBasedPreInvocationAdviceTests

Maxvgrad:
Getting NPE if @PreFilter argument filterType is not provided
and method accept more then one argument.

Add related exception message.

fixes gh-6803

### DIFF
--- a/core/src/main/java/org/springframework/security/access/expression/method/ExpressionBasedPreInvocationAdvice.java
+++ b/core/src/main/java/org/springframework/security/access/expression/method/ExpressionBasedPreInvocationAdvice.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -81,6 +81,9 @@ public class ExpressionBasedPreInvocationAdvice implements
 						"A PreFilter expression was set but the method argument type"
 								+ arg.getClass() + " is not filterable");
 			}
+		} else if (mi.getArguments().length > 1) {
+			throw new IllegalArgumentException(
+					"Unable to determine the method argument for filtering. Specify the filter target.");
 		}
 
 		if (filterTarget.getClass().isArray()) {
--- a/core/src/test/java/org/springframework/security/access/expression/method/ExpressionBasedPreInvocationAdviceTests.java
+++ b/core/src/test/java/org/springframework/security/access/expression/method/ExpressionBasedPreInvocationAdviceTests.java
@@ -1,0 +1,162 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
+ *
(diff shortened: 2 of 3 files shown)


rhamedy:
When contributing to `Spring` as per the [Contribution guidelines](https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.md) you are required to add `@author` in changed/added source files 🙂

Maxvgrad:
@rhamedy I've added an @author annotation with my real name same as in commit 👍

core/src/test/java/org/springframework/security/access/intercept/method/MockMethodInvocation.java
```diff
@@ -15,15 +15,23 @@
...
-
 @SuppressWarnings("unchecked")
 public class MockMethodInvocation implements MethodInvocation {
 	private Method method;
 	private Object targetObject;
+	private Object[] arguments = new Object[0];
+
+	public MockMethodInvocation(Object targetObject, Class clazz, String methodName, Class[] parameterTypes,
+								Object[] arguments) throws NoSuchMethodException {
+		this.method = clazz.getMethod(methodName, parameterTypes);
```
jzheaux: Is there any reason you aren't calling `this(targetObject, clazz, methodName, parameterTypes)`?

If not, would you mind changing this? I realize it's a minor point, but it's one less thing to keep track off.

core/src/test/java/org/springframework/security/access/expression/method/ExpressionBasedPreInvocationAdviceTests.java
```diff
@@ -0,0 +1,162 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests {@link ExpressionBasedPreInvocationAdvice}
+ *
+ * @author Maksim Vinogradov
+ * @since 5.1
```
jzheaux: This would go into 5.2, so please update this value.

core/src/main/java/org/springframework/security/access/expression/method/ExpressionBasedPreInvocationAdvice.java
```diff
@@ -81,6 +81,9 @@ else if (mi.getArguments().length == 1) {
 						"A PreFilter expression was set but the method argument type"
 								+ arg.getClass() + " is not filterable");
 			}
+		} else if (mi.getArguments().length > 1) {
+			throw new IllegalArgumentException(
+					"Unable to define the method argument for filtering. Specify the filter target.");
```
jzheaux: I think the error message here should use a different word than "define" since that might imply something else. Could we instead do something like "decide", "determine", or "locate"?

```suggestion
					"Unable to determine the method argument for filtering. Specify the filter target.");
```

Maxvgrad:
@jzheaux Thank you for code review! I've ressolved all discussions 👍

jzheaux:
Thanks, @Maxvgrad! This is now merged into `master`.

PR #6806, opened 2019-04-22
JWT converters improvements

ch4mpy:
Make it possible to inject granted-authorities converter in JwtAuthenticationConverter constructor

JwtGrantedAuthoritiesConverter modifications (which currently is both a granted-authority and a scope converter.):
  * rename to JwtScopesGrantedAuthoritiesConverter because it extracts authorities from token scopes only
  * extract a TokenAttributesScopesConverter and use it as collaborator
  * make prefix scope authorities configurable (so that it can be removed)
  * add constructors to inject scopes converter and scope authorities prefix
  * modify default constructor so that it uses a TokenAttributesScopesConverter extracting scopes from "scope" and "scp" claims and adding "SCOPE_" prefix (keep current behavior)

TokenAttributesScopesConverter behavior modification: scan and merge all "claims to scan" (instead of first found only)


ch4mpy:
Close as ticket was closed before submitting. Submitted just for illustration purpose.

Issue #6807, opened 2019-04-22
Potential incoherent issue and expiry instants in Oauth2 tokens


ch4mpy:
As pointed in https://github.com/spring-projects/spring-security/issues/6634 (and https://github.com/spring-projects/spring-security/issues/6557), it seems possible to build tokens with issue and expiry instants and differing with what is exposed in claims (or attributes for opaque tokens).

I propose here a rather radical way to prevent this: removing issue and expiration instants members from `AbstractOAuth2Token` in favor of claims.

I've no illusion this PR will be rejected at this stage, just a support for discussion.

jgrandja:
Closing - see comments in #6808

ch4mpy:
I understand you don't want to introduce breaking changes at this stage. Few remarks on the working branch I linked to this ticket:
  * neither marking `getIssuedAt()` and `getExpiresAt()` as abstract (and providing implementations retrieving it from attributes in concreate classes) nor removing `issuedAt` and `expiresAt` properties in `AbstractOAuth2Token` requires breaking changes
  * I don't intend to author anything concerning core objects. To me this is team responsibility: I have neither skills nor background for that. This branch is just a support to explore possibilities and get an idea of impacts. Don't expect atomic changes on it, again it is not intended to be integrated in your codebase.
  * breaking changes on this branch are due to other potential modifications orthogonal to initial concern:
    - promoting `attributes` as `AbstractOAuth2Token` as all OAuth2 tokens types have some (even refresh-tokens are issued and might expire)
    - making `AbstractOAuth2Token` a `Principal`. I was first surprised that `principal` property in `AbstractOAuth2TokenAuthenticationToken` is neither a `java.security.Principal` nor an `AuthenticatedPrincipal`. This allows to change that and quite a few things suddenly get clearer and simpler to me
  * most changes are due to many test fixtures updates and the code here was written in a big working day (so nothing that huge, actually...)

ch4mpy:
On a purely theoretical point of vue, I'm always puzzled when I have more than a single representation of a value in an app.

Here we have two different representations for issue and expiration dates in the same classes: one as `AbstractOAuth2Token` member (and as so in each subclass) and one as claim (or attribute for introspection tokens). 

IMO this is a design smell. One of the two should be removed and I think it's members.

To prevent breaking changes, it's possible to (just a suggestion):
[...]

PR #6808, opened 2019-04-22
AbstractOAuth2Token modifications:

ch4mpy:
To illustrate
 https://github.com/spring-projects/spring-security/issues/6807

Make it possible to inject granted-authorities converter in JwtAuthenticationConverter constructor

JwtGrantedAuthoritiesConverter modifications (which currently is both a granted-authority and a scope converter.):
  * rename to JwtScopesGrantedAuthoritiesConverter because it extracts authorities from token scopes only
  * extract a TokenAttributesScopesConverter and use it as collaborator
  * make prefix scope authorities configurable (so that it can be removed)
  * add constructors to inject scopes converter and scope authorities prefix
  * modify default constructor so that it uses a TokenAttributesScopesConverter extracting scopes from "scope" and "scp" claims and adding "SCOPE_" prefix (keep current behavior)

TokenAttributesScopesConverter behavior modification: scan and merge all "claims to scan" (instead of first found only)

### DIFF
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/AbstractOAuth2Token.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/AbstractOAuth2Token.java
@@ -15,12 +15,14 @@
  */
 package org.springframework.security.oauth2.core;
 
+import java.io.Serializable;
+import java.time.Instant;
+import java.util.Collections;
+import java.util.Map;
+
 import org.springframework.lang.Nullable;
 import org.springframework.security.core.SpringSecurityCoreVersion;
 import org.springframework.util.Assert;
-
-import java.io.Serializable;
-import java.time.Instant;
 
 /**
  * Base class for OAuth 2.0 Token implementations.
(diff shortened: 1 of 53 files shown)


oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2AccessToken.java
```diff
@@ -128,4 +150,19 @@ public int hashCode() {
 			return this.getValue().hashCode();
 		}
 	}
+
+	@Override
+	public Instant getIssuedAt() {
+		return getClaimAsInstant("iat");
```
jgrandja: The `iat` claim is defined by the OpenID Connect related specifications as well as `Jwt` spec. However, this claim is likely not available for standard OAuth 2.0 providers like GitHub and Facebook. But it doesn't mean they don't provide their own attribute value for `issueAt` but using a different attribute name.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2AccessToken.java
```diff
@@ -128,4 +150,19 @@ public int hashCode() {
...
+	}
+
+	@Override
+	public Instant getExpiresAt() {
+		return getClaimAsInstant("exp");
```
jgrandja: Same reasons as described for `iat`.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/AbstractOAuth2Token.java
```diff
@@ -70,49 +61,45 @@ public String getTokenValue() {
...
 
 	/**
 	 * Returns the expiration time on or after which the token MUST NOT be accepted.
 	 *
 	 * @return the expiration time of the token or null
 	 */
-	public @Nullable Instant getExpiresAt() {
-		return this.expiresAt;
-	}
+	public abstract @Nullable Instant getExpiresAt();
```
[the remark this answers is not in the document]
ch4mpy: Does it make sens to have issue and expiration instants both as plein members AND claims (or attributes) in most token implementations ?

To me, issue and expiration claims (or attributes) access is polymorphic as claims could have different name or serialization depending on the token type. Some token types could even have no claim (attributes) and require instant members for that.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/AbstractOAuth2Token.java
```diff
@@ -32,33 +34,22 @@
...
-
-	/**
-	 * Sub-class constructor.
-	 *
-	 * @param tokenValue the token value
-	 */
-	protected AbstractOAuth2Token(String tokenValue) {
-		this(tokenValue, null, null);
-	}
+	private final Map<String, Object> attributes;
```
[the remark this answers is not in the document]
ch4mpy: `issued_at` and `expires_at` seemed good candidates for being stored in a `Map<String, Object> attributes`, just like for other token types.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2AccessToken.java
ch4mpy: couldn't it be solved with more polymorphisme?

ch4mpy:
> This PR cannot be accepted

I can imagine. As per my opening comment in https://github.com/spring-projects/spring-security/issues/6807:
> I've no illusion this PR will be rejected at this stage, just a support for discussion.

Issue #6228, opened 2018-12-04
Authentication scheme case-insensitivity


jgrandja:
/cc @vpavic @jzheaux @rwinch @tnwang @nlebas

Let's revisit the Authentication scheme case-insensitivity as it seems we're not all on the same page as of yet. This is related to #6150 #5586 #6195 

I've re-read the 3 related specs and have provided references below along with my comments.

### **Key references**
** Take note of the **highlighted**

[**2.1.  Authorization Request Header Field**](https://tools.ietf.org/html/rfc6750#section-2.1)

When sending the access token in the "Authorization" request header
   field **defined by HTTP/1.1 [RFC2617]**, the client uses the **"Bearer"**
   authentication scheme to transmit the access token.

   For example:

     GET /resource HTTP/1.1
     Host: server.example.com
     Authorization: Bearer mF_9.B5f-4.1JqM

   The syntax of the "Authorization" header field for this scheme
   **follows the usage of the Basic scheme defined in Section 2 of
   [RFC2617]**.  Note that, as with Basic, **it does not conform to the
   generic syntax defined in Section 1.2 of [RFC2617] but is compatible
   with the general authentication framework** being developed for
   HTTP 1.1 [HTTP-AUTH], although it does not follow the preferred
   practice outlined therein in order to reflect existing deployments.
   The syntax for Bearer credentials is as follows:

     b64token    = 1*( ALPHA / DIGIT /
                       "-" / "." / "_" / "~" / "+" / "/" ) *"="
     credentials = "Bearer" 1*SP b64token

   Clients SHOULD make authenticated requests with a bearer token using
   the "Authorization" request header field with the "Bearer" HTTP
   authorization scheme.  Resource servers MUST support this method.

### **Comments:**
- Bearer scheme follows the usage of the Basic scheme defined in Section 2 RFC 2617
- Basic scheme **does not conform to the generic syntax** defined in Section 1.2 of RFC 2617 (which is case-insensitivity for auth-scheme) **but is compatible with the general authentication framework**. The key point in the last sentence is COMPATIBILITY. So IMO, accepting a lower-case bearer or basic auth-scheme promotes this compatibility.

### **Key references**
** Take note of the **highlighted**

[**1.2 Access Authentication Framework**](https://tools.ietf.org/html/rfc2617#section-1.2)

   HTTP provides a simple **challenge-response authentication mechanism**
   that MAY be **used by a server to challenge a client request** and by a
[...]

PR #6813, opened 2019-04-24
Detect authorization bearer token properly

TheFonz2017:
DefaultBearerTokenResolver has a bug that makes it miss authorization bearer tokens, if the "Bearer" is written in lower case as "bearer". The latter can be observed, when running Zuul as an API gateway that forwards authorization bearer tokens to downstream services. This seems to be fixed in spring-security master, but is still an issue in 5.1.x.

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolver.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolver.java
@@ -87,7 +87,7 @@ public final class DefaultBearerTokenResolver implements BearerTokenResolver {
 
 	private static String resolveFromAuthorizationHeader(HttpServletRequest request) {
 		String authorization = request.getHeader(HttpHeaders.AUTHORIZATION);
-		if (StringUtils.hasText(authorization) && authorization.startsWith("Bearer")) {
+		if (StringUtils.hasText(authorization) && authorization.toLowerCase().startsWith("bearer")) {
 			Matcher matcher = authorizationPattern.matcher(authorization);
 
 			if (!matcher.matches()) {
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolver.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolver.java
@@ -36,7 +36,9 @@ import org.springframework.util.StringUtils;
  */
 public final class DefaultBearerTokenResolver implements BearerTokenResolver {
 
-	private static final Pattern authorizationPattern = Pattern.compile("^Bearer (?<token>[a-zA-Z0-9-._~+/]+)=*$");
+	private static final Pattern authorizationPattern = Pattern.compile(
+															"^Bearer (?<token>[a-zA-Z0-9-._~+/]+)=*$", 
+															Pattern.CASE_INSENSITIVE);
 
 	private boolean allowFormEncodedBodyParameter = false;
 


pivotal-issuemaster:
@TheFonz2017 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=6813)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=6813) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@TheFonz2017 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=6813)!

TheFonz2017:
Note: there is an additional issue in this class. That is in [line 39](https://github.com/spring-projects/spring-security/pull/6813/files#diff-85e3a885f488a0b7c65b4afe5bc9b909L39). 
There, a matching pattern is given, which also assumes that the `Bearer` part is written exactly like this. This is wrong, I think.

jzheaux:
@TheFonz2017 thanks for the PR! 

This was actually already addressed in 5.2, so we'd likely instead backport https://github.com/spring-projects/spring-security/commit/63f2b6094f59cc9ded6a83ac3def4a1726890a8b instead of merging this PR.

In addition, this is currently an open discussion over on https://github.com/spring-projects/spring-security/issues/6228, and we need to make sure to resolve that before proceeding with a backport.

TheFonz2017:
Thanks for the update. I will wait then.
Just some more clarifications:
- I am using **Spring Netflix Zuul** in combination with with **Spring Security OAuth**, **Spring Security JWT** and **Spring Cloud Security** (see dependencies below) and the `@EnableOAuth2Sso` annotation
- This auto-configures OAuth2 in Zuul and as a result, a Zuul-filter will be added to the Zuul filter chain called [`OAuth2TokenRelayFilter`](https://github.com/spring-cloud/spring-cloud-security/blob/master/spring-cloud-security/src/main/java/org/springframework/cloud/security/oauth2/proxy/OAuth2TokenRelayFilter.java) which is responsible for adding an `authorization: bearer <token>` header to the proxied request for downstream services to consume. 
- As a result of [Line 83](https://github.com/spring-cloud/spring-cloud-security/blob/master/spring-cloud-security/src/main/java/org/springframework/cloud/security/oauth2/proxy/OAuth2TokenRelayFilter.java#L83) in `OAuth2TokenRelayFilter`, the resulting header is all lower-case (i.e. authorization: bearer <token>). In particular, this is the case since `OAuth2AuthenticationDetails#getTokenType()` returns a lower-case `bearer`. 
- Thus, if the discussion about case-sensitivity were to conclude with "We'll do it case sensitive in Spring", you will also have to fix the `org.springframework.security.oauth2.provider.authentication.OAuth2AuthenticationDetails` implementation.

- For all those suffering from the same issue right now: I fixed this issue with a workaround (until Spring Security 5.2 is out) by implementing a custom Zuul PRE-filter, that re-writes the authorization header before it is proxied downstream. See code below.

### Dependencies I used - for reference

```
                <dependency>
			<groupId>org.springframework.cloud</groupId>
			<artifactId>spring-cloud-starter-netflix-zuul</artifactId>
		</dependency>

		<!-- Important: A Eureka Client needs to be on the classpath for Zuul to 
			automatically detect services registered to Eureka, and do an automatic routing 
			based on registered service name. E.g. a service registered as 'users' will 
			be available as /users on the Zuul proxy and requests will be forwarded to 
			the 'users' service. -->
		<dependency>
			<groupId>org.springframework.cloud</groupId>
			<artifactId>spring-cloud-starter-netflix-eureka-client</artifactId>
		</dependency>

		<!-- Required for health checks and info pages advertised by the service 
			and read by Eureka -->
		<dependency>
			<groupId>org.springframework.boot</groupId>
			<artifactId>spring-boot-starter-actuator</artifactId>
		</dependency>

		<!-- For OkHttp-client based routing filter. -->
		<dependency>
			<groupId>com.squareup.okhttp3</groupId>
			<artifactId>okhttp</artifactId>
		</dependency>

		<!-- For @ConfigurationProperties usage in GroovyFilterPathConfig.java -->
		<dependency>
			<groupId>org.springframework.boot</groupId>
			<artifactId>spring-boot-configuration-processor</artifactId>
			<optional>true</optional>
		</dependency>

		<!-- Groovy Language for Groovy Filters. Enable this, if you would like 
			to use Groovy filters. Otherwise you can get rid of this dependency. -->
		<dependency>
			<groupId>org.codehaus.groovy</groupId>
			<artifactId>groovy-all</artifactId>
			<version>2.4.9</version>
		</dependency>

		<dependency>
			<groupId>org.springframework.cloud</groupId>
			<artifactId>spring-cloud-starter-oauth2</artifactId>
		</dependency>

		<dependency>
			<groupId>org.springframework.boot</groupId>
			<artifactId>spring-boot-starter-security</artifactId>
		</dependency>

		<dependency>
			<groupId>org.springframework.boot</groupId>
			<artifactId>spring-boot-starter-web</artifactId>
		</dependency>

		<dependency>
			<groupId>org.springframework.security.oauth</groupId>
			<artifactId>spring-security-oauth2</artifactId>
			<version>2.3.5.RELEASE</version>
		</dependency>

		<dependency>
			<groupId>org.springframework.security</groupId>
			<artifactId>spring-security-jwt</artifactId>
			<version>1.0.10.RELEASE</version>
		</dependency>

		<dependency>
			<groupId>org.springframework.boot</groupId>
			<artifactId>spring-boot-starter-test</artifactId>
			<scope>test</scope>
		</dependency>
	</dependencies>


	<dependencyManagement>
		<dependencies>
			<dependency>
				<groupId>org.springframework.cloud</groupId>
				<artifactId>spring-cloud-dependencies</artifactId>
				<version>Greenwich.RELEASE</version>
				<type>pom</type>
				<scope>import</scope>
			</dependency>

			<dependency>
				<groupId>org.springframework.security</groupId>
				<artifactId>spring-security-bom</artifactId>
				<version>5.1.5.RELEASE</version>
				<type>pom</type>
				<scope>import</scope>
			</dependency>

		</dependencies>
	</dependencyManagement>
```

### Authorization Header Re-Writing Zuul PRE-Filter

```java
import static org.springframework.cloud.netflix.zuul.filters.support.FilterConstants.PRE_TYPE;

import java.util.Map;

import org.springframework.cloud.netflix.zuul.filters.ZuulProperties;
import org.springframework.cloud.netflix.zuul.filters.discovery.DiscoveryClientRouteLocator;

import com.netflix.zuul.ZuulFilter;
import com.netflix.zuul.context.RequestContext;
import com.netflix.zuul.exception.ZuulException;

public class ZuulAuthorizationHeaderProxyFilter extends ZuulFilter {

    @SuppressWarnings("unused")
    private ZuulProperties zuulProperties;
    
    @SuppressWarnings("unused")
    private DiscoveryClientRouteLocator routeLocator;
    
    /**
     * Creates a new Zuul filter for canary testing using the given properties.
     * @param zuulProperties the Zuul configurations as given in application.yml
     * @param routeLocator the route locator.
     */
    public ZuulAuthorizationHeaderProxyFilter(ZuulProperties zuulProperties, DiscoveryClientRouteLocator routeLocator) {
        this.zuulProperties = zuulProperties;
        this.routeLocator = routeLocator;
    }
    
    @Override
    public boolean shouldFilter() {
        // Only run, if there is a Zuul request header set for the authorization,
        // and if it contains a bearer token.
        RequestContext ctx = RequestContext.getCurrentContext();
        Map<String, String> zuulRequestHeaders = ctx.getZuulRequestHeaders();
        
        String authorizationHeaderValue = zuulRequestHeaders.get("authorization");
        authorizationHeaderValue = authorizationHeaderValue != null ? authorizationHeaderValue : zuulRequestHeaders.get("Authorization");
        
        return (authorizationHeaderValue != null) && (authorizationHeaderValue.trim().toLowerCase().startsWith("bearer"));
    }

    @Override
    public Object run() throws ZuulException {
        RequestContext ctx = RequestContext.getCurrentContext();
        Map<String, String> zuulRequestHeaders = ctx.getZuulRequestHeaders();
        
        String authorizationHeaderValue = zuulRequestHeaders.get("authorization");
        authorizationHeaderValue = authorizationHeaderValue != null ? authorizationHeaderValue : zuulRequestHeaders.get("Authorization");
        
        if(authorizationHeaderValue == null)
            throw new ZuulException("Error! Could not retrieve authorization header from Zuul request headers.", 500, "Header not found. This must be an illegal state.");
        
        authorizationHeaderValue = authorizationHeaderValue.replace("bearer", "Bearer");
        
        zuulRequestHeaders.put("authorization", authorizationHeaderValue);
        
        return null;
    }

    @Override
    public String filterType() {
        return PRE_TYPE;
    }

    @Override
    public int filterOrder() {
        return 11; // one after the OAuth2TokenRelayFilter
    }
}
```

PR #6817, opened 2019-04-25
Resource Server - Multi-Tenant Jwt Decoder by Issuer

gburboz:
Resolves issue #6778 : Resource Server - Multi-Tenant Jwt Decoder by Issuer

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MultiTenantDelegatingJwtDecoder.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MultiTenantDelegatingJwtDecoder.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2002-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.oauth2.jwt;
+
+import com.nimbusds.jose.JOSEObject;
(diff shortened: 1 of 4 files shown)


samples/boot/oauth2resourceserver-multitenancy/src/main/resources/application.yml
```diff
@@ -3,6 +3,14 @@ spring:
     oauth2:
       resourceserver:
         jwt:
+          issuer-uri: ${mockwebserver.url}
           jwk-set-uri: ${mockwebserver.url}/.well-known/jwks.json
+        multi-tenant-jwt:
```
gburboz: I did not have mock server that supports JWTs with iss claims hence have put these as sample. They need not necessarily support JWT access tokens with iss claim in it. Please suggest how to use mock JWTs here w/ iss claim

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MultiTenantDelegatingJwtDecoder.java
```diff
@@ -0,0 +1,115 @@
...
+		} else {
+			throw new JwtException(String.format("Unable to determine JwtDecoder"));
+		}
+		return jwtDecoder.decode(token);
+	}
+
+	private String parseAndFindIssuer(String token) {
+		try {
+			Base64URL[] parts = JOSEObject.split(token);
+			JSONObject payload = JSONObjectUtils.parse(parts[1].decodeToString());
```
gburboz: Unfortunately this parsing is repeated again when `NimbusJwtDecoder` is invoked. If we limit this delegates to `NimbusJwtDecoder ` and refactor the these two class we may be able to avoid double parsing.

gburboz:
Alternate approach to **avoid double parsing** is [NimbusJwtMultiTenantDecoder.java](https://github.com/gburboz/spring-security/blob/feature/6779-rs-multitenant-v2/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtMultiTenantDecoder.java) with refactored [NimbusJwtDecoder.java](https://github.com/gburboz/spring-security/blob/feature/6779-rs-multitenant-v2/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java)

jzheaux:
Closing in favor of https://github.com/spring-projects/spring-security/issues/6778#issuecomment-502902575 so the JWT is not parsed multiple times.

Issue #6811, opened 2019-04-23
Align code in oauth2-client extensions for WebClient


jgrandja:
The `WebClient` extensions for OAuth 2.0 Client are supported via `ServletOAuth2AuthorizedClientExchangeFilterFunction` (Servlet) and `ServerOAuth2AuthorizedClientExchangeFilterFunction` (WebFlux).

The `ServerOAuth2AuthorizedClientExchangeFilterFunction` (WebFlux) uses a collaborator `OAuth2AuthorizedClientResolver` to realize part of the feature set. However, `ServletOAuth2AuthorizedClientExchangeFilterFunction` (Servlet) doesn't have such a collaborator and therefore all the code is contained within. We should consider adding an `OAuth2AuthorizedClientResolver` equivalent for the Servlet version to align the code and provide consistency.

Furthermore, as we continue to add support for other grant types, e.g. Resource Owner Password Credentials #6003, the code within each `ExchangeFilterFunction` could grow quite a bit making it more complex and harder to maintain. It is also preferred that the code in each `ExchangeFilterFunction` is aligned/consistent and reused wherever possible. However, we need to be careful with reuse and ensure we don't introduce a blocking operation within the reactive implementation.

As part of this decomposition exercise, we should consider re-structuring components/collaborators so they can potentially be reused by other technology stacks, e.g. WebFlux's `WebSocketClient` #6711, `RestTemplate` or Feign Client.

Related #6683, #6780

This issue is divided into the following tasks.

### Servlet

- [x] [#17](https://github.com/spring-projects/spring-security-team-tasks/issues/17) Introduce `OAuth2AuthorizedClientProvider`
- [x] [#18](https://github.com/spring-projects/spring-security-team-tasks/issues/18) Implement `authorization_code` `OAuth2AuthorizedClientProvider`
- [x] [#19](https://github.com/spring-projects/spring-security-team-tasks/issues/19) Implement `client_credentials` `OAuth2AuthorizedClientProvider`
- [x] [#20](https://github.com/spring-projects/spring-security-team-tasks/issues/20) Add `refresh_token` `OAuth2AccessTokenResponseClient`
- [x] [#21](https://github.com/spring-projects/spring-security-team-tasks/issues/21) Implement `refresh_token` `OAuth2AuthorizedClientProvider`
- [x] [#22](https://github.com/spring-projects/spring-security-team-tasks/issues/22) Implement delegating `OAuth2AuthorizedClientProvider`
[...]

Issue #5633, opened 2018-08-02
OAuth2LoginAuthenticationProvider uses OAuth2AuthorizationCodeAuthenticationProvider


jgrandja:
The `OAuth2LoginAuthenticationProvider` performs most of the same logic that exists in `OAuth2AuthorizationCodeAuthenticationProvider`. We should re-factor to allow for reuse.

rwinch:
Related to this is that OAuth2LoginAuthenticationToken could extend OAuth2AuthorizationCodeAuthenticationToken which seems logical and would allow the existing OAuth2 Log In code to be passive.

pdelert:
@jgrandja @rwinch can I take this enhancement ?

jgrandja:
@pdelert That would be great! Thank you.

pdelert:
@jgrandja here are my small changes, please have a look here #7356

rh-id:
hi @jgrandja , @pdelert . do you mind if i takeover this issue?

jgrandja:
That would be great @rh-id. But let's wait for confirmation from @pdelert as they have an open PR #7356 but I haven't heard back from them in a while. Please wait 1-2 weeks for a confirmation and if we don't hear from them then go ahead with the PR. Thanks!

rh-id:
hi @jgrandja i had created PR #8170

PR #6845, opened 2019-05-06
Introduce OAuth2AuthorizedClient Manager/Provider

jgrandja:
This PR addresses the Servlet implementation for #6811  - the Reactive implementation will be submitted in a separate PR.

The primary responsibility of an implementation of an `OAuth2AuthorizedClientProvider` is to **authorize** (or **re-authorize**) a client. This interface would be implemented by specific grant types, e.g. `authorization_code`, `refresh_token`, `client_credentials`, `password`, custom/extension grants, etc.


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2ClientAuthorizationProvider.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.client;
+
+/**
+ *
+ * @author Joe Grandja
+ * @since 5.2
+ */
+public interface OAuth2ClientAuthorizationProvider<T extends AbstractOAuth2ClientAuthorizationContext> {
```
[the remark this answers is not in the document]
jgrandja: I actually did consider the name `OAuth2AuthorizedClientManager` but decided otherwise since it doesn't really **manage** `OAuth2AuthorizedClient` instances. It's sole purpose it to authorize a client and therefore **provide** a new `OAuth2AuthorizedClient`. As far as managing  `OAuth2AuthorizedClient`(s), this is the role of the `OAuth2AuthorizedClientService`.

So I don't feel adding `Manager` to the name makes sense.

I also considered the name `OAuth2AuthorizedClientProvider` but liked `OAuth2ClientAuthorizationProvider` as it aligned with `AuthenticationProvider`. However, I still think `OAuth2AuthorizedClientProvider` could work as well.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultRefreshTokenTokenResponseClient.java
```diff
@@ -0,0 +1,133 @@
...
+ * an access token credential at the Authorization Server's Token Endpoint.
+ *
+ * @author Joe Grandja
+ * @since 5.2
+ * @see OAuth2AccessTokenResponseClient
+ * @see OAuth2RefreshTokenGrantRequest
+ * @see OAuth2AccessTokenResponse
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc6749#section-6">Refreshing an Access Token</a>
+ */
+public final class DefaultRefreshTokenTokenResponseClient implements OAuth2AccessTokenResponseClient<OAuth2RefreshTokenGrantRequest> {
```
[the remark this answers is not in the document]
jgrandja: The naming convention follows the existing implementation `DefaultClientCredentialsTokenResponseClient` and `DefaultAuthorizationCodeTokenResponseClient`, which are the default implementations used for a specific grant and in a Servlet environment.

Also, `DefaultOAuth2UserService` follows this convention.

Do you still think we should break away from this naming convention?

> I realize that we don't currently expose the RestOperations

It is exposed via `DefaultRefreshTokenTokenResponseClient.setRestOperations()`

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2ClientAuthorizationProvider.java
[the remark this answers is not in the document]
rwinch: For reactive will need to return a different type `Mono<OAuth2AuthorizedClient>` also. Nowhere else do we try to combine reactive and not reactive, so I don't view this as a good reason for generics.

Similarly, we don't generalize service APIs but instead the types being passed in. Looking at Spring there are variants of `HandlerMethodArgumentResolver` for servlet, messaging, and reactive. I think the same will apply here.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultRefreshTokenTokenResponseClient.java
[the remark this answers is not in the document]
rwinch: Thanks for pointing that out. 

I believe this naming happened for historical reasons. Initially we had some `Default*` implementations that did not expose any implementation details. Later, we exposed them.

With the other existing and related OAuth classes being named `Default*` I agree that we should leave it as is.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2ClientAuthorizationProvider.java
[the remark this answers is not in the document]
rwinch: I can by into `Provider`, but I strongly believe it should start with `OAuth2AuthorizedClient` since that is what it is providing.

For what it is worth, I believe that `OAuth2ClientAuthorizationProvider` does not align with `AuthenticationProvider`. `AuthenticationProvider` returns an `Authentication`. This API returns an `OAuth2AuthorizedClient` and not an `OAuth2ClientAuthorization` nor a `OAuth2Client` as the name `OAuth2ClientAuthorizationProvider` would imply.

rwinch:
It's worth asking ourselves...are we trying to just look up an access token or combine it with a lookup of the current access token with resolving it if it isn't found.

Another concern I have with the current APIs is that method inputs are things like `ClientRegistration` which might not be very easily accessed without injecting multiple collaborators. I rather like how we currently support `WebClient` support where the user can provide just a `clientRegistrationId` or the `ClientRegistration` and the support will handle the details for the user. I also like how things like `HttpServletRequest` can be defaulted to the current `ThreadLocal`.

To help us understand what is going to be of the most use to our users, I'd like the PR to include some sample usage of the API (in services not just controllers) too.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2ClientAuthorizationProvider.java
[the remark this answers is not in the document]
jgrandja: Ok I'll rename to `OAuth2AuthorizedClientProvider`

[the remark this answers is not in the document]
jgrandja: > For reactive will need to return a different type Mono<OAuth2AuthorizedClient>

Yes, I'm aware of that but this discussion is about the input parameter which can be reused. As an example, both `WebClientReactiveAuthorizationCodeTokenResponseClient` and `DefaultAuthorizationCodeTokenResponseClient` use the same input `OAuth2AuthorizationCodeGrantRequest`. The same pattern applies here.

If you still believe we should get rid of the generic type than are you suggesting the interface to be changed as follows?

`OAuth2AuthorizedClient authorize(AbstractOAuth2ClientAuthorizationContext authorizationContext);`

jgrandja:
> are we trying to just look up an access token or combine it with a lookup of the current access token with resolving it if it isn't found

The `OAuth2ClientAuthorizationProvider` should handle both use-cases - authorizing a new client OR re-authorizing an existing client (e.g. access token is expired).

The flow is triggered by the input to `ServletOAuth2ClientAuthorizationContext.authorize(ClientRegistration ...` (authorize new client) OR `ServletOAuth2ClientAuthorizationContext.reauthorize(OAuth2AuthorizedClient ...` (re-authorize existing client because the access token is expired)

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2ClientAuthorizationProvider.java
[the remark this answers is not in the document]
rwinch: > Yes, I'm aware of that but this discussion is about the input parameter which can be reused. 

For reactive the output cannot be reused, so we would never use the reactive generic type for the input.

This really doesn't align with code else where and I think just complicates things. Rather than having a generic API I'd like it to be focused on servlet, messaging, or reactive. We won't be able to reuse the implementation and I don't see us ever having a class consume the interface in a way that is not specific to servlet, reactive, or messaging.

> If you still believe we should get rid of the generic type than are you suggesting the interface to be changed as follows?
> 
> OAuth2AuthorizedClient authorize(AbstractOAuth2ClientAuthorizationContext authorizationContext);

I think it would be changed to the following unless you can come up with scenarios where a class would be able to consume the interface with AbstractOAuth2ClientAuthorizationContext as the context vs a concrete subclass.

```
OAuth2AuthorizedClient authorize(ServletOAuth2ClientAuthorizationContext authorizationContext)
```

rwinch:
Can we put together a sample of its usage? Also demo what our internal classes will look like when consuming this API?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2ClientAuthorizationProvider.java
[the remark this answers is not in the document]
rwinch: Thanks. When you do can you rename the PR subject too please?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -364,103 +367,59 @@ private void populateDefaultOAuth2AuthorizedClient(Map<String, Object> attrs) {
...
-	private OAuth2AuthorizedClient getAuthorizedClient(String clientRegistrationId, Map<String, Object> attrs) {
-		ClientRegistration clientRegistration = this.clientRegistrationRepository.findByRegistrationId(clientRegistrationId);
-		if (clientRegistration == null) {
-			throw new IllegalArgumentException("Could not find ClientRegistration with id " + clientRegistrationId);
-		}
-		if (isClientCredentialsGrantType(clientRegistration)) {
-			return authorizeWithClientCredentials(clientRegistration, attrs);
+	private Mono<OAuth2AuthorizedClient> authorizedClient(Map<String, Object> attrs) {
+		OAuth2AuthorizedClient authorizedClient = getOAuth2AuthorizedClient(attrs);
+		if (reauthorizeRequired(authorizedClient)) {
```
jzheaux: Would this logic of determining whether or not to refresh be better suited inside the authorization provider? Doing so would relieve zuul filters, gateway filters and the like from having to replicate it.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -157,12 +160,12 @@ public void destroy() throws Exception {
 	/**
 	 * Sets the {@link OAuth2AccessTokenResponseClient} to be used for getting an {@link OAuth2AuthorizedClient} for
 	 * client_credentials grant.
+	 * @deprecated Supply {@link #setClientAuthorizationProvider(OAuth2ClientAuthorizationProvider)} with a configured {@link OAuth2AccessTokenResponseClient} for the client_credentials grant.
 	 * @param clientCredentialsTokenResponseClient the client to use
 	 */
+	@Deprecated
 	public void setClientCredentialsTokenResponseClient(
 			OAuth2AccessTokenResponseClient<OAuth2ClientCredentialsGrantRequest> clientCredentialsTokenResponseClient) {
-		Assert.notNull(clientCredentialsTokenResponseClient, "clientCredentialsTokenResponseClient cannot be null");
```
jzheaux: Since this is a WIP, I'm assuming this method just isn't filled out, yet. But, I'll call out that this method should not silently ignore the user's intent, even if it is deprecated. Instead, it should adapt to the new way or it should throw an exception that indicates it's no longer supported.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2AuthorizationContext.java
```diff
@@ -0,0 +1,272 @@
...
+		return getAuthorizedClient() == null;
+	}
+
+	/**
+	 * Returns {@code true} if the client needs to be re-authorized, otherwise {@code false}.
+	 *
+	 * @return {@code true} if the client needs to be re-authorized, otherwise {@code false}.
+	 */
+	public boolean reauthorizationRequired() {
+		return getAuthorizedClient() != null;
```
jzheaux: Can you explain your reasoning a bit here? "Reauthorization is required if we have an authorized client" doesn't seem logical.

Actually, given the name, I'd expect logic like what `hasTokenExpired` has.

Would it make sense to either 1. actually check if reauthorization is indeed required or 2. rename the method to `reauthorizationRequested`? 

If it's not ideal to place the expiration logic here in the context, then I think that `reauthorizationRequested` aligns better with the application's intent when calling the `reauthorize` static method, i.e. the application doesn't actually care whether reauthorization truly is required, it wants it reauthorized either way.

I think it would be easy to make this mistake, say if I were building a Zuul filter, or similar:

```java
OAuth2AuthorizationContext.Builder contextBuilder = reauthorize(authorizedClient);
if (contextBuilder.reauthorizeRequired()) {
    // ... application assumes that this means the token is expired
}
```

rwinch: I second this sentiment. More particularly, I'm not sure that it makes sense to embed logic into a domain object.

I'm also curious what defines re-authorization. Does this indicate that authorization has expired, or is it more inclusive (i.e. if the client just wants to request a new access token).

I wonder if it makes more sense to try and update the APIs be explicit and align with the OAuth APIs. OAuth2AuthorizedClientProvider would have two methods to get an access token and another to refresh it.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshTokenOAuth2AuthorizedClientProvider.java
```diff
@@ -0,0 +1,141 @@
...
+		OAuth2AccessTokenResponse tokenResponse =
+				this.accessTokenResponseClient.getTokenResponse(refreshTokenGrantRequest);
+
+		OAuth2AuthorizedClient authorizedClient = new OAuth2AuthorizedClient(
+				context.getClientRegistration(),
+				context.getPrincipal().getName(),
+				tokenResponse.getAccessToken(),
+				tokenResponse.getRefreshToken());
+
+		this.authorizedClientRepository.saveAuthorizedClient(
```
rwinch: It is quite surprising to me that the provider will save the result but not try and load it first. I think the API should not save the authorization. I am even more convinced the API should not do save without trying to perform a lookup as I find this asymmetry confusing.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2AuthorizationContext.java
jgrandja: @jzheaux @rwinch Let me explain the meaning behind reauthorization. A client may need to be reauthorized in the following cases:

1. access token is expired and an `OAuth2AuthorizedClient.refreshToken` is available, e.g. `authorization_code` or `password` client. NOTE: `client_credentials` client does not receive a refresh token as per spec recommendation.
2. access token is expired and an `OAuth2AuthorizedClient.refreshToken` is NOT available, e.g. `client_credentials` client do not have refresh token so this is a renewal (not a refresh). Furthermore, this case may apply to a `authorization_code` or `password` client if a refresh token is not available depending on the client and/or provider configuration.

It's important to note that a refresh token request can only happen when there is a refresh token available and the `refresh_token` grant is performed. In all other cases, it's getting a new access token OR renewing an expired access token for reauthorization cases.

> Would it make sense to either 1. actually check if reauthorization is indeed required or 2. rename the method to reauthorizationRequested?

It doesn't make sense to contain the access token expiration logic in the context. The expiration logic should be contained within the http client as it is today with `ServletOAuth2AuthorizedClientExchangeFilterFunction`. This is where the user would configure the `accessTokenExpiresSkew` as well. The http client should determine if the access token is expired or not and build the `OAuth2AuthorizationContext` for authorization or reauthorization.

> 2. rename the method to reauthorizationRequested?

I think this name makes a lot of sense. Let me propose the new names based on this?

```java
requestAuthorization(ClientRegistration clientRegistration)
boolean authorizationRequested()

requestReauthorization(OAuth2AuthorizedClient authorizedClient)
boolean reauthorizationRequested()
```

Or possibly an alternative name for `reauthorization`:

```java
requestRenewal(OAuth2AuthorizedClient authorizedClient)
boolean renewalRequested()
```

@jzheaux @rwinch What are your thoughts on these names?

jgrandja: @rwinch 

> I wonder if it makes more sense to try and update the APIs be explicit and align with the OAuth APIs. OAuth2AuthorizedClientProvider would have two methods to get an access token and another to refresh it.

I don't think it makes sense to add 2 operations (e.g. `authorize` and `refresh`) to `OAuth2AuthorizedClientProvider`. The `refresh()` can only happen in the `refresh_token` case - `RefreshTokenOAuth2AuthorizedClientProvider`. Even if we renamed the operation to something more generic like `renew()`, not all `OAuth2AuthorizedClientProvider` would implement this operation (and vice-versa) so there would be `null` implementations. For example, `authorization_code` `OAuth2AuthorizedClientProvider` would only implement `authorize()` (not `renew()`) and `refresh_token` `OAuth2AuthorizedClientProvider` would implement `renew()` but never `authorize()`.

My preference would be to keep the one `authorize()` operation and it's up to the implementation to determine if the request is for authorization or reauthorization depending on what's in the context.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshTokenOAuth2AuthorizedClientProvider.java
jgrandja: > It is quite surprising to me that the provider will save the result but not try and load it first

The `OAuth2AuthorizedClient` is supplied via the context - why would it look it up again? I'm not sure I follow?

rwinch: That is what I'm trying to get at. It is strange to require the user to lookup the inputs via the repository API and the provider will save the result.

I think we should either have the API allow looking up the OAuth2AuthorizedClient or (likely preferred) not save the result.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2AuthorizationContext.java
jzheaux: Thanks for the explanation on not embedding the logic in the context object - I agree that it would be better to rename the methods.

I think what I'd do is:

```java
static Builder withAuthorizedClient(OAuth2AuthorizedClient authorizedClient)
static Builder withClientRegistration(ClientRegistration clientRegistration)

boolean authorizationRequested()
boolean reauthorizationRequested()
```

I worry that some semantics might be lost, but it's nice that it aligns with the team standard for other builders. The semantic distance could be nice, too, in case our definition of what it means to request reauthorization changes.

If that's not a concern, then I like:

```java
static Builder forReauthorization(OAuth2AuthorizedClient authorizedClient)
static Builder forAuthorization(ClientRegistration clientRegistration)
```

That would create `OAuth2AuthorizationContext.forReauthorization(...)`, which I think reads nicely.

I could buy `requestReauthorization` because of the symmetry with `reauthorizationRequested`, but I lean away from it since we are working with a builder. It seems to me that the only verb method we'd want for a builder would be `build()` itself.

jgrandja: I like `forReauthorization()` and `forAuthorization()` the best. I'm going to go with this for now for next review.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2AuthorizedClientProvider.java
```diff
@@ -0,0 +1,45 @@
...
+	/**
+	 * Attempt to authorize (or re-authorize) the {@link OAuth2AuthorizationContext#getClientRegistration() client} in the provided context.
+	 * Implementations must return {@code null} if authorization is not supported for the specified client,
+	 * e.g. the provider doesn't support the {@link ClientRegistration#getAuthorizationGrantType() authorization grant} type configured for the client.
+	 *
+	 * @param context the context that holds authorization-specific state for the client
+	 * @return the {@link OAuth2AuthorizedClient} or {@code null} if authorization is not supported for the specified client
+	 */
+	@Nullable
+	OAuth2AuthorizedClient authorize(OAuth2AuthorizationContext context);
```
rwinch: [I still think](https://github.com/spring-projects/spring-security/pull/6845#discussion_r294888403) we need to revisit this API.

It looks to me like there is a lot of duplicated code in `OAuth2AuthorizedClientArgumentResolver` and `ServletOAuth2AuthorizedClientExchangeFilterFunction` despite having the new `OAuth2AuthorizedClientProvider` abstraction. For example, OAuth2AuthorizedClientArgumentResolver [resolveArgument method](https://github.com/spring-projects/spring-security/blob/dd6c5b0cf592cdfc8e969635f92173393d31dca7/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/method/annotation/OAuth2AuthorizedClientArgumentResolver.java#L103-L141) looks very similar to [authorizeClient method](https://github.com/spring-projects/spring-security/blob/dd6c5b0cf592cdfc8e969635f92173393d31dca7/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java#L394-L416).

I expect that anyone wanting to do something similar for `HttpClient` or `RestTemplate` will have similar issues. I wonder if it makes more sense for the provider to be able to to do more (i.e. retrieve and save the `OAuth2AuthorizedClient`).

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizedClientManager.java
```diff
@@ -0,0 +1,81 @@
...
+	 * the {@link ClientRegistration#getAuthorizationGrantType() authorization grant} type configured for the client.
+	 *
+	 * @param clientRegistrationId the identifier for the client's registration
+	 * @param principal the {@code Principal} {@link Authentication} (to be) associated to the authorized client
+	 * @param request the {@code HttpServletRequest}
+	 * @param response the {@code HttpServletResponse}
+	 * @return the {@link OAuth2AuthorizedClient} or {@code null} if authorization is not supported for the specified client
+	 */
+	@Nullable
+	OAuth2AuthorizedClient authorize(String clientRegistrationId, Authentication principal,
```
rwinch: Did you consider having a request object here?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizedClientManager.java
```diff
@@ -0,0 +1,81 @@
...
+	 * e.g. a {@link OAuth2AuthorizedClient#getRefreshToken() refresh token} is not available OR
+	 * the {@link OAuth2AuthorizedClient#getAccessToken() access token} is not expired.
+	 *
+	 * @param authorizedClient the authorized client
+	 * @param principal the {@code Principal} {@link Authentication} associated to the authorized client
+	 * @param request the {@code HttpServletRequest}
+	 * @param response the {@code HttpServletResponse}
+	 * @return the re-authorized {@link OAuth2AuthorizedClient} or the provided {@code authorizedClient} if not re-authorized
+	 */
+	OAuth2AuthorizedClient reauthorize(OAuth2AuthorizedClient authorizedClient, Authentication principal,
```
rwinch: Is there a reason we need to support reauthorize separately? It seems that the argument resolver does not need to use reauthorize. I'm guessing that ServletOAuth2AuthorizedClientExchangeFilterFunction could be restructured so that it does not require reauthorize as well

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -119,16 +109,13 @@
 
 	private static final String REQUEST_CONTEXT_OPERATOR_KEY = RequestContextSubscriber.class.getName();
 
-	private Clock clock = Clock.systemUTC();
-
 	private Duration accessTokenExpiresSkew = Duration.ofMinutes(1);
 
 	private ClientRegistrationRepository clientRegistrationRepository;
 
 	private OAuth2AuthorizedClientRepository authorizedClientRepository;
```
rwinch: Are clientRegistrationRepository and authorizedClientRepository still necessary? I think that we should be able to rely on only OAuth2AuthorizedClientManager

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -119,16 +109,13 @@
...
 
-	private OAuth2AccessTokenResponseClient<OAuth2ClientCredentialsGrantRequest> clientCredentialsTokenResponseClient =
-			new DefaultClientCredentialsTokenResponseClient();
+	private OAuth2AuthorizedClientManager authorizedClientManager;
```
rwinch: We should consider adding a constructor that uses authorizedClientManager only

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/AuthorizationCodeOAuth2AuthorizedClientProvider.java
```diff
@@ -0,0 +1,57 @@
...
+ * An implementation of an {@link OAuth2AuthorizedClientProvider}
+ * for the {@link AuthorizationGrantType#AUTHORIZATION_CODE authorization_code} grant.
+ *
+ * @author Joe Grandja
+ * @since 5.2
+ * @see OAuth2AuthorizedClientProvider
+ */
+public final class AuthorizationCodeOAuth2AuthorizedClientProvider implements OAuth2AuthorizedClientProvider {
+
+	public AuthorizationCodeOAuth2AuthorizedClientProvider() {
```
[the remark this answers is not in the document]
jgrandja: I've been meaning to remove all the default constructors...there are a couple that were left behind.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/method/annotation/OAuth2AuthorizedClientArgumentResolver.java
```diff
@@ -149,37 +168,34 @@ private String resolveClientRegistrationId(MethodParameter parameter) {
...
 			OAuth2AccessTokenResponseClient<OAuth2ClientCredentialsGrantRequest> clientCredentialsTokenResponseClient) {
 		Assert.notNull(clientCredentialsTokenResponseClient, "clientCredentialsTokenResponseClient cannot be null");
-		this.clientCredentialsTokenResponseClient = clientCredentialsTokenResponseClient;
+		updateAuthorizedClientManager(clientCredentialsTokenResponseClient);
+	}
+
+	private void updateAuthorizedClientManager(
+			OAuth2AccessTokenResponseClient<OAuth2ClientCredentialsGrantRequest> clientCredentialsTokenResponseClient) {
+
+		if (this.authorizedClientManager instanceof DefaultOAuth2AuthorizedClientManager) {
```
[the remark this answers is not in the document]
jgrandja: good idea re: `defaultManager` member. I'll update

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2ReauthorizeRequest.java
```diff
@@ -0,0 +1,57 @@
...
+ * Represents a request the {@link OAuth2AuthorizedClientManager} uses to
+ * {@link OAuth2AuthorizedClientManager#reauthorize(OAuth2ReauthorizeRequest) re-authorize}
+ * the provided {@link OAuth2AuthorizedClient#getClientRegistration() client}.
+ *
+ * @author Joe Grandja
+ * @since 5.2
+ * @see OAuth2AuthorizeRequest
+ * @see OAuth2AuthorizedClientManager
+ */
+public class OAuth2ReauthorizeRequest extends OAuth2AuthorizeRequest {
```
rwinch: I don't think it makes sense for `OAuth2ReauthorizeRequest` to extend `OAuth2AuthorizeRequest`. What does it mean for `OAuth2ReauthorizeRequest` to be passed into `OAuth2AuthorizedClientManager.authorize`? The types should enforce correct usage.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/ClientCredentialsOAuth2AuthorizedClientProvider.java
```diff
@@ -0,0 +1,108 @@
...
+	 * @return the {@link OAuth2AuthorizedClient} or {@code null} if authorization (or re-authorization) is not supported
+	 */
+	@Override
+	@Nullable
+	public OAuth2AuthorizedClient authorize(OAuth2AuthorizationContext context) {
+		Assert.notNull(context, "context cannot be null");
+
+		ClientRegistration clientRegistration = context.getClientRegistration();
+		OAuth2AuthorizedClient authorizedClient = context.getAuthorizedClient();
+		if (!AuthorizationGrantType.CLIENT_CREDENTIALS.equals(clientRegistration.getAuthorizationGrantType()) ||
```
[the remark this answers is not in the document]
jgrandja: Ok I'll adjust

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2AuthorizationContext.java
```diff
@@ -0,0 +1,194 @@
...
+ *
+ * @author Joe Grandja
+ * @since 5.2
+ * @see OAuth2AuthorizedClientProvider
+ */
+public final class OAuth2AuthorizationContext {
+	private ClientRegistration clientRegistration;
+	private OAuth2AuthorizedClient authorizedClient;
+	private Authentication principal;
+	private Map<String, Object> attributes;
```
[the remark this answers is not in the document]
jgrandja: Ok, I'll remove from this PR and add `contextAttributesMapper` back in for `password` grant PR, as that's where it will be used initially

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2ReauthorizeRequest.java
jgrandja: > I don't think it makes sense for `OAuth2ReauthorizeRequest` to extend `OAuth2AuthorizeRequest`

This makes perfect sense to me. A request to re-authorize, (e.g. token is expired) is considered an authorization request as well. Therefore, a re-authorization request is an authorization request.

> What does it mean for `OAuth2ReauthorizeRequest` to be passed into `OAuth2AuthorizedClientManager.authorize()`

Why would this even happen? The API is clear `authorize(OAuth2AuthorizeRequest)` so even if a `OAuth2ReauthorizeRequest` was passed into `authorize()` it would only be used as an `OAuth2AuthorizeRequest` and **never cast** to an `OAuth2ReauthorizeRequest`.
Can you expand on the use case because as far as I see it the user would not be using the API correctly in this instance if they were to pass in an `OAuth2ReauthorizeRequest`, but the method implementation would be correctly implemented based on the clear definition of the operation and it's input.

Also, the added benefit of having the hierarchy `OAuth2ReauthorizeRequest` extends `OAuth2AuthorizeRequest` is the `DefaultOAuth2AuthorizedClientManager` will only need one `setContextAttributesMapper(Function<OAuth2AuthorizeRequest, Map<String, Object>> contextAttributesMapper)`. It would not be ideal to have a second `setContextAttributesMapper(Function<OAuth2ReauthorizeRequest, Map<String, Object>> contextAttributesMapper)` (typed `OAuth2ReauthorizeRequest`)

samples/boot/oauth2webclient/src/main/java/sample/config/WebClientConfig.java
```diff
@@ -32,8 +35,20 @@
 
 	@Bean
 	WebClient webClient(ClientRegistrationRepository clientRegistrationRepository, OAuth2AuthorizedClientRepository authorizedClientRepository) {
-		ServletOAuth2AuthorizedClientExchangeFilterFunction oauth2 = new ServletOAuth2AuthorizedClientExchangeFilterFunction(clientRegistrationRepository, authorizedClientRepository);
+		OAuth2AuthorizedClientProvider authorizedClientProvider =
```
jzheaux: This feels like a lot of code to ask an application to write. Is simplifying this a consideration for another PR?


A reasonable accommodation might be:

```java
OAuth2ClientAuthorizationManager manager = OAuth2ClientAuthorizationManager.withDefaults(clients, authz);
ServletOAuth2AuthorizedClientExchangeFilterFunction oauth2 =
        new ServletOAuth2AuthorizedClientExchangeFilterFunction(manager);
```

Or, if it's an option, leaving the `ServletOAuth2AuthorizedClientExchangeFilterFunction(ClientRegistrationRepository, OAuth2AuthorizedClientRepository)` constructor as un-deprecated would be another solution.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -137,11 +132,48 @@
...
+	/**
+	 * Constructs a {@code ServletOAuth2AuthorizedClientExchangeFilterFunction} using the provided parameters.
+	 *
+	 * @deprecated Use {@link #ServletOAuth2AuthorizedClientExchangeFilterFunction(OAuth2AuthorizedClientManager)} instead.
+	 * 				See {@link DefaultOAuth2AuthorizedClientManager} and {@link OAuth2AuthorizedClientProviderBuilder}.
+	 *
+	 * @param clientRegistrationRepository the repository of client registrations
+	 * @param authorizedClientRepository the repository of authorized clients
+	 */
+	@Deprecated
```
jzheaux: Why is this approach now deprecated?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2AuthorizationContext.java
```diff
@@ -0,0 +1,194 @@
...
+		return (T) this.getAttributes().get(name);
+	}
+
+	/**
+	 * Returns a new {@link Builder} initialized with the {@link ClientRegistration}.
+	 *
+	 * @param clientRegistration the {@link ClientRegistration client registration}
+	 * @return the {@link Builder}
+	 */
+	public static Builder forClient(ClientRegistration clientRegistration) {
```
jzheaux: Now that this method refers more directly to the builder's constructor parameters, `withClientRegistration` seems much more reasonable. (And `withOAuth2AuthorizedClient` as well). These also would fall in line with the overall naming convention in the codebase.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2AuthorizationContext.java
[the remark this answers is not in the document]
jgrandja: Just to clarify from my last comment, I will remove `DefaultOAuth2AuthorizedClientManager.contextAttributesMapper` but will not remove `OAuth2AuthorizationContext.attributes` as it's being used by `RefreshTokenOAuth2AuthorizedClientProvider`. You can see the usage in this [test ](https://github.com/spring-projects/spring-security/pull/6845/files#diff-83f70a51829e726be056afb992ffc279R162)

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/ClientCredentialsOAuth2AuthorizedClientProvider.java
```diff
@@ -0,0 +1,108 @@
...
+		OAuth2ClientCredentialsGrantRequest clientCredentialsGrantRequest =
+				new OAuth2ClientCredentialsGrantRequest(clientRegistration);
+		OAuth2AccessTokenResponse tokenResponse =
+				this.accessTokenResponseClient.getTokenResponse(clientCredentialsGrantRequest);
+
+		return new OAuth2AuthorizedClient(clientRegistration, context.getPrincipal().getName(), tokenResponse.getAccessToken());
+	}
+
+	private boolean hasTokenExpired(AbstractOAuth2Token token) {
+		return token.getExpiresAt().isBefore(Instant.now().minus(this.clockSkew));
```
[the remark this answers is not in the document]
jgrandja: The default is `Clock.systemUTC()`. Has there been any issues with this default? have other user's reported anything related to this?

Agreed, we can add this as a separate enhancement afterwards if need be.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2ReauthorizeRequest.java
rwinch: The API should enforce proper usage. It is surprising behavior that the API allows an
OAuth2AuthorizedClient to be provided to authorize via the OAuth2ReauthorizeRequest but it will be completely ignored.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2AuthorizedClientProviderBuilder.java
```diff
@@ -0,0 +1,267 @@
...
+
+	private OAuth2AuthorizedClientProviderBuilder() {
+	}
+
+	/**
+	 * Returns a new {@link OAuth2AuthorizedClientProviderBuilder} for configuring the supported authorization grant(s).
+	 *
+	 * @return the {@link OAuth2AuthorizedClientProviderBuilder}
+	 */
+	public static OAuth2AuthorizedClientProviderBuilder withProvider() {
```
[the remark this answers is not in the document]
jgrandja: I agree the name `OAuth2AuthorizedClientBuilder#builder()` is better and our convention.

> is this static factory in the right place

I externalized `OAuth2AuthorizedClientProviderBuilder` from the core API `OAuth2AuthorizedClientProvider` because `OAuth2AuthorizedClientProvider` should not have any dependencies or any knowledge on how to build the specific implementations, e.g. `AuthorizationCodeOAuth2AuthorizedClientProvider`, `RefreshTokenOAuth2AuthorizedClientProvider`, etc. Furthermore, user's may implement their own `OAuth2AuthorizedClientProvider` if they have custom grants which they would provide via `OAuth2AuthorizedClientProviderBuilder.withProvider().provider(OAuth2AuthorizedClientProvider)`.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2AuthorizedClientProviderBuilder.java
```diff
@@ -0,0 +1,267 @@
...
+ * @author Joe Grandja
+ * @since 5.2
+ * @see OAuth2AuthorizedClientProvider
+ * @see AuthorizationCodeOAuth2AuthorizedClientProvider
+ * @see RefreshTokenOAuth2AuthorizedClientProvider
+ * @see ClientCredentialsOAuth2AuthorizedClientProvider
+ * @see DelegatingOAuth2AuthorizedClientProvider
+ */
+public final class OAuth2AuthorizedClientProviderBuilder {
+	private final Map<Class<?>, Builder> builders = new HashMap<>();
```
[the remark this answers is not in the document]
jgrandja: The order doesn't matter given how the `OAuth2AuthorizedClientProvider` are implemented. Either way, I'll change it to `LinkedHashMap`

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2AuthorizedClientProviderBuilder.java
```diff
@@ -0,0 +1,267 @@
...
+
+	/**
+	 * Configures an {@link OAuth2AuthorizedClientProvider} to be composed with the {@link DelegatingOAuth2AuthorizedClientProvider}.
+	 * This may be used for implementations of extension authorization grants.
+	 *
+	 * @return the {@link OAuth2AuthorizedClientProviderBuilder}
+	 */
+	public OAuth2AuthorizedClientProviderBuilder provider(OAuth2AuthorizedClientProvider provider) {
+		Assert.notNull(provider, "provider cannot be null");
+		this.builders.computeIfAbsent(provider.getClass(), k -> () -> provider);
```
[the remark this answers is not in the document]
jgrandja: Do you think a user would do this?

```
OAuth2AuthorizedClientProvider authorizedClientProvider =
		OAuth2AuthorizedClientProviderBuilder.withProvider()
				.authorizationCode()
				.refreshToken()
				.clientCredentials()
				.clientCredentials()     // The previous (first) instance is returned
				.build();
```

If if they did, they are able to override the configuration of the first `ClientCredentialsGrantBuilder` so I don't see an issue here.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2AuthorizedClientProviderBuilder.java
```diff
@@ -0,0 +1,267 @@
...
+		return OAuth2AuthorizedClientProviderBuilder.this;
+	}
+
+	/**
+	 * Configures support for the {@code authorization_code} grant.
+	 *
+	 * @return the {@link OAuth2AuthorizedClientProviderBuilder}
+	 */
+	public OAuth2AuthorizedClientProviderBuilder authorizationCode() {
+		this.builders.computeIfAbsent(AuthorizationCodeGrantBuilder.class, k -> new AuthorizationCodeGrantBuilder());
```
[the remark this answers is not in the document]
jgrandja: Good catch! I'll make the provider class the key for all.

samples/boot/oauth2webclient/src/main/java/sample/config/WebClientConfig.java
jgrandja: In the first example, was that a typo `OAuth2ClientAuthorizationManager`? Did you mean `OAuth2AuthorizedClientManager`? Or is that a new class you are proposing?

For the 2nd example, `ServletOAuth2AuthorizedClientExchangeFilterFunction(ClientRegistrationRepository, OAuth2AuthorizedClientRepository)` will stay deprecated given that `OAuth2AuthorizedClientManager` is now the 1st class citizen for the `ExchangeFilterFunction`.

jgrandja: I'm going to change the `WebClientConfig` to be this instead.

```java
@Bean
WebClient webClient(OAuth2AuthorizedClientManager authorizedClientManager) {
	ServletOAuth2AuthorizedClientExchangeFilterFunction oauth2 =
			new ServletOAuth2AuthorizedClientExchangeFilterFunction(authorizedClientManager);
	oauth2.setDefaultOAuth2AuthorizedClient(true);
	return WebClient.builder()
			.apply(oauth2.oauth2Configuration())
			.build();
}

@Bean
OAuth2AuthorizedClientManager authorizedClientManager(ClientRegistrationRepository clientRegistrationRepository,
														OAuth2AuthorizedClientRepository authorizedClientRepository) {
	OAuth2AuthorizedClientProvider authorizedClientProvider =
			OAuth2AuthorizedClientProviderBuilder.withProvider()
					.authorizationCode()
					.refreshToken()
					.clientCredentials()
					.build();
	DefaultOAuth2AuthorizedClientManager authorizedClientManager = new DefaultOAuth2AuthorizedClientManager(
			clientRegistrationRepository, authorizedClientRepository);
	authorizedClientManager.setAuthorizedClientProvider(authorizedClientProvider);

	return authorizedClientManager;
}
```

This simplifies configuring a `WebClient` with the addition of a `OAuth2AuthorizedClientManager` `@Bean`. We can have this `@Bean` auto-configured by Boot but I'm not sure if this is the best option. Reason being is that most user's will not require all grant types to be configured. Thoughts?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
jgrandja: Because `OAuth2AuthorizedClientManager` is now the 1st class citizen for the `ExchangeFilterFunction`.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2AuthorizationContext.java
jgrandja: Makes sense. I'll change to `withClientRegistration` and `withAuthorizedClient` (leaving out `OAuth2`)

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2ReauthorizeRequest.java
jgrandja: I'd like to hear your answer regarding my question...

> Why would this even happen?...

Is this really a valid scenario? Like I already mentioned this would be a user error on the usage of the API. But the API would still behave as expected, ignoring the `OAuth2ReauthorizeRequest  ` attributes.

Furthermore, based on this understanding does this issue become invalid? See [comment](https://github.com/spring-projects/spring-security/issues/5633#issuecomment-410059242). Because based on this issue, an `OAuth2LoginAuthenticationToken` could get passed into an `OAuth2AuthorizationCodeAuthenticationProvider` which is identically scenario we are discussing here. @jzheaux Let's get your input here as I'm still not convinced.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2AuthorizedClientProviderBuilder.java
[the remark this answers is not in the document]
jzheaux: Agreed that there is little reason to call `clientCredentials` or other related methods twice.

What this reminds me of, though, is `JwtConfigurer`, which has both `jwtDecoder()` and `jwkSetUri()`. Whichever is called last is the one that takes precedence. In my mind:

`jwtDecoder` : `provider` and
`jwkSetUri` : `authorizationCode`

Hypothetically, if an API were to return an instance of `OAuth2AuthorizedClientProviderBuilder` (like `ClientRegistrations` returns a `ClientRegistraions.Builder`), I don't see how an application could override the configuration performed by that API.

For example, an application can do `ClientRegistrations.fromIssuerLocation(location).jwkSetUri(...)` to override what is provided by `ClientRegistrations` since `ClientRegistration.Builder` takes the last configuration as the correct one.

samples/boot/oauth2webclient/src/main/java/sample/config/WebClientConfig.java
jzheaux: Sorry, yes, that was a typo - good catch.

I agree that a Boot-provided instance will help out a great deal. I'd expect it to align with other related Boot defaults. For example, doesn't `oauth2Login` get configured by default for the `authorization_code` flow? I would think that the Boot-provided one would do the same, or at least base its decision on the same information.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/ClientCredentialsOAuth2AuthorizedClientProvider.java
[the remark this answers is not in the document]
jzheaux: Yes, there was [a user report](https://github.com/jzheaux/spring-security-oauth2-resource-server/issues/57), which is why `JwtTimestampValidator` has a `setClock` method.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2AuthorizedClientProviderBuilder.java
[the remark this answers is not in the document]
jgrandja: I see the examples you provided but just to re-clarify, if a user calls `clientCredentials()` a second (or more) time it will return the same instance each time:

1st time
`clientCredentials(configurer -> configurer.accessTokenResponseClient(accessTokenResponseClient))`

2nd time
`clientCredentials(configurer -> configurer.accessTokenResponseClient(accessTokenResponseClient))`

The 2nd call overrides the previous `accessTokenResponseClient` set so overriding works  here.

Am I missing something or not understanding you last comment?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
jgrandja: @rwinch I spoke to @jzheaux offline about this and he has a valid point as far as re-considering **not deprecating** this constructor. It can serve as a convenience constructor. Thoughts?

samples/boot/oauth2webclient/src/main/java/sample/config/WebClientConfig.java
jgrandja: > doesn't `oauth2Login` get configured by default for the authorization_code flow

Technically true, because `oauth2Login` does leverage `authorization_code` flow. But I think you meant `oauth2Client(()`, which by default configures  `authorization_code`.

I still feel that not all users will want to enable all grants by default in this scenario. Of course it wouldn't hurt and they can easily override the default. I'm just not sure if enabling `password` grant by default is a good idea. We'll have to give this some further thought as far as what the default grants will be.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2AuthorizationContext.java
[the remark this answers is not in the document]
jgrandja: @rwinch There indeed was a gap in my initial implementation of `DefaultOAuth2AuthorizedClientManager`. There now is a default implementation of `contextAttributesMapper` that provides the support for `RefreshTokenOAuth2AuthorizedClientProvider` and very possibly future implementations of `OAuth2AuthorizedClientProvider` (e.g. dynamic scopes). Please see this [test](https://github.com/spring-projects/spring-security/pull/6845/files#diff-8502d25303650e7a215cd70bb233634bR258) for usage.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/ClientCredentialsOAuth2AuthorizedClientProvider.java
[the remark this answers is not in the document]
jgrandja: I've added #7114

samples/boot/oauth2webclient/src/main/java/sample/config/WebClientConfig.java
jgrandja: Ticket #7115

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
rwinch: I saw the discussion in the comments and I'm ok with keeping the constructor around (not deprecating it).

jgrandja:
@rwinch @jzheaux I've pushed all the updates from last review. There are only 2 items that still need to be addressed:

- [Comment 1](https://github.com/spring-projects/spring-security/pull/6845#discussion_r304001531)
- [Comment 2](https://github.com/spring-projects/spring-security/pull/6845#discussion_r304348865)

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/OAuth2AuthorizedClientProviderBuilder.java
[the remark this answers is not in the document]
jzheaux: Okay, I see. Since the configurer method returns the same instance, overriding works. This was my main concern with doing first-wins.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2ReauthorizeRequest.java
rwinch: > Is this really a valid scenario?

That is exactly my point. The API should not allow this to happen, but it does.

> But the API would still behave as expected,

As a user I find this confusing. If I create an `OAuth2ReauthorizeRequest` providing the `OAuth2AuthorizedClient`, it is surprising to me that the `OAuth2AuthorizedClient` would be ignored.

jzheaux: Is there a need for the two types and the separate methods? Looking at the default implementations, there seems to be a large overlap. If both are indeed authorize requests, as the hierarchy implies, perhaps `OAuth2AuthorizeRequest` could optionally hold an `Oauth2AuthorizedClient`.

jgrandja: I personally like the clarity of the API for `OAuth2AuthorizedClientManager` having both `authorize()` and `reauthorize()`. However, I'm open to discussing the removal of `reauthorize()`. If we were to do that than we have this:


```java
public interface OAuth2AuthorizedClientManager {
	@Nullable
	OAuth2AuthorizedClient authorize(OAuth2AuthorizeRequest authorizeRequest);
}

public class OAuth2AuthorizeRequest {
	private final String clientRegistrationId;
	private final OAuth2AuthorizedClient authorizedClient;
	private final Authentication principal;
	private final HttpServletRequest servletRequest;
	private final HttpServletResponse servletResponse;

	public OAuth2AuthorizeRequest(String clientRegistrationId, Authentication principal,
									HttpServletRequest servletRequest, HttpServletResponse servletResponse) {

	}

	public OAuth2AuthorizeRequest(OAuth2AuthorizedClient authorizedClient, Authentication principal,
									HttpServletRequest servletRequest, HttpServletResponse servletResponse) {

	}
	

```

`OAuth2AuthorizeRequest` would provide 2 constructors and would be similar to the members in `OAuth2AuthorizationContext` which is passed downstream to the `OAuth2AuthorizedClientProvider`(s).

@rwinch What are your thoughts on this proposed design?

rwinch: I'm ok with this proposal as it solve my concern and aligns with our previous discussions (i.e. we can have two separate methods with two request types or the same method with a single request type).

jgrandja:
@rwinch @jzheaux I've addressed the remaining 2 items. Ready for review.

Issue #5543, opened 2018-07-19
Create OpenID Connect Discovery Client


jgrandja:
Create OpenID Connect Discovery Client which is capable of requesting [Provider Metadata](http://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata) via the supplied `issuer`.

Related #5523, #5540, #5350

**NOTE:** There is existing logic for OpenID Connect Discovery in `OidcConfigurationProvider`. We should extract this logic to allow for re-use in Client and Resource Server.

jgrandja:
We're going to hold off on this feature for 5.1. This [branch](https://github.com/jgrandja/spring-security/tree/gh-5543-discovery-client) contains this feature and we'll consider adding it in a later release.

pthorson:
Note: `OidcConfigurationProvider` moved to `ClientRegistrations` in https://github.com/spring-projects/spring-security/commit/742716fae3695317f4587ca41b91710ebce40a56

jgrandja:
Closing as this is not needed at the moment.

nevmerzhitsky:
Sorry, why did you not include the OpenID Connect Discovery Client to the library? There is an automated way to take `jwks_uri` key from the URLs like `oauth2/.well-known/openid-configuration` by the lib to pass it into `createRemoteJWKSet()`?

PR #6847, opened 2019-05-07
OAuth 2.0 Discovery Clients

jzheaux:
A support class for configuration that is common to the various
builders that use discovery.

The following is how you would configure ClientRegistrations to use
oauth-authorization-server discovery:

```java
ClientRegistrations.fromIssuerLocation("https://idp.example.org",
        OAuth2MetadataClientBuilder::oauth2Discovery);
```

Or to use OIDC Discovery as well as OAuth 2.0 Discovery do:

```java
ClientRegistrations.fromIssuerLocation("http://idp.example.org",
        client -> client.oidcDiscovery().oauth2Discovery());
```

### DIFF
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2MetadataClientBuilder.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2MetadataClientBuilder.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.oauth2.core;
+
+
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Function;
+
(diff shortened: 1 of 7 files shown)


oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2MetadataClientBuilder.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Function;
+
+import org.springframework.util.Assert;
+import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.util.UriComponentsBuilder;
+
+public class OAuth2MetadataClientBuilder {
```
jzheaux: I'm not convinced that this is the time to introduce this as a `public` class. However, I see no other way to share this with `ClientRegistrations`, `JwtDecoders`, and `ReactiveJwtDecoders`.

Some alternatives include:

* Making the constructor `protected`. I don't see this as a huge win since its still a public contract that folks simply then need to extend to use.
* Have two copies of this class. This seems odd, really, since there is nothing Resource Server-specific or Client Application-specific about this class. It would also mean copying this into other packages where we may want to do discovery down the road (e.g. introspection configuration)
* Narrow the class's contract down to only generating the URIs. The overlapping concerns of these three classes (`ClientRegistrations`, `JwtDecoders`, and `ReactiveJwtDecoders`) are the need to address a metadata endpoint using the issuer. This is the same way that Nimbus exposes this feature, and it is clearly the way that the spec outlines the contract - an issuer in exchange for metadata. While this class could more narrowly just calculate the URIs, it seems a very odd thing to expose since it seems an internal implementation detail to a metadata client (it wouldn't make sense to me to hand the fully-qualified discovery url to a discovery client since that discovery client would really prefer to calculate it itself).

For these reasons, I do lean towards keeping this class `public` with a `URI` -> `Map<String, Object>` contract. 

Also, I'd prefer the class be `final`, though really this PR is in draft form and has yet some cleanup to do. My hope is to begin conversation about what is an effective consolidation of the duplicated code in these three classes, and perhaps others down the road.

jgrandja:
Related #5543. Also see this [comment](https://github.com/spring-projects/spring-security/issues/5543#issuecomment-408189111) that points to a branch with similar implementation.

jzheaux:
Closing in favor of https://github.com/spring-projects/spring-security/pull/6765

Issue #6809, opened 2019-04-22
Confusing example - WebMvcConfigurer vs WebSecurityConfigurerAdapter


adamzimny:
### Summary
In Spring Security documentation in the hello-web example here:
   https://docs.spring.io/spring-security/site/docs/5.1.5.RELEASE/reference/htmlsingle/#hello-web-security-java-configuration
the code is:
```
@EnableWebSecurity
public class WebSecurityConfig implements WebMvcConfigurer {
``` 
This suggests that custom implementation of `WebMvcConfigurer` interface is the standard way of Spring Security configuration.

Further down in the documentation usage of `WebSecurityConfigurerAdapter` is mentioned and multiple examples of `void configure(HttpSecurity http)` customization are given. 

This raises confusion about the role of `WebMvcConfigurer` in Spring Security and the use cases for `WebMvcConfigurer` vs `WebSecurityConfigurerAdapter`.

Most likely the intention in the example was:
```
@EnableWebSecurity
public class WebSecurityConfig implements WebSecurityConfigurerAdapter {
```
instead of
```
@EnableWebSecurity
public class WebSecurityConfig implements WebMvcConfigurer {
```

### Version
5.1.5

### Sample

Not only I got confused, Stackoverflow discussing the same:
https://stackoverflow.com/questions/53894649/difference-between-webmvcconfigurer-and-websecurityconfigureradapter

jzheaux:
@adamzimny I agree with your analysis, thanks for the report. Would you be able to submit a PR with the needed change?

pthorson:
Can same be checked in for 5.1.x branch? Thank you.

jzheaux:
Actually, I'm thinking that it's not quite as simple as switching out the class reference.

The issue here is that the documentation states:

> You can find the most basic example of a Spring Security Java Configuration below:

And if we add the `extends WebSecurityConfigurerAdapter`, then that's actually a false statement as the simplest configuration requires no extends at all. I had originally missed this when I recommended this for a PR.

I agree with @rwinch [here](https://github.com/spring-projects/spring-security/pull/6850#issuecomment-494156453) that the best fix would be to remove the `implements WebMvcConfigurer` and update the way `WebSecurityConfigurerAdapter` is introduced in the documentation. This could likely be done by changing only a couple of sentences.

PR #6850, opened 2019-05-08
Remove Invalid WebMvcConfigurer from Sample Documentation

adamzimny:
Obvious Fix
Fixes gh-6809

### DIFF
--- a/docs/manual/src/docs/asciidoc/_includes/servlet/preface/java-configuration.adoc
+++ b/docs/manual/src/docs/asciidoc/_includes/servlet/preface/java-configuration.adoc
@@ -18,17 +18,16 @@ You can find the most basic example of a Spring Security Java Configuration belo
 [[jc-hello-wsca]]
 [source,java]
 ----
-import org.springframework.beans.factory.annotation.Autowired;
-
 import org.springframework.context.annotation.*;
-import org.springframework.security.config.annotation.authentication.builders.*;
 import org.springframework.security.config.annotation.web.configuration.*;
+import org.springframework.security.core.userdetails.*;
+import org.springframework.security.provisioning.InMemoryUserDetailsManager;
 
 @EnableWebSecurity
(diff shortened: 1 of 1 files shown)


rwinch:
Thanks for the PR! Since this is the hello world (minimal sample) we can remove `extends WebSecurityConfigurerAdapter` as well. Can you please update the PR to do that?

adamzimny:
Hi Rob,

you are right that `extends WebSecurityConfigurerAdapter` is not needed in the example itself. But if you read on and get to chapter 6.3 (and 6.4 and 6.5) then these chapters refer to the hello world example and explain why we extend from WebSecurityConfigurerAdapter and how it's functionality can be used.
This is actually how I came across the issue, I copy/pasted the hello world code snippet to my application, then continued reading and then at chapter 6.3 things no longer made much sense.

rwinch:
I think that along with that we should update the documentation to explain that when extending `WebSecurityConfigurerAdapter` is needed too (i.e. when we need to modify the default behavior of web based security).

adamzimny:
This is explained in chapters 6.2 - 6.5. You may be right that this documentation fragment is not perfect and it could use some refactoring. But the intention of this patch is just to fix an obvious bug - wrong class name used in the example.

pivotal-issuemaster:
@adamzimny Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=6850)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=6850) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@adamzimny Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=6850)!

jzheaux:
Closed in favor of 362356d

PR #6852, opened 2019-05-09
Make scheduler configurable on ReactiveAuthenticationManagerAdapter

ttddyy:
Currently, authentication logic will be performed on hardcoded elastic scheduler in `ReactiveAuthenticationManagerAdapter`.
This needs to be configurable in order to perform the authentication logic in dedicated scheduler.

Also, once authentication is performed, rest of the logic(Controller/Service) also be performed on the published scheduler.
To allow the rest of the logic after authentication to run on different scheduler, adding another optional scheduler to publish on.

This PR is based on the usecase I have described here: https://github.com/reactor/reactor-core/pull/1644#issuecomment-490701518

### DIFF
--- a/core/src/main/java/org/springframework/security/authentication/ReactiveAuthenticationManagerAdapter.java
+++ b/core/src/main/java/org/springframework/security/authentication/ReactiveAuthenticationManagerAdapter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,6 +18,7 @@ package org.springframework.security.authentication;
 import org.springframework.security.core.Authentication;
 import org.springframework.util.Assert;
 import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Scheduler;
 import reactor.core.scheduler.Schedulers;
 
 /**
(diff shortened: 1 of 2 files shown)


rwinch:
Thanks for the PR @ttddyy!

Can you please help me better understand what problems you are trying to solve?

ttddyy:
Hi @rwinch 

The change has two aspect.

First one is that we want to run the auth logic on controlled executor rather than the elastic one. Currently, it is hardcoded, so I want to make it changeable.

The second point is after auth process has finished, we would like to switch the executor. Currently, since auth logic has switched to the elastic executor, subsequent logic (such as performing controller/service) also continue to be executed on the switched elastic executor.
In my case, after performing auth on the controlled executor, now we want to switch to another executor that is managed differently to perform application logic.
(Switching back may be debatable action to do, according to the discussion on https://github.com/reactor/reactor-core/pull/1644, but having such capability is useful IMO)

rwinch:
Ok I understand the what, but why are you needing to do this? Specifically, why do you need to switch back? What are you switching back to?

ttddyy:
I would like to publish to another executor that executes application logic. The main reason is to have metrics around the executors that perform auth logic and app logic *separately*. This way we can get independent metrics for auth and make decision on converting the auth part to full reactive or just use adapter approach.

rwinch:
Can't you specify the executor for the logic you are executing? Why does this need to be part of Spring Security's code?

ttddyy:
OK, yeah executor for application can be specified outside of spring-security such as another webfilter or so. I'll update the PR to only focus on making auth-manager scheduler modifiable.

core/src/main/java/org/springframework/security/authentication/ReactiveAuthenticationManagerAdapter.java
```diff
@@ -50,4 +54,13 @@ public ReactiveAuthenticationManagerAdapter(AuthenticationManager authentication
 			})
 			.filter( a -> a.isAuthenticated());
 	}
+
+	/**
+	 * Set a scheduler that will be published on to perform the authentication logic.
+	 * @param scheduler a scheduler to be published on
+	 */
+	public void setScheduler(Scheduler scheduler) {
```
rwinch: Please add a null check and a test to validate the null check

ttddyy: Updated the PR with null check and test

ttddyy:
Hi @rwinch
This PR is ready, can you care to merge?
Thanks,

rwinch:
Thanks for the PR and the ping @ttddyy! This is now merged into master

Issue #6818, opened 2019-04-26
Switch to proxy-less configuration by leveraging @Configuration(proxyBeanMethods = false)


sdeleuze:
As part of our Spring efficiency effort, we have recently migrated almost all `@Confguration` in Spring Framework 5.2 and Spring Boot 2.2 to be usable without CGLIB proxies. It would be great for end users and consistent from a portfolio perspective if Spring Security 5.2 could do the same.

[This Spring Framework commit](https://github.com/spring-projects/spring-framework/commit/1a8b3fba9421461869d65318470b3432b62c1f9b) introduced a new `proxyBeanMethods` boolean attribute on `@Configuration` that when set to `false` avoid to create a CGLIB proxy for configuration classes. To be able to use this new capability you "just" have to replace eventual `@Bean` method invocations that where relying on CGLIB by explicit method parameter that will be injected without proxy.

@bclozel did that work on Spring MVC and Spring WebFlux configuration via [this commit](https://github.com/spring-projects/spring-framework/commit/47c8d1de7bde543089792cb9517ec540993b113c). You can also take inspiration from [Spring Boot autoconfigurations](https://github.com/spring-projects/spring-boot/tree/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure)  which have been almost all updated with `@Configuration(proxyBeanMethods = false)`. As a bonus, this would make Spring Security configurations GraalVM native image friendly for free ;-)

A possible process is to manually look through each `@Configuration` annotated class, do a "find usage" on each `@Bean` annotated method, replace invocation by parameter injection, and turn each `@Configuration` to `@Configuration(proxyBeanMethods = false)`. Notice that if you need more flexibility with the bean retrieved, you can leverage `ObjectProvider<Foo>` parameters [like that](https://github.com/spring-projects/spring-boot/blob/36c1c051b8106371d37f8a747cea0e369a8e5c84/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cache/CaffeineCacheConfiguration.java#L50).

Side notes: I found https://github.com/spring-projects/spring-security/issues/6613 on a connected topic, but targeting Spring Security 6.0 which will be too late for what is discussed here. Notice also that we took the decision to do breaking changes to Spring MVC and WebFlux configuration classes given the fact this topic is high priority for the Spring Boot 2.2 release train.

rwinch:
[...]

PR #6853, opened 2019-05-09
Replace bean method calls with injection

eleftherias:
This is so that our configuration classes do not rely on CGLIB to proxy bean methods.

Fixes gh-6818

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/AuthenticationConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/AuthenticationConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -107,8 +107,7 @@ public class AuthenticationConfiguration {
 		if (this.authenticationManagerInitialized) {
 			return this.authenticationManager;
 		}
-		AuthenticationManagerBuilder authBuilder = authenticationManagerBuilder(
-				this.objectPostProcessor, this.applicationContext);
+		AuthenticationManagerBuilder authBuilder = this.applicationContext.getBean(AuthenticationManagerBuilder.class);
 		if (this.buildingAuthenticationManager.getAndSet(true)) {
 			return new AuthenticationManagerDelegator(authBuilder);
 		}
--- a/config/src/main/java/org/springframework/security/config/annotation/method/configuration/GlobalMethodSecurityConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/method/configuration/GlobalMethodSecurityConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -29,10 +29,7 @@ import org.springframework.beans.factory.BeanFactoryAware;
 import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.beans.factory.SmartInitializingSingleton;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.context.annotation.AdviceMode;
-import org.springframework.context.annotation.Bean;
-import org.springframework.context.annotation.Configuration;
-import org.springframework.context.annotation.ImportAware;
+import org.springframework.context.annotation.*;
 import org.springframework.core.annotation.AnnotationAttributes;
 import org.springframework.core.annotation.AnnotationUtils;
 import org.springframework.core.type.AnnotationMetadata;
@@ -110,7 +107,6 @@ public class GlobalMethodSecurityConfiguration
 	 * <li>{@link #accessDecisionManager()}</li>
 	 * <li>{@link #afterInvocationManager()}</li>
 	 * <li>{@link #authenticationManager()}</li>
-	 * <li>{@link #methodSecurityMetadataSource()}</li>
 	 * <li>{@link #runAsManager()}</li>
 	 *
 	 * </ul>
@@ -119,19 +115,19 @@ public class GlobalMethodSecurityConfiguration
 	 * Subclasses can override this method to provide a different
 	 * {@link MethodInterceptor}.
 	 * </p>
+	 * @param methodSecurityMetadataSource the default {@link MethodSecurityMetadataSource}.
 	 *
 	 * @return the {@link MethodInterceptor}.
-	 * @throws Exception
 	 */
 	@Bean
-	public MethodInterceptor methodSecurityInterceptor() throws Exception {
+	public MethodInterceptor methodSecurityInterceptor(MethodSecurityMetadataSource methodSecurityMetadataSource) {
 		this.methodSecurityInterceptor = isAspectJ()
 				? new AspectJMethodSecurityInterceptor()
 				: new MethodSecurityInterceptor();
 		methodSecurityInterceptor.setAccessDecisionManager(accessDecisionManager());
 		methodSecurityInterceptor.setAfterInvocationManager(afterInvocationManager());
 		methodSecurityInterceptor
-				.setSecurityMetadataSource(methodSecurityMetadataSource());
+				.setSecurityMetadataSource(methodSecurityMetadataSource);
 		RunAsManager runAsManager = runAsManager();
 		if (runAsManager != null) {
 			methodSecurityInterceptor.setRunAsManager(runAsManager);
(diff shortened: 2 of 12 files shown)


config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/AuthenticationConfiguration.java
```diff
@@ -103,12 +103,10 @@ public static InitializeAuthenticationProviderBeanManagerConfigurer initializeAu
 		return new InitializeAuthenticationProviderBeanManagerConfigurer(context);
 	}
 
-	public AuthenticationManager getAuthenticationManager() throws Exception {
+	public AuthenticationManager getAuthenticationManager(AuthenticationManagerBuilder authBuilder) throws Exception {
```
jgrandja: This is a major API change so it could only be allowed in Spring Security 6.0. Is this issue scheduled for 6.0? For more info around versioning see [Semantic Versioning](https://semver.org/spec/v2.0.0.html)

config/src/main/java/org/springframework/security/config/annotation/method/configuration/GlobalMethodSecurityConfiguration.java
```diff
@@ -119,19 +115,19 @@
...
 	 * {@link MethodInterceptor}.
 	 * </p>
+	 * @param methodSecurityMetadataSource the default {@link MethodSecurityMetadataSource}.
 	 *
 	 * @return the {@link MethodInterceptor}.
-	 * @throws Exception
 	 */
 	@Bean
-	public MethodInterceptor methodSecurityInterceptor() throws Exception {
+	public MethodInterceptor methodSecurityInterceptor(MethodSecurityMetadataSource methodSecurityMetadataSource) {
```
jgrandja: Same reason as above re: Semantic Versioning

config/src/main/java/org/springframework/security/config/annotation/web/configuration/WebSecurityConfiguration.java
```diff
@@ -64,15 +62,13 @@
  * @since 3.2
  */
 @Configuration
-public class WebSecurityConfiguration implements ImportAware, BeanClassLoaderAware {
```
jgrandja: Same reason as above re: Semantic Versioning

eleftherias: This change isn't neccessary, I did not consider the configurations that are extending this one. Will revert it back to implementing `BeanClassLoaderAware`.

config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/AuthenticationConfiguration.java
rwinch: I struggle with this a little bit because we want to guarantee that the `AuthenticationManagerBuilder` is actually a `Bean`. Perhaps we could change to do this:

```java
AuthenticationManagerBuilder authBuilder = this.applicationContext.getBean(AuthenticationManagerBuilder.class);
```

config/src/main/java/org/springframework/security/config/annotation/web/configuration/WebSecurityConfiguration.java
```diff
@@ -123,16 +120,19 @@ public WebInvocationPrivilegeEvaluator privilegeEvaluator() throws Exception {
...
+	 * {@link AutowiredWebSecurityConfigurersIgnoreParents} containing the
 	 * {@code <SecurityConfigurer<FilterChainProxy, WebSecurityBuilder>} instances used to
 	 * create the web configuration
 	 * @throws Exception
 	 */
 	@Autowired(required = false)
 	public void setFilterChainProxySecurityConfigurer(
 			ObjectPostProcessor<Object> objectPostProcessor,
-			@Value("#{@autowiredWebSecurityConfigurersIgnoreParents.getWebSecurityConfigurers()}") List<SecurityConfigurer<Filter, WebSecurity>> webSecurityConfigurers)
+			AutowiredWebSecurityConfigurersIgnoreParents autowiredWebSecurityConfigurersIgnoreParents)
```
rwinch: Can you explain why this change is necessary? A concern I have is that setting `List<SecurityConfigurer<Filter, WebSecurity>>` is agnostic to have we got it and only the annotation instructs how to look up the value. Another concern is that `AutowiredWebSecurityConfigurersIgnoreParents` is package private.

config/src/main/java/org/springframework/security/config/annotation/web/configuration/WebSecurityConfiguration.java
```diff
@@ -71,8 +70,6 @@
 
 	private List<SecurityConfigurer<Filter, WebSecurity>> webSecurityConfigurers;
 
-	private ClassLoader beanClassLoader;
```
rwinch: I see that we don't need beanClassLoader, but since it is unrelated can we please put it in another PR?

config/src/main/java/org/springframework/security/config/annotation/web/reactive/ServerHttpSecurityConfiguration.java
```diff
@@ -43,7 +43,7 @@
  * @since 5.0
  */
 @Configuration
-class ServerHttpSecurityConfiguration implements WebFluxConfigurer {
+class ServerHttpSecurityConfiguration {
```
rwinch: Can you explain why we need to expose this via a Bean vs the Configuration? Does setting proxy proxyBeanMethods mean the `ServerHttpSecurityConfiguration` is no longer registered as a Bean?

config/src/main/java/org/springframework/security/config/annotation/web/reactive/ServerHttpSecurityConfiguration.java
```diff
@@ -113,6 +119,10 @@ public CurrentSecurityContextArgumentResolver reactiveCurrentSecurityContextArgu
 
 	@Bean(HTTPSECURITY_BEAN_NAME)
 	@Scope("prototype")
+	public ServerHttpSecurity serverHttpSecurity() {
```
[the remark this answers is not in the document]
eleftherias: There is a test that uses the `httpSecurity()` method. 
https://github.com/spring-projects/spring-security/blob/047bd16b5156de0bb595ff6b324a63cef92b1fe1/config/src/test/java/org/springframework/security/config/annotation/web/reactive/ServerHttpSecurityConfigurationBuilder.java#L31
If we separate the bean method from the method that creates the object then we avoid calling the bean method in the test.
Unless we aren't concerned about the test using the bean method because it's creating a new instance of `ServerHttpSecurityConfiguration`?

config/src/main/java/org/springframework/security/config/annotation/web/reactive/ServerHttpSecurityConfiguration.java
eleftherias: `ServerHttpSecurityConfiguration` will still be registered as a Bean. 
The reason for this change is to avoid calling `authenticationPrincipalArgumentResolver()` from the `configureArgumentResolvers` override. 
I noticed spring boot had a similar situation when disabling proxying and this was their approach 
https://github.com/spring-projects/spring-boot/blob/dfb2e8a278e7211c658104ae45a19b68aec57a21/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/neo4j/Neo4jDataAutoConfiguration.java#L136

config/src/main/java/org/springframework/security/config/annotation/web/reactive/ServerHttpSecurityConfiguration.java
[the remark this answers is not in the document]
rwinch: When it just performs new it won't create a Bean since nothing is processing the annotation.

config/src/main/java/org/springframework/security/config/annotation/web/reactive/ServerHttpSecurityConfiguration.java
rwinch: Ok thanks for the clarification. I think that this makes sense.

config/src/main/java/org/springframework/security/config/annotation/web/configuration/WebSecurityConfiguration.java
eleftherias: This change is not neccessary. I have reverted it.

config/src/main/java/org/springframework/security/config/annotation/method/configuration/GlobalMethodSecurityConfiguration.java
rwinch: I don't think this is avoidable. Given the high priority of this and the low likely hood of causing users any problems, I think this breaking change is ok.

eleftherias:
@rwinch Thanks for pointing that out. I limited my search to `@Configuration`, didn't think about classes that could be using other annotations!

config/src/test/java/org/springframework/security/config/annotation/authentication/configuration/AuthenticationConfigurationTests.java
```diff
@@ -172,51 +171,49 @@ public AuthenticationManager authenticationManager() {
 			return this.authenticationManager;
 		}
 	}
-		//
```
rwinch: There seem to be some formatting changes in the code within this commit. While overall, they are a good thing I think we should move it to another PR or at minimum a separate commit. The reason is it helps us know the intent of the changes when being reviewed or when looking through commit history if we are trying to figure out a bug at some point.

rwinch:

PR #6855, opened 2019-05-10
Translate messages.properties into Japanese

shimamchan:
Add error message translation of Japanese


rwinch:
Thanks for the PR @shimamchan! This is now merged into master

PR #6887, opened 2019-05-17
Simplify various redundant branches

OrangeDog:
Obvious Fix
There may be more code like this lying around, but I haven't gone looking for it.


OrangeDog:
@rwinch really? In security code especially it's important to be as clear as possible to reduce the places for bugs to hide.

Here you've code with two possible outcomes, expressed with five separate branches (not even an else) and three returns. Code style guidelines are all well and good, but if they result in nonsense like this then they should be ignored.

rwinch:
@OrangeDog The case that an if block with multiple conditions is more readable is not wrong, but it isn't right either. I find it more difficult to read because it is doing 3 things on a single line (evaluating the first expression, evaluating the second expression, evaluating the if statement with the two evaluated expressions).

eleftherias:
@OrangeDog are you able to update the PR to simplify the code based on the previous comment
> you can simplify by just removing https://github.com/spring-projects/spring-security/blob/9b77da08a68b94b6930dbdc158c99a154c058a61~1/config/src/main/java/org/springframework/security/config/annotation/web/configurers/SessionManagementConfigurer.java#L502-L508

PR #6893, opened 2019-05-21
Add constructor to JwtAuthenticationToken that takes a principal name

### LINKED ISSUE #6865: Add constructor to JwtAuthenticationToken that takes a principal name
`OAuth2IntrospectionAuthenticationToken` takes a name as a constructor parameter and otherwise defaults to the token's SUB claim for the principal name.

`JwtAuthenticationToken`, on the other hand, provides no way to customize this.

It'd be nice to do the same for `JwtAuthenticationToken`:

```java
public JwtAuthenticationToken(Jwt jwt, Collection<? extends GrantedAuthority> authorities, String name) {
    // ...
```

This encourages composition over inheritance while also keeping the application from having to create a custom class to override this default behavior.

HaydenMeloche:
Closes #6865 
I added two constructors to allow a custom principal name to be passed to the class. Otherwise the constructor will just set the name value to be token's subject.

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationToken.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationToken.java
@@ -36,6 +36,8 @@ import org.springframework.security.oauth2.jwt.Jwt;
 public class JwtAuthenticationToken extends AbstractOAuth2TokenAuthenticationToken<Jwt> {
 	private static final long serialVersionUID = SpringSecurityCoreVersion.SERIAL_VERSION_UID;
 
+	private final String name;
+
 	/**
 	 * Constructs a {@code JwtAuthenticationToken} using the provided parameters.
 	 *
@@ -43,6 +45,7 @@ public class JwtAuthenticationToken extends AbstractOAuth2TokenAuthenticationTok
 	 */
 	public JwtAuthenticationToken(Jwt jwt) {
 		super(jwt);
+		this.name = jwt.getSubject();
 	}
 
 	/**
@@ -54,6 +57,20 @@ public class JwtAuthenticationToken extends AbstractOAuth2TokenAuthenticationTok
 	public JwtAuthenticationToken(Jwt jwt, Collection<? extends GrantedAuthority> authorities) {
 		super(jwt, authorities);
 		this.setAuthenticated(true);
+		this.name = jwt.getSubject();
+	}
+
+	/**
+	 * Constructs a {@code JwtAuthenticationToken} using the provided parameters.
+	 *
+	 * @param jwt the JWT
+	 * @param authorities the authorities assigned to the JWT
+	 * @param name the principal name
+	 */
+	public JwtAuthenticationToken(Jwt jwt, Collection<? extends GrantedAuthority> authorities, String name) {
+		super(jwt, authorities);
+		this.setAuthenticated(true);
+		this.name = name;
 	}
 
 	/**
@@ -69,6 +86,6 @@ public class JwtAuthenticationToken extends AbstractOAuth2TokenAuthenticationTok
 	 */
 	@Override
 	public String getName() {
-		return this.getToken().getSubject();
+		return this.name;
 	}
 }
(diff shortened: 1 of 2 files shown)


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationToken.java
```diff
@@ -36,13 +36,27 @@
...
+		this.name = this.getToken().getSubject();
+	}
+
+	/**
+	 * Constructs a {@code JwtAuthenticationToken} using the provided parameters.
+	 *
+	 * @param jwt  the JWT
+	 * @param name principal name
+	 */
+	public JwtAuthenticationToken(Jwt jwt, String name) {
```
jzheaux: I think we can live without this constructor for now. We want to be careful to not get too many constructors--
we'd probably eventually introduce a builder instead of this constructor.

Just adding `(Jwt, Collection, String)` should be fine for this PR, unless you have a concern with leaving this constructor out.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationToken.java
```diff
@@ -36,13 +36,27 @@
 public class JwtAuthenticationToken extends AbstractOAuth2TokenAuthenticationToken<Jwt> {
 	private static final long serialVersionUID = SpringSecurityCoreVersion.SERIAL_VERSION_UID;
 
+	private String name;
```
jzheaux: Please make this `final` since this member doesn't need to change.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationToken.java
```diff
@@ -56,6 +70,19 @@ public JwtAuthenticationToken(Jwt jwt, Collection<? extends GrantedAuthority> au
 		this.setAuthenticated(true);
 	}
 
+	/**
+	 * Sub-class constructor.
```
jzheaux: Thanks for adding a JavaDoc. Would you please have it match a bit closer to the other constructor JavaDoc?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationToken.java
```diff
@@ -36,13 +36,27 @@
...
+	private String name;
+
 	/**
 	 * Constructs a {@code JwtAuthenticationToken} using the provided parameters.
 	 *
 	 * @param jwt the JWT
 	 */
 	public JwtAuthenticationToken(Jwt jwt) {
 		super(jwt);
+		this.name = this.getToken().getSubject();
```
jzheaux: This make sense, since `getToken()` is `final`, but what would you think of just doing:

```suggestion
		this.name = jwt.getSubject();
```

So that readers of this code have fewer methods to jump through to understand the code.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationToken.java
```diff
@@ -56,6 +70,19 @@ public JwtAuthenticationToken(Jwt jwt, Collection<? extends GrantedAuthority> au
 		this.setAuthenticated(true);
 	}
 
+	/**
```
jzheaux: Why did `(Jwt, Collection)` not get changed? Seems like we'd need to set the name there, too.

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationTokenTests.java
```diff
@@ -98,6 +107,19 @@ public void constructorWhenUsingOnlyJwtThenConstructedCorrectly() {
 		assertThat(token.isAuthenticated()).isFalse();
 	}
 
+	@Test
```
jzheaux: Would you please add a test that shows that when using the `(Jwt, Collection)` constructor (this is the one most folks are using in 5.1) continues to work correctly, e.g. `getName()` shows in that case whatever `jwt.getSubject()` returns?

HaydenMeloche:
@jzheaux Thanks for your review, I believe I made all the changes you requested. Can you take another look? 😄

jzheaux:
Thanks for the PR, @HaydenMeloche! This is now merged into `master` via f84ab3a2558d5258d99e169725c87c9ee37cd020. Also, I added a polish commit of 936d28d328e505c75ec028e24ba5bcdea40e69f8.

PR #6898, opened 2019-05-23
Support URI vars in formLogin and logout MockMvc requests

drumonii:
Allows URI variables in `formLogin` and `logout` requests of `MockMvc`

login example:
```
mockMvc.perform(formLogin("{basePath}/admin/login", "/api")
        .user("username")
        .password("password"))
        .andExpect(authenticated()
             ...
        .andExpect(status().isOk())
        .andExpect(...);
```
yields `/api/admin/login` as the `formLogin` URL.

logout example:
```
mockMvc.perform(logout("{basePath}/admin/logout", "/api"))
        .andExpect(unauthenticated()
        .andExpect(status().isOk())
        .andExpect(...);
```
yields `/api/admin/logout` as the `logout` URL.

### DIFF
--- a/test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuildersFormLoginTests.java
+++ b/test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuildersFormLoginTests.java
@@ -66,6 +66,22 @@ public class SecurityMockMvcRequestBuildersFormLoginTests {
 		assertThat(request.getRequestURI()).isEqualTo("/login");
 	}
 
+	@Test
+	public void customWithUriVars() {
+		MockHttpServletRequest request = formLogin().loginProcessingUrl("/uri-login/{var1}/{var2}", "val1", "val2")
+				.user("username", "admin").password("password", "secret").buildRequest(this.servletContext);
+
+		CsrfToken token = (CsrfToken) request
+				.getAttribute(CsrfRequestPostProcessor.TestCsrfTokenRepository.TOKEN_ATTR_NAME);
+
+		assertThat(request.getParameter("username")).isEqualTo("admin");
+		assertThat(request.getParameter("password")).isEqualTo("secret");
+		assertThat(request.getMethod()).isEqualTo("POST");
+		assertThat(request.getParameter(token.getParameterName()))
+				.isEqualTo(token.getToken());
+		assertThat(request.getRequestURI()).isEqualTo("/uri-login/val1/val2");
+	}
+
 	// gh-3920
 	@Test
 	public void usesAcceptMediaForContentNegotiation() throws Exception {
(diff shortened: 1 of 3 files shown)


eleftherias:
Thanks for the PR @drumonii.
I am concerned that this feature will cause confusion that outweighs the benefit of it.
There are many cases that we would need to consider that result in an error.
For example:
- if a user doesn't add the correct number of parameters `formLogin("{basePath}/admin/login/{other}", "/api")`, 
- or if they don't add a placeholder in the login URL `formLogin("user/admin/login", "/api")`

Additionally, it is not obvious how the URI variables will be substituted into the login URL.
When using something like `org.springframework.web.util.UriTemplate`, you are able to specify the variables by name using a `Map`, which makes it obvious what substitution will occur.

I would suggest that users in need of templating use the spring `UriTemplate`, like this:
```
UriTemplate template = new UriTemplate("{basePath}/admin/login");
Map<String, String> uriVariables = new HashMap<String, String>();
uriVariables.put("basePath", "/api");
mockMvc.perform(formLogin(template.expand(uriVariables))
```

drumonii:
Thanks for the feedback @eleftherias

This is essentially a convenience method that is modeled after `mockMvc`'s `perform` actions such as `get`, `post`, etc. that all take `Object... uriVars` as the last method parameter. Since `formLogin` and `logout` are `mockMvc` `perform` actions, I would have expected them to act similar. So I thought it would be appropriate for them reciprocate with `Object... uriVars` in their method signatures. I have tried to mimick the uriVars `MockHttpServletRequestBuilder` logic for URI variables as part of `formLogin` and `logout`.

So I would presume those the existing `mockMvc` `perform` actions have the same issues as you pointed. But to answer your concerns: 

- given the  incorrect number of parameters, an IAE is thrown
- given there aren't placeholders but parameters passed, then the URL specified is used as is with no exception thrown

eleftherias:
Thanks for the quick response @drumonii 

Perhaps a better place for the URI variables is in the method `loginProcessingUrl` for `formLogin` and `logoutUrl` for `logout`. 
Then adding a template and variables would look like this
```
mockMvc.perform(formLogin().loginProcessingUrl("{basePath}/admin/login", "/api"))
...
```

This makes it clearer that the URI variables supplied are used to construct the login URL, and not something else.

drumonii:
@eleftherias sounds reasonable to me! I've force pushed those changes for `formLogin` and `logout` with their respective builders.

test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuildersFormLogoutTests.java
```diff
@@ -58,4 +58,17 @@ public void custom() throws Exception {
 		assertThat(request.getRequestURI()).isEqualTo("/admin/logout");
 	}
 
+	@Test
+	public void customWithUriVars() throws Exception {
```
eleftherias: This `throws` clause is redundant.

test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuildersFormLoginTests.java
```diff
@@ -66,6 +66,22 @@ public void custom() throws Exception {
 		assertThat(request.getRequestURI()).isEqualTo("/login");
 	}
 
+	@Test
+	public void customWithUriVars() throws Exception {
```
eleftherias: This `throws` clause is redundant.

drumonii:
@eleftherias done! I too noticed the `throws Exception` was unnecessary but left it since the other tests had it declared despite being never thrown.

eleftherias:
Thanks for the PR @drumonii! This is now merged into master.

PR #6900, opened 2019-05-23
Finer variables for OAuth2 redirectUriTemplate expansion

### LINKED ISSUE #6239: Add scheme/protocol variable for OAuth2 redirectUriTemplate
Currently most of us use the default redirect URL: 
```
{baseUrl}/{action}/oauth2/code/{registrationId}
```
These variables are populated in *DefaultOAuth2AuthorizationRequestResolver#expandRedirectUri*.

Due to our platform setup and constraints we need all our redirect URIs to start with **https://** 
We would like to reuse the default URL but change only the protocol part.

Suggestion: add variables **scheme**, **serverName** and **serverPort**  (naming aligned with *ServletRequest* for redirect URI templating.

zeratul021:
Fixes #6239

Provided baseScheme, baseHost, basePort and basePath variables that comprise the baseUrl variable.
To avoid errors during templating, I've added all variables with empty string default.

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolver.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolver.java
@@ -27,6 +27,8 @@ import org.springframework.security.oauth2.core.endpoint.PkceParameterNames;
 import org.springframework.security.web.util.UrlUtils;
 import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+import org.springframework.web.util.UriComponents;
 import org.springframework.web.util.UriComponentsBuilder;
 
 import javax.servlet.http.HttpServletRequest;
@@ -54,6 +56,7 @@ import java.util.Map;
  */
 public final class DefaultOAuth2AuthorizationRequestResolver implements OAuth2AuthorizationRequestResolver {
 	private static final String REGISTRATION_ID_URI_VARIABLE_NAME = "registrationId";
+	private static final char PATH_DELIMITER = '/';
 	private final ClientRegistrationRepository clientRegistrationRepository;
 	private final AntPathRequestMatcher authorizationRequestMatcher;
 	private final StringKeyGenerator stateGenerator = new Base64StringKeyGenerator(Base64.getUrlEncoder());
@@ -127,7 +130,7 @@ public final class DefaultOAuth2AuthorizationRequestResolver implements OAuth2Au
 					") for Client Registration with Id: " + clientRegistration.getRegistrationId());
 		}
 
-		String redirectUriStr = this.expandRedirectUri(request, clientRegistration, redirectUriAction);
+		String redirectUriStr = expandRedirectUri(request, clientRegistration, redirectUriAction);
 
 		OAuth2AuthorizationRequest authorizationRequest = builder
 				.clientId(clientRegistration.getClientId())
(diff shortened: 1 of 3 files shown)


zeratul021:
@rwinch I've managed to craft something with some assumptions, expecting feedback.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolverTests.java
```diff
@@ -50,6 +51,7 @@
 	public void setUp() {
 		this.registration1 = TestClientRegistrations.clientRegistration().build();
 		this.registration2 = TestClientRegistrations.clientRegistration2().build();
+		this.registration3 = TestClientRegistrations.clientRegistration3().build();
```
rwinch: The registration1 doesn't care about the value other than it is different than registration. For this case, registration3 needs to use the new variables, so (if it is kept around) it should be named more explicitly.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/DefaultServerOAuth2AuthorizationRequestResolver.java
```diff
@@ -154,23 +156,52 @@ else if (AuthorizationGrantType.IMPLICIT.equals(clientRegistration.getAuthorizat
...
-		uriVariables.put("baseUrl", baseUrl);
+				.fragment(null)
+				.build();
+		String scheme = uriComponents.getScheme();
+		uriVariables.put("baseScheme", scheme == null ? "" : scheme);
+		String host = uriComponents.getHost();
+		uriVariables.put("baseHost", host == null ? "" : host);
+		// following logic is based on HierarchicalUriComponents#toUriString()
+		int port = uriComponents.getPort();
+		uriVariables.put("basePort", port == -1 ? "" : ":" + port);
```
rwinch: Add a test for -1

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolver.java
```diff
@@ -149,20 +152,49 @@ private String resolveRegistrationId(HttpServletRequest request) {
...
+				.replaceQuery(null)
+				.fragment(null)
+				.build();
+		String scheme = uriComponents.getScheme();
+		uriVariables.put("baseScheme", scheme == null ? "" : scheme);
+		String host = uriComponents.getHost();
+		uriVariables.put("baseHost", host == null ? "" : host);
+		// following logic is based on HierarchicalUriComponents#toUriString()
+		int port = uriComponents.getPort();
+		uriVariables.put("basePort", port == -1 ? "" : ":" + port);
```
rwinch: Add test for -1 port

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/registration/TestClientRegistrations.java
```diff
@@ -55,6 +55,21 @@
 				.clientSecret("client-secret");
 	}
 
+	public static ClientRegistration.Builder clientRegistration3() {
```
rwinch: Move this to be a private static method in `DefaultOAuth2AuthorizationRequestResolverTests`

zeratul021: OK

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolverTests.java
zeratul021: OK

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/DefaultServerOAuth2AuthorizationRequestResolver.java
zeratul021: I see the test class contains much less tests then non-reactive counterpart. Also, I've already moved the client registration to that specific class as requested. Would it suffice to have the test in one place?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizationRequestResolver.java
zeratul021: OK

rwinch:
Thanks for the PR @zeratul021! This is now merged into master

Issue #6885, opened 2019-05-16
Allow http basic auth to be configured using nested builder


eleftherias:
Support nested builder
------
Currently HTTP basic authentication can be configured in the following way.
```
http
	.httpBasic()
	.realmName("Realm");
```

We should allow configuration using a nested builder, as follows.
```
http
	.httpBasic(httpBasicConfig -> httpBasicConfig.realmName("Realm"));
```
The remaining methods should also be available in the nested builder, `authenticationEntryPoint`, `authenticationDetailsSource`.

Remove need for `and()`
------
Any subsequent configurations to HttpSecurity will **not** need to be chained using `and()`.
```
http
	.httpBasic(httpBasicConfig -> httpBasicConfig.realmName("Realm"))
	.cors();
```

Nest `disable()`
------
HTTP basic authentication should be disabled using a nested builder as well
```
http
    .httpBasic(HttpBasicConfigurer::disable);
```
note: `HttpBasicConfigurer` may not be the appropriate class name.

rwinch:
To keep things simple I'd say we should hold off on `httpBasic(HttpBasicConfigurer::disable)` since users will be able to do `httpBasic(httpBasicConfig -> httpBasicConfig.disable())`.

eleftherias:
@rwinch I believe those are equivalent (assuming that we continue to use the `HttpBasicConfigurer`). The IDE will actually prompt you to replace the lambda with the method reference.

rwinch:
Oops. My mistake. You are right :)

Issue #5557, opened 2018-07-22
Please add support for nested builders in the DSL 


joshlong:
This could take advantage of lambdas in Java 8. 

This would mean we could configure things in Spring Security without worrying about indents.

```
foo 
  .a ( aSpec  ->aSpec.foo() .bar()) 
  .b ( bSpec -> bSpec x().y())
```

Spring Integration and Spring Cloud Gateway both support similar nested builder style DSLs.

philwebb:
I'm really glad Josh raised this one (thanks @mbhave for pointing me at the issue). I've been thinking for some time that we might be able to offer a really nice lambda first configuration API now that we have Java 8 as the baseline. I even started to sketch out some ideas a few months ago, but I had to park them because of other work.

I'm personally of the opinion that trying to add a new lambda DSL on top of the existing `spring-security-config` might end up overwhelming the user. I wonder if we can create a brand new `spring-security-dsl` project that's designed from the ground-up to only be configured using lambdas?

The last time I though about this, I was keen to investigate a `Customizer` style callback interface that you could then use to apply additional security rules via lambdas. This might be controversial, but I also think we should move away from a fluent API and instead insist each security rule is on its own line. This might make the API a little more verbose, but I think it will result in configuration code that's easier to read. I was also inspired by AssertJ and I wondered if we could flip some of the ordering around. Rather than start with the pattern matching, and then define the rule, what if we started with the rule?

Perhaps some sample code would do a better job describing what I mean:

```java
# Existing
@Override
protected void configure(HttpSecurity http) throws Exception {
	http.authorizeRequests()
			.mvcMatchers("/actuator/beans").hasRole("BEANS")
			.requestMatchers(EndpointRequest.to("health", "info")).permitAll()
			.requestMatchers(EndpointRequest.toAnyEndpoint().excluding(MappingsEndpoint.class)).hasRole("ACTUATOR")
			.requestMatchers(PathRequest.toStaticResources().atCommonLocations()).permitAll()
			.antMatchers("/foo").permitAll()
			.antMatchers("/**").hasRole("USER")
			.and()
		.cors()
			.and()
		.httpBasic();
}
```

```java
# Lambda
@Bean
public HttpSecurityCustomizer httpSecurityCustomizer() {
	return (http) -> {
		http.requests((request) -> {
			request.mustHaveRole("BEANS").whenMatchesMvcPath("/actuator/beans");
[...]

PR #6901, opened 2019-05-23
Merge pull request #6901 from eleftherias/gh-6885-http-basic-dsl

eleftherias:
Issue: gh-5557
Fixes: gh-6885

### DIFF
--- a/config/src/main/java/org/springframework/security/config/Customizer.java
+++ b/config/src/main/java/org/springframework/security/config/Customizer.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.config;
+
+/**
+ * Callback interface that accepts a single input argument and returns no result,
+ * with the ability to throw a (checked) exception.
+ *
+ * @author Eleftheria Stein
+ * @param <T> the type of the input to the operation
+ * @since 5.2
+ */
+@FunctionalInterface
+public interface Customizer<T> {
+
+	/**
+	 * Performs the customizations on the input argument.
+	 *
+	 * @param t the input argument
+	 * @throws Exception if any error occurs
+	 */
+	void customize(T t) throws Exception;
+
+	/**
+	 * Returns a {@link Customizer} that does not alter the input argument.
+	 *
+	 * @return a {@link Customizer} that does not alter the input argument.
+	 */
+	static <T> Customizer<T> withDefaults() {
+		return t -> {};
+	}
+}
--- a/config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,6 +19,7 @@ import org.springframework.context.ApplicationContext;
 import org.springframework.http.HttpMethod;
 import org.springframework.security.authentication.AuthenticationManager;
 import org.springframework.security.authentication.AuthenticationProvider;
+import org.springframework.security.config.Customizer;
 import org.springframework.security.config.annotation.AbstractConfiguredSecurityBuilder;
 import org.springframework.security.config.annotation.ObjectPostProcessor;
 import org.springframework.security.config.annotation.SecurityBuilder;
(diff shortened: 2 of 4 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/configurers/HttpBasicConfigurer.java
```diff
@@ -95,6 +95,17 @@
...
+	}
+
+	/**
+	 * Sets up the {@link HttpBasicConfigurer} using the default values for all of
+	 * its attributes.
+	 *
+	 * @return {@link HttpBasicConfigurer} for additional customization
+	 * @throws Exception
+	 */
+	public HttpBasicConfigurer<B> withDefaults() throws Exception {
```
rwinch: I think I'd prefer to revert and potentially use a generic noop Consumer. The reason is that `withDefaults` is exposed for those not using lambda's and becomes confusing since the defaults are already applied by default. Another reason is that way we don't need a withDefaults for every configurer that we have.

I know we already talked about creating a method reference for users, but in hindsight I'd suggest we just require users to create an empty lambda. The JDK doesn't even provide one. We can always introduce a generic no op consumer later on.

config/src/main/java/org/springframework/security/config/annotation/web/HttpConfigurerConsumer.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Callback interface used to process an {@link AbstractHttpConfigurer} with the ability to
+ * throw a (checked) exception.
+ *
+ * @author Eleftheria Stein
+ * @param <C> the configurer type
+ */
+@FunctionalInterface
+public interface HttpConfigurerConsumer<C extends AbstractHttpConfigurer> {
```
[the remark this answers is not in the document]
eleftherias: The `java.util.function.Consumer` does't allow exceptions to be thrown from the `accept` method.
In our case the `realmName` method on `HttpBasicConfigurer` throws an exception, so you would not be able to use that method in a lambda with the Java `Consumer`.
I have seen a custom consumer in Spring Boot and I assume it is for the same reason. 
https://github.com/spring-projects/spring-boot/blob/36c1c051b8106371d37f8a747cea0e369a8e5c84/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/runner/ContextConsumer.java

config/src/main/java/org/springframework/security/config/annotation/web/configurers/HttpBasicConfigurer.java
eleftherias: I agree.
I believe users could still do the following to keep the defaults.
```
http
    .httpBasic(httpBasicConfigurer -> {});
```

config/src/main/java/org/springframework/security/config/annotation/web/HttpConfigurerConsumer.java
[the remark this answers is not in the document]
rwinch: Thanks for the pointer. That makes sense. A few thoughts:

* I think we should change the signature to allow throwing `Throwable` rather than `Exception`
* I'm thinking we may want to make this more central (i.e. within core) since it is likely we will need to use this in other places.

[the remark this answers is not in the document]
eleftherias: @rwinch I moved the consumer to the core module and renamed it. I also made the `Throwable` a type argument so that it doesn't need to bubble up as a `Throwable` in the calling methods.

config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
```diff
@@ -1094,6 +1095,42 @@ private ApplicationContext getContext() {
...
+	 * customized using {@link HttpBasicConfigurer#realmName(String)}.
+	 *
+	 * <pre>
+	 * &#064;Configuration
+	 * &#064;EnableWebSecurity
+	 * public class HttpBasicSecurityConfig extends WebSecurityConfigurerAdapter {
+	 *
+	 * 	&#064;Override
+	 * 	protected void configure(HttpSecurity http) throws Exception {
+	 * 		http.authorizeRequests().antMatchers(&quot;/**&quot;).hasRole(&quot;USER&quot;).and().httpBasic(httpBasicConfig -> {});
```
rwinch: I'd fix this formatting to have newlines in it. I know there are other places we are missing newlines, but that occurred by mistake.

config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
```diff
@@ -1094,6 +1095,42 @@ private ApplicationContext getContext() {
...
+	 * 	}
+	 *
+	 * 	&#064;Override
+	 * 	protected void configure(AuthenticationManagerBuilder auth) throws Exception {
```
rwinch: I'd just omit this from the configuration as we now recommend exposing a UserDetailsService via a Bean. This also allows the user to focus on this specific configuration

eleftherias: I've updated the Javadoc with both suggestions.

config/src/main/java/org/springframework/security/config/Customizer.java
```diff
@@ -0,0 +1,37 @@
...
+ * Callback interface that accepts a single input argument and returns no result,
+ * with the ability to throw a (checked) exception.
+ *
+ * @author Eleftheria Stein
+ * @param <T> the type of the input to the operation
+ * @since 5.2
+ */
+@FunctionalInterface
+public interface Customizer<T> {
+
```
rwinch: Sorry this went in a full circle. However, now that we have a custom interface again, perhaps we can reintroduce something like


```
<T> Customizer<T> withDefaults() {
    return t -> {};
}
```

sbrannen: That would need to be `static` -- right?

eleftherias: @rwinch I added `withDefaults` to the `Customizer`.
@sbrannen Yes, the method is static.

rwinch:

Issue #6925, opened 2019-05-31
OAuth2TokenValidator examples use non existing method name for creating the NimbusJwtDecoder


lgraf:
### Summary
The code examples for `OAuth2TokenValidator` use methods that are not exist, for creating a `NimbusJwtDecoder`/`NimbusReactiveJwtDecoder`.

This effects both validator examples (`JwtTimestampValidator` and custom `AudienceValidator`) in both documentations (servlet/reactive).

I have seen that changes to this api are planned (#6010), but i think this change can be helpful in the interim. If you like i provide a PR for adjusting the docs.

### Actual Behavior
The docs use the method `withOidcIssuerLocation(issuerUri)` which does not exists.
```
NimbusJwtDecoder jwtDecoder = (NimbusJwtDecoder)
        JwtDecoders.withOidcIssuerLocation(issuerUri);
```
[Servlet - JwtTimestampValidator example](https://github.com/spring-projects/spring-security/blame/master/docs/manual/src/docs/asciidoc/_includes/servlet/preface/java-configuration.adoc#L724)
[Servlet - AudienceValidator example](https://github.com/spring-projects/spring-security/blame/master/docs/manual/src/docs/asciidoc/_includes/servlet/preface/java-configuration.adoc#L764)

```
NimbusReactiveJwtDecoder jwtDecoder = (NimbusReactiveJwtDecoder)
             ReactiveJwtDecoders.withOidcIssuerLocation(issuerUri);
```
[Reactive - JwtTimestampValidator example](https://github.com/spring-projects/spring-security/blame/master/docs/manual/src/docs/asciidoc/_includes/reactive/oauth2/resource-server.adoc#L344)
[Reactive - AudienceValidator example](https://github.com/spring-projects/spring-security/blame/master/docs/manual/src/docs/asciidoc/_includes/reactive/oauth2/resource-server.adoc#L386)

Sorry for using blame links, but i can't find a way to create a link to a line in file for (rendered) `.adoc` files.

### Expected Behavior
The docs use the method `fromOidcIssuerLocation(issuerUri)`
```
NimbusJwtDecoder jwtDecoder = (NimbusJwtDecoder)
        JwtDecoders.fromOidcIssuerLocation(issuerUri);
```
```
NimbusReactiveJwtDecoder jwtDecoder = (NimbusReactiveJwtDecoder)
             ReactiveJwtDecoders.withOidcIssuerLocation(issuerUri);
```

### Version
5.1.5.RELEASE / master

[...]

PR #6927, opened 2019-05-31
Fix typo in OAuth2TokenValidator doc

lgraf:
Use correct method names for creating JwtDecoder like described in gh-6925

Obvious Fix


jzheaux:
Thanks, @lgraf, this is now merged into `master`.

Issue #6958, opened 2019-06-05
Fix HttpSecurity jee() Javadoc example for mappableRoles


eleftherias:
### Summary

The example configuration in the Javadoc for the `jee()` method in `HttpSecurity` incorrectly demonstrates that the parameters for `mappableRoles()` are prefixed with "ROLE_".
  
The method implementation of `mappableRoles()` adds the prefix "ROLE_" to any parameter, therefore if a user were to use the demonstrated configuration, then the list of `mappableRoles` would be `["ROLE_ROLE_USER", "ROLE_ROLE_ADMIN"]`.

### Actual Behavior
The Javadoc states that `mappableRoles` are configured as follows:
```
@Override
protected void configure(HttpSecurity http) throws Exception {
        http.authorizeRequests().antMatchers("/**").hasRole("USER").and()
        // Example jee() configuration
                        .jee().mappableRoles("ROLE_USER", "ROLE_ADMIN");
}
```

### Expected Behavior
The Javadoc should provide the following example configuration:
```
@Override
protected void configure(HttpSecurity http) throws Exception {
        http.authorizeRequests().antMatchers("/**").hasRole("USER").and()
        // Example jee() configuration
                        .jee().mappableRoles("USER", "ADMIN");
 }
```

PR #6959, opened 2019-06-05
Fix HttpSecurity Javadoc for jee() method

eleftherias:
Fixes: gh-6958


rwinch:

PR #6966, opened 2019-06-07
OAuth2Login: user login will fail or skipped if previous session exis…

### LINKED ISSUE #6890: OAuth2Login should process authenticated requests
Use case:

We have a system A use OAuth2Login to implement user login(Code flow). And another system B works as OAuth2 provider.

Step 1: A supervisor login system A. Everything works well.
Step 2: Supervisor close browser without logout. System A session is still available.
Step 3: A non-privilege user login system B to do something, Everything works well. Session of system B is reset to non-privilege user.
Step 4: The non-privilege user **login system A from login page**, but found that the login user is actually supervisor. (Reuse the previous supervisor session)

Expected: At step 4, since non-privilege user start a new login process from login page, we expected non-privilege user login finally.

After debug, we found that at step 4, the login process of system A have got the oauth2 code (Code flow) and redirected to the redirect url, but the redirect url is not processed correctly and skipped since the following code.

https://github.com/spring-projects/spring-security/blob/b1195e7789521845752b91677285f9ff5eeeb6b0/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java#L851-L855

Since the existence of the previous session, the process of redirect url is skipped. And the previous session is reused incorrectly.

AFAIK the process of redirect url is used to validate oauth2 code and retrieve user/token info and cannot be skipped. So it looks a security issue.

And since the process of redirect url is skipped, the saved authentication request cannot be removed from (in memory) session and potential run out server memory.

wangzw:
Fix #6890

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2LoginConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2LoginConfigurer.java
@@ -432,9 +432,7 @@ public final class OAuth2LoginConfigurer<B extends HttpSecurityBuilder<B>> exten
 				this.loginProcessingUrl);
 		this.setAuthenticationFilter(authenticationFilter);
 		super.loginProcessingUrl(this.loginProcessingUrl);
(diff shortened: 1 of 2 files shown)


jgrandja:
Thanks for the PR @wangzw. However, I ended up reverting 2 commits that fixed this issue. See 06943d2 and 6e76df8.

Thanks again for reporting.

PR #6968, opened 2019-06-08
Accept Converter in ReactiveJwtAuthenticationConverterAdapter

ttddyy:
Currently, `ReactiveJwtAuthenticationConverterAdapter` takes `JwtAuthenticationConverter` as its constructor argument. However, this limits the usage of this adapter.
In this commit, widen the constructor to take `Converter<Jwt, AbstractAuthenticationToken>` and allow this adapter to be used by generic converters.


jzheaux:
Thanks for the PR, @ttddyy! This is now merged into `master`.

Issue #6967, opened 2019-06-07
Make Spring web configuration classes use proxyBeanMethods=false by default


rwinch:
### Summary
Similar to https://github.com/spring-projects/spring-framework/issues/23004 we should make Spring Security's configuration classes proxyBeanMethods=false by default.

Related gh-6818

eleftherias:
@aclement We have now set `proxyBeanMethods=false` for spring security configurations and merged this change to master.

PR #6970, opened 2019-06-10
Disable bean proxying in configuration classes

eleftherias:
Fixes gh-6967

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/AuthenticationConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/AuthenticationConfiguration.java
@@ -57,7 +57,7 @@ import java.util.stream.Collectors;
  * @since 3.2
  *
  */
-@Configuration
+@Configuration(proxyBeanMethods = false)
 @Import(ObjectPostProcessorConfiguration.class)
 public class AuthenticationConfiguration {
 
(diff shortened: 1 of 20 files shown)


config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/EnableGlobalAuthentication.java
```diff
@@ -49,7 +49,7 @@
  * {@link AuthenticationManagerBuilder}. For example:
  *
  * <pre class="code">
- * &#064;Configuration
+ * &#064;Configuration(proxyBeanMethods = false)
```
eleftherias: The spring security annotations such as `@EnableGlobalAuthentication` do not set `proxyBeanMethods=false` by default, to avoid breaking changes.
I have updated the Javadoc to encourage users to add the additional annotation `@Configuration(proxyBeanMethods = false)`.

rwinch:
Thanks for the PR @eleftherias! I think documenting to specify `proxyBeanMethods=false` helps for passivity, but I don't think we want to do that at the cost of preventing users getting the benefits of disabling proxyBeanMethods going forward. I'd recommend removing this from the docs.

eleftherias:
@rwinch `proxyBeanMethods=false` has been removed from the Javadoc.

rwinch:

Issue #6976, opened 2019-06-12
Add DefaultAuthenticationManagerResolver


jzheaux:
There ought to be a simple way to configure multi-tenancy for common use cases like by domain, by path, and by header.

Being able to do something like

```java
@Bean 
public AuthenticationManagerResolver<HttpServletRequest> byHeader
        (Converter<String, AuthenticationManager> lookupByTenant) {
    return DefaultAuthenticationManagerResolver.resolveByHeaderValue
            ("X-Tenant-Id", lookupByTenant);
}
```

might be nice.

rwinch:
Some thoughts...perhaps create `DefaultAuthenticationManagerResolver` which allows injecting a Converter for the request to be converted into an id and then another Converter for the id to be converted into a manager. Then static factory methods on it that resemble byHeader as shown above

jzheaux:
We could achieve the same thing with:

```java
static <T> AuthenticationManagerResolver<HttpServletRequest> byRequestConverter
        (Converter<HttpServletRequest, T> requestConverter, 
                Converter<T, AuthenticationManager> resolver)
```

However, I do like the idea of a concrete implementation that can hold state. I've updated the PR accordingly.

PR #6977, opened 2019-06-12
Add MultiTenantAuthenticationManagerResolver

jzheaux:
A class with a number of handy request-based implementations of
AuthenticationManager Resolver

Fixes; gh-6976

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/MultiTenantAuthenticationManagerResolver.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/MultiTenantAuthenticationManagerResolver.java
@@ -1,0 +1,174 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.web.authentication;
+
+import java.util.Optional;
+import javax.servlet.http.HttpServletRequest;
+
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.security.authentication.AuthenticationManager;
+import org.springframework.security.authentication.AuthenticationManagerResolver;
+import org.springframework.util.Assert;
+import org.springframework.web.util.UriComponents;
+import org.springframework.web.util.UriComponentsBuilder;
+
+/**
+ * An implementation of {@link AuthenticationManagerResolver} that separates the tasks of
+ * extracting the request's tenant identifier and looking up an {@link AuthenticationManager}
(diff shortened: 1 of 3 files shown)


web/src/main/java/org/springframework/security/web/authentication/ServletAuthenticationManagerResolvers.java
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * A few simple factories for creating {@link AuthenticationManagerResolver}s based on various
+ * aspects of the {@link HttpServletRequest}
+ *
+ * @author Josh Cummings
+ * @since 5.2
+ * @see AuthenticationManagerResolver
+ */
+public final class ServletAuthenticationManagerResolvers {
```
rwinch: Let's rename this to AuthenticationManagerResolvers

rwinch: I'd like to also consider moving this to a class with a name more focused on multi tenancy. Please consider something like creating a class named `MultiTenancyAuthenticationManagerResolver` with static factory methods that do what is happening here. The `MultiTenancyAuthenticationManagerResolver` class allows injecting a strategy to resolve the tenant id from the request and then the manager from the tenant id.

web/src/main/java/org/springframework/security/web/authentication/MultiTenantAuthenticationManagerResolver.java
```diff
@@ -0,0 +1,174 @@
...
+	 *	AuthenticationManagerResolver<HttpServletRequest> resolver =
+	 *			resolveByFirstHostnameLabel(authenticationManagers::get);
+	 * </pre>
+	 *
+	 * {@link HttpServletRequest}
+	 * @param resolver A {@link String}-resolving {@link AuthenticationManagerResolver}
+	 * @return A hostname-resolving {@link AuthenticationManagerResolver}
+	 */
+	public static AuthenticationManagerResolver<HttpServletRequest>
+			resolveByFirstHostnameLabel(Converter<String, AuthenticationManager> resolver) {
```
rwinch: I'm not sure about the phrasing for this. Perhaps it is just me, but I typically hear the phrase subdomain rather than hostname label. Have you given thought to resolveBySubdomain?

jzheaux: I'm all for pragmatism, and "subdomain" may be more pragmatic. I think it reads more cleanly, too. 

The reason that I avoided it is because of [Wikipedia's definition of a subdomain](https://en.wikipedia.org/wiki/Subdomain): 

> A subdomain is a domain that is part of a larger domain; the only domain that is not also a subdomain is the root domain.[1] For example, west.example.com and east.example.com/ are subdomains of the example.com domain, which in turn is a subdomain of the com top-level domain (TLD)

If that's the case, then "west" or "east" are not subdomains, even though that's what we all often call them.

Knowing this, what are your thoughts? We could call it `resolveBySubdomain` (which is what I originally had), but would it cause cognitive dissonance?

rwinch: Thanks for the comment. I understand your reservations. I guess I wouldn't necessarily assume that the value of the tenant is the subdomain itself, but that the tenant is resolved from a subdomain. Perhaps a wording of `resolveFromSubdomain` would be better?

rwinch:
Thanks @jzheaux! This looks good to me

PR #6978, opened 2019-06-12
Add "getJwtValidator" on JwtDecoder implementations

ttddyy:
Currently, once `JwtDecoder` is created, there is no way to retrieve the validator in it.

This is a problem when `JwtDecoder` is created via `JwtDecoders#fromOidcIssuerLocation()`(or via `OAuth2ResourceServerAutoConfiguration` in boot).
Since oidc flow creates a validator that checks issuer, that validator needs to be in place.
It can replace the validators via `setJwtValidator()` but no way to **add** to existing ones except using reflection to obtain it.

This PR adds `NimbusJwtDecoder#getJwtValidator` and `NimbusReactiveJwtDecoder#getJwtValidator`.
With combining getter and setter, new validator can be added on existing ones.(still bit hacky, but at least it can add to existing one)

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
@@ -100,6 +100,16 @@ public final class NimbusJwtDecoder implements JwtDecoder {
 	}
 
 	/**
+	 * Get {@link Jwt} validator.
+	 * 
+	 * @since 5.2
+	 * @return jwt validator
+	 */
+	public OAuth2TokenValidator<Jwt> getJwtValidator() {
+		return this.jwtValidator;
+	}
+
+	/**
 	 * Use the following {@link Converter} for manipulating the JWT's claim set
 	 *
 	 * @param claimSetConverter the {@link Converter} to use
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusReactiveJwtDecoder.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusReactiveJwtDecoder.java
@@ -121,6 +121,16 @@ public final class NimbusReactiveJwtDecoder implements ReactiveJwtDecoder {
 	}
 
 	/**
+	 * Get {@link OAuth2TokenValidator} that validates incoming {@link Jwt}s.
+	 * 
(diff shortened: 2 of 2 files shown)


jzheaux:
Thanks for the PR, @ttddyy. I'm a little confused by the use case, though.

Can't you just do this:

```java
@Autowired
public void addValidators(JwtDecoder jwtDecoder) {
    OAuth2TokenValidator<Jwt> defaults = JwtValidators.createDefaultWithIssuer(issuer);
    ((NimbusJwtDecoder) jwtDecoder).setJwtValidator(
        new DelegatingOAuth2TokenValidator<Jwt>(defaults, myCustomValidator));
}
```

Since `JwtDecoders#fromOidcIssuerLocation` simply calls `JwtValidators`, this will give you the same setup as, say:

```java
@Autowired
public void addValidators(JwtDecoder jwtDecoder) {
    OAuth2TokenValidator<Jwt> defaults = jwtDecoder.getJwtValidator();
    ((NimbusJwtDecoder) jwtDecoder).setJwtValidator(
        new DelegatingOAuth2TokenValidator<Jwt>(defaults, myCustomValidator));
}
```

I think it's good to add things that make the code simpler, but to me, these seem equally simple. Can you clarify what isn't simple about what you are trying to do?

ttddyy:
Hi @jzheaux 

Ah, ok, I was thinking bit mixed up here.
`JwtDecoders.fromOidcIssuerLocation` creates validator by issuer and decoder by jwks.
(I was thinking the other way, validator by jwks and decoder by issuer, so that jwks is only available after oidc workflow)

So, I think `JwtValidators#createDefaultWithIssuer(issuer)` would work for adding custom validators as you suggested.

I have created a PR for spring-boot to consider `OAuth2TokenValidator<Jwt>` bean if available when creating `JwtDecoder` in `OAuth2ResourceServerJwtConfiguration` and `ReactiveOAuth2ResourceServerJwkConfiguration`.

While writing tests, I still needed to access inside `jwtValidator` variable [here](https://github.com/spring-projects/spring-boot/pull/17226/files#diff-460d57d638e95e86705d668ad74fc383R383), as well as for `DelegatingOAuth2TokenValidator` [here](https://github.com/spring-projects/spring-boot/pull/17226/files#diff-460d57d638e95e86705d668ad74fc383R389) for validation.

So, I think getters are nice to have, but not a show stopper since usages would be most likely to be in test for now.

jzheaux:
Okay, cool, @ttddyy, I'll close this for now, please feel free to reopen if you run into something that is difficult without the getter.

PR #7012, opened 2019-06-17
Validate token passed in query parameters same as headers

### LINKED ISSUE #7011: Reactive OAuth2 using query parameters for access_token can cause HTTP 500s
### Summary
When using ServerBearerTokenAuthenticationConverter with `setAllowUriQueryParameter(true)` then there is a possibility that the web server will return a HTTP 500 when the query parameter is not filled in. E.g: /hello?access_token=

Edit: This problem potentially also affects DefaultBearerTokenResolver

### Actual Behavior
ServerBearerTokenAuthenticationConverter.convert method throws a IllegalArgumentException due to the string being passed to BearerTokenAuthenticationToken::new being empty.

### Expected Behavior
The expected behaviour should be that the query parameter is validated in the ServerBearerTokenAuthenticationConverter.token method and a OAuth2AuthenticationException is thrown.

The validation should be similar/same as if the token was being provided from HTTP Header.

### Version
```
    implementation group: 'org.springframework.boot', name: 'spring-boot-starter-webflux', version: '2.1.5.RELEASE'
    implementation group: 'org.springframework.boot', name: 'spring-boot-starter-security', version: '2.1.5.RELEASE'
    implementation group: 'org.springframework.security', name: 'spring-security-oauth2-resource-server', version: '5.1.5.RELEASE'
    implementation group: 'org.springframework.security', name: 'spring-security-oauth2-jose', version: '5.1.5.RELEASE'
```

### Sample
https://github.com/bhavikkumar/reactive-spring-auth

bhavikkumar:
Validates access tokens passed through query parameters in the same manner as if they were passed in Authorization HTTP header as a bearer token.

This resolves #7011

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolver.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolver.java
@@ -39,6 +39,9 @@ public final class DefaultBearerTokenResolver implements BearerTokenResolver {
 	private static final Pattern authorizationPattern = Pattern.compile(
 		"^Bearer (?<token>[a-zA-Z0-9-._~+/]+)=*$",
 		Pattern.CASE_INSENSITIVE);
+	private static final Pattern queryParametersPattern = Pattern.compile(
+			"^(?<token>[a-zA-Z0-9-._~+/]+)=*$",
+			Pattern.CASE_INSENSITIVE);
 
 	private boolean allowFormEncodedBodyParameter = false;
 
@@ -90,17 +93,7 @@ public final class DefaultBearerTokenResolver implements BearerTokenResolver {
 	private static String resolveFromAuthorizationHeader(HttpServletRequest request) {
 		String authorization = request.getHeader(HttpHeaders.AUTHORIZATION);
 		if (StringUtils.startsWithIgnoreCase(authorization, "bearer")) {
-			Matcher matcher = authorizationPattern.matcher(authorization);
-
-			if (!matcher.matches()) {
-				BearerTokenError error = new BearerTokenError(BearerTokenErrorCodes.INVALID_TOKEN,
-						HttpStatus.UNAUTHORIZED,
-						"Bearer token is malformed",
-						"https://tools.ietf.org/html/rfc6750#section-3.1");
-				throw new OAuth2AuthenticationException(error);
-			}
-
-			return matcher.group("token");
+			return validateToken(authorization, authorizationPattern);
 		}
 		return null;
 	}
@@ -112,7 +105,7 @@ public final class DefaultBearerTokenResolver implements BearerTokenResolver {
 		}
 
 		if (values.length == 1) {
-			return values[0];
+			return validateToken(values[0], queryParametersPattern);
 		}
 
 		BearerTokenError error = new BearerTokenError(BearerTokenErrorCodes.INVALID_REQUEST,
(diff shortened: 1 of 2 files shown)


bhavikkumar:
I've made the assumption that the query parameters here have already been decoded and therefore validated against a pattern. 

If this is not true then just checking for empty query parameter will also resolve the issue.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolver.java
```diff
@@ -126,4 +130,17 @@ private boolean isParameterTokenSupportedForRequest(HttpServletRequest request)
 		return ((this.allowFormEncodedBodyParameter && "POST".equals(request.getMethod()))
 				|| (this.allowUriQueryParameter && "GET".equals(request.getMethod())));
 	}
+
+	private String validateQueryParamter(String parameterToken) {
```
jzheaux: I think it would help if the two methods `resolveFromAuthorizationHeader` and `resolveFromRequestParameters` looked like each other. What if you took the contents of this method and inlined them into `resolveFromRequestParameters` so that the two methods structurally read the same?

Possibly another way would be to change both methods to call a common validation method.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/server/ServerBearerTokenAuthenticationConverter.java
```diff
@@ -38,20 +38,22 @@
  * from the {@link ServerWebExchange}.
  *
  * @author Rob Winch
- * @since 5.1
```
jzheaux: Please leave this line on here. Thanks!

jzheaux:
Also, @bhavikkumar, would you please [format your commit message like this](https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.md#format-commit-messages), including at least the phrase "Fixes: gh-7011"?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/server/ServerBearerTokenAuthenticationConverter.java
bhavikkumar: Good spot. I missed that when committing.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolver.java
bhavikkumar: I've updated the PR. It now uses a common validation method now.

bhavikkumar:
@rwinch Fixed the reactive flow in https://github.com/spring-projects/spring-security/pull/7020

jzheaux:
@bhavikkumar, in taking another look at this PR and matching it against the RFC, I think there's a mismatch since the RFC doesn't specify a format for bearer token query parameters. I think it would be better to stick to the RFC, and, in fact, #7020 aligns with this on the reactive side. 

So, to better align with the RFC and also the reactive solution of #7011, would you please update your PR to complain when the query parameter is empty (not if it doesn't follow the same pattern as a bearer token header value)?

bhavikkumar:
I just wrote a quick test around this for the non reactive flow, it seems to work as expected. Closing this PR.

jzheaux:
@bhavikkumar, sorry, I'm confused. #7020 only addresses the reactive stack. I think your PR is still necessary for the servlet stack?

bhavikkumar:
I had not verified the servlet stack, but based on the code it looked like it needed the same fix. However, when I implemented it to check, it worked fine without the fix. I received the expected 401 response.

I didn't follow it through to where it was being handled. I can push up the example if you want to verify.

jzheaux:
I see, @bhavikkumar, thank you.

Issue #6498, opened 2019-02-01
Simplify access to OAuth 2.0 token attributes


jzheaux:
Related to #5200 and https://github.com/spring-projects/spring-security/pull/6352#discussion_r253137232

We should introduce `OAuth2TokenAttributes`, a class that holds an OAuth 2.0 Token's attributes. To this point, each authentication token has exposed attributes simply as a `Map`.

It works out best, though, when an new class's usage can be demonstrated, ensuring that the API is correct. So, to complete this ticket, a few things need to be done:

- [ ] **Introduce `OAuth2TokenAttributes`**

This is a simple domain object that contains a `Map`:

```java
public class OAuth2TokenAttributes {
    <T> T getAttribute(String name);
    Map<String, Object> getAttributes();
}
```

It should probably go in `oauth2-core` in `org.springframework.security.oauth2.core`.

- [ ] **Change `OAuth2IntrospectionAuthenticationToken`'s constructor**

It should take an `OAuth2TokenAttributes` for its principal instead of a `Map`. While this is a breaking change, it's allowable since that API has not GA'd yet

- [ ] **Update the `oauth2resourceserver-opaque` sample**

This sample should be adjusted to use the token's principal (`OAuth2TokenAttributes`)

### Extra Background

Spring Security offers different ways to access OAuth 2.0 token attributes, depending on how they were obtained.

For example, a Resource Server can authenticate via JWT tokens and obtain their contents in the following way:

```java
@GetMapping("/protected")
public String method(@AuthenticationPrincipal Jwt jwt) {
    String subject = jwt.getSubject();
    // ...
}

// and

@GetMapping("/protected")
public String method(@AuthenticationPrincipal(expression="subject") String subject) {
    // ...
}
```

When using introspection, though, the principal is simply a map, so the pattern is:

```java
@GetMapping("/protected")
public String method(@AuthenticationPrincipal Map<String, Object> attributes) {
    String subject = (String) attributes.get("sub");
    // ...
}

// and 

@GetMapping("/protected")
public String method(@AuthenticationPrincipal(expression="['sub']") String subject) {
    // ...
}
```

It would be nice if application code didn't have to know the token verification strategy to extract the attributes from the principal.

Further, there is potentially some clean up here with clarifying the meaning around attributes and claims. Claims are unverified attributes. Or, in other words, a JWT has a claim set, but the ensuing `Authentication` object has attributes since the JWT at that point is verified.

There is already one way to get attributes in an agnostic way:

[...]

PR #7014, opened 2019-06-18
Added OAuth2TokenAttributes to wrap attributes

clementkng:
To simplify access to OAuth 2.0 token attributes

Fixes gh-6498

### DIFF
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenAttributes.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenAttributes.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.oauth2.core;
+
+import java.util.Collections;
+import java.util.Map;
+
+/**
+ * A domain object that wraps the attributes of an OAuth 2.0 token.
+ *
+ * @author Clement Ng
+ * @since 5.2
+ */
+public final class OAuth2TokenAttributes {
+	private final Map<String, Object> attributes;
+
+	/**
+	 * Constructs an {@code OAuth2TokenAttributes} using the provided parameters.
+	 *
+	 * @param attributes the attributes of the OAuth 2.0 token
+	 */
+	public OAuth2TokenAttributes(Map<String, Object> attributes) {
+		this.attributes = Collections.unmodifiableMap(attributes);
+	}
+
+	/**
+	 * Gets the attributes of the OAuth 2.0 token in map form.
+	 *
+	 * @return a {@link Map} of the attribute's objects keyed by the attribute's names
+	 */
+	public Map<String, Object> getAttributes() {
+		return attributes;
+	}
+
+	/**
+	 * Gets the attribute of the OAuth 2.0 token corresponding to the name.
+	 *
+	 * @param name the name to lookup in the attributes
+	 * @return the object corresponding to the name in the attributes
+	 */
+	public <A> A getAttribute(String name) {
+		return (A) this.attributes.get(name);
+	}
+}
--- a/oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/authentication/OAuth2IntrospectionAuthenticationTokenTests.java
+++ b/oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/authentication/OAuth2IntrospectionAuthenticationTokenTests.java
@@ -30,6 +30,7 @@ import org.junit.Test;
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.authority.AuthorityUtils;
 import org.springframework.security.oauth2.core.OAuth2AccessToken;
+import org.springframework.security.oauth2.core.OAuth2TokenAttributes;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatCode;
@@ -46,14 +47,15 @@ public class OAuth2IntrospectionAuthenticationTokenTests {
 	private final OAuth2AccessToken token =
 			new OAuth2AccessToken(OAuth2AccessToken.TokenType.BEARER,
 				"token", Instant.now(), Instant.now().plusSeconds(3600));
-	private final Map<String, Object> attributes = new HashMap<>();
 	private final String name = "sub";
+	private Map<String, Object> attributesMap = new HashMap<>();
+	private final OAuth2TokenAttributes attributes = new OAuth2TokenAttributes(attributesMap);
 
 	@Before
 	public void setUp() {
-		this.attributes.put(SUBJECT, this.name);
-		this.attributes.put(CLIENT_ID, "client_id");
-		this.attributes.put(USERNAME, "username");
+		this.attributesMap.put(SUBJECT, this.name);
+		this.attributesMap.put(CLIENT_ID, "client_id");
+		this.attributesMap.put(USERNAME, "username");
 	}
 
 	@Test
@@ -67,7 +69,8 @@ public class OAuth2IntrospectionAuthenticationTokenTests {
 	@Test
 	public void getNameWhenHasNoSubjectThenReturnsNull() {
 		OAuth2IntrospectionAuthenticationToken authenticated =
-				new OAuth2IntrospectionAuthenticationToken(this.token, Collections.singletonMap("claim", "value"),
+				new OAuth2IntrospectionAuthenticationToken(this.token,
+						new OAuth2TokenAttributes(Collections.singletonMap("claim", "value")),
 						Collections.emptyList());
 		assertThat(authenticated.getName()).isNull();
 	}
@@ -76,7 +79,7 @@ public class OAuth2IntrospectionAuthenticationTokenTests {
 	public void getNameWhenTokenHasUsernameThenReturnsUsernameAttribute() {
 		OAuth2IntrospectionAuthenticationToken authenticated =
 				new OAuth2IntrospectionAuthenticationToken(this.token, this.attributes, Collections.emptyList());
-		assertThat(authenticated.getName()).isEqualTo(this.attributes.get(SUBJECT));
+		assertThat(authenticated.getName()).isEqualTo(this.attributes.getAttribute(SUBJECT));
 	}
 
 	@Test
@@ -92,7 +95,8 @@ public class OAuth2IntrospectionAuthenticationTokenTests {
 				.isInstanceOf(IllegalArgumentException.class)
 				.hasMessageContaining("attributes cannot be empty");
 
-		assertThatCode(() -> new OAuth2IntrospectionAuthenticationToken(this.token, Collections.emptyMap(), null))
+		assertThatCode(() -> new OAuth2IntrospectionAuthenticationToken(this.token,
+									new OAuth2TokenAttributes(Collections.emptyMap()), null))
 				.isInstanceOf(IllegalArgumentException.class)
 				.hasMessageContaining("attributes cannot be empty");
 	}
@@ -100,7 +104,8 @@ public class OAuth2IntrospectionAuthenticationTokenTests {
 	@Test
 	public void constructorWhenPassingAllAttributesThenTokenIsAuthenticated() {
 		OAuth2IntrospectionAuthenticationToken authenticated =
-				new OAuth2IntrospectionAuthenticationToken(this.token, Collections.singletonMap("claim", "value"),
+				new OAuth2IntrospectionAuthenticationToken(this.token,
+						new OAuth2TokenAttributes(Collections.singletonMap("claim", "value")),
 						Collections.emptyList(), "harris");
 		assertThat(authenticated.isAuthenticated()).isTrue();
 	}
@@ -109,7 +114,7 @@ public class OAuth2IntrospectionAuthenticationTokenTests {
 	public void getTokenAttributesWhenHasTokenThenReturnsThem() {
 		OAuth2IntrospectionAuthenticationToken authenticated =
 				new OAuth2IntrospectionAuthenticationToken(this.token, this.attributes, Collections.emptyList());
-		assertThat(authenticated.getTokenAttributes()).isEqualTo(this.attributes);
+		assertThat(authenticated.getTokenAttributes()).isEqualTo(this.attributes.getAttributes());
 	}
 
 	@Test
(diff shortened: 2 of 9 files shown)


oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenAttributes.java
```diff
@@ -0,0 +1,48 @@
...
+	public OAuth2TokenAttributes() {
+		this.attributes = new HashMap<>();
+	}
+
+	public OAuth2TokenAttributes(Map<String, Object> attributes) {
+		Assert.notEmpty(attributes, "attributes cannot be empty");
+		this.attributes = attributes;
+	}
+
+	public Map<String, Object> getAttributes() {
```
jzheaux: Let's add a `getAttribute` method so an application can look up an attribute by name without casting, e.g.:

```java
public <A> A getAttribute(String name) {
    return (A) this.attributes.get(name);
}
```

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenAttributes.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.springframework.util.Assert;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ *
+ * @author Clement Ng
+ *
```
jzheaux: Please add `@since 5.2`. Also, I'm sure you are already planning on adding a description of the class, but just calling it out here.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenAttributes.java
```diff
@@ -0,0 +1,48 @@
...
+	public OAuth2TokenAttributes(Map<String, Object> attributes) {
+		Assert.notEmpty(attributes, "attributes cannot be empty");
+		this.attributes = attributes;
+	}
+
+	public Map<String, Object> getAttributes() {
+		return attributes;
+	}
+
+	public void putAttributes(String key, String value) {
```
jzheaux: Let's please leave the class immutable.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenAttributes.java
```diff
@@ -0,0 +1,48 @@
...
+public class OAuth2TokenAttributes {
+	private Map<String, Object> attributes;
+
+	public OAuth2TokenAttributes() {
+		this.attributes = new HashMap<>();
+	}
+
+	public OAuth2TokenAttributes(Map<String, Object> attributes) {
+		Assert.notEmpty(attributes, "attributes cannot be empty");
+		this.attributes = attributes;
```
jzheaux: Please make a copy of the attributes and make it an immutable copy.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenAttributes.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ *
+ * @author Clement Ng
+ *
+ */
+public class OAuth2TokenAttributes {
```
jzheaux: It'd be great to have a test class for `OAuth2TokenAttributes`.

clementkng: I was considering it, but wasn't sure what would be tested w/i the test class that wasn't already covered in `OAuth2IntrospectionAuthenticationTokenTests`. Should I pull some attribute-specific tests from `OAuth2IntrospectionAuthenticationTokenTests`?

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenAttributes.java
```diff
@@ -0,0 +1,48 @@
...
+	private Map<String, Object> attributes;
+
+	public OAuth2TokenAttributes() {
```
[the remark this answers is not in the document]
clementkng: I added the default constructor because `constructorWhenAttributesAreNullOrEmptyThenThrowsException()` tests that an empty attributes map throws an illegal argument exception, but the setup for all the tests originally set `attributes` to an empty map (which is OK because it doesn't go through the constructor for `OAuth2IntrospectionAuthenticationToken`). One potential alternative is to have the ``OAuth2IntrospectionAuthenticationToken.attributes` method do the checking that the map is not empty and the `OAuth2TokenAttributes` are not null. Would this be better?

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenAttributes.java
clementkng: If we make the `attributes` map immutable via `Collections.unmodifiableMap`, I believe the `setUp()` method in the `OAuth2IntrospectionAuthenticationTokenTests` will have to be collapsed into the creation of the instance variables (since right now it puts values into an OAuth2.0 attributes map, but now we need the map to be prepopulated w/ data before constructing the `OAuth2.0TokenAttributes`). Would this be OK?

jzheaux: Yes, that is fine. We do want to be careful when modifying tests, but it is better than breaking encapsulation.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenAttributes.java
jzheaux: Fair point, we can add tests for this class if and when it becomes more sophisticated.

clementkng:
Thanks @jzheaux I've addressed the feedback above. Let me know if you have any additional comments!

oauth2/oauth2-core/src/test/java/org/springframework/security/oauth2/core/OAuth2TokenAttributesTest.java
```diff
@@ -0,0 +1,31 @@
...
+package org.springframework.security.oauth2.core;
+
+import org.junit.Test;
+
+/**
+ * Tests for {@link OAuth2TokenAttributes}
+ *
+ * @author Clement Ng
+ */
+public class OAuth2TokenAttributesTest {
```
jzheaux: I think this is a stray file, let's remove it.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenAttributes.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * A domain object that wraps around the attributes in an {@link org.springframework.security.core.Authentication}
+ * token.
+ *
+ * @author Clement Ng
+ * @since 5.2
+ */
+public final class OAuth2TokenAttributes {
+	private Map<String, Object> attributes;
```
jzheaux: Since it's intended for this to be immutable, let's make it `final`.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenAttributes.java
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.security.oauth2.core;
+
+import java.util.Collections;
+import java.util.Map;
+
+/**
+ * A domain object that wraps around the attributes in an {@link org.springframework.security.core.Authentication}
```
jzheaux: Let's mention OAuth 2.0 in this description. Maybe:

```suggestion
 * A domain object that wraps the attributes of an OAuth 2.0 token
```

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenAttributes.java
```diff
@@ -0,0 +1,43 @@
...
+
+	public OAuth2TokenAttributes(Map<String, Object> attributes) {
+		this.attributes = Collections.unmodifiableMap(attributes);
+	}
+
+	public Map<String, Object> getAttributes() {
+		return attributes;
+	}
+
+	public <A> A getAttribute(String name) {
```
jzheaux: Would you please add some simple Javadoc to each public method you are introducing?

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/authentication/OAuth2IntrospectionAuthenticationTokenTests.java
```diff
@@ -46,15 +46,13 @@
...
-	private final Map<String, Object> attributes = new HashMap<>();
 	private final String name = "sub";
-
-	@Before
-	public void setUp() {
-		this.attributes.put(SUBJECT, this.name);
-		this.attributes.put(CLIENT_ID, "client_id");
-		this.attributes.put(USERNAME, "username");
-	}
+	private Map<String, Object> attributesMap = new HashMap<String, Object>() {{
```
jzheaux: While this is a clever construct for initializing a `Map`, it's not used across the codebase. So, just for consistency, let's please still use the `setUp` method.

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/authentication/OAuth2IntrospectionAuthenticationTokenTests.java
```diff
@@ -76,7 +75,7 @@ public void getNameWhenHasNoSubjectThenReturnsNull() {
 	public void getNameWhenTokenHasUsernameThenReturnsUsernameAttribute() {
 		OAuth2IntrospectionAuthenticationToken authenticated =
 				new OAuth2IntrospectionAuthenticationToken(this.token, this.attributes, Collections.emptyList());
-		assertThat(authenticated.getName()).isEqualTo(this.attributes.get(SUBJECT));
+		assertThat(authenticated.getName()).isEqualTo(this.attributes.getAttributes().get(SUBJECT));
```
jzheaux: I think you could call `this.attributes.getAttribute(SUBJECT)`, right?

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/authentication/OAuth2IntrospectionAuthenticationTokenTests.java
clementkng: Sounds good, I didn't know @Before would allow `setUp` to populate the `attributesMap` before the OAuth2TokenAttributes `attributes` was made.

clementkng:
@jzheaux Thanks for the feedback, I've incorporated it into the PR.

jzheaux:
Looks great, @clementkng. The ticket outlines that the sample should also be updated - would you mind rolling that into this PR so that we can close out the ticket all at once?

clementkng:
@jzheaux I'm not sure how to update a ticket by rolling that into the PR. I assume you mean check off the ticket items, but how do I connect that to the PR via a commit?

jzheaux:
@clementkng, sorry, what I mean is just add a commit to this PR that updates the `oauth2resourceserver-opaque` sample. So after making the appropriate changes to the sample, you'd do:

```bash
git add samples/
git commit
git push origin gh-6498-OAuth2TokenAttributes-class
```

Where your commit message would be something like:

```bash
Update Opaque Token Sample

Issue: gh-6948
```

clementkng:
@jzheaux Sorry, I think I misunderstood the item on that ticket. I've looked through the `sample` folders w/i the `oauth2resourceserver-opaque` (as well as the entire directory itself), and can't seem to find usages of `OAuth2IntrospectionAuthenticationToken` to update. I quickly scanned through the `README.adoc` as well and couldn't find anything to update. I'm not sure what the appropriate changes to the sample should be.

jzheaux:
@clementkng no problem, sorry that's not so clear, let's see if I can help.

So, we want to change this here:

```diff
- public String index(@AuthenticationPrincipal(expression="['sub']") String subject) {
+ public String index(@AuthenticationPrincipal OAuth2TokenAttributes attributes) {
```

In [`OAuth2ResourceServerController`](https://github.com/spring-projects/spring-security/blob/master/samples/boot/oauth2resourceserver-opaque/src/main/java/sample/OAuth2ResourceServerController.java).

clementkng:
@jzheaux Just to check my understanding, once we add the `attributes` argument, we have the method return `String.format("Hello, %s!", attributes.getAttribute("sub"));`, which is to demonstrate the pattern we wanted to achieve at the end of the Extra Information section of the original ticket.

jzheaux:
Correct, @clementkng

clementkng:
@jzheaux Got it, there's currently an `NullPointerException at OAuth2ResourceServerApplicationITests.java:58` and an `AssertionError at NimbusReactiveJwtDecoderTests.java:122` that I'm going to dig into.

clementkng:
@jzheaux is there a way to `./gradlew` run individual integration tests? I'm currently having a hard time rooting out the issues in the sample.

clementkng:
Actually, I was able to figure out how to run the individual test. I've been able to narrow the issue down to the fact that the OAuth2TokenAttributes `attributes` being passed into the `index` method is null, and was able to see that the test fails when `.with(bearerToken(this.noScopesToken))` is called. For now, I'm not quite able to link up the two b/c I'm not very familiar w/ how the resource server and integration test code works. Would you happen to have any advice on how to parse this?

jzheaux:
When you use `@AuthenticationPrincipal SomeType someObject`, and the value comes back as `null`, it means that either the user is not authenticated or that `SomeType` is not the actual type of the principal.

Since that endpoint requires authentication, we know that it's because the principal is not actually of type `OAuth2TokenAttributes`.

So, it's probably best to check the `OAuth2IntrospectionAuthenticationToken` to see what is being set as the principal value. The principal should be of type `OAuth2TokenAttributes`.

clementkng:
@jzheaux Thanks for explaining! My understanding was that changing the constructor of `OAuth2IntrospectionAuthenticationToken` to take in a `OAuth2TokenAttributes` rather than a `Map` was enough to ensure that the principal value taken in was an `OAuth2TokenAttributes`. Am I missing something there?

Another thing I noticed was that the tests still pass w/ the old `@AuthenticationPrincipal` access pattern (`expression="['sub']") String subject`), but should this pass in a value of `null` since we've changed the aforementioned `SomeType` to be `OAuth2TokenAttributes`? If so, this might indicate something else is wrong.

jzheaux:
@clementkng, if you take a look at the `super` constructor that's invoked, you'll see that the second parameter matches up with the principal. Because the code sends the result of the `attributes` method call as that second parameter, a `Map` is getting set as the principal.

The reason that the `['sub']` expression continues to work is because the principal hasn't changed yet from a `Map`.

clementkng:
@jzheaux I see, thanks for pointing that out to me. I was able to update the sample. Please let me know if you have any other feedback!

jzheaux:
Thanks, @clementkng, this is now merged into `master`. And congrats on another great contribution!

clementkng:
@jzheaux Thanks for your help as well! When I first started out on this ticket, I had no idea how much the `OAuth2IntrospectionAuthenticationToken` class was involved in the code, so thanks for guiding me throughout trickier parts of the ticket.

PR #7042, opened 2019-06-26
Allow upgrading between different BCrypt encodings

larsgrefer:
This will allow the upgradeEncoding mechanism to upgrade from a lower-strength bcrypt to a higher-strength bcrypt encoding.

### DIFF
--- a/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.java
@@ -126,13 +126,13 @@ public class BCryptPasswordEncoder implements PasswordEncoder {
 	public boolean upgradeEncoding(String encodedPassword) {
 		if (encodedPassword == null || encodedPassword.length() == 0) {
 			logger.warn("Empty encoded password");
-			return false;
+			return true;
 		}
 
 		Matcher matcher = BCRYPT_PATTERN.matcher(encodedPassword);
 		if (!matcher.matches()) {
 			logger.warn("Encoded password does not look like BCrypt");
-			return false;
+			return true;
 		}
 		else {
 			int strength = Integer.parseInt(matcher.group(2));
--- a/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.java
@@ -126,13 +126,12 @@ public class BCryptPasswordEncoder implements PasswordEncoder {
 	public boolean upgradeEncoding(String encodedPassword) {
 		if (encodedPassword == null || encodedPassword.length() == 0) {
 			logger.warn("Empty encoded password");
-			return true;
+			return false;
 		}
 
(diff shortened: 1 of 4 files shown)


crypto/src/main/java/org/springframework/security/crypto/password/DelegatingPasswordEncoder.java
```diff
@@ -217,9 +217,15 @@ private String extractId(String prefixEncodedPassword) {
 	}
 
 	@Override
-	public boolean upgradeEncoding(String encodedPassword) {
-		String id = extractId(encodedPassword);
-		return !this.idForEncode.equalsIgnoreCase(id);
+	public boolean upgradeEncoding(String prefixEncodedPassword) {
```
rwinch: Please add a test for these changes

crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.java
```diff
@@ -125,6 +122,24 @@ public boolean matches(CharSequence rawPassword, String encodedPassword) {
 		return BCrypt.checkpw(rawPassword.toString(), encodedPassword);
 	}
 
+	@Override
+	public boolean upgradeEncoding(String encodedPassword) {
+		if (encodedPassword == null || encodedPassword.length() == 0) {
+			logger.warn("Empty encoded password");
```
rwinch: Please add additional tests

larsgrefer:
I'm not sure what the default behavior for invalid inputs should be.

`org.springframework.security.crypto.password.DelegatingPasswordEncoder#upgradeEncoding` returns `true` for all invalid inputs (null, no parseable id, unknown id) while `org.springframework.security.crypto.password.PasswordEncoder#upgradeEncoding` always returns false.

The default of `false` makes sense, because it prevents useless re-encodings on every login.

How should `org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder#upgradeEncoding` behave for invalid inputs (null or non-bcrypt)? `true`, `false` or `IllegalArgumentException` ?

rwinch:
Good question @larsgrefer
 
`DelegatingPasswordEncoder` returns true for null, no id, unknown id because we expect its configuration to change over time. All we care about is that the value is encoded with the configured PasswordEncoder. Note that these are actually valid as there is the potential to fallback if there is an unknown id, null id, etc.

`BCryptPasswordEncoder` should return false for a null input because it will blow up if it is encoded. For non-bcrypt it should fail with an Exception.

rwinch:
Thanks for the updates @larsgrefer. It appears that [the tests](https://api.travis-ci.org/v3/job/551839898/log.txt) fail.

```
org.springframework.security.crypto.bcrypt.BCryptPasswordEncoderTests > upgradeFromNullOrEmpty FAILED
    org.junit.ComparisonFailure at BCryptPasswordEncoderTests.java:191
```

larsgrefer:
@rwinch The tests are fixed now

rwinch:
Thanks for all your work on this PR @larsgrefer! This is now merged into master via d3d6a8743e9b17653320d967cc42d2e4dc50adba I added a little polish to the tests via 742df2cd1dc30a929674e341a1366effb1327364

PR #7046, opened 2019-06-27
DSL nested builder for HTTP security

eleftherias:
Issue: gh-5557

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -23,6 +23,7 @@ import java.util.Map;
 
 import javax.servlet.http.HttpServletRequest;
 
+import org.springframework.security.config.Customizer;
 import org.springframework.security.config.annotation.web.HttpSecurityBuilder;
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
 import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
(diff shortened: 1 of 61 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
```diff
@@ -1,5 +1,5 @@
 /*
```
eleftherias: `featurePolicy()` has not been updated to allow lambda configuration.
This is because there is no reasonable default for the `policyDirectives` argument.
Instead, it can be the last configuration for `headers()`.
```
http
				.headers(headers ->
					headers
						.defaultsDisabled()
						.referrerPolicy(referrerPolicy ->
								referrerPolicy.policy(ReferrerPolicy.SAME_ORIGIN)
						)
						.featurePolicy("")
				);
```

config/src/main/java/org/springframework/security/config/annotation/web/configurers/openid/OpenIDLoginConfigurer.java
```diff
@@ -148,6 +149,24 @@ public AttributeExchangeConfigurer attributeExchange(String identifierPattern) {
...
+	 * {@link AttributeExchangeConfigurer#identifierPattern(String)}
+	 *
+	 * @param attributeExchangeCustomizer the {@link Customizer} to provide more options for
+	 * the {@link AttributeExchangeConfigurer}
+	 * @return a {@link OpenIDLoginConfigurer} for further customizations
+	 * @throws Exception
+	 */
+	public OpenIDLoginConfigurer<H> attributeExchange(Customizer<AttributeExchangeConfigurer> attributeExchangeCustomizer)
+			throws Exception {
+		AttributeExchangeConfigurer attributeExchangeConfigurer = new AttributeExchangeConfigurer(".*");
```
eleftherias: The default attribute exchange identifier pattern is ".*" to match everything. It can be updated using the new `identifierPattern()` method.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/openid/OpenIDLoginConfigurer.java
```diff
@@ -443,6 +491,16 @@ public AttributeConfigurer attribute(String name) {
...
 			private String type;
 
+			/**
+			 * Creates a new instance named "default-attribute".
+			 * The name can by updated using {@link #name(String)}.
+			 *
+			 * @see AttributeExchangeConfigurer#attribute(String)
+			 */
+			private AttributeConfigurer() {
+				this.name = "default-attribute";
```
eleftherias: The default attribute name is "default-attribute". This is primarily to avoid a `NullPointerException`, users should override the name in the `name()` setter.

PR #7047, opened 2019-06-27
Fixed validation in ClientRegistration.Builder

### LINKED ISSUE #7040: Can't use a custom authorization grant type in a ClientRegistration
### Summary

Documentation suggests that defining additional grant types should be supported. Though, the ClientRegistration builder won't validate a registration using a custom grant type.

### Actual Behavior

1. Define a custom `org.springframework.security.oauth2.core.AuthorizationGrantType`.
2. Build a `org.springframework.security.oauth2.client.registration.ClientRegistration` with this custom type.

When calling `build()`, any unsupported grant type is validated as an authorization code grant type, and it fails.

### Expected Behavior

Custom grant type should not be validated with the wrong validator. It could be nice to be able to provide a custom validator.

### Version

`spring-security-oauth2-client 5.1.5.RELEASE`

edouardhue:
ClientRegistration.Builder defaulted to validating as an
authorization code registration, though a custom grant type could be in
use. The actual grant type is now verified for every case.
 - Fixed validation in ClientRegistration.Builder
 - New test that fails unless the issue is fixed.

Also made OAuth2AuthorizationGrantRequestEntityUtils public to help
implementing custom token response clients.

Fixes #7040

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
@@ -486,7 +486,7 @@ public final class ClientRegistration implements Serializable {
 				this.validateClientCredentialsGrantType();
 			} else if (AuthorizationGrantType.IMPLICIT.equals(this.authorizationGrantType)) {
 				this.validateImplicitGrantType();
-			} else {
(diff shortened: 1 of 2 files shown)


edouardhue:
@rwinch Ok, that's undone. I'll consider another ticket for this need.

rwinch:
Thanks for the updates @edouardhue! This is now merged into master and backported to 5.1.x

Issue #5660, opened 2018-08-13
ServerHttpSecurity can't set multiple authentication managers


neokyle1:
### Summary

The [documentation](https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/config/web/server/ServerHttpSecurity.html) seems to support allowing different authentication managers for different authentication specs (HttpBasicSpec and FormLoginSpec). However, those seem to always be overwritten when the .build() method is called:

[HttpBasicSpec overwritten](https://github.com/spring-projects/spring-security/blob/5.0.6.RELEASE/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java#L505)
[FormLoginSpec overwritten](https://github.com/spring-projects/spring-security/blob/5.0.6.RELEASE/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java#L509)

### Actual Behavior

HttpBasicSpec and FormLoginSpec can not use different AuthenticationManagers

### Expected Behavior

HttpBasicSpec and FormLoginSpec should be able to use different AuthenticationManagers

### Configuration

```java
    public SecurityWebFilterChain springSecurityFilterChain(ServerHttpSecurity http) {
        http
            // ...
            .httpBasic()
                    .authenticationManager(authenticationManager1)
            .and()
            .formLogin()
                    .authenticationManager(authenticationManager2);
        return http.build();
    }
```

### Version

spring-security 5.0.6.RELEASE

The problematic code referenced in the summary appears to still exist in the master branch.

### Sample

N/A

srinivasmkoppal:
what is authenticationManager1

PR #7048, opened 2019-06-28
Allow custom ReactiveAuthenticationManager for basic and form auth

ttddyy:
I just stumbled upon this issue(#5660), so fixing it.

----

Prior to this change, `HttpBasicSpec#authenticationManager` and
`FormLoginSpec#authenticationManager` were always overridden by
`ServerHttpSecurity#authenticationManager`.

This commit makes sure override only happens when custom authentication
manager was not specified.

Fixes: gh-5660


rwinch:
Thanks for the PR @ttddyy! This is now merged into master and backported via gh-7053

Issue #7049, opened 2019-06-28
NimbusJwtDecoder.withPublicKey() does not work with JWT that contains kid


firnkes:
### Summary
When using NimbusJwtDecoder.withPublicKey() to decode a JWT token that contains a kid the decoding always fails as the provided PublicKey does not match.

### Actual Behavior

``` Signed JWT rejected: Another algorithm expected, or no matching key(s) found ``` is raised, because no match key is found (JWT token contains key, provided public key not).

### Expected Behavior
As I specify the public key to used, the jwt token should be tried to be verified using this key. There is no need to find a matching key.
### Version

5.2.0.M2

### Sample

```
String verificationKey = "someKey";
String token = "someTokenWithKid";

KeyFactory kf = KeyFactory.getInstance("RSA");
byte[] decoded = Base64.getDecoder().decode(verificationKey);
EncodedKeySpec keySpec = new X509EncodedKeySpec(decoded);
RSAPublicKey publicKey = (RSAPublicKey) kf.generatePublic(keySpec);
NimbusJwtDecoder decoder = NimbusJwtDecoder.withPublicKey(publicKey).build();
decoder.decode(token);

```

jzheaux:
Thanks for the report, @firnkes, I agreed that this would be an improvement on the default Nimbus behavior.

I've submitted #7055 accordingly.

PR #7055, opened 2019-06-28
Single-key Key Selector

jzheaux:
Fixes: gh-7049

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
@@ -29,10 +29,6 @@ import javax.crypto.SecretKey;
 
 import com.nimbusds.jose.JWSAlgorithm;
 import com.nimbusds.jose.RemoteKeySourceException;
-import com.nimbusds.jose.jwk.JWKSet;
-import com.nimbusds.jose.jwk.RSAKey;
-import com.nimbusds.jose.jwk.source.ImmutableJWKSet;
-import com.nimbusds.jose.jwk.source.ImmutableSecret;
 import com.nimbusds.jose.jwk.source.JWKSource;
 import com.nimbusds.jose.jwk.source.RemoteJWKSet;
 import com.nimbusds.jose.proc.JWSKeySelector;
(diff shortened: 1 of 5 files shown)


oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/SingleKeyJWSKeySelector.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.springframework.util.Assert;
+
+/**
+ * An internal implementation of {@link JWSKeySelector} that always returns the same key
+ *
+ * @author Josh Cummings
+ * @since 5.2
+ */
+final class SingleKeyJWSKeySelector<C extends SecurityContext> implements JWSKeySelector<C> {
```
rwinch: Should we consider submitting this to Nimbus?

jzheaux: Thanks for checking on this. Yes, I'm planning on it.

jzheaux: Here is the Nimbus PR: https://bitbucket.org/connect2id/nimbus-jose-jwt/pull-requests/55/add-singlekeyjwskeyselector/diff

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -414,9 +403,8 @@ public NimbusJwtDecoder build() {
 		}
 
 		JWTProcessor<SecurityContext> processor() {
-			JWKSource<SecurityContext> jwkSource = new ImmutableSecret<>(this.secretKey);
 			JWSKeySelector<SecurityContext> jwsKeySelector =
-					new JWSVerificationKeySelector<>(this.jwsAlgorithm, jwkSource);
+					new SingleKeyJWSKeySelector<>(this.jwsAlgorithm, this.secretKey);
```
jgrandja: I don't think this should have changed since it's the `SecretKey` builder and the issue raised is for `PublicKey`.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/SingleKeyJWSKeySelector.java
```diff
@@ -0,0 +1,54 @@
...
+	private final List<Key> keySet;
+	private final JWSAlgorithm expectedJwsAlgorithm;
+
+	public SingleKeyJWSKeySelector(JWSAlgorithm expectedJwsAlgorithm, Key key) {
```
jgrandja: You can remove the `public` modifier since the class is `package-private`

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
jzheaux: Makes sense, I've raised #7056 so that this gets articulated.

PR #7057, opened 2019-06-28
Allow upgrading between different SCrypt encodings

larsgrefer:
see #7042


rwinch:
Thanks for the PR @larsgrefer! This is now merged into master with a bit of polish e1f155ba8998b7cd4a307c1e0a703dc6bf7e6693

Issue #7038, opened 2019-06-25
SecurityConfigurer.init Javadoc should state that applying other configurers should be performed in this method


rwinch:
### Summary

The [Javadoc for SecurityConfigurer.init()](https://docs.spring.io/spring-security/site/docs/4.2.12.RELEASE/apidocs/org/springframework/security/config/annotation/SecurityConfigurer.html#init-B-) should clarify that applying configurers should be performed in init as well.

clementkng:
Hi @rwinch, could I try taking a stab at this issue?

rwinch:
@clementkng Thanks for volunteering! Yes the issue is yours.

clementkng:
> applying configurers should be performed in init as well

@rwinch I'm a little confused by this statement. Where else are configurers applied? Does this statement mean that each `configure(SecurityBuilder)` is invoked (in the top of the spec, describing the `SecurityConfigurer` interface)?

rwinch:
Most of the methods on `HttpSecurity` apply a configurer. For example, invoking `formLogin` invokes a configurer. If you are to do that within another Configurer..it must be done in the init method

PR #7080, opened 2019-07-05
Changed docs to reflect that init should apply configurers

clementkng:
Fixes gh-7038


rwinch:
Thanks for the PR @clementkng! This is now merged into master

PR #7082, opened 2019-07-05
Add Chinese Traditional localized messages.

tan9:


rwinch:
Thanks for the PR @tan9! This is now merged into master

Issue #2790, opened 2014-04-25
SEC-2576: ArrayIndexOutOfBoundsException in IpAddressMatcher


spring-issuemaster:
[Andreas Woitzick](https://jira.spring.io/secure/ViewProfile.jspa?name=andreas.woitzick) (Migrated from [SEC-2576](https://jira.spring.io/browse/SEC-2576?redirect=false)) said:

When dealing with performance issues in our software we looked a little bit deeper in the sources of IpAddressMatcher class. We figured out that the matches-methods will throw an ArrayIndexOutOfBoundsException when the netmask length given at construction time is too large. "Too large" means the netmask byte array is longer the byte array obtained from the ip address.

See attached test class (ArrayIndexOutOfBoundsExceptionTest).

We think to fix these Exception is quite easy, if the byte array for the netmask is determined at construction time and some basic checks are added. (We expect this solution will also perform better if the matches method of an already constructed IpAddressMatcher is called more than once.)

Attached IpAddressMatcher class contains these changes.

For informational purposes we submit the attached IpAddressMatcher2 class with some more performance optimizations (... and API changes). These changes are only seen useful for our special needs.

Best regards
Andreas Woitzick

jzheaux:
Agreed that this is an issue. To address it, I think the code should validate the constructor input. For example, it should assert that the `requiredAddress` byte length times 8 is greater than or equal to `nMaskBits`.

The code should do the same for `remoteAddress`.

clementkng:
Hi @jzheaux, can I try taking this issue?

jzheaux:
Yes, of course, @clementkng! I'm looking forward to your PR.

PR #7084, opened 2019-07-05
Throws exception when passed IP address with too long mask

clementkng:
Originally, `matches(String address)` would first call `parse` on an address, when I believe the correct behavior is to first check for the validity of the subnet mask (an invalid mask for an IP address is invalid regardless of what method it's a part of). This is why invalid masks could be passed into the constructor but then throw exceptions on valid calls to `matches`. So I changed both the constructor and the `matches` method to first validate that the mask in the IP address passed in is not too long.

Draft b/c I'm not sure if my assumption fully holds for all cases.

Fixes gh-2790

### DIFF
--- a/web/src/main/java/org/springframework/security/web/util/matcher/IpAddressMatcher.java
+++ b/web/src/main/java/org/springframework/security/web/util/matcher/IpAddressMatcher.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,6 +22,7 @@ import javax.servlet.http.HttpServletRequest;
 
 import org.springframework.security.web.util.matcher.RequestMatcher;
 import org.springframework.util.StringUtils;
+import org.springframework.util.Assert;
 
 /**
  * Matches a request based on IP Address or subnet mask matching against the remote
@@ -55,6 +56,9 @@ public final class IpAddressMatcher implements RequestMatcher {
 			nMaskBits = -1;
 		}
 		requiredAddress = parseAddress(ipAddress);
+		Assert.isTrue(requiredAddress.getAddress().length * 8 >= nMaskBits,
+				String.format("IP address %s is too short for bitmask of length %d",
+						ipAddress, nMaskBits));
 	}
 
 	public boolean matches(HttpServletRequest request) {
--- a/web/src/test/java/org/springframework/security/web/util/matcher/IpAddressMatcherTests.java
+++ b/web/src/test/java/org/springframework/security/web/util/matcher/IpAddressMatcherTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -83,5 +83,25 @@ public class IpAddressMatcherTests {
 		matcher = new IpAddressMatcher("192.168.0.159/0");
 		assertThat(matcher.matches("123.4.5.6")).isTrue();
 		assertThat(matcher.matches("192.168.0.159")).isTrue();
+	}
+
+	// SEC-2576
+	@Test
+	public void ipv4RequiredAddressMaskTooLongThenIllegalArgumentException() {
+		String ipv4AddressWithTooLongMask = "192.168.1.104/33";
+		assertThatCode(() -> new IpAddressMatcher(ipv4AddressWithTooLongMask))
+				.isInstanceOf(IllegalArgumentException.class)
+				.hasMessage(String.format("IP address %s is too short for bitmask of " +
+						"length %d", "192.168.1.104", 33));
+	}
+
+	// SEC-2576
+	@Test
+	public void ipv6RequiredAddressMaskTooLongThenIllegalArgumentException() {
+		String ipv6AddressWithTooLongMask = "fe80::21f:5bff:fe33:bd68/129";
+		assertThatCode(() -> new IpAddressMatcher(ipv6AddressWithTooLongMask))
+				.isInstanceOf(IllegalArgumentException.class)
+				.hasMessage(String.format("IP address %s is too short for bitmask of " +
+						"length %d", "fe80::21f:5bff:fe33:bd68", 129));
 	}
 }


web/src/main/java/org/springframework/security/web/util/matcher/IpAddressMatcher.java
```diff
@@ -46,23 +47,25 @@
...
+				String.format("Mask bits too long for this address %s", ipAddress));
 	}
 
 	public boolean matches(HttpServletRequest request) {
 		return matches(request.getRemoteAddr());
 	}
 
 	public boolean matches(String address) {
-		InetAddress remoteAddress = parseAddress(address);
+		String [] addressAndMask = preprocessAddressAndMask(address);
```
jzheaux: The address passed into this method doesn't have a mask, so the preprocessing is unnecessary here.

As such, the `preprocessAddressAndMask` method is probably also unnecessary for now.

clementkng: So does this mean that we don't have to check `remote address` anymore, since there's no mask that will be assigned there?

jzheaux: Because of this line:

```java
if ((remAddr[i] & mask[i]) != (reqAddr[i] & mask[i])) {
```

The code needs to ensure that both `remAddr` and `reqAddr` are not too short. It makes sense to me that the code would verify the required address in the constructor and the remote address in the `matches` method.

clementkng: I assume you're referring to this [line](https://github.com/spring-projects/spring-security/blob/3ea9d376b2e6e0d46828de779fc5eb00ab0a8699/web/src/main/java/org/springframework/security/web/util/matcher/IpAddressMatcher.java#L90). I can add a comparison that the `remAddr.length * 8 >= nMaskBits`, but I'm having trouble coming up w/ a too short IP address. Stuff like `192.168.1` and shorter variants seem to be matching just fine.

jzheaux: Oops, I was behind a version. I'm referring to [this line](https://github.com/spring-projects/spring-security/blob/3ea9d376b2e6e0d46828de779fc5eb00ab0a8699/web/src/main/java/org/springframework/security/web/util/matcher/IpAddressMatcher.java#L84).

I wonder if it would break if you configured to match against an IPv6 address and called `matches` with an IPv4 address?

clementkng: @jzheaux In that case [this line](https://github.com/spring-projects/spring-security/blob/3ea9d376b2e6e0d46828de779fc5eb00ab0a8699/web/src/main/java/org/springframework/security/web/util/matcher/IpAddressMatcher.java#L67) will realize that the classes of the IP addresses are different (`java.net.Inet6Address` vs `java.net.Inet4Address`) and then simply return `False`. IMO this feels more correct than an `IllegalArgumentException` stating that a mask length is too long, which isn't indicative that the addresses you're comparing are of different types. Unless it is valid to try to mix and match ipv4 and ipv6 addresses?

clementkng: @jzheaux In that case [this line](https://github.com/spring-projects/spring-security/blob/3ea9d376b2e6e0d46828de779fc5eb00ab0a8699/web/src/main/java/org/springframework/security/web/util/matcher/IpAddressMatcher.java#L67) will realize that the classes of the IP addresses are different (`java.net.Inet6Address` vs `java.net.Inet4Address`) and then simply return `False`. IMO this feels more correct than an `IllegalArgumentException` stating that a mask length is too long, which isn't indicative that the addresses you're comparing are of different types. Unless it is valid to try to mix and match ipv4 and ipv6 addresses?

jzheaux: No, it's not a valid use case -- was just thinking out loud about how you might make the remote address byte array too long.

I agree that an `IPv4Address` instance will always be 4 bytes long and an `IPv6Address` will always be 16 bytes long, so there's no need to check the length of remote address.

clementkng: Thanks @jzheaux, any other feedback?

web/src/main/java/org/springframework/security/web/util/matcher/IpAddressMatcher.java
```diff
@@ -46,15 +47,18 @@
 	 */
 	public IpAddressMatcher(String ipAddress) {
 
+		String ipAddr = ipAddress;
```
jzheaux: While this may be an improvement, @clementkng, let's please leave renaming the variable for another PR, unless there is a reason it's necessary?

clementkng: @jzheaux I changed it because ‘ipaddress’ is overwritten if we parse a mask, meaning that I can’t use it in a meaningful error message (to show the ip and mask together).

clementkng: I guess an alternative would be to rewrite the message in a way i.e. `IP address <address> too short for bitmask <mask>.` Let me know what you think.

jzheaux: I agree that a message with placeholders is a good solution. I like it because of the smaller change footprint.

clementkng: @jzheaux I changed the message and changed the variables back.

clementkng:
@jzheaux Thanks for the review. Is there any other feedback?

web/src/main/java/org/springframework/security/web/util/matcher/IpAddressMatcher.java
```diff
@@ -76,6 +80,8 @@ public boolean matches(String address) {
 		byte[] reqAddr = requiredAddress.getAddress();
 
 		int nMaskFullBytes = nMaskBits / 8;
+		Assert.isTrue(remAddr.length >= nMaskFullBytes,
```
jzheaux: Actually, this check can be removed, correct @clementkng? You and I could not find a use case where this check would fail.

I'd recommend taking it out; however, if it stays, then let's make sure that the error messages align to both include the ip address and the bitmask length.

clementkng: @jzheaux good catch, forgot about that! I removed the check.

jzheaux:
Thanks, @clementkng! This is now merged into `master`.

Issue #7089, opened 2019-07-09
Error description by BearerTokenAccessDeniedHandler is misleading


edouardhue:
### Summary

The error description for the insufficient_scope case in BearerTokenAccessDeniedHandler assumes that the denial is caused by an insufficient scope, while it could have many other causes. It is misleading and should be reworded.

### Actual Behavior

The error description in `org.springframework.security.oauth2.server.resource.web.access.BearerTokenAccessDeniedHandler` for error code `insufficient_scope` states that "The token provided has insufficient scope [%s] for this request".

### Expected Behavior

RFC 6750 describes this error code with :
>          The request requires higher privileges than provided by the
>          access token.  The resource server SHOULD respond with the HTTP
>          403 (Forbidden) status code and MAY include the "scope"
>          attribute with the scope necessary to access the protected
>          resource.

An implementation could grant other authorities based on other claims from the token, or from any other source. Access could be denied because of insufficient role, origin... or any other insufficient privilege. Thus, the error description should not assume that the scope is always insufficient.

It could be reworded as : "The token provided has insufficient privileges for this request". This would fix #6280 by the way.

### Configuration

Not applicable.

### Version

spring-security-oauth2-resource-server-5.1.5.RELEASE

jzheaux:
@edouardhue, I think this makes sense.

Since the RFC says "The request requires higher privileges than provided by the access token.", I think an error message of "The token provided has insufficient privileges for this request" would be more accurate. 

The existing message is also a potential source of confusion since the spec allows for missing scopes to be included in the error message whereas the existing message includes provided scopes.

Could you provide a PR for the change, both on the reactive and servlet bearer token denial handlers?

edouardhue:
Sure, gimme a few days!

edouardhue:
@jzheaux Reading again https://tools.ietf.org/html/rfc6750#section-3, I think the scope attribute that is added if a scope is found in the token should be removed too. According to rfc6750, it should be "indicating the required scope of the access token for accessing the requested resource", but it lists the scopes extracted from the token. Plus, at section 3.1, the rfc states that in the case of insufficient_scope error, "the resource server […] MAY include the "scope" attribute with the scope necessary to access the protected resource." It is misleading to respond with the scopes from the token, as an user would understand that its token actually has all the required scopes. I suggest removing this attribute as it is optional and its value would be hard to compute.

edouardhue:
@jzheaux I just submitted #7096.

PR #7096, opened 2019-07-11
Removal of scope in WWW-Authenticate header on OAuth2 insufficient_scope error responses

edouardhue:
Fixes #gh-7089. It was found that current implementations of `BearerTokenAccessDeniedHandler` include information in their error responses about the scopes granted to the rejected OAuth2 token, whereas the OAuth RFCs states that these messages should describe the *required* scopes. This fix removes any mention of the scopes in these messages, as computing the required scopes within the `AccessDeniedHandler` looks impossible.


jzheaux:
Thanks for the PR, @edouardhue!

In preparation for merging, will you please squash your commits and format the remaining like this:

```bash
Commit Title

Any further description of what was done.

Fixes gh-7089
```

jzheaux:
Thanks for the PR, @edouardhue! This is now merged into `master`.

Issue #6945, opened 2019-06-03
More flexibility on authorizations lookup on OAuth2 resource-server


ch4mpy:
The default authentication managers / builders for OAuth2 builds authorities from "scope" claim.

IMO, this comes with serious limitations, main reason being it mixes the notions of 
* authorization: what the user is allowed to on a resource server
* scope: what the user allows a client to do in his name with resource-servers

If authorisation-server serves authorities, shouldn't we expect user authorities on resource-servers to be a scoped view (I mean a sub-set) of all user's authorities?
 
To stick to spring-security OAuth2 boot samples, I'd expect "message" to be a scope and ["message:read", "message:write"] to be authorities within this scope ("message:" prefix used as a way to prevent collisions with authorities from other scopes).

To illustrate the confusion, still looking at the samples, _subject_ user is granted with "ROLE_USER" authority only on the authorisation-server and then either "message:read", "message:write" or "none" (but never "ROLE_USER") on resource servers, depending on wich **client** grants where used. Isn't it wired?

Of course it's not so complicated to switch the `AuthenticationManager` on `JwtConfigurer` or `OpaqueTokenConfigurer` to something aware of "authorities".

Question are:
* why using "scope" to grant OAuth2 authorities in default resource-server `AuthenticationManager`? specifically when default authorization-server user token converter puts it in "authorities" claim (even `SubjectAttributeUserTokenConverter` from samples does so...)
* is it planed to register "public" claims for Spring authorities as specified by [JWT RFC](https://tools.ietf.org/html/rfc7519#section-4.2) or even [introspection one](https://tools.ietf.org/html/rfc7662#section-3.1)? Private claims work just fine thougt.
* is it planed to add some flexibility to granted-authorities retrieving from token ?

Regarding last point, it is currently easier to do for JWT than opaque tokens introspection as it is possible to configure a `Converter<Jwt, Collection<GrantedAuthority>` when for introspection, the all AuthenticationManager has to be configured.

As a decoded token is no more than a claim-set, in both cases, the ability to configure a `Converter<Map<String, Object>, Collection<GrantedAuthority>>` (or even `Converter<Map<String, Object>, Authentication>` for more flexibility) might be just enough.

jzheaux:
@ch4mpy these are good questions - do you have a proposal? I think this would help me focus my thoughts. The title of the ticket is "More flexibility on authorizations lookup", though it already seems to me to be quite flexible. In what way are you feeling encumbered?

[...]

PR #7098, opened 2019-07-12
gh-6945 configurable authoritiesClaim on JwtGrantedAuthoritiesConverter

ch4mpy:
As requested by @jzheaux in spring-projects/spring-security#6945, expose a setter for authoritiesClaimName.

Also expose a setter for authorities prefix.

Defaulted prefix to "SCOPE_" and authorities claim name to "scope" for backward compatibility but respectively "" and "authorities" would be much better defaults


ch4mpy:
Issue was split in two parts, creating two PRs instead of this single one

PR #7099, opened 2019-07-12
Configuration of session fixation and concurrency through nested builder

eleftherias:
Issue: gh-5557


eleftherias:
Closing and moving changes to gh-7046.

Issue #7100, opened 2019-07-13
JwtGrantedAuthoritiesConverter should allow configuring the authorities claim name


jzheaux:
Related to #6945

It would be nice if an application could indicate a custom claim name for `GrantedAuthority`s without creating a custom converter.

Something like:

```java
JwtGrantedAuthoritiesConverter converter = new JwtGrantedAuthoritiesConverter();
converter.setAuthoritiesClaimName("authorities");
```

Would be easier to do than:

```java
Converter<Jwt, List<GrantedAuthorities>> converter = jwt -> {
    List<String> authorities = (List<String>) jwt.getClaim("authorities");
    return authorities.stream()
        .map(authority -> "SCOPE_" + authority)
        .map(SimpleGrantedAuthority::new)
        .collect(Collectors.toList());
};
```

So, let's add a setter to `JwtGrantedAuthoritiesConverter` called `setAuthoritiesClaimName` so that the converter knows which claim to look for.

andifalk:
Hi @jzheaux 
has anybody already claimed this first timers issue?
Would make sense to implement this one together with https://github.com/spring-projects/spring-security/issues/7101. If this is ok for you I would be glad to take over implementation for both.

jzheaux:
@andifalk, it's yours!

PR #7104, opened 2019-07-13
 gh-7100 JwtGrantedAuthoritiesConverter: configurable authorities claim names

ch4mpy:
gh-7100

JwtGrantedAuthoritiesConverter: configurable authorities clai


jzheaux:
I think this is an interesting idea, @ch4mpy, though it changes the original intent of this class, which is to identify a single claim (amongst a couple of candidates), and extract claims from it. If more community members confirm that they are extracting authorities from more than one claim in the same JWT, then we can take a look at simplifying this through a new composite class.

Issue #7101, opened 2019-07-13
JwtGrantedAuthoritiesConverter should allow configuring the authority prefix


jzheaux:
Related to #6945 and #7100 

It would be nice to configure `JwtGrantedAuthoritiesConverter` with what scope prefix to use for the resulting `GrantedAuthority`s:

```java
JwtGrantedAuthoritiesConverter converter = new JwtGrantedAuthoritiesConverter();
converter.setAuthorityPrefix("ROLE_");
```

This would involve adding a new method to `JwtGrantedAuthoritiesConverter`, namely `setAuthorityPrefix`. The code would then use this prefix instead of the hardcoded "SCOPE_" prefix.

The default value should still be "SCOPE_".

ch4mpy:
Changed the spec a bit: have prefix configurable per authority claim name

I did this because authorities claim name actually has to be a collection of claim names (default is `["scope", "scp"]`) and, also, it is very convenient to add a claim to scan for more authorities. But, in such a case, it is likely that required prefix would change with claim name.

IMO, legit use-case would be parsing `authorities` claim with no prefix at all and eventually adding `SCOPE_` prefixed authorities retrieved from `scope` and `scp` claims.

jzheaux:
I disagree that the claim name needs to be a collection of claim names. The fact that it originally is a collection of claim names is an internal implementation detail to the class.

Let's keep this class very simple - if the application needs to do something complex like adding a different prefix for different claims, then it can implement the interface quite easily itself.

ch4mpy:
I suggest you provide the implementation you already have in mind, then.

jzheaux:
Thank you, @ch4mpy, for the PRs. I can tell that you think deeply and want your contributions to be helpful to the community.

I'm a little confused by [your last comment](https://github.com/spring-projects/spring-security/issues/7101#issuecomment-511815763), though. Do you feel like we are at an impasse? 

Actually, I've just been focusing on your comments here to make sure that the description in the ticket reflects the work to be done before doing a PR review. Once we're aligned here on the contract, I figure that the PR step will go quickly.

Do you have concerns about simply doing `setAuthorityPrefix(String)`? 

Since a single prefix for a single claim is a more common scenario than multiple claims and multiple prefixes, it should be the simpler of the two to articulate.

[...]

PR #7105, opened 2019-07-13
Gh 7101

ch4mpy:
gh-7101

JwtGrantedAuthoritiesConverter: configurable authorities prefix


jzheaux:
I think this is an interesting idea, @ch4mpy, though it may be a little too early for it. If extracting authorities from more than one claim - and using a different prefix accordingly - becomes common enough, then we could take a look at adding a composite class to simplify that configuration.

Issue #7035, opened 2019-06-25
Infinite loop in role hierarchy resolving


karelmaxa:
### Summary

_spring-security_ allows to define hierarchical roles to offer a convenient means of simplifying the access-control configuration data. When you provide hierarchy definition with cycle that does not include currently resolved role, resolving process will fall into infinite loop.

### Actual Behavior

Resolving of role hierarchy is not able to correctly detect cycle when currently resolved role is not included in the cycle. Resolving process falls into infinite loop.

### Expected Behavior

`CycleInRoleHierarchyException` should be thrown.

### Version

5.2.0.M3, 5.1.5.RELEASE

### Sample

Add provided test fragment into `RoleHierarchyImplTests#testCyclesInRoleHierarchy` [1].

```
try {
    roleHierarchyImpl.setHierarchy("ROLE_C > ROLE_B\nROLE_B > ROLE_A\nROLE_A > ROLE_B");
    fail("Cycle in role hierarchy was not detected!");
} catch (CycleInRoleHierarchyException e) {
}
```
I am probably able to resolve the issue and submit pull request. But i have to say that current implementation of role hierarchy resolving looks little strange. For example this condition [2] is completely useless?. 

[1] https://github.com/spring-projects/spring-security/blob/5.1.5.RELEASE/core/src/test/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImplTests.java#L139
[2] https://github.com/spring-projects/spring-security/blob/5.1.5.RELEASE/core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImpl.java#L220

pavelhoral:
I would add that there are strange methods marked with `SEC-863` comment in `RoleHierarchyImpl`. I understand that there was a need to check authorities based on their string representation, but more obvious implementation would be to use `HashMap`. Such implementation would not require additional methods such as `getRolesReachableInOneOrMoreSteps` and `addReachableRoles` And those methods are not even used consistently. You can be sure that when building the hierarchy you are working exclusively with `SimpleGrantedAuthority` with proper `equals` implementation. So those two methods are AFAIK useful only in `#getReachableGrantedAuthorities` method call and are just making the class unnecessarily more complex.

rwinch:
@pavelhoral Thank you for the Pull Request. A PR would be most welcome. I'd suggest trying to split up the changes into distinct goals as large changes take longer to get merged.

pavelhoral:
[...]

PR #7106, opened 2019-07-15
Fix infinite loop in role hierarchy resolving

karelmaxa:
This PR resolves issue with infinite loop in role hierarchy resolving. 

The original issue has been resolved in first two commits (`2d36062`, `d3eaef6`). Last commit (`be0ad67`) simplifies the implementation in general meaning and removes unnecessary logic (logic related to the comparison of custom `GrantedAuthority` implementations). We are convinced that the readability of the component is much more better, but it is not necessary to merge this commit as well.

Fixes gh-7035


rwinch:
Thanks for the PR @karelmaxa! This is now merged into master

Issue #7114, opened 2019-07-17
Allow configurable Clock in OAuth2AuthorizedClientProvider impls


jgrandja:
We should add a `Clock` member to allow for user customization for the following:

- `RefreshTokenOAuth2AuthorizedClientProvider`
- `RefreshTokenReactiveOAuth2AuthorizedClientProvider`
- `ClientCredentialsOAuth2AuthorizedClientProvider`
- `ClientCredentialsReactiveOAuth2AuthorizedClientProvider`

**NOTE:** PR #7116 needs to be merged before this task can proceed.

PR #7116, opened 2019-07-17
Introduce Reactive OAuth2AuthorizedClient Manager/Provider

jgrandja:
This PR addresses the reactive implementation of #6811.


jgrandja:
@rwinch @jzheaux I just merged #6845 and rebased this PR off master. As an FYI, all the review feedback from #6845 has also been applied to this PR. I believe this is close to merging. It would be ideal if we can merge this for M4 as well. Feel free to add polish in order to get this merged while I am away. Thank you.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/WebClientReactiveRefreshTokenTokenResponseClient.java
```diff
@@ -0,0 +1,137 @@
...
+					.uri(clientRegistration.getProviderDetails().getTokenUri())
+					.headers(tokenRequestHeaders(clientRegistration))
+					.body(tokenRequestBody(refreshTokenGrantRequest))
+					.exchange()
+					.flatMap(response -> {
+						if (!response.statusCode().is2xxSuccessful()) {
+							OAuth2Error oauth2Error = new OAuth2Error(INVALID_TOKEN_RESPONSE_ERROR_CODE,
+									"An error occurred while attempting to retrieve the OAuth 2.0 Access Token Response: " +
+											"HTTP Status Code " + response.rawStatusCode(), null);
+							throw new OAuth2AuthorizationException(oauth2Error);
```
robotmrv: possible resource leak  
from org.springframework.web.reactive.function.client.WebClient.RequestHeadersSpec#exchange javadoc 
> NOTE: You must always use one of the body or entity methods of the response to ensure resources are released. See ClientResponse for more details.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/WebClientReactiveRefreshTokenTokenResponseClient.java
```diff
@@ -0,0 +1,137 @@
...
+ * @author Joe Grandja
+ * @since 5.2
+ * @see ReactiveOAuth2AccessTokenResponseClient
+ * @see OAuth2RefreshTokenGrantRequest
+ * @see OAuth2AccessTokenResponse
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc6749#section-6">Section 6 Refreshing an Access Token</a>
+ */
+public final class WebClientReactiveRefreshTokenTokenResponseClient implements ReactiveOAuth2AccessTokenResponseClient<OAuth2RefreshTokenGrantRequest> {
+	private static final String INVALID_TOKEN_RESPONSE_ERROR_CODE = "invalid_token_response";
+	private WebClient webClient = WebClient.builder().build();
```
robotmrv: what if someone wants to use JettyClientHttpConnector  
would not this code create Netty resources in any case so they would be redundant?  
Or users could even exclude reactor netty from dependencies 
```xml
<dependency>
	<groupId>org.springframework.boot</groupId>
	<artifactId>spring-boot-starter-webflux</artifactId>
	<exclusions>
		<exclusion>
			<groupId>org.springframework.boot</groupId>
			<artifactId>spring-boot-starter-reactor-netty</artifactId>
		</exclusion>
	</exclusions>
</dependency>
<dependency>
	<groupId>org.springframework.boot</groupId>
	<artifactId>spring-boot-starter-jetty</artifactId>
</dependency>
<dependency>
	<groupId>org.eclipse.jetty</groupId>
	<artifactId>jetty-reactive-httpclient</artifactId>
</dependency>
```
like this  
then they will get NoClassDefFoundError in any case  
may be it is worth to make mandatory parameter or just lazy ?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/WebClientReactiveRefreshTokenTokenResponseClient.java
jgrandja: @robotmrv I'm not sure I'm following? The `ClientResponse.body` method is used for a successful response...

> response.body(oauth2AccessTokenResponse())

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/WebClientReactiveRefreshTokenTokenResponseClient.java
jgrandja: @robotmrv I don't see any issues here. Spring WebFlux is responsible for creating the correct `WebClient` instance with required backing implementations, whether Netty or Jetty - the dependencies in the classpath will determine this. FYI, the use can also override the default `WebClient` via `WebClientReactiveRefreshTokenTokenResponseClient.setWebClient`

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/WebClientReactiveRefreshTokenTokenResponseClient.java
robotmrv: > The ClientResponse.body method is used for a successful response...  

exactly  
it is requirement to consume body in any case  
there are several issues related to this problem (e.g. https://github.com/spring-projects/spring-framework/issues/22005#issuecomment-453479197 )  
see last paragraph from reference guide https://docs.spring.io/spring/docs/current/spring-framework-reference/web-reactive.html#webflux-client-exchange  
So my proposal is just to use in case of error 
```java
return clientResponse.bodyToMono(Void.class)
    .then(Mono.error(new OAuth2AuthorizationException(oauth2Error)));
```

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/WebClientReactiveRefreshTokenTokenResponseClient.java
robotmrv: > FYI, the user can also override the default WebClient via WebClientReactiveRefreshTokenTokenResponseClient.setWebClient

yes, but `WebClient.builder().build();` is invoked in any case (during instance initialization), so user without dependency on reactor-netty will get NoClassDefFoundError because by default DefaultWebClientBuilder uses ReactorClientHttpConnector and as far as I see it is not overridable for the static builder in this case  
org.springframework.web.reactive.function.client.DefaultWebClientBuilder#initExchangeFunction
```java
private ExchangeFunction initExchangeFunction() {
	if (this.exchangeFunction != null) {
		return this.exchangeFunction;
	}
	else if (this.connector != null) {
		return ExchangeFunctions.create(this.connector, this.exchangeStrategies);
	}
	else {
		return ExchangeFunctions.create(new ReactorClientHttpConnector(), this.exchangeStrategies);//<----here this case
	}
}
```

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/result/method/annotation/OAuth2AuthorizedClientArgumentResolver.java
```diff
@@ -70,17 +114,50 @@ public boolean supportsParameter(MethodParameter parameter) {
...
 	}
+
+	private Mono<ServerOAuth2AuthorizeRequest> authorizeRequest(String registrationId, ServerWebExchange exchange) {
+		Mono<Authentication> defaultedAuthentication = currentAuthentication();
+
+		Mono<String> defaultedRegistrationId = Mono.justOrEmpty(registrationId)
+				.switchIfEmpty(clientRegistrationId(defaultedAuthentication))
+				.switchIfEmpty(Mono.error(() -> new IllegalArgumentException("The clientRegistrationId could not be resolved. Please provide one")));
+
+		Mono<Optional<ServerWebExchange>> defaultedExchange = Mono.justOrEmpty(exchange)
```
rwinch: We should not use Optional

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/AuthorizationCodeReactiveOAuth2AuthorizedClientProvider.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @author Joe Grandja
+ * @since 5.2
+ * @see ReactiveOAuth2AuthorizedClientProvider
+ */
+public final class AuthorizationCodeReactiveOAuth2AuthorizedClientProvider implements ReactiveOAuth2AuthorizedClientProvider {
+
+	/**
+	 * Attempt to authorize the {@link OAuth2AuthorizationContext#getClientRegistration() client} in the provided {@code context}.
+	 * Returns {@code null} if authorization is not supported,
```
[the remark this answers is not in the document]
jgrandja: Done

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/ClientCredentialsReactiveOAuth2AuthorizedClientProvider.java
```diff
@@ -0,0 +1,110 @@
...
+		// is the same as acquiring a new access token (authorization).
+
+		return Mono.just(new OAuth2ClientCredentialsGrantRequest(clientRegistration))
+				.flatMap(this.accessTokenResponseClient::getTokenResponse)
+				.map(tokenResponse -> new OAuth2AuthorizedClient(
+						clientRegistration, context.getPrincipal().getName(), tokenResponse.getAccessToken()));
+	}
+
+	private boolean hasTokenExpired(AbstractOAuth2Token token) {
+		return token.getExpiresAt().isBefore(Instant.now().minus(this.clockSkew));
```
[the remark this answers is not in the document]
jgrandja: This enhancement is tracked in this ticket #7114

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/WebClientReactiveRefreshTokenTokenResponseClient.java
rwinch: This is true. I created https://github.com/spring-projects/spring-framework/issues/23491 to address this

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/WebClientReactiveRefreshTokenTokenResponseClient.java
```diff
@@ -0,0 +1,137 @@
...
+		Assert.notNull(refreshTokenGrantRequest, "refreshTokenGrantRequest cannot be null");
+		return Mono.defer(() -> {
+			ClientRegistration clientRegistration = refreshTokenGrantRequest.getClientRegistration();
+			return this.webClient.post()
+					.uri(clientRegistration.getProviderDetails().getTokenUri())
+					.headers(tokenRequestHeaders(clientRegistration))
+					.body(tokenRequestBody(refreshTokenGrantRequest))
+					.exchange()
+					.flatMap(response -> {
+						if (!response.statusCode().is2xxSuccessful()) {
```
robotmrv: It is possible to catch IllegalArgumentException for non-standard status code  
Not sure about real probability of the problem but  
there are several related issues 
https://github.com/spring-cloud/spring-cloud-sleuth/issues/1382  
https://github.com/spring-projects/spring-framework/issues/23367  
github.com/spring-projects/spring-framework/issues/23366

rwinch: Thanks @robotmrv You are right we need to ensure we don't cause a leak by calling `statusCode()` when the status is unknown. Instead we should do something like:

```java
HttpStatus status = HttpStatus.resolve(response.rawStatusCode());
if (status != null && !status.is2xxSuccessful()) {
```

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/WebClientReactiveRefreshTokenTokenResponseClient.java
rwinch: @robotmrv Spring Framework's latest SNAPSHOTs now detect the connector based upon the classpath

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/WebClientReactiveRefreshTokenTokenResponseClient.java
jgrandja: Thanks for catching this @robotmrv. I'll update similar to your proposal.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/WebClientReactiveRefreshTokenTokenResponseClient.java
robotmrv: @jgrandja I see that this case was not covered in your final commit.  
Just check that You did not miss it.

jgrandja: I did forget! Thanks for letting me know @robotmrv :) I added this polish f0515a0

Issue #7113, opened 2019-07-17
Prevent disabled user from logging in on reactive applications


eleftherias:
### Summary

In servlet applications, if a user account is disabled, locked or expired, then the user is not permitted to log in.
In reactive applications, the user is still permitted to log in regardless of their account status.

The reactive applications should behave like the servlet application by default, and prevent disabled, locked or expired accounts from logging in.

eleftherias:
This functionality is handled by the pre and post authentication checks in the `AbstractUserDetailsAuthenticationProvider` for servlet applications.
Reactive applications do not have pre authentication checks and have a noop post authentication check, as seen in `UserDetailsRepositoryReactiveAuthenticationManager`.
Instead, they should have the same checks that the servlet applications have.
Additionally, there should be a higher level class, similar to `AbstractUserDetailsAuthenticationProvider` for reactive applications, where these checks are defined.

PR #7119, opened 2019-07-21
Add support to prevent authentication when user is locked, disabled o…

eddumelendez:
…r expired for reactive apps

Currently, reactive applications doesn't perform validation when user
is locked, disabled or expired. This commit introduces these validations.

Fixes gh-7113

### DIFF
--- a/core/src/main/java/org/springframework/security/authentication/AbstractUserDetailsReactiveAuthenticationManager.java
+++ b/core/src/main/java/org/springframework/security/authentication/AbstractUserDetailsReactiveAuthenticationManager.java
@@ -167,8 +167,7 @@ public abstract class AbstractUserDetailsReactiveAuthenticationManager implement
 	 * from an implementation-specific location.
 	 *
 	 * @param username The username to retrieve
-	 * @return the user information (never <code>null</code> - instead an exception should
-	 * the thrown)
+	 * @return the user information. If authentication fails, a Mono error is returned.
 	 */
 	protected abstract Mono<UserDetails> retrieveUser(String username);
 
--- a/core/src/main/java/org/springframework/security/authentication/AbstractUserDetailsReactiveAuthenticationManager.java
+++ b/core/src/main/java/org/springframework/security/authentication/AbstractUserDetailsReactiveAuthenticationManager.java
@@ -33,6 +33,15 @@ import org.springframework.security.crypto.password.PasswordEncoder;
 import org.springframework.util.Assert;
 
 /**
+ * A base {@link ReactiveAuthenticationManager} that allows subclasses to override and work with
+ * {@link UserDetails} objects.
+ *
+ * <p>
+ * Upon successful validation, a <code>UsernamePasswordAuthenticationToken</code> will be
+ * created and returned to the caller. The token will include as its principal either a
+ * <code>String</code> representation of the username, or the {@link UserDetails} that was
+ * returned from the authentication repository.
+ *
  * @author Eddú Meléndez
  * @since 5.2
  */
--- a/core/src/main/java/org/springframework/security/authentication/AbstractUserDetailsReactiveAuthenticationManager.java
+++ b/core/src/main/java/org/springframework/security/authentication/AbstractUserDetailsReactiveAuthenticationManager.java
@@ -1,0 +1,166 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.authentication;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Scheduler;
+import reactor.core.scheduler.Schedulers;
+
+import org.springframework.context.support.MessageSourceAccessor;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.SpringSecurityMessageSource;
+import org.springframework.security.core.userdetails.ReactiveUserDetailsPasswordService;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/springframework/security/authentication/UserDetailsRepositoryReactiveAuthenticationManager.java
```diff
@@ -17,105 +17,34 @@
...
-	 * Sets the strategy which will be used to validate the loaded <tt>UserDetails</tt>
-	 * object after authentication occurs.
-	 *
-	 * @param postAuthenticationChecks The {@link UserDetailsChecker}
-	 * @since 5.2
-	 */
-	public void setPostAuthenticationChecks(UserDetailsChecker postAuthenticationChecks) {
-		Assert.notNull(this.postAuthenticationChecks, "postAuthenticationChecks cannot be null");
-		this.postAuthenticationChecks = postAuthenticationChecks;
+	public void setUserDetailsService(ReactiveUserDetailsService userDetailsService) {
```
eleftherias: Is this method necessary? The `ReactiveUserDetailsService userDetailsService ` is a required parameter in the constructor, so it would already be set when an instance of this class is created.

core/src/test/java/org/springframework/security/authentication/UserDetailsRepositoryReactiveAuthenticationManagerTests.java
```diff
@@ -39,6 +39,7 @@
```
eleftherias: Please update the Apache license header in this file to the year 2019.

core/src/main/java/org/springframework/security/authentication/UserDetailsRepositoryReactiveAuthenticationManager.java
```diff
@@ -17,105 +17,34 @@
 package org.springframework.security.authentication;
```
eleftherias: Please update the Apache license header in this file to the year 2019.

core/src/test/java/org/springframework/security/authentication/UserDetailsRepositoryReactiveAuthenticationManagerTests.java
```diff
@@ -171,4 +172,90 @@ public void authenticateWhenPostAuthenticationChecksNotSet() {
 
 		verifyZeroInteractions(this.postAuthenticationChecks);
 	}
+
+	@Test
+	public void testAuthenticateFailsIfAccountExpired() {
```
eleftherias: Please update the test names to have the form _methodWhenConditionThenResult_. This one could be `authenticateWhenAccountExpiredThenException`.

core/src/test/java/org/springframework/security/authentication/UserDetailsRepositoryReactiveAuthenticationManagerTests.java
```diff
@@ -171,4 +172,90 @@ public void authenticateWhenPostAuthenticationChecksNotSet() {
...
+		this.manager.setPasswordEncoder(this.encoder);
+		this.manager.setUserDetailsService(new MockUserDetailsServiceUserAccountExpired());
+
+		UsernamePasswordAuthenticationToken token = new UsernamePasswordAuthenticationToken(
+				this.user, this.user.getPassword());
+		try {
```
eleftherias: It would be nice to have the exception declared in the annotation `@Test(expected = LockedException.class)`, instead of using the try-fail-catch pattern. In other tests we catch the exception only if we need to validate an attribute, such as the message.

core/src/test/java/org/springframework/security/authentication/UserDetailsRepositoryReactiveAuthenticationManagerTests.java
```diff
@@ -171,4 +172,90 @@ public void authenticateWhenPostAuthenticationChecksNotSet() {
 
 		verifyZeroInteractions(this.postAuthenticationChecks);
 	}
+
+	@Test
+	public void testAuthenticateFailsIfAccountExpired() {
+		this.manager.setPasswordEncoder(this.encoder);
+		this.manager.setUserDetailsService(new MockUserDetailsServiceUserAccountExpired());
```
eleftherias: Could we use `this.userDetailsService` here instead? It would be better if this test had the same pattern as the other tests by using `when(this.userDetailsService.findByUsername(any())).thenReturn(Mono.just(expiredUser));`

eddumelendez:
Hi @eleftherias I have solved the comments. Thanks for the feedback

core/src/main/java/org/springframework/security/authentication/AbstractUserDetailsReactiveAuthenticationManager.java
```diff
@@ -0,0 +1,175 @@
...
+		Assert.notNull(this.postAuthenticationChecks, "postAuthenticationChecks cannot be null");
+		this.postAuthenticationChecks = postAuthenticationChecks;
+	}
+
+	/**
+	 * Allows subclasses to retrieve the <code>UserDetails</code>
+	 * from an implementation-specific location.
+	 *
+	 * @param username The username to retrieve
+	 * @return the user information (never <code>null</code> - instead an exception should
```
eleftherias: The Javadoc should state that a Mono error is returned on failure, and not that an exception is thrown, since the signature does not include `throws Exception`. See the `authenticate` Javadoc for more details https://github.com/spring-projects/spring-security/blob/7e845409f185f759c1ea739bb7935964549b8765/core/src/main/java/org/springframework/security/authentication/ReactiveAuthenticationManager.java#L35-L37

core/src/test/java/org/springframework/security/authentication/UserDetailsRepositoryReactiveAuthenticationManagerTests.java
```diff
@@ -171,4 +172,56 @@ public void authenticateWhenPostAuthenticationChecksNotSet() {
...
+		this.manager.setPasswordEncoder(this.encoder);
+
+		UserDetails expiredUser = User.withUsername("user")
+				.password("password")
+				.roles("USER")
+				.accountExpired(true)
+				.build();
+		when(this.userDetailsService.findByUsername(any())).thenReturn(Mono.just(expiredUser));
+
+		UsernamePasswordAuthenticationToken token = new UsernamePasswordAuthenticationToken(
```
eleftherias: Can we use `expiredUser` to create the token instead of `this.user`? It will mimic a more realistic scenario. Similarly for `lockedUser` and `disabledUser` in the other tests.

eddumelendez:
@eleftherias new commits has been submitted with the changes.

eleftherias:
Thanks for the PR @eddumelendez! This is now merged into master.

Issue #7107, opened 2019-07-16
Add support for nested builders in the DSL for reactive apps


eleftherias:
Allow spring security for reactive apps to be configured using a nested buider, taking advantage of Java 8 lambdas.

An example configuration would look like this:
```
http
	.authorizeExchange(exchanges ->
		exchanges
			.anyExchange().authenticated()
	)
	.formLogin(formLogin ->
		formLogin
			.loginPage("/custom-login")
	)
	.redirectToHttps(redirectToHttps ->
		redirectToHttps
			.httpsRedirectWhen(new PathPatternParserServerWebExchangeMatcher("/secure"))
	);
```

Relates to gh-5557.

PR #7121, opened 2019-07-22
Support nested builder in DSL for reactive apps

eleftherias:
Fixes: gh-7107

### DIFF
--- a/config/src/test/java/org/springframework/security/config/web/server/HeaderSpecTests.java
+++ b/config/src/test/java/org/springframework/security/config/web/server/HeaderSpecTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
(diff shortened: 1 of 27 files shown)


config/src/test/java/org/springframework/security/config/web/server/HeaderSpecTests.java
```diff
@@ -49,7 +50,7 @@
  */
 public class HeaderSpecTests {
 
-	private ServerHttpSecurity.HeaderSpec headers = ServerHttpSecurity.http().headers();
+	private ServerHttpSecurity serverHttpSecurity = ServerHttpSecurity.http();
```
rwinch: To be consistent with the variable we typically use, please change to `http`. We also use http to make it look [similar to the XML based configuration](https://spring.io/blog/2013/07/11/spring-security-java-config-preview-readability/#comparison-to-xml-namespace).

rwinch:
Thanks this is now merged into master!

PR #7125, opened 2019-07-24
Change constant SAVED_REQUEST to public

quaff:
Application code can refer to it.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/savedrequest/HttpSessionRequestCache.java
+++ b/web/src/main/java/org/springframework/security/web/savedrequest/HttpSessionRequestCache.java
@@ -37,7 +37,7 @@ import org.springframework.security.web.util.matcher.RequestMatcher;
  * @since 3.0
  */
 public class HttpSessionRequestCache implements RequestCache {
-	static final String SAVED_REQUEST = "SPRING_SECURITY_SAVED_REQUEST";
+	public static final String SAVED_REQUEST = "SPRING_SECURITY_SAVED_REQUEST";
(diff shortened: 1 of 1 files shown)


rwinch:
Why do you need to refer to it? You can use the `RequestCache` to read and write the value.

quaff:
It's a trivial request, If you think not worthy, never mind.
I'm implementing my own `HttpSession`, I don't want SAVED_REQUEST trigger setDirty(true), code like this:
```
	@Override
	public Object getAttribute(String key) {
		return attrMap.get(key);
	}

	@Override
	public void setAttribute(String key, Object object) {
		attrMap.put(key, object);
		if (!key.equals("SPRING_SECURITY_SAVED_REQUEST"))
			setDirty(true);
	}
	@Override
	public void removeAttribute(String key) {
		attrMap.remove(key);
		if (!key.equals("SPRING_SECURITY_SAVED_REQUEST"))
			setDirty(true);
	}
```

rwinch:
I'm closing this because it unnecessarily breaks encapsulation. Instead you can set the attribute on the request cache and ensure the value you set it to matches that of the custom HttpSession.

quaff:
@rwinch How to autowire my own `RequestCache` implementation using xml configuration? It doesn't works even if mark it as primary, thanks a lots.

rwinch:
See https://docs.spring.io/spring-security/site/docs/5.1.x/reference/htmlsingle/#nsa-request-cache-attributes

quaff:
👍

PR #7083, opened 2019-07-05
Allow configuration of SessionAuthenticationStrategy for CSRF

samiconductor:
Reasoning https://github.com/spring-projects/spring-security/issues/5300#issuecomment-501335705

Relates to #5300

PR #7127, opened 2019-07-24
Fix CSRF session authentication strategy since version

samiconductor:
Bump `@since` version on methods for configurable CSRF session authentication strategy.

Fixes #7083


rwinch:
Thanks for the PR @samiconductor! This is now merged into master

Issue #7128, opened 2019-07-24
Remove exception from security configuration methods


eleftherias:
### Summary

The new additions to `HttpSecurity` and `ServerHttpSecurity` that support lambdas throw an exception.
These exceptions can be removed because the `Customizer` will no longer throw an exception.

PR #7131, opened 2019-07-24
Remove exceptions from lambda security configuration

eleftherias:
Fixes: gh-7128


rwinch:
Thanks for the PR @eleftherias! This is merged into master

Issue #6011, opened 2018-10-23
ldap-server XML directive should support UnboundIdContainer


jzheaux:
On the Java config side, the code selects an in-memory container based on dependencies on the classpath, choosing between `ApacheDSContainer` and `UnboundIdContainer`.

XML should have this support, too, in order to simplify transition, should a user need to switch to the unbound id container without needing to jump over to the Java DSL as well.

eddumelendez:
Hi @jzheaux 

I am looking into this issue. I am wondering which should be the approach to test it

* Create an example of that specific use case
* Add a submodule for `unboundid` to test it with a different classpath than `apacheds` and then merge it into `spring-security-config`

Thanks in advance

jzheaux:
Hi, @eddumelendez 

First, yes, I like the idea of an example. I think that it would be compelling to enhance the existing ldap sample with instructions on how the user can modify the dependency list and see the sample use unboundid instead of apacheds.

Regarding tests, there really isn't a simple way yet to induce a selection between the two. I lean towards introducing a Gradle test task that would run with a different classpath. So, for example, there might be a task called "unboundidTest". By "submodule", do you mean introducing something like "unboundid-test" alongside "main", "test", and "integration-test"?

Eventually, it would be nice to have a way to target which to use, independent from the classpath, but `LdapServerBeanDefinitionParser` doesn't support this, and it's a bit gnarly to set it up directly for a unit test. Though, technically, constructing an `LdapServerBeanDefinitionParser` directly might be an option, too.

eddumelendez:
@jzheaux my current implementation is based on classpath detection. What if used this issue to improve `LdapServerBeanDefinitionParser` and support `embedded-server` attribute which will be `APACHEDS` or `UNBOUNDID` but by default will be always `APACHEDS` to maintain the backward compatibility. WDYT?

jzheaux:
@eddumelendez I guess my concern with that is it would never be used outside the project. For example, a user would only need that setting if they themselves had `apacheds` and `unboundid` both on the classpath, which seems a bit unlikely.

That said, in light of the level of effort given the other two scenarios, yes, I think that it's reasonable to addition.

eddumelendez:
@jzheaux sorry for the delay. I already submitted a draft for this. still thinking about the testing for `unboundid`. we can continue the discussion in the draft :)

PR #7149, opened 2019-07-26
Add unbounid support in xml

eddumelendez:
Currently, unboundid was added as a support for embbeded LDAP and it
is used on the Java Config. This commit introduces support from XML side.
Also, give the chance to users to move from apacheds to unboundid using
a new attribute `mode`.

Fixes gh-6011

### DIFF
--- a/config/src/main/java/org/springframework/security/config/ldap/LdapServerBeanDefinitionParser.java
+++ b/config/src/main/java/org/springframework/security/config/ldap/LdapServerBeanDefinitionParser.java
@@ -17,9 +17,6 @@ package org.springframework.security.config.ldap;
 
 import java.io.IOException;
 import java.net.ServerSocket;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Map;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -77,16 +74,6 @@ public class LdapServerBeanDefinitionParser implements BeanDefinitionParser {
 
 	private static final String APACHEDS_CONTAINER_CLASSNAME = "org.springframework.security.ldap.server.ApacheDSContainer";
 	private static final String UNBOUNDID_CONTAINER_CLASSNAME = "org.springframework.security.ldap.server.UnboundIdContainer";
-
-	private Map<String, EmbeddedLdapServer> embeddedServers;
-
-	public LdapServerBeanDefinitionParser() {
-		Map<String, EmbeddedLdapServer> embeddedLdapServers = new HashMap<>();
-		embeddedLdapServers.put("apacheds", new EmbeddedLdapServer(BeanIds.EMBEDDED_APACHE_DS, APACHEDS_CLASSNAME, APACHEDS_CONTAINER_CLASSNAME));
-		embeddedLdapServers.put("unboundid", new EmbeddedLdapServer(BeanIds.EMBEDDED_UNBOUNDID, UNBOUNID_CLASSNAME, UNBOUNDID_CONTAINER_CLASSNAME));
-
-		this.embeddedServers = Collections.unmodifiableMap(embeddedLdapServers);
-	}
 
 	public BeanDefinition parse(Element elt, ParserContext parserContext) {
 		String url = elt.getAttribute(ATT_URL);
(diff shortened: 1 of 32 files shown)


rwinch:
Thanks for putting this together. I assume you are still working on it since it is marked WIP and the build is failing?

eddumelendez:
@rwinch this PR is ready to be reviewed

config/src/main/java/org/springframework/security/config/ldap/LdapServerBeanDefinitionParser.java
```diff
@@ -142,34 +150,61 @@ private RootBeanDefinition createEmbeddedServer(Element element,
...
+				.containsBeanDefinition(BeanIds.EMBEDDED_APACHE_DS) ||
+				parserContext.getRegistry().containsBeanDefinition(BeanIds.EMBEDDED_UNBOUNDID)) {
 			parserContext.getReaderContext().error(
 					"Only one embedded server bean is allowed per application context",
 					element);
 		}
 
-		parserContext.getRegistry().registerBeanDefinition(BeanIds.EMBEDDED_APACHE_DS,
-				apacheContainer);
+		if (isApacheDSEnabled(mode)) {
```
rwinch: I think the logic is a bit off here.

* If the mode is empty and apacheds is on the classpath use apacheds
* if the mode is apacheds, then always use apacheds (if it is not on the classpath it would then fail with an exception)
* if the mode is empty and unbounded is on the classpath use unbounded
* if the mode is unbounded, then always use unbounded (if it is not on the classpath it would then fail with an exception)

config/src/main/resources/org/springframework/security/config/spring-security-5.2.xsd
```diff
@@ -222,6 +222,19 @@
                 </xs:documentation>
          </xs:annotation>
       </xs:attribute>
+      <xs:attribute name="mode">
+         <xs:annotation>
+            <xs:documentation>Explicitly specifies which embedded ldap server should use. Values are `apacheds` and
+                `unboundid` Default `apacheds`.
```
rwinch: The default should be empty which selects based on the classpath

eddumelendez:
Thanks for the feedback @rwinch I will be working on these changes but I think tests will need to be out of the module and should be as a integration test or sample? WDYT?

config/src/main/java/org/springframework/security/config/ldap/LdapServerBeanDefinitionParser.java
eddumelendez: I took a new approach to provide this logic

eddumelendez:
@rwinch PR updated.

config/src/main/java/org/springframework/security/config/ldap/LdapServerBeanDefinitionParser.java
```diff
@@ -142,34 +166,78 @@ private RootBeanDefinition createEmbeddedServer(Element element,
...
+		if (embeddedLdapServer != null) {
+			parserContext.getRegistry().registerBeanDefinition(embeddedLdapServer.getBeanId(),
+					ldapContainer);
+		}
 
 		return (RootBeanDefinition) contextSource.getBeanDefinition();
 	}
 
+	private RootBeanDefinition getRootBeanDefinition(String mode) {
+		if (StringUtils.hasLength(mode)) {
```
rwinch: It seems like there is a lot of duplication in the code. This block looks very similar to resolveEmbeddedLdapServer

config/src/main/java/org/springframework/security/config/ldap/LdapServerBeanDefinitionParser.java
```diff
@@ -142,34 +166,78 @@ private RootBeanDefinition createEmbeddedServer(Element element,
...
+			if (isEmbeddedServerEnabled(mode)) {
+				return new RootBeanDefinition(this.embeddedServers.get(mode).getContainerClass(), null, null);
+			}
+		}
+		else {
+			for (Map.Entry<String, EmbeddedLdapServer> entry : this.embeddedServers.entrySet()) {
```
rwinch: This iterates over the entries, but the order is not guaranteed in a `HashMap`. We want to be sure that ApacheDS is used first and then Unboundid (since this is more passive). Since there are only two entries, perhaps we could avoid using a `Collection` or `Map`.

config/src/main/java/org/springframework/security/config/ldap/LdapServerBeanDefinitionParser.java
```diff
@@ -142,34 +166,78 @@ private RootBeanDefinition createEmbeddedServer(Element element,
...
+			parserContext.getRegistry().registerBeanDefinition(embeddedLdapServer.getBeanId(),
+					ldapContainer);
+		}
 
 		return (RootBeanDefinition) contextSource.getBeanDefinition();
 	}
 
+	private RootBeanDefinition getRootBeanDefinition(String mode) {
+		if (StringUtils.hasLength(mode)) {
+			if (isEmbeddedServerEnabled(mode)) {
```
rwinch: Internally `isEmbeddedServerEnabled` resolves the embedded server and then it isn't used which is a bit strange.

config/src/main/java/org/springframework/security/config/ldap/LdapServerBeanDefinitionParser.java
```diff
@@ -142,34 +166,78 @@ private RootBeanDefinition createEmbeddedServer(Element element,
...
+				.containsBeanDefinition(BeanIds.EMBEDDED_APACHE_DS) ||
+				parserContext.getRegistry().containsBeanDefinition(BeanIds.EMBEDDED_UNBOUNDID)) {
 			parserContext.getReaderContext().error(
 					"Only one embedded server bean is allowed per application context",
 					element);
 		}
 
-		parserContext.getRegistry().registerBeanDefinition(BeanIds.EMBEDDED_APACHE_DS,
-				apacheContainer);
+		EmbeddedLdapServer embeddedLdapServer = resolveEmbeddedLdapServer(mode);
```
rwinch: Calling `getRootBeanDefinition` above (https://github.com/spring-projects/spring-security/pull/7149/files#diff-ddb1e7b48066bb067a15640ca4b9388cR170) calls `isEmbeddedServerEnabled` which then calls `resolveEmbeddedLdapServer`. Then `resolveEmbeddedLdapServer` is called again. I think we want to simplify this logic.

config/src/main/java/org/springframework/security/config/ldap/LdapServerBeanDefinitionParser.java
```diff
@@ -142,34 +166,78 @@ private RootBeanDefinition createEmbeddedServer(Element element,
...
+		}
+		else {
+			for (Map.Entry<String, EmbeddedLdapServer> entry : this.embeddedServers.entrySet()) {
+				EmbeddedLdapServer ldapServer = entry.getValue();
+				if (ClassUtils.isPresent(ldapServer.getClassName(), getClass().getClassLoader())) {
+					return ldapServer;
+				}
+			}
+		}
+		return null;
```
rwinch: By the time `createEmbeddedServer` is called, aren't we certain that an embedded server needs to be created? When is it ever expected it will be null? When would we expect `isEmbeddedServerEnabled` to be false?

eddumelendez:
@rwinch I have updated the PR taking into account your comments. complexity have been reduced in the implementation and test covering all the scenarios has been added.

rwinch:
Thanks for the fast turnaround @eddumelendez! This is now merged into master

Issue #4310, opened 2017-04-26
HTTP Host header attack


fraenku:
The class `UrlUtils` is using the method` getServerName() `of the `HttpServletRequest.`
This method indeed is not secure since it could be manipulated through the host-header

See also: http://www.skeletonscribe.net/2013/05/practical-http-host-header-attacks.html)
or: https://find-sec-bugs.github.io/bugs.htm#SERVLET_SERVER_NAME

See also: https://github.com/ESAPI/esapi-java-legacy/blob/develop/src/main/java/org/owasp/esapi/filters/SecurityWrapperRequest.java
for a list of potential request headers which are manipulable.

ibaskine:
This issue is detected by Burp Scanner, which makes it difficult selling products based on Spring Security to security conscious customers. Are there any plans fixing this issue or is there any workaround?

fraenku:
Some progress on this? issue is open since April...
The issue above leaded to an attack where an URL in an e-mail could be manipulated (which can be easily used for phising-attacks)

savas127:
We are also having diffuculty to explain this vulnerability to our customers. They suggest us to add a white list to check hostname.

ibaskine:
Just wander. Can I use code snippet below for getting server name?

new java.net.URI(request.getRequestURL().toString()).getHost()

gjoseph:
FWIW SourceClear reports this as a vulnerability (https://www.sourceclear.com/vulnerability-database/security/incorrect-headers-handling/java/sid-20558) - any chance this could get patched up soon?

jzheaux:
Since the `Host` header is untrusted information, we should likely consider whitelisting as a solution. `StrictHttpFirewall` is the most suitable, so I'd propose adding a method `StrictHttpFirewall#setAllowedHostnames(Predicate)`, which would ensure the hostname in `HttpServletRequest` is a known good value.

The application would need to set this value like so:

```java
void configure(WebSecurity web) {
    StrictHttpFirewall firewall = new StrictHttpFirewall();
    firewall.setAllowedHostnames(hostname -> hostname.equals("example.org"));
    web
        .httpFirewall(firewall);
}
```

With this solution, `UrlUtils` would still invoke `request.getServerName()`, but it would be a known good value at that point.

@fraenku would this address the issue?

avodonosov:
Does anyone know a use case where spring-security itself opens a vulnerability due to Host manipulation?

[...]

PR #7158, opened 2019-07-27
Add support for allowedHostnames in StrictHttpFirewall

eddumelendez:
Introduce a new method `setAllowedHostnames` which perform the validation
against untrusted hostnames.

Fixes gh-4310

### DIFF
--- a/web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
+++ b/web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2017 the original author or authors.
+ * Copyright 2012-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,6 +26,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.function.Predicate;
 
 /**
  * <p>
@@ -66,10 +67,15 @@ import java.util.Set;
  * Rejects URLs that contain a URL encoded percent. See
  * {@link #setAllowUrlEncodedPercent(boolean)}
  * </li>
+ * <li>
+ * Rejects hosts that are not allowed. See
+ * {@link #setAllowedHostnames(Predicate)}
+ * </li>
  * </ul>
  *
  * @see DefaultHttpFirewall
  * @author Rob Winch
+ * @author Eddú Meléndez
  * @since 4.2.4
  */
 public class StrictHttpFirewall implements HttpFirewall {
@@ -97,6 +103,8 @@ public class StrictHttpFirewall implements HttpFirewall {
 	private Set<String> decodedUrlBlacklist = new HashSet<>();
 
 	private Set<String> allowedHttpMethods = createDefaultAllowedHttpMethods();
+
+	private Predicate<String> allowedHostnames = hostname -> true;
 
 	public StrictHttpFirewall() {
 		urlBlacklistsAddAll(FORBIDDEN_SEMICOLON);
@@ -297,6 +305,13 @@ public class StrictHttpFirewall implements HttpFirewall {
 		}
 	}
 
+	public void setAllowedHostnames(Predicate<String> allowedHostnames) {
+		if (allowedHostnames == null) {
+			throw new IllegalArgumentException("allowedHostnames cannot be null");
+		}
+		this.allowedHostnames = allowedHostnames;
+	}
+
 	private void urlBlacklistsAddAll(Collection<String> values) {
 		this.encodedUrlBlacklist.addAll(values);
 		this.decodedUrlBlacklist.addAll(values);
@@ -311,6 +326,7 @@ public class StrictHttpFirewall implements HttpFirewall {
 	public FirewalledRequest getFirewalledRequest(HttpServletRequest request) throws RequestRejectedException {
 		rejectForbiddenHttpMethod(request);
 		rejectedBlacklistedUrls(request);
+		rejectedUntrustedHosts(request);
 
 		if (!isNormalized(request)) {
 			throw new RequestRejectedException("The request was rejected because the URL was not normalized.");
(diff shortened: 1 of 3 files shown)


eddumelendez:
are changes needed in `DummyRequest`?

roricolivares:
Any update on this change? Our latest release is being flagged due to this issue now showing up on  SourceClear. https://www.sourceclear.com/vulnerability-database/security/incorrect-headers-handling/java/sid-20558.

eddumelendez:
thanks for the feedback @jzheaux. Changes has been submitted

web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
```diff
@@ -98,6 +104,8 @@
 
 	private Set<String> allowedHttpMethods = createDefaultAllowedHttpMethods();
 
+	private Predicate<String> allowedHostnames;
```
jzheaux: Let's set this to a default value, like `hostname -> true`.

web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
```diff
@@ -297,6 +305,12 @@ public void setAllowUrlEncodedPercent(boolean allowUrlEncodedPercent) {
 		}
 	}
 
+	public void setAllowedHostnames(Predicate<String> allowedHostnames) {
+		if (allowedHostnames != null) {
```
jzheaux: Let's please have this throw an exception if a `null` predicate is given. Something like:

```java
if (allowedHostnames == null) {
    throw new IllegalArgumentException("allowedHostnames cannot be null");
}
```

web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
```diff
@@ -352,6 +367,13 @@ private void rejectedBlacklistedUrls(HttpServletRequest request) {
 		}
 	}
 
+	private void rejectedUntrustedHosts(HttpServletRequest request) {
+		String serverName = request.getServerName();
+		if (this.allowedHostnames != null && !this.allowedHostnames.test(serverName)) {
```
jzheaux: Sorry, the concern isn't that `this.allowedHostnames` would be `null` at this point, but that `serverName` would be `null`. Because of the change to `DummyRequest`, it's possible for `getServerName` to return `null`.

So, we'd want to do:

```suggestion
		if (serverName != null && !this.allowedHostnames.test(serverName)) {
```

eddumelendez:
fixed :)

jzheaux:
Thanks again, @eddumelendez! In preparation for merging, would you go ahead and squash your commits?

eddumelendez:
@jzheaux done :)

jzheaux:
Nice work, @eddumelendez, this is now merged into `master`.

avodonosov:
@jzheaux , @eddumelendez , thanks for fixing that. I see the 5.x series release is created.

Can we hope it will also be released in 4.x soon?

How about 3.x?

jzheaux:
@avodonosov There aren't plans to backport this since it introduces a public API, `StrictHttpFirewall#setAllowableHostnames`.

Note, though, that adding the functionality yourself is quite simple as the `HttpFirewall` can be customized:

```java
void configure(WebSecurity web) {
    web
        .httpFirewall(new MyCustomFirewall());
}
```

avodonosov:
@jzheaux , why introducing a public API prevents backporting? (If it was breaking public API then I understood).

Thanks for the reply though.

Yes, it's possible to implement the same whitelist solution ourselves.

jzheaux:
> Yes, it's possible to implement the same whitelist solution ourselves.

I'm glad to hear that. Let me know if you run into any trouble.

> why introducing a public API prevents backporting?

Spring Security follows semantic versioning. One of the key points is [number 7](https://semver.org/#spec-item-7) (**emphasis mine**):

> **Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API.** It MUST be incremented if any public API functionality is marked as deprecated. It MAY be incremented if substantial new functionality or improvements are introduced within the private code. It MAY include patch level changes. Patch version MUST be reset to 0 when minor version is incremented.

We've made rare exceptions to this in the past, but this doesn't appear to be one of those.

What it means is that the public API doesn't get added to except in minor version releases, e.g. 5.2, 5.3, etc.

If there were a known vulnerability related to the Host header, it might be fixed in a different way from this general solution, so it's not really sensible to backport the whitelisting feature on those grounds.

avodonosov:
@jzheaux, I'd like to share 2 things about semantic versioning:

1. It doesn't prohibit backporting to version 4. (I'm not sure, but from
    your comment seems you imply semantic versioning prohibits it)
   The current release is 4.2.13, the fixed release will be 4.3.0 - all according to the quote you provide. 

2. Semantic versioning doesn't work for Java when we want an old version
    of a library to be used simultaneously with the new version. Simply because
    Java doesn't allow several versions of classes with an equal fully qualified name.

    How often we need several version simultaneously? Almost always.

    First, the dependency tree of any moderate size application includes some
    libraries in several tree nodes, with different versions.

    The second (more relevant for spring-security): migration easiness.
    We may want our old rest api endpoints to remain protected by spring-security 4,
    because it's already tested stable code. But for new endpoints, we'd like to use the modern,
    latest version 5.

    If this is not allowed, then we have a much more difficult "all or nothing" situation:
    unless you found resources to rewrite your existing code, you must continue
    using version 4 for old endpoints, thus bogging down deeper and deeper into the old version,
    which has limited or sometimes no maintenance (in some large system I know some
    components still use spring-security 3).

In order to allow several versions of your library in an application, give them different artifact name (spring-security3, spring-security4, spring-security5) and different package names.

commons-collections, commons-lang use this approach

jzheaux:
You are correct that this change could be made in a hypothetical 4.3.x, though no such release is planned.

I'm not certain what you are driving at with the rest of your comments, but if you'd like to pursue them further, I'd recommend you create a separate ticket as it feels like we are getting off-topic here. 

The decision to follow semantic versioning is a Spring-wide one, though, so if you are trying to make a case against it, you'd need to make it first with Spring Framework.

avodonosov:
Well, I have this point about semver for years, so the mention just triggered me to present it.

There is a valid thinking behind it, but it should be applied with understanding. Simultaneous loading of several versions is crucial. Javascript allows this for free - js packages (in all major package management solutions) do not have a global namespace to interfere (as java fully qualified class names). But js semver doesn't translate to Java so directly and mechanically as people think. For Java the apache common's approach is the best.

Sorry, I'm continuing the offtopic. Also, I'm not ready to push the whole Spring project - I'm afraid it will require more energy than I have and  I'm not sure of success with whatever energy. But I decided to mentions it as you referred semver.

Issue #7060, opened 2019-07-01
Multipartfile request with no authentication is still consumed even after an AccessDeniedException is thrown


a-sayyed:
### Summary
As explained by @wilkinsona in this related issue: https://github.com/spring-projects/spring-boot/issues/17345, a multipartfile request with no authentication to a secure endpoint, results in an `AccessDeniedException` (when the `HiddenHttpMethodFilter` is disabled). This is handled in `ExceptionTranslationFilter.handleSpringSecurityException(HttpServletRequest, HttpServletResponse, FilterChain, RuntimeException)` which results in the creation of a `DefaultSavedRequest`. This calls `javax.servlet.ServletRequest.getParameterMap()` which causes the multipart request to be consumed and parsed.

### Actual Behavior
Multipart file request with no Authentication to a secure endpoint results in the request being parsed and consumed anyway, then the client gets a `401 Unauthorized`

### Expected Behavior
Multipart file request with no Authentication to a secure endpoint should result in the request not being parsed or consumed, and the client gets a `401 Unauthorized` as soon as an  `AccessDeniedException` is thrown.

### Configuration
Please see the attached sample. You will need to add the property `spring.mvc.hiddenmethod.filter.enabled=false` to the `application.properties` file

### Version
2.1.6.RELEASE

### Sample
https://github.com/a-sayyed/spring-jetty-secure-multipartfile-upload-bug
![2019-06-28_09h29_33](https://user-images.githubusercontent.com/27724355/60420483-2828ef00-9be8-11e9-97da-c7cc969e1299.png)

rwinch:
Thanks for the report @a-sayyed!

I'm not sure this can be avoided all together. This is a common issue for multipart requests. For example, [CSRF needs to process the parameters](https://docs.spring.io/spring-security/site/docs/current/reference/htmlsingle/#csrf-multipart) to determine if the request is valid.

So I can better understand your issue, can you please explain what specific problem this is causing you? I understand a file can be uploaded, but this doesn't seem any more of a problem than a user can make requests to the server (the difference is the file is written to disk vs just being in memory).

a-sayyed:
Thank you for the explanation @rwinch!
My problem is exactly what you mentioned, I understand that this is similar to normal "bad requests", however this can quite easily result in a DDoS attack. Is it not possible to move the Multipartfilter behind the security filter chain? would that solve this issue?

rwinch:
[...]

PR #7167, opened 2019-07-31
ignore Multipart requests in HttpSessionRequestCache.requestMatcher

a-sayyed:
references: https://github.com/spring-projects/spring-security/issues/7060

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/RequestCacheConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/RequestCacheConfigurer.java
@@ -142,21 +142,11 @@ public final class RequestCacheConfigurer<H extends HttpSecurityBuilder<H>> exte
 			return null;
 		}
 	}
+
 	@SuppressWarnings("unchecked")
 	private RequestMatcher createDefaultSavedRequestMatcher(H http) {
-		ContentNegotiationStrategy contentNegotiationStrategy = http
-				.getSharedObject(ContentNegotiationStrategy.class);
-		if (contentNegotiationStrategy == null) {
-			contentNegotiationStrategy = new HeaderContentNegotiationStrategy();
-		}
-
 		RequestMatcher notFavIcon = new NegatedRequestMatcher(new AntPathRequestMatcher(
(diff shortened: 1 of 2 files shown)


a-sayyed:
I have a question regarding testing, since I'm not very familiar with the test setup.
I found the `RequestCacheConfigurerTests` class, and thought about writing an extra test method,
something like:

```
// SEC-7060
@Test
public void postWhenRequestIsMultipartThenRequestShouldNotBeCached() throws Exception {
	this.spring.register(DefaultSecurityConfig.class, MultipartFileController.class).autowire();
	MockMultipartFile aFile = new MockMultipartFile("aFile", "aFile", MediaType.TEXT_PLAIN_VALUE, 
RandomStringUtils.random(1024).getBytes());

	this.mvc.perform(multipart("/upload")
		.file(aFile))
		.andDo(print())
		.andExpect(status().isUnauthorized())
		.andReturn();
	}

@Controller
static class MultipartFileController {
	@PostMapping("/upload")
	public ResponseEntity<String> upload(MultipartFile multipartFile){
		return ResponseEntity.ok().build();
	}
}
```
I have a few questions:
- is this the way to go?
- do I need to register the controller, or it doesn't matter in this context?
- how do I really check if the request parameters are not cached?

rwinch:
See if this test helps https://github.com/spring-projects/spring-security/blob/5.2.0.M4/config/src/test/java/org/springframework/security/config/annotation/web/configurers/RequestCacheConfigurerTests.java#L126

a-sayyed:
> Thanks for working this issue. See if this test helps https://github.com/spring-projects/spring-security/blob/5.2.0.M4/config/src/test/java/org/springframework/security/config/annotation/web/configurers/RequestCacheConfigurerTests.java#L126
> 
> The gist is:
> 
>     * make a multipart request before authenticating and save that session object
> 
>     * Authenticate with the session object from before
> 
>     * ensure you are not redirected to the URL from the multipart request

Thanks @rwinch for the hint, I have added the test as you suggested.

rwinch:
Thanks for the PR @a-sayyed! This is now merged into master

Issue #7157, opened 2019-07-26
ServerRequestCacheWebFilter causes WebSession to be read every request


rwinch:
### Summary
Currently the `ServerRequestCacheWebFilter` causes the `WebSession` to be read on every request. We should be able to limit this to only happening after authentication success happens.

rwinch:
One way we can do this is to cache the request in a cookie rather than the session

jzheaux:
While this is a nice feature to be able to cache the request in a cookie, is it still the case that if an application wants to cache the request in a session, the `WebSession` will be hit on every request? Is this something that we should address in a separate ticket?

rwinch:
The only place WebFlux support requires reading the session on every request is the request cache. This means that moving this to use a cookie will solve the problem.

rwinch:
This is important for the improvements to session management improvements

rwinch:
Closed via 28c0d1459cab448f1616831b60379393931a8586

PR #7171, opened 2019-08-01
Cache request URI in cookie instead of session

eleftherias:
Fixes: gh-7157

### DIFF
--- a/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
@@ -152,7 +152,7 @@ import org.springframework.security.web.server.header.XXssProtectionServerHttpHe
 import org.springframework.security.web.server.savedrequest.NoOpServerRequestCache;
 import org.springframework.security.web.server.savedrequest.ServerRequestCache;
 import org.springframework.security.web.server.savedrequest.ServerRequestCacheWebFilter;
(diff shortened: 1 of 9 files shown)


jzheaux:
@eleftherias thanks for this PR - it doesn't sound much like the description in #7157. Would you mind adding some of your thinking that led you to this conclusion? Maybe add it as a comment in the original ticket.

rwinch:
@jzheaux There was some offline discussion about this between @eleftherias. Since she is on PTO I went ahead and added a comment to the original issue

jzheaux:
One thing to think about here is that there might be end-users who are affected by this change during an upgrade.

For example, let's say that an end-user was in the middle of logging in when the the upgrade happened. Their remembered URI could be in the session, but when Spring Security tries to do the redirect after login, it will now look for it in a cookie (and fail to find it).

I believe the impact is probably minor, but let's weigh it before merging. Also, we should add a release note about it.

If it appears the impact will be large, it might be possible to introduce a strategy that consults multiple request caches, if we wanted to try and be passive.

rwinch:
I think perhaps we should split this into two items. One introducing the cookie cache and then separately consider making it it a default.

eleftherias:
I create gh-8033 for the server cookie cache.

rwinch:
Closing this in favor of  gh-8033 for support of storing in a cookie. I don't think we should make storing in a Cookie as a default as it isn't as secure as using the session. Instead, I think we should proceed by making the request unique (i.e. add a query parameter) so it is only looked up on log in success.

PR #7176, opened 2019-08-01
Use org.mockito.ArgumentMatchers in favor of org.mockito.Matchers

larsgrefer:


jzheaux:
Thanks for another great cleanup, @larsgrefer! This is now merged into `master`.

PR #7174, opened 2019-08-01
Remove the unused emma plugin

larsgrefer:

PR #7173, opened 2019-08-01
Migrate VersionsResourceTasks groovy->java

larsgrefer:
Issue: gh-4939

PR #7077, opened 2019-07-04
Migrate TrangPlugin groovy->java

larsgrefer:
Issue: gh-4939

PR #7183, opened 2019-08-05
Use the 'io.freefair.aspectj' gradle plugin

larsgrefer:
After I've migrated (or removed) the other gradle plugins in #7077 #7173 and #7174 the AspectJ plugin was the last remaining one. Instead of migrating the current groovy based plugin to java, I just used the AspectJ gradle plugin I've already written last year.


rwinch:
Thanks for the PR @larsgrefer! We'd love to rely on proper AspectJ Gradle plugin. However, it appears this PR broke the build. Can you take a look at that?

larsgrefer:
@rwinch The build is fixed now.

rwinch:
Is there a reason that we need to move the source code into a different source folder? I'd prefer to keep the code in the src/main/java folder.

larsgrefer:
The `io.freefair.aspectj` plugin uses `src/main/aspectj` as source root for the code to be compiled by `ajc` by default. (The same way like `src/main/groovy` would contain code for `groovyc` and `src/main/kotlin` would contain the code compiled by `kotlinc`).

If you like, this can be changed in the build script:
```gradle
sourceSets.main.aspectj.srcDir "src/main/java"
```

Another option would be the `io.freefair.aspectj.post-compile-weaving` plugin.
This plugin leaves the actual compilation to javac (or groovyc, kotlinc, ...) and only performs the weaving as bytecode-to-bytecode transformation (by passing the compiler output into ajc's `--inpath`)

rwinch:
I think I'd like to use `sourceSets.main.aspectj.srcDir "src/main/java"`. My reasoning is that while it makes sense from a Gradle plugin perspective that aspectj code is place in it's own source folder it feels a bit invasive.

Spring allows switching between proxy and compile time aspects with just configuration. It is a bit disruptive to switch the folder that the annotated Java code is in.

larsgrefer:
> I think I'd like to use sourceSets.main.aspectj.srcDir "src/main/java". My reasoning is that while it makes sense from a Gradle plugin perspective that aspectj code is place in it's own source folder it feels a bit invasive.

For the `aspects` project this is actually the only option because its aspect is an `*.aj` file, which has to be compiled by `ajc`.

For `samples/javaconfig/aspectj` and `samples/xml/aspectj` I've chosen the `post-compile-weaving` approach because it's even less invasive and provides better integration:
 - The `post-compile-weaving` plugin can also weave groovy, scala and kotlin classes
 - Since the acutal compilation is perfomed by `javac`, additional tooling like lombok could be used without problems.

rwinch:
Thanks for the PR @larsgrefer! This is now merged into master

PR #7219, opened 2019-08-05
Use try-with-resources instead of try-finally

larsgrefer:


rwinch:
Thanks for the PR @larsgrefer This is now merged into master

Issue #7228, opened 2019-08-06
RequestContextSubscriber could put null value in Reactor Context


robotmrv:
### Summary
There are several issues with current implementation of ServletOAuth2AuthorizedClientExchangeFilterFunction.RequestContextSubscriber introduced by  https://github.com/spring-projects/spring-security/pull/6526

### Actual Behavior
1. RequestContextSubscriber could populate reactor Context with null values - that is forbidden since https://github.com/reactor/reactor-core/pull/1800 (it is possible if Flux or Mono was subscribed outside of WebRequest or security context)
2. RequestContextSubscriber replaces source context and does not respect previous values (possibly from other library) 
3. It does not create holder for its values in a context as it was suggested in javadoc
```
 * Note that contexts are optimized for low cardinality key/value storage, and a user
 * might want to associate a dedicated mutable structure to a single key to represent his
 * own context instead of using multiple {@link #put}, which could be more costly.
 * Past five user key/value pair, the {@link Context} will use a copy-on-write
 * implementation backed by a new {@link java.util.Map} on each {@link #put}.
```
https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/util/context/Context.java#L36-L44

subscriber creates 4 keys (occupies almost all "optimmized" implementations of Context) 
and as far as I understand `CONTEXT_DEFAULTED_ATTR_NAME` key is just a marker  

Maybe it would be better to put all this values into some holder so you can reduce number of keys from 4 to 1. And it allows remove extra `CONTEXT_DEFAULTED_ATTR_NAME`   

### Expected Behavior
1. RequestContextSubscriber should not populate Context with null values
2. RequestContextSubscriber should not replace previous Context 
3. RequestContextSubscriber should create DataHolder for its nullable data and do not "pollute" Context

### Version
5.1.6.RELEASE
5.2.0.M4

robotmrv:
Hi, @jgrandja  
will this fix be backported to 5.1.x?

jgrandja:
Thanks for the reminder @robotmrv! I just backported.

PR #7235, opened 2019-08-07
Fix NPE in RequestContextSubscriber

robotmrv:
RequestContextSubscriber could cause NPE if Mono/Flux.subscribe
was invoked outside of Web Context.
In addition it replaced source Context with its own without respect
to old data.
Now Request Context Data is Propagated within holder class and
it is added to existing reactor Context

Fixes gh-7228

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
@@ -19,6 +19,7 @@ package org.springframework.security.oauth2.client.web.reactive.function.client;
 import org.reactivestreams.Subscription;
 import org.springframework.beans.factory.DisposableBean;
 import org.springframework.beans.factory.InitializingBean;
+import org.springframework.lang.Nullable;
 import org.springframework.security.authentication.AnonymousAuthenticationToken;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.GrantedAuthority;
@@ -95,6 +96,7 @@ import java.util.function.Consumer;
  *
  * @author Rob Winch
  * @author Joe Grandja
+ * @author Roman Matiushchenko
  * @since 5.1
  * @see OAuth2AuthorizedClientManager
  */
@@ -174,7 +176,7 @@ public final class ServletOAuth2AuthorizedClientExchangeFilterFunction
 
 	@Override
 	public void afterPropertiesSet() throws Exception {
-		Hooks.onLastOperator(REQUEST_CONTEXT_OPERATOR_KEY, Operators.lift((s, sub) -> createRequestContextSubscriber(sub)));
+		Hooks.onLastOperator(REQUEST_CONTEXT_OPERATOR_KEY, Operators.liftPublisher((s, sub) -> createRequestContextSubscriberIfNecessary(sub)));
 	}
 
 	@Override
@@ -378,14 +380,22 @@ public final class ServletOAuth2AuthorizedClientExchangeFilterFunction
 	}
 
 	private void populateRequestAttributes(Map<String, Object> attrs, Context ctx) {
-		if (ctx.hasKey(HTTP_SERVLET_REQUEST_ATTR_NAME)) {
-			attrs.putIfAbsent(HTTP_SERVLET_REQUEST_ATTR_NAME, ctx.get(HTTP_SERVLET_REQUEST_ATTR_NAME));
-		}
-		if (ctx.hasKey(HTTP_SERVLET_RESPONSE_ATTR_NAME)) {
-			attrs.putIfAbsent(HTTP_SERVLET_RESPONSE_ATTR_NAME, ctx.get(HTTP_SERVLET_RESPONSE_ATTR_NAME));
-		}
-		if (ctx.hasKey(AUTHENTICATION_ATTR_NAME)) {
-			attrs.putIfAbsent(AUTHENTICATION_ATTR_NAME, ctx.get(AUTHENTICATION_ATTR_NAME));
+		RequestContextDataHolder holder = RequestContextSubscriber.getRequestContext(ctx);
+		if (holder != null) {
+			HttpServletRequest request = holder.getRequest();
+			if (request != null) {
+				attrs.putIfAbsent(HTTP_SERVLET_REQUEST_ATTR_NAME, request);
+			}
+
+			HttpServletResponse response = holder.getResponse();
+			if (response != null) {
+				attrs.putIfAbsent(HTTP_SERVLET_RESPONSE_ATTR_NAME, response);
+			}
+
+			Authentication authentication = holder.getAuthentication();
+			if (authentication != null) {
+				attrs.putIfAbsent(AUTHENTICATION_ATTR_NAME, authentication);
+			}
 		}
 	}
 
@@ -472,7 +482,7 @@ public final class ServletOAuth2AuthorizedClientExchangeFilterFunction
 					.build();
 	}
 
-	private <T> CoreSubscriber<T> createRequestContextSubscriber(CoreSubscriber<T> delegate) {
+	<T> CoreSubscriber<T> createRequestContextSubscriberIfNecessary(CoreSubscriber<T> delegate) {
 		HttpServletRequest request = null;
 		HttpServletResponse response = null;
 		ServletRequestAttributes requestAttributes =
@@ -482,6 +492,10 @@ public final class ServletOAuth2AuthorizedClientExchangeFilterFunction
 			response = requestAttributes.getResponse();
 		}
 		Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+		if (authentication == null && request == null && response == null) {
+			//do not need to create RequestContextSubscriber with empty data
+			return delegate;
+		}
 		return new RequestContextSubscriber<>(delegate, request, response, authentication);
 	}
 
@@ -553,34 +567,37 @@ public final class ServletOAuth2AuthorizedClientExchangeFilterFunction
 		}
 	}
 
-	private static class RequestContextSubscriber<T> implements CoreSubscriber<T> {
-		private static final String CONTEXT_DEFAULTED_ATTR_NAME = RequestContextSubscriber.class.getName().concat(".CONTEXT_DEFAULTED_ATTR_NAME");
+	static class RequestContextSubscriber<T> implements CoreSubscriber<T> {
+		static final String REQUEST_CONTEXT_DATA_HOLDER =
+				RequestContextSubscriber.class.getName().concat(".REQUEST_CONTEXT_DATA_HOLDER");
 		private final CoreSubscriber<T> delegate;
-		private final HttpServletRequest request;
-		private final HttpServletResponse response;
-		private final Authentication authentication;
+		private final Context context;
 
-		private RequestContextSubscriber(CoreSubscriber<T> delegate,
-											HttpServletRequest request,
-											HttpServletResponse response,
-											Authentication authentication) {
+		RequestContextSubscriber(CoreSubscriber<T> delegate,
+								HttpServletRequest request,
+								HttpServletResponse response,
+								Authentication authentication) {
 			this.delegate = delegate;
-			this.request = request;
-			this.response = response;
-			this.authentication = authentication;
+
+			Context parentContext = this.delegate.currentContext();
+			Context context;
+			if (parentContext.hasKey(REQUEST_CONTEXT_DATA_HOLDER)) {
+				context = parentContext;
+			} else {
+				context = parentContext.put(REQUEST_CONTEXT_DATA_HOLDER, new RequestContextDataHolder(request, response, authentication));
+			}
+
+			this.context = context;
+		}
+
+		@Nullable
+		private static RequestContextDataHolder getRequestContext(Context ctx) {
+			return ctx.getOrDefault(REQUEST_CONTEXT_DATA_HOLDER, null);
 		}
 
 		@Override
 		public Context currentContext() {
-			Context context = this.delegate.currentContext();
-			if (context.hasKey(CONTEXT_DEFAULTED_ATTR_NAME)) {
-				return context;
-			}
-			return Context.of(
-					CONTEXT_DEFAULTED_ATTR_NAME, Boolean.TRUE,
-					HTTP_SERVLET_REQUEST_ATTR_NAME, this.request,
-					HTTP_SERVLET_RESPONSE_ATTR_NAME, this.response,
-					AUTHENTICATION_ATTR_NAME, this.authentication);
+			return this.context;
 		}
 
 		@Override
@@ -601,6 +618,35 @@ public final class ServletOAuth2AuthorizedClientExchangeFilterFunction
 		@Override
 		public void onComplete() {
 			this.delegate.onComplete();
+		}
+	}
+
+	static class RequestContextDataHolder {
+		private final HttpServletRequest request;
+		private final HttpServletResponse response;
+		private final Authentication authentication;
+
+		RequestContextDataHolder(@Nullable HttpServletRequest request,
+								@Nullable HttpServletResponse response,
+								@Nullable Authentication authentication) {
+			this.request = request;
+			this.response = response;
+			this.authentication = authentication;
+		}
+
+		@Nullable
+		private HttpServletRequest getRequest() {
+			return this.request;
+		}
+
+		@Nullable
+		private HttpServletResponse getResponse() {
+			return this.response;
+		}
+
+		@Nullable
+		private Authentication getAuthentication() {
+			return this.authentication;
 		}
 	}
 }
--- a/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunctionTests.java
+++ b/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunctionTests.java
@@ -72,6 +72,10 @@ import org.springframework.web.context.request.ServletRequestAttributes;
 import org.springframework.web.reactive.function.BodyInserter;
 import org.springframework.web.reactive.function.client.ClientRequest;
 import org.springframework.web.reactive.function.client.WebClient;
+import reactor.core.CoreSubscriber;
+import reactor.core.publisher.BaseSubscriber;
+import reactor.core.publisher.Mono;
+import reactor.util.context.Context;
 
 import java.net.URI;
 import java.time.Duration;
@@ -84,6 +88,7 @@ import java.util.Optional;
 import java.util.function.Consumer;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
 import static org.assertj.core.api.AssertionsForClassTypes.assertThatThrownBy;
 import static org.mockito.Mockito.*;
 import static org.springframework.http.HttpMethod.GET;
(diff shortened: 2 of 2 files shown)


pivotal-issuemaster:
@robotmrv Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=7235)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=7235) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@robotmrv Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=7235)!

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunctionTests.java
```diff
@@ -635,6 +642,58 @@ public void filterWhenRequestAttributesNotSetAndHooksInitHooksResetThenDefaultsN
 		assertThat(getBody(request)).isEmpty();
 	}
 
+	// gh-7228
+	@Test
+	public void publisherWhenHooksInitAndSubscribedOutsideOfWebContextThenShouldNotThrowNpe() throws Exception {
+		Mono<Context> publisher = Mono.subscriberContext();
+		this.function.afterPropertiesSet();			// Hooks.onLastOperator() initialized
+//		this.function.destroy();					// Hooks.onLastOperator() released
```
rwinch: Is there a reason that the tests do not invoke .destroy()?

robotmrv: leave it here just to mark that Hook is not released.  
but maybe it is better to move `this.function.destroy();` to the `cleanup()`

robotmrv:
I have doubts about  `ServletOAuth2AuthorizedClientExchangeFilterFunction.createRequestContextSubscriber()`  
It creates RequestContextSubscriber on every subscribe()/block() even outside of Security or WebContext and propagates empty DataHolder.  
Maybe it would be better to create RequestContextSubscriber only in case `if (authentication != null || requestAttributes != null)` ?

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunctionTests.java
robotmrv: moved `this.function.destroy()` to the `cleanup()` method

rwinch:
@robotmrv - It looks like something needs to change to me as well. The ThreadLocal should only be invoked inside of defaultRequest which is guaranteed to be invoked on the original Servlet container Thread. I'm going to let @jgrandja take a look at this since he worked on some of the cleanup in this class.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunctionTests.java
```diff
@@ -635,6 +643,101 @@ public void filterWhenRequestAttributesNotSetAndHooksInitHooksResetThenDefaultsN
 		assertThat(getBody(request)).isEmpty();
 	}
 
+	// gh-7228
+	@Test
+	public void publisherWhenHooksInitAndSubscribedOutsideOfWebContextThenShouldNotThrowNpe() throws Exception {
```
jgrandja: If I comment out the following in `createRequestContextSubscriber()`

```
//		if (authentication == null && request == null && response == null) {
//			return delegate;
//		}
```

....the test still passes. I'm not sure this test makes sense?

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunctionTests.java
```diff
@@ -635,6 +643,101 @@ public void filterWhenRequestAttributesNotSetAndHooksInitHooksResetThenDefaultsN
...
+		Mono<Context> publisher = Mono.subscriberContext();
+		this.function.afterPropertiesSet();			// Hooks.onLastOperator() initialized
+
+		publisher.block();
+	}
+
+	// gh-7228
+	@Test
+	public void publisherWhenHooksInitAndSubscribedOutsideOfWebContextAndOutsideOfSecurityContextAndWithParentContextThenShouldNotReplaceParentContextAndShouldNotAddRequestContext() throws Exception {
```
jgrandja: I'm not sure about this test and the next one. It's quite low level and not directly testing the API of `ServletOAuth2AuthorizedClientExchangeFilterFunction`. Most tests are directly testing `filter()` and  `defaultRequest()`, whereas these 2 tests are directly using the Reactor API's.

The preference would be to change these tests so they are aligned with the existing `filter()` tests.

robotmrv: The main issue was that current instrumentation affects all application   
So this tests check if it is possible to performe operations on Publishers without web/sucurity context with initialized reactor instrumentation (which is ServletOAuth2AuthorizedClientExchangeFilterFunction functionality)

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunctionTests.java
robotmrv: because this test does not checks this part (publisherWhenHooksInitAndSubscribedOutsideOfWebContextAndOutsideOfSecurityContextAndWithParentContextThenShouldNotReplaceParentContextAndShouldNotAddRequestContext() tests it)   
The check you commented is intended to skip reactor context population with empty RequestContextDataHolder 
 values as it is not useful data (especially outside of web context). This is just optimization ( here my motivation https://github.com/spring-projects/spring-security/pull/7235#issuecomment-520127370 )

robotmrv: NPE is fixed by introducing RequestContextDataHolder which could contain null values,   
It also reduces number of occupied keys in reactor context

robotmrv:
@jgrandja  I've fixed codestyle according to convention standard and removed `publisherWhenHooksInitAndSubscribedOutsideOfWebContextThenShouldNotThrowNpe()`  
Regarding other tests:  
- They check reactor Hook which is part of the ServletOAuth2AuthorizedClientExchangeFilterFunction 
- `RequestContextSubscriber` is low level implementation which should be checked 
- these tests cover scenarios from gh-7228 affect of global reactor hook on whole application - it should not affect unrelated functionality (e.g. reactive DB init on application startup etc.)
- I do not know how to test it better  
so I leave them as is  
If you still have concerns about them please let me know.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -484,6 +494,9 @@ private ClientRequest bearer(ClientRequest request, OAuth2AuthorizedClient autho
 			response = requestAttributes.getResponse();
 		}
 		Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+		if (authentication == null && request == null && response == null) {
+			return delegate;
```
jgrandja: This is misleading since it doesn't create and return a `RequestContextSubscriber` instance, but instead returns the "current" `Subscriber`. I think we should return `null` for these cases and ensure `Hooks.onLastOperator` is NOT called in `afterPropertiesSet()`. Please add a test for this as well.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunctionTests.java
```diff
@@ -146,9 +153,10 @@ public void setup() {
 	}
 
 	@After
-	public void cleanup() {
+	public void cleanup() throws Exception {
```
jgrandja: Please revert the 2 changes in `cleanup()` as I believe it was added as a result of the removed test.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunctionTests.java
jgrandja: Both of these tests do not directly test the API of `ServletOAuth2AuthorizedClientExchangeFilterFunction` but instead are testing internal implementation details. In order to simplify the tests, please make `RequestContextSubscriber` package-private along with the constructor. This will allow you to test the constructed state of `RequestContextSubscriber` using `currentContext()`. Please remove these 2 tests after you have added the new ones. Thanks.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunctionTests.java
robotmrv: No.  
It removes reactor Hook from the filter,  i.e. makes cleanup of the global state.   
So it should be better here.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
robotmrv: Hooks.onLastOperator() - applies global lift function on `subscribe()/block()` and it should be better installed once per application life circle.  
As Lift function is invoked only at the moment of `subscribe()` lazily (not at the moment of application context / Filter initialization) it is not possible to unsure is there web/security context or not.  

Not sure we are on the same page with case I try to fix  
One of the cases:  
I want to initialize application with some data. I have my reactive service API to do it. 
I make `ApplicationRunner` or just Application listener which invokes my reactive API on startup and call subscribe()/block(). 
So If my API implementation uses reactor context (e.g. reactive transactions, or just some custom implementation) with initialized hook from filter and without current fix I will get NPE, because `RequestContextSubscriber` will populate reactor context with `null` values (as there is no Web and Security Context) and this is forbidden.  

Another scenario - my reactive API is invoked form RabbitMQ listener and there is no web context - as the result NPE.  

So what PR fixes:  
1. it ensures that no `null` values are propagated to the reactor context.  
2. As reactor context recommended to be as small as possible (see javadocs) this fix reduces number of used entries from 4 to 1/0.  
3. `RequestContextSubscriber` replaced original context so data from other hooks could be lost. This fix ensures that `RequestContextSubscriber` will not replace all parent context but adds values to it.  
4. In described case `RequestContextDataHolder` will be created with `null` values and will be useless as it is not required at all for the operation (application initialization), so it is make sense not to populate reactor context with useless data because of p.2 and not to create additional `Subscriber` to reduce allocation and invocation stack size on each emit or `currentContext()`

Lift function can't return `null` as later we will get NPE so `delegate` subscriber looks fine to me as it means NOOP lift function.   

@jgrandja, do you still want to return `null` from lift function and create `RequestContextSubscriber` in any case with empty `RequestContextDataHolder`?

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunctionTests.java
robotmrv: as this PR fixes scenario without web/security context (see https://github.com/spring-projects/spring-security/pull/7235#discussion_r316789550 ) and in this scenario `ExchangeFilterFunction` API of `ServletOAuth2AuthorizedClientExchangeFilterFunction` is not used  
but `ServletOAuth2AuthorizedClientExchangeFilterFunction#afterPropertiesSet` is used and it is part of API it is makes sense to test these scenarios without usage of `ExchangeFilterFunction` part of API.  

Other scenarios with direct usage of `ExchangeFilterFunction` API are covered by existing tests. 
I will make `RequestContextSubscriber` package-private but test will not change very much as they could not use `ExchangeFilterFunction` API by scenario.   

These test could be moved to a new test class but I do not see point in this as they still test `ServletOAuth2AuthorizedClientExchangeFilterFunction#afterPropertiesSet` API.

robotmrv:
@jgrandja , 
I've updated tests, please check

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunctionTests.java
```diff
@@ -635,6 +641,69 @@ public void filterWhenRequestAttributesNotSetAndHooksInitHooksResetThenDefaultsN
 		assertThat(getBody(request)).isEmpty();
 	}
 
+	// gh-7228
+	@Test
+	public void afterPropertiesSetWhenHooksInitAndPublisherSubscribedOutsideOfWebContextAndOutsideOfSecurityContextThenSubscriptionShouldNotThrowException() throws Exception {
```
jgrandja: Let's shorten this name to `afterPropertiesSetWhenHooksInitAndOutsideWebSecurityContextThenShouldNotThrowException`

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunctionTests.java
```diff
@@ -635,6 +641,69 @@ public void filterWhenRequestAttributesNotSetAndHooksInitHooksResetThenDefaultsN
...
+		this.function.afterPropertiesSet();			// Hooks.onLastOperator() initialized
+		assertThatCode(() -> Mono.subscriberContext().block())
+				.as("RequestContext Hook brakes application outside of web/security context")
+				.doesNotThrowAnyException();
+	}
+
+	@Test
+	public void createRequestContextSubscriberIfNecessaryWhenOutsideOfWebContextAndOutsideOfSecurityContextThenShouldReturnOriginalSubscriber() throws Exception {
```
jgrandja: Please rename to `createRequestContextSubscriberIfNecessaryWhenOutsideWebSecurityContextThenReturnOriginalSubscriber`

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunctionTests.java
```diff
@@ -635,6 +641,69 @@ public void filterWhenRequestAttributesNotSetAndHooksInitHooksResetThenDefaultsN
...
+		BaseSubscriber<Object> originalSubscriber = new BaseSubscriber<Object>() {};
+		CoreSubscriber<Object> resultSubscriber = this.function.createRequestContextSubscriberIfNecessary(originalSubscriber);
+		assertThat(resultSubscriber).isSameAs(originalSubscriber);
+	}
+
+	// gh-7228
+	@Test
+	public void requestSubscriberWhenHasRequestAndHasResponseThenShouldAddToParentContextRequestContextDataHolder() throws Exception {
```
jgrandja: Please rename to `createRequestContextSubscriberWhenRequestResponseProvidedThenCreateWithParentContext`

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunctionTests.java
```diff
@@ -635,6 +641,69 @@ public void filterWhenRequestAttributesNotSetAndHooksInitHooksResetThenDefaultsN
...
+		testRequestContextSubscriber(new MockHttpServletRequest(), new MockHttpServletResponse(), null);
+	}
+
+	// gh-7228
+	@Test
+	public void requestSubscriberWhenHasAuthenticationThenShouldAddToParentContextRequestContextDataHolder() throws Exception {
```
jgrandja: Please rename to `createRequestContextSubscriberWhenAuthenticationProvidedThenCreateWithParentContext`

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunctionTests.java
```diff
@@ -635,6 +641,69 @@ public void filterWhenRequestAttributesNotSetAndHooksInitHooksResetThenDefaultsN
...
+		testRequestContextSubscriber(null, null, this.authentication);
+	}
+
```
jgrandja: Can you also add a test where `RequestContextSubscriber` constructor uses `parentContext`

```java
			if (parentContext.hasKey(REQUEST_CONTEXT_DATA_HOLDER)) {
				context = parentContext;

``

Comment (robotmrv) [reply oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunctionTests.java]: @jgrandja 
I've added test and renamed existing as you suggested.  
please check

File: oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunctionTests.java [hunk 11]
```diff
@@ -643,32 +643,53 @@ public void filterWhenRequestAttributesNotSetAndHooksInitHooksResetThenDefaultsN
...
 	// gh-7228
 	@Test
-	public void requestSubscriberWhenHasAuthenticationThenShouldAddToParentContextRequestContextDataHolder() throws Exception {
+	public void createRequestContextSubscriberWhenAuthenticationProvidedThenCreateWithParentContext() throws Exception {
 		testRequestContextSubscriber(null, null, this.authentication);
 	}
 
+	@Test
+	public void createRequestContextSubscriberWhenParentContextHasDataHolderThenShouldReuseParentContext() throws Exception {
+		RequestContextDataHolder testValue = mock(RequestContextDataHolder.class);
```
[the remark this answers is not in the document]
robotmrv: @jgrandja done

jgrandja:
Thank you for all the updates @robotmrv. This is now in master.

PR #7240, opened 2019-08-08
Use List.toArray of 0

larsgrefer:


config/src/main/java/org/springframework/security/config/annotation/authentication/configurers/provisioning/JdbcUserDetailsManagerConfigurer.java
```diff
@@ -191,7 +191,7 @@ public JdbcUserDetailsManager getUserDetailsService() {
 
 	protected DatabasePopulator getDatabasePopulator() {
 		ResourceDatabasePopulator dbp = new ResourceDatabasePopulator();
-		dbp.setScripts(initScripts.toArray(new Resource[initScripts.size()]));
+		dbp.setScripts(initScripts.toArray(new Resource[0]));
```
rwinch: Can you explain why you would initialize it to a 0 length array? This means that `List` must now allocate another Array of the correct size which is more garbage that would need collected.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/JeeConfigurer.java
```diff
@@ -100,9 +101,7 @@ public JeeConfigurer() {
 	 */
 	public JeeConfigurer<H> mappableAuthorities(String... mappableRoles) {
 		this.mappableRoles.clear();
-		for (String role : mappableRoles) {
-			this.mappableRoles.add(role);
-		}
+		this.mappableRoles.addAll(Arrays.asList(mappableRoles));
```
rwinch: This allocates an additional object which is not preferred since it in turn will need garbage collected.

config/src/main/java/org/springframework/security/config/annotation/authentication/configurers/provisioning/JdbcUserDetailsManagerConfigurer.java
larsgrefer: IntelliJ says the following:
>In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow. However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version. Also passing pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray call which may result in extra nulls at the end of the array, if the collection was concurrently shrunk during the operation.

Other sources for this can be found here:
- https://stackoverflow.com/a/29444594/3574494
- https://shipilev.net/blog/2016/arrays-wisdom-ancients/

config/src/main/java/org/springframework/security/config/annotation/web/configurers/JeeConfigurer.java
larsgrefer: You're right. Unfortunately `HashSet` just uses looped `add` calls in it's `addAll` implementation.

Generally `addAll` has the benefit that the number of elements which will be added is known in advance, so the internal data structure can be resized to the correct number. Array-based collections like `ArrayList` can use `System.arraycopy` internally, while looped `add` calls may require multiple internal resizes.

larsgrefer:
`Collections.singletonList` just provides slightly better performance.

I can't tell if someone actually calls `set()` on those lists.

larsgrefer:
I've modified this PR to only contain one change instead of all three, so the changes can be discussed and merged/declined independently.

rwinch:
> Collections.singletonList just provides slightly better performance.
>
> I can't tell if someone actually calls set() on those lists.

Understood it can provide a slight improvement, but I'm not sure it is worth the risk of breaking users. As you said we can discuss that separately.

Thanks for the quick turnaround @larsgrefer! This is now merged into master

PR #7241, opened 2019-08-08
(fix) enable public access for DigestAuthUtils so it can be used

bevensteven:
I'm guessing there might be some reasons for why `DigestAuthUtils` isn't already publicly accessible, but decided to file this pull request in case it is doable.

Sometimes developers need access to lower level tools to create a digest for digest auth, this pull request addresses this by making the class publicly available for use. 

Some questions:
1. Does this expose any vulnerabilities?
2. If made public, do the signatures of any of the methods in this util need to change?

Thanks!


rwinch:
Thanks for the PR @bevensteven These are not intended to be public as we want to keep DigestAuthUtils encapsulated.

Issue #6254, opened 2018-12-07
Document BearerTokenResolver in reference


jgrandja:
There is no documentation for `BearerTokenResolver` and it's default implementation `DefaultBearerTokenResolver` in the reference. We should document and provide code samples on how to use and customize.

kostya05983:
Hello @jgrandja , I can take this one, If I understand correctly, I should add docs in docs module of spring security, shouldn't I? And where can I put code samples of this?

jzheaux:
@kostya05983 Thanks for volunteering - yes, you'd place docs in the docs module.

As for a sample, you could add one that goes alongside the other oauth2resourceserver samples, though there are quite a few already. Unless you feel customizing a bearer token resolver to be quite nuanced, I'd lean towards just sticking with snippets in the documentation.

Does that sound doable?

kostya05983:
@jgrandja , thanks for guide, yes, it do.

PR #7250, opened 2019-08-11
add documentation BearerTokenResolver with snippets

kostya05983:
Add docs to BearerTokenResolver with snippets, snippet is taken from existing code base. 
gh-6254

### DIFF
--- a/docs/manual/src/docs/asciidoc/_includes/reactive/oauth2/access-token.adoc
+++ b/docs/manual/src/docs/asciidoc/_includes/reactive/oauth2/access-token.adoc
@@ -33,3 +33,46 @@ SecurityWebFilterChain configure(ServerHttpSecurity http) throws Exception {
 ----
 
 You can now leverage Spring Security's <<webclient>> or <<webflux-roac,@RegisteredOAuth2AuthorizedClient>> support to obtain and use the access token.
+
+== BearerTokenResolver
+
+With interface BearerTokenResolver you can provide a strategy to resolve a bearer token.
+
+The interface provides the next method:
+
+[source,java]
+----
+/**
+* Resolve any <a href="https://tools.ietf.org/html/rfc6750#section-1.2" target="_blank">Bearer Token</a>
+* value from the request.
+*
+* @param request the request
+* @return the Bearer Token value or {@code null} if none found
+* @throws OAuth2AuthenticationException if the found token is invalid
+*/
+String resolve(HttpServletRequest request);
+----
+
+In code base, you can find two implementation of this interface:
+HeaderBearerTokenResolver and DefaultBearerTokenResolver (based on RFC 6750).
+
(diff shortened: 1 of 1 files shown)


kostya05983:
@jzheaux , please see it.

docs/manual/src/docs/asciidoc/_includes/reactive/oauth2/access-token.adoc
```diff
@@ -33,3 +33,46 @@ SecurityWebFilterChain configure(ServerHttpSecurity http) throws Exception {
 ----
 
 You can now leverage Spring Security's <<webclient>> or <<webflux-roac,@RegisteredOAuth2AuthorizedClient>> support to obtain and use the access token.
+
+== BearerTokenResolver
+
+With interface BearerTokenResolver you can provide a strategy to resolve a bearer token.
```
jzheaux: When referring to class names, and other inlined code, please use backticks, e.g. `BearerTokenResolver`

docs/manual/src/docs/asciidoc/_includes/reactive/oauth2/access-token.adoc
```diff
@@ -33,3 +33,46 @@ SecurityWebFilterChain configure(ServerHttpSecurity http) throws Exception {
...
+
+The interface provides the next method:
```
jzheaux: Instead of showing the interface, could you show example usage? Specifically, I'm thinking it would be nice for the documentation to show how to do form-based bearer tokens and using a custom header name.

kostya05983: @jzheaux , 1) Can you explain more about it? is It like Loosely Coupling? When we take parameters and generate a token?  2) We can provide a link to code with HeaderBearerTokenResolver.

jzheaux: Good questions. 

For the first, there is a setting, `DefaultBearerTokenResolver#setAllowFormEncodedBodyParameter`. The spec allows for passing the bearer token [via a form parameter](https://tools.ietf.org/html/rfc6750#section-2.2) instead of a header.

For the second (and in general), I recommend that the code be inlined, so the reader can see example usage in the context of your explanation. For example, you can imagine documentation written like this:

> Or, we may need to use a custom header name, like when using Google Cloud's IAP Proxy authentication:
> ```java
>    http
>        .oauth2ResourceServer()
>             .bearerTokenResolver(new HeaderBearerTokenResolver("x-goog-iap-jwt-assertion"))
>  ```

jzheaux:
@kostya05983 thanks for your contributions here! I've merged your PR via 69a4848850a2b9c93e5bd2d05681e4e9ffed748b, and then I realized that it needed to be in the servlet docs, so I moved it and added a couple of examples via 5d0e80c85cb28dad6e5532162466e5eb5d5c6319

PR #7252, opened 2019-08-11
Fix NPE in TokenBasedRememberMeServices with null UserDetails

### LINKED ISSUE #7251: TokenBasedRememberMeServices.processAutoLoginCookie (TokenBasedRememberMeServices.java:134) java.lang.NullPointerException
TokenBasedRememberMeServices assumes user exists when it could not for various reasons:

1. The user could have been deleted (Most likely)
2. The cookie could be invalid

https://github.com/spring-projects/spring-security/blob/master/web/src/main/java/org/springframework/security/web/authentication/rememberme/TokenBasedRememberMeServices.java#L123-L124

There is no check and such an exception is not handled gracefully. 
At the very least, it should throw a InvalidCookieException

codeconsole:
Fixes  #7251

Checks to see if the userService found a userDetails or not.  If the user has since been deleted, this will end up with an uncaught null pointer exception and result in the user not being able to visit the site.


eleftherias:
@codeconsole I've added your commit to master and followed up with a polish commit that adds the changes I mentioned.

PR #7254, opened 2019-08-12
Use UTF-8 for compilation

larsgrefer:
See: https://travis-ci.org/spring-projects/spring-security/builds/570999589#L1514-L1525

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,5 +12,5 @@ max_line_length = 120
 [*.{java,xml}]
 indent_style = tab
 indent_size = 4
-charset = latin1
+charset = utf-8
 continuation_indent_size = 8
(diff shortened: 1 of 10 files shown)


rwinch:
@larsgrefer Thanks for the PR. I think we should just change the file encoding to be UTF-9 rather than changing the encoding of the build. What are your thoughts? If you agree, do you think you could make the changes?

larsgrefer:
>  think we should just change the file encoding to be UTF-9 rather than changing the encoding of the build. What are your thoughts?

I'd definitely vote for UTF-8, but the CONTRIBUTING.md explicitly requests `ISO-8859-1` 

> If you agree, do you think you could make the changes?

I'll try 👍

rwinch:
> I'd definitely vote for UTF-8, but the CONTRIBUTING.md explicitly requests ISO-8859-1

I had missed that we had that documented to use ISO-8859-1.

Thanks for the quick turnaround!

PR #7256, opened 2019-08-13
Add setter for authority prefix in JwtGrantedAuthoritiesConverter

andifalk:
Fixes gh-7101

Prior to this change mapped authorities are always prefixed
with default value 'SCOPE_'. To change this default behaviour the
converter had to be replaced completely with a custom one.
This commit adds an additional setter to configure a custom
authority prefix like e.g. 'ROLE_'. Without specifying a custom prefix
the default prefix still remains 'SCOPE_'.
This way existing authorization checks using the standard 'ROLE_'
prefix can be reused without lots of effort.

Required CLA has been submitted as well

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverter.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverter.java
@@ -25,6 +25,7 @@ import org.springframework.core.convert.converter.Converter;
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.authority.SimpleGrantedAuthority;
 import org.springframework.security.oauth2.jwt.Jwt;
+import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
 /**
@@ -40,6 +41,8 @@ public final class JwtGrantedAuthoritiesConverter implements Converter<Jwt, Coll
 	private static final Collection<String> WELL_KNOWN_AUTHORITIES_CLAIM_NAMES =
 			Arrays.asList("scope", "scp");
 
+	private String authorityPrefix = DEFAULT_AUTHORITY_PREFIX;
+
 	/**
 	 * Extract {@link GrantedAuthority}s from the given {@link Jwt}.
 	 *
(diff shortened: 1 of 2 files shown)


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverter.java
```diff
@@ -50,11 +53,16 @@
 	public Collection<GrantedAuthority> convert(Jwt jwt) {
 		Collection<GrantedAuthority> grantedAuthorities = new ArrayList<>();
 		for (String authority : getAuthorities(jwt)) {
-			grantedAuthorities.add(new SimpleGrantedAuthority(DEFAULT_AUTHORITY_PREFIX + authority));
+			grantedAuthorities.add(new SimpleGrantedAuthority(authorityPrefix + authority));
 		}
 		return grantedAuthorities;
 	}
 
+	public void setAuthorityPrefix(String authorityPrefix) {
```
jzheaux: Will you please add some simple JavaDoc here?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverter.java
```diff
@@ -50,11 +53,16 @@
 	public Collection<GrantedAuthority> convert(Jwt jwt) {
 		Collection<GrantedAuthority> grantedAuthorities = new ArrayList<>();
 		for (String authority : getAuthorities(jwt)) {
-			grantedAuthorities.add(new SimpleGrantedAuthority(DEFAULT_AUTHORITY_PREFIX + authority));
+			grantedAuthorities.add(new SimpleGrantedAuthority(authorityPrefix + authority));
```
jzheaux: To conform with our team standard, would you please prefix the field with `this.`?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverter.java
andifalk: will add the suggested javadoc

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverter.java
andifalk: will change this :+1:

jzheaux:
Thanks for the PR, @andifalk! This is now merged into `master`.

Issue #7166, opened 2019-07-31
SessionAuthenticationStrategy make HttpSecurity.sessionManagement().maximumSessions(1) unavailability


HomminLee:
I am not good at English. So, please forgive my grammatical mistakes.

I have a problem while config `HttpSecurity.sessionManagement().maximumSessions(1)`. I hope a user only have one valid session at the same time.

 If I login at browser A，and then login at browser B. It's run perfect and session in browser A will invalid.

But if I login at browser A,  then login at browser A(
In other words, I'm called the `/login` interface twice in the same browser.), then login at browser B. I found both browsers can access the restricted interfaces.

---

I viewed source code, found there have three `SessionAuthenticationStrategy`: `ConcurrentSessionControlAuthenticationStrategy`, `ChangeSessionIdAuthenticationStrategy`, `RegisterSessionAuthenticationStrategy`. I tried to debug with source code and the following happened:

While the second time login at browser A, `ConcurrentSessionControlAuthenticationStrategy` find the same sessionId, so it do noting. But `ChangeSessionIdAuthenticationStrategy` change sessionId into a new sessionId, and `RegisterSessionAuthenticationStrategy` think it's a new session and store. so there is two session in session registry.

Did I miss some configuration to resolve this problem?

My spring security version is 5.1.5.RELEASE. Think you very much!

rwinch:
> But if I login at A browser, then login at A browser, then login at B browser. I found two session is both valid.

Can you explain by what you mean by this? How do you login A browser and login A browser again?

HomminLee:
Thank you for taking the time to look at my issue.I tried to complete my question.

rwinch:
Thank you. Can you please provide a minimal sample to reproduce the issue?

HomminLee:
I created a simple project to verification my problem, and get the same happening. It only included a security configuration and a rest controller. I upload my code at [https://github.com/HomminLee/security-session-demo](https://github.com/HomminLee/security-session-demo). Also, I write down the steps in README.md.

eleftherias:
Thank you for the sample and detailed explanation.
We are working on a fix for this, but in the meantime there are 2 options you can use to prevent this problem from happening.

Option 1
You can prevent any additional users from logging in when the maximum number of sessions has been reached, by using `maxSessionsPreventsLogin(true)`.
You configuration would look like this.
```
http
    .sessionManagement()
        .maximumSessions(1)
            .maxSessionsPreventsLogin(true);
```

[...]

PR #7258, opened 2019-08-14
Expire as many sessions as exceed maximum allowed

eleftherias:
Fixes: gh-7166

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/session/ConcurrentSessionControlAuthenticationStrategy.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/session/ConcurrentSessionControlAuthenticationStrategy.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,6 +15,7 @@
  */
 package org.springframework.security.web.authentication.session;
 
+import java.util.Comparator;
 import java.util.List;
 
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/springframework/security/web/authentication/session/ConcurrentSessionControlAuthenticationStrategy.java
```diff
@@ -156,18 +158,12 @@ protected void allowableSessionsExceeded(List<SessionInformation> sessions,
...
-		for (SessionInformation session : sessions) {
-			if ((leastRecentlyUsed == null)
-					|| session.getLastRequest()
-							.before(leastRecentlyUsed.getLastRequest())) {
-				leastRecentlyUsed = session;
-			}
+		// Determine least recently used sessions, and mark them for invalidation
+		sessions.sort(Comparator.comparing(SessionInformation::getLastRequest));
+		int maximumSessionsExceededBy = sessions.size() - allowableSessions + 1;
+		for (int i = 0; i < maximumSessionsExceededBy; i++) {
```
rwinch: `sessions` is defined as a `List` which means we need to treat this as though it could be any implementation of a `List`. If it were a `LinkedList`, then iterating over the `List` using `List.get(int)` is very inefficient.

Instead of using `List.get(int)` we should use the iterator (i.e. `for (SessionInformation session : sessions)`). A few options that use an Iterator are:

- keep a separate variable for determining which sessions should be expired and break out of the `for` loop if that is met. 
- produce [subList](https://docs.oracle.com/javase/8/docs/api/java/util/List.html#subList-int-int-) of sessions to expire and iterate over those

eleftherias: @rwinch Thanks, I've updated it to use subList.

rwinch:
Thanks @eleftherias! This is now merged into master. I also backported to 5.1.x via gh-7262

PR #7271, opened 2019-08-16
Add setter for authorities claim name in JwtGrantedAuthoritiesConverter

andifalk:
Fixes gh-7100

Prior to this change authorities are always mapped using well known
claim names ('scp' or 'scope'). To change this default behaviour the
converter had to be replaced completely with a custom one.
This commit adds an additional setter to configure a custom
claim name like for example 'roles' to map authorities.
Without specifying a custom claim name
the default claims to be used still remains to
the well known ones ('scp' and 'scope').

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverter.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverter.java
@@ -43,6 +43,8 @@ public final class JwtGrantedAuthoritiesConverter implements Converter<Jwt, Coll
 
 	private String authorityPrefix = DEFAULT_AUTHORITY_PREFIX;
 
+	private String authoritiesClaimName;
+
 	/**
 	 * Extract {@link GrantedAuthority}s from the given {@link Jwt}.
 	 *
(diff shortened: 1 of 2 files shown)


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverter.java
```diff
@@ -70,8 +72,20 @@ public void setAuthorityPrefix(String authorityPrefix) {
...
+	 * @since 5.2
+	 */
+	public void setAuthoritiesClaimName(String claimName) {
+		Assert.hasText(claimName, "claimName cannot be empty");
+		this.claimNames = Collections.singletonList(claimName);
+	}
+
 	private String getAuthoritiesClaimName(Jwt jwt) {
-		for (String claimName : WELL_KNOWN_AUTHORITIES_CLAIM_NAMES) {
+		for (String claimName : this.claimNames) {
```
jzheaux: If the application sets the authorities claim name, then there is no need to search for it. Did you run into difficulties with something more like:

```java
if (this.authoritiesClaimName != null) {
    return this.authoritiesClaimName;
}

for (String claimName : WELL_KNOWN...
```

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverterTests.java
```diff
@@ -139,6 +139,59 @@ public void convertWhenTokenHasEmptyScopeAndNonEmptyScpThenScopeAttributeIsTrans
 		assertThat(authorities).isEmpty();
 	}
 
+	@Test
+	public void convertWhenTokenHasEmptyScopeAndEmptyScpAttributeThenTranslatesToNoAuthorities() {
```
jzheaux: It looks like some of these tests are meant to improve the overall coverage. Since they aren't directly related to the enhancement, will you please put them in a separate commit, in this same PR?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverter.java
andifalk: No, I  just missed that. There is really no need to search for custom authorities claim. Changed it as suggested.

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverterTests.java
andifalk: Yes, these are meant to improve test coverage. Will move these into separate commit.

jzheaux:
Thanks for another PR, @andifalk! This is now merged into `master`.

Issue #7249, opened 2019-08-10
Authentication Mechanisms Should Default their ServerSecurityContextRepository


jzheaux:
The Javadoc for `ServerHttpSecurity#securityContextRepository` states:

> The strategy used with {@code ReactorContextWebFilter}. It does not impact how the {@code SecurityContext} is saved which is configured on a per {@link AuthenticationWebFilter} basis.

This is, however, outdated since form login does use this value if it's specified by the application:

```java
if (this.securityContextRepository != null) {
	this.formLogin.securityContextRepository(this.securityContextRepository);
}
```

I had a chat with @rwinch, and the desired functionality is the following:

If an application wires a security context repository for a given authentication mechanisms, that takes precedence for that mechanism. For example:

```java
http
    .securityContextRepository(a)
    .formLogin()
        .securityContextRepository(b);
```

```java
http
    .formLogin()
        .securityContextRepository(b);
```

In both cases, `b` would be used for form login.

If not, then, the authentication mechanism should use the global one:

```java
http
    .securityContextRepository(a)
        .formLogin();
```

In this case, `a` would be used for form login.

If none is specified by the application, each authentication mechanisms should use what it already defaults to.

So, to complete this task, there are three steps:

1. Update the logic in form login, which currently accepts the global value regardless
2. Update the logic in http basic and oauth2 login, which currently don't consider the global value at all
3. Update the JavaDoc to indicate that setting the `securityContextRepository` does affect the underlying authentication mechanisms.

eddumelendez:
I will take a look at this one @jzheaux

jzheaux:
Sounds great, @eddumelendez, it's yours.

PR #7275, opened 2019-08-18
Allow to set default securityContextRepository for each authenticatio…

eddumelendez:
…n mechanisms

Fixes gh-7249

### DIFF
--- a/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
@@ -268,7 +268,7 @@ public class ServerHttpSecurity {
 
 	private ReactiveAuthenticationManager authenticationManager;
 
-	private ServerSecurityContextRepository securityContextRepository = new WebSessionServerSecurityContextRepository();
+	private ServerSecurityContextRepository securityContextRepository;
 
 	private ServerAuthenticationEntryPoint authenticationEntryPoint;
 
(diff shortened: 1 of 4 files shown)


eddumelendez:
@jzheaux how should `securityContextRepositoryWebFilter()` resolve the `securityContextRepository`? Since the new approach is:

1. Use `securityContextRepository` in the authentication mechanism
2. Use `securityContextRepository` in the global value
3. Use the default one

My current PR is setting the global `securityContextRepository` in the `build()` method by each authentication mechanism which I assume is wrong but didn't notice until running integration tests with configs with multiple authentication mechanism. Should the global value remain with an instance of `WebSessionServerSecurityContextRepository`?

Note: there have been some changes in other tests since http basic and auth2 login never read the `securityContextRepository` and there were some of them changing that value but it was ignored.

Thanks in advance

jzheaux:
@eddumelendez Good observations.

>  Should the global value remain with an instance of WebSessionServerSecurityContextRepository?

No, good catch, I think that this should be set to `null`, and then also change `securityContextRepositoryWebFilter()` to use `WebSessionServerSecurityContextRepository` by default instead of returning `null`.

config/src/test/java/org/springframework/security/config/annotation/web/reactive/EnableWebFluxSecurityTests.java
```diff
@@ -153,6 +153,7 @@ public void defaultPopulatesReactorContext() {
 		client
 			.get()
 			.uri("/")
+			.headers(headers -> headers.setBasicAuth("user", "password"))
```
jzheaux: Why was this added? It seems like an odd side-effect of the change.

eddumelendez: yes, you are right

eddumelendez:
@jzheaux thanks again for all your feedback. I have submitted a new commit

jzheaux:
Great, @eddumelendez, go ahead and squash your commits in preparation for merging.

eddumelendez:
@jzheaux done :)

Issue #7170, opened 2019-07-31
Mock Jwt should ensure that CSRF is not required


jzheaux:
According to https://github.com/spring-projects/spring-security/issues/7118, when using `jwt()` (Servlet) or `mockJwt()` (WebFlux) test support, a tester must also provide a CSRF token.

This is because the CSRF token is lifted when a bearer token header is present in the request.

But, when using `jwt()` or `mockJwt()`, no bearer token header is provided as it's unnecessary.

In order to align with the expected behavior of resource server endpoints, when `jwt()` or `mockJwt()` is used, the tester should not also have to supply a CSRF token.

PR #7276, opened 2019-08-18
Removes the need for a CSRF token with mock JWT

henriquels25:
Fixes: gh-7170

Changes the JwtRequestPostProcessor to remove the check for a CSRF token in the request.

If there is a more elegant way to change the RequestMatcher of the request CsrfFilter, please let me know.

### DIFF
--- a/test/src/main/java/org/springframework/security/test/web/support/WebTestUtils.java
+++ b/test/src/main/java/org/springframework/security/test/web/support/WebTestUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2014 the original author or authors.
+ * Copyright 2002-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


test/src/main/java/org/springframework/security/test/web/support/WebTestUtils.java
```diff
@@ -114,6 +115,21 @@ public static void setCsrfTokenRepository(HttpServletRequest request,
...
 	}
 
+	/**
+	 * Sets the {@link RequestMatcher} for the specified {@link HttpServletRequest}.
+	 *
+	 * @param request the {@link RequestMatcher} to obtain the
+	 * {@link RequestMatcher}
+	 * @param requestMatcher the {@link RequestMatcher} to set
+	 */
+	public static void setCsrfRequestMatcher(HttpServletRequest request,
```
rwinch: We cannot do this because it dirties the context since additional requests without a jwt token would also ignore CSRF protection. The CSRF request should be disabled for just this request.

It might be better to:

1) Update CsrfFilter.shouldNotFilter to return true if a specific request attribute is set
2) Modify the code to set the attribute when jwt is being invoked

jzheaux:
Closing in favor of https://github.com/spring-projects/spring-security/pull/7368 which addresses the concerns listed in https://github.com/spring-projects/spring-security/pull/7276#discussion_r315332028

PR #7291, opened 2019-08-21
XML Exposes Only Single AuthenticationEventPublisher Bean

### LINKED ISSUE #7282: Every XML sec:authentication-manager creates a new global instance of AuthenticationEventPublisher, cannot autowire
### Summary

When you use multiple `sec:authentication-manager`s, you can't `@Autowire` the `AuthenticationEventPublisher` because every `sec:authentication-manager` will instantiate its own public bean

### Detailed explanation
I work on a plugin-based authentication architecture. All plugins must be switched by Spring profiles and should coexist if possible (all active have no effect).

Currently I have two kinds of plugins:

- SSO plugins have their own filter and a dedicated authentication manager, e.g.
- For username-password-based plugins (e.g. Active Directory, database, custom lookup) I had to invent an ExtensibleAuthenticationManager that basically wraps a `ProviderManager` and scans for `AuthenticationProvider`s that support `UsernamePasswordToken`

```
public class ExtensibleAuthenticationManager implements AuthenticationManager, InitializingBean
{
    private final static Logger log = LogManager.getLogger();

    private AuthenticationManager parentAuthenticationManager;
    private ProviderManager providerManager;
    private Class<? extends Authentication> limitToClass;

    @Autowired
    private ListableBeanFactory beanFactory;
    @Autowired
    private MessageSource messageSource;
    @Autowired(required = false)
    private AuthenticationEventPublisher authenticationEventPublisher;

```

### Actual Behavior

[Reference](https://github.com/spring-projects/spring-security/blob/4.2.x/config/src/main/java/org/springframework/security/config/authentication/AuthenticationManagerBeanDefinitionParser.java)

Since Spring Security creates a globally-visible bean of type `DefaultAuthenticationEventPublisher` every time it creates an authentication manager, you can autowire it

### Expected Behavior

There should be either one globally-visible bean of type `AuthenticationEventPublisher` that can be autowired in other beans if required, *or* the definition of a new `AuthenticationManager` should inject an instance of `DefaultAuthenticationEventPublisher` that is private to the authentication manager.

I would prefer the latter, because if one writes their own security plugin there will be actually no public AuthenticationEventPublisher to Autowire, and developer is forced to find their own way (i.e. declare explicitly) to autowire into their authenticatio components

### Configuration

```
<beans profile="CEDACRISEC" xmlns="http://www.springframework.org/schema/beans" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
       xmlns:sec="http://www.springframework.org/schema/security"
       xsi:schemaLocation="http://www.springframework.org/schema/beans     http://www.springframework.org/schema/beans/spring-beans-4.3.xsd
                           http://www.springframework.org/schema/security  http://www.springframework.org/schema/security/spring-security-4.2.xsd"
>
    <sec:authentication-manager id="cedacriAuthenticationManager">
        <sec:authentication-provider ref="cedacriAuthenticationProvider"/>
    </sec:authentication-manager>

    <bean id="formLoginFilter_cedacri" class="it.phoenix.web.security.cedacri.CedacriAuthenticationFilter">


---------------------------


<beans profile="ICCREASEC"
       xmlns:sec="http://www.springframework.org/schema/security"
       xmlns="http://www.springframework.org/schema/beans"
       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
       xsi:schemaLocation="http://www.springframework.org/schema/beans
                        http://www.springframework.org/schema/beans/spring-beans-4.3.xsd
                        http://www.springframework.org/schema/security
                        http://www.springframework.org/schema/security/spring-security-4.2.xsd">

    <bean id="preAuthFilter_iccrea" class="it.phoenix.web.security.spring.PreAuthenticatedIccreaProcessingFilter">
        <property name="sessionAuthenticationStrategy"
                  ref="org.springframework.security.web.authentication.session.CompositeSessionAuthenticationStrategy#0"/>
        <property name="authenticationManager" ref="iccreaAuthenticationManager"/>
        <property name="authenticationSuccessHandler" ref="authenticationHandler"/>
        <property name="authenticationFailureHandler" ref="authenticationHandler"/>
    </bean>

    <sec:authentication-manager id="iccreaAuthenticationManager">
        <sec:authentication-provider ref="iccreaPreAuthProvider"/>
    </sec:authentication-manager>

---------------------------

<beans profile="PREAUTHSEC" xmlns:sec="http://www.springframework.org/schema/security" xmlns="http://www.springframework.org/schema/beans"
       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
       xsi:schemaLocation="http://www.springframework.org/schema/beans
                        http://www.springframework.org/schema/beans/spring-beans-4.3.xsd
                        http://www.springframework.org/schema/security
                        http://www.springframework.org/schema/security/spring-security-4.2.xsd"
>

    <bean id="preAuthFilter_preauthsec" class="it.phoenix.web.security.spring.PreAuthenticatedProcessingFilter">
        <property name="sessionAuthenticationStrategy"
                  ref="org.springframework.security.web.authentication.session.CompositeSessionAuthenticationStrategy#0"/>
        <property name="authenticationFailureHandler" ref="authenticationHandler" />
        <property name="authenticationSuccessHandler" ref="authenticationHandler" />
        <property name="authenticationManager" ref="preauthAuthenticationManager"/>
        <property name="principalRequestHeader" value="${httpheader.preauth.user:SM_USER}"/>
        <property name="rolesRequestHeader" value="${httpheader.preauth.role:SM_ROLES}"/>
        <property name="rolesDelimiter" value="${httpheader.preauth.rolesDelimiter:,}"/>
    </bean>

    <sec:authentication-manager id="preauthAuthenticationManager">
        <sec:authentication-provider ref="profilesAuthenticationProvider"/>
    </sec:authentication-manager>

</beans>

---------------------------

<sec:http name="httpSecurityContext" auto-config="false" use-expressions="true" access-decision-manager-ref="accessDecisionManager"
              authentication-manager-ref="usernamePasswordAuthenticationManager"
              entry-point-ref="authenticationEntryPoint"
    >
   .............

    <bean id="usernamePasswordAuthenticationManager" class="it.phoenix.web.security.spring.ExtensibleAuthenticationManager" lazy-init="true">
        <property name="limitToClass" value="org.springframework.security.authentication.UsernamePasswordAuthenticationToken" />
    </bean>
</sec:http>


```

### Version

Using version 4.2.11 but should affect all 4.2.x

### Related

#4400 
#4269

djechelon:
The DefaultAuthenticationEventPublisher is now an inner bean of the ProviderManager.
I would advise to avoid publishing multiple copies of generated inner beans to the global scope, rather either publish a singleton once or generate inner beans which are not visible to others

Note: I haven't pushed any test (yet), and also could not run unit tests in local IDE. Probably some bot here will run anti regression tests.

For the tests, I have taken a look at AuthenticationManagerBuilderTests.

Here are some ideas:

- Test that no more AuthenticationEventPublisher beans are in the Spring context
Before implementing the change: executing the Builder resulted in both a bean of type ProviderManager and a bean of type AuthenticationEventPublisher to be visible in Spring scope. After implementation only ProviderManager should be visible.
Enhancement: it is adivsable that AuthenticationManagerBuilder creates one and only one bean for segregation purposes (not tested)

- Test that authentication events are correctly propagated
This is actually an integration test. I must make sure that my change is consistent. DefaultAuthenticationEventPublisher has optional dependency over ApplicationEventPublisher. Need to make sure that on a bootstrapped setup, upon an authentication attempt, the correct event is published to the Context.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/authentication/AuthenticationManagerBeanDefinitionParser.java
+++ b/config/src/main/java/org/springframework/security/config/authentication/AuthenticationManagerBeanDefinitionParser.java
@@ -28,6 +28,7 @@ import org.springframework.beans.factory.config.RuntimeBeanReference;
 import org.springframework.beans.factory.parsing.BeanComponentDefinition;
 import org.springframework.beans.factory.parsing.CompositeComponentDefinition;
 import org.springframework.beans.factory.support.BeanDefinitionBuilder;
+import org.springframework.beans.factory.support.GenericBeanDefinition;
 import org.springframework.beans.factory.support.ManagedList;
 import org.springframework.beans.factory.support.RootBeanDefinition;
 import org.springframework.beans.factory.xml.BeanDefinitionParser;
@@ -56,6 +57,8 @@ public class AuthenticationManagerBeanDefinitionParser implements BeanDefinition
 	private static final String ATT_REF = "ref";
 
 	private static final String ATT_ERASE_CREDENTIALS = "erase-credentials";
+
+	private static final String AUTHENTICATION_EVENT_PUBLISHER_BEAN_NAME = "authenticationEventPublisher";
 
 	@Override
 	public BeanDefinition parse(Element element, ParserContext pc) {
@@ -86,11 +89,16 @@ public class AuthenticationManagerBeanDefinitionParser implements BeanDefinition
 		if ("false".equals(element.getAttribute(ATT_ERASE_CREDENTIALS))) {
 			providerManagerBldr.addPropertyValue("eraseCredentialsAfterAuthentication", false);
 		}
-		// Add the default event publisher
-		BeanDefinition publisher = new RootBeanDefinition(DefaultAuthenticationEventPublisher.class);
-		String pubId = pc.getReaderContext().generateBeanName(publisher);
-		pc.registerBeanComponent(new BeanComponentDefinition(publisher, pubId));
-		providerManagerBldr.addPropertyReference("authenticationEventPublisher", pubId);
+
+		if (!pc.getRegistry().containsBeanDefinition(AUTHENTICATION_EVENT_PUBLISHER_BEAN_NAME)) {
+			// Add the default event publisher to the context
+			GenericBeanDefinition publisher = new GenericBeanDefinition();
+			publisher.setBeanClass(DefaultAuthenticationEventPublisher.class);
+			pc.getRegistry().registerBeanDefinition(AUTHENTICATION_EVENT_PUBLISHER_BEAN_NAME, publisher);
+		}
+
+		providerManagerBldr.addPropertyReference("authenticationEventPublisher",
+				AUTHENTICATION_EVENT_PUBLISHER_BEAN_NAME);
 		pc.registerBeanComponent(new BeanComponentDefinition(providerManagerBldr.getBeanDefinition(), id));
 		if (StringUtils.hasText(alias)) {
 			pc.getRegistry().registerAlias(id, alias);
--- a/config/src/main/java/org/springframework/security/config/authentication/AuthenticationManagerBeanDefinitionParser.java
+++ b/config/src/main/java/org/springframework/security/config/authentication/AuthenticationManagerBeanDefinitionParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -28,7 +28,6 @@ import org.springframework.beans.factory.config.RuntimeBeanReference;
 import org.springframework.beans.factory.parsing.BeanComponentDefinition;
 import org.springframework.beans.factory.parsing.CompositeComponentDefinition;
 import org.springframework.beans.factory.support.BeanDefinitionBuilder;
-import org.springframework.beans.factory.support.GenericBeanDefinition;
 import org.springframework.beans.factory.support.ManagedList;
 import org.springframework.beans.factory.support.RootBeanDefinition;
 import org.springframework.beans.factory.xml.BeanDefinitionParser;
@@ -58,7 +57,7 @@ public class AuthenticationManagerBeanDefinitionParser implements BeanDefinition
 
 	private static final String ATT_ERASE_CREDENTIALS = "erase-credentials";
 
-	private static final String AUTHENTICATION_EVENT_PUBLISHER_BEAN_NAME = "authenticationEventPublisher";
+	private static final String AUTHENTICATION_EVENT_PUBLISHER_BEAN_NAME = "defaultAuthenticationEventPublisher";
 
 	@Override
 	public BeanDefinition parse(Element element, ParserContext pc) {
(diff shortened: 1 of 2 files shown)


rwinch:
I think perhaps it is worth leaving as a root bean and checking for the existence of the bean before creating a new one.

As for the tests, a good place to do this would be in [AuthenticationManagerBeanDefinitionParserTests](https://github.com/spring-projects/spring-security/blob/5.2.0.M4/config/src/test/java/org/springframework/security/config/authentication/AuthenticationManagerBeanDefinitionParserTests.java)

eleftherias:
@djechelon Would you be able to update this PR with the feedback Rob provided in the previous comment?

jzheaux:
If this is to be merged, it would need to be merged to `main` instead, and then it could be backported. `4.2.x` is no longer supported at this time.

djechelon:
I have rebased my work on the `main` branch and followed @rwinch's advice... uhm... _my own way_

I also took some time to create a couple of tests which pass locally.

config/src/main/java/org/springframework/security/config/authentication/AuthenticationManagerBeanDefinitionParser.java
```diff
@@ -57,6 +58,8 @@
 
 	private static final String ATT_ERASE_CREDENTIALS = "erase-credentials";
 
+	private static final String AUTHENTICATION_EVENT_PUBLISHER_BEAN_NAME = "authenticationEventPublisher";
```
eleftherias, line 61: `defaultAuthenticationEventPublisher` is a more descriptive name.

config/src/main/java/org/springframework/security/config/authentication/AuthenticationManagerBeanDefinitionParser.java
```diff
@@ -86,11 +89,16 @@ public BeanDefinition parse(Element element, ParserContext pc) {
...
 		}
-		// Add the default event publisher
-		BeanDefinition publisher = new RootBeanDefinition(DefaultAuthenticationEventPublisher.class);
-		String pubId = pc.getReaderContext().generateBeanName(publisher);
-		pc.registerBeanComponent(new BeanComponentDefinition(publisher, pubId));
-		providerManagerBldr.addPropertyReference("authenticationEventPublisher", pubId);
+
+		if (!pc.getRegistry().containsBeanDefinition(AUTHENTICATION_EVENT_PUBLISHER_BEAN_NAME)) {
+			// Add the default event publisher to the context
+			GenericBeanDefinition publisher = new GenericBeanDefinition();
```
eleftherias, line 95: This should remain a `RootBeanDefinition`. The only code that needs to change is the addition of the `if` statement and the bean name.

config/src/test/java/org/springframework/security/config/authentication/AuthenticationManagerBeanDefinitionParserTests.java
```diff
@@ -54,16 +55,42 @@
...
+			+ "            <user name='john' password='{noop}doe' authorities='ROLE_C,ROLE_D' />"
+			+ "        </user-service>"
+			+ "    </authentication-provider>"
+			+ "</authentication-manager>";
+	// @formatter:on
+
 	@Rule
 	public final SpringTestRule spring = new SpringTestRule();
 
+	@Autowired
```
eleftherias, line 72: This can be moved back to where it was before, since it's not used in the new tests.

eleftherias (changes requested):
Thanks for the updates @djechelon! 
I have left some comments inline.
Please also update the copyright header dates at the top of both files.

config/src/test/java/org/springframework/security/config/authentication/AuthenticationManagerBeanDefinitionParserTests.java
djechelon: Apologies, it's probably because of formatting-spree

config/src/main/java/org/springframework/security/config/authentication/AuthenticationManagerBeanDefinitionParser.java
djechelon: I think I got it the correct way now

djechelon:
Please not it's the first time I use review tool on Github, so I don't know if I should be the one doing `Resolve conversation`

eleftherias:
Thanks for the PR @djechelon! This is now merged into main via fe99c3b.
Please excuse the commit message, that was an error on my part.

PR #7301, opened 2019-08-22
Remove redundant throws clauses

larsgrefer:
Removes exceptions that are declared in a method's signature but never thrown by the method itself or its implementations/derivatives.


rwinch:
Thanks for the PR @larsgrefer! This is now merged into master

PR #7307, opened 2019-08-23
Add additional tests for OAuth2 client configuration

eleftherias:
Added a test to cover the case when an `OAuth2AccessTokenResponseClient` bean does not exist.


eleftherias:
Test is not necessary.

PR #7329, opened 2019-08-29
Authorization Response should also match on query parameters

ttddyy:
This is about OAuth2 Authorization Code Grant flow.

When a service is registered with `redirect_uri` that has query parameter (e.g: `http://some/auth?param=foo`), current `OAuth2AuthorizationCodeGrantFilter` doesn't work well.

This is because `OAuth2AuthorizationCodeGrantFilter#shouldProcessAuthorizationResponse` compares registered redirect_uri (AuthRequest: `http://some/auth?param=foo`) against the redirected request URI without query parameters. (AuthResponse: `http://some/auth`)

This patch updates the `redirect_uri` comparison not only checking base uri but also checks query params in the registered `redirect_uri` whether they are all included in the authorization response's query parameters.

Also, in the patch, it now retrieves final redirect_uri from `authorizationRequest` since `authorizationResponse.getRedirectUri()` does not have query parameters.
This might be debatable and also may change based on #7324.

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationCodeGrantFilter.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationCodeGrantFilter.java
@@ -41,6 +41,7 @@ import org.springframework.util.Assert;
 import org.springframework.util.MultiValueMap;
 import org.springframework.util.StringUtils;
 import org.springframework.web.filter.OncePerRequestFilter;
+import org.springframework.web.util.UriComponents;
 import org.springframework.web.util.UriComponentsBuilder;
 
 import javax.servlet.FilterChain;
@@ -48,6 +49,10 @@ import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
 
 /**
  * A {@code Filter} for the OAuth 2.0 Authorization Code Grant,
@@ -77,6 +82,7 @@ import java.io.IOException;
  * </ul>
  *
  * @author Joe Grandja
+ * @author Tadaya Tsuyukubo
  * @since 5.1
  * @see OAuth2AuthorizationCodeAuthenticationToken
  * @see OAuth2AuthorizationCodeAuthenticationProvider
@@ -145,14 +151,104 @@ public class OAuth2AuthorizationCodeGrantFilter extends OncePerRequestFilter {
 		if (authorizationRequest == null) {
 			return false;
 		}
-		String requestUrl = UrlUtils.buildFullRequestUrl(request.getScheme(), request.getServerName(),
-				request.getServerPort(), request.getRequestURI(), null);
 		MultiValueMap<String, String> params = OAuth2AuthorizationResponseUtils.toMultiMap(request.getParameterMap());
-		if (requestUrl.equals(authorizationRequest.getRedirectUri()) &&
-				OAuth2AuthorizationResponseUtils.isAuthorizationResponse(params)) {
+
+		if (OAuth2AuthorizationResponseUtils.isAuthorizationResponse(params) &&
+				isValidRedirectUrl(request, authorizationRequest.getRedirectUri())) {
 			return true;
 		}
 		return false;
+	}
+
+	private boolean isValidRedirectUrl(HttpServletRequest httpServletRequest, String redirectUrl) {
+		UriComponents request = UriComponentsBuilder.fromHttpUrl(UrlUtils.buildFullRequestUrl(httpServletRequest))
+				.replaceQueryParam(OAuth2ParameterNames.CODE)
+				.replaceQueryParam(OAuth2ParameterNames.STATE)
+				.build();
+
+		UriComponents redirect = UriComponentsBuilder.fromUriString(redirectUrl).build();
+
+		// Simple check first
+		if (request.toUriString().equals(redirectUrl)) {
+			return true;
+		}
+
+		// Compare each part of url one by one
+
+		if (!Objects.equals(request.getScheme(), redirect.getScheme())) {
+			return false;
+		}
+
+		if (!Objects.equals(request.getUserInfo(), redirect.getUserInfo())) {
+			return false;
+		}
+
+		if (!Objects.equals(request.getHost(), redirect.getHost())) {
+			return false;
+		}
+
+		if (!Objects.equals(request.getPort(), redirect.getPort())) {
+			return false;
+		}
+
+		if (!Objects.equals(request.getPath(), redirect.getPath())) {
+			return false;
+		}
+
+		if (redirect.getQuery() == null) {
+			return true;  // no need to check request query params
+		}
+
+		// Compare request param/values are in exact order specified in redirect url.
+		// Request(Authorization Response) can have additional parameters appended which is allowed by spec.
+		// Since urls are hierarchical, UriComponents uses HierarchicalUriComponents which keeps the exact
+		// order of parameters.
+		MultiValueMap<String, String> requestParamMap = request.getQueryParams();
+		MultiValueMap<String, String> redirectParamMap = redirect.getQueryParams();
+
+		return containsRedirectUriParamsInOrder(requestParamMap, redirectParamMap);
+
+	}
+
+	private boolean containsRedirectUriParamsInOrder(MultiValueMap<String, String> requestMap, MultiValueMap<String, String> redirectMap) {
+		Iterator<Map.Entry<String, List<String>>> requestIterator = requestMap.entrySet().iterator();
+		Iterator<Map.Entry<String, List<String>>> redirectIterator = redirectMap.entrySet().iterator();
+
+		while (requestIterator.hasNext() && redirectIterator.hasNext()) {
+			Map.Entry<String, List<String>> requestEntry = requestIterator.next();
+			Map.Entry<String, List<String>> redirectEntry = redirectIterator.next();
+
+			String requestParam = requestEntry.getKey();
+			String redirectParam = redirectEntry.getKey();
+
+			if (!requestParam.equals(redirectParam)) {
+				return false;  // param doesn't match
+			}
+
+			List<String> requestValues = requestEntry.getValue();
+			List<String> redirectValues = redirectEntry.getValue();
+
+			if (requestValues.size() < redirectValues.size()) {
+				return false;  // request param values don't have ones specified in redirect param
+			}
+			// request may have additional param values; thus, iterate over redirect values
+			for (int i = 0; i < redirectValues.size(); i++) {
+				String requestValue = requestValues.get(i);
+				String redirectValue = redirectValues.get(i);
+
+				if (!requestValue.equals(redirectValue)) {
+					return false;  // request param value doesn't match
+				}
+			}
+		}
+
+		// request may have additional params which is ok by spec
+
+		if (redirectIterator.hasNext()) {
+			return false;  // request has less params than redirect params
+		}
+
+		return true;
 	}
 
 	private void processAuthorizationResponse(HttpServletRequest request, HttpServletResponse response)
@@ -166,7 +262,8 @@ public class OAuth2AuthorizationCodeGrantFilter extends OncePerRequestFilter {
 
 		MultiValueMap<String, String> params = OAuth2AuthorizationResponseUtils.toMultiMap(request.getParameterMap());
 		String redirectUri = UriComponentsBuilder.fromHttpUrl(UrlUtils.buildFullRequestUrl(request))
-				.replaceQuery(null)
+				.replaceQueryParam(OAuth2ParameterNames.CODE)
+				.replaceQueryParam(OAuth2ParameterNames.STATE)
 				.build()
 				.toUriString();
 		OAuth2AuthorizationResponse authorizationResponse = OAuth2AuthorizationResponseUtils.convert(params, redirectUri);
--- a/config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2AuthorizationCodeGrantRedirectTests.java
+++ b/config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2AuthorizationCodeGrantRedirectTests.java
@@ -1,0 +1,156 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.config.annotation.web.configurers.oauth2.client;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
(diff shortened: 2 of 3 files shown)


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationExchangeUtils.java
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * Utility methods for OAuth 2.0 Authorization Exchange(Request and Response).
+ *
+ * @author Tadaya Tsuyukubo
+ * @since 5.2
+ * @see OAuth2AuthorizationRequest
+ * @see OAuth2AuthorizationResponse
+ */
+public final class OAuth2AuthorizationExchangeUtils {
```
jgrandja: Can you please remove this and keep the changes contained in `OAuth2AuthorizationResponseUtils` and `OAuth2AuthorizationCodeGrantFilter`, as well as the Reactive counterparts. I would like to get this into a patch release.

NOTE: In `processAuthorizationResponse()`, I believe you can build the `redirect-uri` for `OAuth2AuthorizationResponse` as follows:

```java
		MultiValueMap<String, String> params = OAuth2AuthorizationResponseUtils.toMultiMap(request.getParameterMap());
		String redirectUri = UriComponentsBuilder.fromHttpUrl(UrlUtils.buildFullRequestUrl(request))
				.replaceQueryParam(OAuth2ParameterNames.CODE, (Object) null)
				.replaceQueryParam(OAuth2ParameterNames.STATE, (Object) null)
				.build()
				.toUriString();
		OAuth2AuthorizationResponse authorizationResponse = OAuth2AuthorizationResponseUtils.convert(params, redirectUri);
```

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationResponse.java
```diff
@@ -122,6 +133,7 @@ public static Builder error(String errorCode) {
 		private String errorCode;
 		private String errorDescription;
 		private String errorUri;
+		private MultiValueMap<String, String> parameters = new LinkedMultiValueMap<>();
```
jgrandja: This is not needed. Please see previous comment on how to build the `OAuth2AuthorizationResponse.redirectUri` with query parameters.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationExchangeUtils.java
ttddyy: The `OAuth2AuthorizationResponseUtils` is package scoped to `org.springframework.security.oauth2.client.web`, so cannot reference from provider/validator(`org.springframework.security.oauth2.client.authentication`).

So, ok to make it public??

ttddyy:
I have updated the PR
- just kept the `OAuth2AuthorizationExchangeUtils` as is due to utility class scope
- removed `parameters` from `OAuth2AuthorizationResponse`
- updated the way to get `redirectUri` from http request in `OAuth2AuthorizationCodeGrantFilter`
- put back to use `authorizationResponse` to get `redirectUrl`

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationExchangeUtils.java
jgrandja: `OAuth2AuthorizationExchangeValidator` does not need to change. What I was suggesting previously is to add the following to `OAuth2AuthorizationResponseUtils`:

```java
	static OAuth2AuthorizationResponse convert(HttpServletRequest request) {
		MultiValueMap<String, String> params = toMultiMap(request.getParameterMap());
		MultiValueMap<String, String> redirectParams = new LinkedMultiValueMap<>(params);
		redirectParams.remove(OAuth2ParameterNames.CODE);
		redirectParams.remove(OAuth2ParameterNames.STATE);
		String redirectUri = UriComponentsBuilder.fromHttpUrl(UrlUtils.buildFullRequestUrl(request))
				.replaceQueryParams(redirectParams)
				.build()
				.toUriString();
		return convert(params, redirectUri);
	}
```

This utility method can than be used in `shouldProcessAuthorizationResponse` and `processAuthorizationResponse`. This will contain the changes only in `OAuth2AuthorizationCodeGrantFilter` and `OAuth2AuthorizationCodeGrantWebFilter`

ttddyy: ok, but I think there is be a problem when multiple parameters are presented.

For example, following should match:
- `http://localhost/auth?foo=FOO&bar=BAR`
- `http://localhost/auth?bar=BAR&foo=FOO`

If it does simple string match of `redirect_uri` in `OAuth2AuthorizationExchangeValidator`, they are considered as different.

jgrandja: I'm not sure if this is an issue. I'm expecting the provider to perform the redirect using the registered `redirect-uri` **as-is** so it should be the same order as specified in `ClientRegistration.redirect-uri`.

jgrandja: I just tested with Okta and it's performing an exact match.

The registered `redirect-uri` is `http://locahost:8080/authorized/okta2?param1=value1&param2=value2`.

If I trigger the Authorization Request using `http://locahost:8080/authorized/okta2?param2=value2&param1=value1` it fails.

ttddyy:
Updated the PR.

Now, simply it checks the redirected url by removing `code` and `state` parameters.

Limitations would be:
- the redirected url parameters need to be exactly same order in `OAuth2AuthorizationRequest#redirectUri`
- Also, after successful authentication, the authorization server cannot add any additional parameters to the redirecting url (in order for exact match)

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationCodeGrantFilter.java
```diff
@@ -166,7 +171,8 @@ private void processAuthorizationResponse(HttpServletRequest request, HttpServle
 
 		MultiValueMap<String, String> params = OAuth2AuthorizationResponseUtils.toMultiMap(request.getParameterMap());
 		String redirectUri = UriComponentsBuilder.fromHttpUrl(UrlUtils.buildFullRequestUrl(request))
-				.replaceQuery(null)
+				.replaceQueryParam(OAuth2ParameterNames.CODE)
```
jgrandja: @ttddyy After reviewing the spec:

> The redirection endpoint URI MUST be an absolute URI as defined by
   [RFC3986] Section 4.3.  The endpoint URI MAY include an
   "application/x-www-form-urlencoded" formatted (per Appendix B) query
   component ([RFC3986] Section 3.4), which MUST be retained when **adding
   additional query parameters**.  The endpoint URI MUST NOT include a
   fragment component.

NOTE the bold highlight. As per spec, the Authorization Server will ADD additional query parameters to the `redirect-uri`. This means that the query parameters in the `redirect_uri` (if any) should be in the correct order. Also, in addition to `code` and `state`, the Authorization Server may add additional query parameters. I've tested this with google and additional parameters are added on top of `code` and `state`. Can you please add a test for this scenario and adjust the code for this.

ttddyy:
@jgrandja 
Updated the logic to handle additional params from auth server.
For test, added a case to simulate additional parameter added by authorization server.
Also, rebased to the latest master.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationCodeGrantFilter.java
```diff
@@ -166,7 +197,8 @@ private void processAuthorizationResponse(HttpServletRequest request, HttpServle
...
+				.replaceQueryParam(OAuth2ParameterNames.STATE)
 				.build()
 				.toUriString();
 		OAuth2AuthorizationResponse authorizationResponse = OAuth2AuthorizationResponseUtils.convert(params, redirectUri);
```
jgrandja: If the authorization server adds additional parameters to the authorization response, than the `OAuth2AuthorizationResponse.redirectUri` will contain the additional parameters and result in a failed validation check in `OAuth2AuthorizationCodeAuthenticationProvider` and the call to `OAuth2AuthorizationExchangeValidator.validate()` when it compares the `redirectUri` between the authorization request and response. This revealed that we need an integration test for this scenario. Can you please add an integration test and fix so it passes. Take a look at `OAuth2LoginTests` or `OAuth2LoginConfigurerTests` for an example for the integration test.

ttddyy: ah, true.
So, now the question is how you want to share the validation logic.
My original PR introduced `OAuth2AuthorizationExchangeUtils` class to share the validation logic between `OAuth2AuthorizationCodeGrantFilter` and `OAuth2AuthorizationExchangeValidator`. (also `OidcAuthorizationCodeAuthenticationProvider`). ([link](https://github.com/ttddyy/spring-security/commit/52dcd13e31439491de9c93086c9e7a9b74886d07#diff-da51724beb161d9b530f7494b5b78daeR36) bit outdated)

This is because `OAuth2AuthorizationResponseUtils` seems good place to put the validation logic but it is package scoped class.
Since `~CodeGrantFilter` and `~ExchangeValidator` are in different package, we need a way to share this validation logic. So, if you don't oppose, I'm going to put back `OAuth2AuthorizationExchangeUtils` to share the redirect url validation logic.

jgrandja: > So, now the question is how you want to share the validation logic.
My original PR introduced `OAuth2AuthorizationExchangeUtils` class to share the validation logic between `OAuth2AuthorizationCodeGrantFilter` and `OAuth2AuthorizationExchangeValidator`

We do not want to share the logic between `OAuth2AuthorizationCodeGrantFilter` and `OAuth2AuthorizationExchangeValidator`. Reason being is that `OAuth2AuthorizationCodeGrantFilter` is in the `web` sub-package and the `OAuth2AuthorizationExchangeValidator` is in the `authentication` sub-package, which is NOT a web component.

The validation check in `OAuth2AuthorizationExchangeValidator` can simply be changed to:

```java
if (!authorizationResponse.getRedirectUri().startsWith(authorizationRequest.getRedirectUri()))
```

It's safe to use `startsWith`, since we know that any additional query parameters will be **appended** to the registered `redirect-uri`

> The redirection endpoint URI MUST be an absolute URI as defined by
[RFC3986] Section 4.3. The endpoint URI MAY include an
"application/x-www-form-urlencoded" formatted (per Appendix B) query
component ([RFC3986] Section 3.4), which MUST be retained when **adding
additional query parameters**. The endpoint URI MUST NOT include a
fragment component.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationCodeGrantFilter.java
```diff
@@ -145,16 +147,45 @@ private boolean shouldProcessAuthorizationResponse(HttpServletRequest request) {
 		if (authorizationRequest == null) {
 			return false;
 		}
-		String requestUrl = UrlUtils.buildFullRequestUrl(request.getScheme(), request.getServerName(),
-				request.getServerPort(), request.getRequestURI(), null);
 		MultiValueMap<String, String> params = OAuth2AuthorizationResponseUtils.toMultiMap(request.getParameterMap());
-		if (requestUrl.equals(authorizationRequest.getRedirectUri()) &&
+
+		if (isValidRedirectUrl(request, authorizationRequest.getRedirectUri()) &&
```
jgrandja: @ttddyy This can be simplified as follows:

```java
		String requestUrl = UrlUtils.buildFullRequestUrl(request);
		MultiValueMap<String, String> params = OAuth2AuthorizationResponseUtils.toMultiMap(request.getParameterMap());
		if (requestUrl.startsWith(authorizationRequest.getRedirectUri()) &&
				OAuth2AuthorizationResponseUtils.isAuthorizationResponse(params)) {
			return true;
		}
```

ttddyy: @jgrandja 
There is a problem with `startWith()` check which is caught by existing test.

Let's say here is changed to use `startWith()` as below:

```java
String requestUrl = UriComponentsBuilder.fromHttpUrl(UrlUtils.buildFullRequestUrl(request))
    .replaceQueryParam(OAuth2ParameterNames.CODE)
    .replaceQueryParam(OAuth2ParameterNames.STATE)
    .build()
    .toUriString();
MultiValueMap<String, String> params = OAuth2AuthorizationResponseUtils.toMultiMap(request.getParameterMap());

if (requestUrl.startsWith(authorizationRequest.getRedirectUri()) &&
    OAuth2AuthorizationResponseUtils.isAuthorizationResponse(params)) {
    return true;
}
return false;
```

This logic matches when:
- requestUrl = `http://localhost/callback/client-1-no-match`
- authorizationRequest.getRedirectUri() = `http://localhost/callback/client-1`

Thus, simple `startWith()` check doesn't work.

So, if we don't want to share the code between filter and validator:
- put same logic in both place
- or, maybe perform through check(`isValidRedirectUrl`) in filter, but simply only do `startWith()` check in validator

ttddyy:
Updated the PR adding an integration test, `OAuth2AuthorizationCodeGrantRedirectTests`.
Currently some of the tests are failing as expected since the validation logic is not yet updated.
Once we solve the validation logic issue, I'll update the PR.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationCodeGrantFilter.java
jgrandja: @ttddyy Yes, you are right, there is a problem with using `startsWith()`. Instead we should perform exact matching on all parts of the URI, with the exception of the query parameters since there may be additional parameters sent in the authorization response.

In the `shouldProcessAuthorizationResponse()` method, can you make use of `UriComponentsBuilder` for both the `request` (authorization response) and `authorizationRequest.getRedirectUri()` and perform exact matching on each part, eg. scheme, userInfo, host, port, path. The query parameters (if any) in the `authorizationRequest.getRedirectUri()` should exist in the same order in `request` (authorization response).

On a side note, the validation logic for `redirect-uri` in `OAuth2AuthorizationExchangeValidator` is redundant. The matching happens in `shouldProcessAuthorizationResponse()` so we don't need to double validate in `OAuth2AuthorizationExchangeValidator`. Please remove the existing validation which will simply things.

jgrandja:
@ttddyy 

> the validation logic for redirect-uri in `OAuth2AuthorizationExchangeValidator` is redundant. The matching happens in `shouldProcessAuthorizationResponse()` so we don't need to double validate in `OAuth2AuthorizationExchangeValidator`.

I removed the redundant validation in #7708 to help with this PR.

ttddyy:
> I removed the redundant validation in #7708 to help with this PR.
Thanks, will work on this ticket early next week with rebasing to the latest.

ttddyy:
@jgrandja 
Hi, the PR has updated. Can you check.
Thanks,

jgrandja:
@ttddyy Thank you for the updates. After reviewing, I still feel the logic can be simplified and the code can be reduced further. Our ultimate goal with PR's is to introduce the minimal amount of code required to either fix a bug or introduce a new feature. The main reason is because the more code we add the more we need to maintain. It is definitely a challenging task but it is important.

Having said that, I decided to go ahead and apply the fix myself in this commit 3c86239. Thank you for your understanding.

Issue #7284, opened 2019-08-20
Resource Server should support WebClient Bearer Token propagation


jzheaux:
A resource server can propagate the bearer token by fashioning an `ExchangeFilterFunction`:

```java
ExchangeFilterFunction oauth2 = (request, next) -> {
    AbstractOAuth2TokenAuthenticationToken<?> authentication = 
        (AbstractOAuth2TokenAuthenticationToken<?>) 
            SecurityContextHolder.getContext().getAuthentication();
    request.headers().setBearerAuth(authentication.getToken().getTokenValue());
    return next.exchange(request);
};
```

It would be nice to remove some of this boilerplate.

jzheaux:
Related to https://github.com/spring-cloud/spring-cloud-security/issues/217

Issue #5334, opened 2018-05-11
Add BearerExchangeFilterFunction


rwinch:
### Summary
Create an `ExchangeFilterFunction` that adds Bearer tokens

jgrandja:
This may be resolved as a result of sub-tasks required in #6811

PR #7330, opened 2019-08-29
Add Servlet and ServerBearerExchangeFilterFunction

jzheaux:
Fixes gh-5334
Fixes gh-7284

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/ServletBearerExchangeFilterFunction.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/ServletBearerExchangeFilterFunction.java
@@ -1,0 +1,248 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.oauth2.server.resource.web;
+
+import java.util.Map;
+import java.util.function.Consumer;
+
+import org.reactivestreams.Subscription;
+import reactor.core.CoreSubscriber;
+import reactor.core.publisher.Hooks;
+import reactor.core.publisher.Mono;
+import reactor.core.publisher.Operators;
+import reactor.util.context.Context;
+
+import org.springframework.beans.factory.DisposableBean;
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.lang.Nullable;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.oauth2.core.AbstractOAuth2Token;
+import org.springframework.web.reactive.function.client.ClientRequest;
+import org.springframework.web.reactive.function.client.ClientResponse;
+import org.springframework.web.reactive.function.client.ExchangeFilterFunction;
+import org.springframework.web.reactive.function.client.ExchangeFunction;
+import org.springframework.web.reactive.function.client.WebClient;
+
+/**
+ * An {@link ExchangeFilterFunction} that adds the
+ * <a href="https://tools.ietf.org/html/rfc6750#section-1.2" target="_blank">Bearer Token</a>
+ * from an existing {@link AbstractOAuth2Token} tied to the current {@link Authentication}.
+ *
+ * Suitable for Servlet applications, applying it to a typical {@link org.springframework.web.reactive.function.client.WebClient}
+ * configuration:
+ *
+ * <pre>
+ *  @Bean
+ *  WebClient webClient() {
+ *      ServletBearerExchangeFilterFunction bearer = new ServletBearerExchangeFilterFunction();
+ *      return WebClient.builder()
+ *              .apply(bearer.oauth2Configuration())
+ *              .build();
+ *  }
+ * </pre>
+ *
+ * @author Josh Cummings
+ * @since 5.2
+ */
+public class ServletBearerExchangeFilterFunction
+		implements ExchangeFilterFunction, InitializingBean, DisposableBean {
+
+	private static final String AUTHENTICATION_ATTR_NAME = Authentication.class.getName();
+
+	private static final String REQUEST_CONTEXT_OPERATOR_KEY = RequestContextSubscriber.class.getName();
+
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public void afterPropertiesSet() throws Exception {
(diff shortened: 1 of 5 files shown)


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/ServletBearerExchangeFilterFunction.java
```diff
@@ -0,0 +1,247 @@
...
+		private CoreSubscriber<T> delegate;
+		private final Context context;
+
+		private RequestContextSubscriber(CoreSubscriber<T> delegate,
+				Authentication authentication) {
+
+			this.delegate = delegate;
+			Context parentContext = this.delegate.currentContext();
+			Context context;
+			if (parentContext.hasKey(REQUEST_CONTEXT_DATA_HOLDER_ATTR_NAME)) {
```
robotmrv: maybe it would be better not to create empty RequestContextDataHolder?
by 
```java
if (authentication == null || parentContext.hasKey(REQUEST_CONTEXT_DATA_HOLDER_ATTR_NAME)) {}
```

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/ServletBearerExchangeFilterFunction.java
```diff
@@ -0,0 +1,247 @@
...
+	private static final String AUTHENTICATION_ATTR_NAME = Authentication.class.getName();
+
+	private static final String REQUEST_CONTEXT_OPERATOR_KEY = RequestContextSubscriber.class.getName();
+
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public void afterPropertiesSet() throws Exception {
+		Hooks.onLastOperator(REQUEST_CONTEXT_OPERATOR_KEY, Operators.lift((s, sub) -> createRequestContextSubscriber(sub)));
```
robotmrv: `Operators.liftPublisher()` will not create unnecessary transformations into `Scannable` of a publisher

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/server/ServerBearerExchangeFilterFunction.java
```diff
@@ -0,0 +1,117 @@
...
+
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public Mono<ClientResponse> filter(ClientRequest request, ExchangeFunction next) {
+		return oauth2Token(request.attributes())
+				.map(oauth2Token -> bearer(request, oauth2Token))
+				.flatMap(next::exchange)
+				.switchIfEmpty(Mono.defer(() -> next.exchange(request)));
```
robotmrv: maybe would be better to write
```java
return oauth2Token(request.attributes())
    .map(oauth2Token -> bearer(request, oauth2Token))
    .defaultIfEmpty(request)
    .flatMap(next::exchange);
```
This should create less Publishers/functions and it does not duplicate `next::exchange`

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/ServletBearerExchangeFilterFunction.java
```diff
@@ -0,0 +1,247 @@
...
+
+	/**
+	 * Provides defaults for the {@link Authentication} using
+	 * {@link SecurityContextHolder}. It also can default the {@link AbstractOAuth2Token} using the
+	 * {@link #authentication(Authentication)}.
+	 * @return the {@link Consumer} to populate the attributes
+	 */
+	public Consumer<WebClient.RequestHeadersSpec<?>> defaultRequest() {
+		return spec -> spec.attributes(attrs -> {
+			populateDefaultAuthentication(attrs);
```
robotmrv: I wonder what case does it cover?
Isn't it covered by `Hooks.onLastOperator()` from afterPropertiesSet() - it does the same. 
It seems to me that this is just legacy solution copied from 
https://github.com/spring-projects/spring-security/blob/master/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
Or is it just an altarnative for case when Hook is not enabled and chaining is not necessary?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/ServletBearerExchangeFilterFunction.java
jzheaux: Thanks for the tip! I'll update that.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/ServletBearerExchangeFilterFunction.java
jzheaux: This is an alternative for when more sophisticated chaining is wanted by the application, so `oauth2Configuration` would not suffice. Order obviously matters with filters, so, for example, I might need to default the request, add my own custom filter, and then add this filter after that.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/ServletBearerExchangeFilterFunction.java
```diff
@@ -0,0 +1,248 @@
...
+	 * {@inheritDoc}
+	 */
+	@Override
+	public Mono<ClientResponse> filter(ClientRequest request, ExchangeFunction next) {
+		return mergeRequestAttributesIfNecessary(request)
+				.filter(req -> req.attribute(AUTHENTICATION_ATTR_NAME).isPresent())
+				.map(req -> getOAuth2Token(req.attributes()))
+				.map(token -> bearer(request, token))
+				.flatMap(next::exchange)
+				.switchIfEmpty(Mono.defer(() -> next.exchange(request)));
```
robotmrv: I know there are activities to reduce number of `Optional` usage and replacement of `filter/map/flatMap()` to the `handle()` in the spring core  
so it could be rewritten to  
may be it is a little bit more imperative, but it creates less operators and does not use create `Optional` to check attribute presence 
```java
public Mono<ClientResponse> filter(ClientRequest request, ExchangeFunction next) {
	return mergeRequestAttributesIfNecessary(request)
			.handle((req, sink) -> {
				Map<String, Object> attributes = req.attributes();
				if (attributes.get(AUTHENTICATION_ATTR_NAME) == null) {
					sink.next(request);
				} else {
					AbstractOAuth2Token token = getOAuth2Token(attributes);
					sink.next(bearer(req, token));
				}
			})
			.flatMap(next::exchange);
}

private Mono<ClientRequest> mergeRequestAttributesIfNecessary(ClientRequest request) {
	if (request.attributes().get(AUTHENTICATION_ATTR_NAME) != null) {
		return Mono.just(request);
	}
	return mergeRequestAttributesFromContext(request);
}

private Mono<ClientRequest> mergeRequestAttributesFromContext(ClientRequest request) {
	return Mono.subscriberContext()
			.map(ctx -> ClientRequest.from(request)
					.attributes(attrs -> populateRequestAttributes(attrs, ctx))
					.build()
			);
}
```

PR #7333, opened 2019-08-30
Propagate additional parameters from exchange response

ttddyy:
This is about OAuth2 Authorization Code Grant flow.

Currently, when constructing `OAuth2AuthorizationCodeAuthenticationToken` from exchange response, `additionalParameters` are not passed to the token in `OAuth2AuthorizationCodeAuthenticationProvider`. Consequently, when saving the authorized client with `OAuth2AuthorizedClientRepository#saveAuthorizedClient`, `additionalParameters` are not available.

This prevents persisting additional parameters when storing access-token and refresh-token.

This PR propagates `additionalParameters` to `OAuth2AuthorizationCodeAuthenticationToken` and  `OAuth2AuthorizedClient`.
So that, they can be used in later part - `OAuth2AuthorizedClientRepository#saveAuthorizedClient`.

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/ClientCredentialsOAuth2AuthorizedClientProvider.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/ClientCredentialsOAuth2AuthorizedClientProvider.java
@@ -82,7 +82,8 @@ public final class ClientCredentialsOAuth2AuthorizedClientProvider implements OA
 		OAuth2AccessTokenResponse tokenResponse =
 				this.accessTokenResponseClient.getTokenResponse(clientCredentialsGrantRequest);
 
-		return new OAuth2AuthorizedClient(clientRegistration, context.getPrincipal().getName(), tokenResponse.getAccessToken());
+		return new OAuth2AuthorizedClient(clientRegistration, context.getPrincipal().getName(), tokenResponse.getAccessToken(),
+				null, tokenResponse.getAdditionalParameters());
 	}
 
 	private boolean hasTokenExpired(AbstractOAuth2Token token) {
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/ClientCredentialsReactiveOAuth2AuthorizedClientProvider.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/ClientCredentialsReactiveOAuth2AuthorizedClientProvider.java
@@ -78,7 +78,8 @@ public final class ClientCredentialsReactiveOAuth2AuthorizedClientProvider imple
 		return Mono.just(new OAuth2ClientCredentialsGrantRequest(clientRegistration))
 				.flatMap(this.accessTokenResponseClient::getTokenResponse)
 				.map(tokenResponse -> new OAuth2AuthorizedClient(
-						clientRegistration, context.getPrincipal().getName(), tokenResponse.getAccessToken()));
+						clientRegistration, context.getPrincipal().getName(), tokenResponse.getAccessToken(),
+						null, tokenResponse.getAdditionalParameters()));
 	}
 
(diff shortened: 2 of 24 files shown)


jgrandja:
@ttddyy Can you please provide me a use case on why you need `OAuth2AuthorizationCodeAuthenticationToken.additionalParameters` to be propagated to `OAuth2AuthorizedClient`? As of now, `OAuth2AuthorizationCodeAuthenticationToken.additionalParameters` is populated with the `OAuth2AccessTokenResponse.getAdditionalParameters()` in `OAuth2AuthorizationCodeReactiveAuthenticationManager`, which represents the additional parameters returned in the Access Token Response.

After the client completes the `authorization_code` grant flow, the Access Token additional parameters are not valid anymore as the client simply uses the `OAuth2AccessToken` going forward. So IMO, propagating the `additionalParameters` to `OAuth2AuthorizedClient` does not make sense.

Can you please provide a concrete use case on why you feel this is needed.

Also, in the future, I would highly encourage you to log an issue to start the discussion first before you prepare a PR. This will ensure you don't waste your time preparing a PR that might not get accepted.

ttddyy:
Hi @jgrandja 

My usecase is handling the id-token.

The `id-token` comes as `OAuth2AccessTokenResponse#additionalParameters`.
I want to store it with other tokens - access token and refresh token.

So, for reactive side, `OAuth2AuthorizationCodeReactiveAuthenticationManager` is passing the additional parameters, but for servlet side, `OAuth2AuthorizationCodeAuthenticationProvider` is missing to pass the additional params to `OAuth2AuthorizationCodeAuthenticationToken`.
This is the first part.

The second part is:
While persisting tokens, instead of passing access-token and refresh-token as separate argument, I'm passing `OAuth2AuthorizedClient` as an argument to the service method since this is the representation of the authorized client and contains all oauth related information. I can pass id-token separately, but it is more natural that id-token is part of authorized client information.
This is the reason for adding `additionalParameters` to `OAuth2AuthorizedClient`.

jgrandja:
@ttddyy 

> The id-token comes as `OAuth2AccessTokenResponse#additionalParameters`

True but this only happens in the OpenID Connect `authorization_code` flow, which is `OAuth2LoginAuthenticationProvider` NOT `OAuth2AuthorizationCodeAuthenticationProvider` (as you have mentioned). NOTE: The `OAuth2AuthorizationCodeAuthenticationProvider` handles the standard OAuth 2.0 `authorization_code` grant.

> it is more natural that id-token is part of authorized client information

The [ID Token](https://openid.net/specs/openid-connect-core-1_0.html#IDToken) is a security token that contains Claims about the Authentication of **an End-User**.

However, the `OAuth2AuthorizedClient` is an OAuth 2.0 **Client** that was authorized by the End-User (Resource Owner).

These are 2 completely independent entities, hence, the separation of `OAuth2AuthorizedClient` and `OidcUser`. However, they are associated via `OAuth2AuthorizedClient.principalName`.

Furthermore, the `OAuth2AccessToken` and `OAuth2RefreshToken` are **longer-lived tokens** compared to the **shorter-lived token** of the `OidcIdToken`. The `id_token` is bound to the session which typically expires in 30 mins (by default) in most containers and therefore does not make sense to persist on a longer term basis as is the case for `OAuth2AuthorizedClient`.

As an FYI, you can always access the `OidcIdToken` via `Authentication.getPrincipal()` when it's an instance of `OAuth2LoginAuthenticationToken` after completing an OIDC flow. The principal will be an `OidcUser` and than you can `OidcUser.getIdToken()`.

I hope this makes sense?

ttddyy:
@jgrandja 
Thanks for the detailed explanation.

I misunderstood client and end-user code grant.
Initially, I configured to use `http.oauth2Client()` rather than `http.oauth2Login()` for oidc flow. After setting `http.oauth2Login()` different mechanism(filters) is configured to handle the path of code exchange.
I will spend more time on understanding and configuring it to our environment.

Thanks for the clarification. I'am closing this issue since it is not the implementation for end-user code grant flow.

Issue #7327, opened 2019-08-29
OAuth2AuthorizationRequest not removed from session


AndreasKl:
When #6215 was fixed only the adding of new `OAuth2AuthorizationRequest`s was fixed, not the removal of those. With a distributed session store we observed an increase in session size for users having long running sessions.

A dump of the keys of the session attributes revealed a huge `HashMap` of `OAuth2AuthorizationRequest`. This is due to `org.springframework.security.oauth2.client.web.server.WebSessionOAuth2ServerAuthorizationRequestRepository#removeAuthorizationRequest` only removing the `OAuth2AuthorizationRequest` from the `HashMap` and not updating the session attributes leaving no clue to the session repository that the session was amended.

The expected behaviour would be that the **stateToAuthzRequest** `HashMap` should not grow without limit and `OAuth2AuthorizationRequest` should be removed after it was used to create a new session.

Used version: spring-security-oauth2-client-5.1.6.RELEASE.jar
however the issue exists on master: https://github.com/spring-projects/spring-security/blob/master/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/WebSessionOAuth2ServerAuthorizationRequestRepository.java#L85

jgrandja:
Thanks for the report @AndreasKl. You are right, `WebSessionOAuth2ServerAuthorizationRequestRepository.removeAuthorizationRequest()` does not update the sessions attributes with the updated `Map` of the removal of `OAuth2AuthorizationRequest`.

Would you be interested in submitting a PR for this fix?

AndreasKl:
@jgrandja Would be happy to provide the PR.

Do you think it is of any value, if I also provide a PR with an optional facility that allows to limit the amount of `OAuth2AuthorizationRequest`s for a single session. This could be an additional session attribute backed by a size limited `LinkedHashMap` containing the `state` values as keys and an empty object as value (this would have the benefit that existing sessions can be de-serialized).

jgrandja:
@AndreasKl 

> Do you think it is of any value, if I also provide a PR with an optional facility that allows to limit the amount of OAuth2AuthorizationRequests for a single session

The existing issue #5145 tracks this enhancement. We would need this enhancement implemented on both the Servlet and Reactive side. It would be great if you can provide a PR?

AndreasKl:
@jgrandja Thanks for the hint, wasn't aware there is already an open issue.

PR #7334, opened 2019-08-30
Remove OAuth2AuthorizationRequest when a distributed session is used

AndreasKl:
Dirties the WebSession by putting the amended AUTHORIZATION_REQUEST map into
the WebSession even it was already in the map. This causes common SessionRepository
implementations like Redis to persist the updated attribute.

Fixes gh-7327

Author: Andreas Kluth <mail@andreaskluth.net>


jgrandja:
@AndreasKl Thanks for the PR @AndreasKl. This is now in master. FYI, I applied a little polish to the  test.

Issue #7273, opened 2019-08-17
Redirect using a relative URL


jzheaux:
Spring Security originally redirected on authentication success with an absolute URL, conforming to https://www.rfc-editor.org/rfc/rfc2616#section-14.30. More broadly, any redirects to within the application should use relative URL

However, this RFC is now obsolete. To that end, the reactive stack redirects using a relative URL. It would be good to bring the servlet code into alignment.

davidcostanzo:
When I posted my earlier comment, I was completely lost.  The proposed fix for this issue put me on a path to understanding the control flow.  I think the absolute URL is created in [LoginUrlAuthenticationEntryPoint. buildRedirectUrlToLoginPage()](https://github.com/spring-projects/spring-security/blob/5.0.x/web/src/main/java/org/springframework/security/web/authentication/LoginUrlAuthenticationEntryPoint.java#L178-L207), which goes out of its way to convert a relative URL to an absolute one, conforming to the obsolete RFC.

I was able to work around this problem in my application by creating a subclass of `LoginUrlAuthenticationEntryPoint` that overrides `buildRedirectUrlToLoginPage` to return the URL that `determineUrlToUseForThisRequest` returns, which is a relative URL in my case.  In my application's `SecurityFilterChain` bean, it additionally configures the `HttpSecurity` with

```
   http.exceptionHandling(exceptionHandling -> {
       exceptionHandling.authenticationEntryPoint(new MyLoginUrlAuthenticationEntryPoint("/login"));
   });
```
This ignores any port mapper configuration, so it won't work for everyone.

jzheaux:
Note that I removed the `breaks-passivity` label by putting this behind a setting, `setFavorRedirectUris`. [This default will get switched](https://github.com/spring-projects/spring-security/issues/16300) in Spring Security 7.

PR #7358, opened 2019-09-04
DefaultRedirectStrategy Host-less Redirect

jzheaux:
Fixes gh-7273


jzheaux:
Because of features like the port mapper, I want to think about this a bit more before proceeding with a PR.

Issue #6762, opened 2019-04-09
Introduce a Map-based AuthenticationManagerResolver


jzheaux:
Recently, #6722 introduced a contract for resolving an `AuthenticationManager` via some context, like a request.

A natural implementation for this would be one that contained a `Map<RequestMatcher, AuthenticationManager>` and selected the `AuthenticationManager` from the configured `Map`.

This aligns with other request-matching API implementations like `DelegatingAuthenticationEntryPoint`  and `RequestMatcherDelegatingAccessDeniedHandler`.

clevertension:
i can't find `AuthenticationMatcher` in the repo, do you mean `AuthenticationManager`?

jzheaux:
Good catch, @clevertension. I've updated the description.

PR #7366, opened 2019-09-05
Add Request AuthenticationManagerResolvers

jzheaux:
Fixes gh-6762

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/RequestMatcherDelegatingAuthenticationManagerResolver.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/RequestMatcherDelegatingAuthenticationManagerResolver.java
@@ -1,0 +1,146 @@
+/*
+ * Copyright 2002-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.web.authentication;
+
+import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


jzheaux:
@bdemers I remember you saying that you were taking a look at choosing JWT vs Opaque Token based on path. Will you take a look at this PR and see whether it would be helpful for what you are trying to do?

bdemers:
@jzheaux This is great! I just had a chance to play around with this!

One thing that _feels_ odd is configuring the path matching in two different places.
here: https://github.com/spring-projects/spring-security/blob/2a3f03ab005b1eb915f7ae49a8d8697c441286dc/samples/boot/oauth2resourceserver-multitenancy/src/main/java/sample/OAuth2ResourceServerSecurityConfiguration.java#L63
and again here: https://github.com/spring-projects/spring-security/blob/2a3f03ab005b1eb915f7ae49a8d8697c441286dc/samples/boot/oauth2resourceserver-multitenancy/src/main/java/sample/OAuth2ResourceServerSecurityConfiguration.java#L76-L77

These types of use-case may not be common enough to worry about that (and the OAuth configs are at the _top_ level of `HttpSecurity`...)

Any idea of how this type of strategy will play with Boot's autoconfigure classes/properties (`security.oauth2.*`).  For example, currently, you cannot define a `jwk.key-set-uri` and a `token-info-uri` for the same provider. (again this might not be a common enough use case currently to worry about just yet)

That said, I'd love to use this as is today :)

bdemers:
@jzheaux I was playing around with this again today, and noticed what the `RequestMatchingAuthenticationManagerResolver.setDefaultAuthenticationManager()` could do.

That seems to remove my concern about the paths being defined twice.  

I could do something like:

```java
LinkedHashMap<RequestMatcher, AuthenticationManager> authenticationManagers = new LinkedHashMap<>();
            authenticationManagers.put(new AntPathRequestMatcher("/opaque/**"), opaque());
            RequestMatchingAuthenticationManagerResolver authenticationManagerResolver = new RequestMatchingAuthenticationManagerResolver(authenticationManagers);
            authenticationManagerResolver.setDefaultAuthenticationManager(jwt());
            return authenticationManagerResolver;
```

Assuming I'm seeing this correctly, this would (more or less) function the same way as the out of the box JWT Decoding (with some configuration differences).  Where all paths are handled by the JWT decoder, except the special handling `/opaque/**`.

Very cool stuff!

bdemers:
Woot!!

Issue #5145, opened 2018-03-20
OAuth2AuthorizationRequest should expire


jgrandja:
We should ensure the `OAuth2AuthorizationRequest` expires after the configured time limit.

jgrandja:
@rwinch I've been thinking about assigning this as a first-timer issue. However, as I looked into this further, I'm not sure if we really need to implement this. The current and only implementation for `AuthorizationRequestRepository` is `HttpSession`-based, including it's reactive counterpart. So the `OAuth2AuthorizationRequest`(s) will be removed whenever a session is terminated anyway. Although we can be more proactive in expiring `OAuth2AuthorizationRequest`(s) before a session is terminated, I'm not sure there is value here. What are your thoughts?

jgrandja:
@rwinch You proposed this change to me way back and I logged the issue as I thought it was a good idea at the time as well.

> Can you remind me why we want to expire it?

If the user triggers the Authorization Code flow OR the `oauth2Login()` flow and does not proceed through to the end of the flow, for example, closes the browser at the consent screen, than the `OAuth2AuthorizationRequest` will essentially be _orphaned_ in the `AuthorizationRequestRepository` until the session is expired. IMO, I don't see an issue with this? Do you?

> Where is the configured limit coming from?

This would be part of the feature if implemented. I'm guessing we would have a setter in the implementation(s) of `AuthorizationRequestRepository`.

rwinch:
Thanks for the reminder.

It's probably not a huge deal, but I'd like to leave this open. It is technically a memory leak and sessions can be kept around quite a long time.

AndreasKl:
@rwinch @jgrandja We recently observed this issues which caused an increase in per call overhead in our gateway, due to near 4k of `OAuth2AuthorizationRequest` attached to a few sessions. 

As I did not want to invalidate existing sessions a companion map was added to the session holding a configurable amount of `OAuth2AuthorizationRequest` `state` keys.

If you like I could provide a PR for this change? Imho I would prefer a limit based solution over a time based one as a possible attacker could just create a lot of `OAuth2AuthorizationRequest` just by coincident.

jgrandja:
@AndreasKl I believe this solution should have both time-limit and size-limit. The size-limit will prevent resource exhaustion and the time-limit will ensure resources are not wasted.

[...]

PR #7381, opened 2019-09-05
Add expireAt to OAuth2AuthorizationRequest and clean-up expired requests on save

AndreasKl:
Fixes gh-5145.

### DIFF
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java
@@ -15,6 +15,8 @@
  */
 package org.springframework.security.oauth2.core.endpoint;
 
+import java.time.Clock;
+import java.time.Instant;
 import org.springframework.security.core.SpringSecurityCoreVersion;
 import org.springframework.security.oauth2.core.AuthorizationGrantType;
 import org.springframework.util.Assert;
@@ -55,6 +57,7 @@ public final class OAuth2AuthorizationRequest implements Serializable {
 	private Map<String, Object> additionalParameters;
 	private String authorizationRequestUri;
 	private Map<String, Object> attributes;
+	private Instant expiresAt;
 
 	private OAuth2AuthorizationRequest() {
 	}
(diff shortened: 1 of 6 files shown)


oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java
```diff
@@ -55,6 +57,7 @@
 	private Map<String, Object> additionalParameters;
 	private String authorizationRequestUri;
 	private Map<String, Object> attributes;
+	private Instant expiresAt;
```
jgrandja: An Authorization Request does not have an `expiresAt` attribute/parameter as per spec. This really is an implementation detail and should be externalized to the expiring component.

AndreasKl:
> Thanks for the PR @AndreasKl.
> 
> There is quite a bit of coupling in the `OAuth2AuthorizationRequestResolver` and `AuthorizationRequestRepository` and it's implementations (and Reactive). I feel we should rework things to remove this coupling and potentially allow for reuse in other areas within the OAuth support, eg. expiring PK from a JWK Cache.
> 
> The expiring capability is really a cross-cutting concern. Can you try the following solution?
> 
>     * Intercept calls to `loadAuthorizationRequest()` and store the information necessary
> 
>     * At a scheduled time (configurable with skew), the component will attempt to `loadAuthorizationRequest()` and if it returns than determine if time has expired and than `removeAuthorizationRequest`
> 
> 
> Please take a look at the [Cache abstraction](https://docs.spring.io/spring/docs/current/spring-framework-reference/integration.html#cache) to see if it would meet our needs.

If you have time could you have a look at: https://github.com/spring-projects/spring-security/issues/5145#issuecomment-528597139 . I'm now even more convinced that a size based solution could be a better match; we see a few `OAuth2AuthorizationRequest` that live for over 10 minutes on our production system.

Off the top of my head, I think the cache abstraction would impose a quite large runtime requirement to the user of spring security. The user than has to deploy a distributed cache for be able to prevent what is effectively a way to cause an OOM in an application.

I definitely see the need for reducing the duplication and fully understand that `OAuth2AuthorizationRequest` should not contain a `createdAt` or a `expiresAt`, I'll look into other options to store the data and generalize the clean up.

jgrandja:
> I think the cache abstraction would impose a quite large runtime requirement to the user of spring security

I'm not convinced it's the right solution to use either but we should at least rule it out and give reasons for it. I'd prefer a very light solution.

> I'm now even more convinced that a size based solution could be a better match

I still think we should account for time-based as well. For low-load client application, an `OAuth2AuthorizationRequest` could hang out for a long time. Size-based takes care of high-load client applications. I feel we should account for both.

Issue #7418, opened 2019-09-10
ServletBearerExchangeFilterFunction should propagate Authentication


jzheaux:
Related to https://github.com/spring-projects/spring-security/issues/7353#issuecomment-529121870

`ServletBearerExchangeFilterFunction` consults the `SecurityContextHolder` directly when looking up the current `Authentication`. This is because it was initially designed to only read from the main thread.

This is limiting, however, in situations where the `Authentication` may be needed off of the main thread. Consider the toy example below:

```java
ServletBearerExchangeFilterFunction bearer = new ServletBearerExchangeFilterFunction();
WebClient webClient = WebClient.builder()
		.filter((request, next) -> Mono.deferWithContext(ctx -> next.exchange(request)))
		.filter(bearer)
		.build();

return webClient.get()
		.uri("https://endpoint.example.org")
		.retrieve()
		.bodyToMono(String.class)
		.timeout(Duration.ofMillis(10000))
                .retry(3, TimeoutException.class::isInstance)
		.block();
```

With this setup, the `ServletBearerExchangeFilterFunction` will not be executed on the main thread, meaning that `SecurityContextHolder` will not be available.

`ServletBearerExchangeFilterFunction` could be enhanced, though, by introducing a `Hooks.onLastOperator` that adds the `Authentication` to the reactor context so that more complex scenarios like the above work.

PR #7419, opened 2019-09-10
Bearer WebClient Filter Authentication Propagation

jzheaux:
Fixes: gh-7418

### DIFF
--- a/config/src/test/java/org/springframework/security/config/annotation/web/configuration/OAuth2ResourceServerConfigurationTests.java
+++ b/config/src/test/java/org/springframework/security/config/annotation/web/configuration/OAuth2ResourceServerConfigurationTests.java
@@ -1,0 +1,165 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/reactive/function/client/ServletBearerExchangeFilterFunction.java
```diff
@@ -61,14 +61,16 @@
 	}
 
 	private Mono<AbstractOAuth2Token> oauth2Token() {
-		return currentAuthentication()
+		return Mono.subscriberContext()
```
jgrandja: A user looking at this code is going to wonder where the `Authentication` is being populated in the Context. We should javadoc to make it clear.

config/src/test/java/org/springframework/security/config/annotation/web/configuration/OAuth2ResourceServerConfigurationTests.java
```diff
@@ -0,0 +1,91 @@
...
+public class OAuth2ResourceServerConfigurationTests {
+	@Rule
+	public final SpringTestRule spring = new SpringTestRule();
+
+	@Autowired
+	private MockMvc mockMvc;
+
+	// gh-7418
+	@Test
+	public void requestWhenAuthenticatedThenBearerTokenPropagated() throws Exception {
```
jgrandja: In addition to the `MockMvc` tests, we should also add `WebClient` tests. Please add a test that uses `ServletBearerExchangeFilterFunction` and another test that **does not add** `ServletBearerExchangeFilterFunction`. I'm suspecting the `WebClient` test that doesn't use `ServletBearerExchangeFilterFunction` will still populate the Reactor `Context` which will result in a memory leak (as discussed).

Issue #7440, opened 2019-09-16
Reference documentation refers to Java 5.0 and Spring Security 3.0  


andifalk:
### Summary
Current reference documentation tells that "Spring Security 3.0 requires a Java 5.0 Runtime Environment or higher". The whole sentence is wrong: Spring Security 3.0 should be Spring Security 5.2 and minimum Java Runtime must be Java 8 or newer.

### Actual Behavior
In https://docs.spring.io/spring-security/site/docs/5.2.0.RC1/reference/html5/#runtime-environment the docs tell that 

> Spring Security 3.0 requires a Java 5.0 Runtime Environment or higher.

In https://docs.spring.io/spring-security/site/docs/5.2.0.RC1/reference/html5/#core-components the docs tell that

> In Spring Security 3.0, the contents of the spring-security-core jar were stripped down to the bare minimum. 

### Expected Behavior
In https://docs.spring.io/spring-security/site/docs/5.2.0.RC1/reference/html5/#runtime-environment the docs should instead tell that 

> Spring Security 5.2 requires a Java 8 Runtime Environment or higher.

In https://docs.spring.io/spring-security/site/docs/5.2.0.RC1/reference/html5/#core-components the docs should tell that

> As of Spring Security 3.0, the contents of the spring-security-core jar were stripped down to the bare minimum. 

### Version
Spring Security 5.2.0.RC1

PR #7441, opened 2019-09-16
Fix wrong java runtime version mentioned in reference doc

andifalk:
Fixes gh-7440

The reference documentation still referred to Java 5.0 as minimum
runtime version which is wrong.
This commit changes this to the correct
- Java 8 runtime version as required minimum version.
- In addition it corrects a fuzzy wording regarding stripping down the
`spring-security-core` jar.


jzheaux:
Thanks for the PR, @andifalk! This is now merged into `master`.

Issue #7466, opened 2019-09-23
Add ability to customize OAuth2AuthorizationRequestRedirectWebFilter in OAuth2LoginSpec


rchigvintsev:
It would be nice to be able to customize `OAuth2AuthorizationRequestRedirectWebFilter` during configuring of OAuth2 login in the following way: `http.oauth2Login().authorizationRequestRepository(requestRepository)` .

Also for now `OAuth2AuthorizationRequestRedirectWebFilter` creates its own instance of `ServerRequestCache`. I think common `ServerRequestCache` (that is configured using `RequestCacheSpec`) should be used in this case.

PR #7467, opened 2019-09-23
Allow to customize OAuth2AuthorizationRequestRedirectWebFilter in OAuth2LoginSpec

rchigvintsev:
Fixes gh-7466


jgrandja:
Thanks for the PR @rchigvintsev! This is now in master.

I went ahead and applied some polish given that our 5.2 GA is on Monday.
Typically we would have gone through a review process where I would have requested changes from you but there is no time at this point. Feel free to check the polish commit 08d2c93 and let me know if you have any questions.
Thanks again.

rchigvintsev:
@jgrandja I couldn't decide how to write the tests and where to place them for quite a long time. :) Thank you for polishing!

Issue #7422, opened 2019-09-11
Servlet ExchangeFilterFunctions should align


jzheaux:
`ServletOAuth2AuthorizedClientExchangeFilterFunction` and `ServletBearerExchangeFilterFunction` both have the same concern of registering a `CoreSubscriber` that populates request details like the current `Authentication`.

It'd be nice if `ServletOAuth2AuthorizedClientExchangeFilterFunction` did this in the same way as `ServletBearerExchangeFilterFunction`.

PR #7476, opened 2019-09-25
Align Servlet ExchangeFilterFunction CoreSubscriber

### LINKED ISSUE #7423: Refine subscription strategy for Servlet ExchangeFilterFunctions
Related to #7422 

There are two Servlet-based ExchangeFilterFunctions in the code base:

`ServletBearerExchangeFilterFunction`
`ServletOAuth2AuthorizedClientExchangeFilterFunction`

They both rely on the registration of a `CoreSubscriber` that populates the Reactor `Context` with the current `Authentication`.

The subscriber instances and the actual registrations are different. This means that there are two keys in the Reactor Context, both of which contain the current `Authentication` when likely only one would suffice.

jgrandja:
Fixes gh-7422

The intention of the new `SecurityReactorContextConfiguration` is that it's meant to be used/shared by both `ServletOAuth2AuthorizedClientExchangeFilterFunction` and `ServletBearerExchangeFilterFunction`. However, the changes in this PR work only for `ServletOAuth2AuthorizedClientExchangeFilterFunction`.

When this PR is merged than I believe we can remove `OAuth2ResourceServerConfiguration` and apply a very small change in `ServletBearerExchangeFilterFunction`, which will than resolve #7423.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configuration/SecurityReactorContextConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configuration/SecurityReactorContextConfiguration.java
@@ -1,0 +1,143 @@
+/*
+ * Copyright 2002-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.config.annotation.web.configuration;
+
+import org.reactivestreams.Subscription;
+import org.springframework.beans.factory.DisposableBean;
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.web.context.request.RequestContextHolder;
+import org.springframework.web.context.request.ServletRequestAttributes;
+import reactor.core.CoreSubscriber;
+import reactor.core.publisher.Hooks;
+import reactor.core.publisher.Operators;
+import reactor.util.context.Context;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * {@link Configuration} that adds a {@code Publisher} for the last operator created
+ * in every {@code Mono} or {@code Flux}.
+ *
+ * <p>
+ * The {@code Publisher} is solely responsible for adding
+ * the current {@code HttpServletRequest}, {@code HttpServletResponse} and {@code Authentication}
+ * to the Reactor {@code Context} so that it's accessible in every flow, if required.
+ *
+ * @author Joe Grandja
+ * @since 5.2
+ * @see OAuth2ImportSelector
+ */
+@Configuration(proxyBeanMethods = false)
+class SecurityReactorContextConfiguration {
+
+	@Bean
+	SecurityReactorContextSubscriberRegistrar securityReactorContextSubscriberRegistrar() {
+		return new SecurityReactorContextSubscriberRegistrar();
+	}
+
+	static class SecurityReactorContextSubscriberRegistrar implements InitializingBean, DisposableBean {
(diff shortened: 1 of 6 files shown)


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -100,8 +93,10 @@
  * @since 5.1
  * @see OAuth2AuthorizedClientManager
  */
-public final class ServletOAuth2AuthorizedClientExchangeFilterFunction
-		implements ExchangeFilterFunction, InitializingBean, DisposableBean {
+public final class ServletOAuth2AuthorizedClientExchangeFilterFunction implements ExchangeFilterFunction {
```
jgrandja: @rwinch What are your thoughts on removing `InitializingBean` and `DisposableBean`? Just a reminder that this was backported to 5.1.x via this [commit](https://github.com/spring-projects/spring-security/commit/091cbe54585b44b5656cbb33a6d2e29a82e7f05f#diff-972147f7b036129a221ef5e0a9da5886R108). In hindsight, it probably should not have since it introduced new API's in a patch release. So I feel removing it here is fine but want to get your take on this. If you prefer I leave it in than I can simply do no-op in both methods.

rwinch: I would just make them noops

rwinch: Actually in hindsight we are already breaking compatibility because for the functionality to work the same way `SecurityReactorContextSubscriberRegistrar` needs to be added. I'd go ahead and remove it so it is more apparent.

jgrandja: Ok so I'll leave the PR as-is, which has already removed `InitializingBean` and `DisposableBean`

config/src/main/java/org/springframework/security/config/annotation/web/configuration/SecurityReactorContextConfiguration.java
```diff
@@ -0,0 +1,143 @@
...
+	SecurityReactorContextSubscriberRegistrar securityReactorContextSubscriberRegistrar() {
+		return new SecurityReactorContextSubscriberRegistrar();
+	}
+
+	static class SecurityReactorContextSubscriberRegistrar implements InitializingBean, DisposableBean {
+		private static final String SECURITY_REACTOR_CONTEXT_OPERATOR_KEY = "org.springframework.security.SECURITY_REACTOR_CONTEXT_OPERATOR";
+
+		@Override
+		public void afterPropertiesSet() throws Exception {
+			Hooks.onLastOperator(SECURITY_REACTOR_CONTEXT_OPERATOR_KEY,
```
robotmrv: @jgrandja  @rwinch  
as far as function from `Hooks.onLastOperator` is invoked for each last operator in a chain it means that such simple code will print `201`: 
```java
public static void main(String[] args) {
    final AtomicInteger counter = new AtomicInteger();
    Hooks.onLastOperator(pub -> {
        counter.incrementAndGet();
        return pub;
    });
    Flux.range(0, 200)
            .flatMap(it -> Mono.defer(() -> Mono.just(it)))//just do some work
            .blockLast();
    System.out.println("counter: " + counter);
}
```
it means that 200 publishers and probably subscribers and attributes maps will be created just for nothing.  
To reduce allocation this code could be rewritten to 
```java
public void afterPropertiesSet() {
	Function<? super Publisher<Object>, ? extends Publisher<Object>> lifter =
			Operators.liftPublisher((s, sub) -> createRequestContextSubscriberIfNecessary(sub));
	Hooks.onLastOperator(SECURITY_REACTOR_CONTEXT_OPERATOR_KEY, pub -> {
		HttpServletRequest request = null;
		HttpServletResponse response = null;
		ServletRequestAttributes requestAttributes =
				(ServletRequestAttributes) RequestContextHolder.getRequestAttributes();
		if (requestAttributes != null) {
			request = requestAttributes.getRequest();
			response = requestAttributes.getResponse();
		}
		if (authentication == null && request == null && response == null) {
			//do not need to create RequestContextSubscriber with empty data
			return pub;
		}
		return lifter.apply(pub);
	});
}
...
<T> CoreSubscriber<T> createRequestContextSubscriberIfNecessary(CoreSubscriber<T> delegate) {
	if (delegate.currentContext().hasKey(SECURITY_CONTEXT_ATTRIBUTES)) {
        //already enriched. no nead to create Subscriber
		return delegate;
	}
... // the rest logic is the same 
}
```

jgrandja: @robotmrv Your tips on the Reactive end of things has been very valuable. Thank you for keeping an eye out on improvements. I applied your suggested changes :) Thanks again!

jgrandja:
Merged via 2a5bd6e

Issue #7375, opened 2019-09-05
SamlAuthenticationProvider should propagate actual validation errors


fhanik:
Per https://github.com/spring-projects/spring-security/pull/7260#discussion_r319314499 it may be a good idea to propagate an `Saml2AuthenticationException` in the authentication provider with validation details.

This can be caught in the authentication entry point, or other location, to influence UI rendering or messaging.

PR #7477, opened 2019-09-25
propagate saml authentication exception #7375

fhanik:
PowerMock removed as a test dependency. Testing converted using real data.

Fixes gh-7375


saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/authentication/TestSaml2CryptoSupport.java
```diff
@@ -0,0 +1,169 @@
...
+import java.security.NoSuchAlgorithmException;
+import java.security.NoSuchProviderException;
+import java.security.cert.X509Certificate;
+import java.util.List;
+import javax.crypto.SecretKey;
+
+import static java.util.Arrays.asList;
+import static org.opensaml.security.crypto.KeySupport.generateKey;
+
+final class TestSaml2CryptoSupport {
```
jzheaux: We've reserved `TestXYZ` for test domain objects. Maybe this could be called `Saml2CryptoTestSupport`?

Issue #7492, opened 2019-09-29
AbstractUserDetailsReactiveAuthenticationManager default Scheduler should be disposed


robotmrv:
### Summary
AbstractUserDetailsReactiveAuthenticationManager creates new parallel `Scheduler` which creates `Thread`s with `daemon=false`.
```java
Scheduler scheduler = Schedulers.newParallel("password-encoder");
```
As states `Schedulers.newParallel` javadoc 
> daemon – false if the Scheduler requires an explicit Scheduler.dispose() to exit the VM.

### Actual Behavior
default `Scheduler` is not disposed 
1. on `AbstractUserDetailsReactiveAuthenticationManager` bean destruction 
2. or if custom `Scheduler` was set

### Expected Behavior
default `Scheduler` should be disposed 
1. on `AbstractUserDetailsReactiveAuthenticationManager` bean destruction 
2. or if custom `Scheduler` was set

### Version
`5.2.0.BUILD-SNAPSHOT`

rwinch:
@robotmrv I went ahead and changed this fix to use a daemon thread instead. We don't need the thread to block shutting down the JVM.

robotmrv:
@rwinch 
I thought about this solution (https://github.com/spring-projects/spring-security/commit/cb5c58eeaadee97d9b622454428dc6a9169ab1a1#r35174748) 
`Schedulers.newParallel("password-encoder", Schedulers.DEFAULT_POOL_SIZE, true)` 
but there is another problem  
if user wants to refresh application context, default scheduler will not be disposed so unused redundant `Threads` will be present.  
So for this case it should be disposed in anyway

rwinch:
Thanks. Then perhaps it should be a static member variable w/ the default values

robotmrv:
yes, static default `Scheduler` will solve that problem

PR #7493, opened 2019-09-29
Dispose default Scheduler

robotmrv:
AbstractUserDetailsReactiveAuthenticationManager creates parallel `Scheduler` with `daemon=false` `Thread`s. It is recommended to dispose such Schedulers to be able exit the VM

Fixes gh-7492


rwinch:
Thanks for the PR @robotmrv! This is now merged into master with a little polish on the implementation details fc8a0184b0622b00e1521ab126c05de1350c13ba

Issue #7513, opened 2019-10-07
SAML2 Provider AuthNRequest Hardcoded Protocol Binding


blucas:
### Summary

When acting as a SP sending an `AuthnRequest` to an IdP, the request is created with a hard-coded [protocol binding](https://github.com/spring-projects/spring-security/blob/master/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationRequestFactory.java#L46).  I was testing against http://samltest.id and it looks like they validate that the binding used is `HTTP-POST`.  I am a novice at SAML.  Googling around, it sounds like `HTTP-POST` is the recommended approach to responding to an `AuthnRequest`. Here is an [example](https://github.com/web2py/web2py/pull/2199) and [another](https://stackoverflow.com/a/2239909).

### Actual Behavior

IdP validates protocol-binding and rejects `AuthnRequest`:
```
2019-10-07 00:40:38,302 - DEBUG [org.opensaml.saml.common.binding.impl.DefaultEndpointResolver:?] - Endpoint Resolver org.opensaml.saml.common.binding.impl.DefaultEndpointResolver: Candidate endpoint binding 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST' did not match 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect'
2019-10-07 00:40:38,302 - DEBUG [org.opensaml.saml.common.binding.impl.DefaultEndpointResolver:?] - Endpoint Resolver org.opensaml.saml.common.binding.impl.DefaultEndpointResolver: Candidate endpoint binding 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect' not permitted by input criteria
```

You can see by the last log statement that `HTTP-Redirect` is not permitted

### Expected Behavior

The SP should use `HTTP-POST` protocol-binding when creating an `AuthnRequest` so that validation on the IdP side passes.

### Version

Spring-Security 5.2.0
Spring-Boot 2.2.0.RC-1

@fhanik FYI.  Also, thanks for contributing this.  It is very much appreciated.

fhanik:
@blucas Thank you for the report. Glad to already see feedback.

I will mark this as an _enhancement_, and look into a configuration option for POST vs REDIRECT.
The work around right now is to inject your own `Saml2AuthenticationRequestFactory` object.

[...]

PR #7516, opened 2019-10-08
Support configuration of protocol binding for authentication requests

blucas:
Fixes gh-7513.

Note: This is not intended as the final solution, but as a starting point for discussion.  I believe a better solution may be to add `protocolBinding` to `Saml2AuthenticationRequest` and `RelyingPartyRegistration`.  But I'd like to hear feedback on that approach before submitting any code.

@fhanik FYI.

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationRequestFactory.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationRequestFactory.java
@@ -16,6 +16,7 @@
 
 package org.springframework.security.saml2.provider.service.authentication;
 
+import org.opensaml.saml.common.xml.SAMLConstants;
 import org.springframework.util.Assert;
 
 import org.joda.time.DateTime;
@@ -32,6 +33,7 @@ import java.util.UUID;
 public class OpenSamlAuthenticationRequestFactory implements Saml2AuthenticationRequestFactory {
 	private Clock clock = Clock.systemUTC();
 	private final OpenSamlImplementation saml = OpenSamlImplementation.getInstance();
+	private String protocolBinding = SAMLConstants.SAML2_POST_BINDING_URI;
 
 	/**
 	 * {@inheritDoc}
@@ -43,7 +45,7 @@ public class OpenSamlAuthenticationRequestFactory implements Saml2Authentication
 		auth.setIssueInstant(new DateTime(this.clock.millis()));
 		auth.setForceAuthn(Boolean.FALSE);
 		auth.setIsPassive(Boolean.FALSE);
-		auth.setProtocolBinding("urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect");
+		auth.setProtocolBinding(protocolBinding);
 		Issuer issuer = this.saml.buildSAMLObject(Issuer.class);
 		issuer.setValue(request.getIssuer());
 		auth.setIssuer(issuer);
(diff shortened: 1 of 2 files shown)


fhanik:
Awesome @blucas 

Can you add a test case to your pull request, you can use this one so you get familiar with our style:
https://github.com/fhanik/spring-security/commit/f195ac7f67cb2d4a5a28ab5006121c0f92bea816

The naming convention for methods is
1. method name of method being tested
2. 'When'
3. condition for tests
4. 'Then'
5. expected results

blucas:
@fhanik - I didn't add a test as I wasn't convinced this approach was the best solution to the problem.  Do you have any thoughts around exposing the `protocolBinding` to `Saml2AuthenticationRequest` and `RelyingPartyRegistration`?

blucas:
@fhanik - I've added the test and fixed the failure.  Please consider my proposal around exposing the `protocolBinding` to `Saml2AuthenticationRequest` and `RelyingPartyRegistration`

fhanik:
Hi @blucas ,

I've spent some time thinking about this. And the different options as you noted are

1. Just default to POST, no configuration option
2. Allow the `OpenSamlAuthenticationRequestFactory` to be configured for POST/REDIRECT
3. Expose it in the `Saml2AuthenticationRequest` object
4. Expose it in the `RelyingPartyRegistration` object
5. Make is available in Spring Boot's application.yml file

The further down the list we reach, the broader the impact it has on the overall codebase.

Then I started thinking about the use case. It is my belief that `POST` will probably cover 95%+ of all use cases. I can see why `REDIRECT` may be frowned upon because you're risking an assertion to be shared elsewhere through the browser, for example via the `Referrer` header.

I believe that option 1 and 2 are both sufficient. I believe going further down the list, why it may seem flexible, it's a lot of work for very little gain. Specially if we don't see this value change frequently. So I'm thinking a minimalist approach for the sake of simplicity and not creating an overarching configuration base.

What are your thoughts? Am I missing something ?

blucas:
@fhanik I completely agree with everything you said.  I do believe `POST` as default is the right direction and covers majority of use cases.  Additionally it doesn't expose the SAML object to other services (such as access logs).

I feel like the `setter` approach provided in this PR is the bare minimum we can propose and will suffice for my purposes at the very least.

The reason I mention exposing it further (3 and 4 above) is it feels more like the "spring way" of doing things.  Plus is provides more specific configuration (one RPR could use `POST` while another could use `REDIRECT`).  But having said that, I don't think many people would make sure of that configurability/flexibility.

In summary.  I believe the approach of this PR is enough.  Any further configuration requests could be handled at a later date.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationRequestFactory.java
```diff
@@ -67,4 +70,18 @@ public void setClock(Clock clock) {
 		Assert.notNull(clock, "clock cannot be null");
 		this.clock = clock;
 	}
+
+	/**
+	 * Sets the {@code protocolBinding} to use when generating authentication requests
+	 *
```
fhanik: Can you add to this documentation
1. What values are acceptable
2. @throws for when invalid parameter is set

I wonder if we should switch to `IllegalArgumentException` since this is not really SAML, but just a simple getter that takes a value.

saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationRequestFactoryTests.java
```diff
@@ -0,0 +1,75 @@
...
+	@Test
+	public void createAuthenticationRequestWhenSetUriThenReturnsCorrectBinding() {
+		factory.setProtocolBinding(SAMLConstants.SAML2_REDIRECT_BINDING_URI);
+		AuthnRequest authn = getAuthNRequest();
+		Assert.assertEquals(SAMLConstants.SAML2_REDIRECT_BINDING_URI, authn.getProtocolBinding());
+	}
+
+	@Test
+	public void createAuthenticationRequestWhenSetUnsupportredUriThenThrowsSaml2Exception() {
+		exception.expect(Saml2Exception.class);
```
fhanik: Is `IllegalArgumentException` perhaps better

fhanik:
Thank @blucas , I'll check out the CI failures tomorrow

blucas:
@fhanik - I'm not too sure why the build is failing, I haven't touched CAS code... thanks for looking into it for me.

PR #7547, opened 2019-10-19
Add support for multiple JWK Set URLs

seedhartha:
This is required when using Spring Boot application as a Resource Server
together with Keycloak, where each realm has it's own JWK Set.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
@@ -1833,6 +1833,17 @@ public class ServerHttpSecurity {
 				return this;
 			}
 
+			/**
+			 * Configures a {@link ReactiveJwtDecoder} using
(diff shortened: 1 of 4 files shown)


jzheaux:
@seedhartha Thanks for your interest in making Spring Security better.

It's not really in the best interest of a resource server to hit many endpoints to find one that accepts a token. It's possible that two endpoints, for example, are served in the end by the same provider and may not be using GUIDs as `kid` values (causing a collision).

This also causes a problem from a performance perspective in the case of a new key from one provider. When the lookup fails, the cache gets invalidated and needs to be repopulated.

What I'd recommend instead is to correlate the information provided in the JWT [against a whitelist of `AuthenticationManager`s](https://github.com/jzheaux/springone2019/commit/02ba58d0a06ea2e11e3b66384a0168c67a291f44#diff-860d7d887699c901e6f3103d7db4f117R28) to determine which endpoint to hit.

You could do [the same kind of thing with `JwtDecoder`](https://github.com/spring-projects/spring-security/pull/6779/files#diff-8695b60cf3fb4e49ace5677e3dae5f89R19) if you wish.

Issue #7546, opened 2019-10-18
DefaultReactiveOAuth2AuthorizedClientManager.saveAuthorizedClient does not save authorized client


philsttr:
## Summary

`DefaultReactiveOAuth2AuthorizedClientManager.saveAuthorizedClient` never actually saves the authorized client, because it ignores the `Mono<Void>` returned from `authorizedClientRepository.saveAuthorizedClient`

### Actual Behavior

`DefaultReactiveOAuth2AuthorizedClientManager.saveAuthorizedClient` does not save the authorized client.

https://github.com/spring-projects/spring-security/blob/6ad328f909525bedabbc9ac4978d7ee7f0e84ced/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultReactiveOAuth2AuthorizedClientManager.java#L108-L111

`this.authorizedClientRepository.saveAuthorizedClient` returns a `Mono<Void>`, which is ignored, and never included in the stream, and therefore never subscribed.

### Expected Behavior

`DefaultReactiveOAuth2AuthorizedClientManager.saveAuthorizedClient` saves the authorized client.

Perhaps:

```java
	.flatMap(exchange -> 
		this.authorizedClientRepository.saveAuthorizedClient(authorizedClient, principal, exchange)
			.thenReturn(authorizedClient))
```

### Configuration

n/a

### Version

Spring Boot 2.2.0.RELEASE
Spring Security 5.2.0.RELEASE

### Sample

n/a

PR #7551, opened 2019-10-20
Make saveAuthorizedClient save the authorized client

philsttr:
Make `DefaultReactiveOAuth2AuthorizedClientManager.saveAuthorizedClient` actually save the authorized client

Previously, `saveAuthorizedClient` never actually saved the authorized client, because it ignored the `Mono<Void>` returned from `authorizedClientRepository.saveAuthorizedClient`.

Now, it does not ignore the `Mono<Void>` returned from `authorizedClientRepository.saveAuthorizedClient`, and includes it in the stream, and therefore it will properly save the authorized client.

Added logic to unit test to verify the `Mono<Void>` returned from `authorizedClientRepository.saveAuthorizedClient` is subscribed.

Fixes gh-7546

I have signed the CLA


jgrandja:
Thank you for catching that @philsttr and for providing the fix. This is now in master.

Issue #7153, opened 2019-07-26
Deprecate OpenID 2.0 Support


rwinch:
### Summary

The OpenID 1.0 and 2.0 protocols have been deprecated and users are [encouraged to migrate](https://openid.net/specs/openid-connect-migration-1_0.html) to [OpenID Connect](https://openid.net/connect/). The [openid4java](https://github.com/jbufu/openid4java) library that is used by Spring Security has not seen an [update since 2015 (4 years)](https://github.com/jbufu/openid4java/releases). We need to deprecate the OpenID support.

We should add the following deprecation notice:

> The OpenID 1.0 and 2.0 protocols have been deprecated and users are [encouraged to migrate](https://openid.net/specs/openid-connect-migration-1_0.html) to [OpenID Connect](https://openid.net/connect/).

To the following locations:

* Any files in the spring-security-openid module
* The openid portion of the reference 
* The openid sample application
* The openid related code in spring-security-config for bot Java Configuration and XML configuration

NOTE: We are NOT deprecating OpenID Connect support. Only the support for OpenID 2.0 protocol will be deprecated (the protocol itself was deprecated).

eddumelendez:
Hi @rwinch, do you think this is good for `first-timers-only`?

rwinch:
Hey @eddumelendez Thanks for the question. I think this is a lot of files to touch for a first timers issue.

ThomasVitale:
Hi @rwinch, I would be available to take this issue.

rwinch:
Thanks @ThomasVitale! The issue is yours if you still want it

ThomasVitale:
I will start working on it then, thanks @rwinch

farooqkhan003:
@ThomasVitale by any chance if you didn't get time to work on this, can I take this issue?

ThomasVitale:
Hi @farooqkhan003, I have started with it and then didn't have time to complete it. So far I have added the deprecation notice to all files in the openid module. Would you like to take over and complete the remaining parts of the task? If so, @rwinch can we split the delivery in 2 parts or should it be a single one? In any case, I have created a PR with the changes I have done so far.

dadikovi:
@farooqkhan003 Are you still working on this issue? If not, @rwinch can I take it?

farooqkhan003:
@dadikovi unfortunately I didn't get chance to work on this issue.

rwinch:
@dadikovi Please take it. The issue is yours

[...]

PR #7554, opened 2019-10-22
Deprecate OpenID 2.0 Support

ThomasVitale:
Add deprecation notice to all files in the spring-security-openid module

Fixes gh-7153


ThomasVitale:
The commits in this PR have been delivered as part of https://github.com/spring-projects/spring-security/pull/8450

Issue #7548, opened 2019-10-19
DelegatingReactiveAuthenticationManager signals error instead of continuing


DarrenForsythe:
### Summary

The `DelegatingReactiveAuthenticationManager` should continue to the next authentication manger on error, rather than signalling an error and stopping the stream.

### Actual Behavior

Adding Opaque and JWT support to a WebFlux resource server via, 

```
    public class MultiTokenSupport
            implements ReactiveAuthenticationManagerResolver<ServerHttpRequest> {

        private final ReactiveAuthenticationManager authenticationManager;

        public MultiTokenSupport(
                JwtReactiveAuthenticationManager jwtAuthenticationProvider,
                OpaqueTokenReactiveAuthenticationManager opaqueTokenAuthenticationProvider) {
            authenticationManager = new DelegatingReactiveAuthenticationManager(jwtAuthenticationProvider,
                    opaqueTokenAuthenticationProvider);
        }

        @Override
        public Mono<ReactiveAuthenticationManager> resolve(ServerHttpRequest serverHttpRequest) {
            return Mono.just(authenticationManager);
        }
    }
```

JWT tokens will work as expected, however Opaque tokens will fail as the `JwtReactiveAuthenticationManager` creates an error for the Opaque token.

### Expected Behavior

The DelegatingReactiveAuthenticationManager should continue the flux instead of singnalling an error from an AuthenticationManager. Which would mirror the Servlet/Web implementation closer.

A rough impl. that does work, 

```
        authenticationManager = authentication -> Flux
                .fromIterable(Arrays.asList(jwtAuthenticationProvider,
                        opaqueTokenAuthenticationProvider))
                .concatMapDelayError(reactiveAuthenticationManager -> reactiveAuthenticationManager
                        .authenticate(authentication)
                .next();
```

This solution does introduce more complexities around the error handling as there maybe 0 to many errors to unwrap, which may be best to be wrapped in a custom exception that Spring Security would still return a 401. 

### Version

Spring Boot 2.2.0, Spring Security 5.2

DarrenForsythe:
I also noted this in the test class, 

> // delay to try and force delegate2 to finish (i.e. make sure we didn't use flatMap)

If the second delegate was finishing an error would occur as it would return null instead of Mono.empty() causing an NPE to be throwing by reactor. Is there an expectation that all the elements would be applied async and the first one found returned?

jzheaux:
[...]

PR #7555, opened 2019-10-22
Do not short circuit DelegatingReactiveAuthenticationManager.

DarrenForsythe:
Currently the DelegatingReactiveAuthenticationManager will short cirtcuit on an error signaled from a delegated reactive authentication manager. This pull requests switches the concatMap to concatMapDelayError method call to invoke each delegated authentication manager.

If there is no authentication provided as a result the last error signaled is used.

Fixes gh-7548


jzheaux:
Thanks, again, @DarrenForsythe, for working to make Spring Security better.

I had a quick chat with @rwinch about this, and he confirmed that `DelegatingReactiveAuthenticationManager`'s current short-circuiting behavior is intentional -- if a delegate authentication manager returns `Mono.empty()`, then it will move on to the next authentication manager in the list.

Since the class you've indicated is behaving as designed, I'm going to close the PR.

Issue #7556, opened 2019-10-23
UserCache in DaoAuthenticationProvider Auto Configuration


rastadrian:
### Summary

When the AuthenticationConfiguration builds the AuthenticationManager along with a DaoAuthenticationProvider.

There's not a straightforward way to set a `UserCache` in that `DaoAuthenticationProvider` as it is not Spring managed, it is just instantiated by the `InitializeUserDetailsBeanManagerConfigurer`.

### Actual Behavior

If a UserCache is registered to the application's context, the Authentication Provider Configurer won't set said UserCache into the Provider.

### Expected Behavior

**Given** that an `UserCache` is registered in the `ApplicationContext`.
**When** the `AuthenticationConfiguration` configures the `InitializeUserDetailsBeanManagerConfigurer` .
**Then** the Configurer should set the registered `UserCache` in the `DaoAuthenticationProvider` that is created for the `AuthenticationManagerBuilder`.

### Configuration

A project configured with these dependencies:

```
spring-boot-starter-web
spring-boot-starter-security
```
A SecurityConfiguration with `@EnableWebSecurity`.

A registered implementation of an `UserDetailsService` and of an `UserCache`.

### Version

Spring Framework 5.2.0.RELEASE
Spring Boot 2.2.0.RELEASE

Also present in older versions.

### Sample

https://github.com/rastadrian/spring-security-usercache-sample

rwinch:
We try and focus on the the most frequently used beans for automatic detection. I don't view `UserCache` as such a type. If you need to configure a `UserCache`, expose a `DaoAuthenticationProvider` as a Bean with the `UserCache`.

PR #7557, opened 2019-10-23
Adds UserCache to the DaoAuthenticationprovider through AuthenticationConfiguration

rastadrian:
When the AuthenticationConfiguration sets the UserDetailsBeanConfigurer it creates a DaoAuthenticationProvider without the capabilities of setting an ApplicationContext registered UserCache bean.

Fixes gh-7556


eleftherias:
Thanks for the PR @rastadrian.
Given what was mentioned in the issue gh-7556, about only focusing on the most frequently used beans for automatic detection, we do not want to set the user cache on the provider.
I noticed that you also made an enhancement to the Javadoc in the `InitializeUserDetailsBeanManagerConfigurer` class. We can merge the Javadoc change if you want to update your PR to only include that line.

rastadrian:
Hi @eleftherias, I have updated the PR to only include the javadoc change.

eleftherias:
Thanks for the speedy update @rastadrian! This is now merged into master.

PR #7559, opened 2019-10-23
Fix typo 'is' -> 'if' in javadoc

rand0m86:
Just a typo fix in javadoc for `authenticationEntryPoint` method. Nothing fancy here.


jzheaux:
Thanks, @rand0m86, this is now merged into `master`.

PR #7562, opened 2019-10-23
Make save and load authorized clients work without exchange

### LINKED ISSUE #7544: DefaultReactiveOAuth2AuthorizedClientManager never calls UnAuthenticatedServerOAuth2AuthorizedClientRepository
### Summary

When using `UnAuthenticatedServerOAuth2AuthorizedClientRepository` together with the `DefaultReactiveOAuth2AuthorizedClientManager` authorized clients are not correctly loaded/saved.

This seems to relate to #7468 (@jgrandja )

It seems that when using the `UnAuthenticatedServerOAuth2AuthorizedClientRepository` it is (and can) never be called from `DefaultReactiveOAuth2AuthorizedClientManager` as it tries to `flatMap` a ` Mono<ServerWebExchange`  (in the method `loadAuthorizedClient`). In situations there `UnAuthenticatedServerOAuth2AuthorizedClientRepository` can/should be used the WebExchange well be null/empty the `Mono` will be empty and therefore the code in `flatMap` will not execute.
If `ServerWebExchange` would be present `UnAuthenticatedServerOAuth2AuthorizedClientRepository` would throw an Exception there therefore needs to be a way for the ClientManager to call the repository without a WebExchange.

### Version
Spring Security 5.2.0 via Spring Boot 2.2.0

frzme:
Previously saveAuthorizedClient  and loadAuthorizedClient would not save nor load the authorized client when no current serverWebExchange was available. In case of the UnAuthenticatedServerOAuth2AuthorizedClientRepository it is however nescessary for the save and load call to also happen without and exchange.

I'm not entirely certain about the impact of the behavior change I introduced in the test `authorizeWhenNotAuthorizedAndSupportedProviderAndExchangeUnavailableThenAuthorizedButNotSaved` -> `authorizeWhenNotAuthorizedAndSupportedProviderAndExchangeUnavailableThenAuthorizedAndSaved`

Please verify

Fixes #7544 

I have signed the CLA

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultReactiveOAuth2AuthorizedClientManager.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultReactiveOAuth2AuthorizedClientManager.java
@@ -34,6 +34,7 @@ import reactor.core.publisher.Mono;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.Function;
 
 /**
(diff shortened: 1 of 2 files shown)


jgrandja:
Closing. Please see [comment](https://github.com/spring-projects/spring-security/issues/7544#issuecomment-548800979).

Issue #7512, opened 2019-10-06
ClientRegistrations.fromIssuerLocation for Oauth2 AuthorizationServer requires jwks url even though jwks is not required in the metadata spec


knutejoh:
### Summary
When using ClientRegistrations.fromIssuerLocation for setting up Oauth2 AuthorizationServer the code requires jwks url to be a part of the returned metadata in the .well-known/oauth-authorization-server  even though this is not required in the metadata spec (see https://tools.ietf.org/html/rfc8414) 

### Actual Behavior
Nullpointer thrown from line 222 (.jwkSetUri(metadata.getJWKSetURI().toASCIIString())) in org.springframework.security.oauth2.client.registration.ClientRegistrations

### Expected Behavior
No nullpointer and the client configured correctly with the provided metadata

### Configuration
Here is an example metadata file that should work
{
  "issuer": "https://issuerurl:port",
  "authorization_endpoint": "https://issuerurl:port/oauth/authorize",
  "token_endpoint": "https://issuerurl:port/oauth/token",
  "scopes_supported": [
    "user:check-access",
    "user:full",
    "user:info",
    "user:list-projects",
    "user:list-scoped-projects"
  ],
  "response_types_supported": [
    "code",
    "token"
  ],
  "grant_types_supported": [
    "authorization_code",
    "implicit"
  ],
  "code_challenge_methods_supported": [
    "plain",
    "S256"
  ]
}

### Version
Spring Securiy Oauth2 Client 5.2.0.RELEASE

rhamedy:
I remember working on `ClientRegistrations.fromIssuerLocation` with @jzheaux help, however, the part that throws `NPE` wasn't touched. 

Both [JwtDecodersTests.java](https://github.com/spring-projects/spring-security/blob/master/oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/JwtDecodersTests.java) and [ClientRegistrationsTests.java](https://github.com/rhamedy/spring-security/blob/3c1eedb3b8dc6bbccf698f6281839544a9853132/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/registration/ClientRegistrationsTest.java) have a `DEFAULT_RESPONSE_TEMPLATE` that includes a value for `jwks_uri` field. The specification definitely marked `jwks_uri` as `OPTIONAL`. 

I would love to make a pull request should we decide to refactor/add null-check 🙂

jzheaux:
[...]

PR #7573, opened 2019-10-27
Make jwks_uri optional for RFC 8414 and required for OpenID Connect

rhamedy:
Relating to https://github.com/spring-projects/spring-security/issues/7512

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
@@ -21,6 +21,7 @@ import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Supplier;
 
 import com.nimbusds.oauth2.sdk.GrantType;
 import com.nimbusds.oauth2.sdk.ParseException;
@@ -86,10 +87,7 @@ public final class ClientRegistrations {
 	 */
 	public static ClientRegistration.Builder fromOidcIssuerLocation(String issuer) {
 		Assert.hasText(issuer, "issuer cannot be empty");
-		Map<String, Object> configuration = getConfiguration(issuer, oidc(URI.create(issuer)));
-		OIDCProviderMetadata metadata = parse(configuration, OIDCProviderMetadata::parse);
-		return withProviderConfiguration(metadata, issuer)
-				.userInfoUri(metadata.getUserInfoEndpointURI().toASCIIString());
+		return getBuilder(issuer, oidc(URI.create(issuer)));
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


rhamedy:
Hi @jzheaux, I would appreciate the feedback. 

I also noticed that the `OIDCProviderMetadata.class` constructor has the following validation

```
if (jwkSetURI == null)
	throw new IllegalArgumentException("The public JWK set URI must not be null");
``` 

As per our conversation on the issue, I have pushed my changes and kept the error message the same as above unless you want it to be different.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
```diff
@@ -137,42 +135,89 @@
...
+	private static Supplier<ClientRegistration.Builder> oidcRfc8414(URI issuer) {
+		URI uri = UriComponentsBuilder.fromUri(issuer)
 				.replacePath(OIDC_METADATA_PATH + issuer.getPath()).build(Collections.emptyMap());
+		return () -> {
+			RequestEntity<Void> request = RequestEntity.get(uri).build();
+			Map<String, Object> configuration = rest.exchange(request, typeReference).getBody();
+			AuthorizationServerMetadata metadata = parse(configuration, AuthorizationServerMetadata::parse);
+			validateJwksUri(metadata);
+			ClientRegistration.Builder builder = withProviderConfiguration(metadata, issuer.toASCIIString())
+					.jwkSetUri(metadata.getJWKSetURI().toASCIIString());
```
jzheaux: According to RFC 8414, it's okay for jwk set uri to be `null`, so we should only set it on the builder if it is there in the response.

It makes me also wonder if there are any other properties that are required by OIDC discovery, but are optional in RFC 8414.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
```diff
@@ -137,42 +135,89 @@
...
-	private static Map<String, Object> getConfiguration(String issuer, URI... uris) {
+	@SafeVarargs
+	private static ClientRegistration.Builder getBuilder(String issuer, Supplier<ClientRegistration.Builder>... suppliers) {
 		String errorMessage = "Unable to resolve Configuration with the provided Issuer of \"" + issuer + "\"";
-		for (URI uri : uris) {
+		for (Supplier<ClientRegistration.Builder> supplier : suppliers) {
 			try {
-				RequestEntity<Void> request = RequestEntity.get(uri).build();
-				return rest.exchange(request, typeReference).getBody();
+				ClientRegistration.Builder builder = supplier.get();
```
jzheaux: We could probably simplify this a bit:

```suggestion
				return supplier.get();
```

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
rhamedy: > According to RFC 8414, it's okay for jwk set uri to be null, so we should only set it on the builder if it is there in the response.

In that case, the block of code could be reused by both `oidcRfc8414` and `oauth`.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
rhamedy: 👍

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
rhamedy: I had a look at the [OIDC Discovery](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationResponse) and [RFC 8414](https://tools.ietf.org/html/rfc8414#section-2). Turns out the `jwks_uri` seem to be the only one.

rhamedy:
@jzheaux pushed my changes relating to your feedback. I will squash my commits once there are no other changes needed to make. 👍

rhamedy:
I find it difficult to come up with good method names. If you think the new method names could be improved, then please feel free to suggest.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
```diff
@@ -86,10 +87,7 @@
 	 */
 	public static ClientRegistration.Builder fromOidcIssuerLocation(String issuer) {
 		Assert.hasText(issuer, "issuer cannot be empty");
-		Map<String, Object> configuration = getConfiguration(issuer, oidc(URI.create(issuer)));
-		OIDCProviderMetadata metadata = parse(configuration, OIDCProviderMetadata::parse);
-		return withProviderConfiguration(metadata, issuer)
-				.userInfoUri(metadata.getUserInfoEndpointURI().toASCIIString());
+		return oidc(URI.create(issuer)).get();
```
jzheaux: I think we still want to route this through `getBuilder` since that keeps the exception handling the same as before.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
```diff
@@ -137,42 +135,80 @@
...
+		return getRfc8414Builder(issuer, uri);
+	}
+
+	private static Supplier<ClientRegistration.Builder> getRfc8414Builder(URI issuer, URI uri) {
+		return () -> {
+			RequestEntity<Void> request = RequestEntity.get(uri).build();
+			Map<String, Object> configuration = rest.exchange(request, typeReference).getBody();
+			AuthorizationServerMetadata metadata = parse(configuration, AuthorizationServerMetadata::parse);
+			ClientRegistration.Builder builder = withProviderConfiguration(metadata, issuer.toASCIIString())
+					.jwkSetUri(metadata.getJWKSetURI() != null ? metadata.getJWKSetURI().toASCIIString() : null);
```
jzheaux: Let's follow a similar construct here as was done with userInfoEndpoint. I believe that will make it a bit easier to read.

```suggestion
			URI jwkSetUri = metadata.getJWKSetURI();
			if (jwkSetUri != null) {
			    builder.jwkSetUri(jwkSetUri.toASCIIString());
			}
```

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
```diff
@@ -137,42 +135,80 @@
...
-		return UriComponentsBuilder.fromUri(issuer)
+	private static Supplier<ClientRegistration.Builder> oidc(URI issuer) {
+		URI uri = UriComponentsBuilder.fromUri(issuer)
 				.replacePath(issuer.getPath() + OIDC_METADATA_PATH).build(Collections.emptyMap());
+
+		return () -> {
+			RequestEntity<Void> request = RequestEntity.get(uri).build();
+			Map<String, Object> configuration = rest.exchange(request, typeReference).getBody();
+			OIDCProviderMetadata metadata = parse(configuration, OIDCProviderMetadata::parse);
+			validateJwksUri(metadata);
```
jzheaux: As you pointed out, the `OIDCProviderMetadata` constructor checks for this already, so I think we are guaranteed at this point that the JWK set URI won't be `null`, or am I missing something?

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/registration/ClientRegistrationsTest.java
```diff
@@ -168,6 +168,30 @@ private void assertIssuerMetadata(ClientRegistration registration,
 				"grant_types_supported", "token_endpoint", "token_endpoint_auth_methods_supported", "userinfo_endpoint");
 	}
 
+	@Test
```
jzheaux: To conform with other bug resolutions, would you please add a comment above each of these tests indicating the ticket that it resolves?

```suggestion
	// gh-7512
	@Test
```

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
rhamedy: Good catch! 👍

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
rhamedy: I should have not doubted my first suggestion. I agree, easier to read.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
rhamedy: We could definitely remove that and rely on `OIDCProviderMetadata` to always validate that. I added it as an extra validation. 

https://github.com/spring-projects/spring-security/pull/7573#issuecomment-547732488

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/registration/ClientRegistrationsTest.java
rhamedy: 👍

rhamedy:
Hi @jzheaux, replied to you comments and pushed the recent changes up 👍

rhamedy:
Thank you for approving the changes. I have squashed my commits 👍

jzheaux:
Thanks, @rhamedy! This is now merged into `master`.

PR #7590, opened 2019-10-31
Clarify usage of hasAnyRole and hasAnyAuthority

tinexw:
A valid `@PreAuthorize` expression with `hasAnyRole` looks like this:
```@PreAuthorize("hasAnyRole('USER', 'ADMIN')")```

However, this was not quite clear to me from the documentation. It states
`hasAnyRole([role1,role2])`. I initially thought that I had to add brackets around the complete list of roles e.g. `@PreAuthorize("hasAnyRole(['USER', 'ADMIN'])")`.

This leads to an exception though:
```Caused by: org.springframework.expression.spel.SpelParseException: Expression [hasAnyRole(['USER', 'ADMIN'])] @18: EL1043E: Unexpected token. Expected 'rsquare(])' but was 'comma(,)'```

which makes it clear that the brackets are wrong.

I think it would be clearer if the documentation would use this expression: `hasAnyRole([role1],[role2])`.

### DIFF
--- a/docs/manual/src/docs/asciidoc/_includes/servlet/authorization/expression-based.adoc
+++ b/docs/manual/src/docs/asciidoc/_includes/servlet/authorization/expression-based.adoc
@@ -21,21 +21,31 @@ This provides some common expressions which are available in both web and method
 |===
 | Expression | Description
 
-| `hasRole([role])`
+| `hasRole(String role)`
 | Returns `true` if the current principal has the specified role.
+
+For example, `hasRole('admin')`
+
 By default if the supplied role does not start with 'ROLE_' it will be added.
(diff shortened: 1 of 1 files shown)


eleftherias:
Thanks for the PR @tinexw!

I can see that the documentation is not very clear.
Perhaps we can follow the example of `hasPermission(Object target, Object permission)` which is in the same section.
Then the expressions would be `hasRole(String role)` and similarly `hasAnyRole(String... roles)`.

It may also be useful to add an example in the description (similar to `hasPermission`).

tinexw:
Good suggestion @eleftherias  I changed it accordingly.

eleftherias:
Thanks for the PR @tinexw! This is now merged into master.

PR #7620, opened 2019-11-04
Oidclogin test support

jzheaux:


test/src/main/java/org/springframework/security/test/web/support/WebMvcTestUtils.java
```diff
@@ -0,0 +1,70 @@
...
+import org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter;
+
+/**
+ * A utility class for testing applications that use Spring Security's Web MVC support
+ *
+ * @author Josh Cummings
+ * @since 5.3
+ */
+public class WebMvcTestUtils {
+	public static void registerArgumentResolver(HttpServletRequest request, HandlerMethodArgumentResolver argumentResolver) {
```
jzheaux: @rwinch I like the idea of being able to work with controllers that use `@RegisteredOAuth2AuthorizedClient`; however, I'm concerned about how well this approach will work in a multi-threaded test environment.

I could resolve it with a `synchronized` block that locks on the non-null `adapter` instance, but I wonder if you can see a better approach.

The consequence, as far as I can tell, is that multiple threads could evaluate `!containsArgumentResolver` to `true` simultaneously, and each add an instance of `argumentResolver` to the list. The tests would still work, but the resulting list of resolves seems a bit messy.

jzheaux:
@rwinch, obviously feel free to comment on anything you like. I'm posting this PR so that you can take a look at a couple of multi-threading concerns I have.

test/src/main/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestPostProcessors.java
```diff
@@ -1052,6 +1078,176 @@ public MockHttpServletRequest postProcessRequest(MockHttpServletRequest request)
...
+			return this;
+		}
+
+		@Override
+		public MockHttpServletRequest postProcessRequest(MockHttpServletRequest request) {
+			OidcUser oidcUser = getOidcUser();
+			OAuth2AuthenticationToken token = new OAuth2AuthenticationToken(oidcUser, oidcUser.getAuthorities(), this.clientRegistration.getRegistrationId());
+			OAuth2AuthorizedClient client = new OAuth2AuthorizedClient(this.clientRegistration, token.getName(), this.accessToken);
+
+			ClientRegistrationRepository clients = new InMemoryClientRegistrationRepository(this.clientRegistration);
```
jzheaux: Again regarding multi-threading, @rwinch, I believe this to be thread-safe since it is not mutable state, e.g. there's no save method on `ClientRegistrationRepository` and the tester can't supply a `ClientRegistration` instance.

However, if these features were added, I'd be at a loss for how to properly inject a `ClientRegistrationRepository` in a thread-safe manner.

test/src/main/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestPostProcessors.java
```diff
@@ -1024,6 +1045,161 @@ public MockHttpServletRequest postProcessRequest(MockHttpServletRequest request)
...
+			this.clientRegistration = clientRegistration;
+			return this;
+		}
+
+		@Override
+		public MockHttpServletRequest postProcessRequest(MockHttpServletRequest request) {
+			OidcUser oidcUser = getOidcUser();
+			OAuth2AuthenticationToken token = new OAuth2AuthenticationToken(oidcUser, oidcUser.getAuthorities(), this.clientRegistration.getRegistrationId());
+			OAuth2AuthorizedClient client = new OAuth2AuthorizedClient(this.clientRegistration, token.getName(), this.accessToken);
+			OAuth2AuthorizedClientRepository authorizedClientRepository = new HttpSessionOAuth2AuthorizedClientRepository();
```
jzheaux: @jgrandja I looked into simply pulling this from the application context. The issue I discovered is that Spring Boot registers an `AuthenticatedPrincipalOAuth2AuthorizedClientRepository`, which will store this `client` in an `InMemoryOAuth2AuthorizedClientService`. Since that is likely shared between tests, it could cause a problem in concurrent tests.

I've chosen to stick with `HttpSessionOAuth2AuthorizedClientRepository` and documentation that indicates that tests should use that repository so that the argument resolver picks up the client created here. 

With this approach, `@RegisteredOAuth2AuthorizedClient` will work with still pretty minimal code in the test, e.g.:

```java
@Test 
public void test() {
    this.mvc.perform(get("/")
        .with(oidcLogin().clientRegistration(clientRegistration));
}

@Configuration
static class ClientConfig {
    @Bean 
    ClientRegistrationRepository clients() {
        return new InMemoryClientRegistrationRepository(clientRegistration);
    }

    @Bean 
    OAuth2AuthorizedClientRepository authorizedClients() {
        return new HttpSessionOAuth2AuthorizedClientRepository();
    }
}
```

This could be potentially reduced down the road if there's a simple, thread-safe way to reconfigure or override the `OAuth2AuthorizedClientArgumentResolver` within the confines of the test.

/cc @rwinch for additional thoughts.

jgrandja: @jzheaux I like the way the test reads. It's simple, concise and consistent with `HttpSecurity.oauth2Login()` Servlet DSL. For the initial feature release, I think it's fine to require the use `HttpSessionOAuth2AuthorizedClientRepository` in tests. We can figure out ways to improve this later.

jzheaux:
Merged via b35e18ff3104703deee6b8ca6eccce9025a123b2

PR #7623, opened 2019-11-06
Add missing schemas

quaff:
Application will crash if no available internet access since commit https://github.com/spring-projects/spring-security/commit/2e63def05b3f014b09604aa2ba1c866457a97be2#diff-3cade1a1761d3de1f3ae69fe1261dda1
This commit refer to official https://github.com/spring-projects/spring-framework/blob/master/spring-beans/src/main/resources/META-INF/spring.schemas


fabianfrz:
Thanks a lot, I fixed the problem with your help. 

If someone experiences the same behavior, change the xsd:schema-location to look like this:

```
         http://www.springframework.org/schema/security
         https://www.springframework.org/schema/security/spring-security.xsd
```

PR #7639, opened 2019-11-08
Add option to use jwksCache in NimbusJwtDecoder

nmz:
Added option provide jwksCache so that the ttl can be configured

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
@@ -33,6 +33,8 @@ import javax.crypto.SecretKey;
 
 import com.nimbusds.jose.JWSAlgorithm;
 import com.nimbusds.jose.RemoteKeySourceException;
+import com.nimbusds.jose.jwk.source.DefaultJWKSetCache;
+import com.nimbusds.jose.jwk.source.JWKSetCache;
 import com.nimbusds.jose.jwk.source.JWKSource;
 import com.nimbusds.jose.jwk.source.RemoteJWKSet;
 import com.nimbusds.jose.proc.JWSKeySelector;
@@ -212,10 +214,17 @@ public final class NimbusJwtDecoder implements JwtDecoder {
 		private String jwkSetUri;
 		private Set<SignatureAlgorithm> signatureAlgorithms = new HashSet<>();
 		private RestOperations restOperations = new RestTemplate();
+		private JWKSetCache jwkSetCache = new DefaultJWKSetCache();
 
 		private JwkSetUriJwtDecoderBuilder(String jwkSetUri) {
 			Assert.hasText(jwkSetUri, "jwkSetUri cannot be empty");
 			this.jwkSetUri = jwkSetUri;
+		}
+
+		public JwkSetUriJwtDecoderBuilder jwkSetCache(JWKSetCache jwkSetCache) {
+			Assert.notNull(jwkSetCache, "jwkSetCache cannot be null");
+			this.jwkSetCache = jwkSetCache;
+			return this;
 		}
 
 		/**
@@ -279,7 +288,7 @@ public final class NimbusJwtDecoder implements JwtDecoder {
 
 		JWTProcessor<SecurityContext> processor() {
 			ResourceRetriever jwkSetRetriever = new RestOperationsResourceRetriever(this.restOperations);
-			JWKSource<SecurityContext> jwkSource = new RemoteJWKSet<>(toURL(this.jwkSetUri), jwkSetRetriever);
+			JWKSource<SecurityContext> jwkSource = new RemoteJWKSet<>(toURL(this.jwkSetUri), jwkSetRetriever, jwkSetCache);
 			ConfigurableJWTProcessor<SecurityContext> jwtProcessor = new DefaultJWTProcessor<>();
 			jwtProcessor.setJWSKeySelector(jwsKeySelector(jwkSource));
 
--- a/oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
+++ b/oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
@@ -245,6 +245,12 @@ public class NimbusJwtDecoderTests {
 	}
 
 	@Test
+	public void jwkSetCacheWhenNullThenThrowsException() {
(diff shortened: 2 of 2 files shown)


pivotal-issuemaster:
@nmz Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=7639)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=7639) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@nmz Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=7639)!

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -212,12 +214,19 @@ public static SecretKeyJwtDecoderBuilder withSecretKey(SecretKey secretKey) {
...
 		private Set<SignatureAlgorithm> signatureAlgorithms = new HashSet<>();
 		private RestOperations restOperations = new RestTemplate();
+		private JWKSetCache jwkSetCache = new DefaultJWKSetCache();
 
 		private JwkSetUriJwtDecoderBuilder(String jwkSetUri) {
 			Assert.hasText(jwkSetUri, "jwkSetUri cannot be empty");
 			this.jwkSetUri = jwkSetUri;
 		}
 
+		public JwkSetUriJwtDecoderBuilder jwkSetCache(JWKSetCache jwkSetCache) {
```
jzheaux: I'd prefer exposing an interface that is more powerful than the `JWKSetCache` interface. If an application really wants to supply a `JWKSetCache`, then it's quite simple to construct a `DefaultJWTProcessor` directly and pass that to the `NimbusJwtDecoder` constructor.

What if this took a `org.springframework.cache.Cache` instead? Then this builder could internally wrap that in a `CacheJWKSetCache` implementation, similar to the design of `RestOperationsResourceRetreiver`.

The nice thing about doing that is that an application can then use whatever caching mechanism they wish, be it Caffeine, Hazelcast, or something else. 

Another nice this is that using `Cache` allows the entries to be key-value based which is better suited for a multi-tenant resource server that is caching keys from multiple issuers.

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
```diff
@@ -244,6 +244,12 @@ public void jwsAlgorithmWhenNullThenThrowsException() {
 		Assertions.assertThatCode(() -> builder.jwsAlgorithm(null)).isInstanceOf(IllegalArgumentException.class);
 	}
 
+	@Test
```
jzheaux: Please also add a unit test that confirms that the provided cache gets used.

jzheaux:
@nmz are you able to apply the changes requested?

andersflemmen:
@jzheaux if you want the same setup as with the NimbusJwtDecoder, but be able to set the cache time, it's not really just as simple as using the DefaultJwtProcessor though? Why shouldn't it be possible to set the cache time directly? 5 minutes is a pretty short default cache lifespan.

jzheaux:
Good questions, @andersflemmen. Let's see if I can address a couple of them.

> Why shouldn't it be possible to set the cache time directly? 

Spring Security favors exposing components over individual properties. Consider, for example, a related use case of changing the socket timeout for the JWK Set request. Instead of exposing a method for changing the socket timeout, Spring Security exposes a method for supplying a custom `RestOperations`. This keeps the builder from getting cluttered with several small configuration options while at the same time giving the application a great deal of configuration power.

That said, another route could be to introduce a more generic builder - `NimbusJwtDecoder.withJwkSource(JWKSource)` - similar to the one that `NimbusReactiveJwtDecoder` already has. In that case, the application could instantiate its own `JWKSource`:

```java
@Bean 
JwtDecoder jwtDecoder() {
    JWKSource<SecurityContext> jwkSource = new RemoteJWKSet(url, retriever, cache);
    return NimbusJwtDecoder.withJwkSource(jwkSource).build();
}
```

`NimbusJwtDecoder.withJwkSetUri(String)` could be enhanced to delegate to this builder.

>  5 minutes is a pretty short default cache lifespan.

This would be something to take up with the Nimbus folks. Perhaps they'd accept a pull request to lengthen the default ttl.

andersflemmen:
@jzheaux Thank you for clarifying, that makes sense. Would be nice to see either this pull request completed, or an withJwkSource option.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
20fps: @jzheaux Sry, but how should it help to multi-tenant resource server? I mean in multi-tenant environment we will have a separate `AuthenticationManager` for each tenant, that means unique `jwks-uri` for each tenant -> so unique `jwkSetCache` for each tenant, there is no way multiple keys will be stored with the current implementation

jzheaux: Good question, @20fps. I guess it comes down to what parts of your infra are multi-tenant. If your cache is multi-tenant, then each `AuthenticationManager` could have a `JwtDecoder` that shares the same cache - each JWK Set keyed, for example, by that tenant's JWK Set Uri. 

If each tenant has its own cache, then that point is moot. My main point is that by introducing a key to the cache grants additional flexibility - multi-tenancy is one way that flexibility could be leveraged.

20fps: @jzheaux still don't see it can happen since we have the interface method `jwkSetCache#get`, it means we cannot pass any key there, so the only way spring Cache can be used is the hardcoded key only with only one element. Do you have other ideas how to resolve it?

20fps: The possible way could be to allow passing the key from outside, then the method should be like jwkSetCache(Object key, Cache jwkSetCache), so we can have this data inside inner implementation and do something like `get() { return jwkSetCache.get(key); }`, what do you think? Just don't see another way without reimplementing stuff like `RemoteJWKSet`

jzheaux:
Closed in favor of https://github.com/spring-projects/spring-security/pull/8332

Issue #3087, opened 2015-02-23
SEC-2865: Session URL rewriting being disabled should only strip JSESSIONID


spring-projects-issues:
[Rob Winch](https://jira.spring.io/secure/ViewProfile.jspa?name=rwinch) (Migrated from [SEC-2865](https://jira.spring.io/browse/SEC-2865?redirect=false)) said:

This is important so URL rewriting for something like Spring Session can take place

Things to consider:
- Users may have customized the COOKIE name (i.e. it may be a path variable other than JSESSIONID)
- Users may have path variables in their initial URL

OrangeDog:
Just to clarify, it is known that by default `.sessionManagement()` disables ALL rewriting?
While technically correct, rather than "disallows HTTP sessions to be included in the URL", the documentation should be clear on this.

jzheaux:
Good point, @OrangeDog. Would you like to submit a PR with some improved documentation around this?

PR #7644, opened 2019-11-13
Clarify behaviour of enableSessionUrlRewriting

OrangeDog:
See #3087

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/SessionManagementConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/SessionManagementConfigurer.java
@@ -185,8 +185,9 @@ public final class SessionManagementConfigurer<H extends HttpSecurityBuilder<H>>
 	/**
 	 * If set to true, allows HTTP sessions to be rewritten in the URLs when using
 	 * {@link HttpServletResponse#encodeRedirectURL(String)} or
-	 * {@link HttpServletResponse#encodeURL(String)}, otherwise disallows HTTP sessions to
-	 * be included in the URL. This prevents leaking information to external domains.
+	 * {@link HttpServletResponse#encodeURL(String)}, otherwise disallows all URL
+	 * rewriting, including resource chain functionality.
+	 * This prevents leaking information to external domains.
 	 *
 	 * @param enableSessionUrlRewriting true if should allow the JSESSIONID to be
 	 * rewritten into the URLs, else false (default)


config/src/main/java/org/springframework/security/config/annotation/web/configurers/SessionManagementConfigurer.java
```diff
@@ -185,8 +185,9 @@ public SessionManagementConfigurer() {
 	/**
 	 * If set to true, allows HTTP sessions to be rewritten in the URLs when using
 	 * {@link HttpServletResponse#encodeRedirectURL(String)} or
-	 * {@link HttpServletResponse#encodeURL(String)}, otherwise disallows HTTP sessions to
-	 * be included in the URL. This prevents leaking information to external domains.
+	 * {@link HttpServletResponse#encodeURL(String)}, otherwise disallows all URL
+	 * rewriting, including resource chain functionality.
```
jzheaux: Two questions here:

- Why is it necessary to list "resource chain functionality" directly?

- Since the first part of this JavaDoc says "if set to true, allows HTTP sessions to be ....", then it makes sense to reiterate this in the converse, so I'd suggest:

```suggestion
	 * rewriting, including disallowing HTTP sessions from being included in the URL.
```

OrangeDog: Well this is why I suggested the developers clarify it...

The surprising behaviour is that this also controls non-Spring Security features. From experience, the resource chain (seemingly) randomly not working a Spring project is the most common issue.

You could just make "HTTP sessions" an aside remark for the whole thing, but then it's more obvious that the method is badly named or there's a bug.

jzheaux: I think that this method aligns with the servlet JavaDoc, which says for `HttpServletResponse#encodeURL`:

> Encodes the specified URL by including the session ID,
or, if encoding is not needed, returns the URL unchanged.
The implementation of this method includes the logic to
determine whether the session ID needs to be encoded in the URL.
For example, if the browser supports cookies, or session
tracking is turned off, URL encoding is unnecessary.

The way I read this is that implementers of `HttpServletResponse` should add the `jsessionid`, encoding it as necessary, but if no `jsessionid` is to be added, leave the URL unchanged. I agree that `HttpServletResponse#encodeURL` is a bit of an odd name, given the JavaDoc.

[The servlet spec](https://javaee.github.io/servlet-spec/downloads/servlet-3.1/Final/servlet-3_1-final.pdf), which refers in 7.1.3 to the function of adding `jsessionid` as a request parameter as "URL Rewriting".

Because of these, the method `enableSessionUrlRewriting` seems like a natural name, evoking that it is to make sure `jsessionid` is not managed as a request parameter.

Maybe what I'm missing is a more detailed description of the problem you are seeing when setting this to `true`. What are you expecting the Servlet container to do with the URL that gets skipped because `HttpServletResponse#encodeURL` is bypassed?

OrangeDog: The problem is when it's `false`: #3087

Other things (e.g. Spring's resource chain) use `encodeURL` to do other stuff.

jzheaux:
@OrangeDog, I created a sample application that demonstrates [resource chaining working by default in conjunction with Spring Security](https://github.com/jzheaux/spring-security-examples/tree/master/gh-7644).

Given that, I don't believe it's accurate to say that `enableSessionUrlRewriting` disables resource chaining.

What would you think about something like:

```java
 * <p>
 * This is achieved by guarding {@link HttpServletResponse#encodeURL} and {@link HttpServletResponse#encodeRedirectURL} invocations.
 * Any code that also overrides either of these two methods,
 * like {@link org.springframework.web.servlet.resource.ResourceUrlEncodingFilter},
 * needs to come after the security filter chain or risk being skipped.
```

Since JavaDoc is quite simple to adjust, if I don't hear anything else from you in a week or so, I'll go ahead and change it to read like the above. Either way, don't hesitate to reach out if you find additional refinements.

Issue #7661, opened 2019-11-20
Check BCrypt hashed value of a byte array 


rafahop:
### Summary
Since 5.2.0.RELEASE (388a7b62b906bd56deadb7ca45248fa1a63bdf12) BCrypt offers an overloaded [hashpw](https://github.com/spring-projects/spring-security/blob/0d35194b47d1ed0e2558e106aa8658a6aeea17a5/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCrypt.java#L747) method to hash a byte array. However there is no [checkpw](https://github.com/spring-projects/spring-security/blob/0d35194b47d1ed0e2558e106aa8658a6aeea17a5/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCrypt.java#L907) equivalent to check wether a byte array matches a previously hashed string.

If hashing a byte[], we are forced to wrap it into a String when calling the existing `checkpw` and this does not necessarily returns the correct result:

```
byte[] pw = new byte[] { -5 };
assertThat(BCrypt.checkpw(new String(pw), BCrypt.hashpw(pw, BCrypt.gensalt()))).isTrue();
```

PR #7662, opened 2019-11-20
Check BCrypt hashes of byte array passwords

rafahop:
BCrypt: add missing overloaded version of `checkpw` accepting a byte array.

Fixes gh-7661


crypto/src/test/java/org/springframework/security/crypto/bcrypt/BCryptTests.java
```diff
@@ -14,18 +14,33 @@
 
 package org.springframework.security.crypto.bcrypt;
 
-import org.junit.Test;
```
eleftherias: Please do not change the import ordering unless it's necessary. 
It can make the revision history unclear as to when an import was added.

crypto/src/test/java/org/springframework/security/crypto/bcrypt/BCryptTests.java
```diff
@@ -14,18 +14,33 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
 
 /**
  * JUnit unit tests for BCrypt routines
  * @author Damien Miller
  */
 public class BCryptTests {
 
+	private static class TestObject<T> {
```
eleftherias: This class is a great addition to the test, it helps with readability and clarifies what the tests are doing.
Since it's not directly related to the `checkpw` addition, could you please move it to a separate PR or separate commit?

rafahop: Thanks for the feedback @eleftherias ! 
It does make sense splitting it into a separate PR and I have done so here: #7694 

Please note that to keep it as simplest as possible, this `TestObject` won't be using generics in that PR as the generic type would only be required when adding the ability to test a byte[] - which will be added as part of this change though!

I'll sort this branch to contain only the changes for `checkpw` once #7694 is approved and merged in 👍

crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCrypt.java
```diff
@@ -908,6 +908,17 @@ public static boolean checkpw(String plaintext, String hashed) {
 		return equalsNoEarlyReturn(hashed, hashpw(plaintext, hashed));
 	}
 
+	/**
+	 * Check that a password (as a byte array) matches a previously hashed
+	 * one
+	 * @param passwordb	the password to verify, as a byte array
+	 * @param hashed		the previously-hashed password
+	 * @return	true if the passwords match, false otherwise
+	 */
```
eleftherias: Please add the `@since` version 5.3 tag.

crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCrypt.java
```diff
@@ -908,6 +908,17 @@ public static boolean checkpw(String plaintext, String hashed) {
 		return equalsNoEarlyReturn(hashed, hashpw(plaintext, hashed));
 	}
 
+	/**
+	 * Check that a password (as a byte array) matches a previously hashed
+	 * one
+	 * @param passwordb	the password to verify, as a byte array
+	 * @param hashed		the previously-hashed password
```
eleftherias: Something is off with the formatting in this line, perhaps an extra tab?

rafahop:
All done @eleftherias ! 👍

eleftherias:
Thanks for the PR @rafahop! This is now merged into master.

PR #7666, opened 2019-11-21
Fix InitializeAuthenticationProviderBeanManagerConfigurer Javadoc

xak2000:
Currently the javadoc of class `InitializeAuthenticationProviderBeanManagerConfigurer` is copy-pasted from `InitializeUserDetailsBeanManagerConfigurer`. This PR just fixes the Javadoc and removes an unnecessary import.

This also fixes copy-pasted class and variable names in `InitializeAuthenticationProviderBeanManagerConfigurer` class and also fixes variable name in `InitializeUserDetailsBeanManagerConfigurer`: `userDetailsBeanNames` -> `beanNames` as the names returned for bean type that depends on method parameter and they are not always `UserDetailsService` type (the method `getBeanOrNull` is called several times with different 'type' argument actually).


eleftherias:
Thanks for the PR @xak2000! This is now merged into master.

PR #7673, opened 2019-11-26
fix typo: require require

ryenus:
Title says it all.


eleftherias:
Thanks for the PR @ryenus! This is now merged into master.

PR #7675, opened 2019-11-26
Correctly configure authorization requests repository for OAuth2 login

alek-sys:
To use custom `ServerAuthorizationRequestRepository` both `OAuth2AuthorizationRequestRedirectWebFilter` and `OAuth2LoginAuthenticationWebFilter` should use the same repo provided in the configuration. Currently the former filter is correctly configured, but the latter always uses default, WebSession based repository. So authorization code created before redirect to authorization endpoint will never be found to complete OAuth2 login when custom `ServerAuthorizationRequestRepository` is used.

This change also makes OAuth2Client and OAuth2Login authentication converters consistent.


jgrandja:
Thanks for the PR @alek-sys! This is now in master and back ported to 5.2.x. Thank you for reporting this.

PR #7678, opened 2019-11-27
Log full failed authentication exception in BasicAuthenticationFilter

filiphr:
Was having some problems when my implementations were throwing some runtime exceptions. Which lead to `InternalAuthenticationServiceException` being thrown. However, the log didn't have any stacktrace. With this change it would be possible to see the stacktrace when such exception occurs.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilter.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilter.java
@@ -183,7 +183,7 @@ public class BasicAuthenticationFilter extends OncePerRequestFilter {
 			SecurityContextHolder.clearContext();
 
 			if (debug) {
-				this.logger.debug("Authentication request for failed: " + failed, failed);
+				this.logger.debug("Authentication request for failed!", failed);
 			}
 
 			this.rememberMeServices.loginFail(request, response);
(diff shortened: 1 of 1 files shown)


web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilter.java
```diff
@@ -183,7 +183,7 @@ protected void doFilterInternal(HttpServletRequest request,
 			SecurityContextHolder.clearContext();
 
 			if (debug) {
-				this.logger.debug("Authentication request for failed: " + failed);
+				this.logger.debug("Authentication request for failed: " + failed, failed);
```
eleftherias: Please change the debug message to be the same as `BearerTokenAuthenticationFilter` https://github.com/spring-projects/spring-security/blob/56f524259588e225368327a4ee19eaaf2416ba99/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/BearerTokenAuthenticationFilter.java#L135

filiphr:
Thanks for the Review @eleftherias, applied the suggestion. Please squash all commits when merging, I couldn't do it via the GitHub UI

eleftherias:
Thanks for the PR @filiphr! This is now merged into master.

filiphr:
Thanks for merging, thanks for linking it with the issue. I didn't really search if there is one, since it was so simple :)

PR #7687, opened 2019-11-29
switches web authentication principal resolver to use reactive context (#6598)

Dav1dde:
Implementation for #6598 as noted by @eleftherias 

Replacing `exchange.getPrincipal()` with `ReactiveSecurityContextHolder.getContext().map(SecurityContext::getAuthentication)` in `AuthenticationPrincipalArgumentResolver`.
Other implementations of `AuthenticationPrincipalArgumentResolver` are already doing that (messaging).

One thing I noticed was, the logic for `AuthenticationPrincipalArgumentResolver` exists multiple times (like 2-3 duplicates), maybe it's time to refactor the logic out into an abstract base class.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/reactive/result/method/annotation/AuthenticationPrincipalArgumentResolver.java
+++ b/web/src/main/java/org/springframework/security/web/reactive/result/method/annotation/AuthenticationPrincipalArgumentResolver.java
@@ -28,8 +28,9 @@ import org.springframework.expression.Expression;
 import org.springframework.expression.ExpressionParser;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
 import org.springframework.expression.spel.support.StandardEvaluationContext;
-import org.springframework.security.core.Authentication;
 import org.springframework.security.core.annotation.AuthenticationPrincipal;
+import org.springframework.security.core.context.ReactiveSecurityContextHolder;
+import org.springframework.security.core.context.SecurityContext;
 import org.springframework.util.StringUtils;
 import org.springframework.web.reactive.BindingContext;
 import org.springframework.web.reactive.result.method.HandlerMethodArgumentResolverSupport;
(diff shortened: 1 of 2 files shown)


eleftherias:
Thanks for the PR @Dav1dde! This is now merged into master.

web/src/main/java/org/springframework/security/web/reactive/result/method/annotation/AuthenticationPrincipalArgumentResolver.java
```diff
@@ -69,9 +70,9 @@ public boolean supportsParameter(MethodParameter parameter) {
 	public Mono<Object> resolveArgument(MethodParameter parameter, BindingContext bindingContext,
 			ServerWebExchange exchange) {
 		ReactiveAdapter adapter = getAdapterRegistry().getAdapter(parameter.getParameterType());
-		return exchange.getPrincipal()
-			.ofType(Authentication.class)
-			.flatMap( a -> {
+		return ReactiveSecurityContextHolder.getContext()
+			.map(SecurityContext::getAuthentication)
```
robotmrv: @Dav1dde  @eleftherias 
By javadoc `SecurityContext::getAuthentication` could return `null`  
`Mono.map` forbids `null`  
should this case be handled, or it is very unlikely to happen

Dav1dde: Looking at the implementation of `ReactiveSecurityContextHolder` this may very well be the case (the API allows it).

This would also need to be fixed in multiple other `AuthenticationPrinicipalArgumentResolver` implementations, e.g. in [messaging](https://github.com/spring-projects/spring-security/blob/master/messaging/src/main/java/org/springframework/security/messaging/handler/invocation/reactive/AuthenticationPrincipalArgumentResolver.java#L129)

Issue #7654, opened 2019-11-18
Allow configuring authenticationManagerResolver for SAML2


eleftherias:
### Summary

It should be easier to configure the `authoritiesExtractor`, `authoritiesMapper` and `responseTimeValidationSkew` on the `OpenSamlAuthenticationProvider`. 
See gh-7642.

Allowing an `authenticationManagerResolver` for SAML2 login will allow customizing the `OpenSamlAuthenticationProvider`.

This should be similar to [OAuth2ResourceServerConfigurer.authenticationManagerResolver](https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.html#authenticationManagerResolver-org.springframework.security.authentication.AuthenticationManagerResolver-).

fhanik:
@eleftherias I've been reviewing this, and implemented a possible solution in [1d71a62](https://github.com/fhanik/spring-security/commit/1d71a62c44b3b25f170ea9e391ccc016811be573).

It does however become difficult to justify the use of a `AuthenticationManagerResolver` simply for the ability to configure setters on the authentication provider. 

The `AbstractAuthenticationFilterConfigurer` calls [`setAuthenticationManager`](https://github.com/spring-projects/spring-security/blob/f5f965b6aac76596e1745056d4b015b60a7cbd85/config/src/main/java/org/springframework/security/config/annotation/web/configurers/AbstractAuthenticationFilterConfigurer.java#L291-L292) making the end result of the configuration less obvious.

I will continue reviewing this, but may opt to just make the authentication provider configurable, or the options on it, rather than adding a resolver.

eleftherias:
@fhanik The idea with having the `AuthenticationManagerResolver` configurable is that it would support multi-tenancy in the future.
For now, we only need the ability to set the options on the provider.
I will leave it up to you to decide which option is best.

PR #7693, opened 2019-12-02
Enable AuthenticationManager configuration in saml2Login

fhanik:
Allow configuration of AuthenticationManagerResolver in saml2Login()
    
Fixes gh-7654


config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
```diff
@@ -103,6 +105,14 @@
 
 	private RelyingPartyRegistrationRepository relyingPartyRegistrationRepository;
 
+	private AuthenticationManagerResolver<HttpServletRequest> authenticationManagerResolver;
+
+	public Saml2LoginConfigurer authenticationManagerResolver
```
jzheaux: Reminder to add JavaDoc, including a `@since` tag.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/servlet/filter/Saml2WebSsoAuthenticationFilter.java
```diff
@@ -50,13 +53,21 @@ public Saml2WebSsoAuthenticationFilter(RelyingPartyRegistrationRepository relyin
 	public Saml2WebSsoAuthenticationFilter(
 			RelyingPartyRegistrationRepository relyingPartyRegistrationRepository,
 			String filterProcessesUrl) {
+		this(relyingPartyRegistrationRepository, filterProcessesUrl, null);
+	}
+
+	public Saml2WebSsoAuthenticationFilter(
```
jzheaux: Reminder to please add JavaDoc, including a `@since` tag.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/servlet/filter/Saml2WebSsoAuthenticationFilter.java
```diff
@@ -86,7 +97,16 @@ public Authentication attemptAuthentication(HttpServletRequest request, HttpServ
 				localSpEntityId,
 				rp.getCredentials()
 		);
-		return getAuthenticationManager().authenticate(authentication);
+		return getAuthenticationManager(request).authenticate(authentication);
```
jzheaux: It makes me a little nervous that an instance of this filter could contain both an `AuthenticationManager` and an `AuthenticationManagerResolver` instance. It could be a source of bugs in the future.

Have you already considered pushing the `AuthenticationManagerResolver` instead up to the parent? You could possibly have the parent hold onto only an instance of `AuthenticationManagerResolver` and have the `getAuthenticationManager(HttpServletRequest) method`.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurerTests.java
```diff
@@ -0,0 +1,290 @@
...
+				);
+			}
+
+			public boolean supports(Class<?> authentication) {
+				return authentication.isAssignableFrom(Saml2AuthenticationToken.class);
+			}
+		};
+	}
+
+	private static RelyingPartyRegistration getSaml2AuthenticationConfiguration() {
```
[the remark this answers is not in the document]
fhanik: Good point. Support data has been extracted.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/servlet/filter/Saml2WebSsoAuthenticationFilter.java
```diff
@@ -50,13 +53,21 @@ public Saml2WebSsoAuthenticationFilter(RelyingPartyRegistrationRepository relyin
...
 			RelyingPartyRegistrationRepository relyingPartyRegistrationRepository,
 			String filterProcessesUrl) {
+		this(relyingPartyRegistrationRepository, filterProcessesUrl, null);
+	}
+
+	public Saml2WebSsoAuthenticationFilter(
+				RelyingPartyRegistrationRepository relyingPartyRegistrationRepository,
+				String filterProcessesUrl,
+				AuthenticationManagerResolver<HttpServletRequest> authenticationManagerResolver) {
 		super(filterProcessesUrl);
```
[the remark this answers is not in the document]
fhanik: I've removed the `AuthenticationManagerResolver`. It creates confusion around the original goal of being able to create and configure an authentication provider for the SAML chain.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
fhanik: Thanks. Done. and this method changed from `authenticationManagerResolver` to `authenticationProvider`

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
```diff
@@ -103,6 +103,20 @@
...
+	private AuthenticationProvider authenticationProvider;
+
+	/**
+	 * Allows a configuration of a {@link AuthenticationProvider} to be used during SAML 2 authentication.
+	 * If none is specified, the system will create one and insert it into the ProviderManager list
+	 * @param authenticationProvider the authentication provider to be used
+	 * @return the {@link Saml2LoginConfigurer} for further configuration
+	 * @since 5.2.2
+	 */
+	public Saml2LoginConfigurer authenticationProvider(AuthenticationProvider authenticationProvider) {
```
jgrandja: Instead of exposing an `authenticationProvider(..)` configuration option, we may want to consider providing `authoritiesExtractor(...)` and `authoritiesMapper(...)` under `saml2Login()` OR under a _nested config class_. This would be consistent with `oauth2Login()`, `oauth2Client()` and `oauth2ResourceServer()` as neither expose `authenticationProvider()` and instead handle the configuration of the provider in `configure(http)`.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurerTests.java
```diff
@@ -0,0 +1,185 @@
...
+import java.util.Collections;
+import javax.servlet.ServletException;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static org.springframework.security.config.annotation.web.configurers.saml2.TestRelyingPartyRegistrations.getSaml2AuthenticationConfiguration;
+
+public class Saml2LoginConfigurerTests {
```
jgrandja: Please add class-level javadoc.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/saml2/TestRelyingPartyRegistrations.java
```diff
@@ -0,0 +1,53 @@
...
+package org.springframework.security.config.annotation.web.configurers.saml2;
+
+import org.springframework.security.saml2.credentials.Saml2X509Credential;
+import org.springframework.security.saml2.provider.service.registration.RelyingPartyRegistration;
+import org.springframework.security.saml2.provider.service.servlet.filter.Saml2WebSsoAuthenticationFilter;
+
+import static org.springframework.security.config.annotation.web.configurers.saml2.TestSaml2Credentials.getSigningCredential;
+import static org.springframework.security.config.annotation.web.configurers.saml2.TestSaml2Credentials.getVerificationCertificate;
+
+public class TestRelyingPartyRegistrations {
```
jgrandja: Please add class-level javadoc.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/saml2/TestSaml2Credentials.java
```diff
@@ -0,0 +1,113 @@
...
+import java.nio.charset.StandardCharsets;
+import java.security.PrivateKey;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+
+import static org.springframework.security.saml2.credentials.Saml2X509Credential.Saml2X509CredentialType.DECRYPTION;
+import static org.springframework.security.saml2.credentials.Saml2X509Credential.Saml2X509CredentialType.SIGNING;
+import static org.springframework.security.saml2.credentials.Saml2X509Credential.Saml2X509CredentialType.VERIFICATION;
+
+public class TestSaml2Credentials {
```
jgrandja: Please add class-level javadoc.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
```diff
@@ -211,11 +226,16 @@ public void init(B http) throws Exception {
...
+		}
+		else {
+			saml2WebSsoAuthenticationFilter.setAuthenticationManager(this.authenticationManager);
+		}
 	}
 
-	private AuthenticationProvider getAuthenticationProvider() {
-		AuthenticationProvider provider = new OpenSamlAuthenticationProvider();
-		return postProcess(provider);
+	private AuthenticationManager getAuthenticationManager() {
```
jzheaux: I think this method should probably register the provider using `http.authenticationProvider()` and then return the shared `AuthenticationManager`. The default behavior for these configurers is to register a provider with the shared provider manager. If the code calls `saml2Login().authenticationManager()`, it's clear that they want to replace this default behavior for this filter.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/servlet/filter/Saml2WebSsoAuthenticationFilter.java
fhanik: Completed

config/src/test/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurerTests.java
fhanik: Completed.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/saml2/TestRelyingPartyRegistrations.java
fhanik: Completed.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/saml2/TestSaml2Credentials.java
fhanik: Completed.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
fhanik: Simplified it to a return void and renamed it to 
[`registerDefaultAuthenticationProvider`](https://github.com/fhanik/spring-security/commit/146502a967b5c0d2c95796ba8a5300c52dac65ff)

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
fhanik: Thanks @jgrandja . This was my first consideration. That leaves the implementation with an inability to affect or easily swap out SAML response parsing and validation, something that can definitely change between providers. The idea here was to follow Spring logic that one should be able to swap out each component, not just the values of a component.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
```diff
@@ -103,6 +111,71 @@
...
+
+	/**
+	 * Allows a configuration of a {@link AuthenticationManager} to be used during SAML 2 authentication.
+	 * If none is specified, the system will create one inject it into the {@link Saml2WebSsoAuthenticationFilter}
+	 * @param authenticationManager the authentication manager to be used
+	 * @return the {@link Saml2LoginConfigurer} for further configuration
+	 * @throws IllegalArgumentException if authenticationManager is null
+	 * @throws IllegalArgumentException if {@link #authenticationConfiguration} has already been invoked to
+	 * 									configure the default manager
+	 * @since 5.2.2
```
jzheaux: Because these are changes to the public API and are not addressing a bug, we should [target a minor release instead](https://semver.org/#spec-item-7).

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
```diff
@@ -313,5 +403,50 @@ private Saml2AuthenticationRequestFactory getResolver(B http) {
 		}
 	}
 
+	/**
+	 * Configuration held for the default authentication manager used in SAML authentication.
+	 */
+	public class AuthenticationConfigurationSpec<B extends HttpSecurityBuilder<B>> {
```
jzheaux: I'd recommend leaving this out for the time being. The main reason is that applications can either call  `authenticationManager()` or they can use a post-processor.

A second reason is that the DSL is not typically a place where all the properties of a given component are exposed.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProvider.java
```diff
@@ -122,7 +123,7 @@
  * @see <a href="https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf#page=38">SAML 2 StatusResponse</a>
  * @see <a href="https://wiki.shibboleth.net/confluence/display/OS30/Home">OpenSAML 3</a>
  */
-public final class OpenSamlAuthenticationProvider implements AuthenticationProvider {
+public final class OpenSamlAuthenticationProvider implements AuthenticationManager, AuthenticationProvider {
```
jzheaux: This might be a breaking change. For example, if users have exposed `OpenSamlAuthenticationProvider` as a bean in 5.2.x, e.g.:

```java
@Bean
public OpenSamlAuthenticationProvider provider() {
    // ... 
}
```

Then startup may fail when upgrading to 5.3.x as there will be more than one instance of `AuthenticationManager` in the application context.

You can see related behavior by changing `public interface AuthenticationProvider` to `public interface AuthenticationProvider extends AuthenticationManager` and running `MethodSecurityInterceptorWithAopConfigTests`. It's not the same thing, but it may give you some places to poke around in order to get a better idea of the implications of the change.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/saml2/TestSaml2Credentials.java
```diff
@@ -0,0 +1,116 @@
...
+import static org.springframework.security.saml2.credentials.Saml2X509Credential.Saml2X509CredentialType.DECRYPTION;
+import static org.springframework.security.saml2.credentials.Saml2X509Credential.Saml2X509CredentialType.SIGNING;
+import static org.springframework.security.saml2.credentials.Saml2X509Credential.Saml2X509CredentialType.VERIFICATION;
+
+/**
+ * Preconfigured SAML credentials for SAML integration tests.
+ */
+public class TestSaml2Credentials {
+
+	static Saml2X509Credential getVerificationCertificate() {
```
jzheaux: A minor point, but most other `TestXXXs` classes don't prefix the methods with `get`. I think it would be more consistent to call this method `verificationCertificate`, for example.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
fhanik: I've also added in a _spec_ for configuring the defaults settings.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
fhanik: Do we hold back additive changes? Aren't those simply backwards compatible?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
fhanik: The [`Oauth2LoginConfigurer`](https://github.com/spring-projects/spring-security/blob/65f6025cef07113c4a165dcc9ad6c50cd1c66475/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2LoginConfigurer.java#L512-L515) chooses to use beans or shared objects to configure the same.
Would that be preferred if one doesn't want to override the default implementation

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProvider.java
fhanik: That's a good point. I will revert that change.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/saml2/TestSaml2Credentials.java
fhanik: Resolved

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProvider.java
fhanik: That's a good point. I will remove the redundant interface

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
fhanik: Should I expose them as `shared objects` (like `Oauth2LoginConfigurer` does) or do we simply close the original issue as `won't fix` since the postprocessor does allow you to set it?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
fhanik: Even for additive APIs? Rather than simply pushing out the code and let it be used?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
fhanik: This was removed. A post processor approach has been documented in the SAML documentation

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
fhanik: I misunderstood. Updated to 5.3

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
fhanik: I've opted to simply document the post processor approach in the documentation.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurerTests.java
```diff
@@ -0,0 +1,259 @@
...
+				.getAuthentication();
+		Assert.assertNotNull("Expected a valid authentication object.", authentication);
+		assertThat(authentication.getAuthorities()).hasSize(1);
+		assertThat(authentication.getAuthorities()).first()
+				.isInstanceOf(SimpleGrantedAuthority.class).hasToString(expected);
+	}
+
+
+	@EnableWebSecurity
+	static class Saml2LoginConfigWithCustomAuthenticationManager extends Saml2LoginConfigBase {
```
jzheaux: Would

```suggestion
	@Import(Saml2LoginConfigBase.class)
	static class Saml2LoginConfigWithCustomAuthenticationManager extends WebSecurityConfigurerAdapter {
```

be a bit simpler? I guess it feels a little odd to extend `WebSecurityConfigurerAdapter` when only adding beans into the application context, it seems like it's doing two jobs.

docs/manual/src/docs/asciidoc/_includes/servlet/saml2/saml2-login.adoc
```diff
@@ -223,6 +223,66 @@ public interface Saml2AuthenticationRequestFactory {
...
+
+        http.saml2Login()
+                .addObjectPostProcessor(processor)
+        ;
+        super.configure(http);
+    }
+}
+----
+
+The authentication manager for the security filter can also be overwritten, using your own
```
jzheaux: This sentence seems to imply by omission that the user's only alternative is creating a custom `AuthenticationManager` instance.

It seems to me that the following is perfectly acceptable and arguably as easy as the first option:

```java
OpenSamlAuthenticationProvider authenticationProvider = new OpenSamlAuthenticationProvider(...);
authenticationProvider.setResponseTimeValidationSkew(...)
...
AuthenticationManager authenticationManager = new ProviderManager
    (Arrays.asList(authenticationProvider));
http
    .saml2Login()
        .authenticationManager(authenticationManager);
```

docs/manual/src/docs/asciidoc/_includes/servlet/saml2/saml2-login.adoc
```diff
@@ -223,6 +223,66 @@ public interface Saml2AuthenticationRequestFactory {
...
+            @Override
+            public <O extends OpenSamlAuthenticationProvider> O postProcess(O provider) {
+                provider.setResponseTimeValidationSkew(RESPONSE_TIME_VALIDATION_SKEW);
+                provider.setAuthoritiesMapper(AUTHORITIES_MAPPER);
+                provider.setAuthoritiesExtractor(AUTHORITIES_EXTRACTOR);
+                return provider;
+            }
+        };
+
+        http.saml2Login()
```
jzheaux: For consistency could you do:

```suggestion
        http
            .saml2Login()
```

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
```diff
@@ -103,6 +103,25 @@
...
+	private Saml2WebSsoAuthenticationFilter saml2WebSsoAuthenticationFilter;
+
+	/**
+	 * Allows a configuration of a {@link AuthenticationManager} to be used during SAML 2 authentication.
+	 * If none is specified, the system will create one inject it into the {@link Saml2WebSsoAuthenticationFilter}
+	 * @param authenticationManager the authentication manager to be used
+	 * @return the {@link Saml2LoginConfigurer} for further configuration
+	 * @throws IllegalArgumentException if authenticationManager is null
+	 * 									configure the default manager
+	 * @since 5.3.0
```
jzheaux: Please do

```suggestion
	 * @since 5.3
```

instead.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurerTests.java
fhanik: Agreed

docs/manual/src/docs/asciidoc/_includes/servlet/saml2/saml2-login.adoc
fhanik: It's almost the same. The default filter chain will have a `ProviderManager` with more than one authentication manager. This replaces the provider manager with a singleton list.

I like this suggestion and will add it to the documentation

docs/manual/src/docs/asciidoc/_includes/servlet/saml2/saml2-login.adoc
fhanik: Fixed

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
fhanik: Fixed

PR #7726, opened 2019-12-12
Fix remember-me cookie set/cancel inconsistency:

okohub:
AbstractRememberMeServices is setting remember-me cookie with checking request is secure or secure usage is independently set to a fixed flag.
But when cancelling a cookie, cookie is not being marked secure or not. It produces an inconsistency when using secure flag as a part to identity of cookie. This commit intended to fix this situation.


okohub:
@eleftherias added tests, thank you :)

eleftherias:
Thanks for the PR @okohub! This is now merged into master.

Issue #7724, opened 2019-12-11
Add JwtAuthenticationManagerResolver


jzheaux:
Related to #6778 and #5351 

It would be nice if there were an `AuthenticationManagerResolver` implementation that simplified the resolution of `AuthenticationManager`s by the `iss` JWT claim.

Issue #5351, opened 2018-05-15
Multi-tenancy support for OAuth2


jzheaux:
### Summary

Today, it isn't clear how to best configure Spring Security to support a multi-tenant OAuth2 client.

Here is an example of one approach out in the wild:

https://github.com/thomasdarimont/spring-boot-2-keycloak-oauth-example/blob/feature/mulit-tenancy/src/main/java/demo/SpringBoot2App.java#L127

Though whether `JwtDecoder` is the ideal place and how tenants might possibly be treated in a more first-class way is yet to be seen. Opening this issue to get the conversation started.

jgrandja:
Related #5385

metacubed:
In a multi-tenant configuration, each tenant would likely come with its own set of allowed issuers. A tenant-specific request would trust only a subset of the issuers configured for the resource server.

We would need to introduce a selection mechanism to pick the allowed issuer(s) for a request. This `IssuerListSelector` could have custom implementations that are specific to the resource server.

For example, the `keycloak-auth` project uses the `request.getServerName()` to decide which decoder implementation to use. Other request attributes could also be used, such as the request sub-domain, a header, etc.

canayozel:
@jzheaux since this is released with spring security 5.2.0 is there any documentation or example to follow ?

jzheaux:
@canayozel Multi-tenancy support for clients is had through multiple client registrations: https://docs.spring.io/spring-security/site/docs/current/reference/htmlsingle/#oauth2login-custom-provider-properties

For resource server, we have [a sample](https://github.com/spring-projects/spring-security/tree/master/samples/boot/oauth2resourceserver-multitenancy), but are [still on the hook for adding documentation](https://github.com/spring-projects/spring-security/issues/7532).

bertramn:
Hmm not too sure how the [multi tenant example](https://github.com/spring-projects/spring-security/blob/master/samples/boot/oauth2resourceserver-multitenancy/src/main/java/sample/OAuth2ResourceServerSecurityConfiguration.java) is of use to anyone wanting to run a Resource Server that can validate JWTs from different issuers. It just flip-flops the `AuthenticationManager` back and forth between paths. And I thought we finally could replace our custom implementation.

jzheaux:
[...]

PR #7733, opened 2019-12-12
Add JwtIssuerAuthenticationManagerResolver

jzheaux:
Fixes gh-7724

### DIFF
--- a/config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
+++ b/config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
@@ -28,10 +28,19 @@ import java.time.Instant;
 import java.time.ZoneId;
 import java.util.Base64;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.stream.Collectors;
 import javax.annotation.PreDestroy;
 
+import com.nimbusds.jose.JWSAlgorithm;
+import com.nimbusds.jose.JWSHeader;
+import com.nimbusds.jose.JWSObject;
+import com.nimbusds.jose.Payload;
+import com.nimbusds.jose.crypto.RSASSASigner;
+import com.nimbusds.jose.jwk.JWKSet;
+import com.nimbusds.jose.jwk.RSAKey;
+import net.minidev.json.JSONObject;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
 import org.hamcrest.core.AllOf;
@@ -82,14 +91,15 @@ import org.springframework.security.oauth2.core.DefaultOAuth2AuthenticatedPrinci
 import org.springframework.security.oauth2.core.OAuth2Error;
 import org.springframework.security.oauth2.core.OAuth2TokenValidator;
 import org.springframework.security.oauth2.core.OAuth2TokenValidatorResult;
+import org.springframework.security.oauth2.jose.TestKeys;
 import org.springframework.security.oauth2.jwt.Jwt;
-import org.springframework.security.oauth2.jwt.JwtClaimNames;
 import org.springframework.security.oauth2.jwt.JwtDecoder;
 import org.springframework.security.oauth2.jwt.JwtException;
 import org.springframework.security.oauth2.jwt.JwtTimestampValidator;
 import org.springframework.security.oauth2.jwt.NimbusJwtDecoder;
 import org.springframework.security.oauth2.server.resource.authentication.BearerTokenAuthentication;
 import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationConverter;
+import org.springframework.security.oauth2.server.resource.authentication.JwtIssuerAuthenticationManagerResolver;
 import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationToken;
 import org.springframework.security.oauth2.server.resource.introspection.NimbusOpaqueTokenIntrospector;
 import org.springframework.security.oauth2.server.resource.introspection.OpaqueTokenIntrospector;
@@ -127,6 +137,8 @@ import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 import static org.springframework.security.config.Customizer.withDefaults;
 import static org.springframework.security.oauth2.core.TestOAuth2AccessTokens.noScopes;
+import static org.springframework.security.oauth2.jwt.JwtClaimNames.ISS;
+import static org.springframework.security.oauth2.jwt.JwtClaimNames.SUB;
 import static org.springframework.security.oauth2.jwt.NimbusJwtDecoder.withJwkSetUri;
 import static org.springframework.security.oauth2.jwt.NimbusJwtDecoder.withPublicKey;
 import static org.springframework.security.oauth2.jwt.TestJwts.jwt;
@@ -149,7 +161,7 @@ import static org.springframework.web.bind.annotation.RequestMethod.POST;
 public class OAuth2ResourceServerConfigurerTests {
 	private static final String JWT_TOKEN = "token";
 	private static final String JWT_SUBJECT = "mock-test-subject";
-	private static final Map<String, Object> JWT_CLAIMS = Collections.singletonMap(JwtClaimNames.SUB, JWT_SUBJECT);
+	private static final Map<String, Object> JWT_CLAIMS = Collections.singletonMap(SUB, JWT_SUBJECT);
 	private static final Jwt JWT = jwt().build();
 	private static final String JWK_SET_URI = "https://mock.org";
 	private static final JwtAuthenticationToken JWT_AUTHENTICATION_TOKEN =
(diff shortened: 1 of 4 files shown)


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationManagerResolver.java
```diff
@@ -0,0 +1,159 @@
...
+ * anyone can stand up an authorization server and issue valid tokens to a resource server. The simplest way
+ * to achieve this is to supply a whitelist of trusted issuers in the constructor.
+ *
+ * This class derives the Issuer from the `iss` claim found in the {@link HttpServletRequest}'s
+ * <a href="https://tools.ietf.org/html/rfc6750#section-1.2" target="_blank">Bearer Token</a>.
+ *
+ * @author Josh Cummings
+ * @since 5.3
+ */
+public final class JwtAuthenticationManagerResolver implements AuthenticationManagerResolver<HttpServletRequest> {
```
fhanik: Is it intended that there is no documentation as part of this new feature, sample perhaps?
Is that a separate issue?

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
```diff
@@ -1332,6 +1344,49 @@ public void getAuthenticationManagerWhenConfiguredAuthenticationManagerThenTakes
...
+				.with(bearerToken(jwtTwo)))
+				.andExpect(status().isOk())
+				.andExpect(content().string("test-subject"));
+
+		mockWebServer(String.format(metadata, issuerThree, issuerThree));
+		mockWebServer(jwkSet);
+
+		this.mvc.perform(get("/authenticated")
+				.with(bearerToken(jwtThree)))
+				.andExpect(status().isUnauthorized());
```
fhanik: Can we narrow down the validation criteria, to perhaps verify that the failure was indeed from `invalid issuer`?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationManagerResolver.java
```diff
@@ -0,0 +1,159 @@
...
+	private static final OAuth2Error DEFAULT_INVALID_TOKEN = invalidToken("Invalid token");
+
+	private final AuthenticationManagerResolver<String> issuerAuthenticationManagerResolver;
+
+	private Converter<HttpServletRequest, String> issuerConverter = new JwtClaimIssuerConverter();
```
fhanik: Not `final`, but I don't see a way to configure this variable.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationManagerResolver.java
```diff
@@ -0,0 +1,159 @@
...
+		Assert.notNull(issuerAuthenticationManagerResolver, "issuerAuthenticationManagerResolver cannot be null");
+		this.issuerAuthenticationManagerResolver = issuerAuthenticationManagerResolver;
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public AuthenticationManager resolve(HttpServletRequest request) {
+		String issuer = this.issuerConverter.convert(request);
```
fhanik: What happens here if the JWT is encrypted? Maybe not a supported feature? (note in javadoc)

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationManagerResolver.java
```diff
@@ -0,0 +1,159 @@
...
+		AuthenticationManager authenticationManager = this.issuerAuthenticationManagerResolver.resolve(issuer);
+		if (authenticationManager == null) {
+			throw new OAuth2AuthenticationException(invalidToken("Invalid issuer " + issuer));
```
fhanik: The parent interface or this class don't document runtime exceptions. Should it documented?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationManagerResolver.java
```diff
@@ -0,0 +1,159 @@
...
+
+		TrustedIssuerJwtAuthenticationManagerResolver(Predicate<String> trustedIssuer) {
+			this.trustedIssuer = trustedIssuer;
+		}
+
+		@Override
+		public AuthenticationManager resolve(String issuer) {
+			if (this.trustedIssuer.test(issuer)) {
+				return this.authenticationManagers.computeIfAbsent(issuer, k -> {
+					JwtDecoder jwtDecoder = JwtDecoders.fromIssuerLocation(issuer);
```
fhanik: Do we need to document that changes on the authorization server will require a restart, since we only resolve metadata on the first request. Or should this be controlled? How do we do it when we don't use a resolver?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationManagerResolver.java
```diff
@@ -0,0 +1,159 @@
...
+		Assert.notEmpty(trustedIssuers, "trustedIssuers cannot be empty");
+		trustedIssuers = Collections.unmodifiableCollection(trustedIssuers);
+		this.issuerAuthenticationManagerResolver =
+				new TrustedIssuerJwtAuthenticationManagerResolver(trustedIssuers::contains);
+	}
+
+	/**
+	 * Construct a {@link JwtAuthenticationManagerResolver} using the provided parameters
+	 *
+	 * @param issuerAuthenticationManagerResolver a strategy for resolving the {@link AuthenticationManager} by the issuer
```
fhanik: I'd expand the javadoc a little bit here, since this can be a _shoot yourself in the foot_ constructor. The javadoc should state that the burden of trust verification based on issuer falls on the `issuerAuthenticationManagerResolver` implementation.

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationManagerResolverTests.java
```diff
@@ -0,0 +1,163 @@
...
+ * Tests for {@link JwtAuthenticationManagerResolver}
+ */
+public class JwtAuthenticationManagerResolverTests {
+	private static final String DEFAULT_RESPONSE_TEMPLATE = "{\n"
+			+ "    \"issuer\": \"%s\", \n"
+			+ "    \"jwks_uri\": \"%s/.well-known/jwks.json\" \n"
+			+ "}";
+
+	private String jwt = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyLCJpc3MiOiJ0cnVzdGVkIn0.YtGPgaMeQmxQD5ZDag8KkB9QO-5tG-EPQNn13sJ189Q";
+	private String evil = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyLCJpc3MiOiJcIiJ9.EpoIFAhHc_YDphrqwsaN5rZkDLpBp_w6IfXJaSjUau4";
```
fhanik: It would be easier for future test maintenance if this string was JSON. Then in the setup method, it gets converted to an encoded JWT.

That way I can change the JSON, and continue running the tests as I develop.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationManagerResolver.java
jzheaux: I think that we are okay to leave this out since this is the way that it behaves like when there is no resolver.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationManagerResolver.java
jzheaux: Spring Security doesn't support JWE, so I'm not sure that this needs to be called out. That said, I did clarify that this is for JWS tokens (see class-level JavaDoc). Feel free to resolve this convo if you agree.

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationManagerResolverTests.java
```diff
@@ -0,0 +1,186 @@
...
+
+	@Test
+	public void resolveWhenBearerTokenMalformedThenException() {
+		JwtAuthenticationManagerResolver authenticationManagerResolver =
+				new JwtAuthenticationManagerResolver("trusted");
+		MockHttpServletRequest request = new MockHttpServletRequest();
+		request.addHeader("Authorization", "Bearer jwt");
+		assertThatCode(() -> authenticationManagerResolver.resolve(request))
+				.isInstanceOf(OAuth2AuthenticationException.class)
+				.hasMessageNotContaining("Invalid issuer");
```
fhanik: Maybe test for the actual exception you're expecting, rather than the one you're _not_ expecting. Or be satisfied that it fails.

jzheaux: The message that is returned is one that I don't have control over. I want to confirm that the issuer is not tested itself.

jzheaux:
@dyroberts how well would this PR address some of the concerns you raised in #5385?

With it, you could do either:

```java
JwtAuthenticationManagerResolver authenticationManagerResolver =
        new JwtAuthenticationManagerResolver("list", "of", "issuers");

http
    .oauth2ResourceServer()
        .authenticationManagerResolver(authenticationManagerResolver)
```

or

```java
Map<String, AuthenticationManager> authenticationManagers = ...;
JwtAuthenticationManagerResolver authenticationManagerResolver =
        new JwtAuthenticationManagerResolver(authenticationManagers::get);

http
    .oauth2ResourceServer()
        .authenticationManagerResolver(authenticationManagerResolver);
```

dyroberts:
@jzheaux Looks really good! The only thing I could see as a slight improvement would be some auto-config to create a `JwtAuthenticationManagerResolver` from application properties. But maybe that's not inline with the spring-security conventions of today? Anyway, this really does cover well the case that I was referring to.

jzheaux:
@dyroberts we are still looking into the idea of mapping this to config properties. While this does appear to be a common way to handle multi-tenancy in resource servers, it's not clear whether or not multi-tenancy itself is so common as to want to add Spring Boot support.

That said, this class leaves open this possibility.

jzheaux:
@gburboz @xsreality @bertramn How well would this PR address the concerns you raised in #6778 and #5351?

xsreality:
Hi Josh, the PR looks good. My primary concern was to be able to add/remove tenants at runtime without needing to restart the application. The updated doc makes it clear how to do that. Thanks!

davidmelia:
@jzheaux will you be adding a reactive version of this class?  Thanks

jzheaux:
@davidmelia I think it makes sense to explore that, yes - I left a bit more detail in [the ticket you raised](https://github.com/spring-projects/spring-security/issues/7857).

RavindraSengar:
Hi,

I have similar use case.
One Backend API endpoint/resource
2 consumers of this BE API both will have their corresponding OAUTH2/OKTA authorization servers. Each with a different RSA key configured for creation of the JWT tokens.
In BE API, we want to accept tokens from either auth server and authenticate.
Problem - BE API/Resource is configured to use only 1 set of security.oauth2.resource.jwk.key-set-uri and security.jwt.issuer Is there a way we can mention 2 set of security.oauth2.resource.jwk.key-set-uri and security.jwt.issuer ? Or how to authenticate tokens from 2 different OKTA/OAUTH authorization servers ?

Project setup Details -

org.springframework.boot" version "2.3.1.RELEASE"
io.spring.dependency-management" version "1.0.8.RELEASE"
springSecurityVersion = "5.3.3.RELEASE"

I have gone through this support documentation but need some help to implement this. 
Not sure this is right place to ask this question. but asking it anyway. 
Thanks.

jzheaux:
Thanks for reaching out, @RavindraSengar! In the future, please don't double-post as it creates noise. I've left some suggestions for you about multiple issuers over in the other ticket where you asked a similar question.

Issue #7721, opened 2019-12-11
RequestCacheSpec not used on RedirectServerAuthenticationEntryPoint for OAuth2LoginSpec.configure


sdoxsee:
### Summary

The `RequestCache` set in the RequestCacheSpec is not used on RedirectServerAuthenticationEntryPoint for OAuth2LoginSpec.configure

### Actual Behavior

From `ServerHttpSecurity`
```
public class OAuth2LoginSpec {

        //...

	protected void configure(ServerHttpSecurity http) {

                //...

		oauthRedirectFilter.setRequestCache(http.requestCache.requestCache);

                //...

		MediaTypeServerWebExchangeMatcher htmlMatcher = new MediaTypeServerWebExchangeMatcher(
				MediaType.TEXT_HTML);
		htmlMatcher.setIgnoredMediaTypes(Collections.singleton(MediaType.ALL));
		Map<String, String> urlToText = http.oauth2Login.getLinks();
		if (urlToText.size() == 1) {
			http.defaultEntryPoints.add(new DelegateEntry(htmlMatcher, new RedirectServerAuthenticationEntryPoint(urlToText.keySet().iterator().next())));
		} else {
			http.defaultEntryPoints.add(new DelegateEntry(htmlMatcher, new RedirectServerAuthenticationEntryPoint("/login")));
		}

		http.addFilterAt(oauthRedirectFilter, SecurityWebFiltersOrder.HTTP_BASIC);
		http.addFilterAt(authenticationFilter, SecurityWebFiltersOrder.AUTHENTICATION);
	}
```

`RedirectServerAuthenticationEntryPoint`'s RequestCache remains the default `WebSessionServerRequestCache` rather than the one set on the `RequestCacheSpec` for `ServerHttpSecurity`

### Expected Behavior

I would think that `RedirectServerAuthenticationEntryPoint` should use the `RequestCacheSpec`'s `RequestCache`

### Configuration
See sample below

### Version

Spring Security 5.2.1

### Sample
See README and sample code on https://github.com/sdoxsee/gateway-sample

sdoxsee:
Thanks @fhanik!

PR #7734, opened 2019-12-12
Use the custom ServerRequestCache for Oauth2LoginSpec

fhanik:
Fixes gh-7721

Test case provided

### DIFF
--- a/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
@@ -76,9 +76,11 @@ import org.springframework.security.oauth2.client.userinfo.ReactiveOAuth2UserSer
 import org.springframework.security.oauth2.client.web.server.AuthenticatedPrincipalServerOAuth2AuthorizedClientRepository;
 import org.springframework.security.oauth2.client.web.server.OAuth2AuthorizationCodeGrantWebFilter;
 import org.springframework.security.oauth2.client.web.server.OAuth2AuthorizationRequestRedirectWebFilter;
+import org.springframework.security.oauth2.client.web.server.ServerAuthorizationRequestRepository;
 import org.springframework.security.oauth2.client.web.server.ServerOAuth2AuthorizationCodeAuthenticationTokenConverter;
 import org.springframework.security.oauth2.client.web.server.ServerOAuth2AuthorizationRequestResolver;
 import org.springframework.security.oauth2.client.web.server.ServerOAuth2AuthorizedClientRepository;
+import org.springframework.security.oauth2.client.web.server.WebSessionOAuth2ServerAuthorizationRequestRepository;
 import org.springframework.security.oauth2.client.web.server.authentication.OAuth2LoginAuthenticationWebFilter;
 import org.springframework.security.oauth2.core.endpoint.OAuth2AuthorizationRequest;
 import org.springframework.security.oauth2.core.oidc.user.OidcUser;
@@ -984,7 +986,7 @@ public class ServerHttpSecurity {
 
 		private ServerWebExchangeMatcher authenticationMatcher;
 
-		private ServerAuthenticationSuccessHandler authenticationSuccessHandler = new RedirectServerAuthenticationSuccessHandler();
+		private ServerAuthenticationSuccessHandler authenticationSuccessHandler;
 
 		private ServerAuthenticationFailureHandler authenticationFailureHandler;
 
@@ -1175,7 +1177,7 @@ public class ServerHttpSecurity {
 			authenticationFilter.setRequiresAuthenticationMatcher(getAuthenticationMatcher());
 			authenticationFilter.setServerAuthenticationConverter(getAuthenticationConverter(clientRegistrationRepository));
 
-			authenticationFilter.setAuthenticationSuccessHandler(this.authenticationSuccessHandler);
+			authenticationFilter.setAuthenticationSuccessHandler(getAuthenticationSuccessHandler(http));
 			authenticationFilter.setAuthenticationFailureHandler(getAuthenticationFailureHandler());
 			authenticationFilter.setSecurityContextRepository(this.securityContextRepository);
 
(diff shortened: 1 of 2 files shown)


sdoxsee:
Thanks @fhanik and @jgrandja! Any chance this could make it into 5.2.2 as well as it's a bug fix?

sdoxsee:
Also, the `authenticationSuccessHandler` has its own `RequestCache` but the `authenticationSuccessHandler` may or may not be provided as custom. Which `RequestCache` should it be using? It's own or the same one as the others?

fhanik:
Hi @sdoxsee 

I believe the success handler should be using the same [as the others](https://github.com/spring-projects/spring-security/blob/c40a17b4d1c8eefff416aefd7c3fd96a1c6fe914/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java#L3115-L3116)

https://github.com/spring-projects/spring-security/blob/c40a17b4d1c8eefff416aefd7c3fd96a1c6fe914/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java#L3115-L3116

sdoxsee:
Thanks @fhanik. I think that's for FormLoginSpec as opposed to OAuth2LoginSpec's configure?

fhanik:
Correct. So yes, it's still an issue in the OAuth2LoginSpec.

I'll update the PR to encapsulate both use scenarios.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -984,7 +986,7 @@ public ServerHttpSecurity oauth2Login(Customizer<OAuth2LoginSpec> oauth2LoginCus
 
 		private ServerWebExchangeMatcher authenticationMatcher;
 
-		private ServerAuthenticationSuccessHandler authenticationSuccessHandler = new RedirectServerAuthenticationSuccessHandler();
+		private ServerAuthenticationSuccessHandler authenticationSuccessHandler = null;
```
jgrandja: No need for `= null` since it will default to `null`

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -1175,6 +1177,11 @@ protected void configure(ServerHttpSecurity http) {
 			authenticationFilter.setRequiresAuthenticationMatcher(getAuthenticationMatcher());
 			authenticationFilter.setServerAuthenticationConverter(getAuthenticationConverter(clientRegistrationRepository));
 
+			if (this.authenticationSuccessHandler == null) {
```
jgrandja: Can we move this to `getAuthenticationSuccessHandler()` similar to `getAuthenticationFailureHandler()`

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -1183,11 +1190,15 @@ protected void configure(ServerHttpSecurity http) {
...
+			String authenticationEntryPointRedirectPath;
 			if (urlToText.size() == 1) {
-				http.defaultEntryPoints.add(new DelegateEntry(htmlMatcher, new RedirectServerAuthenticationEntryPoint(urlToText.keySet().iterator().next())));
+				authenticationEntryPointRedirectPath = urlToText.keySet().iterator().next();
 			} else {
-				http.defaultEntryPoints.add(new DelegateEntry(htmlMatcher, new RedirectServerAuthenticationEntryPoint("/login")));
+				authenticationEntryPointRedirectPath = "/login";
 			}
+			RedirectServerAuthenticationEntryPoint entryPoint = new RedirectServerAuthenticationEntryPoint(authenticationEntryPointRedirectPath);
+			entryPoint.setRequestCache(http.requestCache.requestCache);
```
jgrandja: I'm not sure if `RedirectServerAuthenticationEntryPoint` needs the same `requestCache`? I think we only want to save the request when an authentication fails only. Does a test fail without this change?

sdoxsee: Hey @jgrandja. That's my main, issue. I'm using a react application on a different port and and want to save the "referer" header into a `RequestCache` instead of the normal entry point. I'm unable to get to the `RequestCache` on the `RedirectServerAuthenticationEntryPoint` without this change. I'm thinking something similar to FormLoginSpec: https://github.com/spring-projects/spring-security/blob/master/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java#L3118

sdoxsee: I've found a better way to accomplish what I'm looking for by doing employing x-forward* on my proxy and `server.forward-headers-strategy: framework` on my spring-cloud-gateway (see proxy [reference](https://docs.spring.io/spring-security/site/docs/current/reference/html/webflux-oauth2.html#webflux-oauth2-login-sample-redirect)) but still wonder...

There are at least 3 places in the oauth2login where a `RequestCache` is set (defaultEntryPoint, authenticationSuccessHandler, and OAuth2AuthorizationRequestRedirectWebFilter) . Currently, only `OAuth2AuthorizationRequestRedirectWebFilter` is configurable using the `RequestCacheSpec`. Should they all have their own independent `RequestCache` implementations and how should oauth2Login differ from formLogin WRT `RequestCache`?

fhanik: @jgrandja I am under the impression that all components that participate in a flow would need to share a request cache. And the cache would need to remain the same between components that save the request and read the saved request.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
fhanik: Completed

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
fhanik: Fixed

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
jgrandja: @fhanik @sdoxsee Sorry for the confusion on my end. I made the assumption that a failed request is saved in `ExceptionTranslationWebFilter` instead of `RedirectServerAuthenticationEntryPoint`, as this is the way the Servlet counterpart `ExceptionTranslationFilter` is implemented. However, the reactive implementation is different and indeed the `RedirectServerAuthenticationEntryPoint` is responsible for saving the failed request.

@sdoxsee 

> Should they all have their own independent RequestCache implementations

No, all components participating in the same flow should share the same request cache.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -76,9 +76,11 @@
 import org.springframework.security.oauth2.client.web.server.AuthenticatedPrincipalServerOAuth2AuthorizedClientRepository;
 import org.springframework.security.oauth2.client.web.server.OAuth2AuthorizationCodeGrantWebFilter;
 import org.springframework.security.oauth2.client.web.server.OAuth2AuthorizationRequestRedirectWebFilter;
+import org.springframework.security.oauth2.client.web.server.ServerAuthorizationRequestRepository;
```
jgrandja: Remove unused import

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -76,9 +76,11 @@
...
 import org.springframework.security.oauth2.client.web.server.ServerOAuth2AuthorizationCodeAuthenticationTokenConverter;
 import org.springframework.security.oauth2.client.web.server.ServerOAuth2AuthorizationRequestResolver;
 import org.springframework.security.oauth2.client.web.server.ServerOAuth2AuthorizedClientRepository;
+import org.springframework.security.oauth2.client.web.server.WebSessionOAuth2ServerAuthorizationRequestRepository;
```
jgrandja: Remove unused import

jgrandja:
@fhanik I noticed you applied the "for: backport-to-5.2.x" to this PR, but it needs to be applied to the associated issue instead. The backport bot does not work on PR's.

Issue #4557, opened 2017-09-20
Provide XML namespace support for OAuth2Login


tiwatsuka:
### Summary

Does it have any plan to add  the way to configure OAuth2Login with XML namespace?
I think there are some projects still using XML configuration for security settings.
If we can define the beans related OAuth2Login just by setting about ClientRegistration with XML, it's very useful.

Related #5184

jgrandja:
@tiwatsuka We haven't decided whether to add XML namespace support. It definitely won't make it into 5.0 GA. We may consider it in a subsequent release.

tiwatsuka:
@jgrandja I see. I'll expect that it is finally added. Thank you for your consideration.

jgrandja:
@tiwatsuka Just to confirm, we will be adding XML namespace support. We'll do our best to get this in 5.1.

pthorson:
So 5.2 then or not at all? I have a large existing Spring MVC project with XML configuration that I need to add OAuth to today. Can I mix the namespace and Java config?

jgrandja:
@pthorson This won't get into 5.2. Unless you'd be interested in submitting a PR? There are other higher priority items at the moment and there hasn't been much demand to add this support as of yet. 

That being said, you should be able to mix namespace and Java config - see [Importing XML Configuration](https://docs.spring.io/spring-boot/docs/2.1.0.RELEASE/reference/htmlsingle/#using-boot-importing-xml-configuration)

pthorson:
Understood. Not sure of appetite here for a PR. That being said, is OAuth 2.0 and Open ID Connect possible now without Spring Boot being used? Based on the matrix it looks like Spring Boot 1.5x was previously required.

rwinch:
@pthorson The new support does not require Spring Boot (however, it is certainly easier to use with it)

bendem:
Hi, I'd like to request once again that feature. I just had to migrate our xml config for security to java config because there was no equivalent to `org.springframework.security.config.annotation.web.builders.HttpSecurity#oauth2Login` in xml. The Java config is less clear and more error prone. It's really annoying that there is no parity between the capabilities of `org.springframework.security.config.annotation.web.builders.HttpSecurity` and the spring-security xml namespace.

jgrandja:
[...]

PR #7743, opened 2019-12-13
Add XML namespace support for oauth2Login()

jgrandja:
Fixes gh-4557

### DIFF
--- a/config/src/test/resources/org/springframework/security/config/http/OAuth2LoginBeanDefinitionParserTests-SingleClientRegistration.xml
+++ b/config/src/test/resources/org/springframework/security/config/http/OAuth2LoginBeanDefinitionParserTests-SingleClientRegistration.xml
@@ -1,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2002-2019 the original author or authors.
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~       https://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<b:beans xmlns:b="http://www.springframework.org/schema/beans"
+		xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		xmlns="http://www.springframework.org/schema/security"
+		xsi:schemaLocation="
+			http://www.springframework.org/schema/security
+			https://www.springframework.org/schema/security/spring-security.xsd
+			http://www.springframework.org/schema/beans
+			https://www.springframework.org/schema/beans/spring-beans.xsd">
+
+	<http auto-config="true">
+		<oauth2-login/>
+	</http>
+
+	<client-registrations>
+		<client-registration registration-id="google-login"
+							 client-id="google-client-id"
+							 client-secret="google-client-secret"
+							 client-authentication-method="basic"
+							 authorization-grant-type="authorization_code"
+							 redirect-uri="{baseUrl}/login/oauth2/code/{registrationId}"
+							 scope="openid,profile,email"
+							 client-name="Google"
+							 provider-id="google"/>
+		<provider provider-id="google"
+				  authorization-uri="https://accounts.google.com/o/oauth2/v2/auth"
+				  token-uri="https://www.googleapis.com/oauth2/v4/token"
+				  userinfo-uri="https://www.googleapis.com/oauth2/v3/userinfo"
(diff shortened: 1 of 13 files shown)


jgrandja:
@rh-id Here is the initial draft ready for you to take over.

Take a look at the 2nd commit and you will see what needs to be implemented.

There is one [test](https://github.com/spring-projects/spring-security/pull/7743/files#diff-355c215b3327f2d76f1a9268b69505b2R44) that fails but should pass after the `BeanDefinitionParser`'s are implemented.

To start things off, I would suggest that you get familiar with `HttpSecurity.oauth2Login()` first since this is the equivalent implementation required for the xml namespace support. If you know this well than the xml namespace implementation should make things easier for you.

Here are a couple of tests to check out:

`OAuth2LoginApplicationTests`  - spring-security-samples-boot-oauth2login

`org.springframework.security.config.annotation.web.configurers.oauth2.client.OAuth2LoginConfigurerTests`


Let me know if you have any questions.

config/src/test/resources/org/springframework/security/config/http/OAuth2LoginBeanDefinitionParserTests-SingleClientRegistration.xml
```diff
@@ -0,0 +1,58 @@
...
+<b:beans xmlns:b="http://www.springframework.org/schema/beans"
+		xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		xmlns="http://www.springframework.org/schema/security"
+		xsi:schemaLocation="
+			http://www.springframework.org/schema/security
+			https://www.springframework.org/schema/security/spring-security.xsd
+			http://www.springframework.org/schema/beans
+			https://www.springframework.org/schema/beans/spring-beans.xsd">
+
+	<http auto-config="true">
```
jgrandja: @rwinch @jzheaux When you have a moment, please review this `oauth2Login()` xml sample configuration and provide feedback. This sample config explicitly configures all possible configurations. However, most are optional as defined in the schema. Also, `issuer-uri` is also defined in the schema but it's not in this xml instance.

rh-id:
hi @jgrandja , I had push to https://github.com/jgrandja/spring-security/pull/2 should i just push to this project instead?

There are some questions that I might need to ask.
1. the class "SecurityFilters" and "FilterComparator". "SecurityFilters" enum stored the filter and order information while "FilterComparator" had filter ordering as comparator. does it mean that we have to maintain both classes so that it will have the same filter order?

2. I assume both "BeanDefinitionParser" classes will need to be decoupled from "org.springframework.security.oauth2.client" package hence I will need to use class name instead of class instance itself as parameter when instantiating "RootBeanDefinition" right? if so then ClientRegistration might have problem since it doesn't have public constructor.

jgrandja:
@rh-id The plan is to work off of this PR. You will see 2 commits in this PR with the initial work. In the 2nd commit "Add template for BeanDefinitionParser implementations", you will see instructions on what to implement. All updates should build off of these 2 commits. I'll close the PR you submitted to my fork.

> the class "SecurityFilters" and "FilterComparator". "SecurityFilters" enum stored the filter and order information while "FilterComparator" had filter ordering as comparator. does it mean that we have to maintain both classes so that it will have the same filter order?

`FilterComparator` is used for Java Config and it has already been updated as part of `HttpSecurity.oauth2Login()` so you shouldn't have to make any changes here. You will only need to update `SecurityFilters` since this is used for XML Config.

> I assume both "BeanDefinitionParser" classes will need to be decoupled from "org.springframework.security.oauth2.client" package hence I will need to use class name instead of class instance itself as parameter when instantiating "RootBeanDefinition" right? if so then ClientRegistration might have problem since it doesn't have public constructor.

I'm not sure I follow you here. It's not clear to me what the issue is. FYI, `ClientRegistration` can be instantiated using `withRegistrationId()` factory method.

rh-id:
hi @jgrandja sorry I'm not sure if I get you as well. yes, I understand that we will work based on this PR and I had proceed to finish the implementation of the XML support and test it. 

should I create a PR to your fork ? or create  PR in this project? my latest commit at: https://github.com/rh-id/spring-security

and thanks for the info.

jgrandja:
@rh-id 

> sorry I'm not sure if I get you as well

Ok lets simplify things. Please run the following:

`git fetch upstream pull/7743/head:oauth2login-xmlconfig`

NOTE: `upstream` in my setup is `git@github.com:spring-projects/spring-security.git`. It might be `origin` in your setup.

You should now have a local branch called `oauth2login-xmlconfig` that checks out this PR. You will notice 2 commits. Your work/commits should add on top of these 2 commits. After you have completed the work then you can simply submit a new PR. I'll close off this Draft PR.

Does this clarify things?

One important thing to mention is that [5.3.0.RC1](https://github.com/spring-projects/spring-security/milestones) is due Feb 5 and this feature needs to be merged by then. This gives approx. 3 1/2 weeks but I would like to see a complete implementation within the next 2 weeks if possible to provide enough time for review. Do you feel you have the time to complete based on these timelines? I don't want you to feel rushed or cause unnecessary stress if you have other job priorities. If you feel you cannot deliver with these timelines then that's fine as well and I'll take it over. Please let me know either way so we can co-ordinate.

rh-id:
hi @jgrandja i had create the PR https://github.com/spring-projects/spring-security/pull/7821

jgrandja:
Closing in favour of #7821

PR #7744, opened 2019-12-14
Don't cache requests with `Accept: text/event-stream` by default.

joshiste:
The eventstream requests is typically not directly invoked by the browser.
And even more unfortunately the browser api for `Eventsource` doesn't allow the set additional headers as `X-Requested-With: XMLHttpRequest`..


fhanik:
Merged to `master` and `5.2.x`

PR #7755, opened 2019-12-18
Resolve JavaType only once for whitelisted class

matthiasstock:
### Summary

The Jackson 2 `JavaType` is unnecessarily resolved twice when a class is whitelisted. This should be the most likely case. The delegate doing this may be a `ClassNameIdResolver` whose operations seem rather complex, so it would be a good idea to avoid a call.

### Actual Behavior

JavaTypes are resolved twice for a whitelisted class.

### Expected Behavior

JavaTypes are resolved only once for a whitelisted class.

### DIFF
--- a/core/src/main/java/org/springframework/security/jackson2/SecurityJackson2Modules.java
+++ b/core/src/main/java/org/springframework/security/jackson2/SecurityJackson2Modules.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2018 the original author or authors.
+ * Copyright 2015-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/springframework/security/jackson2/SecurityJackson2Modules.java
```diff
@@ -225,7 +225,7 @@ public JavaType typeFromId(DatabindContext context, String id) throws IOExceptio
 			JavaType result = delegate.typeFromId(context, id);
```
eleftherias: Please update the Apache license header in this file. For more details see the [contribution guidelines](https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.md#update-apache-license-header-to-modified-files-as-necessary).

eleftherias:
Thanks for the PR @matthiasstock! This is now merged into master.

Issue #7699, opened 2019-12-04
Introduce Reactive OAuth2Authorization success/failure handlers


philsttr:
### Summary

When an OAuth2 Client sends a token to a downstream OAuth2 Resource Server, and the OAuth2 Resource Server returns an HTTP 401 response, I would like to prevent the OAuth2 Client from sending that same token in future requests (and instead retrieve a new token for future requests).

Currently, spring-security-oauth2-client provides good support for re-retrieving/refreshing tokens before they expire.  However, spring-security-oauth2-client does not provide good support for re-retrieving/refreshing access tokens when they have become invalid for reasons other than expiration.  A good signal that a token is no longer valid is when an OAuth2 Resource Server returns an HTTP 401 response to the client.

Both `ServerOAuth2AuthorizedClientRepository` and `ReactiveOAuth2AuthorizedClientService` provide a `removeAuthorizedClient` method to remove the cached `OAuth2AuthorizedClient`.  It would be nice if spring-security-oauth2-client provided _something_ that actually called that method when an HTTP 401 response is received from an OAuth2 Resource Server.

A possible implementation might be:

1. the `DefaultReactiveOAuth2AuthorizedClientManager` (and the future `AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager` from #7589) could put a `Mono<Void>` in the subscriber context that will remove the token from the cache when subscribed
   * `DefaultReactiveOAuth2AuthorizedClientManager` would put a `Mono<Void>` that invokes its `ServerOAuth2AuthorizedClientRepository.removeAuthorizedClient`
   * `AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager` would put a `Mono<Void>` that invokes its `ReactiveOAuth2AuthorizedClientService.removeAuthorizedClient`
2. A new `ExchangeFilterFunction` (or one of the existing ones) checks for HTTP 401 responses, and chains the `Mono<Void>` from step 1 if it exists
   * usage of the subscriber context would decouple the `ExchangeFilterFunction` from the implementation of how to remove the cached authorized client  (i.e. separate filter function implementations would not be needed depending on whether a `ServerOAuth2AuthorizedClientRepository` or `ReactiveOAuth2AuthorizedClientService` is being used to persist `OAuth2AuthorizedClient` s)
[...]

Issue #7583, opened 2019-10-30
RefreshTokenOAuth2AuthorizedClientProvider does not handle expired refresh token


janzyka:
### Summary
Refresh tokens don't support timeouts - if refresh token expires application will keep trying to refresh access token through it forever.

Also all class around OAuth2 client are final which doesn't allow to fix the problem on my side temporary.

### Actual Behavior
1. App authorizes client and gets access & refresh tokens
2. App is making requests through access token
3. Access token expires, app is able to refresh
4. Now refresh token expires (!)
5. Attempt to refresh fails -> request fails
6. Any subsequent access leads to attempt to refresh through expired refresh token
7. Back to 5.

=> application will never recover

### Expected Behavior

In order of preference

Option 1 - allow & handle refresh tokens expiry
Option 2 - add flag to optionally disable refresh tokens in `PasswordOAuth2AuthorizedClientProvider`
Option 3 - if refresh was attempted and failed remove client from `authorizeClientRepository` so the subsequent call starts from scratch allowing to recover

Idea 1 - don't make all classes final, it is difficult to add own behaviour. I remember nearly all classes in Spring used to be opened for extension.
### Configuration

```
    @Bean
    public WebClient webClient(ClientRegistrationRepository clientRegistrationRepository, OAuth2AuthorizedClientService authorizedClientService) {

        AuthorizedClientServiceOAuth2AuthorizedClientManager manager = new AuthorizedClientServiceOAuth2AuthorizedClientManager(clientRegistrationRepository, authorizedClientService);
        manager.setAuthorizedClientProvider(new DelegatingOAuth2AuthorizedClientProvider(
                new RefreshTokenOAuth2AuthorizedClientProvider(),
                new PasswordOAuth2AuthorizedClientProvider()));

        Map<String, Object> passwordAttributes = new HashMap<>();
        passwordAttributes.put(OAuth2AuthorizationContext.USERNAME_ATTRIBUTE_NAME, "user");
        passwordAttributes.put(OAuth2AuthorizationContext.PASSWORD_ATTRIBUTE_NAME, "password");

        manager.setContextAttributesMapper(request -> passwordAttributes);

        ServletOAuth2AuthorizedClientExchangeFilterFunction oauth2 = new ServletOAuth2AuthorizedClientExchangeFilterFunction(manager);

        oauth2.setDefaultClientRegistrationId("oauth-service-to-service-client");

        return WebClient.builder()
                .filter(oauth2)
[...]

PR #7756, opened 2019-12-18
Introduce Reactive OAuth2Authorization success/failure handlers

philsttr:
All `ReactiveOAuth2AuthorizedClientManager`s now have authorization success/failure handlers.
Default success handlers will save authorized clients for future requests.
Default failure handlers will remove previously saved authorized clients.

`ServerOAuth2AuthorizedClientExchangeFilterFunction` also makes use of a
failure handler in the case of unauthorized or forbidden http status code.

The main use cases now handled are
- remove authorized client when an authorization server indicates that a refresh token is no longer valid (when authorization server returns invalid_grant)
- remove authorized client when a resource server indicates that an access token is no longer valid (when resource server returns invalid_token)

Introduced `ClientAuthorizationException` to capture details needed when removing an authorized client.
All `ReactiveOAuth2AccessTokenResponseClient`s now throw a `ClientAuthorizationException` on failures.

Created `AbstractWebClientReactiveOAuth2AccessTokenResponseClient` to unify common logic between all `ReactiveOAuth2AccessTokenResponseClient`s.

Fixes gh-7699

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,7 +17,10 @@ package org.springframework.security.oauth2.client;
 
 import org.springframework.security.core.Authentication;
 import org.springframework.security.oauth2.client.registration.ReactiveClientRegistrationRepository;
+import org.springframework.security.oauth2.client.web.DefaultReactiveOAuth2AuthorizedClientManager;
+import org.springframework.security.oauth2.core.OAuth2AuthorizationException;
 import org.springframework.util.Assert;
+import org.springframework.web.server.ServerWebExchange;
 import reactor.core.publisher.Mono;
 
 import java.util.Collections;
@@ -26,10 +29,33 @@ import java.util.function.Function;
 
 /**
  * An implementation of an {@link ReactiveOAuth2AuthorizedClientManager}
- * that is capable of operating outside of a {@code ServerHttpRequest} context,
+ * that is capable of operating outside of the context of a {@link ServerWebExchange},
  * e.g. in a scheduled/background thread and/or in the service-tier.
  *
- * <p>This is a reactive equivalent of {@link org.springframework.security.oauth2.client.AuthorizedClientServiceOAuth2AuthorizedClientManager}</p>
+ * <p>(When operating <em>within</em> the context of a {@link ServerWebExchange},
+ * use {@link DefaultReactiveOAuth2AuthorizedClientManager} instead.)</p>
+ *
+ * <p>This is a reactive equivalent of {@link org.springframework.security.oauth2.client.AuthorizedClientServiceOAuth2AuthorizedClientManager}.</p>
+ *
+ * <h2>Authorized Client Persistence</h2>
+ *
+ * <p>This client manager utilizes a {@link ReactiveOAuth2AuthorizedClientService}
+ * to persist {@link OAuth2AuthorizedClient}s.</p>
+ *
+ * <p>By default, when an authorization attempt succeeds, the {@link OAuth2AuthorizedClient}
+ * will be saved in the authorized client service.
+ * This functionality can be changed by configuring a custom {@link ReactiveOAuth2AuthorizationSuccessHandler}
+ * via {@link #setAuthorizationFailureHandler(ReactiveOAuth2AuthorizationFailureHandler)}.</p>
+ *
+ * <p>By default, when an authorization attempt fails due to an
+ * {@value org.springframework.security.oauth2.core.OAuth2ErrorCodes#INVALID_GRANT} error,
+ * the previously saved {@link OAuth2AuthorizedClient}
+ * will be removed from the authorized client service.
+ * (The {@value org.springframework.security.oauth2.core.OAuth2ErrorCodes#INVALID_GRANT}
+ * error generally occurs when a refresh token that is no longer valid
+ * is used to retrieve a new access token.)
+ * This functionality can be changed by configuring a custom {@link ReactiveOAuth2AuthorizationFailureHandler}
+ * via {@link #setAuthorizationFailureHandler(ReactiveOAuth2AuthorizationFailureHandler)}.</p>
  *
  * @author Ankur Pathak
  * @author Phil Clay
@@ -45,6 +71,8 @@ public final class AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager
 	private final ReactiveOAuth2AuthorizedClientService authorizedClientService;
 	private ReactiveOAuth2AuthorizedClientProvider authorizedClientProvider = context -> Mono.empty();
 	private Function<OAuth2AuthorizeRequest, Mono<Map<String, Object>>> contextAttributesMapper = new DefaultContextAttributesMapper();
+	private ReactiveOAuth2AuthorizationSuccessHandler authorizationSuccessHandler;
+	private ReactiveOAuth2AuthorizationFailureHandler authorizationFailureHandler;
 
 	/**
 	 * Constructs an {@code AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager} using the provided parameters.
@@ -59,6 +87,8 @@ public final class AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager
 		Assert.notNull(authorizedClientService, "authorizedClientService cannot be null");
 		this.clientRegistrationRepository = clientRegistrationRepository;
 		this.authorizedClientService = authorizedClientService;
+		this.authorizationSuccessHandler = new SaveAuthorizedClientReactiveOAuth2AuthorizationSuccessHandler(authorizedClientService);
+		this.authorizationFailureHandler = new RemoveAuthorizedClientReactiveOAuth2AuthorizationFailureHandler(authorizedClientService);
 	}
 
 	@Override
@@ -66,7 +96,7 @@ public final class AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager
 		Assert.notNull(authorizeRequest, "authorizeRequest cannot be null");
 
 		return createAuthorizationContext(authorizeRequest)
-				.flatMap(this::authorizeAndSave);
+				.flatMap(authorizationContext -> authorize(authorizationContext, authorizeRequest.getPrincipal()));
 	}
 
 	private Mono<OAuth2AuthorizationContext> createAuthorizationContext(OAuth2AuthorizeRequest authorizeRequest) {
@@ -90,12 +120,33 @@ public final class AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager
 						}));
 	}
 
-	private Mono<OAuth2AuthorizedClient> authorizeAndSave(OAuth2AuthorizationContext authorizationContext) {
+	/**
+	 * Performs authorization, and notifies either the {@link #authorizationSuccessHandler}
+	 * or {@link #authorizationFailureHandler}, depending on the authorization result.
+	 *
+	 * @param authorizationContext the context to authorize
+	 * @param principal the principle to authorize
+	 * @return a {@link Mono} that emits the authorized client after the authorization attempt succeeds
+	 *         and the {@link #authorizationSuccessHandler} has completed,
+	 *         or completes with an exception after the authorization attempt fails
+	 *         and the {@link #authorizationFailureHandler} has completed
+	 */
+	private Mono<OAuth2AuthorizedClient> authorize(
+			OAuth2AuthorizationContext authorizationContext,
+			Authentication principal) {
 		return this.authorizedClientProvider.authorize(authorizationContext)
-				.flatMap(authorizedClient -> this.authorizedClientService.saveAuthorizedClient(
+				// Notify the authorizationSuccessHandler of the successful authorization
+				.flatMap(authorizedClient -> authorizationSuccessHandler.onAuthorizationSuccess(
 								authorizedClient,
-								authorizationContext.getPrincipal())
+								principal,
+								Collections.emptyMap())
 						.thenReturn(authorizedClient))
+				// Notify the authorizationFailureHandler of the failed authorization
+				.onErrorResume(OAuth2AuthorizationException.class, authorizationException -> authorizationFailureHandler.onAuthorizationFailure(
+								authorizationException,
+								principal,
+								Collections.emptyMap())
+						.then(Mono.error(authorizationException)))
 				.switchIfEmpty(Mono.defer(()-> Mono.justOrEmpty(authorizationContext.getAuthorizedClient())));
 	}
 
@@ -122,6 +173,36 @@ public final class AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager
 	}
 
 	/**
+	 * Sets the handler that handles successful authorizations.
+	 *
+	 * <p>A {@link SaveAuthorizedClientReactiveOAuth2AuthorizationSuccessHandler}
+	 * is used by default.</p>
+	 *
+	 * @param authorizationSuccessHandler the handler that handles successful authorizations.
+	 * @see SaveAuthorizedClientReactiveOAuth2AuthorizationSuccessHandler
+	 * @since 5.3
+	 */
+	public void setAuthorizationSuccessHandler(ReactiveOAuth2AuthorizationSuccessHandler authorizationSuccessHandler) {
+		Assert.notNull(authorizationSuccessHandler, "authorizationSuccessHandler cannot be null");
+		this.authorizationSuccessHandler = authorizationSuccessHandler;
+	}
+
+	/**
+	 * Sets the handler that handles authorization failures.
+	 *
+	 * <p>A {@link RemoveAuthorizedClientReactiveOAuth2AuthorizationFailureHandler}
+	 * is used by default.</p>
+	 *
+	 * @param authorizationFailureHandler the handler that handles authorization failures.
+	 * @see RemoveAuthorizedClientReactiveOAuth2AuthorizationFailureHandler
+	 * @since 5.3
+	 */
+	public void setAuthorizationFailureHandler(ReactiveOAuth2AuthorizationFailureHandler authorizationFailureHandler) {
+		Assert.notNull(authorizationFailureHandler, "authorizationFailureHandler cannot be null");
+		this.authorizationFailureHandler = authorizationFailureHandler;
+	}
+
+	/**
 	 * The default implementation of the {@link #setContextAttributesMapper(Function) contextAttributesMapper}.
 	 */
 	public static class DefaultContextAttributesMapper implements Function<OAuth2AuthorizeRequest, Mono<Map<String, Object>>> {
@@ -134,4 +215,5 @@ public final class AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager
 			return Mono.fromCallable(() -> mapper.apply(authorizeRequest));
 		}
 	}
+
 }
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/ReactiveOAuth2AuthorizationFailureHandler.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/ReactiveOAuth2AuthorizationFailureHandler.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.client;
+
+import org.springframework.security.core.Authentication;
+import org.springframework.security.oauth2.core.OAuth2AuthorizationException;
+import reactor.core.publisher.Mono;
+
+import java.util.Map;
+
+/**
+ * Handles when an OAuth 2.0 Authorized Client
+ * fails to authorize (or re-authorize)
+ * via the authorization server or resource server.
+ *
+ * @author Phil Clay
+ * @since 5.3
+ */
+@FunctionalInterface
+public interface ReactiveOAuth2AuthorizationFailureHandler {
+
+	/**
+	 * Called when an OAuth 2.0 Authorized Client
+	 * fails to authorize (or re-authorize)
+	 * via the authorization server or resource server.
+	 *
+	 * @param authorizationException the exception that contains details about what failed
+	 * @param principal the {@code Principal} that was attempted to be authorized
+	 * @param attributes an immutable {@code Map} of extra optional attributes present under certain conditions.
+	 *                   For example, this might contain a {@link org.springframework.web.server.ServerWebExchange ServerWebExchange}
+	 *                   if the authorization was performed within the context of a {@code ServerWebExchange}.
+	 * @return an empty {@link Mono} that completes after this handler has finished handling the event.
+	 */
+	Mono<Void> onAuthorizationFailure(
+			OAuth2AuthorizationException authorizationException,
+			Authentication principal,
+			Map<String, Object> attributes);
+}
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/ReactiveOAuth2AuthorizationSuccessHandler.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/ReactiveOAuth2AuthorizationSuccessHandler.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.client;
+
+import org.springframework.security.core.Authentication;
+import reactor.core.publisher.Mono;
+
+import java.util.Map;
+
+/**
+ * Handles when an OAuth 2.0 Authorized Client
+ * has been successfully authorized (or re-authorized)
+ * via the authorization server.
+ *
+ * @author Phil Clay
+ * @since 5.3
+ */
+@FunctionalInterface
+public interface ReactiveOAuth2AuthorizationSuccessHandler {
+
+	/**
+	 * Called when an OAuth 2.0 Authorized Client
+	 * has been successfully authorized (or re-authorized)
+	 * via the authorization server.
+	 *
+	 * @param authorizedClient the client that was successfully authorized
+	 * @param principal the {@code Principal} that was authorized
+	 * @param attributes an immutable {@code Map} of extra optional attributes present under certain conditions.
+	 *                   For example, this might contain a {@link org.springframework.web.server.ServerWebExchange ServerWebExchange}
+	 *                   if the authorization was performed within the context of a {@code ServerWebExchange}.
+	 * @return an empty {@link Mono} that completes after this handler has finished handling the event.
+	 */
+	Mono<Void> onAuthorizationSuccess(
+			OAuth2AuthorizedClient authorizedClient,
+			Authentication principal,
+			Map<String, Object> attributes);
+
+}
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RemoveAuthorizedClientReactiveOAuth2AuthorizationFailureHandler.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RemoveAuthorizedClientReactiveOAuth2AuthorizationFailureHandler.java
@@ -1,0 +1,169 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.client;
+
+import org.springframework.security.core.Authentication;
+import org.springframework.security.oauth2.client.web.server.ServerOAuth2AuthorizedClientRepository;
+import org.springframework.security.oauth2.core.OAuth2AuthorizationException;
+import org.springframework.security.oauth2.core.OAuth2ErrorCodes;
+import org.springframework.util.Assert;
+import org.springframework.web.server.ServerWebExchange;
+import reactor.core.publisher.Mono;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * An authorization failure handler that removes authorized clients from a
+ * {@link ServerOAuth2AuthorizedClientRepository}
+ * or a {@link ReactiveOAuth2AuthorizedClientService}.
+ * for specific OAuth 2.0 error codes.
+ *
+ * @author Phil Clay
+ * @since 5.3
+ */
+public class RemoveAuthorizedClientReactiveOAuth2AuthorizationFailureHandler implements ReactiveOAuth2AuthorizationFailureHandler {
+
+	/**
+	 * The default OAuth2 error codes that will trigger removal of the authorized client.
+	 * @see OAuth2ErrorCodes
+	 */
+	public static final Set<String>	DEFAULT_REMOVE_AUTHORIZED_CLIENT_ERROR_CODES = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(
+			/*
+			 * Returned from resource servers when an access token provided is expired, revoked,
+			 * malformed, or invalid for other reasons.
+			 *
+			 * Note that this is needed because the ServerOAuth2AuthorizedClientExchangeFilterFunction
+			 * delegates this type of failure received from a resource server
+			 * to this failure handler.
+			 */
+			OAuth2ErrorCodes.INVALID_TOKEN,
+			/*
+			 * Returned from authorization servers when a refresh token is invalid, expired, revoked,
+			 * does not match the redirection URI used in the authorization request, or was issued to another client.
+			 */
+			OAuth2ErrorCodes.INVALID_GRANT)));
+
+	/**
+	 * A delegate that removes clients from either a
+	 * {@link ServerOAuth2AuthorizedClientRepository}
+	 * or a
+	 * {@link ReactiveOAuth2AuthorizedClientService}
+	 * if the error code is one of the {@link #removeAuthorizedClientErrorCodes}.
+	 */
+	private final OAuth2AuthorizedClientRemover delegate;
+
+	/**
+	 * The OAuth2 Error Codes which will trigger removal of an authorized client.
+	 * @see OAuth2ErrorCodes
+	 */
+	private final Set<String> removeAuthorizedClientErrorCodes;
+
+	@FunctionalInterface
+	private interface OAuth2AuthorizedClientRemover {
+		Mono<Void> removeAuthorizedClient(
+				String clientRegistrationId,
+				Authentication principal,
+				Map<String, Object> attributes);
+	}
+
+	/**
+	 * @param authorizedClientRepository The repository from which authorized clients will be removed
+	 * 		  if the error code is one of the {@link #DEFAULT_REMOVE_AUTHORIZED_CLIENT_ERROR_CODES}.
+	 */
+	public RemoveAuthorizedClientReactiveOAuth2AuthorizationFailureHandler(ServerOAuth2AuthorizedClientRepository authorizedClientRepository) {
+		this(authorizedClientRepository, DEFAULT_REMOVE_AUTHORIZED_CLIENT_ERROR_CODES);
+	}
+
+	/**
+	 * @param authorizedClientRepository The repository from which authorized clients will be removed
+	 * 		 if the error code is one of the {@code removeAuthorizedClientErrorCodes}.
+	 * @param removeAuthorizedClientErrorCodes the OAuth2 Error Codes which will trigger removal of an authorized client.
+	 * @see OAuth2ErrorCodes
+	 */
+	public RemoveAuthorizedClientReactiveOAuth2AuthorizationFailureHandler(
+			ServerOAuth2AuthorizedClientRepository authorizedClientRepository,
+			Set<String> removeAuthorizedClientErrorCodes) {
+		Assert.notNull(authorizedClientRepository, "authorizedClientRepository cannot be null");
+		Assert.notNull(removeAuthorizedClientErrorCodes, "removeAuthorizedClientErrorCodes cannot be null");
+		this.removeAuthorizedClientErrorCodes = Collections.unmodifiableSet(new HashSet<>(removeAuthorizedClientErrorCodes));
+		this.delegate = (clientRegistrationId, principal, attributes) ->
+				authorizedClientRepository.removeAuthorizedClient(
+						clientRegistrationId,
+						principal,
+						(ServerWebExchange) attributes.get(ServerWebExchange.class.getName()));
+	}
+
+	/**
+	 * @param authorizedClientService the service from which authorized clients will be removed
+	 * 		  if the error code is one of the {@code removeAuthorizedClientErrorCodes}.
+	 */
+	public RemoveAuthorizedClientReactiveOAuth2AuthorizationFailureHandler(ReactiveOAuth2AuthorizedClientService authorizedClientService) {
+		this(authorizedClientService, DEFAULT_REMOVE_AUTHORIZED_CLIENT_ERROR_CODES);
+	}
+
+	/**
+	 * @param authorizedClientService the service from which authorized clients will be removed
+	 * 		  if the error code is one of the {@code removeAuthorizedClientErrorCodes}.
+	 * @param removeAuthorizedClientErrorCodes the OAuth2 Error Codes which will trigger removal of an authorized client.
+	 * @see OAuth2ErrorCodes
+	 */
+	public RemoveAuthorizedClientReactiveOAuth2AuthorizationFailureHandler(
+			ReactiveOAuth2AuthorizedClientService authorizedClientService,
+			Set<String> removeAuthorizedClientErrorCodes) {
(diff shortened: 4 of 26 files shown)


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/AbstractWebClientReactiveOAuth2AccessTokenResponseClient.java
```diff
@@ -0,0 +1,229 @@
...
+ * @author Phil Clay
+ * @since 5.3
+ * @param <T> type of grant request
+ * @see <a href="https://tools.ietf.org/html/rfc6749#section-3.2">RFC-6749 Token Endpoint</a>
+ * @see WebClientReactiveAuthorizationCodeTokenResponseClient
+ * @see WebClientReactiveClientCredentialsTokenResponseClient
+ * @see WebClientReactivePasswordTokenResponseClient
+ * @see WebClientReactiveRefreshTokenTokenResponseClient
+ */
+abstract class AbstractWebClientReactiveOAuth2AccessTokenResponseClient<T extends AbstractOAuth2AuthorizationGrantRequest>
```
philsttr: The introduction of this base class deduped a lot of stuff that was in the various `*WebClientReactiveOAuth2AccessTokenResponseClient` implementations.

It also fixed some response processing discrepancies/problems in the implementations.

For example, 
* Password, RefreshToken, and ClientCredentials previously checked HTTP Status code, but AuthorizationCode did not
* Password, RefreshToken, and ClientCredentials did not parse the authorization server response for a non 200 HTTP Status code (therefore, the oauth2 error code returned from the authorization server was not captured.)

In addition, this changes the exception type thrown to be `ClientAuthenticationException` (which extends `OAuth2AuthorizationException`) instead of `OAuth2AuthorizationException`

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -144,6 +187,31 @@ private static ReactiveOAuth2AuthorizedClientManager createDefaultAuthorizedClie
...
+		} else if (authorizedClientManager instanceof AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager) {
+			return new DefaultResponseHandler(
+					((AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager) authorizedClientManager)::getAuthorizationFailureHandler);
+		} else if (authorizedClientManager instanceof UnAuthenticatedReactiveOAuth2AuthorizedClientManager) {
+			return new DefaultResponseHandler(
+					((UnAuthenticatedReactiveOAuth2AuthorizedClientManager) authorizedClientManager)::getAuthorizationFailureHandler);
+		} else {
+			return (request, response) -> response;
+		}
+	}
```
philsttr: I'm not terribly happy with this.  It's quite ugly.  Suggestions welcome.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -330,64 +403,61 @@ public void setAccessTokenExpiresSkew(Duration accessTokenExpiresSkew) {
...
-				.filter(t -> this.defaultOAuth2AuthorizedClient && t instanceof OAuth2AuthenticationToken)
-				.cast(OAuth2AuthenticationToken.class)
-				.map(OAuth2AuthenticationToken::getAuthorizedClientRegistrationId);
-	}
-
-	private Mono<ServerWebExchange> currentServerWebExchange() {
-		return Mono.subscriberContext()
-				.filter(c -> c.hasKey(ServerWebExchange.class))
-				.map(c -> c.get(ServerWebExchange.class));
-	}
```
philsttr: Replaced with fields above.  No need to recreate these publishers every time.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunctionITests.java
```diff
@@ -0,0 +1,330 @@
...
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static org.springframework.security.oauth2.client.web.reactive.function.client.ServletOAuth2AuthorizedClientExchangeFilterFunction.clientRegistrationId;
+
+public class ServerOAuth2AuthorizedClientExchangeFilterFunctionITests {
```
philsttr: Contains equivalent tests as `ServletOAuth2AuthorizationClientExchangeFilterFunctionITests` _plus_ some additional tests related to the new failure handler.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2AuthorizationException.java
```diff
@@ -30,7 +32,18 @@
...
+		this(error, error.toString());
+	}
+
+	/**
+	 * Constructs an {@code OAuth2AuthorizationException} using the provided parameters.
+	 *
+	 * @param error the {@link OAuth2Error OAuth 2.0 Error}
+	 * @param message the exception message
+	 */
+	public OAuth2AuthorizationException(OAuth2Error error, String message) {
```
philsttr: Added constructors that take a message, because I wanted the ability to explicitly specify a message that provides more context, without changing the message that was returned in a response from an authorization/resource server.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/web/reactive/function/OAuth2AccessTokenResponseBodyExtractor.java
```diff
@@ -58,6 +58,10 @@
 		ParameterizedTypeReference<Map<String, Object>> type = new ParameterizedTypeReference<Map<String, Object>>() {};
 		BodyExtractor<Mono<Map<String, Object>>, ReactiveHttpInputMessage> delegate = BodyExtractors.toMono(type);
 		return delegate.extract(inputMessage, context)
+				.onErrorMap(e -> new OAuth2AuthorizationException(
+						invalidTokenResponse("An error occurred parsing the Access Token response: " + e.getMessage()), e))
+				.switchIfEmpty(Mono.error(() -> new OAuth2AuthorizationException(
+						invalidTokenResponse("Empty OAuth 2.0 Access Token Response"))))
```
philsttr: These were a couple use cases that were not handled previously that I encountered during testing.

It also wraps things like `DecodingException` in an `OAuth2AuthorizationException`.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/ReactiveOAuth2AuthorizationSuccessHandler.java
```diff
@@ -0,0 +1,41 @@
...
+
+	/**
+	 * Called when an OAuth 2.0 Authorized Client
+	 * has been successfully authorized (or re-authorized)
+	 * via the authorization server.
+	 *
+	 * @param authorizationSuccessEvent details of the successful authorization event that occurred
+	 * @return an empty {@link Mono} that completes after this handler has finished handling the event.
+	 */
+	Mono<Void> onAuthorizationSuccess(OAuth2AuthorizationSuccessEvent authorizationSuccessEvent);
```
jgrandja: As I mentioned during the design iteration, I wasn't convinced with the `*Event` naming. Event objects are used in a pub-sub interaction model, which is not the case here. The success/failure handlers are meant to be direct collaborators of the manager and therefore are coupled in the design. I just get a feel that using `*Event` in the class name and `*notify*` in internal methods will cause some confusion with the API. 

The simple option is to change the name. I can't think of a better name at the moment.

The other option is we flatten the arguments. For example:

```java
public interface ReactiveOAuth2AuthorizationSuccessHandler {

	Mono<Void> onAuthorizationSuccess(OAuth2AuthorizedClient authorizedClient, Authentication principal, Map<String, Object> attributes);

	default Mono<Void> onAuthorizationSuccess(OAuth2AuthorizedClient authorizedClient, Authentication principal) {
		return onAuthorizationSuccess(authorizedClient, principal, Collections.emptyMap());
	}
}
```

What are your thoughts?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/ReactiveOAuth2AuthorizationFailureHandler.java
```diff
@@ -0,0 +1,41 @@
...
+
+	/**
+	 * Called when an OAuth 2.0 Authorized Client
+	 * fails to authorize (or re-authorize)
+	 * via the authorization server or resource server.
+	 *
+	 * @param authorizationFailureEvent details of the failed authorization event that occurred
+	 * @return an empty {@link Mono} that completes after this handler has finished handling the event.
+	 */
+	Mono<Void> onAuthorizationFailure(OAuth2AuthorizationFailureEvent authorizationFailureEvent);
```
jgrandja: Same comments apply from `ReactiveOAuth2AuthorizationSuccessHandler`

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager.java
```diff
@@ -121,6 +198,58 @@ public void setContextAttributesMapper(Function<OAuth2AuthorizeRequest, Mono<Map
...
 	}
 
+	/**
+	 * Gets the handler that handles successful authorizations.
+	 *
+	 * @return the handler that handles successful authorizations.
+	 * @see DefaultAuthorizationSuccessHandler
+	 * @since 5.3
+	 */
+	public ReactiveOAuth2AuthorizationSuccessHandler getAuthorizationSuccessHandler() {
```
jgrandja: We typically don't expose a getter for an optional collaborator - only the setter. However, we should javadoc the default collaborator.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager.java
```diff
@@ -121,6 +198,58 @@ public void setContextAttributesMapper(Function<OAuth2AuthorizeRequest, Mono<Map
...
+	}
+
+	/**
+	 * Gets the handler that handles authorization failures.
+	 *
+	 * @return the handler that handles authorization failures.
+	 * @see DefaultAuthorizationFailureHandler
+	 * @since 5.3
+	 */
+	public ReactiveOAuth2AuthorizationFailureHandler getAuthorizationFailureHandler() {
```
jgrandja: We typically don't expose a getter for an optional collaborator - only the setter. However, we should javadoc the default collaborator.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager.java
```diff
@@ -134,4 +263,106 @@ public void setContextAttributesMapper(Function<OAuth2AuthorizeRequest, Mono<Map
...
 	}
+
+	/**
+	 * The default authorization success handler,
+	 * which will save the authorized client in a
+	 * {@link ReactiveOAuth2AuthorizedClientService}.
+	 *
+	 * @since 5.3
+	 */
+	public static class DefaultAuthorizationSuccessHandler implements ReactiveOAuth2AuthorizationSuccessHandler {
```
jgrandja: I'd rather make this `private` or implement via a `private` helper method. The implementation is quite minimal so code reuse wouldn't make much of a difference here. I'm ok with minimal code duplication vs. exposing a public API. Easier from a maintenance standpoint on our end and allows us to change things more easily.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager.java
```diff
@@ -134,4 +263,106 @@ public void setContextAttributesMapper(Function<OAuth2AuthorizeRequest, Mono<Map
...
+
+	/**
+	 * The default authorization failure handler,
+	 * which will remove the authorized client from a
+	 * {@link ReactiveOAuth2AuthorizedClientService}
+	 * for specific OAuth 2.0 error codes.
+	 *
+	 * @since 5.3
+	 */
+	public static class DefaultAuthorizationFailureHandler implements ReactiveOAuth2AuthorizationFailureHandler {
```
jgrandja: Same comments apply from `DefaultAuthorizationSuccessHandler `

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultReactiveOAuth2AuthorizedClientManager.java
```diff
@@ -167,6 +263,58 @@ public void setContextAttributesMapper(Function<OAuth2AuthorizeRequest, Mono<Map
...
 	}
 
+	/**
+	 * Gets the handler that handles successful authorizations.
+	 *
+	 * @return the handler that handles successful authorizations.
+	 * @see DefaultAuthorizationSuccessHandler
+	 * @since 5.3
+	 */
+	public ReactiveOAuth2AuthorizationSuccessHandler getAuthorizationSuccessHandler() {
```
jgrandja: We typically don't expose a getter for an optional collaborator - only the setter. However, we should javadoc the default collaborator.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultReactiveOAuth2AuthorizedClientManager.java
```diff
@@ -167,6 +263,58 @@ public void setContextAttributesMapper(Function<OAuth2AuthorizeRequest, Mono<Map
...
+	}
+
+	/**
+	 * Gets the handler that handles authorization failures.
+	 *
+	 * @return the handler that handles authorization failures.
+	 * @see DefaultAuthorizationFailureHandler
+	 * @since 5.3
+	 */
+	public ReactiveOAuth2AuthorizationFailureHandler getAuthorizationFailureHandler() {
```
jgrandja: We typically don't expose a getter for an optional collaborator - only the setter. However, we should javadoc the default collaborator.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultReactiveOAuth2AuthorizedClientManager.java
```diff
@@ -190,4 +338,128 @@ public void setContextAttributesMapper(Function<OAuth2AuthorizeRequest, Mono<Map
...
 	}
+
+	/**
+	 * The default authorization success handler,
+	 * which will save the authorized client in a
+	 * {@link ServerOAuth2AuthorizedClientRepository}.
+	 *
+	 * @since 5.3
+	 */
+	public static class DefaultAuthorizationSuccessHandler implements ReactiveOAuth2AuthorizationSuccessHandler {
```
jgrandja: I'd rather make this `private` or implement via a `private` helper method. The implementation is quite minimal so code reuse wouldn't make much of a difference here. I'm ok with minimal code duplication vs. exposing a public API. Easier from a maintenance standpoint on our end and allows us to change things more easily.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultReactiveOAuth2AuthorizedClientManager.java
```diff
@@ -190,4 +338,128 @@ public void setContextAttributesMapper(Function<OAuth2AuthorizeRequest, Mono<Map
...
+
+	/**
+	 * The default authorization failure handler,
+	 * which will remove the authorized client from a
+	 * {@link ServerOAuth2AuthorizedClientRepository}
+	 * for specific OAuth 2.0 error codes.
+	 *
+	 * @since 5.3
+	 */
+	public static class DefaultAuthorizationFailureHandler implements ReactiveOAuth2AuthorizationFailureHandler {
```
jgrandja: Same comments apply from `DefaultAuthorizationSuccessHandler `

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -101,6 +143,7 @@
 	public ServerOAuth2AuthorizedClientExchangeFilterFunction(ReactiveOAuth2AuthorizedClientManager authorizedClientManager) {
 		Assert.notNull(authorizedClientManager, "authorizedClientManager cannot be null");
 		this.authorizedClientManager = authorizedClientManager;
+		this.responseHandler = createDefaultResponseHandler(authorizedClientManager);
```
jgrandja: I'm not sure this makes sense here. If the application supplies `ReactiveOAuth2AuthorizedClientManager` then it's responsible for configuring all it's collaborator's. I think we only need to set the default handlers when constructing the default manager in `createDefaultAuthorizedClientManager()`.

I see that you were trying to reuse the manager handlers in `createDefaultResponseHandler()`. Given that I commented on removing the getters for the handlers, this won't be possible. As mentioned in another comment, the implementation is quite minimal so code reuse wouldn't make much of a difference here. I'm ok with minimal code duplication vs. exposing a public API. Easier from a maintenance standpoint on our end and allows us to change things more easily.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -430,15 +503,231 @@ private UnAuthenticatedReactiveOAuth2AuthorizedClientManager(
...
+							principal))
+					.then(Mono.error(authorizationException));
+		}
+
+		/**
+		 * Gets the handler that handles successful authorizations.
+		 * @return the handler that handles successful authorizations.
+		 * @see DefaultReactiveOAuth2AuthorizedClientManager.DefaultAuthorizationSuccessHandler
+		 */
+		public ReactiveOAuth2AuthorizationSuccessHandler getAuthorizationSuccessHandler() {
```
jgrandja: Please remove as per previous comment. As well, this class is `private`.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -430,15 +503,231 @@ private UnAuthenticatedReactiveOAuth2AuthorizedClientManager(
...
+			return authorizationSuccessHandler;
+		}
+
+		/**
+		 * Gets the handler that handles authorization failures.
+		 * @return the handler that handles authorization failures.
+		 * @see DefaultReactiveOAuth2AuthorizedClientManager.DefaultAuthorizationFailureHandler
+		 */
+		public ReactiveOAuth2AuthorizationFailureHandler getAuthorizationFailureHandler() {
```
jgrandja: Please remove as per previous comment. As well, this class is `private`.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/ReactiveOAuth2AuthorizationSuccessHandler.java
philsttr: I can't think of a better name for the class.  I'm fine with flattening the args as you have proposed.  I'll change it and see what it looks like.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
philsttr: I'm trying to make the basic functionality of removing cached clients on HTTP 401 from a resource server easy to use out of the box.

If all the default success/failure handler classes (which provide the basic functionality) are made private, and the getters are removed from the client managers, then it's going to be fairly painful for an application that uses the `ServerOAuth2AuthorizedClientExchangeFilterFunction(ReactiveOAuth2AuthorizedClientManager authorizedClientManager)` constructor to remove cached clients.

Right now, the `ServerOAuth2AuthorizedClientExchangeFilterFunction`'s response handler (which delegates to the client manager's failure handler to perform cached client removal) is private.  We would need to make a public interface for this.  This is not terrible, but I was trying to keep this all private.

An app that wanted to use the `ServerOAuth2AuthorizedClientExchangeFilterFunction(ReactiveOAuth2AuthorizedClientManager authorizedClientManager)` constructor, and remove cached clients on HTTP 401 responses from the resource server would need to:
* Implement a `ReactiveOAuth2AuthorizationFailureHandler` that is a copy/paste of `AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager.DefaultAuthorizationFailureHandler` or `DefaultReactiveOAuth2AuthorizedClientManager.DefaultAuthorizationFailureHandler` 
* Create the `ReactiveOAuth2AuthorizedClientManager` instance, and set its failure handler to the custom implementation
* Implement a copy/paste of `ServerOAuth2AuthorizedClientExchangeFilterFunction.DefaultResponseHandler` that delegates to the `ReactiveOAuth2AuthorizationFailureHandler`
* Create the `ServerOAuth2AuthorizedClientExchangeFilterFunction` instance, and set its response handler

All this is very painful, especially for just enabling basic functionality.

How do you suggest making the removal of cached clients on HTTP 401 from a resource server easy to use by an app that uses the `ServerOAuth2AuthorizedClientExchangeFilterFunction(ReactiveOAuth2AuthorizedClientManager authorizedClientManager)` constructor?  (maybe I'm missing something)

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/ReactiveOAuth2AuthorizationSuccessHandler.java
jgrandja: @jzheaux @rwinch Can you review this [comment](https://github.com/spring-projects/spring-security/pull/7756#discussion_r363483374) and suggest the approach you are in favour of? Either rename the input to the handler operations or flatten the arguments. I'm kind of divided with this one. I think I prefer the one input object as it will make it easier to add a new attribute (later on) without changing the interface.

jzheaux: @jgrandja Agreed that we don't want to use `*Event` since it could easily be confused with `ApplicationEventPublisher`, `AuthenticationEventPublisher`, et al. 

Given that `AuthenticationSuccessHandler` looks like this:

```
void onAuthenticationSuccess(HttpServletRequest request,
			HttpServletResponse response, Authentication authentication)
```

And `AuthenticationFailureHandler` looks like this:

```java
void onAuthenticationFailure(HttpServletRequest request,
			HttpServletResponse response, AuthenticationException exception)
```

And that this pattern is continued in the reactive stack, I'd favor modeling these other Spring Security success/failure handlers more closely (as you already said "flatten the arguments").

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
jgrandja: @philsttr 

I've been giving this some further thought and I agree that we should make it easier for users to leverage an existing basic (default) implementation.

So let's say we provide 2 implementations: `SaveAuthorizedClientReactiveOAuth2AuthorizationSuccessHandler` and `RemoveAuthorizedClientReactiveOAuth2AuthorizationFailureHandler`.

Both would be public outer classes and therefore could be used by any requiring component. For example, `DefaultReactiveOAuth2AuthorizedClientManager` and `AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager` would instantiate both as their default handlers. A setter would be exposed for each handler but no getter. We could apply the same changes to `ServerOAuth2AuthorizedClientExchangeFilterFunction` as well.

For `SaveAuthorizedClientReactiveOAuth2AuthorizationSuccessHandler`, we could provide 2 constructors - one that accepts `ServerOAuth2AuthorizedClientRepository` and the other that accepts `ReactiveOAuth2AuthorizedClientService`.

What are your thoughts?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/ReactiveOAuth2AuthorizationSuccessHandler.java
philsttr: Flattened the args as requested.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager.java
philsttr: Removed

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager.java
philsttr: Removed

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager.java
philsttr: Pulled out into public outer class as per other discussion in ServerOAuth2AuthorizedClientExchangeFilterFunction

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultReactiveOAuth2AuthorizedClientManager.java
philsttr: Pulled out into public outer class as per other discussion in ServerOAuth2AuthorizedClientExchangeFilterFunction

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultReactiveOAuth2AuthorizedClientManager.java
philsttr: Pulled out into public outer class as per other discussion in ServerOAuth2AuthorizedClientExchangeFilterFunction

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
philsttr: Removed

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
philsttr: Removed

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
philsttr: I've updated the PR with these changes.

I still don't quite like that the `ServerOAuth2AuthorizedClientExchangeFilterFunction(ReactiveOAuth2AuthorizedClientManager authorizedClientManager)` doesn't automatically handle removal of the authorized client on auth failure, without additional configuration of calling `setAuthorizationFailureHandler`, but I can live with this, given the constraints of keeping the getters private.

I've document this limitation in the javadocs.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
philsttr: Resolved by not automatically supporting removal of authorized client on auth failure when this constructor is used.  Requires user to call setAuthorizationFailureHandler to enable this functionality.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultReactiveOAuth2AuthorizedClientManager.java
philsttr: Removed

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultReactiveOAuth2AuthorizedClientManager.java
philsttr: Removed

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2AuthorizationException.java
```diff
@@ -30,7 +32,18 @@
...
-		super(error.toString());
+		this(error, error.toString());
+	}
+
+	/**
+	 * Constructs an {@code OAuth2AuthorizationException} using the provided parameters.
+	 *
+	 * @param error the {@link OAuth2Error OAuth 2.0 Error}
+	 * @param message the exception message
+	 */
```
jgrandja: Please add `@since 5.3`

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2AuthorizationException.java
```diff
@@ -41,7 +54,18 @@ public OAuth2AuthorizationException(OAuth2Error error) {
...
-		super(error.toString(), cause);
+		this(error, error.toString(), cause);
+	}
+
+	/**
+	 * Constructs an {@code OAuth2AuthorizationException} using the provided parameters.
+	 *
+	 * @param error the {@link OAuth2Error OAuth 2.0 Error}
+	 * @param cause the root cause
+	 */
```
jgrandja: Please add @since 5.3

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RemoveAuthorizedClientReactiveOAuth2AuthorizationFailureHandler.java
```diff
@@ -0,0 +1,170 @@
...
+ * @author Phil Clay
+ * @since 5.3
+ */
+public class RemoveAuthorizedClientReactiveOAuth2AuthorizationFailureHandler implements ReactiveOAuth2AuthorizationFailureHandler {
+
+	/**
+	 * The default OAuth2 error codes that will trigger removal of the authorized client.
+	 * @see OAuth2ErrorCodes
+	 */
+	public static final Set<String>
```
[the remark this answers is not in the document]
philsttr: Hmm.  IMHO, I don't think the cached client should be removed for all failures.  That would be too aggressive, since not all OAuth2 error codes indicate that a token is invalid.

I'd prefer that the default implementation follows the spec as closely as possible, and only remove tokens for the OAuth2 error codes for which the spec says the token is invalid.

As a side note, these aren't HTTP statuses, they're OAuth2 error codes.  And they are the specific error codes that (according to the spec) indicate that a token is no longer valid, and a new token should be retrieved.  The only reason I included the ability to configure these error codes at all is because there might be some out-of-spec servers that return incorrect OAuth2 error codes for invalid tokens.  E.g. in https://github.com/spring-projects/spring-security/issues/7583#issuecomment-549086681, @janzyka mentioned that their corporate server returns "invalid_request" instead of "invalid_grant" when the token is no longer valid.

If more complex (out-of-spec) filtering is needed other than oauth2 error codes, then a user could create a new implementation of `ReactiveOAuth2AuthorizationFailureHandler`.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/AbstractWebClientReactiveOAuth2AccessTokenResponseClient.java
```diff
@@ -0,0 +1,229 @@
...
+		return body;
+	}
+
+	/**
+	 * Returns the scopes to include as a property in the token request.
+	 *
+	 * @param grantRequest the grant request
+	 * @return the scopes to include as a property in the token request.
+	 */
+	protected abstract Set<String> scopesToRequest(T grantRequest);
```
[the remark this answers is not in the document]
philsttr: Hmm.  I thought `scopesToRequest` was a bit more self-documenting, considering there is also a `defaultScopes` method.

But I can change it to scopes, and just rely on the javadoc.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/AbstractWebClientReactiveOAuth2AccessTokenResponseClient.java
```diff
@@ -0,0 +1,229 @@
...
+
+	/**
+	 * Populates the given {@link OAuth2AccessTokenResponse} with additional details
+	 * from the grant request.
+	 *
+	 * @param grantRequest the request for which the response was received.
+	 * @param tokenResponse the original token response
+	 * @return a token response optionally populated with additional details from the request.
+	 */
+	protected OAuth2AccessTokenResponse populateTokenResponse(T grantRequest, OAuth2AccessTokenResponse tokenResponse) {
```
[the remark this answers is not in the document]
philsttr: It's overridden by WebClientReactiveRefreshTokenTokenResponseClient, which has slightly different semantics.

As an alternative, I'm going to make this method, and the other protected methods use default visibility instead of protected, so they can't be utilized outside of the package

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager.java
```diff
@@ -90,15 +120,60 @@ public AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager(
...
+	 * or {@link #authorizationFailureHandler}, depending on the authorization result.
+	 *
+	 * @param authorizationContext the context to authorize
+	 * @param principal the principle to authorize
+	 * @return a {@link Mono} that emits the authorized client after the authorization attempt succeeds
+	 *         and the {@link #authorizationSuccessHandler} has completed,
+	 *         or completes with an exception after the authorization attempt fails
+	 *         and the {@link #authorizationFailureHandler} has completed
+	 */
+	private Mono<OAuth2AuthorizedClient> authorizeAndNotify(
```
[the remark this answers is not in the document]
philsttr: I thought authorizeAndNotify was a little more self-documenting, since there is already an authorize method.  But I'll change it.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -430,15 +541,209 @@ private UnAuthenticatedReactiveOAuth2AuthorizedClientManager(
...
+	 * @since 5.3
+	 */
+	private class AuthorizationFailureForwarder implements ClientResponseHandler {
+
+		/**
+		 * A map of HTTP Status Code to OAuth 2.0 Error codes for
+		 * HTTP status codes that should be interpreted as
+		 * authentication or authorization failures.
+		 */
+		private final Map<Integer, String> httpStatusToOAuth2ErrorCodeMap;
```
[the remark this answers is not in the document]
philsttr: Only HTTP 401 and HTTP 403 indicate authentication/authorization failures.  IMHO other HTTP status should not be forwarded to the authorization failure handler.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -330,80 +420,102 @@ public void setAccessTokenExpiresSkew(Duration accessTokenExpiresSkew) {
...
+	 * so that the same token is no longer used in future requests to the Resource Server.</p>
+	 *
+	 * <p>The failure handler used by default depends on which constructor was used
+	 * to construct this {@link ServerOAuth2AuthorizedClientExchangeFilterFunction}.
+	 * See the constructors for more details.</p>
+	 *
+	 * @param authorizationFailureHandler the handler that handles authentication and authorization failures.
+	 * @since 5.3
+	 */
+	public void setAuthorizationFailureHandler(ReactiveOAuth2AuthorizationFailureHandler authorizationFailureHandler) {
```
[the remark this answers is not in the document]
philsttr: IMHO, I don't think it makes sense to have an authorizationSuccessHandler here, and adding one might even cause confusion.

I say it might cause confusion, because someone might be tempted to set it to a `SaveAuthorizedClientReactiveOAuth2AuthorizationSuccessHandler`, which is not needed, and will just create contention within the backing authorized client repository/service.

I say it doesn't make sense, since it's almost impossible to tell if authorization was successful just by analyzing a response from a resource server.  For example, does an HTTP 400 error indicate that authorization was successful?  What about an HTTP 500 error?  Authorization might or might not have been successful when the resource server returns those status codes, depending on where in the resource server's stack the error occurred.

On the other side, it is immediately obvious when authorization fails, since the resource server will return 401 or 403.  So it makes sense to have a failure handler.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultReactiveOAuth2AuthorizedClientManager.java
```diff
@@ -70,57 +111,107 @@ public DefaultReactiveOAuth2AuthorizedClientManager(ReactiveClientRegistrationRe
...
+	 *
+	 * @param authorizationContext the context to authorize
+	 * @param principal the principle to authorize
+	 * @param serverWebExchange the currently active exchange
+	 * @return a {@link Mono} that emits the authorized client after the authorization attempt succeeds
+	 *         and the {@link #authorizationSuccessHandler} has completed,
+	 *         or completes with an exception after the authorization attempt fails
+	 *         and the {@link #authorizationFailureHandler} has completed
+	 */
+	private Mono<OAuth2AuthorizedClient> authorizeAndNotify(
```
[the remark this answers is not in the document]
philsttr: I thought authorizeAndNotify was a little more self-documenting, since there is already an authorize method. But I'll change it.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -430,15 +541,209 @@ private UnAuthenticatedReactiveOAuth2AuthorizedClientManager(
...
+		 * or {@link #authorizationFailureHandler}, depending on the authorization result.
+		 *
+		 * @param authorizationContext the context to authorize
+		 * @param principal the principle to authorize
+		 * @return a {@link Mono} that emits the authorized client after the authorization attempt succeeds
+		 *         and the {@link #authorizationSuccessHandler} has completed,
+		 *         or completes with an exception after the authorization attempt fails
+		 *         and the {@link #authorizationFailureHandler} has completed
+		 */
+		private Mono<OAuth2AuthorizedClient> authorizeAndNotify(
```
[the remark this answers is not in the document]
philsttr: I thought authorizeAndNotify was a little more self-documenting, since there is already an authorize method. But I'll change it.

philsttr:
Not sure what is causing that [ci failure](https://travis-ci.org/spring-projects/spring-security/builds/637720406#L771).  Help appreciated.

```
* What went wrong:
Execution failed for task ':spring-security-data:compileJava'.
> Could not resolve all files for configuration ':spring-security-data:detachedConfiguration8'.
   > Could not resolve org.jetbrains.kotlinx:kotlinx-coroutines-bom:1.3.3.
     Required by:
         project :spring-security-data
      > Cannot choose between the following variants of org.jetbrains.kotlinx:kotlinx-coroutines-bom:1.3.3:
          - enforcedRuntimeElements
          - runtimeElements
        All of them match the consumer attributes:
          - Variant 'enforcedRuntimeElements' capability org.jetbrains.kotlinx:kotlinx-coroutines-bom:1.3.3:
              - Unmatched attributes:
                  - Found org.gradle.category 'enforced-platform' but wasn't required.
                  - Found org.gradle.status 'release' but wasn't required.
                  - Found org.gradle.usage 'java-runtime' but wasn't required.
          - Variant 'runtimeElements' capability org.jetbrains.kotlinx:kotlinx-coroutines-bom:1.3.3:
              - Unmatched attributes:
                  - Found org.gradle.category 'platform' but wasn't required.
                  - Found org.gradle.status 'release' but wasn't required.
                  - Found org.gradle.usage 'java-runtime' but wasn't required.
```

jgrandja:
@philsttr The CI failure is not related to your updates. We're working on it.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RemoveAuthorizedClientReactiveOAuth2AuthorizationFailureHandler.java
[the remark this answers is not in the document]
jgrandja: Good points. We'll keep it.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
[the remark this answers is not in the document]
jgrandja: Agreed, having a `setAuthorizationSuccessHandler()` might cause confusion here. We'll leave it out.

jgrandja:
@philsttr Thanks again for all your great work on this feature. I applied some minor javadoc edits and also moved `SaveAuthorizedClientReactiveOAuth2AuthorizationSuccessHandler` and `RemoveAuthorizedClientReactiveOAuth2AuthorizationFailureHandler` to the `...client.web` package. This is now in master.

FYI, I logged #7840 for the Servlet implementation.

philsttr:
Thanks Joe!  And thanks for all the brainstorming and reviews.

Also, I'm unfortunately not going to be able to work on the servlet implementation.  I wish I could, but my company doesn't have an immediate need for the servlet impl (unlike the reactive impl), and I have more pressing stuff on my plate in the new year.

jgrandja:
@philsttr No worries Phil. I totally understand. I'll take care of it. Thanks again.

Issue #7716, opened 2019-12-10
Add and remove users from MapReactiveUserDetailsService


PyvesB:
### Summary

When using Spring's security with the Webflux framework, removing and adding users dynamically isn't straightforward.

### Actual Behavior

[MapReactiveUserDetailsService](https://github.com/spring-projects/spring-security/blob/master/core/src/main/java/org/springframework/security/core/userdetails/MapReactiveUserDetailsService.java) only has two public methods: find a user by name and update a user's password. These are defined by its two parent interfaces, [ReactiveUserDetailsService](https://github.com/spring-projects/spring-security/blob/master/core/src/main/java/org/springframework/security/core/userdetails/ReactiveUserDetailsService.java) and [ReactiveUserDetailsPasswordService](https://github.com/spring-projects/spring-security/blob/master/core/src/main/java/org/springframework/security/core/userdetails/ReactiveUserDetailsPasswordService.java). 

Unless you use the `MapReactiveUserDetailsService(Map<String, UserDetails> users)` constructor and keep a reference to the original map, there is no way to update users dynamically. This workaround seems quite brittle and won't work when using one of the class's two other constructors.

### Expected Behavior

`MapReactiveUserDetailsService` should provide methods to conveniently add and remove users at any point in time. Should these methods be defined in the `ReactiveUserDetailsService` interface itself?

### Version

Latest 5.2.1.RELEASE.

fhanik:
For feature parity it would be reasonable to expect a [`ReactiveUserDetailsManager`](https://github.com/spring-projects/spring-security/blob/b93528138e2b2f7ad34d10a040e6b7ac50fc587a/core/src/main/java/org/springframework/security/provisioning/UserDetailsManager.java#L28-L59) interface similar to its [servlet counterpart](https://github.com/spring-projects/spring-security/blob/b93528138e2b2f7ad34d10a040e6b7ac50fc587a/core/src/main/java/org/springframework/security/provisioning/UserDetailsManager.java#L28-L59) 

I agree that keeping a reference to the user map is not an ideal solution. One [should not](https://github.com/spring-projects/spring-security/blob/b93528138e2b2f7ad34d10a040e6b7ac50fc587a/core/src/main/java/org/springframework/security/provisioning/InMemoryUserDetailsManager.java#L60-L64) expect that the reference is the source of data inside the implementation, but rather the seed for the initial user set.

Would you be interested in contributing a PR?

PyvesB:
[...]

PR #7762, opened 2019-12-20
Implement methods to add and remove users from MapReactiveUserDetailsService

PyvesB:
This is a start on #7716.

@fhanik suggested having an interface similar to the servlet [UserDetailsManager](https://github.com/spring-projects/spring-security/blob/b93528138e2b2f7ad34d10a040e6b7ac50fc587a/core/src/main/java/org/springframework/security/provisioning/UserDetailsManager.java), therefore, I've implemented reactive versions of `createUser`, `updateUser`, `deleteUser`, `userExists` and `changePassword`, with behaviours closely aligned to what is done in [InMemoryUserDetailsManager](https://github.com/spring-projects/spring-security/blob/b93528138e2b2f7ad34d10a040e6b7ac50fc587a/core/src/main/java/org/springframework/security/provisioning/InMemoryUserDetailsManager.java).

There is still a pending discussion about which interface should define these methods (see https://github.com/spring-projects/spring-security/issues/7716#issuecomment-565707483), so for now they are just standalone methods without any Javadocs.


marcusdacoregio:
Hi @PyvesB, thanks for the PR.

You should follow what `UserDetailsManager` does and do the same creating a `ReactiveUserDetailsManager`.
The class `MapReactiveUserDetailsService` can implement `ReactiveUserDetailsManager` and define the methods as you have done.

core/src/main/java/org/springframework/security/core/userdetails/MapReactiveUserDetailsService.java
```diff
@@ -82,6 +91,62 @@ public MapReactiveUserDetailsService(Collection<UserDetails> users) {
...
+	}
+
+	public Mono<Void> deleteUser(String username) {
+		return Mono.just(username)
+				.doOnNext(this.users::remove)
+				.then();
+	}
+
+	public Mono<Boolean> userExists(String username) {
+		return Mono.just(username)
```
marcusdacoregio, line 117: You should use `getKey(username)` here

core/src/main/java/org/springframework/security/core/userdetails/MapReactiveUserDetailsService.java
```diff
@@ -82,6 +91,62 @@ public MapReactiveUserDetailsService(Collection<UserDetails> users) {
...
+	public Mono<Void> updateUser(UserDetails user) {
+		return userExists(user.getUsername())
+				.doOnNext(exists -> Assert.isTrue(exists, "user should exist"))
+				.thenReturn(getKey(user.getUsername()))
+				.doOnNext(key -> this.users.put(key, user))
+				.then();
+	}
+
+	public Mono<Void> deleteUser(String username) {
+		return Mono.just(username)
```
marcusdacoregio, line 111: You should use `getKey(username)` here

marcusdacoregio (changes requested):
Hi @PyvesB, thanks for the PR. I've left some feedback inline.

Could you please update your tests to use JUnit 5 and make sure that it follows the standard naming convention (`methodInTestWhenScenarioThenExpected`)?

About the interface, we can stay with the methods only inside this implementation for now, and if we have another implementation in the future that does the same thing, we must extract an interface and refactor those classes.

PyvesB:
Hello @marcusdacoregio, I no longer use spring-security in any of my current projects, but I've set up a dev environment again and addressed your comments. I wasn't expecting the review to come two years after someone from the maintainer team suggested contributing!

> Could you please update your tests to use JUnit 5 and make sure that it follows the standard naming convention (`methodInTestWhenScenarioThenExpected`)?

This should probably be documented in [the contributing guidelines](https://github.com/spring-projects/spring-security/blob/main/CONTRIBUTING.adoc#submit-junit-test-cases-for-all-behavior-changes), which currently only state that "New test methods should not start with test". Something that isn't documented, recommended by JUnit 5, nor even followed by half the existing tests in that file really can't be called a "standard convention". 😉

marcusdacoregio:
Hi @PyvesB. Thanks for taking the time to work on the suggestions. I'm sorry that your PR was inactive for so long.

Since you no longer use Spring Security, I am not sure if we should merge this PR now. I am aware that we have the same for the `Servlet` side, but it's not really a desire that Spring Security manages users. Also, I don't see this being a common requirement between Spring Security's users, if you can point me to issues where this is requested, it would help.

I'll close this for now but we can always reopen if the necessity is clear.

PR #7764, opened 2019-12-20
Set secure on cookie when logging out

okohub:
It is better to mark cookie secure flag with request.isSecure() to ensure cookie identity is same

### DIFF
--- a/web/src/test/java/org/springframework/security/web/authentication/logout/CookieClearingLogoutHandlerTests.java
+++ b/web/src/test/java/org/springframework/security/web/authentication/logout/CookieClearingLogoutHandlerTests.java
@@ -75,6 +75,18 @@ public class CookieClearingLogoutHandlerTests {
 	}
 
 	@Test
+	public void configuredCookieIsNotSecure() {
+		MockHttpServletResponse response = new MockHttpServletResponse();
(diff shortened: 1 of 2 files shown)


web/src/test/java/org/springframework/security/web/authentication/logout/CookieClearingLogoutHandlerTests.java
```diff
@@ -61,6 +62,18 @@ public void configuredCookiesAreCleared() {
...
+		MockHttpServletResponse response = new MockHttpServletResponse();
+		MockHttpServletRequest request = new MockHttpServletRequest();
+		request.setSecure(true);
+		request.setContextPath("/app");
+		CookieClearingLogoutHandler handler = new CookieClearingLogoutHandler("my_cookie");
+		handler.logout(request, response, mock(Authentication.class));
+		assertThat(response.getCookies()).hasSize(1);
+		assertThat(response.getCookies()[0].getSecure()).isTrue();
+	}
+
```
eleftherias: Please add a test checking that when the request is not secure, then the cookie is not secure.

okohub:
@eleftherias I added the test you requested, thanks.

eleftherias:
Thanks for the PR @okohub! This is now merged into master.

PR #7769, opened 2019-12-23
Fix return type of method AbstractConfiguredSecurityBuilder.objectPostProcessor()

Ammar0:
The return type of the method AbstractConfiguredSecurityBuilder.objectPostProcessor() is wrong.
As mentioned in the Javadoc, the method returns the SecurityBuilder for further customizations.

```
/**
 * @param <O> The object that this builder returns
 * @param <B> The type of this builder (that is returned by the base class)
 */
public abstract class AbstractConfiguredSecurityBuilder<O, B extends SecurityBuilder<O>>
		extends AbstractSecurityBuilder<O> {
}
```


jzheaux:
Good catch, this does appear to be a bug as the cast would fail:

```java
http
    .objectPostProcessor(processor)
        .getFilters() // ClassCastException
```

I've merged this as b4619f31ee791213260b30df5e4ef62de534c5a6. Note that I changed the commit description to align with [our contribution guidelines](https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.md#format-commit-messages). The description now also outlines the problem, since there is no accompanying issue to tie the commit to.

PR #7772, opened 2019-12-26
Fix var typo and code readability

rafaelrenanpacheco:


eleftherias:
Thanks for the PR @rafaelrenanpacheco! This is now merged into master.

Issue #5185, opened 2018-03-29
Provide XML namespace support for OAuth 2.0 Resource Server


jgrandja:
We should consider providing XML namespace support that is equivalent to `HttpSecurity.oauth2ResourceServer()`.

**NOTE:** Spring Security OAuth 2.x implements this via [ResourceServerBeanDefinitionParser](https://github.com/spring-projects/spring-security-oauth/blob/master/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/config/xml/ResourceServerBeanDefinitionParser.java). See the [sample xml](https://github.com/spring-projects/spring-security-oauth/blob/master/spring-security-oauth2/src/test/resources/org/springframework/security/oauth2/config/xml/resource-server-context.xml) config.

PR #7775, opened 2019-12-28
OAuth 2.0 Resource Server XML Support

jzheaux:
Fixes: gh-5185


config/src/main/java/org/springframework/security/config/http/OAuth2ResourceServerBeanDefinitionParser.java
```diff
@@ -0,0 +1,336 @@
...
+			return new RootBeanDefinition(DefaultBearerTokenResolver.class);
+		} else {
+			return new RuntimeBeanReference(bearerTokenResolverRef);
+		}
+	}
+}
+
+final class JwtBeanDefinitionParser implements BeanDefinitionParser {
+	static final String DECODER_REF = "decoder-ref";
+	static final String JWK_SET_URI = "jwk-set-uri";
```
jgrandja: Let's rename the attribute to `jwkset-uri` - NOTE: This is how it's named in `<client-registration>` element

config/src/main/java/org/springframework/security/config/http/OAuth2ResourceServerBeanDefinitionParser.java
```diff
@@ -0,0 +1,336 @@
...
+		return new JwtAuthenticationConverter();
+	}
+
+	JwtBeanDefinitionParser() {}
+}
+
+final class OpaqueTokenBeanDefinitionParser implements BeanDefinitionParser {
+	static final String INTROSPECTOR_REF = "introspector-ref";
+	static final String INTROSPECTION_URI = "introspection-uri";
+	static final String CLIENT_ID = "client-id";
```
jgrandja: I wonder if this should be renamed to `introspection-client-id` - similar for `client-secret`?

config/src/main/java/org/springframework/security/config/http/OAuth2ResourceServerBeanDefinitionParser.java
jzheaux: On the DSL, this is called `jwkSetUri`, and `jwkSetUri` appears to be the prevailing case pattern in variable and method names. The most natural translation I would think would be to place a dash wherever the case changes.

Based on this, I'd recommend changing `<client-registration>` to be `jwk-set-uri` so that it aligns with the DSL.

config/src/main/java/org/springframework/security/config/http/OAuth2ResourceServerBeanDefinitionParser.java
jzheaux: These match the names of the properties in Spring Boot, which is why I went with these names. Also, the repetition of the name feels just a little bit cumbersome when typing.

I suppose I'm open, though - can you elaborate on why you feel this is important?

config/src/main/java/org/springframework/security/config/http/OAuth2ResourceServerBeanDefinitionParser.java
jgrandja: That makes sense. I'll update to `jwk-set-uri`  for `<client-registration>`

config/src/main/java/org/springframework/security/config/http/OAuth2ResourceServerBeanDefinitionParser.java
jgrandja: I missed the fact that these properties are same as in Spring Boot. Let's keep as-is then.

config/src/main/resources/org/springframework/security/config/spring-security-5.3.rnc
```diff
@@ -572,6 +572,48 @@ provider.attlist &=
 	## The URI used to discover the configuration information for an OAuth 2.0 or OpenID Connect 1.0 Provider.
 	attribute issuer-uri {xsd:token}?
 
+oauth2-resource-server =
+	## Configures authentication support as an OAuth 2.0 Resource Server.
+	element oauth2-resource-server {oauth2-resource-server.attlist}
```
jgrandja: I'm confused how `<jwt>` and `<opaque-token>` are `<xs:choice>`'s in the `xsd`, however, both elements are not defined as choices in the `rnc` file? It looks like both those elements do not have `<oauth2-resource-server>` as their parent. Maybe I'm missing something?

It should be similar to `websocket-message-broker`:

```xml
websocket-message-broker =
	element websocket-message-broker { websocket-message-broker.attrlist, (intercept-message* & expression-handler?) }
```


```(intercept-message* & expression-handler?)``` are choices

PR #7784, opened 2020-01-03
Fixed desc of PasswordEncoder

artmiar:
It does not offer decoding passwords. Only encoding and matching encoded passwords


eleftherias:
Thanks for the PR @artmiar! This is now merged into master.

Issue #5558, opened 2018-07-22
Please support using Kotlin in a DSL


joshlong:
there are some exciting options for DSL designers in the Kotlin world. It'd be nice to support a Kotlin-DSL that does in a more idiomatic way what Spring Security does.

It would be able to take advantage of [nested builder DSLs, like this](https://github.com/spring-projects/spring-security/issues/5557)

rwinch:
Closing this in favor of #5557

metacubed:
@eleftherias and @joshlong, what do you think about simplifying this further?

```kotlin
override fun configure(http: HttpSecurity) {
    http {
        authorizeRequests {
            + "/path" to permitAll
            + "/user/{userName}" to "#userName == 'user'"
            + RegexRequestMatcher(".*", null) to authenticated
        }
    }
}
```

This option implements a `+` operator as an alternative to the `authorize` method.
This would only be executed at config/startup time, so the additional `Pair` creation would probably be ok.

eleftherias:
Thanks for the suggestion @metacubed.
I would refrain from using the `+` operator in this context, because it is already a well defined infix binary operator. 
Using it at the start of the line with only one parameter strays from its intended use-case and may confuse users.

eleftherias:
A native Kotlin DSL for servlet based applications is available as of Spring Security 5.3.
The native Kotlin DSL for WebFlux based applications will be part of Spring Security 5.4.
Spring Security 5.4.0.M1 is scheduled for 2020/05/06.

PR #7785, opened 2020-01-03
Idiomatic Kotlin DSL for configuring HTTP security

eleftherias:
Issue: gh-5558


eleftherias:
An example configuration is provided in [the Kotlin sample](https://github.com/spring-projects/spring-security/blob/master/samples/boot/kotlin/src/main/kotlin/org/springframework/security/samples/config/SecurityConfig.kt).

Issue #7795, opened 2020-01-08
Allowing to configure logo in default login page


vvtelu:
### Summary

With the current implementation - in Spring Security Web 5.2.1 - To make configurable logo in the default login page

### Expected Behavior

Showing the configurable logo in the default login page

### Configuration

Through REST API Call to get the logo details and image related attributes.

### Version

5.2.1 RELEASE

jzheaux:
Thanks for putting some thought into this, @vvtelu. However, it's not intended that this page be customizable. The default login page is meant to be a reasonable default. If you want to customize the page, then you can do so by [creating your own login page](https://docs.spring.io/spring-security/site/docs/current/guides/html5/form-javaconfig.html).

PR #7796, opened 2020-01-08
Allow logo in login page

vvtelu:
Allowing to configure logo in default login page #7795


jzheaux:
Thanks for putting some thought into this, @vvtelu. However, it's not intended that this page be customizable. The default login page is meant to be a reasonable default. If you want to customize the page, then you can do so by [creating your own login page](https://docs.spring.io/spring-security/site/docs/current/guides/html5/form-javaconfig.html).

Issue #7713, opened 2019-12-09
ProviderManager should have a varargs constructor


jzheaux:
`oauth2ResourceServer()` introduces a couple of `authenticationManager()` methods for complete control over the authentication mechanism:

```java
http
    .oauth2ResourceServer()
        .jwt()
            .authenticationManager(...)
```

This can be leveraged to construct a custom `JwtAuthenticationProvider`:

```java
AuthenticationProvider jwt = new JwtAuthenticationProvider(... ;

http
    .oauth2ResourceServer()
        .jwt()
            .authenticationManager(jwt::authenticate)
```

However, the method reference shortcut may seem a bit terse for some tastes.

As an alternative, code can construct an instance of `ProviderManager`:

```java
AuthenticationProvider jwt = new JwtAuthenticationProvider(... ;

http
    .oauth2ResourceServer()
        .jwt()
            .authenticationManager(new ProviderManager(Arrays.asList(jwt)))
```

This could be improved by adding a varargs constructor to `ProviderManager`:

```java
AuthenticationProvider jwt = new JwtAuthenticationProvider(... ;

http
    .oauth2ResourceServer()
        .jwt()
            .authenticationManager(new ProviderManager(jwt))
```

ThomasVitale:
Hi @jzheaux, I would like to work on this issue.

jzheaux:
@ThomasVitale it's yours

ThomasVitale:
@jzheaux do you have any recommendation to handle the ambiguity when passing `null` to the constructor? Would it be ok to force a `Collections.emptyList()` argument instead of `null` in case of no `AuthenticationProviders` passed to the `ProviderManager`? What about backward compatibility?

jzheaux:
@ThomasVitale good question

```java
new ProviderManager(null)
```

is not a supported use of that class, so I don't think there's much to worry about in the way of disambiguation. That is, if folks are doing that for some reason, they are on their own to make it work as they upgrade.

That said, we ought to continue to correctly error when null values are provided, so you might consider adding something like:

```java
if (providers.contains(null)) {
	throw new IllegalArgumentException(
			"providers cannot contain null values");
}
```
To the `checkState()` method.

> Would it be ok to force a Collections.emptyList()

Probably not. Spring Security prefers to throw an exception in misconfiguration scenarios instead of trying to leniently interpret the code's intent.

ThomasVitale:
@jzheaux thanks for your answer, it solved my doubts.

[...]

PR #7806, opened 2020-01-10
ProviderManager should have a varargs constructor

ThomasVitale:
- Added varargs constructor to ProviderManager.
- Updated ProviderManagerTests to use the new syntax and fix the check for empty list.

Fixes gh-7713

### DIFF
--- a/core/src/test/java/org/springframework/security/authentication/ProviderManagerTests.java
+++ b/core/src/test/java/org/springframework/security/authentication/ProviderManagerTests.java
@@ -35,11 +35,10 @@ import static org.mockito.Mockito.*;
  *
  * @author Ben Alex
  */
-@SuppressWarnings("unchecked")
 public class ProviderManagerTests {
 
 	@Test(expected = ProviderNotFoundException.class)
-	public void authenticationFailsWithUnsupportedToken() throws Exception {
+	public void authenticationFailsWithUnsupportedToken() {
 		Authentication token = new AbstractAuthenticationToken(null) {
 			public Object getCredentials() {
 				return "";
@@ -55,7 +54,7 @@ public class ProviderManagerTests {
 	}
 
 	@Test
-	public void credentialsAreClearedByDefault() throws Exception {
+	public void credentialsAreClearedByDefault() {
 		UsernamePasswordAuthenticationToken token = new UsernamePasswordAuthenticationToken(
 				"Test", "Password");
 		ProviderManager mgr = makeProviderManager();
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/springframework/security/authentication/ProviderManagerTests.java
```diff
@@ -96,7 +94,7 @@ public void authenticationSucceedsWhenFirstProviderReturnsNullButSecondAuthentic
 
 	@Test(expected = IllegalArgumentException.class)
 	public void testStartupFailsIfProvidersNotSet() {
-		new ProviderManager(null);
```
jzheaux: I think we should still have a test for `new ProviderManager((AuthenticationProvider) null)` and for `new ProviderManager((List<AuthenticationProvider>) null)`.

The reason is that an application might do:

```java
AuthenticationProvider provider = constructProvider();
return new ProviderManager(provider);
```

but `constructProvider` returns `null`.

I believe this also implies that the extra validation logic in `checkState()` will turn out to be necessary.

core/src/test/java/org/springframework/security/authentication/ProviderManagerTests.java
```diff
@@ -189,7 +186,7 @@ public void parentAuthenticationIsUsedIfProvidersDontAuthenticate() {
 		Authentication authReq = mock(Authentication.class);
 		when(parent.authenticate(authReq)).thenReturn(authReq);
 		ProviderManager mgr = new ProviderManager(
-				Arrays.asList(mock(AuthenticationProvider.class)), parent);
+				Collections.singletonList(mock(AuthenticationProvider.class)), parent);
```
jzheaux: Thanks for these cleanup items, too. Will you please put them in a separate commit (same PR)? That is, any code changes that aren't directly related to the new constructor, let's separate them into their own commit. This simplifies maintenance.

jzheaux:
@ThomasVitale Will you be able to apply the requested changes?

core/src/test/java/org/springframework/security/authentication/ProviderManagerTests.java
ThomasVitale: Got it, thanks for the clarification. I have added the extra validation login in `checkState()` and tests for both null scenarios.

core/src/test/java/org/springframework/security/authentication/ProviderManagerTests.java
ThomasVitale: Ok, now they are in their own commit.

ThomasVitale:
@jzheaux I have just delivered the requested changes, sorry for the delay. Does it look ok now?

jzheaux:
Nice, @ThomasVitale! This is now merged into `master`.

Also, I added a very slight polish via d22b476983d476b46e2ad97fcad7cd59f650ad5d

PR #7826, opened 2020-01-14
Improve Bearer Token Error Handling

jzheaux:
This PR improves how Resource Server handles Bearer Token errors. 

Specifically:

* It adds `InvalidBearerTokenException` to `DefaultAuthenticationEventPublisher`.
* It differentiates between JWT processing errors that are token-related vs configuration or service-related.

These are in the same PR because Resource Server cannot currently determine whether or not a JWT processing error is really an invalid token, and I'd like to address that before registering that exception with `DefaultAuthenticationEventPublisher`.

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationProvider.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationProvider.java
@@ -20,9 +20,11 @@ import java.util.Collection;
 import org.springframework.core.convert.converter.Converter;
 import org.springframework.security.authentication.AbstractAuthenticationToken;
 import org.springframework.security.authentication.AuthenticationProvider;
+import org.springframework.security.authentication.AuthenticationServiceException;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
 import org.springframework.security.core.GrantedAuthority;
+import org.springframework.security.oauth2.jwt.BadJwtException;
 import org.springframework.security.oauth2.jwt.Jwt;
 import org.springframework.security.oauth2.jwt.JwtDecoder;
 import org.springframework.security.oauth2.jwt.JwtException;
@@ -80,8 +82,10 @@ public final class JwtAuthenticationProvider implements AuthenticationProvider {
 		Jwt jwt;
 		try {
 			jwt = this.jwtDecoder.decode(bearer.getToken());
-		} catch (JwtException failed) {
+		} catch (BadJwtException failed) {
 			throw new InvalidBearerTokenException(failed.getMessage(), failed);
+		} catch (JwtException failed) {
+			throw new AuthenticationServiceException(failed.getMessage(), failed);
 		}
 
 		AbstractAuthenticationToken token = this.jwtAuthenticationConverter.convert(jwt);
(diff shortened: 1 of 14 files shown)


jzheaux:
`OAuth2AuthenticationException` is a generic exception that is thrown in numerous circumstances, including security misconfiguration scenarios. It doesn't seem like a good fit to tie to an event since you'd get alerted to problems that aren't related to bearer token authentication failures.

In the ticket, you said:

> I would like to log requests for invalid attempts for authenticating via OAuth 2.0 bearer tokens (via JWT)

By this, I interpret your intent to be `Authorization: Bearer` scenarios, which is what this PR addresses. Are there other scenarios you are looking to cover?

jzheaux:
@sdavids13 Can you describe other scenarios that you'd like to event on? This PR is certainly something that we can iterate on.

sdavids13:
Yes, I think this PR fits the bill with what I need/requested. I was merely concerned that someone might inadvertently use the `OAuth2AuthenticationException` when there is a legitimate authentication error and not get a corresponding audit event. Looking at the documentation of [OAuth2AuthenticationException](https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/oauth2/core/OAuth2AuthenticationException.html):

> This exception is thrown for all OAuth 2.0 related Authentication errors.
> There are a number of scenarios where an error may occur, for example:
> - The authorization request or token request is missing a required parameter
> - Missing or invalid client identifier
> - Invalid or mismatching redirection URI
> - The requested scope is invalid, unknown, or malformed
> - The resource owner or authorization server denied the access request
> - Client authentication failed
> - The provided authorization grant (authorization code, resource owner credentials) is invalid, expired, or revoked

I would think for all of those problems we would want to know if someone is attempting to authenticate but being denied for any one of those reasons.

As of right now, I think this PR results in a better out-of-box experience and addressed my primary issue, my only fear is that we aren't catching all events which we may want to audit. But, at the same time, we could merge this PR, and backlog a different ticket which may ask those other questions as to which other events should be audited.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationProvider.java
```diff
@@ -80,8 +82,10 @@ public Authentication authenticate(Authentication authentication) throws Authent
 		Jwt jwt;
 		try {
 			jwt = this.jwtDecoder.decode(bearer.getToken());
-		} catch (JwtException failed) {
+		} catch (BadJwtException failed) {
 			throw new InvalidBearerTokenException(failed.getMessage(), failed);
+		} catch (JwtException failed) {
+			throw new AuthenticationServiceException(failed.getMessage(), failed);
```
jgrandja: I'm wondering if this is correct? When `AuthenticationServiceException` is thrown, it typically is caught by an `AbstractAuthenticationProcessingFilter`, which `BearerTokenAuthenticationFilter` is not.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtReactiveAuthenticationManager.java
```diff
@@ -71,7 +73,11 @@ public void setJwtAuthenticationConverter(
 		this.jwtAuthenticationConverter = jwtAuthenticationConverter;
 	}
 
-	private OAuth2AuthenticationException onError(JwtException e) {
-		return new InvalidBearerTokenException(e.getMessage(), e);
+	private AuthenticationException onError(JwtException e) {
+		if (e instanceof BadJwtException) {
+			return new InvalidBearerTokenException(e.getMessage(), e);
+		} else {
+			return new AuthenticationServiceException(e.getMessage(), e);
```
jgrandja: Same as previous comment

jzheaux:
Merged to `master` via fbdecdafb88140d3adce63bf3b3f2e6bb091c222

Issue #7655, opened 2019-11-18
Provide JDBC implementation of OAuth2AuthorizedClientService


jgrandja:
We should provide a JDBC implementation of `OAuth2AuthorizedClientService`.

Related #7889

PR #7855, opened 2020-01-22
Add JDBC implementation of OAuth2AuthorizedClientService

jgrandja:
Fixes gh-7655


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/JdbcOAuth2AuthorizedClientService.java
```diff
@@ -0,0 +1,268 @@
...
+ * </pre>
+ *
+ * @author Joe Grandja
+ * @since 5.3
+ * @see OAuth2AuthorizedClientService
+ * @see OAuth2AuthorizedClient
+ * @see JdbcTemplate
+ */
+@Repository
+public class JdbcOAuth2AuthorizedClientService implements OAuth2AuthorizedClientService {
```
jgrandja: @rwinch FYI, I have considered introducing `OAuth2AuthorizedClientDao` as it seems it would be a logical design choice, however, I felt that introducing a new interface in this instance would complicate the design and seemed a bit too much. 

If we did introduce `OAuth2AuthorizedClientDao` with the initial `JdbcOAuth2AuthorizedClientDao` implementation, what would the `OAuth2AuthorizedClientService` look like? It wouldn't do much more than delegate to the Dao. What would the name be? `DelegatingOAuth2AuthorizedClientService` would be too generic of a name since we could have another type of delegate that would delegate to another type of backing implementation other than a Dao. So then the name would have to be something like `DaoDelegatingOAuth2AuthorizedClientService`, which is still not ideal since we have `Dao` and `Service` in the name.

Now let's say we did introduce `OAuth2AuthorizedClientDao`, would we have another implementation? Possibly an O/R mapping implementation like Hibernate? I'm not convinced it's needed since we now will have JDBC. Redis is definitely a likely candidate for the next implementation but would that implement `OAuth2AuthorizedClientDao`? I don't think it's the right fit. Also, with the combination of `HttpSessionOAuth2AuthorizedClientRepository` and Spring Session backed by Redis, a user could have `OAuth2AuthorizedClient` stored in Redis today.

What are your thoughts?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/JdbcOAuth2AuthorizedClientService.java
```diff
@@ -0,0 +1,268 @@
...
+	public void saveAuthorizedClient(OAuth2AuthorizedClient authorizedClient, Authentication principal) {
+		Assert.notNull(authorizedClient, "authorizedClient cannot be null");
+		Assert.notNull(principal, "principal cannot be null");
+
+		OAuth2AuthorizedClientData authorizedClientData = mapData(authorizedClient, principal);
+		Object[] args = {
+				authorizedClientData.clientRegistrationId,
+				authorizedClientData.principalName,
+				authorizedClientData.accessTokenType,
+				authorizedClientData.accessTokenValue,		// TODO Encrypt
```
jgrandja: @rwinch Do you have a preference/recommendation on the approach for encrypting/securing the access token and refresh token?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/JdbcOAuth2AuthorizedClientService.java
```diff
@@ -0,0 +1,268 @@
...
+	protected final ClientRegistrationRepository clientRegistrationRepository;
+
+	/**
+	 * Constructs a {@code JdbcOAuth2AuthorizedClientService} using the provided parameters.
+	 *
+	 * @param dataSource the data source
+	 * @param clientRegistrationRepository the repository of client registrations
+	 */
+	public JdbcOAuth2AuthorizedClientService(
+			DataSource dataSource, ClientRegistrationRepository clientRegistrationRepository) {
```
rwinch: I think it would be good to allow injecting `JdbcOperations` instead.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/JdbcOAuth2AuthorizedClientService.java
```diff
@@ -0,0 +1,268 @@
...
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.jdbc.core.PreparedStatementSetter;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.oauth2.client.registration.ClientRegistration;
+import org.springframework.security.oauth2.client.registration.ClientRegistrationRepository;
+import org.springframework.security.oauth2.core.OAuth2AccessToken;
+import org.springframework.security.oauth2.core.OAuth2RefreshToken;
+import org.springframework.stereotype.Repository;
+import org.springframework.transaction.annotation.EnableTransactionManagement;
+import org.springframework.transaction.annotation.Transactional;
```
rwinch: We don't use `@Transactional` in Spring Security because that means users must have declarative transactions enabled and we have no way to enforce it (nor should we need to). `JdbcTemplate` will handle the transactions for you.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/JdbcOAuth2AuthorizedClientService.java
rwinch: I agree. There isn't a need for a Dao

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/JdbcOAuth2AuthorizedClientService.java
rwinch: I think we should externalize the mapping (reading/writing) of the data into one of Spring's strategies. Then user's can inject their own strategies.

oauth2/oauth2-client/src/test/resources/oauth2-client-schema.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE oauth2_authorized_client (
```
rwinch: Can we publish this on the classpath similar to how we do with the users.ddl? We could also update the Javadoc of the service to mention the classpath resource.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/JdbcOAuth2AuthorizedClientService.java
jgrandja: Which specific strategy were you thinking? `Converter` or you had another one in mind?

jgrandja:
@rwinch 

> I cannot help to think that we might want to think about discriminating based on other attributes...i.e. the scopes associated with the token

The only "finder" operation we have in `OAuth2AuthorizedClientService` is:

`loadAuthorizedClient(String clientRegistrationId, String principalName)`

so this wouldn't work for searching based on `scopes`. We would either need to add another operation or possibly create another API that provides various "finder" operations. Is there a specific use case that you are trying to address that has been reported by the community?

It might make sense to log a ticket with the details you had in mind so we can address this separately?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/JdbcOAuth2AuthorizedClientService.java
```diff
@@ -0,0 +1,516 @@
...
+		 */
+		public Authentication getPrincipal() {
+			return this.principal;
+		}
+	}
+
+	/**
+	 * The data (entity) representation of an {@link OAuth2AuthorizedClient}.
+	 */
+	public static class OAuth2AuthorizedClientData {
```
rwinch: What is the need for this type vs allowing the user to inject a `RowMapper` that produces an `OAuth2AuthorizedClient`?

jgrandja: @rwinch The `OAuth2AuthorizedClientData` is an entity representation that IMO is simpler to deal with compared to returning an `OAuth2AuthorizedClient`. The logic in `OAuth2AuthorizedClientDataConverter` is a bit involved and would require a bit too much work for the user to implement. Instead they can simply reuse `OAuth2AuthorizedClientDataConverter` and implement a `RowMapper<OAuth2AuthorizedClientData>` that performs some data translation/conversion (if necessary), eg. decoding/decrypting the access/refresh token.

jgrandja: Take a look at the test `JdbcOAuth2AuthorizedClientServiceTests.tableDefinitionWhenCustomThenAbleToOverride`. It demonstrates what the user would need to implement if they preferred to change the table definition from the standard definition it expects. The `RowMapper<OAuth2AuthorizedClientData>` needs to be implemented but both `Converter`'s are reused.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/JdbcOAuth2AuthorizedClientService.java
```diff
@@ -0,0 +1,514 @@
...
+		Assert.hasText(principalName, "principalName cannot be empty");
+
+		Object[] args = {clientRegistrationId, principalName};
+		int[] argTypes = {Types.VARCHAR, Types.VARCHAR};
+		PreparedStatementSetter pss = new ArgumentTypePreparedStatementSetter(args, argTypes);
+
+		List<OAuth2AuthorizedClientData> result = this.jdbcOperations.query(
+				LOAD_AUTHORIZED_CLIENT_SQL, pss, this.authorizedClientDataRowMapper);
+
+		return !result.isEmpty() ? (T) this.authorizedClientDataConverter.convert(result.get(0)) : null;
```
rwinch: Since we are only looking for 0 or 1 results, it seems like queryForObject might make more sense here.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/JdbcOAuth2AuthorizedClientService.java
rwinch: I don't like having this additional type. If `OAuth2AuthorizedClient` is difficult to use, we should just make that easier vs creating another type.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/JdbcOAuth2AuthorizedClientService.java
jgrandja: I used `queryForObject()` originally but it throws an `IncorrectResultSizeDataAccessException` if not found, which is not ideal. With `List query()`, I can avoid the `try catch` and exception being thrown.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/JdbcOAuth2AuthorizedClientService.java
jgrandja: I removed `OAuth2AuthorizedClientData`

oauth2/oauth2-client/src/main/resources/org/springframework/security/oauth2/client/oauth2-client-schema.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE oauth2_authorized_client (
```
rwinch: I would not be shy about making values larger rather than shorter. Storage space is typically not an issue these days. I'd consider making the token value and the refresh token value's both blobs since we should not ever need to search by them.

jgrandja: I made both token columns `blob` and increased the size for a couple of other columns too. This is now merged.

jgrandja:
Merged via de8b558

PR #7875, opened 2020-01-30
Enabled custom success and failure handlers for OAuth2LoginSpec

ndrone:
Added the ability to be able to add custom ServerAuthenticationSuccessHandler and ServerAuthenticationFailureHandler to the OAuth2LoginSpec. If no custom success handler or failure handler is provided the default RedirectServerAuthenticationSuccessHandler or RedirectServerAuthenticationFailureHandler will be used.


jgrandja:
@ndrone Thanks for the PR. However, this is already available since 5.2. See the [code](https://github.com/spring-projects/spring-security/blob/cb9fd091502187cd73610d12b8a19045962be252/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java#L1022).

I'll close this PR as a duplicate.

PR #7879, opened 2020-01-30
Remove redundant branches from SessionManagementConfigurer

OrangeDog:
I've complained about the style in which these are written before, but this was a prime example of how it over-complicates the control flow, leading to (in this case thankfully not dangerous) errors.

Making use of `else` and `&&` would simplify this, and other configuration code, even further.
For this case something like this would be best, but I've had changes like this rejected before.
```
if (this.invalidSessionStrategy == null && this.invalidSessionUrl != null) {
    this.invalidSessionStrategy = new SimpleRedirectInvalidSessionStrategy(
            this.invalidSessionUrl);
}
return this.invalidSessionStrategy;
```

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/SessionManagementConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/SessionManagementConfigurer.java
@@ -557,17 +557,13 @@ public final class SessionManagementConfigurer<H extends HttpSecurityBuilder<H>>
 		if (this.invalidSessionStrategy != null) {
 			return this.invalidSessionStrategy;
 		}
-		if (this.invalidSessionUrl != null) {
-			this.invalidSessionStrategy = new SimpleRedirectInvalidSessionStrategy(
-					this.invalidSessionUrl);
-		}
+		
 		if (this.invalidSessionUrl == null) {
 			return null;
 		}
-		if (this.invalidSessionStrategy == null) {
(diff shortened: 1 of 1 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/configurers/SessionManagementConfigurer.java
```diff
@@ -561,13 +561,6 @@ InvalidSessionStrategy getInvalidSessionStrategy() {
 			this.invalidSessionStrategy = new SimpleRedirectInvalidSessionStrategy(
 					this.invalidSessionUrl);
 		}
-		if (this.invalidSessionUrl == null) {
```
jzheaux: @OrangeDog Thanks for looking into this.

Taking a look at `getExpiredSessionStrategy` below, it looks like there may be an opportunity for cleanup there as well.

What would you think of making these two methods look the same, something like:

```java
if (strategy != null) {
    return strategy;
}

if (url == null) {
    return null;
}

strategy = // get the default strategy
return strategy;
```

I believe this would get rid of the dead branches as well as keep each if statement simple.

jzheaux:
Thanks, @OrangeDog, this is now merged into `master` via ee6df1701b1ca03ffad39c54de252b3cb8319cf7.

Note that I locally rebased the branch and then updated the commit to comply with our [commit formatting guideline](https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.md#format-commit-messages)s.

Issue #7871, opened 2020-01-29
Query parameters in authorization-url are double-encoded


manuelbl:
### Summary

If the autorization URL of a OAuth 2.0 provider contains query parameters with URL encoded characters, they are encoded again and the request is rejected by the IdP.

Example: If the authorization URL in `application.yml` is configured as:

`authorization-uri: 'https://iam.abc.com/idp/oauth2/authorize?claims=%7B%22urn%3Aabc%3Aplace_of_birth%22%3Anull%7D`

Then the resulting URL becomes (somewhat shortened):

`https://iam.abc.com/idp/oauth2/authorize?acr_values=quo1&claims=%257B%2522urn%253Aabc%253Aplace_of_birth%2522%253Anull%257D&response_type=code&client_id=c1&scope=openid%20profile%20email&state=83l9MUgoab0w&redirect_uri=http://localhost:8081/login/oauth2/code/abc&nonce=w4VgoHfH9f71'`

The `claims` parameter is now double encoded. The IdP doesn't understand it anymore and rejects the authorization request.

### Actual Behavior

The query parameter value is double encoded and no longer valid for the IdP.

Correct value in configuration:

`claims=%7B%22urn%3Aabc%3Aplace_of_birth%22%3Anull%7D`

Invalid value in request:

`claims=claims=%257B%2522urn%253Aabc%253Aplace_of_birth%2522%253Anull%257D`

### Expected Behavior

No double encoding should be applied. The query parameter value should be retained without changes.

Correct value in configuration:

`claims=%7B%22urn%3Aabc%3Aplace_of_birth%22%3Anull%7D`

Correct value in request:

`claims=%7B%22urn%3Aabc%3Aplace_of_birth%22%3Anull%7D`

### Notes

If an invalid URL without URL encoding is used::

`claims={"urn:abc:place_of_birth":null}`

the resulting URL is invalid as well (and rejected by the IdP):

`claims={%22urn:abc:date_of_birth%22:null}`

Furthermore, if URLs were to be entered without encoding the query parameters, many parameter values could not be entered as the URL would no longer be parseable.

I am aware of issue #5760 and the associated PR #6299. Unfortunately, it fixes the problem only partially.

[...]

PR #7881, opened 2020-01-30
Prevent double-escaping of authorize URL parameters

manuelbl:
If the authorization URL in the OAuth2 provider configuration contained query parameters with percent escaped characters, these characters were escaped a second time – damaging the URL. This commit fixes it.

It is relevant to support the OIDC claims parameter (see https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter).

Fixes gh-7871

**Implementation notes**

As `UriComponentsBuilder` cannot fully decode query parameters (no implementation for undoing the percent escaping), the approach is to assemble already escaped components. Before adding query parameters to the authorization URL, they must therefore be properly escaped.

The method for percent escaping uses separate `UriComponentsBuilder` instances. That's not straight-forward but the only way to access the method `HierarchicalUriComponents.encodeUriComponent`. It ensures that the minimal necessary escaping according to RFC 3986 is applied.

An alternative would be to use `java.net.URLEncoder.encode`. Contrary to its name, this method applies *HTML form encoding*, which is similar but far more aggressive, i.e. it escapes additional characters such as `/`, `:` etc.

### DIFF
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java
@@ -23,8 +23,10 @@ import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
 import org.springframework.util.StringUtils;
 import org.springframework.web.util.UriComponentsBuilder;
+import org.springframework.web.util.UriUtils;
 
 import java.io.Serializable;
+import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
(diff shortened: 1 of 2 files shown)


oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java
```diff
@@ -376,29 +375,40 @@ public OAuth2AuthorizationRequest build() {
 
 		private String buildAuthorizationRequestUri() {
 			MultiValueMap<String, String> parameters = new LinkedMultiValueMap<>();
-			parameters.set(OAuth2ParameterNames.RESPONSE_TYPE, this.responseType.getValue());
-			parameters.set(OAuth2ParameterNames.CLIENT_ID, this.clientId);
+			parameters.set(OAuth2ParameterNames.RESPONSE_TYPE, percentEncode(this.responseType.getValue()));
```
jgrandja: Please replace `percentEncode` with Spring's `UriUtils.encodeQueryParam(paramValue, StandardCharsets.UTF_8)`

oauth2/oauth2-core/src/test/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequestTests.java
```diff
@@ -307,4 +307,19 @@ public void buildWhenAuthorizationUriIncludesQueryParameterThenAuthorizationRequ
 						"response_type=code&client_id=client-id&state=state&" +
 						"redirect_uri=https://example.com/authorize/oauth2/code/registration-id");
 	}
+
+	@Test
+	public void buildWhenAuthorizationUriIncludesEscapedQueryParameterThenAuthorizationRequestUrlIncludesIt() {
+		OAuth2AuthorizationRequest authorizationRequest =
+				TestOAuth2AuthorizationRequests.request()
```
jgrandja: Can you also add some `additionalParameters` that require encoding.

jgrandja:
Thanks again for providing this fix @manuelbl ! This is now in master and backported to 5.2.x.

Issue #7842, opened 2020-01-17
Set default redirect in OidcClientInitiatedServerLogoutSuccessHandler


mraible:
### Summary

I ❤️ the new `OidcClientInitiatedServerLogoutSuccessHandler` in Spring Security 5.2! The only problem I see with it is you have to set a `postLogoutRedirectUri` property. This will likely vary between development, test, and production, so it seems like kind of a pain to have to maintain a new property that could be calculated.

By default, it'd be nice if the `postLogoutRedirectUri` defaulted to the current context path (`/`). If people wanted to override it, it'd be cool if they could use a relative URL (that starts with `/`) or an absolute URL (current behavior).

I'm entering this here as requested by @jgrandja on [Stack Overflow](https://stackoverflow.com/questions/59650868/how-do-i-logout-when-using-spring-webflux-and-oauth-2-0-login).

### Actual Behavior

You need to set a `postLogoutRedirectUri`.

### Expected Behavior

I'd like this value to default to `/`.

### Configuration

OIDC with Okta and Keycloak (because JHipster).

### Version

Spring Security 5.2

jzheaux:
@mraible I'm glad you like it!

By `/`, is it your intent to redirect to the root of the authorization server or the root of the client application?

@jgrandja Have you already considered the possibility of this being part of the `ClientRegistration`? It seems similar in nature to `ClientRegistration#redirectUriTemplate`.

mraible:
By /, I mean the root of the client application.

jgrandja:
@jzheaux I'm not sure it needs to be part of `ClientRegistration`. It seems it would be fairly trivial to implement a default `postLogoutRedirectUri` in the handler itself.

jzheaux:
> By /, I mean the root of the client application.

Got it, @mraible. So, the existing semantics of this handler are to only include a `post_logout_redirect_uri` parameter if it's specified. I'm concerned that setting a default would surprise users who are upgrading - they'd have to now call `setPostLogoutRedirectUri(null)` or similar to keep their existing behavior.

What about if `setPostLogoutRedirectUri` understood the `{baseUrl}` placeholder in the setter?

```java
handler.setPostLogoutRedirectUri("{baseUrl}")
```

This would work similarly to the `redirect-uri-template` configuration. Would that address your deployment concerns?

mraible:
> This would work similarly to the `redirect-uri-template` configuration. Would that address your deployment concerns?

[...]

PR #7938, opened 2020-02-05
baseUrl placeholder for OidcLogoutSuccessHandlers

jzheaux:
Fixes gh-7842

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/web/logout/OidcClientInitiatedLogoutSuccessHandler.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/web/logout/OidcClientInitiatedLogoutSuccessHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,6 +18,7 @@ package org.springframework.security.oauth2.client.oidc.web.logout;
 
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
+import java.util.Collections;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
@@ -27,7 +28,9 @@ import org.springframework.security.oauth2.client.registration.ClientRegistratio
 import org.springframework.security.oauth2.client.registration.ClientRegistrationRepository;
 import org.springframework.security.oauth2.core.oidc.user.OidcUser;
 import org.springframework.security.web.authentication.logout.SimpleUrlLogoutSuccessHandler;
+import org.springframework.security.web.util.UrlUtils;
 import org.springframework.util.Assert;
+import org.springframework.web.util.UriComponents;
 import org.springframework.web.util.UriComponentsBuilder;
 
 /**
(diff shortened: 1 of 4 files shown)


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/web/logout/OidcClientInitiatedLogoutSuccessHandler.java
```diff
@@ -101,6 +122,26 @@ private String idToken(Authentication authentication) {
...
+	 * </pre>
+	 *
+	 * will make so that {@code post_logout_redirect_uri} will be set to the base url for the client
+	 * application.
+	 *
+	 * @param postLogoutRedirectUriTemplate - A template for creating the {@code post_logout_redirect_uri}
+	 * query parameter
+	 * @since 5.3
+	 */
+	public void setPostLogoutRedirectUriTemplate(String postLogoutRedirectUriTemplate) {
```
jgrandja: I don't really like this option of adding another configuration method that is similar to the existing `setPostLogoutRedirectUri(URI postLogoutRedirectUri)`. It may confuse the user.

I thought of an alternative method that we can try out. What if we use a delegation-based strategy where the delegator calculates the `baseUrl` and then calls `setPostLogoutRedirectUri()` on the delegate before delegating. I think this would work and might be enough to provide @mraible as a custom configuration on his end and no changes on our end. Let's try this out in a sample and see how it goes.

jzheaux: Since the decision would be based on the incoming request, and this is a setter, I think that would cause some concurrency issues.

> I don't really like this option of adding another configuration method

Based on an offline convo with @jgrandja, I think what we could do is deprecate the older method that takes a `URI` and rename this one.

Other classes that have configurable URI values in Spring Security take a String like `mvcMatchers`, `ClientRegistration`, and `RelyingPartyRegistration` because of this expressed desire to be able to have placeholders.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/web/server/logout/OidcClientInitiatedServerLogoutSuccessHandler.java
```diff
@@ -104,14 +119,48 @@ private String idToken(Authentication authentication) {
...
+				.buildAndExpand(Collections.singletonMap("baseUrl", uriComponents.toUriString()))
+				.toUri();
+	}
+
 	/**
 	 * Set the post logout redirect uri to use
 	 *
 	 * @param postLogoutRedirectUri - A valid URL to which the OP should redirect after logging out the user
 	 */
 	public void setPostLogoutRedirectUri(URI postLogoutRedirectUri) {
```
jgrandja: This needs to be `@Deprecated` as well

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/web/server/logout/OidcClientInitiatedServerLogoutSuccessHandler.java
```diff
@@ -104,14 +119,48 @@ private String idToken(Authentication authentication) {
...
+	 * </pre>
+	 *
+	 * will make so that {@code post_logout_redirect_uri} will be set to the base url for the client
+	 * application.
+	 *
+	 * @param postLogoutRedirectUriTemplate - A template for creating the {@code post_logout_redirect_uri}
+	 * query parameter
+	 * @since 5.3
+	 */
+	public void setPostLogoutRedirectUriTemplate(String postLogoutRedirectUriTemplate) {
```
jgrandja: Method name should be `setPostLogoutRedirectUri()`, just like Servlet implementation

jzheaux:
Merged to `master` via 968ebb194bed1c58422cbcd49716aaafb84c9631

PR #7953, opened 2020-02-06
fix #7952 Don't force downcasting of RequestAttributes to ServletRequestAttributes

### LINKED ISSUE #7952: ClassCastException for ServletRequestAttributes
https://github.com/spring-projects/spring-security/commit/0fea57d6a1e0bf25a47e854553e478704fa849ff

In this commit, it casts the RequestAttribute to ServletRequestAttributes, however, in our project, we have some other RequestAttribute implementations based on AbstractRequestAttributes but not based on ServletRequestAttributes, and this fails with ClassCastException. This was working in Spring boot 2.1, but breaks in 2.2 with this change. Can this be changed to check if RequestAttribute is an instance of ServletRequestAttributes first?

smaldini:
Fixes gh-7952


jgrandja:
Thank you for providing this fix @smaldini ! This is now in master and backported to 5.2.x.

PR #7981, opened 2020-02-13
BCryptPasswordEncoder handle null rawPassword

ge1mina023:
Detecting the NullPointerException of rawPassword

I believe that the method of BCryptPasswordEncoder's matches() should have       a way to detect whether the rawPassword is null.Because Spring Security will delete the password of authentication stored in the SecurityContext after I log in, so my other authenticate request will get a NullPointerException from BCryptPasswordEncoder's matches().There should have a way to detect the NullPointerException of rawPassword

### DIFF
--- a/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.java
@@ -123,7 +123,6 @@ public class BCryptPasswordEncoder implements PasswordEncoder {
 			logger.warn("Empty raw password");
 			return false;
 		}
-
 		return BCrypt.checkpw(rawPassword.toString(), encodedPassword);
 	}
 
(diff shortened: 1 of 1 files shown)


crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.java
```diff
@@ -119,6 +119,11 @@ public boolean matches(CharSequence rawPassword, String encodedPassword) {
 			return false;
 		}
 
+		if(rawPassword.toString() == null || rawPassword.toString.length() == 0){
```
jzheaux: Let's have this match the earlier check:

```suggestion
		if (rawPassword == null || rawPassword.length() == 0) {
```

Also, for readability, we should probably place this at the beginning of the method, so that the method parameters are testing in the same order that they are defined.

crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.java
```diff
@@ -119,6 +119,11 @@ public boolean matches(CharSequence rawPassword, String encodedPassword) {
...
+			logger.warn("Empty raw password");
+			return false;
+		}
+
```
jzheaux: Would you also add a couple of tests to `BCryptPasswordEncoderTests`? I think that `doesntMatchNullEncodedValue` and `doesntMatchEmptyEncodedValue` are probably good examples to follow.

jzheaux:
@ge1mina023 in addition to the feedback I left above, would you please compile the code locally with 

```bash
./gradle check
```

and squash your commits?

jzheaux:
Closed in favor of https://github.com/spring-projects/spring-security/pull/8330

Issue #7840, opened 2020-01-16
Introduce OAuth2Authorization success/failure handlers


jgrandja:
This ticket addresses the Servlet implementation of #7699

philsttr:
@jgrandja Feel free to tag me as a reviewer on the implementation if you want an extra set of eyes.

jgrandja:
@philsttr I just submitted the PR for the Servlet implementation #7986. I'm planning on merging this by end of next week. If you have time to review that would be great. No worries if you're busy next week.

jgrandja:
@philsttr I forgot to mention...when I was implementing the Servlet version, I noticed a code path that may not be valid. This specific [test](https://github.com/spring-projects/spring-security/blob/e5fca61810650db1b204f4d8137e8bacbf7e1fbf/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunctionTests.java#L638) throws an `OAuth2AuthorizationException` directly from the `ExchangeFunction`, which ultimately branches into this [code path](https://github.com/spring-projects/spring-security/blob/e5fca61810650db1b204f4d8137e8bacbf7e1fbf/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java#L626). I'm not sure this scenario could happen. Maybe I'm missing something that you are seeing? Could you look at this again and provide some clarification?

philsttr:
That scenario will not happen by the classes and functionality _provided by spring security out-of-the-box_.

I specifically added that code path to allow downstream filters to be able to inspect a resource server response and throw an `OAuth2AuthorizationException` to trigger the logic in the upstream filter function.  This could be utilized by applications talking to a non-compliant resource server.  For example, maybe their resource server returns a 400 status... or maybe it's an rpc-style server that returns an error in the response body.

jgrandja:
Gotchya! That makes sense. Ok I will add it back into the Servlet impl.

jgrandja:
@philsttr There were some package tangles in the Servlet and Reactive implementations. I resolved it via c8cc971. When you get a chance, take a look at the changes and let me know if you have any concerns. Thanks.

philsttr:
@jgrandja LGTM!

PR #7986, opened 2020-02-15
Add OAuth2Authorization success/failure handlers

jgrandja:
Fixes gh-7840

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
@@ -13,15 +13,18 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.springframework.security.oauth2.client.web.reactive.function.client;
 
+import org.springframework.http.HttpStatus;
+import org.springframework.lang.Nullable;
 import org.springframework.security.authentication.AnonymousAuthenticationToken;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.authority.AuthorityUtils;
 import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.oauth2.client.ClientAuthorizationException;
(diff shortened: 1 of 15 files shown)


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -550,4 +649,100 @@ private UnsupportedOperationException unsupported() {
...
+		 * by notifying the authorization failure handler if the http status
+		 * code is in the {@link #httpStatusToOAuth2ErrorCodeMap}.
+		 *
+		 * @param request the request being processed
+		 * @param httpStatusCode the http status returned by the resource server
+		 * @param exception The root cause exception for the failure (nullable)
+		 * @return a {@link Mono} that completes empty after the authorization failure handler completes
+		 */
+		private Mono<Void> handleHttpStatus(ClientRequest request, int httpStatusCode, @Nullable Exception exception) {
+			return Mono.justOrEmpty(this.httpStatusToOAuth2ErrorCodeMap.get(httpStatusCode))
```
philsttr: This applies to both the servlet impl and the reactive impl.... But this could theoretically check for (and parse) a `WWW-Authenticate` header returned from the resource server.  It likely contains the details necessary to create an `OAuth2Error` object.  This would be preferred (and then perhaps fallback to the httpStatusCode -> oauth2ErrorCode mapping if details could not be extracted from `WWW-Authenticate`).

For example:

```
HTTP/1.1 403 Forbidden
WWW-Authenticate: Bearer error="insufficient_scope", error_description="The request requires higher privileges than provided by the access token.", error_uri="https://tools.ietf.org/html/rfc6750#section-3.1"
```

jgrandja: Good idea! Let me see what I can do here.

jgrandja: @philsttr I added the improvement as suggested. This is now in master.

jgrandja:
Merged via 69156b7

Issue #7995, opened 2020-02-19
JwtIssuerReactiveAuthenticationManagerResolver eagerly creates Exceptions


robotmrv:
### Summary

`JwtIssuerReactiveAuthenticationManagerResolver` eagerly creates `Exception`s which are not thrown always   
see https://github.com/spring-projects/spring-security/blob/a90e579350562897ca5f092f591769374cd0d278/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerReactiveAuthenticationManagerResolver.java#L122-L125 
and 
https://github.com/spring-projects/spring-security/blob/a90e579350562897ca5f092f591769374cd0d278/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerReactiveAuthenticationManagerResolver.java#L143-L146
### Actual Behavior

Exceptions are instantiated always 

### Expected Behavior

Exceptions should be instantiated lazily only when they are needed 

### Version

5.3.0.BUILD-SNAPSHOT

PR #7996, opened 2020-02-19
Instantiate exceptions lazily

robotmrv:
Add lazy Exception instantiation and reduce amount of
reactor operators to reduce allocations

Fixes gh-7995

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerReactiveAuthenticationManagerResolver.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerReactiveAuthenticationManagerResolver.java
@@ -16,9 +16,9 @@
 
 package org.springframework.security.oauth2.server.resource.authentication;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Predicate;
@@ -54,6 +54,7 @@ import org.springframework.web.server.ServerWebExchange;
  * <a href="https://tools.ietf.org/html/rfc6750#section-1.2" target="_blank">Bearer Token</a>.
  *
  * @author Josh Cummings
+ * @author Roman Matiushchenko
  * @since 5.3
  */
 public final class JwtIssuerReactiveAuthenticationManagerResolver
@@ -79,8 +80,7 @@ public final class JwtIssuerReactiveAuthenticationManagerResolver
 	public JwtIssuerReactiveAuthenticationManagerResolver(Collection<String> trustedIssuers) {
 		Assert.notEmpty(trustedIssuers, "trustedIssuers cannot be empty");
 		this.issuerAuthenticationManagerResolver =
-				new TrustedIssuerJwtAuthenticationManagerResolver
-						(Collections.unmodifiableCollection(trustedIssuers)::contains);
+				new TrustedIssuerJwtAuthenticationManagerResolver(new ArrayList<>(trustedIssuers)::contains);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerReactiveAuthenticationManagerResolver.java
```diff
@@ -166,7 +167,7 @@ public JwtIssuerReactiveAuthenticationManagerResolver(Collection<String> trusted
 					.filter(this.trustedIssuer)
 					.flatMap(iss ->
```
robotmrv: if assume that predicate is stateless (as far as result is cached indefinitely) this could be rewritten to  
```java
@Override
public Mono<ReactiveAuthenticationManager> resolve(String issuer) {
	if (this.trustedIssuer.test(issuer)) {
		return Mono.empty();
	}
	return this.authenticationManagers.computeIfAbsent(issuer, k ->
			Mono.<ReactiveAuthenticationManager>fromCallable(() ->
					new JwtReactiveAuthenticationManager(ReactiveJwtDecoders.fromIssuerLocation(k))
			)
			.subscribeOn(Schedulers.boundedElastic())
			.cache());
}
```
this will reduce 3 operators and allocations

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerReactiveAuthenticationManagerResolver.java
```diff
@@ -133,26 +133,26 @@ public JwtIssuerReactiveAuthenticationManagerResolver(Collection<String> trusted
...
-					.flatMap(this::issuer);
-		}
-
-		private Mono<String> issuer(BearerTokenAuthenticationToken token) {
-			try {
-				String issuer = JWTParser.parse(token.getToken()).getJWTClaimsSet().getIssuer();
-				return Mono.justOrEmpty(issuer).switchIfEmpty(
-						Mono.error(new InvalidBearerTokenException("Missing issuer")));
-			} catch (Exception e) {
-				return Mono.error(new InvalidBearerTokenException(e.getMessage()));
```
robotmrv: original exception is not passed as `cause`. Is it intentionally?

jzheaux:
Thanks for the polish, @robotmrv. In preparation for merging, will you make two adjustments to the commit messages, please?

1. Both commits refer to additional cleanup items. Would you please ensure that one commit is about lazy exceptions and the other is about the additional cleanup?
2. One of the commits has a long commit message - it's preferred that commit messages be separated into [a title of fewer than 50 characters](https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.md#format-commit-messages) and an optional description for more detail.

robotmrv:
@jzheaux  no problem 
I thought they would eventually be squashed into one

jzheaux:
I see, @robotmrv. The advantage of separating the commits, in this case, is that it clarifies what precisely was needed to fix the issue vs what was a related polish. This can be useful for reference in the future.

robotmrv:
@jzheaux, done 
I've change a little bit second commit. Use `Mono#map` instead of `Mono#handle` as it is more efficient.  
It now `throw`s from mapper lambda but it seems that `reactor-core` team is ok with this (see Simon`s Basle comment to the answer https://stackoverflow.com/questions/53595420/correct-way-of-throwing-exceptions-with-reactor/53596358#53596358)

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerReactiveAuthenticationManagerResolver.java
jzheaux: That's likely an oversight, especially since the servlet version wraps the exception. Would you like to update your PR to include this change as well?

robotmrv: no problem

robotmrv: could it be separate commit or should it be integrated into first/second?

robotmrv: done it polish commit

jzheaux:
Thanks, @robotmrv! This is now merged into `master`.

PR #8009, opened 2020-02-23
reduce once create session

publisherfk:


pivotal-issuemaster:
@publisherfk Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=8009)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=8009) to manually synchronize the status of this Pull Request.

eleftherias:
Thanks for the suggestion @publisherfk.
It may not be obvious from the code, but part of `getSession()`'s behaviour is to create a session if one does not exist.
We cannot accept this change, since we want to create a session even if debug is not enabled.
You can see the expected behaviour in the following test.
https://github.com/spring-projects/spring-security/blob/283e451cadb1e713f44932dc1fad3eabc9fa46e9/web/src/test/java/org/springframework/security/web/context/SecurityContextPersistenceFilterTests.java#L118-L127

PR #8032, opened 2020-02-28
Allow to configure principal name for Resource Server auth

alek-sys:
Add new configuration property `.jwt().principalNameClaim("claim-name")` to configure `JwtAuthenticationConverter` to use value of that claim as a token name, so `authentication.getName()` can contain any value from the token.

If claim name is not provided or JWT token doesn't contain it, default of 'sub' claim will be used to keep changes backward compatible.

Currently the only way to have different name for `JwtAuthenticationToken` authentication object is to override the whole `jwtAuthenticationConverter` and implement a custom converter. However extracting a different principal name seems to be a common requirement, so I think this option can be useful. 

I was thinking to introduce new `JwtAuthenticationConverterSpec` class to chain it more logically, like `jwt().jwtAuthenticationConverter().principalNameClaim("...")` but that is potenitally not backward compatible (requires an extra `.and()` to back to `JwtSpec` level).

### DIFF
--- a/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
@@ -1794,8 +1794,8 @@ public class ServerHttpSecurity {
 		public class JwtSpec {
 			private ReactiveAuthenticationManager authenticationManager;
 			private ReactiveJwtDecoder jwtDecoder;
-			private Converter<Jwt, ? extends Mono<? extends AbstractAuthenticationToken>> jwtAuthenticationConverter
-					= new ReactiveJwtAuthenticationConverterAdapter(new JwtAuthenticationConverter());
+			private Converter<Jwt, ? extends Mono<? extends AbstractAuthenticationToken>> jwtAuthenticationConverter;
+			private String principalNameClaim;
 
 			/**
 			 * Configures the {@link ReactiveAuthenticationManager} to use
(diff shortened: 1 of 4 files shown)


jzheaux:
@alek-sys Thanks for your initiative and ideas.

I hesitate to add more to the `jwt()` DSL as my hope is to keep it quite small. Once the principal name is added, then granted authorities must shortly follow. 

> override the whole `jwtAuthenticationConverter`

There may be some merit to adding the setter, but I think the benefit is quite marginal. To me, it appears to be the difference between:

```java
JwtGrantedAuthoritiesConverter converter = new JwtGrantedAuthoritiesConverter();
return jwt -> {
    Collection<GrantedAuthorities> authorities = converter.convert(jwt);
    return new JwtAuthenticationToken(jwt, authorities, jwt.getClaim("my_claim"));
}
```
and:

```java
JwtAuthenticationConverter converter = new JwtAuthenticationConverter();
converter.setPrincipalNameClaim("my_claim");
return converter;
```

If what `JwtAuthenticationConverter` did was quite complex, it would be a clearer win to add the setter, but, as it is, hardly any additional customization is necessary without the setter.

Are there scenarios I'm missing where the setter would help out a lot more?

alek-sys:
Thanks for feedback @jzheaux. 

My motivation for this PR was to make principal name easily configurable when using `.oauth2ResourceServer`, similar to `.oauth2Login` when `user-name-attribute` is defined for a provider. 

I found I need it while working on Spring Cloud Gateway [sample app](https://github.com/spring-cloud-services-samples/animal-rescue/blob/master/backend/src/main/java/io/spring/cloud/samples/animalrescue/backend/CloudFoundrySecurityConfiguration.java#L48), which uses token relay pattern and passes IdToken via `Authorization` header to a proxied app. So if the app needs to display a username, default `JwtAuthenticationConverter` has to be overriden, otherwise default `sub` claim is used and it contains internal user ID, at least for Cloud Foundry UAA authorization server. 

But now when I put it all together, I feel like this is a very much edge case (gateway + token relay of an id token + UAA + UI on the proxied app). So feel free to close this PR without merging, more I think about it more I doubt this feature will be in high demand with typical `oauth2ResourceServer` use cases (service-to-service communication).

jzheaux:
Good info, @alek-sys. I've closed this PR, then.

Issue #7539, opened 2019-10-17
Strategy for reading the actual CSRF token from the request


rwinch:
### Summary

Currently the CsrfTokenRepository allows loading the expected CsrfToken from the request. It would be nice if the actual CSRF token could be a strategy that could be replaced.

Related gh-7537 gh-7538

PR #8082, opened 2020-03-05
Fixes gh-4001 : CSRF token BREACH Attack

rh-id:
Fixes gh-4001

Apply Base64 encoding at `getToken` string as well

### DIFF
--- a/web/src/main/java/org/springframework/security/web/csrf/CookieCsrfTokenRepository.java
+++ b/web/src/main/java/org/springframework/security/web/csrf/CookieCsrfTokenRepository.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2016 the original author or authors.
+ * Copyright 2012-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,7 +17,7 @@
 package org.springframework.security.web.csrf;
 
 import java.util.UUID;
-
+import java.util.function.Function;
 import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -53,13 +53,15 @@ public final class CookieCsrfTokenRepository implements CsrfTokenRepository {
 
 	private String cookieDomain;
 
+	private Function<String, CsrfToken> generateTokenProvider = (value) -> new DefaultCsrfToken(this.headerName,
+			this.parameterName, value);
+
 	public CookieCsrfTokenRepository() {
 	}
 
 	@Override
 	public CsrfToken generateToken(HttpServletRequest request) {
-		return new DefaultCsrfToken(this.headerName, this.parameterName,
-				createNewToken());
+		return this.generateTokenProvider.apply(createNewToken());
 	}
 
 	@Override
@@ -97,7 +99,7 @@ public final class CookieCsrfTokenRepository implements CsrfTokenRepository {
 		if (!StringUtils.hasLength(token)) {
 			return null;
 		}
-		return new DefaultCsrfToken(this.headerName, this.parameterName, token);
+		return this.generateTokenProvider.apply(token);
 	}
 
 	/**
@@ -105,7 +107,9 @@ public final class CookieCsrfTokenRepository implements CsrfTokenRepository {
 	 *
 	 * @param parameterName the name of the HTTP request parameter that should be used to
 	 * provide a token
+	 * @deprecated use {@link #setGenerateToken(generateTokenProvider)} and pass the parameterName instead.
 	 */
+	@Deprecated
 	public void setParameterName(String parameterName) {
 		Assert.notNull(parameterName, "parameterName is not null");
 		this.parameterName = parameterName;
@@ -116,7 +120,9 @@ public final class CookieCsrfTokenRepository implements CsrfTokenRepository {
 	 *
 	 * @param headerName the name of the HTTP header that should be used to provide the
 	 * token
+	 * @deprecated use {@link #setGenerateToken(generateTokenProvider)} and pass the headerName instead.
 	 */
+	@Deprecated
 	public void setHeaderName(String headerName) {
 		Assert.notNull(headerName, "headerName is not null");
 		this.headerName = headerName;
@@ -195,4 +201,21 @@ public final class CookieCsrfTokenRepository implements CsrfTokenRepository {
 		this.cookieDomain = cookieDomain;
 	}
 
+	/**
+	 * Sets generate token provider<br/>
+	 * <br/>
+	 * Example : <br/>
+	 * <br/>
+	 * {@code (headerName, parameterName, value) -> new  DefaultCsrfToken(headerName, parameterName, value)}<br/>
+	 *
+	 * @param generateTokenProvider provider to be used for generateToken and
+	 *                              loadToken
+	 * @since 5.4
+	 * @see GenerateTokenProvider
+	 * @see XorCsrfToken
+	 */
+	public void setGenerateToken(GenerateTokenProvider<? extends CsrfToken> generateTokenProvider) {
+		this.generateTokenProvider = (value) -> generateTokenProvider.generateToken(this.headerName, this.parameterName,
+				value);
+	}
 }
--- a/web/src/main/java/org/springframework/security/web/csrf/GenerateTokenProvider.java
+++ b/web/src/main/java/org/springframework/security/web/csrf/GenerateTokenProvider.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.web.csrf;
+
+/**
+ * Functional interface to provide CSRF token generation logic
+ *
+ * @author Ruby Hartono
+ *
+ * @param <T> the type of the returned CsrfToken
+ * @since 5.4
+ */
+@FunctionalInterface
+public interface GenerateTokenProvider<T extends CsrfToken> {
+
+	/**
+	 * Generate CsrfToken from parameters
+	 *
+	 * @param headerName    header name
+	 * @param parameterName parameter name
+	 * @param value         token value
+	 * @return CsrfToken generated from parameters
+	 */
+	T generateToken(String headerName, String parameterName, String value);
+}
--- a/web/src/test/java/org/springframework/security/web/csrf/DefaultCsrfTokenTests.java
+++ b/web/src/test/java/org/springframework/security/web/csrf/DefaultCsrfTokenTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2013 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 package org.springframework.security.web.csrf;
+
+import static org.assertj.core.api.Assertions.assertThat;
 
 import org.junit.Test;
 
@@ -54,5 +56,31 @@ public class DefaultCsrfTokenTests {
 	@Test(expected = IllegalArgumentException.class)
 	public void constructorEmptyTokenValue() {
 		new DefaultCsrfToken(headerName, parameterName, "");
+	}
+
+	@Test
+	public void matchesTokenValue() {
+		String tokenStr = "123456";
+		DefaultCsrfToken token = new DefaultCsrfToken(headerName, parameterName, tokenStr);
+		String csrfToken = token.getToken();
+		String csrfToken2 = token.getToken();
+
+		assertThat(token.getToken()).isEqualTo(csrfToken);
+		assertThat(token.getToken()).isEqualTo(csrfToken2);
+		assertThat(csrfToken).isEqualTo(csrfToken2);
+		assertThat(token.matches(csrfToken)).isTrue();
+		assertThat(token.matches(csrfToken2)).isTrue();
+	}
+
+	@Test
+	public void notMatchesTokenValue() {
+		DefaultCsrfToken token1 = new DefaultCsrfToken(headerName, parameterName, "token1");
+		DefaultCsrfToken token2 = new DefaultCsrfToken(headerName, parameterName, "token2");
+		String csrfToken1 = token1.getToken();
+		String csrfToken2 = token2.getToken();
+
+		assertThat(csrfToken1).isNotEqualTo(csrfToken2);
+		assertThat(token1.matches(csrfToken2)).isFalse();
+		assertThat(token2.matches(csrfToken1)).isFalse();
 	}
 }
--- a/config/src/test/java/org/springframework/security/config/http/CsrfConfigTests.java
+++ b/config/src/test/java/org/springframework/security/config/http/CsrfConfigTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -640,7 +640,7 @@ public class CsrfConfigTests {
 			MockHttpServletRequest request = result.getRequest();
 			CsrfToken token = WebTestUtils.getCsrfTokenRepository(request).loadToken(request);
 			assertThat(token).isNotNull();
-			assertThat(token.getToken()).isEqualTo(this.token.apply(result));
+			assertThat(token.matches(this.token.apply(result))).isTrue();
 		}
 	}
 
--- a/docs/manual/src/docs/asciidoc/_includes/about/exploits/csrf.adoc
+++ b/docs/manual/src/docs/asciidoc/_includes/about/exploits/csrf.adoc
@@ -411,3 +411,10 @@ Overriding the HTTP method occurs in a filter.
 That filter must be placed before Spring Security's support.
 Note that overriding only happens on a `post`, so this is actually unlikely to cause any real problems.
 However, it is still best practice to ensure it is placed before Spring Security's filters.
+
+
+[[csrf-considerations-breach-mitigation]]
+==== BREACH Security Exploit Mitigation
+In order to protect our application from BREACH attacks https://en.wikipedia.org/wiki/BREACH consider to use `XorCsrfToken` implementation instead of `DefaultCsrfToken`.
+`XorCsrfToken` implementation provides protection from BREACH by always returning random string for the token value.
+this will make it harder to guess the actual length of each response payload when using HTTP compression.
--- a/docs/manual/src/docs/asciidoc/_includes/servlet/exploits/csrf.adoc
+++ b/docs/manual/src/docs/asciidoc/_includes/servlet/exploits/csrf.adoc
@@ -414,3 +414,62 @@ We have <<csrf-considerations-multipart-body,already discussed>> the trade-offs 
 
 In Spring's Servlet support, overriding the HTTP method is done using https://docs.spring.io/spring-framework/docs/5.2.x/javadoc-api/org/springframework/web/filter/reactive/HiddenHttpMethodFilter.html[HiddenHttpMethodFilter].
 More information can be found in https://docs.spring.io/spring/docs/5.2.x/spring-framework-reference/web.html#mvc-rest-method-conversion[HTTP Method Conversion] section of the reference documentation.
+
+[[servlet-csrf-considerations-breach-mitigation]]
+=== BREACH Security Exploit Mitigation
+We have <<csrf-considerations-breach-mitigation,already discussed>> usage on `XorCsrfToken` implementation.
+
+In Spring's Servlet support, https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/web/csrf/CookieCsrfTokenRepository.html[CookieCsrfTokenRepository] and https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/web/csrf/HttpSessionCsrfTokenRepository.html[HttpSessionCsrfTokenRepository] use https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/web/csrf/DefaultCsrfToken.html[DefaultCsrfToken] instance type as return type for method `generateToken(HttpServletRequest request)` and `loadToken(HttpServletRequest request)`.
+
+In order to override the default implementation in both classes `setGenerateToken(GenerateTokenProvider<? extends CsrfToken> generateTokenProvider)` is introduced.
+We could set the https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/web/csrf/GenerateTokenProvider.html[GenerateTokenProvider] from provided static method https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/web/csrf/XorCsrfToken.html[XorCsrfToken.createGenerateTokenProvider()] as per below:
+
+.setGenerateToken from provided static method
+====
+[source,java]
+----
+CookieCsrfTokenRepository cookieCsrfTokenRepository = new CookieCsrfTokenRepository();
+cookieCsrfTokenRepository.setGenerateToken(XorCsrfToken.createGenerateTokenProvider())
+----
+====
+
+or we could set by defining our own implementation
+
+.setGenerateToken own implementation
+====
+[source,java]
+----
+CookieCsrfTokenRepository cookieCsrfTokenRepository = new CookieCsrfTokenRepository();
+cookieCsrfTokenRepository.setGenerateToken((headerName, parameterName, value) -> new XorCsrfToken(headerName, parameterName, value))
+----
+====
+
+Note that `GenerateTokenProvider` is a `FunctionalInterface` that accepts three parameters `(headerName, parameterName, value)`,
+`headerName` and `parameterName` values are derived/assigned from `CookieCsrfTokenRepository` / `HttpSessionCsrfTokenRepository` fields,
+`value` is a string token value generated/loaded from `CookieCsrfTokenRepository` / `HttpSessionCsrfTokenRepository`.
+
+We could customize `headerName` and `parameterName` in a couple of ways:
+
+.setGenerateToken hard coded
+====
+[source,java]
+----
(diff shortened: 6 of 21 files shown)


rh-id:
Are these classes part of the Reactive implementation? if yes I will update it as well if this commit is ok

`/spring-security-web/src/main/java/org/springframework/security/web/server/csrf/CsrfWebFilter.java`
`/spring-security-web/src/main/java/org/springframework/security/web/server/csrf/CsrfToken.java`
`/spring-security-web/src/main/java/org/springframework/security/web/server/csrf/DefaultCsrfToken.java`

rh-id:
hi @rwinch , I notice this test failing `CsrfInputTagTests.java`. and from the code `CsrfInputTag` it didn't encode the token value, so should the token value encoded and returned from `DefaultCsrfToken.getToken()` or the token should be encoded in `CsrfInputTag.handleToken` ? i think it should be done in `CsrfInputTag`

	@Override
	public String handleToken(CsrfToken token){
		String tokenValue = "";
		try {
			tokenValue = URLEncoder.encode(token.getToken(), StandardCharsets.UTF_8.toString());
		}catch (UnsupportedEncodingException ex) {
			// leave blank
		}
		
		return "<input type=\"hidden\" name=\"" + token.getParameterName()
				+ "\" value=\"" + tokenValue + "\" />";
	}

rh-id:
Travis hitting multiple errors in `CsrfConfigurerTests` but my local only hit 1 test error `CsrfConfigurerTests.postWhenCsrfEnabledAndSessionIsExpiredThenRespondsWithForbidden` at line `356` and I really had no idea why.

web/src/test/java/org/springframework/security/web/csrf/DefaultCsrfTokenTests.java
```diff
@@ -55,4 +56,20 @@ public void constructorNullTokenValue() {
 	public void constructorEmptyTokenValue() {
 		new DefaultCsrfToken(headerName, parameterName, "");
 	}
+
+	// gh-4001
+	@Test
+	public void matchingTokenValue() {
```
rwinch: We should have a test for not matching too

web/src/main/java/org/springframework/security/web/csrf/DefaultCsrfToken.java
```diff
@@ -72,6 +81,73 @@ public String getParameterName() {
 	 * @see org.springframework.security.web.csrf.CsrfToken#getToken()
 	 */
 	public String getToken() {
-		return this.token;
+		Random randomSize = new Random();
```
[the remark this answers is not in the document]
rh-id: I added random size to add more randomness to the token. But I think will have issue with `matches` method logic since like you mentioned required to use constant time comparison.

Will remove this and fix it to 5 instead.

web/src/main/java/org/springframework/security/web/csrf/DefaultCsrfToken.java
```diff
@@ -72,6 +81,73 @@ public String getParameterName() {
 	 * @see org.springframework.security.web.csrf.CsrfToken#getToken()
 	 */
 	public String getToken() {
```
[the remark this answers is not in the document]
rh-id: I do agree that it should be new implementation. It might seemed the new implementation will be related with gh-7539 ?

It doesn't feel right because of the following issues:

1.  It seemed there is a "hidden rules" in generating CsrfToken string. it should be HTML safe, hence why I use Base64

2. Since the `getToken` use Base64 encoding, `matches` also require to decode Base64 in order to get the value.

It seemed that there should be `serialization` and `deserialization` strategy for the token string to deserialized to CsrfToken. Which might require CSRF token string to have data structure other than a simple UUID.

web/src/main/java/org/springframework/security/web/csrf/DefaultCsrfToken.java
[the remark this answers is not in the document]
rwinch: I'm thinking a larger value is better to reduce the likelihood that we get duplicate values. I'd suggest having the value be the same size as the actual token.

web/src/main/java/org/springframework/security/web/csrf/DefaultCsrfToken.java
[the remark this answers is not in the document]
rwinch: I don't believe you need to implement gh-7539 with this ticket. That is more about how the token is read from the request vs what is stored. The logic is more like:

```
storeCsrfToken.matches(readCsrfToken.getValue())
```

That means only the stored value needs to be a custom implementation.

web/src/main/java/org/springframework/security/web/csrf/SecureCsrfToken.java
```diff
@@ -0,0 +1,151 @@
...
+import org.springframework.util.Assert;
+
+/**
+ * A CSRF token that is used to protect against CSRF attacks with extra protection from BREACH exploit
+ *
+ * @author Ruby Hartono
+ * @since 5.4
+ */
+@SuppressWarnings("serial")
+public final class SecureCsrfToken implements CsrfToken {
```
rwinch: I'd prefer a name that is more descriptive. The term secure can and will change over time. I'm not sure I like any of these, but a few ideas to get the ball rolling XorCsrfToken (currently my favorite), SaltedCsrfToken, etc

web/src/main/java/org/springframework/security/web/csrf/SecureCsrfToken.java
```diff
@@ -0,0 +1,151 @@
...
+	 *                      parameter of parametername).
+	 */
+	public SecureCsrfToken(String headerName, String parameterName, String token) {
+		Assert.hasLength(headerName, "headerName cannot be null or empty");
+		Assert.hasLength(parameterName, "parameterName cannot be null or empty");
+		Assert.hasLength(token, "token cannot be null or empty");
+		this.headerName = headerName;
+		this.parameterName = parameterName;
+		this.tokenBytes = Utf8.encode(token);
+		this.secureRandom = new SecureRandom();
```
rwinch: There will be lots of instances of this object that are created. That means we should be able to inject and reuse the `SecureRandom` instance.

web/src/main/java/org/springframework/security/web/csrf/SecureCsrfToken.java
```diff
@@ -0,0 +1,151 @@
...
+	 * (non-Javadoc)
+	 *
+	 * @see org.springframework.security.web.csrf.CsrfToken#getToken()
+	 */
+	public String getToken() {
+		ByteBuffer byteBuffer = ByteBuffer.allocate(this.randomByteSize);
+		this.secureRandom.ints(Byte.MIN_VALUE, Byte.MAX_VALUE).limit(randomByteSize)
+				.forEach((randInt) -> byteBuffer.put((byte) randInt));
+
+		byte[] randomBytes = byteBuffer.array();
```
rwinch: I don't think we need to use `ByteBuffer` here as I think it is unnecessary complexity and object creation. I believe the following would work:

```suggestion
		byte[] randomBytes = new byte[this.randomByteSize];
		this.secureRandom.nextBytes(randomBytes);
```

web/src/main/java/org/springframework/security/web/csrf/SecureCsrfToken.java
```diff
@@ -0,0 +1,151 @@
...
+
+		byte[] xoredCsrf = xorCsrf(randomBytes, this.tokenBytes);
+
+		ByteBuffer combinedBuffer = ByteBuffer.allocate(this.randomByteSize + xoredCsrf.length);
+		combinedBuffer.put(randomBytes);
+		combinedBuffer.put(xoredCsrf);
+
+		// returning randomBytes + XOR csrf token
+		return Base64.getEncoder().encodeToString(combinedBuffer.array());
```
rwinch: I'd prefer to avoid the use of `ByteBuffer` if possible. This will avoid unnecessary overhead (object creation which will mean more Garbage Collection) in a method that is used heavily within the system. It seems that `System.arraycopy` would work well.

```suggestion
		byte[] combinedBytes = new byte[randomBytes.length + xoredCsrf.length];
		System.arraycopy(randomBytes, 0, combinedBytes, 0, randomBytes.length);
		System.arraycopy(xoredCsrf, 0, combinedBytes, randomBytes.length, xoredCsrf.length);

		// returning randomBytes + XOR csrf token
		return Base64.getEncoder().encodeToString(combinedBytes);

File: web/src/main/java/org/springframework/security/web/csrf/SecureCsrfToken.java [hunk 8]
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	private static byte[] xorCsrf(byte[] randomBytes, byte[] csrfBytes) {
```
rwinch: We could further refine this method to return `randomBytes + the xor(randomBytes,csrfBytes)` to avoid an additional `byte[]` being created, but I think that would go past the point of readability vs optimizations.

web/src/main/java/org/springframework/security/web/csrf/SecureCsrfToken.java
```diff
@@ -0,0 +1,151 @@
...
+	 */
+	public SecureCsrfToken(String headerName, String parameterName, String token) {
+		Assert.hasLength(headerName, "headerName cannot be null or empty");
+		Assert.hasLength(parameterName, "parameterName cannot be null or empty");
+		Assert.hasLength(token, "token cannot be null or empty");
+		this.headerName = headerName;
+		this.parameterName = parameterName;
+		this.tokenBytes = Utf8.encode(token);
+		this.secureRandom = new SecureRandom();
+		this.randomByteSize = this.tokenBytes.length;
```
rwinch: I don't think we need to store this since at this point we always assume it is the length of the bytes.

web/src/main/java/org/springframework/security/web/csrf/SecureCsrfToken.java
```diff
@@ -0,0 +1,151 @@
...
+
+		if (paramToken.length == tokenSize) {
+			return MessageDigest.isEqual(this.tokenBytes, paramToken);
+		} else if (paramToken.length < tokenSize) {
+			return false;
+		}
+
+		// extract token and random bytes
+		int paramXorTokenOffset = paramToken.length - tokenSize;
+		ByteBuffer paramXoredToken = ByteBuffer.allocate(tokenSize);
```
rwinch: Again, I'd prefer to avoid the use of `ByteBuffer` here if possible

web/src/main/java/org/springframework/security/web/csrf/SecureCsrfToken.java
rh-id: Initially I thought of naming it with `XorCsrfToken` or `XoredCsrfToken`. it is just that doesn't sound right for naming a class with XOR gate. perhaps `SaltedCsrfToken` is better in this case ?

web/src/main/java/org/springframework/security/web/csrf/SecureCsrfToken.java
rh-id: you mean returning byte array with sequences of `randomBytes + the xor(randomBytes,csrfBytes)` ? I don't see how you could avoid additional `byte[]` being created. 

even so, returning that value will over complicate the logic no?

web/src/main/java/org/springframework/security/web/csrf/SecureCsrfToken.java
rh-id: will use `System.arraycopy` instead for extraction

rh-id:
Hi @rwinch , thanks for the review. I don't think it is suitable to add more `CsrfTokenRepository` implementation only for this token. I'm thinking of adding new constructor to `CookieCsrfTokenRepository` and `HttpSessionCsrfTokenRepository` with `SecureRandom` instance. if the `SecureRandom` instance is null then will use `DefaultCsrfToken` else use `SecureCsrfToken`

as for reactive implementation, does the class `SecureCsrfToken` needed to be copied over to the other package `org.springframework.security.web.server.csrf` ? if so I think it is better to separate the PR / issue. once this PR is merge I will continue with the reactive implementation.

As for documentation I will add on more javadoc on `SecureCsrfToken`

rwinch:
> I'm thinking of adding new constructor to CookieCsrfTokenRepository and HttpSessionCsrfTokenRepository with SecureRandom instance. if the SecureRandom instance is null then will use DefaultCsrfToken else use SecureCsrfToken

I like that approach. I'd like to suggest we use a static factory method instead of making the constructor public. That way we can have a method that creates the SecureRandom for them and another that allows it to be injected (for power users).

> as for reactive implementation, does the class SecureCsrfToken needed to be copied over to the other package org.springframework.security.web.server.csrf ? if so I think it is better to separate the PR / issue. 

Works for me. Can you please create a ticket?

> As for documentation I will add on more javadoc on SecureCsrfToken

We will need to get these updated as well:

* https://github.com/spring-projects/spring-security/blob/master/docs/manual/src/docs/asciidoc/_includes/about/exploits/csrf.adoc
* https://github.com/spring-projects/spring-security/blob/master/docs/manual/src/docs/asciidoc/_includes/servlet/exploits/csrf.adoc

We want to ensure the reference documents how users can leverage the new feature or they won't know it is available.

web/src/main/java/org/springframework/security/web/csrf/SecureCsrfToken.java
rwinch: tl;dr - Please leave as is.

Sorry I shouldn't have even said anything here. I was more thinking out loud and leaving notes for my future self. The main takeaway is:

> but I think that would go past the point of readability vs optimizations.

Which this could be optimized, but isn't worth making the code complicated.

web/src/main/java/org/springframework/security/web/csrf/SecureCsrfToken.java
rwinch: I like Xor because that is what it is actually doing. Salted could cause us problems down the road if there is another implementation that does something similar.

What are your thoughts on the name for this @jzheaux @jgrandja @eleftherias ?

jzheaux: My concern with `XorCsrfToken` is that as security recommendations change, it might end up doing more than XOR in the future. Maybe `PrototypeCsrfToken` since it's unique per-use - others like that are `UniqueCsrfToken` or `SingleUseCsrfToken`. Having the value change on each usage seems to be the goal of the implementation.

Aside from that concern, I don't see any problem with `XorCsrfToken`.

rwinch: If it does more, then it is likely still doing XOR right? If security concerns change drastically, we'd likely name it something else?

rh-id:
Hi @rwinch I had updated the codes, I notice that there might be issue on this change. if implementation were to be updated on `XorCsrfToken` the `CookieCsrfTokenRepository` and `HttpSessionCsrfTokenRepository` might need to be updated as well similar to adding `setSecureRandom`.

If there are new implementation of the `CsrfToken` then both `CookieCsrfTokenRepository` and `HttpSessionCsrfTokenRepository` will need to cater the logic.

perhaps should change the static method factory logic with `CookieCsrfTokenRepository.useXorCsrfToken` and remove the `setSecureRandom` ?

rh-id:
hi @rwinch thanks for the review. the issue with the approach `setGenerateToken(Producer<CsrfToken>)` is that there are required fields to be passed which are `headerName` and `parameterName` to the producer and there might be specific parameter for different CsrfToken implementation for example `SecureRandom` in this case. creating producer class with multiple CsrfToken class implementation might be an issue as well.

to make this more extensible and flexible perhaps we could do it like this?

		BiFunction<String, String, XorCsrfToken> tokenProducer = (headerName,
				parameterName) -> new XorCsrfToken(headerName, parameterName, UUID.randomUUID().toString());
		tokenProducer = tokenProducer.andThen( token -> {
			return new XorCsrfToken(token.getHeaderName(), token.getParameterName(), token.getTokenValue(), new SecureRandom());
			});


so, new method will be `setGenerateToken(BiFunction<String, String, ? extends CsrfToken> generateTokenProducer)`

the inner implementation will then passed in the `headerName` and `parameterName` and apply the function to generate the token
`() -> tokenRef.apply(this.headerName, this.parameterName)`

and the static functions `XorCsrfToken.create()` will return `BiFunction<String, String, XorCsrfToken>` with default UUID string value and SecureRandom

web/src/main/java/org/springframework/security/web/csrf/CookieCsrfTokenRepository.java
```diff
@@ -150,7 +153,7 @@ private String getRequestContext(HttpServletRequest request) {
 
 	/**
 	 * Factory method to conveniently create an instance that has
-	 * {@link #setCookieHttpOnly(boolean)} set to false.
+	 * {@link #setCookieHttpOnly(SecureRandom)} set to false.
```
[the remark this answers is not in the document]
rh-id: yes sorry about that

rwinch:
I was imagining that `setHeaderName` and `setParameterName` would only impact the default `generateToken`. Something like:

```
Producer<CsrfToken> generator = new DefaultCsrfTokenGenerator(parameterName, headerName);
repository.setGenerateToken(generator);
```


If it was overridden the configured headerName and parameterName would not longer be used. We might even deprecate those methods in favor of setting the generateToken.

rh-id:
hi @rwinch I had update the code to apply a slightly different approach.

`Producer<CsrfToken> generator = new DefaultCsrfTokenGenerator(parameterName, headerName);`
I'm also thinking this but there are problems with this approach.

1. checked from the class `HttpSessionCsrfTokenRepository` and `CookieCsrfTokenRepository` there are default values `parameterName` and `headerName` so if we use that approach user might need to maintain the default value themselves or we hack `DefaultCsrfTokenGenerator` to set new `parameterName` and `headerName` whenever the setter was called by user. and there might be possible that the values in `DefaultCsrfTokenGenerator` and `CookieCsrfTokenRepository` might not be in sync.

2. new classes will need to be created whenever there is new implementation of `CsrfToken` `DefaultCsrfTokenGenerator` with `DefaultCsrfToken`, `XorCsrfTokenGenerator` with `XorCsrfToken`.

web/src/main/java/org/springframework/security/web/csrf/GenerateTokenProvider.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Functional interface to provide CSRF token generation logic
+ *
+ * @author Ruby Hartono
+ *
+ * @param <T> the type of the returned CsrfToken
+ * @since 5.4
+ */
+@FunctionalInterface
+public interface GenerateTokenProvider<T extends CsrfToken> {
```
rwinch, line 27: We shouldn't need this interface.

web/src/main/java/org/springframework/security/web/csrf/CookieCsrfTokenRepository.java
```diff
@@ -55,14 +53,15 @@
 
 	private String cookieDomain;
 
-	private Optional<SecureRandom> optionalSecureRandom = Optional.empty();
+	private Function<String, CsrfToken> generateTokenProvider = (value) -> new DefaultCsrfToken(this.headerName,
```
rwinch, line 56: Rather than passing in `createNewToken()` as an argument to the `Function` this should be a `Producer` and it can create the token itself.

rwinch: This is the same `generateTokenProvider` found in `CookieCsrfTokenRepository`. We could move this logic as a static method on `DefaultCsrfToken` just like the static methods on `XorCsrfToken`.

rh-id: the `loadToken` method for 
`CookieCsrfTokenRepository` required to passed in string token as value and construct the `CsrfToken` instance which is why it shouldn't be a `Producer` that can create token itself.

rh-id: even if it was replaced with something like `Producer<CsrfToken> generator = new DefaultCsrfTokenGenerator(parameterName, headerName);`

it will fail on this test

	@Test
	public void generateTokenCustom() {
		String headerName = "headerName";
		String parameterName = "paramName";
		this.repository.setHeaderName(headerName);
		this.repository.setParameterName(parameterName);

		CsrfToken generateToken = this.repository.generateToken(this.request);

		assertThat(generateToken).isNotNull();
		assertThat(generateToken.getHeaderName()).isEqualTo(headerName);
		assertThat(generateToken.getParameterName()).isEqualTo(parameterName);
		assertThat(generateToken.getToken()).isNotEmpty();
	}

because the `headerName` and `parameterName` was not in sync

it can be solved by setting it together something like :

	public void setParameterName(String parameterName) {
		Assert.notNull(parameterName, "parameterName is not null");
		this.parameterName = parameterName;
		this.generateTokenProvider.setParameterName(parameterName);
	}

however it will caused another issue for another kind of test:

		String headerName = "headerName";
		String parameterName = "paramName";
		this.repository.setHeaderName(headerName);
		this.repository.setParameterName(parameterName);
		Producer prod = new DefaultCsrfTokenGenerator(headerName, parameterName);

		this.repository.setGenerateToken(prod);
		prod.setHeaderName("otherHeader");

		CsrfToken generateToken = this.repository.generateToken(this.request);

		assertThat(generateToken).isNotNull();
		assertThat(generateToken.getHeaderName()).isEqualTo(headerName); // this will fail
		assertThat(generateToken.getParameterName()).isEqualTo(parameterName);
		assertThat(generateToken.getToken()).isNotEmpty();

the above test will fail because the value was not in sync

rh-id: The Function & FunctionalInterface approach ensures that we are passing `this.headerName` and `this.parameterName` as reference so that it will always be in sync 

this is to ensure backwards compatibility for `setParameterName` and `setHeaderName` while depreciating it before removed in the future.

rwinch: Sorry. I think you are right that the default generator needs to be able to access the headerName and parameterName values in the event they are updated. Alternatively, setting the headerName and parameterName need to also update the generator.

rh-id: So is this approach fine ? should we deprecate the `setHeaderName` and `setParameterName` in both `CookieCsrfTokenRepository` and `HttpSessionCsrfTokenRepository` as well?

rwinch: Yes

rh-id:
hi @rwinch i had updated the code to use `Base64.getUrlEncoder` instead, because CSRF token might be passed in to the URLs query param.

I had also updated the documentation.

web/src/main/java/org/springframework/security/web/csrf/GenerateTokenProvider.java
rwinch: I think you missed this comment. We shouldn't need this interface. We should use a `Producer<CsrfToken>` instead.

rh-id: We still need this interface. As mentioned from the other comments.

Do you mean to rename this from `GenerateTokenProvider<T extends CsrfToken>` to `Producer<CsrfToken>` ?

rwinch: Ahh sorry. I meant to say`Supplier<CsrfToken.

All occurrences of `GenerateTokenProvider` can be replaced with a `Supplier<CsrfToken>`.

rh-id: do you mean `java.util.function.Supplier` ?

nope, it cannot be replaced. `GenerateTokenProvider` accepts 3 arguments (`headerName`, `parameterName`, and `value`) while `Supplier` accepts no arguments.

rwinch: It would be done as constructor arguments to the `Supplier`. Something like:

```java
public static Supplier<CsrfToken> create(String headerName, String parameterName) {
    return () -> new DefaultCsrfToken(headerName, paraemterName, createTokenValue());
}
```

rh-id: How do you suggest we modify `CookieCsrfTokenRepository.loadToken()` ?
```
@Override
public CsrfToken loadToken(HttpServletRequest request) {
	Cookie cookie = WebUtils.getCookie(request, this.cookieName);
	if (cookie == null) {
		return null;
	}
	String token = cookie.getValue();
	if (!StringUtils.hasLength(token)) {
		return null;
	}
	return this.generateTokenProvider.apply(token);
}
```

rwinch: That is a good point. At this point I think it makes sense to change the API to be a `Converter<String,CsrfToken>` that allows passing in the csrf token string and the output is the `CsrfToken` implementation. When generating a new instance, the `CsrfTokenRepository` is in charge of generating the token and passing it into the `Converter<String,CsrfToken>`.

rh-id: you mean `org.springframework.core.convert.converter.Converter<S,T>` ?

So it would be like below?
```
CookieCsrfTokenRepository repo = new CookieCsrfTokenRepository();
repo.setGenerateToken((tokenStrValue) -> new DefaultCsrfToken("XSRF-TOKEN", "_csrf", tokenStrValue));
```

if so this means that `setHeaderName` and `setParameterName` will not work at all 
```
CookieCsrfTokenRepository repo = new CookieCsrfTokenRepository();
repo.setGenerateToken((tokenStrValue) -> new DefaultCsrfToken("XSRF-TOKEN", "_csrf", tokenStrValue));

repo.setParameterName("customParam"); // the token generated won't follow this value
repo.setHeaderName("customHeader"); // the token generated won't follow this value
```

and another issue is that users will be forced to maintain/hardcode default value of `parameterName` and `headerName` which were defined in both `CookieCsrfTokenRepository` and `HttpSessionCsrfTokenRepository` with default values `_csrf` and `X-CSRF-TOKEN`.

As for the current approach, it still consider `setHeaderName` and `setParameterName` from `CookieCsrfTokenRepository`
```
CookieCsrfTokenRepository repo = new CookieCsrfTokenRepository();
repo.setGenerateToken((headerName, parameterName, tokenStrValue) -> new DefaultCsrfToken(headerName, parameterName, tokenStrValue));

repo.setParameterName("customParam"); // the token generated will follow this value
repo.setHeaderName("customHeader"); // the token generated will follow this value
```

For the current approach, if users decides to maintain their own `headerName` and `parameterName` it could be done as per below
```
CookieCsrfTokenRepository repo = new CookieCsrfTokenRepository();
repo.setGenerateToken((headerName, parameterName, tokenStrValue) -> new DefaultCsrfToken("customHeader", "customParameter", tokenStrValue));

repo.setParameterName("customParam2"); // the token generated will NOT follow this value
repo.setHeaderName("customHeader2"); // the token generated will NOT follow this value
```

sjohnr:
Hi @rh-id. It's been some time and there hasn't been an update on this PR. I noticed the fork is deleted. I was able to recover your commit, but it seems we'd need to start a new PR to keep going. Do you mind if I take it over at this point? I'll keep your name on the existing commit.

rh-id:
Hi @sjohnr please go ahead.

sjohnr:
Closing in favor of #10778.

PR #8109, opened 2020-03-12
Fix exception from empty basic auth header token

zeeshanadnan:
fixes spring-projectsgh-7976

### DIFF
--- a/web/src/test/java/org/springframework/security/web/authentication/www/BasicAuthenticationConverterTests.java
+++ b/web/src/test/java/org/springframework/security/web/authentication/www/BasicAuthenticationConverterTests.java
@@ -111,4 +111,12 @@ public class BasicAuthenticationConverterTests {
 		assertThat(authentication.getName()).isEqualTo("rod");
 		assertThat(authentication.getCredentials()).isEqualTo("");
 	}
+
+	@Test(expected = BadCredentialsException.class)
+	public void requestWhenEmptyBasicAuthorizationHeaderTokenThenError() {
+		MockHttpServletRequest request = new MockHttpServletRequest();
+		request.addHeader("Authorization", "Basic ");
+		converter.convert(request);
+	}
+
 }
--- a/web/src/test/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilterTests.java
+++ b/web/src/test/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilterTests.java
@@ -424,4 +424,20 @@ public class BasicAuthenticationFilterTests {
 		assertThat(SecurityContextHolder.getContext().getAuthentication()).isNull();
 	}
 
+	@Test
+	public void requestWhenEmptyBasicAuthorizationHeaderTokenThenUnauthorized() throws Exception {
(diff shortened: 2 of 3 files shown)


web/src/test/java/org/springframework/security/web/authentication/www/BasicAuthenticationConverterTests.java
```diff
@@ -111,4 +111,15 @@ public void convertWhenEmptyPassword() {
 		assertThat(authentication.getName()).isEqualTo("rod");
 		assertThat(authentication.getCredentials()).isEqualTo("");
 	}
+
+	@Test
+	public void testWhenEmptyBasicAuthorizationHeaderTokenThenIgnored() {
```
eleftherias: Although this test name follows the same pattern as other tests in this class, we would like any new tests to follow the convention methodWhenConditionThenResult.
A possible name could be:
```suggestion
	public void requestWhenEmptyBasicAuthorizationHeaderTokenThenIgnored() {
```

zeeshanadnan:
@eleftherias updated the PR with the requested changes. Also in case of empty basic auth token filter chain should not proceed. Changed it too. Any further code improvements and comments is highly appreciated. Thanks.

web/src/test/java/org/springframework/security/web/authentication/www/BasicAuthenticationConverterTests.java
```diff
@@ -111,4 +112,12 @@ public void convertWhenEmptyPassword() {
 		assertThat(authentication.getName()).isEqualTo("rod");
 		assertThat(authentication.getCredentials()).isEqualTo("");
 	}
+
+	@Test
```
eleftherias: Please add the expected exception to the `@Test` annotation.
For an example see https://github.com/spring-projects/spring-security/blob/7411f25474113907d4d264c38b2fe458c0d27778/web/src/test/java/org/springframework/security/web/authentication/www/BasicAuthenticationConverterTests.java#L94-L100

zeeshanadnan: I followed the convention of using AssertJ in new tests. So used `assertThatThrownBy`.

web/src/test/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilterTests.java
```diff
@@ -424,4 +424,20 @@ public void doFilterWhenTokenAndFilterCharsetDoNotMatchThenUnauthorized() throws
 		assertThat(SecurityContextHolder.getContext().getAuthentication()).isNull();
 	}
 
+	@Test
+	public void requestWhenEmptyBasicAuthorizationHeaderTokenThenIgnored() throws Exception {
```
eleftherias: Please update the test name, since the header is not ignored, but rather returns unauthorized.
```suggestion
	public void requestWhenEmptyBasicAuthorizationHeaderTokenThenUnauthorized() throws Exception {
```

eleftherias:
Thank for the PR @zeeshanadnan! This is now merged into master.

Issue #7160, opened 2019-07-29
NimbusJwtDecoder and NimbusReactiveJwtDecoder should determine algorithm from JWK Set Endpoint


jzheaux:
`NimbusJwtDecoder` and `NimbusReactiveJwtDecoder` could be improved by determining the JWS algorithm from the JWK Set endpoint.

This is something already supported in Nimbus via:

```java
URL url = new URL("https://idp.example.org/.well-known/jwks.json");
JWSKeySelector selector = 
        JWSAlgorithmFamilyJWSKeySelector.fromJWKSetURL(url);
```

The result would be that on startup, the application would hit the JWK Set endpoint, inspect the `kty` field of the JWKs returned, and infer the algorithm or algorithms that the decoder should support accordingly. Perhaps this could be delayed, though, until the first request comes in.

If an application wants to skip this auto-configuration, it can easily do so by specifying which algorithm to use:

```java
String jwkSetUri = "https://idp.example.org/.well-known/jwks.json";
JwtDecoder decoder = NimbusJwtDecoder.fromJwkSetUri(jwkSetUri)
        .jwsAlgorithm(JwsAlgorithm.RS256);
```

Care will need to be taken to ensure that this change is passive. For example, `NimbusJwtDecoder` selects `RS256` by default. For those still picking the defaults, it'd be unfortunate if the algorithms selected by reading the JWKS response didn't include `RS256`.

mkheck:
Hey @jzheaux, I wouldn't mind taking a run at this. Please let me know your thoughts.

mkheck:
Hi @jzheaux, not sure what happened to my earlier comment, but I'd like to take this on if it makes sense to you. Please let me know.

jzheaux:
Yes, @mkheck, I think this would be a good fit. Forgive the delay - I've added some more detail to the description, which I wanted to confirm before handing it off.

PR #8112, opened 2020-03-13
Adds support for JWSKeySelector introspection given a valid JWK Set.

hitchan:
CLA has been submitted.

This PR adds support for creating a JWSVerificationKeySelector dynamically based on the availability of JWK algorithms from a given JWK Set provided by an OIDC discovery document ([Issue 7160](https://github.com/spring-projects/spring-security/issues/7160)). Non-standard or unknown algorithms will be silently ignored (see SignatureAlgorithm class). A warning log has also been added to let developers know if there was a problem fetching the JWK Set, although this does not halt the application from starting.

The original functionality is still in place that adds the RS256 algorithm if none are provided during initial configuration for backwards compatibility.

A new unit test as well as an extended mock JWK set for verification.

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
@@ -16,21 +16,6 @@
 
 package org.springframework.security.oauth2.jwt;
 
-import java.io.IOException;
-import java.net.MalformedURLException;
-import java.net.URL;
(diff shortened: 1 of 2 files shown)


hitchan:
I can't figure out why the test hangs at a certain point. I will resubmit once i have time to sort it out.

izio7:
Hi @hitchan, the same feature is also needed for the org/springframework/security/oauth2/jwt/NimbusReactiveJwtDecoder.java

rwinch:
@hitchan If you are using ok http mock server it will hang if you make a request and haven't queued a response

Issue #8138, opened 2020-03-17
LDAP server configuration should support port=0


jzheaux:
It can be handy when testing to state that your ldap server should use port `0`, that is, indicate that any available port should be selected.

This is especially helpful when tests are running in parallel and could potentially cause a port conflict on startup.

PR #8139, opened 2020-03-17
Allow port=0 for LDAP Servers

jzheaux:
Fixes gh-8138

Note that a change to the XSD was necessary in order to allow for 0 in `port`.

Also, note that I've left one test in each `LdapServerBeanDefinitionParserTests` and `LdapProviderBeanDefinitionParserTests` that don't use port=0 to demonstrate and test default usage.


jzheaux:
This is now merged into `master` via 2d8c65db5620306ad7cb7828d9af3520308a7f77 and 4d99ee289649e351f29476f6ebcd8661d5c51931

Issue #8182, opened 2020-03-24
Add Figures to Resource Server Docs


jzheaux:
Related to #8039

PR #8184, opened 2020-03-24
Add Figures to Resource Server Docs

jzheaux:
Fixes gh-8182


jzheaux:
@rwinch Yes, I mulled over this for some time, and I think there's value in bumping it out in the hierarchy - perhaps into its own section. 

Wherever it goes, I believe it's important to lead with examples, meaning that I believe it should follow the minimal sample. Following "Runtime Expectations" felt the most natural since that section is still in the business of talking about the minimal sample.

The documentation as it stands in this PR begins with a small amount of architectural overview about Bearer Token as a whole only because there is no example that can be formulated at that point in the docs. But, once we get to JWT and opaque token, we can show an example, say "now what did we just see?", and proceed with architectural detail.

I see this as quite similar to what the early servlet docs do: Start with a minimal boot example in section 8, and then go into detail on how things work in section 9.

jzheaux:
Merged via e62b8a758509495302118a8d7bc4e86d555bd1a5 to `master`

PR #8332, opened 2020-04-06
Provide possibility to use custom cache to store JWK Set

20fps:
Update according to the changes requested here https://github.com/spring-projects/spring-security/pull/7639 by @jzheaux to have a possibility to use spring Cache to store JWK Set.

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
@@ -34,6 +34,8 @@ import javax.crypto.SecretKey;
 import com.nimbusds.jose.JOSEException;
 import com.nimbusds.jose.JWSAlgorithm;
 import com.nimbusds.jose.RemoteKeySourceException;
+import com.nimbusds.jose.jwk.JWKSet;
+import com.nimbusds.jose.jwk.source.JWKSetCache;
 import com.nimbusds.jose.jwk.source.JWKSource;
 import com.nimbusds.jose.jwk.source.RemoteJWKSet;
 import com.nimbusds.jose.proc.JWSKeySelector;
@@ -49,6 +51,7 @@ import com.nimbusds.jwt.proc.ConfigurableJWTProcessor;
 import com.nimbusds.jwt.proc.DefaultJWTProcessor;
 import com.nimbusds.jwt.proc.JWTProcessor;
 
+import org.springframework.cache.Cache;
 import org.springframework.core.convert.converter.Converter;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
@@ -68,6 +71,7 @@ import org.springframework.web.client.RestTemplate;
  *
  * @author Josh Cummings
  * @author Joe Grandja
+ * @author Mykyta Bezverkhyi
  * @since 5.2
  */
 public final class NimbusJwtDecoder implements JwtDecoder {
@@ -215,6 +219,7 @@ public final class NimbusJwtDecoder implements JwtDecoder {
 		private String jwkSetUri;
 		private Set<SignatureAlgorithm> signatureAlgorithms = new HashSet<>();
 		private RestOperations restOperations = new RestTemplate();
+		private Cache cache;
 
 		private JwkSetUriJwtDecoderBuilder(String jwkSetUri) {
 			Assert.hasText(jwkSetUri, "jwkSetUri cannot be empty");
@@ -264,6 +269,20 @@ public final class NimbusJwtDecoder implements JwtDecoder {
 			return this;
 		}
 
+		/**
+		 * Use the given {@link Cache} to store
+		 * <a href="https://tools.ietf.org/html/rfc7517#section-5">JWK Set</a>.
+		 *
+		 * @param cache the {@link Cache} to be used to store JWK Set
+		 * @return a {@link JwkSetUriJwtDecoderBuilder} for further configurations
+		 * @since 5.4
+		 */
+		public JwkSetUriJwtDecoderBuilder cache(Cache cache) {
+			Assert.notNull(cache, "cache cannot be null");
+			this.cache = cache;
+			return this;
+		}
+
 		JWSKeySelector<SecurityContext> jwsKeySelector(JWKSource<SecurityContext> jwkSource) {
 			if (this.signatureAlgorithms.isEmpty()) {
 				return new JWSVerificationKeySelector<>(JWSAlgorithm.RS256, jwkSource);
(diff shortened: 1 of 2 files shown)


oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -309,6 +328,33 @@ private static URL toURL(String url) {
...
 
+		private static class CacheJWKSetCache implements JWKSetCache {
+
+			private final Cache jwkSetCache;
+			private final String cacheKey;
+
+			CacheJWKSetCache(Cache jwkSetCache) {
+				Assert.notNull(jwkSetCache, "jwkSetCache cannot be null");
+				this.jwkSetCache = jwkSetCache;
+				this.cacheKey = UUID.randomUUID().toString();
```
jzheaux, line 339: Using a GUID for the key may make it tricky to work with outside of this class. Since the idea here is that you can also use this cache in other places in the code, I think that the JWK Set URL makes more sense as the key.

This also aligns with the sensibility that when caching a method, it's usually keyed by its parameters.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -309,6 +328,33 @@ private static URL toURL(String url) {
 			}
 		}
 
+		private static class CacheJWKSetCache implements JWKSetCache {
```
jzheaux, line 331: Taking another look at this, the way the Nimbus cache refreshes can lead to more than one refresh call:

```java
JWKSet jwkSet = jwkSetCache.get();
if (jwkSetCache.requiresRefresh() || jwkSet == null) {
    // ... do refresh
    jwkSetCache.put(jwkSet);
}
```

If many threads run the above `if` statement at the same time, they will all induce calls to the authorization server.

I think it would be better to wrap a cache around a `ResourceRetreiver` implementation and use `Cache#get(key, callable)` since that provides synchronization guarantees. Then, you'd create a no-op `JWKSetCache` that never caches anything.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -264,6 +270,19 @@ public JwkSetUriJwtDecoderBuilder restOperations(RestOperations restOperations)
...
 		}
 
+		/**
+		 * Use the given {@link Cache} to store
+		 * <a href="https://tools.ietf.org/html/rfc7517#section-5">JWK Set</a>.
+		 *
+		 * @param jwkSetCache the {@link Cache} to be used to store JWK Set
+		 * @return a {@link JwkSetUriJwtDecoderBuilder} for further configurations
+		 */
+		public JwkSetUriJwtDecoderBuilder jwkSetCache(Cache jwkSetCache) {
```
jzheaux, line 280: I'm a bit on the fence here, but I lean towards this method being called `cache` instead of `jwkSetCache`. The reason for that is the `restOperations` method isn't called `jwkSetRestOperations`.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -264,6 +270,19 @@ public JwkSetUriJwtDecoderBuilder restOperations(RestOperations restOperations)
 			return this;
 		}
 
+		/**
+		 * Use the given {@link Cache} to store
+		 * <a href="https://tools.ietf.org/html/rfc7517#section-5">JWK Set</a>.
+		 *
+		 * @param jwkSetCache the {@link Cache} to be used to store JWK Set
+		 * @return a {@link JwkSetUriJwtDecoderBuilder} for further configurations
+		 */
```
jzheaux, line 279: Will you please add a `@since 5.4` annotation here?

20fps: Thanks for your feedback, makes perfect sence, I'll address your comments as soon as I can

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -309,6 +336,37 @@ private static URL toURL(String url) {
...
+			private final ResourceRetriever resourceRetriever;
+
+			CachingResourceRetriever(Cache cache, ResourceRetriever resourceRetriever) {
+				this.cache = cache;
+				this.resourceRetriever = resourceRetriever;
+			}
+
+			@Override
+			public Resource retrieveResource(URL url) {
+				return cache.get(url.toString(), () -> resourceRetriever.retrieveResource(url));
```
jzheaux, line 366: This might not work with some caches since `Resource` isn't `Serializable`. I'd recommend something closer to

```suggestion
				String jwks = cache.get(url.toString(), () -> resourceRetreiver.retreiveResource(url).getContent());
				return new Resource(jwks, "UTF-8");
```
return cache.get(url.toString(), () -> resourceRetriever.retrieveResource(url));

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -280,9 +300,16 @@ public JwkSetUriJwtDecoderBuilder restOperations(RestOperations restOperations)
 			}
 		}
 
+		JWKSource<SecurityContext> jwkSource(ResourceRetriever resourceRetriever) {
+			return Optional.ofNullable(this.cache)
```
jzheaux, line 304: As awesome as `Optional` is, it's too slow to be used in framework code. Would you please use a plain if statement instead?

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
20fps: Yes, I guess we'll have this issues with Redis cache for example, nice one, thanks

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -309,6 +336,38 @@ private static URL toURL(String url) {
...
+			private final ResourceRetriever resourceRetriever;
+
+			CachingResourceRetriever(Cache cache, ResourceRetriever resourceRetriever) {
+				this.cache = cache;
+				this.resourceRetriever = resourceRetriever;
+			}
+
+			@Override
+			public Resource retrieveResource(URL url) {
+				String jwkSet = cache.get(url.toString(), () -> resourceRetriever.retrieveResource(url).getContent());
```
jzheaux, line 366: I think we may want to wrap any exception in an `IOException`, like `RestOperationsResourceRetriever` does. The reason for that is so that Nimbus's `RemoteJWKSet` can react appropriately and throw its `RemoteKeySourceException`. 

My concern is that adding this cache will change the exception callers see from a `RemoteKeySourceException` to a `ValueRetreivalException`.

jzheaux:
In preparation for merging, @20fps, would you please squash your commits and format the remaining one to look something like:

```
Add Cache to NimbusJwtDecoderJwkSetUriBuilder

PR gh-8332
```

20fps:
> In preparation for merging, @20fps, would you please squash your commits and format the remaining one to look something like:
> 
> ```
> Add Cache to NimbusJwtDecoderJwkSetUriBuilder
> 
> PR gh-8332
> ```

np, done

jzheaux:
Thanks, @20fps, this is now merged into `master` via 9133cc24e48cfc345a38cb5d6c1c76e5576a304e. Also, I pushed a polish 10aa9743ed7e790a2fa14f44d4d55a91ce1c7026 as well as a docs ad8c49acaed8d132b1d410fd8f7edaa4bb077bc7 commit.

Issue #8337, opened 2020-04-06
Allow specify header in ServerBearerTokenAuthenticationConverter


TJReinert:
### Summary

I'd like to be able to use the current ServerBearerTokenAuthenticationConverter but want to use a different header (`HttpHeaders.PROXY_AUTHORIZATION`) for my Spring Cloud Gateway Instance.

### Actual Behavior

Authentication must be taken from the `HttpHeaders.AUTHORIZATION`.

### Expected Behavior

I should be able to supply what Header I would like to extract my token from. If not specified default to `HttpHeaders.AUTHORIZATION`.

### Version

All versions, enhancement request.

### Notes

These changes are straight forward and I can open a PR for these changes if deemed necessary.

jzheaux:
That makes sense, @TJReinert, would you be interested in submitting a PR to add that feature?

TJReinert:
Sure thing. Please coach me if there are improvements desired -- first submission to an opensource project.

PR #8341, opened 2020-04-07
Allow custom header during bearer token extraction

TJReinert:
Added ability to specify the header that the
ServerBearerTokenAuthenticationConverter uses to
extract a Bearer Token.

Fixes gh-8337

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/server/ServerBearerTokenAuthenticationConverter.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/server/ServerBearerTokenAuthenticationConverter.java
@@ -50,6 +50,7 @@ public class ServerBearerTokenAuthenticationConverter
 		Pattern.CASE_INSENSITIVE);
 
 	private boolean allowUriQueryParameter = false;
+	private String bearerTokenHeaderName = HttpHeaders.AUTHORIZATION;
 
 	public Mono<Authentication> convert(ServerWebExchange exchange) {
 		return Mono.justOrEmpty(token(exchange.getRequest()))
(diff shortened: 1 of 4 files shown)


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/server/ServerBearerTokenAuthenticationConverter.java
```diff
@@ -90,8 +91,20 @@ public void setAllowUriQueryParameter(boolean allowUriQueryParameter) {
...
 
-	private static String resolveFromAuthorizationHeader(HttpHeaders headers) {
-		String authorization = headers.getFirst(HttpHeaders.AUTHORIZATION);
+	/**
+	 * Set this value to configure what header is checked when resolving a Bearer Token.
+	 * This value is defaulted to {@link HttpHeaders#AUTHORIZATION}.
+	 *
+	 * This allows other headers to be used such as {@link HttpHeaders#PROXY_AUTHORIZATION}
+	 *
+	 * @param authorizationHeader the header to check when retrieving the Bearer Token.
```
jzheaux, line 100: Since this will go out in 5.4, will you please add a `@since 5.4` to the JavaDoc?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/server/ServerBearerTokenAuthenticationConverter.java
```diff
@@ -90,8 +91,20 @@ public void setAllowUriQueryParameter(boolean allowUriQueryParameter) {
...
+	 */
+	public void setAuthorizationHeader(String authorizationHeader) {
```
jzheaux, line 102: I'm thinking that you aren't precisely setting the header, but actually the header name. Also, it's not the authorization header if we are changing it, right? 

Could we instead call this `setBearerTokenHeaderName`?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/server/ServerBearerTokenAuthenticationConverter.java
```diff
@@ -90,8 +91,20 @@ public void setAllowUriQueryParameter(boolean allowUriQueryParameter) {
...
+		this.authorizationHeader = authorizationHeader;
+	}
+
+	private static String resolveFromAuthorizationHeader(HttpHeaders headers, String authorizationHeader) {
+		String authorization = headers.getFirst(authorizationHeader);
```
jzheaux, line 107: Is there a reason `authorizationHeader` needs to be passed in? It seems like it would be a smaller change to use `this.`.

TJReinert: I chose to add a parameter since the method is static. I could not access the ```this.authorizationHeader```.

I don't see a reason it needed to be static. See last push.

TJReinert:
The CONTRIBUTING doc asked that we squash commits, does that include future changes?

jzheaux:
Looks good, @TJReinert! I'm going to add some documentation around your change, and then I think we'll be ready to merge.

I don't think I understand your question about future changes, though. Could you elaborate?

TJReinert:
Apologies, must be losing my mind near end of day. 

I saw that it is asked that we squish our changes. Does that just apply to the 'initial' commit for the pull request? Or should all changes that are the result of a code review also be squished into a single commit. (Causing the need for a force push and comments on the PR to be unattached to a commit).

jzheaux:
Ah, @TJReinert, I see. Yes, it's usually the latter. It's preferred to minimize PR commits to simplify backporting. We aren't backporting in this case, but it's still good hygiene and makes the commit history easier to follow.

jzheaux:
@TJReinert, thanks again! This is now merged into `master` via 2f8eb16d76942b973fcad35660e4db56f58eeb49. Also, I've added a documentation via 419d7264f98a3347160e80bd1b8e07e2f66aa8d6

Issue #8307, opened 2020-04-02
AuthorizeRequestsDsl doesn't allow HTTP Method to be specified


adamu:
### Summary

`AbstractRequestMatcherRegistry` provides the [`mvcMatchers(HttpMethod method, String... mvcPatterns)`](https://github.com/spring-projects/spring-security/blob/b93528138e2b2f7ad34d10a040e6b7ac50fc587a/config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java#L161) function to allow creation of an `MvcMatcher` for a specific HTTP verb, but there is no corresponding `authorize()` function in [`AuthorizeRequestsDsl`](https://github.com/spring-projects/spring-security/blob/2fb3d3d5a2ce1b15829fd36a77db0ab8e34b5b2a/config/src/main/kotlin/org/springframework/security/config/web/servlet/AuthorizeRequestsDsl.kt).

### Actual Behavior

If we want to create an MvcMatcher that matches an HTTP Verb for use with the DSL, we have to construct the matcher manually using a similar method to `AbstractRequestMatcherRegistry`'s protected [`createMvcMatchers()`](https://github.com/spring-projects/spring-security/blob/b93528138e2b2f7ad34d10a040e6b7ac50fc587a/config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java#L170-L191) method, or otherwise use the non-DSL [`authorizeRequests()`](https://github.com/spring-projects/spring-security/blob/2fb3d3d5a2ce1b15829fd36a77db0ab8e34b5b2a/config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java#L1254)

### Expected Behavior

Should be able to specify the HTTP method using the `authorize()` function:

```kotlin
authorizeRequests {
    authorize(POST, "/admin/**", hasRole("ADMIN"))
}
```

### Version

5.3.0.RELEASE

adamu:
Here is part of the log:

```
DEBUG: Created Pull Request #XX(branch="renovate/aws-java-sdk-monorepo")
DEBUG: Checking #XX for automerge(branch="renovate/aws-java-sdk-monorepo")
{
  "automerge": false,
  "automergeType": "pr",
  "automergeComment": "automergeComment"
}
DEBUG: No automerge(branch="renovate/aws-java-sdk-monorepo")
```

And the config (abbreviated):


```
{
[...]

PR #8350, opened 2020-04-08
Add authorize() DSL method that accepts HttpMethod

adamu:
Attempted an implementation for #8307, which adds an implementation of `authorize()` for `AuthorizeRequestsDsl`, so that we can write code like this:

```kotlin
http {
    authorizeRequests {
        authorize(GET, "/path", permitAll)
        authorize(PUT, "/path", denyAll)
    }
}
```

### DIFF
--- a/config/src/main/kotlin/org/springframework/security/config/web/servlet/AuthorizeRequestsDsl.kt
+++ b/config/src/main/kotlin/org/springframework/security/config/web/servlet/AuthorizeRequestsDsl.kt
@@ -35,6 +35,7 @@ class AuthorizeRequestsDsl : AbstractRequestMatcherDsl() {
     private val MVC_PRESENT = ClassUtils.isPresent(
             HANDLER_MAPPING_INTROSPECTOR,
             AuthorizeRequestsDsl::class.java.classLoader)
+    private val PATTERN_TYPE = if (MVC_PRESENT) PatternType.MVC else PatternType.ANT
 
     /**
      * Adds a request authorization rule.
@@ -64,11 +65,7 @@ class AuthorizeRequestsDsl : AbstractRequestMatcherDsl() {
      * (i.e. "hasAuthority('ROLE_USER') and hasAuthority('ROLE_SUPER')")
      */
     fun authorize(pattern: String, access: String = "authenticated") {
-        if (MVC_PRESENT) {
-            authorizationRules.add(PatternAuthorizationRule(pattern, PatternType.MVC, null, access))
-        } else {
-            authorizationRules.add(PatternAuthorizationRule(pattern, PatternType.ANT, null, access))
-        }
+        authorizationRules.add(PatternAuthorizationRule(pattern, PATTERN_TYPE, null, access))
     }
 
     /**
@@ -89,11 +86,7 @@ class AuthorizeRequestsDsl : AbstractRequestMatcherDsl() {
      * (i.e. "hasAuthority('ROLE_USER') and hasAuthority('ROLE_SUPER')")
      */
     fun authorize(pattern: String, servletPath: String, access: String = "authenticated") {
-        if (MVC_PRESENT) {
-            authorizationRules.add(PatternAuthorizationRule(pattern, PatternType.MVC, servletPath, access))
-        } else {
-            authorizationRules.add(PatternAuthorizationRule(pattern, PatternType.ANT, servletPath, access))
-        }
+        authorizationRules.add(PatternAuthorizationRule(pattern, PATTERN_TYPE, servletPath, access))
     }
 
     /**
--- a/config/src/main/kotlin/org/springframework/security/config/web/servlet/AuthorizeRequestsDsl.kt
+++ b/config/src/main/kotlin/org/springframework/security/config/web/servlet/AuthorizeRequestsDsl.kt
@@ -65,7 +65,9 @@ class AuthorizeRequestsDsl : AbstractRequestMatcherDsl() {
      * (i.e. "hasAuthority('ROLE_USER') and hasAuthority('ROLE_SUPER')")
      */
     fun authorize(pattern: String, access: String = "authenticated") {
-        authorizationRules.add(PatternAuthorizationRule(pattern, PATTERN_TYPE, null, access))
+        authorizationRules.add(PatternAuthorizationRule(pattern = pattern,
+                                                        patternType = PATTERN_TYPE,
+                                                        rule = access))
     }
 
     /**
@@ -86,7 +88,10 @@ class AuthorizeRequestsDsl : AbstractRequestMatcherDsl() {
      * (i.e. "hasAuthority('ROLE_USER') and hasAuthority('ROLE_SUPER')")
      */
     fun authorize(pattern: String, servletPath: String, access: String = "authenticated") {
-        authorizationRules.add(PatternAuthorizationRule(pattern, PATTERN_TYPE, servletPath, access))
+        authorizationRules.add(PatternAuthorizationRule(pattern = pattern,
+                                                        patternType = PATTERN_TYPE,
+                                                        servletPath = servletPath,
+                                                        rule = access))
     }
 
     /**
--- a/config/src/main/kotlin/org/springframework/security/config/web/servlet/AuthorizeRequestsDsl.kt
+++ b/config/src/main/kotlin/org/springframework/security/config/web/servlet/AuthorizeRequestsDsl.kt
@@ -16,6 +16,7 @@
 
 package org.springframework.security.config.web.servlet
 
+import org.springframework.http.HttpMethod
 import org.springframework.security.config.annotation.web.builders.HttpSecurity
 import org.springframework.security.config.annotation.web.configurers.ExpressionUrlAuthorizationConfigurer
 import org.springframework.security.web.util.matcher.AnyRequestMatcher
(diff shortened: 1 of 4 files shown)


adamu:
@eleftherias I have split the PR into 3 commits, using your suggested titles.

I also added the extra test.

However, there is some behaviour I am unsure about. For `AuthorizeRequestsByMvcConfigWithAndWithoutHttpMethod`:

```kotlin
http {
    authorizeRequests {
        authorize("/no_specified_method", permitAll)
        authorize(HttpMethod.GET, "/specified_method", permitAll)
        authorize(HttpMethod.PUT, "/specified_method", denyAll)
    }
}
```

With this spec, even though `/no_specified_method` is mapped with a `@RequestMapping`, a `PUT` request to `/no_specified_method` will result in a 403 error.

Additionally, if we change the spec for `/specified_method` so that the `PUT` is permitted and `GET` is denied, the `PUT` request will be denied.

I think this is the behaviour of `MvcMatchersAuthorizedUrl` and is not directly related to the behaviour of this DSL, but I wanted to mention it.

config/src/test/kotlin/org/springframework/security/config/web/servlet/AuthorizeRequestsDslTests.kt
```diff
@@ -271,4 +274,92 @@ class AuthorizeRequestsDslTests {
...
+    @Test
+    fun `request when secured by mvc with http method then responds`() {
+        this.spring.register(AuthorizeRequestsByMvcConfigWithHttpMethod::class.java).autowire()
+
+        this.mockMvc.get("/path")
+            .andExpect {
+                status { isOk }
+            }
+
+        this.mockMvc.put("/path")
```
eleftherias, line 311: PUT requests should include a CSRF token
```suggestion
        this.mockMvc.put("/path") {
            with(csrf())
        }
```

config/src/test/kotlin/org/springframework/security/config/web/servlet/AuthorizeRequestsDslTests.kt
```diff
@@ -271,4 +274,92 @@ class AuthorizeRequestsDslTests {
...
+            .andExpect {
+                status { isForbidden }
+            }
+    }
+
+    @EnableWebSecurity
+    @EnableWebMvc
+    open class AuthorizeRequestsByMvcConfigWithAndWithoutHttpMethod : WebSecurityConfigurerAdapter() {
```
eleftherias, line 319: I had a different test in mind when I mentioned using HTTP method and servlet path.

We should have a test similar to `` `request when secured by mvc with servlet path then responds based on servlet path` ``, but using the HTTP method.

The configuration would look something like
```
http {
    authorizeRequests {
        authorize(HttpMethod.GET, "/path", "/spring", denyAll)
    }
}
```

adamu: Sorry I misunderstood. I'll look at replacing the second test this weekend.

eleftherias: Absolutely, good catch.

adamu:
@eleftherias I have made the following changes:
1. Added the `HttpMethod` for the `antMatchers()` too.
1. Added a corresponding test for that.
1. Added the servlet path test..

Travis failed, but it seems to be due to style problems in `CONTRIBUTING.md`, which I've not changed in this PR.

config/src/test/kotlin/org/springframework/security/config/web/servlet/AuthorizeRequestsDslTests.kt
```diff
@@ -72,12 +76,29 @@ class AuthorizeRequestsDslTests {
...
+            }
+    }
+
     @EnableWebSecurity
     open class AuthorizeRequestsByRegexConfig : WebSecurityConfigurerAdapter() {
         override fun configure(http: HttpSecurity) {
             http {
                 authorizeRequests {
                     authorize(RegexRequestMatcher("/path", null), permitAll)
+                    authorize(RegexRequestMatcher("/post", "POST"), permitAll)
```
eleftherias, line 100: It may be confusing to have a GET request calling "/post". It is best to rename the endpoint to something else.

config/src/main/kotlin/org/springframework/security/config/web/servlet/AuthorizeRequestsDsl.kt
```diff
@@ -152,9 +200,13 @@ class AuthorizeRequestsDsl : AbstractRequestMatcherDsl() {
                     is MatcherAuthorizationRule -> requests.requestMatchers(rule.matcher).access(rule.rule)
                     is PatternAuthorizationRule -> {
                         when (rule.patternType) {
-                            PatternType.ANT -> requests.antMatchers(rule.pattern).access(rule.rule)
+                            PatternType.ANT -> requests.antMatchers(rule.httpMethod, rule.pattern).access(rule.rule)
```
eleftherias, line 203: We should only use `.antMatchers(HttpMethod, String)` if `httpMethod` is not null, similar to `mvcMatchers`

adamu: I can do this, although I notice there is a discrepency between the docs for `mvcMatchers()` and `antMatchers()`.

The [documentation for `antMatchers(HttpMethod httpMethod, String... antPatterns)`](https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.html#antMatchers-org.springframework.http.HttpMethod-java.lang.String...-) says:

> `method` - the `HttpMethod` to use or `null` for any HttpMethod.

And indeed, the non-`HttpMethod` version [just calls the HttpMethod version with `null`](https://github.com/spring-projects/spring-security/blob/b93528138e2b2f7ad34d10a040e6b7ac50fc587a/config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java#L286-L288).

However, the [documentation for `mvcMatchers(HttpMethod method, String... mvcPatterns)`](https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.html#mvcMatchers-org.springframework.http.HttpMethod-java.lang.String...-) does not contain the "or `null`" option for the `method` argument (maybe it should?):

> method - the HTTP method to match on

Although, `mvcMatchers(String... mvcPatterns)` does [internally use a protected method that documents the same behaviour](https://github.com/spring-projects/spring-security/blob/b93528138e2b2f7ad34d10a040e6b7ac50fc587a/config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java#L163-L170).

I think it's cleaner for them both to use the option to pass in `null`, as it is publicly documented behaviour and will remove the need for the conditional check here. But as it's not documented for `mvcMatchers()`, I'm hesitant. What do you think?

I agree they should be unified, so happy to add the conditional check to `antMatchers()` too if you prefer that.

config/src/test/kotlin/org/springframework/security/config/web/servlet/AuthorizeRequestsDslTests.kt
adamu: Renamed to `/onlyPostPermitted`

config/src/main/kotlin/org/springframework/security/config/web/servlet/AuthorizeRequestsDsl.kt
adamu: I pushed the version without the conditionals for now, so you can see what I mean.

eleftherias: Thanks that looks good to me.

config/src/main/kotlin/org/springframework/security/config/web/servlet/AuthorizeRequestsDsl.kt
```diff
@@ -152,12 +200,10 @@ class AuthorizeRequestsDsl : AbstractRequestMatcherDsl() {
...
                         when (rule.patternType) {
-                            PatternType.ANT -> requests.antMatchers(rule.pattern).access(rule.rule)
-                            PatternType.MVC -> {
-                                val mvcMatchersAuthorizeUrl = requests.mvcMatchers(rule.pattern)
-                                rule.servletPath?.also { mvcMatchersAuthorizeUrl.servletPath(rule.servletPath) }
-                                mvcMatchersAuthorizeUrl.access(rule.rule)
-                            }
+                            PatternType.ANT -> requests.antMatchers(rule.httpMethod, rule.pattern).access(rule.rule)
+                            PatternType.MVC -> requests.mvcMatchers(rule.httpMethod, rule.pattern)
+                                .servletPath(rule.servletPath)
```
eleftherias, line 205: I think we should keep the null check for `servletPath`.
I can see that we have null checks in `MvcRequestMatcher`, but I consider this scenario to be different than the `httpMethod`. 
We are never explicitly replacing an unset `servletPath` with null, as we did for `httpMethod`.

adamu: I see. I removed it because:

1. The default value of `servletPath` on `MvcRequestMatcher` is `null` and the documentation for [`setServletPath()`](https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/web/servlet/util/matcher/MvcRequestMatcher.html#setServletPath-java.lang.String-) states "The default is undefined which means any servlet path". This seems to be saying that `null` is equivalent to being unset, which means any servlet path.
1. In `PatternAuthorizationRule`, the default value is also `null` when unspecified. So we can take advantage of the above point to keep the code simpler here.

Would you still like me to add the null check back?

adamu: I added it back, but I'm happy to take it out again if you prefer!

eleftherias:
Thanks again @adamu!

Issue #8391, opened 2020-04-14
Dependency issues in resource server


pgerhard:
I am attempting to create a resource server using for OAuth2 using an opaque Token. According to the documentation (https://docs.spring.io/spring-security/site/docs/current/reference/html5/#dependencies) only the dependency `spring-security-oauth2-resource-server` is required (I am not using JWT so I do not require `spring-security-oauth2-jose`. However the application fails to start to due to a ClassNotFoundException

```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'opaqueTokenIntrospector' defined in class path resource [org/springframework/boot/autoconfigure/security/oauth2/resource/servlet/OAuth2ResourceServerOpaqueTokenConfiguration$OpaqueTokenIntrospectionClientConfiguration.class]: Post-processing of merged bean definition failed; nested exception is java.lang.IllegalStateException: Failed to introspect Class [org.springframework.security.oauth2.server.resource.introspection.NimbusOpaqueTokenIntrospector] from ClassLoader [jdk.internal.loader.ClassLoaders$AppClassLoader@e17e1c67]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:572) ~[spring-beans-5.2.5.RELEASE.jar:5.2.5.RELEASE]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:517) ~[spring-beans-5.2.5.RELEASE.jar:5.2.5.RELEASE]
	at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:323) ~[spring-beans-5.2.5.RELEASE.jar:5.2.5.RELEASE]
	at org.springframework.beans.factory.support.AbstractBeanFactory$$Lambda$455/0000000000000000.getObject(Unknown Source) ~[na:na]
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:222) ~[spring-beans-5.2.5.RELEASE.jar:5.2.5.RELEASE]
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:321) ~[spring-beans-5.2.5.RELEASE.jar:5.2.5.RELEASE]
[...]

PR #8394, opened 2020-04-15
Documented dependencies for opaque Resource Server

pgerhard:
As discussed in #8391, extended the documentation for the opaque Resource Server in order to clearly state which dependencies are required in order for the minimal deployment to work.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -59,7 +59,7 @@ Check out the https://stackoverflow.com/questions/tagged/spring-security[Spring 
 https://spring.io/services[Commercial support] is available too.
 
 == Contributing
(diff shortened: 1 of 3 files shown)


rwinch:
Thanks for the PR @pgerhard! Could you please update the commit to match our Contributor guidelines. In particular it should end with an empty new line and then `Closes gh-8391`

pgerhard:
@rwinch oops. Sorry I will get right on that. While doing this I just noticed that the link to the Contributor guidelines in the README.md is dead. Can I fix this in the PR or do we need an extra ticket?

rwinch:
@pgerhard Ahh good catch. Please send a separate PR for that.

rwinch:
I just went ahead and merged. Thanks again for the PR!

Issue #8144, opened 2020-03-18
ApacheDSContainer should allow a zero port


jzheaux:
ApacheDS's `TcpTransport` allows for a zero port, selecting any available port at startup. It does not correctly propagate that port up through the object graph, though.

For example, giving a `TcpTransport` a port of zero:

```java
server = new LdapServer();
TcpTransport transport = new TcpTransport(port);
server.setTransports(transport);
server.start();
```

will result in a server that is listening on a random port, but `server.getPort()` still returns `0`.

Because ApacheDS has not had a GA release in many years, it's unlikely that this enhancement will get applied to the ApacheDS project.

Still, it would be nice if `ApacheDSContainer` could accept `0` as a port value. One immediate benefit from this is reducing the likelihood of a port collision when multiple LDAP servers are started up simultaneously.

For this to happen, the following test would need to pass:

```java
ApacheDSContainer container = new ApacheDSContainer("dc=springframework,dc=org",
		"classpath:test-server.ldif");
container.setPort(0);
container.afterPropertiesSet();
assertNotEquals(container.getPort(), 0);
assertNotEquals(container.server.getPort(), 0);
assertNotEquals(container.server.getPortSSL(), 0);
```

It may be possible to post-process the `server` instance in `ApacheDSContainer#afterPropertiesSet` into a state where the selected port is correctly returned.

Note that `UnboundIdContainerTests` has a test that would likely be good to port over into `ApacheDSContainerTests` for testing this feature.

rwinch:
Perhaps we can add another accessor on 'ApacheDsContsiner' that would be be used for getting the local port vs configured port?

jzheaux:
That could be nice, yes. I think it should be `getConfiguredPort` so that the same pattern can be applied to `UnboundIdContainer` which already supports a zero port.

evgeniycheban:
I can take this task.

jzheaux:
Thanks, @evgeniycheban! It's yours.

rwinch:
Marking as duplicate of the PR

PR #8416, opened 2020-04-18
Allow port=0 for ApacheDSContainer

evgeniycheban:
Fixes gh-8144

### DIFF
--- a/ldap/src/main/java/org/springframework/security/ldap/server/ApacheDSContainer.java
+++ b/ldap/src/main/java/org/springframework/security/ldap/server/ApacheDSContainer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,6 +15,7 @@
  */
 package org.springframework.security.ldap.server;
 
+import java.net.InetSocketAddress;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -39,6 +40,7 @@ import org.apache.directory.server.protocol.shared.store.LdifFileLoader;
 import org.apache.directory.server.protocol.shared.transport.TcpTransport;
 import org.apache.directory.shared.ldap.exception.LdapNameNotFoundException;
 import org.apache.directory.shared.ldap.name.LdapDN;
+import org.apache.mina.transport.socket.SocketAcceptor;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.DisposableBean;
 import org.springframework.beans.factory.InitializingBean;
@@ -69,6 +71,7 @@ import org.springframework.util.Assert;
  * @author Luke Taylor
  * @author Rob Winch
  * @author Gunnar Hillert
+ * @author Evgeniy Cheban
  * @deprecated Use {@link UnboundIdContainer} instead because ApacheDS 1.x is no longer
  * supported with no GA version to replace it.
  */
@@ -80,6 +83,7 @@ public class ApacheDSContainer implements InitializingBean, DisposableBean, Life
 	final DefaultDirectoryService service;
 	LdapServer server;
 
+	private TcpTransport transport;
 	private ApplicationContext ctxt;
 	private File workingDir;
 
@@ -88,6 +92,7 @@ public class ApacheDSContainer implements InitializingBean, DisposableBean, Life
 	private final JdbmPartition partition;
 	private final String root;
 	private int port = 53389;
+	private int localPort;
 
 	private boolean ldapOverSslEnabled;
 	private File keyStoreFile;
@@ -143,7 +148,7 @@ public class ApacheDSContainer implements InitializingBean, DisposableBean, Life
 		server.setDirectoryService(service);
 		// AbstractLdapIntegrationTests assume IPv4, so we specify the same here
 
-		TcpTransport transport = new TcpTransport(port);
+		this.transport = new TcpTransport(port);
 		if (ldapOverSslEnabled) {
 				transport.setEnableSSL(true);
 				server.setKeystoreFile(this.keyStoreFile.getAbsolutePath());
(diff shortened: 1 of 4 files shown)


ldap/src/main/java/org/springframework/security/ldap/server/ApacheDSContainer.java
```diff
@@ -143,6 +148,9 @@ public void afterPropertiesSet() throws Exception {
 		server.setDirectoryService(service);
 		// AbstractLdapIntegrationTests assume IPv4, so we specify the same here
 
+		if (this.port == 0) {
+			this.port = getRandomPort();
+		}
```
rwinch, line 153: Sorry for the confusion. We are already finding a random port in this way within our configuration support. 

The problem is sometimes the port gets resolved to an open port and assigned, the port is then taken up by another process or thread, and then this code tries to use the now taken port.

To fix this we want to be able to pass in `0` to `TcpTransport` to avoid a race condition.

evgeniycheban: Thanks for the review. Do I understand correctly that we should also add an accessor to `ApacheDSContainer` for configured port value i.e `0`?

rwinch: It's not for the configured value, but for the actual port that is resolved by `TcpTransport`.

evgeniycheban: I think it should be `getLocalPort` for the actual port.

rwinch: Yes we want the name of the method to be getLocalPort

evgeniycheban: Done.

ldap/src/main/java/org/springframework/security/ldap/server/ApacheDSContainer.java
```diff
@@ -151,6 +155,10 @@ public void afterPropertiesSet() throws Exception {
 		}
 		server.setTransports(transport);
 		start();
+
+		SocketAcceptor socketAcceptor = server.getSocketAcceptor(transport);
+		InetSocketAddress localAddress = socketAcceptor.getLocalAddress();
+		this.localPort = localAddress.getPort();
```
rwinch, line 161: I think this logic should probably be moved to the `start` method since a user could invoke `start` directly.

evgeniycheban: Should we also move the `server` initialization to the `start` method?

rwinch: Let's try and make as little changes in this PR as possible. If we see additional room for improvement, let's create separate tickets.

evgeniycheban: Done.

evgeniycheban:
@rwinch I made changes to `LdapAuthenticationProviderConfigurer`.

evgeniycheban:
> Thanks for the updates. We still need changes in LdapServerBeanDefinitionParser. We should also make sure our integration tests are always using port 0 and not use `new ServerSocket(0).getLocalPort()` anymore.

I made changes for `LdapServerBeanDefinitionParser ` as well.
What should we do with tests that use `ServerSocket(0).getLocalPort()` to get a list of available ports? For example `ApacheDSContainerTests#multipleInstancesSimultanciously`.

evgeniycheban:
@rwinch Could you please take a look at the latest changes when you have a moment?

ldap/src/main/java/org/springframework/security/ldap/DefaultSpringSecurityContextSource.java
```diff
@@ -171,4 +173,14 @@ private static String buildProviderUrl(List<String> urls, String baseDn) {
 
 	}
 
+	/**
+	 * Create and initialize an instance which will connect to the embedded LDAP server.
+	 *
+	 * @param container the embedded DS container
+	 * @return an instance which will connect to the embedded LDAP server
+	 */
+	public static DefaultSpringSecurityContextSource createEmbeddedContextSource(EmbeddedDsContainer container) {
```
rwinch, line 182: I don't think we should need to add this method. I'd prefer to keep as many things private scope as possible.

ldap/src/main/java/org/springframework/security/ldap/server/EmbeddedDsContainer.java
```diff
@@ -0,0 +1,40 @@
...
+package org.springframework.security.ldap.server;
+
+/**
+ * Common interface for embedded DS containers.
+ *
+ * @author Evgeniy Cheban
+ * @see ApacheDSContainer
+ * @see UnboundIdContainer
+ */
+public interface EmbeddedDsContainer {
```
rwinch, line 25: I don't think we should need to add a new public interface.

ldap/src/main/java/org/springframework/security/ldap/DefaultSpringSecurityContextSource.java
evgeniycheban: I added this factory method for XML config because we need to create and post process a bean of embedded container first and then use actual port value to build a url.

rwinch: This can likely be done with a package private method within the config itself.

evgeniycheban: Could you please provide some example of how this can be done in BeanDefinitionParser?

ldap/src/main/java/org/springframework/security/ldap/server/EmbeddedDsContainer.java
evgeniycheban: Removed.

ldap/src/main/java/org/springframework/security/ldap/DefaultSpringSecurityContextSource.java
evgeniycheban: Got it. Added `EmbeddedLdapServerConfigBean` with factory method.

rwinch:
Thanks for the Pull Request and your patience as we worked on finding the best solution! This is now merged into master via 0fa339f75b274d480f25b466d842d72eaeebe34d :smile:

Issue #2884, opened 2014-06-23
SEC-2664:  ActiveDirectoryLdapAuthenticationProvider should wrap communication exceptions in InternalAuthenticationServiceException


spring-issuemaster:
[Pradeep Agrawal](https://jira.spring.io/secure/ViewProfile.jspa?name=pradeepagrawal8184) (Migrated from [SEC-2664](https://jira.spring.io/browse/SEC-2664?redirect=false)) said:

Jun 23, 2014 2:53:43 PM org.apache.catalina.core.StandardWrapperValve invoke
SEVERE: Servlet.service() for servlet [default] in context with path [/myapp] threw exception
org.springframework.ldap.CommunicationException: myazuread.cloudapp.net:389; nested exception is javax.naming.CommunicationException: myazuread.cloudapp.net:389 [Root exception is java.net.UnknownHostException: myazuread.cloudapp.net]
    at org.springframework.ldap.support.LdapUtils.convertLdapException(LdapUtils.java:100)
    at org.springframework.security.ldap.authentication.ad.ActiveDirectoryLdapAuthenticationProvider.bindAsUser(ActiveDirectoryLdapAuthenticationProvider.java:170)
    at org.springframework.security.ldap.authentication.ad.ActiveDirectoryLdapAuthenticationProvider.doAuthentication(ActiveDirectoryLdapAuthenticationProvider.java:111)
    at org.springframework.security.ldap.authentication.AbstractLdapAuthenticationProvider.authenticate(AbstractLdapAuthenticationProvider.java:61)
    at org.springframework.security.authentication.ProviderManager.authenticate(ProviderManager.java:156)
    at org.springframework.security.web.authentication.UsernamePasswordAuthenticationFilter.attemptAuthentication(UsernamePasswordAuthenticationFilter.java:94)
    at org.springframework.security.web.authentication.AbstractAuthenticationProcessingFilter.doFilter(AbstractAuthenticationProcessingFilter.java:194)
    at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:323)
    at org.springframework.security.web.authentication.logout.LogoutFilter.doFilter(LogoutFilter.java:105)
    at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:323)
    at org.springframework.security.web.context.SecurityContextPersistenceFilter.doFilter(SecurityContextPersistenceFilter.java:87)
[...]

PR #8418, opened 2020-04-19
ActiveDirectoryLdapAuthenticationProvider uses InternalAuthenticationServiceException

dadikovi:
…ication exceptions in InternalAuthenticationServiceException 
GitHub issue: gh-2884

**This is just a proposal without new test cases or documentations.**

Please let me know what kind of test/documentation changes should I make together with this change.

### DIFF
--- a/core/src/main/resources/org/springframework/security/messages.properties
+++ b/core/src/main/resources/org/springframework/security/messages.properties
@@ -31,6 +31,7 @@ DigestAuthenticationFilter.usernameNotFound=Username {0} not found
 JdbcDaoImpl.noAuthority=User {0} has no GrantedAuthority
 JdbcDaoImpl.notFound=User {0} not found
 LdapAuthenticationProvider.badCredentials=Bad credentials
+LdapAuthenticationProvider.badLdapConnection=Connection to LDAP server failed
 LdapAuthenticationProvider.credentialsExpired=User credentials have expired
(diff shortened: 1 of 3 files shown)


rwinch:
Thanks for the PR! Can you please add a test and update the commit message to match our conventions listed in [CONTRIBUTING](https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.adoc#format-commit-messages)? Something like:

```
ActiveDirectoryLdapAuthenticationProvider Wraps CommunicationException

CommunicationException are now wrapped in InternalAuthenticationServiceException

Closes gh-2884
```

dadikovi:
I sent another iteration of this PR, but I have a few question:

- CommunicationException is thrown from bindAsUser and searchForUser methods as well, so I changed the doAuthentication method (bindAsUser is now in the try-block as well), because I wanted to keep this wrapping logic in one place. (Please review my decision :) )
- I had to introduce a new error message, I put it in messages.properties. Since I'm new here I don't know if I should take care of the translations or not?
- I can imagine that someone expects that a CommunicationException is thrown if LDAP is not accessible, so this change is an API-breaking change for them. Should I write some changelog or documentation about it?

dadikovi:
(Also, I will fix the issues as soon as I can)

rwinch:
Thanks for the Pull Request! This is now merged into master via f7b33da5775d2bef504c1671a83a52e48759d13b :smile:

Issue #8421, opened 2020-04-21
OAuth2 Resource Server docs not in sync - authorityPrefix can't be set to ""


judomu:
In spring-security documentation (https://docs.spring.io/spring-security/site/docs/current/reference/html5/):

> Or, you can remove the prefix altogether by calling JwtGrantedAuthoritiesConverter#setAuthorityPrefix("").

But that will not work

```
	public void setAuthorityPrefix(String authorityPrefix) {
		Assert.hasText(authorityPrefix, "authorityPrefix cannot be empty");
		this.authorityPrefix = authorityPrefix;
	}
```

So the documentation is wrong or the implementation, imho the option to set authorityPrefix to "" makes sense to me.

rwinch:
Thanks for the report. I think it is the implementation that is broken. Would you like to submit a PR to fix it?

judomu:
Sure, I will provide the fix

rwinch:
Thanks @judomu! The issue is yours

rwinch:
This is now merged into master and 5.2.x

PR #8424, opened 2020-04-21
Enables empty authorityPrefix in JwtGrantedAuthoritisConverter

judomu:
- docs stated that empty authorityPrefix are allowed but implementation denied to use `""`
- commit removes the `hasText`-limitation but restricts to `notNull`

Fixes gh-8421


rwinch:

PR #8450, opened 2020-05-01
Deprecate OpenID 2.0 support

dadikovi:
This commit puts deprecation notice on docs, sample applications and configurations (java and xml).

Fixes gh-7153

### DIFF
--- a/docs/manual/src/docs/asciidoc/_includes/servlet/authentication/index.adoc
+++ b/docs/manual/src/docs/asciidoc/_includes/servlet/authentication/index.adoc
@@ -33,6 +33,7 @@ This also gives a good idea of the high level flow of authentication and how pie
 * <<servlet-rememberme, Remember Me>> - How to remember a user past session expiration
 * <<servlet-jaas, JAAS Authentication>> - Authenticate with JAAS
 * <<servlet-openid,OpenID>> - OpenID Authentication (not to be confused with OpenID Connect)
+// FIXME: The one above is deprecated. Should it be removed from here as well?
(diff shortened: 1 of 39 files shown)


dadikovi:
Thanks for the feedback @rwinch!

- Modified the deprecation note, so it points to spring-security-oauth2 now, which contains the OIDC support.
- Cherry-picked ThomasVitale's changes, and then in a separate commit added the oauth2 note to them as well.

> We should ensure that all the parsers related to the configuration are deprecated as well (this will trigger tooling to give deprecations)

I couldn't understand this part. I should modify the rnc files?

docs/manual/src/docs/asciidoc/_includes/servlet/authentication/index.adoc
```diff
@@ -33,6 +33,7 @@ This also gives a good idea of the high level flow of authentication and how pie
 * <<servlet-rememberme, Remember Me>> - How to remember a user past session expiration
 * <<servlet-jaas, JAAS Authentication>> - Authenticate with JAAS
 * <<servlet-openid,OpenID>> - OpenID Authentication (not to be confused with OpenID Connect)
+// FIXME: The one above is deprecated. Should it be removed from here as well?
```
rwinch, line 36: We do not need to remove it here. We still want users to find it and figure out it has been deprecated.
```suggestion
```

dadikovi:
Thanks @rwinch for your response, I sent another commit with the requested changes:

- modified rnc file, then generated the xsd-s, so my earlier manual modifications could be overwritten.
- removed FIXME comment
- extracted parser method which is called only if there is related tag in the xml configuration, and which has deprecation notice.

rwinch:
Thanks for the updated Pull Request! This is now merged into master :smile:

Issue #7745, opened 2019-12-14
SecurityMockMvcConfigurer regression in Spring Boot 2.2.1


JohnWu-Pro:
### Summary

SecurityMockMvcConfigurer$DelegateFilter is not null-safe, if no `springSecurityFilterChain` is available.

### Actual Behavior

```
java.lang.NullPointerException
	at org.springframework.security.test.web.servlet.setup.SecurityMockMvcConfigurer$DelegateFilter.doFilter(SecurityMockMvcConfigurer.java:132)
	at org.springframework.mock.web.MockFilterChain.doFilter(MockFilterChain.java:134)
	at org.springframework.test.web.servlet.MockMvc.perform(MockMvc.java:183)
	at io.restassured.module.mockmvc.internal.MockMvcRequestSenderImpl.performRequest(MockMvcRequestSenderImpl.java:219)
	at io.restassured.module.mockmvc.internal.MockMvcRequestSenderImpl.sendRequest(MockMvcRequestSenderImpl.java:448)
	at io.restassured.module.mockmvc.internal.MockMvcRequestSenderImpl.put(MockMvcRequestSenderImpl.java:505)
	at io.restassured.module.mockmvc.internal.MockMvcRequestSenderImpl.put(MockMvcRequestSenderImpl.java:101)
	at com.example.demo.rest.DemoResourceTest.testEcho(DemoResourceTest.java:36)
```

### Expected Behavior

No exception should be thrown even when no `springSecurityFilterChain` is available.

### Configuration

Our application uses spring security, and `org.springframework.security.test.web.servlet.setup.SecurityMockMvcConfigurers` is also in the test classpath. In a REST endpoint test (Mockito+RestAssured based, not full integration test, and security is not of interest), `springSecurityFilterChain` is not available.

### Version

The issue exists in spring-security-test 2.2.x. In spring-security-test 2.0.x, it works fine.

Related to issue https://github.com/spring-projects/spring-security/issues/7688

### Sample

```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>
    <parent>
        <groupId>org.springframework.boot</groupId>
[...]

PR #8451, opened 2020-05-01
SecurityMockMvcConfigurer$DelegateFilter is not null-safe

dadikovi:
This commit adds null-checks in relevant methods, so instead of NPE an IllegalStateException will be thrown with additional details.

Fixes gh-7745

### DIFF
--- a/test/src/main/java/org/springframework/security/test/web/servlet/setup/SecurityMockMvcConfigurer.java
+++ b/test/src/main/java/org/springframework/security/test/web/servlet/setup/SecurityMockMvcConfigurer.java
@@ -118,38 +118,44 @@ final class SecurityMockMvcConfigurer extends MockMvcConfigurerAdapter {
 		}
 
 		Filter getDelegate() {
-			return this.delegate;
+			Filter result = this.delegate;
+			if (result == null) {
+				throw new IllegalStateException("delegate cannot be null. Ensure a Bean with the name "
+						+ BeanIds.SPRING_SECURITY_FILTER_CHAIN
+						+ " implementing Filter is present or inject the Filter to be used.");
+			}
+			return result;
 		}
 
 		@Override
 		public void init(FilterConfig filterConfig) throws ServletException {
-			this.delegate.init(filterConfig);
+			getDelegate().init(filterConfig);
 		}
 
 		@Override
 		public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
 				throws IOException, ServletException {
-			this.delegate.doFilter(request, response, chain);
+			getDelegate().doFilter(request, response, chain);
 		}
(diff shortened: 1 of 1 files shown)


test/src/main/java/org/springframework/security/test/web/servlet/setup/SecurityMockMvcConfigurer.java
```diff
@@ -123,33 +123,47 @@ Filter getDelegate() {
...
 			return this.delegate.equals(obj);
 		}
 
 		@Override
 		public String toString() {
+			this.requireDelegate();
 			return this.delegate.toString();
 		}
+
+		private void requireDelegate() {
```
rwinch, line 161: Rather than do this could we update getDelegate to do a null check. Also keeping in mind that there is a race condition working on the member variable. So it would be something like:

```java
Filter result = this.delegate;
if (result == null) {
   // throw
}
return result;
```

test/src/main/java/org/springframework/security/test/web/servlet/setup/SecurityMockMvcConfigurer.java
```diff
@@ -123,33 +123,47 @@ Filter getDelegate() {
...
 
 		@Override
 		public boolean equals(Object obj) {
+			this.requireDelegate();
 			return this.delegate.equals(obj);
 		}
 
 		@Override
 		public String toString() {
+			this.requireDelegate();
```
rwinch, line 157: We don't use `this.` on methods...only members.

dadikovi: Okay, I fixed it.
If you think I could send another PR with a new checkstyle rule, something like this:

`<module name="Regexp">
			<property name="format" value="this\s*\.[a-zA-Z]+\(\)"/>
			<property name="illegalPattern" value="true"/>
			<property name="message" value="this. on method"/>
		</module>`

Off course this PR would include the fixing of existing matches on that pattern as well (currently there are 100+)

test/src/main/java/org/springframework/security/test/web/servlet/setup/SecurityMockMvcConfigurer.java
dadikovi: Force-pushed this modification, thanks!

test/src/main/java/org/springframework/security/test/web/servlet/setup/SecurityMockMvcConfigurer.java
rwinch: That would be great. Thanks!

rwinch:

Issue #8149, opened 2020-03-19
NPE for ill-constructed HTTP requests with non-standard HTTP method


Robbert1:
I expected the matcher to not trigger NPE's (resulting in HTTP 500's) but rather choosing to match or not for unknown methods.

This line triggers the NPE as getMethod returns null for any unmappable method string.
https://github.com/spring-projects/spring-security/blob/06fdb83fb89840c511b2bc46f72b7c49229c9dab/web/src/main/java/org/springframework/security/web/server/csrf/CsrfWebFilter.java#L190

eleftherias:
Thanks for the report @Robbert1.
Would you be interested in submitting a PR?

parikshitdutta:
@eleftherias @rwinch how is it for contribution? I can look into it.

eleftherias:
Thanks for the offer @parikshitdutta. The issue is yours!

parikshitdutta:
Hi @eleftherias, Please take a look at PR #8452, or Please assign it to respective reviewer.

Thank you.
+ @rwinch

rwinch:
Closed in favor of gh-8452

PR #8452, opened 2020-05-01
Fixed NPE for HTTP requests with non-standard HTTP method for CsrfWebFilter

parikshitdutta:
Fixed NPE for HTTP requests with non-standard HTTP method in CsrfWebFilter

1. CsrfWebFilter code updated to deal with null value while non-standard HTTP methods used
2. Respective test added to validate the scenario with non-standard (custom) HTTP method

Expected Result:
Non-standard HTTP methods should match for CsrfProtection Requirement now, instead of throwing NPE and resulting HTTP 5xx 

Fixes gh-8149

### DIFF
--- a/web/src/main/java/org/springframework/security/web/server/csrf/CsrfWebFilter.java
+++ b/web/src/main/java/org/springframework/security/web/server/csrf/CsrfWebFilter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -60,6 +60,7 @@ import static java.lang.Boolean.TRUE;
  * </p>
  *
  * @author Rob Winch
+ * @author Parikshit Dutta
  * @since 5.0
  */
 public class CsrfWebFilter implements WebFilter {
@@ -187,7 +188,7 @@ public class CsrfWebFilter implements WebFilter {
 		@Override
 		public Mono<MatchResult> matches(ServerWebExchange exchange) {
 			return Mono.just(exchange.getRequest())
-				.map(r -> r.getMethod())
+				.flatMap(r -> Mono.justOrEmpty(r.getMethod()))
 				.filter(m -> ALLOWED_METHODS.contains(m))
 				.flatMap(m -> MatchResult.notMatch())
 				.switchIfEmpty(MatchResult.match());
--- a/web/src/test/java/org/springframework/security/web/server/csrf/CsrfWebFilterTests.java
+++ b/web/src/test/java/org/springframework/security/web/server/csrf/CsrfWebFilterTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,6 +20,8 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.Mock;
 import org.mockito.junit.MockitoJUnitRunner;
+
+import org.springframework.http.HttpMethod;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.mock.http.server.reactive.MockServerHttpRequest;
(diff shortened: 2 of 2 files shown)


web/src/main/java/org/springframework/security/web/server/csrf/CsrfWebFilter.java
```diff
@@ -187,7 +187,7 @@ public static void skipExchange(ServerWebExchange exchange) {
 		@Override
 		public Mono<MatchResult> matches(ServerWebExchange exchange) {
 			return Mono.just(exchange.getRequest())
-				.map(r -> r.getMethod())
+				.map(r -> (r.getMethod() != null ? r.getMethod() : Mono.empty()))
```
rwinch, line 190: This doesn't look right to me. You shouldn't map to a publisher. It should be a `.flatMap` and both code paths need to be a `Publisher`

```suggestion
				.flatMap(r -> Mono.justOrEmpty(r.getMethod()))
```

web/src/test/java/org/springframework/security/web/server/csrf/CsrfWebFilterTests.java
```diff
@@ -284,4 +305,12 @@ String ok() {
 			return "ok";
 		}
 	}
+
+	static class MockNonStandardHttpRequest {
```
rwinch, line 309: Is there a need to externalize this or could you just directly use `MockServerHttpRequest.method` directly?

web/src/test/java/org/springframework/security/web/server/csrf/CsrfWebFilterTests.java
```diff
@@ -183,6 +188,22 @@ public void filterWhenPostAndEstablishedCsrfTokenAndHeaderValidTokenThenContinue
 		chainResult.assertWasSubscribed();
 	}
 
+	@Test
+	public void filterWhenNonStandardHTTPMethodUsedThenMatchesRequireCSRFProtection() {
+		MockServerWebExchange nonStandardHttpRequest = from(MockNonStandardHttpRequest.nonStandardHTTPMethod("/"));
+
+		ServerWebExchangeMatcher serverWebExchangeMatcher = spy(CsrfWebFilter.DEFAULT_CSRF_MATCHER);
```
rwinch, line 195: I don't think we need to test the `CsrfWebFilter` and the matcher here. I think we can just check `CsrfWebFilter.DEFAULT_CSRF_MATCHER` directly. If we need to test both, we should probably create separate tests.

web/src/main/java/org/springframework/security/web/server/csrf/CsrfWebFilter.java
parikshitdutta: Just to understand the required change, my understanding was the return from "getMethod" in question is time bounded, not a value for future, therefore it would be okay to use condition in map.

Please enlighten me if I am missing something.

web/src/test/java/org/springframework/security/web/server/csrf/CsrfWebFilterTests.java
parikshitdutta: The idea was to build MockServerWebExchange from NonStandard HTTP Request (i.e: "PURGE", "REVOKE" etc.), by using user defined string which is not a valid HTTP Method.

Now mocking without externalizing meaning I have to use get, post or other defined builders in MockServerHttpRequest (MockServerHttpRequest itself not extensible) or even if I try to do in different way, most likely i would end up writing redundant code.

Please share your thought.

web/src/main/java/org/springframework/security/web/server/csrf/CsrfWebFilter.java
rwinch: The issue is that if `r.getMethod()` is `null` then, it is a `.map(Mono.empty())` means that it is now a `Mono<Mono<Empty>>`. Instead you want just `Mono.empty()`.

web/src/test/java/org/springframework/security/web/server/csrf/CsrfWebFilterTests.java
rwinch: It looks like it is only being used once, so I think it can safely be consolidated without having redundant code.

web/src/main/java/org/springframework/security/web/server/csrf/CsrfWebFilter.java
parikshitdutta: totally make sense! thanks for clarifying.

parikshitdutta:
Hi @rwinch, Please review the changes as you find time, I will look forward to your feedback.

rwinch:

Issue #8459, opened 2020-05-05
spring-security-test: @WithMockOidcUser


nenaraab:
**Expected Behavior**
similar to [`@WithMockUser`](https://docs.spring.io/spring-security/site/docs/current/reference/html5/#test-method-withmockuser):

```java
    @Test
    @WithMockOidcUser(name = "Any@unknown.org" )
    public void authenticateWithoutPermission_200() throws Exception {
        mockMvc.perform(get("/authenticate"))
                .andExpect(status().isOk());
    }
```

**Current Behavior**
I'm unable to use Spring WebMVC tests for my secured methods and those methods that expects the security context to be filled. They require the user id (name) from the Oidc token also to perform authorization checks with `@PreAuthorize`. 

**Context**
My application requires a `OidcUser` as AuthenticationPrincipal and uses [Method Security](https://docs.spring.io/spring-security/site/docs/current/reference/html5/#method-security-expressions).

eleftherias:
@nenaraab This functionality is available as of Spring Security 5.3.
Check our [this section](https://docs.spring.io/spring-security/site/docs/5.3.2.RELEASE/reference/html5/#testing-oidc-login) of the reference documentation.

rwinch:
@eleftherias This functionality is a little different given it is annotation based. We can leave this closed as a duplicate of the PR gh-8461 There is some discussion on that ticket as to why we are waiting on the issue.

nenaraab:
Hi @eleftherias 

thanks a lot for the hint with the `oidcLogin` Mock MVC post processor!

Honestly i did not find it, when I looked for spring test utilities... but actually i was looking for a testing approach for methods that were secured with custom `@PreAuthorize` annotations and `@WithMockUser` didn't worked.

So, in my simple demo app i was able to leverage your suggested aproach, e.g.:

```java
@Test
//@WithMockOidcUser(name = "Alice_salesOrders@test.com", authorities = {"read:salesOrders"})
public void readWith_Alice_salesOrders_200() throws Exception {
        mockMvc.perform(get("/salesOrders")
                    .with(oidcLogin()
                        .idToken(token -> token.claim("sub", "Alice_salesOrders@test.com"))
                        .authorities(new SimpleGrantedAuthority("read:salesOrders"))))
                .andExpect(status().isOk());
}
```
Not nice, but i think this becomes even nicer with an `RequestPostProcessor`... :-)

[...]

Issue #7618, opened 2019-11-04
Add oidcLogin MockMvc Test Support


jzheaux:
Related to #6634.

It'd be handy to have test support for `oidcLogin`:

```java
this.mvc.perform(get("/").with(oidcLogin()));
```
or
```java
this.mvc.perform(get("/")
    .with(oidcLogin().authorities(new SimpleGrantedAuthority("SCOPE_read")));
```

The result would be an `OAuth2AuthenticationToken` as well as an `OAuth2AuthorizedClient`.

Ideally, tests would work with controllers using `@AuthenticatedPrincipal` as well as `@RegisteredOAuth2AuthorizedClient` annotations.

Issue #6557, opened 2019-02-24
Provide with an OAuth2 aware MockMvc


ch4mpy:
### Summary

Authorization header has to be added to each and every request when test involves any other OAuth2 authentication than Anonymous. This could be automatic.

### Actual Behavior

If Authorization header is forgotten, authentication context remains anonymous, whatever I try to set it with @WithMockUser or alike

### Expected Behavior

Authorization header is added if an authentication context is set for test-case

### Version

All versions

### Sample

As can be seen on [this lib](https://github.com/ch4mpy/spring-rest-oauth/tree/master/spring-rest-oauth-common/spring-rest-oauth-common-test) I solved this [proxying MockMvc request builder](https://github.com/ch4mpy/spring-rest-oauth/blob/master/spring-rest-oauth-common/spring-rest-oauth-common-test/src/main/java/com/c4soft/sro/common/test/OAuth2MockMvcHelper.java) to add a specific Authorization header I can intercept with a mocked token store, each time an `OAuth2Authentication` is present in security context.

I also created tooling to set-up such test Oauth2 authentication (`@WithMockOAuth2Client` & `@WithMockOAuth2User`) and a parent class for OAuth2 secured controllers tests (`OAuth2ControllerTest` which configures token service mock to intercept the special Authorization header), but this will be the object of coming cards.

Overall result in [controller unit tests](https://github.com/ch4mpy/spring-rest-oauth/blob/master/spring-rest-oauth-api-parent/spring-rest-oauth-api-uaa/src/test/java/com/c4soft/sro/api/uaa/web/UserControllerTest.java):
``` java
@WebMvcTest(UserController.class)
@Import({ResourceServerConfig.class})
@EnableSpringDataWebSupport
public class UserControllerTest extends OAuth2ControllerTest {

    @MockBean
    UserRepository userRepo;

    @Test
    @WithMockOAuth2User(user = @WithMockUser(username = "admin", authorities = "READ_USERS"))
    public void whenAuthenticatedWithReadUserPrivilegeThenListUsersReturnsUsersPage() throws Exception {
        final List<User> users = Arrays.asList(admin, user);
        when(userRepo.findAll(any(Pageable.class))).thenAnswer(invocation ->
                new PageImpl<>(users, (Pageable) invocation.getArguments()[0], users.size()));

        api.get("/users/")
                .andExpect(status().isOk())
                .andExpect(jsonPath("$._embedded.elements", hasSize(users.size())))
[...]

PR #8461, opened 2020-05-05
spring-security-test @WithMockOidcuser gh-8459

nenaraab:
Adds `@WithMockOidcUser` to `spring-security-test` library that allows you to fill the SecurityContext properly when using `oauth2Login` as I was used to with `@WithMockUser`

Example Web Mvc test:
```java
@Test
@WithMockOidcUser(name = "Alice_salesOrders@test.com", authorities = {"read:salesOrders"})
public void readWith_Alice_salesOrders_200() {
        mockMvc.perform(get("/salesOrders"))
                .andExpect(status().isOk());
}
```
Example **other Spring Tests**:
```java
@Autowired
DataService dataService;
    
@Test
@WithMockOidcUser(authorities = {"Admin"})
public void accessService_authorizedUser() {
        String data = dataService.readSensitiveData();
        assertFalse(data.isEmpty());
}
```

### Github issue
Find further details discussed here: #https://github.com/spring-projects/spring-security/issues/8459

### DIFF
--- a/test/src/main/java/org/springframework/security/test/context/support/WithMockOidcUser.java
+++ b/test/src/main/java/org/springframework/security/test/context/support/WithMockOidcUser.java
@@ -33,13 +33,14 @@ import java.lang.annotation.*;
  * <ul>
  * <li>The Authentication that is populated in the {@link SecurityContext} is of type {@link OAuth2AuthenticationToken}.</li>
  * <li>The principal on the Authentication is Spring Security’s User object of type {@code OidcUser}.</li>
- * <li>The default User has the username of "user", {@link #value()} or {@link #name()} and does not have to exist.</li>
- * <li>The default User has and a single GrantedAuthority {@link GrantedAuthority} or those that are specified by {@link #authorities()}.</li>
+ * <li>The default User has the user name "user". You can overwrite it the name with {@link #value()} or {@link #name()}.</li>
+ * <li>The default User has "ROLE_USER" and "SCOPE_openid" as {@link GrantedAuthority}.
+ * You can overwrite them with {@link #scopes()} or {@link #authorities()}. </li>
  * </ul>
  *
  * @author Nena Raab
  * @see WithUserDetails
- * @since 5.3
+ * @since 5.4
  */
 @Target({ ElementType.METHOD, ElementType.TYPE })
 @Retention(RetentionPolicy.RUNTIME)
@@ -47,7 +48,6 @@ import java.lang.annotation.*;
 @Documented
 @WithSecurityContext(factory = WithMockOidcUserSecurityContextFactory.class)
 public @interface WithMockOidcUser {
-	String DEFAULT_SCOPE = "SCOPE_openid";
 
 	/**
 	 * Convenience mechanism for specifying the username. The default is "user". If
@@ -57,7 +57,7 @@ public @interface WithMockOidcUser {
 	String value() default "user";
 
 	/**
-	 * The user name oder user id (subject) to be used. Note that {@link #value()} is a synonym for
+	 * The user name or user id (subject) to be used. Note that {@link #value()} is a synonym for
 	 * {@link #name()}, but if {@link #name()} is specified it will take
 	 * precedence.
 	 * @return
@@ -66,33 +66,36 @@ public @interface WithMockOidcUser {
 
 	/**
 	 * <p>
-	 * The authorities, that should be mapped to {@code GrantedAuthority}.
-	 * The default is "SCOPE_openid". A {@link GrantedAuthority} will be created for each value.
+	 * The scopes that should be mapped to {@code GrantedAuthority}.
+	 * The default is "openid". Each value in scopes gets prefixed with "SCOPE_"
+	 * and added to the list of {@link GrantedAuthority}.
 	 * </p>
-	 **
+	 * <p>
+	 * If {@link #authorities()} is specified this property cannot be changed from the default.
+	 * </p>
+	 *
 	 * @return
 	 */
-	String[] authorities() default { "SCOPE_openid" };
+	String[] scopes() default { "openid" };
 
 	/**
-	 * The name of the OIDC token claim that contains the subject identifier that identifies the End-User.
-	 * The default is "sub".
+	 * <p>
+	 * The authorities that should be mapped to {@code GrantedAuthority}.
+	 * </p>
+	 *
+	 * <p>
+	 * If this property is specified then {@link #scopes()} is not used. This differs from
+	 * {@link #scopes()} in that it does not prefix the values passed in automatically.
+	 * </p>
 	 * @return
 	 */
-	String nameTokenClaim() default "sub";
-
-	/**
-	 * The password to be used. The default is "clientId".
-	 * @return
-	 */
-	String clientId() default "clientId";
+	String[] authorities() default { };
 
 	/**
 	 * Determines when the {@link SecurityContext} is setup. The default is before
 	 * {@link TestExecutionEvent#TEST_METHOD} which occurs during
 	 * {@link org.springframework.test.context.TestExecutionListener#beforeTestMethod(TestContext)}
 	 * @return the {@link TestExecutionEvent} to initialize before
-	 * @since 5.1
 	 */
 	@AliasFor(annotation = WithSecurityContext.class)
 	TestExecutionEvent setupBefore() default TestExecutionEvent.TEST_METHOD;
--- a/test/src/main/java/org/springframework/security/test/context/support/WithMockOidcUser.java
+++ b/test/src/main/java/org/springframework/security/test/context/support/WithMockOidcUser.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.test.context.support;
+
+import org.springframework.core.annotation.AliasFor;
+import org.springframework.security.core.GrantedAuthority;
+import org.springframework.security.core.context.SecurityContext;
+import org.springframework.security.oauth2.client.authentication.OAuth2AuthenticationToken;
+import org.springframework.security.test.context.support.*;
+import org.springframework.test.context.TestContext;
+import org.springframework.test.web.servlet.MockMvc;
+
+import java.lang.annotation.*;
+
+/**
+ * When used with {@link WithSecurityContextTestExecutionListener} this annotation can be
+ * added to a test method to emulate running with a mocked user. In order to work with
+ * {@link MockMvc} The {@link SecurityContext} that is used will have the following
+ * properties:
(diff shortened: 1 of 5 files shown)


rwinch:
Thanks for the PR @nenaraab!

When we did gh-7618, I think we avoided annotation support for OIDC because it was not very easy for users to customize the `OidcUser` and `OAuth2AuthenticationToken` objects (the attributes are rich objects which cannot easily be modified by an annotation). Instead, we felt that user's could set `SecurityContextHolder` directly for method support or use the [`oidcLogin` support for `MockMvc`](https://docs.spring.io/spring-security/site/docs/current/reference/html5/#testing-oidc-login) style testing.

Given you find value in this, it is worth us reconsidering. However, I want to ensure that we remember all the concerns we had before we merge this. I'm assigning this to @jzheaux since he worked on the test support for some of the other OAuth bits and he might remember better what we discussed. As a heads up he is out on PTO so his response will be delayed.

Thanks again for your PR!

nenaraab:
Hi @rwinch and @jzheaux, that's cool, thanks for initial feedback and reconsidering.

In our context the usage of `@WithMockOidcUser` is quite meaningful and elegant :-), have a look at this sample:
https://github.com/SAP/cloud-security-xsuaa-integration/blob/cas/samples/spring-security-cas/src/test/java/com/sap/cloud/security/samples/SalesOrderControllerTest.java

Btw, to make the configuration of `@WithMockOidcUser` more generic, there is the option to enhance `@WithMockOidcUser` with custom claims, as documented here:
https://stackoverflow.com/questions/13569079/can-java-annotation-have-complex-return-type-like-hashmap

Enjoy your day!

rwinch:
Here is a link to some of our previous discussions around annotation based testing with more complex objects https://github.com/spring-projects/spring-security/issues/6557

test/src/main/java/org/springframework/security/test/context/support/WithMockOidcUser.java
```diff
@@ -0,0 +1,99 @@
...
+	 * @return
+	 */
+	String[] authorities() default { "SCOPE_openid" };
+
+	/**
+	 * The name of the OIDC token claim that contains the subject identifier that identifies the End-User.
+	 * The default is "sub".
+	 * @return
+	 */
+	String nameTokenClaim() default "sub";
```
jzheaux, line 82: Generally speaking, apps need to know what the principal's name is, not which claim it's in. For that reason, I'd recommend leaving this out.

test/src/main/java/org/springframework/security/test/context/support/WithMockOidcUserSecurityContextFactory.java
```diff
@@ -0,0 +1,91 @@
...
+		String userId = StringUtils.hasLength(withUser.name()) ? withUser
+				.name() : withUser.value();
+		if (userId == null) {
+			throw new IllegalArgumentException(withUser
+					+ " cannot have null userId on both userId and value properties");
+		}
+
+		List<GrantedAuthority> grantedAuthorities = new ArrayList<>();
+		for (String authority : withUser.authorities()) {
+			grantedAuthorities.add(new SimpleGrantedAuthority(authority));
```
jzheaux, line 55: To align this with the `OidcUser` that `OidcUserService` creates, I think [an `OidcUserAuthority` should also be added](https://github.com/spring-projects/spring-security/blob/master/test/src/main/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestPostProcessors.java#L1553).

test/src/main/java/org/springframework/security/test/context/support/WithMockOidcUserSecurityContextFactory.java
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	private class OidcIdTokenFactory {
+		private Map<String, Object> claims = new HashMap<>();
+		final Instant expiredAt = new GregorianCalendar().toInstant().plusSeconds(600);
+		final Instant issuedAt = new GregorianCalendar().toInstant().minusSeconds(3);
+
+		OidcIdTokenFactory(String userId, String clientId, String userIdClaimName) {
+			claims.put("client_id", clientId); // mandatory
+			claims.put("iat", issuedAt.getEpochSecond());
```
jzheaux, line 77: It might be easier to maintain if these used the claim names defined in `IdTokenClaimNames`.

test/src/main/java/org/springframework/security/test/context/support/WithMockOidcUserSecurityContextFactory.java
```diff
@@ -0,0 +1,91 @@
...
+				principal, principal.getAuthorities(), withUser.clientId());
+
+		SecurityContext context = SecurityContextHolder.createEmptyContext();
+		context.setAuthentication(authentication);
+		return context;
+	}
+
+	private class OidcIdTokenFactory {
+		private Map<String, Object> claims = new HashMap<>();
+		final Instant expiredAt = new GregorianCalendar().toInstant().plusSeconds(600);
```
jzheaux, line 72: Is there a reason that you are using `new GregorianCalendar().toInstant()` instead of `Instant.now()`?

test/src/test/java/org/springframework/security/test/context/support/WithMockOidcUserSecurityContextTests.java
```diff
@@ -0,0 +1,105 @@
...
+		when(withUser.name()).thenReturn("");
+	}
+
+	@Test(expected = IllegalArgumentException.class)
+	public void userNameValueNullRaiseException() {
+		when(withUser.value()).thenReturn(null);
+		factory.createSecurityContext(withUser);
+	}
+
+	public void valueDefaultsUserIdWhenUserNameIsNull() {
```
jzheaux, line 54: Would you please re-arrange the method names so that they look like `valueWhenUserNameIsNullThenDefaultsUserId`? In new classes, the team has standardized on `methodNameWhenCircumstanceThenResult`.

test/src/main/java/org/springframework/security/test/context/support/WithMockOidcUser.java
```diff
@@ -0,0 +1,99 @@
...
+	 * @return
+	 */
+	String clientId() default "clientId";
+
+	/**
+	 * Determines when the {@link SecurityContext} is setup. The default is before
+	 * {@link TestExecutionEvent#TEST_METHOD} which occurs during
+	 * {@link org.springframework.test.context.TestExecutionListener#beforeTestMethod(TestContext)}
+	 * @return the {@link TestExecutionEvent} to initialize before
+	 * @since 5.1
```
jzheaux, line 95: Reminder to remove this `@since`, which probably came over from a copy-paste

test/src/main/java/org/springframework/security/test/context/support/WithMockOidcUser.java
```diff
@@ -0,0 +1,99 @@
...
+ * <ul>
+ * <li>The Authentication that is populated in the {@link SecurityContext} is of type {@link OAuth2AuthenticationToken}.</li>
+ * <li>The principal on the Authentication is Spring Security’s User object of type {@code OidcUser}.</li>
+ * <li>The default User has the username of "user", {@link #value()} or {@link #name()} and does not have to exist.</li>
+ * <li>The default User has and a single GrantedAuthority {@link GrantedAuthority} or those that are specified by {@link #authorities()}.</li>
+ * </ul>
+ *
+ * @author Nena Raab
+ * @see WithUserDetails
+ * @since 5.3
```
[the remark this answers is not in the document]
nenaraab: sure, thanks for the detailed feedback!

test/src/main/java/org/springframework/security/test/context/support/WithMockOidcUser.java
```diff
@@ -0,0 +1,99 @@
...
+
+	/**
+	 * <p>
+	 * The authorities, that should be mapped to {@code GrantedAuthority}.
+	 * The default is "SCOPE_openid". A {@link GrantedAuthority} will be created for each value.
+	 * </p>
+	 **
+	 * @return
+	 */
+	String[] authorities() default { "SCOPE_openid" };
```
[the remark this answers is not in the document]
nenaraab: My suggestion would be to use scopes instead of authorities, as authorities has Spring Security specific means (gets prefixed with "SCOPE_") and i can not see an advantage of having both, beside of the fact when someone has changed the prefix.

test/src/main/java/org/springframework/security/test/context/support/WithMockOidcUser.java
```diff
@@ -0,0 +1,99 @@
...
+	 * The default is "sub".
+	 * @return
+	 */
+	String nameTokenClaim() default "sub";
+
+	/**
+	 * The password to be used. The default is "clientId".
+	 * @return
+	 */
+	String clientId() default "clientId";
```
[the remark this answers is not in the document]
nenaraab: yes, for sure!

test/src/main/java/org/springframework/security/test/context/support/WithMockOidcUserSecurityContextFactory.java
```diff
@@ -0,0 +1,91 @@
...
+		final Instant issuedAt = new GregorianCalendar().toInstant().minusSeconds(3);
+
+		OidcIdTokenFactory(String userId, String clientId, String userIdClaimName) {
+			claims.put("client_id", clientId); // mandatory
```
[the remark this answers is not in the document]
nenaraab: No, its not required (any more) it can be removed. Thanks!

test/src/main/java/org/springframework/security/test/context/support/WithMockOidcUserSecurityContextFactory.java
nenaraab: 👍

test/src/main/java/org/springframework/security/test/context/support/WithMockOidcUser.java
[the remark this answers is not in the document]
jzheaux: No problem, I don't have strong feelings about it, and we can always wait for a community member to request it.

The rest is just for future reference:

> mutual exclusivity

Btw, while I see your point about mutual exclusivity, I don't see it as a problem as you do. The DSL allows `oauth2ResourceServer().jwt()` OR `oauth2ResourceServer().opaqueToken()`, for example.

> scopes only

agreed

> authorities only ... add "SCOPE_" prefix

The application only has to do this if it's using the "SCOPE_" prefix in its authorization rules, in which case the tester already knows about it. I don't see this as an issue.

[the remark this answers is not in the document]
nenaraab: then - lets provide both :-)

nenaraab:
@jzheaux 
I'm done with the changes, happy to receive further feedback, in case there are further findings :-)
Thanks again for your valuable review and explanations!

test/src/main/java/org/springframework/security/test/context/support/WithMockOidcUserSecurityContextFactory.java
nenaraab: The only thing it does now: it adds another `GrantedAuthority` with "ROLE_USER", not sure why required...

jzheaux:
@nenaraab thanks so much for this contribution and for responding to my change requests. After letting this sit for some time now, I feel like we don't want to add claims to the annotation. 

Without claims, it provides no information that isn't already available in `Authentication`:

```java
public String method(Authentication authentication) {
    String name = authentication.getName();
    Collection<GrantedAuthority> authorities = authentication.getAuthorities();
}
```

in which case tying the application to `OidcUser` is needless. I'd prefer not to simplify the use case where folks are tying their applications needlessly to `OidcUser`.

Given that, I'm going to close this, but please feel free to comment further if I've missed a use case of yours.

Issue #5758, opened 2018-08-31
No Spring-Security logging with Spring-Boot and WebFlux


kensiprell:
### Summary

I cannot get Spring-Security to log anything in a simple Spring-Boot WebFlux application. I  assume I must be doing something wrong because my searches have only turned up this unanswered [SO question](https://stackoverflow.com/questions/49877221/spring-security-5-logging-for-reactive-applications).

### Actual Behavior

No Spring-Boot logging entries appear in terminal window after running `./gradlew bootRun`.

### Expected Behavior

Detailed logging entries as if using MVC. I followed the same steps to reproduce as shown below, substituting "Web" for "Reactive Web," and it works as expected.

### Configuration

macOS 10.13.6
java version "1.8.0_181"

### Version

I tried Spring-Booot versions 2.0.4, 2.1.0.M2, and 2.1.0.SNAPSHOT.

### Sample

https://github.com/kensiprell/webflux-spring-security-demo

### Steps to Reproduce

1. Go to https://start.spring.io.

2. Choose Gradle, Kotlin or Java, and a Boot version.

3. Add 'Security' and 'Reactive Web' to the dependencies and generate the project.

4. Unarchive demo.zip and `cd demo/`.

5. Edit `./src/main/resources/application.properties` and add a line that should increase the logging level:
`logging.level.root=DEBUG`
or
`logging.level.org.springframework.security=DEBUG`

6. `./gradlew bootRun`

7. Open http://localhost:8080 in a browser.

rwinch:
You are not doing anything wrong. There really isn't any logging enabled within Spring Security's reactive support. Much of this has to do with the fact that the logging is likely a blocking operation. When I discussed this previously with the Reactor team they had said that there were thoughts around supporting reactive logging, but I just had not followed up on that. It seems like the recommendation is still to configuring logging to be async and that this is a limitation. We can and should add logging to Spring Security.

wtatum:
I think I have a good understanding of what makes this so challenging to resolve, but I'd like to quickly voice concern that this isn't being looked into more. There are some use cases that are inherently reactive in nature (a good example is Spring Cloud Gateway) but also strongly benefit from a traceable, auditable security framework.

[...]

PR #8504, opened 2020-05-08
Add debug logging to Reactive Web

mouellet:
Goal is to provide insight to devs on:
- Authentication & Authorization success/failures
- WebSession & SecurityContext
- Request matchers, cache & authn/authz flow

Fixes gh-5758

Please feel free to suggest better log messages (they mostly come from the servlet side)!


rwinch:
Thank you @mouellet! This is now merged into master. I believe this is a significant improvement on where things stand today. However, I do want to let you know that we are hoping to revisit our logging story and align with some of the logging changes Spring Framework did in 5.0 (i.e. more targeted logging). In the meantime, this is a drastic improvement to the code today. Thanks again!

Issue #7572, opened 2019-10-26
formLogin() does not work with REST Docs


ValtteriL:
### Summary

I was asked to open this issue here after Spring REST Docs team investigated my issue why REST Docs did not get configured with formLogin().

> `SecurityMockMvcRequestBuilders$FormLoginRequestBuilder`  does not implement `Mergeable` so the default configuration that is set up on the `MockMvc` instance is not applied. Crucially in the case of Spring REST Docs, this means that the `ConfigurerApplyingRequestPostProcessor` that applies the REST Docs context to the request is lost.

The original issue can be found here: https://github.com/spring-projects/spring-restdocs/issues/655

The following code causes `java.lang.IllegalStateException: REST Docs configuration not found. Did you forget to apply a MockMvcRestDocumentationConfigurer when building the MockMvc instance?`

```
@RunWith(SpringRunner.class)
@SpringBootTest
@AutoConfigureMockMvc
@AutoConfigureRestDocs
public class LoginLogoutTest {

    @Autowired
    private MockMvc mockMvc;

    @Test
    public void adminCanLoginLogout() throws Exception {
        mockMvc.perform(formLogin().user(TestConfig.ADMIN_USERNAME).password(TestConfig.PASSWORD))
            .andExpect(status().isOk())
            .andExpect(authenticated().withUsername(TestConfig.ADMIN_USERNAME))
            .andDo(document("login"));

        mockMvc.perform(logout())
            .andExpect(status().isOk())
            .andExpect(unauthenticated())
            .andDo(document("logout"));
    }
}
```

### Actual Behavior

On `mvn test` the code causes `java.lang.IllegalStateException: REST Docs configuration not found. Did you forget to apply a MockMvcRestDocumentationConfigurer when building the MockMvc instance?` on the line with `document("login")`.

### Expected Behavior

No errors. The REST Docs should be correctly configured.

### Version

All versions

### Sample

[restdocs-formlogin.zip](https://github.com/spring-projects/spring-security/files/3774584/restdocs-formlogin.zip)
This sample causes the same error on `mvn test`.

rhamedy:
Out of curiosity and the following comment 

> SecurityMockMvcRequestBuilders$FormLoginRequestBuilder does not implement Mergeable so the default configuration that is set up on the MockMvc instance is not applied. Crucially in the case of Spring REST Docs, this means that the ConfigurerApplyingRequestPostProcessor that applies the REST Docs context to the request is lost.

[...]

PR #8507, opened 2020-05-09
formLogin() does not work with REST Docs

dadikovi:
This commit adds a new test which fails right now, and also suggest a change.

Fixes gh-7572


dadikovi:
- I added a new test case to SecurityMockMvcRequestBuildersFormLoginTests.java, which replicates the behavior of REST docs (it fails now). I also added lots of comments with links to the original code parts in spring-restdocs.
- Also started to implement a naive fix for the issue. However, at this point this issue cannot be solved because an interface in spring-test should be extended. I added a few comments here as well which describes what kind of extension is needed.

Dear @rwinch, please double-check my findings if you have time. If you can confirm them, I will create PR to the related spring project as well, after it I will continue working on this PR until the test succeeds :) Thank you in advance!

test/src/main/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuilders.java
```diff
@@ -124,6 +137,22 @@ public LogoutRequestBuilder logoutUrl(String logoutUrl, Object... uriVars) {
 
 		private LogoutRequestBuilder() {
 		}
+
+		@Override public boolean isMergeEnabled() {
```
rwinch, line 141: Can you please put `@Override` on its own line to be consistent with Spring Security code conventions?

```suggestion
		@Override
		public boolean isMergeEnabled() {
```

test/src/main/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuilders.java
```diff
@@ -124,6 +137,22 @@ public LogoutRequestBuilder logoutUrl(String logoutUrl, Object... uriVars) {
...
+			return true;
+		}
+
+		@Override public Object merge( Object parent ) {
```
rwinch, line 145: Can you please put `@Override` on its own line and remove the extra space to be consistent with Spring Security code conventions?

```suggestion
		@Override
		public Object merge(Object parent ) {
```

test/src/main/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuilders.java
```diff
@@ -124,6 +137,22 @@ public LogoutRequestBuilder logoutUrl(String logoutUrl, Object... uriVars) {
...
+			if (parent == null) {
+				return this;
+			}
+			if (parent instanceof MockHttpServletRequestBuilder) {
```
rwinch, line 149: I think we can just check if it is an instance of `Mergeable` since that is what we are casting it to. The check for what we pass in is a `MockHttpServletRequestBuilder` will be done when we invoke `MockHttpServletRequestBuilder.merge`

test/src/main/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuilders.java
```diff
@@ -260,6 +298,25 @@ public FormLoginRequestBuilder acceptMediaType(MediaType acceptMediaType) {
...
+		public boolean isMergeEnabled() {
+			return true;
+		}
+
+		@Override
+		public Object merge( Object parent ) {
+			if (parent == null) {
+				return this;
+			}
+			if (parent instanceof MockHttpServletRequestBuilder) {
```
rwinch, line 313: I think we can just check if it is an instance of Mergeable since that is what we are casting it to. The check for what we pass in is a MockHttpServletRequestBuilder will be done when we invoke MockHttpServletRequestBuilder.merge

test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuildersFormLoginTests.java
```diff
@@ -82,6 +94,35 @@ public void customWithUriVars() {
...
+			((ConfigurableSmartRequestBuilder) defaultRequestBuilder).with(new MockPostProcessor());
+		}
+
+		if (defaultRequestBuilder != null && requestBuilder instanceof Mergeable ) {
+			requestBuilder = (RequestBuilder) ((Mergeable) requestBuilder).merge(defaultRequestBuilder);
+		}
+
+		MockHttpServletRequest request = requestBuilder.buildRequest(this.servletContext);
+
+		Assert.isInstanceOf(ConfigurableSmartRequestBuilder.class, requestBuilder);
```
rwinch, line 121: Please use assertj

test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuildersFormLoginTests.java
```diff
@@ -82,6 +94,35 @@ public void customWithUriVars() {
...
+	 */
+	@Test
+	public void postProcessorsAreMergedDuringMockMvcPerform() throws Exception {
+
+		RequestBuilder requestBuilder = formLogin()
+				.user("my-user")
+				.password("my-password")
+				.loginProcessingUrl("/my-path");
+
+		RequestBuilder defaultRequestBuilder = MockMvcRequestBuilders.get("/");
```
rwinch, line 110: If you assign this to `MockHttpServletRequestBuilder` we don't need to do the `instanceof` or casting to `ConfigurableSmartRequestBuilder` or `Mergeable`

test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuildersFormLoginTests.java
```diff
@@ -82,6 +94,35 @@ public void customWithUriVars() {
...
+		if (defaultRequestBuilder instanceof ConfigurableSmartRequestBuilder) {
+			((ConfigurableSmartRequestBuilder) defaultRequestBuilder).with(new MockPostProcessor());
+		}
+
+		if (defaultRequestBuilder != null && requestBuilder instanceof Mergeable ) {
```
rwinch, line 115: I don't think we need a null check here

test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuildersFormLogoutTests.java
```diff
@@ -71,4 +79,37 @@ public void customWithUriVars() {
...
+	 * spring-restdocs uses postprocessors to do its trick. It will work only if these are merged together
+	 * with our request builders. (gh-7572)
+	 * @throws Exception
+	 */
+	@Test
+	public void postProcessorsAreMergedDuringMockMvcPerform() throws Exception {
+
+		RequestBuilder requestBuilder = logout();
+
+		RequestBuilder defaultRequestBuilder = MockMvcRequestBuilders.get("/");
```
rwinch, line 92: If you assign this to `MockHttpServletRequestBuilder` we don't need to do the `instanceof` or casting to `ConfigurableSmartRequestBuilder` or `Mergeable`

test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuildersFormLogoutTests.java
```diff
@@ -71,4 +79,37 @@ public void customWithUriVars() {
...
+		if (defaultRequestBuilder instanceof ConfigurableSmartRequestBuilder ) {
+			((ConfigurableSmartRequestBuilder) defaultRequestBuilder).with(new MockPostProcessor());
+		}
+
+		if (defaultRequestBuilder != null && requestBuilder instanceof Mergeable ) {
```
rwinch, line 97: I don't think we need this null check

test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuildersFormLogoutTests.java
```diff
@@ -71,4 +79,37 @@ public void customWithUriVars() {
...
+			requestBuilder = (RequestBuilder) ((Mergeable) requestBuilder).merge(defaultRequestBuilder);
+		}
+
+		MockHttpServletRequest request = requestBuilder.buildRequest(this.servletContext);
+
+		Assert.isInstanceOf(ConfigurableSmartRequestBuilder.class, requestBuilder);
```
rwinch, line 103: Please use assertj

test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuildersFormLogoutTests.java
```diff
@@ -71,4 +79,37 @@ public void customWithUriVars() {
...
+		Assert.isTrue(request.equals(((SmartRequestBuilder) requestBuilder).postProcessRequest(request)),
```
rwinch, line 104: Please use assertj

test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuildersFormLoginTests.java
```diff
@@ -82,6 +94,35 @@ public void customWithUriVars() {
...
+			requestBuilder = (RequestBuilder) ((Mergeable) requestBuilder).merge(defaultRequestBuilder);
+		}
+
+		MockHttpServletRequest request = requestBuilder.buildRequest(this.servletContext);
+
+		Assert.isInstanceOf(ConfigurableSmartRequestBuilder.class, requestBuilder);
+		Assert.isTrue(request.equals(((SmartRequestBuilder) requestBuilder).postProcessRequest(request)),
```
rwinch, line 122: Please use assertj

test/src/main/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuilders.java
```diff
@@ -260,6 +298,25 @@ public FormLoginRequestBuilder acceptMediaType(MediaType acceptMediaType) {
...
 		}
+
+
+		@Override
+		public boolean isMergeEnabled() {
+			return true;
+		}
+
+		@Override
+		public Object merge( Object parent ) {
```
rwinch, line 309: Remove extra space for consistent formatting

```suggestion
		public Object merge(Object parent ) {
```

test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuildersFormLoginTests.java
```diff
@@ -82,6 +95,32 @@ public void customWithUriVars() {
...
+	@Test
+	public void postProcessorsAreMergedDuringMockMvcPerform() throws Exception {
+
+		RequestBuilder requestBuilder = formLogin()
+				.user("my-user")
+				.password("my-password")
+				.loginProcessingUrl("/my-path");
+
+		MockHttpServletRequestBuilder defaultRequestBuilder = MockMvcRequestBuilders.get("/");
+		defaultRequestBuilder.with(new MockPostProcessor());
```
rwinch, line 112: Looking at this test again, I think some of this is a little incidentally verified. For example, if I remove the line above this test still passes. That means that it isn't verifying that the postProcessor is merged. I'm thinking that it might be better to use `MockMvc` to perform the validation. Something like this:

```java
	@Test
	public void postProcessorsAreMergedDuringMockMvcPerform() throws Exception {
		RequestPostProcessor postProcessor = mock(RequestPostProcessor.class);
		when(postProcessor.postProcessRequest(any())).thenAnswer(i -> i.getArgument(0));
		MockMvc mockMvc = MockMvcBuilders.standaloneSetup(new Object())
				.defaultRequest(MockMvcRequestBuilders.get("/").with(postProcessor))
				.build();

		mockMvc.perform(formLogin());

		verify(postProcessor).postProcessRequest(any());
	}
	
```

Thoughts?

test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuildersFormLogoutTests.java
```diff
@@ -71,4 +80,34 @@ public void customWithUriVars() {
...
+	 * with our request builders. (gh-7572)
+	 * @throws Exception
+	 */
+	@Test
+	public void postProcessorsAreMergedDuringMockMvcPerform() throws Exception {
+
+		RequestBuilder requestBuilder = logout();
+
+		MockHttpServletRequestBuilder defaultRequestBuilder = MockMvcRequestBuilders.get("/");
+		defaultRequestBuilder.with(new MockPostProcessor());
```
rwinch, line 94: Similar to the other test, this doesn't seem to be doing anything since if we remove the line above the test will still pass. I think we should likely use `MockMvc` here as I suggested above.

test/src/main/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuilders.java
```diff
@@ -124,6 +137,24 @@ public LogoutRequestBuilder logoutUrl(String logoutUrl, Object... uriVars) {
 
 		private LogoutRequestBuilder() {
 		}
+
```
rwinch, line 140: Let's make these public methods above the private constructor to follow Spring Security's conventions

test/src/main/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuilders.java
```diff
@@ -260,6 +300,25 @@ public FormLoginRequestBuilder acceptMediaType(MediaType acceptMediaType) {
 
 		private FormLoginRequestBuilder() {
 		}
+
```
rwinch, line 303: Let's make these public methods above the private constructor to follow Spring Security's conventions

test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuildersFormLoginTests.java
dadikovi: Your approach seems to be much better, I will use this :)

dadikovi:
Executed the requested changes, sorry for styling issues, this time executed the checkstyle build, hope there wont be other problems. The previous travis build contained test failures as well.
If they are still there I will check them too.

dadikovi:
Yeah, well, it looks like I broke a lots of tests with this change. I'll do some research and will fix them. :)

test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuildersFormLoginTests.java
```diff
@@ -82,6 +101,23 @@ public void customWithUriVars() {
...
+	@Test
+	public void postProcessorsAreMergedDuringMockMvcPerform() throws Exception {
+		RequestPostProcessor postProcessor = mock(RequestPostProcessor.class);
+		when(postProcessor.postProcessRequest(any())).thenAnswer(i -> i.getArgument(0));
+		MockMvc mockMvc = MockMvcBuilders.standaloneSetup(new Object())
+				.defaultRequest(MockMvcRequestBuilders.get("/").with(postProcessor))
+				.build();
+
+		mockMvc.perform(formLogin());
+		verify(postProcessor).postProcessRequest(any());
```
rwinch, line 118: It seems like `formLogin()` is not working quite correctly when merging. For example, the following does not work correctly:

```suggestion
		MvcResult mvcResult = mockMvc.perform(formLogin()).andReturn();

		assertThat(mvcResult.getRequest().getMethod()).isEqualTo(HttpMethod.POST.name());
		verify(postProcessor).postProcessRequest(any());
```

I think the problem is that merge should always return `this` and we should flip `buildRequest` to pass in `this.mergeable` into `builder` (which needs to be an instance of `MockHttpServletRequestBuilder`).

dadikovi: I can see that the problem here that after this code:

```
if (this.mergeable != null) {
  builder = (RequestBuilder) this.mergeable.merge(builder);
}
```

the request method will be GET instead of POST (because the mergeable requestBuilder is type of GET). The URL changed as well. 
As I can see, there is no way to change the method and the URL of a requestBuilder (only in the constructor), so this approach that we use builder constructed by mergeable, and then we modify the necessary attributes wont work. :/
So it should be done the other way: we should get the postprocessors from the mergable and put them into our requestBuilder. However, this also cannot be done, since there is no API to get the postprocessors from the mergable (see comments from my original commit in this PR).
So my understanding is that to achive this some modifications are needed outside of spring-security as well:
- either MockHttpServletRequestBuilder should contain setter methods for url and method fields
- or the ConfigurableSmartRequestBuilder should contain getter method for the postprocessors.

dadikovi: @rwinch which is the better solution?

rwinch: This commit https://github.com/rwinch/spring-security/commit/695a9668b374740f5c7161265f4e112797539699 seems to solve the problem for me. The branch is https://github.com/rwinch/spring-security/commits/gh-7572

dadikovi: @rwinch I tried you version, but this test: postProcessorsAreMergedDuringMockMvcPerform still fails.

So its time to shine for me, and to show the first real contribution under this PR :D

I had to add this line at the end of buildRequest method (before the return statement), and so the merged postprocessors are called now:

`loginRequest.postProcessRequest(request);`

I sent another two commits, I will check if travis approved it (if not, I'll fix it)

dadikovi:
@rwinch finally all tests pass. :)

test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuildersFormLoginTests.java
```diff
@@ -82,6 +89,23 @@ public void customWithUriVars() {
...
+	 */
+	@Test
+	public void postProcessorsAreMergedDuringMockMvcPerform() throws Exception {
+		RequestPostProcessor postProcessor = mock(RequestPostProcessor.class);
+		when(postProcessor.postProcessRequest(any())).thenAnswer(i -> i.getArgument(0));
+		MockMvc mockMvc = MockMvcBuilders.standaloneSetup(new Object())
+				.defaultRequest(MockMvcRequestBuilders.get("/").with(postProcessor))
+				.build();
+
+		mockMvc.perform(formLogin());
```
rwinch, line 105: Could you add the assertion that it is a post back? I think we should also assert the other parts of the request that formLogin sets.

test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuildersFormLogoutTests.java
```diff
@@ -71,4 +79,22 @@ public void customWithUriVars() {
...
+	 */
+	@Test
+	public void postProcessorsAreMergedDuringMockMvcPerform() throws Exception {
+		RequestPostProcessor postProcessor = mock(RequestPostProcessor.class);
+		when(postProcessor.postProcessRequest(any())).thenAnswer(i -> i.getArgument(0));
+		MockMvc mockMvc = MockMvcBuilders.standaloneSetup(new Object())
+				.defaultRequest(MockMvcRequestBuilders.get("/").with(postProcessor))
+				.build();
+
+		mockMvc.perform(logout());
```
rwinch, line 95: Similar to formLogin we should have assertions for what logout contributes to the request.

dadikovi:
Thank you for the feedback, added the missing assertions :)

dadikovi:
@rwinch added the additional assertions. Please note that I had to modify the visibility of TestCsrfTokenRepository's constructor from private to package private.

test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuildersFormLoginTests.java
```diff
@@ -107,9 +112,27 @@ public void postProcessorsAreMergedDuringMockMvcPerform() throws Exception {
...
+		assertThat(mvcResult.getRequest().getHeader("Accept"))
+				.isEqualTo(MediaType.toString(Arrays.asList(MediaType.APPLICATION_FORM_URLENCODED)));
+		assertThat(mvcResult.getRequest().getParameter("username")).isEqualTo("user");
+		assertThat(mvcResult.getRequest().getParameter("password")).isEqualTo("password");
+		assertThat(mvcResult.getRequest().getRequestURI()).isEqualTo("/login");
+		assertCsrfToken(mvcResult.getRequest());
 		verify(postProcessor).postProcessRequest(any());
 	}
 
+	private void assertCsrfToken(MockHttpServletRequest request) {
```
rwinch, line 124: I think this assertion is a bit different than some of our other tests. We really only need to ensure that the HTTP parameter `_csrf` exists and is not empty.

test/src/test/java/org/springframework/security/test/web/servlet/request/SecurityMockMvcRequestBuildersFormLogoutTests.java
```diff
@@ -96,8 +102,23 @@ public void postProcessorsAreMergedDuringMockMvcPerform() throws Exception {
...
 		assertThat(mvcResult.getRequest().getMethod()).isEqualTo(HttpMethod.POST.name());
-
+		assertThat(mvcResult.getRequest().getHeader("Accept"))
+				.isEqualTo(MediaType.toString(Arrays.asList(MediaType.TEXT_HTML, MediaType.ALL)));
+		assertThat(mvcResult.getRequest().getRequestURI()).isEqualTo("/logout");
+		assertCsrfToken(mvcResult.getRequest());
 		verify(postProcessor).postProcessRequest(any());
 	}
 
+	private void assertCsrfToken(MockHttpServletRequest request) {
```
rwinch, line 112: Same as above. I think this assertion is a bit different than some of our other tests. We really only need to ensure that the HTTP parameter `_csrf` exists and is not empty.

dadikovi:
@rwinch Thanks for the feedback, I simplified the assertion of the csrf token. :)
Well, its really a fun for me to get in touch with the spring community, and I've already learned some new stuff (especially about unit testing). So thank _you_ for reviewing my iterations so patiently. (even if probably it would be more cost-effective for you just implement such tasks without involving any contributors) :)

rwinch:
Thanks for the PR @dadikovi This has been merged via 88028d82edc2702209da3a52663004608b3c287e

Issue #8128, opened 2020-03-17
 Blocking call to UUID.randomUUID in WebSessionServerCsrfTokenRepository (blockhound)


cbornet:
### Summary

Detected by blockhound: `WebSessionServerCsrfTokenRepository` and `CookieServerCsrfTokenRepository` make blocking calls to `UUID.randomUUID` when generating the token.

It would be nice to have a non-blocking SecureRandom to solve this.
It can of course be offloaded to the boundedElastic scheduler but that looks sub optimal.

https://github.com/spring-projects/spring-security/blob/747d8817cbadc307f7407c26fc88b2ff63c37149/web/src/main/java/org/springframework/security/web/server/csrf/WebSessionServerCsrfTokenRepository.java#L112

### Version

5.2.2.RELEASE

rwinch:
Thanks for the report @cbornet!

I'm not aware of a non-blocking secure random source. Are you? If we don't have a non-blocking secure random source then I agree our best bet is to use the boundedElastic.

Would you be interested in submitting a PR?

cbornet:
> I'm not aware of a non-blocking secure random source. Are you?

No, I'm not either. I guess even reading /dev/random with NIO is offloading to a thread-pool. Would be nice to have it in Java one day though.

I'll do the PR for boundedElastic, no problem

cbornet:
I want to put a `publishOn` in generateToken but I don't find a good place
I coud do 
```java
	@Override
	public Mono<CsrfToken> generateToken(ServerWebExchange exchange) {
		return Mono.just(exchange)
			.publishOn(Schedulers.boundedElastic())
			.fromCallable(() -> createCsrfToken());
	}
```
but maybe there's a better way without wrapping exchange (which is not used) ?
```

rwinch:
That looks correct to me.

rwinch:
@cbornet Are you still interested in submitting this?

cbornet:
Yes. Sorry I've been quite busy. I'll do it this week.

PR #8534, opened 2020-05-15
Create the CSRF token on the bounded elactic scheduler

cbornet:
The CSRF token is created with a call to UUID.randomUUID which is blocking.
This change ensures this blocking call is done on the bounded elastic scheduler which supports blocking calls.

Fixes gh-8128


rwinch:
Thanks for the report! A fix is now merged into master :) I've also backported as far back as 5.0.x

Issue #7621, opened 2019-11-05
Object ID Identicy conversion to long fails on old schema


nucatus:
### Summary

In case of an old ACL schema where the `object_id_identity` is of type _bigint_ instead of the current _varchar(36)_, the conversion of the identity to a known type fails. 

### Actual Behavior

Conversion of ACL object ID identity fails.

### Expected Behavior

Conversion of the ACL object ID identity should succeed to a known type, or long if no `class_id_type` is provided.

### Version

5.2.x

rwinch:
Thanks for the report. Can you provide a sample of that reproduces this issue?

nucatus:
Well, I was just looking at the history of the `aclSchema` and noticed that the type of the `object_id_identity` field changed from `BIGINT NOT NULL` to `VARCHAR(36) NOT NULL` when support for non-Long IDs was added.

In our project we have a schema where the type is `BIGINT UNSIGNED NOT NULL`. When data is extracted from the ResultSet, the mysql driver will convert this type to a BigInteger. In the previous version, this was not an issue, because the conversion to Long was carried on the else branch of the `hasConversionService()`. This changed with this [update](https://github.com/spring-projects/spring-security/commit/918a4cd3230c7dcd273a47ada5288c535ffcb030#diff-bdb889847e56650fc7c52f9de584ba22R121), where this check is buggy, because `canConvertFromStringTo(Long.class)` will always return true, no matter the source type, and it's passing the call to the `convertService`.

Indeed, this works with the original schema, but in our case, where our schema is slightly different and the type that the mysql driver returns is a `BigInteger`, it fails because it is looking to convert it with the `conversionService`, instead of falling on the else branch where the conversation works for other, unknown, types.

igorpele:
Hi, I would like to work on this. As far as I understand, it only has to be verified that the conversion from String to Long is executed only if the identifier is an instance of String.

dadikovi:
@igorpele do you still want to work on this issue? (I'm just asking because I'm looking for an issue to work, but I can find other one off course)

igorpele:
@dadikovi as I got no response on this issue, you are welcome to take it.

PR #8538, opened 2020-05-16
Object ID Identicy conversion to long fails on old schema

dadikovi:
This change fixed a bug which tried to convert non-string object as string

Fixes gh-7621


dadikovi:
So the issue was here that in case of an old ACL schema the identity is converted to BigInteger by mysql driver, and handling of this datatype was buggy.

PR #8542, opened 2020-05-18
Security issue: mentioning the default strength of `BCryptPasswordEncoder`

islamazhar:
Hi, 

As mentioned in the Spring security doc on `BCryptPasswordEncoder`

> The strength of   BCryptPasswordEncoder should be tuned to take about 1 second to verify a password on your system.

However, the default implementation of `BCryptPasswordEncoder` uses a **default strength of 10**. 
On my system (Intel Core i5 CPU-1.60Hz 8 GM RAM), I found that the default implementation takes around **220~250 ms** to verify a password which is clearly way less than 1 second lower limit. 

I think it should be worth **mentioning the default strength of BCryptPasswordEncoder** since all of the Spring projects I have worked with developers tend not to change the default strength which according to the Spring security doc is not secure.  

This can make the developers using the default strength of `BCryptPasswordEncoder` more conscious about setting a correct secure strength.


islamazhar:
I searched a bit more and found some interesting links which may be useful for adding a more appropriate context to the issue. 
 
- The python `bcrypt` library mentions about the default strength [here](https://pypi.org/project/bcrypt/) in *Adjustable Work Factor* section. 
- The NodeJS `bcrypt` documentation mentions about the strength factor [here](https://github.com/kelektiv/node.bcrypt.js#a-note-on-rounds). 
- An interesting [answer](https://security.stackexchange.com/questions/17207/recommended-of-rounds-for-bcrypt) on *security stack exchange* about the number of recommended rounds for `bcrypt`.

rwinch:

PR #8548, opened 2020-05-18
Configuration defaults to SessionRegistry bean

candrews:
If a SessionRegistry is necessary, check for one in the ApplicationContext before creating one.

See https://github.com/spring-projects/spring-session/issues/1629


rwinch:
Thanks for the Pull Request. Could you please add a test that verifies this works when there is one `SessionRegistry` bean (should use that bean) and more than one `SessionRegistry` bean (should not break with `NoUniqueBeanDefinitionException`.

candrews:
> Could you please add a test that verifies this works when there is one `SessionRegistry` bean (should use that bean) and more than one `SessionRegistry` bean (should not break with `NoUniqueBeanDefinitionException`.

I've added the two tests as requested.

Thanks again!

rwinch:

PR #8580, opened 2020-05-21
Replacing md5 with sha256 token based remember me services

islamazhar:
This is related to issue #8580


rwinch:
Thanks for the PR. We need to remain passive, so we would want to allow configuring the algorithm vs changing it. We could consider changing it by default in Spring Security 6 though.

An alternative would be to do something like we did with passwords and have a prefix that specifies the algorithm that is being used. If no algorithm is specified it would default to MD5

rwinch:
Closing because the requested changes have not been made. Please feel free to discuss on the related issue before opening a new PR.

Issue #5867, opened 2018-09-18
Reactive equivalents for security expression handling and url-based access control


danielfernandez:
### Summary

In the new `thymeleaf-extras-springsecurity5` I'm trying to make all the functionality already existing for Servlet-based applications available also for WebFlux ones.

But I'm unable to find in the reactive side of Spring Security any equivalents for the functionality providen by `SecurityExpressionHandler<FilterInvocation>` or `WebInvocationPrivilegeEvaluator` in Servlet-based web applications. These allow Thymeleaf to evaluate Spring Security expressions such as `hasRole(x)`, and also perform url-based access control (privilege evaluator).

Also, digging into Spring Security's code, I cannot find any infrastructure for quickly developing reactive equivalents to these. For example, not only there seems to be no reactive object to be used as an expression evaluation *root*, but actually there is no reactive *invocation* class --let's say a hypothetic `WebFilterInvocation`-- that would encapsulate the equivalent `ServerWebExchange` in order to support matching operations…

Am I looking in the wrong place (the `spring-security-web` module)? Was this left out on purpose? If not, maybe with some detail on what would be needed I might be able to help, at least with the not-so-low-level infrastructure...

### Version

Spring Security `5.1.0.BUILD-SNAPSHOT`

rwinch:
@danielfernandez Thanks for reaching out. There is not SpEL expression support for request mappings within WebFlux applications. This largely has to do with the fact that we can now use method references and lambdas. For example:

```java
@Bean
SecurityWebFilterChain springSecurityFilterChain(ServerHttpSecurity http) {
	http
		// ...
		.authorizeExchange()
			.pathMatchers("/users").access(this::isRob)
			.anyExchange().authenticated()
			.and()
		.httpBasic().and()
		.formLogin()
			.loginPage("/login");
	return http.build();
}

private Mono<AuthorizationDecision> isRob(Mono<Authentication> authentication,
		AuthorizationContext authorizationContext) {
	return authentication
			.map(Authentication::getName)
			.map(username -> username.startsWith("rob@"))
			.map(AuthorizationDecision::new);
}
```

[...]

PR #8584, opened 2020-05-22
Support Mono<Boolean> for Method Security

sheiy:
#4841

### DIFF
--- a/core/src/main/java/org/springframework/security/access/expression/ExpressionUtils.java
+++ b/core/src/main/java/org/springframework/security/access/expression/ExpressionUtils.java
@@ -18,12 +18,23 @@ package org.springframework.security.access.expression;
 import org.springframework.expression.EvaluationContext;
 import org.springframework.expression.EvaluationException;
 import org.springframework.expression.Expression;
+import reactor.core.publisher.Mono;
 
 public final class ExpressionUtils {
 
 	public static boolean evaluateAsBoolean(Expression expr, EvaluationContext ctx) {
 		try {
 			return expr.getValue(ctx, Boolean.class);
+		}
+		catch (EvaluationException e) {
+			throw new IllegalArgumentException("Failed to evaluate expression '"
+					+ expr.getExpressionString() + "'", e);
+		}
+	}
+
+	public static Mono<Boolean> evaluateAsMonoBoolean(Expression expr, EvaluationContext ctx) {
+		try {
+			return (Mono<Boolean>) expr.getValue(ctx, Mono.class);
 		}
 		catch (EvaluationException e) {
 			throw new IllegalArgumentException("Failed to evaluate expression '"
--- a/core/src/main/java/org/springframework/security/access/prepost/PreInvocationAuthorizationReactiveAdvice.java
+++ b/core/src/main/java/org/springframework/security/access/prepost/PreInvocationAuthorizationReactiveAdvice.java
@@ -39,6 +39,6 @@ public interface PreInvocationAuthorizationReactiveAdvice extends AopInfrastruct
 	 *                               annotations.
 	 * @return Mono<true> if authorised, Mono<false> otherwise
 	 */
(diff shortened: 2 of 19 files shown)


pivotal-issuemaster:
@sheiy Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=8584)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=8584) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@sheiy Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=8584)!

core/src/main/java/org/springframework/security/access/expression/ExpressionUtils.java
```diff
@@ -30,4 +31,14 @@ public static boolean evaluateAsBoolean(Expression expr, EvaluationContext ctx)
 					+ expr.getExpressionString() + "'", e);
 		}
 	}
+
+	public static Mono<Boolean> evaluateAsMonoBoolean(Expression expr, EvaluationContext ctx) {
```
rwinch, line 35: Adding a Reactive return type to a class used by non-reactive applications can potentially break non-reactive applications since depending on the JDK it may load all method definitions as it loads the class. In general, optional dependencies must be placed in distinct classes.

core/src/main/java/org/springframework/security/access/prepost/PreInvocationAuthorizationReactiveAdvice.java
```diff
@@ -0,0 +1,44 @@
...
+	 * to decide whether the method call is authorised.
+	 *
+	 * @param authentication         the information on the principal on whose account the
+	 *                               decision should be made
+	 * @param mi                     the method invocation being attempted
+	 * @param preInvocationAttribute the attribute built from the @PreFilter and @PostFilter
+	 *                               annotations.
+	 * @return Mono<true> if authorised, Mono<false> otherwise
+	 */
+	Mono<Boolean> before(Authentication authentication, MethodInvocation mi,
```
rwinch, line 42: Generally on the reactive side we use `Mono<Authentication>` so that if the `Authentication` is not needed, it is not subscribed to (i.e. if a method had `permitAll` we wouldn't need to load `Authentication` from session).

core/src/main/java/org/springframework/security/access/prepost/PrePostAdviceReactiveMethodInterceptor.java
```diff
@@ -58,7 +58,7 @@
 	 * @param preInvocationAdvice the {@link PreInvocationAuthorizationAdvice} to use
 	 * @param postInvocationAdvice the {@link PostInvocationAuthorizationAdvice} to use
 	 */
-	public PrePostAdviceReactiveMethodInterceptor(MethodSecurityMetadataSource attributeSource, PreInvocationAuthorizationAdvice preInvocationAdvice, PostInvocationAuthorizationAdvice postInvocationAdvice) {
+	public PrePostAdviceReactiveMethodInterceptor(MethodSecurityMetadataSource attributeSource, PreInvocationAuthorizationReactiveAdvice preInvocationAdvice, PostInvocationAuthorizationAdvice postInvocationAdvice) {
```
rwinch, line 61: This is a breaking change that we cannot make. We would need to add another constructor

rwinch:
Another note is that we may want to ensure what we do for evaluating SpEL expressions works for https://github.com/spring-projects/spring-security/issues/5867

sheiy:
@rwinch Thanks for your reply. I'm sorry for that I'm not familiar with spring-security.
I will close this PR and wait for your commit of ReactiveAuthorizationManager<Expression>.

Issue #8536, opened 2020-05-15
authorization_code grant should use same ServerRequestCache


jgrandja:
`ServerHttpSecurity.OAuth2ClientSpec.configure()` registers `OAuth2AuthorizationRequestRedirectWebFilter` and `OAuth2AuthorizationCodeGrantWebFilter` for the `authorization_code` grant flow. However, these `WebFilter` 's do not share a `ServerRequestCache`.

We should apply the necessary changes so that the configured `ServerRequestCache` is used by both `WebFilter` 's.

Related #8120

parikshitdutta:
Hi @jgrandja, can I look into it as well?

jgrandja:
Thanks @parikshitdutta. The issue is yours.

parikshitdutta:
Hi @jgrandja, regret taking more time, super cyclone massacred our place. Anyways, please have a look at PR #8587. I will look forward to your feedback.

Thank you.

jgrandja:
I'm really sorry to hear this @parikshitdutta. I hope you and your family are ok? There is no rush for this feature as it's scheduled for 5.4, which is being released early Sep. I will review later this week. Take care  of yourself and family.

parikshitdutta:
Hi @jgrandja, We are blessed to be safe and secure, but it has devastated our place. Things are getting back to normal. Thanks for asking and caring for us.

I will look forward to your feedback as you finish reviewing changes. On a separate note, is there a way, I can get used to the release calendar and planned features for forthcoming releases?

jgrandja:
Hey @parikshitdutta - you can see the scheduled releases [here](https://github.com/spring-projects/spring-security/milestones).

I'm glad to hear things are getting back to normal. I wish you and your family luck.

PR #8587, opened 2020-05-23
Add ServerRequestCache setter in OAuth2AuthorizationCodeGrantWebFilter

parikshitdutta:
- added setter to make ServerRequestCache injectable
- added respective test to validate serverRequestCache can not be set to null
- added test that asserts the supplied ServerRequestCache is actually used
- updated ServerHttpSecurity.OAuth2ClientSpec to set the configured requestCache
- updated tests in OAuth2ClientSpecTests to verify provided ServerRequestCache in OAuth2ClientSpec is used as custom configuration
- updated tests in OAuth2ClientSpecTests to verify provided ServerRequestCache in OAuth2ClientSpec is used as custom lambda configuration

closes gh-8536 [#8536]


jgrandja:
Thanks for the PR @parikshitdutta. This is now in master.
Hope you and your family are doing better.

Issue #5607, opened 2018-07-30
Register RestOperations @Bean to be used as default for oauth2-client flows


jgrandja:
We should register a `RestOperations` `@Bean` in `OAuth2ClientConfiguration` that is configured with default settings and is used as the default via `setRestOperations(restOperations)` in the various `oauth2-client` flows - see #5601.

The **well-known** `@Bean` should be named **springSecurityOAuth2RestOperations** and automatically be injected/used as the default in all `oauth2-client` flows. The user has the ability to register a `@Bean` with the same name, which will effectively override the default.

For now, the default `@Bean` will be configured with default connect and read timeout settings for 30 secs. However, the user may override this `@Bean` and provide custom settings for Proxy, SSL, HTTP Client backing implementation, etc.

rwinch:
I'm not sure we should be quite as granular as springSecurityOAuth2RestOperations. Otherwise we run into a pretty slippery slope on bean names. Do we add springSecurityOAuth2ClientRestOperations,  springSecurityOAuth2ResourceServerRestOperations, etc? Likely springSecurityRestOperations is good enough or potentially even just a Bean rather than the name.

jgrandja:
> Likely springSecurityRestOperations is good enough or potentially even just a Bean rather than the name.

I'm fine with `springSecurityRestOperations`. But I'm not sure about...

>  even just a Bean rather than the name

How do we know `RestOperations` bean is meant for oauth2-client flows? The user may have configured a bean for another use case within the app.

jgrandja:
Instead of registering a _well-known_ `@Bean` **name**, we should discover the `@Bean` in `OAuth2ClientConfiguration` **by type** only.

For the `@Bean` type, we should consider borrowing the design from `JwtDecoderFactory` for a possible solution.

For example, `OidcIdTokenDecoderFactory` implements `JwtDecoderFactory`, which is supplied a `ClientRegistration` to determine the type of `JwtDecoder` to create for the specific client.

```java
public interface JwtDecoderFactory<C> {
	// context - provides contextual information used to create a specific JwtDecoder
	JwtDecoder createDecoder(C context);
}

public final class OidcIdTokenDecoderFactory implements JwtDecoderFactory<ClientRegistration> {
	@Override
	public JwtDecoder createDecoder(ClientRegistration clientRegistration) {
                ....
	}
```

We could have a similar design for creating `RestOperations`, as follows:

```java
public interface RestOperationsFactory<C> {
[...]

PR #8588, opened 2020-05-24
Allow for custom rest template

Budlee:

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoderProviderConfiguration.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoderProviderConfiguration.java
@@ -36,23 +36,32 @@ import java.util.Map;
  * @author Rafiullah Hamedy
  * @since 5.2
  */
-class JwtDecoderProviderConfigurationUtils {
+public class JwtDecoderProviderConfiguration {
 	private static final String OIDC_METADATA_PATH = "/.well-known/openid-configuration";
 	private static final String OAUTH_METADATA_PATH = "/.well-known/oauth-authorization-server";
-	private static final RestTemplate rest = new RestTemplate();
-	private static final ParameterizedTypeReference<Map<String, Object>> typeReference =
+	private final RestTemplate rest = new RestTemplate();
+	private final ParameterizedTypeReference<Map<String, Object>> typeReference =
 			new ParameterizedTypeReference<Map<String, Object>>() {};
 
-	static Map<String, Object> getConfigurationForOidcIssuerLocation(String oidcIssuerLocation) {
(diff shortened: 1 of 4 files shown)


Budlee:
Reference to issue https://github.com/spring-projects/spring-security/issues/5607

jzheaux:
@Budlee, thanks for taking some time to think this through. Making `RestTemplate` more configurable here would certainly be nice.

This is still an open discussion, but the options I can see are:

```java
ReactiveJwtDecoders#fromIssuerLocation(String issuer, RestTemplate rest)
```

or providing a new class that implements `ReactiveJwtDecoderFactory`:

```java
public class IssuerReactiveJwtDecoderFactory implements ... {
    private final RestTemplate rest;

    @Override
    public ReactiveJwtDecoder createDecoder(String issuer) {
        // ...
    }

    public void setRestTemplate(RestTemplate rest) {
        // ...
    }
}
```

with equivalents needed for imperative.

Either way, I think that the Utils class may still move around a bit, so I'd prefer to keep that package-private as it is. AFAICT, it's not necessary to expose it.

As far as multi-tenancy, I think that your proposal doesn't yet address the use case you raised for having a different RestTemplate per issuer. To have something that sophisticated, I think you'd still need to supply your own `AuthenticationManagerResolver<String>` in the constructor. 

Even still, it may be valuable to expose something like a `ReactiveJwtDecoderFactory<String>` in `JwtIssuerReactiveAuthenticationManagerResolver`, but that will probably depend on how the larger conversation about `RestTemplate` goes.

rwinch:
Another option might be to allow the user to use `RestTemplate` directly.

Budlee:
@jgrandja  is there any decision on what is the way forward to expose this to the RestTemplate to the user? Is the idea to commit to #8624 as the solution

jgrandja:
@Budlee I'm going to spend some time this week to figure out a solution. This one is a bit tricky as there are quite a few touch points in the code and a couple of special use cases.

jgrandja:
@budlee The original design intent for `JwtDecoders` and `ReactiveJwtDecoders` is that it was meant to be used as a utility class, hence the static methods. It was provided as a convenience for most use cases but not all. For more advanced use cases, eg. where you need to configure proxy settings for the underlying HTTP client, you can configure a `JwtDecoder` (or `ReactiveJwtDecoder`) and expose it as a `@Bean` - see [example 1](https://docs.spring.io/spring-security/site/docs/current/reference/html5/#oauth2resourceserver-jwt-decoder-dsl) and [example 2](https://docs.spring.io/spring-security/site/docs/current/reference/html5/#oauth2resourceserver-jwt-timeouts).

After reviewing your PR, we do not want to expose `JwtDecoderProviderConfiguration` or provide a `ReactiveJwtDecoders.fromIssuerLocation(String issuer, RestTemplate rest)`. This was never the design intent for `JwtDecoders` or `ReactiveJwtDecoders`.

I'm going to close this PR based on the above reasoning.

Issue #8609, opened 2020-05-29
OAuth2AuthorizationCodeGrantWebFilter should handle OAuth2AuthorizationException


wjlc:
@jgrandja 
> The `code` (`authorization_code`) is a temporary credential that can be used one-time only, hence, the `AUTHORIZATION_REQUEST_NOT_FOUND` error.

The problem is not in `authorization_code`. `AuthenticationWebFilter` has `authenticationFailureHandler` to handle invalid authentication(`authorization_code`). `OidcAuthorizationCodeReactiveAuthenticationManager` throws `OAuth2AuthenticationException` which is `AuthenticationException` and browser is redirected to `/login?error`

The problem is in 
1. `ServerOAuth2AuthorizationCodeAuthenticationTokenConverter` it throws `OAuth2AuthorizationException` and 
2. `AuthenticationWebFilter` doesn't handle any errors from `ServerAuthenticationConverter`

Only one ServerAuthenticationConverter implementation will redirect browser to login page
`ServerHttpBasicAuthenticationConverter` cause it return `Mono.empty()` in case of any authentication problem

_Originally posted by @iilkevych in https://github.com/spring-projects/spring-security/issues/7884#issuecomment-634278181_

wjlc:
https://github.com/spring-projects/spring-security/issues/7884#issue-558408713

iilkevych:
Some `ReactiveAuthenticationManager` implementations also can throw `OAuth2AuthorizationException` It seems to be right to check all them in `AuthenticationWebFilter` using one base exception `AccessDeniedException`
Which is actually handled already

jgrandja:
@wjlc @iilkevych Thank you for this report.

The issue is in `OAuth2AuthorizationCodeGrantWebFilter` as it does not handle `OAuth2AuthorizationException`, which can be thrown by `ServerOAuth2AuthorizationCodeAuthenticationTokenConverter`.

The fix should be in `OAuth2AuthorizationCodeGrantWebFilter` to handle `OAuth2AuthorizationException` and delegate to `authenticationFailureHandler` on errors.

Would either of you be interested in submitting a PR for this?

wjlc:
@shazin This is not the correct place for the fix. plz look at issue and replicate it. [https://github.com/spring-projects/spring-security/issues/7884#issue-558408713](url)

wjlc:
[...]

PR #8616, opened 2020-05-30
OAuth2AuthorizationCodeGrantWebFilter should handle OAuth2AuthorizationException #8609

shazin:
The issue is in OAuth2AuthorizationCodeGrantWebFilter as it does not handle OAuth2AuthorizationException, which can be thrown by ServerOAuth2AuthorizationCodeAuthenticationTokenConverter.

The fix should be in OAuth2AuthorizationCodeGrantWebFilter to handle OAuth2AuthorizationException and delegate to authenticationFailureHandler on errors.

Fixes gh-8609

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/OAuth2AuthorizationCodeGrantWebFilter.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/OAuth2AuthorizationCodeGrantWebFilter.java
@@ -23,6 +23,8 @@ import org.springframework.security.core.AuthenticationException;
 import org.springframework.security.core.authority.AuthorityUtils;
 import org.springframework.security.core.context.ReactiveSecurityContextHolder;
 import org.springframework.security.core.context.SecurityContext;
+import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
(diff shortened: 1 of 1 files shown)


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/OAuth2AuthorizationCodeGrantWebFilter.java
```diff
@@ -184,6 +186,8 @@ private void updateDefaultAuthenticationConverter() {
 		return this.authenticationManager.authenticate(token)
 				.switchIfEmpty(Mono.defer(() -> Mono.error(new IllegalStateException("No provider found for " + token.getClass()))))
 				.flatMap(authentication -> onAuthenticationSuccess(authentication, webFilterExchange))
+				.onErrorResume(OAuth2AuthorizationException.class, e -> this.authenticationFailureHandler
```
jgrandja, line 189: This is not the correct place for the fix. `OAuth2AuthorizationException` is not thrown by `AuthenticationManager`, instead it's thrown by `authenticationConverter.convert(exchange)` (`ServerOAuth2AuthorizationCodeAuthenticationTokenConverter`)

jgrandja:
@shazin I'm closing this PR as I just pushed the fix in these commits 4c902bb da4b626

Issue #6374, opened 2019-01-08
Provide Cookie implementation of AuthorizationRequestRepository


jgrandja:
We should consider providing a `Cookie` based implementation of `AuthorizationRequestRepository`.

afrancoc2000:
Hi @jgrandja, How many votes do you need?

jgrandja:
Hi @afrancoc2000. There is no specific number of votes we target but this one has enough to start prioritizing. However, we are pretty jammed with higher priority tasks that need to get into the upcoming 5.2 release so this one will take lower priority. It would be helpful if someone from the community would contribute a PR for this as I believe it would be fairly trivial. Would you be interested in submitting a PR?

jgrandja:
@afrancoc2000 Can you provide details on why you require a cookie implementation of `AuthorizationRequestRepository`. I'd like to understand your use case before we proceed with any work. Thanks.

afrancoc2000:
Sure the reason is microservices applications, today the best way of handling this type of apps is making them stateless so any app can respond to any request. 

The authorization can be easily saved and restored from a jwt cookie thats better than having a session because for sessions I need an extra component like a redis cache, that means more infrastructure, more points of failure and the posiblility of conflicts between sessions, problems that could be easily solved by replacing a couple lines of code.

Also that allows me to link an external frontend more easily just by passing the jwt cookie that is a standard in the industry.

I would love to help I was looking at the example of @naturalprogrammer in this links: [link 1](https://dzone.com/articles/spring-security-5-oauth-20-login-and-signup-in-sta), [link 2](https://github.com/naturalprogrammer/spring-lemon) but found out that storing the OAuth2AuthorizationRequest  is not enough and I would prefer to do the implementation with a jwt cookie saving and restoring the security context, haven't got till there yet.

jgrandja:
@afrancoc2000 The `AuthorizationRequestRepository` is responsible for persisting the `OAuth2AuthorizationRequest` as part of the `authorization_code` grant flow. Please see [Authorization Request](https://tools.ietf.org/html/rfc6749#section-4.1.1) for more details.

It does not save a `Jwt` as it seems that is your understanding?

> The authorization can be easily saved and restored from a jwt cookie

Please correct me if I'm misunderstanding your comments?

fnaeem78:
[...]

PR #8621, opened 2020-06-02
Add HttpCookieOAuth2AuthorizationRequestRepository

Gittenburg:
Fixes gh-6374


jgrandja:
@Gittenburg Thanks for the PR, however, we're not looking to add a `Cookie` implementation for the reasons mentioned in this [comment](https://github.com/spring-projects/spring-security/issues/6374#issuecomment-743241379).

PR #8651, opened 2020-06-04
Don't consume request body. Fixes gh-8650

### LINKED ISSUE #8650: DefaultOAuth2AuthorizationRequestResolver erroneously consumes POST request body
**Describe the bug**
DefaultOAuth2AuthorizationRequestResolver consumes the request body, by calling `getParameter` on the request, even if the request doesn't match the authorization request base uri.

The bug is in `resolve(HttpServletRequest request)` method, which calls `getAction`, even if the registrationId is null.

**To Reproduce**
Create a filterchain with DefaultOAuth2AuthorizationRequestResolver in it. Send a POST request with a body. Observe that in the servlet, the request body appears to be empty.

**Expected behavior**
I expect that the body doesn't appear to be empty. 

**Sample**
No sample, but I'm adding a PR with a test.

eamelink:
Per the servlet spec, getParameter(name) consumes the request body for POST requests. This commit prevents DefaultOAuth2AuthorizationRequestResolver from consuming the request body for non-Authorization requests.

Fixes #8650


jzheaux:
Thanks, @eamelink! This is now merged into `master` via cd3fd6762f621f8d36ff5792cc427d1609dd3540

PR #8653, opened 2020-06-04
Adds cookie based RequestCache

zeeshanadnan:
fixes spring-projectsgh-8034

### DIFF
--- a/web/src/main/java/org/springframework/security/web/savedrequest/CookieRequestCache.java
+++ b/web/src/main/java/org/springframework/security/web/savedrequest/CookieRequestCache.java
@@ -1,0 +1,131 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.web.savedrequest;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.springframework.security.web.util.UrlUtils;
+import org.springframework.security.web.util.matcher.AnyRequestMatcher;
+import org.springframework.security.web.util.matcher.RequestMatcher;
+import org.springframework.util.Assert;
+import org.springframework.web.util.UriComponents;
+import org.springframework.web.util.UriComponentsBuilder;
+import org.springframework.web.util.WebUtils;
+
+import javax.servlet.http.Cookie;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.util.Base64;
+
+
+/**
+ * An Implementation of {@code RequestCache} which saves the original request URI in a cookie.
+ *
+ * @author Zeeshan Adnan
+ */
+public class CookieRequestCache implements RequestCache {
+
+	private RequestMatcher requestMatcher = AnyRequestMatcher.INSTANCE;
+	protected final Log logger = LogFactory.getLog(this.getClass());
+
+	private static final String COOKIE_NAME = "REDIRECT_URI";
+	private static final int COOKIE_MAX_AGE = -1;
+
+	@Override
+	public void saveRequest(HttpServletRequest request, HttpServletResponse response) {
+		if (requestMatcher.matches(request)) {
+			String redirectUrl = UrlUtils.buildFullRequestUrl(request);
+			Cookie savedCookie = new Cookie(COOKIE_NAME, encodeCookie(redirectUrl));
+			savedCookie.setMaxAge(COOKIE_MAX_AGE);
+			savedCookie.setSecure(request.isSecure());
+			savedCookie.setPath(request.getContextPath());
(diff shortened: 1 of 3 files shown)


web/src/main/java/org/springframework/security/web/savedrequest/CookieRequestCache.java
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	/**
+	 * If the {@code cookieName} property is set, the request is stored in
+	 * the cookie using this attribute name. Default is
+	 * "saved_request".
+	 *
+	 * @param cookieName is the new name of the cookie.
+	 */
+	public void setCookieName(String cookieName) {
```
eleftherias, line 106: I don't think we need to make the cookie name configurable.
We can always add this functionality later if a user requests it.
Similarly, I don't think we need `setCookieMaxAge` and `setCookiePath` for now.

web/src/main/java/org/springframework/security/web/savedrequest/CookieRequestCache.java
```diff
@@ -0,0 +1,130 @@
...
+import org.springframework.util.Assert;
+import org.springframework.web.util.WebUtils;
+
+import javax.servlet.http.Cookie;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+
+/**
+ * {@code RequestCache} which stores the {@code SavedRequest} in a cookie.
```
eleftherias, line 32: We should let users know that this cache will not store the complete request, but just the request URI.
Unlike `HttpSessionRequestCache`, it does not store headers etc.

web/src/main/java/org/springframework/security/web/savedrequest/CookieRequestCache.java
```diff
@@ -0,0 +1,130 @@
...
+			response.addCookie(savedCookie);
+		} else {
+			logger.debug("Request not saved as configured RequestMatcher did not match");
+		}
+	}
+
+	@Override
+	public SavedRequest getRequest(HttpServletRequest request, HttpServletResponse response) {
+		Cookie savedRequestCookie = WebUtils.getCookie(request, cookieName);
+		return savedRequestCookie != null ? new SimpleSavedRequest(savedRequestCookie.getValue()) : null;
```
eleftherias, line 63: This should use `DefaultSavedRequest` instead of `SimpleSavedRequest`

web/src/main/java/org/springframework/security/web/savedrequest/CookieRequestCache.java
```diff
@@ -0,0 +1,130 @@
...
+	protected final Log logger = LogFactory.getLog(this.getClass());
+
+	private String cookieName = "saved_request";
+	private String cookiePath = null;
+	private int cookieMaxAge = -1;
+
+	@Override
+	public void saveRequest(HttpServletRequest request, HttpServletResponse response) {
+		if (requestMatcher.matches(request)) {
+			String redirectUrl = UrlUtils.buildFullRequestUrl(request);
```
eleftherias, line 48: The URL should be encoded before being saved.
A URL may contain a comma, which is an invalid character in a cookie value.
See `CookieServerRequestCache` for an example
https://github.com/spring-projects/spring-security/blob/28d2cfa14aeea54c20cd87f294fc7546063b0713/web/src/main/java/org/springframework/security/web/server/savedrequest/CookieServerRequestCache.java#L127

web/src/main/java/org/springframework/security/web/savedrequest/CookieRequestCache.java
zeeshanadnan: @eleftherias  i did not find a suitable constructor to do it in an easy way. Also in the `getMatchingRequest` method its always going to return null. Is it going to make a difference or i am missing something? Any suggestion on a better approach will be very helpful. Thanks.

eleftherias: @zeeshanadnan It can use the builder, 
```
new DefaultSavedRequest.Builder().setRequestURI(savedRequestCookie.getValue()).build()
```

web/src/test/java/org/springframework/security/web/savedrequest/CookieRequestCacheTests.java
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * @author Zeeshan Adnan
+ */
+public class CookieRequestCacheTests {
+
+	private static final String DEFAULT_COOKIE_NAME = "saved_request";
+
+	@Test
+	public void originalRequestSetsSavedRequestInACookieOnResponse() {
```
eleftherias, line 38: In an effort to have test names be consistent across the codebase, we would like all new tests to follow the naming convention `methodWhenConditionThenResult`.
For example, this test could be renamed to something like `saveRequestWhenMatchesThenSavedRequestInACookieOnResponse`.

web/src/main/java/org/springframework/security/web/savedrequest/CookieRequestCache.java
```diff
@@ -0,0 +1,131 @@
...
+ *
+ * @author Zeeshan Adnan
+ */
+public class CookieRequestCache implements RequestCache {
+
+	private RequestMatcher requestMatcher = AnyRequestMatcher.INSTANCE;
+	protected final Log logger = LogFactory.getLog(this.getClass());
+
+	private static final String COOKIE_NAME = "saved_request";
+	private static final String COOKIE_PATH = null;
```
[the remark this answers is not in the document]
zeeshanadnan: Initially my thought was user should be able to control the cookie path so i included a property and setter method. But yes you are right the default value should be request's context path.

zeeshanadnan:
Sorry @eleftherias for the late updates.

eleftherias:
Thanks for the PR @zeeshanadnan!

Issue #8589, opened 2020-05-24
Add logging


Budlee:
**Expected Behavior**

The Jwt Validators that are provided on the oauth jose package have no logging in them when they fail. It makes it hard when debugging to understand when a Jwt Token has failed and for what reason.
I'm not sure if this is on purpose for security reasons to not log the failure

**Current Behavior**

No loggin in the default JWT Validators

**Context**

This has slowed down debugging and understanding which validator has failed.
Debug messages when the logging is set to debug would be helpful.

To get an insight into what JWTs are failing have patched the validators and manually setup the JwtDecoder/ReactiveJwtDecoder.

They can not be extended as they are all Final so patching is the only option.

jzheaux:
Thanks for raising this, @Budlee. We certainly want to ensure that there's enough logging available.

Before knowing what the right change is, let's get on the same page about what is missing. In Spring Security, each authentication filter logs failures at the DEBUG level, so you can do:

```yaml
logging:
  level:
    org.springframework.security: DEBUG
```
and you'll see failure information, for example when a bearer token fails validation.

On the Servlet side, you'll see an exception like:

```
...
Caused by: com.nimbusds.jose.proc.BadJWSException: Signed JWT rejected: Invalid signature
	at com.nimbusds.jwt.proc.DefaultJWTProcessor.process(DefaultJWTProcessor.java:405) ~[nimbus-jose-jwt-8.17.1.jar:8.17.1]
	at com.nimbusds.jwt.proc.DefaultJWTProcessor.process(DefaultJWTProcessor.java:330) ~[nimbus-jose-jwt-8.17.1.jar:8.17.1]
	at org.springframework.security.oauth2.jwt.NimbusReactiveJwtDecoder.createClaimsSet(NimbusReactiveJwtDecoder.java:531) ~[spring-security-oauth2-jose-5.4.0-SNAPSHOT.jar:5.4.0-SNAPSHOT]
	... 63 common frames omitted
```

On the WebFlux side, you'll see a simple message:

```
Authentication failed: Failed to validate the token
```

I think things can be improved on the WebFlux side, and perhaps they can be improved on the Servlet side as well. 

Can you elaborate on what information that you aren't able to get that you need?

Budlee:
@jzheaux this is why i did not see it as I am using reactive for pretty much everything currently.

Would you accept a PR that adds more logging to the reactive implementation and  are there any constaints?

jzheaux:
Got it, makes sense.

[...]

PR #8665, opened 2020-06-08
gh-8589 Additional Jwt validation debug messages

Budlee:
Adding extra logging for JWT Validation 
https://github.com/spring-projects/spring-security/issues/8589

### DIFF
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/DelegatingOAuth2TokenValidator.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/DelegatingOAuth2TokenValidator.java
@@ -16,17 +16,19 @@
 
 package org.springframework.security.oauth2.core;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.springframework.util.Assert;
+
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-
-import org.springframework.util.Assert;
+import java.util.stream.Collectors;
 
 /**
  * A composite validator
  *
  * @param <T> the type of {@link AbstractOAuth2Token} this validator validates
- *
  * @author Josh Cummings
  * @since 5.1
  */
@@ -63,10 +65,9 @@ public final class DelegatingOAuth2TokenValidator<T extends AbstractOAuth2Token>
 	public OAuth2TokenValidatorResult validate(T token) {
 		Collection<OAuth2Error> errors = new ArrayList<>();
 
-		for ( OAuth2TokenValidator<T> validator : this.tokenValidators) {
+		for (OAuth2TokenValidator<T> validator : this.tokenValidators) {
 			errors.addAll(validator.validate(token).getErrors());
 		}
-
 		return OAuth2TokenValidatorResult.failure(errors);
 	}
 }
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/DelegatingOAuth2TokenValidator.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/DelegatingOAuth2TokenValidator.java
@@ -16,16 +16,17 @@
 
 package org.springframework.security.oauth2.core;
 
-import org.springframework.util.Assert;
-
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+
+import org.springframework.util.Assert;
 
 /**
  * A composite validator
  *
  * @param <T> the type of {@link AbstractOAuth2Token} this validator validates
+ *
  * @author Josh Cummings
  * @since 5.1
  */
@@ -62,9 +63,10 @@ public final class DelegatingOAuth2TokenValidator<T extends AbstractOAuth2Token>
 	public OAuth2TokenValidatorResult validate(T token) {
 		Collection<OAuth2Error> errors = new ArrayList<>();
 
-		for (OAuth2TokenValidator<T> validator : this.tokenValidators) {
+		for ( OAuth2TokenValidator<T> validator : this.tokenValidators) {
 			errors.addAll(validator.validate(token).getErrors());
 		}
+
 		return OAuth2TokenValidatorResult.failure(errors);
 	}
 }
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/DelegatingOAuth2TokenValidator.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/DelegatingOAuth2TokenValidator.java
@@ -16,14 +16,11 @@
 
 package org.springframework.security.oauth2.core;
 
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
 import org.springframework.util.Assert;
 
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.stream.Collectors;
 
 /**
  * A composite validator
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtClaimValidator.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtClaimValidator.java
@@ -66,7 +66,7 @@ public final class JwtClaimValidator<T> implements OAuth2TokenValidator<Jwt> {
 			return OAuth2TokenValidatorResult.success();
 		} else {
 			if (logger.isDebugEnabled()) {
-				logger.debug("Failed to validate OAuth2 token claim " + claim);
+				logger.debug(error.getDescription());
 			}
 			return OAuth2TokenValidatorResult.failure(error);
 		}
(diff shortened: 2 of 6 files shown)


Budlee:
Hey @jzheaux, I think that  would be good to have as well.
The stack trace for reactive would be horriblly long is my only thought. But as a trace maybe a good.

The other thing is in this specific case is that I have logging for each JWT Validator that fails. This is quite nice in the log.

Maybe it should change but the current validate gets the first validation fail even if there is many
org.springframework.security.oauth2.jwt.NimbusReactiveJwtDecoder#validateJwt
```java
if ( result.hasErrors() ) {
			String message = result.getErrors().iterator().next().getDescription();
			throw new JwtValidationException(message, result.getErrors());
		}
```
This seems silly as the DelegatedOauth collects all
org.springframework.security.oauth2.core.DelegatingOAuth2TokenValidator#validate
```java
for (OAuth2TokenValidator<T> validator : this.tokenValidators) {
			errors.addAll(validator.validate(token).getErrors());
		}
```


Also I think this `result.getErrors().iterator().next().getDescription();` is an NPE waiting to happen

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtTimestampValidator.java
```diff
@@ -71,6 +77,9 @@ public OAuth2TokenValidatorResult validate(Jwt jwt) {
 
 		if (expiry != null) {
 			if (Instant.now(this.clock).minus(clockSkew).isAfter(expiry)) {
+				if (logger.isDebugEnabled()){
```
jzheaux, line 80: I see your point that since there is the potential to collect multiple errors, but only one is ultimately in the exception message, that not logging in these places causes an information gap.

What would you think about having the error message and the debug message align? For example, doing something like:

```java
String reason = String.format("Jwt expired at %s", isoDateTime(jwt.getExpiresAt()));
if (logger.isDebugEnabled()) { 
    logger.debug("Failed to validate because " + reason);
}
OAuth2Error error = new OAuth2Error(..., reason, ...);
```

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtReactiveAuthenticationManager.java
```diff
@@ -74,6 +78,9 @@ public void setJwtAuthenticationConverter(
 	}
 
 	private AuthenticationException onError(JwtException e) {
+		if (logger.isDebugEnabled()) {
```
jzheaux, line 81: I'm not seeing the added benefit of this one since `AuthenticationWebFilter` displays the same information. Is it necessary?

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/DelegatingOAuth2TokenValidator.java
```diff
@@ -16,17 +16,19 @@
 
 package org.springframework.security.oauth2.core;
 
+import org.apache.commons.logging.Log;
```
jzheaux, line 19: It seems there are some unused imports here.

jzheaux:
That's great feedback, @Budlee. I'll try and address a couple of your comments:

> This seems silly as the DelegatedOauth collects all org.springframework.security.oauth2.core.DelegatingOAuth2TokenValidator#validate

What would you supply as the message for the exception if not the first validation error message?

> Also I think this result.getErrors().iterator().next().getDescription(); is an NPE waiting to happen

I agree that this could be polished. 

The `OAuth2TokenValidatorResult` constructor disallows null or empty lists, but it doesn't check if `Collection<OAuth2Error>` contains null elements. It might be good add that assertion.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/DelegatingOAuth2TokenValidator.java
Budlee: Sorry. I'll remove

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtTimestampValidator.java
Budlee: Sounds good. I'll reimplement this

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtReactiveAuthenticationManager.java
Budlee: Agree. I've removed

Budlee:
@jzheaux,
I've resolved the issues that you had and i've add the tidy up for that potential NPE.

As the validators are now all logging it does make sense for just the first error to go back

Please let me know if you want any more changes

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusReactiveJwtDecoder.java
```diff
@@ -177,10 +181,13 @@ private Jwt createJwt(JWT parsedJwt, JWTClaimsSet jwtClaimsSet) {
 
 	private Jwt validateJwt(Jwt jwt) {
 		OAuth2TokenValidatorResult result = this.jwtValidator.validate(jwt);
-
 		if ( result.hasErrors() ) {
-			String message = result.getErrors().iterator().next().getDescription();
-			throw new JwtValidationException(message, result.getErrors());
+			final Collection<OAuth2Error> errors = result.getErrors();
+			final Optional<String> validationErrorString = errors.stream()
```
jzheaux, line 186: I agree that this more thorough check is an improvement. For now, using `Optional` and `Stream` are too taxing on the garbage collector to be used in Spring Security. 

Will you please change this to a for loop that finds the first non-null description? You could probably do that in a separate method.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtTimestampValidator.java
jzheaux: This looks good for `nbf`. Will you please make the same change for `exp`?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtReactiveAuthenticationManager.java
```diff
@@ -39,6 +42,7 @@
  * @since 5.1
  */
 public final class JwtReactiveAuthenticationManager implements ReactiveAuthenticationManager {
+	private static final Log logger = LogFactory.getLog(JwtReactiveAuthenticationManager.class);
```
jzheaux, line 45: Looks like this might be unnecessary for now - in fact, I think this class can be removed from the PR

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtClaimValidator.java
```diff
@@ -61,6 +65,9 @@ public OAuth2TokenValidatorResult validate(Jwt token) {
 		if (test.test(claimValue)) {
 			return OAuth2TokenValidatorResult.success();
 		} else {
+			if (logger.isDebugEnabled()) {
+				logger.debug("Failed to validate OAuth2 token claim " + claim);
```
jzheaux, line 69: It'd be nice to have the `error` description align with the debug message here, too, like you did with `nbf`.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/DelegatingOAuth2TokenValidator.java
```diff
@@ -16,17 +16,16 @@
 
 package org.springframework.security.oauth2.core;
 
+import org.springframework.util.Assert;
```
jzheaux, line 19: Since these are just whitespace changes, let's leave this out of the PR.

Budlee: Fixed

Budlee:
@jzheaux i've updates

jzheaux:
Thanks for the updates, @Budlee! 

It looks like the build is failing. Will you try rebasing, and double-check by running `./gradlew check`? If there are any checkstyle errors, please fix them.

Then, at that point in preparation for merging, would you please squash your commits and [format the commit message](https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.md#format-commit-messages).

jzheaux:
Thanks, @Budlee! This is now merged into `master` via b2728059ae903d027d0da0f4a25bed6df213a0ce and 90e5f45e1f3a2c0b1e2d63faaabf3df6b86ee34f.

Note that I separated your two contributions into two separate commits for easier backporting. First, the additional logging and second the NPE polish.

Issue #8661, opened 2020-06-08
Ability to easily read attribute values from SAML response


kostic017:
**Expected Behavior**
I should be able to read attribute values after SAML response is parsed and validated without parsing it myself.

**Current Behavior**
As far I can tell `OpenSamlAuthenticationProvider` parses and validates the response, and then throws away the `Assertion` object. What I'm left with is encrypted response string.

**Context**
I want to show the user's first and last name on my website.

I could parse XML myself after successful login, but that doesn't seem very pretty.

I think that in the earlier iterations `SecurityContextHolder.getContext().getAuthentication().getCredentials()` returned `SAMLCredential` object with `getAttribute` methods.

Sorry if this is already possible, please tell me how it's supposed to be done in that case. If it's not possible, what nice workaround I could use?

I know that it's possible to extract roles using custom authorities extractor, but what about other attributes?

jzheaux:
@kostic017 thanks for reaching out; I think adding attributes to the `Saml2AuthenticatedPrincipal` would be a way to address this.

To add it, it'll be important to maintain Spring Security's independence from OpenSAML, meaning that the resulting authentication shouldn't have any OpenSAML-specific material in it.

One way this could work would be to have `OpenSamlAuthenicationProvider` iterate through all the attributes and convert them to their corresponding Java types. I think it would be reasonable to add a `Map<String, Object> getAttributes()` default method to `Saml2AuthenticatedPrincipal` in that case.

Does that sound like what you are looking for and would you be interested in submitting a PR to add it?

kostic017:
Well, as you can probably tell, I've created a pull request... (force push to my branch, looks ugly here)

fpagliar:
Thanks so much for raising this issue!!! This is a big blocker for us.

kostic017:
@fpagliar Here you can find a nice workaround until fix gets released
https://stackoverflow.com/q/62210172/10479742
https://stackoverflow.com/q/58400571/10479742

fpagliar:
@kostic017 thank you! 
Yeah you can also use the AuthoritiesExtractor to intercept the selected valid decrypted assertion and store what you need. But it is horribly hacky.

PR #8667, opened 2020-06-09
Getting response attributes from Saml2AuthenticatedPrincipal

kostic017:
It's not possible to read attribute values from SAML response after successful login because `Saml2AuthenticatedPrincipal` doesn't have `getAttributes()` method like `OAuth2AuthenticatedPrincipal` does. I've added the `getAttributes()` method to the interface. The map is being populated in `OpenSamlAuthenticationProvider`.

With this patch, I won't have to parse SAML response the second time by myself if I want to read some attribute value.

fixes gh-8661

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProvider.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,11 +17,13 @@ package org.springframework.security.saml2.provider.service.authentication;
 
 import java.security.cert.X509Certificate;
 import java.time.Duration;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -31,7 +33,19 @@ import javax.annotation.Nonnull;
 import net.shibboleth.utilities.java.support.resolver.CriteriaSet;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.joda.time.DateTime;
 import org.opensaml.core.criterion.EntityIdCriterion;
+import org.opensaml.core.xml.XMLObject;
+import org.opensaml.core.xml.config.XMLObjectProviderRegistrySupport;
+import org.opensaml.core.xml.io.Marshaller;
+
+import org.opensaml.core.xml.schema.XSAny;
+import org.opensaml.core.xml.schema.XSBoolean;
+import org.opensaml.core.xml.schema.XSBooleanValue;
+import org.opensaml.core.xml.schema.XSDateTime;
+import org.opensaml.core.xml.schema.XSInteger;
+import org.opensaml.core.xml.schema.XSString;
+import org.opensaml.core.xml.schema.XSURI;
 import org.opensaml.saml.common.assertion.ValidationContext;
 import org.opensaml.saml.common.assertion.ValidationResult;
 import org.opensaml.saml.common.xml.SAMLConstants;
@@ -45,6 +59,8 @@ import org.opensaml.saml.saml2.assertion.SubjectConfirmationValidator;
 import org.opensaml.saml.saml2.assertion.impl.AudienceRestrictionConditionValidator;
 import org.opensaml.saml.saml2.assertion.impl.BearerSubjectConfirmationValidator;
 import org.opensaml.saml.saml2.core.Assertion;
+import org.opensaml.saml.saml2.core.Attribute;
+import org.opensaml.saml.saml2.core.AttributeStatement;
 import org.opensaml.saml.saml2.core.EncryptedAssertion;
 import org.opensaml.saml.saml2.core.EncryptedID;
 import org.opensaml.saml.saml2.core.NameID;
@@ -205,8 +221,9 @@ public final class OpenSamlAuthenticationProvider implements AuthenticationProvi
 			List<Assertion> validAssertions = validateResponse(token, response);
 			Assertion assertion = validAssertions.get(0);
 			String username = getUsername(token, assertion);
+			Map<String, List<Object>> attributes = getAssertionAttributes(assertion);
 			return new Saml2Authentication(
-					new SimpleSaml2AuthenticatedPrincipal(username), token.getSaml2Response(),
+					new SimpleSaml2AuthenticatedPrincipal(username, attributes), token.getSaml2Response(),
 					this.authoritiesMapper.mapAuthorities(getAssertionAuthorities(assertion)));
 		} catch (Saml2AuthenticationException e) {
 			throw e;
@@ -492,6 +509,60 @@ public final class OpenSamlAuthenticationProvider implements AuthenticationProvi
 			}
 		}
 		throw last;
+	}
+
+	private Map<String, List<Object>> getAssertionAttributes(Assertion assertion) {
+		Map<String, List<Object>> attributeMap = new LinkedHashMap<>();
+		for (AttributeStatement attributeStatement : assertion.getAttributeStatements()) {
+			for (Attribute attribute : attributeStatement.getAttributes()) {
+
+				List<Object> attributeValues = new ArrayList<>();
+				for (XMLObject xmlObject : attribute.getAttributeValues()) {
+					Object attributeValue = getXmlObjectValue(xmlObject);
+					if (attributeValue != null) {
+						attributeValues.add(attributeValue);
+					}
+				}
+				attributeMap.put(attribute.getName(), attributeValues);
+
+			}
+		}
+		return attributeMap;
+	}
+
+	private Object getXmlObjectValue(XMLObject xmlObject) {
+		if (xmlObject == null) {
+			return null;
+		}
+		if (xmlObject instanceof XSAny) {
+			return getXSAnyObjectValue((XSAny) xmlObject);
+		}
+		if (xmlObject instanceof XSString) {
+			return ((XSString) xmlObject).getValue();
+		}
+		if (xmlObject instanceof XSInteger) {
+			return ((XSInteger) xmlObject).getValue();
+		}
+		if (xmlObject instanceof XSURI) {
+			return ((XSURI) xmlObject).getValue();
+		}
+		if (xmlObject instanceof XSBoolean) {
+			XSBooleanValue xsBooleanValue = ((XSBoolean) xmlObject).getValue();
+			return xsBooleanValue != null ? xsBooleanValue.getValue() : null;
+		}
+		if (xmlObject instanceof XSDateTime) {
+			DateTime dateTime = ((XSDateTime) xmlObject).getValue();
+			return dateTime != null ? Instant.ofEpochMilli(dateTime.getMillis()) : null;
+		}
+		return null;
+	}
+
+	private Object getXSAnyObjectValue(XSAny xsAny) {
+		Marshaller marshaller = XMLObjectProviderRegistrySupport.getMarshallerFactory().getMarshaller(xsAny);
+		if (marshaller != null) {
+			return this.saml.serialize(xsAny);
+		}
+		return xsAny.getTextContent();
 	}
 
 	private Saml2Error validationError(String code, String description) {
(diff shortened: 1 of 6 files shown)


saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProvider.java
```diff
@@ -474,6 +480,33 @@ else if (subject.getEncryptedID() != null) {
 		return username;
 	}
 
+	private Map<String, Object> getAssertionAttributes(Assertion assertion) {
+		Map<String, Object> attributeMap = new HashMap<>();
```
jzheaux, line 484: Let's please use a `LinkedHashMap` so that it preserves the order in which OpenSAML gave them to us.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProvider.java
```diff
@@ -474,6 +480,33 @@ else if (subject.getEncryptedID() != null) {
...
+		for (AttributeStatement attributeStatement : assertion.getAttributeStatements()) {
+			for (Attribute attribute : attributeStatement.getAttributes()) {
+				List<String> attributeValues = new ArrayList<>();
+				for (XMLObject xmlObject : attribute.getAttributeValues()) {
+					attributeValues.add(getXmlObjectValue(xmlObject));
```
jzheaux, line 489: It seems odd to add null values to the list. Is it necessary to preserve that aspect of the attribute statement?

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProvider.java
```diff
@@ -474,6 +480,33 @@ else if (subject.getEncryptedID() != null) {
...
+	}
+
+	private String getXmlObjectValue(XMLObject xmlObject) {
+		if (xmlObject == null) {
+			return null;
+		}
+		if (xmlObject instanceof XSString) {
+			((XSString) xmlObject).getValue();
+		}
+		if (xmlObject instanceof XSAnyImpl) {
```
jzheaux, line 504: I think there are a few more types here that are relevant like `XSBoolean`, `XSDateTime`, `XSInteger`, and `XSURI`.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AuthenticatedPrincipal.java
```diff
@@ -16,13 +16,37 @@
...
  * @author Clement Stoquart
  * @since 5.2.2
  */
 public interface Saml2AuthenticatedPrincipal extends AuthenticatedPrincipal {
+	/**
+	 * Get the Saml2 token attribute by name
+	 *
+	 * @param name the name of the attribute
+	 * @param <A> the type of the attribute
+	 * @return the attribute or {@code null} otherwise
```
jzheaux, line 37: To each public method you add in this PR, please add a `@since 5.4`.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/SimpleSaml2AuthenticatedPrincipal.java
```diff
@@ -27,13 +29,25 @@
 class SimpleSaml2AuthenticatedPrincipal implements Saml2AuthenticatedPrincipal, Serializable {
 
 	private final String name;
+	private final Map<String, Object> attributes;
 
 	SimpleSaml2AuthenticatedPrincipal(String name) {
```
jzheaux, line 34: Is there still a use case in the application for this constructor? Since this is a package-private class, we should consider removing it.

jzheaux:
In addition to the inline feedback, would you also please add unit tests that show the feature working? I'd recommend tests that show how attributes statements of various arrangements and data types will work.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProvider.java
```diff
@@ -494,6 +494,52 @@ private NameID decrypt(Saml2AuthenticationToken token, EncryptedID assertion)
...
+		if (xmlObject instanceof XSURI) {
+			return ((XSURI) xmlObject).getValue();
+		}
+		if (xmlObject instanceof XSBoolean) {
+			XSBooleanValue xsBooleanValue = ((XSBoolean) xmlObject).getValue();
+			return xsBooleanValue != null ? xsBooleanValue.getValue() : null;
+		}
+		if (xmlObject instanceof XSDateTime) {
+			DateTime dateTime = ((XSDateTime) xmlObject).getValue();
+			return dateTime != null ? Instant.ofEpochSecond(dateTime.getMillis()) : null;
```
kostic017, line 538: I wasn't sure what to do with `org.joda.time.DateTime`. `XSDateTime` also has `org.joda.time.formatter.DateTimeFormatter` reference (if that's relevant).

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProvider.java
```diff
@@ -494,6 +494,52 @@ private NameID decrypt(Saml2AuthenticationToken token, EncryptedID assertion)
...
+		}
+		return attributeMap;
+	}
+
+	private Object getXmlObjectValue(XMLObject xmlObject) {
+		if (xmlObject == null) {
+			return null;
+		}
+		if (xmlObject instanceof XSAnyImpl) {
+			return ((XSAnyImpl) xmlObject).getTextContent();
```
jzheaux, line 521: Since `XSAny` can have much more than text content, it makes me wonder whether we'll be losing information by only returning the text content.

I think let's skip this type for now (don't add it into the `Map`), we can add it later when we've got a better understanding of what use cases are still missing.

kostic017: Well... That's the only type of attribute values that we get from our server :)
We only get the name of an attribute value for some reason, without the type argument.

saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/authentication/SimpleSaml2AuthenticatedPrincipalTests.java
```diff
@@ -19,11 +19,13 @@
...
 import org.junit.Test;
 
+import java.util.LinkedHashMap;
+
 public class SimpleSaml2AuthenticatedPrincipalTests {
 
 	@Test
 	public void createSimpleSaml2AuthenticatedPrincipal() {
-		SimpleSaml2AuthenticatedPrincipal principal = new SimpleSaml2AuthenticatedPrincipal("user");
+		SimpleSaml2AuthenticatedPrincipal principal = new SimpleSaml2AuthenticatedPrincipal("user", new LinkedHashMap<>());
```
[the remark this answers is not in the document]
kostic017: I'll see what I can do about that, I don't write tests often :)

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProvider.java
kostic017: I don't know how common this behavior is but [looks valid](https://lists.oasis-open.org/archives/saml-dev/201106/msg00001.html).

While we have an `XSAnyImpl` instance, it's QName is `{urn:oasis:names:tc:SAML:2.0:assertion}AttributeValue` and there is no marshaler for that, so I'm getting NPE both with the unit and e2e test.

This method was used in the [spring-security-saml](https://github.com/spring-projects/spring-security-saml/blob/708ab9839701afeafe3e430f4cb287cd58194c62/core/src/main/java/org/springframework/security/saml/SAMLCredential.java#L216) to get attribute value as a string.

jzheaux: Perhaps something like this would be more robust:

```java
Map<QName, Marshaller> marshallers =
		XMLObjectProviderRegistrySupport.getMarshallerFactory().getMarshallers();
if (marshallers.containsKey(xmlObject.getSchemaType())) {
	return this.saml.serialize(xmlObject);
}
if (marshallers.containsKey(xmlObject.getElementQName())) {
	return this.saml.serialize(xmlObject);
}
return ((XSAnyImpl) xmlObject).getTextContent();
```
That way, an application can override the marshalling behavior if they need to by configuring OpenSAML's `MarshallerFactory`.

kostic017: I'm ok with that as long as I can unregister the marshaller if it gets registered by default in some future versions since text content satisfies my needs.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AuthenticatedPrincipal.java
```diff
@@ -16,13 +16,39 @@
...
+	 * @since 5.4
+	 */
+	@Nullable
+	default <A> A getAttribute(String name) {
```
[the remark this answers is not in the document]
kostic017: Yeah, I prefer the first option, but for one user attribute (tenant) in my app I'm expecting multiple values. Maybe we should provide both options?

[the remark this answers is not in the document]
kostic017: I forgot about `getAttributes` method. I could use that in that one case if we only provide the first option.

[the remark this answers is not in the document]
jzheaux: I agree that to get the underlying list you can call `getAttributes().get("tenant")`.

Since there is the possibility of adding both somewhere down the road, I'm leaning more heavily now towards renaming this method to `getFirstAttribute`. I believe this aligns nicely with `MultiValueMap` which is similar in concept and also has a method called `getFirst()`.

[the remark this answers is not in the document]
kostic017: Hm, after changing `getAttributes` return type, I can't do this anymore
```
List<String> tenants = (List<String>) principal.getAttributes().get("tenant");
```
since the compiler doesn't allow casting from `List<Object>` to `List<String>`.

And this doesn't look very nice: 
```
List<Object> tenants2 = principal.getAttributes().get("tenant");
```

If I had a method like this
```
public <A> List<A> getAttributeValue(String name) {
    return (List<A>) getAttributes().get(name);
}
```
I could do
```
List<String> tenants = principal.getAttributeValue("tenant");
```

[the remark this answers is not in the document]
kostic017: On the other hand, what if those values have different types... I'm so used to working with only one value type.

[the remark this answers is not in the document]
jzheaux: Good catch! A couple thoughts come to mind:

1. Since this caught you and I off-guard, it would be nice if there was a test in `SimpleSaml2AuthenticatedPrincipalTests` that demonstrated the various behaviors. That way our expectations are codified in a test.
2. I think adding a `List` getter makes sense. An application can set the reference type to `List<Object>` if their attribute contains multiple types.

Also, I'd recommend that the interface stick with `getAttribute` and `getFirstAttribute`. The reason for that is it aligns with similar Spring Security support for OAuth 2.0, which uses `getClaim` and `getAttribute`. As another example, this is the same convention followed by the Servlet spec with `HttpServletRequest#getParameter`, `#getAttribute`, and `#getHeader`.

If you agree, then the contract would become:

```java
Map<String, List<Object>> getAttributes();

<A> List<A> getAttribute(String name);

<A> A getFirstAttribute(String name);
```

kostic017:
@jzheaux I think you added duplicate label by mistake.

jzheaux:
Thanks for the additional changes, @kostic017.

In preparation for merging, will you please do three more things?

* Make sure that any file you modified has a copyright header that extends to `2020`?
* Squash your commits into one
* Format your final commit message to include the phrase `Closes gh-xxx` in the commit body, e.g.

```
Add SAML Attribute Support

Closes gh-8661
```

jzheaux:
Regarding the duplicate label, it "duplicates" the linked issue. When looking at the tickets completed in a milestone, it helps remove some noise from the report.

jzheaux:
Excellent, @kostic017! This is now merged into `master` via eed33228f4c29b2806c4be21b81fea0ed1892f2c.

I also applied a small polish via 360db53dd24d2d12316dcbdaaa7bc4cba59ea0a5 and some documentation via 8cbdcfe75610f865917a0d371211e22dcc7a10e6

Issue #8675, opened 2020-06-10
Change getMethod() to return configured value in SimpleSavedRequest


sentinalll:
**Expected Behavior**

getMethod() returns value configured for "method" property

**Current Behavior**

returns null

**Context**

It causes failures when the request is resolved on authentication
As a workaround SimpleSavedRequest can be extended with overridden getMethod method

jzheaux:
Good catch, @sentinalll. I think the intent of this class is to save the method indicated in the constructor that takes a `SavedRequest` instance.

Would you be interested in submitting a PR to update the `SimpleSavedRequest(SavedRequest)` constructor to set the `method` field? Also, to have `getMethod()` return the `method` field?

sentinalll:
Hi @jzheaux, sorry for the delay. Sure, I can work on a PR. With `SimpleSavedRequest` I am trying to handle `ajax` request when the session is already expired. Check if it is `ajax` request -> save `Referer` URL in `SimpleSavedRequest` and put it as a session attribute -> set location header to authorization URI and status unauthorized.

jzheaux:
So, am I following correctly to assume that you only need to store the URL and you don't actually need the method?

sentinalll:
Yes, you are right. I am using `SimpleSavedRequest(String redirectUrl)` constructor. And it fails on a callback from Authorization Server with an error:

```
java.lang.NullPointerException: Name is null
	at java.lang.Enum.valueOf(Enum.java:236) ~[na:1.8.0_202]
	at org.springframework.web.bind.annotation.RequestMethod.valueOf(RequestMethod.java:35) ~[spring-web-5.2.6.RELEASE.jar:5.2.6.RELEASE]
	at org.springframework.web.servlet.mvc.condition.RequestMethodsRequestCondition.matchRequestMethod(RequestMethodsRequestCondition.java:148) ~[spring-webmvc-5.2.6.RELEASE.jar:5.2.6.RELEASE]
	at org.springframework.web.servlet.mvc.condition.RequestMethodsRequestCondition.getMatchingCondition(RequestMethodsRequestCondition.java:127) ~[spring-webmvc-5.2.6.RELEASE.jar:5.2.6.RELEASE]
	at org.springframework.web.servlet.mvc.method.RequestMappingInfo.getMatchingCondition(RequestMappingInfo.java:221) ~[spring-webmvc-5.2.6.RELEASE.jar:5.2.6.RELEASE]
[...]

PR #8686, opened 2020-06-12
Updated SimpleSavedRequest#getMethod

sentinalll:
Before:
1. SimpleSavedRequest#getMethod returned null
2. SimpleSavedRequest(SavedRequest request) constructor did not set the method field from the request

After:
1. SimpleSavedRequest#getMethod returns method property value
2. SimpleSavedRequest(SavedRequest request) constructor sets the method field from request

Closes gh-8675

I have submitted the CLA

### DIFF
--- a/web/src/main/java/org/springframework/security/web/savedrequest/SimpleSavedRequest.java
+++ b/web/src/main/java/org/springframework/security/web/savedrequest/SimpleSavedRequest.java
@@ -57,6 +57,7 @@ public class SimpleSavedRequest implements SavedRequest {
 		}
 		this.locales = request.getLocales();
 		this.parameters = request.getParameterMap();
+		this.method = request.getMethod();
 	}
 
 	@Override
@@ -71,7 +72,7 @@ public class SimpleSavedRequest implements SavedRequest {
 
 	@Override
 	public String getMethod() {
-		return null;
+		return this.method;
 	}
 
 	@Override
--- a/web/src/test/java/org/springframework/security/web/savedrequest/SimpleSavedRequestTests.java
+++ b/web/src/test/java/org/springframework/security/web/savedrequest/SimpleSavedRequestTests.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.web.savedrequest;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import javax.servlet.http.Cookie;
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class SimpleSavedRequestTests {
+
+	@Test
+	public void constructorWhenGivenSavedRequestThenCopies() {
+		SavedRequest savedRequest = new SimpleSavedRequest(prepareSavedRequest());
+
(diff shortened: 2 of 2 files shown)


web/src/test/java/org/springframework/security/web/savedrequest/SimpleSavedRequestTests.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.Map;
+import javax.servlet.http.Cookie;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class SimpleSavedRequestTests {
```
jzheaux, line 31: Would you please add tests to show that the `String` constructor works, too?

web/src/main/java/org/springframework/security/web/savedrequest/SimpleSavedRequest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2004, 2005, 2006 Acegi Technology Pty Limited
+ * Copyright 2004, 2005, 2006, 2020 Acegi Technology Pty Limited
```
jzheaux, line 2: Thank you for updating this; however, the Acegi copyright headers are a funny exception to the rule for now. Please leave them as-is.

web/src/test/java/org/springframework/security/web/savedrequest/SimpleSavedRequestTests.java
```diff
@@ -0,0 +1,93 @@
...
+		when(savedRequestMock.getCookies()).thenReturn(Collections.singletonList(new Cookie("cookiename", "cookievalue")));
+		when(savedRequestMock.getHeaderNames()).thenReturn(Collections.singletonList("headername"));
+		when(savedRequestMock.getHeaderValues("headername")).thenReturn(Collections.singletonList("headervalue"));
+		when(savedRequestMock.getLocales()).thenReturn(Collections.singletonList(Locale.ENGLISH));
+		when(savedRequestMock.getParameterMap()).thenReturn(Collections.singletonMap("key", new String[] {"value"}));
+		this.savedRequest = new SimpleSavedRequest(savedRequestMock);
+	}
+
+	@Test
+	public void getMethod() {
```
jzheaux, line 48: Since this is a new test, will you please change these test names to follow this convention:

`{methodName}When{Input}Then{Output}`

You might also consider consolidating these tests into `constructorWhenGivenSavedRequestThenCopies` or something similar.

web/src/test/java/org/springframework/security/web/savedrequest/SimpleSavedRequestTests.java
```diff
@@ -0,0 +1,93 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class SimpleSavedRequestTests {
+
+	private SimpleSavedRequest savedRequest;
+
+	@Before
+	public void setUp() {
+		SavedRequest savedRequestMock = mock(SavedRequest.class);
```
jzheaux, line 37: Instead of a mock, if you constructed a `SimpleSavedRequest` with a redirect URL and then called the setters, you'd get a similar outcome while at the same time testing more of the class.

web/src/test/java/org/springframework/security/web/savedrequest/SimpleSavedRequestTests.java
sentinalll: Hi, I've added one more test to cover the case with String constructor. Please take a look

web/src/main/java/org/springframework/security/web/savedrequest/SimpleSavedRequest.java
sentinalll: Okay, I'll revert it.

web/src/test/java/org/springframework/security/web/savedrequest/SimpleSavedRequestTests.java
sentinalll: Thank you for your review. I've changed the test according to your suggestions.

web/src/test/java/org/springframework/security/web/savedrequest/SimpleSavedRequestTests.java
sentinalll: Improved as suggested

jzheaux:
Thanks for the PR, @sentinalll! This is now merged into `master`.

Issue #7154, opened 2019-07-26
Replace internal Stream usage with for loops


rwinch:
### Summary

While some prefer the readability of using the Stream APIs, the GC overhead of creating additional objects (including intermediate objects) can cause significant decrease in performance There are some simple [benchmarks](https://gist.github.com/mp911de/7f203d787c6fdaf97dc91e6dc3aa73d3) that illustrate the problem.

We should replace Stream usage with for loops throughout Spring Security's code base.

Related gh-7155

eddumelendez:
Hi @rwinch, do you think this is good for `first-timers-only`?

kostya05983:
Hello @rwinch, I have an experience with stream api and others streams such as rxjava, I can take it in work. Where can I start? is stream api used in all modules?

rwinch:
Thanks for the offer @kostya05983!

A search shows the following locations contain an import to stream APIs. Note that we should only change private internal code. Please note:

* We should not change any public APIs only internal implementation code
* We do not need to change test code
* We do not need to change sample code

```
samples/boot/oauth2authorizationserver/src/main/java/sample/AuthorizationServerConfiguration.java
core/src/main/java/org/springframework/security/authorization/AuthorityReactiveAuthorizationManager.java
core/src/main/java/org/springframework/security/converter/RsaKeyConverters.java
core/src/main/java/org/springframework/security/core/userdetails/MapReactiveUserDetailsService.java
config/src/test/java/org/springframework/security/config/http/MiscHttpConfigTests.java
config/src/test/java/org/springframework/security/config/http/SecurityFiltersAssertions.java
config/src/test/java/org/springframework/security/config/doc/SpringSecurityXsdParser.java
config/src/test/java/org/springframework/security/config/doc/XsdDocumentedTests.java
config/src/test/java/org/springframework/security/config/doc/XmlNode.java
web/src/main/java/org/springframework/security/web/header/writers/ClearSiteDataHeaderWriter.java
web/src/test/java/org/springframework/security/web/server/header/ClearSiteDataServerHttpHeadersWriterTests.java
web/src/main/java/org/springframework/security/web/server/header/CompositeServerHttpHeadersWriter.java
web/src/main/java/org/springframework/security/web/server/header/ClearSiteDataServerHttpHeadersWriter.java
web/src/main/java/org/springframework/security/web/server/authentication/DelegatingServerAuthenticationSuccessHandler.java
[...]

PR #8695, opened 2020-06-16
ProviderManager Should Use CollectionUtils#contains

### LINKED ISSUE #8689: ProviderManager#checkState() throws NullPointerException
**Describe the bug**

When use [`java.util.List.of()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/List.html#of(E)), [`ProviderManager#checkState()`](https://github.com/spring-projects/spring-security/blob/5.3.3.RELEASE/core/src/main/java/org/springframework/security/authentication/ProviderManager.java#L148) throws `NullPointerException`.

`checkState()` uses `List#contains(null)` since https://github.com/spring-projects/spring-security/issues/7713#issuecomment-569741294, but [it may throws `NullPointerException`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/List.html#contains(java.lang.Object)):

> NullPointerException - if the specified element is null and this list does not permit null elements ([optional](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Collection.html#optional-restrictions))

**To Reproduce**

    new ProviderManager(List.of(myProvider));

**Expected behavior**

`NPE` is not thrown.

yukihane:
Use `stream().anyMatch(Objects::isNull)` insead of `contains(null)`, which may throw `NullPointerException`.

Fixes #8689

### DIFF
--- a/core/src/main/java/org/springframework/security/authentication/ProviderManager.java
+++ b/core/src/main/java/org/springframework/security/authentication/ProviderManager.java
@@ -30,6 +30,7 @@ import org.springframework.security.core.AuthenticationException;
 import org.springframework.security.core.CredentialsContainer;
 import org.springframework.security.core.SpringSecurityMessageSource;
 import org.springframework.util.Assert;
+import org.springframework.util.CollectionUtils;
 
 /**
  * Iterates an {@link Authentication} request through a list of
@@ -145,7 +146,7 @@ public class ProviderManager implements AuthenticationManager, MessageSourceAwar
 			throw new IllegalArgumentException(
 					"A parent AuthenticationManager or a list "
 							+ "of AuthenticationProviders is required");
-		} else if (providers.contains(null)) {
+		} else if (CollectionUtils.contains(providers.iterator(), null)) {
 			throw new IllegalArgumentException(
 					"providers list cannot contain null values");
 		}
--- a/core/src/test/java/org/springframework/security/authentication/ProviderManagerTests.java
+++ b/core/src/test/java/org/springframework/security/authentication/ProviderManagerTests.java
@@ -102,6 +102,31 @@ public class ProviderManagerTests {
 		new ProviderManager((AuthenticationProvider) null);
 	}
 
+	@Test(expected = IllegalArgumentException.class)
+	public void testStartupFailsIfProvidersContainNullElement() {
+		new ProviderManager(Arrays.asList(mock(AuthenticationProvider.class), null));
+	}
+
+	@Test
+	public void testUsingNullNotPermittedList() {
+		// imitated Java9 List.of(e) object, which disallows null elements and
+		// throws NPE when contains(null) called
+		List<AuthenticationProvider> providers = new ArrayList<AuthenticationProvider>() {
+			private static final long serialVersionUID = 1L;
+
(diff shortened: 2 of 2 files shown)


pivotal-issuemaster:
@yukihane Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=8695)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=8695) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@yukihane Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=8695)!

core/src/main/java/org/springframework/security/authentication/ProviderManager.java
```diff
@@ -145,7 +145,7 @@ private void checkState() {
 			throw new IllegalArgumentException(
 					"A parent AuthenticationManager or a list "
 							+ "of AuthenticationProviders is required");
-		} else if (providers.contains(null)) {
+		} else if (providers.stream().anyMatch(Objects::isNull)) {
```
jzheaux, line 148: At the framework level, we find that the Stream API is not fast enough. Would you please use a for loop instead?

core/src/test/java/org/springframework/security/authentication/ProviderManagerTests.java
```diff
@@ -102,6 +102,26 @@ public void testStartupFailsIfProvidersNotSetAsVarargs() {
 		new ProviderManager((AuthenticationProvider) null);
 	}
 
+	@Test
+	public void testUsingNullNotPermittedList() {
```
jzheaux, line 106: I think this test has the right idea. It seems like the list ought to have a `null` element, though, to fully describe the use case.

core/src/main/java/org/springframework/security/authentication/ProviderManager.java
rwinch: In particular we avoid the Stream APIs because they create extra objects which leads to excessive Garbage Collection. This can cause the JDK to pause and deteriorate the performance. See gh-7154

I believe we could use Spring's [CollectionUtils.contains](https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/util/CollectionUtils.java#L125)

yukihane: I've understood it and amended to use `CollectionUtils.contains`.

core/src/test/java/org/springframework/security/authentication/ProviderManagerTests.java
yukihane: I assumed it already existed.
I've created additional test case.

jzheaux:
Thanks, @yukihane! This is now merged into master via 5302fb776cb39d304fd6e29f9d44a30579d2fbd8 and c177b391d4b4bab839b8ab9afe3f6b1ff99496a0

Issue #7027, opened 2019-06-20
Make RestTemplate used by ClientRegistrations (Discovery) configurable


shibuyaku:
### Summary

spring-security-oauth2-client uses a RestTemplate for openid/oauth Discovery that is not configurable. This is unuseable in scenarios where you need to adjust the RestTemplate. Example: You need to use a proxy and configure auth. 

### Actual Behavior
ClientRegistrations class uses a RestTemplate for doing OpenId Discovery that is not configurable, since it is not using RestTemplateBuilder or something comparable.

OpenId discovery is done by querying issuerUri + "/.well-known/openid-configuration" (for oidc) or isserUri + "/.well-known/oauth-authorization-server" (for oauth).

Current implementation:
```RestTemplate rest = new RestTemplate()```

### Expected Behavior
ClientRegistrations should use a configurable RestTemplate for doing OpenId Discovery. One should be able to configure the requestFactory, interceptors, errorHandler and so on of that RestTemplate.

### Version
5.2.0.M3 and 5.1.5.RELEASE

### Sample
https://github.com/spring-projects/spring-security/blob/1739ef8d3c75871367e8bf9a0daf46547abf6d45/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java#L58

Related #5607

**Comment by shibuyaku** [comment]
We usually use org.springframework.boot.web.client.RestTemplateBuilder to achieve this in our code, but I can not immediately tell if that would be a valid approach in the spring-security project.

If so:
```RestTemplateBuilder().build()``` instead of ``RestTemplate()``` would allow a user to configure everything he wants using org.springframework.boot.web.client.RestTemplateCustomizer.

**Comment by shibuyaku** [comment]
FYI: Found 6 more occurences of new RestTemplate() that will also be affected and unconfigurable, 2 of them in spring-security-oauth2-client.

We stumbled upon this because the RestTemplate() will try to use the host/port of a proxy that is configured via environment variables, but ignores the user and password. 

Complete list:
```
org.springframework.security.oauth2.client.registration.ClientRegistrations
org.springframework.security.oauth2.client.userinfo.DefaultOAuth2UserService
org.springframework.security.oauth2.client.userinfo.CustomUserTypesOAuth2UserService

org.springframework.security.oauth2.jwt.JwtDecoders
org.springframework.security.oauth2.jwt.NimbusJwtDecoder
[...]

PR #8732, opened 2020-06-19
Register RestOperations @Bean for OAuth 2.0 Client

jgrandja:
Issue gh-8882

This PR addresses the Servlet implementation for OAuth 2.0 Client.

The changes in this PR will also be required for OAuth 2.0 Client WebFlux.

In addition to OAuth 2.0 Client, similar changes will need to be applied to OAuth 2.0 Resource Server and the JOSE stack.

I've done all the legwork as to which changes are required for the follow-up PR's:

### OAuth 2.0 Client WebFlux

```java
WebClientReactiveAuthorizationCodeTokenResponseClient
WebClientReactiveRefreshTokenTokenResponseClient
WebClientReactiveClientCredentialsTokenResponseClient
WebClientReactivePasswordTokenResponseClient

DefaultReactiveOAuth2UserService
```

### OAuth 2.0 Resource Server

```java
NimbusOpaqueTokenIntrospector
```

### OAuth 2.0 Resource Server WebFlux

```java
NimbusReactiveOpaqueTokenIntrospector
```

### JOSE

```java
NimbusJwtDecoder
NimbusJwtDecoderJwkSupport
JwtDecoderProviderConfigurationUtils
JwtDecoders
```

### JOSE WebFlux

```java
NimbusReactiveJwtDecoder
ReactiveRemoteJWKSource
ReactiveJwtDecoders
```

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configuration/OAuth2ClientConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configuration/OAuth2ClientConfiguration.java
@@ -16,7 +16,6 @@
 package org.springframework.security.config.annotation.web.configuration;
 
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.beans.factory.support.BeanDefinitionRegistryPostProcessor;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
@@ -24,19 +23,16 @@ import org.springframework.context.annotation.Import;
 import org.springframework.context.annotation.ImportSelector;
 import org.springframework.core.type.AnnotationMetadata;
 import org.springframework.security.config.oauth2.client.DefaultOAuth2AuthorizedClientManagerPostProcessor;
-import org.springframework.security.config.oauth2.client.OAuth2ClientBeanNames;
 import org.springframework.security.config.oauth2.client.OAuth2ClientRestOperationsPostProcessor;
 import org.springframework.security.oauth2.client.OAuth2AuthorizedClientProvider;
 import org.springframework.security.oauth2.client.OAuth2AuthorizedClientProviderBuilder;
 import org.springframework.security.oauth2.client.endpoint.OAuth2AccessTokenResponseClient;
 import org.springframework.security.oauth2.client.endpoint.OAuth2ClientCredentialsGrantRequest;
 import org.springframework.security.oauth2.client.registration.ClientRegistrationRepository;
-import org.springframework.security.oauth2.client.registration.ClientRegistrations;
 import org.springframework.security.oauth2.client.web.DefaultOAuth2AuthorizedClientManager;
 import org.springframework.security.oauth2.client.web.OAuth2AuthorizedClientRepository;
 import org.springframework.security.oauth2.client.web.method.annotation.OAuth2AuthorizedClientArgumentResolver;
 import org.springframework.util.ClassUtils;
-import org.springframework.web.client.RestOperations;
 import org.springframework.web.method.support.HandlerMethodArgumentResolver;
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
 
@@ -55,8 +51,7 @@ import java.util.Optional;
  * @see OAuth2ImportSelector
  */
 @Import({OAuth2ClientConfiguration.OAuth2ClientSecurityConfiguration.class,
-		OAuth2ClientConfiguration.OAuth2ClientWebMvcImportSelector.class,
-		OAuth2ClientConfiguration.OAuth2ClientRegistrationsConfiguration.class})
+		OAuth2ClientConfiguration.OAuth2ClientWebMvcImportSelector.class})
 final class OAuth2ClientConfiguration {
 
 	@Configuration(proxyBeanMethods = false)
@@ -132,16 +127,6 @@ final class OAuth2ClientConfiguration {
 		void setAccessTokenResponseClient(
 				Optional<OAuth2AccessTokenResponseClient<OAuth2ClientCredentialsGrantRequest>> accessTokenResponseClient) {
 			accessTokenResponseClient.ifPresent(client -> this.accessTokenResponseClient = client);
-		}
-	}
-
-	@Configuration(proxyBeanMethods = false)
-	static class OAuth2ClientRegistrationsConfiguration {
-
-		@Autowired
-		@Qualifier(OAuth2ClientBeanNames.REST_OPERATIONS)
-		void configure(RestOperations restOperations) {
-			ClientRegistrations.setRestOperations(restOperations);
 		}
 	}
 }
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configuration/OAuth2ClientConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configuration/OAuth2ClientConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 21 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2ClientConfigurerUtils.java
```diff
@@ -96,4 +98,9 @@ private OAuth2ClientConfigurerUtils() {
 		}
 		return (!authorizedClientServiceMap.isEmpty() ? authorizedClientServiceMap.values().iterator().next() : null);
 	}
+
+	static <B extends HttpSecurityBuilder<B>> RestOperations getRestOperationsBean(B builder) {
+		return builder.getSharedObject(ApplicationContext.class).getBean(
```
rwinch, line 103: This looks like it would fail if there isn't a bean by that name and type or if there is more than one bean of this type.

rwinch: Is there a reason you think we should use a bean name here? We don't typically do the lookups by bean name.

config/src/main/java/org/springframework/security/config/annotation/web/configuration/OAuth2ClientConfiguration.java
```diff
@@ -91,23 +115,33 @@ public void addArgumentResolvers(List<HandlerMethodArgumentResolver> argumentRes
...
 		}
 	}
+
+	@Configuration(proxyBeanMethods = false)
+	static class OAuth2ClientRegistrationsConfiguration {
+
+		@Autowired
+		@Qualifier(OAuth2ClientBeanNames.REST_OPERATIONS)
+		void configure(RestOperations restOperations) {
+			ClientRegistrations.setRestOperations(restOperations);
```
rwinch, line 144: Setting a static method from a bean definition is not a typical pattern I find in Spring. Instead, I'd expect to be able to provide the RestOperations to the builder. A few concerns around this design:

First, it is not easy to provide different ``RestOperation``s for different issuers in a thread safe way. The builders should ideally be thread safe. This is especially true, given the performance improvements being made to Spring and talks around making bean creation multi-threaded.

Another and more immediate concern is that it appears that there is likely a bean ordering issue here. What guarantees this will happen before `ClientRegistrations` is used to create a client registration?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2ClientConfigurerUtils.java
jgrandja: > This looks like it would fail if there isn't a bean by that name and type or if there is more than one bean of this type

`OAuth2ClientRestOperationsPostProcessor` defined in `OAuth2ClientConfiguration` will register the `@Bean` using `OAuth2ClientBeanNames.REST_OPERATIONS` so it won't fail. It's expected to be registered.

> Is there a reason you think we should use a bean name here?

I thought this is what we decided on, as per [comment](https://github.com/spring-projects/spring-security/issues/5607#issuecomment-615496025)

Furthermore, an application may registered one or more `RestOperations` `@Bean` so we need to ensure we pick up the correct `RestOperations` `@Bean` to be used for the client flows.

Also, see tests in `OAuth2ClientConfigurationTests`  starting at `loadContextWhenPostProcessedThenRestOperationsRegistered()` which demonstrates that the `@Bean` is registered by default and how a user could override the `@Bean`

config/src/main/java/org/springframework/security/config/annotation/web/configuration/OAuth2ClientConfiguration.java
jgrandja: > Setting a static method from a bean definition is not a typical pattern I find in Spring.

Understood. This is the simplest solution I came up. We're kind of in a bind with this class since it's not designed to be a `@Bean` and instead a utility class.

> I'd expect to be able to provide the RestOperations to the builder

There is no builder in `ClientRegistrations`? I'm don't understand. Can you clarify what you mean here?

> Another and more immediate concern is that it appears that there is likely a bean ordering issue here

See test `OAuth2ClientConfigurationTests.loadContextWhenRestOperationsRegisteredThenClientRegistrationsUses()` as it passes.

Honestly, I'm not a fan of this static setter either and would be fine not exposing it. But then how do we address #7027? Do we introduce #5543 and deprecate `ClientRegistrations`?

config/src/main/java/org/springframework/security/config/annotation/web/configuration/OAuth2ClientConfiguration.java
```diff
@@ -46,19 +50,34 @@
...
+	static class OAuth2ClientSecurityConfiguration {
+
+		@Bean
+		BeanDefinitionRegistryPostProcessor oauth2ClientRestOperationsPostProcessor() {
+			return new OAuth2ClientRestOperationsPostProcessor();
+		}
+
+		@Bean
+		BeanDefinitionRegistryPostProcessor defaultOAuth2AuthorizedClientManagerPostProcessor() {
+			return new DefaultOAuth2AuthorizedClientManagerPostProcessor();
```
jzheaux, line 67: I'm having trouble seeing the value of registering `OAuth2AuthorizedClientManager` with a known bean name.

Doing so with `RestOperations` makes sense since there are many use cases when an application publishes a `RestOperations` but doesn't intend it to be used by Spring Security.

What is it about `OAuth2AuthorizedClientManager` that requires the same treatment as `RestOperations`? Is this a needed pattern for other OAuth 2.0 Client beans?

jgrandja: The `OAuth2AuthorizedClientManager` is the central interface for client since 5.2. It is used primarily by `ServerOAuth2AuthorizedClientExchangeFilterFunction` and `OAuth2AuthorizedClientArgumentResolver`. It indirectly uses `RestOperations`. The relationship chain is as follows:

`OAuth2AuthorizedClientManager` has 1 or more `OAuth2AuthorizedClientProvider`
`OAuth2AuthorizedClientProvider` has an `OAuth2AccessTokenResponseClient`
`OAuth2AccessTokenResponseClient` has a `RestOperations`

If an application configures a custom `RestOperations` then it will automatically be configured with the "default" `OAuth2AuthorizedClientManager` and associated `OAuth2AuthorizedClientProvider`'s, and the user could leverage this default `@Bean` instead of explicitly configuring it. It really is meant to be a convenience mechanism.

jgrandja: One important thing to note is that an application may configure more than 1 `OAuth2AuthorizedClientManager` in their setup. For example, they may have a background service that performs `client_credentials` grant and another `OAuth2AuthorizedClientManager` that handles `authorization_code` and `refresh_token` in the web context.

The "default" `OAuth2AuthorizedClientManager` will be wired into the `OAuth2AuthorizedClientArgumentResolver` and may be used by `ServletOAuth2AuthorizedClientExchangeFilterFunction`.

config/src/main/java/org/springframework/security/config/annotation/web/configuration/OAuth2ClientConfiguration.java
rwinch: > There is no builder in ClientRegistrations? I don't understand. Can you clarify what you mean here?

Sorry for being unclear. An option would be to introduce a builder to ClientRegistrations that allows passing in the `RestOperations` that would be used.

> Honestly, I'm not a fan of this static setter either and would be fine not exposing it. But then how do we address #7027? Do we introduce #5543 and deprecate ClientRegistrations?

I think an option would be that Spring Security leverages the configured RestOperations to pass into the newly added builder in ClientRegistrations.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2ClientConfigurerUtils.java
rwinch: Rather than relying on `OAuth2ClientRestOperationsPostProcessor` could we just default the `RestOperations` if none was found? This is what we do in other circumstances.

jgrandja: > This is what we do in other circumstances

Can you please be specific and provide a code sample I could reference?

jgrandja: Are you referring to the way we conditionally `@Import` using `ImportSelector`? For example, `OAuth2ClientConfigurationOAuth2ClientWebMvcImportSelector`?

config/src/main/java/org/springframework/security/config/annotation/web/configuration/OAuth2ClientConfiguration.java
jgrandja: > An option would be to introduce a builder to ClientRegistrations

I'm not sure how this would work @rwinch? Maybe I'm missing something.

`ClientRegistrations` was designed as a utility `final` class with `static` methods.

What would `ClientRegistrations.Builder.build()` return? An instance of `ClientRegistrations`?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2ClientConfigurerUtils.java
rwinch: We often try to get a Bean if it doesn't exist we create a default instance. An [example is PasswordEncoder](https://github.com/spring-projects/spring-security/blob/5.4.0-M1/config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/AuthenticationConfiguration.java#L320-L330)

jgrandja:
Closing this PR as it needs more work. I'll submit a new PR when I get a chance to revisit this and apply the necessary changes.

PR #8734, opened 2020-06-20
Add RoleHierarchyBuilder for typesafe JavaConfig style role hierarchy configuration

baso53:
Added support for JavaConfig style builder for `RoleHierarchyImpl`.

I was pretty surprised that there isn't some typesafe way of configuring RoleHierarchy, in my opinion this is very much needed, so I created this PR.

There are also some changes in the `RoleHierarchyImpl` in the `getReachableGrantedAuthorities` method, which was pretty confusing, so I refactored it to use Java stream API to make it more functional. I don't know what's the policy on changing existing code, but it's just a refactoring, so if it's advised not to change, it can be reverted.

The tests for the builder are basically the same as for the `RoleHierarchyImpl`, which ensures that everything from there works with the builder as well.

BTW, I didn't see that the usual flow for new features is to open up a new issue - I just saw it now when opening this PR. Should I open a new issue or can it be discussed here?


marcusdacoregio:
Closing this given there hasn't been the updates (the discussion to improve the implementation did not happen)

Issue #6591, opened 2019-03-05
`@WithUserDetails` `setupBefore` does not work with JUnit 5


zmcode:
### Summary

When using JUnit 5 it seems that setting the `setupBefore` parameter of `@WithUserDetails` to `TestExecutionEvent.TEST_EXECUTION` does not work as intended. In fact it appears that the security context is created before the execution of a `@BeforeEach` method. 

### Actual Behavior

I wrote the following test class to showcase the issue

```java
@ExtendWith(SpringExtension.class)
@ContextConfiguration
class WithUserDetailsTest {

  private static String USERNAME;

  @BeforeEach
  void setUp() {
    USERNAME = "dummy";
  }

  @Test
  @WithUserDetails(
      userDetailsServiceBeanName = "testUserDetailsService",
      setupBefore = TestExecutionEvent.TEST_EXECUTION
  )
  void test() {
    TestUserDetails principal = (TestUserDetails) SecurityContextHolder.getContext()
        .getAuthentication().getPrincipal();
    Assertions.assertNotNull(principal.getUsername());
  }

  @Configuration
  static class TestUserDetailsConfiguration {

    @Bean("testUserDetailsService")
    UserDetailsService testUserDetailsService() {
      return s -> new TestUserDetails(USERNAME);
    }
  }

  static class TestUserDetails implements UserDetails {

    private static final long serialVersionUID = -6779736987183888865L;

    private String username;

    TestUserDetails(String username) {
      this.username = username;
    }

    @Override
    public String getUsername() {
      return username;
    }

    ...
  }
}
```
What happens here is that the test keeps failing because when the `UserDetails` is actually loaded the `USERNAME` variable is still null.

### Expected Behavior

I would expect the `@BeforeEach` method to be executed before the `SecurityContext` initialization therefore allowing the creation of a `UserDetails` with a non null username.

### Version

I am using Spring Boot Dependency `2.1.3.RELEASE` (=> Spring Security Test `5.1.4.RELEASE`).

jfriedenstab:
I stumbled upon this issue, too. I would like to create a custom `UserDetails` object in the `@BeforeEach` method which can then be used by `@WithUserDetails`.
Are there any plans to provide a fix for this?

piotrkaron:
For anyone with this problem, as this is still not resolved.
[...]

PR #8738, opened 2020-06-21
Fix SecurityContext creation for TEST_EXECUTION

MGabr:
Hi,

Currently, there is support for setting up a SecurityContext after @Before by
using TestExecutionEvent.TEST_EXECUTION. The current implementation, however,
already creates the SecurityContext in @Before and just does not set it yet.
This leads to issues like #6591. For the case of @WithUserDetails, the
creation of the SecurityContext already looks up a user from the repository.
If the user was inserted in @Before, the user is not found despite using
TestExecutionEvent.TEST_EXECUTION. This can be especially problematic with
data driven test frameworks like mongoUnit which initialize the database in @Before 
and drop the database after each test. In this case workarounds like described in
#6591 do not work.

This commit changes the creation of the SecurityContext to happen after @Before 
if using TestExecutionEvent.TEST_EXECUTION.

Closes gh-6591.

I have submitted the CLA.

### DIFF
--- a/test/src/main/java/org/springframework/security/test/context/support/WithSecurityContextTestExecutionListener.java
+++ b/test/src/main/java/org/springframework/security/test/context/support/WithSecurityContextTestExecutionListener.java
@@ -17,6 +17,7 @@ package org.springframework.security.test.context.support;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedElement;
+import java.util.function.Supplier;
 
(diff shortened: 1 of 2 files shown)


rwinch:
Thanks for the PR! This is merged into master via 97ee6d66f1b0d72c7c6e66730b7add4b737f026a I also added a test to validate that the creation of the SecurityContext was delayed via 97ee6d66f1b0d72c7c6e66730b7add4b737f026a

Issue #8735, opened 2020-06-20
Support JWTBearer Client Assertion Authentication


visweshwar:
**Expected Behavior**

One of the protocols support by OAuth is Client Assertion as authentication

A client uses an assertion to authenticate to the authorization server's token endpoint by using the "client_assertion_type" and "client_assertion" parameters

https://tools.ietf.org/html/rfc7521#section-6.1

**Current Behavior**
 
No Support

**Context**

This is something supported by jose.nimbus  and as consumer and provider of OAuth we want to secure our authentication to the token endpoint using this protocol.

jgrandja:
Closing as duplicate #6881

PR #8737, opened 2020-06-20
Adding Enhancement for Client Authentication using JWT Secret assertion

visweshwar:
One of the protocols support by OAuth is Client Assertion as authentication

A client uses an assertion to authenticate to the authorization server's token endpoint by using the "client_assertion_type" and "client_assertion" parameters

https://tools.ietf.org/html/rfc7521#section-6.1

Current Behavior

No Support

Context

This is something supported by jose.nimbus and as consumer and provider of OAuth we want to secure our authentication to the token endpoint using this protocol.

#8735

PR #8739, opened 2020-06-21
Feature/client assertion auth

visweshwar:
One of the protocols support by OAuth is Client Assertion as authentication

A client uses an assertion to authenticate to the authorization server's token endpoint by using the "client_assertion_type" and "client_assertion" parameters

https://tools.ietf.org/html/rfc7521#section-6.1

Current Behavior

No Support

Context

This is something supported by jose.nimbus and as consumer and provider of OAuth we want to secure our authentication to the token endpoint using this protocol.

#8735


jgrandja:
Closing as duplicate of #8737

Issue #6489, opened 2019-01-29
Simplify retrieving Introspection-specific attributes


jzheaux:
Related to #5200 and https://github.com/spring-projects/spring-security/pull/6352#discussion_r251856850

One way to simplify would be by creating an `IntrospectionClaimAccessor` similar to `JwtClaimAccessor`.

qavid:
Hi @jzheaux, I would like to work on this issue, could you please provide more details?

jzheaux:
@qavid  - I took a look at your commit, and I like where you seem to be headed with it. Can you clarify where you'd like additional detail?

jzheaux:
@qavid are you still thinking about submitting a PR for this issue?

qavid:
@jzheaux I would really like to, but i don't have much time. I'll try this weekend and I will let you know.

qavid:
Hi @jzheaux, I have submitted PR, but I have one question.
Documentation for AuthenticatedPrincipal#getName states that returned name is never null but according to RFC7519 (JWT) and RFC7662 (Token Introspection) SUB claim is in both cases optional. What should be returned in case there isn't SUB claim in JWT or Token Introspection? Previous implementation of DefaultOAuth2AuthenticatedPrincipal would also return null in this case.

I wasn't sure if DefaultOAuth2AuthenticatedPrincipal should be removed so I have removed it in a separate commit.

PR #8740, opened 2020-06-21
Simplify retrieving Introspection-specific attributes

qavid:
Fixes gh-6489

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/OAuth2IntrospectionAuthenticatedPrincipal.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/OAuth2IntrospectionAuthenticatedPrincipal.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.oauth2.server.resource.introspection;
+
+import static org.springframework.security.core.authority.AuthorityUtils.NO_AUTHORITIES;
+
+import java.io.Serializable;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Map;
+
+import org.springframework.security.core.GrantedAuthority;
+import org.springframework.security.oauth2.core.OAuth2AuthenticatedPrincipal;
+import org.springframework.util.Assert;
+
+/**
+ * A domain object that wraps the attributes of OAuth 2.0 Token Introspection.
+ *
+ * @author David Kovac
+ * @since 5.4
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7662#section-2.2">Introspection Response</a>
+ */
+public final class OAuth2IntrospectionAuthenticatedPrincipal implements OAuth2AuthenticatedPrincipal,
+		OAuth2IntrospectionClaimAccessor, Serializable {
+	private final Map<String, Object> attributes;
+	private final Collection<GrantedAuthority> authorities;
+	private final String name;
+
+	/**
+	 * Constructs an {@code OAuth2IntrospectionAuthenticatedPrincipal} using the provided parameters.
+	 *
+	 * @param attributes  the attributes of the OAuth 2.0 Token Introspection
+	 * @param authorities the authorities of the OAuth 2.0 Token Introspection
+	 */
+	public OAuth2IntrospectionAuthenticatedPrincipal(Map<String, Object> attributes,
+			Collection<GrantedAuthority> authorities) {
+
+		this(null, attributes, authorities);
+	}
+
+	/**
+	 * Constructs an {@code OAuth2IntrospectionAuthenticatedPrincipal} using the provided parameters.
+	 *
+	 * @param name        the name attached to the OAuth 2.0 Token Introspection
+	 * @param attributes  the attributes of the OAuth 2.0 Token Introspection
+	 * @param authorities the authorities of the OAuth 2.0 Token Introspection
+	 */
+	public OAuth2IntrospectionAuthenticatedPrincipal(String name, Map<String, Object> attributes,
+			Collection<GrantedAuthority> authorities) {
+
+		Assert.notEmpty(attributes, "attributes cannot be empty");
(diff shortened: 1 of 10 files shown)


oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtAuthenticatedPrincipal.java
```diff
@@ -14,70 +14,75 @@
...
  *
- * @author Clement Ng
- * @author Josh Cummings
- * @since 5.2
+ * @author David Kovac
+ * @since 5.4
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7519#section-4.1">Registered Claim Names</a>
  */
-public final class DefaultOAuth2AuthenticatedPrincipal implements OAuth2AuthenticatedPrincipal, Serializable {
+public final class JwtAuthenticatedPrincipal implements OAuth2AuthenticatedPrincipal, JwtClaimAccessor, Serializable {
```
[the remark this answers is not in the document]
jzheaux: Additionally, `OAuth2IntrospectionAuthenticatedPrincipal` only makes sense when OAuth 2.0 Introspection was performed to retrieve the token attributes. For applications that introspected the token, but not with OAuth 2.0 Introspection, they can use `DefaultOAuth2AuthenticatedPrincipal` for additional flexibility.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/OAuth2IntrospectionAuthenticatedPrincipal.java
```diff
@@ -0,0 +1,108 @@
...
+import org.springframework.util.Assert;
+
+/**
+ * A domain object that wraps the attributes of OAuth 2.0 Token Introspection.
+ *
+ * @author David Kovac
+ * @since 5.4
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7662#section-2.2">Introspection Response</a>
+ */
+public final class OAuth2IntrospectionAuthenticatedPrincipal implements OAuth2AuthenticatedPrincipal,
```
jzheaux, line 37: I wonder if this should be package-private for the time being. It seems like the major reason to introduce this class is to make it easier to do:

```java
public String method(@AuthenticationPrincipal(expression="subject") String subject) ...

public String method(@AuthenticationPrincipal OAuth2IntrospectionClaimAccessor accessor) ...
```
and so on.

There's no use case that I can see where an application must have an instance of `OAuth2IntrospectionAuthenticatedPrincipal` since it holds the same information as a `DefaultOAuth2AuthenticatedPrincipal`.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtAuthenticatedPrincipal.java
```diff
@@ -14,70 +14,75 @@
...
-	 * Constructs an {@code DefaultOAuth2AuthenticatedPrincipal} using the provided parameters.
+	 * Constructs an {@code JwtAuthenticatedPrincipal} using the provided parameters.
 	 *
-	 * @param attributes the attributes of the OAuth 2.0 token
-	 * @param authorities the authorities of the OAuth 2.0 token
+	 * @param attributes  the attributes of the OAuth 2.0 JSON Web Token
+	 * @param authorities the authorities of the OAuth 2.0 JSON Web Token
 	 */
-	public DefaultOAuth2AuthenticatedPrincipal(Map<String, Object> attributes,
+	public JwtAuthenticatedPrincipal(Map<String, Object> attributes,
```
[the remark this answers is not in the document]
qavid: I removed JwtAuthenticatedPrincipal, you are right it makes no sense and I don't see benefit in having it

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtAuthenticatedPrincipal.java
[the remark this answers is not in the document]
qavid: I mentioned it in issue that I was not sure if it can be removed (because it is a public class), but introspecting token without OAuth2 introspection is a good point

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/OAuth2IntrospectionAuthenticatedPrincipal.java
qavid: I have changed access level to package-private

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/OAuth2IntrospectionAuthenticatedPrincipal.java
```diff
@@ -0,0 +1,108 @@
...
+import org.springframework.util.Assert;
+
+/**
+ * A domain object that wraps the attributes of OAuth 2.0 Token Introspection.
+ *
+ * @author David Kovac
+ * @since 5.4
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7662#section-2.2">Introspection Response</a>
+ */
+final class OAuth2IntrospectionAuthenticatedPrincipal implements OAuth2AuthenticatedPrincipal,
```
jzheaux, line 37: Sorry for some belated second thoughts about this. Let's go ahead and make this public so that it can be used by `OpaqueTokenRequestPostProcessor`, `OpaqueTokenMutator`, and the various introspection-related tests.

Since `BearerTokenAuthentication` is generic, I'd recommend that its tests continue to use `DefaultOAuth2AuthenticatedPrincipal`.

qavid: No problem, I hope I got it right with "introspection-related tests"

jzheaux:
Thanks again, @qavid. In preparation for merging, will you please squash your commits and [format the commit message](https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.md#format-commit-messages)?

qavid:
I  have squashed commits and updated commit message.

Btw,
> [format the commit message](https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.md#format-commit-messages)

link is pointing to non-existing document. Same incorrect link is on this [wiki page](https://github.com/spring-projects/spring-security/wiki/Contributor-Guidelines)
Correct URL is: https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.adoc (note .adoc suffix instead of .md)

jzheaux:
Thanks, @qavid, I've updated my message as well as my reply template.

jzheaux:
Thanks, @qavid! This is now merged into `master` via af1c96b425e18d6ed4f722c40ad2488dcd234503. Note that I added a small polish in 221c33f5581547bff6791f08ddc8de4a4d497fca and updated the Opaque Token Sample in 19f06284cd6ef8a3bd6f38df79355befec61368e to document how the feature works.

PR #8742, opened 2020-06-22
Add support for dynamic JWS signature algorithm with JWKs - Issue 7160

hitchan:
CLA has been submitted.

This PR adds support for creating a JWSVerificationKeySelector dynamically based on the availability of JWK algorithms from a given JWK Set provided by an OIDC discovery document (Issue 7160). Non-standard or unknown algorithms will be silently ignored (see SignatureAlgorithm class). A warning log has also been added to let developers know if there was a problem fetching the JWK Set, although this does not halt the application from starting.

The original functionality is still in place that adds the RS256 algorithm if none are provided during initial configuration for backwards compatibility.

This functionality has been implemented for both standard and reactive versions.

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoderBuilder.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoderBuilder.java
@@ -17,7 +17,10 @@ package org.springframework.security.oauth2.jwt;
 
 import com.nimbusds.jose.Algorithm;
 import com.nimbusds.jose.JWSAlgorithm;
-import com.nimbusds.jose.jwk.*;
+import com.nimbusds.jose.jwk.JWK;
+import com.nimbusds.jose.jwk.JWKMatcher;
+import com.nimbusds.jose.jwk.JWKSelector;
+import com.nimbusds.jose.jwk.KeyUse;
 import com.nimbusds.jose.jwk.source.JWKSource;
 import com.nimbusds.jose.proc.SecurityContext;
 import org.apache.commons.logging.Log;
@@ -29,13 +32,14 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.*;
 import java.util.function.Consumer;
-import java.util.function.Function;
 import java.util.stream.Collectors;
 
 /**
  * An abstraction of the common functionality for the two main JwtDecoderBuilder instances
  * ({@link NimbusJwtDecoder}, and {@link NimbusReactiveJwtDecoder}).
- * @param <T>
+ * @param <T> The parent class type
+ *
+ * @author Nick Hitchan
  */
 public abstract class JwtDecoderBuilder<T> {
 
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoderBuilder.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoderBuilder.java
@@ -94,13 +94,16 @@ public abstract class JwtDecoderBuilder<T> {
 
 	/**
 	 * Fetches {@link SignatureAlgorithm}s based on the configured {@link JWKSource}s keys.
+	 * If algorithms have already been provided, then dynamic JWK algorithm fetching will not occur.
 	 * @param jwkSource
 	 * @return A set of {@link JWSAlgorithm}s to be used for JWT signature verification.
 	 */
 	protected Set<JWSAlgorithm> getSignatureAlgorithms(JWKSource<SecurityContext> jwkSource) {
 		Set<SignatureAlgorithm> jwkAlgorithms = getDefaultAlgorithms();
 		try {
-			jwkAlgorithms.addAll(fetchSignatureVerificationAlgorithms(jwkSource));
+			if (signatureAlgorithms.isEmpty()) {
+				jwkAlgorithms.addAll(fetchSignatureVerificationAlgorithms(jwkSource));
+			}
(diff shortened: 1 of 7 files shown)


hitchan:
Some of the integration tests are hanging, will update the PR once the issue has been resolved.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoderBuilder.java
```diff
@@ -0,0 +1,223 @@
...
+
+	/**
+	 * Fetches {@link SignatureAlgorithm}s based on the configured {@link JWKSource}s keys.
+	 * @param jwkSource
+	 * @return A set of {@link JWSAlgorithm}s to be used for JWT signature verification.
+	 */
+	protected Set<JWSAlgorithm> getSignatureAlgorithms(JWKSource<SecurityContext> jwkSource) {
+		Set<SignatureAlgorithm> jwkAlgorithms = getDefaultAlgorithms();
+		try {
+			jwkAlgorithms.addAll(fetchSignatureVerificationAlgorithms(jwkSource));
```
jzheaux, line 103: Spring Security builders are not typically additive - instead they replace. This allows Spring Security to gracefully backoff when an application wants to manually configure a value.

What that means here is that if the application has configured any algorithms, then the auto-fetch doesn't get run.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoderBuilder.java
```diff
@@ -0,0 +1,223 @@
...
+	/**
+	 * Converts a list of {@link JWK}s into a set of {@link SignatureAlgorithm}s.
+	 * @param jwks
+	 * @return A set of {@link SignatureAlgorithm} instances that may be used to validate a JWT (JWS).
+	 */
+	private Set<SignatureAlgorithm> fetchSignatureVerificationAlgorithms(List<JWK> jwks) {
+		if (jwks == null) {
+			return Collections.emptySet();
+		}
+		return jwks.stream().map(jwk -> {
```
jzheaux, line 172: The Stream API is a bit too slow and hard on the garbage collector to be used in Spring Security - will you please change this to a for loop?

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoderBuilder.java
```diff
@@ -0,0 +1,223 @@
...
+	private Set<SignatureAlgorithm> fetchSignatureVerificationAlgorithms(JWKSource<SecurityContext> jwkSource) {
+		return fetchSignatureVerificationAlgorithms(fetchSignatureVerificationJwks(jwkSource));
+	}
+
+	/**
+	 * Given a valid {@link ReactiveJWKSource}, fetches, and parses out the algorithms of available JWKs.
+	 * @param jwkSource
+	 * @return A set of {@link SignatureAlgorithm} instances that may be used to validate a JWT (JWS).
+	 */
+	private Set<SignatureAlgorithm> fetchSignatureVerificationAlgorithms(ReactiveJWKSource jwkSource) {
```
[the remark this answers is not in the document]
hitchan: Thanks for the feedback. I agree we definitely could, to be honest I don't have too much experience on the reactive side of things as of yet. This comment cleared up some doubts I had.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoderBuilder.java
hitchan: Understood, I agree that is definitely a better approach. I will make the required changes.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoderBuilder.java
hitchan: Yes that is no problem at all.

hitchan:
@jzheaux Should I be basing these changes off of master? I am not able to get a clean build & successful test to run off of the master branch.

jzheaux:
What I'm seeing on your branch is that the `JwtIssuerAuthenticationManagerResolver` tests are failing. This is likely because they are also using `MockWebServer`, but are only mocking the discovery endpoint at this point and not also the JWK Set endpoint.

If you added that mocking to those tests, I think you'd see those start to pass.

Also, I see that the multitenancy sample is failing, and I believe that's due to the `jwsAlgorithm` being moved from `NimbusJwtDecoderJwkSetUriBuilder` out to `JwtDecoderBuilder`. I think inlining the support you wrote back into `NimbusJwtDecoderJwkSetUriBuilder` will resolve that.

hitchan:
@jzheaux I decided to start a new branch here https://github.com/hitchan/spring-security/tree/feature/7160-2 and inline a lot of these changes and simplify them with your comments in mind.

Could you please take a look at this branch and provide any insight as to why these (i think) simple changes cause widespread unit test failure? Most of them are "ClassDefNotFound" problems. I am unable to determine the connection between some new methods in a class and the class not being able to be found by the classloader.

What I have tried so far
 - master branch -> build -> success.
 - master branch -> Add one unused method into NimbusJwtDecoder -> 100's of unit test failures. I'm really stuck here & unsure of what to do at this point.

hitchan:
I'm going to close this pull request in favor of the [more simplified implementation here](https://github.com/spring-projects/spring-security/pull/8752)

Issue #8730, opened 2020-06-19
Spring Boot 2.3.x refuses JWT tokens with custom `typ` header


jrehwaldt:
Spring Boot 2.3.x.RELEASE applies additional non-standard validation on JWT tokens due to upgrade to NimbusDS 8.x.

With Spring Boot 2.2.2.RELEASE a JWT token with header field `typ: "JWS"` validates fine. In Spring Boot 2.3.0.RELEASE such a token is rejected due to [a breaking change in NimbusDS 8.x](https://bitbucket.org/connect2id/nimbus-jose-jwt/commits/ee6aec33c6349ebc85de37f4b76da7f1feeb0ee5?w=1#chg-CHANGELOG.txt), which requires `typ` to be set to either `JWT` or omitted.

As per [RFC 7515 there's no dedicated `typ` mandated](https://tools.ietf.org/html/rfc7515#section-4.1.9).

The issue is that the builder design in `NimbusJwtDecoder` (in my case `NimbusJwtDecoder.withPublicKey(publicKey).build()`) does not allow to go back to the old behavior, nor does it allow to change the `JWSTypeVerifier`/`JWETypeVerifier` set in NimbusDS's `DefaultJWTProcessor`. Therefore, it is currently necessary to duplicate the complete builder code, which is making me uneasy considering it's a security relevant part and a future change in Spring Security/Nimbus might render my code insecure.

**Current Behavior**

JWT tokens with `typ: JWS` are refused since Spring Boot 2.3.x

**Expected Behavior**

Either one of:
* unusual, [but not forbidden types](https://tools.ietf.org/html/rfc7515#section-4.1.9) are allowed
* Builder allows to specify custom allowed types

**Context**

Dependency: `org.springframework.security:spring-security-oauth2-jose:5.3.3.RELEASE`
Class: `org.springframework.security.oauth2.jwt.NimbusJwtDecoder`

Unfortunately, I am not in control of the authorization server, which generates these custom JWTs. At the same time, I do not want to copy security relevant code.

Possible solutions:
1. restore previous behavior by setting `JWSTypeVerifier`/`JWETypeVerifier` to no-op verifiers in the builder(s)
2. allow to configure allowed types via config/builder
4. keep as is and require users to duplicate the builder code to apply either 1. or 2.

<details>
<summary>Stacktrace</summary>

```
{
   "cause":null,
   "stackTrace":[
      {
         "classLoaderName":"app",
         "moduleName":null,
[...]

PR #8745, opened 2020-06-22
Allow customizing `JWTProcessor` passed to `NimbusJwtDecoder`

jrehwaldt:
Add ability to post-process Nimbus's `JWTProcessor` before passing it to `NimbusJwtDecoder` when using the builders.

Fixes gh-8730

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
@@ -219,10 +219,12 @@ public final class NimbusJwtDecoder implements JwtDecoder {
 		private Set<SignatureAlgorithm> signatureAlgorithms = new HashSet<>();
 		private RestOperations restOperations = new RestTemplate();
 		private Cache cache;
+		private Consumer<ConfigurableJWTProcessor<SecurityContext>> jwtProcessorCustomizer;
 
 		private JwkSetUriJwtDecoderBuilder(String jwkSetUri) {
 			Assert.hasText(jwkSetUri, "jwkSetUri cannot be empty");
 			this.jwkSetUri = jwkSetUri;
+			this.jwtProcessorCustomizer = (processor) -> {};
 		}
 
 		/**
@@ -282,6 +284,20 @@ public final class NimbusJwtDecoder implements JwtDecoder {
 			return this;
 		}
 
+		/**
+		 * Use the given {@link Consumer} to customize the {@link JWTProcessor ConfigurableJWTProcessor} before
+		 * passing it to the build {@link NimbusJwtDecoder}.
+		 *
+		 * @param jwtProcessorCustomizer the callback used to alter the processor
+		 * @return a {@link JwkSetUriJwtDecoderBuilder} for further configurations
+		 * @since 5.4
+		 */
+		public JwkSetUriJwtDecoderBuilder jwtProcessorCustomizer(Consumer<ConfigurableJWTProcessor<SecurityContext>> jwtProcessorCustomizer) {
+			Assert.notNull(jwtProcessorCustomizer, "jwtProcessorCustomizer cannot be null");
+			this.jwtProcessorCustomizer = jwtProcessorCustomizer;
+			return this;
+		}
+
 		JWSKeySelector<SecurityContext> jwsKeySelector(JWKSource<SecurityContext> jwkSource) {
 			if (this.signatureAlgorithms.isEmpty()) {
 				return new JWSVerificationKeySelector<>(JWSAlgorithm.RS256, jwkSource);
@@ -311,6 +327,8 @@ public final class NimbusJwtDecoder implements JwtDecoder {
 
 			// Spring Security validates the claim set independent from Nimbus
 			jwtProcessor.setJWTClaimsSetVerifier((claims, context) -> { });
+
+			this.jwtProcessorCustomizer.accept(jwtProcessor);
 
 			return jwtProcessor;
 		}
@@ -414,11 +432,13 @@ public final class NimbusJwtDecoder implements JwtDecoder {
 	public static final class PublicKeyJwtDecoderBuilder {
 		private JWSAlgorithm jwsAlgorithm;
 		private RSAPublicKey key;
+		private Consumer<ConfigurableJWTProcessor<SecurityContext>> jwtProcessorCustomizer;
 
 		private PublicKeyJwtDecoderBuilder(RSAPublicKey key) {
 			Assert.notNull(key, "key cannot be null");
 			this.jwsAlgorithm = JWSAlgorithm.RS256;
 			this.key = key;
+			this.jwtProcessorCustomizer = (processor) -> {};
 		}
 
 		/**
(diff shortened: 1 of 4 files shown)


oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -461,7 +476,9 @@ public PublicKeyJwtDecoderBuilder signatureAlgorithm(SignatureAlgorithm signatur
 		 * @return the configured {@link NimbusJwtDecoder}
 		 */
 		public NimbusJwtDecoder build() {
-			return new NimbusJwtDecoder(processor());
+			ConfigurableJWTProcessor<SecurityContext> processor = processor();
```
jzheaux, line 479: It seems like the delta would be less if you had `processor()` use the `jwtProcessorCustomizer` instead. Any reason that it needs to be inside the `build()` method?

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -437,7 +439,20 @@ public PublicKeyJwtDecoderBuilder signatureAlgorithm(SignatureAlgorithm signatur
 			return this;
 		}
 
-		JWTProcessor<SecurityContext> processor() {
+		/**
+		 * Allows to customize the {@link JWTProcessor ConfigurableJWTProcessor} before passing it to
+		 * the build {@link NimbusJwtDecoder}.
+		 *
+		 * @param jwtProcessorCustomizer the callback used to alter the processor
+		 * @return a {@link PublicKeyJwtDecoderBuilder} for further configurations
```
jzheaux, line 447: Please add a `@since 5.4` to any public methods that you add in this PR.

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
```diff
@@ -346,6 +347,23 @@ public void decodeWhenSignatureMismatchesAlgorithmThenThrowsException() throws E
 				.isInstanceOf(BadJwtException.class);
 	}
 
+	// gh-8730
+	@Test
+	public void decodeWhenCustomizingJwtProcessor() throws Exception {
+		RSAPublicKey publicKey = TestKeys.DEFAULT_PUBLIC_KEY;
+		RSAPrivateKey privateKey = TestKeys.DEFAULT_PRIVATE_KEY;
+		JWSHeader header = new JWSHeader.Builder(JWSAlgorithm.RS256).type(new JOSEObjectType("custom")).build();
```
jzheaux, line 355: Is your use case that `typ` is set to `JWS`? If so, you might consider changing the test to that to prove that it'll work for your use case.

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
```diff
@@ -346,6 +347,23 @@ public void decodeWhenSignatureMismatchesAlgorithmThenThrowsException() throws E
 				.isInstanceOf(BadJwtException.class);
 	}
 
+	// gh-8730
+	@Test
+	public void decodeWhenCustomizingJwtProcessor() throws Exception {
```
jzheaux, line 352: While I agree that this test name makes sense, you might consider naming it to something that reflects the use case you are wanting the functionality to address, e.g. `decodeWhenCustomTypHeaderThenProcesses` or something like that.

jrehwaldt:
@jzheaux Thank you for the comments. I incorporated your advice and added the functionality to the other two builder.

I had to "swap" the logic in the test `withJwkSetUriWhenUsingCustomTypeHeaderThenRefuseOmittedType` for the following reason: `NimbusJwtDecoderTests.JWK_SET` does not match `TestKeys.DEFAULT_PRIVATE_KEY`, which makes it impossible to construct a signed JWT with custom header `typ`. If you have a better idea I am happy to improve/align this test with the other two.

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
```diff
@@ -524,6 +573,30 @@ public void decodeWhenCacheIsConfiguredAndValueLoaderErrorsThenThrowsJwtExceptio
 				.hasMessageContaining("An error occurred while attempting to decode the Jwt");
 	}
 
+	// gh-8730
+	@Test
+	public void withJwkSetUriWhenUsingCustomTypeHeaderThenRefuseOmittedType() throws Exception {
```
jrehwaldt, line 578: I had to "swap" the logic in the test withJwkSetUriWhenUsingCustomTypeHeaderThenRefuseOmittedType for the following reason: `NimbusJwtDecoderTests.JWK_SET` does not match `TestKeys.DEFAULT_PRIVATE_KEY`, which makes it impossible to construct a signed JWT with custom header typ. If you have a better idea I am happy to improve/align this test with the other two.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -282,6 +284,20 @@ public JwkSetUriJwtDecoderBuilder cache(Cache cache) {
 			return this;
 		}
 
+		/**
+		 * Allows to customize the {@link JWTProcessor ConfigurableJWTProcessor} before passing it to
```
jzheaux, line 288: Will you please reword this to match the JavaDoc for the other methods? Maybe something closer to:

```
Use the given {@link Consumer} to customize the {@link JWTProcessor ConfigurableJWTProcessor} before ...
```

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
jzheaux: I think this is fine. Note that the major goal of a test like this is just to make sure that your customization is used. So Mockito would be fine in this case, too; just call `verify` at the end of the test.

jrehwaldt: I opted against Mockito since it doesn't guarantee the `jwtProcessor` passed to the callback is actually the one used to verify (aka passed to `NimbusJwtDecorder`). Code like `jwtProcessorCustomizer.accept(new DefaultJWTProcessor())` would verify just fine with Mockito.

jrehwaldt:
@jzheaux Let me know if there's anything else you'd like me to improve.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -312,6 +328,8 @@ public JwkSetUriJwtDecoderBuilder cache(Cache cache) {
 			// Spring Security validates the claim set independent from Nimbus
 			jwtProcessor.setJWTClaimsSetVerifier((claims, context) -> { });
 
+			jwtProcessorCustomizer.accept(jwtProcessor);
```
jzheaux, line 331: For consistency with the rest of the code here, will you please prefix member variable references with `this.`?

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
```diff
@@ -346,6 +347,30 @@ public void decodeWhenSignatureMismatchesAlgorithmThenThrowsException() throws E
...
+		RSAPublicKey publicKey = TestKeys.DEFAULT_PUBLIC_KEY;
+		RSAPrivateKey privateKey = TestKeys.DEFAULT_PRIVATE_KEY;
+		JWSHeader header = new JWSHeader.Builder(JWSAlgorithm.RS256).type(new JOSEObjectType("JWS")).build();
+		JWTClaimsSet claimsSet = new JWTClaimsSet.Builder()
+				.expirationTime(Date.from(Instant.now().plusSeconds(60)))
+				.build();
+		SignedJWT signedJwt = signedJwt(privateKey, header, claimsSet);
+		NimbusJwtDecoder decoder = withPublicKey(publicKey)
+				.signatureAlgorithm(SignatureAlgorithm.RS256)
+				.jwtProcessorCustomizer(p -> p.setJWSTypeVerifier((type, context) -> {}))
```
jzheaux, line 362: I'd recommend that the test follow the same allowlisting pattern that Nimbus is using. For example, it'd be more secure to do:

```java
.jwtProcessorCustomizer(p -> p.setJWSTypeVerifier(
        new DefaultJOSEObjectTypeVerifier(new JOSEObjectType("JWS"))
    )
)
```

jrehwaldt:
@jzheaux I applied the last comments and squashed the commits.

jzheaux:
Great, @jrehwaldt, thanks for your contribution! This is now merged into `master` via d31fff11b3bb91e484228f29af469a9a0661dd32.

Note that I did a small polish on the commit so that the title was less than 50 characters and the final message was `Closes gh-xxxx` instead of `Fixes gh-xxxx`.

I hope to see more contributions from you in the future!

PR #8747, opened 2020-06-22
optimize-methods

lenve:
The method of SecurityContextHolderStrategy#createEmptyContext  is the  same in three implement class (GlobalSecurityContextHolderStrategy,ThreadLocalSecurityContextHolderStrategy and InheritableThreadLocalSecurityContextHolderStrategy).

If we use the method of  createEmptyContext, a feature of Java8, we can define a default implement in SecurityContextHolderStrategy interface.Then we do not need the same code in these three implement classes any more.


rwinch:
Thank for the PR. We don't want to provide a default implementation as we want any implementer to explicitly choose what implementation of SecurityContext they would use.

PR #8756, opened 2020-06-24
Use memory-saving Collections.singletonList in JdbcAclService.readAclById()

stsypanov:
We don't need array-based `Arrays.asList()` as we always have only one instance `ObjectIdentity` at call site.


stsypanov:
@eleftherias done

eleftherias:
Thanks for the PR @stsypanov! This is now merged into master.

Issue #7890, opened 2020-02-03
Provide reactive JDBC implementation of ReactiveOAuth2AuthorizedClientService


jgrandja:
This issue addresses the reactive counterpart of #7655.

Related #7889

drgeb:
Is there any example of the JDBCReactiveOAuth2AuthorizedClientService that I can start to look at ?

jgrandja:
@drgeb The issue is meant to deliver `JdbcReactiveOAuth2AuthorizedClientService`. The work here hasn't started so there are no examples.

ovidiupopa91:
Hi @jgrandja  if it's ok I can take a look and provide the ```JdbcReactiveOAuth2AuthorizedClientService```

jgrandja:
@ovidiupopa91 That would be great! The issue is yours.

rwinch:
Another thing to keep in mind is that Spring Framework 5.3 is adding support for r2dbc https://github.com/spring-projects/spring-framework/pull/25065 It might be valuable to build off of that and wait until we are on Spring Framework 5.3

jgrandja:
Thanks for the heads up @rwinch. I totally agree that we should wait until Spring Framework 5.3 is out and leverage the provided implementations.

@ovidiupopa91 Let's hold off on this for now. I would rather write less code than more so let's wait until Spring Framework 5.3 is out and then we can proceed.

rwinch:
We could start something in a branch that is based of the current Spring Framework snapshots and then it would be ready when Framework released.

jgrandja:
Makes sense @rwinch.

@ovidiupopa91 Please take a look at the new features added in [spring-projects/spring-framework#25065](https://github.com/spring-projects/spring-framework/pull/25065) and leverage for this implementation. Let me know if you have any questions.

ovidiupopa91:
Hi @jgrandja . Great, I'll do that. I will probably open a draft PR to get your feedback as soon as possible.

ovidiupopa91:
Hi @jgrandja . Can you please reassign me to the issue? (just to avoid what happened when I tried to contribute to the spring-authorization-server project :D )

PR #8765, opened 2020-06-26
Provide a R2dbc implementation of ReactiveOuath2AuthorizedClientService

ovidiupopa91:
Implement R2dbcReactiveOuath2AuthorizedClientService which persists the
Oauth2AuthorizedClient in a sql database

Update gradle properties to use the 5.3.0-SNAPSHOT version of spring
framework (take advantage of the r2dbc integration)

Add optional depedency to the spring-r2dbc module

Add test compile dependencies to r2dbc-h2 and r2dbc-test

Closes gh-7890


jgrandja:
@ovidiupopa91 Thanks for the PR ! Overall this looks good. Spring Framework 5.3.0 GA is planned for Oct 27 so I'll review this a couple weeks beforehand and we'll merge shortly after Oct 27. Just wanted to give you an update.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/R2dbcReactiveOAuth2AuthorizedClientService.java
```diff
@@ -0,0 +1,323 @@
...
+				.findByRegistrationId(oAuth2AuthorizedClientHolder.getClientRegistrationId())
+				.switchIfEmpty(Mono
+						.error(dataRetrievalFailureException(oAuth2AuthorizedClientHolder.getClientRegistrationId())))
+				.map(clientRegistration -> new OAuth2AuthorizedClient(clientRegistration,
+						oAuth2AuthorizedClientHolder.getPrincipalName(),
+						oAuth2AuthorizedClientHolder.getoAuth2AccessToken(),
+						oAuth2AuthorizedClientHolder.getRefreshToken()));
+	}
+
+	private Mono<Integer> updateAuthorizedClient(OAuth2AuthorizedClient authorizedClient, Authentication principal) {
```
jgrandja, line 214: Let's move `updateAuthorizedClient` and `insertAuthorizedClient` right under `saveAuthorizedClient`

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/R2dbcReactiveOAuth2AuthorizedClientService.java
```diff
@@ -0,0 +1,323 @@
...
+		}
+		return executeSpec.fetch().rowsUpdated();
+	}
+
+	private Throwable dataRetrievalFailureException(String clientRegistrationId) {
+		return new DataRetrievalFailureException("The ClientRegistration with id '" + clientRegistrationId
+				+ "' exists in the data source, " + "however, it was not found in the ClientRegistrationRepository.");
+	}
+
+	private static final class OAuth2AuthorizedClientParametersMapper
```
jgrandja, line 237: Change to `public static class`

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/R2dbcReactiveOAuth2AuthorizedClientService.java
```diff
@@ -0,0 +1,323 @@
...
+			}
+
+			parameters.put("refreshTokenValue", Parameter.fromOrEmpty(refreshTokenValue, ByteBuffer.class));
+			parameters.put("refreshTokenIssuedAt", Parameter.fromOrEmpty(refreshTokenIssuedAt, LocalDateTime.class));
+			return parameters;
+		}
+
+	}
+
+	private static final class OAuth2AuthorizedClientRowMapper
```
jgrandja, line 281: Change to `public static class`

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/R2dbcReactiveOAuth2AuthorizedClientService.java
```diff
@@ -0,0 +1,323 @@
...
+	 * The default is {@link OAuth2AuthorizedClientRowMapper}.
+	 *
+	 * @param authorizedClientRowMapper the {@link BiFunction} used for mapping the current row in {@code io.r2dbc.spi.Row} to {@link OAuth2AuthorizedClientHolder}
+	 */
+	public void setAuthorizedClientRowMapper(BiFunction<Row, RowMetadata, OAuth2AuthorizedClientHolder> authorizedClientRowMapper) {
+		Assert.notNull(authorizedClientRowMapper, "authorizedClientRowMapper cannot be null");
+		this.authorizedClientRowMapper = authorizedClientRowMapper;
+	}
+
+	public static class OAuth2AuthorizedClientHolder {
```
jgrandja, line 170: Add `final` modifier

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/R2dbcReactiveOAuth2AuthorizedClientService.java
```diff
@@ -0,0 +1,323 @@
...
+		@Override
+		public Map<String, Parameter> apply(OAuth2AuthorizedClient authorizedClient, Authentication principal) {
+
+			final Map<String, Parameter> parameters = new HashMap<>();
+
+			final ClientRegistration clientRegistration = authorizedClient.getClientRegistration();
+			final OAuth2AccessToken accessToken = authorizedClient.getAccessToken();
+			final OAuth2RefreshToken refreshToken = authorizedClient.getRefreshToken();
+
+			parameters.put("clientRegistrationId", Parameter.from(clientRegistration.getRegistrationId()));
```
jgrandja, line 249: Let's use `fromOrEmpty()` or `Parameter(@Nullable Object value, Class<?> type)` so that the type is also set.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/R2dbcReactiveOAuth2AuthorizedClientService.java
```diff
@@ -0,0 +1,323 @@
...
+			parameters.put("principalName", Parameter.from(principal.getName()));
+			parameters.put("accessTokenType", Parameter.from(accessToken.getTokenType().getValue()));
+			parameters.put("accessTokenValue",
+					Parameter.from(ByteBuffer.wrap(accessToken.getTokenValue().getBytes(StandardCharsets.UTF_8))));
+			parameters.put("accessTokenIssuedAt",
+					Parameter.from(LocalDateTime.ofInstant(accessToken.getIssuedAt(), ZoneOffset.UTC)));
```
jgrandja, line 255: I wonder if we should use `Timestamp.from()` same as in `JdbcOAuth2AuthorizedClientService`?

jgrandja (changes requested):
Thanks for the PR @ovidiupopa91. Please see comments inline.

Also, please rebase on master.

Then run the build and you'll notice code formatting errors, which you can auto-correct by following the console prompt.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/R2dbcReactiveOAuth2AuthorizedClientService.java
```diff
@@ -0,0 +1,323 @@
...
+			" access_token_expires_at = :accessTokenExpiresAt, " +
+			" access_token_scopes = :accessTokenScopes," +
+			" refresh_token_value = :refreshTokenValue, " +
+			" refresh_token_issued_at = :refreshTokenIssuedAt" +
+			" WHERE " +
+			PK_FILTER;
+	private DatabaseClient databaseClient;
+	private ReactiveClientRegistrationRepository clientRegistrationRepository;
+	private BiFunction<OAuth2AuthorizedClient, Authentication, Map<String, Parameter>> authorizedClientParametersMapper;
+	private BiFunction<Row, RowMetadata, OAuth2AuthorizedClientHolder>  authorizedClientRowMapper;
```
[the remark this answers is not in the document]
ovidiupopa91: Hi @jgrandja ,

if we add ```clientRegistrationRepository ``` as a member of the ```OAuth2AuthorizedClientRowMapper``` it means that the return type of this function would be ```Mono<OAuth2AuthorizedClient>```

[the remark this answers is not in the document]
jgrandja: Yes, I see what you're saying. Give it a try and if it works out great otherwise we can keep as-is.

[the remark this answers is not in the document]
ovidiupopa91: Changed the method signature to ``` public Mono<OAuth2AuthorizedClient> apply(Row row, RowMetadata rowMetadata)``` in ```OAuth2AuthorizedClientRowMapper``` class. Tests are passing. 

The only thing that I don't like how it looks is 
```
return (Mono<T>) this.databaseClient.sql(LOAD_AUTHORIZED_CLIENT_SQL)
				.bind("clientRegistrationId", clientRegistrationId).bind("principalName", principalName)
				.map(this.authorizedClientRowMapper::apply)
				.first().flatMap(authorizedClient -> authorizedClient);
```
Because ```first()``` is already returning a Mono, basically the result of he mapper will actually be ```Mono<Mono<OAuth2AuthorizedClient>>``` which means that I have to do a``` flatMap``` to get the underlying ```Mono<OAuth2AuthorizedClient>```

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/R2dbcReactiveOAuth2AuthorizedClientService.java
ovidiupopa91: Timestamp is not a valid datatype. Based on https://r2dbc.io/spec/0.8.3.RELEASE/spec/html/, sql type TIMESTAMP should be mapped as ```java.time.LocalDateTime``` if it represents a date and time without a timezone.

gradle.properties
```diff
@@ -6,3 +6,4 @@ kotlinVersion=1.3.70
 org.gradle.jvmargs=-Xmx3g -XX:MaxPermSize=2048m -XX:+HeapDumpOnOutOfMemoryError
 org.gradle.parallel=true
 org.gradle.caching=true
+springVersion=5.3.0-SNAPSHOT
```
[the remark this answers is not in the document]
ovidiupopa91: I will also have to update ```ext.reactorVersion = "Dysprosium-SR+"``` because spring-tx is using ```ext.reactorVersion = "2020.0.0-SNAPSHOT"```. If I don't do that, test are failing with ```NoSuchMethodError``` in ```org.springframework.transaction.reactive.TransactionContextManager.currentContext()```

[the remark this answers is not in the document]
ovidiupopa91: I will also have to update ```ext.reactorVersion = "Dysprosium-SR+"``` because spring-tx is using ```ext.reactorVersion = "2020.0.0-SNAPSHOT"```. If I don't do that, test are failing with ```NoSuchMethodError``` in ```org.springframework.transaction.reactive.TransactionContextManager.currentContext()``` which is using ```Mono.deferContextual``` which is not available with Dysprosium-SR+ release train

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/R2dbcReactiveOAuth2AuthorizedClientService.java
ovidiupopa91: Fixed

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/R2dbcReactiveOAuth2AuthorizedClientService.java
ovidiupopa91: Fixed

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/R2dbcReactiveOAuth2AuthorizedClientService.java
ovidiupopa91: Fixed

ovidiupopa91:
Hi @jgrandja, 

fixed all comments, format and checkstyle issues. Not sure why the build is failing when running following tasks:

1. ```Task :spring-security-web:jacocoTestReport```
2. ```Task:spring-security-config:integrationTest```

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/R2dbcReactiveOAuth2AuthorizedClientService.java
```diff
@@ -0,0 +1,370 @@
...
+	@Override
+	@SuppressWarnings("unchecked")
+	public <T extends OAuth2AuthorizedClient> Mono<T> loadAuthorizedClient(String clientRegistrationId,
+			String principalName) {
+		Assert.hasText(clientRegistrationId, "clientRegistrationId cannot be empty");
+		Assert.hasText(principalName, "principalName cannot be empty");
+
+		return (Mono<T>) this.databaseClient.sql(LOAD_AUTHORIZED_CLIENT_SQL)
+				.bind("clientRegistrationId", clientRegistrationId).bind("principalName", principalName)
+				.map(this.authorizedClientRowMapper::apply).first().flatMap((authorizedClient) -> authorizedClient);
```
jgrandja, line 142: Agreed, I'm also not in favour of the `flatMap()` that is required because of the `Mono<OAuth2AuthorizedClient>` return type in the row mapper.

Let's revert the return type to `OAuth2AuthorizedClientHolder` having these members:

```java
	public static final class OAuth2AuthorizedClientHolder {
		private final String clientRegistrationId;
		private final String principalName;
		private final OAuth2AccessToken accessToken;
		private final OAuth2RefreshToken refreshToken;


```

jgrandja (changes requested):
Thanks for the updates @ovidiupopa91. I left a couple of minor comments.

Also, please squash the commits after you apply the last updates and we should be ready to merge.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/R2dbcReactiveOAuth2AuthorizedClientService.java
ovidiupopa91: This means that the ```authorizedClientParametersMapper``` will also be reverted to 
`
private BiFunction<OAuth2AuthorizedClient, Authentication, Map<String, Parameter>> authorizedClientParametersMapper;
`
or do you prefer using the same signature

ovidiupopa91:
```
protected Function<OAuth2AuthorizedClientHolder, Map<String, Parameter>> authorizedClientParametersMapper;

protected BiFunction<Row, RowMetadata, OAuth2AuthorizedClientHolder> authorizedClientRowMapper;
```

and two constructors in ```OAuth2AuthorizedClientHolder```
```
public OAuth2AuthorizedClientHolder(OAuth2AuthorizedClient authorizedClient, Authentication principal) {
			Assert.notNull(authorizedClient, "authorizedClient cannot be null");
			Assert.notNull(principal, "principal cannot be null");
			this.clientRegistrationId = authorizedClient.getClientRegistration().getRegistrationId();
			this.principalName = principal.getName();
			this.accessToken = authorizedClient.getAccessToken();
			this.refreshToken = authorizedClient.getRefreshToken();
		}

		public OAuth2AuthorizedClientHolder(String clientRegistrationId, String principalName,
				OAuth2AccessToken accessToken, OAuth2RefreshToken refreshToken) {
			this.clientRegistrationId = clientRegistrationId;
			this.principalName = principalName;
			this.accessToken = accessToken;
			this.refreshToken = refreshToken;
		}
```
The first one is used when saving an authorized client. The second one is used when loading an authorized client.

oauth2/oauth2-client/src/main/resources/org/springframework/security/oauth2/client/oauth2-client-schema-h2.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE oauth2_authorized_client (
```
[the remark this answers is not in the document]
ovidiupopa91: Remove oauth2-client-schema-h2.sql

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/R2dbcReactiveOAuth2AuthorizedClientService.java
jgrandja: Let's use `OAuth2AuthorizedClientHolder` in both `authorizedClientParametersMapper` and `authorizedClientRowMapper`. Using a wrapper object in the signature provides us the flexibility in "adding" members to the wrapper object (if later required) without breaking the `BiFunction` signature.

The two constructors in `OAuth2AuthorizedClientHolder` makes sense. I would delegate the `OAuth2AuthorizedClient` constructor to the other constructor.

jgrandja: I just noticed that there are a lot more changed files in latest commit. Can you ensure that you **rebase** on top of master and not merge.

ovidiupopa91: @jgrandja I did a rebase. But, because there was a breaking API change in spring-framework, test module, I had to update Kotlin tests from spring-security-config (https://github.com/spring-projects/spring-framework/commit/d04c5f8b2cb274b8f44c47b3783e5d29f3e21b43#diff-5a70a1d6caaa66c8feec94bbaf723ee682cdafda132c988ef46a365cd6886f62)

ovidiupopa91: > The two constructors in `OAuth2AuthorizedClientHolder` makes sense. I would delegate the `OAuth2AuthorizedClient` constructor to the other constructor.

By doing this, following checks
```
Assert.notNull(authorizedClient, "authorizedClient cannot be null");
Assert.notNull(principal, "principal cannot be null");
```
will no longer be in place since the delegate constructor call must be the first statement in a constructor

jgrandja: Looks like it has been fixed, see https://github.com/spring-projects/spring-security/commit/9cf3129d7afa2abb439aba6aadfee0a2c8c784bf

jgrandja: I see. Ok then don't bother with delegating constructor as the `Assert` is necessary

ovidiupopa91: > Looks like it has been fixed, see [9cf3129](https://github.com/spring-projects/spring-security/commit/9cf3129d7afa2abb439aba6aadfee0a2c8c784bf)

yep, right. Looks like I rebased before it was fixed in the main repository :) I will revert the changes

jgrandja:
Thanks for the great work @ovidiupopa91 ! This is now in master.

Issue #8778, opened 2020-06-30
Mongolian translation required for messages.propeperties


hurelhuyag:
**Expected Behavior**.

I wanted to show spring security messages in mongolian

**Current Behavior**

There is no file named messages_mn_MN.properties

PR #8780, opened 2020-06-30
mongolian translation for messages.properties

hurelhuyag:
#8778


eleftherias:
Thanks for the PR @hurelhuyag!

Issue #8772, opened 2020-06-29
OpenSAML should get initialized as part of container lifecycle


jzheaux:
OpenSAML gets initialized by Spring Security lazily via a package-private class `OpenSamlImplementation`. This setup makes it tricky for an application to override the defaults that Spring Security specifies.

For example, to use a different `ParserPool` configuration, one needs to ensure that they call `XMLObjectProviderRegistrySupport.setParserPool` only after `OpenSamlAuthenticationProvider` or `OpenSamlAuthenticationRequestFactory` is classloaded since those are the two classes that use an instance of `OpenSamlImplementation`.

It may be cleaner to initialize OpenSAML as part of the container lifecycle, for example as a `BeanFactoryPostProcessor`. In this way, an application can register their own `BeanFactoryPostProcessor` to override the behavior.

At that point, the code can be removed in `OpenSamlImplementation` that initializes OpenSAML.

PR #8781, opened 2020-07-01
Initialize OpenSAML as part of container lifecycle

jzheaux:
Closes gh-8772

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configuration/Saml2ServiceProviderConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configuration/Saml2ServiceProviderConfiguration.java
@@ -1,0 +1,152 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.config.annotation.web.configuration;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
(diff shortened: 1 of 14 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/configuration/Saml2ServiceProviderConfiguration.java
```diff
@@ -0,0 +1,152 @@
...
+		BeanFactoryPostProcessor saml2BeanFactoryPostProcessor() {
+			return new OpenSamlServiceProviderBeanFactoryPostProcessor();
+		}
+
+		private static class OpenSamlServiceProviderBeanFactoryPostProcessor
+				implements BeanFactoryPostProcessor, Ordered {
+
+			@Override
+			public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory) throws BeansException {
+				if (ConfigurationService.get(XMLObjectProviderRegistry.class) == null) {
```
jzheaux, line 93: @rwinch What are your thoughts about initializing OpenSAML in `spring-security-config`? What I'm trying to achieve is remove it from the static initialization of `OpenSamlImplementation`, and this seems to be the easiest way to hide the implementation while still giving the application a hook to override what's done here.

As a side note, I think there's an argument to be made for Spring Security not initializing OpenSAML at all; however, 5.2. and 5.3 both initialize OpenSAML, so it would be a surprise to upgrade to 5.4 and find that no longer happening.

jzheaux: Also, of course, feel free to comment on other parts of the PR, but the above is my main question. If you agree that this is a viable route, then it's the one I'm planning on taking.

rwinch: I really don't like pushing this into configuration. It makes it harder to use without Spring configuration and also complicates things a bit (now users need to understand BeanFactoryPostProcessor to understand the initialization). Furthermore, I don't like the initialization tied to configuration. If someone creates a new `OpenSamlAuthenticationProvider` it should work without being tied to the Bean container.

Instead, I'd prefer exposing a single static method that initializes Open SAML that any class leveraging Open SAML could invoke. 

I think ideally we'd remove the static initialization all together. Perhaps SAML has a way to provide the `ParserPool` as an argument vs using a default static value?

jzheaux:
Closed in favor of 2276fcf34ab476578ff8653a98f986435f1e135f

PR #8790, opened 2020-07-02
Don't use oidc scopes_supported for scope as default in ClientRegistrations

### LINKED ISSUE #8514: scopes_supported metadata should not be used as default in ClientRegistrations
**Describe the bug**
The oauth2-client requests all available scopes on the authorization server by default. That includes also scopes for other clients.

`ClientRegistration.getScopes()` copies all `scopes_supported` from authorization server metadata into scope config for this client. So the oauth2-client requested on each authorization all of these scopes.

As [Keycloak has recently fixed a bug](https://issues.redhat.com/browse/KEYCLOAK-8071
) now throwing error on invalid scope requests, this bug will occur for more people.

**To Reproduce**

* Install current keycloak >= 10.0.0
* Create new client
* Remove some default keycloak or a new created client scopes from `Assigned Default Client Scopes` and `Assigned Optional Client Scopes` of this client
* Try to start a session with this client and oauth2-client.
* Don't configure `spring.security.oauth2.client.registration.${name}.scope` (This is a workaround to fix the problem)

**Expected behavior**

As `scopes_supported` lists all scopes for all clients on this authorization server it's not a good default.

>   scopes_supported
>      RECOMMENDED.  JSON array containing a list of the OAuth 2.0
>      [RFC6749] "scope" values that this authorization server supports.
>      Servers MAY choose not to advertise some supported scope values
>      even when this parameter is used.
Source: https://tools.ietf.org/html/rfc8414#section-2

Also, considering that the `scope` parameter is optional.

>    scope
>          OPTIONAL.  The scope of the access request as described by
>          Section 3.3.
Source https://tools.ietf.org/html/rfc6749#section-4.1.1

The best default should be just leave it empty.

martin-v:
fixes #8514

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
@@ -34,7 +34,6 @@ import org.springframework.http.RequestEntity;
 import org.springframework.security.oauth2.core.AuthorizationGrantType;
 import org.springframework.security.oauth2.core.ClientAuthenticationMethod;
 import org.springframework.security.oauth2.core.oidc.IdTokenClaimNames;
-import org.springframework.security.oauth2.core.oidc.OidcScopes;
 import org.springframework.util.Assert;
 import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.RestTemplate;
@@ -239,7 +238,6 @@ public final class ClientRegistrations {
 
 		return ClientRegistration.withRegistrationId(name)
 				.userNameAttributeName(IdTokenClaimNames.SUB)
-				.scope(OidcScopes.OPENID) // default to "openid" which must be supported
 				.authorizationGrantType(AuthorizationGrantType.AUTHORIZATION_CODE)
 				.clientAuthenticationMethod(method)
 				.redirectUri("{baseUrl}/{action}/oauth2/code/{registrationId}")
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
@@ -25,7 +25,6 @@ import java.util.function.Supplier;
 
 import com.nimbusds.oauth2.sdk.GrantType;
 import com.nimbusds.oauth2.sdk.ParseException;
-import com.nimbusds.oauth2.sdk.Scope;
 import com.nimbusds.oauth2.sdk.as.AuthorizationServerMetadata;
 import com.nimbusds.openid.connect.sdk.op.OIDCProviderMetadata;
 import net.minidev.json.JSONObject;
@@ -236,12 +235,11 @@ public final class ClientRegistrations {
 			throw new IllegalArgumentException("Only AuthorizationGrantType.AUTHORIZATION_CODE is supported. The issuer \"" + issuer +
 					"\" returned a configuration of " + grantTypes);
 		}
-		List<String> scopes = getScopes(metadata);
 		Map<String, Object> configurationMetadata = new LinkedHashMap<>(metadata.toJSONObject());
 
 		return ClientRegistration.withRegistrationId(name)
 				.userNameAttributeName(IdTokenClaimNames.SUB)
-				.scope(scopes)
+				.scope(OidcScopes.OPENID) // default to "openid" which must be supported
 				.authorizationGrantType(AuthorizationGrantType.AUTHORIZATION_CODE)
 				.clientAuthenticationMethod(method)
 				.redirectUri("{baseUrl}/{action}/oauth2/code/{registrationId}")
@@ -266,16 +264,6 @@ public final class ClientRegistrations {
 		}
 		throw new IllegalArgumentException("Only ClientAuthenticationMethod.BASIC, ClientAuthenticationMethod.POST and "
 				+ "ClientAuthenticationMethod.NONE are supported. The issuer \"" + issuer + "\" returned a configuration of " + metadataAuthMethods);
-	}
-
-	private static List<String> getScopes(AuthorizationServerMetadata metadata) {
-		Scope scope = metadata.getScopes();
-		if (scope == null) {
-			// If null, default to "openid" which must be supported
-			return Collections.singletonList(OidcScopes.OPENID);
-		} else {
-			return scope.toStringList();
-		}
 	}
 
 	private ClientRegistrations() {}
--- a/config/src/test/java/org/springframework/security/config/oauth2/client/ClientRegistrationsBeanDefinitionParserTests.java
+++ b/config/src/test/java/org/springframework/security/config/oauth2/client/ClientRegistrationsBeanDefinitionParserTests.java
@@ -152,7 +152,7 @@ public class ClientRegistrationsBeanDefinitionParserTests {
 		assertThat(googleRegistration.getClientAuthenticationMethod()).isEqualTo(ClientAuthenticationMethod.BASIC);
 		assertThat(googleRegistration.getAuthorizationGrantType()).isEqualTo(AuthorizationGrantType.AUTHORIZATION_CODE);
 		assertThat(googleRegistration.getRedirectUri()).isEqualTo("{baseUrl}/{action}/oauth2/code/{registrationId}");
-		assertThat(googleRegistration.getScopes()).isEqualTo(StringUtils.commaDelimitedListToSet("openid,profile,email"));
+		assertThat(googleRegistration.getScopes()).isEqualTo(StringUtils.commaDelimitedListToSet("openid"));
 		assertThat(googleRegistration.getClientName()).isEqualTo(serverUrl);
 
 		ProviderDetails googleProviderDetails = googleRegistration.getProviderDetails();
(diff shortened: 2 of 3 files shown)


pivotal-issuemaster:
@martin-v Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=8790)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=8790) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@martin-v Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=8790)!

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
```diff
@@ -268,16 +266,6 @@ private static ClientAuthenticationMethod getClientAuthenticationMethod(String i
 				+ "ClientAuthenticationMethod.NONE are supported. The issuer \"" + issuer + "\" returned a configuration of " + metadataAuthMethods);
 	}
 
-	private static List<String> getScopes(AuthorizationServerMetadata metadata) {
```
jgrandja, line 271: Instead of removing this method, let's apply this logic:

if `metadata.getScopes()` contains `openid` then
    default to `openid`, `profile`, `email`
else
  default to `openid`

This will ensure backwards compatibility when `CommonOAuth2Provider.GOOGLE` and `CommonOAuth2Provider.OKTA` is used.

martin-v: But for `CommonOAuth2Provider` the scopes are defined, and this is also tested in `CommonOAuth2ProviderTests` so these stick with `openid`, `profile`, `email`. You are probaply mislead by the bad test-data naming in `ClientRegistrationsBeanDefinitionParserTests`

martin-v: @jgrandja Did you get a chance to look at that remark?

jgrandja: @martin-v Apologies, my last comment was not that clear.

Let's assume the user has configured this:

```yml
spring:
  security:
    oauth2:
      client:
        registration:
          google:
            client-id: id
            client-secret: secret
        provider:
          google:
            issuer-uri: https://accounts.google.com
```

Google [discovery endpoint](https://accounts.google.com/.well-known/openid-configuration) returns `openid, profile, email` in the `scopes_supported` parameter, which initializes the `ClientRegistration` with the 3 scopes.

The changes in this PR would initialize the `ClientRegistration` with just the `openid` scope, which changes the runtime behaviour as the **UserInfo endpoint would not be called** by `OidcUserService` as it doesn't contain at least one of the [accessible scopes](https://github.com/spring-projects/spring-security/blob/833bfd0c22a4f026b10d9878e410dbadc89d2b3e/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserService.java#L170).

The logic proposed in previous comment would preserve this behaviour. Furthermore, if the `openid` scope is returned in `scopes_supported` then we know the provider supports OpenID Connect and the UserInfo endpoint so it makes sense to also default to `profile, email` to enable the UserInfo endpoint call.

Makes sense?

martin-v: `profile`, `email` are optional scopes in openid-connect, see https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims 

One option is if the `profile`, `email` is listed in `scopes_supported` then we add these, but this still can cause the problem from #8514 (If it is listed in 'scopes_supported', this does not mean that it is supported for that particular client)

martin-v: I didn't check but probably google diliveres always all 3 scopes, because it's allowed to deliver more scopes than requested, this whould also solve this issue.

Another question is how common is to use these lines
```
        provider:
          google:
            issuer-uri: https://accounts.google.com
```
Because as I understand it, it is optional and 'CommonOAuth2Provider' takes care of the configuration of the provider.

jgrandja: > One option is if the `profile`, `email` is listed in `scopes_supported` then we add these

Agreed, this logic makes the most sense.

> If it is listed in 'scopes_supported', this does not mean that it is supported for that particular client

I'm not sure I understand? FYI, if `profile` and `email` is listed in `scopes_supported` then **it can be requested by any client**.

jgrandja: > Another question is how common is to use these lines...

Whether it's common or not, the application can use this configuration so we need to account for this scenario.

martin-v: > > If it is listed in 'scopes_supported', this does not mean that it is supported for that particular client
> 
> I'm not sure I understand? FYI, if `profile` and `email` is listed in `scopes_supported` then **it can be requested by any client**.

No, thats exactly the problem in #8514. `scopes_supported` lists all scopes for all clients supported by the server, not for a specific client. At least this is what Keycloak is doing. If client _A_ supports scope _z_, _y_ and client _B_ supports scope _y_, _x_ then Keycloak lists _z_,_y_,_x_

jgrandja: @martin-v Regarding my comment...

> FYI, if `profile` and `email` is listed in `scopes_supported` then it can be requested by any client.

To be clear, the [standard OIDC scopes](https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims) `profile email address phone` may be requested (implicitly) by **ANY client** registered with an OpenID Connect provider. The client **does not need** to be explicitly configured to request these standard OIDC scopes. As far as custom scopes go, the client has to be explicitly configured to request custom (non-standard) scopes.

The solution I have proposed in this [comment](https://github.com/spring-projects/spring-security/pull/8790#discussion_r449603024) is dealing with the standard OIDC scopes only. Your previous comment is referring to non-standard scopes so this is not related to what we need to address in this issue.

It sounds like you don't agree with the proposed solution? If true, then please explain why. Have you tried out the proposed solution?

martin-v: One change I would made at least to the proposed solution is, not depend on `openid` to request other scopes, instead request `profile` if `profile` exist in `scopes_supported`. Same for email, etc..

But my general problem with this solution is that I didn't find in the OIDC spec that `profile email address phone` MUST be supported on the server side. Can you quote the relevant part of the spec?

If the spec don't require `profile email address phone` we should not expect that they are allowed for each client on the server, even if `scopes_supported` claims that one of the clients support these scopes. Because we still do not know which client supports which scope.

I see the problem with backwards compatibility, so probably add something like the proposed solution but add a deprecation warning that this behaviour will be removed in the next minor/major version. What do you think about that?

jgrandja: @martin-v I was mistaken with my earlier [comment](https://github.com/spring-projects/spring-security/pull/8790#discussion_r449603024) regarding:

> This will ensure backwards compatibility when `CommonOAuth2Provider.GOOGLE` and `CommonOAuth2Provider.OKTA` is used.

This is a non-issue if the application uses the default settings. However, **it may** be an issue if the application customizes the properties as mentioned in this [comment](https://github.com/spring-projects/spring-security/pull/8790#discussion_r452221523). However, even with this custom configuration the application will need to configure the `scope` property as well to ensure the correct scopes are requested.

The solution I originally proposed doesn't actually make sense. We cannot assume that all `ClientRegistation`'s are configured for `oauth2Login()`, therefore we should not default to `openid` or any of the additional OIDC scopes. There may be `ClientRegistation`'s that are configured for OAuth 2.0 `authorization_code` or `client_credentials` and defaulting to `openid` would be invalid.

I think the change is pretty simple now - we do not assign any defaults to `ClientRegistration.scope()`.

jgrandja: @martin-v Did you see my last comment? I'm looking to get this merged this week. Are you able to apply the change I mentioned. If not, I can apply and merge.

martin-v: @jgrandja sry, didn't see it.

If we leave it empty we still have the problem mentioned in https://github.com/spring-projects/spring-security/issues/8514#issuecomment-652438261 but we can't fix both issues.

martin-v: And I can do the change, if you still prefer the empty solution.

jgrandja: @martin-v Regarding...

> If we leave it empty we still have the problem mentioned in #8514 

I realize that `openid` is a required scope for OIDC. However, as I mentioned in my previous [comment](https://github.com/spring-projects/spring-security/pull/8790#discussion_r458179423)

> We cannot assume that all `ClientRegistation`'s are configured for `oauth2Login()`, therefore we should not default to `openid` or any of the additional OIDC scopes. There may be `ClientRegistation`'s that are configured for OAuth 2.0 `authorization_code` or `client_credentials` and defaulting to `openid` would be invalid.

Please go ahead with the change on NOT assigning any defaults to `ClientRegistration.scope()`. This change will NOT be backported and will only go into the 5.4 release.

jgrandja: @martin-v 5.4.0 is being released Sep 2 and I need to get this change in. Are you able to update, as per my last comment, this week?

martin-v: Sorry, didn't notice your response. I will update the PR in a few minutes.

jgrandja:
Thanks for the updates @martin-v ! This is now in master.

PR #8794, opened 2020-07-03
Fix #8484 Provide mechanism for parsing IdP metadata

### LINKED ISSUE #8484: Support for dynamic configuration using IDP metadata URL for SAML SSO integration
**Expected Behavior**

SAML IDP registration can be configured via either IDP metadata URL or file-based metadata instead of manually adding configuration property like entity-id, sso-url, X.509 Certificate, etc.. Benefit would be simpler IDP registration configuration and also it can support dynamic configuration like certificate rotation, etc..
```
spring:
  security:
    saml2:
      relyingparty:
        registration:
          idpone:
            identityprovider:
              metadata-url:
	  idptwo:
            identityprovider:
              metadata-url:
```

**Current Behavior**

Currently  **spring-security-saml2-service-provider** does not support dynamic configuration for IDP using metadata URL. The IDP configuration needs to be added through RelyingPartyRegistration with the different properties like entity-id, sso-url, etc.. Whereas this configuration can be replaced with a single IDP metadata URL which contains all information related to IDP. Also, as an additional option, please allow the same to be configured via file-based metadata.

```
spring:
  security:
    saml2:
      relyingparty:
        registration:
          idpone:
            identityprovider:
              entity-id:
              sso-url:
              verification:
                credentials:
                  - certificate-location:
	  idptwo:
            identityprovider:
              entity-id:
              sso-url:
              verification:
                credentials:
                  - certificate-location:
```

**Context**

My current IDP provides IDP configuration information via metadata URL but due to lack of support in **spring-security-saml2-service-provider**, we have to manually configure each individual property. As a current alternative, we have option to switch back to old way of configuring SAML IDP using **Spring Security SAML Extension**(https://docs.spring.io/autorepo/docs/spring-security-saml/1.0.x-SNAPSHOT/reference/htmlsingle/#quick-start-idp-metadata)

jkubrynski:

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/SamlMetadataParser.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/SamlMetadataParser.java
@@ -1,0 +1,132 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.saml2.provider.service.registration;
+
+import net.shibboleth.utilities.java.support.xml.XMLParserException;
+import org.opensaml.core.config.ConfigurationService;
+import org.opensaml.core.xml.config.XMLObjectProviderRegistry;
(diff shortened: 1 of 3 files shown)


jzheaux:
With the introduction of https://github.com/spring-projects/spring-security/issues/8877, I think this PR can now be simplified.

There is a class in Spring Security's OAuth 2.0 support called `ClientRegistrations` which takes a discovery endpoint and formulates a `ClientRegistration.Builder` from it, adding in all the provider information. I think it would be useful to do the same in the SAML 2.0 support with a class called `RelyingPartyRegistrations`. `RelyingPartyRegistrations` would construct a `RestTemplate` that uses the `HttpMessageConverter` that #8877 introduced.

@jkubrynski, would you be interested in updating this PR to introduce `RelyingPartyRegistrations` and use the `OpenSamlRelyingPartyRegistrationBuilderHttpMessageConverter` to convert the HTTP response to a `RelyingPartyRegistration.Builder`?

jkubrynski:
Sorry @jzheaux but I don't have more time left to support those PRs. You can work on this one on your side or simply close it.

jzheaux:
Closed in favor of f82190b414b788074c339794a250f6273a204cff

PR #8800, opened 2020-07-07
Fix ProviderManager Javadoc typo

lenve:
On lines 240 and 257 of the ProviderManager class, the comments may have some problems.

The original note was: If the parent AuthenticationManager was attempted and successful than it will publish an AuthenticationSuccessEvent

The correct one should be: If the parent AuthenticationManager was attempted and successful then it will publish an AuthenticationSuccessEvent

Replace the word than with then.


rwinch:

Issue #8804, opened 2020-07-07
Configure HTTP Security without extending WebSecurityConfigurerAdapter


eleftherias:
**Expected Behavior**

Similar to the WebFlux security configuration, we should add the capability to configure HTTP Security by registering a `SecurityFilterChain` bean, in Servlet applications.

```
@Bean
SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
	return http
			.antMatcher("/**")
			.authorizeRequests(authorize -> authorize
					.anyRequest().authenticated()
			)
			.build();
}
```

**Current Behavior**

The equivalent configuration by extending the `WebSecurityConfigurerAdapter` looks like this
```
@Configuration
static class SecurityConfig extends WebSecurityConfigurerAdapter {
	@Override
	protected void configure(HttpSecurity http) throws Exception {
		http
			.antMatcher("/**")
			.authorizeRequests(authorize -> authorize
					.anyRequest().authenticated()
			);
	}
}
```

cnry1:
this  good

gbaso:
What about the `AuthenticationManagerBuilder` configuration? Suppose I need to conditionally register additional authentication providers. Do I still need to extend WebSecurityConfigurerAdapter? Exposing a `AuthenticationManager` bean by calling build on the builder throws an exception at startup:

```log
[...]
Caused by: java.lang.IllegalStateException: Cannot apply org.springframework.security.config.annotation.authentication.configuration.AuthenticationConfiguration$EnableGlobalAuthenticationAutowiredConfigurer@7c7d3134 to already built object
        at org.springframework.security.config.annotation.AbstractConfiguredSecurityBuilder.add(AbstractConfiguredSecurityBuilder.java:182) ~[spring-security-config-5.4.1.jar:5.4.1]
        at org.springframework.security.config.annotation.AbstractConfiguredSecurityBuilder.apply(AbstractConfiguredSecurityBuilder.java:138) ~[spring-security-config-5.4.1.jar:5.4.1]
        at org.springframework.security.config.annotation.authentication.configuration.AuthenticationConfiguration.getAuthenticationManager(AuthenticationConfiguration.java:119) ~[spring-security-config-5.4.1.jar:5.4.1]
        at org.springframework.security.config.annotation.web.configuration.HttpSecurityConfiguration.authenticationManager(HttpSecurityConfiguration.java:106) ~[spring-security-config-5.4.1.jar:5.4.1]
[...]

PR #8805, opened 2020-07-07
Remove need for WebSecurityConfigurerAdapter

eleftherias:
Closes gh-8804

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/builders/WebSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/builders/WebSecurity.java
@@ -283,7 +283,8 @@ public final class WebSecurity extends
 		Assert.state(
 				!securityFilterChainBuilders.isEmpty(),
 				() -> "At least one SecurityBuilder<? extends SecurityFilterChain> needs to be specified. "
-						+ "Typically this done by adding a @Configuration that extends WebSecurityConfigurerAdapter. "
+						+ "Typically this is done by exposing a SecurityFilterChain bean "
+						+ "or by adding a @Configuration that extends WebSecurityConfigurerAdapter. "
 						+ "More advanced users can invoke "
 						+ WebSecurity.class.getSimpleName()
 						+ ".addSecurityFilterChainBuilder directly");
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configuration/HttpSecurityConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configuration/HttpSecurityConfiguration.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.config.annotation.web.configuration;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Scope;
+import org.springframework.security.authentication.AuthenticationManager;
+import org.springframework.security.config.annotation.ObjectPostProcessor;
+import org.springframework.security.config.annotation.authentication.builders.AuthenticationManagerBuilder;
+import org.springframework.security.config.annotation.authentication.configuration.AuthenticationConfiguration;
+import org.springframework.security.config.annotation.web.builders.HttpSecurity;
+import org.springframework.security.config.annotation.web.configurers.DefaultLoginPageConfigurer;
+import org.springframework.security.web.context.request.async.WebAsyncManagerIntegrationFilter;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.springframework.security.config.Customizer.withDefaults;
+
+/**
+ * {@link Configuration} that exposes the {@link HttpSecurity} bean.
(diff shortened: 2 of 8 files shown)


config/src/test/java/org/springframework/security/config/annotation/web/configuration/HttpSecurityConfigurationPowermockTests.java
```diff
@@ -0,0 +1,88 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+
+/**
+ * Tests for {@link HttpSecurityConfiguration}.
+ *
+ * @author Eleftheria Stein
+ */
+@RunWith(PowerMockRunner.class)
```
rwinch, line 49: I wonder if we could avoid the use of powermock here? Instead we could validate that using the resulting WebAsncManager uses SecurityContextCallableProcessingInterceptor by ensuring the SecurityContext is propagated

config/src/main/java/org/springframework/security/config/annotation/web/builders/WebSecurity.java
```diff
@@ -208,6 +205,25 @@ public WebSecurity addSecurityFilterChainBuilder(
...
+	 * Typically this method is invoked automatically within the framework from
+	 * {@link WebSecurityConfigurerAdapter#init(WebSecurity)}
+	 * </p>
+	 *
+	 * @since 5.4
+	 * @param securityFilterChain the {@link SecurityFilterChain} instance
+	 * @return the {@link WebSecurity} for further customizations
+	 */
+	public WebSecurity addSecurityFilterChain(SecurityFilterChain securityFilterChain) {
+		this.securityFilterChains.add(securityFilterChain);
```
rwinch, line 223: I wonder if we could/should use `this.securityFilterChainBuilders.add(() -> securityFilterChain)` instead. This would mean that performBuild could stay the same and be simplified (it only operates on a single collection instead of two.

config/src/main/java/org/springframework/security/config/annotation/web/configuration/WebSecurityConfiguration.java
```diff
@@ -95,12 +100,27 @@ public static DelegatingApplicationListener delegatingApplicationListener() {
 	public Filter springSecurityFilterChain() throws Exception {
 		boolean hasConfigurers = webSecurityConfigurers != null
 				&& !webSecurityConfigurers.isEmpty();
-		if (!hasConfigurers) {
+		boolean hasFilterChain = !securityFilterChains.isEmpty();
+		if (hasConfigurers && hasFilterChain) {
+			throw new IllegalStateException(
+					"Found WebSecurityConfigurerAdapter as well as SecurityFilterChain." +
```
rwinch, line 106: Is there a reason to limit to one or the other?

rwinch:
I also wonder if we should consider deprecating WebSecurityConfigurerAdapter and update the samples? Perhaps that is something for another issue?

config/src/test/java/org/springframework/security/config/annotation/web/configuration/HttpSecurityConfigurationPowermockTests.java
eleftherias: I'm not sure how to do that. Is there a similar test somewhere else in the codebase?

config/src/main/java/org/springframework/security/config/annotation/web/configuration/WebSecurityConfiguration.java
eleftherias: If we allow both, it brings up concerns around ordering.

Using the `@Order` annotation doesn't make sense because they are different components, so their order cannot be compared.

The other option I see is to always apply one before the other, for example, first apply all `WebSecurityConfigurerAdapter` configurations and then all `SecurityFilterChain` configurations. But I can see that causing more confusion.

eleftherias:
@rwinch I created gh-8821 to update the samples.

I had not thought about deprecating `WebSecurityConfigurerAdapter`. 
It may come as a surprise to users, since it is such a core part of Spring Security. 
Maybe we can consider deprecating it in 5.5, once everyone is familiar with the `SecurityFilterChain` bean approach.

rwinch:
If we are going to deprecate it, I'd rather give users more time to update. The deprecation can give instructions on how to migrate (which shouldn't be too painful). We'd likely remove the support next year in 6.0, but certainly no sooner.

config/src/test/java/org/springframework/security/config/annotation/web/configuration/HttpSecurityConfigurationPowermockTests.java
rwinch: I don't know if there is. I'd like to avoid using Powermock where possible. I believe you would need a downstream Filter use `WebAsncManager` in a `Thread` that asserts the `SecurityContext`

config/src/main/java/org/springframework/security/config/annotation/web/configuration/HttpSecurityConfiguration.java
```diff
@@ -0,0 +1,117 @@
...
+	private AuthenticationManager authenticationManager() throws Exception {
+		if (this.authenticationManager != null) {
+			return this.authenticationManager;
+		} else {
+			return this.authenticationConfiguration.getAuthenticationManager();
+		}
+	}
+
+	private Map<Class<?>, Object> createSharedObjects() {
+		Map<Class<?>, Object> sharedObjects = new HashMap<>();
```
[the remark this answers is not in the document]
eleftherias: My understanding is that the `WebSecurityConfigurerAdapter` adds the additional shared objects because they may be overridden within that class, for example using `configure(AuthenticationManagerBuilder auth)`.
When using the `SecurityFilterChain` bean, you would need to expose a `UserDetailsService` bean and I confirmed that the `UserDetailsService` is used even if it's not explicitly added to the shared objects here. 
It was similar for the `ContentNegotiationStrategy` and `AuthenticationTrustResolver`.

imod:
I see the value to use the `SecurityFilterChain`, but how do I register multiple `AuthenticationProvider` with this approach?

With the adapter I could simply do:
```
	@Override
	protected void configure(AuthenticationManagerBuilder auth) throws Exception {
		auth.authenticationProvider(new Provider1());
		auth.authenticationProvider(new Provider2());
		auth.authenticationProvider(new Provider3());
	}	
```

Issue #8789, opened 2020-07-02
Add Static Factories to Saml2X509Credential


jzheaux:
`Saml2X509Credential` constructors are somewhat complicated to use. 

For example, there are possible constructs that don't make sense:

```java
new Saml2X509Credential(privateKey, x509Cerificate, Saml2X509CredentialType.ENCRYPTION)
```

The above is awkward in the context of SAML 2.0 since the common use case for encryption is to encrypt something using the other party's public certificate, not their private key as well.

Also, some constructs are possible at compile-time, but then throw an exception at runtime:

```java
new Saml2X509Credential(x509Cerificate, Saml2X509CredentialType.DECRYPTION)
```
When run, the above will error because no private key was supplied.

Both of these scenarios can be addressed by adding some simple static factories to `Saml2X509Credential` like:

```java
public static Saml2X509Credential encryption(X509Certificiate certificate) { ... }

public static Saml2X509Credential decryption(PrivateKey key, X509Certificate certificate) { ... }
```

The nice thing about these as well is that they remove the application's need to work with the `Saml2X509CredentialType` class, further simplifying their application.

ThomasVitale:
Hi @jzheaux, can I help with this task?

jzheaux:
Nice, @ThomasVitale! The issue is yours.

ThomasVitale:
@jzheaux I opened a PR here: https://github.com/spring-projects/spring-security/pull/8822 I added static factories for the 4 main operations plus unit tests. How does it look?

PR #8822, opened 2020-07-09
Add Static Factories to Saml2X509Credential

ThomasVitale:
- Add static factories to Saml2X509Credential for verification, encryption,
signing, and decryption.
- Add unit tests for new static factories in Saml2X509Credential.

Fixes gh-8789

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/credentials/Saml2X509Credential.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/credentials/Saml2X509Credential.java
@@ -46,6 +46,40 @@ public class Saml2X509Credential {
 	private final Set<Saml2X509CredentialType> credentialTypes;
 
 	/**
+	 * Create a {@link Saml2X509Credential} that can be used for encryption.
+	 * @param certificate the certificate to use for encryption
+	 */
+	public static Saml2X509Credential encryption(X509Certificate certificate) {
+		return new Saml2X509Credential(certificate, Saml2X509CredentialType.ENCRYPTION);
+	}
+
+	/**
+	 * Create a {@link Saml2X509Credential} that can be used for verification.
+	 * @param certificate the certificate to use for verification
+	 */
+	public static Saml2X509Credential verification(X509Certificate certificate) {
+		return new Saml2X509Credential(certificate, Saml2X509CredentialType.VERIFICATION);
+	}
+
+	/**
+	 * Create a {@link Saml2X509Credential} that can be used for decryption.
+	 * @param privateKey the private key to use for decryption
+	 * @param certificate the certificate to use for decryption
+	 */
+	public static Saml2X509Credential decryption(PrivateKey privateKey, X509Certificate certificate) {
+		return new Saml2X509Credential(privateKey, certificate, Saml2X509CredentialType.DECRYPTION);
+	}
+
+	/**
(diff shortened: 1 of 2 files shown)


jzheaux:
@ThomasVitale, thanks for the PR!

Based on the build output, it appears some tests may have failed. Are you able to take a look at those?

ThomasVitale:
@jzheaux Yeah, something failed but it didn't seem linked to the new code in this PR. After pulling the latest changes from the master branch and running the build again, now there's no failure.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/credentials/Saml2X509Credential.java
```diff
@@ -45,6 +45,44 @@
 	private final X509Certificate certificate;
 	private final Set<Saml2X509CredentialType> credentialTypes;
 
+	/**
+	 * Creates a Saml2X509Credentials representing Identity Provider credentials
```
jzheaux, line 49: I don't think it's entirely true that encryption credentials are for the Identity Provider. Or, while that's certainly the common case, I think it makes more sense to affirm that in something that has more context like `RelyingPartyRegistration`.

Would you please change this to something like:

```
Create a {@link Saml2X509Credential} that can be used for encryption.

@param certificate the certificate to use for encryption
```

jzheaux: Also, based on the [Code Style guidelines](https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting) would you please use the imperative ("Create" instead of "Creates") in your JavaDoc?

ThomasVitale: @jzheaux I updated the JavaDoc according to your comments and the code style guidelines.

ThomasVitale:
The build is failing with different errors not related to my changes, including an unavailable "commons-codec" dependency. How can I fix that?

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/credentials/Saml2X509Credential.java
```diff
@@ -45,6 +45,42 @@
...
+	/**
+	 * Create a {@link Saml2X509Credential} that can be used for verification.
+	 * @param certificate the certificate to use for verification
+	 */
+	public static Saml2X509Credential verification(X509Certificate certificate) {
+		return new Saml2X509Credential(certificate, Saml2X509CredentialType.VERIFICATION);
+	}
+
+	/**
+	 * Create a {@link Saml2X509Credential} representing Service Provider credentials
```
jzheaux, line 65: Will you please change the remaining two JavaDocs to match the first two?

ThomasVitale: Done

jzheaux:
Thanks, @ThomasVitale! This is now merged into `master` via 3978cc591f197f0baa816e43affbdbb71049438d and 97ccbe5df20c1f9999784d2d0d84d159965bdaca.

Note that `credentials.Saml2X509Credential` was deprecated in favor of `core.Saml2X509Credential recently so I added a polish commit to move the methods to the new class in `core`.

PR #8848, opened 2020-07-19
Exclude magic number for default BCrypt strength

SvensenSeven:
@pivotal-issuemaster This is an Obvious Fix

### DIFF
--- a/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCrypt.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCrypt.java
@@ -370,6 +370,7 @@ public class BCrypt {
 	};
 	static final int MIN_LOG_ROUNDS = 4;
 	static final int MAX_LOG_ROUNDS = 31;
+	static final int DEFAULT_ROUNDS = 10;
 
 	// Expanded Blowfish key
 	private int P[];
--- a/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.java
@@ -43,7 +43,7 @@ public class BCryptPasswordEncoder implements PasswordEncoder {
 
 
 	public BCryptPasswordEncoder() {
-		this(-1);
+		this(BCrypt.DEFAULT_ROUNDS);
 	}
 
 	/**
@@ -65,7 +65,7 @@ public class BCryptPasswordEncoder implements PasswordEncoder {
 	 * @param random  the secure random instance to use
 	 */
 	public BCryptPasswordEncoder(BCryptVersion version, SecureRandom random) {
-		this(version, -1, random);
+		this(version, BCrypt.DEFAULT_ROUNDS, random);
 	}
 
 	/**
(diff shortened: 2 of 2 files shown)


crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.java
```diff
@@ -90,11 +90,11 @@ public BCryptPasswordEncoder(BCryptVersion version, int strength) {
 	 * @param random   the secure random instance to use
 	 */
 	public BCryptPasswordEncoder(BCryptVersion version, int strength, SecureRandom random) {
-		if (strength != -1 && (strength < BCrypt.MIN_LOG_ROUNDS || strength > BCrypt.MAX_LOG_ROUNDS)) {
+		if (strength < BCrypt.MIN_LOG_ROUNDS || strength > BCrypt.MAX_LOG_ROUNDS) {
```
rwinch, line 93: Unfortunately, this breaks passivity for users who are providing the strength as -1. For that reason, unless we have a really good reason I don't think we can make this change.

crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCrypt.java
```diff
@@ -370,6 +370,7 @@
 	};
 	static final int MIN_LOG_ROUNDS = 4;
 	static final int MAX_LOG_ROUNDS = 31;
+	static final int DEFAULT_ROUNDS = 10;
```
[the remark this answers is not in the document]
SvensenSeven: Thanks for the reply. Yes, the motivation was to clean up the code.

Yes, the definition of DEFAULT_ROUNDS could be moved from BCrypt to BCryptPasswordEncoder
```suggestion
```

crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.java
SvensenSeven: > Unfortunately, this breaks passivity for users who are providing the strength as -1

While I agree that this is true, I hardly could imaging a developer using -1 to get a default strength (Spring Security doesn't use it itself). At least, such behaviour is not documented. 
Javadoc comment for 'strength' parameter states that
`* @param strength the log rounds to use, between 4 and 31`

Moreover, to get a default strength a default constructor could be used BCryptPasswordEncoder(), which is natural, rather than BCryptPasswordEncoder(-1).

Also providing a default strength value directly in the statement
`this.strength = strength == -1 ? 10 : strength;`
skips the check
`(strength < BCrypt.MIN_LOG_ROUNDS || strength > BCrypt.MAX_LOG_ROUNDS)`

jzheaux: Good points, @SvensenSeven, the developer would certainly have very reasonable workarounds.

However if allowing -1 isn't causing some kind of problem, then remaining passive needs to take priority.

jzheaux:
Hi, @SvensenSeven! Are you able to make the requested changes?

jzheaux:
Closing as the change is not passive and does not fix any bugs.

Issue #8783, opened 2020-07-01
Support custom filter in Server Kotlin DSL


eleftherias:
We should add support for adding a custom filter in the Server Kotlin DSL.
Similar to `ServerHttpSecurity`, we should have 3 ways to add a filter, `addFilterAt`, `addFilterBefore` and `addFilterAfter`.

The configuration should look like this
```
@Bean
open fun springWebFilterChain(http: ServerHttpSecurity): SecurityWebFilterChain {
    return http {
        addFilterAfter(TestWebFilter(), SecurityWebFiltersOrder.SECURITY_CONTEXT_SERVER_WEB_EXCHANGE)
    }
}
```

evgeniycheban:
I can take this.

eleftherias:
Thanks @evgeniycheban! The issue is yours.

PR #8850, opened 2020-07-19
Support custom filter in Server Kotlin DSL

evgeniycheban:
Closes gh-8783

### DIFF
--- a/config/src/main/kotlin/org/springframework/security/config/web/server/ServerHttpSecurityDsl.kt
+++ b/config/src/main/kotlin/org/springframework/security/config/web/server/ServerHttpSecurityDsl.kt
@@ -20,6 +20,7 @@ import org.springframework.security.oauth2.client.registration.ReactiveClientReg
 import org.springframework.security.web.server.SecurityWebFilterChain
 import org.springframework.security.web.server.util.matcher.ServerWebExchangeMatcher
 import org.springframework.web.server.ServerWebExchange
+import org.springframework.web.server.WebFilter
 
 /**
  * Configures [ServerHttpSecurity] using a [ServerHttpSecurity Kotlin DSL][ServerHttpSecurityDsl].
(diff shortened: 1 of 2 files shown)


config/src/main/kotlin/org/springframework/security/config/web/server/ServerHttpSecurityDsl.kt
```diff
@@ -89,6 +90,33 @@ class ServerHttpSecurityDsl(private val http: ServerHttpSecurity, private val in
         this.http.securityMatcher(securityMatcher)
     }
 
+    /**
+     * Adds a [WebFilter] at a specific position.
```
eleftherias, line 94: It would be nice to have an example of how to use this method, similar to what we have in `HttpSecurityDsl`.

The same applies for `addFilterBefore` and `addFilterAfter`.

config/src/test/kotlin/org/springframework/security/config/web/server/ServerHttpSecurityDslTests.kt
```diff
@@ -123,4 +129,74 @@ class ServerHttpSecurityDslTests {
             }
         }
     }
+
+    @Test
+    fun `add filter at is applied`() {
```
eleftherias, line 134: Please make the test names more descriptive.

For example, this test could be renamed to:

`add filter at applies custom at specified filter position`

config/src/main/kotlin/org/springframework/security/config/web/server/ServerHttpSecurityDsl.kt
evgeniycheban: Done.

config/src/test/kotlin/org/springframework/security/config/web/server/ServerHttpSecurityDslTests.kt
evgeniycheban: Done.

eleftherias:
Thanks for the PR @evgeniycheban! This is now merged into master.

Issue #8852, opened 2020-07-20
Make class SimpleSaml2AuthenticatedPrincipal public


JoakimLofgren:
**Expected Behavior**

We are able to serialize/deserialize the class `Saml2Authentication` and `SimpleSaml2AuthenticatedPrincipal` using an object mapper.

**Current Behavior**

We cannot whitelist `SimpleSaml2AuthenticatedPrincipal` as it is package private and the deserialization crashes.

**Context**

We are using `spring-session-data-mongo` to store the `HttpSession` in a mongo collection.
This approach uses a object mapper to convert it to JSON/BSON.

We can whitelist and fix the Saml2Authentication storage with the following kotlin snippets:
```kotlin
objectMapper.addMixIn(Saml2Authentication::class.java, Saml2AuthenticationMixin::class.java)

...

/**
 * Used to whitelist [Saml2Authentication] for [SecurityJackson2Modules].
 */
class Saml2AuthenticationMixin @JsonCreator constructor(
    @JsonProperty("principal") principal: AuthenticatedPrincipal,
    @JsonProperty("saml2Response") saml2Response: String,
    @JsonProperty("authorities") authorities: Collection<GrantedAuthority>
) : Saml2Authentication(principal, saml2Response, authorities) // Nothing special
```

But I cannot add the following mixin as the class is package private:
```kotlin
/**
 * Used to whitelist [SimpleSaml2AuthenticatedPrincipal] for [SecurityJackson2Modules].
 */
class SimpleSaml2AuthenticatedPrincipalMixin // Nothing special
```

This is the exception we get without being able to whitelist the class:
```
com.fasterxml.jackson.databind.JsonMappingException: The class with 
org.springframework.security.saml2.provider.service.authentication.SimpleSaml2AuthenticatedPrincipal and name of 
org.springframework.security.saml2.provider.service.authentication.SimpleSaml2AuthenticatedPrincipal is not whitelisted. If you 
believe this class is safe to deserialize, please provide an explicit mapping using Jackson annotations or by providing a Mixin. If 
the serialization is only done by a trusted source, you can also enable default typing. See https://github.com/spring-projects
/spring-security/issues/4370 for details (through reference chain: 
org.springframework.session.data.mongo.MongoSession["attrs"]->
java.util.HashMap["SPRING_SECURITY_CONTEXT"]->
org.springframework.security.core.context.SecurityContextImpl["authentication"]->
[...]

PR #8861, opened 2020-07-21
Update SimpleSaml2AuthenticatedPrincipal class name

JoakimLofgren:
Rename it to DefaultSaml2AuthenticatedPrincipal to be more in line with
the respective class in the OAuth2 module.

Also make the class public to be able to whitelist the SAML2 auth classes
in Jackson object mappers for deserialization in e.g. Spring Session MongoDB.

Closes gh-8852

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/DefaultSaml2AuthenticatedPrincipal.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/DefaultSaml2AuthenticatedPrincipal.java
@@ -16,6 +16,8 @@
 
 package org.springframework.security.saml2.provider.service.authentication;
 
+import org.springframework.util.Assert;
+
 import java.io.Serializable;
 import java.util.List;
 import java.util.Map;
@@ -24,14 +26,17 @@ import java.util.Map;
  * Default implementation of a {@link Saml2AuthenticatedPrincipal}.
  *
  * @author Clement Stoquart
- * @since 5.2.2
+ * @since 5.4
  */
-class SimpleSaml2AuthenticatedPrincipal implements Saml2AuthenticatedPrincipal, Serializable {
+public class DefaultSaml2AuthenticatedPrincipal implements Saml2AuthenticatedPrincipal, Serializable {
 
 	private final String name;
 	private final Map<String, List<Object>> attributes;
 
-	SimpleSaml2AuthenticatedPrincipal(String name, Map<String, List<Object>> attributes) {
+	public DefaultSaml2AuthenticatedPrincipal(String name, Map<String, List<Object>> attributes) {
+		Assert.notNull(name, "name cannot be null");
+		Assert.notNull(attributes, "attributes cannot be null");
+
 		this.name = name;
 		this.attributes = attributes;
 	}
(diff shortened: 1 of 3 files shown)


saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/DefaultSaml2AuthenticatedPrincipal.java
```diff
@@ -26,12 +26,12 @@
...
  * @since 5.2.2
  */
-class SimpleSaml2AuthenticatedPrincipal implements Saml2AuthenticatedPrincipal, Serializable {
+public class DefaultSaml2AuthenticatedPrincipal implements Saml2AuthenticatedPrincipal, Serializable {
 
 	private final String name;
 	private final Map<String, List<Object>> attributes;
 
-	SimpleSaml2AuthenticatedPrincipal(String name, Map<String, List<Object>> attributes) {
+	DefaultSaml2AuthenticatedPrincipal(String name, Map<String, List<Object>> attributes) {
```
jzheaux, line 34: I'm thinking the constructor should be public as well so that applications can use it in their own authentication providers.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/DefaultSaml2AuthenticatedPrincipal.java
```diff
@@ -26,12 +26,12 @@
...
 		this.name = name;
```
jzheaux, line 35: Now that this is going to be for general use, will you please add validation checks in the constructor here, similar to:

```java
Assert.notNull(name, "name cannot be null");
```

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/DefaultSaml2AuthenticatedPrincipal.java
JoakimLofgren: Ah, missed that.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/DefaultSaml2AuthenticatedPrincipal.java
JoakimLofgren: Same for `attributes` or just `name` ?

JoakimLofgren: I did both :)

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/DefaultSaml2AuthenticatedPrincipal.java
```diff
@@ -26,12 +28,15 @@
  * @author Clement Stoquart
  * @since 5.2.2
```
jzheaux, line 29: Since this class is just becoming generally available in 5.4, it makes more sense to indicate that in the `@since`. Otherwise, developers may wonder why they can't find `DefaultSaml2AuthenticatedPrincipal` in 5.3, for example.

Will you please change this to `@since 5.4`?

jzheaux: You are also welcome to add yourself as an author, if you like.

jzheaux:
Thanks, @JoakimLofgren! This is now merged into `master`.

PR #8878, opened 2020-07-24
 Fix #6594 Legacy Migration Incompatibility

### LINKED ISSUE #6594: Why does MessageDigestPasswordEncoder consider '{}' to be part of the salt?
### Summary

I'm upgrading from Spring Security 4.x to 5.x.

The ReflectionSaltSource from Spring 4 lets us configure a custom salt. But that's removed in Spring Security 5. I then found out that I should use MessageDigestPasswordEncoder. It has a detailed java-doc but unfortunately the doc is like a bag of words without conveying any structured information (I tried multiple times; my bad if I was ignorant).

Anyways I thought I should do the following based on my limited understanding of the javadoc.

`Old system with 4.x` - myEncodedPassword and mySalt are passed separately to the encoder.

`New System with 5.x` - Pass one field with the value `{mySalt}myEncodedPassword` to the MessageDigestPasswordEncoder

### Actual Behavior

The Problem was that when MessageDigestPasswordEncoder sees `{mySalt}encodedPassword`, it uses `{mySalt}` (with the {}) as the salt instead of using `mySalt` as the salt . I'm confused.

### Expected Behavior

I'd expect the salt to be `mySalt` and not `{mySalt}`

### Version

group='org.springframework.security', module='spring-security-core', version='5.1.4.RELEASE'

### Sample

```
@Grab(group='org.springframework.security', module='spring-security-core', version='5.1.4.RELEASE')
import org.springframework.security.crypto.password.MessageDigestPasswordEncoder

String password = 'myPassword'
String salt_1 = 'mySalt'
String salt_2 = '{mySalt}'
// http://www.lorem-ipsum.co.uk/hasher.php generated below hashes
String encodedPasswordWithSalt_1 = '57bc828628811a10496215e217b7ae9b714c859fc7a8b1c678c9a0cc40aac422'
String encodedPasswordWithSalt_2 = 'a18b53fc58843def1e08e00a718f40d6f8eda0b97ef97824b5078c1fad93c0c5'

MessageDigestPasswordEncoder encoder = new MessageDigestPasswordEncoder('SHA-256')

// expected to match but did not
println "expected=true, actual=" + encoder.matches(password, "{${salt_1}}${encodedPasswordWithSalt_1}") 

// why does this match?
println "expected=false, actual=" + encoder.matches(password, "{${salt_1}}${encodedPasswordWithSalt_2}") 
```

dkbast:
Changes the behaviour of extractSalt to match the documentation (and the expected behaviour). Closes spring-projects#6594

Before:
prefixEncodedPassword = "{this_is_the_salt}hash"
extractSalt(prefixEncodedPassword) == "{this_is_the_salt}"

Now:
prefixEncodedPassword = "{this_is_the_salt}hash"
extractSalt(prefixEncodedPassword) == "this_is_the_salt"

### DIFF
--- a/crypto/src/main/java/org/springframework/security/crypto/password/MessageDigestPasswordEncoder.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/password/MessageDigestPasswordEncoder.java
@@ -169,6 +169,6 @@ public class MessageDigestPasswordEncoder implements PasswordEncoder {
 		if (end < 0) {
 			return "";
 		}
-		return prefixEncodedPassword.substring(start, end + 1);
+		return prefixEncodedPassword.substring(start + 1, end);
 	}
 }


pivotal-issuemaster:
@dkbast Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=8878)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=8878) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@dkbast Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=8878)!

rwinch:
Can you expand on what you believe the problem is? This PR breaks the build.

dkbast:
> Can you expand on what you believe the problem is? This PR breaks the build.

When working with legacy code it is common to have the salt and hash in separate columns of the database. The salt used to decode is randomly generated and does not include the forced curly brackets which the MessageDigestPasswordEncoder enforces.

The expected behaviour would be to pass a string like "{salt}hash" and have the MessageDigestPasswordEncoder use "salt" as the salt and not "{salt}" since the hash was generated with "salt".

This is the same problem encountered in #6594 

The questioning the issue is: what is the point in having a backwards compatible algorithm if it cannot be applied to old data.

I marked this as draft since I decided to rewrite the class in my project and didn't get around finding a general solution.

madorb:
any updates on this?

rwinch:
We cannot make these changes as it breaks passivity.

Issue #8453, opened 2020-05-02
security:client-registrations doesn't take propertyconfigurer properties


richardcs:
Using Spring Security 5.3.1.RELEASE

I use XML based configuration for most security setup as I have customizations that need to be dynamically processed. 

Using <security:client-registrations... Everything works fine if I hardcode the client-id and client-secret but using a propertyConfigurer the values aren't used. 

I have this setup for the propertyConfigurer

` <context:property-placeholder location="file:${app.properties}"/>`

app.properties contains these values:

`oauth2.client.id=<redacted>`
`oauth2.client.secret=<redacted>`

client-registrations is setup like this:

```
<security:client-registrations>
    <security:client-registration registration-id="github"
                         client-id="${oauth2.client.id}"
                         client-secret="${oauth2.client.secret}"
                         provider-id="github"/>
  </security:client-registrations>
```

Values are not propagated even though in other beans (not in the security namespace) are propagated through the propertyConfigurer.

jgrandja:
@richardcs Property placeholders are not taken into account with the initial support for xml config. This would be an enhancement we can certainly add.

Would you be interested in submitting a PR for this?

richardcs:
I can definitely take a look. Could you give me some pointers about how that enhancement would be added?

jgrandja:
@richardcs I would have to spend some time myself to figure out how it can be done. Please see if you can figure it out and if you're having issues then I can jump in and help out.

rwinch:
@richardcs Thanks for looking into this See if this commit helps you https://github.com/spring-projects/spring-security/commit/4542f00b14ba6dc8bb8c1cf4365dc7dad5fe5a3c

evgeniycheban:
I can work on it.

jgrandja:
Thanks @evgeniycheban ! The issue is yours.

evgeniycheban:
@jgrandja I've submitted PR.
Please take a look when you have a moment.

abhishek-bafna-amdhan:
@jzheaux I am wondering if a similar issue might also apply to SAML integration within Spring Security. I have encountered the same issue and created an issue https://github.com/spring-projects/spring-security/issues/14645

PR #8880, opened 2020-07-26
Resolve oauth2 client-id, client-secret placeholders

evgeniycheban:
Closes gh-8453

### DIFF
--- a/config/src/main/java/org/springframework/security/config/oauth2/client/ClientRegistrationsBeanDefinitionParser.java
+++ b/config/src/main/java/org/springframework/security/config/oauth2/client/ClientRegistrationsBeanDefinitionParser.java
@@ -42,6 +42,7 @@ import org.springframework.util.xml.DomUtils;
 
 /**
  * @author Ruby Hartono
+ * @author Evgeniy Cheban
  * @since 5.3
  */
 public final class ClientRegistrationsBeanDefinitionParser implements BeanDefinitionParser {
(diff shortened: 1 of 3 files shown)


config/src/main/java/org/springframework/security/config/oauth2/client/ClientRegistrationsBeanDefinitionParser.java
```diff
@@ -104,8 +105,10 @@ public BeanDefinition parse(Element element, ParserContext parserContext) {
 				}
 			}
 			getOptionalIfNotEmpty(clientRegistrationElt.getAttribute(ATT_CLIENT_ID))
+					.map(parserContext.getReaderContext().getEnvironment()::resolvePlaceholders)
```
jgrandja, line 108: There may be other placeholders that need to be resolved other than `client-id` and `client-secret` and I would like to avoid adding `parserContext.getReaderContext().getEnvironment()::resolvePlaceholders` to every element.

If you look at this [commit](https://github.com/spring-projects/spring-security/commit/4542f00b14ba6dc8bb8c1cf4365dc7dad5fe5a3c#diff-6a79f880c5e53b19f12658d702e68cbaR325), could we simply register a `org.springframework.beans.factory.config.PropertyPlaceholderConfigurer` `@Bean` to take care of it?

evgeniycheban: Thanks for the review @jgrandja.
In this case, we do not register `ClientRegistration` as a bean, so we need to manually resolve property placeholders.

Maybe it would be better to move `parserContext.getReaderContext().getEnvironment()::resolvePlaceholders` to `getOptionalIfNotEmpty` method?

jgrandja: Yes you are right @evgeniycheban - `ClientRegistration` are not beans so that won't work.

> Maybe it would be better to move `parserContext.getReaderContext().getEnvironment()::resolvePlaceholders` to `getOptionalIfNotEmpty` method?

Agreed.

jgrandja:
Thanks for the PR @evgeniycheban ! This is now in master.

Issue #8865, opened 2020-07-21
ServerBearerTokenAuthenticationConverter throws exceptions instead of signalling error


qavid:
**Describe the bug**
ServerBearerTokenAuthenticationConverter throws exceptions in case
bearer token doesn't match pattern 
or 
authorization header and access token query param are both present in request.

**To Reproduce**
Send invalid bearer token in authorization header. For example, you can add test bellow to ServerBearerTokenAuthenticationConverterTests.

```
@Test
public void simpleTest() {
	MockServerHttpRequest.BaseBuilder<?> request = MockServerHttpRequest
			.get("/")
			.header(HttpHeaders.AUTHORIZATION, "bearer !!");

	assertThatCode(() -> this.converter.convert(MockServerWebExchange.from(request))).doesNotThrowAnyException();
}
```

**Expected behavior**
ServerBearerTokenAuthenticationConverter should signal error after subscribing to returned publisher not to throw error after calling convert method.

I'm able to submit a PR.

rwinch:
Thank you for reporting this and volunteering to submit a PR @qavid! We'd love to see a pull request for this issue.

qavid:
Hi @rwinch , I'll work on it next week/weekend after my vacation.

PR #8894, opened 2020-08-02
Resolve Bearer token after subscribing to publisher

qavid:
Bearer token was resolved immediately after calling method convert. In situations when malformed token was provided or authorization header and access token query param were present in request exception was thrown instead of signalling error.
After this change Bearer token is resolved on subscription and invalid states are handled by signaling error to subscriber.

Closes gh-8865


rwinch:
Thanks for the PR @qavid! I've merged this and it will be part of our next release

PR #8899, opened 2020-08-03
Add DefaultRelyingPartyResolver

jzheaux:

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


config/src/test/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurerTests.java
```diff
@@ -276,22 +274,11 @@ protected void configure(HttpSecurity http) throws Exception {
...
-			};
-
 			http
 				.authorizeRequests(authz -> authz
 						.anyRequest().authenticated()
 				)
-				.saml2Login(saml2 -> saml2
-						.addObjectPostProcessor(processor)
-				);
+				.saml2Login(saml2 -> {});
```
evgeniycheban, line 281: `Customizer#withDefaults`looks more readable to me.

saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/web/DefaultRelyingPartyRegistrationResolverTests.java
```diff
@@ -0,0 +1,67 @@
...
+import org.springframework.security.saml2.provider.service.registration.RelyingPartyRegistrationRepository;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.springframework.security.saml2.provider.service.registration.TestRelyingPartyRegistrations.relyingPartyRegistration;
+
+/**
+ * Tests for {@link DefaultRelyingPartyRegistrationResolver}
+ */
+public class DefaultRelyingPartyRegistrationResolverTests {
+	private RelyingPartyRegistration registration = relyingPartyRegistration().build();
```
evgeniycheban, line 33: This field can be `final`. Same for `repository` and `resolver`.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
```diff
@@ -335,6 +339,16 @@ private Saml2AuthenticationRequestFactory getResolver(B http) {
 			}
 			return resolver;
 		}
+
+		private Saml2AuthenticationRequestContextResolver getContextResolver(B http) {
+			Saml2AuthenticationRequestContextResolver resolver = getBeanOrNull(http, Saml2AuthenticationRequestContextResolver.class);
+			if (resolver == null) {
+				return new DefaultSaml2AuthenticationRequestContextResolver(
```
evgeniycheban, line 346: I think it would be better to have a single `return` statement.
`new DefaultSaml2AuthenticationRequestContextResolver` can be moved into `getDefaultContextResolver` method.

jzheaux: I'm of a different opinion on this one. I think returning early is more readable. 

However, since that's probably in the eye of the beholder, what I'd propose for now is aligning with how other similar methods in the class, like `getResolver`. We can consider changing both methods at another time.

jzheaux:
Thanks for the feedback, @evgeniycheban, it was very helpful.

I incorporated it and committed the changes in 015281ff53a3279896729bc0a36e07cd6cbbdb0c to `master`.

PR #8907, opened 2020-08-04
Refactor - apply java code convention

tramyu:
That code syntax doesn't fit the Java convention.

So, I added whitespace


eleftherias:
Thanks for the PR @tramyu!

PR #8963, opened 2020-08-15
#8868 SecurityWebApplicationContextUtils should use WebApplicationCon…

### LINKED ISSUE #8868: SecurityWebApplicationContextUtils should use WebApplicationContextUtils
Initially SecurityWebApplicationContextUtils [copied code](https://github.com/spring-projects/spring-security/blob/5.3.3.RELEASE/web/src/main/java/org/springframework/security/web/context/support/SecurityWebApplicationContextUtils.java#L58) from [WebApplicationContextUtils.findWebApplicationContext](https://github.com/spring-projects/spring-framework/blob/v5.2.8.RELEASE/spring-web/src/main/java/org/springframework/web/context/support/WebApplicationContextUtils.java#L148) to ensure compatibility with older Spring Framework versions. We no longer need to support versions of Spring Framework that old, so we should now call `WebApplicationContextUtils.findWebApplicationContext`

govi20:
PR close #8868 also contains minor refactoring changes.

### DIFF
--- a/taglibs/src/main/java/org/springframework/security/taglibs/authz/AccessControlListTag.java
+++ b/taglibs/src/main/java/org/springframework/security/taglibs/authz/AccessControlListTag.java
@@ -69,7 +69,7 @@ public class AccessControlListTag extends TagSupport {
 
 	// ~ Methods
 	// ========================================================================================================
-
+	@Override
 	public int doStartTag() throws JspException {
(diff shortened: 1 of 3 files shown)


eleftherias:
Thanks for the PR @govi20! This was already merged in gh-9194. 
For that reason, I'm closing this as a duplicate.

Issue #8840, opened 2020-07-17
Consider providing setter of authenticationDetailsSource field in BearerTokenAuthenticationFilter


thecodinglog:
**Expected Behavior**

A custom `AuthenticationDetailsSource` can be injected into `BearerTokenAuthenticationFilter`.

**Current Behavior**

The `authenticationDetailsSource` field is _final_ so it can't.

**Context**

In order to authenticate with bearer token, I'm writing an `AuthenticationManager` having authentication logic using Spring Security.
To do this, I have to put additional information into the details of Authentication and change the authenticationDetailsSource field of `org.springframework.security.oauth2.server.resource.web.BearerTokenAuthenticationFilter` class. But the field is declared as **final** and **forced to use WebAuthenticationDetailsSource** class.

I think it may be possible to provide a setter so that a newly defined `AuthenticationDetailsSource` can be injected from outside, but I am wondering what is forcing it.

Can please you consider providing setter of `authenticationDetailsSource` field in `BearerTokenAuthenticationFilter`?

version : spring-security 5.3.2.RELEASE

jzheaux:
Thanks for the suggestion, @thecodinglog. I think it's reasonable to consider this since other filters also have the same setter.

I'm thinking, though, that adding support for an `AuthenticationConverter` might make a bit more sense, similar to [how things are done with `AuthenticationFilter`](https://github.com/spring-projects/spring-security/blob/master/web/src/main/java/org/springframework/security/web/authentication/AuthenticationFilter.java#L176). Then, you could use whatever `AuthenticationDetailsSource` you want or simply replace the authentication token with one that better suits your needs.

If the code were changed from:

```java
String token;

try {
    token = this.bearerTokenResolver.convert(request);
} catch (OAuth2AuthenticationException invalid) {
    this.authenticationEntryPoint.commence(request, response, invalid);
    return;
}
```

to:

```java
Authentication token;

try {
    token = this.authenticationConverter.convert(request);
} catch (OAuth2AuthenticationException invalid) {
    this.authenticationEntryPoint.commence(request, response, invalid);
    return;
}
```

would that allow you to add extra detail that you need?

thecodinglog:
The method you suggested is much more intuitive and make sense than adding a setter. It also perfectly matches what I want.

What I was trying to do was to use the hashed body, requested url and remote address from Request object for authentication.

Thank you.

jzheaux:
[...]

PR #8975, opened 2020-08-19
Add BearerTokenAuthenticationConverter

thecodinglog:
BearerTokenAuthenticationConverter is introduced to solve the problem of not being able to change AuthenticationDetailsSource.

Closes gh-8840

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -39,6 +39,7 @@ import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
 import org.springframework.security.oauth2.jwt.Jwt;
 import org.springframework.security.oauth2.jwt.JwtDecoder;
 import org.springframework.security.oauth2.jwt.NimbusJwtDecoder;
+import org.springframework.security.oauth2.server.resource.authentication.BearerTokenAuthenticationConverter;
 import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationConverter;
 import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationProvider;
 import org.springframework.security.oauth2.server.resource.authentication.OpaqueTokenAuthenticationProvider;
@@ -51,6 +52,7 @@ import org.springframework.security.oauth2.server.resource.web.DefaultBearerToke
 import org.springframework.security.oauth2.server.resource.web.access.BearerTokenAccessDeniedHandler;
 import org.springframework.security.web.AuthenticationEntryPoint;
 import org.springframework.security.web.access.AccessDeniedHandler;
+import org.springframework.security.web.authentication.AuthenticationConverter;
 import org.springframework.security.web.util.matcher.AndRequestMatcher;
 import org.springframework.security.web.util.matcher.MediaTypeRequestMatcher;
 import org.springframework.security.web.util.matcher.NegatedRequestMatcher;
@@ -78,6 +80,8 @@ import org.springframework.web.accept.HeaderContentNegotiationStrategy;
  * authentication failures are handled
  * <li>{@link #bearerTokenResolver(BearerTokenResolver)} - customizes how to resolve a
  * bearer token from the request</li>
+ * <li>{@link #bearerTokenAuthenticationConverter(AuthenticationConverter)}</li> -
+ * customizes how to convert a bear token authentication from the request
  * <li>{@link #jwt(Customizer)} - enables Jwt-encoded bearer token support</li>
  * <li>{@link #opaqueToken(Customizer)} - enables opaque bearer token support</li>
  * </ul>
@@ -159,6 +163,8 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 
 	private BearerTokenRequestMatcher requestMatcher = new BearerTokenRequestMatcher();
 
+	private AuthenticationConverter authenticationConverter;
+
 	public OAuth2ResourceServerConfigurer(ApplicationContext context) {
 		Assert.notNull(context, "context cannot be null");
 		this.context = context;
@@ -186,6 +192,13 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 	public OAuth2ResourceServerConfigurer<H> bearerTokenResolver(BearerTokenResolver bearerTokenResolver) {
 		Assert.notNull(bearerTokenResolver, "bearerTokenResolver cannot be null");
 		this.bearerTokenResolver = bearerTokenResolver;
+		return this;
+	}
+
+	public OAuth2ResourceServerConfigurer<H> bearerTokenAuthenticationConverter(
+			AuthenticationConverter authenticationConverter) {
+		Assert.notNull(authenticationConverter, "authenticationConverter cannot be null");
+		this.authenticationConverter = authenticationConverter;
 		return this;
 	}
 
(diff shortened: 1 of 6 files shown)


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/BearerTokenAuthenticationFilter.java
```diff
@@ -139,10 +134,12 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
 	 * Set the {@link BearerTokenResolver} to use. Defaults to
 	 * {@link DefaultBearerTokenResolver}.
 	 * @param bearerTokenResolver the {@code BearerTokenResolver} to use
+	 * @deprecated Instead, use {@link BearerTokenAuthenticationConverter} explicitly
+	 * @see BearerTokenAuthenticationConverter
 	 */
 	public void setBearerTokenResolver(BearerTokenResolver bearerTokenResolver) {
```
jzheaux, line 140: Will you please add a `@Deprecated` annotation as well?

jzheaux (changes requested):
Thanks for the PR, @thecodinglog! I've left some feedback inline.

Also, would you be able to update `OAuth2ResourceServerConfigurer` to use `setAuthenticationConverter` instead of `BearerTokenResolver`?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/BearerTokenAuthenticationFilter.java
thecodinglog: Sure, I will re-request review soon.

benba:
Thanks for this PR @thecodinglog , I needed exactly this to implement a feature that also needs to add additional information into the details.

I've tried to use it on the current Spring Security version while waiting for 5.5.0, and I was wandering if it wouldn't be easier if `BearerTokenAuthenticationConverter` was an interface that extends `AuthenticationConverter` and define `setBearerTokenResolver` and `setAuthenticationDetailsSource` methods.
The implementation would then be into something like `DefaultBearerTokenAuthenticationConverter`.

The reason I asked this because today, you are forced to subclass `BearerTokenAuthenticationConverter` to be able to customize `BearerTokenAuthenticationFilter` (since the `BearerTokenAuthenticationFilter.setAuthenticationConverter` takes a `BearerTokenAuthenticationConverter`), and one may need a completely different implementation.

Another option could be to use a bare `AuthenticationConverter` and handle the backward compatibility in `BearerTokenAuthenticationFilter.setBearerTokenResolver` with a cast to `BearerTokenAuthenticationConverter` if the `authenticationConverter` is of this type.

thecodinglog:
Thanks for such a good opinion @benba .

This PR focused on being able to change the `AuthenticationDetailsSource` in `BearerTokenAuthenticationFilter`.
Adding a setter to the `BearerTokenAuthenticationFilter` is easiest way but there was an opinion that it would be better to make the implementation similar to other `AuthenticationConverter` implementations. I agreed that and I did like that.

I agree that `BearerTokenAuthenticationConverter` is more flexible if it is an interface, but I disagree to create an interface where the interface's methods are only setters. Because it doesn't clearly explain what the interface does.

If you need an interface for the setter It would be better to use `BearerTokenAuthenticationConverter` in combination with interfaces such as `AuthenticationConverter` and `BearerTokenResolverReplacable`. But this seems too verbose.

As you mentioned, you may need a completely different implementation for `BearerTokenAuthenticationConverter`.
For this, I don't like it too, but I didn't make it a _final_ class so that it could be inherited.

I'm not sure I understand the last opinion which you mentioned but you mean that do the `setBearerTokenResolver` of the `BearerTokenAuthenticationFilter` class as below?

```java
    if(bearerTokenResolver instanceof BearerTokenAuthenticationConverter)
        this.authenticationConverter = (BearerTokenAuthenticationConverter)bearerTokenResolver;
    else
        this.authenticationConverter.setBearerTokenResolver(bearerTokenResolver);
}
```

If so, Since `BearerTokenResolver` and `AuthenticationConverter` are interfaces that have completely different functions, but I think that it is not intuitive from the user's point of view. 
In order to use the above method, I think the users might carefully read the related documentation or see the implementation to get a sense of how to use it.

Thanks again!😃

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/BearerTokenAuthenticationConverter.java
```diff
@@ -0,0 +1,86 @@
...
+
+	private AuthenticationDetailsSource<HttpServletRequest, ?> authenticationDetailsSource = new WebAuthenticationDetailsSource();
+
+	private BearerTokenResolver bearerTokenResolver;
+
+	public BearerTokenAuthenticationConverter() {
+		this.bearerTokenResolver = new DefaultBearerTokenResolver();
+	}
+
+	public BearerTokenAuthenticationConverter(BearerTokenResolver bearerTokenResolver) {
```
jzheaux, line 47: Since this class has a default value for the `AuthenticationConverter`, let's please remove this constructor.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/BearerTokenAuthenticationConverter.java
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Converts from a HttpServletRequest to {@link BearerTokenAuthenticationToken} that can
+ * be authenticated. Null authentication possible if there was no Authorization header
+ * with Bearer Token.
+ *
+ * @author Jeongjin Kim
+ * @since 5.5
+ */
+public class BearerTokenAuthenticationConverter implements AuthenticationConverter {
```
jzheaux, line 37: Let's please leave the class final for now. We can open it up later if necessary.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -189,6 +194,13 @@ public OAuth2ResourceServerConfigurer(ApplicationContext context) {
 		return this;
 	}
 
+	public OAuth2ResourceServerConfigurer<H> bearerTokenAuthenticationConverter(
```
jzheaux, line 197: I think this should take an `AuthenticationConverter` so that applications can switch out the implementation

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/BearerTokenAuthenticationFilter.java
```diff
@@ -144,10 +139,13 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
...
 	 * {@link DefaultBearerTokenResolver}.
 	 * @param bearerTokenResolver the {@code BearerTokenResolver} to use
+	 * @deprecated Instead, use {@link BearerTokenAuthenticationConverter} explicitly
+	 * @see BearerTokenAuthenticationConverter
 	 */
+	@Deprecated
 	public void setBearerTokenResolver(BearerTokenResolver bearerTokenResolver) {
 		Assert.notNull(bearerTokenResolver, "bearerTokenResolver cannot be null");
-		this.bearerTokenResolver = bearerTokenResolver;
+		this.authenticationConverter.setBearerTokenResolver(bearerTokenResolver);
```
jzheaux, line 148: I'd recommend something like this:

```java
Assert.isTrue(this.authenticationConverter instanceof BearerTokenAuthenticationConverter, "bearerTokenResolver and authenticationConverter cannot both be customized in this filter. Since you've customized the authenticationConverter, please consider configuring the bearerTokenResolver there.");
((BearerTokenAuthenticationConverter) this.authenticationConverter).setBearerTokenResolver(bearerTokenResolver);
```

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/BearerTokenAuthenticationFilter.java
```diff
@@ -171,4 +169,15 @@ public void setAuthenticationFailureHandler(final AuthenticationFailureHandler a
 		this.authenticationFailureHandler = authenticationFailureHandler;
 	}
 
+	/**
+	 * Set the {@link BearerTokenAuthenticationConverter} to use. Defaults to
+	 * {@link BearerTokenAuthenticationConverter}.
+	 * @param authenticationConverter the {@code AuthenticationConverter} to use
+	 * @since 5.5
+	 */
+	public void setAuthenticationConverter(BearerTokenAuthenticationConverter authenticationConverter) {
```
jzheaux, line 178: Let's please use the `AuthenticationConverter` interface here so that applications can switch out the implementation completely.

jzheaux (changes requested):
Thanks again, @thecodinglog, and my apologies for the delay in responding. I've left some additional feedback inline.

Also, as a housekeeping item, will you please update the copyright headers in the files in this PR to have an end date of 2021? Thanks!

jzheaux:
Thanks, @thecodinglog! This is now merged into `master` via 31f310fd22b0c2f6b1a0e01ec16821364245bb8b.

I also added a polish via b774e91734b4ae875c7f88a4d9e77be70c110a8d that simplified a method name and rearranged some of the methods and fields.

Issue #8900, opened 2020-08-03
Add AuthorizationManager


rwinch:
We should add an `AuthorizationManager` which is an imperative version of [`ReactiveAuthorizationManager`](https://github.com/spring-projects/spring-security/blob/5.3.3.RELEASE/core/src/main/java/org/springframework/security/authorization/ReactiveAuthorizationManager.java#L31). The class should look something like:

```java
public interface AuthorizationManager<T> {
	AuthorizationDecision check(Supplier<Authentication> authentication, T object);

	default void verify(Supplier<Authentication> authentication, T object) {
		AuthorizationDecision decision = check(authentication, object);
		if (!decision.isGranted()) {
			throw new AccessDeniedException("Access Denied");
		}
	}
}
```

Using something that allows delaying looking up the `Authentication` like `Supplier<Authentication>` vs an `Authentication` directly.

We should also add support for `AuthorizationManager` in `HttpSecurity.authorizeRequests()`.

Finally, we should change around the existing classes that use `AccessDecisionManager` should migrate to `AuthorizationManager` and `AccessDecisionManager` should be marked as deprecated.

evgeniycheban:
@rwinch I would like to work on this.

jzheaux:
It's yours, @evgeniycheban

evgeniycheban:
@jzheaux I created a very draft PR.

Please take a look when you have a moment. Am I moving in the right direction?

evgeniycheban:
Thanks for the review @jzheaux 

@rwinch @jzheaux I have a few questions.
1. Should we implement an `AuthorizationManager` for Method Security within this task?
2. By default, `HttpSecurity.authorizeRequests()` uses `ExpressionUrlAuthorizationConfigurer`, should we add an `AuthorizationManager` that supports `SpEL` expressions?

I've currently added support for `AuthorizationManager` to `UrlAuthorizationConfigurer` in the [2275b04](https://github.com/spring-projects/spring-security/pull/8996/commits/2275b043c4c49d3871aed5600eb755bcc0c693c6)

jzheaux:
Good questions, @evgeniycheban.

> 1. Should we implement an AuthorizationManager for Method Security within this task?

One of the goals of this ticket is to deprecate `AccessDecisionManager`. Given that, we should see if we can eliminate any internal use of it and deprecate any public use.

[...]

PR #8996, opened 2020-09-07
Add AuthorizationManager

evgeniycheban:
Closes gh-8900

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/AuthorizeHttpRequestsConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/AuthorizeHttpRequestsConfigurer.java
@@ -1,0 +1,289 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.config.annotation.web.configurers;
+
+import java.util.List;
+
+import javax.servlet.http.HttpServletRequest;
+
+import org.springframework.context.ApplicationContext;
+import org.springframework.http.HttpMethod;
+import org.springframework.security.authorization.AuthenticatedAuthorizationManager;
+import org.springframework.security.authorization.AuthorityAuthorizationManager;
+import org.springframework.security.authorization.AuthorizationDecision;
+import org.springframework.security.authorization.AuthorizationManager;
+import org.springframework.security.config.annotation.ObjectPostProcessor;
+import org.springframework.security.config.annotation.web.AbstractRequestMatcherRegistry;
+import org.springframework.security.config.annotation.web.HttpSecurityBuilder;
+import org.springframework.security.web.access.intercept.AuthorizationFilter;
+import org.springframework.security.web.access.intercept.DelegatingAuthorizationManager;
+import org.springframework.security.web.access.intercept.RequestAuthorizationContext;
+import org.springframework.security.web.servlet.util.matcher.MvcRequestMatcher;
+import org.springframework.security.web.util.matcher.RequestMatcher;
+import org.springframework.util.Assert;
+
+/**
+ * Adds a URL based authorization using {@link AuthorizationManager}.
+ *
+ * @param <H> the type of {@link HttpSecurityBuilder} that is being configured.
+ * @author Evgeniy Cheban
+ */
+public final class AuthorizeHttpRequestsConfigurer<H extends HttpSecurityBuilder<H>>
+		extends AbstractHttpConfigurer<AuthorizeHttpRequestsConfigurer<H>, H> {
+
+	private final AuthorizationManagerRequestMatcherRegistry registry;
+
+	/**
+	 * Creates an instance.
+	 * @param context the {@link ApplicationContext} to use
+	 */
+	public AuthorizeHttpRequestsConfigurer(ApplicationContext context) {
+		this.registry = new AuthorizationManagerRequestMatcherRegistry(context);
+	}
+
+	/**
+	 * The {@link AuthorizationManagerRequestMatcherRegistry} is what users will interact
+	 * with after applying the {@link AuthorizeHttpRequestsConfigurer}.
+	 * @return the {@link AuthorizationManagerRequestMatcherRegistry} for further
+	 * customizations
+	 */
+	public AuthorizationManagerRequestMatcherRegistry getRegistry() {
+		return this.registry;
+	}
+
+	@Override
+	public void configure(H http) {
+		AuthorizationManager<HttpServletRequest> authorizationManager = this.registry.createAuthorizationManager();
+		AuthorizationFilter authorizationFilter = new AuthorizationFilter(authorizationManager);
+		http.addFilter(postProcess(authorizationFilter));
+	}
+
+	private AuthorizationManagerRequestMatcherRegistry addMapping(List<? extends RequestMatcher> matchers,
+			AuthorizationManager<RequestAuthorizationContext> manager) {
+		for (RequestMatcher matcher : matchers) {
+			this.registry.addMapping(matcher, manager);
+		}
+		return this.registry;
+	}
+
+	/**
+	 * Registry for mapping a {@link RequestMatcher} to an {@link AuthorizationManager}.
+	 *
+	 * @author Evgeniy Cheban
+	 */
+	public final class AuthorizationManagerRequestMatcherRegistry
+			extends AbstractRequestMatcherRegistry<AuthorizedUrl> {
+
+		private final DelegatingAuthorizationManager.Builder managerBuilder = DelegatingAuthorizationManager.builder();
+
+		private List<RequestMatcher> unmappedMatchers;
+
+		private int mappingCount;
+
+		private AuthorizationManagerRequestMatcherRegistry(ApplicationContext context) {
+			setApplicationContext(context);
+		}
+
+		private void addMapping(RequestMatcher matcher, AuthorizationManager<RequestAuthorizationContext> manager) {
+			this.unmappedMatchers = null;
+			this.managerBuilder.add(matcher, manager);
+			this.mappingCount++;
+		}
+
+		private AuthorizationManager<HttpServletRequest> createAuthorizationManager() {
+			Assert.state(this.unmappedMatchers == null,
+					() -> "An incomplete mapping was found for " + this.unmappedMatchers
+							+ ". Try completing it with something like requestUrls().<something>.hasRole('USER')");
+			Assert.state(this.mappingCount > 0,
+					"At least one mapping is required (for example, authorizeHttpRequests().anyRequest().authenticated())");
+			return postProcess(this.managerBuilder.build());
+		}
+
+		@Override
+		public MvcMatchersAuthorizedUrl mvcMatchers(String... mvcPatterns) {
+			return mvcMatchers(null, mvcPatterns);
+		}
+
+		@Override
+		public MvcMatchersAuthorizedUrl mvcMatchers(HttpMethod method, String... mvcPatterns) {
+			return new MvcMatchersAuthorizedUrl(createMvcMatchers(method, mvcPatterns));
+		}
+
+		@Override
+		protected AuthorizedUrl chainRequestMatchers(List<RequestMatcher> requestMatchers) {
+			this.unmappedMatchers = requestMatchers;
+			return new AuthorizedUrl(requestMatchers);
+		}
+
+		/**
+		 * Adds an {@link ObjectPostProcessor} for this class.
+		 * @param objectPostProcessor the {@link ObjectPostProcessor} to use
+		 * @return the {@link AuthorizationManagerRequestMatcherRegistry} for further
+		 * customizations
+		 */
+		public AuthorizationManagerRequestMatcherRegistry withObjectPostProcessor(
+				ObjectPostProcessor<?> objectPostProcessor) {
+			addObjectPostProcessor(objectPostProcessor);
+			return this;
+		}
+
+		/**
+		 * Return the {@link HttpSecurityBuilder} when done using the
+		 * {@link AuthorizeHttpRequestsConfigurer}. This is useful for method chaining.
+		 * @return the {@link HttpSecurityBuilder} for further customizations
+		 */
+		public H and() {
+			return AuthorizeHttpRequestsConfigurer.this.and();
+		}
+
+	}
+
+	/**
+	 * An {@link AuthorizeHttpRequestsConfigurer.AuthorizedUrl} that allows optionally
+	 * configuring the {@link MvcRequestMatcher#setServletPath(String)}.
+	 *
+	 * @author Evgeniy Cheban
+	 */
+	public final class MvcMatchersAuthorizedUrl extends AuthorizedUrl {
+
+		private MvcMatchersAuthorizedUrl(List<MvcRequestMatcher> matchers) {
+			super(matchers);
+		}
+
+		/**
+		 * Configures <code>servletPath</code> to {@link MvcRequestMatcher}s.
+		 * @param servletPath the servlet path
+		 * @return the {@link MvcMatchersAuthorizedUrl} for further customizations
+		 */
+		@SuppressWarnings("unchecked")
+		public MvcMatchersAuthorizedUrl servletPath(String servletPath) {
+			for (MvcRequestMatcher matcher : (List<MvcRequestMatcher>) getMatchers()) {
+				matcher.setServletPath(servletPath);
+			}
+			return this;
+		}
+
+	}
+
+	/**
+	 * An object that allows configuring the {@link AuthorizationManager} for
+	 * {@link RequestMatcher}s.
+	 *
+	 * @author Evgeniy Cheban
+	 */
+	public class AuthorizedUrl {
+
+		private final List<? extends RequestMatcher> matchers;
+
+		/**
+		 * Creates an instance.
+		 * @param matchers the {@link RequestMatcher} instances to map
+		 */
+		AuthorizedUrl(List<? extends RequestMatcher> matchers) {
+			this.matchers = matchers;
+		}
+
+		protected List<? extends RequestMatcher> getMatchers() {
+			return this.matchers;
+		}
+
+		/**
+		 * Specify that URLs are allowed by anyone.
+		 * @return the {@link AuthorizationManagerRequestMatcherRegistry} for further
+		 * customizations
+		 */
+		public AuthorizationManagerRequestMatcherRegistry permitAll() {
+			return access((a, o) -> new AuthorizationDecision(true));
+		}
+
+		/**
+		 * Specify that URLs are not allowed by anyone.
+		 * @return the {@link AuthorizationManagerRequestMatcherRegistry} for further
+		 * customizations
+		 */
+		public AuthorizationManagerRequestMatcherRegistry denyAll() {
+			return access((a, o) -> new AuthorizationDecision(false));
+		}
+
+		/**
+		 * Specifies a user requires a role.
+		 * @param role the role that should be required which is prepended with ROLE_
+		 * automatically (i.e. USER, ADMIN, etc). It should not start with ROLE_
+		 * @return {@link AuthorizationManagerRequestMatcherRegistry} for further
+		 * customizations
+		 */
+		public AuthorizationManagerRequestMatcherRegistry hasRole(String role) {
+			return access(AuthorityAuthorizationManager.hasRole(role));
+		}
+
+		/**
+		 * Specifies that a user requires one of many roles.
+		 * @param roles the roles that the user should have at least one of (i.e. ADMIN,
+		 * USER, etc). Each role should not start with ROLE_ since it is automatically
+		 * prepended already
+		 * @return the {@link AuthorizationManagerRequestMatcherRegistry} for further
+		 * customizations
+		 */
+		public AuthorizationManagerRequestMatcherRegistry hasAnyRole(String... roles) {
+			return access(AuthorityAuthorizationManager.hasAnyRole(roles));
+		}
+
+		/**
+		 * Specifies a user requires an authority.
+		 * @param authority the authority that should be required
+		 * @return the {@link AuthorizationManagerRequestMatcherRegistry} for further
+		 * customizations
+		 */
+		public AuthorizationManagerRequestMatcherRegistry hasAuthority(String authority) {
+			return access(AuthorityAuthorizationManager.hasAuthority(authority));
+		}
+
+		/**
+		 * Specifies that a user requires one of many authorities.
+		 * @param authorities the authorities that the user should have at least one of
+		 * (i.e. ROLE_USER, ROLE_ADMIN, etc)
+		 * @return the {@link AuthorizationManagerRequestMatcherRegistry} for further
+		 * customizations
+		 */
+		public AuthorizationManagerRequestMatcherRegistry hasAnyAuthority(String... authorities) {
+			return access(AuthorityAuthorizationManager.hasAnyAuthority(authorities));
+		}
+
+		/**
+		 * Specify that URLs are allowed by any authenticated user.
+		 * @return the {@link AuthorizationManagerRequestMatcherRegistry} for further
+		 * customizations
+		 */
+		public AuthorizationManagerRequestMatcherRegistry authenticated() {
+			return access(AuthenticatedAuthorizationManager.authenticated());
+		}
+
+		/**
+		 * Allows specifying a custom {@link AuthorizationManager}.
+		 * @param manager the {@link AuthorizationManager} to use
+		 * @return the {@link AuthorizationManagerRequestMatcherRegistry} for further
+		 * customizations
+		 */
+		public AuthorizationManagerRequestMatcherRegistry access(
+				AuthorizationManager<RequestAuthorizationContext> manager) {
+			Assert.notNull(manager, "manager cannot be null");
+			return AuthorizeHttpRequestsConfigurer.this.addMapping(this.matchers, manager);
+		}
+
+	}
+
+}
--- a/config/src/test/java/org/springframework/security/config/annotation/web/configurers/AuthorizeHttpRequestsConfigurerTests.java
+++ b/config/src/test/java/org/springframework/security/config/annotation/web/configurers/AuthorizeHttpRequestsConfigurerTests.java
@@ -1,0 +1,628 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.config.annotation.web.configurers;
+
+import org.junit.Rule;
+import org.junit.Test;
+
+import org.springframework.beans.factory.BeanCreationException;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Bean;
+import org.springframework.security.authorization.AuthorizationManager;
+import org.springframework.security.config.annotation.ObjectPostProcessor;
+import org.springframework.security.config.annotation.web.AbstractRequestMatcherRegistry;
+import org.springframework.security.config.annotation.web.builders.HttpSecurity;
+import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
+import org.springframework.security.config.test.SpringTestRule;
+import org.springframework.security.core.authority.SimpleGrantedAuthority;
+import org.springframework.security.web.SecurityFilterChain;
+import org.springframework.security.web.access.intercept.AuthorizationFilter;
+import org.springframework.security.web.access.intercept.DelegatingAuthorizationManager;
+import org.springframework.security.web.access.intercept.RequestAuthorizationContext;
+import org.springframework.test.web.servlet.MockMvc;
+import org.springframework.test.web.servlet.request.MockHttpServletRequestBuilder;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.servlet.config.annotation.EnableWebMvc;
+
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+import static org.mockito.Mockito.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+import static org.springframework.security.config.Customizer.withDefaults;
+import static org.springframework.security.test.web.servlet.request.SecurityMockMvcRequestPostProcessors.csrf;
+import static org.springframework.security.test.web.servlet.request.SecurityMockMvcRequestPostProcessors.user;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+/**
+ * Tests for {@link AuthorizeHttpRequestsConfigurer}.
+ *
+ * @author Evgeniy Cheban
+ */
+public class AuthorizeHttpRequestsConfigurerTests {
+
+	@Rule
+	public final SpringTestRule spring = new SpringTestRule();
+
+	@Autowired
+	MockMvc mvc;
+
+	@Test
+	public void configureWhenAuthorizedHttpRequestsAndNoRequestsThenException() {
+		assertThatExceptionOfType(BeanCreationException.class)
+				.isThrownBy(() -> this.spring.register(NoRequestsConfig.class).autowire()).withMessageContaining(
+						"At least one mapping is required (for example, authorizeHttpRequests().anyRequest().authenticated())");
+	}
+
+	@Test
+	public void configureWhenAnyRequestIncompleteMappingThenException() {
+		assertThatExceptionOfType(BeanCreationException.class)
+				.isThrownBy(() -> this.spring.register(IncompleteMappingConfig.class).autowire())
+				.withMessageContaining("An incomplete mapping was found for ");
+	}
+
+	@Test
+	public void configureWhenMvcMatcherAfterAnyRequestThenException() {
+		assertThatExceptionOfType(BeanCreationException.class)
+				.isThrownBy(() -> this.spring.register(AfterAnyRequestConfig.class).autowire())
+				.withMessageContaining("Can't configure mvcMatchers after anyRequest");
+	}
+
+	@Test
+	public void configureMvcMatcherAccessAuthorizationManagerWhenNotNullThenVerifyUse() throws Exception {
+		CustomAuthorizationManagerConfig.authorizationManager = mock(AuthorizationManager.class);
+		this.spring.register(CustomAuthorizationManagerConfig.class, BasicController.class).autowire();
+		this.mvc.perform(get("/")).andExpect(status().isOk());
+		verify(CustomAuthorizationManagerConfig.authorizationManager).check(any(), any());
+	}
+
+	@Test
+	public void configureMvcMatcherAccessAuthorizationManagerWhenNullThenException() {
+		CustomAuthorizationManagerConfig.authorizationManager = null;
+		assertThatExceptionOfType(BeanCreationException.class)
+				.isThrownBy(() -> this.spring.register(CustomAuthorizationManagerConfig.class).autowire())
+				.withMessageContaining("manager cannot be null");
+	}
+
+	@Test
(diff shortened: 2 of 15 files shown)


web/src/main/java/org/springframework/security/web/access/RequestMatcherDelegatingAuthorizationManager.java
```diff
@@ -0,0 +1,72 @@
...
+	private final List<RequestMatcherEntry<AuthorizationManager<HttpServletRequest>>> managers;
+
+	public RequestMatcherDelegatingAuthorizationManager(
+			List<RequestMatcherEntry<AuthorizationManager<HttpServletRequest>>> managers) {
+		this.managers = managers;
+	}
+
+	@Override
+	public AuthorizationDecision check(Supplier<Authentication> authentication, HttpServletRequest request) {
+		return this.managers.stream().filter((entry) -> entry.getMatcher().matches(request))
```
jzheaux, line 48: Let's [avoid the use of the Stream API](https://github.com/spring-projects/spring-security/issues/7154).

web/src/main/java/org/springframework/security/web/access/IpAddressAuthorizationManager.java
```diff
@@ -0,0 +1,60 @@
...
+import org.springframework.security.web.util.matcher.RequestMatcher;
+import org.springframework.util.Assert;
+
+/**
+ * An {@link AuthorizationManager} that determines if the current request is authorized
+ * based on a {@link IpAddressMatcher} evaluation.
+ *
+ * @author Evgeniy Cheban
+ */
+public class IpAddressAuthorizationManager implements AuthorizationManager<HttpServletRequest> {
```
jzheaux, line 36: I'm not entirely convinced that this class should be public as these days authorizing by IP address is quite fragile. 

If necessary, I think the DSL's implementation of `hasIpAddress()` could use a private `AuthorizationManager` implementation or a lambda.

jzheaux (changes requested):
Thanks, @evgeniycheban, this is going in the right direction. I've left a couple of comments inline.

web/src/main/java/org/springframework/security/web/access/RequestMatcherDelegatingAuthorizationManager.java
evgeniycheban: Done.

jzheaux (changes requested):
Great progress, @evgeniycheban, it's nice to see this starting to take shape. I've left some additional feedback inline.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/AbstractInterceptUrlConfigurer.java
```diff
@@ -142,29 +103,21 @@ private FilterSecurityInterceptor createFilterSecurityInterceptor(H http,
...
 		AbstractInterceptUrlRegistry() {
 		}
 
-		/**
-		 * Allows setting the {@link AccessDecisionManager}. If none is provided, a
-		 * default {@link AccessDecisionManager} is created.
-		 * @param accessDecisionManager the {@link AccessDecisionManager} to use
-		 * @return the {@link AbstractInterceptUrlConfigurer} for further customization
-		 */
-		public R accessDecisionManager(AccessDecisionManager accessDecisionManager) {
```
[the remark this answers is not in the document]
evgeniycheban: In the current revision this setter in place.

core/src/main/java/org/springframework/security/authorization/AuthorizationManager.java
```diff
@@ -0,0 +1,55 @@
...
+@FunctionalInterface
+public interface AuthorizationManager<T> {
+
+	/**
+	 * Determines if access is granted for a specific authentication and object.
+	 * @param authentication the Authentication to check
+	 * @param object the object to check
+	 * @return a decision
+	 */
+	AuthorizationDecision check(Supplier<Authentication> authentication, T object);
```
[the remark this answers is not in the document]
evgeniycheban: Maybe it would be better to return an empty `Optional` instead of `null`?

core/src/main/java/org/springframework/security/authorization/AffirmativeBasedAuthorizationManager.java
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Simple concrete implementation of {@link AuthorizationManager} that grants access if
+ * any {@link AccessDecisionVoter} returns an affirmative response.
+ *
+ * @param <T> the type of object that the authorization check is being done one.
+ * @author Evgeniy Cheban
+ */
+public final class AffirmativeBasedAuthorizationManager<T> implements AuthorizationManager<T> {
+
+	private final List<AccessDecisionVoter<?>> decisionVoters;
```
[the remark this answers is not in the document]
evgeniycheban: I think this implementation requires less changes and we can reuse voters.
I'm not sure how this can be implemented for Method Security, the `AffirmativeBasedAuthorizationManager` is also used in `GlobalMethodSecurityConfiguration`.

core/src/main/java/org/springframework/security/access/intercept/AbstractSecurityInterceptor.java
```diff
@@ -375,10 +404,24 @@ public boolean isValidateConfigAttributes() {
...
 	public abstract SecurityMetadataSource obtainSecurityMetadataSource();
 
+	/**
+	 * Sets the {@link AccessDecisionManager} instance.
+	 * @param accessDecisionManager the {@link AccessDecisionManager} instance
+	 * @deprecated use {@link #setAuthorizationManager(AuthorizationManager)} instead
+	 */
+	@Deprecated
 	public void setAccessDecisionManager(AccessDecisionManager accessDecisionManager) {
 		this.accessDecisionManager = accessDecisionManager;
```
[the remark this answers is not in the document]
evgeniycheban: Agreed.

core/src/main/java/org/springframework/security/authorization/AffirmativeBasedAuthorizationManager.java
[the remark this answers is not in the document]
jzheaux: While I agree that this might be less change, we don't want to introduce a public API that continues to have the same level of complexity as the existing API. The primary goal of this PR is to simplify the authorization API. Generally speaking, we ought to, going forward, avoid using the `ConfigAttribute` approach.

As you mentioned, this is more deeply embedded in method security. I wonder if this class could be package-private to method security configuration, and we can introduce another ticket to continue to rework method security.

core/src/main/java/org/springframework/security/authorization/AuthorizationManager.java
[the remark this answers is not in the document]
jzheaux: `Optional` adds some extra overhead; it's been observed to cause excess GC at the framework level. While `Optional` is nice, since this could potentially be called on every request, I'd prefer `null`.

core/src/main/java/org/springframework/security/authorization/AffirmativeBasedAuthorizationManager.java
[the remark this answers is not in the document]
evgeniycheban: I forget that `AffirmativeBasedAuthorizationManager` is also used for Web Socket configuration in `AbstractSecurityWebSocketMessageBrokerConfigurer`, maybe for now we can use an adapter for `AccessDecisionManager` and rework Method Security and Web Socket configuration in another ticket?

core/src/main/java/org/springframework/security/access/intercept/AbstractSecurityInterceptor.java
```diff
@@ -161,7 +167,8 @@ public void afterPropertiesSet() {
 	private void validateAttributeDefs(Collection<ConfigAttribute> attributeDefs) {
 		Set<ConfigAttribute> unsupportedAttrs = new HashSet<>();
 		for (ConfigAttribute attr : attributeDefs) {
-			if (!this.runAsManager.supports(attr) && !this.accessDecisionManager.supports(attr)
+			if (!this.runAsManager.supports(attr) && this.accessDecisionManager != null
```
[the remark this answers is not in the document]
evgeniycheban: This test must be done together with `RunAsManager` and `AfterInvocationManager`.
Maybe this test is no longer necessary?

core/src/main/java/org/springframework/security/access/intercept/AbstractSecurityInterceptor.java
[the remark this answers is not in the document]
evgeniycheban: What do you think about `getAccessDecisionManager`?

We can try to cast an `AuthorizationManager` to an `AuthorizationManagerAdapter` and return an `AccessDecisionManager` otherwise throw an exception with message indicating that user should use `getAuthorizationManager` instead.

core/src/main/java/org/springframework/security/access/intercept/AbstractSecurityInterceptor.java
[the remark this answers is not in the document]
evgeniycheban: I think we can solve it like this:

```java
private Predicate<Class<?>> secureObjectValidator = (o) -> true;
private Predicate<ConfigAttribute> configAttributeValidator = (o) -> true;

public void setAccessDecisionManager(AccessDecisionManager accessDecisionManager) {
       this.secureObjectValidator = accessDecisionManager::supports;
       this.configAttributeValidator = accessDecisionManager::supports;
       this.authorizationManager = new AuthorizationManagerAdapter<>(accessDecisionManager,
		    (o) -> obtainSecurityMetadataSource().getAttributes(o));
}
```

What do you think, @jzheaux?

core/src/main/java/org/springframework/security/access/intercept/AbstractSecurityInterceptor.java
```diff
@@ -115,6 +118,8 @@
 
 	private AccessDecisionManager accessDecisionManager;
```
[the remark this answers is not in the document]
evgeniycheban: It might make sense to consider introducing a new filter that will use an `AuthorizationManager` similar to `AuthorizationWebFilter`.

core/src/main/java/org/springframework/security/access/intercept/AbstractSecurityInterceptor.java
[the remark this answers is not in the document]
jzheaux: Ideally, `AuthorizationManagerAdapter` would not be public since it perpetuates the usage of a deprecated API. 

It might be possible to make it a private inner class of `AbstractSecurityInterceptor`, in which case you could access its private `AccessDecisionManager`. The configurers would need to know whether to call `setAccessDecisionManager` or `setAuthorizationManager` in that case. The nice thing about that is that we can also know if the application provided an `AccessDecisionManager` and can correctly error if `setAuthorizationManager` is called thereafter.

evgeniycheban:
@jzheaux I've updated the PR.

core/src/main/java/org/springframework/security/access/intercept/AbstractSecurityInterceptor.java
```diff
@@ -201,17 +212,8 @@ protected InterceptorStatusToken beforeInvocation(Object object) {
 					"An Authentication object was not found in the SecurityContext"), object, attributes);
 		}
 		Authentication authenticated = authenticateIfRequired();
-		if (this.logger.isTraceEnabled()) {
```
jzheaux, line 204: I see that you moved these to the adapter, which is understandable since they need the collection of config attributes.

Have you considered adding equivalent logs to `FilterInvocationAuthorizationManager`? It's helpful to see logs that explain the authorization context, especially reasons for authorization success of failure.

core/src/main/java/org/springframework/security/access/intercept/AbstractSecurityInterceptor.java
```diff
@@ -201,17 +212,8 @@ protected InterceptorStatusToken beforeInvocation(Object object) {
...
-			this.logger.trace(LogMessage.format("Authorizing %s with attributes %s", object, attributes));
-		}
 		// Attempt authorization
-		attemptAuthorization(object, attributes, authenticated);
-		if (this.logger.isDebugEnabled()) {
-			this.logger.debug(LogMessage.format("Authorized %s with attributes %s", object, attributes));
-		}
-		if (this.publishAuthorizationSuccess) {
-			publishEvent(new AuthorizedEvent(object, attributes, authenticated));
```
jzheaux, line 213: I wonder if `FilterInvocationAuthorizationManager` should fire an `AuthorizedEvent`. I'm thinking out loud here just a bit, but the important thing to remember is that upgrading to the latest shouldn't mean your events stop firing.

core/src/main/java/org/springframework/security/access/intercept/AbstractSecurityInterceptor.java
```diff
@@ -375,8 +379,25 @@ public boolean isValidateConfigAttributes() {
...
+		this.authorizationManager = new AuthorizationManagerAdapter<>(accessDecisionManager);
+	}
+
+	/**
+	 * Sets the {@link AuthorizationManager} instance.
+	 * @param authorizationManager the {@link AuthorizationManager} instance
+	 * @throws IllegalArgumentException if an {@link AccessDecisionManager} is already set
+	 */
+	public void setAuthorizationManager(AuthorizationManager<?> authorizationManager) {
+		Assert.isTrue(!(this.authorizationManager instanceof AuthorizationManagerAdapter),
```
jzheaux, line 398: In newer APIs Spring Security checks for null in the setter. Does that work here?

core/src/main/java/org/springframework/security/authorization/AuthorizationDecision.java
```diff
@@ -18,10 +18,21 @@
 
 /**
  * @author Rob Winch
+ * @author Evgeniy Cheban
  * @since 5.0
  */
 public class AuthorizationDecision {
```
jzheaux, line 24: A bit more thinking out loud, here. I wonder if `AuthorizationDecision` would benefit from a collection of reasons. Similar to `OAuth2TokenValidatorResult`, it's nice to be able to see why validation succeeded or failed.

The reason I'm thinking about this is that `AuthorizedEvent` and `AuthorizedFailureEvent` are both capable, through the config attributes, of saying why it succeeded or failed. It might be reasonable to add the `AuthorizationDecision` to the event.

core/src/test/java/org/springframework/security/access/intercept/aopalliance/MethodSecurityInterceptorTests.java
```diff
@@ -126,9 +127,9 @@ public void gettersReturnExpectedData() {
 	}
 
 	@Test
-	public void missingAccessDecisionManagerIsDetected() throws Exception {
-		this.interceptor.setAccessDecisionManager(null);
-		assertThatIllegalArgumentException().isThrownBy(() -> this.interceptor.afterPropertiesSet());
+	public void missingAccessDecisionManagerIsDetected() {
```
jzheaux, line 130: Should the tests also verify that a null authorization manager is detected?

core/src/test/java/org/springframework/security/authorization/AuthorityAuthorizationManagerTests.java
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * Tests for {@link AuthorityAuthorizationManager}.
+ *
+ * @author Evgeniy Cheban
+ */
+public class AuthorityAuthorizationManagerTests {
+
+	@Test
+	public void testWhenRoleNull() {
```
jzheaux, line 37: To better align with the team standard, would you mind renaming these methods to `methodUnderTestWhenConditionThenResult`?

For example, this method could be instead `hasRoleWhenNullThenException`.

web/src/main/java/org/springframework/security/web/access/DefaultWebInvocationPrivilegeEvaluator.java
```diff
@@ -91,14 +90,8 @@ public boolean isAllowed(String contextPath, String uri, String method, Authenti
 		if (authentication == null) {
 			return false;
 		}
-		try {
-			this.securityInterceptor.getAccessDecisionManager().decide(authentication, filterInvocation, attributes);
-			return true;
-		}
-		catch (AccessDeniedException ex) {
-			logger.debug(LogMessage.format("%s denied for %s", filterInvocation, authentication), ex);
```
jzheaux, line 99: Is it still possible to preserve the log here? Or, is there a reason that the log is no longer needed?

web/src/main/java/org/springframework/security/web/util/matcher/AntPathRequestMatcher.java
```diff
@@ -159,9 +160,14 @@ public boolean matches(HttpServletRequest request) {
 
 	@Override
 	public MatchResult matcher(HttpServletRequest request) {
-		if (this.matcher == null || !matches(request)) {
+		if (!matches(request)) {
 			return MatchResult.notMatch();
 		}
+
+		if (this.matcher == null) {
```
jzheaux, line 167: Changing the semantics in a minor release worries me a bit. Why is a null matcher a match now instead of not a match?

web/src/main/java/org/springframework/security/web/util/matcher/OrRequestMatcher.java
```diff
@@ -28,17 +28,18 @@
...
 
-	private final List<RequestMatcher> requestMatchers;
+	private final List<? extends RequestMatcher> requestMatchers;
 
 	/**
 	 * Creates a new instance
 	 * @param requestMatchers the {@link RequestMatcher} instances to try
 	 */
-	public OrRequestMatcher(List<RequestMatcher> requestMatchers) {
+	public OrRequestMatcher(List<? extends RequestMatcher> requestMatchers) {
```
jzheaux, line 42: I'd recommend trying to keep the matchers the same as they are today. It might mean registering more authorization manager instances in the configurer.

jzheaux (changes requested):
Thanks, @evgeniycheban, things are looking good. I've left some feedback inline.

This review was a little longer - note that GitHub compresses and hides some conversations, so remember to expand those when going through the review.

core/src/main/java/org/springframework/security/access/intercept/AbstractSecurityInterceptor.java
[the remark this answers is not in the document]
jzheaux: Agreed that it's worth considering. It doesn't change the fact that `AbstractSecurityInterceptor` needs to stop using `AccessDecisionManager`, but it would probably be worth it to file a ticket and start that conversation.

web/src/main/java/org/springframework/security/web/util/matcher/AntPathRequestMatcher.java
evgeniycheban: This matcher is only used to extract variables. If we have a `MATCH_ALL` pattern then `matcher` will be set to `null` and this method returns `MatchResult.notMatch` so I think this is a bug.

jzheaux: I see. Is the change necessary for this PR? If not, please create a separate ticket to get the bug addressed. This makes it easier to backport to earlier branches as necessary.

evgeniycheban: Yes, this is necessary, since I use `MatchResult` in `FilterInvocationAuthorizationManager` to get variables. Without this change, tests where the `MATCH_ALL` pattern is used will fail.

core/src/main/java/org/springframework/security/authorization/AuthorizationDecision.java
evgeniycheban: Agreed, but instead of adding the `AuthorizationDecision` to the event I wonder if it would be better to create a `ConfigAttribute` collection from a collection of reasons to create the event, so that user won't notice any changes after the update. What do you think?

evgeniycheban: The `AuthorizationDecision` can also be added to the event and the collection of `ConfigAttribute` can be marked as deprecated.

web/src/main/java/org/springframework/security/web/access/DefaultWebInvocationPrivilegeEvaluator.java
evgeniycheban: Yes, I'll use the `verify` method instead. I'm also wondering if we should now move the `null` checks for `authentication` and `attributes` to the `AuthorizationManagerAdapter`, since the `attributes` can be `null` when the `AuthorizationManager` is configured using the `access(AuthorizationManager)` method.

Checks like this are already done in `AbstractSecurityInterceptor`, it might be worth combining them in `AuthorizationManagerAdapter`.

evgeniycheban:
@jzheaux I've updated the PR according to our conversation.

core/src/test/java/org/springframework/security/access/intercept/aopalliance/MethodSecurityInterceptorTests.java
evgeniycheban: These tests are in `AbstractSecurityInterceptorTests`.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/AbstractInterceptUrlConfigurer.java
```diff
@@ -98,62 +101,62 @@ public void configure(H http) throws Exception {
...
+				AuthorizationManager<FilterInvocationAuthorizationContext> authorizationManager) {
+			for (RequestMatcher requestMatcher : requestMatchers) {
+				addMapping(requestMatcher, authorizationManager);
+			}
+			resetUnmappedMatchers();
+		}
+
+		final void addMapping(RequestMatcher requestMatcher,
+				AuthorizationManager<FilterInvocationAuthorizationContext> authorizationManager) {
+			this.managerBuilder.add(new RequestMatcherEntry<>(requestMatcher, authorizationManager));
```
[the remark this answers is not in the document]
jzheaux: One option might be to change `WebExpressionAuthorizationManager`'s `ExpressionHandler` into a setter so that it can be post-processed by `createAuthorizationManager`. `WebExpressionAuthorizationManager` would use `DefaultWebSecurityExpressionHandler` by default, but would be updated by `createAuthorizationManager`.

Then, while there is still parallel registration, at least the user's expression handler configuration doesn't get missed and the authorization manager would get all the rules in the right order.

Another option feels pretty hacky, but it would be to introduce a private class called `AuthorizationManagerConfigAttribute` that holds onto both items. Then this registry could hold onto config attributes and authorization managers in the same map. This has one minor benefit of removing the `resetUnmappedMatchers` method.

core/src/main/java/org/springframework/security/authorization/AuthorizationDecision.java
jzheaux: Yes, I like this idea. For now, let's try and leave the events unchanged, and then those things that are firing the events can produce a collection of `ConfigAttribute`s.

If that works well, you might consider re-consolidating the logging and event-throwing logic back into `AbstractSecurityInterceptor`. One reason this would be nice is it would prevent having to have the `publishAuthorizationSuccess` configuration in multiple places.

jzheaux (changes requested):
Thanks for your patience with this round of feedback, @evgeniycheban. I got behind after some recent PTO, I look forward to the next iteration!

web/src/main/java/org/springframework/security/web/access/DefaultWebInvocationPrivilegeEvaluator.java
jzheaux: I see that you changed to the `verify` method, as you explained. I'm wondering if it would still work to use `check` and then verify if the response `isDenied`? It's nice to avoid creating exceptions where we can since it's expensive.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/AbstractInterceptUrlConfigurer.java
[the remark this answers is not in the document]
evgeniycheban: In the first option, the expression should be parsed only after the `expressionHandler` is set, since the user can use a custom `expressionParser` in the `expressionHandler`, this seems to complicate the initialization of the `WebExpressionAuthorizationManager` and I'd like to avoid using lifecycle hooks such as `afterPropertiesSet` in this case, so I think the second option is better.

evgeniycheban:
@jzheaux I've updated the PR, please take a look at the changes in [45f9cdb](https://github.com/spring-projects/spring-security/pull/8996/commits/45f9cdbda9e68d9188e8fc5125457b4e6acec585)

web/src/main/java/org/springframework/security/web/access/DefaultWebInvocationPrivilegeEvaluator.java
evgeniycheban: Sorry, I misunderstood you about logging, I thought that the exception should still be visible in the logs.

evgeniycheban: @jzheaux It might be worth considering adding a default method 
`boolean isGranted(Supplier<Authentication> authentication, T object)` to the `AuthorizationManager`. What do you think?

jzheaux:
@evgeniycheban, thanks for all the work you've put into this.

I've been thinking more about what you said here:

> It might make sense to consider introducing a new filter that will use an AuthorizationManager similar to AuthorizationWebFilter.

And I had a chat with @rwinch this morning about the same thing.

As you've likely felt as well, this PR is needing to jump through a lot of hoops to make sure that it's a passive upgrade for applications that are customizing low-level classes like `SecurityMetadataSource` and `ConfigAttribute`.

It would be nice to simplify that. What would you think of adding a new DSL method instead that configures a new filter, instead of trying to adapt the existing DSL and `FilterSecurityInterceptor`? 

In this way, applications that are doing low-level customization of `FilterSecurityInterceptor` will still work. The new filter would not need to accommodate `SecurityMetadataSource`, `AccessDecisionManager`, `ConfigAttribute`, etc. It would also allow us to consider casting a wider deprecation net in a future PR.

The new DSL might look something like this:

```java
http
    .authorizeHttpRequests((requests) -> requests
        .mvcMatchers // ...
    )
```

`authorizeHttpRequests` would use a new configurer and construct and register an `AuthorizationFilter`, similar to `AuthorizationWebFilter`. The configurer would look similar to `ServerHttpSecurity.AuthorizeExchangeSpec` (for example, it wouldn't use the support for unmapped matchers in `AbstractConfigAttributeRequestMatcherRegistry`).

Also, in this case, we'd create separate tickets to look at events in `AuthorizationFilter` and reasons in `AuthorizationDecision`. We'd also deprecate `AccessDecisionManager` at a later time.

evgeniycheban:
> @evgeniycheban, thanks for all the work you've put into this.
> 
> I've been thinking more about what you said here:
> 
> > It might make sense to consider introducing a new filter that will use an AuthorizationManager similar to AuthorizationWebFilter.
> 
> And I had a chat with @rwinch this morning about the same thing.
> 
> As you've likely felt as well, this PR is needing to jump through a lot of hoops to make sure that it's a passive upgrade for applications that are customizing low-level classes like `SecurityMetadataSource` and `ConfigAttribute`.
> 
> It would be nice to simplify that. What would you think of adding a new DSL method instead that configures a new filter, instead of trying to adapt the existing DSL and `FilterSecurityInterceptor`?
> 
> In this way, applications that are doing low-level customization of `FilterSecurityInterceptor` will still work. The new filter would not need to accommodate `SecurityMetadataSource`, `AccessDecisionManager`, `ConfigAttribute`, etc. It would also allow us to consider casting a wider deprecation net in a future PR.
> 
> The new DSL might look something like this:
> 
> ```java
> http
>     .authorizeHttpRequests((requests) -> requests
>         .mvcMatchers // ...
>     )
> ```
> 
> `authorizeHttpRequests` would use a new configurer and construct and register an `AuthorizationFilter`, similar to `AuthorizationWebFilter`. The configurer would look similar to `ServerHttpSecurity.AuthorizeExchangeSpec` (for example, it wouldn't use the support for unmapped matchers in `AbstractConfigAttributeRequestMatcherRegistry`).
> 
> Also, in this case, we'd create separate tickets to look at events in `AuthorizationFilter` and reasons in `AuthorizationDecision`. We'd also deprecate `AccessDecisionManager` at a later time.

I like the idea of adding a new DSL, thanks for the feedback.

jzheaux:
@evgeniycheban glad to hear that you like it. Are you able to update this PR to use the proposed DSL method?

evgeniycheban:
@jzheaux sure, I've been busy at work these days, I'll update the PR this week.

evgeniycheban:
@jzheaux I've updated the PR, please take a look.

web/src/main/java/org/springframework/security/web/access/intercept/DelegatingAuthorizationManager.java
```diff
@@ -0,0 +1,121 @@
...
+			if (matchResult.isMatch()) {
+				AuthorizationManager<RequestAuthorizationContext> manager = mapping.getEntry();
+				if (this.logger.isTraceEnabled()) {
+					this.logger.trace(LogMessage.format("Checking authorization on %s using %s", request, manager));
+				}
+				return manager.check(authentication,
+						new RequestAuthorizationContext(request, matchResult.getVariables()));
+			}
+		}
+		this.logger.trace("Did not find RequestMatcher that match the request returned null");
```
jzheaux, line 82: Let's rephrase to:

```suggestion
		this.logger.trace("Abstaining since did not find matching RequestMatcher")
```

config/src/main/java/org/springframework/security/config/annotation/web/configurers/AuthorizeHttpRequestsConfigurer.java
```diff
@@ -0,0 +1,286 @@
...
+			setApplicationContext(context);
+		}
+
+		private void addMapping(RequestMatcher matcher, AuthorizationManager<RequestAuthorizationContext> manager) {
+			this.unmappedMatchers = null;
+			this.managerBuilder.add(new RequestMatcherEntry<>(matcher, manager));
+		}
+
+		private AuthorizationManager<HttpServletRequest> createAuthorizationManager() {
+			Assert.state(this.unmappedMatchers == null,
```
jzheaux, line 107: While the `DelegatingAuthorizationManager` constructor will make sure there is at least one mapping, it might be helpful to the end user to see a more contextual error here like:

```
At least one mapping is required (for example, authorizeHttpRequests().anyRequest().authenticated())
```

This is similar to the error message thrown with `authorizeRequests`.

jzheaux (changes requested):
Thanks, @evgeniycheban! I think this change made for a more straightforward PR.

I've left some feedback inline.

evgeniycheban:
@jzheaux I've made changes according to the review comments, please take a look.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/AuthorizeHttpRequestsConfigurer.java
evgeniycheban: I changed the message in the `DelegatingAuthorizationManager`'s constructor.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/AuthorizeHttpRequestsConfigurerTests.java
```diff
@@ -0,0 +1,618 @@
...
+		// @formatter:off
+		MockHttpServletRequestBuilder requestWithUser = get("/")
+				.with(user("user")
+				.roles("USER"));
+		// @formatter:on
+		this.mvc.perform(requestWithUser).andExpect(status().isOk());
+	}
+
+	@EnableWebSecurity
+	static class IncompleteMappingConfig extends WebSecurityConfigurerAdapter {
```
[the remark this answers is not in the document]
evgeniycheban: Done.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/AuthorizeHttpRequestsConfigurerTests.java
```diff
@@ -0,0 +1,618 @@
...
+	}
+
+	@EnableWebSecurity
+	static class AfterAnyRequestConfig extends WebSecurityConfigurerAdapter {
+
+		@Override
+		protected void configure(HttpSecurity http) throws Exception {
+			// @formatter:off
+			http
+				.authorizeHttpRequests((authorizeHttpRequests) ->
```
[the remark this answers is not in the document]
evgeniycheban: Done.

core/src/main/java/org/springframework/security/authorization/AuthorizationDecision.java
```diff
@@ -18,16 +18,35 @@
...
 
+	/**
+	 * Granted {@link AuthorizationDecision}.
+	 */
+	public static final AuthorizationDecision GRANTED = new AuthorizationDecision(true);
```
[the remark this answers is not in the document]
evgeniycheban: Done.

core/src/test/java/org/springframework/security/authorization/AuthenticatedAuthorizationManagerTests.java
```diff
@@ -0,0 +1,80 @@
...
+				AuthorityUtils.createAuthorityList("ROLE_ANONYMOUS"));
+		Object object = new Object();
+
+		assertThat(manager.check(authentication, object).isGranted()).isFalse();
+	}
+
+	@Test
+	public void authenticatedWhenUserNotAuthenticatedThenDeniedDecision() {
+		AuthenticatedAuthorizationManager<Object> manager = AuthenticatedAuthorizationManager.authenticated();
+		Supplier<Authentication> authentication = () -> {
```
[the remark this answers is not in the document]
evgeniycheban: Done.

web/src/main/java/org/springframework/security/web/access/intercept/DelegatingAuthorizationManager.java
```diff
@@ -0,0 +1,121 @@
...
+	public static final class Builder {
+
+		private final List<RequestMatcherEntry<AuthorizationManager<RequestAuthorizationContext>>> mappings = new ArrayList<>();
+
+		/**
+		 * Adds a {@link RequestMatcherEntry}.
+		 * @param entry the {@link RequestMatcher} to add
+		 * @return the {@link Builder} for further customizations
+		 */
+		public Builder add(RequestMatcherEntry<AuthorizationManager<RequestAuthorizationContext>> entry) {
```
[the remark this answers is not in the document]
evgeniycheban: Done.

web/src/main/java/org/springframework/security/web/access/intercept/DelegatingAuthorizationManager.java
evgeniycheban: Done.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/AuthorizeHttpRequestsConfigurer.java
jzheaux: This feels odd since if an application is using `DelegatingAuthorizationManager` independently from the DSL, then the error message will describe a remedy that doesn't apply.

evgeniycheban: @jzheaux I added validation to `createAuthorizationManager `

jzheaux:
Thanks, @evgeniycheban! Let's now have @rwinch take a look as I anticipate that he will also have some feedback.

jzheaux (approved):

rwinch (changes requested):
Thanks for the PR @evgeniycheban! I'm really happy how this turned out.

I think we should revisit https://github.com/spring-projects/spring-security/pull/8996#discussion_r506652742

If it is a bug, then I think we should create a separate ticket and commit with tests for it. This makes it easier for users to track changes (that it changed and why it changed). It also makes it easier for us to backport bug fixes.

core/src/main/java/org/springframework/security/authorization/AuthorizationManager.java
```diff
@@ -0,0 +1,67 @@
...
+	 * Determines if access should be granted for a specific authentication and object.
+	 * @param authentication the {@link Supplier} of the {@link Authentication} to check
+	 * @param object the {@link T} object to check
+	 * @return true if access is granted, false if denied
+	 */
+	default boolean isGranted(Supplier<Authentication> authentication, T object) {
+		AuthorizationDecision decision = check(authentication, object);
+		return decision == null || decision.isGranted();
+	}
+
```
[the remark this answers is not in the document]
evgeniycheban: This is an old diff, please see the last (62283faa) commit.

core/src/main/java/org/springframework/security/authorization/AuthorityAuthorizationManager.java
```diff
@@ -0,0 +1,75 @@
...
+ * evaluating if the {@link Authentication} contains a specified authority.
+ *
+ * @param <T> the type of object being authorized.
+ * @author Evgeniy Cheban
+ */
+public final class AuthorityAuthorizationManager<T> implements AuthorizationManager<T> {
+
+	private final Collection<String> authorities;
+
+	public AuthorityAuthorizationManager(Collection<String> authorities) {
```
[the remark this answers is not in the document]
evgeniycheban: This is an old diff, please see the last (62283fa) commit.

evgeniycheban:
@rwinch I reverted changes for `MvcRequestMatcher` and `AntPathRequestMatcher`.

evgeniycheban:
@rwinch Currently, `MvcRequestMatcher.matcher` doesn't check `HttpMethod` and `servletPath` like `MvcRequestMatcher.matches` does. Should I create a separate ticket for this or leave my changes in this PR?

jzheaux:
@evgeniycheban If the changes aren't required for this PR, then it's best to file a separate ticket, especially since the changes aren't related to the ticket that this PR is resolving.

The reason for that is it simplifies backporting bug fixes to other branches. Also, it makes release notes clearer for users that are affected by the bugs as they'll see a separate line item in the notes.

jzheaux:
@evgeniycheban, I added #9284 and #9285 to address `MvcRequestMatcher` and `AntPathRequestMatcher`, respectively.

jzheaux:
Thanks for all your work on this PR, @evgeniycheban! It's very exciting.

In addition to the above two tasks, I've created #9286, #9287, and #9288 for other improvements that we identified during the PR.

PR #9013, opened 2020-09-11
Use AssertJ for exception testing

philwebb:
This PR contains three commits that migrate to AssertJ for exception testing. This should help with an eventual migration to JUnit 5.

### DIFF
--- a/acl/src/test/java/org/springframework/security/acls/domain/AclImplTests.java
+++ b/acl/src/test/java/org/springframework/security/acls/domain/AclImplTests.java
@@ -166,7 +166,7 @@ public class AclImplTests {
 		assertThat(acl.getEntries().get(2).getSid()).isEqualTo(new GrantedAuthoritySid("ROLE_TEST2"));
 	}
 
-	@Test(expected = NotFoundException.class)
+	@Test
 	public void insertAceFailsForNonExistentElement() {
 		MutableAcl acl = new AclImpl(this.objectIdentity, 1, this.authzStrategy, this.pgs, null, null, true,
 				new PrincipalSid("joe"));
@@ -174,7 +174,8 @@ public class AclImplTests {
 		// Insert one permission
 		acl.insertAce(0, BasePermission.READ, new GrantedAuthoritySid("ROLE_TEST1"), true);
 		service.updateAcl(acl);
-		acl.insertAce(55, BasePermission.READ, new GrantedAuthoritySid("ROLE_TEST2"), true);
+		assertThatExceptionOfType(NotFoundException.class)
+				.isThrownBy(() -> acl.insertAce(55, BasePermission.READ, new GrantedAuthoritySid("ROLE_TEST2"), true));
 	}
 
 	@Test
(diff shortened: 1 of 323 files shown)


core/src/test/java/org/springframework/security/core/authority/mapping/SimpleAuthoritiesMapperTests.java
```diff
@@ -25,18 +25,19 @@
...
 public class SimpleAuthoritiesMapperTests {
 
-	@Test(expected = IllegalArgumentException.class)
+	@Test
 	public void rejectsInvalidCaseConversionFlags() throws Exception {
 		SimpleAuthorityMapper mapper = new SimpleAuthorityMapper();
 		mapper.setConvertToLowerCase(true);
 		mapper.setConvertToUpperCase(true);
-		mapper.afterPropertiesSet();
+		assertThatIllegalArgumentException().isThrownBy(mapper::afterPropertiesSet);
```
jzheaux, line 40: Since this method no longer throws an Exception, should that be cleaned up? There may be other methods to which this clean up might also need applying.

ldap/src/test/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProviderTests.java
```diff
@@ -250,100 +239,72 @@ public void passwordNeedsResetIsCorrectlyMapped() {
...
-		catch (InternalAuthenticationServiceException ex) {
-			// Since GH-8418 ldap communication exception is wrapped into
-			// InternalAuthenticationServiceException.
-			// This test is about the wrapped exception, so we throw it.
-			throw ex.getCause();
-		}
+		}).withCauseInstanceOf(org.springframework.ldap.CommunicationException.class);
 	}
 
-	@Test(expected = org.springframework.security.authentication.InternalAuthenticationServiceException.class)
```
jzheaux, line 340: Should this instead be replaced with `@Test`?

web/src/test/java/org/springframework/security/web/server/authentication/SwitchUserWebFilterTests.java
```diff
@@ -183,11 +179,12 @@ public void switchUserWhenUsernameIsMissingThenThrowException() {
 				.from(MockServerHttpRequest.post("/login/impersonate"));
 		final WebFilterChain chain = mock(WebFilterChain.class);
 		final SecurityContextImpl securityContext = new SecurityContextImpl(mock(Authentication.class));
-		this.exceptionRule.expect(IllegalArgumentException.class);
-		this.exceptionRule.expectMessage("The userName can not be null.");
-		this.switchUserWebFilter.filter(exchange, chain)
-				.subscriberContext(ReactiveSecurityContextHolder.withSecurityContext(Mono.just(securityContext)))
-				.block();
+		assertThatIllegalArgumentException()
+				.isThrownBy(() -> this.switchUserWebFilter.filter(exchange, chain)
```
jzheaux, line 183: This is another spot where I'm confused about the line wrapping. Compare this to line 220 and 221 where the content is identical, but separated across two lines.

jzheaux (changes requested):
Thanks for the PR, @philwebb! I've left some feedback inline.

acl/src/test/java/org/springframework/security/acls/jdbc/JdbcMutableAclServiceTests.java
```diff
@@ -161,91 +162,80 @@ public void testLifecycle() {
...
 				.readAclsById(Arrays.asList(getTopParentOid(), getMiddleParentOid(), getChildOid()));
 		assertThat(map).hasSize(3);
-		// Replace our current objects with their retrieved versions
-		topParent = (MutableAcl) map.get(getTopParentOid());
-		middleParent = (MutableAcl) map.get(getMiddleParentOid());
-		child = (MutableAcl) map.get(getChildOid());
+		// Get the retrieved versions
+		MutableAcl retrievedTopParent = (MutableAcl) map.get(getTopParentOid());
+		MutableAcl retrievedMiddleParent = (MutableAcl) map.get(getMiddleParentOid());
+		MutableAcl retrievedChild = (MutableAcl) map.get(getChildOid());
```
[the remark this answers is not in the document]
philwebb: The use of lambdas means that the variables need to be effectively final. The original version reassigned variables which meant compilation failed when they were referenced in the lambda.

acl/src/test/java/org/springframework/security/acls/domain/AclImplTests.java
```diff
@@ -97,58 +98,38 @@ public void tearDown() {
...
+	@Test
 	public void constructorsRejectNullObjectIdentity() {
-		try {
-			new AclImpl(null, 1, this.authzStrategy, this.pgs, null, null, true, new PrincipalSid("joe"));
-			fail("Should have thrown IllegalArgumentException");
-		}
-		catch (IllegalArgumentException expected) {
-		}
-		new AclImpl(null, 1, this.authzStrategy, this.mockAuditLogger);
+		assertThatIllegalArgumentException().isThrownBy(
```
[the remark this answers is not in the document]
philwebb: The plugin uses the Formatter logic from Eclipse. I'm not aware of any documentation about when it decides to split lines. In this example, the length of constructor seems to be causing the difference. I can wrap those blocks in // no-format if you like?

ldap/src/test/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProviderTests.java
philwebb: Yes, I must have accidentally deleted it. I'll push a fix then we can rebased when all the other issues are addressed.

web/src/test/java/org/springframework/security/web/server/authentication/SwitchUserWebFilterTests.java
philwebb: This is related to the length of the contained line. Extracting the holder to a variable helps quite a bit. I'll push a fix.

philwebb:
@jzheaux I've pushed https://github.com/spring-projects/spring-security/pull/9013/commits/4f381c0d5a2274fd89324264279005a12a220830 to address some of the comments. When we're 100% happy I can rebased them properly.

jzheaux (approved):
Thanks for the changes, @philwebb.

Thank you for cleaning up the stray `throws Exception` clause. There are others, though it appears that some are obsoleted by your changes, while some were stray before your changes. Instead of going through each one, I'll introduce a separate commit for removing the rest.

Issue #9018, opened 2020-09-15
[Question} Why CookieServerCsrfTokenRepository does not use Schedulers.boundedElastic()?


tt4g:
**Describe the bug**

Call `UUID#randomUUID()` in `CookieServerCsrfTokenRepository#createNewToken()` and `UUID#randomUUID()` is blocking I/O operation (reports #8128).
However, `CookieServerCsrfTokenRepository#generateToken(ServerWebExchange)` is not change Scheduler.

https://github.com/spring-projects/spring-security/blob/57c5ec26e7af1a4c2b8da894840fe55bf6897b9f/web/src/main/java/org/springframework/security/web/server/csrf/CookieServerCsrfTokenRepository.java#L68-L71

https://github.com/spring-projects/spring-security/blob/57c5ec26e7af1a4c2b8da894840fe55bf6897b9f/web/src/main/java/org/springframework/security/web/server/csrf/CookieServerCsrfTokenRepository.java#L148-L150

https://github.com/spring-projects/spring-security/blob/57c5ec26e7af1a4c2b8da894840fe55bf6897b9f/web/src/main/java/org/springframework/security/web/server/csrf/CookieServerCsrfTokenRepository.java#L156-L158

**To Reproduce**

`CookieServerCsrfTokenRepository#generateToken(ServerWebExchange)`

**Expected behavior**

Use `Schedulers.boundedElastic()`.

**Sample**

**N/A**

Is it the design of not using the `Schedulers.boundedElastic()`?

jzheaux:
Hi, @tt4g, thanks for the report.

I believe that it makes sense to change the token generation to use `Schedulers.boundedElastic()` like so:

```java
Mono.fromCallable(this::createCsrfToken).subscribeOn(Schedulers.boundedElastic());
```

since `UUID.randomUUID` is blocking.

Would you be able to submit a PR to update the class?

tt4g:
I will submit PR in a few days.

tt4g:
@jzheaux PR has been sent. Could you please confirm?

PR #9021, opened 2020-09-16
Use Schedulers.boundedElastic for UUID.randomUUID

tt4g:
The CSRF token is generated by UUID.randomUUID() which is I/O blocking operation.
This commit changes the subscriber thread to the bounded elactic scheduler.

Fix gh-9018


jzheaux:
Thanks for the PR, @tt4g! Will you please add the ticket number to the end your commit message like so:

```
...the bounded elactic scheduler.

Closes gh-9018
```

tt4g:
@jzheaux Thank you for quick response. I'm update commit message.

jzheaux:
Thanks, @tt4g, for the quick responses! This is now merged into `master`.

Issue #9017, opened 2020-09-15
WebSessionServerCsrfTokenRepository#generateToken() don't use Schedulers.boundedElastic()


tt4g:
**Describe the bug**

in `master` branch, `Schedulers.boundedElastic()` is not used because the `Mono` chain is broken.

https://github.com/spring-projects/spring-security/blob/57c5ec26e7af1a4c2b8da894840fe55bf6897b9f/web/src/main/java/org/springframework/security/web/server/csrf/WebSessionServerCsrfTokenRepository.java#L53-L57

**To Reproduce**

Call `WebSessionServerCsrfTokenRepository#generateToken()`.

**Expected behavior**

Use `Schedulers.boundedElastic()` .

**Sample**

**N/A**

I think I need to revert back to 8866fa6fb065b206724f36d1c9ee1d6bee46a4bd:

https://github.com/spring-projects/spring-security/blob/8866fa6fb065b206724f36d1c9ee1d6bee46a4bd/web/src/main/java/org/springframework/security/web/server/csrf/WebSessionServerCsrfTokenRepository.java#L53-L55

PR #9022, opened 2020-09-16
Fix broken Mono chain

tt4g:
This commit restore broken Mono chain in WebSessionServerCsrfTokenRepository.generateToken(ServerWebExchange).

Closes gh-9017


jzheaux:
Thanks for the PR, @tt4g! This is now merged into `master`.

PR #9039, opened 2020-09-22
support configurable sign algorithm

amergey:

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
@@ -18,6 +18,8 @@ package org.springframework.security.saml2.provider.service.registration;
 
 import java.security.PrivateKey;
 import java.security.cert.X509Certificate;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
@@ -26,6 +28,8 @@ import java.util.List;
 import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.Function;
+
+import org.opensaml.xmlsec.signature.support.SignatureConstants;
 
 import org.springframework.security.saml2.core.Saml2X509Credential;
 import org.springframework.util.Assert;
@@ -438,10 +442,12 @@ public final class RelyingPartyRegistration {
 
 		private final Saml2MessageBinding singleSignOnServiceBinding;
 
+		private List<String> signingMethodAlgorithms;
+
 		private AssertingPartyDetails(String entityId, boolean wantAuthnRequestsSigned,
 				Collection<Saml2X509Credential> verificationX509Credentials,
 				Collection<Saml2X509Credential> encryptionX509Credentials, String singleSignOnServiceLocation,
-				Saml2MessageBinding singleSignOnServiceBinding) {
+				Saml2MessageBinding singleSignOnServiceBinding, List<String> signingMethodAlgorithms) {
 			Assert.hasText(entityId, "entityId cannot be null or empty");
 			Assert.notNull(verificationX509Credentials, "verificationX509Credentials cannot be null");
 			for (Saml2X509Credential credential : verificationX509Credentials) {
@@ -457,12 +463,14 @@ public final class RelyingPartyRegistration {
 			}
 			Assert.notNull(singleSignOnServiceLocation, "singleSignOnServiceLocation cannot be null");
 			Assert.notNull(singleSignOnServiceBinding, "singleSignOnServiceBinding cannot be null");
+			Assert.notEmpty(signingMethodAlgorithms, "signingMethodAlgorithms cannot be empty");
 			this.entityId = entityId;
 			this.wantAuthnRequestsSigned = wantAuthnRequestsSigned;
 			this.verificationX509Credentials = verificationX509Credentials;
 			this.encryptionX509Credentials = encryptionX509Credentials;
 			this.singleSignOnServiceLocation = singleSignOnServiceLocation;
 			this.singleSignOnServiceBinding = singleSignOnServiceBinding;
+			this.signingMethodAlgorithms = signingMethodAlgorithms;
 		}
 
 		/**
@@ -542,6 +550,15 @@ public final class RelyingPartyRegistration {
 			return this.singleSignOnServiceBinding;
 		}
 
+		/**
+		 * Return the list of preferred signature algorithm URIs, in preference order.
+		 * @return the list of signature algorithm URIs
+		 * @since 5.5
+		 */
+		public List<String> getSigningMethodAlgorithms() {
+			return this.signingMethodAlgorithms;
+		}
+
 		public static final class Builder {
 
 			private String entityId;
@@ -555,6 +572,8 @@ public final class RelyingPartyRegistration {
 			private String singleSignOnServiceLocation;
 
 			private Saml2MessageBinding singleSignOnServiceBinding = Saml2MessageBinding.REDIRECT;
+
+			private List<String> signingMethodAlgorithms = new ArrayList<>();
 
 			/**
 			 * Set the asserting party's <a href=
@@ -640,14 +659,30 @@ public final class RelyingPartyRegistration {
 			}
 
 			/**
+			 * Apply this {@link Consumer} to the list of signature algorithm URIs
+			 * @param signingMethodAlgorithmsConsumer a {@link Consumer} of the list of
+			 * signature algorithm URIs
+			 * @return this {@code Builder}
+			 * @since 5.5
+			 */
+			public Builder signingMethodAlgorithms(Consumer<List<String>> signingMethodAlgorithmsConsumer) {
+				signingMethodAlgorithmsConsumer.accept(this.signingMethodAlgorithms);
+				return this;
+			}
+
+			/**
 			 * Creates an immutable ProviderDetails object representing the configuration
 			 * for an Identity Provider, IDP
 			 * @return immutable ProviderDetails object
 			 */
 			public AssertingPartyDetails build() {
+				List<String> signingMethodAlgorithmsCopy = this.signingMethodAlgorithms.isEmpty()
+						? Arrays.asList(SignatureConstants.ALGO_ID_SIGNATURE_RSA_SHA256)
+						: Collections.unmodifiableList(this.signingMethodAlgorithms);
+
 				return new AssertingPartyDetails(this.entityId, this.wantAuthnRequestsSigned,
 						this.verificationX509Credentials, this.encryptionX509Credentials,
-						this.singleSignOnServiceLocation, this.singleSignOnServiceBinding);
+						this.singleSignOnServiceLocation, this.singleSignOnServiceBinding, signingMethodAlgorithmsCopy);
 			}
 
 		}
--- a/saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProviderTests.java
+++ b/saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProviderTests.java
@@ -52,6 +52,7 @@ import org.opensaml.saml.saml2.core.impl.EncryptedAssertionBuilder;
 import org.opensaml.saml.saml2.core.impl.EncryptedIDBuilder;
 import org.opensaml.saml.saml2.core.impl.NameIDBuilder;
 import org.opensaml.xmlsec.encryption.impl.EncryptedDataBuilder;
+import org.opensaml.xmlsec.signature.support.SignatureConstants;
 import org.w3c.dom.Element;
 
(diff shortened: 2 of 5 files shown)


jzheaux:
@amergey, my apologies, but I think I led you down the wrong path here by placing the algorithm in `Saml2AuthenticationRequestContext`.

Thinking about this further, the selected algorithm must be one that the Asserting Party can support. The place to look that up is `RelyingPartyRegistration.AssertingPartyDetails`.

What we need to do, then, is add a private `List<String> signingMethods` field to `RelyingPartyRegistration.AssertingPartyDetails` and then use that information to resolve the `SigningSignatureParameters` in `OpenSamlAuthenticationRequestFactory` instead of getting the setting from `Saml2AuthenticationRequestContext`.

Would you be able to make those changes to the PR? If not, I'd be happy to add a polish to apply them.

amergey:
@jzheaux I have updated the PR according to your comments

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
```diff
@@ -438,10 +440,12 @@ private static Saml2X509Credential fromDeprecated(
 
 		private final Saml2MessageBinding singleSignOnServiceBinding;
 
+		private String signAlgorithmUri;
```
jzheaux, line 443: So that this aligns better with `<IDPSSODescriptor>`, will you please change this to be a `Collection<String>`? The reason is that an asserting party can specify more than one algorithm. Also, will you please call it `signingAlgorithms` since that's closer to what it's called [in the `<Extensions>` element.](https://www.oasis-open.org/committees/download.php/51890/SAML%20MD%20simplified%20overview.pdf).

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
```diff
@@ -542,6 +548,15 @@ public Saml2MessageBinding getSingleSignOnServiceBinding() {
 			return this.singleSignOnServiceBinding;
 		}
 
+		/**
+		 * Return the algorithm URI used to sign XML document
```
jzheaux, line 552: It would be nice if the JavaDoc here explained how this value relates to the asserting party's `<IDPSSODescriptor>`.

saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProviderTests.java
```diff
@@ -134,7 +135,7 @@ public void authenticateWhenInvalidDestinationThenThrowAuthenticationException()
 		Response response = TestOpenSamlObjects.response(DESTINATION + "invalid", ASSERTING_PARTY_ENTITY_ID);
 		response.getAssertions().add(TestOpenSamlObjects.assertion());
 		TestOpenSamlObjects.signed(response, TestSaml2X509Credentials.assertingPartySigningCredential(),
-				RELYING_PARTY_ENTITY_ID);
+				RELYING_PARTY_ENTITY_ID, SignatureConstants.ALGO_ID_SIGNATURE_RSA_SHA256);
```
jzheaux, line 138: Ideally, these tests that are unrelated to your change would not also be changed. Would it be better to introduce a new method into `TestOpenSamlObjects`?

jzheaux (changes requested):
Thanks for the updates, @amergey! I've left some feedback inline.

Also, will you please rebase? This file has changed since you began the PR.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
amergey: I do not get this, how can an asserting party have more than one algorithm ?

amergey:
I have rebased the PR, there is a test failure since, but I do not see how it is related to my changes

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
jzheaux: An asserting party can support more than one signing method, listing them in order of preference in their `<IDPSSODescriptor>`. The reliant party should use the highest one on the list that they can.

For example, an asserting party can say "relying parties can either sign with RSA-SHA256 or ECDSA-SHA256". So, if a relying party can use RSA-SHA256, it should, since that was listed first.

OpenSAML ships with support for this negotiation step which you can invoke by setting all the configured algorithms in `BasicSignatureSigningConfiguration#setSignatureAlgorithms`.

amergey: got it, thanks

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
```diff
@@ -542,6 +549,15 @@ public Saml2MessageBinding getSingleSignOnServiceBinding() {
 			return this.singleSignOnServiceBinding;
 		}
 
+		/**
+		 * Return the list of preferred signature algorithm URIs, in preference order.
+		 * @return the list of signature algorithm URIs
+		 * @since 5.5
+		 */
+		public List<String> getSignAlgorithmUris() {
```
jzheaux, line 557: I realize that OpenSAML calls this a signing algorithm, but since the metadata element looks like:

```xml
<alg:SigningMethod MinKeySize="256" Algorithm="http://www.w3.org/2001/04/xmldsigmore#ecdsa-sha256"/>
```

it would be more appropriate to name this `getSigningMethodAlgorithms`.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
```diff
@@ -639,6 +657,17 @@ public Builder singleSignOnServiceBinding(Saml2MessageBinding singleSignOnServic
 				return this;
 			}
 
+			/**
+			 * Set the signature algorithms to use when signing, in preference order.
+			 * @param signAlgorithmUris list of signature algorithm URI
+			 * @return this {@code Builder}
+			 * @since 5.5
+			 */
+			public Builder signAlgorithmUris(List<String> signAlgorithmUris) {
```
jzheaux, line 666: For consistency with the rest of the `Builder`, let's have this take a `Consumer<Collection<String>>` instead of a `List<String>`. You can see the `signingX509Credentials` method for an example.

The reason that using a `Consumer` is nice is it makes it simpler to return an instance of `Builder` in an API where an application can add algorithms to the existing list.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
```diff
@@ -556,6 +572,8 @@ public Saml2MessageBinding getSingleSignOnServiceBinding() {
 
 			private Saml2MessageBinding singleSignOnServiceBinding = Saml2MessageBinding.REDIRECT;
 
+			private List<String> signAlgorithmUris = Arrays.asList(SignatureConstants.ALGO_ID_SIGNATURE_RSA_SHA256);
```
jzheaux, line 575: I think it would probably be best to leave this empty from the start and then add it in the `build` method if it's still empty.

The reason for that is that this will likely be called by `OpenSamlAssertingPartyMetadataConverter`, which will take the list of signing algorithms from the asserting party's metadata and do something like:

```java
builder.assertingPartyDetails((details) -> details
    .signingMethodAlgorithms((a) -> a.addAll(algorithms))
);
```

If there's something already in the list, then it would need to also first clear the original list:

```java
builder.assertingPartyDetails((details) -> details
    .signingMethodAlgorithms((a) -> {
        a.clear();
        a.addAll(algorithms);
    })
);
```

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
```diff
@@ -438,10 +441,12 @@ private static Saml2X509Credential fromDeprecated(
 
 		private final Saml2MessageBinding singleSignOnServiceBinding;
 
+		private List<String> signingAlgorithmUris;
```
jzheaux, line 444: This should be a `Collection` since the `List`-specific methods are not needed.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
```diff
@@ -457,12 +462,14 @@ private AssertingPartyDetails(String entityId, boolean wantAuthnRequestsSigned,
 			}
 			Assert.notNull(singleSignOnServiceLocation, "singleSignOnServiceLocation cannot be null");
 			Assert.notNull(singleSignOnServiceBinding, "singleSignOnServiceBinding cannot be null");
+			Assert.notNull(signingAlgorithmUris, "signingAlgorithmUris cannot be null");
```
jzheaux, line 465: This should ensure that the list also isn't empty. You can achieve this by calling `Assert.notEmpty` instead. I think it would make sense to change the error message to align with that as well.

jzheaux (changes requested):
Thanks for the fast turnaround, @amergey! 

This is starting to take shape. I've left a bit more feedback inline.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
amergey: List specific are not needed, but the order matters, it the main reason I used list there

amergey:
@jzheaux  I have updated the PR according to your comments, the remaining point is related to collection for signing algorithm. I keep list because of the order, but If you still think collection is better, let me know, I will do the change

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
```diff
@@ -639,15 +657,32 @@ public Builder singleSignOnServiceBinding(Saml2MessageBinding singleSignOnServic
...
 			 * Creates an immutable ProviderDetails object representing the configuration
 			 * for an Identity Provider, IDP
 			 * @return immutable ProviderDetails object
 			 */
 			public AssertingPartyDetails build() {
+				if (this.signingMethodAlgorithms.isEmpty()) {
+					this.signingMethodAlgorithms.add(SignatureConstants.ALGO_ID_SIGNATURE_RSA_SHA256);
+				}
+				List<String> signingMethodAlgorithmsCopy = new ArrayList(this.signingMethodAlgorithms);
+				this.signingMethodAlgorithms.clear();
```
jzheaux, line 682: What is your reasoning for clearing the list out? `build` does not clear out the other lists or reset any of the other fields to their original state.

jzheaux (changes requested):
Thanks for the adjustments, @amergey. I've left one more inline comment.

In preparation for merging, will you please squash your commits and format the commit message to include the ticket number, like so:

```
Commit Title

Closes gh-8952
```

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
jzheaux: I see, fair point. Let's leave it as-is.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
amergey: Good point, It was to solve the initialization issue where default algo was always there because of the fact that a default AssertingPartyDetails is created with RelyingPartyRegistration instance but after looking at this more carefully, it seems the wrong way to fix this, I will improve it

jzheaux:
Thanks, @amergey! This is now merged into `master`.

Note that I added a polish of e1826a0bd8eee7dc63267c28d4ce6f6ff294e7e8 that changes the name of the method since "method" and "algorithm" are synonymous. So `getSigningMethodAlgorithms` is now `getSigningAlgorithms`. I also added support to the `withRelyingPartyRegistration` method to make sure the algorithms get copied over. Finally, I reordered the methods so that they appear in the order that those items typically appear in role descriptors.

Also, I added documentation via b8f8fab54f508cf7ff4f12b63ba137bf23163059.

Issue #9030, opened 2020-09-17
Include Port in DNS SRV type lookups


KathrynN:
**Expected Behavior**

JndiDnsResolver should be all I need to discover a service given the domain and service-type.

**Current Behavior**

JndiDnsResolver only gives me the service hostname / ip address, but not which port.

**Context**

I'm building a system using SRV DNS service discovery with randomly assigned ports (to allow multiple services to be able to stack on one server).  This non-static port means that I have to rewrite the JndiDnsResolver myself to capture the port, even though the JndiDnsResolver has the method to do it.

I think it would be a quick fix, and am happy to share what I've done, which will add an additional method to the JndiDnsResolver that also captures the port without changing the result for older methods.

Or, if anyone knows if this is already implemented and I'm using the wrong version of spring boot security, let me know that too! 👍 
```
        <dependency>
            <groupId>org.springframework.security</groupId>
            <artifactId>spring-security-remoting</artifactId>
            <version>5.3.4.RELEASE</version>
        </dependency>
```

eleftherias:
@KathrynN That makes sense to me.

Are you interested in submitting a pull request with your changes?

KathrynN:
Absolutely!  I will send it up over the weekend (I'm not feeling very well today).  Thanks!

KathrynN:
I have popped in a pull request (as you can see above!) but I can't build the master branch on my machine, let alone run the tests.  I will try to have a go over next week when I have recovered enough to wrangle gradle into submission.

PR #9052, opened 2020-09-26
Added method to return both ip and port for srv dns lookup requests.

KathrynN:
Closes gh-9030

### DIFF
--- a/remoting/src/main/java/org/springframework/security/remoting/dns/JndiDnsResolver.java
+++ b/remoting/src/main/java/org/springframework/security/remoting/dns/JndiDnsResolver.java
@@ -61,14 +61,26 @@ public class JndiDnsResolver implements DnsResolver {
 
 	@Override
 	public String resolveServiceEntry(String serviceType, String domain) {
-		return resolveServiceEntry(serviceType, domain, this.ctxFactory.getCtx());
+		return resolveServiceEntry(serviceType, domain, this.ctxFactory.getCtx()).getHostName();
 	}
 
 	@Override
 	public String resolveServiceIpAddress(String serviceType, String domain) {
 		DirContext ctx = this.ctxFactory.getCtx();
-		String hostname = resolveServiceEntry(serviceType, domain, ctx);
+		String hostname = resolveServiceEntry(serviceType, domain, ctx).getHostName();
 		return resolveIpAddress(hostname, ctx);
+	}
+
+	/**
+	 * @author Kathryn Newbould
+	 * @since 5.4.1
+	 * @return String of ip address and port, format [ip_address]:[port] of service if found
+	 * @throws DnsLookupException if not found
+	 */
+	public String resolveServiceIpAddressAndPort(String serviceType, String domain) {
+		DirContext ctx = this.ctxFactory.getCtx();
+		ConnectionInfo hostInfo = resolveServiceEntry(serviceType, domain, ctx);
+		return resolveIpAddress(hostInfo.getHostName(), ctx) + ":" + hostInfo.getPort();
 	}
 
 	// This method is needed, so that we can use only one DirContext for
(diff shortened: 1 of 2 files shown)


remoting/src/test/java/org/springframework/security/remoting/dns/JndiDnsResolverTests.java
```diff
@@ -92,6 +92,16 @@ public void testResolveServiceIpAddress() throws Exception {
 		assertThat(ipAddress).isEqualTo("63.246.7.80");
 	}
 
+	@Test
+	public void resolveServiceIpAddressWithPort() throws Exception {
```
KathrynN, line 96: NB - considered prefixing this with "test" to match format of rest of file, but after checking the guidelines it specifically says new tests should not prefix "test" so just let me know if I misread that.

eleftherias: You are correct not to prefix the test with "test". 
As we add more tests, we generally follow the format `methodWhenConditionThenResult`.
This one could be named something like `resolveServiceIpAddressAndPortWhenExistsThenReturnsIpAdressAndPort`.

remoting/src/main/java/org/springframework/security/remoting/dns/JndiDnsResolver.java
```diff
@@ -61,16 +61,28 @@ public String resolveIpAddress(String hostname) {
...
 	public String resolveServiceIpAddress(String serviceType, String domain) {
 		DirContext ctx = this.ctxFactory.getCtx();
-		String hostname = resolveServiceEntry(serviceType, domain, ctx);
+		String hostname = resolveServiceEntry(serviceType, domain, ctx).getHostName();
 		return resolveIpAddress(hostname, ctx);
 	}
 
+	/**
+	 * @author Kathryn Newbould
+	 * @since 5.4.1
```
eleftherias, line 76: This change will be in 5.5 , since 5.4.1 is a patch release and should only include bug fixes.
```suggestion
	 * @since 5.5
```

remoting/src/main/java/org/springframework/security/remoting/dns/JndiDnsResolver.java
```diff
@@ -61,16 +61,28 @@ public String resolveIpAddress(String hostname) {
...
 	@Override
 	public String resolveServiceIpAddress(String serviceType, String domain) {
 		DirContext ctx = this.ctxFactory.getCtx();
-		String hostname = resolveServiceEntry(serviceType, domain, ctx);
+		String hostname = resolveServiceEntry(serviceType, domain, ctx).getHostName();
 		return resolveIpAddress(hostname, ctx);
 	}
 
+	/**
+	 * @author Kathryn Newbould
```
eleftherias, line 75: We generally keep the `@author` tags on the class Javadoc. Could you move this up to the JndiDnsResolver Javadoc?

remoting/src/main/java/org/springframework/security/remoting/dns/JndiDnsResolver.java
```diff
@@ -107,15 +120,18 @@ private String resolveServiceEntry(String serviceType, String domain, DirContext
 				int priority = Integer.parseInt(record[0]);
 				int weight = Integer.parseInt(record[1]);
 				// we have a new highest Priority, so forget also the highest weight
+				int SERVICE_RECORD_PORT_INDEX = 2;
```
eleftherias, line 123: Since we're already using the integers directly for the priority, weight and hostname, I think we should continue that pattern here and use `record[2]` rather than setting a constant.

eleftherias (changes requested):
Thanks @KathrynN!

I have left a couple of comments inline. 

I also recommend rebasing on the latest version of master. There are some changes in c5023127190c8f80bf83bafabcbea4df5cf44898 that may cause conflicts.

eleftherias:
@KathrynN Have you had a chance to look at my feedback?

KathrynN:
Hi!  Yes, so sorry.  I've been rather unwell this last 2 months or so (almost co-inciding with my renewed interest in opensource... hmmmm) but I'm sort of alive again.  I'll try to get to this over the weekend.  Sorry again for just vanishing!

eleftherias:
I'm sorry to hear you've been unwell and I hope you're feeling better @KathrynN. 
Would you like to continue working on this PR or would you prefer if someone else took over?

eleftherias:
Merged via 2af322c06d63dae06690db73d8e295e982e96ce2 and polished via 488683f5a5658003b96ee7d01c9e5c952a3f1f64.

Issue #9044, opened 2020-09-24
Add HSM Support for Decrypting Assertions


jzheaux:
It would be nice to allow for a custom decryption strategy in `OpenSamlAuthenticationProvider`. This would simplify delegating the assertion and principal decryption to a separate service.

jzheaux:
Currently, `OpenSamlAuthenticationProvider` has a private field `Converter<Saml2AuthenticationToken, Decrypter>`. I'm not convinced that `Decrypter` is the correct thing to expose, though. I'd prefer that it be an interface. Since OpenSAML doesn't provide an interface for decryption, we could possibly use `Consumer`.

So, instead of:

```java
Decrypter decrypter = // ... your custom decrypter
provider.setDecrypter((token) -> decrypter);
```

you'd do:

```java
YourDecrypter decrypter = // ... your custom decrypter
provider.setResponseDecrypter((tuple) -> {
    EncryptedAssertion encrypted = tuple.getResponse().getEncryptedAssertions().get(0);
    Assertion assertion = decrypter.decrypt(encrypted);
    tuple.getResponse().getAssertions().add(assertion);
});
```

Or, for decrypting an assertion's subject, you'd do:

```java
YourDecrypter decrypter = // ... your custom decrypter
provider.setAssertionDecrypter((tuple) -> {
    EncryptedID encrypted = tuple.getAssertion().getSubject().getEncryptedID();
    NameID name = decrypter.decrypt(encrypted);
    tuple.getAssertion().getSubject().setNameID(name);
});
```

While a touch more verbose, it provides a great deal more flexibility and encourages composition over inheritance.

ryan13mt:
@jzheaux would something like this meet your explanation above? https://github.com/spring-projects/spring-security/pull/9055

jzheaux:
Thanks for the PR, @ryan13mt, I've left my feedback there.

ryan13mt:
Hello @jzheaux  thanks for closing this issue and merging to master.

I am finding a problem with the polishing commit. You added a check in d0581c9#diff-e1434c9a229d033aa3fe922ab7b9637a54ba5a2521711e3b362e70bc7f787833R511 where you are checking if a response is signed and if true, doing the elements decryption. Is there a reason this check was added?

[...]

PR #9055, opened 2020-09-28
Add HSM Support for Decrypting Assertions

ryan13mt:
To solve issue: https://github.com/spring-projects/spring-security/issues/9044

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProvider.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProvider.java
@@ -157,6 +157,7 @@ import org.springframework.util.StringUtils;
  * asserting party, IDP, verification certificates.
  * </p>
  *
+ * @author Ryan Cassar
  * @since 5.2
  * @see <a href=
  * "https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf#page=38">SAML 2
@@ -210,6 +211,32 @@ public final class OpenSamlAuthenticationProvider implements AuthenticationProvi
 	private Converter<Saml2AuthenticationToken, SignatureTrustEngine> signatureTrustEngineConverter = new SignatureTrustEngineConverter();
 
 	private Converter<Saml2AuthenticationToken, Decrypter> decrypterConverter = new DecrypterConverter();
+
+	private Consumer<ResponseToken> assertionDecrypter = (responseToken) -> {
+		List<Assertion> assertions = new ArrayList<>();
+		for (EncryptedAssertion encryptedAssertion : responseToken.getResponse().getEncryptedAssertions()) {
+			try {
+				Decrypter decrypter = this.decrypterConverter.convert(responseToken.getToken());
+				Assertion assertion = decrypter.decrypt(encryptedAssertion);
+				assertions.add(assertion);
+			}
+			catch (DecryptionException ex) {
+				throw createAuthenticationException(Saml2ErrorCodes.DECRYPTION_ERROR, ex.getMessage(), ex);
+			}
+		}
+		responseToken.getResponse().getAssertions().addAll(assertions);
+	};
+
+	private Consumer<ResponseToken> principalDecrypter = (responseToken) -> {
+		try {
+			Decrypter decrypter = this.decrypterConverter.convert(responseToken.getToken());
+			Assertion assertion = CollectionUtils.firstElement(responseToken.getResponse().getAssertions());
+			assertion.getSubject().setNameID((NameID) decrypter.decrypt(assertion.getSubject().getEncryptedID()));
+		}
+		catch (DecryptionException ex) {
+			throw createAuthenticationException(Saml2ErrorCodes.DECRYPTION_ERROR, ex.getMessage(), ex);
+		}
+	};
 
 	/**
 	 * Creates an {@link OpenSamlAuthenticationProvider}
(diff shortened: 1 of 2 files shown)


saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProvider.java
```diff
@@ -512,6 +566,12 @@ private Saml2ResponseValidatorResult validateResponse(Saml2AuthenticationToken t
 	}
 
 	private List<Assertion> decryptAssertions(Decrypter decrypter, Response response) {
+		Consumer<Response> assertionDecrypter = getAssertionDecrypter();
+		if (assertionDecrypter != null) {
```
jzheaux, line 570: Instead of checking for null, could we instead set a default value to `assertionDecrypter` that performs the default operation?

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProvider.java
```diff
@@ -332,6 +336,48 @@ public void setResponseTimeValidationSkew(Duration responseTimeValidationSkew) {
...
+	 *	OpenSamlAuthenticationProvider provider = new OpenSamlAuthenticationProvider();
+	 *	provider.setAssertionDecrypter((response) -> {
+	 *  	EncryptedAssertion encrypted = response.getEncryptedAssertions().get(0);
+	 *  	Assertion assertion = decrypter.decrypt(encrypted);
+	 *  	response.getAssertions().add(assertion);
+	 *	});
+	 * </pre>
+	 * @param assertionDecrypter response consumer
+	 */
+	public void setAssertionDecrypter(Consumer<Response> assertionDecrypter) {
```
jzheaux, line 356: In the class, there is a tuple class called `ResponseToken` which has the deserialized SAML Response as well as the `Saml2AuthenticationToken`. It would be nice to have both of these when customizing decryption. The reason for that is if you are doing local decryption, you'll need to look up the keys from the `Saml2AuthenticationToken`.

Could we change this contract to `setAssertionDecrypter(Consumer<ResponseToken> assertionDecrypter)`?

jzheaux (changes requested):
Thanks, @ryan13mt, for another PR! I've left some feedback inline.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProvider.java
ryan13mt: I committed a new change to match with your suggestion. To do it i had to create a class variable for the Decrypter since it's now being used from the 2 new consumers.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProvider.java
ryan13mt: Agreed and done in the latest commit.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProvider.java
```diff
@@ -211,6 +211,32 @@
 
 	private Converter<Saml2AuthenticationToken, Decrypter> decrypterConverter = new DecrypterConverter();
 
+	private Decrypter decrypter;
```
jzheaux, line 214: This isn't thread-safe. Because the `decrypter` instance is changed with each authentication, two concurrent authentication attempts will clobber each other's value here.

Instead, the default values below will need to construct a `Decrypter` using the details from `Saml2AuthenticationToken`. For `assertionDecrypter`, I think this will look something like:

```java
Decrypter decrypter = this.decrypterConverter.convert(responseToken.getToken());
for (EncryptedAssertion encryptedAssertion : responseToken.getResponse().getEncryptedAssertions()) {
// ...
```

with something similar being done in `principalDecrypter`.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProvider.java
```diff
@@ -332,6 +358,50 @@ public void setResponseTimeValidationSkew(Duration responseTimeValidationSkew) {
...
+	 *		Response response = responseToken.getResponse();
+	 *  	EncryptedAssertion encrypted = response.getEncryptedAssertions().get(0);
+	 *  	Assertion assertion = decrypter.decrypt(encrypted);
+	 *  	response.getAssertions().add(assertion);
+	 *	});
+	 * </pre>
+	 * @param assertionDecrypter response token consumer
+	 */
+	public void setAssertionDecrypter(Consumer<ResponseToken> assertionDecrypter) {
+		this.assertionDecrypter = assertionDecrypter;
```
jzheaux, line 380: In these setters, will you please check for null using `Assert.notNull`?

jzheaux (changes requested):
Thanks for the updates, @ryan13mt! I like where this is headed; I've left some feedback inline.

Also, just a reminder to add tests for the setters. I believe it would suffice to use mock `Consumer` instances.

ryan13mt:
@jzheaux locally this is failing as well like the other pull request. Not really sure on what the reason is.

Execution failed for task ':spring-security-samples-boot-oauth2resourceserver-webflux:integrationTest'
Execution failed for task ':spring-security-samples-boot-oauth2webclient-webflux:test'

jzheaux (changes requested):
Thanks for the updates, @ryan13mt! In preparation for merging, will you please squash your commits? Also, if you like, please consider adding yourself to the list of authors.

ryan13mt:
Thanks @jzheaux hope i squashed my commits the right way. Is there anything else needed to close this PR please?

ryan13mt:
Hi again @jzheaux is there any other input needed from my end cause i am going to be reassigned into a new project and would have very limited time to contribute further to this and the other pull request https://github.com/spring-projects/spring-security/pull/9056#


Thanks and sorry for the spam 🤣

jzheaux:
Thanks for your contribution, @ryan13mt! This is now merged into `master` via 535ae3e27dc951e9aa5e1e0ff09d46be4e897eea as well as a polish commit of d0581c9a2600b5bea5f2e0cab8fe4c99c46b8fa3 to adjust the method names.

In a few minutes, it should be available in `5.5.0-SNAPSHOT` in https://repo.spring.io/libs-snapshot. Please try it out to make sure it meets your needs.

ryan13mt:
Hello @jzheaux thanks for closing this issue and merging to master. 

I am finding a problem with the polishing commit. You added a check in https://github.com/spring-projects/spring-security/commit/d0581c9a2600b5bea5f2e0cab8fe4c99c46b8fa3#diff-e1434c9a229d033aa3fe922ab7b9637a54ba5a2521711e3b362e70bc7f787833R511 where you are checking if a response is signed and if true, doing the elements decryption. Is there a reason this check was added?

I believe this is not correct since there are use cases where the idp will sign and encrypt the assertion but will leave the response unsigned as is in our case. Thus having an unsigned SAML Response with an encrypted signed Assertion should still decrypt the assertion and then checking that it has been signed before throwing the error `"Either the response or one of the assertions is unsigned. Please either sign the response or all of the assertions.".` or in my case it's throwing  `"No assertions found in response."` since the encrypted assertions have not being decrypted yet.

With this new check, the library is now requiring that a response must be signed if it contains an encrypted assertion.


Thanks

Issue #9028, opened 2020-09-17
File-based Configuration for Asserting Party Metadata


ryan13mt:
**Expected Behavior**

```
String metadataLocation = "C:\\local\\temp\\saml\\meta.xml";

RelyingPartyRegistration registration = RelyingPartyRegistrations.fromMetadataLocation(metadataLocation)
 .registrationId("registration-id").build();
```

**Current Behavior**

```
String metadataLocation = "https://testAssertingParty.com/meta.xml";

RelyingPartyRegistration registration = RelyingPartyRegistrations.fromMetadataLocation(metadataLocation)
 .registrationId("registration-id")
.build();
```

**Context**

Currently the `RelyingPartyRegistrations.fromMetadataLocation(metadataLocation)` only accepts a URL. Basing on my old implementation with **Spring Security SAML Extension** it would be good if we could also provide a local directory path to this method. Some asserting parties will not host their metadata files but will instead email it to the relying party.

**Suggested Solution**

The OpenSamlRelyingPartyRegistrationBuilderHttpMessageConverter class introduced in **5.4** can still be used but the _parsing logic_ should be extracted into a separate utility class that can be used elsewhere. 

An example of the changes needed in `RelyingPartyRegistrations.fromMetadataLocation(metadataLocation)`:

```
public static RelyingPartyRegistration.Builder fromMetadataLocation(String metadataLocation) {
	try {
		if (metadataLocation.matches("^(https?)://.*$")) {
			RestOperations rest = new RestTemplate(Arrays.asList(new OpenSamlRelyingPartyRegistrationBuilderHttpMessageConverter()));
			return rest.getForObject(metadataLocation, RelyingPartyRegistration.Builder.class);
		} else {
			OpenSamlRelyingPartyRegistrationBuilderConverter converter = new OpenSamlRelyingPartyRegistrationBuilderConverter();
			return converter.read(RelyingPartyRegistration.Builder.class, new FileInputStream(metadataLocation));
		}
	}
	catch (RestClientException | FileNotFoundException ex) {
		if (ex.getCause() instanceof Saml2Exception) {
			throw (Saml2Exception) ex.getCause();
		}
		throw new Saml2Exception(ex);
	}
}
```

This way if the metadataLocation is a url, it will use the OpenSamlRelyingPartyRegistrationBuilderHttpMessageConverter, if not, it will use the Utility class that holds the parsing code which in my example is named OpenSamlRelyingPartyRegistrationBuilderConverter.

[...]

PR #9056, opened 2020-09-28
Changed metadata converter to accept files as well

ryan13mt:
Pull request to close https://github.com/spring-projects/spring-security/issues/9028

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/OpenSamlAssertingPartyMetadataConverter.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/OpenSamlAssertingPartyMetadataConverter.java
@@ -1,0 +1,161 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.saml2.provider.service.registration;
+
+import java.io.InputStream;
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+import java.util.ArrayList;
+import java.util.List;
+
+import net.shibboleth.utilities.java.support.xml.ParserPool;
+import org.opensaml.core.config.ConfigurationService;
+import org.opensaml.core.xml.XMLObject;
+import org.opensaml.core.xml.config.XMLObjectProviderRegistry;
+import org.opensaml.core.xml.io.Unmarshaller;
+import org.opensaml.saml.common.xml.SAMLConstants;
+import org.opensaml.saml.saml2.metadata.EntitiesDescriptor;
+import org.opensaml.saml.saml2.metadata.EntityDescriptor;
+import org.opensaml.saml.saml2.metadata.IDPSSODescriptor;
+import org.opensaml.saml.saml2.metadata.KeyDescriptor;
+import org.opensaml.saml.saml2.metadata.SingleSignOnService;
+import org.opensaml.security.credential.UsageType;
+import org.opensaml.xmlsec.keyinfo.KeyInfoSupport;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+
+import org.springframework.security.saml2.Saml2Exception;
+import org.springframework.security.saml2.core.OpenSamlInitializationService;
+import org.springframework.security.saml2.core.Saml2X509Credential;
+
+class OpenSamlAssertingPartyMetadataConverter {
+
+	static {
+		OpenSamlInitializationService.initialize();
+	}
+
+	private final XMLObjectProviderRegistry registry;
+
+	private final ParserPool parserPool;
+
+	/**
+	 * Creates a {@link OpenSamlAssertingPartyMetadataConverter}
+	 */
+	OpenSamlAssertingPartyMetadataConverter() {
+		this.registry = ConfigurationService.get(XMLObjectProviderRegistry.class);
+		this.parserPool = this.registry.getParserPool();
+	}
+
+	RelyingPartyRegistration.Builder convert(InputStream inputStream) {
+		EntityDescriptor descriptor = entityDescriptor(inputStream);
+		IDPSSODescriptor idpssoDescriptor = descriptor.getIDPSSODescriptor(SAMLConstants.SAML20P_NS);
+		if (idpssoDescriptor == null) {
+			throw new Saml2Exception("Metadata response is missing the necessary IDPSSODescriptor element");
+		}
+		List<Saml2X509Credential> verification = new ArrayList<>();
+		List<Saml2X509Credential> encryption = new ArrayList<>();
+		for (KeyDescriptor keyDescriptor : idpssoDescriptor.getKeyDescriptors()) {
+			if (keyDescriptor.getUse().equals(UsageType.SIGNING)) {
+				List<X509Certificate> certificates = certificates(keyDescriptor);
+				for (X509Certificate certificate : certificates) {
+					verification.add(Saml2X509Credential.verification(certificate));
+				}
+			}
+			if (keyDescriptor.getUse().equals(UsageType.ENCRYPTION)) {
+				List<X509Certificate> certificates = certificates(keyDescriptor);
+				for (X509Certificate certificate : certificates) {
+					encryption.add(Saml2X509Credential.encryption(certificate));
+				}
+			}
+			if (keyDescriptor.getUse().equals(UsageType.UNSPECIFIED)) {
+				List<X509Certificate> certificates = certificates(keyDescriptor);
+				for (X509Certificate certificate : certificates) {
+					verification.add(Saml2X509Credential.verification(certificate));
+					encryption.add(Saml2X509Credential.encryption(certificate));
+				}
+			}
+		}
+		if (verification.isEmpty()) {
+			throw new Saml2Exception(
+					"Metadata response is missing verification certificates, necessary for verifying SAML assertions");
+		}
+		RelyingPartyRegistration.Builder builder = RelyingPartyRegistration.withRegistrationId(descriptor.getEntityID())
+				.assertingPartyDetails((party) -> party.entityId(descriptor.getEntityID())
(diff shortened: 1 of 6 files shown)


saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/OpenSamlAssertingPartyMetadataConverter.java
```diff
@@ -77,44 +67,28 @@
  * @author Josh Cummings
  * @since 5.4
  */
-public class OpenSamlRelyingPartyRegistrationBuilderHttpMessageConverter
-		implements HttpMessageConverter<RelyingPartyRegistration.Builder> {
+public class OpenSamlAssertingPartyMetadataConverter {
```
jzheaux, line 70: I think this should be two classes. One is something that implements `HttpMessageConverter` and one that just deals with the logic of converting an `InputStream` into a `RelyingPartyRegistration.Builder`. This existing class's public contract would remain as-is and the new class would be called `OpenSamlAssertingPartyMetadataConverter`. This existing class would call the new class.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/OpenSamlAssertingPartyMetadataConverter.java
```diff
@@ -171,50 +145,50 @@ else if (singleSignOnService.getBinding().equals(Saml2MessageBinding.REDIRECT.ge
...
+				}
+			}
 		}
+		return encryption;
+	}
+
+	private List<X509Certificate> certificates(KeyDescriptor keyDescriptor) {
 		try {
-			XMLObject object = unmarshaller.unmarshall(element);
-			if (object instanceof EntitiesDescriptor) {
```
jzheaux, line 192: Is this a change you intended? It looks like you unintentionally removed support for `EntitiesDescriptor`.

saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/registration/OpenSamlRelyingPartyRegistrationBuilderHttpMessageConverterTests.java
```diff
@@ -54,11 +54,11 @@
 	private static final String SINGLE_SIGN_ON_SERVICE_TEMPLATE = "<md:SingleSignOnService Binding=\"urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect\" "
 			+ "Location=\"sso-location\"/>";
 
-	private OpenSamlRelyingPartyRegistrationBuilderHttpMessageConverter converter;
+	private OpenSamlAssertingPartyMetadataConverter converter;
```
jzheaux, line 57: This is just a reminder to add some file-based tests as I imagine you are just looking for initial feedback right now.

jzheaux (changes requested):
Thanks for the PR, @ryan13mt! I've left some feedback inline.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/OpenSamlAssertingPartyMetadataConverter.java
ryan13mt: But won't the use of the `ResourceLoader` always return an `InputStream` instead of an `HttpInputMessage` with this new approach? Irrespective if it's a file or a url location that is what i have seen from the tests i have done. This was the reason i removed the `HttpMessageConverter` completely.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/OpenSamlAssertingPartyMetadataConverter.java
ryan13mt: Sorry for that, must have been moved to another place of the class when formatting wrongly

saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/registration/OpenSamlRelyingPartyRegistrationBuilderHttpMessageConverterTests.java
ryan13mt: Changed tests to handle an input stream instead of an http message

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/OpenSamlAssertingPartyMetadataConverter.java
jzheaux: The `HttpMessageConverter` class is still valuable for applications that are using `RestTemplate` to retrieve the resource. 

While the basic `UrlResourceLoader` may work in the most basic HTTP cases, any sophisticated tweaking -- connection timeouts, proxy configuration, etc. -- will require using `RestTemplate`.

jzheaux: Sure, no problem.

There are two use cases that these classes cater to:

* Applications that want to supply a URI and get a `RelyingPartyRegistration`. That's fulfilled by `RelyingPartyRegistrations`.
* Applications that need a `RestTemplate` to get a `RelyingPartyRegistration`. That's fulfilled by `OpenSamlRelyingPartyRegistrationBuilderHttpMessageConverter`, in conjunction with a `RestTemplate` that the application itself would instantiate.

Both of these classes - `RelyingPartyRegistrations` and `OpenSamlRelyingPartyRegistrationBuilderHttpMessageConverter` - are going to have an `InputStream` that they want converted into a `RelyingPartyRegistration.Builder`, so they will each have a private instance of `OpenSamlAssertingPartyMetadataConverter`.

When this change is complete, `RelyingPartyRegistrations` will no longer use a `RestTemplate`, which means it will no longer call `OpenSamlRelyingPartyRegistrationBuilderHttpMessageConverter`.

Does that clarify?

ryan13mt: Understood and now it makes total sense. I have restored the `OpenSamlRelyingPartyRegistrationBuilderHttpMessageConverter` and the parsing logic is now extracted into `OpenSamlAssertingPartyMetadataConverter`. I left the tests for the Http converter and added a class that also tests for the Url or File based metadata files.

ryan13mt:
@jzheaux i added everything including tests etc. Could you help me pinpoint the reason why the build is failing  please?

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistrations.java
```diff
@@ -46,6 +47,7 @@ private RelyingPartyRegistrations() {
 	 * applications will also need to provide a registrationId, like so:
 	 *
 	 * <pre>
+	 *  String metadataLocation = "file:C:\\saml\\metadata.xml"
```
jzheaux, line 50: I think this JavaDoc should show some different possible values. This is important so that readers don't get the idea that `RelyingPartyRegistrations` is only for file-based resolution.

Something like the following might do:

```
Return a {@link RelyingPartyRegistration.Builder} based off of the given SAML 2.0
Asserting Party (IDP) metadata location.

Valid locations can be classpath- or file-based or they can be HTTP endpoints. Some 
valid endpoints might include:

<pre>
  metadataLocation = "classpath:asserting-party-metadata.xml";
  metadataLocation = "file:asserting-party-metadata.xml";
  metadataLocation = "https://ap.example.org/metadata";
</pre>

Note that by default the registrationId ...
```

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistrations.java
```diff
@@ -56,14 +58,15 @@ private RelyingPartyRegistrations() {
 	 * about the asserting party. Thus, you will need to remember to still populate
 	 * anything about the relying party, like any private keys the relying party will use
 	 * for signing AuthnRequests.
-	 * @param metadataLocation
+	 * @param metadataLocation This can be anything that the
```
jzheaux, line 61: Let's please say something like "classpath- or file-based locations or HTTP endpoints" instead of tying the description to using `ResourceLoader`.

saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/registration/OpenSamlRelyingPartyRegistrationBuilderHttpMessageConverterTests.java
```diff
@@ -114,31 +112,6 @@ public void readWhenDescriptorFullySpecifiedThenConfigures() throws Exception {
 				.isEqualTo(x509Certificate(CERTIFICATE));
 	}
 
-	// gh-9051
```
jzheaux, line 117: Can you explain why these tests were removed?

jzheaux (changes requested):
Thanks for the updates, @ryan13mt. I've left a few more pieces of feedback inline.

saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/registration/OpenSamlRelyingPartyRegistrationBuilderHttpMessageConverterTests.java
ryan13mt: Sorry for that. I will be reverting the class to how it was

ryan13mt:
Hi @jzheaux applied all the changes you suggested

jzheaux:
Thanks again, @ryan13mt! This is now merged into `master` via 9a11cc84ade6bfb23de5e645dac954ff3b33364c and a small polish of e6ff57c11654d3b3b36abab09bfcb257d8912439.

Please try out the feature using the `5.5.0-SNAPSHOT`, which you can find in [Spring's snapshot Maven repo](https://repo.spring.io/libs-snapshot).

SPodlipskaya:
Thanks!!!

SPodlipskaya:
@ryan13mt @jzheaux 
Hi there 
Tried this feature in 5.5.0-SNAPSHOT on my local (macos) - it works correct
When I try it on Debian - it writes the following: 

`HttpURLConnection required for [classpath:/config.xml] but got: sun.net.www.protocol.file.FileURLConnection:file:/var/lib/tomcat8/webapps/ROOT/WEB-INF/classes/config.xml`


Do you have any ideas? 

Thank you

jzheaux:
Thanks for trying out the feature, @sssppp6. I'm not sure what the problem is, can you provide some additional detail? For example, can you share a minimal GitHub sample the reproduces the issue?

jzheaux:
Thanks, @sssppp6. Were you able to get things sorted out?

Issue #9008, opened 2020-09-09
Disable parallel deployment build in GitHub Actions CI


jgrandja:
We need to ensure that the `deployArtifacts` and `finalizeDeployArtifacts` DO NOT run in parallel.

This can be achieved by adding the `--no-parallel` argument and also by splitting the `deployArtifacts` and `finalizeDeployArtifacts` into separate commands.

See this commit 651c94b on what changes need to be applied.

ThomasVitale:
@jgrandja If needed, I'm available to help with this task. It's for the deploy-pipeline-test branch, right?

jgrandja:
That would be great @ThomasVitale ! Yes, it's for the `deploy-pipeline-test` branch.

ThomasVitale:
@jgrandja I've just opened a PR with the change.

jgrandja:
Closed via dd5baf9

PR #9062, opened 2020-09-30
Disable parallel deployment build in GitHub Actions pipelines

ThomasVitale:
Make deployArtifacts and finalizeDeployArtifacts not run in parallel in CI pipeline. 

Fixes gh-9008


jgrandja:
Thanks for the PR @ThomasVitale ! This is now merged into `deploy-pipeline-test` via dd5baf9.

Issue #9127, opened 2020-10-13
IllegalStateException when session is already invalidated


kouroups:
In AbstractPreAuthenticatedProcessingFilter.java in the following block of code

`if (AbstractPreAuthenticatedProcessingFilter.this.invalidateSessionOnPrincipalChange) {
				SecurityContextHolder.clearContext();
				HttpSession session = request.getSession(false);
				if (session != null) {
					AbstractPreAuthenticatedProcessingFilter.this.logger.debug("Invalidating existing session");
					session.invalidate();
					request.getSession();
				}
}
`
if the session is not null then the expectation is to invalidate it and get a new session but if the session is already invalidated then an IllegalStateException is thrown and an error is returned to the user. 

Stacktrace: 

`	java.lang.IllegalStateException: invalidate: Session already invalidated
	at org.apache.catalina.session.StandardSession.invalidate(StandardSession.java:1217)
	at org.apache.catalina.session.StandardSessionFacade.invalidate(StandardSessionFacade.java:170)
	at org.springframework.security.web.authentication.preauth.AbstractPreAuthenticatedProcessingFilter$PreAuthenticatedProcessingRequestMatcher.matches(AbstractPreAuthenticatedProcessingFilter.java:372)
	at org.springframework.security.web.authentication.preauth.AbstractPreAuthenticatedProcessingFilter.doFilter(AbstractPreAuthenticatedProcessingFilter.java:120)
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:334)`

Expectation is to invalidate the session and get a new one so given that session is already invalidate then the exception should be caught and allow it to get a new session

jzheaux:
This seems reasonable, but I'd like to better understand your specific use case. Can you explain under what circumstances the existing session is getting invalidated before it reaches this filter?

kouroups:
Hey Josn,  I tried to reproduce it using various ways but I wasn't able, so I suspect is race conditions. But I see it quite often happening in logs. And looks odd to me to invalidate an existing invalidated session

jzheaux:
Closing due to https://github.com/spring-projects/spring-security/pull/9128#issuecomment-984972101

stefan-g:
@jzheaux  I'm facing the same issue. I have a  restcontroller with a /logout endpoint where the session is invalidated.
The exception does't happen all the time. Is there a way to prevent that the session is saved after it was invalidated? 

[...]

PR #9128, opened 2020-10-13
Catch iIlegalStateException when session is already invalidate

kouroups:
When a session is already invalidated then an uncaught exception is thrown. This is a proposed fix for  issue [9127](https://github.com/spring-projects/spring-security/issues/9127).


jzheaux:
I'd prefer to leave this class as-is.

The reason is that some applications may rely on the existing exception handling behavior, even if your application does not.

If you would like to have a more lenient `invalidate`, then you can wrap `HttpServletRequest#getSession` [with relative ease](https://github.com/jzheaux/spring-security-examples/tree/master/gh-9128) in a custom filter. Or, since it appears to be [caused by a race condition](https://github.com/spring-projects/spring-security/issues/9127#issuecomment-718697912), perhaps that can be addressed instead.

PR #9145, opened 2020-10-22
[docs]Add white space before strong notation.

mahaker:
### Before

`ofother ...`

https://docs.spring.io/spring-security/site/docs/current/reference/html5/#appendix-faq-concurrent-session-same-browser

スクリーンショット 2020-10-22 13 50 27

### After

`of other ...`

Generated PDF.

スクリーンショット 2020-10-22 14 01 16


jzheaux:
Good catch, @mahaker! This is now merged.

PR #9146, opened 2020-10-22
Add refresh token expiration support

alek-sys:
In order to support Authorization Server refresh token grant https://github.com/spring-projects-experimental/spring-authorization-server/pull/128


jgrandja:
Thanks for the PR @alek-sys ! This is now in master.

Issue #8951, opened 2020-08-10
Some classes should implement MessageSourceAware


amergey:
**Describe the bug**
The javadoc of org.springframework.security.core.SpringSecurityMessageSource
says

> * The default <code>MessageSource</code> used by Spring Security.
> <p>
> All Spring Security classes requiring message localization will by default use this
> class. However, all such classes will also implement <code>MessageSourceAware</code> so
> that the application context can inject an alternative message source. Therefore this
> class is only used when the deployment environment has not specified an alternative
> message source.

but specifying an alternative message source is not working everywhere in spring security because some classes are not implementing MessageSourceAware.
I found 4 classes using SpringSecurityMessageSource without implementing MessageSourceAware:

- org.springframework.security.authentication.AbstractUserDetailsReactiveAuthenticationManager

- org.springframework.security.web.access.ExceptionTranslationFilter

- org.springframework.security.web.authentication.preauth.x509.SubjectDnX509PrincipalExtractor

- org.springframework.security.web.authentication.rememberme.AbstractRememberMeServices

These should probably implement MessageSourceAware, all others seems to be good.

amergey:
Would it help if I start a PR for this ?

rwinch:
Thanks for the report an offer to create a PR! We'd love to get a PR from you.

PR #9151, opened 2020-10-26
Implement MessageSourceAware where missing

amergey:
Closes gh-8951

### DIFF
--- a/core/src/main/java/org/springframework/security/authentication/AbstractUserDetailsReactiveAuthenticationManager.java
+++ b/core/src/main/java/org/springframework/security/authentication/AbstractUserDetailsReactiveAuthenticationManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,6 +22,8 @@ import reactor.core.publisher.Mono;
 import reactor.core.scheduler.Scheduler;
 import reactor.core.scheduler.Schedulers;
 
+import org.springframework.context.MessageSource;
+import org.springframework.context.MessageSourceAware;
 import org.springframework.context.support.MessageSourceAccessor;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.SpringSecurityMessageSource;
@@ -45,7 +47,8 @@ import org.springframework.util.Assert;
  * @author Eddú Meléndez
  * @since 5.2
  */
-public abstract class AbstractUserDetailsReactiveAuthenticationManager implements ReactiveAuthenticationManager {
+public abstract class AbstractUserDetailsReactiveAuthenticationManager
+		implements ReactiveAuthenticationManager, MessageSourceAware {
 
 	protected final Log logger = LogFactory.getLog(getClass());
 
@@ -162,6 +165,15 @@ public abstract class AbstractUserDetailsReactiveAuthenticationManager implement
 	public void setPostAuthenticationChecks(UserDetailsChecker postAuthenticationChecks) {
 		Assert.notNull(this.postAuthenticationChecks, "postAuthenticationChecks cannot be null");
 		this.postAuthenticationChecks = postAuthenticationChecks;
+	}
+
+	/**
+	 * @since 5.5
+	 */
+	@Override
+	public void setMessageSource(MessageSource messageSource) {
+		Assert.notNull(messageSource, "messageSource cannot be null");
+		this.messages = new MessageSourceAccessor(messageSource);
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/springframework/security/authentication/AbstractUserDetailsReactiveAuthenticationManager.java
```diff
@@ -164,6 +167,11 @@ public void setPostAuthenticationChecks(UserDetailsChecker postAuthenticationChe
 		this.postAuthenticationChecks = postAuthenticationChecks;
 	}
 
+	@Override
```
jzheaux, line 170: Will you please add JavaDoc for these setters? In the JavaDoc, please add `@since 5.5`.

core/src/main/java/org/springframework/security/authentication/AbstractUserDetailsReactiveAuthenticationManager.java
```diff
@@ -22,6 +22,8 @@
 import reactor.core.scheduler.Scheduler;
 import reactor.core.scheduler.Schedulers;
 
+import org.springframework.context.MessageSource;
```
jzheaux, line 25: Please make sure that each file uses the most recent copyright year. In this case, that would mean changing `2002-2019` to `2002-2020`.

core/src/main/java/org/springframework/security/authentication/AbstractUserDetailsReactiveAuthenticationManager.java
```diff
@@ -164,6 +167,11 @@ public void setPostAuthenticationChecks(UserDetailsChecker postAuthenticationChe
 		this.postAuthenticationChecks = postAuthenticationChecks;
 	}
 
+	@Override
+	public void setMessageSource(MessageSource messageSource) {
+		this.messages = new MessageSourceAccessor(messageSource);
```
jzheaux, line 172: It would also be nice if the setters ensured a non-null parameter, like with `Assert.notNull`.

jzheaux (changes requested):
Thanks for the PR, @amergey! I've left some feedback inline.

jzheaux (changes requested):
Thanks for the updates, @amergey. Lastly, would you please add tests that confirm that setting the message source has the desired effect on each class?

amergey:
> Thanks for the updates, @amergey. Lastly, would you please add tests that confirm that setting the message source has the desired effect on each class?

@jzheaux I do not see what to test here, the bug was these classes were not implementing MessageSourceAware as they should, fixing the bug is juts to implement it and I do not see what relevant test we could add there. 
( MessageSourceAware behavior is already covered by tests like AutowireBeanFactoryObjectPostProcessorTests. )

jzheaux:
> the bug was these classes were not implementing MessageSourceAware as they should, fixing the bug is juts to implement it

Good points. It's quite common in Spring Security to test that setters have non-null guards. For example, you can see this in `JdbcDaoImplTests#setMessageSourceWhenNullThenThrowsException`. I believe this PR should at least test that the setters check for `null`.

Additionally, I was thinking that tests would confirm that the `MessageSource` used by the class is the one that the setter provided. But, after trying it out myself on `SubjectDnX509PrincipalExtractorTests`, it seems to create a test that heavily overlaps other tests. Also, this kind of test isn't done anywhere else in the repo, so maybe we can leave those out.

Can you add tests that confirm that the setters check for null?

amergey:
> Can you add tests that confirm that the setters check for null?
sure ! I have added 2 tests per class similar to what is done elsewhere

jzheaux:
@amergey Thanks for the update! 

Sorry, I feel like I miscommunicated on my feedback - I'd prefer to not expose fields for the sake of testing. I think it's okay to only add tests that confirm that the setters check for null.

Do you feel strongly that the second of two tests in each class should stay? If so, let's use default scope instead of `protected` scope. If not, let's remove the tests.

amergey:
> @amergey Thanks for the update!
> 
> Sorry, I feel like I miscommunicated on my feedback - I'd prefer to not expose fields for the sake of testing. I think it's okay to only add tests that confirm that the setters check for null.
> 
> Do you feel strongly that the second of two tests in each class should stay? If so, let's use default scope instead of `protected` scope. If not, let's remove the tests.

I just did that like this to be consistent with what existed. If you think it is better without the second test, then I can remove them and move scope to private

jzheaux:
> I just did that like this to be consistent with what existed.

Thanks, @amergey. Can you humor me for a moment and point out where Spring Security is testing setters by making the field `protected`? I want to make sure I'm not missing something. I agree that the first of the two tests is consistent with the rest of Spring Security, but I'm having trouble seeing the second of the two tests as being consistent with what exists.

amergey:
> > I just did that like this to be consistent with what existed.
> 
> Thanks, @amergey. Can you humor me for a moment and point out where Spring Security is testing setters by making the field `protected`? I want to make sure I'm not missing something. I agree that the first of the two tests is consistent with the rest of Spring Security, but I'm having trouble seeing the second of the two tests as being consistent with what exists.

When I said consistent I did not have in mind testing setters by making the field protected,  it was about having the "message" field itself protected (which seems to be protected when a class implements MessageSourceAware). 
I have seen the second test in another MessageSourceAware implementation, but I need to find which one I used as sample

jzheaux (approved):

jzheaux:
I see, @amergey, thanks for clarifying.

Issue #9152, opened 2020-10-26
It is not necessary to fetch all user sessions if unlimited sessions are set in the ConcurrentSessionControlAuthenticationStrategy.


ApRobot:
**Expected Behavior**

In `ConcurrentSessionControlAuthenticationStrategy.onAuthentication`, If `maximumSessions` value settled to -1, which means there are no restrictions for user sessions count, then no need fetch all user sessions in that case. We should just skipping any following computations.

**Current Behavior**

`ConcurrentSessionControlAuthenticationStrategy.onAuthentication` start with `getAllSessions` whatever or not there is restriction on session count, and if is not, then just does nothing with retreived data.

**Context**

Save little peace of computing resources.

PR #9153, opened 2020-10-26
Change guard expressions order

ApRobot:
Check of allowed user sessions count moved to head for avoid unnecessary fetching all user sessions.

Fixes gh-9152


eleftherias (changes requested):
Thanks @ApRobot! This is a nice enhancement. It looks like your branch is slightly behind, could you rebase on the latest master branch and then we can merge this.

eleftherias:
Thanks @ApRobot! This is now merged into master.

PR #9157, opened 2020-10-27
Reactive doc points to unit tests

gberche-orange:
As a spring security user
* in order to adapt the servlet configuration syntax such as multiple [filter chains](https://spring.io/guides/topicals/spring-security-architecture#_request_matching_for_dispatch_and_authorization) to reactive syntax
* I need example/documentation of such syntax (as method names do not match)

This PR adds to the reactive documentation pointer to unit tests as further examples of configurations. 

It might be possible to user asciidoc syntax to directly source the `MultiSecurityHttpConfig` instead of pointing to github (but this falls outside of my asciidoc so far). This may make maintenance of this section overtime easier if ever tests would detect broken path when `MultiSecurityHttpConfig` source code gets refactored.

### DIFF
--- a/docs/manual/src/docs/asciidoc/_includes/reactive/webflux.adoc
+++ b/docs/manual/src/docs/asciidoc/_includes/reactive/webflux.adoc
@@ -125,5 +125,65 @@ class HelloWebfluxSecurityConfig {
 This configuration explicitly sets up all the same things as our minimal configuration.
 From here you can easily make the changes to the defaults.
 
-You can find more examples of explicit configuration in unit tests, by searching https://github.com/spring-projects/spring-security/search?q=path%3Aconfig%2Fsrc%2Ftest%2F+EnableWebFluxSecurity[EnableWebFluxSecurity in the `config/src/test/` directory], e.g. https://github.com/spring-projects/spring-security/blob/9cf3129d7afa2abb439aba6aadfee0a2c8c784bf/config/src/test/java/org/springframework/security/config/annotation/web/reactive/EnableWebFluxSecurityTests.java#L349-L366[MultiSecurityHttpConfig] illustrating multiple `SecurityWebFilterChain` beans.
+You can find more examples of explicit configuration in unit tests, by searching https://github.com/spring-projects/spring-security/search?q=path%3Aconfig%2Fsrc%2Ftest%2F+EnableWebFluxSecurity[EnableWebFluxSecurity in the `config/src/test/` directory], e.g. https://github.com/spring-projects/spring-security/blob/9cf3129d7afa2abb439aba6aadfee0a2c8c784bf/config/src/test/java/org/springframework/security/config/annotation/web/reactive/EnableWebFluxSecurityTests.java#L349-L366[MultiSecurityHttpConfig] illustrating multiple `SecurityWebFilterChain` beans, which is also explained in section below.
+
+=== Multiple chains support
+
+We can configure multiple `SecurityWebFilterChain` instances.
+
+For example, the following is an example of having a different configuration for URL's that start with `/v2/`
+
+[source,java]
+----
+@Configuration
+@EnableWebFluxSecurity
+public class MultiWebfluxHttpSecurityConfig {
+
+	@Bean
+	public MapReactiveUserDetailsService userDetailsService() {
+		UserDetails user = User.withDefaultPasswordEncoder()
+			.username("user")
+			.password("user")
+			.roles("USER")
+			.build();
+		return new MapReactiveUserDetailsService(user);
+	}
+
+
+	@Order(SecurityProperties.BASIC_AUTH_ORDER - 10)                                                     <1>
+	@Bean
+	public SecurityWebFilterChain osbUnRestrictedSpringSecurityFilterChain(ServerHttpSecurity http) {
+		http
+			.csrf().disable()
+			//Scope this filter only to /v2 requests, otherwise this will handle other filters as well
+			//see background at https://spring.io/guides/topicals/spring-security-architecture#_creating_and_customizing_filter_chains
+			.securityMatcher(new PathPatternParserServerWebExchangeMatcher("/v2/**"))                    <2>
+			.authorizeExchange()
+				.anyExchange().permitAll();
+		return http.build();
+	}
+
+	@Bean
+	public SecurityWebFilterChain actuatorSpringSecurityFilterChain(ServerHttpSecurity http) {           <3>
+		http.authorizeExchange((exchanges) -> exchanges
+			.matchers(EndpointRequest.to(HealthEndpoint.class)).permitAll()
+			.matchers(EndpointRequest.toAnyEndpoint().excluding(HealthEndpoint.class)).hasRole("USER"));
+		//http basic and form login are configured for all matchers above. If a different config is needed, then
+		//we need to split it into a distinct spring-security filter
+		http.httpBasic(Customizer.withDefaults());                                                       <4>
+		http.formLogin(Customizer.withDefaults());
+		return http.build();
+	}
+
+
+}
+----
+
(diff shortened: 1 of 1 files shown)


docs/manual/src/docs/asciidoc/_includes/reactive/webflux.adoc
```diff
@@ -124,3 +124,6 @@ class HelloWebfluxSecurityConfig {
 
 This configuration explicitly sets up all the same things as our minimal configuration.
 From here you can easily make the changes to the defaults.
+
+You can find more examples of explicit configuration in unit tests, by searching https://github.com/spring-projects/spring-security/search?q=path%3Aconfig%2Fsrc%2Ftest%2F+EnableWebFluxSecurity[EnableWebFluxSecurity in the `config/src/test/` directory], e.g. https://github.com/spring-projects/spring-security/blob/9cf3129d7afa2abb439aba6aadfee0a2c8c784bf/config/src/test/java/org/springframework/security/config/annotation/web/reactive/EnableWebFluxSecurityTests.java#L349-L366[MultiSecurityHttpConfig] illustrating multiple `SecurityWebFilterChain` beans.
```
jzheaux, line 128: I like the additional sentence here that points the reader to the unit tests.

The specific example of multiple chains seems too narrow for this point in the documentation. Could this part of the sentence be moved to a section where multiple chains is discussed? Such a section might need to be added.

jzheaux (changes requested):
Thanks, @gberche-orange for the PR! I've left my feedback inline.

gberche-orange:
Thanks @jzheaux for your feedback. I've added a section about webflux support for multiple filter, inspired by my project at https://github.com/orange-cloudfoundry/osb-reverse-proxy/blob/6e38440c53992b343606c872375c1a087c8ccb40/src/main/java/com/orange/oss/osbreverseproxy/SecurityConfig.java#L99-L151 which contains a bit more background which I trimmed from the contributed sample.

As a disclaimer, I'm still pretty new to spring-security, and may benefit from guidance/feedback on how to simplify this configuration.

docs/manual/src/docs/asciidoc/_includes/reactive/webflux.adoc
```diff
@@ -124,3 +124,66 @@ class HelloWebfluxSecurityConfig {
 
 This configuration explicitly sets up all the same things as our minimal configuration.
 From here you can easily make the changes to the defaults.
+
+You can find more examples of explicit configuration in unit tests, by searching https://github.com/spring-projects/spring-security/search?q=path%3Aconfig%2Fsrc%2Ftest%2F+EnableWebFluxSecurity[EnableWebFluxSecurity in the `config/src/test/` directory], e.g. https://github.com/spring-projects/spring-security/blob/9cf3129d7afa2abb439aba6aadfee0a2c8c784bf/config/src/test/java/org/springframework/security/config/annotation/web/reactive/EnableWebFluxSecurityTests.java#L349-L366[MultiSecurityHttpConfig] illustrating multiple `SecurityWebFilterChain` beans, which is also explained in section below.
```
jzheaux, line 128: Even with the new section, I still think that this sentence should stay focused on one thing and not two. Could we simplify it to:

```
You can find more examples of explicit configuration in unit tests, by searching https://github.com/spring-projects/spring-security/search?q=path%3Aconfig%2Fsrc%2Ftest%2F+EnableWebFluxSecurity[EnableWebFluxSecurity in the `config/src/test/` directory].
```

docs/manual/src/docs/asciidoc/_includes/reactive/webflux.adoc
```diff
@@ -124,3 +124,66 @@ class HelloWebfluxSecurityConfig {
...
+
+=== Multiple chains support
```
jzheaux, line 130: Before each section it's helpful to have an anchor, like so:

```
[[jc-webflux-multiple-filter-chains]]
=== Multiple Chains Support
...
```

docs/manual/src/docs/asciidoc/_includes/reactive/webflux.adoc
```diff
@@ -124,3 +124,66 @@ class HelloWebfluxSecurityConfig {
...
+			.password("user")
+			.roles("USER")
+			.build();
+		return new MapReactiveUserDetailsService(user);
+	}
+
+
+	@Order(SecurityProperties.BASIC_AUTH_ORDER - 10)                                                     <1>
+	@Bean
+	public SecurityWebFilterChain osbUnRestrictedSpringSecurityFilterChain(ServerHttpSecurity http) {
```
jzheaux, line 155: I'd recommend simplifying the example as much as possible, so as to keep the focus on the principle you are illustrating.

Consider starting with `MultiSecurityHttpConfig` from `EnableWebFluxSecurityTests` and making it more complex, if you see the need.

jzheaux (changes requested):
Thanks for the changes, @gberche-orange! I've left some additional feedback inline.

docs/manual/src/docs/asciidoc/_includes/reactive/webflux.adoc
gberche-orange: I felt that while unit tests are a useful workaround for lack of documentation, they are still hard to grasp for the users of a library. I feel that pointing to a specific test explained in more details in a dedicated session lowers the barrier to reading unit tests by users.

I've however removed the `MultiSecurityHttpConfig` as requested.

docs/manual/src/docs/asciidoc/_includes/reactive/webflux.adoc
gberche-orange: thanks, added anchor

docs/manual/src/docs/asciidoc/_includes/reactive/webflux.adoc
gberche-orange: In this specific example, the `MultiSecurityHttpConfig` seems to rely on defaults for its second `SecurityWebFilterChain`. 

https://github.com/spring-projects/spring-security/blob/9cf3129d7afa2abb439aba6aadfee0a2c8c784bf/config/src/test/java/org/springframework/security/config/annotation/web/reactive/EnableWebFluxSecurityTests.java#L349-L366

and it was not clear to me how readers would discover those defaults that lead to others requests such as `/test`be accepted.
https://github.com/spring-projects/spring-security/blob/9cf3129d7afa2abb439aba6aadfee0a2c8c784bf/config/src/test/java/org/springframework/security/config/annotation/web/reactive/EnableWebFluxSecurityTests.java#L247-L264

To match the existing servlet api example at https://github.com/spring-projects/spring-security/blob/1ca47f5cd0c5ca574f39d25bed82f8b83ced7732/docs/manual/src/docs/asciidoc/_includes/servlet/java-configuration/index.adoc#multiple-httpsecurity I've therefore reused an explicit distinct config inspired from `WebfluxFormSecurityConfig`.

gberche-orange:
Thanks @jzheaux for your additional review.

It is quite challenging for me as a newcomer and 1st time contributor to reverse engineer documentation from unit tests and to come up with documentation that match quality that that core maintainers would produce. 

If there are significant improvements required to this PR, you might want consider then feel free to then to use this PR as a starting point, and consider further improving it to match the expected documentation quality.

jzheaux (approved):

jzheaux:
Thanks for the PR, @gberche-orange! This is now merged into `master` with ab9a310485b43e7a5cd08e0ffba3cb641cb43991 and 69336fb3ec40e1c62ad1700e15fa0d6e03cb57d4.

Also, I polished it the new section with 8b7751f5f400d1663b3a8f60885b1837983906ea to update some of the indentation and standardize the explanations to fit the typical writing style of the reference manual.

gberche-orange:
Great, thanks @jzheaux !

PR #9163, opened 2020-10-29
Preconfigure class loader for message source

### LINKED ISSUE #8999: Default class loader in SpringSecurityMessageSource only works if Spring Security and Spring Framework share a classloader
**Describe the bug**

`SpringSecurityMessageSource.getAccessor()` is used in multiple places to set the default message source but the configured default class loader does not work if the spring security class loader is a different one to the spring framework class loader.

The super class has a null `bundleClassLoader` and defaults the bean class loader to:

```
ClassUtils.getDefaultClassLoader();
```

If no context class loader is present, it will be the class loader which loaded the ClassUtils class.

I would suggest that `getAccessor` should do something like:

```
	public static MessageSourceAccessor getAccessor() {
		final SpringSecurityMessageSource source = new SpringSecurityMessageSource();
		source.setBundleClassLoader(SpringSecurityMessageSource.class.getClassLoader());
		return new MessageSourceAccessor(source);
	}
```

It is possible to work around this by setting my own message source with the class loader correctly configured, but this is tedious and requires multiple internally created objects to be reconfigured.

aclemons:
The helper method SpringSecurityMessageSource.getAccessor does not
configure a bundle classloader on the message source instance it
creates. This means it will use the default configured by
ResourceBundleMessageSource which is ClassUtils.getDefaultClassLoader().
In a classpath constrained environment like OSGi, this will very likely
be a different classloader to the one which loaded the Spring Security
classes.

This commit explicitly configures a classloader on the message source
with a better default value.

Closes #8999

### DIFF
--- a/core/src/main/java/org/springframework/security/core/SpringSecurityMessageSource.java
+++ b/core/src/main/java/org/springframework/security/core/SpringSecurityMessageSource.java
@@ -38,7 +38,9 @@ public class SpringSecurityMessageSource extends ResourceBundleMessageSource {
 	}
 
 	public static MessageSourceAccessor getAccessor() {
-		return new MessageSourceAccessor(new SpringSecurityMessageSource());
+		final SpringSecurityMessageSource source = new SpringSecurityMessageSource();
+		source.setBundleClassLoader(SpringSecurityMessageSource.class.getClassLoader());
+		return new MessageSourceAccessor(source);
 	}
 
 }


pivotal-issuemaster:
@aclemons Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=9163)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=9163) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@aclemons Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=9163)!

rwinch:
I don't think we should optimize for scenarios like OSGi where there may be different `ClassLoader`s for Spring and Spring Security. The default reflects the value that should be used in most cases and we don't have a lot of OSGi users these days.

If we want to optimize this, then I think we could leverage `BeanClassLoaderAware` to get a hold of the `ClassLoader` that should be used and inject that into an explicit instance of `SpringSecurityMessageSource` which is in turn injected into any `Bean` that requires it.

aclemons:
I don't quite follow your suggestion. When I noticed that the default message source was not properly configured, I looked to configure explicitly the message source on those beans myself. Some of them implement MessageSourceAware so I can configure a SpringSecurityMessageSource in my application context for those.

Then there is code like this in [ExceptionTranslationFilter](https://github.com/spring-projects/spring-security/blob/5.4.1/web/src/main/java/org/springframework/security/web/access/ExceptionTranslationFilter.java#L92) or [AbstractRememberMeServices](https://github.com/spring-projects/spring-security/blob/5.4.1/web/src/main/java/org/springframework/security/web/authentication/rememberme/AbstractRememberMeServices.java#L74) etc where the message source is not configurable and I was stuck.

It then seemed simplest to just be explicit in this static helper that is used ~20 times in the code base and configure the correct class loader. My change could be slightly adjusted to also behave like ClassUtils.getDefaultClassLoader() to take the current thread's context class loader into account if we wanted.

Is this optimising for OSGi? I'm not sure - does this code correctly load the message source when using modules on newer jdks?

jzheaux:
https://github.com/spring-projects/spring-security/pull/9151 might be of value here as it is adding `setMessageSource` to those two classes.

rwinch:
@aclemons We haven't had any complaints about this behavior from others so I'm not aware of any other ways for it to reproduce. If you can provide a reproducer on when the issue occurs (other than OSGi) it would help us to understand the problem better. At this point it seems speculative.

aclemons:
Sorry, I just noticed Spring Security does not ship with module declarations (module-info.class) so my question about newer jdks won't matter until it does.

jzheaux:
Now that #9151 has been merged, it appears this PR can be closed. Feel free to reopen, though, if there is more to discuss.

Issue #8535, opened 2020-05-15
Support customization of BearerTokenResolver in JwtIssuerAuthenticationManagerResolver


marci74:
**Expected Behavior**

It should be possible to set or customize the BearerTokenResolver inside the JwtIssuerAuthenticationManagerResolver

**Current Behavior**

The customization is not possible.

**Context**

I have to support multiple token issuers and bearer tokens in query parameters. I would like to use the new JwtIssuerAuthenticationManagerResolver but I cannot alter the "allowUriQueryParameter"
property of the DefaultBearerTokenResolver which is being used inside the JwtIssuerAuthenticationManagerResolver.JwtClaimIssuerConverter class.

jzheaux:
Thanks for the report, @marci74. I think what needs to be done here is to remove the dependency on the bearer token resolver altogether.

Since the resolver can return an `AuthenticationManager`, it could return one that retrieves the token from the given authentication, e.g.:

```java
public AuthenticationManager resolve(HttpServletRequest request) {
    return authentication -> {
        BearerTokenAuthenticationToken token = (BearerTokenAuthenticationToken) authentication;
        String issuer = this.issuerConverter.convert(token);
        // ...
        return authenticationManager.authenticate(token);
    };
}
```

Something similar can be done on the reactive side as well.

Would you be interested in filing a PR to address this?

AbstractConcept:
@jzheaux I have run into a very similar issue, if not the same one.

Using an external `oauth2ResourceServer`, in a multi-tenant environment, it should be possible to override / select a custom `JwtDecoder` / `jwtAuthenticationProvider` with a custom `jwtAuthenticationConverter`, as not all of the JWT tokens are the same, and this leads to errors.

Currently, according to the official documentation, if there is a need to support multi-tenancy by JWT Claim, one can use `JwtIssuerAuthenticationManagerResolver` that has to be configured by a class extending `WebSecurityConfigurerAdapter`, however, it comes with an overriden `public AuthenticationManager resolve(String issuer)` method, that contains a predefined, hardcoded basic `JwtAuthenticationProvider`, that also contains a predefined, hardcoded `JwtAuthenticationConverter`, which in turn contains a predefined, hardcoded `JwtGrantedAuthoritiesConverter`, the root of my problems.

[...]

PR #9168, opened 2020-10-29
Enable customization of BearerTokenResolver by adding a setter for JwtClaimIssuerConverter on JwtIssuerAuthenticationManagerResolver

arvidOtt:
**Description**
When using the `JwtIssuerAuthenticationManagerResolver` there should be a way to replace the `DefaultBearerTokenResolver` (or here the entire `JwtClaimIssuerConverter`) by a custom implementation which is requested in #8535. 

I decided to set the entire `JwtClaimIssuerConverter` because it is declared as type `Converter<HttpServletRequest,String>` and implementing a setter or new constructor on it would required to change the type of the attribute directly to `JwtClaimIssuerConverter` class.

**Context**
I use the JwtIssuerAuthenticationManagerResolver in a web socket backend which receives the JWT not in the Authorization Header but in a custom X-Authorization Cookie Header. Therefore, I need to use my own BearerTokenResolver implementation which gets the JWT from there. This seems not to be possible at the moment.

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
@@ -131,7 +131,7 @@ public final class JwtIssuerAuthenticationManagerResolver implements Authenticat
 	}
 
 	/**
-	 * Set a custom issuer converter
+	 * Set a custom bearer token resolver
 	 *
 	 * @since 5.5
 	 */
@@ -142,10 +142,10 @@ public final class JwtIssuerAuthenticationManagerResolver implements Authenticat
 
 	private static class JwtClaimIssuerConverter implements Converter<HttpServletRequest, String> {
 
-		private BearerTokenResolver resolver;
+		private final BearerTokenResolver resolver;
 
 		JwtClaimIssuerConverter() {
-			this.resolver = new DefaultBearerTokenResolver();
+			this(new DefaultBearerTokenResolver());
 		}
 
 		JwtClaimIssuerConverter(BearerTokenResolver bearerTokenResolver) {
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
@@ -135,14 +135,23 @@ public final class JwtIssuerAuthenticationManagerResolver implements Authenticat
 	 *
 	 * @since 5.5
 	 */
-	public void setIssuerConverter(Converter<HttpServletRequest, String> issuerConverter) {
-		Assert.notNull(issuerConverter, "issuerConverter cannot be null");
-		this.issuerConverter = issuerConverter;
+	public void setBearerTokenResolver(BearerTokenResolver bearerTokenResolver) {
+		Assert.notNull(bearerTokenResolver, "bearerTokenResolver cannot be null");
+		this.issuerConverter = new JwtClaimIssuerConverter(bearerTokenResolver);
 	}
 
 	private static class JwtClaimIssuerConverter implements Converter<HttpServletRequest, String> {
 
-		private final BearerTokenResolver resolver = new DefaultBearerTokenResolver();
+		private BearerTokenResolver resolver;
+
+		JwtClaimIssuerConverter() {
+			this.resolver = new DefaultBearerTokenResolver();
+		}
+
+		JwtClaimIssuerConverter(BearerTokenResolver bearerTokenResolver) {
+			Assert.notNull(bearerTokenResolver, "bearerTokenResolver cannot be null");
+			this.resolver = bearerTokenResolver;
+		}
 
 		@Override
 		public String convert(@NonNull HttpServletRequest request) {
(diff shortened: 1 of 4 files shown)


arvidOtt:
@jzheaux this is basically a slim version of #9005. I hope one of them can be done soon.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
```diff
@@ -130,6 +130,16 @@ public AuthenticationManager resolve(HttpServletRequest request) {
 		return authenticationManager;
 	}
 
+	/**
+	 * Set a custom issuer converter
+	 *
+	 * @since 5.5
+	 */
+	public void setIssuerConverter(Converter<HttpServletRequest, String> issuerConverter) {
```
jzheaux, line 138: Instead of `setIssuerConverter`, could you do:

```java
    this.issuerConverter = new JwtClaimIssuerConverter(bearerTokenResolver);
}
```

That way, the API uses a familiar interface that they most likely already configured in their app.

jzheaux (changes requested):
Thanks, @arvidOtt! I've left some feedback inline.

Also, would you please add unit tests to demonstrate that the feature works?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
arvidOtt: Okay :) Then I would also need to add another constructor to the `JwtClaimIssuerConverter` to pass the resolver / converter correct?

arvidOtt: @jzheaux I hope this is the way your feedback was meant.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
```diff
@@ -130,9 +130,28 @@ public AuthenticationManager resolve(HttpServletRequest request) {
 		return authenticationManager;
 	}
 
+	/**
+	 * Set a custom issuer converter
```
jzheaux, line 134: Will you please update the JavaDoc so that it talks about `BearerTokenResolver` instead?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
```diff
@@ -130,9 +130,28 @@ public AuthenticationManager resolve(HttpServletRequest request) {
...
+	 * @since 5.5
+	 */
+	public void setBearerTokenResolver(BearerTokenResolver bearerTokenResolver) {
+		Assert.notNull(bearerTokenResolver, "bearerTokenResolver cannot be null");
+		this.issuerConverter = new JwtClaimIssuerConverter(bearerTokenResolver);
+	}
+
 	private static class JwtClaimIssuerConverter implements Converter<HttpServletRequest, String> {
 
-		private final BearerTokenResolver resolver = new DefaultBearerTokenResolver();
```
jzheaux, line 135: I believe this can still be `final` since it is going to be set by either constructor.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
```diff
@@ -130,9 +130,28 @@ public AuthenticationManager resolve(HttpServletRequest request) {
...
+		private BearerTokenResolver resolver;
+
+		JwtClaimIssuerConverter() {
+			this.resolver = new DefaultBearerTokenResolver();
```
jzheaux, line 148: It might be more efficient to do:

```java
this(new DefaulBearerTokenResolver());
```

jzheaux (changes requested):
Thanks for the updates, @arvidOtt! I've left some additional feedback inline.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
arvidOtt: Done.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
arvidOtt: Done.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
arvidOtt: Done 👍

arvidOtt:
@jzheaux I adapted the PR according to your feedback. Will write some test cases for the setter method next 🧪

arvidOtt:
Added unit tests for the new feature.

jzheaux (approved):

jzheaux:
Thanks, @arvidOtt! This is now merged into `master` via d0d655e18df6ae1af42317e10efcf0bb93da875d.

arvidOtt:
@jzheaux thank you for your support!

PR #9194, opened 2020-11-08
SecurityWebApplicationContextUtils cleanup gh-8868

adi-devops:
Change to cleanup SecurityWebApplicationContextUtils. Based on the recommendation in the tagged issue. 

> We no longer need to support versions of Spring Framework that old, so we should now call WebApplicationContextUtils.findWebApplicationContext

Closes gh-8868


rwinch:

Issue #9195, opened 2020-11-08
Maximum Age for the cookie created by CookieCsrfTokenRepository


sedran:
**Expected Behavior**

Maximum age of cookies created by `CookieCsrfTokenRepository` should be configurable.

**Current Behavior**

Currently, `CookieCsrfTokenRepository` generated cookies with a maximum age value of -1.

**Context**

Currently, `CookieCsrfTokenRepository` generated cookies with a maximum age value of -1.
This means the cookie will expire as soon as the browser is closed.
Under some circumtances, a cookie with a specific maximum age may be desired.

I've faced with a security tester who did not approve my project's production deployment 
until I've set a maximum age value to the csrf token cookie. Now I have to extend `CookieCsrfTokenRepository`
and override its behavior in every project that my team has developped.

I've implemented this feature and I want to send a pull request.

PR #9196, opened 2020-11-08
Allow maximum age of csrf cookie to be configured

sedran:
Allows maxAge of the generated cookie by CookieCsrfTokenRepository
to be configurable.

Prior to this commit, maximum age was set with a value of -1.

After this commit, it will be configured by the user with an either
positive or negative value. If the user does not provide a value,
it will be set -1.

An IllegalArgumentException will be thrown when
this value is set to zero.

Closes gh-9195

### DIFF
--- a/web/src/main/java/org/springframework/security/web/csrf/CookieCsrfTokenRepository.java
+++ b/web/src/main/java/org/springframework/security/web/csrf/CookieCsrfTokenRepository.java
@@ -57,6 +57,8 @@ public final class CookieCsrfTokenRepository implements CsrfTokenRepository {
 
 	private Boolean secure;
 
+	private int cookieMaxAge = -1;
+
 	public CookieCsrfTokenRepository() {
 	}
 
@@ -71,7 +73,7 @@ public final class CookieCsrfTokenRepository implements CsrfTokenRepository {
 		Cookie cookie = new Cookie(this.cookieName, tokenValue);
 		cookie.setSecure((this.secure != null) ? this.secure : request.isSecure());
 		cookie.setPath(StringUtils.hasLength(this.cookiePath) ? this.cookiePath : this.getRequestContext(request));
-		cookie.setMaxAge((token != null) ? -1 : 0);
+		cookie.setMaxAge((token != null) ? this.cookieMaxAge : 0);
 		cookie.setHttpOnly(this.cookieHttpOnly);
 		if (StringUtils.hasLength(this.cookieDomain)) {
 			cookie.setDomain(this.cookieDomain);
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/springframework/security/web/csrf/CookieCsrfTokenRepository.java
```diff
@@ -192,4 +194,30 @@ public void setSecure(Boolean secure) {
...
+	 * <p>
+	 * A zero value causes the cookie to be deleted immediately therefore it is not a
+	 * valid value and in that case an {@link IllegalArgumentException} will be thrown.
+	 * @param cookieMaxAge an integer specifying the maximum age of the cookie in seconds;
+	 * if negative, means the cookie is not stored; if zero, the method throws an
+	 * {@link IllegalArgumentException}
+	 * @since 5.5
+	 */
+	public void setCookieMaxAge(int cookieMaxAge) {
+		Assert.isTrue(cookieMaxAge != 0, "cookieMaxAge is not zero");
```
rwinch, line 219: Can you please change this to be cookieMaxAge cannot be zero? As it is, the error the user will see if it is `0` would state it is not zero which would be confusing.

```suggestion
		Assert.isTrue(cookieMaxAge != 0, "cookieMaxAge cannot be zero");
```

rwinch (changes requested):
Thanks for the PR! I've provided feedback inline.

web/src/main/java/org/springframework/security/web/csrf/CookieCsrfTokenRepository.java
sedran: Thank you for your feedback. In this class there are other assertions written in this way. I've tried to follow the convention in the class. Can I change the other assertions in the same file in order to be consistent?

sedran:
The automated build fails because of a dependency error from spring maven repository. How can we solve that?

```
> Could not HEAD 'https://repo.spring.io/libs-snapshot/org/infinispan/infinispan-bom/11.0.5.Final/infinispan-bom-11.0.5.Final.pom'. 
Received status code 401 from server: Unauthorized
```

jzheaux:
Sorry for the issue you experienced with the build, @sedran - it was due to an infrastructural change with Spring's Maven repositories. 

If you rebase, then the error should be addressed.

sedran:
> Sorry for the issue you experienced with the build, @sedran - it was due to an infrastructural change with Spring's Maven repositories.
> 
> If you rebase, then the error should be addressed.

I have amend/force-pushed my branch to trigger auto build as soon as I see the CI status of the project is green. 
Now I can see "All checks have passed" text and there is no conflict. I also implemented the requested changes by @rwinch .

Is there anything I can do for this PR to be merged? Please let me know.

Thanks.

rwinch (approved):
Thanks for the updates and the ping. The changes look good to me

Issue #9199, opened 2020-11-11
Saml2 support overriding the SAMLRequest parameters


adi-devops:
**Expected Behavior**
As a service provider, I should be able to override the following parameter in a SAMLRequest

- <saml:AuthnContextClassRef>
- ForceAuthn
- IsPassive

**Current Behavior**
Currently we are having the default values configured which is 

- <saml:AuthnContextClassRef> = empty
- ForceAuthn =false [OpenSamlAuthenticationRequestFactory.java#L171](https://github.com/spring-projects/spring-security/blob/master/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationRequestFactory.java#L171)
- IsPassive = false [OpenSamlAuthenticationRequestFactory.java#L172](https://github.com/spring-projects/spring-security/blob/master/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationRequestFactory.java#L172)

**Context**
As a service provider we want to enforce the user is always authenticated while access some sensitive service so we would like to set the following values. 

<saml:AuthnContextClassRef> = urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport
ForceAuthn =true

jzheaux:
If these values are the same all the time, then the easiest way is to [register your own custom `AuthnRequestMarshaller` with OpenSAML](https://docs.spring.io/spring-security/site/docs/current/reference/html5/#servlet-saml2login-opensaml-customization).

Or, if the value needs to be changed on a per-request basis, you can [register your own `Saml2AuthenticationRequestContextResolver`](https://docs.spring.io/spring-security/site/docs/current/reference/html5/#servlet-saml2login-sp-initiated-factory-custom-authnrequest).

Will either of those work in your case?

adi-devops:
Awesome that would work for me. Maybe in the future if there are multiple requests we could expose it as a property that can be configured?

jzheaux:
Yes, @adi-devops, that's a good point. ForceAuthn and AuthnContextClassRef have been requested a few times now.

Could you elaborate on what you mean by "expose it as a property"? I'm not sure it's common enough for a Boot property, but I could imagine a scenario where adding properties to `Saml2AuthenticationRequestContext` could be helpful.

adi-devops:
[...]

PR #9209, opened 2020-11-18
enhancement to support overriding SAMLRequest

adi-devops:
This PR is to enhance the SAMLRequest by providing support to override a few of the parameters mentioned in the issue [GH-9199](https://github.com/spring-projects/spring-security/issues/9199).

**Support Added**

1. isPassive
2. ForceAuth
3. AuthnContextClassRef

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationRequestFactory.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationRequestFactory.java
@@ -35,11 +35,16 @@ import org.opensaml.core.config.ConfigurationService;
 import org.opensaml.core.xml.config.XMLObjectProviderRegistry;
 import org.opensaml.core.xml.io.MarshallingException;
 import org.opensaml.saml.common.xml.SAMLConstants;
+import org.opensaml.saml.saml2.core.AuthnContextClassRef;
+import org.opensaml.saml.saml2.core.AuthnContextComparisonTypeEnumeration;
 import org.opensaml.saml.saml2.core.AuthnRequest;
 import org.opensaml.saml.saml2.core.Issuer;
+import org.opensaml.saml.saml2.core.RequestedAuthnContext;
+import org.opensaml.saml.saml2.core.impl.AuthnContextClassRefBuilder;
 import org.opensaml.saml.saml2.core.impl.AuthnRequestBuilder;
 import org.opensaml.saml.saml2.core.impl.AuthnRequestMarshaller;
 import org.opensaml.saml.saml2.core.impl.IssuerBuilder;
+import org.opensaml.saml.saml2.core.impl.RequestedAuthnContextBuilder;
 import org.opensaml.saml.security.impl.SAMLMetadataSignatureSigningParametersResolver;
 import org.opensaml.security.SecurityException;
 import org.opensaml.security.credential.BasicCredential;
@@ -84,6 +89,8 @@ public class OpenSamlAuthenticationRequestFactory implements Saml2Authentication
 
 	private IssuerBuilder issuerBuilder;
 
+	private RequestedAuthnContextBuilder reqAuthnContextBuilder;
+
 	private Converter<Saml2AuthenticationRequestContext, String> protocolBindingResolver = (context) -> {
 		if (context == null) {
 			return SAMLConstants.SAML2_POST_BINDING_URI;
@@ -103,13 +110,16 @@ public class OpenSamlAuthenticationRequestFactory implements Saml2Authentication
 		this.authnRequestBuilder = (AuthnRequestBuilder) registry.getBuilderFactory()
 				.getBuilder(AuthnRequest.DEFAULT_ELEMENT_NAME);
 		this.issuerBuilder = (IssuerBuilder) registry.getBuilderFactory().getBuilder(Issuer.DEFAULT_ELEMENT_NAME);
+		this.reqAuthnContextBuilder = (RequestedAuthnContextBuilder) registry.getBuilderFactory()
+				.getBuilder(RequestedAuthnContext.DEFAULT_ELEMENT_NAME);
 	}
 
 	@Override
 	@Deprecated
 	public String createAuthenticationRequest(Saml2AuthenticationRequest request) {
 		AuthnRequest authnRequest = createAuthnRequest(request.getIssuer(), request.getDestination(),
-				request.getAssertionConsumerServiceUrl(), this.protocolBindingResolver.convert(null));
+				request.getAssertionConsumerServiceUrl(), this.protocolBindingResolver.convert(null), null, false,
+				false);
 		for (org.springframework.security.saml2.credentials.Saml2X509Credential credential : request.getCredentials()) {
 			if (credential.isSigningCredential()) {
 				X509Certificate certificate = credential.getCertificate();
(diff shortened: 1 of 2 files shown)


saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AuthenticationRequestContext.java
```diff
@@ -87,6 +98,32 @@ public String getRelayState() {
 		return this.relayState;
 	}
 
+	/**
+	 * Returns the desired ({@code RequestedAuthnContext} that this SP wishes to set in
+	 * the request.This property populates the {@code AuthNRequest.RequestedAuthnContext}
+	 * XML attribute.
+	 * @return the AuthnContextClassRef value
+	 */
+	public RequestedAuthnContext getRequestAuthnContext() {
```
jzheaux, line 107: We don't want any OpenSAML in the public API. 

Instead, I'd recommend changing this to a collection to `String` values and renaming the method to `getRequestAuthnContextRefs`.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AuthenticationRequestContext.java
```diff
@@ -87,6 +98,32 @@ public String getRelayState() {
...
+		return this.requestAuthnContext;
+	}
+
+	/**
+	 * Returns the isPassive value, if present in the parameters
+	 * @return the isPassive value, default is set to false
+	 */
+	public boolean getIsPassive() {
```
jzheaux, line 115: Please add a `@since 5.5` tag to each of the methods that you introduce.

Also, consider adding a link in the description that points to an authoritative source explaining what each of these values mean.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AuthenticationRequestContext.java
```diff
@@ -39,15 +40,25 @@
...
 
+	private final RequestedAuthnContext requestAuthnContext;
+
+	private final boolean isPassive;
+
+	private final boolean forceAuth;
+
 	protected Saml2AuthenticationRequestContext(RelyingPartyRegistration relyingPartyRegistration, String issuer,
-			String assertionConsumerServiceUrl, String relayState) {
+			String assertionConsumerServiceUrl, String relayState, RequestedAuthnContext requestAuthnContext,
```
jzheaux, line 50: Since this is a protected method, we can't change the signature; otherwise, it will break folks who upgrade. Please add a new constructor, leaving this constructor as-is.

I'd recommend the new constructor be `private` and have it take an instance of the `Builder`.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AuthenticationRequestContext.java
```diff
@@ -87,6 +98,32 @@ public String getRelayState() {
...
+	 */
+	public boolean getIsPassive() {
+		return this.isPassive;
+	}
+
+	/**
+	 * Returns the ForceAuthn value, if present in the parameters
+	 * @return this forceAuthn value, default is set to false
+	 */
+	public boolean getForceAuth() {
```
jzheaux, line 123: I think this should be called `getForceAuthn` since that's the name of the SAML attribute.

jzheaux (changes requested):
Thanks, @adi-devops! I've left some feedback inline.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AuthenticationRequestContext.java
adi-devops: Great point 👍 . I was also thinking of having it as a List of String, however, one challenge is we won't be able to set the `Comparison` value [AuthnContextComparisonTypeEnumeration](https://github.com/apigee/java-opensaml2/blob/master/src/main/java/org/opensaml/saml2/core/AuthnContextComparisonTypeEnumeration.java#L25). Shall we set it to `exact` as default?

jzheaux: Yes, the spec indicates that "exact" is the default.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AuthenticationRequestContext.java
```diff
@@ -87,6 +113,41 @@ public String getRelayState() {
 		return this.relayState;
 	}
 
+	/**
+	 * Returns the desired ({@code AuthnContextClassRef} that this SP wishes to set in the
```
jzheaux, line 117: While SP and "relying party" are both acceptable terms. For consistency with `RelyingPartyRegistration`, would you please use "relying party" instead?

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AuthenticationRequestContext.java
```diff
@@ -87,6 +113,41 @@ public String getRelayState() {
...
+
+	/**
+	 * Returns the isPassive value, if present in the parameters
+	 * @return the isPassive value, default is set to false
+	 * @since 5.5
+	 * @see <a href=
+	 * "https://wiki.shibboleth.net/confluence/display/SP3/isPassive">OpenSAML 3
+	 * isPassive</a>
+	 */
+	public boolean getIsPassive() {
```
jzheaux, line 135: Thinking about this just a bit more, it seems like there's value in having these use `Boolean` instead of `boolean`.

The reason for that is so that an application can tell Spring Security that it wants *no value* for an attribute by setting it to `null`.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AuthenticationRequestContext.java
```diff
@@ -164,14 +231,44 @@ public Builder relayState(String relayState) {
 			return this;
 		}
 
+		/**
+		 * Sets the {@code RequestedAuthnContext} parameter that will accompany this
+		 * AuthNRequest.
+		 * @param requestedAuthnContext the RequestedAuthnContext value.
+		 * @return this object
+		 */
+		public Builder requestAuthnContextRefs(List<String> requestAuthnContextRefs) {
```
jzheaux, line 240: Spring Security's builders tend to take a `Consumer` here instead of the list itself. That pattern allows for specifying a default list that the application can incrementally change.

Would you please change this to:

```java
public Builder requestAuthnContextRefs(Consumer<List<String>> requestAuthnContextRefsConsumer)
```

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AuthenticationRequestContext.java
```diff
@@ -164,14 +231,44 @@ public Builder relayState(String relayState) {
...
+			this.requestAuthnContextRefs = requestAuthnContextRefs;
+			return this;
+		}
+
+		/**
+		 * Sets the {@code ForceAuth} parameter that will accompany this AuthNRequest
+		 * @param forceAuth the foceAuth value. if null or empty, defaults to false.
+		 * @return this object
+		 */
```
jzheaux, line 249: Would you please add `@since 5.5` to the builder methods as well?

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationRequestFactory.java
```diff
@@ -160,17 +166,30 @@ public Saml2RedirectAuthenticationRequest createRedirectAuthenticationRequest(
...
 		AuthnRequest auth = this.authnRequestBuilder.buildObject();
 		auth.setID("ARQ" + UUID.randomUUID().toString().substring(1));
 		auth.setIssueInstant(new DateTime(this.clock.millis()));
-		auth.setForceAuthn(Boolean.FALSE);
-		auth.setIsPassive(Boolean.FALSE);
 		auth.setProtocolBinding(protocolBinding);
+		auth.setForceAuthn(forceAuth);
+		auth.setIsPassive(isPassive);
+		if (requestAuthnContextRefs != null) {
+			RequestedAuthnContext requestAuthnContext = new RequestedAuthnContextBuilder().buildObject();
```
jzheaux, line 182: Let's read the builders from `XMLObjectProviderFactorySupport`. That way, if the application has a custom builder, then this class will use it. You can see how this is done with the other builders in the constructor.

jzheaux (changes requested):
Thanks, @adi-devops, for the updates! I've left a bit more feedback inline.

Also, in preparation for merging, will you please squash your commits and in the resulting commit ensure that you have the phrase `Closes gh-9199` at the end?

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AuthenticationRequestContext.java
adi-devops: Yup, that is a great point.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AuthenticationRequestContext.java
adi-devops: That is an interesting partner did not know that 👍

jzheaux:
Blocked by #9277. Let's revisit once that PR is closed.

jzheaux:
Now that #9277 is merged, an application can do the following to set each of these properties:

First, `isPassive`:

```java
@Bean 
Saml2AuthenticationRequestResolver authenticationRequests(RelyingPartyRegistrationResolver registrations) {
    Saml2AuthenticationRequestResolver resolver = new OpenSaml4AuthenticationRequestResolver(registrations);
    resolver.setAuthnRequestCustomizer((params) -> params.getAuthnRequest().setIsPassive(Boolean.TRUE));
    return resolver;
}
```

Second, `ForceAuthn`:

```java
@Bean 
Saml2AuthenticationRequestResolver authenticationRequests(RelyingPartyRegistrationResolver registrations) {
    Saml2AuthenticationRequestResolver resolver = new OpenSaml4AuthenticationRequestResolver(registrations);
    resolver.setAuthnRequestCustomizer((params) -> params.getAuthnRequest().setForceAuthn(Boolean.TRUE));
    return resolver;
}
```

And third, `AuthContextClassRef`:

```java
@Bean 
Saml2AuthenticationRequestResolver authenticationRequests(RelyingPartyRegistrationResolver registrations) {
    AuthnContextClassRef ref = // .. construct with OpenSAML
    Saml2AuthenticationRequestResolver resolver = new OpenSaml4AuthenticationRequestResolver(registrations);
    resolver.setAuthnRequestCustomizer((params) -> params.getAuthnRequest()
            .getRequestedAuthnContext().getAuthnContextClassRefs().add(ref));
    return resolver;
}
```

Because of this, I think that we don't need to introduce additional fields into `Saml2AuthenticationRequestContext` et al.

Thank you so much for your efforts. I'm going to close this; please feel free to reopen if you feel this doesn't address the issue.

PR #9211, opened 2020-11-20
getRemoteUser() returns principal name

### LINKED ISSUE #3357: SEC-3198: SecurityContextHolderAwareRequestWrapper#getRemoteUser ignores Authenticaion#getName
[Michael Osipov](https://jira.spring.io/secure/ViewProfile.jspa?name=michael-o) (Migrated from [SEC-3198](https://jira.spring.io/browse/SEC-3198?redirect=false)) said:

The public method `getRemoteUser` tries to `return auth.getPrincipal().toString()` but there is no guarantee that the principal which is an arbitrary implementation will acttually return the remote user. The assumption that it does is wrong. Since `Authentication` implements `Principal`, it would be the safest bet to `return auth.getName()` which is by interface guaranteed to be a proper name.

sjrd218:
Returns the name of the authenticated principle instead of falling through to the toString() method which may render a string representation of the entire object rather than a username.

This behavior is helpful in OAuth2 and Saml2 configurations.

Closes #3357

### DIFF
--- a/web/src/main/java/org/springframework/security/web/servletapi/SecurityContextHolderAwareRequestWrapper.java
+++ b/web/src/main/java/org/springframework/security/web/servletapi/SecurityContextHolderAwareRequestWrapper.java
@@ -22,6 +22,7 @@ import java.util.Collection;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletRequestWrapper;
 
+import org.springframework.security.authentication.AbstractAuthenticationToken;
 import org.springframework.security.authentication.AuthenticationTrustResolver;
 import org.springframework.security.authentication.AuthenticationTrustResolverImpl;
 import org.springframework.security.core.Authentication;
@@ -105,6 +106,9 @@ public class SecurityContextHolderAwareRequestWrapper extends HttpServletRequest
 		}
 		if (auth.getPrincipal() instanceof UserDetails) {
 			return ((UserDetails) auth.getPrincipal()).getUsername();
+		}
+		if (auth.getPrincipal() instanceof AbstractAuthenticationToken) {
+			return ((AbstractAuthenticationToken) auth.getPrincipal()).getName();
 		}
 		return auth.getPrincipal().toString();
 	}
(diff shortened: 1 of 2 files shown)


sjrd218:
Closing this PR. I can't run the integration tests on my computer. They always fail. Feel free to use the code if you wish. It takes too much time to try to submit PRs to this project and I can't justify spending any more time trying to make this work.

jzheaux:
@sjrd218, I'm sorry things didn't work well for you. If you are still interested, I wonder if the slow-down you are experiencing is something that I can help with.

Would you be able to post details about the issues that you were having? For example, what were the integration test failures, and what was it that was taking too much time? 

There might be solutions to both, and it would be helpful for others who experience the same issues that you did.

jzheaux:
@sjrd218 after reviewing the PR, I believe it's still a valuable contribution. 9c373ef4f80e2e159e7f53719763ddf0ab78c204 and a polish of f614a8230c84a505597de0bd6380e5e2fea117ea are now merged into `master`.

lnxmad:
This is a showstopper for my team.. Any way to release this bugfix sooner than 5.5.0.M2?

jzheaux:
Hi, @lnxmad. There are no releases planned before `5.5.0-M2`.

Can your problem be addressed by using a custom principal whose `toString` outputs what you need? Or, you can introduce a custom filter that wraps the `HttpServletRequest` with a `HttpServletRequestWrapper` implementation that overrides the `getRemoteUser` method.

If neither of those work, then sharing more detail about your issue might help.

lnxmad:
I did consider a filter to wrap request, however `AuthenticationTrustResolver` is used and doesn't appear to be a bean to inject. I'd have to create a new instance (may not be an issue). As far as the principal object, I am using out of the box OIDC and haven't tracked down a good place to create new principal to extend/override the toString(). I can take another look at that.

Thanks for the quick response!

jzheaux:
You might consider injecting a custom `OidcUserService`. Your custom one can delegate to the default one and then change out the principal, like so:

```java
@Bean 
OAuth2UserService<OidcUserRequest, OidcUser> oidcUserService() {
    OidcUserService delegate = new OidcUserService(); // the default
    return (userRequest) -> {
        OidcUser oidcUser = delegate.loadUser(userRequest);
        return new MyCustomOidcUser(oidcUser);
    }
}
```

Issue #9212, opened 2020-11-21
Better meaning error when principalClaimName is set and claim value is not a String


omlip:
**Context**

When configuring an Spring application as an OAuth Resource Server and we use Jwt, we can set manually the `principalClaimName` on JwtAuthenticationConverter object.

Example like this
```kotlin
fun getCustomAuthenticationConverter(authoritiesClaimName : String): Converter<Jwt, Mono<AbstractAuthenticationToken>> {
    return ReactiveJwtAuthenticationConverterAdapter(
            JwtAuthenticationConverter().apply {
                setPrincipalClaimName("anInt")
                setJwtGrantedAuthoritiesConverter(GrantedAuthoritiesExtractor(authoritiesClaimName))
            }
    )
}
```

**Current Behavior**

If we use a specific `principalClaimName` on a JwtAuthenticationConverter and if the value behind the claim name is not a String, a ClassCastException is thrown in the convert method.

```java
	@Override
	public final AbstractAuthenticationToken convert(Jwt jwt) {
		Collection<GrantedAuthority> authorities = extractAuthorities(jwt);
		if (this.principalClaimName == null) {
			return new JwtAuthenticationToken(jwt, authorities);
		}
		String name = jwt.getClaim(this.principalClaimName); // this line throw a ClassCastException
		return new JwtAuthenticationToken(jwt, authorities, name);
	}
```

We can easily reproduce the situation by adding a test in class `JwtAuthenticationConverterTests`.

```java
	@Test
	public void convertWhenPrincipalClaimNameSetWithOtherValueThanString() {
		this.jwtAuthenticationConverter.setPrincipalClaimName("user_id");
		Jwt jwt = TestJwts.jwt().claim("user_id", 100).build();
                //TODO: complete with assertions
	}
```

**Expected Behavior**

I think we should adapt the code to not lead to a ClassCastException but maybe throw a proper exception which "explain" what the problem is.

Maybe by adapting like this ?
```java
Object claimValue = jwt.getClaim(this.principalClaimName);
Assert.isInstanceOf(String.class, claimValue, "The principal value is not a String");

String name = (String)claimValue;
```

or maybe with the usage of Jwt.claimAsString(...) method ?

What do you think ? 
Let me know if I can help.

[...]

PR #9215, opened 2020-11-23
Avoid ClassCastException if principalClaim value is not a …

omlip:
Fix https://github.com/spring-projects/spring-security/issues/9212

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationConverter.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationConverter.java
@@ -29,6 +29,7 @@ import org.springframework.util.Assert;
  * @author Rob Winch
  * @author Josh Cummings
  * @author Evgeniy Cheban
+ * @author Olivier Antoine
  * @since 5.1
  */
 public class JwtAuthenticationConverter implements Converter<Jwt, AbstractAuthenticationToken> {
@@ -43,8 +44,8 @@ public class JwtAuthenticationConverter implements Converter<Jwt, AbstractAuthen
 		if (this.principalClaimName == null) {
 			return new JwtAuthenticationToken(jwt, authorities);
 		}
-		String name = jwt.getClaim(this.principalClaimName);
-		return new JwtAuthenticationToken(jwt, authorities, name);
+		String principalClaimValue = jwt.getClaimAsString(this.principalClaimName);
+		return new JwtAuthenticationToken(jwt, authorities, principalClaimValue);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationConverter.java
```diff
@@ -43,8 +44,9 @@ public final AbstractAuthenticationToken convert(Jwt jwt) {
 		if (this.principalClaimName == null) {
 			return new JwtAuthenticationToken(jwt, authorities);
 		}
-		String name = jwt.getClaim(this.principalClaimName);
-		return new JwtAuthenticationToken(jwt, authorities, name);
+		Object principalClaimValue = jwt.getClaim(this.principalClaimName);
```
jzheaux, line 47: Would it be better to call `jwt.getClaimAsString(this.principalClaimName)`? The benefit is that the converter would work for principals of other types. I think this aligns with how `AbstractAuthenticationToken#getName` also coerces the principal into a string.

cc @jgrandja

omlip: @jzheaux I initially submit this issue to avoid a ClassCastException and I suggested both ideas to achieve this.
We went for the other one.

After this, I agree with you that we should keep the alignment with JwtClaimAccessor#getSubject method by using jwt.getClaimAsString(this.principalClaimName). It will also reduce the situation where we can have a ClassCastException.

I can change the code, just let me know what to do 👍

jzheaux: Makes sense, @omlip.

Yes, let's go ahead and make the change to `getClaimAsString`. This is also consistent with `DefaultOAuth2User#getName` which coerces the configured claim into a `String`.

omlip: ok thanks, will adapt soon.

omlip:
@jzheaux It seems I did something wrong using Git (I rebase against master) this is why we have all the commits in this.

I guess it is not so evident to review this actually. Do we leave this PR as it is or I create a new one and we close/archive this one ?

jzheaux:
I think if you do a rebase, it will resolve the issue. If we imagine your fork is called `origin` and this is called `spring-projects`, you can do the following:

```bash
git checkout master
git reset --hard spring-projects/master
git pull spring-projects master
git checkout issues/9212
git rebase master
git push origin +issues/9212
```

jzheaux:
Thanks, @omlip! In preparation for merging, will you please squash your four commits into one and ensure that the resulting commit message contains the phrase "Closes gh-9212"?

omlip:
Sure

jzheaux:
Thanks, @omlip! This is now merged into `master`.

PR #9219, opened 2020-11-25
Fix javadoc in Pbkdf2PasswordEncoder

rikles:
Fix javadoc in Pbkdf2PasswordEncoder

- Add `@author Loïc Guibert`
- Add `@since 5.5` to new constructors

See PR gh-9147


jzheaux:
Thanks, @rikles, for the polish. This is now merged into `master`.

Issue #9210, opened 2020-11-20
OidcIdToken cannot be serialized to JSON if token contains claim of type JSONArray


mengelbrecht:
**Describe the bug**
If an IdP sends an ID token with claim `amr`, the Jackson `ObjectMapper` with `SecurityJackson2Modules` cannot serialize the ID token to JSON (related: https://github.com/spring-projects/spring-security/issues/4370).
The `amr` claim in the ID token has the type `com.nimbusds.jose.shaded.json.JSONArray` for which there is no default mixin.

Tested with Spring-Security 5.4.1.

**To Reproduce**
These steps resemble a normal `oauth2Login` configuration where additionally the ID token is serialized to JSON.
1.  Include an `amr` claim in the ID token
2. Decode the string token value using an `JwtDecoder` created by `OidcIdTokenDecoderFactory` to a `Jwt`.
3. Create an `OidcIdToken` from the `Jwt`.
4. Serialize the `OidcIdToken` to a JSON string using an `ObjectMapper` with the `SecurityJackson2Modules`.

**Expected behavior**
The `amr` claim should be an `ArrayList` instead of `JSONArray`.

**Workaround**
Define a mixin for the `JSONArray` class.

jgrandja:
Thanks for the report @mengelbrecht. The default converter for the `amr` claim is `Collection<String>` in [OidcIdTokenDecoderFactory.createDefaultClaimTypeConverters()](https://github.com/spring-projects/spring-security/blob/2abf59b695b3ad14719299ed17ff47b181eed802/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcIdTokenDecoderFactory.java#L112).

Can you put together a test that reproduces the issue as I'm not seeing how the `amr` claim is decoded as a `com.nimbusds.jose.shaded.json.JSONArray`.

mengelbrecht:
I could not create a test because the JWT has to be signed and the validator wants to fetch the jwks which fails in my test.

After a little more digging I could reproduce it using this Kotlin snippet. The code outputs `com.nimbusds.jose.shaded.json.JSONArray` as type for the converted object.
```kotlin
    val jsonArray = com.nimbusds.jose.shaded.json.JSONArray().apply { add("test") }
    val converted = ClaimConversionService.getSharedInstance().convert(
        jsonArray,
        TypeDescriptor.valueOf(Any::class.java),
[...]

Issue #4498, opened 2017-08-16
Support client authentication using X.509 certificate


jgrandja:

myspri:
Hello @jgrandja, I have been talking to some security experts(OpenID/OAuth2 spec writers)  and got the feedback that using MTLS is the most desirable form of client authentication.

https://tools.ietf.org/html/draft-ietf-oauth-mtls-03

Was wondering if this is still in your radar.  
Thanks!

jgrandja:
@myspri Apologize for the delayed response. Yes, we are still planning on implementing this feature. However, we have not scheduled it as of yet.

hablutzel1:
Note that the spec has became the standard RFC 8705, "OAuth 2.0 Mutual-TLS Client Authentication and Certificate-Bound Access Tokens".

In the other hand, there are already some open source implementations like the one from IdentityServer4 (see  https://github.com/IdentityServer/IdentityServer4/tree/master/samples/Clients/src/ConsoleMTLSClient).

maskedpirate:
Any updates on if this has been scheduled?

rwinch:
@maskedpirate No updates. Would you be interested in submitting a pull request?

praveenbommali:
any update !!!

greghall76:
I'd love to here an update on this. I'm interested in running in OpenShift w/ their internal OAuth server. From looking at their oauth-proxy sidecar, I believe OpenShift's OAuth server is expecting a k8s service account client crt as the auth method.

rwinch:
No updates at this time.

magnuskkarlsson:
I think I got it to work in Spring Boot 3

All feedback are welcome!

You need to set 

spring.security.oauth2.client.registration.keycloak.client-authentication-method = none

so you send client_id in request body, otherwise will the Authorization Server never know from which client the request comes from.

Then add custom truststore and keystore for accessTokenResponseClient. The solution is quite long since Apache HttpComponents HttpClient, ignores standard Java truststore (-Djavax.net.ssl.trustStore) and keystore (-Djavax.net.ssl.keyStore). Otherwise we would be done by only setting client-authentication-method = none

`    @Bean
    public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
        http //
...
                .oauth2Client(oauth2Client -> oauth2Client //
                        .authorizationCodeGrant(authorizationCodeGrant -> authorizationCodeGrant //
[...]

PR #9222, opened 2020-11-27
OidcIdToken cannot be serialized to JSON if token contains claim of type JSONArray or JSONObject

ovidiupopa91:
ObjectToListStringConverter and ObjectToMapStringObjectConverter were checking if the source object is of type List or Map and if the first element or key is a String. If we have a JSONArray containing Strings the above check will pass, meaning that a JSONArray will be returned which is not serializable (same applies to JSONObject)

With this change, even if the check is passing a new List or Map will be returned.

Closes gh-9210

### DIFF
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/converter/ObjectToListStringConverter.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/converter/ObjectToListStringConverter.java
@@ -56,12 +56,6 @@ final class ObjectToListStringConverter implements ConditionalGenericConverter {
 		if (source == null) {
 			return null;
 		}
-		if (source instanceof List) {
-			List<?> sourceList = (List<?>) source;
-			if (!sourceList.isEmpty() && sourceList.get(0) instanceof String) {
-				return source;
-			}
-		}
 		if (source instanceof Collection) {
 			Collection<String> results = new ArrayList<>();
 			for (Object object : ((Collection<?>) source)) {
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/converter/ObjectToMapStringObjectConverter.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/converter/ObjectToMapStringObjectConverter.java
@@ -53,9 +53,6 @@ final class ObjectToMapStringObjectConverter implements ConditionalGenericConver
 			return null;
 		}
 		Map<?, ?> sourceMap = (Map<?, ?>) source;
-		if (!sourceMap.isEmpty() && sourceMap.keySet().iterator().next() instanceof String) {
-			return source;
-		}
 		Map<String, Object> result = new HashMap<>();
 		sourceMap.forEach((k, v) -> result.put(k.toString(), v));
 		return result;
--- a/oauth2/oauth2-core/src/test/java/org/springframework/security/oauth2/core/converter/ClaimConversionServiceTests.java
+++ b/oauth2/oauth2-core/src/test/java/org/springframework/security/oauth2/core/converter/ClaimConversionServiceTests.java
@@ -26,6 +26,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import com.nimbusds.jose.shaded.json.JSONArray;
+import com.nimbusds.jose.shaded.json.JSONObject;
 import org.assertj.core.util.Lists;
 import org.junit.Test;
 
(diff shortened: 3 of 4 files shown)


oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/converter/ObjectToMapStringObjectConverter.java
```diff
@@ -54,7 +54,7 @@ public Object convert(Object source, TypeDescriptor sourceType, TypeDescriptor t
 		}
 		Map<?, ?> sourceMap = (Map<?, ?>) source;
 		if (!sourceMap.isEmpty() && sourceMap.keySet().iterator().next() instanceof String) {
```
ovidiupopa91, line 56: In my opinion this if block can be removed, as it will be covered by ```Map<String, Object> result = new HashMap<>();
		sourceMap.forEach((k, v) -> result.put(k.toString(), v));``` and a new Map should be created in both cases.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/converter/ObjectToListStringConverter.java
```diff
@@ -59,7 +59,7 @@ public Object convert(Object source, TypeDescriptor sourceType, TypeDescriptor t
 		if (source instanceof List) {
```
ovidiupopa91, line 59: In my opinion this check can be removed, as it will be covered by ```if (source instanceof Collection) ``` where a new ArrayList is created

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/converter/ObjectToListStringConverter.java
```diff
@@ -59,7 +59,7 @@ public Object convert(Object source, TypeDescriptor sourceType, TypeDescriptor t
...
 			List<?> sourceList = (List<?>) source;
 			if (!sourceList.isEmpty() && sourceList.get(0) instanceof String) {
-				return source;
+				return new ArrayList<String>((List<? extends String>) sourceList);
```
mengelbrecht, line 62: What happens if the source is a `List<Object>` where the first element is a `String` but the others are not? This probably applies to the `ObjectToMapStringObjectConverter` as well.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/converter/ObjectToListStringConverter.java
mengelbrecht: Also the code current code does not cover the case of  a `List<Object>` where the first element is a String but the other elements are not. If this code block is removed it will be handled correctly by the `if (source instanceof Collection)` block which converts all elements to a `String`.

ovidiupopa91: I agree. This is what I did in my initial implementation, but I wanted to double check if someone else is sharing the same opinion. I will apply the stash which contains this.

jgrandja:
Thanks for the PR @ovidiupopa91 ! This is now in master and backported to 5.2.x.

jgrandja:
@ovidiupopa91 If you're available for another PR in Spring Security or Spring Authorization Server let me know and I'll find one for you.

ovidiupopa91:
Hi @jgrandja . Sure, I would be happy to contribute to a new PR. Either of them would be fine.

jgrandja:
Thanks @ovidiupopa91 ! I will be in touch next week with a couple of options to choose from . Enjoy your weekend!

jgrandja:
@ovidiupopa91 Here are a couple of options to choose from:

**OAuth 2.0 Mutual-TLS Client Authentication**
Client [spring-security#4498](https://github.com/spring-projects/spring-security/issues/4498)
Server [spring-authorization-server#101](https://github.com/spring-projects-experimental/spring-authorization-server/issues/101)

**OpenID Connect Dynamic Client Registration 1.0**
[spring-authorization-server#57](https://github.com/spring-projects-experimental/spring-authorization-server/issues/57)

Do any of these features interest you?

Keep in mind that these are major features and will require some effort and time.
I'm not sure what your free time is like but keep this in mind when deciding.

If you prefer a smaller feature let me know and I'll find something.

FYI, I'll be on PTO starting Dec 11 and returning Jan 4.

ovidiupopa91:
Hi @jgrandja. I can give it a try with 

**OpenID Connect Dynamic Client Registration 1.0**
[spring-authorization-server#57](https://github.com/spring-projects-experimental/spring-authorization-server/issues/57)

When it comes to my free time, I'm usually allocating 2-3hours/day. I don't have a problem regarding the effort, but it also depends when this feature is planned to be rolled out. If this must happen in the next 1-2 weeks I'm not sure if I can finish till then. 

Let me know if this is ok with you too. Maybe we can have a chat before you leave, discussing this feature before actually starting to implement it. (we can do it on gitter) 

I will also be on PTO starting Dec 21 and returning on Jan 6.

Issue #9115, opened 2020-10-08
Add NameIdFormat support to RelyingPartyRegistration


Gayathri92:
`RelyingPartyRegistration` contains [SAML 2.0 Metadata](https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf) for both the relying party and its associated asserting party.

It would be nice to include `NameIDFormat` support in both `RelyingPartyRegistration` and `RelyingPartyRegistration.AssertingPartyDetails`. The application could then communicate a set of preferred formats to the asserting party (with #9297, for example) as well as lookup the asserting party's preferred formats.

jzheaux:
Thanks for reaching out, @Gayathri92. Would you please add some more detail about your use case?

spring-issuemaster:
If you would like us to look at this issue, please provide the requested information. If the information is not provided within the next 7 days this issue will be closed.

spring-issuemaster:
Closing due to lack of requested feedback. If you would like us to look at this issue, please provide the requested information and we will re-open the issue.

amergey:
@jzheaux I saw your mention, I will try to submit a PR for this

amergey:
We have some customers where specifying NameIdFormat is mandated by their IDP (or the way they configure it). It seems quite common with PingFederate (but we have that we other as well).

NameIdFormat is a part of NameIDPolicy which support others attributes
SAML Core specification:

jzheaux:
Good points, @amergey. I think I created some confusion by not first updating the description in this ticket. It's updated now.

The reason I'd like to leave changes to `OpenSamlAuthenticationRequestFactory` out of this PR is I'd like to see what happens with https://github.com/spring-projects/spring-security/pull/9277 first since that may obviate the need to change the factory.

Would you be able to update the PR to match the issue's description?

amergey:
@jzheaux Do you mean updating this PR by keeping only nameid format ? (so remove other part of nameid policy but keeping related work on metadata ?) . Also correct me if I am wrong but it seems to me NameID format is defined at SP level, there is no IDP metadatas related to NameID format.

jzheaux:
Sorry for the delay, @amergey, I was on holiday.

> Do you mean updating this PR by keeping only nameid format

Yes, since that's the only one that appears in metadata.

> NameID format is defined at SP level, there is no IDP metadatas related to NameID format.

[...]

PR #9227, opened 2020-12-01
make SP NameIDPolicy configurable in RelyingPartyRegistration

amergey:
closes gh-9115


jzheaux:
Thanks for the PR, @amergey! This PR seems broader than the scope of #9115, so I think we first should take a step back and make sure that we're on the same page about what support should go into the framework and where.

To start that, will you please add some detail to #9115 about your use case?

amergey:
I have updated #9115, I am not the reporter so, I add a comment, I cannot edit the issue.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
```diff
@@ -81,6 +81,8 @@
 
 	private final Saml2MessageBinding assertionConsumerServiceBinding;
 
+	private final String nameIDFormat;
```
jzheaux, line 84: For consistency, will you please lowercase acronyms like so: `nameIdFormat`. Please do the same for the associated getter and setter.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
```diff
@@ -177,6 +180,15 @@ public Saml2MessageBinding getAssertionConsumerServiceBinding() {
 		return this.assertionConsumerServiceBinding;
 	}
 
+	/**
+	 * Get the NameID format.
+	 * @return the NameID format
+	 * @since 5.5
```
jzheaux, line 186: Please update `@since` flags to 5.7

jzheaux (changes requested):
@amergey, thanks for your patience on this review. I've left some feedback inline.

Would you please update the copyright year on any edited files and also squash your commits?

We can schedule this for 5.7.0-M1.

amergey:
I tried to made requested changes, but this PR is quite old already and I have trouble to rebase it to main, it would have been better to make these suggestions last year ....

jzheaux:
>  it would have been better to make these suggestions last year ....

I'm sorry it took so long to get back to this. I corrected the rebase and this is now merged via `a17dfb8456cf3a517d44cfcccdaa7a1d2ae6121f` for 6.0 and `dbe4d704f844d7d3a4b7048e004c07e1496033d2` for 5.7.

Issue #9256, opened 2020-12-03
Exception when declaring multiple AuthenticationManager beans


ibaskine:
**Describe the bug**
After upgrading to Spring Security 5.4.1 my web application throws the following exception:

org.springframework.beans.factory.NoUniqueBeanDefinitionException: No qualifying bean of type 'org.springframework.security.authentication.AuthenticationManager' available: expected single matching bean but found 4: org.springframework.security.config.authentication.AuthenticationManagerFactoryBean#0,org.springframework.security.authentication.ProviderManager#0,org.springframework.security.authenticationManager,stdProviderManager at org.springframework.beans.factory.config.DependencyDescriptor.resolveNotUnique(DependencyDescriptor.java:220) at org.springframework.beans.factory.support.DefaultListableBeanFactory.doResolveDependency(DefaultListableBeanFactory.java:1345) at org.springframework.beans.factory.support.DefaultListableBeanFactory.resolveDependency(DefaultListableBeanFactory.java:1287) at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor$AutowiredMethodElement.inject(AutowiredAnnotationBeanPostProcessor.java:714)

I traced the problem to a new class org.springframework.security.config.annotation.web.configuration.HttpSecurityConfiguration. As this class annotated with @Configuration Spring tries to create corresponding bean definition and to autowire AuthenticationManager to it. As my application has multiple AuthenticationManager beans this attempt fails.

My application doesn't need HttpSecurityConfiguration bean, but I could not filter it out from component scab as the class if of package level so I cannot add reference to it to filter annotation.

**To Reproduce**
Add a second AuthenticatioManager to any Web Application using Spring Security.

**Expected behavior**
As a minimum there shall be a way of excluding HttpSecurityConfiguration from component scan

**Sample**

A link to a GitHub repository with a [minimal, reproducible sample](https://stackoverflow.com/help/minimal-reproducible-example).

Reports that include a sample will take priority over reports that do not.
At times, we may require a sample, so it is good to try and include a sample up front.

eleftherias:
Thanks @ibaskine.

Which version are you upgrading from?

Could you share your security configuration and the full stack trace?

ibaskine:
@eleftherias Thank you for taking look at it. I am upgrading from 5.3.4.RELEASE. 

[...]

PR #9264, opened 2020-12-05
Qualify reference to AuthenticationManager bean in HttpSecurityConfiguration

jcihocki:
Quick reference fix for gh-9256, if you're interested.

https://github.com/spring-projects/spring-security/issues/9256


eleftherias:
Thanks for the suggestion @jcihocki, but we can't guarantee that users will be using the default Bean ID.
They may be exposing an AuthenticationManager with a different Bean ID, in which case this wouldn't work.

Issue #9275, opened 2020-12-10
Current build failure due to com.nimbusds:oauth2-oidc-sdk:8.28.1


fhanik:
Current build generates an error as it seems that the latest (8.28.1) `oauth2-oidc-sdk` artifact may be corrupt

> management "com.nimbusds:oauth2-oidc-sdk:8.27"

Temporarily resolves the issue

**Detailed Stack Trace**

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':spring-security-oauth2-core:compileJava'.
> Could not resolve all files for configuration ':spring-security-oauth2-core:optional'.
   > Could not resolve com.nimbusds:oauth2-oidc-sdk:latest.release.
     Required by:
         project :spring-security-oauth2-core
      > Failed to list versions for com.nimbusds:oauth2-oidc-sdk.
         > Unable to load Maven meta-data from https://repo.spring.io/libs-snapshot/com/nimbusds/oauth2-oidc-sdk/maven-metadata.xml.
            > Could not HEAD 'https://repo.spring.io/libs-snapshot/com/nimbusds/oauth2-oidc-sdk/maven-metadata.xml'. Received status code 401 from server: Unauthorized

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/6.7.1/userguide/command_line_interface.html#sec:command_line_warnings

BUILD FAILED in 1m 7s
105 actionable tasks: 105 executed

A build scan cannot be produced as an error occurred gathering build data.
Please report this problem via https://ge.spring.io/help and include the following via copy/paste:

----------
Gradle version: 6.7.1
Plugin version: 3.2

org.gradle.api.artifacts.ResolveException: Could not resolve all dependencies for configuration ':spring-security-oauth2-core:optional'.
        at org.gradle.api.internal.artifacts.ivyservice.ErrorHandlingConfigurationResolver.wrapException(ErrorHandlingConfigurationResolver.java:104)
        at org.gradle.api.internal.artifacts.ivyservice.ErrorHandlingConfigurationResolver.access$000(ErrorHandlingConfigurationResolver.java:48)
[...]

PR #9276, opened 2020-12-10
LdapAuthoritiesPopulator should be postProcessed

fhanik:
Currently we can set an `LdapAuthoritiesPopulator` and override the default one. However, the current implementation doesn't allow us to create a delegate to the default populator, because it isn't post processed

This simple fix will allow an easier customization of the `LdapAuthoritiesPopulator`


fhanik:
Note: I ran into this [issue creating](https://github.com/spring-projects/spring-security/issues/9275) this PR and used a temporary work around for my local build

https://github.com/spring-projects/spring-security/issues/9275

jzheaux (changes requested):
Thanks for the PR, @fhanik! Will you please update the copyright on these classes so that the end date is `2020`?

There is a checkstyle error in the build. I realize that locally you might not have gotten that given the `401` issue you were experiencing earlier in the build, but I think if you try again, you'll see that error now.

jzheaux:
@fhanik, are you able to make the requested changes?

jzheaux:
Thanks, @fhanik this is now merged via ae0860801146b860586280594d7fc6431b766b18. I also added a polish to simplify the test, change the copyright date, and address the checkstyle errors.

Issue #9205, opened 2020-11-16
Match requests based on servlet dispatcher type


eleftherias:
Add a matcher to match requests and apply security rules based on the servlet Dispatcher type (FORWARD,INCLUDE,REQUEST,ASYNC,ERROR).

**Context**

This will allow users to define what happens in the case of an ERROR dispatch. 
This may help in cases such as gh-9181.

nmck257:
Created PR #9278 - let me know if I missed any functionality?

PR #9278, opened 2020-12-11
Introduced DispatcherType request matcher

nmck257:
Created a DispatcherTypeRequestMatcher and corresponding methods
for configuring an HttpSecurity object. This enables filtering of
security rules based on the dispatcher type of the incoming servlet
request.

Additionally, removed duplicative code for parsing HttpMethod objects
from the AntPathRequestMatcher and RegexRequestMatcher.

Closes gh-9205

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java
@@ -20,14 +20,18 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
+import javax.servlet.DispatcherType;
+
 import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.context.ApplicationContext;
 import org.springframework.http.HttpMethod;
+import org.springframework.lang.Nullable;
 import org.springframework.security.config.annotation.ObjectPostProcessor;
 import org.springframework.security.config.annotation.web.configurers.AbstractConfigAttributeRequestMatcherRegistry;
 import org.springframework.security.web.servlet.util.matcher.MvcRequestMatcher;
 import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
 import org.springframework.security.web.util.matcher.AnyRequestMatcher;
+import org.springframework.security.web.util.matcher.DispatcherTypeRequestMatcher;
 import org.springframework.security.web.util.matcher.RegexRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher;
 import org.springframework.util.Assert;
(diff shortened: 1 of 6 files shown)


nmck257:
FYI - checks are passing now.

(sorry for the noise -- struggling to get my local checkstyle/checkformat config to match up w/ the project)

config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java
```diff
@@ -83,7 +87,7 @@ public C anyRequest() {
 	 * @return the object that is chained after creating the {@link RequestMatcher}
 	 */
 	public C antMatchers(HttpMethod method) {
-		return antMatchers(method, new String[] { "/**" });
```
eleftherias, line 86: Please move any changes that are not directly related to this issue to a different pull request or a different commit within this pull request.

config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java
```diff
@@ -206,6 +210,35 @@ public C regexMatchers(String... regexPatterns) {
...
+	 * {@link org.springframework.security.web.util.matcher.DispatcherTypeRequestMatcher}
+	 * instances.
+	 * @param method the {@link HttpMethod} to use or {@code null} for any
+	 * {@link HttpMethod}.
+	 * @param dispatcherTypes the dispatcher types to match against
+	 * @return the object that is chained after creating the {@link RequestMatcher}
+	 */
+	public C dispatcherTypeMatchers(HttpMethod method, DispatcherType... dispatcherTypes) {
+		Assert.state(!this.anyRequestConfigured, "Can't configure dispatcherTypeMatchers after anyRequest");
+		return chainRequestMatchers(Arrays.stream(dispatcherTypes)
```
eleftherias, line 224: Please use a for-loop instead of a Stream. 
The GC overhead of creating additional objects (including intermediate objects) can cause significant decrease in performance.
See #7154 for more details.

eleftherias (changes requested):
Thanks @nmck257!
I have left some feedback inline.

config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java
nmck257: Okay -- I'll split this (and the change from a private static `valueOf` method to `HttpMethod::resolve`) into a separate commit

eleftherias:
Thanks @nmck257!

Are you interested in submitting a similar PR for the server equivalent?
This would involve creating a `DispatcherTypeServerWebExchangeMatcher` that implements `ServerWebExchangeMatcher`.

nmck257:
> Are you interested in submitting a similar PR for the server equivalent?
> This would involve creating a `DispatcherTypeServerWebExchangeMatcher` that implements `ServerWebExchangeMatcher`.

Sure @eleftherias -- I should have time to do so a little later this week.

nmck257:
@eleftherias - want to make sure I understand the intent for the `DispatcherTypeServerWebExchangeMatcher`.

The `ServerWebExchange` and `ServerHttpRequest` interfaces don't expose a `DispatcherType` directly.

But, I do see that if a request is adapted through `ServletHttpHandlerAdapter` into a `ServletServerHttpRequest`, then that particular `ServerHttpRequest` implementation has an `HttpServletRequest` field which will preserve the `DispatcherType` from the servlet context. And that field is available through `AbstractServerHttpRequest::getNativeRequest` with some casting.

So... is that the scope? The new matcher should check whether the incoming request is an `instanceof` `ServletServerHttpRequest`, and if so, pull back the `DispatcherType` from the request's servlet background for the comparison? And non-match requests which aren't `ServletServerHttpRequest`?

eleftherias:
Thanks for pointing that out @nmck257.

I was too quick to assume that there should be a server equivalent, when in fact this doesn't make sense for server based applications.
That means we don't need the additional PR.

Sorry about the confusion and thank you for setting us on the right track.

nmck257:
Got it - no worries.

Thanks @eleftherias !

simaotwx:
> I was too quick to assume that there should be a server equivalent, when in fact this doesn't make sense for server based applications.

What do you mean by this? The PR says

> This enables filtering of security rules based on the dispatcher type of the incoming servlet request.

"incoming servlet request" → server

Why is this not for server? I don't understand that.
Also, does this work in Kotlin using the DSL?

Issue #9284, opened 2020-12-16
Consider Aligning MvcRequestMatcher's matching methods


jzheaux:
See https://github.com/spring-projects/spring-security/pull/8996#issuecomment-746312303

evgeniycheban:
@jzheaux I can take this.

PR #9300, opened 2020-12-19
Consider Aligning MvcRequestMatcher's matching methods

evgeniycheban:
Closes gh-9284

### DIFF
--- a/web/src/main/java/org/springframework/security/web/servlet/util/matcher/MvcRequestMatcher.java
+++ b/web/src/main/java/org/springframework/security/web/servlet/util/matcher/MvcRequestMatcher.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2019 the original author or authors.
+ * Copyright 2012-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


web/src/main/java/org/springframework/security/web/servlet/util/matcher/MvcRequestMatcher.java
```diff
@@ -95,6 +96,12 @@ private MatchableHandlerMapping getMapping(HttpServletRequest request) {
 
 	@Override
 	public MatchResult matcher(HttpServletRequest request) {
+		if (this.method != null && !this.method.name().equals(request.getMethod())) {
```
jzheaux, line 99: Instead of duplicating the lines, would it be simpler to have `matches(request)` and `matcher(request)` call the same underlying private method?

jzheaux (changes requested):

web/src/main/java/org/springframework/security/web/servlet/util/matcher/MvcRequestMatcher.java
evgeniycheban: Done.

evgeniycheban:
@jzheaux Something is wrong with the PR build.
> Could not GET 'https://repo.spring.io/libs-snapshot/com/fasterxml/jackson/datatype/jackson-datatype-jsr310/2.12.1/jackson-datatype-jsr310-2.12.1.pom'. Received status code 401 from server: Unauthorized

jzheaux (approved):

jzheaux:
Thanks, @evgeniycheban! This is now merged into `master`

Issue #9285, opened 2020-12-16
Reconsider AntPathRequestMatcher matching logic


jzheaux:
See https://github.com/spring-projects/spring-security/pull/8996#discussion_r506652742 and https://github.com/spring-projects/spring-security/pull/8996#pullrequestreview-528354323

evgeniycheban:
@jzheaux I can take this.

PR #9301, opened 2020-12-21
Reconsider AntPathRequestMatcher matching logic

evgeniycheban:
Closes gh-9285

### DIFF
--- a/web/src/main/java/org/springframework/security/web/util/matcher/AntPathRequestMatcher.java
+++ b/web/src/main/java/org/springframework/security/web/util/matcher/AntPathRequestMatcher.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -48,6 +48,7 @@ import org.springframework.web.util.UrlPathHelper;
  * @author Luke Taylor
  * @author Rob Winch
  * @author Eddú Meléndez
+ * @author Evgeniy Cheban
  * @since 3.1
  * @see org.springframework.util.AntPathMatcher
  */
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/springframework/security/web/util/matcher/AntPathRequestMatcher.java
```diff
@@ -159,9 +160,12 @@ public boolean matches(HttpServletRequest request) {
 
 	@Override
 	public MatchResult matcher(HttpServletRequest request) {
-		if (this.matcher == null || !matches(request)) {
+		if (!matches(request)) {
```
jzheaux, line 163: It seems like `this.matcher == null` should be the first check in order to avoid potential NPEs in `matches(HttpServletRequest)`.

web/src/test/java/org/springframework/security/web/util/matcher/AntPathRequestMatcherTests.java
```diff
@@ -196,6 +197,13 @@ public void matchesWithInvalidMethod() {
 		assertThat(matcher.matches(request)).isFalse();
 	}
 
+	@Test
+	public void matcherWhenMatchAllPatternThenMatchResult() {
```
jzheaux, line 201: Since this is a bug, will you please add `// gh-9285` above the test?

jzheaux (changes requested):

web/src/main/java/org/springframework/security/web/util/matcher/AntPathRequestMatcher.java
evgeniycheban: The `matcher` will only be null if the `pattern` is `MATCH_ALL`, and we need to check the `HttpMethod` first.

web/src/test/java/org/springframework/security/web/util/matcher/AntPathRequestMatcherTests.java
evgeniycheban: Done.

web/src/main/java/org/springframework/security/web/util/matcher/AntPathRequestMatcher.java
jzheaux: Good point. Is the null check necessary, then? It seems like it could be removed.

evgeniycheban: The null check is necessary since `matcher` is used to extract variables, for the `MATCH_ALL` pattern there will be an empty `Map` of variables.

jzheaux (approved):

PR #9303, opened 2020-12-22
Fixed NullPointerException with WWW-Authenticate

tristanessquare:
Fixed possible NullPointerException that happens when the OAuth2 ResourceServer does not return a valid WWW-Authenticate header format.

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/http/OAuth2ErrorResponseErrorHandler.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/http/OAuth2ErrorResponseErrorHandler.java
@@ -70,6 +70,9 @@ public class OAuth2ErrorResponseErrorHandler implements ResponseErrorHandler {
 			return null;
 		}
 		BearerTokenError bearerTokenError = getBearerToken(wwwAuthenticateHeader);
+		if (bearerTokenError == null) {
+			return new OAuth2Error(OAuth2ErrorCodes.SERVER_ERROR, null, null);
+		}
 		String errorCode = (bearerTokenError.getCode() != null) ? bearerTokenError.getCode()
 				: OAuth2ErrorCodes.SERVER_ERROR;
 		String errorDescription = bearerTokenError.getDescription();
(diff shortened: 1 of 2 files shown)


Tusenka (approved):

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/http/OAuth2ErrorResponseErrorHandler.java
```diff
@@ -70,7 +70,7 @@ private OAuth2Error readErrorFromWwwAuthenticate(HttpHeaders headers) {
 			return null;
 		}
 		BearerTokenError bearerTokenError = getBearerToken(wwwAuthenticateHeader);
-		String errorCode = (bearerTokenError.getCode() != null) ? bearerTokenError.getCode()
+		String errorCode = (bearerTokenError != null && bearerTokenError.getCode() != null) ? bearerTokenError.getCode()
```
jzheaux, line 73: This doesn't appear to address the issue completely since `bearerTokenError` is referred to later on in the method, potentially causing more NPEs.

If wonder if it would be better to do:

```java
if (bearerTokenError == null) {
    return new OAuth2Error(OAuth2ErrorCodes.SERVER_ERROR, null, null);
}
String errorCode = ...
```

jzheaux (changes requested):
Thanks for the PR, @tristanessquare! I've left some feedback inline.

jzheaux:
Hi, @tristanessquare! Are you able to make the requested changes?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/http/OAuth2ErrorResponseErrorHandler.java
tristanessquare: That's a good point. I will do that.

jzheaux:
Thanks, @tristanessquare! I think the changes make sense.

Since this is addressing a bug, would you please also add a test to `OAuth2ErrorResponseErrorHandlerTests` that confirms that the bug is fixed? Something that would have failed before the change, but now passes would be ideal.

jzheaux (approved):

jzheaux:
Thanks for the PR, @tristanessquare! This is now merged into `master` via 56db058fd098cdb268702b817500da9418f770e3

PR #9336, opened 2021-01-12
Make user info response status check error only

BenjaminFaal:
This would otherwise fail on anything in the 2xx range thats not 200 OK. For example when i use the user-info-uri: https://vl.api.np.km.playstation.net/vl/api/v1/mobile/users/me/info it returns 201 Created (dont ask me why Sony chose that). Anyway i think this status check will still cover most cases by checking if its a 2xx status.

Also this just works in the non reactive version: org.springframework.security.oauth2.client.userinfo.DefaultOAuth2UserService because it try catches the request.

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultReactiveOAuth2UserService.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultReactiveOAuth2UserService.java
@@ -108,7 +108,7 @@ public class DefaultReactiveOAuth2UserService implements ReactiveOAuth2UserServi
 					authenticationMethod);
 			// @formatter:off
 			Mono<Map<String, Object>> userAttributes = requestHeadersSpec.retrieve()
-					.onStatus((s) -> s != HttpStatus.OK, (response) ->
+					.onStatus((s) -> !s.is2xxSuccessful(), (response) ->
 						parse(response)
 							.map((userInfoErrorResponse) -> {
 								String description = userInfoErrorResponse.getErrorObject().getDescription();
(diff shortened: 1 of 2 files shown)


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultReactiveOAuth2UserService.java
```diff
@@ -108,7 +108,7 @@
 					authenticationMethod);
 			// @formatter:off
 			Mono<Map<String, Object>> userAttributes = requestHeadersSpec.retrieve()
-					.onStatus((s) -> s != HttpStatus.OK, (response) ->
+					.onStatus((s) -> !s.is2xxSuccessful(), (response) ->
```
jgrandja, line 111: It doesn't make sense to accept all 2xx status code. For example, `203 Non-Authoritative Information` or `205 Reset Content`.

jgrandja (commented):
@BenjaminFaal I don't feel we can accept this PR based on a non-compliant provider. Have you reported this to Sony?

BenjaminFaal:
Hmm thats a valid point i always thought 2xx was considered success. You are right that Sony should fix this but the nonreactive version: org.springframework.security.oauth2.client.userinfo.DefaultOAuth2UserService just works because it only try-catches the RestClientException and UnknownContentTypeException exceptions which a 201 does not cause.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultReactiveOAuth2UserService.java
jgrandja: @BenjaminFaal I reviewed `DefaultOAuth2UserService`. The associated `RestTemplate` uses `DefaultResponseErrorHandler` (indirectly), which checks for errors using `HttpStatus.isError()`.

Therefore, we can align `DefaultReactiveOAuth2UserService` with `DefaultOAuth2UserService` by applying the change `.onStatus(HttpStatus::isError, (response) -> ...`

Can you please also add a test for your use case, where the status returned is 201.

BenjaminFaal: @jgrandja done, see c4f6f3f3d8b6437f9de331c07b70761a08db4196

BenjaminFaal:
@jgrandja done, see c4f6f3f3d8b6437f9de331c07b70761a08db4196

jgrandja:
Thanks for the PR @BenjaminFaal ! This is now in master and I'll have this backported as well.

Issue #9366, opened 2021-01-22
Remove notEmpty check for authorities in DefaultOAuth2User


mayur9991:
**Describe the bug**
Currently, DefaultOAuth2User constructor has a check to make sure that authorities parameter is not empty.

```java
public DefaultOAuth2User(Collection<? extends GrantedAuthority> authorities, Map<String, Object> attributes, String nameAttributeKey) {
		Assert.notEmpty(authorities, "authorities cannot be empty");
		Assert.notEmpty(attributes, "attributes cannot be empty");
		Assert.hasText(nameAttributeKey, "nameAttributeKey cannot be empty");
		if (!attributes.containsKey(nameAttributeKey)) {
			throw new IllegalArgumentException("Missing attribute '" + nameAttributeKey + "' in attributes");
		}
		this.authorities = Collections.unmodifiableSet(new LinkedHashSet<>(this.sortAuthorities(authorities)));
		this.attributes = Collections.unmodifiableMap(new LinkedHashMap<>(attributes));
		this.nameAttributeKey = nameAttributeKey;
	}
```

This causes a problem when you have a custom authorities extractor and authorities list is empty for particular user.

```
java.lang.IllegalArgumentException: authorities cannot be empty
	at org.springframework.util.Assert.notEmpty(Assert.java:467)
	at org.springframework.security.oauth2.core.user.DefaultOAuth2User.<init>(DefaultOAuth2User.java:63)
	at org.springframework.security.oauth2.core.oidc.user.DefaultOidcUser.<init>(DefaultOidcUser.java:89)
	at com.xebialabs.platform.sso.oidc.service.XLOidcUserService.loadUser(XLOidcUserService.java:33)
	at com.xebialabs.xlrelease.auth.oidc.service.XlrOidcUserService.loadUser(XlrOidcUserService.java:28)
	at com.xebialabs.xlrelease.auth.oidc.service.XlrOidcUserService.loadUser(XlrOidcUserService.java:16)
	at org.springframework.security.oauth2.client.oidc.authentication.OidcAuthorizationCodeAuthenticationProvider.authenticate(OidcAuthorizationCodeAuthenticationProvider.java:174)
	at org.springframework.security.authentication.ProviderManager.authenticate(ProviderManager.java:199)
```

**Expected behavior**
The implementation should be more inline with other part of spring-security. For example `DefaultOAuth2AuthenticatedPrincipal` or `AbstractAuthenticationToken` which allows to have empty list.

[...]

PR #9380, opened 2021-01-27
Allow null or empty authorities for DefaultOAuth2User

mayur9991:
Make DefaultOAuth2User more inline with other part of
spring-security.
For example,
- DefaultOAuth2AuthenticatedPrincipal
- AbstractAuthenticationToken

Closes gh-9366


jgrandja:
Thank you for the PR @mayur9991. This is now in master as well as backported.

PR #8591, opened 2020-05-25
Using SHA-256 to calculate the digital signature to be put in the cookie

aswinmahendiran:
### Fix for the issue gh-8549 (Replacing `MD5` hashing for remember me token)

- Using **`SHA-256`** instead of **`MD5`** algorithm to calculate the digital signature to be put in the cookie.

- Updated the tests to use **`sha256Hex`** instead of **`md5Hex`**

- Updated the documentation to use **`sha256Hex`** instead of **`md5Hex`**

Issue #8549, opened 2020-05-18
Replacing MD5 hashing for remember me token


islamazhar:
**Expected Behavior**

To hash passwords and secret keys a secure hashing algorithm (e.g., SHA256) should be used.

**Current Behavior**

The current `Simple Hash-Based Token Approach` uses **MD5 hash digest** for hashing `password` and `secret-key` in `Remember-Me Authentication`. 

Corresponding [Javadoc link](https://docs.spring.io/spring-security/site/docs/current/reference/html5/#remember-me-hash-token) and [source code link](https://github.com/spring-projects/spring-security/blob/bfb401eeed57f63f2504d04cb5fdf93e183206e8/web/src/main/java/org/springframework/security/web/authentication/rememberme/TokenBasedRememberMeServices.java#L158)  

**Context**
MD5 is already proven to be a weak hashing algorithm and vulnerable against collision attacks [1] and modular differential attacks [2].  Hence I suggest using a secure hashing algorithm such as  SHA-256 instead of already broken MD5 for `remember me token`. 

**References**
[1] [Den Boer and A. Bosselaers, “Collisions for the compression function of MD5,” in Workshop on the Theory and Application of CryptographicTechniques, pp. 293–304, Springer, 1993](https://link.springer.com/chapter/10.1007/3-540-48285-7_26)

[2] [Wang, Xiaoyun, and Hongbo Yu. "How to break MD5 and other hash functions." Annual international conference on the theory and applications of cryptographic techniques. Springer, Berlin, Heidelberg, 2005.](http://merlot.usc.edu/csac-f06/papers/Wang05a.pdf)

rwinch:
Thanks for creating this issue. 

In modern applications, I'd recommend using [Spring Session's Remember Me](https://docs.spring.io/spring-session/docs/current/reference/html5/#spring-security-rememberme) support which allows you to easily offload the session into a data store without the need for cryptography.

I do agree that MD5 is not a good choice. Would you like to provide a pull request that provides the option to use SHA-256?

aswinmahendiran:
Hello @islamazhar thanks for raising this issue. 

Hello @rwinch I have raised the PR for the fix. I have updated the tests as well. Please check the PR (#8591).

This is my first PR in any Open Source project. Happy to start my journey in the security field I love the most!!

aswinmahendiran:
Hi @rwinch it would be great if you can a look at the PR. I am excited to add my first contribution

rwinch:
[...]

PR #9392, opened 2021-02-01
Implement SHA256 algorithm in Remember Me tokens

sargas:
This pull requests addresses #8549 by allowing for the use of SHA-256 when generating the cookies used in Remember Me cookies, as well as enabling `TokenBasedRememberMeServices` to accept such cookies. This allows for the use of a cryptographically secure algorithm for encoding the password and key, making it harder for an attacker to potentially obtain these even from an older cookie.

There are a few existing alternatives that provide a more secure implementation of the Remember Me functionality: `PersistentTokenBasedRememberMeServices` (which uses random strings), or Spring Session's `SpringSessionRememberMeServices`. This PR helps secure the "simple" implementation that is suggested first in the documentation, so it likely appears as more approach to users and does not require a database. Alternatively, users can subclass `TokenBasedRememberMeServices` to provide a more secure hashing algorithm, and this PR seeks not to break any code that is doing so.

This PR extends the cookie format used by `TokenBasedRememberMeServices` to have an optional component before the hashed-string containing the algorithm used. Existing cookies are still accepted, and the existing (and overridable) `makeTokenSignature(long,String,password)` is used if no algorithm is specified in the cookie, or when generating cookies without any hashing algorithm specified. The breaking change to remove support for cookies that do not have a hashing algorithm specified, potentially in Spring Security 6, would allow for the default hashing algorithm to be changed in the future without breaking existing Remember Me cookies.

Unlike previous pull requests from last May (PR #8580 and #8591), this is designed to preserve existing functionality unless the user explicitly chooses an algorithm (currently MD5 and SHA256).

This has one test failure I know of - `XsdDocumentedTests.countWhenReviewingDocumentationThenAllElementsDocumented`. I added documentation to `namespace.adoc` and updated the schema in `spring-security-5.5.rnc`, but the test class is still referring to `spring-security-5.4.rnc`. If I switch the test class to use `spring-security-5.5.rnc`, then there are two test failures that look related to the lack of documentation in `namespace.adoc` for https://github.com/spring-projects/spring-security/commit/54d3839f63a41f81aff43e88f1acd57e5a138838 .

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/rememberme/TokenBasedRememberMeServices.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/rememberme/TokenBasedRememberMeServices.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2004, 2005, 2006 Acegi Technology Pty Limited
+ * Copyright 2004-2021 Acegi Technology Pty Limited
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,6 +20,7 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.Arrays;
 import java.util.Date;
+import java.util.Objects;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -51,11 +52,18 @@ import org.springframework.util.StringUtils;
  * This is also necessary so that the user's password is available and can be checked as
  * part of the encoded cookie.
  * <p>
- * The cookie encoded by this implementation adopts the following form:
+ * The cookie encoded by this implementation adopts either the following form:
  *
  * <pre>
  * username + &quot;:&quot; + expiryTime + &quot;:&quot;
  * 		+ Md5Hex(username + &quot;:&quot; + expiryTime + &quot;:&quot; + password + &quot;:&quot; + key)
+ * </pre>
+ *
+ * or, if a hashing algorithm is configured, the form:
+ *
+ * <pre>
+ * username + &quot;:&quot; + expiryTime + &quot;:&quot; + algorithmName + &quot;:&quot;
+ * 		+ hashingAlgorithm(username + &quot;:&quot; + expiryTime + &quot;:&quot; + password + &quot;:&quot; + key)
  * </pre>
  *
  * <p>
@@ -68,6 +76,11 @@ import org.springframework.util.StringUtils;
  * this implementation (as we do not want to rely on a database for remember-me services).
  * High security applications should be aware of this occasionally undesired disclosure of
  * a valid username.
+ * <p>
+ * The hashing algorithm used, if unspecified, is the MD5 message-digest algorithm in this
+ * implementation. If the configured hashing algorithm is changed, existing cookies will
+ * still be honored if the key or password are unchanged, but the users' cookies will use
+ * the new hashing algorithm the next time they interactively login.
  * <p>
  * This is a basic remember-me implementation which is suitable for many applications.
  * However, we recommend a database-based implementation if you require a more secure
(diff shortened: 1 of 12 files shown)


web/src/main/java/org/springframework/security/web/authentication/rememberme/TokenBasedRememberMeServices.java
```diff
@@ -83,22 +96,34 @@
...
+	}
+
+	/**
+	 * @since 5.5
+	 */
+	public TokenBasedRememberMeServices(String key, UserDetailsService userDetailsService,
+			RememberMeHashingAlgorithm hashingAlgorithm) {
 		super(key, userDetailsService);
+		this.hashingAlgorithm = hashingAlgorithm;
 	}
```
marcusdacoregio, line 112: We should prevent the algorithm from being null in the constructor

```suggestion
	public TokenBasedRememberMeServices(String key, UserDetailsService userDetailsService) {
		super(key, userDetailsService);
	}

	public TokenBasedRememberMeServices(String key, UserDetailsService userDetailsService, RememberMeHashingAlgorithm hashingAlgorithm) {
		this(key, userDetailsService);
		Assert.notNull(hashingAlgorithm, "hashingAlgorithm cannot be null");
		this.hashingAlgorithm = hashingAlgorithm;
	}
```

web/src/main/java/org/springframework/security/web/authentication/rememberme/TokenBasedRememberMeServices.java
```diff
@@ -83,22 +96,34 @@
...
 
 	@Override
 	protected UserDetails processAutoLoginCookie(String[] cookieTokens, HttpServletRequest request,
 			HttpServletResponse response) {
-		if (cookieTokens.length != 3) {
+		if (!(cookieTokens.length == 3 || cookieTokens.length == 4)) {
```
marcusdacoregio, line 117: I think we can simplify the checks here and create an util class:

```java
private static class CookieTokens {

		private final String username;

		private final long expirationTime;

		private final RememberMeHashingAlgorithm algorithm;

		private final String signature;

		private CookieTokens(String[] cookieTokens) {
			if (!(cookieTokens.length == 3 || cookieTokens.length == 4)) {
				throw new InvalidCookieException(
						"Cookie token did not contain 3 or 4 tokens, but contained '" + Arrays.asList(cookieTokens) + "'");
			}
			this.username = getUsername(cookieTokens);
			this.expirationTime = getTokenExpiryTime(cookieTokens);
			this.algorithm = getTokenHashingAlgorithm(cookieTokens);
			this.signature = getSignature(cookieTokens);
		}

		private String getUsername(String[] cookieTokens) {
			return cookieTokens[0];
		}

		private long getTokenExpiryTime(String[] cookieTokens) {
			try {
				return Long.parseLong(cookieTokens[1]);
			}
			catch (NumberFormatException nfe) {
				throw new InvalidCookieException(
						"Cookie token[1] did not contain a valid number (contained '" + cookieTokens[1] + "')");
			}
		}

		private RememberMeHashingAlgorithm getTokenHashingAlgorithm(String[] cookieTokens) {
			if (cookieTokens.length == 3) {
				return null;
			}
			return RememberMeHashingAlgorithm.findByIdentifier(cookieTokens[2]);
		}

		private String getSignature(String[] cookieTokens) {
			return (cookieTokens.length == 4) ? cookieTokens[3] : cookieTokens[2];
		}

	}
```

And this method should use the properties from this class:

```java
CookieTokens properties = new CookieTokens(cookieTokens);
if (isTokenExpired(properties.expirationTime)) {
	throw new InvalidCookieException("Cookie token[1] has expired (expired on '" + new Date(properties.expirationTime)
			+ "'; current time is '" + new Date() + "')");
}
...
```

web/src/main/java/org/springframework/security/web/authentication/rememberme/TokenBasedRememberMeServices.java
```diff
@@ -129,14 +155,39 @@ private long getTokenExpiryTime(String[] cookieTokens) {
...
+	 *
+	 * Default value is the hashing algorithm applied to
+	 * "username:tokenExpiryTime:password:key"
+	 * @since 5.5
+	 */
+	protected String makeTokenSignature(long tokenExpiryTime, String username, String password,
+			RememberMeHashingAlgorithm algorithm) {
+		if (algorithm == RememberMeHashingAlgorithm.UNSET) {
+			return makeTokenSignature(tokenExpiryTime, username, password);
+		}
```
marcusdacoregio, line 187:
```suggestion
		if (algorithm == null) {
			algorithm = DEFAULT_HASHING_ALGORITHM;
		}
```

By doing this you can get rid of the `protected String makeTokenSignature(long tokenExpiryTime, String username, String password)` method.

web/src/main/java/org/springframework/security/web/authentication/rememberme/TokenBasedRememberMeServices.java
```diff
@@ -234,4 +287,20 @@ private static boolean equals(String expected, String actual) {
 		return (s != null) ? Utf8.encode(s) : null;
 	}
 
+	/**
+	 * @since 5.5
+	 */
+	public RememberMeHashingAlgorithm getHashAlgorithm() {
```
marcusdacoregio, line 293: Is this method really needed?

web/src/main/java/org/springframework/security/web/authentication/rememberme/TokenBasedRememberMeServices.java
```diff
@@ -234,4 +287,20 @@ private static boolean equals(String expected, String actual) {
...
+		return this.hashingAlgorithm;
+	}
+
+	/**
+	 * Sets the hashing algorithm used in new cookies
+	 * @throws NullPointerException when the algorithm is null
+	 * @since 5.5
+	 */
+	public void setHashAlgorithm(RememberMeHashingAlgorithm hashingAlgorithm) {
```
marcusdacoregio, line 302: Please rename to `hashingAlgorithm`

marcusdacoregio (changes requested):
Hi @sargas, nice work. I've left my feedback inline.

The target release for this feature is now `5.7`, can you change the `@since` tags to this version?
Also, please rebase your branch with `main`.

About the tests, once we have the implementation polished, I can help you figure it out.

marcusdacoregio:
Due to lack of feedback on this PR, I'm closing this in favor of https://github.com/spring-projects/spring-security/pull/10675.

I took the commit from this PR and applied the polish on top of it.

Issue #9399, opened 2021-02-02
Consider downgrading to Nimbus 8


jzheaux:
Due to a version incompatibility between `oauth2-oidc-sdk:8.x` and `nimbus-jose-jwt:9.x`, various users have reported friction with using Spring Security 5.4.x.

Since Spring Security doesn't rely on any `nimbus-jose-jwt:9.x` features, and since Nimbus will be supporting `nimbus-jose-jwt:8.20.x` for a longer time than normal, it seems like a win-win to change Spring Security 5.4.x to use `nimbus-jose-jwt:8.20.x` instead.

Anyone who needs 9.x features from `nimbus-jose-jwt` could declare their own dependency version in their applications.

spring-projects-issues:
Fixed via e79141a1888da237f55d53237b8294ea3397c76b

ThanksForAllTheFish:
Unfortunately this broke `nimbus-jose-jwt` shaded `json-smart`, since that is available only in 9+. A new vulnerability was recently found in `json-smart`, https://nvd.nist.gov/vuln/detail/CVE-2021-27568. or at least, in my current setup (with Spring Boot), simply setting `nimbus-jose-jwt.version=9.1.3` does not work, complaining at runtime of missing `json-smart` when using `spring-security` 5.4.5, but seems good with `spring-security` 5.4.2. Relevant gradle properties to have a seemingly working setup without `json-smart` with Spring Boot 2.4.3:

- in build.gradle
```
configurations.all {
  exclude module: "json-smart"
}
dependencies {
  implementation 'org.springframework.boot:spring-boot-starter-security'
  implementation 'org.springframework.boot:spring-boot-starter-oauth2-client'
  implementation 'org.springframework.boot:spring-boot-starter-oauth2-resource-server'
}
```
- in `gradle.properties`
```
spring-security.version=5.4.2
nimbus-jose-jwt.version=9.1.3
oauth2-oidc-sdk.version=9.2.2 //not sure this is needed
```
I will come up with a full issue, just don't have time right now to create a new project and tinker with its setup.

erlioniel:
@ThanksForAllTheFish that's absolutely correct. We are struggling with the same CVE right now. However I think upgrade to nimbus 9.x won't solve the issue, but will hide it. The fix for json-smart library [doesn't exist right now](https://github.com/netplex/json-smart-v2/issues/62) and doesn't look like it will be in the nearest future.

[...]

PR #9453, opened 2021-02-12
Downgrade to Nimbus JOSE JWT 8.+

wilkinsona:
This pull request corrects the fix for https://github.com/spring-projects/spring-security/issues/9399. The previous attempt accidentally left Nimbus JOSE JWT constrained at 9.+ in the main build. The Boot-based samples didn't catch the problem as the constraints that were added were ineffective as Boot's dependency management was taking precedence.

https://ge.spring.io/s/gsspmqm777k4w is a build scan with these changes, showing [many 8.x dependencies](https://ge.spring.io/s/gsspmqm777k4w/dependencies?dependencies=nimbus-jose-jwt:8&expandAll) and [no 9.x dependencies](https://ge.spring.io/s/gsspmqm777k4w/dependencies?dependencies=nimbus-jose-jwt:9&expandAll).


wilkinsona:
I'm not sure what we should do for next week's Boot 2.4 release but we have a few options. In my order of preference, they are:

1. Upgrade to a new Security 5.4.5 release and downgrade to JOSE JWT 8.x
2. Stay on Security 5.4.2 and JOSE JWT 9.1.3
3. Update to Security 5.4.4 but stay on JOSE JWT 9.1.3
4. Upgrade to Security 5.4.4 and JOSE JWT 9.5

1 is my favourite as it gets the fix into the hands of Boot's users with as little churn for them as possible. 4 is my least favourite as we would be inflicting a minor dependency upgrade in one Boot maintenance release knowing that we're then going to inflict a major dependency downgrade in the next maintenance release. I think this will damage users' confidence in Boot's dependency management. 3 is quite risky as I don't know if Security is binary compatible with 9.1.3 having been compiled against 9.5.

wilkinsona:
The build failure appears to be unrelated. It was caused by a bad connection to PCF one:

```
Caused by: org.springframework.web.client.ResourceAccessException: I/O error on GET request for "https://simplesaml-for-spring-saml.apps.pcfone.io/saml2/idp/metadata.php": Connection timed out (Connection timed out); nested exception is java.net.ConnectException: Connection timed out (Connection timed out)
```

Issue #9425, opened 2021-02-10
ACL can't be owned by a GrantedAuthoritySid


bberto:
**Describe the bug**
ACL can't be owned by a GrantedAuthoritySid

**To Reproduce**
As user with a role named TEST, after successfully changing an ACL ownership in this way:
```
acl.setOwner(new GrantedAuthoritySid("ROLE_TEST"));
aclService.updateAcl(acl);
```
I can't perform further modifications to the ACL (eg. changing back the ownership to me).

see:
https://github.com/spring-projects/spring-security/blob/master/acl/src/main/java/org/springframework/security/acls/domain/AclAuthorizationStrategyImpl.java#L92

**Expected behavior**
Any user with ROLE_TEST granted authority should be able to change the ACL.

jgrandja:
@bberto I'm trying to understand the issue but it's not clear at the moment. Can you put together a test that demonstrates the issue and another test that you would expect to pass.

bberto:
@jgrandja I created a pull request including the test and a proposed fix

jgrandja:
@bberto Thank you for providing the PR to demonstrate your use case.

The sample code you provided:

`acl.setOwner(new GrantedAuthoritySid("ROLE_TEST"))`

is not correct from a logical entity relationship viewpoint.

An `Acl` is an entity and so is the `Acl.getOwner()`. A `Principal` is an entity, which owns an `Acl`.
However, a `GrantedAuthority` **is an attribute** of a `Principal` and NOT an entity representation.
Therefore, an `Acl` cannot be owned by a `GrantedAuthoritySid` since it's an attribute representation and NOT an entity - only an entity can own an `Acl`.

Hope this makes sense?

Based on this, I'm going to close this issue and associated PR.

bberto:
Your explanation is reasonable. But, reading [documentation](https://docs.spring.io/spring-security/site/docs/current/reference/html5/#domain-acls), seems that Acl is not owned by a `Principal` but by a `Sid`:

> Sid: The ACL module needs to refer to principals and GrantedAuthority[] s. **A level of indirection is provided by the Sid interface**, which is an abbreviation of "security identity". Common classes include PrincipalSid (to represent the principal inside an Authentication object) and GrantedAuthoritySid


> ACL_SID allows us to uniquely identify any principal **or authority** in the system ("SID" stands for "security identity")
> [...]
[...]

PR #9454, opened 2021-02-12
Allow ACL to be owned by GrantedAuthoritySid

bberto:
Allow ACL to be owned by GrantedAuthoritySid

Closes gh-9425

### DIFF
--- a/acl/src/test/java/org/springframework/security/acls/domain/AclAuthorizationStrategyImplTests.java
+++ b/acl/src/test/java/org/springframework/security/acls/domain/AclAuthorizationStrategyImplTests.java
@@ -16,13 +16,13 @@
 
 package org.springframework.security.acls.domain;
 
+
 import java.util.Arrays;
 import java.util.List;
 
 import org.junit.After;
 import org.junit.Test;
 import org.junit.runner.RunWith;
-import static org.mockito.Mockito.when;
 import org.mockito.Mock;
 import org.mockito.junit.MockitoJUnitRunner;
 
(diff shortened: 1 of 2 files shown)


jgrandja:
Closing as per [comment](https://github.com/spring-projects/spring-security/issues/9425#issuecomment-784264364)

acl/src/test/java/org/springframework/security/acls/domain/AclAuthorizationStrategyImplTests.java
```diff
@@ -41,15 +43,12 @@
 	@Mock
 	Acl acl;
 
-	GrantedAuthority authority;
+	GrantedAuthority authority = new SimpleGrantedAuthority("ROLE_SYSTEM_ADMIN");
 
 	AclAuthorizationStrategyImpl strategy;
 
-	@Before
-	public void setup() {
```
jgrandja, line 49: Please revert this change and preserve the `@Before` method and create a `withUserRoles()` static helper method  instead.

jgrandja (changes requested):
Thanks for the PR @bberto. Please see review comment.

acl/src/test/java/org/springframework/security/acls/domain/AclAuthorizationStrategyImplTests.java
bberto: switching back to `@Before` method I realized that helper method is not needed

jgrandja:
Thanks for the PR @bberto. This is now in master.

PR #9472, opened 2021-02-23
Limit OAuth2 login links to redirect-based flows

### LINKED ISSUE #9457: oauth2Login() generates authorization links for "client_credentials" grant type
**Describe the bug**

In Spring Security WebFlux, `OAuth2LoginSpec#getLinks` populates the returned map with authorization links for _all_ registered client registrations, even for those using the `"client_credentials"` grant type meant for server-to-server authentication. This has the following consequences:

This means that if you have two client registration A with grant type `authorization_code` and another registration B with `client_credentials`, the authorization entry point created by Spring Security WebFlux redirects to the default login page (`/login`) rather than directly to the authorization flow of A (`/oauth2/authorization/A`). Even worse, that login page will show an authorization link for registration B even though it is not possible for a user to log in through the `client_credentials` OAuth2 flow. In fact, following this link yields a 500 error as `DefaultServerOAuth2AuthorizationRequestResolver` (sensibly) does not support `client_credentials`.

**To Reproduce**

0. Create a Spring Boot application with WebFlux and Spring Security
0. Define two client registrations within `spring.security.oauth2.client.registration` in `application.yml`:
    * `a` with `authorization-grant-type: authorization_code`
    * `b` with `authorization-grant-type: client_credentials`
0. Configure a `ServerHttpSecurity` bean protecting some path with `oauth2Login()`
0. Start the application
0. Navigate to the protected path

**Expected behavior**

I get redirected to `/oauth2/authorization/a` and then to the OAuth2 provider's login page.

denisw:
This prevents the generated login page from showing links for
authorization grant types like "client_credentials" which are
not redirect-based, and thus not meant for interactive use in
the browser.

Closes #9457.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
@@ -80,6 +80,7 @@ import org.springframework.security.oauth2.client.web.server.ServerOAuth2Authori
 import org.springframework.security.oauth2.client.web.server.ServerOAuth2AuthorizedClientRepository;
 import org.springframework.security.oauth2.client.web.server.WebSessionOAuth2ServerAuthorizationRequestRepository;
 import org.springframework.security.oauth2.client.web.server.authentication.OAuth2LoginAuthenticationWebFilter;
+import org.springframework.security.oauth2.core.AuthorizationGrantType;
 import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
 import org.springframework.security.oauth2.core.OAuth2AuthorizationException;
 import org.springframework.security.oauth2.core.endpoint.OAuth2AuthorizationRequest;
@@ -3356,8 +3357,11 @@ public class ServerHttpSecurity {
 				return Collections.emptyMap();
 			}
 			Map<String, String> result = new HashMap<>();
-			registrations.iterator().forEachRemaining(
-					(r) -> result.put("/oauth2/authorization/" + r.getRegistrationId(), r.getClientName()));
+			registrations.iterator().forEachRemaining((r) -> {
+				if (AuthorizationGrantType.AUTHORIZATION_CODE.equals(r.getAuthorizationGrantType())) {
+					result.put("/oauth2/authorization/" + r.getRegistrationId(), r.getClientName());
+				}
+			});
 			return result;
 		}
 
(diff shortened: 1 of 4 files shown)


config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -3356,8 +3357,12 @@ private ServerWebExchangeMatcher createAttemptAuthenticationRequestMatcher() {
 				return Collections.emptyMap();
 			}
 			Map<String, String> result = new HashMap<>();
-			registrations.iterator().forEachRemaining(
-					(r) -> result.put("/oauth2/authorization/" + r.getRegistrationId(), r.getClientName()));
+			registrations.iterator().forEachRemaining((r) -> {
+				if (AuthorizationGrantType.AUTHORIZATION_CODE.equals(r.getAuthorizationGrantType())
+						|| AuthorizationGrantType.IMPLICIT.equals(r.getAuthorizationGrantType())) {
```
jgrandja, line 3362: Let's remove `AuthorizationGrantType.IMPLICIT` since it's `@Deprecated`. Also, the only supported redirect-based flows is `AUTHORIZATION_CODE`.

jgrandja (changes requested):
Thanks for the PR @denisw. Please see review comment.

Also, could you apply the same change to `OAuth2LoginConfigurer` and add tests for both.
Thanks!

denisw:
@jgrandja Thanks for the feedback! Yes, I will apply the change to `OAuth2LoginConfigurer` as well. Regarding tests, this piece of the code doesn't seem to be covered very well at the moment, so I think I need to figure out how to test this properly. I'll try, though.

jgrandja:
@denisw 

> I need to figure out how to test this properly

For `OAuth2LoginConfigurer` test, see `OAuth2LoginApplicationTests` (oauth2login Boot sample)

and for `OAuth2LoginSpec`, see `org.springframework.security.config.web.server.OAuth2LoginTests`

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
denisw: I removed the `IMPLICIT` case.

denisw:
@jgrandja I added a test case for the WebFlux-side change. I'll continue with the `OAuth2LoginConfigurer` side now.

denisw:
@jgrandja I also updated `OAuth2LoginConfigurer` and added an appropriate test. The PR should now be complete and ready for re-review.

jgrandja:
Thanks for all the updates @denisw ! FYI, I applied some minor polish to the tests.

This is now in master and I'll have it backported as well.

Issue #9027, opened 2020-09-16
BasicAuthenticationFilter breaks SwitchUserFilter


smithtonson:
**Describe the bug**
When using the SwitchUserFilter with Basic auth, BasicAuthenticationFilter.authenticationIsRequired() fails when switched forcing a re-authentication.

**To Reproduce**
Enable both basic auth and the SwitchUserFilter, switch to a new user and then make another request.
BasicAuthenticationFilter.authenticationIsRequired() will return true since the current logged in user doesn't match the WWW-Authenticate header

**Expected behavior**
BasicAuthenticationFilter.authenticationIsRequired() should also check if existingAuth has a GrantedAuthority of ROLE_PREVIOUS_ADMINISTRATOR, and check that username against ```username```
```
		if (existingAuth instanceof UsernamePasswordAuthenticationToken
				&& !existingAuth.getName().equals(username)) {
			return true;
		}
```

smithtonson:
An alternative fix (what I've implemented for myself locally in a CustomSwitchUserFilter) would be to replace creating a UsernamePasswordAuthenticationToken in SwitchUserFilter and instead create a new authentication token like ```SwitchUserToken``` this would bypass the ```existingAuth instanceof UsernamePasswordAuthenticationToken``` check.

ianroberts:
I've offered a possible fix in PR #9480, pretty much as described under "expected behaviour" above.

eleftherias:
Thanks for reaching out @smithtonson.
I suspect this is an issue when using Basic auth and switching the user in the browser.

Typically with Basic auth, you would keep your application stateless and include the credentials in every request.
In that case, it doesn't make sense to use the `SwitchUserFilter`, since the application doesn't have a way to remember that the user was switched.

You also have the option to provide the username and password using Basic auth, receive a Session ID and use the Session ID in all subsequent requests.
In this case, it is sensible to use the `SwitchUserFilter`.
This is the sequence of calls that I used and it worked successfully.
```
# get Session ID using basic auth
curl -v -u admin:password  http://localhost:8080/

# use Session ID when switching users
curl -v -b "JSESSIONID=..." -d "username=user1" http://localhost:8080/login/impersonate

# use same Session ID to impersonate user1
curl -v -b "JSESSIONID=..." http://localhost:8080/info

# received info for user1
```

A problem occurs if you are in a browser, because it will also send the `Authorization: Basic ...` header in the final request.

[...]

PR #9480, opened 2021-03-01
Fix interaction between basic auth and "switch user"

ianroberts:
As per gh-9027, the "switch user" mechanism and the basic authentication filter do not currently work well together.  When impersonating another user, the basic authentication username will be that of the administrator (the one doing the impersonating) but the username cached in the `Authentication` in the security context will be the user being impersonated.  The basic auth filter checks the basic auth username against the cached username and forces a re-authentication if they differ, which in turn breaks the "switch user" function.

This PR aims to fix this issue by detecting when the cached authentication is a "switch user" case, and then checking the current basic auth username against the original username we switched _from_ (extracted from the `SwitchUserGrantedAuthority`) rather than against the impersonated username we have switched _to_.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilter.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilter.java
@@ -31,10 +31,12 @@ import org.springframework.security.authentication.AuthenticationManager;
 import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.context.SecurityContextHolder;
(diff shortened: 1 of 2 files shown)


eleftherias (commented):
Thanks for the PR @ianroberts.

However, this isn't the approach we want to take to handle this scenario.

Any logic regarding switching the user should remain in the `SwitchUserFilter` and the related classes.
The `BasicAuthenticationFilter` shouldn't have any knowledge of user switching. 

I also have some general concerns about whether this feature makes sense and what user switching means in the context of HTTP basic auth. I will continue that discussion on the original issue.
Please hold off on any changes until we have come to a decision on the original issue.

ianroberts:
Fair enough - my specific case is a bit weird (not real basic auth, but rather a special SSO reverse proxy in front of my app which fakes a basic auth header when calling the backend), I can see why you'd hesitate to use this approach in the general case.

eleftherias:
I'm going to close this based on the discussion on gh-9027.
The original issue will remain open.

PR #9488, opened 2021-03-05
Consistent username and password parameter handling

weixsun:
Consider consistent username and password parameter processing of trim in `UsernamePasswordAuthenticationFilter`.

The scenario is as follows:
An application registration logic considers that the username with spaces before and after it is legal, then there will be problems using Spring Security's UsernamePasswordAuthenticationFilter.

> The scene is legal but unreasonable


jzheaux:
Thanks for the suggestion, @weixsun, though I think this would break existing users of Spring Security, which don't allow whitespace at the beginning or end of a username. Since this is the way Spring Security has worked for quite some time, I hesitate to make this change.

In your case, it's quite straightforward to change this behavior, though, by either extending `UsernamePasswordAuthenticationFilter#attemptAuthentication` or by using `AuthenticationFilter` and a custom `AuthenticationConverter`.

PR #9513, opened 2021-03-23
Expire OAuth2AuthorizationRequest when saving to the session

candrews:
Closes gh-5145

This straightforward approach cleans up expired `OAuth2AuthorizationRequest`s as they're loaded (and subsequently saved) to the session. It's basically the same approach as https://github.com/spring-projects/spring-security/pull/7381 but storing creation instead of expiration and using a wrapper class instead of modifying `OAuth2AuthorizationRequest`.

This approach introduces no new dependencies or requirements (such as Spring Cache abstraction).

From https://github.com/spring-projects/spring-security/pull/7381#pullrequestreview-295160506

> The expiring capability is really a cross-cutting concern. Can you try the following solution?
> 
>     * Intercept calls to `loadAuthorizationRequest()` and store the information necessary
> 
>     * At a scheduled time (configurable with skew), the component will attempt to `loadAuthorizationRequest()` and if it returns than determine if time has expired and than `removeAuthorizationRequest`
> 
> 
> Please take a look at the [Cache abstraction](https://docs.spring.io/spring/docs/current/spring-framework-reference/integration.html#cache) to see if it would meet our needs.

For `HttpSessionOAuth2AuthorizationRequestRepository`, I don't believe that could work because the session can only be loaded and saved while the request for it is being processed. In other words, the task that runs at the scheduled itme would not be able to call  `loadAuthorizationRequest()` and get the data back, nor would it be able to call `removeAuthorizationRequest` to write it. Even if it could call those methods, they wouldn't work right; session persistence is handed by servlet filters (see Spring Session) and those filters wouldn't be intercepting session use to handle persistence outside of a request.

It make sense to create a new implementation of the `AuthorizationRequestRepository<OAuth2AuthorizationRequest>` interface at some point which uses a database or Spring Cache abstraction for persistence, as that would allow for expiration to run as a scheduled job. However, that's not of interest to me at the moment :)

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
@@ -69,8 +69,8 @@ public final class HttpSessionOAuth2AuthorizationRequestRepository
 			return null;
 		}
 		Map<String, OAuth2AuthorizationRequestReference> authorizationRequests = this.getAuthorizationRequests(request);
-		OAuth2AuthorizationRequestReference wrappedWithCreated = authorizationRequests.get(stateParameter);
-		return (wrappedWithCreated != null) ? wrappedWithCreated.wrapped : null;
+		OAuth2AuthorizationRequestReference authorizationRequestReference = authorizationRequests.get(stateParameter);
+		return (authorizationRequestReference != null) ? authorizationRequestReference.authorizationRequest : null;
 	}
 
 	@Override
@@ -103,7 +103,7 @@ public final class HttpSessionOAuth2AuthorizationRequestRepository
 			return null;
 		}
 		Map<String, OAuth2AuthorizationRequestReference> authorizationRequests = this.getAuthorizationRequests(request);
-		OAuth2AuthorizationRequestReference wrappedWithCreatedOriginalRequest = authorizationRequests
+		OAuth2AuthorizationRequestReference authorizationRequestReference = authorizationRequests
 				.remove(stateParameter);
 		if (!authorizationRequests.isEmpty()) {
 			request.getSession().setAttribute(this.sessionAttributeName, authorizationRequests);
@@ -111,7 +111,7 @@ public final class HttpSessionOAuth2AuthorizationRequestRepository
 		else {
 			request.getSession().removeAttribute(this.sessionAttributeName);
 		}
-		return (wrappedWithCreatedOriginalRequest != null) ? wrappedWithCreatedOriginalRequest.wrapped : null;
+		return (authorizationRequestReference != null) ? authorizationRequestReference.authorizationRequest : null;
 	}
 
 	@Override
@@ -193,12 +193,13 @@ public final class HttpSessionOAuth2AuthorizationRequestRepository
 
 		private final Instant expiresAt;
 
-		private final OAuth2AuthorizationRequest wrapped;
+		private final OAuth2AuthorizationRequest authorizationRequest;
 
-		private OAuth2AuthorizationRequestReference(OAuth2AuthorizationRequest wrapped, Instant created) {
-			Assert.notNull(wrapped, "wrapped cannot be null");
-			this.expiresAt = created;
-			this.wrapped = wrapped;
+		private OAuth2AuthorizationRequestReference(OAuth2AuthorizationRequest authorizationRequest,
+				Instant expiresAt) {
+			Assert.notNull(authorizationRequest, "authorizationRequest cannot be null");
+			this.expiresAt = expiresAt;
+			this.authorizationRequest = authorizationRequest;
 		}
 
 	}
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
@@ -16,6 +16,10 @@
 
 package org.springframework.security.oauth2.client.web;
 
+import java.io.Serializable;
+import java.time.Clock;
+import java.time.Duration;
+import java.time.Instant;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -23,6 +27,7 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.servlet.http.HttpSession;
 
+import org.springframework.security.core.SpringSecurityCoreVersion;
 import org.springframework.security.oauth2.core.endpoint.OAuth2AuthorizationRequest;
 import org.springframework.security.oauth2.core.endpoint.OAuth2ParameterNames;
 import org.springframework.util.Assert;
@@ -30,9 +35,14 @@ import org.springframework.util.Assert;
 /**
  * An implementation of an {@link AuthorizationRequestRepository} that stores
  * {@link OAuth2AuthorizationRequest} in the {@code HttpSession}.
+ * <p>
+ * <b>NOTE:</b> {@link OAuth2AuthorizationRequest}s expire after two minutes, the default
+ * duration can be configured via
+ * {@link #setOAuth2AuthorizationRequestExpiresIn(Duration)}.
  *
  * @author Joe Grandja
  * @author Rob Winch
+ * @author Craig Andrews
  * @since 5.0
  * @see AuthorizationRequestRepository
  * @see OAuth2AuthorizationRequest
@@ -45,6 +55,10 @@ public final class HttpSessionOAuth2AuthorizationRequestRepository
 
 	private final String sessionAttributeName = DEFAULT_AUTHORIZATION_REQUEST_ATTR_NAME;
 
+	private Clock clock = Clock.systemUTC();
+
+	private Duration oAuth2AuthorizationRequestExpiresIn = Duration.ofSeconds(120);
+
 	@Override
 	public OAuth2AuthorizationRequest loadAuthorizationRequest(HttpServletRequest request) {
 		Assert.notNull(request, "request cannot be null");
@@ -52,8 +66,10 @@ public final class HttpSessionOAuth2AuthorizationRequestRepository
 		if (stateParameter == null) {
 			return null;
 		}
-		Map<String, OAuth2AuthorizationRequest> authorizationRequests = this.getAuthorizationRequests(request);
-		return authorizationRequests.get(stateParameter);
+		Map<String, WrappedWithCreated<OAuth2AuthorizationRequest>> authorizationRequests = this
+				.getAuthorizationRequests(request);
+		WrappedWithCreated<OAuth2AuthorizationRequest> wrappedWithCreated = authorizationRequests.get(stateParameter);
+		return (wrappedWithCreated != null) ? wrappedWithCreated.wrapped : null;
 	}
 
 	@Override
@@ -67,8 +83,9 @@ public final class HttpSessionOAuth2AuthorizationRequestRepository
 		}
 		String state = authorizationRequest.getState();
 		Assert.hasText(state, "authorizationRequest.state cannot be empty");
-		Map<String, OAuth2AuthorizationRequest> authorizationRequests = this.getAuthorizationRequests(request);
-		authorizationRequests.put(state, authorizationRequest);
+		Map<String, WrappedWithCreated<OAuth2AuthorizationRequest>> authorizationRequests = this
+				.getAuthorizationRequests(request);
+		authorizationRequests.put(state, new WrappedWithCreated<>(this.clock.instant(), authorizationRequest));
 		request.getSession().setAttribute(this.sessionAttributeName, authorizationRequests);
 	}
 
@@ -79,15 +96,17 @@ public final class HttpSessionOAuth2AuthorizationRequestRepository
 		if (stateParameter == null) {
 			return null;
 		}
-		Map<String, OAuth2AuthorizationRequest> authorizationRequests = this.getAuthorizationRequests(request);
-		OAuth2AuthorizationRequest originalRequest = authorizationRequests.remove(stateParameter);
+		Map<String, WrappedWithCreated<OAuth2AuthorizationRequest>> authorizationRequests = this
+				.getAuthorizationRequests(request);
+		WrappedWithCreated<OAuth2AuthorizationRequest> wrappedWithCreatedOriginalRequest = authorizationRequests
+				.remove(stateParameter);
 		if (!authorizationRequests.isEmpty()) {
 			request.getSession().setAttribute(this.sessionAttributeName, authorizationRequests);
 		}
 		else {
 			request.getSession().removeAttribute(this.sessionAttributeName);
 		}
-		return originalRequest;
+		return (wrappedWithCreatedOriginalRequest != null) ? wrappedWithCreatedOriginalRequest.wrapped : null;
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


candrews:
@jgrandja can you please take a look at this PR?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
```diff
@@ -113,14 +132,62 @@ private String getStateParameter(HttpServletRequest request) {
...
 		return authorizationRequests;
 	}
 
+	/**
+	 * Sets the {@link Clock} used in {@link Instant#now(Clock)} when setting the instant
+	 * created for {@link OAuth2AuthorizationRequest}.
+	 * @param clock the clock
+	 * @since 5.5
+	 */
+	public void setClock(Clock clock) {
```
jgrandja, line 157: Let's not expose at this time.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
```diff
@@ -113,14 +132,62 @@ private String getStateParameter(HttpServletRequest request) {
...
+	}
+
+	/**
+	 * Sets the {@link Duration} for which {@link OAuth2AuthorizationRequest} should
+	 * expire.
+	 * @param oAuth2AuthorizationRequestExpiresIn the {@link Duration} a
+	 * {@link OAuth2AuthorizationRequest} is considered not expired. Must not be negative.
+	 * @since 5.5
+	 */
+	public void setOAuth2AuthorizationRequestExpiresIn(Duration oAuth2AuthorizationRequestExpiresIn) {
```
jgrandja, line 169: Rename to `authorizationRequestTimeToLive()`

jgrandja (changes requested):
Thanks for the PR @candrews ! Please see review comments.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
```diff
@@ -113,14 +132,62 @@ private String getStateParameter(HttpServletRequest request) {
...
+		Assert.notNull(oAuth2AuthorizationRequestExpiresIn, "oAuth2AuthorizationRequestExpiresIn cannot be null");
+		Assert.state(!oAuth2AuthorizationRequestExpiresIn.isNegative(),
+				"oAuth2AuthorizationRequestExpiresIn cannot be negative");
+		this.oAuth2AuthorizationRequestExpiresIn = oAuth2AuthorizationRequestExpiresIn;
+	}
+
+	private static final class WrappedWithCreated<T extends Serializable> implements Serializable {
```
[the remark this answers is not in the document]
candrews: > Let's rename to `OAuth2AuthorizationRequestReference`

sound good

> Also, is the `T` generic type declaration necessary?

Nope :)
And it's gone.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
```diff
@@ -113,14 +132,62 @@ private String getStateParameter(HttpServletRequest request) {
...
+
+		private static final long serialVersionUID = SpringSecurityCoreVersion.SERIAL_VERSION_UID;
+
+		private final Instant created;
```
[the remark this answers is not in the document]
candrews: :+1:

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
```diff
@@ -113,14 +132,62 @@ private String getStateParameter(HttpServletRequest request) {
...
+
+		private final T wrapped;
+
+		private WrappedWithCreated(Instant created, T wrapped) {
```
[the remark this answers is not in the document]
candrews: :+1:

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
```diff
@@ -113,14 +132,62 @@ private String getStateParameter(HttpServletRequest request) {
...
+			super();
```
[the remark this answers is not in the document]
candrews: :+1:

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
```diff
@@ -113,14 +132,62 @@ private String getStateParameter(HttpServletRequest request) {
...
+			this.created = created;
+			Assert.notNull(wrapped, "wrapped cannot be null");
```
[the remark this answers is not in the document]
candrews: :+1:

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepositoryTests.java
```diff
@@ -237,6 +242,32 @@ public void removeAuthorizationRequestWhenNotSavedThenNotRemoved() {
 		assertThat(removedAuthorizationRequest).isNull();
 	}
 
+	@Test
+	public void testExpiredAuthorizationRequestsRemoved() {
```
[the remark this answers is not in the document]
candrews: Renamed to `removeAuthorizationRequestWhenExpired`

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepositoryTests.java
```diff
@@ -16,6 +16,10 @@
 
 package org.springframework.security.oauth2.client.web;
 
+import java.time.Clock;
```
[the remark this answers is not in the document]
candrews: :+1:

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
```diff
@@ -16,23 +16,33 @@
 
 package org.springframework.security.oauth2.client.web;
 
+import java.io.Serializable;
```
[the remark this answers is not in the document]
candrews: :+1:

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
candrews: I changed to default (package) scope.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
candrews: :+1:

candrews:
> Thanks for the PR @candrews ! Please see review comments.

I've addressed feedback in a fixup commit I added. I'm assuming that the commits will be squashed when the PR is merged.

Thank you for your review!

candrews:
@jgrandja Is there anything else you'd like me to do?

Thanks again

candrews:
I've updated this PR to also set a maximum number of `OAuth2AuthorizationRequest` per session, which is important based on my own testing and was suggested at https://github.com/spring-projects/spring-security/issues/5145#issuecomment-528266887

@jgrandja can you please take a peek?

jgrandja:
@candrews Thanks. I will look at this shortly.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
```diff
@@ -113,14 +137,70 @@ private String getStateParameter(HttpServletRequest request) {
...
+
+	private static final class OAuth2AuthorizationRequestReference implements Serializable {
+
+		private static final long serialVersionUID = SpringSecurityCoreVersion.SERIAL_VERSION_UID;
+
+		private final Instant expiresAt;
+
+		private final OAuth2AuthorizationRequest wrapped;
+
+		private OAuth2AuthorizationRequestReference(OAuth2AuthorizationRequest wrapped, Instant created) {
```
jgrandja, line 198: created -> expiresAt

jgrandja, line 198: wrapped -> authorizationRequest

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
```diff
@@ -113,14 +137,70 @@ private String getStateParameter(HttpServletRequest request) {
...
 		return authorizationRequests;
 	}
 
+	/**
+	 * Sets the {@link Clock} used in {@link Instant#now(Clock)} when setting the instant
+	 * created for {@link OAuth2AuthorizationRequest}.
+	 * @param clock the clock
+	 * @since 5.5
+	 */
+	void setClock(Clock clock) {
```
jgrandja, line 159: Please remove this as it's not needed. The clock will always be in sync since this implementation operates independently within it's own application instance. Clock and clock skew is needed when components are interacting in a distributed way.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
```diff
@@ -113,14 +137,70 @@ private String getStateParameter(HttpServletRequest request) {
...
+		this.authorizationRequestTimeToLive = authorizationRequestTimeToLive;
+	}
+
+	/**
+	 * Sets the maximum number of {@link OAuth2AuthorizationRequest} that can be
+	 * stored/active for a session. If the maximum number are present in a session when an
+	 * attempt is made to save another one, then the oldest will be removed.
+	 * @param maxActiveAuthorizationRequests must not be negative.
+	 */
+	public void setMaxActiveAuthorizationRequestsPerSession(int maxActiveAuthorizationRequestsPerSession) {
```
jgrandja, line 184: Can you explain the specific use case why this is needed?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
```diff
@@ -45,15 +55,22 @@
...
 	public OAuth2AuthorizationRequest loadAuthorizationRequest(HttpServletRequest request) {
 		Assert.notNull(request, "request cannot be null");
 		String stateParameter = this.getStateParameter(request);
 		if (stateParameter == null) {
 			return null;
 		}
-		Map<String, OAuth2AuthorizationRequest> authorizationRequests = this.getAuthorizationRequests(request);
-		return authorizationRequests.get(stateParameter);
+		Map<String, OAuth2AuthorizationRequestReference> authorizationRequests = this.getAuthorizationRequests(request);
+		OAuth2AuthorizationRequestReference wrappedWithCreated = authorizationRequests.get(stateParameter);
```
jgrandja, line 72: Please rename all variables that have `wrappedWith*`

jgrandja (changes requested):
Thanks for the updates @candrews. Please see review comments.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
candrews: It's possible to easily put a lot of `OAuth2AuthorizationRequest`s in this map, resulting in Bad Things :tm: happening. This limit prevents that from being possible.

I contacted security@pivotal.io with the details. Let me know if I should post them here.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
candrews: This method is package private, so nothing other than Spring itself should be using it.

Being able to set the clock is important for testing. See https://github.com/candrews/spring-security/blob/9b1a85d992f3ac0d725f67ac03ee6d9f44659b21/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepositoryTests.java#L249

Is there another approach that should be taken to make the clock available to be set by tests?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
candrews: :+1:

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java
candrews: :+1:

candrews: :+1:

candrews:
How's this looking now?

jgrandja:
@candrews I've been giving this enhancement some further thought and I would prefer to keep this a patch change.

Something about `setAuthorizationRequestTimeToLive()`, `setMaxActiveAuthorizationRequestsPerSession()`, and `setClock()` doesn't feel right from a design standpoint.

Furthermore, I'm not sure how much these configuration-related settings will be used given that gh-5145 did not receive any upvotes and the issue is now over 3 years old.

`HttpSessionOAuth2AuthorizationRequestRepository` is designed to be a simplified (default) version of `AuthorizationRequestRepository`. However, if an application requires more advanced capabilities than it can configure a custom `AuthorizationRequestRepository` via [oauth2Client() or oauth2Login()](https://docs.spring.io/spring-security/site/docs/current/reference/html5/#storing-the-authorization-request).

Can you provide me more details on your use case and why you need this enhancement?

jgrandja:
@candrews 

> If you'd like me to give the specifics, let me know and I will.

No, I'm good. I got the message and much appreciated.

I have another idea brewing for this patch. Let me think it over a bit. I'll get back to you by end of day tomorrow the latest.

jgrandja:
@candrews Here is a template for the solution.

I've added implementation notes but let me know if you have any questions.


```java
/*
 * 1) OAuth2AuthorizationRequest's are stored as Map<String, OAuth2AuthorizationRequestWrapper>, keyed by getState()
 * 2) When saveAuthorizationRequest(), constrain limit by grouping by OAuth2AuthorizationRequestWrapper.getRegistrationId().
 * 		For example, if there are 3 existing `google-registration-id` then reject the save by throwing IllegalStateException.
 * 		I think constraining to 3 for the limit-based strategy is ok?
 * 3) When loadAuthorizationRequest(), remove all isExpired() before returning the Map
 */
private static class OAuth2AuthorizationRequestWrapper implements Serializable {

	private static final Duration DEFAULT_TIME_TO_LIVE = Duration.ofMinutes(5);

	private final OAuth2AuthorizationRequest authorizationRequest;

	private final Instant createdAt;

	private final Instant expiresAt;

	private OAuth2AuthorizationRequestWrapper(OAuth2AuthorizationRequest authorizationRequest) {
		this.authorizationRequest = authorizationRequest;
		this.createdAt = Instant.now();
		// I think 5 minutes is long enough to allow for the authorization (consent) flow to complete?
		this.expiresAt = this.createdAt.plus(DEFAULT_TIME_TO_LIVE);
	}

	private OAuth2AuthorizationRequest getAuthorizationRequest() {
		return this.authorizationRequest;
	}

	private String getRegistrationId() {
		return this.authorizationRequest.getAttribute(OAuth2ParameterNames.REGISTRATION_ID);
	}

	private String getState() {
		return this.authorizationRequest.getState();
	}

	private boolean isExpired() {
		// NOTE:
		// Given that Instant.now() uses the system clock and
		// createdAt and expiresAt is instantiated by the same system clock,
		// there is no need for a configurable Clock or any clock skew adjustment
		return Instant.now().isAfter(this.expiresAt);
	}
}
```

candrews:
>  1) OAuth2AuthorizationRequest's are stored as Map<String, OAuth2AuthorizationRequestWrapper>, keyed by getState()

That's already in place. Based on earlier feedback, the class is named `OAuth2AuthorizationRequestReference`, not `OAuth2AuthorizationRequestWrapper`, though.

>  2) When saveAuthorizationRequest(), constrain limit by grouping by OAuth2AuthorizationRequestWrapper.getRegistrationId().
>  	For example, if there are 3 existing `google-registration-id` then reject the save by throwing IllegalStateException.
>  		I think constraining to 3 for the limit-based strategy is ok?

If we throw `IllegalStateException`, then the user won't be able to authorize using that registration id for the remainder of their session or until the 2 minute expiration occurs clearing that `OAuth2AuthorizationRequestWrapper` from the map.

Perhaps instead of throwing `IllegalStateException`, we should remove the oldest `OAuth2AuthorizationRequestWrapper` for that registration id and continue?

I've just made that change and pushed the commit for it.

>  3) When loadAuthorizationRequest(), remove all isExpired() before returning the Map

Already done.

I kept `setClock` though as a package private method, as it facilitates testing. If you have a better approach, please let me know.

Also, the commit history is getting excessive. Would you like me to squash all the commits and force push update this PR?

candrews:
@jgrandja here's an alternative, very simple approach that also addresses the root issue:
https://github.com/spring-projects/spring-security/pull/9649

jgrandja:
@candrews I'll close this one in favour of gh-9649

PR #9521, opened 2021-03-25
Update ComparableVersion to version from Maven 3.6.3

candrews:
Update ComparableVersion to be a copy of the class in Maven 3.6.3.

Among other improvements, this includes two problems of potential interest:
* https://github.com/apache/maven/commit/3ac9718f8267b2fb9636f1a58acc6bb53356c329 "[MNG-6655] Add support for "release" qualifier "
* https://github.com/apache/maven/commit/445421051c5d364839d3bcdf2fa35e3ecd06e9bc "potential NPE"

I discovered that this class was out of date when I ran a static analysis tool which identified the NPE issue.

### DIFF
--- a/core/src/main/java/org/springframework/security/core/ComparableVersion.java
+++ b/core/src/main/java/org/springframework/security/core/ComparableVersion.java
@@ -17,27 +17,26 @@
 package org.springframework.security.core;
 
 import java.math.BigInteger;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


rwinch:
Thanks for the PR @candrews I've merged this into master as an enhancement. We only use this internally for validating Spring version compatibility, so the possible inputs is fairly limited and unlikely to find the NPE or need the enhancement. If we actually get a reproducible bug, we can consider backporting it.

PR #9524, opened 2021-03-28
csrf verified thrown  NullPointerException

Exclamation-mark:
Fix:
When CSRF verification is turned on, it is not carried in the request, NullPointerException exception is always thrown when CSRF parameter is used

### DIFF
--- a/crypto/src/main/java/org/springframework/security/crypto/encrypt/AesBytesEncryptor.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/encrypt/AesBytesEncryptor.java
@@ -52,14 +52,40 @@ public final class AesBytesEncryptor implements BytesEncryptor {
 
 	private static final String AES_GCM_ALGORITHM = "AES/GCM/NoPadding";
 
+	/**
+	 * Constructs an encryptor that uses AES encryption. Example: <code>
+	 * AesBytesEncryptor encryptor = new AesBytesEncryptor(yourPassword, 5c0744940b5c369b);
+	 * </code>
+	 * @param password the password value
+	 * @param salt the hex-encoded salt value
+	 */
 	public AesBytesEncryptor(String password, CharSequence salt) {
 		this(password, salt, null);
 	}
 
+	/**
+	 * Constructs an encryptor that uses AES encryption. Example: <code>
+	 * AesBytesEncryptor encryptor =
+	 * 		new AesBytesEncryptor(yourPassword, 5c0744940b5c369b, KeyGenerators.secureRandom(16));
+	 * </code>
+	 * @param password the password value
+	 * @param salt the hex-encoded salt value
+	 * @param ivGenerator the generator used to generate the initialization vector
+	 */
 	public AesBytesEncryptor(String password, CharSequence salt, BytesKeyGenerator ivGenerator) {
 		this(password, salt, ivGenerator, CipherAlgorithm.CBC);
 	}
 
+	/**
+	 * Constructs an encryptor that uses AES encryption. Example: <code>
+	 * AesBytesEncryptor encryptor =
+	 * 		new AesBytesEncryptor(yourPassword, 5c0744940b5c369b, KeyGenerators.secureRandom(16), CipherAlgorithm.GCM);
+	 * </code>
+	 * @param password the password value
+	 * @param salt the hex-encoded salt value
+	 * @param ivGenerator the generator used to generate the initialization vector
+	 * @param alg the {@link CipherAlgorithm} to be used
+	 */
 	public AesBytesEncryptor(String password, CharSequence salt, BytesKeyGenerator ivGenerator, CipherAlgorithm alg) {
 		this(CipherUtils.newSecretKey("PBKDF2WithHmacSHA1",
 				new PBEKeySpec(password.toCharArray(), Hex.decode(salt), 1024, 256)), ivGenerator, alg);


pivotal-issuemaster:
@Exclamation-mark Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=9524)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=9524) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@Exclamation-mark Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=9524)!

jzheaux:
Hi, @Exclamation-mark, thanks for the PR.

Do you have a test or a sample that confirms this? The reason I ask is because when I read the code I see:

```java
private static boolean equalsConstantTime(String expected, String actual) {
    byte[] expectedBytes = bytesUtf8(expected);
    byte[] actualBytes = bytesUtf8(actual);
    return MessageDigest.isEqual(expectedBytes, actualBytes); // checks for null in isEqual
}

private static byte[] bytesUtf8(String s) {
    return (s != null) ? Utf8.encode(s) : null; // checks for null
}
```

When I run a simple test on `master` that calls `CsrfFilter.equalsConstantTime("asdf", null)`, it does not throw an NPE, so I'm missing how the proposed change is necessary.

Exclamation-mark:
Thank you for your reply.
I guess your JDK version is higher than 8.
The following code will throw NullPointerException in jdk8, but not in jdk11.
```java
/**
 * The following code restores the workflow of the equalsConstantTime method in CsrFilter
 * */
public class TestCsrfFilterEquals {

    public static void main(String[] args) {
        System.out.println("asdf and null: " + equalsConstantTime("asdf", null));
        System.out.println("asdf and asdf: " + equalsConstantTime("asdf", "asdf"));
    }

    private static final Charset CHARSET = StandardCharsets.UTF_8;

    /**
     * Get the bytes of the String in UTF-8 encoded form.
     */
    public static byte[] encode(CharSequence string) {
        try {
            ByteBuffer bytes = CHARSET.newEncoder().encode(CharBuffer.wrap(string));
            byte[] bytesCopy = new byte[bytes.limit()];
            System.arraycopy(bytes.array(), 0, bytesCopy, 0, bytes.limit());
            return bytesCopy;
        }
        catch (CharacterCodingException ex) {
            throw new IllegalArgumentException("Encoding failed", ex);
        }
    }

    private static boolean equalsConstantTime(String expected, String actual) {
        byte[] expectedBytes = bytesUtf8(expected);
        byte[] actualBytes = bytesUtf8(actual);
        return MessageDigest.isEqual(expectedBytes, actualBytes);
    }

    private static byte[] bytesUtf8(String s) {
        // need to check if Utf8.encode() runs in constant time (probably not).
        // This may leak length of string.
        return (s != null) ? encode(s) : null;
    }
}
```
The root cause of this problem is the different implementation of java.security.MessageDigest.isEqual() method in different versions of JDK.The implementation of this method in jdk8 does not verify whether the two arrays are empty.
```java

  /**
   * MessageDigest.isEqual()'s implementation in jdk 1.8.0_40
   * */
   public static boolean isEqual(byte[] digesta, byte[] digestb) {
        if (digesta.length != digestb.length) {
            return false;
        }

        int result = 0;
        // time-constant comparison
        for (int i = 0; i < digesta.length; i++) {
            result |= digesta[i] ^ digestb[i];
        }
        return result == 0;
    }
```

jzheaux:
Thanks for the info, @Exclamation-mark. Is there something about 8u45 that makes it difficult to upgrade to it (or the latest)?

Exclamation-mark:
> Thanks for the info, @Exclamation-mark. Is there something about 8u45 that makes it difficult to upgrade to it (or the latest)?

It's easy to upgrade on Linux. However, JDK‘s Installation package on windows is not as rich as Linux(in China). 
In addition, I think my commit can make CsrfFilter more perfect (it does’t increase the judgment of null, but ensures that NPE will not be thrown).
Of course, I'm not sure if the work done for this small optimization is worth it. It's up to you to decide.

web/src/main/java/org/springframework/security/web/csrf/CsrfFilter.java
```diff
@@ -174,17 +174,18 @@ public void setAccessDeniedHandler(AccessDeniedHandler accessDeniedHandler) {
 	 * @return
 	 */
 	private static boolean equalsConstantTime(String expected, String actual) {
-		byte[] expectedBytes = bytesUtf8(expected);
-		byte[] actualBytes = bytesUtf8(actual);
+		if (actual == null) {
```
jzheaux, line 177: I think the code could catch more cases by doing:

```java
if (expected == actual) {
    return true;
}
if (expected == null || actual == null) {
    return false;
}
```

`MessageDigest` follows the above pattern when comparing the `byte` arrays.

jzheaux (changes requested):
Thanks for the PR, @Exclamation-mark, and the extra information you provided about JDK versions.

I've left some feedback inline. In addition to that feedback, would you please do the following:

1. Add a test to `CsrfFilterTests` that would fail before your change, but passes after
2. Update the copyright year to `2021` to any files that you change
3. Squash your commits and [format the resulting commit's message](https://github.com/spring-projects/spring-security/blob/3c33535681ccb39ea94f5a68f6bcf7f6e75af5d9/CONTRIBUTING.adoc#format-commit-messages)

web/src/main/java/org/springframework/security/web/csrf/CsrfFilter.java
Exclamation-mark: ok

Exclamation-mark: ok.

Exclamation-mark: ok.
This PR was accidentally closed by me. I submitted a new pr #9561. 
Also, when I was running tests with a lower version of JDK, I found that csrfwebfilter had the same NPE captured, so I made the same modification to csrfwebfilter.
Also, thank you for your guidance. This is my first time to participate in an open source project.

rwinch:
Duplicate of gh-9561

Issue #9334, opened 2021-01-09
web.ignoring().mvcMatchers is confuse in someway about the debug output in the console


manueljordan:
Hello 

Not sure if the following is a possible bug.

Having 

* Spring Boot `2.4.1` 
* Spring dev tools `2.4.1`
* Spring Security `5.4.2`

I understand the following:

```
@Override
public void configure(WebSecurity web) throws Exception {
	web.ignoring().mvcMatchers("/resources/static/css/**", "/static/css/**");
}
```
Ignore the second url pattern (it would be the same than the first). 
I know that it instructs to Spring Security to ignore that URLs to avoid any kind of security.

When I use `logging.level.org.springframework.security=debug`

I can see in the console

**Alpha**

```
[  restartedMain] o.s.s.web.DefaultSecurityFilterChain     : Will secure Mvc [pattern='/resources/static/css/**'] with []
[  restartedMain] o.s.s.web.DefaultSecurityFilterChain     : Will secure Mvc [pattern='/static/css/**'] with []
```

If those URLs are supposed to be ignored, why appears the `Will secure Mvc` term? it gives the impression they are being secured yet

Furthermore when the app is used appears for a simple GET to render the home page:

**Beta**

```
[nio-8080-exec-3] o.s.security.web.FilterChainProxy        : Securing GET /demo/browser/home
[nio-8080-exec-3] w.c.HttpSessionSecurityContextRepository : Retrieved SecurityContextImpl [...]
[nio-8080-exec-3] s.s.w.c.SecurityContextPersistenceFilter : Set SecurityContextHolder to SecurityContextImpl [...]
[nio-8080-exec-3] o.s.s.w.a.i.FilterSecurityInterceptor    : Authorized filter invocation [GET /demo/browser/home] with attributes [authenticated]
[nio-8080-exec-3] o.s.security.web.FilterChainProxy        : Secured GET /demo/browser/home
[nio-8080-exec-3] s.s.w.c.SecurityContextPersistenceFilter : Cleared SecurityContextHolder to complete request
[nio-8080-exec-4] o.s.security.web.FilterChainProxy        : Securing GET /demo/css/general.css
[nio-8080-exec-4] w.c.HttpSessionSecurityContextRepository : Retrieved SecurityContextImpl [...]
[nio-8080-exec-4] s.s.w.c.SecurityContextPersistenceFilter : Set SecurityContextHolder to SecurityContextImpl [...]
[...]

PR #9526, opened 2021-03-29
Print ignore message DefaultSecurityFilterChain

manueljordan:
When either `web.ignoring().mvcMatchers(...)` or
`web.ignoring().antMatchers(...)` methods are used, for all their
variations, the DefaultSecurityFilterChain class now indicates
correctly through its ouput what paths are ignored according the
`ignoring()` settings.

Closes gh-9334

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/builders/WebSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/builders/WebSecurity.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,6 +17,7 @@
 package org.springframework.security.config.annotation.web.builders;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 import javax.servlet.Filter;
@@ -29,6 +30,7 @@ import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
+import org.springframework.core.log.LogMessage;
 import org.springframework.http.HttpMethod;
 import org.springframework.security.access.PermissionEvaluator;
 import org.springframework.security.access.expression.SecurityExpressionHandler;
@@ -55,6 +57,7 @@ import org.springframework.security.web.debug.DebugFilter;
 import org.springframework.security.web.firewall.HttpFirewall;
 import org.springframework.security.web.firewall.RequestRejectedHandler;
 import org.springframework.security.web.firewall.StrictHttpFirewall;
+import org.springframework.security.web.server.restriction.IgnoreRequestMatcher;
 import org.springframework.security.web.servlet.util.matcher.MvcRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher;
 import org.springframework.util.Assert;
@@ -101,7 +104,7 @@ public final class WebSecurity extends AbstractConfiguredSecurityBuilder<Filter,
 
 	private WebInvocationPrivilegeEvaluator privilegeEvaluator;
 
-	private DefaultWebSecurityExpressionHandler defaultWebSecurityExpressionHandler = new DefaultWebSecurityExpressionHandler();
+	private final DefaultWebSecurityExpressionHandler defaultWebSecurityExpressionHandler = new DefaultWebSecurityExpressionHandler();
 
 	private SecurityExpressionHandler<FilterInvocation> expressionHandler = this.defaultWebSecurityExpressionHandler;
 
@@ -373,6 +376,8 @@ public final class WebSecurity extends AbstractConfiguredSecurityBuilder<Filter,
 		@Override
 		public MvcMatchersIgnoredRequestConfigurer mvcMatchers(HttpMethod method, String... mvcPatterns) {
 			List<MvcRequestMatcher> mvcMatchers = createMvcMatchers(method, mvcPatterns);
+			Arrays.asList(mvcPatterns).stream().forEach((t) -> printWarnSecurityMessage(method, t));
+			mvcMatchers.stream().forEach((t) -> t.ignore());
 			WebSecurity.this.ignoredRequests.addAll(mvcMatchers);
 			return new MvcMatchersIgnoredRequestConfigurer(getApplicationContext(), mvcMatchers);
 		}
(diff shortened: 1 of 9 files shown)


web/src/main/java/org/springframework/security/web/DefaultSecurityFilterChain.java
```diff
@@ -48,7 +49,18 @@ public DefaultSecurityFilterChain(RequestMatcher requestMatcher, Filter... filte
 	}
 
 	public DefaultSecurityFilterChain(RequestMatcher requestMatcher, List<Filter> filters) {
-		logger.info(LogMessage.format("Will secure %s with %s", requestMatcher, filters));
+		if (requestMatcher instanceof IgnoreRequestMatcher) {
```
jzheaux, line 52: Maybe we could start with something simpler like this:

```java
if (filters.isEmpty()) {
    logger.info(LogMessage.format("Will not secure %s", requestMatcher));
} else {
    logger.info(LogMessage.format("Will secure %s with %s", requestMatcher, filters));
}
```
If Spring Security has no filters assigned to a given request matcher, then it's true that Spring Security will not be securing that request.

config/src/main/java/org/springframework/security/config/annotation/web/builders/WebSecurity.java
```diff
@@ -395,6 +432,33 @@ public WebSecurity and() {
 			return WebSecurity.this;
 		}
 
+		/**
+		 * @param method the HttpMethod, it could be null too.
+		 * @param pathPattern the path pattern to be ignored
+		 * @since 5.5
+		 */
+		private void printWarnSecurityMessage(HttpMethod method, String pathPattern) {
```
jzheaux, line 440: This warning message may need to wait as there are numerous other scenarios to cover in order for that to be complete, e.g. `regexMatchers`, `anyRequest()`, and `requestMatcher(AnyRequestMatcher.INSTANCE)`. It may be best to begin with the simplest case and do:

```java
@Override
public IgnoredRequestConfigurer anyRequest() {
	WebSecurity.this.logger.warn("Applying explicit instruction to ignore all paths");
	WebSecurity.this.logger.warn("This disables Spring Security and is not recommended");
	return super.anyRequest();
}
```

This aligns with other configured code that guards usage of `anyRequest()` but does not also attempt to parse strings to discern the intent of the application. Later on, it may be possible for the code to be more intelligent without committing to an increased public API.

jzheaux (changes requested):
Thanks for the PR, @manueljordan! 

I'm hesitant to add this much code to support logging. In this PR, we should not:

1. Increase the visibility of existing APIs
2. Introduce new interfaces

I've left some additional feedback inline.

manueljordan:
Hello @jzheaux 

Thanks for the quick reply

>Increase the visibility of existing APIs

If I do that it breaks the DRY principle. The changes of visibility applies mostly for _utils_ methods (static methods), here the situation is about the nested static class, it is tricky, I thought to move it to an outer/external class but I could be more problematic

> Introduce new interfaces

I thought to work based on hierarchy (subclass), but since I did realize it applies only to 2 subclasses of the whole tree hierarchy, I took that decision, it opens the same option to ignore something if something new appears in the future, such as: _somethingMatchers()_


About

```
if (filters.isEmpty()) {
    logger.info(LogMessage.format("Will not secure %s", requestMatcher));
} else {
    logger.info(LogMessage.format("Will secure %s with %s", requestMatcher, filters));
}
```

I thought the same, but - if my memory does not fail me in some scenarios the _filters_ appears empty and not empty (yes, for _ignoring()_), so that approach was discarded, so the key is work with the `RequestMatcher` type, where now two of them are of the `IgnoreRequestMatcher` type.

> This warning message may need to wait as there are numerous other scenarios to cover in order for that to be complete, e.g. regexMatchers, anyRequest(), and requestMatcher(AnyRequestMatcher.INSTANCE). It may be best to begin with the simplest case and do:

That working scenario only applies to `web.ignoring().mvcMatchers(...)` and `web.ignoring().antMatchers(...)`, if my memory does not fail me `web.ignoring().X` applies only to `mvcMatchers(...)` and `antMatchers(...)`.  Furthermore I confirmed through the `@Test` methods that `configure(WebSecurity)` _overrides by complete_ the `configure(HttpSecurity)` settings, so that warning message is important.

For the other methods mentioned, they do not have a common interface to identity quickly if is for an ignore scenario or not. It is the problem from the beginning, there is no a quick way to know if a URL (path) is for ignoring or not.

I am open to follow your guidance, you are the expert here.

Consider remember, I am starting to play with this API practically from the scratch, and your vision is bigger than my current knowledge, but I explained the experiences and the decisions taken according each case

jzheaux:
@manueljordan, sorry for the delay. I'm getting back to this PR now.

> If I do that it breaks the DRY principle. 

Either that, or we need to simplify what we are trying to do. If we are changing the public API in this PR, we are probably trying to go too far too fast.

> I thought the same, but - if my memory does not fail me in some scenarios the filters appears empty and not empty

Either way, if there are no filters for a given request matcher, then requests that match are being effectively ignored.

manueljordan:
Hello @jzheaux 

>Either that, or we need to simplify what we are trying to do. If we are changing the public API in this PR, we are probably trying to go too far too fast.

What do you suggest do here? You are the expert here - and I always take your suggestions friendly. I hope you take in the same way my thoughts. But according with my research in fix this, fall in break the DRY would be an option but you know ... break something like that is not wise ... it seems you want that I should split my current PR in more PRs covering each case by separate. Am I correct?

>Either way, if there are no filters for a given request matcher, then requests that match are being effectively ignored.

Your approach even if is valid - with my PR I am giving an important warning to the developer about security constraints that are being overriding and therefore exposing private things. Do you want split that warning messages how other PR? - something I like wrote in the previous sentence...

Let me know your thoughts

jzheaux:
> What do you suggest here?

I'd prefer we go with the comments I made [in my review](https://github.com/spring-projects/spring-security/pull/9526#pullrequestreview-623663042).

> with my PR I am giving an important warning to the developer

Agreed. I think it's reasonable to add an even stronger message:

```java
// ...
for (RequestMatcher ignoredRequest : this.ignoredRequests) {
    WebSecurity.this.logger.warn("You are asking Spring Security to ignore " + matcher + ". This is not recommended -- please use permitAll via HttpSecurity#authorizeHttpRequests instead.");
    securityFilterChains.add(new DefaultSecurityFilterChain(ignoredRequest));
}
// ...
```

My point is not that we shouldn't warn, but that this PR should not introduce so much infra to facilitate logging. We should be able to add some sensible warnings without copy-pasting and without affecting the public API.

jzheaux:
@manueljordan Do you feel like you can make the changes requested? If not, no worries, I can add a polish commit.

jzheaux:
This is now merged into `5.7.x` via 01ed617d5f33afd832233dca9f010253e7a6d3a3 and as well as the discussed polish commit cbd87fac898ea365e6f1cb07fa019684f23d6b37. The main thrust of the polish commit is to still alert the user to this bad practice while leaving the public API intact.

Issue #9514, opened 2021-03-23
DelegatingSecurityContextTaskScheduler does not allow to use a specific SecurityContext


gbaso:
**Expected Behavior**

`DelegatingSecurityContextTaskScheduler` should have a constructor accepting a `TaskScheduler` and a `SecurityContext`.

**Current Behavior**

Unlike similar classes like `DelegatingSecurityContextAsyncTaskExecutor` and `DelegatingSecurityContextScheduledExecutorService`, `DelegatingSecurityContextTaskScheduler` does not extends `AbstractDelegatingSecurityContextSupport` and does not contains a `SecurityContext` field. As of now, it's just a plain wrapper.

Looking at past issues, it looks like `DelegatingSecurityContextTaskScheduler` was requested in #3125 and introduced in #6043

jzheaux:
Thanks for the report, @gbaso. I agree that `DelegatingSecurityContextTaskScheduler` should wrap the `Runnable` instances it receives in `DelegatingSecurityContextRunnable`s.

Would you be able to provide a PR with that change?

gbaso:
@jzheaux I submitted #9532 to fix the issue

PR #9532, opened 2021-03-31
Add SecurityContext to delegating TaskScheduler

gbaso:
Wrap DelegatingSecurityContextTaskScheduler's Runnable tasks in
DelegatingSecurityContextRunnables, allowing to specify a
SecurityContext to use for tasks execution.

Closes gh-9514

### DIFF
--- a/core/src/main/java/org/springframework/security/scheduling/DelegatingSecurityContextTaskScheduler.java
+++ b/core/src/main/java/org/springframework/security/scheduling/DelegatingSecurityContextTaskScheduler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,8 +19,12 @@ package org.springframework.security.scheduling;
 import java.util.Date;
 import java.util.concurrent.ScheduledFuture;
 
+import org.springframework.core.task.TaskExecutor;
 import org.springframework.scheduling.TaskScheduler;
 import org.springframework.scheduling.Trigger;
+import org.springframework.security.concurrent.DelegatingSecurityContextRunnable;
+import org.springframework.security.core.context.SecurityContext;
+import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/springframework/security/scheduling/DelegatingSecurityContextTaskScheduler.java
```diff
@@ -30,47 +35,64 @@
  * @author Richard Valdivieso
  * @since 5.1
  */
-public class DelegatingSecurityContextTaskScheduler implements TaskScheduler {
+public class DelegatingSecurityContextTaskScheduler extends AbstractDelegatingSecurityContextSupport
```
jzheaux, line 38: Ideally, this class should favor composition over inheritance. Since `DelegatingSecurityContextCallable` and `DelegatingSecurityContextRunnable` have factory methods that do what you need, would you be able to inline the support's `wrap` methods here in `DelegatingSecurityContextTaskScheduler`?

core/src/main/java/org/springframework/security/scheduling/DelegatingSecurityContextTaskScheduler.java
```diff
@@ -30,47 +35,64 @@
...
+		implements TaskScheduler {
 
-	private final TaskScheduler taskScheduler;
+	private final TaskScheduler delegate;
```
jzheaux, line 41: Since this variable name change is not necessary to close the ticket, would you please move this change to a separate commit in this PR?

core/src/test/java/org/springframework/security/scheduling/CurrentDelegatingSecurityContextTaskSchedulerTests.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Tests using the current {@link SecurityContext} on
+ * {@link DelegatingSecurityContextTaskScheduler}
+ *
+ * @author Giacomo Baso
+ * @since 5.5
+ *
+ */
+public class CurrentDelegatingSecurityContextTaskSchedulerTests
```
jzheaux, line 31: Is there a reason that these task scheduler test classes should be separated like this? Most Spring Security test classes follow the pattern `XXXTests` where `XXX` is the class under test.

core/src/test/java/org/springframework/security/scheduling/CurrentDelegatingSecurityContextTaskSchedulerTests.java
```diff
@@ -0,0 +1,44 @@
...
+		extends AbstractDelegatingSecurityContextTaskSchedulerTests {
+
+	@Before
+	public void setUp() throws Exception {
+		super.currentSecurityContextPowermockSetup();
```
jzheaux, line 36: While I realize it was done this way in the past, I don't think that mocking the `create` methods for `DelegatingSecurityContextCallable` and `DelegatingSecurityContextRunnable` is necessary.

jzheaux (changes requested):
Thanks for the PR, @gbaso! I've left some feedback inilne.

core/src/test/java/org/springframework/security/scheduling/CurrentDelegatingSecurityContextTaskSchedulerTests.java
gbaso: Other related test classes follow the same pattern, in particular `AbstractDelegatingSecurityContextExecutorTests` and subclasses. This allows repeating the tests with a different configuration, i.e. to use the current `SecurityContext` or a provided one.

Would you suggest a different approach?

gbaso: Given that those test patterns do not actually check for the use of a specific `SecurityContext` and your comment regarding the mocking of `DelegatingSecurityContextCallable#create`, I've decided to revert to the old tests suite `DelegatingSecurityContextTaskSchedulerTests`, with the addition of a mocked `SecurityContext`.

core/src/main/java/org/springframework/security/scheduling/DelegatingSecurityContextTaskScheduler.java
gbaso: Moved to ba53e5d

core/src/test/java/org/springframework/security/scheduling/CurrentDelegatingSecurityContextTaskSchedulerTests.java
gbaso: See https://github.com/spring-projects/spring-security/pull/9532#discussion_r608824828

core/src/main/java/org/springframework/security/scheduling/DelegatingSecurityContextTaskScheduler.java
gbaso: wrap is not inlined in 1851a42

core/src/test/java/org/springframework/security/scheduling/DelegatingSecurityContextTaskSchedulerTests.java
```diff
@@ -58,7 +61,8 @@
 	@Before
 	public void setup() {
 		MockitoAnnotations.initMocks(this);
-		this.delegatingSecurityContextTaskScheduler = new DelegatingSecurityContextTaskScheduler(this.scheduler);
+		this.delegatingSecurityContextTaskScheduler = new DelegatingSecurityContextTaskScheduler(this.scheduler,
```
jzheaux, line 64: Should there be a test that confirms that the security context was indeed used?

core/src/main/java/org/springframework/security/scheduling/DelegatingSecurityContextTaskScheduler.java
```diff
@@ -32,45 +36,66 @@
...
+		return this.delegate.scheduleWithFixedDelay(wrap(task), startTime, delay);
 	}
 
 	@Override
 	public ScheduledFuture<?> scheduleWithFixedDelay(Runnable task, long delay) {
-		return this.taskScheduler.scheduleWithFixedDelay(task, delay);
+		return this.delegate.scheduleWithFixedDelay(wrap(task), delay);
+	}
+
+	protected final Runnable wrap(Runnable delegate) {
```
jzheaux, line 97: Let's please leave this private for now until there is a clear use case for making it protected. Users who extend this class can easily create their own wrap method.

jzheaux (changes requested):
Thanks for the updates, @gbaso! I've left some additional feedback inline.

core/src/main/java/org/springframework/security/scheduling/DelegatingSecurityContextTaskScheduler.java
gbaso: `wrap` is now private

core/src/test/java/org/springframework/security/scheduling/DelegatingSecurityContextTaskSchedulerTests.java
gbaso: I've added two tests in d455211, for explicit/current security context

jzheaux (changes requested):
Thanks for the updates, @gbaso!

Thanks for your descriptive commit messages. To simplify maintenance, will you please squash your commits into a single atomic commit?

gbaso:
@jzheaux PR should be ready now

jzheaux:
Thanks, @gbaso! I've scheduled this for 5.6.0-M1.

jzheaux:
Thanks again, @gbaso! This is now merged into `main` via 80743a267c91a8556e3e2c59e698d9b8ee4fc199. I also added a small polish to add the `@since` attribute to the new constructor's JavaDoc via e1e31939a34551a74e5d89318dcc240e7218f176

Issue #6053, opened 2018-11-07
Support JWT as an Authorization Grant for client


jgrandja:
This feature will partially implement [JSON Web Token (JWT) Profile for OAuth 2.0 Client Authentication and Authorization Grants](https://tools.ietf.org/html/rfc7523).

[Section 2.1. Using JWTs as Authorization Grants](https://tools.ietf.org/html/rfc7523#section-2.1) will be the focus for this feature implementation.

> JWT Bearer Token can be used to request
   an access token when a client wishes to utilize an existing trust
   relationship, expressed through the semantics of the JWT, without a
   direct user-approval step at the authorization server

One of the primary use cases for using a JWT as an authorization grant is to **_exchange_** it for another JWT (at the Token Endpoint) with narrowed `scope`. This is useful when a service (a) wants to call another downstream service (b) with only` scope` that service (b) understands (supports).

**NOTE:** This ticket addresses client-side support only.

Related #5199 #8175

gandrade:
Would be great have this feature targeting next milestone.

ThomasVitale:
Hi @jgrandja, if someone was available to support me in clarifying and reviewing the technical design for this feature, then I would like to work on it. I am already familiar with the OAuth 2.0 specs and work daily on security features for enterprise applications. But since it would be my first feature-size task on this project, I could use some guidance.

jgrandja:
@ThomasVitale Thanks for the offer! I actually have a POC in this [sample](https://github.com/jgrandja/oauth2-protocol-patterns/blob/master/microservice-b/src/main/java/org/springframework/security/oauth2/client/JwtBearerOAuth2AuthorizedClientProvider.java#L41).

However, I think it would still be valuable to review the spec and the POC code to determine if there is any other work left to do. I believe it's pretty close to being done but there may be other work that needs to be completed.

Let me know what you think?

matt-domsch-sp:
FYI, the OpenID Foundation Fast Federation Working Group is defining a standard that will require RFC 7523 support. Lack of this in Spring will slow development and adoption of this emerging standard.
https://bitbucket.org/openid/fastfed/src/master/text_spec/FastFed-1.0-Draft-01.txt

jgrandja:
Thanks for the info @matt-domsch-sp. Would you happen to know which providers currently support RFC 7523?

matt-domsch-sp:
I'm aware of these uses in the wild:

[...]

PR #9535, opened 2021-04-01
Add jwt-bearer authorization grant

jgrandja:
Add `jwt-bearer` authorization grant support.

### Related specifications

**RFC 7523**
[2.1 Using JWTs as Authorization Grants](https://tools.ietf.org/html/rfc7523#section-2.1)

**RFC 7521**
[4.1 Using Assertions as Authorization Grants](https://tools.ietf.org/html/rfc7521#section-4.1)

Closes gh-6053


jgrandja:
Merged via 7694aa2 eff4cdc

H-LREB:
@jgrandja Thank you for the merge.

I have a question about the class [JwtBearerOAuth2AuthorizedClientProvider](https://github.com/spring-projects/spring-security/blob/master/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/JwtBearerOAuth2AuthorizedClientProvider.java#L62). At line 62,  the check of token expiration has been omitted. What happens if token is expired? In my understanding, the Provider will still return `null`. Won't this cause the expired token to be reused, resulting in a 401 error?

jgrandja:
> What happens if token is expired?

Yes, the request will fail at the resource with 401 and the `ServletOAuth2AuthorizedClientExchangeFilterFunction.AuthorizationFailureForwarder` will handle the error and ultimately remove the `OAuth2AuthorizedClient` via the default `RemoveAuthorizedClientOAuth2AuthorizationFailureHandler`.

`RemoveAuthorizedClientOAuth2AuthorizationFailureHandler` is also used on authorization/re-authorization failures within `OAuth2AuthorizedClientManager`. Please review the [OAuth2AuthorizedClientManager](https://docs.spring.io/spring-security/site/docs/current/reference/html5/#oauth2Client-authorized-manager-provider) reference for further details.

H-LREB:
Thank you @jgrandja  for this information. 

I understand the importance of the declaring a `AuthorizationFailureForwarder`, as it is necessary to clear the `OAuth2AuthorizedClient` when it becomes invalid. 
However, in this case, the error is not exceptional or related to some unexpected behaviour. We know for sure that tokens will expire (e.g. every 5 minutes, or every hour ...etc.). This means that developers will have to handle those situations (which happen periodically).
Not only does this give more work / responsibility to development teams (e.g. by doing a retry of the business request using `WebClient`) but it also makes 401 errors « normal ». This makes it harder to recognize the 401 errors resulting from attack attempts.
In my humble opinion, it is much simpler and more efficient if Spring Security handles token expiration in `JwtBearerOAuth2AuthorizedClientProvider` (line 62), the same way it does with `PasswordOAuth2AuthorizedClientProvider.`

What do you think?

jgrandja:
@H-LREB This is now fixed via gh-9700

H-LREB:
That is a wonderful news.
Thank you!

Issue #9548, opened 2021-04-05
OAuth2 should not redirect to login for Accept: application/json requests


candrews:
**Describe the bug**
Requests that have the header `"X-Requested-With: XMLHttpRequest"` (correctly) return HTTP 401.

However, requests that have the `Accept` header set to any value other than those listed at https://github.com/spring-projects/spring-security/blob/5.4.5/config/src/main/java/org/springframework/security/config/annotation/web/configurers/AbstractAuthenticationFilterConfigurer.java#L256 should also result in an HTTP 401. Currently, however, they are responded to with a 302 redirect to the login page.

See also https://github.com/spring-projects/spring-security/issues/6812

**To Reproduce**
Here's a test that can be added to `OAuth2LoginConfigurerTests` demonstrating the problem:

```java
	@Test
	public void oauth2LoginWithOneClientConfiguredAndRequestAcceptJSONNotAuthenticatedThenDoesNotRedirectForAuthorization()
			throws Exception {
		loadConfig(OAuth2LoginConfig.class);
		String requestUri = "/";
		this.request = new MockHttpServletRequest("GET", requestUri);
		this.request.setServletPath(requestUri);
		this.request.addHeader("Accept", MediaType.APPLICATION_JSON);
		this.springSecurityFilterChain.doFilter(this.request, this.response, this.filterChain);
		assertThat(this.response.getStatus()).isEqualTo(401);
		assertThat(this.response.getRedirectedUrl()).doesNotMatch("http://localhost/oauth2/authorization/google");
	}
```

**Expected behavior**
Requests with the `Accept` header set to `application/json` (for example) should receive an HTTP 401 response.

**Sample**
I've provided a unit test so hopefully that's acceptable in place of a full sample project.

jgrandja:
@candrews See [comment](https://github.com/spring-projects/spring-security/pull/9549#issuecomment-814377416)

jgrandja:
@candrews Closing as per [comment](https://github.com/spring-projects/spring-security/pull/9549#pullrequestreview-630091067).

jgrandja:
[...]

Issue #9266, opened 2020-12-06
WebFlux security should not overwrite the default entry point with a delegating entry point


foo4u:
**Describe the bug**

When `ServerHttpSecurity.build()` creates a `SecurityWebFilterChain`, it replaces the default entry point with the last delegating entry point in `this.defaultEntryPoints`.

https://github.com/spring-projects/spring-security/blob/3cba4eccdcd5f17932ea9f365bd6df2684820819/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java#L1434-L1445

Specifically:

```java
result.setDefaultEntryPoint(this.defaultEntryPoints.get(this.defaultEntryPoints.size() - 1).getEntryPoint());
```

Since delegating entry points should be applied conditionally, this breaks the contract for certain default entry points, such as [`OAuth2LoginSpec.setDefaultEntryponits`](https://github.com/spring-projects/spring-security/blob/3cba4eccdcd5f17932ea9f365bd6df2684820819/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java#L3236), which are designed to be conditional. For example, the `OAuth2LoginSpec` is designed to not redirect on XHR requests but this behavior breaks that contract.

This cannot be worked around by configuring a new default entry point (e.g., `.exceptionHandling().authenticationEntryPoint(...)`) because doing so triggers another bug, which I'll open another issue about.

**To Reproduce**

The issue is reproducible most easily with OAuth2 login. However, the problem is not unique to this scenario. 

1. Configure Spring WebFlux security with OAuth2 login

```java
@Configuration
@EnableWebFluxSecurity
public class SecurityConfig
{
    @Bean
    public SecurityWebFilterChain securityWebFilterChain(ServerHttpSecurity http)
    {
        return http
                .authorizeExchange(exchanges -> exchanges.anyExchange().authenticated())
                .oauth2Login()
                .and()
                .build();
    }
}
```

https://github.com/foo4u/spring-security-bugs/blob/38ba4cd007eb3a7d75f5912f5a013b7ee4ac197d/src/main/java/com/example/demo/SecurityConfig.java#L9-L22

2. Make an XHR request to an endpoint requiring authentication with XHR headers expecting an HTTP 401.

```java
	@Test
	void respondsWithHttp401()
	{
		client
				.get()
				.accept(MediaType.APPLICATION_JSON)
				.header("X-Requested-With","XMLHttpRequest")
				.exchange()
[...]

PR #9549, opened 2021-04-05
OAuth2 should not login redirect non-user-browser requests

candrews:
Requests with an Accept indicating a request didn't originating as a
user browsing a URL shouldn't redirect to the login page. For example,
requests with `Accept: application/json` or `X-Requested-With:
XMLHttpRequest` shouldn't go to the login page; they should result an
HTTP 401 error.

Closes: gh-9548

This change brings oauth2 into alignment with the intent expressed in this commit:
https://github.com/spring-projects/spring-security/commit/66858e22ad381c83b47d373296a4cafbdb1a2615

### DIFF
--- a/config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2LoginConfigurerTests.java
+++ b/config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2LoginConfigurerTests.java
@@ -355,7 +355,9 @@ public class OAuth2LoginConfigurerTests {
 		String requestUri = "/";
 		this.request = new MockHttpServletRequest("GET", requestUri);
 		this.request.setServletPath(requestUri);
+		this.request.addHeader("Accept", MediaType.TEXT_HTML);
 		this.springSecurityFilterChain.doFilter(this.request, this.response, this.filterChain);
+		assertThat(this.response.getStatus()).isEqualTo(302);
 		assertThat(this.response.getRedirectedUrl()).matches("http://localhost/oauth2/authorization/google");
 	}
 
@@ -369,7 +371,8 @@ public class OAuth2LoginConfigurerTests {
 		this.request.setServletPath(requestUri);
 		this.request.addHeader(HttpHeaders.ACCEPT, new MediaType("image", "*").toString());
 		this.springSecurityFilterChain.doFilter(this.request, this.response, this.filterChain);
-		assertThat(this.response.getRedirectedUrl()).matches("http://localhost/login");
+		assertThat(this.response.getStatus()).isEqualTo(302);
+		assertThat(this.response.getRedirectedUrl()).matches("http://localhost/oauth2/authorization/google");
 	}
 
 	// gh-5347
@@ -393,7 +396,19 @@ public class OAuth2LoginConfigurerTests {
 		this.request.setServletPath(requestUri);
 		this.request.addHeader("X-Requested-With", "XMLHttpRequest");
 		this.springSecurityFilterChain.doFilter(this.request, this.response, this.filterChain);
-		assertThat(this.response.getRedirectedUrl()).doesNotMatch("http://localhost/oauth2/authorization/google");
+		assertThat(this.response.getStatus()).isEqualTo(401);
+	}
+
+	@Test
+	public void oauth2LoginWithOneClientConfiguredAndRequestAcceptJSONNotAuthenticatedThenDoesNotRedirectForAuthorization()
+			throws Exception {
+		loadConfig(OAuth2LoginConfig.class);
+		String requestUri = "/";
+		this.request = new MockHttpServletRequest("GET", requestUri);
+		this.request.setServletPath(requestUri);
+		this.request.addHeader("Accept", MediaType.APPLICATION_JSON);
+		this.springSecurityFilterChain.doFilter(this.request, this.response, this.filterChain);
+		assertThat(this.response.getStatus()).isEqualTo(401);
 	}
 
 	@Test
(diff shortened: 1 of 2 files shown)


candrews:
I don't see why requests for `favicon.ico` were treated differently, nor could I find out why from the git history.

rwinch:
@candrews Thanks for the pull request. Can we update the pull request to do a single thing rather than changing favicon.ico, application/json, etc. It's much easier for us to merge, backport, communicate to users what has changed, etc by doing this.

The favicon is treated specially because it is requested in the background by the browser which will trigger it to be stored as the saved request. That means after successful log in the server will redirect to the saved request which (depending on the race condition and if it is public or not) might be the favicon.

jgrandja:
@candrews The current behaviour is correct.

The [OAuth 2.0 Authorization Code grant](https://tools.ietf.org/html/rfc6749#section-4.1) is a **redirection-based flow**:

> The authorization code grant type is used to obtain both access
   tokens and refresh tokens and is optimized for confidential clients.
   Since **this is a redirection-based flow, the client must be capable of
   interacting with the resource owner's user-agent (typically a web
   browser) and capable of receiving incoming requests (via redirection)
   from the authorization server**.

`oauth2Login()` is implemented with the OAuth 2.0 Authorization Code grant flow. Therefore, the authentication flow should be handled by the browser agent **NOT the XHR client**.

Please see this [comment](https://github.com/spring-projects/spring-security/issues/9266#issuecomment-813400866) for further details.

Let me know if this makes sense?

candrews:
> `oauth2Login()` is implemented with the OAuth 2.0 Authorization Code grant flow. Therefore, the authentication flow should be handled by the browser agent **NOT the XHR client**.

That makes sense to me. It seems to be saying that if a request is made by the XHR client, then it cannot be handled by the OAuth 2.0 Authorization Code grant flow. As you said,
> Therefore, the authentication flow should be handled by the browser agent NOT the XHR client.

Which I believe means that this PR is correct: XHR requests should result in immediate denial because they cannot be handled by the OAuth 2.0 Authorization Code grant flow.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2LoginConfigurerTests.java
```diff
@@ -393,7 +396,19 @@ public void oauth2LoginWithOneClientConfiguredAndRequestXHRNotAuthenticatedThenD
...
+	@Test
+	public void oauth2LoginWithOneClientConfiguredAndRequestAcceptJSONNotAuthenticatedThenDoesNotRedirectForAuthorization()
+			throws Exception {
+		loadConfig(OAuth2LoginConfig.class);
+		String requestUri = "/";
+		this.request = new MockHttpServletRequest("GET", requestUri);
+		this.request.setServletPath(requestUri);
+		this.request.addHeader("Accept", MediaType.APPLICATION_JSON);
+		this.springSecurityFilterChain.doFilter(this.request, this.response, this.filterChain);
+		assertThat(this.response.getStatus()).isEqualTo(401);
```
jgrandja, line 411: This is not the expected behaviour if ONLY `oauth2Login()` is configured.

However, if `httpBasic()` is also configured then it would handle the unauthenticated request (via content negotiation) and return `401`.

jgrandja (commented):
@candrews `OAuth2LoginConfigurer` is implemented correctly. I think the confusion is related to how an application should be configured if it's expecting to handle XHR requests.

If only `oauth2Login()` is configured, then unauthenticated XHR requests will return a `302`. 

However, if both `oauth2Login()` and `httpBasic()` is configured then a `401` will be returned as the `AuthenticationEntryPoint` for `httpBasic()` will handle the request via content negotiation.

This behaviour is the same for `formLogin()` and `httpBasic()`.

I put together some tests that demonstrate the correct behaviour depending on the authentication configured in the application, e.g. `httpBasic()`, `formLogin()` and `oauth2Login()`. Please see this [commit](https://github.com/jgrandja/spring-security/commit/befd50a65540647115692d8b5882f5892dae3f9e#diff-1d22fe32e7c9b147d919b2d57d85b926d9251f4008b233706c0cc9cc2bdc0eacR42).

I'm closing this PR and associated issue as the implementation is correct.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2LoginConfigurerTests.java
candrews: Unfortunately, I'm still not understanding why `httpBasic()` and `formLogin()` behave differently from `oauth2Login()` with regards to content negotiation. Shouldn't they all behave the same?

jgrandja: > how does one setup `oauth2Login()` so it uses content negotiation and behaves like `formLogin()`

`oauth2Login()` DOES behave the same as `formLogin()`. Do you look at the tests I put together in the previous [comment](https://github.com/jgrandja/spring-security/commit/befd50a65540647115692d8b5882f5892dae3f9e#diff-1d22fe32e7c9b147d919b2d57d85b926d9251f4008b233706c0cc9cc2bdc0eacR42)?

See `formLogin_requestWhenXHRNotAuthenticatedThenRedirectToLogin` and `oauth2Login_requestWhenXHRNotAuthenticatedThenRedirectToLogin`. These tests demonstrate the same behaviour.

jgrandja: @candrews 

> how does one setup `oauth2Login()` so it uses content negotiation

See this [commit](https://github.com/jgrandja/spring-security/commit/93c82779219515520086fc5420fb72dd81eb64ca) as I believe this is the configuration you are looking for

PR #9559, opened 2021-04-06
Add null check in CsrfFilter and CsrfWebFilter

Exclamation-mark:
Solve the problem that CsrfFilter and CsrfWebFilter throws NPE exception when comparing two byte array is equal in low JDK version.
When JDK version is lower than 1.8.0_45,method java.security.MessageDigest#isEqual does not verify whether the two arrays are null.
And the above two class call this method without null judgment.

ZiQiang Zhao<1694392889@qq.com>
Because I'm not familiar with git, I mistakenly closed the old pr #9524 . @jzheaux


rwinch:
Duplicate of gh-9561

PR #9561, opened 2021-04-06
Add null check in CsrfFilter and CsrfWebFilter

Exclamation-mark:
Solve the problem that CsrfFilter and CsrfWebFilter throws NPE exception when comparing two byte array is equal in low JDK version.
When JDK version is lower than 1.8.0_45,method java.security.MessageDigest#isEqual does not verify whether the two arrays are null.
And the above two class call this method without null judgment.

ZiQiang Zhao  <1694392889@qq.com>
Because I'm not familiar with git, I mistakenly closed the old pr #9524 . @jzheaux

### DIFF
--- a/web/src/main/java/org/springframework/security/web/csrf/CsrfFilter.java
+++ b/web/src/main/java/org/springframework/security/web/csrf/CsrfFilter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2013 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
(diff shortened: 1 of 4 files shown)


jzheaux (changes requested):
Thanks, @Exclamation-mark! Are you able to add a test into `CsrfWebFilterTests` as well?

jzheaux:
Thanks for the PR, @Exclamation-mark! This is now merged via 87ed5270233284d051f1d71ed91c29e12f585337

Issue #9528, opened 2021-03-30
Javadoc error when linking to Spring Security's javadoc using Java 8


wilkinsona:
It looks like Spring Security's javadoc is being built with Java 11. This prevents a Java 8 build from linking to it with the following error:

```
javadoc: warning - Error fetching URL: https://docs.spring.io/spring-security/site/docs/5.5.0-M3/api/
```

The error isn't very clear, but it's due to a missing `package-list` file at https://docs.spring.io/spring-security/site/docs/5.5.0-M3/api/package-list. This is a regression from M2 where https://docs.spring.io/spring-security/site/docs/5.5.0-M2/api/package-list exists.

Until we raise the minimum supported version of Java, I think it would be preferable to use Java 8 to generate the project's javadoc to avoid this sort of problem. It would also be consistent with other Spring projects which, AFAIK, are all using Java 8 for javadoc generation at the moment.

If Java 11 has to be used, perhaps you could put a workaround in place that's similar to [the one that the Gradle team used](https://github.com/gradle/gradle/pull/8193)?

wilkinsona:
Thanks, @jzheaux. Did you consider switching to Java 8 so that Spring Security's javadoc is consistent with the other projects'?

PR #9562, opened 2021-04-06
Add package-list

jzheaux:
Closes gh-9528

### DIFF
--- a/buildSrc/src/main/groovy/io/spring/gradle/convention/JavadocApiPlugin.groovy
+++ b/buildSrc/src/main/groovy/io/spring/gradle/convention/JavadocApiPlugin.groovy
@@ -23,6 +23,7 @@ import java.util.Set;
 import java.util.regex.Pattern;
 
 import org.gradle.api.Action;
+import org.gradle.api.JavaVersion
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.plugins.JavaPluginConvention;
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/groovy/io/spring/gradle/convention/JavadocApiPlugin.groovy
```diff
@@ -14,23 +14,20 @@
...
+
+import org.gradle.api.Action
+import org.gradle.api.JavaVersion
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.api.plugins.JavaPluginConvention
+import org.gradle.api.tasks.SourceSet
+import org.gradle.api.tasks.javadoc.Javadoc
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
```
rwinch, line 30: It seems like these lines should probably be part of another commit?

buildSrc/src/main/groovy/io/spring/gradle/convention/JavadocApiPlugin.groovy
```diff
@@ -49,6 +46,13 @@ public class JavadocApiPlugin implements Plugin<Project> {
 
 		api.setGroup("Documentation");
 		api.setDescription("Generates aggregated Javadoc API documentation.");
+		api.doLast {
+			if (JavaVersion.current().isJava11Compatible()) {
+				File packageList = new File(api.destinationDir, 'package-list')
```
rwinch, line 51: It is better to use project.file to create references to files as it is more robust. See https://docs.gradle.org/current/userguide/working_with_files.html#sec:single_file_paths

> Never use new File(relative path) because this creates a path relative to the current working directory (CWD). Gradle can make no guarantees about the location of the CWD, which means builds that rely on it may break at any time.

Obviously this works if the path is absolute, but absolute paths have the problem of causing a cache miss if the build happens in one directory and then tries to run again in another directory. This is a common scenario when using Gradle enterprise cache (which we use) with a build happening on one machine that adds a cache entry and then someone building on another machine in a different directory.

rwinch: I think it would be preferable to use [Project.copy](https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#copy-org.gradle.api.Action-) to copy the files as it is build in.

rwinch (changes requested):

Issue #7155, opened 2019-07-26
Remove internal Optional usage in favor of null checks


rwinch:
### Summary

While some prefer the readability of using the Optional API, the GC overhead of creating additional objects (including intermediate objects) can cause significant decrease in performance There are some simple [benchmarks](https://gist.github.com/mp911de/7f203d787c6fdaf97dc91e6dc3aa73d3) that illustrate the problem.

We should replace Stream usage with for loops throughout Spring Security's code base.

Related gh-7154

eddumelendez:
Hi @rwinch, do you think this is good for `first-timers-only`?

rwinch:
Thanks for asking @eddumelendez I think it is likely a lot of code to touch for a first timers.

PR #9632, opened 2021-04-13
Add RequestedUrlRedirectInvalidSessionStrategy implemention of InvalidSessionStrategy

candrews:
Performs a redirect to the original request URL when an invalid requested session is detected.

In effect, when a user's session times out, the user is redirected to URL they originally requested instead of some fixed URL.

I have a number of projects using this implementation, so I suspect it's useful to the wide Spring Security world.

I'm hoping you can give a quick review of this idea. If it seems like it will be acceptable to Spring Security, I'll update this PR with tests.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/session/RequestedUrlRedirectInvalidSessionStrategy.java
+++ b/web/src/main/java/org/springframework/security/web/session/RequestedUrlRedirectInvalidSessionStrategy.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2002-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.web.session;
+
+import java.io.IOException;
+import java.util.Optional;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import org.springframework.security.web.DefaultRedirectStrategy;
+import org.springframework.security.web.RedirectStrategy;
+import org.springframework.web.util.UrlPathHelper;
+
+/**
+ * Performs a redirect to the original request URL when an invalid requested session is
+ * detected by the {@code SessionManagementFilter}.
+ *
+ * @author Craig Andrews
+ */
+public final class RequestedUrlRedirectInvalidSessionStrategy implements InvalidSessionStrategy {
+
+	private final Log logger = LogFactory.getLog(getClass());
+
+	private final RedirectStrategy redirectStrategy = new DefaultRedirectStrategy();
+
+	private boolean createNewSession = true;
+
+	private final UrlPathHelper urlPathHelper = new UrlPathHelper();
+
+	@Override
+	public void onInvalidSessionDetected(HttpServletRequest request, HttpServletResponse response) throws IOException {
+		String destinationUrl = this.urlPathHelper.getOriginatingRequestUri(request)
+				+ Optional.ofNullable(this.urlPathHelper.getOriginatingQueryString(request)).filter((s) -> !s.isEmpty())
+						.map((s) -> "?" + s).orElse("");
+		this.logger.debug("Starting new session (if required) and redirecting to '" + destinationUrl + "'");
+		if (this.createNewSession) {
+			request.getSession();
+		}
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/springframework/security/web/session/RequestedUrlRedirectInvalidSessionStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+	private final RedirectStrategy redirectStrategy = new DefaultRedirectStrategy();
+
+	private boolean createNewSession = true;
+
+	private final UrlPathHelper urlPathHelper = new UrlPathHelper();
+
+	@Override
+	public void onInvalidSessionDetected(HttpServletRequest request, HttpServletResponse response) throws IOException {
+		String destinationUrl = this.urlPathHelper.getOriginatingRequestUri(request)
+				+ Optional.ofNullable(this.urlPathHelper.getOriginatingQueryString(request)).filter((s) -> !s.isEmpty())
```
rwinch, line 51: We avoid the use of Optional See gh-7155

web/src/main/java/org/springframework/security/web/session/RequestedUrlRedirectInvalidSessionStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+
+	private final RedirectStrategy redirectStrategy = new DefaultRedirectStrategy();
+
+	private boolean createNewSession = true;
+
+	private final UrlPathHelper urlPathHelper = new UrlPathHelper();
+
+	@Override
+	public void onInvalidSessionDetected(HttpServletRequest request, HttpServletResponse response) throws IOException {
+		String destinationUrl = this.urlPathHelper.getOriginatingRequestUri(request)
```
rwinch, line 50: Rather than `UrlPathHelper` which requires multiple operations and string concatenation, did consider `ServletUriComponentsBuilder` and then clearing out the scheme, host, and port.

web/src/main/java/org/springframework/security/web/session/RequestedUrlRedirectInvalidSessionStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+	 * possible looping issues where the same session ID is sent with the redirected
+	 * request). Alternatively, ensure that the configured URL does not pass through the
+	 * {@code SessionManagementFilter}.
+	 * @param createNewSession defaults to {@code true}.
+	 */
+	public void setCreateNewSession(boolean createNewSession) {
+		this.createNewSession = createNewSession;
+	}
+
+}
```
rwinch, line 71: As you mentioned, we will also want tests.

rwinch (changes requested):
Thanks for the PR! Overall, I understand and agree with the proposal. I've provided feedback inline.

web/src/main/java/org/springframework/security/web/session/RequestedUrlRedirectInvalidSessionStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+	private boolean createNewSession = true;
+
+	private final UrlPathHelper urlPathHelper = new UrlPathHelper();
+
+	@Override
+	public void onInvalidSessionDetected(HttpServletRequest request, HttpServletResponse response) throws IOException {
+		String destinationUrl = this.urlPathHelper.getOriginatingRequestUri(request)
+				+ Optional.ofNullable(this.urlPathHelper.getOriginatingQueryString(request)).filter((s) -> !s.isEmpty())
+						.map((s) -> "?" + s).orElse("");
+		this.logger.debug("Starting new session (if required) and redirecting to '" + destinationUrl + "'");
```
rwinch, line 53: Please avoid string concatenation for logging without using guards.

candrews: I noticed that there was no guard in `SimpleRedirectInvalidSessionStrategy` so I submitted a PR addressing that: https://github.com/spring-projects/spring-security/pull/9648

I've also added the guard here and updated the PR accordingly

web/src/main/java/org/springframework/security/web/session/RequestedUrlRedirectInvalidSessionStrategy.java
candrews: I've now added a test and updated the PR accordingly

web/src/main/java/org/springframework/security/web/session/RequestedUrlRedirectInvalidSessionStrategy.java
candrews: That sounds like a good idea. I've made that change and updated the PR accordingly.

web/src/main/java/org/springframework/security/web/session/RequestedUrlRedirectInvalidSessionStrategy.java
candrews: That's good to know, thank you for linking to that issue with the explanation!

I've updated the PR accordingly.

candrews:
Thank you for the review and feedback - I've addressed all of the feedback provided, so I think this PR is ready for another review.

rwinch (approved):
This looks good. I've scheduled it for 5.6.0-M1

rwinch:

Issue #9653, opened 2021-04-18
Dynamic post-logout-redirect-url in OidcClientInitiatedLogoutSuccessHandler


sebsprenger:
**Expected Behavior**

Provide a way to provide a redirect URL which depends on the request after OIDC logout.

**Current Behavior**

`OidcClientInitiatedLogoutSuccessHandler.setPostLogoutRedirectUri(String postLogoutRedirectUri)` can be used to statically set the redirect URL for all logout requests. It lacks a way to create the redirect URL dynamically depending on the `HttpServletRequest`.

**Context**

In my scenario I want to carry over query parameters to the redirect URL.
A user may be on `/app1/page1` and wants to logout, so passes a query param to the `/logout?clientId=47` endpoint (which is provided by Spring security) which then shall redirect afterwards to `/app1/`. Accordingly `/logout?clientId=11` may lead to `/app2/`.

My initial WebSecurityConfig looked like the following
```
var handler = new OidcClientInitiatedLogoutSuccessHandler(oauthClientRepo);
handler.setPostLogoutRedirectUri("/app1/")

http.logout()
  .logoutUrl("/logout")
  .logoutSuccessHandler(handler)
```

This config always redirects to `/app1/` regardless of the `clientId` query parameter.

As a workaround a wrote my own `LogoutSuccessHandler` which is basically a copy of `OidcClientInitiatedLogoutSuccessHandler` and changed only the part which builds the redirect URL to consider the query parameter that I envisioned. I dislike the code duplication as I want to benefit from updates  made by others. The change that I did is very focussed (only the creation of the URL) and can be delegated to a plugin.

jgrandja:
Thanks for the report @sebsprenger. I left a [comment](https://github.com/spring-projects/spring-security/pull/9654/files#r615944105) and @jzheaux will be able to help with this further.

jzheaux:
@sebsprenger, thanks for the suggestion and the related PR.

A change may be reasonable, but I'd first like to understand your description better:

> In my scenario I want to carry over query parameters to the redirect URL.
A user may be on `/app1/page1` and wants to logout, so passes a query param to the `/logout?clientId=47` endpoint (which is provided by Spring security) which then shall redirect afterwards to `/app1/`. Accordingly `/logout?clientId=11` may lead to `/app2/`.

[...]

PR #9654, opened 2021-04-18
Allow plugins to adjust the postLogoutRedirectUri

sebsprenger:
OidcClientInitiatedLogoutSuccessHandler only allowed a static redirect URI
during OIDC logout. Now there is an interface which allows more flexibility
as it has access to the request object.

Closes gh-9653

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/web/logout/PostLogoutRedirectUriProvider.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/web/logout/PostLogoutRedirectUriProvider.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2002-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/web/logout/PostLogoutRedirectUriProvider.java
```diff
@@ -0,0 +1,34 @@
...
+import javax.servlet.http.HttpServletRequest;
+
+/**
+ * Strategy to create a post logout redirect URI during the OIDC logout flow.
+ *
+ * @author Sebastian Sprenger
+ * @since 5.4
+ * @see org.springframework.security.oauth2.client.oidc.web.logout.OidcClientInitiatedLogoutSuccessHandler
+ **/
+public interface PostLogoutRedirectUriProvider {
```
jgrandja, line 30: I don't think we should introduce a new API for this. We may consider making `OidcClientInitiatedLogoutSuccessHandler.postLogoutRedirectUri()` `protected` OR there may be a way to leverage `URI` template variables adding to the existing `baseUrl`

jzheaux:
Given https://github.com/spring-projects/spring-security/issues/9653#issuecomment-992183173 this PR seems unnecessary for now, so I'll close it. If the need arises again, I'm happy to take another look.

Issue #9651, opened 2021-04-17
JwtDecoders from issuer uri produces ClassCastException


tiborkoch:
**Describe the bug**
Using the latest Spring Boot snapshot version (2.5.0-SNAPSHOT) that has Spring Security 5.5.0-RC1 as dependency, a `ClassCastException` is thrown when creating a `JwtDecoder` from an issuer uri.

```
Caused by: java.lang.ClassCastException: class com.nimbusds.jose.Algorithm cannot be cast to class com.nimbusds.jose.JWSAlgorithm (com.nimbusds.jose.Algorithm and com.nimbusds.jose.JWSAlgorithm are in unnamed module of loader 'app')
	at org.springframework.security.oauth2.jwt.JwtDecoderProviderConfigurationUtils.getSignatureAlgorithms(JwtDecoderProviderConfigurationUtils.java:93) ~[spring-security-oauth2-jose-5.5.0-RC1.jar:5.5.0-RC1]
```

It can be fixed by using 
```
JWSAlgorithm jwsAlgorithm = JWSAlgorithm.parse(jwk.getAlgorithm().getName());
jwsAlgorithms.add(jwsAlgorithm);
```
instead of casting.

**To Reproduce**
Start the app in the provided repository.

**Expected behavior**
No errors on startup.

**Sample**
https://github.com/tiborkoch/spring-security-oauth2resourceserver

UbaidurRehman1:
I was testing latest Spring Boot snapshot version (2.5.0-SNAPSHOT)  and I got the same issue. Folllowing is the stacktrace:
Spring Boot Version: 2.5.0-SNAPSHOT
Spring Security Version: 5.5.0-RC1

```
Caused by: java.lang.ClassCastException: class com.nimbusds.jose.Algorithm cannot be cast to class com.nimbusds.jose.JWSAlgorithm (com.nimbusds.jose.Algorithm and com.nimbusds.jose.JWSAlgorithm are in unnamed module of loader 'app')
	at org.springframework.security.oauth2.jwt.JwtDecoderProviderConfigurationUtils.getSignatureAlgorithms(JwtDecoderProviderConfigurationUtils.java:93) ~[spring-security-oauth2-jose-5.5.0-RC1.jar:5.5.0-RC1]
[...]

PR #9658, opened 2021-04-20
Fix ClassCastException in JwtDecoders fromIssuerLocation

tiborkoch:
Fix for https://github.com/spring-projects/spring-security/issues/9651

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoderProviderConfigurationUtils.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoderProviderConfigurationUtils.java
@@ -90,7 +90,8 @@ final class JwtDecoderProviderConfigurationUtils {
 			List<? extends JWK> jwks = jwkSource.get(new JWKSelector(jwkMatcher), null);
 			for (JWK jwk : jwks) {
 				if (jwk.getAlgorithm() != null) {
-					jwsAlgorithms.add((JWSAlgorithm) jwk.getAlgorithm());
+					JWSAlgorithm jwsAlgorithm = JWSAlgorithm.parse(jwk.getAlgorithm().getName());
+					jwsAlgorithms.add(jwsAlgorithm);
 				}
 				else {
 					if (jwk.getKeyType() == KeyType.RSA) {
--- a/oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/JwtDecoderProviderConfigurationUtilsTests.java
+++ b/oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/JwtDecoderProviderConfigurationUtilsTests.java
@@ -20,6 +20,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Set;
 
+import com.nimbusds.jose.Algorithm;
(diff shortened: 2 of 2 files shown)


oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoderProviderConfigurationUtils.java
```diff
@@ -90,7 +90,8 @@ static void validateIssuer(Map<String, Object> configuration, String issuer) {
 			List<? extends JWK> jwks = jwkSource.get(new JWKSelector(jwkMatcher), null);
 			for (JWK jwk : jwks) {
 				if (jwk.getAlgorithm() != null) {
-					jwsAlgorithms.add((JWSAlgorithm) jwk.getAlgorithm());
+					JWSAlgorithm jwsAlgorithm = JWSAlgorithm.parse(jwk.getAlgorithm().getName());
```
jzheaux, line 93: Will you please add a test that confirms the fix? It should fail without your change and pass with your change.

Also in that test, would you please prefix it with the issue number like so:

```java
// gh-9651
@Test
public void getSignatureAlgorithmsWhenAlgorithmThenParses() {
 // ...
}
```

jzheaux (changes requested):
Thanks for the PR, @tiborkoch! Please see my review comments inline.

In addition, would you please include the issue number in your squashed Git commit? Something like this:

```
Fix ClassCastException

Closes gh-9651
```

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoderProviderConfigurationUtils.java
tiborkoch: added a test and updated the commit message

jzheaux:
Thanks, @tiborkoch! This is now merged into `master`.

Blockost:
I'm having the exact same issue but with `spring-security-oauth2-jose@5.6.1`. I set `spring.security.oauth2.resourceserver.jwt.jwk-set-uri` to my Auth0 endpoint but without success.

The line that raises the error is the one at 411 of class  `NimbusReactiveJwtDecoder`:
```java
private JWKSelector createSelector(Function<JWSAlgorithm, Boolean> expectedJwsAlgorithms, Header header) {
	JWSHeader jwsHeader = (JWSHeader) header; // <--- Error thrown here !
	if (!expectedJwsAlgorithms.apply(jwsHeader.getAlgorithm())) {
		throw new BadJwtException("Unsupported algorithm of " + header.getAlgorithm());
	}
	return new JWKSelector(JWKMatcher.forJWSHeader(jwsHeader));
}
```

I'm using:
- spring boot and `spring-boot-starter-oauth2-resource-server ` 2.6.3
- It's using `ReactiveJwtDecoder` because I'm building an app with Spring Webflux

Not sure what I can do to fix this... 🤔

dhaeb:
@Blockost how did you deal with this issue? We created our own `NimbusReactiveJwtDecoder` but are not happy to maintain this code.
It seems like that: 
`NimbusJwtDecoder.withJwkSetUri`can handle `JWE`, the reactive version `NimbusReactiveJwtDecoder.withJwkSetUri`  does not work for us...

PR #9659, opened 2021-04-20
Align response processing in DefaultOAuth2UserService to other implementations

### LINKED ISSUE #9629: Improve customization of `DefaultOAuth2UserService` to handle other content types
**Expected Behavior**

`DefaultOAuth2UserService` can be extended to e.g. allow for custom body parsing to handle `application/jwt`for signed and/or encrypted UserInfo Response.

Rough draft: 
```java

public class CustomOAuth2UserService extends DefaultOAuth2UserService {

   @Override
    protected ResponseEntity<Map<String, Object>> getResponse(OAuth2UserRequest userRequest, RequestEntity<?> request) {
      // Custom code to handle requests that aren't simple application/json
      return ...;
     }
}
``` 

We are open for other solutions as well and happy to contribute, if that's something you see worth it as addition to spring-security.

**Current Behavior**

`DefaultOAuth2UserService` has to be copied and "rewritten" - because `getResponse()` is called inside `loadUser(OAuth2UserRequest userRequest)` which forces us to re-create the whole `loadUser(OAuth2UserRequest userRequest)` method. 

https://github.com/spring-projects/spring-security/blob/a325216f19277d4191c97afee1c66f82d056f9dc/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java#L88-L117

**Context**

Related to #9583

knoobie:
fixes #9629


pivotal-issuemaster:
@knoobie Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=9659)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=9659) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@knoobie Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=9659)!

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
```diff
@@ -105,20 +110,20 @@ public OAuth2User loadUser(OAuth2UserRequest userRequest) throws OAuth2Authentic
...
 		OAuth2AccessToken token = userRequest.getAccessToken();
 		for (String authority : token.getScopes()) {
 			authorities.add(new SimpleGrantedAuthority("SCOPE_" + authority));
 		}
-		return new DefaultOAuth2User(authorities, userAttributes, userNameAttributeName);
+		return new DefaultOAuth2User(authorities, userAuthority.getAttributes(), userNameAttributeName);
 	}
 
-	private ResponseEntity<Map<String, Object>> getResponse(OAuth2UserRequest userRequest, RequestEntity<?> request) {
+	private ResponseEntity<OAuth2UserAuthority> getResponse(OAuth2UserRequest userRequest, RequestEntity<?> request) {
```
jgrandja, line 124: Why does this return `OAuth2UserAuthority` instead of `Map<String, Object>`? The UserInfo response returns a set of claim/attribute name/value pairs.

knoobie: If I remember correctly, this was done to align the created `OAuth2UserAuthorityHttpMessageConverter` with e.g. `OAuth2AccessTokenResponseHttpMessageConverter` in the same package. Additionally, I wasn't sure how to satisfy `supports(Class<?> clazz)` - so I ended up with this construct. I don't mind changing this to old value, as it's just a implementation detail. 

Before my change, the following was done with the returned `Map<String, Object> userAttributes`:

`authorities.add(new OAuth2UserAuthority(userAttributes));` - so it felt like a good abstraction, if I move it to the private method. 

Any preference?

jgrandja: See [5.3.2. Successful UserInfo Response](https://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse):

> The UserInfo Claims MUST be returned as the members of a JSON object ...

`OAuth2UserAuthority` would not be the equivalent representation. We should use `OidcUserInfo`.

We can create a new `HttpMessageConverter`:

`public class OidcUserInfoHttpMessageConverter extends AbstractHttpMessageConverter<OidcUserInfo>`

Then we can configure it as follows:

```java
RestTemplate restTemplate = new RestTemplate(
		Collections.singletonList(
				new OidcUserInfoHttpMessageConverter()));
restTemplate.setErrorHandler(new OAuth2ErrorResponseErrorHandler());

DefaultOAuth2UserService userService = new DefaultOAuth2UserService();
userService.setRestOperations(restTemplate);

OidcUserService oidcUserService = new OidcUserService();
oidcUserService.setOauth2UserService(userService);
```

Makes sense?

knoobie: Thanks for your feedback! I haven't compared the spec with the used class - thanks for the input! I try to change it in the upcoming days.

knoobie: @jgrandja Adopted your suggestions and removed my new constructor for `DefaultOAuth2UserService`, because it can be easily done with your code shown above to customize it now.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
```diff
@@ -71,15 +74,14 @@
 
 	private static final String INVALID_USER_INFO_RESPONSE_ERROR_CODE = "invalid_user_info_response";
 
-	private static final ParameterizedTypeReference<Map<String, Object>> PARAMETERIZED_RESPONSE_TYPE = new ParameterizedTypeReference<Map<String, Object>>() {
-	};
-
 	private Converter<OAuth2UserRequest, RequestEntity<?>> requestEntityConverter = new OAuth2UserRequestEntityConverter();
```
knoobie, line 77: Should I improve `OAuth2UserRequestEntityConverter` as well, to allow for easier customizing by providing a setter for the Accept Header? 

Currently it is hardcoded to `APPLICATION_JSON` as well. 

This would allow the full jwt support by just implementing a custom `OidcUserInfoHttpMessageConverter` (application developer responsibility) and simple using `OAuth2UserRequestEntityConverter::setAcceptHeader("application/jwt")` instead of copying the source / implementing a custom solution.

jgrandja: > Should I improve `OAuth2UserRequestEntityConverter` as well, to allow for easier customizing by providing a setter for the Accept Header?

Let's leave this for now. We might consider updating it but this should go into a separate issue/commit.

PR #9671, opened 2021-04-22
Fix issue#9661 Buffer LEAK detected by ResourceLeakDetector in Authen…

kevinat:
Fix issue#9661 Buffer LEAK detected by ResourceLeakDetector in AuthenticationPayloadExchangeConverter

### DIFF
--- a/rsocket/src/main/java/org/springframework/security/rsocket/authentication/AuthenticationPayloadExchangeConverter.java
+++ b/rsocket/src/main/java/org/springframework/security/rsocket/authentication/AuthenticationPayloadExchangeConverter.java
@@ -71,16 +71,21 @@ public class AuthenticationPayloadExchangeConverter implements PayloadExchangeAu
 		if (authenticationMetadata == null) {
 			return null;
 		}
-		ByteBuf rawAuthentication = ByteBufAllocator.DEFAULT.buffer().writeBytes(authenticationMetadata);
-		if (!AuthMetadataCodec.isWellKnownAuthType(rawAuthentication)) {
-			return null;
-		}
-		WellKnownAuthType wellKnownAuthType = AuthMetadataCodec.readWellKnownAuthType(rawAuthentication);
-		if (WellKnownAuthType.SIMPLE.equals(wellKnownAuthType)) {
-			return simple(rawAuthentication);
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@kevinat Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=9671)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-security&pullRequestId=9671) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@kevinat Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-security&pullRequestId=9671)!

jzheaux:
Thanks for the PR, @kevinat!

Is there a way this can be tested? One way that comes to mind would be to rework how `AuthenticationPayloadExchangeConverter` internals use `ByteBuf` so that it can be invoked from a test.

For example, you could do:

```java
Authentication authentication(ByteBuf rawAuthentication) {
    try {
        // ...
    } finally {
        rawAuthentication.release();
    }
}
```

And you could add a test that invokes that method and confirms that the ref count is equal to 0.

kevinat:
@jzheaux This might not be the best way to test, cos that rework could break the logic of the  method `Authentication authentication(Map<String, Object> metadata)`, it basically takes a byte array from metadata, returns an Authentication, that is good enough. Break this into two method would make the code looks not that good.

I can reproduce the LEAK through this:

```java
for (int i = 0; i < 100000; i++) { 
    authenticationPayloadExchangeConverter.authentication(metadata);
}
System.gc();
```

but I don't known how to make the assertion, because the leak report is just an error log not an exception.

Any suggestion for this @jzheaux? and is it OK to make unit test like this?

jzheaux:
Thanks for the additional feedback about testing, @kevinat. Calling `System.gc()` is not guaranteed to perform a GC, so let's leave the test out.

Thanks again for the PR! I've merged this in commit 32ac31c101807b3d46a9ba5c875ccbf61998aa5f and will backport it to previous releases.

Issue #9698, opened 2021-04-30
When ReactiveJwtAuthenticationConverter bean defined, use it in OAuth2 Resource Server config automatically


ThomasVitale:
**Expected Behavior**

When I define a bean of type `ReactiveJwtAuthenticationConverter`, I would expect it to be considered in the OAuth2 Resource Server configuration automatically like it happens with its imperative counterpart (`JwtAuthenticationConverter`).

**Current Behavior**

When I define a bean of type `ReactiveJwtAuthenticationConverter`, it's not considered in the OAuth2 Resource Server configuration. Instead, it requires explicit definition through the DSL as follows.

```
.oauth2ResourceServer(oauth2 -> oauth2.jwt(jwt ->
						jwt.jwtAuthenticationConverter(jwtAuthenticationConverter())))
```

**Context**

It would be nice to align the behaviour between reactive and non-reactive applications since it could generate some confusion.

Proposed solution: https://github.com/spring-projects/spring-security/pull/9699

PR #9699, opened 2021-04-30
Load ReactiveJwtAuthenticationConverter bean in OAuth2 Resource Server config

ThomasVitale:
When a bean of type ReactiveJwtAuthenticationConverter is defined,
the OAuth2 Resource Server configuration will use it automatically
when no other converter is defined through the DSL.

Closes gh-9698

### DIFF
--- a/config/src/test/java/org/springframework/security/config/web/server/OAuth2ResourceServerSpecTests.java
+++ b/config/src/test/java/org/springframework/security/config/web/server/OAuth2ResourceServerSpecTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -64,6 +64,7 @@ import org.springframework.security.oauth2.jwt.ReactiveJwtDecoder;
 import org.springframework.security.oauth2.jwt.TestJwts;
 import org.springframework.security.oauth2.server.resource.BearerTokenAuthenticationToken;
 import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationConverter;
+import org.springframework.security.oauth2.server.resource.authentication.ReactiveJwtAuthenticationConverter;
 import org.springframework.security.oauth2.server.resource.authentication.ReactiveJwtAuthenticationConverterAdapter;
 import org.springframework.security.web.server.SecurityWebFilterChain;
 import org.springframework.security.web.server.authentication.HttpStatusServerEntryPoint;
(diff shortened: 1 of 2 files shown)


jzheaux:
Thanks for the PR, @ThomasVitale! I've scheduled this for the 5.6 release.

ThomasVitale:
@jzheaux Nice! Thanks for checking the PR.

config/src/test/java/org/springframework/security/config/web/server/OAuth2ResourceServerSpecTests.java
```diff
@@ -478,6 +479,59 @@ public void getJwtDecoderWhenNoBeansAndNoDslWiredThenWiringException() {
...
+		assertThat(jwt.getJwtAuthenticationConverter()).isNotEqualTo(beanWiredJwtAuthenticationConverter);
+	}
+
+	@Test
+	public void getJwtAuthenticationConverterWhenNoBeansAndNoDslWiredThenDefaultConverter() {
+		GenericWebApplicationContext context = autowireWebServerGenericWebApplicationContext();
+		ServerHttpSecurity http = new ServerHttpSecurity();
+		http.setApplicationContext(context);
+		ServerHttpSecurity.OAuth2ResourceServerSpec.JwtSpec jwt = http.oauth2ResourceServer().jwt();
+		assertThat(jwt.getJwtAuthenticationConverter()).isNotNull();
```
jzheaux, line 532: Should this instead test that it's an instanceof `ReactiveJwtAuthenticationConverter`?

config/src/test/java/org/springframework/security/config/web/server/OAuth2ResourceServerSpecTests.java
```diff
@@ -478,6 +479,59 @@ public void getJwtDecoderWhenNoBeansAndNoDslWiredThenWiringException() {
...
+				() -> beanWiredJwtAuthenticationConverter);
+		context.registerBean("secondJwtAuthenticationConverter", ReactiveJwtAuthenticationConverter.class,
+				() -> beanWiredJwtAuthenticationConverter);
+		ServerHttpSecurity.OAuth2ResourceServerSpec.JwtSpec jwt = http.oauth2ResourceServer().jwt();
+		jwt.jwtAuthenticationConverter(dslWiredJwtAuthenticationConverter);
+		assertThat(jwt.getJwtAuthenticationConverter()).isEqualTo(dslWiredJwtAuthenticationConverter);
+	}
+
+	@Test
+	public void getJwtAuthenticationConverterWhenTwoBeansWiredThenDefaultTakesPrecedence() {
```
jzheaux, line 512: I'd expect this test to align with `getJwtDecoderWhenTwoBeansWiredThenThrowsWiringException` and `OAuth2ResourceServerConfigurerTests#getJwtAuthenticationConverterWhenDuplicateConverterBeansThenThrowsException`. Is there a reason that this diverges from those?

jzheaux (changes requested):
Thanks, @ThomasVitale! Please see my inline feedback.

jzheaux:
Hi, @ThomasVitale! Are you able to apply the requested changes?

ThomasVitale:
@jzheaux sorry for the late reply. I can deliver the requested changes by the end of this week. Would that be ok?

jzheaux:
Thanks for checking in, @ThomasVitale. Yes, that works great.

config/src/test/java/org/springframework/security/config/web/server/OAuth2ResourceServerSpecTests.java
ThomasVitale: That's an interesting question. The reason is that `ReactiveJwtAuthenticationConverter` is resolved through the `getBeanOrNull()` method defined in `ServerHttpSecurity`. If more than one bean is defined, the logic will use the default one and not throw an error. I based the implementation on how most of the beans for OAuth2 and OIDC are configured in that class (e.g. `OidcUserService` or `ReactiveClientRegistrationRepository`). You made me notice that, in the imperative version, we throw an error if there's more than one bean defined. I guess we want the same logic for the reactive version as well. Thoughts?

jzheaux: Yes, that was my thinking. 

If an application declares more than one `ReactiveJwtAuthenticationConverter` bean, the application should clearly indicate which one should be used. An error is a simple way to do that.

config/src/test/java/org/springframework/security/config/web/server/OAuth2ResourceServerSpecTests.java
ThomasVitale: Good point. Changed to check for the instance type.

config/src/test/java/org/springframework/security/config/web/server/OAuth2ResourceServerSpecTests.java
ThomasVitale: Perfect! I changed it to the agreed behaviour.

jzheaux (approved):

Issue #9701, opened 2021-04-30
Prevent CI from running on forks


vpavic:
The project's GitHub Actions CI build has a hard dependency on some secrets (e.g. `GRADLE_ENTERPRISE_CACHE_USERNAME`), which causes the build to fail on the forks, causing unnecessary noise.

I've been greeted by one such build failure after syncing my fork today:
https://github.com/vpavic/spring-security/runs/2476329503?check_suite_focus=true

jzheaux:
I see that in `pr-build-workflow` there is the following declaration:

```
if: github.repository == 'spring-projects/spring-security'
```

I wonder if the same pattern can be applied to the jobs in `continuous-integration-workflow`.

Also, note that forks can disable actions via `Settings > Actions`.

PR #9717, opened 2021-05-03
Stop CI Jobs on Forks

jzheaux:
Closes gh-9701

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -17,25 +17,33 @@ env:
   COMMIT_SHA: ${{ github.sha }}
   ARTIFACTORY_USERNAME: ${{ secrets.ARTIFACTORY_USERNAME }}
   ARTIFACTORY_PASSWORD: ${{ secrets.ARTIFACTORY_PASSWORD }}
+  RUN_JOBS: ${{ github.repository == 'spring-projects/spring-security' }}
 
 jobs:
-  initiate_error_tracking:
-    name: Initiate job-level error tracking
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -19,9 +19,10 @@ env:
   ARTIFACTORY_PASSWORD: ${{ secrets.ARTIFACTORY_PASSWORD }}
 
 jobs:
-  initiate_error_tracking:
-    name: Initiate job-level error tracking
+  parent:
```
rwinch, line 22: Can you make this more descriptive? Perhaps, `prerequisites`?

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -19,9 +19,10 @@ env:
...
+    name: Pre-requisites for building
     runs-on: ubuntu-latest
+    if: github.repository == 'spring-projects/spring-security'
```
rwinch, line 25: Is there a way for this to be done through the environment so users can activate checks in their fork without changing the source?

rwinch (changes requested):
Thanks for the PR! I've left feedback inline

.github/workflows/continuous-integration-workflow.yml
jzheaux: Hypothetically, `prerequisites` could emit some kind of job outcome that the other tasks could check for. For example, they could possibly check for the existence of the `job-initiate-error-tracking.txt` file.

rwinch (approved):

Issue #9353, opened 2021-01-19
NimbusOpaqueTokenIntrospector uses deprecated MediaType.APPLICATION_JSON_UTF8


pedrodovale:
NimbusOpaqueTokenIntrospector.requestHeaders defines the header _MediaType.APPLICATION_JSON_UTF8_ which seems to be deprecated. In this bug report there's some details about it: https://github.com/spring-projects/spring-framework/issues/22788

The IDP I'm currently using actually rejects the request to the token endpoint because of this, returning a message stating that the response mode is not supported.

It is possible to workaround this issue by setting a new converter using _setRequestEntityConverter_, with a converter that sets the _MediaType.APPLICATION_JSON_ instead of the _MediaType.APPLICATION_JSON_UTF8_. However, shouldn't application/json be the configured MediaType by default?

Thank you in advance.

jzheaux:
Agreed, @pedrodovale, that this would be a good change. Are you able to provide a PR?

PR #9725, opened 2021-05-06
 Update Deprecated Property in Opaque Token Introspectors 

marcusdacoregio:
Update NimbusOpaqueTokenIntrospector and NimbusReactiveOpaqueTokenIntrospector to use MediaType.APPLICATION_JSON instead of the deprecated MediaType.APPLICATION_JSON_UTF8

Closes gh-9353

### DIFF
--- a/oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/introspection/NimbusReactiveOpaqueTokenIntrospectorTests.java
+++ b/oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/introspection/NimbusReactiveOpaqueTokenIntrospectorTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/introspection/NimbusReactiveOpaqueTokenIntrospectorTests.java
```diff
@@ -244,7 +244,7 @@ private WebClient mockResponse(String response) {
 		given(clientResponse.statusCode()).willReturn(HttpStatus.OK);
 		given(clientResponse.bodyToMono(String.class)).willReturn(Mono.just(response));
 		ClientResponse.Headers headers = mock(ClientResponse.Headers.class);
-		given(headers.contentType()).willReturn(Optional.of(MediaType.APPLICATION_JSON_UTF8));
+		given(headers.contentType()).willReturn(Optional.of(MediaType.APPLICATION_JSON));
```
jzheaux, line 247: Should there be a similar test on the servlet side?

jzheaux (commented):
Thanks for the PR, @marcusdacoregio! I've left some feedback inline.

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/introspection/NimbusReactiveOpaqueTokenIntrospectorTests.java
marcusdacoregio: The tests for NimbusOpaqueTokenIntrospector already use the `MediaType.APPLICATION_JSON` instead of the deprecated one so I guess there is no need to change them.

jzheaux:

Issue #9565, opened 2021-04-07
Propagate TestSecurityContextHolder to SecurityContextHolder after MockMvc calls


sniffertine:
**Expected Behavior**

After calling MockMvc, I want to be able to make assertions that depend on the SecurityContext defined by the test.

**Current Behavior**

Currently, after performing any MockMvc request, the SecurityContext is cleared by ThreadLocalSecurityContextHolderStrategy#clearContext() which causes the following exception to be thrown when you use SecurityContext dependent code during assertions:

`java.lang.IllegalArgumentException: Authentication object cannot be null`

Details can be found here in SO: https://stackoverflow.com/questions/51622300/mockmvc-seems-to-be-clear-securitycontext-after-performing-request-java-lang-il/

**Context**

- Goal: I want to keep the SecurityContext during the whole test execution. 
- Workaround: I could disable filters through @AutoConfigureMockMvc(addFilters = false). Does not seem like a nice option in all cases. 
- Current solution in my project using a MockMvcBuilderCustomizer
```
public class MockMvcTestSecurityContextPropagationCustomizer implements MockMvcBuilderCustomizer {

    @Override
    public void customize(ConfigurableMockMvcBuilder<?> builder) {
        builder.alwaysDo(result -> {
            log.debug("resetting SecurityContextHolder to TestSecurityContextHolder");
            SecurityContextHolder.setContext(TestSecurityContextHolder.getContext());
        });
    }

}
```

**Proposal**
My proposal is that this propagation could be done by the framework itself in org.springframework.boot.test.autoconfigure.web.servlet.MockMvcSecurityConfiguration.SecurityMockMvcBuilderCustomizer (or a another separate configuration within MockMvcSecurityConfiguration) but I am not sure if I miss some negative impact by doing that in general.

rwinch:
Thanks for the report. I don't think we can do this automatically as it would be very surprising around flows like log out.

What I could see is Spring Security creating `org.springframework.security.test.web.servlet.response.SecurityMockMvcResultHanlders.copyTestSecurityContextHolderToSecurityContextHolder()` which performs this logic in the returned `ResultHandler`. I'd have to give some thought to the name of this method as it is very verbose and I'd like something a bit more concise.

Then users would be in charge of applying the `ResultHandler` to `MockMvc`.

Would you be interested in submitting a pull request?

marcusdacoregio:
What do you think about `withSecurityContext(Consumer<SecurityContext>)` for the method signature?
[...]

PR #9737, opened 2021-05-10
Propagate TestSecurityContextHolder to SecurityContextHolder

marcusdacoregio:
Create SecurityMockMvcResultHandlers to define security related MockMvc ResultHandlers
Create a method to allow copying the SecurityContext from the TestSecurityContextHolder to SecurityContextHolder

Closes gh-9565

### DIFF
--- a/docs/manual/src/docs/asciidoc/_includes/servlet/test/mockmvc.adoc
+++ b/docs/manual/src/docs/asciidoc/_includes/servlet/test/mockmvc.adoc
@@ -1889,3 +1889,65 @@ mvc
     }
 ----
 ====
+
+=== SecurityMockMvcResultHandlers
+
+Spring Security provides a few ``ResultHandler``s implementations.
+In order to use Spring Security's ``ResultHandler``s implementations ensure the following static import is used:
+
+[source,java]
+----
+import static org.springframework.security.test.web.servlet.response.SecurityMockMvcResultHandlers.*;
+----
+
+==== Exporting the SecurityContext
+
+Often times we want to query a repository to see if some `MockMvc` request actually persisted in the database.
+In some cases our repository query uses the <<data,Spring Data Integration>> to filter the results based on current user's username or any other property.
+Let's see an example:
+
+A repository interface:
+[source,java]
+----
+private interface MessageRepository extends JpaRepository<Message, Long> {
+	@Query("SELECT m.content FROM Message m WHERE m.sentBy = ?#{ principal?.name }")
+	List<String> findAllUserMessages();
+}
+----
+
+Our test scenario:
+
+[source,java]
+----
+mvc
+	.perform(post("/message")
+		.content("New Message")
+		.contentType(MediaType.TEXT_PLAIN)
+	)
+	.andExpect(status().isOk());
+
+List<String> userMessages = messageRepository.findAllUserMessages();
+assertThat(userMessages).hasSize(1);
+----
+
+This test won't pass because after our request finishes, the `SecurityContextHolder` will be cleared out by the filter chain.
+We can then export the `TestSecurityContextHolder` to our `SecurityContextHolder` and use it as we want:
+
+[source,java]
+----
+mvc
+	.perform(post("/message")
+		.content("New Message")
+		.contentType(MediaType.TEXT_PLAIN)
+	)
+	.andDo(exportTestSecurityContext())
+	.andExpect(status().isOk());
+
+List<String> userMessages = messageRepository.findAllUserMessages();
+assertThat(userMessages).hasSize(1);
+----
+
+[NOTE]
+====
+Remember to clear the `SecurityContextHolder` between your tests, or it may leak amongst them
+====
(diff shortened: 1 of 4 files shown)


marcusdacoregio:
A new section in the [documentation](https://docs.spring.io/spring-security/site/docs/current/reference/html5/#test-mockmvc) named **SecurityMockMvcResultHandlers** should be added to show this new feature.

test/src/main/java/org/springframework/security/test/web/servlet/response/SecurityMockMvcResultHandlers.java
```diff
@@ -0,0 +1,61 @@
...
+	/**
+	 * Exports the {@link SecurityContext} from {@link TestSecurityContextHolder} to
+	 * {@link SecurityContextHolder}
+	 */
+	public static ResultHandler exportSecurityContext() {
+		return new ExportSecurityContextHandler();
+	}
+
+	/**
+	 * A {@link ResultHandler} that copy the {@link SecurityContext} from
```
jzheaux, line 46:
```
A {@link ResultHandler} that copies the ...
```

jzheaux (changes requested):
Nice, @marcusdacoregio! Please see my feedback inline.

marcusdacoregio:
Thank you very much for your review @jzheaux.

I've updated the PR based on your suggestions.

jzheaux (approved):

test/src/main/java/org/springframework/security/test/web/servlet/response/SecurityMockMvcResultHandlers.java
```diff
@@ -0,0 +1,61 @@
...
+public final class SecurityMockMvcResultHandlers {
+
+	private SecurityMockMvcResultHandlers() {
+	}
+
+	/**
+	 * Exports the {@link SecurityContext} from {@link TestSecurityContextHolder} to
+	 * {@link SecurityContextHolder}
+	 */
+	public static ResultHandler exportSecurityContext() {
```
rwinch, line 41: I'd like a name that is a bit more explicit as to what is happening. Perhaps `propagateTestSecurityContext`? I think a better name could be found that my suggestion, but the idea is it should convey what is being done.

docs/manual/src/docs/asciidoc/_includes/servlet/test/mockmvc.adoc
```diff
@@ -1889,3 +1889,65 @@ mvc
...
+		.content("New Message")
+		.contentType(MediaType.TEXT_PLAIN)
+	)
+	.andExpect(status().isOk());
+
+List<String> userMessages = messageRepository.findAllUserMessages();
+assertThat(userMessages).hasSize(1);
+----
+
+This test won't pass because after our request finishes, the `TestSecurityContextHolder` will be cleared out by the filter chain.
```
rwinch, line 1933:
```suggestion
This test won't pass because after our request finishes, the `SecurityContextHolder` will be cleared out by the filter chain.
```

rwinch (changes requested):
Thanks for the PR! I've left feedback inline.

test/src/main/java/org/springframework/security/test/web/servlet/response/SecurityMockMvcResultHandlers.java
jzheaux: "propagate", "retain" and "keep" all seem to imply that it will be retained across multiple `MockMvc` invocations in the same test. Would that be the case? If this isn't a concern, then my favorite among these options is "retain". If it is a concern, then I still like "export" best.

Looking more deeply into the support, I see that there is a request post-processor static method called `testSecurityContext()`. In that case, "xyzTestSecurityContext" does seem more consistent.

marcusdacoregio: I think we should indeed mention `TestSecurityContext` in the name. It would be explicit if we do `exportTestSecurityContext`, since it gives the idea of the `TestSecurityContext` being exported somewhere, and this also gives the opportunity to extend it to `exportTestSecurityContext(Consumer<SecurityContext> exportTo)`.

If we use names like **save**, **propagate**, **preserve**, all of them give the idea that the `TestSecurityContextHolder` will not be cleared out.

rwinch: > "propagate", "retain" and "keep" all seem to imply that it will be retained across multiple MockMvc invocations in the same test. Would that be the case? 

I am not a fan of save, retain, preserve, or keep because I interpret these to mean that the `TestSecurityContextHolder` will not be changed which is true regardless of the `ResultHandler` being used. The behavior is copy/export/propagate the `TestSecurityContextHolder` into the `SecurityContext`.

marcusdacoregio:
Thanks @rwinch. I've updated the PR based on your suggestions.

The method's name was changed to `exportTestSecurityContext`.

rwinch (approved):

Issue #9510, opened 2021-03-17
Update SAML JavaDoc to reference specification


jzheaux:
A number of methods in the SAML support point to Shibboleth explanations for definitions of various SAML terms. While it's quite convenient to link to a web page instead of a PDF, linking to the SAML specification is preferred since it's the source of truth.

One such example is `RelyingPartyRegistration#getEntityId`. Instead of linking to https://wiki.shibboleth.net/confluence/display/CONCEPT/EntityNaming it should link to https://www.oasis-open.org/committees/download.php/51890/SAML%20MD%20simplified%20overview.pdf#2.9%20EntityDescriptor

PR #9743, opened 2021-05-13
Update SAML JavaDoc to reference specification

marcusdacoregio:
Closes gh-9510

### DIFF
--- a/saml2/saml2-service-provider/core/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
+++ b/saml2/saml2-service-provider/core/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


marcusdacoregio:
I went through the docs that are still referencing the **Shibboleth** wiki but they are still valid references to their implementation

saml2/saml2-service-provider/core/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
```diff
@@ -134,7 +134,7 @@ public String getRegistrationId() {
 
 	/**
 	 * Get the relying party's <a href=
-	 * "https://wiki.shibboleth.net/confluence/display/CONCEPT/EntityNaming">EntityID</a>.
+	 * "https://www.oasis-open.org/committees/download.php/51890/SAML%20MD%20simplified%20overview.pdf#2.9%20EntityDescriptor">EntityID</a>.
```
jzheaux, line 137: Will you please update the copyright header in this file to include the current year?

jzheaux (changes requested):
Thanks for the PR, @marcusdacoregio! Please see my feedback inline.

saml2/saml2-service-provider/core/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
marcusdacoregio: Sure, thanks for pointing this out

jzheaux:

PR #9746, opened 2021-05-13
Provide KeyInfo as part of the Signature object when an object is signed

fhanik:
KeyInfo is typically provided as part of a SAML objects signature.

Example: https://login.run.pivotal.io/saml/metadata (check the child elements, //ds:Signature/ds:KeyInfo

Also mentioned on stack overflow: https://stackoverflow.com/questions/66369794/no-keyinfo-generator-provided-for-spring-security-saml2-service-provider


jzheaux:
Thanks, @fhanik, for the PR! This is now merged into `main` via adad2da9645f7638cfa27597dc8ba71ce8094ee3. Because the directory structure was refactored since the PR was submitted, I also did a rebase in 8e0a91d12f578d2f55201daf9bf51b799f844800

Issue #9727, opened 2021-05-06
Build failures on Windows Machine due to line feed character


marcusdacoregio:
**Describe the bug**
When running the command `./gradlew clean build` on the project, you should face some errors with the `checkFormat` tasks, asking for you to run `./gradlew format`.
Once you run the `format` command, it replaces all the line endings with the Windows default line feed character. Making a change in the entire codebase

**To Reproduce**
Set the `core.autocrlf` property from git to false (`git config --global core.autocrlf false`) and clone the repository. Run `./gradlew clean build`.

**Expected behavior**
The project should be build with SUCCESS

PR #9757, opened 2021-05-14
Update Check Windows CI Workflow

marcusdacoregio:
- Add slack integration for error reporting
- Rename from `check_windows` to `build_windows`
- Change the build command to match the one in `build_jdk_11`
- Add `build_windows` to the `needs` of deployment and failure check jobs

Issue gh-9727


marcusdacoregio:
The whole CI Build can be verified here https://github.com/spring-projects/spring-security/actions/runs/842913709

jzheaux:

Issue #9113, opened 2020-10-08
[WebFlux] The CSRF token will change If subscribe to `Mono<CsrfToken>` more than once in a single request


tt4g:
**Describe the bug**

I read #7719 and added a `SubscribeCsrfTokenWebFilter` that actively subscribe `Mono<CsrfToken>`.
However, if `Mono<CsrfToken>` is subscribed to after going through the `SubscribeCsrfTokenWebFilter` and during view rendering, the CSRF token will regenerated.

**To Reproduce**

The following code can reproduce the problem.

```java
@Bean
    public WebFilter subscribeCsrfTokenWebFilter() {
        return ((exchange, chain) -> {
            Mono<CsrfToken> csrfToken = exchange.getAttributeOrDefault(CsrfToken.class.getName(), Mono.empty());

            CsrfToken token = csrfToken.block(); // Generate CsrfToken.

            return csrfToken // Generate new CsrfToken!
                  .then(chain.filter(exchange));
        });
    }
```

If the `ServerHttpRequest` did not have a CSRF token, the
Each time `Mono<CsrfToken>` is subscribed, the repository generates and stores a token.

**Expected behavior**

The CSRF token should be generated only once during a single request.

**Sample**

See **To Reproduce**

**Details**

The `CsrfWebFilter` stores `Mono<CsrfToken>` in `ServerWebExchange` attributes that executes `ServerCsrfTokenRepository#generateToken(ServerWebExchange )` and `ServerCsrfTokenRepository#saveToken(ServerWebExchange , CsrfToken)` when the request did not have a CSRF token.

https://github.com/spring-projects/spring-security/blob/dbce9b5b66b6a8395b217c58e349dca3379accd1/web/src/main/java/org/springframework/security/web/server/csrf/CsrfWebFilter.java#L159-L174

This `Mono<CsrfToken>` repeats the generation and saving of a new token for each subscription.
`CookieCsrfTokenRepository` adds `Set-Cookie: XSRF-TOKEN=...` to the response header as many times as it is called.
`Set-Cookie` should be an override action, so all but the last CSRF token subscribed to in the application will be disabled.

tt4g:
I believe that adding `Mono#cache()` to `CsrfWebFilter#generateToken(ServerWebExchange)` will get around this problem.
Is there any reason why this can't be done?

```diff
    private Mono<CsrfToken> generateToken(ServerWebExchange exchange) {
        return this.csrfTokenRepository.generateToken(exchange)
[...]

PR #9760, opened 2021-05-15
Cache Mono that generate the CSRF token

tt4g:
Closes gh-9113

### DIFF
--- a/web/src/test/java/org/springframework/security/web/server/csrf/CsrfWebFilterTests.java
+++ b/web/src/test/java/org/springframework/security/web/server/csrf/CsrfWebFilterTests.java
@@ -16,6 +16,9 @@
 
 package org.springframework.security.web.server.csrf;
 
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.Mock;
@@ -225,6 +228,26 @@ public class CsrfWebFilterTests {
 		client.post().uri("/").contentType(MediaType.APPLICATION_FORM_URLENCODED)
 				.bodyValue(this.token.getParameterName() + "=" + this.token.getToken()).exchange().expectStatus()
 				.isForbidden();
+	}
+
+	// gh-9113
+	@Test
+	public void filterWhenSubscribingCsrfTokenMultipleTimesThenGenerateOnlyOnce() {
+		this.csrfFilter.setCsrfTokenRepository(this.repository);
+		given(this.repository.loadToken(any())).willReturn(Mono.empty());
+		AtomicInteger count = new AtomicInteger();
+		given(this.repository.generateToken(any())).willReturn(Mono.fromCallable(() -> {
+			count.incrementAndGet();
+			return this.token;
+		}));
+		given(this.repository.saveToken(any(), any())).willReturn(Mono.empty());
+		AtomicReference<Mono<CsrfToken>> tokenFromExchange = new AtomicReference<>();
+		given(this.chain.filter(any())).willReturn(
+				Mono.fromRunnable(() -> tokenFromExchange.set(this.get.getAttribute(CsrfToken.class.getName()))));
+		this.csrfFilter.filter(this.get, this.chain).block();
+		tokenFromExchange.get().block();
+		tokenFromExchange.get().block();
+		assertThat(count).hasValue(1);
 	}
 
 	@RestController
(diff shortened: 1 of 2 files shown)


web/src/test/java/org/springframework/security/web/server/csrf/CsrfWebFilterTests.java
```diff
@@ -227,6 +230,31 @@ public void doFilterWhenTokenIsNullThenNoNullPointer() {
...
+		AtomicInteger callSaveToken = new AtomicInteger(0);
+		given(this.repository.saveToken(any(), any())).willReturn(Mono.defer(() -> {
+			callSaveToken.incrementAndGet();
+			return Mono.empty();
+		}));
+		AtomicReference<Mono<CsrfToken>> tokenFromExchange = new AtomicReference<>();
+		given(this.chain.filter(any())).willReturn(
+				Mono.fromRunnable(() -> tokenFromExchange.set(this.get.getAttribute(CsrfToken.class.getName()))));
+		this.csrfFilter.filter(this.get, this.chain).block();
+		tokenFromExchange.get().block();
```
jzheaux, line 251: Instead of incrementing counters, would it be better to block multiple times and confirm that you receive the same value? For example:

```java
CsrfToken firstToken = tokenFromExchange.get().block();
CsrfToken secondToken = tokenFromExchange.get().block();
assertThat(firstToken).isEqualTo(secondToken);
```

This might simplify some of the instrumentation you are doing in the test. Also, since calling `saveToken` multiple times is not necessarily a problem for this test, removing it may make the test more resilient.

jzheaux (changes requested):
Thanks for the PR, @tt4g! Please see my feedback inline.

Also, will you please squash your commits and format the resulting commit like so:

```java
Commit Title

Closes gh-9113
```

web/src/test/java/org/springframework/security/web/server/csrf/CsrfWebFilterTests.java
tt4g: Thank you for your review and suggestions.
I'm working now.

tt4g:
@jzheaux The work is done.

web/src/test/java/org/springframework/security/web/server/csrf/CsrfWebFilterTests.java
```diff
@@ -227,6 +229,21 @@ public void doFilterWhenTokenIsNullThenNoNullPointer() {
 				.isForbidden();
 	}
 
+	@Test
+	public void filterWhenSubscribingCsrfTokenMultipleTimesThenGenerateOnlyOnce() {
```
jzheaux, line 233: I think this test passes whether or not the change in `CsrfWebFilter` is made. Will you please alter the test in such a way as to pass when your change to `CsrfWebFilter` is in place, but to fail when it isn't? I think `generateToken` might need to be instrumented differently.

web/src/test/java/org/springframework/security/web/server/csrf/CsrfWebFilterTests.java
```diff
@@ -227,6 +229,21 @@ public void doFilterWhenTokenIsNullThenNoNullPointer() {
 				.isForbidden();
 	}
 
+	@Test
```
jzheaux, line 232: I failed to mention earlier, will you please add `// gh-9113` to the top of this test for historical purposes?

jzheaux (changes requested):
Thanks for the updates, @tt4g! Please see my feedback inline.

web/src/test/java/org/springframework/security/web/server/csrf/CsrfWebFilterTests.java
tt4g: You are right. My confirmation was lacking.
Since `CsrfWebFilter` adds `Mono` generated from `generateToken` to the `ServerWebExchange` attribute, `Mono` itself is still cached.
I forgot that I have to pay attention to the number of times CSRF `Mono` is subscribed.

However, I could only think of `AtomicInteger` as an easy way to measure the number of calls to `generateToken`.
`MonoCacheTimeTest` was also testing the cache in the same way (https://github.com/reactor/reactor-core/blob/v3.4.6/reactor-core/src/test/java/reactor/core/publisher/MonoCacheTimeTest.java).
Another idea I came up with was to use `TestPublisher`, but I think this makes the test code harder to understand than it is now.

jzheaux (approved):

Issue #9742, opened 2021-05-13
DefaultSpringSecurityContextSource can't handle spaces in baseDn


jtnord:
**Describe the bug**

When given a valid URL such as `ldap://server/dc=org%20with%20space,dc=com,`  the `DefaultSpringSecurityContextSource`  sets the `baseDN` to `dc=org%20with%20space,dc=com` and not `dc=org with space,dc=com`.

This causes lookups to fail when setting the baseDN as it will be set to an invalid DN.

https://github.com/spring-projects/spring-security/blob/02474fdd842c63e6fa1d129d4a7e88afe12c387b/ldap/src/main/java/org/springframework/security/ldap/DefaultSpringSecurityContextSource.java#L69 gets the `baseDN` `raw path` ie URL encoded https://github.com/spring-projects/spring-security/blob/02474fdd842c63e6fa1d129d4a7e88afe12c387b/ldap/src/main/java/org/springframework/security/ldap/LdapUtils.java#L135 which just about works as it is then used to manipulate the URL as a String.

However later on it is just persisted to the Base DN.

https://github.com/spring-projects/spring-security/blob/02474fdd842c63e6fa1d129d4a7e88afe12c387b/ldap/src/main/java/org/springframework/security/ldap/DefaultSpringSecurityContextSource.java#L77

**To Reproduce**

using commit 39c4195c3557985f34dc725b79e8f3ed3d767e20 from https://github.com/jenkinsci/ldap-plugin/pull/93/commits/39c4195c3557985f34dc725b7  
run `mvn test -Dtest=LDAPDNEscapingTest#testSpacesInDNWithRootDN` 

The URL being provided to `DefaultSpringSecurityContextSource` in this case is `ldap://localhost:52752/dc=planet%20express,dc=com`

**Expected behavior**

The lookup of any user succeeds

but it fails.

```
   4.977 [id=31]	WARNING	h.security.LDAPSecurityRealm#throwUnlessConfigIsIgnorable: Failed communication with ldap server Ve2UCorQ7HhmkQyz783bXg== (ldap://localhost:52752), will _not_ try the next configuration
javax.naming.NameNotFoundException: [LDAP: error code 32 - NO_SUCH_OBJECT: failed for MessageType : SEARCH_REQUEST
Message ID : 2
    SearchRequest
        baseDn : 'dc=planet%20express,dc=com'
        filter : '(uid=fry)'
        scope : whole subtree
        typesOnly : false
        Size Limit : no limit
[...]

PR #9769, opened 2021-05-18
Handle encoded spaces in the root dn

sjohnr:
Fixes an issue where provider URLs passed to the constructor of the
DefaultSpringSecurityContextSource can be URL encoded, resulting in
an invalid base dn.

Closes gh-9742

### DIFF
--- a/ldap/src/integration-test/java/org/springframework/security/ldap/DefaultSpringSecurityContextSourceTests.java
+++ b/ldap/src/integration-test/java/org/springframework/security/ldap/DefaultSpringSecurityContextSourceTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -54,7 +54,17 @@ public class DefaultSpringSecurityContextSourceTests {
 
 	@Test
 	public void supportsSpacesInUrl() {
-		new DefaultSpringSecurityContextSource("ldap://myhost:10389/dc=spring%20framework,dc=org");
+		DefaultSpringSecurityContextSource contextSource = new DefaultSpringSecurityContextSource(
+				"ldap://myhost:10389/dc=spring%20framework,dc=org");
+		assertThat(contextSource.getBaseLdapPathAsString()).isEqualTo("dc=spring framework,dc=org");
+	}
+
+	// gh-9742
+	@Test
+	public void constructorWhenUrlEncodedSpacesWithPlusCharacterThenBaseDnIsProperlyDecoded() {
+		DefaultSpringSecurityContextSource contextSource = new DefaultSpringSecurityContextSource(
+				"ldap://blah:123/dc=spring+framework,dc=org ldap://blah:456/dc=spring+framework,dc=org");
+		assertThat(contextSource.getBaseLdapPathAsString()).isEqualTo("dc=spring framework,dc=org");
 	}
 
 	@Test
@@ -94,6 +104,7 @@ public class DefaultSpringSecurityContextSourceTests {
 	public void serverUrlWithSpacesIsSupported() {
 		DefaultSpringSecurityContextSource contextSource = new DefaultSpringSecurityContextSource(
 				this.contextSource.getUrls()[0] + "ou=space%20cadets,dc=springframework,dc=org");
+		assertThat(contextSource.getBaseLdapPathAsString()).isEqualTo("ou=space cadets,dc=springframework,dc=org");
 		contextSource.afterPropertiesSet();
 		contextSource.getContext("uid=space cadet,ou=space cadets,dc=springframework,dc=org", "spacecadetspassword");
 	}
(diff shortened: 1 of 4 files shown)


ldap/src/main/java/org/springframework/security/ldap/DefaultSpringSecurityContextSource.java
```diff
@@ -16,6 +16,7 @@
 
 package org.springframework.security.ldap;
 
+import java.net.URI;
```
jzheaux, line 19: Please update the copyright years for all files in the PR. For example, this one would change from `2002-2016` to `2002-2021`.

jzheaux (changes requested):
Thanks, @sjohnr! Please see my inlined feedback.

ldap/src/integration-test/java/org/springframework/security/ldap/DefaultSpringSecurityContextSourceTests.java
```diff
@@ -54,7 +54,9 @@ public void instantiationSucceedsWithExpectedProperties() {
 
 	@Test
 	public void supportsSpacesInUrl() {
-		new DefaultSpringSecurityContextSource("ldap://myhost:10389/dc=spring%20framework,dc=org");
+		DefaultSpringSecurityContextSource contextSource = new DefaultSpringSecurityContextSource(
```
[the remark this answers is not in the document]
sjohnr: Thanks! I've added a standalone integration test with a new ldif file for this case and verified it failed prior and passed after the code change.

ldap/src/integration-test/java/org/springframework/security/ldap/search/FilterBasedLdapUserSearchWithSpacesTests.java
```diff
@@ -0,0 +1,89 @@
...
+@RunWith(SpringRunner.class)
+@ContextConfiguration(classes = FilterBasedLdapUserSearchWithSpacesTests.ApacheDsContainerWithSpacesConfig.class)
+public class FilterBasedLdapUserSearchWithSpacesTests {
+
+	@Autowired
+	private DefaultSpringSecurityContextSource contextSource;
+
+	// gh-9742
+	@Test
+	public void basicSearchSucceeds() throws Exception {
```
jzheaux, line 51: With new tests, we use the following convention for test names:

```java
public void methodUnderTestWhenConditionThenResult
```

Will you please adjust the tests you added to match that convention?

ldap/src/main/java/org/springframework/security/ldap/DefaultSpringSecurityContextSource.java
```diff
@@ -74,7 +78,12 @@ public DefaultSpringSecurityContextSource(String providerUrl) {
 			rootDn = (rootDn != null) ? rootDn : urlRootDn;
 		}
 		setUrls(urls.toArray(new String[0]));
-		setBase(rootDn);
+		try {
+			setBase((rootDn != null) ? URLDecoder.decode(rootDn, StandardCharsets.UTF_8.toString()) : null);
```
jzheaux, line 82: For added readability, consider moving the exception handling out to a separate method, e.g.

```java
private static String urlDecode(String text) {
    try {
        return URLDecoder.decode(text, StandardCharsets.UTF_8.toString());
    }
    catch (UnsupportedEncodingException ex) {
        return new IllegalStateException(ex);
    }
}
```

ldap/src/main/java/org/springframework/security/ldap/DefaultSpringSecurityContextSource.java
```diff
@@ -74,7 +78,12 @@ public DefaultSpringSecurityContextSource(String providerUrl) {
...
+		}
+		catch (UnsupportedEncodingException ex) {
+			throw new IllegalArgumentException("Unable to URL decode base dn: " + rootDn);
```
jzheaux, line 85: I think `IllegalArgumentException` fits better when the user's input is the problem. However, `UnsupportedEncodingException` appears to be a problem with the application itself. Maybe `IllegalStateException` would be better.

ldap/src/main/java/org/springframework/security/ldap/DefaultSpringSecurityContextSource.java
```diff
@@ -136,7 +145,13 @@ public DefaultSpringSecurityContextSource(List<String> urls, String baseDn) {
 	private static String buildProviderUrl(List<String> urls, String baseDn) {
 		Assert.notNull(baseDn, "The Base DN for the LDAP server must not be null.");
 		Assert.notEmpty(urls, "At least one LDAP server URL must be provided.");
-		String trimmedBaseDn = baseDn.trim();
+		String trimmedAndEncodedBaseDn;
+		try {
+			trimmedAndEncodedBaseDn = URLEncoder.encode(baseDn.trim(), StandardCharsets.UTF_8.toString());
```
jzheaux, line 150: For added readability, consider moving the exception handling out to a separate method.

jzheaux (changes requested):
Thanks for the updates, @sjohnr! Please see my feedback inline.

ldap/src/main/java/org/springframework/security/ldap/DefaultSpringSecurityContextSource.java
```diff
@@ -147,11 +151,33 @@ private static String buildProviderUrl(List<String> urls, String baseDn) {
...
 
 	}
 
+	private static String encodeUrl(String url) {
+		String encodedUrl;
+		try {
+			encodedUrl = URLEncoder.encode(url.trim(), StandardCharsets.UTF_8.toString());
+		}
+		catch (UnsupportedEncodingException ex) {
+			throw new IllegalStateException("Unable to encode URL: " + url, ex);
```
jzheaux, line 167: Since this exception is essentially guaranteed to not be thrown, you might consider simplifying this to:

```java
throw new IllegalStateException(ex);
```

jzheaux (changes requested):
Thanks, @sjohnr! Please see my feedback inline.

Issue #9741, opened 2021-05-12
OAuth2ErrorResponseErrorHandler throws IllegalArgumentException for a nonstandard HTTP status code response


mjustin:
**Describe the bug**
I have a Spring application using `spring-security-oauth2-client`.  It authenticates via OAuth2 using [AuthorizedClientServiceOAuth2AuthorizedClientManager.authorize(OAuth2AuthorizeRequest authorizeRequest)](https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/oauth2/client/AuthorizedClientServiceOAuth2AuthorizedClientManager.html#authorize-org.springframework.security.oauth2.client.OAuth2AuthorizeRequest-).

The authorization endpoint is in [TIBCO Cloud Mashery](https://www.tibco.com/products/api-management).  Due to a typo in the URL, the OAuth endpoint failed with Mashery's custom [596](https://docs.mashery.com/design/GUID-01FE69BD-82F3-4471-B7A8-EF00798C4168.html) error code.  However, rather than return a sensible error in this scenario, the Spring code throws an `IllegalArgumentException` trying to convert the status code to an [`HttpStatus`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/HttpStatus.html):

```none
java.lang.IllegalArgumentException: No matching constant for [596]
	at org.springframework.http.HttpStatus.valueOf(HttpStatus.java:538)
	at org.springframework.http.client.AbstractClientHttpResponse.getStatusCode(AbstractClientHttpResponse.java:33)
	at org.springframework.security.oauth2.client.http.OAuth2ErrorResponseErrorHandler.handleError(OAuth2ErrorResponseErrorHandler.java:51)
	at org.springframework.web.client.ResponseErrorHandler.handleError(ResponseErrorHandler.java:63)
	at org.springframework.web.client.RestTemplate.handleResponse(RestTemplate.java:782)
	at org.springframework.web.client.RestTemplate.doExecute(RestTemplate.java:740)
	at org.springframework.web.client.RestTemplate.exchange(RestTemplate.java:641)
	at org.springframework.security.oauth2.client.endpoint.DefaultClientCredentialsTokenResponseClient.getTokenResponse(DefaultClientCredentialsTokenResponseClient.java:75)
	at org.springframework.security.oauth2.client.endpoint.DefaultClientCredentialsTokenResponseClient.getTokenResponse(DefaultClientCredentialsTokenResponseClient.java:52)
[...]

PR #9770, opened 2021-05-18
Handle custom status codes in OAuth2ErrorResponseErrorHandler

sjohnr:
Fixes an issue where custom status codes in the error response cause an
IllegalArgumentException to be thrown when resolving an HttpStatus.

Closes gh-9741


jgrandja (changes requested):

jgrandja (approved):

Issue #9685, opened 2021-04-27
Add support for any data type in Access Token Response


jgrandja:
The `expires_in` parameter in an OAuth 2.0 Access Token response should be a JSON number.

As per spec, in section 5.1 Successful Response:

> expires_in
         RECOMMENDED.  The lifetime **in seconds** of the access token.  For
         example, the value "3600" denotes that the access token will
         expire in one hour from the time the response was generated.
         If omitted, the authorization server SHOULD provide the
         expiration time via other means or document the default value.

Sample response:

```
{
       "access_token":"2YotnFZFEjr1zCsicMWpAA",
       "token_type":"example",
       "expires_in":3600,
       "refresh_token":"tGzv3JOkF0XG5Qx2TlKWIA",
       "example_parameter":"example_value"
     }
```

`OAuth2AccessTokenResponseHttpMessageConverter` should support `Converter`'s that operate on `Map<String, Object>` NOT the current `Map<String, String>`.

mychaint:
hi @jgrandja are you working on it

jgrandja:
@mychaint Yes, I'll be taking this one.

The change required here is not as simple as it looks. I will need to `@Deprecate` `setTokenResponseConverter()` and `setTokenResponseParametersConverter()` and provide equivalent setters that operate on `Map<String, Object>`.

We can't simply change `Map<String, String>` to `Map<String, Object>` as this would break existing applications.

I've scheduled this for `5.6` so I will likely not get to this for a bit still.

PR #9779, opened 2021-05-18
Access Token Response supports any data type

sjohnr:
Changed the converter used to convert a map into an OAuth2AccessTokenResponse to
support any object as the value, including json numbers and nested objects. Also
deprecated old setters/constructors and added new setters/factory methods.

Closes gh-9685

### DIFF
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/MapOAuth2AccessTokenResponseConverter.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/MapOAuth2AccessTokenResponseConverter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/MapOAuth2AccessTokenResponseConverter.java
```diff
@@ -38,59 +38,112 @@
...
-			return OAuth2AccessToken.TokenType.BEARER;
-		}
-		return null;
+	/**
+	 * Create a new {@link Converter} instance used to convert to an
+	 * {@link OAuth2AccessTokenResponse} that supports both scalar values and nested
+	 * objects.
+	 * @return A new {@link Converter}
+	 */
+	public static Converter<Map<String, ?>, OAuth2AccessTokenResponse> of() {
```
jgrandja, line 66: Providing a factory method instead of deprecating the class will keep this class around for 6.0 which is not what we want given that the generic type parameters in the class are incorrect. We want this class to go away in 6.0. So let's deprecate the class and introduce a new one with the correct generic type parameters.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AccessTokenResponseParametersConverter.java
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.util.StringUtils;
+
+/**
+ * A {@link Converter} that converts the provided {@link OAuth2AccessTokenResponse} to a
+ * {@code Map} representation of the OAuth 2.0 Access Token Response parameters.
+ *
+ * @author Steve Riesenberg
+ * @since 5.5
+ */
+public class OAuth2AccessTokenResponseParametersConverter
```
jgrandja, line 35: Rename to `DefaultOAuth2AccessTokenResponseMapConverter` and make class `final`

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AccessTokenResponseParametersConverter.java
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.util.CollectionUtils;
+import org.springframework.util.StringUtils;
+
+/**
+ * A {@link Converter} that converts the provided {@link OAuth2AccessTokenResponse} to a
+ * {@code Map} representation of the OAuth 2.0 Access Token Response parameters.
+ *
+ * @author Steve Riesenberg
+ * @since 5.5
```
jgrandja, line 33: This should be `5.6`

jgrandja (changes requested):

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/http/converter/OAuth2AccessTokenResponseHttpMessageConverter.java
```diff
@@ -108,26 +103,63 @@ protected void writeInternal(OAuth2AccessTokenResponse tokenResponse, HttpOutput
...
+	 * @deprecated Use {@link #setAccessTokenResponseConverter(Converter)} instead
 	 * @param tokenResponseConverter the {@link Converter} used for converting to an
 	 * {@link OAuth2AccessTokenResponse}
 	 */
+	@Deprecated
 	public final void setTokenResponseConverter(
 			Converter<Map<String, String>, OAuth2AccessTokenResponse> tokenResponseConverter) {
 		Assert.notNull(tokenResponseConverter, "tokenResponseConverter cannot be null");
-		this.tokenResponseConverter = tokenResponseConverter;
+		// gh-6463: Parse parameter values as Object in order to handle potential JSON
```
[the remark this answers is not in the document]
sjohnr: This was the original comment when the conversion to String was added. I'm refactoring this based on the discussion around deprecating the `protected` fields, but was thinking of leaving the comment for historical purposes, until this is removed in 6.0.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/http/converter/OAuth2AccessTokenResponseHttpMessageConverter.java
```diff
@@ -108,26 +103,63 @@ protected void writeInternal(OAuth2AccessTokenResponse tokenResponse, HttpOutput
...
+		// Object and then convert values to String
+		// @formatter:off
+		this.accessTokenResponseConverter = (source) -> tokenResponseConverter.convert(source.entrySet()
```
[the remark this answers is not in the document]
sjohnr: Resolving in favor of discussion around deprecating the `protected` fields.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/http/converter/OAuth2AccessTokenResponseHttpMessageConverter.java
```diff
@@ -108,26 +103,63 @@ protected void writeInternal(OAuth2AccessTokenResponse tokenResponse, HttpOutput
...
+	 * instead
 	 * @param tokenResponseParametersConverter the {@link Converter} used for converting
 	 * to a {@code Map} representation of the Access Token Response parameters
 	 */
+	@Deprecated
 	public final void setTokenResponseParametersConverter(
 			Converter<OAuth2AccessTokenResponse, Map<String, String>> tokenResponseParametersConverter) {
 		Assert.notNull(tokenResponseParametersConverter, "tokenResponseParametersConverter cannot be null");
-		this.tokenResponseParametersConverter = tokenResponseParametersConverter;
+		this.accessTokenResponseParametersConverter = (source) -> new LinkedHashMap<>(
```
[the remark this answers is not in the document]
sjohnr: Resolving in favor of discussion around deprecating the `protected` fields.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/http/converter/OAuth2AccessTokenResponseHttpMessageConverter.java
```diff
@@ -55,9 +56,9 @@
 
 	private GenericHttpMessageConverter<Object> jsonMessageConverter = HttpMessageConverters.getJsonMessageConverter();
 
-	protected Converter<Map<String, String>, OAuth2AccessTokenResponse> tokenResponseConverter = new MapOAuth2AccessTokenResponseConverter();
```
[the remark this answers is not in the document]
sjohnr: I have omitted the accessors for now, given that the `HttpMessageConverter` doesn't have one and the only use would be to override `readInternal`/`writeInternal` entirely. Since the converters are just classes, creating your own at need seems trivial enough. Let me know if you have additional thoughts on this.

oauth2/oauth2-core/src/test/java/org/springframework/security/oauth2/core/endpoint/MapOAuth2AccessTokenResponseConverterTests.java
```diff
@@ -19,20 +19,19 @@
...
 
 import org.springframework.security.oauth2.core.OAuth2AccessToken;
-import org.springframework.security.oauth2.core.OAuth2RefreshToken;
 
 /**
  * Tests for {@link MapOAuth2AccessTokenResponseConverter}.
  *
  * @author Nikita Konev
  */
+@Deprecated
```
jgrandja, line 34: This test class should be removed

oauth2/oauth2-core/src/test/java/org/springframework/security/oauth2/core/endpoint/OAuth2AccessTokenResponseMapConverterTests.java
```diff
@@ -32,6 +29,7 @@
  *
  * @author Nikita Konev
  */
+@Deprecated
```
jgrandja, line 32: This test class should be removed

oauth2/oauth2-core/src/test/java/org/springframework/security/oauth2/core/http/converter/OAuth2AccessTokenResponseHttpMessageConverterTests.java
```diff
@@ -22,6 +22,7 @@
 import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
```
jgrandja, line 25: Update copyright year

jgrandja (changes requested):
Thanks for the updates @sjohnr. After you apply the updates as per review, please go ahead and merge to main.

Also, please ensure the commit comment is up-to-date and as well delete the `gh-9685` branch after you merge.

Issue #9761, opened 2021-05-17
Fix Build Scan in Build Windows CI Job


marcusdacoregio:
**Describe the bug**
The step `Build with Gradle` in the Job `Build Windows` is finishing with the following message: `A build scan was not published as you have not authenticated with server 'ge.spring.io'.`. 

**To Reproduce**
Run the CI Workflow

**Expected behavior**
The build scan should be published for further analysis
Gradle should authenticate itself with the environment variables provided

PR #9786, opened 2021-05-20
Fix Build Scan not being published in Build Windows Job

marcusdacoregio:
Set environment variables via `env` job property
Fix the name of access key variable

Closes gh-9761, gh-9790

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -96,11 +96,11 @@ jobs:
           path: ~/.gradle/caches
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
       - name: Build with Gradle
-        run: |
-          set GRADLE_ENTERPRISE_CACHE_USERNAME="$GRADLE_ENTERPRISE_CACHE_USER"
-          set GRADLE_ENTERPRISE_CACHE_PASSWORD="$GRADLE_ENTERPRISE_CACHE_PASSWORD"
-          set GRADLE_ENTERPRISE_ACCESS_KEY="$GRADLE_ENTERPRISE_SECRET_ACCESS_KEY"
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -96,11 +96,11 @@ jobs:
           path: ~/.gradle/caches
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
       - name: Build with Gradle
-        run: |
-          set GRADLE_ENTERPRISE_CACHE_USERNAME="$GRADLE_ENTERPRISE_CACHE_USER"
-          set GRADLE_ENTERPRISE_CACHE_PASSWORD="$GRADLE_ENTERPRISE_CACHE_PASSWORD"
-          set GRADLE_ENTERPRISE_ACCESS_KEY="$GRADLE_ENTERPRISE_SECRET_ACCESS_KEY"
-          ./gradlew clean build --continue -PartifactoryUsername="$ARTIFACTORY_USERNAME" -PartifactoryPassword="$ARTIFACTORY_PASSWORD"
+        env:
```
jzheaux, line 99: Since this is now OS-independent it might be worth it to combine `build_jdk_11` and `build_windows` like so:

```
runs-on: [ubuntu-latest, windows-latest]
```

Will you please create a ticket to take a closer look at that?

jzheaux (approved):
Thanks, @marcusdacoregio! This PR looks good. I've added some inline feedback about another ticket I'm thinking should be filed.

.github/workflows/continuous-integration-workflow.yml
marcusdacoregio: Sounds good, I create ticket #9790 for that.

marcusdacoregio: Hi @jzheaux, would you please review this PR again as I have made some changes since we created #9790 ?

Issue #9610, opened 2021-04-12
Client credentials not correctly encoded in Basic Auth


DrZ7:
**Summary**
OAuth2AuthorizationGrantRequestEntityUtils.getTokenRequestHeaders does not work properly if client credentials contain special characters. From RFC 6749:
> Clients in possession of a client password MAY use the HTTP Basic
> authentication scheme as defined in [RFC2617] to authenticate with
> the authorization server. The client identifier is encoded using the
> "application/x-www-form-urlencoded" encoding algorithm per
> Appendix B, and the encoded value is used as the username; the client
> password is encoded using the same algorithm and used as the
> password.

**Actual Behavior**
The client with client name or password containing special characters cannot login. The provider returns exception.

**Expected Behavior**
The client with client name or password containing special characters can be authenticated.

**Configuration Sample**
spring.security.oauth2.client.registration.sth.client-secret = sthUI=+2~fubar

**Where**
org.springframework.security.oauth2.client.endpoint.OAuth2AuthorizationGrantRequestEntityUtils.getTokenRequestHeaders(ClientRegistration)

**Related**
This is related to https://github.com/spring-projects/spring-security-oauth/issues/1826

eleftherias:
Thanks for reaching out @DrZ7.
Which provider are you using when you see the exception?
Could you also share the stacktrace here?

DrZ7:
This ticket is the counterpart to https://github.com/spring-projects/spring-security-oauth/issues/1826...
Our provider works according to specification, it decodes client name and password.
Replacing 
`headers.setBasicAuth(clientRegistration.getClientId(), clientRegistration.getClientSecret());`
with 
`headers.setBasicAuth(URLEncoder.encode(clientRegistration.getClientId(), StandardCharsets.UTF_8),
          URLEncoder.encode(clientRegistration.getClientSecret(), StandardCharsets.UTF_8));`
fixes the issue...

stack with special character password and without encoding:
> org.springframework.security.oauth2.client.ClientAuthorizationException: [invalid_token_response] An error occurred while attempting to retrieve the OAuth 2.0 Access Token Response: 401 : [no body]
	at org.springframework.security.oauth2.client.ClientCredentialsOAuth2AuthorizedClientProvider.getTokenResponse(ClientCredentialsOAuth2AuthorizedClientProvider.java:96)
[...]

PR #9791, opened 2021-05-24
URL encode client credentials

sjohnr:
Closes gh-9610

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/AbstractWebClientReactiveOAuth2AccessTokenResponseClient.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/AbstractWebClientReactiveOAuth2AccessTokenResponseClient.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,9 @@
 
 package org.springframework.security.oauth2.client.endpoint;
 
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.Set;
 
@@ -97,7 +100,19 @@ public abstract class AbstractWebClientReactiveOAuth2AccessTokenResponseClient<T
 		headers.setAccept(Collections.singletonList(MediaType.APPLICATION_JSON));
 		if (ClientAuthenticationMethod.CLIENT_SECRET_BASIC.equals(clientRegistration.getClientAuthenticationMethod())
 				|| ClientAuthenticationMethod.BASIC.equals(clientRegistration.getClientAuthenticationMethod())) {
-			headers.setBasicAuth(clientRegistration.getClientId(), clientRegistration.getClientSecret());
+			String clientId = encodeClientCredential(clientRegistration.getClientId());
+			String clientSecret = encodeClientCredential(clientRegistration.getClientSecret());
+			headers.setBasicAuth(clientId, clientSecret);
+		}
+	}
+
+	private static String encodeClientCredential(String clientCredential) {
+		try {
+			return URLEncoder.encode(clientCredential, StandardCharsets.UTF_8.toString());
+		}
+		catch (UnsupportedEncodingException ex) {
+			// Will not happen since UTF-8 is a standard charset
+			throw new IllegalArgumentException(ex);
 		}
 	}
 
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/OAuth2AuthorizationGrantRequestEntityUtils.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/OAuth2AuthorizationGrantRequestEntityUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,9 @@
 
 package org.springframework.security.oauth2.client.endpoint;
 
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 
 import org.springframework.core.convert.converter.Converter;
@@ -48,9 +51,21 @@ final class OAuth2AuthorizationGrantRequestEntityUtils {
 		headers.addAll(DEFAULT_TOKEN_REQUEST_HEADERS);
 		if (ClientAuthenticationMethod.CLIENT_SECRET_BASIC.equals(clientRegistration.getClientAuthenticationMethod())
 				|| ClientAuthenticationMethod.BASIC.equals(clientRegistration.getClientAuthenticationMethod())) {
-			headers.setBasicAuth(clientRegistration.getClientId(), clientRegistration.getClientSecret());
+			String clientId = encodeClientCredential(clientRegistration.getClientId());
+			String clientSecret = encodeClientCredential(clientRegistration.getClientSecret());
+			headers.setBasicAuth(clientId, clientSecret);
 		}
 		return headers;
+	}
+
+	private static String encodeClientCredential(String clientCredential) {
+		try {
+			return URLEncoder.encode(clientCredential, StandardCharsets.UTF_8.toString());
+		}
+		catch (UnsupportedEncodingException ex) {
+			// Will not happen since UTF-8 is a standard charset
+			throw new IllegalArgumentException(ex);
+		}
 	}
 
 	private static HttpHeaders getDefaultTokenRequestHeaders() {
--- a/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/endpoint/OAuth2ClientCredentialsGrantRequestEntityConverterTests.java
+++ b/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/endpoint/OAuth2ClientCredentialsGrantRequestEntityConverterTests.java
@@ -16,6 +16,11 @@
 
 package org.springframework.security.oauth2.client.endpoint;
 
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.InOrder;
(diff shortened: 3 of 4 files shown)


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/AbstractOAuth2AuthorizationGrantRequestEntityConverter.java
```diff
@@ -69,6 +71,16 @@
 	 */
 	abstract MultiValueMap<String, String> createParameters(T authorizationGrantRequest);
 
+	/**
+	 * Sets the flag that controls whether client credentials are encoded using the
+	 * application/x-www-form-urlencoded algorithm in the headers converter.
+	 * @param encodeClientCredentials {@code true} to encode client credentials (default
+	 * is false)
+	 */
+	public void setEncodeClientCredentials(boolean encodeClientCredentials) {
```
jgrandja, line 80: Remove setter as this bug fix is scheduled for `5.5.1` so we want this to be a patch update.

As well, the client identifier and client secret MUST be encoded using application/x-www-form-urlencoded, as per [spec](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1):

> The client identifier is encoded using the
   "application/x-www-form-urlencoded" encoding algorithm per
   Appendix B, and the encoded value is used as the username; the client
   password is encoded using the same algorithm and used as the
   password.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/AbstractWebClientReactiveOAuth2AccessTokenResponseClient.java
```diff
@@ -215,4 +231,14 @@ public void setWebClient(WebClient webClient) {
 		this.webClient = webClient;
 	}
 
+	/**
+	 * Sets the flag that controls whether client credentials are encoded using the
+	 * application/x-www-form-urlencoded algorithm while populating token request headers.
+	 * @param encodeClientCredentials {@code true} to encode client credentials (default
+	 * is false)
+	 */
+	public void setEncodeClientCredentials(boolean encodeClientCredentials) {
```
jgrandja, line 240: Remove as per previous comment

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/AbstractWebClientReactiveOAuth2AccessTokenResponseClient.java
```diff
@@ -97,7 +102,18 @@ private void populateTokenRequestHeaders(T grantRequest, HttpHeaders headers) {
 		headers.setAccept(Collections.singletonList(MediaType.APPLICATION_JSON));
 		if (ClientAuthenticationMethod.CLIENT_SECRET_BASIC.equals(clientRegistration.getClientAuthenticationMethod())
 				|| ClientAuthenticationMethod.BASIC.equals(clientRegistration.getClientAuthenticationMethod())) {
-			headers.setBasicAuth(clientRegistration.getClientId(), clientRegistration.getClientSecret());
+			try {
+				String clientId = this.encodeClientCredentials
+						? URLEncoder.encode(clientRegistration.getClientId(), StandardCharsets.UTF_8.displayName())
```
jgrandja, line 107: I believe `StandardCharsets.UTF_8.displayName()` should be `StandardCharsets.UTF_8.name()`

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/AbstractWebClientReactiveOAuth2AccessTokenResponseClient.java
```diff
@@ -97,7 +102,18 @@ private void populateTokenRequestHeaders(T grantRequest, HttpHeaders headers) {
...
+						: clientRegistration.getClientId();
+				String clientSecret = this.encodeClientCredentials
+						? URLEncoder.encode(clientRegistration.getClientSecret(), StandardCharsets.UTF_8.displayName())
+						: clientRegistration.getClientSecret();
+				headers.setBasicAuth(clientId, clientSecret);
+			}
+			catch (UnsupportedEncodingException ex) {
+				throw new IllegalStateException("Unable to URL-encode client id or client secret", ex);
```
jgrandja, line 115: Let's throw `IllegalArgumentException` instead and ensure there is a test for this as well.

jgrandja (changes requested):

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/AbstractWebClientReactiveOAuth2AccessTokenResponseClient.java
sjohnr: I went with `.toString()` to be consistent with elsewhere.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/AbstractWebClientReactiveOAuth2AccessTokenResponseClient.java
```diff
@@ -97,7 +100,26 @@ private void populateTokenRequestHeaders(T grantRequest, HttpHeaders headers) {
...
 				|| ClientAuthenticationMethod.BASIC.equals(clientRegistration.getClientAuthenticationMethod())) {
-			headers.setBasicAuth(clientRegistration.getClientId(), clientRegistration.getClientSecret());
+			String clientId = encodeClientCredential(clientRegistration.getClientId());
+			String clientSecret = encodeClientCredential(clientRegistration.getClientSecret());
+			headers.setBasicAuth(clientId, clientSecret);
+		}
+	}
+
+	/**
+	 * Encodes client credentials using the "application/x-www-form-urlencoded" algorithm.
```
jgrandja, line 110: Remove javadoc as it's only needed for `public` or `protected` API's

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/OAuth2AuthorizationGrantRequestEntityUtils.java
```diff
@@ -48,11 +51,23 @@ static HttpHeaders getTokenRequestHeaders(ClientRegistration clientRegistration)
...
 		return headers;
 	}
 
+	private static String encodeClientCredential(String clientCredential) {
+		try {
+			return URLEncoder.encode(clientCredential, StandardCharsets.UTF_8.toString());
+		}
+		catch (UnsupportedEncodingException ex) {
+			// Will not happen since UTF-8 is a standard charset
+			throw new IllegalStateException(ex);
```
jgrandja, line 67: Throw `IllegalArgumentException` instead

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/endpoint/OAuth2ClientCredentialsGrantRequestEntityConverterTests.java
```diff
@@ -128,4 +133,35 @@ public void convertWhenGrantRequestValidThenConverts() {
 		assertThat(formParameters.getFirst(OAuth2ParameterNames.SCOPE)).contains(clientRegistration.getScopes());
 	}
 
+	// gh-9610
+	@SuppressWarnings("unchecked")
+	@Test
+	public void convertWhenSpecialCharactersThenConvertsWithEncodedClientCredentials()
+			throws UnsupportedEncodingException {
+		String clientSecretWithAnsiKeyboardSpecialCharacters = "~!@#$%^&*()_+{}|:\"<>?`-=[]\\;',./ ";
```
jgrandja, line 141: Also test `clientId` with special characters

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/endpoint/WebClientReactiveClientCredentialsTokenResponseClientTests.java
```diff
@@ -89,6 +93,35 @@ public void getTokenResponseWhenHeaderThenSuccess() throws Exception {
...
+		// @formatter:off
+		enqueueJson("{\n"
+			+ "  \"access_token\":\"MTQ0NjJkZmQ5OTM2NDE1ZTZjNGZmZjI3\",\n"
+			+ "  \"token_type\":\"bearer\",\n"
+			+ "  \"expires_in\":3600,\n"
+			+ "  \"refresh_token\":\"IwOGYzYTlmM2YxOTQ5MGE3YmNmMDFkNTVk\",\n"
+			+ "  \"scope\":\"create\"\n"
+			+ "}");
+		// @formatter:on
+		String clientSecretWithAnsiKeyboardSpecialCharacters = "~!@#$%^&*()_+{}|:\"<>?`-=[]\\;',./ ";
```
jgrandja, line 108: Also test `clientId` with special characters

jgrandja (changes requested):
Thanks for the updates @sjohnr. Please apply the updates as per review and then go ahead and merge / backport.

Also, please update the commit comment - remove "if enabled"

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/endpoint/OAuth2ClientCredentialsGrantRequestEntityConverterTests.java
```diff
@@ -128,4 +133,37 @@ public void convertWhenGrantRequestValidThenConverts() {
...
+				clientRegistration);
+		RequestEntity<?> requestEntity = this.converter.convert(clientCredentialsGrantRequest);
+		assertThat(requestEntity.getMethod()).isEqualTo(HttpMethod.POST);
+		assertThat(requestEntity.getUrl().toASCIIString())
+				.isEqualTo(clientRegistration.getProviderDetails().getTokenUri());
+		HttpHeaders headers = requestEntity.getHeaders();
+		assertThat(headers.getAccept()).contains(MediaType.APPLICATION_JSON_UTF8);
+		assertThat(headers.getContentType())
+				.isEqualTo(MediaType.valueOf(MediaType.APPLICATION_FORM_URLENCODED_VALUE + ";charset=UTF-8"));
+		String urlEncodedClientCredential = URLEncoder.encode(clientCredentialWithAnsiKeyboardSpecialCharacters,
```
OrangeDog, line 158: @sjohnr Rather than compare it to exactly the same code, actually specify the expected encoded version. e.g. with the case from the RFC:
```
assertEquals("+%25%26%2B%C2%A3%E2%82%AC", encodeClientCredential(" %&+£€"));
```

sjohnr: @OrangeDog that's a good idea too. However, instead of testing the private method, we could probably test the end result with hard-coded characters. Would you be interested in submitting a PR with an additional test using characters like those (not from the ansi keyboard, as mine was), and asserting like:

```
assertThat(actualRequest.getHeader(HttpHeaders.AUTHORIZATION)).isEqualTo("Basic <hard-coded encoded credentials here>");
```

**Note:** I was mostly concerned with using the test case to ensure that the code complied with the spec, not ensuring specific characters are encoded correctly with Java's encoding mechanism, hence re-using the `URLEncoder` class.

Issue #9315, opened 2020-12-26
Spring Security MessageSource files lack of 'ExceptionTranslationFilter.insufficientAuthentication'


Venus-WXP:
this.messages.getMessage("`ExceptionTranslationFilter.insufficientAuthentication`", "Full authentication is required to access this resource")
All of messages_locale.properties files lack of this item.

jzheaux:
Good catch, @Venus-WXP! Are you able to contribute any translations in a PR?

spring-projects-issues:
If you would like us to look at this issue, please provide the requested information. If the information is not provided within the next 7 days this issue will be closed.

sjohnr:
Hi @Venus-WXP,

Looking at the `ExceptionTranslationFilter`, this message is already able to be localized through `MessageSource`, but as you noted, we lack several translations. Are you able to add a translation for one or more languages in your own messages_xyz.properties?

For watchers, we would love a contribution on this issue so don't hesitate if you are able to contribute one or more translations for the message: "Full authentication is required to access this resource"

Currently supported locales are:

- de
- en
- ja

Missing locales (that currently have other messages) are:

- ca
- cs_CZ
- es_ES
- fr
- it
- ko_KR
- lt
- mn_MN
- nl
- pl
- pt_BR
- pt_PT
- ru
- uk_UA
- zh_CN
- zh_TW

rubensa:
Added pull request for Spanish https://github.com/spring-projects/spring-security/pull/9822

PR #9822, opened 2021-05-27
Add Spanish translation

rubensa:
Fix https://github.com/spring-projects/spring-security/issues/9315 for Spanish


web/src/main/resources/org/springframework/security/web/messages_es.properties
```diff
@@ -0,0 +1 @@
+ExceptionTranslationFilter.insufficientAuthentication=Para acceder a este recurso se requiere autenticaci\u00f3n completa
```
sjohnr, line 1: Please also add this translation to `core/src/main/resources/org/springframework/security/messages_es_ES.properties`.

sjohnr (changes requested):
Hi @rubensa. Thanks for this submission! Please see my feedback inline.

Also, will you please squash your commits and format the resulting commit like so:

```
Commit Title

Closes gh-9315
```

rubensa:
@sjohnr Done!  Sorry for the delay.

rubensa:
@sjohnr Just realized that `PersistentTokenBasedRememberMeServices.cookieStolen` in `core/src/main/resources/org/springframework/security/messages_es_ES.properties` was in English so I translated it too.

Issue #6802, opened 2019-04-20
Form Login not possible when a single OAuth2 Provider is configured


netmikey:
### Summary

When using a Form Login, a single OAuth2 provider and the auto-generated login page, the auto-configured `AuthenticationEntryPoint` will redirect to the provider immediately, bypassing the login page and effectively preventing form login.

### Actual Behavior

When trying to access a protected resource, spring security will immediately redirect to the OAuth2 provider's authentication page instead of the local login page.

### Expected Behavior

When Form Login is configured, the login page should never be skipped.

### Configuration

```
    @Override
    protected void configure(HttpSecurity http) throws Exception {
        http.authorizeRequests().anyRequest().authenticated()
            .and()
            .oauth2Login()
            .and()
            .oauth2Client()
            .and()
            .formLogin().permitAll();
    }
```

```
spring.security.oauth2.client.registration.facebook.client-id=some-id
spring.security.oauth2.client.registration.facebook.client-secret=some-secret
```

### Version

5.1.4-RELEASE, not sure as of which version this happens.

### Sample

I don't have a sample, but I found the exact location of the bug:

https://github.com/spring-projects/spring-security/blob/2c136f7b6c5b9ea75430bc660489bb8b22c1e466/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2LoginConfigurer.java#L444-L453

The condition should check whether Form Login is enabled and don't apply the shortcut if it is.

jgrandja:
Thanks for the report @netmikey. Would you be interested in submitting a PR for this fix?

rhamedy:
I could work on this if @netmikey didn't want to 👍

jgrandja:
Thanks @rhamedy for the offer. Let's provide @netmikey a chance to respond. If it's not picked up by end of next week than feel free to take it on. Thanks!

netmikey:
Thanks, @jgrandja. Since I'm pretty busy those days, please go ahead @rhamedy 👍

rhamedy:
Thanks both :slightly_smiling_face: 

@jgrandja anything I should know or keep in mind when working towards a fix in addition to the pointers that @netmikey included in the PR description?

jgrandja:
@rhamedy Nothing that comes to mind at the moment. But feel free to ask any questions along the way. Thanks for taking this on.

rhamedy:
Hi @jgrandja 

[...]

PR #9827, opened 2021-05-27
Allow form login when single OAuth2 Provider is configured

sjohnr:
Closes gh-6802

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2LoginConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2LoginConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 
 package org.springframework.security.config.annotation.web.configurers.oauth2.client;
 
+import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -73,12 +74,14 @@ import org.springframework.security.web.authentication.ui.DefaultLoginPageGenera
 import org.springframework.security.web.savedrequest.RequestCache;
 import org.springframework.security.web.util.matcher.AndRequestMatcher;
 import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
+import org.springframework.security.web.util.matcher.AnyRequestMatcher;
 import org.springframework.security.web.util.matcher.NegatedRequestMatcher;
 import org.springframework.security.web.util.matcher.OrRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestHeaderRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
+import org.springframework.util.ReflectionUtils;
 
 /**
  * An {@link AbstractHttpConfigurer} for OAuth 2.0 Login, which leverages the OAuth 2.0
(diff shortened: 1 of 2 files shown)


config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2LoginConfigurerTests.java
```diff
@@ -860,6 +871,27 @@ protected void configure(HttpSecurity http) throws Exception {
...
+
+		@Override
+		protected void configure(HttpSecurity http) throws Exception {
+			// @formatter:off
+			http
+				.oauth2Login()
+					.clientRegistrationRepository(this.clientRegistrationRepository)
+					.and()
+				.formLogin()
+					.permitAll();
```
jgrandja, line 888: Remove `permitAll()` as it's applied by default.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2LoginConfigurerTests.java
```diff
@@ -440,6 +440,17 @@ public void oauth2LoginWithOneAuthorizationCodeClientAndOtherClientsConfiguredTh
 		assertThat(this.response.getRedirectedUrl()).matches("http://localhost/oauth2/authorization/google");
 	}
 
+	// gh-6802
+	@Test
+	public void oauth2LoginWithOneClientConfiguredAndFormLoginThenRedirectDefaultLoginPage() throws Exception {
```
jgrandja, line 445: Please move this test method just below `oauth2LoginWithOneClientConfiguredThenRedirectForAuthorization()`

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2LoginConfigurer.java
```diff
@@ -503,9 +503,14 @@ private AuthenticationEntryPoint getLoginEntryPoint(B http, String providerLogin
 				new OrRequestMatcher(loginPageMatcher, faviconMatcher), defaultEntryPointMatcher);
 		RequestMatcher notXRequestedWith = new NegatedRequestMatcher(
 				new RequestHeaderRequestMatcher("X-Requested-With", "XMLHttpRequest"));
+		RequestMatcher formLoginNotEnabled = (request) -> {
+			DefaultLoginPageGeneratingFilter defaultLoginPageGeneratingFilter = http
```
jgrandja, line 507: Let's move the variable initialization of `defaultLoginPageGeneratingFilter` outside of the lambda

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2LoginConfigurer.java
```diff
@@ -503,9 +503,14 @@ private AuthenticationEntryPoint getLoginEntryPoint(B http, String providerLogin
...
+					.getSharedObject(DefaultLoginPageGeneratingFilter.class);
+			return defaultLoginPageGeneratingFilter == null || !defaultLoginPageGeneratingFilter.isFormLoginEnabled();
```
jgrandja, line 509: Is there another way to determine `isFormLoginEnabled()` without exposing that new method? Ideally we want to patch this to older branches.

jgrandja (changes requested):
Thanks for the PR @sjohnr! This looks promising! Please see review comments.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2LoginConfigurer.java
sjohnr: I believe that this only works if this is done inside the lambda. That way, you don't have to worry about the order in which the configurers are registered, as it only matters the outcome and whether Form Login is configured in the end or not.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2LoginConfigurer.java
sjohnr: I don't believe so. At least not while using the `DefaultLoginPageGeneratingFilter`. If we rely on the `isEnabled()` alone it breaks existing tests because it changes redirect behavior with other combinations including just regular oauth2 with a single client. I'm not sure off hand another way to share this type of information with configurers, but I'll look into it.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2LoginConfigurer.java
sjohnr: On second thought, it looks like we *should* be able to rely on the `DefaultLoginPageConfigurer` being first, as it is registered by `HttpSecurityConfiguration#httpSecurity()` when subclassing `WebSecurityConfigurerAdapter` in a hello-world style application.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2LoginConfigurer.java
sjohnr: I've updated the PR to utilize `ReflectionUtils` as a way to work around this. Let me know your thoughts, @jgrandja.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OAuth2LoginConfigurerTests.java
```diff
@@ -367,6 +367,17 @@ public void oauth2LoginWithOneClientConfiguredThenRedirectForAuthorization() thr
...
```
jgrandja, line 372: When I run this test, `DefaultLoginPageGeneratingFilter.formLoginEnabled` is `false`, even though it's enabled via `formLogin()` config.

Issue #9811, opened 2021-05-26
spring-security-config kotlin-stdlib-jdk8 dependency isn't optional


bturner:
**Describe the bug**
`spring-security-config` 5.4.x and prior had `<optional>true</optional>` dependencies on Kotlin. In 5.5.0 the `<optional/>` tag has been removed, even though the Gradle file for `spring-security-config` [still indicates it's optional](https://github.com/spring-projects/spring-security/blob/main/config/spring-security-config.gradle#L39).

**To Reproduce**
* Add Maven dependency on `spring-security-config`
* Run `mvn dependency:tree` and observe that `kotlin-stdlib-jdk8` and others are `compile`-scoped dependencies

**Expected behavior**
* Kotlin should not be a hard dependency unless I'm using something that requires it

**Notes**
I'd assume this was intentional, except that I don't see any changes in the Gradle build for `spring-security-config` related to Kotlin--it's still shown as optional there. The new features talk about Reactive extension support for Kotlin, which made me wonder if perhaps that's what triggered this. If it's intentional, my apologies for the noise.

bturner:
[Here's the POM on Central](https://search.maven.org/artifact/org.springframework.security/spring-security-config/5.5.0/jar), which shows:
```xml
    <dependency>
      <groupId>org.jetbrains.kotlin</groupId>
      <artifactId>kotlin-stdlib-jdk8</artifactId>
      <version>1.5.0</version>
      <scope>compile</scope>
    </dependency>
```

Compared to [5.4.6](https://search.maven.org/artifact/org.springframework.security/spring-security-config/5.4.6/jar), which has:
```xml
    <dependency>
      <groupId>org.jetbrains.kotlin</groupId>
      <artifactId>kotlin-reflect</artifactId>
      <version>1.4.30</version>
      <scope>compile</scope>
      <optional>true</optional>
    </dependency>
    <dependency>
      <groupId>org.jetbrains.kotlin</groupId>
      <artifactId>kotlin-stdlib-jdk8</artifactId>
      <version>1.4.30</version>
      <scope>compile</scope>
      <optional>true</optional>
    </dependency>
```

(The `kotlin-reflect` dependency appears to have been removed.)

eleftherias:
@bturner I have pushed a fix to the latest 5.6.0-SNAPSHOT and 5.5.1-SNAPSHOT.
Could you please try it out and let me know if it fixes the issue?

bturner:
[...]

PR #9859, opened 2021-06-03
Remove hard dependency on Kotlin

eleftherias:
Closes gh-9811


eleftherias:
As of Kotlin 1.4, the Kotlin plugin adds the stdlib dependency by default.
In order to prevent that dependency from being added automatically (since it is optional in Spring Security) we can set `kotlin.stdlib.default.dependency` to `false`.
https://kotlinlang.org/docs/gradle.html#dependency-on-the-standard-library

Issue #9378, opened 2021-01-26
Add 'AnyEdit Eclipse' plugin screenshots to improve details for settings for contribution


manueljordan:
In the  [Mind the whitespace](https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.adoc#mind-the-whitespace) section appears 7 bullet points and later does mention about the `AnyEdit Eclipse plugin`

**Expected Behavior**

Get more details about how to configure correctly `AnyEdit Eclipse plugin`, through some screenshots about the correct settings, would be valuable for newcomers that want contribute to this project.

Seems that plugin resolve many of the 7 bullet points mentioned at the top of this post

**Current Behavior**

There are no indications

**Context**

If I want contribute to this project through STS, would be nice have all the bullet points well configured to avoid problems if I send a PR

I know this request is not critical, but would be valuable to have all the settings well configured

rwinch:
At this time a better plugin would be the [spring java format plugin](https://github.com/spring-io/spring-javaformat) which takes care of all the whitespace needs. Perhaps you could submit a PR that directs users to that?

manueljordan:
When I created a Test class and I executed `./gradlew build` I got many errors about:

- format
- wrong import packages
- missing this references

Now I see that all come from that plugin

Please, could you confirm that plugin covers all about the 7 bullet points and other "things"? It to remove by complete the `AnyEdit Eclipse plugin` and do mention to the other one. Even with your confirmation I am going to play with that plugin to confirm and avoid confusions for the developers

manueljordan:
I fixed this situation, both plugins are need it. Let me improve the documentation for this week or the other.

Is viable add Figures right? It to share and indicate the correct order for import packages in STS/Eclipse. If yes, please give/indicate me the correct structure and nomenclature about the directory names (where to store the figures) and files (figures) names.

Thanks for your understanding.

rwinch:
The project now uses the spring java format plugin.

PR #9865, opened 2021-06-03
Improve Contributing instructions

manueljordan:
It improves the Contributing documentation for: 

* Eclipse/STS users about a plugin
* In General about formatting

Closes gh-9378

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -70,7 +70,12 @@ Please carefully follow the whitespace and formatting conventions already presen
 Whitespace management tips
 
 . You can use the https://marketplace.eclipse.org/content/anyedit-tools[AnyEdit Eclipse plugin] to ensure spaces are used and to clean up trailing whitespaces.
+  Once installed this plugin an extra mandatory configuration must be completed to accomplish the _Unix (LF), not dos (CRLF) line endings_ mentioned in the previous list go to: `Preferences` -> `General` -> `Editors` -> `AnyEdit Tools` -> `Auto Convert` tab and check the `Fix line delimiters (CRLF to LF etc)` checkbox.
 . Use Git's `pre-commit.sample` hook to prevent invalid whitespace from being pushed out. You can enable it by moving `.git/hooks/pre-commit.sample` to `.git/hooks/pre-commit` and ensuring it is executable. For more information on hooks refer to https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks[https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks].
+
+= Formatting
+
+For the build process, it for the project or a specific module, is possible that appears some error messages about invalid formatting with details to fix, it because Spring Security uses the link:https://github.com/spring-io/spring-javaformat[spring-javaformat] plugin.
 
 = Add Apache license header to all new classes
 


CONTRIBUTING.adoc
```diff
@@ -70,8 +70,13 @@ Please carefully follow the whitespace and formatting conventions already presen
 Whitespace management tips
 
 . You can use the https://marketplace.eclipse.org/content/anyedit-tools[AnyEdit Eclipse plugin] to ensure spaces are used and to clean up trailing whitespaces.
+  Once installed this plugin an extra mandatory configuration must be completed to accomplish the _Unix (LF), not dos (CRLF) line endings_ mentioned in the previous list go to: `Preferences` -> `General` -> `Editors` -> `AnyEdit Tools` -> `Auto Convert` tab and check the `Fix line delimiters (CRLF to LF etc)` checkbox.
```
jzheaux, line 73: I think this paragraph is a candidate for backporting to additional versions. As such will you please separate this paragraph into a separate commit from the Spring JavaFormat changes?

CONTRIBUTING.adoc
```diff
@@ -70,8 +70,13 @@ Please carefully follow the whitespace and formatting conventions already presen
...
 . Use Git's `pre-commit.sample` hook to prevent invalid whitespace from being pushed out. You can enable it by moving `.git/hooks/pre-commit.sample` to `.git/hooks/pre-commit` and ensuring it is executable. For more information on hooks refer to https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks[https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks].
 
+= Formatting
```
jzheaux, line 76: Instead of this paragraph, I'm thinking that it might be simpler to update [the recommended build command](https://github.com/spring-projects/spring-security/blob/main/CONTRIBUTING.adoc#run-all-tests-prior-to-submission) to:

```bash
./gradlew clean format check
```

jzheaux (changes requested):
Thanks for the improvements, @manueljordan! Please see my feedback inline.

CONTRIBUTING.adoc
manueljordan: Hello!

That section I think is informative, according with:

* https://github.com/spring-projects/spring-security/issues/9378#issuecomment-769319547

CONTRIBUTING.adoc
manueljordan: >As such will you please separate this paragraph into a separate commit from the Spring JavaFormat changes?

Not sure if I am following you. Can you expand the idea pls?

CONTRIBUTING.adoc
jzheaux: While informative, some of it is unnecessary. If the contributor runs the `format` command when running, they won't run into said error messages in the first place, rendering the extra detail needless.

I can see it being of interest to indicate that Spring Security uses this plugin. Instead, we can replace the "Mind the Whitespace" and "Whitespace Management Tips" sections with an admonition:

```
[TIP]
The project uses https://github.com/spring-io/spring-javaformat[Spring JavaFormat] to 
manage whitespace requirements.
```

Since the `format` command makes all the necessary changes independent of IDE, providing IDE-specific whitespace details is just noise.

manueljordan: Now I see the confusion: The `Spring JavaFormat` covers and reports other errors apart of the _whitespace_ format control, I remember clearly 2 right now: 

* order of the import packages 
* missing `this.` use

And other about: 

* line boundaries 
* Missing addition of `Apache license` 
* Something about a missing line (before or after) of the _package_ declaration

that's why I created the new Formatting section, to let know to the Spring Security contributor that all the error messages about formatting are coming from that project and for more details can go there. The error messages appears when is executed either the _build_ and _test_ Gradle commands - if my memory does not fail me with `./gradlew clean format check` fix many (but not all) of that errors - _I am agree to include that command as you requested_ but in the new section

About the `AnyEdit Tools` - it is a specific custom Eclipse plugin, should I create a new _sub_ section for _AnyEdit Eclipse plugin_ within the _Mind the whitespace_ section? - I think it has sense

jzheaux: Gotcha, I see your point.

I'd recommend three things, then:

1. Change the default command recommended in the build to `./gradlew clean format check`. The reason for this is `format` is quick and it will be necessary anyway for most PRs.
2. Leave the Formatting section, but reword it with an emphasis on preparing the contribution, for example:
```
= Follow Formatting Conventions

Spring Security follows the Spring 
https://github.com/spring-projects/spring-framework/wiki/Code-Style[Code Style]
guidelines. Many of these are checked at build time using 
https://github.com/spring-io/spring-javaformat[Spring JavaFormat]. 
If your local build fails due to formatting concerns, please apply the fixes the build recommends.

Other than changes Spring JavaFormat applies, please preserve existing formatting.

= Follow Naming Conventions

Search the codebase using `git grep` and other tools to discover common naming conventions.
```
3. Remove the Mind the Whitespace section (the above would replace it). At this point, it doesn't seem that AnyEdit is necessary to meet the Spring Security formatting guidelines.

If #&#x2060;3 is a concern, can you elaborate on what Spring Security requirements AnyEdit fulfills that Spring JavaFormat does not? Note that Spring Security uses `.gitattributes` to ensure that line endings are LF in the repo. 

I'd very much prefer to keep the contribution instructions as short as possible.

manueljordan: Hello

> Change the default command recommended in the build to ./gradlew clean format check. The reason for this is format is quick and it will be necessary anyway for most PRs.

I think is better to add that command in the new `Formatting` section, it is tightly related. And in the build section (do you mean [Run all tests prior to submission](https://github.com/spring-projects/spring-security/blob/main/CONTRIBUTING.adoc#run-all-tests-prior-to-submission), right?) I am going to add a simple line indicating that if errors about formatting appears should be read it the link to the `Formatting` section.. Therefore all is located in its respective place. 

> The reason for this is format is quick and it will be necessary anyway for most PRs

Agree, but even with that - each section should be based according its own task (build, test, and formatting). Otherwise would be confuse for new contributors. And I think it would be valuable for pipelines for Jenkins

About **2**, the first part has sense. The second part about `Follow Naming Conventions`, I never used `the git grep` when I did do the other contribution - I received all the errors messages through _Spring JavaFormat_ 

> Remove the Mind the Whitespace section (the above would replace it)

No, it is valuable, some settings of that list must be accomplished without the _AnyEdit Eclipse_ plugin. Therefore through the IDE own features

> it doesn't seem that AnyEdit is necessary to meet the Spring Security formatting guidelines.

No, some settings (one manually - the other automatically) must be accomplished with this plugin.

So, some settings are accomplished through the IDE itself and others through the plugin (automatically and manually)

>If #⁠3 is a concern, can you elaborate on what Spring Security requirements AnyEdit fulfills that Spring JavaFormat does not? Note that Spring Security uses .gitattributes to ensure that line endings are LF in the repo.

The `AnyEdit` plugin helps to accomplish some settings required in the `Mind the whitespace` section. It is a mandatory plugin in STS/Eclipse. About `Spring JavaFormat` it works fine, it has two roles for me, detect and fix format. But is possible to create a new contribution without errors about formatting thanks to the experience gained with the `Spring JavaFormat` errors messages (detection) and mostly because `AnyEdit` is installed and doing its work

> I'd very much prefer to keep the contribution instructions as short as possible.

I understand, but I think is better for this scenario keep concrete and detailed the info. The `AnyEdit` plugin is mandatory for Eclipse/STS developers like me and `Spring JavaFormat` must be mentioned - it was confuse for me, the errors messages (from where they are come from?).

jzheaux: > Therefore all is located in its respective place.

As I see it, this document is a step-by-step guide. Step-by-step guides give information at the point the reader needs it. In the past, most formatting considerations needed to be addressed before any code was written. These days, there are IDE-independent Gradle tasks that can apply many of our rules automatically just before committing.

The `Mind the Whitespace` section really is quite out of date at this point. As far as I can tell, there are only two points that remain that are important for the reader to know:

* Preserve existing formatting; i.e. do not reformat code for its own sake
* Search the codebase using `git grep` and other tools to discover common naming conventions, etc.

The reader needs to know these because there are no mechanisms that do it for them.

Given just those two points, "Mind the Whitespace" doesn't seem like the right name anymore. It almost certainly is in the wrong place in the document, too.

But, I see now that this is more than this PR sets out to do, so let's leave it as-is.

As far as AnyEdit, I think it's telling that neither [Spring Boot](https://github.com/spring-projects/spring-boot/wiki/Working-with-the-Code#importing-into-eclipse) nor [Spring Framework](https://github.com/spring-projects/spring-framework/blob/main/import-into-eclipse.md) mentions AnyEdit in their docs, so I guess I'm still looking for you to help me understand what makes Spring Security so different as to need it.

Again, this is probably out of scope for this PR, so let's also leave this section be. If you like, you are always welcome to submit a separate PR that edits the AnyEdit section.

jzheaux: What we're left with for this PR is the formatting section. Please revise it to read like a step in the process. That step should appear in the correct order relative to when the developer would encounter it when preparing their PR. The step should not duplicate Gradle tasks already documented.

manueljordan: > As far as AnyEdit, I think it's telling that neither Spring Boot nor Spring Framework mentions AnyEdit in their docs, so I guess I'm still looking for you to help me understand what makes Spring Security so different as to need it.

I think because there is no mention about something _Unix (LF), not dos (CRLF) line endings_ where the AnyEdit plugin does that work. And I like I said the AnyEdit plugin is practically a mandatory plugin to be installed on Eclipse for any project. Perhaps they are assuming that is already installed.

Let me create in my side a fake branch to do something (create a new class with something) and execute the current instructions to build and test, thus I am going to polish this TR according what is mandatory and not. It will be done this weekend.

manueljordan: I understand your point to not depend perhaps in plugins, but how you can see in the Spring Framework project, it does mention to have installed the Kotlin plugin and accomplish some extra steps - same as I want to contribute here to give explicit instructions according my direct experience. I think my PR is correct but needs a polish

```
Install the Kotlin Plugin for Eclipse in Eclipse.
Install the Eclipse Groovy Development Tools in Eclipse.
Switch to Groovy 2.5 (Preferences -> Groovy -> Compiler -> Switch to 2.5...) in Eclipse.
Change the Forbidden reference (access rule) in Eclipse from Error to Warning (Preferences -> Java -> Compiler -> Errors/Warnings -> Deprecated and restricted API -> Forbidden reference (access rule)).
```

manueljordan:
Hello @jzheaux 

I created for the _config_ module for a new branch 2 new classes - in the `src/java/main` and `src/java/test`, both classes with some invalid formats, for example `this.` was not used or the licence header was not included. 

I did do the following:

```
build (fails indicating to execute the "format" task)
format
build
   Now appears the following: Checkstyle rule violations were found. See the report at: file://.../spring-security/config/build/reports/checkstyle/main.html
   Fixed manually
build
```
What is weird now is that after to had executed the _format_ command the code (classes) were not automatically fixed, only was generated the `build/reports/checkstyle/main.html` file with all the details about the format errors - btw some errors (but not all) were printed in the console.

I remember clearly when I tried  to fix  bug in other pull request, when I executed the `format` command, it fixes automatically some errors format, but now not.

Something has changed? It prior to send the new pull request? I want avoid confusions for the developer/contributor

jzheaux:
Good question, @manueljordan. 

My experience has been that `format` takes care of whitespace. There are also CheckStyle checks, and those are typically up to the developer to work out.

CONTRIBUTING.adoc
jzheaux: What is AnyEdit providing that `.gitattributes` and Spring JavaFormat do not? For example, is AnyEdit necessary in order for the file to be readable in Eclipse? 

As I see it, the developer doesn't need to manage whitespace in the IDE anymore since all of Spring Security's whitespace concerns will be resolved once they call `format`. Unless you can share a concrete use case where AnyEdit corrects something that `format` does not, we should leave AnyEdit changes out of this PR and simply focus on introducing details about Spring JavaFormat.

manueljordan:
Let me include/indicate the explicit scenario - I mean something like: 

Once executed `build` command to build and apply tests - if your source code violates any format rules, you are going to be suggested to execute the `format` command to apply some default automatic fixes, and then again proceed to execute the `build` command, if any error persists, the errors (each one) with a quick description are going to be reported in a html file to be fix it by yourself.

Therefore the "Format" section is need it.

manueljordan:
>My experience has been that format takes care of whitespace

Even with that even true (I don't know because - I already use the _anyedit_ plugin), seems that other errors are not viable to be fixed automatically through the `format` command, such as the _license_ header.

I remember when a text has more of 120 characters by line, after to have applied the `format` command, it is split/divided in two lines. I tested that scenario again but didn't happen again

jzheaux:
> such as the license header

Agreed, though the instructions are not intended to indicate that AnyEdit is what one uses to add the license header.

What this comes down to is that we don't want to maintain instructions for how to use this or that plugin for this or that IDE. That's the job of that plugin and/or that IDE's documentation.

At this point, I think I'd like to instead look into [updating the contributing documentation](https://github.com/jzheaux/spring-security/blob/contributing/CONTRIBUTING.adoc) to better align with Spring Framework and other Spring projects. I'm going to close this PR because of that.

Issue #1890, opened 2011-02-01
SEC-1667: Consider adding a method to SecurityContextHolderStrategy that returns the current state of the SecurityContext without creating a new one


spring-projects-issues:
[Kyrill Alyoshin](https://jira.spring.io/secure/ViewProfile.jspa?name=kyrill007) (Migrated from [SEC-1667](https://jira.spring.io/browse/SEC-1667?redirect=false)) said:

We do have scenarios when we have to call SecurityContextHolder#getContext outside of web requests. (Obviously, the context will not be populated in such cases.) What will happen though is a new empty SecurityContext will be created and put on a ThreadLocal without being cleared by the servlet filter (as is the case during web requests). This will, of course, lead to class loader based memory leaks on hot redeploys. 

It sure would be nice to add, say, getExistingContext() method to SecurityContextHolderStrategy, which would not create a context if it is not available, and just return an existing one or null otherwise. Then we can call SecurityContextHolder.getContextHolderStrategy().getExistingContext() and 
we're safe.

What do you think?

rwinch:
This does not make sense to me. If you are accessing the SecurityContext you need to ensure it is cleared out. If you are outside of the context of a request then you should not be accessing the SecurityContext because it won't be populated. If you want to ensure it is populated, then whatever populates the context should clear it out. A likely option is to use Spring's [concurrency support](https://docs.spring.io/spring-security/site/docs/5.5.x/reference/html5/#concurrency). In particular, you could use [`DelegatingSecurityContextRunnable`](https://docs.spring.io/spring-security/site/docs/5.5.x/reference/html5/#delegatingsecuritycontextrunnable) or one of the higher level support classes.

jzheaux:
Given https://github.com/spring-projects/spring-security/pull/9877, I think this is something that should be reconsidered. In summary, Spring Security supplies a reactive hook that inspects the security context, and this hook is invoked during shutdown, causing a memory leak like the one described in this ticket.

More generally, it seems reasonable when operating at the framework level that there will be circumstances when performing a peek is valuable, somewhat similar to how Spring Security will often ask for an instance of the `HttpSession` without wanting to create it.

Instead of `getExistingContext()`, though, I believe `peekContext()` will be more descriptive. 

To maintain backward compatibility, it would need to be a `default` method on `SecurityContextHolderStrategy` that returns `getContext()`. This is so that `peekContext() == null` means that there is no context. That said, `peekContext()` should be overridden by all implementations to return the context.

[...]

PR #9877, opened 2021-06-08
Clear null authentication to fix ThreadLocal leak

### LINKED ISSUE #9841: Undeploy shows SecurityContextImpl [Null authentication] memory leak on tomcat
**Describe the bug**
Create a spring boot war application and deploy to tomcat server.
Undeploy the app from tomcat causes the following log.

```
SEVERE [Catalina-utility-1] org.apache.catalina.loader.WebappClassLoaderBase.checkThreadLocalMapForLeaks The web application [xxxxxx] created a ThreadLocal with key of type [java.lang.ThreadLocal] (value [java.lang.ThreadLocal@b58535e3]) and a value of type [org.springframework.security.core.context.SecurityContextImpl] (value [SecurityContextImpl [Null authentication]]) but failed to remove it when the web application was stopped. Threads are going to be renewed over time to try and avoid a probable memory leak.
```

**To Reproduce**
Put war file to webapps directory in tomcat and remove the war from webapps.
Spring boot v2.4.1, Java 11 and tomcat 9.0.41.

**Expected behavior**
`SecurityContextImpl [Null authentication]` is not created with ThreadLocal or removed on stopping.

If SecurityContextHolder strategy is MODE_GLOBAL, this doesn't occur.
I don't find how to remove `SecurityContextImpl [Null authentication]`.

shirosaki:
Fix #9841

### DIFF
--- a/core/src/main/java/org/springframework/security/core/context/SecurityContextHolder.java
+++ b/core/src/main/java/org/springframework/security/core/context/SecurityContextHolder.java
@@ -111,6 +111,14 @@ public class SecurityContextHolder {
 	}
 
 	/**
+	 * Peeks the current <code>SecurityContext</code>.
+	 * @return the security context (may be <code>null</code>)
+	 */
+	public static SecurityContext peekContext() {
+		return strategy.peekContext();
+	}
+
+	/**
 	 * Primarily for troubleshooting purposes, this method shows how many times the class
 	 * has re-initialized its <code>SecurityContextHolderStrategy</code>.
 	 * @return the count (should be one unless you've called
--- a/core/src/main/java/org/springframework/security/core/context/SecurityContextHolderStrategy.java
+++ b/core/src/main/java/org/springframework/security/core/context/SecurityContextHolderStrategy.java
@@ -39,6 +39,12 @@ public interface SecurityContextHolderStrategy {
 	SecurityContext getContext();
 
 	/**
+	 * Peeks the current context without creating an empty context.
+	 * @return a context (may be <code>null</code>)
+	 */
+	SecurityContext peekContext();
+
+	/**
 	 * Sets the current context.
 	 * @param context to the new argument (should never be <code>null</code>, although
 	 * implementations must check if <code>null</code> has been passed and throw an
--- a/core/src/main/java/org/springframework/security/core/context/ThreadLocalSecurityContextHolderStrategy.java
+++ b/core/src/main/java/org/springframework/security/core/context/ThreadLocalSecurityContextHolderStrategy.java
@@ -46,6 +46,11 @@ final class ThreadLocalSecurityContextHolderStrategy implements SecurityContextH
 	}
 
 	@Override
+	public SecurityContext peekContext() {
+		return contextHolder.get();
+	}
+
+	@Override
 	public void setContext(SecurityContext context) {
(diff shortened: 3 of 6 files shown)


jzheaux:
Hi, @shirosaki, thanks for the PR!

I don't think that we can safely clear the context here. This code doesn't know if it's the one to have populated the context. This may work in the special case of destroying a context. But, under normal operations, this may inadvertently log out a user.

Perhaps instead #1890 should be reopened and we consider adding a method to `SecurityContextHolder` that retrieves the context without creating one. Then, this PR can be updated to call that method instead.

shirosaki:
@jzheaux thanks for comments.
I also think `getExistingContext()` method which is not create null authentication object would be preferable.
Even if null authentication object is created, `clearContext()` is not called in `SecurityReactorContextConfiguration.java`.

core/src/main/java/org/springframework/security/core/context/SecurityContextHolderStrategy.java
```diff
@@ -38,6 +38,12 @@
 	 */
 	SecurityContext getContext();
 
+	/**
+	 * Peeks the current context without creating an empty context.
+	 * @return a context (may be <code>null</code>)
+	 */
+	SecurityContext peekContext();
```
jzheaux, line 45: For backward compatibility, this will need to have a default implementation. Please see https://github.com/spring-projects/spring-security/issues/1890#issuecomment-860929424 for more details on the recommended implementation.

core/src/main/java/org/springframework/security/core/context/ThreadLocalSecurityContextHolderStrategy.java
```diff
@@ -45,6 +45,11 @@ public SecurityContext getContext() {
 		return ctx;
 	}
 
+	@Override
```
jzheaux, line 48: Will you please update the copyright message for each edited file to include the year `2021`?

core/src/main/java/org/springframework/security/core/context/SecurityContextHolder.java
```diff
@@ -110,6 +110,14 @@ public static SecurityContext getContext() {
 		return strategy.getContext();
 	}
 
+	/**
+	 * Peeks the current <code>SecurityContext</code>.
+	 * @return the security context (may be <code>null</code>)
```
jzheaux, line 115: For each new public method that you introduce, will you please add `@since 5.6` to the JavaDoc?

jzheaux (changes requested):
Thanks, @shirosaki! I've left some feedback inline.

Also, in each of your two commit messages will you please include the phrase "Closes gh-xxxx" where `xxxx` is the related ticket?

jzheaux:
Thanks, @shirosaki! I've merged your contribution via 809ff883b03e44ac4053b51360fcd9ee270476fe and added some of the changes I specified in 78857c62f4b0a0596b4b386c3a8df319f0830f37.

Note also that upon some deeper investigation, `ThreadLocal` also initializes the value when `get` is called, which means that `peekContext` is not sufficient. As such, I altered the approach to delay the reading of the `SecurityContextHolder` until it's time to send the bearer token. In this way, reactive stacks that aren't actually using the `SecurityContextHolder` don't affect it.

Issue #3403, opened 2016-01-25
SEC-3204: InMemoryUserDetailsManager NullPointerException


spring-projects-issues:
# Updated

We should provide better error messages when an invalid user format is provided
# Original

[Henrik Sachse](https://jira.spring.io/secure/ViewProfile.jspa?name=henrik.sachse) (Migrated from [SEC-3204](https://jira.spring.io/browse/SEC-3204?redirect=false)) said:

The `org.springframework.security.provisioning.InMemoryUserDetailsManager` may throw a `NullPointerException` when an invalid user `Properties` file gets loaded (see attachments).

Can be reproduced with the following code having the attached file on the classpath:
```java
String invalidPropertiesFile = "invalid.properties";
ClassPathResource classPathResource = new ClassPathResource(invalidPropertiesFile);
Properties userProperties = new Properties();
userProperties.load(classPathResource.getInputStream());
// NullPointerException thrown here
InMemoryUserDetailsManager inMemoryUserDetailsManager = new InMemoryUserDetailsManager(userProperties);
```

That is because the method `org.springframework.security.core.userdetails.memory.UserAttributeEditor.setAsText(String)` may invoke the `java.beans.PropertyEditorSupport.setValue(Object)` method with `null` in case its parameter is an empty string.
```

**Comment by ghillert** [comment]
@rwinch Was hitting this issue today - 

The PR was closed. I actually have a minor use-case. A user may not necessarily have Authorities associated with their credentials. This works:

```
        users:
          bob: bobspassword, ROLE_ADMIN
          alice: alicepwd, ROLE_ADMIN
```

But this throws a null-pointer:

```
        users:
          bob: bobspassword
          alice: alicepwd
```

I believe the underlying `UserAttributeEditor` should be able to handle this gracefully without the null-pointer.

rwinch:
@ghillert Thanks for the feedback.

I think that it is reasonable to expect you wouldn't get a NPE. However, #256 would simply skip the invalid lines so you would not have the users added and there would be no error. See https://github.com/spring-projects/spring-security/pull/256#issuecomment-213025212

In short, I agree we do not want to have the NPE but the PR did not solve this correctly. This is why the PR is closed and this issue is still open.

Additional feedback welcome!

ghillert:
@rwinch Gotcha. I would expect, though, that if no roles are provided in the properties string that everything gets created without exception (or have that as an option, e.g. `ignoreMissingRoles`).

[...]

PR #9919, opened 2021-06-15
Improve Error Message for Invalid Properties in InMemoryUserDetailsManager 

marcusdacoregio:
Check if the UserAttribute is null when instantiating the InMemoryUserDetailsManager using the Properties argument's constructor

Closes gh-3403


jzheaux (changes requested):

marcusdacoregio:
Thanks for your review @jzheaux. I've updated the PR with your suggestions. Would you please review it again?

jzheaux (approved):

jzheaux:

PR #9924, opened 2021-06-16
Fix double lazy call in CsrfWebFilter

mucer:
If the filter matched the stream was delayed 2 times. Once with "flatMap" and once with "defer". This caused problems with "x-www-form-urlencoded" requests which did not send the body correctly.

We faced a problem that the body was never send, because some kind of ack from the server side was not received. I traced it down the the CSRF filter. With out the second delay, it works fine. But not sure how to write a unit test for this, since I did not figure out, why this causes such problems. If I should add a test, any assistance would be welcome how to test this.


eleftherias:
Thanks for reaching out @mucer.
To write a test we need a better understanding of your scenario.
When you say that the body was not sent correctly, where was it being sent to and from?
Was there an error or was there a different kind of unexpected behavior?
If you have a sample that reproduces the issue, that would be especially helpful.

eleftherias:
Hi @mucer, have you had the chance to look at my response? 
Some additional details would be very helpful in getting this issue resolved.

mucer:
Hi eleftherias,

I tried to create an unit test after your first comment but did not succeed. Currently I am working at a different project. But currently we have a custom bean with exactly this fix which we want to remove sooner or later. So when I switch back to the other project I will look into this issue again.

Without this fix the server won't accept the body after the header was received.

Issue #9923, opened 2021-06-16
Using the SecurityMockServerConfigurers.java requires the com.nimbusds oauth2-oidc-sdk on the classpath


frommeyerc:
**Describe the bug**
When using the `SecurityMockServerConfigurer` e.g. to test with an opaque OAuth2 token tests fail due to an NoClassDefFoundError. The root of the error is that in the `SecurityMockServerConfigurer` the `StringUtils from the com.nimbusds oauth2-oidc-sdk are used. This is an *optional* dependency of spring-security and in our case, where we're not dealing with oidc, not on the classpath.

Currently the Workaround is to add the sdk to the dependencies, but that should not be necessary.

**To Reproduce**
Use `SecurityMockServerConfigurer#mockOpaqueToken`  in a project that has no dependency to the oauth2-oidc-sdk.

**Expected behavior**
Dependency should not be mandatory for unrelated functionality.

PR #9925, opened 2021-06-16
Replace StringUtils from oauth2-oidc-sdk

frommeyerc:
SecurityMockServerConfigurers.java previously used the StringUtils from
the optional oauth2-oidc-sdk dependency. Replacing this with the
StringUtils from the spring framework this should not force adding the
dependency in cases where it is not actually used.

Closes gh-9923


jzheaux:
Thanks for the PR, @frommeyerc! This is now merged into `main` and will be backported.

PR #9937, opened 2021-06-17
Assert Error-Messages already includes dashes

OllisGit:
When the cert-content is not valid, the assert output message is not correct.
Because it outputs too many dashes .The const X509- and PKCS8-PEM_HEADER already includes the dashes.

I took the output message via copy and paste, but it was still not valid ;-(

Only the output is affected, the checks itself is correct.

### DIFF
--- a/core/src/main/java/org/springframework/security/converter/RsaKeyConverters.java
+++ b/core/src/main/java/org/springframework/security/converter/RsaKeyConverters.java
@@ -72,8 +72,8 @@ public final class RsaKeyConverters {
 		return (source) -> {
 			List<String> lines = readAllLines(source);
 			Assert.isTrue(!lines.isEmpty() && lines.get(0).startsWith(PKCS8_PEM_HEADER),
-					"Key is not in PEM-encoded PKCS#8 format, please check that the header begins with -----"
-							+ PKCS8_PEM_HEADER + "-----");
+					"Key is not in PEM-encoded PKCS#8 format, please check that the header begins with "
+							+ PKCS8_PEM_HEADER);
 			StringBuilder base64Encoded = new StringBuilder();
 			for (String line : lines) {
 				if (RsaKeyConverters.isNotPkcs8Wrapper(line)) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/springframework/security/converter/RsaKeyConverters.java
```diff
@@ -72,8 +72,8 @@ private RsaKeyConverters() {
 		return (source) -> {
 			List<String> lines = readAllLines(source);
 			Assert.isTrue(!lines.isEmpty() && lines.get(0).startsWith(PKCS8_PEM_HEADER),
-					"Key is not in PEM-encoded PKCS#8 format, please check that the header begins with -----"
-							+ PKCS8_PEM_HEADER + "-----");
+					"Key is not in PEM-encoded PKCS#8 format, please check that the header begins with "
```
marcusdacoregio, line 75: It would be nice to improve the two tests for these two methods to verify the message as well. One example would be:
```java
@Test
public void pkcs8WhenConvertingPkcs1PrivateKeyThenIllegalArgumentException() {
	assertThatIllegalArgumentException().isThrownBy(() -> this.pkcs8.convert(toInputStream(PKCS1_PRIVATE_KEY)))
				.withMessage("Key is not in PEM-encoded PKCS#8 format, please check that the header begins with " +
						"-----BEGIN PRIVATE KEY-----");
}
```
The same goes for `x509` method.

marcusdacoregio (changes requested):
Thanks for the PR @OllisGit. I've left some feedback inline.

When you are done with the changes, would you please [update the copyright header](https://github.com/spring-projects/spring-security/blob/main/CONTRIBUTING.adoc#update-apache-license-header-to-modified-files-as-necessary) of the class and test class?

Also, can you please open a ticket for this PR? You can create it as a bug. Then you should edit [your commit message](https://github.com/spring-projects/spring-security/blob/main/CONTRIBUTING.adoc#update-apache-license-header-to-modified-files-as-necessary) to have `Closes gh-xxxx` (where xxxx is the number of your ticket) in the end.

marcusdacoregio:
Hi @OllisGit, did you have the chance to take a look at the suggestions?

If not, I would be happy to do that. Please let me know.

Thanks.

OllisGit:
Hi @marcusdacoregio unfortunately I am really busy at the moment, so I am more then happy if you could take over the task.
Thx, in advance
Olli

Issue #9992, opened 2021-06-23
HttpSecurityContextRepository saves `@Transient Authentication` if `HttpSession` already exists


rwinch:
If an `HttpSession` already exists, then `HttpSecurityContextRepository` will save `@Transient Authentication` but should not. An example test:

```java
@Test
public void saveContextWhenTransientAuthenticationAndSessionExistsThenSkipped() {
	HttpSessionSecurityContextRepository repo = new HttpSessionSecurityContextRepository();
	MockHttpServletRequest request = new MockHttpServletRequest();
	request.getSession(); // ensure the session exists
	MockHttpServletResponse response = new MockHttpServletResponse();
	HttpRequestResponseHolder holder = new HttpRequestResponseHolder(request, response);
	SecurityContext context = repo.loadContext(holder);
	SomeTransientAuthentication authentication = new SomeTransientAuthentication();
	context.setAuthentication(authentication);
	repo.saveContext(context, holder.getRequest(), holder.getResponse());
	MockHttpSession session = (MockHttpSession) request.getSession(false);
	assertThat(Collections.list(session.getAttributeNames())).isEmpty();
}
```

rwinch:
Closing in favor of gh-9993

PR #9993, opened 2021-06-23
Prevent Save @Transient Authentication with existing HttpSession

rwinch:
Previously, `@Transient Authentication` would get saved if an existing
`HttpSession` existed but it shouldn't.

This commit always prevents `@Transient Authentication` from being saved.

Closes gh-9992


web/src/main/java/org/springframework/security/web/context/HttpSessionSecurityContextRepository.java
```diff
@@ -327,6 +330,9 @@ public AsyncContext startAsync(ServletRequest servletRequest, ServletResponse se
 		@Override
 		protected void saveContext(SecurityContext context) {
 			final Authentication authentication = context.getAuthentication();
+			if (isTransientAuthentication(authentication)) {
```
jzheaux, line 333: By moving the call to `isTransientAuthentication` out and to the beginning of `saveContext`, it seems like `createNewSessionIfAllowed`'s implementation and contract can be simplified by removing its call to `isTransientAuthentication`.

Should `createNewSessionIfAllowed` continue to call `isTransientAuthentication`?

jzheaux (changes requested):
Looks good, @rwinch. I left a polish suggestion inline.

web/src/main/java/org/springframework/security/web/context/HttpSessionSecurityContextRepository.java
rwinch: Thanks this is a good point. I've removed the unnecessary check in `createNewSessionIfAllowed`. Can you please verify and merge if things look good now?

jzheaux: It looks like the `Authentication` method parameter in `createNewSessionIfAllowed` is no longer necessary either. Can that be removed as well?

jawherallani:
This breaks my app when I switched from 5.6.6 to 5.7, as JwtAuthenticationToken which has @transient is being used:
the regular flow with 5.6.6:
- request A (websocket handshake) is fired, no http session yet, it fails and will retry
- api request B includes a jwt token in the headers, JwtAuthenticationToken authentication takes place and is then used to attach the SecurityContext on the http session as SPRING_SECURITY_CONTEXT attribute.
- request A, being retried, uses the http session to get the SecurityContext that was just saved and so is authenticated.

This flow breaks because of the @Transient now.

Is there something wrong with this flow? and why should @Transient Authentication be prevented from saving?

PR #10002, opened 2021-06-25
Accept predicate in constructor for JwtIssuerAuthenticationManagerRes…

barrypitman:
Add a constructor to JwtIssuerAuthenticationManagerResolver to allow it to accept a Predicate<String> to determine whether an issuer should be trusted or not. This allows for cases where the trusted issuers are not necessarily known at application startup. Since JwtIssuerAuthenticationManagerResolver is final and internal classes are private, this is not possible to extend it to support this use case without duplicating the whole class.

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
@@ -86,6 +86,16 @@ public final class JwtIssuerAuthenticationManagerResolver implements Authenticat
 				new TrustedIssuerJwtAuthenticationManagerResolver(
 						Collections.unmodifiableCollection(trustedIssuers)::contains));
 	}
+	
+	/**
+     * Construct a {@link JwtIssuerAuthenticationManagerResolver} using the provided
+     * parameters
+     * @param trustedIssuer a predicate to determine whether the issuer should be trusted or not
+     */
+    public JwtIssuerAuthenticationManagerResolver(Predicate<String> trustedIssuer) {
+        this.authenticationManager = new ResolvingAuthenticationManager(
+                new TrustedIssuerJwtAuthenticationManagerResolver(trustedIssuer));
+    }
 
 	/**
 	 * Construct a {@link JwtIssuerAuthenticationManagerResolver} using the provided


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
```diff
@@ -86,6 +86,16 @@ public JwtIssuerAuthenticationManagerResolver(Collection<String> trustedIssuers)
 				new TrustedIssuerJwtAuthenticationManagerResolver(
```
jzheaux, line 86: Will you please update the copyright message to now include `2021`?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
```diff
@@ -86,6 +86,16 @@ public JwtIssuerAuthenticationManagerResolver(Collection<String> trustedIssuers)
...
 						Collections.unmodifiableCollection(trustedIssuers)::contains));
 	}
+	
+	/**
+     * Construct a {@link JwtIssuerAuthenticationManagerResolver} using the provided
+     * parameters
+     * @param trustedIssuer a predicate to determine whether the issuer should be trusted or not
+     */
+    public JwtIssuerAuthenticationManagerResolver(Predicate<String> trustedIssuer) {
+        this.authenticationManager = new ResolvingAuthenticationManager(
```
jzheaux, line 96: Please check for null constructor parameters. You can look at the other constructors as examples.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
```diff
@@ -86,6 +86,16 @@ public JwtIssuerAuthenticationManagerResolver(Collection<String> trustedIssuers)
 				new TrustedIssuerJwtAuthenticationManagerResolver(
 						Collections.unmodifiableCollection(trustedIssuers)::contains));
 	}
+	
+	/**
+     * Construct a {@link JwtIssuerAuthenticationManagerResolver} using the provided
+     * parameters
+     * @param trustedIssuer a predicate to determine whether the issuer should be trusted or not
+     */
```
jzheaux, line 94: Please add `@since 5.6`

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
```diff
@@ -86,6 +86,16 @@ public JwtIssuerAuthenticationManagerResolver(Collection<String> trustedIssuers)
...
+    public JwtIssuerAuthenticationManagerResolver(Predicate<String> trustedIssuer) {
+        this.authenticationManager = new ResolvingAuthenticationManager(
+                new TrustedIssuerJwtAuthenticationManagerResolver(trustedIssuer));
+    }
```
jzheaux, line 98: Please add unit tests in `JwtIssuerAuthenticationManagerResolverTests`.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
```diff
@@ -86,6 +86,16 @@ public JwtIssuerAuthenticationManagerResolver(Collection<String> trustedIssuers)
...
```
jzheaux, line 99: Are you able to add this same functionality for `JwtIssuerReactiveAuthenticationManagerResolver`?

jzheaux (changes requested):
Thanks, @barrypitman! I've left some feedback inline.

In addition to that feedback, will you please [format your commit message](https://github.com/spring-projects/spring-security/blob/fe99c3b83bd41dc926832d01fd4313c8e852230a/CONTRIBUTING.adoc#format-commit-messages)?

jzheaux:
Hi, @barrypitman! Are you able to apply the requested changes?

jzheaux:
@barrypitman, thinking about this a bit more in conjunction with https://github.com/spring-projects/spring-security/pull/10476, I think that the best constructor for these use cases is `JwtIssuerAuthenticationManagerResolver(AuthenticationManagerResolver<String>)`, which already exists.

The reason is that introducing every constructor to allow for the myriad configurations applications want to do will become hard to maintain. In isolation, each seems reasonable and small, but the small changes start to add up. 

Instead, I believe we can simplify composing a custom `AuthenticationManagerResolver<String>`.

For example, an `AuthenticationManagerResolver<String>` with a custom `Predicate` would look something like this:

```java
public class MyAuthenticationManagerResolver implements AuthenticationManagerResolver<String> {
  Predicate<String> myPredicate = // ...

  @Cacheable(unless="#result==null")
  public AuthenticationManager resolve(String issuer) {
    if (!myPredicate.test(issuer)) {
      return null;
    }
    return new JwtAuthenticationProvider(JwtDecoders.fromIssuerLocation(issuer))::authenticate;
  }
}

// ...

@Bean 
AuthenticationManagerResolver<String> multitenantResolver() {
  return new JwtIssuerAuthenticationManagerResolver(new MyAuthenticationManagerResolver());
}
```

which, while already quite simple, could be further simplified by Spring Security exposing a default resolver like so:

```java
public static AuthenticationManager fromIssuerLocation(String issuer) {
  return new JwtAuthenticationProvider(JwtDecoders.fromIssuerLocation(issuer))::authenticate;
}
```

As such, I'm going to close this PR. Please let me know if I've missed something that makes it more complex than I've described, and you and I can take another look at simplifying your use case.

barrypitman:
No problem, thanks for the explanation Josh

Issue #9986, opened 2021-06-22
Docs should point to samples in minor version branches


marcusdacoregio:
**Expected Behavior**
Spring Security docs links that point to Spring Security Samples should point to the minor version branches instead of the version tags. Instead of pointing to `5.5.1` for example, it should point to `5.5.x`.

This way we don't have to release Spring Security Samples every time a Spring Security release is done. 

This should be backported to `5.5.x`.

**Current Behavior**

Today the docs are pointing to a specific version tag that doesn't have related samples release

gtiwari333:
This is still an issue. 

eg: https://docs.spring.io/spring-security/site/docs/current/reference/html5/#servlet-saml2login


The SAML 2 example is referring to non-existent  branch **5.5.1**.
> https://github.com/spring-projects/spring-security-samples/tree/5.5.1/servlet/spring-boot/java/saml2-login

marcusdacoregio:
Hi @gtiwari333. 

The **current** in the docs link of Spring Security refers to `5.5.1`. Since this was merged for `5.6.0-M1` and `5.5.2` it should be fixed when we do the next release. 

You can refer to the docs for the snapshot release, like so: https://docs.spring.io/spring-security/site/docs/5.5.2-SNAPSHOT/reference/html5/#servlet-saml2

Issue #9846, opened 2021-06-02
Verify Samples in Build


jzheaux:
To ensure that [the sample projects](https://github.com/spring-projects/spring-security-samples) stay up to date, they should be included as part of the Spring Security build.

marcusdacoregio:
We should use #10019 to decide which Spring Security Samples' branch to checkout

PR #10011, opened 2021-06-28
Update docs to point to minor version branches

marcusdacoregio:
Closes gh-9986

### DIFF
--- a/docs/manual/spring-security-docs-manual.gradle
+++ b/docs/manual/spring-security-docs-manual.gradle
@@ -18,7 +18,7 @@ asciidoctorj {
 	def ghTag = snapshotBuild ? 'main' : project.version
 	def ghUrl = "https://github.com/spring-projects/spring-security/tree/$ghTag"
 	def ghOldSamplesUrl = "https://github.com/spring-projects/spring-security/tree/5.4.x/samples"
-	def ghSamplesUrl = "https://github.com/spring-projects/spring-security-samples/tree/$ghTag"
+	def ghSamplesUrl = "https://github.com/spring-projects/spring-security-samples/tree/$samplesBranch"
 	def securityDocsUrl = "https://docs.spring.io/spring-security/site/docs/$docsTag"
 	def oldSecurityApiUrl = "https://docs.spring.io/spring-security/site/docs/5.4.x/api/"
 	def springVersion = "5.3.x"
--- a/docs/guides/spring-security-docs-guides.gradle
+++ b/docs/guides/spring-security-docs-guides.gradle
@@ -5,8 +5,7 @@ asciidoctor {
 	baseDir = file('src/docs/asciidoc')
 	options eruby: 'erubis'
 
-	def ghTag = snapshotBuild ? 'main' : project.version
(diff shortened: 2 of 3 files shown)


docs/manual/spring-security-docs-manual.gradle
```diff
@@ -16,9 +16,10 @@ asciidoctor {
 asciidoctorj {
 	def docsTag = snapshotBuild ? 'current' : project.version
 	def ghTag = snapshotBuild ? 'main' : project.version
+	def samplesBranch = computeSamplesBranch(project.version as String)
```
jzheaux, line 19: Can this be simplified to:

```groovy
def samplesBranch = project.version.replaceAll(/(\d+\.\d+\.).*/, '$1x');
```

docs/manual/spring-security-docs-manual.gradle
```diff
@@ -62,3 +63,12 @@ def resolvedVersions(Configuration configuration) {
 repositories {
 	maven { url "https://repo.spring.io/release" }
 }
+
+def computeSamplesBranch(String version) {
+	if ('main' == version) {
```
jzheaux, line 68: I don't think I understand... when could `project.version` be `main`? I would have thought `project.version` would always be a snapshot, milestone, or release version of the artifact.

docs/manual/spring-security-docs-manual.gradle
```diff
@@ -16,9 +16,10 @@ asciidoctor {
 asciidoctorj {
 	def docsTag = snapshotBuild ? 'current' : project.version
 	def ghTag = snapshotBuild ? 'main' : project.version
+	def samplesBranch = computeSamplesBranch(project.version as String)
 	def ghUrl = "https://github.com/spring-projects/spring-security/tree/$ghTag"
 	def ghOldSamplesUrl = "https://github.com/spring-projects/spring-security/tree/5.4.x/samples"
-	def ghSamplesUrl = "https://github.com/spring-projects/spring-security-samples/tree/$ghTag"
+	def ghSamplesUrl = "https://github.com/spring-projects/spring-security-samples/tree/$samplesBranch"
```
jzheaux, line 22: Should this same change be made in `spring-security-docs-guides`?

jzheaux (changes requested):

docs/manual/spring-security-docs-manual.gradle
marcusdacoregio: Yeah, and it cover more scenarios, thanks!

docs/manual/spring-security-docs-manual.gradle
marcusdacoregio: I've made a mistake between version and branch. Gonna fix it

marcusdacoregio:
Thanks for your review @jzheaux. I've introduced a new property in `gradle.properties` named `samplesBranch`. This way each Spring Security version should specify which samples' branch it uses and we avoid computing it from the version property, which is more error-prone.

marcusdacoregio:
I've also split this change into two commits so we can cherry-pick it in other branches easier, like #9846

jzheaux (approved):

PR #10025, opened 2021-06-30
Fix Saml2WebSsoAuthenticationRequestFilter javadoc

Kehrlann:
Fix (maybe copy-paste?) error in Saml2WebSsoAuthenticationRequestFilter. It responds to requests on `/saml2/authenticate/{registrationId}`.


eleftherias:
Thanks for the PR @Kehrlann! This is now merged into main.

PR #10050, opened 2021-07-08
Fix a typo in some class names in the oauth documentation

tmyroadctfig:
If I'm looking at the correct class (in nimbus-jose-7.8.1), there needs to be an extra 's' in the name of the classes here. I was searching for the name copied from the IDE and wasn't finding any matches.


tmyroadctfig:

marcusdacoregio:
Thanks for the PR @tmyroadctfig. This is now merged into main!

PR #10055, opened 2021-07-08
Fix Assertion

Russell-Allen:
Just fixing a trivial copy paste error with the variable name.


marcusdacoregio:
Thanks for the pr @Russell-Allen. It's nice that you found this error.

Could you please amend your commit message to `Fix Assertion`?

Russell-Allen:
I created the PR within github, and I am (so far) unable to find a way to amend the commit message interactively.  And it looks like I'd have to force push a new message if I pulled the repo down locally ...

Do I need to do the latter?  Any alternative suggestions?

marcusdacoregio:
Force-pushing commits is not a problem on your own branch. You can go ahead and do it so the message becomes more clear.

marcusdacoregio:
Thanks @Russell-Allen, I went ahead and changed the message. This is now merged into `main`.

Issue #9185, opened 2020-11-03
Add Saml2AuthenticationRequestRepository


jzheaux:
Related to https://github.com/spring-projects/spring-security/issues/9174

It would be nice to remember the AuthNRequest so that data like `InResponseTo` can be validated.

To do this will require a new interface that both `Saml2WebSsoAuthenticationRequestFilter` and `Saml2WebSsoAuthenticationFilter` can use. The first will store the AuthNRequest and the second will read and remove it.

The interface will look something like this:

```java
public interface Saml2AuthenticationRequestRepository<T extends AbstractSaml2AuthenticationRequest> {
    T loadAuthenticationRequest(HttpServletRequest request);

    void saveAuthenticationRequest(T authenticationRequest, HttpServletRequest request, HttpServletResponse response);

    T removeAuthenticationRequest(HttpServletRequest request, HttpServletResponse response);
}
```

And it should initially have a single implementation `HttpSessionSaml2AuthenticationRequestRepository` that stores the request in the session.

`Saml2AuthenticationToken` should also be altered to have a new constructor that takes the `RelyingPartyRegistration`, the SAML 2.0 response, and the `AbstractSaml2AuthenticationRequest` loaded from the repository.

fast-reflexes:
@jzheaux Why are the responses needed? Also, can we assume that a session exists?

jzheaux:
@fast-reflexes, thanks for the questions

> Why are the responses needed?

I'm not sure I understand the question, are you referring to SAML 2.0 responses? They are needed as that is the response an IDP gives in order to complete the authentication process.

> Also, can we assume that a session exists?

No, we should not assume it exists. But, if it does exist, then we can attempt to look up an associated AuthnRequest from it.

fast-reflexes:
@jzheaux Thanks for answering! Nah, I meant the responses in the function signatures of your interface? The SAML2 response is in the request and so is a potential session so why do we need to include the responses as argument? I can do it, just curious :)

Also in the `Saml2WebSsoAuthenticationRequestFilter`, if a session doesn't exist, we just don't save the request or do we take any measures to create a session there and then?

jzheaux:
Ah, got it, @fast-reflexes. 

The reason is from a lesson we learned in OAuth support. Most will store the AuthNrequest in the session, but some may want to store it in the browser, in a cookie, for example. Applications will need the `HttpServletResponse` to do that. The `HttpSession` implementation wouldn't use that parameter.

[...]

Issue #9174, opened 2020-11-02
Add InResponseTo validation support


fast-reflexes:
**Describe the bug**
The InResponseTo field is not validated and the repsonse is not rejected when this does not correspond to a sent request.

**To Reproduce**
1. Set up SAML2 login and run your server locally which redirects to an IDP which will then redirect to an actual deployed SAML2 authentication endpoint.
2. Set the host of your deployed app to 127.0.0.1 in /etc/hosts
3. Ping the host and verify that localhost responds
4. Start the local app and try to log in, you will be directed to your IDP which will redirect back to the authentication endpoint which will not exist on your local machine, thereby resulting in a not found error.
5. Copy the unresponded request as cURL from the console and paste it in a document.
6. Add flags .L and -i and remove cookies headers to be on the safe side.
7. Remove the entry in /etc/hosts and verify that the actual deployed application responds on ping.
7. Run the command with cURL and verify that the deployed application authentication endpoint accepts the cURL command which is a response to a request NOT initiated by the deployed application but by the local application.

**Expected behavior**
If the request contains a InResponseTo which is incorrect, it should be rejected.

**Sample**

A link to a GitHub repository with a [minimal, reproducible sample](https://stackoverflow.com/help/minimal-reproducible-example).

Reports that include a sample will take priority over reports that do not.
At times, we may require a sample, so it is good to try and include a sample up front.

jzheaux:
@fast-reflexes, you are correct - this is because the AuthNRequest is not saved anywhere after the relying party generates and sends it to the asserting party.

I've created a ticket to add that support - https://github.com/spring-projects/spring-security/issues/9185 - where your contribution would be most welcome.

Once that is complete, we can take a closer look at this ticket to see how to best validate `InResponseTo`. In the meantime, would you mind if I changed this ticket to an enhancement and reworded it to `Add InResponseTo validation support`?

fast-reflexes:
@jzheaux Thanks for answering! No, go ahead and reword it!

bitrecycling:
@jzheaux I have the same problem currently. I have 2 thoughts on that topic currently:
1) I think validation is only really relevant for Responses to authn Request and logout request or am I mistaken? Also replay attacks with "inResponseTo"-Responses can be mitigated if "inResponseTo" is validated.

[...]

PR #10060, opened 2021-07-09
Add Saml2AuthenticationRequestRepository

marcusdacoregio:
Closes gh-9185

Related #9174

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -33,6 +33,7 @@ import org.springframework.security.config.annotation.web.configurers.AbstractAu
 import org.springframework.security.config.annotation.web.configurers.AbstractHttpConfigurer;
 import org.springframework.security.config.annotation.web.configurers.CsrfConfigurer;
 import org.springframework.security.core.Authentication;
+import org.springframework.security.saml2.provider.service.authentication.AbstractSaml2AuthenticationRequest;
 import org.springframework.security.saml2.provider.service.authentication.OpenSaml4AuthenticationProvider;
 import org.springframework.security.saml2.provider.service.authentication.OpenSaml4AuthenticationRequestFactory;
 import org.springframework.security.saml2.provider.service.authentication.OpenSamlAuthenticationProvider;
@@ -40,6 +41,8 @@ import org.springframework.security.saml2.provider.service.authentication.OpenSa
 import org.springframework.security.saml2.provider.service.authentication.Saml2AuthenticationRequestFactory;
 import org.springframework.security.saml2.provider.service.registration.RelyingPartyRegistration;
 import org.springframework.security.saml2.provider.service.registration.RelyingPartyRegistrationRepository;
+import org.springframework.security.saml2.provider.service.servlet.HttpSessionSaml2AuthenticationRequestRepository;
+import org.springframework.security.saml2.provider.service.servlet.Saml2AuthenticationRequestRepository;
 import org.springframework.security.saml2.provider.service.servlet.filter.Saml2WebSsoAuthenticationFilter;
 import org.springframework.security.saml2.provider.service.servlet.filter.Saml2WebSsoAuthenticationRequestFilter;
 import org.springframework.security.saml2.provider.service.web.DefaultRelyingPartyRegistrationResolver;
@@ -206,6 +209,7 @@ public final class Saml2LoginConfigurer<B extends HttpSecurityBuilder<B>>
 		}
 		this.saml2WebSsoAuthenticationFilter = new Saml2WebSsoAuthenticationFilter(getAuthenticationConverter(http),
 				this.loginProcessingUrl);
+		setAuthenticationRequestRepository(http, this.saml2WebSsoAuthenticationFilter);
 		setAuthenticationFilter(this.saml2WebSsoAuthenticationFilter);
 		super.loginProcessingUrl(this.loginProcessingUrl);
 		if (StringUtils.hasText(this.loginPage)) {
@@ -250,6 +254,11 @@ public final class Saml2LoginConfigurer<B extends HttpSecurityBuilder<B>>
 		else {
 			this.saml2WebSsoAuthenticationFilter.setAuthenticationManager(this.authenticationManager);
 		}
+	}
+
+	private void setAuthenticationRequestRepository(B http,
+			Saml2WebSsoAuthenticationFilter saml2WebSsoAuthenticationFilter) {
+		saml2WebSsoAuthenticationFilter.setAuthenticationRequestRepository(getAuthenticationRequestRepository(http));
 	}
 
 	private AuthenticationConverter getAuthenticationConverter(B http) {
@@ -302,6 +311,16 @@ public final class Saml2LoginConfigurer<B extends HttpSecurityBuilder<B>>
 		return idps;
 	}
 
+	private Saml2AuthenticationRequestRepository<AbstractSaml2AuthenticationRequest> getAuthenticationRequestRepository(
+			B http) {
+		Saml2AuthenticationRequestRepository<AbstractSaml2AuthenticationRequest> repository = getBeanOrNull(http,
+				Saml2AuthenticationRequestRepository.class);
+		if (repository == null) {
+			return new HttpSessionSaml2AuthenticationRequestRepository();
+		}
+		return repository;
+	}
+
 	private <C> C getSharedOrBean(B http, Class<C> clazz) {
 		C shared = http.getSharedObject(clazz);
 		if (shared != null) {
@@ -339,8 +358,12 @@ public final class Saml2LoginConfigurer<B extends HttpSecurityBuilder<B>>
 		private Filter build(B http) {
 			Saml2AuthenticationRequestFactory authenticationRequestResolver = getResolver(http);
 			Saml2AuthenticationRequestContextResolver contextResolver = getContextResolver(http);
-			return postProcess(
-					new Saml2WebSsoAuthenticationRequestFilter(contextResolver, authenticationRequestResolver));
+			Saml2AuthenticationRequestRepository<AbstractSaml2AuthenticationRequest> repository = getAuthenticationRequestRepository(
+					http);
+			Saml2WebSsoAuthenticationRequestFilter filter = new Saml2WebSsoAuthenticationRequestFilter(contextResolver,
+					authenticationRequestResolver);
+			filter.setAuthenticationRequestRepository(repository);
+			return postProcess(filter);
 		}
 
 		private Saml2AuthenticationRequestFactory getResolver(B http) {
--- a/docs/manual/src/docs/asciidoc/_includes/servlet/saml2/saml2-login.adoc
+++ b/docs/manual/src/docs/asciidoc/_includes/servlet/saml2/saml2-login.adoc
@@ -1610,3 +1610,33 @@ http {
 The success handler will send logout requests to the asserting party.
 
 The request matcher will detect logout requests from the asserting party.
+
+[[servlet-saml2login-store-authn-request]]
+=== Storing the `AuthnRequest`
+
+The `Saml2AuthenticationRequestRepository` is responsible for the persistence of the `AuthnRequest` from the time the `AuthnRequest` <<servlet-saml2login-sp-initiated-factory,is initiated>> to the time the `SAMLResponse` <<servlet-saml2login-authenticate-responses,is received>>.
+The `Saml2AuthenticationTokenConverter` is responsible for loading the `AuthnRequest` from the `Saml2AuthenticationRequestRepository` and saving it into the `Saml2AuthenticationToken`.
+
+The default implementation of `Saml2AuthenticationRequestRepository` is `HttpSessionSaml2AuthenticationRequestRepository`, which stores the `AuthnRequest` in the `HttpSession`.
+
+If you have a custom implementation of `Saml2AuthenticationRequestRepository`, you may configure it by exposing it as a `@Bean` as shown in the following example:
+
+====
+.Java
+[source,java,role="primary"]
+----
+@Bean
+Saml2AuthenticationRequestRepository<AbstractSaml2AuthenticationRequest> authenticationRequestRepository() {
+	return new CustomSaml2AuthenticationRequestRepository();
+}
+----
+
+.Kotlin
+[source,kotlin,role="secondary"]
+----
+@Bean
+open fun authenticationRequestRepository(): Saml2AuthenticationRequestRepository<AbstractSaml2AuthenticationRequest> {
+    return CustomSaml2AuthenticationRequestRepository()
+}
+----
+====
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AuthenticationToken.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AuthenticationToken.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -38,6 +38,32 @@ public class Saml2AuthenticationToken extends AbstractAuthenticationToken {
 
 	private final String saml2Response;
 
+	private final AbstractSaml2AuthenticationRequest authenticationRequest;
+
+	/**
+	 * Creates a {@link Saml2AuthenticationToken} with the provided parameters.
+	 *
+	 * Note that the given {@link RelyingPartyRegistration} should have all its templates
+	 * resolved at this point. See
+	 * {@link org.springframework.security.saml2.provider.service.servlet.filter.Saml2WebSsoAuthenticationFilter}
+	 * for an example of performing that resolution.
+	 * @param relyingPartyRegistration the resolved {@link RelyingPartyRegistration} to
+	 * use
+	 * @param saml2Response the SAML 2.0 response to authenticate
+	 * @param authenticationRequest the {@code AuthNRequest} sent to the asserting party
+	 *
+	 * @since 5.6
+	 */
+	public Saml2AuthenticationToken(RelyingPartyRegistration relyingPartyRegistration, String saml2Response,
+			AbstractSaml2AuthenticationRequest authenticationRequest) {
+		super(Collections.emptyList());
+		Assert.notNull(relyingPartyRegistration, "relyingPartyRegistration cannot be null");
+		Assert.notNull(saml2Response, "saml2Response cannot be null");
+		this.relyingPartyRegistration = relyingPartyRegistration;
+		this.saml2Response = saml2Response;
+		this.authenticationRequest = authenticationRequest;
+	}
+
 	/**
 	 * Creates a {@link Saml2AuthenticationToken} with the provided parameters
 	 *
@@ -52,11 +78,7 @@ public class Saml2AuthenticationToken extends AbstractAuthenticationToken {
 	 * @since 5.4
 	 */
 	public Saml2AuthenticationToken(RelyingPartyRegistration relyingPartyRegistration, String saml2Response) {
-		super(Collections.emptyList());
-		Assert.notNull(relyingPartyRegistration, "relyingPartyRegistration cannot be null");
-		Assert.notNull(saml2Response, "saml2Response cannot be null");
-		this.relyingPartyRegistration = relyingPartyRegistration;
-		this.saml2Response = saml2Response;
+		this(relyingPartyRegistration, saml2Response, null);
 	}
 
 	/**
@@ -81,6 +103,7 @@ public class Saml2AuthenticationToken extends AbstractAuthenticationToken {
 						.entityId(idpEntityId).singleSignOnServiceLocation(idpEntityId))
 				.build();
 		this.saml2Response = saml2Response;
+		this.authenticationRequest = null;
 	}
 
 	/**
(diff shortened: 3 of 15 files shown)


saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AuthenticationToken.java
```diff
@@ -38,6 +38,32 @@
...
+	 * @param relyingPartyRegistration the resolved {@link RelyingPartyRegistration} to
+	 * use
+	 * @param saml2Response the SAML 2.0 response to authenticate
+	 * @param authenticationRequest the {@code AuthNRequest} sent to the asserting party
+	 *
+	 * @since 5.6
+	 */
+	public Saml2AuthenticationToken(RelyingPartyRegistration relyingPartyRegistration, String saml2Response,
+			AbstractSaml2AuthenticationRequest authenticationRequest) {
+		super(Collections.emptyList());
```
jzheaux, line 59: Ideally, this constructor would call the other constructor to reduce maintenance.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/servlet/HttpSessionSaml2AuthenticationRequestRepository.java
```diff
@@ -0,0 +1,72 @@
...
+		if (httpSession == null) {
+			return null;
+		}
+		return (AbstractSaml2AuthenticationRequest) httpSession.getAttribute(this.saml2AuthnRequestAttributeName);
+	}
+
+	@Override
+	public void saveAuthenticationRequest(AbstractSaml2AuthenticationRequest authenticationRequest,
+			HttpServletRequest request, HttpServletResponse response) {
+		HttpSession httpSession = request.getSession(false);
```
jzheaux, line 53: If someone is proactively saving it, I think it's reasonable to call `request.getSession()` instead.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/servlet/HttpSessionSaml2AuthenticationRequestRepository.java
```diff
@@ -0,0 +1,72 @@
...
+		if (httpSession == null || authenticationRequest == null) {
```
jzheaux, line 54: I think if the request is null, that should be the equivalent of removing the request from the session. This aligns with the Servlet spec, which says about `setAttribute`:

```javadoc
If the value passed in is null, this has the same effect as calling removeAttribute().
```

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/servlet/HttpSessionSaml2AuthenticationRequestRepository.java
```diff
@@ -0,0 +1,72 @@
...
+			return;
+		}
+		httpSession.setAttribute(this.saml2AuthnRequestAttributeName, authenticationRequest);
+	}
+
+	@Override
+	public AbstractSaml2AuthenticationRequest removeAuthenticationRequest(HttpServletRequest request,
+			HttpServletResponse response) {
+		AbstractSaml2AuthenticationRequest saml2AuthenticationRequest = this.loadAuthenticationRequest(request);
```
jzheaux, line 63: Some Spring Security classes still add `this.` to method calls, but it is not part of the style guide. Will you please remove `this.` from method calls?

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/servlet/HttpSessionSaml2AuthenticationRequestRepository.java
```diff
@@ -0,0 +1,72 @@
...
+		if (saml2AuthenticationRequest == null) {
+			return null;
+		}
+		HttpSession httpSession = request.getSession(false);
+		httpSession.removeAttribute(this.saml2AuthnRequestAttributeName);
```
jzheaux, line 68: This might throw an NPE since `getSession(boolean)` may return `null`. I'd recommend `request.getSession()` instead since if `loadAuthenticationRequest` returns non-null, then the session is present.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/servlet/filter/Saml2WebSsoAuthenticationFilter.java
```diff
@@ -95,6 +101,7 @@ protected boolean requiresAuthentication(HttpServletRequest request, HttpServlet
 	public Authentication attemptAuthentication(HttpServletRequest request, HttpServletResponse response)
 			throws AuthenticationException {
 		Authentication authentication = this.authenticationConverter.convert(request);
+		this.authenticationRequestRepository.removeAuthenticationRequest(request, response);
```
jzheaux, line 104: It seems that if `authentication == null` this might be an invalid request. Should the authnrequest be removed after it's clear that `authentication != null`?

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/servlet/filter/Saml2WebSsoAuthenticationFilter.java
```diff
@@ -103,4 +110,16 @@ public Authentication attemptAuthentication(HttpServletRequest request, HttpServ
...
+	/**
+	 * Use the given {@link Saml2AuthenticationRequestRepository} to remove the saved
+	 * authentication request.
+	 * @param authenticationRequestRepository the
+	 * {@link Saml2AuthenticationRequestRepository} to use
+	 * @since 5.6
+	 */
+	public void setAuthenticationRequestRepository(
+			Saml2AuthenticationRequestRepository<AbstractSaml2AuthenticationRequest> authenticationRequestRepository) {
+		this.authenticationRequestRepository = authenticationRequestRepository;
```
jzheaux, line 122: Please check for null. `Assert.notNull` is handy here.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverter.java
```diff
@@ -62,6 +66,23 @@ public Saml2AuthenticationTokenConverter(
...
+	 * Constructs a {@link Saml2AuthenticationTokenConverter} with the provided parameters
+	 * @param relyingPartyRegistrationResolver the strategy for resolving
+	 * {@link RelyingPartyRegistration}s
+	 * @param authenticationRequestRepository the repository to load the
+	 * {@link AbstractSaml2AuthenticationRequest} from
+	 * @since 5.6
+	 */
+	public Saml2AuthenticationTokenConverter(
+			Converter<HttpServletRequest, RelyingPartyRegistration> relyingPartyRegistrationResolver,
+			Saml2AuthenticationRequestRepository<AbstractSaml2AuthenticationRequest> authenticationRequestRepository) {
```
jzheaux, line 79: Usually, it's preferred to either have a setter or a constructor parameter, but not both. In this case, since `Saml2AuthenticationRequestRepository` is optional, I'd recommend removing the constructor parameter.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverter.java
```diff
@@ -50,6 +52,8 @@
 
 	private final Converter<HttpServletRequest, RelyingPartyRegistration> relyingPartyRegistrationResolver;
 
+	private Saml2AuthenticationRequestRepository<AbstractSaml2AuthenticationRequest> authenticationRequestRepository;
```
jzheaux, line 55: Is there a reason that this should not be set initially to `HttpSessionAuthenticationRequestRepository`?

saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/servlet/HttpSessionSaml2AuthenticationRequestRepositoryTest.java
```diff
@@ -0,0 +1,133 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+/**
+ * @author Marcus Da Coregio
+ */
+public class HttpSessionSaml2AuthenticationRequestRepositoryTest {
```
jzheaux, line 36: Please call this `Tests` instead of `Test`.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AuthenticationToken.java
```diff
@@ -38,6 +38,32 @@
 
 	private final String saml2Response;
 
+	private AbstractSaml2AuthenticationRequest authenticationRequest;
```
jzheaux, line 41: Since this won't change, I think it can be `final`.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/servlet/filter/Saml2WebSsoAuthenticationFilter.java
jzheaux: This method also probably needs to pass the provided request repository down to the authentication converter, if the converter is of the correct type.

jzheaux (changes requested):
Thanks for the PR, @marcusdacoregio!

In addition to the comments, please also add DSL support and docs. At a minimum, the DSL support should look for a bean of type `Saml2AuthorizationRequestRepository` and wire it in accordingly.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AuthenticationToken.java
marcusdacoregio: If I do this in the new constructor:
```java
public Saml2AuthenticationToken(RelyingPartyRegistration relyingPartyRegistration, String saml2Response,
			AbstractSaml2AuthenticationRequest authenticationRequest) {
	this(relyingPartyRegistration, saml2Response);
	this.authenticationRequest = authenticationRequest;
}
```
I can't make it `final` since it could be initialized inside the first constructor.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/servlet/filter/Saml2WebSsoAuthenticationFilter.java
marcusdacoregio: Yes, it should.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverter.java
marcusdacoregio: There isn't, I'll initialize it with `HttpSessionAuthenticationRequestRepository`.

marcusdacoregio:
Thanks for your review @jzheaux. I've updated the PR with your suggestions and the DSL and docs addition.

docs/manual/src/docs/asciidoc/_includes/servlet/saml2/saml2-login.adoc
```diff
@@ -1610,3 +1610,33 @@ http {
...
+
+[[servlet-saml2login-store-authn-request]]
+=== Storing the Authentication Request
+
+The `Saml2AuthenticationRequestRepository` is responsible for the persistence of the `AbstractSaml2AuthenticationRequest` from the time the Authentication Request is initiated to the time the Authentication Response is received.
+The `Saml2AuthenticationTokenConverter` is responsible for loading the `AbstractSaml2AuthenticationRequest` from the `Saml2AuthenticationRequestRepository` and saving it into the `Saml2AuthenticationToken`.
+
+The default implementation of `Saml2AuthenticationRequestRepository` is `HttpSessionSaml2AuthenticationRequestRepository`, which stores the `AbstractSaml2AuthenticationRequest` in the `HttpSession`.
+
+If you have a custom implementation of `Saml2AuthenticationRequestRepository`, you may configure it by exposing it as a *Bean* as shown in the following example:
```
jzheaux, line 1622: Instead of *emphasizing* words like "Bean", we should mark them as code, like `@Bean`.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AuthenticationToken.java
jzheaux: I think the constructors should be inverted. For example, do:

```java
public Saml2AuthenticationToken(RelyingPartyRegistration relyingPartyRegistration, String saml2Response) {
    this(relyingPartyRegistration, saml2Response, null);
```

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/servlet/filter/Saml2WebSsoAuthenticationFilter.java
```diff
@@ -100,7 +106,34 @@ public Authentication attemptAuthentication(HttpServletRequest request, HttpServ
...
+	public void setAuthenticationRequestRepository(
+			Saml2AuthenticationRequestRepository<AbstractSaml2AuthenticationRequest> authenticationRequestRepository) {
+		Assert.notNull(authenticationRequestRepository, "authenticationRequestRepository cannot be null");
+		this.authenticationRequestRepository = authenticationRequestRepository;
+		setAuthenticationRequestRepositoryIntoAuthenticationConverter(authenticationRequestRepository);
+	}
+
+	private void setAuthenticationRequestRepositoryIntoAuthenticationConverter(
+			Saml2AuthenticationRequestRepository<AbstractSaml2AuthenticationRequest> authenticationRequestRepository) {
+		if (!this.authenticationConverter.getClass().isAssignableFrom(Saml2AuthenticationTokenConverter.class)) {
```
jzheaux, line 132: Could you instead do:

```java
if (this.authenticationConverter instanceof Saml2AuthenticationTokenConverter) {
```

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/servlet/filter/Saml2WebSsoAuthenticationFilter.java
```diff
@@ -83,6 +88,7 @@ public Saml2WebSsoAuthenticationFilter(AuthenticationConverter authenticationCon
 		this.authenticationConverter = authenticationConverter;
 		setAllowSessionCreation(true);
 		setSessionAuthenticationStrategy(new ChangeSessionIdAuthenticationStrategy());
+		setAuthenticationRequestRepository(new HttpSessionSaml2AuthenticationRequestRepository());
```
jzheaux, line 91: I realize that this is what was done with `setSessionAuthenticationStrategy`; however, it's preferable to not call overrideable methods in a constructor. It may be better to do:

```java
private Saml2AuthenticationRequestRepository authenticationRequestRepository =
    new HttpSessionAuthenticationRequestRepository();
```

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverter.java
```diff
@@ -74,7 +79,27 @@ public Saml2AuthenticationToken convert(HttpServletRequest request) {
...
+	/**
+	 * Use the given {@link Saml2AuthenticationRequestRepository} to load authentication
+	 * request.
+	 * @param authenticationRequestRepository the
+	 * {@link Saml2AuthenticationRequestRepository} to use
+	 * @since 5.6
+	 */
+	public void setAuthenticationRequestRepository(
+			Saml2AuthenticationRequestRepository<AbstractSaml2AuthenticationRequest> authenticationRequestRepository) {
+		this.authenticationRequestRepository = authenticationRequestRepository;
```
jzheaux, line 95: Let's please check for null here, using `Assert.notNull`.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverter.java
```diff
@@ -74,7 +79,27 @@ public Saml2AuthenticationToken convert(HttpServletRequest request) {
...
+	}
+
+	private AbstractSaml2AuthenticationRequest loadAuthenticationRequest(HttpServletRequest request) {
+		if (this.authenticationRequestRepository == null) {
```
jzheaux, line 99: Instead of this null check, you might consider following the Null Object pattern, like so:

```java
private Function<HttpServletRequest, AbstractSaml2AuthenticationRequest> loader = (request) -> null;

// ...

public void setAuthenticationRequestRepository(Saml2AuthenticationRequestRepository requests) {
    // ...
    this.loader = requests::loadAuthenticationRequest;
}
```

This improves readability by removing identation and conditionals.

jzheaux (changes requested):
Thanks for the updates, @marcusdacoregio! I've left some additional feedback about the changes you submitted.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
```diff
@@ -339,8 +351,19 @@ private AuthenticationRequestEndpointConfig() {
...
+					authenticationRequestResolver);
+			if (repository != null) {
+				filter.setAuthenticationRequestRepository(repository);
+			}
+			return postProcess(filter);
+		}
+
+		private Saml2AuthenticationRequestRepository<AbstractSaml2AuthenticationRequest> getAuthenticationRequestRepository(
+				B http) {
+			return getSharedOrBean(http, Saml2AuthenticationRequestRepository.class);
```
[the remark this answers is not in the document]
marcusdacoregio: Actually, there isn't. I was just considering that the Filter already initializes an `HttpSessionSaml2AuthenticationRequestRepository` when constructed

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AuthenticationToken.java
marcusdacoregio: Nice!

marcusdacoregio:
Thanks for your effort on the review @jzheaux. I've updated the PR with your suggestions.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverter.java
```diff
@@ -50,6 +54,10 @@
 
 	private final Converter<HttpServletRequest, RelyingPartyRegistration> relyingPartyRegistrationResolver;
 
+	private Saml2AuthenticationRequestRepository<AbstractSaml2AuthenticationRequest> authenticationRequestRepository = new HttpSessionSaml2AuthenticationRequestRepository();
```
jzheaux, line 57: It would be better to remove this reference since it would otherwise need to stay in sync with the `loader` member.

For example, in your constructor you could do:

```java
this.loader = new HttpSessionSaml2AuthenticationRequestFactory()::loadAuthenticationRequest;
```

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
```diff
@@ -33,13 +33,16 @@
 import org.springframework.security.config.annotation.web.configurers.AbstractHttpConfigurer;
```
jzheaux, line 33: Please update this copyright and any others that are not yet updated to 2021.

jzheaux (changes requested):
Thanks, @marcusdacoregio! I've left some additional feedback inline.

marcusdacoregio:
Thanks @jzheaux, I've updated the PR based on your feedback.

docs/manual/src/docs/asciidoc/_includes/servlet/saml2/saml2-login.adoc
```diff
@@ -1610,3 +1610,33 @@ http {
 The success handler will send logout requests to the asserting party.
 
 The request matcher will detect logout requests from the asserting party.
+
+[[servlet-saml2login-store-authn-request]]
+=== Storing the Authentication Request
+
+The `Saml2AuthenticationRequestRepository` is responsible for the persistence of the `AbstractSaml2AuthenticationRequest` from the time the Authentication Request is initiated to the time the Authentication Response is received.
```
jzheaux, line 1617: You might consider linking back to `[[servlet-saml2login-sp-initiated-factory]]` and `[[servlet-saml2login-authenticate-responses]]` to help folks back to the sections that give an overview of how Spring Security handles AuthnRequests and SAML Responses.

jzheaux (changes requested):
Things are really taking shape now, @marcusdacoregio, nice work.

I've left some remaining feedback about JavaDoc, docs, and tests.

saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/servlet/filter/Saml2WebSsoAuthenticationFilterTests.java
```diff
@@ -84,4 +93,65 @@ public void attemptAuthenticationWhenRegistrationIdDoesNotExistThenThrowsExcepti
...
+	@Test
+	public void setAuthenticationRequestRepositoryWhenNullThenThrowsIllegalArgument() {
+		assertThatIllegalArgumentException().isThrownBy(() -> this.filter.setAuthenticationRequestRepository(null))
+				.withMessage("authenticationRequestRepository cannot be null");
+	}
+
+	@Test
+	public void setAuthenticationRequestRepositoryWhenExpectedAuthenticationConverterTypeThenSetLoaderIntoConverter() {
+		Saml2AuthenticationTokenConverter authenticationConverterMock = mock(Saml2AuthenticationTokenConverter.class);
+		MockSaml2AuthenticationRequestRepository mockAuthenticationRequestRepository = new MockSaml2AuthenticationRequestRepository();
```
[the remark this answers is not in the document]
marcusdacoregio: Actually, there isn't. I went for the dedicated mock but turns out that it's the same as a Mockito mock 😃

marcusdacoregio:
Thank you again for the review @jzheaux. I've updated the PR.

About the docs, I've updated it to use the SAML terms, like `AuthnRequest` and `SAMLResponse`, aligned with the rest of the docs as you mentioned.

jzheaux (approved):

jzheaux:

PR #10075, opened 2021-07-14
Avoid using SpEL to change the meaning of the injection point

snicoll:
While I was working with Native, I've noticed that the injection point metadata is changed by a SpEL expression that I found confusing. If the `@Value` does not apply, the regular behaviour of framework is to inject all bean of that types it knows about.

This commit removes the use of SpEL expression and replaces it with an explicit call to the underlying method. This has the effect of not exposing an additional bean. Additionally, we could move that logic in the configuration class itself.


rwinch:
Thanks for the PR @snicoll This is merged into 5.7.0-M1

Issue #4134, opened 2016-11-15
GrantedAuthorityDefaults does not apply on ExpressionInterceptUrlRegistry#hasRole


kazuki43zoo:
### Summary

Since 4.2, we can control the default role prefix(default is `ROLE_`) at the single point of definition (gh-3701).
However, it not apply to access control definitions using the `ExpressionInterceptUrlRegistry#hasRole()`.

```java
@Configuration
public class SecurityConfig extends WebSecurityConfigurerAdapter {

    @Bean
    GrantedAuthorityDefaults grantedAuthorityDefaults() {
        return new GrantedAuthorityDefaults(""); // Remove the ROLE_ prefix
    }

    @Override
    protected void configure(HttpSecurity http) throws Exception {
        // ...
        http.authorizeRequests()
                .anyRequest().hasRole("USER"); // Allow access from user granted USER role
    }

    @Bean
    public UserDetailsService userDetailsService() {
        InMemoryUserDetailsManager userDetailsManager = new InMemoryUserDetailsManager();
        userDetailsManager.createUser(
                User.withUsername("kazuki43zoo")
                        .password("password")
                        .authorities("USER")  // Grant the USER role (without ROLE_ prefix)
                        .build());
        return userDetailsManager;
    }

}
```

### Actual Behavior

If login with the `kazuki43zoo`, 403 Forbidden error has been occurred.

```
Whitelabel Error Page

This application has no explicit mapping for /error, so you are seeing this as a fallback.

Wed Nov 16 01:06:38 JST 2016
There was an unexpected error (type=Forbidden, status=403).
Access is denied
```

### Expected Behavior

Can access the specified resource after authentication success. (200 OK)

### Version

4.2.0

### Workaround

It work as following configuration instead of.

```java
    @Override
    protected void configure(HttpSecurity http) throws Exception {
        // ...
        http.authorizeRequests()
                .anyRequest().access("hasRole('USER')"); // Use the access() instead of hasRole()
    }
```

kazuki43zoo:
Related with this, User Builder#roles append the `ROLE_` prefix to granted authority. Is this behavior working as design ?

```java
    @Bean
    public UserDetailsService userDetailsService() {
        InMemoryUserDetailsManager userDetailsManager = new InMemoryUserDetailsManager();
        userDetailsManager.createUser(
                User.withUsername("kazuki43zoo")
                        .password("password")
                        .roles("USER") // has been append the ROLE_ prefix
[...]

PR #10078, opened 2021-07-15
Replace static "ROLE_" with customized role prefix

quaff:
Fix gh-4134

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -76,6 +76,7 @@ import org.springframework.util.StringUtils;
  *
  * @param <H> the type of {@link HttpSecurityBuilder} that is being configured
  * @author Rob Winch
+ * @author Yanming Zhou
  * @since 3.2
  * @see org.springframework.security.config.annotation.web.builders.HttpSecurity#authorizeRequests()
  */
@@ -94,6 +95,8 @@ public final class ExpressionUrlAuthorizationConfigurer<H extends HttpSecurityBu
 
 	private static final String rememberMe = "rememberMe";
 
+	private final String rolePrefix;
+
 	private final ExpressionInterceptUrlRegistry REGISTRY;
 
 	private SecurityExpressionHandler<FilterInvocation> expressionHandler;
@@ -103,6 +106,15 @@ public final class ExpressionUrlAuthorizationConfigurer<H extends HttpSecurityBu
 	 * @see HttpSecurity#authorizeRequests()
 	 */
 	public ExpressionUrlAuthorizationConfigurer(ApplicationContext context) {
+		String[] grantedAuthorityDefaultsBeanNames = context.getBeanNamesForType(GrantedAuthorityDefaults.class);
+		if (grantedAuthorityDefaultsBeanNames.length == 1) {
+			GrantedAuthorityDefaults grantedAuthorityDefaults = context.getBean(grantedAuthorityDefaultsBeanNames[0],
+					GrantedAuthorityDefaults.class);
+			this.rolePrefix = grantedAuthorityDefaults.getRolePrefix();
+		}
+		else {
+			this.rolePrefix = "ROLE_";
+		}
 		this.REGISTRY = new ExpressionInterceptUrlRegistry(context);
 	}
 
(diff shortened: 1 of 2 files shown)


quaff:
@rwinch Ready for review.

quaff:
Would any team member take a look at this? It's a simple fix.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
```diff
@@ -21,6 +21,7 @@
 import java.util.LinkedHashMap;
```
jzheaux, line 21: Please update the copyright year for the files you modify.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
```diff
@@ -204,8 +205,15 @@ private static String hasIpAddress(String ipAddressExpression) {
 	public final class ExpressionInterceptUrlRegistry extends
 			ExpressionUrlAuthorizationConfigurer<H>.AbstractInterceptUrlRegistry<ExpressionInterceptUrlRegistry, AuthorizedUrl> {
 
+		private String rolePrefix = "ROLE_";
+
 		private ExpressionInterceptUrlRegistry(ApplicationContext context) {
 			setApplicationContext(context);
+			try {
+				this.rolePrefix = context.getBean(GrantedAuthorityDefaults.class).getRolePrefix();
```
jzheaux, line 213: Instead of catching an exception, which can be costly, please call `ApplicationContext#getBeanNamesForType`. This will return an empty array if there is no `GrantedAuthorityDefaults` bean.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
```diff
@@ -204,8 +205,15 @@ private static String hasIpAddress(String ipAddressExpression) {
 	public final class ExpressionInterceptUrlRegistry extends
 			ExpressionUrlAuthorizationConfigurer<H>.AbstractInterceptUrlRegistry<ExpressionInterceptUrlRegistry, AuthorizedUrl> {
 
+		private String rolePrefix = "ROLE_";
```
jzheaux, line 208: Please make this `final`.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
```diff
@@ -127,7 +129,6 @@ private void interceptUrl(Iterable<? extends RequestMatcher> requestMatchers,
 	}
 
 	@Override
-	@SuppressWarnings("rawtypes")
```
jzheaux, line 130: Thank you for the cleanups as well -- will you please place these in a separate commit so that each commit is about just one thing?

jzheaux (changes requested):
Thanks, @quaff! Please see my feedback inline.

quaff:
@jzheaux updated, please review.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
```diff
@@ -176,16 +177,16 @@ ExpressionBasedFilterInvocationSecurityMetadataSource createMetadataSource(H htt
...
+		return "hasAnyRole('" + rolePrefix + anyAuthorities + "')";
 	}
 
-	private static String hasRole(String role) {
+	private static String hasRole(String rolePrefix, String role) {
 		Assert.notNull(role, "role cannot be null");
-		Assert.isTrue(!role.startsWith("ROLE_"),
-				() -> "role should not start with 'ROLE_' since it is automatically inserted. Got '" + role + "'");
-		return "hasRole('ROLE_" + role + "')";
+		Assert.isTrue(rolePrefix.isEmpty() || !role.startsWith(rolePrefix), () -> "role should not start with '"
```
juncevich, line 187: I saw you had changed the old code, but maybe it will be better to extract asserts to another method, e.x. validate or check role? It will be better for readability.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
```diff
@@ -308,27 +320,30 @@ public AuthorizedUrl not() {
...
 		 * @param role the role to require (i.e. USER, ADMIN, etc). Note, it should not
-		 * start with "ROLE_" as this is automatically inserted.
+		 * start with role prefix as this is automatically inserted.
 		 * @return the {@link ExpressionUrlAuthorizationConfigurer} for further
 		 * customization
 		 */
 		public ExpressionInterceptUrlRegistry hasRole(String role) {
-			return access(ExpressionUrlAuthorizationConfigurer.hasRole(role));
+			return access(ExpressionUrlAuthorizationConfigurer
+					.hasRole(ExpressionUrlAuthorizationConfigurer.this.REGISTRY.rolePrefix, role));
```
juncevich, line 332: Suggest extracting `ExpressionUrlAuthorizationConfigurer.this.REGISTRY.rolePrefix` to constant or method.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
quaff: I would prefer keeping this since it's internal here.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExpressionUrlAuthorizationConfigurer.java
quaff: It only appear 2 times, I'm not sure it should extract as
```java
private final String rolePrefix = ExpressionUrlAuthorizationConfigurer.this.REGISTRY.rolePrefix;
```
leave it to @jzheaux to decide.

jzheaux: I think what I would do is move the `rolePrefix` to be a member of `ExpressionUrlAuthorizationConfigurer` and the `ApplicationContext` support to `ExpressionUrlAuthorizationConfigurer`'s constructor.

The registry doesn't seem to have any need on its own for a role prefix, so it's best to leave it out.

Then, I'd change the references to `ExpressionUrlAuthorizationConfigurer.this.rolePrefix` instead of `ExpressionUrlAuthorizationConfigurer.this.REGISTRY.rolePrefix`.

quaff: @jzheaux Updated.

jzheaux (approved):

jzheaux:
Thanks, @quaff! This is now merged into `main`.

PR #10085, opened 2021-07-16
Allow extending the authentication provider

yuezk:
I want to extend the authentication provider classes to add some extra logic, but I couldn't because they are final classes.

Seems that there are no direct reasons for making them `final`. So can we remove the `final` keyword?


jzheaux:
Thank you, @yuezk, for taking the time to make this PR. We intentionally leave classes as `final` until there is a concrete reason to open them up. This simplifies maintenance long-term and also favors composition over inheritance.

For additional security reasons, please take a look at [the relevant Oracle Security Coding Guideline](https://www.oracle.com/java/technologies/javase/seccodeguide.html#4-5).

If there is a concrete use case that you can't resolve with composition, please feel free to ask a question on [StackOverflow](https://stackoverflow.com) and add the link here so that others can find it - I'd be happy to take a look at your question and see how what can be done.

yuezk:
@jzheaux I see. We are using composition in our project and it works fine. Thanks.

PR #10129, opened 2021-07-20
Fixes the Javadoc of ClaimAccessor methods for Map/List

### LINKED ISSUE #10117: ClaimAccessor#getClaimAsMap doesn't return null as documented
**Describe the bug**
`org.springframework.security.oauth2.core.ClaimAccessor#getClaimAsMap`'s Javadoc reads as follows:
```
	 * @return the claim value or {@code null} if it does not exist or cannot be assigned
	 * to a {@code Map}
```

The actual implementation throws an `IllegalArgumentException` instead, if the claim value cannot be converted to a map:

```java
Assert.isTrue(convertedValue != null,
  () -> "Unable to convert claim '" + claim + "' of type '" + claimValue.getClass() + "' to Map.");
```

**To Reproduce**
Use a claim value such as:

```
{
   "claim": "string_value"
}
```

Then call `getClaimAsMap("claim")` and observe that the following exception will be thrown:

```java
java.lang.IllegalArgumentException: Unable to convert claim 'claim' of type 'class java.lang.String' to Map.
```

**Expected behavior**
The Javadoc should match the actual implementation, or vice versa.

shartte:
Fixes #10117 by correcting the Javadoc to match the implementation behavior.

### DIFF
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
@@ -136,8 +136,9 @@ public interface ClaimAccessor {
 	 * Returns the claim value as a {@code Map<String, Object>} or {@code null} if it does
 	 * not exist or cannot be assigned to a {@code Map}.
 	 * @param claim the name of the claim
-	 * @return the claim value or {@code null} if it does not exist or cannot be assigned
-	 * to a {@code Map}
+	 * @return the claim value or {@code null} if it does not exist
+	 * @throws IllegalArgumentException If the claim exists, but cannot be converted to a
+	 * {@code Map}
 	 */
 	@SuppressWarnings("unchecked")
 	default Map<String, Object> getClaimAsMap(String claim) {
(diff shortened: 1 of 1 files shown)


shartte:

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
```diff
@@ -136,8 +136,9 @@ default URL getClaimAsURL(String claim) {
 	 * Returns the claim value as a {@code Map<String, Object>} or {@code null} if it does
 	 * not exist or cannot be assigned to a {@code Map}.
 	 * @param claim the name of the claim
-	 * @return the claim value or {@code null} if it does not exist or cannot be assigned
-	 * to a {@code Map}
+	 * @return the claim value or {@code null} if it does not exist
+	 * @throws IllegalArgumentException If the claim exists, but cannot be converted to a
```
jzheaux, line 140: For consistency with the rest of the Javadoc, will you please leave "If" lowercased in both of your changes?

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
```diff
@@ -136,8 +136,9 @@ default URL getClaimAsURL(String claim) {
 	 * Returns the claim value as a {@code Map<String, Object>} or {@code null} if it does
```
jzheaux, line 136: Please update the copyright messages to go to "2021"

jzheaux (changes requested):
Thanks for the PR, @shartte! I've left some feedback inline.

Also, in preparation for merging will you please squash your commits and format them according to [the contributing guidelines](https://github.com/spring-projects/spring-security/blob/main/CONTRIBUTING.adoc#format-commit-messages)?

shartte:
I am sorry, but this is turning out to be more busywork than I had anticipated to update two lines of javadoc.
I'll leave this for someone else.

jzheaux:
> I am sorry, but this is turning out to be more busywork than I had anticipated

I can see how for a small change some of these guidelines may seem like a lot to do. I'm sorry if I made it seem like busywork. Holding to these contribution guidelines plays an important role in keeping the code quality high - it was not my intent to just give you meaningless work to keep you busy.

Thank you for taking the time to contribute, and I hope to see you again down the road.

Issue #10130, opened 2021-07-20
Support customizing headers of a request in AbstractWebClientReactiveOAuth2AccessTokenResponseClient


sjohnr:
**Expected Behavior**

Headers added to the request via `.headers((headers) -> populateTokenRequestHeaders(grantRequest, headers))` of the `WebClient` should be customizable so the application can add additional headers or customize the default set of headers provided by `AbstractWebClientReactiveOAuth2AccessTokenResponseClient`.

**Current Behavior**

Currently, the request can only be customized by providing a custom `WebClient` and registering an `ExchangeFilterFunction`. This limits the application to only the current `ClientRequest`, possibly requiring parsing and mutating the request after it has already been built, with the inability to access the current `AbstractOAuth2AuthorizationGrantRequest`.

**Context**

See gh-10042

sjohnr:
@vboulaye, I looked at `AbstractWebClientReactiveOAuth2AccessTokenResponseClient` while opening up this ticket. So far, it looks like the only customization points are package-private methods that can be overridden by a sub-class. However, the implementation on the servlet side uses the `Converter` interface, with a default implementation. For example, given `<T extends AbstractOAuth2AuthorizationGrantRequest>`:

``` java
	private Converter<T, HttpHeaders> headersConverter = this::populateTokenRequestHeaders;
```

Similar to the existing method but now returning headers:

``` java
	private HttpHeaders populateTokenRequestHeaders(T grantRequest) { ... }
```

And provide a nicely documented setter. I think something like this could be the way to go.

PR #10131, opened 2021-07-20
enable customization of headers in AbstractWebClientReactiveOAuth2AccessTokenResponseClient

vboulaye:
Hi,

This is an attempt at fixing #10130 

I tried adding an headers consumer into the `AbstractWebClientReactiveOAuth2AccessTokenResponseClient` to customize the headers after they are set by `populateTokenRequestHeaders(grantRequest, headers)`
Is this the proper way to do it ?

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/AbstractWebClientReactiveOAuth2AccessTokenResponseClient.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/AbstractWebClientReactiveOAuth2AccessTokenResponseClient.java
@@ -24,6 +24,7 @@ import java.util.Set;
 
 import reactor.core.publisher.Mono;
 
+import org.springframework.core.convert.converter.Converter;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.MediaType;
 import org.springframework.security.oauth2.client.registration.ClientRegistration;
@@ -65,6 +66,8 @@ public abstract class AbstractWebClientReactiveOAuth2AccessTokenResponseClient<T
 
 	private WebClient webClient = WebClient.builder().build();
 
+	private Converter<T, HttpHeaders> headersConverter = this::populateTokenRequestHeaders;
+
 	AbstractWebClientReactiveOAuth2AccessTokenResponseClient() {
 	}
 
@@ -74,7 +77,12 @@ public abstract class AbstractWebClientReactiveOAuth2AccessTokenResponseClient<T
 		// @formatter:off
 		return Mono.defer(() -> this.webClient.post()
 				.uri(clientRegistration(grantRequest).getProviderDetails().getTokenUri())
-				.headers((headers) -> populateTokenRequestHeaders(grantRequest, headers))
+				.headers((headers) -> {
+					HttpHeaders headersToAdd = getHeadersConverter().convert(grantRequest);
+					if (headersToAdd != null) {
+						headers.addAll(headersToAdd);
+					}
+				})
 				.body(createTokenRequestBody(grantRequest))
 				.exchange()
 				.flatMap((response) -> readTokenResponse(grantRequest, response))
@@ -92,9 +100,10 @@ public abstract class AbstractWebClientReactiveOAuth2AccessTokenResponseClient<T
 	/**
 	 * Populates the headers for the token request.
 	 * @param grantRequest the grant request
-	 * @param headers the headers to populate
+	 * @return the headers populated for the token request
 	 */
-	private void populateTokenRequestHeaders(T grantRequest, HttpHeaders headers) {
+	private HttpHeaders populateTokenRequestHeaders(T grantRequest) {
+		HttpHeaders headers = new HttpHeaders();
 		ClientRegistration clientRegistration = clientRegistration(grantRequest);
 		headers.setContentType(MediaType.APPLICATION_FORM_URLENCODED);
 		headers.setAccept(Collections.singletonList(MediaType.APPLICATION_JSON));
@@ -104,6 +113,7 @@ public abstract class AbstractWebClientReactiveOAuth2AccessTokenResponseClient<T
 			String clientSecret = encodeClientCredential(clientRegistration.getClientSecret());
 			headers.setBasicAuth(clientId, clientSecret);
 		}
+		return headers;
 	}
 
 	private static String encodeClientCredential(String clientCredential) {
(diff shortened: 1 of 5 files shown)


sjohnr:
Hi @vboulaye! See [this comment](https://github.com/spring-projects/spring-security/issues/10130#issuecomment-883712855) for a starting point. Let me know if you have any more questions.

vboulaye:
Hi, I added the header converter + some unit tests for each of the AbstractWebClientReactiveOAuth2AccessTokenResponseClient variants.
let me know if something is missing.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/endpoint/WebClientReactiveAuthorizationCodeTokenResponseClientTests.java
```diff
@@ -340,4 +345,64 @@ private OAuth2AuthorizationCodeGrantRequest pkceAuthorizationCodeGrantRequest()
...
+				+ "   \"scope\": \"openid profile\"\n"
+				+ "}\n";
+		// @formatter:on
+		this.server.enqueue(jsonResponse(accessTokenSuccessResponse));
+		this.tokenResponseClient.getTokenResponse(request).block();
+		InOrder inOrder = inOrder(headersConverter1, headersConverter2);
+		inOrder.verify(headersConverter1).convert(request);
+		inOrder.verify(headersConverter2).convert(request);
+		RecordedRequest actualRequest = this.server.takeRequest();
+		assertThat(actualRequest.getHeader(HttpHeaders.AUTHORIZATION)).isEqualTo("Basic CUSTOM");
```
sjohnr, line 405: I'm thinking it would be better to test for the ability to customize the `Authorization` header in the above test which is focused on `addHeadersConverter`, and let this test focus on the original basic credentials, e.g. flipping the mock and assertions between the two tests. If you agree, this comment would apply to all test classes.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/endpoint/WebClientReactiveAuthorizationCodeTokenResponseClientTests.java
```diff
@@ -340,4 +345,64 @@ private OAuth2AuthorizationCodeGrantRequest pkceAuthorizationCodeGrantRequest()
...
+
+	@Test
+	public void addHeadersConverterWhenNullThenThrowIllegalArgumentException() {
+		assertThatIllegalArgumentException().isThrownBy(() -> this.tokenResponseClient.addHeadersConverter(null))
+				.withMessage("headersConverter cannot be null");
+	}
+
+	@Test
+	public void convertWhenHeadersConverterAddThenCalled() throws InterruptedException {
+		Converter<OAuth2AuthorizationCodeGrantRequest, HttpHeaders> headersConverter1 = mock(Converter.class);
```
sjohnr, line 362: I recommend renaming to `addedHeadersConverter`.

sjohnr (changes requested):
Hi @vboulaye, my apologies for the delay. Thanks for the updates. I have a few more suggestions below.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/AbstractWebClientReactiveOAuth2AccessTokenResponseClient.java
```diff
@@ -74,7 +77,7 @@
 		// @formatter:off
 		return Mono.defer(() -> this.webClient.post()
 				.uri(clientRegistration(grantRequest).getProviderDetails().getTokenUri())
-				.headers((headers) -> populateTokenRequestHeaders(grantRequest, headers))
+				.headers((headers) -> headers.addAll(headersConverter.convert(grantRequest)))
```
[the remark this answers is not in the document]
vboulaye: regarding `getHeadersConverter()`shouldn't we also open its access to `protected` to make the override it easier ?

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/endpoint/WebClientReactiveAuthorizationCodeTokenResponseClientTests.java
vboulaye: I tried to do this in `WebClientReactiveAuthorizationCodeTokenResponseClientTests`, but in the first test I had to change the header otherwise it adds 2 values to Authorization.
I wonder if it would'nt be better to use the `setHeadersConverter()` calls here ?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/AbstractWebClientReactiveOAuth2AccessTokenResponseClient.java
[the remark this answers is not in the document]
sjohnr: That is a fair point, but it's not done in `AbstractOAuth2AuthorizationGrantRequestEntityConverter`. I'll have to check on why that is. For now, let's leave it package-private but you have a point, it would be difficult to override outside this package.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/endpoint/WebClientReactiveAuthorizationCodeTokenResponseClientTests.java
sjohnr: Ah, you are right, we don't want to test modifying credentials with `addHeadersConverter`. So perhaps just remove the `headersConverter2` in this test and rename `headersConverter1` to `headersConverter`.

vboulaye: ok, I removed the headersConverter2 on the `setHeadersConverter` test, and added the mock in all the `addHeadersConverter` tests. this way we check that both replacing or adding headers work.

vboulaye:
Hi,
I think I handled all the review comments.
Except if you want me to change the visibility of the `getHeadersConverter()` this should be ready to be merged.

sjohnr:
Thanks @vboulaye! This is now in main.

PR #10133, opened 2021-07-21
Propagate original exception for "Malformed Jwk set"

ttddyy:
In our log, we found there are a couple of entries like this:

```
org.springframework.security.oauth2.jwt.JwtException: An error occurred while attempting to decode the Jwt: Malformed Jwk set

	at org.springframework.security.oauth2.jwt.NimbusJwtDecoder.createJwt(NimbusJwtDecoder.java:166)
```

Since [NimbusJwtDecoder](https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java#L167) does not propagate the original exception, it is hard to figure out the real reason why this exception is thrown.

This PR simply adds the original exception to the new throwing exception as its cause.

I understand there is a trace level entry that logs the original `ex`, but would be nice to include it in the throwing exception for analysis.


marcusdacoregio:
Hi @ttddyy, thank you very much for your contribution.

I'll close this because it has been fixed in commit https://github.com/spring-projects/spring-security/commit/f1691370d6ed756910e29a989275c88ebd748654

marcusdacoregio:
Fixed via https://github.com/spring-projects/spring-security/commit/f1691370d6ed756910e29a989275c88ebd748654

Issue #10077, opened 2021-07-14
Saml2LoginConfigure Null pointer exception due to Java module introduction


AbdulR3hman:
**Describe the bug**

The issue is a Java null pointer exception resulted from checking for OpenSAML version within the SAML2 login configurer when used with Java Modules rather than Classpath.

Exception:

```Java
Caused by: java.lang.NullPointerException: null
	at spring.security.config@5.5.1/org.springframework.security.config.annotation.web.configurers.saml2.Saml2LoginConfigurer$AuthenticationRequestEndpointConfig.getResolver(Saml2LoginConfigurer.java:349) ~[spring-security-config-5.5.1.jar:na]
	at spring.security.config@5.5.1/org.springframework.security.config.annotation.web.configurers.saml2.Saml2LoginConfigurer$AuthenticationRequestEndpointConfig.build(Saml2LoginConfigurer.java:340) ~[spring-security-config-5.5.1.jar:na]
	at 
``` 

This is due to the code that doesn't check if the nullable Version class is null:

https://github.com/spring-projects/spring-security/blob/bff377904a60ce2ed97b382be5c057d2ed6dd683/config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java#L346-L355

This was introduced by this issue: https://github.com/spring-projects/spring-security/issues/9095

**To Reproduce**

Simply import saml2 provider: (with spring-boot version: 2.5.2)

```xml
        <dependency>
            <groupId>org.springframework.security</groupId>
            <artifactId>spring-security-saml2-service-provider</artifactId>
        </dependency
```

And build with maven without `module-info.java` file, which means it will be using the classpath instead. This will work, however migrating to module will make the Version of Opensaml completely null resulting in the above exception.

**Expected behavior**

Check for null Version class as it is allowed to be nullable `org.opensaml.core.Version` .

jzheaux:
Thanks for the report, @AbdulR3hman.

It won't fix the problem to check for null since Spring Security must know the version of OpenSAML to select the correct implementation.

Perhaps the configurer could look for a class that is present in OpenSAML 4 that does not exist in OpenSAML 3. Are you able to submit a PR that replaces `Version.getVersion().startsWith("4")` with something like `ClassUtils.isPresent("some.opensaml4.Class", Version.class.getClassLoader())`?

AbdulR3hman:
[...]

PR #10134, opened 2021-07-21
OpenSAML Version null-pointer exception check

AbdulR3hman:
This will enable the app to check for the version using both classpath and module-path (java 9+) when using modules. 

This issue has been discussed in this ticket: https://github.com/spring-projects/spring-security/issues/10077 and followed recommendation of @jzheaux on how to handle this.


jzheaux:
Thanks for the PR, @AbdulR3hman! The build appears to be failing - are you able to fix that and update your PR?

Also, for any updates, please squash your commits and then [format your commit message](https://github.com/spring-projects/spring-security/blob/main/CONTRIBUTING.adoc#format-commit-messages) like so:

```
Improve OpenSAML Version Check

Closes gh-10077
```

Using this format works better with our automation.

Issue #10135, opened 2021-07-21
MappedJwtClaimSetConverter#withDefaults doesn't remove claims from JWT as documented


fguenci:
**Describe the bug**
As stated in [Spring Security Documentation](https://docs.spring.io/spring-security/site/docs/current/reference/html5/#oauth2resourceserver-jwt-claimsetmapping-remove), to remove a claim from a JWT just pass a converter for the claim in `MappedJwtClaimSetConverter.withDefaults()` that return null.
Actually this setting doesn't remove the claim from JWT.

I think the problem is in class `org.springframework.security.oauth2.core.converter.ClaimTypeConverter`, method `convert`.
```java
	public Map<String, Object> convert(Map<String, Object> claims) {
		if (CollectionUtils.isEmpty(claims)) {
			return claims;
		}

		Map<String, Object> result = new HashMap<>(claims);
		this.claimTypeConverters.forEach((claimName, typeConverter) -> {
			if (claims.containsKey(claimName)) {
				Object claim = claims.get(claimName);
				Object mappedClaim = typeConverter.convert(claim);
				if (mappedClaim != null) {
					result.put(claimName, mappedClaim);
				}
			}
		});

		return result;
	}
```
I think that the `result` map should contain all the mapped claims even if its value is null because null value claims are removed later. 

**To Reproduce**
Steps to reproduce:
I want to remove the NBF claim from jwt. 
To do that I set in my jwtdecoder a converter that return null for this claim:

```java
		var jwtDecoder = NimbusJwtDecoder.withJwkSetUri(jwkSetUri).build();
		var converter = MappedJwtClaimSetConverter.withDefaults(Collections.singletonMap(JwtClaimNames.NBF, nbfClaimValue -> null));
		jwtDecoder.setClaimSetConverter(converter);
```

**Expected behavior**
The decoded JWT doen't contains the NBF claim.

jzheaux:
I agree, @fguenci. Are you able to submit a PR with the change, including a test in `MappedJwtClaimSetConverterTests` that matches this use case?

fguenci:
ok, i'll try

PR #10142, opened 2021-07-27
MappedJwtClaimSetConverter#withDefaults doesn't remove claims from JWT as documented

fguenci:
Previously, ClaimTypeConverter doesn't overwrite claim's value converted to null.

This commit allows ClaimTypeConverter to overwrite and return claim with
value converted to null.

Closes https://github.com/spring-projects/spring-security/issues/10135

### DIFF
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/converter/ClaimTypeConverter.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/converter/ClaimTypeConverter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverterTests.java
```diff
@@ -125,6 +125,15 @@ public void convertWhenUsingCustomConverterThenAllOtherDefaultsAreStillUsed() {
 
 	@Test
 	public void convertWhenConverterReturnsNullThenClaimIsRemoved() {
```
jzheaux, line 127: Will you please add to the beginning of this test a comment tying it to the related GitHub issue?

For example:

```
// gh-10135
@Test 
public ...
```

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/converter/ClaimTypeConverter.java
```diff
@@ -56,10 +56,7 @@ public ClaimTypeConverter(Map<String, Converter<Object, ?>> claimTypeConverters)
 		this.claimTypeConverters.forEach((claimName, typeConverter) -> {
```
jzheaux, line 56: Will you please update these files to use a copyright end date of 2021?

jzheaux (changes requested):
Thanks, @fguenci, this looks great! I've left some feedback inline to help with overall consistency with the rest of the codebase.

Also, in preparation for merging, will you please squash commits and format the final commit like so:

```
Title ~50 Characters or Less

Any additional description with a  width of ~70 characters

Closes gh-10135
```

maybe something like this:

```
Preserve Null Claim Values

Closes gh-10135
```

If you like, you can find more details in the [contributing guidelines](https://github.com/spring-projects/spring-security/blob/main/CONTRIBUTING.adoc#format-commit-messages).

Issue #9647, opened 2021-04-16
Move and rename OAuth2IntrospectionClaimAccessor/Names


jgrandja:
We should move and rename `OAuth2IntrospectionClaimAccessor` and `OAuth2IntrospectionClaimNames` to allow for reuse.

Currently, the classes reside in `org.springframework.security.oauth2.server.resource` and are used by `oauth2-resource-server` for introspecting opaque access tokens. This is a "specialized flow" used by an OAuth 2.0 Client, since authentication is required at the Introspection Endpoint using client credentials or a bearer token.

Moving both classes to `org.springframework.security.oauth2.core` (`oauth2-core` module) will allow for reuse on the Authorization Server side [spring-authorization-server#52](https://github.com/spring-projects-experimental/spring-authorization-server/issues/52).

As well, renaming `OAuth2IntrospectionClaimAccessor` to `OAuth2TokenIntrospectionClaimAccessor` and `OAuth2IntrospectionClaimNames` to `OAuth2TokenIntrospectionClaimNames` is explicit on "what is being introspected" (a token) and aligns with the spec language.

**NOTE:**

This is scheduled for `5.6.x` with the following plan:
1) Add `OAuth2TokenIntrospectionClaimAccessor` and `OAuth2TokenIntrospectionClaimNames`
2) `@Deprecate` `OAuth2IntrospectionClaimAccessor` and `OAuth2IntrospectionClaimNames`

qavid:
@jgrandja @jzheaux I have prepared PR #9903 for this issue. Please take a look at it.

PR #10151, opened 2021-08-01
getClaimAsBoolean() should not be falsy

### LINKED ISSUE #10148: getClaimAsBoolean should not be falsy
Related to https://github.com/spring-projects/spring-security/issues/10117#issuecomment-883638049

`ClaimAccessor#getClaimAsBoolean` currently coerces any type into a `Boolean`, which is somewhat surprising when it is a map or list.

For example, the following snippet will somewhat surprisingly pass:

```java
ClaimAccessor claims = () -> Collections.singletonMap("claim", new HashMap<>());
assertThat(claims.getClaimAsBoolean("claim")).isFalse();
```

In reality, the above is a usage error. It would be better for the application to complain so that the developer can adjust their system.

`getClaimAsBoolean` should match core Java behavior more closely. It should only coerce booleans and `String`s (like `"true"` and `"FALSE"`) into `Boolean`s.

The logic to change is in `ObjectToBooleanConverter` where it does:

```java
return Boolean.valueOf(source.toString());
```

This should instead do something like the following:

```java
if (source instanceof String) {
    return Boolean.valueOf((String) source);
}

return null;
```

Then, `getClaimAsBoolean` should introduce an assertion, similar to the assertions in the other `getClaimAsXXX` methods:

```java
Object claimValue = getClaims().get(claim);
Boolean convertedValue = ClaimConversionService.getSharedInstance().convert(claimValue, Boolean.class);
Assert.isNotNull(convertedValue, () -> "Unable to convert claim '" + claim + "' of type '" + claimValue.getClass() + "' to Boolean.");
return convertedValue;
```

Thereafter, if an application really needs the old behavior, it can register a custom converter like so:

```java
ClaimConversionService.getInstance().addConverter(new MyObjectToFalsyBooleanConverter());
```

during application startup.

ahmedmq:
Closes #10148


oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
```diff
@@ -94,10 +94,18 @@ default String getClaimAsString(String claim) {
 	 * Returns the claim value as a {@code Boolean} or {@code null} if it does not exist.
 	 * @param claim the name of the claim
 	 * @return the claim value or {@code null} if it does not exist
+	 * @throws IllegalArgumentException if the claim value cannot be converted to a
+	 * {@code Boolean}
```
sjohnr, line 98: For consistency, I think you should drop the `@throws`, as the `IllegalArgumentException` is self explanatory when it occurs.

sjohnr (commented):
Thanks for the submission @ahmedmq! One minor comment is below, and you can squash your changes and force push to the same branch.

ahmedmq:
@sjohnr - Not sure what has gone wrong, it ran all tests successfully before pushing

sjohnr:
@ahmedmq, you'll need to rebase on `main` and check OAuth2TokenIntrospectionClaimAccessorTests.

```
OAuth2TokenIntrospectionClaimAccessorTests > isActiveWhenActiveClaimValueIsNullThenReturnFalse() FAILED
    java.lang.NullPointerException at OAuth2TokenIntrospectionClaimAccessorTests.java:56
```

Quite a bit has changed related to gh-9647 and gh-10135.

sjohnr:
@ahmedmq, are you able to rebase on `main`? Let me know if you have trouble with this step.

ahmedmq:
@sjohnr - I am not sure of the expected behaviour when the claimValue is null. For e.g in the test case `isActiveWhenActiveClaimValueIsNullThenReturnFalse()` under `OAuth2TokenIntrospectionClaimAccessorTests` we have

`this.claims.put(OAuth2TokenIntrospectionClaimNames.ACTIVE, null);`

Since in the above the claimValue is null, what should the boolean converter return? Should it return false by default

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ClaimAccessor.java
```diff
@@ -96,8 +96,14 @@ default String getClaimAsString(String claim) {
...
 	 */
 	default Boolean getClaimAsBoolean(String claim) {
-		return !hasClaim(claim) ? null
-				: ClaimConversionService.getSharedInstance().convert(getClaims().get(claim), Boolean.class);
+		if (!hasClaim(claim)) {
+			return null;
+		}
+		Object claimValue = getClaims().get(claim);
+		Boolean convertedValue = ClaimConversionService.getSharedInstance().convert(claimValue, Boolean.class);
+		Assert.isTrue(convertedValue != null,
```
sjohnr, line 104: Can you use `Assert.isNotNull` here? Looks like other checks in this class don't do that, however, which is fine for now.

sjohnr (changes requested):
@ahmedmq, thanks for updating the PR. The implementation of the converter looks correct with your changes applied based on #10148. Meaning, it would now return an `IllegalArgumentException`, so I would expect the test you mentioned to need to change. This change will break passivity, and so will go in the 5.6 release.

Also, please see below feedback inline.

Issue #6511, opened 2019-02-07
CsrfConfigurer accessDeniedHandler logic


pbialonc:
### Summary
Is it by design that:
CsrfConfigurer does not reuse accessDeniedHandler logic from ExceptionHandlingConfigurer?

### Actual Behavior

CsrfConfigurer is using getAccessDeniedHandler to get accessDeniedHandler and creates its own if no handler specified.
In case we've specified only defaultDeniedHandlerMappings in ExceptionHandlingConfigurer there is RequestMatcherDelegatingAccessDeniedHandler being created which is not set as accessDeniedHandler but is only forwarded to exceptionTranslationFilter.
In this case CsrfConfigurer creates it's own accessDeniedHandler which does not include defaultDeniedHandlerMappings logic.

### Expected Behavior

CsrfConfigurer reuses accessDeniedHandler and defaultDeniedHandlerMappings created in ExceptionHandlingConfigurer

### Version

5.1.3

rwinch:
It appears like it should be using the default access denied handler. If you are having a problem, please provide a complete and minimal sample to reproduce the issue.

pbialonc:
Here is minimal project reproducing the issue:
https://github.com/pbialonc/csrf

CsrfApplicationTests fails on shouldReturnTeapot as only accessDeniedHandler logic is applied for csrf configurer and defaultAccessDeniedHandlerFor does not take any effect.

This is not an issue for me as I could use global accessDeniedHandler or build custom handler, but I think this is an issue in general.

clevertension:
![image](https://user-images.githubusercontent.com/346723/53680343-32f47580-3d15-11e9-82f8-b753e0fb270b.png)

i think it is resonable that if we `setAccessDeniedHandler` explicitly, the `defaultAccessDeniedHandler` will be invalid, that is the mean of `default` or exactly be called `default for temporarily`

spring-issuemaster:
If you would like us to look at this issue, please provide the requested information. If the information is not provided within the next 7 days this issue will be closed.

jzheaux:
Pardon our dust here as we do some issue cleanup. Feedback was already provided earlier, and I don't think the ticket has been fully addressed, yet, so let's keep the issue open.

christophejan:
I agree that when `accessDeniedHandler` is set explicitly, everything work fine.

But as @pbialonc said:
[...]

PR #10154, opened 2021-08-03
Fix CsrfConfigurer default AccessDeniedHandler consistency

christophejan:
Fix when AccessDeniedHandler is specified per RequestMatcher on
ExceptionHandlingConfigurer.

This introduces evolutions on :
- `CsrfConfigurer#getDefaultAccessDeniedHandler`,
to retrieve an AccessDeniedHandler similar to the one used by
ExceptionHandlingConfigurer.
- `OAuth2ResourceServerConfigurer#accessDeniedHandler`, to continue to
handle CsrfException with the default AccessDeniedHandler implementation

Fixes: gh-6511

I have submitted the CLA

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/CsrfConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/CsrfConfigurer.java
@@ -237,8 +237,8 @@ public final class CsrfConfigurer<H extends HttpSecurityBuilder<H>>
 
 	/**
 	 * Gets the default {@link AccessDeniedHandler} from the
-	 * {@link ExceptionHandlingConfigurer#getAccessDeniedHandler()} or create a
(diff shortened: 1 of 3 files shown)


christophejan:
This is just a minimal fix for [CsrfConfigurer accessDeniedHandler logic #6511](https://github.com/spring-projects/spring-security/issues/6511)

The following points bothered me a little (I would like to know if you think they are major drawbacks) :
1. I’m not sure that calling a method of another configurer during configure phase is a good practice (was already the case before)
1. CsrfConfigurer will use an AccessDeniedHandler similar (same behavior) to the one used by ExceptionHandlingConfigurer but not the same instance of AccessDeniedHandler.

Best regards

rwinch:
Thanks for the PR This is merged via 4318a519711802c1cf06779670ec890d900fe9e4

PR #10173, opened 2021-08-13
Add cause to opaque token introspection failures

ascopes:
This adds the cause of opaque OAuth2 introspection failures to the exception thrown by Spring Security, which aids in debugging authentication issues using OAuth2.


ascopes:
Discovered some other issues in similar locations, will put up a single PR to address each of these.

PR #10175, opened 2021-08-15
Bugfix/oauth2 issuer treatment and exception handling

ascopes:
----
Bug fixes:

Fixes to OAuth2 introspection processing, fixing a couple of misinterpretations of
the implementation spec, and amending logging to aid in debugging.

-   Amended treatment of OAuth2 'iss' claim
    
    Prior to this commit, the OAuth2 resource server code is failing any issuer
    that is not a valid URL. This does not correspond to
    https://datatracker.ietf.org/doc/html/rfc7662#page-7 which redirects to
    https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.1, defining an
    issuer as being a "StringOrURI", which is defined at
    https://datatracker.ietf.org/doc/html/rfc7519#page-5 as being
    an "arbitrary string value" that "MUST be a URI" only for
    "any value containing a ':'". 

    While OIDC does require this field to be a valid URL, the use of the resource
    server itself should not require all integrations to be using OIDC under the
    hood, since that is irrelevant implementation detail as far as a REST API
    using OAuth2 resource server introspection for authentication and authorization
    is concerned.
    
    The issue currently is that an issuer that is not a valid URL may be
    provided, which will automatically result in the request being aborted
    due to being invalid.
    
    I have removed the check entirely, since while the claim could be invalid,
    it is still a response that the OAuth2 introspection endpoint has provided.
    In the liklihood that interpretations of this behaviour are different for
    the OAuth2 server implementation in use, this currently stops Spring
    Security from being able to be used at all without implementing a custom
    introspector from scratch.
    
    It is also worth noting that the spec does not specify whether it is
    valid to normalize issuers or not if they are valid URLs. This may cause
    other unintended side effects as a result of this change, so it is
    safer to disable it entirely. Best practises such as those outlined at
    https://curity.io/resources/learn/jwt-best-practices/#4-always-check-the-issuer always suggest
    that issuers must "match _exactly_ the value you expect it to be" which
    is not necessarily guaranteed by converting to java.net.URL (since there is no
    concrete specification to say that the URL class must conform to JWT semantics).

    

    **This issue is currently causing issues in existing applications we have that use this library**, 
    which is what has sparked the change to update this. The change itself may be
    considered breaking if the documentation refers to the issuer as being stored
    as a URL object. However, if the documentation only implies that .toString can be performed,
    then the underlying logic of this change would likely just be considered implementation
    detail instead.

----
Minor fixes and improvements:

-   Fixed potential NullPointerException in opaque token introspection
    
    It appears Nimbus does not check the presence of the Content-Type
    header before parsing it in some versions, and since prior to this
    commit, the code is .toString()-ing the result, a malformed response
    (such as that from a misbehaving cloud gateway) that does not include
    a Content-Type would currently throw a NullPointerException.
    
    In addition to this, I have added a little more information to the
    log output for this module on the standard and reactive implementations
    to aid in debugging authorization/authentication issues much more
    easily. A failed introspection will now log the reason to debug logs
    to make this simpler to debug.

- Fix typo in headers asciidoc. Would have opened a separate PR for
  this but it felt like a little bit of an overkill to remove two characters!

- Fixed final field warnings in opaque token introspectors

-   Ensuring consistency in error handling of opaque providers/managers
    
    The OpaqueTokenAuthenticationProvider now propagates the cause of
    introspection exceptions in the same way that the reactive
    OpaqueTokenReactiveAuthenticationManager does.
    
    Fixed a final field warning on both OpaqueTokenAuthenticationProvider
    and OpaqueTokenReactiveAuthenticationManager.

-   Replace usages of deprecated OAuth2IntrospectionClaimNames
    
    Replace all usages of OAuth2IntrospectionClaimNames with
    the suggested OAuth2TokenIntrospectionClaimNames.
    
    There does not appear to be any further usages of OAuth2IntrospectionClaimNames,
    so it should be suitable for removal when appropriate in accordance with the
    deprecation policy.

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/NimbusReactiveOpaqueTokenIntrospector.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/NimbusReactiveOpaqueTokenIntrospector.java
@@ -121,7 +121,7 @@ public class NimbusReactiveOpaqueTokenIntrospector implements ReactiveOpaqueToke
 		MediaType contentType = responseEntity.headers().contentType().orElseThrow(() -> {
 			this.logger.trace("Did not receive Content-Type from introspection endpoint in response");
 
-			throw new OAuth2IntrospectionException(
+			return new OAuth2IntrospectionException(
 					"Introspection endpoint response was invalid, as no Content-Type header was provided");
 		});
 
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/NimbusReactiveOpaqueTokenIntrospector.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/NimbusReactiveOpaqueTokenIntrospector.java
@@ -24,10 +24,13 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
+import com.nimbusds.oauth2.sdk.ErrorObject;
 import com.nimbusds.oauth2.sdk.TokenIntrospectionResponse;
 import com.nimbusds.oauth2.sdk.TokenIntrospectionSuccessResponse;
 import com.nimbusds.oauth2.sdk.http.HTTPResponse;
 import com.nimbusds.oauth2.sdk.id.Audience;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import reactor.core.publisher.Mono;
 
 import org.springframework.core.io.buffer.DataBuffer;
@@ -53,6 +56,8 @@ import org.springframework.web.reactive.function.client.WebClient;
  * @since 5.2
  */
 public class NimbusReactiveOpaqueTokenIntrospector implements ReactiveOpaqueTokenIntrospector {
+
+	private final Log logger = LogFactory.getLog(getClass());
 
 	private final URI introspectionUri;
 
@@ -113,14 +118,31 @@ public class NimbusReactiveOpaqueTokenIntrospector implements ReactiveOpaqueToke
 	}
 
 	private Mono<HTTPResponse> adaptToNimbusResponse(ClientResponse responseEntity) {
+		MediaType contentType = responseEntity.headers().contentType().orElseThrow(() -> {
+			this.logger.trace("Did not receive Content-Type from introspection endpoint in response");
+
+			throw new OAuth2IntrospectionException(
+					"Introspection endpoint response was invalid, as no Content-Type header was provided");
+		});
+
+		// Nimbus expects JSON, but does not appear to validate this header first.
+		if (!contentType.isCompatibleWith(MediaType.APPLICATION_JSON)) {
+			this.logger.trace("Did not receive JSON-compatible Content-Type from introspection endpoint in response");
+
+			throw new OAuth2IntrospectionException("Introspection endpoint response was invalid, as content type '"
+					+ contentType + "' is not compatible with JSON");
+		}
+
 		HTTPResponse response = new HTTPResponse(responseEntity.rawStatusCode());
-		response.setHeader(HttpHeaders.CONTENT_TYPE, responseEntity.headers().contentType().get().toString());
+		response.setHeader(HttpHeaders.CONTENT_TYPE, contentType.toString());
 		if (response.getStatusCode() != HTTPResponse.SC_OK) {
+			this.logger.trace("Introspection endpoint returned non-OK status code");
+
 			// @formatter:off
 			return responseEntity.bodyToFlux(DataBuffer.class)
 					.map(DataBufferUtils::release)
 					.then(Mono.error(new OAuth2IntrospectionException(
-							"Introspection endpoint responded with " + response.getStatusCode()))
+							"Introspection endpoint responded with HTTP status code " + response.getStatusCode()))
 					);
 			// @formatter:on
 		}
@@ -138,7 +160,10 @@ public class NimbusReactiveOpaqueTokenIntrospector implements ReactiveOpaqueToke
 
 	private TokenIntrospectionSuccessResponse castToNimbusSuccess(TokenIntrospectionResponse introspectionResponse) {
 		if (!introspectionResponse.indicatesSuccess()) {
-			throw new OAuth2IntrospectionException("Token introspection failed");
+			ErrorObject errorObject = introspectionResponse.toErrorResponse().getErrorObject();
+			String message = "Token introspection failed with response " + errorObject.toJSONObject().toJSONString();
+			this.logger.trace(message);
+			throw new OAuth2IntrospectionException(message);
 		}
 		return (TokenIntrospectionSuccessResponse) introspectionResponse;
 	}
@@ -147,6 +172,7 @@ public class NimbusReactiveOpaqueTokenIntrospector implements ReactiveOpaqueToke
 		// relying solely on the authorization server to validate this token (not checking
 		// 'exp', for example)
 		if (!response.isActive()) {
+			this.logger.trace("Did not validate token since it is inactive");
 			throw new BadOpaqueTokenException("Provided token isn't active");
 		}
 	}
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/NimbusReactiveOpaqueTokenIntrospector.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/NimbusReactiveOpaqueTokenIntrospector.java
@@ -17,7 +17,6 @@
 package org.springframework.security.oauth2.server.resource.introspection;
 
 import java.net.URI;
-import java.net.URL;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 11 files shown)


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/NimbusReactiveOpaqueTokenIntrospector.java
```diff
@@ -114,14 +118,35 @@ public NimbusReactiveOpaqueTokenIntrospector(String introspectionUri, WebClient
 	}
 
 	private Mono<HTTPResponse> adaptToNimbusResponse(ClientResponse responseEntity) {
+		MediaType contentType = responseEntity
+				.headers()
+				.contentType()
+				.orElseThrow(() -> {
+					this.logger.debug("Did not receive Content-Type from introspection endpoint in response");
+
+					throw new OAuth2IntrospectionException(
```
ascopes, line 127:
```suggestion
					return new OAuth2IntrospectionException(
```
Since orElseThrow will throw the return value anyway

jzheaux:
Thanks, @ascopes, for the PR and for your thorough research and feedback!

The PR generally seems acceptable - could you please also address the following?

* Spring Security prefers [trace logging for exceptional cases](https://github.com/spring-projects/spring-framework/wiki/Logging#development-logging).
* It appears you should also run `./gradlew format` to satisfy the build.
* I think it would be helpful to demonstrate the NPE in a test - the test should fail before your change and succeed after

ascopes:
> Thanks, @ascopes, for the PR and for your thorough research and feedback!
> ...

Thanks for the points! Will try and find some time in the morning to incorporate these changes and push it.

ascopes:
> demonstrate the NPE in a test

Confirmation of NPE on the non-reactive Nimbus introspector when Content-Type is missing using the `main` branch.
image

> It appears you should also run ./gradlew format to satisfy the build.

Applied. Thanks for making me aware of that!

> Spring Security prefers trace logging for exceptional cases.

Should be fixed now.

---

Should be ready for a review again, @jzheaux. Thanks for taking the time to go through this.

---

On a side note... it appears `OAuth2IntrospectionClaimAccessor` is deprecated in favour of `OAuth2TokenIntrospectionClaimAccessor`, but the `OAuth2IntrospectionAuthenticatedPrincipal` implementation only implements `OAuth2IntrospectionClaimAccessor` and not the replacement class. I have updated the latter to also implement the new version. Likewise I spotted one other place where issuers are being treated as URLs. Since this is part of the public API, I have updated the interface to mark that method as deprecated, and have then added a new method `getIssuerName` that returns the String issuer instead.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringOpaqueTokenIntrospector.java
```diff
@@ -58,16 +57,16 @@
  */
 public class SpringOpaqueTokenIntrospector implements OpaqueTokenIntrospector {
 
-	private final Log logger = LogFactory.getLog(getClass());
+	private static final String AUTHORITY_PREFIX = "SCOPE_";
```
juncevich, line 60: You use 2nd-time AUTHORITY_PREFIX  as constant. The first one was in NimbusReactiveOpaqueTokenIntrospector.java. Is it better to extract constant in one place? What are you thinking about it?

ascopes: I was thinking this, but it appears that this constant is used in many more places than just here. Since this was already in the file, I just moved the line it was on.

I was tempted to do this but I didn't want to make this PR too large and increase the scope of it too much. I was considering whether or not to draft a separate PR once this fix has gone through to get all the declarations for "ROLE_" and "SCOPE_" and put them in one place.

juncevich: Oh, I see. You are absolutely right about not blowing this PR. Good idea about extracting "ROLE_" and "SCOPE_". For me, it looks like an independent PR.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenIntrospectionClaimAccessor.java
```diff
@@ -130,14 +130,25 @@ default String getSubject() {
...
+	@Deprecated
 	default URL getIssuer() {
 		return getClaimAsURL(OAuth2TokenIntrospectionClaimNames.ISS);
 	}
 
+	/**
+	 * Returns the issuer {@code (iss)} of the token as a string.
+	 * @return the issuer of the token as a String
+	 */
+	default String getIssuerName() {
```
jzheaux, line 148: Let's please leave this out for the time being. Applications can do:

```java
String issuer = accessor.getClaim("iss");
```

jzheaux: Given that `JwtClaimAccessor` does the same thing, there may be some more discussion around this, and I'd like to be consistent. Feel free to open a ticket to discuss further.

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/introspection/NimbusOpaqueTokenIntrospectorTests.java
```diff
@@ -119,15 +122,15 @@
 			+ "     }";
 	// @formatter:on
 
-	private static final ResponseEntity<String> ACTIVE = response(ACTIVE_RESPONSE);
+	private static final ResponseEntity<String> ACTIVE = jsonResponse(ACTIVE_RESPONSE);
```
jzheaux, line 125: This change doesn't seem necessary in order to achieve the desired improvements. In order to keep these commits focused on improving introspection's behavior, will you remove changes like these?

jzheaux (changes requested):
Thanks, @ascopes, for the updates. 

Now that SpringOne has passed, I've had time again to take another look. Please see my feedback below.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenIntrospectionClaimAccessor.java
ascopes: @jzheaux sure thing, will revert.

ascopes: @jzheaux sure thing, will revert.

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/introspection/NimbusOpaqueTokenIntrospectorTests.java
ascopes: Yep, totally forgot about this, as originally I had additional code to deal with something else in here too (was a couple of weeks ago now so I don't remember the full details, but I know I worked out a way around this). Will revert. Thanks for pointing this out.

ascopes:
Will rebase onto spring-projects/spring-security@master once I have applied the requested changes.

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/introspection/NimbusOpaqueTokenIntrospectorTests.java
ascopes: Should be reverted now.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2TokenIntrospectionClaimAccessor.java
ascopes: Reverted.

ascopes:
@jzheaux made the suggested changes, I assume you are still happy for things like the typo fix in [the documentation](https://github.com/spring-projects/spring-security/pull/10175/files#diff-e368979d9a0a2de6d6e8ad9d2c88e3b09da886b56e0fbb5a05852b932d12db30) to go through with this, since it is relatively minor?

Let me know if there is anything else you want me to do! Also, with regards to @juncevich's [comment](https://github.com/spring-projects/spring-security/pull/10175#discussion_r695556566) above, do you agree with making a separate PR or issue afterwards to consider this?

jzheaux:
Thanks for the updates, @ascopes. Yes, I think that fixing the typo in the documentation is fine as it is an obvious fix and is in a separate commit.

In preparation for merging, will you please squash your feedback updates into the appropriate commits? Also please make sure to run `format` for each commit so that needed formatting changes are applied atomically.

> do you agree with making a separate PR or issue afterwards to consider this?

I'd probably recommend an issue first. Scope is a concept unique to the `oauth2` modules, so it's not yet clear to me what object would go into `core` that would contain both the role and scope concepts.

ascopes:
> Thanks for the updates, @ascopes. Yes, I think that fixing the typo in the documentation is fine as it is an obvious fix and is in a separate commit.
> 
> In preparation for merging, will you please squash your feedback updates into the appropriate commits? Also please make sure to run `format` for each commit so that needed formatting changes are applied atomically.
> 
> > do you agree with making a separate PR or issue afterwards to consider this?
> 
> I'd probably recommend an issue first. Scope is a concept unique to the `oauth2` modules, so it's not yet clear to me what object would go into `core` that would contain both the role and scope concepts.

Should be updated now; tests appear to be passing, checkstyle appears to be passing, and I've run format on the respective commits.

Also rebased onto main again.

Please can you re-review when you have time? Thanks :-)

jzheaux (approved):

jzheaux:
Thanks for your detailed contribution, @ascopes! This is now merged into `main`.

Issue #10172, opened 2021-08-13
AuthenticationPrincipalArgumentResolver making troubles with primitives


djechelon:
**Describe the bug**

I discovered an issue in the @AuthenticationPrincipal annotation handler

I have a custom `UserDetails` sub-interface displaying a `userId` attribute of `int` type, corresponding to the user's primary key in the database. The primary key is always present if there is a principal, so it's type int everywhere.

I retrieve it using @AuthenticationPrincipal in my REST controllers

```
public interface OrbitUserDetails extends UserDetails {

    int getUserId();

}

public class Controller {

    public ResponseEntity<?> doSomething(...., @AuthenticationPrincipal(expression="userId") int actorUserId) {
    }

}
```

The problem is that the `AuthenticationPrincipalArgumentResolver` does not correctly resolve int, a null argument is derived and is causing an **unclear exception** at controller invocation.
The exception does not clearly indicate the problem and I had to debug deep into Spring to find out.

Code fragment from [Spring Security](https://github.com/spring-projects/spring-security/blob/5.4.6/web/src/main/java/org/springframework/security/web/method/annotation/AuthenticationPrincipalArgumentResolver.java#L100):

```
		Object principal = authentication.getPrincipal();
		AuthenticationPrincipal annotation = findMethodAnnotation(AuthenticationPrincipal.class, parameter);
		String expressionToParse = annotation.expression();
		if (StringUtils.hasLength(expressionToParse)) {
			StandardEvaluationContext context = new StandardEvaluationContext();
			context.setRootObject(principal);
			context.setVariable("this", principal);
			context.setBeanResolver(this.beanResolver);
			Expression expression = this.parser.parseExpression(expressionToParse);
			principal = expression.getValue(context);
		}
		if (principal != null && !parameter.getParameterType().isAssignableFrom(principal.getClass())) {
			if (annotation.errorOnInvalidType()) {
				throw new ClassCastException(principal + " is not assignable to " + parameter.getParameterType());
			}
			return null;
		}
```

At line 115, the `principal` object is set to an Integer, but at line 117 the cast fails (because, manifestly, `int` cannot be assigned from an integer), but since `errorOnInvalidType` is false by default, the returned argument is null.

The argument is then passed to the MVC method argument resolver, which cannot invoke the REST method `doSomething` but can't explain why exactly

This is the stack trace

```
java.lang.IllegalArgumentException: null
[...]

PR #10213, opened 2021-08-19
Document AuthenticationPrincipal type cannot be primitive

eleftherias:
Closes gh-10172


eleftherias:
Closing in favour of gh-10216, which instead adds support for primitive types.

PR #10216, opened 2021-08-20
Allow AuthenticationPrincipal argument type to be primitive

eleftherias:
Closes gh-10172


marcusdacoregio (approved):
Nice work, @eleftherias. It seems good to me.

PR #10230, opened 2021-09-01
Introduce @OverrideMethodSecurity to avoid boilerplate code

quaff:
```java
public class CrudRepository<T> {

  public void create(T entity) {

  }
  public void update(T entity) {

  }
  public void delete(T entity) {

  }

}
```
Before this commit:
```java
@Secured({ "ROLE_USER" })
public class ContractRepository extends CrudRepository<Contract> {

  @Override
  public void create(Contract contract) {
    super.create(contract);
  }

  @Override
  public void update(Contract contract) {
    super.update(contract);
  }

  @Override
  public void delete(Contract contract) {
    super.delete(contract);
  }

}
```
After this commit:
```java
@Secured({ "ROLE_USER" })
@OverrideMethodSecurity
public class ContractRepository extends CrudRepository<Contract> {

}
```

### DIFF
--- a/core/src/main/java/org/springframework/security/access/annotation/OverrideMethodSecurity.java
+++ b/core/src/main/java/org/springframework/security/access/annotation/OverrideMethodSecurity.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2002-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


jgrandja:
Thanks for the PR @quaff.

I don't feel we should add `@OverrideMethodSecurity` as it serves only as a _"convenience feature"_  for avoiding boilerplate code for method overrides.

With new API's, our goal is to provide features that provide high value and are used widely. I'm not sure of the demand for this convenience feature?

FYI, whenever we introduce new API's, we are required to support it based on the Spring OSS support policy and associated deprecation schedule. This ultimately provides us less flexibility in making passive changes. So we need to ensure new API's are well thought out and are needed by the community. The only way we know if the community needs it is by logging a new feature request and tracking the upvotes. Since there is no issue associated with this new feature, we don't know the demand for this feature.

Based on this, I'm going to close this but feel free to log a new issue for this feature request and we'll see if there is demand within the community.

Issue #10227, opened 2021-08-27
Exception messages cannot be properly internationalized


Hccake:
**Describe the bug**

AbstractUserDetailsAuthenticationProvider class defines an attribute messages, and gives the initial value for it  。

```java
	protected MessageSourceAccessor messages = SpringSecurityMessageSource.getAccessor();
```

But this class also implements MessageSourceAware，AbstractApplicationContext# initMessageSource () method will ensure that the container will always be a messageSource , when the user does not customize messageSource , an empty DelegatingMessageSource will be registered.

This leads to AbstractUserDetailsAuthenticationProvider messageSource attribute is never in the initial value, because always triggers MessageSourceAware#setMessageSource ()  

Exception messages thrown in this class will always be defaultMessage unless the user manually registers the messageSource and scans basename "org.springframework.security.messages".  

**Expected behavior**

Hope in the user without configuration messageSource, back to use SpringSecurityMessageSource, this ensures that the internationalization of correct error message

eleftherias:
Thanks for reaching out @Hccake.
Have you registered a `MessageSource` bean in your application?
```
@Bean
public MessageSource messageSource() {
    ReloadableResourceBundleMessageSource messageSource = new ReloadableResourceBundleMessageSource();
    messageSource.addBasenames("classpath:org/springframework/security/messages");
    return messageSource;
}
```

Hccake:
@eleftherias 

When I register such a messageSource, it works.

But if MessageSourceAccessor is provided by default, why should users register messageSource themselves ?

eleftherias:
Thank you for confirming that, @Hccake.

I'm not sure I understand your question. Perhaps this section of the reference [documentation](https://docs.spring.io/spring-security/site/docs/current/reference/html5/#localization) is able to answer it?

Hccake:
What I mean by this is whether spring-Security can also provide support for internationalization without users manually registering messageSource.

Hccake:
@eleftherias 

hello，I have created a PR for this issue, can you have a look

eleftherias:
The message source is necessary to distinguish between the localization that is provided by Spring Security (which can be seen [here](https://github.com/spring-projects/spring-security/tree/main/core/src/main/resources/org/springframework/security)) and any custom messages that you have added.

If you feel that we are missing any language, we are happy to accept a PR directly in Spring Security with the internationalized messages.

Hccake:
@eleftherias 

[...]

PR #10238, opened 2021-09-06
Use SpringSecurityMessageSource as international information back message source.

Hccake:
In hibernate-Validation, the default ValidatehiResourceBundle will exist regardless of whether the user has customized the ResourceBundle or not, as the final fallback option  

In spring-security，when the user does not specify the message source, should also be able to use the SpringSecurityMessageSource international configuration.  

Closes [gh-10227](https://github.com/spring-projects/spring-security/issues/10227)

### DIFF
--- a/core/src/main/java/org/springframework/security/core/SpringSecurityMessageSource.java
+++ b/core/src/main/java/org/springframework/security/core/SpringSecurityMessageSource.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2004, 2005, 2006 Acegi Technology Pty Limited
+ * Copyright 2004, 2005, 2006, 2021 Acegi Technology Pty Limited
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,7 +16,6 @@
 
 package org.springframework.security.core;
 
-import org.springframework.context.support.MessageSourceAccessor;
 import org.springframework.context.support.ResourceBundleMessageSource;
 
 /**
@@ -24,9 +23,9 @@ import org.springframework.context.support.ResourceBundleMessageSource;
  * <p>
  * All Spring Security classes requiring message localization will by default use this
  * class. However, all such classes will also implement <code>MessageSourceAware</code> so
- * that the application context can inject an alternative message source. Therefore this
- * class is only used when the deployment environment has not specified an alternative
- * message source.
+ * that the application context can inject an alternative message source. When message
+ * source is specified in the deployment environment, it is used as the rollback message
+ * source
  * </p>
  *
  * @author Ben Alex
@@ -35,10 +34,6 @@ public class SpringSecurityMessageSource extends ResourceBundleMessageSource {
 
 	public SpringSecurityMessageSource() {
 		setBasename("org.springframework.security.messages");
-	}
-
-	public static MessageSourceAccessor getAccessor() {
-		return new MessageSourceAccessor(new SpringSecurityMessageSource());
 	}
 
 }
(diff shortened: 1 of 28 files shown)


core/src/main/java/org/springframework/security/core/SpringSecurityMessageSource.java
```diff
@@ -16,17 +16,16 @@
...
 /**
  * The default <code>MessageSource</code> used by Spring Security.
  * <p>
  * All Spring Security classes requiring message localization will by default use this
  * class. However, all such classes will also implement <code>MessageSourceAware</code> so
- * that the application context can inject an alternative message source. Therefore this
- * class is only used when the deployment environment has not specified an alternative
- * message source.
+ * that the application context can inject an alternative message source. When message
+ * Source is specified in the deployment environment, it is used as the rollback message
```
juncevich, line 27: Is it misspell in the word `Source`? With capital `S`.

core/src/main/java/org/springframework/security/core/SpringSecurityMessageSourceAccessor.java
```diff
@@ -0,0 +1,230 @@
...
+import org.springframework.lang.Nullable;
+
+/**
+ * Like {@link MessageSourceAccessor}, This is a helper class for easy access to message
+ * from a MessageSource, providing various overloaded getMessage methods.
+ *
+ * <p>
+ * Unlike MessageSourceAccessor, it sets the SpringSecurityMessageSource to fallback
+ * messageSource. When a user provides messageSource cannot find information, will make
+ * SpringSecurityMessageSource out looking .
```
juncevich, line 34: Maybe extra space between `looking` and `.`.

core/src/main/java/org/springframework/security/core/SpringSecurityMessageSourceAccessor.java
```diff
@@ -0,0 +1,230 @@
...
+import java.util.Locale;
+
+import org.springframework.context.MessageSource;
+import org.springframework.context.MessageSourceResolvable;
+import org.springframework.context.NoSuchMessageException;
+import org.springframework.context.support.MessageSourceAccessor;
+import org.springframework.lang.Nullable;
+
+/**
+ * Like {@link MessageSourceAccessor}, This is a helper class for easy access to message
```
juncevich, line 28: Also about `T` in `This`.

Hccake:
@juncevich 

Hello, I have fixed the error in these comments

Hccake:
@eleftherias Hello, is there any progress on this PR?

core/src/main/java/org/springframework/security/core/SpringSecurityMessageSource.java
```diff
@@ -37,8 +36,4 @@ public SpringSecurityMessageSource() {
 		setBasename("org.springframework.security.messages");
 	}
 
-	public static MessageSourceAccessor getAccessor() {
```
jgrandja, line 40: This cannot be removed as it breaks compatibility. Spring Security follows [semantic versioning](https://semver.org/).

jgrandja (commented):
@Hccake The changes in this PR cannot be accepted as it breaks compatibility in 2 parts:

1) Removal of `SpringSecurityMessageSource.getAccessor()`
2) The behaviour has changed in all `MessageSourceAware` components as it now uses the proposed `SpringSecurityMessageSourceAccessor(messageSource)` instead of `MessageSourceAccessor(messageSource)`. Given that `SpringSecurityMessageSourceAccessor` uses a delegation-based strategy, by first delegating to the provided `messageSource` and then falling back to `SpringSecurityMessageSource`, this effectively changes the behaviour. This new behaviour could potentially break existing applications that do not expect a fallback (or delegation-based) strategy.

Furthermore, I took a look at the commit log and there haven't been any major improvements to `SpringSecurityMessageSource` since 2007, so I'm not convinced this delegation-based strategy is needed.

I'm going to close this based on the above reasoning.

Issue #10226, opened 2021-08-26
Replace SecurityContextHolder#addListener with SecurityContextHolder#setSecurityContextHolderStrategy


jzheaux:
Since listening for security context changes happens at the strategy level, `SecurityContextHolder#addListener` leaks that abstraction.

Instead, let's replace `addListener` with `setSecurityContextHolderStrategy` and make `ListeningSecurityContextHolderStrategy` public. This will allow the listening support to remain with the strategy.

`ListeningSecurityContextHolderStrategy` should not need to peek the `SecurityContext` since the semantic difference between a null `SecurityContext` and one with a null `Authentication` is negligible.

An additional benefit to `setSecurityContextHolderStrategy` over `addListener` is that it simplifies removing a custom strategy that has references to objects that need garbage collecting.

PR #10246, opened 2021-09-09
Replace SecurityContextHolder#addListener

jzheaux:
Closes gh-10226

### DIFF
--- a/core/src/main/java/org/springframework/security/core/context/SecurityContextHolder.java
+++ b/core/src/main/java/org/springframework/security/core/context/SecurityContextHolder.java
@@ -56,6 +56,8 @@ public class SecurityContextHolder {
 
 	public static final String MODE_GLOBAL = "MODE_GLOBAL";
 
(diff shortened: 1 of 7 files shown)


rwinch (approved):
The changes look good.

core/src/main/java/org/springframework/security/core/context/SecurityContextHolder.java
```diff
@@ -69,34 +71,41 @@
 	}
 
 	private static void initialize() {
```
jonatan-ivanov, line 73: This should be called during context init so shouldn't be called concurrently, right?

jonatan-ivanov (approved):

core/src/main/java/org/springframework/security/core/context/SecurityContextHolder.java
jzheaux: This is called whenever the strategy changes. For example, if `setContextHolderStrategy` were called concurrently, then `initialize()` would be called concurrently. This isn't the way the API is intended to be used, though.

PR #10262, opened 2021-09-13
OAuth2 Token Response Customization For Gh 10260

bishoybasily:
For Gh #10261


bishoybasily:
Hi @jgrandja, Could you please give me your feedback about this one?
Thank You!

PR #10276, opened 2021-09-16
Fix typo

heowc:


jzheaux:
Thanks for contributing, @heowc! Even though this is a quick fix, will you please also update the copyright headers in each of the files you modified? The end year should be "2021" in each one.

heowc:
> Thanks for contributing, @heowc! Even though this is a quick fix, will you please also update the copyright headers in each of the files you modified? The end year should be "2021" in each one.

Fixed 😄

jzheaux:
Thanks, @heowc! This is now merged into `main`.

Issue #10247, opened 2021-09-09
Improve Method Security Logging


jzheaux:
Logging should be refactored for when `@EnableGlobalMethodSecurity` is used. For example, on startup, method security's DEBUG and TRACE logs are verbose without adding much value.

Logging should be added for when `@EnableMethodSecurity` is used.

We should keep [Spring Framework's logging guidelines](https://github.com/spring-projects/spring-framework/wiki/Logging#development-logging) in mind as we go forward.

Issue #9287, opened 2020-12-16
Add Reasons to AuthorizationDecisions


jzheaux:
It's helpful to know why a security decision was made.

`AuthorityReactiveAuthorizationManager` and `AuthorityAuthorizationManager` could do this by constructing an `AuthorityAuthorizationDecision`:

```java
public class AuthorityAuthorizationManager implements AuthorizationManager {
    // ...

    public static class AuthorityAuthorizationDecision extends AuthorizationDecision {
        private Collection<GrantedAuthority> authorities;

        public AuthorityAuthorizationDecision(boolean decision, GrantedAuthority... authorities) {
            super(decision);
            // ...
        }

        public Collection<GrantedAuthority> getAuthorities() { 
            return this.authorities;
        }
    }
}
```

Likewise with `AuthenticatedReactiveAuthorizationManager`, `AuthenticatedAuthorizationManager` and `AuthenticatedAuthorizationDecision`.

Each implementation would likely need to override `toString` to assist with logging authorization events.

rwinch:
What is the concrete use case for this?

jzheaux:
I was thinking of [event logging](https://github.com/spring-projects/spring-security/pull/8996#discussion_r506649210). It's helpful in logs to know why access was granted to a resource, for example.

Initially, I was thinking of combining this ticket with #9286. Your question may be a hint that they go together.

PR #10279, opened 2021-09-17
Improve Method Security logging

marcusdacoregio:
This PR removes the trace logs from the `PrePostAnnotationSecurityMetadataSource` which were producing too much noise in logs without adding much value.

There is already some logs that provide value when the `@Pre/PostAuthorize` annotations are being used with `@EnableGlobalMethodSecurity`:

```
2021-09-17 10:01:49.539 DEBUG 55564 --- [           main] .PrePostAnnotationSecurityMetadataSource : @org.springframework.security.access.prepost.PreAuthorize(value="hasRole('USER')") found on specific method: public java.lang.String example.IndexController.index()
2021-09-17 10:01:49.544 DEBUG 55564 --- [           main] m.DelegatingMethodSecurityMetadataSource : Caching method [CacheKey[example.IndexController; public java.lang.String example.IndexController.index()]] with attributes [[authorize: 'hasRole('USER')', filter: 'null', filterTarget: 'null']]
```

Additionally, it adds logs for when `@EnableMethodSecurity` is used with `@Pre/PostAuthorize`, `@Secured` and JSR-250 `@PermitAll`, `@DenyAll`, `@RolesAllowed`.

```
2021-09-17 10:06:31.277 DEBUG 55847 --- [nio-8080-exec-4] s.s.a.m.PreAuthorizeAuthorizationManager : Authorizing method pre-invocation 'ReflectiveMethodInvocation: public java.lang.String example.IndexController.index(); target is of class [example.IndexController]' with attribute 'ExpressionAttribute [Expression=hasRole('USER')]', authentication name 'user' and decision 'AuthorizationDecision [granted=true]'
2021-09-17 10:06:31.278 DEBUG 55847 --- [nio-8080-exec-4] horizationManagerBeforeMethodInterceptor : Authorized ReflectiveMethodInvocation: public java.lang.String example.IndexController.index(); target is of class [example.IndexController]
2021-09-17 10:06:31.285 DEBUG 55847 --- [nio-8080-exec-4] .s.a.m.PostAuthorizeAuthorizationManager : Authorizing method post-invocation 'org.springframework.security.authorization.method.MethodInvocationResult@3bcc1d1e' with attribute 'ExpressionAttribute [Expression=returnObject == 'index']', authentication name 'user' and decision 'AuthorizationDecision [granted=true]'
2021-09-17 10:06:31.285 DEBUG 55847 --- [nio-8080-exec-4] thorizationManagerAfterMethodInterceptor : Authorized ReflectiveMethodInvocation: public java.lang.String example.IndexController.index(); target is of class [example.IndexController]
```

And when the authorization fails:

```
2021-09-17 10:09:00.755 DEBUG 57196 --- [nio-8080-exec-4] s.s.a.m.PreAuthorizeAuthorizationManager : Authorizing method pre-invocation 'ReflectiveMethodInvocation: public java.lang.String example.IndexController.index(); target is of class [example.IndexController]' with attribute 'ExpressionAttribute [Expression=hasRole('ADMIN')]', authentication name 'user' and decision 'AuthorizationDecision [granted=false]'
2021-09-17 10:09:00.756 TRACE 57196 --- [nio-8080-exec-4] horizationManagerBeforeMethodInterceptor : Failed to authorize 'ReflectiveMethodInvocation: public java.lang.String example.IndexController.index(); target is of class [example.IndexController]' with authorization manager 'org.springframework.security.authorization.method.PreAuthorizeAuthorizationManager@37808e23'
```

Closes gh-10247


core/src/main/java/org/springframework/security/authorization/method/PostAuthorizeAuthorizationManager.java
```diff
@@ -76,7 +81,13 @@ public AuthorizationDecision check(Supplier<Authentication> authentication, Meth
 				mi.getMethodInvocation());
 		this.expressionHandler.setReturnObject(mi.getResult(), ctx);
 		boolean granted = ExpressionUtils.evaluateAsBoolean(attribute.getExpression(), ctx);
-		return new AuthorizationDecision(granted);
+		AuthorizationDecision decision = new AuthorizationDecision(granted);
+		if (this.logger.isDebugEnabled()) {
+			this.logger.debug(LogMessage.format(
+					"Authorizing method post-invocation '%s' with attribute '%s', authentication name '%s' and decision '%s'",
+					mi, attribute, authentication.get().getName(), decision));
```
jzheaux, line 88: `getName()` could be PII, like maybe an email address.

You could log `authentication.get()` more safely since an application can override the `toString()` method, but I think it may be best to leave it out of the message altogether. Leaving it out also means that we don't activate a supplier unnecessarily, which is better for performance if that means a session lookup.

jzheaux (changes requested):
Thanks, @marcusdacoregio! I've left feedback inline.

core/src/main/java/org/springframework/security/authorization/method/Jsr250AuthorizationManager.java
```diff
@@ -81,7 +86,13 @@ public void setRolePrefix(String rolePrefix) {
 	@Override
 	public AuthorizationDecision check(Supplier<Authentication> authentication, MethodInvocation methodInvocation) {
 		AuthorizationManager<MethodInvocation> delegate = this.registry.getManager(methodInvocation);
-		return delegate.check(authentication, methodInvocation);
+		AuthorizationDecision decision = delegate.check(authentication, methodInvocation);
+		if (this.logger.isDebugEnabled()) {
+			this.logger.debug(LogMessage.format(
+					"Authorizing method invocation '%s' with authentication name '%s' and decision '%s'",
```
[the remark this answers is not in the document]
marcusdacoregio: "Authorized" may sound that the permission was granted, which might not be always the case. What do you think about the message `Authorization decision '%s' on method invocation '%s'`?.

The log makes sense here to help to know what was the `AuthorizationDecision`. And, in `Pre/PostAuthorizeAuthorizationManager` it also logs what's in the expression attribute.

marcusdacoregio:
Thanks very much for your review @jzheaux. I've updated the PR and replied this comment https://github.com/spring-projects/spring-security/pull/10279#discussion_r714339702

jzheaux (changes requested):
Thanks, @marcusdacoregio! I've left some inline feedback.

marcusdacoregio:
Thanks @jzheaux, nice suggestions.  I updated the PR and it's ready for your review again.

core/src/main/java/org/springframework/security/authorization/method/Jsr250AuthorizationManager.java
```diff
@@ -81,7 +81,11 @@ public void setRolePrefix(String rolePrefix) {
 	@Override
 	public AuthorizationDecision check(Supplier<Authentication> authentication, MethodInvocation methodInvocation) {
 		AuthorizationManager<MethodInvocation> delegate = this.registry.getManager(methodInvocation);
-		return delegate.check(authentication, methodInvocation);
+		AuthorizationDecision decision = delegate.check(authentication, methodInvocation);
+		if (decision == null) {
+			return null;
+		}
+		return new AuthorizationManagerAuthorizationDecision(decision.isGranted(), delegate);
```
jzheaux, line 88: I think `AuthorityAuthorizationManager` might be a more reasonable place since it could return the granted authorities needed. It could return `AuthorityAuthorizationDecision`, for example.

Returning a service component as a member may not be as helpful as you are thinking since it will not typically have any getters.

jzheaux (changes requested):

In addition to that feedback, will you please separate out the work you are doing with custom `AuthorizationDecision`s into a separate commit, resolving https://github.com/spring-projects/spring-security/issues/9287 at the same time?

core/src/main/java/org/springframework/security/authorization/method/Jsr250AuthorizationManager.java
marcusdacoregio: Yeah, I think it was an oversight because I did not see the `AuthorityAuthorizationManager`

marcusdacoregio:
Thanks @jzheaux, I've updated the PR based on your suggestions

core/src/main/java/org/springframework/security/authorization/AuthorityAuthorizationDecision.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Represents an {@link AuthorizationDecision} based on a collection of authorities
+ *
+ * @author Marcus Da Coregio
+ * @since 5.6
+ */
+class AuthorityAuthorizationDecision extends AuthorizationDecision {
+
+	private final Collection<String> authorities;
```
jzheaux, line 29: I'd recommend that this be `Collection<GrantedAuthority>` as this gives more flexibility to applications to use their custom granted authority implementations.

core/src/main/java/org/springframework/security/authorization/AuthorityAuthorizationDecision.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.util.Collection;
+
+/**
+ * Represents an {@link AuthorizationDecision} based on a collection of authorities
+ *
+ * @author Marcus Da Coregio
+ * @since 5.6
+ */
+class AuthorityAuthorizationDecision extends AuthorizationDecision {
```
jzheaux, line 27: I think this should be `public` so that applications that call `AuthorizationManager` can get the extra details out.

core/src/main/java/org/springframework/security/authorization/method/ExpressionAttributeAuthorizationDecision.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.springframework.security.authorization.AuthorizationDecision;
+
+/**
+ * Represents an {@link AuthorizationDecision} based on a {@link ExpressionAttribute}
+ *
+ * @author Marcus Da Coregio
+ * @since 5.6
+ */
+class ExpressionAttributeAuthorizationDecision extends AuthorizationDecision {
```
jzheaux, line 27: I think this should be `public` so that applications that call `AuthorizationManager` can get the extra details out.

core/src/main/java/org/springframework/security/authorization/AuthorityReactiveAuthorizationManager.java
```diff
@@ -48,8 +48,8 @@
 				.flatMapIterable(Authentication::getAuthorities)
 				.map(GrantedAuthority::getAuthority)
 				.any(this.authorities::contains)
-				.map(AuthorizationDecision::new)
-				.defaultIfEmpty(new AuthorizationDecision(false));
+				.map((granted) -> ((AuthorizationDecision) new AuthorityAuthorizationDecision(granted, this.authorities)))
```
jzheaux, line 51: Given my other comment about changing this constructor to take a collection of `GrantedAuthority`s, I think that this class should store a collection of `GrantedAuthority`s.

core/src/main/java/org/springframework/security/authorization/AuthorityAuthorizationManager.java
```diff
@@ -124,7 +124,7 @@ private AuthorityAuthorizationManager(String... authorities) {
 	@Override
 	public AuthorizationDecision check(Supplier<Authentication> authentication, T object) {
 		boolean granted = isGranted(authentication.get());
-		return new AuthorizationDecision(granted);
+		return new AuthorityAuthorizationDecision(granted, this.authorities);
```
jzheaux, line 127: Given my other comment about changing this constructor to take a collection of `GrantedAuthority`s, I think that this class should store a collection of `GrantedAuthority`s.

jzheaux (changes requested):
Thanks, @marcusdacoregio. I've left feedback inline about the classes recently introduced.

PR #10300, opened 2021-09-19
fixes #10236

### LINKED ISSUE #10236: Buggy  PostInvocationAdviceProvider::supports(Class<?> clazz)   implementation
The current implementation doesn't allow to support `MethodInvocation` - inherited classes :
````java
@Override
	public boolean supports(Class<?> clazz) {
		return clazz.isAssignableFrom(MethodInvocation.class);
	}
````
Looks like the intention was the opposite :

````java
@Override
	public boolean supports(Class<?> clazz) {
		return MethodInvocation.class.isAssignableFrom(clazz);
	}
````

Please change [here](https://github.com/spring-projects/spring-security/blob/006b9b960797d279b31cf8c8d16f1549c5632b2c/core/src/main/java/org/springframework/security/access/prepost/PostInvocationAdviceProvider.java#L76)

### PR DESCRIPTION
**Pull request by jvmlet**

### CODE REVIEW

**Review by jzheaux — CHANGES_REQUESTED** [review]
Thanks for the fix, @jvmlet!

Even though this is a simple fix, can you please still do the following:

1. Add a unit test that fails before your change, but passes after
2. Update the copyright year in the class
3. Change your commit message to have a descriptive title, like so:
```
Fix isAssignableFrom Order

Closes gh-10246
```
This makes it easier to find commits when scanning Git logs.

**Comment by jvmlet** [comment]
@jzheaux , please review the updated PR

**Review by jzheaux — CHANGES_REQUESTED** [review]
Thanks for the updates, @jvmlet!

It looks like the PR build is failing due to some formatting concerns. Will you please run:

```
./gradlew format check
```

Some things will be fixed automatically, but there may be one item that you need to adjust manually.

When you are done, could you please rebase with `main` and squash your commits into a single commit?

jvmlet:
Sorry, @jzheaux ,I messed up the squash. I deleted my branch and will  recreate the PR from the fresh branch

PR #10304, opened 2021-09-21
Fix typo in digest.adoc

1993heqiang:


jgrandja:
Thanks for the update @1993heqiang ! This is now in main.

Issue #7722, opened 2019-12-11
Saml2WebSsoAuthenticationFilter ignores the authentication details


horca:
Hello Spring Security team,

#### Problem
When using the `UsernamePasswordAuthenticationFilter` we are setting custom authentication details with `AuthenticationDetailsSource` to be used further down the security chain. Unfortunately, the `Saml2WebSsoAuthenticationFilter` does not populate the authentication details even though it inherits from `AbstractAuthenticationProcessingFilter` which allows to define the details source.

#### Solution
After the `Saml2AuthenticationToken` is created, populate the details. Afterwards, when creating the `Saml2Authentication` within the authetication provider, just copy the details over.

fhanik:
I feel these two are related, and one solution could satisfy both
https://github.com/spring-projects/spring-security/issues/7465

horca:
I'm not really sure if this issue relates to the UserService and AuthenticatedPrincipal... Please see [this commit](https://github.com/horca/spring-security/commit/a524e40dbdc9d53a61837e117580ec80a77bd843) for more information on what I'm trying to achieve (same mechanism as UsernamePasswordAuthenticationFilter).

horca:
Setting the details within the authentication provider is not needed as they are copied inside the `ProviderManager`

jzheaux:
@horca, thanks for laying out some of the details here. I think it makes sense to have the SAML 2.0 support align with other authentication filters in Spring Security.

Would you be interested in submitting a PR to have the filter set the details object on the `Saml2AuthenticationToken` instance and have the authentication provider set the details on the `Saml2Authentication` instance?

adi-devops:
@jzheaux I would love to give it a try if its open for contribution. Can you please explain litte more on what is the implementation plan?

PR #10306, opened 2021-09-21
Saml2WebSsoAuthenticationFilter adds authentication details

Kehrlann:
Closes gh-7722.

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/servlet/filter/Saml2WebSsoAuthenticationFilter.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/servlet/filter/Saml2WebSsoAuthenticationFilter.java
@@ -19,6 +19,7 @@ package org.springframework.security.saml2.provider.service.servlet.filter;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import org.springframework.security.authentication.AbstractAuthenticationToken;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
 import org.springframework.security.saml2.core.Saml2Error;
@@ -109,6 +110,7 @@ public class Saml2WebSsoAuthenticationFilter extends AbstractAuthenticationProce
 					"No relying party registration found");
 			throw new Saml2AuthenticationException(saml2Error);
 		}
+		setDetails(request, authentication);
 		this.authenticationRequestRepository.removeAuthenticationRequest(request, response);
 		return getAuthenticationManager().authenticate(authentication);
 	}
@@ -135,6 +137,13 @@ public class Saml2WebSsoAuthenticationFilter extends AbstractAuthenticationProce
 		if (this.authenticationConverter instanceof Saml2AuthenticationTokenConverter) {
 			Saml2AuthenticationTokenConverter authenticationTokenConverter = (Saml2AuthenticationTokenConverter) this.authenticationConverter;
 			authenticationTokenConverter.setAuthenticationRequestRepository(authenticationRequestRepository);
+		}
+	}
+
+	private void setDetails(HttpServletRequest request, Authentication authentication) {
+		if (AbstractAuthenticationToken.class.isAssignableFrom(authentication.getClass())) {
+			Object details = this.authenticationDetailsSource.buildDetails(request);
+			((AbstractAuthenticationToken) authentication).setDetails(details);
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


Kehrlann:
Hi there!

We've been needing authentication details in our product, specifically `WebAuthenticationDetails`. I saw gh-7722 and decided to submit this.

Notes:
- I added a test in `Saml2WebSsoAuthenticationFilterTests` rather than more integration-style `Saml2LoginConfigurerTests`, but I am unsure whether it is the right place. Maybe they should be in both?
- There is a cast to `AbstractAuthenticationToken`. Technically the authentication converter returns a `Saml2AuthenticationToken` by default. Howver, that is not required. I can't really imagine a setup where a user provides both their own converter AND their own auth provider (they should probably provide their own filter if they do this), but I prefer being extra cautious. Please advise if that is too cautious.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/servlet/filter/Saml2WebSsoAuthenticationFilter.java
```diff
@@ -109,6 +110,12 @@ public Authentication attemptAuthentication(HttpServletRequest request, HttpServ
 					"No relying party registration found");
 			throw new Saml2AuthenticationException(saml2Error);
 		}
+
+		if (AbstractAuthenticationToken.class.isAssignableFrom(authentication.getClass())) {
+			Object details = this.authenticationDetailsSource.buildDetails(request);
+			((AbstractAuthenticationToken) authentication).setDetails(details);
+		}
+
```
marcusdacoregio, line 118: Could you please extract this in a private method called `setDetails` so it becomes more readable?

marcusdacoregio (changes requested):
Hi @Kehrlann, thank you very much for the PR. I've left some feedback inline.

saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/servlet/filter/Saml2WebSsoAuthenticationFilterTests.java
```diff
@@ -119,6 +123,42 @@ public void attemptAuthenticationWhenSavedAuthnRequestThenRemovesAuthnRequest()
 		verify(authenticationRequestRepository).removeAuthenticationRequest(this.request, this.response);
 	}
 
+	@Test
+	public void attemptAuthenticationAddsDetails() {
+		Saml2AuthenticationRequestRepository<AbstractSaml2AuthenticationRequest> authenticationRequestRepository = mock(
```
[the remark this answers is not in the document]
Kehrlann: Good catch 👍

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/servlet/filter/Saml2WebSsoAuthenticationFilter.java
```diff
@@ -109,6 +110,7 @@ public Authentication attemptAuthentication(HttpServletRequest request, HttpServ
 					"No relying party registration found");
 			throw new Saml2AuthenticationException(saml2Error);
 		}
+		this.setDetails(request, authentication);
```
marcusdacoregio, line 113: In Spring Security's codebase, there are some places that we still call an instance method using `this`, but we are moving away from it. Would you please remove the `this.` and just use `setDetails(request, authentication)`?

saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/servlet/filter/Saml2WebSsoAuthenticationFilterTests.java
```diff
@@ -119,6 +123,39 @@ public void attemptAuthenticationWhenSavedAuthnRequestThenRemovesAuthnRequest()
...
+		this.filter.setAuthenticationDetailsSource(authenticationDetailsSource);
+		this.request.setPathInfo("/some/other/path/idp-registration-id");
+		this.filter.attemptAuthentication(this.request, this.response);
+		Assertions.assertEquals(details, token.getDetails());
+	}
+
+	@Test
+	public void attemptAuthenticationWhenAuthenticationNotAbstractAuthenticationTokenDoesNotAddDetails() {
+		Saml2AuthenticationRequestRepository<AbstractSaml2AuthenticationRequest> authenticationRequestRepository = mock(
+				Saml2AuthenticationRequestRepository.class);
```
marcusdacoregio, line 145: I think the `authenticationRequestRepository` is not needed in this test.

marcusdacoregio (changes requested):
Thank you @Kehrlann.

I've left a couple more suggestions in line. Once you fix them we can merge the PR!

marcusdacoregio (approved):

marcusdacoregio:
Thank you very much @Kehrlann. This is now merged into `main`.

Issue #10320, opened 2021-09-24
Provide an easier way to configure an ObjectPostProcessor


sjohnr:
**Expected Behavior**

Using an `ObjectPostProcessor` should be as easy as possible, in order to allow less verbose and error-prone overrides or additional configuration of framework-provided components. Ideally, it would be possible with a lambda expression.

Given:

```java
OAuth2AuthorizationServerConfigurer<HttpSecurity> configurer =
	new OAuth2AuthorizationServerConfigurer<>(); // extends AbstractHttpConfigurer
```

Perhaps we could provide a lambda expression with the target type in the signature. For example:

```java
configurer
	.postProcess((OAuth2AuthorizationCodeRequestAuthenticationProvider authenticationProvider) ->
		authenticationProvider.setAuthenticationValidatorResolver(createDefaultAuthenticationValidatorResolver()));
```

Or the alternate syntax of specifying the target type in angle brackets:

```java
configurer
	.<OAuth2AuthorizationCodeRequestAuthenticationProvider>postProcess(authenticationProvider ->
		authenticationProvider.setAuthenticationValidatorResolver(createDefaultAuthenticationValidatorResolver()));
```

Another option could be to specify the type of the object as a class:

```java
configurer
	.postProcess(OAuth2AuthorizationCodeRequestAuthenticationProvider.class, authenticationProvider ->
		authenticationProvider.setAuthenticationValidatorResolver(createDefaultAuthenticationValidatorResolver()));
```

**Current Behavior**

Defining an `ObjectPostProcessor` cannot be done via a lambda expression, and requires a complex anonymous class. For example:

```java
configurer.withObjectPostProcessor(new ObjectPostProcessor<OAuth2AuthorizationCodeRequestAuthenticationProvider>() {
	@Override
	public <O extends OAuth2AuthorizationCodeRequestAuthenticationProvider> O postProcess(O object) {
		object.setAuthenticationValidatorResolver(createDefaultAuthenticationValidatorResolver());
		return object;
	}
});
```

However, the typical way is to override a component by re-defining the entire component. If it has dependencies also used by the framework, this further requires exposing things as beans that might have been defaulted. In [Spring Authorization Server](https://github.com/spring-projects/spring-authorization-server), this would be components such as `OAuth2AuthorizationService` and `OAuth2AuthorizationConsentService`, but there are many more such examples.

**Context**

[...]

PR #10331, opened 2021-09-29
Add support for ObjectPostProcessors from lambda expressions

sjohnr:
Closes gh-10320


rwinch (changes requested):
Thanks for the PR @sjohnr!

I'm not sure how this would work if you have more than one type of object that is getting processed without needing to perform instance of checks due to erasure. For example, if you add a Map to ConcreteHttpConfigurer that will also be `postProcess`, you will get a `ClassCastException` with the existing test because the argument type of the `Consumer` is lost and there is no Generic type for the class embedded in the lambda like an anonymous class.

Without the ability to postProcess multiple types without instanceof checks, I'm not too sure this enhancement is that desirable.

sjohnr:
Thanks @rwinch, I missed that in my brainstorming the idea. I've pushed another draft of the idea. It's not quite as nice syntactically, but may solve some of the problems with resolving the generics. This solution may point to moving this to another location. Or it may make it less desirable still. Any other thoughts/ideas?

rwinch:
Thanks for the revision. This does indeed solve the ClassCastException, but (as you possibly alluded to) I struggle considering this an improvement on the current code. This solution requires an anonymous inner class and a lambda vs just a single anonymous inner class.

sjohnr:
For sure, I'm with you. I'll take this back and keep working on it. Thanks!

Issue #10326, opened 2021-09-27
DefaultBearerTokenResolver triggers processing of multipart content


regmebaby:
Affected Artifact:
```
      <groupId>org.springframework.security</groupId>
      <artifactId>spring-security-oauth2-resource-server</artifactId>
      <version>5.4.7</version>
```

**Describe the bug**
**Given** that JWT Authentication is configured for a simple Spring Boot Web Application (`http.oauth2ResourceServer(OAuth2ResourceServerConfigurer::jwt);`) and I protect my endpoints (`http.authorizeRequests(requests -> requests.mvcMatchers("/api/**").authenticated());`)

**when** I upload a large file to a controller (`@RequestPart(name = "file") MultipartFile file`) without proving a Bearer Token or with invalid Bearer Token (e.g. using Postman)

**then** the large file is uploaded and processed by the application before I receive the response "401 unauthenticated" (e.g. after  7 Minutes, depending on the file size).

**Cause:** `DefaultBearerTokenResolver` calls `resolveFromRequestParameters` even though `isParameterTokenSupportedForRequest` will be false (as `allowFormEncodedBodyParameter` is false by default). The included `request.getParameterValues("access_token")` causes the consumption of the whole multipart content.

**Another Problem:** Access Tokens have short life time (e.g. 5 minutes). We found ourselves in the situation that the processing triggered by the resolver (`request.getParameterValues("access_token")`) takes that long that the token is always considered expired by the subsequent logic of the `BearerTokenAuthenticationFilter` / `JwtAuthenticationProvider`.

We might want to consider to change the code to resolve from request parameters (and check for multiple bearer tokens in the request) etc. only if `isParameterTokenSupportedForRequest` evaluates true to overcome the outlined two problems.

**To Reproduce**
See Description

**Expected behavior**
- The response "401 unauthenticated" should be returned quite fast. Not after minutes.
- The upload of a large file (upload taking longer than JWT life time) should succeed as the according security filter is passed before the file content is processed / uploaded / consumed.

**Sample**

No sample at hand. Please let me know whether it is necessary to provide one.

sjohnr:
@regmebaby, thanks for the report. Do you have any interest in submitting a PR to fix the issue in `DefaultBearerTokenResolver`? I think defaulting the `parameterToken` to null and only resolving it if `isParameterTokenSupportedForRequest(request)` returns `true` would work for your case as well as existing cases.

regmebaby:
> 
> 
[...]

PR #10340, opened 2021-10-04
Conditionally resolve bearer token from request parameters

pneuschwander:
### What this pull request changes

Before this commit, the token resolver logic unconditionally resolved
the request parameters to check whether a token is present to detect
whether multiple tokens are present in the request and reject those
requests as invalid.

This commit changes this behaviour to resolve the request parameters
only if parameter token is supported for the specific request
considering the already present logic to detect whether parameter
token is supported for the request.

After this commit, the request parameters will be resolved only if
parameter token is supported for the request. One can configure the
parameter token support through the existing boolean toggles of the
implementation.

### Why this change is proposed

gh-10326 describes the impact of resolving the request parameters for large requests.
Resolving the request parameters conditionally as proposed by this pull request should solve the issue.

Closes gh-10326


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolver.java
```diff
@@ -48,8 +48,9 @@
 
 	@Override
 	public String resolve(HttpServletRequest request) {
-		String authorizationHeaderToken = resolveFromAuthorizationHeader(request);
-		String parameterToken = resolveFromRequestParameters(request);
+		final String authorizationHeaderToken = resolveFromAuthorizationHeader(request);
+		final String parameterToken = isParameterTokenSupportedForRequest(request)
+				? resolveFromRequestParameters(request) : null;
```
sjohnr, line 53: I've discussed this change with the team, and I have a suggestion for improving this a bit more (which I didn't think of when we first discussed this issue). For context, here is the [spec](https://datatracker.ietf.org/doc/html/rfc6750#section-2.2) verbiage:

> 2.2.  Form-Encoded Body Parameter
> 
>    When sending the access token in the HTTP request entity-body, the
>    client adds the access token to the request-body using the
>    "access_token" parameter.  The client MUST NOT use this method unless
>    all of the following conditions are met:
> 
>    o  The HTTP request entity-header includes the "Content-Type" header
>       field set to "application/x-www-form-urlencoded".
> 
>    o  The entity-body follows the encoding requirements of the
>       "application/x-www-form-urlencoded" content-type as defined by
>       HTML 4.01 [W3C.REC-html401-19991224].
> 
>    o  The HTTP request entity-body is single-part.
> 
>    o  The content to be encoded in the entity-body MUST consist entirely
>       of ASCII [USASCII] characters.
> 
>    o  The HTTP request method is one for which the request-body has
>       defined semantics.  In particular, this means that the "GET"
>       method MUST NOT be used.

It seems that all of those points are applicable to this scenario. In your case, you have a multi-part file upload (presumably with `Content-Type: multipart/form-data;`), which need not be concerned about accepting the bearer token as a request parameter in the body. To me, this means that we can split our token validation into two parts:
* `isParameterTokenSupportedForRequest(request)`
* `isParameterTokenEnabledForRequest(request)`

Something like:

```java
	private boolean isParameterTokenSupportedForRequest(HttpServletRequest request) {
		return (("POST".equals(request.getMethod())
				&& MediaType.APPLICATION_FORM_URLENCODED_VALUE.equals(request.getContentType()))
				|| "GET".equals(request.getMethod()));
	}

	private boolean isParameterTokenEnabledForRequest(HttpServletRequest request) {
		return ((this.allowFormEncodedBodyParameter && "POST".equals(request.getMethod())
				&& MediaType.APPLICATION_FORM_URLENCODED_VALUE.equals(request.getContentType()))
				|| (this.allowUriQueryParameter && "GET".equals(request.getMethod())));
	}
```

If you now check the new `isParameterTokenEnabledForRequest()` before returning the `parameterToken`, I think we should be able to avoid changing existing tests. We would want to add new tests for the `Content-Type: multipart/form-data;` and `Content-Type: application/x-www-form-urlencoded` cases specifically.

While we could also add a flag to enable/disable this new check (the new version of `isParameterTokenSupportedForRequest()`), I think we are fairly safe in the 5.6 release to state that we are enhancing spec compliance and security, and not make it configurable.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
```diff
@@ -668,6 +670,23 @@ public void requestWhenBearerTokenResolverAllowsRequestBodyAndRequestContainsTwo
...
 
+	@Test
+	public void requestWhenBearerTokenResolverAllowsRequestBodyAndRequestContainsTwoTokensInRequestBodyThenInvalidRequest()
+			throws Exception {
+		this.spring.register(AllowBearerTokenInRequestBodyConfig.class, JwtDecoderConfig.class, BasicController.class)
+				.autowire();
+		JwtDecoder decoder = this.spring.getContext().getBean(JwtDecoder.class);
+		given(decoder.decode(anyString())).willReturn(JWT);
+		// @formatter:off
+		MockHttpServletRequestBuilder request = post("/authenticated")
```
sjohnr, line 681: With the below suggestion, you may need to add a `Content-Type` header to this test.

config/src/test/java/org/springframework/security/config/http/OAuth2ResourceServerBeanDefinitionParserTests.java
```diff
@@ -235,25 +235,26 @@ public void getWhenBearerTokenBeforeNotBeforeThenInvalidToken() throws Exception
 	}
 
 	@Test
-	public void getWhenBearerTokenInTwoPlacesThenInvalidRequest() throws Exception {
-		this.spring.configLocations(xml("JwkSetUri")).autowire();
+	public void getWhenBearerTokenInTwoPlacesAndQueryParameterNotSupportedThenAcceptsRequest() throws Exception {
```
sjohnr, line 238: Because this is a `GET`, I think we should be able to restore the old test if we add a new check to the `isParameterTokenSupportedForRequest()` method. See below.

config/src/test/java/org/springframework/security/config/http/OAuth2ResourceServerBeanDefinitionParserTests.java
```diff
@@ -235,25 +235,26 @@ public void getWhenBearerTokenBeforeNotBeforeThenInvalidToken() throws Exception
...
-				.andExpect(status().isBadRequest())
-				.andExpect(invalidRequestHeader("Found multiple bearer tokens in the request"));
+		this.mvc.perform(get("/").header("Authorization", "Bearer " + token).param("access_token", "token"))
+				.andExpect(status().isNotFound());
 		// @formatter:on
 	}
 
 	@Test
-	public void getWhenBearerTokenInTwoParametersThenInvalidRequest() throws Exception {
+	public void getWhenBearerTokenInTwoParametersAndQueryParameterNotSupportedThenUnauthorized() throws Exception {
```
sjohnr, line 249: Because this is a `GET`, I think we should be able to restore the old test if we add a new check to the `isParameterTokenSupportedForRequest()` method. See below.

config/src/test/java/org/springframework/security/config/http/OAuth2ResourceServerBeanDefinitionParserTests.java
```diff
@@ -486,6 +487,20 @@ public void requestWhenBearerTokenResolverAllowsRequestBodyAndRequestContainsTwo
 		// @formatter:on
 	}
 
+	@Test
+	public void requestWhenBearerTokenResolverAllowsRequestBodyAndRequestContainsTwoTokensInRequestBodyThenInvalidRequest()
+			throws Exception {
+		this.spring.configLocations(xml("MockJwtDecoder"), xml("AllowBearerTokenInBody")).autowire();
+		// @formatter:off
+		MockHttpServletRequestBuilder request = post("/authenticated")
```
sjohnr, line 495: With the below suggestion, you may need to add a `Content-Type` header to this test.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/server/ServerBearerTokenAuthenticationConverter.java
```diff
@@ -77,10 +78,7 @@ private String token(ServerHttpRequest request) {
 			}
 			return authorizationHeaderToken;
 		}
-		if (parameterToken != null && isParameterTokenSupportedForRequest(request)) {
-			return parameterToken;
-		}
-		return null;
+		return parameterToken;
```
sjohnr, line 81: I'd suggest the same here as above. :smile:

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolverTests.java
```diff
@@ -106,7 +106,8 @@ public void resolveWhenHeaderWithInvalidCharactersIsPresentThenAuthenticationExc
 	}
 
 	@Test
-	public void resolveWhenValidHeaderIsPresentTogetherWithFormParameterThenAuthenticationExceptionIsThrown() {
+	public void resolveWhenValidHeaderIsPresentTogetherWithFormParameterAndSupportedThenAuthenticationExceptionIsThrown() {
```
sjohnr, line 109: With the above suggestion, can we also restore the old test here?

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolverTests.java
```diff
@@ -117,7 +118,18 @@ public void resolveWhenValidHeaderIsPresentTogetherWithFormParameterThenAuthenti
 	}
 
 	@Test
-	public void resolveWhenValidHeaderIsPresentTogetherWithQueryParameterThenAuthenticationExceptionIsThrown() {
```
sjohnr, line 120: These tests seem to already line up well with my suggestion. The only thought I have is: Can we can keep the existing test names, and only have new names for truly new tests?

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/web/server/ServerBearerTokenAuthenticationConverterTests.java
```diff
@@ -155,7 +155,8 @@ public void resolveWhenHeaderWithInvalidCharactersIsPresentAndNotSubscribedThenN
 	}
 
 	@Test
-	public void resolveWhenValidHeaderIsPresentTogetherWithQueryParameterThenAuthenticationExceptionIsThrown() {
+	public void resolveWhenValidHeaderIsPresentTogetherWithQueryParameterThenAuthenticationExceptionIsThrown1() {
```
sjohnr, line 158: If this test is not overlapping with other tests, is it possible to change the name? In this test, perhaps also mention `AndQueryParameterSupported` as above?

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/web/server/ServerBearerTokenAuthenticationConverterTests.java
```diff
@@ -166,6 +167,16 @@ public void resolveWhenValidHeaderIsPresentTogetherWithQueryParameterThenAuthent
 		// @formatter:on
 	}
 
+	@Test
+	public void resolveWhenValidHeaderIsPresentTogetherWithQueryParameterThenAuthenticationExceptionIsThrown2() {
```
sjohnr, line 171: Similar to above, perhaps  add `AndQueryParameterNotSupported` here as well?

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/web/server/ServerBearerTokenAuthenticationConverterTests.java
```diff
@@ -204,7 +215,8 @@ public void resolveWhenQueryParameterIsPresentAndNotSupportedThenTokenIsNotResol
 	}
 
 	@Test
-	void resolveWhenQueryParameterHasMultipleAccessTokensThenOAuth2AuthenticationException() {
+	void resolveWhenQueryParameterHasMultipleAccessTokensThenOAuth2AuthenticationException1() {
```
sjohnr, line 218: Similarly, should we be more specific with the name here?

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/web/server/ServerBearerTokenAuthenticationConverterTests.java
```diff
@@ -217,6 +229,14 @@ void resolveWhenQueryParameterHasMultipleAccessTokensThenOAuth2AuthenticationExc
 
 	}
 
+	@Test
+	void resolveWhenQueryParameterHasMultipleAccessTokensThenOAuth2AuthenticationException2() {
```
sjohnr, line 233: Same thought as above.

sjohnr (changes requested):
Greetings, @pneuschwander! Great start to this PR, thanks for working on it!

I was thinking I wouldn't have too many comments, but because of the number of tests (wonderful problem to have), I found a few extra things for you to think about. Sorry about the flood of comments! (It's not an indication that there's anything wrong.) See below comments inline, including a fairly lengthy discussion on a suggestion I have.

Lastly, would you mind adding `// gh-10326` above each new unit test that is added in this PR?

Thanks!

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolver.java
sjohnr: Clarifying comment to above :wink: : "We should be able to avoid changing existing tests", with the exception of adding a `Content-Type` header to `POST` requests.

pneuschwander:
@sjohnr Thank you for your feedback. I modified the PR accordingly. Could you please review the changes and let me know whether there is more that should be done before we can merge? 🙂

Issue #10335, opened 2021-09-30
Add `saml2.ValidIssuers` parameters into SAML 2.0 Assertion Signature Validator


marcusdacoregio:
Related #10263

PR #10341, opened 2021-10-05
Add saml2.ValidIssuers parameter into SAML 2.0 Assertion Validators

marcusdacoregio:
Adds the saml2.ValidIssuers parameter into SAML 2.0 Assertion Validators

Closes gh-10335


jzheaux (changes requested):

PR #10345, opened 2021-10-05
Fix Antora cross-references that lead to other pages.

ahus1:
Also using AsciiDoc style listings instead of Markdown style listings, and using explicit section IDs on all cross-references.

**Background:** I'm maintaining the IntelliJ AsciiDoc plugin that also supports Antora. When I publish new releases, I sometimes pick some public AsciiDoc/Antora documentation and run the plugin's validations. 
Some minor left-overs of the conversion showed up, and I submitted a PR. 

Thanks for providing this this project including its docs!


jzheaux:
Perfect, thanks for the clarification, @ahus1. This is now merged into `main`.

Issue #8974, opened 2020-08-19
5.5.X only works with spring-security-5.4.xsd schema (XML-based config)


Emkas:
**Describe the bug**
`SecurityNamespaceHandler` throws `BeanDefinitionParsingException` when we use version 5.0.X (<= 5.0.18.RELEASE) or 5.1.X (<= 5.1.12.RELEASE) with versioned schema in XML-based configuration. On those branches only `spring-security-4.2.xsd` can be used (see also 'Workaround').

On branch 5.2 [this change](https://github.com/spring-projects/spring-security/commit/9b2af944fa925b73ec3a884b6456b3e06a3cece0#diff-8b27fe873c6ab63906cc4798e1a302e6) skipped directly from 4.2 to 5.2. No updates for 5.0 and 5.1 where provided. So handler is still looking for version 4.2:

```java
private boolean matchesVersionInternal(Element element) {
	String schemaLocation = element.getAttributeNS(
			"http://www.w3.org/2001/XMLSchema-instance", "schemaLocation");
	return schemaLocation.matches("(?m).*spring-security-4\\.2.*.xsd.*")
			|| schemaLocation.matches("(?m).*spring-security.xsd.*")
			|| !schemaLocation.matches("(?m).*spring-security.*");
}
```

**To Reproduce**
1. Use raported version of Spring Security.
2. Use XML-based config.
3. Use (the right) versioned schema (e.g. `spring-security-5.0.xsd`)

**Expected behavior**
Error free XML parsing, when file is valid against XSD with version compatible with Spring Security version (`spring-security-5.0.xsd` for 5.0.X and `spring-security-5.1.xsd` for 5.1.X). 

**Sample**
`ContextConfig`  with namespace like below isn't working with 5.0.18.RELEASE (Aug 05, 2020):

```xml
<beans xmlns="http://www.springframework.org/schema/beans"
	xmlns:security="http://www.springframework.org/schema/security"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:schemaLocation="
		http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd
		http://www.springframework.org/schema/security http://www.springframework.org/schema/security/spring-security-5.0.xsd">
</beans>
```

**Workaround**
Use unversioned schema.

Emkas:
I'm very sorry for any mistake in this report (if any). This is the first time I'm reporting something on Github :-)

[...]

PR #10348, opened 2021-10-06
SecurityNamespaceHandler: update schema version to 5.5

Emkas:
Closes gh-8974

Pull request is on 5.5.x branch since it cannot be cherry-picked to any other branch without modifications.

(My first small change in this project, so I will be grateful for all suggestions)


Emkas:
I can provide the same change on main, but since recent changes I'm having troubles with project build (missing plugin "org.asciidoctor.jvm.convert")

jzheaux:
Hi, @Emkas, thanks for the contribution!

Since this has come up a couple of times now, I wonder if there is a better way to do this so that it's not so tricky to remember.

At the very least, I think we should add a test to `SecurityNamespaceHandler` that confirms this behavior against the contents of `spring.schemas`. Better might be to have `SecurityNamespaceHandler` read `spring.schemas` to find the most recent schema version; then the code would refer to that instead.

Are you able to add one or both of these to this PR?

Additionally, please ensure a complete and passing build for your next push. It appears that there are some failing tests related to this PR.

Emkas:
> @jzheaux Do you want the same change on main (with 5.6 instead of 5.5) or would you do it on your own inside your team?

OK, I see that you've done some changes (split & getLast -> Pattern.match) so I'll assume that you'll merge this into main on you own :-)

PR #10357, opened 2021-10-09
ClaimAccessor#getClaimAsMap doesn't return null as documented

qavid:
Closes gh-10117

### DIFF
--- a/oauth2/oauth2-core/src/test/java/org/springframework/security/oauth2/core/ClaimAccessorTests.java
+++ b/oauth2/oauth2-core/src/test/java/org/springframework/security/oauth2/core/ClaimAccessorTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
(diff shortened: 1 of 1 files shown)


qavid:
@jzheaux, @jgrandja
I didn't wanted to change implementation of `getClaimAsXXX` methods prior discussion, but there is a problem with assertion in situations when claim value is `null` - `claimValue.getClass()` in this case throws NullPointerException.
See tests: `getClaimAsMapWhenValueIsNullThenThrowNullPointerException`, `getClaimAsStringListWhenValueIsNullThenNullPointerException`.

I suppose that correct implementation should return null if claim value is null and can't be converted to not null value.
`if (claimValue == null && convertedValue == null) return null;`

Same logic should be used in my second PR #10356 related to `ClaimAccessor`.

jzheaux:
Thanks for the updates, @qavid. I think it's okay to leave the tests in, but would you place them in a separate commit, please?

I think you make a good point about `getClaimAs` consistency. I think it would be good to add that point to the ticket you create.

PR #10368, opened 2021-10-13
if there is no mapped port, just use the original

asgh:
If there is no mapped port configured just use the original port, but switch to the https scheme. This is helpful for services running on http that are behind a proxy running on https - there is no need to setup mapping in that case, just keep the incoming port and don't change it.

If there is mapping setup (for example the default of 80 -> 443) then this commit changes nothing.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/LoginUrlAuthenticationEntryPoint.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/LoginUrlAuthenticationEntryPoint.java
@@ -158,15 +158,10 @@ public class LoginUrlAuthenticationEntryPoint implements AuthenticationEntryPoin
 		urlBuilder.setPathInfo(loginForm);
 		if (this.forceHttps && "http".equals(scheme)) {
 			Integer httpsPort = this.portMapper.lookupHttpsPort(serverPort);
-			if (httpsPort != null) {
-				// Overwrite scheme and port in the redirect URL
-				urlBuilder.setScheme("https");
-				urlBuilder.setPort(httpsPort);
-			}
-			else {
-				logger.warn(LogMessage.format("Unable to redirect to HTTPS as no port mapping found for HTTP port %s",
-						serverPort));
-			}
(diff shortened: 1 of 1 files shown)


jzheaux:
If a proxy is listening on 443 and the proxied application on 8000 (that is, not 80 and not 8080), then wouldn't this PR try and redirect to `https://{loginUrl}:8000`? I guess I'm not seeing how this addresses the issue you are pointing out.

I wonder if you would be better served using the [`ForwardedHeaderFilter`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/filter/ForwardedHeaderFilter.html) which allows the proxied application to compute redirects relative to the proxy. This seems better since the proxied application would then not compute port mappings at all.

I'm going to close this; however, if you still feel this change is needed, please consider sharing a sample application that demonstrates the interaction of your proxy and application. It may be easier to see at that point where Spring Security can be best simplified.

asgh:
@jzheaux  But it's not using the listen port (that would be even worse), it's using the URL of the request, so it "sees" the port as the browser sees it, and it should just stay on that port.

It should stay on that scheme as well, but there doesn't seem to be a way to know what the original one is. (And that's causing me quite a lot of trouble right now.)

Basically it's inconsistent - one part of the code just looks at what the request is, another part is trying to calculate things based on what running.

But in any case, since the port is the Request port, and not the Listen port, I think it makes sense to just leave the port as is and not try to port map it. (And if you do want to port map it, maybe include the scheme as well?)

Also, there are more oddities in the code - you are calling a portResolver, which in turn also calls a portMapper (but not the same portMapper - it has its own), but it throws away the port map result unless it's IE6. Also the port mapper has two functions: lookupHttpPort and lookupHttpsPort, and both do the exact same thing, except the first one uses an inefficient loop.

Personally I would throw away the entire port mapper stuff (it seems like a buggy relic of an earlier time), and just use the port of the incoming request and be done. If you do want a port mapper then it should really include the scheme, otherwise it's pretty useless.

Issue #10333, opened 2021-09-29
Resolve HttpSecurityDsl Package Tangle


jzheaux:
Resolve package tangle between `org.springframework.security.config.web` and `org.springframework.security.config.annotation.web`.

nor-ek:
Okay, so the task is to move all Dsl files from `org.springframework.security.config.web.servlet` that have dependencies to `org.springframework.security.config.annotation.web`?

You can assign me to that task. I'll do this on the branch and that will wait till `6.x`. I'm open to other tasks for `first-timers`.

jzheaux:
@nor-ek, the `6.0.x` branch is now available. Thanks for taking a look.

nor-ek:
@jzheaux sorry for the mess above. Finally, I've got [Pull Request](https://github.com/spring-projects/spring-security/pull/10474).

Issue #1033, opened 2008-04-15
SEC-774: JdbcDaoImpl no longer allows extending behavior


spring-projects-issues:
["Manuel Dominguez Sarmiento":https://jira.spring.io/secure/ViewProfile.jspa?name=mads1980](Migrated from ["SEC-774":https://jira.spring.io/browse/SEC-774?redirect=false]) said:

Methods/properties that used to be protected in JdbcDaoImpl are now private. Mapping strategies can no longer be customized. We attempted migration to Spring Security 2.0 from Acegi Security 1.0.6 but could not go forward because of this, since we have customized code relying on this feature.

....
    private MappingSqlQuery authoritiesByUsernameMapping;
    private MappingSqlQuery groupAuthoritiesByUsernameMapping;
    private MappingSqlQuery usersByUsernameMapping;
...

```
/**
 * Extension point to allow other MappingSqlQuery objects to be substituted in a subclass
 */
private void initMappingSqlQueries() {
    ...
}
```

PR #10429, opened 2021-10-21
Gh-10333 move HttpSecurityDsl to another package

nor-ek:
According to [gh-1033](https://github.com/spring-projects/spring-security/issues/10333) issue I've moved files from `org.springframework.security.config.web.servlet` to `org.springframework.security.config.annotation.web.configurers` in `src/main/kotlin`. 

Firstly files used to be moved to `org.springframework.security.config.annotation.web` but I've noticed that in `src/main/java` similar files are in `.configurers` the subpackage.


jzheaux:
Hi, @nor-ek. `6.0.x` is now `main`. Would you please take your branch, rebase off of `main` and then submit a new PR?

Also, regarding your question about packages, you can confirm this by doing the following command:

```
gw check && gw s101 | grep feedbackDependencies
```

The result will contain a line like this:

```
1-Nov-2021 09:50:43,309 [main] INFO  - OKAY! feedbackDependencies didn't change! Still at 28 (fail-on-feedback-dependencies)
```

The `s101` command analyzes the codebase for package tangles. If the result says "Still at 28", then the issue is not correctly addressed yet.

Issue #10425, opened 2021-10-21
Allow custom OAuth2ErrorHttpMessageConverter with OAuth2ErrorResponseErrorHandler


khamlaoui:
**Expected Behavior**

When we get errors from IDP (in our case its Okta), we should get a correctly formed OAuth2Error with the error code and description.

**Current Behavior**

The root cause error is lost, because the error conversion is not compliant with **OKTA** error response.
**Okta** error example:
`{
  "errorCode": "invalid_client",
  "errorSummary": "Invalid value for 'client_id' parameter.",
  "errorLink": "invalid_client",
  "errorId": "oaeUy44O3E3Q6mR4ZIcy50gKg",
  "errorCauses": []
}`
The **OAuth2ErrorConverter** try to map mandatory parameter with name **error** however with Okta this attribute is named with **errorCode**

At the end, we lost the Okta error, and we get **IllegalArgumentException**  raised by the constructor of **OAuth2Error** 

A quick fix, could be allowing setter for **OAuth2ErrorHttpMessageConverter** in **OAuth2ErrorResponseErrorHandler** or add a constructor with **OAuth2ErrorHttpMessageConverter** .

Thanks

PR #10432, opened 2021-10-21
Allow custom OAuth2ErrorHttpMessageConverter with OAuth2ErrorResponse…

khamlaoui:
**Expected Behavio**r

When we get errors from IDP (in our case its Okta), we should get a correctly formed OAuth2Error with the error code and description.

**Current Behavior**

The root cause error is lost, because the error conversion is not compliant with OKTA error response.
Okta error example:
{ "errorCode": "invalid_client", "errorSummary": "Invalid value for 'client_id' parameter.", "errorLink": "invalid_client", "errorId": "oaeUy44O3E3Q6mR4ZIcy50gKg", "errorCauses": [] }
The OAuth2ErrorConverter try to map mandatory parameter with name error however with Okta this attribute is named with errorCode

At the end, we lost the Okta error, and we get IllegalArgumentException raised by the constructor of OAuth2Error

A quick fix, could be allowing setter for OAuth2ErrorHttpMessageConverter in OAuth2ErrorResponseErrorHandler or add a constructor with OAuth2ErrorHttpMessageConverter .

Closes gh-10425

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/http/OAuth2ErrorResponseErrorHandler.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/http/OAuth2ErrorResponseErrorHandler.java
@@ -23,10 +23,12 @@ import com.nimbusds.oauth2.sdk.token.BearerTokenError;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.client.ClientHttpResponse;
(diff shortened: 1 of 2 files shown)


sjohnr:
> @sjohnr do i have to change the since to 5.7 ?

If you could, that would be great. Actually, looks like you have already, thanks!

I think Rob's point is that we should stick to our stated policy on semantic versioning, not so much that there's a specific risk. Thanks for the contribution @khamlaoui! Appreciate the fast turnaround even so.

Issue #7411, opened 2019-09-09
Update the default strength for password encoders


larsgrefer:
I propose to increase the default strength for some password encoders.

The BCrypt encoder currently uses `10` as default strength, with which it only takes ~80ms to encode a password (on my Early 2015 MacBook Pro). I think the default should be increased to (at least) 12.

The Argon2 encoder only needs ~40ms, so its defaults should be increased too.

The SCrypt encoder needs ~80ms with its default parameters.

larsgrefer:
The Spring Security Reference Documentation says:
> It is recommended that the "work factor" be tuned to take about 1 second to verify a password on your system.

jzheaux:
Thanks, @larsgrefer for checking on this.

@rwinch do we have a hardware profile that we target when determining when to increase this? My concern would be if we bump the number up to early, lots of folks suddenly see password hashing taking quite a bit longer than 1 second.

larsgrefer:
MacBook Pro (2021, M1 Pro):

```
BCryptPasswordEncoder : 73ms
SCryptPasswordEncoder : 28ms
Argon2PasswordEncoder : 6ms
Pbkdf2PasswordEncoder : 295ms
```

PR #10447, opened 2021-10-28
Bump default BCrypt strength to 12

larsgrefer:
see #7411


jzheaux:
Thanks for the PR, @larsgrefer. Based on https://github.com/OWASP/CheatSheetSeries/issues/601 it seems like OWASP continues to recommend [10 as the minimum work factor](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Password_Storage_Cheat_Sheet.md#introduction).

I lean towards sticking with OWASP's recommendations - why do you think it should be increased to 12?

larsgrefer:
In my opinion, 10 is way too fast for any near-modern hardware. See https://github.com/spring-projects/spring-security/issues/7411#issuecomment-531031619 for some numbers.

The Spring Security documentation itself states:

>It is recommended that the "work factor" be tuned to take about 1 second to verify a password on your system.

With below 100ms, 10 is nearly two orders of magnitude too fast.

Also the current default of 10 was not adjusted since 8565116f203b9c6186b044bc655c9d5d5b2e6450 10 years ago.

With the concerns you mentioned in https://github.com/spring-projects/spring-security/issues/7411#issuecomment-529945087 in mind I'd recommend to bump it "only" to 12 for Spring Security 5.x and then bump it to 13 for 6.x

Sc00bz:
Just had a look and you should be more worried about increasing the defaults for Argon2, scrypt, and PBKDF2. Since those don't even meet the bare minimum settings. Also there's a bug in PBKDF2's defaults.

Argon2's setting are `m=4 MiB, t=3, p=1` either set it to 10 MiB or 7 iterations.
scrypt's settings are `N=2^14, r=8, p=1` (16 MiB) either set `N` to 2^16 (64 MiB) or `p` to 4.
PBKDF2 settings are SHA1, 185k iterations, 256 bits of output. This should be changed to SHA512 and 120k iterations.
bcrypt setting is cost 10. Cost 9 is actually quite strong and is similar to PBKDF2-SHA512 with 230k iterations in strength.

Anyway those are just the bare minimums.

Sc00bz:
PBKDF2's bug reported here: #10489

jzheaux:
@larsgrefer, I appreciate the research you've done here. I'd like to adhere to the [OWASP password recommendations](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Password_Storage_Cheat_Sheet.md#introduction) since they are a single reliable source we can use for decision making. If 10 is too low, let's start with OWASP to get their recommendation increased to 12.

larsgrefer:
@jzheaux OWASP explicitly asked Spring Security to use a strength of 12: https://github.com/OWASP/CheatSheetSeries/issues/601#issuecomment-966578036

PR #10474, opened 2021-11-05
move HttpSecurityDsl and common files to annotation package

nor-ek:
According to gh-10333 issue I've moved files from `org.springframework.security.config.web.servlet` to `org.springframework.security.config.annotation.web` in `src/main/kotlin`.

Result after running `gradle check && gradle s101 | grep feedbackDependencies` command
> 5-Nov-2021 11:53:32,153 [main] INFO  - GREAT! feedbackDependencies has decreased by 28.0 to 0 (fail-on-feedback-dependencies)


config/src/test/kotlin/org/springframework/security/config/annotation/web/RememberMeDslTests.kt
```diff
@@ -14,32 +14,26 @@
  * limitations under the License.
  */
 
-package org.springframework.security.config.web.servlet
+package org.springframework.security.config.annotation.web
 
-import io.mockk.Called
```
eleftherias, line 19: Nice catch finding these unused imports @nor-ek!
Since this change is not related to resolving package tangles, please move it to a separate commit.
This will ensure the commit message accurately describes the changes.

eleftherias (changes requested):
Thanks for the PR @nor-ek!
I have left one comment inline.

config/src/test/kotlin/org/springframework/security/config/annotation/web/RememberMeDslTests.kt
nor-ek: @eleftherias please check it out. I've noticed more changes like that and packed them into `Import cleanup` commit. 

I've also cleaned up titles and messages.

jzheaux:
Thanks, @nor-ek! This was merged into `main` via 4f186f2c1fb520dc95fa5307db9dde299b642e3a.

Also, I re-established the package tangle baseline 43749058011955ba9d44c32a9b4f84c7b93220e7.

PR #10483, opened 2021-11-10
Add GitHubReleasePlugin with createGitHubRelease task

sjohnr:
Closes gh-10457

### DIFF
--- a/buildSrc/src/main/java/org/springframework/gradle/github/milestones/GitHubMilestonePlugin.java
+++ b/buildSrc/src/main/java/org/springframework/gradle/github/milestones/GitHubMilestonePlugin.java
@@ -29,7 +29,7 @@ public class GitHubMilestonePlugin implements Plugin<Project> {
 				githubCheckMilestoneHasNoOpenIssues.setGroup("Release");
 				githubCheckMilestoneHasNoOpenIssues.setDescription("Checks if there are any open issues for the specified repository and milestone");
 				githubCheckMilestoneHasNoOpenIssues.setMilestoneTitle((String) project.findProperty("nextVersion"));
-				if (project.hasProperty("githubAccessToken")) {
+				if (project.hasProperty("gitHubAccessToken")) {
 					githubCheckMilestoneHasNoOpenIssues.setGitHubAccessToken((String) project.findProperty("gitHubAccessToken"));
 				}
 			}
--- a/buildSrc/src/main/java/org/springframework/gradle/github/release/CreateGitHubReleaseTask.java
+++ b/buildSrc/src/main/java/org/springframework/gradle/github/release/CreateGitHubReleaseTask.java
@@ -1,0 +1,130 @@
+/*
+ * Copyright 2002-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.gradle.github.release;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+import org.gradle.api.Action;
+import org.gradle.api.DefaultTask;
+import org.gradle.api.Project;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.Optional;
+import org.gradle.api.tasks.TaskAction;
+
+import org.springframework.gradle.github.RepositoryRef;
+import org.springframework.gradle.github.changelog.GitHubChangelogPlugin;
+
+/**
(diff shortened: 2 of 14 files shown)


buildSrc/src/main/java/org/springframework/gradle/github/release/GitHubReleasePlugin.java
```diff
@@ -0,0 +1,41 @@
...
+	@Override
+	public void apply(Project project) {
+		project.getTasks().register("createGitHubRelease", CreateGitHubReleaseTask.class, new Action<CreateGitHubReleaseTask>() {
+			@Override
+			public void execute(CreateGitHubReleaseTask createGitHubRelease) {
+				createGitHubRelease.setGroup("Release");
+				createGitHubRelease.setDescription("Create a github release");
+				createGitHubRelease.dependsOn("generateChangelog");
+
+				createGitHubRelease.setGitHubAccessToken((String) project.findProperty("gitHubAccessToken"));
```
eleftherias, line 36: I wonder if we should check if `gitHubAccessToken` exists before using it, similar to `GitHubMilestonePlugin`.

buildSrc/src/main/java/org/springframework/gradle/github/release/GitHubReleaseApi.java
```diff
@@ -0,0 +1,87 @@
...
+	 */
+	public void publishRelease(RepositoryRef repository, Release release) {
+		String url = this.baseUrl + "/repos/" + repository.getOwner() + "/" + repository.getName() + "/releases";
+		String json = this.gson.toJson(release);
+		RequestBody body = RequestBody.create(MediaType.parse("application/json"), json);
+		Request request = new Request.Builder().url(url).post(body).build();
+		try {
+			Response response = this.httpClient.newCall(request).execute();
+			if (!response.isSuccessful()) {
+				throw new RuntimeException("Could not create release " + release + ". Got response " + response);
```
eleftherias, line 64: It may be more readable to use `release.getName()` in the exception messages instead of printing the entire object.

buildSrc/src/main/java/org/springframework/gradle/github/release/CreateGitHubReleaseTask.java
```diff
@@ -0,0 +1,93 @@
...
+				.name(this.version)
+				.body(body)
+				.preRelease(false)
+				.build();
+		github.publishRelease(this.repository, release);
+	}
+
+	private String readReleaseNotes() {
+		Project project = getProject();
+		File inputFile = project.file(Paths.get(project.getBuildDir().getPath(), "changelog/release-notes.md"));
```
eleftherias, line 58: Rather than hardcoding "changelog/release-notes.md" here, we can have a public constant in `GitHubChangelogPlugin`, for example `public static final String RELEASE_NOTES_PATH = "changelog/release-notes.md";` and access it here as `GitHubChangelogPlugin.RELEASE_NOTES_PATH`. (the variable name could be anything)

buildSrc/src/main/java/org/springframework/gradle/github/release/RepositoryRef.java
```diff
@@ -0,0 +1,57 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.gradle.github.release;
+
+/**
+ * Steve Riesenberg
+ */
+public class RepositoryRef {
```
eleftherias, line 22: This looks like it's the same class as `org.springframework.gradle.github.milestones.RepositoryRef`. Could we extract it into `org.springframework.gradle.github` and reuse it in both `milestones` and `release` packages?

eleftherias (commented):
Looks good @sjohnr! I have left some suggestions inline.

buildSrc/src/main/java/org/springframework/gradle/github/release/GitHubReleasePlugin.java
sjohnr: It looks like the GitHubMilestonePlugin does not need an access token to work, since it's accessing public data. I will explore the right way to fail with a meaningful error message in this plugin, since it requires access token.

sjohnr:
@eleftherias, I have updated the PR, including adding additional options `createRelease` and `branch`, and made `gitHubAccessToken` optional with a check, so that we can test with dry runs. This task also tags the release, so it accomplishes both #10456 and #10457.

I've tested it and it's working well, so we should be able to begin work on automating releases in a pipeline. Let me know if you spot anything else.

buildSrc/src/main/java/org/springframework/gradle/github/milestones/GitHubMilestonePlugin.java
```diff
@@ -29,7 +29,7 @@ public void execute(GitHubMilestoneHasNoOpenIssuesTask githubCheckMilestoneHasNo
 				githubCheckMilestoneHasNoOpenIssues.setGroup("Release");
 				githubCheckMilestoneHasNoOpenIssues.setDescription("Checks if there are any open issues for the specified repository and milestone");
 				githubCheckMilestoneHasNoOpenIssues.setMilestoneTitle((String) project.findProperty("nextVersion"));
-				if (project.hasProperty("githubAccessToken")) {
```
eleftherias, line 32: I think this should be a separate commit since it's independent from adding the new plugin.

eleftherias (changes requested):
I left one comment @sjohnr. In my option you can merge this after that change. I don't think another review is needed.

Issue #10337, opened 2021-10-01
Invalid_request failures in JwtTokenValidators are always turned into invalid_token errors


jason076:
**Describe the bug**
Returning any failure in a `OAuth2TokenValidator` `validate` function always results in a `InvalidBearerTokenException`  with the error code `BearerTokenErrorCodes.INVALID_TOKEN` and returns a 401. The Error handling does not respect the Error code the validate function returns. This leads to  `OAuth2ErrorCodes.INVALID_REQUEST ` get converted to `BearerTokenErrorCodes.INVALID_TOKEN`.

**To Reproduce**
Return `OAuth2ErrorCodes.INVALID_REQUEST` in `OAuth2TokenValidator` `validate`. This was already the case in an bug I reported earlier: https://github.com/spring-projects/spring-security/issues/10319 

**Expected behavior**
If I return `OAuth2ErrorCodes.INVALID_REQUEST ` failure from  `OAuth2TokenValidator` `validate`  I expect an `OAuth2AuthenticationException` with the `BeareErrorCode` `BearerTokenErrorCodes.INVALID_REQUEST` and the status code 400.

**Sample**
The problematic line in the sourcode is the following https://github.com/spring-projects/spring-security/blob/8c74d6cea5836def47102dfa7f3b7687992f95aa/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationProvider.java#L101

Each `BadJwtException` is converted to `InvalidBearerTokenException` with the error code `BearerTokenErrorCodes.INVALID_TOKEN`

jzheaux:
Hi, @jason076, thanks for the report.

> If I return `OAuth2ErrorCodes.INVALID_REQUEST` failure from `OAuth2TokenValidator` validate I expect an `OAuth2AuthenticationException` with the `BeareErrorCode BearerTokenErrorCodes.INVALID_TOKEN`

Why would you want the error to change? [The errors are defined](https://datatracker.ietf.org/doc/html/rfc6750#section-3.1) as follows (emphasis *mine*):

> invalid_request
         The request is missing a required parameter, includes an
         unsupported parameter or parameter value, repeats the same
         parameter, uses more than one method for including an access
         token, or is otherwise malformed.  The resource server SHOULD
         respond with the HTTP 400 (Bad Request) status code.

> invalid_token
         The access token provided is *expired, revoked, malformed, or
[...]

PR #10500, opened 2021-11-12
Fix JwtClaimValidator error type

jason076:
Previously JWTValidator could return any OAuth2Error like invalid_request. But validators were only allowed
to return invalid_token errors. Therefore Validators are now enforced to return invalid_token errors
by introducing the OAuth2InvalidTokenError class.

Closes gh-10337

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtClaimValidator.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtClaimValidator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -54,7 +54,7 @@ public final class JwtClaimValidator<T> implements OAuth2TokenValidator<Jwt> {
 		Assert.notNull(test, "test can not be null");
 		this.claim = claim;
 		this.test = test;
-		this.error = new OAuth2Error(OAuth2ErrorCodes.INVALID_REQUEST, "The " + this.claim + " claim is not valid",
+		this.error = new OAuth2Error(OAuth2ErrorCodes.INVALID_TOKEN, "The " + this.claim + " claim is not valid",
 				"https://tools.ietf.org/html/rfc6750#section-3.1");
 	}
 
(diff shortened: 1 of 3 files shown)


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcIdTokenValidator.java
```diff
@@ -146,8 +146,8 @@ public void setClock(Clock clock) {
 		this.clock = clock;
 	}
 
-	private static OAuth2Error invalidIdToken(Map<String, Object> invalidClaims) {
-		return new OAuth2Error("invalid_id_token", "The ID Token contains invalid claims: " + invalidClaims,
+	private static OAuth2InvalidTokenError invalidIdToken(Map<String, Object> invalidClaims) {
```
jzheaux, line 149: Since client error codes aren't specified in the OIDC spec, I think it's fine to have this be `invalid_id_token`.

What are your reasons for wanting to change OIDC client error reporting?

jzheaux (changes requested):
Thanks for the PR, @jason076! I've left my feedback inline.

Additionally, please double-check that you have run `./gradlew format check`. It appears that the build is failing on some checkstyle requirements.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2InvalidTokenError.java
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * <p>
+ * At a minimum, an error response will contain an error code. The error code will always
+ * be "invalid_token"
+ *
+ * @author Jonas Erbe
+ * @since 5.0
+ * @see OAuth2ErrorCodes
+ */
+public class OAuth2InvalidTokenError extends OAuth2Error {
```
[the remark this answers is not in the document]
jason076: Personally I favor stronger typing over redundance. The new error class makes returning the wrong error impossible since the compiler fails. But you are right, the public API can't be changed, so I will remove the class.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcIdTokenValidator.java
jason076: This was a missunderstanding on my side, I will revert that change.

jason076:
Can you trigger the pipeline again please? I have fixed the style issues.

jzheaux (approved):

jzheaux:
Thanks, @jason076! This is now merged into `main`.

Issue #10334, opened 2021-09-30
Be able to configure RestTemplate of OidcIdTokenDecoderFactory


H3llK33p3r:
**Expected Behavior**

In our daily deployment our services are behind http proxy servers. 

Due to this, we are not able to use the open id connect feature in spring-security-oauth2 because OidcIdTokenDecoderFactory does not allow to customize the restOperations used. 

Like the customizer of oauth2Login, it will be great to be able to set a restOperation in the OidcIdTokenDecoderFactory in order to use it in the private method org.springframework.security.oauth2.client.oidc.authentication.OidcIdTokenDecoderFactory#buildDecoder when NimbusJwtDecoder builder is used. 

Allowing this we will able to create our own managed bean for JWT decoder that can be used by  a WebSecurityConfigurerAdapter

```
	@Override
	protected void configure(HttpSecurity http) throws Exception {
		http
				.oauth2Login(customizer -> {
								.......
				});
	}
```

**Current Behavior**

No rest template customization is available and we are not able to use spring security in order to validate JWT token.

jzheaux:
Hi, @H3llK33p3r, do you have multiple client registrations? If not, you can simply do:

```java
@Bean 
JwtDecoderFactory<ClientRegistration> jwtDecoderFactory() {
    JwtDecoder jwtDecoder = NimbusJwtDecoder.withJwkSetUri(jwkSetUri)
            .restOperations(restOperations).build();
    return (clientRegistration) -> jwtDecoder;
}
```

Or, if you are multi-tenant, then you can instead do something like the following:

```java
@Component
public class MyJwtDecoderFactory implements JwtDecoderFactory<ClientRegistration> {
    private final Map<String, JwtDecoder> jwtDecoders = new ConcurrentHashMap<>();

    @Override 
    public JwtDecoder createDecoder(ClientRegistration clientRegistration) {
        return this.jwtDecoders.computeIfAbsent(clientRegistration.getRegistrationId(),
                (r) -> NimbusJwtDecoder
                        .withJwkSetUri(clientRegistration.getProviderDetails().getJwkSetUri())
                        .restOperations(restOperations)
                        .build();
    }
}
```

Would either of those work for your situation?

Either way, if you still feel that the `RestOperations` should be exposed, please feel free to share your thoughts on https://github.com/spring-projects/spring-security/issues/8882. That ticket seems to be the one where something like this would be addressed.

H3llK33p3r:
[...]

PR #10513, opened 2021-11-16
make RestOperations configurable in OidcIdTokenDecoderFactory

### LINKED ISSUE #10512: Allow RestOperations to be configured for OidcIdTokenDecoderFactory
**Expected Behavior**

`OidcIdTokenDecoderFactory` when used with a JWKs doesn't provide any option to configure the `RestOperations` used by the `NimbusJwtDecoder` when fetching the JWKs. I expected to be able to provide an instance of my own.

**Current Behavior**

`RestOperations` uses the `hard-coded` default of a `new RestTemplate()` created in `JwkSetUriJwtDecoderBuilder` without any user-settable option.

**Context**

My concrete context was to selectively set a proxy, but I can easily image user reasons to modify the `RestOperations` being used. I ended up providing a `JwtDecoderFactory` of my own that more or less contained a copy of `OidcIdTokenDecoderFactory`.

I will create a PR modelled after the way `RestOperations` are configurable in `DefaultAuthorizationCodeTokenResponseClient`.

bodewig:
This is a simple fix for #10512 which allows the `RestOperations` used when fetching a JWKs to be configured externally.

closes #10512

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcIdTokenDecoderFactory.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcIdTokenDecoderFactory.java
@@ -49,11 +49,18 @@ import org.springframework.security.oauth2.jwt.JwtTimestampValidator;
 import org.springframework.security.oauth2.jwt.NimbusJwtDecoder;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
+import org.springframework.web.client.RestOperations;
+import org.springframework.web.client.RestTemplate;
 
 /**
  * A {@link JwtDecoderFactory factory} that provides a {@link JwtDecoder} used for
  * {@link OidcIdToken} signature verification. The provided {@link JwtDecoder} is
  * associated to a specific {@link ClientRegistration}.
+ *
+ * <p>
+ * This implementation may use a {@link RestOperations} when requesting a key set at the
+ * Authorization Server's JWKS Endpoint.
+ * </p>
  *
  * @author Joe Grandja
  * @author Rafael Dominguez
@@ -80,6 +87,8 @@ public final class OidcIdTokenDecoderFactory implements JwtDecoderFactory<Client
 			createDefaultClaimTypeConverters());
 
 	private final Map<String, JwtDecoder> jwtDecoders = new ConcurrentHashMap<>();
+
+	private RestOperations restOperations = new RestTemplate();
 
 	private Function<ClientRegistration, OAuth2TokenValidator<Jwt>> jwtValidatorFactory = new DefaultOidcIdTokenValidatorFactory();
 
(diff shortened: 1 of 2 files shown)


bodewig:
I've closed #10512 as it is a duplicate of #10334 - which probably means I should close this as well.

jgrandja:
@bodewig I was just reviewing this when you closed it.

It doesn't make sense from a design standpoint to expose `OidcIdTokenDecoderFactory.setRestOperations()`, since not all `JwtDecoder`'s would require a `RestOperations`.

> I ended up providing a `JwtDecoderFactory` of my own that more or less contained a copy of `OidcIdTokenDecoderFactory`

This approach makes sense and is the reason why a `JwtDecoderFactory` `@Bean` was made discoverable to enable customization.

A stripped down version of `OidcIdTokenDecoderFactory` that only supports `jwkSetUri` would look like this:

```java
public final class CustomOidcIdTokenDecoderFactory implements JwtDecoderFactory<ClientRegistration> {

	private static final String MISSING_SIGNATURE_VERIFIER_ERROR_CODE = "missing_signature_verifier";

	private static final ClaimTypeConverter DEFAULT_CLAIM_TYPE_CONVERTER = new ClaimTypeConverter(
			OidcIdTokenDecoderFactory.createDefaultClaimTypeConverters());

	private final Map<String, JwtDecoder> jwtDecoders = new ConcurrentHashMap<>();

	private final Function<ClientRegistration, OAuth2TokenValidator<Jwt>> jwtValidatorFactory =
			(clientRegistration) -> new DelegatingOAuth2TokenValidator<>(
					new JwtTimestampValidator(),
					new OidcIdTokenValidator(clientRegistration));

	private Function<ClientRegistration, Converter<Map<String, Object>, Map<String, Object>>> claimTypeConverterFactory =
			(clientRegistration) -> DEFAULT_CLAIM_TYPE_CONVERTER;

	private RestOperations restOperations = new RestTemplate();


	@Override
	public JwtDecoder createDecoder(ClientRegistration clientRegistration) {
		Assert.notNull(clientRegistration, "clientRegistration cannot be null");
		return this.jwtDecoders.computeIfAbsent(clientRegistration.getRegistrationId(), (key) -> {
			NimbusJwtDecoder jwtDecoder = buildDecoder(clientRegistration);
			jwtDecoder.setJwtValidator(this.jwtValidatorFactory.apply(clientRegistration));
			jwtDecoder.setClaimSetConverter(this.claimTypeConverterFactory.apply(clientRegistration));
			return jwtDecoder;
		});
	}

	private NimbusJwtDecoder buildDecoder(ClientRegistration clientRegistration) {
		String jwkSetUri = clientRegistration.getProviderDetails().getJwkSetUri();
		if (!StringUtils.hasText(jwkSetUri)) {
			OAuth2Error oauth2Error = new OAuth2Error(MISSING_SIGNATURE_VERIFIER_ERROR_CODE,
					"Failed to find a Signature Verifier for Client Registration: '"
							+ clientRegistration.getRegistrationId()
							+ "'. Check to ensure you have configured the JwkSet URI.",
					null);
			throw new OAuth2AuthenticationException(oauth2Error, oauth2Error.toString());
		}
		return NimbusJwtDecoder.withJwkSetUri(jwkSetUri).restOperations(this.restOperations).build();
	}

	public void setRestOperations(RestOperations restOperations) {
	}

}
```

IMO, this is a valid custom implementation with the added benefit of the same internal components of `OidcIdTokenDecoderFactory` being reused.

bodewig:
@jgrandja thank you. I assumed this PR would get declined for the same reasons the earlier issue has been closed.

I do agree not all code paths in `OidcIdTokenDecoderFactory` require rest operations, obviously, which is why "my" implementation delegates to a `OidcIdTokenDecoderFactory` instance in the non-JWKS cases. Happy to see your implementation looks extremely similar to mine :-)

I was hoping for an easier to maintain solution to the "we are forced to use a proxy" use-case, but can certainly live with the additional custom code.

PR #10541, opened 2021-11-22
Update MapReactiveUserDetailsService.java

lizardcoder9999:
Remove 	Assert.notEmpty(users, "users cannot be null or empty"); because when configuring a application for the first time there wont be any users for it to find which causes errors. Having default users can be a security concern for a application.

### DIFF
--- a/core/src/main/java/org/springframework/security/core/userdetails/MapReactiveUserDetailsService.java
+++ b/core/src/main/java/org/springframework/security/core/userdetails/MapReactiveUserDetailsService.java
@@ -56,7 +56,6 @@ public class MapReactiveUserDetailsService implements ReactiveUserDetailsService
 	 * @param users the {@link UserDetails} to use
 	 */
 	public MapReactiveUserDetailsService(Collection<UserDetails> users) {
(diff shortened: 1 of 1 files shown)


lizardcoder9999:
what does waiting-for-triage tag mean? sorry if its a silly question

sjohnr:
Hi @lizardcoder9999, thanks for your interest in the project!

The waiting-for-triage label means it's waiting for a team member to look at.

If I understand correctly, you're aiming for the "first time developer" experience. I believe in that case (possibly the case you're running into), it is even more advantageous for you to be warned against creating an instance with no users, as you will be completely unable to log into the application otherwise. And because you cannot add users to the `MapReactiveUserDetailsService` after it has been created, there would be no purpose of passing an empty list.

**Note:** The `MapReactiveUserDetailsService` is useful for demo applications, but is designed to be replaced by your own implementation of `ReactiveUserDetailsService`, usually backed by a database.

I am going to close this PR for now, as the above will hopefully explain why it works the way it does. If I have misunderstood anything, please let me know and we can discuss it further.

lizardcoder9999:
thanks i see now

PR #10546, opened 2021-11-23
Fix return type for NoOpPasswordEncoder bean in documentation

hpoettker:
Fixes the return type of the bean factory method for a `NoOpPasswordEncoder` in the reference documentation.

The return type must be `PasswordEncoder` as that is the return type of `NoOpPasswordEncoder::getInstance`. The code snippet for Kotlin is correct.


marcusdacoregio:
Thanks @hpoettker. This is now merged into `main`.

Issue #10548, opened 2021-11-23
MappedJwtClaimSetConverter calls Converter#convert with null


mitasov-ra:
**Describe the bug**
Documentation of `org.springframework.core.convert.converter.Converter#convert` clearly says "never null":
```
* @param source the source object to convert, which must be an instance of {@code S} (never {@code null})
```

But [lines 156-158][lines] of `org.springframework.security.oauth2.jwt.MappedJwtClaimSetConverter` haven't got null-checks in them:

```java
Object claim = claims.get(claimName); // claim is nullable
Object mappedClaim = converter.convert(claim); // convert accepts non-null !!!
mappedClaims.compute(claimName, (key, value) -> mappedClaim);
```

This results in `NullPointerException` when accepting JWT without particular claims (which of them that haven't got custom converters for them).

What's interesting is that default converters (e.g. [convertInstant]) has null-checks INSIDE of them:
```java
private static Instant convertInstant(Object source) {
    if (source == null) {
        return null;
    }
// ... snip ...
}
```

**To Reproduce**
Set custom ClaimSetConverter in `NimbusJwtDecoder`:
```java
decoder.setClaimSetConverter(MappedJwtClaimSetConverter
    .withDefaults(Map.of(JwtClaimNames.SUB, source -> {
        return "Custom conversion result";
    })));
```
Then make a request with JWT without "sub" claim.

**Expected behavior**

Conversion is not called for absent claims.

[lines]: https://github.com/spring-projects/spring-security/blob/04161b9288a6ff667a68aa4c1e341faf7b535814/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java#L156-L158

[convertInstant]: https://github.com/spring-projects/spring-security/blob/04161b9288a6ff667a68aa4c1e341faf7b535814/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java#L120-L127

eleftherias:
Thanks for reaching out @mitasov-ra!
I believe this issue describes 2 problems.

The first problem is described in the "To Reproduce" section, which is that a `NullPointerException` is thrown when accepting JWT without particular claims. 
I believe this could be a duplicate of gh-10142.
Could you confirm which version of Spring Security you are using?

The second problem is described in the issue title, which is that `MappedJwtClaimSetConverter` calls `Converter#convert` with `null`.
[...]

PR #10549, opened 2021-11-23
Fixed NPE in MappedJwtClaimSetConverter

mitasov-ra:
Issue #10548
Conversion is no longer called for missing claims.
Conversion is no longer called for claims with null value.

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverter.java
@@ -118,18 +118,12 @@ public final class MappedJwtClaimSetConverter implements Converter<Map<String, O
 	}
 
 	private static Instant convertInstant(Object source) {
-		if (source == null) {
-			return null;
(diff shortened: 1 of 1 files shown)


eleftherias (changes requested):
@mitasov-ra, let's continue this discussion on the related issue gh-10549.
I will note that we would not be able to accept this PR as-is, because the suggested changes break some of the existing functionality.
We want to allow users to add new claims by providing a custom converter.
This is illustrated in the tests, specifically https://github.com/spring-projects/spring-security/blob/74e3abc992e5372a42787dfb757ca732b5459f75/oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/MappedJwtClaimSetConverterTests.java#L144-L153

The behavior is also described the Javadoc for the `MappedJwtClaimSetConverter` constructor
>  A converter that returns a non-null value **adds** or replaces that claim in the claim set.

eleftherias:
@mitasov-ra, I'm going to close this PR for this reasons mentioned in my previous comment.
I will keep the associated issue, gh-10549, open.

mitasov-ra:
Got it, didn't spot that "adds" part in docs.

PR #10553, opened 2021-11-26
Allow Multipart form POSTs to include access tokens

jamietanna:
As part of changes in 6db58cbf8a26296a15eb9f9c187176f099acaa77, we
modified the logic for whether an access token could be sent as a
parameter in a request's body.

However, this doesn't account for Multipart POSTs, for instance where an
image and an access token may be provided together.

This isn't called out as an option for RFC 6750, but is being used in
practice, so we should aim to support it.

In use i.e. by the [Micropub standard](https://micropub.spec.indieweb.org/#reserved-properties)


sjohnr:
Hi @jamietanna, thanks for your interest in the project!

Unfortunately, I don't know much about the Micropub standard or the community behind that document, but my understanding is that it is not an internet standards track document, correct?

Unfortunately, this PR seems to regress issue gh-10326, which specifically aligned us with RFC 6750 to avoid issues with multipart form uploads. If support for multipart form uploads is required in your case, I think it would be fairly reasonable to provide your own `BearerTokenResolver`. In case I'm missing something obvious, can you explain more about your use case and whether it can (or can't) be solved with a custom implementation of `BearerTokenResolver`?

sjohnr:
Thanks @jamietanna for the follow up and understanding!

Issue #10573, opened 2021-12-01
Prevent using both `authorizeRequests` and `authorizeHttpRequests`


marcusdacoregio:
When users configure the `SecurityFilterChain` they can opt to use `authorizeHttpRequests` or `authorizeRequests`. Using both configurations does not make sense since they do the same thing.

Using `authorizeHttpRequests` is preferable now, since `authorizeRequests` is gonna be deprecated.

We should ensure that only one of them is configured inside the filter chain.

rwinch:
@marcusdacoregio Can you please ensure this version is updated to 5.7.0-M1 ?

PR #10574, opened 2021-12-01
Prevent using both `authorizeRequests` and `authorizeHttpRequests`

marcusdacoregio:
Closes gh-10573

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
@@ -2889,8 +2889,15 @@ public final class HttpSecurity extends AbstractConfiguredSecurityBuilder<Defaul
 		}
 	}
 
+	@SuppressWarnings("unchecked")
 	@Override
 	protected DefaultSecurityFilterChain performBuild() {
+		ExpressionUrlAuthorizationConfigurer<?> expressionConfigurer = getConfigurer(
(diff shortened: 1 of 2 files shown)


marcusdacoregio:
Thanks for the feedback @rwinch.

Totally agreed. Users who add filters manually know exactly what they are doing. I'm going to apply the changes.

marcusdacoregio:
I've updated the PR to avoid having `authorizeHttpRequests` and `authorizeRequests` in the same configuration.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/AuthorizeHttpRequestsConfigurer.java
```diff
@@ -70,8 +70,13 @@ public AuthorizationManagerRequestMatcherRegistry getRegistry() {
 		return this.registry;
 	}
 
+	@SuppressWarnings("unchecked")
 	@Override
 	public void configure(H http) {
+		ExpressionUrlAuthorizationConfigurer<?> configurer = http
```
rwinch, line 76: Can we move this to where we are building http? I'd prefer the configurers not need to know about each other.

rwinch (changes requested):

marcusdacoregio:
Thanks @rwinch. I've updated the PR.

rwinch (approved):
The code looks great!

The only other change is the PR should be pointed to 5.7 branch rather than main since the ticket is (rightfully) assigned to 5.7

Issue #10554, opened 2021-11-26
`WebInvocationPrivilegeEvaluator` Bean should support multiple `SecurityFilterChain`s


fast-reflexes:
## Background
Having multiple `WebSecurityConfigurerAdapter`'s which are ordered and each processes a limited set of paths is a very handy tool for creating a structured security configuration. The advantages are:

* May add a catch all that denies all access for any request not explicitly whitelisted by previous configs.
* With both a regular API with basic auth, internal API for UI which uses login via SAML2 and some parts of an app open, configuring it all in one adapter would be messy (if even possible) and therefore also error-prone.

We have had a setup with multiple adapters for a long time and it has worked very well. Spring does not disallow it and there are parts in the code which even suggests that this is an intended use (for example the property `securityFilterChainBuilders` in singleton `WebSecurity` which is a `List` of builders and not a single object). The role of method `HttpSecurity.requestMatchers` also encourages such use.

## Bug
Nonetheless, parts of Spring Security does not take this into account. More specifically, the `WebSecurity` singleton has a property named `filterSecurityInterceptor` which is populated with the `FilterSecurityInterceptor` from the LAST `WebSecurityConfigurerAdapter` processed. This `FilterSecurityInterceptor` is then added to the bean `DefaultWebInvocationPrivilegeEvaluator` created via bean method `privilegeEvaluator` in `WebSecurityConfiguration`. This means that any use of property `filterSecurityInterceptor` in singleton `WebSecurity`, or any use of bean `DefaultWebInvocationPrivilegeEvaluator` will only take the last processed `WebSecurityConfigurerAdapter` into account, which does not seem correct. Luckily, it seems to me that the only use of property `filterSecurityInterceptor` is to create bean `DefaultWebInvocationPrivilegeAdapter` and the only use of this bean is in newly added filter `ErrorPageSecurityFilter`, which is also where we can see this bug at play. Nonetheless, the design seems flawed.

Note that this bug is NOT any of the following:

* `ErrorPageSecurityFilter` inherits from `HttpFilter` and not `Filter` causing `ClassDefNotFound` exception in some environments:
  * https://github.com/spring-projects/spring-boot/issues/28790 
  * https://stackoverflow.com/questions/70087604/why-do-i-have-classnotfoundexception-javax-servlet-http-httpfilter-for-a-war-af/70087906
[...]

PR #10575, opened 2021-12-01
Configure WebInvocationPrivilegeEvaluator for multiple `SecurityFilterChain`s

marcusdacoregio:
Closes gh-10554, gh-10590

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/builders/WebSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/builders/WebSecurity.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.List;
 
 import javax.servlet.Filter;
+import javax.servlet.ServletContext;
 import javax.servlet.http.HttpServletRequest;
 
 import org.apache.commons.logging.Log;
@@ -33,6 +34,7 @@ import org.springframework.http.HttpMethod;
 import org.springframework.security.access.PermissionEvaluator;
 import org.springframework.security.access.expression.SecurityExpressionHandler;
 import org.springframework.security.access.hierarchicalroles.RoleHierarchy;
+import org.springframework.security.authorization.AuthorizationManager;
 import org.springframework.security.config.annotation.AbstractConfiguredSecurityBuilder;
 import org.springframework.security.config.annotation.ObjectPostProcessor;
 import org.springframework.security.config.annotation.SecurityBuilder;
@@ -47,9 +49,12 @@ import org.springframework.security.web.DefaultSecurityFilterChain;
 import org.springframework.security.web.FilterChainProxy;
 import org.springframework.security.web.FilterInvocation;
 import org.springframework.security.web.SecurityFilterChain;
+import org.springframework.security.web.access.AuthorizationManagerWebInvocationPrivilegeEvaluator;
 import org.springframework.security.web.access.DefaultWebInvocationPrivilegeEvaluator;
+import org.springframework.security.web.access.RequestMatcherDelegatingWebInvocationPrivilegeEvaluator;
 import org.springframework.security.web.access.WebInvocationPrivilegeEvaluator;
 import org.springframework.security.web.access.expression.DefaultWebSecurityExpressionHandler;
+import org.springframework.security.web.access.intercept.AuthorizationFilter;
 import org.springframework.security.web.access.intercept.FilterSecurityInterceptor;
 import org.springframework.security.web.debug.DebugFilter;
 import org.springframework.security.web.firewall.HttpFirewall;
@@ -57,7 +62,9 @@ import org.springframework.security.web.firewall.RequestRejectedHandler;
 import org.springframework.security.web.firewall.StrictHttpFirewall;
 import org.springframework.security.web.servlet.util.matcher.MvcRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher;
+import org.springframework.security.web.util.matcher.RequestMatcherEntry;
 import org.springframework.util.Assert;
+import org.springframework.web.context.ServletContextAware;
 import org.springframework.web.filter.DelegatingFilterProxy;
 
 /**
@@ -81,7 +88,7 @@ import org.springframework.web.filter.DelegatingFilterProxy;
  * @see WebSecurityConfiguration
  */
 public final class WebSecurity extends AbstractConfiguredSecurityBuilder<Filter, WebSecurity>
-		implements SecurityBuilder<Filter>, ApplicationContextAware {
+		implements SecurityBuilder<Filter>, ApplicationContextAware, ServletContextAware {
 
 	private final Log logger = LogFactory.getLog(getClass());
 
@@ -107,6 +114,8 @@ public final class WebSecurity extends AbstractConfiguredSecurityBuilder<Filter,
 
 	private Runnable postBuildAction = () -> {
 	};
+
+	private ServletContext servletContext;
 
 	/**
 	 * Creates a new instance
@@ -252,6 +261,8 @@ public final class WebSecurity extends AbstractConfiguredSecurityBuilder<Filter,
 	 * {@link WebSecurityConfigurerAdapter}.
 	 * @param securityInterceptor the {@link FilterSecurityInterceptor} to use
 	 * @return the {@link WebSecurity} for further customizations
+	 * @deprecated Use {@link #privilegeEvaluator(WebInvocationPrivilegeEvaluator)}
+	 * instead
 	 */
 	public WebSecurity securityInterceptor(FilterSecurityInterceptor securityInterceptor) {
 		this.filterSecurityInterceptor = securityInterceptor;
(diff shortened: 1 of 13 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/builders/WebSecurity.java
```diff
@@ -306,6 +315,28 @@ protected Filter performBuild() throws Exception {
...
+			for (Filter filter : securityFilterChain.getFilters()) {
+				if (filter instanceof FilterSecurityInterceptor fsi) {
+					builder.withFilterSecurityInterceptor(fsi, this.servletContext);
+					continue;
+				}
+				if (filter instanceof AuthorizationFilter af) {
+					builder.withAuthorizationManager(af.getAuthorizationManager());
+				}
+			}
+			delegates.add(builder.build());
```
rwinch, line 335: Do not add if there is no delegates because we don't want to have to invoke match unnecessarily.

config/src/main/java/org/springframework/security/config/annotation/web/builders/WebSecurity.java
```diff
@@ -284,6 +292,7 @@ protected Filter performBuild() throws Exception {
 		for (SecurityBuilder<? extends SecurityFilterChain> securityFilterChainBuilder : this.securityFilterChainBuilders) {
 			securityFilterChains.add(securityFilterChainBuilder.build());
 		}
+		buildPrivilegeEvaluator(securityFilterChains);
```
rwinch, line 295: Why is this here the default instead of replacing the default for [`getPrivilegeEvaluator()`](https://github.com/spring-projects/spring-security/blob/a9d98e05f0471dbb888a48a60729279e053649d8/config/src/main/java/org/springframework/security/config/annotation/web/builders/WebSecurity.java#L252). Putting it in the getter ensures that if anyone tries to obtain it the defaulting is consistent.

web/src/main/java/org/springframework/security/web/access/AuthorizationManagerWebInvocationPrivilegeEvaluator.java
```diff
@@ -0,0 +1,61 @@
...
+	@Override
+	public boolean isAllowed(String uri, Authentication authentication) {
+		return isAllowed(null, uri, null, authentication);
+	}
+
+	@Override
+	public boolean isAllowed(String contextPath, String uri, String method, Authentication authentication) {
+		FilterInvocation filterInvocation = new FilterInvocation(contextPath, uri, method);
+		try {
+			this.authorizationManager.verify(() -> authentication, filterInvocation.getHttpRequest());
```
rwinch, line 53: Invoking verify is somewhat costly because it throws an Exception. Instead, the code should invoke the check method.

web/src/main/java/org/springframework/security/web/access/RequestMatcherDelegatingWebInvocationPrivilegeEvaluator.java
```diff
@@ -0,0 +1,179 @@
...
+		return null;
+	}
+
+	public static class DelegatePrivilegeEvaluator {
+
+		private final Function<HttpServletRequest, Boolean> requestMatcher;
+
+		private final List<WebInvocationPrivilegeEvaluator> privilegeEvaluators;
+
+		public DelegatePrivilegeEvaluator(Function<HttpServletRequest, Boolean> requestMatcher) {
```
rwinch, line 127: The code should not allow one to be created without any WebInvocationProivilegeEvaluator instances because we don't want the `RequestMatcher` to be invoked unnecessarily (it decreases the performance every time a check is performed.

web/src/main/java/org/springframework/security/web/access/RequestMatcherDelegatingWebInvocationPrivilegeEvaluator.java
```diff
@@ -0,0 +1,179 @@
...
+		}
+
+		public DelegatePrivilegeEvaluator(Function<HttpServletRequest, Boolean> requestMatcher,
+				List<WebInvocationPrivilegeEvaluator> privilegeEvaluators) {
+			Assert.notNull(requestMatcher, "requestMatcher cannot be null");
+			this.requestMatcher = requestMatcher;
+			this.privilegeEvaluators = privilegeEvaluators;
+		}
+
+		public boolean matches(HttpServletRequest request) {
```
rwinch, line 138: Perhaps change this to return the `RequestMatcher`.

web/src/main/java/org/springframework/security/web/access/RequestMatcherDelegatingWebInvocationPrivilegeEvaluator.java
```diff
@@ -0,0 +1,179 @@
...
+			return this.requestMatcher.apply(request);
+		}
+
+		public List<WebInvocationPrivilegeEvaluator> getPrivilegeEvaluators() {
+			return this.privilegeEvaluators;
+		}
+
+	}
+
+	public static final class Builder {
```
rwinch, line 148: Perhaps remove this Builder for now since all of the operations are happening on the DelegatePrivilegeEvaluator anyway (i.e. an invalid DelegatePrivilegeEvaluator can be created).

web/src/main/java/org/springframework/security/web/access/RequestMatcherDelegatingWebInvocationPrivilegeEvaluator.java
```diff
@@ -0,0 +1,179 @@
...
+			}
+		}
+		return null;
+	}
+
+	public static class DelegatePrivilegeEvaluator {
+
+		private final Function<HttpServletRequest, Boolean> requestMatcher;
+
+		private final List<WebInvocationPrivilegeEvaluator> privilegeEvaluators;
```
rwinch, line 125: This should be immutable list

config/src/main/java/org/springframework/security/config/annotation/web/builders/WebSecurity.java
```diff
@@ -306,6 +315,28 @@ protected Filter performBuild() throws Exception {
...
+					builder.withFilterSecurityInterceptor(fsi, this.servletContext);
+					continue;
+				}
+				if (filter instanceof AuthorizationFilter af) {
+					builder.withAuthorizationManager(af.getAuthorizationManager());
+				}
+			}
+			delegates.add(builder.build());
+		}
+		this.privilegeEvaluator = new RequestMatcherDelegatingWebInvocationPrivilegeEvaluator(delegates);
```
rwinch, line 337: instead of initializing privilegeEvaluator it should create an instance and be assigned in getPrivilegeEvaluator. See https://github.com/spring-projects/spring-security/pull/10575/files#diff-2a3768a3ea186b6702859073cc35bc02aa6c8ebf6b20f19a67323f7c20d3bd38R295

web/src/main/java/org/springframework/security/web/access/RequestMatcherDelegatingWebInvocationPrivilegeEvaluator.java
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	private DelegatePrivilegeEvaluator getDelegate(String contextPath, String uri, String method) {
+		FilterInvocation filterInvocation = new FilterInvocation(contextPath, uri, method);
+		for (DelegatePrivilegeEvaluator delegate : this.delegates) {
+			if (delegate.matches(filterInvocation.getHttpRequest())) {
+				return delegate;
+			}
+		}
+		return null;
```
rwinch, line 118: Instead of null, change the return type to `List<WebInvocationPrivilegeEvaluator>`. If nothing is found return an empty `List`.

rwinch (changes requested):
Thank you for the PR. I've provided feedback.

web/src/main/java/org/springframework/security/web/access/RequestMatcherDelegatingWebInvocationPrivilegeEvaluator.java
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	public static class DelegatePrivilegeEvaluator {
+
+		private final Function<HttpServletRequest, Boolean> requestMatcher;
```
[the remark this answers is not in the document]
marcusdacoregio: Since we don't have the `RequestMatcher` available when constructing it, we will stick to the `Function` for now

marcusdacoregio:
Thanks for the review @rwinch. I've updated the PR based on your suggestions.

web/src/main/java/org/springframework/security/web/access/RequestMatcherDelegatingWebInvocationPrivilegeEvaluator.java
```diff
@@ -0,0 +1,128 @@
...
+
+	public static class RequestMatcherPrivilegeEvaluator {
+
+		private final Function<HttpServletRequest, Boolean> requestMatcher;
+
+		private final WebInvocationPrivilegeEvaluator privilegeEvaluator;
+
+		public RequestMatcherPrivilegeEvaluator(Function<HttpServletRequest, Boolean> requestMatcher,
+				WebInvocationPrivilegeEvaluator privilegeEvaluator) {
+			Assert.notNull(requestMatcher, "requestMatcher cannot be null");
```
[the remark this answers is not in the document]
rwinch: It looks like there is still the need for an empty check on privilegeEvalutators

rwinch (changes requested):

web/src/main/java/org/springframework/security/web/access/AuthorizationManagerWebInvocationPrivilegeEvaluator.java
```diff
@@ -0,0 +1,61 @@
...
+import org.springframework.util.Assert;
+
+/**
+ * An implementation of {@link WebInvocationPrivilegeEvaluator} which delegates the checks
+ * to an instance of {@link AuthorizationManager}
+ *
+ * @author Marcus Da Coregio
+ * @since 5.7
+ */
+public final class AuthorizationManagerWebInvocationPrivilegeEvaluator implements WebInvocationPrivilegeEvaluator {
```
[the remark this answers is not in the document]
marcusdacoregio: I've splitted it in another commit that closes gh-10590

web/src/main/java/org/springframework/security/web/access/RequestMatcherDelegatingWebInvocationPrivilegeEvaluator.java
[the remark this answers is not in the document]
marcusdacoregio: There will be some scenarios where it makes sense to have an empty list. When using `web.ignoring(...)` we have a `SecurityFilterChain` with no filters, therefore no `WebInvocationPrivilegeEvaluator`s. So, in this scenario we can accept an empty list

rwinch (approved):
Thanks for the PR! LGTM

Issue #10571, opened 2021-11-30
Add hasIpAddress to Reactive Kotlin DSL


sjohnr:
This issue adds support for `hasIpAddress` in the Reactive Kotlin DSL.

Related #10570

PR #10579, opened 2021-12-01
Add hasIpAddress to Reactive Kotlin DSL

sjohnr:
Closes gh-10571

### DIFF
--- a/config/src/main/kotlin/org/springframework/security/config/web/server/AuthorizeExchangeDsl.kt
+++ b/config/src/main/kotlin/org/springframework/security/config/web/server/AuthorizeExchangeDsl.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


sjohnr:
(I snuck the missing since commit in here, but this is mainly about hasIpAddress support in the DSL)

config/src/main/kotlin/org/springframework/security/config/web/server/AuthorizeExchangeDsl.kt
```diff
@@ -108,6 +109,12 @@ class AuthorizeExchangeDsl {
     fun hasAnyAuthority(vararg authorities: String): ReactiveAuthorizationManager<AuthorizationContext> =
             AuthorityReactiveAuthorizationManager.hasAnyAuthority<AuthorizationContext>(*authorities)
 
+    /**
+     * Require a specific IP or range of IP addresses.
+     */
```
eleftherias, line 114: We also need a `@since 5.7` tag here.

eleftherias (changes requested):
@sjohnr I left one comment inline and the copyright headers need to be updated to 2021 in the Kotlin files. Everything else looks good.

Issue #10566, opened 2021-11-30
`AuthorityAuthorizationManager` incorrectly compares `GrantedAuthority`


legas1:
**Describe the bug**
Hi, in commit https://github.com/spring-projects/spring-security/commit/86c24da38b9b4452d2208fe4ea46e9c35e0d6f93 there was a slight change in comparison of allowed authorities to endpoints. However in our use case this was huge breaking change.

Lines:
https://github.com/spring-projects/spring-security/commit/86c24da38b9b4452d2208fe4ea46e9c35e0d6f93#diff-8c62a1a24d0860e1da929cdb7cdbf50a8e7daa565fd03f05673299917891f33bR40
https://github.com/spring-projects/spring-security/commit/86c24da38b9b4452d2208fe4ea46e9c35e0d6f93#diff-eb4576063aa24fd635f700152f04a2590973ce56a20899414957e53a2584da74R38

In our Kotlin environment, we implement enum which implements GrantedAuthority and custom Authentication class, when this enum reaches ReactiveAuthorizationManager check method, its compared to SimpleGrantedAuthority (class implementing GrantedAuthority, breaking change from changes above) instead of underlying string authority, which results in non equality and thus denying access to endpoint. 

**To Reproduce**
1. Have custom implementation of AuthenticationManager which have custom Authentication implementation and custom GrantedAuthority implementation.
2. Attempt to use protected endpoint with some authority.
3. Access denied.

**Expected behavior**
Allow access to given endpoint.

**Sample**
https://github.com/legas1/ss.grantedauthority.demo

Does it make sense or do you need more clarification? Or I am completely wrong about my understanding of upper changes? The solution in my opinion would be to revert those changes or maybe adjust equal method of SimpleGrantedAuthority to count with GrantedAuthority interface.

Cheers, Daniel

legas1:
Added example.

terminux:
When using a custom `GrantedAuthority`, there will indeed be bugs that fail to match, i submitted a PR to fix it. #10588

marcusdacoregio:
Fixed via https://github.com/spring-projects/spring-security/pull/10588

marcusdacoregio:
Thanks folks. This is now merged into `5.7.x` and backported to `5.6.x`. The next version `5.6.1` is gonna have the fix.

PR #10588, opened 2021-12-06
Fix the bug that the custom GrantedAuthority comparison fails

terminux:
Closes gh-10566

### DIFF
--- a/core/src/test/java/org/springframework/security/authorization/AuthorityAuthorizationManagerTests.java
+++ b/core/src/test/java/org/springframework/security/authorization/AuthorityAuthorizationManagerTests.java
@@ -16,12 +16,14 @@
 
 package org.springframework.security.authorization;
 
+import java.util.Collections;
 import java.util.function.Supplier;
 
 import org.junit.jupiter.api.Test;
 
 import org.springframework.security.authentication.TestingAuthenticationToken;
 import org.springframework.security.core.Authentication;
+import org.springframework.security.core.GrantedAuthority;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatIllegalArgumentException;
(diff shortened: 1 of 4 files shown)


core/src/test/java/org/springframework/security/authorization/AuthorityAuthorizationManagerTests.java
```diff
@@ -91,6 +93,18 @@ public void hasAnyAuthorityWhenContainNullThenException() {
 				.withMessage("authorities cannot contain null values");
 	}
 
+	@Test
+	public void hasAnyAuthorityWhenUserHasCustomAuthorityThenGrantedDecision() {
```
marcusdacoregio, line 97: Can you also provide a test for when the `Authentication` has the wrong authorities but uses a custom implementation?

core/src/test/java/org/springframework/security/authorization/AuthorityReactiveAuthorizationManagerTests.java
```diff
@@ -88,6 +89,15 @@ public void checkWhenHasAuthorityAndAuthorizedThenReturnTrue() {
 		assertThat(granted).isTrue();
 	}
 
+	@Test
+	public void checkWhenHasCustomAuthorityAndAuthorizedThenReturnTrue() {
```
marcusdacoregio, line 93: Can you also provide a test for when the `Authentication` has the wrong authorities but uses a custom implementation?

marcusdacoregio (changes requested):
Hi @terminux. Thanks for the PR.

I've left some minimum feedback inline.

core/src/test/java/org/springframework/security/authorization/AuthorityAuthorizationManagerTests.java
terminux: Done.

core/src/test/java/org/springframework/security/authorization/AuthorityReactiveAuthorizationManagerTests.java
terminux: Done.

marcusdacoregio (approved):
Awesome!

PR #10614, opened 2021-12-18
Set the AuthnRequestsSigned attribute in the Saml2 metadata filter

xpmatteo:
Previously, the AuthnRequestsSigned attribute was not set in the
metadata xml that is produced by the Saml2 metadata filter.  As a
result, the maintainers of the IDP I had to register with were rejecting
our application.

Now, the attribute is always set, namely to "true" if the asserting
party wants signed authentication requests, and false otherwise. This
declares to the asserting party whether authentication requests will be
signed or not.


jzheaux:
Hi, @xpmatteo. Thanks for the contribution.

While I see your intent, I don't think we want to assume RP configurations based on AP configurations. 

Instead, what about adding a `Converter` method that allows the `EntityDescriptor` to be edited before getting serialized, e.g. `setParametersConsumer(Consumer<EntityDescriptorParameters> parametersConsumer)`? This would be similar to `OpenSaml4LogoutRequestResolver#setParametersConsumer`.

jzheaux:
Closed in favor of https://github.com/spring-projects/spring-security/issues/10839

PR #10675, opened 2022-01-04
Implement SHA-256 in token based Remember-Me services

marcusdacoregio:
Closes gh-8549

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/RememberMeConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/RememberMeConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2015 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -33,6 +33,7 @@ import org.springframework.security.web.authentication.rememberme.AbstractRememb
 import org.springframework.security.web.authentication.rememberme.PersistentTokenBasedRememberMeServices;
 import org.springframework.security.web.authentication.rememberme.PersistentTokenRepository;
 import org.springframework.security.web.authentication.rememberme.RememberMeAuthenticationFilter;
+import org.springframework.security.web.authentication.rememberme.RememberMeHashingAlgorithm;
 import org.springframework.security.web.authentication.rememberme.TokenBasedRememberMeServices;
 import org.springframework.security.web.authentication.ui.DefaultLoginPageGeneratingFilter;
 import org.springframework.util.Assert;
@@ -90,6 +91,8 @@ public final class RememberMeConfigurer<H extends HttpSecurityBuilder<H>>
 	private AuthenticationSuccessHandler authenticationSuccessHandler;
 
 	private String key;
+
+	private RememberMeHashingAlgorithm hashingAlgorithm = RememberMeHashingAlgorithm.UNSET;
 
 	private RememberMeServices rememberMeServices;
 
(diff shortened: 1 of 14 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/configurers/RememberMeConfigurer.java
```diff
@@ -186,6 +189,24 @@ public RememberMeConfigurer<H> key(String key) {
...
+	 * password. If unset, the MD5 message digest algorithm will be used.
+	 * <p>
+	 * This configuration is ignored if
+	 * {@link #tokenRepository(PersistentTokenRepository)} or
+	 * {@link #rememberMeServices(RememberMeServices)} are used.
+	 * @param hashingAlgorithm the algorithm used when creating new cookies
+	 * @return the {@link RememberMeConfigurer} for further customization
+	 * @since 5.7
+	 */
+	public RememberMeConfigurer<H> hashingAlgorithm(RememberMeHashingAlgorithm hashingAlgorithm) {
```
jzheaux, line 204: Since `RememberMeServices` can be constructed with relative ease, I'd recommend leaving this out of the DSL for the time being.

web/src/main/java/org/springframework/security/web/authentication/rememberme/TokenBasedRememberMeServices.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2004, 2005, 2006 Acegi Technology Pty Limited
```
jzheaux, line 2: The ones that say "Acegi Technology Pty Limited", we leave as-is

web/src/main/java/org/springframework/security/web/authentication/rememberme/TokenBasedRememberMeServices.java
```diff
@@ -51,13 +51,20 @@
...
  *
  * <pre>
  * username + &quot;:&quot; + expiryTime + &quot;:&quot;
  * 		+ Md5Hex(username + &quot;:&quot; + expiryTime + &quot;:&quot; + password + &quot;:&quot; + key)
  * </pre>
  *
+ * or, if a hashing algorithm is configured, the form:
+ *
+ * <pre>
+ * username + &quot;:&quot; + expiryTime + &quot;:&quot; + algorithmName + &quot;:&quot;
```
jzheaux, line 64: I wonder if this should be thought of in the same way as how Spring Security deals with passwords.

For example, you might consider having an algorithm-for-matches and an algorithm-for-encoding, both defaulting to MD5.

* When encoding, encode with the algorithm-for-encoding, adding the prefix either way (even if the app hasn't configured it directly)
* When decoding
** if there is no algorithm in the cookie, use the algorithm-for-matches
** if there is an algorithm in the cookie, use that algorithm

This will allow folks to safely upgrade to SHA-256 without losing their old MD5 hash cookies.

web/src/main/java/org/springframework/security/web/authentication/rememberme/TokenBasedRememberMeServices.java
```diff
@@ -83,64 +95,86 @@
...
 	}
 
+	/**
+	 * Constructs a {@link TokenBasedRememberMeServices} using the provided parameters
+	 * @param key the key to use in the hashing algorithm
+	 * @param userDetailsService the {@link UserDetailsService} to retrieve users from
+	 * @param hashingAlgorithm the hashing algorithm to use in the Cookie value
+	 */
+	public TokenBasedRememberMeServices(String key, UserDetailsService userDetailsService,
+			RememberMeHashingAlgorithm hashingAlgorithm) {
```
jzheaux, line 113: I think that since this is an optional parameter, it would be ideal as a setter.

docs/modules/ROOT/pages/servlet/appendix/namespace/http.adoc
```diff
@@ -1798,6 +1798,11 @@ Specifies the period in seconds for which the remember-me cookie should be valid
 By default it will be valid for 14 days.
 
 
+[[nsa-remember-me-hashing-algorithm]]
+* **hashing-algorithm**
```
jzheaux, line 1802: Likewise, would recommend that this be left out for now in favor of applications using `services-ref`.

jzheaux (changes requested):
Thanks, @marcusdacoregio. I've left my comments inline.

marcusdacoregio:
Hi @jzheaux, I'm closing this in favor of https://github.com/spring-projects/spring-security/pull/11464. Can you please review that one? I've addressed your feedback there.

PR #10679, opened 2022-01-05
Improve GRADLE build Performance

xiayingfeng:
[Compiler daemon](https://docs.gradle.org/current/userguide/performance.html#compiler_daemon). The Gradle Java plugin allows you to run the compiler as a separate process by setting `options.fork = true`. This feature can lead to much less garbage collection and make Gradle’s infrastructure faster. This project has more than 1000 source files. We can consider enabling this feature.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -118,6 +118,10 @@ subprojects {
 
 
 allprojects {
+    tasks.withType(JavaCompile).configureEach {
+        options.fork = true
+    }
+
 	if (!['spring-security-bom', 'spring-security-docs'].contains(project.name)) {
 		apply plugin: 'io.spring.javaformat'
 		apply plugin: 'checkstyle'


rwinch:
Thank you for the PR!

I did a quick comparison on the build performance and it appears that it is slower when running with the PR's changes of using a forked compiler (56 seconds) than the original code of not using a forked compiler (34 seconds). My test was rather simple and naive, so please correct me if I am missing something. I first ran `./gradlew clean` to ensure nothing was built. Then I ran `./gradlew classes testClasses integrationTestClasses --no-build-cache` to compile everything and looked at the build time. Note that `--no-build-cache` is important as we make use of Gradle Enterprise Cache which will impact the build if it has already been built.

The [documentation](https://docs.gradle.org/current/userguide/performance.html#compiler_daemon) states:

> It’s unlikely to be useful for small projects, but you should definitely consider it if a single task is compiling close to a thousand or more source files together.

While the entire project has more than a thousand source files, there is a separate compile task per source set per project, so the count is much smaller per "single task".

In short, I think this option actually slows the Spring Security build down. Please let me know if you see any problems with my logic here.

rwinch:
I'm going to go ahead and close this. Please feel free to comment if I was missing something.

PR #10689, opened 2022-01-08
Fix for gh10663 encryptedID

bitrecycling:


bitrecycling:
Build and tests don't fail on my machine with ./gradlew clean build integrationTest. The log mentions a html page (which I cannot access) with test results and the test that breaks, but that is not enough information to figure out what's wrong. So I can hardly fix it. Can you give some more insight or fix the test?

bitrecycling:
@jzheaux I tried to reproduce and fix

jzheaux:
Thanks, @bitrecycling! I've merged this into `5.7.x` via 700cae8d3b2590397dc38ac1eb67d1b19203ec0d and a polish of 3c45d46bd74db9975568dddc3358a739131d062f. It's also been merged into `main` as part of the 6.0 release.

bitrecycling:
> Thanks, @bitrecycling!
@jzheaux glad I could help, also I might be using it soon :)

Issue #10567, opened 2021-11-30
Check for `opensaml` version on OpenSaml support classes


marcusdacoregio:
It would be nice for users to know exactly which version of `opensaml` classes to use. A check could be performed in order to ensure that the correct version of `opensaml` is being used in the current `OpenSamlX` implementation.

One option would be to add the check inside the `OpenSamlX` constructors.

See:
 - https://github.com/spring-projects/spring-security/issues/10547

igorpele:
Hi, I would like to try this.
As far as I could see there is already a version check in the `Saml2LoginConfigurer` and `Saml2LogoutConfigurer`.
Additionally it should be checked in the according `OpenSaml4LoginRequestResolver` or `OpenSaml4LoginRequestResolver` (and all the other OpenSaml3... and OpenSaml4... classes) whether the correct version of opensaml is in the classpath?

marcusdacoregio:
Hi @igorpele, thanks for the interest. The ticket is yours.

Yes, I can think of creating a util class to retrieve the version, like it's being done in `Saml2LoginConfigurer` and `Saml2LogoutConfigurer`.

And inside of each of those constructors, you should verify if the version starts with `3` or `4`. And for the SAML 2.0 Logout classes, the minimum version should be `4.1` as per https://github.com/spring-projects/spring-security/issues/10547#issuecomment-982394169.  

Please, feel free to keep discussing if it's not crystal clear yet.

igorpele:
Hi @marcusdacoregio, thanks for the hints.
Is SAML2 SSO supported with OpenSAML version 4.0.1 as it throws an Exception in the `OpenSaml4AuthenticationProvider` as already mentioned in #10547?
The `OpenSaml4LogoutRequestResolver` and `OpenSaml4LogoutResponseResolver` will after this change only work with versions >= 4.1.0, but should this also be true for the `OpenSaml4AuthenticationProvider`?

marcusdacoregio:
> Is SAML2 SSO supported with OpenSAML version 4.0.1 as it throws an Exception in the OpenSaml4AuthenticationProvider as already mentioned in #10547?

I think we have to check if SSO is working with version `4.0.1`, if not the minimum version should be `4.1.x` then.

> The OpenSaml4LogoutRequestResolver and OpenSaml4LogoutResponseResolver will after this change only work with versions >= 4.1.0, but should this also be true for the OpenSaml4AuthenticationProvider?

[...]

PR #10693, opened 2022-01-10
Check for opensaml version on OpenSaml support classes

igorpele:
In OpenSaml support classes the Saml2VersionUtils class is used to check whether a supported version of OpenSaml is found on the classpath.

Closes gh-10567


marcusdacoregio:
Hi @igorpele, just a heads up that we are discussing this solution and I should get back to you soon.

marcusdacoregio:
Waiting for https://github.com/spring-projects/spring-security/pull/10817 to get merged, since the way to verify the OpenSAML version may change

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/Saml2VersionUtils.java
```diff
@@ -0,0 +1,99 @@
...
+import org.opensaml.core.Version;
+
+import org.springframework.util.Assert;
+
+/**
+ * A Utils class that allows checking the version of the OpenSAML library in the
+ * classpath.
+ *
+ * @author Igor Pelesić
+ * @since 5.7
```
marcusdacoregio, line 28: Could you change it to `@since 5.8`?

marcusdacoregio (changes requested):

marcusdacoregio:
Hey @igorpele, thanks for your contribution.

Unfortunately, this has not made it into our 5.8 release and we do not have another release planned for the 5.x line.
In Spring Security 6, we are only using OpenSAML4, therefore there is no need for a version check.

We might still do something like this if we add OpenSAML5 #11658.

Issue #10451, opened 2021-10-28
Automate release process


sjohnr:
This issue lists the required tasks for automating the release. I've organized the release process (see [RELEASE.adoc](https://github.com/spring-projects/spring-security/blob/main/RELEASE.adoc)) into four stages:

1. **Prepare release** - This stage updates dependencies. It could be performed up to a week prior to the release.
2. **Perform release** - This stage pushes artifacts to Maven Central and creates a github release.
3. **Prepare next iteration** - This stage performs milestone management.
4. **Announce release** - This stage let's the team and/or community know the release is complete. Some of the steps may not be automated.

### Prepare Release

inputs:
* `nextVersion` - Version being released

tasks:
* #11257
* #10453
* #10454 (`./gradlew updateDependencies -PupdateMode=GITHUB_ISSUE -PnextVersion=$nextVersion`)
* check: build success

### Perform Release

inputs:
* `previousVersion` - Version to delete on https://spring.io/projects/spring-security#learn
* `nextVersion` - Version being released

tasks:
* check: ensure all issues are closed in current milestone (`./gradlew gitHubCheckMilestoneHasNoOpenIssues -PnextVersion=$nextVersion`)
* #10455
* #11020
* #11023
* #11022 
* #10928 (`./wait-for-done.sh $nextVersion`)
* #10456
* generate changelog (`./gradlew generateChangelog -PnextVersion=$nextVersion`)
* #10457
* #11308

### Prepare Next Iteration

inputs:
* `previousVersion` - Version just released
* `nextVersion` - Next milestone version

tasks:
* #10458
* #10460
* #10461
* #11021

### Announce Release

inputs:
* `nextVersion` - Version being released

tasks:
* #11307

sjohnr:
Note: Need to handle milestones being released to `repo.spring.io/milestone` vs `repo1.maven.org/maven2` during the "wait for release to appear on Maven Central" step.

sjohnr:
@eleftherias I added gh-11020 and gh-11021 for updating `antora.yml`, and gh-11022 as well as gh-11023 to handle triggering CI updates. These are the issues I encountered with the automation during the `6.0.0-M2` release.

sjohnr:
Note: I've moved the open issues associated with release automation into the 6.0.x bucket to manage them all together. I will be creating a single issue for each maintenance branch that the automation is being backported to when adding all of this work to each respective branch.

sjohnr:
[...]

PR #10701, opened 2022-01-13
Add workflow that triggers release on due date

eleftherias:
Add Gradle task that calculates the next release milestone based on the current version and check if it is due today.

The second step in the workflow is only a placeholder.

Issue gh-10451
Issue gh-10455

### DIFF
--- a/.github/workflows/release-next-version.yml
+++ b/.github/workflows/release-next-version.yml
@@ -1,0 +1,91 @@
+name: Release Next Version
+
+on:
+  push:
+  schedule:
+    - cron: '0 0 * * MON' # Every Monday
+  workflow_dispatch: # Manual trigger
+
+env:
+  SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+  GRADLE_ENTERPRISE_CACHE_USER: ${{ secrets.GRADLE_ENTERPRISE_CACHE_USER }}
+  GRADLE_ENTERPRISE_CACHE_PASSWORD: ${{ secrets.GRADLE_ENTERPRISE_CACHE_PASSWORD }}
+  GRADLE_ENTERPRISE_SECRET_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_SECRET_ACCESS_KEY }}
+  RUN_JOBS: ${{ github.repository == 'spring-projects/spring-security' }}
+
+jobs:
+  prerequisites:
+    name: Pre-requisites for building
+    runs-on: ubuntu-latest
+    outputs:
+      runjobs: ${{ steps.continue.outputs.runjobs }}
+    steps:
+      - id: continue
+        name: Determine if should continue
+        if: env.RUN_JOBS == 'true'
+        run: echo "::set-output name=runjobs::true"
+  check_release_due:
+    name: Check if the release is due today
+    needs: [prerequisites]
+    runs-on: ubuntu-latest
+    if: needs.prerequisites.outputs.runjobs
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK 17
+        uses: actions/setup-java@v1
+        with:
+          java-version: '17'
+      - name: Setup gradle user name
+        run: |
+          mkdir -p ~/.gradle
+          echo 'systemProp.user.name=spring-builds+github' >> ~/.gradle/gradle.properties
+      - name: Cache Gradle packages
+        uses: actions/cache@v2
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+      - name: Check release
+        env:
+          GRADLE_ENTERPRISE_CACHE_USERNAME: ${{ secrets.GRADLE_ENTERPRISE_CACHE_USER }}
+          GRADLE_ENTERPRISE_CACHE_PASSWORD: ${{ secrets.GRADLE_ENTERPRISE_CACHE_PASSWORD }}
+          GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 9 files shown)


buildSrc/src/main/java/org/springframework/gradle/github/milestones/GitHubMilestoneNextVersionDueTodayTask.java
```diff
@@ -0,0 +1,87 @@
...
+			this.nextVersion = this.milestones.calculateNextReleaseMilestone(repository, currentVersion);
+			System.out.println("Next version was not provided. Calculated next version as " + this.nextVersion);
+		}
+		boolean milestoneDueToday = this.milestones.isMilestoneDueToday(this.repository, this.nextVersion);
+		if (!milestoneDueToday) {
+			throw new IllegalStateException("The milestone with the title " + this.nextVersion
+					+ " in the repository " + this.repository + " is not due yet");
+		}
+		System.out.println("The milestone with the title " + this.nextVersion + " in the repository "
+				+ this.repository + " is due today");
```
sjohnr, line 55: After discussion, I did a quick test, and it looks like if we only output the next version if `milestoneDueToday` is true, and (possibly) a blank line otherwise, we can capture that output with the `-q` flag.

For example:

```sh
nextVersion=$(./gradlew -q gitHubCheckMilestoneDueToday)
if [[ "$nextVersion" != "" ]]
then
    echo "Starting release for next version"
fi
```

**Note:** We could also set a task output and check that in a task further down, if we want to avoid the bash if-statement.

Perhaps we could split this into two tasks:

1. `gitHubCheckMilestoneDueToday` fails the build if the given/calculated version is not due today (can be used as a guard against accidentally releasing at the wrong time)
2. `gitHubCalculateNextReleaseMilestoneForToday` (or similar) which simply outputs the next version, or nothing/blank line

.github/workflows/release-next-version.yml
```diff
@@ -0,0 +1,90 @@
...
+      - name: Setup gradle user name
+        run: |
+          mkdir -p ~/.gradle
+          echo 'systemProp.user.name=spring-builds+github' >> ~/.gradle/gradle.properties
+      - name: Deploy artifacts
+        run: |
+          export GRADLE_ENTERPRISE_CACHE_USERNAME="$GRADLE_ENTERPRISE_CACHE_USER"
+          export GRADLE_ENTERPRISE_CACHE_PASSWORD="$GRADLE_ENTERPRISE_CACHE_PASSWORD"
+          export GRADLE_ENTERPRISE_ACCESS_KEY="$GRADLE_ENTERPRISE_SECRET_ACCESS_KEY"
+          echo "Starting release for next version"
```
sjohnr, line 72: As mentioned below, I feel like capturing the next version (either in a variable or a task output) really unlocks the potential for what we would do in this placeholder task.

sjohnr (commented):
This looks great! See discussion comments below.

buildSrc/src/main/java/org/springframework/gradle/github/milestones/GitHubMilestoneNextVersionDueTodayTask.java
eleftherias: I agree it makes sense to split it up into 2 tasks. I was just talking to Rob and he suggested looking into output / input files, since that is a standard way to share information between tasks. I'll start with these updates over the next few days.

eleftherias: As you suggested I split it up into two tasks:
1. `gitHubNextReleaseMilestone` which outputs the next version to a file called `next-release.yml`
2. `gitHubCheckMilestoneDueToday` which depends on `gitHubNextReleaseMilestone` and uses the generated file `next-release.yml` as input and fails the build if the given/calculated version is not due today

.github/workflows/release-next-version.yml
eleftherias: The task that performs the release could depend on the task `gitHubNextReleaseMilestone`, similar to how `gitHubCheckMilestoneDueToday` does it.

.github/workflows/release-next-version.yml
```diff
@@ -0,0 +1,91 @@
+name: CI
+
+on:
+  push:
+  schedule:
+    - cron: '0 0 * * MON' # Every Monday
```
sjohnr, line 6: I believe the [`schedule` trigger](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) runs only on the default branch. I'm thinking we may need a higher level workflow to trigger these builds on each branch.

Should we solve that as part of this PR or go ahead and merge this and work on that separately? If separately, perhaps remove the schedule and only include the manual trigger here?

I personally think the manual trigger adds value on other branches even without a scheduled run!

sjohnr (approved):
Thanks @eleftherias! I've tested this and it works very well! See one comment below, but beyond that this seems ready to merge and add to the 5.7 branch as well.

.github/workflows/release-next-version.yml
eleftherias: At the moment the only way to run a scheduled workflow on the 5.7.x branch is to add another workflow, for example `release-next-version-5.7.x.yml`, to the main branch. This isn't ideal since the CI code won't be tied to the source code, they'll be on different branches.
The `release-next-version-5.7.x.yml` would look like this:
```
# ...
jobs:
  # ...
  release:
    name: Release next version
    needs: [check_release_due]
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
        with:
          ref: 5.7.x
# ...
```
I would like to hold off on adding this until we have a complete release pipeline, in case GitHub has added some new functionality by then.

The relevant GitHub Actions issue is [here](https://github.community/t/scheduled-builds-of-non-default-branch/16306).

sjohnr: Thanks! I agree, let's hold off. Just wanted to note it for future me/you/anyone else.

PR #10778, opened 2022-01-25
BREACH attack protection for CSRF tokens

sjohnr:


web/src/main/java/org/springframework/security/web/csrf/XorCsrfTokenRepository.java
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	/**
+	 * Wrap an underlying {@link CsrfTokenRepository} with the ability to provide
+	 * {@link CsrfToken} objects that are salted with a random value on every request.
+	 * @param delegate The {@link CsrfTokenRepository} to delegate to for persistence
+	 * @return An {@link XorCsrfTokenRepository} that wraps the given
+	 * {@link CsrfTokenRepository}
+	 */
+	public static CsrfTokenRepository wrap(CsrfTokenRepository delegate) {
```
sjohnr, line 89: Note: I'm unsure whether we want a static method or just use the constructor since historically other implementations have just had a `public` constructor.

docs/modules/ROOT/pages/features/exploits/csrf.adoc
```diff
@@ -412,3 +412,12 @@ Overriding the HTTP method occurs in a filter.
 That filter must be placed before Spring Security's support.
 Note that overriding happens only on a `post`, so this is actually unlikely to cause any real problems.
 However, it is still best practice to ensure that it is placed before Spring Security's filters.
+
+[[csrf-considerations-breach-mitigation]]
+==== BREACH Security Exploit Mitigation
+
+When a web server has HTTP compression enabled, web applications hosted on that server may be vulnerable to a https://en.wikipedia.org/wiki/BREACH[BREACH attack]. Any sensitive information displayed by the web application could become compromised, including a CSRF token. As part of a defense in depth strategy, Spring Security provides the `XorCsrfTokenRepository` implementation that wraps another `CsrfTokenRepository`. When using `XorCsrfTokenRepository`, a random string is always returned for the token value, which makes it more difficult to guess the actual length of each response payload when using HTTP compression.
```
rwinch, line 419: New lines per sentence make it easier to diff when changes are made.

web/src/main/java/org/springframework/security/web/csrf/XorCsrfTokenRepository.java
rwinch: I'd just keep the constructor. Static methods are useful when you have lots of constructors and need more descriptive ways of creating an instance. It is also useful when the constructor can take the same number of arguments/types but the arguments are different. For example, `static ABC createFromToken(String)` and `static AbC createFromUrl(String)` are necessary because the number of arguments and types of the arguments are the same, but the value provided is different meaning.

web/src/test/java/org/springframework/security/web/csrf/XorCsrfTokenRepositoryTests.java
```diff
@@ -0,0 +1,175 @@
...
+		assertThat(csrfToken1.matches(token1)).isTrue();
+		assertThat(csrfToken1.matches(token2)).isTrue();
+		assertThat(csrfToken2.matches(token1)).isTrue();
+		assertThat(csrfToken2.matches(token2)).isTrue();
+	}
+
+	@Test
+	public void matchesWhenTokensNotEqualThenFalse() {
+		given(this.delegate.generateToken(this.request)).willAnswer(
+				(invocation) -> new DefaultCsrfToken("headerName", "parameterName", UUID.randomUUID().toString()));
```
rwinch, line 162: Avoid use of random as much as possible because it slows the tests down and we want tests to be consistent. If you want a random value, generate one and copy the value into the test.

rwinch (changes requested):
I've reviewed this inline.

rwinch (changes requested):
We have spoken offline and are going to explore some alternatives. 

Now that it is implemented, something that doesn't sit well is the fact that the repository is performing logic other than persistence. I'm wondering if we need to put the logic of reading the request and writing the request attributes into another API that can do the secure random. This also has the advantage of not impacting persistence and only changing the token a single time per request.

We will explore if [https://github.com/spring-projects/spring-security/blob/6dbd88a5a4857c279aed681529[…]main/java/org/springframework/security/web/csrf/CsrfFilter.java](https://github.com/spring-projects/spring-security/blob/6dbd88a5a4857c279aed681529b584e94b62cfa1/web/src/main/java/org/springframework/security/web/csrf/CsrfFilter.java#L110-L111)
and
[https://github.com/spring-projects/spring-security/blob/6dbd88a5a4857c279aed681529[…]main/java/org/springframework/security/web/csrf/CsrfFilter.java](https://github.com/spring-projects/spring-security/blob/6dbd88a5a4857c279aed681529b584e94b62cfa1/web/src/main/java/org/springframework/security/web/csrf/CsrfFilter.java#L120-L123)
were placed into a new API.

@sjohnr has agreed to a branches that demonstrate the ideas.

sjohnr:
@rwinch I have pushed a branch ([draft 1](https://github.com/spring-projects/spring-security/compare/5.8.x...sjohnr:spring-security:gh-4001-draft-1-functional?expand=1)) based on our discussion, and looking for general feedback on the approach. This draft uses functional hooks.

The next approach we could try would be to introduce a new interface. I don't have a good idea for what it would be called but let me know if you would like to see that approach next, or if you have another idea for the next branch.

rwinch:
I think this is a good start. I agree with you that we should create specific interfaces. An interesting observation is that CsrfAuthenticationStrategy can reuse the logic for setting the CSRF on an attribute.

I think that the names you have used could be used as a valid interface names. RequestAttributeHandler and CsrfTokenRequestResolver (we might need to play with this a bit).  We should have a single class implement both APIs since they are distinct jobs, but need to be coordinated.

sjohnr:
Replaced by gh-11731

PR #10800, opened 2022-02-02
Fix broken link to SAML2 login example

kdombeck:
Fix this link to point to 
https://github.com/spring-projects/spring-security-samples/tree/5.6.x/servlet/spring-boot/java/saml2/login 
instead of 
https://github.com/spring-projects/spring-security-samples/tree/5.6.x/servlet/spring-boot/java/saml2-login


sjohnr:
Thanks very much @kdombeck!

PR #10808, opened 2022-02-04
Apply defaults

ab0159w:
Not all defaults are overridden in provided sample, which did not work


ab0159w:
> Thanks for reaching out @ab0159w. Could you share some more information about your use-case and why the given configuration did not work for you? Given my understanding, it is working as I would expect.

Angular application calling spring cloud gateway secured with spring security (reactive) and keycloak. Will investigate further later in the week to see if I can see the exact cause.

eleftherias:
Thanks @ab0159w, a sample that reproduces the problem would be especially helpful.

eleftherias:
@ab0159w Have you had the chance to investigate this further?

ab0159w:
@eleftherias Sorry, have not investigated further yet.

ab0159w:
@eleftherias removed the .applyPermitDefaultValues() and it is working for me, must have been allowed-origins or allowed-headers that was causing the issue initially. Closing PR.

Issue #10798, opened 2022-02-02
DefaultLoginPageGeneratingFilter - fix incomplete localization of the default login page


razilein:
**Expected Behavior**

All labels and buttons in the default login page should be localized on the basis of message.properties

**Current Behavior**

When using another language than english only errormessages and tooltips are localized.
The labels, placeholder and button-text are staying english.

Thats because these are harcoded. For instance "Please sign in".
https://github.com/spring-projects/spring-security/blob/main/web/src/main/java/org/springframework/security/web/authentication/ui/DefaultLoginPageGeneratingFilter.java#L221

**Context**

The localization is incomplete and always stays english.

sjohnr:
Hi @razilein, thanks for your interest in the project. Would you be interested in submitting a pull request for this issue? I think it would be really nice to have full localization support on the default login page!

Also, have you tried implementing a custom login page, such as in [this sample](https://github.com/spring-projects/spring-security-samples/tree/main/servlet/java-configuration/authentication/username-password)?

razilein:
Hello @sjohnr,

yes I had a custom login page before, but I switched to the default-one.
My custom login page had no major differences to the default one.

I struggled a bit with the options maximumSessions and maxSessionsPreventsLogin with my custom page.
With the default login page this feature worked fine for me, so I switched to it and want to keep it.

razilein:
I've created #10811 to solve my problem.

sjohnr:
Closing this issue for now. Please see [this comment](https://github.com/spring-projects/spring-security/pull/10811#issuecomment-1041923727) for an explanation.

PR #10811, opened 2022-02-05
Allow complete localization for default login page

razilein:
* localize all labels and buttons
* add german messagekeys

Closes gh-10798


inpercima (commented):
Looks good to me.

sjohnr:
Hi @razilein. First, I want to thank you for taking the time to provide this PR! It's great to have community members such as you so willing to contribute!

I wanted to let you know that the Spring Security team has discussed possible improvements such as this one to the default login page, and we have decided not to proceed. The reason is that, while it seems very nice to provide customization hooks, the default login page is provided for convenience while getting started and is designed to be replaced with a custom one (see [here](https://github.com/spring-projects/spring-security-samples/blob/main/servlet/java-configuration/authentication/username-password/form/src/main/resources/templates/login.html)). Further, the topic of localization is large enough that it's not something we'll be pursuing at this time given other priorities. With that in mind, we'd like to keep the default login page filter as simple as possible and not include localization support.

I'm going to close this PR and the associated issue for now, but we will most certainly keep the door open for further discussion. I welcome you and the rest of the community to provide feedback, and we can revisit this topic if necessary.

PR #10812, opened 2022-02-05
Change in SimpleGrantedAuthority#equals

croissant676:
Prevent NPE when calling equals with parameter null


jzheaux:
Thanks, @croissant676, for the contribution. For it to be accepted, please do the following:

* Ensure it meets the formatting guidelines (check out the project locally and run `./gradlew format check`)
* Add a test that fails before your fix, but passes after your fix

jzheaux:
@croissant676, looking at this more closely, it appears to be unnecessary. In fact, when I create a test like so:

```java
assertThat(new SimpleGrantedAuthority("ROLE_TEST").equals(null)).isFalse()
```

the test passes just fine since the `instanceof` returns `false`.

Because of that, I'm going to decline this PR. That said, if you can add a test that demonstrates the need for the change, I'm happy to reopen.

Issue #10695, opened 2022-01-11
Make source code compatible with JDK 8


marcusdacoregio:
The code in the `5.7.x` branch is requiring JDK 11 https://github.com/spring-projects/spring-security/blob/f04cd641b035fc1a3bbed671cc84a3c7c19c759a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java#L162

We should fix this and the build should fail if using a language feature that is not compatible with JDK 8.

marcusdacoregio:
Fixed via https://github.com/spring-projects/spring-security/commit/60ed3602f6281d1a34c643484dfcb3440e2243d5

marcusdacoregio:
Reopening this because the solution was a workaround to enable teams from moving forward.

The optimal solution could be to use Gradle Toolchain or any equivalent to tell which JDK to use to build which module.

PR #10817, opened 2022-02-08
Build modules using Java 8

marcusdacoregio:
This PR does the following:

- Revert the commit https://github.com/spring-projects/spring-security/commit/60ed3602f6281d1a34c643484dfcb3440e2243d5 that was used as a workaround;
- Fix https://github.com/spring-projects/spring-security/issues/10695 by creating a `static` method to initialize the handlers;
- Use Java 8 for all modules but OpenSAML4;
- Because the `Saml2Login/LogoutConfigurerTests` are using some OpenSAML4 classes, these tests have to be compiled using Java 11. For this, those tests were removed from the `compileTestJava` task and a new Gradle task `compileSaml2TestJava` is being used to compile them;
- Another task `saml2Tests` is being used to run the tests compiled by `compileSaml2TestJava`.
- Instead of using the `ModuleDescriptor` to retrieve the OpenSAML version, now we are checking if the class `org.opensaml.core.xml.persist.impl.PassthroughSourceStrategy` exists in the classpath. This class was introduced in OpenSAML4.

### DIFF
--- a/config/spring-security-config.gradle
+++ b/config/spring-security-config.gradle
@@ -113,7 +113,6 @@ dependencies {
 	testRuntimeOnly 'org.hsqldb:hsqldb'
 }
 
-
 rncToXsd {
 	rncDir = file('src/main/resources/org/springframework/security/config/')
 	xsdDir = rncDir
@@ -130,3 +129,33 @@ tasks.withType(KotlinCompile).configureEach {
 }
 
 build.dependsOn rncToXsd
+
+compileTestJava {
+	exclude "org/springframework/security/config/annotation/web/configurers/saml2/**", "org/springframework/security/config/http/Saml2*"
+}
+
+task compileSaml2TestJava(type: JavaCompile) {
+	javaCompiler = javaToolchains.compilerFor {
+		languageVersion = JavaLanguageVersion.of(11)
+	}
+	source = sourceSets.test.java.srcDirs
+	include "org/springframework/security/config/annotation/web/configurers/saml2/**", "org/springframework/security/config/http/Saml2*"
+	classpath = sourceSets.test.compileClasspath
+	destinationDirectory = new File("${buildDir}/classes/java/test")
+	options.sourcepath = sourceSets.test.java.getSourceDirectories()
+}
+
+task saml2Tests(type: Test) {
+	javaLauncher = javaToolchains.launcherFor {
+		languageVersion = JavaLanguageVersion.of(11)
+	}
+	filter {
+		includeTestsMatching "org.springframework.security.config.annotation.web.configurers.saml2.*"
+	}
+	useJUnitPlatform()
+	dependsOn compileSaml2TestJava
+}
+
+tasks.named('test') {
+	finalizedBy 'saml2Tests'
+}
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
@@ -32,8 +32,6 @@ import org.springframework.security.config.annotation.web.configurers.AbstractHt
 import org.springframework.security.config.annotation.web.configurers.CsrfConfigurer;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.saml2.provider.service.authentication.AbstractSaml2AuthenticationRequest;
-import org.springframework.security.saml2.provider.service.authentication.OpenSaml4AuthenticationProvider;
-import org.springframework.security.saml2.provider.service.authentication.OpenSaml4AuthenticationRequestFactory;
 import org.springframework.security.saml2.provider.service.authentication.OpenSamlAuthenticationProvider;
 import org.springframework.security.saml2.provider.service.authentication.OpenSamlAuthenticationRequestFactory;
 import org.springframework.security.saml2.provider.service.authentication.Saml2AuthenticationRequestFactory;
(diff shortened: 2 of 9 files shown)


config/spring-security-config.gradle
```diff
@@ -130,3 +129,33 @@ tasks.withType(KotlinCompile).configureEach {
 }
 
 build.dependsOn rncToXsd
+
+compileTestJava {
```
jzheaux, line 133: Is it true that we don't want to run these tests against JDK 8? It seems to me that we want to run the tests twice: Once for JDK 8 and once for JDK 11.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
```diff
@@ -319,12 +318,11 @@ private Saml2AuthenticationRequestFactory getAuthenticationRequestFactory(B http
 		if (resolver != null) {
 			return resolver;
 		}
-		if (version().startsWith("4")) {
-			return new OpenSaml4AuthenticationRequestFactory();
-		}
-		else {
-			return new OpenSamlAuthenticationRequestFactory();
+		if (ClassUtils.isPresent("org.opensaml.core.xml.persist.impl.PassthroughSourceStrategy",
```
jzheaux, line 321: I wonder if it would still be appropriate to have a `version()` method for readability. It would simply determine the version via `ClassUtils.isPresent`.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
```diff
@@ -319,12 +318,11 @@ private Saml2AuthenticationRequestFactory getAuthenticationRequestFactory(B http
...
+				Version.class.getClassLoader())) {
```
jzheaux, line 322: It feels a little odd to be calling for `Version`'s classloader given that we aren't working with the `Version` class. I wonder if passing `null` here would be better.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LogoutConfigurer.java
```diff
@@ -470,13 +462,14 @@ private Saml2LogoutResponseValidator logoutResponseValidator() {
...
-			if (this.logoutResponseResolver == null) {
-				if (version().startsWith("4")) {
-					return new OpenSaml4LogoutResponseResolver(relyingPartyRegistrationResolver);
-				}
-				return new OpenSaml3LogoutResponseResolver(relyingPartyRegistrationResolver);
+			if (this.logoutResponseResolver != null) {
+				return this.logoutResponseResolver;
 			}
-			return this.logoutResponseResolver;
+			if (ClassUtils.isPresent("org.opensaml.core.xml.persist.impl.PassthroughSourceStrategy",
```
jzheaux, line 468: It might be more readable to have one place for all the decision logic. Right now there is some here and some in `OpenSaml4LogoutSupportFactory`

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LogoutConfigurer.java
```diff
@@ -519,4 +512,38 @@ public void logout(HttpServletRequest request, HttpServletResponse response, Aut
 
 	}
 
+	private static class OpenSaml4LogoutSupportFactory {
```
jzheaux, line 515: Is this class still necessary? I think it was necessary when it was acting as a classpath guard. However, since you are using reflection now, I don't think this class offers this same benefit.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/NamespaceHttpX509Tests.java
```diff
@@ -21,13 +21,11 @@
...
 import java.security.cert.X509Certificate;
 
-import javax.security.auth.x500.X500Principal;
 import javax.servlet.http.HttpServletRequest;
 
-import org.bouncycastle.asn1.x500.X500Name;
-import org.bouncycastle.asn1.x500.style.BCStyle;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+import sun.security.x509.X500Name;
```
jzheaux, line 28: Can you clarify how these changes pertain to the PR?

jzheaux (changes requested):
Thanks, @marcusdacoregio! I've left my feedback inline.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/NamespaceHttpX509Tests.java
marcusdacoregio: Those changes came from the revert that I did to the old commit, where we weren't using toolchain to build the modules

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LogoutConfigurer.java
marcusdacoregio: Those classes make the code cleaner by moving the try-catch from the reflective operations to another method

config/spring-security-config.gradle
marcusdacoregio: We want to run they using JDK 11 only because they are using OpenSAML4 classes

rwinch (approved):

PR #10833, opened 2022-02-14
Java 17, removed deprecated User.withDefaultPasswordEncoder()

hduerkop:


marcusdacoregio:
Hi @hduerkop. 

As the javadoc for the `User#withDefaultPasswordEncoder()` method states (highlight by me):

> Using this method is not considered safe for production, but is acceptable for demos and getting started. For production purposes, ensure the password is encoded externally. See the method Javadoc for additional details. **There are no plans to remove this support. It is deprecated to indicate that this is considered insecure for production purposes.**

We do not want to remove this support, the deprecation is to indicate that it is insecure for production purposes.

I'm closing this but feel free to create a ticket and discuss more about it if you want.

PR #10845, opened 2022-02-17
Add gradle task to update project version in gradle.properties

eleftherias:
Running `./gradlew updateProjectVersion` will update the version in `gradle.properties` to the next milestone version.

Running `./gradlew updateProjectVersion -Pcommit=true` will update the version and create a commit with the message "Release VERSION", where VERSION is the next milestone version.

**Warning**: Make sure to undo the commit to prevent accidentally pushing it!


sjohnr (approved):
Looks good, @eleftherias! I'll work next on what happens after this is committed, specifically watching for the maven central release and then creating a GitHub release. Shall I also update the scheduled task that runs on Mondays to call this new task, or do you want to do that?

eleftherias:
Thanks @sjohnr!
I'll go ahead and update the scheduled task. 
I will add a check for `./gradlew gitHubCheckMilestoneHasNoOpenIssues` to the pipeline as per gh-10451 before committing and pushing the version update.

PR #10849, opened 2022-02-19
Add support for validation of InResponseTo attribute when validating SAML2 responses

fast-reflexes:
… a SAML2 response given that this attribute is set and that the request can be found.

https://github.com/spring-projects/spring-security/issues/9174

What do you think so far?

Left to do that I know of:
* Perhaps add some test
* Go through submission guidelines again

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/core/Saml2ErrorCodes.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/core/Saml2ErrorCodes.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -30,6 +30,13 @@ public interface Saml2ErrorCodes {
 	 * https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf#page=46
 	 */
 	String UNKNOWN_RESPONSE_CLASS = "unknown_response_class";
+
+	/**
+	 * The serialized AuthNRequest could not be deserialized correctly.
+	 *
+	 * @since 5.7
+	 */
+	String MALFORMED_REQUEST_DATA = "malformed_request_data";
 
 	/**
 	 * The response data is malformed or incomplete. An invalid XML object was received,
@@ -115,5 +122,12 @@ public interface Saml2ErrorCodes {
 	 * correspond to any relying party registration.
 	 */
 	String RELYING_PARTY_REGISTRATION_NOT_FOUND = "relying_party_registration_not_found";
+
+	/**
+	 * The InResponseTo content of the response does not match the ID of the AuthNRequest.
+	 *
+	 * @since 5.7
+	 */
+	String INVALID_IN_RESPONSE_TO = "invalid_in_response_to";
 
 }
--- a/saml2/saml2-service-provider/src/opensaml4Main/java/org/springframework/security/saml2/provider/service/authentication/OpenSaml4AuthenticationProvider.java
+++ b/saml2/saml2-service-provider/src/opensaml4Main/java/org/springframework/security/saml2/provider/service/authentication/OpenSaml4AuthenticationProvider.java
@@ -37,6 +37,7 @@ import org.apache.commons.logging.LogFactory;
 import org.opensaml.core.config.ConfigurationService;
 import org.opensaml.core.xml.XMLObject;
 import org.opensaml.core.xml.config.XMLObjectProviderRegistry;
+import org.opensaml.core.xml.config.XMLObjectProviderRegistrySupport;
 import org.opensaml.core.xml.schema.XSAny;
 import org.opensaml.core.xml.schema.XSBoolean;
 import org.opensaml.core.xml.schema.XSBooleanValue;
@@ -57,13 +58,17 @@ import org.opensaml.saml.saml2.assertion.impl.DelegationRestrictionConditionVali
 import org.opensaml.saml.saml2.core.Assertion;
 import org.opensaml.saml.saml2.core.Attribute;
 import org.opensaml.saml.saml2.core.AttributeStatement;
+import org.opensaml.saml.saml2.core.AuthnRequest;
 import org.opensaml.saml.saml2.core.AuthnStatement;
 import org.opensaml.saml.saml2.core.Condition;
 import org.opensaml.saml.saml2.core.EncryptedAssertion;
 import org.opensaml.saml.saml2.core.OneTimeUse;
 import org.opensaml.saml.saml2.core.Response;
 import org.opensaml.saml.saml2.core.StatusCode;
+import org.opensaml.saml.saml2.core.Subject;
 import org.opensaml.saml.saml2.core.SubjectConfirmation;
+import org.opensaml.saml.saml2.core.SubjectConfirmationData;
+import org.opensaml.saml.saml2.core.impl.AuthnRequestUnmarshaller;
 import org.opensaml.saml.saml2.core.impl.ResponseUnmarshaller;
 import org.opensaml.saml.saml2.encryption.Decrypter;
 import org.opensaml.saml.security.impl.SAMLSignatureProfileValidator;
@@ -85,6 +90,7 @@ import org.springframework.security.saml2.core.Saml2Error;
 import org.springframework.security.saml2.core.Saml2ErrorCodes;
 import org.springframework.security.saml2.core.Saml2ResponseValidatorResult;
 import org.springframework.security.saml2.provider.service.registration.RelyingPartyRegistration;
+import org.springframework.security.saml2.provider.service.registration.Saml2MessageBinding;
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.StringUtils;
@@ -349,6 +355,37 @@ public final class OpenSaml4AuthenticationProvider implements AuthenticationProv
 		this.responseAuthenticationConverter = responseAuthenticationConverter;
 	}
 
+	private static Saml2ResponseValidatorResult validateInResponseTo(AbstractSaml2AuthenticationRequest storedRequest,
+			String inResponseTo) {
+		if (!StringUtils.hasText(inResponseTo)) {
+			return Saml2ResponseValidatorResult.success();
+		}
+		AuthnRequest request;
+		try {
+			request = parseRequest(storedRequest);
+		}
+		catch (Exception ex) {
+			String message = "The stored AuthNRequest could not be properly deserialized [" + ex.getMessage() + "]";
+			return Saml2ResponseValidatorResult
+					.failure(new Saml2Error(Saml2ErrorCodes.MALFORMED_REQUEST_DATA, message));
+		}
+		if (request == null) {
+			String message = "The response contained an InResponseTo attribute [" + inResponseTo + "]"
+					+ " but no saved AuthNRequest request was found";
+			return Saml2ResponseValidatorResult
+					.failure(new Saml2Error(Saml2ErrorCodes.INVALID_IN_RESPONSE_TO, message));
+		}
+		else if (!request.getID().equals(inResponseTo)) {
+			String message = "The InResponseTo attribute [" + inResponseTo + "] does not match the ID of the "
+					+ "AuthNRequest [" + request.getID() + "]";
+			return Saml2ResponseValidatorResult
+					.failure(new Saml2Error(Saml2ErrorCodes.INVALID_IN_RESPONSE_TO, message));
+		}
+		else {
+			return Saml2ResponseValidatorResult.success();
+		}
+	}
+
 	/**
 	 * Construct a default strategy for validating the SAML 2.0 Response
 	 * @return the default response validator strategy
@@ -365,6 +402,10 @@ public final class OpenSaml4AuthenticationProvider implements AuthenticationProv
 						response.getID());
 				result = result.concat(new Saml2Error(Saml2ErrorCodes.INVALID_RESPONSE, message));
 			}
+
+			String inResponseTo = response.getInResponseTo();
+			result = result.concat(validateInResponseTo(token.getAuthenticationRequest(), inResponseTo));
+
 			String issuer = response.getIssuer().getValue();
 			String destination = response.getDestination();
 			String location = token.getRelyingPartyRegistration().getAssertionConsumerServiceLocation();
@@ -447,7 +488,7 @@ public final class OpenSaml4AuthenticationProvider implements AuthenticationProv
 		try {
 			Saml2AuthenticationToken token = (Saml2AuthenticationToken) authentication;
 			String serializedResponse = token.getSaml2Response();
-			Response response = parse(serializedResponse);
+			Response response = parseResponse(serializedResponse);
 			process(token, response);
 			AbstractAuthenticationToken authenticationResponse = this.responseAuthenticationConverter
 					.convert(new ResponseToken(response, token));
@@ -469,7 +510,7 @@ public final class OpenSaml4AuthenticationProvider implements AuthenticationProv
 		return authentication != null && Saml2AuthenticationToken.class.isAssignableFrom(authentication);
 	}
 
-	private Response parse(String response) throws Saml2Exception, Saml2AuthenticationException {
+	private Response parseResponse(String response) throws Saml2Exception, Saml2AuthenticationException {
 		try {
 			Document document = this.parserPool
 					.parse(new ByteArrayInputStream(response.getBytes(StandardCharsets.UTF_8)));
@@ -479,6 +520,28 @@ public final class OpenSaml4AuthenticationProvider implements AuthenticationProv
 		catch (Exception ex) {
 			throw createAuthenticationException(Saml2ErrorCodes.MALFORMED_RESPONSE_DATA, ex.getMessage(), ex);
 		}
+	}
+
+	private static AuthnRequest parseRequest(AbstractSaml2AuthenticationRequest request) throws Exception {
+		if (request == null) {
+			return null;
+		}
+		String samlRequest = request.getSamlRequest();
+		if (!StringUtils.hasText(samlRequest)) {
+			return null;
+		}
+		if (request.getBinding() == Saml2MessageBinding.REDIRECT) {
+			samlRequest = Saml2Utils.samlInflate(Saml2Utils.samlDecode(samlRequest));
+		}
+		else {
+			samlRequest = new String(Saml2Utils.samlDecode(samlRequest), StandardCharsets.UTF_8);
+		}
+		Document document = XMLObjectProviderRegistrySupport.getParserPool()
+				.parse(new ByteArrayInputStream(samlRequest.getBytes(StandardCharsets.UTF_8)));
+		Element element = document.getDocumentElement();
+		AuthnRequestUnmarshaller unmarshaller = (AuthnRequestUnmarshaller) XMLObjectProviderRegistrySupport
+				.getUnmarshallerFactory().getUnmarshaller(AuthnRequest.DEFAULT_ELEMENT_NAME);
+		return (AuthnRequest) unmarshaller.unmarshall(element);
 	}
 
 	private void process(Saml2AuthenticationToken token, Response response) {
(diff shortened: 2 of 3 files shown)


saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/core/Saml2ErrorCodes.java
```diff
@@ -116,4 +116,9 @@
 	 */
 	String RELYING_PARTY_REGISTRATION_NOT_FOUND = "relying_party_registration_not_found";
 
+	/**
+	 * The InResponseTo content of the response does not match the ID of the AuthNRequest.
```
jzheaux, line 120: Please add `@since 5.7`

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/AbstractSaml2AuthenticationRequest.java
```diff
@@ -89,6 +93,14 @@ public String getAuthenticationRequestUri() {
 		return this.authenticationRequestUri;
 	}
 
+	/**
+	 * Returns the id of this AuthNRequest.
+	 * @return the id of this request
+	 */
```
jzheaux, line 99: Please add `@since 5.7` here and to any other public methods, fields, etc. that you are introducing.

saml2/saml2-service-provider/src/opensaml3Main/java/org/springframework/security/saml2/provider/service/authentication/OpenSamlAuthenticationProvider.java
```diff
@@ -554,12 +554,19 @@ else if (logger.isDebugEnabled()) {
 			Response response = responseToken.getResponse();
 			Saml2AuthenticationToken token = responseToken.getToken();
 			Saml2ResponseValidatorResult result = Saml2ResponseValidatorResult.success();
+			AbstractSaml2AuthenticationRequest request = token.getAuthenticationRequest();
```
jzheaux, line 557: Since OpenSAML 3 is EOL, I'd prefer to only add this support to `OpenSaml4AuthenticationProvider`.

saml2/saml2-service-provider/src/opensaml4Main/java/org/springframework/security/saml2/provider/service/authentication/OpenSaml4AuthenticationProvider.java
```diff
@@ -359,12 +359,19 @@ public void setResponseAuthenticationConverter(
...
 			Saml2ResponseValidatorResult result = Saml2ResponseValidatorResult.success();
+			AbstractSaml2AuthenticationRequest request = token.getAuthenticationRequest();
 			String statusCode = getStatusCode(response);
 			if (!StatusCode.SUCCESS.equals(statusCode)) {
 				String message = String.format("Invalid status [%s] for SAML response [%s]", statusCode,
 						response.getID());
 				result = result.concat(new Saml2Error(Saml2ErrorCodes.INVALID_RESPONSE, message));
 			}
+			String inResponseTo = response.getInResponseTo();
+			if (StringUtils.hasText(inResponseTo) && request != null && !inResponseTo.equals(request.getId())) {
```
jzheaux, line 370: Typically in Spring Security, if an if statement has 3+ clauses, we break it up into a separate method, for example (pseudocode):

```java
private static Saml2ResponseValidationResult validateInResponseTo(Response response, AbstractAuthenticationRequest request) {
    String inReponseTo = response.getInResponseTo();
    if (!StringUtils.hasText(inResponseTo)) {
        return Saml2ResponseValidationResult.success();
    }
    if (request == null) {
        return Saml2ResponseValidationResult.success();
    }
    // ....
    return Saml2ResponseValidatorResult.error(new Saml2Error(...));
}
```

jzheaux (changes requested):
Thanks, @fast-reflexes! I've left some feedback inline.

Additionally, will you please format your commit message like this:

```
Commit Title

Closes gh-9174
```

saml2/saml2-service-provider/src/opensaml4Main/java/org/springframework/security/saml2/provider/service/authentication/OpenSaml4AuthenticationProvider.java
```diff
@@ -359,12 +359,26 @@ public void setResponseAuthenticationConverter(
...
 			Saml2AuthenticationToken token = responseToken.getToken();
 			Saml2ResponseValidatorResult result = Saml2ResponseValidatorResult.success();
+			AbstractSaml2AuthenticationRequest request = token.getAuthenticationRequest();
 			String statusCode = getStatusCode(response);
 			if (!StatusCode.SUCCESS.equals(statusCode)) {
 				String message = String.format("Invalid status [%s] for SAML response [%s]", statusCode,
 						response.getID());
 				result = result.concat(new Saml2Error(Saml2ErrorCodes.INVALID_RESPONSE, message));
 			}
+			String inResponseTo = response.getInResponseTo();
```
jzheaux, line 369: We probably should check the `inResponseTo` value in each assertion as well in `createAssertionValidator`.

saml2/saml2-service-provider/src/opensaml4Main/java/org/springframework/security/saml2/provider/service/web/authentication/OpenSaml4AuthenticationRequestResolver.java
```diff
@@ -55,8 +55,10 @@ public OpenSaml4AuthenticationRequestResolver(RelyingPartyRegistrationResolver r
 	@Override
 	public <T extends AbstractSaml2AuthenticationRequest> T resolve(HttpServletRequest request) {
 		return this.authnRequestResolver.resolve(request, (registration, authnRequest) -> {
-			authnRequest.setIssueInstant(Instant.now(this.clock));
+			Instant issuingInstant = Instant.now(this.clock);
```
jzheaux, line 58: I appreciate the extra effort to add a timing-based validation, though I'm concerned that this may be too opinionated as it will necessarily introduce timeout configurations. Would you mind if we stay focused on `InResponseTo` for now? You can open a separate ticket to discuss the timeout feature, if you like.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/AbstractSaml2AuthenticationRequest.java
```diff
@@ -89,6 +98,24 @@ public String getAuthenticationRequestUri() {
 		return this.authenticationRequestUri;
 	}
 
+	/**
+	 * Returns the id of this AuthNRequest.
+	 * @return the id of this request
+	 * @since 5.7
+	 */
+	public String getId() {
```
jzheaux, line 106: Thinking about this more, if the only place the id is needed is in OpenSAML-specific classes, then it's reasonable to leave this property off and have `OpenSaml4AuthenticationProvider` parse the `AuthnRequest` using OpenSAML and read the ID. This serves the function of keeping the public API as small as possible.

jzheaux (changes requested):
Thanks, @fast-reflexes! I've left some feedback inline.

In addition to that feedback, will you do me the favor of making your commit title just a tad shorter? ~50 characters is ideal. This helps when scanning through Git logs.

saml2/saml2-service-provider/src/opensaml4Main/java/org/springframework/security/saml2/provider/service/web/authentication/OpenSaml4AuthenticationRequestResolver.java
fast-reflexes: Cool, the related problem to `InRespnseTo` is that if we assume that the existence of a saved request indicates that we should validate `InResponseTo` in the response, then a saved stale request efficiently and improperly blocks IdP initiated logins (e.g., we first try to log in using SP initiated login which is for some reason aborted with the saved request still in the backend associated with our cookie, 15 minutes later we go to the IdP and try to login via a link to our app, now the app will reject the login attempt because there is a stale stored request which indicated that we should validate `InResponseTo` in the login attempt which is missing since it is a IdP initiated login. An alternative would be to only validate `InResponseTo` if it is present in the response but that kind of circumvents one of the main ideas behind it, namely that we should know when to expect an `InResponseTo` and when not. You think we should just ignore this situation and remove the eviction?

jzheaux: Good points. I think that if there is no `InResponseTo`, then validation should pass. You can see `OpenSamlLogoutResponseValidator#validateInResponseTo` for an example.

`InResponseTo` helps mitigate replay attacks, but this validation doesn't apply if the IdP is configured to do unsolicited login. So, if `InResponseTo` is missing, there's no problem unless it is evidence of misconfiguration. If it's misconfiguration, the IdP needs to be changed to not allow unsolicited responses and the SP needs to be customized to always reject missing `InResponseTo` responses.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/AbstractSaml2AuthenticationRequest.java
fast-reflexes: Hmm... can't we just store the `AuthnRequest` instead of the serialized message string in `AbstractSaml2AuthenticationRequest`? Can't really find anywhere where this string is used for anything particular and in the tests where it is used, it's deserialized into an `AuthnRequest` so it would make more sense to store the actual request... what do you think?

fast-reflexes: I mean of course the request string is used but what do you say about storing the `AuthnRequest` and then add a method `getSerializedSamlRequest()` to `AbstractSaml2AuthenticationRequest` which returns the serialized request for use in `SamlWebSsoAuthenticationRequestFilter` (so we move the serialization process to reside inside `AbstractSaml2AuthenticationRequest`). My concern is otherwise that the deserialization of the request in `OpenSaml4AuthenticationProvider` will be a bit hacky and cumbersome and to me, it makes more sense to store the domain class of the request instead of the serialized string :)

jzheaux: `AbstractSaml2AuthenticationRequest` cannot contain an `AuthnRequest` because `AbstractSaml2AuthenticationRequest` is not OpenSAML-specific.

jzheaux: > My concern is otherwise that the deserialization of the request in OpenSaml4AuthenticationProvider will be a bit hacky

It's not clear to me why this would be hacky. It does need to be serialized so it can go into the session. Whether it's Java or it's Spring, something is thus going to have to deserialize it.

fast-reflexes:
I updated based on your feedback. 

* What I mean is that converting the request to a string and then decoding it is less straightforward than storing the domain representation of a request (be it locked down to the OpenSAML implementation or some thin wrapper around it). Transforming it into a string is a preparation for the actual outgoing request and therefore something that I think should be done right before sending the actual request. It makes testing more cumbersome when the request is stored as a string.
* I really think we should have a flag / property for whether applications want to use validation of `InResponseTo` or not. Of course they can add their own validators and so on but it's more cumbersome than just opting in or out of such a feature.

These are my remarks as a programmer and a Spring user :) Here, I'm nothing but someone who wants to get more involved and learn to contribute so I respect your decisions to 100%, just letting you know what I think :)

Let me know what you think of the current state of the code :)

saml2/saml2-service-provider/src/opensaml4Main/java/org/springframework/security/saml2/provider/service/authentication/OpenSaml4AuthenticationProvider.java
```diff
@@ -349,6 +356,25 @@ public void setResponseAuthenticationConverter(
 		this.responseAuthenticationConverter = responseAuthenticationConverter;
 	}
 
+	private static Saml2ResponseValidatorResult validateInResponseTo(AuthnRequest request, String inResponseTo,
+			Saml2ResponseValidatorResult result) {
+		if (StringUtils.hasText(inResponseTo)) {
```
jzheaux, line 361: It would be nice to have less indentation to achieve greater readability, like in `OpenSamlLogoutResponseValidator#validateInResponseTo`.

What if you did

```java
if (!StringUtils.hasText(inResponseTo)) {
    return Saml2ResponseValidatorResult.success();
}

if (request == null) {
    return new Saml2Error(...);
}

if (inResponseTo.equals(request.getInResponseTo())) {
    return Saml2ResponseValidatorResult.success();
}

return new Saml2Error(...);
```

The Spring Security codebase tends to favor early returns over additional indentation.

saml2/saml2-service-provider/src/opensaml4Test/java/org/springframework/security/saml2/provider/service/authentication/OpenSaml4AuthenticationProviderTests.java
```diff
@@ -190,9 +192,7 @@ public void authenticateWhenMissingSubjectThenThrowAuthenticationException() {
 		Response response = response();
 		Assertion assertion = assertion();
 		assertion.setSubject(null);
-		TestOpenSamlObjects.signed(assertion, TestSaml2X509Credentials.assertingPartySigningCredential(),
-				RELYING_PARTY_ENTITY_ID);
-		response.getAssertions().add(assertion);
+		response.getAssertions().add(signed(assertion));
```
jzheaux, line 195: Thanks for thinking about making the other tests easier. Since they aren't part of the feature you are adding, will you please move these other test changes to a separate commit in this PR?

jzheaux (changes requested):
Thanks, @fast-reflexes! Glad this is taking shape. I've left some feedback inline.

jzheaux:
> What I mean is that converting the request to a string and then decoding it is less straightforward than storing the domain representation of a request (be it locked down to the OpenSAML implementation or some thin wrapper around it).

I see your point, but not how to apply it. `AuthnRequest` is not `Serializable`, so how would you expect the `AuthnRequest` to be able to be stored in the session?

> I really think we should have a flag / property for whether applications want to use validation of InResponseTo or not. 

We can always add this in a separate PR down the road. I think boolean properties should be added quite slowly and with more usage data than we have now. You can welcome to add a ticket and then we can see if the community votes for it.

saml2/saml2-service-provider/src/opensaml4Main/java/org/springframework/security/saml2/provider/service/authentication/OpenSaml4AuthenticationProvider.java
```diff
@@ -737,7 +801,14 @@ protected ValidationResult validateAddress(SubjectConfirmation confirmation, Ass
 				@Override
 				protected ValidationResult validateInResponseTo(SubjectConfirmation confirmation, Assertion assertion,
 						ValidationContext context, boolean required) {
-					// applications should validate their own in response to
+					String inResponseTo = confirmation.getSubjectConfirmationData().getInResponseTo();
```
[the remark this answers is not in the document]
fast-reflexes: Very good idea, didn't realize how this class worked until now. This will of course require us to ALWAYS parse the request twice since we won't know beforehand if there are `InResponseTo` properties in the assertions or not. You think that's ok?

saml2/saml2-service-provider/src/opensaml4Main/java/org/springframework/security/saml2/provider/service/authentication/OpenSaml4AuthenticationProvider.java
fast-reflexes: Affirmative! Couldn't find the method you were talking about in that class though but I understand what you mean!

fast-reflexes:
> 
> I see your point, but not how to apply it. `AuthnRequest` is not `Serializable`, so how would you expect the `AuthnRequest` to be able to be stored in the session?
> 
Aha.. so THAT'S the problem ... hmm .. ok, then I'll rest my case for now!

saml2/saml2-service-provider/src/opensaml4Main/java/org/springframework/security/saml2/provider/service/authentication/OpenSaml4AuthenticationProvider.java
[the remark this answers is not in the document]
jzheaux: I don't think that's quite true. I think you would only add the `SC_VALID_IN_RESPONSE_TO` parameter in the event that the assertion has a subject confirmation data has an `InResponseTo` attribute. For example, you can change the if statement on line 756 to check for the `InResponseTo` field. If it passes, parse the `AuthnRequest` and supply the id.

saml2/saml2-service-provider/src/opensaml4Main/java/org/springframework/security/saml2/provider/service/authentication/OpenSaml4AuthenticationProvider.java
```diff
@@ -685,13 +748,44 @@ private static Saml2AuthenticationException createAuthenticationException(String
...
+			AuthnRequest request = parseRequest(storedRequest);
+			requestId = (request != null) ? request.getID() : null;
+		}
+		catch (Exception ex) {
+		}
+		if (StringUtils.hasText(requestId)) {
+			context.put(SAML2AssertionValidationParameters.SC_VALID_IN_RESPONSE_TO, requestId);
+		}
+		else {
+			context.put(SAML2AssertionValidationParameters.SC_VALID_IN_RESPONSE_TO, "");
```
fast-reflexes, line 774: Here we set the empty string so that we know that the validation will fail (since the very prerequisite for even doing the validation is that the assertion has a non-empty `InResponseTo`). The reason for failing is either that there was no stored request, it could not be deserialized or it lacked id. You think this solution is ok?

fast-reflexes: Could trigger a false positive if we have multiple confirmations and one has the (invalid) empty string as `InResponseTo` whereas a later one has a proper one ...

fast-reflexes: Perhaps better to stay with the override anyways? 
* Won't be more code than is added by this solution
* Easier to understand what's going on
* Will cover all cases since we can control directly whether to consider valid or not

jzheaux: I think you can just leave it unset. The OpenSAML code works as follows:

- If no `InResponseTo` in subject confirmation -> valid
- Or, if no `SC_VALID_IN_RESPONSE_TO` set -> invalid
- Or, if `SC_VALID_IN_RESPONSE_TO` mismatches `InResponseTo` -> invalid

> Could trigger a false positive

If any assertions are wrong, then we should fail. We don't accept a response that is in response to multiple requests.

> override anyways

Ideally, we don't write code that we don't have to as this lowers maintenance. I don't see why it's easier to understand what is going on. The code that you have written is virtually the same as OpenSAML's `BearerSubjectConfirmationValidator#validateInResponseTo`.

fast-reflexes: Yeah of course, can be unset, I don't know what I was thinking... and what I meant by 

> Could trigger a false positive if we have multiple confirmations and one has the (invalid) empty string as 
> InResponseTo whereas a later one has a proper one ...

was that if an assertion has multiple subject confirmations (and I had set the `SC_VALID_IN_RESPONSE_TO`to the empty string), then the false positive would be a subject with 2 confirmations where ONE had the WRONG `InResponseTo` and one had the empty string as `InResponse` to. Such confirmation would falsely pass even though neither of the `InResponseTo` matches the request id (given that the request could not be unserialized so that `SC_VALID_IN_RESPONSE_TO` was set to the empty string). 

Since we can leave it unset, it should not be a problem. 

It seems it is enough with one confirmation passing the test on the subject (see `SAML20AssertionValidator::validateSubjectConfirmation`)

I've updated the code, let me know what you think!

fast-reflexes:
Great, thanks @jzheaux ! I looked into your changes to learn the style... what do you mean by 

`Refactored methods to have only immutable method parameters`

Just curious :)

Issue #10481, opened 2021-11-10
Add authorizeHttpRequests to Kotlin DSL


jzheaux:
#8900 introduced `HttpSecurity#authorizeHttpRequests` and the Kotlin DSL should have equivalent support.

PR #10895, opened 2022-02-21
Add authorizeHttpRequests to Kotlin DSL

ysavchen:
Closes gh-10481

Comparing to `AuthorizeRequestsDslTests`, the newly added `AuthorizeHttpRequestsDslTests` does not include the following test:
```
@Test
fun `request when secured by mvc path variables then responds based on path variable value`() {
    this.spring.register(MvcMatcherPathVariablesConfig::class.java).autowire()
    this.mockMvc.get("/user/user")
        .andExpect {
            status { isOk() }
        }

    this.mockMvc.get("/user/deny")
        .andExpect {
            status { isForbidden() }
        }
}

@EnableWebSecurity
@EnableWebMvc
open class MvcMatcherPathVariablesConfig {
    @Bean
    open fun securityFilterChain(http: HttpSecurity): SecurityFilterChain {
        http {
            authorizeHttpRequests {
                authorize("/user/{userName}", "#userName == 'user'")
            }
        }
        return http.build()
    }

    @RestController
    internal class PathController {
        @RequestMapping("/user/{user}")
        fun path(@PathVariable user: String) {
        }
    }
}
```
As far as I understand `authorizeHttpRequests` API does not provide SpEL support and this test seems obsolete.
But due to default value for access parameter, there is a mismatch between `authorizeRequests` and `authorizeHttpRequets` API:
```
autorhizeRequests {
    authorize("/user/{userName}", "#userName == 'user'")   //invokes authorize with pattern and access
}

authorizeHttpRequests {
    authorize("/user/{userName}", "#userName == 'user'")   //invokes authorize with pattern and servletPath as access type is not String anymore
}
```

### DIFF
--- a/config/src/main/kotlin/org/springframework/security/config/annotation/web/AbstractRequestMatcherDsl.kt
+++ b/config/src/main/kotlin/org/springframework/security/config/annotation/web/AbstractRequestMatcherDsl.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,6 +17,8 @@
 package org.springframework.security.config.annotation.web
 
 import org.springframework.http.HttpMethod
+import org.springframework.security.authorization.AuthorizationManager
+import org.springframework.security.web.access.intercept.RequestAuthorizationContext
 import org.springframework.security.web.util.matcher.AnyRequestMatcher
 import org.springframework.security.web.util.matcher.RequestMatcher
 
@@ -36,13 +38,24 @@ abstract class AbstractRequestMatcherDsl {
     protected data class MatcherAuthorizationRule(val matcher: RequestMatcher,
                                                   override val rule: String) : AuthorizationRule(rule)
 
+    protected data class MatcherAuthorizationManagerRule(val matcher: RequestMatcher,
+                                                         override val rule: AuthorizationManager<RequestAuthorizationContext>) : AuthorizationManagerRule(rule)
+
     protected data class PatternAuthorizationRule(val pattern: String,
                                                   val patternType: PatternType,
                                                   val servletPath: String? = null,
                                                   val httpMethod: HttpMethod? = null,
                                                   override val rule: String) : AuthorizationRule(rule)
 
+    protected data class PatternAuthorizationManagerRule(val pattern: String,
+                                                         val patternType: PatternType,
+                                                         val servletPath: String? = null,
+                                                         val httpMethod: HttpMethod? = null,
+                                                         override val rule: AuthorizationManager<RequestAuthorizationContext>) : AuthorizationManagerRule(rule)
+
     protected abstract class AuthorizationRule(open val rule: String)
+
+    protected abstract class AuthorizationManagerRule(open val rule: AuthorizationManager<RequestAuthorizationContext>)
 
     protected enum class PatternType {
         ANT, MVC
--- a/config/src/main/kotlin/org/springframework/security/config/annotation/web/AuthorizeHttpRequestsDsl.kt
+++ b/config/src/main/kotlin/org/springframework/security/config/annotation/web/AuthorizeHttpRequestsDsl.kt
@@ -1,0 +1,253 @@
+/*
+ * Copyright 2002-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.config.annotation.web
+
+import org.springframework.http.HttpMethod
+import org.springframework.security.authorization.AuthenticatedAuthorizationManager
+import org.springframework.security.authorization.AuthorityAuthorizationManager
+import org.springframework.security.authorization.AuthorizationDecision
+import org.springframework.security.authorization.AuthorizationManager
+import org.springframework.security.config.annotation.web.builders.HttpSecurity
+import org.springframework.security.config.annotation.web.configurers.AuthorizeHttpRequestsConfigurer
+import org.springframework.security.core.Authentication
+import org.springframework.security.web.access.intercept.RequestAuthorizationContext
+import org.springframework.security.web.util.matcher.AnyRequestMatcher
+import org.springframework.security.web.util.matcher.RequestMatcher
+import org.springframework.util.ClassUtils
+import java.util.function.Supplier
+
+/**
+ * A Kotlin DSL to configure [HttpSecurity] request authorization using idiomatic Kotlin code.
+ *
+ * @author Yuriy Savchenko
+ * @since 5.7
+ */
+class AuthorizeHttpRequestsDsl : AbstractRequestMatcherDsl() {
+    private val authorizationRules = mutableListOf<AuthorizationManagerRule>()
+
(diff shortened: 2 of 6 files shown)


config/src/main/kotlin/org/springframework/security/config/annotation/web/AbstractRequestMatcherDsl.kt
```diff
@@ -42,8 +44,19 @@ abstract class AbstractRequestMatcherDsl {
                                                   val httpMethod: HttpMethod? = null,
                                                   override val rule: String) : AuthorizationRule(rule)
 
+    protected data class MatcherAuthorizationManagerRule(val matcher: RequestMatcher,
```
jzheaux, line 47: Please place this immediately after `MatcherAuthorizationRule`, as you did with `AuthorizationManagerRule`

config/src/main/kotlin/org/springframework/security/config/annotation/web/AuthorizeHttpRequestsDsl.kt
```diff
@@ -0,0 +1,253 @@
...
+     * If the current request will not be processed by Spring MVC, a reasonable default
+     * using the pattern as an ant pattern will be used.
+     *
+     * @param pattern the pattern to match incoming requests against.
+     * @param servletPath the servlet path to match incoming requests against. This
+     * only applies when using an MVC pattern matcher.
+     * @param access the [AuthorizationManager] to secure the matching request
+     * (i.e. created via hasAuthority("ROLE_USER"))
+     */
+    fun authorize(pattern: String,
```
jzheaux, line 132: Regarding your concern about method matching, you might need to do the overloading yourself, that is, without default parameters. You can have a method that takes two Strings and then invokes an overloaded method that takes two Strings and an AuthorizationManager.

config/src/main/kotlin/org/springframework/security/config/annotation/web/AuthorizeRequestsDsl.kt
```diff
@@ -54,7 +54,7 @@ class AuthorizeRequestsDsl : AbstractRequestMatcherDsl() {
      * Adds a request authorization rule for an endpoint matching the provided
      * pattern.
      * If Spring MVC is on the classpath, it will use an MVC matcher.
-     * If Spring MVC is not an the classpath, it will use an ant matcher.
+     * If Spring MVC is not on the classpath, it will use an ant matcher.
```
jzheaux, line 57: Thanks for the cleanup. Will you please move these cleanups to a separate commit so that they are easier to backport to earlier versions?

jzheaux (changes requested):
Thanks, @ysavchen! I've left some feedback inline.

I'd also like to have @eleftherias take a look since she is quite a bit more familiar with the Kotlin DSL than I am.

config/src/main/kotlin/org/springframework/security/config/annotation/web/AuthorizeHttpRequestsDsl.kt
ysavchen: Thanks for the review, @jzheaux!
I analyzed this problem a bit more thoroughly.
1. There is some inconsistency between lambda config and Kotlin DSL for `authorizeRequests`
    - In lambda config, if endpoint matching is specified, a user must specify the access
    - Kotlin DSL config allows to omit the access, then "authenticated" will be applied by default
    ```
    // lambda config
    http.authorizeRequests { authorize ->
        authorize.mvcMatchers("/hello").authenticated()  // user must explicitly specify access
    }
    
    // Kotlin DSL
    http {
        authorizeRequets {
            authorize("/hello")  // user can omit the access, then "authenticated" is applied implicitly
        }
    }
    ```
2. Problem in my case (rather migration problem than method overloading)
The case when a problem may appear is a user has `authorizeRequests` config with SpEL in Kotlin DSL
```
authorizeRequests {
    authorize("/user/{userName}", "#userName == 'user'")
}
```
and migrates to new API simply adding Http to the method name
```
authorizeHttpRequests {
    authorize("/user/{userName}", "#userName == 'user'")
}
```
The code compiles successfully and the user may not pay attention that his/her SpEL expression does not configure access anymore.
If the app's API is covered by tests, then the user will get the error and no problem here.
Intellij also shows a hint that the second parameter is changed to servletPath.

If I remove "authenticated" default in this method:
```
fun authorize(pattern: String,
              servletPath: String,
              access: AuthorizationManager<RequestAuthorizationContext>)
```
Then the compiler shows an error, and forces the user pay attention that his/her SpEL config is not valid anymore.

But removing for this particular method adds a special case to the framework API.
So, I think the choice is to leave default "authenticated" config (app tests or Intellij hint solution) or remove "authenticated" by default for all methods in Kotlin DSL for `authorizeHttpRequests` (compilation error solution).

3. I realized that the test case with path variable value can be convered the following config instead of SpEL:
```
@Bean
open fun securityFilterChain(http: HttpSecurity): SecurityFilterChain {
    val access = AuthorizationManager { _: Supplier<Authentication>, context: RequestAuthorizationContext ->
        AuthorizationDecision(context.variables["userName"] == "user")
    }
    http {
        authorizeHttpRequests {
            authorize("/user/{userName}", access)
        }
    }
    return http.build()
}
```

eleftherias: Another option is omitting the `servletPath` from this method entirely. 
That's the approach we use for `authorizeRequests`. 
The `servletPath` only applies to MVC matchers, so it could be confusing in the Kotlin DSL because the pattern could end up being used in an Ant matcher or MVC matcher.

eleftherias (changes requested):
Thanks for the PR @ysavchen! 
I would like to have a test for what happens when we use both `authorizeRequests` and `authorizeHttpRequests` in the Kotlin DSL.
I've also left one commend inline.

jzheaux:
Hi, @ysavchen. Are you able to make the requested changes?

config/src/main/kotlin/org/springframework/security/config/annotation/web/AuthorizeHttpRequestsDsl.kt
ysavchen: I'm not sure if I've got it right, but `authorizeRequests` uses `servletPath`.
If an Ant matcher is used, then `servletPath` is ignored:
```
 when (rule.patternType) {
       PatternType.ANT -> requests.antMatchers(rule.httpMethod, rule.pattern).access(rule.rule)
       PatternType.MVC -> requests.mvcMatchers(rule.httpMethod, rule.pattern)
                .apply { if(rule.servletPath != null) servletPath(rule.servletPath) }
                .access(rule.rule)
 }
 ```
 Removing the `servletPath` from the method in `authorizeHttpRequests` makes it different from `authorizeRequests` and a bit reduces the API.<br>
 I can propose a solution for the migration problem by using an explicit type for `servletPath` parameter (new 4th option).
 Thus, the `authorize` methods of `authorizeHttpRequests` provide the same functionality and defaults as `authorizeRequests`, and a user will get a compilation error during migration.
 The API looks like:
 ```
 authorizeHttpRequests {
     authorize("/path", servletPath("/spring"), hasRole("ADMIN"))
 }
 ```
 Please, correct if I've got smth wrong in the option with omitting the `servletPath` parameter.

config/src/main/kotlin/org/springframework/security/config/annotation/web/AbstractRequestMatcherDsl.kt
ysavchen: Done

ysavchen:
Sorry for delay @jzheaux. I needed some time to analyze the options.
I've added a comment regarding the used solution.
Tests are updated.

config/src/main/kotlin/org/springframework/security/config/annotation/web/AuthorizeHttpRequestsDsl.kt
eleftherias: Thanks @ysavchen. You are correct about the `servletPath`.
You've also made me see another issue with the `authorizeRequests` DSL, for which I created the issue gh-10957.

I think your suggestion to remove "authenticated" by default for all methods in Kotlin DSL for `authorizeHttpRequests` (compilation error solution) is the best approach.

ysavchen: Thanks @eleftherias 
Updated PR

jzheaux (approved):

Issue #10846, opened 2022-02-17
OAuth2AuthorizedClientArgumentResolver couldn't use ReactiveOAuth2AuthorizedClientManager registered in the Context 


muhdalavu:
### Related servlet bug
#8700 

**Describe the bug**
`OAuth2AuthorizedClientArgumentResolver` does not use the `ReactiveOAuth2AuthorizedClientManager` provided by the Spring configuration, therefore `@RegisteredOAuth2AuthorizedClient` annotated parameters do not use potential customization done in the config.

**To Reproduce**
Let's say you want to customize the access token response for a refresh like described in the [doc](Let's say you want to customize the access token response for a refresh like described in the [doc](https://docs.spring.io/spring-security/site/docs/5.3.2.RELEASE/reference/html5/#customizing-the-access-token-response-2))

```
@Bean
    ReactiveOAuth2AuthorizedClientManager authorizedClientManager(
            ServerOAuth2AuthorizedClientRepository serverOAuth2AuthorizedClientRepository,
            ServerOAuth2AuthorizedClientRepository authorizedClientRepository,
            ReactiveOAuth2AuthorizedClientService authorizedClientService,
            ReactiveClientRegistrationRepository clientRegistrationRepository) {
        WebClientReactiveRefreshTokenTokenResponseClient webClientReactiveRefreshTokenTokenResponseClient = new WebClientReactiveRefreshTokenTokenResponseClient();
        webClientReactiveRefreshTokenTokenResponseClient.setWebClient(webClient());
        webClientReactiveRefreshTokenTokenResponseClient.setBodyExtractor(new OAuth2AccessTokenResponseBodyExtractor());
        ReactiveOAuth2AuthorizedClientProvider authorizedClientProvider =
                ReactiveOAuth2AuthorizedClientProviderBuilder.builder()
                        .authorizationCode()
                        .refreshToken(refreshTokenGrantBuilder -> refreshTokenGrantBuilder.accessTokenResponseClient(webClientReactiveRefreshTokenTokenResponseClient))
                        .build();
        DefaultReactiveOAuth2AuthorizedClientManager authorizedClientManager =
                new DefaultReactiveOAuth2AuthorizedClientManager(clientRegistrationRepository, authorizedClientRepository);
        authorizedClientManager.setAuthorizedClientProvider(authorizedClientProvider);
        return authorizedClientManager;
    }
```
  

[...]

PR #10898, opened 2022-02-22
fix gh_10846

muhdalavu:
#10846 

Fix for [OAuth2AuthorizedClientArgumentResolver couldn't use ReactiveOAuth2AuthorizedClientManager registered in the Context](https://github.com/spring-projects/spring-security/issues/10846)

### DIFF
--- a/config/src/test/java/org/springframework/security/config/annotation/web/reactive/ReactiveOAuth2ClientImportSelectorTest.java
+++ b/config/src/test/java/org/springframework/security/config/annotation/web/reactive/ReactiveOAuth2ClientImportSelectorTest.java
@@ -1,0 +1,169 @@
+/*
+ * Copyright 2002-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


config/src/test/java/org/springframework/security/config/annotation/web/reactive/ReactiveOAuth2ClientImportSelectorTest.java
```diff
@@ -0,0 +1,142 @@
...
+	public void setApplicationContext(ApplicationContext context) {
+		// @formatter:off
+		this.client = WebTestClient
+				.bindToApplicationContext(context)
+				.build();
+		// @formatter:on
+	}
+
+	@Test
+	public void requestWhenAuthorizedClientManagerConfiguredThenUsed() {
```
jzheaux, line 71: It would be nice to confirm that the previous behavior still works. Can you please add tests that show that if only repositories are defined, then the argument resolver still works? These tests would go in an earlier commit, but in this same PR.

muhdalavu:
Sure will do it.

jzheaux:
Thanks, @muhdalavu. This is now merged into `5.7.x` as well as `main`.

Issue #10905, opened 2022-02-24
Add Jackson Support for saml2 Module


ugrave:
**Describe the bug**
I try to use Jackson to de/serialize http jdb session.
I create a ConversionService to use Jackson Object mapper with registred modules of `SecurityJackson2Modules`.

**To Reproduce**
Using the following ConversionService:

````java
@Configuration(proxyBeanMethods = false)
public class SessionConfig implements BeanClassLoaderAware {

  private ClassLoader beanClassLoader;
  
  @Bean
  ConversionService springSessionConversionService() {
    ObjectMapper objectMapper = objectMapper();
    var conversionService = new GenericConversionService();
    conversionService.addConverter(new JacksonSerializer(objectMapper));
    conversionService.addConverter(new JacksonDeserializer(objectMapper));
    return conversionService;
  }

  private ObjectMapper objectMapper() {
    ObjectMapper objectMapper = new ObjectMapper();
    objectMapper.registerModules(SecurityJackson2Modules.getModules(this.beanClassLoader));
    return objectMapper;
  }

  @Override
  public void setBeanClassLoader(ClassLoader classLoader) {
    this.beanClassLoader = classLoader;
  }

  private static class JacksonSerializer implements GenericConverter {

    private final ObjectMapper objectMapper;

    public JacksonSerializer(ObjectMapper objectMapper) {
      this.objectMapper = objectMapper;
    }

    @Override
    public Set<ConvertiblePair> getConvertibleTypes() {
      return Set.of(new ConvertiblePair(Object.class, byte[].class));
    }

    @Override
    public Object convert(Object source, TypeDescriptor sourceType, TypeDescriptor targetType) {
      try {
        return objectMapper.writeValueAsBytes(source);
      } catch (JsonProcessingException e) {
        throw new UncheckedIOException(e);
      }
    }
  }

  private static class JacksonDeserializer implements GenericConverter {

    private final ObjectMapper objectMapper;

    public JacksonDeserializer(ObjectMapper objectMapper) {
      this.objectMapper = objectMapper;
    }

    @Override
    public Set<ConvertiblePair> getConvertibleTypes() {
      return Set.of(new ConvertiblePair(byte[].class, Object.class));
    }

    @Override
    public Object convert(Object source, TypeDescriptor sourceType, TypeDescriptor targetType) {
      try {
        return objectMapper.readValue((byte[]) source, targetType.getType());
      } catch (IOException e) {
        throw new UncheckedIOException(e);
      }
[...]

PR #10907, opened 2022-02-24
Add Jackson Support for saml2 Module

ugrave:
This PR does the following:

* Adds a new Jackson Module `org.springframework.security.saml2.jackson2.Saml2Jackson2Module` which contains mixins for SAML2 classes
* Add new mixin for `Collections.modifiableMap` to `org.springframework.security.jackson2.CoreJackson2Module`

Closes gh-10905

### DIFF
--- a/core/src/main/java/org/springframework/security/jackson2/UnmodifiableMapDeserializer.java
+++ b/core/src/main/java/org/springframework/security/jackson2/UnmodifiableMapDeserializer.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2002-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.jackson2;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * Custom deserializer for {@link UnmodifiableMapMixin}.
+ *
+ * @author Ulrich Grave
+ * @since 5.7
+ * @see UnmodifiableMapMixin
+ */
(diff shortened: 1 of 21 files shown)


ugrave:
I have added a `org.springframework.security.jackson2.UnmodifiableMapMixin` to the CoreJackson2Module. Now i see there is already a mixin for UnmodifiableMaps bur this is placed in the OAuth2ClientJackson2Module (`org.springframework.security.oauth2.client.jackson2.UnmodifiableMapMixin`)
Question is how sould i handle it. Move/Copy the mixing from the OAuth2 module to the general jackson module and replace my version.

saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/jackson2/AbstractMixinTests.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.List;
+import java.util.Map;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.BeforeEach;
+
+import org.springframework.security.jackson2.SecurityJackson2Modules;
+import org.springframework.security.saml2.provider.service.authentication.DefaultSaml2AuthenticatedPrincipal;
+
+abstract class AbstractMixinTests {
```
jzheaux, line 31: Instead of relying on inheritance here, I think this could be a static factory class that each of the tests invokes. For example, there is `TestRelyingPartyRegistrations` that constructs different kinds of relying party registration configurations. This could instead be `TestSaml2JsonPayloads`.

core/src/main/java/org/springframework/security/jackson2/UnmodifiableMapDeserializer.java
```diff
@@ -0,0 +1,54 @@
...
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * Custom deserializer for {@link UnmodifiableMapMixin}.
+ *
+ * @author Ulrich Grave
+ * @since 6.0
```
jzheaux, line 34: Will you change these to `5.7`, please? I think we can get this feature out to the community earlier than 6.0, which would be nice for everyone.

jzheaux (changes requested):
Thanks, @ugrave! I appreciate your thoroughness in the PR, I've left some feedback inline.

In preparation for merging, will you please squash your commits?

saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/jackson2/AbstractMixinTests.java
ugrave: Move payloads to factory class. Not sure if i also should move the creation of the ObjectMapper to a factory or still use the before method.

jzheaux: I think the before method is fine, it's probably simple enough to repeat inside each test instead of inheriting.

jzheaux:
Thanks, @ugrave! 

This is merged via df84826c950c35d0b28d5b04a60dabba7d865c40. I also added a polish of 6c3d183a94196f1f6c320d29f70e022f38195177 and 931fb6a328d13ae66699fc4a0bba62540f65611b to remove some duplication and remove the `authenticated` attribute from `Saml2AuthenticationMixin`.

Issue #10903, opened 2022-02-23
ServerHttpBasicAuthenticationConverter uses platform's default charset


frozenice:
**Expected Behavior**

`ServerHttpBasicAuthenticationConverter` should use utf-8 (or a configurable charset) for `String` construction. My understanding is that the majority of HTTP clients, especially browsers, use utf-8 nowadays for basic auth encoding (*citation needed*).

**Current Behavior**

In `org.springframework.security.web.server.ServerHttpBasicAuthenticationConverter` the *Authorization* header is parsed and turned into a `UsernamePasswordAuthenticationToken`. The header content after "`Basic `" is base64-decoded with `Base64.getDecoder().decode(value)` to get the raw bytes. These bytes are then converted to a `String` ([here](https://github.com/spring-projects/spring-security/blob/main/web/src/main/java/org/springframework/security/web/server/ServerHttpBasicAuthenticationConverter.java#L56)):

```java
new String(base64Decode(credentials))
```

This uses a `String` constructor without specifying a charset (which, in my opinion, is always bad). Without an explicit charset it will use the *platform's default charset* - so, for example, utf-8 on Linux and iso-8859-1 on Windows.

**Context**

This is not the first time an issue with basic auth encoding is brought up (see e. g. #2969, but that is for `BasicAuthenticationFilter` to first use iso-8859-1, then utf-8, instead of utf-8 only, funnily), but I haven't found one for `ServerHttpBasicAuthenticationConverter`.

I have a workaround (disabling httpBasic and adding my own `AuthenticationWebFilter`), but that's really ugly and missing some features (the matcher / entryPoint stuff from `org.springframework.security.config.web.server.ServerHttpSecurity.HttpBasicSpec#configure`).

Notes:

- `org.springframework.security.web.authentication.www.BasicAuthenticationFilter` uses utf-8 as a default. It is configurable via `org.springframework.security.web.authentication.www.BasicAuthenticationFilter#setCredentialsCharset` (although I don't know how or even if that is exposed for configuration).
- `org.springframework.security.web.server.ServerHttpBasicAuthenticationConverter` uses the platform's default charset, unconfigurably.
[...]

PR #10911, opened 2022-02-25
Use utf-8 in ServerHttpBasicAuthenticationConverter

frozenice:
This changes `ServerHttpBasicAuthenticationConverter` to use utf-8 rather than the platform's default encoding, as per #10903.

- use an explicit charset (utf-8)
- remove useless argument to substring
- add test with non-ASCII characters
- rename test to use proper character name

### DIFF
--- a/web/src/main/java/org/springframework/security/web/server/ServerHttpBasicAuthenticationConverter.java
+++ b/web/src/main/java/org/springframework/security/web/server/ServerHttpBasicAuthenticationConverter.java
@@ -16,6 +16,7 @@
 
 package org.springframework.security.web.server;
 
+import java.nio.charset.StandardCharsets;
 import java.util.Base64;
 import java.util.function.Function;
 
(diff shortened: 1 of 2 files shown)


sjohnr:
Hi @frozenice. Apologies on the delay reviewing this. I'm working on a webinar that will be done this week and I will circle back next week!

web/src/main/java/org/springframework/security/web/server/ServerHttpBasicAuthenticationConverter.java
```diff
@@ -51,9 +52,8 @@
 		if (!StringUtils.startsWithIgnoreCase(authorization, "basic ")) {
 			return Mono.empty();
 		}
-		String credentials = (authorization.length() <= BASIC.length()) ? ""
-				: authorization.substring(BASIC.length(), authorization.length());
-		String decoded = new String(base64Decode(credentials));
+		String credentials = (authorization.length() <= BASIC.length()) ? "" : authorization.substring(BASIC.length());
+		String decoded = new String(base64Decode(credentials), StandardCharsets.UTF_8);
```
sjohnr, line 56: Based on discussion in the parent issue, I think I'm now leaning towards actually making this configurable with a default. This would have the benefit of matching up well with the servlet implementation. Would you mind adding that option and a few corresponding tests with for example `ISO_8859_1`?

sjohnr (changes requested):
Sorry for the delay @frozenice. Thanks for the discussion and the PR. I have a bit of feedback below.

frozenice:
Fixed the checkstyle errors. Btw the `request` parameter of `org.springframework.security.web.server.authentication.ServerHttpBasicAuthenticationConverterTests#apply` is not used.

sjohnr:
Thanks @frozenice! I've squashed your commits and this is now in `main` as 2b6bc5dd0b3ba14f277062299f1e3692a744344d. I have also added a small polish commit 946e24e1c255604d553fe7a24a2508ca60c0841a to add javadoc, copyright year, and an additional test. I'll add this to the 5.7 release as well.

PR #10912, opened 2022-02-25
Change HashSet to LinkedHashSet for RelyingPartyRegistration credentials

fhanik:
Changing the collections type in the RelyingPartyRegistration.Builder to `LinkedHashSet` from `HashSet`

RelyingPartyRegistration constructor converts the credentials (`credentials`, `decryptionX509Credentials`, `signingX509Credentials`) to unmodifiable, ordered `LinkedList` objects.

There is a chance, possibly minuscule but not non existent, that the Builder collections which uses HashSet/HashMap underneath may return a different order for these credentials. This may lead to very difficult debugging steps when a certain key (the first in the list) is expected to have been used for signing but wasn't.


fhanik:
Would love to have this simple fix back ported to maintenance branches too. Thank you!

jzheaux:
Thanks, @fhanik! This is merged into `5.7.x` via 70b52a001b150e157c846a41fa3e5b9f1e5fe0fd. It will also be ported to `main`, `5.6.x`, and `5.5.x`

Issue #10828, opened 2022-02-11
Saml2AuthenticationRequestRepository does not work in combination with Spring Session


tompson:
### Summary

The `HttpSessionSaml2AuthenticationRequestRepository` saves the `Saml2AuthenticationRequest` in the session and tries to load it after the IdP authenticated the user. 

This does not work when using Spring Session because the session cookie is not sent to the server after the IdP authenticated the user.

Spring Session creates a session cookie with `SameSite=Lax` which causes the browser not to send the cookie when sending the `POST` request after the IdP authentication. 

### To Reproduce

Create a minimal Spring Boot application with Spring Security SAML and Spring Session active.

### Expected behaviour

After login at the IdP and being redirected to the application the user should be signed in and seeing the secured URL.

### Actual behaviour 

A new session is created and the user is at the login page again. When he tries to load the secured URL he is able to request it.

eleftherias:
Thanks for reaching out @tompson.

Note that we have a workaround shared in https://github.com/spring-projects/spring-session/issues/1577.

We will need to think about what changes in Spring Security or Spring Session could make this scenario work, since there is no clear solution at the moment.
Feel free to share any suggestions.

jzheaux:
@tompson, can you clarify how this is an issue with `Saml2AuthenticationRequestRepository`? It sounds like it's an issue with the `RequestCache`. 

`Saml2AuthenticationRequestRepository` stores the `AuthnRequest` for the purpose of validating the `InResponseTo` value in the SAML 2.0 Response, but such validation has not yet been added.

tompson:
@eleftherias thanks for that hint, I experimented with different settings for `SameSite` myself but it is not an option for us to reduce protection against CSRF

@jzheaux I stumbled across this when trying to solve https://github.com/spring-projects/spring-security/issues/10550 and I realised later that it is just an optional validation of `InResponseTo` - so it is no blocker for us

onmishkin:
> Thanks, @tompson. I wonder if you could create an implementation of `RequestCache` that keys the saved requests by the `AuthnRequest#ID` instead of by the session id.

[...]

PR #10961, opened 2022-03-11
RelyingPartyRegistrations#fromMetadataLocation: prioritize REDIRECT binding for sso and logout

Kehrlann:
## Context

Since the introduction of SameSite cookie restrictions, SAML's SSO and Logout with POST break the SavedRequest behavior. See issues:
-  Saml2AuthenticationRequestRepository does not work in combination with Spring Session #10828 
- Spring session does not redirect back to original request in [spring-session](https://github.com/spring-projects/spring-session/issues/1577)

In their service metadata, some services have multiple `SingleSignOnService` entries, because they support both `POST` and `REDIRECT` bindings, see [examples](#examples) . Same is theoretically possible for `SingleLogoutService`, but I have not personally seen examples in the wild.

Currently, the `OpenSamlMetadataAssertingPartyDetailsConverter` takes whichever binding is first. So if `POST` comes first, it is selected, leading to the above "issues".

## Proposed solution

Instead of selection the first `SingleSignOnService` available, iterate over all `SingleSignOnService`s and select the first `REDIRECT`-bound.

Note: I'm proposing this against `5.7.x` but I'm unsure whether I should do this against `main` instead. I feel it is a feature change that should go in the next minor, but it could come before Spring Security 6.

#### Okta

```
<md:SingleSignOnService 
    Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" 
    Location="https://REDACTED.okta.com/app/REDACTED/sso/saml"
/>
<md:SingleSignOnService 
    Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect" 
    Location="https://REDACTED.okta.com/app/REDACTED/sso/saml"
/>
```

#### Workspace One

```
<md:SingleSignOnService 
    Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect" 
    Location="https://REDACTED.workspaceair.com/SAAS/auth/federation/sso"
/>
<md:SingleSignOnService 
    Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" 
    Location="https://REDACTED.workspaceair.com/SAAS/auth/federation/sso"
/>
```

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/OpenSamlMetadataAssertingPartyDetailsConverter.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/OpenSamlMetadataAssertingPartyDetailsConverter.java
@@ -139,7 +139,9 @@ class OpenSamlMetadataAssertingPartyDetailsConverter {
 				continue;
 			}
 			party.singleSignOnServiceLocation(singleSignOnService.getLocation()).singleSignOnServiceBinding(binding);
-			break;
+			if (binding.equals(Saml2MessageBinding.REDIRECT)) {
+				break;
+			}
 		}
 		for (SingleLogoutService singleLogoutService : idpssoDescriptor.getSingleLogoutServices()) {
 			Saml2MessageBinding binding;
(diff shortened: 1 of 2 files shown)


Kehrlann:
This PR has absolutely no effect on the aforementioned bugs. I mixed AuthN *requests* and *responses*,  the binding is for the *request*, the SAML Asserting Party will _always_ POST back, no matter what was used in the request.

Issue #10968, opened 2022-03-14
Allow overriding encodeCookie and decodeCookie


bhuism:
So that we can implement other ways of serialising the request to the cookie value

sjohnr:
@bhuism thanks for the suggestion. For clarity, can you describe your use case for adjusting the behavior of this class?

Also, I think the preference generally would be to provide customization hooks, such as a `Function<String, String> cookieEncoder`, etc. instead of encouraging/allowing inheritance-based overrides. While this class (`CookieRequestCache`) is not `final`, I think enhancements would ideally treat it as such.

Note that it's usually best to discuss an issue prior to opening a PR.

bhuism:
@sjohnr Morning, I will discuss it first next time. Idea in this case that the cookie value can be crypted, so that it can not be changed by a client. 

Customisation hook via function is perfectly fine offcource, why is this preferred btw?

sjohnr:
@bhuism, thanks for the reply! Customization hooks are preferred because they encourage composition and delegation instead of inheritance. You can then build up bigger and more complex implementations by combining one or more existing implementations together with your own logic.

rwinch:
Spring Security allows customization of how the SecurityContext is persisted via SecurityContextRepository implementations, CSRF token via CsrfTokenRepository, etc. There are already hooks for how everything in Spring Security is persisted. If you find something specific that you cannot customize, please let us know and we will work with you on how to do it or adding something that you can do it.

PR #10969, opened 2022-03-14
Allow for overriding encodeCookie and decodeCookie

bhuism:
#10968 Allow for overriding encodeCookie and decodeCookie

### DIFF
--- a/web/src/main/java/org/springframework/security/web/savedrequest/CookieRequestCache.java
+++ b/web/src/main/java/org/springframework/security/web/savedrequest/CookieRequestCache.java
@@ -113,11 +113,11 @@ public class CookieRequestCache implements RequestCache {
 		response.addCookie(removeSavedRequestCookie);
 	}
 
-	private static String encodeCookie(String cookieValue) {
+	protected String encodeCookie(String cookieValue) {
 		return Base64.getEncoder().encodeToString(cookieValue.getBytes());
 	}
 
-	private static String decodeCookie(String encodedCookieValue) {
+	protected String decodeCookie(String encodedCookieValue) {
 		return new String(Base64.getDecoder().decode(encodedCookieValue.getBytes()));
 	}
 


web/src/main/java/org/springframework/security/web/savedrequest/CookieRequestCache.java
```diff
@@ -113,11 +113,11 @@ public void removeRequest(HttpServletRequest request, HttpServletResponse respon
 		response.addCookie(removeSavedRequestCookie);
 	}
 
-	private static String encodeCookie(String cookieValue) {
+	protected String encodeCookie(String cookieValue) {
```
jzheaux, line 116: I think it would be more powerful to introduce `Converter`s that can turn a `Cookie` into a `String` and vice-versa. Something like:

```java
public void setCookieValueDecoder(Converter<HttpServletRequest, String> decoder);

public void setValueCookieEncoder(Converter<String, Cookie> encoder);
```

In addition to allowing you to customize how the value is encoded and decoded, this approach makes more features available as well.

I specifically like the encoder since this allows an application to completely construct the cookie, allowing for things like setting the same site parameter or customizing the max age. This contract would also allow the application to customize the cookie name.

On the flip side, the decoder could look up the cookie and extract the URI themselves. This contract allows the application to customize the cookie name.

web/src/main/java/org/springframework/security/web/savedrequest/CookieRequestCache.java
```diff
@@ -113,11 +113,11 @@ public void removeRequest(HttpServletRequest request, HttpServletResponse respon
...
 		return Base64.getEncoder().encodeToString(cookieValue.getBytes());
 	}
 
-	private static String decodeCookie(String encodedCookieValue) {
+	protected String decodeCookie(String encodedCookieValue) {
 		return new String(Base64.getDecoder().decode(encodedCookieValue.getBytes()));
 	}
```
jzheaux, line 123: Please change the copyright header at the top of the file to read `2022` instead of `2020`.

jzheaux (changes requested):
Thanks, @bhuism! In addition to my inline comments, will you please update your commit message to follow [the contributing guidelines](https://github.com/spring-projects/spring-security/blob/main/CONTRIBUTING.adoc#format-commit-messages)?

Issue #10956, opened 2022-03-10
Multiple .requestMatchers().mvcMatchers() override previous one


vova-yatsyk-theraven:
**Describe the bug**
Defining multiple `.requestMatchers().mvcMatchers()` are overriding previous one.
```
http
        .requestMatchers()
            .mvcMatchers("/api-1")
            .mvcMatchers("/api-2")
            .mvcMatchers("/api-3")
        .and()
```
In the example above matcher for `"/api-3"` will override the one for `"/api-1"`, and result matcher list will contain only two latest matchers: `"/api-2"` and `"/api-3"`.

**Expected behavior**
All matches should be used together, joined by `OrRequestMatcher`.

**Possible issue**
`MvcMatchersRequestMatcherConfigurer` that returned after `.mvcMatchers()` contains only the last pattern, but it should collect all pattern combined together.
I think the line https://github.com/spring-projects/spring-security/blob/main/config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java#L3119 from the following code:

```
@Override
public MvcMatchersRequestMatcherConfigurer mvcMatchers(HttpMethod method, String... mvcPatterns) {
    List<MvcRequestMatcher> mvcMatchers = createMvcMatchers(method, mvcPatterns);
    setMatchers(mvcMatchers);
    return new MvcMatchersRequestMatcherConfigurer(getContext(), mvcMatchers);
}
```
should be changed to return all matches: `this.matchers`, like below:
```
@Override
public MvcMatchersRequestMatcherConfigurer mvcMatchers(HttpMethod method, String... mvcPatterns) {
    List<MvcRequestMatcher> mvcMatchers = createMvcMatchers(method, mvcPatterns);
    setMatchers(mvcMatchers);
    return new MvcMatchersRequestMatcherConfigurer(getContext(), this.matchers);
}
```

**Version**
Reproduced on v5.3.4.
But main and the latest [v5.6.2](https://github.com/spring-projects/spring-security/blob/5.6.2/config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java) contains the same code.

marcusdacoregio:
Thanks for the report @vova-yatsyk-theraven, I was able to reproduce the bug here. This is now scheduled for release `5.7.0-M3` and will be backported to `5.5.x` and `5.6.x`.

marcusdacoregio:
As a workaround, you can specify it like this:
```java
http
[...]

PR #11019, opened 2022-03-22
Fix mvcMatchers overriding previous paths

marcusdacoregio:
Closes gh-10956

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
@@ -3290,20 +3290,26 @@ public final class HttpSecurity extends AbstractConfiguredSecurityBuilder<Defaul
 	 */
 	public final class MvcMatchersRequestMatcherConfigurer extends RequestMatcherConfigurer {
 
+		private final List<MvcRequestMatcher> mvcMatchers;
+
 		/**
 		 * Creates a new instance
 		 * @param context the {@link ApplicationContext} to use
-		 * @param matchers the {@link MvcRequestMatcher} instances to set the servlet path
-		 * on if {@link #servletPath(String)} is set.
+		 * @param mvcMatchers the {@link MvcRequestMatcher} instances to set the servlet
+		 * path on if {@link #servletPath(String)} is set.
+		 * @param allMatchers the {@link RequestMatcher} instances to continue the
+		 * configuration
 		 */
-		private MvcMatchersRequestMatcherConfigurer(ApplicationContext context, List<MvcRequestMatcher> matchers) {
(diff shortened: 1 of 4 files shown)


config/src/test/java/org/springframework/security/config/annotation/web/configurers/HttpSecurityRequestMatchersTests.java
```diff
@@ -175,6 +211,70 @@ public void loadConfig(Class<?>... configs) {
...
 
+	@EnableWebSecurity
+	@Configuration
+	@EnableWebMvc
+	static class MultiMvcMatcherInLambdaConfig {
+
+		@Bean
+		@Order(Ordered.HIGHEST_PRECEDENCE)
+		SecurityFilterChain first(HttpSecurity http) throws Exception {
+			http.requestMatchers(
```
jzheaux, line 222: Please use `//@ formatter:off` and `:on` so that the `http` formatting is preserved like so:

```java
http
    .requestMatchers((requests) -> requests
        .mvcMatchers("/test-1")
        .mvcMatchers("/test-2")
        .mvcMatchers("/test-3")
    )
    .authorizeRequests((authorize) -> authorize
        .anyRequest().denyAll()
    )
    .httpBasic(Customizer.withDefaults());
```

config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
```diff
@@ -3290,20 +3290,26 @@ public HttpSecurity regexMatcher(String pattern) {
 	 */
 	public final class MvcMatchersRequestMatcherConfigurer extends RequestMatcherConfigurer {
 
+		private final List<MvcRequestMatcher> mvcMatchers;
```
jzheaux, line 3293: I imagine that this problem exists in other parts of Spring Security Config, like `ChannelRequestMatcherRegistry`/`MvcMatchersRequiresChannelUrl` and `StandardInterceptUrlRegistry`/`MvcMatchersAuthorizedUrl`.

If it doesn't, then I wonder if those indicate a pattern that `RequestMatcherConfigurer`/`MvcMatchersRequestMatcherConfigurer` should follow. Or if it does, let's fix those classes at the same time.

jzheaux (changes requested):

marcusdacoregio:
Thanks for the review @jzheaux.

This problem does not exist on others `Configurers` because folks are required to call `requiresInsecure()/requiresSecure()` or `hasRole()` after the `mvcMatchers()`, therefore you can't chain the call to `mvcMatchers()`. However, I added some tests to make sure that the behavior is preserved.

jzheaux (approved):

Issue #10673, opened 2021-12-31
Claims contain an instance of java.net.URL and are used in hash-based containers


jyrimatti:
Hi,

class `org.springframework.security.oauth2.core.user.OAuth2UserAuthority` includes its `attributes` for the calculation of `hashCode` and `equals`.
In case of `org.springframework.security.oauth2.core.oidc.user.OidcUserAuthority extends OAuth2UserAuthority` those attributes come from `org.springframework.security.oauth2.core.oidc.OidcIdToken` and `org.springframework.security.oauth2.core.oidc.OidcUserInfo` via `org.springframework.security.oauth2.core.oidc.user.OidcUserAuthority.collectClaims(OidcIdToken, OidcUserInfo)` function. It seems that due to https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcIdTokenDecoderFactory.java#L101 whenever (always?) the claims contain a `org.springframework.security.oauth2.core.oidc.IdTokenClaimNames.ISS`, it is there as an instance of class `java.net.URL`.

This is a problem, since due to historical reasons, URL performs DNS lookups whenever its equals/hashCode is used. Instances of URL must not be used in any containers requiring use of equals/hashCode.

A simple solution would be to change it to be an instance of `java.net.URI`. Would this break something? I didn't craft a pull request since I have no idea if this kind of change would be a major breaking change for spring-security...?

jgrandja:
Thanks for the report @jyrimatti. This will be a breaking change so I've scheduled it for `6.0.x`.

I think the required change is to ensure the `Map` returned from `OidcUserAuthority.collectClaims()` replaces all `URL` values with `URL.toExternalForm()`.

Would you be interested in submitting a PR for this?

jyrimatti:
Yes, sure, I can submit one.

Just to make sure: Are you certain this fix is enough? java.net.URL is not deprecated and I guess it's never going to be, but as https://bugs.java.com/bugdatabase/view_bug.do?bug_id=4434494 says:

> People are encouraged to use URI for parsing and URI comparison, and leave URL class for accessing the URI itself, getting at the protocol handler, interacting with the protocol etc.

[...]

PR #11030, opened 2022-03-25
Convert URLs to String before equals/hashcode

jyrimatti:
java.net.URL performs DNS lookups whenever its equals/hashCode is
used. Thus attribute values of type java.net.URL need to be converted
to something else before they are used for equals/hashCode.

This PR performs the conversion in equals/hashCode. Thus the
actual attributes remain the same as originally given in the
constructor. This brings some overhead from constructing a new
map in every equals/hashCode invocation, but since there likely
aren't that many attributes, I think performance doesn't need to
be considered here.

Closes gh-10673

### DIFF
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/user/OAuth2UserAuthority.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/user/OAuth2UserAuthority.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,9 +16,11 @@
 
 package org.springframework.security.oauth2.core.user;
 
+import java.net.URL;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Objects;
 
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.SpringSecurityCoreVersion;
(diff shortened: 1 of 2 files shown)


oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/user/OAuth2UserAuthority.java
```diff
@@ -85,13 +86,13 @@ public boolean equals(Object obj) {
 		if (!this.getAuthority().equals(that.getAuthority())) {
 			return false;
 		}
-		return this.getAttributes().equals(that.getAttributes());
+		return convertURLs(this.getAttributes()).equals(convertURLs(that.getAttributes()));
```
jgrandja, line 89: We should try to optimize the `equals()` and `hashCode()` since it creates a new `Map` on each call. One option would be to introduce a member variable `private final Map<String, Object> convertedAttributes` that is created at construction time and used in `equals()` and `hashCode()` or an internal `Comparator`  instance. I think I prefer the latter.

jgrandja (changes requested):
Apologies for the delay @jyrimatti. Please see review comment.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/user/OAuth2UserAuthority.java
jyrimatti: Hi,

sure, if you think it's worth the complexity. Sorry for the delay, I was in my COVIDs this week :)

I first tried the simple approach of just adding a `private final Map<String, Object> convertedAttributes` but got failing tests from
```
OAuth2AuthenticationTokenMixinTests.serializeWhenMixinRegisteredThenSerializes()
OAuth2AuthenticationTokenMixinTests.serializeWhenRequiredAttributesOnlyThenSerializes()
```
I guess it's forbidden for some reason to add new members to this class?

I'm afraid I didn't understand the suggestion to use a `Comparator`. It wouldn't help with hashCode and there's no need for natural ordering, so wouldn't a `Comparator` just bring additional complexity?

Here's the updated PR. I had to duplicate equals/hashCode functionality of a map. I feel this is somewhat complicated, please let me know if I missed some obvious simple implementation.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/user/OAuth2UserAuthority.java
```diff
@@ -100,4 +126,12 @@ public String toString() {
 		return this.getAuthority();
 	}
 
+	/**
+	 * @return {@code URL} converted to a string since {@code URL} shouldn't be used for
+	 * equality/hashCode. For other instances the value is returned as is.
+	 */
+	static final Object convertURL(Object value) {
```
jgrandja, line 133: Let's make this `private`

oauth2/oauth2-core/src/test/java/org/springframework/security/oauth2/core/user/OAuth2UserAuthorityTests.java
```diff
@@ -58,4 +60,11 @@ public void constructorWhenAllParametersProvidedAndValidThenCreated() {
 		assertThat(userAuthority.getAttributes()).isEqualTo(ATTRIBUTES);
 	}
 
+	@Test
+	public void convertURLReplacesURLWithString() throws MalformedURLException {
```
jgrandja, line 64: Please remove this test and add a test for `equals()` and `hashCode()` using 2 different instances of `OAuth2UserAuthority`

jgrandja (changes requested):
Thanks for the updates @jyrimatti ! Almost ready to merge after a couple of minor updates.

Also, can you please squash the commits on next push. Thanks!

PR #11034, opened 2022-03-26
Fix saml2 authentication-requests documentation

grafjo:
looks like `ping` is some registration id used in the past.


jzheaux:
Thanks, @grafjo, this is now merged into `main`, `5.7.x`, and backported to `5.6.x`.

grafjo:
@jzheaux you're welcome - thx for the fast merge!

Issue #11053, opened 2022-04-05
AllowSessionCreation not propagated from authentication filters to handlers


t-beckmann:
**Expected Behavior**

When disallowing sesion creation no attempts on creating a session shall be done. Configuring flag at filters should affect handlers as well.

**Current Behavior**

When disabling the allowSessionCreation flag of OAuth2/SAML using the setter in AbstractAuthenticationProcessingFilter, the value is not applied to the authentication handlers attached to it. As a result, e.g. the default created SimpleUrlAuthenticationFailureHandler will attempt session creation.

**Context**

In my use case the server is configured to stateless session creation policy in order to avoid sticky sessions at the load balancer as well as session synchronization between cluster nodes. I know this mode of operation requires a lot of additional changes in order to operate properly, which I succeeded in doing already.

As a workaround I currently propagate the allowSessionCreation manually and since AbstractAuthenticationProcessingFilter.getFailureHandler() is not public an ugly hack is required for not losing defaults:

```
// following configures allowSessionCreation on filters which Spring unfortunately does not do itself
final var AbstractAuthenticationProcessingFilter_getFailureHandler = ReflectionUtils.findMethod(AbstractAuthenticationProcessingFilter.class, "getFailureHandler");
ReflectionUtils.makeAccessible(AbstractAuthenticationProcessingFilter_getFailureHandler);
final var authProcessingFilterPostProcessor = new ObjectPostProcessor<AbstractAuthenticationProcessingFilter>() {
	@Override
	public <O extends AbstractAuthenticationProcessingFilter> O postProcess(final O filter) {
		filter.setAllowSessionCreation(false);

		final var failureHandler = ReflectionUtils.invokeMethod(AbstractAuthenticationProcessingFilter_getFailureHandler, filter);
		if (failureHandler instanceof SimpleUrlAuthenticationFailureHandler) {
			((SimpleUrlAuthenticationFailureHandler) failureHandler).setAllowSessionCreation(false);
		}
		return filter;
	}
};
```

**Using an ObjectPostProcessor probably is acceptable. However in order doing so I propose changing `AbstractAuthenticationProcessingFilter.getFailureHandler()` to public access.**

t-beckmann:
Notice, the related `AuthenticationFilter.getFailureHandler()` is public already.

eleftherias:
Thanks for reaching out @t-beckmann.
This looks like a duplicate of gh-4242.
There are also some alternative workarounds mentioned in that issue.

[...]

PR #11054, opened 2022-04-05
Make getters of handlers publically accessible

t-beckmann:
Changes AbstractAuthenticationProcessingFilter getters of success and failure handlers to public visibility so that allowSessionCreation flag can be configured, see also gh-11053.


t-beckmann:

PR #11062, opened 2022-04-05
Use SecurityContextHolderStrategy for defaults

jzheaux:
This introduces support for a `SecurityContextHolderStrategy` bean when used with the minimal configuration:

```java
http
    .authorizeHttpRequests((authorize) -> authorize.anyRequest().authenticated())
    .formLogin(Customizer.withDefaults())
    .httpBasic(Customizer.withDefaults())
```

as well as

```java
http
    .authorizeRequests((authorize) -> authorize.anyRequest().authenticated())
    .formLogin(Customizer.withDefaults())
    .httpBasic(Customizer.withDefaults())
```

including the subsequent lookup of the `SecurityContext` using `@AuthenticationPrincipal`.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configuration/WebMvcSecurityConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configuration/WebMvcSecurityConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,6 +24,8 @@ import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.expression.BeanFactoryResolver;
 import org.springframework.expression.BeanResolver;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.core.context.SecurityContextHolderStrategy;
 import org.springframework.security.web.method.annotation.AuthenticationPrincipalArgumentResolver;
 import org.springframework.security.web.method.annotation.CsrfTokenArgumentResolver;
 import org.springframework.security.web.method.annotation.CurrentSecurityContextArgumentResolver;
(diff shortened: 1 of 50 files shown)


web/src/main/java/org/springframework/security/web/access/ExceptionTranslationFilter.java
```diff
@@ -239,6 +243,17 @@ public void setMessageSource(MessageSource messageSource) {
 		this.messages = new MessageSourceAccessor(messageSource);
 	}
 
+	/**
+	 * Sets the {@link SecurityContextHolderStrategy} to use. The default action is to use
+	 * the {@link SecurityContextHolderStrategy} stored in {@link SecurityContextHolder}.
+	 *
+	 * @since 6.0
```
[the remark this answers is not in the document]
jzheaux: Good catch. Originally, this was targeted for 6.0. Then there was an attempt made to get it in for 5.7, which failed. Now that there is a 5.8 release, I've updated the `@since` tags to `5.8`

config/src/main/java/org/springframework/security/config/annotation/web/configuration/WebMvcSecurityConfiguration.java
```diff
@@ -72,6 +78,9 @@ RequestDataValueProcessor requestDataValueProcessor() {
 	@Override
 	public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
 		this.beanResolver = new BeanFactoryResolver(applicationContext.getAutowireCapableBeanFactory());
+		if (applicationContext.getBeanNamesForType(SecurityContextHolderStrategy.class).length > 0) {
+			this.securityContextHolderStrategy = applicationContext.getBean(SecurityContextHolderStrategy.class);
```
rwinch, line 82: Will this fail with multiple beans of SecurityContextHolderStrategy?

rwinch (changes requested):

config/src/main/java/org/springframework/security/config/annotation/web/configurers/SecurityContextConfigurer.java
```diff
@@ -83,6 +84,19 @@ public SecurityContextConfigurer<H> securityContextRepository(SecurityContextRep
...
+	/**
+	 * Specifies the shared {@link SecurityContextHolderStrategy} that is to be used
+	 * @param securityContextHolderStrategy the {@link SecurityContextHolderStrategy} to
+	 * use
+	 * @return the {@link SecurityContextConfigurer} for further customizations
+	 * @since 5.8
+	 */
+	public SecurityContextConfigurer<H> securityContextHolderStrategy(
+			SecurityContextHolderStrategy securityContextHolderStrategy) {
+		getBuilder().setSharedObject(SecurityContextHolderStrategy.class, securityContextHolderStrategy);
```
rwinch, line 96: I have a concern that the `HttpSecurity` and `AuthenticationPrincipalArgumentResolver` may have different `SecurityContextHoderStrategy` associated with them.

rwinch (changes requested):

config/src/main/java/org/springframework/security/config/annotation/web/configurers/SecurityContextConfigurer.java
jzheaux: We'll leave the DSL support out for now. It would be rather unexpected to use a different `SecurityContextHolderStrategy` for each filter chain. That said, if this turns out to be needed, the argument resolver could possibly be updated to look at the request attribute to get the security context.

config/src/main/java/org/springframework/security/config/annotation/web/configuration/WebMvcSecurityConfiguration.java
jzheaux: Yes, this will fail and that was my intent. However, Rob and I discussed and the prevailing pattern in Spring Security is to return null if there are multiple of the same bean, allowing the application to fall back to the defaults. This may be worth discussing in the future.

PR #11071, opened 2022-04-07
Fix Infinite recursion WebSecurityConfigurerAdapter.java

Milan-Toliya:
Fixed Infinite recursion issue during authenticate and loadUserByUsername method execution.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configuration/WebSecurityConfigurerAdapter.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configuration/WebSecurityConfigurerAdapter.java
@@ -512,9 +512,7 @@ public abstract class WebSecurityConfigurerAdapter implements WebSecurityConfigu
 		@Override
 		public Authentication authenticate(Authentication authentication) throws AuthenticationException {
 			if (this.delegate != null) {
-				if (this == this.delegate) {
-                    throw new IllegalStateException("Infinite recursion occured while authenticate.");
-                }
+				Assert.state(this != this.delegate, "Infinite recursion detected. Please refer to this comment for common resolutions: https://github.com/spring-projects/spring-security/issues/8369#issuecomment-614862388");
 				return this.delegate.authenticate(authentication);
 			}
 			synchronized (this.delegateMonitor) {
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configuration/WebSecurityConfigurerAdapter.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configuration/WebSecurityConfigurerAdapter.java
@@ -460,6 +460,9 @@ public abstract class WebSecurityConfigurerAdapter implements WebSecurityConfigu
 		@Override
 		public UserDetails loadUserByUsername(String username) throws UsernameNotFoundException {
 			if (this.delegate != null) {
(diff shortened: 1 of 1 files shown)


marcusdacoregio:
Hi @Milan-Toliya, thanks for contributing to Spring Security. Have you opened an issue before working on the code changes?

Can you help me understand what this PR fixes and what is the use case?

Milan-Toliya:
Hi @marcusdacoregio I did not open an issue before submitting this patch.

With respect to Use case, I was trying to implement security filter in the web app. I Implemented WebSecurityConfigurerAdapter and used InMemoryUserDetailsManager to store user details.

I did Implemented WebSecurityConfigurerAdapter twice in the same project. It was causing issue in boot up, I fixed that by adding `@Order(0)` in one class. Now It was time to test `/authenticate` API. During testing of authentication end point it was going in infinite recursion which is not acceptable in any use case.

I have eliminated Infinite recursion by **throwing an Exception** when `this == this.delegate`.

I may have missed important details to share here as I'm new to the spring community. Please Let me know if more details required.

Milan-Toliya:
Hi @marcusdacoregio 

I tried to reproduce but no luck. I will share code once I am able to reproduce it.

Steps to reproduce: 

1. Implement JWT Token with longest expire time. Then did some changes in both implementation of `WebSecurityConfigurerAdapter`. (NOTE: I have two Implementation of `WebSecurityConfigurerAdapter`)
2. Those changes made in code lead to state such that, `UserDetailsService`(Interface) is auto wired with `UserDetailsServiceDelegator` (Inner class of the WebSecurityConfigurerAdapter)
3. Used JWT token generated in step 1 to GET data from the secure END (i.e. `/helloword`)
4. JWT Filter will perform this action `userDetailsService.loadUserByUsername(username)` (**ISSUE** Infinite recursion loop in the method `userDetailsService.loadUserByUsername` when `userDetailsService` is instance of `UserDetailsServiceDelegator`)
5. I am not able to simulate 2nd step right now  `UserDetailsService`(Interface) is auto wiring with `InMemoryUserDetailsManager` which it actually should in positive case. Not sure how I managed to got in 2nd step's state earlier.

Refer Below Image for classes mention above.
image

NOTE: I did not got issue with `AuthenticationManagerDelegator.authenticate()` but By looking at the code it clearly visible that it suffers from the same infinite recursion loop when `this == this.delegate` (refer the code change I have proposed or look at the source code )


Thanks.

Milan-Toliya:
> Hi @Milan-Toliya, thanks for contributing to Spring Security. Have you opened an issue before working on the code changes?
> 
> Can you help me understand what this PR fixes and what is the use case?

Hi, @marcusdacoregio I have opened an issue for this PR.
https://github.com/spring-projects/spring-security/issues/11088

marcusdacoregio (changes requested):
Hi @Milan-Toliya, while we cannot simulate the root problem that is happening, this was brought to the team's attention and we think that your changes make sense since it's "cheaper" to throw an `IllegalStateException` than waiting for the `StackOverflowError` to happen.

If users have configured things in an unsupported way, throwing an `IllegalStateException` feels like an improvement from an error handling perspective.

marcusdacoregio:
Closing in favor of https://github.com/spring-projects/spring-security/issues/12343

Issue #11065, opened 2022-04-06
Custom RelayState with the new OpenSamlAuthenticationRequestResolver


exeba:
With the deprecated `Saml2AuthenticationRequestContext` it was possible to set a custom RelayState, but with the new `OpenSamlAuthenticationRequestResolver` it is always initialized to:

```
String relayState = UUID.randomUUID().toString();
```

I don't see a way to overcome this without implementing my own AuthenticationRequestResolver, is there a reason for this?

jzheaux:
Hi, @exeba. The purpose of deprecating the other classes was to simplify the API to a single interface. So, to have a custom relay state, you create a custom resolver that delegates like so:

```java
@Bean 
Saml2AuthenticationRequestResolver authenticationRequestResolver(
        RelyingPartyRegistrationResolver registrations) {
    OpenSaml4AuthenticationRequestResolver authenticationRequests = 
            new OpenSaml4AuthenticationRequestResolver(registrations);
    return (request) -> {
        Saml2PostAuthenticationRequest post = authenticationRequests.resolve(request);
        RelyingPartyRegistration registration = registrations.resolve(request, "registration-id");
        String relayState = resolveCustomRelayState(request);
        return Saml2PostAuthenticationRequest.withRelyingPartyRegistration(registration)
                .samlRequest(post.getSamlRequest())
                .authenticationRequestUri(post.getAuthenticationRequestUri())
                .relayState(relayState)
                .build();
    }
}
```

That said, I think this can be simplified. For example, it would be nicer to not have to waste the CPU on generating a relay state if the app is going to override it. To that end, I think we should add a `setRelayStateResolver(Converter<HttpServletRequest, String> relayStateResolver)` method. The default value would be one that randomly generates the relay state.

The resulting configuration would then instead be:

```java
@Bean 
Saml2AuthenticationRequestResolver authenticationRequestResolver(
        RelyingPartyRegistrationResolver registrations) {
    OpenSaml4AuthenticationRequestResolver authenticationRequests = 
            new OpenSaml4AuthenticationRequestResolver(registrations);
    authenticationRequests.setRelayStateResolver(this::resolveCustomRelayState);
    return authenticationRequests;
}
```

Would this address your concern and, if so, are you able to submit a PR that adds the setter?

cjburgawork:
@jzheaux, Thanks for posting the code for the relay state resolvers.

[...]

PR #11072, opened 2022-04-07
Allow custom relay state

exeba:
This addresses issue #11065.

I've a couple of questions:

 - [OpenSamlAuthenticationRequestResolver.java](https://github.com/spring-projects/spring-security/compare/main...exeba:custom-relay-state?expand=1#diff-156c59b85e9f8743eb095ea6dc8b4603a9ad1999c0f7f4c9f75359aec2af9805) used to have all final fields. Is it ok to add a non-final field?
 - I tried to add a test case but it seems that ./gradelw test just compiles without executing the tests, am I missing something?

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/authentication/OpenSamlAuthenticationRequestResolver.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/authentication/OpenSamlAuthenticationRequestResolver.java
@@ -129,7 +129,7 @@ class OpenSamlAuthenticationRequestResolver {
 		if (authnRequest.getID() == null) {
 			authnRequest.setID("ARQ" + UUID.randomUUID().toString().substring(1));
 		}
-		String relayState = relayStateResolver.convert(request);
+		String relayState = this.relayStateResolver.convert(request);
 		Saml2MessageBinding binding = registration.getAssertingPartyDetails().getSingleSignOnServiceBinding();
 		if (binding == Saml2MessageBinding.POST) {
 			if (registration.getAssertingPartyDetails().getWantAuthnRequestsSigned()) {
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/authentication/OpenSamlAuthenticationRequestResolver.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/authentication/OpenSamlAuthenticationRequestResolver.java
@@ -36,6 +36,7 @@ import org.opensaml.saml.saml2.core.impl.IssuerBuilder;
 import org.opensaml.saml.saml2.core.impl.NameIDBuilder;
 import org.w3c.dom.Element;
 
+import org.springframework.core.convert.converter.Converter;
 import org.springframework.security.saml2.Saml2Exception;
 import org.springframework.security.saml2.core.OpenSamlInitializationService;
 import org.springframework.security.saml2.core.Saml2ParameterNames;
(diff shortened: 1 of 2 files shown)


saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/authentication/OpenSamlAuthenticationRequestResolver.java
```diff
@@ -71,6 +72,8 @@
 
 	private final NameIDBuilder nameIdBuilder;
 
+	private Converter<HttpServletRequest, String> relayStateResolver = (request) -> UUID.randomUUID().toString();
```
jzheaux, line 75: It's fine that this is non-final since it has a default configuration value.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/authentication/OpenSamlAuthenticationRequestResolver.java
```diff
@@ -122,7 +129,7 @@
 		if (authnRequest.getID() == null) {
 			authnRequest.setID("ARQ" + UUID.randomUUID().toString().substring(1));
 		}
-		String relayState = UUID.randomUUID().toString();
+		String relayState = relayStateResolver.convert(request);
```
jzheaux, line 132: Please change to `this.relayStateResolver...` to fit with Spring Security's coding conventions.

saml2/saml2-service-provider/src/opensaml4Main/java/org/springframework/security/saml2/provider/service/web/authentication/OpenSaml4AuthenticationRequestResolver.java
```diff
@@ -77,6 +78,14 @@ public void setClock(Clock clock) {
 		this.clock = clock;
 	}
 
+	/**
+	 * Use this {@link Converter<HttpServletRequest, String>} to compute the RelayState
+	 * @param relayStateResolver the {@link Converter<HttpServletRequest, String>} to use
+	 */
```
jzheaux, line 84: Please add `@since 5.7`

saml2/saml2-service-provider/src/opensaml4Main/java/org/springframework/security/saml2/provider/service/web/authentication/OpenSaml4AuthenticationRequestResolver.java
```diff
@@ -77,6 +78,14 @@ public void setClock(Clock clock) {
...
+	public void setRelayStateResolver(Converter<HttpServletRequest, String> relayStateResolver) {
+		this.authnRequestResolver.setRelayStateResolver(relayStateResolver);
```
jzheaux, line 86: While unlikely, please check for `null` here by doing:

```java
Assert.notNull(relayStateResolver, "relayStateResolver cannot be null");
```

jzheaux (changes requested):
Thanks, @exeba, for the quick turnaround!

For tests, please try running `./gradlew format check` and let me know if the tests run at that point.

exeba:
@jzheaux I tried runnig `./gradlew format check`.
It seems that the majority of tests are run, but not the ones under `saml2/saml2-service-provider/src/test/`
The only way I can get a failure is with a syntax error. Maybe it is a problem with my environment

jzheaux:
Thanks, @exeba. Sorry for the delay, this is now merged in 4dfc349914cf7df9c4405d7f34b9f3a570b77cc1 and will go into the 5.8 release.

Issue #9667, opened 2021-04-21
SecurityExpressionHandler#createEvaluationContext should defer lookup of Authentication


jzheaux:
Like `AuthorizationManager`, it would be nice to be able to defer the looking up `Authentication` in `SecurityExpressionHandler`.

A default `createEvaluationContext` method could be added:

```java
default EvaluationContext createEvaluationContext(Supplier<Authentication> authentication, T invocation) {
    return createEvaluationContext(authentication.get(), invocation);
}
```

And then `DefaultXXXSecurityExpressionHandler`s would be updated to defer evaluation until the authentication was inspected.

evgeniycheban:
@jzheaux I can take this.

evgeniycheban:
@jzheaux I've opened the PR, please take a look.

jzheaux:
In order to accommodate changing to `Supplier<Authentication>` in `SecurityExpressionRoot`, the `protected Authentication authentication` member variable is changed to `private Supplier<Authentication> authentication`.

Classes that extend `SecurityExpressionRoot` and access the `protected` member variable should instead call `getAuthentication()`.

PR #11079, opened 2022-04-07
Add AuthorizationManager to Messaging

jzheaux:
Instead of extending `AbstractSecurityWebSocketMessageBrokerConfigurer`, folks can do:

```java
@EnableWebSocketSecurity
@Configuration
public class SecurityConfig {
    @Bean  
    AuthorizationManager<Message<?>> authorizationManager(
            MessageMatcherDelegatingAuthorizationManager.Builder messages) {
        messages
            .simpDestMatchers("/path/**").permitAll()
            .anyMessage().denyAll();
        return messages.build();
    }
}
```

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/socket/AbstractSecurityWebSocketMessageBrokerConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/socket/AbstractSecurityWebSocketMessageBrokerConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -81,9 +81,12 @@ import org.springframework.web.socket.sockjs.transport.TransportHandlingSockJsSe
  *
  * @author Rob Winch
  * @since 4.0
+ * @see WebSocketMessageBrokerSecurityConfiguration
+ * @deprecated Use {@link EnableWebSocketSecurity} instead
  */
 @Order(Ordered.HIGHEST_PRECEDENCE + 100)
 @Import(ObjectPostProcessorConfiguration.class)
+@Deprecated
 public abstract class AbstractSecurityWebSocketMessageBrokerConfigurer extends AbstractWebSocketMessageBrokerConfigurer
 		implements SmartInitializingSingleton {
 
--- a/config/src/main/java/org/springframework/security/config/annotation/web/socket/WebSocketMessageBrokerSecurityConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/socket/WebSocketMessageBrokerSecurityConfiguration.java
@@ -1,0 +1,145 @@
+/*
+ * Copyright 2002-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.config.annotation.web.socket;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+import org.springframework.beans.factory.SmartInitializingSingleton;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.annotation.Import;
+import org.springframework.core.Ordered;
+import org.springframework.core.annotation.Order;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.handler.invocation.HandlerMethodArgumentResolver;
+import org.springframework.messaging.simp.config.ChannelRegistration;
+import org.springframework.messaging.support.ChannelInterceptor;
+import org.springframework.security.authorization.AuthorizationManager;
+import org.springframework.security.authorization.SpringAuthorizationEventPublisher;
+import org.springframework.security.messaging.access.intercept.AuthorizationChannelInterceptor;
+import org.springframework.security.messaging.access.intercept.MessageMatcherDelegatingAuthorizationManager;
+import org.springframework.security.messaging.context.AuthenticationPrincipalArgumentResolver;
+import org.springframework.security.messaging.context.SecurityContextChannelInterceptor;
+import org.springframework.security.messaging.web.csrf.CsrfChannelInterceptor;
+import org.springframework.security.messaging.web.socket.server.CsrfTokenHandshakeInterceptor;
+import org.springframework.util.Assert;
+import org.springframework.web.servlet.handler.SimpleUrlHandlerMapping;
+import org.springframework.web.socket.config.annotation.WebSocketMessageBrokerConfigurer;
+import org.springframework.web.socket.server.HandshakeInterceptor;
+import org.springframework.web.socket.server.support.WebSocketHttpRequestHandler;
+import org.springframework.web.socket.sockjs.SockJsService;
+import org.springframework.web.socket.sockjs.support.SockJsHttpRequestHandler;
+import org.springframework.web.socket.sockjs.transport.TransportHandlingSockJsService;
+
+@Order(Ordered.HIGHEST_PRECEDENCE + 100)
+@Import(MessageMatcherAuthorizationManagerConfiguration.class)
+final class WebSocketMessageBrokerSecurityConfiguration
+		implements WebSocketMessageBrokerConfigurer, SmartInitializingSingleton {
+
+	private static final String SIMPLE_URL_HANDLER_MAPPING_BEAN_NAME = "stompWebSocketHandlerMapping";
+
+	private MessageMatcherDelegatingAuthorizationManager b;
+
+	private static final AuthorizationManager<Message<?>> ANY_MESSAGE_AUTHENTICATED = MessageMatcherDelegatingAuthorizationManager
+			.builder().anyMessage().authenticated().build();
+
+	private final ChannelInterceptor securityContextChannelInterceptor = new SecurityContextChannelInterceptor();
+
+	private final ChannelInterceptor csrfChannelInterceptor = new CsrfChannelInterceptor();
+
+	private AuthorizationChannelInterceptor authorizationChannelInterceptor = new AuthorizationChannelInterceptor(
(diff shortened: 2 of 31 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/configuration/EnableWebSecurity.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
```
rwinch, line 2: I think this can be reverted

config/src/main/java/org/springframework/security/config/annotation/web/messaging/AuthorizationManagerMessageMatcherRegistry.java
```diff
@@ -0,0 +1,354 @@
...
+	 */
+	public Constraint matchers(MessageMatcher<?>... matchers) {
+		List<MessageMatcher<?>> builders = new ArrayList<>(matchers.length);
+		for (MessageMatcher<?> matcher : matchers) {
+			builders.add(matcher);
+		}
+		return new Constraint(builders);
+	}
+
+	public AuthorizationManager<Message<?>> build() {
```
rwinch, line 176: I think this is more of a builder than a registry since it exposes a build method. It makes me wonder if we are building an AuthorizationManager if we out to just provide a builder on MessageMatcherDelegatingAuthorizationManager as you mentioned. We'd need to double check for package tangles though.

config/src/main/java/org/springframework/security/config/annotation/web/socket/AbstractSecurityWebSocketMessageBrokerConfigurer.java
```diff
@@ -81,9 +81,15 @@
  *
  * @author Rob Winch
  * @since 4.0
+ * @see WebSocketMessageBrokerSecurityConfiguration
+ * @deprecated Use
+ * {@link org.springframework.security.config.annotation.web.configuration.EnableWebSecurity}
```
rwinch, line 86: Should this be EnableWebSocketSecurity?

config/src/main/java/org/springframework/security/config/annotation/web/messaging/AuthorizationManagerMessageMatcherRegistry.java
```diff
@@ -0,0 +1,354 @@
...
+public final class AuthorizationManagerMessageMatcherRegistry {
+
+	private final MessageMatcherDelegatingAuthorizationManager.Builder builder = MessageMatcherDelegatingAuthorizationManager
+			.builder();
+
+	private final ApplicationContext context;
+
+	private PathMatcher pathMatcher;
+
+	public AuthorizationManagerMessageMatcherRegistry(ApplicationContext context) {
```
rwinch, line 56: Does this need to be public or can it be package scope?

config/src/main/java/org/springframework/security/config/annotation/web/messaging/AuthorizationManagerMessageMatcherRegistry.java
```diff
@@ -0,0 +1,354 @@
...
+ *
+ * @author Josh Cummings
+ * @since 5.7
+ */
+public final class AuthorizationManagerMessageMatcherRegistry {
+
+	private final MessageMatcherDelegatingAuthorizationManager.Builder builder = MessageMatcherDelegatingAuthorizationManager
+			.builder();
+
+	private final ApplicationContext context;
```
rwinch, line 52: Does it make more sense to always inject PathMatcher rather than use `ApplicationContext`. If we need to delay the lookup perhaps a Supplier<PathMatcher>`.

config/src/main/java/org/springframework/security/config/annotation/web/socket/WebSocketMessageBrokerSecurityConfiguration.java
```diff
@@ -0,0 +1,171 @@
...
+ *
+ * <pre>
+ * &#064;Configuration
+ * &#064;EnableWebSecurity
+ * public class WebSocketSecurityConfig {
+ *
+ * 	&#064;Bean
+ * 	AuthorizationManager&lt;Message&lt;?&gt;&gt; (AuthorizationManagerMessageMatcherRegistry messages) {
+ * 		messages.simpDestMatchers(&quot;/user/queue/errors&quot;).permitAll()
+ * 				.simpDestMatchers(&quot;/admin/**&quot;).hasRole(&quot;ADMIN&quot;).anyMessage()
```
rwinch, line 67: I wonder if this could be better formatted with anyMessage on a new line

config/src/main/java/org/springframework/security/config/annotation/web/socket/WebSocketMessageBrokerSecurityConfiguration.java
```diff
@@ -0,0 +1,171 @@
...
+/**
+ * Allows configuring WebSocket Authorization.
+ *
+ * <p>
+ * For example:
+ * </p>
+ *
+ * <pre>
+ * &#064;Configuration
+ * &#064;EnableWebSecurity
```
rwinch, line 61: Does it need `@EnableWebSocketSecurity`

config/src/main/java/org/springframework/security/config/annotation/web/socket/WebSocketMessageBrokerSecurityConfiguration.java
```diff
@@ -0,0 +1,171 @@
...
+import org.springframework.web.socket.sockjs.transport.TransportHandlingSockJsService;
+
+/**
+ * Allows configuring WebSocket Authorization.
+ *
+ * <p>
+ * For example:
+ * </p>
+ *
+ * <pre>
```
rwinch, line 59: It might be better to have this documented on `@EnableWebSocketSecurity` since it is public while this class is not.

messaging/src/main/java/org/springframework/security/messaging/access/intercept/AuthorizationChannelInterceptor.java
```diff
@@ -0,0 +1,95 @@
...
+	public AuthorizationChannelInterceptor(AuthorizationManager<Message<?>> preSendAuthorizationManager) {
+		Assert.notNull(preSendAuthorizationManager, "preSendAuthorizationManager cannot be null");
+		this.preSendAuthorizationManager = preSendAuthorizationManager;
+	}
+
+	@Override
+	public Message<?> preSend(Message<?> message, MessageChannel channel) {
+		this.logger.debug(LogMessage.of(() -> "Authorizing message send"));
+		AuthorizationDecision decision = this.preSendAuthorizationManager.check(AUTHENTICATION_SUPPLIER, message);
+		this.eventPublisher.publishAuthorizationEvent(AUTHENTICATION_SUPPLIER, message, decision);
```
rwinch, line 75: I think that eventPublisher can be null. Perhaps it would be good to default eventPublisher to a no-op implementation.

messaging/src/main/java/org/springframework/security/messaging/util/matcher/MessageMatcherEntry.java
```diff
@@ -0,0 +1,44 @@
...
+
+package org.springframework.security.messaging.util.matcher;
+
+/**
+ * A rich object for associating a {@link MessageMatcher} to another object.
+ *
+ * @author Josh Cummings
+ * @since 5.7
+ */
+public class MessageMatcherEntry<T> {
```
rwinch, line 25: What are the benefits of this being a top level class vs inner class of MessageMatcherDelegatingAuthorizationManager

rwinch (changes requested):
Thanks for the pull request. I have provided feedback inline. Please also review for package tangles.

config/src/main/java/org/springframework/security/config/annotation/web/messaging/AuthorizationManagerMessageMatcherRegistry.java
jzheaux: The reason for deferring is to solve a dependency cycle. Calling for the `PathMatcher` instance from the `ApplicationContext` causes the `DelegatingWebSocketMessageBrokerConfiguration` to be constructed before this configurer can be registered to it.

I'll change it to a `Supplier` and we can go from there.

jzheaux:
@rwinch I added XML configuration and responded to your other points.

Note that regarding XML configuration, I originally coded it with an opt-out feature to revert back to using `SecurityMetadataSource`. However, I couldn't find any reason to actually use the feature in an existing application, so I removed it. If you see a reason to add the flag now, this can be added easily as I have the code preserved in an earlier commit in the PR.

jzheaux:
Update: Because of #9667, I think that it's important to still allow applications to opt-in to using `AuthorizationManager`. The reason is that #9667 cannot call the `Authentication`-based `protected` methods in `AbstractSecurityExpressionHandler`. If applications have overridden those, then their custom code would stop being called when they upgrade.

So, the PR now introduces `use-authorization-manager`. When the value is `true` or when the application uses the `authorization-manager-ref` attribute, Spring Security will use the `AuthorizationManager` API. In Spring Security 6.0, this attribute can be deprecated and defaulted to `true`.

core/src/main/java/org/springframework/security/access/expression/SecurityExpressionHandler.java
```diff
@@ -42,7 +43,10 @@
 	 * Provides an evaluation context in which to evaluate security expressions for the
 	 * invocation type.
 	 */
-	EvaluationContext createEvaluationContext(Authentication authentication, T invocation);
+	default EvaluationContext createEvaluationContext(Authentication authentication, T invocation) {
```
rwinch, line 46: Is there a reason to default this method in the same commit?

rwinch (changes requested):
Thanks. I left a comment inline.

core/src/main/java/org/springframework/security/access/expression/SecurityExpressionHandler.java
jzheaux: Shoot, good catch! It's left over from an experiment. I'll revert.

messaging/src/main/java/org/springframework/security/messaging/util/matcher/MessageMatcherEntry.java
jzheaux: Inlined.

config/src/main/java/org/springframework/security/config/annotation/web/socket/WebSocketMessageBrokerSecurityConfiguration.java
jzheaux: Removed JavaDoc since package-private class.

config/src/main/java/org/springframework/security/config/annotation/web/messaging/AuthorizationManagerMessageMatcherRegistry.java
jzheaux: Removed class.

rwinch (approved):

Issue #11073, opened 2022-04-07
Missing reactive DelegatingRequestMatcherHeaderWriter


Dav1dde:
**Expected Behavior**

Non-Reactive Spring Security has a `DelegatingRequestMatcherHeaderWriter`

**Current Behavior**

There is no implementation of `ServerHttpHeadersWriter` which allows the user to write certain headers for only certain paths.

**Context**

I am trying to not have one endpoint set `X-Frame-Options`, which you can do in non-reactive Spring Security using a `DelegatingRequestMatcherHeaderWriter`.

marcusdacoregio:
Hi @Dav1dde.

Are you interested in submitting a PR that adds the new implementation?

Dav1dde:
Yeah for sure, I can look into it. I pretty much already have an implementation in my project which just needs some cleaning up. I assume the reactive implementation would use a `ServerWebExchangeMatcher`? Does it make sense to have a 2nd version or an overloaded constructor that accepts a `PathPattern` directly?

marcusdacoregio:
You can pretty much follow the implementation of `DelegatingRequestMatcherHeaderWriter`. I think it makes more sense to use `ServerWebExchangeMatcher` and let folks decide which implementation they want.

Dav1dde:
@marcusdacoregio After many tries to configure the code style in IntelliJ correctly and failing I finally fixed the import order by hand and submitted the PR ;)

I tried to keep it as close as possible to the servlet version, implementation and test

marcusdacoregio:
What were the problems that you faced when working with the codebase?

Dav1dde:
I just couldnt get the auto formatting or syntax checking get to work. Another issue was that the `gradlew` from the cli always errored (with some weird errors in docstrings with non ascii characters), it worked from IntelliJ though.

I tried to get it to work by importing the eclipse settings, setting up checkstyle, neither worked. Also the gradle format just didnt format the imports properly.

marcusdacoregio:
Unfortunately, there's a limitation with the formatting tool that does not organize the imports properly. Thanks for your feedback, I'll try to make it easier for folks to set up their environment by following the CONTRIBUTING doc.

I have created https://github.com/spring-projects/spring-security/issues/11157 to track the improvement in the CONTRIBUTING doc.

PR #11106, opened 2022-04-14
Add DelegatingServerHttpHeadersWriter

Dav1dde:
Servlet Spring Security has DelegatingRequestMatcherHeaderWriter
the reactive world of Spring Security was missing a class to
conditionally write headers.

Closes gh-11073

### DIFF
--- a/web/src/main/java/org/springframework/security/web/server/header/ServerWebExchangeDelegatingServerHttpHeadersWriter.java
+++ b/web/src/main/java/org/springframework/security/web/server/header/ServerWebExchangeDelegatingServerHttpHeadersWriter.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2002-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.web.server.header;
+
+import reactor.core.publisher.Mono;
+
+import org.springframework.security.web.server.util.matcher.ServerWebExchangeMatcher;
+import org.springframework.security.web.server.util.matcher.ServerWebExchangeMatcherEntry;
+import org.springframework.util.Assert;
+import org.springframework.web.server.ServerWebExchange;
+
+/**
+ * Delegates to a provided {@link ServerHttpHeadersWriter} if
+ * {@link ServerWebExchangeMatcher#matches(ServerWebExchange)} returns a match.
+ *
+ * @author David Herberth
+ * @since 5.8
+ */
+public final class ServerWebExchangeDelegatingServerHttpHeadersWriter implements ServerHttpHeadersWriter {
+
+	private final ServerWebExchangeMatcherEntry<ServerHttpHeadersWriter> headersWriter;
+
+	/**
+	 * Creates a new instance
+	 * @param headersWriter the {@link ServerWebExchangeMatcherEntry} holding a
+	 * {@link ServerWebExchangeMatcher} and the {@link ServerHttpHeadersWriter} to invoke
+	 * if the matcher returns a match.
+	 */
+	public ServerWebExchangeDelegatingServerHttpHeadersWriter(
+			ServerWebExchangeMatcherEntry<ServerHttpHeadersWriter> headersWriter) {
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/springframework/security/web/server/header/DelegatingServerHttpHeadersWriter.java
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.web.server.ServerWebExchange;
+
+/**
+ * Delegates to a provided {@link ServerHttpHeadersWriter} when
+ * {@link ServerWebExchangeMatcher#matches(ServerWebExchange)} returns a match.
+ *
+ * @author David Herberth
+ * @since 5.7
+ */
+public final class DelegatingServerHttpHeadersWriter implements ServerHttpHeadersWriter {
```
marcusdacoregio, line 32: I think the name of the class should be `ServerWebExchangeDelegatingServerHttpHeadersWriter` to align with those already existing, like `ServerWebExchangeDelegatingServerAccessDeniedHandler` and `ServerWebExchangeDelegatingReactiveAuthenticationManagerResolver`.

web/src/main/java/org/springframework/security/web/server/header/DelegatingServerHttpHeadersWriter.java
```diff
@@ -0,0 +1,65 @@
...
+
+	private final ServerWebExchangeMatcher webExchangeMatcher;
+
+	private final ServerHttpHeadersWriter delegateHeadersWriter;
```
marcusdacoregio, line 36: What do you think about using `ServerWebExchangeMatcherEntry<ServerHttpHeadersWriter>` class here? The constructor could benefit from it as well, receiving just one parameter.
```suggestion
	private final ServerWebExchangeMatcherEntry<ServerHttpHeadersWriter> delegate;
```

web/src/main/java/org/springframework/security/web/server/header/DelegatingServerHttpHeadersWriter.java
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.security.web.server.util.matcher.ServerWebExchangeMatcher;
+import org.springframework.util.Assert;
+import org.springframework.web.server.ServerWebExchange;
+
+/**
+ * Delegates to a provided {@link ServerHttpHeadersWriter} when
+ * {@link ServerWebExchangeMatcher#matches(ServerWebExchange)} returns a match.
+ *
+ * @author David Herberth
+ * @since 5.7
```
marcusdacoregio, line 30: Since 5.7 is already on RC1, can you change it to `@since 5.8`?

marcusdacoregio (changes requested):
Hey @Dav1dde, thank you very much for the time and effort you put into this.

Please, take a look at my feedback inline.

web/src/main/java/org/springframework/security/web/server/header/DelegatingServerHttpHeadersWriter.java
Dav1dde: Thanks I didn't know this class existed, I think having it as an overload for the constructor is good/fine, but I would like to keep the current constructor, simply because it makes the calling code more readable:

```java
new ServerWebExchangeDelegatingServerHttpHeadersWriter(
    ServerWebExchangeMatchers.pathMatchers("/**"), 
    StaticServerHttpHeadersWriter.builder().headers("foo", "bar").build()
);

// or this:
new ServerWebExchangeDelegatingServerHttpHeadersWriter(new ServerWebExchangeMatcherEntry(
    ServerWebExchangeMatchers.pathMatchers("/**"), 
    StaticServerHttpHeadersWriter.builder().headers("foo", "bar").build()
));
```

web/src/main/java/org/springframework/security/web/server/header/DelegatingServerHttpHeadersWriter.java
Dav1dde: I am happy to rename this, but I tried following the other `ServerHttpHeadersWriter` implementations in the same [package](https://github.com/spring-projects/spring-security/tree/main/web/src/main/java/org/springframework/security/web/server/header), where none start with `ServerWebExchange*`, e.g. the `CompositeServerHttpHeadersWriter` which has a similar usecase to the delegating implementation.

marcusdacoregio: I see. But simply reading `DelegatingServerHttpHeadersWriter` doesn't tell me exactly what this implementation is doing, actually, it tells me that it delegates to another implementation, there is no clue that a `ServerWebExchangeMatcher` will be used.

Dav1dde: That makes sense!

marcusdacoregio:
Thanks so much, @Dav1dde.

Merged via https://github.com/spring-projects/spring-security/commit/0e2fc51bad2a4f246e5e0c4e88d0df10ae18734b

Issue #11153, opened 2022-04-19
Add shouldFilterAllDispatcherTypes to Kotlin DSL


marcusdacoregio:
Related:
 - #11092

PR #11168, opened 2022-04-25
Add shouldFilterAllDispatcherTypes to Kotlin DSL

marcusdacoregio:
Closes gh-11153


eleftherias (commented):
Since this is a boolean, it can be a variable instead of a function. 
It's a similar situation to `clearAuthentication` in the `LogoutDsl`:
https://github.com/spring-projects/spring-security/blob/9a57b42786accb4d806288f589c2750ca43de563/config/src/main/kotlin/org/springframework/security/config/annotation/web/LogoutDsl.kt#L46-L48

eleftherias (approved):

Issue #11105, opened 2022-04-14
Add AuthorizationManager that uses ExpressionHandler


jzheaux:
In order to better facilitate the use of `AuthorizationManager` in JSP, Thymeleaf, and XML configuration support, we should add an expression-based `AuthorizationManager`.

This can also help folks who may be migrating from `authorizeRequests` which supports expressions and also expression-based voters.

evgeniycheban:
@jzheaux I'd like to take this.

evgeniycheban:
@jzheaux I've opened the PR, please take a look when you have time.

PR #11171, opened 2022-04-27
Add AuthorizationManager that uses ExpressionHandler

evgeniycheban:
Closes gh-11105

### DIFF
--- a/web/src/main/java/org/springframework/security/web/access/expression/DefaultHttpSecurityExpressionHandler.java
+++ b/web/src/main/java/org/springframework/security/web/access/expression/DefaultHttpSecurityExpressionHandler.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2002-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.web.access.expression;
+
+import org.springframework.security.access.expression.AbstractSecurityExpressionHandler;
+import org.springframework.security.access.expression.SecurityExpressionHandler;
+import org.springframework.security.access.expression.SecurityExpressionOperations;
+import org.springframework.security.authentication.AuthenticationTrustResolver;
+import org.springframework.security.authentication.AuthenticationTrustResolverImpl;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.web.access.intercept.RequestAuthorizationContext;
+import org.springframework.util.Assert;
+
+/**
+ * A {@link SecurityExpressionHandler} that uses a {@link RequestAuthorizationContext} to
+ * create a {@link WebSecurityExpressionRoot}.
+ *
+ * @author Evgeniy Cheban
+ * @since 5.8
+ */
+public class DefaultHttpSecurityExpressionHandler extends AbstractSecurityExpressionHandler<RequestAuthorizationContext>
+		implements SecurityExpressionHandler<RequestAuthorizationContext> {
+
+	private AuthenticationTrustResolver trustResolver = new AuthenticationTrustResolverImpl();
+
+	private String defaultRolePrefix = "ROLE_";
+
+	@Override
+	protected SecurityExpressionOperations createSecurityExpressionRoot(Authentication authentication,
+			RequestAuthorizationContext context) {
+		WebSecurityExpressionRoot root = new WebSecurityExpressionRoot(authentication, context.getRequest());
+		root.setRoleHierarchy(getRoleHierarchy());
+		root.setPermissionEvaluator(getPermissionEvaluator());
+		root.setTrustResolver(this.trustResolver);
+		root.setDefaultRolePrefix(this.defaultRolePrefix);
+		return root;
+	}
+
+	/**
+	 * Sets the {@link AuthenticationTrustResolver} to be used. The default is
+	 * {@link AuthenticationTrustResolverImpl}.
+	 * @param trustResolver the {@link AuthenticationTrustResolver} to use
+	 */
+	public void setTrustResolver(AuthenticationTrustResolver trustResolver) {
+		Assert.notNull(trustResolver, "trustResolver cannot be null");
+		this.trustResolver = trustResolver;
+	}
+
+	/**
+	 * Sets the default prefix to be added to
+	 * {@link org.springframework.security.access.expression.SecurityExpressionRoot#hasAnyRole(String...)}
+	 * or
+	 * {@link org.springframework.security.access.expression.SecurityExpressionRoot#hasRole(String)}.
+	 * For example, if hasRole("ADMIN") or hasRole("ROLE_ADMIN") is passed in, then the
+	 * role ROLE_ADMIN will be used when the defaultRolePrefix is "ROLE_" (default).
+	 * @param defaultRolePrefix the default prefix to add to roles. The default is
+	 * "ROLE_".
+	 */
+	public void setDefaultRolePrefix(String defaultRolePrefix) {
+		Assert.hasText(defaultRolePrefix, "defaultRolePrefix cannot be empty");
+		this.defaultRolePrefix = defaultRolePrefix;
+	}
+
+}
--- a/web/src/main/java/org/springframework/security/web/access/expression/WebExpressionAuthorizationManager.java
+++ b/web/src/main/java/org/springframework/security/web/access/expression/WebExpressionAuthorizationManager.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2002-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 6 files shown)


web/src/main/java/org/springframework/security/web/access/expression/WebExpressionAuthorizationManager.java
```diff
@@ -0,0 +1,98 @@
...
+ * @since 5.7
+ */
+public final class WebExpressionAuthorizationManager
+		implements AuthorizationManager<RequestAuthorizationContext>, InitializingBean {
+
+	private SecurityExpressionHandler<RequestAuthorizationContext> expressionHandler = new DefaultHttpSecurityExpressionHandler();
+
+	private Expression expression;
+
+	private final String expressionString;
```
rwinch, line 48: I don't think this needs to be tracked separately because you can just set the Expression directly and use Expression.getExpressionString() when you need to update the Expression when a new SecurityExpressionHandler is set.

This would also allow afterPropertiesSet to be removed

web/src/main/java/org/springframework/security/web/access/expression/WebExpressionAuthorizationManager.java
```diff
@@ -0,0 +1,98 @@
...
+
+	/**
+	 * Determines the access by evaluating the provided expression.
+	 * @param authentication the {@link Supplier} of the {@link Authentication} to check
+	 * @param context the {@link RequestAuthorizationContext} to check
+	 * @return an {@link AuthorizationDecision} based on the evaluated expression
+	 */
+	@Override
+	public AuthorizationDecision check(Supplier<Authentication> authentication, RequestAuthorizationContext context) {
+		EvaluationContext ctx = this.expressionHandler.createEvaluationContext(authentication.get(), context);
```
rwinch, line 77: I wonder if there is something we can do here to continue a delayed lookup of the `Authentication` rather than eagerly obtaining it. Ideally an expression of `permitAll` would not cause the `Authentication` to be accessed so that it does not require the `HttpSession` to be accessed unnecessarily which can be costly for requests to static resources when the `HttpSession` is persisted in an external store like `Redis`.

I think perhaps in another PR we should be made to make this possible. First it would add another constructor to `WebExpressionRoot` that takes a `Supplier<Authentication>`. Next a default method to `SecurityExpressionHandler` that is the same as  createEvaluationContext but accepts a `Supplier<Authentication`. Each implementation would override the default implementation to pass the `Supplier<Authentication>` into the new ExpressionRoot. It would deprecate the non `Supplier<Authentication>` method and constructor.

web/src/main/java/org/springframework/security/web/access/expression/WebSecurityExpressionRoot.java
```diff
@@ -39,6 +40,17 @@ public WebSecurityExpressionRoot(Authentication a, FilterInvocation fi) {
 		this.request = fi.getRequest();
```
rwinch, line 40: This constructor can use the newly added constructor `this(authentication, fi.getRequest())`

rwinch (changes requested):
Thanks for the PR @evgeniycheban I've commented inline

web/src/main/java/org/springframework/security/web/access/expression/WebExpressionAuthorizationManager.java
evgeniycheban: Just wondering if it's okay to parse an expression twice when a custom `SecurityExpressionHandler` is set.
I personally prefer using `afterPropertiesSet` here.

web/src/main/java/org/springframework/security/web/access/expression/WebExpressionAuthorizationManager.java
evgeniycheban: I agree. Also we can use the same approach for Pre/Post `AuthorizationManager`.
I'd like to work on this in another PR.

web/src/main/java/org/springframework/security/web/access/expression/WebExpressionAuthorizationManager.java
evgeniycheban: Done.

web/src/main/java/org/springframework/security/web/access/expression/WebSecurityExpressionRoot.java
evgeniycheban: Done.

web/src/main/java/org/springframework/security/web/access/expression/WebExpressionAuthorizationManager.java
jzheaux: I believe there is a ticket for this: https://github.com/spring-projects/spring-security/issues/9667

rwinch (approved):

rwinch:
Thank you for the changes. I've approved it from my side and will leave it to @jzheaux from here

web/src/main/java/org/springframework/security/web/access/expression/DefaultHttpSecurityExpressionHandler.java
```diff
@@ -0,0 +1,77 @@
...
+	 * Sets the default prefix to be added to
+	 * {@link org.springframework.security.access.expression.SecurityExpressionRoot#hasAnyRole(String...)}
+	 * or
+	 * {@link org.springframework.security.access.expression.SecurityExpressionRoot#hasRole(String)}.
+	 * For example, if hasRole("ADMIN") or hasRole("ROLE_ADMIN") is passed in, then the
+	 * role ROLE_ADMIN will be used when the defaultRolePrefix is "ROLE_" (default).
+	 * @param defaultRolePrefix the default prefix to add to roles. Default "ROLE_"
+	 */
+	public void setDefaultRolePrefix(String defaultRolePrefix) {
+		Assert.hasText("defaultRolePrefix", "defaultRolePrefix cannot be empty");
```
jzheaux, line 73: Please change this to

```suggestion
		Assert.hasText(defaultRolePrefix, "defaultRolePrefix cannot be empty");
```

web/src/main/java/org/springframework/security/web/access/expression/WebSecurityExpressionRoot.java
```diff
@@ -35,8 +36,18 @@ public class WebSecurityExpressionRoot extends SecurityExpressionRoot {
...
-		super(a);
-		this.request = fi.getRequest();
+		this(a, fi.getRequest());
+	}
+
+	/**
+	 * Creates an instance for the given {@link Authentication} and
+	 * {@link HttpServletRequest}.
+	 * @param authentication the {@link Authentication} to use
+	 * @param request the {@link HttpServletRequest} to use
```
jzheaux, line 46: Please add `@since 5.8`

web/src/main/java/org/springframework/security/web/access/expression/WebExpressionAuthorizationManager.java
```diff
@@ -0,0 +1,83 @@
...
+		boolean granted = ExpressionUtils.evaluateAsBoolean(this.expression, ctx);
+		return new AuthorizationDecision(granted);
```
jzheaux, line 75: We should add `ExpressionAuthorizationDecision`, similar to other implementations like `AuthorityAuthorizationDecision` that include the context a manager used to make its decision. 

This will serve a similar role to ``ConfigAttribute``s in Spring Security's event handling.

jzheaux (changes requested):

jzheaux:
Also, @evgeniycheban, if you would please rebase against `5.8.x`, that would be very helpful.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/AuthorizeHttpRequestsConfigurer.java
```diff
@@ -316,6 +317,17 @@ public AuthorizationManagerRequestMatcherRegistry authenticated() {
...
 		}
 
+		/**
+		 * Allows specifying a SpEL expression which will be evaluated for request
+		 * authorization.
+		 * @param expressionString the raw expression string
+		 * @return the {@link AuthorizationManagerRequestMatcherRegistry} for further
+		 * customizations
+		 */
+		public AuthorizationManagerRequestMatcherRegistry access(String expressionString) {
```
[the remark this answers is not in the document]
evgeniycheban: Done.

web/src/main/java/org/springframework/security/web/access/expression/DefaultHttpSecurityExpressionHandler.java
```diff
@@ -0,0 +1,77 @@
...
+import org.springframework.security.core.Authentication;
+import org.springframework.security.web.access.intercept.RequestAuthorizationContext;
+import org.springframework.util.Assert;
+
+/**
+ * A {@link SecurityExpressionHandler} that uses a {@link RequestAuthorizationContext} to
+ * create a {@link WebSecurityExpressionRoot}.
+ *
+ * @author Evgeniy Cheban
+ * @since 5.7
```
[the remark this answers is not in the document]
evgeniycheban: Done.

web/src/main/java/org/springframework/security/web/access/expression/DefaultHttpSecurityExpressionHandler.java
evgeniycheban: Done.

web/src/main/java/org/springframework/security/web/access/expression/WebSecurityExpressionRoot.java
evgeniycheban: Done.

web/src/main/java/org/springframework/security/web/access/expression/WebExpressionAuthorizationManager.java
evgeniycheban: Added `org.springframework.security.web.access.expression.ExpressionAuthorizationDecision` that accepts an `Expression` as additional parameter.

evgeniycheban:
@jzheaux Thanks for the review. I've made changes according to your comments and rebased this branch on top of `5.8.x`.

jzheaux (approved):

jzheaux:
Thanks again, @evgeniycheban! This is now merged into `5.8.x`. I also added a polish to introduce support for request variables. It also adds a few more tests: ffaf5b4e618e2af6e391c6ff0971ff48cdcac5ca

PR #11172, opened 2022-04-27
messages.properties cleanup

Emkas:
I propose a few changes in messages*.properties which will make the files more 'unite' and consistent.

Here is the list:

* Add missing insufficientAuthentication property in messages_*.properties
* Add missing untranslated properties in messages_lt
* Fixed bad property name in messages_it
* Remove unnecessary dots from messages_cs_CZ
* Remove trailing space from messages_ru
* Add missing badLdapConnection property in messages_*.properties
* Remove empty messages_nl.properties file


eleftherias (changes requested):
Thanks for the PR @Emkas!

I don't think there is value in adding the placeholders for untranslated properties, such as `#ExceptionTranslationFilter.insufficientAuthentication=Full authentication is required to access this resource`.
I'm also hesitant to remove the `message_nl.properties` file in case it causes unexpected behaviour in an application that is expecting it.

The remaining changes look good! 
If you could update the PR with the above recommendations I will go ahead and merge it.

Emkas:
@eleftherias, thank you for your feedback!
You're the boss, buuuuuut:

1. I can't imagine how removing empty properties file could break anything. Some one must read this not as a `ResourceBundle` but as file. Even then, the only thing someone can do with it is a check for existence.
2. Placeholders are in most of properties right now. For me this is good, because I can translate everything without specialty tool and without jumping through other files (of course if I don't need any context). As for now there are lacks in both messages.properties and messages_en.properties so I had to check both if missing parts are used in code. I was looking for some consistency about this, but I couldn't find any so my guess was that this kind of placeholders are being used in Spring Security.

I mostly did as you asked, but there not much left, as I removed 3 biggest changes :-) Still in two changes I left some placeholders (`ExceptionTranslationFilter.insufficientAuthentication`). Let me know what you think.

eleftherias:
Thanks for the quick response @Emkas!

I read through your justification for the 2 additional changes and I agree with you.
Feel free to add those commits back.
Thanks for bringing up these points and for your contribution!

Emkas:
I'm very pleased to hear that! 2 commits are back in place. As for now, everything is rebased.

Finally only removed part is removing of empty `_nl.properties`.

eleftherias:
Thanks @Emkas! This is now merged into main.

Emkas:
Great :)

Issue #11175, opened 2022-04-28
@EnableMethodSecurity does not resolve @PreAuthorize on interfaces


noelbundick-msft:
**Describe the bug**
When using `@EnableGlobalMethodSecurity(prePostEnabled = true)` alongside Spring Data REST, it is possible to add `@PreAuthorize("hasRole('SOMETHING')")` on the repository interface, which [secures the entire repository](https://docs.spring.io/spring-data/rest/docs/current/reference/html/#security). You can also use `@PreAuthorize` on individual methods

When using the newer `@EnableMethodSecurity` attribute, `@PreAuthorize` only works on individual methods, and does not work on an interface. This is a change in behavior that could result in potential accidental data leakage when upgrading to the latest Spring Security bits.

**To Reproduce**

* Wire up Spring Data REST in a project with `@EnableGlobalMethodSecurity(prePostEnabled = true)`
* Add `@PreAuthorize("hasRole('BOGUS')")` to your repository interface
* Make a GET request, observe that it is rejected
* Upgrade to `@EnableMethodSecurity`
* Make a request, observe that a response is returned instead of rejected. You are now leaking data to unauthorized callers

**Expected behavior**
`@PreAuthorize` to be processed the same way as before

**Sample**

https://github.com/noelbundick-msft/spring-security-methodsecurity-bug

evgeniycheban:
Hi @noelbundick-msft!

I've checked the sample you provided and I see that the `findPreAuthorizeAnnotation` performs search for `@PreAuthorize` annotation on the class that declares the called method in this case `SimpleJpaRepository` instead of `$Proxy` that implements repository interface.

This is definitely a bug, thanks for reporting.

The code I used to reproduce this:
```java
@EnableMethodSecurity
// @EnableGlobalMethodSecurity(prePostEnabled = true)
@Configuration
public class MethodSecurityConfig {}
```
```java
@PreAuthorize("hasRole('BOGUS')")
public interface ThingRepository extends JpaRepository<Thing, Long> {
  @Override
  @PreAuthorize("hasRole('BOGUS')")
  Optional<Thing> findById(Long id);
}
```
```java
Thing testThing = new Thing();
testThing.setName("testName");
this.repository.save(testThing); // Should throw `AccessDeniedException` but it doesn't.
```

rwinch:
Thanks for the report, I closed this in favor of gh-11177

PR #11177, opened 2022-04-29
@EnableMethodSecurity doesn't resolve annotations on interfaces through a Proxy

evgeniycheban:
@EnableMethodSecurity doesn't resolve annotations on interfaces through a Proxy

Removed proxy unwrapping in case of resolving Method Security annotations,
this cause an issue when interfaces which are implemented by the proxy was skipped,
resulting in a missing security checks on those methods.

Closes gh-11175


rwinch:
Thank you for the PR @evgeniycheban Can you please add some tests that would fail without the changes you have made and pass once the changes you made are merged?

evgeniycheban:
@rwinch I've added tests for `@Secured` and `@PreAuthorize` annotations that covers these changes.

rwinch (approved):

Issue #11180, opened 2022-05-03
Schedule release automation on maintenance branches


sjohnr:
**Expected Behavior**

The `release-next-version.yml` workflow should be triggered on each maintenance branch.

**Current Behavior**

GitHub Actions only performs scheduled actions against `main`.

sjohnr:
Related gh-10451

PR #11183, opened 2022-05-04
Add workflow for scheduling releases on maintenance branches

sjohnr:
Closes gh-11180

### DIFF
--- a/.github/workflows/release-scheduler.yml
+++ b/.github/workflows/release-scheduler.yml
@@ -1,0 +1,29 @@
+name: Release Scheduler
+
+on:
+  schedule:
+    - cron: '15 15 * * MON' # Every Monday at 3:15pm UTC
+  workflow_dispatch:
+
+jobs:
+  dispatch_scheduled_releases:
+    name: Dispatch scheduled releases
+    if: ${{ github.repository == 'spring-projects/spring-security' }}
+    strategy:
+      matrix:
+        #
+        # List of active maintenance branches.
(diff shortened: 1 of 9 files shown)


.github/workflows/release-next-version.yml
```diff
@@ -1,21 +1,18 @@
 name: Release Next Version
```
rwinch, line 1: We should consider renaming release-next-version to something like update-scheduled-release-version

buildSrc/src/main/java/org/springframework/gradle/github/release/TriggerReleaseTask.java
```diff
@@ -0,0 +1,68 @@
...
+import org.gradle.api.DefaultTask;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.TaskAction;
+
+import org.springframework.gradle.github.RepositoryRef;
+
+/**
+ * @author Steve Riesenberg
+ */
+public class TriggerReleaseTask extends DefaultTask {
```
rwinch, line 28: I wonder if it might be GitHubDispatchWorkflowTask along with the ability to specify the workflow id to invoke.

rwinch (changes requested):
I've provided some feedback inline

sjohnr:
@rwinch I've updated the PR with some renaming and made the gradle task reusable.

rwinch (approved):

PR #11184, opened 2022-05-05
Add workflow step for scheduling the next milestone

sjohnr:
This step introduces the concept of a release train to determine the appropriate release dates and schedules milestones as needed. It uses the following rules:

- If the current SNAPSHOT version is a minor release (patchVersion == 0), checks to see if a GA release exists. If not, a release train is scheduled as follows:
  - Finds the next available release train starting month (including this month), either January or July.
  - Schedules a release train with M1, M2, M3, RC1 and GA versions on either January/February/March/April/May or July/August/September/October/November.
- If the current SNAPSHOT version is a patch release (patchVersion != 0), checks to see if a GA release exists. If not, an individual milestone is scheduled on the next even month whose release day is in the future (can include this month).
- In either case, the release day used to schedule the release is based on the configured weekOfMonth (1-4) and dayOfWeek (1-5), e.g. 3rd Monday of the month (3, 1).

If a milestone already exists, nothing is created. Once created, milestone due dates can be updated manually as desired.

Closes https://github.com/spring-projects/spring-security/issues/10458


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -294,6 +296,20 @@ jobs:
...
+    timeout-minutes: 90
+    if: ${{ endsWith(needs.prerequisites.outputs.project_version, '-SNAPSHOT') }}
+    env:
+      TOKEN: ${{ github.token }}
+      VERSION: ${{ needs.prerequisites.outputs.project_version }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: spring-io/spring-gradle-build-action@v1
+      - name: Schedule next release (if not already scheduled)
+        run: ./gradlew scheduleNextRelease -PnextVersion=$VERSION -PgitHubAccessToken=$TOKEN
```
rwinch, line 312: Is nextVersion an accurate name here? To me it appears that this is the current version because the release has been performed. The scheduleNextRelease task refers to it as version so perhaps that is a better argument? 

i'm also curious if Gradle really needs to read the version from the prerequisites.outputs. Could Gradle just access the project.version instead?

rwinch (changes requested):

.github/workflows/continuous-integration-workflow.yml
rwinch: I think I see and agree now. nextVersion is in the context of what is being scheduled and that had not been done yet even though the release has already been done. Thanks!

PR #11187, opened 2022-05-07
createEvaluationContext should defer lookup of Authentication

evgeniycheban:
`createEvaluationContext` should defer lookup of `Authentication`

- Added `createEvaluationContext` method that accepts `Supplier<Authentication>`
- Refactored classes that use `EvaluationContext` to use lazy initialization of `Authentication`

Closes gh-9667

### DIFF
--- a/core/src/main/java/org/springframework/security/access/expression/AbstractSecurityExpressionHandler.java
+++ b/core/src/main/java/org/springframework/security/access/expression/AbstractSecurityExpressionHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -35,6 +35,7 @@ import org.springframework.util.Assert;
  * objects.
  *
  * @author Luke Taylor
+ * @author Evgeniy Cheban
  * @since 3.1
  */
 public abstract class AbstractSecurityExpressionHandler<T>
@@ -114,6 +115,10 @@ public abstract class AbstractSecurityExpressionHandler<T>
 
 	public void setPermissionEvaluator(PermissionEvaluator permissionEvaluator) {
 		this.permissionEvaluator = permissionEvaluator;
+	}
+
+	protected BeanResolver getBeanResolver() {
+		return this.beanResolver;
 	}
 
 	@Override
--- a/core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java
+++ b/core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,6 +19,7 @@ package org.springframework.security.access.expression;
 import java.io.Serializable;
 import java.util.Collection;
 import java.util.Set;
+import java.util.function.Supplier;
 
 import org.springframework.security.access.PermissionEvaluator;
 import org.springframework.security.access.hierarchicalroles.RoleHierarchy;
@@ -26,16 +27,18 @@ import org.springframework.security.authentication.AuthenticationTrustResolver;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.authority.AuthorityUtils;
+import org.springframework.util.Assert;
 
 /**
  * Base root object for use in Spring Security expression evaluations.
  *
  * @author Luke Taylor
+ * @author Evgeniy Cheban
  * @since 3.0
  */
 public abstract class SecurityExpressionRoot implements SecurityExpressionOperations {
 
-	protected final Authentication authentication;
+	private final Supplier<Authentication> authentication;
 
 	private AuthenticationTrustResolver trustResolver;
 
@@ -72,10 +75,18 @@ public abstract class SecurityExpressionRoot implements SecurityExpressionOperat
 	 * @param authentication the {@link Authentication} to use. Cannot be null.
 	 */
 	public SecurityExpressionRoot(Authentication authentication) {
-		if (authentication == null) {
-			throw new IllegalArgumentException("Authentication object cannot be null");
-		}
-		this.authentication = authentication;
+		this(() -> authentication);
+	}
+
+	/**
+	 * Creates a new instance that uses lazy initialization of the {@link Authentication}
+	 * object.
+	 * @param authentication the {@link Supplier} of the {@link Authentication} to use.
+	 * Cannot be null.
+	 * @since 5.8
+	 */
+	public SecurityExpressionRoot(Supplier<Authentication> authentication) {
+		this.authentication = new AuthenticationSupplier(authentication);
 	}
 
 	@Override
@@ -111,7 +122,7 @@ public abstract class SecurityExpressionRoot implements SecurityExpressionOperat
 
 	@Override
 	public final Authentication getAuthentication() {
-		return this.authentication;
+		return this.authentication.get();
 	}
 
 	@Override
@@ -126,7 +137,7 @@ public abstract class SecurityExpressionRoot implements SecurityExpressionOperat
 
 	@Override
 	public final boolean isAnonymous() {
-		return this.trustResolver.isAnonymous(this.authentication);
+		return this.trustResolver.isAnonymous(getAuthentication());
 	}
 
 	@Override
@@ -136,13 +147,13 @@ public abstract class SecurityExpressionRoot implements SecurityExpressionOperat
 
 	@Override
 	public final boolean isRememberMe() {
-		return this.trustResolver.isRememberMe(this.authentication);
+		return this.trustResolver.isRememberMe(getAuthentication());
 	}
 
 	@Override
 	public final boolean isFullyAuthenticated() {
-		return !this.trustResolver.isAnonymous(this.authentication)
-				&& !this.trustResolver.isRememberMe(this.authentication);
+		Authentication authentication = getAuthentication();
+		return !this.trustResolver.isAnonymous(authentication) && !this.trustResolver.isRememberMe(authentication);
 	}
 
 	/**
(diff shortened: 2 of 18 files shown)


core/src/main/java/org/springframework/security/access/expression/AbstractSecurityExpressionHandler.java
```diff
@@ -75,6 +78,23 @@ public final EvaluationContext createEvaluationContext(Authentication authentica
...
+	/**
+	 * Invokes the internal template methods to create {@code StandardEvaluationContext}
+	 * and {@code SecurityExpressionRoot} objects.
+	 * @param authentication the {@code Supplier} of the current authentication object
+	 * @param invocation the invocation (filter, method, channel)
+	 * @return the context object for use in evaluating the expression, populated with a
+	 * suitable root object.
+	 */
+	@Override
+	public final EvaluationContext createEvaluationContext(Supplier<Authentication> authentication, T invocation) {
```
jzheaux, line 90: While I appreciate the value of following the prevailing pattern, I think it would be better for `DefaultWeb`/`DefaultHttp`/`DefaultMethod`/`DefaultMessageSecurityExpressionHandler` to implement this method directly instead of replicating the protected methods.

jzheaux: Obviously, we can add the protected methods later if that becomes necessary.

core/src/main/java/org/springframework/security/core/SupplierAuthentication.java
```diff
@@ -0,0 +1,147 @@
...
+
+import org.springframework.util.Assert;
+
+/**
+ * An {@link Authentication} that lazily initializes another {@link Authentication}.
+ *
+ * @author Evgeniy Cheban
+ * @since 5.7
+ */
+public final class SupplierAuthentication implements Authentication {
```
jzheaux, line 30: Cool solution. Instead of doing this, I'd prefer to see if we can keep contracts explicit to it's clear that a given API is deferring the lookup.

If this is ultimately needed to make `SecurityExpressionRoot` work, then maybe it can be inlined to that class.

core/src/main/java/org/springframework/security/access/expression/method/MethodSecurityEvaluationContext.java
```diff
@@ -52,6 +53,10 @@ class MethodSecurityEvaluationContext extends MethodBasedEvaluationContext {
 		super(mi.getThis(), getSpecificMethod(mi), mi.getArguments(), parameterNameDiscoverer);
 	}
 
+	MethodSecurityEvaluationContext(MethodInvocation mi, ParameterNameDiscoverer parameterNameDiscoverer) {
```
jzheaux, line 56: Thanks for this cleanup. Will you please place this constructor into a separate commit since it's not directly about `Supplier<Authentication>`?

jzheaux (changes requested):
Thanks, @evgeniycheban! I've left some feedback inline. Also, will you please rebase this against `5.8.x` as I imagine that's were this change will land?

core/src/main/java/org/springframework/security/core/SupplierAuthentication.java
evgeniycheban: I agree. It would be much easier if we can change `SecurityExpressionRoot.authentication` to `private Supplier<Authentication>`, for those users who use `this.authentication` shouldn't be too difficult to change their code to `getAuthentication()`.

If we'll go with `private Supplier<Authentication>` approach it might be reasonable to introduce a `Supplier` that memorises the value:
```
private static final class AuthenticationSupplier implements Supplier<Authentication> {
    private final Supplier<Authentication> delegate;
    private volatile Authentication value;
    
    @Override
    public Authentication get() {
	if (this.value == null) {
	    synchronized (this.delegate) {
		if (this.value == null) {
		    this.value = this.delegate.get();;
	        }
            }
	}
	return this.value;
    }
}
```
Therefore for the `expression` like `hasRole('USER') or hasRole('ADMIN')` we only obtain an `Authentication` object once instead of calling `Supplier#get` for every `hasRole` in the `expression`.

core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java
```diff
@@ -78,6 +81,16 @@ public SecurityExpressionRoot(Authentication authentication) {
...
 	}
 
+	/**
+	 * Creates a new instance that uses lazy initialization of the {@link Authentication}
+	 * object.
+	 * @param authentication the {@link Supplier} of the {@link Authentication} to use.
+	 * Cannot be null.
+	 */
+	public SecurityExpressionRoot(Supplier<Authentication> authentication) {
+		this.authentication = new SupplierAuthentication(authentication);
```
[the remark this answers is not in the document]
rwinch: I'm not certain of this, but I think that changing the member variable to private will break SpEL expressions that use `#authentication`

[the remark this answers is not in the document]
jzheaux: Hmm, okay. I wondered if that was the case since I can access `principal` in a SpEL expression without there being a `principal` member variable. I'll do some experimenting and see.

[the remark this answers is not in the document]
jzheaux: I ran this test against `SecurityExpressionRoot` with `protected Authentication authentication`, `private Authentication authentication`, and `private Supplier<Authentication> authentication`:

```java
@Test
public void authenticationIsCorrectlyResolved() {
	Authentication authentication = new TestingAuthenticationToken("user", "password");
	ExpressionParser parser = new SpelExpressionParser();
	EvaluationContext context = new StandardEvaluationContext(new SecurityExpressionRoot(authentication) {});
	Expression root = parser.parseExpression("authentication.name == 'user'");
	assertThat(root.getValue(context)).isEqualTo(true);
	Expression variable = parser.parseExpression("#authentication.name == 'user'");
	assertThatExceptionOfType(SpelEvaluationException.class).isThrownBy(() -> variable.getValue(context));
}
```

And it passed for all three setups because the `Expression` uses the method resolver to resolve `authentication`. In no case could `authentication` be accessed as `#authentication`. Hypothetically, if we do need this, I think that we can do something like `context.setVariable("authentication", authentication)` during setup.

To remove noise, I ran a test against two custom objects unrelated to Security, other than to emulate the change from a `protected` field of Type A to a `private` field of Type B. In both classes, the public getter method remains the same:

```java
@Test
public void valueFieldIsNotDirectlyReferenced() {
	ExpressionParser parser = new SpelExpressionParser();
	Expression method = parser.parseExpression("value.length() == 2");
	EvaluationContext privateField = new StandardEvaluationContext(new WithPrivateField());
	EvaluationContext protectedField = new StandardEvaluationContext(new WithProtectedField());
	assertThat(method.getValue(privateField)).isEqualTo(true);
	assertThat(method.getValue(protectedField)).isEqualTo(true);
	Expression field = parser.parseExpression("#value.length() == 2");
	assertThatExceptionOfType(SpelEvaluationException.class).isThrownBy(() -> field.getValue(privateField));
	assertThatExceptionOfType(SpelEvaluationException.class).isThrownBy(() -> field.getValue(protectedField));
}

private static class WithPrivateField {
	private Integer value = 13;

	public String getValue() {
		return String.valueOf(this.value);
	}
}

private static class WithProtectedField {
	protected String value = "13";

	public String getValue() {
		return this.value;
	}
}
```

The test again passes.

Also note that if I remove the method `getValue()` from `WithProtectedField`, the test fails with "Property or field 'value' cannot be found on object of type 'org.springframework.security.access.expression.AbstractSecurityExpressionHandlerTests$WithProtectedField' - maybe not public or not valid?". Only if I also make the field `public` does it pass.

Since the property in `SecurityExpressionRoot` is not `public`, I don't think it has any bearing on how expressions are evaluated.

Let me know if you see other rocks we should try and uncover.

core/src/main/java/org/springframework/security/access/expression/AbstractSecurityExpressionHandler.java
evgeniycheban: Done.

core/src/main/java/org/springframework/security/access/expression/method/MethodSecurityEvaluationContext.java
evgeniycheban: This constructor is used in `DefaultMethodSecurityExpressionHandler.createEvaluationContext` so let’s keep it in one commit.

core/src/main/java/org/springframework/security/authentication/AuthenticationTrustResolverImpl.java
```diff
@@ -48,6 +50,9 @@ public boolean isAnonymous(Authentication authentication) {
 		if ((this.anonymousClass == null) || (authentication == null)) {
 			return false;
 		}
+		if (authentication instanceof SupplierAuthentication supplierAuthentication) {
```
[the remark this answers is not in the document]
evgeniycheban: Removed.

evgeniycheban:
@jzheaux I've updated the PR and rebased this branch on top of `5.8.x`, please take a look.
I've made changes to `SecurityExpressionRoot.authentication` now it’s `private Supplier<Authentication>` that can memorise the value.
Also I've added an implementation of new `createEvaluationContext` method for `DefaultHttpSecurityExpressionHandler`.

core/src/main/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandler.java
```diff
@@ -77,12 +79,26 @@ public StandardEvaluationContext createEvaluationContextInternal(Authentication
...
+		return ctx;
+	}
+
 	/**
 	 * Creates the root object for expression evaluation.
 	 */
 	@Override
 	protected MethodSecurityExpressionOperations createSecurityExpressionRoot(Authentication authentication,
 			MethodInvocation invocation) {
+		return createSecurityExpressionRoot(() -> authentication, invocation);
```
jzheaux, line 97: This new code seems unnecessary. Can the extra private method be removed?

jzheaux (changes requested):
Looking good, @evgeniycheban. I added one cleanup mention.

Additionally, are you able to add an implementation for `DefaultMessageSecurityExpressionHandler`?

core/src/main/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandler.java
evgeniycheban: I reused `createSecurityExpressionRoot` in the `createEvaluationContext` above.

jzheaux:
Thanks, @evgeniycheban! This is now merged into `5.8.x` as well as `main`.

Issue #11188, opened 2022-05-08
Consider replacing an inner loop with Set of authority strings in AuthorityAuthorizationManager#isAuthorized


evgeniycheban:
Currently `AuthorityAuthorizationManager#isAuthorized` uses an inner loop to determine if user has required authority:
```
private boolean isAuthorized(Authentication authentication) {
    for (GrantedAuthority grantedAuthority : authentication.getAuthorities()) {
	for (GrantedAuthority authority : this.authorities) {
	    if (authority.getAuthority().equals(grantedAuthority.getAuthority())) {
		return true;
            }
	}
    }
    return false;
}
```
It can be replaced with a `Set` of authority strings which will be more efficient because `HashSet` has constant lookup time O(1).

evgeniycheban:
@jzheaux I've just found `AuthorityUtils#authorityListToSet` method that does the same I did in `getAuthoritySet`, can I open another PR that will add this minor change?

PR #11191, opened 2022-05-10
Polish gh-11188

evgeniycheban:
Refers https://github.com/spring-projects/spring-security/issues/11188#issuecomment-1121701811


sjohnr:
Thanks @evgeniycheban. I'll add this to `5.8.x` and `main` as well.

cc: @jzheaux

Issue #11195, opened 2022-05-10
Add relyingPartyRegistrationId to AbstractSaml2AuthenticationRequest


jzheaux:
It would be nice if implementations of `AbstractSaml2AuthenticationRequest` stored the `relyingPartyRegistrationId`.

One benefit is it introduces the ability to look up the registration id during login. Currently, the only way to do this is by path or by implementing a custom `RelyingPartyRegistrationResolver`.

The needed changes are as follows:

* Adding the registration id as a property in `AbstractSaml2AuthenticationRequest` (see `Saml2LogoutRequest` and `Saml2LogoutResponse` for examples).
* Set that property in the `withRelyingPartyRegistration` static methods
* Read the property in `Saml2AuthenticationTokenConverter` to determine the registration id, passing that to the configured `relyingPartyRegistrationResolver`

Note that these changes should be made in such a way that receiving an unsolicited SAML response still works (e.g. there is no stored AuthnRequest).

ugrave:
Created a PR #11198 which adds the registration id to the request.

PR #11198, opened 2022-05-12
Add relyingPartyRegistrationId to AbstractSaml2AuthenticationRequest

ugrave:
Closes gh-11195

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/AbstractSaml2AuthenticationRequest.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/AbstractSaml2AuthenticationRequest.java
@@ -20,6 +20,7 @@ import java.io.Serializable;
 import java.nio.charset.Charset;
 
 import org.springframework.security.core.SpringSecurityCoreVersion;
+import org.springframework.security.saml2.provider.service.registration.RelyingPartyRegistration;
 import org.springframework.security.saml2.provider.service.registration.Saml2MessageBinding;
 import org.springframework.util.Assert;
 
@@ -46,6 +47,8 @@ public abstract class AbstractSaml2AuthenticationRequest implements Serializable
 
 	private final String authenticationRequestUri;
 
+	private final String relyingPartyRegistrationId;
+
 	/**
 	 * Mandatory constructor for the {@link AbstractSaml2AuthenticationRequest}
 	 * @param samlRequest - the SAMLRequest XML data, SAML encoded, cannot be empty or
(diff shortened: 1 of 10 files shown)


saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/AbstractSaml2AuthenticationRequest.java
```diff
@@ -46,20 +47,26 @@ public abstract class AbstractSaml2AuthenticationRequest implements Serializable
...
 	 * @param authenticationRequestUri - The authenticationRequestUri, a URL, where to
 	 * send the XML message, cannot be empty or null
+	 * @param relyingPartyRegistrationId the registration id of the relying party
 	 */
-	AbstractSaml2AuthenticationRequest(String samlRequest, String relayState, String authenticationRequestUri) {
+	AbstractSaml2AuthenticationRequest(String samlRequest, String relayState, String authenticationRequestUri,
+			String relyingPartyRegistrationId) {
 		Assert.hasText(samlRequest, "samlRequest cannot be null or empty");
 		Assert.hasText(authenticationRequestUri, "authenticationRequestUri cannot be null or empty");
+		Assert.hasText(relyingPartyRegistrationId, "relyingPartyRegistrationId cannot be null or empty");
```
jzheaux, line 65: Let's remove this so that payloads Jackson-serialized in 5.7 can be deserialized in 5.8. We can add a ticket to require this in Spring Security 6 where we can make a breaking change.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2PostAuthenticationRequest.java
```diff
@@ -50,7 +51,8 @@ public Saml2MessageBinding getBinding() {
 	 */
 	public static Builder withRelyingPartyRegistration(RelyingPartyRegistration registration) {
 		String location = registration.getAssertingPartyDetails().getSingleSignOnServiceLocation();
-		return new Builder().authenticationRequestUri(location);
+		return new Builder().authenticationRequestUri(location)
```
jzheaux, line 54: Instead of calling a mutator method, I think that the `Builder` constructor should take a registration id as a value and that value should be immutable.

jzheaux (changes requested):
Thanks, @ugrave! I've left some feedback inline.

jzheaux:
Also, @ugrave, will you please rebase off of `5.8.x` since that's the release that this will go out on?

ugrave:
I had some problems during rebase. I created a new PR #11230 with 5.8 as base/target.

Issue #11204, opened 2022-05-12
NPE on requests with no content-type header with MultipartFormData enabled


ZhivkoDelchev:
**Bug description**
Performing non-GET requests with no content-type header results in a NullPointerException from CsrfWebFilter when MultipartFormData is enabled.

**To Reproduce**
Perform non-GET request with no body & `content-type` header with MultipartFormData enabled.

**Expected behavior**
All methods without body & `content-type` header should work.

**Stacktrace**

```
500 Server Error for HTTP POST "...."
java.lang.NullPointerException: null
   at org.springframework.security.web.server.csrf.CsrfWebFilter.tokenFromMultipartData(CsrfWebFilter.java:154)
   Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException: 
Error has been observed at the following site(s):
   *__checkpoint ⇢ org.springframework.security.web.server.csrf.CsrfWebFilter [DefaultWebFilterChain]
   *__checkpoint ⇢ com.tis.platform.webgateway.filter.FormDataCacheFilterFactory$FormDataCacheFilter [DefaultWebFilterChain]
   *__checkpoint ⇢ org.springframework.security.web.server.header.HttpHeaderWriterWebFilter [DefaultWebFilterChain]
   *__checkpoint ⇢ org.springframework.security.config.web.server.ServerHttpSecurity$ServerWebExchangeReactorContextWebFilter [DefaultWebFilterChain]
   *__checkpoint ⇢ org.springframework.security.web.server.WebFilterChainProxy [DefaultWebFilterChain]
   *__checkpoint ⇢ org.springframework.boot.actuate.metrics.web.reactive.server.MetricsWebFilter [DefaultWebFilterChain]
   *__checkpoint ⇢ com.tis.commons.traceid.reactive.TraceIdWebFilter [DefaultWebFilterChain]
   *__checkpoint ⇢ HTTP POST "...." [ExceptionHandlingWebHandler]
Original Stack Trace:
      at org.springframework.security.web.server.csrf.CsrfWebFilter.tokenFromMultipartData(CsrfWebFilter.java:154)
      at org.springframework.security.web.server.csrf.CsrfWebFilter.containsValidCsrfToken(CsrfWebFilter.java:143)
      at org.springframework.security.web.server.csrf.CsrfWebFilter.lambda$validateToken$5(CsrfWebFilter.java:136)
[...]

PR #11205, opened 2022-05-12
Reverse content type check

ZhivkoDelchev:
When MultipartFormData is enabled currently the CsrfWebFilter compares
the content-type header against MULTIPART_FORM_DATA MediaType which
leads to NullPointerExecption when there is no content-type header.
This change reverse the check to compare the MULTIPART_FORM_DATA
MediaType against the content-type which contains null check and avoids
the exception.

Closes gh-11204


rwinch (approved):

ZhivkoDelchev:
Formatting issue fixed.

rwinch:
Thanks for the Pull Request @ZhivkoDelchev! This is now merged into main :smile:

PR #11230, opened 2022-05-16
Add relyingPartyRegistrationId to AbstractSaml2AuthenticationRequest

ugrave:
Closes gh-11195


saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/AbstractSaml2AuthenticationRequest.java
```diff
@@ -108,7 +125,15 @@ public String getAuthenticationRequestUri() {
 
 		String relayState;
 
-		protected Builder() {
```
jzheaux, line 111: We cannot remove this constructor since that could break applications that are extending this class. Instead, please mark it as `@Deprecated` so that it can be removed in a future major release.

jzheaux (approved):

jzheaux:
Thanks, @ugrave! This is now merged into `5.8.x` and forward-ported to `main`.

Issue #11197, opened 2022-05-11
ClientRegistrations#fromIssuerLocation should have shorter (or configurable) Http Connect/Read timeouts


Kehrlann:
**Expected Behavior**

When using `ClientRegistrations.fromIssuerLocation("http://192.0.2.0")` or any non-responding address[1], the http connect timeout should be "reasonable", in seconds so that users get quick feedback when a issuerUri just doesn't respond. Reasonable _may_ be somewhere between 5 and 10s.

Or the `ClientRegistrations.rest` RestTemplate could be configurable.

**Current Behavior**

The current timeout is whatever the default timeout is for RestTemplate / underlying HTTP client implementation. In a "default" Boot application, that timeout is 75s. This means the application will hang for 75 seconds, and Tomcat won't even start up. 

That is too long in certain contexts, e.g. a Boot app deployed to Kubernetes. The app would take 75 seconds to respond, which could be longer than the user's readiness probe - the pod would get thrashed without giving useful information to the user.

**Context**

We have seen this issue in k8s deployments, where our pods would crash because a DNS record was (incorrectly) changed and an OIDC issuer did not respond anymore.

**Sample**

A Boot app with OAuth2 login enabled, and the following config:

```yml
spring:
  security:
    oauth2:
      client:
        registration:
          default:
            client-id: "foo"
            client-secret: "bar"
            provider: my-provider
        provider:
          my-provider:
            issuer-uri: https://192.0.2.0 # or any other non-responding IP
```

This can be reproduced programmatically by calling `ClientRegistrations.fromIssuerLocation("http://192.0.2.0");`

Happy to contribute a PR.

#### Footnotes

[1] Used `192.0.2.0` as [RFC5737 - IPv4 Address Blocks Reserved for Documentation](https://www.rfc-editor.org/rfc/rfc5737), works as an example for MacOS but I haven't tried on other systems.

sjohnr:
Hi @Kehrlann! I think this is closely related to gh-8882, which aims to solve this for a variety of components in the OAuth 2.0 landscape.

See [ClientRegistrations](https://github.com/spring-projects/spring-security/issues/8882#user-content-client-registrations) section, which states:

> NOTE: The underlying HTTP Client used in ClientRegistrations was purposely encapsulated and there is no plan to expose it.

[...]

PR #11232, opened 2022-05-16
ClientRegistrations#rest defines 30s connect and read timeouts

Kehrlann:
Example PR for #11197 . Note: the `RestTemplate` uses a plain `SimpleHttpRequestFactory` which is the default through `RestTemplate - extends -> InterceptingHttpAccessor - extends ->  HttpAccessor`.

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
@@ -30,6 +30,7 @@ import net.minidev.json.JSONObject;
 
 import org.springframework.core.ParameterizedTypeReference;
 import org.springframework.http.RequestEntity;
+import org.springframework.http.client.SimpleClientHttpRequestFactory;
 import org.springframework.security.oauth2.core.AuthorizationGrantType;
 import org.springframework.security.oauth2.core.ClientAuthenticationMethod;
 import org.springframework.security.oauth2.core.oidc.IdTokenClaimNames;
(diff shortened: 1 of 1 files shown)


Kehrlann:
Closing re: closing gh-11197 , forgot about this PR. Thanks Joe!

Kehrlann:
On second thought, re-opened the PR. Those are sensible defaults.

sjohnr:
Hi @Kehrlann, I think we can look at possibly merging this after we add more flexibility around `ClientRegistrations`. See gh-11321. However, I wonder if we can discuss the timeout we would choose to default to. While probably not very scientific, I'm thinking of **two** questions that would drive this choice:

* Is it reasonable?
* Is it unobtrusive?

To me, 10 seconds is very reasonable as most requests should not take that long. However, it seems like it could be obtrusive when chosen as a default, because there may be many users that can or will experience events in production that suffer the effects of this short timeout, which could further compound any ongoing issues in their environment. It could be argued the opposite of course, that it helps by terminating open connections and not allowing them to compound, but this is always subjective and takes load testing and other factors to determine appropriate settings, hence we would ideally provide a way for full customization as discussed elsewhere.

While it's probably impossible to pick the "right" default, I think 30 seconds also seems reasonable while being less obtrusive. What do you think? Are there other factors that are worth considering?

Kehrlann:
Hi @sjohnr , 30 seconds sounds very un-obtrusive, while still being below the "1 minute healthcheck" threshold I've seen in the wild in some k8s scenarios.

Updated the PR accordingly.

sjohnr:
Merged into `5.8.x` and `main` as bea7761a1ceb5ff0f10ce0b918dc7db08fc90963.

Issue #10939, opened 2022-03-08
Not possible to take full advantage of BeanPostProcessor capabilities for beans configured via Spring Security


andreybpanfilov:
I do believe Spring Security Team performed a great work to allow us, application developers, configure sophisticated security things in simple way, however I have found out that sometimes it is required to somehow intercept authentication providers and http filters, and it is much simpler to take advantage of BeanPostProcessor capabilities rather than to reproduce OoTB configuration. Unfortunately in some cases it is not possible due to "oversights" in Spring Security, please check some examples below.

[AbstractDaoAuthenticationConfigurer#configure
](https://github.com/spring-projects/spring-security/blob/main/config/src/main/java/org/springframework/security/config/annotation/authentication/configurers/userdetails/AbstractDaoAuthenticationConfigurer.java#L85)

```java
	@Override
	public void configure(B builder) throws Exception {
                // here BPP is required to return instance of DaoAuthenticationProvider 
                // instead of more common as sufficient AuthenticationProvider
		this.provider = postProcess(this.provider);
		builder.authenticationProvider(this.provider);
	}
```

[LdapAuthenticationProviderConfigurer#getLdapAuthoritiesPopulator
](https://github.com/spring-projects/spring-security/blob/main/config/src/main/java/org/springframework/security/config/annotation/authentication/configurers/ldap/LdapAuthenticationProviderConfigurer.java#L134)

```java
	private LdapAuthoritiesPopulator getLdapAuthoritiesPopulator() {
		if (this.ldapAuthoritiesPopulator != null) {
			return this.ldapAuthoritiesPopulator;
		}
		DefaultLdapAuthoritiesPopulator defaultAuthoritiesPopulator = new DefaultLdapAuthoritiesPopulator(
				this.contextSource, this.groupSearchBase);
		defaultAuthoritiesPopulator.setGroupRoleAttribute(this.groupRoleAttribute);
		defaultAuthoritiesPopulator.setGroupSearchFilter(this.groupSearchFilter);
		defaultAuthoritiesPopulator.setSearchSubtree(this.groupSearchSubtree);
		defaultAuthoritiesPopulator.setRolePrefix(this.rolePrefix);
		this.ldapAuthoritiesPopulator = postProcess(defaultAuthoritiesPopulator);

                // here we are potentially returning bean which was not post processed
		return defaultAuthoritiesPopulator;
	}
```

[TrustedIssuerJwtAuthenticationManagerResolver#resolve
[...]

PR #11261, opened 2022-05-16
Fix visibility on addSessionAuthenticationStrategy in SessionManagementConfigurer

jebeaudet:
Hi

I believe this is a simple oversight since the class is `final` and the documentation points toward a public facing method.

A backport would be appreciated if possible!

Thanks

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/SessionManagementConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/SessionManagementConfigurer.java
@@ -260,7 +260,7 @@ public final class SessionManagementConfigurer<H extends HttpSecurityBuilder<H>>
 	 * @param sessionAuthenticationStrategy
 	 * @return the {@link SessionManagementConfigurer} for further customizations
 	 */
-	SessionManagementConfigurer<H> addSessionAuthenticationStrategy(
+	public SessionManagementConfigurer<H> addSessionAuthenticationStrategy(
 			SessionAuthenticationStrategy sessionAuthenticationStrategy) {
 		this.sessionAuthenticationStrategies.add(sessionAuthenticationStrategy);
 		return this;


sjohnr:
Hi @jebeaudet, thanks for the patch. Before proceeding, could you explain a bit more about the issue you're facing and how you hope to solve it with this change?

From looking at the code, it looks like the design of the `SessionManagementConfigurer` does not intend for applications to use this method directly. However, it appears there may be a workaround through the `CsrfConfigurer` if you're interested in doing this in the meantime.

jebeaudet:
Oh ok yea I see that it's used in a package protected fashion.

I had a use case where I wanted to add an audit on authentication via a session (not only for new authentication which is handled by the `ApplicationEventPublisher`). But after some tests I realized that this was not the proper place to do so anyway so I added a filter before `SessionManagementFilter`. 

I opened this PR anyway because I thought this was a simple oversight but now that you mention `CsrfConfigurer`, I realize this might have been intended but the intent is not clear as to why. IMO, it could be a valid use case to do something special with the `SessionAuthenticationStrategy` but I don't have a valid use case on top of my head.

TL;DR, feel free to close if you want, thanks!

sjohnr:
@jebeaudet thanks!

> IMO, it could be a valid use case to do something special with the SessionAuthenticationStrategy but I don't have a valid use case on top of my head.

It very well could be, though there could also be other approaches. This does sound vaguely familiar though, and I wonder if it's related to the idea of intercepting the result of AuthenticationProvider#authenticate, mentioned in [this comment](https://github.com/spring-projects/spring-security/issues/10939#issuecomment-1064721373) of #10939.

In any case, feel free to open another issue if there's something you feel isn't possible or add a comment to the above issue if it's relevant to you. I'll go ahead and close this for now as I'm not sure it would be desired to open up the method in this PR.

Issue #11196, opened 2022-05-11
Some Security Expressions cause NPE when used within @Query


slawekludwiczak:
**Describe the bug**
When I define method to provide currently authenticated user within repository, I get NullPointerException.
```
EL1022E: The function 'isAuthenticated' mapped to an object of type 'class org.springframework.security.access.expression.SecurityExpressionRoot' which cannot be invoked
org.springframework.expression.spel.SpelEvaluationException: EL1022E: The function 'isAuthenticated' mapped to an object of type 'class org.springframework.security.access.expression.SecurityExpressionRoot' which cannot be invoked
...
Caused by: java.lang.NullPointerException: Cannot invoke "org.springframework.security.authentication.AuthenticationTrustResolver.isAnonymous(org.springframework.security.core.Authentication)" because "this.trustResolver" is null
```
According to the [docs](https://docs.spring.io/spring-security/reference/features/integrations/data.html#data-query) you can use **all of the Common Security Expressions within the Query.**

**To Reproduce**
Use `isAuthenticated()`, or `isAnonymous()` inside `@Query`, i.e.:
```
@Query("SELECT u FROM User u WHERE u.username = :#{authentication.name} AND 1=:#{isAuthenticated()? 1:0}")
Optional<User> findCurrentUser();
```
You can run tests from sample repository.
Tested with Spring Boot 2.6.7 and 2.7.0-RC1

**Expected behavior**
No NPE exception

**Sample**
https://github.com/slawekludwiczak/spring-data-security-bug

Reports that include a sample will take priority over reports that do not.
At times, we may require a sample, so it is good to try and include a sample up front.

sjohnr:
Hi @slawekludwiczak, thanks for the report and the helpful sample!

I tried this out and can confirm what you're seeing. It seems to be caused by the `SecurityEvaluationContextExtension`:

https://github.com/spring-projects/spring-security/blob/f34ea188e29cafd6805236de21314905f1c03e0e/data/src/main/java/org/springframework/security/data/repository/query/SecurityEvaluationContextExtension.java#L109-L110

This instance of `SecurityExpressionRoot` doesn't ever seem to get a `trustResolver`. It seems this may have been a problem since this integration was introduced. Perhaps @rwinch could weigh in on context of this issue?

I'm not sure the changes needed in the repository, but I wonder if this could be worked around in the meantime by using a `@PostFilter` annotation? It may not be supported with `Optional`, and require returning a `Collection` of some kind instead.

[...]

PR #11263, opened 2022-05-17
Some Security Expressions cause NPE when used within @Query

evgeniycheban:
Added `trustResolver`, `roleHierarchy`, `permissionEvaluator`, `defaultRolePrefix`
fields to `SecurityEvaluationContextExtension` along with setter methods to override defaults.

Closes gh-11196

### DIFF
--- a/data/src/main/java/org/springframework/security/data/repository/query/SecurityEvaluationContextExtension.java
+++ b/data/src/main/java/org/springframework/security/data/repository/query/SecurityEvaluationContextExtension.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,10 +17,17 @@
 package org.springframework.security.data.repository.query;
 
 import org.springframework.data.spel.spi.EvaluationContextExtension;
+import org.springframework.security.access.PermissionEvaluator;
+import org.springframework.security.access.expression.DenyAllPermissionEvaluator;
 import org.springframework.security.access.expression.SecurityExpressionRoot;
+import org.springframework.security.access.hierarchicalroles.NullRoleHierarchy;
+import org.springframework.security.access.hierarchicalroles.RoleHierarchy;
+import org.springframework.security.authentication.AuthenticationTrustResolver;
+import org.springframework.security.authentication.AuthenticationTrustResolverImpl;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.context.SecurityContext;
 import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.util.Assert;
 
 /**
  * <p>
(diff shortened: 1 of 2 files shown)


data/src/main/java/org/springframework/security/data/repository/query/SecurityEvaluationContextExtension.java
```diff
@@ -77,12 +84,21 @@
...
+ * @author Evgeniy Cheban
  * @since 4.0
  */
 public class SecurityEvaluationContextExtension implements EvaluationContextExtension {
 
 	private Authentication authentication;
 
+	private AuthenticationTrustResolver trustResolver = new AuthenticationTrustResolverImpl();
+
+	private RoleHierarchy roleHierarchy = new NullRoleHierarchy();
```
sjohnr, line 96: This seems fine, but I think the `RoleHierarchy` can be `null` by default, since it is checked internally in the `SecurityExpressionRoot` and only used if provided. What do you think?

data/src/main/java/org/springframework/security/data/repository/query/SecurityEvaluationContextExtension.java
```diff
@@ -77,12 +84,21 @@
...
+
+	private PermissionEvaluator permissionEvaluator = new DenyAllPermissionEvaluator();
+
+	private String defaultRolePrefix = "ROLE_";
```
sjohnr, line 100: This also seems fine, and I noticed this pattern elsewhere in the codebase. However, since the `SecurityContextRoot` defaults to `"ROLE_"`, this seems duplicated. Could/should we reduce the number of places we specify the default role prefix by defaulting to `null` and checking whether to set it on the `SecurityContextRoot` if not `null`?

sjohnr (commented):
Thanks @evgeniycheban! Just a couple of comments for you.

data/src/main/java/org/springframework/security/data/repository/query/SecurityEvaluationContextExtension.java
evgeniycheban: Agreed. Perhaps, we should also consider adding a `null` check for `trustResolver` in `SecurityExpressionRoot` or assign a default value.

evgeniycheban:
@sjohnr I've made changes according to your comments also please see my [comment](https://github.com/spring-projects/spring-security/pull/11263#discussion_r881154172).

data/src/main/java/org/springframework/security/data/repository/query/SecurityEvaluationContextExtension.java
sjohnr: Thanks for working on this @evgeniycheban. Sadly, I wasn’t able to follow up on this quickly enough. I did have a conversation with the team about this and learned that Spring Security conventions prefer defaults over null checks. I’m still newer on the team so I’m still learning! So what you had in both of these cases was preferred. Would you mind force pushing 2bfd933 back (unless you applied any polish). I’m sorry for the trouble.

evgeniycheban: Done.

data/src/main/java/org/springframework/security/data/repository/query/SecurityEvaluationContextExtension.java
evgeniycheban: Remains without changes.

sjohnr (approved):

Issue #11262, opened 2022-05-17
5.6.3 SAML request encoding: on redirect binding, base64 encoded message contains CRLF


junytse:
**Describe the bug**

When SAML SP generate a SAML request to redirect user to IDP, the base64 encoded message has `\r\n` inserted every 76 character. When redirect binding is chosen, this will result in `%0A` appear in the url. The below issue may occur:
1. Some IDP implementation cannot parse line feeds in the encoded message, and it will result in failure on validation;
2. For [security reasons](https://book.hacktricks.xyz/pentesting-web/crlf-0d-0a), some web server or firewall implementation will have CRLF filter to also filter out `%0A` in URL. When this happen, the message sent out is modified and signature validation will fail;
3. It violates SAML spec "http://docs.oasis-open.org/security/saml/v2.0/saml-bindings-2.0-os.pdf" chapter 3.4.4.1 about redirect binding:
    > 3. The compressed data is subsequently base64-encoded according to the rules specified in IETF
RFC 2045 [RFC2045]. **Linefeeds or other whitespace MUST be removed from the result.**

**To Reproduce**
1. Config your SAML IDP to support redirect binding and let SAML SP (using Spring Security SAML SP) to send using redirect binding;
2. Trigger authentication request from SAML SP in a browser with developer mode on;
3. Check the request that redirect from SP to IDP, url params contains SAML message, which has `%0A`.

**Expected behavior**
The BASE64 encoded message should not have line feeds for redirect binding.

**Sample**

[...]

PR #11270, opened 2022-05-21
Use Base64 encoder with no CRLF in output for SAML 2.0 messages

junytse:
Saml2 message BASE64-encode results will not have line wrapped. 

gh-11262


jzheaux:
Thanks for the PR, @junytse. Will you please do me a favor and format your commit to include the issue that is being closed?

Like so:

```
Use Base64 encoder with no CRLF in output for SAML 2.0 messages

Closes gh-11262
```

This makes it easier to track changes in the future.

junytse:
> Thanks for the PR, @junytse. Will you please do me a favor and format your commit to include the issue that is being closed?
> 
> Like so:
> 
> ```
> Use Base64 encoder with no CRLF in output for SAML 2.0 messages
> 
> Closes gh-11262
> ```
> 
> This makes it easier to track changes in the future.

Updated.

Issue #10433, opened 2021-10-21
Resource server configured with JwtIssuerAuthenticationManagerResolver should publish authentication events


cselagea:
**Expected Behavior**

A single-tenant resource server with default configuration uses `ProviderManager` for authentication, which publishes authentication success and failure events using a `DefaultAuthenticationEventPublisher`.

A multi-tenant resource server configured to use `JwtIssuerAuthenticationManagerResolver` should publish the same authentication events.

**Current Behavior**

No authentication events are published.

**Context**

After it is determined that the issuer of the JWT can be trusted (e.g. via `TrustedIssuerJwtAuthenticationManagerResolver`) and the JWT is deemed valid, I'd like to deduce the tenant and store the tenant ID in a `ThreadLocal` context.

I could accomplish this using a filter placed after the Spring Security filter chain, but I think a cleaner solution would be listening for `AuthenticationSuccessEvent`, getting the issuer from the `JwtAuthenticationToken`, and then populating the tenant context.

cselagea:
I think this could be accomplished by adding an `AuthenticationEventPublisher` field initialized to a `NullEventPublisher` in `JwtIssuerAuthenticationManagerResolver` and then providing a setter for the publisher field.

```java
private AuthenticationEventPublisher eventPublisher = new NullEventPublisher();

public void setAuthenticationEventPublisher(AuthenticationEventPublisher eventPublisher) {
	Assert.notNull(eventPublisher, "AuthenticationEventPublisher cannot be null");
	this.eventPublisher = eventPublisher;
}
```

Then in `TrustedIssuerJwtAuthenticationManagerResolver`, instead of returning a `JwtAuthenticationProvider` directly, which has no event-publishing capabilities, an `AuthenticationEventPublishingJwtAuthenticationProvider` (or whatever it would be named) would be returned. This class would implement the `AuthenticationManager` contract by delegating to a `JwtAuthenticationProvider` instance and then use the publisher to publish `AuthenticationSuccessEvent`.

As for authentication failure, since `BearerTokenAuthenticationFilter` catches `AuthenticationException`, I suppose you wouldn't want to catch `AuthenticationException` in `AuthenticationEventPublishingJwtAuthenticationProvider`, so authentication failure events would be published by `BearerTokenAuthenticationFilter`.

cselagea:
@jzheaux could I give it a shot to create a PR for this?

jzheaux:
[...]

PR #11281, opened 2022-05-26
Publish authentication events when using JwtIssuerAuthenticationManagerResolver

cselagea:
Unlike a single-tenant resource server, which uses `ProviderManager` by default and hence publishes authentication success and failure events, a multi-tenant resource server using `JwtIssuerAuthenticationManagerResolver` does not, unless `JwtIssuerAuthenticationManagerResolver` was constructed using a custom event-publishing `issuerAuthenticationManagerResolver`.

Now `JwtIssuerAuthenticationManagerResolver` leverages `ProviderManager` too, so no matter how `JwtIssuerAuthenticationManagerResolver` is constructed authentication events will be published. A `NullEventPublisher` is used by default and a setter is added to enable wiring in a custom `AuthenticationEventPublisher`.

Closes gh-10433

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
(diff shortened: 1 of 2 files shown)


jzheaux:
@cselagea, my apologies for the delay on this PR. The more I think about it, the more it feels like this PR leaks an API that should sit at the `AuthenticationManager` level. Before continuing with this PR, I've added [a comment](https://github.com/spring-projects/spring-security/issues/10433#issuecomment-1150013345) to the original ticket that I'd like to get your feedback on.

jzheaux:
Closed based on https://github.com/spring-projects/spring-security/issues/10433#issuecomment-1157902677

Issue #11286, opened 2022-05-26
Support multiple SingleLogoutService bindings


emopti-jrufer:
**Expected Behavior**

Should be specify mulitple SingleLogoutService bindings. This would be incorporated into  Saml2LogoutRequestFilter, Saml2LogoutResponseFilter and OpenSamlMetadataResolver.

**Current Behavior**

RelyingPartyRegistration only support adding a single singleLogoutServiceBinding.

**Context**

Trying to migrate off of Spring SAML Extension. Existing behavior in the Spring SAML Extension allows support for multiple SLO bindings. 

Have not identified a workaround. This is preventing us from migrating since we currently support both REDIRECT and POST bindings for the SingleLogoutService and are unable to migrate all customers to utilize a single binding.

jzheaux:
Hi, @emopti-jrufer, thanks for the suggestion and the PR. We'll take a look and see if it's feasible to add into the next milestone release.

PR #11287, opened 2022-05-26
Support multiple SingleLogoutService bindings.

emopti-jrufer:
Closes gh-11286

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
@@ -28,6 +28,7 @@ import org.opensaml.xmlsec.signature.support.SignatureConstants;
 
 import org.springframework.security.saml2.core.Saml2X509Credential;
 import org.springframework.util.Assert;
+import org.springframework.util.CollectionUtils;
 
 /**
  * Represents a configured relying party (aka Service Provider) and asserting party (aka
@@ -81,7 +82,7 @@ public final class RelyingPartyRegistration {
 
 	private final String singleLogoutServiceResponseLocation;
 
-	private final Saml2MessageBinding singleLogoutServiceBinding;
+	private final Collection<Saml2MessageBinding> singleLogoutServiceBindings;
 
 	private final String nameIdFormat;
 
@@ -93,7 +94,7 @@ public final class RelyingPartyRegistration {
 
 	private RelyingPartyRegistration(String registrationId, String entityId, String assertionConsumerServiceLocation,
 			Saml2MessageBinding assertionConsumerServiceBinding, String singleLogoutServiceLocation,
-			String singleLogoutServiceResponseLocation, Saml2MessageBinding singleLogoutServiceBinding,
+			String singleLogoutServiceResponseLocation, Collection<Saml2MessageBinding> singleLogoutServiceBindings,
 			AssertingPartyDetails assertingPartyDetails, String nameIdFormat,
 			Collection<Saml2X509Credential> decryptionX509Credentials,
 			Collection<Saml2X509Credential> signingX509Credentials) {
@@ -101,8 +102,8 @@ public final class RelyingPartyRegistration {
 		Assert.hasText(entityId, "entityId cannot be empty");
 		Assert.hasText(assertionConsumerServiceLocation, "assertionConsumerServiceLocation cannot be empty");
 		Assert.notNull(assertionConsumerServiceBinding, "assertionConsumerServiceBinding cannot be null");
-		Assert.isTrue(singleLogoutServiceLocation == null || singleLogoutServiceBinding != null,
-				"singleLogoutServiceBinding cannot be null when singleLogoutServiceLocation is set");
+		Assert.isTrue(singleLogoutServiceLocation == null || !CollectionUtils.isEmpty(singleLogoutServiceBindings),
+				"singleLogoutServiceBindings cannot be null or empty when singleLogoutServiceLocation is set");
 		Assert.notNull(assertingPartyDetails, "assertingPartyDetails cannot be null");
 		Assert.notNull(decryptionX509Credentials, "decryptionX509Credentials cannot be null");
 		for (Saml2X509Credential c : decryptionX509Credentials) {
@@ -121,7 +122,7 @@ public final class RelyingPartyRegistration {
 		this.assertionConsumerServiceBinding = assertionConsumerServiceBinding;
 		this.singleLogoutServiceLocation = singleLogoutServiceLocation;
 		this.singleLogoutServiceResponseLocation = singleLogoutServiceResponseLocation;
-		this.singleLogoutServiceBinding = singleLogoutServiceBinding;
+		this.singleLogoutServiceBindings = Collections.unmodifiableList(new LinkedList<>(singleLogoutServiceBindings));
 		this.nameIdFormat = nameIdFormat;
 		this.assertingPartyDetails = assertingPartyDetails;
 		this.decryptionX509Credentials = Collections.unmodifiableList(new LinkedList<>(decryptionX509Credentials));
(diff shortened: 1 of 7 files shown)


saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
```diff
@@ -192,9 +194,26 @@ public Saml2MessageBinding getAssertionConsumerServiceBinding() {
 	 * relying party's &lt;SPSSODescriptor&gt;.
 	 * @return the SingleLogoutService Binding
 	 * @since 5.6
+	 * @deprecated use {@link #getSingleLogoutServiceBindings()}
 	 */
+	@Deprecated
```
jzheaux, line 199: I hesitate to mark this as deprecated. 

While I agree that the naming isn't ideal, it's quite common in SAML to have a situation where the spec supports multiple values but the majority of the time there is only one. Consider, for example, `Saml2AuthenticatedPrincipal#getFirstAttribute`. I imagine that it will be quite common that an application only supports one binding and now have that extra boilerplate of `get(0)` to do. If they are going to need to do it anyway, and we are already doing it, we might as well keep the method around. 

In that case, the JavaDoc would change to say that it returns the first one in the list.

I'm open to being convinced otherwise if you are strongly in favor of deprecation.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
```diff
@@ -93,16 +95,16 @@ public final class RelyingPartyRegistration {
...
 			AssertingPartyDetails assertingPartyDetails, String nameIdFormat,
 			Collection<Saml2X509Credential> decryptionX509Credentials,
 			Collection<Saml2X509Credential> signingX509Credentials) {
 		Assert.hasText(registrationId, "registrationId cannot be empty");
 		Assert.hasText(entityId, "entityId cannot be empty");
 		Assert.hasText(assertionConsumerServiceLocation, "assertionConsumerServiceLocation cannot be empty");
 		Assert.notNull(assertionConsumerServiceBinding, "assertionConsumerServiceBinding cannot be null");
-		Assert.isTrue(singleLogoutServiceLocation == null || singleLogoutServiceBinding != null,
-				"singleLogoutServiceBinding cannot be null when singleLogoutServiceLocation is set");
+		Assert.isTrue(singleLogoutServiceLocation == null || !CollectionUtils.isEmpty(singleLogoutServiceBindings),
```
jzheaux, line 106: I believe you can do `Assert#notEmpty` here

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/Saml2MessageBinding.java
```diff
@@ -26,13 +32,21 @@
  */
 public enum Saml2MessageBinding {
 
-	POST("urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST"), REDIRECT(
-			"urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect");
+	POST("urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST", (request) -> "POST".equalsIgnoreCase(request.getMethod())
```
jzheaux, line 35: We cannot do this work here since `HttpServletRequest` references are restricted to the `web` package.

Additionally, Spring Security prefers to place behavior as much as possible into components that implement interfaces. Generally, this simplifies applications overriding the behavior.

jzheaux (changes requested):
Thanks, @emopti-jruler! I've left my feedback inline.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
emopti-jrufer: I was mainly trying to preserve the contract. Would you be opposed to throwing an exception if there are multiple? My reasoning is that the term first would imply ordering which then would have to carry to other classes such as the SamlMetadataResolver.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
emopti-jrufer: `Assert#notEmpty` would not be appropriate since this assertion is first checking if `singleLogoutServiceLocation` has been set before checking `singleLogoutServiceBinding `.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
```diff
@@ -854,8 +868,8 @@ public Builder assertionConsumerServiceBinding(Saml2MessageBinding assertionCons
 		 * @return the {@link Builder} for further configuration
 		 * @since 5.6
 		 */
-		public Builder singleLogoutServiceBinding(Saml2MessageBinding singleLogoutServiceBinding) {
-			this.singleLogoutServiceBinding = singleLogoutServiceBinding;
+		public Builder singleLogoutServiceBinding(Saml2MessageBinding... singleLogoutServiceBinding) {
```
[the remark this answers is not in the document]
emopti-jrufer: This was done to prevent breaking changes. Are you okay with adding a second method in the builder with the consumer to preserve original contract?

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
```diff
@@ -855,7 +870,28 @@ public Builder assertionConsumerServiceBinding(Saml2MessageBinding assertionCons
 		 * @since 5.6
 		 */
 		public Builder singleLogoutServiceBinding(Saml2MessageBinding singleLogoutServiceBinding) {
-			this.singleLogoutServiceBinding = singleLogoutServiceBinding;
+			return this.singleLogoutServiceBindings((saml2MessageBindings) -> {
```
jzheaux, line 873: Is there a reason the new method needs to be called? I'm wondering if

```java
this.singleLogoutServiceBindings = Collections.singletonList(singleLogoutServiceBinding);
```

would work?

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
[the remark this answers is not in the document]
jzheaux: It's not clear to me how my suggestion introduces a breaking change. Can you please elaborate?

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
jzheaux: Sure, that sounds reasonable.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/binding/DefaultSaml2MessageBindingResolver.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.util.Collection;
+import java.util.List;
+
+import jakarta.servlet.http.HttpServletRequest;
+
+import org.springframework.security.saml2.provider.service.registration.Saml2MessageBinding;
+import org.springframework.util.Assert;
+
+public class DefaultSaml2MessageBindingResolver implements Saml2MessageBindingResolver {
```
jzheaux, line 27: It doesn't seem likely enough to switch out binding resolution behavior to introduce a new interface. If you feel strongly about introducing this configurability, please open a separate ticket so we can discuss.

For this PR, please don't introduce new interfaces or public classes.

jzheaux (changes requested):
Thanks, @emopti-jrufer! I've left some additional feedback inline.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/binding/DefaultSaml2MessageBindingResolver.java
emopti-jrufer: I agree that it is not likely switching out binding resolution. Maybe I misunderstood your prior comment about preferring to place behavior in components that implement interfaces (https://github.com/spring-projects/spring-security/pull/11287#discussion_r883967313). Would it be preferred to have this logic in a single utility class or do you have other suggestions?

jzheaux: @emopti-jrufer sorry, I can see how that wasn't clear. I don't think there's much value in an abstraction given that it would require changing or adding to the public API. 

I believe the result is that the behavior gets copied between classes.

jzheaux: Yes, so long as references to `HttpServletRequest` stay in the `web` package and the public API is not affected, then a package-private abstraction should be fine.

jzheaux (changes requested):

jzheaux (approved):

jzheaux:
Thanks, @emopti-jrufer! This is now merged into `main` and `5.8.x`

Issue #11318, opened 2022-06-01
Remove dependency on `conmmons-codec`  by using `java.util.Base64`


j3graham:
Currently some tests and one production class make use of commons-codec. The extra external dependency could be removed with the move to  a Java 17 baseline

Updating the tests is straightforward. 

The usage in `Saml2AuthenticationTokenConverter` needs some extra attention.

Currently the token converter uses the `STRICT` policy on the Base64 decoder. This disallows decoding of some "non-canonical" inputs, that could not have been generated by a compliant encoder.

Unfortunately `java.util.Base64` does not support the equivalent behaviour.

There is a test for the token converter that fails without this strictness, so extra checking would be needed.

I have written code to do this additional checking.

Independently, it would be interesting to confirm whether SAML requires rejection of non-canonical Base64, and if so, how strict a definition of canonical  is appropriate to enforce. 

PR to follow.

Refs:
`java.util.base64` accepts non-canonical encodings: https://www.mail-archive.com/core-libs-dev@openjdk.java.net/msg66424.html
Definition of "canonical Base64": https://datatracker.ietf.org/doc/html/rfc4648#section-3.5

rwinch:
Closing in favor of gh-1322

PR #11319, opened 2022-06-01
Remove dependency on conmmons-codec by using java.util.Base64

j3graham:
see gh-11318

First commit removes  test usages  of Base64 and some other minor commons-codec functions.

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverter.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverter.java
@@ -18,13 +18,13 @@ package org.springframework.security.saml2.provider.service.web;
 
 import java.io.ByteArrayOutputStream;
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.Base64;
 import java.util.function.Function;
 import java.util.zip.Inflater;
 import java.util.zip.InflaterOutputStream;
 
 import jakarta.servlet.http.HttpServletRequest;
-import org.apache.commons.codec.CodecPolicy;
-import org.apache.commons.codec.binary.Base64;
 
 import org.springframework.http.HttpMethod;
 import org.springframework.security.saml2.core.Saml2Error;
@@ -47,7 +47,11 @@ import org.springframework.util.Assert;
  */
 public final class Saml2AuthenticationTokenConverter implements AuthenticationConverter {
 
-	private static Base64 BASE64 = new Base64(0, new byte[] { '\n' }, false, CodecPolicy.STRICT);
+	// MimeDecoder allows extra line-breaks as well as other non-alphabet values.
+	// This matches the behaviour of the commons-codec decoder.
+	private static final Base64.Decoder BASE64 = Base64.getMimeDecoder();
+
+	private static final Base64Checker BASE_64_CHECKER = new Base64Checker();
 
 	private final RelyingPartyRegistrationResolver relyingPartyRegistrationResolver;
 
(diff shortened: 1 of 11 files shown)


rwinch:
Thanks for the PR @j3graham To start, I think another change might be to remove commons-codec occurences from `*.gradle` files so the dependency is no longer added to the classpath.

```
rg commons-codec
messaging/spring-security-messaging.gradle
18:	testImplementation 'commons-codec:commons-codec'

build.gradle
92:			components.withModule("commons-codec:commons-codec") { selection ->
95:					selection.reject("commons-codec updates break saml tests");

web/spring-security-web.gradle
23:	testImplementation 'commons-codec:commons-codec'

dependencies/spring-security-dependencies.gradle
26:		api "commons-codec:commons-codec:1.15"
```

I'll leave it to @jzheaux to review this further.

j3graham:
Dependencies removed.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverter.java
```diff
@@ -124,12 +127,59 @@ private String samlInflate(byte[] b) {
 			InflaterOutputStream inflaterOutputStream = new InflaterOutputStream(out, new Inflater(true));
 			inflaterOutputStream.write(b);
 			inflaterOutputStream.finish();
-			return out.toString(StandardCharsets.UTF_8.name());
+			return out.toString(StandardCharsets.UTF_8);
```
jzheaux, line 130: Thanks for this cleanup. Will you please place it in a separate commit so that it can be easily backported?

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverter.java
```diff
@@ -110,6 +112,7 @@ private String inflateIfRequired(HttpServletRequest request, byte[] b) {
 
 	private byte[] samlDecode(String base64EncodedPayload) {
 		try {
+			BASE_64_CHECKER.checkAcceptable(base64EncodedPayload);
```
jzheaux, line 115: What purpose does `BASE_64_CHECKER` serve? I imagine that if `BASE64.decode` fails, then the exception handling will be the same as before.

jzheaux (changes requested):
Thanks, @j3graham! I've left some feedback inline.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverter.java
j3graham: The old decoder was more strict about rejecting possibly invalid Base64 strings. The checker will reject those cases even though `java.util.Base64` would accept it. There is a test that fails if this checking is not present.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverter.java
j3graham: Done

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverter.java
j3graham: The other option would be to remove the failing test, but I can't assess the security significance of that change.

j3graham:
New PR for 5.8.x: https://github.com/spring-projects/spring-security/pull/11322

 javax->jakarta changes  were a bit awkward.

Perhaps we can abandon this one in favour of the 5.8.x one and let the merge to main happen.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverter.java
jzheaux: Sounds good. We can always relax the check later on if needed.

j3graham:
@jzheaux Should I fix up this PR, or will the one from 5.8 get merged independently?

PR #11367, opened 2022-06-12
Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -233,6 +233,8 @@ jobs:
           DOCS_SSH_KEY: ${{ secrets.DOCS_SSH_KEY }}
           DOCS_HOST: ${{ secrets.DOCS_HOST }}
   perform_release:
+    permissions:
+      contents: write  # for Git to git push
     name: Perform release
     needs: [prerequisites, deploy_artifacts, deploy_docs, deploy_schema]
     runs-on: ubuntu-latest
--- a/.github/workflows/milestone-spring-releasetrain.yml
+++ b/.github/workflows/milestone-spring-releasetrain.yml
@@ -7,6 +7,8 @@ env:
   TITLE: ${{ github.event.milestone.title }}
 jobs:
   spring-releasetrain-checks:    
+    permissions:
+      contents: none
     name: Check DueOn is on a Release Date
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/algolia-index.yml
+++ b/.github/workflows/algolia-index.yml
@@ -5,6 +5,9 @@ on:
     - cron: '0 10 * * *' # Once per day at 10am UTC
   workflow_dispatch: # Manual trigger
 
+permissions:
+  contents: read
+
 jobs:
   update:
     name: Update Algolia Index
--- a/.github/workflows/antora-generate.yml
+++ b/.github/workflows/antora-generate.yml
@@ -10,6 +10,9 @@ on:
 env:
   GH_ACTIONS_REPO_TOKEN: ${{ secrets.GH_ACTIONS_REPO_TOKEN }}
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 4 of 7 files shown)


rwinch (approved):
Thanks for the PR @naveensrinivasan  Looks good to me. @sjohnr Can you please take a second look over this and if it looks good merge?

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -233,6 +233,8 @@ jobs:
           DOCS_SSH_KEY: ${{ secrets.DOCS_SSH_KEY }}
           DOCS_HOST: ${{ secrets.DOCS_HOST }}
   perform_release:
+    permissions:
```
sjohnr, line 236: This file doesn't seem to have the default permissions set at the top. Should they be?

.github/workflows/milestone-spring-releasetrain.yml
```diff
@@ -7,6 +7,8 @@ env:
   TITLE: ${{ github.event.milestone.title }}
 jobs:
   spring-releasetrain-checks:    
+    permissions:
```
sjohnr, line 10: Should this file also have a default permission set at the top?

sjohnr (changes requested):
Thanks @naveensrinivasan! See comments below. Also, according to the scorecard reference you provided:

> The highest score is awarded when the permissions definitions in each workflow's yaml file are set as read-only at the [top level](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions) and the required write permissions are declared at the [run-level](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idpermissions). One point is reduced from the score if all jobs have their permissions defined but the top level permissions are not defined. This configuration is secure, but there is a chance that when a new job is added to the workflow, its job permissions could be left undefined because of human error.

If I'm understanding this correctly, we should have every workflow file in this repo contain at a minimum:

```yaml
permissions:
  contents: read
```

Am I understanding that correctly? I see the following missing files:
* backport-bot.yml
* release-scheduler.yml
* update-scheduled-release-version.yml

.github/workflows/continuous-integration-workflow.yml
naveensrinivasan: I can't determine what permissions are required at the top level.

.github/workflows/milestone-spring-releasetrain.yml
naveensrinivasan: I can't determine what permissions are required at the top level.

.github/workflows/continuous-integration-workflow.yml
sjohnr: Sorry for the delay, I was on break. Ok, thanks for the update, we can proceed with what you have here and go from there.

naveensrinivasan: 👍

sjohnr:
Thanks @naveensrinivasan, this is merged as 8f93a7fc94d566b2da79816fcf75925146ad8be6. I've also included a polish commit 3f4efedd23e615a6cfd686cb899fc48573ff3279 to add default permissions to each workflow.

vpavic:
FWIW, these kind of concerns are IMO better addressed at either repository or organization level under Actions settings:
image

That makes workflows easier to maintain as you only override permissions where needed.

Note that I left a similar comment in https://github.com/spring-projects/spring-boot/pull/31344#issuecomment-1158989081.

PR #11380, opened 2022-06-16
Fix method call example on documentation

vkhashimoto:


jzheaux:
Thanks, @vkhashimoto!

PR #11460, opened 2022-07-03
chore: Included githubactions in the dependabot config

naveensrinivasan:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


jzheaux:
Thanks, @naveensrinivasan. Does this take care of all dependencies? Given this configuration, it's not clear to me how it would know to update the `dependencies/spring-security-dependencies.gradle` file in its PRs. If it's only going to take care of dependencies listed in `build.gradle`, for example, I'm don't think this is the solution we want.

Also, I think it would be preferable if this could be configured in Gradle instead of in the `.github` directory so that the solution isn't specific to GHA.

naveensrinivasan:
> Thanks, @naveensrinivasan. Does this take care of all dependencies? Given this configuration, it's not clear to me how it would know to update the `dependencies/spring-security-dependencies.gradle` file in its PRs. If it's only going to take care of dependencies listed in `build.gradle`, for example, I'm don't think this is the solution we want.
> 
> Also, I think it would be preferable if this could be configured in Gradle instead of in the `.github` directory so that the solution isn't specific to GHA.

This PR is only for GitHub actions. I haven't included others because teams have mixed feelings on dependabot for other ecosystems like Gradle. HTH

jzheaux:
Thanks, @naveensrinivasan, that does help. Will this manage dependencies defined in places other than `build.gradle`? Spring Security's dependencies are defined in `dependencies/spring-security-dependencies.gradle`. I seem to remember that the Dependabot integration for Gradle has this limitation.

naveensrinivasan:
> Thanks, @naveensrinivasan, that does help. Will this manage dependencies defined in places other than `build.gradle`? Spring Security's dependencies are defined in `dependencies/spring-security-dependencies.gradle`. I seem to remember that the Dependabot integration for Gradle has this limitation.

Yes AFAIK.

jzheaux:
Related https://github.com/marketplace/actions/gradle-dependency-submission

marcusdacoregio:
@jzheaux I think this can be considered now that we have the `dependabot.yml` in place for Gradle dependencies.

marcusdacoregio:
Closing in favor of https://github.com/spring-projects/spring-security/issues/14298

Issue #11451, opened 2022-07-01
ServerJwtDslTests & WebTestClientHtmlUnitDriverBuilderTests tests failing with 404 issue


igorbolic:
**Describe the bug**
Running the Gradle build from master branch the `ServerJwtDslTests` & `WebTestClientHtmlUnitDriverBuilderTests` tests are failing when accessing root path in test cases.

The following test cases are failing:

**ServerJwtDslTests**  -> `request when JWT configured with public key and valid token then responds with ok()`

```
java.lang.AssertionError: Status expected:<200 OK> but was:<404 NOT_FOUND>
	at org.springframework.test.util.AssertionErrors.fail(AssertionErrors.java:59)
	at org.springframework.test.util.AssertionErrors.assertEquals(AssertionErrors.java:122)
	at org.springframework.test.web.reactive.server.StatusAssertions.lambda$assertStatusAndReturn$4(StatusAssertions.java:236)
	at org.springframework.test.web.reactive.server.ExchangeResult.assertWithDiagnostics(ExchangeResult.java:232)
	at org.springframework.test.web.reactive.server.StatusAssertions.assertStatusAndReturn(StatusAssertions.java:236)
	at org.springframework.test.web.reactive.server.StatusAssertions.isOk(StatusAssertions.java:68)
	at org.springframework.security.config.web.server.ServerJwtDslTests.request when JWT configured with public key and valid token then responds with ok(ServerJwtDslTests.kt:95)
```

**WebTestClientHtmlUnitDriverBuilderTests**  -> `helloWorld()`

```
java.lang.AssertionError: 
Expecting actual:
  ""
to contain:
  "Hello World" 
	at org.springframework.security.htmlunit.server.WebTestClientHtmlUnitDriverBuilderTests.helloWorld(WebTestClientHtmlUnitDriverBuilderTests.java:54)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
```

**To Reproduce**
Runt the `./gradlew clean build` from the master branch

**Expected behavior**
The test should pass in green light.

**Sample**

[...]

PR #11462, opened 2022-07-04
Set the useTrailingSlashMatch to true for tests

igorbolic:
The Spring MVC changed the default behavior for trailing slash match
with https://github.com/spring-projects/spring-framework/issues/28552.
This causes failures in Spring Security's tests.

Setting the `useTrailingSlashMatch` to `true` ensures that Spring
Security will work for users who have modified the default configuration.
Specifing the request mapper with trailing slash path ensures that the tests
are successful when default behavior is used.

Closes gh-11451


jzheaux:
Thanks, @igorbolic, this is now merged into `main`

PR #11464, opened 2022-07-05
Add SHA256 as an algorithm option for Remember Me token hashing

marcusdacoregio:
Closes gh-8549

### DIFF
--- a/docs/modules/ROOT/pages/servlet/authentication/rememberme.adoc
+++ b/docs/modules/ROOT/pages/servlet/authentication/rememberme.adoc
@@ -19,13 +19,14 @@ In essence, a cookie is sent to the browser upon successful interactive authenti
 ====
 [source,txt]
 ----
-base64(username + ":" + expirationTime + ":" +
-md5Hex(username + ":" + expirationTime + ":" password + ":" + key))
+base64(username + ":" + expirationTime + ":" + algorithmName + ":"
+algorithmHex(username + ":" + expirationTime + ":" password + ":" + key))
 
 username:          As identifiable to the UserDetailsService
 password:          That matches the one in the retrieved UserDetails
 expirationTime:    The date and time when the remember-me token expires, expressed in milliseconds
 key:               A private key to prevent modification of the remember-me token
+algorithmName:     The algorithm used to generate and to verify the remember-me token signature
 ----
 ====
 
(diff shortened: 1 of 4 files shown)


skjelmo (commented):
I love it that this PR offers the possibility to configure a secure hashing algorithm for the Remember Me token 🙌 

The possibility to use a secure hashing algorithm is way more important than the concern I'm raising here, but I assume this is the time and place to ask the following question.

I see from the [previous PR](https://github.com/spring-projects/spring-security/pull/10675) that there was a [suggestion to use setter ingestion](https://github.com/spring-projects/spring-security/pull/10675#discussion_r883961181) with the argument that the hashing algorithm is an optional parameter. 

I agree with the argument, and the parameter is obviously optional from the code's point of view. However, I'm not confident that the assumption holds when the default option is cryptographically insecure.

Would it not be better for users who needs a cryptographically secure hashing algorithm to have it be set as a final field with constructor injection, so that it could not be changed (downgraded to an insecure hashing algorithm) by a setter during runtime? Or is my concern way off? 

The configuration would still be optional without a [breaking change so that it doesn't have to wait until Spring Security 6](https://github.com/spring-projects/spring-security/pull/8580#pullrequestreview-416529809). 

Lastly, I really hope this PR makes it through the review process and into Spring Security. From what I can see, there has been [1](https://github.com/spring-projects/spring-security/pull/8580), [2](https://github.com/spring-projects/spring-security/pull/8591), [3](https://github.com/spring-projects/spring-security/pull/9392), [4](https://github.com/spring-projects/spring-security/pull/10675) PRs before this attempt to address [issue #8549 ](https://github.com/spring-projects/spring-security/issues/8549).

docs/modules/ROOT/pages/servlet/authentication/rememberme.adoc
```diff
@@ -113,6 +114,53 @@ A `key` is shared between this authentication provider and the `TokenBasedRememb
...
+To do that you can specify your customized `TokenBasedRememberMeServices` as a Bean and use it in the configuration.
+
+====
+.Java
+[source,java,role="primary"]
+----
+@Bean
+SecurityFilterChain securityFilterChain(HttpSecurity http, RememberMeServices rememberMeServices) throws Exception {
+	http
+			.authorizeRequests((authorize) -> authorize
```
jzheaux, line 130: While `authorizeRequests` is not yet deprecated, `authorizeHttpRequests` is preferred at this point. Is there a reason to use `authorizeRequests` in this part of the documentation?

web/src/main/java/org/springframework/security/web/authentication/rememberme/TokenBasedRememberMeServices.java
```diff
@@ -172,15 +217,33 @@ public void onLoginSuccess(HttpServletRequest request, HttpServletResponse respo
...
+				tokenLifetime, request, response);
 		if (this.logger.isDebugEnabled()) {
 			this.logger.debug(
 					"Added remember-me cookie for user '" + username + "', expiry: '" + new Date(expiryTime) + "'");
 		}
 	}
 
+	/**
+	 * Sets the algorithm to be used to encode the token signature
+	 * @param encodingAlgorithm the encoding algorithm
```
jzheaux, line 231: I think that https://github.com/spring-projects/spring-security/pull/11464#pullrequestreview-1031035009 makes a good point about adding this to the constructor. I see now also that this is how `DelegatingPasswordEncoder` is designed, with the encoding encoder being set in the constructor and the matching encoder being set as a setter.

web/src/main/java/org/springframework/security/web/authentication/rememberme/TokenBasedRememberMeServices.java
```diff
@@ -172,15 +217,33 @@ public void onLoginSuccess(HttpServletRequest request, HttpServletResponse respo
...
+	 */
+	public void setEncodingAlgorithm(RememberMeTokenAlgorithm encodingAlgorithm) {
+		Assert.notNull(encodingAlgorithm, "encodingAlgorithm cannot be null");
+		this.encodingAlgorithm = encodingAlgorithm;
+	}
+
+	/**
+	 * Sets the algorithm to be used to match the token signature
+	 * @param matchingAlgorithm the matching algorithm
+	 */
```
jzheaux, line 241: Please add `@since 5.8`

jzheaux (changes requested):
Great, @marcusdacoregio! I've left some inlined feedback.

docs/modules/ROOT/pages/servlet/authentication/rememberme.adoc
marcusdacoregio: It was an oversight, thanks for pointing that out.

marcusdacoregio:
Thanks @jzheaux and @skjelmo.

The PR is now updated with the suggested changes. Looking forward to your review again.

jzheaux (approved):

Issue #11468, opened 2022-07-06
Receive AuthnRequest Id and Response InResponseTo in Saml2AuthenticationRequestRepository


scottshidlovsky:
**Expected Behavior**
The ability to have access to AuthnRequest Id and Response InResponseTo when saving and loading AbstractSaml2AuthenticationRequests via Saml2AuthenticationRequestRepository

**Current Behavior**
When saving and loading the request via Saml2AuthenticationRequestRepository the user receives AbstractSaml2AuthenticationRequest and HTTP request and response. There is no easy way to access the AuthnRequest Id and the Response InResponseTo when saving and loading requests.

**Context**
When using spring session, the requests cannot be loaded using the default HttpSessionSaml2AuthenticationRequestRepository because the cookie using SameSite = Lax as described in https://github.com/spring-projects/spring-security/issues/10828. We are looking to create a custom Saml2AuthenticationRequestRepository that can store and load the Requests based on the ID instead of using the session. 

The workaround we are using for now is to create custom assertionValidator and responseValidator which filter out the errors for InResponseTo validation. 
```
openSamlAuthenticationProvider.setResponseValidator { token ->
            val result = OpenSaml4AuthenticationProvider.createDefaultResponseValidator().convert(token)
            val newResult = Saml2ResponseValidatorResult.success()
            result?.errors?.forEach { error ->
                if (error.errorCode != org.springframework.security.saml2.core.Saml2ErrorCodes.INVALID_IN_RESPONSE_TO) {
                    newResult.concat(error)
                }
            }
            newResult
        }
```

For retrieving the AuthnRequest ID we could also use similar code to OpenSaml4AuthenticationProvider#getAuthnRequestId but it would require our own implementation because it is currently private. 

Its also possible this is more appropriate to be a stack overflow question on how to get the ID and InResponseTo from a HttpServletRequest

PR #11489, opened 2022-07-12
Add ID to Saml2 Post and Redirect Requests

scottshidlovsky:
Closes gh-11468

### DIFF
--- a/saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/web/authentication/OpenSamlAuthenticationRequestResolverTests.java
+++ b/saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/web/authentication/OpenSamlAuthenticationRequestResolverTests.java
@@ -16,8 +16,8 @@
 
 package org.springframework.security.saml2.provider.service.web.authentication;
 
-import org.junit.Before;
-import org.junit.Test;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
 import org.opensaml.xmlsec.signature.support.SignatureConstants;
 
 import org.springframework.mock.web.MockHttpServletRequest;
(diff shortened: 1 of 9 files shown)


saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/web/authentication/OpenSamlAuthenticationRequestResolverTests.java
```diff
@@ -65,6 +65,7 @@ public void resolveAuthenticationRequestWhenSignedRedirectThenSignsAndRedirects(
 		assertThat(result.getSigAlg()).isEqualTo(SignatureConstants.ALGO_ID_SIGNATURE_RSA_SHA256);
 		assertThat(result.getSignature()).isNotEmpty();
 		assertThat(result.getBinding()).isEqualTo(Saml2MessageBinding.REDIRECT);
+		assertThat(result.getID()).isNotEmpty();
```
scottshidlovsky, line 68: These tests don't appear to be running. Intellij couldn't run them and when I added an exception to be thrown in the tests, there are no failures running with gradle. Is that something that should be looked into for this ticket or a new issue created?

jzheaux (changes requested):
Thanks, @scottshidlovsky! I've left some feedback inline.

saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/web/authentication/OpenSamlAuthenticationRequestResolverTests.java
scottshidlovsky: You were correct. I updated to use JUnit 5 and the tests are running now. There were two failing tests when I ran them with JUnit 5. ```resolveAuthenticationRequestWhenSignedThenCredentialIsRequired``` and ```resolveAuthenticationRequestWhenSHA1SignRequestThenSigns``` because ```OpenSamlAuthenticationRequestResolver#resolve``` does not check if authnRequestConsumer is null before calling it on line 132. I added an empty consumer to fix this issue.

scottshidlovsky:
> Thanks, @scottshidlovsky! I've left some feedback inline.

Thank you for the review! I believe I have addressed all your feedback.

jzheaux (approved):

Issue #11428, opened 2022-06-21
Deprecate AuthenticationProvider


jzheaux:
`AuthenticationManager` and `AuthenticationProvider` have the same primary signature. In an effort to simplify the API, `AuthenticationProvider` should be deprecated.

Here is an initial list:

- [ ] Add `DelegatingAuthenticationManager` to replace `ProviderManager`
- [ ] Update configuration to use `DelegatingAuthenticationManager`
- [ ] Create `AuthenticationManager` implementations for existing `AuthenticationProvider`s
- [ ] Consider creating `AuthenticationProviderManagerAdapter` to adapt `AuthenticationProvider`s into `AuthenticationManager`s
- [ ] Deprecate `authentication-provider` XML support
- [ ] Deprecate DSL support for `authenticationProvider()`
- [ ] Consider `@Bean` support for multiple `AuthenticationManager`s

jgrandja:
I agree that we should simplify the current design of `AuthenticationManager` and `AuthenticationProvider`, however, I'm not sure this makes sense:

> Create `AuthenticationManager` implementations for existing `AuthenticationProvider`s

Most (likely all) `AuthenticationProvider` implementations authenticate a specific credential (e.g. username/password, `authorization_code`, `Jwt`, etc.) and return the authenticated representation of it. However, they do not "manage" the credential or anything at all really. So I'm hung up on the naming here. I don't feel we should have an `AuthenticationManager` if it doesn't really manage anything.

A more logical name would be `Authenticator` since it authenticates a credential. The `ProviderManager` implementation could be re-implemented as a `DelegatingAuthenticator`.

I realize this is a much bigger change since we would introduce a new interface `Authenticator` and deprecate both `AuthenticationManager` and `AuthenticationProvider`. However, I don't feel we should introduce a whole new set of `AuthenticationManager` implementations.

Have we considered deprecating `AuthenticationManager` and keeping `AuthenticationProvider` but removing `supports()`?

jzheaux:
Interesting idea; I can see your point about the semantics around the word "manager".

Yes, it does sound like a big change. Some other things that come to mind are `AuthorizationManager`, `ReactiveAuthenticationManager`, and `ReactiveAuthorizationManager`. There may be others. If we change servlet authentication to `Authenticator`, it seems like we'd need to change these others, following the same semantic reasoning. 

[...]

PR #11595, opened 2022-07-18
Spring Security Observability for Authentication and Authorization

jzheaux:
By doing:

```java
@Bean 
ObservationRegistry observationRegistry() {
    ObservationRegistry observationRegistry = ObservationRegistry.create();
    observationRegistry.observationConfig().observationHandler(new ObservationTextPublisher());
    return observationRegistry;
}

@Bean 
ApplicationListener<SecurityEvent> securityEventListener(ObservationRegistry registry) {
    return DelegatingObservationSecurityEventListener.withDefaults(registry).build();
}
```

Then all `AuthenticationSuccessEvent`s, `AuthenticationFailureEvent`s, and `AuthorizationDeniedEvent`s will be published to Micrometer. `AuthorizationGrantedEvent`s are not published by default in Spring Security. The event details will follow a specific schema. Also, all other  `SecurityEvent`s will be published as well, though only by name. Potentially, Spring Boot could construct a default `DelegatingObservationSecurityEventListener` instance like the above.

*NOTE*: all security events must have a prevailing observation to attach to. For a servlet-based application, this is the Security Filter Chain observation. In addition to the Security Filter Chain observation, there is one dedicated to authentication and another to authorization.

If you want to use the authorization observation, you will need to use `AuthorizationFilter`, which is activated by using `authorizeHttpRequests`:

```java
@Bean
SecurityFilterChain app(HttpSecurity http) throws Exception {
	return http
		.httpBasic()
		.authorizeHttpRequests((request) -> request.anyRequest().authenticated())
		.build();
}
```

In that case, authorization events will be tied to the authorization observation.

Using the authentication observation is automatic as the default `AuthenticationManager` is wrapped.

*Note*: Micrometer's `ObservationTextPublisher` will log events to the console.


jgrandja:
@jzheaux I took a look at the code but I couldn't figure out how a user could hook into observability using a custom `SecurityEvent` that the application has defined.

Would you be able to put together a test so I can better understand?

The use case is this:
- An application has defined their own custom `SecurityEvent` called `OAuth2AuthorizationConsentDeniedEvent`. There is custom code they have hooked into for the authorization consent flow and are instantiating an `OAuth2AuthorizationConsentDeniedEvent` with all the relevant context.
- What do they call to publish the event?
- And what do they provide/configure to receive the event in order to build/format the appropriate message that will eventually be propagated to its source?
- How is the source configured? For example, standard logging vs. a custom hook that gives me complete flexibility to further propagate the event/message to whatever source I choose?

jzheaux:
@jgrandja, great questions. I will put together a test. In the meantime, here are some quick answers:

> What do they call to publish the event?

This PR does not add any publishing capabilities since it is based off of Spring's event-publishing infra. To publish an event, use Spring's `ApplicationEventPublisher`.

> And what do they provide/configure to receive the event ... in order to build/format

```java
Observation.KeyValuesProvider myKeyValuesProvider = new MyEventKeyValuesProvider();
ObservationSecurityEventListener<MyEvent> listener =
    new ObservationSecurityEventListener<>(observationRegistry, myKeyValuesProvider);
DelegatingObservationSecurityEventListener.withDefaults(observationRegistry)
    .add(ResolvableType.forClass(MyEvent.class), listener).build();
```

Briefly, in Micrometer there is such a thing as a `KeyValuesProvider`. When creating an event, you will need to also [create a key values provider](https://github.com/spring-projects/spring-security/pull/11595/files#diff-db964c535d24c2609f09803585638ae9baa6e6662aca9827e52a3e62803e100eR31) that specifies to Micrometer what the relevant parts of that event are.

There is no formatting at this step in the flow.

> How is the destination configured?

Micrometer has the concept of an `ObservationHandler` which takes care of shipping those observations to different destinations, like a log aggregator. You can see an example in the PR's description where it uses `ObservationTextPublisher`.

core/src/main/java/org/springframework/security/event/ObservationSecurityEventListener.java
```diff
@@ -0,0 +1,62 @@
...
+
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public void onApplicationEvent(T event) {
+		String name = "spring.security." + event.getEventType();
+		Observation.Context context = SecurityEventObservationContext.fromEvent(event);
+		Observation.createNotStarted(name, context, this.registry).keyValuesProvider(this.keyValuesProvider)
+				.observe(() -> {
```
jonatan-ivanov, line 58: What is the goal here?
What this will do is measuring the empty code block.

I'm just guessing but I think here signaling that something happened on an existing Observation would make more sense. E.g.:
somebody starts an Observation outside of Spring Security (e.g.: and HTTP request was received) and here we can signal that something happened during that observation:
```java
Observation observation = observationRegistry.getCurrentObservation();
if (observation != null) {
    observation.event(Observation.Event.of("something.happened"));
}
```

jonatan-ivanov: I think measuring a code block is also a scenario that makes sense for Spring Security, e.g.: instrumenting `AuthenticationProvider` or any component that might need time to do its job, e.g.: using BCrypt, calling out on the network (http, ldap, jdbc, etc), reading files, etc.

core/src/main/java/org/springframework/security/authentication/DelegatingAuthenticationManager.java
```diff
@@ -0,0 +1,160 @@
...
+	public DelegatingAuthenticationManager(List<AuthenticationManager> delegates) {
+		Assert.notEmpty(delegates, "delegates cannot be null or empty");
+		this.delegates = delegates;
+	}
+
+	@Override
+	public Authentication authenticate(Authentication authentication) throws AuthenticationException {
+		ObservationContext context = new ObservationContext(authentication);
+		Observation observation = Observation.createNotStarted("spring.security.authentication", context, this.registry)
+				.keyValuesProvider(this.keyValuesProvider);
```
jonatan-ivanov, line 49: The Observation needs to be started.
```suggestion
		Observation observation = Observation.createNotStarted("spring.security.authentication", context, this.registry)
				.keyValuesProvider(this.keyValuesProvider)
				.start();
```

core/src/main/java/org/springframework/security/authentication/DelegatingAuthenticationManager.java
```diff
@@ -0,0 +1,160 @@
...
+			return authenticate(context, authentication);
+		}
+		catch (AuthenticationException ex) {
+			this.eventPublisher.publishAuthenticationFailure(ex, authentication);
+			observation.error(ex);
+			throw ex;
+		}
+		finally {
+			observation.stop();
+		}
```
jonatan-ivanov, line 60: I'm not sure how sensitive the order is here, but you might be able to spare some boilerplate by doing something like this:

```suggestion
		try {
		    return observation.observeChecked(() -> authenticate(context, authentication));
		}
		catch (AuthenticationException ex) {
			this.eventPublisher.publishAuthenticationFailure(ex, authentication);
			throw ex;
		}
```

In this case you don't need to start the Observation so please ignore my comment above about start and `publishAuthenticationFailure` will be called after the observation was stopped (this also means that it is not measured). This seems fine to me though. 

Or you can move the error handling to the `authenticate` method.

core/src/main/java/org/springframework/security/authentication/DelegatingAuthenticationManager.java
```diff
@@ -0,0 +1,160 @@
...
+
+	}
+
+	static class KeyValuesProvider implements Observation.KeyValuesProvider<ObservationContext> {
+
+		@Override
+		public KeyValues getLowCardinalityKeyValues(ObservationContext context) {
+			Authentication request = context.getAuthenticationRequest();
+			KeyValues kvs = KeyValues.of("authentication.request.type", request.getAuthenticationType());
+			if (context.getAuthenticationResult() != null) {
```
jonatan-ivanov, line 111: Can we add fallback values if there is no result?
```java
else {
    return kvs.and("authentication.result.type", "n/a").and("authentication.method", "n/a");
}
```

core/src/main/java/org/springframework/security/event/ObservationSecurityEventListener.java
jzheaux: Excellent! Glad to hear that feature is available now.

jzheaux: This is a place where the PR is in flux right now, but because of existing [plans to deprecate `AuthenticationProvider`](https://github.com/spring-projects/spring-security/issues/11428), instrumentation will likely go into [an `AuthenticationManager` implementation](https://github.com/spring-projects/spring-security/pull/11595/files#diff-4f74389b3bef57442d5aa75201d3647ba476b992d66300eecfa74e575bde3629R30) instead of into `AuthenticationProvider`s.

jzheaux:
@jgrandja based on some changes in Observability M4, I need to amend a few of my statements.

> And what do they provide/configure to receive the event ... in order to build/format

First, if it extends `SecurityEvent`, you need do nothing. The event's name will be published to Micrometer by default.

Second, if you have event metadata that you want to transmit, then you can provide a strategy that computes that metadata for a given event, like so:

```java
@Bean 
ApplicationListener<SecurityEvent> securityEventListener(ObservationRegistry registry) {
    Builder builder = DelegatingObservationSecurityEventListener.withDefaults(registry);

    // specify the custom class and the key values strategy
    builder.add(MyEvent.class, (event) -> KeyValues.of("my.key", event.getValue()));

    return builder.build();
}
```

There is no formatting at this step in the flow.

core/src/main/java/org/springframework/security/event/AuthenticationFailureEventKeyValuesProvider.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * A {@link io.micrometer.observation.Observation.KeyValuesProvider} for
+ * {@link org.springframework.security.access.event.AuthorizationFailureEvent}s
+ *
+ * @author Josh Cummings
+ * @since 6.0
+ */
+public final class AuthenticationFailureEventKeyValuesProvider
+		implements Observation.KeyValuesProvider<SecurityEventObservationContext<AbstractAuthenticationFailureEvent>> {
```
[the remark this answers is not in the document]
jzheaux: Sounds great. I updated the authentication observation to use `ObservationConvention`. As for the event handling, I think since I'm now calling `observation.event`, these classes don't make as much sense anymore as-is.

Please see `ObservationSecurityEventListener` for details. Essentially, an observation security event can hold a `KeyValues` instance to indicate which are the relevant attributes of the underlying `SecurityEvent` to publish.

core/src/main/java/org/springframework/security/event/ObservationSecurityEventListener.java
jzheaux: Changed to `AuthenticationFilter`, which all the major authentication mechanisms will change over to in the 6.0 release.

jzheaux:
Closed in favor of https://github.com/spring-projects/spring-security/pull/11906

Issue #10957, opened 2022-03-11
Kotlin DSL authorize cannot default access string when servletPath is set


eleftherias:
The Kotlin `AuthorizeRequestsDsl` has multiple signatures for the `authorize` method including the following two:
```kotlin
authorize(pattern: String, access: String = "authenticated")
```
and 
```kotlin
authorize(pattern: String, servletPath: String, access: String = "authenticated")
```

Although the `access` string is defaulted in the second method, users cannot take advantage of the default value, because supplying 2 strings will always result in the first method being evaluated.

For example, `authorize("/path", "x")` is equivalent to `authorize(pattern = "/path", access = "x")` and not `authorize(pattern = "/path", servletPath = "x")`.

PR #11599, opened 2022-07-20
Remove default value for access parameter

ysavchen:
Closes gh-10957

The test below shows the issue with "authenticated" access as a default parameter.
After fix the test is not relevant as the problem is solved at the compilation phase.
This is a breaking change, for example, when a user has `authorize("/path")` instead of `authorize("/path", authenticated)` in config.
```
@Test
fun `request when authorize with servletPath and default access`() {
    this.spring.register(ServletPathAndDefaultAccessConfig::class.java).autowire()

    this.mockMvc.perform(MockMvcRequestBuilders.get("/servletPath/path")
        .with { request ->
            request.servletPath = "/servletPath"
            request
        })
        .andExpect(status().isUnauthorized)

    this.mockMvc.perform(MockMvcRequestBuilders.get("/servletPath/path")
        .with(httpBasic("user", "password"))
        .with { request ->
            request.servletPath = "/servletPath"
            request
        })
        .andExpect(status().isOk)
}

@EnableWebSecurity
@EnableWebMvc
open class ServletPathAndDefaultAccessConfig {
    @Bean
    open fun securityFilterChain(http: HttpSecurity): SecurityFilterChain {
        http {
            authorizeRequests {
                // before fix:
                // a user intends to invoke authorize(pattern, servletPath, access ["authenticated" by default]),
                // but in fact get authorize(pattern, access)
                authorize("/path", "/servletPath")

                // after fix:
                // a user must explicitly specify "authenticated" to invoke authorize(pattern, servletPath, access)
                authorize("/path", "/servletPath", authenticated)
            }
            httpBasic { }
        }
        return http.build()
    }

    @RestController
    internal class PathController {
        @GetMapping("/path")
        fun path() {
        }
    }

    @Bean
    open fun userDetailsService(): UserDetailsService {
        val userDetails = User.withDefaultPasswordEncoder()
            .username("user")
            .password("password")
            .roles("USER")
            .build()
        return InMemoryUserDetailsManager(userDetails)
    }
}
```

### DIFF
--- a/config/src/main/kotlin/org/springframework/security/config/annotation/web/AuthorizeRequestsDsl.kt
+++ b/config/src/main/kotlin/org/springframework/security/config/annotation/web/AuthorizeRequestsDsl.kt
@@ -46,7 +46,7 @@ class AuthorizeRequestsDsl : AbstractRequestMatcherDsl() {
      * (i.e. "hasAuthority('ROLE_USER') and hasAuthority('ROLE_SUPER')")
      */
     fun authorize(matches: RequestMatcher = AnyRequestMatcher.INSTANCE,
-                  access: String = "authenticated") {
+                  access: String) {
         authorizationRules.add(MatcherAuthorizationRule(matches, access))
     }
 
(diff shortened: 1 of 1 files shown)


marcusdacoregio:
Thanks @ysavchen, this is now merged into `main`.

Issue #11602, opened 2022-07-20
Ignored requestCache.saveRequest, because sendRedirect is in front


o1i0:
**Describe the bug**
I am using CookieRequestCache .
However, as SendRedirect proceeds first, cookie settings are not included.

**To Reproduce**

**Expected behavior**


**Sample**

https://github.com/spring-projects/spring-security/blob/bf138c5154e87b55ab6f1c8940116351d6e1aedc/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationRequestRedirectFilter.java#L189-L190

jzheaux:
@o1i0, thanks for the report. In your PR, will you please include a unit test that behaves the way you are expecting? It should fail without your change and pass with it.

Including a unit test will give me a better idea of your use case to see if there is a better way to address it.

o1i0:
@jzheaux,
Thanks for the quick check.

Added a test case where the response should be requestCache.saveRequest before isCommitted.
https://github.com/spring-projects/spring-security/pull/11603/files#diff-09f439ada4e90ffa29137b421089e9fcf474193dcef6fb5ee15e78e5c2ed2e20R195-R196

Please check.

PR #11603, opened 2022-07-20
Move sendRedirect after requestCache.saveRequest

o1i0:
gh-11602


oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationRequestRedirectFilterTests.java
```diff
@@ -191,6 +192,8 @@ public void doFilterWhenNotAuthorizationRequestAndClientAuthorizationRequiredExc
 		request.setServletPath(requestUri);
 		MockHttpServletResponse response = new MockHttpServletResponse();
 		FilterChain filterChain = mock(FilterChain.class);
+		willAnswer((invocation) -> assertThat((invocation.<HttpServletResponse>getArgument(1)).isCommitted()).isFalse())
```
jzheaux, line 195: Will you please move this assertion into its own test and then add a comment to the top of that test that looks like this:

```java
// gh-11602
```

jzheaux (changes requested):
Thanks for the PR, @o1i0! I've left my feedback inline.

Also, will you please squash your commits and ensure that the phrase `Closes gh-11602` is at the bottom of the commit message?

jzheaux:
Thanks, @o1i0! This is now merged into `5.6.x`, `5.7.x`, `5.8.x`, `6.0.x`, and `main`.

Issue #11626, opened 2022-07-26
Keep Webflux Security Documentation consistent


jsattler:
**Expected Behavior**
[Webflux Security Documentation](https://github.com/spring-projects/spring-security/blob/main/docs/modules/ROOT/pages/reactive/configuration/webflux.adoc) should be consistent with the examples and only use `@EnableWebFluxSecurity` without the additional `@Configuration`.

**Current Behavior**
One example only uses `@EnableWebFluxSecurity` the rest use an additional `@Configuration`. 

**Context**
It makes it harder for the reader to understand if the `@Configuration` is needed or not.
I suggest to keep it consistent and remove the `@Configuration` with a hint that `@EnableWebfluxSecurity` is already annotated with `@Configuration`.

jsattler:
#11627

sjohnr:
Hi @jsattler. Apologies, I thought too quickly to merge your PR but went ahead and reverted it, due to gh-6613. However, I agree with you that it can be (and is) confusing to users. Because the plan is to *remove* the meta-annotation, I believe we will actually be reversing this throughout our documentation and adding the `@Configuration` to the examples in order to be consistent with the rest of Spring. With that in mind, I'm going to close this issue and add a note on gh-6613 to update the documentation.

If you're interested in working on that issue, feel free to add a comment over there as we'd love to have your help on it! Totally optional of course.

PR #11627, opened 2022-07-26
Remove @Configuration from webflux config examples

jsattler:
Remove `@Configuration` from Webflux Security Configuration to be consistent with other examples.

#11626


sjohnr:
Reverted via a72c5a55db787509b808e75fc7b73916f8ab4a12

Issue #11621, opened 2022-07-26
RemoteJwkSet is not refreshed when encountering an unknown KID


tinolazreg:
**Describe the bug**
If you are setting up a NimbusJwtDecoder using Spring Security, we have the possibility to provide a Cache that will be used for storing the JWK Set. This is a good feature, since we can then override the default TTL, instead of using DefaultJWKSetCache, which has a default refresh time of 5 minutes.

Reference to documentation: https://docs.spring.io/spring-security/reference/servlet/oauth2/resource-server/jwt.html#oauth2resourceserver-jwt-timeouts

The issue occurs when you use your own cache implementation, and you try to decode a JWT with an unknown KID. I would expect this to trigger a refresh of the JWK set, but this is not what is happening.

Example code of setup:
```java
        var jwkSetCache = new ConcurrentMapCache("jwkSetCache", CacheBuilder.newBuilder()
                .expireAfterWrite(Duration.ofMinutes(30))
                .build().asMap(), false);
        var decoder = NimbusJwtDecoder.withJwkSetUri(jwkSetUri)
                .cache(jwkSetCache)
                .build();
```

We end up here with an unknown KID: https://bitbucket.org/connect2id/nimbus-jose-jwt/src/43b5435c548d70ce250afe6f932a6a55b3833bb8/src/main/java/com/nimbusds/jose/jwk/source/RemoteJWKSet.java#lines-486, and we expect the JWK set to be refetched.

The problem is that the NimbusJwtDecoder.java always uses the CachingResourceRetriever, even when it encounters an unknown KID: https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java#L406-L407. This will lead to the JWK set only being fetched from the cache, and not refetched from the remote source.

**To Reproduce**
1. Setup NimbusJwtDecoder with a Cache and remote JWK set.
2. Try to decode a JWT signed with a KID not present in the JWK set.
3. Expect NimbusJwtDecoder to refetch JWK set from jwksetUri.

If we don't specify a cache to the NimbusJwtDecoder builder, it works as expected. The jwk set is refetched when encountering a signed JWS with an unknown KID. 

**Expected behavior**
I expect JWK set to be refetched when encountering a KID not present in the current cached JWK set.

sjohnr:
Thanks for reaching out @tinolazreg, and thanks for the detailed explanation!

[...]

PR #11638, opened 2022-07-27
Refresh remote JWK when unknown KID error occurs

sjohnr:
Closes gh-11621

### DIFF
--- a/oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
+++ b/oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
@@ -36,12 +36,16 @@ import java.util.concurrent.Callable;
 
 import javax.crypto.SecretKey;
 
+import com.nimbusds.jose.JOSEException;
 import com.nimbusds.jose.JOSEObjectType;
 import com.nimbusds.jose.JWSAlgorithm;
 import com.nimbusds.jose.JWSHeader;
 import com.nimbusds.jose.JWSSigner;
 import com.nimbusds.jose.crypto.MACSigner;
 import com.nimbusds.jose.crypto.RSASSASigner;
+import com.nimbusds.jose.jwk.JWKSet;
+import com.nimbusds.jose.jwk.RSAKey;
+import com.nimbusds.jose.jwk.gen.RSAKeyGenerator;
 import com.nimbusds.jose.jwk.source.JWKSource;
 import com.nimbusds.jose.proc.BadJOSEException;
 import com.nimbusds.jose.proc.DefaultJOSEObjectTypeVerifier;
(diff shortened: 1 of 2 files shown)


oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
```diff
@@ -644,22 +648,100 @@ public void decodeWhenCacheThenStoreRetrievedJwkSetToCache() {
...
+		assertThat(requestEntityCaptor.getValue().getHeaders().getAccept()).contains(MediaType.APPLICATION_JSON,
+				APPLICATION_JWK_SET_JSON);
+	}
+
+	// gh-11621
+	@Test
+	public void decodeWithoutCacheSpecifiedAndUnknownKidShouldTriggerFetchOfJwkSet() throws JOSEException {
+		RestOperations restOperations = mock(RestOperations.class);
+
+		RSAKey rsaJWK = new RSAKeyGenerator(2048).keyID("new_kid").generate();
```
jzheaux, line 712: Let's please not regenerate the kid on each run of the test. Ideally, the JWK and associated JWT will be constants at the top of the test. Same for the other test.

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
```diff
@@ -678,6 +760,22 @@ public void decodeWhenCacheIsConfiguredAndValueLoaderErrorsThenThrowsJwtExceptio
 		// @formatter:on
 	}
 
+	// gh-11621
+	@Test
+	public void decodeWhenCacheIsConfiguredAndParseFailsOnCachedValueThenThrowsJwtException() {
+		Cache cache = new ConcurrentMapCache("test-jwk-set-cache");
+		cache.put(JWK_SET_URI, "{}");
```
jzheaux, line 767: I wonder why the code would add a malformed JWK Set to the cache in the first place. Is there a way to instead ensure that what goes into the cache is well-formed?

jzheaux (changes requested):
Nice, @sjohnr! I've left some feedback inline.

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
sjohnr: Good question. This is one of those code-coverage for its own sake tests. Though technically it could happen outside of this component, I don't believe it's possible to happen normally. Shall I just remove the test?

jzheaux: I think the test that's needed is to ensure that what goes into the cache is well-formed.

Issue #6613, opened 2019-03-14
Remove @Configuration meta-annotation from @Enable annotations


vpavic:
Currently, all Spring Security's `@Enable` annotations are meta-annotated with `@Configuration`. While convenient, this is not consistent with the rest of the Spring projects and most notably Spring Framework's `@Enable` annotations.

Additionally, the introduction of support for `@Configuration(proxyBeanMethods=false)` in Spring Framework provides a compelling reason to remove `@Configuration` meta-annotation from Spring Security's `@Enable` annotations and allow users to opt into their preferred configuration mode.

sjohnr:
Note that we will want to update the documentation examples to include `@Configuration` in places where it's currently missing.

jsattler:
@sjohnr I would like to help on this, as I was anyhow looking into the documentation part.

sjohnr:
It's yours @jsattler! Keep in mind we'll want to update the reference documentation, javadoc and [samples](https://github.com/spring-projects/spring-security-samples). I'll open a ticket to update the samples, and if you don't feel you can tackle that part just let me know.

jsattler:
@sjohnr I will take care of the reference documentation, javadoc and samples. The actual removal of the `@Configuration` from the `@Enable*` annotations will be tackled in a separate PR, correct?

sjohnr:
Yes, please do update tests if necessary! I think it can be done under this issue, but if you would like a separate issue let me know. And yes, this is a breaking change so I believe it will only be done in 6.0 (`main`). @rwinch can correct me if I'm wrong.

jsattler:
Quite some files changed, hope that's okay for a single PR. If you have any suggestions to split this up, please let me know.

rwinch:
A note for self. I used the following to check for missing `@Configuration`:

```python
import re
import sys


enable_regex = r".*@Enable[a-zA-Z]+(Security|Authentication).*"
config_regex = r".*@Configuration.*"
A = 2
B = 2

file_name = sys.argv[1]
try:
	file = open(file_name, 'r')
	lines = file.readlines()
except Exception as err:
	print ("Could not open file:", file_name, repr(err))
	sys.exit()


def find_regex_in_range(lines, regex, range):
	for index in range:
		peek_line = lines[index - 1]
		if (re.match(regex, peek_line)):
			return True
	return False

def print_range(lines, range):
	for index in range:
		peek_line = lines[index - 1]
		print(f"{index} {peek_line}", end = "")

[...]

PR #11653, opened 2022-07-30
Remove Configuration meta-annotation from Enable* annotations

jsattler:
Before, Spring Security's @Enable* annotations were meta-annotated with @Configuration.
While convenient, this is not consistent with the rest of the Spring projects and most notably
Spring Framework's @Enable annotations. Additionally, the introduction of support for
@Configuration(proxyBeanMethods=false) in Spring Framework provides a compelling reason to
remove @Configuration meta-annotation from Spring Security's @Enable annotations and allow
users to opt into their preferred configuration mode.

Closes gh-6613

Signed-off-by: Joshua Sattler <joshua.sattler@mailbox.org>


rwinch:
Thanks for the PR @jsattler This is merged into main with a little clean up from my side a5069d7e35112dffa7b6a8eb255fd7189770654e

Issue #11510, opened 2022-07-14
Add isFullyAuthenticated to AuthenticatedTrustResolver


jzheaux:
Since `isFullyAuthenticated` is a composition of `isAnonymous` and `isRememberMe`, a default convenience method can be added:

```java
default boolean isFullyAuthenticated(Authentication authentication) {
    return !isAnonymous(authentication) && !isRememberMe(authentication)
}
```

This would be nice for `SecurityExpressionRoot#isFullyAuthenticated` and [`AuthenticatedAuthorizationManager#fullyAuthenticated`](https://github.com/spring-projects/spring-security/pull/11495)

karthikeyan-r:
Can i work on this implementation ?

jzheaux:
Yep! The issue is yours.

PR #11654, opened 2022-07-30
#11510: Add isFullyAuthenticated to AuthenticatedTrustResolver

karthikeyan-r:
Spring Security [11510](https://github.com/spring-projects/spring-security/issues/11510) - Implemented isFullyAuthenticated function which is composition of isAnonymous & isRememberMe

### DIFF
--- a/core/src/main/java/org/springframework/security/authentication/AuthenticationTrustResolver.java
+++ b/core/src/main/java/org/springframework/security/authentication/AuthenticationTrustResolver.java
@@ -54,16 +54,20 @@ public interface AuthenticationTrustResolver {
 	boolean isRememberMe(Authentication authentication);
 
 	/**
-	 * Indicates whether the passed <code>Authentication</code> token represents an anonymous user that
-	 * has been remembered (i.e. user that has been fully authenticated). This is a composition of 
-	 * <code>isAnonymous</code> and <code>isRememberMe</code> implementation
+	 * Indicates whether the passed <code>Authentication</code> token represents a fully
+	 * authenticated user (that is, neither anonymous or remember-me). This is a
+	 * composition of <code>isAnonymous</code> and <code>isRememberMe</code>
+	 * implementation
 	 * <p>
 	 * @param authentication to test (may be <code>null</code> in which case the method
 	 * will always return <code>false</code>)
-	 * @return <code>true</code> the passed authentication token represented an anonymous principal & is
-	 * authenticated using a remember-me token, <code>false</code> otherwise
+	 * @return <code>true</code> the passed authentication token represented an anonymous
+	 * principal & is authenticated using a remember-me token, <code>false</code>
+	 * otherwise
+	 * @since 5.8
 	 */
 	default boolean isFullyAuthenticated(Authentication authentication) {
-	    return !isAnonymous(authentication) && !isRememberMe(authentication)
+		return !isAnonymous(authentication) && !isRememberMe(authentication);
 	}
+
 }
--- a/core/src/main/java/org/springframework/security/authentication/AuthenticationTrustResolver.java
+++ b/core/src/main/java/org/springframework/security/authentication/AuthenticationTrustResolver.java
@@ -53,4 +53,17 @@ public interface AuthenticationTrustResolver {
 	 */
 	boolean isRememberMe(Authentication authentication);
 
+	/**
+	 * Indicates whether the passed <code>Authentication</code> token represents an anonymous user that
+	 * has been remembered (i.e. user that has been fully authenticated). This is a composition of 
+	 * <code>isAnonymous</code> and <code>isRememberMe</code> implementation
+	 * <p>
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/springframework/security/authentication/AuthenticationTrustResolver.java
```diff
@@ -53,4 +53,17 @@ public interface AuthenticationTrustResolver {
...
 
+	/**
+	 * Indicates whether the passed <code>Authentication</code> token represents an anonymous user that
+	 * has been remembered (i.e. user that has been fully authenticated). This is a composition of 
+	 * <code>isAnonymous</code> and <code>isRememberMe</code> implementation
+	 * <p>
+	 * @param authentication to test (may be <code>null</code> in which case the method
+	 * will always return <code>false</code>)
+	 * @return <code>true</code> the passed authentication token represented an anonymous principal & is
+	 * authenticated using a remember-me token, <code>false</code> otherwise
```
jzheaux, line 64: Will you please add `@since 5.8`? Remember to build with `./gradlew format` to ensure it is placed in the correct location in the JavaDoc.

jzheaux (changes requested):
Thanks, @karthikeyan-r! Sorry for the delay, I've left some feedback inline.

Will you please also adjust the commit message to be like the following:

```
Commit Title

Closes gh-11510
```

core/src/main/java/org/springframework/security/authentication/AuthenticationTrustResolver.java
```diff
@@ -53,4 +53,17 @@ public interface AuthenticationTrustResolver {
 	 */
 	boolean isRememberMe(Authentication authentication);
 
+	/**
+	 * Indicates whether the passed <code>Authentication</code> token represents an anonymous user that
+	 * has been remembered (i.e. user that has been fully authenticated). This is a composition of
```
jzheaux, line 58: I don't think that this is true. The logic confirms that this user is *not* anonymous and is *not* remember me. Can we please adjust the wording?

Maybe:

```
Indicates whether the passed {@code Authentication} token represents a fully authenticated user (that is, neither anonymous or remember-me).
```

jzheaux:
@karthikeyan-r, please also take a look at the ticket description to find the places to use this new method. Please change those places to use it.

jzheaux:
Hi, @karthikeyan-r. Are you able to make the requested changes?

karthikeyan-r:
Apologies for the delay @jzheaux, I was travelling & couldnt check and respond to your message. I will work on these comment & push code back in a day.

core/src/main/java/org/springframework/security/authentication/AuthenticationTrustResolver.java
karthikeyan-r: added `@since` annotation and formatted using `gradlew` task

core/src/main/java/org/springframework/security/authentication/AuthenticationTrustResolver.java
karthikeyan-r: updated the comment

karthikeyan-r:
Regarding changes in `AuthenticatedAuthorizationManager#fullyAuthenticated` as mentioned in this [issue](https://github.com/spring-projects/spring-security/issues/11510#issue-1305070202),

`FullyAuthenticatedAuthorizationStrategy` is using composition of `isGranted` and `isRememberMe` - not _**`isAnonymous`**_. Please clarify on this.
```
private static final class FullyAuthenticatedAuthorizationStrategy extends AuthenticatedAuthorizationStrategy {

		@Override
		boolean isGranted(Authentication authentication) {
			return super.isGranted(authentication) && !this.trustResolver.isRememberMe(authentication);
		}

	}
```

jzheaux:
Now it's my turn to delay, @karthikeyan-r! Thanks for your question.

If you follow the method `isGranted` to its implementation, you'll see that it is called `isAnonymous` also.

I believe it is correct to change the implementation of `FullyAuthenticatedAuthorizationStrategy` to do:

```java
return this.trustResolver.isFullyAuthenticated(authentication);
```

karthikeyan-r:
Thanks for your suggestion @jzheaux, I have updated implementation

jzheaux:
Thanks, @karthikeyan-r! It looks like there is a failing test. Can you check that out, please?

karthikeyan-r:
@jzheaux resolved that test failure. Added null check for authentication to fix that.

karthikeyan-r:
@jzheaux any comments on this fix ?

jzheaux:
Thanks for the updates, @karthikeyan-r! I'll take things from here. We won't be able to merge it just yet since we are in the RC phase of the next release, but I'll merge it right after that.

jzheaux:
Thanks, @karthikeyan-r! This is now merged into `main` with 5fcbb9f4ed792ca24836efaad37d5454a7dd881d and will appear in the `6.1.0` release.

PR #11704, opened 2022-08-15
pom.xml to build.gradle

jujunchen:
When I read the document, I found that there was an error in the gradle part. 
The wrong word pom.xml was used. It should be build.gradle


sjohnr:
Duplicate of gh-11705.

PR #11705, opened 2022-08-15
pom.xml to build.gradle

jujunchen:
When I read the document, I found that there was an error in the gradle part.
The wrong word pom.xml was used. It should be build.gradle


jujunchen:
> Thanks @jujunchen. It seems the `<dependencyManagement>` should be changed as well. Do you agree? If so, would you be able to make that change as well and squash your commits?

ok

PR #11713, opened 2022-08-16
Fix IP address parse error message in IpAddressMatcher#parseAddress()

cyb3r4nt:
There is no whitespace between error message and IP address value  `IpAddressMatcher#parseAddress()`
If IP value is wrong, then error text looks like `Failed to parse addressi.am.ip`.
There should be some separator between those two text tokens.

Also wrapped the address value with single quotes.
Will this add any confusion for the caller?
Or colon and `"Failed to parse address: $value` looks better?


jzheaux:
Thanks, @cyb3r4nt! This is now merged into `main`.

Issue #11716, opened 2022-08-16
webflux logout not working when project defines a context path (spring.webflux.base-path)


vanwobe:
**Describe the bug**
GET request to http://localhost:8080/path/logout provides a confirmation page.
After confirming the logout there is a post request to the root path resulting in a 404 error.

**To Reproduce**
take the hello-security project: https://docs.spring.io/spring-security/reference/reactive/getting-started.html
and add to the application.properties:
spring.webflux.base-path=/path

**Expected behavior**
after confirming the logout a POST request is expected for:
http://localhost:8080/path/logout
instead we see a POST to:
http://localhost:8080/logout

**Sample**
https://start.spring.io/starter.zip?type=maven-project&language=java&packaging=jar&jvmVersion=1.8&groupId=example&artifactId=hello-security&name=hello-security&description=Hello%20Security&packageName=example.hello-security&dependencies=webflux,security

vanwobe:
This is not an enhancement but really a bug. The logout is defective in case a context path is used.

PR #11717, opened 2022-08-16
LogoutPageGeneratingWebFilter should honor context path

vanwobe:
fixes [gh-11716](https://github.com/spring-projects/spring-security/issues/11716)

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/AuthenticationConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/AuthenticationConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


web/src/test/java/org/springframework/security/web/server/ui/LogoutPageGeneratingWebFilterTests.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
```
jzheaux, line 2: Will you please use `2022` for the end date?

web/src/main/java/org/springframework/security/web/server/ui/LogoutPageGeneratingWebFilter.java
```diff
@@ -58,14 +58,15 @@ private Mono<Void> render(ServerWebExchange exchange) {
```
jzheaux, line 58: Relatedly, will you please update the copyright header for this file to `2022`?

jzheaux (changes requested):
Thanks, @vanwobe! I've given some inline feedback. In addition, will you please adjust the commit to include an additional like like this:

```
Closes gh-11716
```

this helps with our GitHub integration.

jzheaux:
Thanks for the updates, @vanwobe. Instead of adding the closes message to a separate commit, will you please squash your commits so that the one commit has both the explanation of what it does as well as the closes message?

You can see an example in [our contribution guidelines](https://github.com/spring-projects/spring-security/blob/main/CONTRIBUTING.adoc#format-commit-messages).

jzheaux:
@vanwobe, I'm seeing several more commits. Would you like a hand at getting the number of commits down to one?

Issue #10552, opened 2021-11-26
AuthenticationEventPublisher is not injected in AuthenticationManagerBuilder for the HttpSecurity prototype built in HttpSecurityConfiguration


zhenchuan9r:
When building a SecurityFilterChain from HttpSecurity, a prototype is built in HttpSecurityConfiguration. However, this prototype is not aware of AuthenticationEventPublisher in the bean factory. Although it is possible to set AuthenticationEventPublisher later, I think it is inconvenient and is not an expected behaviour. I would like to know if this pattern is by design or is a place for an improvement.

zhenchuan9r:
The HttpSecurity prototype is by default installed with a null AuthenticationEventPublisher.

eleftherias:
Thanks for starting this discussion @zhenchuan9r.

You mentioned this in the original post, but for anyone else arriving at this issue, I will point out that you can configure an `AuthenticationEventPublisher` bean yourself:

```
@Bean
public AuthenticationEventPublisher authenticationEventPublisher
		(ApplicationEventPublisher applicationEventPublisher) {
	return new DefaultAuthenticationEventPublisher(applicationEventPublisher);
}
```

When using the `WebSecurityConfigurerAdapter`, it will create an `AuthenticationEventPublisher` bean for you. 
This seems to me like a side-effect and I'm not sure if we would want to make this the default behaviour in `HttpSecurityConfiguration`.
I will have a discussion with the rest of the team and update this issue with additional details.

sjohnr:
For additional context on this issue, see [this comment](https://github.com/spring-projects/spring-security/issues/11162#issue-1214211098) by @Kehrlann and this [sample repo](https://github.com/Kehrlann/spring-security-issue-securityfilterchain-authenticationeventpublisher) that demonstrates the issue.

I think we could make the upgrade path to using an `@Bean` instead of `WebSecurityConfigurerAdapter` easier if we add consistency here.

After a bit of discussion, here are some notes:

* `HttpSecurityConfiguration.httpSecurity()` is (more or less) the equivalent to `WebSecurityConfigurerAdapter.getHttp()`
* `WebSecurityConfigurerAdapter` wires an `AuthenticationEventPublisher` into the local `AuthenticationManagerBuilder` while `HttpSecurityConfiguration` does not
* `WebSecurityConfigurerAdapter` will create the `AuthenticationEventPublisher` if it does not exist as a bean
[...]

PR #11719, opened 2022-08-17
Configure event publisher for HttpSecurity beans

sjohnr:
Closes gh-10552


sjohnr:
Fixed via 0aac515737e385feca81286a6f0bfd1ef2bb90ab

PR #11720, opened 2022-08-18
 issue 3065 / SEC-2839 - fix 

koosg:
Fix issue 3065 / SEC-2839: an old bug where it was possible to parse when the init did not succeed completely. Essentially, it was forgotten to refetch the parser after reloading the parsers.


jzheaux:
Hi, @koosg. Thanks for putting together this PR. There seems to be something wrong with it, though, as it suggests that you'd like to modify several hundred files and merge several hundred commits.

I think this is a change that is needed for all supported branches. Would you please try submitting your PR again, making the change on the `5.6.x` branch? The PR should only contain your changes and have only one commit.

koosg:
Will do!

Op vr 26 aug. 2022 20:17 schreef Josh Cummings ***@***.***>:

koosg:
changed target branch, should be the 5.8

jzheaux:
Thanks, @koosg! This is now merged into `5.6.x`, `5.7.x`, `5.8.x`, and `main`

Issue #11625, opened 2022-07-26
Simplify AuthorizationManager composition


jzheaux:
To compose two authorization managers, some boilerplate is necessary. For example, to achieve a logical "or" between two authorization managers, they can be composed like so:

```java
(authentication, object) -> {
    AuthorizationDecision decision = this.first.check(authentication, object);
    if (decision != null && decision.isGranted()) {
        return decision;
    }
    return this.second.check(authentication, object);
}
```

It would be nice to have an easy way to compose authorization managers in an "any" or "all" fashion, similar to the `AffirmativeBased` and `UnanimousBased` access decision managers.

One way to do this would be to introduce a static class `AuthorizationManagers` with convenience methods like so:

```java
public static <T> AuthorizationManager<T> any(AuthorizationManager<T>... authorizationManagers);

public static <T> AuthorizationManager<T> all(AuthorizationManager<T>... authorizationManagers);
```

Then, the above code could be simplified to:

```java
AuthorizationManagers.any(this.first, this.second);
```

evgeniycheban:
Hi, @jzheaux I can take this.

gourav:
Hi @jzheaux 
May I work on this ?

jzheaux:
@evgeniycheban, @gourav thanks both for your interest.

Since @evgeniycheban was first to post, I'll assign to him. @gourav please consider helping out with the review process once a PR is ready.

KaVeKa:
I believe there is a small difference in the implementation between AffirmativeBased and AuthorizationManagers.anyOf(). 
More precisely in handling `AccessDecisionVoter.ABSTAIN` vs a null instance of AuthorizationDecision.

Given 2 AuthorizationDecisions: `null` and `new AuthorizationDecision(false)`
AffirmativeBased will deny access because it checks all votes, while the AuthorizationManagers.anyOf() will allow access because of the shortcut on `decision == null`.

It is documented clearly in the javadoc in AuthorizationManagers, but the [migration docs](https://docs.spring.io/spring-security/reference/5.8/migration/servlet/authorization.html#_i_use_affirmativebased) make it seem like this is completely equivalent.

Just wanted to note this small discrepancy in the documentation :-)

PR #11756, opened 2022-08-26
Simplify AuthorizationManager composition

evgeniycheban:
Closes gh-11625

### DIFF
--- a/core/src/main/java/org/springframework/security/authorization/AuthorizationManagers.java
+++ b/core/src/main/java/org/springframework/security/authorization/AuthorizationManagers.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2002-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.authorization;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A factory class to create an {@link AuthorizationManager} instances.
+ *
+ * @author Evgeniy Cheban
+ * @since 5.8
+ */
+public final class AuthorizationManagers {
+
+	/**
+	 * Creates an {@link AuthorizationManager} that grants access if at least one
+	 * {@link AuthorizationManager} granted or abstained, if <code>managers</code> are
+	 * empty then denied decision is returned.
+	 * @param <T> the type of object that is being authorized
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/springframework/security/authorization/AuthorizationManagers.java
```diff
@@ -0,0 +1,72 @@
...
+	@SafeVarargs
+	public static <T> AuthorizationManager<T> anyOf(AuthorizationManager<T>... managers) {
+		return (authentication, object) -> {
+			for (AuthorizationManager<T> manager : managers) {
+				AuthorizationDecision decision = manager.check(authentication, object);
+				if (decision == null || decision.isGranted()) {
+					return decision;
+				}
+			}
+			return AuthorizationDecision.DENIED;
```
jzheaux, line 44: Instead of returning `DENIED` here, I think we should return a composite authorization decision. Several authorization managers return extensions of `AuthorizationDecision` that have additional decision context, and we don't want to lose that.

You might consider adding `CompositeAuthorizationDecision` and returning an instance that contains all of the denial decisions.

core/src/main/java/org/springframework/security/authorization/AuthorizationManagers.java
```diff
@@ -0,0 +1,72 @@
...
+	@SafeVarargs
+	public static <T> AuthorizationManager<T> allOf(AuthorizationManager<T>... managers) {
+		return (authentication, object) -> {
+			for (AuthorizationManager<T> manager : managers) {
+				AuthorizationDecision decision = manager.check(authentication, object);
+				if (decision != null && !decision.isGranted()) {
+					return decision;
+				}
+			}
+			return AuthorizationDecision.GRANTED;
```
jzheaux, line 65: Instead of returning `GRANTED` here, I think we should return a composite authorization decision. Several authorization managers return extensions of `AuthorizationDecision` that have additional decision context, and we don't want to lose that.

You might consider adding `CompositeAuthorizationDecision` and returning an instance that contains all of the granted decisions.

core/src/main/java/org/springframework/security/authorization/AuthorizationDecision.java
```diff
@@ -18,10 +18,23 @@
...
+ * @author Evgeniy Cheban
  * @since 5.0
  */
 public class AuthorizationDecision {
 
+	/**
+	 * Represents granted {@link AuthorizationDecision}.
+	 * @since 5.8
+	 */
+	public static final AuthorizationDecision GRANTED = new AuthorizationDecision(true);
```
jzheaux, line 30: I'm still a little hesitant to add these constants. 

They feel similar to `public static final Authentication AUTHENTICATED;` and `public static final Authentication UNAUTHENTICATED;`, which I don't think we'd want to do.

jzheaux (changes requested):

core/src/main/java/org/springframework/security/authorization/AuthorizationManagers.java
evgeniycheban: Done.

core/src/main/java/org/springframework/security/authorization/AuthorizationDecision.java
evgeniycheban: I've removed these constants.

jzheaux (approved):

PR #11773, opened 2022-08-31
xsd document security-context-explicit-save defaults to true

ch4mpy:
Fix missing xsd update after gh-11762


rwinch:
Thanks for the PR @ch4mpy! This is now merged into main

PR #11797, opened 2022-09-08
build and publish RC versions

mojavelinux:
Include RC versions when building docs site.

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -14,10 +14,10 @@ git:
 content:
   sources:
     - url: https://github.com/spring-io/spring-generated-docs
-      branches: 'spring-projects/spring-security/{main,5.{{6..9},{1..9}+({0..9})}.{x,+({0..9})}}'
+      branches: 'spring-projects/spring-security/{main,5.{{6..9},{1..9}+({0..9})}.{x,+({0..9}?(-RC{1..9}))}}'
     - url: https://github.com/spring-projects/spring-security
-      branches: '{main,5.{{6..9},{1..9}+({0..9})}.x}'
-      tags: '5.{{6..9},{1..9}+({0..9})}.+({0..9})'
+      branches: [main, '5.{{6..9},{1..9}+({0..9})}.x']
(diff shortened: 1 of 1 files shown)


antora-playbook.yml
```diff
@@ -14,10 +14,10 @@ git:
 content:
   sources:
     - url: https://github.com/spring-io/spring-generated-docs
-      branches: 'spring-projects/spring-security/{main,5.{{6..9},{1..9}+({0..9})}.{x,+({0..9})}}'
+      branches: 'spring-projects/spring-security/{main,5.{{6..9},{1..9}+({0..9})}.{x,+({0..9}?(-RC{1..9}))}}'
     - url: https://github.com/spring-projects/spring-security
-      branches: '{main,5.{{6..9},{1..9}+({0..9})}.x}'
-      tags: '5.{{6..9},{1..9}+({0..9})}.+({0..9})'
+      branches: [main, '5.{{6..9},{1..9}+({0..9})}.x']
+      tags: '5.{{6..9},{1..9}+({0..9})}.+({0..9})?(-RC{1..9})'
```
sjohnr, line 20: As an aside, I'm wondering if this is generally the place we would manage which and how many versions of the reference docs show up in the picker on the Antora generated site?

@rwinch do we have plans at any point to trim down what versions show up? Right now, it's already quite the list with every version in 5.6/5.7/6.0.

sjohnr (commented):
@mojavelinux is this PR ready to merge? Or is there an prerequisite or step to wait on before merging?

antora-playbook.yml
mojavelinux: For now, this controls what appears in that panel. It would be possible to build versions, but to not display them in that panel, by changing the UI template that makes this list.

We will also be working on being able to build single versions without impacting the versions that have already been published. So this is not the end of the story with regards to what is set here.

rwinch: Yes. We will refine the UI a bit later as it is a lower priority at this point.

sjohnr: Thanks @mojavelinux.

PR #11798, opened 2022-09-08
reconfigure main branch for local builds and as scheduler for docs workflows

mojavelinux:
- set up placeholder and trigger for Deploy Docs workflow in docs-build branch
- set up placeholder and trigger for Rebuild Search Index workflow in docs-build branch
- remove obsolete Deploy Reference workflow
- upgrade Antora to 3.1 and tune settings
- reconfigure docs build for local build only
- add patch to support using linked worktree as Antora content source
- remove Antora extensions only needed for the production docs build

PR #11809, opened 2022-09-09
repurpose 5.8.x branch to provide local docs build

mojavelinux:
* remove unused workflows, scripts, and configuration (now handled by docs-build branch)
* upgrade Antora to 3.1 (and Node.js to 16)
* tune playbook settings
* reconfigure docs build for local build only
* add patch to support using linked worktree as Antora content source
* remove Antora extensions not needed for local builds

Related to https://github.com/spring-projects/spring-security/pull/11798


mojavelinux:
Notice that we don't need to use Collector for a local build since we can rely on Gradle to handle the call to the generateAntora task.

mojavelinux:
Here are the steps to building the docs in this branch.

1. Check out branch (or create a linked worktree)
2. Set the JDK version using `sdk env`
3. Run Antora using Gradle: `./gradlew :spring-security-docs:antora`
4. View the docs in the docs/build/site folder

Issue #7834, opened 2020-01-15
Add constructor to JwtAuthenticationToken that takes principal


xak2000:
Currently, `JwtAuthenticationToken` always pass one `Jwt` instance to super constructor (`AbstractOAuth2TokenAuthenticationToken`) as `token`, `principal` and `credentials`. So, that implies that all three values must always be a jwt token itself.

I don't see any reason why `principal` in JWT token must always be the token itself. It already represents both `token` and `credentials`. But for `principal` I want more flexibility: an ability to also load user (represented by JWT "sub" claim) from DB or external service.

In my application I want to use a JWT token, but I also want to create a custom class that will represent a principal, like `UserDetials` implementation or just a new custom class.

`AbstractOAuth2TokenAuthenticationToken` has a constructor, that takes `token`, `principal` and `credentials` separately.

It would be good, if `JwtAuthenticationToken` also have a constructor, that at least takes `principal` separately.

If this will be added, then it will also be good if `JwtAuthenticationConverter` will have an optional property

```
Converter<Jwt, Object> jwtPrincipalConverter;
```

which, if set, will be used to convert `Jwt` to `principal` (by loading it from DB, external service, or just by creating a more application-friendly `User` object from `Jwt` token) before calling new `JwtAuthenticationToken` constructor.

jzheaux:
@xak2000

> I also want to create a custom class that will represent a principal, like UserDetials implementation or just a new custom class.

`OAuth2AuthenticatedPrincipal` was introduced in 5.2 - does the contract there suit your needs (instead of `UserDetails`, for example)?

5.2 also introduced `BearerTokenAuthentication`, a more generic implementation of `Authentication`. Instead of `Jwt`, it takes an `OAuth2AuthenticatedPrincipal` as its principal.

Given that, you might consider doing:

```java
Converter<Jwt, BearerTokenAuthentication> converter = jwt -> {
    Collection<GrantedAuthority> authorities = // ... 
    OAuth2AccessToken token = new OAuth2AccessToken(BEARER, 
            jwt.getTokenValue(), jwt.getIssuedAt(), jwt.getExpiresAt());
    OAuth2AuthenticatedPrincipal principal = // ... custom principal
    return new BearerTokenAuthentication(principal, token, authorities);
};
```

And then calling the `jwtAuthenticationConverter` DSL method.

Of course, if you need a completely custom class, the contract allows you to do that too:

```java
[...]

PR #11816, opened 2022-09-13
Extendable JwtAuthenticationConverter

michaldo:
There is a code in https://github.com/spring-projects/spring-security/blob/5.7.3/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java#L405 
OAuth2ResourceServerConfigurer
```
Converter<Jwt, ? extends AbstractAuthenticationToken> getJwtAuthenticationConverter() {
    if (this.jwtAuthenticationConverter == null) {
        if (this.context.getBeanNamesForType(JwtAuthenticationConverter.class).length > 0) {
            this.jwtAuthenticationConverter = this.context.getBean(JwtAuthenticationConverter.class);
        }
        else {
            this.jwtAuthenticationConverter = new JwtAuthenticationConverter();
        }
    }
    return this.jwtAuthenticationConverter;
}
```

*Configurable* JwtAuthenticationConverter has limited functionality, because its ability to modify JWTAuthenticationToken is limited to *name* and *authorities*.

This PR extends modification ability by *Extendable* JwtAuthenticationConverter. 

(BTW, ideally would be *Replaceable* JwtAuthenticationConverter: interface JwtAuthenticationConverter extends Converter<Jwt, AbstractAuthenticationToken>+ DefaultJwtAuthenticationConverter implements JwtAuthenticationConverter

My business case is similar to: https://github.com/spring-projects/spring-security/issues/7834

I am quite satisfied how JWT is converted in JWTAuthenticationToken, but rest of my application needs custom Authentication or at least custom Principal. I'm going to implement own JwtAuthenticationConverter which takes JWTAuthenticationToken and returns something customized

Current JwtAuthenticationConverter allows me only customize Authentication *name* and *authorities*. I want customize at least Principal at most whole Authentication

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationConverter.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationConverter.java
@@ -39,7 +39,7 @@ public class JwtAuthenticationConverter implements Converter<Jwt, AbstractAuthen
 	private String principalClaimName = JwtClaimNames.SUB;
 
 	@Override
-	public final AbstractAuthenticationToken convert(Jwt jwt) {
+	public AbstractAuthenticationToken convert(Jwt jwt) {
 		Collection<GrantedAuthority> authorities = this.jwtGrantedAuthoritiesConverter.convert(jwt);
 
 		String principalClaimValue = jwt.getClaimAsString(this.principalClaimName);


michaldo:
I realized now that I solve my business case with 
```
.oauth2ResourceServer().jwt(jwtConfigurer -> jwtConfigurer.jwtAuthenticationConverter(myCustomConverter))
```
I prefer that option over extending JwtAuthenticationConverter. I don't like extend class and completely change behavior.

For me current ability to declare ` jwtConfigurer.jwtAuthenticationConverter()` is good enough. 

Ideally, I would like to JwtAuthenticationConverter be an interface: I could implement it as I want and register as Spring beans. Then my bean would be automatically used in JwtAuthenticationProvider.

Feel free to reject the PR or request change for JwtAuthenticationConverter as a interface and DefaultJwtAuthenticationConverter with current implementation

sjohnr:
Hi @michaldo! Please see the section of the docs on [JwtAuthenticationConverter](https://docs.spring.io/spring-security/reference/servlet/oauth2/resource-server/jwt.html#oauth2resourceserver-jwt-authorization-extraction) for more details on customizing. I'm going to go ahead and close this PR.

Issue #11822, opened 2022-09-15
`JwtAuthenticationProvider` should not aggressively set authentication details


ch4mpy:
**Describe the bug**
`JwtAuthenticationProvider` delegates the instantiation of an `Authentication` to a `Converter<Jwt, ? extends AbstractAuthenticationToken>` but it then alters returned value `details` property.

For framework user providing a `Converter<Jwt, AbstractAuthenticationToken>` bean which sets authentication details, this means he won't get the details he had set when later accessing the authentication (from security expressions for instance).

Please note that `JwtReactiveAuthenticationManager`, the reactive pendant of `JwtAuthenticationProvider` written by @rwinch , does not set `details` after the Authentication is returned by the converter.

**To Reproduce**
- Create a `DemoAuthentication extends AbstractAuthenticationToken` class which sets details in constructor and make it immutable (override `setDetails` to throw an exception)
- Configure a resource-server with JWT decoder and a `Converter<Jwt, DemoAuthentication>`

An exception will be thrown by `JwtAuthenticationProvider` as soon as a request is submitted with a valid JWT.

**Expected behavior**
This kind of side effect on Authentication instance is a problem when the framework user has provided a jwtAuthenticationConverter which intentionally set authentication `details`.

Ideally, `details` property would  be set by the `jwtAuthenticationConverter` (not the `JwtAuthenticationProvider`).

Other sollution with minimum code impact would be preventing `JwtAuthenticationProvider` to touch authentication `details` when it is not null.

**Sample**
https://github.com/ch4mpy/spring-security-11822
- edit application.properties to point to your favorite authorization-server
- get a JWT access-token
- edit `TOKEN` value in `SpringSecurity11822ApplicationTests`
- run `SpringSecurity11822ApplicationTests`

PR #11823, opened 2022-09-15
Prevent JwtAuthenticationProvider from setting authentication details when jwtAuthenticationConverter returned an authentication instance with non null details

ch4mpy:
This could fix gh-11822

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationProvider.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -56,6 +56,7 @@ import org.springframework.util.Assert;
  *
  * @author Josh Cummings
  * @author Joe Grandja
+ * @author Jerome Wacongne ch4mp&#64;c4-soft.com
  * @since 5.1
  * @see AuthenticationProvider
  * @see JwtDecoder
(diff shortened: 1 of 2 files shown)


sjohnr (changes requested):
Thanks @ch4mpy! It would be helpful if you could add a test that ensures the provider is not overwriting details.

ch4mpy:
@sjohnr done

ch4mpy:
@sjohnr anything more needed on this PR?

sjohnr:
@ch4mpy, thanks for asking! I don't think so, but I'm heads down on a few other things and will circle back to this a bit later.

sjohnr (changes requested):
@ch4mpy, when thinking more about this PR, we will need to make an adjustment. For the 5.8/6.0 releases, we are aiming to provide an upgrade to 5.8 that contains no breaking changes. Anything that breaks passivity would require a configuration option that allows users to opt-in to the change to test their compatibility in 5.8, so that an upgrade to 6.0 (where the change is the default) is seamless.

This change is technically a breaking change, as it would have been possible previously to return a non-null `details` which is overridden in the filter and will no longer be. Because the details can be used for things like authorization, we should not include this change in 5.8 without allowing it to be opt-in.

I think the simplest change is to move this PR to be based on 6.0. Can you make this change? I've also added minor suggestions below.

ch4mpy:
@sjohnr rebased on main and updated copyright.

P.S.
Well, the scenario in which one sets detail in authentication converter and this details being overriden by the framework is the exact one a team I know went through and none expected to have details "lost". Reason for me opening the ticket as a "bug".  Of course, whith my recent contribution, it didn't take me long to spot their problem and provide them with a work around.

I hadn't considered it an expected "feature", which is why I initialy based the PR on 5.8

ch4mpy:
Any plan to merge this sometime? 
This would allow me to use immutable `Authentication` implementations in servlets with JWT decoder (can currently do it in reactive apps only).

sjohnr:
Hi @ch4mpy!

> Any plan to merge this sometime?

Apologies, I wasn't able to circle back to this in time for RC1. I'm going to schedule this for 6.1 as I'd prefer to hold off on any code enhancements in the RC phase of 6.0. I'll merge this after the GA release.

sjohnr:
This is merged via 7ad4ebd07ad0c3c8a507cf68bb35308b6595042a

Issue #7856, opened 2020-01-22
OidcUserAuthority should not automatically include ROLE_USER authority


reda-alaoui:
### Summary

I am building a Spring based application that delegates authentication to an OIDC provider like Keycloak. The OIDC could be under my control or not. e.g. Today we use Keycloak, tomorrow we may be using Linkedin.

The users must be granted access to the application by an admin. Since the OIDC may not be under my control, I don't want a random Linkedin user be able to authenticate AND start using the application.

To prevent that, all protected uris require ROLE_ACCESS in addition to a successful authentication. ROLE_ACCESS has to be provided manually by an application admin.

### Actual Behavior

I noticed that, by default, the OIDC authentication has always the ROLE_USER assigned. I am lucky since it does not clash with my expected ROLE_ACCESS, but I find this very disturbing and I am afraid more default roles will be included in the future.

### Expected Behavior

`org.springframework.security.oauth2.core.oidc.user.OidcUserAuthority` should not include a default role, ever.

### Version

5.2.1

### Sample

```java
public class OidcUserAuthority extends OAuth2UserAuthority {
	private final OidcIdToken idToken;
	private final OidcUserInfo userInfo;

	/**
	 * Constructs a {@code OidcUserAuthority} using the provided parameters.
	 *
	 * @param idToken the {@link OidcIdToken ID Token} containing claims about the user
	 */
	public OidcUserAuthority(OidcIdToken idToken) {
		this(idToken, null);
	}

	/**
	 * Constructs a {@code OidcUserAuthority} using the provided parameters
	 * and defaults {@link #getAuthority()} to {@code ROLE_USER}.
	 *
	 * @param idToken the {@link OidcIdToken ID Token} containing claims about the user
	 * @param userInfo the {@link OidcUserInfo UserInfo} containing claims about the user, may be {@code null}
	 */
	public OidcUserAuthority(OidcIdToken idToken, OidcUserInfo userInfo) {
		this("ROLE_USER", idToken, userInfo);
	}
```

jgrandja:
@reda-alaoui `ROLE_USER` is a sensible default role for a "regular" authenticated user. I'd like to understand how this is an issue with your setup?

>  I am afraid more default roles will be included in the future

No, there won't be any further default roles assigned. User authority mapping is application specific and is typically configured by the application to assign custom roles that are understood by that application.

Refer to the reference:
[...]

PR #11887, opened 2022-09-21
Change default authority for oauth2Login()

sjohnr:
Previously, the default authority was `ROLE_USER` when using `oauth2Login()` for both OAuth2 and OIDC providers.

* Default authority for `OAuth2UserAuthority` is now `OAUTH2_USER`
* Default authority for `OidcUserAuthority` is now `OIDC_USER`

Documentation has been updated to include this implementation detail.

Closes gh-7856

### DIFF
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/oidc/user/OidcUserAuthority.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/oidc/user/OidcUserAuthority.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2017 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/oidc/user/OidcUserAuthority.java
```diff
@@ -54,7 +54,7 @@ public OidcUserAuthority(OidcIdToken idToken) {
 	 * may be {@code null}
 	 */
```
jgrandja, line 55: Javadoc needs updating from ROLE_USER to OIDC_USER

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/user/OAuth2UserAuthority.java
```diff
@@ -47,7 +47,7 @@ public class OAuth2UserAuthority implements GrantedAuthority {
 	 * @param attributes the attributes about the user
```
jgrandja, line 47: Javadoc needs updating from ROLE_USER to OAUTH2_USER

jgrandja (changes requested):
Thanks for the PR @sjohnr.

There are 2 minor updates needed and then please go ahead and merge.

Thanks!

sjohnr:
Merged via 181ee7410b2f4503d1d8513be68df510775fa958.

Issue #10506, opened 2021-11-15
Update PasswordEncoder Minimums


jzheaux:
Based on https://github.com/spring-projects/spring-security/pull/10447#issuecomment-966593724, `Argon2PasswordEncoder`, `ScryptPasswordEncoder`, and `Pbkdf2PasswordEncoder` should have their minimums updated.

Please also see gh-7411 gh-4788

larsgrefer:
see also https://github.com/spring-projects/spring-security/issues/7411#issuecomment-531031619

ioanadinuit:
Hi, did someone take this issue? I would like to contribuite.

jzheaux:
@ioanadinuit, thanks for volunteering! The ticket is yours.

One thing we need to keep in mind is what will happen to folks when they upgrade to Spring Security 6. 

Ideally, existing hashes will still work which is not as simple with `Pbkdf2PasswordEncoder`, which does not support upgrading. It may be necessary to implement https://github.com/spring-projects/spring-security/issues/9833 before changing PBKDF2 default settings.

jzheaux:
@Sc00bz, I'd prefer to go off of the recommendations [on the OWASP cheat sheet](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Password_Storage_Cheat_Sheet.md#introduction). They vary just slightly from your recommendations in [your comment](https://github.com/spring-projects/spring-security/pull/10447#issuecomment-966593724). Do you have any concerns with going with the recommendations from the cheat sheet?

rwinch:
Please also see gh-7411

Sc00bz:
@jzheaux That OWASP cheat sheet is based on my recommendations.

The four changes I suggested to this project's defaults:

----

> Argon2's setting are `m=4 MiB, t=3, p=1` either set it to 10 MiB or 7 iterations.

OWASP cheat sheet has for Argon2:
> `m=37 MiB, t=1, p=1`
> `m=15 MiB, t=2, p=1`

The curator trimmed the list at the first two. Numbers came from these formulas:
Argon2i: `m≥74219/(3*t-1)*α, t≥3, p=1`
Argon2{id,d}: `m≥74219/(3*t-1)*α, t≥1, p=1`

`m≥36.24 MiB, t=1, p=1`
`m≥14.50 MiB, t=2, p=1`
`m≥9.06 MiB, t=3, p=1` ****
...
`m≥4.26 MiB, t=6, p=1`
[...]

Issue #10489, opened 2021-11-11
Update default configuration for Pbkdf2PasswordEncoder


Sc00bz:
**Expected Behavior**

`Pbkdf2PasswordEncoder` should not let you generate a hash longer than the hash function. This should either return an error or reduce the output size to match the hash output size: 160 for SHA1, 256 for SHA256, and 512 for SHA512.

**Current Behavior**

`Pbkdf2PasswordEncoder` doesn't prevent bad settings.

**Context**

`Pbkdf2PasswordEncoder` defaults to SHA1 and 256 bits of output. This causes the defender to waste time generating the last 96 bits of output that an attacker is just going to ignore. Also PBKDF2-SHA1 with 185k iterations does not meet the bare minimum for good settings:
PBKDF2-SHA512: 120,000
PBKDF2-SHA256: 350,000
PBKDF2-SHA1: 720,000

I suggest you at least change PBKDF2's default to SHA512 and 120k iterations or more. Also you could change `matches()` to truncate the hash to save time on checking.

Issue #9833, opened 2021-05-28
Support phc Password Storage


rwinch:
We should support [phc password storage](https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md) as described in [OWASP Password Storage Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#upgrading-legacy-hashes).

PR #11904, opened 2022-09-26
Update default configuration for PasswordEncoders

jgrandja:
Update default configuration for `Argon2PasswordEncoder`, `SCryptPasswordEncoder` and `Pbkdf2PasswordEncoder`, as per recommended minimums outlined in [OWASP Cheat Sheet Series ](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html).

### Argon2PasswordEncoder

**Recommendation:**
Use Argon2id with a minimum configuration of 15 MiB of memory, an iteration count of 2, and 1 degree of parallelism.

**Previous default configuration:**
memory=4, iterations=3, parallelism=1

**New default configuration:**
memory=16, iterations=2, parallelism=1

### SCryptPasswordEncoder

**Recommendation:**
Use scrypt with a minimum CPU/memory cost parameter of (2^16), a minimum block size of 8 (1024 bytes), and a parallelization parameter of 1.

**Previous default configuration:**
cpuCost=16384, memoryCost=8, parallelism=1

**New default configuration:**
cpuCost=65536, memoryCost=8, parallelism=1

The default salt length was also updated from 64 to 16.

### Pbkdf2PasswordEncoder

**Recommendation:**
If FIPS-140 compliance is required, use PBKDF2 with a work factor of 310,000 or more and set with an internal hash function of HMAC-SHA-256.

**Previous default configuration:**
algorithm=SHA1, iterations=185000, hashLength=256

**New default configuration:**
algorithm=SHA256, iterations=310000, hashLength=256

The default salt length was also updated from 8 to 16.

Closes gh-10506, Closes gh-10489

### DIFF
--- a/crypto/src/main/java/org/springframework/security/crypto/factory/PasswordEncoderFactories.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/factory/PasswordEncoderFactories.java
@@ -53,7 +53,9 @@ public final class PasswordEncoderFactories {
 	 * <li>noop -
 	 * {@link org.springframework.security.crypto.password.NoOpPasswordEncoder}</li>
 	 * <li>pbkdf2 - {@link Pbkdf2PasswordEncoder}</li>
-	 * <li>scrypt - {@link SCryptPasswordEncoder}</li>
+	 * <li>scrypt - {@link SCryptPasswordEncoder#defaultsForSpringSecurity_v4_1()}</li>
+	 * <li>scrypt@SpringSecurity_v5_8 -
+	 * {@link SCryptPasswordEncoder#defaultsForSpringSecurity_v5_8()}</li>
 	 * <li>SHA-1 - {@code new MessageDigestPasswordEncoder("SHA-1")}</li>
 	 * <li>SHA-256 - {@code new MessageDigestPasswordEncoder("SHA-256")}</li>
 	 * <li>sha256 -
@@ -74,7 +76,8 @@ public final class PasswordEncoderFactories {
 		encoders.put("MD5", new org.springframework.security.crypto.password.MessageDigestPasswordEncoder("MD5"));
 		encoders.put("noop", org.springframework.security.crypto.password.NoOpPasswordEncoder.getInstance());
 		encoders.put("pbkdf2", new Pbkdf2PasswordEncoder());
-		encoders.put("scrypt", new SCryptPasswordEncoder());
+		encoders.put("scrypt", SCryptPasswordEncoder.defaultsForSpringSecurity_v4_1());
+		encoders.put("scrypt@SpringSecurity_v5_8", SCryptPasswordEncoder.defaultsForSpringSecurity_v5_8());
 		encoders.put("SHA-1", new org.springframework.security.crypto.password.MessageDigestPasswordEncoder("SHA-1"));
 		encoders.put("SHA-256",
 				new org.springframework.security.crypto.password.MessageDigestPasswordEncoder("SHA-256"));
--- a/crypto/src/main/java/org/springframework/security/crypto/factory/PasswordEncoderFactories.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/factory/PasswordEncoderFactories.java
@@ -52,7 +52,9 @@ public final class PasswordEncoderFactories {
 	 * <li>MD5 - {@code new MessageDigestPasswordEncoder("MD5")}</li>
 	 * <li>noop -
 	 * {@link org.springframework.security.crypto.password.NoOpPasswordEncoder}</li>
-	 * <li>pbkdf2 - {@link Pbkdf2PasswordEncoder}</li>
+	 * <li>pbkdf2 - {@link Pbkdf2PasswordEncoder#defaultsForSpringSecurity_v5_5()}</li>
+	 * <li>pbkdf2@SpringSecurity_v5_8 -
+	 * {@link Pbkdf2PasswordEncoder#defaultsForSpringSecurity_v5_8()}</li>
 	 * <li>scrypt - {@link SCryptPasswordEncoder#defaultsForSpringSecurity_v4_1()}</li>
 	 * <li>scrypt@SpringSecurity_v5_8 -
 	 * {@link SCryptPasswordEncoder#defaultsForSpringSecurity_v5_8()}</li>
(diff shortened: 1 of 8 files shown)


jgrandja:
@Sc00bz @larsgrefer Thank you for all your valuable feedback so far.

When you have a moment, I would appreciate it if you can take a look at the updates in this PR and provide any feedback. Thank you.

Sc00bz:
> cpuCost=65536, memoryCost=8, parallelism=1

Your names are incorrect "cpuCost" is "N" (number of blocks) and "memoryCost" is "r" or "rate" (block size). Note N\*r is memory cost and you can think of parallelism as a time cost.

Pbkdf2PasswordEncoder.java, line 53:
```
private static final int DEFAULT_SALT_LENGTH = 8;
```
You might get complaints with 64 bits being low (128 to 256 bits is normal). Note 64 bits is fine because 99.99%+ of salts will be globally unique.

SCryptPasswordEncoder.java, line 69:
```
private static final int DEFAULT_SALT_LENGTH = 64;
```
512 bits for a salt is overkill.

jgrandja:
@Sc00bz 

> Your names are incorrect "cpuCost" is "N" (number of blocks) and "memoryCost" is "r" or "rate" (block size)

Yeah, I realize the member names might be misleading but if you look at the javadoc: 

```java
/**
 * Creates a new instance
 *
 * @param cpuCost cpu cost of the algorithm (as defined in scrypt this is N). must be
 * power of 2 greater than 1. Default is currently 16,384 or 2^14)
 * @param memoryCost memory cost of the algorithm (as defined in scrypt this is r)
 * Default is currently 8.
 * ...
 */
public SCryptPasswordEncoder(int cpuCost, int memoryCost, int parallelization, int keyLength, int saltLength) {
```

`cpuCost` represents "N" and `memoryCost` represents "r".

I'll update the `salt` for `pbkdf2` and `scrypt`.

Thanks for the feedback.

crypto/src/main/java/org/springframework/security/crypto/factory/PasswordEncoderFactories.java
```diff
@@ -71,7 +71,14 @@ public static PasswordEncoder createDelegatingPasswordEncoder() {
...
 		encoders.put("MD5", new org.springframework.security.crypto.password.MessageDigestPasswordEncoder("MD5"));
 		encoders.put("noop", org.springframework.security.crypto.password.NoOpPasswordEncoder.getInstance());
-		encoders.put("pbkdf2", new Pbkdf2PasswordEncoder());
+
+		// Preserve the original settings for PBKDF2 for backwards compatibility.
+		// The default configuration has been updated to the recommended minimums.
+		// See gh-10506 Update PasswordEncoder Minimums
+		Pbkdf2PasswordEncoder pbkdf2PasswordEncoder = new Pbkdf2PasswordEncoder("", 8, 185000, 256);
+		pbkdf2PasswordEncoder.setAlgorithm(Pbkdf2PasswordEncoder.SecretKeyFactoryAlgorithm.PBKDF2WithHmacSHA1);
+		encoders.put("pbkdf2", pbkdf2PasswordEncoder);
```
rwinch, line 80: I think we should have an instance of Pbkdf2 added that uses the new defaults added because without it there is no way to encode passwords with pbkdf2 new defaults.

rwinch (changes requested):
I like your idea around the minimums based upon the Spring Security version. I'd like to propose we take it a little further to provide an easier migration path for our users by allowing users to easily opt into changes in 5.8. Rather than changing the default constructor, I'd like to propose we do a few things:

 Starting in 5.8 

- Introduce a static factory method that is named with the major and minor version of Spring Security that introduced the defaults for each encoder. For example defaultsForSecurity5_0()
- Deprecate default constructors
- Introduce a static factory method that is named with 6.0 version of Spring Security that introduces the defaults provided here.

In 6.0 remove the default constructor, but preserve the static factory methods as deprecated methods so that applications can continue to work.

Please also ensure there is a test for each instance of the new password encoder validating a password that was encoded from the older settings. Please hard code the encoded password.

crypto/src/main/java/org/springframework/security/crypto/factory/PasswordEncoderFactories.java
jgrandja: Added new `encodingId` `pbkdf2-sha256` that is configured with the new defaults.

jgrandja:
@rwinch I applied the static factory methods to `Pbkdf2PasswordEncoder` to help users with the migration to `5.8`.

However, I did not apply the static factory methods to `Argon2PasswordEncoder` or `SCryptPasswordEncoder` since they both support `upgradeEncoding()`. I don't feel it makes sense here but if there are reasons I'm missing please let me know.

crypto/src/main/java/org/springframework/security/crypto/factory/PasswordEncoderFactories.java
```diff
@@ -71,7 +73,8 @@ public static PasswordEncoder createDelegatingPasswordEncoder() {
 		encoders.put("MD4", new org.springframework.security.crypto.password.Md4PasswordEncoder());
 		encoders.put("MD5", new org.springframework.security.crypto.password.MessageDigestPasswordEncoder("MD5"));
 		encoders.put("noop", org.springframework.security.crypto.password.NoOpPasswordEncoder.getInstance());
-		encoders.put("pbkdf2", new Pbkdf2PasswordEncoder());
+		encoders.put("pbkdf2", Pbkdf2PasswordEncoder.defaultsForSpringSecurity_v5_5());
+		encoders.put("pbkdf2-sha256", Pbkdf2PasswordEncoder.defaultsForSpringSecurity_v5_8());
```
rwinch, line 77: I think the -sha256 should be also related to v5_8 or all alg parameters since there are other parameters that could change. I realize that the recommendations are based on the algorithm, but these recommendations are for now and may change. That would make it difficult for us to provide additional settings later.

My quick thought is that I'd prefer the id to contain all parameters even if it is verbose as this is easier for users to understand and will be similar to other encodings work (I.e. [phc](https://github.com/spring-projects/spring-security/issues/9833), bcrypt, etc)

rwinch (changes requested):
I've left feedback inline. In addition, please update the other encoders to have static factory methods as we discussed offline.

crypto/src/main/java/org/springframework/security/crypto/factory/PasswordEncoderFactories.java
jgrandja: > My quick thought is that I'd prefer the id to contain all parameters even if it is verbose

I still think it would be too verbose and not that easy to understand.

I decided to align the `encodingId` to the static factory method name, for example:

- `argon2@SpringSecurity_v5_8` -> `Argon2PasswordEncoder.defaultsForSpringSecurity_v5_8()`
- `scrypt@SpringSecurity_v5_8` -> `SCryptPasswordEncoder.defaultsForSpringSecurity_v5_8()`
- `pbkdf2@SpringSecurity_v5_8` -> `Pbkdf2PasswordEncoder.defaultsForSpringSecurity_v5_8()`

PR #11906, opened 2022-09-27
Observability

jzheaux:


core/src/main/java/org/springframework/security/authentication/AuthenticationObservationConvention.java
```diff
@@ -0,0 +1,49 @@
...
+import io.micrometer.observation.ObservationConvention;
+
+public final class AuthenticationObservationConvention
+		implements ObservationConvention<AuthenticationObservationContext> {
+
+	@Override
+	public KeyValues getLowCardinalityKeyValues(AuthenticationObservationContext context) {
+		KeyValues kvs = KeyValues.empty();
+		if (context.getAuthenticationRequest() != null) {
+			kvs = kvs.and("authentication.request.type", context.getAuthenticationRequest().getAuthenticationType());
```
jonatan-ivanov, line 30: This is not necessarily the scope of this PR but we should consider using the `DocumentedObservation` instead of defining keys as String.

jonatan-ivanov, line 30: I would recommend adding a default "unknown", "n/a", "empty", "null", etc. value if the data is not available.

The reason behind this, is that that way you will always have the same set of tags on all of your observations so it will be easier for the users to navigate and also some metrics backends (e.g.: Prometheus) don't like if the number of tags are different.

So this would look like:
```java
if (context.getAuthenticationRequest() != null) {
    kvs = kvs.and("authentication.request.type", context.getAuthenticationRequest().getAuthenticationType());
}
else {
    kvs = kvs.and("authentication.request.type", "n/a");
}
```
or
```java
String requestType = context.getAuthenticationRequest() != null ? context.getAuthenticationRequest().getAuthenticationType() : "n/a";
kvs = kvs.and("authentication.request.type", requestType);
```

core/src/main/java/org/springframework/security/authentication/ObservationAuthenticationManager.java
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * An {@link AuthenticationManager} that observes the authentication
+ *
+ * @author Josh Cummings
+ * @since 6.0
+ */
+public final class ObservationAuthenticationManager implements AuthenticationManager {
+
+	static final String OBSERVATION_NAME = "spring.security.authentications";
```
jonatan-ivanov, line 34: That's also where the `ObservationDocumentiation` can help, see my previous comment, not necessarily the scope of this PR.

core/src/main/java/org/springframework/security/authentication/ObservationAuthenticationManager.java
```diff
@@ -0,0 +1,62 @@
...
+		this.registry = registry;
+		this.delegate = delegate;
+	}
+
+	@Override
+	public Authentication authenticate(Authentication authentication) throws AuthenticationException {
+		AuthenticationObservationContext context = new AuthenticationObservationContext();
+		context.setAuthenticationRequest(authentication);
+		context.setAuthenticationManager(this.delegate.getClass());
+		return Observation.createNotStarted(OBSERVATION_NAME, context, this.registry)
```
jonatan-ivanov, line 54: There are two main flavors of `createNotStarted`: one is where to pass a name, the other is where you pass a convention.

It is recommended to pass a convention to `createNotStarted`, you can define the name there, there is a method you can override, with that you can delete the `observationConvention` call and the `OBSERVATION_NAME` constant here.

core/src/main/java/org/springframework/security/authentication/ObservationReactiveAuthenticationManager.java
```diff
@@ -0,0 +1,64 @@
...
+		this.delegate = delegate;
+	}
+
+	@Override
+	public Mono<Authentication> authenticate(Authentication authentication) throws AuthenticationException {
+		AuthenticationObservationContext context = new AuthenticationObservationContext();
+		context.setAuthenticationRequest(authentication);
+		context.setAuthenticationManager(this.delegate.getClass());
+		Observation observation = Observation.createNotStarted(OBSERVATION_NAME, context, this.registry)
+				.observationConvention(this.convention).start();
```
jonatan-ivanov, line 54: Same as the previous one, please use the overload of `createNotStarted` that accepts a convention and define the name in the convention.

core/src/main/java/org/springframework/security/authorization/AuthorizationObservationConvention.java
```diff
@@ -0,0 +1,69 @@
...
+public final class AuthorizationObservationConvention
+		implements ObservationConvention<AuthorizationObservationContext<?>> {
+
+	@Override
+	public KeyValues getLowCardinalityKeyValues(AuthorizationObservationContext<?> context) {
+		KeyValues kvs = KeyValues.empty();
+		Authentication authentication = context.getAuthentication();
+		if (authentication != null) {
+			String type = authentication.getAuthenticationType();
+			kvs = KeyValues.of("authentication.type", type);
```
jonatan-ivanov, line 37: Same as my previous comment: please define a default value.

web/src/main/java/org/springframework/security/web/AroundFilterObservation.java
```diff
@@ -0,0 +1,147 @@
...
+
+	class SimpleAroundFilterObservation implements AroundFilterObservation {
+
+		private final Iterator<Observation> observations;
+
+		private final Observation before;
+
+		private final Observation after;
+
+		private AtomicReference<Observation> currentObservation = new AtomicReference<>(null);
```
jonatan-ivanov, line 55: I'm not sure if it is more convenient or not but `Observation.Scope` has a reference to the `Observation` that you can get via `getCurrentObservation()` so you might get away with one `AtomicReference`.

web/src/main/java/org/springframework/security/web/AroundFilterObservation.java
```diff
@@ -0,0 +1,147 @@
...
+		}
+
+		@Override
+		public void start() {
+			if (this.observations.hasNext()) {
+				stop();
+				Observation observation = this.observations.next();
+				observation.start();
+				Observation.Scope scope = observation.openScope();
+				this.currentScope = new AtomicReference<>(scope);
```
jonatan-ivanov, line 72: Shouldn't we make the `AtomicReference` `final` and use `currentScope.set(scope)` instead?

Because this is changing the reference of the `currentScope` and visibility is not guaranteed, it's not thread safe. We can either use a `volatile Observation.Scope` or an `AtomicReference` but in that case, the `AtomicReference` should be `final` and `.set(...)` should be used to change the value.

web/src/main/java/org/springframework/security/web/FilterChainProxy.java
```diff
@@ -147,6 +154,12 @@ public class FilterChainProxy extends GenericFilterBean {
 
 	private static final String FILTER_APPLIED = FilterChainProxy.class.getName().concat(".APPLIED");
 
+	static final String REQUEST_OBSERVATION_NAME = "http.requests";
```
jonatan-ivanov, line 157: If this is for the whole request, Spring MVC already covers this, it is called `http.server.requests`.

web/src/main/java/org/springframework/security/web/FilterChainProxy.java
```diff
@@ -147,6 +154,12 @@ public class FilterChainProxy extends GenericFilterBean {
...
+
+	static final String UNSECURED_OBSERVATION_NAME = "http.unsecured.requests";
```
jonatan-ivanov, line 159: Can we match these to what MVC uses, like  `http.server.requests.secured` and `http.server.requests.unsecured`?

web/src/main/java/org/springframework/security/web/server/WebFilterChainProxy.java
```diff
@@ -35,8 +43,24 @@
  */
 public class WebFilterChainProxy implements WebFilter {
 
+	private static final BiFunction<WebHandler, List<WebFilter>, WebFilterChain> DEFAULT_CHAIN = DefaultWebFilterChain::new;
+
+	static final String REQUEST_OBSERVATION_NAME = "server.http.requests";
```
jonatan-ivanov, line 48: Can we match these to Webflux? (See my previous comment about MVC)

web/src/main/java/org/springframework/security/web/FilterChainProxy.java
jzheaux: I'd prefer to have a simple way to turn all Spring Security observations off. Not that it can't be done another way, but right now I've just updated the PR so that all start with `spring.security.`. In this way, an application can create a very simple predicate to turn on Spring Security observations.

I see the value of aligning, too... it would be nice if there were a way to get both. One way may be to introduce a class like `SpringSecurityObservationPredicates#noObservations()` or similar. Is there another potential route?

web/src/main/java/org/springframework/security/web/AroundFilterObservation.java
jzheaux: agreed

web/src/main/java/org/springframework/security/web/FilterChainProxy.java
jonatan-ivanov: I like the `SpringSecurityObservationPredicates#noObservations()` approach.

core/src/main/java/org/springframework/security/core/Authentication.java
```diff
@@ -134,4 +134,8 @@ public interface Authentication extends Principal, Serializable {
 	 */
 	void setAuthenticated(boolean isAuthenticated) throws IllegalArgumentException;
 
+	default String getAuthenticationType() {
```
rwinch, line 137: While this is convenient, I'd prefer not to add extra noise to the interface. If this was really something that was necessary, I'd expect that Object had a convenience method for this (which doesn't make sense to me).

rwinch (changes requested):
My biggest comment is, I'd like to see if it makes sense / is possible to make observability an optional dependency. This might not happen in this release, but may just be possible with the current codebase to do passively. This means that core components can only use the observability APIs privately.

There are also a number of classes / interfaces missing class level Javadoc that would benefit from it.

config/src/main/java/org/springframework/security/config/annotation/method/configuration/Jsr250MethodSecurityConfiguration.java
```diff
@@ -40,28 +50,54 @@
...
+		this.jsr250AuthorizationManager.manager.setRolePrefix(grantedAuthorityDefaults.getRolePrefix());
 	}
 
 	@Autowired(required = false)
 	void setSecurityContextHolderStrategy(SecurityContextHolderStrategy securityContextHolderStrategy) {
-		this.securityContextHolderStrategy = securityContextHolderStrategy;
+		this.interceptor.setSecurityContextHolderStrategy(securityContextHolderStrategy);
+	}
+
+	private static final class ObservationJsr250AuthorizationManager implements AuthorizationManager<MethodInvocation> {
```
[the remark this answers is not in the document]
jzheaux: They would need to `@Import` `AutoProxyRegistrar` and then do:

```java
@Bean
Advisor jsr(ObservationRegistry registry) {
    Jsr250AuthorizationManager delegate = new Jsr250AuthorizationManager();
    return AuthorizationManagerBeforeMethodInterceptor.jsr250(
            new ObservationAuthorizationManager<>(registry, delegate));
}
```

core/src/main/java/org/springframework/security/core/Authentication.java
jzheaux: It's less about convenience (for me) and more about allowing authentication implementations to configure what gets published to Micrometer as the authentication type.

That said, I think this was added fairly early on before I understood Micrometer's convention API. I'll remove this and we can make customizing the observation convention a possibility in a future release.

config/src/main/java/org/springframework/security/config/annotation/method/configuration/Jsr250MethodSecurityConfiguration.java
```diff
@@ -40,28 +50,54 @@
...
+
+		private Jsr250AuthorizationManager manager = new Jsr250AuthorizationManager();
```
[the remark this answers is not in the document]
jzheaux: I don't want to reuse the class because I think that it is going to go away. Right now it's there because of a bean loading cycle. In a Boot application -- when this class uses `@Autowired` to get the `ObservationRegistry` -- the registry gets instantiated too early by Boot and Boot's `ObservationRegistry` auto-configuration breaks.

config/src/main/java/org/springframework/security/config/annotation/method/configuration/Jsr250MethodSecurityConfiguration.java
```diff
@@ -40,28 +50,54 @@
...
+
+		private Supplier<ObservationAuthorizationManager<MethodInvocation>> observation;
+
+		private ObservationJsr250AuthorizationManager(ApplicationContext context) {
+			this.observation = SingletonSupplier.of(() -> {
+				ObservationRegistry registry = ObservationRegistry.NOOP;
```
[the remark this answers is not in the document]
jzheaux: Yes, I'd like to as well. I think this is dependent on what the Boot team says regarding https://github.com/spring-projects/spring-security/pull/11906#discussion_r991640168

core/src/main/java/org/springframework/security/authentication/AuthenticationObservationConvention.java
jzheaux: Added https://github.com/spring-projects/spring-security/issues/11995

web/src/main/java/org/springframework/security/web/FilterChainProxy.java
jzheaux: For simplicity, I ultimately decided to leave the wrapping observation out. We can always add it later if needed.

config/src/main/java/org/springframework/security/config/annotation/method/configuration/Jsr250MethodSecurityConfiguration.java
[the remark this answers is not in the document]
jzheaux: Note that in the updated PR, this class is no longer needed.

config/src/main/java/org/springframework/security/config/annotation/method/configuration/Jsr250MethodSecurityConfiguration.java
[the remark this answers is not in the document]
jzheaux: Andy helped me untangle the cycle, and now this class and the related ones are no longer needed.

Issue #9631, opened 2021-04-13
XSS protection should be set to 0 by default per updated OWASP recommendation


candrews:
**Expected Behavior**

By default, Spring Security should send this HTTP header:
`X-XSS-Protection: 0`

**Current Behavior**

By default, Spring Security sends this HTTP header:
`X-XSS-Protection: 1; mode=block`

**Context**
OWASP has updated its recommendation regarding the `X-XSS-Protection` HTTP header.

OWASP used to recommend the header be set to `1; mode=block`, which is what Spring Security does by default today. See https://docs.spring.io/spring-security/site/docs/current/reference/html5/#headers-xss-protection

However, they now recommend the header be set to `0`

See https://github.com/OWASP/CheatSheetSeries/issues/376

rwinch:
Thanks for the report. Given this breaks passivity, I've assigned it for the 6.x release

blacelle:
While this can be easily configured in [HeadersConfigurer.xssProtection()](https://github.com/spring-projects/spring-security/blob/main/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java#L761):

	.xssProtection()
	.xssProtectionEnabled(false)


It is more complicated with [ServerHttpSecurity.HeaderSpec.xssProtection()](https://github.com/spring-projects/spring-security/blob/main/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java#L2854) as the header can only be removed (instead of configured to 0 / the recommended value).

Kehrlann:
Hey there 👋 I'm interested in picking this one up.

PR #11908, opened 2022-09-27
Make X-Xss-Protection header value configurable in ServerHttpSecurity

Kehrlann:
Make `X-Xss-Protection` header configurable in `ServerHttpSecurity`, re issue: gh-9631.
XSS Protection is configurable through `ServerHttpSecurity#xssProtection#headerValue`, with a `HeaderValue` enum  (DISABLED / ENABLED / ENABLED_MODE_BLOCK). This is because the state `!enabled && block` is invalid, thus we restrict the API to the three valid states only.

I ported the changes over to the Servlet stack, which has more configuration exposed in `HttpSecurity`, so the deprecations bubble up to that class too.

The idea in 6.0 is to change the default to `HeaderValue = DISABLED`, and also remove the `enabled` and `block` booleans.

---

As part of this change, I noticed that in servlet the "mode=block" header values are different in the Servlet and Reactive stacks:
- `X-Xss-Protection: 1; mode=block` in Servlet
- `X-Xss-Protection: 1 ; mode=block` in Reactive, with an extra space

Should we plan to align the Reactive header with the Servlet version?

### DIFF
--- a/config/src/main/kotlin/org/springframework/security/config/web/server/ServerXssProtectionDsl.kt
+++ b/config/src/main/kotlin/org/springframework/security/config/web/server/ServerXssProtectionDsl.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,9 +16,13 @@
 
 package org.springframework.security.config.web.server
 
+import org.springframework.security.web.server.header.XXssProtectionServerHttpHeadersWriter.HeaderValue
+
 /**
  * A Kotlin DSL to configure the [ServerHttpSecurity] XSS protection header using
  * idiomatic Kotlin code.
+ *
+ * @property headerValue the value of the X-XSS-Protection header. OWASP recommends [HeaderValue.DISABLED].
  *
  * @author Eleftheria Stein
  * @since 5.4
@@ -26,6 +30,7 @@ package org.springframework.security.config.web.server
 @ServerSecurityMarker
 class ServerXssProtectionDsl {
     private var disabled = false
+    var headerValue: HeaderValue? = null
 
     /**
      * Disables cache control response headers
@@ -36,6 +41,7 @@ class ServerXssProtectionDsl {
 
     internal fun get(): (ServerHttpSecurity.HeaderSpec.XssProtectionSpec) -> Unit {
         return { xss ->
+            headerValue?.also { xss.headerValue(headerValue) }
             if (disabled) {
                 xss.disable()
             }
--- a/config/src/main/kotlin/org/springframework/security/config/web/servlet/headers/XssProtectionConfigDsl.kt
+++ b/config/src/main/kotlin/org/springframework/security/config/web/servlet/headers/XssProtectionConfigDsl.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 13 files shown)


config/src/main/kotlin/org/springframework/security/config/web/server/ServerXssProtectionDsl.kt
```diff
@@ -16,16 +16,21 @@
...
  *
+ * @property headerValue the value of the X-XSS-Protection header. OWASP recommends [HeaderValue.DISABLED].
+ *
  * @author Eleftheria Stein
  * @since 5.4
  */
 @ServerSecurityMarker
 class ServerXssProtectionDsl {
     private var disabled = false
+    var headerValue: HeaderValue? = null
```
sjohnr, line 33: I believe we want to update the copyright year for this file.

config/src/main/kotlin/org/springframework/security/config/web/servlet/headers/XssProtectionConfigDsl.kt
```diff
@@ -28,11 +29,15 @@ import org.springframework.security.config.annotation.web.configurers.HeadersCon
...
  * If false, will explicitly disable specify that X-XSS-Protection is disabled.
+ * @property headerValue the value of the X-XSS-Protection header. OWASP recommends [HeaderValue.DISABLED].
  */
 @HeadersSecurityMarker
 class XssProtectionConfigDsl {
+    @Deprecated("use headerValue instead")
     var block: Boolean? = null
+    @Deprecated("use headerValue instead")
     var xssProtectionEnabled: Boolean? = null
+    var headerValue: HeaderValue? = null
```
sjohnr, line 40: I believe we will want to update the copyright year for this file (and a few others in this PR).

web/src/main/java/org/springframework/security/web/header/writers/XXssProtectionHeaderWriter.java
```diff
@@ -77,37 +71,86 @@ public void writeHeaders(HttpServletRequest request, HttpServletResponse respons
...
+	 * specify mode as blocked. The content will be replaced with "#". For example:
+	 *
+	 * <pre>
+	 * X-XSS-Protection: 1 ; mode=block
+	 * </pre>
+	 * @param headerValue the new header value
+	 * @since 5.8
+	 */
+	public void setHeaderValue(HeaderValue headerValue) {
+		this.headerValue = headerValue;
```
sjohnr, line 130: Is `null` a valid value? Otherwise we might want to add a not null assertion (`Assert.notNull`) and a test?

web/src/main/java/org/springframework/security/web/server/header/XXssProtectionServerHttpHeadersWriter.java
```diff
@@ -73,40 +71,94 @@ public Mono<Void> writeHttpHeaders(ServerWebExchange exchange) {
...
+	 * specify mode as blocked. The content will be replaced with "#". For example:
+	 *
+	 * <pre>
+	 * X-XSS-Protection: 1 ; mode=block
+	 * </pre>
+	 * @param headerValue the new headerValue
+	 * @since 5.8
+	 */
+	public void setHeaderValue(HeaderValue headerValue) {
+		this.headerValue = headerValue;
```
sjohnr, line 131: Add an `Assert.notNull` here as well?

sjohnr (commented):
Thanks @Kehrlann!

> Should we plan to align the Reactive header with the Servlet version?

While I don't know the answer off-hand, shall we create a separate issue to track this? You can label it with `for: team-attention` and ask about it at our next standup if needed.

You may know the process already, but in case it's new:

When merging this, you can merge into `5.8.x`, then immediately merge `5.8.x` into `main`. Afterwards, you would want to update `whats-new.adoc` in `5.8.x` and then `git merge -s ours 5.8.x` into `main` so that What's New is only updated in 5.8.

Let us know if you would like another review before merging. My review below is mostly cosmetic.

sjohnr:
Merged via 93250013e48fd249e219bc14d6f30beb5bf91bf4

sjohnr:
Note that we will want to update What's New with this change as a separate commit in 5.8.

Issue #11918, opened 2022-09-29
Add CsrfTokenRepository.loadDeferredToken(HttpServletRequest, HttpServletResponse)


sjohnr:
We should add a default method `CsrfTokenRepository.loadToken(HttpServletRequest, HttpServletResponse)` that returns `DeferredCsrfToken` (implements `Supplier<CsrfToken>`).

Issue #11892, opened 2022-09-22
CsrfTokenRequestAttributeHandler -> CsrfTokenRequestHandler


rwinch:
We should rename `CsrfTokenRequestAttributeHandler` to `CsrfTokenRequestHandler` and refactor `CsrfFilter` to no longer need `CsrfTokenRepository` but instead use `CsrfTokenRequestHandler`

PR #11731, opened 2022-08-19
Add new interfaces for BREACH protection

sjohnr:
Draft 2
---

* Add `CsrfTokenRequestAttributeHandler` and `CsrfTokenRequestResolver` interfaces
* Add `CsrfTokenRequestProcessor` and `XorCsrfTokenRequestProcessor` classes
* Move `csrfRequestAttributeName` to `CsrfTokenRequestProcessor`

Issue gh-4001

PR #11924, opened 2022-09-29
Add CsrfTokenRepository.loadDeferredToken

sjohnr:
* Move DeferredCsrfToken to top-level and implement Supplier<CsrfToken>
* Move RepositoryDeferredCsrfToken to top-level and make package-private
* Add CsrfTokenRepository.loadToken(HttpServletRequest, HttpServletResponse)
* Update CsrfFilter and make DeferredCsrfToken package-private
* Rename CsrfTokenRepositoryRequestHandler to CsrfTokenRequestAttributeHandler

Issue gh-11892
Closes gh-11918

### DIFF
--- a/web/src/main/java/org/springframework/security/web/csrf/CsrfFilter.java
+++ b/web/src/main/java/org/springframework/security/web/csrf/CsrfFilter.java
@@ -82,30 +82,21 @@ public final class CsrfFilter extends OncePerRequestFilter {
 
 	private final Log logger = LogFactory.getLog(getClass());
 
-	private final CsrfTokenRequestHandler requestHandler;
+	private final CsrfTokenRepository tokenRepository;
 
 	private RequestMatcher requireCsrfProtectionMatcher = DEFAULT_CSRF_MATCHER;
 
 	private AccessDeniedHandler accessDeniedHandler = new AccessDeniedHandlerImpl();
 
-	/**
-	 * Creates a new instance.
-	 * @param csrfTokenRepository the {@link CsrfTokenRepository} to use
-	 * @deprecated Use {@link CsrfFilter#CsrfFilter(CsrfTokenRequestHandler)} instead
-	 */
-	@Deprecated
-	public CsrfFilter(CsrfTokenRepository csrfTokenRepository) {
-		this(new CsrfTokenRepositoryRequestHandler(csrfTokenRepository));
-	}
+	private CsrfTokenRequestHandler requestHandler = new CsrfTokenRequestAttributeHandler();
 
 	/**
 	 * Creates a new instance.
-	 * @param requestHandler the {@link CsrfTokenRequestHandler} to use. Default is
-	 * {@link CsrfTokenRepositoryRequestHandler}.
+	 * @param tokenRepository the {@link CsrfTokenRepository} to use
 	 */
-	public CsrfFilter(CsrfTokenRequestHandler requestHandler) {
-		Assert.notNull(requestHandler, "requestHandler cannot be null");
-		this.requestHandler = requestHandler;
+	public CsrfFilter(CsrfTokenRepository tokenRepository) {
+		Assert.notNull(tokenRepository, "tokenRepository cannot be null");
+		this.tokenRepository = tokenRepository;
 	}
 
 	@Override
@@ -116,7 +107,8 @@ public final class CsrfFilter extends OncePerRequestFilter {
 	@Override
 	protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
 			throws ServletException, IOException {
-		DeferredCsrfToken deferredCsrfToken = this.requestHandler.handle(request, response);
+		DeferredCsrfToken deferredCsrfToken = this.tokenRepository.loadDeferredToken(request, response);
+		this.requestHandler.handle(request, response, deferredCsrfToken::get);
 		if (!this.requireCsrfProtectionMatcher.matches(request)) {
 			if (this.logger.isTraceEnabled()) {
 				this.logger.trace("Did not protect against CSRF since request did not match "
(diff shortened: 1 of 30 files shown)


web/src/main/java/org/springframework/security/web/csrf/CsrfTokenRepository.java
```diff
@@ -55,4 +58,20 @@ public interface CsrfTokenRepository {
...
+	 * {@link Supplier#get()} to return the same {@link CsrfToken} without the cost of
+	 * loading or generating the token again.
+	 * @param request the {@link HttpServletRequest} to use
+	 * @param response the {@link HttpServletResponse} to use
+	 * @return a {@code Supplier} that will load the {@link CsrfToken}
+	 * @since 5.8
+	 */
+	default Supplier<CsrfToken> loadToken(HttpServletRequest request, HttpServletResponse response) {
+		return new RepositoryDeferredCsrfToken(this, request, response);
+	}
```
sjohnr, line 75: This is the first main change highlighted in this PR.

web/src/main/java/org/springframework/security/web/csrf/CsrfFilter.java
```diff
@@ -116,7 +108,8 @@ protected boolean shouldNotFilter(HttpServletRequest request) throws ServletExce
 	@Override
 	protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
 			throws ServletException, IOException {
-		DeferredCsrfToken deferredCsrfToken = this.requestHandler.handle(request, response);
+		Supplier<CsrfToken> deferredCsrfToken = this.tokenRepository.loadToken(request, response);
+		this.requestHandler.handle(request, response, deferredCsrfToken);
```
sjohnr, line 112: This is the second main change highlighted in this PR.

web/src/main/java/org/springframework/security/web/csrf/DeferredCsrfToken.java
```diff
@@ -16,18 +16,22 @@
...
+
 /**
  * An interface that allows delayed access to a {@link CsrfToken} that may be generated.
  *
  * @author Rob Winch
+ * @author Steve Riesenberg
  * @since 5.8
  */
-public interface DeferredCsrfToken {
+interface DeferredCsrfToken extends Supplier<CsrfToken> {
```
sjohnr, line 28: Making this package-private is the third main change of this PR.

web/src/main/java/org/springframework/security/web/csrf/CsrfFilter.java
```diff
@@ -128,7 +121,8 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
 		CsrfToken csrfToken = deferredCsrfToken.get();
 		String actualToken = this.requestHandler.resolveCsrfTokenValue(request, csrfToken);
 		if (!equalsConstantTime(csrfToken.getToken(), actualToken)) {
-			boolean missingToken = deferredCsrfToken.isGenerated();
+			boolean missingToken = DeferredCsrfToken.class.isAssignableFrom(deferredCsrfToken.getClass())
```
rwinch, line 124: This is implementation specific as there is nothing in the interface that says that the `Supplier<CsrfToken>` will return a `DeferredCsrfToken` if it is generated.

With `DeferredCsrfToken` package private, it would be difficult for someone to properly implement a custom `CsrfTokenRepository` and overrride the `loadToken(request, response)` method since they cannot use `DeferredCsrfToken`.

I wonder if we should change the return type of the repository to `DeferredCsrfToken`

web/src/main/java/org/springframework/security/web/csrf/CsrfTokenRepository.java
```diff
@@ -55,4 +58,20 @@ public interface CsrfTokenRepository {
...
+	 * <p>
+	 * The returned {@code Supplier} is cached to allow subsequent calls to
+	 * {@link Supplier#get()} to return the same {@link CsrfToken} without the cost of
+	 * loading or generating the token again.
+	 * @param request the {@link HttpServletRequest} to use
+	 * @param response the {@link HttpServletResponse} to use
+	 * @return a {@code Supplier} that will load the {@link CsrfToken}
+	 * @since 5.8
+	 */
+	default Supplier<CsrfToken> loadToken(HttpServletRequest request, HttpServletResponse response) {
```
rwinch, line 73: consider renaming to loadDeferredToken

rwinch (changes requested):
I've provided feedback inline.

Additionally, I'm curious how this fits into the big picture of the breach protection

sjohnr:
Thanks for the feedback, @rwinch!

> Additionally, I'm curious how this fits into the big picture of the breach protection

The goal was to make a simpler API for customizing the request handling. The alternative we discussed would be to add a post-processor inside the `CsrfTokenRepositoryRequestHandler`. After experimenting with this a bit, I felt that I could simplify the API and wanted to give it a try and get some feedback.

With the change in this PR, support for breach fits back into the paradigm of a separate implementation of `CsrfTokenRequestHandler`. It is given the `Supplier<CsrfToken>`, which it can modify, and then pass to the delegate (or super) implementation from this PR. If we go this route, I won't need to make very many changes to the branch from the [draft PR](https://github.com/spring-projects/spring-security/pull/11731).

web/src/main/java/org/springframework/security/web/csrf/CsrfFilter.java
rwinch: I'd prefer to add `DeferredCsrfToken`. It differs a bit from `SecurityContextRepository` in that we need to know if the `CsrfToken` was generated or loaded and as such we should have an explicit API. The `GENERATED_CSRF_TOKEN_ATTR_NAME` is a new contract (much like having a new interface) but is more difficult for users to use/discover.

sjohnr:
Thanks @rwinch. I've updated the PR based on your feedback.

sjohnr:
Merged via 475b3bb6bbdc737ba67820d8ca4e557b3e030e7c

rwinch (approved):
Thanks @sjohnr This looks great. I think it has moved things along in a better direction than the work that I had done.

sjohnr:
Appreciate the kind words @rwinch! Thanks for all the help, support and guidance too.

Issue #11932, opened 2022-10-03
Add opt-in strategy in for Authentication(Web)Filter should return a 500 on AuthenticationServiceExceptions


Kehrlann:
Opt-in strategy for upcoming breaking change, see gh-9429

Issue #9429, opened 2021-02-10
Authentication(Web)Filter should return a 500 on AuthenticationServiceExceptions


jzheaux:
When `(Reactive)AuthenticationManager` throws or resolves to an `AuthenticationServiceException`, `Authentication(Web)Filter` should instead return a 500 HTTP response.

Kehrlann:
Hey there 👋 I'm interested in picking this one up.

Kehrlann:
Hey @jzheaux , a few questions regarding this.

Here's my interpretation of the story (for `AuthenticationFilter`):

### When the user does not override the `failureHandler`

When the `AuthenticationManager` throws an `AuthenticationServiceException` ;
Then the exception is re-thrown, ultimately producing a HTTP 500

### When the user provides a custom `failureHandler`

When the `AuthenticationManager` throws an `AuthenticationServiceException` ;
That exception is passed to the user's failureHandler and process by that

---

Let me know if I'm missing something.

PR #11933, opened 2022-10-03
Add (Server)AuthenticationEntryPointFailureHandlerAdapter

Kehrlann:
Issue gh-11932, gh-9429

(Server)AuthenticationEntryPointFailureHandler should produce HTTP 500 instead when an AuthenticationServiceException is thrown, instead of HTTP 401. This commit deprecates the current behavior and introduces an opt-in (Server)AuthenticationEntryPointFailureHandlerAdapter with the expected behavior.

It also deprecates
BearerTokenAuthenticationFilter#setAuthenticationEntryPoint in favor of BearerTokenAuthenticationFilter#setAuthenticationFailureHandler, so that BearerTokenAuthenticationFilter may use the newly introduced AuthenticationEntryPointRethrowingFailureHandler as a default failure handler in future versions.

### DIFF
--- a/etc/checkstyle/checkstyle-suppressions.xml
+++ b/etc/checkstyle/checkstyle-suppressions.xml
@@ -53,4 +53,7 @@
 	<suppress files="WithSecurityContextTestExecutionListenerTests\.java" checks="SpringMethodVisibility"/>
 	<suppress files="AbstractOAuth2AuthorizationGrantRequestEntityConverter\.java" checks="SpringMethodVisibility"/>
 	<suppress files="JoseHeader\.java" checks="SpringMethodVisibility"/>
+
+	<!-- Lambdas that we can't replace with a method reference because a closure is required -->
+	<suppress files="BearerTokenAuthenticationFilter\.java" checks="SpringLambda"/>
 </suppressions>
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/authentication/BearerTokenAuthenticationFilter.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/authentication/BearerTokenAuthenticationFilter.java
@@ -27,7 +27,6 @@ import org.springframework.core.log.LogMessage;
 import org.springframework.security.authentication.AuthenticationDetailsSource;
 import org.springframework.security.authentication.AuthenticationManager;
 import org.springframework.security.authentication.AuthenticationManagerResolver;
-import org.springframework.security.authentication.AuthenticationServiceException;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
 import org.springframework.security.core.context.SecurityContext;
@@ -40,6 +39,7 @@ import org.springframework.security.oauth2.server.resource.web.BearerTokenAuthen
 import org.springframework.security.oauth2.server.resource.web.BearerTokenResolver;
 import org.springframework.security.oauth2.server.resource.web.DefaultBearerTokenResolver;
 import org.springframework.security.web.AuthenticationEntryPoint;
+import org.springframework.security.web.authentication.AuthenticationEntryPointFailureHandlerAdapter;
 import org.springframework.security.web.authentication.AuthenticationFailureHandler;
 import org.springframework.security.web.authentication.WebAuthenticationDetailsSource;
 import org.springframework.security.web.context.NullSecurityContextRepository;
@@ -73,12 +73,12 @@ public class BearerTokenAuthenticationFilter extends OncePerRequestFilter {
 
 	private AuthenticationEntryPoint authenticationEntryPoint = new BearerTokenAuthenticationEntryPoint();
 
-	private AuthenticationFailureHandler authenticationFailureHandler = (request, response, exception) -> {
-		if (exception instanceof AuthenticationServiceException) {
-			throw exception;
-		}
-		this.authenticationEntryPoint.commence(request, response, exception);
-	};
+	private AuthenticationFailureHandler authenticationFailureHandler = new AuthenticationEntryPointFailureHandlerAdapter(
+			(request, response, authException) -> {
+				// This is a lambda and not a method reference so that the FailureHandler
+				// reflects entrypoint updates
+				this.authenticationEntryPoint.commence(request, response, authException);
+			});
 
 	private BearerTokenResolver bearerTokenResolver = new DefaultBearerTokenResolver();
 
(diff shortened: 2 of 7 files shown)


web/src/main/java/org/springframework/security/web/server/authentication/ServerAuthenticationEntryPointRethrowingFailureHandler.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Adapts a {@link ServerAuthenticationEntryPoint} into a
+ * {@link ServerAuthenticationFailureHandler}. When the failure is an
+ * {@link AuthenticationServiceException}, it re-throws, to produce an HTTP 500 error.
+ *
+ * @author Daniel Garnier-Moiroux
+ * @since 5.8
+ */
+public class ServerAuthenticationEntryPointRethrowingFailureHandler implements ServerAuthenticationFailureHandler {
```
jzheaux, line 35: Please make the class `final`. I realize that the deprecated version is open, but that was probably an oversight. We can open it later on if needed.

web/src/main/java/org/springframework/security/web/authentication/AuthenticationEntryPointRethrowingFailureHandler.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Adapts a {@link AuthenticationEntryPoint} into a {@link AuthenticationFailureHandler}.
+ * When the failure is an {@link AuthenticationServiceException}, it re-throws, to produce
+ * an HTTP 500 error.
+ *
+ * @author Daniel Garnier-Moiroux
+ * @since 5.8
+ */
+public class AuthenticationEntryPointRethrowingFailureHandler implements AuthenticationFailureHandler {
```
jzheaux, line 38: I don't like the names since, if the other implementation didn't already exist, I don't believe this one would be named this way. The fact that it rethrows `AuthenticationServiceException` doesn't seem such a defining characteristic so as to embed it in the name. Also, "rethrowing" isn't very descriptive in the end.

I think `AuthenticationEntryPointFailureHandlerAdapter` or `AuthenticationEntryPointAdaptingFailureHandler` would be better. There is precedent for `XXXAdapter` classes in Spring Security, like `ReactiveJwtAuthenticationConverterAdapter`.

jzheaux (changes requested):
Thanks, @Kehrlann! I've left my feedback inline.

web/src/main/java/org/springframework/security/web/authentication/AuthenticationEntryPointRethrowingFailureHandler.java
Kehrlann: Ah, thank you, I like names better as well.

jzheaux (changes requested):
Thanks, @Kehrlann! Sorry, I missed one item in my first review. I've left it as some inline feedback; however, I can apply as a polish, if needed.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -272,6 +300,9 @@ public void configure(H http) {
 		BearerTokenAuthenticationFilter filter = new BearerTokenAuthenticationFilter(resolver);
 		filter.setBearerTokenResolver(bearerTokenResolver);
 		filter.setAuthenticationEntryPoint(this.authenticationEntryPoint);
```
jzheaux, line 302: Since `setAuthenticationEntryPoint` is deprecated, I think it would be best to no longer call it.

What if the code instead did `filter.setAuthenticationFailureHandler(getAuthenticationFailureHandler())` where `getAuthenticationFailureHandler` is something like (pseudocode):

```java
if (authenticationFailureHandler != null) {
    return;
}
return new Adapter(authenticationEntryPoint);
```

config/src/main/java/org/springframework/security/config/http/OAuth2ResourceServerBeanDefinitionParser.java
```diff
@@ -139,6 +144,9 @@ public BeanDefinition parse(Element oauth2ResourceServer, ParserContext pc) {
 		filterBuilder.addConstructorArgValue(authenticationManagerResolver);
 		filterBuilder.addPropertyValue(BEARER_TOKEN_RESOLVER, bearerTokenResolver);
 		filterBuilder.addPropertyValue(AUTHENTICATION_ENTRY_POINT, authenticationEntryPoint);
```
jzheaux, line 146: Same comment about no longer calling deprecated methods.

jzheaux (changes requested):
Thanks, @Kehrlann! I've left a piece of feedback about the latest updates.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
Kehrlann: I agree that it'd be better, but I think we risk changing behavior. Notably because `BearerTokenAuthenticationFilter#doFilterInternal` explicitly calls `authenticationEntryPoint.commence(...)`, without using the `FailureHandler` yet, with this code:

```java
		String token;
		try {
			token = this.bearerTokenResolver.resolve(request);
		}
		catch (OAuth2AuthenticationException invalid) {
			this.logger.trace("Sending to authentication entry point since failed to resolve bearer token", invalid);
			this.authenticationEntryPoint.commence(request, response, invalid);
			return;
		}
		// ...
	}
```

So for the following FilterChain, the behavior would be different:

```
AuthenticationEntryPoint someEntryPoint = getEntryPoint();
http
	.authorizeRequests()
		.anyRequest().authenticated()
		.and()
	.oauth2ResourceServer()
		.authenticationEntryPoint(authenticationEntryPoint())
		.jwt();
```

Kehrlann: I also noticed that I marked `OAuth2ResourceServerConfigurer#authenticationEntryPoint` as deprecated, but the Entry Point is also used in `OAuth2ResourceServerConfigurer#registerDefaultEntryPoint`, so that deprecation is incorrect, we won't be able to "just" remove the field in 6.0 .

I'm unsure what the best course of action is.

jzheaux: I see. Sorry, I had missed that `BearerTokenAuthenticationFilter` uses `this.authenticationEntryPoint` directly.

Since `BearerTokenAuthenticationFilter` continues to use `authenticationEntryPoint`, then it can't be deprecated at this point. That means waiting on it using the new adapter as well, at least until there is time to refactor `BearerTokenAuthenticationFilter`.

At this point, I think it's best to remove the `BearerTokenAuthenticationFilter` work from this PR. I think that in down the road [the DSL will change to use `AuthenticationFilter`](https://github.com/spring-projects/spring-security/issues/11983), which would render the deprecation of `BearerTokenAuthenticationFilter#setAuthenticationEntryPoint` moot.

jzheaux (approved):

jzheaux:
Thanks, @Kehrlann! This has been merged into `5.8.x` in 200b7fecd347c29b4508351d4ee07cfc170888af. Also, good news, I was able to find a passive way to use a flag, so I adjusted the solution in 099aaa33ff3ad4151f338e110f5766547dca8a12.

Issue #11347, opened 2022-06-07
Simplify Java Configuration RequestMatcher Usage


rwinch:
By default, we should reject using the built in RequestMatcher implementations other than MvcRequestMatcher in a MvcRequestMatcher application.

Alternatively we can deprecate/remove antMatcher, regexMatchers, and mvcMatchers from the DSL and have a requestMatcher that is automatic based on the classpath. Users that want to opt out of this would need to use `requestMatchers(new RegExRequestMatcher(...))`. This makes it much more difficult to get the wrong configuration.

evgeniycheban:
I can take this.

rwinch:
@evgeniycheban Thank you for the volunteering. This is going to be a pretty tricky issue with quite a bit of design necessary. For that reason, I'm not sure if it is an ideal for contribution issue.

Issue #9159, opened 2020-10-28
Add 'filterChainMatcher' as an alias of 'requestMatcher'


akuma8:
When configuring security rules, it will be very helpful to understand what we are configuring. 
Example with this kind of configuration : 
```
@Configuration
public class UIResourceProtection extends WebSecurityConfigurerAdapter {

    @Override
    public void configure( HttpSecurity http ) throws Exception {

        http.requestMatchers().antMatchers("/product/**") // be more explicit here 
                .and()
                .authorizeRequests()
                .antMatchers(HttpMethod.PATCH,"/product/update")
                .hasRole("user")
                .antMatchers("/product/private/**").hasRole("super_user")
                .and()
                .oauth2ResourceServer()
                .jwt().jwtAuthenticationConverter( ... ) ;
    }
}
```
To help beginners to understand what they are doing, it will be helpful to name explicitly what we configure:

```
@Configuration
public class UIResourceProtection extends WebSecurityConfigurerAdapter {

    @Override
    public void configure( HttpSecurity http ) throws Exception {

        http.filterChainMatchers().antMatchers("/product/**") // with this "filterChainMatchers()" alias we know that we are defining the filter chain selector and its more explicit.
                .and()
                .authorizeRequests()
                .antMatchers(HttpMethod.PATCH,"/product/update")
                .hasRole("user")
                .antMatchers("/product/private/**").hasRole("super_user")
                .and()
                .oauth2ResourceServer()
                .jwt().jwtAuthenticationConverter( ... ) ;
    }
}
```
Spring Security is not easy to understand, naming things in a clear way could only simplify the life of users, especially beginners.

rwinch:
Are you speaking about making it more explicit between the matchers on `HttpSecurity` and the authorization rules?

akuma8:
> Are you speaking about making it more explicit between the matchers on `HttpSecurity` and the authorization rules?

Yes, in order to know exactly what we are configuring. Filter chains are the core of authorization rules and currently we do not explicitly know how one is declaratively selected by the framework. By **declaratively** I mean, by reading the security configuration code, the filter chain matcher should be obvious to us.

rwinch:
I think that makes sense and have thought about it a bit. I'd prefer not exposing implementation details like `filterChainMatchers` in the DSL. I think something like this would be what I would suggest:

```
http
    .httpSecurityMatcher(matcher)
    ...
```

Thoughts?

akuma8:
[...]

PR #11934, opened 2022-10-03
Simplify Java Configuration RequestMatcher Usage

marcusdacoregio:
If Spring MVC is present in the classpath, use MvcRequestMatcher by default. This commit also adds a new securityMatcher method in HttpSecurity

Closes gh-11347
Closes gh-9159


config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
```diff
@@ -3567,15 +3583,321 @@ public HttpSecurity requestMatchers(Customizer<RequestMatcherConfigurer> request
...
+	 * &#064;EnableWebSecurity
+	 * public class RequestMatchersSecurityConfig {
+	 *
+	 * 	&#064;Bean
+	 * 	public SecurityFilterChain securityFilterChain(HttpSecurity http) throws Exception {
+	 * 		http
+	 * 			.securityMatchers((matchers) -&gt; matchers
+	 * 				.requestMatchers(&quot;/api/**&quot;, &quot;/oauth/**&quot;)
+	 * 			)
+	 * 			.securityMatchers((authorize) -&gt; authorize
```
rwinch, line 3626: Should this be authorizeHttpRequests?

config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
```diff
@@ -3567,15 +3583,321 @@ public HttpSecurity requestMatchers(Customizer<RequestMatcherConfigurer> request
...
+	 * 				.requestMatchers(&quot;/**&quot;).hasRole(&quot;USER&quot;)
```
rwinch, line 3627: We should document any matcher that is `/**' to be `anyRequest()` instead

config/src/main/java/org/springframework/security/config/annotation/web/builders/HttpSecurity.java
```diff
@@ -3567,15 +3583,321 @@ public HttpSecurity requestMatchers(Customizer<RequestMatcherConfigurer> request
...
```
rwinch, line 3748: We should document any matcher that is `/**' to be `anyRequest()` instead

docs/modules/ROOT/pages/whats-new.adoc
```diff
@@ -10,3 +10,5 @@ Below are the highlights of the release.
 * https://github.com/spring-projects/spring-security/pull/11771[gh-11771] - `HttpSecurityDsl` should support `apply` method
 * https://github.com/spring-projects/spring-security/pull/11232[gh-11232] - `ClientRegistrations#rest` defines 30s connect and read timeouts
 * https://github.com/spring-projects/spring-security/pull/11464[gh-11464] - Remember Me supports SHA256 algorithm
+* https://github.com/spring-projects/spring-security/issues/11347[gh-11347] - Support RequestMatcher Validation
```
rwinch, line 13: I think this has evolved. Rather than validation we have changed the DSL to make it easier to make the right choice.

rwinch (changes requested):
Do we need to update the XML support to automatically default to MVC if HandlerMappingIntrospector is on the classpath?

It might be good to add static factory methods to AntRequestMatcher and RegexRequetsMatcher as well. If you desire, this can be done as a separate ticket.

marcusdacoregio:
Thanks @rwinch, I've updated the PR with the suggested changes.

rwinch (approved):

PR #11936, opened 2022-10-03
Add X-Xss-Protection headerValue to XML config

Kehrlann:
See gh-11908

### DIFF
--- a/config/src/main/resources/org/springframework/security/config/spring-security-5.8.rnc
+++ b/config/src/main/resources/org/springframework/security/config/spring-security-5.8.rnc
@@ -1298,6 +1298,9 @@ xss-protection.attlist &=
 xss-protection.attlist &=
 	## Add mode=block to the header or not, default is on.
 	attribute block {xsd:boolean}?
+xss-protection.attlist &=
+	## Specify the value for the X-Xss-Protection header. When set, overrides both enabled and block attributes.
+	attribute header-value {"0"|"1"|"1; mode=block"}?
 
 content-type-options =
 	## Add a X-Content-Type-Options header to the resopnse. Value is always 'nosniff'.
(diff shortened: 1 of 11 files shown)


config/src/main/resources/org/springframework/security/config/spring-security-5.8.rnc
```diff
@@ -1298,6 +1298,9 @@ xss-protection.attlist &=
 xss-protection.attlist &=
 	## Add mode=block to the header or not, default is on.
 	attribute block {xsd:boolean}?
+xss-protection.attlist &=
+	## Specify the value for the X-Xss-Protection header. When set, overrides both enabled and block attributes.
+	attribute header-value {"0","1","1\;mode=block"}?
```
sjohnr, line 1303: My IDE (IntelliJ) complains about this line, with

    group of "string" or "data" element

Is this the correct change to make?

sjohnr (changes requested):
Thanks @Kehrlann! See inline comment below.

config/src/main/resources/org/springframework/security/config/spring-security-5.8.rnc
Kehrlann: Good catch, thanks!

Changed to `"0" | "1" | "1; mode=block"`, it generates a very similar XSD but IntelliJ doesn't complain.

Note, I am no RELAX NG expert. I "made it work" but I don't really understand how...

sjohnr:
Merged via 0e215a21ad491534b6615760b3b02099d978c8fe. I regenerated the 5.8 xsd, as it was out of sync, and added c98de7af2f6328468ae3303a9320d19c1649a9ab in 6.0 for the rnc/xsd after the merge.

PR #11942, opened 2022-10-03
automatically manage docs version (no collector)

mojavelinux:
The purpose of this PR is to remove the hard-coded version in the docs component descriptor (docs/antora.yml). This variation does this without using Antora Collector. Instead, it creates a task dependency in the Gradle build and temporarily rewrites the original component descriptor for the duration of the task execution.


rwinch:
Closing this in favor of gh-11943

PR #11943, opened 2022-10-03
automatically manage docs version (with collector)

mojavelinux:
The purpose of this PR is to remove the hard-coded version in the docs component descriptor (docs/antora.yml). This variation does so using Antora Collector. This approach does not rely on temporary changes to the component descriptor.


rishiraj88:
Thanks

Issue #11899, opened 2022-09-26
Use MvcRequestMatcher by default if Spring MVC is present


marcusdacoregio:
In some places, like XML configuration (`AuthorizationFilterParser`, `FilterChainBeanDefinitionParser`, `MatcherType`), the default `RequestMatcher` implementation is `AntPathRequestMatcher`. We should switch the default to `MvcRequestMatcher` if Spring MVC is on the classpath.

See:
 - https://github.com/spring-projects/spring-security/issues/11347

PR #11950, opened 2022-10-04
Use MvcRequestMatcher by default if Spring MVC is present

marcusdacoregio:
Closes gh-11899


config/src/test/resources/org/springframework/security/config/http/OAuth2ResourceServerBeanDefinitionParserTests-JwtRestOperations.xml
```diff
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-  ~ Copyright 2002-2020 the original author or authors.
+  ~ Copyright 2002-2022 the original author or authors.
```
rwinch, line 3: This file doesn't seem to need to be changed.

rwinch (changes requested):
I had one comment around a file that doesn't look like it needs to change. Other than that, it appears to look good to me.

marcusdacoregio:
Thanks @rwinch, I've updated the PR.

rwinch (approved):

Issue #11938, opened 2022-10-03
Add static factory methods for AntPathRequestMatcher and RegexRequestMatcher


marcusdacoregio:
With [this PR](https://github.com/spring-projects/spring-security/pull/11934) merged, the DSL methods that create an `AntPathRequestMatcher` or a `RegexRequestMatcher` are deprecated. Instead, folks have to specify the actual `RequestMatcher` implementation that they want, something like `.requestMatcher(new AntPathRequestMatcher("/path"))`.

We should add a static factory method to the `RequestMatcher` implementations to make it easier to create instances of the matchers, something like:

```java
.requestMatcher(AntPathRequestMatcher.fromPattern("/path"));
```

PR #11965, opened 2022-10-06
Add static factory method to AntPathRequestMather and RegexRequestMather

marcusdacoregio:
Closes gh-11938

### DIFF
--- a/web/src/main/java/org/springframework/security/web/util/matcher/AntPathRequestMatcher.java
+++ b/web/src/main/java/org/springframework/security/web/util/matcher/AntPathRequestMatcher.java
@@ -69,6 +69,43 @@ public final class AntPathRequestMatcher implements RequestMatcher, RequestVaria
 
 	/**
 	 * Creates a matcher with the specific pattern which will match all HTTP methods in a
+	 * case-sensitive manner.
+	 * @param pattern the ant pattern to use for matching
+	 * @since 5.8
+	 */
+	public static AntPathRequestMatcher antMatcher(String pattern) {
+		Assert.hasText(pattern, "pattern cannot be empty");
+		return new AntPathRequestMatcher(pattern);
+	}
+
+	/**
+	 * Creates a matcher that will match all request with the supplied HTTP method in a
+	 * case-sensitive manner.
+	 * @param method the HTTP method. The {@code matches} method will return false if the
+	 * incoming request doesn't have the same method.
+	 * @since 5.8
+	 */
+	public static AntPathRequestMatcher antMatcher(HttpMethod method) {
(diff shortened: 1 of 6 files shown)


web/src/main/java/org/springframework/security/web/util/matcher/RegexRequestMatcher.java
```diff
@@ -53,6 +54,26 @@ public final class RegexRequestMatcher implements RequestMatcher {
...
+		Assert.hasText(pattern, "pattern cannot be empty");
+		return new RegexRequestMatcher(pattern, null);
+	}
+
+	/**
+	 * Creates an instance that matches to all requests with the same {@link HttpMethod}.
+	 * @param method the HTTP method to match. Must not be null.
+	 * @since 5.8
+	 */
+	public static RegexRequestMatcher forMethod(HttpMethod method) {
```
rwinch, line 72: Right now there is a problem if someone needs to do a static import for both AntPathRequestMatcher.forMethod and RegexRequestMatcher.forMethod the usage would need to be fully qualified. I'd say we should use antMatcher and regexMatcher for the method names.

web/src/main/java/org/springframework/security/web/util/matcher/AntPathRequestMatcher.java
```diff
@@ -67,6 +67,17 @@ public final class AntPathRequestMatcher implements RequestMatcher, RequestVaria
...
+	 * Creates a matcher that will match all request with the supplied HTTP method in a
+	 * case-sensitive manner.
+	 * @param method the HTTP method. The {@code matches} method will return false if the
+	 * incoming request doesn't have the same method.
+	 */
+	public static AntPathRequestMatcher forMethod(HttpMethod method) {
+		Assert.notNull(method, "method cannot be null");
+		return new AntPathRequestMatcher(MATCH_ALL, method.name());
+	}
+
```
rwinch, line 80: There doesn't seem to be a way to specify a pattern or a method and pattern via static factory method

web/src/main/java/org/springframework/security/web/util/matcher/RegexRequestMatcher.java
```diff
@@ -53,6 +54,26 @@ public final class RegexRequestMatcher implements RequestMatcher {
...
+		Assert.notNull(method, "method cannot be null");
+		return new RegexRequestMatcher(".*", method.name());
+	}
+
```
rwinch, line 76: There doesn't seem to be a static factory method for both the method and pattern.

rwinch (changes requested):

PR #12073, opened 2022-10-22
Add configurable authorities split regex

PatrickWalter214:
This PR adds the ability to split the authorities claim on a configurable regex, which was previously set ot a hardcoded regex " ". The IAM solution I used provided the authorities separated by another character than ' ' and I realized that with the JwtGrantedAuthoritiesConverter I was not able to set a custom regex. The solution in this PR makes the JwtGrantedAuthoritiesConverter more flexible by supporting this use case while keeping the previous default behaviour.

I signed the individual CLA to provide this PR.

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverter.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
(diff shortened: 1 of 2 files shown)


jzheaux:
Thanks for the PR, @PatrickWalter214!

Since this is a new feature, it needs to go into the next minor release, which is 6.1. Will you please:

1. Rebase the PR with `main` so it gets merged to the correct version
2. Change the `@since` to be `6.1`

PatrickWalter214:
Thank you for reviewing my PR @jzheaux!

I performed the steps you mentioned with my latest commit.

jzheaux:
Thanks, @PatrickWalter214! This is now merged into `main`.

jzheaux:
Oops, @PatrickWalter214, my muscle memory got the best of me. :sweat_smile:  This is not mergeable until after the 6.0 release. I've created [a new PR](https://github.com/spring-projects/spring-security/pull/12124) with your commit and will merge it after 6.0 releases.

Issue #12081, opened 2022-10-25
Update the RP-initiated Logout links


rozagerardo:
**Describe the bug**
We're using the following link when we refer to the OIDC RP-Initiated Logout feature:

https://openid.net/specs/openid-connect-session-1_0.html#RPLogout

But it seems this spec is not related to the feature anymore, in fact, the #RPLogout fragment doesn't seem to correspond to any section/element in the document.

This seems to be the correct link to the RP-Initiated spec:

https://openid.net/specs/openid-connect-rpinitiated-1_0.html 

**To Reproduce**
Browse the following section in our docs:

https://docs.spring.io/spring-security/reference/servlet/oauth2/login/advanced.html#oauth2login-advanced-oidc-logout

and click on the following link:

> One of the strategies available is [RP-Initiated Logout](https://openid.net/specs/openid-connect-session-1_0.html#RPLogout).

The link will redirect to an OIDC spec that doesn't contain much RP-Initiated Logout info, which might be a little misleading.

**Expected behavior**
The links should point to the correct OIDC Spec:

https://openid.net/specs/openid-connect-rpinitiated-1_0.html 

**Sample**
N/A

rozagerardo:
Here is a small PR to fix this, in case you agree the links should be updated:

https://github.com/spring-projects/spring-security/pull/12082

PR #12082, opened 2022-10-25
Update RP-Initiated Logout target URLs.

rozagerardo:
The URLs we're using are not actually pointing to the OIDC RP-Initiated Logout Specs.

Fixes: gh-12081


jzheaux:
Hi, @rozagerardo! Thanks for the PR. Will you please rebase in on the earliest branch that has the issue? We've recently moved to forward-porting fixes from the earliest supported branch, which is `5.6.x` in this case.

rozagerardo:
Sure @jzheaux , sorry about that,  I missed the guidelines indicating to create the branch from the oldest maintenance branch :)

I now fixed that, re-formated the files as required, and edited the PR to point to `5.6.x` 👍 

Just to be clear I understood this correctly, even though we started using the links since `5.2.x`, we're pointing to `5.6.x` because we're currently maintaining versions from `5.6.x` and above (as per the Milestones that are being actively developed) correct?

Cool, hope the PR looks good now, thanks for the patience!

jzheaux:
Yes, @rozagerardo, that's right.

I've just merged this to `5.6.x` and will update here once it's merged to the other branches. Thanks again!

PR #12099, opened 2022-10-28
Document how to opt-in for SHA256 in RememberMe

marcusdacoregio:
Closes gh-12097

### DIFF
--- a/docs/modules/ROOT/pages/migration.adoc
+++ b/docs/modules/ROOT/pages/migration.adoc
@@ -1703,6 +1703,167 @@ authenticationFilter.setAuthenticationFailureHandler(handler)
 ----
 ====
 
+[[servlet-opt-in-sha256-rememberme]]
+=== Use SHA-256 in Remember Me
+
+The `TokenBasedRememberMeServices` implementation now supports SHA-256 for the Remember Me token and this is the default in Spring Security 6.
+This change makes the implementation more secure by default since MD5 is already proven to be a weak hashing algorithm and vulnerable against collision attacks and modular differential attacks.
+
+The new generated tokens now have the information of which algorithm was used to generate the token and that information is used in order to match it.
+If the algorithm name is not present, then the `matchingAlgorithm` property is used to check the token.
+This allows for a smooth transition from MD5 to SHA-256.
+
+To opt into the new Spring Security 6 default to encode the tokens while still being able to decode tokens encoded with MD5, you can set the `encodingAlgorithm` property to SHA-256 and the `matchingAlgorithm` property to MD5.
+See the xref:servlet/authentication/rememberme.adoc#_tokenbasedremembermeservices[reference documentation] and the {security-api-url}org/springframework/security/web/authentication/rememberme/TokenBasedRememberMeServices.html[API docs] for more information.
+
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/migration.adoc
```diff
@@ -479,6 +479,13 @@ class WebSocketSecurityConfig: WebSocketMessageBrokerConfigurer {
 ----
 ====
 
+[[servlet-opt-in-sha256-rememberme]]
+==== Use SHA-256 in Remember Me
+
+The `TokenBasedRememberMeServices` implementation now supports SHA-256 for the Remember Me token and this is the default in Spring Security 6.
+You can start using it in 5.8 by setting the `encodingAlgorithm` property to SHA-256 while still decoding tokens that are using MD5.
+See the documentation for xref:servlet/authentication/rememberme.adoc#_tokenbasedremembermeservices[TokenBasedRememberMeServices] for more information.
+
```
jzheaux, line 488: Please show sample configuration inline. You can include a partial like "Optimize Querying of `RequestCache`" does, or you can do them inline like "Use `AuthorizationManager` for Method Security".

It would be helpful to show a before and after where applicable as this can help the Spring Boot migrator team use this as a reference to build recipes. The diff can also be instructive visually for the reader.

docs/modules/ROOT/pages/migration.adoc
```diff
@@ -479,6 +479,13 @@ class WebSocketSecurityConfig: WebSocketMessageBrokerConfigurer {
 ----
 ====
 
+[[servlet-opt-in-sha256-rememberme]]
+==== Use SHA-256 in Remember Me
+
+The `TokenBasedRememberMeServices` implementation now supports SHA-256 for the Remember Me token and this is the default in Spring Security 6.
```
jzheaux, line 485: I think it would be helpful to tell the reader how to declaratively maintain the 5.8 default, just in case moving to SHA-256 breaks things.

jzheaux, line 485: I think it would be helpful to explain why this change was made. Moving from MD5 to SHA-256 is obvious to you and I, but not everyone lives in the security space on a day-to-day basis. I'd strongly recommend at least explaining that MD5 is an insecure default.

docs/modules/ROOT/pages/migration.adoc
```diff
@@ -479,6 +479,13 @@ class WebSocketSecurityConfig: WebSocketMessageBrokerConfigurer {
...
+You can start using it in 5.8 by setting the `encodingAlgorithm` property to SHA-256 while still decoding tokens that are using MD5.
```
jzheaux, line 486: It may be helpful to demonstrate how an application will manage existing remember me tokens and what the migration steps for getting to `SHA256` for both matching and encoding will look like. Is there a way for us to guide them through that process?

jzheaux (changes requested):
Nice, @marcusdacoregio! I've left my feedback inline.

jzheaux (changes requested):
Much improved, @marcusdacoregio! I've left some additional feedback inline.

jzheaux (approved):
Nice!

PR #12124, opened 2022-11-01
Add configurable authorities split regex

jzheaux:
Before this commit splitting the authorities claim was done by a hardcoded regex " ". This commit allows to configure to set any regex to split the authorities claim while keeping the previously hardcoded regex as a default.


jzheaux:
Thanks, @PatrickWalter214! This is now merged into `main` in 9c9fd9f4bd997461967b27bdeb4a5eca85580467 with a slight polish on the name cd0f02de49248a0fd0a2021201034a5f6c680417

PatrickWalter214:
Thank you @jzheaux for considering and taking up my proposed change, I'm happy that it reached `main`.

Issue #12129, opened 2022-11-02
SAML logout: Incorrect log messages


palakova:
**Describe the bug**
There are following incorrect log messages, referring to Response instead of Request and vice versa:

OpenSamlLogoutRequestValidator
- "Failed to find issuer in **LogoutResponse**"
- "Failed to find destination in **LogoutResponse**"

Saml2LogoutResponseFilter
- "Did not process logout **request** since failed to find associated RelyingPartyRegistration"

PR #12130, opened 2022-11-02
Fix SAML logout log messages

palakova:
Fixes SAML logout log messages incorrectly referring to logout response instead of logout request and vice versa.

Closes gh-12129


jzheaux:
Thanks, @palakova! This is now merged into `5.6.x`, `5.7.x`, `5.8.x`, and `main`.

Issue #12100, opened 2022-10-28
Document how to use the new `requestMatchers` and `securityMatchers`


marcusdacoregio:
See:
 - #11899 
 - #11939 
 - https://github.com/spring-projects/spring-security/issues/11347

craph:
@marcusdacoregio what is the difference between `requestMatcher` and `securityMatcher` ?

PR #12137, opened 2022-11-03
Document how to use the new requestMatchers and securityMatchers

marcusdacoregio:
Closes gh-12100

### DIFF
--- a/docs/modules/ROOT/pages/migration.adoc
+++ b/docs/modules/ROOT/pages/migration.adoc
@@ -2188,6 +2188,423 @@ versionFour.setAssertionValidator(validator)
 ----
 ====
 
+[[use-new-requestmatchers]]
+=== Use the new `requestMatchers` methods
+
+In Spring Security 5.8, the {security-api-url}org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.html#antMatchers(java.lang.String...)[`antMatchers`], {security-api-url}org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.html#mvcMatchers(java.lang.String...)`mvcMatchers`, and {security-api-url}org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.html#regexMatchers(java.lang.String...)[`regexMatchers`] methods were deprecated in favor of new xref::servlet/authorization/authorize-http-requests.adoc#_request_matchers[`requestMatchers` methods].
+
+The new `requestMatchers` methods were added xref::servlet/authorization/authorize-http-requests.adoc[to `authorizeHttpRequests`], `authorizeRequests`, CSRF configuration, `WebSecurityCustomizer` and any other places that had the specialized `RequestMatcher` methods.
+The deprecated methods are removed in Spring Security 6.
+
+These new methods have more secure defaults since they choose the most appropriate `RequestMatcher` implementation for your application.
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/migration.adoc
```diff
@@ -1973,6 +1973,423 @@ Saml2AuthenticationRequestResolver authenticationRequestResolver() {
 Since Spring Security only supports the `POST` binding for authentication, there is not very much value in overriding the protocol binding at this point in time.
 ====
 
+[[use-new-requestmatchers]]
+==== Use the new `requestMatchers` methods
```
jzheaux, line 1977: Because this has four equals signs, I think it will appear under the SAML deprecations, which seems like an oversight.

docs/modules/ROOT/pages/migration.adoc
```diff
@@ -1973,6 +1973,423 @@ Saml2AuthenticationRequestResolver authenticationRequestResolver() {
...
+                .anyRequest().authenticated()
+            );
+        return http.build();
+    }
+
+}
+----
+====
+
+Note that we are using some static factory methods to make the code more readable.
```
jzheaux, line 2171: We usually avoid "we" in the docs. Something like "Note that the above sample uses static factory methods to improve readability."

You might also consider linking parts of "some static factory methods" to the JavaDoc for those factory methods.

docs/modules/ROOT/pages/migration.adoc
```diff
@@ -1973,6 +1973,423 @@ Saml2AuthenticationRequestResolver authenticationRequestResolver() {
...
+        .authorizeHttpRequests((authz) -> authz
+            .requestMatchers("/api/admin/**").hasRole("ADMIN")
+            .anyRequest().authenticated()
+        );
+    return http.build();
+}
+----
+====
+
+this can be rewritten using the `securityMatchers` methods:
```
jzheaux, line 2262: Small grammar nitpick. If this is a continuation of the previous sentence, it should be lowercased as it is. If not, it should be capitalized. This does not read like a continuation: "Consider the following configuration this can be rewritten ...". "Consider the following configuration" and "This can be rewritten..." make more sense as separate sentences.

docs/modules/ROOT/pages/migration.adoc
```diff
@@ -1973,6 +1973,423 @@ Saml2AuthenticationRequestResolver authenticationRequestResolver() {
...
+                .anyRequest().authenticated()
+            );
+        return http.build();
+    }
+
+}
+----
+====
+
+is equivalent too:
```
jzheaux, line 2063: I think you mean "is equivalent to"

docs/modules/ROOT/pages/migration.adoc
```diff
@@ -1973,6 +1973,423 @@ Saml2AuthenticationRequestResolver authenticationRequestResolver() {
...
+[source,java,role="primary"]
+----
+@Bean
+public WebSecurityCustomizer webSecurityCustomizer() {
+	return (web) -> web.ignoring().antMatchers("/ignore1", "/ignore2");
+}
+----
+====
+
+can be changed to:
```
jzheaux, line 2186: This reads a little funny as a continuation: "you can replace the deprecated methods with the requestMatchers methods can be changed to ...". It might read more clearly as "You can replace the deprecated antMatchers methods:" ... "with their requestMatchers counterparts:"

jzheaux (changes requested):
Great, @marcusdacoregio! I've left feedback inline.

Issue #11939, opened 2022-10-03
Remove deprecated RequestMatcher methods from Java Configuration


marcusdacoregio:
https://github.com/spring-projects/spring-security/pull/11934 deprecated some methods in order to enforce the right `RequestMatcher` usage. We should remove those deprecated methods to avoid misconfiguration.

PR #12150, opened 2022-11-08
Remove antMatcher usage from Multiple HttpSecurity docs

grafjo:
This is a change based on 6.0.0-RC1 release documentation.

`http.antMatcher()` is not longer available and was replaced with `http.securityMatcher()`, so use this in the Java Config Multiple HttpSecurity Instances example, too.

That part of documentation was missing in https://github.com/spring-projects/spring-security/issues/11939


marcusdacoregio:
Thanks @grafjo, this is now merged into `main`.

Issue #12144, opened 2022-11-07
Incorrect scope map fix


DamianFekete:
**Describe the bug**
Scope mapping handling changed with https://github.com/spring-projects/spring-security/issues/12112.

https://github.com/spring-projects/spring-security/commit/2915a70bf786e2bd0780d686d432b9ba85617522#diff-73bd44f873d78e3d71e6a0fa18644a304d562a4a9fd2e303e913f6ed20a0ad16R78-R83

![image](https://user-images.githubusercontent.com/150717/200307800-4eb81fe1-da5a-4f2b-a884-aa8ddc075958.png)

- `OidcAuthorizationCodeAuthenticationProvider.authenticate()` calls `OidcAuthorizationCodeAuthenticationProvider.getResponse()`)
- `DefaultAuthorizationCodeTokenResponseClient.getTokenResponse` does NOT add the scopes anymore. If no scopes are returned by default by the IdP, the scopes list is empty.
   * The comment `If AccessTokenResponse.scope is empty, then we assume all requested scopes were granted.` seems to say something completely different.
- Back in `OidcAuthorizationCodeAuthenticationProvider.authenticate()` the user info has to be loaded: `this.userService.loadUser()`.
- `OidcUserService.loadUser` calls [`this.shouldRetrieveUserInfo(userRequest)`](https://github.com/spring-projects/spring-security/blob/006b9b960797d279b31cf8c8d16f1549c5632b2c/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserService.java#L177) which returns `false` now, because the scopes (`userRequest.getAccessToken().getScopes()`) is empty.
- Because of this the userInfo is not loaded (it is `null`) and can't be used for example in the `userAuthoritiesMapper`.

**To Reproduce**
Our scopes are configured like this:
```
spring.security.oauth2.client.registration.default.scope=openid,profile,entitlements
```

Use  the `userAuthoritiesMapper` with a token-uri endpoint that doesn't return a list of scopes.
```
        http.oauth2Login()
                .userInfoEndpoint().userAuthoritiesMapper(this.userAuthoritiesMapper());
```

In the authorities mapper try to use the `oidcUserAuthority.getUserInfo()` (which is now null).

**Expected behavior**
`oidcUserAuthority.getUserInfo()` should not be `null`.

**Sample**
No sample yet.

Ping @sjohnr

tobiaskrauss:
[...]

PR #12160, opened 2022-11-08
Request user info when AS returns no scopes

sjohnr:
Closes gh-12144


sjohnr:
(Will rebase on older branch shortly)

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserService.java
```diff
@@ -173,8 +173,15 @@ private boolean shouldRetrieveUserInfo(OidcUserRequest userRequest) {
 				.equals(userRequest.getClientRegistration().getAuthorizationGrantType())) {
 			// Return true if there is at least one match between the authorized scope(s)
 			// and accessible scope(s)
+			// As per spec, in Section 5.1 Successful Access Token Response
+			// https://www.rfc-editor.org/rfc/rfc6749#section-5.1
```
jgrandja, line 177: This comment does not make sense in this context. I think we should be more specific.

sjohnr:
Merged via fde26e003a648a670be40c690a028970f0d1137b

PR #12162, opened 2022-11-08
configure Antora and CI workflow to support partial build

mojavelinux:
- upgrade to Antora 3.2.0-alpha.2
- upgrade release-line-extension to 1.0.0-alpha.2
- add atlas-extension
- add local autoconfig-partial-build extension to activate partial build
- configure deploy-docs.yml workflow to detect and set up partial build
- configure publish-docs.sh script to sync partial build

The deploy-docs workflow now accepts an input which will activate a partial build. The input is the refname that you want to build. You can test this using the following `gh` call:

```
gh workflow --repo spring-projects/spring-security run deploy-docs.yml --ref docs-build -f build-refname=5.7.x
```

(the `--repo` is optional if the current directory is inside a clone of the repository)

As of now, the partial build will not be published. The sync script is configured to do a dry run in this case so we have a chance to verify its behavior in this scenario.

### DIFF
--- a/.github/actions/publish-docs.sh
+++ b/.github/actions/publish-docs.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+FROM=build/site
 HOST="$1"
 HOST_PATH="$2"
 SSH_PRIVATE_KEY="$3"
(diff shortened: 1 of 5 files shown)


mojavelinux:
I think I have a simpler way to configure rsync in the case of a partial build. I'm still exploring experimenting with that approach. But we can at least see how it works now.

mojavelinux:
Keep in mind that a partial build means build a single version. That version must have already been built in a full build. If it was not previously build, Antora automatically reverts to a full build.

mojavelinux:
> I'm still exploring experimenting with that approach.

This is now done. It intelligently syncs the files built while ignoring directories it did not build.

.github/actions/publish-docs.sh
```diff
@@ -13,10 +14,16 @@ fi
 
 (
   set -e
+  set -f
```
mojavelinux, line 17: `set -f` turns off pathname expansion in bash, which makes using rsync a lot simpler.

antora-playbook.yml
```diff
@@ -1,8 +1,10 @@
 antora:
   extensions:
+  - ./lib/antora/extensions/autoconfig-partial-build.js
```
mojavelinux, line 3: This extension detects when it's the appropriate time to do a partial build and reconfigures Atlas to accordingly.

rwinch (approved):

PR #12238, opened 2022-11-18
NimbusJwtDecoder unknown KID scenario is not correctly tested

jonkjenn:
Two different methods are used for accessing the cache and only one of them is correctly mocked. This means the cache will always refresh unrelated to the unknown KID. https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java#L415
The `RemoteJWKSet` will never actually find a missing KID and do the update because of this. The code seems to be correct though and the test succeeds with the fixed mock.

Relates to this issue https://github.com/spring-projects/spring-security/issues/11621


jzheaux:
Thanks, @jonkjenn! Will you please rebase to `5.7.x` and squash your commits?

Please also add a brief description to the commit like:

```
Polish NimbusJwtDecoderTests

- Add missing mock

Closes gh-12238
 ```

jonkjenn:
@jzheaux Updated

PR #12266, opened 2022-11-22
upgrade Antora and Antora Collector to latest alpha; remove patch to support linked worktrees (5.8.x)

mojavelinux:


sjohnr:
Thanks @mojavelinux! This is now merged via fe252f5057377a1cdf8ef0efdee0a707be646590

This was applied via a normal forward merge from 5.7.x and then I inspected the result to ensure it matched the contents of this PR. Please let me know if I did anything incorrectly, but I believe we should be in sync now.

PR #12269, opened 2022-11-22
Fix: Replace tenantRepository with tenants

selllami:
Fix issue : https://docs.spring.io/spring-security/reference/servlet/oauth2/resource-server/multitenancy.html

Details: 
Replace `this.tenantRepository` with `this.tenants`


marcusdacoregio:
Thanks @selllami, this was merged via https://github.com/spring-projects/spring-security/commit/626e53d1214b5b6a639648a7c0bd8c44744d50bf

Issue #12275, opened 2022-11-23
Wrong username attribute when mapping authorities with delegate OidcUserService


daniel-shuy:
**Describe the bug**
When using the Delegation-based Strategy with `OidcUserService` as documented at https://docs.spring.io/spring-security/reference/6.0/servlet/oauth2/login/advanced.html#oauth2login-advanced-map-authorities-oauth2userservice, and the `spring.security.oauth2.client.provider.<provider>.user-name-attribute` is set, the username retrieved by `SecurityContextHolder.getContext().getAuthentication().getName()` is wrong.

This is because `DefaultOidcUser` has the following constructors:
- `DefaultOidcUser(Collection<? extends GrantedAuthority> authorities, OidcIdToken idToken, OidcUserInfo userInfo)`
- `DefaultOidcUser(Collection<? extends GrantedAuthority> authorities, OidcIdToken idToken, OidcUserInfo userInfo, String nameAttributeKey)`
where the first calls the second, with the `nameAtrributeKey` defaulting to `sub`.

The authority mapping replaces the authorities, but copies over the existing `idToken` and `userInfo`. It is however unable to copy over the `nameAttributeKey` because the [OidcUser](https://docs.spring.io/spring-security/site/docs/6.0.0/api/org/springframework/security/oauth2/core/oidc/user/OidcUser.html) interface does not have a getter for `nameAttributeKey`.

Perhaps a builder could be added to `DefaultOidcUser` that copies the values over from an existing `OidcUser` (including the `nameAttributeKey`). This avoids having to change the `OidcUser` interface.

**To Reproduce**
1. Configure an OIDC client with `spring-boot-starter-oauth2-client`.
2. Configure a custom `OAuth2UserService` to map authorities. For the simplest example to reproduce this issue:
```java
@Configuration
@EnableWebSecurity
public class OAuth2LoginSecurityConfig {

	@Bean
	public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
		http
			.oauth2Login(oauth2 -> oauth2
			    .userInfoEndpoint(userInfo -> userInfo
			        .oidcUserService(this.oidcUserService())
			        ...
			    )
			);
		return http.build();
	}

	private OAuth2UserService<OidcUserRequest, OidcUser> oidcUserService() {
		final OidcUserService delegate = new OidcUserService();

		return (userRequest) -> {
			OidcUser oidcUser = delegate.loadUser(userRequest);

[...]

PR #12281, opened 2022-11-24
Fix Delegation-based Strategy with OidcUserService/OidcReactiveOAuth2UserService examples

daniel-shuy:
The examples for Delegation-based Strategy with `OidcUserService`/`OidcReactiveOAuth2UserService` is bugged. If the `userNameAttributeName` is overridden with `spring.security.oauth2.client.provider.<provider>.user-name-attribute`, `SecurityContextHolder.getContext().getAuthentication().getName()` will return the value of `sub` (the default username attribute) regardless of what is set.

This fixes the example by obtaining the username attribute from `OidcUserRequest` and passing it to the new `DefaultOidcUser`.

See gh-12275 for the original discussion.

### DIFF
--- a/docs/modules/ROOT/pages/reactive/oauth2/login/advanced.adoc
+++ b/docs/modules/ROOT/pages/reactive/oauth2/login/advanced.adoc
@@ -453,7 +453,13 @@ public class OAuth2LoginSecurityConfig {
 						// 2) Map the authority information to one or more GrantedAuthority's and add it to mappedAuthorities
 
 						// 3) Create a copy of oidcUser but use the mappedAuthorities instead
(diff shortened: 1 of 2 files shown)


sjohnr:
@daniel-shuy just an update that I will revisit this in the coming week to get this merged, as I think improving the docs will help. See also [this comment](https://github.com/spring-projects/spring-security/pull/12282#issuecomment-1973625964) regarding the related enhancement.

sjohnr:
Thanks @daniel-shuy! This is now merged via 488901dd4e17544c88dff95e1b41b6fd23e5bcf3 with minor polish 9751636735d240ca1dbb3e0593e2302b7b8f95cb.

PR #12282, opened 2022-11-24
Improve API for delegation-based strategy

daniel-shuy:
The current API for delegation-based strategy with [OidcUserService](https://docs.spring.io/spring-security/reference/6.0/servlet/oauth2/login/advanced.html#oauth2login-advanced-map-authorities-oauth2userservice)/[OidcReactiveOAuth2UserService](https://docs.spring.io/spring-security/reference/6.0/reactive/oauth2/login/advanced.html#webflux-oauth2-login-advanced-map-authorities-reactiveoauth2userservice) is error-prone because it requires understanding the inner workings of [OidcUserService](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserService.java)/[OidcReactiveOAuth2UserService](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcReactiveOAuth2UserService.java) (see gh-12275 and gh-12281 for the original discussion).

This improves the API by:
- Making `OidcUserService#getUser(OidcUserRequest, OidcUserInfo, Set)` `protected`
- `OidcReactiveOAuth2UserService`: Abstract out logic to set `DefaultOidcUser` `nameAttributeKey` from `OidcUserRequest` into a new `protected` `getUser(OidcUserRequest, OidcUserInfo, Set)` method

This allows the delegation-based strategy to be massively simplified (especially for reactive).

Using the example in the doc, the configuration can be simplified from:
```java
@Bean
public ReactiveOAuth2UserService<OidcUserRequest, OidcUser> oidcUserService() {
	final OidcReactiveOAuth2UserService delegate = new OidcReactiveOAuth2UserService();

	return (userRequest) -> {
		// Delegate to the default implementation for loading a user
		return delegate.loadUser(userRequest)
				.flatMap((oidcUser) -> {
					OAuth2AccessToken accessToken = userRequest.getAccessToken();
					Set<GrantedAuthority> mappedAuthorities = new HashSet<>();

					// TODO
					// 1) Fetch the authority information from the protected resource using accessToken
					// 2) Map the authority information to one or more GrantedAuthority's and add it to mappedAuthorities

					// 3) Create a copy of oidcUser but use the mappedAuthorities instead
					ProviderDetails providerDetails = userRequest.getClientRegistration().getProviderDetails();
					String userNameAttributeName = providerDetails.getUserInfoEndpoint().getUserNameAttributeName();
					if (StringUtils.hasText(userNameAttributeName)) {
						oidcUser = new DefaultOidcUser(mappedAuthorities, oidcUser.getIdToken(), oidcUser.getUserInfo(), userNameAttributeName);
					} else {
						oidcUser = new DefaultOidcUser(mappedAuthorities, oidcUser.getIdToken(), oidcUser.getUserInfo());
					}

					return Mono.just(oidcUser);
				});
	};
}
```
to:
```java
@Bean
public ReactiveOAuth2UserService<OidcUserRequest, OidcUser> oidcUserService() {
	return new OidcReactiveOAuth2UserService() {
		@Override
		protected OidcUser getUser(OidcUserRequest userRequest, OidcUserInfo userInfo, Set<GrantedAuthority> authorities) {
			OAuth2AccessToken accessToken = userRequest.getAccessToken();
			Set<GrantedAuthority> mappedAuthorities = new HashSet<>();

			// TODO
			// 1) Fetch the authority information from the protected resource using accessToken
			// 2) Map the authority information to one or more GrantedAuthority's and add it to mappedAuthorities

			// Delegate to the default implementation for getting a user
			OidcUser oidcUser = super.loadUser(userRequest, userInfo, mappedAuthorities);

			return oidcUser;
		}
	};
}
```

### DIFF
--- a/docs/modules/ROOT/pages/reactive/oauth2/login/advanced.adoc
+++ b/docs/modules/ROOT/pages/reactive/oauth2/login/advanced.adoc
@@ -439,24 +439,21 @@ public class OAuth2LoginSecurityConfig {
 
 	@Bean
 	public ReactiveOAuth2UserService<OidcUserRequest, OidcUser> oidcUserService() {
-		final OidcReactiveOAuth2UserService delegate = new OidcReactiveOAuth2UserService();
+		return new OidcReactiveOAuth2UserService() {
+			@Override
+			protected OidcUser getUser(OidcUserRequest userRequest, OidcUserInfo userInfo, Set<GrantedAuthority> authorities) {
+				OAuth2AccessToken accessToken = userRequest.getAccessToken();
+				Set<GrantedAuthority> mappedAuthorities = new HashSet<>();
 
-		return (userRequest) -> {
-			// Delegate to the default implementation for loading a user
-			return delegate.loadUser(userRequest)
-					.flatMap((oidcUser) -> {
-						OAuth2AccessToken accessToken = userRequest.getAccessToken();
-						Set<GrantedAuthority> mappedAuthorities = new HashSet<>();
+				// TODO
+				// 1) Fetch the authority information from the protected resource using accessToken
+				// 2) Map the authority information to one or more GrantedAuthority's and add it to mappedAuthorities
 
-						// TODO
-						// 1) Fetch the authority information from the protected resource using accessToken
-						// 2) Map the authority information to one or more GrantedAuthority's and add it to mappedAuthorities
+				// Delegate to the default implementation for getting a user
+				OidcUser oidcUser = super.loadUser(userRequest, userInfo, mappedAuthorities);
 
-						// 3) Create a copy of oidcUser but use the mappedAuthorities instead
-						oidcUser = new DefaultOidcUser(mappedAuthorities, oidcUser.getIdToken(), oidcUser.getUserInfo());
-
-						return Mono.just(oidcUser);
-					});
+				return oidcUser;
+			}
 		};
 	}
 }
@@ -479,24 +476,19 @@ class OAuth2LoginSecurityConfig {
     }
 
     @Bean
-    fun oidcUserService(): ReactiveOAuth2UserService<OidcUserRequest, OidcUser> {
-        val delegate = OidcReactiveOAuth2UserService()
+    fun oidcUserService(): ReactiveOAuth2UserService<OidcUserRequest, OidcUser> = object : OidcReactiveOAuth2UserService() {
+        override fun getUser(userRequest: OidcUserRequest, userInfo: OidcUserInfo, authorities: Set<GrantedAuthority>): OidcUser {
+            val accessToken = userRequest.accessToken
+            val mappedAuthorities = mutableSetOf<GrantedAuthority>()
 
-        return ReactiveOAuth2UserService { userRequest ->
-            // Delegate to the default implementation for loading a user
-            delegate.loadUser(userRequest)
-                .flatMap { oidcUser ->
-                    val accessToken = userRequest.accessToken
-                    val mappedAuthorities = mutableSetOf<GrantedAuthority>()
+            // TODO
+            // 1) Fetch the authority information from the protected resource using accessToken
+            // 2) Map the authority information to one or more GrantedAuthority's and add it to mappedAuthorities
 
-                    // TODO
-                    // 1) Fetch the authority information from the protected resource using accessToken
-                    // 2) Map the authority information to one or more GrantedAuthority's and add it to mappedAuthorities
-                    // 3) Create a copy of oidcUser but use the mappedAuthorities instead
-                    val mappedOidcUser = DefaultOidcUser(mappedAuthorities, oidcUser.idToken, oidcUser.userInfo)
+            // Delegate to the default implementation for getting a user
+            val mappedOidcUser = super.getUser(userRequest, userInfo, mappedAuthorities)
 
-                    Mono.just(mappedOidcUser)
-                }
+            mappedOidcUser
         }
     }
 }
--- a/docs/modules/ROOT/pages/servlet/oauth2/login/advanced.adoc
+++ b/docs/modules/ROOT/pages/servlet/oauth2/login/advanced.adoc
@@ -626,23 +626,21 @@ public class OAuth2LoginSecurityConfig {
 	}
 
 	private OAuth2UserService<OidcUserRequest, OidcUser> oidcUserService() {
-		final OidcUserService delegate = new OidcUserService();
+		return new OidcUserService() {
+			@Override
+			protected OidcUser getUser(OidcUserRequest userRequest, OidcUserInfo userInfo, Set<GrantedAuthority> authorities) {
+				OAuth2AccessToken accessToken = userRequest.getAccessToken();
+				Set<GrantedAuthority> mappedAuthorities = new HashSet<>();
 
-		return (userRequest) -> {
-			// Delegate to the default implementation for loading a user
-			OidcUser oidcUser = delegate.loadUser(userRequest);
+				// TODO
+				// 1) Fetch the authority information from the protected resource using accessToken
+				// 2) Map the authority information to one or more GrantedAuthority's and add it to mappedAuthorities
 
-			OAuth2AccessToken accessToken = userRequest.getAccessToken();
-			Set<GrantedAuthority> mappedAuthorities = new HashSet<>();
+				// Delegate to the default implementation for getting a user
+				OidcUser oidcUser = super.getUser(userRequest, userInfo, mappedAuthorities);
 
-			// TODO
-			// 1) Fetch the authority information from the protected resource using accessToken
-			// 2) Map the authority information to one or more GrantedAuthority's and add it to mappedAuthorities
-
-			// 3) Create a copy of oidcUser but use the mappedAuthorities instead
-			oidcUser = new DefaultOidcUser(mappedAuthorities, oidcUser.getIdToken(), oidcUser.getUserInfo());
-
-			return oidcUser;
+				return oidcUser;
+			}
 		};
 	}
 }
@@ -668,21 +666,17 @@ class OAuth2LoginSecurityConfig  {
     }
 
     @Bean
-    fun oidcUserService(): OAuth2UserService<OidcUserRequest, OidcUser> {
-        val delegate = OidcUserService()
-
-        return OAuth2UserService { userRequest ->
-            // Delegate to the default implementation for loading a user
-            var oidcUser = delegate.loadUser(userRequest)
-
+    fun oidcUserService(): OAuth2UserService<OidcUserRequest, OidcUser> = object : OidcUserService() {
+        override fun getUser(userRequest: OidcUserRequest, userInfo: OidcUserInfo, authorities: Set<GrantedAuthority>): OidcUser {
             val accessToken = userRequest.accessToken
             val mappedAuthorities = HashSet<GrantedAuthority>()
 
             // TODO
             // 1) Fetch the authority information from the protected resource using accessToken
             // 2) Map the authority information to one or more GrantedAuthority's and add it to mappedAuthorities
-            // 3) Create a copy of oidcUser but use the mappedAuthorities instead
-            oidcUser = DefaultOidcUser(mappedAuthorities, oidcUser.idToken, oidcUser.userInfo)
+
+            // Delegate to the default implementation for getting a user
+            val oidcUser = super.getUser(userRequest, userInfo, mappedAuthorities)
 
             oidcUser
         }
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcReactiveOAuth2UserService.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcReactiveOAuth2UserService.java
@@ -111,15 +111,19 @@ public class OidcReactiveOAuth2UserService implements ReactiveOAuth2UserService<
 					for (String scope : token.getScopes()) {
 						authorities.add(new SimpleGrantedAuthority("SCOPE_" + scope));
 					}
-					String userNameAttributeName = userRequest.getClientRegistration().getProviderDetails()
-							.getUserInfoEndpoint().getUserNameAttributeName();
-					if (StringUtils.hasText(userNameAttributeName)) {
-						return new DefaultOidcUser(authorities, userRequest.getIdToken(), userInfo,
-								userNameAttributeName);
-					}
-					return new DefaultOidcUser(authorities, userRequest.getIdToken(), userInfo);
+					return getUser(userRequest, userInfo, authorities);
 				});
 		// @formatter:on
+	}
+
+	protected OidcUser getUser(OidcUserRequest userRequest, OidcUserInfo userInfo, Set<GrantedAuthority> authorities) {
+		String userNameAttributeName = userRequest.getClientRegistration().getProviderDetails()
+				.getUserInfoEndpoint().getUserNameAttributeName();
+		if (StringUtils.hasText(userNameAttributeName)) {
+			return new DefaultOidcUser(authorities, userRequest.getIdToken(), userInfo,
+					userNameAttributeName);
+		}
+		return new DefaultOidcUser(authorities, userRequest.getIdToken(), userInfo);
 	}
 
 	private Mono<OidcUserInfo> getUserInfo(OidcUserRequest userRequest) {
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserService.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserService.java
@@ -145,7 +145,7 @@ public class OidcUserService implements OAuth2UserService<OidcUserRequest, OidcU
 		return DEFAULT_CLAIM_TYPE_CONVERTER.convert(oauth2User.getAttributes());
 	}
 
-	private OidcUser getUser(OidcUserRequest userRequest, OidcUserInfo userInfo, Set<GrantedAuthority> authorities) {
+	protected OidcUser getUser(OidcUserRequest userRequest, OidcUserInfo userInfo, Set<GrantedAuthority> authorities) {
 		ProviderDetails providerDetails = userRequest.getClientRegistration().getProviderDetails();
 		String userNameAttributeName = providerDetails.getUserInfoEndpoint().getUserNameAttributeName();
 		if (StringUtils.hasText(userNameAttributeName)) {


daniel-shuy:
Any update on this? Its been awhile

sjohnr:
Thanks for the ping @daniel-shuy! I have not had time to look at this recently, but I will add it to my TODO list for after the 6.1 release next week.

sjohnr:
@daniel-shuy, thanks for the PR!

However, I don't find myself in favor of introducing a `protected` method and inheritance in order to facilitate easier delegation-based examples. I'm also having a hard time understanding what scenario is being solved for here.

I understand that the docs demonstrate using delegation and use the example of mapping authorities as the use case for the customization. However, the previous example of [Using a GrantedAuthoritiesMapper](https://docs.spring.io/spring-security/reference/reactive/oauth2/login/advanced.html#webflux-oauth2-login-advanced-map-authorities-grantedauthoritiesmapper) would suffice in most cases. Do you truly have an advanced scenario that requires delegation? Or were you led that way by the docs only due to the example? Can you describe your use case for the customization here?

daniel-shuy:
@sjohnr

> I understand that the docs demonstrate using delegation and use the example of mapping authorities as the use case for the customization. However, the previous example of [Using a GrantedAuthoritiesMapper](https://docs.spring.io/spring-security/reference/reactive/oauth2/login/advanced.html#webflux-oauth2-login-advanced-map-authorities-grantedauthoritiesmapper) would suffice in most cases. Do you truly have an advanced scenario that requires delegation? Or were you led that way by the docs only due to the example? Can you describe your use case for the customization here?

Yes, to configure `spring-security-oauth2-client` for Keycloak.

Keycloak stores the roles in the access token payload as e.g.
```json5
{
  // ...
  "realm_access": {
    "roles": [
      // Roles
    ]
  },
  "resource_access": {
    "client": {
      "roles": [
        // Authorities
      ]
    },
    // ...
  },
  // ...
}
```

Therefore to map Keycloak roles to Spring Security authorities, a custom `OAuth2UserService`/`ReactiveOAuth2UserService` needs to be created (a `GrantedAuthoritiesMapper` does not have access to the `OidcUserRequest` to get the access token payload):
```java
@Bean
public ReactiveOAuth2UserService<OidcUserRequest, OidcUser> oidcUserService() {
	final OidcReactiveOAuth2UserService delegate = new OidcReactiveOAuth2UserService();

	return (userRequest) -> {
		// Delegate to the default implementation for loading a user
		return delegate.loadUser(userRequest)
				.flatMap((oidcUser) -> {
					OAuth2AccessToken accessToken = userRequest.getAccessToken();
					Set<GrantedAuthority> mappedAuthorities = new HashSet<>();

					// map Keycloak roles in token to Spring Security authorities

					ProviderDetails providerDetails = userRequest.getClientRegistration().getProviderDetails();
					String userNameAttributeName = providerDetails.getUserInfoEndpoint().getUserNameAttributeName();
					if (StringUtils.hasText(userNameAttributeName)) {
						oidcUser = new DefaultOidcUser(mappedAuthorities, oidcUser.getIdToken(), oidcUser.getUserInfo(), userNameAttributeName);
					} else {
						oidcUser = new DefaultOidcUser(mappedAuthorities, oidcUser.getIdToken(), oidcUser.getUserInfo());
					}

					return Mono.just(oidcUser);
				});
	};
}
```

The current API is very hard to understand (requires understanding the inner workings of [OidcUserService](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserService.java)/[OidcReactiveOAuth2UserService](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcReactiveOAuth2UserService.java)) and error prone (the bug in the documentation still isn't fixed, see gh-12281).

daniel-shuy:
> However, I don't find myself in favor of introducing a protected method and inheritance in order to facilitate easier delegation-based examples.

What about adding a setter to [OidcUserService](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserService.java)/[OidcReactiveOAuth2UserService](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcReactiveOAuth2UserService.java) to take in a `Converter<OidcUserRequest, Collection<GrantedAuthority>>`, like [JwtAuthenticationConverter](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationConverter.java)?

daniel-shuy:
@sjohnr 

> First, I would suggest using Keycloak's mappers to map claims to the `id_token`, as this is by far the easiest way to get access to roles. Having said that, I'm not an expert in Keycloak administration so perhaps that doesn't make sense or is not possible for you.

Yes it is possible, but I would prefer not having to change the Keycloak configuration as much as possible, just to get it working with Spring Security, as I am trying to write an open source library that can be reused by others, to replace [keycloak-spring-boot-starter](https://www.keycloak.org/docs/latest/securing_apps/#_spring_boot_adapter), which is now deprecated (https://www.keycloak.org/2022/02/adapter-deprecation.html).

---

> In any case, because the `Converter` interface only accepts a single parameter, and the mapping to authorities requires both `OidcUserRequest` and `OidcUserInfo`

Ah, I was thinking to simply map them to `SimpleGrantedAuthority`s, but good point, it should map them to `OidcUserAuthority`s/`OAuth2UserAuthority`s. What about something like:
```java
public interface OidcUserRequestAuthoritiesConverter implements Converter<OidcUserRequest, Collection<String>> {
}

public interface OAuth2UserRequestAuthoritiesConverter implements Converter<OAuth2UserRequest, Collection<String>> {
}
```

Then a setter (e.g. `setOidcUserRequestGrantedAuthoritiesConverter`/`setOAuth2UserRequestAuthoritiesConverter`) can be added to [OidcUserService](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserService.java) and [DefaultOAuth2UserService](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java) respectively to map the `OidcUserRequest`/`OAuth2UserRequest` to a `Collection<String>`, and `loadUser()` can be enhanced to map the `Collection<String>` to `OidcUserAuthority`s/`OAuth2UserAuthority`s.

---

> We could do something very similar in `DefaultOAuth2UserService`. What do you think?

Both `OidcUserService` and `DefaultOAuth2UserService` need to be modified. Even though `OidcUserService` delegates to `DefaultOAuth2UserService`, the authorities from the `OAuth2User` returned by `DefaultOAuth2UserService` are not used by the `OidcUserService`.

sjohnr:
> I am trying to write an open source library that can be reused by others, to replace [keycloak-spring-boot-starter](https://www.keycloak.org/docs/latest/securing_apps/#_spring_boot_adapter), which is now deprecated (https://www.keycloak.org/2022/02/adapter-deprecation.html).

I see. Thanks!

> Ah, I was thinking to simply map them to `SimpleGrantedAuthority`s, but good point, it should map them to `OidcUserAuthority`s/`OAuth2UserAuthority`s. What about something like:
> 
> ```java
> public interface OidcUserRequestAuthoritiesConverter implements Converter<OidcUserRequest, Collection<String>> {
> }
> 
> public interface OAuth2UserRequestAuthoritiesConverter implements Converter<OAuth2UserRequest, Collection<String>> {
> }
> ```

I'm not sure this will work. As I mentioned, the `OidcUserAuthority` requires the `OidcUserInfo` which is looked up earlier in the `loadUser` method. I also think that if we want to introduce a new hook into `OidcUserService`, it would ideally add quite a bit more flexibility.

We could introduce a `@FunctionalInterface`, but it needs to take 2 parameters (`OidcUserRequest` and `OidcUserInfo`) and return a `Collection<GrantedAuthority>` (or specifically a `Collection<? extends GrantedAuthority>` like `GrantedAuthoritiesMapper` does). In this way, the entire set of authorities can be replaced instead of just added to.

Also, `AuthorityUtils` is available to map from `String...` to a `List<GrantedAuthority>` so we don't need to do that mapping for users.

Lastly, I think any new hook we add here should extend from the hook we add for `DefaultOAuth2UserService` (with generics), since these two classes form a hierarchy.

daniel-shuy:
Oops, sorry, I meant to write:
```java
public interface OidcUserRequestAuthoritiesConverter {
    Collection<String> toAuthorities(OidcUserRequest userRequest, OidcUserInfo userInfo);
}

public interface OAuth2UserRequestAuthoritiesConverter {
    Collection<String> toAuthorities(OAuth2UserRequest userRequest, Map<String, Object> userAttributes);
}
```

---

> We could introduce a `@FunctionalInterface`, but it needs to take 2 parameters (`OidcUserRequest` and `OidcUserInfo`) and return a `Collection<GrantedAuthority>` (or specifically a `Collection<? extends GrantedAuthority>` like `GrantedAuthoritiesMapper` does). In this way, the entire set of authorities can be replaced instead of just added to.
> 
> Also, `AuthorityUtils` is available to map from `String...` to a `List<GrantedAuthority>` so we don't need to do that mapping for users.

I have a question - I see [`OidcUserService#loadUser(OidcUserRequest)` creating a single `OidcUserAuthority` while mapping the token scopes as `SimpleGrantedAuthority`s](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserService.java#L131) and [`DefaultOAuth2UserService#loadUser(OAuth2UserRequest)` creating a single `OAuth2UserAuthority`, while mapping the token scopes as `SimpleGrantedAuthority`s](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java#L111). Should the mapper map all roles to [`OidcUserAuthority`s](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/oidc/user/OidcUserAuthority.java#L67)/[`OAuth2UserAuthority`s](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/user/OAuth2UserAuthority.java#L58), or like `OidcUserService`/`DefaultOAuth2UserService`, create one `OidcUserAuthority`/`OAuth2UserAuthority` and map the rest as `SimpleGrantedAuthority`s?

> Lastly, I think any new hook we add here should extend from the hook we add for `DefaultOAuth2UserService` (with generics), since these two classes form a hierarchy.

I don't think that is possible, as `OAuth2UserAuthority` requires a `Map<String, Object>` (user attributes), but `OidcUserInfo` does not extend `Map<String, Object>`.

sjohnr:
> Oops, sorry, I meant to write:
> 
> ```java
> public interface OidcUserRequestAuthoritiesConverter {
>     Collection<String> toAuthorities(OidcUserRequest userRequest, OidcUserInfo userInfo);
> }
> 
> public interface OAuth2UserRequestAuthoritiesConverter {
>     Collection<String> toAuthorities(OAuth2UserRequest userRequest, Map<String, Object> userAttributes);
> }
> ```

That's looking better for sure! However, I don't think they should return `Collection<String>`. I think the user should decide what representation of `GrantedAuthority` to return. What do you think of this?

> I have a question - I see [`OidcUserService#loadUser(OidcUserRequest)` creating a single `OidcUserAuthority` while mapping the token scopes as `SimpleGrantedAuthority`s](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserService.java#L131) and [`DefaultOAuth2UserService#loadUser(OAuth2UserRequest)` creating a single `OAuth2UserAuthority`, while mapping the token scopes as `SimpleGrantedAuthority`s](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java#L111). Should the mapper map all roles to [OidcUserAuthority](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/oidc/user/OidcUserAuthority.java#L67)s/[OAuth2UserAuthority](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/user/OAuth2UserAuthority.java#L58)s, or like `OidcUserService`/`DefaultOAuth2UserService`, create one `OidcUserAuthority`/`OAuth2UserAuthority` and map the rest as `SimpleGrantedAuthority`s?

I think there would need to be an advantage to changing it, which I'm not sure about right now. If there is, we'd probably want to address that separately. 

Currently, we return a `Collection` of `GrantedAuthority` and later invoke `GrantedAuthoritiesMapper` where the user has to `instanceof OidcUserAuthority` or `instanceof OAuth2UserAuthority` to pull that authority out of the collection in order to use it, which may be tedious.

That's why I'm excited about introducing this new option in `OidcUserService` and `OAuth2UserService` as it feels like it could be easier to use!

> I don't think that is possible, as `OAuth2UserAuthority` requires a `Map<String, Object>` (user attributes), but `OidcUserInfo` does not extend `Map<String, Object>`.

That's a fair point. I was just thinking, if possible the hooks should be closely related. But you're right, probably not a hierarchy.

daniel-shuy:
> That's looking better for sure! However, I don't think they should return `Collection<String>`. I think the user should decide what representation of `GrantedAuthority` to return. What do you think of this?

I'm kind of on the fence about this. If the mapper should follow [OidcUserService#loadUser(OidcUserRequest)](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserService.java#L131)/[DefaultOAuth2UserService(OAuth2UserRequest)](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java#L111), to create a single `OidcUserAuthority`/`OAuth2UserAuthority` and map the other roles as `SimpleGrantedAuthority`s, then the issue with returning `Collection<GrantedAuthority>`, is that it requires the user to know the inner workings of `OidcUserService`/`OAuth2UserService`. But on the other hand it offers more flexibility compared to returning a `Collection<String>`.

Another option is to simply map everything as `SimpleGrantedAuthority`s. In that case the `OidcUserInfo` or `OAuth2User` attributes (`Map<String, Object>`) won't even be required.

What is the purpose/advantage of using `OidcUserAuthority`/`OAuth2UserAuthority` over `SimpleGrantedAuthority`?

sjohnr:
> What is the purpose/advantage of using `OidcUserAuthority`/`OAuth2UserAuthority` over `SimpleGrantedAuthority`?

I'm not 100% sure the background of the existing feature, but I believe it's purpose is to allow the `instanceof` checks I mentioned. It allows compatibility with the existing `GrantedAuthoritiesMapper` interface.

> I'm kind of on the fence about this. If the mapper should follow [OidcUserService#loadUser(OidcUserRequest)](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserService.java#L131)/[DefaultOAuth2UserService(OAuth2UserRequest)](https://github.com/daniel-shuy/spring-security/blob/77f62d8b6cc2ae6229f1345cabe303b215043ef0/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java#L111), to create a single `OidcUserAuthority`/`OAuth2UserAuthority` and map the other roles as `SimpleGrantedAuthority`s, then the issue with returning `Collection<GrantedAuthority>`, is that it requires the user to know the inner workings of `OidcUserService`/`OAuth2UserService`.

We will need to enhance the docs along with this change. But if we add this capability, I think the inner workings of the class become less important for users, because usually you would choose one of:

a) customize using this new option, so you don't usually require additional mapping and don't need the special authority
b) customize using the existing `GrantedAuthoritiesMapper` option, which uses the existing documented feature
c) customize both, in which case you are aware of the details of both custom implementations and don't need to know about anything else

Does that make sense?

> Another option is to simply map everything as `SimpleGrantedAuthority`s. In that case the `OidcUserInfo` or `OAuth2User` attributes (`Map<String, Object>`) won't even be required.

I want to be careful not to expand the scope of this enhancement. While we could **add** mappings to a collection containing that existing authority, as stated above I think there's better flexibility and a simpler contract by allowing the user to fully override the mapped collection. We just need to document that is what the user is doing by customizing.

daniel-shuy:
Sounds good! I'm in favor of giving users the flexibility while enhancing the docs

sjohnr:
@daniel-shuy just checking in. Are you able to make updates to this PR based on my [comment](https://github.com/spring-projects/spring-security/pull/12282#issuecomment-1555083781) above?

sjohnr:
@daniel-shuy I'm going to close this PR in favor of the approach I outlined above. I've created a [branch](https://github.com/sjohnr/spring-security/tree/gh-14672-oidc-user-mapper) with changes for the servlet side to demonstrate. I will proceed with the reactive side shortly so we can try for 6.3.0-M3. Please let me know if you have any thoughts.

Issue #12032, opened 2022-10-17
Replace deprecated set-state set-output GitHub Action's commands


marcusdacoregio:
See the warnings on this run https://github.com/spring-projects/spring-security/actions/runs/3268576453

PR #12289, opened 2022-11-24
Replace deprecated set-state set-output GitHub Action's commands

kirklin:
The save-state, set-output command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Closes gh-12032

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -31,15 +31,15 @@ jobs:
       runjobs: ${{ steps.continue.outputs.runjobs }}
       project_version: ${{ steps.continue.outputs.project_version }}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - id: continue
         name: Determine if should continue
         run: |
           # Run jobs if in upstream repository
-          echo "::set-output name=runjobs::true"
+          echo "runjobs=true" >>$GITHUB_OUTPUT
           # Extract version from gradle.properties
           version=$(cat gradle.properties | grep "version=" | awk -F'=' '{print $2}')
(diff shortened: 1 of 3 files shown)


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -71,9 +71,9 @@ jobs:
     runs-on: ubuntu-latest
     if: needs.prerequisites.outputs.runjobs
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Set up gradle
-        uses: spring-io/spring-gradle-build-action@v1
+        uses: kirklin/spring-gradle-build-action@v1
```
marcusdacoregio, line 76: I don't think we want to use an action that is outside the Spring control, instead of doing this, we should open an issue in https://github.com/spring-io/spring-gradle-build-action in order to get a new version released with the deprecations replaced.

Please, let's keep `spring-io/spring-gradle-build-action@v1` for now.

marcusdacoregio (changes requested):
Thanks, @kirklin, I've left my feedback inline.

Also, can you please squash your commits into one?

.github/workflows/continuous-integration-workflow.yml
kirklin: ok!I will sync the update to `spring-io/spring-gradle-build-action@v1`.

marcusdacoregio:
Thanks @kirklin, this is now merged via https://github.com/spring-projects/spring-security/commit/9c6d90490d92cd301999da1a50c1a8ee6f1ac1fa

Issue #9642, opened 2021-04-14
Reject use of Stream


rwinch:
Similar to gh-9641 we should reject usage of Stream

See gh-7154

PR #12312, opened 2022-11-28
Use array stream to convert pattern to RequestMatcher

damingerdai:


marcusdacoregio:
Hi @damingerdai, thanks for the PR.

Unfortunately we reject use of Stream and Optional in the code, see reasons in https://github.com/spring-projects/spring-security/issues/9642

PR #12318, opened 2022-11-29
Fix password encoder migration guide

mschneid:


jzheaux:
Thanks, @mschneid! This is now merged into `5.8.x`.

Issue #12232, opened 2022-11-17
SecuredAuthorizationManager should cache annotation's value


jzheaux:
`SecurityAuthorizationManager` currently caches an `AuthorizationManager` per method/class pair.

The memory footprint this requires can be improved by instead caching the annotation's value and passing that into an instance of the class indicated in #12231.

evgeniycheban:
@jzheaux I can take this.

PR #12335, opened 2022-12-03
SecuredAuthorizationManager should cache annotation's value

evgeniycheban:
Closes gh-12232

### DIFF
--- a/core/src/test/java/org/springframework/security/authorization/AuthoritiesAuthorizationManagerTests.java
+++ b/core/src/test/java/org/springframework/security/authorization/AuthoritiesAuthorizationManagerTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/springframework/security/authorization/AuthoritiesAuthorizationManagerTests.java
```diff
@@ -74,14 +74,14 @@ void checkWhenUserHasNotAnyAuthorityThenDeniedDecision() {
 	}
 
 	@Test
-	void hasRoleWhenRoleHierarchySetThenGreaterRoleTakesPrecedence() {
+	void checkWhenRoleHierarchySetThenGreaterRoleTakesPrecedence() {
```
jzheaux, line 77: Will you please put this in a separate commit with a comment that describes why the change was made? By doing that, each commit will stay focused on one atomic change.

jzheaux (changes requested):
Thanks, @evgeniycheban! This looks great. I left a piece of feedback inline.

Also, now that we are in 2023, will you please update the copyright year for these files?

core/src/test/java/org/springframework/security/authorization/AuthoritiesAuthorizationManagerTests.java
evgeniycheban: Done.

jzheaux:
Thanks, @evgeniycheban! This is now merged into `main`.

PR #12352, opened 2022-12-07
Introduce JwtDecoders.fromIssuerAndJwksUri

nkonev:
For my configured server, it doesn't have OAuth2 handle like `https:/my-company/.well-known/oauth-authorization-server/ISSUER`, which is need in `JwtDecoderProviderConfigurationUtils.getConfigurationForIssuerLocation(issuer) -> oauth(uri)`.
The second problem with it is it generates JWT where issuer (`iss`) is just string, not url, according https://www.rfc-editor.org/rfc/rfc7519#section-4.1.1.

So it's helpful for me to just create JwtDecoder with issuer and `jwks_uri` provided by myself. 

For more context:
I am adapting this manual https://curity.io/resources/learn/spring-boot-api/
Why don't I just copy the entire `JwtDecoders.withProviderConfiguration` ? because in `JwtDecoderProviderConfigurationUtils::addJWSAlgorithms` both class and method are not public.

PTAL @jzheaux 

---
Actually, it's better variant of https://github.com/spring-projects/spring-security/pull/12348.

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoders.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoders.java
@@ -95,6 +95,26 @@ public final class JwtDecoders {
 	}
 
 	/**
+	 * Creates a {@link JwtDecoder} using the provided
+	 * <a href= "https://tools.ietf.org/html/rfc7519#section-4.1.1">Issuer</a> an jwksUri.
+	 *
+	 * Note that this method is useful for cases when JWT token has {@code iss} param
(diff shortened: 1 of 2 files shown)


jzheaux:
The purpose of `JwtDecoders` is to query a configuration endpoint. Given that you don't have one, I think that the better approach is to use `NimbusJwtDecoder` directly like so:

```java
@Bean 
JwtDecoder jwtDecoder() {
    NimbusJwtDecoder jwtDecoder = NimbusJwtDecoder.withJwkSetUri(jwkSetUri).build();
    jwtDecoder.setJwtValidator(JwtValidators.createDefaultFromIssuer(issuer));
    return jwtDecoder;
}
```

PR #12366, opened 2022-12-11
Adds hints for Authentication manager

nkonev:
Fixes https://github.com/spring-projects/spring-framework/issues/29599 (As a community guy I cannot move issue from the Framework to here)

@marcusdacoregio I saw you added some hints https://github.com/spring-projects/spring-security/commit/ef879aadd65bbd2edc76be340b554741da87358b, so I think I can ask you to review.

### DIFF
--- a/core/src/main/java/org/springframework/security/aot/hint/CoreSecurityRuntimeHints.java
+++ b/core/src/main/java/org/springframework/security/aot/hint/CoreSecurityRuntimeHints.java
@@ -19,14 +19,18 @@ package org.springframework.security.aot.hint;
 import java.util.List;
 import java.util.stream.Stream;
 
+import org.springframework.aop.SpringProxy;
+import org.springframework.aop.framework.Advised;
 import org.springframework.aot.hint.MemberCategory;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/springframework/security/aot/hint/CoreSecurityRuntimeHints.java
```diff
@@ -19,14 +19,18 @@
 import java.util.List;
 import java.util.stream.Stream;
 
+import org.springframework.aop.SpringProxy;
```
nkonev, line 22: @sdeleuze Should we move SpringProxy, Advised and DecoratingProxy to the Spring Framework ? They can be useful for other

marcusdacoregio:
Hi @nkonev, thanks for the contribution.

You can go ahead and close that ticket in Spring Framework.

About the PR, I don't think we want to solve it this way. Instead, a delegate that delays the initialization of the bean would probably work in this case. I created https://github.com/spring-projects/spring-security/issues/12367 for that

nkonev:
Ok, I suspected that it isn't a simple fix :)
Closing

PR #12376, opened 2022-12-14
Make sure the replacing path include the path before the context path.

### LINKED ISSUE #12375: {baseUrl} does not return the proper path if the server is behind the reverse proxy
If a server is behind a reverse proxy, for example, the application correct URL is "https://www.exmaple.com/**t1**/app-context/", but the {baseUrl} returns  as "https://www.exmaple.com/app-context/", the "**/t1**" that represents a server behind the reverse proxy is missing

zhanedw-statcan:
Fixes #12375


marcusdacoregio:
I'll close this as it seems that there was no progress on the discussion in https://github.com/spring-projects/spring-security/issues/12375

PR #12433, opened 2022-12-19
switch to published partial build extension and remove local extension (docs-build)

mojavelinux:
Remove the local copy of the partial build extension and switched to the one published in the `@springio/antora-extensions` package. The extension will now use isomorphic-git instead of the local git command. This PR also upgrades the `@spring/asciidoctor-extensions` package, which no longer contains the Antora extensions.


marcusdacoregio:
Thanks @mojavelinux, this is now merged.

PR #12437, opened 2022-12-19
Fix typo in OAuth 2.0 testing docs

eleftherias:
The `RequestPostProcessor` that takes no arguments is `oauth2Login`, which is what this paragraph describes.

This should be forward-ported up to 6.0.x.


marcusdacoregio:
Thank you @eleftherias.

This is now merged into `5.7.x` and forward-ported to `5.8.x`, `6.0.x` and `main`.

Issue #12430, opened 2022-12-18
Allow authorization request resolver to be changed for the OAuth2 client configuration


spoptchev:
**Describe the bug**

The `authorizationRequestResolver` can't be changed in the OAuth2 client configuration. Because of this, we can't use the Twitter oAuth2 flow, which needs PKCE.

**To Reproduce**

To reproduce the issue you can use our [example code](https://github.com/zone-2-technologies/twitter-oauth2-client-pkce). Just run the code in the `spring-security-pkce-issue` project. Please follow the `README.md` in the repository root directory for instructions.

Additionally, we have prepared a [fix](https://github.com/spoptchev/spring-security/tree/authorization-resolver-oauth2-client).

**Expected behavior**

The OAuth2 client configuration should be able to be changed so that the `OAuth2AuthorizationRequestCustomizers.withPkce()` customizer can be added to the authorization request resolver.

**Please be aware that we explicitly do not need an OAuth2 login for our application**. We just want to use the OAuth2 client configuration.

jgrandja:
@spoptchev You are correct, `OAuth2ClientSpec.authorizationRequestResolver(ServerOAuth2AuthorizationRequestResolver)` is not available.

Would you be interested in submitting a PR for this enhancement? It should be similar to `OAuth2LoginSpec.authorizationRequestResolver(ServerOAuth2AuthorizationRequestResolver)`.

spoptchev:
Hi @jgrandja, sure, I opened a PR (https://github.com/spring-projects/spring-security/pull/12438).

PR #12438, opened 2022-12-19
Allow authorization request resolver to be changed for the OAuth2 client configuration

spoptchev:
Fixes gh-12430. 

See issue for details.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
@@ -3841,6 +3841,7 @@ public class ServerHttpSecurity {
 		 * @param authorizationRequestResolver the resolver used for resolving
 		 * {@link OAuth2AuthorizationRequest}'s
 		 * @return the {@link OAuth2ClientSpec} for further configuration
+		 * @since 6.1
 		 */
 		public OAuth2ClientSpec authorizationRequestResolver(
 				ServerOAuth2AuthorizationRequestResolver authorizationRequestResolver) {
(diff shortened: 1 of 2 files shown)


config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -3831,9 +3831,30 @@ public final class OAuth2ClientSpec {
...
+
 		private OAuth2ClientSpec() {
 		}
 
+		/**
+		 * Sets the resolver used for resolving {@link OAuth2AuthorizationRequest}'s.
+		 * @param authorizationRequestResolver the resolver used for resolving
+		 * {@link OAuth2AuthorizationRequest}'s
+		 * @return the {@link OAuth2ClientSpec} for further configuration
+		 */
```
jgrandja, line 3844: Add `@since 6.1`

jgrandja (changes requested):
Thanks for the PR @spoptchev !

Can you also please add a test similar to `OAuth2LoginTests.oauth2LoginWhenCustomObjectsThenUsed()`. We should test if a custom `authorizationRequestResolver` is set then it is used. I think all you need to modify is `OAuth2ClientSpecTests.oauth2ClientWhenCustomObjectsThenUsed()`.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
spoptchev: Done 👍

spoptchev:
@jgrandja I added the test. Let me know if something is missing.

jgrandja:
Thanks for the updates @spoptchev. This is now merged.

FYI, I added a polish commit with some minor updates in order to get this merged as I am off until the new year.

Enjoy the holidays!

spoptchev:
@jgrandja thank you for the integration.

Happy holidays 🎄

PR #12440, opened 2022-12-19
Take refresh token expiry into account.

### LINKED ISSUE #12439: PasswordOAuth2AuthorizedClientProvider doesn't take refresh token expiry into account
**Describe the bug**
PasswordOAuth2AuthorizedClientProvider doesn't take refresh token expiry into account. PasswordOAuth2AuthorizedClientProvider only checks that there is a refresh token and then returns null assuming that the RefreshTokenOAuth2AuthorizedClientProvider will get a new token. But if the refresh token has expired, the authorize flow will fail as the RefreshTokenOAuth2AuthorizedClientProvider can't refresh and the PasswordOAuth2AuthorizedClientProvider should have acquired a new authorization token.

**To Reproduce**
create a client provider using OAuth2AuthorizedClientProviderBuilder and add password() and refreshToken(). Use the provider to authorize and thus obtain an authorization token. Wait until the refresh token expires. Try to call authorize again and exception is thrown even though the provider has enough information to acquire a valid token.

**Expected behavior**
Client provider falls back to PasswordOAuth2AuthorizedClientProvider and acquires a new token.

PS. the expiry checks on both PasswordOAuth2AuthorizedClientProvider and RefreshTokenOAuth2AuthorizedClientProvider get the expiry Instant with token.getExpiresAt() which is labeled with Nullable, but they call the .minus method without null check.

jhahkala:
Take refresh token expiry into account.
Fix possible NPE on the expiry check

fixes #12439

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/ClientCredentialsOAuth2AuthorizedClientProvider.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/ClientCredentialsOAuth2AuthorizedClientProvider.java
@@ -98,7 +98,11 @@ public final class ClientCredentialsOAuth2AuthorizedClientProvider implements OA
 	}
 
 	private boolean hasTokenExpired(OAuth2Token token) {
-		return this.clock.instant().isAfter(token.getExpiresAt().minus(this.clockSkew));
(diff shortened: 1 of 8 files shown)


jgrandja:
Closing as per [comment](https://github.com/spring-projects/spring-security/issues/12439#issuecomment-1359713291)

Issue #12454, opened 2022-12-23
Cookie no support for field 'version' and 'comment'


twosom:
**Expected Behavior**

`jakarta.servlet.http.Cookie` now support <a href="http://www.ietf.org/rfc/rfc6265.txt">RFC 6265</a>.

and `getComment`, `setComment`, `getVersion`, `setVersion` 
With the adoption of support for RFC 6265, those method should no longer be used.

and now, those methods's body is empty or return null.
therefore, we should add `@Deprecated` on `SavedCookie.class`'s constructor and some getter.

**Current Behavior**

in getter, it always return 0 or null. and setter have no operation.

**Context**
Suggested solution: #12531

PR #12455, opened 2022-12-23
refactoring test related with Cookie.

twosom:
Fixed the code related to Cookie's deprecated methods.
#12454 
Closes gh-12454

### DIFF
--- a/web/src/main/java/org/springframework/security/web/savedrequest/SavedCookie.java
+++ b/web/src/main/java/org/springframework/security/web/savedrequest/SavedCookie.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -31,22 +31,28 @@ public class SavedCookie implements Serializable {
 
 	private static final long serialVersionUID = SpringSecurityCoreVersion.SERIAL_VERSION_UID;
 
-	private final java.lang.String name;
+	private final String name;
 
-	private final java.lang.String value;
+	private final String value;
 
-	private final java.lang.String comment;
+	private final String comment;
 
-	private final java.lang.String domain;
+	private final String domain;
 
 	private final int maxAge;
 
-	private final java.lang.String path;
+	private final String path;
 
 	private final boolean secure;
 
 	private final int version;
 
+	/**
+	 * @deprecated use
+	 * {@link SavedCookie#SavedCookie(String, String, String, int, String, boolean)}
+	 * instead
+	 */
+	@Deprecated
 	public SavedCookie(String name, String value, String comment, String domain, int maxAge, String path,
 			boolean secure, int version) {
 		this.name = name;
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/springframework/security/web/jackson2/SavedCookieMixin.java
```diff
@@ -29,24 +29,22 @@
...
+ * 		mapper.registerModule(new WebServletJackson2Module());
  * </pre>
  *
  * @author Jitendra Singh.
- * @since 4.2
  * @see WebServletJackson2Module
  * @see org.springframework.security.jackson2.SecurityJackson2Modules
  */
-@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY)
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS)
```
jzheaux, line 39: I think we should leave this class as-is. The reason is that we want to remain backward compatible to circumstances where `comment` and `version` are still in use. This will allow applications to change when they are ready.

jzheaux (changes requested):
Thanks, @twosom! Instead of removing usage, we should follow the same pattern of deprecating the Spring Security APIs that use the deprecated Cookie APIs. I've left additional detail inline.

web/src/main/java/org/springframework/security/web/jackson2/SavedCookieMixin.java
```diff
@@ -29,24 +29,22 @@
  *
  * <pre>
  * 		ObjectMapper mapper = new ObjectMapper();
- *		mapper.registerModule(new WebServletJackson2Module());
+ * 		mapper.registerModule(new WebServletJackson2Module());
  * </pre>
  *
  * @author Jitendra Singh.
- * @since 4.2
```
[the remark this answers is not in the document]
twosom: Oops... that was a mistake! sorry.

web/src/main/java/org/springframework/security/web/savedrequest/SavedCookie.java
```diff
@@ -31,22 +31,28 @@ public class SavedCookie implements Serializable {
```
jzheaux, line 31: As a matter of hygiene, please update the copyright to `2023`.

jzheaux (changes requested):
@twosom, thanks! The updates are great. I've left some feedback inline.

In addition to the feedback, will you please squash your commits? In the resulting commit, please include `Closes gh-12454` as the last line to integrate with our automation.

web/src/main/java/org/springframework/security/web/savedrequest/SavedCookie.java
```diff
@@ -31,22 +31,28 @@ public class SavedCookie implements Serializable {
...
 	private final boolean secure;
 
 	private final int version;
 
+	/**
+	 * @deprecated use
+	 * {@link SavedCookie#SavedCookie(String, String, String, int, String, boolean)}
+	 * instead
+	 */
+	@Deprecated
```
marcusdacoregio, line 55: I think we can use `@Deprecated(forRemoval = true, since = "6.1")` to add more detail to it.

twosom: ok! I'll add more detail in `@Deprecated` annotation.

twosom:
@marcusdacoregio @jzheaux 
I'll Close this PR and reopen it.

PR #12471, opened 2022-12-26
Duplicate words.

sakaeda11:
The words "The default implementation" are duplicated


jzheaux:
Thanks, @sakaeda11! Will you please rebase this on `6.0.x` as that's the earliest supported branch where this bug occurs?

jzheaux:
Thanks, @sakaeda11! This is now merged into `6.0.x` and `main`.

PR #12474, opened 2022-12-28
Update multitenancy.adoc

deleze:
The Java example at line 421 should use the injected `jwtValidator` and not from the current class referenced by `this. jwtValidator`.


jzheaux:
Thanks, @deleze! This is now merged into `6.0.x` and `main`.

Issue #12233, opened 2022-11-17
SecuredAuthorizationManager should allow customizing underlying authorization manager


jzheaux:
Dependent on https://github.com/spring-projects/spring-security/issues/12232, exposing this authorization manager will allow applications to customize it. This can be handy in circumstances where, for example, authorization is externalized to service or database.

evgeniycheban:
I can take this.

remal:
@jzheaux, @evgeniycheban, how can this functionality be used?

My goal is to add `RoleHierarchy` to `@Secured` functionality. Unfortunately, I couldn't find a way to customize `SecuredAuthorizationManager` by any customizer or `BeanPostProcessor`.

PR #12488, opened 2023-01-06
SecuredAuthorizationManager should allow customizing underlying authorization manager

evgeniycheban:
SecuredAuthorizationManager should allow customizing underlying authorization manager

Closes gh-12233

### DIFF
--- a/core/src/main/java/org/springframework/security/authorization/method/SecuredAuthorizationManager.java
+++ b/core/src/main/java/org/springframework/security/authorization/method/SecuredAuthorizationManager.java
@@ -17,6 +17,7 @@
 package org.springframework.security.authorization.method;
 
 import java.lang.reflect.Method;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/springframework/security/authorization/method/SecuredAuthorizationManager.java
```diff
@@ -43,10 +45,22 @@
...
 	private final Map<MethodClassKey, Set<String>> cachedAuthorities = new ConcurrentHashMap<>();
 
+	/**
+	 * Sets an {@link AuthorizationManager} that accepts a collection of authority
+	 * strings.
+	 * @param delegate the {@link AuthorizationManager} that accepts a collection of
+	 * authority strings to use
+	 * @since 6.1
+	 */
+	public void setDelegate(AuthorizationManager<Collection<String>> delegate) {
```
jzheaux, line 59: I'd prefer that this method name be more descriptive. What about `setAuthoritiesAuthorizationManager`?

jzheaux (changes requested):

core/src/main/java/org/springframework/security/authorization/method/SecuredAuthorizationManager.java
evgeniycheban: Done.

jzheaux (approved):

PR #12497, opened 2023-01-06
add playbook template to support local build in content branches

mojavelinux:


mojavelinux:
This is a prerequisite step for rolling out the centralized playbook file for content branches.

PR #12500, opened 2023-01-06
upgrade Antora plugin and configure playbook provider to support local build

mojavelinux:


mojavelinux:
This PR can be merged as soon as https://github.com/spring-projects/spring-security/pull/12522 is merged (which provides the template).

mojavelinux:
With this change, it will no longer be necessary for the author to maintain or modify the playbook per content/software branch. Instead, the playbook is managed centrally in the docs-build branch.

The plugin does sideload or download the template to cached-antora-playbook.yml, which will be used if present. This still allows the author to play around with the playbook if desired. To get a fresh copy, the author can delete the file and the plugin will retrieve a fresh copy.

mojavelinux:
@rwinch I have updated this PR so it only contains the configuration to switch to the managed local build controlled by the playbook provider.

Note that the Gradle Antora Plugin now requires Java 17, so it's necessary to switch to Java 17 to run Antora within the branch. (It does not impact the production build since it runs everything using Java 17). If this is a problem, please file an issue in the Gradle Antora Plugin and we can discuss it.

mojavelinux:
Based on what CI is telling me, this is going to be a problem since it causes the main build to fail.

FYI, it's no longer necessary to having the antora task run as part of the main build since the antora task in a branch is only for author preview. But I can understand that you want to verify that it works for when the author needs it, so I supposed it's not unreasonable to run it in CI. But I wouldn't be bothering the devs with it as part of their main build...or as part of a distribution build.

mojavelinux:
I've upgrade to a version of the plugin that is compatible with Java 11. This PR is now ready.

PR #12502, opened 2023-01-06
turn off scheduler for Deploy Docs workflow

mojavelinux:
This need is already filled by the push trigger. When a commit is pushed to a branch, the workflow automatically kicks off a partial build for that branch/version. A full build of the site occurs when a release is made or on demand. There's no reason to run this workflow otherwise.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -5,8 +5,8 @@ on:
     tags: '**'
   repository_dispatch:
     types: request-build-reference # legacy
-  schedule:
-  - cron: '0 10 * * *' # Once per day at 10am UTC
+  #schedule:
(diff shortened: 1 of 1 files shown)


rwinch:
The only reason I can think of is for pushes to the UI, but that is a pretty small case and we tend to build right away vs waiting for a schedule. In any case, I think we probably need to switch to a versioned UI.

rwinch (approved):

mojavelinux:
> The only reason I can think of is for pushes to the UI

And you could still have that. But this particular scheduler wouldn't help with that anyway. It was only scheduling single version builds, which would not have updated the UI. It's reasonable to do a full build of the site weekly (which would require additional updates to this workflow file).

PR #12449, opened 2022-12-23
chore: Use cache in continuous-integration-workflow.yml

jongwooo:
Signed-off-by: jongwooo <jongwooo.han@gmail.com>

## Description

Updated `continuous-integration-workflow.yml` to cache dependencies using [actions/setup-java](https://github.com/actions/setup-java#caching-packages-dependencies). `setup-java@v3` or newer has caching **built-in**.

### About caching workflow dependencies
Jobs on GitHub-hosted runners start in a clean virtual environment and **must download dependencies each time**, causing increased network utilization, longer runtime, and increased cost. To help speed up the time it takes to recreate files like dependencies, GitHub can cache files that frequently use in workflows.

### Solutions
Java projects can run faster on GitHub Actions by enabling dependency caching on the [setup-java](https://github.com/actions/setup-java) action. `setup-java` supports caching for both Gradle and Maven projects.

**AS-IS**

```yaml
- name: Set up JDK 17
  uses: actions/setup-java@v3
  with:
    java-version: '17'
    distribution: 'temurin'
```

**TO-BE**

```yaml
- name: Set up JDK 17
  uses: actions/setup-java@v3
  with:
    java-version: '17'
    distribution: 'temurin'
    cache: 'gradle'
```

> It’s literally a one line change to pass the `cache: 'gradle'` input parameter.

## References
[actions/setup-java#caching-packages-dependencies](https://github.com/actions/setup-java#caching-packages-dependencies)
[How to build Gradle projects with GitHub Actions](https://tomgregory.com/build-gradle-projects-with-github-actions/#5_Enable_Gradle_caching)

PR #12503, opened 2023-01-07
chore: Use cache in continuous-integration-workflow.yml

jongwooo:
Signed-off-by: jongwooo <jongwooo.han@gmail.com>

## Description

Reopen #12449

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -55,6 +55,7 @@ jobs:
         with:
           java-version: '11'
           distribution: 'adopt'
+          cache: 'gradle'
       - name: Set up Gradle
         uses: gradle/gradle-build-action@v2
       - name: Set up gradle user name


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -53,8 +53,9 @@ jobs:
       - name: Set up JDK 11
         uses: actions/setup-java@v3
         with:
-          java-version: '11'
-          distribution: 'adopt'
+          java-version: '17'
```
jzheaux, line 56: I think you are only wanting to add the `cache` line, correct? Please remove these other two changes.

jzheaux (changes requested):
Thanks for updating this PR to be rebased on `5.7.x`, @jongwooo. I've left some feedback inline.

.github/workflows/continuous-integration-workflow.yml
jongwooo: @jzheaux Thank you for your review! I pushed a commit.

jzheaux (approved):

Issue #10143, opened 2021-07-27
OAuth2 OidcUserInfo is null because scopes coming from Google do not match


hedgehog1833:
I am using Spring Boot 2.5.3 with Spring Security 5.5.1 for my web app and have setup authentication via OAuth2 with Google next to native users that authenticated with email and password. I extended the `OidcUserService` to save OAuth-users with their name and email in a database. My `loadUser(OidcUserRequest userRequest)`-method looks like this:
```
public class CustomOidcUserService extends OidcUserService {

  ............

  @Override
  public OidcUser loadUser(OidcUserRequest userRequest) throws OAuth2AuthenticationException {
    OidcUser user = super.loadUser(userRequest);
    OidcUserInfo userInfo = user.getUserInfo();

    if (!userAlreadyExists(userInfo.getGivenName(), userInfo.getEmail())) {
      persistOAuthUser(userInfo);
    }

    return user;
  }

  ..........

}
```
Now the line with if-statement throws a NullPointer because the UserInfo is null. I did some digging and found that the UserInfo is not retreived because in the `OidcUserService` in Line 177 the expression `CollectionUtils.containsAny(userRequest.getAccessToken().getScopes(), this.accessibleScopes)` is false.
Google returns exactly these scopes in the `OAuth2AccessTokenResponse` (call made in `DefaultAuthorizationCodeTokenResponseClient`):
- "https://www.googleapis.com/auth/userinfo.profile"
- "https://www.googleapis.com/auth/userinfo.email"
- "openid"

While the configured accessibleScopes contain only "profile" and "email", without the host and so on so there is no match and the UserInfo is not retreived.

So now of course I can work around this and add these scopes to my UserService and then my UserInfo is set as expected. But that seems weird to me. Did Google change something here? Or am I still missing something?

My application.yml config also has nothing special regarding oauth:
```
security:
    oauth2:
      client:
        registration:
          google:
            client-id: clientId
            client-secret: clientSecret
```

I hope posting here was correct.... ☺️

marcusdacoregio:
Hi @hedgehog1833, thanks for the report.

Configuring the accessible scopes is the best option as per #6886.

I'm closing this but feel free to write here if you think that should be different, so we may have a discussion with @jgrandja.

hedgehog1833:
Hi @marcusdacoregio, thanks for your reply!
[...]

PR #12513, opened 2023-01-10
Get userInfo, if scopes in token is empty

sharat87:
In a WebFlux app, with OIDC authentication, the userInfo request is handled here: https://github.com/spring-projects/spring-security/blob/556891b4fa8e6b2e32be5b8dcadae485283d761a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcReactiveOAuth2UserService.java#L125-L128

Here, we call the `OidcUserRequestUtils.shouldRetrieveUserInfo` static function, to determine if we should do `userInfo`. In this function, we check if the scopes in the token, and the configured scopes match: https://github.com/spring-projects/spring-security/blob/556891b4fa8e6b2e32be5b8dcadae485283d761a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserRequestUtils.java#L60-L65

Now consider this check for a non-Webflux (non-reactive) OIDC authentication: https://github.com/spring-projects/spring-security/blob/556891b4fa8e6b2e32be5b8dcadae485283d761a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserService.java#L172-L184

We can see two major differences:

1. The non-Webflux flow, when it encounters missing scope field, or empty scope field, it _will_ call the `userInfo` endpoint. Whereas the Webflux flow _will not_.

2. The non-Webflux flow, also checks `this.accessibleScopes.isEmpty()`. Overriding this `accessibleScopes` has been [recommended](https://github.com/spring-projects/spring-security/issues/10143#issuecomment-888605197), to handle cases like missing scope in token. But this affordance is missing in the Webflux flow.

So, when using Webflux, I don't see a way out. Can we add these (at least the first) one, to the Webflux's flow as well? That's what this PR does. If the scopes are missing/empty in the token, _do_ call the `userInfo` endpoint.

Thank you for your time.

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserRequestUtils.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserRequestUtils.java
@@ -60,7 +60,8 @@ final class OidcUserRequestUtils {
 		if (AuthorizationGrantType.AUTHORIZATION_CODE.equals(clientRegistration.getAuthorizationGrantType())) {
 			// Return true if there is at least one match between the authorized scope(s)
 			// and UserInfo scope(s)
-			return CollectionUtils.containsAny(userRequest.getAccessToken().getScopes(),
+			return CollectionUtils.isEmpty(userRequest.getAccessToken().getScopes())
+				|| CollectionUtils.containsAny(userRequest.getAccessToken().getScopes(),
 					userRequest.getClientRegistration().getScopes());
 		}
 		return false;


sjohnr:
@sharat87, apologies as I seem to have missed this PR. Thanks for contributing!

Please note that it's usually best to open an issue first, and discuss possible solutions prior to submitting a PR. I noticed you submitted the PR using the GitHub editor. Are you able to download the project and run tests to verify that everything is working? There should be _at least_ one test that needs to change, and if not, would you please add a new test that covers this change?

sjohnr:
@sharat87 have you seen my [above comment](https://github.com/spring-projects/spring-security/pull/12513#issuecomment-1450618826)?

sjohnr:
Closing in favor of gh-13259.

PR #12520, opened 2023-01-10
Index Latest Patch Only

rwinch:
Rather than index every version, we should index only the latest patch release of each major.minor to avoid overwhelming users with redundant results.


mojavelinux:
This change is superseded by #12523. It's not sufficient to add a filter in the handlebars template because it will not produce valid JSON. But since the model for the template is controlled by the extension, we can do the filtering before calling the template.

rwinch:
Thanks @mojavelinux! Closing in favor of gh-12523.

PR #12531, opened 2023-01-12
Refactor SavedCookie for Cookie's deprecated method and fix typo in RememberMeAuthenticationFilter

twosom:
This PR solved issue for #12454 

and continue for #12455


jzheaux:
@twosom, will you please do two more bookkeeping items in preparation for merging? I would do them, but it appears that when I split the commit, it removes you as the contributor and I'd prefer you get credit.

1. Please separate into two commits. One for the typo fix and one for the deprecation work. That way each commit is an atomic change. It helps when backporting to earlier versions.
2. Please have the deprecation commit close the related issue. The way you do that is to add a line at the end line `Closes gh-12454`

Thanks!

PR #12532, opened 2023-01-12
Propagate security context via channel interceptor

artembilan:
The `SecurityContextChannelInterceptor` currently has an ability to populate a security context with an `Authentication` from message header. But doesn't look like there is a way in the framework to populate such a header and therefore propagate context downstream the messaging flow

* Extend a `SecurityContextChannelInterceptor` to support `postReceive()` contract for context population and `afterReceiveCompletion()` to clean it up, respectively
* Optimize `setup()` function logic to not re-populate context if it already has the same `Authentication`
* Add a logic to the `preSend()` to populate header with the value from current security context if it is missed in a message

The change is driven by this Spring Integration feature: https://docs.spring.io/spring-integration/reference/html/security.html#security-context-propagation.
There is now only a single class in the Spring Integration Security modules and it is that `SecurityContextPropagationChannelInterceptor`.
When we will have such a propagation feature in Spring Security for Messaging by itself, there won't be a reason in that module in Spring Integration at all.

### DIFF
--- a/messaging/src/main/java/org/springframework/security/messaging/context/SecurityContextPropagationChannelInterceptor.java
+++ b/messaging/src/main/java/org/springframework/security/messaging/context/SecurityContextPropagationChannelInterceptor.java
@@ -1,0 +1,164 @@
+/*
+ * Copyright 2002-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.messaging.context;
+
+import java.util.Stack;
+
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageChannel;
+import org.springframework.messaging.MessageHandler;
+import org.springframework.messaging.simp.SimpMessageHeaderAccessor;
+import org.springframework.messaging.support.ExecutorChannelInterceptor;
+import org.springframework.messaging.support.MessageBuilder;
+import org.springframework.security.authentication.AnonymousAuthenticationToken;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.authority.AuthorityUtils;
+import org.springframework.security.core.context.SecurityContext;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.core.context.SecurityContextHolderStrategy;
+import org.springframework.util.Assert;
+
+/**
+ * An {@link ExecutorChannelInterceptor} that takes an {@link Authentication} from the
+ * current {@link SecurityContext} (if any) in the
+ * {@link #preSend(Message, MessageChannel)} callback and stores it into an
+ * {@link #authenticationHeaderName} message header. Then sets the context from this
+ * header in the {@link #beforeHandle(Message, MessageChannel, MessageHandler)} and
+ * {@link #postReceive(Message, MessageChannel)} both of which typically happen on a
+ * different thread.
+ * <p>
+ * Note: cannot be used in combination with a {@link SecurityContextChannelInterceptor} on
+ * the same channel since both these interceptors modify a security context on a handling
+ * and receiving operations.
+ *
+ * @author Artem Bilan
+ * @since 6.2
+ * @see SecurityContextChannelInterceptor
+ */
+public final class SecurityContextPropagationChannelInterceptor implements ExecutorChannelInterceptor {
+
+	private static final ThreadLocal<Stack<SecurityContext>> originalContext = new ThreadLocal<>();
+
+	private SecurityContextHolderStrategy securityContextHolderStrategy = SecurityContextHolder
+		.getContextHolderStrategy();
+
+	private SecurityContext empty = this.securityContextHolderStrategy.createEmptyContext();
+
+	private final String authenticationHeaderName;
+
+	private Authentication anonymous = new AnonymousAuthenticationToken("key", "anonymous",
+			AuthorityUtils.createAuthorityList("ROLE_ANONYMOUS"));
+
+	/**
+	 * Create a new instance using the header of the name
+	 * {@link SimpMessageHeaderAccessor#USER_HEADER}.
+	 */
+	public SecurityContextPropagationChannelInterceptor() {
+		this(SimpMessageHeaderAccessor.USER_HEADER);
+	}
+
+	/**
+	 * Create a new instance that uses the specified header to populate the
+	 * {@link Authentication}.
+	 * @param authenticationHeaderName the header name to populate the
+	 * {@link Authentication}. Cannot be null.
+	 */
+	public SecurityContextPropagationChannelInterceptor(String authenticationHeaderName) {
+		Assert.notNull(authenticationHeaderName, "authenticationHeaderName cannot be null");
+		this.authenticationHeaderName = authenticationHeaderName;
+	}
+
+	public void setSecurityContextHolderStrategy(SecurityContextHolderStrategy strategy) {
+		this.securityContextHolderStrategy = strategy;
+		this.empty = this.securityContextHolderStrategy.createEmptyContext();
+	}
+
+	/**
+	 * Configure an Authentication used for anonymous authentication. Default is: <pre>
+	 * new AnonymousAuthenticationToken(&quot;key&quot;, &quot;anonymous&quot;,
+	 * 		AuthorityUtils.createAuthorityList(&quot;ROLE_ANONYMOUS&quot;));
+	 * </pre>
+	 * @param authentication the Authentication used for anonymous authentication. Cannot
+	 * be null.
+	 */
+	public void setAnonymousAuthentication(Authentication authentication) {
+		Assert.notNull(authentication, "authentication cannot be null");
+		this.anonymous = authentication;
+	}
+
+	@Override
+	public Message<?> preSend(Message<?> message, MessageChannel channel) {
+		Authentication authentication = this.securityContextHolderStrategy.getContext().getAuthentication();
+		if (authentication == null) {
+			authentication = this.anonymous;
+		}
+		return MessageBuilder.fromMessage(message).setHeader(this.authenticationHeaderName, authentication).build();
+	}
+
+	@Override
+	public Message<?> beforeHandle(Message<?> message, MessageChannel channel, MessageHandler handler) {
+		return postReceive(message, channel);
+	}
+
+	@Override
+	public Message<?> postReceive(Message<?> message, MessageChannel channel) {
+		setup(message);
+		return message;
+	}
+
+	@Override
+	public void afterMessageHandled(Message<?> message, MessageChannel channel, MessageHandler handler, Exception ex) {
+		cleanup();
+	}
+
+	private void setup(Message<?> message) {
+		Authentication authentication = message.getHeaders().get(this.authenticationHeaderName, Authentication.class);
+		SecurityContext currentContext = this.securityContextHolderStrategy.getContext();
+		Stack<SecurityContext> contextStack = originalContext.get();
+		if (contextStack == null) {
+			contextStack = new Stack<>();
+			originalContext.set(contextStack);
+		}
+		contextStack.push(currentContext);
+		SecurityContext context = this.securityContextHolderStrategy.createEmptyContext();
+		context.setAuthentication(authentication);
+		this.securityContextHolderStrategy.setContext(context);
+	}
+
+	private void cleanup() {
+		Stack<SecurityContext> contextStack = originalContext.get();
+		if (contextStack == null || contextStack.isEmpty()) {
+			this.securityContextHolderStrategy.clearContext();
+			originalContext.remove();
+			return;
+		}
+		SecurityContext context = contextStack.pop();
+		try {
+			if (this.empty.equals(context)) {
+				this.securityContextHolderStrategy.clearContext();
+				originalContext.remove();
+			}
+			else {
+				this.securityContextHolderStrategy.setContext(context);
+			}
+		}
+		catch (Throwable ex) {
+			this.securityContextHolderStrategy.clearContext();
+		}
+	}
+
+}
--- a/messaging/src/test/java/org/springframework/security/messaging/context/SecurityContextPropagationChannelInterceptorTests.java
+++ b/messaging/src/test/java/org/springframework/security/messaging/context/SecurityContextPropagationChannelInterceptorTests.java
@@ -1,0 +1,160 @@
+/*
+ * Copyright 2002-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.messaging.context;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageChannel;
+import org.springframework.messaging.MessageHandler;
+import org.springframework.messaging.simp.SimpMessageHeaderAccessor;
+import org.springframework.messaging.support.MessageBuilder;
+import org.springframework.security.authentication.AnonymousAuthenticationToken;
+import org.springframework.security.authentication.TestingAuthenticationToken;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.core.context.SecurityContextHolderStrategy;
+import org.springframework.security.core.context.SecurityContextImpl;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+@ExtendWith(MockitoExtension.class)
+public class SecurityContextPropagationChannelInterceptorTests {
+
+	@Mock
+	MessageChannel channel;
+
+	@Mock
+	MessageHandler handler;
+
+	MessageBuilder<String> messageBuilder;
+
+	Authentication authentication;
+
+	SecurityContextPropagationChannelInterceptor interceptor;
+
+	@BeforeEach
+	public void setup() {
+		this.authentication = new TestingAuthenticationToken("user", "pass", "ROLE_USER");
+		this.messageBuilder = MessageBuilder.withPayload("payload");
+		this.interceptor = new SecurityContextPropagationChannelInterceptor();
+	}
+
+	@AfterEach
+	public void cleanup() {
+		this.interceptor.afterMessageHandled(this.messageBuilder.build(), this.channel, this.handler, null);
+		SecurityContextHolder.clearContext();
+	}
+
+	@Test
+	public void preSendDefaultHeader() {
+		SecurityContextHolder.getContext().setAuthentication(this.authentication);
+		Message<?> message = this.interceptor.preSend(this.messageBuilder.build(), this.channel);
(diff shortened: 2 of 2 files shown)


messaging/src/test/java/org/springframework/security/messaging/context/SecurityContextChannelInterceptorTests.java
```diff
@@ -175,6 +169,24 @@ public void beforeHandleUserSet() {
...
 
+	@Test
+	public void postReceiveUserSet() {
+		this.messageBuilder.setHeader(SimpMessageHeaderAccessor.USER_HEADER, this.authentication);
+		this.interceptor.postReceive(this.messageBuilder.build(), this.channel);
+		assertThat(SecurityContextHolder.getContext().getAuthentication()).isSameAs(this.authentication);
+	}
+
+	@Test
+	public void authenticationIsPropagatedFromPreSendToPostReceive() {
```
jzheaux, line 180: I don't think I'm understanding this test. What is the real-world scenario that it is simulating?

messaging/src/test/java/org/springframework/security/messaging/context/SecurityContextChannelInterceptorTests.java
```diff
@@ -138,13 +139,6 @@ public void preSendUserNotSet() {
 		assertAnonymous();
 	}
 
-	// SEC-2845
-	@Test
```
jzheaux, line 142: I don't think we want to remove tests, especially those linked to an issue. Is there a reason for removing this one?

jzheaux (changes requested):
@artembilan, thanks for the PR! I've left some questions inline.

messaging/src/test/java/org/springframework/security/messaging/context/SecurityContextChannelInterceptorTests.java
artembilan: The use-case is exactly what the change in this PR is needed. We put a message into a `Queue` and we don't know which thread, where and when we going to pull from that queue. In preSend we store auth ctx into message header and in the `postReceive` we do restore that ctx into the current thread which has just called `receive` on a queue.

All of that logic is really covered by the `SecurityContextChannelInterceptor` with a small adjustments to a missed functionality for `postReceive()`.

messaging/src/test/java/org/springframework/security/messaging/context/SecurityContextChannelInterceptorTests.java
artembilan: Oh! I didn't mention that in the commit comments. 
Sorry.
I removed it deliberately just because this test is really a plain copy of `preSendUserNotSet()`.
Perhaps some logic is missed because their names are different, but that already not relevant to this PR.

Well, I can restore it if you insist, but let's see if this test can be improved somehow to reflect its purpose or so!

artembilan: According the name of this test, if we do `setAnonymousAuthentication()`, this test becomes identical to the `preSendUsesCustomAnonymous()`. Just because that one also does set a user into a message to intercept.
So, one of this tests is redundant this or other way.

messaging/src/main/java/org/springframework/security/messaging/context/SecurityContextChannelInterceptor.java
```diff
@@ -91,7 +98,18 @@ public void setAnonymousAuthentication(Authentication authentication) {
 
 	@Override
 	public Message<?> preSend(Message<?> message, MessageChannel channel) {
-		setup(message);
```
jzheaux, line 94: It looks like what this previously did was set the security context to whatever user was in the message (anonymous if no user).

With the new logic, that won't happen anymore. Instead, if there is no user, then an authenticated security context will remain as-is.

Also, it seems that if it is an authenticated security context, then it doesn't get pushed to the stack. This creates problems if `cleanup` is called later on. For example, if the message header is empty and `SecurityContextHolder` is set to a value, when `afterSendCompletion` is done, the security context is emptied instead of set back to what it what before.

To remain passive, I think that `setup` needs to be called unconditionally.

Would it make sense to do the following instead:

```java
setup(message);
if (message.getHeaders().get(this.authenticationHeaderName) == null) {
    Authentication authentication = this.securityContextHolderStrategy.getContext().getAuthentication();
    message = MessageBuilder.fromMessage(message).setHeader(this.authenticationHeaderName, authentication).build();
}
```

In this way, the `SecurityContextHolder` stack is preserved and the message gets the current security context as a value (anonymous if there is no security context).

messaging/src/test/java/org/springframework/security/messaging/context/SecurityContextChannelInterceptorTests.java
jzheaux: The thing I'm not understanding is why would you want the `SecurityContextHolder` to be cleared after `afterMessageHandled` is called? 

The following flow feels more natural: `preSend` saves the current security context and sets a new one according to the message contents, then `cleanup` restores what it was before.

messaging/src/main/java/org/springframework/security/messaging/context/SecurityContextChannelInterceptor.java
```diff
@@ -118,14 +141,19 @@ public void setSecurityContextHolderStrategy(SecurityContextHolderStrategy strat
 
 	private void setup(Message<?> message) {
 		SecurityContext currentContext = this.securityContextHolderStrategy.getContext();
+		Object user = message.getHeaders().get(this.authenticationHeaderName);
+		Authentication authentication = getAuthentication(user);
+
+		if (authentication.equals(currentContext.getAuthentication())) {
```
jzheaux, line 147: I'm not sure this is a good idea. Wouldn't this cause `cleanup` to break? If `setup` is called three times and `cleanup` is subsequently called three times, if one of the stack pushes was skipped, then `cleanup` will restore incorrect values. It seems like this would create problems for the scope of each authentication change.

jzheaux (changes requested):
Thanks, @artembilan! I've left some additional feedback inline.

messaging/src/main/java/org/springframework/security/messaging/context/SecurityContextChannelInterceptor.java
artembilan: And that is exactly not OK for the change I'm proposing.
The `setup(message);` does set an `anonymous` if no header in the message, but my goal is to take the current one and set it into header.

Perhaps `cleanup()` must not pop if we haven't pushed...

The goal of this change is to add a new behavior to populate the header from the current context not changing it and not pushing anything to the stack. If header is present and it is not the same what we have on a context, then yes we do push.

The previous version if this interceptor is indeed what you are describing: push if header present or not and this does not give us what we need on the receiving side. Perhaps it will work if we would add that header in the other interceptor upfront this one, but why if there is already one in the current `SecurityContext`.

Let me see what we can do with `cleanup()` to follow this new logic!

jzheaux: > The setup(message); does set an anonymous if no header in the message, but my goal is to take the current one and set it into header.

It makes me wonder if the logic you are wanting to have should be in a separate interceptor. This interceptor wasn't originally intended as propagation but as an analogy to `SecurityContextHolderFilter`. It can't behave in opposing ways based on the same input (no header).

artembilan: Sure! I also thought about that.
Currently the `SecurityContextChannelInterceptor` really does propagate the ctx via headers if the value is present in the message.
To avoid duplication an interceptor you suggest and I've thought for would only load the current ctx into a header and this `SecurityContextChannelInterceptor` would be used as a second one to propagate the ctx.
And that is going to be a typical configuration.
This way end-users may end up with the question why do we need two interceptors when only one would do the trick for both use-cases.

If you still think that `SecurityContextChannelInterceptor` must remain as it is, then I'll work on a  `SecurityContextPropagationChannelInterceptor`  even if it is going to have some logic similar to the `SecurityContextChannelInterceptor`.
The only different with this two interceptors would be only in a way how they deal with current ctx and header: one of them takes header and set it into a current ctx.
Another takes current ctx in a `preSend` and populate int on other threads in the `postReceive()` and `beforeHandle()`.
Well, probably `afterSendCompletion()` won't do anything in this new interceptor, since we really don't affect the current ctx, but just takes it into the message header.

Let me know your decision and I'll proceed from here!

Thank you for feedback!

jzheaux: The existing interceptor is designed to populate the security context with the header; this PR wants to populate the header with the security context. While there is some overlap, I think these two goals are ultimately at odds. One is (conceptually) `MessageToSecurityContextInterceptor` and the other is `SecurityContextToMessageInterceptor`.

> And that is going to be a typical configuration.

It seems to me that you'd want to populate the security context (`SecurityContextChannelInterceptor`) on the inbound channel and populate the message (this PR) on the outbound channel. So when we talk about typical configuration, I am doubting that they would be in the same interceptor chain. Indeed, it seems like they'd cancel out each other just a bit.

If it sounds like I'm understanding correctly, then yes please, let's move this functionality into a separate filter. I wonder if `SecurityContextOutboundChannelInterceptor` would be an appropriate name. Since my recommendation is based on my beginner-level understanding of message channels, please feel free to help me see things differently.

artembilan: Thanks, Josh, for confirmation!

I'm looking not into a separate `SecurityContextPropagationChannelInterceptor` impl.

To clean things a little:
No, we don't talk about channel adapters. 
We talk about a thread boundaries provided by the `ExecutorChannel` (with some variations), where `send` operation happens on one thread and then `handle` on the other. That's exactly why we have that special `ExecutorChannelInterceptor` abstraction. 
Another one is a `QueueChannel` where `send` puts the message onto the queue and `receive` pulls it from there, which typically happens on a scheduled thread.
Both of those threads just two separate ends of the same channel, therefore we have a scope of the message as a unit to carry all the information via interceptor hooks: https://www.enterpriseintegrationpatterns.com/patterns/messaging/MessageChannel.html

artembilan:
@jzheaux ,

any feedback, please?

artembilan:
Here is some simple Spring Boot sample to demonstrate how a new `SecurityContextPropagationChannelInterceptor` helps to propagate security context from one thread to another via `MessageChannel` in between: https://github.com/artembilan/sandbox/blob/master/spring-integration-security-context-propagation/HELP.md.
Thanks

messaging/src/main/java/org/springframework/security/messaging/context/SecurityContextPropagationChannelInterceptor.java
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	@Override
+	public void afterMessageHandled(Message<?> message, MessageChannel channel, MessageHandler handler, Exception ex) {
+		cleanup(message);
+	}
+
+	private void setup(Message<?> message) {
+		Authentication authentication = message.getHeaders().get(this.authenticationHeaderName, Authentication.class);
+		if (authentication != null) {
```
jzheaux, line 112: Instead of conditionally adding to the stack, can you use the null object pattern that `SecurityContextChannelInterceptor` does?

messaging/src/main/java/org/springframework/security/messaging/context/SecurityContextPropagationChannelInterceptor.java
```diff
@@ -0,0 +1,164 @@
...
+	 * by this interceptor from the
+	 * {@link #afterMessageHandled(Message, MessageChannel, MessageHandler, Exception)}
+	 * where handling operation is finished on the other thread. Must be called explicitly
+	 * downstream after {@link org.springframework.messaging.PollableChannel#receive()}
+	 * operation and processing with the message is done.
+	 * @since 6.2
+	 */
+	public void cleanup(Message<?> message) {
+		Authentication authentication = message.getHeaders().get(this.authenticationHeaderName, Authentication.class);
+		if (authentication != null
```
jzheaux, line 139: If the message always has a value set, then I think this if statement can go away as well.

messaging/src/main/java/org/springframework/security/messaging/context/SecurityContextPropagationChannelInterceptor.java
jzheaux: In addition to removing the if statement, it also aligns well with the Spring Security Filter Chain which adds an anonymous authentication as well.

messaging/src/main/java/org/springframework/security/messaging/context/SecurityContextPropagationChannelInterceptor.java
```diff
@@ -0,0 +1,164 @@
...
+	 * {@link #authenticationHeaderName} header value is equal to the current
+	 * {@link Authentication}, and restores one from the stack(if any). Used automatically
+	 * by this interceptor from the
+	 * {@link #afterMessageHandled(Message, MessageChannel, MessageHandler, Exception)}
+	 * where handling operation is finished on the other thread. Must be called explicitly
+	 * downstream after {@link org.springframework.messaging.PollableChannel#receive()}
+	 * operation and processing with the message is done.
+	 * @since 6.2
+	 */
+	public void cleanup(Message<?> message) {
```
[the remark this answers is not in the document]
jzheaux: I'm not really clear on why someone can't call `afterMessageHandled`, for example.

jzheaux (changes requested):
Thanks for moving this out into another class, and also for the sample! Please see my next round of questions/comments inline.

messaging/src/main/java/org/springframework/security/messaging/context/SecurityContextPropagationChannelInterceptor.java
artembilan: This interceptor is not about authentication, but rather accepting existing auth from message header or ignoring otherwise.
If you would like to have an anonymous auth, then use a `SecurityContextChannelInterceptor` in the flow after this propagation has done its "void" logic.
Plus it doesn't look very optimal to set a context where it was not requested to be set.

messaging/src/main/java/org/springframework/security/messaging/context/SecurityContextPropagationChannelInterceptor.java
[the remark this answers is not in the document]
artembilan: The `afterMessageHandled(Message<?> message, MessageChannel channel, MessageHandler handler, Exception ex)` is an interceptor contract and it is called by the framework in specific places.
Yes, one of them is indeed from message handler, when `this.messageHandler.handleMessage(message);` is done.
But in case of `PollableChannel` and that `postReceive()`, respectively, the handler is not involved in the process yet, therefore we cannot `cleanup()` automatically.
The suggestion is to do that manually downstream, where one of the option is like this `handle(mySecurityContextPropagationChannelInterceptor::cleanup)` - just to make it realigned with our Messaging API.

messaging/src/main/java/org/springframework/security/messaging/context/SecurityContextPropagationChannelInterceptor.java
artembilan: The message is not always have that value set. We indeed can fail with missed one here: what is the point to call this clean up if no auth header.
But at the same time the header is no always populated: see `preSend()`.

I really don't see a benefit from the `anonymous` propagation since we don't auth here, but, yeah, propagate whatever is there in the context or void otherwise.

artembilan:
Just pushed the rest of requested changes.

Thanks

messaging/src/main/java/org/springframework/security/messaging/context/SecurityContextPropagationChannelInterceptor.java
jzheaux: My main concern is conditionally adding and removing from a stack. It's an opportunity for memory leaks. 

Consider, for example, if the authentication changes between `postReceive` and `afterMessageHandled`. Then, the stack will never get cleaned up.

I don't think that comparing authentication instances is a reliable way to ensure the stack values get popped correctly.

The check performed at cleanup is trying to figure out if something was placed on the stack so that it can be removed. Instead, having a placeholder value would make the logic simpler and less likely to experience a memory leak.

Would adding an anonymous value here (I see this as meaning "no authentication was propagated") prevent you from being able to use the interceptor or make it difficult?

messaging/src/main/java/org/springframework/security/messaging/context/SecurityContextPropagationChannelInterceptor.java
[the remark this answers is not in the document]
jzheaux: I guess I'm unclear since the propagation filter in Spring Messaging doesn't do this same thing. Is there a way that this PR could focus on enabling the existing functionality that Spring Messaging wants to port over? We can consider other use cases in the future.

messaging/src/main/java/org/springframework/security/messaging/context/SecurityContextPropagationChannelInterceptor.java
artembilan: The `afterMessageHandled()` is not called after `postReceive()`. If that is the case, then another instance of the interceptor has to be used downstream. And that's exactly why I have introduced that `public cleanup(Message)`. There is no connection between received message from the `PollableChannel` and downstream handler. But the point is that auth has to be populated to the context to make that handler downstream called in a secure way.

artembilan: OK. I'll look today into your anonymous suggestion.
No, it won't make the interceptor not working since the point is to make this as transparent for Messaging as possible and fully rely on existing features in Spring Security. Therefore if anonymous is OK to reject the call for `@PreAuthorize`, I'm fine to go ahead. Only my concern that it is sub-optimal to always populate into the stack whatever was not requested.

artembilan: Another note: with an anonymous support in this interceptor it is going to be very similar to what we have so far in the `SecurityContextChannelInterceptor`. And therefore we slowly coming back to whatever I suggested to do in my original commit for this PR.

messaging/src/main/java/org/springframework/security/messaging/context/SecurityContextPropagationChannelInterceptor.java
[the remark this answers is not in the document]
artembilan: I'm not sure what you are talking about. That was really a main point of this PR: move context propagation support out of Spring Integration and make it available for the whole Spring Messaging.

messaging/src/main/java/org/springframework/security/messaging/context/SecurityContextPropagationChannelInterceptor.java
jzheaux: I see your point here about them being similar in many ways.

I think they are fundamentally different since `preSend` wants to do the opposite thing in each. The nice thing is we can always combine them in an abstract class or a proper subclass if that's what it comes down to. I don't mind people declaring that their intent is to authenticate vs propagate.

messaging/src/main/java/org/springframework/security/messaging/context/SecurityContextPropagationChannelInterceptor.java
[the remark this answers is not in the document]
jzheaux: What I'm saying is the Spring Integration interceptor, AFAICT, does not expose a `cleanup` method like this PR does. In the original one, `cleanup` is private.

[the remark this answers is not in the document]
artembilan: Well, I looked into that one more time and I find it as a but since my current interceptor in Spring Integration doesn't clean up thread local on `postReceive`. That `SecurityContextHolder.clearContext()` will be enough for context, but `EMPTY_CONTEXT` will remain. Therefore I still would like to have this public `cleanup()` for end-user to be called in the end of secured message handling downstream.

I can come up with some extra `Closeable` header to be added to the message, but that doesn't look cleaner or transparent for end-user: still has to be called some way from end-user code.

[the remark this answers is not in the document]
jzheaux: Thanks for the explanation, that's helpful.

> that interceptor ... a plain SecurityContextHolder.clearContext(); is enough

Either one seems problematic, TBH. Again, though, this can be the result of my not understanding how things work on the Integration side of things. My thinking is that an interceptor should be self-sufficient; neither a call to `cleanup` nor a call to `SecurityContextHolder.clearContext()` should be necessary outside the interceptor as that leaks the interceptor's internals.

> The afterMessageHandled() is not called after postReceive(). 

(Since `cleanup` has been mentioned in both threads, I'm bringing this comment :point_up: from the other thread to give us one place to focus.)

I don't think I understand how an interceptor works, then, but I may have missed the full context of this comment. When I use the sample you provided (thank you, again), both `postRecieve` and `afterMessageHandled` are invoked through `PollingConsumer`. Can you give a scenario -- possibly in the sample -- where something other than the interceptor needs to call `cleanup`?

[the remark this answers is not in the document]
artembilan: Oh! I see now 😄 .
That my `PollingConsumer` indeed has a respective logic:
```
			this.handler.handleMessage(theMessage);
			if (!CollectionUtils.isEmpty(interceptorStack)) {
				triggerAfterMessageHandled(theMessage, null, interceptorStack);
			}
```

I introduced it there 10 years ago. 
My bad.
Yes, removing this clean up from public API.

Sorry for the noise!

jzheaux (approved):

PR #12536, opened 2023-01-13
move code comment to callout

mojavelinux:
Code comments should be placed in callouts, especially when they contain formatting and links. Otherwise, it breaks the layout of the page, as you can see in this code listing:

screenshot

### DIFF
--- a/docs/modules/ROOT/pages/servlet/architecture.adoc
+++ b/docs/modules/ROOT/pages/servlet/architecture.adoc
@@ -71,29 +71,26 @@ The following listing shows pseudo code of `DelegatingFilterProxy`:
 .`DelegatingFilterProxy` Pseudo Code
 ====
 .Java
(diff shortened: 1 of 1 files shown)


rwinch:
Thanks for finding and fixing this one. Is there a way for this to work as a comment. Part of the reason we do it in comments is to ensure copy / pasted code still has the information in it.

mojavelinux:
Yep. That was the UI PR I sent about the code phrase. But using formatting
substitutions on a code block is very non portable and thus I consider it
an anti pattern.

mojavelinux:
In theory, it would be possible to have the copy button include the content from the callout list, similar to how the code modifies the copy content for a folded code block. But if there is still a preference to have the comment directly in the code, I would strongly advise to have it just be plain text. The link and formatting isn't going to get picked up on a copy anyway.

PR #12543, opened 2023-01-13
upgrade Asciidoctor Tabs, the Antora Gradle plugin, and Node.js

mojavelinux:
Note that this PR depends on https://github.com/spring-io/antora-ui-spring/pull/62


mojavelinux:
This PR also upgrades Node.js to 19. That should make the production build slightly faster. It's not needed in the content branches as the impact there would probably not be measurable. The content branch builds can use whatever version of Node.js the plugin pulls in by default.

PR #12545, opened 2023-01-14
use single gradle command/daemon for production docs build

mojavelinux:
I'm not requesting that you merge this right away, but to consider it. This PR reconfigures the production docs build to use the same gradle command/daemon throughout the build. It first installs Gradle onto the PATH. Then it rewrites the collector commands to use that gradle command instead of gradlew. By doing so, the build does not have to keep downloading Gradle for each reference where a different version is requested by gradlew.

Right now, this works across all references. If there comes a time when it does not, it would be possible to reconfigure the extension to leave those references alone and only use the shared gradle command for ones where it does work.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -18,24 +18,21 @@ jobs:
       uses: actions/checkout@v3
       with:
         fetch-depth: 5
-    - name: Set Up Gradle
-      uses: spring-io/spring-gradle-build-action@v1
-      with:
-        java-version: '17'
-        distribution: temurin
-    - name: Scrub Gradle Cache
-      # Remove some files from the Gradle cache, so they aren't cached by GitHub Actions.
-      # Restoring these files from a GitHub Actions cache might cause problems for future builds.
-      run: |
-        rm -f /home/runner/.gradle/caches/modules-2/modules-2.lock
-        rm -f /home/runner/.gradle/caches/modules-2/gc.properties
     - name: Set up refname build
       if: github.event.inputs.build-refname
       run: |
-        git fetch --depth 1 https://github.com/$GITHUB_REPOSITORY ${{ github.event.inputs.build-refname }}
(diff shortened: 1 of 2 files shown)


mojavelinux:
This PR also simplifies the GitHub Actions workflow by using the gradle/gradle-build-action@v2 action directly. It already manages the Gradle cache, so there is no need for the extra build logic.

rwinch:
Thanks for the PR @mojavelinux 

I think we want to continue to use https://github.com/spring-io/spring-gradle-build-action because it does things like set the user property which is necessary for our Gradle Enterprise arrangement, validates the wrapper, etc. Internally gradle-build-action uses the gradle one, so perhaps we can delete the step that is clearing entries from the Gradle cache.

I agree that running this command for each tag/branch is quite slow, but in regards to using the same Gradle version for the whole build, I'm not sure this is going to help the build performance much. There are 4 Gradle distributions being used currently and the difference for the build time on my machine is negligible between having the Gradle distributions cached or not. Perhaps I'm missing something here?

mojavelinux:
You are right that it doesn't seem to have a measurable impact on the build time (as I thought it might). Over time, I think the number of different Gradle distributions (and thus daemons) will add up, but GitHub Actions is so fast at downloading them, and the build cache doesn't really help that this is probably not a worthwhile exercise.

>  Internally gradle-build-action uses the gradle one, so perhaps we can delete the step that is clearing entries from the Gradle cache.

That's the main step I was trying to get rid of in making that change, so perhaps we should just isolate that as a separate PR.

> I think we want to continue to use spring-io/spring-gradle-build-action because it does things like...

I think the main issue is that it's using an older version of the gradle action, hence all the warnings in GitHub Actions. Perhaps the solution there is to get that action updated so its using the v3 version of the Java action and the v2 version of the Gradle action.

rwinch:
> Perhaps the solution there is to get that action updated so its using the v3 version of the Java action and the v2 version of the Gradle action.

Thanks. I've done that via https://github.com/spring-projects/spring-security/pull/12563 and https://github.com/spring-io/spring-gradle-build-action/pull/1

PR #12550, opened 2023-01-17
prime atlas extension for use in playbook template

mojavelinux:
- declare atlas extension as package
- declare release line extension as package
- enable release line extension by default
- configure atlas extension, but do not enable by default


mojavelinux:
This PR configures the playbook for per branch builds so they can be used with the Atlas extension when desired. From the branch's docs directory, enable the extension using this command:

```
../gradlew antora --extension @antora/atlas-extension
```

This isn't a critical feature, which is why it's not enabled by default. However, it can be used for a final xref check before a commit.

rwinch:
Thanks for the Pull Request! This is now merged :smile:

Issue #12031, opened 2022-10-17
Allow configuring SecurityContextRepository for BasicAuthenticationFilter


rwinch:
We should allow configuring SecurityContextRepository on BasicAuthenticationFilter so that users do not need to rely on `SecurityContextPersistenceFitler`. Users that want BasicAuthenticationFilter to use a custom repository right now can do this:

```java
http
	.httpBasic(b -> b.addObjectPostProcessor(new ObjectPostProcessor<BasicAuthenticationFilter>() {
			@Override
			public <O extends BasicAuthenticationFilter> O postProcess(O filter) {
				filter.setSecurityContextRepository(new HttpSessionSecurityContextRepository());
				return filter;
			}
		}));
```

However, it should be easier to configure with an explicit option.

evgeniycheban:
Hi, @rwinch @jzheaux I'd like to work on it.

evgeniycheban:
Hi, @rwinch @jzheaux I've created a PR that adds a new method 
`securityContextRepository(SecurityContextRepository securityContextRepository)` that can be used to customize it.

PR #12556, opened 2023-01-18
Allow configuring SecurityContextRepository for BasicAuthenticationFilter

evgeniycheban:
Allow configuring SecurityContextRepository for BasicAuthenticationFilter

Closes gh-12031

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HttpBasicConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HttpBasicConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/configurers/HttpBasicConfigurer.java
```diff
@@ -192,6 +212,7 @@ public void configure(B http) {
 		AuthenticationManager authenticationManager = http.getSharedObject(AuthenticationManager.class);
 		BasicAuthenticationFilter basicAuthenticationFilter = new BasicAuthenticationFilter(authenticationManager,
 				this.authenticationEntryPoint);
+		basicAuthenticationFilter.setSecurityContextRepository(this.securityContextRepository);
```
sjohnr, line 215: I'm wondering if it makes more sense to provide no default for `securityContextRepository` in this configure (therefore it defaults to `null`), and only set on the filter if not `null`?

sjohnr (commented):
Thanks @evgeniycheban! I've added on review comment below.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/HttpBasicConfigurer.java
evgeniycheban: Assigning a default value and performing a `null` check in the setter seems like a common strategy that is used elsewhere, so I would stay here with this approach and, perhaps, later align other methods in this class.

sjohnr: Hi @evgeniycheban! Sorry for the delay in responding.

> Assigning a default value and performing a `null` check in the setter seems like a common strategy that is used elsewhere

I was thinking specifically of configurers where it seems the pattern is often to let the configured `Filter` contain the default and only override the filter (via the setter) if not null.

Can you share an example in other configurers where you see this strategy used? Or are you referring mostly to other classes (besides configurers) in Spring Security?

evgeniycheban: @sjohnr yes, I was referring to other classes where this pattern is commonly used.

evgeniycheban: @sjohnr I've updated the PR to remove the default value and add a `null` check.

sjohnr: Awesome! I’ll work on getting this merged shortly.

sjohnr:
Thanks @evgeniycheban! Merged via 59829321a8045c26f19c725f3ae081ac297e4079

Issue #12378, opened 2022-12-14
Fix CSRF protection provided by @EnableWebSocketSecurity / Stomp


emopti-jrufer:
CSRF protection provided by `@EnableWebSocketSecurity` is broken. I have identified 2 things that prevent the `CsrfChannelInterceptor` from validating the CSRF token from the CONNECT headers.

First problem encounted was the stack trace below. The `CsrfChannelInterceptor` was trying to access the `CsrfToken` when processing the CONNECT message. The changes to made in 475b3bb6bbdc737ba67820d8ca4e557b3e030e7c and d94677f87e3238900435041d8dc20b33b3e36719 to defer the loading of the token caused the deferred supplier to try to access the http upgrade `HttpServletRequest` after it has gone out of scope.

Attempting to come up with a fix for the first problem I added an additional HandshakeInterceptor that forced eager loading the CsrfToken while they upgrade HttpServeletRequest was still in scope. I added the source below for clarity.

This overcame the first problem but then the `CsrfChannelInterceptor` started throwing `InvalidCsrfTokenException`. After investigating it became apparent the `CsrfChannelInterceptor` was only design to work with tokens that were not masked. Since `XorCsrfTokenRequestAttributeHandler` is now the default implementation for `CsrfTokenRequestHandler` the `CsrfChannelInterceptor` needs to be modified to have the same or similar logic as the `CsrfFilter#doFilterInternal`.

```
 stack_trace: org.springframework.messaging.MessageDeliveryException: Failed to send message to ExecutorSubscribableChannel[clientInboundChannel]
	at org.springframework.messaging.support.AbstractMessageChannel.send(AbstractMessageChannel.java:149)
	at org.springframework.messaging.support.AbstractMessageChannel.send(AbstractMessageChannel.java:125)
	at org.springframework.web.socket.messaging.StompSubProtocolHandler.handleMessageFromClient(StompSubProtocolHandler.java:310)
	at org.springframework.web.socket.messaging.SubProtocolWebSocketHandler.handleMessage(SubProtocolWebSocketHandler.java:335)
	at org.springframework.web.socket.handler.WebSocketHandlerDecorator.handleMessage(WebSocketHandlerDecorator.java:75)
	at org.springframework.web.socket.handler.WebSocketHandlerDecorator.handleMessage(WebSocketHandlerDecorator.java:75)
[...]

PR #12562, opened 2023-01-18
Add XorCsrfChannelInterceptor

sjohnr:


messaging/src/main/java/org/springframework/security/messaging/web/socket/server/CsrfTokenHandshakeInterceptor.java
```diff
@@ -41,11 +44,15 @@ public final class CsrfTokenHandshakeInterceptor implements HandshakeInterceptor
...
 		HttpServletRequest httpRequest = ((ServletServerHttpRequest) request).getServletRequest();
-		CsrfToken token = (CsrfToken) httpRequest.getAttribute(CsrfToken.class.getName());
-		if (token == null) {
+		DeferredCsrfToken deferredCsrfToken = (DeferredCsrfToken) httpRequest
+				.getAttribute(DeferredCsrfToken.class.getName());
+		if (deferredCsrfToken == null) {
 			return true;
 		}
-		attributes.put(CsrfToken.class.getName(), token);
+		CsrfToken csrfToken = deferredCsrfToken.get();
```
rwinch, line 52: I think it is worth leaving a comment as to why we cannot use the DeferredCsrfToken directly and why we must copy the values over into a new token.

It might be nice to have a test that verifies we copied things over. The test should have similar documentation listed above.

rwinch (changes requested):
Thanks @sjohnr Overall this looks good. Once you have resolved the comments, from my perspective you can merge this.

sjohnr:
**Note:** Once this PR is merged, we can close gh-12378 by making `XorCsrfChannelInterceptor` the default in `WebSocketMessageBrokerSecurityConfiguration.csrfChannelInterceptor`.

Issue #12538, opened 2023-01-13
Custom RelayState with OpenSamlLogoutRequestResolver


ghaege:
**Describe the bug**
RelayState Param with SamlLogoutRequest is ignored.
Instead its replaced with a hardcoded UUID and cannot be customized.

**To Reproduce**
If we call for example the logout api via: http://logout-host:1234/logout?RelayState=http://app-host:5678
the RelayState Param is never processed.
```
final class OpenSamlLogoutRequestResolver {
..
    String relayState = UUID.randomUUID().toString();
```

**Expected behavior**
RelayState should processed or at least be customizable via 
```
@Bean 
Saml2LogoutRequestResolver logoutRequestResolver(
        RelyingPartyRegistrationResolver registrations) {
    
    OpenSaml4LogoutRequestResolver logoutRequest = 
            new OpenSaml4LogoutRequestResolver(registrations);
    logoutRequest.setRelayStateResolver(this::resolveCustomRelayState);
    
    return logoutRequest;
}
```

**Solution**
The solution should be very similar to #11065

jzheaux:
Sounds great, @ghaege. Can you provide a PR that adds the feature?

ghaege:
yes, but it might take a couple days

PR #12565, opened 2023-01-19
fixed Custom RelayState with OpenSamlLogoutRequestResolver

ghaege:


jzheaux (changes requested):
Thanks, @ghaege! New features need to go into the latest minor release, so will you please rebase on `main` so that this goes out in 6.1?

Also, will you please update your commit messages in the following two ways:
- The title should be in the present tense and should probably say "Add" instead of "Fix" (since this is new behavior)
- The commit message should have "Closes gh-12538" at the bottom, like this:

```
Add RelayState Customizer to SAML Logout

Closes gh-12538
```

ghaege:
replaced by https://github.com/spring-projects/spring-security/pull/12582

PR #12569, opened 2023-01-20
Using modern Java features

krzyk:
Used few modern Java features to reduce boilerplate and hopefully increase readability.
switch expressions, few lambdas, patterns for instanceof, also replaced few cases of `new Long` with correct `Long.valueOf` (because constructor for Long is deprecated).

I didn't replace all switches, because in some cases moving to switch expressions or new switch statements didn't reduce the amount of code.


zucchivan (commented):
LGTM! 

Good job. =)

jzheaux:
Thanks, @krzyk! This is now merged into `main`.

PR #12571, opened 2023-01-21
add docs to local Antora extensions

mojavelinux:
Add docs to the local Antora extensions in the docs-build (playbook) branch. These extensions are only used for the building the production docs site.


sjohnr:
Merged via 53acce43bc5d9bb49ea8ea8fc5166888d3ae8668

PR #12582, opened 2023-01-25
Add RelayState Customizer to SAML Logout

ghaege:
Add RelayState Customizer to SAML Logout

Closes gh-12538

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/authentication/logout/OpenSaml4LogoutRequestResolver.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/authentication/logout/OpenSaml4LogoutRequestResolver.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/authentication/logout/OpenSaml4LogoutRequestResolver.java
```diff
@@ -23,18 +23,19 @@
 import jakarta.servlet.http.HttpServletRequest;
```
jzheaux, line 23: In each file you modify, please update the copyright year to `2023`.

jzheaux (changes requested):
Thanks, @ghaege! I appreciate the contribution, and sorry that it took me a bit to get back to you. I've left some feedback inline. When you submit your update, please squash your commits into a single one.

In addition to that, please have that single commit include the bug that it being closed, like so:

```
Your Commit Title

Closes gh-12538
```

Thanks!

ghaege:
Thanks, @jzheaux for your feedback.
But since your recommendations are mainly by "polishing-nature and spring internal rules" and I don't have time, I would like to ask you to do the changes.

Thanx for your understanding.

jzheaux:
Thanks, @ghaege. No problem, we'll make sure this gets in. 

To keep each commit atomic, I am going to need to split the unrelated cleanups [into a separate commit](https://github.com/spring-projects/spring-security/blob/main/CONTRIBUTING.adoc#squash-commits). This will remove your authorship, and I want to make sure you get credit. So, what I propose is that I will instead add the relay state feature myself and add you as a co-author in the commit message. Does that sound okay to you? I'll go ahead and proceed this way if I don't hear from you in the next few days.

If you want to resubmit the cleanups once you have time again, then you can always rebase your `gh-12538` branch with `main` and resubmit the PR.

ghaege:
Perfect - do it like you proposed.

jzheaux (approved):

Issue #12294, opened 2022-11-25
Allow java.lang.Long in whitelist for Jackson security


mvitz:
**Expected Behavior**

When using `spring-security-oauth2-client` in combination with `spring-session-data-redis` and enabling JSON serializing via

```
@Bean
public RedisSerializer<Object> springSessionDefaultRedisSerializer() {
    var om = new ObjectMapper();
    om.registerModules(SecurityJackson2Modules.getModules(loader));
    return new GenericJackson2JsonRedisSerializer(om);
}
```

the session should be successfully deserialized when containing a type of `java.lang.Long`.

**Current Behavior**

The deserialization fails with the exception:
```
org.springframework.data.redis.serializer.SerializationException: Could not read JSON: The class with java.lang.Long and name of java.lang.Long is not in the allowlist. If you believe this class is safe to deserialize, please provide an explicit mapping using Jackson annotations or by providing a Mixin. If the serialization is only done by a trusted source, you can also enable default typing. See https://github.com/spring-projects/spring-security/issues/4370 for details (through reference chain: org.springframework.security.core.context.SecurityContextImpl["authentication"]->org.springframework.security.oauth2.client.authentication.OAuth2AuthenticationToken["principal"]->org.springframework.security.oauth2.core.oidc.user.DefaultOidcUser["authorities"]->org.springframework.security.oauth2.core.oidc.user.OidcUserAuthority["idToken"]->org.springframework.security.oauth2.core.oidc.OidcIdToken["claims"])
```

**Context**

At least our Okta returns a claim `ver` with a `java.lang.Long` of `1`. When persisting the spring-security session with spring-session as JSON within redis this is serialized as

```
...
                "ver": [
                  "java.lang.Long",
                  1
                ],
...
```

and can not be deserialized because `java.lang.Long` is not within the allow list.

As a workaround, I can fall back to storing the session not as JSON.

aboutZZ:
I had same issue, you can add your jackson mixins:
```java
@Bean
public OAuth2AuthorizationService oauth2AuthorizationService(JdbcTemplate jdbcTemplate, RegisteredClientRepository registeredClientRepository) {
    log.trace("[APP] Initializing bean OAuth2AuthorizationService");
    JdbcOAuth2AuthorizationService service = new JdbcOAuth2AuthorizationService(jdbcTemplate, registeredClientRepository);
[...]

PR #12584, opened 2023-01-25
Extend jackson allow list

TolstovM:
Fix for this issue https://github.com/spring-projects/spring-security/issues/12294
This exception occurs when session is stored in redis and `RedisIndexedSessionConfiguration` is used.
Also I faced with similar problem when put `Collection.emptySet()` in user's authorities, so I added it to allow list too.


marcusdacoregio:
Closing as declined, see https://github.com/spring-projects/spring-security/issues/12294#issuecomment-1568526815 for more details

Issue #12539, opened 2023-01-13
Saml2LogoutRequestMixin relayState property should be binding


ghaege:
**Describe the bug**
It seems relayState property should be binding

**To Reproduce**
```
class Saml2LogoutRequestMixin {
..
  @JsonProperty("relayState") Saml2MessageBinding relayState,
```
**Expected behavior**
```
class Saml2LogoutRequestMixin {
..
  @JsonProperty("binding") Saml2MessageBinding binding,
```

PR #12588, opened 2023-01-26
Saml2LogoutRequestMixin relayState property should be binding

marcusdacoregio:
Closes gh-12539


jzheaux:
@marcusdacoregio are you able to add a test that confirms that the behavior will continue to work this way?

Issue #12604, opened 2023-01-31
Add support for AuthnRequestsSigned setting 


1livv:
**Expected Behavior**

https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf#740 defines it.
It is useful when a service provider wants to sign the request no matter what.

**Current Behavior**
Right now the SAML request is signed only if the asserting party has set WantAuthnRequestsSigned.

**Context**
Trying to migrate from the old SAML library and we use AuthnRequestsSigned setting to always sign the request.

PR #12605, opened 2023-01-31
Add support for AuthnRequestsSigned setting

1livv:
closes gh-12604


1livv:
Would it be possible to have this merged in the 5.8.x alongside main even it's a new feature ? We're still using spring 5

jzheaux:
Sorry, @1livv, that I missed your earlier comment about 5.8. Generally, our policy is to only backport bugs to 5.8.

In the meantime, I believe you have the option to override the provider setting in `RelyingPartyRegistration`. Have you already tried that as a workaround?

1livv:
Thanks for the workaround @jzheaux that works ? Do I need to make my PR against main now ?

jzheaux:
Yes, @1livv, it would be best to rebase on `main`.

1livv:
@jzheaux  rebased it

Issue #12662, opened 2023-02-13
Initialize as complete a MethodSecurityExpressionHandler object as possible to ensure migration from @EnableGlobalMethodSecurity to @EnableMethodSecurity will work


orange-guo:
**Expected Behavior**

The method with the name `defaultExpressionHandler` of class `PrePostMethodSecurityConfiguration` should initialize the following dependencies of `MethodSecurityExpressionHandler`
`PermissionEvaluator`, `RoleHierarchy`, `AuthenticationTrustResolver`, `GrantedAuthorityDefaults`

**Current Behavior**

The method with the name `defaultExpressionHandler` of class `PrePostMethodSecurityConfiguration` initialized only `GrantedAuthorityDefaults`

**Context**
My customized `RoleHierarchy` is not working When I migrate the `@EnableGlobalMethodSecurity` to `@EnableMethodSecurity`.
After looking at the source code of the implementation class `GlobalMethodSecurityConfiguration` associated with `@EnableGlobalMethodSecurity`, I found it will initialize the dependencies of `defaultMethodExpressionHandler` as much as possible.
But `@EnableMethodSecurity`'s associated implementation class `PrePostMethodSecurityConfiguration` is not like this.
I think the behavior should be as consistent as possible so there are no issues when migrating.

orange-guo:
PrePostMethodSecurityConfiguration.java
```java
private static MethodSecurityExpressionHandler defaultExpressionHandler(
			ObjectProvider<GrantedAuthorityDefaults> defaultsProvider, ApplicationContext context) {
		DefaultMethodSecurityExpressionHandler handler = new DefaultMethodSecurityExpressionHandler();
		defaultsProvider.ifAvailable((d) -> handler.setDefaultRolePrefix(d.getRolePrefix()));
		handler.setApplicationContext(context);
		return handler;
	}
```

GlobalMethodSecurityConfiguration.java
```java
@Override
	public void afterSingletonsInstantiated() {
		try {
			initializeMethodSecurityInterceptor();
		}
		catch (Exception ex) {
			throw new RuntimeException(ex);
		}
		PermissionEvaluator permissionEvaluator = getSingleBeanOrNull(PermissionEvaluator.class);
		if (permissionEvaluator != null) {
			this.defaultMethodExpressionHandler.setPermissionEvaluator(permissionEvaluator);
		}
		RoleHierarchy roleHierarchy = getSingleBeanOrNull(RoleHierarchy.class);
[...]

PR #12673, opened 2023-02-15
FIX gh-12662

orange-guo:
Initialize as complete the default MethodSecurityExpressionHandler object as possible to ensure migration from @EnableGlobalMethodSecurity to @EnableMethodSecurity will work
see https://github.com/spring-projects/spring-security/issues/12662


jzheaux:
Thanks very much for submitting a PR, @orange-guo. I'm afraid we won't be able to accept it due to https://github.com/spring-projects/spring-security/issues/12662#issuecomment-1432040864. If you'd like further consideration, please feel free to continue the conversation on the issue.

PR #12694, opened 2023-02-17
Fixed test in DefaultLoginPageGeneratingFilterTests

twosom:
In the existing `handlesNonIso8859CharsInErrorMessage` method of `DefaultLoginPageGeneratingFilterTests`, 
it only retrieved a `Korean ISO-8859` message but did not verify whether it was properly included in the HTML document that is actually returned as the response. 
Therefore, a verification logic was added.


jzheaux:
Thanks, @twosom! This is now merged into `5.8.x`, `6.0.x`, and `main`.

PR #12740, opened 2023-02-18
Add test to SimpleUrlAuthenticationSuccessHandlerTests

twosom:
In the `onAuthenticationSuccess` method of the `SimpleUrlAuthenticationSuccessHandler`, the `clearAuthenticationAttributes` method is used to remove `AuthenticationException` objects that may be left in the session.

However, since there was no verification of whether this logic was working properly in test cases of `SimpleUrlAuthenticationSuccessHandler`, I added tests to verify it.

### DIFF
--- a/web/src/test/java/org/springframework/security/web/authentication/SimpleUrlAuthenticationSuccessHandlerTests.java
+++ b/web/src/test/java/org/springframework/security/web/authentication/SimpleUrlAuthenticationSuccessHandlerTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


jzheaux:
Thanks for the PR, @twosom! So that we can confirm that this feature continues to work on all branches, will you please rebase this off of `5.7.x`? Then when we merge, I'll forward port this to `5.8.x`, `6.0.x`, and `main`.

twosom:
> Thanks for the PR, @twosom! So that we can confirm that this feature continues to work on all branches, will you please rebase this off of `5.7.x`? Then when we merge, I'll forward port this to `5.8.x`, `6.0.x`, and `main`.

Thanks @jzheaux! I rebased it to 5.7.x :-)

jzheaux:
Thanks, @twosom! This is now merged into `5.7.x`, `5.8.x`, `6.0.x`, and `main`.

Issue #12664, opened 2023-02-13
SAML migration questions: RelyingPartyRegistrations#fromMetadata() doesn't work with a federation catalog that has a mix of SP:s and IdP:s et.c


stnor:
I am trying to migrate from the old SAML extension project to the new. on Spring 5.8.x (not boot).

It would be good if i could use `RelyingPartyRegistrations.collectionFromMetadataLocation()` could skip "SP" entries instead of throwing exceptions.

Right now I am getting `org.springframework.security.saml2.Saml2Exception: Metadata response is missing the necessary IDPSSODescriptor element`

Ideally there should be a flag to skip entities without IDPSSODescriptor. In this federation, there are `SPSSODescriptor`:s mixed in the same metadata as the IdP:s in this case.

See `https://fed.skolfederation.se/prod/md/skolfederation-3_1.xml` (A Federation for school owners (IdP) ca 200+ and e-learning resources (SPs) in Sweden).

Since the classes are package private and final, it is hard to work around the issue at present.

The only possible workaround seems to be to copy classes.. 

Also, how does one parse and store the other metadata, that was read by the old implementation, such as "organisation.name" when `RelyingPartyRegistration` is final and there are no hooks in the code afaik. Couldn't it be an interface instead? Or expose the XMLObject?

I have a dropdown list to select the IdP by OrgName in my implementation today, that's using the old project.

I'm unable to find a migration guide, and the docs are pretty sparse.

Thanks.

jzheaux:
Hi, @stnor, thanks for the detailed report.

Looks like there are a couple of things here. 

- Address SPSSODescriptor error handling
- Add to SAML Metadata documentation for getting the underlying XMLObject (already supported, undocumented, though)

I've added https://github.com/spring-projects/spring-security/issues/12667 to address the second one. As for the first one, yes, I think that makes sense to add and am happy to use this ticket to address that.

> I'm unable to find a migration guide

It is in our Wiki: https://github.com/spring-projects/spring-security/wiki/SAML-2.0-Migration-Guide - note that the main page didn't have it in the bullet list, so I've added it there to make it easier to see.

stnor:
Great, thanks for the prompt response.

I am looking at the migration guide right now, and I am unable to tell if the URL:s are the same form that page. Granted, I can look at the code, but ideally the URL:s should be documented if they are changed.

I am using this configuration at the moment. I can't tell if they're considered "default" or not.
```
[...]

PR #12742, opened 2023-02-18
Ignore non-IDPSSODescriptor entries when reading metadata 

stnor:
Fixes https://github.com/spring-projects/spring-security/issues/12664

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/OpenSamlMetadataAssertingPartyDetailsConverter.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/OpenSamlMetadataAssertingPartyDetailsConverter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/OpenSamlMetadataAssertingPartyDetailsConverter.java
```diff
@@ -71,13 +71,16 @@ Collection<RelyingPartyRegistration.AssertingPartyDetails.Builder> convert(Input
...
+				if (descriptor.getIDPSSODescriptor(SAMLConstants.SAML20P_NS) != null) {
+					builders.add(convert(descriptor));
+				}
 			}
 			return builders;
 		}
 		if (xmlObject instanceof EntityDescriptor) {
 			EntityDescriptor descriptor = (EntityDescriptor) xmlObject;
-			return Arrays.asList(convert(descriptor));
+			return descriptor.getIDPSSODescriptor(SAMLConstants.SAML20P_NS) == null ?
```
jzheaux, line 82: I think this changes the semantics of the method, which we shouldn't do in a maintenance release. It will also break `RelyingPartyRegistrations#fromMetadata` since it relies on the return value being non-empty. I'd recommend this check be removed so that the original exception is still thrown.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/OpenSamlMetadataAssertingPartyDetailsConverter.java
```diff
@@ -71,13 +71,16 @@ Collection<RelyingPartyRegistration.AssertingPartyDetails.Builder> convert(Input
 		if (xmlObject instanceof EntitiesDescriptor) {
 			EntitiesDescriptor descriptors = (EntitiesDescriptor) xmlObject;
 			for (EntityDescriptor descriptor : descriptors.getEntityDescriptors()) {
-				builders.add(convert(descriptor));
+				if (descriptor.getIDPSSODescriptor(SAMLConstants.SAML20P_NS) != null) {
+					builders.add(convert(descriptor));
+				}
 			}
 			return builders;
```
jzheaux, line 78: To preserve semantics, I believe this should still error if `builders` is empty at this point.

jzheaux (changes requested):
Thanks, @stnor, this is helpful. I've left some feedback inline. Also, will you please add a unit test to confirm that the functionality works?

Finally, once you've rebased onto `5.7.x` remember to also squash your commit and include the issue that is being closed like so:

```
Ignore SPSSODescriptor elements

Closes gh-12664
```

Issue #12751, opened 2023-02-20
OrRequestMatcher and AndRequestMatcher should accept List<? extends RequestMatcher>


chschu:
**Expected Behavior**

By allowing construction of an `OrRequestMatcher`/`AndRequestMatcher` from a `List<? extends RequestMatcher>`, the following expressions would compile:
```
new OrRequestMatcher(Stream.of(new AntPathRequestMatcher("foo")).toList());
```

**Current Behavior**

The expression doesn't compile, because Java infers the type `List<AntPathRequestMatcher>` for the parameter value, and there is no constructor accepting that type.

**Context**

With complex generic expressions (which are common when using the Java Streams API), explicit type parameters must be used to use the two matchers:
```
new OrRequestMatcher(Stream.<RequestMatcher>of(new AntPathRequestMatcher("foo")).toList());
```
This is hard to write and even harder to read in more complex cases.

jzheaux:
Thanks for the PR, @chschu.

Based on https://wiki.eclipse.org/Evolving_Java-based_APIs_2#Evolving_API_classes_-_API_methods_and_constructors, I'm not certain that we can make this change and remain binary compatible.

That said, I wonder if a `RequestMatchers` class with `allOf` and `anyOf` would be nice, similar to what is done in `AuthorizationManagers` (less the generic parameter type). In that case, `allOf` and `anyOf` could take the list of matchers as you describe.

chschu:
Thanks for the review, @jzheaux.

I just compiled a class file against the less general version of `AndRequestMatcher`, and executed it with the more general version on the classpath. This worked, so binary compatibility doesn't seem to be an issue. According to `javap`, generic parameter information is indeed present in the class file, but only the type-erased constructor signature is used during constructor lookup.

Anyway, I like your idea of a `RequestMatchers` class with some static methods even better. If there are no other objections regarding that change, I'd happily adapt my pull request and leave `AndRequestMatcher` and `OrRequestMatcher` unchanged.

PR #12752, opened 2023-02-20
Add factory class for RequestMatcher composition

chschu:
This PR implements the improvement described in gh-12751.

The added test cases verify that compilation succeeds.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/util/matcher/RequestMatchers.java
+++ b/web/src/main/java/org/springframework/security/web/util/matcher/RequestMatchers.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2002-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.web.util.matcher;
+
+import java.util.List;
+
+/**
+ * A factory class to create {@link RequestMatcher} instances.
+ *
+ * @author Christian Schuster
+ * @since 6.1
+ */
+public final class RequestMatchers {
+
+	/**
+	 * Creates a {@link RequestMatcher} that matches if at least one of the given
+	 * {@link RequestMatcher}s matches, if <code>matchers</code> are empty then the
+	 * returned matcher never matches.
+	 * @param matchers the {@link RequestMatcher}s to use
+	 * @return the any-of composed {@link RequestMatcher}
+	 * @see OrRequestMatcher
+	 */
+	public static RequestMatcher anyOf(RequestMatcher... matchers) {
+		return (matchers.length > 0) ? new OrRequestMatcher(List.of(matchers)) : (request) -> false;
+	}
+
+	/**
+	 * Creates a {@link RequestMatcher} that matches if all the given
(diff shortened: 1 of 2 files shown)


jzheaux:
Based on https://wiki.eclipse.org/Evolving_Java-based_APIs_2#Evolving_API_classes_-_API_methods_and_constructors, I'm not certain that we can make this change and remain binary compatible.

That said, I wonder if a `RequestMatchers` class with `allOf` and `anyOf` would be nice, similar to what is done in `AuthorizationManagers` (less the generic parameter type). In that case, `allOf` and `anyOf` could take the list of matchers as you describe.

jzheaux:
Thanks, @chschu! Please see my comments [in the ticket](https://github.com/spring-projects/spring-security/issues/12751) before we proceed further.

web/src/main/java/org/springframework/security/web/util/matcher/RequestMatchers.java
```diff
@@ -0,0 +1,56 @@
...
+public final class RequestMatchers {
+
+	/**
+	 * Creates a {@link RequestMatcher} that matches if at least one of the
+	 * given {@link RequestMatcher}s matches, if <code>matchers</code>
+	 * are empty then the returned matcher never matches.
+	 * @param matchers the {@link RequestMatcher}s to use
+	 * @return the any-of composed {@link RequestMatcher}
+	 */
+	public static RequestMatcher anyOf(RequestMatcher... matchers) {
```
jzheaux, line 18: Will you please use `AndRequestMatcher` here? The reason is so that the result also gets `AndRequestMatcher#matcher`.

web/src/main/java/org/springframework/security/web/util/matcher/RequestMatchers.java
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	/**
+	 * Creates a {@link RequestMatcher} that matches if all the
+	 * given {@link RequestMatcher}s match, if <code>matchers</code>
+	 * are empty then the returned matcher always matches.
+	 * @param matchers the {@link RequestMatcher}s to use
+	 * @return the all-of composed {@link RequestMatcher}
+	 */
+	public static RequestMatcher allOf(RequestMatcher... matchers) {
```
jzheaux, line 36: Will you please use `OrRequestMatcher` here? The reason is so that the result also gets `OrRequestMatcher#matcher`.

web/src/main/java/org/springframework/security/web/util/matcher/RequestMatchers.java
```diff
@@ -0,0 +1,56 @@
...
+	/**
+	 * Creates a {@link RequestMatcher} that matches if the
+	 * given {@link RequestMatcher} does not match.
+	 * @param matcher the {@link RequestMatcher} to use
+	 * @return the inverted {@link RequestMatcher}
+	 */
+	public static RequestMatcher not(RequestMatcher matcher) {
+		return request -> !matcher.matches(request);
+	}
+}
```
jzheaux, line 56: The project code formatter is going to require a private constructor and possibly a couple of other things. WIll you please run `./gradlew format check` to get the complete list?

jzheaux (changes requested):
Thanks, @chschu! I've left some feedback inline.

web/src/main/java/org/springframework/security/web/util/matcher/RequestMatchers.java
chschu: Done.

I had to handle the case with zero matchers explicitly, because the constructor of `OrRequestMatcher` throws an exception in that case.

web/src/main/java/org/springframework/security/web/util/matcher/RequestMatchers.java
chschu: Done.

I had to handle the case with zero matchers explicitly, because the constructor of `AndRequestMatcher` throws an exception in that case.

jzheaux (approved):

jzheaux:
Thanks, @chschu! This is now merged into `main`.

PR #12753, opened 2023-02-21
Add ability to customize AuthenticationManager and AuthorizationManager observation conventions

### LINKED ISSUE #12534: Ability to customize the `AuthorizationObservationConvention`
**Expected Behavior**

The ability to customize the obersations made by spring security (adjust low/high cardinality keys, etc). For instance I would like to add the high cardinality key of the `Principal#getName()` to my traces.

**Current Behavior**

You cannot currently customize the provided `AuthorizationObservationConvention` without registering your own `ObersationRegistry` which replaces the entire registry.

**Context**

Here is where the `AuthorizationObservationConvention` is currently used and instantiated as part of the class: https://github.com/spring-projects/spring-security/blob/9d876fce820e8d517bd0c947a3113d46082070c5/core/src/main/java/org/springframework/security/authorization/ObservationAuthorizationManager.java#L39

In spring boot, they made adjustments ahead of the 3.0 release to make their own observation conventions customizable for HTTP clients. It would make sense for spring-security to do the same thing here and allow providing your own conventions/overridden conventions.

See spring boots implementation here: https://github.com/spring-projects/spring-boot/commit/07766c436cb89128eac5389d33f76329b758d8df

braunsonm:
Fixes #12534


braunsonm:
cc @jzheaux

core/src/main/java/org/springframework/security/authentication/ObservationAuthenticationManager.java
```diff
@@ -56,4 +57,11 @@ public Authentication authenticate(Authentication authentication) throws Authent
 		});
 	}
 
+	public ObservationConvention<AuthenticationObservationContext> getConvention() {
```
jzheaux, line 60: For now, please leave out the getters. Doing so has a tendency to simplify backward compatibility over time, which is nice.

core/src/main/java/org/springframework/security/authentication/ObservationAuthenticationManager.java
```diff
@@ -56,4 +57,11 @@ public Authentication authenticate(Authentication authentication) throws Authent
...
+		return this.convention;
+	}
+
+	public void setConvention(ObservationConvention<AuthenticationObservationContext> convention) {
```
jzheaux, line 64: Will you please add a quick JavaDoc here that at least marks the version when it was added? For example:

```java
/**
 * Use this convention for reporting observation data
 * 
 * @since 6.1
 */
```

jzheaux: Please also do this for the other setters.

core/src/main/java/org/springframework/security/authentication/ObservationAuthenticationManager.java
```diff
@@ -56,4 +57,11 @@ public Authentication authenticate(Authentication authentication) throws Authent
...
+		this.convention = convention;
```
jzheaux, line 65: Please check for `null` here. A simple way to do that is:

```java
Assert.notNull(convention, "The observation convention cannot be null");
```

jzheaux: Please also do this for the other setters.

core/src/main/java/org/springframework/security/authentication/ObservationAuthenticationManager.java
jzheaux: Please also remove the other getters.

jzheaux (changes requested):
Thanks, @braunsonm! Sorry it took me a bit to get to reviewing your PR; I've left some feedback inline.

In addition to that inline feedback, will you please add unit tests to confirm that your new functionality works? Please, for example, add a test to `ObservationAuthenticationManagerTests` and one to each of the other corresponding test classes.

When you are ready to go, please squash all of your commits into one with a commit message that looks something like this:

```
Support Customizing Observation Conventions

Closes gh-12534
```

This format integrates with GitHub to close the ticket automatically, which is helpful.

braunsonm:
Should be good. There isn't much to test on a setter but I assumed you wanted the assertion tested

jzheaux (approved):

jzheaux:
Thanks, @braunsonm! This is now merged into `main`. I also merged a couple of polish commits, one for general formatting and one to adjust the method name to `setObservationConvention` for clarity.

PR #12760, opened 2023-02-21
AuthorizeHttpRequestsConfigurer#XXXrole methods should ensure the role does not begin with the role prefix

### LINKED ISSUE #12581: AuthorizeHttpRequestsConfigurer#XXXrole methods should ensure the role does not begin with the role prefix
Similar to how `User.UserBuilder` works, configuration should fail if configurations like the following are attempted:

```java
hasRole('ROLE_USER')
hasAnyRole('ROLE_USER')
```

Otherwise, scenarios like [this SO question](https://stackoverflow.com/questions/75225700/hasrolerole-user-not-working-in-spring-boot3-with-springsecurity6/75227337#75227337) crop up.

bist220:
There are no check for role prefix in AuthorizeHttpRequestsConfigurer#XXXrole methoids. This PR adds check for the same. Now the configuration will fail if role/s start with prefix for hasRole and hasAnyRole methods.

Closes #12581

### DIFF
--- a/core/src/main/java/org/springframework/security/authorization/AuthorityAuthorizationManager.java
+++ b/core/src/main/java/org/springframework/security/authorization/AuthorityAuthorizationManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


jzheaux:
Thanks, @bist220 for the PR! This is now merged into `5.7.x`, `5.8.x`, `6.0.x`, and `main`.

PR #12779, opened 2023-02-24
Update doc to use SecurityContextHolderFilter

skryvets:
Replacing deprecated `SecurityContextPersistenceFilter` with `SecurityContextHolderFilter`.


marcusdacoregio:
Hi @skryvets, I don't think I follow this change because the `SecurityContextPersistenceFilter` is the one the has the `forceEagerSessionCreation` property

skryvets:
Sorry, yes, you're right I just noticed it. 

I was trying to understand how it works. The latest spring security doesn't have `SecurityContextPersistenceFilter` and uses `SecurityContextHolderFilter`, that's why I thought it should be changed. I completely missed the fact that the latter doesn't have forceEagerSessionCreation.

I see now that "holder" doesn't manipulate session and 100% relies on `SecurityContextRepository`/`SecurityContextHolderStrategy`.

Please disregard

Issue #12795, opened 2023-02-27
RequestedUrlRedirectInvalidSessionStrategy doesn't take servlet context path into account


mches:
**Describe the bug**
When using `RequestedUrlRedirectInvalidSessionStrategy` as the invalid session strategy within the session management of the security filter chain, with a non-empty servlet context path configured, the context path is duplicated in the redirect URLs. For example, if the context path is "/app" and the requested URI is "/app/requested", then the redirect URL becomes "/app/app/requested".

**To Reproduce**
Configure a non-empty context path such as "/app". Set an instance of `RequestedUrlRedirectInvalidSessionStrategy` as the invalid session strategy in the security filter chain. Access the app at "http://localhost:8080/app/requested" with an invalid session ID. Get redirected to "/app/app/requested".

**Expected behavior**
The redirect should not duplicate the context path segments. It should be the same as the request URI, "/app/requested".

**Sample**

https://github.com/spring-projects/spring-security/compare/5.7.x...mches:spring-security:gh-12795

PR #12796, opened 2023-02-27
Take servlet context path into account and ensure HTTP method is preserved on redirect to requested URL

mches:
* Makes `RequestedUrlRedirectInvalidSessionStrategy` take servlet context path into account in putting together the redirect URL.
* Makes the 3xx status code used by `DefaultRedirectStrategy` configurable.
* Makes `RequestedUrlRedirectInvalidSessionStrategy` use status `307 Temporary Redirect` to prevent user agents from changing the HTTP method.

I'd like to see this backported to 5.8.x at least.

Closes gh-12795.
Closes gh-12797.


mches:
Closed in favor of gh-12817

Issue #12797, opened 2023-02-28
RequestedUrlRedirectInvalidSessionStrategy can cause the HTTP method to change depending on the user agent


mches:
**Describe the bug**
`RequestedUrlRedirectInvalidSessionStrategy#onInvalidSessionDetected(…)` uses `DefaultRedirectStrategy#sendRedirect(…)`, which delegates to `HttpServletResponse#sendRedirect(…)` which always sets HTTP status `302 Found` regardless of the HTTP method of the original request. I've observed this can cause the HTTP method to change from POST to GET depending on the user agent, as described in [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302). The same documentation recommends  to redirect with HTTP status `307 Temporary Redirect` for HTTP methods other than GET and HEAD to prevent the user agent from changing the method.

> Even if the specification requires the method (and the body) not to be altered when the redirection is performed, not all user-agents conform here - you can still find this type of bugged software out there. It is therefore recommended to set the 302 code only as a response for [GET](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET) or [HEAD](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD) methods and to use [307 Temporary Redirect](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307) instead, as the method change is explicitly prohibited in that case.

**To Reproduce**
Set an instance of `RequestedUrlRedirectInvalidSessionStrategy` as the invalid session strategy in the security filter chain. Make a POST request with an invalid session ID with Chrome browser. Observe a method not supported error due to Chrome using the GET method where only POST is allowed.

**Expected behavior**
The POST request with an invalid session ID should be redirected with HTTP status `307 Temporary Redirect`. Chrome will be prohibited from changing the HTTP method.

**Sample**

https://github.com/spring-projects/spring-security/compare/5.7.x...mches:spring-security:gh-12797

PR #12817, opened 2023-03-02
Make it possible to take servlet context path into account and ensure HTTP method is preserved on redirect to requested URL

mches:
* Makes it possible to customize the redirect strategy of `RequestedUrlRedirectInvalidSessionStrategy`, so that the servlet context path can be taken into account if desired.
* Makes the HTTP status code used by `DefaultRedirectStrategy` configurable, so that `307 Temporary Redirect` can be chosen if desired.

Closes gh-12795
Closes gh-12797

### DIFF
--- a/web/src/main/java/org/springframework/security/web/DefaultRedirectStrategy.java
+++ b/web/src/main/java/org/springframework/security/web/DefaultRedirectStrategy.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,6 +24,8 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 import org.springframework.core.log.LogMessage;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpStatus;
 import org.springframework.security.web.util.UrlUtils;
 import org.springframework.util.Assert;
 
@@ -32,6 +34,7 @@ import org.springframework.util.Assert;
  * the framework.
  *
  * @author Luke Taylor
+ * @author Mark Chesney
  * @since 3.0
  */
 public class DefaultRedirectStrategy implements RedirectStrategy {
@@ -39,6 +42,8 @@ public class DefaultRedirectStrategy implements RedirectStrategy {
 	protected final Log logger = LogFactory.getLog(getClass());
 
 	private boolean contextRelative;
+
+	private HttpStatus statusCode = HttpStatus.FOUND;
 
 	/**
 	 * Redirects the response to the supplied URL.
@@ -55,7 +60,14 @@ public class DefaultRedirectStrategy implements RedirectStrategy {
 		if (this.logger.isDebugEnabled()) {
 			this.logger.debug(LogMessage.format("Redirecting to %s", redirectUrl));
 		}
-		response.sendRedirect(redirectUrl);
+		if (this.statusCode == HttpStatus.FOUND) {
+			response.sendRedirect(redirectUrl);
+		}
+		else {
+			response.setHeader(HttpHeaders.LOCATION, redirectUrl);
+			response.setStatus(this.statusCode.value());
+			response.getWriter().flush();
+		}
 	}
 
 	protected String calculateRedirectUrl(String contextPath, String url) {
@@ -94,6 +106,20 @@ public class DefaultRedirectStrategy implements RedirectStrategy {
 	 */
 	protected boolean isContextRelative() {
 		return this.contextRelative;
+	}
+
+	/**
+	 * Sets the HTTP status code to use. The default is {@link HttpStatus#FOUND}.
+	 * <p>
+	 * Note that according to RFC 7231, with {@link HttpStatus#FOUND}, a user agent MAY
+	 * change the request method from POST to GET for the subsequent request. If this
+	 * behavior is undesired, {@link HttpStatus#TEMPORARY_REDIRECT} can be used instead.
+	 * @param statusCode the HTTP status code to use.
+	 * @since 6.2
+	 */
+	public void setStatusCode(HttpStatus statusCode) {
+		Assert.notNull(statusCode, "statusCode cannot be null");
+		this.statusCode = statusCode;
 	}
 
 }
--- a/web/src/test/java/org/springframework/security/web/DefaultRedirectStrategyTests.java
+++ b/web/src/test/java/org/springframework/security/web/DefaultRedirectStrategyTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,6 +18,7 @@ package org.springframework.security.web;
 
 import org.junit.jupiter.api.Test;
 
+import org.springframework.http.HttpStatus;
 import org.springframework.mock.web.MockHttpServletRequest;
 import org.springframework.mock.web.MockHttpServletResponse;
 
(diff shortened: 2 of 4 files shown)


web/src/main/java/org/springframework/security/web/DefaultRedirectStrategy.java
```diff
@@ -56,7 +62,13 @@ public void sendRedirect(HttpServletRequest request, HttpServletResponse respons
 		if (this.logger.isDebugEnabled()) {
 			this.logger.debug(LogMessage.format("Redirecting to %s", redirectUrl));
 		}
-		response.sendRedirect(redirectUrl);
+		if (this.statusCode == null) {
```
jzheaux, line 65: Please change this to check if `this.statusCode == HttpStatus.TEMPORARY_REDIRECT`. This reason is so that if `setStatusCode(302)` is called, the default behavior is restored.

web/src/main/java/org/springframework/security/web/DefaultRedirectStrategy.java
```diff
@@ -56,7 +62,13 @@ public void sendRedirect(HttpServletRequest request, HttpServletResponse respons
...
+			response.sendRedirect(redirectUrl);
+		}
+		else {
+			response.setHeader(HttpHeaders.LOCATION, redirectUrl);
+			response.setStatus(this.statusCode.value());
```
jzheaux, line 70: To ensure that the commit expectations are similar, please add a call to `response.getWriter().flush()`.

web/src/main/java/org/springframework/security/web/DefaultRedirectStrategy.java
```diff
@@ -97,4 +109,12 @@ protected boolean isContextRelative() {
 		return this.contextRelative;
 	}
 
+	public void setStatusCode(HttpStatus statusCode) {
+		this.statusCode = statusCode;
```
jzheaux, line 113: Please disallow `null` values.

web/src/main/java/org/springframework/security/web/DefaultRedirectStrategy.java
```diff
@@ -97,4 +109,12 @@ protected boolean isContextRelative() {
 		return this.contextRelative;
 	}
 
+	public void setStatusCode(HttpStatus statusCode) {
```
jzheaux, line 112: Please add JavaDoc describing how to use the method and when it comes in handy. It might be helpful to indicate that this method should be used to disambiguate 302 behavior (303 to rewrite the method, 307 to preserve it).

web/src/main/java/org/springframework/security/web/DefaultRedirectStrategy.java
```diff
@@ -97,4 +109,12 @@ protected boolean isContextRelative() {
...
+		this.statusCode = statusCode;
+	}
+
+	protected HttpStatus getStatusCode() {
```
jzheaux, line 116: Spring Security tends to avoid getters; will you please remove this one for the time being?

web/src/main/java/org/springframework/security/web/DefaultRedirectStrategy.java
```diff
@@ -41,6 +44,9 @@ public class DefaultRedirectStrategy implements RedirectStrategy {
 
 	private boolean contextRelative;
 
+	@Nullable
+	private HttpStatus statusCode;
```
jzheaux, line 48: Instead of setting this to `null`, please set it to 302.

jzheaux (changes requested):
Thanks for the PR, @mches! And sorry for the delay; I think this PR got lost in the shuffle.

I've left some inline feedback and look forward to your updates.

jzheaux:
Also, since this is a new feature, @mches, will you please change the branch to `main` and ensure that any new APIs have a `@since 6.2` JavaDoc?

web/src/main/java/org/springframework/security/web/DefaultRedirectStrategy.java
mches: Do you actually mean `this.statusCode == HttpStatus.FOUND` and `setStatusCode(HttpStatus.FOUND)` respectively?

web/src/main/java/org/springframework/security/web/DefaultRedirectStrategy.java
mches: Got it, but actually, `HttpStatus.FOUND`, right?

web/src/test/java/org/springframework/security/web/DefaultRedirectStrategyTests.java
```diff
@@ -64,4 +71,19 @@ public void contextRelativeShouldThrowExceptionIfURLDoesNotContainContextPath()
 				.isThrownBy(() -> rds.sendRedirect(request, response, "https://redirectme.somewhere.else"));
 	}
 
+	@ParameterizedTest
+	@EnumSource(value = HttpStatus.class, names = { "FOUND", "SEE_OTHER", "TEMPORARY_REDIRECT" })
```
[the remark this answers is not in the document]
mches: > It would also be valuable to assert that the response is not committed, so as to clearly indicate the difference between this approach and calling `sendRedirect`.

Referring to https://github.com/spring-projects/spring-security/pull/12817#discussion_r1307930631, it seems like you ultimately decided you'd prefer the response to be committed in any case. Is that right?

mches:
> Also, since this is a new feature, @mches, will you please change the branch to `main` and ensure that any new APIs have a `@since 6.2` JavaDoc?

I was hoping for this to be usable with Spring Boot 2.7.x, but I get it.

mches:
> Thanks for the PR, @mches! And sorry for the delay; I think this PR got lost in the shuffle.
> 
> I've left some inline feedback and look forward to your updates.

Thank you for the review, @jzheaux. I've responded inline. 

But also,

1. If the target is Spring Security 6.2 which has a Java 17 prerequisite, is it acceptable to use `var` in new tests?
2. Is it acceptable to include BDD-style comments on new tests? For example, `// given`, `// when`, `// then`
3. Is it acceptable to introduce builder-style setters to `DefaultRedirectStrategy` and `RequestedUrlRedirectInvalidSessionStrategy`, for added convenience?

Compare with builder-style setters:

```java
@Configuration
@EnableWebSecurity
public class Config {
    @Bean
    public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
        http
                .sessionManagement((sessionManagement) -> sessionManagement
                        .invalidSessionStrategy(new RequestedUrlRedirectInvalidSessionStrategy()
                                .setRedirectStrategy(new DefaultRedirectStrategy()
                                        .setContextRelative(true)
                                        .setStatusCode(HttpStatus.TEMPORARY_REDIRECT))))
                ...;
        return http.build();
    }
}
```

versus without:

```java
@Configuration
@EnableWebSecurity
public class Config {
    @Bean
    public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
        var redirectStrategy = new DefaultRedirectStrategy();
        redirectStrategy.setContextRelative(true);
        redirectStrategy.setStatusCode(HttpStatus.TEMPORARY_REDIRECT);
        var invalidSessionStrategy = new RequestedUrlRedirectInvalidSessionStrategy();
        invalidSessionStrategy.setRedirectStrategy(redirectStrategy);
        http
                .sessionManagement((sessionManagement) -> sessionManagement
                        .invalidSessionStrategy(invalidSessionStrategy))
                ...;
        return http.build();
    }
}
```

web/src/test/java/org/springframework/security/web/session/SessionManagementFilterTests.java
```diff
@@ -46,9 +49,11 @@
 /**
  * @author Luke Taylor
  * @author Rob Winch
+ * @author Mark Chesney
  */
 public class SessionManagementFilterTests {
 
+	@BeforeEach
```
mches, line 56: `SecurityContextHolderAwareRequestWrapperTests` and a number of other tests don't clear the context after each/all tests, so it becomes necessary to counteract side effects that crop up when such tests are run together.

```java
…
package org.springframework.security.web.servletapi;
…
public class SecurityContextHolderAwareRequestWrapperTests {

	@BeforeEach
	public void tearDown() {
		SecurityContextHolder.clearContext();
	}
…
}
```

jzheaux (approved):

PR #12818, opened 2023-03-02
Added support to display child status codes if they exist in SAML response

### LINKED ISSUE #11725: OpenSaml4AuthenticationProvider should include secondary statusCode messages on error
**Expected Behavior**

Two cases:
1. When the `isPassive` flag is set to `true`, and the request is sent to an IdP that doesn't support passive mode, the expected statusCode is `urn:oasis:names:tc:SAML:2.0:status:NoPassive`:

2. Similarly, with the `isPassive` flag, is set to `true`, and the request is sent to an IdP that supports passive mode, but the user doesn't have a session yet with the IdP, the expected statusCode is `urn:oasis:names:tc:SAML:2.0:status:NoPassive`

**Current Behavior**

In the `createDefaultResponseValidator` method, when the request goes through case 1, the samlResponse looks like this:
Screen Shot 2022-08-18 at 2 13 54 PM

However, the `createDefaultResponseValidator` uses the outer statusCode `urn:oasis:names:tc:SAML:2.0:status:Responder`:
Screen Shot 2022-08-18 at 2 07 23 PM

Similarly, in case 2, the samlResponse looks like this:
Screen Shot 2022-08-18 at 2 14 33 PM

and output from `createDefaultResponseValidator` is `urn:oasis:names:tc:SAML:2.0:status:Requester`:
Screen Shot 2022-08-18 at 3 06 09 PM

**Context**
#### How has this issue affected you? 
I can't tell if the source of the error is `NoPassive` or something else to decide how to proceed with the sign-in flow. 

#### What are you trying to accomplish?
I'm implementing a dynamic passive value for multi-tenants, and when the IdP doesn't support passive, or there's no session at the IdP, I'm detecting the error and using the `redirectStrategy` to send the user back to the main page. It's a public page that tries to passively log the user in if there's a session with the IdP.

#### What other alternatives have you considered?
None.

#### Are you aware of any workarounds?
No.

Anubhav-2000:
Fixes #11725 
Please let me know if i have implemented something incorrectly. First time submitting a PR for bug fix :)


Anubhav-2000:
Hi @jzheaux, could you take a look at this fix?

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/OpenSaml4AuthenticationProvider.java
```diff
@@ -371,13 +371,14 @@ public static Converter<ResponseToken, Saml2ResponseValidatorResult> createDefau
 			Response response = responseToken.getResponse();
 			Saml2AuthenticationToken token = responseToken.getToken();
 			Saml2ResponseValidatorResult result = Saml2ResponseValidatorResult.success();
-			String statusCode = getStatusCode(response);
-			if (!StatusCode.SUCCESS.equals(statusCode)) {
-				String message = String.format("Invalid status [%s] for SAML response [%s]", statusCode,
-						response.getID());
-				result = result.concat(new Saml2Error(Saml2ErrorCodes.INVALID_RESPONSE, message));
+			List<String> statusCodes = getStatusCodes(response);
```
jzheaux, line 374: Instead of getting all the status codes ad infinitum, please just get the top-level one. 

If it is of type `REQUESTER`, `RESPONDER`, or `VERSION_MISMATCH`, then add a single error message that includes this status code and any single child status code.

Otherwise, if it isn't `SUCCESS`, then add a single error message like it already does.

This is based on the logic outlined in: https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf#page=39

jzheaux (changes requested):
Thanks, @Anubhav-2000! Sorry for the delay in responding. I've left some inline feedback.

Also, will you please add a test that confirms the error message includes a child status code?

jzheaux:
Hi, @Anubhav-2000! Are you able to make the requested changes?

youngkih:
Hi @jzheaux ! May I follow up on this according to your feedback ?

youngkih:
Filed https://github.com/spring-projects/spring-security/pull/14743, wonder if you guys could help take a look

jzheaux:
Thanks, @Anubhav-2000 for getting things started! @youngkih was able to finish in a separate PR. I'm closing this in favor of https://github.com/spring-projects/spring-security/pull/14743

Issue #12873, opened 2023-03-16
DaoAuthenticationProvider is not usable on RHEL 8.7 with enforced FIPS mode


psvo:
**Describe the bug**

Creating instance of DaoAuthenticationProvider fails due to "PBKDF2WithHmacSHA256 SecretKeyFactory not available" when running on RHEL 8.7 with enforced FIPS mode.

The problem is that the DaoAuthenticationProvider creates a default delegating password encoder and one of the delegates fails to instantiate due to limited JCE provider availability when FIPS is enforced.

There's no workaround, because the DaoAuthenticationProvider has only the default constructor which fails due to unconditionally calling `org.springframework.security.crypto.factory.PasswordEncoderFactories#createDelegatingPasswordEncoder`.

The error is:
```
org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.security.authentication.dao.DaoAuthenticationProvider]: Constructor threw exception; nested exception is java.lang.IllegalArgumentException: Invalid algorithm 'PBKDF2WithHmacSHA256'.
	... more
Caused by: java.security.NoSuchAlgorithmException: PBKDF2WithHmacSHA256 SecretKeyFactory not available
	at javax.crypto.SecretKeyFactory.<init>(SecretKeyFactory.java:122) ~[?:?]
	at javax.crypto.SecretKeyFactory.getInstance(SecretKeyFactory.java:168) ~[?:?]
	at org.springframework.security.crypto.password.Pbkdf2PasswordEncoder.setAlgorithm(Pbkdf2PasswordEncoder.java:226) ~[spring-security-crypto-5.8.2.jar:5.8.2]
	at org.springframework.security.crypto.password.Pbkdf2PasswordEncoder.<init>(Pbkdf2PasswordEncoder.java:179) ~[spring-security-crypto-5.8.2.jar:5.8.2]
	at org.springframework.security.crypto.password.Pbkdf2PasswordEncoder.defaultsForSpringSecurity_v5_8(Pbkdf2PasswordEncoder.java:207) ~[spring-security-crypto-5.8.2.jar:5.8.2]
	at org.springframework.security.crypto.factory.PasswordEncoderFactories.createDelegatingPasswordEncoder(PasswordEncoderFactories.java:81) ~[spring-security-crypto-5.8.2.jar:5.8.2]
[...]

PR #12874, opened 2023-03-16
Add new DaoAuthenticationProvider constructor

psvo:
Add a new constructor to the DaoAuthenticationProvider, which allows providing a custom PasswordEncoder to prevent instantiation of the default delegating PasswordEncoder in the default constructor.

This provides a way to instantiate the DaoAuthenticationProvider on JDKs where the default delegating PasswordEncoder cannot be instantiated due to limited JCE providers for compliance reasons (e.g., FIPS).

Closes gh-12873

### DIFF
--- a/core/src/main/java/org/springframework/security/authentication/dao/DaoAuthenticationProvider.java
+++ b/core/src/main/java/org/springframework/security/authentication/dao/DaoAuthenticationProvider.java
@@ -61,7 +61,11 @@ public class DaoAuthenticationProvider extends AbstractUserDetailsAuthentication
 	private UserDetailsPasswordService userDetailsPasswordService;
 
 	public DaoAuthenticationProvider() {
-		setPasswordEncoder(PasswordEncoderFactories.createDelegatingPasswordEncoder());
+		this(PasswordEncoderFactories.createDelegatingPasswordEncoder());
+	}
+
+	public DaoAuthenticationProvider(PasswordEncoder passwordEncoder) {
+		setPasswordEncoder(passwordEncoder);
 	}
 
 	@Override


core/src/main/java/org/springframework/security/authentication/dao/DaoAuthenticationProvider.java
```diff
@@ -61,7 +61,11 @@ public class DaoAuthenticationProvider extends AbstractUserDetailsAuthentication
 	private UserDetailsPasswordService userDetailsPasswordService;
 
 	public DaoAuthenticationProvider() {
-		setPasswordEncoder(PasswordEncoderFactories.createDelegatingPasswordEncoder());
+		this(PasswordEncoderFactories.createDelegatingPasswordEncoder());
+	}
+
+	public DaoAuthenticationProvider(PasswordEncoder passwordEncoder) {
```
marcusdacoregio, line 67: Can you add a test where it makes sure that the provided `PasswordEncoder` is used?

core/src/main/java/org/springframework/security/authentication/dao/DaoAuthenticationProvider.java
```diff
@@ -61,7 +61,11 @@ public class DaoAuthenticationProvider extends AbstractUserDetailsAuthentication
 	private UserDetailsPasswordService userDetailsPasswordService;
 
 	public DaoAuthenticationProvider() {
-		setPasswordEncoder(PasswordEncoderFactories.createDelegatingPasswordEncoder());
+		this(PasswordEncoderFactories.createDelegatingPasswordEncoder());
+	}
+
```
marcusdacoregio, line 66: I think that it would be great to add a little bit of javadoc here, with a little description and the `@since 5.8.3` tag

marcusdacoregio (changes requested):
Hi @psvo, thank you for the PR.

This problem seems to be introduced in `5.8.x`, therefore I think that the target branch should be `5.8.x`. Additionally, the `5.6.x` branch is not supported anymore.

marcusdacoregio:
@psvo I was discussing this with a team member and we don't think adding a new constructor to a patch release will improve things. Since that change broke existing applications, we should aim to keep the behavior as before (do not throw an exception for the new encoders). A new constructor could be added for 6.1, the next minor release, and we should provide a bug fix for 5.8.x and 6.0.x.

That said, please hold the changes until we are 100% sure what to do here.

psvo:
> That said, please hold the changes until we are 100% sure what to do here.

Ok

PR #12930, opened 2023-03-27
Make eclipse/vscode project import work

jvalkeal:
- Gradle projects contain cycles which comes from dependencies to test sources which is not a problem in gradle but eclipse metadata generation is getting confused. Thus we need settings to relax errors org.eclipse.jdt.core.circularClasspath=warning org.eclipse.jdt.core.incompleteClasspath=warning
- Additionally .classpath entries needs to be changes having without_test_code=false test=false
- Aspects end up getting source dirs `build/classes/java/main` and `build/resources/main` which never have sources. Vscode complains about that, eclipse is fine. Remove those from entries.
- In tests `htmlunit` depends on `xml-apis`. `xml-apis` are now part of jdk and eclipse complains about that. Excluse these in a gradle build.
- Both eclipse and vscode don't currently work with buildship, due to project cycles and buildship cannot be configured. It's possible to create metadata from `eclipse` task manually which then can be imported. For this we need to disable automatic import in vscode using buildship. This goes to `.vscode/settings.json` workspace config.
- Then with these changes user can do something like git clean -fxd && ./gradlew clean build cleanEclipse eclipse -x checkstyleNohttp -x test -x integrationTest and import projects manually.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+	"java.import.gradle.enabled": false
+}
--- a/buildSrc/src/main/groovy/io/spring/gradle/convention/AbstractSpringJavaPlugin.groovy
+++ b/buildSrc/src/main/groovy/io/spring/gradle/convention/AbstractSpringJavaPlugin.groovy
@@ -54,6 +54,7 @@ public abstract class AbstractSpringJavaPlugin implements Plugin<Project> {
 		pluginManager.apply("io.spring.convention.javadoc-options");
 		pluginManager.apply("io.spring.convention.checkstyle");
 		pluginManager.apply(CopyPropertiesPlugin);
+		pluginManager.apply("io.spring.convention.eclipse");
 
 		project.jar {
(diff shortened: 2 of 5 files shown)


jvalkeal:
Eclipse imports cleanly
spring-security-eclipse-import-1

jvalkeal:
STS still has old validations which cannot be turned off https://github.com/spring-projects/sts4/issues/1016

Something to fix later as we have some real validation errors in some xml files.

spring-security-sts-import-1

jvalkeal:
vscode imports cleanly

spring-security-vscode-import-1

marcusdacoregio:
Hi @jvalkeal, I remember @rwinch comment about applying those changes to `5.7.x` and merging them all the way up to `main`. Have you considered that?

jvalkeal:
I rebased and force pushed this PR branch atop of `6.0.x`. That is an oldest branch we have any change of using eclipse/vscode. `5.x` creates multiple jars from a saml project using gradle features and that would not work as eclipse only understands single classpath per project.

Issue #12947, opened 2023-03-30
Remove OpenSaml deprecation warnings 


Brummolix:
**Expected Behavior**

When running Spring Security with saml support it should not log deprecation warnings.

**Current Behavior**

While using OpenSamlLogoutRequestResolver the application logs for example:

`09:09:04 [WARN ] DEPRECATED#warn:125 - Java class method 'setSessionIndex', (interface org.opensaml.saml.saml2.core.SessionIndex): This will be removed in the next major version of this software; replacement is setValue
`

**Context**

Deprecated (for removal) methods should not be used any longer.
I will provide a PR with a fix.

PR #12948, opened 2023-03-30
Replace deprecated OpenSaml methods

Brummolix:
Fixes gh-12947


Brummolix:

PR #12949, opened 2023-03-30
TestingAuthenticationToken takes broader collection type

lmartelli:
So that callers do not have to cast.


jzheaux:
Closing in favor of https://github.com/spring-projects/spring-security/pull/12953

Issue #12346, opened 2022-12-06
Spring Security SAML signature validation issue 


lukaszmigdalek:
**Describe the bug**
After upgrading the Spring Security to version 5.8.0 signature validation of the SAML Logout Response won't work. 
According to the [SAML specification](https://docs.oasis-open.org/security/saml/v2.0/saml-bindings-2.0-os.pdf) order of query params for signature should be:
`To construct the signature, a string consisting of the concatenation of the RelayState (if present),
SigAlg, and SAMLRequest (or SAMLResponse) query string parameters (each one URLencoded) is constructed in one of the following ways (ordered as below):
SAMLRequest=value&RelayState=value&SigAlg=value
SAMLResponse=value&RelayState=value&SigAlg=value`

But in the Spring Security 5.8.0/6.0.0, content is just query parameters without the `Signature` param and without care about mentioned above order of params.

The issue was caused by [this](https://github.com/spring-projects/spring-security/commit/5cbc1a47da4dea6c3bba02236a1f149959ef9ff0#diff-a622c52bae8f625138d6cced9a4a822114dbc24bdcec1fc46511f9f97920e7c6L206) change.

**To Reproduce**
IdP redirects to the URL where the query param `RelayState` is not in the middle e.g.
`SAMLResponse=someContentHere&SigAlg=http://www.w3.org/2000/09/xmldsig#rsa-sha1&RelayState=a7d72218-c3e6-44b1-b3ba-e0dbca8ea8a9`

Example IdP: OneLogin

**Expected behavior**
Content for signature validation should be ordered as described in the SAML 2.0 specification.
Also, encoded query param values (SigAlg, RelayState, SAMLRequest, etc) should be unchanged as described [here](https://github.com/spring-projects/spring-security/issues/11235).

jzheaux:
Hi, @lukaszmigdalek. There are some details I'm not quite following just yet.

> IdP redirects to the URL where the query param RelayState is not in the middle

This sounds like a problem with the IdP. Are you saying that Spring Security should fail if the parameters from the IdP are in the wrong order? (Perhaps you are right, I just first want to make sure I understand the problem you are running into.)

> But in the Spring Security 5.8.0/6.0.0, content is just query parameters without the Signature param and without care about mentioned above order of params

I'm not able to reproduce this. When I run `OpenSaml4LogoutRequestResolver`, the result of `Saml2LogoutRequest#getParametersQuery` is in the order you describe and does contain the `Signature` parameter.

[...]

PR #12963, opened 2023-04-03
SAML 2.0 HTTP Redirect Binding query params may appear in any order

lukaszmigdalek:
According to my latest comment:
https://github.com/spring-projects/spring-security/issues/12346#issuecomment-1375203031

SAML 2.0 HTTP Redirect Binding specification says that query params may appear in any order.
"Note that when verifying signatures, the order of the query string parameters on the resulting URL to be
 verified is not prescribed by this binding. The parameters may appear in any order. Before verifying a
 signature, if any, the relying party MUST ensure that the parameter values to be verified are ordered as
 required by the signing rules above."

Added support for a different order of query params based on the previous implementation:
https://github.com/spring-projects/spring-security/commit/88f9529329676c10d87ab6381ec44a96fd144d87

gh-12346 
https://github.com/spring-projects/spring-security/issues/12346

### DIFF
--- a/core/src/main/java/org/springframework/security/authentication/dao/DaoAuthenticationProvider.java
+++ b/core/src/main/java/org/springframework/security/authentication/dao/DaoAuthenticationProvider.java
@@ -40,8 +40,9 @@ import org.springframework.util.Assert;
 public class DaoAuthenticationProvider extends AbstractUserDetailsAuthenticationProvider {
 
 	/**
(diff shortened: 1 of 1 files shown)


jzheaux:
Thanks, @lukaszmigdalek, for the PR. This is now merged into `5.8.x`, `6.0.x`, and `main`. It will go out in `5.8.4`, `6.0.4`, and `6.1.1`.

PR #12976, opened 2023-04-07
Add missing null-check

lawley:
Following the behaviour of getHeader(String name), avoid call to validateAllowedHeaderValue with a null value (which results in an exception).


rwinch:
Thank you for the PR @lawley! 

Can you help me understand how you are coming across this problem? I do not think it is allowed for the `Enumeration` to contain a `null` value. If the header is undefined, then the `Enumeration` does not contain `null` values but is an empty one.

It does appear that the `Enumeration` itself can be `null` if the container does not allow access to the header information which means we should add a `null` check to the `Enumeration` itself. However, I do not understand how a `null` value in the `Enumeration` would occur.

marcusdacoregio:
Closing due to the lack of requested feedback.

PR #12996, opened 2023-04-11
Fix javadoc for migration from WebSecurityConfigurerAdapter

xak2000:
This PR just fixes the code sample in javadocs that errorneously uses `WebSecurity` parameter in a bean-creation method that doesn't need to have any parameters, as `WebSecurity` parameter is already provided in `WebSecurityCustomizer.customize(WebSecurity web)` functional method. Moreover, the sample doesn't even compile because of parameter names collision (the bean factory paramter and lambda parameter).

This fix could also be backported to 5.7.x branch.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configuration/WebSecurityConfigurerAdapter.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configuration/WebSecurityConfigurerAdapter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


jzheaux:
@xak2000, thanks for the PR. We've recently changed to use forward porting. To that end, will you please rebase this PR to `5.7.x` and then upon merging, I'll forward port to `5.8.x`, `6.0.x`, and `main`?

xak2000:
@jzheaux Rebased the PR to `5.7.x` and changed target branch to `5.7.x`. Should I also rebase on the fresh `5.7.x` version or it's good in the current state?

jzheaux:
Thanks for checking, @xak2000. No need to do a fresh rebase. This is now merged to `5.7.x` as well as `5.8.x`. In the end, I didn't need to forward-port to `6.0.x` or `main` as I had though, since `WebSecurityConfigurerAdapter` is removed in those versions.

PR #13007, opened 2023-04-13
Change access level of `OidcClientInitiatedLogoutSuccessHandler::determineTargetUrl` method to public

yuezk:
Currently, there is no easy way to customize or extend this handler, especially the `determineTargetUrl()`. The reason is that the class itself is `final`, and the modifier of `determineTargetUrl()` is `protect`. 

The scenario is: We want to add the `state` parameter to the final target URL, but we can't find a point to modify the target URL. We finally achieve this by passing a customized `RedirectStrategy` and modifying the target URL in its `sendRedirect()` method.

I think we can make `determineTargetUrl()` accessible from the outside, and we can define a class that extends `SimpleUrlLogoutSuccessHandler` that embeds a field of `OidcClientInitiatedLogoutSuccessHandler` and override the `determineTargetUrl()` method so that we can modify the target URL.


jzheaux:
Hi, @yuezk, and thanks for the PR.

I think instead the class should not be `final`. While this is not typical for most Spring Security APIs, this is the way that `AbstractAuthenticationTargetUrlRequestHandler` is designed. This is especially important since the top-level method returns `void`, making it trickier to use a delegate. Also, it brings this class into alignment with how `SimpleUrlLogoutSuccessHandler`, a similar implementation, is open for extension.

Would it address your use case if you could extend `OidcClientInitiatedLogoutSuccessHandler`?

yuezk:
@jzheaux Yes, makes sense. And it will address our case.

yuezk:
Hi @jzheaux, is there any problem with this PR, can we proceed? Thanks.

jzheaux:
Thanks for your patience, @yuezk. Your PR was waiting for the `6.1.1` release and for `6.2` development to officially start; sorry that I didn't explain that earlier.

This is now merged into `main` is scheduled for the `6.2.0-M1` release.

yuezk:
@jzheaux Thanks. Will it be backported to 5.8.x?

PR #13010, opened 2023-04-13
Deprecate .and() and non lambda DSL methods

marcusdacoregio:


docs/modules/ROOT/pages/migration-7/index.adoc
```diff
@@ -0,0 +1,8 @@
+[[preparing]]
+= Preparing for 7.0
+
+While Spring Security 7.0 is a bit far away, it is important to start preparing for it now.
```
rwinch, line 4: I'd change this to state that a date for Spring Security 7.0 has not been set

rwinch (changes requested):
Overall it looks good. Currently the deprecation notices only state that it is for removal. Please ensure all the deprecation notices include that it will be removed in 7.0. I provided one comment on the documentation as well. After these changes are made, please feel free to merge.

PR #13070, opened 2023-04-19
Suggestion: Remove duplicate of ContentTypeOptionsServerHttpHeadersWriter

joerg-richter-5234:
I stumbled upon this when i imported XContentTypeOptionsServerHttpHeadersWriter#X_CONTENT_OPTIONS
 and my test for the header 'X-Content-Type-Options' failed.

XContentTypeOptionsServerHttpHeadersWriter does not seem to in use.
The constant XContentTypeOptionsServerHttpHeadersWriter#X_CONTENT_OPTIONS is "X-Content-Options". 

ContentTypeOptionsServerHttpHeadersWriter is in use and covered by a test: HeaderSpecTests. 
The ContentTypeOptionsServerHttpHeadersWriter#X_CONTENT_OPTIONS is "X-Content-Type-Options". 

X-Content-Options does seem to be listed on https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers. 

If this is correct, i would like to suggest to remove or deprecate XContentTypeOptionsServerHttpHeadersWriter 
as a duplicate of ContentTypeOptionsServerHttpHeadersWriter.

grafik


marcusdacoregio:
Hi, @joerg-richter-5234, thanks for the finding.

Since this is a public API we cannot just delete it in a minor or patch version. I think the best approach here would be to deprecate that class in `main` and remove it in 7.0, what do you think? If that makes sense, you should target the `main` branch and add the following to the class:

```java
/**
 * ...
 * @deprecated For removal in 7.0. Use {@link ContentTypeOptionsServerHttpHeadersWriter} instead.
 */
@Deprecated(since = "6.2", forRemoval = true)
public class XContentTypeOptionsServerHttpHeadersWriter implements ServerHttpHeadersWriter {
```

joerg-richter-5234:
@marcusdacoregio cheers. I agree that your move is more sensible. will do.

joerg-richter-5234:
As recommended. I opened a pull request to deprecate XContentTypeOptionsServerHttpHeadersWriter in main. 
See https://github.com/spring-projects/spring-security/pull/13119

PR #13077, opened 2023-04-23
Update architecture.adoc

neshkeev:
`RoleHierarchy` doesn't have the `setHierarchy` method, so the snippet doesn't work as is. The method is declared inside `RoleHierarchyImpl`


jzheaux:
Thanks for the PR, @neshkeev! This is now merged into `main`.

PR #13078, opened 2023-04-24
Update acls.adoc

neshkeev:
Fix the spring security samples link


jzheaux:
Thanks again, @neshkeev! This is now merged into `main`.

Issue #11448, opened 2022-06-29
Allow configuring the ActiveDirectoryLdapAuthenticationProvider in AuthenticationManagerBuilder


petrdvorak:
Currently, we are able to set standard LDAP provider via:

```java
@Configuration
public class WebSecurityConfig extends WebSecurityConfigurerAdapter {

    @Override
    public void configure(AuthenticationManagerBuilder auth) throws Exception {
        final LdapAuthenticationProviderConfigurer<AuthenticationManagerBuilder> ldapAuthenticationBuilder
            = auth.ldapAuthentication();
        // ... proceed with additional configuration
    }

}
```

However, the `LdapAuthenticationProviderConfigurer` is hardcoded to create `LdapAuthenticationProvider` in the `build` method ([here](https://github.com/spring-projects/spring-security/blob/main/config/src/main/java/org/springframework/security/config/annotation/authentication/configurers/ldap/LdapAuthenticationProviderConfigurer.java#L98)).

There is no way to setup the configurer to build the `ActiveDirectoryLdapAuthenticationProvider`, which uses a different internal logic on top of the same base `AbstractLdapAuthenticationProvider` class.

To be able to configure Active Directory the same way we currently can configure classic LDAP, we would like to see either of these options:

### Option 1: Own configurer for Active Directory

... providing the following new method:

```java
@Configuration
public class WebSecurityConfig extends WebSecurityConfigurerAdapter {

    @Override
    public void configure(AuthenticationManagerBuilder auth) throws Exception {
        final LdapAuthenticationProviderConfigurer<AuthenticationManagerBuilder> ldapAuthenticationBuilder
            = auth.activeDirectoryAuthentication();
        // ... proceed with additional configuration
    }

}
```

### Option 2: Picking the right class from a registered bean

... instead of creating the class, the configurer could automatically detect a bean:

```java
    @Bean
    @ConditionalOnProperty(name = "my.props.ldap.security.method", havingValue = "active-directory")
    public ActiveDirectoryLdapAuthenticationProvider activeDirectoryLdapAuthenticationProvider(LdapConfiguration configuration) {
        final String activeDirectoryDomain = configuration.getActiveDirectoryDomain();
        final String ldapUrl = configuration.getLdapUrl();
        final String ldapRoot = configuration.getLdapRoot();
        return new ActiveDirectoryLdapAuthenticationProvider(activeDirectoryDomain, ldapUrl, ldapRoot);
    }
```

[...]

Issue #4490, opened 2017-08-09
ActiveDirectoryLdapAuthenticationProvider does not support different AuthoritiesPopulator


GoGoris:
### Summary

The ActiveDirectoryLdapAuthenticationProvider is final and does not use an AuthoritiesPopulator and LdapAuthenticator. 

https://stackoverflow.com/questions/16113115/how-do-i-use-a-custom-authorities-populator-with-spring-security-and-the-actived

### Actual Behavior

I cannot override the loadUserAuthorities method in the ActiveDirectoryLdapAuthenticationProvider.

### Expected Behavior

I expected the ActiveDirectoryAuthenticationProvider to work in the same way as the LdapAuthenticationProvider, with a separate LdapAuthenticator and AuthoritiesPopulator that I can override if needed. 

### Version

The latest version on master still has a seperate loadUserAuthorities method instead of using an AuthoritiesPopulator: https://github.com/spring-projects/spring-security/blob/01579be7d42a0f7c0e75b35a50e46165c14f9c7f/ldap/src/main/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProvider.java

Haarolean:
Six years later this is still an issue. Can we get some triaging as a start?

jzheaux:
@Haarolean, I think this would be a nice feature to add. Would you like to contribute a PR?

Haarolean:
@jzheaux thanks for triaging the issue. Sure, please assign :)

alecdesjardins:
What is the status on this? This would be extremely useful.

Haarolean:
Raised a PR #14539

Haarolean:
@jzheaux just noticed I've made `DefaultActiveDirectoryAuthoritiesPopulator` final, shouldn't we make it extendable on par with `DefaultLdapAuthoritiesPopulator`?

PR #13100, opened 2023-04-26
Impl setting a custom LdapAuthenticationProvider w/ AbstractLdapAuthenticationManagerFactory

Haarolean:
Hi,

I suggested some adjustments within #11448, but as nobody replied I decided to raise a PR.

Additionally,

1) I've considered the approach of adding a property like `useActiveDirectoryComponents`, but since `ActiveDirectoryLdapAuthenticationProvider` would require additional params (like a domain name) it doesn't look applicable. So I rejected this approach.
2) Since `LdapAuthoritiesPopulator` is not applicable to `ActiveDirectoryLdapAuthenticationProvider` (or `AbstractLdapAuthenticationProvider` generally), the code looks like it does. This might be changed after #4490 is implemented. 

Let me know what you think.
Cheers

### DIFF
--- a/config/src/main/java/org/springframework/security/config/ldap/AbstractLdapAuthenticationManagerFactory.java
+++ b/config/src/main/java/org/springframework/security/config/ldap/AbstractLdapAuthenticationManagerFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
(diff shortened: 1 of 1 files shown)


jzheaux:
Thanks, @Haarolean. I'm going to close this PR at this point since providing the full authentication provider could instead be done with `new ProviderManager(authenticationProvider)`. Let's please continue discussing on #11448 as needed.

Issue #11624, opened 2022-07-26
Simplify RequestMatcherDelegatingAuthorizationManager.Builder matcher registration


jzheaux:
It would be easier to use `RequestMatcherDelegatingAuthorizationManager.Builder` if it used builder methods similar to `MessageMatcherDelegatingAuthorizationManager.Builder`.

Specifically, it would be nice to be able to do things like the following:

```java
RequestMatcherDelegatingAuthorizationManager.builder()
    .requestMatchers(new RegexRequestMatcher(...)).hasAuthority("message:read")
    .requestMatchers(new RegexRequestMatcher(...)).hasAuthority("message:read")
    .anyRequest().authenticated()
    .build();
```

Then, when applications need to decorate `AuthorizationManager` behavior, they can still take advantage of the authorization DSL. 

Given #11347, this ticket should only introduce `anyRequest` and `requestMatchers`.

evgeniycheban:
@jzheaux I can work on it.

marcusdacoregio:
Closing in favor of https://github.com/spring-projects/spring-security/pull/13110

PR #13110, opened 2023-04-27
Simplify RequestMatcherDelegatingAuthorizationManager.Builder matcher registration

evgeniycheban:
Simplify RequestMatcherDelegatingAuthorizationManager.Builder matcher registration

Closes gh-11624

### DIFF
--- a/web/src/main/java/org/springframework/security/web/access/intercept/RequestMatcherDelegatingAuthorizationManager.java
+++ b/web/src/main/java/org/springframework/security/web/access/intercept/RequestMatcherDelegatingAuthorizationManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,9 +26,12 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 import org.springframework.core.log.LogMessage;
+import org.springframework.security.authorization.AuthenticatedAuthorizationManager;
+import org.springframework.security.authorization.AuthorityAuthorizationManager;
 import org.springframework.security.authorization.AuthorizationDecision;
 import org.springframework.security.authorization.AuthorizationManager;
 import org.springframework.security.core.Authentication;
+import org.springframework.security.web.util.matcher.AnyRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher.MatchResult;
 import org.springframework.security.web.util.matcher.RequestMatcherEntry;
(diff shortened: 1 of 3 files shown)


web/src/main/java/org/springframework/security/web/access/intercept/RequestMatcherDelegatingAuthorizationManager.java
```diff
@@ -127,11 +134,49 @@ public Builder add(RequestMatcher matcher, AuthorizationManager<RequestAuthoriza
...
 			return this;
 		}
 
+		/**
+		 * Maps an array of string patterns to {@link AuthorizationManager}.
+		 * @param patterns the patterns to map
+		 * @return the {@link AuthorizedUrl} for further customizations
+		 * @since 6.1
+		 */
+		public AuthorizedUrl requestMatchers(String... patterns) {
```
marcusdacoregio, line 149: The `requestMatchers` method should keep the same behavior as `AbstractRequestMatcherRegistry` by creating a `MvcRequestMatcher` if possible and use `AntPathRequestMatcher` as a fallback. https://github.com/spring-projects/spring-security/blob/1631cac1506d7fdda1f3864357075af4965a78cd/config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java#L181

marcusdacoregio (changes requested):
Hi @evgeniycheban, I have left feedback inline.

Can you please ensure that the `@since` tags are updated to `6.2` because we already have a release candidate for 6.1?

web/src/main/java/org/springframework/security/web/access/intercept/RequestMatcherDelegatingAuthorizationManager.java
evgeniycheban: Hi @marcusdacoregio, thanks for the review. I'm not sure if it would be a good idea to propagate an `ApplicationContext` to this builder, maybe we could create here `AntPathRequestMatcher` by default
and let the user register `MvcRequestMatcher` using `requestMatchers(RequestMatcher... matchers)`.

What do you think?

marcusdacoregio: I think you are right, can you make that behavior explicit in the javadoc?

evgeniycheban: I've reflected this behavior in the javadoc and updated `@since` tags to `6.2`.

marcusdacoregio: Hi @evgeniycheban, I am concerned about having the ant matcher the default if Spring MVC is in the classpath. This deviates from the default that Spring Security adopted in 6.0. For now, what if `requestMatchers` only takes a `RequestMatcher` instance? We can think more (in another issue) about how to communicate the MVC nuance to the builder.

evgeniycheban: Hi @marcusdacoregio, makes sense, let's stay only with `requestMatchers(RequestMatcher... matchers)` for now and think about supporting `MvcRequestMatcher` in the future releases. I will update the PR soon.

evgeniycheban: I've updated the PR.

marcusdacoregio:
Thanks @evgeniycheban, this is now merged into `main`

PR #13119, opened 2023-05-01
Suggestion: Deprecate XContentTypeOptionsServerHttpHeadersWriter

joerg-richter-5234:
- Deprecation of XContentTypeOptionsServerHttpHeadersWriter as duplicate of ContentTypeOptionsServerHttpHeadersWriter
- XContentTypeOptionsServerHttpHeadersWriter#X_CONTENT_OPTIONS defines "X-Content-Options", instead of "X-Content-Type-Options"
- X-Content-Options does seem to be listed. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

Following the recommendation by marcusdacoregio on the original request: https://github.com/spring-projects/spring-security/pull/13070

### DIFF
--- a/web/src/main/java/org/springframework/security/web/server/header/XContentTypeOptionsServerHttpHeadersWriter.java
+++ b/web/src/main/java/org/springframework/security/web/server/header/XContentTypeOptionsServerHttpHeadersWriter.java
@@ -25,7 +25,9 @@ import org.springframework.web.server.ServerWebExchange;
  *
  * @author Rob Winch
  * @since 5.0
+ * @deprecated For removal in 7.0. Use {@link ContentTypeOptionsServerHttpHeadersWriter} instead.
  */
+@Deprecated(since = "6.2", forRemoval = true)
 public class XContentTypeOptionsServerHttpHeadersWriter implements ServerHttpHeadersWriter {
 
 	public static final String X_CONTENT_OPTIONS = "X-Content-Options";


marcusdacoregio:
Hello @joerg-richter-5234. After hearing @jzheaux's feedback, I believe we should reconsider our approach to deprecating the `XContentTypeOptionsServerHttpHeadersWriter`.

While I initially [suggested deprecating the `XContentTypeOptionsServerHttpHeadersWriter`](https://github.com/spring-projects/spring-security/pull/13070#issuecomment-1523768533) and keeping the `ContentTypeOptionsServerHttpHeadersWriter`, I now realize that this may cause conflicts in the future if a header called `Content-Type-Options` becomes standardized and we need to add support for it. Additionally, it's worth noting that the servlet version is also named `XContentTypeOptionsHeaderWriter`, so it may be beneficial to keep the names aligned.

With that in mind, I suggest we switch our approach and deprecate the `ContentTypeOptionsServerHttpHeadersWriter` instead.

In addition to that would you be interested in [opening a PR that fixes](https://github.com/spring-projects/spring-security/issues/13155) the constant value in `XContentTypeOptionsServerHttpHeadersWriter`? If so, that new PR should target the `5.7.x` branch.

joerg-richter-5234:
@marcusdacoregio  Sure. I'll pick up https://github.com/spring-projects/spring-security/issues/13155 and target 5.7.x. 

Regarding: `ContentTypeOptionsServerHttpHeadersWriter`
I'll follow your lead on this. What your preferred workflow: Edit this PR or open a new PR ?

Issue #12870, opened 2023-03-16
Add meta-annotation support for EnableMethodSecurity


twelvet-s:
**Describe the bug**
After using the version Latest corresponding to Boot3 and opening the annotation @ EnableMethodSecurity, @ PreAuthorize is still invalid, but using the old @ EnableGlobalMethodSecurity is normal to enter @ PreAuthorize. What is missing or a bug

**To Reproduce**
Use @ EnableMethodSecurity

**Expected behavior**
@PreAuthorize needs to work

evgeniycheban:
Hi, @twelvet-s can you provide a code example for reproducing this issue?

twelvet-s:
> 你好， @twelvet-s你能提供一个代码示例来重现这个问题吗？
![0090065fe1a24dc9f9730aa1c37e285](https://user-images.githubusercontent.com/31510505/235282122-778bfaba-cf7b-413c-ac87-aabed4949cc8.png)

When I use the annotation '@ enableGlobalMethodSecurity', it is normal to enter and determine permissions. After using '@ enableMethodSecurity', it becomes invalid

evgeniycheban:
Hi, @twelvet-s, thanks for the reply, unfortunately, I couldn't reproduce it.

What is meant by "it becomes invalid"? What output are you getting when using `@EnableMethodSecurity`?
Could you please share a repository with the minimal configuration that is close to your project to reproduce this issue?
That would help me with debugging it.

twelvet-s:
https://github.com/twelvet-s/twelvet/tree/master-boot-3

Under this branch

twelvet-s:
Using @ enableGlobalMethodSecurity is a normal entry permission judgment, but it has been replaced by a new one, which has no effect

twelvet-s:
> 

When using the new annotation @ enableMethodSecurity, it will not enter the judgment of the @ PreAuthorize annotation. The deprecated @ enableGlobalMethodSecurity can

evgeniycheban:
It seems that when `@EnableMethodSecurity` is defined on another annotation it doesn't get imported, could you try using  `@EnableMethodSecurity` on the configuration class itself? If that is the case, I will provide a fix.

twelvet-s:
> 

Yes, I really hope to provide some help. Is this a bug? If so, I am also looking forward to fixing it because I tend to use it this way. Thank you very much for your help

evgeniycheban:
[...]

PR #13120, opened 2023-05-01
Add meta-annotation support for EnableMethodSecurity

evgeniycheban:
EnableMethodSecurity annotation does not get imported when defined as a meta-annotation

Closes gh-12870


marcusdacoregio:
Thanks @evgeniycheban, this is now merged into `main`

PR #13137, opened 2023-05-06
Debug message in OpenSamlAuthenticationProviders is never logged

### LINKED ISSUE #13136: Saml logging: Debug message in OpenSamlAuthenticationProvider is never logged since it's wrapped by isTraceEnabled()
**Describe the bug**
In SAML OpenSamlAuthenticationProvider and OpenSaml4AuthenticationProvider there is a debug message that never gets logged if the log level is debug since it's wrapped by a `isTraceEnabled()`

stnor:
Fixes #13136


jzheaux:
Sorry, @stnor, I merged that too quickly. The if statements are not in error. As you can see, the message is logged with more detail if trace is enabled, but is still logged in the `else` block when debug is enabled.

stnor:
Ah, sorry.

Issue #13134, opened 2023-05-05
authorization-grant-type does not support common name "jwt-bearer"


alexis-valois:
**Describe the bug**
It is possible to use "urn:ietf:params:oauth:grant-type:jwt-bearer" as a valid authorization-grant-type. 
However, Spring-Security does nothing if the grant_type uses the "short" notation : "jwt-bearer".
In fact, the filter will just return an empty Mono in this situation)

The authentication POST call is then forged using the authorization-grant-type value as the "grant_type" field value.
Some identity providers expect to see the exact string "jwt-bearer", which is problematic because there's no way to configure spring-security oauth client as such.

**To Reproduce**

Set a spring-security oauth-client project using those application.yml configurations. You'll see that nothing happens, regarding the authentication tokens fetch operations, even if we apply `.oauth2Client()` in a `SecurityWebFilterChain` bean.

# Application.java
```
@SpringBootApplication
public class Application {
    public static void main(String[] args) {
        SpringApplication.run(Application .class, args);
    }
}
```
# application.yml
```
security:
    oauth2:
      client:
        registration:
          <provider>:
            client-authentication-method : client_secret_post
            authorization-grant-type: jwt-bearer
            client-id: <client id>
            client-secret: <client secret>
```
# Configuration Beans
```
@Configuration
@EnableWebFluxSecurity
public class OAuth2ClientSecurityConfiguration {

    @Bean
    public ReactiveOAuth2AuthorizedClientManager authorizedClientManager(
            ReactiveClientRegistrationRepository clientRegistrationRepository,
            ServerOAuth2AuthorizedClientRepository authorizedClientService) {

        var jwtBearerAuthorizedClientProvider  =
               new JwtBearerReactiveOAuth2AuthorizedClientProvider();

        ReactiveOAuth2AuthorizedClientProvider authorizedClientProvider =
                ReactiveOAuth2AuthorizedClientProviderBuilder.builder()
                        .provider(jwtBearerAuthorizedClientProvider)
                        .build();

        var authorizedClientManager =
                new DefaultReactiveOAuth2AuthorizedClientManager(
                        clientRegistrationRepository, authorizedClientService);
        authorizedClientManager.setAuthorizedClientProvider(authorizedClientProvider);

        return authorizedClientManager;
    }

    @Bean
    public SecurityWebFilterChain configure(ServerHttpSecurity http) {
        return http
                .oauth2Client()
                .and()
                .build();
    }

    @Bean
[...]

PR #13178, opened 2023-05-14
Address jwt-bearer short (common) notation

alexis-valois:
Closes gh-13134

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/JwtBearerOAuth2AuthorizedClientProvider.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/JwtBearerOAuth2AuthorizedClientProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -58,8 +58,9 @@ public final class JwtBearerOAuth2AuthorizedClientProvider implements OAuth2Auth
 	 * {@code context}. Returns {@code null} if authorization (or re-authorization) is not
 	 * supported, e.g. the client's {@link ClientRegistration#getAuthorizationGrantType()
 	 * authorization grant type} is not {@link AuthorizationGrantType#JWT_BEARER
-	 * jwt-bearer} OR the {@link OAuth2AuthorizedClient#getAccessToken() access token} is
-	 * not expired.
+	 * urn:ietf:params:oauth:grant-type:jwt-bearer} OR
+	 * {@link AuthorizationGrantType#JWT_BEARER_COMMON_NOTATION jwt-bearer} OR the
+	 * {@link OAuth2AuthorizedClient#getAccessToken() access token} is not expired.
 	 * @param context the context that holds authorization-specific state for the client
 	 * @return the {@link OAuth2AuthorizedClient} or {@code null} if authorization is not
 	 * supported
@@ -69,7 +70,9 @@ public final class JwtBearerOAuth2AuthorizedClientProvider implements OAuth2Auth
 	public OAuth2AuthorizedClient authorize(OAuth2AuthorizationContext context) {
 		Assert.notNull(context, "context cannot be null");
 		ClientRegistration clientRegistration = context.getClientRegistration();
-		if (!AuthorizationGrantType.JWT_BEARER.equals(clientRegistration.getAuthorizationGrantType())) {
+		if (!AuthorizationGrantType.JWT_BEARER.equals(clientRegistration.getAuthorizationGrantType())
+				|| !AuthorizationGrantType.JWT_BEARER_COMMON_NOTATION
+						.equals(clientRegistration.getAuthorizationGrantType())) {
 			return null;
 		}
 		OAuth2AuthorizedClient authorizedClient = context.getAuthorizedClient();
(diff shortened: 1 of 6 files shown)


jzheaux:
@alexis-valois, thanks for the suggestion. I'm not familiar with the term "Jwt Bearer Common Notation". Is this in a spec somewhere? For it to be "first-class" like this in the code, it usually needs to be in an RFC.

Just for some context, is there a reason your authorization server is unable to follow the spec? It seems like that is the ideal place to try and address this issue.

alexis-valois:
Hello @jzheaux. Here's a bit more of context about the situation that led me to open this issue : 

at my organization, we are trying to authenticate with an external service provider that has its own oauth2 Identity Server(we do not own or have control over that auth server). 

From my understanding, the product used on their side for that purpose is [IdentityServer4](https://github.com/IdentityServer/IdentityServer4) wich is a C# dotnetcore solution. 

On our side, with use Java with SpringBoot and SpringSecurity to interface with them. 

Our service provider is not able to make any change in their authentication system for us. 

So, even if we were able to make the change directly on IdentityServer4's repo, that would not help us with our problem, because that would mean that they would have to update their system and re-deploy a new one with the new IdentityServer4 build. 

Anyway, in a ideal world, I think that both solutions should be able to recognize "jwt-bearer" as a synonym to "urn:ietf:params:oauth:grant-type:jwt-bearer", as its the case with "client_credentials" and others.

To answer your question, there's no offical name such as "short notation" or "common notation" : thats just the best name I came up with.

According to the [RFC 7521 page 7](https://www.rfc-editor.org/rfc/rfc7521#page-7), I think the use of "urn:ietf:params:oauth:grant-type" is suggested, but not enforced. That may explain why IndentityServer4 uses "jwt-bearer"

jzheaux:
Gotcha. Thanks for the extra background.

Typically, we avoid adding implementation details that aren't spec-compliant. That said, I think there is something else we can do here.

I added #13181 to make it easier to configure a custom provider. This takes the custom config you posted in #13134 down to:

```java
@Bean
ReactiveOAuth2AuthorizedClientProvider authorizedClientProvider() {
   return new JwtBearerReactiveOAuth2AuthorizedClientProvider(...);
}
```

And then, this PR could introduce a constructor that takes a custom `AuthorizationGrantType`, which I think creates a better balance in how much configuration code you are needing to write to integrate with IdentityServer4. It would look like this instead then:

```java
@Bean
ReactiveOAuth2AuthorizedClientProvider authorizedClientProvider() {
   return new JwtBearerReactiveOAuth2AuthorizedClientProvider(new AuthorizationGrantType("jwt-bearer"));
}
```

Would this achieve what you are hoping for? 

If so, would you please update your PR to add a no-op constructor and one that takes the `AuthorizationGrantType` to check? Ideally, you'd do this in both the reactive and imperative implementations. And finally, it would be helpful if you could add unit tests that confirm the new constructor works.

alexis-valois:
Thanks for the reply @jzheaux 

Will that change makes it possible to pass the check at that line ? : https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/JwtBearerOAuth2AuthorizedClientProvider.java#L72

For what I understand, event if I supply an AuthorizationGrantType with the value "jwt-bearer", the "if" statement will still evaluate to "false" and the "authorize" method will still return an empty Mono, without actually generating a JwtAssertion, calling the authProvider and so on. 

Or am I missing something in your solution...

alexis-valois:
Yes ! I understand thx, im on it

jgrandja:
@jzheaux 

> And then, this PR could introduce a constructor that takes a custom `AuthorizationGrantType`

I'm not sure about this proposal. Even if we went with it, there still will be an issue with `JwtBearerGrantRequest` since it sets `AuthorizationGrantType.JWT_BEARER` and does not allow it to be changed.

The intent of `OAuth2AuthorizedClientProviderBuilder.builder().provider()` is to allow configuring a custom `OAuth2AuthorizedClientProvider`, so a similar implementation of `JwtBearerOAuth2AuthorizedClientProvider` can be extended there. IMO, the code is quite minimal in `JwtBearerOAuth2AuthorizedClientProvider` so copying it to a custom implementation should be straight forward.

From an API standpoint, I'm not sure this makes sense `public JwtBearerOAuth2AuthorizedClientProvider(AuthorizationGrantType authorizationGrantType)`.

jgrandja:
@alexis-valois If all you need to do is replace `grant_type=urn:ietf:params:oauth:grant-type:jwt-bearer` to `grant_type=jwt-bearer` then you can configure a custom `JwtBearerGrantRequestEntityConverter`. See [Customizing the Access Token Request](https://docs.spring.io/spring-security/reference/servlet/oauth2/client/authorization-grants.html#_customizing_the_access_token_request_5) in the reference.

alexis-valois:
@jgrandja I tried using a custom JwtBearerGrantRequestEntityConverter but unfortunatly, that did not do what I was looking for. That solution indeed allows me to replace the "grant_type" value, but the jwt-bearer flow itself (sending a request for an oauth token, checking if it is valid, refecth another one when it expires, etc) is not enabled, because the If condition mentionned above doesn't pass, and a simple Empty mono is returned instead. 

Ideally, we would prefer not to copy the JwtBearerOAuth2AuthorizedClientProvider in our codebase to implement it ourself. We would prefer keeping the most we can from Spring-Security to benefit from any updates and security fixes that will occur in the future.

FYI : I tried the solution proposed by @jzheaux locally and it fixes all my problems.

alexis-valois:
@jgrandja I saw that configuration in the docs, but I wasn't sure if that was possible to do the same with the Reactive App implementation. Is it ?

jgrandja:
@alexis-valois 

> I wasn't sure if that was possible to do the same with the Reactive App implementation. Is it ?

Yes it is. Please see [reference doc](https://docs.spring.io/spring-security/reference/reactive/oauth2/client/authorization-grants.html#_customizing_the_access_token_request_5).

alexis-valois:
@jgrandja that works indeed ! I've had trouble finding my way in the docs, sorry for the confusion. I'll close the PR and issue. Thanks again !

jgrandja:
Glad you got it working @alexis-valois ! I'll close the issue as well.

PR #13179, opened 2023-05-15
Use consistent list of micrometer tags in web observation handler

dfrommi:
The tag `spring.security.reached.filter.name` is only set if a filter-name is available, otherwise the tag is omitted entirely. This leads to issues with metric-exporters that don't support dynamic tags, but rather expect tag-names of a metric to be always the same. The most prominent example is the Prometheus-exporter.

Instead of omitting the tag if no filer-name is set, a none-value is applied instead, making the tag-list consistent in all cases

### DIFF
--- a/web/src/main/java/org/springframework/security/web/ObservationFilterChainDecorator.java
+++ b/web/src/main/java/org/springframework/security/web/ObservationFilterChainDecorator.java
@@ -37,6 +37,7 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 import org.springframework.core.log.LogMessage;
+import org.springframework.util.StringUtils;
 
 /**
  * A {@link org.springframework.security.web.FilterChainProxy.FilterChainDecorator} that
@@ -519,8 +520,8 @@ public final class ObservationFilterChainDecorator implements FilterChainProxy.F
 			return KeyValues.of(CHAIN_SIZE_NAME, String.valueOf(context.getChainSize()))
 					.and(CHAIN_POSITION_NAME, String.valueOf(context.getChainPosition()))
 					.and(FILTER_SECTION_NAME, context.getFilterSection())
-					.and(FILTER_NAME, (context.getFilterName() != null && !context.getFilterName().isEmpty())
-							? context.getFilterName() : KeyValue.NONE_VALUE);
+					.and(FILTER_NAME, (StringUtils.hasText(context.getFilterName())) ? context.getFilterName()
+							: KeyValue.NONE_VALUE);
 		}
 
 		@Override
--- a/web/src/main/java/org/springframework/security/web/ObservationFilterChainDecorator.java
+++ b/web/src/main/java/org/springframework/security/web/ObservationFilterChainDecorator.java
@@ -22,6 +22,7 @@ import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 
+import io.micrometer.common.KeyValue;
 import io.micrometer.common.KeyValues;
 import io.micrometer.observation.Observation;
 import io.micrometer.observation.ObservationConvention;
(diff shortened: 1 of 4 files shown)


web/src/main/java/org/springframework/security/web/ObservationFilterChainDecorator.java
```diff
@@ -568,13 +569,12 @@ public String getContextualName(FilterChainObservationContext context) {
 
 		@Override
 		public KeyValues getLowCardinalityKeyValues(FilterChainObservationContext context) {
-			KeyValues kv = KeyValues.of(CHAIN_SIZE_NAME, String.valueOf(context.getChainSize()))
+			return KeyValues.of(CHAIN_SIZE_NAME, String.valueOf(context.getChainSize()))
```
jzheaux, line 572: Will you please make the same change in `ObservationWebFIlterChainDecorator`, the reactive version?

web/src/main/java/org/springframework/security/web/ObservationFilterChainDecorator.java
```diff
@@ -568,13 +569,12 @@ public String getContextualName(FilterChainObservationContext context) {
...
 					.and(CHAIN_POSITION_NAME, String.valueOf(context.getChainPosition()))
-					.and(FILTER_SECTION_NAME, context.getFilterSection());
-			if (context.getFilterName() != null) {
-				kv = kv.and(FILTER_NAME, context.getFilterName());
-			}
-			return kv;
+					.and(FILTER_SECTION_NAME, context.getFilterSection())
+					.and(FILTER_NAME, (context.getFilterName() != null)
+							? context.getFilterName()
+							: KeyValue.NONE_VALUE);
```
jzheaux, line 577: Are you able to add a unit test in `ObservationFilterChainDecoratorTests` and `ObservationWebFilterChainDecoratorTests` that confirms that the observation handler gets `NONE_VALUE` for the value in this case?

jzheaux (changes requested):
Thanks, @dfrommi, for the PR! I've left some feedback inline.

web/src/main/java/org/springframework/security/web/ObservationFilterChainDecorator.java
dfrommi: Done

web/src/main/java/org/springframework/security/web/ObservationFilterChainDecorator.java
dfrommi: Sure, I will try my best.
Before I start from scratch, is there already a test that verifies tagging? I couldn't find one. Thanks!

jzheaux: Here's a sample to get you started:

```java
@Test
void decorateFiltersWhenCompletesThenHasSpringSecurityReachedFilterNameTag() throws Exception {
	ObservationHandler<Observation.Context> handler = mock(ObservationHandler.class);
	given(handler.supportsContext(any())).willReturn(true);
	ObservationRegistry registry = ObservationRegistry.create();
	registry.observationConfig().observationHandler(handler);
	ObservationFilterChainDecorator decorator = new ObservationFilterChainDecorator(registry);
	FilterChain chain = mock(FilterChain.class);
	Filter filter = spy(new BasicAuthenticationFilter());
	FilterChain decorated = decorator.decorate(chain, List.of(filter));
	decorated.doFilter(new MockHttpServletRequest("GET", "/"), new MockHttpServletResponse());
	ArgumentCaptor<Observation.Context> context = ArgumentCaptor.forClass(Observation.Context.class);
	verify(handler, times(3)).onScopeClosed(context.capture());
	assertThat(context.getValue().getLowCardinalityKeyValue("spring.security.reached.filter.name"))
		.isEqualTo("BasicAuthenticationFilter");
}
```

Let me know if you get stuck. I think this would need to change to simulate the circumstance where the filter name is null.

dfrommi:
@jzheaux Thanks for the test-example.

As requested, I've added the same change to the reactive class as well.
I've also extended the condition slightly to use the `NONE_VALUE` also for empty strings. I think this makes perfect sense and allowed me to test it by using an anonymous class.
For the `null` case, I'd need something that returns `null` for `getClass().getSimpleName()` and I couldn't find anything that does.

jzheaux (approved):

jzheaux:
Thanks, @dfrommi, for the PR! This is merged into `6.0.x` and `main` and will go out in the next point release.

Issue #9900, opened 2021-06-11
Support for JWT Header TYP as "at+jwt"


tekgainers:
**Expected Behavior**

Currently, if the JWT is having typ as "at+jwt", the token is rejected with message "Failed to authenticate since the JWT was invalid". Spring Security Oauth2 Resource Server with JWT as bearer token should accept typ as "at+jwt" as well.

**Current Behavior**

Currently, if the JWT is having typ as "at+jwt", the token is rejected with message "Failed to authenticate since the JWT was invalid".

My Authorization server is issuing JWT access token with typ as "at+jwt" as per the following draft: https://datatracker.ietf.org/doc/html/draft-ietf-oauth-access-token-jwt-11

How has this issue affected you? JWT token is rejected although this is correct as per Authorization server
What are you trying to accomplish? Validate JWT Bearer token using Spring Security OAuth2 Resource Server capabilities.
What other alternatives have you considered? At this point, in the documentation I didn;t find a way where I can customize the JWT Validator. Although, documentation is around our own Audience or other validations once the JWT is accepted by Spring Security.
Are you aware of any workarounds? Not at this point of time.

tekgainers:
@jzheaux can you please have a look if this is bug or this needs enhancement in library?

Thanks in advance.

jzheaux:
The reason for this behavior is Nimbus defaults to checking that the `typ` is `JWT`. I am not certain whether Nimbus yet supports [the `at+jwt` draft](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-access-token-jwt-11#section-4) - Spring Security would likely rely on Nimbus for this kind of support.

To get the resource server to accept `"typ" : "at+jwt"`, you can configure Nimbus like so:

```java
@Bean 
JwtDecoder jwtDecoder() {
	DefaultJOSEObjectTypeVerifier<SecurityContext> verifier =
		new DefaultJOSEObjectTypeVerifier<>(new JOSEObjectType("at+jwt"));
	NimbusJwtDecoder decoder = NimbusJwtDecoder.withJwkSetUri(this.uri)
		.jwtProcessorCustomizer((processor) -> processor.setJWSTypeVerifier(verifier))
		.build()
        // ... any other decoder settings
	return decoder;
}
```

For passivity reasons, I imagine that Spring Security will continue to use the Nimbus defaults. Once the draft gets finalized, the may be an opportunity to make this configuration simpler.

ymajoros:
[...]

PR #13186, opened 2023-05-16
Allow at+jwt, according to RFC-9068

ymajoros:
Closes 13185

### DIFF
--- a/docs/modules/ROOT/pages/migration/oauth2.adoc
+++ b/docs/modules/ROOT/pages/migration/oauth2.adoc
@@ -1,0 +1,172 @@
+= OAuth 2.0 Changes
+
+== Validate `typ` Header with `JwtTypeValidator`
+
+`NimbusJwtDecoder` in Spring Security 7 will move `typ` header validation to `JwtTypeValidator` intsead of relying on Nimbus.
+This brings it in line with `NimbusJwtDecoder` validating claims instead of relying on Nimbus to validate them.
+
+If you are changing Nimbus's default type validation in a `jwtProcessorCustomizer` method, then you should move that to `JwtTypeValidator` or an implementation of `OAuth2TokenValidator` of your own.
+
+To check if you are prepared for this change, add the default `JwtTypeValidator` to your list of validators, as this will be included by default in 7:
+
+[tabs]
+======
+Java::
++
+[source,java,role="primary"]
+----
+@Bean
+JwtDecoder jwtDecoder() {
+	NimbusJwtDecoder jwtDecoder = NimbusJwtDecoder.withIssuerLocation(location)
+        .validateTypes(false) <1>
+        // ... your remaining configuration
+        .build();
+	jwtDecoder.setJwtValidator(JwtValidators.createDefaultWithValidators(
+		new JwtIssuerValidator(location), JwtTypeValidator.jwt())); <2>
+	return jwtDecoder;
+}
+----
+
+Kotlin::
++
+[source,kotlin,role="secondary"]
+----
+@Bean
+fun jwtDecoder(): JwtDecoder {
+    val jwtDecoder = NimbusJwtDecoder.withIssuerLocation(location)
+        .validateTypes(false) <1>
+        // ... your remaining configuration
+        .build()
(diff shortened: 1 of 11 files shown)


jzheaux:
@ymajoros, thank you very much for the PR. I don't think that it should be in `JwtDecoders`, but `NimbusJwtDecoder`. Also, it is missing the needed validation. Please see [my comment](https://github.com/spring-projects/spring-security/issues/9900#issuecomment-1555274171) about what needs to be done to add this support.

jzheaux:
@ymajoros, I don't think we should add this feature without also introducing validation. Allow me to elaborate on what I meant [here](https://github.com/spring-projects/spring-security/pull/13186#issuecomment-1577358744).

I think it would be valuable for `NimbusJwtDecoder`'s builders to turn off the Nimbus `typ` validation so that applications can better choose which validation types, `JWT`, `at+jwt`, etc. they want to support.

The builders could change like so:

```java
NimbusJwtDecoder.withIssuerLocation(issuer)
    .useNimbusTypeVerifier(false).build();
```

Then, we can change `JwtValidators` to add `typ` validation by default similar to Nimbus, allowing the following:

```java
NimbusJwtDecoder decoder = NimbusJwtDecoder.withIssuerLocation(issuer)
    .useNimbusTypeVerifier(false).build();
decoder.setJwtValidator(JwtValidators.createDefaultWithIssuer(issuer));
```

to remain an otherwise passive change.

Then, this PR would also introduce a new validator for `at+jwt`, something like:

```java
JwtValidators.createDefaultForAtJwt(issuer);
```

So then to turn on `at+jwt` support, that would look like this:

```java
NimbusJwtDecoder decoder = NimbusJwtDecoder.withIssuerLocation(issuer)
    .useNimbusTypeVerifier(false).build();
decoder.setJwtValidator(JwtValidators.createDefaultForAtJwt(issuer));
```

Is this something you'd still be interested in putting together? If not, I can mark this as `ideal-for-contribution` so others know that it's available.

ymajoros:
> @ymajoros, I don't think we should add this feature without also introducing validation. Allow me to elaborate on what I meant [here](https://github.com/spring-projects/spring-security/pull/13186#issuecomment-1577358744).
> 
> I think it would be valuable for `NimbusJwtDecoder`'s builders to turn off the Nimbus `typ` validation so that applications can better choose which validation types, `JWT`, `at+jwt`, etc. they want to support.
> 
> The builders could change like so:
> 
> ```java
> NimbusJwtDecoder.withIssuerLocation(issuer)
>     .useNimbusTypeVerifier(false).build();
> ```
> 
> Then, we can change `JwtValidators` to add `typ` validation by default similar to Nimbus, allowing the following:
> 
> ```java
> NimbusJwtDecoder decoder = NimbusJwtDecoder.withIssuerLocation(issuer)
>     .useNimbusTypeVerifier(false).build();
> decoder.setJwtValidator(JwtValidators.createDefaultWithIssuer(issuer));
> ```
> 
> to remain an otherwise passive change.
> 
> Then, this PR would also introduce a new validator for `at+jwt`, something like:
> 
> ```java
> JwtValidators.createDefaultForAtJwt(issuer);
> ```
> 
> So then to turn on `at+jwt` support, that would look like this:
> 
> ```java
> NimbusJwtDecoder decoder = NimbusJwtDecoder.withIssuerLocation(issuer)
>     .useNimbusTypeVerifier(false).build();
> decoder.setJwtValidator(JwtValidators.createDefaultForAtJwt(issuer));
> ```
> 
> Is this something you'd still be interested in putting together? If not, I can mark this as `ideal-for-contribution` so others know that it's available.

Hello, thanks for the analysis. TBH, I created this years ago because a colleague from security asked for it, but I don't even work there anymore and I won't follow this in any case. I just think being able to follow standards is always a good option, but I have no sponsor or personal interest in this anymore. 

Thanks for your feedback, anyway.

jzheaux:
No problem at all @ymajoros -- I like making sure before making a change like that to a PR, so thank you for responding as quickly as you did.

I've moved this to `ideal-for-contribution` to invite others to contribute and complete the PR.

jzheaux:
I've been talking with the team about this, and since it likely has a flag that will require migration between 6.5 and 7, I'll take this up myself instead.

PR #13212, opened 2023-05-23
Fix type on method-security.adoc

mariodmpereira:


jzheaux:
Thanks, @mariodmpereira! This is now merged into `main`.

Issue #13203, opened 2023-05-21
Additional filters registered when using Custom DSL


bwgjoseph:
I'm using 

- Spring Boot 2.7.12
- Spring Security 5.8.3

(I did test with Spring Boot 3.1.0, and I am facing the same issue)

**Describe the bug**
While playing around with [Custom DSL](https://docs.spring.io/spring-security/reference/5.8/servlet/configuration/java.html#jc-custom-dsls), I noticed 2 additional filters are being applied when using `custom dsl`.

- DefaultLoginPageGeneratingFilter
- DefaultLogoutPageGeneratingFilter

Not exactly sure if this is intended behavior, but I thought it shouldn't be?

**To Reproduce**
Here's the following code

```java
@Configuration(proxyBeanMethods = false)
@EnableWebSecurity(debug = true)
public class WebSecurityConfig {
    @Bean
    public SecurityFilterChain docsFilterChain(HttpSecurity http) throws Exception {
        return http
            // .apply(DummyDsl.dummyDsl())
            // .and()
            .build();
    }
}

public class DummyDsl extends AbstractHttpConfigurer<DummyDsl, HttpSecurity> {
    @Override
    public void init(HttpSecurity http) throws Exception {
        http.formLogin(AbstractHttpConfigurer::disable);
    }

    public static DummyDsl dummyDsl() {
        return new DummyDsl();
    }
}
```

When `custom dsl` is not applied, this is the filter chain

```
Security filter chain: [
  DisableEncodeUrlFilter
  WebAsyncManagerIntegrationFilter
  SecurityContextHolderFilter
  HeaderWriterFilter
  CsrfFilter
  LogoutFilter
  RequestCacheAwareFilter
  SecurityContextHolderAwareRequestFilter
  AnonymousAuthenticationFilter
  ExceptionTranslationFilter
]
```

But when it is being applied, this is the filter chain

```
Security filter chain: [
  DisableEncodeUrlFilter
  WebAsyncManagerIntegrationFilter
  SecurityContextHolderFilter
  HeaderWriterFilter
  CsrfFilter
  LogoutFilter
  DefaultLoginPageGeneratingFilter
  DefaultLogoutPageGeneratingFilter
  RequestCacheAwareFilter
  SecurityContextHolderAwareRequestFilter
  AnonymousAuthenticationFilter
  ExceptionTranslationFilter
]
```

**Expected behavior**
I should expect that the filter chain should be same across both setup, unless I am missing something (I hope not)?

[...]

PR #13235, opened 2023-05-26
Does not apply a Configurer when disabled from another DSL

marcusdacoregio:
Closes gh-13203


jzheaux (changes requested):
It would be nice to have a test that directly demonstrates that the reported arrangement now works. For example, one where `formLogin` is disabled in a configurer, and it's clear that the default login/logout filters are not added.

Issue #13144, opened 2023-05-08
Provide meaningful error when invalid client-authentication-method is provided


rwinch:
If an invalid client-authentication-method is configured (eg "basic"), then the authorization code request is made with no credentials which makes troubleshooting more difficult. Instead, a meaningful error message should be provided on the client if the client-authentication-method is invalid.

PR #13240, opened 2023-05-26
Error On Unsupported Client Authentication Methods

jzheaux:
Related to #13144


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultAuthorizationCodeTokenResponseClient.java
```diff
@@ -132,4 +134,26 @@ public void setRestOperations(RestOperations restOperations) {
...
+			ClientRegistration clientRegistration = authorizationCodeGrantRequest.getClientRegistration();
+			ClientAuthenticationMethod clientAuthenticationMethod = clientRegistration.getClientAuthenticationMethod();
+			String registrationId = clientRegistration.getRegistrationId();
+			boolean supportedClientAuthenticationMethod = clientAuthenticationMethod
+					.equals(ClientAuthenticationMethod.NONE)
+					|| clientAuthenticationMethod.equals(ClientAuthenticationMethod.CLIENT_SECRET_BASIC)
+					|| clientAuthenticationMethod.equals(ClientAuthenticationMethod.CLIENT_SECRET_POST);
+			Assert.isTrue(supportedClientAuthenticationMethod, () -> String.format(
+					"This class supports `client_secret_basic`, `client_secret_post`, and `none` by default. Client [%s] is using [%s] instead. Please use a supported client authentication method, or use `setRequestEntityConverter` to supply an instance that supports [%s].",
+					registrationId, clientAuthenticationMethod, clientAuthenticationMethod));
```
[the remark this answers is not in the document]
sjohnr: We should also add something similar to the other converters (or their corresponding access token clients) and reactive side as well?

sjohnr (changes requested):
Looks like a good start, @jzheaux! I've added my thoughts below.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultAuthorizationCodeTokenResponseClient.java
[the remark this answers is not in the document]
jzheaux: The concern with adding the logic into the converters is that it would break the extensibility of the converter. For example, if the converter errors, I cannot augment the behavior by adding `ClientAuthorizationMethod` of my own by subclassing the converter or delegating to it.

IMO, it is the client that should enforce what client authentication methods it supports. The reason I wrapped the converter is to make it simpler to override in the event that the extra checks break someone. Perhaps it would be better instead to check input conditions in the `getTokenResponse` method. 

I think the concern is that there is no way to articulate what client authentication methods this client supports. But, really it is the server that states what it supports. So, instead, another way would be to check the response code. If it is a 401, then we throw an exception that details the client authentication method specified and how it may be incorrect, something like:

```java
catch (RestClientException ex) {
  if (status code is 401 && clientAuthenticationMethod is neither BASIC nor POST) {
      throw new OAuth2AuthorizationException("The client returned a 401. It appears you could be using a custom client authentication method [%s] instead of the more typical client_secret_basic or client_secret_post. Note that Spring Security recently changed the value of basic to client_secret_basic and post to client_secret_post. Please double-check your configuration for ... etc. ... you may want to change your method to ... or you may need to implement a custom request entity converter ... ")
  } else {
      // existing behavior
  }
}
```

or similar.

Even better, if this can read the `WWW-Authenticate` header in the response, the error message can be even more helpful.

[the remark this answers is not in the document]
sjohnr: That all makes sense, @jzheaux. After discussing with you, we're in agreement that handling the 401 error provides the right opportunity for a more descriptive error message that helps developers discover the root cause of missing client authentication.

Furthermore, due to the customization options available with request entity converters, the approach in this PR to wrap the converter internally seems the best way to passively add this validation. If anyone customizes the access token response client, they will undoubtedly customize the `requestEntityConverter`, which would cause this validation to back off. There may be times when they still need the validation after customizing, but the 80/20 rule seems applicable here.

We discussed some other ideas, and the only one I'm still wondering about is if there's something we can add in Spring Boot to detect `basic` or `post`. I feel like a warning could be logged in Boot for those two strings in particular. It shouldn't fail though. But that is a separate conversation, and I may be off there still.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/AbstractWebClientReactiveOAuth2AccessTokenResponseClient.java
```diff
@@ -106,6 +101,34 @@ public Mono<OAuth2AccessTokenResponse> getTokenResponse(T grantRequest) {
...
+
+	private void validateClientAuthenticationMethod(T grantRequest) {
+		ClientRegistration clientRegistration = grantRequest.getClientRegistration();
+		ClientAuthenticationMethod clientAuthenticationMethod = clientRegistration.getClientAuthenticationMethod();
+		boolean supportedClientAuthenticationMethod = clientAuthenticationMethod.equals(ClientAuthenticationMethod.NONE)
+				|| clientAuthenticationMethod.equals(ClientAuthenticationMethod.CLIENT_SECRET_BASIC)
+				|| clientAuthenticationMethod.equals(ClientAuthenticationMethod.CLIENT_SECRET_POST);
+		if (!supportedClientAuthenticationMethod) {
+			throw new IllegalArgumentException(String.format(
+					"This class supports `client_secret_basic`, `client_secret_post`, and `none` by default. Client [%s] is using [%s] instead. Please use a supported client authentication method, or use `setRequestEntityConverter` to supply an instance that supports [%s].",
```
sjohnr, line 117: Is the error message accurate here? I'm guessing the `requestEntityConverter` is intended to remain hidden, and it should specify something like "... set/add headersConverter or set/add parametersConverter ..." instead?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultClientCredentialsTokenResponseClient.java
```diff
@@ -58,7 +60,7 @@
 
 	private static final String INVALID_TOKEN_RESPONSE_ERROR_CODE = "invalid_token_response";
 
-	private Converter<OAuth2ClientCredentialsGrantRequest, RequestEntity<?>> requestEntityConverter = new OAuth2ClientCredentialsGrantRequestEntityConverter();
+	private Converter<OAuth2ClientCredentialsGrantRequest, RequestEntity<?>> requestEntityConverter = new ClientAuthenticationMethodVerifyingRequestEntityConverter();
```
sjohnr, line 63: Probably just preference again, but I feel like instead of delegation, introducing a boolean flag here would work as well. It seems nice to me also because it makes things slightly easier to understand and debug.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultJwtBearerTokenResponseClient.java
```diff
@@ -124,4 +126,26 @@ public void setRestOperations(RestOperations restOperations) {
...
+			ClientRegistration clientRegistration = jwtBearerGrantRequest.getClientRegistration();
+			ClientAuthenticationMethod clientAuthenticationMethod = clientRegistration.getClientAuthenticationMethod();
+			String registrationId = clientRegistration.getRegistrationId();
+			boolean supportedClientAuthenticationMethod = clientAuthenticationMethod
+					.equals(ClientAuthenticationMethod.NONE)
+					|| clientAuthenticationMethod.equals(ClientAuthenticationMethod.CLIENT_SECRET_BASIC)
+					|| clientAuthenticationMethod.equals(ClientAuthenticationMethod.CLIENT_SECRET_POST);
+			Assert.isTrue(supportedClientAuthenticationMethod, () -> String.format(
+					"This class supports `client_secret_basic`, `client_secret_post`, and `none` by default. Client [%s] is using [%s] instead. Please use a supported client authentication method, or use `setRequestEntityConverter` to supply an instance that supports [%s].",
+					registrationId, clientAuthenticationMethod, clientAuthenticationMethod));
```
sjohnr, line 145: Could we re-use some of this validation across the implementations?

sjohnr (changes requested):
Thanks for the updates @jzheaux, this is looking really good! I've added just a few more comments below.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultClientCredentialsTokenResponseClient.java
```diff
@@ -129,4 +132,26 @@ public void setRestOperations(RestOperations restOperations) {
 		this.restOperations = restOperations;
 	}
 
+	private static final class ClientAuthenticationMethodVerifyingRequestEntityConverter
+			implements Converter<OAuth2ClientCredentialsGrantRequest, RequestEntity<?>> {
```
sjohnr, line 136: Looks like the reusable class was only used in `DefaultAuthorizationCodeTokenResponseClient`?

sjohnr (approved):
Thanks @jzheaux, I added one new item below. I'm good once this is addressed. 👍

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultClientCredentialsTokenResponseClient.java
jzheaux: Good catch, thanks.

PR #12555, opened 2023-01-18
Add empty authorities by default to easily build a user

stillya:
Issue - [spring-projectsgh](https://github.com/spring-projects/spring-security/issues/12533)

PR #13290, opened 2023-06-08
The current version breaks backwards compatibility

dkorotych:
Broken backwards compatibility. There is currently no way to reset the `this.authorities` list.
For example, the code will not work correctly,

```java
final User.UserBuilder builder = User.withUserDetails(userDetails);
if (CollectionUtils.isEmpty(authorities)) {
  builder.authorities(Collections.emptyList());
} else {
  builder.authorities(authorities.toArray(String[]::new));
}
userDetailsManager.updateUser(builder.build());
```
it will always merge new and old authorities, but the previous version will overwrite them


jzheaux:
This appears related to https://github.com/spring-projects/spring-security/pull/12555

PR #13307, opened 2023-06-12
Replace deprecated methods

EvaristeGalois11:
Hi, I replaced some deprecated methods.

This doesn't want to be a full cleansing of every single deprecated methods of course, I just tried to replace the simple cases where a logically equivalent alternative was present.

Also I didn't consider some cases because #12569 is already covering them.

Bye!


jzheaux:
Thanks, @EvaristeGalois11! This is now merged into `main`.

PR #13411, opened 2023-06-22
fix String comparison

realmovestar:
It seems to me that the original author intended to return early without executing the `Utf8...` part if the values of two tokens are the same. However, even if the values are the same, the `Utf8...` part is executed because the `==` operator is used.


jzheaux:
Hi, @realmovestar, thanks for the suggestion. It's intended to not call `equals` since that itself exits early, potentially exposing applications to timing attacks. The intent of the method is to perform equals in a constant-time manner.

As such, I'm closing this PR to leave the method as-is. If it seems I've misunderstood your intent, please don't hesitate to reach out.

PR #13412, opened 2023-06-22
Assert is missing object. It was useless before Spring Framework 6.1 and will not compile on 6.1

cigaly:
Hope title explains all


jzheaux:
Thanks, @cigaly! This is now merged into `main`.

Issue #13366, opened 2023-06-16
Consider returning `this` from commonly used setter methods instead of `void`


ThomasKasene:
This is perhaps a little controversial, but I've sometimes wondered about the possibility of converting some of the more commonly used setters to return `this` rather than `void`.

Case in point, I recently wanted to be able to do something like this:
```java
@Bean
ReactiveJwtDecoder jwtDecoder(/* ... */) {
    return NimbusReactiveJwtDecoder
            .withIssuerLocation(...)
            .webClient(...)
            .jwtValidator(...)
            .build();
}
```
But I was thwarted by the limitations of the `JwkSetUriReactiveJwtDecoderBuilder` not allowing me to specify a `jwtValidator`, so I had to resort to an ugly detour via a variable:
```java
@Bean
ReactiveJwtDecoder jwtDecoder(/* ... */) {
    var jwtDecoder = NimbusReactiveJwtDecoder
            .withIssuerLocation(...)
            .webClient(...)
            .build();
    jwtDecoder.setJwtValidator(...);
    return jwtDecoder;
}
```

For a little while I was considering submitting a PR to add `jwtValidator(...)` to the builder, but soon realized that this class contains 4 such builders, and I'd have to duplicate the functionality between them all (and then double that for the non-reactive variant.)

An option, which I believe is becoming more and more common, is to enable developers to chain setters by returning `this`. If the `setJwtValidator(...)` method I used above returned the `NimbusReactiveJwtDecoder` instance, I could at the very least get a compromise going:
```java
@Bean
ReactiveJwtDecoder jwtDecoder(/* ... */) {
    return NimbusReactiveJwtDecoder
            .withIssuerLocation(...)
            .webClient(...)
            .build()
            .setJwtValidator(...);
}
```

It _should_ be backwards compatible, and as an added benefit, the validation that already exists within `setJwtValidator(...)` will be applied so we wouldn't have to duplicate that as well.

This is just one example, but it's a common theme perhaps especially in Spring Security.

Thanks for listening to my TED talk! 😄

marcusdacoregio:
Hi @ThomasKasene, I don't think returning `this` on setters is a pattern that Spring Security follows, that's why it has those `with...` methods.

> For a little while I was considering submitting a PR to add jwtValidator(...) to the builder, but soon realized that this class contains 4 such builders, and I'd have to duplicate the functionality between them all (and then double that for the non-reactive variant.)

[...]

PR #13423, opened 2023-06-23
Enabled all `*JwtDecoderBuilder`s to set a custom JWT validator

ThomasKasene:
This was discussed briefly as an example to make a point in #13366, but as per a suggestion there, I went ahead and added a `jwtValidator(...)` method to all the `*JwtDecoderBuilder`s to enable customizing the `JwtDecoder` and `ReactiveJwtDecoder` further without having to resort to awkward temporary variables:

```java
@Bean
ReactiveJwtDecoder jwtDecoder(/* ... */) {
    return NimbusReactiveJwtDecoder
            .withIssuerLocation(...)
            .webClient(...)
            .jwtValidator(...)
            .build();
}
```
as opposed to this:
```java
@Bean
ReactiveJwtDecoder jwtDecoder(/* ... */) {
    var jwtDecoder = NimbusReactiveJwtDecoder
            .withIssuerLocation(...)
            .webClient(...)
            .build();
    jwtDecoder.setJwtValidator(...);
    return jwtDecoder;
}
```

Disclaimer: I recognize that I was never given a "go ahead" in #13366 to do this, so I understand if this addition is not the desired direction for the project.

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoders.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoders.java
@@ -89,9 +89,8 @@ public final class JwtDecoders {
 	@SuppressWarnings("unchecked")
 	public static <T extends JwtDecoder> T fromIssuerLocation(String issuer) {
 		Assert.hasText(issuer, "issuer cannot be empty");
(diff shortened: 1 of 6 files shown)


ThomasKasene:
This is not quite in line with the team's wishes as it introduces a secondary way of accomplishing the same thing as the existing setter, which might be confusing to some according to [this comment](https://github.com/spring-projects/spring-security/issues/13366#issuecomment-1619251100).

I'll keep the branch though, and maybe submit something for Spring Security 7.

PR #13428, opened 2023-06-26
Allow JWT issuer AuthenticationManagerResolvers to accept predicate issuer validator

### LINKED ISSUE #13427: Create JwtIssuerReactiveAuthenticationManagerResolver with predicate
**Expected Behavior**

`JwtIssuerReactiveAuthenticationManagerResolver` internally uses `TrustedIssuerJwtAuthenticationManagerResolver`, which does a predicate test on JWT issuer. `JwtIssuerReactiveAuthenticationManagerResolver` constructors only allow collection/array of issuer URLs, and `TrustedIssuerJwtAuthenticationManagerResolver` is not public. There should be another constructor that accepts a `Predicate<String>` (which gets delegated to `TrustedIssuerJwtAuthenticationManagerResolver`), or `TrustedIssuerJwtAuthenticationManagerResolver` should be made public.

**Current Behavior**

`JwtIssuerReactiveAuthenticationManagerResolver` only allows collection/array of issuer URLs. It should also generalize to allow any predicate on issuer.

**Context**

This helps simplify the usage of Spring Security in multi-tenancy context. As a developer, I only care to write code that validates the issuer URL; the underlying caching/lazy-loading of `ReactiveJwtDecoder` should be managed by Spring, which is already done by `TrustedIssuerJwtAuthenticationManagerResolver`.

singhbaljit:
Close #13427

* Make `TrustedIssuerJwtAuthenticationManagerResolver` public so constructors to validate issuer with a predicate can be used.
* Makes the servlet and reactive versions a little more consistent (i.e. logging)
* Use `Set` instead of `ArrayList`; it has O(1) time complexity for look ups.

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
@@ -227,8 +227,7 @@ public final class JwtIssuerAuthenticationManagerResolver implements Authenticat
 				return authenticationManager;
 			}
 			else {
-				this.logger.debug(LogMessage
-						.format("Did not resolve AuthenticationManager since issuer is not trusted", issuer));
+				this.logger.debug("Did not resolve AuthenticationManager since issuer is not trusted");
 			}
 			return null;
 		}
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
@@ -91,6 +91,7 @@ public final class JwtIssuerAuthenticationManagerResolver implements Authenticat
 	 * Construct a {@link JwtIssuerAuthenticationManagerResolver} using the provided
 	 * parameters
 	 * @param trustedIssuers an array of trusted issuers
+	 * @since 6.2
 	 */
 	public static JwtIssuerAuthenticationManagerResolver fromTrustedIssuers(String... trustedIssuers) {
 		return fromTrustedIssuers(Set.of(trustedIssuers));
@@ -100,6 +101,7 @@ public final class JwtIssuerAuthenticationManagerResolver implements Authenticat
 	 * Construct a {@link JwtIssuerAuthenticationManagerResolver} using the provided
 	 * parameters
 	 * @param trustedIssuers a collection of trusted issuers
+	 * @since 6.2
 	 */
 	public static JwtIssuerAuthenticationManagerResolver fromTrustedIssuers(Collection<String> trustedIssuers) {
 		Assert.notEmpty(trustedIssuers, "trustedIssuers cannot be empty");
@@ -110,6 +112,7 @@ public final class JwtIssuerAuthenticationManagerResolver implements Authenticat
 	 * Construct a {@link JwtIssuerAuthenticationManagerResolver} using the provided
 	 * parameters
 	 * @param trustedIssuers a predicate to validate issuers
+	 * @since 6.2
 	 */
 	public static JwtIssuerAuthenticationManagerResolver fromTrustedIssuers(Predicate<String> trustedIssuers) {
 		Assert.notNull(trustedIssuers, "trustedIssuers cannot be null");
@@ -225,7 +228,7 @@ public final class JwtIssuerAuthenticationManagerResolver implements Authenticat
 			}
 			else {
 				this.logger.debug(LogMessage
-						.format("Did not resolve AuthenticationManager since issuer '%s' is not trusted", issuer));
+						.format("Did not resolve AuthenticationManager since issuer is not trusted", issuer));
 			}
 			return null;
 		}
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
@@ -67,7 +67,9 @@ public final class JwtIssuerAuthenticationManagerResolver implements Authenticat
 	 * Construct a {@link JwtIssuerAuthenticationManagerResolver} using the provided
 	 * parameters
 	 * @param trustedIssuers an array of trusted issuers
+	 * @deprecated use {@link #fromTrustedIssuers(String...)}
 	 */
+	@Deprecated(since = "6.2", forRemoval = true)
 	public JwtIssuerAuthenticationManagerResolver(String... trustedIssuers) {
 		this(Set.of(trustedIssuers));
 	}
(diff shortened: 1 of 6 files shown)


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
```diff
@@ -67,31 +66,34 @@ public final class JwtIssuerAuthenticationManagerResolver implements Authenticat
...
 	 */
 	public JwtIssuerAuthenticationManagerResolver(Collection<String> trustedIssuers) {
 		Assert.notEmpty(trustedIssuers, "trustedIssuers cannot be empty");
 		this.authenticationManager = new ResolvingAuthenticationManager(
 				new TrustedIssuerJwtAuthenticationManagerResolver(
-						Collections.unmodifiableCollection(trustedIssuers)::contains));
+						// Set#copyOf prevents null elements, removes duplicates
+						// Set#copyOf is a no-op if already unmodifiable, see the varargs constructor
+						// Set#contains also has O(1) time complexity
+						Set.copyOf(trustedIssuers)::contains));
```
jzheaux, line 87: Thanks for this cleanup. Will you please place the Set optimizations in a separate commit so that it can be backported to earlier versions?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
```diff
@@ -67,31 +66,34 @@ public final class JwtIssuerAuthenticationManagerResolver implements Authenticat
 	/**
 	 * Construct a {@link JwtIssuerAuthenticationManagerResolver} using the provided
 	 * parameters
-	 * @param trustedIssuers a list of trusted issuers
+	 * @param trustedIssuers an array of trusted issuers
```
jzheaux, line 69: Thanks for this cleanup. Will you please place the JavaDoc fixes in a separate commit so that it can be backported to earlier versions?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
```diff
@@ -67,31 +66,34 @@ public final class JwtIssuerAuthenticationManagerResolver implements Authenticat
...
 	 * parameters
-	 * @param trustedIssuers a list of trusted issuers
+	 * @param trustedIssuers a collection of trusted issuers
 	 */
 	public JwtIssuerAuthenticationManagerResolver(Collection<String> trustedIssuers) {
 		Assert.notEmpty(trustedIssuers, "trustedIssuers cannot be empty");
 		this.authenticationManager = new ResolvingAuthenticationManager(
 				new TrustedIssuerJwtAuthenticationManagerResolver(
-						Collections.unmodifiableCollection(trustedIssuers)::contains));
+						// Set#copyOf prevents null elements, removes duplicates
```
jzheaux, line 84: Thanks for the comments, please feel free to remove them at this point.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
```diff
@@ -167,15 +170,25 @@ public String convert(@NonNull BearerTokenAuthenticationToken authentication) {
...
 		private final Map<String, AuthenticationManager> authenticationManagers = new ConcurrentHashMap<>();
 
 		private final Predicate<String> trustedIssuer;
 
-		TrustedIssuerJwtAuthenticationManagerResolver(Predicate<String> trustedIssuer) {
+		/**
+		 * Constructs a {@link TrustedIssuerJwtAuthenticationManagerResolver} instance using the predicate
+		 * @param trustedIssuer a predicate to validate issuer
+		 */
+		public TrustedIssuerJwtAuthenticationManagerResolver(Predicate<String> trustedIssuer) {
```
jzheaux, line 190: I prefer the other idea that you listed of adding it as a top-level constructor. That allows this class to remain private for now.

jzheaux (changes requested):
Thanks, @singhbaljit! First, please see my inline feedback.

Second, will you please add some tests of the new functionality?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
singhbaljit: Done

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
singhbaljit: Done

singhbaljit:
#### re: tests
I'm honestly not sure what new tests to add. The existing tests already cover the functionality. I'm just exposing it publicly.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
jzheaux: What about instead deprecating the trusted-issuer constructors and changing to static factory methods? I think this is nice in that it addressed the backward compatibility concern as well as aligns more closely with related components like `NimbusJwtDecoder`.

I think the method `fromTrustedIssuers` makes sense:

```
AuthenticationManager#fromTrustedIssuers(String...)
AuthenticationManager#fromTrustedIssuers(Collection<String>)
AuthenticationManager#fromTrustedIssuers(Predicate<String>)
```

singhbaljit: That's a great idea! I wasn't sure there was an appetite for such a change. Do you want to cherry-pick the other 3 commits? I'll refactor the last commit with the proposed static factory methods.

singhbaljit: @jzheaux I have made the requested changes. Please have another look.

jzheaux (approved):

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtIssuerAuthenticationManagerResolver.java
```diff
@@ -227,8 +227,8 @@ public AuthenticationManager resolve(String issuer) {
 				return authenticationManager;
 			}
 			else {
-				this.logger.debug(LogMessage
-						.format("Did not resolve AuthenticationManager since issuer is not trusted", issuer));
+				this.logger.debug(
+						LogMessage.format("Did not resolve AuthenticationManager since issuer is not trusted", issuer));
```
singhbaljit, line 231: `LogMessage#format` is not needed since there's nothing for format anymore.

singhbaljit:
Also, you probably want to fixup the `Remove Logging of Untrusted Data` commit into `make logging consistent for Jwt issuer resolvers` commit if you want to back port those changes.

jzheaux:
Thanks, @singhbaljit! This is now merged into `main`.

jzheaux:
Also, the documentation was updated in 3b7e7f63d81396dc441f5dac60c569b8910f4ac3

PR #13440, opened 2023-06-30
Return content-type from saved request

repolevedavaj:
We detected an issue in the process of handling the original request, after redirecting the Browser to the authenticated entry point: As the `SavedRequestAwareWrapper` is not exposing the content-type of the original request, Spring cannot resolve the handler method correctly, if it specifies a `consumes` parameter (see attached example: [post-redirect.zip](https://github.com/spring-projects/spring-security/files/11915298/post-redirect.zip)).

```
@PostMapping(path = "example-post", consumes = {
        MediaType.APPLICATION_FORM_URLENCODED_VALUE,
        MediaType.ALL_VALUE // only needed if browser has been redirected to the authentication entry point
})
public String exampleHandler() {
    ...
}
```


repolevedavaj:
Is there any update on this PR?

jzheaux:
Thanks, @repolevedavaj! This is now merged into `main`.

PR #13458, opened 2023-07-07
Fix Adapeter to Adapter

kawin-nai:
Changed WebSecurityConfigurerAdapeter to WebSecurityConfigurerAdapter in Spring docs 5.8.x


jzheaux:
Thanks, @kawin-nai! This is now merged into `5.8.x`.

Issue #13468, opened 2023-07-08
Wrongly placed source and target arguments while calling ProviderManager#copyDetails(target, source)


mainul35:
**Describe the bug**
In the class `org.springframework.security.authentication.ProviderManager`, the parameters of method `copyDetails(target, source)` is wrongly oriented at line 184.

**To Reproduce**
I was debugging the code to find out how I can fetch the refresh token, and meanwhile found this issue.

**Expected behavior**
The correct code will be `copyDetails(result, authentication)` at line 184 of class `org.springframework.security.authentication.ProviderManager`

jzheaux:
Thanks for reaching out @mainul35. The current orientation is the correct one.

The idea is that there may be request details that need to be passed into the result. after authentication is complete. If the result details are already populated, the request details aren't copied.

Also, `result` is what is returned from the method so I don't see a lot of value copying the other direction.

If you feel I'm mistaken, please feel free to add further comments, and we can continue taking a look!

mainul35:
Hello @jzheaux, I want to know 2 things.

1. May I know, for which authentication provider implementation class I may expect to have the AuthenticationDetails not set in the result object? I could only checkby debugging the OidcAuthorizationCodeAuthenticationProvider class and I found, either the Authentication object will be returned null or must throw an exception or must contain the AuthenticationDetails along with other information. So, for this case the `copyDetails(authentication, result)` method seems no longer necessary.

2. For other AuthenticationProvider classes, I went through the codes and found similar logics - either, returning null for Authentication Object or the same object which was passed through the parameter is returned or exception thrown. If none of these happened and a new Authentication object is created then it was populated with the AuthenticationDetails object from the source object already in the authenticate(...) method. In this case, isn't this `copyDetails(authentication, result)` method doing redundant work?

PR #13469, opened 2023-07-08
Bugfix/copy source object to target

mainul35:
In this PR I changed the argument positions. Previously the source object was passed in the place of target and target object was passed in the place or source argument. 

Issue reference: gh-13468


mainul35:
Hello @jzheaux @lmartelli @cigaly @marcusdacoregio @EvaristeGalois11 , can someone please review my code? Thank you.

PR #13472, opened 2023-07-09
Refactor for readability

limvik:
I refactored it like this

1. Using `putIfAbsent` method instead of `containsKey` and `put` method of the `Map`.
2. Using pattern matching for `instanceof`

I hope this helps improve readability.


jzheaux:
Thanks, @limvik! This is now merged into `6.1.x`, and `main`.

Issue #13310, opened 2023-06-13
ArrayIndexOutOfBoundsException in XorCsrfTokenRequestAttributeHandler


mhankus:
java.lang.ArrayIndexOutOfBoundsException is thrown in XorCsrfTokenRequestAttributeHandler during attack
Affects version spring-security 6.0.3 (I have not tested 6.1)

```
java.lang.ArrayIndexOutOfBoundsException: arraycopy: last destination index 36 out of bounds for byte[8]
        at org.springframework.security.web.csrf.XorCsrfTokenRequestAttributeHandler.xorCsrf(XorCsrfTokenRequestAttributeHandler.java:119) ~[spring-security-web-6.0.3.jar!/:6.0.3]
        at org.springframework.security.web.csrf.XorCsrfTokenRequestAttributeHandler.getTokenValue(XorCsrfTokenRequestAttributeHandler.java:99) ~[spring-security-web-6.0.3.jar!/:6.0.3]
        at org.springframework.security.web.csrf.XorCsrfTokenRequestAttributeHandler.resolveCsrfTokenValue(XorCsrfTokenRequestAttributeHandler.java:73) ~[spring-security-web-6.0.3.jar!/:6.0.3]
        at org.springframework.security.web.csrf.CsrfFilter.doFilterInternal(CsrfFilter.java:121) ~[spring-security-web-6.0.3.jar!/:6.0.3]
        at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:116) ~[spring-web-6.0.9.jar!/:6.0.9]
        at org.springframework.security.web.ObservationFilterChainDecorator$ObservationFilter.wrapFilter(ObservationFilterChainDecorator.java:185) ~[spring-security-web-6.0.3.jar!/:6.0.3]
        at org.springframework.security.web.ObservationFilterChainDecorator$ObservationFilter.doFilter(ObservationFilterChainDecorator.java:172) ~[spring-security-web-6.0.3.jar!/:6.0.3]
        at org.springframework.security.web.ObservationFilterChainDecorator$VirtualFilterChain.doFilter(ObservationFilterChainDecorator.java:133) ~[spring-security-web-6.0.3.jar!/:6.0.3]
        at org.springframework.security.web.header.HeaderWriterFilter.doHeadersAfter(HeaderWriterFilter.java:90) ~[spring-security-web-6.0.3.jar!/:6.0.3]
[...]

PR #13550, opened 2023-07-17
Fixing ArrayIndexOutOfBoundsException in XorCsrfTokenRequestAttribute…

kevin2jordan:
## Issue Description: https://github.com/spring-projects/spring-security/issues/13310
ArrayIndexOutOfBoundsException in XorCsrfTokenRequestAttributeHandler

SecurityIssue

## Steps to reproduce the issue

1. Modify the CSRF token value from the client side.
2. Submit the request with the modified token value.

## Expected behaviour:
The underlying method getTokenValue should validate the encoded token length and return null if the value is incorrect. Generating stack traces for exceptions is much more expensive and may impact performance.

## Change Details:

1. Handle the ArrayIndexOutOfBoundsException.
2. Added the test case.


jzheaux:
Thanks, @kevin2jordan! This is now merged into `main`.

PR #13553, opened 2023-07-17
Set NullRequestCache when session management disabled

dukcode:
```java
  @Bean
  public SecurityFilterChain securityFilterChain(HttpSecurity http) throws Exception {

    http
        .authorizeHttpRequests(request -> request.anyRequest().authenticated())
        .sessionManagement(s -> s.sessionCreationPolicy(SessionCreationPolicy.STATELESS));

    return http.build();
  }

```

When I configure `sessionManagement` to `STATELESS` like upper code, spring-security sets`RequestCache` of `ExceptionTranslationFilter` to `NullRequestCache` like next image.

image

```java
  @Bean
  public SecurityFilterChain securityFilterChain(HttpSecurity http) throws Exception {

    http
        .authorizeHttpRequests(request -> request.anyRequest().authenticated())
        .sessionManagement(AbstractHttpConfigurer::disable);

    return http.build();
  }

```
But when I configure `sessionManagement` disabled, `RequestCache` of `ExceptionTranslationFilter` is `HttpSessionRequestCache`. Please see image and output from Spring Security debug mode.

image

image

I think that `RequestCache` should be `NullRequestCache`.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/SessionManagementConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/SessionManagementConfigurer.java
@@ -358,6 +358,12 @@ public final class SessionManagementConfigurer<H extends HttpSecurityBuilder<H>>
 	}
 
 	@Override
+	public H disable() {
+		getBuilder().setSharedObject(RequestCache.class, new NullRequestCache());
+		return super.disable();
+	}
+
+	@Override
 	public void init(H http) {
 		SecurityContextRepository securityContextRepository = http.getSharedObject(SecurityContextRepository.class);
 		boolean stateless = isStateless();


marcusdacoregio:
Hi @dukcode, have you discussed this with the team before submitting the PR? Asking because I do not think that we want to do this since `sessionManagement()` is really only about how the session is handled around authentication. If you disable it, it won't make any configuration, and, therefore, will rely on the other defaults, like the `requestCache()`.

dukcode:
Hi @marcusdacoregio!

In `ExceptionTranslationFilter`, it calls `requestCache.saveRequest(request, response)`

When we set `sessionManagement()` to `disable`, `ExceptionTranslationFilter` make exception because `RequestCache` is set to `HttpSessionRequestCache`.

However, when we set `sessionManagement()` to `STATELESS`, `RequestCache` is set to `NullRequestCache`. So, there is no exception.

So I think `RequestCache` should be `NullRequestCache` when we set `sessionManagement()` to disable like `STATELESS`.

marcusdacoregio:
> When we set sessionManagement() to disable, in ExceptionTranslationFilter http session is created because RequestCache is set to HttpSessionRequestCache.

This is expected based on what I mentioned before, you are disabling session management at all but are keeping the defaults of the other configurations. If you do not want to have sessions at all, just set it to `STATELESS`, otherwise, if you are disabling it, you must also configure the `RequestCache` to use `NullRequestCache`.

dc-oe:
We are seeing the requestcache draining our form data when looking for continue. It looks similar to what issue 13731 is seeing. Our current solution is to disable the optimization and get it back to 5.8 behavior. You can't drain the input stream looking for something that isn't there. We would like to know if we can fix this through configuration. We use XML and not annotations.

PR #13603, opened 2023-08-01
Improve RequestMatcher Validation

nucleussoftwareopen:
Improve RequestMatcher Validation


marcusdacoregio:
Hi, @nucleussoftwareopen. It is always better to open an issue to discuss and have some guidance before opening a PR.
Moreover, we are not accepting commits to the `5.4.x` since it [reached its end of life](https://spring.io/projects/spring-security#support).

PR #13605, opened 2023-08-01
Fix typo in docs

bigfang:
Since Gradle 7.x, the `compile` configuration is deprecated, use `implementation` instead.


jzheaux:
Thanks, @bigfang! This is now merged into `6.0.x`, `6.1.x`, and `main`.

PR #13614, opened 2023-08-04
Improve StrictHttpFirewall error messaging

bjornharvold:
Better error strings for invalid header and parameter values.

This PR was created because we have had an outstanding ticket with Cloudflare (https://support.cloudflare.com/hc/en-us/requests/2858014?page=1) for over a month where we have been trying to get to the bottom of what / who is sending an illegal header value. because StrictHttpFirewall doesn't tell us what the header name is, it's been a guessing game.

Work done: Updated exception strings to include the header / parameter name for when the value is invalid.


jzheaux:
Thanks, @bjornharvold! This is now merged into `main`.

PR #13596, opened 2023-07-28
Add return statement of the roleHierachy method in the servlet/author…

galmegiz:
Add return statement in the Reference

this closes #13595

PR #13618, opened 2023-08-06
docs : Add return value

merong245:
In the example code, the method does not have a return value.
It's a small issue, but needs fixing.


jzheaux:
Thanks, @merong245! This has already been recently fixed in https://github.com/spring-projects/spring-security/pull/13596

PR #13619, opened 2023-08-06
Adopt dedicated AssertJ assertions for more expressive test failure messages

timtebeek:
Noticed that there were some instances where AssertJ wasn't used to it's full expressive potential, so I ran our [AssertJ best practices recipe](https://docs.openrewrite.org/recipes/java/testing/assertj/assertj) ([via](https://app.moderne.io/recipes/builder/bGVuS?organizationId=RGVmYXVsdA%3D%3D)) to adopt dedicated assertions. This should help the message when an assertion fails, such that it's not "expected true but was false", but instead something like "expected collection to contain "apple", but was ["banana"]".
```diff
-assertThat(objectIdentities.size()).isEqualTo(1);
+assertThat(objectIdentities).hasSize(1);

-assertThat(objects.contains("apple")).isTrue();
+assertThat(objects).contains("apple");
```
These are mostly automated changes with some manual fixes; I recommend a squash merge if possible. I hope this is appreciated! :)

### DIFF
--- a/acl/src/test/java/org/springframework/security/acls/jdbc/JdbcAclServiceTests.java
+++ b/acl/src/test/java/org/springframework/security/acls/jdbc/JdbcAclServiceTests.java
@@ -112,7 +112,7 @@ public class JdbcAclServiceTests {
 		given(this.jdbcOperations.query(anyString(), eq(args), any(RowMapper.class))).willReturn(result);
 		ObjectIdentity objectIdentity = new ObjectIdentityImpl(MockLongIdDomainObject.class, 1L);
 		List<ObjectIdentity> objectIdentities = this.aclService.findChildren(objectIdentity);
-		assertThat(objectIdentities.size()).isEqualTo(1);
+		assertThat(objectIdentities).hasSize(1);
 		assertThat(objectIdentities.get(0).getIdentifier()).isEqualTo("5577");
 	}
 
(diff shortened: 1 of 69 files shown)


oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/jackson2/OAuth2AuthenticationTokenMixinTests.java
```diff
@@ -139,7 +139,7 @@ public void deserializeWhenMixinRegisteredThenDeserializes() throws Exception {
 				.isEqualTo(expectedAuthentication.getAuthorizedClientRegistrationId());
 		DefaultOAuth2User expectedOauth2User = (DefaultOAuth2User) expectedAuthentication.getPrincipal();
 		DefaultOAuth2User oauth2User = (DefaultOAuth2User) authentication.getPrincipal();
-		assertThat(oauth2User.getAuthorities().containsAll(expectedOauth2User.getAuthorities())).isTrue();
+		assertThat(oidcUser.getAuthorities().containsAll(expectedOidcUser.getAuthorities())).isTrue();
```
jzheaux, line 142: Why is this change needed?

jzheaux (changes requested):
Thanks, @timtebeek! This is very helpful.

I've left a comment inline about one of the changes.

Also, if you wish at this point, you are welcome to squash your commits to a single commit.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/jackson2/OAuth2AuthenticationTokenMixinTests.java
timtebeek: Apologies! I had manually reverted an automated change there, but made a mistake in copying the before & after lines. This has since been corrected in 650a2fc. Thanks for noticing! Goes to show I shouldn't make manual changes and fix the automation & run again. (:

timtebeek:
Glad to hear you think it's helpful! I've fixed my manual mistake and think we should be good for a (squash) merge.

I'd noticed [another small possible improvement](https://github.com/openrewrite/rewrite-testing-frameworks/issues/400), which is why it's fun for me to do these PRs; can circle back on that in the future if that's appreciated; should make for a smaller PR by then.

timtebeek:
Anything else you'd like to see before a squash merge @jzheaux ? I'd hope to get this in before there's merge conflicts if at all possible, but understand this is far from a critical change.

timtebeek:
Gone ahead and merged the followup commits into the first one such that it's now just a single commit; hope that helps!

jzheaux:
Thanks, @timtebeek! This is now merged into `main`.

PR #13621, opened 2023-08-07
add java doc warning comment into ipAddressMatcher for potential DNS resolution

maoling:
- In our use case,  our api gateway used `ipAddressMatcher` util to check whether a request's ip is in the ip-black list, but sometimes, the `remoteAddress` or `X-Forwarded-For` is in a bad format(maybe malicious requests) as following.  They're not Ip, but hostname. ipAddressMatcher calls the  `InetAddress#getByNam`e  which causes a DNS resolution, then the cloud vm alarms with this behavior.
  > remoteAddress:cj3gg4k0jvnb2t86jq6gi8twr36r38hzd.oast.fun:44194, X-Forwarded
  -For:"cj3gg4k0jvnb2t86jq6gi8twr36r38hzd.oast.fun, 1.2.3.4, 1.4.5.6"

- This PR just adds java doc warning comment, If your guys prefer the implementation by removing using the `InetAddress#getByName` and check ipv4/ipv6 inside, I will create another PR to rewrite it.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/util/matcher/IpAddressMatcher.java
+++ b/web/src/main/java/org/springframework/security/web/util/matcher/IpAddressMatcher.java
@@ -44,7 +44,7 @@ public final class IpAddressMatcher implements RequestMatcher {
 	 * Takes a specific IP address or a range specified using the IP/Netmask (e.g.
 	 * 192.168.1.0/24 or 202.24.0.0/14).
 	 * @param ipAddress the address or range of addresses from which the request must
-	 * come.
+	 * come. Note: ipAddress should not be a hostname to avoid the DNS resolution for potential security issue
 	 */
 	public IpAddressMatcher(String ipAddress) {
 		if (ipAddress.indexOf('/') > 0) {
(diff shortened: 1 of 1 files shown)


jzheaux (changes requested):
I don't think the new JavaDoc adds anything that wasn't already said earlier. The existing JavaDoc says:

```java
* Takes a specific IP address or a range specified using the IP/Netmask (e.g.
* 192.168.1.0/24 or 202.24.0.0/14).
```

so the correct format is already indicated.

That said, I like your second idea to validate the parameters. Will you please change the code to validate that the first character is `[`, `:`, or a digit? Something like this:

```java
Assert.isTrue(ipAddress.charAt(0) == `[` || 
    ipAddress.charAt(0) == ':' || 
    Character.digit(ipAddress.charAt(0), 16), "ipAddress must start with a [, :, or a hexadecimal digit");
```

This is needed in the constructor as well as the `matches(String)` method when preparing the `address` for comparison.

jzheaux:
Hi, @maoling. Are you able to make the requested changes?

FdHerrera:
Hi guys, hope you're doing well

Since no updates has been made in this pull request, I went ahead and created [this PR](https://github.com/spring-projects/spring-security/pull/14491) to address the changes suggested.

Let me know if its ok for you

jzheaux:
Closing in favor of #14491

PR #13648, opened 2023-08-14
Add getters to `OAuth2AuthorizedClientId`

ch4mpy:
The current implementation of `OAuth2AuthorizedClientId` is enough to compare IDs, but I have a use case where I retrieve some if this IDs and need to access the registration ID, which is not possible right now.


willns1 (approved):

jzheaux (changes requested):
Thanks, @ch4mpy. WIll you do me a couple of favors here, please?

1. Update the copyright year
2. Add JavaDoc. I realize that getter JavaDoc is sometimes a little cookie cutter; in this case it at least gives a place for a `@since` attribute.

ch4mpy:
@jzheaux finally done

sjohnr:
Thanks @ch4mpy! This is now merged as b69939c1e55c9dd477e55e91f6d783ae720c18ec with polish commit 05d3beb6a845712a10a405655170007340ce2cc0.

PR #13649, opened 2023-08-14
Replace deprecated method

limvik:
Hello, I replaced `HttpMethod.resolve()` to `HttpMethod.valueOf()`

`HttpMethod.resolve()` is deprecated since Spring Framework 6.0.

Thank you.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/util/matcher/DispatcherTypeRequestMatcher.java
+++ b/web/src/main/java/org/springframework/security/web/util/matcher/DispatcherTypeRequestMatcher.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/springframework/security/web/util/matcher/DispatcherTypeRequestMatcher.java
```diff
@@ -66,7 +66,7 @@ public DispatcherTypeRequestMatcher(DispatcherType dispatcherType, @Nullable Htt
 	@Override
 	public boolean matches(HttpServletRequest request) {
 		if (this.httpMethod != null && StringUtils.hasText(request.getMethod())
-				&& this.httpMethod != HttpMethod.resolve(request.getMethod())) {
+				&& this.httpMethod != HttpMethod.valueOf(request.getMethod())) {
```
jesperronn, line 69: as I read the deprecated method:

https://github.com/spring-projects/spring-framework/blob/6.0.x/spring-web/src/main/java/org/springframework/http/HttpMethod.java#L138

it seems to have a null-check. 

I wonder if you also need a null check here (`request.getMethod()` == null)?

jesperronn:
looks good to me, but if request.getMethod() can be null it needs a null check? 

If request.getMethod() == null and no null check is introduced, then it will change behaviour slightly

```
# Before:
HttpMethod.resolve(null) => null

# After:
HttpMethod.valueOf(null) => new HttpMethod(null) 

```

I dont know if that is a significant change or it will ever make a difference.

web/src/main/java/org/springframework/security/web/util/matcher/DispatcherTypeRequestMatcher.java
jzheaux: It's not needed since that condition is checked on line 68 with `StringUtils.hasText(request.getMethod())`

PR #13664, opened 2023-08-18
refactor: code readability

YunByungil:
Wouldn't it be more intuitive to write it like this?
I wanted to participate in this contribution
Please think it's cute


marcusdacoregio:
Hi, @YunByungil, thanks for the contribution. However, I think that such a cosmetic change does not improve the readability of the method. I agree that if we use something like `assertThat(...).isEmpty()` reads better but, again, the places we are not using that are sporadic.

Issue #13715, opened 2023-08-19
Allow OpenSamlMetadataResolver to switch between pretty print vs normal displayed metadata rendering


adrianpoplesanu:
**Expected Behavior**

Allow OpenSamlMetadataResolver to generate metadata that is not pretty printed. This can be achieved by adding a flag that controls whether the outputted metadata string gets pretty printed or not. The default value can be set to true as to preserve the current default behaviour.

**Current Behavior**

Currently the metadata is always pretty printed through the use of `SerializeSupport.prettyPrintXML(element)` which has `format-pretty-print=true` statically set.

**Context**

Generating a signed metadata, that gets pretty printed after signing, leads to the metadata not being validated when the signature is checked afterwards.
2 possible workarounds that are not satisfactory and hard to maintain with library version updates:
1. code duplication, or 
2. reflection override of the static value in SerializeSupport

sumeetpri:
@adrianpoplesanu  How to generate signed metadata , I am have similar requirement but not able to generate signed metadata ? I have posted my question at https://stackoverflow.com/questions/76926392/adding-signature-to-service-provider-metadata-in-spring-security-saml

adrianpoplesanu:
@sumeetpri
I signed the metadata by using an entity descriptor customizer. I created a class that extends `Consumer<EntityDescriptorParameters>` that signs the entity descriptor within `accept` and passed it in `OpenSamlMetadataResolver` via `setEntityDescriptorCustomizer`. I believe you can use any signing implementation, however i copied(because of the visibility modifiers) the one from OpenSamlSigningUtils over in my project and used that.

PR #13716, opened 2023-08-19
Allow OpenSamlMetadataResolver to also render un-pretty printed metadata content

adrianpoplesanu:
closes gh-13715

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/metadata/OpenSamlMetadataResolver.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/metadata/OpenSamlMetadataResolver.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -72,6 +72,8 @@ public final class OpenSamlMetadataResolver implements Saml2MetadataResolver {
 	private Consumer<EntityDescriptorParameters> entityDescriptorCustomizer = (parameters) -> {
 	};
 
+	private boolean usePrettyPrint = true;
+
 	public OpenSamlMetadataResolver() {
 		this.entityDescriptorMarshaller = (EntityDescriptorMarshaller) XMLObjectProviderRegistrySupport
 			.getMarshallerFactory()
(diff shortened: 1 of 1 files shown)


saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/metadata/OpenSamlMetadataResolver.java
```diff
@@ -121,6 +123,10 @@ public void setEntityDescriptorCustomizer(Consumer<EntityDescriptorParameters> e
 		this.entityDescriptorCustomizer = entityDescriptorCustomizer;
 	}
 
+	public void setUseFormatPrettyPrintXML(boolean useFormatPrettyPrintXML) {
```
jzheaux, line 126: Would you please simplify this method to be called `setUsePrettyPrint`?

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/metadata/OpenSamlMetadataResolver.java
```diff
@@ -121,6 +123,10 @@ public void setEntityDescriptorCustomizer(Consumer<EntityDescriptorParameters> e
 		this.entityDescriptorCustomizer = entityDescriptorCustomizer;
 	}
```
jzheaux, line 125: Will you please add a brief JavaDoc? Something like this:

```java
/**
  * Configure whether to pretty-print the metadata XML. This can be
  * helpful when signing the metadata payload.
  * @since 6.2
  **/
```

jzheaux (changes requested):
Thanks, @adrianpoplesanu! I've left some feedback inline.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/metadata/OpenSamlMetadataResolver.java
adrianpoplesanu: yes, done, i also updated the flag's name for consistency, thank you!

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/metadata/OpenSamlMetadataResolver.java
adrianpoplesanu: yes, done, i added your suggestion, thank you!

jzheaux (approved):

jzheaux:
Thanks, @adrianpoplesanu! This is now merged into `main`.

PR #13752, opened 2023-08-30
Max Sessions on WebFlux

marcusdacoregio:
- https://github.com/spring-projects/spring-security/issues/13791

### DIFF
--- a/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
@@ -62,6 +62,7 @@ import org.springframework.security.core.Authentication;
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.authority.AuthorityUtils;
 import org.springframework.security.core.authority.mapping.GrantedAuthoritiesMapper;
+import org.springframework.security.core.session.ReactiveSessionRegistry;
 import org.springframework.security.core.userdetails.ReactiveUserDetailsService;
 import org.springframework.security.oauth2.client.InMemoryReactiveOAuth2AuthorizedClientService;
 import org.springframework.security.oauth2.client.ReactiveOAuth2AuthorizedClientService;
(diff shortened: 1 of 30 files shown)


maradanasai:
HI @marcusdacoregio Thanks for this. Can you please take it forward?

maradanasai:
Can't wait for this feature released in 6.20

marcusdacoregio:
Hi, @maradanasai. Unfortunately, this will not make it into the 6.2 release. I will try to get it added for the first milestone of 6.3 though.

web/src/main/java/org/springframework/security/web/server/authentication/session/PreventLoginMaximumSessionsExceededHandler.java
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.security.web.authentication.session.SessionAuthenticationException;
+
+/**
+ * Throws a {@link SessionAuthenticationException} when the maximum number of sessions for
+ * a user has been reached.
+ *
+ * @author Marcus da Coregio
+ * @since 6.3
+ */
+public final class PreventLoginMaximumSessionsExceededHandler implements ServerMaximumSessionsExceededHandler {
```
jzheaux, line 30: I'd recommend the word `Server` also be in the class name. The reason is if we were to add the same thing on the servlet side, we want the class names to be different for clarity.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -1907,6 +1957,234 @@ public AuthorizeExchangeSpec access(ReactiveAuthorizationManager<AuthorizationCo
...
+				return this;
+			}
+
+			/**
+			 * Sets the {@link Function} to use to determine the maximum number of
+			 * sessions for a user.
+			 * @param maxSessions the {@link Function} to use
+			 * @return the {@link ConcurrentSessionsSpec} to continue customizing
+			 */
+			public ConcurrentSessionsSpec maxSessions(Function<Authentication, Mono<Integer>> maxSessions) {
```
jzheaux, line 2050: Having a `Function` as an API can be a bit more challenging. Have you considered having two methods:

```java
.maxSessions(Function)

.maxSessions(Integer)
```

In this way, folks can supply a global max, should that be preferred, in a way that is clearly correct.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -1907,6 +1957,234 @@ public AuthorizeExchangeSpec access(ReactiveAuthorizationManager<AuthorizationCo
...
+			 * @param reactiveSessionRegistry the {@link ReactiveSessionRegistry} to use
+			 * @return the {@link ConcurrentSessionsSpec} to continue customizing
+			 */
+			public ConcurrentSessionsSpec sessionRegistry(ReactiveSessionRegistry reactiveSessionRegistry) {
+				SessionManagementSpec.this.sessionRegistry = reactiveSessionRegistry;
+				return this;
+			}
+
+			/**
+			 * Sets the {@link Function} to use to determine the maximum number of
```
jzheaux, line 2045: It's not clear to me how I'd state that for some authentications, there is no maximum. For example, is it:

```java
.maxSessions((authentication) -> containsSomeAuthority(authentication) ? Mono.just(Integer.MAX_VALUE) : Mono.just(3))
```

or something else?

(I can see in the reference that it is -1, but this may be value to place in the JavaDoc, too. Also, since it is a magic number, you might consider a static constant like `UNLIMITED`. This adds clarity to the user's intent and it also gives you the ability to change that value in the future should it become necessary.)

web/src/main/java/org/springframework/security/web/server/authentication/session/ServerSessionAuthenticationStrategy.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Allows pluggable support for WebSession-related behaviour when an authentication
+ * occurs.
+ *
+ * @author Marcus da Coregio
+ * @since 6.3
+ * @see AuthenticationWebFilter#setSessionAuthenticationStrategy(ServerSessionAuthenticationStrategy)
+ */
+public interface ServerSessionAuthenticationStrategy {
```
jzheaux, line 33: I wonder if a new interface is necessary in this case since we already have `ServerAuthenticationSuccessHandler`, whose method signature is the same.

It seems like the implementations could be added to a `DelegatingServerAuthenticationSuccessHandler` instance. Do you try that already?

jzheaux (changes requested):
This looks great, @marcusdacoregio. Per your request, I've reviewed the following:

* Interface naming; I also took the liberty to comment on some class names
* DSL names and contracts

web/src/main/java/org/springframework/security/web/server/authentication/session/ServerSessionAuthenticationStrategy.java
marcusdacoregio: Thanks! I've changed the implementation to use the `ServerAuthenticationSuccessHandler` instead of a new interface.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
marcusdacoregio: Instead of the user dealing with magic numbers, I think it would be better to define a `SessionLimit` POJO with a `SessionLimit#of(int)` and `SessionLimit#unlimited()` static factory methods to achieve a better declarative configuration.

In addition to that, there are now `maximumSessions(SessionLimit)` and `maximumSessions(Function<Authentication, Mono<SessionLimit>)` available.

PR #13769, opened 2023-09-01
Closes #13754 Add hasScope and hasAnyScope for authorization rules

### LINKED ISSUE #13754: Add hasScope for authorization rules
`AuthorizationManager` makes the following possible in the DSL:

```java
http
        .anyRequest().access(hasRole("USER"))
    )
```

It would be nice to be able to do the same with scopes, like so:

```java
http
        .anyRequest().access(hasScope("resource:read"))
    )
```

This could be done in a static factory class like `OAuth2AuthorizationManagers` in `oauth2-core`. It could have `hasScope` and `hasAnyScope`. They might be implemented by delegating to `AuthorityAuthorizationManager` like this:

```java
public static <T> AuthorityAuthorizationManager<T> hasScope(String scope) {
    return AuthorityAuthorizationManager.hasAuthority("SCOPE_" + scope);
}
```

petrovskimario:

### DIFF
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ScopeAuthorizationManagerFactory.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ScopeAuthorizationManagerFactory.java
@@ -33,7 +33,7 @@ public final class ScopeAuthorizationManagerFactory {
 	}
 
 	public static <T> AuthorityAuthorizationManager<T> hasAnyScope(String... scopes) {
-		String[] mappedScopes = Arrays.stream(scopes).map(s -> "SCOPE_" + s).toArray(String[]::new);
+		String[] mappedScopes = Arrays.stream(scopes).map((String s) -> "SCOPE_" + s).toArray(String[]::new);
 		return AuthorityAuthorizationManager.hasAnyAuthority(mappedScopes);
 	}
 }
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ScopeAuthorizationManagerFactory.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ScopeAuthorizationManagerFactory.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2002-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.springframework.security.oauth2.core;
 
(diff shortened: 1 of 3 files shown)


oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ScopeAuthorizationManagerFactory.java
```diff
@@ -0,0 +1,40 @@
...
+
+	private ScopeAuthorizationManagerFactory() {
+	}
+
+	public static <T> AuthorityAuthorizationManager<T> hasScope(String scope) {
+		return AuthorityAuthorizationManager.hasAuthority("SCOPE_" + scope);
+	}
+
+	public static <T> AuthorityAuthorizationManager<T> hasAnyScope(String... scopes) {
+		String[] mappedScopes = Arrays.stream(scopes).map((String s) -> "SCOPE_" + s).toArray(String[]::new);
```
jzheaux, line 36: Will you please first test that the scopes don't already begin with "SCOPE_"? If they do, this method should error.

jzheaux, line 36: For performance reasons, please [avoid stream usage in Spring Security](https://github.com/spring-projects/spring-security/issues/7154).

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ScopeAuthorizationManagerFactory.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package org.springframework.security.oauth2.core;
+
+import java.util.Arrays;
+
+import org.springframework.security.authorization.AuthorityAuthorizationManager;
+
+/**
+ * @author Mario Petrovski
```
jzheaux, line 24: Please add a `@since 6.2` annotation

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ScopeAuthorizationManagerFactory.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public final class ScopeAuthorizationManagerFactory {
```
jzheaux, line 26: Spring Security tends to name factories by pluralizing what is being created. Since this can create `AuthorizationManager`s, will you please call this `OAuth2AuthorizationManagers` instead?

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/ScopeAuthorizationManagerFactory.java
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright 2002-2018 the original author or authors.
```
jzheaux, line 2: Will you please set this to `2002-2023`?

jzheaux (commented):
Thanks, @petrovskimario! I've left some feedback inline.

Also, even though the functionality here is straightforward, will you please add a `Tests` class that confirms they work as expected? That way they can be changed in the future with confidence.

petrovskimario:
Hello @jzheaux 

Thank you for your suggestions.
I think i have completed it now.

Please review and let me know if something can be improved.

jzheaux:
Thanks, @petrovskimario! This is now merged into `main` in addition to a polish @ a3227f041ce405fa55d0feab078595fdca50c692 that I added.

PR #13772, opened 2023-09-03
Allow instantiation with a custom webclient

TuranDev:
This is to allow the passing in of a custom WebClient when instantiating any child of the class AbstractWebClientReactiveOAuth2AccessTokenResponseClient. This removes the overhead of having to first instantiate it with a default WebClient and then the setting of the required custom one.

It also lends itself well to spring boot to reduce the amount of equivalent vanilla instances of beans floating around when not needed.


sjohnr:
@TuranDev thanks for the PR!

However, I'm not sure we want to do this, or at least not without some discussion first. Please note that it's often best to open an issue to discuss changes such as this prior to submitting a PR.

Regarding this specific change, I understand you find it convenient to instantiate these classes by passing a custom `WebClient` at the same time (eliminating a single setter call). Yet there are other fields in the `abstract` class that others may feel are equally valuable to have in a constructor. I'm not sure how we would decide (without discussion and feedback with the community) which constructor arguments would make the most sense. It could also be argued (because of the variety of fields or for other reasons) that the setters are just fine and the need for convenience isn't that high. For that reason, I'm going to close this in favor of an issue to discuss and solicit feedback. Would you mind opening an issue instead?

Issue #13776, opened 2023-09-06
Saml-Metadata with special characters is corrupted


JannickWeisshaupt:
**Describe the bug**
Saml metadata is corrupted once it contains special characters.

**To Reproduce**

1. Configure the metadata to contain special characters, e.g. via OpenSamlMetadataResolver.EntityDescriptorParameters.
2. Generate the metadata via /saml2/service-provider-metadata/{descriptor}
3. The metadata is corrupted and invalid xml

**Expected behavior**
The metadata is not corrupted, valid xml and obeys the Saml Standard

**Cause**
In Saml2MetadataFilter Line 111
`response.setContentLength(metadata.getMetadata().length());`
is generally wrong as the length in bytes is required which differs from Java String length, e.g. "ä".length() = 1 but should be 2

PR #13777, opened 2023-09-06
fix corrupted saml2 metadata once special characters are present

JannickWeisshaupt:
Fixes corrupted Saml2Metadata once special characters are present. The cause of this bug is that the length in bytes, which is correct for the content length is not equal to the java string length.

See Issue #13776


marcusdacoregio:
Hi @JannickWeisshaupt, can you add a test that verifies the correct behavior? In addition to that, can you please target the `6.1.x` branch?

JannickWeisshaupt:
Hi @marcusdacoregio, should be done!

marcusdacoregio:
Nice @JannickWeisshaupt. In preparation for merging, can you please squash your commits into one? Ideally, the single commit should have the following message:

```
Fix corrupted saml2 metadata when special characters are present

Closes gh-13776
```

marcusdacoregio (approved):

JannickWeisshaupt:
Commits are squashed @marcusdacoregio

marcusdacoregio:
Thanks, @JannickWeisshaupt. This is now merged into `5.8.x`, `6.0.x`, `6.1.x` and `main`

Issue #13780, opened 2023-09-06
Cannot reproduce scrypt encoded password


ChrisCho-H:
`SCryptPasswordEncoder` provides `encode()` method but the salt is generated internally each time call the method.
To reproduce scrypt encoded password, it should also allow to pass salt in parameter.

it could be very useful in many ways.
For example,
1. user enters password which is gonna be used as a key to encrypt data
2. for security, encode password with scrypt
3. encrypted data is stored in any storage(and not encoded key)
4. to decrypt data, user enters password and then reproduce scrypt key to decrypt

marcusdacoregio:
Hi, @ChrisCho-H. The idea of the salt is really to be randomly generated each time you encode a password, so two same strings become different hashes.

ChrisCho-H:
I know. What I meant is to generate random byte outside of module and use that as a salt parameter, so that reproduce scrypt key.

marcusdacoregio:
I don't see how the question that you linked relates to what you are asking. If you want, you can specify the `saltGenerator` inside the SCrypt implementation.

PR #13781, opened 2023-09-06
feat: encode with salt param to reproduce scrypt key

ChrisCho-H:
`SCryptPasswordEncoder` provides `encode()` method but the salt is generated internally each time call the method.
To reproduce scrypt encoded password, it should also allow to pass salt in parameter.

just made overloading method to allow it.

[gh-13780](https://github.com/spring-projects/spring-security/issues/13780)


marcusdacoregio:
Closed via https://github.com/spring-projects/spring-security/issues/13780#issuecomment-1714554960

PR #13788, opened 2023-09-09
Add a factory method for RoleHierarchyImpl

making:
Currenctly Hierarchical Roles requires the setter method to define the hierarchy.

```java
RoleHierarchyImpl hierarchy = new RoleHierarchyImpl();
hierarchy.setHierarchy("...");
```

This PR adds a factory method for this use case to make it easier

```java
RoleHierarchyImpl hierarchy = RoleHierarchyImpl.of("...");
```

### DIFF
--- a/core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImpl.java
+++ b/core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,7 +21,6 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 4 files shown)


jzheaux:
@making, I like this idea; however, I'd prefer that it not be on the implementation.

Usually, this is addressed in Spring Security with a factory class like `RoleHierarchies`. The nice thing about that is that we could do the following:

```java
public final class RoleHierarchies {
    public static RoleHierarchy of(String hierarchy) { ... }
    public static RoleHierarchy of(Map<String, List<String>> hierarchy) { ... }
}
```

Are you able to change the PR along these lines?

making:
I’ll do it

making:
The reason I created this factory method in the impl class is because the format of the string representation is highly dependent on `RoleHierarchyImpl`.
If I don't have to care about it I'll move to RoleHierarchies.

making:
> `Map<String, List<String>> hierarchy`

Is this a map where directly assigned authority is the key and reachable authorities is the value?

jzheaux:
@making Good points.

> The reason I created this factory method in the impl class is because the format of the string representation is highly dependent on `RoleHierarchyImpl`.

Good point. Maybe instead what we do is deprecate the `setHierarchy` method as well as add the factory method. My concern is that it's confusing to have a setter and a constructor that do the same thing in the same class.

> And is this intended to be a factory method for a different implementation class than `RoleHierarchyImpl`?

No, I was thinking for the same implementation. It could use `RoleHierarchyUtils` to adapt, for example.

That said, a builder is probably better than taking a map, so instead let's just add the factory method to `RoleHierarchyImpl`, as you originally proposed, and also deprecate `setHierarchy`. I'll create separate ticket to look at a more fluent way to describe a hierarchy. Will that work?

jzheaux:
Thanks again for your submission, @making! This is now merged into `main`. I added a polish in https://github.com/spring-projects/spring-security/pull/13788/commits/9ffa54d20819724aae916fd51fee576784074ecc.

PR #13914, opened 2023-10-02
Expaned visibility in DefaultServerOAuth2AuthorizationRequestResolver…

Tr1ple-F:
Why should you do this:

I have the case where the consumer (i.e. the frontend asking for a OAuth2 redirect) needs to change the redirect uri on the fly. For that purpose we added a whitelist of permutations which the frontend can send in the request header to manually specify which "sub-site" it wants to redirect to. To avoid code duplication it would be better if one could just extend the "expandRedirectUri" Method and provide their custom implementation

@rwinch I think a lot of your classes are very useful and our project does a lot of customization on top of the standard OAuth2 mechanism. I would be willing do flesh out the OAuth2/OIDC Spring Documentation in exchange for visibility/scope changes

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
@@ -220,7 +220,7 @@ public final class ClientRegistration implements Serializable {
 
 		private Map<String, Object> configurationMetadata = Collections.emptyMap();
 
-		ProviderDetails() {
+		public ProviderDetails() {
 		}
 
 		/**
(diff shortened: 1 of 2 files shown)


Tr1ple-F:
To add regarding documentation - I think there should be more documentation regarding:
- The default implementation of CSRF with the BREACH protection
- Customizing redirect uri's
- General information regarding which classes do what part of the OAuth2 handling
- How to make good use of the OAuth2 redirect strategy

sjohnr:
Hi @Tr1ple-F, thanks for the PR! However, it's usually better to start with an issue so that we can discuss the direction we'd like to take something before writing code.

> To avoid code duplication it would be better if one could just extend the "expandRedirectUri" Method and provide their custom implementation

In this case (as in many cases), I think we would prefer not to simply open up visibility or encourage an inheritance-based approach, but instead lean towards composition. Or if necessary, we could add a utility method to share code amongst multiple implementations.

Since the `ServerOAuth2AuthorizationRequestResolver` (and corresponding non-reactive `OAuth2AuthorizationRequestResolver`) already allow easily replacing the mechanism for resolving an `OAuth2AuthorizationRequest`, I think it would be preferable to focus on making the expansion of URI variables reusable. See (for example) [`RelyingPartyRegistrationPlaceholderResolvers`](https://github.com/spring-projects/spring-security/blob/main/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/RelyingPartyRegistrationPlaceholderResolvers.java) for possible inspiration.

If we can refactor `expandRedirectUri()` into a reusable static method, you could easily implement your own `ServerOAuth2AuthorizationRequestResolver`. Would something like that achieve your goal?

Tr1ple-F:
Yes that helps me. Thank you!

sjohnr:
@Tr1ple-F are you planning on submitting a new PR? You are more than welcome to re-open and push new updates to this PR with an update. We can easily change the title of the PR to match.

PR #13969, opened 2023-10-06
Update http.adoc: IP number does not follow IP number format

GeniusDP:
ip number does follow the ip number format, let it be 192.168.0.107


marcusdacoregio:
Thanks, @GeniusDP. This is now merged into `main`.

Issue #13988, opened 2023-10-10
Allow one to customize the AuthenticationConverter in BasicAuthenticationFilter


mmoayyed:
**Expected Behavior**

Presently in version `6.2.0-M3`, there is no possibility to customize the `authenticationConverter` that is provided by `BasicAuthenticationFilter`, which means post-processing of the filter does not allow one to control the authentication conversion. Customizing the authentication converter allows one to decide where, when and how the filter should actually process a given request, i.e.:

```java
// A custom implementation of authenticationConverter can return null here
var authRequest = authenticationConverter.convert(request);
if (authRequest == null) {
    this.logger.trace("Did not process authentication request");
    chain.doFilter(request, response);
    return;
}
```

The proposal in summary is,

1. Provide a setter for **authenticationConverter** in BasicAuthenticationFilter
2. Similar to `OidcLogoutAuthenticationConverter`, allow one to customize the request matching functionality.

**Current Behavior**

Not possible to customize the authentication conversion process for this filter without reflection or a brand new filter. The main driver for this is to allow the filter match on certain requests, letting SS to handle those, while ignoring other (authenticated) requests and letting the app handle those.

**Context**

If HttpSecurity is configured for basic-authentication, it is not possible (or at least seems this way) to decide when and for which requests the basic auth filter should execute. Compared with `OidcLogoutAuthenticationConverter` one is given a customizable request matcher. In contrast, the `BasicAuthenticationFilter`, matches on everything and anything that is able to produce the right kind of credentials. It seems impossible for the filter to back away, when a request contains credentials, allowing the app to handle that request. Perhaps that can be done using multiple filter chains, web customizers, etc all of which seem somewhat unnecessarily complex compared to the option here.

As ever, thank you!

mmoayyed:
PS I am happy to provide a PR here if you feel like this proposal might be appropriate to implement.

marcusdacoregio:
Hi, @mmoayyed. Thanks for the report.

I agree that the `AuthenticationConverter` should be customizable via a setter allowing the user to specify an `ObjectPostProcessor`.

Are you able to create a PR that targets the `main` branch and add the setter and related tests?
`6.2.0-RC1` will be released on Monday. Ideally, this PR should come at maximum on Wednesday so we have time to add it to the release. If not, we will only be able to add such support in `6.3`. If you do not have time to provide the PR in time, please let me know so I can work on it.

mmoayyed:
[...]

PR #13989, opened 2023-10-10
Allow AuthenticationConverter to be settable in BasicAuthenticationFilter

mmoayyed:
Closes https://github.com/spring-projects/spring-security/issues/13988

### DIFF
--- a/web/src/test/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilterTests.java
+++ b/web/src/test/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilterTests.java
@@ -21,6 +21,7 @@ import java.nio.charset.StandardCharsets;
 import jakarta.servlet.FilterChain;
 import jakarta.servlet.ServletRequest;
 import jakarta.servlet.ServletResponse;
+import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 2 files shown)


web/src/test/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilterTests.java
```diff
@@ -81,6 +82,15 @@ public void setUp() {
...
 		this.filter = new BasicAuthenticationFilter(this.manager, new BasicAuthenticationEntryPoint());
+		this.filter.setAuthenticationConverter(new BasicAuthenticationConverter() {
+			@Override
+			public UsernamePasswordAuthenticationToken convert(final HttpServletRequest request) {
+				if (request.getServletPath().equalsIgnoreCase("/ignored-request")) {
+					return null;
+				}
+				return super.convert(request);
+			}
+		});
```
marcusdacoregio, line 93: You can set the `AuthenticationConverter` into the filter in a specific test, there is no need to add it for every test method in `@BeforeEach`.

marcusdacoregio: In addition to that, we recommend using a delegate than extend the existing `BasicAuthenticationConverter`, something like:

```java
static class MyAuthenticationConverter implements AuthenticationConverter {

    private final BasicAuthenticationConverter delegate = new BasicAuthenticationConverter();

    @Override
	public UsernamePasswordAuthenticationToken convert(final HttpServletRequest request) {
		// check if request matches, maybe use RequestMatcher here
		return this.delegate.convert(request);
	}

}
```

Sometimes people refer to the tests as implementation examples, so it is better to align with what the team recommends and uses.

marcusdacoregio (changes requested):
Thanks @mmoayyed, I've left some feedback inline.

When you are done, can you please squash your commits? Thanks!

mmoayyed:
Thank you for the feedback. I will try to get to this tomorrow but if that is too late for the release you are more than welcome to take over from here or otherwise I was see what I can do in the coming days.

web/src/test/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilterTests.java
mmoayyed: A delegate is not possible here because the setter is:

```java
public void setAuthenticationConverter(BasicAuthenticationConverter authenticationConverter) {}
```

It's not possible to switch to:

```java
public void setAuthenticationConverter(AuthenticationConverter authenticationConverter) {}
```

...and change the field type to `AuthenticationConverter` because that breaks other things in the filter.

How would you like to proceed?

marcusdacoregio:
Thanks, @mmoayyed. I took over from where you left it and added the support into `main`.

mmoayyed:
Thank you very much!

PR #14060, opened 2023-10-29
Fix obsolete comment and typos

martin-lukas:
Found a weird reference to `BasicProcessingFilter` in the `BasicAuthenticationFilter` class. Seems to be a left-over after a renaming of the filter classes was done (commit 1286741c7c50fa7730e31852ea6fa4e03bc387ce).

While I was at it, fixed some more typos in comments in the core module.


marcusdacoregio:
Thanks, @martin-lukas. This is now merged.

PR #14090, opened 2023-11-02
fix #14089 Bugs occur when roles and authorities are built at the sam…

### LINKED ISSUE #14089: Bugs occur when roles and authorities are built at the same time
**Describe the bug**
Bugs occur when roles and authorities are built at the same time.
The content is overwritten by the latter

**To Reproduce**

when use code under line to build UserDetails, the value of authorities will be empty.

```java
           UserDetails userDetails = User.withDefaultPasswordEncoder()
                    .username("admin")
                    .password("123456")
                    .roles("admin")
                    .authorities("admin:get")
                    .build();
```

image

the reason is at line 444 under package org.springframework.security.core.userdetails and class User

image

**Expected behavior**

Since a method of continuous execution is provided, the results should be as expected.

so it should be :

```java
    this.authorities.addAll(authorities);
```

**Versions**
spring-security-core:6.1.5

Reports that include a sample will take priority over reports that do not.
At times, we may require a sample, so it is good to try and include a sample up front.

vanga-top:
…e time

### DIFF
--- a/core/src/main/java/org/springframework/security/core/userdetails/User.java
+++ b/core/src/main/java/org/springframework/security/core/userdetails/User.java
@@ -441,7 +441,7 @@ public class User implements UserDetails, CredentialsContainer {
 		 */
 		public UserBuilder authorities(Collection<? extends GrantedAuthority> authorities) {
 			Assert.notNull(authorities, "authorities cannot be null");
-			this.authorities = new ArrayList<>(authorities);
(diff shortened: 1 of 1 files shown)


sjohnr:
Closing per [comment](https://github.com/spring-projects/spring-security/issues/14089#issuecomment-1791006250).

PR #14124, opened 2023-11-09
Made JdbcUserDetailsManager implement UserDetailsPasswordService 

geirhe:
DaoAuthenticationProvider#130 tries to upgrade the encoding of a poorly encoded password, which seems to end up in limbo if JdbcUserDetailsManager does not implement UserDetailsPasswordService, patterned after InMemoryUserDetailsManager.

I noticed this when setting up a bare-bones spring authorization server project with persistence. My oauth clients would get updated passwords, my users would not.

I also had to tweak another test that started failing when implementing this interface, which makes me think this was how it was meant to be from the start, but then the actual implementation was just forgotten?


config/src/test/resources/CustomJdbcUserServiceSampleConfig.sql
```diff
@@ -1,4 +1,4 @@
-create table users(principal varchar_ignorecase(50) not null primary key,credentials varchar_ignorecase(50) not null);
+create table users(principal varchar_ignorecase(50) not null primary key,credentials varchar_ignorecase(100) not null);
```
geirhe, line 1: Increased size to allow the test to succeed - the encoded password is longer than 50 characters.

config/src/test/java/org/springframework/security/config/annotation/authentication/NamespaceJdbcUserServiceTests.java
```diff
@@ -114,6 +114,8 @@ void configure(AuthenticationManagerBuilder auth, DataSource dataSource) throws
 				.usersByUsernameQuery("select principal,credentials,true from users where principal = ?")
 				// jdbc-user-service@authorities-by-username-query
 				.authoritiesByUsernameQuery("select principal,role from roles where principal = ?")
+				// jdbc-user-service@change-password-query
```
geirhe, line 117: This test started failing when implementing the password update interface, which prompted me to tweak the test so the queries matched the schema.

geirhe:
Related issue/bug report: https://github.com/spring-projects/spring-security/issues/14126

Issue #14132, opened 2023-11-13
Remove deprecations from CookieCsrfTokenRepository


marcusdacoregio:
Remove methods that were deprecated as part of this PR https://github.com/spring-projects/spring-security/pull/12237

PR #14139, opened 2023-11-14
Remove deprecated methods from CookieServerCsrfTokenRepository

milaneuh:
From : https://github.com/spring-projects/spring-security/issues/14132 

Removing the Deprecated methods used in CookieServerCsrfTokenRepository.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/csrf/CookieCsrfTokenRepository.java
+++ b/web/src/main/java/org/springframework/security/web/csrf/CookieCsrfTokenRepository.java
@@ -169,14 +169,6 @@ public final class CookieCsrfTokenRepository implements CsrfTokenRepository {
 		this.cookieName = cookieName;
 	}
 
-	/**
-	 * @deprecated Use {@link #setCookieCustomizer(Consumer)} instead.
-	 */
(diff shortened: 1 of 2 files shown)


marcusdacoregio:
Hi, @milaneuh. Thanks for the PR.

Since this is only scheduled for 7.0, getting this merged will take a while. I think one feedback is that we also have to add documentation to help users migrate, and that documentation would be added to the [Preparing to 7.0](https://docs.spring.io/spring-security/reference/migration-7/index.html) section.

milaneuh:
> Hi, @milaneuh. Thanks for the PR.
> 
> Since this is only scheduled for 7.0, getting this merged will take a while. I think one feedback is that we also have to add documentation to help users migrate, and that documentation would be added to the [Preparing to 7.0](https://docs.spring.io/spring-security/reference/migration-7/index.html) section.

Hi, can I help in any way for the documentation ?

marcusdacoregio:
Yes, you can add a new `Web` section under **Preparing for 7.0** similar to the `Configuration` section. There you should guide users to avoid using the deprecated methods and, instead, use the customizer.

Issue #13773, opened 2023-09-04
UsernamePasswordAuthenticationFilter supports configuring AuthenticationManagerResolver


pop1213:
**Expected Behavior**

`UsernamePasswordAuthenticationFilter` supports configuring `AuthenticationManagerResolver`, just like the `BearerTokenAuthenticationFilter`.
**Current Behavior**

**Context**

I expect to dynamically retrieve the `AuthenticationManager`, for example, based on a parameter in the request. Let's consider a scenario where my user information is stored in different databases or tables. I want to obtain different AuthenticationManagers based on the request parameter, and then call different `UserDetailsService` implementations to retrieve user information.

jzheaux:
Thanks for the suggestion, @pop1213. Are you able to provide a PR to add that constructor? I believe it would be:

```java
public BasicAuthenticationFilter(AuthenticationManagerResolver authenticationManagerResolver,
    AuthenticationEntryPoint authenticationEntryPoint) { 
    // ...
}
```

PR #14150, opened 2023-11-16
UsernamePasswordAuthenticationFilter supports configuring Authenticat…

pop1213:
`UsernamePasswordAuthenticationFilter `supports configuring `AuthenticationManagerResolver`, just like the `BearerTokenAuthenticationFilter`.

Closes gh-13773


jzheaux:
Thanks for the PR, @pop1213! Before proceeding, I'd like to get on the same page about what you are trying to achieve and what you've already tried.

For example, have you already tried using the general-purpose `AuthenticationFilter`? It might be a better use of the PR to consider enhancing that to suit your needs instead of `UsernamePasswordAuthenticationFilter`.

PR #14169, opened 2023-11-19
Fix typo in the 'Authorizing Requests' example

aaron-to-go:
The example request matcher listed "/resources/**", while the explanation (3) used "/static/" instead.


marcusdacoregio:
Thanks, @aaron-to-go. This is now merged into `main`.

PR #14179, opened 2023-11-21
Fix: path in explanation to match path in example

cdenicola:
Updated explanation of "Authorizing Requests" example to match the path of the example for point 3 listing the URLs

Description
```
any user can access a request if the URL starts with "/static/", equals "/signup", or equals "/about".
```
now matches example
```
.requestMatchers("/static/**", "/signup", "/about").permitAll()
```


marcusdacoregio:
Thanks, @cdenicola. This is a duplicate of https://github.com/spring-projects/spring-security/pull/14169

PR #14183, opened 2023-11-22
Fix typo in getClaimAsMap docstring

ParkerM:
This fixes a typo in the docstring for `ClaimAccessor#getClaimAsMap` that mistakenly refers to the expected type as a List instead of a Map.


marcusdacoregio:
Thanks, @ParkerM. This is now merged into `main`.

Issue #13730, opened 2023-08-22
OAuth2 Resource Server is exposing server information.


akamom:
**Given** I created a custom JWT decoder as described in the [Spring OAuth2 Resource Server Documentation](https://docs.spring.io/spring-security/reference/servlet/oauth2/resource-server/jwt.html#oauth2resourceserver-jwt-decoder-builder) with `NimbusJwtDecoder`.

Example:
```java
  @Bean
  public JwtDecoder getJwtDecoder() throws ParseException {
        NimbusJwtDecoder jwtDecoder =  NimbusJwtDecoder
              .withJwkSetUri(this.properties.getJwkSetUri())
              .jwsAlgorithms(algorithms -> {
                  algorithms.add(SignatureAlgorithm.ES256);
                  algorithms.add(SignatureAlgorithm.ES384);
                  algorithms.add(SignatureAlgorithm.ES512);
              })
              .build();
        jwtDecoder.setJwtValidator(this.getTokenValidator());
        return jwtDecoder
  }
```

**And** I configure my Spring App as Oauth2 Resource Server, like:
```java 
@EnableWebSecurity
@Configuration
public class SecurityConfig {

    @Bean
    public SecurityFilterChain oauth2FilterChain(HttpSecurity http) throws Exception {
        http
            .csrf(AbstractHttpConfigurer::disable)
            .authorizeHttpRequests(authorize -> authorize
                    .requestMatchers("/api/v1/**").authenticated()
            )
            .oauth2ResourceServer(oauth2 -> oauth2
                    .jwt(Customizer.withDefaults()));

        return http.build();
    }
}
```

**When** I send a request to `/api/v1/**` with a JWT token that has a malformed JWT header
```http
POST /api/v1/test HTTP/1.1
Host: localhost:80
Authorization: Bearer <MELFORMED_HEADER>.<PAYLOAD>.<SIGNATURE>
```

**Expected Behavior**
**Then** I expect that a 401 is responded with a `www-authenticate` header which doesn't expose any app internal information.

Something like:
```
www-authenticate: Bearer error="invalid_token", error_description="An error occurred while attempting to decode the Jwt: Invalid unsecured/JWS/JWE", error_uri="https://tools.ietf.org/html/rfc6750#section-3.1"
```

**Current Behavior**
**Then** currently a 401 is responded with a `www-authenticate` header which does expose that the app is using a `Nimbusdecoder`.
```
[...]

PR #14185, opened 2023-11-22
Dropped Nimbus Error Message gh 13730

Caguilar92:
Caught ParseExcepetion, dropped nimbus error message, and replace with Malformed header.

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
@@ -144,6 +144,9 @@ public final class NimbusJwtDecoder implements JwtDecoder {
 		}
 		catch (Exception ex) {
 			this.logger.trace("Failed to parse token", ex);
+			if (ex instanceof ParseException) {
+				throw new BadJwtException(String.format(DECODING_ERROR_MESSAGE_TEMPLATE, "Malformed token"), ex);
+			}
 			throw new BadJwtException(String.format(DECODING_ERROR_MESSAGE_TEMPLATE, ex.getMessage()), ex);
 		}
 	}
--- a/oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
+++ b/oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
@@ -97,6 +97,8 @@ import static org.mockito.Mockito.verifyNoMoreInteractions;
  */
 public class NimbusJwtDecoderTests {
 
+	private static final String MALFORMED_TOKEN = "eyJhbGciOiJSUzI1NiJ9.eyJuYmYiOnt9LCJleHAiOjQ2ODQyMjUwODd9";
+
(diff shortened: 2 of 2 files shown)


oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -144,6 +144,9 @@ private JWT parse(String token) {
 		}
 		catch (Exception ex) {
 			this.logger.trace("Failed to parse token", ex);
+			if(ex.getCause() instanceof ParseException) {
+				throw new BadJwtException(String.format(DECODING_ERROR_MESSAGE_TEMPLATE, "Malformed header"), ex);
```
jzheaux, line 148: I think that the body could also be malformed. Can this say "Malformed token" instead?

gradle/libs.versions.toml
```diff
@@ -90,7 +90,7 @@ org-seleniumhq-selenium-selenium-support = "org.seleniumhq.selenium:selenium-sup
 org-skyscreamer-jsonassert = "org.skyscreamer:jsonassert:1.5.1"
 org-slf4j-log4j-over-slf4j = { module = "org.slf4j:log4j-over-slf4j", version.ref = "org-slf4j" }
 org-slf4j-slf4j-api = { module = "org.slf4j:slf4j-api", version.ref = "org-slf4j" }
-org-springframework-data-spring-data-bom = "org.springframework.data:spring-data-bom:2021.2.17"
+org-springframework-data-spring-data-bom = "org.springframework.data:spring-data-bom:2021.2.18"
```
jzheaux, line 93: Please leave this change out of the PR.

jzheaux (changes requested):
Thanks for the PR, @Caguilar92! In addition to my inline feedback, will you also please add a test that demonstrates the changed behavior?

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
Caguilar92: It appears like the exception here is thrown by the JWTParser when it cannot decode the header.

https://bitbucket.org/connect2id/nimbus-jose-jwt/src/release-9.24.4/src/main/java/com/nimbusds/jwt/JWTParser.java/#-71

jzheaux: It can also be thrown here: https://bitbucket.org/connect2id/nimbus-jose-jwt/src/d8a4358e56328637974b343a0834da67e20a6e30/src/main/java/com/nimbusds/jwt/SignedJWT.java#lines-98, which `JWTParser#parse` calls.

Further, the JavaDoc says:

```
* @throws ParseException If the string couldn't be parsed to a valid 
* unsecured, signed or encrypted JWT.
```

which I think pays dividends to follow.

Caguilar92: I agree, and I don't mean to go against the grain,  but the case where the body of the token is malformed seems to have already been addressed, indicated by this test case.
https://github.com/spring-projects/spring-security/blob/c336ca49fbeeea5b8acf6b87ccfcb3685358688f/oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java#L188-L195. Upon reviewing the issue submitted gh-13730, the exception that contained the unexpected message was thrown because the user introduced a malformed header, hence the reason for adding  "Malformed header" to the exception message, since a malformed token body is already addressed.

Caguilar92: I understand now. I agree, "malformed token" would be a more appropriate message.

jzheaux:
@Caguilar92, thanks for the updates. In preparation for merging, will you please ensure that there is only one commit in your PR?

gradle/libs.versions.toml
jzheaux: Sure, @Caguilar92. Please try rebasing your PR branch with `main`. 

It should be something like this. Let's say that the name for your remote is `origin` and the name for the spring-projects remote is called `spring-projects`. You would do:

```
git checkout main
git reset --hard spring-projects main
git checkout gh-13730
git rebase main 
// double-check here that there is now only one commit
git push origin +gh-13730
```

jzheaux: Also, congrats on doing your first PR! :)

Caguilar92: @jzheaux, thank you for your patience, guidance, and feedback. I really appreciate it!

PR #9881, opened 2021-06-08
Added support for the CAS gateway feature.

yvangg:
This is the CAS gateway feature.
This is a implementation based on the following discussion (PR): https://github.com/spring-projects/spring-security/pull/40#

I did some modifications in order to perform the CAS gateway redirection only if the CAS cookie is present on the client application request. Of course this will not fit in all situations (where CAS domain is different from client applications won't work), but for many of situations this is a pretty good approach.

Regards.

PR #14193, opened 2023-11-24
Added support for the CAS gateway feature

leleuj:
This PR aims to support the CAS `gateway` feature in the `spring-security-cas` library.

It follows: https://github.com/spring-projects/spring-security/pull/40 and especially: https://github.com/spring-projects/spring-security/pull/9881 as discussed with @marcusdacoregio.

Unit tests pass and I tested it manually.


marcusdacoregio:
Thank you very much @leleuj. I'll try to review this PR soon.

cas/src/main/java/org/springframework/security/cas/web/CasAuthenticationEntryPoint.java
```diff
@@ -98,7 +98,7 @@ protected String createServiceUrl(HttpServletRequest request, HttpServletRespons
 	 */
 	protected String createRedirectUrl(String serviceUrl) {
 		return CommonUtils.constructRedirectUrl(this.loginUrl, this.serviceProperties.getServiceParameter(), serviceUrl,
-				this.serviceProperties.isSendRenew(), false);
+				this.serviceProperties.isSendRenew(), this.serviceProperties.isGateway());
```
marcusdacoregio, line 101: Is it true that we should always send `gateway=true` to CAS, even if this is not a gateway request? In other words, if I do `serviceProperties.setGateway(true)` does that mean that every redirect to CAS will include `gateway=true`?

cas/src/main/java/org/springframework/security/cas/web/CasAuthenticationFilter.java
```diff
@@ -238,8 +241,14 @@ public Authentication attemptAuthentication(HttpServletRequest request, HttpServ
 		}
 		String serviceTicket = obtainArtifact(request);
 		if (serviceTicket == null) {
-			this.logger.debug("Failed to obtain an artifact (cas ticket)");
-			serviceTicket = "";
+			if (this.gateway) {
+				this.logger.debug("Gateway request with no CAS ticket");
+				return new CasGatewayNoServiceTicketAuthenticationToken();
```
marcusdacoregio, line 246: Am I missing something or this will not be passed to the `CasAuthenticationProvider`?

marcusdacoregio (changes requested):
As far as I understood from https://github.com/spring-projects/spring-security/pull/9881 and https://github.com/spring-projects/spring-security/pull/40 the idea behind this feature is to add a filter that would trigger the redirect to the CAS service with the `gateway=true` parameter. 

Based on [previous reviews](https://github.com/spring-projects/spring-security/pull/9881#pullrequestreview-807877540) from @rwinch, the design was pretty much aligned and there were some details left like having a mechanism to avoid hitting the CAS server for every request.

I might be missing something but I do not see how the changes and the new components proposed in this PR work together to achieve the desired behavior. If you could outline the idea or maybe put together a sample (you can use [this](https://github.com/spring-projects/spring-security-samples/tree/main/servlet/spring-boot/java/cas/login)) it would be great to help me understand the proposed design.

leleuj:
"_the idea behind this feature is to add a filter that would trigger the redirect to the CAS service with the gateway=true parameter_":
 
Not exactly, the idea behind this feature is not to add a new filter, this is just a possible implementation.

The idea behind this feature is to be able to call the CAS server with `gateway=true`, meaning that if the user is not auhtenticated, it comes back to the client app being anonymous and if the user is authenticated, it comes back to the client app being authenticated.

leleuj:
When revisiting this implementation, I realized that the `gateway=true` parameter is sent to the CAS server and this is the role of the entry point (to call the CAS server for login). So I understandably wrote this easy definition:

```java
    @Bean
    public ServiceProperties serviceProperties() {
        ServiceProperties serviceProperties = new ServiceProperties();
        serviceProperties.setService("http://localhost:8081/login/cas");
        serviceProperties.setSendRenew(false);
        serviceProperties.setGateway(true);
        return serviceProperties;
    }

    @Bean
    @Primary
    public AuthenticationEntryPoint authenticationEntryPoint() {
        CasAuthenticationEntryPoint entryPoint = new CasAuthenticationEntryPoint();
        entryPoint.setLoginUrl("http://localhost:8080/cas/login");
        entryPoint.setServiceProperties(serviceProperties());
        return entryPoint;
    }
```

It makes sense to attach the `gateway` configuration to the `serviceProperties`. The `renew` and `gateway` are similar protocol settings and it seems relevant to me to have them at the same place.

This is the way it was done in the original official CAS filter: https://github.com/apereo/java-cas-client#client-configuration-using-webxml

This is how I did it for Shiro and pac4j as well.

You can handle both the gateway mode and a regular CAS authentication using two `serviceProperties` definition and two entry points as it is possible to have multiple entry points in Spring Security.

The use case would a public/protected home page (with `gateway`) and a protected area for the rest of the webapp (without `gateway`).

Of course, this is a choice of design to have the `gateway` setting like the `renew` setting in the `serviceProperties`.  I can change it back to the previous proposal, but as I said, I think it's better this way.

Having a new filter for the `gateway` setting is possible, but it feels awkward to me to add a dedicated filter for this setting. Why not a specific filter to send the `renew` setting as well? I prefer to have the `gateway` setting integrated seamlessly in the general CAS support.

cas/src/main/java/org/springframework/security/cas/web/CasAuthenticationEntryPoint.java
leleuj: Yes, it means that all calls to CAS using **these** entry point and `serviceProperties` will have `gateway=true`, but you can have two `serviceProperties` (with and without `gateway`) and two entry points to have URLs in your client app where you use `gateway` and other URLs where you don't.

The use case would a public/protected home page (with `gateway`) and a protected area for the rest of the webapp (without `gateway`).

leleuj:
I had an issue with my test demo. I will test again and keep you posted.

leleuj:
Forget my previous comments for now.

leleuj:
You were right: I don't understand how the provider was called in the gateway mode.

Anyway, I restarted my work from the provided demo: https://github.com/spring-projects/spring-security-samples/tree/main/servlet/spring-boot/java/cas/login

cas/src/main/java/org/springframework/security/cas/web/CasAuthenticationFilter.java
leleuj: You were right, this was never passed to the `CasAuthenticationProvider`, but this isn't necessary.

leleuj:
I still keep the idea that no additional filter is required, the `CasAuthenticationFilter` can be used in `gateway` mode thanks to the `serviceProperties`.

Multiple filters can be defined on different URLs if need be.

leleuj:
The `CasAuthenticationFilter` is aware of the `gateway` configuration.

This is used to know if a specific token must be created for the "gateway mode" or if the regular CAS login must be performed:

```java
if (serviceTicket == null) {
	if (this.gateway) {
		this.logger.debug("Gateway request with no CAS ticket");
		return new CasGatewayNoServiceTicketAuthenticationToken();
	}
	else {
		this.logger.debug("Failed to obtain an artifact (cas ticket)");
		serviceTicket = "";
	}
}
```

This is the final authentication available: `authenticated=true, roles=ROLE_ANONYMOUS`.

leleuj:
For each request, the `CasAuthenticationFilter` and especially its `requiresAuthentication` method is called. Now, it is checked if the `gateway` mode is enabled and if the token is an "expired" gateway one:

```java
final Authentication auth = SecurityContextHolder.getContext().getAuthentication();
if (this.gateway && auth instanceof CasGatewayNoServiceTicketAuthenticationToken token) {
	final Instant creationDate = token.getCreationDate();
	final Instant startRetainDate = Instant.now().minusSeconds(this.gatewayAuthenticationRetainDelayInSeconds);
	// removes authentication after a certain retain delay
	if (creationDate.isBefore(startRetainDate)) {
		this.logger.debug("Removes authentication because of gateway setting");
		SecurityContextHolder.getContext().setAuthentication(null);
	}
}
```

If so, it is removed from the context and a new authentication is triggered, meaning a new round-trip to the CAS server is done (once again using the `gateway` support).

leleuj:
Sorry for the false start. Are things clearer now?

cas/src/main/java/org/springframework/security/cas/web/CasAuthenticationFilter.java
```diff
@@ -238,8 +244,14 @@ public Authentication attemptAuthentication(HttpServletRequest request, HttpServ
...
```
marcusdacoregio, line 249: What happens on a public URL when there is no `AuthenticationException`? Will the `CasAuthenticationEntryPoint` be invoked?
To be able to invoke this filter to different URLs, the `filterProcessingUrl` should allow customization, however, I'm still not convinced that it is the responsibility of this filter to be aware of the gateway request.

If we take a look at how OAuth2 components work, there are at least two filters, one that performs the redirect to the Authorization Server and another that processes the authorization response.

marcusdacoregio (changes requested):
Thanks, @leleuj, I left some feedback inline.

I still think that we should have a new filter with the responsibility of redirecting to the CAS server with `gateway=true` if needed.

I was taking a look at https://github.com/spring-projects/spring-security/pull/9881 and it looks like there is support to prevent making too many requests via `CasCookieGatewayRequestMatcher`

It is fine to have the `gateway` property inside `ServiceProperties`, that way the `CasAuthenticationFilter` can proper handle an `unsuccessfulAuthentication` if there is no service ticket.

cas/src/main/java/org/springframework/security/cas/authentication/CasGatewayNoServiceTicketAuthenticationToken.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * An {@link org.springframework.security.core.Authentication} implementation that
+ * represents a CAS gateway request with no service ticket returned (user is not
+ * authenticated at the CAS server level).
+ *
+ * @author Jerome LELEU
+ * @since 6.3.0
+ */
+public final class CasGatewayNoServiceTicketAuthenticationToken extends AbstractAuthenticationToken {
```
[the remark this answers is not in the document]
leleuj: OK. I get the point, but indeed, this may not be needed depending on the chosen strategy.

cas/src/main/java/org/springframework/security/cas/web/CasAuthenticationFilter.java
leleuj: The `CasAuthenticationEntryPoint` will only be invoked if the URL is configured with `authenticated`.

And the `CasGatewayNoServiceTicketAuthenticationToken` cannot be an `AnonymousAuthenticationToken` as it leads to an infinite loop.

On the other hand, this prevents triggering a new CAS round-trip and only leads to a 403 error page for another URL protected with non-gateway CAS login. It would certainly need a custom `AccessDeniedHandler`.

So although re-using the `CasAuthenticationFilter` for the `gateway` setting seems an alluring idea, I think I was wrong on this: I will create a new filter. I will do that next Monday.

I think it's better to close this PR and submit a new PR, isn't it?

leleuj: OK. I will do that today. Thanks for your time and advices

leleuj: @marcusdacoregio I just submitted a polished-to-main-branch PR from #9881.

Just let me know if you have additional comments.

cas/src/main/java/org/springframework/security/cas/web/TriggerCasGatewayFilter.java
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
+			throws IOException, ServletException {
+
+		HttpServletRequest request = (HttpServletRequest) req;
+		HttpServletResponse response = (HttpServletResponse) res;
+
+		if (this.requestMatcher.matches(request)) {
+			throw new TriggerCasGatewayException("Try a CAS gateway authentication");
```
marcusdacoregio, line 61: Based on @rwinch comment [here](https://github.com/spring-projects/spring-security/pull/9881/files#r750692748), I also think the exception should not be thrown here, instead, if the request matches, we should redirect to the CAS server with `gateway=true`.

cas/src/main/java/org/springframework/security/cas/web/TriggerCasGatewayFilter.java
```diff
@@ -0,0 +1,74 @@
...
+ * {@link TriggerCasGatewayException} when the current request matches against the
+ * configured {@link RequestMatcher}.
+ * <p>
+ * The default implementation you can use is {@link DefaultCasGatewayRequestMatcher}.
+ *
+ * @author Michael Remond
+ */
+public class TriggerCasGatewayFilter extends GenericFilterBean {
+
+	private RequestMatcher requestMatcher;
```
marcusdacoregio, line 47: The default could be the `CasCookieGatewayRequestMatcher` and it can be customized via `setRequestMatcher`.

The constructor can be removed.

cas/src/main/java/org/springframework/security/cas/web/CasAuthenticationEntryPoint.java
marcusdacoregio, line 1: There is probably no need to change this class since we will not rely on it to redirect to the CAS server, instead the `TriggerCasGatewayFilter` will do that for us.

cas/src/main/java/org/springframework/security/cas/web/CasAuthenticationFilter.java
```diff
@@ -226,6 +236,29 @@ protected final void successfulAuthentication(HttpServletRequest request, HttpSe
...
+		// the saved url
+		if (StringUtils.hasText(obtainArtifact(request))) {
+			super.unsuccessfulAuthentication(request, response, failed);
+		}
+		else {
+			SavedRequest savedRequest = this.requestCache.getRequest(request, response);
+			if (savedRequest != null) {
+				this.redirectStrategy.sendRedirect(request, response, savedRequest.getRedirectUrl());
+			}
+		}
```
marcusdacoregio, line 259: I'm finding this strategy to decide if that was a gateway request too sensitive.

Is it possible to make CAS redirect to the service URL with a query param? Something like `my.company.com/cas/login?gateway=true&service=https://server3.company.com/webapp/login/cas%3Fgateway=true`. With that, upon successful/failed authentication, the CAS server would redirect to `https://server3.company.com/webapp/login/cas?gateway=true` and we can be sure whether that was a gateway request or not.

marcusdacoregio (changes requested):
Thanks, @leleuj. It is looking nice.

I've left some feedback inline.

cas/src/main/java/org/springframework/security/cas/web/TriggerCasGatewayFilter.java
leleuj: I understand. I think the solution here is to use the `CasAuthenticationEntryPoint` to perform the gateway redirection.

cas/src/main/java/org/springframework/security/cas/web/TriggerCasGatewayFilter.java
leleuj: I see your point. I can set the `CasCookieGatewayRequestMatcher` by default.

cas/src/main/java/org/springframework/security/cas/web/CasAuthenticationEntryPoint.java
leleuj: There is some logic in the `CasAuthenticationEntryPoint` we may not want to duplicate in the `TriggerCasGatewayFilter` (like the `encodeServiceUrlWithSessionId` property).

Are you sure you don't want to use the `CasAuthenticationEntryPoint` in the `TriggerCasGatewayFilter?

cas/src/main/java/org/springframework/security/cas/web/CasAuthenticationFilter.java
leleuj: You're right, this strategy is a bit too sensitive (checking if we already have a saved request).

Though, using a parameter in the `serviceUrl` will trigger a lot of complexity. Indeed, the service tickets created by the CAS server are linked to the original service URL.
It means that we can't reuse the existing `ticketValidator` to validate service tickets created on a gateway request as the service URL won't match, it has `gateway=true` for example while we try to validate the service ticket for the service without `gateway=true`.

I can give it a try if you want. I just want you to confirm.

marcusdacoregio: Let's discard the parameter for now. What about setting some attribute in the session (if the session exists) and checking that attribute later on

marcusdacoregio: By the way, sorry for the back and forth on this, I haven't had time to test how this works on my own yet.

cas/src/main/java/org/springframework/security/cas/web/CasAuthenticationEntryPoint.java
leleuj: Yes, we can forget about the `encodeServiceUrlWithSessionId` which is related to CAS 2 (15 years old). I won't use the `CasAuthenticationEntryPoint` and ignore this setting.
It should even be removed (at least deprecated) in the `CasAuthenticationEntryPoint`.

cas/src/main/java/org/springframework/security/cas/web/CasAuthenticationFilter.java
leleuj: OK. I will try to use the session for that.

leleuj: No problem. Good design takes time ;-)

leleuj: In fact, removing the exception has a drawback: without exception, there is no saved request so I need to do that manually in the `TriggerCasGatewayFilter`

marcusdacoregio: No problem in that case, the `OAuth2AuthorizationRequestRedirectFilter` also saves the request before the redirect.

leleuj: OK. Good

leleuj: @marcusdacoregio I have updated the PR to take into account all your feedbacks:
- the exception has been removed
- the `TriggerCasGatewayFilter` performs the redirection to the CAS server with `gateway=true`
- the `CasAuthenticationEntryPoint` is no longer changed
- the `gateway` nature of the CAS round-trip is saved into the web session.

BTW, I have added a `TriggerCasGatewayFilterTests` to test the `TriggerCasGatewayFilter` and cropped all commits into one.

We should be good for merging. Just let me know.

leleuj: @marcusdacoregio Did you get some time to check this PR?

marcusdacoregio: I'll take a look at it today

marcusdacoregio:
Hello, @leleuj. I applied a polish commit on top of yours and I'll wait for the CI to pass to merge it. As soon as the SNAPSHOTs are available I'll update [the sample](https://github.com/spring-projects/spring-security-samples/tree/main/servlet/spring-boot/java/cas/login) to include the gateway feature.

marcusdacoregio (approved):

leleuj:
I'm not sure I understand your answer: "_We only add enhancements to minor versions._"

Would it be possible in a new version 5.9.x or only in version 6.3?

marcusdacoregio:
Sorry if I was not clear. We follow the semantic versioning (_major.minor.patch_), and our policy is to add new features only when we increment the minor version. Since the next minor release is `6.3` (we are currently in `6.2`), that release is the target for this PR.

The `5.x` line ended with `5.8`.

leleuj:
OK. That makes sense. Thanks for your time and help.

PR #14196, opened 2023-11-26
Add RoleHierarchyBuilder

### LINKED ISSUE #13300: Easier way to define Role Hierarchy
**Disclaimer**: I'm hesitant to create this issue for sometime as this is quite subjective, but I thought that I still want to make the suggestion to see if there's any other ways to achieve it (that I've perhaps missed out on)

---

Currently, if I want to define a `RoleHierarchy`, I can do it like

```java
@Bean
public RoleHierarchy roleHierarchy() {
    RoleHierarchyImpl roleHierarchy = new RoleHierarchyImpl();
    String roleHierarchyFromMap = "ROLE_ADMIN > ROLE_STAFF \n ROLE_STAFF > ROLE_USER \n ROLE_STAFF > ROLE_GUEST";
    roleHierarchy.setHierarchy(roleHierarchyFromMap);
    return roleHierarchy;
}
```

The result would be something like

```log
2023-06-09 21:17:39.881 DEBUG 20580 --- [  restartedMain] o.s.s.a.h.RoleHierarchyImpl              : setHierarchy() - The following role hierarchy was set: ROLE_ADMIN > ROLE_STAFF
 ROLE_STAFF > ROLE_USER
 ROLE_STAFF > ROLE_GUEST
2023-06-09 21:17:39.884 DEBUG 20580 --- [  restartedMain] o.s.s.a.h.RoleHierarchyImpl              : buildRolesReachableInOneStepMap() - From role ROLE_ADMIN one 
can reach role ROLE_STAFF in one step.
2023-06-09 21:17:39.885 DEBUG 20580 --- [  restartedMain] o.s.s.a.h.RoleHierarchyImpl              : buildRolesReachableInOneStepMap() - From role ROLE_STAFF one 
can reach role ROLE_USER in one step.
2023-06-09 21:17:39.885 DEBUG 20580 --- [  restartedMain] o.s.s.a.h.RoleHierarchyImpl              : buildRolesReachableInOneStepMap() - From role ROLE_STAFF one 
can reach role ROLE_GUEST in one step.
2023-06-09 21:17:39.886 DEBUG 20580 --- [  restartedMain] o.s.s.a.h.RoleHierarchyImpl              : buildRolesReachableInOneOrMoreStepsMap() - From role ROLE_STAFF one can reach [ROLE_USER, ROLE_GUEST] in one or more steps.
2023-06-09 21:17:39.888 DEBUG 20580 --- [  restartedMain] o.s.s.a.h.RoleHierarchyImpl              : buildRolesReachableInOneOrMoreStepsMap() - From role ROLE_ADMIN one can reach [ROLE_USER, ROLE_STAFF, ROLE_GUEST] in one or more steps.
```

Which is correct, but it is more prone to mistake (with more complex setup), and "uglier" to declare this way via `\n`

Luckily, for most use case, I can replace it with `RoleHierarchyUtils.roleHierarchyFromMap` but not for this case where there's a repeated key. And since we are using `Map`, `key` has to be unique.

```java
@Bean
public RoleHierarchy roleHierarchy() {
    RoleHierarchyImpl roleHierarchy = new RoleHierarchyImpl();
    Map<String, List<String>> roleHierarchyMap = new HashMap<>();

    roleHierarchyMap.put("ROLE_ADMIN", List.of("ROLE_STAFF"));
    roleHierarchyMap.put("ROLE_STAFF", List.of("ROLE_USER"));
    // Because it is Map, I can't quite define this
    // roleHierarchyMap.put("ROLE_STAFF", List.of("ROLE_GUEST"));
    String roleHierarchyFromMap = RoleHierarchyUtils.roleHierarchyFromMap(roleHierarchyMap);

    roleHierarchy.setHierarchy(roleHierarchyFromMap);
    return roleHierarchy;
}
```

Note that for this use case, `ROLE_STAFF` inherit `ROLE_USER` AND inherit `ROLE_GUEST` **but** `ROLE_USER` does not inherit `ROLE_GUEST`. Hence, I can't simply do the following

```java
@Bean
public RoleHierarchy roleHierarchy() {
    RoleHierarchyImpl roleHierarchy = new RoleHierarchyImpl();
    Map<String, List<String>> roleHierarchyMap = new HashMap<>();

    roleHierarchyMap.put("ROLE_ADMIN", List.of("ROLE_STAFF"));
    roleHierarchyMap.put("ROLE_STAFF", List.of("ROLE_USER", "ROLE_GUEST"));
    String roleHierarchyFromMap = RoleHierarchyUtils.roleHierarchyFromMap(roleHierarchyMap);

    roleHierarchy.setHierarchy(roleHierarchyFromMap);
    return roleHierarchy;
}
```

Which is a limitation for this use case.

I have another way to declare. Since Java now supports `TextBlock`, I can still define as

```java
@Bean
public RoleHierarchy roleHierarchy() {
    RoleHierarchyImpl roleHierarchy = new RoleHierarchyImpl();
    Map<String, List<String>> roleHierarchyMap = new HashMap<>();

    String roleHierarchyFromMap = """
            ROLE_ADMIN > ROLE_STAFF
            ROLE_STAFF > ROLE_USER
            ROLE_STAFF > ROLE_GUEST
            """;

    roleHierarchy.setHierarchy(roleHierarchyFromMap);
    return roleHierarchy;
}
```

Which still works. But I do still like the way that I use `Map` to define my hierarchy, which is still clearer, better and straightforward IMO. And I think it's easier to support the declaration via `application.yaml` and construct through `@ConfigurationProperties` and set it.

I wonder if there's any way to support my use case which make the declaration easier. 

Thanks!

FdHerrera:
Create builder class to declaratively create a RoleHierarchy with a fluent API, add test. Closes #13300
gh-13300

### DIFF
--- a/core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImpl.java
+++ b/core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImpl.java
@@ -20,8 +20,11 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Stream;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -30,6 +33,9 @@ import org.springframework.core.log.LogMessage;
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.authority.AuthorityUtils;
 import org.springframework.security.core.authority.SimpleGrantedAuthority;
+import org.springframework.util.Assert;
+
+import static org.springframework.security.access.hierarchicalroles.RoleHierarchyUtils.roleHierarchyFromMap;
 
 /**
  * <p>
@@ -98,6 +104,35 @@ public class RoleHierarchyImpl implements RoleHierarchy {
 	 * (i.e. fully resolved hierarchy from {@link #rolesReachableInOneStepMap})
 	 */
 	private Map<String, Set<GrantedAuthority>> rolesReachableInOneOrMoreStepsMap = null;
+
+	/**
+	 * Factory method that creates a {@link Builder} instance with the default role prefix
+	 * "ROLE_"
+	 * @return a {@link Builder} instance with the default role prefix "ROLE_"
+	 */
+	public static Builder withDefaultRolePrefix() {
+		return withRolePrefix("ROLE_");
+	}
+
+	/**
+	 * Factory method that creates a {@link Builder} instance with the specified role
+	 * prefix.
+	 * @param rolePrefix the prefix to be used for the roles in the hierarchy.
+	 * @return a new {@link Builder} instance with the specified role prefix
+	 * @throws IllegalArgumentException if the provided role prefix is null
+	 */
+	public static Builder withRolePrefix(String rolePrefix) {
+		Assert.notNull(rolePrefix, "rolePrefix must not be null");
+		return new Builder(rolePrefix);
+	}
+
+	/**
+	 * Factory method that creates a {@link Builder} instance with no role prefix.
+	 * @return a new {@link Builder} instance with no role prefix.
+	 */
+	public static Builder withNoRolePrefix() {
+		return withRolePrefix("");
+	}
 
 	/**
 	 * Set the role hierarchy and pre-calculate for every role the set of all reachable
@@ -209,6 +244,80 @@ public class RoleHierarchyImpl implements RoleHierarchy {
 			logger.debug(LogMessage.format(
 					"buildRolesReachableInOneOrMoreStepsMap() - From role %s one can reach %s in one or more steps.",
 					roleName, visitedRolesSet));
+		}
+
+	}
+
+	/**
+	 * Builder class for constructing a {@link RoleHierarchyImpl} based on a hierarchical
+	 * role structure.
+	 *
+	 * @author Federico Herrera
+	 * @since 6.3
+	 */
+	public static final class Builder {
+
+		private final String rolePrefix;
+
+		private final Map<String, List<String>> roleBranches;
+
+		private Builder(String rolePrefix) {
+			this.rolePrefix = rolePrefix;
+			this.roleBranches = new LinkedHashMap<>();
+		}
+
+		/**
+		 * Creates a new hierarchy branch to define a role and its child roles.
+		 * @param role the highest role in this branch
+		 * @return a {@link RoleBranchBuilder} to define the child roles for the
+		 * <code>role</code>
+		 */
+		public RoleBranchBuilder role(String role) {
+			Assert.hasText(role, "role must not be empty");
+			return new RoleBranchBuilder(this, rolePrefix.concat(role));
+		}
+
+		/**
+		 * Builds and returns a {@link RoleHierarchyImpl} describing the defined role
+		 * hierarchy.
+		 * @return a {@link RoleHierarchyImpl}
+		 */
+		public RoleHierarchyImpl build() {
+			String roleHierarchyRepresentation = roleHierarchyFromMap(roleBranches);
+			RoleHierarchyImpl roleHierarchy = new RoleHierarchyImpl();
+			roleHierarchy.setHierarchy(roleHierarchyRepresentation);
+			return roleHierarchy;
+		}
+
+		/**
+		 * Builder class for constructing child roles within a role hierarchy branch.
+		 */
+		public static final class RoleBranchBuilder {
+
+			private final Builder parentBuilder;
+
+			private final String role;
+
+			private RoleBranchBuilder(Builder parentBuilder, String role) {
+				this.parentBuilder = parentBuilder;
+				this.role = role;
+			}
+
+			/**
+			 * Specifies implied role(s) for the current role in the hierarchy.
+			 * @param impliedRoles role name(s) implied by the role.
+			 * @return the same {@link Builder} instance
+			 * @throws IllegalArgumentException if <code>impliedRoles</code> is null,
+			 * empty or contains any null element.
+			 */
+			public Builder implies(String... impliedRoles) {
+				Assert.notEmpty(impliedRoles, "at least one implied role must be provided");
+				Assert.noNullElements(impliedRoles, "implied role name(s) cannot be empty");
+				parentBuilder.roleBranches.put(role,
+						Stream.of(impliedRoles).map(parentBuilder.rolePrefix::concat).toList());
+				return parentBuilder;
+			}
+
 		}
 
 	}
--- a/core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImpl.java
+++ b/core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImpl.java
@@ -24,7 +24,6 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.stream.Stream;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -34,8 +33,6 @@ import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.authority.AuthorityUtils;
 import org.springframework.security.core.authority.SimpleGrantedAuthority;
 import org.springframework.util.Assert;
-
-import static org.springframework.security.access.hierarchicalroles.RoleHierarchyUtils.roleHierarchyFromMap;
 
 /**
  * <p>
@@ -109,6 +106,7 @@ public class RoleHierarchyImpl implements RoleHierarchy {
 	 * Factory method that creates a {@link Builder} instance with the default role prefix
 	 * "ROLE_"
 	 * @return a {@link Builder} instance with the default role prefix "ROLE_"
+	 * @since 6.3
 	 */
 	public static Builder withDefaultRolePrefix() {
 		return withRolePrefix("ROLE_");
@@ -120,18 +118,11 @@ public class RoleHierarchyImpl implements RoleHierarchy {
 	 * @param rolePrefix the prefix to be used for the roles in the hierarchy.
 	 * @return a new {@link Builder} instance with the specified role prefix
 	 * @throws IllegalArgumentException if the provided role prefix is null
+	 * @since 6.3
 	 */
 	public static Builder withRolePrefix(String rolePrefix) {
 		Assert.notNull(rolePrefix, "rolePrefix must not be null");
 		return new Builder(rolePrefix);
-	}
-
-	/**
-	 * Factory method that creates a {@link Builder} instance with no role prefix.
-	 * @return a new {@link Builder} instance with no role prefix.
-	 */
-	public static Builder withNoRolePrefix() {
-		return withRolePrefix("");
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+ * Relies on  {@link RoleHierarchyUtils#roleHierarchyFromMap(Map)} to build the string.
+ */
+public class RoleHierarchyBuilder {
+	private final Map<String, List<String>> roleBranches;
+
+	private RoleHierarchyBuilder() {
+		this.roleBranches = new LinkedHashMap<>();
+	}
+
+	public static RoleHierarchyBuilder newBuilder() {
```
jzheaux, line 22: Since any role hierarchy will require at least one role, I think it would be nice to call this `role`, removing one call.

core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+		}
+
+		/**
+		 * Implied role(s).
+		 * @param impliedRoles role name(s) implied by the role.
+		 * @return the same {@link RoleHierarchyBuilder}
+		 * @throws IllegalArgumentException if <code>impliedRoles</code> is null or empty
+		 * or contains any null element.
+		 */
+		public RoleHierarchyBuilder inheritsFromAllOf(String... impliedRoles) {
```
jzheaux, line 62: Can we call this `implies`? I think the simpler name makes the DSL more readable. It also aligns the method name with the parameter name and JavaDoc.

core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.List;
+import java.util.Map;
+
+import static org.springframework.security.access.hierarchicalroles.RoleHierarchyUtils.roleHierarchyFromMap;
+
+/**
+ * Class to provide a fluent API to build a {@link String} representing a {@link RoleHierarchy}.
+ * Relies on  {@link RoleHierarchyUtils#roleHierarchyFromMap(Map)} to build the string.
+ */
+public class RoleHierarchyBuilder {
```
jzheaux, line 15: Please make the class `final` until it's clear that extendability is needed.

core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.springframework.security.access.hierarchicalroles.RoleHierarchyUtils.roleHierarchyFromMap;
+
+/**
+ * Class to provide a fluent API to build a {@link String} representing a {@link RoleHierarchy}.
+ * Relies on  {@link RoleHierarchyUtils#roleHierarchyFromMap(Map)} to build the string.
```
jzheaux, line 13: Please add `@since 6.3`. Please also consider adding an `@author` tag.

core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+
+	/**
+	 * Builds the hierarchy using {@link RoleHierarchyUtils#roleHierarchyFromMap(Map)}
+	 * @return a string representation of a role hierarchy.
+	 */
+	public String build() {
+		return roleHierarchyFromMap(roleBranches);
+	}
+
+	public static class RoleBranchBuilder {
```
jzheaux, line 45: Please also make this class `final`.

core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+		private final RoleHierarchyBuilder parentBuilder;
+		private final String role;
+
+		RoleBranchBuilder(String role, RoleHierarchyBuilder roleHierarchyBuilder) {
```
jzheaux, line 49: Please make the constructor `private`.

jzheaux (changes requested):
Thanks for the PR, @FdHerrera! Given that this is ultimately in the service of `RoleHierarchyImpl`, will you please move this into that class? In that way, I believe it will also better align with https://github.com/spring-projects/spring-security/pull/13788

core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyBuilder.java
jzheaux: Note that I believe it is moved into `RoleHierarchyImpl`, then the two `role` methods won't conflict since this one will live in `RoleHierarchyImpl` itself.

FdHerrera:
Thank you very much for your feedback, @jzheaux ! 

I'll start implementing your suggestions.

core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyBuilder.java
FdHerrera: Hello @jzheaux 

While implementing this suggestion I think placing the call to the builder in the implementation is kind of mixing the roles of the entities.

I suggest, instead, leave the building of the `RoleHierarchyImpl` to live only in the builder, and refactor the final `build()` method to create a fully featured `RoleHierarchyImpl` (using the utility method `RoleHierarchyUtills#roleHierarchyFromMap` and `RoleHierarchyImpl#setHierarchy`, this way, it will have a complete and clearer scope and will reduce the impact on related classes, it will also allow doing something like the follow:

```
@Bean
public RoleHierarchy roleHierarchy() {
    return RoleHierarchyBuilder.newBuilder()
        .role("ROLE_A").impliesAllOf("ROLE_B")
        .role("ROLE_B").impliesAllOf("ROLE_C", "ROLE_D")
        .build();
}
```

Elegant and concise, really straightforward. 
Wdyt?

FdHerrera: Thank you for your feedback and help to further understand your idea, @jzheaux !

Makes total sense so I'm planning to move the builder to the class `RoleHierarchyImpl` to follow the convention of the classes you mention.

Will try to implement  a `role(String)` method in that class to better align with the last comment in this thread.

Thank you for sharing your thoughts!

core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImpl.java
```diff
@@ -213,4 +222,65 @@ else if (roleName.equals(lowerRole.getAuthority())) {
...
+			}
+
+			/**
+			 * Implied role(s).
+			 * @param impliedRoles role name(s) implied by the role.
+			 * @return the same {@link Builder}
+			 * @throws IllegalArgumentException if <code>impliedRoles</code> is null or
+			 * empty or contains any null element.
+			 */
+			public Builder impliesAllOf(String... impliedRoles) {
```
jzheaux, line 276: I think it would be better if this were called `implies`. The reason is that `allOf` is implied by the fact that it is a varargs.

I think the `allOf` prefix makes sense for things like `AuthorizationManagers` since you can have an `anyOf` there too, which makes the clarification necessary in that case.

core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImpl.java
```diff
@@ -99,6 +104,10 @@ public class RoleHierarchyImpl implements RoleHierarchy {
 	 */
 	private Map<String, Set<GrantedAuthority>> rolesReachableInOneOrMoreStepsMap = null;
 
+	public static Builder.RoleBranchBuilder role(String role) {
```
jzheaux, line 107: I believe [this comment](https://github.com/spring-projects/spring-security/issues/13300#issuecomment-1606115443) raises something that this PR hasn't considered yet, which is role prefixes.

What if instead we went back to your earlier suggestion and did the following:

```java
public static Builder builder() {
    return new Builder("ROLE_");
}

public static Builder withRolePrefix(String rolePrefix) {
    return new Builder(rolePrefix);
}
```

The reason is that, in Spring Security, things like `hasRole("USER")` imply that `ROLE_` will be prefixed automatically, as opposed to `hasAuthority(`ROLE_USER`)` where the application must supply the prefix. `role` methods in this class should do the same thing.

I think this arrangement of methods is better aligned since they both return the same `Builder` class and we also don't introduce a `role` method that doesn't automatically prefix the underlying authority.

core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImpl.java
```diff
@@ -213,4 +222,65 @@ else if (roleName.equals(lowerRole.getAuthority())) {
...
+	 */
+	public static final class Builder {
+
+		private final Map<String, List<String>> roleBranches;
+
+		private Builder() {
+			this.roleBranches = new LinkedHashMap<>();
+		}
+
+		public RoleBranchBuilder role(String role) {
```
jzheaux, line 241: Thanks for adding JavaDoc. Will you please add a JavaDoc here, too?

jzheaux (changes requested):
Thanks, @FdHerrera! Things are shaping up nicely. I've left a little bit more feedback inline.

Also, in preparation for merging, will you please squash your commits into a single commit that looks something like this:

```
Add RoleHierarchy.Builder

Closes gh-13300
```

core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImpl.java
jzheaux: Looking back at my first review, I realize that one of my comments used `implies` and another used `impliesAllOf`. Sorry for potentially being confusing about my position.

core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImpl.java
FdHerrera: Sure, makes sense, `implies` should also prefix ROLE to each one? Something like
```
role("FOO").implies("BAR")
```

resulting in: `ROLE_FOO > ROLE_BAR` ?

FdHerrera: Or whatever prefix is defined using `withRolePrefix`

I'm thinking maybe `builderWithRolePrefix` ? Or just overload `builder` with a parameter `rolePrefix`?

Just because `RoleHierarchyImpl.withRolePrefix(rolePrefix)` doesn't sound very intuitive to me. Wdyt?

jzheaux: > `role("FOO").implies("BAR")`

Yes, that's the idea.

> Or whatever prefix is defined using `withRolePrefix`

Good question. The `with` prefix comes from a naming convention in Spring Security. `withXXX` means "give me a builder based on XXX" and `fromXXX` means "construct an instance using values XXX". An example is `NimbusJwtDecoder#withJwkSetUri`.

If what you are looking for is to align the method names, maybe we consider something like `withDefaults()` or `withDefaultRolePrefix()` instead of `builder()`. I kind of like the subtle reminder that `withDefaultRolePrefix()` will include the default role prefix for you.

FdHerrera: Liked the `withDefaultRolePrefix()`

Went ahead with this approach and implemented another one with no prefix, for cases where no standard prefix is used.

That led to 3 factory methods to build a hierarchy: `withDefaultRolePrefix`, `withRolePrefix` and `withNoRolePrefix`

Let me know if it looks good for you! Ty in advance

FdHerrera:
Thank you for the thorough review and constructive feedback! I appreciate the collaborative back-and-forth @jzheaux .

I implemented your suggestions and squashed the commits, I think is looking pretty good let me hear your thoughts!

FdHerrera:
> I believe it will also better align with #13788

Also, regarding this comment, should I change the commit and description to close that issue instead of gh-13300?

jzheaux:
Thanks, @FdHerrera! I've added a polish to make it align more closely with the rest of the codebase stylistically.

For now, I think we should leave out `withNoRolePrefix` and wait to see how the community uses the class. In other places in Spring Security, `ROLE_` is the default and you supply a string value to override the default. There isn't the notion elsewhere in the code base of "no prefix". I went ahead and adjusted that in my polish commit. Let me know if me removing that method causes a problem.

FdHerrera:
Looks good @jzheaux thank you for your help!

jzheaux (approved):

PR #14235, opened 2023-12-01
Add a custom converter in SpringOpaqueTokenIntrospector

### LINKED ISSUE #14198: Convert custom claim in NimbusOpaqueTokenIntrospector
**Expected Behavior**

Our current OAuth2 provider returns a custom claim for the roles, which I would like to convert.

Currently I cannot override the method here
https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/NimbusOpaqueTokenIntrospector.java#L212 because the convert method is private.

Could this method be made protected or public or offer other alternatives to convert custom claims?

**Current Behavior**

Everything is private for the NimbusOpaqueTokenIntrospector so i have no chance to convert custom claims.

**Context**

Example what I we currently need to convert roles

```
   private Collection<GrantedAuthority> getAuthorities( final TokenIntrospectionSuccessResponse response, final Map<String, Object> claims ) {
      final Collection<GrantedAuthority> authorities = new ArrayList<>();
      if ( response.getScope() != null ) {
         final List<String> scopes = Collections.unmodifiableList( response.getScope().toStringList() );
         claims.put( SCOPE, scopes );

         for ( final String scope : scopes ) {
            authorities.add( new SimpleGrantedAuthority( authorityPrefix + scope ) );
         }
      }

      final List<String> roles = response.getStringListParameter( roleClaim );
      if ( !CollectionUtils.isEmpty( roles ) ) {
         claims.put( roleClaim, roles );
         roles.forEach( role -> authorities.add( new SimpleGrantedAuthority( rolePrefix + role ) ) );
      }
      return authorities;
   }
```

ahmd-nabil:
Closes #14198

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringOpaqueTokenIntrospector.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringOpaqueTokenIntrospector.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 
 import org.apache.commons.logging.Log;
@@ -69,7 +70,7 @@ public class SpringOpaqueTokenIntrospector implements OpaqueTokenIntrospector {
 
 	private Converter<String, RequestEntity<?>> requestEntityConverter;
 
-	private Converter<OAuth2TokenIntrospectionClaimAccessor, OAuth2AuthenticatedPrincipal> authenticationConverter;
+	private Converter<OAuth2TokenIntrospectionClaimAccessor, ? extends OAuth2AuthenticatedPrincipal> authenticationConverter = this::defaultAuthenticationConverter;
 
 	/**
 	 * Creates a {@code OpaqueTokenAuthenticationProvider} with the provided parameters
@@ -85,7 +86,6 @@ public class SpringOpaqueTokenIntrospector implements OpaqueTokenIntrospector {
 		RestTemplate restTemplate = new RestTemplate();
 		restTemplate.getInterceptors().add(new BasicAuthenticationInterceptor(clientId, clientSecret));
 		this.restOperations = restTemplate;
-		this.authenticationConverter = this.defaultAuthenticationConverter();
 	}
 
 	/**
@@ -100,7 +100,6 @@ public class SpringOpaqueTokenIntrospector implements OpaqueTokenIntrospector {
 		Assert.notNull(restOperations, "restOperations cannot be null");
 		this.requestEntityConverter = this.defaultRequestEntityConverter(URI.create(introspectionUri));
 		this.restOperations = restOperations;
-		this.authenticationConverter = this.defaultAuthenticationConverter();
 	}
 
 	private Converter<String, RequestEntity<?>> defaultRequestEntityConverter(URI introspectionUri) {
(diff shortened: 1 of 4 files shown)


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringOpaqueTokenIntrospector.java
```diff
@@ -222,7 +227,17 @@ private OAuth2AuthenticatedPrincipal convertClaimsSet(Map<String, Object> claims
 			}
 			return v;
 		});
-		return new OAuth2IntrospectionAuthenticatedPrincipal(claims, authorities);
+		return this.authenticationConverter.convert(new OAuth2IntrospectionAuthenticatedPrincipal(claims, authorities));
```
jzheaux, line 230: I think it would provide more value if the default authentication converter took the claim set and added the authorities instead of being a pass-through. The reason is that this allows the code to align with `JwtAuthenticationConverter` down the road.

In the case of `JwtAuthenticationConverter`, the instance moves from `Jwt` (essentially the claim set) to an `Authentication` (something with authorities), and I think that we should do the same thing here.

More concretely, I think the default converter should be invoked like this:

```java
return this.authenticationConverter.convert(() -> claims);
```

and the default converter should do the work of computing the granted authorities and constructing an `OAuth2IntrospectionAuthenticatedPrincipal`. Will that work for you?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringOpaqueTokenIntrospector.java
```diff
@@ -222,7 +227,17 @@ private OAuth2AuthenticatedPrincipal convertClaimsSet(Map<String, Object> claims
...
+	}
+
+	private Converter<OAuth2TokenIntrospectionClaimAccessor, OAuth2AuthenticatedPrincipal> defaultAuthenticationConverter() {
+		return (accessor) -> (OAuth2IntrospectionAuthenticatedPrincipal) accessor;
+	}
+
+	public void setAuthenticationConverter(
```
jzheaux, line 237: Will you please add JavaDoc here that describes what this method is and when to use it? It would also be nice if the doc explained what the default authentication converter does. It should also have a `@since 6.3` attribute.

jzheaux (changes requested):
Thanks, @ahmd-nabil, for the PR! I've left my feedback inline.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringOpaqueTokenIntrospector.java
ahmd-nabil: thanks @jzheaux for feedback. currently working on it. I was thinking to change `convertClaimsSet ` method return type to `Map<String, Object>` to return claims and then use `this.authenticationConverter.convert(() -> claims)` inside `introspect()` method instead of `convertClaimsSet()`. does this seem right ?

jzheaux: Yes, I think that makes a lot of sense.

jzheaux (changes requested):
This looks great, @ahmd-nabil. I think it's fine if you want to rearrange the private method structure a bit. Once you feel ready with that, will you please squash your commits into one that looks something like this:

```
Add Introspection Authentication Converter

Closes gh-14198
```

and then we can merge things.

jzheaux (approved):

PR #14255, opened 2023-12-07
Fix wrong document about architecture

305088020:
In the paragraph of custom filter

I found that one part of the document was written incorrectly, so I revised it.


305088020:
Thinks！I regard `authentication filters` as `AuthorizationFilter`😅

Issue #13990, opened 2023-10-10
Consider adding dependency convergence detection


sjohnr:
We should consider adding dependency convergence detection to our build to prevent issues like gh-13843. For example, the following dependencies must agree on the version of `com.nimbusds:nimbus-jose-jwt`.

* `spring-security-oauth2-client` -> `com.nimbusds:oauth2-oidc-sdk` -> `com.nimbusds:nimbus-jose-jwt`
* `spring-security-oauth2-jose` -> `com.nimbusds:nimbus-jose-jwt`

See [comment](https://github.com/spring-projects/spring-security/issues/13843#issuecomment-1746737794) and related issue for more information.

marcusdacoregio:
@sjohnr if there aren't more dependencies that have to be checked, this might have been resolved via https://github.com/spring-projects/spring-security/issues/14047

sjohnr:
@marcusdacoregio based on [this comment](https://github.com/spring-projects/spring-security/issues/13843#issuecomment-1746737794), I think the idea was to add general purpose detection when convergence issues arise. But I don't have too much experience with that. gh-14047 definitely solves the most immediate problem, so I'll let others weigh in on whether more general detection is still needed. @ThomasKasene

ThomasKasene:
I'll weigh in my own experience - make of it what you will 😄

My team maintains a large number of Spring Boot applications which all use Spring Security for OAuth 2.0 stuff. When Boot 3.1.1 (I think) was released, Dependabot gave us a bunch of PRs to upgrade, but all the builds broke because we use the `maven-enforcer-plugin` to detect dependency convergence errors. The culprit was, of course, the fact that `spring-security-oauth2-jose` and `oauth2-oidc-sdk` depended on different versions of `nimbus-jose-jwt`.

We _could_ have overridden the versions of the offending dependency in all the apps, and that would've been the "correct" thing to do. Doing so would've meant having to look out for changes/fixes in Spring Security to this issue (like the one provided through #13843). In the meantime, we'd have to manage any Dependabot PRs about `nimbus-jose-jwt` in all the apps, which would possibly all be invalid anyway because both Spring Security and `oauth2-oidc-sdk` use very old versions of `nimbus-jose-jwt`.

In the end, a general lack of bandwidth forced us to jump to plan B, which was simply to not upgrade to Spring Boot 3.1.1 at all. The fix didn't arrive until 3.1.5, so our adoption rate crashed down to zero in the meantime.

[...]

Issue #14245, opened 2023-12-05
Consider removing com.nimbusds:oauth2-oidc-sdk dependency


jgrandja:
We should strongly consider removing the `com.nimbusds:oauth2-oidc-sdk` dependency as it has caused a number of issues over the years ever since we added it's usage in the OAuth2 support.

The root issue is the result of the out-of-sync release process used for the `oauth2-oidc-sdk` and `nimbus-jose-jwt` artifacts. Since `oauth2-oidc-sdk` depends on `nimbus-jose-jwt`, it's important to keep the versions of these artifacts in sync - only if `oauth2-oidc-sdk` is being used. However, there are cases where this poses to be a challenge, when artifacts are unexpectedly downgraded as described in this [comment](https://github.com/spring-projects/spring-security/issues/13843#issuecomment-1737985635). As a result, we're also forced to downgrade versions to prevent dependency convergence, and this is not ideal as we should always strive to be using the latest versions as a best practice.

The support overhead that `com.nimbusds:oauth2-oidc-sdk` dependency has added over the years has far outweighed the value of reuse of the library, so I recommend we remove the use of the library altogether and only depend on `com.nimbusds:nimbus-jose-jwt` and ensure we're always on the latest version.

We are currently using `com.nimbusds:oauth2-oidc-sdk` in the following classes:

- `OAuth2ErrorResponseErrorHandler`
- `ClientRegistrations`
- `DefaultReactiveOAuth2UserService`
- `OAuth2AccessTokenResponseBodyExtractor`
- `NimbusOpaqueTokenIntrospector`
- `NimbusReactiveOpaqueTokenIntrospector`

After reviewing the code in the above classes, I believe updating the code with our own implementation would be minimal effort and it ultimately would save us from any further support overhead.

Related gh-13843, gh-13990

jzheaux:
The `OpaqueTokenIntrospector` classes can be deprecated at this point I think. We kept them around for a while in case folks needed the Nimbus support, but now there's a Nimbus-free implementation for both servlet and reactive.

xak2000:
I'll add my 5 cents.

The mentioned unexpected downgrade of `nimbus-jose-jwt` from 9.31 to 9.24.4 actually caused problems in our project as our code was already dependent on some methods, added in 9.26. For instance, `RSAKey.Builder.notBeforeTime` and `RSAKey.Builder.expirationTime`. After upgrading to spring-security 6.1.5, these methods disappeared. :)

[...]

PR #14256, opened 2023-12-07
Dependency convergence detection

andreilisa:
Adding dependency convergence detection

Related to [13990](https://github.com/spring-projects/spring-security/issues/13990)


buildSrc/settings.gradle
```diff
@@ -1,3 +1,35 @@
...
+		r.dependencies.add("commons-codec:commons-codec:1.15")
+		r.dependencies.add("com.fasterxml.jackson.core:jackson-databind:2.13.4.2")
+		r.dependencies.add("com.fasterxml.jackson.core:jackson-core:2.13.4")
+		r.dependencies.add("com.google.guava:guava:30.0-jre")
+		r.dependencies.add("org.apache.httpcomponents:httpclient:4.5.14")
+		r.dependencies.add("com.thoughtworks.xstream:xstream:1.4.19")
+		r.dependencies.add("org.apache.httpcomponents:httpcore:4.4.16")
+		r.dependencies.add("net.bytebuddy:byte-buddy:1.14.6")
+		r.dependencies.add("org.junit.jupiter:junit-jupiter-api:5.10.1")
+	}
```
marcusdacoregio, line 30: Can you explain why this is needed? I wonder if we will need to keep updating this rule often.

marcusdacoregio: Hi, @andreilisa. I haven't got time to go through it yet. I'm not also sure if we should force such dependencies. I'll try to review it soon to have an idea if it can cause any trouble for us in the future.

marcusdacoregio:
Hi, @andreilisa. Thank you so much for your time in investigating how to integrate the dependency convergence detection into the build. However, I feel that even if applying this change, we are not going to fix the problem (which doesn't exist anymore, this is just a safety mechanism to prevent that for happening again) and it may cause more trouble resolving our dependencies. There is also some thoughts being given to get rid of the dependency that causes trouble #14245, allowing us to get rid of https://github.com/spring-projects/spring-security/blob/main/buildSrc/src/main/java/org/springframework/security/convention/versions/VerifyDependenciesVersionsPlugin.java.

With that said, I'll close this as not planned for now and as we mature the idea, we might reopen it in the future.

Issue #14186, opened 2023-11-22
Add support for nested property names in oauth2.providers.userNameAttribute


leeavital:
**Expected Behavior**

I was trying to configure my app to use the pagerduty oauth2 provider, I was using the following configuration:

```
spring:
  security:
    oauth2:
      client:
        registration:
          pagerduty:
            provider: pagerduty
            clientId: <redacted>
            clientSecret: <redacted>
            authorizationGrantType: authorization_code
            redirectUri: "{baseUrl}/login/oauth2/code/{registrationId}"
            clientAuthenticationMethod: client_secret_post

        provider:
          pagerduty:
            authorizationUri: "https://identity.pagerduty.com/oauth/authorize"
            tokenUri: "https://identity.pagerduty.com/oauth/token"
            userInfoUri: "https://api.pagerduty.com/users/me"
            userNameAttribute:  user.email
```

The pagerduty `/me` api returns users with all the interesting properties  nested under the `user` field, like so:

```
{
  "user": {
    "id": ...,
    "email": "fancy@pants.com",
```

I tried setting `userNameAttribute: user.email`  expecting the name field to be extracted as the `name` property on the `user` object.   But  I get an error from DefaultOAuth2User: `"Missing attribute 'user.email' in attributes`.

**Current Behavior**

Ideally a user would successfully be extracted, and login would be successful. 

 
**Context**

I wound up exposing a custom OAuth2UserService class, but it's 90% of a copy paste of `DefaultOAuth2UserService`, and this seems like something that another user info API might reasonably do.

jzheaux:
Hi, @leeavital, thanks for the suggestion. I think this sounds reasonable.

I believe there are three steps that are needed, though, to make this work.

First, `DefaultOAuth2User` should have a new constructor and deprecate the existing one. The new one should look like this:

```java
public DefaultOAuth2User(Map<String, Object> attributes, Collection<? extends GrantedAuthority> authorities, String name)
```

Second, Spring Security should update existing production code (not tests) to use the new constructor.

Third, `DefaultOAuth2UserService` should change to process the userNameAttribute as a simple SpEL expression like so:

```java
SimpleEvaluationContext context = SimpleEvaluationContext
		.forPropertyAccessors(new MapAccessor())
		.withRootObject(userAttributes).build();
SpelExpressionParser parser = new SpelExpressionParser();
Expression expression = parser.parseExpression(userNameAttributeName);
String name = (String) expression.getValue(context);
[...]

PR #14265, opened 2023-12-08
Nested username attribute in DefaultOAuth2User

ahmd-nabil:
ISSUE gh-14186

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -76,6 +76,9 @@ public class DefaultOAuth2UserService implements OAuth2UserService<OAuth2UserReq
 
 	private Converter<OAuth2UserRequest, RequestEntity<?>> requestEntityConverter = new OAuth2UserRequestEntityConverter();
 
+	private Converter<OAuth2UserRequest, Converter<Map<String, Object>, Map<String, Object>>> attributesConverter = (
+			request) -> (attributes) -> attributes;
+
 	private RestOperations restOperations;
 
 	public DefaultOAuth2UserService() {
@@ -108,7 +111,7 @@ public class DefaultOAuth2UserService implements OAuth2UserService<OAuth2UserReq
 		}
 		RequestEntity<?> request = this.requestEntityConverter.convert(userRequest);
 		ResponseEntity<Map<String, Object>> response = getResponse(userRequest, request);
-		Map<String, Object> userAttributes = response.getBody();
+		Map<String, Object> userAttributes = this.attributesConverter.convert(userRequest).convert(response.getBody());
 		Set<GrantedAuthority> authorities = new LinkedHashSet<>();
 		authorities.add(new OAuth2UserAuthority(userAttributes));
 		OAuth2AccessToken token = userRequest.getAccessToken();
@@ -116,6 +119,32 @@ public class DefaultOAuth2UserService implements OAuth2UserService<OAuth2UserReq
 			authorities.add(new SimpleGrantedAuthority("SCOPE_" + authority));
 		}
 		return new DefaultOAuth2User(authorities, userAttributes, userNameAttributeName);
+	}
+
+	/**
+	 * Use this strategy to adapt user attributes into a format understood by Spring
+	 * Security; by default, the original attributes are preserved.
+	 *
+	 * <p>
+	 * This can be helpful, for example, if the user attribute is nested. Since Spring
+	 * Security needs the username attribute to be at the top level, you can use this
+	 * method to do:
+	 *
+	 * <pre>
+	 *     DefaultOAuth2UserService userService = new DefaultOAuth2UserService();
+	 *     userService.setAttributesConverter((userRequest) -> (attributes) ->
+	 *         Map&lt;String, Object&gt; userObject = (Map&lt;String, Object&gt;) attributes.get("user");
+	 *         attributes.put("user-name", userObject.get("user-name"));
+	 *         return attributes;
+	 *     });
+	 * </pre>
+	 * @param attributesConverter the attribute adaptation strategy to use
+	 * @since 6.3
+	 */
+	public void setAttributesConverter(
+			Converter<OAuth2UserRequest, Converter<Map<String, Object>, Map<String, Object>>> attributesConverter) {
+		Assert.notNull(attributesConverter, "attributesConverter cannot be null");
+		this.attributesConverter = attributesConverter;
 	}
 
 	private ResponseEntity<Map<String, Object>> getResponse(OAuth2UserRequest userRequest, RequestEntity<?> request) {
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
@@ -16,9 +16,9 @@
 
 package org.springframework.security.oauth2.client.userinfo;
 
+import java.util.Collection;
 import java.util.LinkedHashSet;
 import java.util.Map;
-import java.util.Set;
 
 import org.springframework.core.ParameterizedTypeReference;
 import org.springframework.core.convert.converter.Converter;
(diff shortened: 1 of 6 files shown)


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
```diff
@@ -157,6 +141,53 @@ private ResponseEntity<Map<String, Object>> getResponse(OAuth2UserRequest userRe
...
+		return authorities;
+	}
+
+	private String getName(Map<String, Object> attributes, String userNameAttributeName) {
+		Assert.notEmpty(attributes, "attributes cannot be empty");
+		Assert.hasText(userNameAttributeName, "userNameAttributeName cannot be empty");
+		SimpleEvaluationContext context = SimpleEvaluationContext.forPropertyAccessors(new MapAccessor())
+			.withRootObject(attributes)
+			.build();
+		SpelExpressionParser parser = new SpelExpressionParser();
```
jzheaux, line 182: Please move this to be a member variable as it doesn't need to be re-constructed each time.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
```diff
@@ -157,6 +141,53 @@ private ResponseEntity<Map<String, Object>> getResponse(OAuth2UserRequest userRe
...
+		String expressionString = Arrays.stream(userNameAttributeName.split("\\."))
```
jzheaux, line 183: I think it's important to rely on the developer to specify the correct expression. Are you able to simply pass in `userNameAttributeName`?

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserServiceTests.java
```diff
@@ -158,6 +158,40 @@ public void loadUserWhenUserInfoSuccessResponseThenReturnUser() {
 		assertThat(userAuthority.getAttributes()).isEqualTo(user.getAttributes());
 	}
 
+	@Test
+	public void loadUserWhenNestedUserInfoSuccessThenReturnUser() {
+		// @formatter:off
+		String userInfoResponse = "{\n"
+				+ "   \"user\": {\"user-name\": \"user1\"},\n"
```
jzheaux, line 165: I believe the reason you are having trouble with the simple dot syntax is since `user-name` is not a valid SpEL accessor. 

Either change the payload to use `username` for the property or use `user['user-name']` for the expression.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/user/DefaultOAuth2User.java
```diff
@@ -57,12 +57,15 @@ public class DefaultOAuth2User implements OAuth2User, Serializable {
...
 
+	private final String name;
+
 	/**
 	 * Constructs a {@code DefaultOAuth2User} using the provided parameters.
 	 * @param authorities the authorities granted to the user
 	 * @param attributes the attributes about the user
 	 * @param nameAttributeKey the key used to access the user's &quot;name&quot; from
 	 * {@link #getAttributes()}
+	 * @deprecated
```
jzheaux, line 68: Please add a description here as to what is the preferred alternative

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/user/DefaultOAuth2User.java
```diff
@@ -57,12 +57,15 @@ public class DefaultOAuth2User implements OAuth2User, Serializable {
...
 	 */
 	public DefaultOAuth2User(Collection<? extends GrantedAuthority> authorities, Map<String, Object> attributes,
```
jzheaux, line 70: Can this constructor be changed to call the new constructor? This may simplify future maintenance.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/jackson2/OAuth2AuthenticationTokenMixinTests.java
```diff
@@ -194,7 +194,8 @@ private static String asJson(DefaultOAuth2User oauth2User) {
 				"      \"@class\": \"java.util.Collections$UnmodifiableMap\",\n" +
 				"      \"username\": \"user\"\n" +
 				"    },\n" +
-				"    \"nameAttributeKey\": \"username\"\n" +
+				"    \"nameAttributeKey\": \"username\",\n" +
```
jzheaux, line 197: I don't think this is representative of how the JSON will look (no more `nameAttributeKey` property). 

The tests as a whole should confirm two things:

1. Values serialized with no `name` property (given that it is new) should still deserialize correctly
2. Values with no `nameAttributeKey` property (given that it is being removed) should still deserialize correctly

I believe this means that there will need to be a change to `DefaultOAuth2UserMixin`; it may need to use a deserializer, like `ClientRegistrationMixin` does. If it contains `nameAttributeKey`, use the deprecated constructor; otherwise, use the new constructor.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/user/DefaultOAuth2User.java
```diff
@@ -57,12 +57,15 @@ public class DefaultOAuth2User implements OAuth2User, Serializable {
 
 	private final String nameAttributeKey;
```
jzheaux, line 58: This should no longer be necessary

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
```diff
@@ -157,6 +141,53 @@ private ResponseEntity<Map<String, Object>> getResponse(OAuth2UserRequest userRe
 		}
 	}
 
+	private String getUserNameAttributeName(OAuth2UserRequest userRequest) {
```
jzheaux, line 144: I appreciate this cleanup. So that the feature commit has only the changes needed for the feature, will you please place these other refactors into a separate, earlier commit in the PR?

jzheaux (changes requested):
Thanks, @ahmd-nabil! I've left some feedback inline.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/user/DefaultOAuth2User.java
ahmd-nabil: @jzheaux, doing so caused `NullPointerException`s, because I am getting the `name` out of `attributes` map before checking for nulls in the new constructor.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/jackson2/OAuth2AuthenticationTokenMixinTests.java
ahmd-nabil: @jzheaux, here after implementing deserializer for `DefaultOAuth2UserMixin`, I had to implement another deserializer `DefaultOidcUserMixin`. seems reasonable ?? just checking if I am in the right direction

ahmd-nabil: @jzheaux, just to make sure I am going the right direction, working on this I had to create 2 deserializers one for `DefaultOAuth2UserMixin` and another for `DefaultOidcUserMixin` seems reasonable ?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
ahmd-nabil: I will do so once everything else is approved

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/user/DefaultOAuth2User.java
ahmd-nabil: another solution would be by introducing an `init` method to encapsulate common logic, but will require all `authorities`, `attributes`, and `name` to be **non final**.

init method would look something like
```
private void init(Map<String, Object> attributes, Collection<? extends GrantedAuthority> authorities,
					  String name) {
        this.authorities = (authorities != null)
		        ? Collections.unmodifiableSet(new LinkedHashSet<>(this.sortAuthorities(authorities)))
		        : Collections.unmodifiableSet(new LinkedHashSet<>(AuthorityUtils.NO_AUTHORITIES));
        this.attributes = Collections.unmodifiableMap(new LinkedHashMap<>(attributes));
        this.name = name;
}
```
old constructor
```
public DefaultOAuth2User(Collection<? extends GrantedAuthority> authorities, Map<String, Object> attributes,
			String nameAttributeKey) {
        Assert.hasText(nameAttributeKey, "nameAttributeKey cannot be empty");
	        throw new IllegalArgumentException("Missing attribute '" + nameAttributeKey + "' in attributes");
        }
        init(attributes, authorities,(String) attributes.get(nameAttributeKey));
}
```
and new constructor
```
public DefaultOAuth2User(Map<String, Object> attributes, Collection<? extends GrantedAuthority> authorities,
			String name) {
        Assert.hasText(name, "name cannot be empty");
        init(attributes, authorities, name);
}
```

ahmd-nabil:
Hi @jzheaux. sorry for rushing you. But, I 've been waiting for a review for a while, so I am just making sure if PR is still needed. whenever you have the time I would appreciate your review :)

jzheaux:
Thanks for the updates, @ahmd-nabil. Some of your recent changes caused me to see that there might not be a way to remain backward compatible. For example, an application using `user-name-attribute: user-name` will break when upgrading.

Further, the JWT spec does not restrict characters for custom claim names, meaning that it is technically possible for `user.username` to be a valid claim name by itself. This makes it rather tricky to know the user's intent by reading the property value.

Given that, I think we might need to take a different route, and hopefully one that is less invasive. To that end, I've pushed a change to [my fork](https://github.com/spring-projects/spring-security/compare/main...jzheaux:spring-security:gh-14186) that takes a different approach. If you like it, you are welcome to merge it to your branch.

ahmd-nabil:
thanks @jzheaux merged. and as always thanks for your guidance :)

jzheaux:
Great, thanks @ahmd-nabil. Are you able to also do the same for `DefaultReactiveOAuth2UserService` and the associated tests? 

If not, I can add that as a polish. Either way, I think we're ready to separate the existing code cleanup into a separate commit and squash the other changes into a single commit.

ahmd-nabil:
TY @jzheaux, I've separated the commit into 2 of them (I had to copy paste a lot to do that, I wonder if there is a better way).
I am not that good with project reactor but I will give it a try.

jzheaux:
Thanks, @ahmd-nabil! This is now merged into `main`.

Issue #14275, opened 2023-12-11
Provide default implementations in UserDetails


jzheaux:
The methods `isAccountNonExpired`, `isAccountNonLocked`, `isCredentialsNonExpired`, and `isEnabled` are used less often.

It would be nice if they were already implemented, returning `true` for each one.

Note that this would align the interface with how `UserBuilder` treats these properties.

PR #14276, opened 2023-12-11
Add default implementations in UserDetails

ahmd-nabil:
Closes gh-14275

### DIFF
--- a/core/src/main/java/org/springframework/security/core/userdetails/UserDetails.java
+++ b/core/src/main/java/org/springframework/security/core/userdetails/UserDetails.java
@@ -67,14 +67,18 @@ public interface UserDetails extends Serializable {
 	 * @return <code>true</code> if the user's account is valid (ie non-expired),
 	 * <code>false</code> if no longer valid (ie expired)
 	 */
-	boolean isAccountNonExpired();
+	default boolean isAccountNonExpired() {
+		return true;
+	}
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/springframework/security/core/userdetails/UserDetails.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2004, 2005, 2006 Acegi Technology Pty Limited
+ * Copyright 2004-2023 Acegi Technology Pty Limited
```
jzheaux, line 2: @ahmd-nabil the Acegi copyright headers are a special case. Will you please leave those as they are?

jzheaux (changes requested):
Thanks, @ahmd-nabil, that was quick! I have just one inline comment for you before merging.

core/src/main/java/org/springframework/security/core/userdetails/UserDetails.java
ahmd-nabil: sure

ahmd-nabil: @jzheaux done, thanks for your great support btw. I am waiting for your review on PR #14265 whenever you have time.

PR #14282, opened 2023-12-12
Consider InMemorySaml2AuthenticationRequestRepository

jzheaux:
I wonder about the fact that the repository could fill up with stale authnrequests and where the balance is with assisting with cleanup.

The PR addresses this by introducing an additional method, specific to the implementation. The attached documentation describes what I'd anticipate to be its usage.


jzheaux:
Waiting first to see how #14297 turns out.

Issue #4961, opened 2018-01-12
when using a ReactiveAuthorizationManager no AuthenticationEvents are published


joshiste:
When I use the reactive WebFilter in a webflux application no AuthenticationEvents are published

rwinch:
Thanks for the report. This is currently blocked by https://github.com/spring-projects/spring-framework/issues/21025

wyaeld:
@rwinch it is extremely frustrating for your users that this just silently fails with no clue as to why.  If you are claiming that Spring Security works for a reactive stack, please update your documentation to be extremely explicit about limitations.  Auditing is considered to be a first class capability of Spring Security.

Given @jhoeller comments on https://jira.spring.io/browse/SPR-16481 can you please clarify if there is a workaround or listener class implementation that someone can do in a Reactive Stack with Spring Security.  I'm currently working on an application where auditing is not optional.

rwinch:
An overdue response to @wyaeld Just because Spring Security does not publish events, doesn't mean you cannot achieve this. You are able to easily plug in custom success/failure handlers which can publish the events.

wyaeld:
@rwinch Appreciate the reply.  Not sure if we are on the same page.  The issue that was a core feature of Spring Security no longer worked, without any documentation, using your reactive implementation.

That's a pretty big issue for someone depending on the reliability of your components.  As a brief glance, it doesn't appear documentation currently indicates any caveats, but since this doesn't appear to have been treated as a major problem.  The handful of linked issues appears to indicate some others are still struggling with it, nearly 3 years after it first reported.

The lack of support or response on this issue forced us to reassess our framework choices, and ultimately select something with a smaller, but more reasonable to understand and maintain implementation of reactive patterns.

rwinch:
Thanks for the reply @wyaeld Sorry for the delayed response. Glad you found something that solved your problem.

marcusdacoregio:
Hi all, I've reworded this ticket to be more clear that it will address the publishing of authorization events on WebFlux.

MarcinAman:
Are there any updates on this issue? I see a PR yet it was closed without any visible outcomes.

PR #14361, opened 2023-12-21
Publish Authorization Events on WebFlux

marcusdacoregio:
Closes gh-4961


marcusdacoregio:
Converting to draft because we still have to figure out how to propagate the reactor context from the publisher to the subscriber

Issue #14352, opened 2023-12-19
Introducing new `TypedAuthentication<T>` Interface - 7.0 Proposal


Crain-32:
*Keeping Original Post for Context
**Expected Behavior**
The [Authentication](https://github.com/spring-projects/spring-security/blob/c19f3d9d0610fcfcfd5ad966b607828264bd58f8/core/src/main/java/org/springframework/security/core/Authentication.java) Interface is core to Spring Security, however there are some genuine pain points with it and the Developer Experience with Spring Security. This suggestion should be seen in line with the efforts of #13266 .
To help improve this I suggest adding a single Generic to `Authentication`,
```java
public interface Authentication<T> extends Serializable {
    	Collection<? extends GrantedAuthority> getAuthorities();
    	
    	Object getCredentials();
    	
    	Object getDetails();
    	// Alternative would be keep Object getPrincipal() and include a new Function
    	T getPrincipal();
    	
    	boolean isAuthenticated();

        void setAuthenticated(boolean isAuthenticated);
}
// New Interface 
public interface UserAuthentication extends Authentication<UserDetails> {
     @Override
     default Collection<? extends GrantedAuthority> getAuthorities() {
         return this.getPrincipal().getAuthorities();
     }
}
// Example of a Future Interface
public interface JwtAuthentication extends Authentication<ImaginarySpringSecurityJwt> {}
```

This is similar to what was mentioned [here](https://github.com/spring-projects/spring-security/issues/11348#issuecomment-1150232726), however taking a more step based approach, and focusing first on the main downstream usage.
To reduce adoption friction I'd also expect the following interface to be created,

`public interface LegacyAuthentication extends Authentication<Object> {}`

**Current Behavior**
Weak Typing with
`Authentication.getPrinciple() -> Object`

**Context**

The `Authentication` Interface is very old, and thus it falls into a hard spot of *needing* improvements, but being so core to the framework that adjusting it will have serious considerations for everyone involved.

[...]

PR #14369, opened 2023-12-22
Closes gh-14352

Crain-32:
The goal of #14352 is to begin introducing a more strongly typed `Authentication` object into the Spring Security Ecosystem. This should hopefully (My fingers are crossed) help slowly reduce some overhead for developers across the board as the general flow of User -> Authentication Object can be made more clear with people being able to see they should put a `UserDetails` Object in the Principal to align with Spring Security's Testing Support, along with being able to see the flow of Types through Spring Security, where now you just see `Object` passed everywhere.

For this PR I left it at adding in the interface and replicating the simple `UsernamePasswordAuthenticationToken` with the new interface, along with creating a new `AbstractUserDetailsAuthentication<?, ?>` which mirrors the `AbstractAuthenticationToken`, but requires the Principal object to be a `UserDetails` implementation. I believe swapping the `UsernamePasswordAuthenticationToken` for the `UsernamePasswordTypedAuthenticationToken<?>` to be a breaking change, and will leave that level of integration to the hands of the larger Spring Team.

Additionally I swapped the `AbstractAuthenticationToken.eraseSecret(String)`'s instance check to be a more modern capture group, as I was already reviewing that entire class.

Last thing to note is at the time of this PR #14352 is still in "waiting-for-triage", but I had a free afternoon and felt like trying this out, figured that might also help triage effort as at least for me, seeing code makes life a lot easier.

### DIFF
--- a/core/src/main/java/org/springframework/security/authentication/UsernamePasswordTypedAuthenticationToken.java
+++ b/core/src/main/java/org/springframework/security/authentication/UsernamePasswordTypedAuthenticationToken.java
@@ -1,0 +1,85 @@
+package org.springframework.security.authentication;
+
+import org.springframework.security.core.SpringSecurityCoreVersion;
+import org.springframework.security.core.userdetails.UserDetails;
+import org.springframework.util.Assert;
+
+/**
+ * A {@link AbstractUserDetailsAuthentication} implementation that is designed for simple
+ * presentation of a username and password.
+ * <p>
+ *
+ * @param <D> The Details Object the Token can map to.
+ * @author Peter Eastham
+ */
+public class UsernamePasswordTypedAuthenticationToken<D> extends AbstractUserDetailsAuthentication<String, D> {
+
+	private static final long serialVersionUID = SpringSecurityCoreVersion.SERIAL_VERSION_UID;
+
+	private String credentials;
+
+	/**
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/springframework/security/authentication/UsernamePasswordTypedAuthenticationToken.java
```diff
@@ -0,0 +1,85 @@
...
+
+	/**
+	 * Creates a token based on a supplied <tt>UserDetails</tt> instance and
+	 * <tt>String</tt> Object.
+	 *
+	 * @param principal a nonnull <tt>UserDetails</tt> for the principal.
+	 * @throws IllegalArgumentException if {@code principal.getAuthorities()} is null, or
+	 *                                  any Authorities in it are null.
+	 */
+	private UsernamePasswordTypedAuthenticationToken(UserDetails principal, String credentials, boolean authenticated) {
```
Crain-32, line 29: Another thing to note. I was very torn on if this should be a public or private constructor. In a review of how `UsernamePasswordAuthenticationToken` is handled, I arrived to the conclusion the maximum visibility should be `protected`, however with no sub-classes or other reasons to expose it, I left it as private.

sjohnr:
Closing this PR for now per [this comment](https://github.com/spring-projects/spring-security/issues/14352#issuecomment-1879132538).

Issue #13625, opened 2023-08-07
AnnotationConfigurationException when using PreAuthorize, CGLIB and EnableMethodSecurity


rawfg:
**Describe the bug**
I perforrmed the [method security migration](https://docs.spring.io/spring-security/reference/servlet/authorization/method-security.html#migration-enableglobalmethodsecurity), to replace deprecated `EnableGlobalMethodSecurity` annotations with the new ` EnableMethodSecurity` annotations, and I've got an exception when using CGLIB  and the `PreAuthorize` annotation.

**To Reproduce**
1. Clone the [sample repository](https://github.com/rawfg/preauth-issue)
2. Run the application
3. Navigate to the http://localhost:8080/hello
4. Provide credentials: _username_: **user**, _password_: **pass**
5. Notice the exception:
```
org.springframework.core.annotation.AnnotationConfigurationException: Found more than one annotation of type interface org.springframework.security.access.prepost.PreAuthorize attributed to public org.springframework.http.ResponseEntity com.example.preauth.SayHelloController.sayHello() Please remove the duplicate annotations and publish a bean to handle your authorization logic.
	at org.springframework.security.authorization.method.AuthorizationAnnotationUtils.findUniqueAnnotation(AuthorizationAnnotationUtils.java:65) ~[spring-security-core-6.1.2.jar:6.1.2]
	at org.springframework.security.authorization.method.PreAuthorizeExpressionAttributeRegistry.findPreAuthorizeAnnotation(PreAuthorizeExpressionAttributeRegistry.java:71) ~[spring-security-core-6.1.2.jar:6.1.2]
```

**Expected behavior**
1. Clone the [sample repository](https://github.com/rawfg/preauth-issue)
2. Comment [line 15](https://github.com/rawfg/preauth-issue/blob/master/src/main/java/com/example/preauth/SecurityConfiguration.java#L15C2-L15C22) in the `SecurityConfiguration` class
3. Uncomment [line 17](https://github.com/rawfg/preauth-issue/blob/master/src/main/java/com/example/preauth/SecurityConfiguration.java#L17) in the `SecurityConfiguration` class
4. Run the application
5. Navigate to the http://localhost:8080/hello
6. Provide credentials: _username_: **user**, _password_: **pass**
7. Application produces _hello_ response

**Sample**

[sample repository](https://github.com/rawfg/preauth-issue)

patrick-sager:
[...]

PR #14407, opened 2024-01-05
Revise `AuthorizationAnnotationUtils`

sbrannen:
## Overview

This commit revises `AuthorizationAnnotationUtils` as follows.

- Removes code duplication by treating both `Class` and `Method` as `AnnotatedElement`.

- Avoids duplicated annotation searches by processing merged annotations in a single Stream instead of first using the `MergedAnnotations` API to find possible duplicates and then again searching for a single annotation via `AnnotationUtils` (which effectively performs the same search using the `MergedAnnotations` API internally).

- Uses `.distinct()` within the Stream to avoid the need for the workaround introduced in gh-13625. Note that the semantics here result in duplicate "equivalent" annotations being ignored. In other words, if `@⁠PreAuthorize("hasRole('someRole')")` is present multiple times as a meta-annotation, no exception will be thrown and the first such annotation found will be used.

- Improves the error message when competing annotations are found by including the competing annotations in the error message.

- Updates `AuthorizationAnnotationUtilsTests` to cover all known, supported use cases.

- Configures correct role in `@⁠RequireUserRole`.

Please note this commit uses `.map(MergedAnnotation::withNonMergedAttributes)` to retain backward compatibility with previous versions of Spring Security. However, that line can be deleted if the Spring Security team decides that it wishes to support merged annotation attributes via custom composed annotations. If that decision is made, the `composedMergedAnnotationsAreNotSupported()` test should be renamed and updated as explained in the comment in that method.

## Related Issues

- #13625
- https://github.com/spring-projects/spring-framework/issues/31803

### DIFF
--- a/core/src/main/java/org/springframework/security/authorization/method/AuthorizationAnnotationUtils.java
+++ b/core/src/main/java/org/springframework/security/authorization/method/AuthorizationAnnotationUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


jzheaux:
Thanks, @sbrannen! This is now merged into `main`.

Issue #14480, opened 2024-01-22
Add Meta-annotation Parameter Support


jzheaux:
It would be nice if Spring Security's method security meta-annotation support allowed for parameters.

For example, it would be nice to be able to do:

```java
@Retention(RetentionPolicy.RUNTIME)
@PreAuthorize("hasAuthority(#value)")
public @interface HasAuthority {
    String value();
}
```

Then, an application could do:

```java
@HasAuthority("message:read")
public String method(...) {
}
```

The annotation expression should be able to handle method parameters, like `@PreAuthorize` already does:

```java
@Retention(RetentionPolicy.RUNTIME)
@PreAuthorize("@myAuthzBean.authorizeParameter(#object)")
public @interface AuthorizeObject {
}
```

allowing an application to do:

```java
@AuthorizeObject
public String method(Object object) {
}
```

If an annotation attribute and a method parameter have the same name, the method parameter value should take priority.

kse-music:
@jzheaux  Can I do it?

jzheaux:
Hi, @kse-music, thanks for the offer. I'm going to take care of this one for the moment since it is in a bit of flux (notice the recent changes in the description).

Please feel free to try out this branch to suggest improvements: https://github.com/jzheaux/spring-security/tree/gh-14480

kse-music:
Oh, according to the previous description, I thought that the meta-annotation parameter was used as an SPEL expression variable, but I didn't expect it used as a placeholder parsed property sources. @jzheaux

jzheaux:
That's right, @kse-music. As I was doing some of my own research, I found that approach to be easier. The reason is that SpEL does not support nested variables, which would be needed in some setups.

Sorry for the confusion. If you are still interested, I'd love to have your help on this PR: https://github.com/spring-projects/spring-security/pull/14494 once I have it out of draft.

sbrannen:
Very interesting idea! 👍 

> The reason is that SpEL does not support nested variables, which would be needed in some setups.

Would you mind providing a concrete example just for the sake of clarity?

jzheaux:
Sure, @sbrannen. In the following arrangement:

```java
@PreAuthorize("hasAuthority(#authority)")
public @interface HasAuthority {
    String authority();
}

// ...

@HasAuthority("#parameter")
public String method(String parameter) {
   // ...
}
```

what happens is if `method("value")` is invoked, the set of variables handed to the evaluation context looks something like:

```
[...]

PR #14494, opened 2024-01-25
Add meta-annotation annotation parameter support

jzheaux:
Closes gh-14480

### DIFF
--- a/core/src/main/java/org/springframework/security/authorization/method/AuthorizationAnnotationUtils.java
+++ b/core/src/main/java/org/springframework/security/authorization/method/AuthorizationAnnotationUtils.java
@@ -19,13 +19,19 @@ package org.springframework.security.authorization.method;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedElement;
 import java.lang.reflect.Method;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
 
 import org.springframework.core.annotation.AnnotationConfigurationException;
 import org.springframework.core.annotation.MergedAnnotation;
 import org.springframework.core.annotation.MergedAnnotations;
 import org.springframework.core.annotation.MergedAnnotations.SearchStrategy;
 import org.springframework.core.annotation.RepeatableContainers;
+import org.springframework.core.convert.support.DefaultConversionService;
+import org.springframework.util.PropertyPlaceholderHelper;
 
 /**
  * A collection of utility methods that check for, and error on, conflicting annotations.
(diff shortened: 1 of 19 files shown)


core/src/main/java/org/springframework/security/authorization/method/AuthorizationAnnotationUtils.java
```diff
@@ -50,6 +56,36 @@
  */
 final class AuthorizationAnnotationUtils {
 
+	static <A extends Annotation> Function<AnnotatedElement, A> withPlaceholderResolver(Class<A> type) {
+		Function<MergedAnnotation<A>, A> map = (mergedAnnotation) -> {
+			A annotation = mergedAnnotation.synthesize();
+			if (mergedAnnotation.getMetaSource() == null) {
+				return annotation;
+			}
```
sbrannen, line 64: I would avoid unnecessarily synthesizing the annotation.

core/src/main/java/org/springframework/security/authorization/method/AuthorizationAnnotationUtils.java
```diff
@@ -50,6 +56,36 @@
...
+			String expression = (String) mergedAnnotation.asMap().get("value");
+			MutablePropertySources propertySources = new MutablePropertySources();
+			Map<String, Object> annotationProperties = mergedAnnotation.getMetaSource().asMap();
+			propertySources.addFirst(new MapPropertySource("annotation", annotationProperties));
+			PropertySourcesPropertyResolver propertyResolver = new PropertySourcesPropertyResolver(propertySources);
+			AnnotatedElement annotatedElement = (AnnotatedElement) mergedAnnotation.getSource();
+			String value = propertyResolver.resolvePlaceholders(expression);
+			return MergedAnnotation.of(annotatedElement, type, Collections.singletonMap("value", value)).synthesize();
```
sbrannen, line 72: I've only glanced at this, but I'm wondering if it might be easier to work with `PropertyPlaceholderHelper` instead.

For example, see the `StaticStringValueResolver` in `spring-test`:

https://github.com/spring-projects/spring-framework/blob/00577ed80acfc22104b41d6a631f433a3fa7e6af/spring-test/src/main/java/org/springframework/test/web/servlet/setup/StandaloneMockMvcBuilder.java#L582-L597

docs/modules/ROOT/pages/servlet/authorization/method-security.adoc
```diff
@@ -900,6 +900,127 @@ open class BankService {
...
+Java::
++
+[source,java,role="primary"]
+----
+@Target({ ElementType.METHOD, ElementType.TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+@PreAuthorize("hasRole('${value}')")
+public @interface HasRole {
+	String value();
+}
```
sbrannen, line 917: I wonder if using property placeholders would confuse users who may assume the placeholders can be sourced from JVM system properties, environment variables, Spring Boot application properties, etc.

Have you considered treating the text as a template containing SpEL expressions which can access the annotation's attributes?

jzheaux: > Have you considered treating the text as a template containing SpEL expressions which can access the annotation's attributes?

One thing that I considered was using SpEL variables; however, I ran into trouble when an annotation-attribute-as-SpEL-variable referred to another SpEL variable:

```java
@PreAuthorize(hasRole(#value))
public @interface HasRole {
    String value();
}

// ...

@HasRole("#parameter")
public String doThis(String parameter) {
    // ...
}
```

I'm not sure of another way that the SpEL expression could access the annotation's attributes that would also allow for the attributes to be SpEL expressions themselves. Can you elaborate?

jzheaux: Alternatively, I could change the delimiter, like so:

```java
@PreAuthorize("hasRole('{value}')")
public @interface HasRole {
	String value();
}
```

so that it matches the syntax for `@RequestMapping` URI templates.

jzheaux:
@kse-music, I've just moved this out of draft. Do you have any feedback you'd like to share?

config/src/main/java/org/springframework/security/config/annotation/method/configuration/EnableMethodSecurity.java
```diff
@@ -97,4 +97,11 @@
 	 */
 	int offset() default 0;
 
+	/**
+	 * Indicate whether to resolve authorization templates with annotation parameters
+	 * @return the resolution policy to use
+	 * @since 6.3
+	 */
+	boolean useAnnotationParameters() default false;
```
rwinch, line 105: I don't think that placing the property here makes sense. The property only enables for pre/post but is present on `@EnableMethodSecurity` which can enable for `@Secured` and JSR 250 annotations.

config/src/main/java/org/springframework/security/config/annotation/method/configuration/EnableReactiveMethodSecurity.java
```diff
@@ -74,4 +74,11 @@
 	 */
 	boolean useAuthorizationManager() default true;
 
+	/**
+	 * Indicate whether to resolve authorization templates with annotation parameters
+	 * @return the resolution policy to use
+	 * @since 6.3
+	 */
+	boolean useAnnotationParameters() default false;
```
rwinch, line 82: I don't think that placing the property here makes sense. The property only enables for pre/post but is present on `@EnableReactiveMethodSecurity` which will likely in the future be similar to `@EnableMethodSecurity` which controls JSR 250 & `@Secured` too

config/src/main/java/org/springframework/security/config/annotation/method/configuration/EnableMethodSecurity.java
rwinch: I'm not sure that useAnnotationParameters makes sense either. Why is the name `Parameters` used? Perhaps `useTemplates`?

core/src/main/java/org/springframework/security/authorization/method/AuthorizationAnnotationUtils.java
```diff
@@ -50,6 +56,41 @@
  */
 final class AuthorizationAnnotationUtils {
 
+	static <A extends Annotation> Function<AnnotatedElement, A> withPlaceholderResolver(Class<A> type) {
+		Function<MergedAnnotation<A>, A> map = (mergedAnnotation) -> {
+			if (mergedAnnotation.getMetaSource() == null) {
+				return mergedAnnotation.synthesize();
+			}
+			PropertyPlaceholderHelper helper = new PropertyPlaceholderHelper("{", "}");
```
rwinch, line 64: It might be nice to have a way to allow users to specify `ignoreUnresolvablePlaceholders = false`. For this reason, a boolean for turning this feature on and off might not be the best approach.

config/src/main/java/org/springframework/security/config/annotation/method/configuration/EnableMethodSecurity.java
jzheaux: I think we can take the annotation setting out for now and leave it as setters for the relevant authorization managers. I'll also update the documentation to reflect that.

Issue #14525, opened 2024-02-01
SAML relying party logout filter is always ordered last


aspan:
When configuring the SAML logout configuration the LogoutFilter is ordered last. It turns out that the `org.springframework.security.config.annotation.web.configurers.saml2.Saml2LogoutConfigurer` configures a standard `org.springframework.security.web.authentication.logout.LogoutFilter` and tries to order it before an already existing `org.springframework.security.web.authentication.logout.LogoutFilter` which doesn't have any effect and the added filter isn't ordered before the existing filter. My solution was to just extend the `org.springframework.security.web.authentication.logout.LogoutFilter` and add that instead so the ordering works.

jzheaux:
Thanks, @aspan! Closing in favor of https://github.com/spring-projects/spring-security/pull/14526

PR #14549, opened 2024-02-05
Saml2 LogoutFilter Should Come Before Common LogoutFilter

aspan:
Closes gh-14525


jzheaux:
Thanks, @aspan! This is now merged into `5.8.x` as well as `6.1.x, `6.2.x`, and `main`.

PR #14649, opened 2024-02-24
Improve JdbcUserDetailsManager.userExists method

Shenker93:
Hello everyone,

This proposal is about tiny enhancements applied to JdbcUserDetailsManager.userExists() method.
As we do not really use an information returned by select query
`select username from users where username = ?`,
and only resultset size is important, my idea is to replace it with an ordinary `count` query.
Some pros of this approach:
* It works faster on DB side and IMO the code is more straightforward in this case, you see what's the purpose just looking on sql
* No need to create unnecessary wrappers on Java side. Currently JdbcTemplate creates a list with a one single purpose - to let the code check its size - and then it goes directly to GC. Not used for iterating over it or any modification/read operations.
* Overloaded version of queryForList() method used there is already marked as  `@Deprecated` as of Spring 5.3 release in favor of queryForList(String, Class, Object...). However that's a separate topic, some other JdbcUserDetailsManager methods also use it, e.g. findUsersInGroup(), that could be resolved in a separate PR to not mix the things up. 

Also, a unit test was added in case if userExists() was invoked with a 'null' parameter. That use-case is not something that really-really should be used as most databases expect `is [not] null` query pattern instead (well, with rare exclusions for null-aware mode support), However as long as this method is designed in this fail-safe manner, let's create a test to be informed about any breaking changes.

### DIFF
--- a/core/src/main/java/org/springframework/security/provisioning/JdbcUserDetailsManager.java
+++ b/core/src/main/java/org/springframework/security/provisioning/JdbcUserDetailsManager.java
@@ -77,7 +77,7 @@ public class JdbcUserDetailsManager extends JdbcDaoImpl implements UserDetailsMa
 
 	public static final String DEF_DELETE_USER_AUTHORITIES_SQL = "delete from authorities where username = ?";
 
-	public static final String DEF_USER_EXISTS_SQL = "select username from users where username = ?";
+	public static final String DEF_USER_EXISTS_SQL = "select count(*) from users where username = ?";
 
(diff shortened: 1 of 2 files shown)


jzheaux:
@Shenker93, I think this is a good idea, though because the value is `public`, we would need to wait for Spring Security 7 to merge this so it can be part of the major-release migration notes.

I've marked the PR accordingly.

rwinch:
Thanks for the PR! I've rebased on the current main and this will be merged as soon as the build passes. Thanks again for taking the time to make Spring Security better :)

Issue #14644, opened 2024-02-22
Support delegating BearerTokenResolver


CrazyParanoid:
Need to add support for retrieving tokens from different headers in one component. This is a common case when a project has multiple authentication schemes. For example, in ProviderManager there are two providers, but one gets the token through DefaultBearerTokenResolver and the other from a non-standard header like -X-Authorization.
For example, component DelegatingBearerTokenResolver:
```
    @Override
    public String resolve(HttpServletRequest request) {
        return delegates.stream()
                .map(d -> d.resolve(request))
                .filter(Objects::nonNull)
                .findAny()
                .orElse(null);
    }
```

CrazyParanoid:
It also makes sense to make a DelegatingServerAuthenticationConverter for different authentication schemes. What’s interesting is there is no way to change the authorizationPattern in the ServerBearerTokenAuthenticationConverter, which prevents this component from being used for non-standard authentication schemes.

jzheaux:
I think a `DelegatingServerAuthenticationConverter` would be nice, @CrazyParanoid. Can you submit a PR for it?

CrazyParanoid:
Added DelegatingBearerTokenResolver. I see that DelegatingServerAuthenticationConverter was added in another PR.

jzheaux:
I'm reopening this since there is more to discuss with out to address the servlet side of this enhancement.

jzheaux:
I don't prefer to add a delegating bearer token resolver as the use case is quite narrow. Instead, I'd prefer to add `DelegatingAuthenticationConverter` so that it can service more use cases. If you are interested a PR is welcome.

I realize that `BearerTokenAuthenticationFilter` doesn't use `AuthenticationConverter` currently, so that doesn't address your concern directly. I believe https://github.com/spring-projects/spring-security/issues/11983 will. In the meantime, you can do:

```java
@Bean 
BearerTokenResolver bearerTokenResolver() {
    BearerTokenResolver one = ...;
    BearerTokenResolver two = ...;
    return (request) -> Optional.ofNullable(one.resolve(request)).orElseGet(() -> two.resolve(request));
}
```

CrazyParanoid:
Hi @jzheaux. Thanks for your feedback! I added support DelegatingAuthenticationConverter to my PR.

CrazyParanoid:
[...]

PR #14654, opened 2024-02-27
Add DelegatingServerAuthenticationConverter

kse-music:
Closes gh-14644

### DIFF
--- a/web/src/main/java/org/springframework/security/web/server/authentication/DelegatingServerAuthenticationConverter.java
+++ b/web/src/main/java/org/springframework/security/web/server/authentication/DelegatingServerAuthenticationConverter.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2002-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.web.server.authentication;
+
+import java.util.List;
+import java.util.function.Function;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+import org.springframework.security.core.Authentication;
+import org.springframework.util.Assert;
+import org.springframework.web.server.ServerWebExchange;
+
+/**
+ * A {@link ServerAuthenticationConverter} that delegates to other
+ * {@link ServerAuthenticationConverter} instances.
+ *
+ * @author DingHao
+ * @since 6.3
+ */
+public final class DelegatingServerAuthenticationConverter implements ServerAuthenticationConverter {
+
+	private final List<ServerAuthenticationConverter> delegates;
+
+	private boolean continueOnError = false;
+
+	private final Log logger = LogFactory.getLog(getClass());
+
+	public DelegatingServerAuthenticationConverter(ServerAuthenticationConverter... converters) {
+		this(List.of(converters));
+	}
+
+	public DelegatingServerAuthenticationConverter(List<ServerAuthenticationConverter> converters) {
+		Assert.notEmpty(converters, "converters cannot be null");
+		this.delegates = converters;
+	}
+
+	@Override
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/springframework/security/web/server/authentication/DelegatingServerAuthenticationConverter.java
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	public DelegatingServerAuthenticationConverter(List<ServerAuthenticationConverter> converters) {
+		Assert.notEmpty(converters, "converters cannot be null");
+		this.delegates = converters;
+	}
+
+	@Override
+	public Mono<Authentication> convert(ServerWebExchange exchange) {
+		return Flux.fromIterable(this.delegates).concatMap((converter) -> converter.convert(exchange)).next();
```
CrazyParanoid, line 50: If the first delegate throws an error, the thread will be terminated. DelegatingServerAuthenticationConverter must iterates all delegates. It seems that you need to use concatMapDelayError.

kse-music: Done

web/src/main/java/org/springframework/security/web/server/authentication/DelegatingServerAuthenticationConverter.java
```diff
@@ -0,0 +1,72 @@
...
+		this(List.of(converters));
+	}
+
+	public DelegatingServerAuthenticationConverter(List<ServerAuthenticationConverter> converters) {
+		Assert.notEmpty(converters, "converters cannot be null");
+		this.delegates = converters;
+	}
+
+	@Override
+	public Mono<Authentication> convert(ServerWebExchange exchange) {
```
CrazyParanoid, line 56: In DelegatingReactiveAuthenticationManager, the continueOnError flag added optional behavior. In the case of DelegatingServerAuthenticationConverter, I think that the initial behavior of the component should be concatMapDelayError, and the continueOnError flag is not needed in this case, that is:
```
	@Override
	public Mono<Authentication> convert(ServerWebExchange exchange) {
		return Flux.fromIterable(delegates)
				.concatMapDelayError(c -> c.convert(exchange)
						.doOnError(logger::debug)
				)
				.next();
	}
```
@jzheaux what do you think about it?

jzheaux: I think that it should fail when there is an error by default. The reason is that Spring Security delegates should return `null` when they don't have anything to do and reserve the right to error if they deem the contents to somehow be insecure.

web/src/main/java/org/springframework/security/web/server/authentication/DelegatingServerAuthenticationConverter.java
```diff
@@ -0,0 +1,72 @@
...
+ * {@link ServerAuthenticationConverter} instances.
+ *
+ * @author DingHao
+ * @since 6.3
+ */
+public class DelegatingServerAuthenticationConverter implements ServerAuthenticationConverter {
+
+	private final List<ServerAuthenticationConverter> delegates;
+
+	private boolean continueOnError = true;
```
jzheaux, line 42: Please have this default to `false`. The reason is two-fold. First please see my comment [here](https://github.com/spring-projects/spring-security/pull/14654/files#r1506834943) and second I think it would be good to keep defaults aligned so that if we change them, they are changed together.

web/src/main/java/org/springframework/security/web/server/authentication/DelegatingServerAuthenticationConverter.java
```diff
@@ -0,0 +1,72 @@
...
+import org.springframework.web.server.ServerWebExchange;
+
+/**
+ * A {@link ServerAuthenticationConverter} that delegates to other
+ * {@link ServerAuthenticationConverter} instances.
+ *
+ * @author DingHao
+ * @since 6.3
+ */
+public class DelegatingServerAuthenticationConverter implements ServerAuthenticationConverter {
```
jzheaux, line 38: Please make this class `final` for now

jzheaux (changes requested):
Thanks, @kse-music! I've left some feedback inline.

web/src/main/java/org/springframework/security/web/server/authentication/DelegatingServerAuthenticationConverter.java
kse-music: Done

web/src/main/java/org/springframework/security/web/server/authentication/DelegatingServerAuthenticationConverter.java
kse-music: Done

kse-music:
Thanks,@jzheaux ,all feedback is done

jzheaux (approved):

Issue #14652, opened 2024-02-26
Provide Native Hints for Beans used in Method Security Annotations


marcusdacoregio:
We should look into how to provide native hints for bean methods used inside Method Security annotations.

Currently, in order to make this work:

```java
@Component
class Authz {
   boolean check(Authentication authentication, String id, String permission) {
      return "admin".equals(authentication.getName());
   }
}

@PreAuthorize("@authz.check(authentication, #id, 'read')")
String findById(String id) {

}
```

We need to register hints:

```java
@Override
public void registerHints(RuntimeHints hints, ClassLoader classLoader) {
	hints.reflection().registerType(Authz.class, MemberCategory.INVOKE_DECLARED_METHODS);
}
```

Related to
 - https://github.com/spring-projects/spring-framework/issues/29548

marcusdacoregio:
@sbrannen provided some insights on how we could achieve that:

- Parse the SpEL expression and then walk the AST to find nodes of type `BeanReference` and then retrieve the `beanName` from that.
- `beanName` is a private field with no getter, we can probably use reflection or parse the bean name from the string returned from `toStringAST()`.
- Once we know the `beanName` we can look up that bean in the `ApplicationContext` and register hints for the bean's concrete type.

sbrannen:
> `beanName` is a private field with no getter, we can probably use reflection or parse the bean name from the string returned from `toStringAST()`.

Please note that `BeanReference` will have a `getName()` method in Spring Framework 6.2.

If it's needed in 6.1.x, we could consider backporting it.

- See https://github.com/spring-projects/spring-framework/issues/32640

PR #14669, opened 2024-02-29
Gh 14597

jzheaux:
Proxy arbitrary objects like so:

```java
public class Flight {
    @PreAuthorize("hasAuthority('airplane:read')")
    Double getAltitude() {
        return 35000d;
    }
}

// ...

var preAuthorize = AuthorizationManagerBeforeMethodInterceptor.preAuthorize();
var factory = new AuthorizationProxyFactory(preAuthorize);
var flight = new Flight();
assertThatNoException().isThrownBy(flight::getAltitude);
Flight proxied = (Flight) factory.proxy(flight);
assertThatExceptionOfType(AccessDeniedException.class).isThrownBy(proxied::getAltitude);
```

For return values of type `Iterator`, `Collection`, `Stream`, `Array`, and `Map`, the values are proxied instead of the collection.

Also when adding `@EnableMethodSecurity`, you can do:

```java
@Component
@AuthorizedResult
public class FlightRepository {
    Iterator<Flights> findAll() {
        return List.of(new Flight()).iterator();
    }
}

// ...
```

And then each `Flight` returned will be proxied.


rwinch:
> For return values of type Iterator, Collection, Stream, Array, and Map, the values are proxied instead of the collection.

We should consider how this works with `@PostFilter` too.

rwinch:
I wonder if we could consider `AuthorizeReturnObject` which
* removes the `d` to align with PreAuthorize and PostAuthorize
* Uses ReturnObject to align with SpEL keyword of returnObject

core/src/test/java/org/springframework/security/authorization/object/AuthorizationProxyFactoryTests.java
```diff
@@ -0,0 +1,147 @@
...
+
+	@Test
+	public void proxyWhenPreAuthorizeThenHonors() {
+		SecurityContextHolder.getContext().setAuthentication(this.user);
+		AuthorizationManagerBeforeMethodInterceptor preAuthorize = AuthorizationManagerBeforeMethodInterceptor
+			.preAuthorize();
+		AuthorizationProxyFactory factory = new AuthorizationProxyFactory(preAuthorize);
+		Flight flight = new Flight();
+		assertThat(flight.getAltitude()).isEqualTo(35000d);
+		Flight secured = (Flight) factory.proxy(flight);
```
marcusdacoregio, line 45: We will have to check if the user need to provide any runtime hints allowing the reflection usage by the proxy in a native-image. An alternative, as part of https://github.com/spring-projects/spring-security/issues/14652, would be to inspect every bean that has `@AuthorizeReturnObject` and register a hint for the return object type.

marcusdacoregio: We can check how `ReflectiveProcessor` implementations works and provide an implementation for our annotation

rwinch (approved):
Thanks @jzheaux This looks good!

core/src/test/java/org/springframework/security/authorization/object/AuthorizationProxyFactoryTests.java
jzheaux: Agreed, let's take a look in #14652

PR #14712, opened 2024-03-08
Add Authorization Denied Handlers for Method Security

marcusdacoregio:
For this feature, we have at least 2 options:

1. Create a `DeniedHandlerMethodInterceptor` that intercepts `AccessDeniedException` thrown from methods annotated with `@DeniedHandler`:
- The `AuthorizationManagerAfterMethodInterceptor` would have to throw a more contextual `AccessDeniedException`, something like `MethodInvocationResultAccessDeniedException` so we can have access to the result object that resulted in an exception to pass it to the `MethodAccessDeniedHandler`.
- When the `@Pre/PostFilter` filters non-collection types, it would have to throw an `AccessDeniedException` for not authorized objects instead of just returning `null` so `DeniedHandlerMethodInterceptor` can catch it.

2. Add the logic inside `AuthorizationManagerAfterMethodInterceptor`, `AuthorizationManagerBeforeMethodInterceptor` and `DefaultMethodSecurityExpressionHandler`:
- The `MethodSecurityEvaluationContext` would have to expose the `Method` in order to access it from `DefaultMethodSecurityExpressionHandler`
- Don't need to throw an exception if the `@DeniedHandler` is present
- Don't need to create contextual `AccessDeniedException`, those classes have access to both `MethodInvocation` and the `result`.

### DIFF
--- a/core/src/main/java/org/springframework/security/authorization/method/AuthorizationManagerAfterMethodInterceptor.java
+++ b/core/src/main/java/org/springframework/security/authorization/method/AuthorizationManagerAfterMethodInterceptor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -54,6 +54,8 @@ public final class AuthorizationManagerAfterMethodInterceptor implements Authori
 	private final Pointcut pointcut;
 
 	private final AuthorizationManager<MethodInvocationResult> authorizationManager;
+
+	private final MethodAuthorizationDeniedPostProcessor defaultPostProcessor = new ThrowingMethodAuthorizationDeniedPostProcessor();
 
 	private int order;
 
@@ -116,8 +118,7 @@ public final class AuthorizationManagerAfterMethodInterceptor implements Authori
 	@Override
 	public Object invoke(MethodInvocation mi) throws Throwable {
 		Object result = mi.proceed();
-		attemptAuthorization(mi, result);
-		return result;
+		return attemptAuthorization(mi, result);
 	}
 
 	@Override
@@ -168,7 +169,7 @@ public final class AuthorizationManagerAfterMethodInterceptor implements Authori
 		this.securityContextHolderStrategy = () -> strategy;
 	}
 
-	private void attemptAuthorization(MethodInvocation mi, Object result) {
+	private Object attemptAuthorization(MethodInvocation mi, Object result) {
 		this.logger.debug(LogMessage.of(() -> "Authorizing method invocation " + mi));
 		MethodInvocationResult object = new MethodInvocationResult(mi, result);
 		AuthorizationDecision decision = this.authorizationManager.check(this::getAuthentication, object);
@@ -176,9 +177,17 @@ public final class AuthorizationManagerAfterMethodInterceptor implements Authori
 		if (decision != null && !decision.isGranted()) {
 			this.logger.debug(LogMessage.of(() -> "Failed to authorize " + mi + " with authorization manager "
 					+ this.authorizationManager + " and decision " + decision));
-			throw new AccessDeniedException("Access Denied");
+			return postProcess(object, decision);
 		}
 		this.logger.debug(LogMessage.of(() -> "Authorized method invocation " + mi));
+		return result;
+	}
+
+	private Object postProcess(MethodInvocationResult mi, AuthorizationDecision decision) {
+		if (decision instanceof MethodAuthorizationDeniedPostProcessor postProcessableDecision) {
+			return postProcessableDecision.postProcessResult(mi, decision);
+		}
+		return this.defaultPostProcessor.postProcessResult(mi, decision);
 	}
 
 	private Authentication getAuthentication() {
--- a/core/src/main/java/org/springframework/security/authorization/method/AuthorizationManagerBeforeReactiveMethodInterceptor.java
+++ b/core/src/main/java/org/springframework/security/authorization/method/AuthorizationManagerBeforeReactiveMethodInterceptor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -32,6 +32,7 @@ import org.springframework.core.MethodParameter;
 import org.springframework.core.ReactiveAdapter;
 import org.springframework.core.ReactiveAdapterRegistry;
 import org.springframework.security.access.prepost.PreAuthorize;
+import org.springframework.security.authorization.AuthorizationDecision;
 import org.springframework.security.authorization.ReactiveAuthorizationManager;
 import org.springframework.security.core.Authentication;
 import org.springframework.util.Assert;
@@ -56,6 +57,8 @@ public final class AuthorizationManagerBeforeReactiveMethodInterceptor implement
 	private final ReactiveAuthorizationManager<MethodInvocation> authorizationManager;
 
 	private int order = AuthorizationInterceptorsOrder.FIRST.getOrder();
+
+	private final MethodAuthorizationDeniedHandler defaultHandler = new ThrowingMethodAuthorizationDeniedHandler();
 
 	/**
 	 * Creates an instance for the {@link PreAuthorize} annotation.
@@ -112,29 +115,63 @@ public final class AuthorizationManagerBeforeReactiveMethodInterceptor implement
 						+ " must return an instance of org.reactivestreams.Publisher "
 						+ "(for example, a Mono or Flux) or the function must be a Kotlin coroutine "
 						+ "in order to support Reactor Context");
-		Mono<Authentication> authentication = ReactiveAuthenticationUtils.getAuthentication();
 		ReactiveAdapter adapter = ReactiveAdapterRegistry.getSharedInstance().getAdapter(type);
-		Mono<Void> preAuthorize = this.authorizationManager.verify(authentication, mi);
 		if (hasFlowReturnType) {
 			if (isSuspendingFunction) {
-				return preAuthorize.thenMany(Flux.defer(() -> ReactiveMethodInvocationUtils.proceed(mi)));
+				return preAuthorized(mi, Flux.defer(() -> ReactiveMethodInvocationUtils.proceed(mi)));
 			}
 			else {
 				Assert.state(adapter != null, () -> "The returnType " + type + " on " + method
 						+ " must have a org.springframework.core.ReactiveAdapter registered");
-				Flux<?> response = preAuthorize
-					.thenMany(Flux.defer(() -> adapter.toPublisher(ReactiveMethodInvocationUtils.proceed(mi))));
+				Flux<Object> response = preAuthorized(mi,
+						Flux.defer(() -> adapter.toPublisher(ReactiveMethodInvocationUtils.proceed(mi))));
 				return KotlinDelegate.asFlow(response);
 			}
 		}
 		if (isMultiValue(type, adapter)) {
-			Publisher<?> publisher = Flux.defer(() -> ReactiveMethodInvocationUtils.proceed(mi));
-			Flux<?> result = preAuthorize.thenMany(publisher);
+			Flux<?> result = preAuthorized(mi, Flux.defer(() -> ReactiveMethodInvocationUtils.proceed(mi)));
 			return (adapter != null) ? adapter.fromPublisher(result) : result;
 		}
-		Mono<?> publisher = Mono.defer(() -> ReactiveMethodInvocationUtils.proceed(mi));
-		Mono<?> result = preAuthorize.then(publisher);
+		Mono<?> result = preAuthorized(mi, Mono.defer(() -> ReactiveMethodInvocationUtils.proceed(mi)));
 		return (adapter != null) ? adapter.fromPublisher(result) : result;
+	}
+
+	private Flux<Object> preAuthorized(MethodInvocation mi, Flux<Object> mapping) {
+		Mono<Authentication> authentication = ReactiveAuthenticationUtils.getAuthentication();
+		return this.authorizationManager.check(authentication, mi)
+			.switchIfEmpty(Mono.just(new AuthorizationDecision(false)))
+			.flatMapMany((decision) -> {
+				if (decision.isGranted()) {
+					return mapping;
+				}
+				return postProcess(decision, mi);
+			});
+	}
+
+	private Mono<Object> preAuthorized(MethodInvocation mi, Mono<Object> mapping) {
+		Mono<Authentication> authentication = ReactiveAuthenticationUtils.getAuthentication();
+		return this.authorizationManager.check(authentication, mi)
+			.switchIfEmpty(Mono.just(new AuthorizationDecision(false)))
+			.flatMap((decision) -> {
+				if (decision.isGranted()) {
+					return mapping;
+				}
+				return postProcess(decision, mi);
+			});
+	}
+
+	private Mono<Object> postProcess(AuthorizationDecision decision, MethodInvocation mi) {
+		return Mono.fromSupplier(() -> {
+			if (decision instanceof MethodAuthorizationDeniedHandler handler) {
+				return handler.handle(mi, decision);
+			}
+			return this.defaultHandler.handle(mi, decision);
+		}).flatMap((result) -> {
+			if (Mono.class.isAssignableFrom(result.getClass())) {
+				return (Mono<?>) result;
+			}
+			return Mono.justOrEmpty(result);
+		});
 	}
 
 	private boolean isMultiValue(Class<?> returnType, ReactiveAdapter adapter) {
--- a/docs/modules/ROOT/pages/servlet/authorization/method-security.adoc
+++ b/docs/modules/ROOT/pages/servlet/authorization/method-security.adoc
@@ -44,6 +44,7 @@ Consider learning about the following use cases:
 * Understanding <<method-security-architecture,how method security works>> and reasons to use it
 * Comparing <<request-vs-method,request-level and method-level authorization>>
 * Authorizing methods with <<use-preauthorize,`@PreAuthorize`>> and <<use-postauthorize,`@PostAuthorize`>>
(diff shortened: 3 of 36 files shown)


core/src/main/java/org/springframework/security/authorization/method/MethodAccessDeniedHandlerResolver.java
marcusdacoregio, line 1: Maybe this could be a `MethodAccessDeniedHandlerFactory` that users would expose as a bean so we do not have to be concerned

core/src/main/java/org/springframework/security/authorization/method/MethodAccessDeniedHandler.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.authorization.method;
+
+import org.springframework.lang.Nullable;
+import org.springframework.security.authorization.AuthorizationDecision;
+
+public interface MethodAccessDeniedHandler {
```
jzheaux, line 22: Have you already considered making this a generic class, like `AuthorizationManager`? I think that I'd personally prefer the type safety of:

```java
Object handle(MethodInvocation mi, AuthorizationDecision decision) { 
    ....
}
```

in my implementation.

core/src/main/java/org/springframework/security/authorization/method/AuthorizationManagerAfterMethodInterceptor.java
```diff
@@ -172,17 +174,30 @@ public void setSecurityContextHolderStrategy(SecurityContextHolderStrategy strat
...
 		this.logger.debug(LogMessage.of(() -> "Authorized method invocation " + mi));
+		return result;
+	}
+
+	private Object handleOrThrow(MethodInvocation mi, Object result, AuthorizationDecision decision) {
+		MethodAccessDeniedHandler deniedHandler = this.accessDeniedHandlerResolver.resolve(mi.getMethod());
+		if (deniedHandler == null) {
+			throw new AccessDeniedException("Access Denied");
+		}
+		return deniedHandler.handle(result, decision);
```
jzheaux, line 200: Is there a reason to only pass the result instead of `MethodInvocationResult`? It seems a little asymmetric that an application wouldn't have access to the method invocation in both cases.

core/src/main/java/org/springframework/security/authorization/method/DeniedHandler.java
```diff
@@ -0,0 +1,34 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ ElementType.METHOD, ElementType.TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+@Inherited
+@Documented
+public @interface DeniedHandler {
+
+	Class<? extends MethodAccessDeniedHandler> value();
```
jzheaux, line 32: There may be value in listing a default here, like `default NullMethodAccessDeniedHandler.class`

core/src/main/java/org/springframework/security/authorization/method/PreFilterAuthorizationMethodInterceptor.java
```diff
@@ -137,10 +147,24 @@ public Object invoke(MethodInvocation mi) throws Throwable {
 		MethodSecurityExpressionHandler expressionHandler = this.registry.getExpressionHandler();
 		EvaluationContext ctx = expressionHandler.createEvaluationContext(this::getAuthentication, mi);
 		Object filterTarget = findFilterTarget(attribute.getFilterTarget(), ctx, mi);
-		expressionHandler.filter(filterTarget, attribute.getExpression(), ctx);
+		Object filteredObject = expressionHandler.filter(filterTarget, attribute.getExpression(), ctx);
+		replaceFilterTargetIfNeeded(attribute.getFilterTarget(), ctx, filteredObject, mi);
```
jzheaux, line 151: I'm a little nervous about this, given the obvious historical efforts to not replace the method parameter (which is why the collection filter clears the existing collection instead of creating a new one. @rwinch can you fill me in on the reason that was done that way?

core/src/main/java/org/springframework/security/authorization/method/MethodAccessDeniedHandlerResolver.java
```diff
@@ -0,0 +1,74 @@
...
+	private DeniedHandler findDeniedHandler(Method method) {
+		DeniedHandler methodAnnotation = AuthorizationAnnotationUtils.findUniqueAnnotation(method, DeniedHandler.class);
+		if (methodAnnotation != null) {
+			return methodAnnotation;
+		}
+		return AuthorizationAnnotationUtils.findUniqueAnnotation(method.getDeclaringClass(), DeniedHandler.class);
+	}
+
+	private MethodAccessDeniedHandler getBeanOrNew(Class<? extends MethodAccessDeniedHandler> handlerClass) {
+		if (this.applicationContext != null && this.applicationContext.getBeanNamesForType(handlerClass).length == 1) {
```
jzheaux, line 54: I'm not so sure about inferring something like this. It makes it a little tricky if I want some methods to use a managed bean when others don't, especially when those are in same the type hierarchy.

For example, if I have something like this:

```java
public class MyAccessDeniedHandler implements MethodAccessDeniedHandler {
    // ...
}
```

that is not a bean, and then

```java
@Component 
public class MyComponentAccessDeniedHandler extends MyAccessDeniedHandler {
    // ...
]
```

then when doing the following:

```java
@DeniedHandler(MyAccessDeniedHandler.class)
```

I believe this resolver could select the incorrect handler.

Instead, I think resolving by bean could either wait for a future enhancement, or perhaps the annotation could instead have two properties, say `handlerClass` and `handlerBeanName`, in order to make the intent clear.

jzheaux (changes requested):
Thanks for putting this together, @marcusdacoregio! I've left some feedback inline.

core/src/main/java/org/springframework/security/authorization/method/MethodAccessDeniedHandlerResolver.java
rwinch: While I understand the concern, I worry about deviating from the semantics of how `Autowired` behaves. A configuration of `@Autowired MyAccessDeniedHandler` would also resolve to `MyComponentAcccessDeniedHandler`. I think changing these semantics is a bit surprising to users. Differences like this would make it more difficult to use.

A benefit of this approach is that if `@DeniedHandler` defaults to `MethodAccessDeniedHandler` then a default Bean of `NullMethodAccessDeniedHandler` could be provided. Users wanting to override that default could provide their own `MethodAccessDeniedHandler` bean.

core/src/main/java/org/springframework/security/authorization/method/PreFilterAuthorizationMethodInterceptor.java
rwinch: Believe it or not the clearing of collections happened before I was a part of the team, so I'm not sure the reasoning. I think the reason was to ensure that the same type that was passed in was used. It could also be because users expected references to the argument to be updated (which I understand to some degree, but I generally don't like mutating inputs)

core/src/main/java/org/springframework/security/util/ThrowableAnalyzer.java
marcusdacoregio, line 1: This class has been copied from `spring-security-web`. We would probably want to deprecate it in `spring-security-web` and recommend usages to use the class from `spring-security-core`

marcusdacoregio:
@rwinch @jzheaux I've pushed a new commit that uses a `HandleAccessDeniedMethodInterceptor` to handle denied exceptions thrown from the method security annotations. This won't work with `@PreFilter` and `@PostFilter` because they do not throw such exceptions and I don't know if changing its semantics right now would be the best choice.

One concern is that the `MethodAccessDeniedHandler` is using the `Object` type for `deniedObject`, this will make users have to be aware of the possible types and it forces the use of the `instanceof` operator. To aliviate the problem, I provided `AbstractMethodInvocationAccessDeniedHandler` with two type-safe methods that can be overriden. Any suggestions how can this be improved?

Another concern is that exceptions can be expensive and it can have a significant impact on application performance (however I haven't done any benchmarking yet). Consider the following class:
```java
class User {

  // ...

  @PreAuthorize("hasRole('ADMIN')")
  @HandleAccessDenied(FirstNameFallbackHandler.class)
  public String getName() {
    return this.name;
  }

  @PreAuthorize("hasRole('ADMIN')")
  @HandleAccessDenied
  public String getAddress() {
    return this.address;
  }

  @PreAuthorize("hasRole('ADMIN')")
  @HandleAccessDenied
  public String getPhoneNumber() {
    return this.phoneNumber;
  }
 
}
```

If we were to use those getters, 3 access denied exceptions would have to be constructed (if it is not an admin) just to be handled by `HandleAccessDeniedMethodInterceptor`. For me, it sounds better not to create an exception and handle the access denied at the moment we validate the security expression. What do you think?

marcusdacoregio:
After talking to @rwinch and @jzheaux we decided to support filtering non-collection types only in `@PreAuthorize` and `@PostAuthorize` for now, this have some advatanges:

- Allow us to know exactly if the access denied happened on the annotation expression itself
- An exception is not instantiated if there is a handler (an exception can still be thrown by the handler)
- Better use of generics by accepting `MethodAccessDeniedHandler<MethodInvocation>` for `@PreAuthorize` and `MethodAccessDeniedHandler<MethodInvocationResult>` for `@PostAuthorize`
- Since Method Security now supports Meta Annotations (#14480), users can create their own annotation and do not depend on a Spring Security on their business code.

core/src/main/java/org/springframework/security/authorization/method/MethodAccessDeniedHandler.java
```diff
@@ -0,0 +1,28 @@
...
+package org.springframework.security.authorization.method;
+
+import org.springframework.lang.Nullable;
+import org.springframework.security.access.AccessDeniedException;
+import org.springframework.security.authorization.AuthorizationDecision;
+
+public interface MethodAccessDeniedHandler<T> {
+
+	@Nullable
+	Object handle(T deniedObject, AuthorizationDecision decision);
```
rwinch, line 26: Consider a method name that makes it more clear that it can replace the deniedObject. I wonder if this is more of a post processor of sorts in which the class name and the method name could be updated to include postProcess. We'd want the interface documented that it is allowed to throw an AccessDeniedException (likely though it would prefer a `DecisionAwareAccessDeniedException` since there is an `AuthorizationDecision` available)

rwinch (changes requested):
Thanks @marcusdacoregio This is really taking shape! I've provided some comments below for your consideration

core/src/main/java/org/springframework/security/authorization/DecisionAwareAccessDeniedException.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.authorization;
+
+import org.springframework.security.access.AccessDeniedException;
+import org.springframework.util.Assert;
+
+public class DecisionAwareAccessDeniedException extends AccessDeniedException {
```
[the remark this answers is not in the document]
jzheaux: Or `AuthorizationDecisionException`. `AuthenticationException` implies that authentication failed, e.g. `AuthenticationFailedException` is a little redundant.

marcusdacoregio:
@rwinch @jzheaux the PR has been updated based on our last discussion.

- Rob mentioned that it would be nice if we could create a new interface as a replacement for `AuthorizationDecision`, currently it is called `AuthorizationResult`, so we can have all the benefits of using an interface instead of a class in the APIs.
- `MethodAccessDeniedHandler` has been renamed to `AuthorizationDeniedPostProcessor`.
- `PreAuthorizeExpressionAttributeRegistry` and `PostAuthorizeExpressionAttributeRegistry` create an `ExpressionAttribute` containing the `postProcessorClass` instead of resolving it in the method interceptors.
- The [`PostProcessableAuthorizationDecision`](https://github.com/spring-projects/spring-security/pull/14712/files#diff-77024f2a8ccb6d5aa830a2170728c07c126c390508ed2cead4f765d94113947b) implements `ResolvableTypeProvider` that allow us to verify its generics at runtime, so we can know whether the post processor class can post process the type that we expect. See the logic [here](https://github.com/spring-projects/spring-security/pull/14712/files#diff-b0269e47cb36525b5fd4c129cd5bed17b9d4a24af7e56f61cf2af7505e4f9fbaR205-R216). I could use some ideas on how to improve that.

rwinch (changes requested):
I've provided feedback below

core/src/main/java/org/springframework/security/authorization/method/DefaultPostInvocationAuthorizationDeniedPostProcessor.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.security.authorization.method;
+
+import org.springframework.security.authorization.AuthorizationException;
+import org.springframework.security.authorization.AuthorizationResult;
+
+public class DefaultPostInvocationAuthorizationDeniedPostProcessor
+		implements AuthorizationDeniedPostProcessor<MethodInvocationResult> {
```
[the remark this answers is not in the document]
marcusdacoregio: I don’t think the class can be generic because it is used inside the PreAuthorize and PostAuthorize classes and annotation’s attributes must be constants. We need an actual implementation of the concrete type to use

docs/modules/ROOT/pages/servlet/authorization/method-security.adoc
```diff
@@ -51,6 +51,7 @@ Consider learning about the following use cases:
 * Coordinating with <<changing-the-order,@Transactional and other AOP-based annotations>>
 * Customizing <<customizing-expression-handling,SpEL expression handling>>
 * Integrating with <<custom-authorization-managers,custom authorization systems>>
+* Providing <<fallback-values-authorization-denied,fallback values when authorization is denied>>
```
jzheaux, line 54: This is likely more valuable to readers than its position in the list. I think it likely belongs under "Authorizing methods with `@PreFilter` and `@PostFilter`". Up to you, though.

docs/modules/ROOT/pages/servlet/authorization/method-security.adoc
```diff
@@ -2208,6 +2209,394 @@ And if they do have that authority, they'll see:
 You can also add the Spring Boot property `spring.jackson.default-property-inclusion=non_null` to exclude the null value, if you also don't want to reveal the JSON key to an unauthorized user.
 ====
 
+[[fallback-values-authorization-denied]]
+== Providing Fallback Values When Authorization is Denied
+
+There are some scenarios where you may not wish to throw an `AccessDeniedException` when a method is invoked without the required permissions.
+Instead, you might wish to return a post-processed result, like a masked result, or a default value in cases where access denied happened before invoking the method.
+
+Spring Security provides support for handling and post-processing method access denied with the `@PreAuthorize` and `@PostAuthorize` annotations respectively.
```
jzheaux, line 2218: I'd recommend linking back to earlier sections that cover `@PreAuthorize` and `@PostAuthorize` fundamentals.

docs/modules/ROOT/pages/servlet/authorization/method-security.adoc
```diff
@@ -2208,6 +2209,394 @@ And if they do have that authority, they'll see:
...
+class SecurityConfig {
+
+    @Bean
+    fun emailMaskingMethodAuthorizationDeniedPostProcessor(): EmailMaskingMethodAuthorizationDeniedPostProcessor {
+        return EmailMaskingMethodAuthorizationDeniedPostProcessor()
+    }
+
+}
+
+class User (val name:String, @PostAuthorize(value = "hasAuthority('user:read')", postProcessorClass = EmailMaskingMethodAuthorizationDeniedPostProcessor::class) val email:String) <3>
```
jzheaux, line 2391: Expressions like this are long enough that it would be valuable to emphasize (perhaps with a code snippet) how meta-annotation support can help here.

jzheaux (changes requested):
This looks awesome, @marcusdacoregio! I've left some minor feedback inline.

core/src/main/java/org/springframework/security/authorization/method/AuthorizationManagerBeforeReactiveMethodInterceptor.java
```diff
@@ -57,6 +58,8 @@ public final class AuthorizationManagerBeforeReactiveMethodInterceptor implement
 
 	private int order = AuthorizationInterceptorsOrder.FIRST.getOrder();
 
+	private final MethodAuthorizationDeniedHandler defaultHandler = new ThrowingMethodAuthorizationDeniedHandler();
```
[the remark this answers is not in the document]
marcusdacoregio: That is not needed because we are using `Mono.fromSupplier`, the `MonoSupplier` already catches the exception and calls `onError`, see [here](https://github.com/reactor/reactor-core/blob/main/reactor-core/src/main/java/reactor/core/publisher/MonoSupplier.java#L128-L136) and [here](https://github.com/reactor/reactor-core/blob/main/reactor-core/src/test/java/reactor/core/publisher/MonoSupplierTest.java#L75-L80).

core/src/main/java/org/springframework/security/authorization/method/PostProcessableAuthorizationDecision.java
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * An {@link org.springframework.security.authorization.AuthorizationDecision} that post
+ * process an {@link AuthorizationResult} by invoking the
+ * {@link AuthorizationDeniedPostProcessor} passed via constructor.
+ *
+ * @param <T> the type of the object used in post-processing
+ * @author Marcus da Coregio
+ * @since 6.3
+ */
+class PostProcessableAuthorizationDecision<T> extends ExpressionAuthorizationDecision {
```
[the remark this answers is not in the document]
marcusdacoregio: I believe we don't need this anymore since we have `PostAuthorizeAuthorizationDecision` and `PreAuthorizeAuthorizationDecision`, both package-private.

core/src/main/java/org/springframework/security/authorization/method/MethodAccessDeniedHandlerResolver.java
jzheaux: Agreed, my point is mainly that the two approaches are incompatible and that we should choose one or the other. I agree that bean-based is the most reasonable default behavior.

jzheaux (approved):

PR #14714, opened 2024-03-10
Use relative URLs in /login redirects

Pistolnik:
## TL;DR
Closes https://github.com/spring-projects/spring-security/issues/7273

## Motivation
I noticed inconsistency in all tests that are using `andExpect(MockMvcResultMatchers.redirectedUrl("/xyz"))`. I was able to test everything using relative urls with one exception - `/login` - absolute paths had to be used there. Same pattern is in the whole repo - `/login` redirect asserts are using `http://localhost/login` (or similar). 
`LoginUrlAuthenticationEntryPoint` is responsible - namely `buildRedirectUrlToLoginPage` that always constructs absolute path. 

There are probably no other than historical reasons for this behavior (as mentioned in https://github.com/spring-projects/spring-security/issues/7273). At time of introducing the the class, [absolute URI was required by RFC 2616](https://www.rfc-editor.org/rfc/rfc2616#section-14.30). The RFC is obsoleted by RFC 7231 since 2014. RFC 7231 [allows relative URIs](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.2) (current RFC 9110 also [allows relative URIs](https://datatracker.ietf.org/doc/html/rfc9110#name-location)).

Expected behavior should be that `LoginUrlAuthenticationEntryPoint` would override relative path to absolute only in cases where it's necessary. That is when `forceHttps` is set to true and incoming requests are `http`.

## Change
After this change, `LoginUrlAuthenticationEntryPoint` will use relative paths if possible (http -> http, https -> https would use relative paths). However if `forceHttps` is set to true, and incoming request is `http`, it'll be redirected using absolute path (`https://example.com/login`).

## More complicated flows
When incoming request is `http`, `forceHttps` is `true`, but `portResolver` is not able to resolve ports correctly: previous behavior was to fall back to redirect to `http` (using absolute path) and log a warning. After the change, the warning is still logged, but relative path will be used.

Potentially problematic is one corner case - when we have incoming `https` request, but using a "matching" `http` port (not `https`!). Let's say we have `8080` for `http` and `8443` for `https` - what should the behavior be for `https://example.com:8080/login`? Previously app redirected to `https://example.com:8443/login` - but only in case when a port "matching" `http` port was used. If different port was used (e.g. 8081), app would redirect to original `https://example.com:8081/login`. I think it was unwanted side effect. After the change, the special behavior is removed. It does not matter if port "matches" `http` when accessing via `https`. Let me know if I should change this.

## Tests
As a result of the small change, many asserts in many tests had to be updated. Behavior is covered by tests (e.g. `testHttpsOperationFromOriginalHttpUrl` `testHttpsOperationFromOriginalHttpsUrl`, `testOperationWhenHttpsRequestsButHttpsPortUnknown` in `LoginUrlAuthenticationEntryPointTests`).
I only updated `testHttpsOperationFromOriginalHttpsUrl` to cover the corner case flow explicitly.

### DIFF
--- a/config/src/test/java/org/springframework/security/config/annotation/web/builders/NamespaceHttpTests.java
+++ b/config/src/test/java/org/springframework/security/config/annotation/web/builders/NamespaceHttpTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2024 the original author or authors.
+ * Copyright 2002-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -77,7 +77,7 @@ import static org.springframework.security.test.web.servlet.request.SecurityMock
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.forwardedUrl;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.header;
-import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.redirectedUrl;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.redirectedUrlPattern;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 
 /**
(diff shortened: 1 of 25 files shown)


jzheaux:
Thanks for the PR, @Pistolnik! I added a polish to place this feature behind a setting until the next major release when we can change the defaults. This will allow the code to remain passive.

When we being development for Spring Security 7 next year, I can reapply the changes you made to the test cases.

PR #14764, opened 2024-03-15
Improve logging in AuthenticationWebFilter

### LINKED ISSUE #14091: Improve logging in NimbusReactiveJwtDecoder
**Catched exception should be logged before JwtException is thrown**

Before a JwtException is thrown, the catched errorMsg should logged on an appropriate level. In order to quickly be able to identify the root cause in the event of a problem.

**Current Behavior**
Plugin: 
- org.springframework.boot: 3.1.3

Artifacts: 
- org.springframework.boot:spring-boot-starter-webflux:3.1.3
- org.springframework.security:spring-security-oauth2-jose:6.1.3

It simply throws:
`throw new BadJwtException("Failed to validate the token", ex);` which is used for logging in AuthenticationWebFilter at DEBUG level only

In a real world application, proper logging to monitor the application state is crucial. At the moment only a verbose message "Failed to validate the token" is logged at DEBUG level, which is not really used in prod profiles? Also the errorMsg doesn't provide any more information than the 401 Unauthorized response code that is returned regardless.

**Context**
After I got my NimbusReactiveJwtDecoder to work with jwks validation. I wanted to see how the implementation handles invalid scenarios, logging etc. And to my surprise, there is in practice no logging at all. There is a lot of exceptions carrying different messages, but eventually they all end up in the WebFilter as: "Authentication failed: Failed to validate the token".

Is it possible that the NimbusReactiveJwtDecoder actually log the catched exceptions?
One practical example is when the jwks public key signature is a mismatch from the private key used to sign the JWT that is used as Authorization Bearer:

[Here](https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusReactiveJwtDecoder.java#L298) you can see how the BadJWSException is thrown when the signature is invalid. Originally thrown from DefaultJWTProcessor in com.nimbusds.jwt.proc

image

image

This flow makes it really hard to use in production environment, because if any of these JwtExceptions would occur (due to invalid Jwt signatures being sent etc.), it would be of great value to see the reason for the actual exception, and preferably not at debug level. Either some bean config that enables additional logging or some other smart way would be hugely beneficial to have.

I'm more than happy to answer additional questions :)

I wish all of you to have a lovely week.
Sincerely, Thomas

ThomasHagelberg:
Closes #14091

### DIFF
--- a/web/src/main/java/org/springframework/security/web/server/authentication/AuthenticationWebFilter.java
+++ b/web/src/main/java/org/springframework/security/web/server/authentication/AuthenticationWebFilter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


jzheaux:
Thanks, @ThomasHagelberg! This is now merged into `main`.

jzheaux:
I've also backported to `6.1.x` and `6.2.x`, so you should see it in the next maintenance release.

Issue #14077, opened 2023-10-31
ConcurrentSessionFilter breaks permitAll endpoint on session expiration


arvyy:
I expect to always be able to invoke endpoints marked with permitAll. However, ConcurrentSessionFilter early aborts such requests at the edge of expiration. https://github.com/spring-projects/spring-security/blob/ed6ff670d102736eea0ac360921c9015151ac630/web/src/main/java/org/springframework/security/web/session/ConcurrentSessionFilter.java#L145  and there doesn't seem a way to sidestep this early return through SessionInformationExpiredStrategy because it doesn't expose FilterChain to call doFilter on.

marcusdacoregio:
Hi, @arvyy,  thanks for the report.

`permitAll` means that it will allow requests with/without authentication to go through, but Spring Security filters will continue doing what they do. Therefore, since you have concurrent session control, if you perform a request with a session that has expired by concurrency, the `ConcurrentSessionFilter` won't allow that request to go through.

Can you elaborate more on your use case? How does the session expire? What are those requests that you always want to go through?

arvyy:
I have a /currentUser endpoint that I want to work regardless if user is logged in or not (in latter case it'd return null). On frontend I create a timer until the end of the session (which I know from an auxiliary cookie). At the end of the timer, I call /currentUser, and if it weren't null (because user was browsing in different tab), I restart the timer. If it was null, I know user has been idle too long and change frontend view to the login screen. This is currently broken, because frontend instead of receiving a json response of an expected form, it instead receives a plain text message "This session has been expired ...".

I understand that permitAll still invokes security filters, and that's desired, because I still want to be able to fetch (just potentially empty) security context's user authentication value. My problem isn't with design as a whole, but with inflexibility of ConcurrentSessionFilter specifically.

marcusdacoregio:
It seems to me that your `/currentUser` endpoint should require authentication and instead of returning `null` when the user is not authenticated, it should return a 401. This way you can provide a `SessionInformationExpiredStrategy` implementation that returns 401 and your frontend code can be adapted to expect it. 

But, if you do not want to do that and instead return `null`, you can provide a `SessionInformationExpiredStrategy` implementation that writes nothing to the response. Does that make sense?

arvyy:
[...]

PR #14765, opened 2024-03-16
Add ContinueRequestSessionInformationExpiredStrategy

Ilpyo-Yang:
### Description
This PR adds a ContinueRequestSessionInformationExpiredStrategy to contiue next filters even though the session expired.

### Changes
+ Add ContinueRequestSessionInformationExpiredStrategy
+ Add testcode

Issue #14077

### DIFF
--- a/web/src/main/java/org/springframework/security/web/session/ContinueRequestSessionInformationExpiredStrategy.java
+++ b/web/src/main/java/org/springframework/security/web/session/ContinueRequestSessionInformationExpiredStrategy.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.web.session;
+
+import jakarta.servlet.ServletException;
+import java.io.IOException;
+
+/**
+ * A {@link SessionInformationExpiredStrategy} continues to execute subsequent filters
+ * even after session expiration.
+ *
+ * @author Rosie Yang
+ * @since 6.3
+ */
+public final class ContinueRequestSessionInformationExpiredStrategy implements SessionInformationExpiredStrategy {
+
+	@Override
+	public void onExpiredSessionDetected(SessionInformationExpiredEvent event) throws ServletException, IOException {
+		event.getFilterChain().doFilter(event.getRequest(), event.getResponse());
+	}
+
+}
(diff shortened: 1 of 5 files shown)


web/src/main/java/org/springframework/security/web/session/ContinueRequestSessionInformationExpiredStrategy.java
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright 2002-2016 the original author or authors.
```
marcusdacoregio, line 2: Can you make sure that the copyright notice includes the current year?

web/src/main/java/org/springframework/security/web/session/ContinueRequestSessionInformationExpiredStrategy.java
```diff
@@ -0,0 +1,35 @@
...
+package org.springframework.security.web.session;
+
+import jakarta.servlet.ServletException;
+import java.io.IOException;
+
+/**
+ * A {@link SessionInformationExpiredStrategy} continues to execute subsequent filters
+ * even after session expiration.
+ *
+ * @author Rosie Yang
```
marcusdacoregio, line 26: Please add `@since 6.3` too

web/src/main/java/org/springframework/security/web/session/SessionInformationExpiredEvent.java
```diff
@@ -35,19 +36,22 @@ public final class SessionInformationExpiredEvent extends ApplicationEvent {
...
+
 	/**
 	 * Creates a new instance
 	 * @param sessionInformation the SessionInformation that is expired
 	 * @param request the HttpServletRequest
 	 * @param response the HttpServletResponse
 	 */
 	public SessionInformationExpiredEvent(SessionInformation sessionInformation, HttpServletRequest request,
-			HttpServletResponse response) {
+			HttpServletResponse response, FilterChain filterChain) {
```
marcusdacoregio, line 48: Instead of changing this constructor, a new constructor should be created accepting all those parameter + the `FilterChain`. The reason is because if we change this constructor it will be a breaking change for anyone that is using `SessionInformationExpiredEvent`.

marcusdacoregio (changes requested):

web/src/main/java/org/springframework/security/web/session/ContinueRequestSessionInformationExpiredStrategy.java
Ilpyo-Yang: Thanks! I changed

web/src/main/java/org/springframework/security/web/session/ContinueRequestSessionInformationExpiredStrategy.java
Ilpyo-Yang: Thanks! I changed too. I realized I need to pay more attention to contribution rules, sorry about that.

web/src/main/java/org/springframework/security/web/session/SessionInformationExpiredEvent.java
Ilpyo-Yang: As you mentioned, I add new constructor below original constructor.  But filterChain value need for new constructor i had to handle this in original constructor too.

marcusdacoregio:
Closing as fixed via https://github.com/spring-projects/spring-security/commit/095929f6e828c01fc6c4893b0b1b9bfc9a77d888

PR #14790, opened 2024-03-21
Provide Runtime Hints for Beans used in Pre/PostAuthorize Expressions

marcusdacoregio:
Closes gh-14652

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/method/configuration/PrePostAuthorizeBeanFactoryInitializationAotProcessor.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/method/configuration/PrePostAuthorizeBeanFactoryInitializationAotProcessor.java
@@ -1,0 +1,199 @@
+/*
+ * Copyright 2002-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


config/src/main/java/org/springframework/security/config/annotation/method/configuration/PrePostAuthorizeBeanFactoryInitializationAotProcessor.java
```diff
@@ -0,0 +1,199 @@
...
+
+			return switch (annotations.size()) {
+				case 0 -> null;
+				case 1 -> annotations.get(0);
+				default -> throw new AnnotationConfigurationException("""
+						Please ensure there is one unique annotation of type @%s attributed to %s. \
+						Found %d competing annotations: %s""".formatted(annotationType.getName(), annotatedElement,
+						annotations.size(), annotations));
+			};
+		}
```
marcusdacoregio, line 195: @jzheaux This was copied from `AuthorizationAnnotationUtils`. My concern is that this logic can change and then we have to remember to update it on multiple places. Any idea on how we can reuse the same logic?

marcusdacoregio:
As per @jzheaux comments: it's not always clear from the bean factory which type it is that we should be proxying, like in the case of Spring Data, which publishes a factory bean (so the interface that has the annotations isn't yet known).
Another challenge ls that with these changes it will be possible and even likely that some classes won't be a bean. (Consider `CrudRepositoyr#findAll` which returns possibly-annotated domain objects).

With that said, we might have to adopt another strategy to find the components to register hints, maybe by doing a component scan.

marcusdacoregio:
I'll close this for now for the following reasons:

- There are some corner cases that we are not yet sure what the approach should be, like those mentioned [here](https://github.com/spring-projects/spring-security/pull/14790#issuecomment-2017969579).
- The support won't be complete, for example, we will provide runtime hints only for SpEL inside `@PreAuthorize` and `@PostAuthorize`, however, there are more annotations that allows usage of SpEL and it might be surprising that those are not supported.
- As of now, users can use `@RegisterReflectionForBinding` as described in https://docs.spring.io/spring-security/reference/native-image/method-security.html

Issue #14750, opened 2024-03-14
Add BearerTokenAuthenticationConverter


franticticktick:
Need to add `BearerTokenAuthenticationConverter` that implements `AuthenticationConverter`. Perhaps, it is worth extract some of the functionality from `ServerBearerTokenAuthenticationConverter` into `AbstractBearerTokenAuthenticationConverter` for example. This will be a template method pattern.

sjohnr:
Hi @CrazyParanoid! Thanks for reaching out. Can you please provide a bit more information about what this issue relates to? For example, are you attempting to use `AuthenticationFilter` configured to support bearer tokens? If so, is there a reason you are unable to use the existing `BearerTokenAuthenticationFilter`? Can you please describe your use case?

CrazyParanoid:
Hi @sjohnr ! Thanks for your feedback. Support for multiple authentication schemes in spring security is very important to me. It seems to me now that it is not entirely complete. I started several tasks to add support for delegating components. For example, #14644 a `DelegatingBearerTokenResolver` would solve my problem, but if the `AuthenticationFilter` is in the configurator, then it would be nice to have a set of delegating components. In that case, I'm interested in the appearance of delegating components (`DelegatingServerLogoutSuccessHandler`, `DelegatingServerAuthenticationConverter`, `RequestMatcherDelegatingAuthenticationSuccessHandler`), as this will help solve issues with several authentication schemes. 
Recently, some of these components have been implemented (#14655 #14654 ), but this is still not enough.

CrazyParanoid:
@sjohnr Thanks for your comments!

> Are you trying to support multiple authentication schemes through a single AuthenticationFilter?

Yes, initially I tried to do this through `BearerTokenAuthenticationFilter`  and I was missing `DelegatingBearerTokenResolver`.  `AuthenticationWebFilter` has all the necessary components, but I needed to implement #14521 I know many cases when legacy jwt-based authentication schemes work for many years and sooner or later there is a need to migrate to Oauth 2.0. 

> Adding some of these components may be very helpful for various scenarios so I don't see a problem with it in general. However, I can't help but feel like you may be choosing a complicated path for solving the problem you are having where a simpler solution may exist.

[...]

PR #14791, opened 2024-03-22
Add BearerTokenAuthenticationConverter

franticticktick:
Closes gh-14750

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
@@ -23,8 +23,6 @@ import java.util.Map;
 import java.util.function.Supplier;
 
 import jakarta.servlet.http.HttpServletRequest;
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
 
 import org.springframework.context.ApplicationContext;
 import org.springframework.core.convert.converter.Converter;
@@ -44,7 +42,6 @@ import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
 import org.springframework.security.oauth2.jwt.Jwt;
 import org.springframework.security.oauth2.jwt.JwtDecoder;
 import org.springframework.security.oauth2.jwt.NimbusJwtDecoder;
-import org.springframework.security.oauth2.server.resource.authentication.BearerTokenAuthenticationToken;
 import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationConverter;
 import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationProvider;
 import org.springframework.security.oauth2.server.resource.authentication.OpaqueTokenAuthenticationProvider;
@@ -69,7 +66,6 @@ import org.springframework.security.web.util.matcher.OrRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestHeaderRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher;
 import org.springframework.util.Assert;
-import org.springframework.util.StringUtils;
 import org.springframework.web.accept.ContentNegotiationStrategy;
 import org.springframework.web.accept.HeaderContentNegotiationStrategy;
 
@@ -200,13 +196,9 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 		return this;
 	}
 
-	/**
-	 * @deprecated please use {@link #authenticationConverter} instead
-	 */
-	@Deprecated
 	public OAuth2ResourceServerConfigurer<H> bearerTokenResolver(BearerTokenResolver bearerTokenResolver) {
 		Assert.notNull(bearerTokenResolver, "bearerTokenResolver cannot be null");
-		this.authenticationConverter = new BearerTokenResolverAuthenticationConverterAdapter(bearerTokenResolver);
+		this.authenticationConverter = new BearerTokenResolverHoldingAuthenticationConverter(bearerTokenResolver);
 		return this;
 	}
 
@@ -214,7 +206,7 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 	 * Sets the {@link AuthenticationConverter} to use.
 	 * @param authenticationConverter the authentication converter
 	 * @return the {@link OAuth2ResourceServerConfigurer} for further configuration
-	 * @since 6.5
+	 * @since 7.0
 	 */
 	public OAuth2ResourceServerConfigurer<H> authenticationConverter(AuthenticationConverter authenticationConverter) {
 		Assert.notNull(authenticationConverter, "authenticationConverter cannot be null");
@@ -299,10 +291,9 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 			resolver = (request) -> authenticationManager;
 		}
 
-		BearerTokenAuthenticationFilter filter = new BearerTokenAuthenticationFilter(resolver);
 		AuthenticationConverter converter = getAuthenticationConverter();
 		this.requestMatcher.setAuthenticationConverter(converter);
-		filter.setAuthenticationConverter(converter);
+		BearerTokenAuthenticationFilter filter = new BearerTokenAuthenticationFilter(resolver, converter);
 		filter.setAuthenticationEntryPoint(this.authenticationEntryPoint);
 		filter.setSecurityContextHolderStrategy(getSecurityContextHolderStrategy());
 		filter = postProcess(filter);
@@ -394,7 +385,7 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 		}
 		else if (this.context.getBeanNamesForType(BearerTokenResolver.class).length > 0) {
 			BearerTokenResolver bearerTokenResolver = this.context.getBean(BearerTokenResolver.class);
-			this.authenticationConverter = new BearerTokenResolverAuthenticationConverterAdapter(bearerTokenResolver);
+			this.authenticationConverter = new BearerTokenResolverHoldingAuthenticationConverter(bearerTokenResolver);
 		}
 		else {
 			this.authenticationConverter = new BearerTokenAuthenticationConverter();
@@ -404,7 +395,7 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 
 	BearerTokenResolver getBearerTokenResolver() {
 		AuthenticationConverter authenticationConverter = getAuthenticationConverter();
-		if (authenticationConverter instanceof BearerTokenResolverAuthenticationConverterAdapter bearer) {
+		if (authenticationConverter instanceof OAuth2ResourceServerConfigurer.BearerTokenResolverHoldingAuthenticationConverter bearer) {
 			return bearer.bearerTokenResolver;
 		}
 		return null;
@@ -614,24 +605,22 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 
 	}
 
-	private static final class BearerTokenResolverAuthenticationConverterAdapter implements AuthenticationConverter {
-
-		private final Log logger = LogFactory.getLog(BearerTokenResolverAuthenticationConverterAdapter.class);
+	private static final class BearerTokenResolverHoldingAuthenticationConverter implements AuthenticationConverter {
 
 		private final BearerTokenResolver bearerTokenResolver;
 
-		BearerTokenResolverAuthenticationConverterAdapter(BearerTokenResolver bearerTokenResolver) {
+		private final AuthenticationConverter authenticationConverter;
+
+		BearerTokenResolverHoldingAuthenticationConverter(BearerTokenResolver bearerTokenResolver) {
 			this.bearerTokenResolver = bearerTokenResolver;
+			BearerTokenAuthenticationConverter authenticationConverter = new BearerTokenAuthenticationConverter();
+			authenticationConverter.setBearerTokenResolver(bearerTokenResolver);
+			this.authenticationConverter = authenticationConverter;
 		}
 
 		@Override
 		public Authentication convert(HttpServletRequest request) {
-			String token = this.bearerTokenResolver.resolve(request);
-			if (!StringUtils.hasText(token)) {
-				this.logger.trace("Did not process request since did not find bearer token");
-				return null;
-			}
-			return new BearerTokenAuthenticationToken(token);
+			return this.authenticationConverter.convert(request);
 		}
 
 	}
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
@@ -23,6 +23,8 @@ import java.util.Map;
 import java.util.function.Supplier;
 
 import jakarta.servlet.http.HttpServletRequest;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 
 import org.springframework.context.ApplicationContext;
 import org.springframework.core.convert.converter.Converter;
@@ -37,10 +39,12 @@ import org.springframework.security.config.annotation.web.configurers.AbstractHt
 import org.springframework.security.config.annotation.web.configurers.CsrfConfigurer;
 import org.springframework.security.config.annotation.web.configurers.ExceptionHandlingConfigurer;
 import org.springframework.security.config.http.SessionCreationPolicy;
+import org.springframework.security.core.Authentication;
 import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
 import org.springframework.security.oauth2.jwt.Jwt;
 import org.springframework.security.oauth2.jwt.JwtDecoder;
 import org.springframework.security.oauth2.jwt.NimbusJwtDecoder;
+import org.springframework.security.oauth2.server.resource.authentication.BearerTokenAuthenticationToken;
 import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationConverter;
 import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationProvider;
 import org.springframework.security.oauth2.server.resource.authentication.OpaqueTokenAuthenticationProvider;
@@ -49,13 +53,14 @@ import org.springframework.security.oauth2.server.resource.introspection.OpaqueT
 import org.springframework.security.oauth2.server.resource.introspection.SpringOpaqueTokenIntrospector;
 import org.springframework.security.oauth2.server.resource.web.BearerTokenAuthenticationEntryPoint;
 import org.springframework.security.oauth2.server.resource.web.BearerTokenResolver;
-import org.springframework.security.oauth2.server.resource.web.DefaultBearerTokenResolver;
 import org.springframework.security.oauth2.server.resource.web.access.BearerTokenAccessDeniedHandler;
+import org.springframework.security.oauth2.server.resource.web.authentication.BearerTokenAuthenticationConverter;
 import org.springframework.security.oauth2.server.resource.web.authentication.BearerTokenAuthenticationFilter;
 import org.springframework.security.web.AuthenticationEntryPoint;
 import org.springframework.security.web.access.AccessDeniedHandler;
 import org.springframework.security.web.access.AccessDeniedHandlerImpl;
 import org.springframework.security.web.access.DelegatingAccessDeniedHandler;
+import org.springframework.security.web.authentication.AuthenticationConverter;
 import org.springframework.security.web.csrf.CsrfException;
 import org.springframework.security.web.util.matcher.AndRequestMatcher;
 import org.springframework.security.web.util.matcher.MediaTypeRequestMatcher;
@@ -64,6 +69,7 @@ import org.springframework.security.web.util.matcher.OrRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestHeaderRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 import org.springframework.web.accept.ContentNegotiationStrategy;
 import org.springframework.web.accept.HeaderContentNegotiationStrategy;
 
@@ -156,7 +162,7 @@ public final class OAuth2ResourceServerConfigurer<H extends HttpSecurityBuilder<
 
 	private AuthenticationManagerResolver<HttpServletRequest> authenticationManagerResolver;
 
-	private BearerTokenResolver bearerTokenResolver;
+	private AuthenticationConverter authenticationConverter;
 
 	private JwtConfigurer jwtConfigurer;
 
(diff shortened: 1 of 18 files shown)


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/AbstractBearerTokenAuthenticationConverter.java
```diff
@@ -0,0 +1,139 @@
...
+import org.springframework.util.StringUtils;
+
+/**
+ * Base class for bearer token converter implementations.
+ *
+ * @author Max Batischev
+ * @author Rob Winch
+ * @since 6.3
+ */
+public abstract class AbstractBearerTokenAuthenticationConverter<R> {
```
jzheaux, line 39: I'd prefer to keep these object hierarchies separate. It reduces the possibility of package tangles. Will you please remove this abstract class?

jzheaux (changes requested):
Will you please change `BearerTokenAuthenticationFilter` to accept an `AuthenticationConverter`? Also, `OAuth2ResourceServerConfigurer` should configure this in the same way as `BearerTokenResolver`. Will you please add that?

Finally, please add tests both for the new authentication converter and corresponding tests in `OAuth2ResourceServerConfigurerTests`.

CrazyParanoid:
Hi @jzheaux! Thanks for your feedback! I would like some advice on the best way to do this. I can add `AuthenticationConverter` to `BearerTokenAuthenticationFilter` and change `doFilterInternal` method:
```
        @Override
		try {
			Authentication authentication = authenticationConverter.convert(request);
			if (authentication == null) {
				this.logger.trace("Did not process request since did not find bearer token");
				filterChain.doFilter(request, response);
				return;
			}
			AuthenticationManager authenticationManager = this.authenticationManagerResolver.resolve(request);
			Authentication authenticationResult = authenticationManager.authenticate(authentication);
			SecurityContext context = this.securityContextHolderStrategy.createEmptyContext();
			context.setAuthentication(authenticationResult);
			this.securityContextHolderStrategy.setContext(context);
			this.securityContextRepository.saveContext(context, request, response);
				this.logger.debug(LogMessage.format("Set SecurityContextHolder to %s", authenticationResult));
			}
			filterChain.doFilter(request, response);
		} catch (AuthenticationException failed) {
			this.securityContextHolderStrategy.clearContext();
			this.logger.trace("Failed to process authentication request", failed);
			this.authenticationFailureHandler.onAuthenticationFailure(request, response, failed);
		}
	}
```
But there are several problems.
Firstly, it is necessary to save backward compatibility with `BearerTokenResolver`, which were set through `OAuth2ResourceServerConfigurer`, for example:
```
SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
			// @formatter:off
			http
					.authorizeRequests()
					.anyRequest().authenticated()
					.and()
					.oauth2ResourceServer()
					.bearerTokenResolver(customResolver())
					.jwt();
			return http.build();
```

Even if i declare method `bearerTokenResolver` as deprecated it should still work. And it seems hard to do, i could try adding a `BearerTokenResolverAuthenticationConverterAdapter` which would do something like this:
```
	private static final class BearerTokenResolverAuthenticationConverterAdapter implements AuthenticationConverter {

		private final BearerTokenResolver bearerTokenResolver;

		BearerTokenResolverAuthenticationConverterAdapter(BearerTokenResolver bearerTokenResolver) {
			Assert.notNull(bearerTokenResolver, "bearerTokenResolver cant be null");
			this.bearerTokenResolver = bearerTokenResolver;
		}

		@Override
		public Authentication convert(HttpServletRequest request) {
			String token = this.bearerTokenResolver.resolve(request);
			if (StringUtils.hasText(token)) {
				return new BearerTokenAuthenticationToken(token);
			}
			return null;
		}

	}
```
And use it only when a custom `BearerTokenResolver` is specified, for example:
```
BearerTokenResolver getBearerTokenResolver() {
		if (this.bearerTokenResolver == null) {
			if (this.context.getBeanNamesForType(BearerTokenResolver.class).length > 0) {
				this.bearerTokenResolver = this.context.getBean(BearerTokenResolver.class);
			} else {
				this.bearerTokenResolver = new DefaultBearerTokenResolver();
				return this.bearerTokenResolver;
			}
		}
		this.authenticationConverter = new BearerTokenResolverAuthenticationConverterAdapter(this.bearerTokenResolver);
		return this.bearerTokenResolver;
	}
```
Not the most beautiful method, but it partially solves the problem. Next you will need to do something with `BearerTokenRequestMatcher`, of course it can be changed to:
```
@Override
			try {
				Authentication authentication = authenticationConverter.convert(request);
				return authentication != null;
			} catch (OAuth2AuthenticationException ex) {
				return false;
			}
		}
```

But I'm not sure about this solution. In addition, there may be a problem with `AuthenticationDetailsSource`, in `BearerTokenAuthenticationFilter` it can be changed, but now it must be changed only through `AuthenticationConverter`.

CrazyParanoid:
Hi @jzheaux! I have added `authenticationConverter` to `OAuth2ResourceServerConfigurer` and to `BearerTokenAuthenticationFilter`. To ensure backward compatibility with `bearerTokenResolver`, several changes had to be made. First, i changed `BearerTokenRequestMatcher`:
```
private static final class BearerTokenRequestMatcher implements RequestMatcher {

		private BearerTokenResolver bearerTokenResolver;

		private AuthenticationConverter authenticationConverter;

		@Override
			try {
				if (this.bearerTokenResolver != null) {
					return this.bearerTokenResolver.resolve(request) != null;
				}
				return this.authenticationConverter.convert(request) != null;
			}
			catch (OAuth2AuthenticationException ex) {
				return false;
			}
		}

		void setBearerTokenResolver(BearerTokenResolver tokenResolver) {
			Assert.notNull(tokenResolver, "resolver cannot be null");
			this.bearerTokenResolver = tokenResolver;
		}

		void setAuthenticationConverter(AuthenticationConverter authenticationConverter) {
			Assert.notNull(authenticationConverter, "authenticationConverter cannot be null");
			this.authenticationConverter = authenticationConverter;
		}

	}
```
Now it works with authenticationConverter or with bearerTokenResolver, but not with both. If a `bearerTokenResolver` was specified, then the `BearerTokenAuthenticationFilter` will be used along with the `authenticationDetailsSource`. By default, `authenticationConverter` will be used if `bearerTokenResolver` is not specified:
```
        @Override
		AuthenticationManagerResolver resolver = this.authenticationManagerResolver;
		if (resolver == null) {
			AuthenticationManager authenticationManager = getAuthenticationManager(http);
			resolver = (request) -> authenticationManager;
		}
		BearerTokenAuthenticationFilter filter = new BearerTokenAuthenticationFilter(resolver);

		BearerTokenResolver bearerTokenResolver = getBearerTokenResolver();
		if (bearerTokenResolver != null) {
			this.requestMatcher.setBearerTokenResolver(bearerTokenResolver);
			filter.setBearerTokenResolver(bearerTokenResolver);
		}
		else {
			AuthenticationConverter converter = getAuthenticationConverter();
			this.requestMatcher.setAuthenticationConverter(converter);
			filter.setAuthenticationConverter(converter);
		}

		filter.setAuthenticationConverter(getAuthenticationConverter());
		filter.setAuthenticationEntryPoint(this.authenticationEntryPoint);
		filter.setSecurityContextHolderStrategy(getSecurityContextHolderStrategy());
		filter = postProcess(filter);
		http.addFilter(filter);
	}
```
In `BearerTokenAuthenticationFilter` `authenticationConverter` will work if `bearerTokenResolver` is not set:
```
                 Authentication authentication;
		try {
			if (this.bearerTokenResolver != null) {
				String token = this.bearerTokenResolver.resolve(request);
				if (!StringUtils.hasText(token)) {
					this.logger.trace("Did not process request since did not find bearer token");
					return;
				}
				authentication = bearerTokenAuthenticationToken(token, request);
			}
			else {
				authentication = this.authenticationConverter.convert(request);
			}
		}
		catch (OAuth2AuthenticationException invalid) {
			this.logger.trace("Sending to authentication entry point since failed to resolve bearer token", invalid);
			this.authenticationEntryPoint.commence(request, response, invalid);
			return;
		}

		if (authentication == null) {
			this.logger.trace("Failed to convert authentication request");
			filterChain.doFilter(request, response);
			return;
		}
```
Tests work as is. `AuthenticationDetailsSource` can only be set if `bearerTokenResolver` has been set. I think `bearerTokenResolver` can be noted as deprecated.

CrazyParanoid:
Hi @jzheaux ! Maybe it’s worth adding `BearerTokenAuthenticationConverter` to the `BearerTokenAuthenticationFilter` as a separate issue? This issue does not seem too easy and may affect the current functionality.

jzheaux:
@CrazyParanoid, thanks for your patience as I have been out of town.

We certainly want to remain passive and don't want to change behavior unnecessarily. Also, though, we don't typically add filter components that cannot be used by existing filters.

I'll take a look at the PR this week and see if I can find a way to simplify things. Thank you for all your research details.

jzheaux:
@franticticktick I realize it's been a while since you opened this PR. I've reviewed it and posted [a PR](https://github.com/franticticktick/spring-security/pull/2) to your repo that you can merge if you agree to the changes. We can see from there what other changes might be needed.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -194,9 +201,16 @@ public OAuth2ResourceServerConfigurer<H> authenticationManagerResolver(
...
 
+	@Deprecated
 	public OAuth2ResourceServerConfigurer<H> bearerTokenResolver(BearerTokenResolver bearerTokenResolver) {
 		Assert.notNull(bearerTokenResolver, "bearerTokenResolver cannot be null");
-		this.bearerTokenResolver = bearerTokenResolver;
+		this.authenticationConverter = new BearerTokenResolverAuthenticationConverterAdapter(bearerTokenResolver);
+		return this;
+	}
+
+	public OAuth2ResourceServerConfigurer<H> authenticationConverter(AuthenticationConverter authenticationConverter) {
```
jzheaux, line 211: Even though, `bearerTokenResolver` doesn't have a JavaDoc, please add one for `authenticationConverter`

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurer.java
```diff
@@ -194,9 +201,16 @@ public OAuth2ResourceServerConfigurer<H> authenticationManagerResolver(
 		return this;
 	}
 
+	@Deprecated
```
jzheaux, line 204: Please add a JavaDoc here that at least has the following:

```java
/**
  * @deprecated please use {@link #authenticationConverter} instead
  */
```

jzheaux (changes requested):
Thanks for the updates, @franticticktick. This is coming together nicely. I've left some feedback inline.

franticticktick:
Hi @jzheaux , i need some more time to tidy up this PR. Also, I have a few questions, I'll write a new comment a bit later.

config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ResourceServerConfigurerTests.java
```diff
@@ -760,10 +762,44 @@ public void getBearerTokenResolverWhenResolverBeanAndAnotherOnTheDslThenTheDslOn
 	}
 
 	@Test
-	public void getBearerTokenResolverWhenNoResolverSpecifiedThenTheDefaultIsUsed() {
```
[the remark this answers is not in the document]
franticticktick: This won't work. We'll never get `DefaultBearerTokenResolver`, since `BearerTokenAuthenticationConverter` now does the main work. At the same time, we have `BearerTokenResolverAuthenticationConverterAdapter`, which replaces any custom `BearerTokenResolver`, of course, it can have `DefaultBearerTokenResolver`, but only if the developer defines it himself. In other words, instead of `DefaultBearerTokenResolver`, we now have `BearerTokenAuthenticationConverter`.

config/src/main/java/org/springframework/security/config/http/OAuth2ResourceServerBeanDefinitionParser.java
```diff
@@ -124,25 +129,40 @@ public BeanDefinition parse(Element oauth2ResourceServer, ParserContext pc) {
...
+		BeanDefinition requestMatcher = buildRequestMatcher(bearerTokenResolver, authenticationConverter);
 		this.entryPoints.put(requestMatcher, authenticationEntryPoint);
 		this.deniedHandlers.put(requestMatcher, this.accessDeniedHandler);
 		this.ignoreCsrfRequestMatchers.add(requestMatcher);
 		BeanDefinitionBuilder filterBuilder = BeanDefinitionBuilder
 			.rootBeanDefinition(BearerTokenAuthenticationFilter.class);
 		BeanMetadataElement authenticationManagerResolver = getAuthenticationManagerResolver(oauth2ResourceServer);
 		filterBuilder.addConstructorArgValue(authenticationManagerResolver);
-		filterBuilder.addPropertyValue(BEARER_TOKEN_RESOLVER, bearerTokenResolver);
+		filterBuilder.addPropertyValue(AUTHENTICATION_CONVERTER, authenticationConverter);
```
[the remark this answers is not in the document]
franticticktick: I added such a check. Also, I removed `RootBeanDefinition` from `getAuthenticationConverter` and `getBearerTokenResolver` methods. We need to handle the case when `bearerTokenResolver` and `authenticationConverter` are null. This will mean that the user left the default configuration and we can set `authenticationConverter` to `RootBeanDefinition`.

```
		BeanMetadataElement bearerTokenResolver = getBearerTokenResolver(oauth2ResourceServer);
		BeanMetadataElement authenticationConverter = getAuthenticationConverter(oauth2ResourceServer);
		if (bearerTokenResolver != null && authenticationConverter != null) {
			throw new BeanDefinitionStoreException(
					"You cannot use bearer-token-ref and authentication-converter-ref in the same oauth2-resource-server element");
		}
		if (bearerTokenResolver == null && authenticationConverter == null) {
			authenticationConverter = new RootBeanDefinition(BearerTokenAuthenticationConverter.class);
		}
```

After these checks we are guaranteed to have either `bearerTokenResolver` or `authenticationConverter` left. And we can create a `requestMatcher` and add PropertyValue to the `filterBuilder`.

franticticktick:
Hey @jzheaux , thanks for the warmth. Could you review my latest changes please?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/authentication/BearerTokenAuthenticationFilter.java
```diff
@@ -181,7 +186,14 @@ public void setSecurityContextRepository(SecurityContextRepository securityConte
 	 */
 	public void setBearerTokenResolver(BearerTokenResolver bearerTokenResolver) {
```
jzheaux, line 187: Will you please mark this as deprecated?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/authentication/BearerTokenAuthenticationFilter.java
```diff
@@ -114,23 +117,25 @@ public BearerTokenAuthenticationFilter(AuthenticationManager authenticationManag
...
+
+		if (authenticationRequest == null) {
 			this.logger.trace("Did not process request since did not find bearer token");
 			filterChain.doFilter(request, response);
 			return;
 		}
 
-		BearerTokenAuthenticationToken authenticationRequest = new BearerTokenAuthenticationToken(token);
-		authenticationRequest.setDetails(this.authenticationDetailsSource.buildDetails(request));
+		if (authenticationRequest instanceof AbstractAuthenticationToken details) {
```
jzheaux, line 136: It's a bit tricky to be having the authentication converter setting the details as well as the filter. Instead of doing it in both places, will you please add to `BearerTokenAuthenticatinFilter` a private static adapter class and hold a default bearerTokenResolver and authenticationDetailsSource as members? This adapter would be the default `authenticationConverter` for `BearerTokenAuthenticationFilter`.

If `BearerTokenAuthenticationFilter#setBearerTokenResolver` or `#setAuthenticationDetailsSource` is called, it will set the adapter's members accordingly if `authenticationConverter` is of the right type. Otherwise, the setters return an error like "setBearerTokenResolver and setAuthenticationConverter cannot be used together". 

I think this makes sense because `BearerTokenAuthenticationConverter` takes care of resolving the token and populating it with details. So it's a usage error to try and do either of those here once you've specified an `AuthenticationConverter`.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/authentication/BearerTokenAuthenticationFilter.java
```diff
@@ -208,7 +220,7 @@ public void setAuthenticationFailureHandler(final AuthenticationFailureHandler a
 	/**
 	 * Set the {@link AuthenticationDetailsSource} to use. Defaults to
 	 * {@link WebAuthenticationDetailsSource}.
-	 * @param authenticationDetailsSource the {@code AuthenticationConverter} to use
+	 * @param authenticationDetailsSource the {@code AuthenticationDetailsSource} to use
 	 * @since 5.5
 	 */
 	public void setAuthenticationDetailsSource(
```
jzheaux, line 226: Will you please deprecate this method since `BearerTokenAuthenticationConverter` takes care of this?

jzheaux (changes requested):
Thanks for the next revision here and for your patience. I've been letting this one metabolize to ensure the solution is getting us closer to being able to use `AuthenticationFilter` while also honoring a separate initiative to make `Authentication` instances immutable.

So, :), in addition to my latest inline feedback, will you please update the `@since` attributes to 7.0 and use the 7.0 rnc?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/authentication/BearerTokenAuthenticationFilter.java
franticticktick: I think we can't fully guarantee backward compatibility this way. I can make an adapter like this:
```
	public static final class BearerTokenAuthenticationConverterAdapter implements AuthenticationConverter {
		private final Log logger = LogFactory.getLog(this.getClass());

		private AuthenticationDetailsSource<HttpServletRequest, ?> authenticationDetailsSource = new WebAuthenticationDetailsSource();
		private BearerTokenResolver bearerTokenResolver = new DefaultBearerTokenResolver();

		@Override
		public Authentication convert(HttpServletRequest request) {
			String token = this.bearerTokenResolver.resolve(request);
			if (!StringUtils.hasText(token)) {
				this.logger.trace("Did not process request since did not find bearer token");
				return null;
			}
			AbstractAuthenticationToken authenticationRequest = new BearerTokenAuthenticationToken(token);
			authenticationRequest.setDetails(this.authenticationDetailsSource.buildDetails(request));

            return authenticationRequest;
		}

		public void setAuthenticationDetailsSource(AuthenticationDetailsSource<HttpServletRequest, ?> authenticationDetailsSource) {
			Assert.notNull(authenticationDetailsSource, "authenticationDetailsSource cannot be null");
			this.authenticationDetailsSource = authenticationDetailsSource;
		}

			Assert.notNull(bearerTokenResolver, "bearerTokenResolver cannot be null");
			this.bearerTokenResolver = bearerTokenResolver;
		}

		public BearerTokenResolver getBearerTokenResolver() {
			return this.bearerTokenResolver;
		}
	}
```
It has both `authenticationDetailsSource` and `bearerTokenResolver`. At the same time, when calling the `setAuthenticationDetailsSource` method, I can check the `authenticationConverter` type:
```
	@Deprecated
	public void setAuthenticationDetailsSource(
			AuthenticationDetailsSource<HttpServletRequest, ?> authenticationDetailsSource) {
		if(this.authenticationConverter instanceof BearerTokenAuthenticationConverterAdapter adapter) {
			Assert.notNull(authenticationDetailsSource, "authenticationDetailsSource cannot be null");
			adapter.setAuthenticationDetailsSource(authenticationDetailsSource);
		} else {
			throw new IllegalStateException("authenticationDetailsSource and authenticationConverter cannot be use together");
		}
	}
```
But there is one problem, if the user wanted to call the `setAuthenticationDetailsSource` method without explicitly defining `BearerTokenResolver`, he will get an error. Pay attention to the test `OAuth2ResourceServerConfigurerTests.requestWhenCustomAuthenticationDetailsSourceThenUsed`.
```
	@Test
	public void requestWhenCustomAuthenticationDetailsSourceThenUsed() throws Exception {
		this.spring.register(CustomAuthenticationDetailsSource.class, JwtDecoderConfig.class, BasicController.class)
			.autowire();
		JwtDecoder decoder = this.spring.getContext().getBean(JwtDecoder.class);
		given(decoder.decode(anyString())).willReturn(JWT);
		this.mvc.perform(get("/authenticated").with(bearerToken(JWT_TOKEN)))
			.andExpect(status().isOk())
			.andExpect(content().string(JWT_SUBJECT));
		verifyBean(AuthenticationDetailsSource.class).buildDetails(any());
	}
```
This test will not pass. `authenticationDetailsSource` via `withObjectPostProcessor`, but the `bearerTokenResolver` was not explicitly set, which means that `BearerTokenAuthenticationFilter` will use `BearerTokenAuthenticationConverter` as the converter. In `OAuth2ResourceServerConfigurer`:
```
	AuthenticationConverter getAuthenticationConverter() {
		if (this.authenticationConverter != null) {
			return this.authenticationConverter;
		}
		if (this.context.getBeanNamesForType(AuthenticationConverter.class).length > 0) {
			this.authenticationConverter = this.context.getBean(AuthenticationConverter.class);
		}
		else if (this.context.getBeanNamesForType(BearerTokenResolver.class).length > 0) {
			BearerTokenResolver bearerTokenResolver = this.context.getBean(BearerTokenResolver.class);
			BearerTokenAuthenticationFilter.BearerTokenAuthenticationConverterAdapter converter = new BearerTokenAuthenticationFilter.BearerTokenAuthenticationConverterAdapter();
			converter.setBearerTokenResolver(bearerTokenResolver);
			this.authenticationConverter = converter;
		}
		else {
			this.authenticationConverter = new BearerTokenAuthenticationConverter();
		}
		return this.authenticationConverter;
	}
```
Simply put, `authenticationDetailsSource` can only be set if `bearerTokenResolver` is explicitly specified via bean or via configurer. Otherwise, `BearerTokenAuthenticationFilter` will use `BearerTokenAuthenticationConverter`. Without explicitly specified `bearerTokenResolver`, setting `authenticationDetailsSource` is simply pointless. But we can do this:
```
	@Deprecated
		logger.warn("If you specify an authenticationConverter it will not be used.");
		BearerTokenAuthenticationConverterAdapter adapter = new BearerTokenAuthenticationConverterAdapter();
		adapter.setBearerTokenResolver(bearerTokenResolver);
		this.authenticationConverter = adapter;
	}

	@Deprecated
	public void setAuthenticationDetailsSource(
			AuthenticationDetailsSource<HttpServletRequest, ?> authenticationDetailsSource) {
		logger.warn("If you specify an authenticationConverter it will not be used. " +
				"Instead the DefaultBearerTokenResolver will be used in combination with the authenticationDetailsSource you specify.");
		BearerTokenAuthenticationConverterAdapter adapter = new BearerTokenAuthenticationConverterAdapter();
		adapter.setAuthenticationDetailsSource(authenticationDetailsSource);
		this.authenticationConverter = adapter;
	}
```
At the same time, we will preserve backward compatibility as much as possible. All tests will remain as is, we will warn the user that authenticationConverter will not be used. @jzheaux what do you think about this?

franticticktick: The correct solution would be like this:
```
	@Deprecated
	public void setAuthenticationDetailsSource(
			AuthenticationDetailsSource<HttpServletRequest, ?> authenticationDetailsSource) {
		if(this.authenticationConverter instanceof BearerTokenAuthenticationConverterAdapter adapter) {
			adapter.setAuthenticationDetailsSource(authenticationDetailsSource);
		} else {
			logger.warn("If you specify an authenticationConverter it will not be used. " +
					"Instead the DefaultBearerTokenResolver will be used in combination with the authenticationDetailsSource you specify.");
			BearerTokenAuthenticationConverterAdapter adapter = new BearerTokenAuthenticationConverterAdapter();
			adapter.setAuthenticationDetailsSource(authenticationDetailsSource);
			this.authenticationConverter = adapter;
		}
	}
```
And:
```
	@Deprecated
		if(this.authenticationConverter instanceof BearerTokenAuthenticationConverterAdapter adapter) {
			adapter.setBearerTokenResolver(bearerTokenResolver);
		} else {
			logger.warn("If you specify an authenticationConverter it will not be used.");
			BearerTokenAuthenticationConverterAdapter adapter = new BearerTokenAuthenticationConverterAdapter();
			adapter.setBearerTokenResolver(bearerTokenResolver);
			this.authenticationConverter = adapter;	
		}
	}
```

jzheaux:
Thanks for all your help with this @franticticktick over a protracted analysis. This will merge into `main` once the build passes.

franticticktick:
Thanks @jzheaux , I was just about to continue working on this pr, but you beat me to it :)

franticticktick:
@jzheaux I have a question, I looked at https://github.com/spring-projects/spring-security/pull/14791/commits/ce0854b87c26cf6162d5e98696ec0e6b9d5cfed4 , are we sure we want to replace `BearerTokenAuthenticationConverter` logic with `BearerTokenResolver` adapter logic? I expected that we would make a separate adapter class.

Issue #14646, opened 2024-02-22
Update Documentation about Kotlin Coroutine Support


jzheaux:
Based one [this comment](https://github.com/spring-projects/spring-security/issues/9401#issuecomment-1833459899) from @rrrship, the documentation should be updated to correctly state that Spring Security fully supports coroutines at this point.

PR #14797, opened 2024-03-23
Updated Documentation about Kotlin Coroutine Support #14646

Chillthrower:
Updated Documentation about Kotlin Coroutine Support #14646

### DIFF
--- a/docs/modules/ROOT/pages/reactive/authorization/method.adoc
+++ b/docs/modules/ROOT/pages/reactive/authorization/method.adoc
@@ -19,7 +19,7 @@ This improves upon `@EnableReactiveMethodSecurity` in a number of ways. `@Enable
 
 1. Uses the simplified `AuthorizationManager` API instead of metadata sources, config attributes, decision managers, and voters.
 This simplifies reuse and customization.
-2. Supports reactive return types. Note that we are waiting on https://github.com/spring-projects/spring-framework/issues/22462[additional coroutine support from the Spring Framework] before adding coroutine support.
+2. Supports reactive return types including Kotlin coroutines.
 3. Is built using native Spring AOP, removing abstractions and allowing you to use Spring AOP building blocks to customize
 4. Checks for conflicting annotations to ensure an unambiguous security configuration
 5. Complies with JSR-250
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/reactive/authorization/method.adoc
```diff
@@ -304,11 +304,10 @@ and it will be invoked after the `@PostAuthorize` interceptor.
 
 == EnableReactiveMethodSecurity
 
-[WARNING]
+[INFO]
```
sjohnr, line 307: Instead of changing the warning to info with the given message, I think we can simply remove the warning. Would you please consider this change?

sjohnr (changes requested):
Hi @Chillthrower thanks for the contribution! See one comment below and let me know if you agree.

In addition, there is a reference to the pending Kotlin coroutines support earlier in this page. I think the this should change from:

> 2. Supports reactive return types. Note that we are waiting on [additional coroutine support from the Spring Framework](https://github.com/spring-projects/spring-framework/issues/22462) before adding coroutine support.

to:

> 2. Supports reactive return types including Kotlin coroutines.

Would you be able to make these changes?

docs/modules/ROOT/pages/reactive/authorization/method.adoc
Chillthrower: I made the changes as per your request if further changes have to be made please let me know

sjohnr: Sorry, I don't think I was clear enough. 😓  I meant remove the note you added as well. The entire paragraph is not needed.

sjohnr:
@Chillthrower did you see my request above for updating another section of this page? I don't see that update.

docs/modules/ROOT/pages/reactive/authorization/method.adoc
Chillthrower: oh my bad, I removed that para and also added "Supports reactive return types including Kotlin coroutines."

sjohnr:
Thanks @Chillthrower! This is now merged as b5c9ba6c1c6c43ae3d04cbcf8e4a7538ce8b16bb.

Issue #14803, opened 2024-03-26
Bad return type for HeadersConfigurer#permissionsPolicy method with customizer


florianberthe:
**Describe the bug**
`permissionsPolicy(Customizer<PermissionsPolicyConfig> permissionsPolicyCustomizer)` method in `HeadersConfigurer` class currently returns an instance of `PermissionsPolicyConfig`.  It breaks the chaining in HeadersConfigurer.

**To Reproduce**
See samples below.
Current implementation:
```
public PermissionsPolicyConfig permissionsPolicy(Customizer<PermissionsPolicyConfig> permissionsPolicyCustomizer) {
    this.permissionsPolicy.writer = new PermissionsPolicyHeaderWriter();
    permissionsPolicyCustomizer.customize(this.permissionsPolicy);
    return this.permissionsPolicy;
}
```

**Expected behavior**
This method should return `HeadersConfigurer<H>` instance.
```
public HeadersConfigurer<H> permissionsPolicy(Customizer<PermissionsPolicyConfig> permissionsPolicyCustomizer) {
    this.permissionsPolicy.writer = new PermissionsPolicyHeaderWriter();
    permissionsPolicyCustomizer.customize(this.permissionsPolicy);
    return HeadersConfigurer.this;
}
```

**Sample**

This code compiles:
```
public static void applyConfig(HttpSecurity http) throws Exception {
        http.headers(headers -> headers
                .defaultsDisabled()
                .xssProtection(xssProtection -> xssProtection.headerValue(HeaderValue.ENABLED_MODE_BLOCK))
                .frameOptions(FrameOptionsConfig::sameOrigin)
                .contentSecurityPolicy(contentSecurityPolicy -> contentSecurityPolicy
                    .policyDirectives("default-src 'none'" +
                        "; frame-ancestors 'self'" +
                        "; style-src 'self' 'unsafe-inline'" +
                        "; form-action 'self'" +
                        "; font-src 'self'" +
                        "; object-src 'none'" +
                        "; media-src 'self'")
                )
                .referrerPolicy(referrerPolicy -> referrerPolicy.policy(ReferrerPolicy.STRICT_ORIGIN))
                .permissionsPolicy(permissionsPolicy -> permissionsPolicy.policy("camera=(self), sync-xhr=(self), geolocation=()"))
            );
    }
```

This code doesn't compile:
```
public static void applyConfig(HttpSecurity http) throws Exception {
        http.headers(headers -> headers
                .defaultsDisabled()
[...]

PR #14804, opened 2024-03-26
Fix HeadersConfigurer#permissionsPolicy method with customizer

florianberthe:
Preserve HeadersConfigurer's chaining like other methods with customizer (New SpringSecurity DSL)

Closes gh-14803


florianberthe:
> Hi, @florianberthe. Thanks for the PR.
> 
> Can you please target the `5.8.x` branch since the problem is present there too?

Do I have to recreate a new PR for that ?

marcusdacoregio:
@florianberthe you don't need to. You can change the target branch to `5.8.x` and then rebase your branch with `5.8.x`.

Issue #14778, opened 2024-03-19
Support customizing the redirect URL in OidcClientInitiatedServerLogoutSuccessHandler


stipx:
**Expected Behavior**

In order to be able to work with some restrictive SSO implementations sometimes additional parameters are needed (like "state") in order that the logout request is succeeding.

**Current Behavior**

In order to achieve this it was necessary to implement a custom logout handler which gets the logout/end_session URL from the client registration and sets the ID-Token hint, the redirect uri and the state and does a redirect then.

So a simple resolver/converter/customizer for the redirect URL would be much easier than implementing a whole new logout handler.

programista-zacny:
I agree.

We needed to redirect to dynamic url when OIDC logout. `location` query param was used.
E.g. `/logout?location=http://somewhere.com`

```java
@RequiredArgsConstructor
public class RedirectOidcServerLogoutSuccessHandler implements ServerLogoutSuccessHandler {

    private final ReactiveClientRegistrationRepository clientRegistrationRepository;

    @Override
    public Mono<Void> onLogoutSuccess(WebFilterExchange exchange, Authentication authentication) {
        OidcClientInitiatedServerLogoutSuccessHandler delegate =
                new OidcClientInitiatedServerLogoutSuccessHandler(clientRegistrationRepository);
        getLocation(exchange).ifPresent(delegate::setPostLogoutRedirectUri);
        return delegate.onLogoutSuccess(exchange, authentication);
    }

    private Optional<String> getLocation(WebFilterExchange exchange) {
        return Optional.ofNullable(exchange.getExchange().getRequest().getQueryParams().getFirst("location"));
    }
}
```

CrazyParanoid:
I can suggest a simple solution - add `setServerLogoutSuccessHandler`. This will allow you to implement your `RedirectServerLogoutSuccessHandler` and pass it to `OidcClientInitiatedServerLogoutSuccessHandler`, for example:
```
    public class RedirectOidcServerLogoutSuccessHandler extends RedirectServerLogoutSuccessHandler {

        private final ReactiveClientRegistrationRepository clientRegistrationRepository;

        public RedirectOidcServerLogoutSuccessHandler(ReactiveClientRegistrationRepository clientRegistrationRepository) {
            this.clientRegistrationRepository = clientRegistrationRepository;
        }

        @Override
        public Mono<Void> onLogoutSuccess(WebFilterExchange exchange, Authentication authentication) {
[...]

PR #14808, opened 2024-03-27
Customize Redirect URI in OidcClientInitiatedServerLogoutSuccessHandler

franticticktick:
Closes gh-14778

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/web/server/logout/OidcClientInitiatedServerLogoutSuccessHandler.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/web/server/logout/OidcClientInitiatedServerLogoutSuccessHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -23,6 +23,7 @@ import java.util.Map;
 
 import reactor.core.publisher.Mono;
 
+import org.springframework.core.convert.converter.Converter;
 import org.springframework.http.server.reactive.ServerHttpRequest;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.oauth2.client.authentication.OAuth2AuthenticationToken;
@@ -35,6 +36,7 @@ import org.springframework.security.web.server.WebFilterExchange;
 import org.springframework.security.web.server.authentication.logout.RedirectServerLogoutSuccessHandler;
 import org.springframework.security.web.server.authentication.logout.ServerLogoutSuccessHandler;
 import org.springframework.util.Assert;
+import org.springframework.web.server.ServerWebExchange;
 import org.springframework.web.util.UriComponents;
 import org.springframework.web.util.UriComponentsBuilder;
 
@@ -56,6 +58,8 @@ public class OidcClientInitiatedServerLogoutSuccessHandler implements ServerLogo
 	private final ReactiveClientRegistrationRepository clientRegistrationRepository;
 
 	private String postLogoutRedirectUri;
+
+	private Converter<RedirectUriParameters, Mono<String>> redirectUriResolver = new DefaultRedirectUriResolver();
 
 	/**
 	 * Constructs an {@link OidcClientInitiatedServerLogoutSuccessHandler} with the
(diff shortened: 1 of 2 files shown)


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/web/server/logout/OidcClientInitiatedServerLogoutSuccessHandler.java
```diff
@@ -189,4 +182,90 @@ public void setLogoutSuccessUrl(URI logoutSuccessUrl) {
...
+
+	/**
+	 * Parameters, required for redirect URI resolving.
+	 *
+	 * @author Max Batischev
+	 * @since 6.4
+	 */
+	public static class RedirectUriParameters {
+
+		private ServerWebExchange serverWebExchange;
```
jzheaux, line 204: Please have these be `final` and in the constructor

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/web/server/logout/OidcClientInitiatedServerLogoutSuccessHandler.java
```diff
@@ -72,22 +76,11 @@ public OidcClientInitiatedServerLogoutSuccessHandler(
 
 	@Override
 	public Mono<Void> onLogoutSuccess(WebFilterExchange exchange, Authentication authentication) {
+		RedirectUriParameters redirectUriParameters = new RedirectUriParameters();
+		redirectUriParameters.setAuthentication(authentication);
+		redirectUriParameters.setServerWebExchange(exchange.getExchange());
 		// @formatter:off
-		return Mono.just(authentication)
```
jzheaux, line 76: Can we please look up the `ClientRegistration` before passing to the resolver?

jzheaux (changes requested):
Thanks, @CrazyParanoid! I've left my feedback inline.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/web/server/logout/OidcClientInitiatedServerLogoutSuccessHandler.java
CrazyParanoid: I can add Final to the class, but I'm not sure if it's worth making all the fields Final and initializing them in the constructor. The `DefaultRedirectUriResolver` requires two attributes `Authentication` and `ServerWebExchange` and this attributes are passed to the `onLogoutSuccess` method, meaning they cannot be null, so they can be made final and added to the constructor, `ClientRegistration` will be found in the resolver (`OidcClientInitiatedServerLogoutSuccessHandler. this.clientRegistrationRepository::findByRegistrationId`). That is, it cannot be declared final and required to be initialized in the constructor. I can easily pass it via setter if I need another one `ClientRegistration`:
```
this.handler.setRedirectUriResolver((params) -> {
			params.setClientRegistration(ClientRegistration.withRegistrationId("clientId")
					.build());
			return Mono.just("URL");});
```

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/web/server/logout/OidcClientInitiatedServerLogoutSuccessHandler.java
CrazyParanoid: Currently, `redirectUriResolver` finds `clientRegistration` in `clientRegistrationRepository`. To do this, it needs to get the `authorizedClientRegistrationId` from `authentication`. It will not be possible to set `clientRegistration` before passing it to `redirectUriResolver`, since by default it has not yet been found at this moment.

jzheaux: I don't think I'm following. Why can't the code lookup the registration id from the authentication before looking up the client registration? That's what it was doing before. What am I missing?

IOW, could you instead only have the resolver perform the steps from [here](https://github.com/spring-projects/spring-security/pull/14808/files#diff-9b5866925549763b9729318297077754e6ffa9c14766158fd3ea6fa466cef86dL83) to [here](https://github.com/spring-projects/spring-security/pull/14808/files#diff-9b5866925549763b9729318297077754e6ffa9c14766158fd3ea6fa466cef86dL89)?

CrazyParanoid: I understood your idea, maybe I initially went the wrong way. Could you please review the changes?

jzheaux (approved):

jzheaux:
@franticticktick, sorry for the delay on this one. Can you do one more thing for me and change the `@since` to 6.5? Then we'll merge once we cut the `6.4.x` branch

franticticktick:
Hi @jzheaux, done.

jzheaux:
Thanks, @franticticktick for this PR. This is now merged into `main`.

Issue #14599, opened 2024-02-13
Support Converter for SpEL return types to AuthorizationDecision


rwinch:
Support custom return types for for SpEL method security through the use of providing a Converter of that return type to the Authorization Decision.

I don't think that I like this idea but I'm documenting it. The reason I'm not sure if I like this idea as the SpEL expression can invoke the converter itself.

jzheaux:
Closed by automation due to typo in commit message. Reopening.

PR #14840, opened 2024-04-04
Support SpEL Returning AuthorizationDecision

jzheaux:
Closes gh-14599

WIth this PR, an application can return an `AuthorizationDecision` from a SpEL expression.

Note that the `MethodAuthorizationDeniedHandler` and `MethodAuthorizationResultPostProcessor` object hierarchies were modified to ensure that any custom type that is returned from a SpEL expression is preserved. No authorization decision wrapping is performed any longer in this case.

This is also valuable for being able to use custom handling and post-processing in the event that the expression throws an exception.


core/src/main/java/org/springframework/security/access/expression/ExpressionUtils.java
```diff
@@ -19,12 +19,42 @@
...
+			Object result = expr.getValue(ctx);
+			if (result instanceof AuthorizationDecision decision) {
+				return decision;
+			}
+			if (result instanceof Boolean granted) {
+				return new ExpressionAuthorizationDecision(granted, expr);
+			}
+			if (result == null) {
+				return null;
+			}
```
marcusdacoregio, line 48: Have you considered using `AuthorizationDecision decision = expr.getValue(ctx, AuthorizationDecision.class)` and let the `ConversionService` do the work?

For example, you can change the `DefaultMethodSecurityExpressionHandler#createEvaluationContext` to include a custom `TypeConverter`, something like:

```java
GenericConversionService conversionService = new GenericConversionService();
conversionService.addConverter(new BooleanToAuthorizationDecisionConverter());
StandardTypeConverter typeConverter = new StandardTypeConverter(conversionService);
ctx.setTypeConverter(typeConverter);
```

core/src/main/java/org/springframework/security/authorization/method/ReactiveExpressionUtils.java
```diff
@@ -56,7 +85,7 @@ static Mono<Boolean> evaluateAsBoolean(Expression expr, EvaluationContext ctx) {
 		});
 	}
 
-	private static Mono<Boolean> createInvalidReturnTypeMono(Expression expr) {
+	private static <T> Mono<T> createInvalidReturnTypeMono(Expression expr) {
 		return Mono.error(() -> new IllegalStateException(
 				"Expression: '" + expr.getExpressionString() + "' must return boolean or Mono<Boolean>"));
```
marcusdacoregio, line 90: I believe this message should be changed as well to include `AuthorizationDecision`

core/src/main/java/org/springframework/security/access/expression/ExpressionUtils.java
```diff
@@ -19,12 +19,42 @@
...
+	/**
+	 * Evaluate a SpEL expression and coerce into an {@link AuthorizationDecision}
+	 * @param expr a SpEL expression
+	 * @param ctx an {@link EvaluationContext}
+	 * @return the resulting {@link AuthorizationDecision}
+	 * @since 6.3
+	 */
+	public static AuthorizationDecision evaluate(Expression expr, EvaluationContext ctx) {
+		try {
+			Object result = expr.getValue(ctx);
```
marcusdacoregio, line 39: I believe we should support `AuthorizationResult` instead of `AuthorizationDecision`. We can even add a new method to `AuthorizationManager` named `authorize` to return an `AuthorizationResult`.

marcusdacoregio (changes requested):
Thanks @jzheaux. I've left some feedback inline.

core/src/main/java/org/springframework/security/access/expression/ExpressionUtils.java
jzheaux: I suppose I don't see the value as the logic is quite simple. Further, the conversion service wouldn't have access to the `Expression` so it couldn't construct an `ExpressionAuthorizationDecision`, which is needed for backward compatibility.

core/src/main/java/org/springframework/security/access/expression/ExpressionUtils.java
jzheaux: For now, I'll make sure that all new APIs use `AuthorizationResult`. We can cast inside of authorization manager instances for now. I think we should add support in `AuthorizationManager` for `AuthorizationResult` in a new ticket.

Issue #14843, opened 2024-04-04
AuthorizationManager should support returning an AuthorizationResult


jzheaux:
It would be nice for `AuthoriationManager` to code to the `AuthorizationResult` interface like so:

```java
default AuthorizationResult authorize(Supplier<Authentication> authentication, T object) {
    return check(authentication, object);
}
```

This will allow implementations to code to an interface instead of a concrete class.

In turn, `check` should be deprecated and classes in Spring Security should call `authorize` instead.

As when deprecating other methods, the existing tests should remain as-is to ensure that deprecated functionality continues to work.

PR #14846, opened 2024-04-04
AuthorizationManager should return AuthorizationResult

CrazyParanoid:
Added a new authorization method to AuthorizationManager that returns AuthorizationResult.

Closes gh-14843

### DIFF
--- a/core/src/main/java/org/springframework/security/authorization/AuthorizationEventPublisher.java
+++ b/core/src/main/java/org/springframework/security/authorization/AuthorizationEventPublisher.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -42,8 +42,37 @@ public interface AuthorizationEventPublisher {
 	 * @param object the secured object
 	 * @param decision the decision about whether the user may access the secured object
 	 * @param <T> the secured object's type
+	 * @deprecated use
+	 * {@link #publishAuthorizationEvent(Supplier, Object, AuthorizationResult)} instead
 	 */
+	@Deprecated
 	<T> void publishAuthorizationEvent(Supplier<Authentication> authentication, T object,
 			AuthorizationDecision decision);
+
+	/**
+	 * Publish the given details in the form of an event, typically
+	 * {@link AuthorizationGrantedEvent} or {@link AuthorizationDeniedEvent}.
+	 *
+	 * Note that success events can be very noisy if enabled by default. Because of this
+	 * implementations may choose to drop success events by default.
+	 * @param authentication a {@link Supplier} for the current user
+	 * @param object the secured object
+	 * @param result {@link AuthorizationResult} the result about whether the user may
+	 * access the secured object
+	 * @param <T> the secured object's type
+	 * @since 6.4
+	 */
+	default <T> void publishAuthorizationEvent(Supplier<Authentication> authentication, T object,
+			AuthorizationResult result) {
+		if (result == null) {
+			publishAuthorizationEvent(authentication, object, null);
+			return;
+		}
+		if (result instanceof AuthorizationDecision decision) {
+			publishAuthorizationEvent(authentication, object, decision);
+			return;
+		}
+		throw new UnsupportedOperationException("result must be of type AuthorizationDecision");
+	}
 
 }
--- a/core/src/main/java/org/springframework/security/authorization/AuthorizationManager.java
+++ b/core/src/main/java/org/springframework/security/authorization/AuthorizationManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -50,8 +50,23 @@ public interface AuthorizationManager<T> {
 	 * @param authentication the {@link Supplier} of the {@link Authentication} to check
 	 * @param object the {@link T} object to check
 	 * @return an {@link AuthorizationDecision} or null if no decision could be made
+	 * @deprecated please use {@link #authorize(Supplier, Object)} instead
 	 */
 	@Nullable
+	@Deprecated
 	AuthorizationDecision check(Supplier<Authentication> authentication, T object);
+
+	/**
+	 * Determines if access is granted for a specific authentication and object.
+	 * @param authentication the {@link Supplier} of the {@link Authentication} to
+	 * authorize
+	 * @param object the {@link T} object to authorize
+	 * @return an {@link AuthorizationResult}
+	 * @since 6.4
+	 */
+	@Nullable
+	default AuthorizationResult authorize(Supplier<Authentication> authentication, T object) {
+		return check(authentication, object);
+	}
 
 }
--- a/core/src/main/java/org/springframework/security/authorization/AuthorizationObservationContext.java
+++ b/core/src/main/java/org/springframework/security/authorization/AuthorizationObservationContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -34,6 +34,8 @@ public class AuthorizationObservationContext<T> extends Observation.Context {
 	private final T object;
 
 	private AuthorizationDecision decision;
+
+	private AuthorizationResult authorizationResult;
 
 	public AuthorizationObservationContext(T object) {
 		Assert.notNull(object, "object cannot be null");
@@ -71,17 +73,43 @@ public class AuthorizationObservationContext<T> extends Observation.Context {
 	/**
 	 * Get the observed {@link AuthorizationDecision}
 	 * @return the observed {@link AuthorizationDecision}
+	 * @deprecated please use {@link #getAuthorizationResult()} instead
 	 */
+	@Deprecated
 	public AuthorizationDecision getDecision() {
-		return this.decision;
+		Assert.isInstanceOf(AuthorizationDecision.class, this.authorizationResult,
+				"Please call getAuthorizationResult instead. If you must call getDecision, please ensure that the result you provide is of type AuthorizationDecision");
+		return (AuthorizationDecision) this.authorizationResult;
 	}
 
 	/**
 	 * Set the observed {@link AuthorizationDecision}
 	 * @param decision the observed {@link AuthorizationDecision}
+	 * @deprecated please use {@link #setAuthorizationResult(AuthorizationResult)} instead
 	 */
+	@Deprecated
 	public void setDecision(AuthorizationDecision decision) {
+		Assert.isInstanceOf(AuthorizationDecision.class, decision,
+				"Please call setAuthorizationResult instead. If you must call getDecision, please ensure that the result you provide is of type AuthorizationDecision");
 		this.decision = decision;
+	}
+
+	/**
+	 * Get the observed {@link AuthorizationResult}
+	 * @return the observed {@link AuthorizationResult}
+	 * @since 6.4
+	 */
+	public AuthorizationResult getAuthorizationResult() {
+		return this.authorizationResult;
+	}
+
+	/**
+	 * Set the observed {@link AuthorizationResult}
+	 * @param authorizationResult the observed {@link AuthorizationResult}
+	 * @since 6.4
+	 */
+	public void setAuthorizationResult(AuthorizationResult authorizationResult) {
+		this.authorizationResult = authorizationResult;
 	}
 
 }
--- a/core/src/main/java/org/springframework/security/authorization/AuthorizationObservationContext.java
+++ b/core/src/main/java/org/springframework/security/authorization/AuthorizationObservationContext.java
@@ -33,8 +33,6 @@ public class AuthorizationObservationContext<T> extends Observation.Context {
 
 	private final T object;
 
-	private AuthorizationDecision decision;
-
 	private AuthorizationResult authorizationResult;
 
 	public AuthorizationObservationContext(T object) {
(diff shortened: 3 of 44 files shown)


jzheaux (changes requested):
Thanks, @CrazyParanoid.

With `check` deprecated, we should no longer call it in from other non-deprecated production code in Spring Security. Can you also make those corresponding changes? Please leave tests as-is, though.

Also, please add tests to ensure that the new method works.

Finally, please take a look at `AuthorizaitonManagerBeforeMethodInterceptor` and other method interceptors to ensure that they are no longer casting expression values to `AuthorizationDecision`. Instead, they should implement the `authorize` method, have their `check` method call it, and then perform the cast there.

CrazyParanoid:
Hi @jzheaux ! I added the `authorize` method to the `AuthorizationManager` and made the default implementation of the `check` method:
```
default AuthorizationDecision check(Supplier<Authentication> authentication, T object) {
		return (AuthorizationDecision) authorize(authentication, object);
	}
```
With this implementation, it will not be possible to leave the tests as is, because in some tests, a mock of the `check` method is made and its invocation is verified:
```
verify(this.mockAuthorizationManager).check(any(), any());
```
And it needs to be changed to `authorize`, since the mock was made for example:
```
given(this.mockAuthorizationManager.check(any(), any())).willReturn(new AuthorizationDecision(true));
```
In addition, there are several delegating components that make call `check` method of delegate.  Here you need to mock the `authorize` and verify its invocation.

core/src/main/java/org/springframework/security/authorization/AuthorizationManager.java
```diff
@@ -50,8 +50,23 @@ default void verify(Supplier<Authentication> authentication, T object) {
...
+	/**
+	 * Determines if access should be granted for a specific authentication and object.
+	 * @param authentication the {@link Supplier} of the {@link Authentication} to
+	 * authorize
+	 * @param object the {@link T} object to authorize
+	 * @return an {@link AuthorizationResult} or null if no result could be made
+	 * @since 6.3
+	 */
+	@Nullable
+	AuthorizationResult authorize(Supplier<Authentication> authentication, T object);
```
jzheaux, line 70: Sorry, I may have misspoken here. We can't introduce a new non-default method to an existing interface as that would not be passive. Will you please switch this so that `authorize` has the default implementation?

core/src/main/java/org/springframework/security/authorization/AuthorizationObservationContext.java
```diff
@@ -69,18 +69,18 @@ public T getObject() {
 	}
 
 	/**
-	 * Get the observed {@link AuthorizationDecision}
-	 * @return the observed {@link AuthorizationDecision}
+	 * Get the observed {@link AuthorizationResult}
+	 * @return the observed {@link AuthorizationResult}
 	 */
-	public AuthorizationDecision getDecision() {
+	public AuthorizationResult getDecision() {
```
jzheaux, line 75: To remain [binary compatible](https://wiki.eclipse.org/Evolving_Java-based_APIs_2#Evolving_API_interfaces_-_API_methods), we cannot change return types. Instead, please add a new getter (called `getAuthorizationResult`) and deprecate this getter.

core/src/main/java/org/springframework/security/authorization/AuthorizationObservationContext.java
```diff
@@ -69,18 +69,18 @@ public T getObject() {
...
 	}
 
 	/**
-	 * Set the observed {@link AuthorizationDecision}
-	 * @param decision the observed {@link AuthorizationDecision}
+	 * Set the observed {@link AuthorizationResult}
+	 * @param decision the observed {@link AuthorizationResult}
 	 */
-	public void setDecision(AuthorizationDecision decision) {
+	public void setDecision(AuthorizationResult decision) {
```
jzheaux, line 83: To remain [binary compatible](https://wiki.eclipse.org/Evolving_Java-based_APIs_2#Evolving_API_interfaces_-_API_methods), we cannot change return types. Instead, please add a new setter (called `setAuthorizationResult`) and deprecate this setter.

core/src/main/java/org/springframework/security/authorization/ObservationAuthorizationManager.java
```diff
@@ -69,16 +69,16 @@ public ObservationAuthorizationManager(ObservationRegistry registry, Authorizati
 	}
 
 	@Override
-	public AuthorizationDecision check(Supplier<Authentication> authentication, T object) {
```
jzheaux, line 72: I believe once you change to `authorize` being the default method, you will be fine to leave implementations as they are.

core/src/main/java/org/springframework/security/authorization/method/AuthorizationManagerBeforeMethodInterceptor.java
```diff
@@ -245,8 +246,8 @@ public void setSecurityContextHolderStrategy(SecurityContextHolderStrategy secur
 
 	private Object attemptAuthorization(MethodInvocation mi) throws Throwable {
 		this.logger.debug(LogMessage.of(() -> "Authorizing method invocation " + mi));
-		AuthorizationDecision decision = this.authorizationManager.check(this::getAuthentication, mi);
-		this.eventPublisher.publishAuthorizationEvent(this::getAuthentication, mi, decision);
+		AuthorizationResult decision = this.authorizationManager.authorize(this::getAuthentication, mi);
+		this.eventPublisher.publishAuthorizationEvent(this::getAuthentication, mi, (AuthorizationDecision) decision);
```
jzheaux, line 250: Let's consider also adding a `publishAuthorizationEvent` implementation that uses `AuthorizeResult`.

jzheaux (changes requested):
Thanks for the updates and for your patience as I got back to this PR, @CrazyParanoid. I've left my feedback inilne.

jzheaux:
@CrazyParanoid, also if you have time, once we are aligned on the servlet changes, it would be great if the PR could have the reactive bits as well. I'm happy to add a polish for that if needed.

CrazyParanoid:
Hi @jzheaux. Thanks for your feedback! I will complete this issue in the next few days.

core/src/main/java/org/springframework/security/authorization/method/AuthorizationManagerBeforeMethodInterceptor.java
CrazyParanoid: I can add a new method:
```
	default <T> void publishAuthorizationEvent(Supplier<Authentication> authentication, T object,
			AuthorizationResult decision) {
		publishAuthorizationEvent(authentication, object,(AuthorizationDecision) decision);
	}
```
Method with `AuthorizationDecision` semantics can be declared as deprecated. This will not break backward compatibility with existing `AuthorizationEventPublisher` implementations

CrazyParanoid: @jzheaux I suggest adding an `AuthorizationEventPublisher` implementation that will not publish events, like `NoopAuthorizationEventPublisher`:
```
public final class NoopAuthorizationEventPublisher implements AuthorizationEventPublisher {

	@Override
	public <T> void publishAuthorizationEvent(Supplier<Authentication> authentication, T object, AuthorizationDecision decision) {

	}

	@Override
	public <T> void publishAuthorizationEvent(Supplier<Authentication> authentication, T object, AuthorizationResult decision) {
	}
}
```
`AuthorizationEventPublisher` is not marked as a functional interface, but in fact it is one. `AuthorizationManagerAfterMethodInterceptor` and `AuthorizationFilter` use `noPublish` method to implement the publisher. This means that calling `this.eventPublisher.publishAuthorizationEvent(this::getAuthentication, request, decision);` will require the `AuthorizationDecision` type explicitly. At the same time, we cannot remove the `publishAuthorizationEvent` method with `AuthorizationDecision` semantic. Therefore, it makes sense to replace `eventPublisher` in these components with `NoopAuthorizationEventPublisher`.

core/src/main/java/org/springframework/security/authorization/AuthorizationEventPublisher.java
```diff
@@ -42,8 +42,26 @@ public interface AuthorizationEventPublisher {
...
+	 * implementations may choose to drop success events by default.
+	 * @param authentication a {@link Supplier} for the current user
+	 * @param object the secured object
+	 * @param decision {@link AuthorizationResult} the decision about whether the user may access the secured object
+	 * @param <T> the secured object's type
+	 * @since 6.4
+	 */
+	default <T> void publishAuthorizationEvent(Supplier<Authentication> authentication, T object,
+			AuthorizationResult decision) {
+		publishAuthorizationEvent(authentication, object,(AuthorizationDecision) decision);
```
jzheaux, line 65: Even though it's a default implementation, we should perform a type-safe check. Maybe you could do the following:

```java
if (!(result instanceof AuthorizationDecision decision)) {
    throw new UnsupportedOperationException("result must be of type AuthorizationDecision");
}
publishAuthorizationEvent(authentication, object, decision);
```

core/src/main/java/org/springframework/security/authorization/AuthorizationObservationContext.java
```diff
@@ -73,15 +73,32 @@ public T getObject() {
 	 * @return the observed {@link AuthorizationDecision}
 	 */
 	public AuthorizationDecision getDecision() {
```
jzheaux, line 75: Let's deprecate this as well.

core/src/main/java/org/springframework/security/authorization/AuthorizationObservationContext.java
```diff
@@ -73,15 +73,32 @@ public T getObject() {
...
-		return this.decision;
+		return (AuthorizationDecision) this.authorizationResult;
 	}
 
 	/**
 	 * Set the observed {@link AuthorizationDecision}
 	 * @param decision the observed {@link AuthorizationDecision}
 	 */
 	public void setDecision(AuthorizationDecision decision) {
```
jzheaux, line 83: Let's deprecate this as well

core/src/main/java/org/springframework/security/authorization/AuthorizationObservationContext.java
```diff
@@ -73,15 +73,32 @@ public T getObject() {
 	 * @return the observed {@link AuthorizationDecision}
 	 */
 	public AuthorizationDecision getDecision() {
-		return this.decision;
+		return (AuthorizationDecision) this.authorizationResult;
```
jzheaux, line 76: Please perform a type check here. The reason is so that the application can give a more informative error than `ClassCastException`. Maybe something like:

```java
Assert.isInstanceOf(AuthorizationDecision.class, "please call getAuthorizationResult instead. If you must call getDecision, please ensure that the result you provide is of type AuthorizationDecision");
return (AuthorizationDecision) this.authorizationResult;
```

core/src/main/java/org/springframework/security/authorization/NoopAuthorizationEventPublisher.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.springframework.security.core.Authentication;
+
+/**
+ * An {@link AuthorizationEventPublisher} implementation that does not authorization publish events
+ *
+ * @author Max Batischev
+ * @since 6.4
+ */
+public final class NoopAuthorizationEventPublisher implements AuthorizationEventPublisher {
```
jzheaux, line 29: I'd rather not introduce this in this PR. If you are able to make this package-private, I don't mind leaving it. Otherwise, let's stick with private inner classes for now, even though that's a bit of duplication.

core/src/test/java/org/springframework/security/authorization/method/AuthorizationManagerBeforeMethodInterceptorTests.java
```diff
@@ -72,7 +72,7 @@ public void beforeWhenMockAuthorizationManagerThenCheck() throws Throwable {
 		AuthorizationManagerBeforeMethodInterceptor advice = new AuthorizationManagerBeforeMethodInterceptor(
 				Pointcut.TRUE, mockAuthorizationManager);
 		advice.invoke(mockMethodInvocation);
-		verify(mockAuthorizationManager).check(any(Supplier.class), eq(mockMethodInvocation));
+		verify(mockAuthorizationManager).authorize(any(Supplier.class), eq(mockMethodInvocation));
```
jzheaux, line 75: I'd prefer these change to have Mockito call the real `authorize` method, thus confirming that folks upgrading can have the same expectations. For example:

```java
given(mockAuthorizationManager.authorize(any(), any()).willCallRealMethod();
```

Then, this last line in the test shouldn't have to change.

Please see if this strategy works and, if so, apply it to the remaining tests so that their verification steps stay the same.

jzheaux (changes requested):
Looking good, @CrazyParanoid, we're nearly there. I've left some additional feedback inline.

jzheaux:
Additionally, the build appears to be failing due to formatting concerns. Please try running from the command line the following:

```
./gradlew format && ./gradlew checkstyleMain checkstyleTest
```

This will correct what it can and then give you a report of the things that you need to change manually from a code convention standpoint.

CrazyParanoid:
Hi @jzheaux , thanks for your feedback! I'll get back to working on this issue soon. But I'm still a little worried about the correctness of the chosen solution - if the `authorize` method is marked as default, it turns out that all lambdas will implement the `check` method, is this a problem and how easy will it be to switch all lambdas to the `authorize` method?

jzheaux:
Good point, @CrazyParanoid, though I think this is something we'll have to take in stride. We don't want to break folks when they upgrade, which is what is driving the decision to mark `authorize` as `default`.

I believe the main consequence is that they will not be able to use a lambda and a custom implementation of `AuthorizationResult` together. Since that feature isn't in Spring Security yet, no one has lost anything.

Either way, it should not be an issue for very long, given that Spring Security 7 will be released next fall where we can remove the `check` method.

How well does that address your concern?

CrazyParanoid:
I'm not sure if this is type safe. For example, we started using `authorize` instead of `check` and get `AuthrorizationResult`. In this case, `eventPublisher` is implemented as a `noPublish` lambda, this lambda requires an object of type `AuthorizationDecision`, but there is no guarantee that `authorize` will return exactly `AuthorizationDecision`, it can return any `AuthrorizationResult`. Which will immediately result in an error in `noPublish`.

CrazyParanoid:
This concerns the publisher, I think that, as you said, it would be better to open another ticket for `NoOpAuthorizationEventPublisher`.

jzheaux:
I agree that the no-publish lambda may not be able to stay. A concrete implementation will likely improve readability as well. It just needs to remain package-private or private (inline) for now.

jzheaux (approved):

jzheaux:
Nice, @CrazyParanoid! This for all this work. I ended up polishing the event support a bit more than expected, so I opened a separate ticket and adjusted the commits accordingly. I also added some polish, largely to remove deprecated usages and references to things like `AuthorizationResult decision`.

This is now merged into `main`.

Issue #11157, opened 2022-04-20
Improve CONTRIBUTING doc


marcusdacoregio:
Some users are having a hard time configuring their environment to work with the Spring Security codebase.

We should review and improve the CONTRIBUTING doc in order to avoid contributors wasting time with stuff like formatting, imports, etc.

See:
 - https://github.com/spring-projects/spring-security/issues/11073#issuecomment-1103116291

jzheaux:
I think setting up the dev environment should go in a separate document. The reason is that it branches out based on your IDE and perhaps your OS. I think [Spring Framework's Build from Source](https://github.com/spring-projects/spring-framework/wiki/Build-from-Source) document could be a guide.

Crain-32:
As I just experienced the horrors of getting the formatting correct, I'd like to say it appears IntelliJ now supports the necessary import [configurations](https://www.jetbrains.com/help/idea/code-style-java.html#imports_table). 

For development environments Spring as a whole could provide simple [dev containers](https://containers.dev/) or [devfiles](https://devfile.io/docs/2.2.0/devfile-ecosystem) with the expected tools (Gradle, JVM Versions, etc) to contribute. (I will admit I haven't reviewed if other issues have covered this topic previously)

PR #14859, opened 2024-04-05
Simplify Disabling application/x-www-form-urlencoded Encoding Client ID and Secret

### LINKED ISSUE #11440: Simplify Disabling application/x-www-form-urlencoded Encoding Client ID and Secret
There are quite a few authorization servers (see how many users have commented on gh-10018) that do not URL Encode the Client ID and Secret as [required by the OAuth specification](https://github.com/spring-projects/spring-security/issues/10018#issuecomment-873132661). There is a way to [customize the configuration ](https://github.com/spring-projects/spring-security/issues/10018#issuecomment-877523592) to support this use-case, but is quite verbose given the number of users experiencing the problem. We should provide an easier way to disable URL encoding.

Crain-32:
Closes #11440 , includes a new test that covers both the URL encoding and URL non-encoding.

Force Push was to keep this at one commit, and I forgot to follow the formatting. 🤦🏻‍♂️ 

I am really bad at following this formatter, I apologize for any notifications this sends.


Crain-32:
I have no idea why the formatter is failing on the imports in the `DefaultOAuth2TokenRequestHeadersConverter`. I'll need someone else to review that.

sjohnr:
> I have no idea why the formatter is failing on the imports in the `DefaultOAuth2TokenRequestHeadersConverter`. I'll need someone else to review that.

I checked the diff, and it looks like the `java.nio.charset.StandardCharsets` import is not correctly ordered within the import list. You can simply move it back to its original position.

Crain-32:
That didn't fix it, I'm begin to think it's the function order?

Honestly since this is a small PR I'm going to let it sit for a bit and dig into #11157 and the IntelliJ formatting to see if we can get IntelliJ to auto format correctly.

sjohnr:
@Crain-32 according to the build output, you can simply run `./gradlew format` to fix formatting issues, and try the build again.

Crain-32:
My brain was reading that as using `format` to check the formatting 🤦🏻‍♂️ 

Fixed.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultOAuth2TokenRequestHeadersConverter.java
```diff
@@ -29,50 +28,54 @@
...
-	private static HttpHeaders DEFAULT_TOKEN_REQUEST_HEADERS = getDefaultTokenRequestHeaders();
+	private static final HttpHeaders DEFAULT_TOKEN_HEADERS = getDefaultTokenRequestHeaders();
 
-	private OAuth2AuthorizationGrantRequestEntityUtils() {
+	private boolean encodeClientCredentials = true;
+
+	private static HttpHeaders getDefaultTokenRequestHeaders() {
+		HttpHeaders headers = new HttpHeaders();
+		headers.setAccept(Collections.singletonList(MediaType.APPLICATION_JSON_UTF8));
+		final MediaType contentType = MediaType.valueOf(MediaType.APPLICATION_FORM_URLENCODED_VALUE + ";charset=UTF-8");
```
sjohnr, line 52: In order to bring this into alignment with the corresponding logic on the reactive side (see `AbstractWebClientReactiveOAuth2AccessTokenResponseClient#populateTokenRequestHeaders`), I wonder if there's anything we can do about this `charset=UTF-8`? For backwards compatibility, it might be nice to make this additional param on the `Content-Type` configurable. If we do this, we can re-use this exact implementation for the reactive side as well. (see opening comment)

docs/modules/ROOT/pages/servlet/oauth2/client/client-authentication.adoc
```diff
@@ -92,7 +92,9 @@ val tokenResponseClient = DefaultAuthorizationCodeTokenResponseClient()
 tokenResponseClient.setRequestEntityConverter(requestEntityConverter)
 ----
 ======
-
+[NOTE]
+If you're using the `client-authentication-method: client_secret_basic` and you need to skip URL encoding,
+create a new `DefaultOAuth2TokenRequestHeadersConverter` and set it in the Request Entity Converter above.
```
sjohnr, line 97: This note seems to be added in the "Authenticate using `private_key_jwt" section? If so, this wouldn't be the best place to add this note. Perhaps it could be a `[TIP]` elsewhere? For example, it could be added to `authorization-grants.adoc` where we cover customizing the access token request for each grant type. But this would require adding the tip 5 times because we unfortunately have structured the content that way currently. Or we could (for now) add an "Authenticate using `client_secret_basic`" section on this page.

sjohnr (changes requested):
Thanks for the PR and work on this so far @Crain-32! I have some comments inline below.

Additionally, I feel that we should provide equivalent support for the reactive side (`AbstractWebClientReactiveOAuth2AccessTokenResponseClient`). It is already set up to allow customizing the headers converter. I've added a comment below regarding the `charset=UTF-8` in the `Content-Type` which is the only discrepancy I noticed between the two implementations. If that bit is made configurable, would you be able to add this to `AbstractWebClientReactiveOAuth2AccessTokenResponseClient` as well (and add corresponding tests)? If it's too much, I can help out with that as a follow up.

Thanks!

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/endpoint/OAuth2PasswordGrantRequestEntityConverterTests.java
```diff
@@ -130,4 +135,33 @@ public void convertWhenGrantRequestValidThenConverts() {
 		assertThat(formParameters.getFirst(OAuth2ParameterNames.SCOPE)).contains(clientRegistration.getScopes());
 	}
 
+	@SuppressWarnings("unchecked")
+	@Test
+	public void convertWhenGrantRequestValidThenConvertsWithoutUrlEncoding() {
```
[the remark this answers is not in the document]
sjohnr: Actually, I think the existing tests are already cover it. I don't think we need to add grant-type specific tests at all. So disregard the above comment.

Crain-32:
Sorry, took this week off in general.
I'll keep to this tomorrow.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultOAuth2TokenRequestHeadersConverter.java
```diff
@@ -29,50 +28,54 @@
...
- * @see OAuth2AuthorizationCodeGrantRequestEntityConverter
+ * @since 6.3
  * @see OAuth2ClientCredentialsGrantRequestEntityConverter
  */
-final class OAuth2AuthorizationGrantRequestEntityUtils {
+public class DefaultOAuth2TokenRequestHeadersConverter<T extends AbstractOAuth2AuthorizationGrantRequest>
+		implements Converter<T, HttpHeaders> {
 
-	private static HttpHeaders DEFAULT_TOKEN_REQUEST_HEADERS = getDefaultTokenRequestHeaders();
+	private static final HttpHeaders DEFAULT_TOKEN_HEADERS = getDefaultTokenRequestHeaders();
```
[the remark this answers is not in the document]
Crain-32: I've swapped it to also having a `setter` and no longer being `static final`.
This also meant I felt a lot better adjusting to using just `Accept:  application/json`, since if there is a real need to go back to `application/json;charset=UTF-8` or `application/x-www-form-urlencoded;charset=UTF-8` for consumers, they can configure that the same way they can configure the `encodeClientCredentialsIfRequired`

docs/modules/ROOT/pages/servlet/oauth2/client/client-authentication.adoc
Crain-32: I'll format and push the code, and we can discuss the documentation side after we feel good about the integration of the `DefaultOAuthtokenREquestHeadersConverter` into the Reactive and Non-Reactive side. Since I can see that impacting the documentation we set up.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultOAuth2TokenRequestHeadersConverter.java
Crain-32: Removing the `charset=UTF-8` breaks about 19 tests. Do you want me to remove that check from the tests? Or maintain a default of `charset=UTF-8`? (More information on a different conversation)

sjohnr: @Crain-32 We don't want to change the behavior for the servlet side (tests should not be changed). But `charset=UTF-8` should not be the default for the new class either. Instead, the `AbstractOAuth2AuthorizationGrantRequestEntityConverter` should be able to easily create an instance of this new class with the addition of `charset=UTF-8` enabled. Then the reactive version would just instantiate and use the class with defaults.

The best way to do this IMHO would be to create a package-private static factory method in this new class that instantiates the class and enables the customized `charset`, which is used by `AbstractOAuth2AuthorizationGrantRequestEntityConverter`. Make sense?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultOAuth2TokenRequestHeadersConverter.java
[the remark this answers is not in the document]
sjohnr: I don't think I prefer to add `setDefaultHeaders(...)`. Instead, we can add fields:

* `private MediaType accept` (defaults to `MediaType.APPLICATION_JSON`)
* `private MediaType contentType` (defaults to `MediaType.APPLICATION_FORM_URLENCODED`)

and private methods:

* `setAccept(MediaType)`
* `setContentType(MediaType)`

These methods will only be called by the new package-private static factory method (see below conversation). The `static final` and `getDefaultTokenRequestHeaders()` can be removed.

docs/modules/ROOT/pages/servlet/oauth2/client/client-authentication.adoc
sjohnr: Let's fix things so that tests are not broken (see above recommendations). As mentioned above, let's also remove `getDefaultTokenRequestHeaders()`.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/DefaultOAuth2TokenRequestHeadersConverter.java
Crain-32: @sjohnr it does make sense to use a factory. Would the preferred naming be something like `....HeaderConverterCustomizer`, or `.....HeaderConverterFactory`? The first feels more in line with the general idea of "Customizer" the JWT Client customizer does, where the second one feels more generic to the Java ecosystem.


Follow up question on that though, would it be better to expose the factory/builder within the public scope instead of the class?
Current plan for customization was just having the consumer use `new Default.....()`, but if we're having the internals use a factory, might as well expose the factory instead of the class.

sjohnr: > Would the preferred naming be something like `....HeaderConverterCustomizer`, or `.....HeaderConverterFactory`?

It should be a package-private static method, not a class. The name doesn't exactly matter because it will be hidden at the package level. However, `withCharsetUtf8()` would be my preferred name, as it tells the story fairly well.

> Current plan for customization was just having the consumer use `new Default.....()`, but if we're having the internals use a factory, might as well expose the factory instead of the class.

I don't think that we should expose this. The charset customization is only being used to make this change non-breaking. The `charset=UTF-8` variants in `MediaType` are deprecated and should not be used. We don't want to encourage their use, so keeping this customization internal and only used by default implementations provided by the framework is preferred. Said another way, this customization should stay hidden in the framework and only used to prevent breaking tests (e.g. backwards compatibility).

testersen:
With this pull request, will there be a configuration we can tweak in the application properties/yaml file we can use to disable url encoding of the identifier? Something like `spring.security.oauth2.client.provider.my-provider.urlencode-identifier = false`?

sjohnr:
@testersen configuration properties are a [spring boot](https://github.com/spring-projects/spring-boot/) feature, so we won't be able to add anything like that here. I don't think that would be the preferred method way to customize anyway, since it is authentication method specific.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/endpoint/OAuth2PasswordGrantRequestEntityConverterTests.java
```diff
@@ -110,9 +110,14 @@ public void convertWhenParametersConverterSetThenCalled() {
 	@SuppressWarnings("unchecked")
 	@Test
 	public void convertWhenGrantRequestValidThenConverts() {
-		ClientRegistration clientRegistration = TestClientRegistrations.password().build();
+		ClientRegistration clientRegistration = TestClientRegistrations.password()
+			.clientId("clientId")
+			.clientSecret("clientSecret=")
+			.build();
```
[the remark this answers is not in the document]
Crain-32: Since the `convertWhenGrantRequestValidThenConvertsWithoutUrlEncoding` is the test that confirms the URL encoding toggle works, it felt like the correct move would be to include a slightly change on an earlier test that confirms the URL encoding default was enabled.

This felt better than copying and pasting an entire test just to add in a `=`, without any other changes. Where as the `convertWhenGrantRequestValidThenConvertsWithoutUrlEncoding` is a copy+paste of another test with the `=` added, and the toggle swapped.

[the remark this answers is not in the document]
Crain-32: Likewise the historical tests + this minor change mean the class has coverage, which sure is a little bit of a weak argument, but adding in several tests for this feels a bit much.

Crain-32:
That last push *should* have covered all the code changes. I'm not sure if `historicalConverter` was the best name for the simple factory, but it sounded good, so I went with it.

Figured it would be best to get an okay on the code -> update docs -> rebase on master -> finalize.

sjohnr:
Thanks for the updates @Crain-32! Since we're wrapping up the release cycle and I felt like this could still fit in this release, I went ahead and merged this PR as d0adb2aa700501fe618d6d9a6b9a5cdc1ac66147 with polish commit 2598bf8c379e2462de86daa18139d71f8c7ef9ba. I also updated the docs in 2dd908dff8bb5ebf376912ea02fadb885a5877fd.

Crain-32:
@sjohnr Thanks for doing that! I know this PR/Issue was a little long lived, but I'm grateful for the feedback and finishing it out for the release. Ideally the next one will go smoother.

Issue #14870, opened 2024-04-09
The continueOnError value is inconsistent with that described in the java doc


kse-music:
**Describe the bug**
The code line [setContinueOnError](https://github.com/spring-projects/spring-security/blame/main/web/src/main/java/org/springframework/security/web/server/authentication/DelegatingServerAuthenticationConverter.java#L68)
The default value of continueOnError is false, but the java doc describes the default value as true

marcusdacoregio:
Closed in favor of https://github.com/spring-projects/spring-security/pull/14871

PR #14871, opened 2024-04-09
Fix continueOnError default value in java doc

kse-music:
Closes gh-14870


marcusdacoregio:
Thanks, @kse-music. This is now merged into `main`

Issue #14600, opened 2024-02-13
Method Security SpEL supports throwing AccessDeniedException


rwinch:
It would be nice if we supported throwing `AccessDeniedException` for Method Security SpEL. This would allow method security to easily convey more information to the framework on why authorization failed (e.g. requires MFA). They can also use AuthorizationManager instances in their SpEL. For example:

```java
@Service
class Authz {
  AuthorizationDecision check(...) {
    throw new AccessDeniedException("Denied");
  }
}
```

```java
@PreAuthorize("@authz.check(...)")
```

PR #14882, opened 2024-04-10
Handle SpEL AuthorizationDeniedExceptions

jzheaux:
Closes gh-14600

### DIFF
--- a/core/src/main/java/org/springframework/security/authorization/method/AuthorizationManagerAfterMethodInterceptor.java
+++ b/core/src/main/java/org/springframework/security/authorization/method/AuthorizationManagerAfterMethodInterceptor.java
@@ -30,6 +30,7 @@ import org.springframework.security.access.AccessDeniedException;
 import org.springframework.security.access.prepost.PostAuthorize;
 import org.springframework.security.authentication.AuthenticationCredentialsNotFoundException;
 import org.springframework.security.authorization.AuthorizationDecision;
+import org.springframework.security.authorization.AuthorizationDeniedException;
 import org.springframework.security.authorization.AuthorizationEventPublisher;
 import org.springframework.security.authorization.AuthorizationManager;
 import org.springframework.security.core.Authentication;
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/springframework/security/authorization/method/AuthorizationManagerAfterMethodInterceptor.java
```diff
@@ -173,16 +174,28 @@ private Object attemptAuthorization(MethodInvocation mi, Object result) {
 		this.logger.debug(LogMessage.of(() -> "Authorizing method invocation " + mi));
 		MethodInvocationResult object = new MethodInvocationResult(mi, result);
 		AuthorizationDecision decision = this.authorizationManager.check(this::getAuthentication, object);
-		this.eventPublisher.publishAuthorizationEvent(this::getAuthentication, object, decision);
-		if (decision != null && !decision.isGranted()) {
-			this.logger.debug(LogMessage.of(() -> "Failed to authorize " + mi + " with authorization manager "
-					+ this.authorizationManager + " and decision " + decision));
-			return postProcess(object, decision);
+		try {
```
marcusdacoregio, line 177: Shouldn't this be moved to the line above?

core/src/main/java/org/springframework/security/authorization/method/AuthorizationManagerAfterReactiveMethodInterceptor.java
```diff
@@ -151,7 +152,23 @@ private Mono<Object> postAuthorize(Mono<Authentication> authentication, MethodIn
 		MethodInvocationResult invocationResult = new MethodInvocationResult(mi, result);
 		return this.authorizationManager.check(authentication, invocationResult)
 			.switchIfEmpty(Mono.just(new AuthorizationDecision(false)))
-			.flatMap((decision) -> postProcess(decision, invocationResult));
+			.flatMap((decision) -> postProcess(decision, invocationResult))
+			.onErrorResume(AuthorizationDeniedException.class, (denied) -> postProcess(denied, invocationResult));
```
marcusdacoregio, line 156: Does this also handle if `postProcess` throws a `AuthorizationDeniedException`?

marcusdacoregio (changes requested):
Thanks @jzheaux. I've left some feedback inline.

core/src/main/java/org/springframework/security/authorization/method/AuthorizationManagerAfterMethodInterceptor.java
jzheaux: Thanks, good catch

core/src/main/java/org/springframework/security/authorization/method/AuthorizationManagerAfterReactiveMethodInterceptor.java
jzheaux: Good point. I've updated the logic.

Issue #14269, opened 2023-12-11
Add reasonable timeout to JwtDecoderProviderConfigurationUtils and NimbusJwtDecoder


vonnahme:
**Expected Behavior**

These classes should use reasonable default timeouts to avoid the possibility of a connection hanging.

**Current Behavior**

A default RestTemplate with no timeout configured is used.

https://github.com/spring-projects/spring-security/blob/6.2.0/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoderProviderConfigurationUtils.java#L66

https://github.com/spring-projects/spring-security/blob/6.2.0/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java#L271

**Context**

We had network issues recently and saw hung threads in these classes due to the connections never being released.  

It appears that nimbus sets a default timeout on http connections they make: https://www.javadoc.io/static/com.nimbusds/nimbus-jose-jwt/9.37.3/com/nimbusds/jose/jwk/source/JWKSourceBuilder.html#DEFAULT_HTTP_CONNECT_TIMEOUT

Since both of these classes rely on nimbus, perhaps the nimbus timeout settings could be re-used?

After researching this a bit, it does seem I could configure this at the JVM by setting 
```
-Dsun.net.client.defaultConnectTimeout=
-Dsun.net.client.defaultReadTimeout=
```
but that's more broad than I would prefer.

jzheaux:
Related to https://github.com/spring-projects/spring-security/pull/11232

vonnahme:
> Thanks for reaching out, @vonnahme. I think it makes sense to provide a timeout, at least to the places where RestTemplate is not configurable like JwtDecoderProviderConfigurationUtils.

Great!

> That said, I'm not sure yet if this would help your situation.

It definitely would have :-)  When a thread gets hung indefinitely, thread pools fill up, the server becomes unresponsive, and teams are scrambling to restart things faster than they can fill up.  Having the timeout wouldn't have stopped the requests from hanging, but would have helped the infrastructure stay healthy / the failure would have stayed local rather than cascading throughout the environment.

> Can you please provide how you are presently constructing a NimbusJwtDecoder instance in your application?

We set the property
```
spring.security.oauth2.resourceserver.jwt.issuer-uri
```
[...]

PR #14890, opened 2024-04-11
Add timeout defaults to JwtDecoders

MrJovanovic13:
This PR adds reasonable timeouts to JwtDecoderProviderConfigurationUtils and NimbusJwtDecoder.
It's fixing [Issue GH-14269](https://github.com/spring-projects/spring-security/issues/14269)

Since the `RestOperations` instance in `JwkSetUriJwtDecoderBuilder` is not static, I couldn't have used a static initializer block and had to use the following solution. I am open to suggestions if you think there is a better solution for this.

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoderProviderConfigurationUtils.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoderProviderConfigurationUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -39,6 +39,7 @@ import com.nimbusds.jwt.proc.ConfigurableJWTProcessor;
 import org.springframework.core.ParameterizedTypeReference;
 import org.springframework.http.RequestEntity;
 import org.springframework.http.ResponseEntity;
+import org.springframework.http.client.SimpleClientHttpRequestFactory;
 import org.springframework.security.oauth2.jose.jws.SignatureAlgorithm;
 import org.springframework.util.Assert;
 import org.springframework.web.client.HttpClientErrorException;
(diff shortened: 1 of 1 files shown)


oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtDecoderProviderConfigurationUtils.java
```diff
@@ -39,6 +39,7 @@
 import org.springframework.core.ParameterizedTypeReference;
 import org.springframework.http.RequestEntity;
 import org.springframework.http.ResponseEntity;
+import org.springframework.http.client.SimpleClientHttpRequestFactory;
```
jzheaux, line 42: Will you please update the copyright year to 2024?

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -271,7 +272,16 @@ public static final class JwkSetUriJwtDecoderBuilder {
 
 		private Set<SignatureAlgorithm> signatureAlgorithms = new HashSet<>();
 
-		private RestOperations restOperations = new RestTemplate();
+		private RestOperations restOperations = this.buildRestTemplateWithDefaultTimeoutValues();
+
+		private RestTemplate buildRestTemplateWithDefaultTimeoutValues() {
```
jzheaux, line 277: For now, will you please isolate your changes to `JwtDecoderProviderConfigurationUtils` since that `RestTemplate` is otherwise not configurable? The reason I say that is that I'd prefer to rely on `RestTemplate` for defaults as a general rule.

jzheaux (changes requested):
Thanks, @MrJovanovic13, for the PR! In addition to my inline feedback, will you also please update your commit message to be formatted like so:

```
Add Default Timeout to JwtDecoders RestTemplate

Any other description you'd like to add that wraps roughly at 60
characters

Closes gh-14269
```

In this way, it causes the automation to close the related ticket and the short title also makes it easy to browse in IDEs.

MrJovanovic13:
Thanks for the comments, I have resolved both of them now.

PR #14899, opened 2024-04-12
Make ``OidcUserService`` overrideable

### LINKED ISSUE #14898: Make OidcUserService overrideable
**Expected Behavior**

I would like to extend [``OidcUserService``](https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserService.java) and override certain methods e.g. ``shouldRetrieveUserInfo``

**Current Behavior**

You can't override like ~80% of the class because it's method/fields ``final`` or ``private``

**Context**

> How has this issue affected you?
> What are you trying to accomplish?
> What other alternatives have you considered?
> Are you aware of any workarounds?

I have to copy paste the complete class -which will likely break when I update the library.
Or I get the reflection "crowbar" and make methods somehow accessible.
Or I create a lot of issues here that are related to my use-case which just costs a lot of time and effort and then I still have to wait for a release.

It also doesn't matter that this was changed in https://github.com/spring-projects/spring-security/commit/96e3e4f8b1ff9545772a7384ff67d490484178db, this is a general problem.

AB-xdev:
Fixes #14898

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserService.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserService.java
@@ -67,22 +67,22 @@ import org.springframework.util.StringUtils;
  */
 public class OidcUserService implements OAuth2UserService<OidcUserRequest, OidcUser> {
 
-	private static final String INVALID_USER_INFO_RESPONSE_ERROR_CODE = "invalid_user_info_response";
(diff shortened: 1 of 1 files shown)


sjohnr:
@AB-xdev thanks for the PR! I appreciate [your comment](https://github.com/spring-projects/spring-security/issues/14898#issue-2239836416) and the fact that discussing your use case costs a lot of time. However, opening up classes to extension in this way isn't the solution that we're looking for. As such, I'm going to close this PR. Please consider discussing your use case and what you're trying to accomplish by adding another comment to the issue (I assigned to myself).

PR #14902, opened 2024-04-13
Fix ArrayIndexOutOfBoundsException in XorCsrfTokenRequestAttributeHan…

kratosmy:
To fix gh-#13310, I make a hotfix to ensure that `System.arraycopy()` won't cause a `ArrayIndexOutOfBoundsException`.

### DIFF
--- a/crypto/src/main/java/org/springframework/security/crypto/password/DelegatingPasswordEncoder.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/password/DelegatingPasswordEncoder.java
@@ -19,8 +19,6 @@ package org.springframework.security.crypto.password;
 import java.util.HashMap;
 import java.util.Map;
 
-import org.springframework.util.StringUtils;
-
 /**
  * A password encoder that delegates to another PasswordEncoder based upon a prefixed
  * identifier.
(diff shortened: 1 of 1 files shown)


jzheaux:
Thanks, @kratosmy! Will you please add a unit test that would fail without your change and pass with your change in place?

kratosmy:
> Thanks, @kratosmy! Will you please add a unit test that would fail without your change and pass with your change in place?

sure!let me add it

kratosmy:
> Thanks, @kratosmy! Will you please add a unit test that would fail without your change and pass with your change in place?

Hi, @jzheaux , I think it is hard to write a unit test for this change because it is regarding to a call to `System.arraycopy()`, which is a base native function in the middle of a static function.

jzheaux:
@kratosmy, I think it's okay to proceed without mocking `System.arraycopy`. It seems like a test where the csrf byte array is longer than the random byte array will suffice. Does that help you to proceed?

kratosmy:
> @kratosmy, I think it's okay to proceed without mocking `System.arraycopy`. It seems like a test where the csrf byte array is longer than the random byte array will suffice. Does that help you to proceed?

Sure will do.

kratosmy:
@jzheaux Hi, I have opened another PR (https://github.com/spring-projects/spring-security/pull/14976) using a new branch name to align with contribution guideline.

kratosmy:
And I also fix the same code in another package: `org.springframework.security.messaging.web.csrf`, please kindly help review, thanks a lot!

Issue #14801, opened 2024-03-25
Sign SP Metadata


siddharth-78:
The EOL SAML version provided a way to sign SP metadata like this:

```
 ExtendedMetadata extendedMetadata = generator.generateExtendedMetadata();
    if (!StringUtils.isEmpty(entityAlias)) {
      extendedMetadata.setAlias(entityAlias);
    }
    extendedMetadata.setSignMetadata(true);
```

I don't find anything of this fashion with the newer saml version (saml2-service-provider 5.8.9), has the support for signing SP metadata dropped? Or am I missing something?

jzheaux:
Hi, @siddharth-78, no this has not been added yet. Are you able to contribute a PR to add `setSignMetadata(boolean)` to `OpenSamlMetadataResolver`? It may be as straightforward as copying `OpenSamlSigningUtils` and `Saml2Utils` into the corresponding package and then calling `OpenSamlSigningUtils#sign` before serializing the value.

Issue #13661, opened 2023-08-18
How to Add a Signature to Service Provider Metadata


sumeetpri:
I'm currently in the process of migrating my application's authentication from Spring Security SAML Extension to Spring Security SAML. I'm using the service provider metadata to register my application with the Identity Provider (IDP). I've managed to generate the metadata for the service provider, but I'm struggling to figure out how to sign the metadata details.

In the previously generated service provider metadata, you can see the presence of the `<ds:Signature></ds:Signature>` section, which was automatically generated by Spring Security SAML Extension. However, in my migration to Spring Security SAML, I haven't found an implementation for signing the metadata in the `OpenSamlMetadataResolver`.

Here's a snippet of my code:

```java
SecurityFilterChain configure(HttpSecurity http) throws Exception {
    OpenSaml4AuthenticationProvider authenticationProvider = new OpenSaml4AuthenticationProvider();

    Saml2MetadataFilter filter = new Saml2MetadataFilter(relyingPartyRegistrationRepository,
        new OpenSamlMetadataResolver());
    http.addFilterBefore(filter, Saml2WebSsoAuthenticationFilter.class);

    http.authorizeHttpRequests(requests -> requests
        .requestMatchers("/saml2/service-provider-metadata/**")
        .permitAll()
    ).saml2Login((saml2) -> saml2.loginProcessingUrl(SAML2_ASSERTION_CONSUMER_SERVICE_URL))
    .build();
}
```

The service provider metadata generated by Spring Security SAML Extension includes a `<ds:Signature>` section as follows:

```xml
<md:EntityDescriptor xmlns:md="urn:oasis:names:tc:SAML:2.0:metadata" ID="example.com" entityID="example.com">
    <ds:Signature xmlns:ds="http://www.w3.org/2000/09/xmldsig#">
        <!-- Signature details here -->
    </ds:Signature>
    <!-- Other metadata details -->
</md:EntityDescriptor>
```

My question is, how can I achieve the same feature of generating and adding the `<ds:Signature>` section in the service provider metadata using Spring Security SAML?

I've already reviewed the `OpenSamlMetadataResolver` implementation, but I couldn't find a built-in way to sign the metadata. Any guidance or code examples on how to achieve this would be greatly appreciated.

sumeetpri:
I am able to add signature , so closing this open issue .

arotondo:
[...]

PR #14916, opened 2024-04-17
Support signing SAML metadata

CrazyParanoid:
Closes gh-14801

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/metadata/OpenSamlMetadataResolver.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/metadata/OpenSamlMetadataResolver.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -74,6 +74,8 @@ public final class OpenSamlMetadataResolver implements Saml2MetadataResolver {
 
 	private boolean usePrettyPrint = true;
 
+	private boolean signMetadata = false;
+
 	public OpenSamlMetadataResolver() {
 		this.entityDescriptorMarshaller = (EntityDescriptorMarshaller) XMLObjectProviderRegistrySupport
 			.getMarshallerFactory()
@@ -111,6 +113,9 @@ public final class OpenSamlMetadataResolver implements Saml2MetadataResolver {
 		SPSSODescriptor spSsoDescriptor = buildSpSsoDescriptor(registration);
 		entityDescriptor.getRoleDescriptors(SPSSODescriptor.DEFAULT_ELEMENT_NAME).add(spSsoDescriptor);
 		this.entityDescriptorCustomizer.accept(new EntityDescriptorParameters(entityDescriptor, registration));
+		if (this.signMetadata) {
+			return OpenSamlSigningUtils.sign(entityDescriptor, registration);
+		}
 		return entityDescriptor;
 	}
 
@@ -128,6 +133,7 @@ public final class OpenSamlMetadataResolver implements Saml2MetadataResolver {
 	/**
 	 * Configure whether to pretty-print the metadata XML. This can be helpful when
 	 * signing the metadata payload.
+	 *
 	 * @since 6.2
 	 **/
 	public void setUsePrettyPrint(boolean usePrettyPrint) {
@@ -236,6 +242,15 @@ public final class OpenSamlMetadataResolver implements Saml2MetadataResolver {
 		catch (Exception ex) {
 			throw new Saml2Exception(ex);
 		}
+	}
+
+	/**
+	 * Configure whether to sign the metadata, defaults to {@code false}.
+	 *
+	 * @since 6.4
+	 */
+	public void setSignMetadata(boolean signMetadata) {
+		this.signMetadata = signMetadata;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/metadata/OpenSamlMetadataResolver.java
```diff
@@ -111,6 +113,9 @@ private EntityDescriptor entityDescriptor(RelyingPartyRegistration registration)
 		SPSSODescriptor spSsoDescriptor = buildSpSsoDescriptor(registration);
 		entityDescriptor.getRoleDescriptors(SPSSODescriptor.DEFAULT_ELEMENT_NAME).add(spSsoDescriptor);
 		this.entityDescriptorCustomizer.accept(new EntityDescriptorParameters(entityDescriptor, registration));
+		if (this.signMetadata) {
```
jzheaux, line 116: Instead of signing each entity descriptor, please sign the entire payload. I believe this will mean that you'll call `sign` separately [here](https://github.com/spring-projects/spring-security/pull/14916/files#diff-81a8047cc4069e0ae01dcb90c3e43e2c829bd24006e7f3df32f72310917b4eb8L101) and [here](https://github.com/spring-projects/spring-security/pull/14916/files#diff-81a8047cc4069e0ae01dcb90c3e43e2c829bd24006e7f3df32f72310917b4eb8L105).

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/metadata/OpenSamlMetadataResolver.java
```diff
@@ -238,6 +244,15 @@ private String serialize(EntitiesDescriptor entities) {
 		}
 	}
 
+	/**
+	 * Configure whether to sign the metadata.
```
jzheaux, line 248: Please indicate the default value in the JavaDoc.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/metadata/OpenSamlMetadataResolver.java
```diff
@@ -238,6 +244,15 @@ private String serialize(EntitiesDescriptor entities) {
...
+	 *
+	 * @since 6.3
```
jzheaux, line 250: Now that we are in the RC phase, this will go into the 6.4 release instead. Please change the `@since` attribute to reflect this.

jzheaux (changes requested):
Thanks, @CrazyParanoid! I've left some feedback inline.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/metadata/OpenSamlMetadataResolver.java
CrazyParanoid: Hi @jzheaux ! Thanks for your feedback. In the case of signing `EntitiesDescriptor` with a list of children, it is not clear which SP and IP pair to use. In other words, which of the `relyingPartyRegistrations` should be passed to the `sign` method?

jzheaux: Good catch! Let's leave it as it is.

CrazyParanoid:
Hi @jzheaux ! Thanks for your feedback. All your comments have been resolved.

longgt:
@CrazyParanoid thanks for your contribution.
Regarding to [Signing with OpenSAML](https://blog.samlsecurity.com/2012/11/signing-with-opensaml.html) here, the final hash should be computed right after we did marshal the EntityDescriptor at below method
https://github.com/spring-projects/spring-security/blob/e92dd02c57a527a483df86eae7163a2c93ccb608/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/metadata/OpenSamlMetadataResolver.java#L215

If we sign EntityDescriptor before calling OpenSamlMetadataResolver.serialize, the marshaller in there will create a brand new XMLSignature object.
It causes the 2 signature elements in below serialized with empty content
```
<ds:DigestValue/>
<ds:SignatureValue/>
```

Where it should be
```
<ds:DigestValue>digest signature</ds:DigestValue>
<ds:SignatureValue>signature value</ds:SignatureValue>
```

https://github.com/alexo/SAML-2.0/blob/96363df652cff76cd7a04f200a7f58b988b7bfe7/java-opensaml/opensaml-xmlsec-impl/src/main/java/org/opensaml/xml/signature/impl/SignatureMarshaller.java#L99

Beside that, the test case is check only whether **DigestValue** and **SignatureValue** is existing or not.
It's better to check whether it contains value or not, in my opinion.

I noted all debug steps at my environment (with Spring Security 6.2.2) here
https://github.com/spring-projects/spring-security/issues/13661#issuecomment-2090554273

Please correct if I misunderstood this situation.
Thanks.

CrazyParanoid:
Hi @longgt ! In the `resolveWhenRelyingPartyAndSignMetadataSetThenMetadataMatches` test the xml is serialized with:
```
<ds:DigestValue>K411jmOwS83oyZ5LDpixHzwmvi6dbmd/GAT2sW2Pdyc=</ds:DigestValue>
<ds:SignatureValue>
DQfd3Gl+04LYqDYEk3O7zwuDbA940PAA3F+BVQxSF+7QdmIEykaSL5Kj+kg3DGtpUgbQPwC7SyRz&#13;
hkTZ78v9JzNu1SP8rx5DxyrdRNS51LOQLQP1qFMeSjzEi1TXlHtp0Z35qF2dAycnIty4oZFk2zNk&#13;
Uvm2N1DkAtC5XSqKssI=
</ds:SignatureValue>
```
I can add a value check to the test, but apparently there is no problem here.

jzheaux (approved):

jzheaux:
Thanks, @CrazyParanoid! This is now merged into `main` and will go out in the next milestone release.

jzheaux:
Because I think this may become the default in the future, I made a polish of 6695784a3fe7ed9edc137fc0e095f344dbb45592 to notify that metadata is unsigned.

Issue #14917, opened 2024-04-17
Replace "Spring Boot 2.x" with "Spring Boot"


sjohnr:
**Expected Behavior**

Docs should reference "Spring Boot" without "2.x".

**Current Behavior**

The docs currently refer to "Spring Boot 2.x" in the following pages:

* `servlet/oauth2/login/core.adoc`
* `reactive/oauth2/login/core.adoc`

sjohnr:
Hi @MrJovanovic13, thank you! Yes, you may take it. If it exists elsewhere, feel free to change it there as well, though I haven't noticed other places.

PR #14919, opened 2024-04-17
Replace "Spring Boot 2.x" with "Spring Boot"

MrJovanovic13:
Closes gh-14917

I've found `Spring Boot 2.x` referenced in a few more doc files other than the ones mentioned in the issue, let me know if those should not be changed and I can revert them.


sjohnr:
Thanks @MrJovanovic13!

Issue #14954, opened 2024-04-25
The MethodAuthorizationDeniedPostProcessor does not exist


kse-music:
**Describe the bug**

The MethodAuthorizationDeniedPostProcessor described on the [setApplicationContext](https://github.com/spring-projects/spring-security/blob/main/core/src/main/java/org/springframework/security/authorization/method/PostAuthorizeExpressionAttributeRegistry.java#L84) method does not exist

PR #14955, opened 2024-04-25
Fix  MethodAuthorizationDeniedPostProcessor does not exist in java doc

kse-music:
Closes gh-14954


jzheaux:

PR #14969, opened 2024-04-26
Minor Documentation update on import needed for using Kotlin DSL

madorb:
Updates the documentation to provide a bit more explicit instruction on how to enable the Kotlin DSL. I found the existing reference to 

> import the `invoke` function

a bit less obvious then I would have thought, and despite being reasonably knowledgable, having done this thing before in the past, and having the code sample, it took me a few minutes to figure out exactly what it was referring to.

### DIFF
--- a/docs/modules/ROOT/pages/reactive/configuration/webflux.adoc
+++ b/docs/modules/ROOT/pages/reactive/configuration/webflux.adoc
@@ -132,7 +132,7 @@ class HelloWebfluxSecurityConfig {
 ======
 
 [NOTE]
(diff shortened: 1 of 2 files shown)


jzheaux:
Thanks, @madorb! WIll you please do me the favor of formatting your commit [like so](https://github.com/spring-projects/spring-security/blob/main/CONTRIBUTING.adoc#submit-a-pull-request)?

An example might be something like:

```
Clarify Package Import

Provide more explict instructions regarding the necessary import 
to make the Kotlin DSL work. For some reason it took me 10 minutes 
to figure this out based on the existing doc.
```

madorb:
@jzheaux done!

marcusdacoregio:
Thanks, @madorb. This is now merged into `main`

PR #14976, opened 2024-04-27
Fix possible ArrayIndexOutOfBoundsException in XorCsrfTokenRequestAtt…

kratosmy:
To fix gh-https://github.com/spring-projects/spring-security/issues/13310, I make a hotfix to ensure that System.arraycopy() won't cause a ArrayIndexOutOfBoundsException.

### DIFF
--- a/messaging/src/main/java/org/springframework/security/messaging/web/csrf/XorCsrfTokenUtils.java
+++ b/messaging/src/main/java/org/springframework/security/messaging/web/csrf/XorCsrfTokenUtils.java
@@ -56,13 +56,16 @@ final class XorCsrfTokenUtils {
 		System.arraycopy(actualBytes, randomBytesSize, xoredCsrf, 0, tokenSize);
 
 		byte[] csrfBytes = xorCsrf(randomBytes, xoredCsrf);
-		return Utf8.decode(csrfBytes);
+		return (csrfBytes != null) ? Utf8.decode(csrfBytes) : null;
 	}
 
-	private static byte[] xorCsrf(byte[] randomBytes, byte[] csrfBytes) {
+	static byte[] xorCsrf(byte[] randomBytes, byte[] csrfBytes) {
+		if (csrfBytes.length < randomBytes.length) {
+			return null;
+		}
 		int len = Math.min(randomBytes.length, csrfBytes.length);
 		byte[] xoredCsrf = new byte[len];
-		System.arraycopy(csrfBytes, 0, xoredCsrf, 0, csrfBytes.length);
+		System.arraycopy(csrfBytes, 0, xoredCsrf, 0, len);
 		for (int i = 0; i < len; i++) {
 			xoredCsrf[i] ^= randomBytes[i];
 		}
--- a/messaging/src/test/java/org/springframework/security/messaging/web/csrf/XorCsrfChannelInterceptorTests.java
+++ b/messaging/src/test/java/org/springframework/security/messaging/web/csrf/XorCsrfChannelInterceptorTests.java
@@ -32,6 +32,7 @@ import org.springframework.security.web.csrf.InvalidCsrfTokenException;
 import org.springframework.security.web.csrf.MissingCsrfTokenException;
 
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+import static org.assertj.core.api.Assertions.assertThatNoException;
 import static org.mockito.Mockito.mock;
 
 /**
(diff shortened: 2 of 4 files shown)


messaging/src/main/java/org/springframework/security/messaging/web/csrf/XorCsrfTokenUtils.java
```diff
@@ -56,13 +56,16 @@ static String getTokenValue(String actualToken, String token) {
...
-		return Utf8.decode(csrfBytes);
+		return (csrfBytes != null) ? Utf8.decode(csrfBytes) : null;
 	}
 
-	private static byte[] xorCsrf(byte[] randomBytes, byte[] csrfBytes) {
+	static byte[] xorCsrf(byte[] randomBytes, byte[] csrfBytes) {
+		if (csrfBytes.length < randomBytes.length) {
+			return null;
+		}
 		int len = Math.min(randomBytes.length, csrfBytes.length);
```
roysav13, line 66: Hi,
Isn't the value of `randomBytes.length` will always be smaller or equals to the value of `csrfBytes.length`?
If so you can remove the usage of the min fucntion and just call the value

messaging/src/main/java/org/springframework/security/messaging/web/csrf/XorCsrfTokenUtils.java
```diff
@@ -56,13 +56,16 @@ static String getTokenValue(String actualToken, String token) {
 		System.arraycopy(actualBytes, randomBytesSize, xoredCsrf, 0, tokenSize);
 
 		byte[] csrfBytes = xorCsrf(randomBytes, xoredCsrf);
-		return Utf8.decode(csrfBytes);
+		return (csrfBytes != null) ? Utf8.decode(csrfBytes) : null;
 	}
 
-	private static byte[] xorCsrf(byte[] randomBytes, byte[] csrfBytes) {
+	static byte[] xorCsrf(byte[] randomBytes, byte[] csrfBytes) {
+		if (csrfBytes.length < randomBytes.length) {
```
roysav13, line 63: don't you want to assert that these values equals?
if the `randomBytes` will be larger there is no downside but if the `csrfBytes` is longer then there might be a condition where some bytes of the csrf token is not xored

messaging/src/test/java/org/springframework/security/messaging/web/csrf/XorCsrfChannelInterceptorTests.java
```diff
@@ -141,6 +142,13 @@ public void preSendWhenUnsubscribeThenIgnores() {
 		this.interceptor.preSend(message(), this.channel);
 	}
 
+	@Test
+	public void preSendWhenCsrfBytesIsLongerThanRandomBytesThenArrayIndexOutOfBoundsExceptionWillNotBeThrown() {
+		this.messageHeaders.setNativeHeader(this.token.getHeaderName(), XOR_CSRF_TOKEN_VALUE);
+		this.messageHeaders.getSessionAttributes().put(CsrfToken.class.getName(), this.token);
```
marcusdacoregio, line 148: Looks like this test is not passing a longer CSRF token than the random bytes, can you verify it please?

kratosmy: Thanks for reviewing. I have made the change.

sjohnr:
@kratosmy thanks for submitting this PR! I appreciate your work in helping resolve this issue. However, after taking a deeper look at the original issue and fix as well as this additional fix, the resulting behavior isn't consistent and the fix can be improved. Since this is a fairly subtle and complex issue, I have decided to fix the issue myself in a separate commit and backport the fix to older OSS supported branches.

Since I won't be directly re-using either the fix or the tests you provided, I am unfortunately going to close this PR as declined. However, I again want to reiterate that your work is appreciated and was instrumental in helping resolve this issue. Thanks for your contribution!

messaging/src/main/java/org/springframework/security/messaging/web/csrf/XorCsrfTokenUtils.java
sjohnr: Thanks for your comments on this PR @roysav13! I will be addressing this comment in a separate fix. See gh-15184.

messaging/src/main/java/org/springframework/security/messaging/web/csrf/XorCsrfTokenUtils.java
sjohnr: Thanks for your comments on this PR @roysav13! I will be addressing this comment in a separate fix. See gh-15184.

PR #14977, opened 2024-04-27
Assert WebSession is not null

### LINKED ISSUE #14975: WebSessionServerOAuth2AuthorizedClientRepository throws NPE when session is null
**Describe the bug**
[This line](https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/WebSessionServerOAuth2AuthorizedClientRepository.java#L70) `session.getAttributes() `throws NPE, if we have WebSession bean configuration like this to disable WebSession:

``` 
@Bean
    fun webSessionManager(): WebSessionManager {
        // Emulate SessionCreationPolicy.STATELESS
        return WebSessionManager { exchange: ServerWebExchange? -> Mono.empty() }
    }
 ``` 

**Expected behavior**
I thing we may do null-safety when we are trying to call getAuthorizedClients in [this place](https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/WebSessionServerOAuth2AuthorizedClientRepository.java#L99) with nullable session.

**Configuration**
Spring Cloud Gateway.

```
@Bean
    fun authorizedClientManager(
        clientRegistrationRepository: ReactiveClientRegistrationRepository,
        authorizedClientRepository: ServerOAuth2AuthorizedClientRepository
    ): ReactiveOAuth2AuthorizedClientManager {
        val authorizedClientProvider = ReactiveOAuth2AuthorizedClientProviderBuilder
            .builder()
            .clientCredentials()
            .build()
        val authorizedClientManager = DefaultReactiveOAuth2AuthorizedClientManager(
            clientRegistrationRepository, authorizedClientRepository
        )
        authorizedClientManager.setAuthorizedClientProvider(authorizedClientProvider)
        return authorizedClientManager
    }

    @Bean
    fun springSecurityFilterChainNonProd(
        httpSecurity: ServerHttpSecurity,
        someIntrospector: SomeIntrospector,
    ): SecurityWebFilterChain? {
        return httpSecurity
            .cors(Customizer.withDefaults())
            .authorizeExchange { auth ->
                auth.anyExchange().authenticated()
            }
            .oauth2ResourceServer { oauth2 ->
                oauth2.opaqueToken {
                    it.introspector(someIntrospector)
                }
            }
            .csrf { csrf -> csrf.disable() }
            .logout { logout -> logout.disable() }
            .build()
    }
```

### PR DESCRIPTION
**Pull request by dukbong**
### Original Code:
The getAuthorizedClients method retrieves a map of OAuth2 authorized clients from a WebSession. In the original code, if the session is null, the method assigns a default empty HashMap. This code does not explicitly assert that WebSession cannot be null, potentially leading to silent failures or unexpected behavior.

Closes #14975

### Suggested Refactoring:
To ensure robustness and improve error handling, you can refactor the method by adding an assertion to prevent a null session. This approach uses Assert.notNull to raise an exception with a clear error message if the WebSession is null, helping developers quickly identify and fix the problem.

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/WebSessionServerOAuth2AuthorizedClientRepository.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/WebSessionServerOAuth2AuthorizedClientRepository.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -96,12 +96,9 @@ public final class WebSessionServerOAuth2AuthorizedClientRepository implements S
 
 	@SuppressWarnings("unchecked")
 	private Map<String, OAuth2AuthorizedClient> getAuthorizedClients(WebSession session) {
-		Map<String, OAuth2AuthorizedClient> authorizedClients = (session != null)
-				? (Map<String, OAuth2AuthorizedClient>) session.getAttribute(this.sessionAttributeName) : null;
-		if (authorizedClients == null) {
-			authorizedClients = new HashMap<>();
-		}
-		return authorizedClients;
+		Assert.notNull(session, "session cannot be null");
+		Map<String, OAuth2AuthorizedClient> authorizedClients = session.getAttribute(this.sessionAttributeName);
+		return (authorizedClients != null) ? authorizedClients : new HashMap<>();
 	}
 
 }
--- a/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/server/WebSessionServerOAuth2AuthorizedClientRepositoryTests.java
+++ b/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/server/WebSessionServerOAuth2AuthorizedClientRepositoryTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,10 +25,12 @@ import org.springframework.security.oauth2.client.registration.ClientRegistratio
 import org.springframework.security.oauth2.client.registration.TestClientRegistrations;
 import org.springframework.security.oauth2.core.OAuth2AccessToken;
 import org.springframework.web.server.WebSession;
+import reactor.core.publisher.Mono;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatIllegalArgumentException;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
 /**
  * @author Rob Winch
@@ -200,6 +202,26 @@ public class WebSessionServerOAuth2AuthorizedClientRepositoryTests {
 			.block();
 		assertThat(loadedAuthorizedClient2).isNotNull();
 		assertThat(loadedAuthorizedClient2).isSameAs(authorizedClient2);
+	}
+	
+	@Test
+	public void saveAuthorizedClientWhenSessionIsNullThenThrowIllegalArgumentException() {
+		MockServerWebExchange mockedExchange = mock(MockServerWebExchange.class);
+		when(mockedExchange.getSession()).thenReturn(Mono.empty());
+		OAuth2AuthorizedClient authorizedClient = new OAuth2AuthorizedClient(this.registration1, this.principalName1,
+				mock(OAuth2AccessToken.class));
+		assertThatIllegalArgumentException().isThrownBy(
+				() -> authorizedClientRepository.saveAuthorizedClient(authorizedClient, null, mockedExchange).block())
+				.withMessage("session cannot be null");
+	}
+	
+	@Test
+	public void removeAuthorizedClientWhenSessionIsNullThenThrowIllegalArgumentException() {
+		MockServerWebExchange mockedExchange = mock(MockServerWebExchange.class);
+		when(mockedExchange.getSession()).thenReturn(Mono.empty());
+		assertThatIllegalArgumentException().isThrownBy(
+				() -> authorizedClientRepository.removeAuthorizedClient(this.registrationId1, null, mockedExchange).block())
+				.withMessage("session cannot be null");
 	}
 
 }

### CODE REVIEW

**Comment by dukbong** [comment]
> Hi @dukbong, thanks for the PR!
> 
> 
> 
> Just a heads up, I closed the original issue with [this explanation](https://github.com/spring-projects/spring-security/issues/14975#issuecomment-2090702647). I appreciate what you're doing in this change by asserting the session cannot be null. I'm wondering though if this change is absolutely necessary?
> 
> 
> 
> > This code does not explicitly assert that WebSession cannot be null, potentially leading to silent failures or unexpected behavior.
> 
> 
> 
> Can you explain the situation in which this would occur (other than the invalid configuration referred to in the original issue)?


Hi @sjohnr 

Initially, I planned to submit a pull request to address the issue, but as I reviewed the code, I found the structure to be questionable, prompting a change in direction.

The `getAuthorizedClient` function is written with the assumption that the session might be null. However, if the session is null, this leads to an exception in the `saveAuthorizedClient` function. To prevent this, I believe it's better to ensure the session isn't null either in the `getAuthorizedClient` function or in the `doOnSuccess` block.

Based on these observations, I created a pull request to address the issue.

- Your feedback is invaluable for my growth and learning.

File: oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/WebSessionServerOAuth2AuthorizedClientRepository.java [hunk 1]
```diff
@@ -94,14 +94,11 @@ public Mono<Void> removeAuthorizedClient(String clientRegistrationId, Authentica
...
-		if (authorizedClients == null) {
-			authorizedClients = new HashMap<>();
-		}
-		return authorizedClients;
+		Assert.notNull(session, "WebSession cannot be null. Ensure the session is properly initialized before accessing authorized clients.");
+	    Map<String, OAuth2AuthorizedClient> authorizedClients = session.getAttribute(this.sessionAttributeName);
+	    return authorizedClients != null ? authorizedClients : new HashMap<>();
 	}
 
+
```
sjohnr, line 103: Please remove this extra newline.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/WebSessionServerOAuth2AuthorizedClientRepository.java
sjohnr, line 1: Please update the copyright year in the header as per [contributing guidelines](https://github.com/spring-projects/spring-security/blob/main/CONTRIBUTING.adoc#submit-a-pull-request).

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/WebSessionServerOAuth2AuthorizedClientRepository.java
```diff
@@ -94,14 +94,11 @@ public Mono<Void> removeAuthorizedClient(String clientRegistrationId, Authentica
...
 	private Map<String, OAuth2AuthorizedClient> getAuthorizedClients(WebSession session) {
-		Map<String, OAuth2AuthorizedClient> authorizedClients = (session != null)
-				? (Map<String, OAuth2AuthorizedClient>) session.getAttribute(this.sessionAttributeName) : null;
-		if (authorizedClients == null) {
-			authorizedClients = new HashMap<>();
-		}
-		return authorizedClients;
+		Assert.notNull(session, "WebSession cannot be null. Ensure the session is properly initialized before accessing authorized clients.");
+	    Map<String, OAuth2AuthorizedClient> authorizedClients = session.getAttribute(this.sessionAttributeName);
+	    return authorizedClients != null ? authorizedClients : new HashMap<>();
```
sjohnr, line 100: You can add parentheses around the ternary condition to satisfy checkstyle and allow the build to pass.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/WebSessionServerOAuth2AuthorizedClientRepository.java
```diff
@@ -94,14 +94,11 @@ public Mono<Void> removeAuthorizedClient(String clientRegistrationId, Authentica
...
 
-	@SuppressWarnings("unchecked")
 	private Map<String, OAuth2AuthorizedClient> getAuthorizedClients(WebSession session) {
-		Map<String, OAuth2AuthorizedClient> authorizedClients = (session != null)
-				? (Map<String, OAuth2AuthorizedClient>) session.getAttribute(this.sessionAttributeName) : null;
-		if (authorizedClients == null) {
-			authorizedClients = new HashMap<>();
-		}
-		return authorizedClients;
+		Assert.notNull(session, "WebSession cannot be null. Ensure the session is properly initialized before accessing authorized clients.");
```
sjohnr, line 98: I think the message can simply be `session cannot be null`.

sjohnr (changes requested):
Thanks for the PR @dukbong. In addition to the feedback inline below, can you please add a test that asserts an `IllegalArgumentException` is thrown when `WebSession` is null?

dukbong:
> Thanks for the PR @dukbong. In addition to the feedback inline below, can you please add a test that asserts an `IllegalArgumentException` is thrown when `WebSession` is null?

Thanks @sjohnr  for agreeing with my feedback.
I will make the changes and additions as per the comments below!

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/WebSessionServerOAuth2AuthorizedClientRepository.java
```diff
@@ -94,14 +94,10 @@ public Mono<Void> removeAuthorizedClient(String clientRegistrationId, Authentica
 		// @formatter:on
 	}
 
-	@SuppressWarnings("unchecked")
```
sjohnr, line 97: I don't think this should be removed. Can you restore this line?

sjohnr (commented):
Thanks @dukbong. I added one minor comment below. Can you also please rebase your changes on main and squash commits? I will schedule this to be merged in the next minor so you won't see it merged for a few weeks.

dukbong:
> Thanks @dukbong. I added one minor comment below. Can you also please rebase your changes on main and squash commits? I will schedule this to be merged in the next minor so you won't see it merged for a few weeks.


Hello. @sjohnr 

As you suggested, I rebased onto the main branch and squashed the commits into one.

If there's anything I did wrong or need to correct, please let me know!
I'll fix it right away!!

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/server/WebSessionServerOAuth2AuthorizedClientRepository.java
sjohnr: Apologies as you were correct to remove this. I'll re-remove it in a polish commit.

sjohnr:
Thanks @dukbong, this is now merged into `5.8.x` as 1695d03b72d9d48f24abd8381f5f120b99a7f7a1. I did apply a polish commit with some adjustments because I didn't run the build to see checkstyle and other build output, so check out 
e34621ec2ced820accc121b053752387d053d9a5 and let me know if you have questions.

PR #15012, opened 2024-05-06
Expose user name attribute name in `OAuth2UserAuthority`

filiphr:
We are using custom `GrantedAuthoritiesMapper` that is doing some custom loading of attributes based on the `OAuth2UserAuthority`. We've had some request to do something based on the user name. Currently, there is no easy way to access the configured username attribute name.

The goal of this PR is to expose this information directly in the `OAuth2UserAuthority` and this make it easy to get the username from the attributes.

An alternative could also be to expose `getUserName` and just use the `userNameAttributeName` in the `OAuth2UserAuthority`.

I hope that you'll consider this PR, if there is anything that you think should be added or done better please let me know


filiphr:
Can someone from the Spring Security team have a look at this? If this is something that you are not interested in then let me know and we can close it.

sjohnr:
Thanks @filiphr! I'll be taking a look at this and send any feedback in the next few days. Overall it seems reasonable and a nice enhancement so I think we should be able to accept it.

sjohnr:
Thanks @filiphr! This is now merged as 99aee99b3457a1ecebc05d543fb417a0b0008d11 with minor polish 7c45ebd81c8b9ca7569e65137aae84de9c48af76.

filiphr:
Thanks a lot for merging this @sjohnr

PR #15033, opened 2024-05-09
instanceof changed to recent syntax

dukbong:
instanceof changed to recent syntax

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -42,7 +42,7 @@ springRelease {
 	weekOfMonth = 3
 	dayOfWeek = 1
 	referenceDocUrl = "https://docs.spring.io/spring-security/reference/{version}/index.html"
-	apiDocUrl = "https://docs.spring.io/spring-security/docs/{version}/api/"
+	apiDocUrl = "https://docs.spring.io/spring-security/site/docs/{version}/api/"
 	replaceSnapshotVersionInReferenceDocUrl = true
 }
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/springframework/security/authentication/AbstractAuthenticationToken.java
```diff
@@ -112,8 +112,8 @@ public void eraseCredentials() {
 	}
 
 	private void eraseSecret(Object secret) {
-		if (secret instanceof CredentialsContainer) {
-			((CredentialsContainer) secret).eraseCredentials();
+		if (secret instanceof CredentialsContainer secret) {
```
marcusdacoregio, line 115: The variable name `secret` is already defined in the scope, can you please rename it to maybe `container`?

marcusdacoregio (changes requested):
Thanks @dukbong, I've left a feedback inline.

marcusdacoregio:
In preparation to merge, can you please squash your commits?

The single commit message could be: `Use instanceof Pattern Matching `

dukbong:
> In preparation to merge, can you please squash your commits?
> 
> The single commit message could be: `Use instanceof Pattern Matching `

I'm sorry. @marcusdacoregio 

My mistake caused the PR to be removed.

I will submit a new PR.

PR #15048, opened 2024-05-12
Add methods to augment allowed headers and parameters in StrictHttpFi…

### LINKED ISSUE #13639: `StrictHttpFirewall#setAllowedHeaderNames` should augment with existing `Predicate`
**Expected Behavior**
`StrictHttpFirewall#setAllowedHeaderNames` either should augment the existing predicate with `Predicate#and` or provide `addAllowHeaderNames` to allow augmenting it to current predicate, so that I get to retain default protection provided by `StrictHttpFirewall`.

**Current Behavior**
The current implementation rejects any Request with non-ascii characters. Current implementation of `StrictHttpFirewall#setAllowedHeaderNames` replaces predicate, making me lose out to default protection. 

I am willing to provide a PR for this should this be accepted as valid request.

baezzys:
This pull request introduces new constants in the `StrictHttpFirewall` class that allow for the augmentation of the sets of allowable header names, header values, parameter names, and parameter values.

This closes #13639

### DIFF
--- a/web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
+++ b/web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
@@ -75,6 +75,7 @@ import org.springframework.util.Assert;
  *
  * @author Rob Winch
  * @author Eddú Meléndez
+ * @author Jinwoo Bae
  * @since 4.2.4
  * @see DefaultHttpFirewall
  */
@@ -134,13 +135,21 @@ public class StrictHttpFirewall implements HttpFirewall {
 
 	private static final Predicate<String> HEADER_VALUE_PREDICATE = (s) -> HEADER_VALUE_PATTERN.matcher(s).matches();
 
-	private Predicate<String> allowedHeaderNames = ASSIGNED_AND_NOT_ISO_CONTROL_PREDICATE;
+	private Predicate<String> allowedHeaderNames = ALLOWED_HEADER_NAMES;
 
-	private Predicate<String> allowedHeaderValues = HEADER_VALUE_PREDICATE;
+	public static final Predicate<String> ALLOWED_HEADER_NAMES = ASSIGNED_AND_NOT_ISO_CONTROL_PREDICATE;
 
-	private Predicate<String> allowedParameterNames = ASSIGNED_AND_NOT_ISO_CONTROL_PREDICATE;
+	private Predicate<String> allowedHeaderValues = ALLOWED_HEADER_VALUES;
 
-	private Predicate<String> allowedParameterValues = (value) -> true;
+	public static final Predicate<String> ALLOWED_HEADER_VALUES = HEADER_VALUE_PREDICATE;
+
+	private Predicate<String> allowedParameterNames = ALLOWED_PARAMETER_NAMES;
+
+	public static final Predicate<String> ALLOWED_PARAMETER_NAMES = ASSIGNED_AND_NOT_ISO_CONTROL_PREDICATE;
+
+	private Predicate<String> allowedParameterValues = ALLOWED_PARAMETER_VALUES;
+
+	public static final Predicate<String> ALLOWED_PARAMETER_VALUES = (value) -> true;
 
 	public StrictHttpFirewall() {
 		urlBlocklistsAddAll(FORBIDDEN_SEMICOLON);
--- a/web/src/test/java/org/springframework/security/web/firewall/StrictHttpFirewallTests.java
+++ b/web/src/test/java/org/springframework/security/web/firewall/StrictHttpFirewallTests.java
@@ -31,6 +31,7 @@ import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 /**
  * @author Rob Winch
  * @author Eddú Meléndez
+ * @author Jinwoo Bae
  */
 public class StrictHttpFirewallTests {
 
@@ -724,9 +725,26 @@ public class StrictHttpFirewallTests {
 	}
 
 	@Test
+	public void getFirewalledRequestWhenHeaderNameNotAllowedWithAugmentedHeaderNamesThenException() {
+		this.firewall
+			.setAllowedHeaderNames(StrictHttpFirewall.ALLOWED_HEADER_NAMES.and((name) -> !name.equals("bad name")));
+		HttpServletRequest request = this.firewall.getFirewalledRequest(this.request);
+		assertThatExceptionOfType(RequestRejectedException.class).isThrownBy(() -> request.getHeader("bad name"));
+	}
+
+	@Test
 	public void getFirewalledRequestGetHeaderWhenNotAllowedHeaderValueThenException() {
 		this.request.addHeader("good name", "bad value");
 		this.firewall.setAllowedHeaderValues((value) -> !value.equals("bad value"));
+		HttpServletRequest request = this.firewall.getFirewalledRequest(this.request);
+		assertThatExceptionOfType(RequestRejectedException.class).isThrownBy(() -> request.getHeader("good name"));
+	}
+
+	@Test
+	public void getFirewalledRequestWhenHeaderValueNotAllowedWithAugmentedHeaderValuesThenException() {
+		this.request.addHeader("good name", "bad value");
+		this.firewall.setAllowedHeaderValues(
+				StrictHttpFirewall.ALLOWED_HEADER_VALUES.and((value) -> !value.equals("bad value")));
 		HttpServletRequest request = this.firewall.getFirewalledRequest(this.request);
 		assertThatExceptionOfType(RequestRejectedException.class).isThrownBy(() -> request.getHeader("good name"));
 	}
@@ -841,9 +859,29 @@ public class StrictHttpFirewallTests {
 	}
 
 	@Test
+	public void getFirewalledRequestWhenParameterValueNotAllowedWithAugmentedParameterValuesThenException() {
+		this.request.addParameter("Something", "bad value");
+		this.firewall.setAllowedParameterValues(
+				StrictHttpFirewall.ALLOWED_PARAMETER_VALUES.and((value) -> !value.equals("bad value")));
+		HttpServletRequest request = this.firewall.getFirewalledRequest(this.request);
+		assertThatExceptionOfType(RequestRejectedException.class)
+			.isThrownBy(() -> request.getParameterValues("Something"));
+	}
+
+	@Test
 	public void getFirewalledRequestGetParameterValuesWhenNotAllowedInParameterNameThenException() {
 		this.firewall.setAllowedParameterNames((value) -> !value.equals("bad name"));
 		this.request.addParameter("bad name", "good value");
+		HttpServletRequest request = this.firewall.getFirewalledRequest(this.request);
+		assertThatExceptionOfType(RequestRejectedException.class)
+			.isThrownBy(() -> request.getParameterValues("bad name"));
+	}
+
+	@Test
+	public void getFirewalledRequestWhenParameterNameNotAllowedWithAugmentedParameterNamesThenException() {
+		this.request.addParameter("bad name", "good value");
+		this.firewall.setAllowedParameterNames(
+				StrictHttpFirewall.ALLOWED_PARAMETER_NAMES.and((value) -> !value.equals("bad name")));
 		HttpServletRequest request = this.firewall.getFirewalledRequest(this.request);
 		assertThatExceptionOfType(RequestRejectedException.class)
 			.isThrownBy(() -> request.getParameterValues("bad name"));


jzheaux (changes requested):
Thanks, @baezzys! There is an issue with having both `set` and `add` that I didn't see earlier, which is that it assumes that the application wants to do `Predicate#and` and not `Predicate#or`. It doesn't give the developer much more power.

Also, I think that `addAllowHeaderValues` makes it seem like you are listing additional ways that headers would be allowed, which is the opposite of the implementation.

Instead, please let's introduce `public static` defaults for each, like so:

```java
public static final Predicate<String> ALLOWED_HEADER_NAMES = ...;

public static final Predicate<String> ALLOWED_HEADER_VALUES = ...;

public static final Predicate<String> ALLOWED_PARAMETER_NAMES = ...;

public static final Predicate<String> ALLOWED_PARAMETER_VALUES = ...;
```

So that an application can do:

```java
firewall.setAllowedHeaderValues(ALLOWED_HEADER_VALUES.and((value) -> !value.contains("\t")))
```

baezzys:
Thank you for the review, @jzheaux. However, since the setters are not static, would it be acceptable to change them to public instead of public static?

jzheaux:
I think they should be `static` so that you don't need an instance of the firewall to refer to them.

baezzys:
Thank you for the feedback, @jzheaux

I have updated the `StrictHttpFirewall` class to change the private Predicates for allowed header names, header values, parameter names, and parameter values to public static. Additionally, I have changed the corresponding setters.

Please review the changes and let me know if any further adjustments are needed.

web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
```diff
@@ -134,13 +135,13 @@ public class StrictHttpFirewall implements HttpFirewall {
 
 	private static final Predicate<String> HEADER_VALUE_PREDICATE = (s) -> HEADER_VALUE_PATTERN.matcher(s).matches();
 
-	private Predicate<String> allowedHeaderNames = ASSIGNED_AND_NOT_ISO_CONTROL_PREDICATE;
```
jzheaux, line 137: Let's keep both. There is the public constant that folks can use to build expressions and then there is whatever in the internal setting. You'd want something like this:

```java
private Predicate<String> allowedHeaderNames = ALLOWED_HEADER_NAMES;

public static final Predicate<String> ALLOWED_HEADER_NAMES = ASSIGNED_AND_NOT_ISO_CONTROL_PREDICATE;
```

web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
```diff
@@ -435,9 +436,9 @@ public void setAllowUrlEncodedLineSeparator(boolean allowUrlEncodedLineSeparator
 	 * @see Character#isISOControl(int)
 	 * @see Character#isDefined(int)
 	 */
-	public void setAllowedHeaderNames(Predicate<String> allowedHeaderNames) {
+	public static void setAllowedHeaderNames(Predicate<String> allowedHeaderNames) {
```
jzheaux, line 439: Please keep the methods non-static. We only need the public constants to be static.

web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
```diff
@@ -435,9 +436,9 @@ public void setAllowUrlEncodedLineSeparator(boolean allowUrlEncodedLineSeparator
...
 		Assert.notNull(allowedHeaderNames, "allowedHeaderNames cannot be null");
-		this.allowedHeaderNames = allowedHeaderNames;
```
jzheaux, line 440: Please keep this line as it was.

web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
```diff
@@ -450,9 +451,9 @@ public void setAllowedHeaderNames(Predicate<String> allowedHeaderNames) {
 	 * @see Character#isISOControl(int)
 	 * @see Character#isDefined(int)
 	 */
-	public void setAllowedHeaderValues(Predicate<String> allowedHeaderValues) {
+	public static void setAllowedHeaderValues(Predicate<String> allowedHeaderValues) {
 		Assert.notNull(allowedHeaderValues, "allowedHeaderValues cannot be null");
-		this.allowedHeaderValues = allowedHeaderValues;
+		ALLOWED_HEADER_VALUES = allowedHeaderValues;
```
jzheaux, line 456: Please keep this line as it was.

web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
```diff
@@ -463,9 +464,9 @@ public void setAllowedHeaderValues(Predicate<String> allowedHeaderValues) {
 	 * @see Character#isISOControl(int)
 	 * @see Character#isDefined(int)
 	 */
-	public void setAllowedParameterNames(Predicate<String> allowedParameterNames) {
+	public static void setAllowedParameterNames(Predicate<String> allowedParameterNames) {
 		Assert.notNull(allowedParameterNames, "allowedParameterNames cannot be null");
-		this.allowedParameterNames = allowedParameterNames;
```
jzheaux, line 468: Please keep this line as it was

web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
```diff
@@ -476,9 +477,9 @@ public void setAllowedParameterNames(Predicate<String> allowedParameterNames) {
 	 * @param allowedParameterValues the predicate for testing parameter values
 	 * @since 5.4
 	 */
-	public void setAllowedParameterValues(Predicate<String> allowedParameterValues) {
+	public static void setAllowedParameterValues(Predicate<String> allowedParameterValues) {
 		Assert.notNull(allowedParameterValues, "allowedParameterValues cannot be null");
-		this.allowedParameterValues = allowedParameterValues;
```
jzheaux, line 481: Please keep this line as it was

web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
```diff
@@ -826,28 +827,28 @@ public String[] getParameterValues(String name) {
 		}
 
 		private void validateAllowedHeaderName(String headerNames) {
-			if (!StrictHttpFirewall.this.allowedHeaderNames.test(headerNames)) {
+			if (!StrictHttpFirewall.this.ALLOWED_HEADER_NAMES.test(headerNames)) {
```
jzheaux, line 830: Please leave this line as it was.

web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
```diff
@@ -826,28 +827,28 @@ public String[] getParameterValues(String name) {
...
 				throw new RequestRejectedException(
 						"The request was rejected because the header name \"" + headerNames + "\" is not allowed.");
 			}
 		}
 
 		private void validateAllowedHeaderValue(String name, String value) {
-			if (!StrictHttpFirewall.this.allowedHeaderValues.test(value)) {
+			if (!StrictHttpFirewall.this.ALLOWED_HEADER_VALUES.test(value)) {
```
jzheaux, line 837: Please leave this line as it was.

web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
```diff
@@ -826,28 +827,28 @@ public String[] getParameterValues(String name) {
...
 				throw new RequestRejectedException("The request was rejected because the header: \"" + name
 						+ " \" has a value \"" + value + "\" that is not allowed.");
 			}
 		}
 
 		private void validateAllowedParameterName(String name) {
-			if (!StrictHttpFirewall.this.allowedParameterNames.test(name)) {
+			if (!StrictHttpFirewall.this.ALLOWED_PARAMETER_NAMES.test(name)) {
```
jzheaux, line 844: Please leave this line as it was.

web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
```diff
@@ -826,28 +827,28 @@ public String[] getParameterValues(String name) {
...
 				throw new RequestRejectedException(
 						"The request was rejected because the parameter name \"" + name + "\" is not allowed.");
 			}
 		}
 
 		private void validateAllowedParameterValue(String name, String value) {
-			if (!StrictHttpFirewall.this.allowedParameterValues.test(value)) {
+			if (!StrictHttpFirewall.this.ALLOWED_PARAMETER_VALUES.test(value)) {
```
jzheaux, line 851: Please leave this line as it was.

web/src/test/java/org/springframework/security/web/firewall/StrictHttpFirewallTests.java
```diff
@@ -41,6 +44,19 @@ public class StrictHttpFirewallTests {
...
 	private MockHttpServletRequest request = new MockHttpServletRequest("GET", "");
 
+	private static final Pattern ASSIGNED_AND_NOT_ISO_CONTROL_PATTERN = Pattern
+		.compile("[\\p{IsAssigned}&&[^\\p{IsControl}]]*");
+
+	private static final Pattern HEADER_VALUE_PATTERN = Pattern.compile("[\\p{IsAssigned}&&[[^\\p{IsControl}]||\\t]]*");
+
+	@BeforeEach
+	public void initializeFirewall() {
+		StrictHttpFirewall.ALLOWED_HEADER_NAMES = (s) -> ASSIGNED_AND_NOT_ISO_CONTROL_PATTERN.matcher(s).matches();
```
jzheaux, line 54: I think these will already be set by the object itself. They won't need to be reset if you follow the earlier suggestions.

web/src/test/java/org/springframework/security/web/firewall/StrictHttpFirewallTests.java
```diff
@@ -723,6 +739,18 @@ public void getFirewalledRequestGetHeaderWhenNotAllowedHeaderNameThenException()
 		assertThatExceptionOfType(RequestRejectedException.class).isThrownBy(() -> request.getHeader("bad name"));
 	}
 
+	@Test
+	public void getFirewalledRequestGetHeaderWhenSetNotAllowedHeaderNamesThenException() {
+		this.firewall.setAllowedHeaderNames((name) -> !name.equals("bad name"));
+		this.firewall
+			.setAllowedHeaderNames(this.firewall.ALLOWED_HEADER_NAMES.and((name) -> !name.equals("worst name")));
```
jzheaux, line 746: After the above changes, this will change to something closer to:

```java
this.firewall.setAllowedHeaderNames(
    StrictHttpFirewall.ALLOWED_HEADER_NAMES.and((name) -> !name.equals("worst name"));
```

web/src/test/java/org/springframework/security/web/firewall/StrictHttpFirewallTests.java
```diff
@@ -731,6 +759,21 @@ public void getFirewalledRequestGetHeaderWhenNotAllowedHeaderValueThenException(
 		assertThatExceptionOfType(RequestRejectedException.class).isThrownBy(() -> request.getHeader("good name"));
 	}
 
+	@Test
+	public void getFirewalledRequestGetHeaderWhenSetNotAllowedHeaderValuesThenException() {
+		this.request.addHeader("good name", "bad value");
+		this.request.addHeader("best name", "worst value");
+
+		this.firewall.setAllowedHeaderValues((value) -> !value.equals("bad value"));
```
jzheaux, line 767: Since the second call overrides the first, you can leave this line out and it won't change the result of the test.

jzheaux (changes requested):
Thanks, @baezzys! I've left some feedback inline.

web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
baezzys: Thank you for your feedback, @jzheaux.

I have updated the code to reflect your suggestions. 

One question I have is regarding the potential confusion that might arise because ALLOWED_HEADER_NAMES is final and thus immutable. Even if the user changes the allowedHeaderNames predicate, the ALLOWED_HEADER_NAMES constant remains unchanged. Could this lead to any misunderstandings or misuses?

web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
baezzys: Done.

web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
baezzys: Done.

web/src/test/java/org/springframework/security/web/firewall/StrictHttpFirewallTests.java
baezzys: Done.

web/src/test/java/org/springframework/security/web/firewall/StrictHttpFirewallTests.java
baezzys: Done.

web/src/test/java/org/springframework/security/web/firewall/StrictHttpFirewallTests.java
baezzys: Thank you for the review.

I have updated the code to retain only the first call.

baezzys:
Hi @jzheaux, Sorry for the late response. I've made the changes based on your feedback. PTAL. Thanks!

web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
baezzys: Done.

web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
baezzys: Done.

web/src/main/java/org/springframework/security/web/firewall/StrictHttpFirewall.java
jzheaux: This is a common standard in Java, so I don't think it will lead to wide confusion. The constant needs to be final because that is not the thing that is being changed. Instead, the setter is updating the value for `allowedHeaderNames`.

jzheaux (approved):

jzheaux:
Thank you, @baezzys, for the PR! This is now merged into `main`.

PR #15066, opened 2024-05-13
Do Not Invalidate Current Session When Its Registered

joaquinjsb:
this change prevents from deleting the current session, which ends up creating the need to log in two times.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/server/authentication/InvalidateLeastUsedServerMaximumSessionsExceededHandler.java
+++ b/web/src/main/java/org/springframework/security/web/server/authentication/InvalidateLeastUsedServerMaximumSessionsExceededHandler.java
@@ -54,6 +54,7 @@ public final class InvalidateLeastUsedServerMaximumSessionsExceededHandler
 				maximumSessionsExceededBy);
 
 		return Flux.fromIterable(leastRecentlyUsedSessionsToInvalidate)
+			.filter(toInvalidate -> !toInvalidate.getSessionId().equals(context.getCurrentSession().getId()))
 			.flatMap((toInvalidate) -> toInvalidate.invalidate().thenReturn(toInvalidate))
 			.flatMap((toInvalidate) -> this.webSessionStore.removeSession(toInvalidate.getSessionId()))
 			.then();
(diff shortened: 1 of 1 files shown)


marcusdacoregio:
Hi @joaquinjsb, can you elaborate more on what is the problem and how this fixes it? As far as I can tell the invalidation is performed before the current session is registered in the `ReactiveSessionRegistry`, thus not returning the current session.

joaquinjsb:
> Hi @joaquinjsb, can you elaborate more on what is the problem and how this fixes it? As far as I can tell the invalidation is performed before the current session is registered in the `ReactiveSessionRegistry`, thus not returning the current session.

what happens in my case is ->
-> login browser 1
-> login browser 2
    -> the ConcurrentSessionControlServerAuthenticationSuccessHandler:handleConcurrency finds currentSession, and the  
         registeredSessions = sessionTuple.getT2(); contains both the current session and the session from browser 1 and passes it down on line 87.

what my fix does is, removes from the registeredSessions, the current session, so that you don't end up deleting both browser 1 and browser 2 sessions, which prevents the login flow from being "standard" on the user side.

marcusdacoregio:
There is a test that verifies that behavior [here](https://github.com/spring-projects/spring-security/blob/main/config/src/test/java/org/springframework/security/config/web/server/SessionManagementSpecTests.java#L86-L118), can you create a new test that reproduces the problem?

joaquinjsb:
I tried running it with a limit of 1, which behaves correctly as you describe, would the problem be probably on SpringSessionBackedReactiveSessionRegistry? do you have any suggestions on how to proceed?

joaquinjsb:
> The best approach would be to create a minimal, reproducible sample for the issues, with tests preferably. Then we can identify exactly where the problem is

here you have both minimal reproducible, and a test which is failing.
https://github.com/joaquinjsb/sessionRedisReactive

joaquinjsb:
I dived in further, the answer to our question is here:
https://github.com/spring-projects/spring-security/blob/a17a599bd54838dd2f166d97555de3240335d07f/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java#L2120C5-L2127C6

the delegate.getSession() creates a new session, then it gets saved in the repository, thus creating two entries at this point

marcusdacoregio:
Thanks @joaquinjsb, this has been closed via https://github.com/spring-projects/spring-security/commit/927840fe88052a40c72fe80a16372f594880e02b

PR #15070, opened 2024-05-14
Use instanceof Pattern Matching

dukbong:
Use instanceof Pattern Matching


marcusdacoregio:
Thanks @dukbong, this is now merged into `main`.

Issue #15093, opened 2024-05-18
Add Kotlin support in `DefaultMethodSecurityExpressionHandler` 


call-me-baki:
**Expected Behavior**
When using the [`@PreFilter`](https://github.com/spring-projects/spring-security/blob/ff94944313d97b7c02ba8732ebbe63e3392ca7a4/core/src/main/java/org/springframework/security/access/prepost/PreFilter.java) and [`@PostFilter`](https://github.com/spring-projects/spring-security/blob/ff94944313d97b7c02ba8732ebbe63e3392ca7a4/core/src/main/java/org/springframework/security/access/prepost/PostFilter.java) annotations with Kotlin's [`Collection`](https://github.com/JetBrains/kotlin/blob/12f5819170c546cbf0cbae4824f06d41d98c3223/libraries/stdlib/native-wasm/src/kotlin/collections/Collection.kt) and [`Map`](https://github.com/JetBrains/kotlin/blob/12f5819170c546cbf0cbae4824f06d41d98c3223/libraries/stdlib/native-wasm/src/kotlin/collections/Map.kt), the filtering logic should be applied as it is in Java code.

**Current Behavior**
Any Kotlin `Collection`s and empty `Map`s throw an [`UnsupportedOperationException`](https://github.com/JetBrains/kotlin/blob/12f5819170c546cbf0cbae4824f06d41d98c3223/libraries/stdlib/jvm/runtime/kotlin/TypeAliases.kt#L17) as these types in Kotlin are immutable, but the implementation of [`DefaultMethodSecurityExpressionHandler`](https://github.com/spring-projects/spring-security/blob/64e2a2ff8b493aa0298043128b7329e2a7362e4b/core/src/main/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandler.java) attempts to mutate their state.  

Minimal reproducible example:
- #15094

**Context**
For `Collection`s, that happens in the following lines:
https://github.com/spring-projects/spring-security/blob/64e2a2ff8b493aa0298043128b7329e2a7362e4b/core/src/main/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandler.java#L154-L155

For `Maps`s, the same happens in the following lines:
https://github.com/spring-projects/spring-security/blob/64e2a2ff8b493aa0298043128b7329e2a7362e4b/core/src/main/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandler.java#L192-L193

[...]

PR #15094, opened 2024-05-18
#15093 minimal reproducible example

call-me-baki:
Represents a minimal reproducible example for the following issue:
- #15093 

The example is in the unit test named `DefaultMethodSecurityExpressionHandlerKotlinTests`. 

To make testing with Kotlin possible, the required dependencies and configuration were added to  `core/spring-security-core.gradle`.


jzheaux:
Thanks, @call-me-baki! Now that we are underway on your PR, I'll close this sample.

PR #15095, opened 2024-05-18
Add Kotlin support to PreFilter and PostFilter annotations

call-me-baki:
Closes the following issue:
- #15093

**Summary**
Adds support for Kotlin in [`DefaultMethodSecurityExpressionHandler`](https://github.com/spring-projects/spring-security/blob/64e2a2ff8b493aa0298043128b7329e2a7362e4b/core/src/main/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandler.java).

It additionally updates the documentation (both `javadoc` and `asciidoc`) to reflect the current implementation details and usage.

**Implementation details**
Attempts to mutate the state of either the `Collection` or the `Map`:
- if an `UnsupportedOperationException` is not thrown, returns the instance with its state mutated.
- if an `UnsupportedOperationException` is thrown, returns a new instance with the filtered state.

**Tests**
The original Java test [`DefaultMethodSecurityExpressionHandlerTests`](https://github.com/spring-projects/spring-security/blob/64e2a2ff8b493aa0298043128b7329e2a7362e4b/core/src/test/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandlerTests.java) has been left unchanged and acts as a regression test.

A new test [`DefaultMethodSecurityExpressionHandlerKotlinTests`](https://github.com/call-me-baki/spring-security/blob/c59c632c6ad842744b11a5ac97ebcecd38b69090/core/src/test/kotlin/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandlerKotlinTests.kt) has been added to assert over the correct behavior when using Kotlin instead of Java.

### DIFF
--- a/core/src/main/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandler.java
+++ b/core/src/main/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -52,6 +52,7 @@ import org.springframework.util.Assert;
  *
  * @author Luke Taylor
  * @author Evgeniy Cheban
+ * @author Blagoja Stamatovski
  * @since 3.0
  */
 public class DefaultMethodSecurityExpressionHandler extends AbstractSecurityExpressionHandler<MethodInvocation>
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandler.java
```diff
@@ -151,9 +153,8 @@ private <T> Object filterCollection(Collection<T> filterTarget, Expression filte
 			}
 		}
 		this.logger.debug(LogMessage.format("Retaining elements: %s", retain));
-		filterTarget.clear();
```
jzheaux, line 154: For passivity reasons, instead of changing the default strategy, what if the code first tried to modify the existing collection, and then if that fails, return the new instance? In this way, applications that are dependent on a specific concrete collection type will still continue to function.

core/src/main/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandler.java
```diff
@@ -109,12 +109,14 @@ private MethodSecurityExpressionOperations createSecurityExpressionRoot(Supplier
 	}
 
 	/**
```
jzheaux, line 111: Will you please change the copyright date on any edited files that are `20xx-20yy` to be `20xx-2024`?

jzheaux (changes requested):
Thanks, @call-me-baki! I've left some feedback inline.

call-me-baki:
Thank you @jzheaux for the feedback! 🙏🏼 

I've addressed the requested changes, would you please have a look again? 

Regarding adding myself to JavaDoc's authors, I'm completely fine if we revert that commit should this change not warrant it.

jzheaux:
Thanks for the updates, @call-me-baki! In preparation for merging, will you please do the following:

1. Run `./gradlew format && ./gradlew :spring-security-core:check` and make any changes the output requires
2. Squash your commits into one
3. Ensure your commit message is brief and also contains `Closes gh-15093`, for example:

```
Add Kotlin support for Pre-PostFilter Annotations

Closes gh-15093
```

call-me-baki:
Hey @jzheaux! Thanks for your support and guidance. 

I ran `./gradlew format && ./gradlew :spring-security-core:check`, there wasn't anything actionable. 
The commits have been squashed and I hope that the commit message is fitting. 

Looking forward contributing to Spring Security in the future!

jzheaux (approved):

jzheaux:
Thanks again, @call-me-baki! This is now merged into `main` and will go out in the next milestone release. I also added a slight polish in aa9bf83c6d0be0c40e868e085dfd8e809afa4e91 to clarify variable names and align logging levels.

Issue #12218, opened 2022-11-16
Restructure SecurityFilterChain logs


jzheaux:

jzheaux:
Yes, @awydini, it's yours.

PR #15096, opened 2024-05-19
Adjust DefaultSecurityFilterChain Logging Level and Simplify Filter Logging

baezzys:
### Overview
This PR addresses issue #12218 

1. Changes the logging level of `DefaultSecurityFilterChain` from INFO to DEBUG to align with `FilterChainProxy`.
2. Logs only the class names of filters instead of their full `toString` of filter.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/DefaultSecurityFilterChain.java
+++ b/web/src/main/java/org/springframework/security/web/DefaultSecurityFilterChain.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -32,6 +32,7 @@ import org.springframework.security.web.util.matcher.RequestMatcher;
  * Standard implementation of {@code SecurityFilterChain}.
  *
  * @author Luke Taylor
+ * @author Jinwoo Bae
  * @since 3.1
  */
 public final class DefaultSecurityFilterChain implements SecurityFilterChain {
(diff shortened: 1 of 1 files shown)


web/src/main/java/org/springframework/security/web/DefaultSecurityFilterChain.java
```diff
@@ -48,10 +50,13 @@ public DefaultSecurityFilterChain(RequestMatcher requestMatcher, Filter... filte
 
 	public DefaultSecurityFilterChain(RequestMatcher requestMatcher, List<Filter> filters) {
 		if (filters.isEmpty()) {
-			logger.info(LogMessage.format("Will not secure %s", requestMatcher));
+			logger.debug(LogMessage.format("Will not secure %s", requestMatcher));
 		}
 		else {
-			logger.info(LogMessage.format("Will secure %s with %s", requestMatcher, filters));
+			logger.debug(LogMessage.format("Will secure %s with filters: %s", requestMatcher,
+					filters.stream()
```
jzheaux, line 57: To remain in line with https://github.com/spring-projects/spring-security/issues/7154 will you please use a for loop instead?

jzheaux (changes requested):
Thanks, @baezzys, this is helpful. I've left some feedback inline.

web/src/main/java/org/springframework/security/web/DefaultSecurityFilterChain.java
baezzys: Thank you for the review. I will update it to use a for loop as suggested.

jzheaux (approved):

jzheaux:
Thanks, @baezzys! This is now merged into `main`. I also added [a polish](https://github.com/spring-projects/spring-security/commit/6956ed693c734bc2359bd625e6d9bd76c0bffed5) to reuse some of the String manipulation support in Spring Framework.

PR #15129, opened 2024-05-22
Fix Kotlin example in authorize-http-requests.adoc

theHacker:
- Consistency: Replaced mix of tabs/spaces with spaces indentation

---

Something was mixed-up with the code example at https://docs.spring.io/spring-security/reference/servlet/authorization/authorize-http-requests.html.
Not sure if it's perfect (Java has `// ...`, Kotlin and XML don't), but it should be an improvement 😄


marcusdacoregio:
Thanks, this is now merged into `main`.

PR #15222, opened 2024-06-11
Fix wrong explanation for @PostAuthorize annotation

Junhyunny:
Hi.

In docs about method based authorization, there is a wrong explanation. This should be `AuthorizationManagerAfterMethodInterceptor` for `@PostAuthorize` annotation.

Thanks.


marcusdacoregio:
Thanks @Junhyunny, this is now merged into `main`.

PR #15255, opened 2024-06-14
Add getter for jwt validator on NimbusJwtDecoder

mrodal:
Closes https://github.com/spring-projects/spring-security/issues/6909


mrodal:
closing since https://github.com/spring-projects/spring-boot/pull/35874 addresses this use case

Issue #15264, opened 2024-06-17
Find Roles for Authority using the RoleHierarchy?


nielsbasjes:
My current understanding (and please correct me if this incorrect) is that 
- an Authority is essentially an externally managed 'flag' that indicates what the user is. In the cases I have seen this is usually a list of groups/teams/... of which the user is a member.
- a Role is essentially a grouping of permissions inside the application. Users in a specific Role can access certain endpoints and do certain operations.

At the technical level these 2 concepts are the same, the only notable difference is that the `Role` has a prefix (default = "ROLE_") in the string that represents it.

Now recently the `RoleHierarchyImpl` has a Builder that supports a much easier way of defining the roles and this way hiding these prefixes has become much cleaner.

This code looks something like this:
```
@Bean
static RoleHierarchy roleHierarchy() {
    return RoleHierarchyImpl.withDefaultRolePrefix()
        .role("ADMIN").implies("STAFF", "TEAMS")
        .role("TEAMS").implies("OPERATORS")
        .role("STAFF").implies("USER")
        .role("USER").implies("GUEST")
        .build();
}
```

Since the Authorities and Roles are technically `the same` I have the suggestion to extend this builder to allow for something like this snippet.
Here the `role()` and `authority()` have 1 difference: The prefix is used or not.

```
@Bean
static RoleHierarchy roleHierarchy() {
    return RoleHierarchyImpl.withDefaultRolePrefix()
        .role("ADMIN").implies("STAFF", "TEAMS")
        .role("TEAMS").implies("OPERATORS")
        .role("STAFF").implies("USER")
        .role("USER").implies("GUEST")

        .authority("TEAM_ABC").implies("TEAMS")
        .build();
}
```

If you like this idea then I'll put up a pull request.

If you think this is a totally wrong approach (which is actually quite likely) then please tell me what the right way of mapping an external Authority to a Role is?

marcusdacoregio:
Hi @nielsbasjes, thanks for the suggestion.

Yes, I think it makes sense since the `RoleHierarchy` is applied when we use `hasAuthority`/`hasAnyAuthority`.
Feel free to provide a PR and we can go from there.

nielsbasjes:
Working on it.
In the text format you can specify
```
   ROLE_A > ROLE_B
   ROLE_A > ROLE_C
```
which gives 'A' both 'B' and 'C'

In the builder this is broken.

This example gives 'A' only 'C' (the last one)
```
[...]

PR #15272, opened 2024-06-18
Support Authorities Without Role Prefix in `RoleHierarchyImpl` Builder

nielsbasjes:
@marcusdacoregio here is my take on https://github.com/spring-projects/spring-security/issues/15264

This is 3 changes (hence 3 commits):
- The proposed way of being able to add the mapping from any authority into 1 or more roles.
- I found that if you add to a hierarchy builder the same role with some extra roles the existing roles are discarded. This is different from what you get with the text format.
- The tests to makes sure this actually works became unreadable (hence I wasn't sure anymore) so I added a few helper functions to make it a lot better to understand what the test does.

Are the documentation changes I created enough?

### DIFF
--- a/core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImpl.java
+++ b/core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImpl.java
@@ -290,6 +290,7 @@ public class RoleHierarchyImpl implements RoleHierarchy {
 		 * @param authority the highest authority in this branch
 		 * @return a {@link ImpliedRoles} to define the child roles for the
 		 * <code>authority</code>
+		 * @since 6.4
 		 */
 		public ImpliedRoles authority(String authority) {
 			Assert.hasText(authority, "authority must not be empty");
--- a/core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImpl.java
+++ b/core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -282,7 +282,18 @@ public class RoleHierarchyImpl implements RoleHierarchy {
 		 */
 		public ImpliedRoles role(String role) {
 			Assert.hasText(role, "role must not be empty");
-			return new ImpliedRoles(role);
+			return new ImpliedRoles(this.rolePrefix.concat(role));
+		}
+
+		/**
+		 * Creates a new hierarchy branch to define an authority and its child roles.
+		 * @param authority the highest authority in this branch
+		 * @return a {@link ImpliedRoles} to define the child roles for the
+		 * <code>authority</code>
+		 */
+		public ImpliedRoles authority(String authority) {
+			Assert.hasText(authority, "authority must not be empty");
+			return new ImpliedRoles(authority);
 		}
 
 		/**
@@ -299,7 +310,7 @@ public class RoleHierarchyImpl implements RoleHierarchy {
 			for (String impliedRole : impliedRoles) {
 				withPrefix.add(new SimpleGrantedAuthority(this.rolePrefix.concat(impliedRole)));
 			}
-			this.hierarchy.put(this.rolePrefix.concat(role), withPrefix);
+			this.hierarchy.put(role, withPrefix);
 			return this;
 		}
 
--- a/core/src/test/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImplTests.java
+++ b/core/src/test/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImplTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2016 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -249,14 +249,23 @@ public class RoleHierarchyImplTests {
 			.implies("B")
 			.role("B")
 			.implies("C", "D")
+			.authority("C")
+			.implies("E", "F", "B")
 			.build();
-		List<GrantedAuthority> flatAuthorities = AuthorityUtils.createAuthorityList("ROLE_A");
-		List<GrantedAuthority> allAuthorities = AuthorityUtils.createAuthorityList("ROLE_A", "ROLE_B", "ROLE_C",
+		List<GrantedAuthority> flatAuthorities1 = AuthorityUtils.createAuthorityList("ROLE_A");
+		List<GrantedAuthority> allAuthorities1 = AuthorityUtils.createAuthorityList("ROLE_A", "ROLE_B", "ROLE_C",
 				"ROLE_D");
 
 		assertThat(roleHierarchyImpl).isNotNull();
-		assertThat(roleHierarchyImpl.getReachableGrantedAuthorities(flatAuthorities))
-			.containsExactlyInAnyOrderElementsOf(allAuthorities);
+		assertThat(roleHierarchyImpl.getReachableGrantedAuthorities(flatAuthorities1))
+			.containsExactlyInAnyOrderElementsOf(allAuthorities1);
+
+		List<GrantedAuthority> flatAuthorities2 = AuthorityUtils.createAuthorityList("C");
+		List<GrantedAuthority> allAuthorities2 = AuthorityUtils.createAuthorityList("C", "ROLE_B", "ROLE_C", "ROLE_D",
+				"ROLE_E", "ROLE_F");
+		assertThat(roleHierarchyImpl.getReachableGrantedAuthorities(flatAuthorities2))
+			.containsExactlyInAnyOrderElementsOf(allAuthorities2);
+
 	}
 
 	@Test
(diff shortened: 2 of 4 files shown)


nielsbasjes:
> Hi @nielsbasjes, thanks for the report.
> 
> > I found that if you add to a hierarchy builder the same role with some extra roles the existing roles are discarded. This is different from what you get with the text format.
> 
> Would you please move this to another PR? From my perspective it isn't exactly a bug but a support that should be added to the builder. A new PR is better so we can track exactly when/what changed.

I have removed that part from this PR.
After you have approved this I'll put that up for separate approval (to avoid needless merge complexity).

nielsbasjes:
@marcusdacoregio is this how you want this merge request? 
If not then just tell me what I need to change.

marcusdacoregio:
Hi @nielsbasjes. I plan to review this PR before 6.4.0-M1 on July 15th. I'll leave my feedback if there is anything else to be changed.

nielsbasjes:
@marcusdacoregio How shall I submit the other change? I'm asking because it is partially overlapping code.

core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImpl.java
```diff
@@ -282,7 +282,18 @@ private Builder(String rolePrefix) {
...
-			return new ImpliedRoles(role);
+			return new ImpliedRoles(this.rolePrefix.concat(role));
+		}
+
+		/**
+		 * Creates a new hierarchy branch to define an authority and its child roles.
+		 * @param authority the highest authority in this branch
+		 * @return a {@link ImpliedRoles} to define the child roles for the
+		 * <code>authority</code>
+		 */
```
marcusdacoregio, line 293: Can you please add `@since 6.4`? It is also important to mention that this method does not include the `rolePrefix`.

core/src/test/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImplTests.java
```diff
@@ -249,29 +250,34 @@ public void testBuilderWithDefaultRolePrefix() {
...
-				"ROLE_D");
 
-		assertThat(roleHierarchyImpl).isNotNull();
-		assertThat(roleHierarchyImpl.getReachableGrantedAuthorities(flatAuthorities))
-			.containsExactlyInAnyOrderElementsOf(allAuthorities);
+		assertHierarchy(roleHierarchyImpl).givesToAuthorities("ROLE_A")
+			.theseAuthorities("ROLE_A", "ROLE_B", "ROLE_C", "ROLE_D");
+
+		assertHierarchy(roleHierarchyImpl).givesToAuthorities("C")
+			.theseAuthorities("C", "ROLE_B", "ROLE_C", "ROLE_D", "ROLE_E", "ROLE_F");
```
marcusdacoregio, line 261: As of now, using `.authority("C").implies("E", "F", "B")`, means that `C` will have `ROLE_E`, `ROLE_F` and `ROLE_B`. What if I want to say that authority `user:write` implies in `user:read`, `user:block`, for example?

I'm thinking that when using `authority(...)` the role prefix should be ignored for the implied authorities too. It should be possible to do `.authority("user:write").implies("user:read", "user:delete")`. What do you think?

marcusdacoregio (changes requested):

core/src/test/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImplTests.java
nielsbasjes: My mindset when writing this was that 
- the authority is a "label" that is provided externally provided (via authentication).
- the role is a "label" that is internal to use to actually give access to things.

So from your example assuming the 'user:...' are externally provided authorities I see this as a possible way of doing this:

```
.authority("user:read")  .implies("USER_READ") 
.authority("user:write") .implies("USER_WRITE")
.authority("user:delete").implies("USER_DELETE")

.authority("TeamXYZMembers").implies("USER_READ", "USER_WRITE")

.role("USER_WRITE")      .implies("USER_READ", "USER_DELETE")
```

This would make all members of "Team XYZ" have the roles USER_READ, USER_WRITE and USER_DELETE.

Primary question: Does this reasoning make sense? It is correct?
If this is correct; Perhaps it should be documented as such?
(if you want I'll put up a separate merge request with such documentation)

The way I normally structure this is that in an application I define roles being groups of users with the same set of tasks like "Administrator", "Manager", "User", etc. Those are then assigned to the user in an external IGA/IAM system (often based on department/job title/...).

Then via the authentication of the specific user; the application receives one or more of these "external roles" as the authorities via the authentication.

From there my application has something like this
```
.authority("User")         .implies("USER_READ") 
.authority("Manager")      .implies("USER_WRITE")
.authority("Administrator").implies("USER_DELETE", "SHOW_DASHBOARD")

.role("USER_WRITE")       .implies("USER_READ")
.role("USER_DELETE")      .implies("USER_WRITE")
```

As said; I you want I'm willing to make a first draft on documenting this.

nielsbasjes: If you want to be able to imply an authority then something like this is needed:
``` .authority("user:write").impliesAuthority("user:read", "user:delete")```

A separate function is needed because there is no way to make any distinction between `authority` or `role` because they are both just a string.

Also
- I currently disagree with this approach given how I look at the `authority` (external) and `role` (internal).
- It is not needed to get the desired effect as shown in the example I posted yesterday.

**Comment by marcusdacoregio** [comment]
Hi @nielsbasjes. I was talking to @jzheaux about this feature and we come to a conclusion that we should refrain from adding the `.authority(...)` method to the builder. The main reason is because that we can cause more confusion with methods with a not so predictable behavior, like we discussed [here](https://github.com/spring-projects/spring-security/pull/15272#discussion_r1652732861), which can easily lead to misconfiguration.

Our suggestion is to add a new `public static Builder withNoRolePrefix()` method. This is nice because the application is either always supplying fully-qualified authorities or always supplying roles, making it easier to infer at a glance when the code is read later on.

What do you think?

**Comment by nielsbasjes** [comment]
> Hi @nielsbasjes. I was talking to @jzheaux about this feature and we come to a conclusion that we should refrain from adding the `.authority(...)` method to the builder. 

Unexpected but you're the committers here.

> The main reason is because that we can cause more confusion with methods with a not so predictable behavior, like we discussed [here](https://github.com/spring-projects/spring-security/pull/15272#discussion_r1652732861), which can easily lead to misconfiguration.

Where do you see "not so predictable behavior"?
- In `.role("X")` X is always a role (i.e. with prefix).
- In `.implies("X")` X is always a role (i.e. with prefix).
- In `.authority("X")` X is always an authority (i.e. without prefix).

From where I'm standing the confusion (which _is_ there) stems mainly from the difference between `authority` and `role` being quite unclear (at least to me when reading the documentation) and being stored in the same structures with a hidden prefix to determine the difference.

My code and explanation (which as I said I'm willing to put into the documentation if so desired) make it clearer and reliably usable ... at least to me it does.

> Our suggestion is to add a new `public static Builder withNoRolePrefix()` method. This is nice because the application is either always supplying fully-qualified authorities or always supplying roles, making it easier to infer at a glance when the code is read later on.

That would be a completely silly change because it is 100% the same as simply doing `.withRolePrefix("")`. I do not see that as a valid code improvement because that makes the distinction between authorities and roles even harder to understand.

In my applications I do not want an external authority to accidentally match a role ever; I have even created this last week to make that even stronger to ensure no problems if someone creates an externally provided authority "ROLE_X":
```
@Bean
fun grantedAuthorityDefaults(): GrantedAuthorityDefaults = GrantedAuthorityDefaults(UUID.randomUUID().toString())
```

> What do you think?

I think it would really help if you document what you see as the distinction between `authority` and `role` because from this discussion it seems you have a totally different idea than I have.

**Comment by marcusdacoregio** [comment]
I don't think I follow clearly the [following statement](https://github.com/spring-projects/spring-security/pull/15272#discussion_r1654443909):

> My mindset when writing this was that
>
> - the authority is a "label" that is provided externally provided (via authentication).
> - the role is a "label" that is internal to use to actually give access to things.

In my opinion, this is probably mostly related to your business needs than how the framework sees it. Both authorities and roles are the same thing, sometimes it makes sense to group users into roles and sometimes you need a more granular control so you can opt-in for authorities. But that is just a matter of naming, when performing any authorization logic the framework does not differ whether it is an authority or a role.

> That would be a completely silly change because it is 100% the same as simply doing .withRolePrefix(""). I do not see that as a valid code improvement because that makes the distinction between authorities and roles even harder to understand.

That's the point, there is no difference between those two in general. The example that you gave earlier should have a different arrangement, IMO:

```java
RoleHierarchyImpl.withDefaultRolePrefix()
  .role("USER_WRITE")       .implies("USER_READ")
  .role("USER_DELETE")      .implies("USER_WRITE")
  .authority("User")         .implies("USER_READ") 
  .authority("Manager")      .implies("USER_WRITE")
  .authority("Administrator").implies("USER_DELETE", "SHOW_DASHBOARD")
```

should be:

```java
RoleHierarchyImpl.withNoRolePrefix()
  .role("ROLE_USER")       .implies("USER_READ")
  .role("USER_WRITE")       .implies("USER_READ")
  .role("USER_DELETE")      .implies("USER_WRITE")
  .authority("User")         .implies("USER_READ") 
  .authority("Manager")      .implies("USER_WRITE")
  .authority("Administrator").implies("USER_DELETE", "SHOW_DASHBOARD")
```

marcusdacoregio:
I don't think I clearly follow the [following statement](https://github.com/spring-projects/spring-security/pull/15272#discussion_r1654443909):

> My mindset when writing this was that
>
> - the authority is a "label" that is provided externally provided (via authentication).
> - the role is a "label" that is internal to use to actually give access to things.

In my opinion, this is probably mostly related to your business needs than how the framework sees it. Both authorities and roles are the same thing, sometimes it makes sense to group users into roles and sometimes you need a more granular control so you can opt-in for authorities. But that is just a matter of naming, when performing any authorization logic the framework does not differ whether it is an authority or a role.

> That would be a completely silly change because it is 100% the same as simply doing .withRolePrefix(""). I do not see that as a valid code improvement because that makes the distinction between authorities and roles even harder to understand.

That's the point, there is no difference between those two in general. The example that you gave earlier from my point of view should have a different arrangement. Roles are often used to group authorities and simplify access control logic:

```java
RoleHierarchyImpl.withDefaultRolePrefix()
  .role("USER_WRITE")        .implies("USER_READ")
  .role("USER_DELETE")       .implies("USER_WRITE")
  .authority("User")         .implies("USER_READ") 
  .authority("Manager")      .implies("USER_WRITE")
  .authority("Administrator").implies("USER_DELETE", "SHOW_DASHBOARD")
```

should be:

```java
RoleHierarchyImpl.withNoRolePrefix()
  .role("ROLE_USER")             .implies("USER_READ")
  .role("ROLE_MANAGER")          .implies("USER_WRITE")
  .role("ROLE_ADMINISTRATOR")    .implies("USER_DELETE", "SHOW_DASHBOARD")
  .role("USER_WRITE")       .implies("USER_READ")
  .role("USER_DELETE")      .implies("USER_WRITE")
```

In the example above, there is three roles and based on the role that the user have authorities are given to them. It sounds like it's the other way around than your understanding, but authorities/roles can be given to users directly, with no hard rule on whether it should be external/internal. This is also (briefly) mentioned [in the documentation](https://docs.spring.io/spring-security/reference/servlet/authorization/method-security.html#favor-granting-authorities).

> I think it would really help if you document what you see as the distinction between authority and role because from this discussion it seems you have a totally different idea than I have.

Absolutely, we can include in the reference docs some guidance on the difference between those two.

nielsbasjes:
> In my opinion, this is probably mostly related to your business needs than how the framework sees it.

Yes, so that seems to be the big problem for me here; The framework has two flags that are technically identical and also do not have any logical distinction between them.
So duplicate APIs that do the same, have the same implementation and only differ because one prefixes everything with a prefix.

That makes the current situation very confusing for me.

I'm confronted with authorities which are of external origin (i.e. usergroups from the IAM/IGA solution) which I need to handle in an as clean as possible way within the application.

My need is to map these external values (without any prefix; stored as authorities) into Roles in a clean way, that is what I'm trying to achieve here. 

Key question: What would you like to do with this? How do you see this? I'm willing to change it as long as I understand how you want it.

For now: I have taken the code I put here as a starting point and created a company internal variant that meets our business needs: Mapping authorities and roles into new roles.


Note: I had to copy some of the code (instead of inherit) because parts of it were `private` instead of `protected` (just making the constructor with parameters protected would have fixed it for me).

PR #15278, opened 2024-06-19
Fix level of log message.

OrangeDog:
Surrounding code appears to intend it to be at trace, not debug.


jzheaux:
Thanks, @OrangeDog, this is now merged into `main`

PR #15296, opened 2024-06-24
docs: update doc for cors intgeration with servlet apps

HusseinOkasha:
- update file docs/modules/ROOT/pages/servlet/integrations/cors.adoc 
     - add setAllowedHeaders("Authorization") to alert other developers that headers like "Authorization" is not allowed by default.


marcusdacoregio:
Hi @HusseinOkasha, thanks for the PR.

I don't think we should add such information without context on why it should be there since the `Authorization` header is very specific to OAuth in Spring Security. It might even cause some confusion for people who are not using the `Authorization` header or OAuth.

For that reason, I'm going to close this, but if we see the community reporting confusion on what to add to the `CorsConfigurationSource` we can consider reopening it.

PR #15307, opened 2024-06-27
Augment role hierarchy in `RoleHierarchyImpl` Builder

nielsbasjes:
This change (as I have the code here) requires https://github.com/spring-projects/spring-security/pull/15272 to be merged (and this should be rebased before merging).

Assume this role mapping in text send to the `RoleHierarchyImpl`:
```
ROLE_X > ROLE_A
ROLE_X > ROLE_B
```

The expectation is that this code fragment of using the `RoleHierarchyImpl` Builder would yield the same effect:
```
.role("X").implies("A")
.role("X").implies("B")
```

The current behavior of the Builder is that this results in the role "X" ONLY being assigned the role "B" (the last one assigned).

With this change the the role "X" will be assigned BOTH the roles "A" and "B" making it the same as the text format again.

FYI: @marcusdacoregio


marcusdacoregio:
Fixed via https://github.com/spring-projects/spring-security/pull/15717

PR #15318, opened 2024-06-28
Update architecture.adoc

dboldureanu:
The list of filters is printed at DEBUG level on the application startup and not INFO level, see DefaultSecurityFilterChain


jzheaux:
Thanks, @dboldureanu! This is now merged into `main`.

PR #15325, opened 2024-06-30
Fix malformed list in "Using Method Parameters" documentation

Marcono1234:
See [section "Using Method Parameters" in the current documentation](https://docs.spring.io/spring-security/reference/servlet/authorization/method-security.html#using_method_parameters). It looks like it is supposed to list the order in which Spring tries to discover parameter names:
1. `@P`
2. `@Param`
3. `javac` `-parameters`
4. `javac` with debug symbols

However, currently it is erroneously displayed as:
1. `@P`
   - `@Param`
   - `javac` `-parameters`
   - `javac` with debug symbols

This pull request tries to fix this.

Though please let me know if I misunderstood this, or if there is a better way to solve this.


Marcono1234:

jzheaux:
Thanks, @Marcono1234! This is now merged into `main`.

PR #15327, opened 2024-06-30
Two small fixes to the Servlet's Authentication Events documentation

stefanganzer:
In https://docs.spring.io/spring-security/reference/servlet/authentication/events.html#_default_event, the given example does not work, as
* in the Java case, the assignment type `AuthenticationEventPublisher` does not have the expected method `setDefaultAuthenticationFailureEvent`, but `DefaultAuthenticationEventPublisher` has.
* in both the Java and Kotlin case the class `GenericAuthenticationFailureEvent` does not exist. `AbstractAuthenticationFailureEvent` could be used instead.


jzheaux:
Thanks, @stefanganzer! This is now merged into `main`.

Issue #15286, opened 2024-06-20
Improve @AuthenticationPrincipal meta-annotations


rwinch:
It would be nice if we could improve the support of `@AuthenticationPrincipal` meta-annotations to align with method security. For example, the following doesn't appear to work:

```java
@Retention(RetentionPolicy.RUNTIME)
@AuthenticationPrincipal
public @interface CurrentUser {
	@AliasFor(annotation = AuthenticationPrincipal.class)
	String expression() default "";
}
```

```java
@CurrentUser(expression = "id") Long currentUserId
```

It would be nice to also support something like this:

```java
@Retention(RetentionPolicy.RUNTIME)
@AuthenticationPrincipal("principal.{property}")
public @interface CurrentUser {
	String property() default "";
}
```

PR #15344, opened 2024-07-02
Improve @AuthenticationPrincipal meta-annotations

kse-music:
Closes gh-15286

### DIFF
--- a/messaging/src/main/java/org/springframework/security/messaging/context/AuthenticationPrincipalArgumentResolver.java
+++ b/messaging/src/main/java/org/springframework/security/messaging/context/AuthenticationPrincipalArgumentResolver.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/messaging/src/main/java/org/springframework/security/messaging/context/AuthenticationPrincipalArgumentResolver.java
+++ b/messaging/src/main/java/org/springframework/security/messaging/context/AuthenticationPrincipalArgumentResolver.java
@@ -17,9 +17,10 @@
 package org.springframework.security.messaging.context;
 
 import java.lang.annotation.Annotation;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.springframework.core.MethodParameter;
-import org.springframework.core.annotation.AnnotationUtils;
 import org.springframework.expression.Expression;
 import org.springframework.expression.ExpressionParser;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
@@ -27,6 +28,9 @@ import org.springframework.expression.spel.support.StandardEvaluationContext;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.handler.invocation.HandlerMethodArgumentResolver;
 import org.springframework.security.core.Authentication;
+import org.springframework.security.core.annotation.AnnotationSynthesizer;
+import org.springframework.security.core.annotation.AnnotationSynthesizers;
+import org.springframework.security.core.annotation.AnnotationTemplateExpressionDefaults;
 import org.springframework.security.core.annotation.AuthenticationPrincipal;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.security.core.context.SecurityContextHolderStrategy;
@@ -83,6 +87,7 @@ import org.springframework.util.StringUtils;
  * </pre>
  *
  * @author Rob Winch
+ * @author DingHao
  * @since 4.0
  */
 public final class AuthenticationPrincipalArgumentResolver implements HandlerMethodArgumentResolver {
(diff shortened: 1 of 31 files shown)


messaging/src/main/java/org/springframework/security/messaging/context/AuthenticationPrincipalArgumentResolver.java
```diff
@@ -83,15 +94,20 @@
...
 public final class AuthenticationPrincipalArgumentResolver implements HandlerMethodArgumentResolver {
 
 	private SecurityContextHolderStrategy securityContextHolderStrategy = SecurityContextHolder
 		.getContextHolderStrategy();
 
+	private final Map<MethodParameter, Annotation> cachedAttributes = new ConcurrentHashMap<>();
+
 	private ExpressionParser parser = new SpelExpressionParser();
 
+	private PrePostTemplateDefaults defaults = new PrePostTemplateDefaults();
```
jzheaux, line 109: I think we'll want to introduce a new class, maybe `AuthenticationPrincipalTemplateDefaults`. By the name and JavaDoc `PrePostTemplateDefaults` is intended for the pre/post annotations.

messaging/src/main/java/org/springframework/security/messaging/context/AuthenticationPrincipalArgumentResolver.java
```diff
@@ -133,26 +149,63 @@ public void setSecurityContextHolderStrategy(SecurityContextHolderStrategy secur
...
+	/**
+	 * Configure AuthenticationPrincipal template resolution
+	 * <p>
+	 * By default, this value is <code>null</code>, which indicates that templates should
+	 * not be resolved.
+	 * @param defaults - whether to resolve AuthenticationPrincipal templates parameters
+	 * @since 6.4
+	 */
+	public void setTemplateDefaults(@Nullable PrePostTemplateDefaults defaults) {
+		if (defaults != null) {
```
jzheaux, line 161: Please instead don't allow null values. We don't want methods to silently not do what they say they will do.

messaging/src/main/java/org/springframework/security/messaging/context/AuthenticationPrincipalArgumentResolver.java
```diff
@@ -133,26 +149,63 @@ public void setSecurityContextHolderStrategy(SecurityContextHolderStrategy secur
...
 		}
-		Annotation[] annotationsToSearch = parameter.getParameterAnnotations();
-		for (Annotation toSearch : annotationsToSearch) {
-			annotation = AnnotationUtils.findAnnotation(toSearch.annotationType(), annotationClass);
-			if (annotation != null) {
-				return annotation;
-			}
-		}
-		return null;
+		return (T) this.cachedAttributes.computeIfAbsent(parameter, (k) -> {
```
jzheaux, line 179: Let's consider turning `AuthorizationAnnotationUtils` into an API that both the argument resolver and method security can use. Do you have time for that? If not, I can create a separate ticket for it.

jzheaux (changes requested):
Thanks for contributing this, @kse-music! I've left some initial feedback inline.

messaging/src/main/java/org/springframework/security/messaging/context/AuthenticationPrincipalArgumentResolver.java
kse-music: Done

messaging/src/main/java/org/springframework/security/messaging/context/AuthenticationPrincipalArgumentResolver.java
kse-music: Done

messaging/src/main/java/org/springframework/security/messaging/context/AuthenticationPrincipalArgumentResolver.java
kse-music: it's my pleasure

kse-music:
Thanks,@jzheaux ,all feedback is done

jzheaux:
Thanks, @kse-music! Given that this needs to be placed in the other `AuthenticationPrincipalArgumentResolver`s, I extracted the logic in `AuthorizationAnnotationUtils` into an API. Please see if it works for you. If so, please also update this PR to change the other non-deprecated instances of `AuthenticationPrincipalArgumentResolver` in the project.

kse-music:
> Thanks, @kse-music! Given that this needs to be placed in the other `AuthenticationPrincipalArgumentResolver`s, I extracted the logic in `AuthorizationAnnotationUtils` into an API. Please see if it works for you. If so, please also update this PR to change the other non-deprecated instances of `AuthenticationPrincipalArgumentResolver` in the project.

All non-deprecated instances of AuthenticationPrincipalArgumentResolver in the project has done.

jzheaux:
Thanks, @kse-music, this is taking shape very nicely.

I believe our last step is to make this simple to configure by picking up the `AnnotationTemplateExpressionDefaults` bean in configuration.

I've added [documentation](https://github.com/spring-projects/spring-security/pull/15344/files#diff-cf8343e91909a111571bdcae3d007cb61824793545041ec143247ce3e69c1c85) so you can see the expectation that someone should be able to do:

```java
@Bean
public AnnotationTemplateExpressionDefaults templateExpressionDefaults() {
    return new AnnotationTemplateExpressionDefaults();
}
```

And now they will be able to use placeholders in any `@AuthenticationPrincipal` meta-annotation definition.

Are you able to add this to the following `spring-security-config` classes: `WebMvcSecurityConfiguration`, `ServerHttpSecurityConfiguration`, and `WebSocketMessageBrokerSecurityConfiguration`? It's a little harder, however, if you are able, it would be great to add to `WebSocketMessageBrokerSecurityBeanDefinitionParser`, too.

kse-music:
@jzheaux ,I tried to add it

kse-music:
@jzheaux I have added the code, please help to review it.Thank you

jzheaux (approved):

jzheaux:
Thanks for all your help with this PR, @kse-music! I added a couple of commits to add more testing and also to deprecate `PrePostTemplateDefaults`. It's now merged into `main`.

jzheaux:
If you are interested, it would be nice to do the same with `@CurrentSecurityContext`. I've added https://github.com/spring-projects/spring-security/issues/15551 to address that. That PR should go a lot more quickly.

kse-music:
> If you are interested, it would be nice to do the same with `@CurrentSecurityContext`. I've added #15551 to address that. That PR should go a lot more quickly.

Yes,I have added PR #15553

Issue #15090, opened 2024-05-17
Make OpenSamlMetadataRelyingPartyRegistrationConverter public


OrangeDog:
**Expected Behavior**

Class and its methods should be `public`.

**Current Behavior**

Class and its methods are package-private.

**Context**

To allow writing alternatives to `RelyingPartyRegistrations` and others by adapting OpenSAML APIs.
For example, in order to implement verification (https://github.com/spring-projects/spring-security/issues/15018#issuecomment-2110477455) and refreshing (https://github.com/spring-projects/spring-security/issues/15027#issuecomment-2100971114).

jzheaux:
I believe this will be addressed in #12116. I'll leave this ticket open for the moment just in case the other evolves differently than I imagine.

OrangeDog:
Workaround:
```java
try {
    Class<?> converterClass = Class.forName("org.springframework.security.saml2.provider.service.registration.OpenSamlMetadataRelyingPartyRegistrationConverter");
    Constructor<?> converterConstructor = converterClass.getDeclaredConstructor();
    converterConstructor.setAccessible(true);
    Object converterInstance = converterConstructor.newInstance();
    Method converterMethod = converterClass.getDeclaredMethod("convert", EntityDescriptor.class);
    converterMethod.setAccessible(true);
    this.converter = value -> {
        try {
            return (RelyingPartyRegistration.Builder) converterMethod.invoke(converterInstance, value);
        } catch (InvocationTargetException ex) {
            if (ex.getTargetException() instanceof RuntimeException cause) {
                throw cause;
            } else {
                throw new RuntimeException("Cannot convert metadata", ex);
            }
        } catch (IllegalAccessException ex) {
            throw new IllegalStateException("Cannot convert metadata", ex);
        }
    };
}
catch (ReflectiveOperationException ex) {
    throw new IllegalStateException("Cannot initialise metadata converter", ex);
}
```

Issue #15018, opened 2024-05-07
RelyingPartyRegistrations does not verify signatures


OrangeDog:
**Describe the bug**
When an `EntitiesDescriptor`, `EntityDescriptor`, or `IDPSSODescriptor` is being parsed, any included `Signature`s must be verified against a supplied trust store. This is especially critical if the metadata is fetched over an insecure channel. However, the methods in `RelyingPartyRegistrations` do not do this. Nor is it clear how someone might do it after the fact.

The underlying OpenSaml library provides all the necessary functionality. Spring Security just needs to call it. The now unsupported [spring-security-saml](https://github.com/spring-attic/spring-security-saml) does this correctly.

**To Reproduce**
The current API does not provide any way to provide trust material, even if it tried to verify signatures.

**Expected behavior**
See above.

**Additional Information**
This may qualify as a security issue, if people were expecting the previous behaviour of SAML metadata being cryptographically verified.

OrangeDog:
Indeed, there's a big warning in the OpenSAML documentation against the design of spring-security-saml-provider:

> It is very dangerous to attempt to use parts of the library in isolation without making use of all of its relevant components. In particular, implementing your own XML processing code, using XML parsing classes other than the ParserPool components provided by the library, using your own security processing code, omitting proper support for SAML metadata, etc. are all risky choices that may lead to security flaws and incomplete, unsafe, and ill-advised SAML solutions. The Shibboleth Project discourages such approaches in the strongest possible terms. Use all of it that applies to the task at hand, or use none of it.

OrangeDog:
A safer approach is to have OpenSAML do the fetch and validation together, as the previous extension did.

Some example configuration:

```java
@Bean(initMethod = "initialize", destroyMethod = "destroy")
public BasicParserPool parserPool() {
    return new BasicParserPool();
}

@Bean
public HttpClient httpClient() {
    return new net.shibboleth.utilities.java.support.httpclient.HttpClientBuilder().buildClient();
}

@Bean
public MetadataFilter metadataFilter(SignatureTrustEngine trustEngine) {
    MetadataFilter roleFilter = new EntityRoleFilter(List.of(IDPSSODescriptor.DEFAULT_ELEMENT_NAME));
    SignatureValidationFilter validationFilter = new SignatureValidationFilter(trustEngine);
    validationFilter.setRequireSignedRoot(METADATA_URI.getScheme().equals("http"));
[...]

Issue #15017, opened 2024-05-07
RelyingPartyRegistrations typically produces unusable registrationId


OrangeDog:
**Describe the bug**
Methods such as `RelyingPartyRegistrations.collectionFromMetadataLocation` use the `entity` of the asserting party as the `registrationId`.

SAML entity IDs are usually URIs (and often a URL to their metadata), and Spring's SAML support (incl. the default login page) requires the `registrationId` to be added to a URL.

This typically results in an error when trying to access e.g. `https://relyingparty.com/saml2/authenticate/https://assertingparty.com/SAML/metadata.xml`

**To Reproduce**
Construct a `RelyingPartyRegistrationRepository` using `RelyingPartyRegistrations` without changing the default `registrationId`.

**Expected behavior**
`registrationId` should always be generated as URL-safe, or should always be escaped when used in a URL.

**Workaround**
```java
List<RelyingPartyRegistration> registrations = RelyingPartyRegistrations.collectionFromMetadataLocation(url)
        .stream()
        .map(builder -> builder
                .entityId("https://relyingparty.com/saml-metadata.xml")
                // etc.
                .build())
        .map(reg -> reg.mutate()
                .registrationId(reg.getAssertingPartyDetails().getEntityId().replaceAll("[:/?#]+", "_"))
                .build())
        .toList();
```

**Caveat**
Using `URLEncoder.encode` to make it safe is not sufficient, as you still get a 400 Bad Request from a default Spring Boot server.

jzheaux:
Thanks, @OrangeDog, for the suggestion. Note that this is covered in the JavaDoc:

```
* Note that by default the registrationId is set to be the given metadata location,
* but this will most often not be sufficient. To complete the configuration, most
* applications will also need to provide a registrationId, like so:
```

Still, I think this would be good to investigate and improve. I wanted to add this context so it's clear why I'm changing the ticket to an enhancement.

OrangeDog:
The SAML Discovery protocol (which the previous extension implemented) uses a query parameter rather than a path fragment so that entity ids can be transferred safely. Is it possible to do that with the current `authenticationRequestUri` template? A standard  URI encoding would work there.

jzheaux:
Nice idea. That's not yet supported in the DSL, but you could achieve that by publishing the `Saml2WebSsoAuthenticationRequestFilter` in the filter chain and setting the `RequestMatcher` accordingly.

[...]

Issue #12116, opened 2022-11-01
Validate asserting party metadata signature


palakova:
Various elements in metadata can be digitally signed. Although signatures are optional, according to [SAML specification](https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf), section _4.3.3.2 Processing Signed Documents and Fragments_, "Metadata consumers MUST validate signatures, when present."

**Expected Behavior**
Similar as the previous Spring Security SAML Extension implementation: 
Enable to submit a collection of public keys via configuration (e.g. as a part of RelyingPartyRegistration) and verify signature(s) in IdP metadata using PKIX algorithm and using the provided public keys as trust anchors.

**Current Behavior**
I believe metadata signature is not verified. I did not find any way to enable this behaviour.

**Context**
Adding metadata signature increases security and conforms to SAML 2.0 specification. 
As an example, Azure AD signs the root `<EntityDescriptor>` element of it's metadata.

jzheaux:
I think this sounds reasonable, though it would likely involve making `OpenSamlAssertingPartyDetailsConverter` public and polishing it accordingly.

It might change to something like `OpenSamlAssertingPartyDetailsDecoder` (similar to `NimbusJwtDecoder`) and may get an interface like `AssertingPartyDetailsDecoder`. The implementation would take the keys as part of its construction.

sumeetpri:
Hi @jzheaux @sjohnr , metadata consumer must validate signature when presented . Is there any option to add signature?

PR #15349, opened 2024-07-02
Add AssertingPartyMetadataRepository

jzheaux:
This PR introduces `OpenSamlAssertingPartyMetadataRepository`, a class that makes what is otherwise hidden behind `RelyingPartyRegistrations` configurable. For example, before this change, you would do:

```java
RelyingPartyRegistration.Builder builder = RelyingPartyRegistrations.fromMetadataLocation("uri");
```

With this change, you can do:

```java
var repository = OpenSamlAssertingPartyMetadataRepository.fromTrustedMetadataLocation("uri").build();
```

or

```java
var repository = new OpenSamlAssertingPartyMetadataRepository(myMetadataResolver);
```

or

```java
var repository = OpenSamlAssertingPartyMetadataRepository.withMetadataLocation("uri")
    .verificationCredentials((c) -> c.addAll(my, set, of, credentials)).build();
```

followed by:

```java
Collection<RelyingPartyRegistration.Builder> builders = StreamSupport.stream(repository.spliterator(), false)
    .map(RelyingPartyRegistration::withAssertingPartyMetadata);
```

`OpenSamlAssertingPartyMetadataRepository` uses an underlying `MetadataResolver` to refresh the metadata in an expiry-aware fashion. This can be used to back a `RelyingPartyRegistrationRepository` like so:

```java
@Component
public class RefreshableRelyingPartyRegistrationRepository implements IterableRelyingPartyRegistrationRepository {
    private final AssertingPartyMetadataRepository parties;

    // ...

    @Override 
    public RelyingPartyRegistration findByRegistrationId(String registrationId) {
        return applyRelyingParty(this.parties.findByEntityId(registrationId));
    }

    @Override 
    public Iterator<RelyingPartyRegistration> iterator() {
        return StreamSupport.stream(this.parties.spliterator(), false).map(this::applyRelyingParty).iterator();
    }

    private RelyingPartyRegistration applyRelyingParty(AssertingPartyMetadata metadata) {
        return RelyingPartyRegistration.withAssertingPartyMetadata(metadata)
            // apply relying party
            .build();
    }
}
```

### DIFF
--- a/docs/modules/ROOT/pages/servlet/saml2/login/overview.adoc
+++ b/docs/modules/ROOT/pages/servlet/saml2/login/overview.adoc
@@ -484,7 +484,7 @@ public RelyingPartyRegistrationRepository relyingPartyRegistrations() throws Exc
     Saml2X509Credential credential = Saml2X509Credential.verification(certificate);
     RelyingPartyRegistration registration = RelyingPartyRegistration
             .withRegistrationId("example")
-            .assertingPartyDetails(party -> party
+            .assertingPartyMetadata(party -> party
                 .entityId("https://idp.example.com/issuer")
                 .singleSignOnServiceLocation("https://idp.example.com/SSO.saml2")
                 .wantAuthnRequestsSigned(false)
@@ -508,7 +508,7 @@ open fun relyingPartyRegistrations(): RelyingPartyRegistrationRepository {
     val credential: Saml2X509Credential = Saml2X509Credential.verification(certificate)
     val registration = RelyingPartyRegistration
         .withRegistrationId("example")
-        .assertingPartyDetails { party: AssertingPartyDetails.Builder ->
+        .assertingPartyMetadata { party: AssertingPartyMetadata.Builder ->
             party
                 .entityId("https://idp.example.com/issuer")
                 .singleSignOnServiceLocation("https://idp.example.com/SSO.saml2")
@@ -699,7 +699,7 @@ RelyingPartyRegistration relyingPartyRegistration = RelyingPartyRegistration.wit
         .entityId("{baseUrl}/{registrationId}")
         .decryptionX509Credentials(c -> c.add(relyingPartyDecryptingCredential()))
         .assertionConsumerServiceLocation("/my-login-endpoint/{registrationId}")
-        .assertingPartyDetails(party -> party
+        .assertingPartyMetadata(party -> party
                 .entityId("https://ap.example.org")
                 .verificationX509Credentials(c -> c.add(assertingPartyVerifyingCredential()))
                 .singleSignOnServiceLocation("https://ap.example.org/SSO.saml2")
@@ -718,7 +718,7 @@ val relyingPartyRegistration =
             c.add(relyingPartyDecryptingCredential())
         }
         .assertionConsumerServiceLocation("/my-login-endpoint/{registrationId}")
-        .assertingPartyDetails { party -> party
+        .assertingPartyMetadata { party -> party
                 .entityId("https://ap.example.org")
                 .verificationX509Credentials { c -> c.add(assertingPartyVerifyingCredential()) }
                 .singleSignOnServiceLocation("https://ap.example.org/SSO.saml2")
@@ -730,7 +730,7 @@ val relyingPartyRegistration =
 [TIP]
 ====
 The top-level metadata methods are details about the relying party.
-The methods inside `assertingPartyDetails` are details about the asserting party.
+The methods inside `AssertingPartyMetadata` are details about the asserting party.
 ====
 
 [NOTE]
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/OpenSamlAssertingPartyMetadataRepository.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/OpenSamlAssertingPartyMetadataRepository.java
@@ -1,0 +1,383 @@
+/*
+ * Copyright 2002-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.saml2.provider.service.registration;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.Set;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+import javax.annotation.Nonnull;
+
+import net.shibboleth.utilities.java.support.component.ComponentInitializationException;
+import net.shibboleth.utilities.java.support.resolver.CriteriaSet;
+import net.shibboleth.utilities.java.support.resolver.ResolverException;
+import org.opensaml.core.criterion.EntityIdCriterion;
+import org.opensaml.core.xml.config.XMLObjectProviderRegistrySupport;
+import org.opensaml.saml.criterion.EntityRoleCriterion;
+import org.opensaml.saml.metadata.IterableMetadataSource;
+import org.opensaml.saml.metadata.resolver.MetadataResolver;
+import org.opensaml.saml.metadata.resolver.filter.impl.SignatureValidationFilter;
+import org.opensaml.saml.metadata.resolver.impl.AbstractBatchMetadataResolver;
+import org.opensaml.saml.metadata.resolver.impl.ResourceBackedMetadataResolver;
+import org.opensaml.saml.metadata.resolver.index.MetadataIndex;
+import org.opensaml.saml.metadata.resolver.index.impl.RoleMetadataIndex;
+import org.opensaml.saml.saml2.metadata.EntityDescriptor;
+import org.opensaml.saml.saml2.metadata.IDPSSODescriptor;
+import org.opensaml.security.credential.Credential;
+import org.opensaml.security.credential.impl.CollectionCredentialResolver;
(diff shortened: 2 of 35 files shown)


jzheaux:
@OrangeDog will you please take a look at this PR and see if it addresses the use cases you described in https://github.com/spring-projects/spring-security/issues/15090, https://github.com/spring-projects/spring-security/issues/15017, and https://github.com/spring-projects/spring-security/issues/15018 (related to https://github.com/spring-projects/spring-security/issues/12116)?

OrangeDog:
- #15090 ❌ - I still need correct expiry-aware metadata refreshing, so will continue to adapt an `org.opensaml.saml.metadata.resolver.MetadataResolver` . Thus I still need a way to turn a single `EntityDescriptor` into a `RelyingPartyRegistration`.
- #15017 ✔- This looks like it works. I will have a look later to test it. However, what happens if the default encoding turns two different entity ids into the same registration id?
- #12116 ❓ - The default configuration needs to be secure (and the current lack of security should have a CVE filed). If untrusted metadata is provided (no signature from a remote source, no trust material available) it needs to fail. It should have the same behaviour as the deprecated implementation which spent a lot of time ensuring correct security. But for me, as I still need `MetadataResolver`, I will be getting that to do it for me (Shibboleth have spent even more time getting the security correct) rather than use your code.

docs/modules/ROOT/pages/servlet/saml2/login/overview.adoc
```diff
@@ -639,6 +639,60 @@ class MyCustomSecurityConfiguration  {
...
+[source,java,role="primary"]
+----
+@Configuration
+@EnableWebSecurity
+public class MyCustomSecurityConfiguration {
+	@Bean
+    RelyingPartyRegistrationsDecoder registrationsDecoder() {
+		PublicKey pub = lookupAssertingPartyPublicKey();
+		Set<Credential> credentials = Set.of(new BasicCredential(pub));
+		return new OpenSamlRelyingPartyRegistrationsDecoder(credentials);
```
OrangeDog, line 659: This constructor does not appear to exist.

OrangeDog:
I've tested what I'm using (see above) and it's definitely an improvement 👍 . Please make `OpenSamlRelyingPartyRegistrationsDecoder.convert(EntityDescriptor)` public though.

For the signing, perhaps it should always use the filter, but with an empty/system trust store if none is provided. Then when decoding, set `requireSignedRoot` based on the URI scheme, or provide some customisation for what is trusted (with a safe default).

----

> With this change, applications can specify query parameters in the `authenticationRequestUri` value through a new method: `authenticationRequestUriQuery`:

Rather than parse a query string template, perhaps a method that just takes the parameter name would be better?

```
http
    .saml2Login((saml2) -> saml2
        .authenticationRequestUri("/saml/login")
        .authenticationRequestUriParam("peerEntityId")
    )
```

or 

```
http
    .saml2Login((saml2) -> saml2
        .authenticationRequestUri((uri) -> uri
            .path("/saml/login")
            .queryParam("peerEntityId")
        )
    )
```

jzheaux:
> But for me, as I still need MetadataResolver, I will be getting that to do it for me (Shibboleth have spent even more time getting the security correct) rather than use your code.
> ...
>  Please make OpenSamlRelyingPartyRegistrationsDecoder.convert(EntityDescriptor) public though.

I think what may be better than this is enhancing `OpenSamlAssertingPartyDetails#withEntityDescriptor` to do the work of preparing the asserting party details instead of the decoder doing that. Then I think you'd not use the decoder at all.

OrangeDog:
> I think what may be better than this is enhancing `OpenSamlAssertingPartyDetails#withEntityDescriptor` to do the work of preparing the asserting party details instead of the decoder doing that. Then I think you'd not use the decoder at all.

I don't care where that code is, as long as I can get at it without reflection hacks during my `EntityDescriptor` -> `RelyingPartyRegistration` conversion in my `findUniqueByAssertingPartyEntityId` adaptor.

jzheaux:
Thanks for all the feedback and insight, @OrangeDog. I'm enjoying this collaboration with you.

Specifically, your comments about `MetadataResolver` inspired a different design that is hopefully more useful in the way of creating a refreshing, expiry-aware repository.

> Rather than parse a query string template, perhaps a method that just takes the parameter name would be better?

I'd like to keep the placeholder there as I think a nice feature one day would be to allow for other placeholders that are already supported in other parts of Spring Security like `{relyingPartyEntityId}` and `{assertingPartyEntityId}`. Given that, I think there is less cognitive load for the user to supply the query as a string, though I'd be open to doing something like:

```java
.authenticationRequestUri("/saml2/authenticate", "registrationId={registrationId}")
```

Such just didn't seem to buy very much since I'm already parsing the `=`.

> The default configuration needs to be secure (and the current lack of security should have a CVE filed). If untrusted metadata is provided (no signature from a remote source, no trust material available) it needs to fail.

I'd prefer to leave the defaults as they are, which means that if no credentials are provided, then no signature verification is performed. This is also what `MetadataResolver` in OpenSAML does by default.

I disagree that this is a CVE on the grounds that HTTPS is the standard for secure endpoints across Spring Security. When deviating from that standard, Security tries to make the secure option the easy one. But going further than that, we can easily get in the way -- for example. it's not true that every HTTP endpoint requires trust material as `http://localhost/` does not.

jzheaux:
Also, in the meantime, I've addressed #15090 in the way [we already discussed](https://github.com/spring-projects/spring-security/pull/15349#issuecomment-2214757347) and committed that to `main`.

OrangeDog:
> a different design that is hopefully more useful

Very interesting. I'll have a deeper look at that (and probably edit this before you wake up).

> HTTPS is the standard for secure endpoints

Not in SAML. That's why everything has XML-level signatures and optional encryption.

> it's not true that every HTTP endpoint requires trust material

That may be true, but disabling security should be a conscious choice by the developer, not the default.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/OpenSamlAssertingPartyMetadataRepository.java
```diff
@@ -0,0 +1,297 @@
...
+
+			@Override
+			public AssertingPartyMetadata next() {
+				return OpenSamlAssertingPartyDetails.withEntityDescriptor(descriptors.next()).build();
+			}
+		};
+	}
+
+	@Override
+	public AssertingPartyMetadata findByEntityId(String entityId) {
```
OrangeDog, line 99: This should have `@Nullable` to match the interface definition.
Then it should return `null` instead of throwing `NoSuchElementException`.

I used `resolveSingle` instead of `resolve` to make that simpler.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
```diff
@@ -338,6 +338,25 @@ public static Builder withAssertingPartyDetails(AssertingPartyDetails assertingP
...
+	 * Presented as a convenience method when working with
+	 * {@link AssertingPartyMetadataRepository} return values. As such, only supports
+	 * {@link AssertingPartyMetadata} instances of type {@link AssertingPartyDetails}.
+	 * @param metadata the metadata used to initialize the
+	 * {@link RelyingPartyRegistration} {@link Builder}
+	 * @return {@link Builder} to create a {@link RelyingPartyRegistration} object
+	 * @since 6.4
+	 */
+	public static Builder withAssertingPartyMetadata(AssertingPartyMetadata metadata) {
+		Assert.isInstanceOf(AssertingPartyDetails.class, "metadata must be of type AssertingPartyDetails");
```
OrangeDog, line 356: This is checking the class of the message string, which always fails.

OrangeDog (changes requested):

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/OpenSamlAssertingPartyMetadataRepository.java
```diff
@@ -0,0 +1,297 @@
...
+
+	public OpenSamlAssertingPartyMetadataRepository(MetadataResolver metadataResolver) {
+		this.metadataResolver = metadataResolver;
+	}
+
+	@Override
+	@NonNull
+	public Iterator<AssertingPartyMetadata> iterator() {
+		CriteriaSet all = new CriteriaSet(new EntityRoleCriterion(IDPSSODescriptor.DEFAULT_ELEMENT_NAME));
+		Iterator<EntityDescriptor> descriptors = resolve(all).iterator();
```
OrangeDog, line 84: Whether this works depends on how the resolver is set up. If there's no index for `EntityRoleCriterion` then it will return nothing. Further, some resolvers e.g. `ChainingMetadataResolver` will not return all possible matches.

It can work with `EvaluableEntityRoleEntityDescriptorCriterion`, but only if the resolver has `resolveViaPredicatesOnly` set to `true`. 

You also cannot rely on `IterableMetadataSource` because not all classes implement that. You can special-case some classes to iterate over their delegates, but again not all of those may be iterable.

Basically, there's no way to implement this without restricting the configuration of resolvers that can be used. Looking for `IterableMetadataSource` is probably the better way.

OrangeDog: e.g.
```java
Stream<EntityDescriptor> allEntities(MetadataResolver resolver) {
    if (resolver instanceof IterableMetadataSource iterable) {
        return StreamSupport.stream(iterable.spliterator(), false);
    } else if (resolver instanceof ChainingMetadataResolver chaining) {
        return chaining.getResolvers().stream().flatMap(this::allEntities);
    } else if (resolver instanceof CompositeMetadataResolver composite) {
        return composite.getResolvers().stream().flatMap(this::allEntities);
    } else {
        return Stream.empty();
    }
}
```

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/registration/RelyingPartyRegistration.java
jzheaux: Because the AssertingPartyMetadataRepository returns an interface, I'd like to move `RelyingPartyRegistration` towards using it as well. Good feedback though, I've updated the PR so that it doesn't need to do the `instanceof` check.

OrangeDog:
I think this needs more test coverage, in particular the parts that didn't work that I pointed out above.

PR #15350, opened 2024-07-02
Fixing URL on README

mateusscheper:
Changing URL from https://docs.spring.io/spring-security/site/docs/current/reference/htmlsingle/ to https://docs.spring.io/spring-security/reference/ since the first one doesn't exist anymore.


mateusscheper:
By the way, why don't we have HTML and PDF documentation anymore?

jzheaux:
Hi, @mateusscheper, thanks for the fix. This is now merged into  `main`.

> By the way, why don't we have HTML and PDF documentation anymore?

Because it was rarely used and poorly maintained. We may add PDF support at some point using https://gitlab.com/antora/antora-assembler/-/blob/main/README.adoc?ref_type=heads#user-content-overview but it isn't scheduled.

Issue #14717, opened 2024-03-11
Support Externalized Method Security Configuration


rwinch:

evgeniycheban:
Hi, @rwinch can you assign this to me? I would like to implement this externalized method security configuration approach.

PR #15355, opened 2024-07-04
Method Security Exclude ExceptionHandler annotation method

kse-music:
Closes gh-15507


jzheaux:
Let's leave this out for now, @kse-music. I think we should consider a more general solution like https://github.com/spring-projects/spring-security/issues/14717 as that is a more expressive and cross-cutting way to customize how Spring Security annotations are applied.

For the time being, I think folks should edit the `Pointcut`, as outlined in [the related ticket](https://github.com/spring-projects/spring-security/issues/15507#issuecomment-2359091682).

Issue #15354, opened 2024-07-03
OAuth2AuthorizationRequest is not extendable 


aranoverse:
* **
# OAuth2AuthorizationRequest is not extendable 

**To Reproduce**
  the following is declaration of [AuthorizationRequestRepository](https://github.com/spring-projects/spring-security/blob/773e86701edcd6f2179ad66204f55e24c3ae3d51/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/AuthorizationRequestRepository.java#L40) <T extends OAuth2AuthorizationRequest> and OAuth2AuthorizationRequest is final class. 
 it means that we can  not create new CustomOAuth2AuthorizationRequest which  extends OAuth2AuthorizationRequest 

```java
  public interface AuthorizationRequestRepository <T extends OAuth2AuthorizationRequest> 
```
 
**Expected behavior** 

[OAuth2AuthorizationRequest ](https://github.com/spring-projects/spring-security/blob/773e86701edcd6f2179ad66204f55e24c3ae3d51/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java#L54) should not be final class

sjohnr:
Closing per [this comment](https://github.com/spring-projects/spring-security/pull/15361#issuecomment-2228914044).

PR #15361, opened 2024-07-04
Make OAuth2AuthorizationRequest extendable

aranoverse:
OAuth2AuthorizationRequest should not be final class [the issue link](https://github.com/spring-projects/spring-security/issues/15354)

### DIFF
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationRequest.java
@@ -51,7 +51,7 @@ import org.springframework.web.util.UriUtils;
  * "https://tools.ietf.org/html/rfc6749#section-4.1.1">Section 4.1.1 Authorization Code
  * Grant Request</a>
  */
-public final class OAuth2AuthorizationRequest implements Serializable {
+public class OAuth2AuthorizationRequest implements Serializable {
 
 	private static final long serialVersionUID = SpringSecurityCoreVersion.SERIAL_VERSION_UID;
 


jzheaux:
Thanks, @aranoverse. Can you explain more about what you are wanting to do in your application? I ask because it's worth leaving things `final` until opening them is needed. It's also worth considering whether the `Builder` class should also be extendible.

aranoverse:
> Thanks, @aranoverse. Can you explain more about what you are wanting to do in your application? I ask because it's worth leaving things `final` until opening them is needed. It's also worth considering whether the `Builder` class should also be extendible.

Thank you for noticing this issue. When using the `AuthorizationRequestRepository`, I want to store some additional data fields. I found that the declaration of `AuthorizationRequestRepository` has `T extends OAuth2AuthorizationRequest`, but I discovered that `OAuth2AuthorizationRequest` is not extendable.

Although I can work around this problem using `attributes`, from a design perspective, I think `T extends OAuth2AuthorizationRequest which is  a not extendable class` is not correct.

sjohnr:
Hi @aranoverse, thanks for your explanation!

> Although I can work around this problem using `attributes`, from a design perspective, I don't think `T extends OAuth2AuthorizationRequest` which is a not extendable class is correct.

You are correct that this does not seem logically consistent. I have spoken with @jgrandja to confirm that in fact the original idea was to leave the interface flexible in case subclasses of `OAuth2AuthorizationRequest` were required. Because of `attributes` and `additionalParameters`, subclasses have not been needed since this class was introduced in 5.0. Therefore, there are no plans to introduce a subclass and it does not seem necessary to make the class non-`final`.

> When using the `AuthorizationRequestRepository`, I want to store some additional data fields.

The field `additionalParameters` is ideal for this situation. I believe you use that instead of `attributes`. For now, I'm going to close this PR as it does not seem necessary to open up the class for extension. I will also close the corresponding issue as well.

However, you are correct that the interface `AuthorizationRequestRepository` does not require generics. We can open a separate issue to address this in 7.0.

If you feel I have misunderstood anything, please let me know.

aranoverse:
> Hi @aranoverse, thanks for your explanation!
> 
> > Although I can work around this problem using `attributes`, from a design perspective, I don't think `T extends OAuth2AuthorizationRequest` which is a not extendable class is correct.
> 
> You are correct that this does not seem logically consistent. I have spoken with @jgrandja to confirm that in fact the original idea was to leave the interface flexible in case subclasses of `OAuth2AuthorizationRequest` were required. Because of `attributes` and `additionalParameters`, subclasses have not been needed since this class was introduced in 5.0. Therefore, there are no plans to introduce a subclass and it does not seem necessary to make the class non-`final`.
> 
> > When using the `AuthorizationRequestRepository`, I want to store some additional data fields.
> 
> The field `additionalParameters` is ideal for this situation. I believe you can use that instead of `attributes`. For now, I'm going to close this PR as it does not seem necessary to open up the class for extension. I will also close the corresponding issue as well.
> 
> However, you are correct that the interface `AuthorizationRequestRepository` does not require generics. We can open a separate issue to address this in 7.0.
> 
> If you feel I have misunderstood anything, please let me know.

Thanks 🍻  !

PR #15362, opened 2024-07-04
Update the OAuth2 jwt and opaque Resource Server documentation

arey:
Prefer usage of the Lambda DSL.
The `OAuth2ResourceServerConfigurer::opaqueToken` and `OAuth2ResourceServerConfigurer::jwt` methods are deprecated since Spring Security 6.1.
With Spring Security 6.3, it looks like both methods are not working.

### DIFF
--- a/docs/modules/ROOT/pages/servlet/oauth2/resource-server/jwt.adoc
+++ b/docs/modules/ROOT/pages/servlet/oauth2/resource-server/jwt.adoc
@@ -879,7 +879,9 @@ public class DirectlyConfiguredJwkSetUri {
                 .requestMatchers("/messages/**").access(hasScope("messages"))
                 .anyRequest().authenticated()
             )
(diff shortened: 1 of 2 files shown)


jzheaux:
Thanks, @arey, I think this is a good idea.

Can you clarify what you mean by "it looks like both methods are not working"? What is it that isn't working for you?

arey:
Hi @jzheaux
Please disregard my last comment : `With Spring Security 6.3, it looks like both methods are not working.`
Instead of displaying a warning following the use of the deprecated method, IntelliJ displays an error. But the code compiles fine.
Saying that, I'd like to see the target method referenced in the documentation.

jzheaux:
Thanks, @arey! This is now merged into `main`.

arey:
You're welcome. Thank you for all your work on Spring Security @jzheaux

PR #15379, opened 2024-07-08
Add @FunctionalInterface to AuthenticationManager

pongdangx2:
AuthgenticationManager Interface is being implemented by using lambda expression.
Therefore, It would be clear not to add more abstract function, If @FunctionalInterface annotation is added.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java
@@ -315,7 +315,7 @@ public abstract class AbstractRequestMatcherRegistry<C> {
 		}
 	}
 
-	private String computeErrorMessage(Collection<? extends ServletRegistration> registrations) {
+	private static String computeErrorMessage(Collection<? extends ServletRegistration> registrations) {
 		String template = "This method cannot decide whether these patterns are Spring MVC patterns or not. "
 				+ "If this endpoint is a Spring MVC endpoint, please use requestMatchers(MvcRequestMatcher); "
 				+ "otherwise, please use requestMatchers(AntPathRequestMatcher).\n\n"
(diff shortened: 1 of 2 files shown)


jzheaux:
Thanks, @pongdangx2. Will you do me a favor and leave the copyright as-is? The Ageci copyright header is a special case.

pongdangx2:
Thank you, @jzheaux  I just restore the copyright line :)

jzheaux:
@pongdangx2 will you please also squash your commits? This makes backporting simpler.

It would also be nice if the commit title were capitalized, like so:

```
Add @FunctionalInterface to AuthenticationManager
```

pongdangx2:
@jzheaux 
I rebase my commits. And another commit is created, that I didn't make. (commit 06a9227e4e021b1667c3add721a6aa68b2487b34)

Could you please see if I should do something?? or I did right..?

jzheaux:
For the rebase to work, @pongdangx2, you need to rebase to the most recent version of `main`.

It's easier to do this if you have a branch with your changes. So, first, I'd recommend doing the following:

```
git checkout -b authentication-manager-functional-interface
git checkout main
```

Now you have your changes saved to a separate branch and are back on `main`. Then, you will want to make your `main` up to date with the Spring Security upstream. For the following, imagine that the name you use locally for your fork is `origin` and the same of the Spring Security upstream is called `spring-projects`:

```
git fetch spring-projects
git reset --hard spring-projects/main
git push origin main
```

Now `main` on your fork should be the same as Spring Security. You can verify this by doing:

```
$> git status
...
Your branch is up to date with 'spring-projects/main'.
```

Having done this, you can rebase your branch and push it by doing:

```
git checkout authentication-manager-functional-interface
git rebase main
git push origin authentication-manager-functional-interface
```

Finally, close this PR and create a new one based on your branch.

While several steps, this approach will save you time later on as you will create a branch the moment you start a new feature or bug fix.

PR #15256, opened 2024-06-15
Refactoring Test Code to Reduce Duplication of Mock Object Creation for Issue#14768

gzhao9:
Hi there,

This PR is to address [issue #14768](https://github.com/spring-projects/spring-security/issues/14768), which involves reducing repeated mock object creation in tests. 

Although my issue submission mentioned 4 separate draft PRs, I combined them into this PR, since they all address the same problem. 

If you prefer these changes to be split into separate PRs, please let me know, and I will gladly close this PR and submit individual ones.

Issue #14768, opened 2024-03-17
Refactoring Test Code to Reduce Duplication of Mock Object Creation


gzhao9:
Dear Spring Security Developers,

I have been reviewing your test code and noticed an issue with repeated creation of mock objects in your tests. Here are four examples where this occurs:

## Example 1

This example is found in [ActiveDirectoryLdapAuthenticationProviderTests](https://github.com/spring-projects/spring-security/blob/0078462e410eefbe27b3550f88fba058231f45bb/ldap/src/test/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProviderTests.java#L4).
The repeated code fragment is as follows:
```java
DirContext ctx = mock(DirContext.class);
given(ctx.getNameInNamespace()).willReturn("");
```
The issue with this code is that `willReturn("")` is unnecessary because Mockito defaults the return of `String` methods to `""`. This piece of code is used 18 times across 8 test cases.
Thus, I propose the following solution:

1. Remove `given(ctx.getNameInNamespace()).willReturn("");`
2. Move `DirContext ctx = mock(DirContext.class);` to `@BeforeEach`

This change reduces the code by 16 lines. I have created a [Draft PR](https://github.com/gzhao9/spring-security/pull/1), and you can view the changes [here](https://github.com/gzhao9/spring-security/pull/1/files).

## Example 2

This code is repeated in 8 test cases across 3 test classes:

- [AuthorizationManagerAfterMethodInterceptorTests](https://github.com/spring-projects/spring-security/blob/0078462e410eefbe27b3550f88fba058231f45bb/core/src/test/java/org/springframework/security/authorization/method/AuthorizationManagerAfterMethodInterceptorTests.java)

- [AuthorizationManagerBeforeMethodInterceptorTests](https://github.com/spring-projects/spring-security/blob/0078462e410eefbe27b3550f88fba058231f45bb/core/src/test/java/org/springframework/security/authorization/method/AuthorizationManagerBeforeMethodInterceptorTests.java)

- [PostFilterAuthorizationMethodInterceptorTests](https://github.com/spring-projects/spring-security/blob/0078462e410eefbe27b3550f88fba058231f45bb/core/src/test/java/org/springframework/security/authorization/method/PostFilterAuthorizationMethodInterceptorTests.java)

The repeated code is as follows:
```java
SecurityContextHolderStrategy strategy = mock(SecurityContextHolderStrategy.class);
// other statements
[...]

PR #15396, opened 2024-07-11
Factor out some common object mocking in tests

gzhao9:
Hi there,

This PR is to address [issue #14768](https://github.com/spring-projects/spring-security/issues/14768), which involves reducing repeated mock object creation in tests. 

I mistakenly closed PR #15256 and then discovered that I do not have permission to reopen it, receiving the message `this branch was force pushed or recreated.`  I'm not sure if I should reopen the original or create this new PR 

Although my issue submission mentioned 4 separate draft PRs, I combined them into this PR, since they all address the same problem.

### DIFF
--- a/ldap/src/test/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProviderTests.java
+++ b/ldap/src/test/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProviderTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -59,6 +59,7 @@ import static org.mockito.Mockito.verify;
 /**
  * @author Luke Taylor
  * @author Rob Winch
+ * @author Gengwu Zhao
  */
 public class ActiveDirectoryLdapAuthenticationProviderTests {
 
@@ -70,9 +71,13 @@ public class ActiveDirectoryLdapAuthenticationProviderTests {
 
 	UsernamePasswordAuthenticationToken joe = UsernamePasswordAuthenticationToken.unauthenticated("joe", "password");
 
+	DirContext ctx;
+
 	@BeforeEach
-	public void setUp() {
+	public void setUp() throws NamingException {
 		this.provider = new ActiveDirectoryLdapAuthenticationProvider("mydomain.eu", "ldap://192.168.1.200/");
+		this.ctx = mock(DirContext.class);
+		given(this.ctx.getNameInNamespace()).willReturn("");
 	}
 
 	@Test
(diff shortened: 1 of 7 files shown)


gzhao9:
@jesperronn Please feel free to let me know if there is anything I need to change. I look forward to your suggestions.

jzheaux (changes requested):
Thanks for the idea, @gzhao9. I think there is already a `MockSecurityContextHolderStrategy` available in `spring-security-core`. Can you please enhance that instead?

I think it would be reasonable to add the default constructor and one that takes a `SecurityContext` to that class.

Please also run `./gradlew format && ./gradlew check` before committing next. It appears there are some formatting issues that those two tasks will help you work through.

When you are done, will you please change your commits to be more descriptive in the following way. For each commit, describe what you've done and include the issue number like so:

```
Use MockSecurityContextHolderStrategy

Issue gh-14768
```

For the final commit, please use `Closes gh-14768` instead as this integrates with our build system to close the issue automatically.

ldap/src/test/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProviderTests.java
```diff
@@ -90,8 +91,6 @@ public void successfulAuthenticationProducesExpectedAuthorities() throws Excepti
 	@Test
 	public void customSearchFilterIsUsedForSuccessfulAuthentication() throws Exception {
 		String customSearchFilter = "(&(objectClass=user)(sAMAccountName={0}))";
-		DirContext ctx = mock(DirContext.class);
-		given(ctx.getNameInNamespace()).willReturn("");
```
jzheaux, line 94: I'd like the tests to remain as similar as possible behaviorally. Will you then please add this to the setup method:

```java
given(ctx.getNameInNamespace()).willReturn("");
```

web/src/test/java/org/springframework/security/web/access/ExceptionTranslationFilterTests.java
```diff
@@ -304,7 +291,11 @@ public void setMessageSourceWhenNotNullThenCanGet() {
 		filter.messages.getMessage(code);
 		verify(source).getMessage(eq(code), any(), any());
 	}
-
+	private FilterChain mockFilterChainWiehException(Object exception) throws ServletException, IOException {
```
jzheaux, line 294: I believe there is a typo here and the method should be called `mockFilterChainWithException`.

jzheaux, line 294: I think this method can take `Exception` for the parameter type. Is there a reason it needs to be `Object`?

web/src/test/java/org/springframework/security/web/access/ExceptionTranslationFilterTests.java
```diff
@@ -304,7 +291,11 @@ public void setMessageSourceWhenNotNullThenCanGet() {
...
+		FilterChain fc = mock(FilterChain.class);
+		willThrow((Throwable) exception).given(fc).doFilter(any(HttpServletRequest.class), any(HttpServletResponse.class));
```
jzheaux, line 296: If you can change the parameter type, this cast should be removed.

jzheaux (changes requested):
Thank you for the PR, @gzhao9. I've left feedback inline.

Additionally, will you please consider whether you can follow the same strategy for all four files when it comes to setting up mocks? It seems that in one you have a separate class, in another you have a field and a `setup` method, and in others you use a local method. It would be nice if the code can be more consistent.

Finally, will you please ensure that your commits are in the following format:

```
Factor out common SecurityContextHolderStrategy mocking
```

or similar, depending on the work done for each file.
```

gzhao9:
> Thanks for the cleanup, @gzhao9. I think there is already a `MockSecurityContextHolderStrategy` available in `spring-security-core`. Can you please enhance that instead?
> 
> I think it would be reasonable to add the default constructor and one that takes a `SecurityContext` to that class.
> 
> Please also run `./gradlew format && ./gradlew check` before committing next. It appears there are some formatting issues that those two tasks will help you work through.
> 
> When you are done, will you please change your commits to be more descriptive in the following way. For each commit, describe what you've done and include the issue number like so:
> 
> ```
> Use MockSecurityContextHolderStrategy
> 
> Issue gh-14768
> ```
> 
> For the final commit, please use `Closes gh-14768` instead as this integrates with our build system to close the issue automatically.

I found `MockSecurityContextHolderStrategy` in `spring-security-core`. However, since they are not in the same package, I would need to change the dependency if I want to reference `MockSecurityContextHolderStrategy` in `org.springframework.security.authorization.method`. I believe adding new dependencies should be considered carefully. Therefore, I only added a private method in each test class.

gzhao9:
I first focus on each file code modification. I will rebase commits when there is no need to modify the code or format.

jzheaux:
Thanks for the updates, @gzhao9! I think there are two other items from the review that haven't been addressed yet. Are you able to do those as well?

gzhao9:
> Thanks for the updates, @gzhao9! I think there are two other items from the review that haven't been addressed yet. Are you able to do those as well?

I have modified the code as required.

jzheaux (approved):

jzheaux:
Thanks, @gzhao9! This is now merged into `main`.

PR #15405, opened 2024-07-12
Add support for "audience" parameter in Token Exchange grant

dalbani:
Following [the introduction of the support for Token Exchange in Spring Security 6.3](https://spring.io/blog/2024/03/19/token-exchange-support-in-spring-security-6-3-0-m3), this MR adds the ability to set the `audience` parameter in the request, as described in https://datatracker.ietf.org/doc/html/rfc8693#:~:text=%C2%B6-,audience,-OPTIONAL.%20The%20logical.

_Work in progress: unit tests need to be added._

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/TokenExchangeOAuth2AuthorizedClientProvider.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/TokenExchangeOAuth2AuthorizedClientProvider.java
@@ -49,6 +49,8 @@ public final class TokenExchangeOAuth2AuthorizedClientProvider implements OAuth2
 
 	private Function<OAuth2AuthorizationContext, OAuth2Token> actorTokenResolver = (context) -> null;
 
+	private Function<OAuth2AuthorizationContext, String> audienceResolver = (context) -> null;
+
 	private Duration clockSkew = Duration.ofSeconds(60);
 
 	private Clock clock = Clock.systemUTC();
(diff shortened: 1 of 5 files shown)


sjohnr:
@dalbani thanks so much for being willing to enhance the token exchange support in the framework!

However, keep in mind that it's usually best to start by opening an issue to discuss improvements prior to opening a pull request. This would save effort in the case where a different direction might be taken due to discussion, or where the request might be closed for some reason.

In this case, the `audience` parameter should not be included in the `OAuth2AuthorizedClientProvider` implementation because it is not a credential but is an optional parameter to be included with the access token request. You can provide a custom `parametersConverter` to the `TokenExchangeGrantRequestEntityConverter`, which would be the intended and preferred way to customize and provide an `audience` parameter. This is demonstrated in the blog post for that reason.

I'm going to close this PR with the above explanation. If you feel I have misunderstood anything, please let me know and we can discuss further.

dalbani:
Hi @sjohnr, thanks a lot for your feedback.
If I understand correctly, your point is that `audience` is an optional parameter, so it make sense to a custom parameters converter.
But then the spec says that `actor_token` is also optional. Why treat it differently than other optional parameters?
Why not make Spring Security support the full breadth of parameters as first class citizen, instead of requiring developers to write custom code?
Oh, I see you mentioned as well that `audience` was not a credential as well. But why is it so important in the decision to support it in `TokenExchangeGrantRequestEntityConverter` or not?

dalbani:
For the record, this feature is related to the use case I raised in https://github.com/spring-projects/spring-security/issues/15408.

sjohnr:
Hi @dalbani!

> But then the spec says that `actor_token` is also optional. Why treat it differently than other optional parameters?
> Why not make Spring Security support the full breadth of parameters as first class citizens, instead of requiring developers to write custom code?

The spec is fairly light on details and provides only two high level use cases: "impersonation" and "delegation". The credential required for impersonation is `subject_token` while the credentials required for delegation are `subject_token` and `actor_token`. While `actor_token` is optional overall, it is not optional for the delegation use case. All other parameters are optional and could be combined or omitted in various ways (including with other parameters not even specified in the spec) which I don't think we could predict.

To keep the `TokenExchangeGrantRequest` simple and straightforward, I feel that using the already existing and dedicated capability for customizing parameters is the natural place for optional parameters like `audience`. Also, now that this code is released and the API is public and being consumed by applications, I want to generally avoid changes when possible.

> Oh, I see you mentioned as well that `audience` was not a credential as well. But why is it so important in the decision to support it in `TokenExchangeGrantRequestEntityConverter` or not?

Hopefully this is answered by the above comment. The goal is to make configuration of all grant types including token exchange extremely consistent with each other as well as simple and intuitive to use.

Issue #15338, opened 2024-07-01
NPE in DefaultOAuth2User.getName function


Eccenux:
**Describe the bug**
Problem is a possible `NullPointerException` (NPE) in `DefaultOAuth2User.getName`.

**To Reproduce**
Get null value on the name attribute during after OAuth authentaction.
Or simply create attributes for `DefaultOAuth2User` as show in the _Sample_.

**Expected behavior**
NPE shouldn't be going around Spring Security. It should maybe use `throw new UsernameNotFoundException(.)`, not sure.

**Sample**

This is a minimal example to reproduce (without setting up an oauth2 server):
```
		Collection<GrantedAuthority> authorities = Arrays.asList(
			new SimpleGrantedAuthority("ROLE_USER")
		);
		var attributes = new HashMap<String, Object>();
		var nameAttributeKey = "Email";
		attributes.put("UserName", "");
		// notice this is in the attributes, but is null
		attributes.put(nameAttributeKey, null);
		var test = new DefaultOAuth2User(authorities, attributes, nameAttributeKey);
		test.getName();
```

This code is a problem:
```
	@Override
	public String getName() {
		return this.getAttribute(this.nameAttributeKey).toString();
	}
```

**Real code**
In a real use case this was the attributes of the DefaultOAuth2User:
> User Attributes: [{UserName=, Email=null, Authenticated=false, Claims=null}]

The name field was "Email", but I guess it could be anything that just so happens to be null. Note that attributes are from an external service. The service is defined by `userInfoUri` from `FactoryBean<T>`. So the attributes are essentially attributes of a user profile as returned by the identity provider (idP).

This is how getName was originally called:
```
public class OAuth2UserServiceImpl extends DefaultOAuth2UserService {
	private static final Logger LOG = LoggerFactory.getLogger(OAuth2UserServiceImpl.class);
	// ...
	@Override
	public OAuth2User loadUser(OAuth2UserRequest userRequest) throws OAuth2AuthenticationException {
		OAuth2User oauth2user = super.loadUser(userRequest);

		final String providerId = userRequest.getClientRegistration().getRegistrationId();
		// this results in NPE
		final String username = oauth2user.getName();
		// ...
}
```

jzheaux:
Thanks, @Eccenux, for the report. It sounds like the reason there is an NPE is because the application is placing a `null` value into the map for the `nameAttributeKey`. Does that sound right?

[...]

PR #15407, opened 2024-07-12
Fix NPE when nameAttributeValue is null (#15338)

HyoJongPark:
This commit resolves NPEs that occur in DefaultOuth2User#getName.
If the attributes have a key, but the value comes in null, the NPE occurs, and we modified the case and added a related test case.

fixes: #15338

### DIFF
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/user/DefaultOAuth2User.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/user/DefaultOAuth2User.java
@@ -69,7 +69,8 @@ public class DefaultOAuth2User implements OAuth2User, Serializable {
 			String nameAttributeKey) {
 		Assert.notEmpty(attributes, "attributes cannot be empty");
 		Assert.hasText(nameAttributeKey, "nameAttributeKey cannot be empty");
-		Assert.notNull(attributes.get(nameAttributeKey), "Attribute value for '" + nameAttributeKey + "' cannot be null");
+		Assert.notNull(attributes.get(nameAttributeKey),
+				"Attribute value for '" + nameAttributeKey + "' cannot be null");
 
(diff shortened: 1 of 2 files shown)


oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/user/DefaultOAuth2User.java
```diff
@@ -69,8 +70,10 @@ public DefaultOAuth2User(Collection<? extends GrantedAuthority> authorities, Map
 		Assert.notEmpty(attributes, "attributes cannot be empty");
 		Assert.hasText(nameAttributeKey, "nameAttributeKey cannot be empty");
 		if (!attributes.containsKey(nameAttributeKey)) {
-			throw new IllegalArgumentException("Missing attribute '" + nameAttributeKey + "' in attributes");
+			throw new IllegalArgumentException("Missing nameAttributeKey '" + nameAttributeKey + "' in attributes");
 		}
+		Assert.notNull(attributes.get(nameAttributeKey), "Name attribute '" + nameAttributeKey + "' cannot be null");
```
jzheaux, line 75: I think `Map#get` returing `null` implies that `Map#containsKey` returns `false`.

Would you please then simply replace the `containsKey` check with the `nonNull` check?

Please change the message also to:

```
Assert.notNull(attributes.get(nameAttributeKey), "Attribute value for '" + nameAttributeKey + "' cannot be null");
```

jzheaux (changes requested):
Thanks! I've left some feedback inline.

Also, when you are ready, will you please squash your commits so that there is only one? Please format that final commit like so:

```
Check for null Name Attribute Value

Closes gh-15338
```

HyoJongPark:
All review has been resolved and reformat commit.!

jzheaux:
Thanks, @HyoJongPark! This is now merged into `main`.

PR #15409, opened 2024-07-12
Clarify url Parameter Usage in AD Provider Constructor

### LINKED ISSUE #7760: ActiveDirectoryLdapAuthenticationProvider does not implement support for multiple urls
### Summary

The javadoc for `ActiveDirectoryLdapAuthenticationProvider` constructors says the param `url` supports multiple URLs. The javadoc however does not define how the multiple URLs needs to be supplied (i.e delimiter specification). Upon trying to supply this param as multiple URLs as a comma or pipe separated list this string is used as is during the ldap binding

### Actual Behavior

By looking at the code, specifically the constructors of `ActiveDirectoryLdapAuthenticationProvider` and method `bindAsUser` it uses the url provided in the constructor as is without checking if multiple URLs are present. Due to this the the line `env.put(Context.PROVIDER_URL, bindUrl);` in `bindAsUser` method will inject the list of URLs without splitting in the env which will be incorrect and the ldap bind does not work.

### Expected Behavior
The code should check if the supplied param url is a delimited list or URLs and then store it as such. During the authentication if an ldap server is unavailable the remaining URLs should be tried. The javadoc/code should also make it clear that if multiple URLs are to be supplied what format they should be in or there should be an additional constructor/setter that accepts list instead of a string. 

### Configuration

    @Configuration
    public class WebSecurityConfig extends WebSecurityConfigurerAdapter {
 
    @Value("${ldap.url:ldap://localhost:389}") private String url;
    @Value("${ldap.domain:domain}") private String domain;
 
    @Override
    protected void configure(HttpSecurity http) throws Exception
    {
        http.authorizeRequests()
              .anyRequest().fullyAuthenticated()
              .httpBasic();
    }
 
    @Override
    public void configure(AuthenticationManagerBuilder auth) throws Exception
    {
        ActiveDirectoryLdapAuthenticationProvider adProvider = 
                    new ActiveDirectoryLdapAuthenticationProvider(domain,url);
        adProvider.setConvertSubErrorCodesToExceptions(true);
        adProvider.setUseAuthenticationRequestCredentials(true);
        auth.authenticationProvider(adProvider);
        auth.eraseCredentials(false);
     }
    }

### Version
spring-security-ldap 4.2.10 and 5.1.5.RELEASE

Haarolean:
Updated javadocs for `ActiveDirectoryLdapAuthenticationProvider` as discussed in #7760. Closes #7760.

### DIFF
--- a/ldap/src/main/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProvider.java
+++ b/ldap/src/main/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProvider.java
@@ -136,9 +136,11 @@ public final class ActiveDirectoryLdapAuthenticationProvider extends AbstractLda
 	ContextFactory contextFactory = new ContextFactory();
 
 	/**
-	 * @param domain the domain name (may be null or empty)
-	 * @param url an LDAP url (or multiple URLs)
-	 * @param rootDn the root DN (may be null or empty)
+	 * @param domain the domain name (can be null or empty)
+	 * @param url an LDAP url (or multiple space-delimited URLs).
+	 * @param rootDn the root DN (can be null or empty)
+	 * @see <a href="https://docs.oracle.com/javase/jndi/tutorial/ldap/misc/url.html">JNDI
+	 * URL format documentation</a>
 	 */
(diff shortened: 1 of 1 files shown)


jzheaux:
Thanks, @Haarolean! This is now merged into `5.8.x`, `6.2.x`, `6.3.x`, and `main`.

Issue #13588, opened 2023-07-27
Add support for requesting protected resources with RestClient


mjeffrey:
**Expected Behavior**
Allow the use RestClient (to be introduced in Spring 6.1) for blocking calls in a non reactive application In Oauth2 Client.
See https://spring.io/blog/2023/07/13/new-in-spring-6-1-restclient.

**Current Behavior**
Only WebClient is supported which means a lot of reactive dependencies are pulled in when using Oauth2 Client even in a blocking application. 

**Context**
To avoid the dependency issue we are using RestTemplate (with interceptors) but would prefer to see a supported solution.

sjohnr:
@mjeffrey thanks for starting the conversation on this!

To clarify, are you specifically directing this issue toward providing support for `RestClient` in OAuth2 equivalent to the existing [`ServletOAuth2AuthorizedClientExchangeFilterFunction`](https://docs.spring.io/spring-security/reference/servlet/oauth2/client/authorized-clients.html#oauth2Client-webclient-servlet) which supports `WebClient`?

mjeffrey:
@sjohnr 
I don't want to ask for too much related to implementation, you guys know this much better than me but I think what you said would do what is needed for us. 

What I'd like to achieve is to be able to use the new RestClient with the Oauth2 client and to not need the reactive libraries.

In my company we are in the process of migrating a number of Spring Boot 2 (keycloak Oauth2 client) to Spring Boot 3 projects.
The Keycloak client no longer supports Spring Boot 3 and so we are moving to Spring Security Oauth2 client.
However the requirement to use WebClient is one of the things holding some teams back.

sjohnr:
Ok, I've changed the title of the issue to what I think aligns with the request as a starting point, and we can go from there. Since the only part of OAuth2 Client for a Servlet environment that directly uses reactive libraries is `ServletOAuth2AuthorizedClientExchangeFilterFunction`, this issue can be used to provide equivalent support.

alexcrownus:
I was about to create a similar issue, glad someone already did. Many thanks @mjeffrey. 

Since support for RestTemplate was dropped because it is now in maintenance mode, there should now be support for the new RestClient for Servlet applications without using WebClient in blocking mode.

tudburley:
[...]

PR #15437, opened 2024-07-17
Add `RestClient` interceptor

sjohnr:
To Do:

- [x] Add unit tests
- [ ] Add integration tests
- [x] Add reference documentation
- [x] Add sample(s)

Closes gh-13588

### DIFF
--- a/docs/modules/ROOT/pages/servlet/oauth2/index.adoc
+++ b/docs/modules/ROOT/pages/servlet/oauth2/index.adoc
@@ -398,13 +398,14 @@ See xref:getting-spring-security.adoc[] for additional options when not using Sp
 
 Consider the following use cases for OAuth2 Client:
 
-* <<oauth2-client-log-users-in,I want to log users in using OAuth 2.0 or OpenID Connect 1.0>>
-* <<oauth2-client-access-protected-resources,I want to obtain an access token for users in order to access a third-party API>>
-* <<oauth2-client-access-protected-resources-current-user,I want to do both>> (log users in _and_ access a third-party API)
-* <<oauth2-client-enable-extension-grant-type,I want to enable an extension grant type>>
-* <<oauth2-client-customize-existing-grant-type,I want to customize an existing grant type>>
-* <<oauth2-client-customize-request-parameters,I want to customize token request parameters>>
-* <<oauth2-client-customize-rest-operations,I want to customize the `RestOperations` used by OAuth2 Client components>>
+* I want to <<oauth2-client-log-users-in,log users in using OAuth 2.0 or OpenID Connect 1.0>>
+* I want to <<oauth2-client-access-protected-resources,use `RestClient` to obtain an access token for users in order to access a third-party API>>
+* I want to <<oauth2-client-access-protected-resources-current-user,do both>> (log users in _and_ access a third-party API)
+* I want to <<oauth2-client-access-protected-resources-webclient,use `WebClient` to obtain an access token for users in order to access a third-party API>>
+* I want to <<oauth2-client-enable-extension-grant-type,enable an extension grant type>>
+* I want to <<oauth2-client-customize-existing-grant-type,customize an existing grant type>>
+* I want to <<oauth2-client-customize-request-parameters,customize token request parameters>>
+* I want to <<oauth2-client-customize-rest-operations,customize the `RestOperations` used by OAuth2 Client components>>
 
 [[oauth2-client-log-users-in]]
 === Log Users In with OAuth2
@@ -584,38 +585,11 @@ Spring Security provides implementations of `OAuth2AuthorizedClientManager` for 
 Spring Security registers a default `OAuth2AuthorizedClientManager` bean for you when one does not exist.
 ====
 
-The easiest way to use an `OAuth2AuthorizedClientManager` is via an `ExchangeFilterFunction` that intercepts requests through a `WebClient`.
-To use `WebClient`, you will need to add the `spring-webflux` dependency along with a reactive client implementation:
+The easiest way to use an `OAuth2AuthorizedClientManager` is via a `ClientHttpRequestInterceptor` that intercepts requests through a `RestClient`, which is already available when `spring-web` is on the classpath.
 
-.Add Spring WebFlux Dependency
-[tabs]
-======
-Gradle::
-+
-[source,gradle,role="primary"]
-----
-implementation 'org.springframework:spring-webflux'
-implementation 'io.projectreactor.netty:reactor-netty'
-----
+The following example uses the default `OAuth2AuthorizedClientManager` to configure a `RestClient` capable of accessing protected resources by placing `Bearer` tokens in the `Authorization` header of each request:
 
-Maven::
-+
-[source,maven,role="secondary"]
-----
-<dependency>
-	<groupId>org.springframework</groupId>
-	<artifactId>spring-webflux</artifactId>
-</dependency>
-<dependency>
-	<groupId>io.projectreactor.netty</groupId>
-	<artifactId>reactor-netty</artifactId>
-</dependency>
-----
-======
-
-The following example uses the default `OAuth2AuthorizedClientManager` to configure a `WebClient` capable of accessing protected resources by placing `Bearer` tokens in the `Authorization` header of each request:
-
-.Configure `WebClient` with `ExchangeFilterFunction`
+.Configure `RestClient` with `ClientHttpRequestInterceptor`
 [tabs]
 =====
 Java::
@@ -623,14 +597,14 @@ Java::
 [source,java,role="primary"]
 ----
 @Configuration
-public class WebClientConfig {
+public class RestClientConfig {
 
 	@Bean
-	public WebClient webClient(OAuth2AuthorizedClientManager authorizedClientManager) {
-		ServletOAuth2AuthorizedClientExchangeFilterFunction filter =
-				new ServletOAuth2AuthorizedClientExchangeFilterFunction(authorizedClientManager);
-		return WebClient.builder()
-				.apply(filter.oauth2Configuration())
+	public RestClient restClient(OAuth2AuthorizedClientManager authorizedClientManager) {
+		OAuth2ClientHttpRequestInterceptor requestInterceptor =
+				new OAuth2ClientHttpRequestInterceptor(authorizedClientManager);
+		return RestClient.builder()
+				.requestInterceptor(requestInterceptor)
 				.build();
 	}
 
@@ -642,13 +616,13 @@ Kotlin::
 [source,kotlin,role="secondary"]
 ----
 @Configuration
-class WebClientConfig {
+class RestClientConfig {
 
 	@Bean
-	fun webClient(authorizedClientManager: OAuth2AuthorizedClientManager): WebClient {
-		val filter = ServletOAuth2AuthorizedClientExchangeFilterFunction(authorizedClientManager)
-		return WebClient.builder()
-			.apply(filter.oauth2Configuration())
+	fun restClient(authorizedClientManager: OAuth2AuthorizedClientManager): RestClient {
+		val requestInterceptor = OAuth2ClientHttpRequestInterceptor(authorizedClientManager)
+		return RestClient.builder()
+			.requestInterceptor(requestInterceptor)
 			.build()
 	}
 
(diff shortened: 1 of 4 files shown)


azdanov:
Hey. If using an [HTTP Interface](https://docs.spring.io/spring-framework/reference/integration/rest-clients.html#rest-http-interface), how would it be possible to dynamically change `clientRegistrationId` since there is no access to the restClient after the proxy is created?

In webclient I could use [ServletOAuth2AuthorizedClientExchangeFilterFunction](https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.html):

```java
@Bean
public WebClient webClient(
        WebClient.Builder builder,
        HttpClient httpClient,
        OAuth2AuthorizedClientManager authorizedClientManager
) {
    ServletOAuth2AuthorizedClientExchangeFilterFunction oauth2 =
            new ServletOAuth2AuthorizedClientExchangeFilterFunction(authorizedClientManager);

    oauth2.setDefaultOAuth2AuthorizedClient(true);

    return builder
            .apply(oauth2.oauth2Configuration())
            .clientConnector(new ReactorClientHttpConnector(httpClient))
            .build();
}
```

sjohnr:
Thanks for asking, @azdanov. See the per-request arrangement in [this comment](https://github.com/spring-projects/spring-security/issues/13588#issuecomment-2123283036). This aspect of the interceptor is still in need of feedback so feel free to weigh in on it. The other option would be to use [`RestClient#attributes`](https://github.com/spring-projects/spring-framework/issues/32027) but I am not sure if using attributes to dynamically configure on a per-request basis is intuitive for users, so I'm not yet leaning towards using them.

However, I'm not clear how you would dynamically change the `clientRegistrationId` in your example with `WebClient` since I don't believe you can use `attributes` there. Using `setDefaultOAuth2AuthorizedClient(true)` would allow detecting the id from the currently logged in user (if they logged in using OAuth2 or OIDC), but you couldn't specify it dynamically. Can you please provide an example of how you would incorporate the `WebClient` from your example into an HTTP interface while dynamically specifying the `clientRegistrationId`?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/function/client/OAuth2ClientHttpRequestInterceptor.java
```diff
@@ -0,0 +1,462 @@
...
+		catch (RestClientResponseException ex) {
+			handleAuthorizationFailure(ex.getResponseHeaders(), ex.getStatusCode());
+			throw ex;
+		}
+		catch (OAuth2AuthorizationException ex) {
+			handleAuthorizationFailure(ex);
+			throw ex;
+		}
+	}
+
```
Lonelysoul-HayashiNoMirai, line 358: Hey bro, I've tried to use ur interceptor in my production code, then I got an exception: IllegalArgumentException: URI with undefined scheme.
After some debugging attempts, I realized that the baseUrl, which I set here:
`return RestClient.builder ()
                .requestFactory (requestFactory)
                .baseUrl (commandCenterBaseUrl)
                .requestInterceptor (requestInterceptor)
                .build ();`
...didn't get passed to the interceptor... well Ok, I added an additional setter for the baseUrl and then...
```suggestion
@Override
    public ClientHttpResponse intercept (
            HttpRequest request, byte[] body, ClientHttpRequestExecution execution
    ) throws IOException {

        if (!isNullOrEmpty (baseUrl)){
            request = prependBaseUrl (request);
        }
        authorizeClient (request);

        try {
            ClientHttpResponse response = execution.execute (request, body);
            handleAuthorizationFailure (response.getHeaders (), response.getStatusCode ());
            return response;
        }
        catch (RestClientResponseException ex){
            handleAuthorizationFailure (ex.getResponseHeaders (), ex.getStatusCode ());
            throw ex;
        }
        catch (OAuth2AuthorizationException ex){
            handleAuthorizationFailure (ex);
            throw ex;
        }
    }

    private HttpRequest prependBaseUrl (HttpRequest request){
        URI uri = request.getURI ();
        URI completeUri = fromHttpUrl (baseUrl)
                .path (uri.getPath ())
                .query (uri.getQuery ())
                .build (true)
                .toUri ();

        return new HttpRequestWrapper (request){

            @Override
            public @NonNull URI getURI (){
                return completeUri;
            }
        };
    }
```
I hope this suggestion helps!

azdanov:
>Using setDefaultOAuth2AuthorizedClient(true) would allow detecting the id from the currently logged in user

This is what I meant, thank you for clarifying things 🙂. That's what I considered dynamic, that the interceptor will do it, and not me.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/function/client/OAuth2ClientHttpRequestInterceptor.java
sjohnr: Thanks for the suggestion @Lonelysoul-HayashiNoMirai! However, I'm not understanding your issue. This class should not be responsible for setting the `baseUrl`. Can you help me understand your issue better by providing a sample? **Note:** I have a TODO to add samples, so this will help immensely.

sjohnr:
> This is what I meant, thank you for clarifying things 🙂. That's what I considered dynamic, that the interceptor will do it, and not me. It would be cool to see a `setDefaultOAuth2AuthorizedClient` option RestClient interceptor too.

Thanks @azdanov, I suspected so but didn't want to assume. I have not yet determined whether it is truly necessary to introduce this functionality into the interceptor. I consider it a convenience feature, and it creates more ways to use the interceptor which can confuse users as much as help them. That's something I'd like to avoid. However, this is good feedback so I will think about it some more.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/function/client/OAuth2ClientHttpRequestInterceptor.java
Lonelysoul-HayashiNoMirai: > Thanks for the suggestion @Lonelysoul-HayashiNoMirai! However, I'm not understanding your issue. This class should not be responsible for setting the `baseUrl`. Can you help me understand your issue better by providing a sample? **Note:** I have a TODO to add samples, so this will help immensely.

I know that this class has nothing to do with "setting the baseUrl", then I should set it in here, right?
image
Even after setting it in the rest client, I'm still getting this exception:
image
When I was debugging, just before the exception was thrown, I clearly saw that the URL was missing the "baseUrl":
Debug-Screenshot
With my modifications in the suggestion, everything is Ok again... that's all, bro

Lonelysoul-HayashiNoMirai:
@sjohnr 
Bro, I found another problem, when using ur interceptor with the @Async annotation, the method handleAuthorizationFailure () throws an exception [IllegalArgumentException: request cannot be null] if the current access token failed...

I've tried to debug it, and I guess the root cause of the problem is that ur code tried to get the RequestAttributes from the RequestContextHolder and then get the request from the RequestAttributes. Since @Async is using, there is nothing in the RequestContextHolder... => failed

~ Could you do me a favour and suggest me a way to fix it?

Request_Is_Null_Exception_Screen

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/function/client/OAuth2ClientHttpRequestInterceptor.java
sjohnr: @Lonelysoul-HayashiNoMirai thanks for reporting this issue and giving the interceptor a try. However, I am not able to reproduce your issue.

I have created a sample in a [branch](https://github.com/sjohnr/spring-security-samples/tree/gh-13588-rest-client) of the [Spring Security Samples](https://github.com/spring-projects/spring-security-samples) project that you can try ([code and instructions here](https://github.com/sjohnr/spring-security-samples/tree/gh-13588-rest-client/servlet/spring-boot/java/oauth2/httpclient)). Would you consider modifying that sample to reproduce the issue you're seeing and sharing it with me so I can investigate further?

sjohnr: I should mention that in order to use the sample above, you will need to build this branch of spring-security and install it in your local maven cache. You can do so via the command `./gradlew publishToMavenLocal`.

Lonelysoul-HayashiNoMirai: @sjohnr 
Thank bro for considering my feedback, really appreciate it...
However, I'm up to my neck in deadlines right now :(
Will check later if I have free time...

In the meantime, if u have some spare time, pls help me check this issue first (this may be the cause of the above issue): https://github.com/spring-projects/spring-security/pull/15437#issuecomment-2255630771

sjohnr:
@Lonelysoul-HayashiNoMirai 

> ~ Could you do me a favour and suggest me a way to fix it?

To use OAuth2 Client features with `@Async`, `TaskExecutor` or other background thread processing, you should use [`AuthorizedClientServiceOAuth2AuthorizedClientManager`](https://docs.spring.io/spring-security/reference/servlet/oauth2/client/core.html#oauth2Client-authorized-manager-provider), by registering a bean like so:

```java
@Bean
public OAuth2AuthorizedClientManager authorizedClientManager(
		ClientRegistrationRepository clientRegistrationRepository,
		OAuth2AuthorizedClientService authorizedClientService) {

	return new AuthorizedClientServiceOAuth2AuthorizedClientManager(
		clientRegistrationRepository, authorizedClientService);
}
```

I believe that will solve the issue, but note that this wouldn't be an issue with the interceptor but only with your configuration. If this doesn't solve the issue, please provide a minimal sample as I am not able to reproduce your issue without this.

sjohnr:
Just a note for those following along: Based on feedback, I've pushed changes to the PR (gh-15437) with an updated API for the interceptor that takes advantage of [`attributes`](https://github.com/spring-projects/spring-framework/issues/32027) for configuring the `clientRegistrationId` on a per-request basis. This effectively consolidates everything to a single primary usage example:

```java
@GetMapping("/cashcards")
public Cashcard[] cashcards() {
	OAuth2ClientHttpRequestInterceptor requestInterceptor =
		new OAuth2ClientHttpRequestInterceptor(this.authorizedClientManager, CLIENT_REGISTRATION_ID);

	// Configure the interceptor to remove invalid authorized clients
	requestInterceptor.setAuthorizedClientRepository(this.authorizedClientRepository);

	return this.restClient.get()
		.uri("http://localhost:8090/cashcards")
		.attributes(OAuth2ClientHttpRequestInterceptor.clientRegistrationId(CLIENT_REGISTRATION_ID))
		.retrieve()
		.body(Cashcard[].class);
}
```

**Note** that you can also call `requestInterceptor.setDefaultClientRegistrationId(...)`, or omit the `clientRegistrationId` entirely and it will be derived from the current user (requires `http.oauth2Login()`). 

cc: @azdanov @barbetb @mcordeiro73 @ThomasVitale @mjeffrey @ThomasKasene @rwinch @jgrandja

Lonelysoul-HayashiNoMirai:
@sjohnr
Finally I got some free time for u, sorry bro.
Obviously, I can't give u the production code, so here u go: I created 2 new projects to reproduce this issue:
[request-cannot-be-null-exception](https://github.com/spring-projects/spring-security/pull/15437#issuecomment-2255630771)

- Server: [OAuth2_Authorization_Resource_Server_Sample](https://github.com/Lonelysoul-HayashiNoMirai/OAuth2_Authorization_Resource_Server_Sample)
- Client: [OAuth2_Client_Sample](https://github.com/Lonelysoul-HayashiNoMirai/OAuth2_Client_Sample)

### Steps To Reproduce:
- Step 1: Run both projects.
- Step 2: Send request to the client app to trigger it send request to the server app:
   + curl --location --request POST 'http://localhost:8085/api/v1/rest-client:fireATestRequest'
- Step 3: Check the log of the client app, if u see "The response: protected-string", that means the interceptor successfully fetched the access token and then the client app is free to access the server's APIs.
- Step 4: Restart the server app to invalidate the token.
- Step 5: Send request to the client app (Step 2) again, exception will be thrown in the client app: IllegalArgumentException: request cannot be null

### Expected Behavior: The client app should discard the invalid token and fetch a new one.

sjohnr:
@Lonelysoul-HayashiNoMirai thank you again for trying the interceptor, and providing samples for the issue you encountered. There are a couple of things going on.

1. You have modified the interceptor with the `baseUrl` issue we discussed, which is not necessary. You are using a `URI` to interact with `RestClient`, and the javadoc for the `uri(URI)` method clearly states:
    > Specify the URI using an absolute, fully constructed URI.

    If you want to let `RestClient` fill in the `baseUrl` for you, you should provide it to `RestClient` when building using `.baseUrl(...)` on the builder, and then use the `uri(String)` or one of the non-URI variants to provide the path.

2. The issue you are encountering with `IllegalArgumentException: request cannot be null` is because you are calling `requestInterceptor.setAuthorizedClientRepository(authorizedClientRepository)` when setting up the interceptor. This sets up failure handling to expect a `request`. There is also `requestInterceptor.setAuthorizedClientService(authorizedClientService)` which is correct for your use case, because you are using `AuthorizedClientServiceOAuth2AuthorizedClientManager` with `@Async`. Please use this setup in order to fix your issue. **Note:** Calling an async method causes the `RequestContextHolder` to lose the context of the original request, which causes the issue you encountered.

Again, thank you for trying things out and looking for issues. It is very helpful to know that this use case requires careful setup in order to avoid issues, and I think clear documentation will help immensely.

Lonelysoul-HayashiNoMirai:
@sjohnr
Well, i don't know what to say, thanks bro, love u:

Point 1: "You have modified..." It works flawlessly.
Point 2: "The issue you are encountering...": It seems to be working fine... yet the first request call (after the server app restart) still fails with 401, it becomes Ok if I resend the request again (pls check the image below), is this expected behavior?, do I need to implement the retry logic myself?:
image

sjohnr:
> is this expected behavior?, do I need to implement the retry logic myself? (May be, I would use spring-retry):

It is expected behavior, yes. Adding retry would be pretty cool, let me know if that works for you!

Lonelysoul-HayashiNoMirai:
> > is this expected behavior?, do I need to implement the retry logic myself? (May be, I would use spring-retry):
> 
> It is expected behavior, yes. Adding retry would be pretty cool, let me know if that works for you!

@sjohnr
FYI, the combination with the spring-retry works perfectly! Once again, thank u for... everything, bro <3

Issue #15378, opened 2024-07-08
Response header ``Vary`` since Spring Security 6.2.0


renetrefft:
After upgrading from Spring Boot 3.1.* to Spring Boot 3.2.0 which includes Spring Security 6.2.0, responses of REST services provided by ``@Controller`` classes have ``Vary`` headers if ``org.springframework.boot:spring-boot-starter-security`` is in classpath.

Is this intended? We immediately noticed this change since our CDN Akamai does not cache responses with this header.

malaquf:
Hi @renetrefft, we've observed the same issue and I believe [this is the change/fix](https://github.com/spring-projects/spring-framework/commit/d27b5d0ab6e8b91a77e272ad57ae83c7d81d810b#diff-d854062b0e7c407eadbc755d6226f401R62) that causes it.

malaquf:
Hi @renetrefft , we also noticed this issue and I believe [this is the commit](https://github.com/spring-projects/spring-security/commit/618847418f415ac818be3ad92ecf1afe0635958c) that introduces it. Now CorsFilter is configured by default, if CorsConfigurationSource is present.

malaquf:
I believe this side effect is not intended, as HandlerMappingIntrospector implements CorsConfigurationSource and is instantiated by default by WebMvcConfigurationSupport in spring boot auto configure.

jzheaux:
I agree that this is likely not the intent. It may be best for Spring Security to be more conservative for the time being and pick up only `UrlBasedCorsConfigurationSource` instances.

baezzys:
Hi @jzheaux I'd like to contribute to this issue. Can I work on it?

marcusdacoregio:
Closed via https://github.com/spring-projects/spring-security/commit/3d4bcf1b4483e927eec05f2c01c5359e0b714267

piotrooo:
I think this change should be interpreted as a **breaking change** since, after the update from 3.3.2 to 3.3.3, it is no longer working.

FYI @marcusdacoregio, @baezzys and @jzheaux 

```java
@Configuration
public class CorsWebConfiguration implements WebMvcConfigurer {
    @Override
    public void addCorsMappings(CorsRegistry registry) {
        registry.addMapping("/web/**")
                .allowedOriginPatterns("https://*.example.com")
                .allowedMethods("GET", "POST", "PUT", "DELETE", "OPTIONS", "HEAD");
    }
}
```

PR #15444, opened 2024-07-21
Restrict automatic CORS configuration to UrlBasedCorsConfigurationSource

baezzys:
Closes [gh-15378](https://github.com/spring-projects/spring-security/issues/15378)

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configuration/HttpSecurityConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configuration/HttpSecurityConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -47,7 +47,7 @@ import org.springframework.security.crypto.password.PasswordEncoder;
 import org.springframework.security.web.context.request.async.WebAsyncManagerIntegrationFilter;
 import org.springframework.web.accept.ContentNegotiationStrategy;
 import org.springframework.web.accept.HeaderContentNegotiationStrategy;
-import org.springframework.web.cors.CorsConfigurationSource;
+import org.springframework.web.cors.UrlBasedCorsConfigurationSource;
 
 import static org.springframework.security.config.Customizer.withDefaults;
 
@@ -55,6 +55,7 @@ import static org.springframework.security.config.Customizer.withDefaults;
  * {@link Configuration} that exposes the {@link HttpSecurity} bean.
  *
  * @author Eleftheria Stein
+ * @author Jinwoo Bae
  * @since 5.4
  */
 @Configuration(proxyBeanMethods = false)
@@ -131,8 +132,7 @@ class HttpSecurityConfiguration {
 	}
 
 	private void applyCorsIfAvailable(HttpSecurity http) throws Exception {
-		String[] beanNames = this.context.getBeanNamesForType(CorsConfigurationSource.class);
-		if (beanNames.length == 1) {
+		if (this.context.getBeanNamesForType(UrlBasedCorsConfigurationSource.class).length > 0) {
 			http.cors(withDefaults());
 		}
 	}
--- a/config/src/test/java/org/springframework/security/config/annotation/web/configuration/HttpSecurityConfigurationTests.java
+++ b/config/src/test/java/org/springframework/security/config/annotation/web/configuration/HttpSecurityConfigurationTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


marcusdacoregio:
Hi @baezzys, thanks for the PR. Can you please target the `6.2.x` branch in your PR and rebase your branch with it?

config/src/main/java/org/springframework/security/config/annotation/web/configuration/HttpSecurityConfiguration.java
```diff
@@ -131,8 +133,13 @@ HttpSecurity httpSecurity() throws Exception {
 	}
 
 	private void applyCorsIfAvailable(HttpSecurity http) throws Exception {
-		String[] beanNames = this.context.getBeanNamesForType(CorsConfigurationSource.class);
-		if (beanNames.length == 1) {
+		Map<String, CorsConfigurationSource> corsConfigurationSources = this.context
+			.getBeansOfType(CorsConfigurationSource.class);
+
+		boolean hasUrlBasedCorsConfigurationSource = corsConfigurationSources.values()
```
marcusdacoregio, line 139: Would you please adopt another strategy instead of [using Stream](https://github.com/spring-projects/spring-security/issues/9642)?

marcusdacoregio (changes requested):

baezzys:
Hi @marcusdacoregio, I have completed the rebase and retargeted the PR to the 6.2.x branch.

config/src/main/java/org/springframework/security/config/annotation/web/configuration/HttpSecurityConfiguration.java
baezzys: I have refactored the code to avoid using Streams. PTAL. Thanks.

config/src/main/java/org/springframework/security/config/annotation/web/configuration/HttpSecurityConfiguration.java
```diff
@@ -131,9 +133,14 @@ HttpSecurity httpSecurity() throws Exception {
...
 	private void applyCorsIfAvailable(HttpSecurity http) throws Exception {
-		String[] beanNames = this.context.getBeanNamesForType(CorsConfigurationSource.class);
-		if (beanNames.length == 1) {
-			http.cors(withDefaults());
+		Map<String, CorsConfigurationSource> corsConfigurationSources = this.context
+			.getBeansOfType(CorsConfigurationSource.class);
+
+		for (CorsConfigurationSource source : corsConfigurationSources.values()) {
+			if (source instanceof UrlBasedCorsConfigurationSource) {
+				http.cors(withDefaults());
```
marcusdacoregio, line 141: I wonder if we should check if the instance is `UrlBasedCorsConfigurationSource` and if the bean name is `corsConfigurationSource`, since this is the [bean name used by the `CorsConfigurer`](https://github.com/spring-projects/spring-security/blob/main/config/src/main/java/org/springframework/security/config/annotation/web/configurers/CorsConfigurer.java#L89), to avoid picking up the wrong `CorsConfigurationSource`.

config/src/test/java/org/springframework/security/config/annotation/web/configuration/HttpSecurityConfigurationTests.java
```diff
@@ -374,6 +374,15 @@ public void configureWhenCorsConfigurationSourceThenApplyCors() {
...
 
+	@Test
+	public void configureWhenNoUrlBasedCorsConfigThenNoCorsApplied() throws Exception {
+		this.spring.register(NonUrlBasedCorsConfig.class, DefaultWithFilterChainConfig.class).autowire();
+		SecurityFilterChain filterChain = this.spring.getContext().getBean(SecurityFilterChain.class);
+		assertThat(filterChain.getFilters()).noneMatch((filter) -> filter instanceof CorsFilter);
+
+		this.mockMvc.perform(formLogin()).andExpect(header().doesNotExist("Access-Control-Allow-Origin"));
+	}
+
```
marcusdacoregio, line 385: Can you please add a test that verifies if the header `Vary` is not present? In summary, simulate the problem reported in https://github.com/spring-projects/spring-security/issues/15378 and assert that it is fixed.

You can add the issue number in the test, like so:

```java
// gh-15378
@Test
void ...() {
}
```

marcusdacoregio (changes requested):
Thanks @baezzys, I've left some feedback inline.

config/src/main/java/org/springframework/security/config/annotation/web/configuration/HttpSecurityConfiguration.java
baezzys: I have updated the code to check if there are any beans of type UrlBasedCorsConfigurationSource using getBeanNamesForType. If such beans exist, CORS configuration is applied.

If this is not what you intended, please feel free to provide further feedback.

config/src/test/java/org/springframework/security/config/annotation/web/configuration/HttpSecurityConfigurationTests.java
baezzys: I have updated the existing test code to verify if the Vary header is not present. Thank you for your feedback!

marcusdacoregio:
Thanks @baezzys, this is now merged into `6.2.x`, `6.3.x` and `main`.

PR #15472, opened 2024-07-25
Docs: Fix wrong annotation class name for Kotlin example

tahakorkem:
Fixed wrong annotation class name `IsAdmin` as `HasRole` for Kotlin example.


jzheaux:
Thanks for the correction, @tahakorkem! This is now merged into `6.3.x` and `main`.

PR #15477, opened 2024-07-27
Post code suggestions on PRs with format issues

timtebeek:
These two workflows together ensure that if a pull request comes in that has formatting issues, that a `github-actions` bot then posts a code suggestion comment for fix any offending lines. Those comments would look something like this

image

I've used this to good effect on the OpenRewrite GitHub org, where we run some additional automations for copyright headers and [coding best practices](https://docs.openrewrite.org/recipes/recipes/openrewritebestpractices).

This has been discussed over the past year or so with @rwinch (👋🏻 ), as a good first step towards gently enforcing standards, as opposed to an unfriendly failing build pipeline with delayed feedback.

The PR leverages https://github.com/googleapis/code-suggester to turn a git diff into code suggestion comments, which we've found to work well for smaller formatting issues. https://github.com/reviewdog/reviewdog could be a possible alternative, as well as others.

I've of course not been able to test these new workflows, as it requires them to be available on the main branch first, but I've done my best to match up with the existing [.github/workflows/pr-build-workflow.yml](https://github.com/spring-projects/spring-security/blob/main/.github/workflows/pr-build-workflow.yml) in terms of actions used. I figured to offer a first step PR here, but of course this [could be generalized](https://github.com/openrewrite/gh-automation/blob/main/.github/workflows/receive-pr.yml) if desired. 

No worries if you'd rather go a different direction. I figured a direct PR would be easier to review than yet another issues to discuss this once more.

### DIFF
--- a/.github/workflows/pr-suggestions-workflow.yml
+++ b/.github/workflows/pr-suggestions-workflow.yml
@@ -1,0 +1,59 @@
+# Description: This workflow is triggered when the `pr-format-workflow` completes to post suggestions on the PR.
+# Since this pull request has write permissions on the target repo, we should **NOT** execute any untrusted code.
+# https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
+---
+name: pr-suggestions-workflow
+
+on:
+  workflow_run:
+    workflows: ["pr-format-workflow"]
+    types:
+      - completed
+
+jobs:
+  post-suggestions:
+    # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      pull-requests: write
+    env:
+      # https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token
+      ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{github.event.workflow_run.head_branch}}
+          repository: ${{github.event.workflow_run.head_repository.full_name}}
+
+      # Download the patch
+      - uses: actions/download-artifact@v4
+        with:
+          name: patch
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          run-id: ${{ github.event.workflow_run.id }}
+      - name: Apply patch
+        run: |
+          git apply git-diff.patch --allow-empty
+          rm git-diff.patch
+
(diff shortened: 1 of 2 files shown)


timtebeek:
@markpollack did I understand correctly you'd be looking to do the same on [spring-ai](https://github.com/spring-projects/spring-ai) through `./mvnw spring-javaformat:apply` and `./mvnw license:update-file-header -Plicense` ? Want me to open a similar PR targeting Maven there?

.github/workflows/pr-suggestions-workflow.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Read pr_number.txt
+        run: |
+          PR_NUMBER=$(cat pr_number.txt)
+          echo "PR_NUMBER=$PR_NUMBER" >> $GITHUB_ENV
+          rm pr_number.txt
+
+      # Post suggestions as a comment on the PR
+      - uses: googleapis/code-suggester@v4
+        with:
+          command: review
```
jzheaux, line 57: Is there a way to include a comment in the patch suggestion? If so, it would be nice to inform the OP of `./gradlew format`. I realize they can just accept the PR suggestion instead, I just see it as informative for the future so they remove one step from their submissions.

jzheaux (changes requested):
I like it, @timtebeek. I've left some feedback inline. Also, since this will affect the whole team, I'll add `team-attention` and we can discuss at our next standup.

jzheaux:
@timtebeek, instead of code suggestions, could it try and push a polish commit to the PR's branch instead?

If the push fails then post a comment to the PR indicating how to fix the formatting themselves.

timtebeek:
> @timtebeek, instead of code suggestions, could it try and push a polish commit to the PR's branch instead?
> 
> If the push fails then post a comment to the PR indicating how to fix the formatting themselves.

Hmm; haven't explored that option yet, and it'd take some care to be sure that this can be done safely. Any PR may contain untrusted code, and allowing an automation to run against that code and push to a branch might be a bit of a challenge to do safely, especially if we do not want to use privileged tokens. The _fun_ part of running an OSS project I suppose.

Let me know your thoughts on this; I don't yet know if I'll find time to explore that alternative in the near future.

timtebeek:
I don't think there's a good safe way to execute a command on a not-yet-vetted PR and push a change back to the branch, especially when allowing anyone to open PRs. Up to you then whether posting comments (possible without any need for tokens) is still enough of an improvement to merge this in, or whether to close this PR.

From our experience enforcing a formatter and consistent license headers across PRs through comments has been a great help to reduce the review load, but to each their own of course. I'm just as fine with not seeing this merged; it's been ~two years since this was first discussed while cycling outside of Barcelona; sad to miss Rob there this year! :)

timtebeek:
I'll close this effort as it seems unlikely to get merged at this point. Let me know if you'd want to add this at some point in the future. All the best!

Issue #3192, opened 2015-05-23
SEC-2983: InMemoryUserDetailsManager#loadUserByUsername should preserve custom UserDetails types


spring-projects-issues:
[William Gorder](https://jira.spring.io/secure/ViewProfile.jspa?name=wgorder) (Migrated from [SEC-2983](https://jira.spring.io/browse/SEC-2983?redirect=false)) said:

Is there a good reason for not returning MutableUser.delegate in the  InMemoryUserDetailsManager.loadUserByUsername() method?  What I have run into is I want to provide my own user details, so I pass them to the create() method but when I get them back from authentication.getPrincipal() its been replaced with the vanilla spring security User object since apparently rather then just returning the UserDetails that we already have it takes the delegate and  creates a new org.springframework.security.core.userdetails.User.

What I am proposing is a something different then the MutableUserDetails interface. The MutableUser object would need to expose its delegate so that it could be returned if it exists.  Currently the only way to get what I want is to create my own version of InMemoryUserDetailsManager and InMemoryUserDetailsManagerConfigurer unless I am missing something.  It just seems that by exposing a create(UserDetails) method it is implied that we would get the UserDetails provided and not something different.  I see this bit of the API as confusing and bit misleading.  There is nothing in the Java Doc implying that that underlying implementation will be changed.

jzheaux:
> Is there a good reason for not returning MutableUser.delegate in the InMemoryUserDetailsManager.loadUserByUsername() method?

I believe one benefit is that `User` implements `CredentialsContainer`, ensuring that the password will be removed completely after authentication.

> Currently the only way to get what I want is to create my own version of InMemoryUserDetailsManager and InMemoryUserDetailsManagerConfigurer 

Things have obviously changed a lot since the 3.x days. Today publishing a `UserDetailsManager` bean or, even simpler, [a `UserDetailsService`](https://github.com/spring-projects/spring-security-samples/blob/main/servlet/spring-boot/java/authentication/username-password/user-details-service/custom-user/src/main/java/example/CustomUserRepositoryUserDetailsService.java) is quite simple and is the recommended route for custom `UserDetails` implementations.

That said, one thing I think we could do is store any `MutableUserDetails` instead of wrapping it like so:

```java
public void createUser(UserDetails user) {
    if (user instanceof MutableUserDetails mutable) {
        this.users.put(user.getUsername().toLowerCase(), mutable);
    } else {
[...]

PR #15498, opened 2024-07-29
Preserve custom user type in InMemoryUserDetailsManager

MrJovanovic13:
This PR fixes issue gh-3192 by allowing `MutableUserDetails` to be stored and loaded inside the `InMemoryUserDetailsManager` instead of always wrapping it into `MutableUser`. The custom user type must implement `CredentialsContainer` in order to be returned by `loadUserByUsername()` and not transformed into `User`.

I wanted to write tests for the new cases, and I saw that methods were not previously covered so I wrote tests for all cases for `createUser()` and `loadUserByUsername()`. However, I have concerns about some tests:
`createUserWhenNotInstanceOfMutableUserDetailsThenShouldWrapIntoMutableUser` and
`createUserWhenInstanceOfMutableUserDetailsThenShouldNotWrapIntoMutableUser` can't directly access `Map<String, MutableUserDetails> users` to get the underlying user & type, so they are relying on `loadUserByUsername` to get the user and as such are not pure. 
If you think this can be improved somehow, I am open to feedback.

### DIFF
--- a/core/src/main/java/org/springframework/security/provisioning/InMemoryUserDetailsManager.java
+++ b/core/src/main/java/org/springframework/security/provisioning/InMemoryUserDetailsManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -30,6 +30,7 @@ import org.springframework.security.access.AccessDeniedException;
 import org.springframework.security.authentication.AuthenticationManager;
 import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
 import org.springframework.security.core.Authentication;
+import org.springframework.security.core.CredentialsContainer;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.security.core.context.SecurityContextHolderStrategy;
 import org.springframework.security.core.userdetails.User;
@@ -96,7 +97,13 @@ public class InMemoryUserDetailsManager implements UserDetailsManager, UserDetai
 	@Override
 	public void createUser(UserDetails user) {
 		Assert.isTrue(!userExists(user.getUsername()), "user should not exist");
-		this.users.put(user.getUsername().toLowerCase(), new MutableUser(user));
+
+		if (user instanceof MutableUserDetails mutable) {
+			this.users.put(user.getUsername().toLowerCase(), mutable);
+		}
+		else {
+			this.users.put(user.getUsername().toLowerCase(), new MutableUser(user));
+		}
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/springframework/security/provisioning/InMemoryUserDetailsManager.java
```diff
@@ -154,6 +161,9 @@ public UserDetails loadUserByUsername(String username) throws UsernameNotFoundEx
 		if (user == null) {
 			throw new UsernameNotFoundException(username);
 		}
+		if (user instanceof CredentialsContainer) {
+			return user;
+		}
 		return new User(user.getUsername(), user.getPassword(), user.isEnabled(), user.isAccountNonExpired(),
 				user.isCredentialsNonExpired(), user.isAccountNonLocked(), user.getAuthorities());
 	}
```
jzheaux, line 169: Can you also please update `updateUser` in the same way?

core/src/test/java/org/springframework/security/provisioning/InMemoryUserDetailsManagerTests.java
```diff
@@ -97,4 +102,88 @@ public void changePasswordWhenCustomSecurityContextHolderStrategyThenUses() {
...
+	public void createUserWhenInstanceOfMutableUserDetailsThenShouldNotWrapIntoMutableUser() {
+		InMemoryUserDetailsManager manager = new InMemoryUserDetailsManager();
+		CustomUser user = new CustomUser(User.withUserDetails(PasswordEncodedUser.user()).build());
+		manager.createUser(user);
+		UserDetails userDetails = manager.loadUserByUsername(user.getUsername());
+		assertThat(userDetails).isInstanceOf(CustomUser.class);
+	}
+
+	@Test
+	public void loadUserByUsernameWhenUserNullThenException() {
```
jzheaux, line 130: Thank you for the extra tests. Will you please place them in a separate commit so that I can backport those to earlier versions?

jzheaux (changes requested):
Thanks, @MrJovanovic13, for the PR! I've left some feedback inline.

core/src/test/java/org/springframework/security/provisioning/InMemoryUserDetailsManagerTests.java
MrJovanovic13: I have placed them in a separate commit.

core/src/main/java/org/springframework/security/provisioning/InMemoryUserDetailsManager.java
MrJovanovic13: Yes, done.

core/src/test/java/org/springframework/security/provisioning/InMemoryUserDetailsManagerTests.java
```diff
@@ -97,4 +102,88 @@ public void changePasswordWhenCustomSecurityContextHolderStrategyThenUses() {
...
 	}
 
+	@Test
+	public void createUserWhenUserAlreadyExistsThenException() {
+		assertThatIllegalArgumentException().isThrownBy(() -> this.manager.createUser(this.user))
+			.withMessage("user should not exist");
+	}
+
+	@Test
+	public void createUserWhenNotInstanceOfMutableUserDetailsThenShouldWrapIntoMutableUser() {
```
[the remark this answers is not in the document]
MrJovanovic13: Thanks for the pointers, that did help. I pushed the refactored tests.

Please let me know if any additional changes are needed, I believe I understood you correctly, the only test I am doubtful on is `createUserWhenInstanceOfMutableUserDetailsThenChangePasswordWorks`.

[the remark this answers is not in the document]
jzheaux: Looks good!

jzheaux (approved):

jzheaux:
Thanks again, @MrJovanovic13! This is now merged into `main`.

PR #15525, opened 2024-08-05
Improve authentication by trimming input and refactoring obtain methods

Seol-JY:
Updated the `UsernamePasswordAuthenticationFilter` to automatically trim leading and trailing whitespace from both the `username` and `password` fields in the login form.

refactors the obtainUsername and obtainPassword methods in UsernamePasswordAuthenticationFilter to ensure they return non-null strings, defaulting to an empty string if the parameter is missing. Additionally, these methods now automatically trim leading and trailing whitespace from the username and password, improving the user experience by preventing errors caused by accidental spaces.

I couldn't find a valid reason not to trim the password.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/UsernamePasswordAuthenticationFilter.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/UsernamePasswordAuthenticationFilter.java
@@ -74,10 +74,10 @@ public class UsernamePasswordAuthenticationFilter extends AbstractAuthentication
 		if (this.postOnly && !request.getMethod().equals("POST")) {
 			throw new AuthenticationServiceException("Authentication method not supported: " + request.getMethod());
 		}
+
 		String username = obtainUsername(request);
-		username = (username != null) ? username.trim() : "";
 		String password = obtainPassword(request);
-		password = (password != null) ? password : "";
+
 		UsernamePasswordAuthenticationToken authRequest = UsernamePasswordAuthenticationToken.unauthenticated(username,
 				password);
 		// Allow subclasses to set the "details" property
@@ -100,7 +100,13 @@ public class UsernamePasswordAuthenticationFilter extends AbstractAuthentication
 	 */
 	@Nullable
 	protected String obtainPassword(HttpServletRequest request) {
-		return request.getParameter(this.passwordParameter);
+		String password = request.getParameter(this.passwordParameter);
+
+		if (password != null) {
+			return password.trim();
+		}
+
+		return "";
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


web/src/main/java/org/springframework/security/web/authentication/UsernamePasswordAuthenticationFilter.java
```diff
@@ -100,7 +100,13 @@ public Authentication attemptAuthentication(HttpServletRequest request, HttpServ
 	 */
 	@Nullable
 	protected String obtainPassword(HttpServletRequest request) {
-		return request.getParameter(this.passwordParameter);
+		String password = request.getParameter(this.passwordParameter);
+
+		if (password != null) {
```
jzheaux, line 105: We can't trim the password since a password might contain a space on the end.

web/src/main/java/org/springframework/security/web/authentication/UsernamePasswordAuthenticationFilter.java
```diff
@@ -74,10 +74,10 @@ public Authentication attemptAuthentication(HttpServletRequest request, HttpServ
 		if (this.postOnly && !request.getMethod().equals("POST")) {
 			throw new AuthenticationServiceException("Authentication method not supported: " + request.getMethod());
 		}
+
 		String username = obtainUsername(request);
-		username = (username != null) ? username.trim() : "";
```
jzheaux, line 78: For backward compatibility, this line needs to stay where it is. If someone is overriding obtainUsername today, then this change would mean when they upgrade, the trimming would stop.

If you need it to be switched off perhaps we could add a property to toggle the trimming behavior.

web/src/main/java/org/springframework/security/web/authentication/UsernamePasswordAuthenticationFilter.java
```diff
@@ -74,10 +74,10 @@ public Authentication attemptAuthentication(HttpServletRequest request, HttpServ
...
 		String password = obtainPassword(request);
-		password = (password != null) ? password : "";
```
jzheaux, line 80: For backward compatibility, this line needs to stay where it is. If someone is overriding obtainPassword today, then this change would mean when they upgrade, the defaulting would stop.

jzheaux (changes requested):
Hi, @Seol-JY, thanks for the PR! I'm not sure that we would want to do this, largely due to backward compatibility concerns. Can you explain what you are trying to accomplish by making this change?

web/src/main/java/org/springframework/security/web/authentication/UsernamePasswordAuthenticationFilter.java
Seol-JY: The intention behind my pull request was to ensure that methods starting with 'obtain' always return a valid string (or an empty string) rather than potentially returning null. I believed this approach would lead to more robust and predictable behavior. However, as you've pointed out, I now realize that I may not have adequately considered backward compatibility issues in my proposed changes.

jzheaux:
Sounds good, @Seol-JY. Given [this comment](https://github.com/spring-projects/spring-security/pull/15525#discussion_r1713613398), I'm going to close this PR. I'd love to see additional contributions from you down the road.

PR #15537, opened 2024-08-08
Use explicit types instead of var

Kehrlann:
- Removed usage of `var` in a few classes
- Removed unused method


rwinch:
Thanks for the PR! I rebased it off of 6.3.x and then merged into main via 71f40f2bc47e81f9c8dcf5fdfdc025da2a2c270c

Issue #14904, opened 2024-04-13
Add Spring Session support to OIDC Back-Channel Logout


pzgadzaj-equinix:
**Describe the bug**
When using Spring boot in version 3.2.1, together with Redis-base session store, session invalidation fails because of lack of Base64 cookie encoding

When back channel logout implementation tries to invalidate the session, It makes a POST with Session cookie created based on session stored in OidcSessionRegistry. Value of the session cookie is not being base64-encoded: https://github.com/spring-projects/spring-security/blob/main/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OidcBackChannelLogoutHandler.java#L108

When the this POST is being handled, Session cookie is by default base64-decoded: https://github.com/spring-projects/spring-session/blob/main/spring-session-core/src/main/java/org/springframework/session/web/http/DefaultCookieSerializer.java#L101

which cause that the session invalidation fails

**To Reproduce**
1. Prepare an application which uses Spring session stored in Redis + OIDC back channel configured
2. Log in to the application using OIDC integration
3. Trigger OIDC back channel logout

**Expected behavior**
1. Session established in step 2 is invalidated

**Sample**

A link to a GitHub repository with a [minimal, reproducible sample](https://stackoverflow.com/help/minimal-reproducible-example).

Reports that include a sample will take priority over reports that do not.
At times, we may require a sample, so it is good to try and include a sample up front.

jzheaux:
Hi, @pzgadzaj-equinix, thanks for reaching out. Spring Session support for OIDC Backchannel Logout is forthcoming as we also need to expose the ability to change the cookie name. Or it may be the case that Spring Session publishes a `LogoutHandler` of its own so that it can apply the `CookieSerializer` directly.

I'll leave this ticket to explore the best route for that.

aelillie:
When supporting Spring Session with OIDC Backchannel Logout you also need to consider how the `OidcSessionRegistry` should be notified about removing the stored session information, as it is expected to be initiated by the `SessionDestroyedEvent` (via `HttpSessionEventPublisher`) and handled in `OAuth2LoginConfigurer.OidcClientSessionEventListener.onApplicationEvent`.

pzgadzaj-equinix:
@xiechangning20 I guess that disabling bas64 encoding is an option but then this also requires using Session ids which don't need to be base64-encoded

[...]

Issue #13841, opened 2023-09-16
Consider OIDC Back-Channel favoring logout_token over CSRF


jzheaux:
The current OIDC back-channel logout support saves the end-user's CSRF token to use in a self-logout call when the back-channel request comes from the authorization server.

This adds more information to `OidcSessionInformation` than may be necessary. Instead, I think it would also work to send the `logout_token` in place of the CSRF token.

The upside is a simpler contract and simpler configuration. The possible downside is the logout token is validated multiple times, once for each session being invalidated.

jzheaux:
Waiting on #14510

jzheaux:
Given the feedback in https://github.com/spring-projects/spring-security/issues/15227, I think that's enough to consider adding this without waiting for #14510.

Issue #15227, opened 2024-06-11
Back-Channel Logout fails with cookie-based CSRF protection


ch4mpy:
## Describe the bug
I'm trying to configure Back-Channel Logout on an [OAuth2 BFF](https://www.baeldung.com/spring-cloud-gateway-bff-oauth2): a reactive Spring Cloud Gateway instance configured with `oauth2Login` and the `TokenRelay=` filter.

As this BFF is used with single-page applications, CSRF protection is cookie-based (and this works for the RP-Initiated Logout: the POST request to `/logout` is correctly protected in this case).

During Back-Channel Logout, the internal request to `/logout` fails with 403 FORBIDDEN due to CSRF authorization failure.

Note that the Back-Channel Logout is successful as soon as:
- I disable CSRF protection, but this makes the all system vulnerable to CSRF attacks (all REST requests from frontends are going through this BFF, initially authorized with a session cookie)
- switch to session store for the CSRF token (but that breaks all `POST`, `PUT`, `PATCH` & `DELETE` requests from single-page & mobile apps).

**This makes Back-Channel Logout unusable in production with single-page & mobile apps**.

## To Reproduce
Enable cookie-based protection against CSRF and then initiate a Back-Channel Logout.

## Expected behavior
The Back-Channel Logout should be successful, whatever store is used for the CSRF token.

Why should the CSRF protection be enforced in a flow where the user agent is not involved?

## Sample
Pre-requisites:
- JDK between 17 & 21 on the path
- node LTS on the path
- Docker Desktop up
```sh
git clone https://github.com/ch4mpy/quiz.git
cd quiz
sh ./build.sh
```
This builds and composes all the services in docker.

Frontend URI is logged at the end of the build script (it contain the building machine hostname): `http://hostname/ui/`

Frontend users in `quiz` realm are `ch4mp`, `moderator`, `trainee`, and `trainer`. All have `secret` as secret.

To trigger a Back-Channel Logout, visit the Keycloak user account in the `quiz` realm and click the logout button from the top right corner: `http://hostname/auth/realms/quiz/account/`

To debug the OAuth2 client, stop the `quiz.bff` Docker container and start the `api/bff` Spring Boot project in debug mode with your favorite IDE.

Switching the CSRF protection strategy is just a matter of editing the `com.c4-soft.springaddons.oidc.client.csrf` property in the BFF `application.yml`.

[...]

Issue #14553, opened 2024-02-06
Back-Channel Logout should use localhost for internal logout request


ch4mpy:
**Describe the bug**
Given I have:
- a Spring OAuth2 client configured with Back-Channel Logout
- an authorization server with Back-Channel Logout configured to call the Spring OAuth2 client

When I try to visit a path requiring to be authorized on the Spring client, then I am redirected to login.

Once authenticated, I can access the protected resource.

When I visit the `end_session_endpoint` of the authorization server, then I can see a `TRACE` log on the Spring client displaying: `Found and removed 1 session(s) from mapping of 1 session(s)`

However, if I refresh the tab pointing to the protected resource, I can still access it.

There are two log lines after the one stating that the session was removed:
```
ExchangeFunctions : [1c75356f] HTTP POST http://host.docker.internal:7080/logout, headers={masked}
ExchangeFunctions : [1c75356f] Cancel signal (to close connection)
```
Is the client trying to call its own `/logout` endpoint but using the wrong hostname? `host.docker.internal` is the hostname used by the authorization server to initiate the Back-Channel Logout (which is running in Docker), but this hostname is unknown from the Spring client.

**Expected behavior**
If the Spring client needs to call itself to actually destroy the user session, shouldn't it call `localhost`?

**Sample**

https://github.com/ch4mpy/spring-security-14553 

run `docker compose up` to create a new Keycloak instance (`admin` / `admin` to access [http://localhost:8080/realms/master](http://localhost:8080/realms/master))

create a new realm by importing `spring-security-realm.json`

create a user

start the spring boot app

visit [http://localhost:7080/](http://localhost:7080/)

the `end_session_endpoint` is at [http://localhost:8080/realms/spring-security/protocol/openid-connect/logout](http://localhost:8080/realms/spring-security/protocol/openid-connect/logout)

jzheaux:
@ch4mpy, good observation. I think we can change this:

```
String logout = UriComponentsBuilder.fromHttpUrl(url)
			.replacePath(this.logoutEndpointName)
			.build()
			.toUriString();
```

to:

```diff
String logout = UriComponentsBuilder.fromHttpUrl(url)
+			.host("localhost")
			.replacePath(this.logoutEndpointName)
			.build()
			.toUriString();
```

to address the issue.

ch4mpy:
@jzheaux thank you.

koyama-tagbangers:
In the production environment, the logout endpoint is accessed via https.
[...]

PR #15540, opened 2024-08-08
Oidc Logout Improvements

jzheaux:
This brings support for #13841 and #14904.

To configure OIDC Back-Channel logout to reuse the Back-Channel endpoint to invalidate each individual session (thus removing the need for the CSRF token), configure OIDC logout in the following way:

```java
.oidcLogout((oidc) -> oidc
    .backChannel(Customizer.withDefaults())
)
```

**NOTE** that for simplicity, this changes the default internal URI. Since it changes it to point back to the OIDC back-channel URL itself, I believe this will go unnoticed by most applications. That said, in case you need your URI to stay as-is, you can specify the URI as follows:

```java
// ...
.oidcLogout((oidc) -> oidc
    .backChannel((backChannel) -> backChannel
        .logoutUri(myCustomInternalUri)
    )
)
```

To configure OIDC Back-Channel Logout with a different cookie name, configure it as follows:

```java
@Bean 
OidcSessionRegistry sessionRegistry() {
    return new InMemoryOidcSessionRegistry();
}

@Bean
OidcBackChannelLogoutHandler oidcLogoutHandler(OidcSessionRegistry sessionRegistry) {
    OidcBackChannelLogoutHandler logoutHandler = new OidcBackChannelLogoutHandler(sessionRegistry);
    // the logout URI can also be configured here
    logoutHandler.setSessionCookieName("SESSION");
    return logoutHandler;
}
```

Or it can also be specified in the DSL directly:

```java
// ...
.oidcLogout((oidc) -> oidc
    .backChannel((backChannel) -> backChannel
        .logoutHandler(logoutHandler)
    )
)
```


jzheaux:
@ch4mpy would you please look over this PR as see if is addresses your concerns about CSRF and Spring Session?

ch4mpy:
@jzheaux I'll have a look during the week-end if it's not too late. But if the CSRF token is not checked anymore during Back-Channel Logout (I believe it shouldn't as this is a server-to-server protocol in which the user agent shouldn't be involved), then it should work for any kind of frontend!

ch4mpy:
@jzheaux I can confirm that, when building the branch for this PR locally, Back-Channel Logout works in [this project of mine](https://github.com/ch4mpy/quiz). It is an Angular application sending REST requests through an OAuth2 BFF: a spring-cloud-gateway instance configured with:
- `oauth2Login`
- cookie-based protection against CSRF (because it is called by a SPA authorizing requests with a session cookie)
- RP-Initiated Logout (keeps working, validating the CSRF token)
- Back-Channel Logout ([which fails](https://github.com/spring-projects/spring-security/issues/15227), is successful with this PR)

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OidcBackChannelLogoutHandler.java
```diff
@@ -137,6 +145,9 @@ String computeLogoutEndpoint(HttpServletRequest request) {
 		int port = uriComponents.getPort();
```
ch4mpy, line 145: shouldn't the `server.port` property be used here instead of the original request port?

config/src/main/java/org/springframework/security/config/web/server/OidcBackChannelServerLogoutHandler.java
```diff
@@ -137,6 +149,9 @@ String computeLogoutEndpoint(ServerHttpRequest request) {
 		int port = uriComponents.getPort();
```
ch4mpy, line 149: Same question as for the servlet implementation: shouldn't the server.port property be used here instead of the original request port?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OidcBackChannelLogoutHandler.java
```diff
@@ -137,6 +145,9 @@ String computeLogoutEndpoint(HttpServletRequest request) {
...
 		uriVariables.put("basePort", (port == -1) ? "" : ":" + port);
 
+		String registrationId = token.getClientRegistration().getRegistrationId();
+		uriVariables.put("registrationId", registrationId);
+
 		return UriComponentsBuilder.fromUriString(this.logoutUri)
```
ch4mpy, line 151: What if the application is configured with a base path?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OidcBackChannelLogoutHandler.java
jzheaux: I've just caught up in the discussion in https://github.com/spring-projects/spring-security/issues/14553 and am now better versed in the complexities around using `localhost`. Instead, I think a more reasonable default is `{baseUrl}/logout/connect/back-channel/{registrationId}`.

The reason it is still necessary to specify the path is that it needs to remain compatible with the previous approach, which will still be available for a time. Also, this is very similar to other OAuth 2.0 URI default configuration values.

ch4mpy: Honestly, the current code is reasonable as, thanks to the URI configuration, I could get back-channel logout work with a SPA using cookie-based protection against CSRF.

We are talking about what could make a better default, and yes, using `{baseUrl}` is probably one.

dreamstar-enterprises:
I maybe barking up the wrong tree here, but this first principles approach worked for me, in a way that I could also understand. I'm sure people could adapt the below to their needs too

 ```
**
 * Configuration properties for server endpoints and authentication settings.
 *
 * @property route53Host AWS Route53 host for production routing
 * @property route53Port AWS Route53 port number
 * @property reverseProxyHost Reverse proxy host for development routing
 * @property reverseProxyPort Reverse proxy port number
 * @property bffServerPrefix BFF (Backend for Frontend) server prefix
 * @property resourceServerHost Resource server host address
 * @property resourceServerPort Resource server port number
 * @property resourceServerPrefix Resource server endpoint prefix
 * @property auth0AuthRegistrationId Auth0 registration identifier
 * @property auth0IssuerUri Auth0 issuer URI
 * @property auth0ApiAudience Auth0 API audience identifier
 * @property auth0DreamstarVivianaUrl Auth0 Dreamstar Viviana URL
 * @property inHouseAuthServerPrefix In-house authorization server prefix
 * @property inHouseAuthRegistrationId In-house authorization registration ID
 * @property sslProperties SSL configuration properties
 */
@ConfigurationProperties(prefix = "dse-viviana-servers")
internal class ServerProperties(
    private val springProfileProperties: SpringProfileProperties,
    private val sslProperties: SslProperties
) {

    /*************************/
    /* SERVERS               */
    /*************************/

    /**
     * Route53 Configuration
     */
    var route53Host: String? = null
    var route53Port: Int? = null

    /**
     * Constructs the complete Route53 URI using HTTPS scheme (External Calls)
     */
    val route53Uri: String
        get() = "${HttpSchemeTypesEnum.HTTPS}://$route53Host:$route53Port"

    /**
     * Reverse Proxy Configuration
     */
    var reverseProxyHost: String? = null
    var reverseProxyPort: Int? = null

    /**
     * Constructs the complete Reverse Proxy URI using HTTP scheme (External Calls)
     */
    val reverseProxyUri: String
        get() = "${HttpSchemeTypesEnum.HTTP}://$reverseProxyHost:$reverseProxyPort"

    /**
     * BFF Server Configuration
     */
    var bffServerHost: String? = null
    var bffServerPort: Int? = null
    var bffServerPrefix: String? = null

    /**
     * Constructs the complete BFF URI using HTTP scheme (Internal Calls)
     */
    val bffUri: String
        get() {
            val scheme = if (sslProperties.sslEnabled) HttpSchemeTypesEnum.HTTPS else HttpSchemeTypesEnum.HTTP
            return "$scheme://$bffServerHost:$bffServerPort"
        }

    /**
     * Determines the client URI based on active profile (External Calls)
     * @return URI string for client communications
     */
    val clientUri: String
        get() = when (springProfileProperties.activeProfile) {
            ProfileTypesEnum.DEVELOPMENT.type -> "$reverseProxyUri$bffServerPrefix"
            ProfileTypesEnum.PRODUCTION.type -> "$route53Uri$bffServerPrefix"
            else -> "$reverseProxyUri$bffServerPrefix"
        }

    /**
     * Resource Server Configuration
     */
    var resourceServerHost: String? = null
    var resourceServerPort: Int? = null
    var resourceServerPrefix: String? = null

    /**
     * Constructs the complete Resource Server URI using HTTP scheme (Internal Routing)
     */
    val resourceServerUri: String
        get() = "${HttpSchemeTypesEnum.HTTP}://$resourceServerHost:$resourceServerPort"

```

PR #15554, opened 2024-08-11
Improve documentation about CredentialsContainer

Chu3laMan:


marcusdacoregio:
Hi @Chu3laMan, thanks for the PR? Why did you replace the `javadoc:` macro? We are moving to use this macro now and it should be used.

Chu3laMan:
> Hi @Chu3laMan, thanks for the PR? Would you mind clarifying why you replaced the `javadoc:` macro? We are moving to use this macro now and it should be used.

Hi @marcusdacoregio thank you for your remark, i will fix it ...

sjohnr (changes requested):
Thank you for the PR @Chu3laMan. I have added a few comments inline below.

Issue #15567, opened 2024-08-12
The variable targetClassToUse is not passed into the synthesize method


kse-music:
As title

PR #15568, opened 2024-08-12
Fix variable targetClassToUse is not passed into the synthesize method

kse-music:
Closes gh-15567


jzheaux:

Issue #15204, opened 2024-06-05
CookieRequestCache should set SameSite on the REDIRECT_URI cookie


2is10:
**Describe the bug**
When using CookieRequestCache, Firefox complains:

> Cookie “REDIRECT_URI” does not have a proper “SameSite” attribute value. Soon, cookies without the “SameSite” attribute or with an invalid value will be treated as “Lax”. This means that the cookie will no longer be sent in third-party contexts. If your application depends on this cookie being available in such contexts, please add the “SameSite=None“ attribute to it. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite

![image](https://github.com/spring-projects/spring-security/assets/647851/fff9ae1c-5cbc-4711-9d1a-247bce26f8fd)

**To Reproduce**
Configure your application to use CookieRequestCache, like so:
```java
http.requestCache(config -> config.requestCache(new CookieRequestCache()));
```
Note that due to #2932 you likely also need to separately pass the CookieRequestCache to SavedRequestAwareAuthenticationSuccessHandler.

**Expected behavior**
No warning in Firefox. The POST /login request should also not show up as “blocked”.

![image](https://github.com/spring-projects/spring-security/assets/647851/b12268ac-da33-4aae-9b84-071b0ac1f65d)

**Sample**

No sample provided. This is easy to try on any project.

marcusdacoregio:
Hi @2is10, thanks for the report.

I believe that we could add some kind of customizer to the cookie in the `CookieRequestCache` implementation, similar to https://github.com/spring-projects/spring-security/pull/15203, what do you think?

I don't think that we should set a default value to that attribute tho, since [the docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#samesitesamesite-value) mention that the attribute is optional and, if not provided, a default value (defined by the browser) will be used.

PR #15685, opened 2024-08-23
Add cookie customizer to CookieRequestCache

fb64:
Issue https://github.com/spring-projects/spring-security/issues/15204

### DIFF
--- a/web/src/main/java/org/springframework/security/web/savedrequest/CookieRequestCache.java
+++ b/web/src/main/java/org/springframework/security/web/savedrequest/CookieRequestCache.java
@@ -18,6 +18,7 @@ package org.springframework.security.web.savedrequest;
 
 import java.util.Base64;
 import java.util.Collections;
+import java.util.function.Consumer;
 
 import jakarta.servlet.http.Cookie;
 import jakarta.servlet.http.HttpServletRequest;
@@ -51,6 +52,9 @@ public class CookieRequestCache implements RequestCache {
 
 	private static final int COOKIE_MAX_AGE = -1;
 
+	private Consumer<Cookie> cookieCustomizer = (cookie) -> {
+	};
+
 	@Override
 	public void saveRequest(HttpServletRequest request, HttpServletResponse response) {
 		if (!this.requestMatcher.matches(request)) {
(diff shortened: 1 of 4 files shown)


web/src/main/java/org/springframework/security/web/savedrequest/CookieRequestCache.java
```diff
@@ -152,4 +157,13 @@ public void setRequestMatcher(RequestMatcher requestMatcher) {
 		this.requestMatcher = requestMatcher;
 	}
 
+	/**
+	 * Sets the {@link Consumer}, allowing customization of cookie.
+	 * @param cookieCustomizer customize for cookie
```
marcusdacoregio, line 162: Can you please add `@since 6.4`?

marcusdacoregio (changes requested):
Thanks for the PR @fb64. I've left a single feedback inline.

Would you please add the same support in `CookieServerRequestCache`? I believe that the only difference is that the `cookieCustomizer` would be a `Consumer<ResponseCookie.Builder>`

fb64:
> Thanks for the PR @fb64. I've left a single feedback inline.
> 
> Would you please add the same support in `CookieServerRequestCache`? I believe that the only difference is that the `cookieCustomizer` would be a `Consumer<ResponseCookie.Builder>`

As far as I know `ResponseCookie` is used by reactive stack and cannot directly used in a standard `HttpServletResponse` unless refactoring the class, that's why I used a consumer of the same `Cookie` object already used. 
It's very similar to https://github.com/spring-projects/spring-security/pull/15203

marcusdacoregio:
Hi @fb64, I'm sorry if what I said was confusing. But I mean that the reactive implementation should be a `Consumer<ResponseCookie.Builder>`. Does that make sense?

fb64:
> Hi @fb64, I'm sorry if what I said was confusing. But I mean that the reactive implementation should use a `Consumer<ResponseCookie.Builder>`. Does that make sense?

Got it, I added it to `CookieServerRequestCache`  sorry for the misunderstanding 😅

marcusdacoregio:
Thanks @fb64, this is now merged into `main`.

PR #15686, opened 2024-08-23
Fix OIDC Logout docs: Session Strategy vs. Registry

hwanders:
# Contents

This contains the following changes to the OIDC logout documentation:
- replace `OidcSessionStrategy` by `OidcSessionRegistry`
- replace `ReactiveOidcSessionStrategy` by `ReactiveOidcSessionRegistry`
- fix broken/inconsistent Java code examples
- fix syntax of adoc source code blocks -- same as in https://github.com/spring-projects/spring-security/pull/15566 to avoid conflicts if that one gets merged first

Furthermore a test name was adjusted to reflect the right terminology.

# Reason
- There is no `OidcSessionStrategy` in Spring Security, the demo code indicates that actually `OidcSessionRegistry` was meant


jzheaux:
Thanks for the corrections, @hwanders! These are now merged into `6.2.x`, `6.3.x`, and `main` and should appear in the docs site shortly.

PR #15711, opened 2024-08-30
Annotation expression template processing should not fail on Class parameter types

kse-music:
relate to gh-15721

### DIFF
--- a/config/src/test/java/org/springframework/security/config/annotation/method/configuration/PrePostMethodSecurityConfigurationTests.java
+++ b/config/src/test/java/org/springframework/security/config/annotation/method/configuration/PrePostMethodSecurityConfigurationTests.java
@@ -998,6 +998,15 @@ public class PrePostMethodSecurityConfigurationTests {
 		verify(expressionHandler, times(4)).createEvaluationContext(any(Supplier.class), any());
 	}
 
+	// gh-15721
+	@Test
+	@WithMockUser(roles = "uid")
+	public void methodWhenMetaAnnotationPropertiesHasClassProperties() {
+		this.spring.register(MetaAnnotationPlaceholderConfig.class).autowire();
+		MetaAnnotationService service = this.spring.getContext().getBean(MetaAnnotationService.class);
+		assertThat(service.getIdPath("uid")).isEqualTo("uid");
+	}
+
 	private static Consumer<ConfigurableWebApplicationContext> disallowBeanOverriding() {
 		return (context) -> ((AnnotationConfigWebApplicationContext) context).setAllowBeanDefinitionOverriding(false);
 	}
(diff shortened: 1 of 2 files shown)


jzheaux:
Thanks, @kse-music! I think this is just an oversight. All legal annotation parameter types should work by default. 

Can you rebase this PR to `6.3.x` and then use a local `ConversionService` instance that includes a `ClassToString` converter by default?

kse-music:
@jzheaux I have added a local conversion service

core/src/main/java/org/springframework/security/authorization/method/AuthorizationAnnotationPropertyConversionService.java
```diff
@@ -0,0 +1,88 @@
...
+import org.springframework.core.convert.support.GenericConversionService;
+
+/**
+ * A {@link ConversionService} configured with converters that provide class to string
+ * conversion
+ *
+ * @author DingHao
+ * @since 6.3
+ */
+public final class AuthorizationAnnotationPropertyConversionService extends GenericConversionService {
```
jzheaux, line 36: While I appreciate the value of creating a custom implementation as `oauth2-core` does, let's start a little simpler for now, especially since this is on a maintenance branch.

What if instead, inside of `AuthorizationAnnotationUtils` you added something similar to:

```java
private final static DefaultConversionService conversionService = new DefaultConversionService();
static {
    conversionService.addConverter(...);
}
```

And then the code would use the `conversionService` instance.

config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java
```diff
@@ -509,7 +509,8 @@ static class DispatcherServletRequestMatcher implements RequestMatcher {
 		public boolean matches(HttpServletRequest request) {
 			String name = request.getHttpServletMapping().getServletName();
 			ServletRegistration registration = this.servletContext.getServletRegistration(name);
-			Assert.notNull(registration, computeErrorMessage(this.servletContext.getServletRegistrations().values()));
+			Assert.notNull(registration,
```
jzheaux, line 512: It looks like this might have wandered in from a merge. Can you rebase on `6.3.x` to remove this change from your commit?

config/src/test/java/org/springframework/security/config/annotation/method/configuration/PrePostMethodSecurityConfigurationTests.java
```diff
@@ -998,6 +998,14 @@ public void adviseWhenPrePostEnabledThenEachInterceptorRunsExactlyOnce() {
 		verify(expressionHandler, times(4)).createEvaluationContext(any(Supplier.class), any());
 	}
 
+	@Test
```
jzheaux, line 1001: Since this is for a bug fix, will you please place a comment indicating the bug that this fixes? Like so:

```
// gh-12345
@Test
...
```

jzheaux (changes requested):
Thanks @kse-music! I've left my feedback inline.

core/src/main/java/org/springframework/security/authorization/method/AuthorizationAnnotationPropertyConversionService.java
kse-music: Done

kse-music:
@jzheaux All feedback has been completed

jzheaux (approved):

jzheaux:
Thanks, @kse-music! This is now merged into `6.3.x` and `main`

PR #15713, opened 2024-08-30
Avoid running the VerifyDependenciesVersionsTask and CheckExpecedBranchVersionTask when inputs are stable

xyloman:
Each plugin has been updated to use the local build cache / avoid unnecessary work. To accomplish this, the inputs for each task will be computed when the plugin runs instead of at configuration time. This allows, for the inputs to be properly fingerprinted by Gradle. In addition, each task now will produce an output file.

- In the case of the `VerifyDependenciesVersionsTask` These changes have avoided a call to the Central Maven Repository on each build saving time and network calls.
- In the case of the `CheckExpectedBranchVersionTask` savings are less but semantically ensure that the task is only executed when the branch name of the git repository changes—reducing the number of functions that run on every incremental build.

### DIFF
--- a/buildSrc/src/main/java/org/springframework/security/CheckExpectedBranchVersionPlugin.java
+++ b/buildSrc/src/main/java/org/springframework/security/CheckExpectedBranchVersionPlugin.java
@@ -16,15 +16,23 @@
 
 package org.springframework.security;
 
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-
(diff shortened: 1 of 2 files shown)


marcusdacoregio:
Thanks @xyloman, this has been merged into `6.2.x`, `6.3.x` and `main` via https://github.com/spring-projects/spring-security/commit/69231c58536bc787a7834044a280986509b4203b

Issue #14633, opened 2024-02-18
Support different OIDC issuer hostnames for frontend/backend endpoints


heruan:
**Expected Behavior**

When the OIDC provider uses different hostnames from frontend and backend endpoints, fetching metadata from the configure issuer hostname does not fail.

**Current Behavior**

If the frontend and backend hostnames differs when fetching metadata, a validation exception is thrown.

**Context**

Consider a Spring application running inside a Kubernetes cluster, and it needs to authenticate against an OIDC server inside the same cluster (say Keycloak). Hostnames used inside the cluster are different from the public ones, and this is supported by Keycloak (frontend and backend hostnames can be different).

The Spring app has this configuration:

```
spring.security.oauth2.client.provider.keycloak.issuer-uri: http://internal:8180/realms/foo
```

When fetching metadata from there, Keycloak returns:

```json
{
    "issuer": "https://external/realms/foo",
    "authorization_endpoint": "https://external/realms/foo/protocol/openid-connect/auth",
    "token_endpoint": "http://internal:8180/realms/foo/protocol/openid-connect/token",
    "introspection_endpoint": "http://internal:8180/realms/foo/protocol/openid-connect/token/introspect",
    "userinfo_endpoint": "http://internal:8180/realms/foo/protocol/openid-connect/userinfo",
    "end_session_endpoint": "https://external/realms/foo/protocol/openid-connect/logout",
    "...": "..."
}
```

As expected, the frontend endpoints use `https://external` and backend endpoints use `http://internal:8180`.

The problem is that when fetching metadata, Spring fails here:

https://github.com/spring-projects/spring-security/blob/9c6b5f90f7787404922450d9ec559415c55ec4c3/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java#L246-L248

Fetching metadata from the issuer is necessary since some endpoints are read only from metadata, e.g the `end_session_endpoint`:

https://github.com/spring-projects/spring-security/blob/9c6b5f90f7787404922450d9ec559415c55ec4c3/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/web/logout/OidcClientInitiatedLogoutSuccessHandler.java#L79-L80

But also other back-channel endpoints without configuration properties in `application.yaml`.

Since I know both internal and external hostnames, how can I make Spring Security fetch metadata from the internal issuer hostname and accept the external hostname in metadata?

[...]

PR #15716, opened 2024-08-31
Allow building a ClientRegistration from provided configuration

heruan:
This adds the `fromOidcConfiguration` method to `ClientRegistrations` to build a `ClientRegistration` from a map representation of an OpenID Provider Configuration Response.

This is useful when the OpenID Provider Configuration is not available at a well-known location, or if custom validation is needed for the issuer location (e.g. if the issuer is only reachable via a back-channel URI that is different from the issuer value in the configuration).

Example usage:

```java
RequestEntity<Void> request = RequestEntity.get(metadataEndpoint).build();
ParameterizedTypeReference<Map<String, Object>> typeReference = new ParameterizedTypeReference<>() {};
Map<String, Object> configuration = rest.exchange(request, typeReference).getBody();
// Validate configuration.get("issuer") as per in the OIDC specification
ClientRegistration registration = ClientRegistrations.fromOidcConfiguration(configuration)
    .clientId("client-id")
    .clientSecret("client-secret")
    .build();
```

Fixes: gh-14633

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
@@ -73,6 +73,46 @@ public final class ClientRegistrations {
 	}
 
 	/**
+	 * Creates a {@link ClientRegistration.Builder} using the provided map representation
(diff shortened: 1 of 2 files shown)


rwinch (approved):

rwinch:

Issue #14575, opened 2024-02-10
Format check fails on fresh repo checkout


willemvd:
**Describe the bug**
After checking out the repository I want to run the `./gradlew format check` to make sure everything is ok before starting a new PR like described in the contribution documentation, but it fails with 

```
$ ./gradlew format check
> Task :spring-security-aspects:checkFormatTest FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Some problems were found with the configuration of task ':spring-security-aspects:checkFormatTest' (type 'CheckFormat').
  - Gradle detected a problem with the following location: '/Users/.../spring-security/aspects/src/test/java'.

    Reason: Task ':spring-security-aspects:checkFormatTest' uses this output of task ':spring-security-aspects:formatTest' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.

    Possible solutions:
      1. Declare task ':spring-security-aspects:formatTest' as an input of ':spring-security-aspects:checkFormatTest'.
      2. Declare an explicit dependency on ':spring-security-aspects:formatTest' from ':spring-security-aspects:checkFormatTest' using Task#dependsOn.
      3. Declare an explicit dependency on ':spring-security-aspects:formatTest' from ':spring-security-aspects:checkFormatTest' using Task#mustRunAfter.

    For more information, please refer to https://docs.gradle.org/8.6/userguide/validation_problems.html#implicit_dependency in the Gradle documentation.
  - Gradle detected a problem with the following location: '/Users/.../spring-security/aspects/src/test/aspectj'.

    Reason: Task ':spring-security-aspects:checkFormatTest' uses this output of task ':spring-security-aspects:formatTest' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.

    Possible solutions:
      1. Declare task ':spring-security-aspects:formatTest' as an input of ':spring-security-aspects:checkFormatTest'.
      2. Declare an explicit dependency on ':spring-security-aspects:formatTest' from ':spring-security-aspects:checkFormatTest' using Task#dependsOn.
      3. Declare an explicit dependency on ':spring-security-aspects:formatTest' from ':spring-security-aspects:checkFormatTest' using Task#mustRunAfter.

    For more information, please refer to https://docs.gradle.org/8.6/userguide/validation_problems.html#implicit_dependency in the Gradle documentation.
```

[...]

PR #15717, opened 2024-08-31
Fix adding more implied roles in the RoleHierarchy Builder.

nielsbasjes:
_[ As requested by @marcusdacoregio in this [comment](https://github.com/spring-projects/spring-security/pull/15272#issuecomment-2317889390) ]_

When defining a RoleHierarchy you can use the text format and the Builder model.

In the text format you can specify
```
   ROLE_A > ROLE_B
   ROLE_A > ROLE_C
```
which gives 'A' both 'B' and 'C'

The same specification in the Builder
```
RoleHierarchyImpl.withDefaultRolePrefix()
    .role("A").implies("B")
    .role("A").implies("C")
    .build();

```
which gives 'A' only 'C'

Rootcause: doing `implies` wipes any previous specifications for the same source role.

This fixes that and makes the Builder behave the same as the text format.

### DIFF
--- a/core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImpl.java
+++ b/core/src/main/java/org/springframework/security/access/hierarchicalroles/RoleHierarchyImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


nielsbasjes:
Apologies for putting in merge request with a checkstyle problem.
I tried to check the format locally but that failed. 
Now I know this is a problem others have already reported https://github.com/spring-projects/spring-security/issues/14575

marcusdacoregio:
Thanks @nielsbasjes, this is now merged into `main`.

I am considering this as an `enhancement` for now, but we can consider backporting it if we receive feedback about users needing it in older versions.

PR #15728, opened 2024-09-03
Corrected AuthenticationEventPublisher Type

ximinghui:
The [AuthenticationEventPublisher](https://docs.spring.io/spring-security/site/docs/6.3.3/api/org/springframework/security/authentication/AuthenticationEventPublisher.html) interface does not provide the ```setAdditionalExceptionMappings(Map<Class<? extends AuthenticationException)``` method, and it is located in the [DefaultAuthenticationEventPublisher](https://docs.spring.io/spring-security/site/docs/6.3.3/api/org/springframework/security/authentication/DefaultAuthenticationEventPublisher.html) implementation class.

### DIFF
--- a/docs/modules/ROOT/pages/servlet/authentication/events.adoc
+++ b/docs/modules/ROOT/pages/servlet/authentication/events.adoc
@@ -107,7 +107,7 @@ public AuthenticationEventPublisher authenticationEventPublisher
     Map<Class<? extends AuthenticationException>,
         Class<? extends AbstractAuthenticationFailureEvent>> mapping =
             Collections.singletonMap(FooException.class, FooEvent.class);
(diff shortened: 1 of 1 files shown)


ximinghui:
Since Spring Security 6.3.3 uses Java 17, the variable type declaration can also be changed to the ```var``` keyword available since Java 10+. However, I don’t see the ```var``` keyword used in the documentation, so I continue to use traditional type declarations to maintain consistency in the code style.

jzheaux:
Thanks, @ximinghui! You are correct that Spring Security does not use `var`. That is because we try and align ourselves with khttps://github.com/spring-projects/spring-framework/wiki/Code-Style#local-variable-type-inference.

PR #15729, opened 2024-09-03
Add Security Hint AOT Processing

jzheaux:
This API allows for hint registration based on Spring Security's infrastructural beans

There are two built-in implementations:

1. `AuthorizeReturnObjectCoreHintsRegistrar` traverses the object graph of all Spring beans for methods that use `@AuthorizeReturnObject`. It creates the initial proxy class for each of those return values. Also, it uses each return value as a launch point to find other objects that use `@AuthorizeReturnObject`.

2. `AuthorizeReturnObjectDataHintsRegistrar` traverses Spring Data repository factory beans since each repository's primary domain object is often hidden behind generic erasure. It thereafter traverses any object it finds in the same way as 1.

Any return objects that were missed by the scan can be added by the application in the following way:

```java
@Bean 
static SecurityHintsRegistrar alsoProxy(AuthorizationProxyFactory proxyFactory) {
    List<Class<?>> classes = List.of(MyExtraClass.class, MyOtherExtraClass.class);
    return new AuthorizeReturnObjectHintsRegistrar(proxyFactory, classes);
}
```

or the application can add a custom `SecurityHintsRegistrar`.

There is also a sample application available here: https://github.com/jzheaux/spring-security-samples/tree/proxy-hints-using-pr/servlet/spring-boot/java/data

### DIFF
--- a/core/src/main/java/org/springframework/security/aot/hint/SecurityHintsAotProcessor.java
+++ b/core/src/main/java/org/springframework/security/aot/hint/SecurityHintsAotProcessor.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2002-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.aot.hint;
+
+import org.springframework.aot.generate.GenerationContext;
+import org.springframework.beans.factory.aot.BeanFactoryInitializationAotContribution;
+import org.springframework.beans.factory.aot.BeanFactoryInitializationAotProcessor;
+import org.springframework.beans.factory.aot.BeanFactoryInitializationCode;
+import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
+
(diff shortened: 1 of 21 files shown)


core/src/main/java/org/springframework/security/aot/hint/SecurityHintsAotProcessor.java
```diff
@@ -0,0 +1,49 @@
...
+
+package org.springframework.security.aot.hint;
+
+import org.springframework.aot.generate.GenerationContext;
+import org.springframework.beans.factory.aot.BeanFactoryInitializationAotContribution;
+import org.springframework.beans.factory.aot.BeanFactoryInitializationAotProcessor;
+import org.springframework.beans.factory.aot.BeanFactoryInitializationCode;
+import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
+
+public final class SecurityHintsAotProcessor implements BeanFactoryInitializationAotProcessor {
```
marcusdacoregio, line 25: Can this class be package-private?

data/src/main/java/org/springframework/security/data/aot/hint/AuthorizeReturnObjectDataHintsRegistrar.java
```diff
@@ -0,0 +1,100 @@
...
+				continue;
+			}
+			Class<?> repository = generics[0];
+			for (Method method : repository.getDeclaredMethods()) {
+				AuthorizeReturnObject returnObject = this.scanner.scan(method, repository);
+				if (returnObject == null) {
+					continue;
+				}
+				// optimistically assume that the entity needs wrapping if any of the
+				// repository methods use @AuthorizeReturnObject
```
marcusdacoregio, line 69: What about projections? Do they get registered too? Users could return a different object, not just the entity:

```java
interface UserRepository extends CrudRepository<User, Long> {

  @Query("...")
  UserProjection getUserWithAdditionalData(Long id);

}
```

marcusdacoregio (changes requested):
That's so cool @jzheaux. I've left some minor feedback inline.

marcusdacoregio (approved):

Issue #2253, opened 2012-08-08
SEC-2028: Support remember me and concurrency control


spring-projects-issues:
[Rob Winch](https://jira.spring.io/secure/ViewProfile.jspa?name=rwinch) (Migrated from [SEC-2028](https://jira.spring.io/browse/SEC-2028?redirect=false)) said:

Currently the concurrency control does not work with remember me. It would be good to support this out of the box.

PR #15748, opened 2024-09-06
Use SessionAuthenticationStrategy for Remember-Me authentication

xhaggi:
As described in gh-2253, the `RememberMeAuthenticationFilter` does not call the configured `SessionAuthenticationStrategy` if there is one, and therefore concurrent session control does not work properly.

This PR adapts the `RememberMeAuthenticationFilter` so that it calls the `SessionAuthenticationStrategy` on successful authentication as the `AbstractAuthenticationProcessingFilter` does. The shared `SessionAuthenticationStrategy` is then used in the `RememberMeConfigurer` to configure the filter like it is done in `FormLoginConfigurer`.

Let me know if anything is missing!

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/RememberMeConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/RememberMeConfigurer.java
@@ -33,6 +33,7 @@ import org.springframework.security.web.authentication.rememberme.PersistentToke
 import org.springframework.security.web.authentication.rememberme.PersistentTokenRepository;
 import org.springframework.security.web.authentication.rememberme.RememberMeAuthenticationFilter;
 import org.springframework.security.web.authentication.rememberme.TokenBasedRememberMeServices;
+import org.springframework.security.web.authentication.session.SessionAuthenticationStrategy;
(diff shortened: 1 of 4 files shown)


jzheaux (changes requested):
Thanks for the PR, @xhaggi! Could you add a unit test that confirms the desired behavior (concurrency control) works, for example when the DSL configures a max number of sessions and the remember me filter exceeds that?

xhaggi:
@jzheaux of course. I have added the unit test.

jzheaux (approved):

xhaggi:
@jzheaux the [build](https://github.com/spring-projects/spring-security/actions/runs/11121314806/job/31517224220) failed because formatting violations were found in the files I changed. I fixed it and force-pushed the changes.

xhaggi:
BTW is there any chance that this will be backported to earlier versions?

jzheaux:
Hi, @xhaggi, no, it's not likely, given that there is a behavioral change and a new public API. That said, what version were you hoping to backport it to and what is preventing you from upgrading?

xhaggi:
>what version were you hoping to backport

I was thinking of the last stable version 6.3.x.

>what is preventing you from upgrading

Nothing, but we have to wait for the next stable 😉

Issue #15778, opened 2024-09-10
Where should i config OAuth2LoginAuthenticationFilter with property authenticationResultConverter?


mengxzh:
![image](https://github.com/user-attachments/assets/149ea8e4-2ff6-4bb5-85ed-f48b1619189a)

This filter init and configure in OAuth2LoginConfigurer 
In both init and configure method, no set this property

![image](https://github.com/user-attachments/assets/54c93750-9ed2-40a8-9610-93aee4cbfa2b)
![image](https://github.com/user-attachments/assets/cd19d6ea-d8f0-4acd-b58e-d4e05c20f8fc)

How should i config this property?

mengxzh:
Spring-Security-Config Version: 6.3.1

mengxzh:
Which MeilStone could merge above merge reuqest?

jzheaux:
You can use an `ObjectPostProcessor` like so:

```java
http
    .oauth2Login((oauth2) -> oauth2
        .addObjectPostProcessor(new ObjectPostProcessor<OAuth2LoginAuthenticationFilter>() {
		@Override
		public <O extends OAuth2LoginAuthenticationFilter> O postProcess(O object) {
			object.setAuthenticationResultConverter(myAuthenticationResultConverter);
			return object;
		}
	})
```

The DSL doesn't expose everything, just the most common configurations. `addObjectPostProcessor` is there to allow you to see the unexposed values.

I'm going to close this as answered, however please feel free to post more comments if it seems I'm misunderstood.

dirkniblick:
Thank you @jzheaux for posting your solution. I am using Spring Cloud Azure (i.e., Active Directory, Entra) to add single sign on (SSO) to my Spring Boot 3.4 web app. I searched high and low for a way to convert the `DefaultOidcUser` principal provided by `AadOAuth2UserService` to my own (and more useful) `UserDetails` and found nothing, until I found this page. I feel like this should be easier to do, but for anybody who wants to do this too, here's my solution:

```java
@Configuration
@EnableWebSecurity
public class WebSecurityConfig {

    @Bean
    public SecurityFilterChain securityFilterChainBean(HttpSecurity http) throws Exception {
        http.with(AadWebApplicationHttpSecurityConfigurer.aadWebApplication(), o -> {})
            .authorizeHttpRequests(requests -> requests
                    .requestMatchers(HttpMethod.OPTIONS).permitAll()
[...]

PR #15779, opened 2024-09-10
OAuth2LoginConfigurer support set authenticationResultConverter

kse-music:
Closes gh-15778


jzheaux:
Thanks for the PR, @kse-music. I'd prefer to have the individual use an `ObjectPostProcessor` for now. If authentication result converter becomes a very common requested configuration, we can take another look at adding it to the DSL.

PR #15781, opened 2024-09-10
Fix getBeansWithName in global authentication configurers

Kehrlann:
There was a small bug with `getBeansWithName` in both `InitializeAuthenticationProviderBeanManagerConfigurer` and `InitializeUserDetailsBeanManagerConfigurer`, where we returned a list containing n copies of the same bean.

This did not have any functional impact as we only ever use the list when there is only one element in it.

---

Please merge forward to main.


rwinch (approved):

rwinch:

PR #15791, opened 2024-09-11
Remove trailing spaces in default UIs

Kehrlann:
Following gh-15723, default UIs had blank lines with only spaces. These get deleted by the spring-javaformat plugin. In order to avoid this behavior, an extra `\s` had been inserted in the tests. The reason for those `\s` is not obvious.

This commit cleans up the `\s` but changing the HTML templates.


rwinch:

PR #15794, opened 2024-09-11
Provide Runtime Hints for Beans used in Pre/PostAuthorize Expressions

marcusdacoregio:
Closes gh-14652


jzheaux (changes requested):
So exciting, @marcusdacoregio. I've left a piece of feedback inline.

Also, it's interesting that this class requires searching the object graph recursively as well. I wonder if there is a better abstraction that could work for the handful of use cases that we have.

For example, traversing hierarchies often lends itself to Visitor pattern and following such may unify these currently disparate classes.

There are other things in common about the use cases as well:

1. There will always be unreachable objects, so there needs to be a way to provide those manually
2. Spring Data-specific awareness would increase reachability

For now, I think this is a fine start, but perhaps a ticket to explore an improved abstraction for these cross-cutting concerns would be beneficial.

jzheaux (approved):

PR #15804, opened 2024-09-13
CVE-2024-29857, CVE-2024-34447 & CVE-2023-2976 : 

gaetan-deltombe:
The CVE-2024-29857 and CVE-2024-34447 are fixed, except for "spring-security-cas".

because of the dependency "org.apereo.cas.client:cas-client-core", the highest available version is 4.0.4.

CVE-2023-2976 : The dependency opensaml-core:4.3.2 bring a vulnerability with guava:31.1-jre, I override the dependency using the 32.0.1 version.

### DIFF
--- a/crypto/spring-security-crypto.gradle
+++ b/crypto/spring-security-crypto.gradle
@@ -3,8 +3,8 @@ apply plugin: 'io.spring.convention.spring-module'
 dependencies {
 	management platform(project(":spring-security-dependencies"))
 	optional 'org.springframework:spring-jcl'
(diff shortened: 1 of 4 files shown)


jzheaux:
Thanks for the PR, @gaetan-deltombe. Unfortunately, `6.1.x` is [no longer supported as of May 2024](https://spring.io/projects/spring-security#support), so it wouldn't do any good to update these since there will be no more OSS releases along this line.

Please feel free to check supported lines for dependency upgrades that you need.

Issue #15818, opened 2024-09-17
ServerBearerTokenAuthenticationConverter does not support form encoded body parameter


jonah1und1:
**Describe the bug**
When using opaque tokens with an OAuth 2 resource server, webflux's `ServerBearerTokenAuthenticationConverter` does not support authentication via url-encoded POST requests with their access token as a body parameter.  
[RFC-6750 Sec. 2.2](https://datatracker.ietf.org/doc/html/rfc6750#section-2.2) describes this behaviour and it is supported by `DefaultBearerTokenResolver` in mvc.

**To Reproduce**
Create a POST endpoint for which authentication with an OAuth 2 resource server with an opaque token is needed. Doing so with the reactive stack does not allow for authentication via a body parameter.  

**Expected behavior**
Providing a body parameter with a valid access token named `access_token` should authenticate the request.  

**Sample**
A sample project with designated test cases for the mvc and reactive stack can be found here:
https://github.com/jonah1und1/spring-security-rfc6750-2.2

jonah1und1:
Any updates on this?  
I happily improve on the issue description and/or pull request if needed.

sjohnr:
@jonah1und1 thanks for checking. We am working on a few high priority items for the release and then I will be reviewing your PR. If you would like to get a head start, please check the PR for modern java features from Java 11+ and switch back to corresponding Java 8 features. We are not yet ready to update most of the source code to Java 17 features because we are following Spring Framework's lead on which language features we are using.

jonah1und1:
@sjohnr Thank you. I will do that.

jonah1und1:
@sjohnr Did you have any time to take a look at the PR yet?

PR #15819, opened 2024-09-17
Add support for access token in body parameter as per rfc 6750 Sec. 2.2

jonah1und1:
Currently, the reactive stack does not allow for authentication via parameter in body of post requests.  
[RFC-6750 Sec. 2.2](https://datatracker.ietf.org/doc/html/rfc6750#section-2.2) allows this. It is also support by mvc stack.  
  
Related ticket: gh-15818.


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/server/authentication/ServerBearerTokenAuthenticationConverter.java
```diff
@@ -122,25 +147,69 @@ public void setBearerTokenHeaderName(String bearerTokenHeaderName) {
 		this.bearerTokenHeaderName = bearerTokenHeaderName;
 	}
 
-	private String resolveFromAuthorizationHeader(HttpHeaders headers) {
+	/**
+	 * Set if transport of access token using form-encoded body parameter is supported.
+	 * Defaults to {@code false}.
+	 * @param allowFormEncodedBodyParameter if the form-encoded body parameter is
+	 * supported
```
sjohnr, line 154: Please add `@since 6.5` tag to note when this [method was added](https://github.com/spring-projects/spring-security/blob/main/CONTRIBUTING.adoc#submit-a-pull-request).

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/server/authentication/ServerBearerTokenAuthenticationConverter.java
```diff
@@ -65,38 +75,53 @@ public Mono<Authentication> convert(ServerWebExchange exchange) {
...
-		if (parameterToken != null && isParameterTokenSupportedForRequest(request)) {
-			return parameterToken;
-		}
-		return null;
+	private Mono<String> token(ServerWebExchange exchange) {
+		final ServerHttpRequest request = exchange.getRequest();
+
+		return Flux.merge(resolveFromAuthorizationHeader(request.getHeaders()).map(s -> Tuples.of(s, TokenSource.HEADER)),
+						  resolveAccessTokenFromRequest(request).map(s -> Tuples.of(s, TokenSource.QUERY_PARAMETER)),
+						  resolveAccessTokenFromBody(exchange).map(s -> Tuples.of(s, TokenSource.BODY_PARAMETER)))
```
sjohnr, line 83: Would it be possible to resolve without the creation of `Tuple`s here? It is generally preferred to avoid creation of additional objects when possible (e.g. `Stream`) and I think `Tuple` would fall in the same category here. Instead of the `TokenSource`, I believe `Mono#switchIfEmpty` could be used.

sjohnr (changes requested):
Thanks much for the PR @jonah1und1. Apologies for the delay. I have some feedback inline. Please note that I will hold off merging until after the GA release, since we are in the RC phase of 6.4 and will need to wait to introduce new APIs (`setAllowFormEncodedBodyParameter()`) until 6.5.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/server/authentication/ServerBearerTokenAuthenticationConverter.java
jonah1und1: I am sorry, I fail to see how to resolve this without some kind of information where the token is from, to then check whether it is supported. Could you give me an hint on how to use `Mono#switchIfEmpty` here?

sjohnr: My mistake. Since you're using `Flux.merge()`, it would actually be `Flux#switchIfEmpty`.

The new `resolveAccessTokenFromBody()` checks a `boolean`, method and content-type before returning a `Mono`. The `resolveAccessTokenFromRequest()` method can be refactored to do the same. In that case, `Flux.merge()` will return some number of tokens (0-3) that don't require checking the source again. It would end up looking something like this:

```java
@Override
public Mono<Authentication> convert(ServerWebExchange exchange) {
	return Mono.defer(() -> token(exchange));
}

private Mono<Authentication> token(ServerWebExchange exchange) {
	ServerHttpRequest request = exchange.getRequest();
	return Flux.merge(resolveFromAuthorizationHeader(request.getHeaders()),
			resolveAccessTokenFromRequest(exchange.getRequest()),
			resolveAccessTokenFromBody(exchange))
		.switchIfEmpty(Mono.error(new OAuth2AuthenticationException(invalidTokenError())))
		.collectList()
		.flatMap((accessTokens) -> {
			if (accessTokens.size() > 1) {
				BearerTokenError error = BearerTokenErrors
					.invalidRequest("Found multiple bearer tokens in the request");
				return Mono.error(new OAuth2AuthenticationException(error));
			}

			return Mono.just(new BearerTokenAuthenticationToken(accessTokens.get(0)));
		});
}
```

jonah1und1: Thank you for your answer!
Unfortunately, I think there might be two problems with this approach:

Firstly, if no token can be found the current implementation of `ServerBearerTokenAuthenticationConverter#convert()` returns an empty mono. An `invalidTokenError()` is only raised if the String containing the token is empty, but not if it's `null`.
This could be mitigated with relative ease imo, by leaving `convert()` in its current state and replacing `switchIfEmpty()` by a second if-clause inside of `.flatMap()`:
```java
 if (accessTokens.isEmpty()) {
    return Mono.empty();
}
```

Secondly, the tests `resolveWhenValidHeaderIsPresentTogetherWithQueryParameterThenAuthenticationExceptionIsThrown()` and `resolveWhenQueryParameterHasMultipleAccessTokensThenOAuth2AuthenticationException()` fail because they do not set `allowUriQueryParameter` to `true` first. While I am not quite sure about the expected behaviour here and whether adding `this.converter.setAllowUriQueryParameter(true)` to both tests is acceptable, doing so might change the current behaviour.
`resolveWhenValidQueryParameterIsPresentTogetherWithBodyParameterThenAuthenticationExceptionIsThrown()`, which was added by me, also fails because the query parameter access token is filtered due to the `POST`-request.

sjohnr: > Firstly, if no token can be found the current implementation of `ServerBearerTokenAuthenticationConverter#convert()` returns an empty mono. An `invalidTokenError()` is only raised if the String containing the token is empty, but not if it's `null`.

Hmm, fair point.

> Secondly, the tests [...] do not set `allowUriQueryParameter` to `true` first.

Also a fair point. This is unfortunate, because it could be argued that leaving this false implies that such parameters should be ignored. The test and original implementation don't agree though and that makes this tricky. I think this comes down to whether it would be considered a bug to read the token from the request (e.g. query string or body parameters) and validate it when it is not explicitly enabled.

The [spec](https://datatracker.ietf.org/doc/html/rfc6750#section-2) states that

> Resource servers MAY support this method.

for both body parameters and query string parameters. I don't see anything in the [Error Codes](https://datatracker.ietf.org/doc/html/rfc6750#section-3.1) section indicating that the request must be validated for these problems if the server doesn't support a particular method for resolving the token.

For those reasons, I'm inclined to open an bug to necessitate a change in behavior for both servlet and reactive implementations that would be a precursor to this enhancement. What do you think?

jonah1und1: Sorry for the late reply!
Sounds sound to open a dedicated bug ticket for that. Do you want me to implement a PR for that?

sjohnr: @jonah1und1 I opened gh-16038. I think it would be nice to get this fixed in the release this month. Feel free to open a PR if you have availability to work on it, or if you're busy let me know and I can submit a fix.

jonah1und1: I created a PR: https://github.com/spring-projects/spring-security/pull/16039

Issue #15771, opened 2024-09-09
OidcBackChannelLogoutTokenValidator should not throw an NPE when issuer is missing


jzheaux:
Related to https://github.com/spring-projects/spring-boot/issues/42172#issue-2510291998, when `ClientRegistration` does not have an `issuerUri` configured, it should give a better error than an NPE.

This needs to be addressed for both the imperative and reactive versions of the class.

Please base this change off the `6.2.x` branch so it can be forward-ported when merged.

wapkch:
Hi @jzheaux , may i work on this?

jzheaux:
Yes, please. :) Thanks for offering, the issue is yours.

PR #15824, opened 2024-09-18
OidcBackChannelLogoutTokenValidator should not construct when missing OIDC Provider Issuer

wapkch:
Closes gh-15771

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OidcBackChannelLogoutTokenValidator.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OidcBackChannelLogoutTokenValidator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/client/OidcBackChannelLogoutTokenValidator.java
```diff
@@ -77,6 +78,9 @@ else if (events.get(BACK_CHANNEL_LOGOUT_EVENT) == null) {
 		if (issuer == null) {
 			errors.add(invalidLogoutToken("iss claim must not be null"));
 		}
+		else if (!StringUtils.hasText(this.issuer)) {
```
jzheaux, line 81: I can see the value of doing it here. 

However, given that we already know this is invalid in the constructor, I'd prefer the check be there instead, preventing the object from getting constructed in an invalid state.

Perhaps something like this:

```java
String issuer = clientRegistration.getProviderDetails.getIsserUri();
Assert.hasText(issuer, "provider issuer cannot be null");
this.issuer = issuer;
...
```

jzheaux: Also, that would more clearly indicate to the application that it should throw a 5xx instead of a 4xx, indicating that it is likely a server configuration problem, not a user request problem.

jzheaux (changes requested):
Thanks for the PR, @wapkch! I left a piece of feedback inline. Please see if you can apply it also to the reactive implementation and update the tests accordingly.

jzheaux (approved):

jzheaux:
Thanks again, @wapkch, this is now merged into `6.2.x`, `6.3.x`, and `main`

PR #15827, opened 2024-09-19
Complete HttpSecurity Deprecation notices

rs017991:
This is a follow-up to https://github.com/spring-projects/spring-security/pull/13010

At that time, existing deprecation notices were upgraded to 'for removal', yet:
1. The annotation itself was not updated to reflect 'for removal'
2. Further, they both suggest migrating to a method that is _also_ deprecated and slated for removal.

Those two issues combine to create a humorous situation where a deprecation _**warning**_ turns into a deprecation _**error**_ if you follow the advice.


rwinch:

PR #15832, opened 2024-09-20
Polish ExpressionTemplateSecurityAnnotationScanner

kse-music:
Since `UniqueSecurityAnnotationScanner `has cached annotations internally


jzheaux:

Issue #15735, opened 2024-09-04
 Consider adding OneTimeTokenService for a clustered environment


CrazyParanoid:
It would be nice to implement session based `OneTimeTokenService`. Now this is difficult to do, because `OneTimeTokenService` accepts `GenerateOneTimeTokenRequest`. It is better to change the design of the `generate` method - replace GenerateOneTimeTokenRequest with HttpServletRequest.
It is not yet clear how you can call `consume` for an http session.

CrazyParanoid:
Maybe the best solution would be to split this component into two? For example `OneTimeTokenGenerator` and `OneTimeTokenRepository`.
Then
```
OneTimeToken ott = this.oneTimeTokenGenerator.generate(generateRequest);
this.oneTimeTokenGenerator.save(ott);
```
With this implementation it is much easier to ensure `OneTimeTokenRepository` works in a cluster environment.

marcusdacoregio:
Hi @CrazyParanoid, thanks for the report.

Can you elaborate more on your use case? Would you like to authenticate the session that started the request instead of the session that submits the token?

An alternative that I had in mind is to provide a `GenerateOneTimeTokenRequestResolver` that can be used in the `GenerateOneTimeTokenFilter`, something like:

```java
public interface GenerateOneTimeTokenRequestResolver {

    GenerateOneTimeTokenRequest resolve(HttpServletRequest request);

}
```

This way you can return a subclass of `GenerateOneTimeTokenRequest` with the appropriate information.

CrazyParanoid:
Hi @marcusdacoregio 

> An alternative that I had in mind is to provide a GenerateOneTimeTokenRequestResolver that can be used in the GenerateOneTimeTokenFilter, something like:

This may be a good solution, but I would first look at the design of the `OneTimeTokenService`. It seems to me that two components for generation a token and storing it would give much more flexibility in the API, as well as the single responsibility principle. 
Overall, separating the token generation logic from its storage will make the API simpler, cleaner, and easier to use.

CrazyParanoid:
> Can you elaborate more on your use case? Would you like to authenticate the session that started the request instead of the session that submits the token?

I'm more interested in the ability of the `OneTimeTokenService` to work in a distributed environment. Maybe `JbcOneTimeTokenService` or `CachedOneTimeTokenService`? Something like this.

marcusdacoregio:
[...]

PR #15842, opened 2024-09-24
Add support JdbcOneTimeTokenService

CrazyParanoid:
Closes gh-15735

### DIFF
--- a/core/src/main/java/org/springframework/security/authentication/ott/JdbcOneTimeTokenService.java
+++ b/core/src/main/java/org/springframework/security/authentication/ott/JdbcOneTimeTokenService.java
@@ -1,0 +1,239 @@
+/*
+ * Copyright 2002-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.authentication.ott;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Timestamp;
+import java.sql.Types;
+import java.time.Clock;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+import java.util.function.Function;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import org.springframework.jdbc.core.ArgumentPreparedStatementSetter;
+import org.springframework.jdbc.core.JdbcOperations;
+import org.springframework.jdbc.core.PreparedStatementSetter;
+import org.springframework.jdbc.core.RowMapper;
+import org.springframework.jdbc.core.SqlParameterValue;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
+import org.springframework.scheduling.support.CronTrigger;
+import org.springframework.util.Assert;
+import org.springframework.util.CollectionUtils;
+import org.springframework.util.StringUtils;
+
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/springframework/security/authentication/ott/OneTimeTokenUtils.java
```diff
@@ -0,0 +1,55 @@
...
+import org.springframework.util.Assert;
+
+/**
+ *
+ * Utility for default generation and checking of {@link OneTimeToken} time to live.
+ *
+ * @author Max Batischev
+ * @since 6.4
+ */
+public final class OneTimeTokenUtils {
```
rwinch, line 32: Let's keep this package scope, alternatively it might be worth just having some duplicated logic in the Repository instances at least for now. It isn't a lot and we can always extract it out later. It also will allow us to avoid a static Clock (see my comment on the `clock`.

core/src/main/java/org/springframework/security/authentication/ott/OneTimeTokenUtils.java
```diff
@@ -0,0 +1,55 @@
...
+
+	public static long DEFAULT_ONE_TIME_TOKEN_TIME_TO_LIVE = 300;
+
+	private static Clock clock = Clock.systemUTC();
```
rwinch, line 36: I don't view it as ideal to have a static Clock for a few reasons.

It makes it easier to have errors when setting it. 

1) For example, the tests right now only set this before a test is ran. However, some tests override the value. All future tests that use the Clock must also set it. Alternatively, we could also set after the tests run
2) The tests that are setting the value must know what to set the Clock to. Any changes to the Clock would likely need to be made in all of the tests.

Ensuring we are using an instance variable (perhaps on the Repository itself) would avoid all of this. It would also allow us to potentially expose the Clock on the Repository to be set in the future.

core/src/main/java/org/springframework/security/authentication/ott/JdbcOneTimeTokenService.java
```diff
@@ -0,0 +1,198 @@
...
+ *
+ * <p>
+ * <b>NOTE:</b> This {@code JdbcOneTimeTokenService} depends on the table definition
+ * described in "classpath:org/springframework/security/core/ott/jdbc/one-time-tokens.sql"
+ * and therefore MUST be defined in the database schema.
+ *
+ * @author Max Batischev
+ * @since 6.4
+ */
+public final class JdbcOneTimeTokenService implements OneTimeTokenService {
```
rwinch, line 49: If a token is never consumed, then it is never deleted even after it expires. We probably need a way to ensure that expired tokens are cleaned up. Likely this would need to be done using something like a `TaskScheduler`. For an example of what it might look like, you can refer to [JdbcIndexedSessionRepository](https://github.com/spring-projects/spring-session/blob/3.3.2/spring-session-jdbc/src/main/java/org/springframework/session/jdbc/JdbcIndexedSessionRepository.java) in Spring Session.

core/src/main/resources/org/springframework/security/core/ott/jdbc/one-time-tokens.sql
```diff
@@ -0,0 +1,6 @@
+create table one_time_tokens
```
rwinch, line 1: Can you please name this one-time-tokens-schema.sql to align with the current conventions in Spring Security?

rwinch (changes requested):
Thank you for the PR @CrazyParanoid!

I've provided feedback inline.

core/src/main/java/org/springframework/security/authentication/ott/OneTimeTokenUtils.java
CrazyParanoid: Yes, you are right. Not the best solution. I will return the logic for generating and checking the token from the `InMemoryOneTimeTokenService`, I will need to think about how to extract it in the next iterations.

CrazyParanoid:
Hi @rwinch! All your comments have been resolved

rwinch:
Thanks for the Pull Request! This is now merged into main via https://github.com/spring-projects/spring-security/commit/1dd79c379b73f958147a70d4175934bd07372967 :smile:

I added a few commits to clean things up a bit. You can see a list of the commits in https://github.com/spring-projects/spring-security/issues/15735 or by viewing the commits merged in 1dd79c379b73f958147a70d4175934bd07372967

PR #15845, opened 2024-09-24
Fix Broken Resource Server Doc Links

ngocnhan-tran1996:


docs/modules/ROOT/pages/servlet/oauth2/resource-server/jwt.adoc
```diff
@@ -974,6 +974,7 @@ Java::
 public JwtAuthenticationConverter jwtAuthenticationConverter() {
     JwtGrantedAuthoritiesConverter grantedAuthoritiesConverter = new JwtGrantedAuthoritiesConverter();
     grantedAuthoritiesConverter.setAuthoritiesClaimName("authorities");
+    grantedAuthoritiesConverter.setAuthoritiesClaimDelimiter(",");
```
jzheaux, line 977: While I agree that sometimes claims can be comma-delimited, that is not the use case that is set up by the description above. Unless otherwise specified, the documentation tries to stay with [the spec](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3) which states that `scope` is space-delimited.

As such, please leave these code snippet changes out for now.

jzheaux (changes requested):
Good catch on the broken links, @ngocnhan-tran1996! I've left some feedback inline.

Also, will you please change your commit message to be more descriptive as this helps future research when there are bugs?

You might have your commit title be something like "Fix Broken Resource Server Doc Links"

docs/modules/ROOT/pages/servlet/oauth2/resource-server/jwt.adoc
ngocnhan-tran1996: Thanks for your review

jzheaux (approved):

jzheaux:
Thanks again for your help here, @ngocnhan-tran1996. In the future, will you please create a branch for your PRs? This helps when it comes to me pushing polishes or rebasing your PR on a different branch before merging.

It's no problem this time; I can merge this, cherry-pick to an earlier version, and then forward-port the changes. Just something to remember for the future.

ngocnhan-tran1996:
> Thanks again for your help here, @ngocnhan-tran1996. In the future, will you please create a branch for your PRs? This helps when it comes to me pushing polishes or rebasing your PR on a different branch before merging.
> 
> It's no problem this time; I can merge this, cherry-pick to an earlier version, and then forward-port the changes. Just something to remember for the future.

I will create a new branch in the next contribute. Thanks for your advice

Issue #15702, opened 2024-08-27
Supporting logout+jwt for back-channel logout with spring-webflux


c1rd3cm:
**Expected Behavior**
We are using an IDP that sends back-channel logout request with a token type set as `logout+jwt` which should be accepted by spring-security when using spring-webflux.

**Current Behavior**
The back-channel is currently rejected because the only token type supported is `jwt`.

```java
Error has been observed at the following site(s):
    *__________Mono.error ⇢ at org.springframework.security.config.web.server.OidcBackChannelLogoutReactiveAuthenticationManager.lambda$decode$2(OidcBackChannelLogoutReactiveAuthenticationManager.java:92)
    *__Mono.onErrorResume ⇢ at org.springframework.security.config.web.server.OidcBackChannelLogoutReactiveAuthenticationManager.decode(OidcBackChannelLogoutReactiveAuthenticationManager.java:88)
    |_           Mono.map ⇢ at org.springframework.security.config.web.server.OidcBackChannelLogoutReactiveAuthenticationManager.authenticate(OidcBackChannelLogoutReactiveAuthenticationManager.java:80)
    |_           Mono.map ⇢ at org.springframework.security.config.web.server.OidcBackChannelLogoutReactiveAuthenticationManager.authenticate(OidcBackChannelLogoutReactiveAuthenticationManager.java:83)
    *________Mono.flatMap ⇢ at org.springframework.security.config.web.server.OidcBackChannelLogoutWebFilter.filter(OidcBackChannelLogoutWebFilter.java:90)

Caused by: com.nimbusds.jose.proc.BadJOSEException: JOSE header typ (type) logout+jwt not allowed
    at com.nimbusds.jose.proc.DefaultJOSEObjectTypeVerifier.verify(DefaultJOSEObjectTypeVerifier.java:148)
    at com.nimbusds.jwt.proc.DefaultJWTProcessor.process(DefaultJWTProcessor.java:378)
    at com.nimbusds.jwt.proc.DefaultJWTProcessor.process(DefaultJWTProcessor.java:340)
    at org.springframework.security.oauth2.jwt.NimbusReactiveJwtDecoder.createClaimsSet(NimbusReactiveJwtDecoder.java:292)
    at org.springframework.security.oauth2.jwt.NimbusReactiveJwtDecoder$JwkSetUriReactiveJwtDecoderBuilder.lambda$processor$13(NimbusReactiveJwtDecoder.java:449)
[...]

PR #15847, opened 2024-09-24
Allow logout+jwt JWT type for reactive

c1rd3cm:
The OIDC back-channel spec recommends using a logout token typ `logout+jwt` (see [here](https://openid.net/specs/openid-connect-backchannel-1_0-final.html#LogoutToken)).

Support of this type was recently added [on the Servlet side](https://github.com/spring-projects/spring-security/commit/9101bf1f7d5ad5d57ebfa2df78135704cdc19b23), so backporting the same on the reactive side to 
close the gap.

Closes gh-15702

### DIFF
--- a/config/src/main/java/org/springframework/security/config/web/server/OidcBackChannelLogoutReactiveAuthenticationManager.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/OidcBackChannelLogoutReactiveAuthenticationManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


jzheaux:
Thanks, @c1rd3cm! Before this gets merged, will you please update the commit message so that it ends with the following on its own line:

```
Closes gh-15702
```

This will allow the associated issue to be closed upon merge. It also makes it easier to perform research in the future when ascertaining why a certain change was made.

c1rd3cm:
Thanks for the review @jzheaux. Sorry, this is my first time submitting a PR, and I saw that I missed the reference to the issue in my commit but I thought that having it the PR would be enough. I updated my commit description and did a force push, I hope this is the right way.

jzheaux:
Looks great, @c1rd3cm. Thanks for the update.

jzheaux:
Thanks again, @c1rd3cm! This is now merged into `main`.

Issue #15846, opened 2024-09-24
Inconsistent ReactiveOAuth2UserService in OAuth2LoginSpec


blacelle:
**Describe the bug**
The way to configure a `ReactiveOAuth2UserService<OAuth2UserRequest, OAuth2User>` in  `OAuth2LoginSpec` is inconsistent.

More specifically:
- `OAuth2LoginReactiveAuthenticationManager` constructor is called over `getOauth2UserService()` which search for a bean of type `ReactiveOAuth2UserService`, else fallback on `DefaultReactiveOAuth2UserService`
- `OidcReactiveOAuth2UserService` is instanciated by its default constructor, without relying on `getOauth2UserService()`

```
		private ReactiveAuthenticationManager createDefault() {
			[...]
			OAuth2LoginReactiveAuthenticationManager oauth2Manager = new OAuth2LoginReactiveAuthenticationManager(
					client, getOauth2UserService());
```

```
		private ReactiveOAuth2UserService<OidcUserRequest, OidcUser> getOidcUserService() {
			ResolvableType type = ResolvableType.forClassWithGenerics(ReactiveOAuth2UserService.class,
					OidcUserRequest.class, OidcUser.class);
			ReactiveOAuth2UserService<OidcUserRequest, OidcUser> bean = getBeanOrNull(type);
			if (bean != null) {
				return bean;
			}
			return new OidcReactiveOAuth2UserService();
		}
```

This leads to inconsistent `ReactiveOAuth2UserService` through the components created by `OAuth2LoginSpec`.

**To Reproduce**

```
@EnableWebFluxSecurity
@Import({

		CustomOAuth2UserService.class,

})
@Slf4j
@SpringBootApplication(scanBasePackages = "none")
public class InconsistentOAuth2UserDetailsServiceApp {
	public static class CustomOAuth2UserService extends DefaultOAuth2UserService {

	}

	@Bean
	public SecurityWebFilterChain configureUi(ServerHttpSecurity http) {

		return http.build();
	}

	public static void main(String[] args) {
		SpringApplication.run(InconsistentOAuth2UserDetailsServiceApp.class);
	}

}
```

**Expected behavior**
Given a bean of type `ReactiveOAuth2UserService`, it should be injected to all components needed one, especially `OidcReactiveOAuth2UserService`, as it is done for `OAuth2LoginReactiveAuthenticationManager`

blacelle:
[...]

PR #15848, opened 2024-09-25
Auto config OidcReactiveOAuth2UserService When Oauth2UserService bean is Present

kse-music:
Closes gh-15846

### DIFF
--- a/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
@@ -4506,7 +4506,9 @@ public class ServerHttpSecurity {
 			if (bean != null) {
 				return bean;
 			}
-			return new OidcReactiveOAuth2UserService();
+			OidcReactiveOAuth2UserService reactiveOAuth2UserService = new OidcReactiveOAuth2UserService();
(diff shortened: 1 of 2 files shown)


config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -4506,7 +4506,9 @@ private ReactiveOAuth2UserService<OidcUserRequest, OidcUser> getOidcUserService(
 			if (bean != null) {
 				return bean;
 			}
-			return new OidcReactiveOAuth2UserService();
+			OidcReactiveOAuth2UserService reactiveOAuth2UserService = new OidcReactiveOAuth2UserService();
```
jzheaux, line 4509: Will you please add a test that demonstrates that when there is an `ReactiveOauth2UserService` bean, this OIDC service uses it correctly?

jzheaux (changes requested):
Thanks, @kse-music! Please see my feedback inline.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
kse-music: Done

jzheaux (approved):

jzheaux:

PR #15853, opened 2024-09-25
JwtDecoders should support issuer hostnames containing underscores

### LINKED ISSUE #15852: JwtDecoderProviderConfigurationUtils incorrectly handles issuer URI
**Describe the bug**
`JwtDecoderProviderConfigurationUtils` use `UriComponentsBuilder` to modify a `URI` in `oidc`, `oidcRfc8414` and `oauth`.
Due to a known issue in UriComponents builder (https://github.com/spring-projects/spring-framework/issues/27774), this produces invalid URIs in the cases where the URI contains an `authority`, but no `host`.
These still come up in practice though, e.g. see the following partial backtrace:

```
Caused by: java.lang.IllegalArgumentException: Unable to resolve the Configuration with the provided Issuer of "http://elated_sutherland:8080/auth/realms/my-realm"
        at org.springframework.security.oauth2.jwt.JwtDecoderProviderConfigurationUtils.getConfiguration(JwtDecoderProviderConfigurationUtils.java:178)
        at org.springframework.security.oauth2.jwt.JwtDecoderProviderConfigurationUtils.getConfigurationForIssuerLocation(JwtDecoderProviderConfigurationUtils.java:90)
        at org.springframework.security.oauth2.jwt.NimbusJwtDecoder.lambda$withIssuerLocation$2(NimbusJwtDecoder.java:226)
        at org.springframework.security.oauth2.jwt.NimbusJwtDecoder$JwkSetUriJwtDecoderBuilder.processor(NimbusJwtDecoder.java:389)
        at org.springframework.security.oauth2.jwt.NimbusJwtDecoder$JwkSetUriJwtDecoderBuilder.build(NimbusJwtDecoder.java:405)
        at org.springframework.boot.autoconfigure.security.oauth2.resource.servlet.OAuth2ResourceServerJwtConfiguration$JwtDecoderConfiguration.lambda$jwtDecoderByIssuerUri$3(OAuth2ResourceServerJwtConfiguration.java:159)
        at org.springframework.security.oauth2.jwt.SupplierJwtDecoder.lambda$new$0(SupplierJwtDecoder.java:36)
        ... 88 common frames omitted
Caused by: org.springframework.web.client.ResourceAccessException: I/O error on GET request for "http:/auth/realms/my-realm/.well-known/openid-configuration": Failed to select a proxy
        at org.springframework.web.client.RestTemplate.createResourceAccessException(RestTemplate.java:915)
        at org.springframework.web.client.RestTemplate.doExecute(RestTemplate.java:895)
        at org.springframework.web.client.RestTemplate.exchange(RestTemplate.java:740)
        at org.springframework.security.oauth2.jwt.JwtDecoderProviderConfigurationUtils.getConfiguration(JwtDecoderProviderConfigurationUtils.java:167)
        ... 94 common frames omitted
```

**To Reproduce**
```java
import org.springframework.web.util.UriComponentsBuilder;

import java.net.URI;
import java.util.Collections;

String rawUrl = "http://elated_sutherland:8080/auth/realms/my-realm";
URI uri = URI.create(rawUrl);
System.out.println(uri.getAuthority());
System.out.println(uri.getHost());
URI newUrl = UriComponentsBuilder.fromUri(uri)
  .replacePath(uri.getPath() + "/.well-known/openid-configuration")
  .build(Collections.emptyMap());
System.out.println(newUrl);
```

shows what happens. It outputs:
```
elated_sutherland:8080
null
http:/auth/realms/my-realm/.well-known/openid-configuration
```

**Expected behavior**
The correct oidc URI should be generated: `http://elated_sutherland:8080/auth/realms/my-realm/.well-known/openid-configuration`.

bodograumann:
Closes #15852

By using `UriComponentsBuilder.fromUriString` we stay in a single semantic context and do not loose information.

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
@@ -37,6 +37,7 @@ import org.springframework.security.oauth2.core.oidc.IdTokenClaimNames;
 import org.springframework.util.Assert;
 import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.RestTemplate;
+import org.springframework.web.util.UriComponents;
 import org.springframework.web.util.UriComponentsBuilder;
 
 /**
@@ -211,13 +212,18 @@ public final class ClientRegistrations {
 		};
 	}
 
-	private static Supplier<ClientRegistration.Builder> oidcRfc8414(URI issuer) {
+	private static Supplier<ClientRegistration.Builder> oidcRfc8414(String issuer) {
+		URI uri = oidcRfc8414Uri(issuer);
+		return getRfc8414Builder(issuer, uri);
+	}
+
+	static URI oidcRfc8414Uri(String issuer) {
+		UriComponents uri = UriComponentsBuilder.fromUriString(issuer).build();
 		// @formatter:off
-		URI uri = UriComponentsBuilder.fromUri(issuer)
-				.replacePath(OIDC_METADATA_PATH + issuer.getPath())
+		return UriComponentsBuilder.newInstance().uriComponents(uri)
+				.replacePath(OIDC_METADATA_PATH + uri.getPath())
 				.build(Collections.emptyMap());
 		// @formatter:on
-		return getRfc8414Builder(issuer, uri);
 	}
 
 	private static Supplier<ClientRegistration.Builder> oauth(URI issuer) {
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
@@ -17,7 +17,6 @@
 package org.springframework.security.oauth2.client.registration;
 
 import java.net.URI;
-import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -146,7 +145,7 @@ public final class ClientRegistrations {
 	 */
 	public static ClientRegistration.Builder fromOidcIssuerLocation(String issuer) {
 		Assert.hasText(issuer, "issuer cannot be empty");
-		return getBuilder(issuer, oidc(URI.create(issuer)));
+		return getBuilder(issuer, oidc(issuer));
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


jzheaux (changes requested):
Thanks for the PR, @bodograumann! WIll you please also add a test for both servlet and reactive that show that this behavior is now repaired?

jzheaux:
Thinking about this more, while it may still be valuable for this PR to have these tests, it seems to me that @rstoyanchev [is open to](https://github.com/spring-projects/spring-framework/issues/27774#issuecomment-1004757583) updating `UriComponentsBuilder`. I think it would be better to focus on addressing the issue in Framework first. Have you already considered submitting a PR there?

bodograumann:
True, fixing the underlying issue in UriComponentsBuilder would be preferrable, but I focused on what I could get done in limited time.
I'll try to look into adding a test, which will require somehow mocking the RestTemplate and maybe I'll also find time to understand the difference between authority and host enough to propose a solution for UriComponentsBuilder.

rstoyanchev:
The challenge with re-parsing the authority is having to extract the parsing logic out of the whole algorithm. Couldn't you re-parse the full URI string in fallback mode? Something like:

```java
URI uri = URI.create(issuer);
if (uri.getHost() == null && getAuthority() != null) {
    uri = UriComponentsBuilder.fromUriString(issuer).build().toURI();
}
```

bodograumann:
That makes sense @rstoyanchev .
Also I couldn't find a way to mock the RestTemplate here, so I create a PR for the spring framework: https://github.com/spring-projects/spring-framework/pull/33614

bodograumann:
> The challenge with re-parsing the authority is having to extract the parsing logic out of the whole algorithm. Couldn't you re-parse the full URI string in fallback mode? Something like:
> 
> ```java
> URI uri = URI.create(issuer);
> if (uri.getHost() == null && getAuthority() != null) {
>     uri = UriComponentsBuilder.fromUriString(issuer).build().toURI();
> }
> ```
> 
> This would impact only such URI's, and minimize any potential disruption for all others.

Given that since https://github.com/spring-projects/spring-framework/commit/bbb53d03c4f9f65b8bfaca92fbd1e88aabf6e347 `UriComponentsBuilder` uses the RFC 3986 parser by default, we are here only switching from RFC 2396 to RFC 3986 URI parsing, not to WhatWG parsing anymore.
Do you still think the impact would be too great, @rstoyanchev ?

As for testing: Could I make the `oidc`, `oidcRfc8414` and `oauth` methods package protected and test them directly? Testing anything involving `getConfiguration` and the RestTemplate seems unfeasible to me, when we want to vary the hostname.

jzheaux:
Thanks for the PR, @bodograumann! And I'm sorry for the extended delay.

In my testing, I found another issue with using `URI` and so I've adjusted these classes and `ClientRegistrations` to not use it. You can see these efforts in 571135d2ab06b24d30c91249ca8c7d0b0285a346 and 49e5d3796e019d8ba03c68481e7c1a4327cad751

PR #15857, opened 2024-09-26
Polish OAuth2ClientConfiguration

kse-music:
Use constructor inject and use `ObjectProvider.getIfUnique` instead of list.size == 1

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configuration/OAuth2ClientConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configuration/OAuth2ClientConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


jzheaux:
Thanks for the cleanup, @kse-music! Will you please add some more detail to your Git commit so that it is clear from reading the title what is going on? In the absence of a ticket that already has this detail, it would also be helpful if the commit had a description as to the benefit of this commit.

Second, I imagine that this pattern of taking a list and checking the size is common in Spring Security. Can you update the PR to clean up other places as well (e.g. `OAuth2ClientConfiguration`, `AbstractSecurityWebSocketMessageBrokerConfigurer`, etc.).

Finally, I think it would be better to do the `Arrays.asList` -> `Collections.singleton` maneuver in a separate PR since there are likely many places where this appears as well.

I can help with additional commits for anything you feel you don't have time for.

kse-music:
@jzheaux After inspection, only `OAuth2ClientConfiguration`, `OAuth2ClientWebFluxSecurityConfiguration`, and `AbstractSecurityWebSocketMessageBrokerConfigurer `were found to have the only bean judgment in the `spring-config module`. All except `AbstractSecurityWebSocketMessageBrokerConfigurer `have been modified because it tag deprecated.
 
If I have overlooked anything or other needs to be revised, please feel free to add it.

rwinch:
> Finally, I think it would be better to do the Arrays.asList -> Collections.singleton maneuver in a separate PR since there are likely many places where this appears as well.

Note that `Arrays.asList` is mutable and `Collections.singleton` is not. Switching to an immutable `Collection` would be considered as a breaking change.

jzheaux:
@rwinch, I learned something new today. :) I had always thought `Arrays.asList` was immutable since I couldn't add or remove from it. However, you can use `set(index, value)`. Good catch.

PR #15862, opened 2024-09-28
Update document

ngocnhan-tran1996:
I add breakline for easier reading

Before
image

After
image


jzheaux:
Another good catch, @ngocnhan-tran1996! Will you please update the commit message to

1. Capitalize the first word
2. Describe what change was made

For example, you could do:

```
Add whitespace to improve readability
```

PR #15876, opened 2024-10-04
Fix DefaultLoginPageConfigurerTests localization failure on system language different than english

CaptainAye:
The PR aims to fix the `DefaultLoginPageConfigurerTests.loginPageWhenErrorThenDefaultLoginPageWithError()` failure given the system language is set to something else than english.

When asserting the Bad credentials login page, the expected template contains the hardcoded "Bad credentials" message.
When the system language is other than english, this test fails locally as the actual code will produce the localized version of this page.

This PR fixes the issue adding the localized message to the expected page template in order to make sure that it will match the actual response regardless the locale setting.


sjohnr:
Thanks for the PR @CaptainAye! This is now merged into `main`.

PR #15898, opened 2024-10-10
Improve readibility of empty collection checks

kwonyonghyun:
In the current codebase, we're using .size() == 0 to check if Lists and Maps are empty. I propose changing this to .isEmpty().

Reasons for the proposed change:
1. Improved readability: .isEmpty() more clearly expresses the intent of checking whether a collection is empty.
2. Adherence to Java Collections Framework conventions: isEmpty() is the standard method for checking if a collection is empty.

If there's a specific reason for using `size() == 0` instead of `isEmpty()`, could you please share the rationale?


jzheaux (changes requested):
Thanks for the cleanup, @kwonyonghyun! Will you please update any non-Acegi copyright messages to be the current year? When you are done, please squash your commits into one and then go ahead and remove the "Proposal" part of your comment; maybe change it to something like:

```
Improve readability of empty collection checks
```

kwonyonghyun:
Thank you for the review @jzheaux ! Updated non-Acegi copyright messages to be the current year and made it into a single commit like you requested :)

jzheaux (approved):

jzheaux:
Thanks, @kwonyonghyun! This is now merged into `main`.

PR #15900, opened 2024-10-11
Polish diamond operator usage

ngocnhan-tran1996:

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/LoginUrlAuthenticationEntryPoint.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/LoginUrlAuthenticationEntryPoint.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2004, 2005, 2006 Acegi Technology Pty Limited
+ * Copyright 2004-2024 Acegi Technology Pty Limited
  *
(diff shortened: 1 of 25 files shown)


web/src/main/java/org/springframework/security/web/authentication/LoginUrlAuthenticationEntryPoint.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2004, 2005, 2006 Acegi Technology Pty Limited
```
jzheaux, line 2: It appears this update is not about improving diamond operator usage. If you want to keep it in, please do two things:

1. Leave any Acegi copyrights as-is
2. Move this to a separate commit for easier backporting

jzheaux (approved):

web/src/main/java/org/springframework/security/web/authentication/LoginUrlAuthenticationEntryPoint.java
ngocnhan-tran1996: I think I keep Acegi copyrights and just remove redundant keyword

jzheaux:
Thanks! This is now merged into `main`.

Issue #15905, opened 2024-10-13
Unhandled exception in CookieRequestCache results in 500 Internal Server Error


2is10:
**Describe the bug**

URL deserialization from the `REDIRECT_URI` cookie in CookieRequestCache fails and results in a 500 Internal Server Error if the cookie contains a non-Base64 character.

The 500 Internal Server Error stems from an IllegalArgumentException in this code in CookieRequestCache.java:
```java
private static String decodeCookie(String encodedCookieValue) {
    return new String(Base64.getDecoder().decode(encodedCookieValue.getBytes()));
}
```

**To Reproduce**

1. Configure Spring Security to use a CookieRequestCache as follows:

```java
http.requestCache(config -> config.requestCache(new CookieRequestCache()));
```

2. With the server running, send an authenticated request with an altered `REDIRECT_URI` cookie value:

```
curl -k 'https://localhost:8443' -H 'Cookie: SESSION=MWY…; REDIRECT_URI=123^^^456'
```

**Expected behavior**

The server should either ignore the invalid `REDIRECT_URI` cookie value and clear the cookie _or_ respond with status 400 Bad Request and an error description, as described in [the HTTP specification](https://httpwg.org/specs/rfc9110.html#status.400).

**Sample**

Any application that uses Spring Security need only insert this one line where its `HttpSecurity` is configured to reproduce the issue:
```java
http.requestCache(config -> config.requestCache(new CookieRequestCache()));
```

**Impact**

A customer of ours ran the Invicti Web Security Scanner on our application, and it triggered this error and reported it to the customer as a low-severity vulnerability ([details](https://www.invicti.com/web-vulnerability-scanner/vulnerabilities/internal-server-error/)). From that page:

> **Impact**
>
> The impact may vary depending on the condition. Generally this indicates poor coding practices, not enough error checking, sanitization and whitelisting.

I hope that you’ll prioritize fixing this issue since it has a simple fix and because even minor security vulnerabilities identified by web scanners make products/applications that use Spring Security look bad.

PR #15914, opened 2024-10-14
Catch base64 decode exception

kse-music:
Closes gh-15905

This is also aligned with `CookieServerRequestCache`

```
onErrorResume(IllegalArgumentException.class, (ex) -> Mono.empty())
```

### DIFF
--- a/web/src/main/java/org/springframework/security/web/savedrequest/CookieRequestCache.java
+++ b/web/src/main/java/org/springframework/security/web/savedrequest/CookieRequestCache.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/springframework/security/web/savedrequest/CookieRequestCache.java
```diff
@@ -73,7 +73,13 @@ public SavedRequest getRequest(HttpServletRequest request, HttpServletResponse r
 		if (savedRequestCookie == null) {
 			return null;
 		}
-		String originalURI = decodeCookie(savedRequestCookie.getValue());
+		String originalURI;
+		try {
+			originalURI = decodeCookie(savedRequestCookie.getValue());
+		}
+		catch (IllegalArgumentException ex) {
+			return null;
```
jzheaux, line 81: Let's please at least log the exception so that it isn't lost. `DEBUG` should be suitable.

jzheaux (changes requested):
Thanks, @kse-music! I've left some feedback inline. Also, please update the copyright years to any non-Acegi copyright messages.

jzheaux:
Thanks for the updates, @kse-music! Can you please update the copyright for the test file as well?

kse-music:
@jzheaux I have been updated

jzheaux (approved):

jzheaux:
Thanks, @kse-music! This is now merged into `5.8.x` as well as the other branches.

PR #15949, opened 2024-10-18
Align DelegatingAuthenticationConverter Constructors

kse-music:
This change is valuable since `new ArrayList<>()` and `List.of` have different behaviors, meaning that these two constructors give two different object states without saying so.

This PR aligns the two constructors.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/DelegatingAuthenticationConverter.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/DelegatingAuthenticationConverter.java
@@ -16,7 +16,6 @@
 
 package org.springframework.security.web.authentication;
 
-import java.util.ArrayList;
 import java.util.List;
 
 import jakarta.servlet.http.HttpServletRequest;
@@ -38,7 +37,7 @@ public final class DelegatingAuthenticationConverter implements AuthenticationCo
 
 	public DelegatingAuthenticationConverter(List<AuthenticationConverter> delegates) {
 		Assert.notEmpty(delegates, "delegates cannot be null");
-		this.delegates = new ArrayList<>(delegates);
+		this.delegates = List.copyOf(delegates);
 	}
 
 	public DelegatingAuthenticationConverter(AuthenticationConverter... delegates) {


web/src/main/java/org/springframework/security/web/authentication/DelegatingAuthenticationConverter.java
```diff
@@ -38,12 +37,11 @@ public final class DelegatingAuthenticationConverter implements AuthenticationCo
 
 	public DelegatingAuthenticationConverter(List<AuthenticationConverter> delegates) {
 		Assert.notEmpty(delegates, "delegates cannot be null");
-		this.delegates = new ArrayList<>(delegates);
+		this.delegates = delegates;
```
CrazyParanoid, line 40: Hi @kse-music , delegates passed to the constructor must be wrapped in a new list, only then can we suggest that they not be changed from the outside. Your solution is not safe in this regard, so I suggest leaving it as it is.

web/src/main/java/org/springframework/security/web/authentication/DelegatingAuthenticationConverter.java
```diff
@@ -38,12 +37,11 @@ public final class DelegatingAuthenticationConverter implements AuthenticationCo
...
 	public DelegatingAuthenticationConverter(List<AuthenticationConverter> delegates) {
 		Assert.notEmpty(delegates, "delegates cannot be null");
-		this.delegates = new ArrayList<>(delegates);
+		this.delegates = List.copyOf(delegates);
 	}
 
 	public DelegatingAuthenticationConverter(AuthenticationConverter... delegates) {
-		Assert.notEmpty(delegates, "delegates cannot be null");
-		this.delegates = List.of(delegates);
+		this(List.of(delegates));
```
jzheaux, line 44: I'm not a huge fan of recreating the list twice. Could we please leave this constructor as-is, even though that duplicates the null check?

jzheaux (changes requested):
@kse-music, thanks for this improvement. Please see my inline feedback.

Also, since the constructors demonstrate different behavior, I think this is a bug and should be fixed in `6.3.x` as well. Are you able to rebase this change to `6.3.x` please?

jzheaux:
Also, @kse-music, I updated the PR title and description to reflect some of the API benefits to this change.

kse-music:
Thanks a lot, @jzheaux. Everything is done as you expected.

jzheaux (approved):

jzheaux:

PR #15970, opened 2024-10-22
Add DefaultResourcesFilter.webauthn()

Kehrlann:
reviewer: @rwinch 

## Description

Use `DefaultResourcesFilter` with a proper static method for webauthn pages.

The filter is registered when:
- the default login page is used
- the default registration page is used

This means the filter may end up registered twice, which may cause a slight performance hit. We accept this tradeoff for code simplicity, since this is not intended for production workloads anyways.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/WebAuthnConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/WebAuthnConfigurer.java
@@ -16,7 +16,6 @@
 
 package org.springframework.security.config.annotation.web.configurers;
 
-import java.lang.reflect.Constructor;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Optional;
@@ -24,9 +23,6 @@ import java.util.Set;
 
 import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.context.ApplicationContext;
-import org.springframework.core.io.ClassPathResource;
-import org.springframework.http.HttpMethod;
-import org.springframework.http.MediaType;
 import org.springframework.security.authentication.ProviderManager;
 import org.springframework.security.config.annotation.web.HttpSecurityBuilder;
 import org.springframework.security.core.userdetails.UserDetailsService;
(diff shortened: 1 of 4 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/configurers/WebAuthnConfigurer.java
```diff
@@ -127,26 +119,15 @@ public void configure(H http) throws Exception {
 		http.addFilterBefore(webAuthnAuthnFilter, BasicAuthenticationFilter.class);
 		http.addFilterAfter(new WebAuthnRegistrationFilter(userCredentials, rpOperations), AuthorizationFilter.class);
 		http.addFilterBefore(new PublicKeyCredentialCreationOptionsFilter(rpOperations), AuthorizationFilter.class);
+		http.addFilter(DefaultResourcesFilter.webauthn());
```
rwinch, line 122: We don't want to register the webauthn JavaScript unless the default log in page is being used otherwise a user cannot override the JavaScript. Users that still want the JavaScript an explicitly register the JavaScript.

We should probably update DefaultWebAuthnRegistrationPageGeneratingFilter to only be registered when the default log in page is enabled.

rwinch (changes requested):
I've placed comments inline

config/src/main/java/org/springframework/security/config/annotation/web/configurers/WebAuthnConfigurer.java
```diff
@@ -127,29 +133,23 @@ public void configure(H http) throws Exception {
...
 				CsrfToken csrfToken = (CsrfToken) request.getAttribute(CsrfToken.class.getName());
 				return Map.of(csrfToken.getHeaderName(), csrfToken.getToken());
 			});
+			http.addFilter(DefaultResourcesFilter.webauthn());
+		}
+
+		if (!this.disableDefaultRegistrationPage) {
+			http.addFilterAfter(new DefaultWebAuthnRegistrationPageGeneratingFilter(userEntities, userCredentials),
+					AuthorizationFilter.class);
+			http.addFilter(DefaultResourcesFilter.webauthn());
```
rwinch, line 152: I don't think we want to register this twice when the default login page and the default registration pages are both enabled.

rwinch (changes requested):
I've left feedback inline

rwinch (approved):

rwinch:

PR #15978, opened 2024-10-23
Make RequestMatcherDelegatingAuthorizationManager post-processable

### LINKED ISSUE #15948: RequestMatcherDelegatingAuthorizationManager should be post-processable
Currently there is not way enforce the [Principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege) with respect to Handler Mappings and method security, e.g. PreAuthorize. 

Some [frameworks](http://github.com/grails/spring-security-core) built on top of Spring Security accomplish this by providing a [AnnotationFilterInvocationDefinition](https://github.com/grails/grails-spring-security-core/blob/7.0.x/plugin/src/main/groovy/grails/plugin/springsecurity/web/access/intercept/AnnotationFilterInvocationDefinition.groovy) to the [FilterSecurityInterceptor](https://github.com/spring-projects/spring-security/blob/dd94b119cab4945b7f1ddae0c46faf94100df95a/web/src/main/java/org/springframework/security/web/access/intercept/FilterSecurityInterceptor.java#L47-L5) that checks both the request mappings and method level annotations.

This approach, however, will no longer be possible in Spring Security 7 due to the fact that the FilterSecurityInterceptor is set for removal in favor of the AuthorizationFilter.

While it is possible to create an instance of the AuthorizationFilter with its own [AuthorizationManager](https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/authorization/AuthorizationManager.html), it is impossible to utilize the convenient creation of a RequestMatcherDelegatingAuthorizationManager.  

A very simple solution for achieving the  Principle of least privilege while honoring method level overrides, would be to instantiate a AuthorizationFilter with a wrapped RequestMatcherDelegatingAuthorizationManager that first checks the authorization level of request mapping and then does the method level check annotation check as an override.  This is not currently possible because [RequestMatcherDelegatingAuthorizationManager](https://github.com/spring-projects/spring-security/blob/7a1718887c0c10149b1f5057a7b532320f136b0a/web/src/main/java/org/springframework/security/web/access/intercept/RequestMatcherDelegatingAuthorizationManager.java) can not be built independently to the AuthorizationFilter, nor can it be customized to perform further checks.

Making it extensible or allowing it to be built outside of the AuthorizationFilter would enable the new approach to be accomplished and promote code reuse.  The goal here is to utilize the RequestMatcherDelegatingAuthorizationManager when creating a custom approach to securing requests that involves method level security.

codeconsole:
Fix extensibility issue since [RequestMatcherDelegatingAuthorizationManager](https://github.com/spring-projects/spring-security/blob/main/web/src/main/java/org/springframework/security/web/access/intercept/RequestMatcherDelegatingAuthorizationManager.java) is `final` and does not expose any public methods other than what is available through [AuthorizationManager](https://github.com/spring-projects/spring-security/blob/main/core/src/main/java/org/springframework/security/authorization/AuthorizationManager.java). Fixes #15948

Allows the following:

```java 
http.authorizeHttpRequests((authorize) -> authorize
    .requestMatchers("/assets/**", "/logout").permitAll()
    .anyRequest().permitAll().withObjectPostProcessor(new ObjectPostProcessor<AuthorizationManager<HttpServletRequest>>() {
        @Override
         public <O extends AuthorizationManager<HttpServletRequest>> O postProcess(O object) {
            return (O) new WrappedRequestMatcherDelegatingAuthorizationManager(context, object);
        }
    })
);
```

which opens up the possibility to provide additional security checks such as Controller annotations by wrapping the current manager and using the outcome of is authorization check to be compared against other checks.

This works similar to what is already possible with the ObjectPostProcessor for [AuthorizationFilter](https://github.com/spring-projects/spring-security/blob/main/web/src/main/java/org/springframework/security/web/access/intercept/AuthorizationFilter.java) except unlike [RequestMatcherDelegatingAuthorizationManager](https://github.com/spring-projects/spring-security/blob/main/web/src/main/java/org/springframework/security/web/access/intercept/RequestMatcherDelegatingAuthorizationManager.java), [AuthorizationFilter](https://github.com/spring-projects/spring-security/blob/main/web/src/main/java/org/springframework/security/web/access/intercept/AuthorizationFilter.java) is not `final` and can be extended. 

There is no benefit in post processing a `final` class that doesn't not expose any additional information that is not already provided by it's interface [AuthorizationManager<HttpServletRequest>](https://github.com/spring-projects/spring-security/blob/main/core/src/main/java/org/springframework/security/authorization/AuthorizationManager.java)

Alternatively, you could just remove `final` from 
https://github.com/spring-projects/spring-security/blob/8a972917fadb730264ee0d7da06e61d13837df2a/web/src/main/java/org/springframework/security/web/access/intercept/RequestMatcherDelegatingAuthorizationManager.java#L49
but either solution works.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/AuthorizeHttpRequestsConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/AuthorizeHttpRequestsConfigurer.java
@@ -171,7 +171,8 @@ public final class AuthorizeHttpRequestsConfigurer<H extends HttpSecurityBuilder
 			Assert.state(this.mappingCount > 0,
 					"At least one mapping is required (for example, authorizeHttpRequests().anyRequest().authenticated())");
 			ObservationRegistry registry = getObservationRegistry();
-			RequestMatcherDelegatingAuthorizationManager manager = postProcess(this.managerBuilder.build());
+			AuthorizationManager<HttpServletRequest> manager = postProcess(
+					(AuthorizationManager<HttpServletRequest>) this.managerBuilder.build());
 			if (registry.isNoop()) {
 				return manager;
 			}
--- a/config/src/test/java/org/springframework/security/config/annotation/web/configurers/AuthorizeHttpRequestsConfigurerTests.java
+++ b/config/src/test/java/org/springframework/security/config/annotation/web/configurers/AuthorizeHttpRequestsConfigurerTests.java
@@ -153,6 +153,7 @@ public class AuthorizeHttpRequestsConfigurerTests {
 		this.spring.register(ObjectPostProcessorConfig.class).autowire();
 		ObjectPostProcessor objectPostProcessor = this.spring.getContext().getBean(ObjectPostProcessor.class);
 		verify(objectPostProcessor).postProcess(any(RequestMatcherDelegatingAuthorizationManager.class));
+		verify(objectPostProcessor).postProcess(any(AuthorizationManager.class));
 		verify(objectPostProcessor).postProcess(any(AuthorizationFilter.class));
 	}
 


jzheaux:
Thanks, @codeconsole! Will you please also add a unit test in `AuthorizeHttpRequestsConfigurer` that confirms the behavior works?

And when you push your update, it will help if you can change your commit to something more like this:

```
Make RequestMatcherDelegatingAuthorizationManager Post-Processable

Closes gh-15978
```

It improves readability in the git history to have a short title. The `Closes` line makes sure our GitHub integration creates tickets for the other supported branches when this is merged.

codeconsole:
@jzheaux Thanks for the feedback.  I have changed the commit message and added the test. 

I also published a snapshot locally and confirmed it works as expected with my application and that I am able to perform the behavior that I needed it to do.

This 1 little change allows me to delete 462 lines of code so I will be excited to see it merged.  😄

codeconsole:
@jzheaux The version of `main` I was based on  was failing so I synced up and force pushed off of the latest main to make sure all the tests passed.  I hope that didn't cause any issues?

jzheaux:
It didn't cause any issues, @codeconsole, thanks for checking. I moved the PR to be based off of `6.2.x` so that we can forward port from there. Also, I accidentally gave you the wrong ticket number to place in the commit, so I just fixed it myself rather than have you re-do it.

jzheaux:
Thanks, @codeconsole! This is now merged into `6.2.x`, `6.3.x`, and `main`.

PR #15996, opened 2024-10-25
Polish

ngocnhan-tran1996:
This PR includes
- Polish diamond operator usage
- Polish Optional usage
- Simplify condition in some methods which can return direct result instead


jzheaux:
Thanks, @ngocnhan-tran1996, for the polishes!

Issue #14461, opened 2024-01-15
Access to accessToken in GrantedAuthoritiesMapper


Xyaren:
**Expected Behavior**

When using oauth2login it is should be possible to use the access token in GrantedAuthoritiesMapper in order to map access token claims to authorities for use in hasRole/hasAuthority.
This is espacially usefull when some claims do only appear in the access token.

**Current Behavior**
GrantedAuthoritiesMapper has currently no access to the accessToken.

**Context**
https://docs.spring.io/spring-security/reference/reactive/oauth2/login/advanced.html#webflux-oauth2-login-advanced-map-authorities-grantedauthoritiesmapper

My (dirty) workaround involves extending the OidcUserService and returning a new OidcUser Object.

jzheaux:
Hi, @Xyaren, thanks for reaching out. What is making you feel like extending `OidcUserService` is a dirty workaround? I'm thinking you might do something like this:

```java
@Component
public class MyOidcUserService implements OAuth2UserService<OidcUserRequest, OidcUser> {
    private final OidcUserService delegate = new OidcUserService();

    @Override
    public OidcUser loadUser(OidcUserRequest userRequest) throws OAuth2AuthenticationException {
        OidcUser user = this.delegate.loadUser(userRequest);
        Collection<GrantedAuthority> authorities = new ArrayList<>(user.getAuthorities());
        // ... add custom authorities
        return new DefaultOidcUser(authorities, user.getIdToken(), user.getUserInfo());
    }
}
```

which doesn't seem dirty to me. I want to make sure I'm not missing something; can you clarify please?

Xyaren:
Hi, thanks for the quick response!
I do have a similar workaround, that I'll attach at the end.

It feels odd wrapping the original service, tear apart the response and stitch it together again in a new instance of the same object. I also was unsure about "reconstructing" the OidcUser due to no access to nameAttributeKey which is required for the all-arg constructor. Therefore I went for extending the default implementation and wrapping the result.
It there anything speaking against having the Access token available within the OidcUser ? 
This would allow users to use a GrantedAuthoritiesMapper to map from any of the sources.

```java
import java.util.Collection;
import java.util.HashSet;
import java.util.Set;
import java.util.stream.Collectors;

import packages.from.my.company.that.i.dont.want.to.name.JwtToAuthorityExtractor;
import lombok.AllArgsConstructor;
import lombok.experimental.Delegate;
import org.jetbrains.annotations.NotNull;
[...]

PR #16003, opened 2024-10-26
Expose getter for nameAttributeKey in OAuth2AuthenticatedPrincipal

andreblanke:
Given a `DefaultOidcUser` or `DefaultOAuth2User`, it is currently not possible to create a faithful copy of the instance since the constructors require a `nameAttributeKey` for the all-arg constructor:

```java
public DefaultOidcUser(Collection<? extends GrantedAuthority> authorities, OidcIdToken idToken,
		String nameAttributeKey)
```

```java
public DefaultOAuth2User(Collection<? extends GrantedAuthority> authorities, Map<String, Object> attributes,
		String nameAttributeKey)
```

which is not accessible.

This issue has been mentioned before in https://github.com/spring-projects/spring-security/issues/14461#issuecomment-1898069450 where a custom `ExtendedOidcUser` decorator class is used to work around it. While that works, I feel that this should be possible without introducing a separate class using just the public API.

The PR aims to change this by adding the `OAuth2AuthenticatedPrincipal.getNameAttributeKey` method in 464b078d82632d3b8e162042fc6d27fb323a1fc2. This is a breaking change for classes implementing the interface.

With the `nameAttributeKey` now available from within the interface, I figured it also makes sense to provide a default implementation for `OAuth2AuthenticatedPrincipal.getName` in c7bcf865715d3f78017131e8e7470022f1508842.

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/jackson2/DefaultOidcUserMixin.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/jackson2/DefaultOidcUserMixin.java
@@ -43,10 +43,17 @@ import org.springframework.security.oauth2.core.oidc.user.DefaultOidcUser;
 @JsonIgnoreProperties(value = { "attributes" }, ignoreUnknown = true)
 abstract class DefaultOidcUserMixin {
 
+	@Deprecated
 	@JsonCreator
 	DefaultOidcUserMixin(@JsonProperty("authorities") Collection<? extends GrantedAuthority> authorities,
 			@JsonProperty("idToken") OidcIdToken idToken, @JsonProperty("userInfo") OidcUserInfo userInfo,
 			@JsonProperty("nameAttributeKey") String nameAttributeKey) {
+	}
+
+	@JsonCreator
+	DefaultOidcUserMixin(@JsonProperty("name") String name,
+			@JsonProperty("idToken") OidcIdToken idToken, @JsonProperty("userInfo") OidcUserInfo userInfo,
+			@JsonProperty("authorities") Collection<? extends GrantedAuthority> authorities) {
 	}
 
 }
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/jackson2/DefaultOidcUserMixin.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/jackson2/DefaultOidcUserMixin.java
@@ -43,17 +43,10 @@ import org.springframework.security.oauth2.core.oidc.user.DefaultOidcUser;
 @JsonIgnoreProperties(value = { "attributes" }, ignoreUnknown = true)
 abstract class DefaultOidcUserMixin {
 
-	@Deprecated
 	@JsonCreator
 	DefaultOidcUserMixin(@JsonProperty("authorities") Collection<? extends GrantedAuthority> authorities,
 			@JsonProperty("idToken") OidcIdToken idToken, @JsonProperty("userInfo") OidcUserInfo userInfo,
 			@JsonProperty("nameAttributeKey") String nameAttributeKey) {
-	}
-
-	@JsonCreator
-	DefaultOidcUserMixin(@JsonProperty("name") String name,
-			@JsonProperty("idToken") OidcIdToken idToken, @JsonProperty("userInfo") OidcUserInfo userInfo,
-			@JsonProperty("authorities") Collection<? extends GrantedAuthority> authorities) {
 	}
 
 }
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserRequestUtils.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserRequestUtils.java
@@ -17,6 +17,7 @@
 package org.springframework.security.oauth2.client.oidc.userinfo;
 
 import java.util.LinkedHashSet;
+import java.util.Map;
 import java.util.Set;
 
 import org.springframework.security.core.GrantedAuthority;
@@ -28,6 +29,7 @@ import org.springframework.security.oauth2.core.oidc.OidcUserInfo;
 import org.springframework.security.oauth2.core.oidc.user.DefaultOidcUser;
 import org.springframework.security.oauth2.core.oidc.user.OidcUser;
 import org.springframework.security.oauth2.core.oidc.user.OidcUserAuthority;
+import org.springframework.security.oauth2.core.user.DefaultOAuth2User;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.StringUtils;
 
@@ -90,10 +92,15 @@ final class OidcUserRequestUtils {
 		for (String scope : token.getScopes()) {
 			authorities.add(new SimpleGrantedAuthority("SCOPE_" + scope));
 		}
+		DefaultOidcUser.Builder userBuilder = new DefaultOidcUser.Builder();
 		if (StringUtils.hasText(userNameAttributeName)) {
-			return new DefaultOidcUser(authorities, userRequest.getIdToken(), userInfo, userNameAttributeName);
+			userBuilder.nameAttributeKey(userNameAttributeName);
 		}
-		return new DefaultOidcUser(authorities, userRequest.getIdToken(), userInfo);
+		return userBuilder
+			.idToken(userRequest.getIdToken())
+			.userInfo(userInfo)
+			.authorities(authorities)
+			.build();
 	}
 
 	private OidcUserRequestUtils() {
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserRequestUtils.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserRequestUtils.java
@@ -17,7 +17,6 @@
 package org.springframework.security.oauth2.client.oidc.userinfo;
 
 import java.util.LinkedHashSet;
-import java.util.Map;
 import java.util.Set;
 
 import org.springframework.security.core.GrantedAuthority;
@@ -29,7 +28,6 @@ import org.springframework.security.oauth2.core.oidc.OidcUserInfo;
 import org.springframework.security.oauth2.core.oidc.user.DefaultOidcUser;
 import org.springframework.security.oauth2.core.oidc.user.OidcUser;
 import org.springframework.security.oauth2.core.oidc.user.OidcUserAuthority;
-import org.springframework.security.oauth2.core.user.DefaultOAuth2User;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.StringUtils;
 
@@ -92,15 +90,10 @@ final class OidcUserRequestUtils {
 		for (String scope : token.getScopes()) {
 			authorities.add(new SimpleGrantedAuthority("SCOPE_" + scope));
 		}
-		DefaultOidcUser.Builder userBuilder = new DefaultOidcUser.Builder();
 		if (StringUtils.hasText(userNameAttributeName)) {
-			userBuilder.nameAttributeKey(userNameAttributeName);
+			return new DefaultOidcUser(authorities, userRequest.getIdToken(), userInfo, userNameAttributeName);
 		}
-		return userBuilder
-			.idToken(userRequest.getIdToken())
-			.userInfo(userInfo)
-			.authorities(authorities)
-			.build();
+		return new DefaultOidcUser(authorities, userRequest.getIdToken(), userInfo);
 	}
 
 	private OidcUserRequestUtils() {
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
@@ -96,11 +96,7 @@ public class DefaultOAuth2UserService implements OAuth2UserService<OAuth2UserReq
 		OAuth2AccessToken token = userRequest.getAccessToken();
 		Map<String, Object> attributes = this.attributesConverter.convert(userRequest).convert(response.getBody());
 		Collection<GrantedAuthority> authorities = getAuthorities(token, attributes, userNameAttributeName);
-		return new DefaultOAuth2User.Builder()
-			.nameAttributeKey(userNameAttributeName)
-			.attributes(attributes)
-			.authorities(authorities)
-			.build();
+		return new DefaultOAuth2User(authorities, attributes, userNameAttributeName);
 	}
 
 	/**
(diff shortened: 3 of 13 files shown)


oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/DefaultOAuth2AuthenticatedPrincipal.java
```diff
@@ -64,7 +64,7 @@ public DefaultOAuth2AuthenticatedPrincipal(String name, Map<String, Object> attr
 		this.attributes = Collections.unmodifiableMap(attributes);
 		this.authorities = (authorities != null) ? Collections.unmodifiableCollection(authorities)
 				: AuthorityUtils.NO_AUTHORITIES;
-		this.name = (name != null) ? name : (String) this.attributes.get("sub");
+		this.name = (name != null) ? name : OAuth2AuthenticatedPrincipal.super.getName();
```
jzheaux, line 67: Let's please leave this as-is.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/DefaultOAuth2AuthenticatedPrincipal.java
```diff
@@ -81,6 +81,11 @@ public Collection<? extends GrantedAuthority> getAuthorities() {
 		return this.authorities;
 	}
 
+	@Override
```
jzheaux, line 84: What if we instead added a constructor that took the `String name` instead of the attribute key? Then, we could simply deprecate the `nameAttributeKey` constructor and have it call the new constructor? In that case, this method and the interface method would go away.

jzheaux: At that point, we'd also want to update production files to use the new constructor (that is, tests can stay the same).

jzheaux (changes requested):
Thanks, @andreblanke! I've left some feedback inline.

jzheaux:
Hi, @andreblanke! Are you able to make the requested changes? No rush, if I don't hear from you in about a week, I'm happy to make the changes myself.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/DefaultOAuth2AuthenticatedPrincipal.java
andreblanke: I like your suggestion with a new constructor. It wouldn't affect downstream users of the interface and aligns well with the getters since there would be one constructor parameter for each.

The only issue I see is that the existing `String nameAttributeKey` and new `String name` constructors would share the same unless the order was changed to have the `name` in front:

```java
public DefaultOAuth2User(String name, Collection<? extends GrantedAuthority> authorities, Map<String, Object> attributes)
```

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/DefaultOAuth2AuthenticatedPrincipal.java
andreblanke: I've reverted the line to

```java
this.name = (name != null) ? name : (String) this.attributes.get("sub");
```

`"sub"` is now used twice within `DefaultOAuth2AuthenticatedPrincipal`.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/DefaultOAuth2AuthenticatedPrincipal.java
jzheaux: Good point. Typically authenticated principals in Spring Security follow this order:

```
(principal, details, authorities)
```

So this is also an opportunity to bring this constructor into alignment with the rest. It should be:

```
(String name, Map<String, Object> attributes, Collection<? extends GrantedAuthority> authorities)
```

In that case, will you also please add a `@JsonCreator` in `DefaultOAuth2UserMixin` that points to the new constructor? Note that the old `@JsonCreator` and constructor should remain in place.

andreblanke: I added the new `String name` constructors to `DefaultOAuth2User` and `DefaultOidcUser` and accompanying `@JsonCreator` in `DefaultOAuth2UserMixin` and `DefaultOidcUserMixin`. `DefaultOidcUser` has quite a few constructors and I aligned the parameter order in all of them, deprecating the old ones.

All production usages use the new constructors (except for one in `OidcUserRequestUtils.getUser` which would require `OidcUserAuthority.collectClaims` to be accessible, so this is still a to-do) while the tests stay the same.

One last open topic would be unit tests for the new constructors, but I figured I'd wait for your feedback before approaching them.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserRequestUtils.java
```diff
@@ -91,9 +91,10 @@ static OidcUser getUser(OidcUserRequest userRequest, OidcUserInfo userInfo) {
 			authorities.add(new SimpleGrantedAuthority("SCOPE_" + scope));
 		}
 		if (StringUtils.hasText(userNameAttributeName)) {
+			// TODO: Get name from OidcUserAuthority.collectClaims.
```
jzheaux, line 94: Since this is simply combining the two maps, perhaps you could duplicate that logic here in order to obtain the name

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
```diff
@@ -96,7 +96,7 @@ public OAuth2User loadUser(OAuth2UserRequest userRequest) throws OAuth2Authentic
 		OAuth2AccessToken token = userRequest.getAccessToken();
 		Map<String, Object> attributes = this.attributesConverter.convert(userRequest).convert(response.getBody());
 		Collection<GrantedAuthority> authorities = getAuthorities(token, attributes, userNameAttributeName);
-		return new DefaultOAuth2User(authorities, attributes, userNameAttributeName);
+		return new DefaultOAuth2User(attributes.get(userNameAttributeName).toString(), attributes, authorities);
```
jzheaux, line 99: Is it guaranteed at this point that attribute.get(userNameAttributeName) is non-null? If not, please do a null check so that the code can offer a more informative error message.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/oidc/user/DefaultOidcUser.java
```diff
@@ -51,34 +51,68 @@ public class DefaultOidcUser extends DefaultOAuth2User implements OidcUser {
...
 		this(authorities, idToken, IdTokenClaimNames.SUB);
 	}
 
+	/**
+	 * Constructs a {@code DefaultOidcUser} using the provided parameters.
+	 * @param idToken the {@link OidcIdToken ID Token} containing claims about the user
+	 * @param authorities the authorities granted to the user
+	 */
+	public DefaultOidcUser(OidcIdToken idToken, Collection<? extends GrantedAuthority> authorities) {
+		this(null, idToken, authorities);
```
jzheaux, line 65: Is this correct? It seems like it should be `idToken.getAttribute(IdTokenClaimNames.SUB)`.

Same question for `DefaultOidcUser(OidcIdToken, OidcUserInfo, Collection<? extends GrantedAuthority>)`

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/jackson2/DefaultOidcUserMixin.java
```diff
@@ -43,10 +43,17 @@
...
 
+	@Deprecated
 	@JsonCreator
 	DefaultOidcUserMixin(@JsonProperty("authorities") Collection<? extends GrantedAuthority> authorities,
 			@JsonProperty("idToken") OidcIdToken idToken, @JsonProperty("userInfo") OidcUserInfo userInfo,
 			@JsonProperty("nameAttributeKey") String nameAttributeKey) {
 	}
 
+	@JsonCreator
+	DefaultOidcUserMixin(@JsonProperty("name") String name,
```
jzheaux, line 54: Will you please add `DefaultOAuth2UserMixinTests` and `DefaultOidcUserMixinTests` to ensure that serializations of either arrangement can correctly deserialize?

jzheaux (changes requested):
Thanks for the updates, @andreblanke! I've left some feedback inilne.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/oidc/user/DefaultOidcUser.java
andreblanke: Both constructors ultimately delegate to

```java
public DefaultOAuth2User(String name, Map<String, Object> attributes,
			Collection<? extends GrantedAuthority> authorities) {
    // ...
    this.name = (name != null) ? name : (String) this.attributes.get("sub");
}
```

which is consistent with the constructor behavior of `DefaultOAuth2AuthenticatedPrincipal`, although using `IdTokenClaimNames.SUB` instead of `"sub"` would be more appropriate.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserRequestUtils.java
andreblanke: Does anything speak against making `OidcUserAuthority.collectClaims` public instead of duplicating the logic in your opinion? I could imagine that by deprecating the `nameAttributeKey` constructors users of Spring Security might also require a snippet such as

```java
Map<String, Object> attributes =  OidcUserAuthority.collectClaims(userRequest.getIdToken(), userInfo);
String name = (String) attributes.get(userNameAttributeName);
new DefaultOidcUser(name, userRequest.getIdToken(), userInfo, authorities);
```

when instantiating `DefaultOidcUser`. With `OidcUserAuthority.collectClaims` being package-private merging the claims would need to be reimplemented for each project.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/oidc/user/DefaultOidcUser.java
andreblanke: However, for that to succeed I'd still need to remove the `Assert.notNull(name, "name cannot be null");` from the constructor which is still present at the moment.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
andreblanke: You're right, `attributes` is nullable and an assertion is appropriate here, basically what my newly introduced `DefaultOAuth2User.getNameFromAttributes` is doing.

However, I can't help but feel that "get name from attributes using nameAttributeKey with assertions" is a frequent enough operation (4 mostly similar occurrences have been touched by this PR) to maybe warrant a separate method.

What would you think about changing the constructors to use `name` instead of `nameAttributeKey` (as done by the PR) but introducing static factory methods for `DefaultOidcUser` and `DefaultOAuth2User` taking `nameAttributeKey` as parameter instead of `name`?  
That would get rid of the old, inconsistent constructors (in terms of both the parameter order and the usage of `nameAttributeKey` instead of `name`) while giving a clear upgrade path. Additionally, `OidcUserAuthority.collectClaims` mentioned above would not need to be duplicated or made accessible.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/userinfo/OidcUserRequestUtils.java
andreblanke: Should no longer be relevant with the introduction of builder classes.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
andreblanke: I've added builders for `DefaultOAuth2User` and `DefaultOidcUser`. Right now the new "canonical constructor" is still public, but it might be better to make it private to be in line with, e.g., `OAuth2AuthorizeRequest`?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/jackson2/DefaultOidcUserMixin.java
andreblanke: It seems that Jackson only supports a single `@JsonCreator` for each class, meaning the backwards compatibility would be a bit more involved (perhaps using a custom deserializer?).

jzheaux: Yes, the next evolution of that mixin would be to use a deserializer.

jzheaux:
Hi, @andreblanke. Thanks for doing all this research with me, I appreciate your contributions. At this point, I think it's pretty clear that this approach is too complex to address your much simpler concern. I think deprecation will need to wait for another time.

Can you confirm that this is what you are effectively needing:

```java
DefaultOAuth2User user = new DefaultOAuth2User(...);
DefaultOAuth2User copy = // ... copy the values somehow
assert user.equals(copy);
assert user != copy;
```

If so, I think we can do something much simpler, which is add a copy constructor to each class:

```java
DefaultOAuth2User user = new DefaultOAuth2User(...);
DefaultOAuth2User copy = new DefaultOAuth2User(user);
assert user.equals(copy);
assert user != copy;

// ... same for DefaultOidcUser
```

If you can confirm that this would work, I will push a change along these lines and have you review it.

andreblanke:
Hi @jzheaux. First of all thank your for your guidance so far. Sorry the PR has been dragging on for a while now. I agree that the current approach is too complex and the scope of the PR has grown quite a bit from my original submission.

Unfortunately a copy constructor wouldn't be sufficient for my use case. My original requirement (which I should've maybe mentioned to avoid the XY problem) was to add a new `GrantedAuthority` to an `OidcUser`.

Since `DefaultOAuth2User.authorities` is unmodifiable, re-creating the `OidcUser` using

```java
new DefaultOidcUser(newMappedAuthorities, originalUser.idToken, originalUser.userInfo)
```

and replacing the principal where it's used is the only option. However, the new `DefaultOidcUser` isn't an exact copy since its `nameAttributeKey` is `IdTokenClaimNames.SUB` as opposed to whatever it was for `originalUser.

jzheaux:
Gotcha, @andreblanke, that makes sense.

Given that you need to add more authorities, could you do something like the following:

```java
OidcUserService userService = new OidcUserService();
userService.setOidcUserMapper((request, info) -> {
	List<GrantedAuthority> authorities = request.getAccessToken().getScopes()
		.stream().map(SimpleGrantedAuthority::new).toList();
	authorities.add(your custom authority);
	return new DefaultOidcUser(authorities, request.getIdToken(), info);
});
```

jzheaux:
I'm going to close this PR at this point, @andreblanke. Thanks for your efforts, and please don't hesitate to reach out if the above snippet doesn't address your concern.

Issue #15988, opened 2024-10-24
Implementations of OpaqueTokenIntrospector fail to URL encode client secret


joelossher:
**Describe the bug**
Both the `SpringOpaqueTokenIntrospector` and `NimbusOpaqueTokenIntrospector` use the `clientId` and `clientSecret` to authenticate the calls to the authorization server.

This is done via basic authentication added using a `BasicAuthenticationInterceptor`. This does not perform any URL encoding.

This issue was addressed in https://github.com/spring-projects/spring-security/issues/9610 for the token granting client, but persists for the introspection client.

The workaround at the moment is to manually encode the secret when instantiating the introspector.

**To Reproduce**
1. Set up a Spring Authorization Server with a client with a secret such as `badSecret%`
1. Configure a `SpringOpaqueTokenIntrospector` or `NimbusOpaqueTokenIntrospector` to use that client
2. Attempt to use the introspector with the Spring Authorization Server.
3. See the server respond with a 400 `invalid_request` error and see the following cause in the logs:
```
Caused by: java.lang.IllegalArgumentException: URLDecoder: Incomplete trailing escape (%) pattern
        at java.base/java.net.URLDecoder.decode(URLDecoder.java:230) ~[?:?]
        at java.base/java.net.URLDecoder.decode(URLDecoder.java:147) ~[?:?]
        at org.springframework.security.oauth2.server.authorization.web.authentication.ClientSecretBasicAuthenticationConverter.convert(ClientSecretBasicAuthenticationConverter.java:85) ~[spring-security-oauth2-authorization-server-1.3.2.jar!/:1.3.2]
        ... 103 more
```

**Expected behavior**
The token introspector should URL encode the secret.

jzheaux:
Thanks for this report, @joelossher. I agree that this should be taken care of.

My primary concern is that there are applications like yours that are self-encoding already (like you are). To start encoding by default at this point would break those applications.

Because there is a constructor that accepts a `RestOperations` instance, there is already a workaround, should the fix cause any issues; we just don't want to break folks in a point release.

For that reason, I'll schedule this fix for 6.5.

jzheaux:
Are you able to provide a PR to add the encoding and a test?

ngocnhan-tran1996:
@jzheaux 

`SpringOpaqueTokenIntrospector` use `BasicAuthenticationInterceptor` for encoding `clientId` and `clientSecret`.
[...]

PR #16008, opened 2024-10-29
Encode clientId and clientSecret for `OpaqueTokenIntrospector` and `ReactiveOpaqueTokenIntrospector`

ngocnhan-tran1996:
Closes gh-15988

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringOpaqueTokenIntrospector.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringOpaqueTokenIntrospector.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2024 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,6 +18,8 @@ package org.springframework.security.oauth2.server.resource.introspection;
 
 import java.io.Serial;
 import java.net.URI;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -77,9 +79,11 @@ public class SpringOpaqueTokenIntrospector implements OpaqueTokenIntrospector {
 	/**
 	 * Creates a {@code OpaqueTokenAuthenticationProvider} with the provided parameters
 	 * @param introspectionUri The introspection endpoint uri
-	 * @param clientId The client id authorized to introspect
-	 * @param clientSecret The client's secret
+	 * @param clientId The URL-encoded client id authorized to introspect
+	 * @param clientSecret The URL-encoded client secret authorized to introspect
+	 * @deprecated Please use {@link SpringOpaqueTokenIntrospector.Builder}
 	 */
+	@Deprecated(since = "6.5", forRemoval = true)
 	public SpringOpaqueTokenIntrospector(String introspectionUri, String clientId, String clientSecret) {
 		Assert.notNull(introspectionUri, "introspectionUri cannot be null");
 		Assert.notNull(clientId, "clientId cannot be null");
@@ -269,6 +273,18 @@ public class SpringOpaqueTokenIntrospector implements OpaqueTokenIntrospector {
 		return authorities;
 	}
 
+	/**
+	 * Creates a {@code SpringOpaqueTokenIntrospector.Builder} with the given
+	 * introspection endpoint uri
+	 * @param introspectionUri The introspection endpoint uri
+	 * @return the {@link SpringOpaqueTokenIntrospector.Builder}
+	 * @since 6.5
+	 */
+	public static Builder withIntrospectionUri(String introspectionUri) {
+		Assert.notNull(introspectionUri, "introspectionUri cannot be null");
+		return new Builder(introspectionUri);
+	}
+
 	// gh-7563
 	private static final class ArrayListFromString extends ArrayList<String> {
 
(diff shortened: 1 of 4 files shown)


jzheaux (changes requested):
@ngocnhan-tran1996 thanks for your patience while we got ready for 6.5 development. I've left my feedback inline.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringOpaqueTokenIntrospector.java
```diff
@@ -295,4 +300,94 @@ default List<String> getScopes() {
 
 	}
 
+	/**
+	 * Used to build {@link SpringOpaqueTokenIntrospector}.
+	 *
+	 * @author Ngoc Nhan
+	 * @since 6.5
+	 */
+	public static final class SpringOpaqueTokenIntrospectorBuilder {
```
jzheaux, line 309: I think you can just call this `Builder` as the fully-qualified name already includes `SpringOpaqueTokenIntrospector`.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringOpaqueTokenIntrospector.java
```diff
@@ -295,4 +300,94 @@ default List<String> getScopes() {
...
+		}
+
+		/**
+		 * Creates a {@code SpringOpaqueTokenIntrospectorBuilder} with the provided
+		 * parameters
+		 * @param introspectionUri The introspection endpoint uri
+		 * @return the {@link SpringOpaqueTokenIntrospectorBuilder}
+		 * @since 6.5
+		 */
+		public static SpringOpaqueTokenIntrospectorBuilder withIntrospectionUri(String introspectionUri) {
```
jzheaux, line 324: As a matter of convention, please move this out of the inner class and into `SpringOpaqueTokenIntrospector`.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringOpaqueTokenIntrospector.java
```diff
@@ -295,4 +300,94 @@ default List<String> getScopes() {
...
+		}
+
+		/**
+		 * Creates a {@code SpringOpaqueTokenIntrospector} with the provided parameters
+		 * @param clientId The client id authorized
+		 * @param clientSecret The client secret for the authorized client
+		 * @return the {@link SpringOpaqueTokenIntrospector}
+		 * @since 6.5
+		 */
+		public SpringOpaqueTokenIntrospector introspectionClientCredentials(String clientId, String clientSecret) {
```
jzheaux, line 370: Please separate these, again as a matter of convention. For additional brevity, can you please name them `clientId` and `clientSecret`.

jzheaux, line 370: It's more common for a builder to have a `build()` method. That will make it more obvious to a coder which method they call when they are done. Can you so that a `build()` method is what calls the constructor?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringOpaqueTokenIntrospector.java
```diff
@@ -295,4 +300,94 @@ default List<String> getScopes() {
...
+
+		/**
+		 * Creates a {@code SpringOpaqueTokenIntrospector} with the provided parameters
+		 * The given {@link RestOperations} should perform its own client authentication
+		 * against the introspection endpoint.
+		 * @param restOperations The client for performing the introspection request
+		 * @return the {@link SpringOpaqueTokenIntrospector}
+		 * @since 6.5
+		 */
+		public SpringOpaqueTokenIntrospector withRestOperations(RestOperations restOperations) {
```
jzheaux, line 386: Let's leave this out for now. If they have a `RestOperations`, they can simply use the `RestOperations` constructor. Since they at that point need to add their own interceptor, I trust that they know what they are doing regarding client id and secret.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringOpaqueTokenIntrospector.java
```diff
@@ -295,4 +300,94 @@ default List<String> getScopes() {
...
+		/**
+		 * Creates a {@code SpringOpaqueTokenIntrospector} with the provided parameters
+		 * @param clientId The client id authorized that should be encode
+		 * @param clientSecret The client secret that should be encode for the authorized
+		 * client
+		 * @param charset the charset to use
+		 * @return the {@link SpringOpaqueTokenIntrospector}
+		 * @since 6.5
+		 */
+		public SpringOpaqueTokenIntrospector introspectionEncodeClientCredentials(String clientId, String clientSecret,
```
jzheaux, line 351: Please leave these convenience methods out for now. I think we can start with `clientId`, `clientSecret`, and `build`.

jzheaux (changes requested):
Thanks, @ngocnhan-tran1996! I've sent you some of my feedback, mostly about Security's code conventions. Also, if you are able, please also update the documentation to use the builder instead of the now-deprecated constructor.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringOpaqueTokenIntrospector.java
```diff
@@ -295,4 +311,87 @@ default List<String> getScopes() {
...
+			return this;
+		}
+
+		/**
+		 * Uses the given parameter to build {@code SpringOpaqueTokenIntrospector}
+		 * @param clientId The client id authorized
+		 * @return the {@link SpringOpaqueTokenIntrospector.Builder}
+		 * @since 6.5
+		 */
+		public Builder clientId(String clientId) {
```
jzheaux, line 352: While I see the value of having both encoded and unencoded methods here, I think that distinction will be lost on most users of the API. Let's please only have `clientId`, which takes an unencoded value and encodes it using `UTF_8` and `clientSecret`, which takes an unencoded value and encodes it using `UTF_8`. The builder should then only have three public methods.

If folks need a different `CharSet` or if they need to encode the values themselves, they should use the `RestOperations` constructor (or the deprecated constructor for the time being).

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringOpaqueTokenIntrospector.java
```diff
@@ -79,7 +81,9 @@ public class SpringOpaqueTokenIntrospector implements OpaqueTokenIntrospector {
 	 * @param introspectionUri The introspection endpoint uri
 	 * @param clientId The client id authorized to introspect
```
jzheaux, line 82: Please update this JavaDoc to say something like:

```
clientId The URL-encoded client id authorized to introspect
```

And the same for `clientSecret`

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringOpaqueTokenIntrospector.java
```diff
@@ -295,4 +311,87 @@ default List<String> getScopes() {
...
+		public Builder clientId(String clientId, Charset charset) {
+			Assert.notNull(clientId, "clientId cannot be null");
+			Assert.notNull(charset, "charset cannot be null");
+			this.clientId = URLEncoder.encode(clientId, charset);
+			return this;
+		}
+
+		/**
+		 * Uses the given parameter to build {@code SpringOpaqueTokenIntrospector}
+		 * @param clientId The client id authorized
```
jzheaux, line 348: Please change this to something like

```
clientId The unencoded client id ...
```

You could also change the description to explain this a bit more:

```
Use this client identifier. Note that the builder will {@link URLEncode} the value that you provide, so please give the unencoded value. This will not make a different for most client ids.
```

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/introspection/SpringReactiveOpaqueTokenIntrospectorTests.java
```diff
@@ -261,6 +262,52 @@ public void constructorWhenRestOperationsIsNullThenIllegalArgumentException() {
...
+					{
+						"active": true,
+						"username": "client%&1"
+					}
+					""";
+			server.setDispatcher(requiresAuth("client%25%261", "secret%40%242", response));
+			String introspectUri = server.url("/introspect").toString();
+			ReactiveOpaqueTokenIntrospector introspectionClient = SpringReactiveOpaqueTokenIntrospector
+				.withIntrospectionUri(introspectUri)
+				.clientId("client%&1", StandardCharsets.UTF_8)
```
jzheaux, line 298: Let's use the unencoded values in this test in order to confirm that you've successfully added the requested functionality.

jzheaux (changes requested):
Thanks for the update, @ngocnhan-tran1996, and for your patience as I came back around to this PR. I've left some additional feedback inline.

Also, are you able to update the documentation [here](https://github.com/spring-projects/spring-security/blob/main/docs/modules/ROOT/pages/reactive/oauth2/resource-server/opaque-token.adoc) and [here](https://github.com/spring-projects/spring-security/blob/main/docs/modules/ROOT/pages/servlet/oauth2/resource-server/opaque-token.adoc) to use the new builder instead of the now-deprecated constructor?

jzheaux (approved):

Issue #16038, opened 2024-11-04
`ServerBearerTokenAuthenticationConverter` validates parameters when not enabled


sjohnr:
`ServerBearerTokenAuthenticationConverter` validates the query parameter `access_token` when `allowUriQueryParameter` is `false`. The [spec](https://datatracker.ietf.org/doc/html/rfc6750#section-2) states that

> Resource servers MAY support this method.

for query string parameters, but does not indicate in the [Error Codes](https://datatracker.ietf.org/doc/html/rfc6750#section-3.1) section that the `access_token` parameter MUST be validated if the server doesn't support that particular method for resolving the token.

**Note:** This also applies to `DefaultBearerTokenResolver`, and includes when `allowFormEncodedBodyParameter` is set to `false`.

jonah1und1:
I created a PR for this, please feel free to review it.

PR #16039, opened 2024-11-05
Do not validate parameters in `ServerBearerTokenAuthenticationConverter` and `DefaultBearerTokenResolver` if not enabled

jonah1und1:
Closes issue: https://github.com/spring-projects/spring-security/issues/16038

### DIFF
--- a/oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolverTests.java
+++ b/oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolverTests.java
@@ -239,7 +239,6 @@ public class DefaultBearerTokenResolverTests {
 
 	@Test
 	public void resolveWhenFormParameterIsPresentAndNotSupportedThenTokenIsNotResolved() {
-		this.resolver.setAllowFormEncodedBodyParameter(false);
 		MockHttpServletRequest request = new MockHttpServletRequest();
 		request.setMethod("POST");
 		request.setContentType("application/x-www-form-urlencoded");
@@ -269,7 +268,6 @@ public class DefaultBearerTokenResolverTests {
 	// gh-16038
 	@Test
 	void resolveWhenRequestContainsTwoAccessTokenFormParametersAndSupportIsDisabledThenTokenIsNotResolved() {
-		this.resolver.setAllowFormEncodedBodyParameter(false);
 		MockHttpServletRequest request = new MockHttpServletRequest();
 		request.setMethod("POST");
 		request.setContentType("application/x-www-form-urlencoded");
@@ -279,11 +277,9 @@ public class DefaultBearerTokenResolverTests {
 
 	// gh-16038
 	@Test
-	void resolveWhenRequestContainsTwoAccessTokenQueryParameterAndSupportIsDisabledThenTokenIsNotResolved() {
-		this.resolver.setAllowUriQueryParameter(false);
+	void resolveWhenRequestContainsTwoAccessTokenQueryParametersAndSupportIsDisabledThenTokenIsNotResolved() {
 		MockHttpServletRequest request = new MockHttpServletRequest();
 		request.setMethod("GET");
-		request.setQueryString("access_token=" + TEST_TOKEN);
 		request.addParameter("access_token", "token1", "token2");
 		assertThat(this.resolver.resolve(request)).isNull();
 	}
--- a/oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolverTests.java
+++ b/oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolverTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -107,6 +107,7 @@ public class DefaultBearerTokenResolverTests {
 
 	@Test
 	public void resolveWhenValidHeaderIsPresentTogetherWithFormParameterThenAuthenticationExceptionIsThrown() {
+		this.resolver.setAllowFormEncodedBodyParameter(true);
 		MockHttpServletRequest request = new MockHttpServletRequest();
 		request.addHeader("Authorization", "Bearer " + TEST_TOKEN);
 		request.setMethod("POST");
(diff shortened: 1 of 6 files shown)


oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/web/server/authentication/ServerBearerTokenAuthenticationConverterTests.java
```diff
@@ -217,6 +219,15 @@ void resolveWhenQueryParameterHasMultipleAccessTokensThenOAuth2AuthenticationExc
 
 	}
 
+	// gh-16038
+	@Test
+	void resoleWhenAllowUriQueryParameterIsFalseThenQueryParameterIsIgnored() {
```
sjohnr, line 224: I think we would be missing a similar test to the ones you added in `DefaultBearerTokenResolverTests`, that ensure when two access tokens are present and `allowUriQueryParameter` is default (`false`) no exception is thrown?

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolverTests.java
```diff
@@ -261,6 +266,28 @@ public void resolveWhenQueryParameterIsPresentAndNotSupportedThenTokenIsNotResol
...
+		request.setMethod("POST");
+		request.setContentType("application/x-www-form-urlencoded");
+		request.addParameter("access_token", "token1", "token2");
+		assertThat(this.resolver.resolve(request)).isNull();
+	}
+
+	// gh-16038
+	@Test
+	void resolveWhenRequestContainsTwoAccessTokenQueryParameterAndSupportIsDisabledThenTokenIsNotResolved() {
+		this.resolver.setAllowUriQueryParameter(false);
```
sjohnr, line 283: This line can be omitted since it is the default.

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolverTests.java
```diff
@@ -261,6 +266,28 @@ public void resolveWhenQueryParameterIsPresentAndNotSupportedThenTokenIsNotResol
 		assertThat(this.resolver.resolve(request)).isNull();
 	}
 
+	// gh-16038
+	@Test
+	void resolveWhenRequestContainsTwoAccessTokenFormParametersAndSupportIsDisabledThenTokenIsNotResolved() {
+		this.resolver.setAllowFormEncodedBodyParameter(false);
```
sjohnr, line 272: This line can be omitted since it is the default.

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolverTests.java
```diff
@@ -235,6 +239,7 @@ public void resolveWhenPostAndFormParameterIsSupportedAndQueryParameterIsPresent
 
 	@Test
 	public void resolveWhenFormParameterIsPresentAndNotSupportedThenTokenIsNotResolved() {
+		this.resolver.setAllowFormEncodedBodyParameter(false);
```
sjohnr, line 242: This line is not needed since it is the default.

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/web/DefaultBearerTokenResolverTests.java
```diff
@@ -261,6 +266,28 @@ public void resolveWhenQueryParameterIsPresentAndNotSupportedThenTokenIsNotResol
...
+		MockHttpServletRequest request = new MockHttpServletRequest();
+		request.setMethod("POST");
+		request.setContentType("application/x-www-form-urlencoded");
+		request.addParameter("access_token", "token1", "token2");
+		assertThat(this.resolver.resolve(request)).isNull();
+	}
+
+	// gh-16038
+	@Test
+	void resolveWhenRequestContainsTwoAccessTokenQueryParameterAndSupportIsDisabledThenTokenIsNotResolved() {
```
sjohnr, line 282: "Parameter" is missing an "s":

```suggestion
	void resolveWhenRequestContainsTwoAccessTokenQueryParametersAndSupportIsDisabledThenTokenIsNotResolved() {
```

sjohnr (changes requested):
Thanks for the PR @jonah1und1! I have provided some feedback on tests below.

jonah1und1:
Thanks for the feedback!
I incorporated your suggestions and added some code to fix failing tests in `OAuth2ResourceServerConfigurerTests.java` and `OAuth2ResourceServerBeanDefinitionParserTests.java`, please feel free to comment.

sjohnr:
Thanks for providing updates @jonah1und1! I wanted to let you know an update from my side as well: I spoke with a team member on our approach here to validate parameters up front. I realize I should have done this prior to responding to the original PR, as the feedback I received pointed out an alternative to this approach.

The alternative approach is to actually have multiple implementations of `BearerTokenResolver` instead of one implementation with multiple `boolean` flags. This would allow a delegating implementation that composes whichever options the user wants to enable. However, there would still be two strategies for composing multiple resolvers together that result in different outcomes:

1. Try multiple delegates and find the first non-null returning one. (similar to the current implementation)
2. Try all delegates and throw an error if more than one returns something. (similar to this PR)

Having the option to do one or the other of these is clearly more flexible, at the expense of requiring user-defined delegation logic. So we would need to provide one of the delegating patterns above as well, and would need to choose which one to provide. In any case, if we took this approach, we would have to go back to the drawing board on this PR.

Long story short, I think I'd like to pause on this for a bit and think about it more before deciding how to move forward. We don't want to make a change only to realize it was the wrong change, so I think it best to hold off and not merge anything into the upcoming release. I apologize for any inconvenience, and thank you for contributing what you have so far. Let me know if anything I said doesn't make sense.

sjohnr:
@jonah1und1 thanks for your patience. I have circled back to this and reviewed the changes and goals of the issue. I have decided to move forward as we originally discussed, by incorporating your changes and adding a polish commit. I did not opt to move to a new approach, as I don't want to "upset the apple cart" and require moving to a new implementation/strategy (multiple delegates) just to address this issue. It is possible the long term solution is to still deprecate the `DefaultBearerTokenResolver` and move to a delegating strategy but we can wait for more feedback and determine if it's necessary in the future.

This PR is merged into `6.3.x` as da94fbe4315556e8541d194df8a815817698fe9e with polish commit 3c0fef59b547359ac3e5824c6b56694c3cb7b8dd.

**Note** that the polish commit aligns the implementation of `DefaultBearerTokenResolver` with `ServerBearerTokenAuthenticationConverter` and the changes from gh-15819 which are merged into `main` as an enhancement.

jonah1und1:
Thank you for considering my changes and merging them :+1:

Issue #16047, opened 2024-11-07
Circular reference when exposing a AuthenticationProvider with a dependency to HttpSecurity 


dru1:
Hi,

there seems to be a strange behavior, when there is a Security Configuration which exposes `AuthenticationProvider`s as Spring `@Bean`s, which use  `HttpSecurity` as a method argument. The application does not start due to circular references, depending on which Spring Security Version is in use:

### Spring Boot 3.2.7, Spring Security 6.2.5:

If there is only one `AuthenticationProvider` there **is** a circular reference: 

```java
@Configuration
@EnableWebSecurity
public class SecurityConfiguration {

    // Security configuration omitted (it doesn't matter)
    // Triggers a circular reference, when only one AuthenticationProvider bean is present

    @Bean
    public AnonymousAuthenticationProvider anonymousAuthenticationProviderOne(@Nonnull HttpSecurity http) throws Exception {
        return new AnonymousAuthenticationProvider("helloWorld1");
    }

}
```

Exception:
```
***************************
APPLICATION FAILED TO START
***************************

Description:

The dependencies of some of the beans in the application context form a cycle:

   org.springframework.security.config.annotation.web.configuration.WebSecurityConfiguration (field private org.springframework.security.config.annotation.web.builders.HttpSecurity org.springframework.security.config.annotation.web.configuration.WebSecurityConfiguration.httpSecurity)
┌─────┐
|  org.springframework.security.config.annotation.web.configuration.HttpSecurityConfiguration.httpSecurity defined in class path resource [org/springframework/security/config/annotation/web/configuration/HttpSecurityConfiguration.class]
↑     ↓
|  anonymousAuthenticationProviderOne defined in class path resource [org/example/SecurityConfiguration.class]
└─────┘
```

If there is more than one `AuthenticationProvider`, everything is fine:

```java
@Configuration
@EnableWebSecurity
public class SecurityConfiguration {

    // Security configuration omitted (it doesn't matter)
    // No circular reference is triggered, when two AuthenticationProvider beans are present

    @Bean
    public AnonymousAuthenticationProvider anonymousAuthenticationProviderOne(@Nonnull HttpSecurity http) throws Exception {
        return new AnonymousAuthenticationProvider("helloWorld1");
    }

    @Bean
    public AnonymousAuthenticationProvider anonymousAuthenticationProviderTwo(@Nonnull HttpSecurity http) throws Exception {
        return new AnonymousAuthenticationProvider("helloWorld2");
    }

[...]

PR #16050, opened 2024-11-08
Delay initialization AuthenticationProvider in Global Authentication

kse-music:
Related #16047

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/InitializeAuthenticationProviderBeanManagerConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/InitializeAuthenticationProviderBeanManagerConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,8 +16,7 @@
 
 package org.springframework.security.config.annotation.authentication.configuration;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.Arrays;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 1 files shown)


jzheaux:
Given that we don't want to support the circular reference described in #16047, is there still value in making this change?

kse-music:
I think it may be useful, because this can avoid the instantiation of AuthenticationProvider caused by the GlobalAuthenticationConfigurer configuration when there are multiple AuthenticationProvider beans in the context. Since there are multiple AuthenticationProvider beans, GlobalAuthenticationConfigurer will not set the AuthenticationProvider, so there is no need to instantiate it here early.

config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/InitializeAuthenticationProviderBeanManagerConfigurer.java
```diff
@@ -94,25 +95,26 @@ private <T> List<BeanWithName<T>> getBeansWithName(Class<T> type) {
 			String[] beanNames = InitializeAuthenticationProviderBeanManagerConfigurer.this.context
 				.getBeanNamesForType(type);
 			for (String beanName : beanNames) {
-				T bean = InitializeAuthenticationProviderBeanManagerConfigurer.this.context.getBean(beanName, type);
-				beanWithNames.add(new BeanWithName<T>(bean, beanName));
+				Supplier<T> bean = () -> InitializeAuthenticationProviderBeanManagerConfigurer.this.context
```
jzheaux, line 98: Instead of placing a supplier here, could we instead change the configure method above in the following two ways:

1. Don't collect the beans; instead call `ApplicationContext#getBeanNamesForType` - this should eliminate `BeanWithName` and `getBeansWithName` altogether
2. Look up the actual bean after the validation passes

jzheaux (changes requested):
Thanks, @kse-music! I've left some feedback inline.

kse-music:
@jzheaux I did it as you expected, and also I create another PR #16144 doing the same thing

jzheaux (approved):

jzheaux:
Thanks, @kse-music! This is now merged into `6.3.x` and `main`

PR #16071, opened 2024-11-12
Add getDelegate() method to ObservationAuthenticationManager

mtommila:
Add a getDelegate() method to ObservationAuthenticationManager to get the underlying AuthenticationManager.

This may be useful when configuring security if you need to get the ProviderManager to customize it. If you try to get it from HttpSecurity.getSharedObject(AuthenticationManager.class) but the ProviderManager has been wrapped with an ObservationAuthenticationManager then there's no easy way to get the ProviderManager otherwise.

### DIFF
--- a/core/src/main/java/org/springframework/security/authentication/ObservationAuthenticationManager.java
+++ b/core/src/main/java/org/springframework/security/authentication/ObservationAuthenticationManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2023 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -66,6 +66,16 @@ public final class ObservationAuthenticationManager implements AuthenticationMan
 	public void setObservationConvention(ObservationConvention<AuthenticationObservationContext> convention) {
 		Assert.notNull(convention, "The observation convention cannot be null");
 		this.convention = convention;
+	}
+
+	/**
+	 * Gets the delegate
+	 * @return The delegate
+	 *
+	 * @since 6.4
+	 */
+	public AuthenticationManager getDelegate() {
+		return this.delegate;
 	}
 
 }
--- a/core/src/test/java/org/springframework/security/authentication/ObservationAuthenticationManagerTests.java
+++ b/core/src/test/java/org/springframework/security/authentication/ObservationAuthenticationManagerTests.java
@@ -99,4 +99,9 @@ public class ObservationAuthenticationManagerTests {
 			.isThrownBy(() -> this.tested.setObservationConvention(null));
 	}
 
+	@Test
+	void getDelegate() {
+		assertThat(this.tested.getDelegate()).isSameAs(this.authenticationManager);
+	}
+
 }


jzheaux:
Thanks for the suggestion, @mtommila. Instead of adding a getter to `ObservationAuthenticationManager`, you can construct an `AuthenticationManager` yourself and wrap it:

```java
@Bean 
AuthenticationManager authenticationManager(ObservationRegistry registry) {
    // ... prepare authentication providers
    ProviderManager toObserve = new ProviderManager(providers);
    return new ObservationAuthenticationManager(registry, toObserve);
}
```

In this way, you'd have direct access to the underlying manager in addition to having more access to its configuration.

If that doesn't seem to help, can you tell me more about your situation and what you want to achieve by getting the delegate `AuthenticationManager`?

mtommila:
Hello,

Constructing my own `AuthenticationManager` and especially all the authentication providers is exactly what I want to _avoid_. The situation is that we are using a third-party product (the CoreMedia content management system), which is built on the Spring framework, including Spring security. It provides a Spring security setup, which we are customizing to add OAuth2 OIDC SSO to it.

So, the `AuthenticationManager` and authentication providers are set up elsewhere (by the underlying CoreMedia system, and by our customization of adding OAuth2 SSO to it). But what I need to customize further is the OAuth2 part slightly so that it works with the authentication provider, and the underlying CoreMedia system. It needs a bit of glue logic to get it to work.

In my glue logic, I need to get the existing, pre-configured `AuthenticationManager` (that is a `ProviderManager`) and from there, get the `AuthenticationProvider`s to find the one of them that is the `OidcAuthorizationCodeAuthenticationProvider` and wrap it with a wrapper that contains my custom glue logic.

The logic is something as follows:
```
  public SecurityFilterChain aadSecurityFilterChain(HttpSecurity http) throws Exception {
    http.securityMatcher(loginUrl, redirectUrl).authorizeHttpRequests((authorize) -> authorize.requestMatchers(loginUrl, redirectUrl).permitAll()).oauth2Login((oauth2Login) -> oauth2Login.authorizationEndpoint((authorizationEndpoint) -> authorizationEndpoint.baseUri(baseUrl)).loginProcessingUrl(redirectUrl)
      .successHandler(successHandler)).sessionManagement((sessionManagement) -> sessionManagement.sessionAuthenticationStrategy(sessionAuthenticationStrategy));
    http.with(new SecurityConfigurerAdapter<>() {
      @Override
      public void configure(HttpSecurity http) {
        wrapAuthenticationProvider(http.getSharedObject(AuthenticationManager.class));
      }
    }, Customizer.withDefaults());
    return http.build();
  }

  @Bean
  public AADAuthenticationProviderWrapper aadProviderWrapper() {
    return new AADAuthenticationProviderWrapper();
  }

  private void wrapAuthenticationProvider(AuthenticationManager authenticationManager) {
    ProviderManager providerManager = (ProviderManager) authenticationManager;
    ListIterator<AuthenticationProvider> iterator = providerManager.getProviders().listIterator();
    boolean wrapped = false;
    while (iterator.hasNext()) {
      AuthenticationProvider authenticationProvider = iterator.next();
      if (authenticationProvider instanceof OidcAuthorizationCodeAuthenticationProvider) {
        AADAuthenticationProviderWrapper aadAuthenticationProviderWrapper = aadProviderWrapper();
        aadAuthenticationProviderWrapper.setAuthenticationProvider(authenticationProvider);
        iterator.set(aadAuthenticationProviderWrapper);
        wrapped = true;
      }
    }
    if (!wrapped) {
      throw new IllegalArgumentException("OidcAuthorizationCodeAuthenticationProvider not found");
    }
  }
```

Is there a better way to wrap the existing OIDC authentication provider?

jzheaux:
Gotcha. One of the strategies that can help when post-processing code that you don't have control over is registering an `ObjectPostProcessor`. It's like `BeanPostProcessor`, but for individual objects managed by the DSL.

It looks something like this:

```java
http
    .oauth2Login((oauth2) -> oauth2
        .addObjectPostProcessor(new ObjectPostProcessor<>() {
            @Override
            public AuthenticationProvider postProcess(AuthenticationProvider provider) {
                if (provider instanceof OidcAuthorizationCodeAuthenticationProvider oidc) {
                    return wrap(oidc);
                }
                return provider; 
            }
        }
    )
```

How well does that work for you?

mtommila:
Hello,

Yes, adding such an `ObjectPostProcessor` would seem to work. Thank you, I was not aware of such a possibility.

jzheaux:
Awesome! Glad it worked for you. I'll close this now; feel free to reach out again if you have trouble.

Issue #15511, opened 2024-08-03
`InMemory[Reactive]OAuth2AuthorizedClientService` does not support changes to the `ClientRegistration` at runtime


kzander91:
**Describe the bug**
We're calling an API that requires JWTs for authentication. The JWTs are obtained from an authorization server using Client Credentials Grant.
The authorization server requires us to periodically rotate the client secret. We store the secret in a separate configuration service (where we can update it out-of-band) and we have implemented a custom `ReactiveClientRegistrationRepository` that retrieves the secret from there.

This setup works fine until the secret is changed: We noticed that our app keeps using the _previous_ secret, even though our `ReactiveClientRegistrationRepository` is returning the _new_ secret.

We have tracked this down to `InMemoryReactiveOAuth2AuthorizedClientService#loadAuthorizedClient` which internally uses our repository implementation:
https://github.com/spring-projects/spring-security/blob/6adc3b3d99be1452b2900e1d5f9a05cee01157fa/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryReactiveOAuth2AuthorizedClientService.java#L60-L67
Note how the `ClientRegistration` retrieved from our `clientRegistrationRepository` (containing the _current_ secret) is effectively ignored in favor of the `ClientRegistration` cached in the `authorizedClients` map (as a member of `OAuth2AuthorizedClient`). This means that the client secret that was current at the time of the first token request is used forever, even if the `ClientRegistration` changes afterwards.
After we rotated the secret, the authorization server rejects the previous one, resulting in all token requests to fail shortly after a secret rotation.

**Expected behavior**
The `InMemoryReactiveOAuth2AuthorizedClientService` should consider the fact that the `ClientRegistration` may change at runtime and thus not ignore it after fetching it from the registration repository.

**Workaround**
* First quick workaround was to reboot our app after every secret rotation.
* Next, we implemented our own `ReactiveOAuth2AuthorizedClientService` that is basically a copy of `InMemoryReactiveOAuth2AuthorizedClientService`, but with the following modification to `loadAuthorizedClient`:
```java
return this.clientRegistrationRepository.findByRegistrationId(clientRegistrationId)
  .mapNotNull(clientRegistration -> {
    OAuth2AuthorizedClientId id = new OAuth2AuthorizedClientId(clientRegistrationId, principalName);
    OAuth2AuthorizedClient cachedAuthorizedClient = this.authorizedClients.get(id);
[...]

PR #16133, opened 2024-11-19
Always return current ClientRegistration in `loadAuthorizedClient`

kzander91:
This changes `InMemoryOAuth2AuthorizedClientService.loadAuthorizedClient` (and its reactive counterpart) to always return `OAuth2AuthorizedClient` instances containing the current `ClientRegistration` as obtained from the `ClientRegistrationRepository`.

Before this change, the first `ClientRegistration` instance was cached, with the effect that any changes made in the `ClientRegistrationRepository` (such as a new client secret) would not have taken effect.

Closes gh-15511

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryOAuth2AuthorizedClientService.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryOAuth2AuthorizedClientService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -80,7 +80,13 @@ public final class InMemoryOAuth2AuthorizedClientService implements OAuth2Author
 		if (registration == null) {
 			return null;
 		}
-		return (T) this.authorizedClients.get(new OAuth2AuthorizedClientId(clientRegistrationId, principalName));
+		OAuth2AuthorizedClient cachedAuthorizedClient = this.authorizedClients
+			.get(new OAuth2AuthorizedClientId(clientRegistrationId, principalName));
+		if (cachedAuthorizedClient == null) {
+			return null;
+		}
+		return (T) new OAuth2AuthorizedClient(registration, cachedAuthorizedClient.getPrincipalName(),
+				cachedAuthorizedClient.getAccessToken(), cachedAuthorizedClient.getRefreshToken());
 	}
 
 	@Override
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryReactiveOAuth2AuthorizedClientService.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryReactiveOAuth2AuthorizedClientService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2019 the original author or authors.
+ * Copyright 2002-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryOAuth2AuthorizedClientService.java
```diff
@@ -80,7 +80,13 @@ public <T extends OAuth2AuthorizedClient> T loadAuthorizedClient(String clientRe
...
 			return null;
 		}
-		return (T) this.authorizedClients.get(new OAuth2AuthorizedClientId(clientRegistrationId, principalName));
+		OAuth2AuthorizedClient cachedAuthorizedClient = this.authorizedClients
+				.get(new OAuth2AuthorizedClientId(clientRegistrationId, principalName));
+		if (cachedAuthorizedClient == null) {
+			return null;
+		}
+		return (T) new OAuth2AuthorizedClient(registration, cachedAuthorizedClient.getPrincipalName(),
+				cachedAuthorizedClient.getAccessToken(), cachedAuthorizedClient.getRefreshToken());
```
kzander91, line 89: Here we're no longer returning the instance that was saved in the map on `saveAuthorizedClient`, which means that if `T` is a subclass of `OAuth2AuthorizedClient`, this may cause a `ClassCastException`.
Is this an acceptable risk?
For context, the JDBC/R2DBC implementations can be configured with mappers to allow custom subtypes.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryReactiveOAuth2AuthorizedClientService.java
```diff
@@ -62,8 +62,19 @@ public <T extends OAuth2AuthorizedClient> Mono<T> loadAuthorizedClient(String cl
...
+					if (cachedAuthorizedClient == null) {
+						return null;
+					}
+					// @formatter:off
+					return new OAuth2AuthorizedClient(clientRegistration,
+							cachedAuthorizedClient.getPrincipalName(),
+							cachedAuthorizedClient.getAccessToken(),
+							cachedAuthorizedClient.getRefreshToken());
+					// @formatter:on
+				});
```
kzander91, line 77: @sjohnr suggested in https://github.com/spring-projects/spring-security/issues/15511#issuecomment-2484441982 that `flatMap` was preferred here, but I disagree, because a single `mapNotNull` surely is less expensive.
However, if code style/consistency is more important here, I'm happy to refactor this.

kzander91:
I configured my IntelliJ code style according to https://github.com/spring-projects/spring-framework/wiki/IntelliJ-IDEA-Editor-Settings. Some formatting and imports changed in the tests.
Did I configure something incorrectly, or were the files just not formatted correctly?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryOAuth2AuthorizedClientService.java
sjohnr: I think the cast to `T` is still necessary given the contract of `OAuth2AuthorizedClientService`. It's not ideal, but the contract is what it is at this time. Do you see a way to avoid the cast?

kzander91: Unless we allow users to configure some sort of mapper to customize the type (like the JDBC/R2DBC implementations), I don't think we can avoid it.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/InMemoryReactiveOAuth2AuthorizedClientService.java
kzander91: Okay then I'll leave it as it is. Feel free to polish afterwards.

kzander91:
@sjohnr Thanks for the review, I ran `gradle format` and have pushed the changes.

sjohnr:
Thanks @kzander91! This is now merged into `6.2.x` as 73f3f75712eb8ba0fdcb11bda43800e5fa7cc4fd with polish 0eb6acde967d5c96fa7b9e7646e8c8ae4170d85b.

PR #16143, opened 2024-11-21
Clarify where "withDefaults()" comes from

pfconrey:
Update documentation to clarify where "withDefaults()" is defined. It took me a while to figure out how to implement the example in the docs.


jzheaux:
Thanks, @pfconrey, for the clarification. 

In preparation for merging, can you please rebase so that there is only one commit in your PR?

pfconrey:
After looking through the code base, there are dozens, if not hundreds of locations where this needs to be clarified. I will withdraw this PR and see if there's time to take up making all the necessary changes.

PR #16144, opened 2024-11-21
Do not eagerly construct UserDetailsService bean in Global Authentication

kse-music:
Related to #16050

As the [comments ](https://github.com/spring-projects/spring-security/pull/16050#issuecomment-2469417774)said


jzheaux:
Thanks, @kse-music, this is now merged into `6.3.x` and `main`

PR #16146, opened 2024-11-22
Spelling error in opensaml.adoc

OrangeDog:


jzheaux:
Thanks for the corrections, @OrangeDog. This is now merged into `main`

PR #16164, opened 2024-11-25
Address BouncyCastle's deprecated AESFastEngine usage

fjacobs:
- Update AESEngine to use the default AES engine, following BouncyCastle's recommendations (see release-1-56 of changelog: https://www.bouncycastle.org/download/bouncy-castle-java/?filter=java%3Drelease-1-56).
- Migrate to the latest API 'newInstance()' method to allow removal of method level @SuppressWarnings("deprecation")
- Remove @SuppressWarnings("deprecation"), as it is not needed anymore

### DIFF
--- a/crypto/src/main/java/org/springframework/security/crypto/encrypt/BouncyCastleAesCbcBytesEncryptor.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/encrypt/BouncyCastleAesCbcBytesEncryptor.java
@@ -16,12 +16,9 @@
 
 package org.springframework.security.crypto.encrypt;
 
-import java.util.function.Supplier;
-
 import org.bouncycastle.crypto.BufferedBlockCipher;
 import org.bouncycastle.crypto.InvalidCipherTextException;
 import org.bouncycastle.crypto.engines.AESEngine;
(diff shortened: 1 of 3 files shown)


sjohnr:
@fjacobs thanks for opening this. I have reviewed the release notes you referenced and agree that the recommendation in certain cases is to use `AESEngine`. However, I think this could be at a minimum a performance impact on applications, so I am not certain we want to apply this without considering that. I'm unfamiliar with the runtime performance characteristics of the different engines other than what is stated in the docs about three levels of performance (slow, middle, fast). Do you have a sense of what the impact of this change might be in terms of performance, or an idea of how we might best gauge that?

sjohnr:
@fjacobs do you have any thoughts regarding the [above comment](https://github.com/spring-projects/spring-security/pull/16164#issuecomment-2501657445)?

rwinch:
Thanks for the PR @fjacobs! I've rebased this from main and removed the compatibility checks since this will be merged into Spring Security 7 and I'd prefer the deprecated usage to be removed entirely. 

This will be merged into main as soon as the build passes. Thanks again for your contribution 😄

Issue #16165, opened 2024-11-25
Code example in documentation is incorrect


sjohnr:
The code example for creating a `OAuth2ClientHttpRequestInterceptor` is incorrect in [Access Protected Resources for the Current User](https://docs.spring.io/spring-security/reference/servlet/oauth2/index.html#oauth2-client-access-protected-resources-current-user). The example:

```java
OAuth2ClientHttpRequestInterceptor requestInterceptor =
		new OAuth2ClientHttpRequestInterceptor(authorizedClientManager, clientRegistrationIdResolver());
return RestClient.builder()
		.requestInterceptor(requestInterceptor)
		.build();
```

should instead be:

```java
OAuth2ClientHttpRequestInterceptor requestInterceptor =
		new OAuth2ClientHttpRequestInterceptor(authorizedClientManager);
requestInterceptor.setClientRegistrationIdResolver(clientRegistrationIdResolver());

return RestClient.builder()
		.requestInterceptor(requestInterceptor)
		.build();
```

PR #16172, opened 2024-11-26
Fix OAuth2 documentation: Correct OAuth2ClientHttpRequestInterceptor usage

harpreets789:
The code example in the "Access Protected Resources for the Current User" -> "Configure RestClient with ClientHttpRequestInterceptor" section was incorrect. Updated the example to use the correct `OAuth2ClientHttpRequestInterceptor ` constructor and added a call to `setClientRegistrationIdResolver`. This ensures accurate usage of the API.

Closes gh-16165.


harpreets789:
The sample code in the documentation has been corrected by instantiating the `OAuth2ClientHttpRequestInterceptor` with only the `authorizedClientManager `and setting the `clientRegistrationIdResolver `separately using the `setClientRegistrationIdResolver()` method.

 Fixing the previous incorrect constructor usage in both the Java and Kotlin code samples.

sjohnr:
Thanks @harpreets789, this is now merged into `main`!

Issue #15982, opened 2024-10-24
Fail when several filter chains have the same securityMatcher


franticticktick:
Related [https://github.com/spring-projects/spring-security/issues/15220](https://github.com/spring-projects/spring-security/issues/15220)
```
@Bean 
@Order(0)
SecurityFilterChain app(HttpSecurity http) throws Exception {
    http
        .securityMatcher("/app/**")
        .authorizeHttpRequests(...)
        .formLogin(...)

    return http.build();
}

@Bean 
@Order(1)
SecurityFilterChain api(HttpSecurity http) throws Exception {
    http
    	.securityMatcher("/app/**")
        .authorizeHttpRequests(...)
        .httpBasic(...)

    return http.build();
}
```
Is it correct to allow filter chains with the same matcher to be created? As far as I understand, this is the same case.

jzheaux:
Interestingly, `DefaultFilterChainValidator` already contains this check. I think it could be valuable to change the way `WebSecurity` works to use this filter chain validator. I'm not sure why it was originally excluded, so it may not work; however, I think it's worth considering.

franticticktick:
`FilterChainProxy` does not validate itself after the bean is created, because the `filterChainValidator` is `NullFilterChainValidator`, not `DefaultFilterChainValidator`. 
```
private FilterChainValidator filterChainValidator = new NullFilterChainValidator();
```
It seems that `DefaultFilterChainValidator` it is not used anywhere.

jzheaux:
It's used by the XML support. I'm not sure why it's not used by the Java support. I think it would be reasonable to try using it to close this ticket.

franticticktick:
`DefaultFilterChainValidator` is implemented in the **config** module, and `FilterChainProxy` is implemented in the **web** module, perhaps this is the reason. I can make a private copy of the `DefaultFilterChainValidator` in the **web** module, but ideally I would want to remove the duplication.

jzheaux:
I don't think that will be needed since where the validator is set is in a config class (`WebSecurity`, I believe). IOW, we should be able to do:

```java
filterChainProxy.setFilterChainValidator(new DefaultFilterChainValidator())
```

`DefaultFilterChainValidator` would likely need to be updated with the error message enhancements already added in `WebSecurity`; otherwise I imagine it may be a drop-in replacement.

franticticktick:
[...]

PR #16186, opened 2024-11-28
Add support checking same security matchers

franticticktick:
Closes gh-15982


config/src/main/java/org/springframework/security/config/http/DefaultFilterChainValidator.java
```diff
@@ -75,11 +74,12 @@ private void checkPathOrder(List<SecurityFilterChain> filterChains) {
...
 		Iterator<SecurityFilterChain> chains = filterChains.iterator();
 		while (chains.hasNext()) {
-			RequestMatcher matcher = ((DefaultSecurityFilterChain) chains.next()).getRequestMatcher();
-			if (AnyRequestMatcher.INSTANCE.equals(matcher) && chains.hasNext()) {
-				throw new IllegalArgumentException("A universal match pattern ('/**') is defined "
-						+ " before other patterns in the filter chain, causing them to be ignored. Please check the "
-						+ "ordering in your <security:http> namespace or FilterChainProxy bean configuration");
+			if (chains.next() instanceof DefaultSecurityFilterChain securityFilterChain) {
+				if (AnyRequestMatcher.INSTANCE.equals(securityFilterChain.getRequestMatcher()) && chains.hasNext()) {
+					throw new IllegalArgumentException("A universal match pattern ('/**') is defined "
```
jzheaux, line 79: By adding this check, the one in `WebSecurity` is now redundant.  Will you please remove the related error message from `WebSecurity` and replace this one here with that one?

Partially, this is because this one refers to an XML block, which wouldn't apply in a Java Config application. Also, I want to preserve recent enhancements to the `WebSecurity` error message

config/src/main/java/org/springframework/security/config/http/DefaultFilterChainValidator.java
```diff
@@ -88,10 +88,13 @@ private void checkForDuplicateMatchers(List<SecurityFilterChain> chains) {
 		while (chains.size() > 1) {
 			DefaultSecurityFilterChain chain = (DefaultSecurityFilterChain) chains.remove(0);
 			for (SecurityFilterChain test : chains) {
-				if (chain.getRequestMatcher().equals(((DefaultSecurityFilterChain) test).getRequestMatcher())) {
-					throw new IllegalArgumentException("The FilterChainProxy contains two filter chains using the"
-							+ " matcher " + chain.getRequestMatcher() + ". If you are using multiple <http> namespace "
-							+ "elements, you must use a 'pattern' attribute to define the request patterns to which they apply.");
+				if (test instanceof DefaultSecurityFilterChain securityFilterChain) {
+					if (chain.getRequestMatcher().equals(securityFilterChain.getRequestMatcher())) {
+						throw new IllegalArgumentException("The FilterChainProxy contains two filter chains using the"
```
jzheaux, line 93: These message is XML-specific, so they arent't as relevant to Java Config users. Still, I think that each application type should see error messages that make sense to them.

To address this, perhaps this class can contain a private inner interface with methods like:

```java
void handleDuplicateRequestMatchers(chain, securityFilterChain);

void handleChainAfterAnyRequest(anyRequestChain, afterChain);
```

(don't worry so much about the names, just giving the idea)

Then, internally there would be a XML and Java Config versions of the interface. This would allow each message to be relevant to the application type. There may be other ways to achieve this as well. What I'd prefer to avoid is adding `if/else` statements throughout error handling logic.

config/src/main/java/org/springframework/security/config/http/DefaultFilterChainValidator.java
jzheaux, line 79: Also, will you please add an exception type `UnreachableFilterChainException` and throw that instead in both cases? Then Spring Boot may be able to pick it up and support it through their throwable analyzer.

web/src/main/java/org/springframework/security/web/util/matcher/OrRequestMatcher.java
```diff
@@ -81,6 +82,23 @@ public MatchResult matcher(HttpServletRequest request) {
 		return MatchResult.notMatch();
 	}
 
+	@Override
+	public boolean equals(Object o) {
```
jzheaux, line 86: Will you please put this in a preceding commit and then also add `hashCode` and `equals` for `AndRequestMatcher` in that commit?

jzheaux (changes requested):
Thanks, @franticticktick! I've left my feedback inline.

Also, if you want to add #16213 to this PR as a separate commit, I think that makes sense.

config/src/main/java/org/springframework/security/config/http/DefaultFilterChainValidator.java
franticticktick: For example, strategy:
```
interface RequestMatchersCheckingStrategy {
		void handleDuplicateRequestMatchers(SecurityFilterChain chain, SecurityFilterChain securityFilterChain);

		void handleChainAfterAnyRequest(SecurityFilterChain anyRequestChain, SecurityFilterChain afterChain);
	}
```
And its two implementations `XmlRequestMatchersCheckingStrategy` and `JavaRequestMatchersCheckingStrategy`. The `checkForDuplicateMatchers` method will look like:
```
private void checkForDuplicateMatchers(List<SecurityFilterChain> chains) {
		while (chains.size() > 1) {
			DefaultSecurityFilterChain chain = (DefaultSecurityFilterChain) chains.remove(0);
			for (SecurityFilterChain test : chains) {
				requestMatchersCheckingStrategy.handleDuplicateRequestMatchers(test, chain);
			}
		}
	}
```
I’ll just note that this interface must be public in order to be able to specify its implementation:
```
filterChainValidator.setRequestMatchersCheckingStrategy(new JavaRequestMatchersCheckingStrategy());
```

jzheaux:
@franticticktick, given #16249, I'd like to change my recommendation as that ticket would remove the notion of "after-any" unreachability. Thus, the idea of having some kind of validation reporter may be more brittle than I had first considered (since one of the handlers could be immediately deprecated).

Instead, I think a new implementation of `FilterChainValidator` would be better. In this way, the XML validation remains unaffected. I think `WebSecurityFilterChainValidator` would be a good name. Do you have any concerns about that approach?

franticticktick:
@jzheaux maybe it makes sense to open a new ticket? In this issue, I can fix bugs in `DefaultFilterChainValidator`, and add `equals` and `hashCode` to `OrRequestMatcher` in a separate commit. In general, i can start working on the `WebSecurityFilterChainValidator` in the [16213](https://github.com/spring-projects/spring-security/issues/16213).

jzheaux:
Good points, @franticticktick. 

My thought is that the improvements to `DefaultFilterChainValidator` are more a polish and doesn't necessarily require a ticket. So perhaps that simply moves to a separate commit in this PR.

Then, there should be another commit that solves #15982 which includes improving the matchers as well as adding the check to `WebSecurityFilterChainValidator`. To that end, I've pushed a commit to add `WebSecurityFilterChainValidator` already so this commit will only be focused on adding the feature.

How does that sound?

franticticktick:
@jzheaux thanks for the recommendations, I added tests for `WebSecurityFilterChainValidator` and a new exception type `UnreachableFilterChainException`.

web/src/main/java/org/springframework/security/web/UnreachableFilterChainException.java
```diff
@@ -0,0 +1,36 @@
...
+
+package org.springframework.security.web;
+
+/**
+ * Thrown if {@link SecurityFilterChain securityFilterChain} is not valid
+ *
+ * @author Max Batischev
+ * @since 6.5
+ */
+public final class UnreachableFilterChainException extends RuntimeException {
```
jzheaux, line 25: Please have this implement `IllegalArgumentException`. In this way, the unit tests can stay the same, demonstrating backward compatibility.

web/src/main/java/org/springframework/security/web/UnreachableFilterChainException.java
```diff
@@ -0,0 +1,36 @@
...
+
+	/**
+	 * Constructs an <code>UnreachableFilterChainException</code> with the specified
+	 * message.
+	 * @param message the detail message
+	 */
+	public UnreachableFilterChainException(String message) {
```
jzheaux, line 32: Can you include the blocking filter chain and the unreachable one as member variables? This will help with some of the programmatic analysis that Boot performs.

config/src/main/java/org/springframework/security/config/http/DefaultFilterChainValidator.java
```diff
@@ -53,9 +54,8 @@
 import org.springframework.security.web.servletapi.SecurityContextHolderAwareRequestFilter;
 import org.springframework.security.web.session.SessionManagementFilter;
 import org.springframework.security.web.util.matcher.AnyRequestMatcher;
-import org.springframework.security.web.util.matcher.RequestMatcher;
 
-public class DefaultFilterChainValidator implements FilterChainProxy.FilterChainValidator {
+public final class DefaultFilterChainValidator implements FilterChainProxy.FilterChainValidator {
```
jzheaux, line 58: This isn't passive, so let's leave it out of the PR. For example, it would break an application that is extending `DefaultFilterChainValidator`.

web/src/main/java/org/springframework/security/web/UnreachableFilterChainException.java
jzheaux, line 25: Typically, Security leaves exceptions open for extension. Will you please remove the `final` keyword?

config/src/test/java/org/springframework/security/config/annotation/web/configuration/WebSecurityConfigurationTests.java
```diff
@@ -323,7 +324,7 @@ public void loadConfigWhenTwoSecurityFilterChainsPresentAndSecondWithAnyRequestT
 		assertThatExceptionOfType(BeanCreationException.class)
 			.isThrownBy(() -> this.spring.register(MultipleAnyRequestSecurityFilterChainConfig.class).autowire())
 			.havingRootCause()
-			.isExactlyInstanceOf(IllegalArgumentException.class);
+			.isExactlyInstanceOf(UnreachableFilterChainException.class);
```
jzheaux, line 327: While this is more accurate of a test at this point, please leave tests that don't have to be modified as-is.

jzheaux (changes requested):
Thanks for the update, @franticticktick! I've left feedback inline.

config/src/test/java/org/springframework/security/config/annotation/web/configuration/WebSecurityConfigurationTests.java
franticticktick: Here we will have to use the `isInstanceOf` check, because `isExactlyInstanceOf` uses a check like:
```
expectedType.equals(actual.getClass()) 
```
and it will not succeed even if `UnreachableFilterChainException` extends `IllegalArgumentException`.

web/src/main/java/org/springframework/security/web/UnreachableFilterChainException.java
franticticktick: Done

web/src/main/java/org/springframework/security/web/UnreachableFilterChainException.java
franticticktick: Done

config/src/main/java/org/springframework/security/config/http/DefaultFilterChainValidator.java
franticticktick: Done

web/src/main/java/org/springframework/security/web/UnreachableFilterChainException.java
franticticktick: Done

config/src/test/java/org/springframework/security/config/annotation/web/configuration/WebSecurityConfigurationTests.java
jzheaux: Hmm, I see, so it needs to change either way. Let's please change it to `.isInstanceOf` to make it more resilient to this kind of change.

web/src/main/java/org/springframework/security/web/UnreachableFilterChainException.java
```diff
@@ -0,0 +1,51 @@
...
+
+package org.springframework.security.web;
+
+/**
+ * Thrown if {@link SecurityFilterChain securityFilterChain} is not valid.
+ *
+ * @author Max Batischev
+ * @since 6.5
+ */
+public class UnreachableFilterChainException extends IllegalArgumentException {
```
jzheaux, line 25: Sorry I didn't mention this in the last review, it's a bit of a new process. Since exceptions are serializable, they are subject to our serialization process.

Please do the following:

1. Add `private static final int serialVersionUID = SpringSecurityCoreVersion.SERIAL_VERSION_UID;` to the class
2. Activate and run `SpringSecurityCoreVersionSerializableTests#serializeCurrentVersionClasses`
3. Add the resulting `.serialized` file to the PR

jzheaux (changes requested):
Thanks for the update, @franticticktick, it looks good.

My apologies there is one item I forgot, please see my inline comment for details.

franticticktick:
@jzheaux thanks for your feedback, everything is done.

franticticktick:
@jzheaux could you please review the third commit? It solves [16213](https://github.com/spring-projects/spring-security/issues/16213)

jzheaux (approved):

jzheaux:
Thanks, @franticticktick. Another very helpful PR!

Issue #16162, opened 2024-11-25
Support fullyAuthenticated in Kotlin authorizeHttpRequests


sgrimm:
**Describe the bug**
Upgrading from Spring Boot 3.3.5 to 3.4.0 includes an upgrade to Spring Security 6.4, which deprecates the `authorizeRequests` block in the HTTP configuration DSL. The deprecation message suggests using `authorizeHttpRequests` instead. But `authorizeHttpRequests` is missing the `fullyAuthenticated` property.

`w: file:///home/runner/work/terraware-server/terraware-server/src/main/kotlin/com/terraformation/backend/auth/SecurityConfig.kt:67:7 '@Deprecated(...) fun authorizeRequests(authorizeRequestsConfiguration: AuthorizeRequestsDsl.() -> Unit): Unit' is deprecated. Since 6.4. Use authorizeHttpRequests instead.`

**To Reproduce**
In a Spring Boot 3.3.5 app, use a security configuration like
```kotlin
@Configuration
@EnableWebSecurity
class SecurityConfig {
  @Bean
  fun securityFilter(http: HttpSecurity): SecurityFilterChain {
    http {
      authorizeRequests {
        authorize("/api/**", fullyAuthenticated)
      }
    }
  }
}
```
Upgrade to Spring Boot 3.4.0 and follow the suggestion to replace `authorizeRequests` with `authorizeHttpRequests`:
```kotlin
@Configuration
@EnableWebSecurity
class SecurityConfig {
  @Bean
  fun securityFilter(http: HttpSecurity): SecurityFilterChain {
    http {
      authorizeHttpRequests {
        authorize("/api/**", fullyAuthenticated)
      }
    }
  }
}
```
Compilation will fail because `fullyAuthenticated` is undefined.

**Expected behavior**
The suggested replacement in the deprecation message should include all the functionality of the older version or there should be a migration guide describing what to use instead.

**Sample**
https://github.com/sgrimm/spring-security-fullyauthenticated

[SecurityConfig.kt in that repo](https://github.com/sgrimm/spring-security-fullyauthenticated/blob/main/src/main/kotlin/com/example/springsecuritydeprecation/SecurityConfig.kt)

**Workaround**
Define `fullyAuthenticated` in the application code:
```kotlin
val fullyAuthenticated = AuthenticatedAuthorizationManager.fullyAuthenticated<RequestAuthorizationContext>()
```

jzheaux:
[...]

PR #16190, opened 2024-11-29
Add support fullyAuthenticated to Kotlin DSL

franticticktick:
Closes gh-16162

### DIFF
--- a/config/src/main/kotlin/org/springframework/security/config/annotation/web/AuthorizeHttpRequestsDsl.kt
+++ b/config/src/main/kotlin/org/springframework/security/config/annotation/web/AuthorizeHttpRequestsDsl.kt
@@ -275,6 +275,13 @@ class AuthorizeHttpRequestsDsl : AbstractRequestMatcherDsl {
     val authenticated: AuthorizationManager<RequestAuthorizationContext> =
         AuthenticatedAuthorizationManager.authenticated()
 
+    /**
+     * Specify that URLs are allowed by users who have authenticated and were not "remembered".
+     * @since 6.5
+     */
+    val fullyAuthenticated: AuthorizationManager<RequestAuthorizationContext> =
+            AuthenticatedAuthorizationManager.fullyAuthenticated()
+
     internal fun get(): (AuthorizeHttpRequestsConfigurer<HttpSecurity>.AuthorizationManagerRequestMatcherRegistry) -> Unit {
         return { requests ->
             authorizationRules.forEach { rule ->
--- a/config/src/test/kotlin/org/springframework/security/config/annotation/web/AuthorizeHttpRequestsDslTests.kt
+++ b/config/src/test/kotlin/org/springframework/security/config/annotation/web/AuthorizeHttpRequestsDslTests.kt
@@ -27,6 +27,8 @@ import org.springframework.context.annotation.Configuration
 import org.springframework.http.HttpMethod
 import org.springframework.security.access.hierarchicalroles.RoleHierarchy
 import org.springframework.security.access.hierarchicalroles.RoleHierarchyImpl
+import org.springframework.security.authentication.RememberMeAuthenticationToken
+import org.springframework.security.authentication.TestAuthentication
 import org.springframework.security.authorization.AuthorizationDecision
 import org.springframework.security.authorization.AuthorizationManager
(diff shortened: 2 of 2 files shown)


config/src/test/kotlin/org/springframework/security/config/annotation/web/AuthorizeHttpRequestsDslTests.kt
```diff
@@ -961,4 +963,45 @@ class AuthorizeHttpRequestsDslTests {
         }
 
     }
+
+    @Test
+    fun `request when fully authenticated configured then responds ok`() {
```
jzheaux, line 968: Will you please add another test that confirms that when a security context is not fully authenticated, then Security correctly forbids? For example, using `RememberMeAuthenticationToken` to simulate a remembered authentication.

config/src/main/kotlin/org/springframework/security/config/annotation/web/AuthorizeHttpRequestsDsl.kt
```diff
@@ -275,6 +275,12 @@ class AuthorizeHttpRequestsDsl : AbstractRequestMatcherDsl {
     val authenticated: AuthorizationManager<RequestAuthorizationContext> =
         AuthenticatedAuthorizationManager.authenticated()
 
+    /**
+     * Specify that URLs are allowed by users who have authenticated and were not "remembered".
+     */
```
jzheaux, line 280: Will you please add a `@since 6.5` to this?

jzheaux (changes requested):
Thanks again, @franticticktick! I've left some feedback inline.

config/src/test/kotlin/org/springframework/security/config/annotation/web/AuthorizeHttpRequestsDslTests.kt
```diff
@@ -961,4 +964,63 @@ class AuthorizeHttpRequestsDslTests {
...
+    }
+
+    @Test
+    fun `request when fully authenticated configured and remember-me token then responds unauthorized`() {
+        this.spring.register(FullyAuthenticatedConfig::class.java).autowire()
+        val rememberMe = RememberMeAuthenticationToken("key", "user",
+                AuthorityUtils.createAuthorityList("ROLE_USER"))
+
+        this.mockMvc.post("/path") {
+            with(SecurityMockMvcRequestPostProcessors.user("user").roles("USER"))
```
jzheaux, line 989: Something doesn't quite make sense about this test. There are two `with` declarations setting the authentication, when there should only be one. Also, it is returning `isUnauthorized` when it should be `isForbidden`. Can you look into this please?

config/src/test/kotlin/org/springframework/security/config/annotation/web/AuthorizeHttpRequestsDslTests.kt
```diff
@@ -961,4 +964,63 @@ class AuthorizeHttpRequestsDslTests {
         }
 
     }
+
+    @Test
+    fun `request when fully authenticated configured then responds ok`() {
+        this.spring.register(FullyAuthenticatedConfig::class.java).autowire()
+
+        this.mockMvc.post("/path") {
```
jzheaux, line 972: It doesn't seem necessary that this is a POST. Can you change both of these to a GET? This will also remove the need to add a CSRF token, simplifying the test.

jzheaux (changes requested):
Looking good, @franticticktick. I found two more items in the test if you could address them, please.

config/src/test/kotlin/org/springframework/security/config/annotation/web/AuthorizeHttpRequestsDslTests.kt
franticticktick: Fixed.

config/src/test/kotlin/org/springframework/security/config/annotation/web/AuthorizeHttpRequestsDslTests.kt
franticticktick: Fixed.

franticticktick:
@jzheaux thanks for feedback, all your comments have been resolved.

jzheaux (approved):

PR #16194, opened 2024-11-30
Improve AuthorizationManager configuration error messages

### LINKED ISSUE #16193: Improve error messaging when use-authorization-manager and access-decision-manager-ref conflict
**Current Behavior**

The user has a project with Spring Security utilizing Access Decision Manager classes. 
They have the following in their security.xml:
```
<http use-expressions="true" access-decision-manager-ref="affirmativeBasedAccessDecisionManager">
```
The project is based on Spring 5.x and they migrate to Spring 6.x.
The user would get an error
```
org.springframework.beans.factory.parsing.BeanDefinitionParsingException: Configuration problem: AuthorizationManager cannot be used in conjunction with `access-decision-manager-ref`
```
without having any `AuthorizationManager` beans configured.

This error message is confusing because it makes the user look for AuthorizationManager usages in the project.

**Expected Behavior**

If the error message pointed the user at the right attribute, it would save investigation time for users who perform the migration. It already gives a specific attribute name for `access-decision-manager-ref`. It should also give a reference to `use-authorization-manager` and highlight that it's true by default.

**Context**

The solution if the user doesn't want to migrate onto the new `AuthorizationManager` yet is to specify

```
<http use-expressions="true" access-decision-manager-ref="affirmativeBasedAccessDecisionManager" use-authorization-manager="false">
```

Spikhalskiy:
Users that are migrating from Spring 5.x don't know that the AuthorizationManager is turned on by default in 6.x and incompatible with their existing Access Decision Manager configuration.
This improvement to the error message with help to bring their attention to the relevant option that is true by default. It will also prevent them from wasting time and looking at where they implement or wire an AuthorizationManager instance.

Closes #16193

### DIFF
--- a/config/src/main/java/org/springframework/security/config/http/AuthorizationFilterParser.java
+++ b/config/src/main/java/org/springframework/security/config/http/AuthorizationFilterParser.java
@@ -76,12 +76,12 @@ class AuthorizationFilterParser implements BeanDefinitionParser {
 	public BeanDefinition parse(Element element, ParserContext parserContext) {
 		if (!isUseExpressions(element)) {
 			parserContext.getReaderContext()
-				.error("AuthorizationManager must be used with `use-expressions=\"true\"", element);
+				.error("AuthorizationManager must be used with `use-expressions=\"true\"; Add `use-authorization-manager=\"false\"` or enable use of expressions in `<http>` block", element);
 			return null;
 		}
 		if (StringUtils.hasText(element.getAttribute(ATT_ACCESS_DECISION_MANAGER_REF))) {
 			parserContext.getReaderContext()
-				.error("AuthorizationManager cannot be used in conjunction with `access-decision-manager-ref`",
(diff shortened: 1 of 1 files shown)


Spikhalskiy:
Please feel free to advise if there is a more colloquial format for such additional information in the error messages in Spring framework

jzheaux:
Referring to https://github.com/spring-projects/spring-security/issues/16193#issuecomment-2518553669, this message is only shown for XML support, so I'm not concerned about there being a conflict. Also, the existing message already uses XML-specific language (`use-expressions="true"`).

I think that an error message should "speak the language" so the remediation is more quickly understood. If the same kind of thing exists in Java Config, that exception message would be different and refer to configuration files, for example.

Unless there is a common situation where a developer would be confused by seeing `<http>`, I would vote to keep it in the message.

Another possibility would be something like this:

```
AuthorizationManager and AccessDecisionManager cannot be used at the same time; either remove `access-decision-manager-ref` or add `use-authorization-manager="false"`
```

I don't think we need to let them know in the error message what the defaults are. We use documentation and IDE support for that. Instead, we should tell them what to do to get up and going again.

Does that address your concern?

Spikhalskiy:
This totally makes sense; thank you for the advice. I adjusted the messages, please let me know if they look better.

Issue #16177, opened 2024-11-26
Issue when using @AuthenticationPrincipal on interfaces


K-e-l-v-i-n-12:
In the current version of Spring Security (version 6.4), a changed behavior occurs when using the `@AuthenticationPrincipal` annotation on an interface. If the annotation is applied directly to the concrete implementation, everything works as expected and the `UserDetails` object is passed correctly. However, if the annotation is applied to an interface, the implementation receives a new or empty UserDetails object.

**Reproduction of the problem:**
The problem can be reproduced with the example project: [K-e-l-v-i-n-12/spring-security](https://github.com/K-e-l-v-i-n-12/spring-security)

**Steps to reproduce:**
- Start the application.
- Call `http://localhost:8080/userByInterface` (any username/password). The response is: Hello, null.
- Call `http://localhost:8080/userByImplementation` (any username/password). The response is: Hello, test.

**Actual behaviour:**
If `@AuthenticationPrincipal` is applied to an interface, the implementation receives an empty UserDetails object.

**Expected behaviour:**
The UserDetails object should be passed correctly even when an interface is used.

**Proposed solution:**
Extend support for @AuthenticationPrincipal to interfaces.

jzheaux:
Apologies. https://github.com/spring-projects/spring-security/commit/87de6cea1bff2e8700400f68186b014c1b75d32f was linked to the incorrect issue.

PR #16201, opened 2024-12-02
Add @AuthenticationPrincipal/@CurrentSecurityContext Interface Support for Expression Templates

kse-music:
Closes gh-16248

### DIFF
--- a/core/src/main/java/org/springframework/security/core/annotation/UniqueSecurityAnnotationScanner.java
+++ b/core/src/main/java/org/springframework/security/core/annotation/UniqueSecurityAnnotationScanner.java
@@ -18,6 +18,7 @@ package org.springframework.security.core.annotation;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedElement;
+import java.lang.reflect.Executable;
 import java.lang.reflect.Method;
 import java.lang.reflect.Parameter;
 import java.util.ArrayList;
@@ -83,6 +84,7 @@ import org.springframework.util.ClassUtils;
  *
  * @param <A> the annotation to search for and synthesize
  * @author Josh Cummings
+ * @author DingHao
  * @since 6.4
  */
 final class UniqueSecurityAnnotationScanner<A extends Annotation> extends AbstractSecurityAnnotationScanner<A> {
@@ -107,7 +109,7 @@ final class UniqueSecurityAnnotationScanner<A extends Annotation> extends Abstra
 	MergedAnnotation<A> merge(AnnotatedElement element, Class<?> targetClass) {
 		if (element instanceof Parameter parameter) {
 			return this.uniqueParameterAnnotationCache.computeIfAbsent(parameter, (p) -> {
-				List<MergedAnnotation<A>> annotations = findDirectAnnotations(p);
+				List<MergedAnnotation<A>> annotations = findParameterAnnotations(p);
 				return requireUnique(p, annotations);
 			});
 		}
@@ -135,6 +137,56 @@ final class UniqueSecurityAnnotationScanner<A extends Annotation> extends Abstra
 						synthesized));
 			}
 		};
+	}
+
+	private List<MergedAnnotation<A>> findParameterAnnotations(Parameter current) {
+		List<MergedAnnotation<A>> directAnnotations = findDirectAnnotations(current);
+		if (!directAnnotations.isEmpty()) {
+			return directAnnotations;
+		}
+		Executable executable = current.getDeclaringExecutable();
+		if (executable instanceof Method method) {
+			Class<?> clazz = method.getDeclaringClass();
+			Set<Class<?>> visited = new HashSet<>();
+			while (clazz != null && clazz != Object.class) {
+				directAnnotations = findClosestParameterAnnotations(method, clazz, current, visited);
+				if (!directAnnotations.isEmpty()) {
+					return directAnnotations;
+				}
+				clazz = clazz.getSuperclass();
+			}
+		}
+		return Collections.emptyList();
+	}
+
+	private List<MergedAnnotation<A>> findClosestParameterAnnotations(Method method, Class<?> clazz, Parameter current,
+			Set<Class<?>> visited) {
+		if (!visited.add(clazz)) {
+			return Collections.emptyList();
+		}
+		List<MergedAnnotation<A>> annotations = new ArrayList<>(findDirectParameterAnnotations(method, clazz, current));
+		for (Class<?> ifc : clazz.getInterfaces()) {
+			annotations.addAll(findClosestParameterAnnotations(method, ifc, current, visited));
+		}
+		return annotations;
+	}
+
+	private List<MergedAnnotation<A>> findDirectParameterAnnotations(Method method, Class<?> clazz, Parameter current) {
+		try {
+			Method methodToUse = clazz.getDeclaredMethod(method.getName(), method.getParameterTypes());
+			for (Parameter parameter : methodToUse.getParameters()) {
+				if (parameter.getName().equals(current.getName())) {
+					List<MergedAnnotation<A>> directAnnotations = findDirectAnnotations(parameter);
+					if (!directAnnotations.isEmpty()) {
+						return directAnnotations;
+					}
+				}
+			}
+		}
+		catch (NoSuchMethodException ex) {
+			// move on
+		}
+		return Collections.emptyList();
 	}
 
 	private List<MergedAnnotation<A>> findMethodAnnotations(Method method, Class<?> targetClass) {
--- a/core/src/test/java/org/springframework/security/core/annotation/UniqueSecurityAnnotationScannerTests.java
+++ b/core/src/test/java/org/springframework/security/core/annotation/UniqueSecurityAnnotationScannerTests.java
@@ -16,7 +16,13 @@
 
 package org.springframework.security.core.annotation;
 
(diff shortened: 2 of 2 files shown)


jzheaux (changes requested):
Thanks for the PR, @kse-music! I've left some feedback inline.

In addition, let's please change the argument resolvers to fall back to `MethodParameter#findParameterAnnotation` (the previous behavior) if they haven't supplied an `AnnotationTemplateExpressionDefaults`. 

The reason for this is historically the argument resolvers have simply picked the first annotation they find instead of enforcing Security's no-duplicate policy and we don't want to change that expectation in a minor release.

kse-music:
@jzheaux All feedback is done.

jzheaux:
> Unless support meta-annotations also need to provide an AnnotationTemplateExpressionDefaults?

That's correct. Please see [this section](https://docs.spring.io/spring-security/reference/servlet/authorization/method-security.html#_templating_meta_annotation_expressions) of the reference for more details.

kse-music:
> That's correct. Please see [this section](https://docs.spring.io/spring-security/reference/servlet/authorization/method-security.html#_templating_meta_annotation_expressions) of the reference for more details.

Thanks, @jzheaux . Now all checks have now passed

web/src/test/java/org/springframework/security/web/method/annotation/AuthenticationPrincipalArgumentResolverTests.java
```diff
@@ -199,6 +200,7 @@ public void resolveArgumentObject() throws Exception {
 	public void resolveArgumentCustomMetaAnnotation() throws Exception {
 		CustomUserPrincipal principal = new CustomUserPrincipal();
 		setAuthenticationPrincipal(principal);
+		this.resolver.setTemplateDefaults(new AnnotationTemplateExpressionDefaults());
```
jzheaux, line 203: Sorry, I misunderstood your question earlier. This line should not be needed. Please see my related comment for details on how to address this.

web/src/main/java/org/springframework/security/web/method/annotation/AuthenticationPrincipalArgumentResolver.java
```diff
@@ -173,9 +179,22 @@ public void setTemplateDefaults(AnnotationTemplateExpressionDefaults templateDef
...
 	 * @return the {@link Annotation} that was found or null.
 	 */
-	@SuppressWarnings("unchecked")
-	private <T extends Annotation> T findMethodAnnotation(MethodParameter parameter) {
-		return (T) this.scanner.scan(parameter.getParameter());
+	private AuthenticationPrincipal findMethodAnnotation(MethodParameter parameter) {
+		if (this.useAnnotationTemplate) {
+			return this.scanner.scan(parameter.getParameter());
+		}
+		AuthenticationPrincipal annotation = parameter.getParameterAnnotation(this.annotationType);
```
jzheaux, line 186: This algorithm does not support `aliasFor`. Since that was also added in 6.4 and is unrelated to the template expressions, let's please change everything from line 186 to 197 to the following:

```java
return MergedAnnotations.from(parameter.getParameter(), MergedAnnotations.SearchStrategy.INHERITED_ANNOTATIONS, RepeatableContainers.none())
				.get(annotationType).synthesize(MergedAnnotation::isPresent).orElse(null);
```

core/src/main/java/org/springframework/security/core/annotation/UniqueSecurityAnnotationScanner.java
```diff
@@ -137,6 +140,64 @@ private MergedAnnotation<A> requireUnique(AnnotatedElement element, List<MergedA
...
+					if (clazz != null) {
+						directAnnotations.addAll(findParameterAnnotations(method, clazz, current));
+					}
+				}
+			}
+		}
+		return directAnnotations;
+	}
+
+	private boolean isOverrideFor(Method method, Method candidate) {
```
jzheaux, line 183: Let's make sure this works by adding other methods that are close in nature, but are not overrides.

core/src/main/java/org/springframework/security/core/annotation/UniqueSecurityAnnotationScanner.java
```diff
@@ -137,6 +140,64 @@ private MergedAnnotation<A> requireUnique(AnnotatedElement element, List<MergedA
...
+					}
+				}
+			}
+		}
+		return directAnnotations;
+	}
+
+	private List<MergedAnnotation<A>> findParameterAnnotations(Parameter current) {
+		List<MergedAnnotation<A>> directAnnotations = new ArrayList<>(findDirectAnnotations(current));
+		if (directAnnotations.isEmpty()) {
```
jzheaux, line 162: To align with other Spring Security code, will you please favor returning early over nesting?

For example, instead of this:

```java
if (directAnnotations.isEmpty()) {
   // do things
}
return directAnnotations;
```

do this:

```java
if (!directAnnotations.isEmpty()) {
    return directAnnotations;
}
// do things
```

This tends to reduce nesting.

core/src/main/java/org/springframework/security/core/annotation/UniqueSecurityAnnotationScanner.java
```diff
@@ -137,6 +140,64 @@ private MergedAnnotation<A> requireUnique(AnnotatedElement element, List<MergedA
...
+			Executable executable = current.getDeclaringExecutable();
+			if (executable instanceof Method method) {
+				Class<?> clazz = method.getDeclaringClass();
+				while (clazz != null) {
+					for (Class<?> ifc : clazz.getInterfaces()) {
+						directAnnotations.addAll(findParameterAnnotations(method, ifc, current));
```
jzheaux, line 168: This approach can lead to an infinite recursion since already visited classes can be revisited. Please keep track of the classes you have already visited so that you know to skip them.

jzheaux (changes requested):
Thanks for the PR, @kse-music, great job. I've left some feedback inline.

jzheaux:
@kse-music, one more thing. I'm not comfortable releasing this big of a change on a maintenance release. So let's split this PR into two: One for the additional support in `SecurityAnnotationScanners` and one for the fallback logic in the argument resolvers. The fallback logic will be merged and released in 6.4 and the rest will release in 6.5.

web/src/main/java/org/springframework/security/web/method/annotation/AuthenticationPrincipalArgumentResolver.java
kse-music: But If do like that, parameter annotation can't resolve from interface or parent that is #16177 mentioned.By research `MergedAnnotations `source code `MergedAnnotations.SearchStrategy` only effect class or method .If the `AnnotatedElement` is parameter , only the annotations on the current element will not include inheritance.

kse-music:
@jzheaux If use fallback logic, it may not be easy to support `aliasFor` , as [comment](https://github.com/spring-projects/spring-security/pull/16201#discussion_r1872494288) said, or you write a search for parameter annotation inheritance?

web/src/main/java/org/springframework/security/web/method/annotation/AuthenticationPrincipalArgumentResolver.java
jzheaux: Please take a look at this and let me know what you think: https://github.com/spring-projects/spring-security/commit/b8bfbace5f993fc452ebe55e2eeef5a7a7d911c8

kse-music: Sorry, I misinterpreted you.

jzheaux:
I've opened https://github.com/spring-projects/spring-security/issues/16248. Would you please change the commit message to point to that ticket instead?

kse-music:
> I've opened #16248. Would you please change the commit message to point to that ticket instead?

Done

jzheaux (approved):

core/src/test/java/org/springframework/security/core/annotation/UniqueSecurityAnnotationScannerTests.java
```diff
@@ -251,6 +260,77 @@ void scanWhenClassInheritingAbstractClassNoAnnotationsThenNoAnnotation() throws
...
+
+		void add(@CustomParameterAnnotation("one") String user);
+
+		List<String> list(@CustomParameterAnnotation("two") String user);
+
+		String get(@CustomParameterAnnotation("three") String user);
+
+	}
+
+	interface OtherUserService {
```
jzheaux, line 301: If I change this to have a super interface and move the annotation parameter there, then the test fails. It doesn't seem that it searches for conflicts beyond the first level of inheritance. Am I doing something wrong?

jzheaux (changes requested):
@kse-music, thanks for your patience while I got back to this. After some testing, I have some additional feedback; are you able to address it?

core/src/test/java/org/springframework/security/core/annotation/UniqueSecurityAnnotationScannerTests.java
kse-music: Done

kse-music:
@jzheaux All feedback is done.

jzheaux (approved):

jzheaux:
Lovely, @kse-music. All merged and thank you again.

Issue #13562, opened 2023-07-18
Simplify MvcRequestMatcher construction


jzheaux:
Constructing an `MvcRequestMatcher` can be confusing since it's unclear where to obtain a `HandlerMappingIntrospector` from.

Spring Security could publish a prototype bean like so to make that easier:

```java
@Bean 
@Scope("prototype")
MvcRequestMatcher.Builder mvcRequestMatcherBuilder(HandlerMappingIntrospector introspector) {
    return new MvcRequestMatcher.Builder(introspector);
}
```

jzheaux:
Due to https://github.com/spring-projects/spring-security/issues/13794, this has been reopened to give us more time to assess the right solution.

jzheaux:
I'm closing this ticket since it is superceded by [`AntPathRequestMatcher`](https://github.com/spring-projects/spring-security/issues/16887)'s and [`MvcRequestMatcher`](https://github.com/spring-projects/spring-security/issues/16886)'s removal.

Please [use `PathPatternRequestMatcher`](https://docs.spring.io/spring-security/reference/servlet/authorization/authorize-http-requests.html#match-by-mvc) instead.

PR #16208, opened 2024-12-02
A RequestMatcherFactory API

jzheaux:
Closes gh-13562

This allows an application to directly configure how `requestMatchers(String)` works, thereby delegating the responsibility of disambiguating request patterns.

For example, in an application where Spring MVC is on `/servlet/path` and there are other servlets, there might exist two endpoints, one at `/example` and one at `/servlet/path/example`. Because Spring MVC patterns exclude the `/servlet/path` prefix, a declaration like this one:

```java
.requestMatchers("/example").permitAll()
```

becomes ambiguous.

With this PR, an application can alleviate this in one of three ways:

First, they can state that all their `String`-based request matcher endpoints are MVC endpoints, a common scenario:

```java
@Bean
RequestMatcherFactory useMvc(HandlerMappingIntrospector introspector) {
    MvcRequestMatcher.Builder builder = new MvcRequestMatcher.Builder(introspector).servletPath("/servlet/path");
    return mvc:pattern;
}
```

Spring Security will consult this bean instead of using the default behavior in `requestMatchers(String)` , `securityMatchers(String)` or `ignoringRequestMatchers(String)` methods and interpret each one as a Spring MVC request.

Second, they can construct a `MvcRequestMatcherFactory` which will check the destination of the request and use MVC matching rules if the request is targeting `DispatcherServlet` or Ant matching rules otherwise:

```java
@Bean
RequestMatcherFactory delegating(MvcRequestMatcherFactory.Builder builder) {
    return builder.servletPath("/servlet/path").build();
}
```

Or third, they can introduce logic of their own like so:

```java
@Bean
RequestMatcherFactory custom(HandlerMappingIntrospector introspector) {
    MvcRequestMatcher.Builder mvc = new MvcRequestMatcher.Builder(introspector).servletPath("/servlet/path");
    return (method, pattern) -> ("/graphql".equals(pattern)) ? 
            new AntPathRequestMatcher("/graphql") : mvc.requestMatcher(method, pattern);
}
```

Though in the third case, it will likely be less work to do the first or second option and then do:

```java
@Bean 
SecurityFilterChain security(HttpSecurity http) throws Exception {
    http
            .requestMatchers(new AntPathRequestMatcher("/graphql")).hasAuthority("graphql")
            .requestMatchers("/my", "/mvc", "/endpoints').hasAuthority("app")
        )
}
```

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/AbstractRequestMatcherRegistry.java
@@ -48,6 +48,7 @@ import org.springframework.security.web.util.matcher.DispatcherTypeRequestMatche
 import org.springframework.security.web.util.matcher.OrRequestMatcher;
 import org.springframework.security.web.util.matcher.RegexRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher;
+import org.springframework.security.web.util.matcher.RequestMatcherFactory;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
 import org.springframework.web.context.WebApplicationContext;
@@ -72,6 +73,8 @@ public abstract class AbstractRequestMatcherRegistry<C> {
 	private static final boolean mvcPresent;
 
 	private static final RequestMatcher ANY_REQUEST = AnyRequestMatcher.INSTANCE;
+
+	private final RequestMatcherFactory requestMatcherFactory = new DefaultRequestMatcherFactory();
 
 	private ApplicationContext context;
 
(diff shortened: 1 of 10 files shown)


rwinch:
I think that a user that wants to change the strategy can leverage `requestMatchers(RequestMatcher...)` rather than us adding another API (complexity). What do you think @jzheaux ?

jzheaux:
I think that's a good point, and there may be some value in just having the user use a builder directly for now:

```java
@Bean 
SecurityFilterChain security(MvcRequestMatcherFactory.Builder builder) {
    MvcRequestMatcherFactory mvc = builder.servletPath("/servlet/path").build();
    http
            .requestMatchers(mvc.patterns("/my/**", "/controllers/**")).permitAll()
         )
         // ...
    return http.build()
}
```

I like this bc it's consistent with how applications would do it if there were multiple dispatcher servlets:


```java
@Bean 
SecurityFilterChain security(MvcRequestMatcherFactory.Builder builder) {
    MvcRequestMatcherFactory mvc = builder.servletPath("/mvc").build();
    MvcRequestMatcherFactory api = builder.servletPath("/api").build();
    http
            .requestMatchers(mvc.patterns("/my/**", "/controllers/**")).permitAll()
            .requestMatchers(api.patterns("/resources/**")).hasAuthority("app")
         )
         // ...
    return http.build()
}
```

Does that address your concern? If so, I can leave out the support for the DSL picking up a `RequestMatcherFactory` bean. And I feel like that would allow removing the `RequestMatcherFactory` interface for the time being as well.

Or, are there more fundamental shifts that are on your mind?

jzheaux:
Some recent developments in the Spring Framework may allow us to pursue something simpler than this solution. So I'm going to close for now and then I'll reopen a draft PR with the relevant changes.

PR #16252, opened 2024-12-10
Avoid requesting an unnecessary attestation statement when creating a webauthn credential

ynojima:
[The attestation option in PublicKeyCredentialCreationOptions](https://www.w3.org/TR/webauthn-3/#dom-publickeycredentialcreationoptions-attestation) is a parameter that controls whether to request attestation statement from the security key. Currently, Spring Security Passkeys requests attestation statement by specifiy `direct` value to `attestation` option, but Spring Security Passkeys doesn't implement attestation statement verification[1]. Therefore, the requested attestation statement is not used at all.
Specifying `direct` to request attestation may trigger browsers to display additional privacy related dialog to users, so it is best to avoid specifying `direct` unnecessarily.

image

[1] Spring Security Passkeys uses [WebAuthnManager.createNonStrictWebAuthnManager()](https://github.com/spring-projects/spring-security/blob/main/web/src/main/java/org/springframework/security/web/webauthn/management/Webauthn4JRelyingPartyOperations.java#L99) to [setup a `WebAuthnManager` instance configured not to verify attestation statements.](https://webauthn4j.github.io/webauthn4j/en/#configuration)


ynojima:
@rwinch  Sorry, I missed updating the document. Could you merge this one too? https://github.com/spring-projects/spring-security/pull/16264

Issue #16251, opened 2024-12-09
Remove Deprecated Usages of RemoteJWKSet


jzheaux:
Nimbus has deprecated `RemoteJWKSet` in favor of using `JWKSourceBuilder`

kwondh5217:
Hi, @jzheaux 

I’d like to work on this issue as my first contribution to Spring Security. Would it be alright if I take this on?

Thank you!

franticticktick:
[Duplicate](https://github.com/spring-projects/spring-security/issues/15866).

kwondh5217:
Hi @franticticktick

I noticed that this issue is marked as a duplicate of #15866. I’d like to clarify: should I continue working on this issue, or would it be better to focus on the original issue (#15866)?

If this issue is still valid, I’d be happy to contribute and help resolve it. Please let me know how I can proceed.

Thank you!

franticticktick:
Hi @kwondh5217, i marked this as a duplicate so that contributors will pay attention to [15866]( https://github.com/spring-projects/spring-security/issues/15866), as there may be nuances that are worth considering. Nothing more :)

kwondh5217:
Hi @franticticktick,
Thank you for explaining! I’ll give it a try.
Thanks again!

PR #16296, opened 2024-12-17
Remove Deprecated Usages of RemoteJWKSet

kwondh5217:
close gh-16251


oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -414,7 +420,80 @@ private static URL toURL(String url) {
 			}
 		}
 
-		private static final class SpringJWKSetCache implements JWKSetCache {
+		private static final class SpringURLBasedJWKSource<C extends SecurityContext> implements JWKSource<C> {
+
+			private final URLBasedJWKSetSource urlBasedJWKSetSource;
```
jzheaux, line 425: Since we are doing our own implementation of `JWKSource`, it should be possible to use `RestOperations` and `Cache` directly instead of implementing additional Nimbus interfaces.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -377,11 +382,12 @@ JWSKeySelector<SecurityContext> jwsKeySelector(JWKSource<SecurityContext> jwkSou
 		}
 
 		JWKSource<SecurityContext> jwkSource(ResourceRetriever jwkSetRetriever, String jwkSetUri) {
-			if (this.cache == null) {
-				return new RemoteJWKSet<>(toURL(jwkSetUri), jwkSetRetriever);
+			URLBasedJWKSetSource urlBasedJWKSetSource = new URLBasedJWKSetSource(toURL(jwkSetUri), jwkSetRetriever);
+			if(this.cache == null) {
```
jzheaux, line 386: Let's update this so that the member variable is set to `NoOpCache`. In this way, the null checking here and downstream is unnecessary.

jzheaux (changes requested):
Thanks, @kwondh5217, for the PR! I've left some feedback inline.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
kwondh5217: Hi @jzheaux, I have a couple of questions regarding the feedback:

1. Regarding "Since we are doing our own implementation of `JWKSource`, it should be possible to use `RestOperations` and `Cache` directly instead of implementing additional Nimbus interfaces":

   Does this mean we should replace the custom `SpringJWKSetCache` implementation with direct usage of Spring's `Cache`?

2. I am also considering whether to streamline the code by either:
   - Using `RestOperations` directly, or
   - Using `RestOperationsResourceRetriever`

Below are the two possible approaches:

- **Direct `RestOperations`**:
    ```java
    private static final class SpringURLBasedJWKSource<C extends SecurityContext> implements JWKSource<C> {

        private final RestOperations restOperations;

        // Implementation logic uses restOperations directly
    }
    ```

- **Using `ResourceRetriever`**:
    ```java
    private static final class SpringURLBasedJWKSource<C extends SecurityContext> implements JWKSource<C> {

        private final ResourceRetriever resourceRetriever;

        // Implementation logic continues to leverage resourceRetriever
    }
    ```

Would it be better to reuse the already implemented `RestOperationsResourceRetriever`, or to use `RestOperations` directly within the class? 

If we go with the direct `RestOperations` approach, should the `RestOperationsResourceRetriever` implementation be removed entirely? 

Thank you for your feedback!

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -414,7 +420,80 @@ private static URL toURL(String url) {
...
+				this.urlBasedJWKSetSource = urlBasedJWKSetSource;
+				this.jwkSetCache = jwkSetCache;
+			}
+
+			@Override
+			public List<JWK> get(JWKSelector jwkSelector, SecurityContext context) throws KeySourceException {
+				if (this.jwkSetCache != null) {
+					JWKSet jwkSet = this.jwkSetCache.get();
+					if (this.jwkSetCache.requiresRefresh() || jwkSet == null) {
+						synchronized (this) {
```
holzerch, line 443: Could you please replace this `synchronized` statement with an `ReentrantLock`. Otherwise this could cause thread pinning with virtual threads.

kwondh5217: @holzerch Thank you for the feedback! 
I’ll replace the synchronized block with a ReentrantLock as suggested.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
jzheaux: Yes, `RestOperationsResourceRetriever` and `SpringJWKSetCache` would both be removed entirely.

jzheaux:
Superceded by https://github.com/spring-projects/spring-security/pull/16537

Issue #16231, opened 2024-12-07
Jwt Principal customization


vaa25:
**Expected Behavior**
I want to use method authorization annotations and @AuthenticationPrincipal clean way with custom business fields using oauth2 authorization with spring-boot-starter-oauth2-resource-server supporting all spring.security.oauth2.resourceserver.jwt.* properties.

```java
@PreAuthorize("principal.id==100")
@GetMapping
public MyPrincipal test(@AuthenticationPrincipal MyPrincipal principal) {
    return principal;
}
```
response:
```json
{"id": 123, "name": "Alex"}
```
It can be achieved if Authentication object will have custom principal object:
```java
public class JwtPrincipal implements MyPrincipal {
  private final Integer id;
  private final String name;
  public JwtPrincipal(Jwt jwt, String principalName) {
    this.id = Integer.parseInt(jwt.getClaim("id"));
    this.name=principalName;
  }
  @Override
  public Integer getId() {
    return id;
  }
  @Override
  public String getName() {
    return name;
  }
}
```
I'd like to add a principal converter as a bean or via OAuth2ResourceServerConfigurer for customization 

```java
public interface JwtPrincipalConverter {
    Object convert(Jwt jwt, String principalName);
}
```

**Current Behavior**

Current implementation sets org.springframework.security.oauth2.jwt.Jwt object as principal causing such implementation:
```java
@PreAuthorize("principal.claims['id']=='100'")
@GetMapping
public Jwt test(@AuthenticationPrincipal Jwt principal) {
    return principal;
}
```
Here I have to write 'claims' in @PreAuthorize that is not business field, and receive Jwt principal as method parameter that is not business object.

**Context**
I believe having custom Principal object makes authorization control more readable and simple and can help to accept requests with different types of authorization into same controller method.

There is some workarounds, but they are not so simple as my proposition and may require third-party library.
1. to use custom jwtAuthenticationConverter. In this case we have to write custom JwtAuthenticationToken and additional code to support spring.security.oauth2.resourceserver.jwt.* properties.
2. to apply third-party library with its own properties and a little bit less additional code.

Based on [stackoverflow question](https://stackoverflow.com/questions/79242659/springboot-oauth2-principal-object)

P.S. I have implemented this feature locally, just waiting for your approve of this idea, guys.

jzheaux:
@vaa25, thanks for the suggestion. Before moving forward, I want to make sure we are on the same page.

[...]

Issue #6237, opened 2018-12-05
Support UserDetailsService components in OAuth2 Resource Server flows


edeandrea:
Spun out from #6219.

We need the ability to apply a `UserDetailsChecker` to the OAuth2 Resource Server flows.

My initial thoughts on implementation would be to have converters (i.e. `Converter<Jwt, AbstractAuthenticationToken>` & `Converter<Jwt, Mono<AbstractAuthenticationToken>>`) that can take an underlying `Converter` and apply a `UserDetailsChecker` to it before returning the `AbstractAuthenticationToken`/`Mono<AbstractAuthenticationToken>`.

That way in the configuration DSL a consumer could choose to have this ability or not via something like

```java
http.oauth2ResourceServer()
  .jwt()
    .jwtAuthenticationConverter(new ReactiveJwtPostAutneticationChecksAuthenticationConverter(userDetailsService));
```
I'm open to discussing the correct approach on how best to integrate it and am happy to supply a PR for the correct approach.

If you'd rather me submit a PR with what I have and discuss the approach on the PR thats fine too.

jzheaux:
@edeandrea Thanks for your proposal!

The tricky part is that there isn't a `UserDetails` in play with Resource Server. Really, the typical semantics of a Resource Server are not user-based, but authority-based.

Likely, any converter we roll into master should play nicely with these semantics. For example, the default, `JwtAuthenticationConverter` returns an authentication token of type `JwtAuthenticationToken` whose principal is a `Jwt`. While nothing in the framework strictly requires either of these to be the case, changing a `Jwt` out for something like a `UserDetails` alters the underlying meaning of the `Authentication`. 

As you can imagine, it's confusing for users to hear "remember that if you use this converter, then the token is no longer the principal, meaning that such-and-such features will work differently for you now."

(And to be clear, none of the above means that the implementation you are proposing is wrong-headed. It just may mean that it isn't a fit to be rolled into master. I am going to mark this ticket for team discussion, though, to see what @rwinch et al think.)

Happily, though, as you have discovered, you can do this with your own `Converter` implementation. To confirm, are you blocked on progress without your proposed changes?

edeandrea:
Thanks @jzheaux No at this moment I'm not blocked - I just hate having to write my own custom stuff when (in theory) things really belong "up" a level :)

As a framework developer here at my company we always prefer to not have to write custom code and instead "wire" spring code together to make things work the way we want to work.

[...]

PR #16311, opened 2024-12-19
gh-16231 add JwtPrincipalConverter.java support

vaa25:
closes gh-16231

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/JwtPrincipalConverter.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/JwtPrincipalConverter.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2002-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.oauth2.server.resource.introspection;
+
+import org.springframework.security.oauth2.jwt.Jwt;
+
+/**
+ * @author Alex Vlasov
+ */
+public interface JwtPrincipalConverter {
+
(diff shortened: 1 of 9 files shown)


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/JwtPrincipalConverter.java
```diff
@@ -0,0 +1,28 @@
...
+package org.springframework.security.oauth2.server.resource.introspection;
+
+import org.springframework.security.oauth2.jwt.Jwt;
+
+/**
+ * @author Alex Vlasov
+ */
+public interface JwtPrincipalConverter {
+
+	Object convert(Jwt jwt, String principalName);
```
jzheaux, line 26: Just connecting the two conversations here. This should return a Spring Security API. That allows Spring to evolve around the API over time and reason about its contents. If it returns `Object`, we lose that ability.

jzheaux: For additional context, please see the code in [this sample application](https://github.com/spring-projects/spring-security-samples/blob/main/servlet/spring-boot/java/authentication/username-password/user-details-service/custom-user/src/main/java/example/CustomUserRepositoryUserDetailsService.java).

vaa25: Authentication.principal is an Object. Any other type would restrict principal type. If Object is not allowed by SS code convention, then could it be some marker interface? That marker interface could be a superinterface  for other principal interfaces like AuthenticatedPrincipal. Or maybe it could be AuthenticatedPrincipal itself. Sorry, I am not very familiar with SS architecture and not figure out what to do in this situation.

jzheaux: No problem. What I'd recommend is that it return `OAuth2AuthenticatedPrincipal`. This allows a converter to transmit and user-level authorities (which can be different from the authorities derived from the JWT).

If implementing extra methods is a concern, I think we can look at adding `default` implementations to some of those methods.

vaa25: Current principal is `Jwt` object. It does not implement `OAuth2AuthenticatedPrincipal`. `JwtPrincipalConverter` should convert to some `Jwt` superclass to be compatible with current principal. What to do with this?

jzheaux:
I'm going to close this as declined in favor of #6237 as I feel that will be a better driver for enabling more use cases than just this one.

PR #16312, opened 2024-12-20
Fixed the issue where the annotation parameter scan skipped first-level conflicts

kse-music:
When searching for parameter annotations, I overlooked a scenario as below:

```java
interface RemoteUserService {

    void batch(@CustomParameterAnnotation("six") String... user);

}

class UserServiceImpl implements RemoteUserService {

    @Override
    public void batch(@CustomParameterAnnotation("seven") String... user) {
      
    }

}

class DefaultUserService extends UserServiceImpl implements RemoteUserService {

    @Override
    public void batch(String... user) {
      
    }

}
```

When I run test should be throw exception :

```java
Parameter parameter = DefaultUserService.class.getDeclaredMethod("batch", String[].class).getParameters()[0];
this.parameterScanner.scan(parameter);
```


kse-music:
@jzheaux Can you help with the review?

jzheaux:
Thanks, @kse-music, this is now merged into `main`

Issue #16306, opened 2024-12-19
Pass Http Request to OAuth2AuthorizationRequestResolver#authorizationRequestCustomizer


ZIRAKrezovic:
**Expected Behavior**

It would be nice if we could use ServerWebExchange when customizing the Authorization Request using DefaultServerOAuth2AuthorizationRequestResolver#authorizationRequestCustomizer

At the moment, we only get the builder instance.

**Current Behavior**

**Context**

I would like to add additional query params to skip or enable login screen on Keycloak using kc_idp_hint query param, depending on where the user is coming from.

sjohnr:
@ZIRAKrezovic, can you please share more about your use case? What request material are you looking to use, what have you tried so far, and are there any workarounds you're aware of?

ZIRAKrezovic:
Hello @sjohnr. I have the following use-case

- Keycloak installed as a Broker to Microsoft, but also supporting internal users that do not have a SSO account
- Spring Security configured to use Keycloak

To eliminate a step for Microsoft SSO users, we pass "kc_idp_hint" to keycloak's own OAuth2 Authorization endpoint to skip the Keycloak login form and redirect to Microsoft SSO

But we'd like to have a workaround to show the login form - to stop the "kc_idp_hint" from being added to the call to OAuth2 Authorization Request in certain cases.

So the use case is as follows - if the user is not logged in and they access the application with special query parameter "show-login-page=true" the kc_idp_hint query parameter is not sent to Keycloak Authorization Endpoint.

Since I cannot access query parameters in `DefaultServerOAuth2AuthorizationRequestResolver#authorizationRequestCustomizer` I am left with the following

- Define a `ServerAuthenticationEntryPoint` to pass the "kc_idp_hint" query param to spring security "/oauth2/authorization/{registrationId}" if the source request contains query parameter "show-login-page=true"
- Reparse the redirect URL constructed by `DefaultServerOAuth2AuthorizationRequestResolver` in a `ServerRedirectStrategy` used by `OAuth2AuthorizationRequestRedirectWebFilter` and add the kc_idp_hint to the target URL if the query param was present in the request handled by `OAuth2AuthorizationRequestRedirectWebFilter`

Relevant code used

```
return http .exceptionHandling(exceptionHandlingSpec ->
                                exceptionHandlingSpec.authenticationEntryPoint(authenticationEntryPoint()))
                .oauth2Login(loginSpec ->
[...]

PR #16320, opened 2024-12-20
Add ServerWebExchange parameter to AuthorizationRequestCustomizer

kse-music:
Closes gh-16306


sjohnr:
@kse-music thanks for the PR. I apologize for the delay in reviewing. However, I don't feel confident that we want to pursue simply introducing a `BiConsumer` here. This isn't generally the best way to expose a new parameter for customization, and I don't see any precedent for it in Spring Security. For that reason, I'm going to close this PR for now and ask that you weigh in on the issue with your use case and ideas so we can find a path forward through discussion before moving to the PR stage. Thanks for understanding.

Issue #16325, opened 2024-12-22
Inconsistent constructor declaration on bean with name '_reactiveMethodSecurityConfiguration'


eugeniudedin:
**Describe the bug**

I get this warn in my spring boot reactive app:

```sh
f.a.AutowiredAnnotationBeanPostProcessor : Inconsistent constructor declaration on bean with name '_reactiveMethodSecurityConfiguration': single autowire-marked constructor flagged as optional - this constructor is effectively required since there is no default constructor to fall back to: org.springframework.security.config.annotation.method.configuration.ReactiveAuthorizationManagerMethodSecurityConfiguration(org.springframework.security.access.expression.method.MethodSecurityExpressionHandler,org.springframework.beans.factory.ObjectProvider,org.springframework.beans.factory.ObjectProvider)
```

A bean (_reactiveMethodSecurityConfiguration in this case) has a constructor marked with @Autowired, but there is no default (no-arg) constructor available.

TheRealGI:
Is there any action needed when this Log message pops up?

sberequek:
Hi, 

I also have this message:

```logs
[org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor] [determineCandidateConstructors] Inconsistent constructor declaration on bean with name '_reactiveMethodSecurityConfiguration': single autowire-marked constructor flagged as optional - this constructor is effectively required since there is no default constructor to fall back to: org.springframework.security.config.annotation.method.configuration.ReactiveAuthorizationManagerMethodSecurityConfiguration(org.springframework.security.access.expression.method.MethodSecurityExpressionHandler,org.springframework.beans.factory.ObjectProvider,org.springframework.beans.factory.ObjectProvider)
```

are there any updates regarding its resolution?

jzheaux:
Closed in favor of https://github.com/spring-projects/spring-security/pull/17197

PR #16326, opened 2024-12-22
Remove Autowire annotation on the unique constructor in ReactiveAuthorizationManagerMethodSecurityConfiguration

kse-music:
Closes gh-16325

There is a `methodSecurityExpressionHandler `bean with @Fallback in the `ReactiveAuthorizationManagerMethodSecurityConfiguration`.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/method/configuration/ReactiveAuthorizationManagerMethodSecurityConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/method/configuration/ReactiveAuthorizationManagerMethodSecurityConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2024 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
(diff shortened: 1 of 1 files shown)


rwinch:
Thanks for the PR! This is a breaking change because currently expressionHandler is optional, but by removing `@Autowired(required = false)` it is now required.

I'm going to close this issue for the reason described above.

kse-music:
@rwinch 

First of all, the `@Autowired(required = false) `annotation on the constructor does not indicate that expressionHandler is optional unless it is annotated on the parameter.

```
ReactiveAuthorizationManagerMethodSecurityConfiguration(@Autowired(required = false) MethodSecurityExpressionHandler expressionHandler,
...
```

In addition, the `ReactiveAuthorizationManagerMethodSecurityConfiguration `configuration already defines a `MethodSecurityExpressionHandler `bean as a` @Fallback` bean. Therefore, I believe the parameter `expressionHandler `being required should not be an issue.

Issue #16338, opened 2024-12-23
Fix Kotlin DSL webAuthn { }


franticticktick:
`WebAuthnDsl` is implemented incorrectly, we need to change the implementation of the `get()` method according to the usual spring security template, as in other kotlin DSLs. Now, due to incorrect implementation, the default configuration will not work.
```
        @Bean
        open fun securityFilterChain(http: HttpSecurity): SecurityFilterChain {
            http{
                formLogin { }
                webAuthn { }
            }
            return http.build()
        }
``` 
In addition, the DSL is missing `disableDefaultRegistrationPage`.

PR #16339, opened 2024-12-23
WebAuthnDsl Bug Fix

franticticktick:
Closes gh-16338

### DIFF
--- a/config/src/main/kotlin/org/springframework/security/config/annotation/web/WebAuthnDsl.kt
+++ b/config/src/main/kotlin/org/springframework/security/config/annotation/web/WebAuthnDsl.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


rwinch:
Thanks for the pull request @franticticktick!

Since this is a bug, would you please base off of 6.4.x? Also...

Would you be able to split this into two commits referencing a different issue/PR. This helps us & users to track what has changed better since issues/PRs are listed in the changelog. One commit would have the fixes for default configuration. The other would add the disableDefaultRegistrationPage. 

I'm fine with this as a single PR if each commit is referencing different issues. Alternatively, you can create multiple PRs instead of multiple issues. Again, the issues/PRs are what go in the changelog so it is valuable to have these as separate items so we can have a concise description of what has changed. Separate commits helps us if we need to do a git bisect at some point to track down a problem.

Thanks again for the PR!

franticticktick:
Thanks @rwinch , I overlooked that. Could you review this PR please?

rwinch:
Thanks for the Pull Request! This is now merged into 6.4.x and main via 882766e54fed3c62a4ab7f8e4f39f945d90905f4 and decf4def95600075faacb7b06c48fcf841bb57f1 :smile:

PR #16341, opened 2024-12-25
Fix logout code snippet for Kotlin

intotherealworld:
### Contents

- Fix logout code snippet for Kotlin: Corrected deleteCookies syntax

### Reason

- The deleteCookies property of org.springframework.security.config.annotation.web.LogoutDsl has a private access modifier, so it cannot be used in the format deleteCookies = ""


rwinch:

PR #16348, opened 2024-12-27
Remove unused code from WebSecurityConfiguration

kse-music:
Remove unused code，`webSecurityConfigurers and beanClassLoader`


rwinch:
Thanks for the PR @kse-music!

Can you please split this up into two commits and pull requests? This helps us to better track changes.

> Use ObjectProvider instead of @Autowired HttpSecurity because when the user provides a SecurityFilterChain bean, it will still inject HttpSecurity. Since HttpSecurity is a prototype bean, this avoids its instantiation.

Can you please add a test that includes a comment to the PR (e.g. gh-16348) to verify that is the case?

kse-music:
@rwinch All the feedback has been addressed as you expected.

rwinch:

Issue #16170, opened 2024-11-26
The selectJwk method of NimbusJwtEncoder class should not throw Exception when jwks size great than one


douxiaofeng99:
**Describe the bug**
I implemented a rotating JWKS using Redis, where a new JWK is generated at regular intervals, and the old JWKs are also retained for a certain period. In this scenario, the selectJwk method of NimbusJwtEncoder retrieves multiple JWKs when selecting the JWKs. This happens because the jwkSelector only sets the algorithm but does not provide any kid, leading to the exception being thrown below.
<img width="833" alt="image" src="https://github.com/user-attachments/assets/7f1c733e-0a1d-45aa-80ec-82f5f992fa74">

**To Reproduce**
Steps to reproduce the behavior.

**Expected behavior**
remove the block:
	if (jwks.size() > 1) {
			throw new JwtEncodingException(String.format(ENCODING_ERROR_MESSAGE_TEMPLATE,
					"Found multiple JWK signing keys for algorithm '" + headers.getAlgorithm().getName() + "'"));
		}
because in the last the method,   return jwks.get(0);  already use first jwk.

**Sample**

A link to a GitHub repository with a [minimal, reproducible sample](https://stackoverflow.com/help/minimal-reproducible-example).

Reports that include a sample will take priority over reports that do not.
At times, we may require a sample, so it is good to try and include a sample up front.

jzheaux:
Thanks for reaching out, @douxiaofeng99. 

Note that while you are able to specify a kid or a thumbprint in the `JwsHeader`, I agree that `NimbusJwtEncoder` could be improved when it comes to supporting key rotation when those are not specified.

I think a good start would be to introduce a setter that allows applications to indicate which JWK to pick. Something like:

```java
setJwkSelector(Converter<List<JWK>, JWK> selector);
```

Then applications could do:

```java
setJwkSelector(List::getFirst); // or
setJwkSelector(List::getLast);  // or
setJwkSelector((jwks) -> jwks.stream().max(Comparator.comparing(JWK::getIssueTime)).orElseThrow())
```

Given the following from [the RFC](https://datatracker.ietf.org/doc/html/rfc7517#section-5.1):

> The value of the "keys" parameter is an array of JWK values.  By
   default, the order of the JWK values within the array does not imply
   an order of preference among them, although applications of JWK Sets
   can choose to assign a meaning to the order for their purposes, if
   desired.

[...]

PR #16354, opened 2025-01-02
The selectJwk method of NimbusJwtEncoder class should not throw Exception when jwks size great than one #16170

douxiaofeng99:
enhancement for #16170


douxiaofeng99:
cancelled for DCO, redo.

PR #16370, opened 2025-01-08
Avoid unnecessary instantiation of HttpSecurity

kse-music:
Use ObjectProvider instead of @Autowired HttpSecurity because when the user provides a SecurityFilterChain bean, it will still inject HttpSecurity. Since HttpSecurity is a prototype bean, this avoids its instantiation.


rwinch:
Thanks for the PR! I went ahead and rebased it and pushed a minor change b21850a1345841072b02b4ac7c5ba6191b1ccec3. Once the build passes this will be merged into main. Thanks again for the PR 😄

Issue #16369, opened 2025-01-07
Use dependency injection in webauthn filters to allow for customization


levimiller-qhrtech:
**Expected Behavior**
I expect to be able to provide my own implementations for PublicKeyCredentialCreationOptionsRepository and HttpMessageConverter when using passkeys/webauthn, specifically to be able to save/load the options outside of a session.

e.g.,
```
@Bean
SecurityFilterChain webAuthnFilterChain(HttpSecurity http) throws Exception {
  return http.webAuthn(webAuthn -> { ... }).build();
}
  
@Bean
B2cPublicKeyCredentialCreationOptionsRepository customCreationOptionsRepo() {
  return new MyCustomB2cPublicKeyCredentialCreationOptionsRepository();
}
```

**Current Behavior**
The webauthn filters just initialize their own instances on construction:
https://github.com/spring-projects/spring-security/blob/main/web/src/main/java/org/springframework/security/web/webauthn/registration/PublicKeyCredentialCreationOptionsFilter.java#L59

WebAuthnRegistrationFilter at least has a setter to update it, but I'd have to use reflection to update PublicKeyCredentialCreationOptionsFilter.

**Context**
I have another server forwarding the webauthn requests to a spring boot server, and will be storing the state externally and not in a session. My current workaround is to use reflection to update the private fields to use my implementation of B2cPublicKeyCredentialCreationOptionsRepository, but I feel that dependencies should be injected so they can be customized.

rwinch:
@levimiller-qhrtech Thanks for creating this ticket. I agree with the requests you have made (we will split it into two features though).

@kse-music put together a PR, but I asked for some changes. I'm waiting to see if he will make some changes I requested (I'm hopeful he will as he is very active/excellent contributor).

kse-music:
@rwinch I have split it into two PR (#16396 and #16397 ), can you help me review the code.

rwinch:
Thanks to @kse-music this is now resolved in main!

PR #16371, opened 2025-01-08
Apply ObjectPostProcessor to the filter in WebAuthnConfigurer

kse-music:
Closes gh-16369

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/WebAuthnConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/WebAuthnConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2024 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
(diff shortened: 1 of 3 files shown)


rwinch:
Thanks for the PR @kse-music However, I'd like it split into two features even though it is in a single ticket. This helps us to track changes more accurately.

One of the pull requests would focus on configuring `PublicKeyCredentialCreationOptionsRepository`. I'd prefer to have a bit more first class support for configuring `PublicKeyCredentialCreationOptionsRepository`. The configuration should have an explicit property for `PublicKeyCredentialCreationOptionsRepository`. If it is set, then it should be used. Otherwise, if there is no value set on the DSL, it should look for a unique Bean of type `PublicKeyCredentialCreationOptionsRepository`. If it exists, then, it should be used.

For the other pull request, it would allow configuring `HttpMessageConverter<Object>` through the DSL only. This is important because serialization/deserialization should be explicit to avoid accidentally opening up a deserialization vulnerability.

PR #16375, opened 2025-01-08
method-security: fix invalid Kotlin syntax

simaotwx:
val/var on function parameters is invalid Kotlin syntax. It has been removed quite some time ago. This change updates the method-security page to reflect that.


rwinch:
Thanks for the Pull Request! Once the build passes it will be merged into main :smile:

Issue #16382, opened 2025-01-09
Add ClientRegistration.clientSettings.requireProofKey to Enable PKCE


rwinch:
[PKCE](https://oauth.net/2/pkce/) is recommended to prevent CSRF and authorization code injection attacks. We should further simplify enabling PKCE for Confidential Clients.

 # Current Behavior

The simplest Spring Boot application that requires PKCE is shown below:

```java
@Bean
SecurityFilterChain springSecurity(HttpSecurity http, OAuth2AuthorizationRequestResolver resolver) throws Exception {
	http
		.authorizeHttpRequests( r -> r
			.anyRequest().authenticated()
		)
		.oauth2Login(l -> l
			.authorizationEndpoint(a -> a
				.authorizationRequestResolver(resolver)
			)
		);
	return http.build();
}

@Bean
OAuth2AuthorizationRequestResolver pkceResolver(ClientRegistrationRepository clientRegistrationRepository) {
	DefaultOAuth2AuthorizationRequestResolver resolver = 
		new DefaultOAuth2AuthorizationRequestResolver(clientRegistrationRepository,"/oauth2/authorization");
	resolver.setAuthorizationRequestCustomizer(OAuth2AuthorizationRequestCustomizers.withPkce());
	return resolver;
}
```

# Shortcomings 

This will be simplified to just a Bean declaration when gh-16380 is implemented, but this still has the following downsides:

- It is not discoverable from the configuration of a typical Spring Security application
- It is not declarative

For something that is considered a best security practice (rather than an edge case), we must do better.

# Solution

I think what makes the most sense is adding a property `ClientRegistration.clientSettings.requireProofKey=(true|false)` (default is false) similar to what was [proposed](https://github.com/spring-projects/spring-security/issues/12219#issuecomment-2161545711) in gh-12219. 

**UPDATE** After speaking with @jgrandja offline, he suggested the property `ClientRegistration.clientSettings.requireProofKey=(true|false)` to align with Autorization Server's `RegisteredClient.[clientSettings](https://github.com/spring-projects/spring-authorization-server/blob/1.4.1/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/RegisteredClient.java#L73).requireProofKey`. In this case, we can introduce the `ClientSettings` in the same package as `ClientRegistration` and it does NOT need to be backed by a `Map` or extend `AbstractSettings`. If needed, we can later introduce `AbstractSettings` and back it by a `Map` at that point.

[...]

PR #16386, opened 2025-01-09
Add ClientRegistration.clientSettings.requireProofKey to Enable PKCE

kse-music:
Closes gh-16382

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/jackson2/ClientRegistrationDeserializer.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/jackson2/ClientRegistrationDeserializer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


rwinch:
Thanks for the PR @kse-music I'm putting this on pause as the team discusses the underlying ticket.

PS: Sorry for not flagging the ticket for team discussion until after you had already created the PR.

rwinch:
@kse-music Thanks again for the PR! 

After speaking with @jgrandja offline, he suggested the property `ClientRegistration.clientSettings.requireProofKey=(true|false)` to align with Autorization Server's `RegisteredClient.[clientSettings](https://github.com/spring-projects/spring-authorization-server/blob/1.4.1/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/RegisteredClient.java#L73).requireProofKey`. In this case, we can introduce the `ClientSettings` in the same package as `ClientRegistration` and it does NOT need to be backed by a `Map` or extend `AbstractSettings`. If needed, we can later introduce `AbstractSettings` and back it by a `Map` at that point.

Also, I would expect that this PR enable PKCE if that setting is set to true.

rwinch:
Thanks for the Pull Request! This is now merged into main via the merge commit 4fc99aa9e1 :smile: 

I included the original commit as 8d3e0844c5 that was rebased along with  additional polish:

* 85d7cc1335 Document requireProofKey
* 004f38639d Move ClientSettings to ClientRegistration
* 4c533569bb Ensure missing ClientRegistration.clientSettings JSON node works
* f9498d3885 PKCE cannot be true and AuthorizationGrantType != AUTHORIZATION_CODE
* ab629cc1ca Add AuthorizationGrantType.toString()
* b0a4dcb89e ClientSettings equals, hashCode, toString
* 2665a92107 Ensure that ClientSettings cannot be null
* 0ed7b18f42 DefaultServerOAuth2AuthorizationRequestResolver requireProofKey support
* 8d3e0844c5 Add ClientRegistration.clientSettings.requireProofKey to Enable PKCE

Issue #16373, opened 2025-01-08
Add support for additional dynamic OIDC Token Exchange Request parameters


kistlers:
**Expected Behavior**

In a project I am currently working on, we must send additional request parameters in the OIDC Token Exchange grant request. I want a simpler way to add and define additional parameters to be sent with the token exchange grant request.  

**Current Behavior**

The current implementation of https://github.com/spring-projects/spring-security/blob/0e3cfd1efbd0aba7e9890f695a622db2db70ddf9/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/TokenExchangeOAuth2AuthorizedClientProvider.java#L44 does not allow customization to this level. What is possible, is adding a static list of additional parameters to the grant request by passing a customized `contextAttributesMapper`  to the https://github.com/spring-projects/spring-security/blob/0e3cfd1efbd0aba7e9890f695a622db2db70ddf9/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultOAuth2AuthorizedClientManager.java#L88.

**Context**

To be more specific, when calling `OAuth2AuthorizedClientManager::authorize`, which will perform a token exchange for this client registration, the request must in my case include an additional parameter to be sent.
This is not possible without completely replacing the implementation of `TokenExchangeOAuth2AuthorizedClientProvider` for the following reasons:
- The `parametersConverter` of the `RestClientTokenExchangeTokenResponseClient` only has access to the `TokenExchangeGrantRequest` as an input that might contain per-exchange dynamic parameters.
- None of the attributes set in the `OAuth2AuthorizeRequest` passed to `OAuth2AuthorizedClientManager::authorize` end up in the `TokenExchangeGrantRequest` object.
- And even if that were the case, a second call to authorize the client with a different dynamic parameter would not result in another token exchange, as the context thinks the client is already authorized because of this snippet in `TokenExchangeOAuth2AuthorizedClientProvider`: 
https://github.com/spring-projects/spring-security/blob/0e3cfd1efbd0aba7e9890f695a622db2db70ddf9/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/TokenExchangeOAuth2AuthorizedClientProvider.java#L76-L81

I was able to circumvent these short-comings by:
- Copying `TokenExchangeOAuth2AuthorizedClientProvider` and extending the check for re-authorization.
[...]

PR #16393, opened 2025-01-10
add the ability to store additional grant request parameters for the OIDC token exchange grant request

kistlers:
This is an attempt at implementing the issue gh-#16373. Check the issue for detailed explanation of what this tries to achieve.


jgrandja:
@kistlers In case you are not aware, our process for feature or enhancement requests is to log an issue first (before PR) to understand the use case and to determine if the feature/enhancement is valid and if it's something the community would actually use.

If it's a valid use case, the issue will sit for a while until we determine if it's in demand (based on upvotes) and then we will schedule it for a release.

I see that you already logged the issue so we can continue the dialogue there. I'll close this PR.

PR #16416, opened 2025-01-14
Fix broken link

ngocnhan-tran1996:
Reference
https://docs.spring.io/spring-security/reference/servlet/test/mockmvc/setup.html
https://docs.spring.io/spring-security/reference/servlet/test/mockmvc/authentication.html


jzheaux:
Thanks for the fix, @ngocnhan-tran1996!

Issue #16422, opened 2025-01-15
Mistake in FILTER_SECTION_NAME attribute


Drimkast:
**It seems mistake in  [`FILTER_SECTION_NAME`](https://github.com/spring-projects/spring-security/blob/c78ac116f9b8d157278208e0f463cd4c2302da6b/web/src/main/java/org/springframework/security/web/ObservationFilterChainDecorator.java#L557)**
Other keys starts from `spring.security` but `FILTER_SECTION_NAME` starts from `security.security`

franticticktick:
Hey @Drimkast, you can open PR for such fixes without a ticket. There is no need to track such changes :)

Drimkast:
> Hey [@Drimkast](https://github.com/Drimkast), you can open PR for such fixes without a ticket. There is no need to track such changes :)

Thanks for info. Next time i'll do PR without ticket :)

jzheaux:
Closing in favor of https://github.com/spring-projects/spring-security/pull/16427

PR #16427, opened 2025-01-16
Use spring.security prefix instead of security.security

kse-music:
Closes gh-16422


jzheaux:
Thanks @kse-music! This is now merged into `main`.

Note that even though this is a bug, I don't think we should backport this as it may affect reports that are operating off of this key name. As such, this will only go out in 6.5. I've also added a note to [What's New](https://github.com/spring-projects/spring-security/commit/352a6a0d5312ba7789597a114d10fed2239d101b#diff-4127e68b7c2f5a9c16f4d1bdb48f607199e31f23391406e6cf575dc24c9b4dc7R11).

Issue #16432, opened 2025-01-16
Make `PublicKeyCredentialRequestOptions` `Serializable`


rwinch:
We should make `PublicKeyCredentialRequestOptions` implement `Serializable`. This will ensure that it an be persisted in distributed `HttpSession` setups.

franticticktick:
Hi @rwinch , could you review the PR?

PR #16438, opened 2025-01-17
Make PublicKeyCredentialRequestOptions Serializable

franticticktick:
Closes gh-16432


rwinch (changes requested):
Thanks for the Pull Request. Historically the serialVersionUID no longer used for new classes implementing Serializable. Going forward we should generate a serialVersionUID.

franticticktick:
Hi @rwinch, thanks for feedback. If we use the generated `serialVersionUID` `SpringSecurityCoreVersionSerializableTests` will not work. It only works with `SpringSecurityCoreVersion`. Maybe we should add such support to `SpringSecurityCoreVersionSerializableTests`?

franticticktick:
Thanks for help @rwinch. Could you review changes please?

config/src/test/resources/serialized/6.5.x/org.springframework.security.web.webauthn.api.ImmutableAuthenticationExtensionsClientInput.serialized
rwinch, line 1: It looks like this failed to write

config/src/test/java/org/springframework/security/SpringSecurityCoreVersionSerializableTests.java
```diff
@@ -508,6 +519,43 @@ class SpringSecurityCoreVersionSerializableTests {
 				(r) -> new AuthenticationSwitchUserEvent(authentication, user));
 		generatorByClassName.put(HttpSessionCreatedEvent.class,
 				(r) -> new HttpSessionCreatedEvent(new MockHttpSession()));
+
+		// webauthn
+		CredProtectAuthenticationExtensionsClientInput.CredProtect credProtect = new CredProtectAuthenticationExtensionsClientInput.CredProtect(
+				CredProtectAuthenticationExtensionsClientInput.CredProtect.ProtectionPolicy.USER_VERIFICATION_OPTIONAL,
+				true);
+		Bytes id = new Bytes(("test").getBytes());
```
rwinch, line 527: I like that you used a constant value here vs generating a new value. Please use a more realistic value. To make this easy I recently added TestBytes.get() gh-16461

config/src/test/java/org/springframework/security/SpringSecurityCoreVersionSerializableTests.java
```diff
@@ -508,6 +519,43 @@ class SpringSecurityCoreVersionSerializableTests {
...
+		generatorByClassName.put(CredProtectAuthenticationExtensionsClientInput.CredProtect.class, (c) -> credProtect);
+		generatorByClassName.put(CredProtectAuthenticationExtensionsClientInput.class,
+				(c) -> new CredProtectAuthenticationExtensionsClientInput(credProtect));
+		generatorByClassName.put(ImmutableAuthenticationExtensionsClientInputs.class, (i) -> inputs);
+		generatorByClassName.put(ImmutableAuthenticationExtensionsClientInput.class,
+				(i) -> ImmutableAuthenticationExtensionsClientInput.credProps);
+		generatorByClassName.put(Bytes.class, (b) -> id);
+		generatorByClassName.put(PublicKeyCredentialDescriptor.class, (d) -> descriptor);
+		// @formatter:off
+		generatorByClassName.put(PublicKeyCredentialRequestOptions.class, (o) -> PublicKeyCredentialRequestOptions.builder()
```
rwinch, line 549: Please use (or at least start with) TestPublicKeyCredentialRequestOptions

config/src/test/java/org/springframework/security/SpringSecurityCoreVersionSerializableTests.java
franticticktick: `ImmutableAuthenticationExtensionClientInput` has generic parameter, so `SpringSecurityCoreVersionSerializableTests` doesn't work. To serialize the class I used this trick:
```
Object instance;
if(clazz.equals(ImmutableAuthenticationExtensionsClientInput.class)) {
	instance = instancioWithParameter((Class<ImmutableAuthenticationExtensionsClientInput>) clazz)
			.create();
} else {
	instance = instancioWithDefaults(clazz).create();
}

private static InstancioApi<?> instancioWithParameter(Class<ImmutableAuthenticationExtensionsClientInput> clazz) {
	InstancioApi<?> instancio = Instancio.of(clazz).withTypeParameters(Boolean.class);
	if (generatorByClassName.containsKey(clazz)) {
		instancio.supply(Select.all(clazz), generatorByClassName.get(clazz));
	}
	return instancio;
}
```
Perhaps there is an easier way to do this. We should think about this.

franticticktick:
`SpringSecurityCoreVersionSerializableTests#serializeAndDeserializeAreEqual` does not work due to `ImmutableAuthenticationExtensionsClientInput`. As I already said, it can be serialized, but due to specifying a parameter with a generic, test `serializeAndDeserializeAreEqual` will fail. I don't see an easy way to solve this problem. @rwinch could you help please?

rwinch:
@franticticktick Thanks for the pull request. This is now merged into master along with a84173794101ab9a7536b202e8a627dbdaabeded to fix the generic types. 

NOTE: This can also be fixed using `ResolvableType.forClassWithGenerics(AuthenticationExtensionsClientInput.class, Boolean.class).resolve()`

Issue #16444, opened 2025-01-19
In the UsernamePasswordAuthenticationFilter, the authRequest is added as an attribute to the HttpServletRequest.


amm0124:
**Expected Behavior**

You expect to receive username and password in the application/json format and be able to extract them again in the unsuccessHandler.

In traditional HTML forms, the default encoding is x-www-urlencoded, but nowadays, application/json format is more commonly preferred in web development. While it is up to the developer to customize UsernamePasswordAuthenticationFilter to handle application/json format for username and password, an issue arises.

The HttpServletRequest in application/json format is a consumable stream, meaning once the data is extracted, it cannot be extracted again. This leads to a problem where, in the unsuccessHandler, you are unable to perform subsequent actions (e.g., incrementing account error counts for the username) because the stream has already been consumed.

To solve this problem, developers may find that completely customizing the filter is inefficient and means they are missing out on useful functionality provided by Spring Security.

Therefore, if authRequest is saved in the HttpServletRequest in UsernamePasswordAuthenticationFilter, it would resolve this issue.
 

**Context**

After receiving the login credentials in the application/json format for form login, you'd like to use the username and password in the unsuccessHandler for subsequent processing like error count increments.

```java

public class CustomLoginFailureHandler implements AuthenticationFailureHandler {
	
	@Override
	public void onAuthenticationFailure(HttpServletRequest request, HttpServletResponse response, AuthenticationException exception) throws IOException {
	
		UsernamePasswordAuthenticationToken authRequest = (UsernamePasswordAuthenticationToken) request.getAttribute("authRequest");
		// For example, you can extract the username using authRequest.getPrincipal() 
		// and increment the account failure count.
		
}
```

rwinch:
I think that a better way to do this would be to add a new optional property`Authentication authenticationRequest` to the `AuthenticationException`. Then in the `AuthenticationProvider` it would include the `Authentication` that was passed in to it in the `AuthenticationException` that is thrown when authentication fails. The failure would then be able to access the `authenticationRequest` property.

What do you think? Would you be interested in creating a pull request for this?

amm0124:
> I think that a better way to do this would be to add a new optional property`Authentication authenticationRequest` to the `AuthenticationException`. Then in the `AuthenticationProvider` it would include the `Authentication` that was passed in to it in the `AuthenticationException` that is thrown when authentication fails. The failure would then be able to access the `authenticationRequest` property.
> 
[...]

PR #16445, opened 2025-01-19
Add authRequest(Token) In HttpServletRequest

amm0124:
gh-16444


rwinch:
I'd prefer a solution that does not require users to know an `HttpServletRequest` attribute name. I'm closing this, but let's discuss a solution on the ticket that you logged.

NOTE: If you need this now, you can always extend UsernamePasswordAuthenticationFilter and override the setDetails method to set the attribute in the subclass.

PR #16493, opened 2025-01-28
Polish

ngocnhan-tran1996:
- Use pattern matching
- Simply if condition, `parameterToken` was checked null before, so we just check `parameterToken` is blank or not instead rechecking `null`


jzheaux:
Thanks for another polish, @ngocnhan-tran1996! This is now merged into `main`.

PR #16499, opened 2025-01-29
Add PathPatternRequestMatcher

jzheaux:
This PR adds `PathPatternRequestMatcher`, a request matcher that doesn't require `HandlerMappingIntrospector`.

Also, it adds a servlet-aware builder for specifying the servlet path (using this request matcher) in a way that looks similar to Kotlin and XML:

```java
import static org.springframework.security.web.servlet.util.matcher.PathPatternRequestMatcher.servletPath;

@Bean
SecurityFilterChain appEndpoints(HttpSecurity http) {
	http
			.requestMatchers(servletPath("/spring-mvc").matcher("/admin/**")).hasAuthority("admin")
			.requestMatchers(servletPath("/spring-mvc").matcher("/my/controller/**")).hasAuthority("controller")
			.anyRequest().authenticated()
		);

	return http.build();
}
```

Future tickets like #16509 are taking a look at a first-class way to represent the servlet inside the authorization DSL. This builder is still helpful since there are many other places in the Java DSL where request matchers are provided that may need the servlet path specified.


web/src/main/java/org/springframework/security/web/servlet/util/matcher/PathPatternRequestMatcher.java
```diff
@@ -0,0 +1,127 @@
...
+	/**
+	 * Creates a {@link PathPatternRequestMatcher} that uses the provided {@code pattern},
+	 * parsing it with {@link PathPatternParser#defaultInstance}.
+	 * <p>
+	 * The {@code pattern} should be relative to the servlet path
+	 * </p>
+	 * @param pattern the pattern used to match
+	 */
+	public static PathPatternRequestMatcher pathPattern(String pattern) {
+		PathPatternParser parser = PathPatternParser.defaultInstance;
```
rwinch, line 73: Should we allow for using the defaultInstant? How did the user verify that this is matches the one used by Spring MVC?

web/src/main/java/org/springframework/security/web/servlet/util/matcher/PathPatternRequestMatcher.java
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * A {@link RequestMatcher} that uses {@link PathPattern}s to match against each
+ * {@link HttpServletRequest}. The provided path should be relative to the servlet (that
+ * is, it should exclude any context or servlet path).
+ *
+ * <p>
+ * To also match the servlet, please see {@link RequestMatchers#servlet}
+ *
+ * <p>
+ * Note that the {@link org.springframework.web.servlet.HandlerMapping} that contains the
```
rwinch, line 40: How do we enable users to ensure that Spring MVC and Spring Security are using the same instance of PathPatternParser. This should be difficult, if not impossible, to do wrong.

web/src/main/java/org/springframework/security/web/util/matcher/RequestMatchers.java
```diff
@@ -60,7 +83,236 @@ public static RequestMatcher not(RequestMatcher matcher) {
...
+	 * The path must be of the format {@code /path}.
+	 * @return a {@link Builder} that treats URIs as relative to the given
+	 * {@code servletPath}
+	 * @since 6.5
+	 */
+	public static Builder servlet(String servletPath) {
+		Assert.notNull(servletPath, "servletPath cannot be null");
+		Assert.isTrue(servletPath.startsWith("/"), "servletPath must start with '/'");
+		Assert.isTrue(!servletPath.endsWith("/"), "servletPath must not end with a slash");
+		Assert.isTrue(!servletPath.contains("*"), "servletPath must not contain a star");
```
rwinch, line 112: I know what I should do, but to me it is not very clear to users what to do if a Servlet is mapped to `*.do`.

rwinch: What do they do if it has a mapping of `/foo` and `*.do`?

rwinch: What does a user do if the servlet maps to `/foo/*`?

web/src/main/java/org/springframework/security/web/util/matcher/RequestMatchers.java
```diff
@@ -60,7 +83,236 @@ public static RequestMatcher not(RequestMatcher matcher) {
...
+	public static Builder request() {
+		return new Builder();
+	}
+
+	/**
+	 * Create {@link RequestMatcher}s whose URIs are relative to the given
+	 * {@code servletPath}.
+	 *
+	 * <p>
+	 * The {@code servletPath} must correlate to a configured servlet in your application.
```
rwinch, line 102: This sounds like `servletPath` might be the equivalent of `<servlet-mapping>/<url-pattern>`, but I suspect it is corresponding to `HttpServletRequest.servletPath`. Perhaps, this could be clarified?

web/src/main/java/org/springframework/security/web/util/matcher/RequestMatchers.java
```diff
@@ -60,7 +83,236 @@ public static RequestMatcher not(RequestMatcher matcher) {
...
 	}
 
+	/**
+	 * Create {@link RequestMatcher}s whose URIs do not have a servlet path prefix
+	 * <p>
+	 * When there is no context path, then these URIs are effectively absolute.
+	 * @return a {@link Builder} that treats URIs as relative to the context path, if any
+	 * @since 6.5
+	 */
+	public static Builder request() {
```
rwinch, line 93: Perhaps this could be `emptyServletPath()` or similar?

web/src/main/java/org/springframework/security/web/util/matcher/RequestMatchers.java
```diff
@@ -60,7 +83,236 @@ public static RequestMatcher not(RequestMatcher matcher) {
...
+		return new Builder();
```
rwinch, line 94: I wonder if this is accurate since it matches any servletPath vs an empty servlet path?

web/src/main/java/org/springframework/security/web/servlet/util/matcher/PathPatternRequestMatcher.java
```diff
@@ -0,0 +1,127 @@
...
+ * {@link org.springframework.web.util.pattern.PathPatternParser} configured in this
+ * class.
+ * </p>
+ *
+ * @author Josh Cummings
+ * @since 6.5
+ */
+public final class PathPatternRequestMatcher implements RequestMatcher {
+
+	private final PathPattern pattern;
```
rwinch, line 51: While PathPattern does not match on the servlet path, is that meaningful for Spring Security applications? I am concerned that this makes it too easy to do the wrong thing?

docs/modules/ROOT/pages/migration-7/web.adoc
```diff
@@ -102,3 +102,12 @@ Xml::
 </b:bean>
 ----
 ======
+
+== Use Absolute Authorization URIs
```
rwinch, line 106: Can we add some examples?

rwinch (changes requested):
Thanks for all of your work on this @jzheaux! I've provided feedback below:

docs/modules/ROOT/pages/servlet/authorization/authorize-http-requests.adoc
```diff
@@ -616,34 +610,55 @@ Kotlin::
...
 
-* If you use the `spring.mvc.servlet.path` Boot property to change the default path (`/`) to something else
-* If you register more than one Spring MVC `DispatcherServlet` (thus requiring that one of them not be the default path)
+	return http.build();
+}
+----
+
+[TIP]
+=====
+There are several other components that create request matchers for you like `PathRequest#toStaticResources#atCommonLocations`
```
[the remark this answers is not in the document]
jzheaux: It is a Spring Boot feature. I'll add a link for additional context.

web/src/main/java/org/springframework/security/web/util/matcher/RequestMatchers.java
```diff
@@ -60,7 +83,236 @@ public static RequestMatcher not(RequestMatcher matcher) {
...
+		 * Create the {@link RequestMatcher}
+		 * @return the composite {@link RequestMatcher}
+		 */
+		public RequestMatcher matcher() {
+			return allOf(this.servletPath, this.methods, this.uris, this.dispatcherTypes, this.matchers);
+		}
+
+	}
+
+	private record HttpMethodRequestMatcher(HttpMethod method) implements RequestMatcher {
```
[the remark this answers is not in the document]
jzheaux: I think we could externalize it, yes. I didn't want to do it prematurely. We have public matchers for most other request aspects (path, header, dispatcher type).

web/src/main/java/org/springframework/security/web/servlet/util/matcher/PathPatternRequestMatcher.java
jzheaux: I'm not quite sure I understand the question here. I think if ppl are publishing their own `PathPatternParser`, they know that they are. As such, they should parse the URI and then call the `PathPattern` constructor.

web/src/main/java/org/springframework/security/web/servlet/util/matcher/PathPatternRequestMatcher.java
jzheaux: I suppose it depends at what level. I feel like looking up the `PathPatternParser` that MVC is using is something good for the DSL to do.

web/src/main/java/org/springframework/security/web/util/matcher/RequestMatchers.java
```diff
@@ -60,7 +83,236 @@ public static RequestMatcher not(RequestMatcher matcher) {
...
+			}
+			return new Builder(this.servletPath, this.methods, this.uris, anyOf(matchers), this.matchers);
+		}
+
+		/**
+		 * Match requests with any of these {@link RequestMatcher}s
+		 * @param requestMatchers the {@link RequestMatchers}s to match
+		 * @return the {@link Builder} for more configuration
+		 */
+		public Builder matching(RequestMatcher... requestMatchers) {
```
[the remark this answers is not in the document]
jzheaux: Possibly, I have it there because I'm imagining `requestMatchers` going away in the DSL over time. I'll try a few things and see if there is a way to remove this.

web/src/main/java/org/springframework/security/web/util/matcher/RequestMatchers.java
jzheaux: They will need to use something else. This is for the purpose of matching `HttpServletRequest#getServletPath`. I think a method name change will be helpful.

Note that I haven't yet been able to get Framework's `PathPattern` to work with extensions, so that may be a follow-up item. As it stands, this specific method doesn't comment on that.

web/src/main/java/org/springframework/security/web/util/matcher/RequestMatchers.java
jzheaux: Perhaps. Though this method is not really about declaring a request in absence of something else. I wouldn't, for example, want `AntPathRequestMatcher#antPattern(String pattern)` to be `AntPathRequestMatcher#emptyMethodPattern(String pattern)` because no method is specified.

I'm open to thinking about other names, but I don't see why we'd want to define this in terms of what is not specified. 

Furthermore, I think the 95% case is that defining a servlet path isn't necessary, so I'd prefer not to define the common case in terms of a concept that commonly is abstracted away from the developer. 

Could you elaborate on why it's important to you?

(as a side note, even though I think that specifying a servlet is not in the 95%, I like that having it as an entry point allows for the returned builder to be reused, and I also like that it makes it clear that such a matcher should only be described once)

web/src/main/java/org/springframework/security/web/servlet/util/matcher/PathPatternRequestMatcher.java
jzheaux: I'm open to other possibilities. One that I considered was to enhance PathPatternRequestMatcher to be something more of an general-purpose request matcher. In this way, its pattern of construction would indicate a place for the servlet path. Perhaps the builder in `RequestMatchers` can move to `PathPatternRequestMatcher`.

What are your thoughts on this?

docs/modules/ROOT/pages/servlet/authorization/authorize-http-requests.adoc
```diff
@@ -616,34 +610,55 @@ Kotlin::
...
-fun mvc(introspector: HandlerMappingIntrospector): MvcRequestMatcher.Builder =
-    MvcRequestMatcher.Builder(introspector).servletPath("/spring-mvc");
-
-@Bean
-fun appEndpoints(http: HttpSecurity, mvc: MvcRequestMatcher.Builder): SecurityFilterChain =
+fun appEndpoints(http: HttpSecurity): SecurityFilterChain {
     http {
         authorizeHttpRequests {
-            authorize(mvc.pattern("/my/controller/**"), hasAuthority("controller"))
+            authorize("/spring-mvc", "/admin/**", hasAuthority("admin"))
```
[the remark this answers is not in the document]
jzheaux: The documentation was mistaken. AFAIK, this is the correct way to specify the servlet path in the Kotlin DSL. See [this test](https://github.com/spring-projects/spring-security/blob/main/config/src/test/kotlin/org/springframework/security/config/annotation/web/AuthorizeHttpRequestsDslTests.kt#L537) as an example.

web/src/main/java/org/springframework/security/web/servlet/util/matcher/PathPatternRequestMatcher.java
jzheaux: I think adding `PathPatternRequestMatcher#withPathPatternParser` will clarify that this is something that may need to be considered. I'll also update the migration docs to assume they need to pick the one up from MVC.

In 7, Spring Security can find MVC's PathPatternParser instance and publish a `PathPatternRequestMatcher.Builder` that people and the DSL can use.

web/src/main/java/org/springframework/security/web/servlet/util/matcher/PathPatternRequestMatcher.java
jzheaux: See https://github.com/spring-projects/spring-security/pull/16499#discussion_r1960716372 for an addition to `PathPatternRequestMatcher` that I think may be useful.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LoginConfigurer.java
```diff
@@ -513,9 +534,9 @@ static class AntPathQueryRequestMatcher implements RequestMatcher {
 
 		private final RequestMatcher matcher;
 
-		AntPathQueryRequestMatcher(String path, String... params) {
+		AntPathQueryRequestMatcher(RequestMatcher pathMatcher, String... params) {
```
rwinch, line 537: Consider renaming this since it is no longer using `AntPathRequestMatcher`

config/src/main/java/org/springframework/security/config/web/PathPatternRequestMatcherBuilderFactoryBean.java
```diff
@@ -0,0 +1,92 @@
...
+ * used to create request matchers in {@link AuthorizeHttpRequestsConfigurer} and other
+ * parts of the DSL.
+ *
+ * @author Josh Cummings
+ * @since 6.5
+ */
+public final class PathPatternRequestMatcherBuilderFactoryBean
+		implements FactoryBean<PathPatternRequestMatcher.Builder>, ApplicationContextAware {
+
+	static final String PATTERN_PARSER_BEAN_NAME = "mvcPatternParser";
```
rwinch, line 38: Consider naming the variable `MVC_PATTERN_PARSER_BEAN_NAME` to align with the value

rwinch (approved):
Thanks for all your work on this! I've provided (optional) feedback. Feel free to merge once you are comfortable.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ott/OneTimeTokenLoginConfigurer.java
```diff
@@ -123,8 +121,9 @@ public final class OneTimeTokenLoginConfigurer<H extends HttpSecurityBuilder<H>>
 	private GenerateOneTimeTokenRequestResolver requestResolver;
 
 	public OneTimeTokenLoginConfigurer(ApplicationContext context) {
-		super(new OneTimeTokenAuthenticationFilter(), OneTimeTokenAuthenticationFilter.DEFAULT_LOGIN_PROCESSING_URL);
+		super(new OneTimeTokenAuthenticationFilter(), null);
```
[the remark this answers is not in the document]
jzheaux: yes, due to the fact that this otherwise induced the calculation of the login URL before the object had access to an `ApplicationContext`. I may be able to remove it though.

web/src/main/java/org/springframework/security/web/servlet/util/matcher/PathPatternRequestMatcher.java
```diff
@@ -0,0 +1,393 @@
...
+	}
+
+	/**
+	 * Create a {@link PathPatternRequestMatcher} whose URIs do not have a servlet path
+	 * prefix
+	 * <p>
+	 * When there is no context path, then these URIs are effectively absolute.
+	 * @return a {@link Builder} that treats URIs as relative to the context path, if any
+	 */
+	public static Builder path() {
```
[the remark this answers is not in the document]
jzheaux: I think I'll change this to `withDefaults()` since it takes on whatever defaults the builder currently has.

PR #16505, opened 2025-01-30
Include AuthenticationRequest in AuthenticationException

amm0124:
- #16444 
- Store the authentication request details in the `authRequest` field of `AuthenticationException` when an authentication exception occurs.
@rwinch

### DIFF
--- a/core/src/main/java/org/springframework/security/core/AuthenticationException.java
+++ b/core/src/main/java/org/springframework/security/core/AuthenticationException.java
@@ -18,6 +18,8 @@ package org.springframework.security.core;
 
 import java.io.Serial;
 
+import org.springframework.util.Assert;
+
 /**
  * Abstract superclass for all exceptions related to an {@link Authentication} object
  * being invalid for whatever reason.
@@ -30,6 +32,16 @@ public abstract class AuthenticationException extends RuntimeException {
 	private static final long serialVersionUID = 2018827803361503060L;
 
 	/**
+	 * The {@link Authentication} object representing the failed authentication attempt.
+	 * <p>
+	 * This field captures the authentication request that was attempted but ultimately
+	 * failed, providing critical information for diagnosing the failure and facilitating
+	 * debugging. If set, the value must not be null.
+	 * </p>
+	 */
+	private Authentication authRequest;
+
+	/**
 	 * Constructs an {@code AuthenticationException} with the specified message and root
 	 * cause.
 	 * @param msg the detail message
@@ -37,6 +49,7 @@ public abstract class AuthenticationException extends RuntimeException {
 	 */
 	public AuthenticationException(String msg, Throwable cause) {
 		super(msg, cause);
+		this.authRequest = null;
 	}
 
 	/**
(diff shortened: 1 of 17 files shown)


core/src/main/java/org/springframework/security/core/AuthenticationException.java
```diff
@@ -46,6 +49,12 @@ public AuthenticationException(String msg, Throwable cause) {
 	 */
 	public AuthenticationException(String msg) {
 		super(msg);
+		this.authRequest = null;
+	}
+
+	public AuthenticationException(String msg, Authentication authRequest) {
```
jzheaux, line 55: Since this is [an optional property](https://github.com/spring-projects/spring-security/issues/16444#issuecomment-2605931950), please add this as a setter. This also prevents a multiplicity of constructors.

jzheaux: Also please [add `@since 6.5`](https://github.com/spring-projects/spring-security/blob/main/CONTRIBUTING.adoc?plain=1#L105) new public API components

core/src/main/java/org/springframework/security/core/AuthenticationException.java
```diff
@@ -46,6 +49,12 @@ public AuthenticationException(String msg, Throwable cause) {
...
+		super(msg);
+		this.authRequest = authRequest;
```
jzheaux, line 57: When provided, please ensure that `authRequest` is not null.

jzheaux (changes requested):
Thanks for the PR, @amm0124! I've left some feedback inline.

core/src/main/java/org/springframework/security/core/AuthenticationException.java
```diff
@@ -46,6 +59,33 @@ public AuthenticationException(String msg, Throwable cause) {
...
+		this.authRequest = null;
+	}
+
+	/**
+	 * Constructs an {@code AuthenticationException} with the specified message and root
+	 * cause.
+	 * @param msg the detail message
+	 * @param authRequest details about the failed authentication request
+	 */
+	public AuthenticationException(String msg, Authentication authRequest) {
```
jzheaux, line 71: Sorry, I wasn't clear about this. Please remove this constructor in favor of the setter you added.

jzheaux (changes requested):
Thanks for the updates! I've left some additional inline feedback.

core/src/main/java/org/springframework/security/core/AuthenticationException.java
amm0124: ok! I checked it!

jzheaux (approved):

jzheaux:
Thanks for the PR, @amm0124! In addition to your contribution, I updated the Spring Security's authentication components to use the setter when throwing exceptions in https://github.com/spring-projects/spring-security/pull/16505/commits/659e7d25f49963ac4b2f6f104a8a69032def89d6. I also did a [slight polish](https://github.com/spring-projects/spring-security/pull/16505/commits/cdb485f5e99bd060612c6eaadb9501b649aa3112) of your original contribution in order to facilitate that.

Issue #16340, opened 2024-12-24
[Azure Oauth2] IllegalArgumentException: Attribute value for "xxx" is null


jloisel:
Hi,

We are having the following issue when trying to configure Azure SSO Authentication using Spring Security Oauth2:

> { json: { log: "2024-12-20 12:42:53,285 [http-nio-8090-exec-40] ERROR o.a.c.c.C.[.[.[.[dispatcherServlet] - Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception java.lang.IllegalArgumentException: Attribute value for 'preferred_username' cannot be null at org.springframework.util.Assert.notNull(Assert.java:181) at org.springframework.security.oauth2.core.user.DefaultOAuth2User.<init>(DefaultOAuth2User.java:72) at org.springframework.security.oauth2.client.userinfo.DefaultOAuth2UserService.loadUser(DefaultOAuth2UserService.java:99) at org.springframework.security.oauth2.client.oidc.userinfo.OidcUserService.loadUser(OidcUserService.java:115) at org.springframework.security.oauth2.client.oidc.userinfo.OidcUserService.loadUser(OidcUserService.java:68) at org.springframework.security.oauth2.client.oidc.authentication.OidcAuthorizationCodeAuthenticationProvider.authenticate(OidcAuthorizationCodeAuthenticationProvider.java:158) at org.springframework.security.authentication.ProviderManager.authenticate(ProviderManager.java:182) at org.springframework.security.authentication.ObservationAuthenticationManager.lambda$authenticate$1(ObservationAuthenticationManager.java:54) at io.micrometer.observation.Observation.observe(Observation.java:565) at org.springframework.security.authentication.ObservationAuthenticationManager.authenticate(ObservationAuthenticationManager.java:53) at org.springframework.security.oauth2.client.web.OAuth2LoginAuthenticationFilter.attemptAuthentication(OAuth2LoginAuthenticationFilter.java:196) at org.springframework.security.web.authentication.AbstractAuthenticationProcessingFilter.doFilter(AbstractAuthenticationProcessingFilter.java:231) at 

Seems linked to #15338.

If we use "sub" username attribute, there is no longer the issue (but sub cannot be mapped to an email on our system), but we can dump the user in an `Oauth2AuthenticationSuccessHandler`:

[...]

PR #16546, opened 2025-02-06
Bug #16340: check username attribute when calling user.getName() (instead of checking in constructor)

jloisel:
See: https://github.com/spring-projects/spring-security/issues/16340

When first loaded, the oauth2 user has not the username attribute in his claims. For this reason, checking for the presence of this attribute in the constructor is too early. Once loaded from the userinfo endpoint, the oauth2 user has the username attribute.

I moved the check from constructor to `getName()` call to prevent this issue. Junits have been rewritten accordingly.


sjohnr:
@jloisel I'm going to close this PR based on [this](https://github.com/spring-projects/spring-security/issues/16340#issuecomment-2755448462) and other comments in gh-16340.

Issue #16498, opened 2025-01-28
Method Security Templates Do Not Use Deep Non-Aliased Attributes


rwinch:
Method Security expressions that use templates do not use deep non-aliased attributes. A complete sample can be found in my [sample repository](https://github.com/rwinch/spring-sample/tree/gh-16498-deep-meta-annotations), but is highlighted below:

```java
@Service
public class Authz {
	public boolean hasPermission(Authentication authentication, Object object, String permission) {
		return true;
	}
}

@Documented
@Retention(RetentionPolicy.RUNTIME)
@Target({ElementType.TYPE, ElementType.METHOD})
@PreAuthorize("@authz.hasPermission(authentication, {object}, {permission})")
public @interface HasPermission {

	String object();

	String permission();
}

@Documented
@Retention(RetentionPolicy.RUNTIME)
@Target({ ElementType.TYPE, ElementType.METHOD})
@HasPermission(object = "{value}", permission = "'read'")
public @interface HasReadPermission {
// If the alias is used, then Spring Security does work. However, there may be templates where a new variable is introduced and thus nothing to alias
//	@AliasFor(annotation = HasPermission.class, value = "object")
	String value();
}

@Service
public class MessageService {

	@HasReadPermission("#name")
	String sayHello(String name) {
		return "Hello " + name;
	}
}

```

This will produce the error `Failed to evaluate expression '@authz.hasPermission(authentication, {value}, 'read')'`.

I'd expect the expression to replace `{value}` with `#name`.

jzheaux:
Closed in favor of https://github.com/spring-projects/spring-security/pull/16550

PR #16550, opened 2025-02-07
Method Security templates support use deep non-aliased attributes

kse-music:
Closes gh-16498


kse-music:
@jzheaux I've added the test case of @AliasFor.

jzheaux:

Issue #16545, opened 2025-02-06
Option to provide custom `logoutTokenDecoderFactory` for OIDC logout


ErwinSteffens:
**Expected Behavior**

Allow to provide custom `logoutTokenDecoderFactory` in `OidcBackChannelLogoutReactiveAuthenticationManager` so we can provide a custom WebClient and can resolve a production connection issue. 

We can work on creating PR for this, but we need some guidance for getting to a correct solution.

**Current Behavior**

We are seeing connection issues with long running HTTP connections to some of our OIDC providers. This shows as the following errors message in our logging: 

```
i.n.c.u.Errors$NativeIoException: recv(..) failed: Connection reset by peer
Wrapped by: o.s.w.r.f.c.WebClientRequestException: recv(..) failed: Connection reset by peer
	at o.s.w.r.f.c.ExchangeFunctions$DefaultExchangeFunction.lambda$wrapException$9(ExchangeFunctions.java:137)
	Suppressed: r.c.p.FluxOnAssembly$OnAssemblyException: 
Error has been observed at the following site(s):
	*__checkpoint ⇢ Request to GET https://xxxx/.well-known/jwks.json [DefaultWebClient]
```

Similar issue in the login process is described [here](https://github.com/spring-projects/spring-security/issues/13274). There was a lot of discussion on how to resolve this, but there is now an option to create a custom `ReactiveJwtDecoderFactory` to validate ID token with a custom `WebClient` instance: 

```java
@Component
public class CustomReactiveOidcIdTokenDecoderFactory implements ReactiveJwtDecoderFactory<ClientRegistration> {

	private final WebClient webClient;

	public CustomReactiveOidcIdTokenDecoderFactory(WebClient webClient) {
		this.webClient = webClient;
	}

	public ReactiveJwtDecoder createDecoder(ClientRegistration registration) {
		var issuerUri = registration.getProviderDetails().getIssuerUri();
		var decoder = NimbusReactiveJwtDecoder.withIssuerLocation(issuerUri)
				.webClient(this.webClient)
				.build();
		decoder.setJwtValidator(JwtValidators.createDefaultWithValidators(new OidcIdTokenValidator(registration)));
		decoder.setClaimSetConverter(new ClaimTypeConverter(ReactiveOidcIdTokenDecoderFactory.createDefaultClaimTypeConverters()));
		return decoder;
	}
}
```

We would like to build similar functionality for the OIDC backchannel logout process but this cannot be accomplished because:
[...]

PR #16552, opened 2025-02-07
Allow to set an authentication manager for back-channel logout in ServerHttpSecurity

ErwinSteffens:
Allow to set custom OidcBackChannelLogoutReactiveAuthenticationManager instance so you can override the logout token decoder factory with for setting a custom WebClient instance. 

Issue: https://github.com/spring-projects/spring-security/issues/16545

Usage:

```java
http
		.oauth2Login(Customizer.withDefaults())
		.oidcLogout(customizer -> customizer
				.backChannel(config -> {
					var manager = new OidcBackChannelLogoutReactiveAuthenticationManager();
					manager.setLogoutTokenDecoderFactory(logoutTokenDecoderFactory);
					config.authenticationManager(manager);
				}));

```

### DIFF
--- a/config/src/main/java/org/springframework/security/config/web/server/OidcBackChannelLogoutReactiveAuthenticationManager.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/OidcBackChannelLogoutReactiveAuthenticationManager.java
@@ -64,14 +64,14 @@ import org.springframework.util.StringUtils;
  * "https://openid.net/specs/openid-connect-backchannel-1_0.html">OIDC Back-Channel
  * Logout</a>
  */
-final class OidcBackChannelLogoutReactiveAuthenticationManager implements ReactiveAuthenticationManager {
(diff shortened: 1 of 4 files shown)


c1rd3cm:
Hi @ErwinSteffens, is there a specific reason this request was closed? We also need to customize the back-channel token validation, and this limitation is currently blocking our work. I noticed that another user opened a similar issue in #18295, which suggests others are running into the same problem. Could you provide any clarification on the status or potential next steps?

Fix3dll:
I am the author of #18295. I would be very happy to hear if you have a solution.

ErwinSteffens:
> Hi @ErwinSteffens, is there a specific reason this request was closed? We also need to customize the back-channel token validation, and this limitation is currently blocking our work. I noticed that another user opened a similar issue in #18295, which suggests others are running into the same problem. Could you provide any clarification on the status or potential next steps?

@c1rd3cm Sorry, late reply. No not sure anymore, but I actually also need it now.

Issue #16540, opened 2025-02-05
mapToUser/mapToGrantedAuthority in JdbcUserDetailsManager to become protected


flozano:
**Expected Behavior**

I'd like to be able to override `mapToUser` and `mapToGrantedAuthority` in order to be able to customize how the user is mapped and what's the actual class to be used, and what's the instance of GrantedAuthority to be used.

**Current Behavior**

```java
	private UserDetails mapToUser(ResultSet rs, int rowNum) throws SQLException {
```

**Context**
JdbcUserDetailsManager is not final and can be extended to override some of its behavior. It already has protected methods. By making this one protected, the extensibility can be much simplified. 

The `loadUsersByUsername` is protected and can be extended already, so this is a minimal change.

jzheaux:
Closing in favor of #16561

PR #16561, opened 2025-02-10
Add customizable RowMappers for user details and authorities in JdbcUserDetailsManager

big-cir:
# Overview
- added setUserDetailsMapper method to allow customization of UserDetails mapping using RowMapper
- added setGrantedAuthorityMapper method to allow customization of GrantedAuthority mapping using RowMapper
- unit tests for updated methods

# Related
- Closes gh-16540

### DIFF
--- a/core/src/main/java/org/springframework/security/provisioning/JdbcUserDetailsManager.java
+++ b/core/src/main/java/org/springframework/security/provisioning/JdbcUserDetailsManager.java
@@ -181,7 +181,7 @@ public class JdbcUserDetailsManager extends JdbcDaoImpl implements UserDetailsMa
 		return getJdbcTemplate().query(getUsersByUsernameQuery(), this::mapToUser, username);
 	}
 
-	private UserDetails mapToUser(ResultSet rs, int rowNum) throws SQLException {
+	protected UserDetails mapToUser(ResultSet rs, int rowNum) throws SQLException {
 		String userName = rs.getString(1);
 		String password = rs.getString(2);
 		boolean enabled = rs.getBoolean(3);
@@ -390,7 +390,7 @@ public class JdbcUserDetailsManager extends JdbcDaoImpl implements UserDetailsMa
 				this::mapToGrantedAuthority);
 	}
 
-	private GrantedAuthority mapToGrantedAuthority(ResultSet rs, int rowNum) throws SQLException {
+	protected GrantedAuthority mapToGrantedAuthority(ResultSet rs, int rowNum) throws SQLException {
 		String roleName = getRolePrefix() + rs.getString(3);
 		return new SimpleGrantedAuthority(roleName);
 	}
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/springframework/security/provisioning/JdbcUserDetailsManager.java
```diff
@@ -181,7 +181,7 @@ protected List<UserDetails> loadUsersByUsername(String username) {
 		return getJdbcTemplate().query(getUsersByUsernameQuery(), this::mapToUser, username);
 	}
 
-	private UserDetails mapToUser(ResultSet rs, int rowNum) throws SQLException {
+	protected UserDetails mapToUser(ResultSet rs, int rowNum) throws SQLException {
```
jzheaux, line 184: Instead of opening these methods up, could we first try using `RowMapper<UserDetails>`? I think it could work like this:

1. add a new member called `RowMapper<UserDetails> userDetailsMapper = this::mapToUser`
2. use this mapper in the `JdbcTemplate` invocation
3. add `setUserDetailsMapper(RowMapper)` to the class

This reflects what [newer Jdbc-based Spring Security components](https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/JdbcOAuth2AuthorizedClientService.java#L238) have been doing.

jzheaux (changes requested):
Thanks for the PR, @big-cir! Can we try composition first? I've left some feedback inline for details on how that could be done.

core/src/main/java/org/springframework/security/provisioning/JdbcUserDetailsManager.java
big-cir: @jzheaux 
Thank you for your feedback! I've implemented the changes as per your suggestion. Could you please confirm if I've correctly understood and applied your recommendations?

core/src/main/java/org/springframework/security/provisioning/JdbcUserDetailsManager.java
```diff
@@ -178,7 +186,7 @@ protected void initDao() throws ApplicationContextException {
 	 */
 	@Override
 	protected List<UserDetails> loadUsersByUsername(String username) {
-		return getJdbcTemplate().query(getUsersByUsernameQuery(), this::mapToUser, username);
+		return getJdbcTemplate().query(getUsersByUsernameQuery(), userDetailsMapper, username);
```
jzheaux, line 189: Please also add two unit tests:

1. One that confirms that setting a `null` value for `setUserDetailsMapper` will throw an exception
2. One that confirms that setting a mock value gets used when querying (call `setUserDetailMapper` with a Mockito mock, search for a user, and verify that the mock row mapper was used)

jzheaux (changes requested):
Hi, @big-cir, yes, that's what I'm thinking. Please see some additional feedback about the method and then let's also proceed with doing the same for `mapToGrantedAuthority`.

big-cir:
Dear @jzheaux,
Thank you very much for your detailed feedback on the setUserDetailsMapper method in JdbcUserDetailsManager.
I've implemented all the changes you suggested. Would you kindly review the updates when you have a moment? I'm eager to ensure the modifications meet the project's standards and your expectations.

Thank you again for your time and expertise.

jzheaux (approved):

PR #16570, opened 2025-02-11
Support JWK Selection Strategy in NimbusJwtEncoder

douxiaofeng99:
Signed-off-by: douxiaofeng99 <18600127780@163.com>

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
@@ -87,17 +87,12 @@ public final class NimbusJwtEncoder implements JwtEncoder {
 
 	private final JWKSource<SecurityContext> jwkSource;
 
-	private Converter<List<JWK>, JWK> jwkSelector= (jwks)->{
-		if (jwks.size() > 1) {
-			throw new JwtEncodingException(String.format(
-					"Failed to select a key since there are multiple for the signing algorithm [%s]; " +
-							"please specify a selector in NimbusJwsEncoder#setJwkSelector",jwks.get(0).getAlgorithm()));
-		}
-		if (jwks.isEmpty()) {
-			throw new JwtEncodingException(
-					String.format(ENCODING_ERROR_MESSAGE_TEMPLATE, "Failed to select a JWK signing key"));
-		}
-		return jwks.get(0);
+	private Converter<List<JWK>, JWK> jwkSelector = (jwks) -> {
+		throw new JwtEncodingException(
+				String.format(
+						"Failed to select a key since there are multiple for the signing algorithm [%s]; "
+								+ "please specify a selector in NimbusJwsEncoder#setJwkSelector",
+						jwks.get(0).getAlgorithm()));
 	};
 
 	/**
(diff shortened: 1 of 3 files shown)


oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -86,6 +87,8 @@ public final class NimbusJwtEncoder implements JwtEncoder {
 
 	private final JWKSource<SecurityContext> jwkSource;
 
+	private Converter<List<JWK>, JWK> jwkSelector;
```
jzheaux, line 90: Will you please give this a default that throws an exception? Like this:

```java
(jwks) -> {
    if (jwks.size() > 1) {
        throw new JwtEncodingException(...)
    }
    return jwks.get(0);
};
```

And have the error message be what [this comment](https://github.com/spring-projects/spring-security/issues/16170#issuecomment-2529603729) indicates.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -95,6 +98,10 @@ public NimbusJwtEncoder(JWKSource<SecurityContext> jwkSource) {
 		this.jwkSource = jwkSource;
 	}
 
+	public void setJwkSelector(Converter<List<JWK>, JWK> jwkSelector) {
```
jzheaux, line 101: Will you please add a JavaDoc description of what this method does. For example:

```java
/**
  * Use this strategy to reduce the list of matching JWKs down to a since one.
  * <p> For example, you can call {@code setJwkSelector(List::getFirst)} in order
  * to have this encoder select the first match.
  * 
  * <p> By default, the class with throw an exception if there is more than one result.
  * @since 6.5
  */
```

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -124,6 +131,10 @@ private JWK selectJwk(JwsHeader headers) {
 					"Failed to select a JWK signing key -> " + ex.getMessage()), ex);
 		}
 
+		if (null != this.jwkSelector) {
```
jzheaux, line 134: Instead of checking for null, set `this.jwkSelector` to have a default and then do:

```java
return this.jwkSelector.convert(jwks);
```

You can then remove the rest of this method.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -46,6 +46,7 @@
 import com.nimbusds.jwt.JWTClaimsSet;
 import com.nimbusds.jwt.SignedJWT;
 
+import org.springframework.core.convert.converter.Converter;
```
jzheaux, line 49: Please update the copyright to end at `2025`, please

jzheaux (changes requested):
Thanks, @douxiaofeng99! I've left some feedback inline.

Also, will you please format your commit message with a title and a closes message like so:

```java
Support JWK Selection Strategy

Closes gh-16170
```

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
douxiaofeng99: Thank you for your feedback. Since this PR has already been closed, am I still able to make modifications? How should I submit the changes? I really appreciate the review.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
douxiaofeng99: Thank you for your feedback. Since this PR has already been closed, am I still able to make modifications? How should I submit the changes? I really appreciate the review.

douxiaofeng99:
@jzheaux， Thanks you very much!

jzheaux (approved):

PR #16587, opened 2025-02-12
Consistently NonNull annotation

ngocnhan-tran1996:


jzheaux:
Thanks for the polish, @ngocnhan-tran1996! This is now merged into `main`.

Issue #15509, opened 2024-08-01
`OidcClientInitiated(Server)LogoutSuccessHandler` might build a redirection URI to the authorization server `end_session_endpoint` with an expired ID token


ch4mpy:
**Describe the bug**
In case of an RP-Initiated Logout triggered after a rather long stale period, it is possible that the `OidcClientInitiated(Server)LogoutSuccessHandler` builds a redirection URI to the authorization server `end_session_endpoint` with an expired ID token, in which case the second part of the logout fails and the user session on the OpenID Provider might not be ended.

This happens because `OidcClientInitiated(Server)LogoutSuccessHandler` internally uses the `(Reactive)ClientRegistrationRepository` instead which does not refresh expired tokens. If it was using the `(Reactive)OAuth2AuthorizedClientManager`, this would probably not happen (at least not until the refresh token is expired too).

**To Reproduce**
This can be reproduced using SPAs:
- Login to an `oauth2Client` configured with RP-Initated Logout
- wait until the ID token is expired without interacting with the OAuth2 client
- send a POST request to the `/logout` endpoint

**Expected behavior**
The redirection to the authorization server `end_session_endpoint` should be built with a valid ID token (at least if the session contains a valid refresh token).

ch4mpy:
@sjohnr I think I progressed in my understanding of the root cause for this issue and re-wrote the description accordingly.

mlitcher:
Just wanted to confirm, Spring Authorization Server does refresh the `id_token` during the refresh token flow and only seems to accept the last one during an RP-initiated logout by default. So, using Spring Cloud Gateway w/ oauth2 client + Spring Authorization Server, you will get an error during RP-initiated logout after the first token refresh due to this issue.

yhao3:
I encountered the same issue when using the Spring OAuth2 client with the Spring Authorization Server.

[...]

PR #16589, opened 2025-02-13
Ensure ID Token is updated after refresh token

yhao3:
Related issue: gh-15509


yhao3:
Hi @sjohnr,

Could you please review this PR? I encountered this issue during development and took the opportunity to resolve it.

Thanks in advance for your time and feedback!

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/RefreshOidcIdTokenHandler.java
```diff
@@ -0,0 +1,61 @@
...
+		this.oidcAuthorizationCodeAuthenticationProvider = oidcAuthorizationCodeAuthenticationProvider;
+	}
+
+	@Override
+	public void onApplicationEvent(OAuth2TokenRefreshedEvent event) {
+		OAuth2AuthorizedClient authorizedClient = event.getAuthorizedClient();
+		OAuth2AccessTokenResponse accessTokenResponse = event.getAccessTokenResponse();
+		OidcIdToken refreshedOidcToken = this.oidcAuthorizationCodeAuthenticationProvider
+			.createOidcToken(authorizedClient.getClientRegistration(), accessTokenResponse);
+		Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
```
sjohnr, line 49: Can you please adjust this to use a `private SecurityContextHolderStrategy securityContextHolderStrategy` field instead? It can be initialized with `SecurityContextHolder.getContextHolderStrategy()` and also include a setter for a custom strategy.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/RefreshOidcIdTokenHandler.java
```diff
@@ -0,0 +1,61 @@
...
+		if (authentication instanceof OAuth2AuthenticationToken oauth2AuthenticationToken) {
+			if (authentication.getPrincipal() instanceof DefaultOidcUser defaultOidcUser) {
+				OidcUser oidcUser = new DefaultOidcUser(defaultOidcUser.getAuthorities(), refreshedOidcToken,
+						defaultOidcUser.getUserInfo(), StandardClaimNames.SUB);
+				SecurityContextHolder.getContext()
+					.setAuthentication(new OAuth2AuthenticationToken(oidcUser, oidcUser.getAuthorities(),
+							oauth2AuthenticationToken.getAuthorizedClientRegistrationId()));
```
sjohnr, line 56: Can you please adjust this to create a new empty `SecurityContext` via `this.securityContextHolderStrategy.createEmptyContext()` and then set the new context via `this.securityContextHolderStrategy.setContext()`?

sjohnr (changes requested):
Thanks for the PR @yhao3. I have performed an initial review and noted a few issues below for you to consider.

yhao3:
Hi @sjohnr,

Thanks for your review and suggestions! I’ve updated the implementation based on your feedback.

sjohnr:
Thanks @yhao3! I'll take another look at this hopefully early next week.

yhao3:
Hi @sjohnr,

Thanks for your response! I’ve made some additional adjustments to `RefreshOidcIdTokenHandler` and added tests. Looking forward to your feedback when you have a chance to review.

sjohnr:
Hi @yhao3, thanks for your updates. Sorry for the delay, it took longer than I expected to work through the updates I felt were needed to get closer to merging this. I will post an update on the original issue for folks to give feedback.

yhao3:
Hi @sjohnr,

Thanks for the update! No worries at all, I really appreciate the time and effort you’re putting into this. I’ll keep an eye on the original issue for further feedback.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcAuthorizedClientRefreshedEventListener.java
```diff
@@ -0,0 +1,219 @@
...
+		}
+
+		// The registrationId must match the one used to log in
+		ClientRegistration clientRegistration = event.getAuthorizedClient().getClientRegistration();
+		if (!authenticationToken.getAuthorizedClientRegistrationId().equals(clientRegistration.getRegistrationId())) {
+			return;
+		}
+
+		// Refresh the OidcUser and send a user refreshed event
+		OidcIdToken idToken = createOidcToken(clientRegistration, accessTokenResponse);
```
filiphr, line 121: Shouldn't we do this only if the current authentication matches the principal in the authorized client in the event? Not sure if it can happen that a refresh happens the client registration is the same, but the current authentication does not match the authorized client principal

filiphr (commented):
This looks like a good solution to solve #15509. I've only added one comment, not sure if it is relevant or not in this scenario

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcAuthorizedClientRefreshedEventListener.java
```diff
@@ -0,0 +1,219 @@
...
+ * {@link OAuth2AuthorizedClientRefreshedEvent} and publishes an event of type
+ * {@link OidcUserRefreshedEvent} in order to refresh an {@link OidcUser}.
+ *
+ * @author Steve Riesenberg
+ * @since 6.5
+ * @see org.springframework.security.oauth2.client.RefreshTokenOAuth2AuthorizedClientProvider
+ * @see OAuth2AuthorizedClientRefreshedEvent
+ * @see OidcUserRefreshedEvent
+ */
+public final class OidcAuthorizedClientRefreshedEventListener
```
ch4mpy, line 63: As it listens for `OAuth2AuthorizedClientRefreshedEvent` shouldn't this listener be an `OAuth2AuthorizedClientRefreshedEventListener`?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcAuthorizedClientRefreshedEventListener.java
```diff
@@ -0,0 +1,219 @@
...
+import org.springframework.security.oauth2.jwt.JwtDecoder;
+import org.springframework.security.oauth2.jwt.JwtDecoderFactory;
+import org.springframework.security.oauth2.jwt.JwtException;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
+ * An {@link ApplicationListener} that listens for events of type
+ * {@link OAuth2AuthorizedClientRefreshedEvent} and publishes an event of type
+ * {@link OidcUserRefreshedEvent} in order to refresh an {@link OidcUser}.
```
ch4mpy, line 55: Shouldn't the listener publish `OAuth2UserRefreshedEvent` if the principal is an `OAuth2User` but not an `OidcUser`?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcAuthorizedClientRefreshedEventListener.java
```diff
@@ -0,0 +1,219 @@
...
+		if (!(authentication instanceof OAuth2AuthenticationToken authenticationToken)
+				|| authenticationToken.getClass() != OAuth2AuthenticationToken.class) {
+			// This event listener only handles the default authentication result. If the
+			// application customizes the authentication result, then a custom event
+			// handler should be provided.
+			return;
+		}
+
+		// The current principal must be an OidcUser
+		if (!(authenticationToken.getPrincipal() instanceof OidcUser existingOidcUser)) {
```
ch4mpy, line 110: Testing for `OAuth2User` would allow to process OAuth2 refresh token flows without OIDC

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcAuthorizedClientRefreshedEventListener.java
```diff
@@ -0,0 +1,219 @@
...
+ * @see OidcUserRefreshedEvent
+ */
+public final class OidcAuthorizedClientRefreshedEventListener
+		implements ApplicationEventPublisherAware, ApplicationListener<OAuth2AuthorizedClientRefreshedEvent> {
+
+	private static final String INVALID_ID_TOKEN_ERROR_CODE = "invalid_id_token";
+
+	private static final String INVALID_NONCE_ERROR_CODE = "invalid_nonce";
+
+	private OAuth2UserService<OidcUserRequest, OidcUser> userService = new OidcUserService();
```
ch4mpy, line 70: Typing with `OidcUserRequest` and `OidcUser` might be too narrow: the `OAuth2User` should be refreshed even if the provider is not configured with OIDC (calling the userinfo endpoint instead of parsing the ID token again)

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/oidc/authentication/OidcAuthorizedClientRefreshedEventListenerTests.java
```diff
@@ -0,0 +1,421 @@
...
+	public void onApplicationEventWhenAccessTokenResponseMissingOpenidScopeThenOidcUserRefreshedEventNotPublished() {
+		OAuth2AccessTokenResponse accessTokenResponse = TestOAuth2AccessTokenResponses.accessTokenResponse()
+			.scopes(Set.of(OidcScopes.OPENID))
+			.build();
+		OAuth2AuthorizedClientRefreshedEvent authorizedClientRefreshedEvent = new OAuth2AuthorizedClientRefreshedEvent(
+				accessTokenResponse, this.authorizedClient);
+		this.eventListener.onApplicationEvent(authorizedClientRefreshedEvent);
+		verifyNoInteractions(this.securityContextHolderStrategy, this.jwtDecoder, this.userService,
+				this.applicationEventPublisher);
+	}
```
yhao3, line 184: These two tests seem to be duplicated. Maybe we should keep only the first one?

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/oidc/authentication/OidcAuthorizedClientRefreshedEventListenerTests.java
```diff
@@ -0,0 +1,421 @@
...
+
+	@Test
+	public void onApplicationEventWhenAccessTokenResponseMissingOpenidScopeThenOidcUserRefreshedEventNotPublished() {
+		OAuth2AccessTokenResponse accessTokenResponse = TestOAuth2AccessTokenResponses.accessTokenResponse()
+			.scopes(Set.of(OidcScopes.OPENID))
```
yhao3, line 177: The test method `onApplicationEventWhenAccessTokenResponseMissingOpenidScopeThenOidcUserRefreshedEventNotPublished` needs to be fixed as it currently does not match its intended purpose.


```suggestion
```

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcAuthorizedClientRefreshedEventListener.java
sjohnr: Thanks for the feedback @filiphr.

Actually, at the point where this event is being processed, the principal name in the authorized client will always be the same as the current authentication. The reason for this is that the authorized client is typically created from the current principal using the same `SecurityContextHolderStrategy` to resolve it. This would only not be the case if the application is customized to somehow populate the principal differently, which I think would be quite uncommon. In any case, if we added a check for it here, it would always pass and so I'm not sure it makes much sense.

filiphr: You are right @sjohnr. It indeed does not make sense that when using the same registration as the one that the login one the principal in the authorized client is different as the one from the authentication.

I think that I missed the prior check and got side tracked with my thoughts.

sjohnr (approved):

toob2:
is this fix included in 1.4.3?

jgrandja:
@toob2 Yes, this is in `6.5.0`

toob2:
> @toob2 Yes, this is in `6.5.0`

this is however still broken, can this ticket be re-opened?

VithouJavaMaestro:
I'm still having the same issue as @toob2

jgrandja:
@VithouJavaMaestro 

> Does this work with RefreshTokenReactiveOAuth2AuthorizedClientProvider?

No, it hasn't been implemented yet. I've added gh-17188 to track it.

toob2:
> @toob2
> 
> > this is however still broken, can this ticket be re-opened?
> 
> I've confirmed that this fix in `6.5.0` does work.
> 
> In order for this solution to work, the `RefreshTokenOAuth2AuthorizedClientProvider` **requires** a `setApplicationEventPublisher(ApplicationEventPublisher)`. Can you double check to ensure an `ApplicationEventPublisher` is set?
> 
> If you are using the default `OAuth2AuthorizedClientManager` `@Bean` then the `ApplicationEventPublisher` will automatically be set. However, if you are supplying your own `OAuth2AuthorizedClientManager` `@Bean`, then this might be the issue because **you need to explicitly configure** `RefreshTokenOAuth2AuthorizedClientProvider.setApplicationEventPublisher(ApplicationEventPublisher)`.

@jgrandja sorry I didnt mention I'm using reactive SCG, will wait for the fix, thanks

zeapo:
@jgrandja thanks for the great work.
There is however a side case where this does not work:
in the RefreshTokenOAuth2AUthorizedClientProvider, we check "hasExpired" on the access token only, however it happens (in our case with azure) that the id token expires before:

Screenshot 2025-07-04 at 13 42 36
should I open an issue regarding that?

jgrandja:
@zeapo 

> in the RefreshTokenOAuth2AUthorizedClientProvider, we check "hasExpired" on the access token only

This is correct. It only needs to check the access token and if it's expired then it renews it via the `refresh_token` grant. The `RefreshTokenOAuth2AuthorizedClientProvider` is not responsible for checking the ID Token.

Issue #16385, opened 2025-01-09
PublicKeyCredentialUserEntityRepository saves `anonymousUser`


justincranford:
Asked in Stack Overflow a week ago.

https://stackoverflow.com/questions/79322876/why-does-spring-security-webauthn-authentication-save-anonymoususer-in-publick

# Minimum Viable Example with Steps to Reproduce

I included a [Minimum Reproducible Example](https://stackoverflow.com/questions/79322876/why-does-spring-security-webauthn-authentication-save-anonymoususer-in-publick#:~:text=a%20new%20account.-,Minimum%20Reproducible%20Example,-https%3A//github.com) with [Steps](https://stackoverflow.com/questions/79322876/why-does-spring-security-webauthn-authentication-save-anonymoususer-in-publick#:~:text=java%20for%20logging.-,Steps,-%3A) in the Stack Overflow post.

App in GitHub: [https://github.com/justincranford/spring-security-webauthn-demo](https://github.com/justincranford/spring-security-webauthn-demo/tree/main/src/main/java/com/justincranford/springsecurity/webauthn/demo)

[Spring versions used](https://github.com/justincranford/spring-security-webauthn-demo/blob/f3307474d82e87aeabca5f57c14fce54879fb877/pom.xml#L9):
- Spring Boot 3.4.1
- Spring Security 6.4.1

# Expected behavior

Expected behavior is `anonymousUser` should not be persisted in [PublicKeyCredentialUserEntityRepository.java](https://github.com/spring-projects/spring-security/blob/main/web/src/main/java/org/springframework/security/web/webauthn/management/PublicKeyCredentialUserEntityRepository.java).

Said another way, expectation is WebAuthn functionality should only ever persist UserEntity and Credential, for authenticated users.

However, I see `anonymousUser` is persisted during WebAuthn Authentication. That seems like a bug.

Or, if there is a valid reason for persisting `anonymousUser`, I would like to understand the design intent, so I can handle it securely and correctly.

# Logs

[...]

PR #16606, opened 2025-02-17
Fix bug PublicKeyCredentialUserEntityRepository saves anonymousUser

Borghii:
Issue gh-16385

### DIFF
--- a/web/src/main/java/org/springframework/security/web/webauthn/management/Webauthn4JRelyingPartyOperations.java
+++ b/web/src/main/java/org/springframework/security/web/webauthn/management/Webauthn4JRelyingPartyOperations.java
@@ -46,6 +46,7 @@ import com.webauthn4j.data.client.challenge.DefaultChallenge;
 import com.webauthn4j.data.extension.authenticator.RegistrationExtensionAuthenticatorOutput;
 import com.webauthn4j.server.ServerProperty;
 
+import org.springframework.security.authentication.AnonymousAuthenticationToken;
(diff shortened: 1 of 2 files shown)


ziqin:
👍 I have tested this PR. It works with my own JDBC-based `PublicKeyCredentialUserEntityRepository` and `UserCredentialRepository` implementation. No anonymousUser entry is created.

web/src/main/java/org/springframework/security/web/webauthn/management/Webauthn4JRelyingPartyOperations.java
```diff
@@ -346,6 +345,18 @@ public PublicKeyCredentialRequestOptions createCredentialRequestOptions(
 			.build();
 	}
 
+	private List<CredentialRecord> findCredentialRecords(Authentication authentication) {
+		AuthenticationTrustResolver trustResolver = new AuthenticationTrustResolverImpl();
```
ziqin, line 349: We can reuse `this.trustResolver` here.

ziqin (changes requested):
We don't need to create a `trustResolver` instance every time `findCredentialRecords` is called.

web/src/main/java/org/springframework/security/web/webauthn/management/Webauthn4JRelyingPartyOperations.java
Borghii: Thanks. I didn't realize

Issue #16590, opened 2025-02-14
Add PermitAll and DenyAll AuthorizationManager instances


jzheaux:
It would be nice to have an `AuthorizationManager` that returned a static response:

```java
public final class SingleResultAuthorizationManager<C> implements AuthorizationManager<C> {
    private final AuthorizationResult result;

    // ...

    public static <C> SingleResultAuthorizationManager<C> permitAll() { ... }

    public static <C> SingleResultAuthorizationManager<C> denyAll() { ... }
}
```

This ticket would also look for and replace code that creates a permit-all or deny-all lambda like in `AuthorizeHttpRequestsConfigurer#AuthorizeUrl` and `RequestMatcherDelegatingAuthorizationManager#AuthorizeUrl`.

franticticktick:
Hi @jzheaux , could you assign this ticket to me please?

jzheaux:
Yes, @franticticktick. Will you in your review please also ping @plll0123 as they were recently involved in a similar polish? I'd like to see if I can encourage them to make more contributions.

PR #16612, opened 2025-02-17
Add SingleResultAuthorizationManager

franticticktick:
Closes gh-16590

### DIFF
--- a/core/src/main/java/org/springframework/security/authorization/SingleResultAuthorizationManager.java
+++ b/core/src/main/java/org/springframework/security/authorization/SingleResultAuthorizationManager.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2002-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.authorization;
+
+import java.util.function.Supplier;
+
+import org.springframework.security.core.Authentication;
+import org.springframework.util.Assert;
+
+/**
+ * An {@link AuthorizationManager} which creates permit-all and deny-all
+ * {@link AuthorizationManager} instances.
+ *
+ * @author Max Batischev
+ * @since 6.5
+ */
+public final class SingleResultAuthorizationManager<C> implements AuthorizationManager<C> {
+
+	private static final SingleResultAuthorizationManager<?> DENY_MANAGER = new SingleResultAuthorizationManager<>(
+			new AuthorizationDecision(false));
+
+	private static final SingleResultAuthorizationManager<?> PERMIT_MANAGER = new SingleResultAuthorizationManager<>(
+			new AuthorizationDecision(true));
+
+	private final AuthorizationResult result;
+
+	public SingleResultAuthorizationManager(AuthorizationResult result) {
+		Assert.notNull(result, "result cannot be null");
+		this.result = result;
+	}
+
+	@Override
+	public AuthorizationDecision check(Supplier<Authentication> authentication, C object) {
+		if (!(this.result instanceof AuthorizationDecision)) {
+			throw new IllegalArgumentException("result should be AuthorizationDecision");
+		}
+		return (AuthorizationDecision) this.result;
+	}
+
+	@Override
+	public AuthorizationResult authorize(Supplier<Authentication> authentication, C object) {
+		return this.result;
+	}
+
+	@SuppressWarnings("unchecked")
+	public static <C> SingleResultAuthorizationManager<C> denyAll() {
+		return (SingleResultAuthorizationManager<C>) DENY_MANAGER;
+	}
+
+	@SuppressWarnings("unchecked")
+	public static <C> SingleResultAuthorizationManager<C> permitAll() {
+		return (SingleResultAuthorizationManager<C>) PERMIT_MANAGER;
+	}
+
+}
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/AuthorizeHttpRequestsConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/AuthorizeHttpRequestsConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2024 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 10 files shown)


franticticktick:
Hey @plll0123, we invite you to join the review so that your interest in contributions does not disappear :)

core/src/main/java/org/springframework/security/authorization/SingleResultAuthorizationManager.java
```diff
@@ -0,0 +1,62 @@
...
+	private static final AuthorizationDecision DENY = new AuthorizationDecision(false);
+
+	private static final AuthorizationDecision PERMIT = new AuthorizationDecision(true);
+
+	/**
+	 * Creates permit-all {@link AuthorizationManager} instance.
+	 * @param <C>
+	 * @return permit-all {@link AuthorizationManager} instance
+	 */
+	public static <C> AuthorizationManager<C> PERMIT_ALL() {
```
jzheaux, line 41: Spring Security conventionally uses `camelCase` for methods, even ones that return a constant; please change this to `permitAll()`.

core/src/main/java/org/springframework/security/authorization/SingleResultAuthorizationManager.java
```diff
@@ -0,0 +1,62 @@
...
+		return (a, o) -> PERMIT;
+	}
+
+	/**
+	 * Creates deny-all {@link AuthorizationManager} instance.
+	 * @param <C>
+	 * @return deny-all {@link AuthorizationManager} instance
+	 */
+	public static <C> AuthorizationManager<C> DENY_ALL() {
```
jzheaux, line 50: Spring Security conventionally uses `camelCase` for methods, even ones that return a constant; please change this to `denyAll()`.

jzheaux, line 50: I think this should return the concrete type. Just like constructors emit the concrete type. Can you please change both method to return `SingleResultAuthorizationManager`?

core/src/main/java/org/springframework/security/authorization/SingleResultAuthorizationManager.java
```diff
@@ -0,0 +1,62 @@
...
+		return (a, o) -> DENY;
+	}
+
+	private SingleResultAuthorizationManager() {
```
jzheaux, line 54: While having this for `permitAll` and `denyAll` is great, we can expand it by taking an `AuthorizationResult` in the constructor and having it be public.

core/src/main/java/org/springframework/security/authorization/SingleResultAuthorizationManager.java
```diff
@@ -0,0 +1,62 @@
...
+		return (a, o) -> PERMIT;
+	}
+
+	/**
+	 * Creates deny-all {@link AuthorizationManager} instance.
+	 * @param <C>
+	 * @return deny-all {@link AuthorizationManager} instance
+	 */
+	public static <C> AuthorizationManager<C> DENY_ALL() {
+		return (a, o) -> DENY;
```
jzheaux, line 51: Instead of a lambda, consider returning a `static final` instance of `SingleResultAuthorizationManager`.

core/src/main/java/org/springframework/security/authorization/SingleResultAuthorizationManager.java
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	private SingleResultAuthorizationManager() {
+	}
+
+	@Override
+	public AuthorizationDecision check(Supplier<Authentication> authentication, C object) {
+		throw new UnsupportedOperationException("Not supported");
```
jzheaux, line 59: Let's still implement this. If the private member `result` isn't of type `AuthenticationDecision`, we can throw and exception.

core/src/main/java/org/springframework/security/authorization/SingleResultAuthorizationManager.java
```diff
@@ -0,0 +1,62 @@
...
+	 */
+	public static <C> AuthorizationManager<C> DENY_ALL() {
+		return (a, o) -> DENY;
+	}
+
+	private SingleResultAuthorizationManager() {
+	}
+
+	@Override
+	public AuthorizationDecision check(Supplier<Authentication> authentication, C object) {
```
jzheaux, line 58: Let's please implement `authorize`, returning whatever single result was set in the constructor.

core/src/main/java/org/springframework/security/authorization/SingleResultAuthorizationManager.java
```diff
@@ -0,0 +1,62 @@
...
+import org.springframework.security.core.Authentication;
+
+/**
+ * An {@link AuthorizationManager} which creates permit-all and deny-all
+ * {@link AuthorizationManager} instances.
+ *
+ * @author Max Batischev
+ * @since 6.5
+ */
+public final class SingleResultAuthorizationManager<C> implements AuthorizationManager<C> {
```
jzheaux, line 30: Please add tests for the class.

jzheaux (changes requested):
Thanks again, @franticticktick! I've left my feedback inline. Note that most if it is based on the premise that I think it class should hold an `AuthorizationResult` member variable, set in the constructor. This will give this class many more uses.

core/src/main/java/org/springframework/security/authorization/SingleResultAuthorizationManager.java
franticticktick: No, we can't do this, java uses `LambdaMetafactory` for lambda creation. This factory creates `CallSite` object, which returns intenface implementation. Therefore these methods must return an interface.

franticticktick: Sorry, you can ignore this comment :)

franticticktick:
Hi @jzheaux, thanks for your feedback. There are several points that need to be discussed. We can add a static method, for example `denyAll`, and return a static final instance:
```
public static <C> SingleResultAuthorizationManager<C> denyAll() {
   return DENY_MANAGER;
}
```
This is only possible if the `SingleResultAuthorizationManager` instance has a raw type:
```
private static final SingleResultAuthorizationManager DENY_MANAGER = new SingleResultAuthorizationManager<>(new AuthorizationDecision(false));
```
Generics are not available in a static context, which forces us to use a raw type. I'm not sure if this is the best pattern in this case, @jzheaux what do you think about it?

jzheaux:
@franticticktick good questions. First, I think it's okay to construct a `SingleResultAuthorizationManager` each time `denyAll()` is called. This is conventionally the most type-safe and I'm fine going that route.

However, since the implementation knows that it won't use the type, then it can safely use `<?>`:

```java
private static final SingleResultAuthorizationManager<?> DENY = new SingleResultAuthorizationManager<>(new AuthorizationDecision(false));
```

That does require a cast in the static method, but again in practice this isn't a concern since this implementation never uses the type at runtime:

```java
public static <T> SingleResultAuthorizationManager<T> denyAll() {
    return (SingleResultAuthorizationManager<T>) DENY;
}
```

franticticktick:
Hey @jzheaux thanks for the feedback. I decided to go with the second solution, although static casting may not be to the liking of clean code fans :)

jzheaux (approved):

Issue #16617, opened 2025-02-18
Make DefaultOneTimeToken Serializable


franticticktick:
Since `OneTimeTokenService` involves storing tokens, it is necessary to make the `DefaultOneTimeToken` serializable.

PR #16618, opened 2025-02-18
Make DefaultOneTimeToken Serializable

franticticktick:
Closes gh-16617


jzheaux:
Hi, @franticticktick. Was there something released in 6.4 that is already `Serializable` that is using these components? If not, then this is a feature enhancement and should be on `main`. Adding functionality is a feature, while fixing existing incomplete or broken functionality is a bug.

The reason the other components are being updated on `6.4.x` is because those components *claim* to be serializable, but they aren't for one reason or another. So, those need fixing on the earliest maintenance branch possible.

franticticktick:
Hi @jzheaux thanks for the clarification. I came across several cases where developers tried to implement `ReactiveOneTimeTokenService `based on `Redis`. This is mainly due to the fact that we cannot yet make `ReactiveOneTimeTokenService `for a distributed environment. In addition, the `ReactiveOneTimeTokenService `implementation based on `Redis `looks convenient. If the `OneTimeToken `is not serializable, then such an implementation is impossible, although there is already a request for this. So I opened a ticket as a bug. Can I leave this ticket as is, but do a PR in `main `with backport in `6.4.x`?

jzheaux:
Good question. While perhaps a little inconvenient, you can create a `RedisOneTimeTokenService` implementation and  store a different object, for example a different implementation of `OneTimeToken` that is serializable.

I see that you've already moved it to `main` so perhaps you agree, but please feel free to say more if I'm missing something.

jzheaux:
Thanks, @franticticktick! This is now merged into `main`.

Issue #16620, opened 2025-02-19
Bug Report: Duplicate Primary Key Exception in Webauthn4J Authentication


Borghii:
# Describe the Bug

When attempting to authenticate using the following method:

`public PublicKeyCredentialUserEntity authenticate(RelyingPartyAuthenticationRequest request) `

from Webauthn4JRelyingPartyOperations  the process follows these steps:

1. Retrieve an existing credential record:

`CredentialRecord existingCredential = this.userCredentials.findByCredentialId(credentialId);`

2. Save the credential record:

`this.userCredentials.save(userCredential);`

![Image](https://github.com/user-attachments/assets/b5e0e1c1-9d7f-4ee0-8b5f-e242f789b1ff)

However, the save method attempts to execute the following SQL insert:

![Image](https://github.com/user-attachments/assets/d5a4dd40-bcd6-4a1d-ad00-7b756b13d899)

	// @formatter:off
	private static final String SAVE_CREDENTIAL_RECORD_SQL = "INSERT INTO " + TABLE_NAME
			+ " (" + COLUMN_NAMES + ") VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)";

This results in a duplicate primary key exception, as the credential record already exists in the database.

I don't know if I'm missing something.

Borghii:
@franticticktick can you revise it?

franticticktick:
Hi @Borghii , thanks for this catch :) it will be fixed via https://github.com/spring-projects/spring-security/pull/16621

PR #16621, opened 2025-02-19
Fix JdbcUserCredentialRepository Save

franticticktick:
Closes gh-16620

### DIFF
--- a/web/src/main/java/org/springframework/security/web/webauthn/management/JdbcUserCredentialRepository.java
+++ b/web/src/main/java/org/springframework/security/web/webauthn/management/JdbcUserCredentialRepository.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2024 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/springframework/security/web/webauthn/management/JdbcUserCredentialRepository.java
```diff
@@ -133,6 +152,21 @@ public void delete(Bytes credentialId) {
 	@Override
 	public void save(CredentialRecord record) {
 		Assert.notNull(record, "record cannot be null");
+		boolean existsRecord = null != this.findByCredentialId(record.getCredentialId());
```
jzheaux, line 155: I think the following logic achieves the same outcome:

```java
int rows = updateCredentialRecord(record); // change to return the number of rows affected
if (rows == 0) {
    insertCredentialRecord(record);
}
```

This would complete more quickly and also be less load on the database.

jzheaux (changes requested):
Hi, @franticticktick, I've left some feedback inline.

jzheaux:
Also, will you please rebase to `6.4.x` since that is where the bug first appears?

web/src/main/java/org/springframework/security/web/webauthn/management/JdbcUserCredentialRepository.java
franticticktick: Good solution. I think we can do this polish in other places with similar logic.

franticticktick:
Hi @jzheaux . Thnaks for your feedback. This component was introduced in current version, so it is not in `6.4.x` branch.

web/src/main/java/org/springframework/security/web/webauthn/management/JdbcUserCredentialRepository.java
```diff
@@ -133,6 +152,18 @@ public void delete(Bytes credentialId) {
 	@Override
 	public void save(CredentialRecord record) {
 		Assert.notNull(record, "record cannot be null");
+		int rows = updateCredentialRecord(record);
+		if (rows == 0) {
+			try {
+				insertCredentialRecord(record);
+			}
+			catch (DuplicateKeyException ex) {
+				updateCredentialRecord(record);
```
jzheaux, line 161: Under what circumstance would you get a `DuplicateKeyException` if an update returns zero rows?

jzheaux (changes requested):
Thanks, @franticticktick! I've left a question for you inline.

web/src/main/java/org/springframework/security/web/webauthn/management/JdbcUserCredentialRepository.java
franticticktick: Good point, it worked before but it's not needed anymore.

jzheaux:
Thanks, @franticticktick! Can you squash your commits please? Then I think we're ready to merge.

franticticktick:
@jzheaux sory please, i usually don't miss things like that.

jzheaux (approved):

jzheaux:
Thanks, @franticticktick! This is now merged into `main`.

PR #16628, opened 2025-02-21
Fix @PostResult example in method-security doc

Kuba15:
Replace @PreFilter with @Postfilter in example


jzheaux:
Thanks, @Kuba15! This is now merged into `6.4.x` and `main`.

PR #16633, opened 2025-02-22
Fix typo `security-api-url` attribute in `faq.adoc`

ngocnhan-tran1996:


jzheaux:
Thanks, @ngocnhan-tran1996, this is now merged into `6.3.x`, `6.4.x`, and `main`.

In the future, would you be able to rebase these documentation fixes to the earliest supported branch where the bug appears? It's nothing big, but it does save time.

PR #16634, opened 2025-02-23
Remove dependency on oauth2-oidc-sdk inside NimbusOpaqueTokenIntrospector

asvanberg:
`NimbusOpaqueTokenIntrospector` now delegates to `SpringOpaqueTokenIntrospector`.

Required a lot of test changes since they are all mock-based relying on the implementation calling a specific method in the `RestOperations` interface and the two implementations differ in which method they call even though they are functionally the same. Tests were changed to use a `MockWebServer` and make real HTTP requests to enable the internals to be freely refactored.

One step of gh-14245


jzheaux:
Thanks for the PR, @asvanberg, however `NimbusOpaqueTokenIntrospector` will be removed in 7, and that would be the earliest that `oauth2-oidc-sdk` is removed. Because of that there won't exist a time when we remove `oauth2-oidc-sdk` and still have `NimbusOpaqueTokenIntrospector` around. Indeed, the opposite will likely be true. Because of that, refactoring the Nimbus implementation likely doesn't buy anything for us at this point.

PR #16646, opened 2025-02-25
Update doc: disable CSRF protection when using REST API for login.

junjieyuan:
If CSRF protection is not disabled, Spring Security will return a 403 for the permitted `/login` API.


sjohnr:
@junjieyuan thanks for the PR. However, this is not an improvement to the docs, as it is not recommended to disable CSRF without careful considerations which are listed elsewhere in the docs. I'm going to close this.

Issue #16501, opened 2025-01-29
Update HandlerMappingIntrospector Usage in CORS support


jzheaux:
The `CorsConfigurationConfigurer` should look up a `PreFlightRequestHandler` instead of a `HandlerMappingIntrospector`. It will still be registered under the same bean name in Framework 7, so that should remain as-is.

If a `PreFlightRequestHandler` is found by the given name, then safely cast it to `CorsConfigurationSource` and construct the `CorsFilter` as normal.

jzheaux:
@evgeniycheban, this ticket is yours. Once you comment, I'll be able to assign it to you officially.

evgeniycheban:
Hi, @jzheaux, please assign it to me.

evgeniycheban:
Hi, @jzheaux I opened a PR to your fork.

PR #16657, opened 2025-02-26
Update HandlerMappingIntrospector Usage in CORS support

evgeniycheban:
Closes gh-16501

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/CorsConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/CorsConfigurer.java
@@ -107,15 +107,14 @@ public class CorsConfigurer<H extends HttpSecurityBuilder<H>> extends AbstractHt
 		 * @return
 		 */
 		private static CorsFilter getMvcCorsFilter(ApplicationContext context) {
-			if (!context.containsBean(HANDLER_MAPPING_INTROSPECTOR_BEAN_NAME)) {
-				throw new NoSuchBeanDefinitionException(HANDLER_MAPPING_INTROSPECTOR_BEAN_NAME, "A Bean named "
-						+ HANDLER_MAPPING_INTROSPECTOR_BEAN_NAME + " of type "
-						+ HandlerMappingIntrospector.class.getName()
-						+ " is required to use MvcRequestMatcher. Please ensure Spring Security & Spring MVC are configured in a shared ApplicationContext.");
+			if (context.getBeanNamesForType(CorsConfigurationSource.class).length > 0) {
+				CorsConfigurationSource corsConfigurationSource = context
+					.getBean(HANDLER_MAPPING_INTROSPECTOR_BEAN_NAME, CorsConfigurationSource.class);
+				return new CorsFilter(corsConfigurationSource);
 			}
-			HandlerMappingIntrospector mappingIntrospector = context.getBean(HANDLER_MAPPING_INTROSPECTOR_BEAN_NAME,
-					HandlerMappingIntrospector.class);
-			return new CorsFilter(mappingIntrospector);
+			throw new NoSuchBeanDefinitionException(CorsConfigurationSource.class,
+					"Failed to find a bean that implements `CorsConfigurationSource`. Please ensure that you are using "
+							+ "`@EnableWebMvc`, are publishing a `WebMvcConfigurer`, or are publishing a `CorsConfigurationSource` bean.");
 		}
 
 	}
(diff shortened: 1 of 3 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/configurers/CorsConfigurer.java
```diff
@@ -109,13 +110,14 @@ static class MvcCorsFilter {
...
 				throw new NoSuchBeanDefinitionException(HANDLER_MAPPING_INTROSPECTOR_BEAN_NAME, "A Bean named "
-						+ HANDLER_MAPPING_INTROSPECTOR_BEAN_NAME + " of type "
-						+ HandlerMappingIntrospector.class.getName()
+						+ HANDLER_MAPPING_INTROSPECTOR_BEAN_NAME + " of type " + PreFlightRequestHandler.class.getName()
 						+ " is required to use MvcRequestMatcher. Please ensure Spring Security & Spring MVC are configured in a shared ApplicationContext.");
 			}
-			HandlerMappingIntrospector mappingIntrospector = context.getBean(HANDLER_MAPPING_INTROSPECTOR_BEAN_NAME,
-					HandlerMappingIntrospector.class);
-			return new CorsFilter(mappingIntrospector);
+			PreFlightRequestHandler mappingIntrospector = context.getBean(HANDLER_MAPPING_INTROSPECTOR_BEAN_NAME,
```
jzheaux, line 116: Along the same lines, since `CorsFilter` doesn't require `PreFlightRequestHandler`, perhaps this should simply work with `CorsConfigurationSource.class`. I believe this would also remove the `isInstanceOf` check below.

jzheaux (changes requested):

config/src/main/java/org/springframework/security/config/annotation/web/configurers/CorsConfigurer.java
```diff
@@ -109,13 +110,14 @@ static class MvcCorsFilter {
 		private static CorsFilter getMvcCorsFilter(ApplicationContext context) {
 			if (!context.containsBean(HANDLER_MAPPING_INTROSPECTOR_BEAN_NAME)) {
 				throw new NoSuchBeanDefinitionException(HANDLER_MAPPING_INTROSPECTOR_BEAN_NAME, "A Bean named "
-						+ HANDLER_MAPPING_INTROSPECTOR_BEAN_NAME + " of type "
-						+ HandlerMappingIntrospector.class.getName()
+						+ HANDLER_MAPPING_INTROSPECTOR_BEAN_NAME + " of type " + PreFlightRequestHandler.class.getName()
```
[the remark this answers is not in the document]
evgeniycheban: Done.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/CorsConfigurer.java
evgeniycheban: I changed it to lookup for `CorsConfigurationSource` bean directly.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/CorsConfigurer.java
```diff
@@ -98,24 +99,20 @@ private CorsFilter getCorsFilter(ApplicationContext context) {
...
 		 * @param context
 		 * @return
 		 */
 		private static CorsFilter getMvcCorsFilter(ApplicationContext context) {
-			if (!context.containsBean(HANDLER_MAPPING_INTROSPECTOR_BEAN_NAME)) {
-				throw new NoSuchBeanDefinitionException(HANDLER_MAPPING_INTROSPECTOR_BEAN_NAME, "A Bean named "
-						+ HANDLER_MAPPING_INTROSPECTOR_BEAN_NAME + " of type "
-						+ HandlerMappingIntrospector.class.getName()
-						+ " is required to use MvcRequestMatcher. Please ensure Spring Security & Spring MVC are configured in a shared ApplicationContext.");
+			if (context.getBeanNamesForType(CorsConfigurationSource.class).length > 0) {
```
jzheaux, line 109: We still need to look need to look it up by name in order to remain passive. Otherwise, if there are multiple in the application context, this logic will throw an error where before it didn't.

jzheaux (changes requested):
Thanks, @evgeniycheban! I think we still need to lookup the `CorsConfigurationSource` by name, please see my inline comment.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/CorsConfigurer.java
evgeniycheban: Hi, @jzheaux, there's no method in `ApplicationContext` which checks that a bean with specific name and type exists, should we check it using `context.getBeanNamesForType(CorsConfigurationSource.class)` and then find the one with name `mvcHandlerMappingIntrospector` and type `CorsConfigurationSource`?

evgeniycheban:
@jzheaux I have updated the PR

jzheaux:
Thanks, @evgeniycheban! This is now merged into `main`. I also added a polish to use `containsBean`: e65e32bb42377e9b8cfa48cd7074d29a5630aab9

PR #16682, opened 2025-02-28
Introduce JwtAudienceValidator

vpavic:
Since audience is among claims that are often validated, this PR introduces `JwtAudienceValidator` to simplify such use case with `spring-security-oauth2-jose`. I've also plugged the new validator into recently introduced `at+jwt` support (thanks for that).


jzheaux:
Nice to see you again, @vpavic! This is merged into `main`.

vpavic:
Likewise! Thanks for the quick merge. :+1:

PR #16709, opened 2025-03-09
Customizable oauth response

dafafdafafadfa:
Additional parameters in oauth response for unconventional oauth providers

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationResponseUtils.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationResponseUtils.java
@@ -16,13 +16,14 @@
 
 package org.springframework.security.oauth2.client.web;
 
-import java.util.Map;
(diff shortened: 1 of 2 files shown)


jgrandja:
@dafafdafafadfa In case you are not aware, our process for feature or enhancement requests is to log an issue first (before PR) to understand the use case and to determine if the feature/enhancement is valid and if it's something the community would actually use.

If it's a valid use case, the issue will sit for a while until we determine if it's in demand (based on upvotes) and then we will schedule it for a release.

I'm going to close this PR but feel free to add an issue with as much detail as possible.

dafafdafafadfa:
> @dafafdafafadfa In case you are not aware, our process for feature or enhancement requests is to log an issue first (before PR) to understand the use case and to determine if the feature/enhancement is valid and if it's something the community would actually use.
> 
> If it's a valid use case, the issue will sit for a while until we determine if it's in demand (based on upvotes) and then we will schedule it for a release.
> 
> I'm going to close this PR but feel free to add an issue with as much detail as possible.

In case you not aware, I don't give a shit what your process is.

PR #16710, opened 2025-03-10
Add not null validation for UserDetailsChecker in AbstractUserDetailsAuthenticationProvider

pogihae:
# Add early validation for authentication checks

## Details
This PR adds validation for `preAuthenticationChecks` and `postAuthenticationChecks` in 
`AbstractUserDetailsAuthenticationProvider`. The primary approach is to add assertions in
the `afterPropertiesSet()` method to detect null values early in the application lifecycle.

Currently, if these checks are set to null, a `NullPointerException` is thrown during the 
authentication process, which might be difficult to troubleshoot. The early validation approach
ensures errors are detected at initialization time with a clear error message.

## Alternative Approach Considered
As an alternative, we could add null checks before invoking the `check()` methods:
```java
if (this.preAuthenticationChecks != null) {
    this.preAuthenticationChecks.check(user);
}

### DIFF
--- a/core/src/main/java/org/springframework/security/authentication/dao/AbstractUserDetailsAuthenticationProvider.java
+++ b/core/src/main/java/org/springframework/security/authentication/dao/AbstractUserDetailsAuthenticationProvider.java
@@ -117,6 +117,8 @@ public abstract class AbstractUserDetailsAuthenticationProvider
 	public final void afterPropertiesSet() throws Exception {
 		Assert.notNull(this.userCache, "A user cache must be set");
 		Assert.notNull(this.messages, "A message source must be set");
+		Assert.notNull(this.preAuthenticationChecks, "A pre authentication checks must be set");
+		Assert.notNull(this.postAuthenticationChecks, "A post authentication checks must be set");
 		doAfterPropertiesSet();
 	}
 


franticticktick:
Hi @pogihae , thanks for this polish. I'm not sure that such a check in `afterPropertiesSet` is the best solution to this problem - `preAuthenticationChecks` and `postAuthenticationChecks` already have default values, and if we haven't changed them, then there's no particular point in checking. 
But we can add nonNull checks to `setPostAuthenticationChecks` and `setPreAuthenticationChecks` methods, thus we ensure that null never gets into `preAuthenticationChecks` and `postAuthenticationChecks`.

pogihae:
Hi @franticticktick . thanks for comment.

You're right. I added the null checks in afterPropertiesSet() because I noticed that other components like message source and user cache also perform null checks there despite having default values. I thought this might be a convention for the codebase, but perhaps I misunderstood.

jzheaux:
Hi, @pogihae. Thanks for the PR. The convention is an older one that Spring Security is replacing with validating the setter parameters as @franticticktick explained.

That said, sticking with a class's internal convention is often helpful to maintaining the class's readability. For this PR, let's continue checking in `afterPropertiesSet`. 

We can also open an issue to move afterPropertiesSet checks into setters/constructors in 7.

jzheaux:
Thanks, @pogihae! This is now merged into `main`.

PR #16714, opened 2025-03-10
Sets correct remote address in WebAuthenticationDetails

clzola:
While I was going trough debug logs of my application on my test server I have noticed this line:

```
Set SecurityContextHolder to JwtAuthenticationToken [Principal=org.springframework.security.oauth2.jwt.Jwt@b2c2a23e, Credentials=[PROTECTED], Authenticated=true, Details=WebAuthenticationDetails [RemoteIpAddress=127.0.0.1, SessionId=null], Granted Authorities=[ROLE_offline_access, ROLE_default-roles-app, ROLE_uma_authorization, ROLE_consumer]]
```

Our application on test server is behind nginx, so because of that attribute `RemoteIpAddress` is set to 127.0.0.1 which is wrong...

So I have updated WebAuthenticationDetails to read remote address first from `X-Forwarded-For`, and then from `X-Real-IP` headers and then finally as fallback to use `request.getRemoteAddr()`. This sould cover almost all use cases I guess...

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/WebAuthenticationDetails.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/WebAuthenticationDetails.java
@@ -29,6 +29,7 @@ import org.springframework.security.core.SpringSecurityCoreVersion;
  *
  * @author Ben Alex
  * @author Luke Taylor
+ * @author Lazar Radinović
  */
 public class WebAuthenticationDetails implements Serializable {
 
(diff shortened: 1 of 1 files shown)


jzheaux:
Spring Security does not look for these headers by design since [Boot supports adapts these headers](https://docs.spring.io/spring-boot/how-to/webserver.html#howto.webserver.use-behind-a-proxy-server).

If you are not using Boot, you can find non-Boot advice [in the Spring Security reference](https://docs.spring.io/spring-security/reference/servlet/appendix/proxy-server.html#page-title).

PR #16729, opened 2025-03-12
Fix typos Open SAML 5 Javadoc referencing Open SAML 4

filiphr:
Fixing some typos that I found while integrating with Open SAML 5


jzheaux:
Thanks, @filiphr! This is now merged into `main`.

PR #16741, opened 2025-03-16
Don't cache WebSocket request

yuezk:
The WebSocket connection request should not be cached. Since the websocket request doesn't accept a custom header, we cannot use `X-Requested-With` as the workaround. Instead, we can match the `Upgrade: websocket` header and ignore it by default.


rwinch:

Issue #16700, opened 2025-03-06
AuthorizationDenied and Granted EventListeners Should Support Generics


jzheaux:
It would be nice if users could listen for authorization events specific to the type of source:

```java
@EventListener
public void onEvent(AuthorizationDeniedEvent<MethodInvocation> event) {
    // ...
}
```

One way this can be done is by having `AuthorizationDeniedEvent` and `AuthorizationGrantedEvent` implement `ResolvableTypeProvider`:

```java
ResolvableType type = ResolvableType.forClassWithGenerics(getClass(), ResolvableType.forInstance(getObject()));
```

franticticktick:
Hi @jzheaux, i would like to work on this ticket, could you assign it to me please?

PR #16762, opened 2025-03-18
Add support ResolvableTypeProvider to AuthorizationEvent

franticticktick:
Closes gh-16700

### DIFF
--- a/config/src/test/java/org/springframework/security/config/annotation/method/configuration/PrePostMethodSecurityConfigurationTests.java
+++ b/config/src/test/java/org/springframework/security/config/annotation/method/configuration/PrePostMethodSecurityConfigurationTests.java
@@ -51,11 +51,13 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.beans.factory.config.BeanPostProcessor;
 import org.springframework.beans.factory.support.BeanDefinitionRegistryPostProcessor;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/springframework/security/authorization/event/AuthorizationEvent.java
```diff
@@ -110,4 +112,14 @@ public AuthorizationResult getAuthorizationResult() {
 		return this.result;
 	}
 
+	/**
+	 * Get {@link ResolvableType} of this class.
+	 * @return {@link ResolvableType}
+	 * @since 6.5
+	 */
+	@Override
+	public ResolvableType getResolvableType() {
```
jzheaux, line 121: Since this class doesn't have generics, I think it will have unintended consequences if placed here. Will you please move this down to `AuthorizationGranted` and `AuthorizationDeniedEvent`?

jzheaux (changes requested):
Thanks, @franticticktick! I've left feedback inline.

jzheaux (approved):

jzheaux:
Thanks for the PR, @franticticktick! Note that I added https://github.com/spring-projects/spring-security/pull/16762/commits/9179db999b2c35cbd859607b4378c569ca159673 to include tests.

Even for simple features like these, tests are helpful for two reasons:

1. Ensure that a feature works. This also helps us have a test to point to when troubleshooting an issue with a reporter.
2. Shows a sample of how to use the feature

PR #16818, opened 2025-03-25
Add toString to IpAddressMatcher

wtigerhyunsu:
Closes gh-16795
Implemented the toString method based on gh-16795.

In the future, if getHostAddress() or getMaskBits() methods are introduced, the internal logic of toString may be updated to use those methods.

Fixed a formatting issue detected in build #5695 (checkFormatTest failure) and resubmitted the PR.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/util/matcher/IpAddressMatcher.java
+++ b/web/src/main/java/org/springframework/security/web/util/matcher/IpAddressMatcher.java
@@ -130,4 +130,11 @@ public final class IpAddressMatcher implements RequestMatcher {
 		}
 	}
 
+	@Override
+	public String toString() {
+		String hostAddress = this.requiredAddress.getHostAddress();
+		return (this.nMaskBits < 0) ? "IpAddress [" + hostAddress + "]"
+				: "IpAddress [" + hostAddress + "/" + this.nMaskBits + "]";
+	}
+
 }
--- a/web/src/test/java/org/springframework/security/web/util/matcher/IpAddressMatcherTests.java
+++ b/web/src/test/java/org/springframework/security/web/util/matcher/IpAddressMatcherTests.java
@@ -153,4 +153,18 @@ public class IpAddressMatcherTests {
 			.withMessage("ipAddress cannot be empty");
 	}
 
+	// gh-16795
+	@Test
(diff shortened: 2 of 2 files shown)


web/src/main/java/org/springframework/security/web/util/matcher/IpAddressMatcher.java
```diff
@@ -130,4 +130,11 @@ private InetAddress parseAddress(String address) {
 		}
 	}
 
+	@Override
+	public String toString() {
+		String hostAddress = this.requiredAddress.getHostAddress();
+		return (this.nMaskBits < 0) ? "IpAddressMatcher[" + hostAddress + "]"
```
jzheaux, line 136: Will you please format this as `IpAddress [127.0.0.1]`. That is:

1. Remove the word `Matcher`, and
2. Add a space between `IpAddress` and the left bracket

jzheaux (changes requested):
Thanks for the PR, @wtigerhyunsu! I'm glad we were able to together on a solution. I've left some feedback inline, would you be able to make those changes please?

Since the feedback is quite brief, please squash your commits the next time you push so that there continues to be only one commit.

web/src/test/java/org/springframework/security/web/util/matcher/IpAddressMatcherTests.java
```diff
@@ -153,4 +153,18 @@ public void constructorWhenRequiredAddressIsEmptyThenThrowsIllegalArgumentExcept
 			.withMessage("ipAddress cannot be empty");
 	}
 
+	// gh-16795
+	@Test
+	public void toStringWhenCidrIsProvidedThenReturnsIpAddressWithCidr() {
+		IpAddressMatcher matcher = new IpAddressMatcher("192.168.1.0/24");
+		assertThat(matcher.toString()).hasToString("IpAddressMatcher[192.168.1.0/24]");
```
jzheaux, line 160: It looks like the tests don't match output of the method. Can you double-check please?

jzheaux (changes requested):
Thanks for the update, @wtigerhyunsu! I've left another piece of inline feedback.

jzheaux (approved):

jzheaux:
Thanks, @wtigerhyunsu! This is now merged into `main`

Issue #16825, opened 2025-03-26
Not all OAuth2 ClientAuthenticationMethods are supported in Jackson2 converters


mapsu:
**Describe the bug**
StdConverters do not support all available OAuth2 ClientAuthenticationMethods
https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/jackson2/StdConverters.java#L48-L65

For example `ClientAuthenticationMethod.PRIVATE_KEY_JWT` authentication method is missing from mapping.
In case `private_key_jwt` is used for client authentication and authorized clients are stored in Redis then when loading and deserializing ClientRegistration from json the `clientAuthenticationMethod` is not correctly converted and then later defaulted to incorrect value.

**Precondition**
- Project using Spring Security configured to store session data to Redis
  - Use `org.springframework.session:spring-session-data-redis` library to achieve this
- Store OAuth2 authorised clients to Http Session instead of InMemory
  - Configure OAuth2AuthorizedClientRepository to HttpSessionOAuth2AuthorizedClientRepository

**To Reproduce**
- Authenticate with a user
  - The authorized client is stored to Redis with client registration details
- When access token expires and OAuth2AuthorizedManager tries to refresh token then incorrect authentication method is used

**Expected behavior**
In token refresh, correct authentication method is used.

**Sample**
I decided to not include a sample this time because it would require setting up redis, e.g. in docker compose, but also authorization server for OAuth2 login.
The bug is quite obvious for me because there's missing ClientAuthenticationMethods missing from converter.

I can create a sample if you deem it's needed.

jzheaux:
Closed in favor of https://github.com/spring-projects/spring-security/pull/16826

PR #16826, opened 2025-03-26
Jackson deserialization of ClientAuthenticationMethods should recognize all values

mapsu:
Closes gh-16825


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/jackson2/StdConverters.java
```diff
@@ -56,9 +56,21 @@ public ClientAuthenticationMethod convert(JsonNode jsonNode) {
 			if (ClientAuthenticationMethod.CLIENT_SECRET_POST.getValue().equalsIgnoreCase(value)) {
 				return ClientAuthenticationMethod.CLIENT_SECRET_POST;
 			}
+			if (ClientAuthenticationMethod.CLIENT_SECRET_JWT.getValue().equalsIgnoreCase(value)) {
```
jzheaux, line 59: Since we already forgot once, I'm concerned we'll forget again. As such, I've added [a commit](https://github.com/spring-projects/spring-security/commit/2885b0f75f6ba42693e0f2eab0c2f267722d9e42) that may help with this. 

Would you consider rebasing and then changing this method to use `ClientAuthenticationMethod.valueOf`?

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/jackson2/ClientRegistrationMixinTests.java
```diff
@@ -0,0 +1,112 @@
...
+				"    \"issuerUri\":\"https://example.com\"," +
+				"    \"configurationMetadata\":{" +
+				"      \"@class\":\"java.util.Collections$UnmodifiableMap\"" +
+				"    }" +
+				"  }," +
+				"  \"clientName\":\"Client Name\"}";
+		// @formatter:on
+	}
+
+	static Stream<Arguments> deserializeWhenMixinRegisteredThenDeserializes() {
```
jzheaux, line 93: I think that if you change to use `ClientAuthenticationMethod.valueOf`, then this test can simplify to just ensure that it deserializes the value to a `ClientAuthenticationMethod`. One way to do this would be to test a custom value with `ClientAuthenticationMethod`'s constructor.

Otherwise, `ClientAuthenticationMethod.valueOf` is being tested separately.

jzheaux (changes requested):
Thanks, @mapsu, for the PR! I've left feedback inline.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/jackson2/StdConverters.java
mapsu: Rebased and changed the implementation. Thank you for the `valueOf` method.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/jackson2/ClientRegistrationMixinTests.java
mapsu: I replaced the ClientRegistrationMixinTests with StdConvertersTest. What do you think?

jzheaux (approved):

jzheaux:
Thanks for the PR, @mapsu! This is now merged into `main`.

Issue #16819, opened 2025-03-25
Sorting in AuthorizationAdvisorProxyFactory should be thread-safe


wbxz987:
Using Spring Security 6.4.4 (via Spring Boot 3.4.4)

**Describe the bug**
When a endpoint annotated with `@AuthorizeReturnObject` returns an object that has an object authorized with e.g. `@PreAuthorize`, `ConcurrentModificationException` is thrown when multiple requests come in parallel.

**To Reproduce**
See the sample repository below for a reproducible test case.

**Expected behavior**
Parallel requests should be handled without errors.

**Sample**
https://github.com/wbxz987/ConcurrentModificationException

The sample repository contains a test, that simulates multiple requests coming in parallel. The test fails because a `ConcurrentModificationException` is thrown.

```
Caused by: java.util.ConcurrentModificationException
	at java.base/java.util.ArrayList.sort(ArrayList.java:1806)
	at org.springframework.core.annotation.AnnotationAwareOrderComparator.sort(AnnotationAwareOrderComparator.java:111)
	at org.springframework.security.authorization.method.AuthorizationAdvisorProxyFactory.proxy(AuthorizationAdvisorProxyFactory.java:168)
	at org.springframework.security.authorization.method.AuthorizeReturnObjectMethodInterceptor.invoke(AuthorizeReturnObjectMethodInterceptor.java:61)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:184)
	at org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:727)
	at org.example.concurrentmodificationexception.controller.Controller$$SpringCGLIB$$0.getModel(<generated>)
```

The test works when downgrading Spring Security to version 6.3.2, and breaks after this commit https://github.com/spring-projects/spring-security/commit/0cab7c8f15ab98e97b10458d63ed3d85601b0903

jzheaux:
Thanks for the detailed report, @wbxz987! We'll target fixing this in the next maintenance release.

In the meantime, you can workaround this by providing your own proxy factory instance like so:

```java
@Primary
@Bean
@Role(BeanDefinition.ROLE_INFRASTRUCTURE)
static AuthorizationProxyFactory proxyFactory() {
	AuthorizationAdvisorProxyFactory proxyFactory = new AuthorizationAdvisorProxyFactory(new ArrayList<>());
[...]

PR #16834, opened 2025-03-27
Sorting in AuthorizationAdvisorProxyFactory should be thread-safe

ngocnhan-tran1996:
Closes gh-16819

### DIFF
--- a/core/src/main/java/org/springframework/security/authorization/method/AuthorizationAdvisorProxyFactory.java
+++ b/core/src/main/java/org/springframework/security/authorization/method/AuthorizationAdvisorProxyFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2024 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/springframework/security/authorization/method/AuthorizationAdvisorProxyFactory.java
```diff
@@ -165,7 +169,13 @@ public static AuthorizationAdvisorProxyFactory withReactiveDefaults() {
 	 */
 	@Override
 	public Object proxy(Object target) {
-		AnnotationAwareOrderComparator.sort(this.advisors);
```
jzheaux, line 168: I think we can likely do this without locking every time and possibly even without a lock.

Can you first try and implement `SmartInitializingBean#afterSingletons` move this sort call to that method? In this way, it will be called only once and also after the related method interceptors have been initialized.

core/src/test/java/org/springframework/security/authorization/AuthorizationAdvisorProxyFactoryTests.java
```diff
@@ -360,6 +362,19 @@ public void proxyWhenDefaultsThenInstanceOfAuthorizationProxy() {
 		assertThat(target).isSameAs(this.flight);
 	}
 
+	@Test
+	public void asyncProxyWhenDefaultsThenNoException() {
+		AuthorizationAdvisorProxyFactory factory = AuthorizationAdvisorProxyFactory.withDefaults();
+		List<CompletableFuture<Void>> futures = new ArrayList<>();
+
+		for (int i = 0; i < 1000; i++) {
```
jzheaux, line 370: Testing concurrency is quite tricky as it is non-deterministic. 

Let's instead consider a test in `PrePostMethodSecurityConfiguration` that can confirm that the advisors in the `AuthorizationAdvisorProxyFactory` bean are ordered.

jzheaux (changes requested):
Thanks, @ngocnhan-tran1996 for the PR! I've left some feedback inline.

Issue #16390, opened 2025-01-09
Simplify support of username as a nested property


rwinch:

yybmion:
Hi @rwinch, I'd like to work on this issue - could you please assign it to me?

PR #16857, opened 2025-04-01
Add support for nested user-name-attribute using dot notation

yybmion:
Closes gh-16390

Hi team, this PR adds support for nested properties in the OAuth2 user-name-attribute using dot notation.
For example, if a provider returns a nested response like `{"data": {"username": "user1"}}`,
users can now directly configure `user-name-attribute: data.username` without writing custom code.

## Implementation
- Added extractNestedAttribute utility method to extract values from nested paths
- Modified loadUser method to handle attributes with dot notation
- Applied the same changes to both servlet and reactive implementations
- Added comprehensive tests to verify functionality

## Testing
Added tests for
- Simple nested attributes 
- Multi-level nested attributes 

All tests pass and existing functionality is preserved.


rwinch:
Thanks for the PR @yybmion! I wonder if this might be better implemented using SpEL to provide more powerful options for resolving the username. What are your thoughts?

rwinch:
Yes. Please provide an implementation that uses SpEL.

yybmion:
Hello @rwinch, I'd like to clarify your feedback on my PR about supporting nested properties in the user-name-attribute.

Did you mean that I should implement support for expressions like `#{data.username}` in the properties and yml configuration files to handle nested structures? (Instead using `data.username`)

Thank you for your guidance!

rwinch:
I think an outline would be:

## Allow Injecting the Principal Name into `DefaultOAuth2User`

Right now `OAuth2UserService` requires defining the `name` property as an attribute name. This limits the flexibility of resolving the name.

Update `DefaultOAuth2User` to allow injecting the name (rather than a property for obtaining the name). You would add a new member variable of `username`. This would likely require using a static factory method to distinguish from the existing constructor since the types of the `nameAttributeKey` and `username` are the same.

You can remove the `nameAttributeKey` member variable and instead populate the `username` using the `attributes[nameAttributeKey]` in the constructor if `nameAttributeKey` was specified.

Deprecate the old `DefaultOAuth2User` constructor in favor of injecting the name directly and update the usage of `DefaultOAuth2User` within Spring Security to no longer use the deprecated constructor.

## Add SpEL Support

Update `ClientRegistration` to have a property named `usernameExpression` and  remove the `userNameAttributeName` property but preserving and deprecating the methods which instead populate the `usernameExpression` member variable. This should be passive since the `usernameAttributeName` is a valid SpEL expression. 

Update `OAuth2UserService` to extract the username using the `usernameExpression` property from the `ClientRegistration` as a SpEL expression with the attributes as the root object. Create the `DefaultOAuth2User` with the username injected into it rather than using the `nameAttributeKey`.

I think creating these as two separate commits is valuable since they are useful on their own. Let me know your thoughts.

yybmion:
Hi! @rwinch  I've implemented the first part of your suggestion - adding the `username` field to `DefaultOAuth2User` and the `withUsername()` static factory method.

For backward compatibility, I kept the `nameAttributeKey` field alongside the new `username` field. The implementation works as follows

```java
public class DefaultOAuth2User {
    private final String nameAttributeKey;  // kept for backward compatibility
    private final String username;          // newly added
    
    @Deprecated
    public DefaultOAuth2User(..., String nameAttributeKey) {
        this.nameAttributeKey = nameAttributeKey;
        // Extract username from attributes[nameAttributeKey]
        this.username = attributes.get(nameAttributeKey).toString();
    }
    
    public static DefaultOAuth2User withUsername(..., String username) {
        // Direct username injection
    }
    
    @Override
    public String getName() {
        return (this.username != null) ? this.username : getAttribute(nameAttributeKey).toString();
    }
}
```

**Issue:** The Jackson serialization tests (`OAuth2AuthenticationTokenMixinTests`)
1. `serializeWhenMixinRegisteredThenSerializes()`
2. `serializeWhenRequiredAttributesOnlyThenSerializes()`

are now failing because Jackson serializes all fields, including both `nameAttributeKey` and `username`. The test expects the old JSON structure but gets the new one with the additional `username` field.

**Question:** What's your preferred approach for handling this?

Thanks!

yybmion:
Hi @rwinch thank you for the detailed guide.

I've implemented the SpEL-based approach as outlined in the original issue. The solution provides

- SpEL expressions for nested property access (e.g., `"data.username"`)
- Full backward compatibility with existing `userNameAttributeName` 
- Support for both sync and reactive implementations

Please see the PR description for detailed changes.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
```diff
@@ -90,13 +97,69 @@ public DefaultOAuth2UserService() {
...
+
+	private String evaluateUsername(Map<String, Object> attributes, String usernameExpression) {
+		Object value = null;
+
+		if (attributes.containsKey(usernameExpression)) {
+			value = attributes.get(usernameExpression);
+		}
+		else {
+			try {
+				ExpressionParser parser = new SpelExpressionParser();
```
rwinch: The ExpressionParser should be reused as it is thread safe. You can safely make this a final member variable.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
```diff
@@ -90,13 +97,69 @@ public DefaultOAuth2UserService() {
...
+				StandardEvaluationContext context = new StandardEvaluationContext();
```
rwinch: Please use `SimpleEvaluationContext` instead since `StandardEvaluationContext` is less safe and unnecessary.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
```diff
@@ -90,13 +97,69 @@ public DefaultOAuth2UserService() {
...
+					null);
+			throw new OAuth2AuthenticationException(oauth2Error, oauth2Error.toString());
+		}
+		return usernameExpression;
+	}
+
+	private String evaluateUsername(Map<String, Object> attributes, String usernameExpression) {
+		Object value = null;
+
+		if (attributes.containsKey(usernameExpression)) {
```
rwinch: You should be able to remove this if statement and always use the expression if it is set as [described](https://github.com/spring-projects/spring-security/pull/16857/files#diff-c6947080fb4a385966afee5999dac1b734c5869f84c155e25b1390eacb65187aL561) on the builder.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
```diff
@@ -322,15 +326,27 @@ public AuthenticationMethod getAuthenticationMethod() {
...
+			 * response.
+			 * @return the SpEL expression for username extraction
+			 * @since 6.5
+			 */
+			public String getUsernameExpression() {
+				if (this.usernameExpression != null) {
+					return this.usernameExpression;
+				}
+
+				return this.userNameAttributeName;
```
rwinch: Please just return the `usernameExpression` without defaulting here. Instead ensure that the `userNameAttributeName` setter sets the `usernameExpression`.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
```diff
@@ -672,7 +700,14 @@ private ProviderDetails createProviderDetails(ClientRegistration clientRegistrat
 			providerDetails.tokenUri = this.tokenUri;
 			providerDetails.userInfoEndpoint.uri = this.userInfoUri;
 			providerDetails.userInfoEndpoint.authenticationMethod = this.userInfoAuthenticationMethod;
+			if (this.usernameExpression != null) {
```
rwinch: If you update the [builder](https://github.com/spring-projects/spring-security/pull/16857/files#diff-c6947080fb4a385966afee5999dac1b734c5869f84c155e25b1390eacb65187aR578) to set the expression when setting the `userNameAttributeName`, then the if statement can be removed

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
```diff
@@ -90,13 +97,69 @@ public DefaultOAuth2UserService() {
...
+		String usernameExpression = getUsernameExpression(userRequest);
 		RequestEntity<?> request = this.requestEntityConverter.convert(userRequest);
 		ResponseEntity<Map<String, Object>> response = getResponse(userRequest, request);
 		OAuth2AccessToken token = userRequest.getAccessToken();
 		Map<String, Object> attributes = this.attributesConverter.convert(userRequest).convert(response.getBody());
-		Collection<GrantedAuthority> authorities = getAuthorities(token, attributes, userNameAttributeName);
-		return new DefaultOAuth2User(authorities, attributes, userNameAttributeName);
+
+		String evaluatedUsername = evaluateUsername(attributes, usernameExpression);
+		Collection<GrantedAuthority> authorities = getAuthorities(token, attributes, usernameExpression);
```
rwinch: This is now passing in `usernameExpression` to `getAuthorities` which means that the `usernameExpression` is now passed into `OAuth2UserAuthority` in place of `userNameAttributeName`.

I think when users use more complex expressions it will break functionality introduced gh-15012 because the `OAuth2UserAuthority` only has an attribute name and not a property. We will need to figure out how to integrate cleanly with gh-15012 From the PR it sounds like we could just expose the username rather than the attribute, but I'm not certain how the integration is being used cc @filiphr

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
```diff
@@ -552,17 +572,25 @@ public Builder userInfoAuthenticationMethod(AuthenticationMethod userInfoAuthent
...
 		/**
-		 * Sets the attribute name used to access the user's name from the user info
-		 * response.
-		 * @param userNameAttributeName the attribute name used to access the user's name
-		 * from the user info response
-		 * @return the {@link Builder}
+		 * @deprecated Use {@link #usernameExpression(String)} instead
 		 */
+		@Deprecated
 		public Builder userNameAttributeName(String userNameAttributeName) {
```
rwinch, line 580: This is the place where I think you should default the usernameExpression to `"['" + userNameAttributeName + `"']"`.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
```diff
@@ -552,17 +572,25 @@ public Builder userInfoAuthenticationMethod(AuthenticationMethod userInfoAuthent
...
 
 		/**
-		 * Sets the attribute name used to access the user's name from the user info
-		 * response.
-		 * @param userNameAttributeName the attribute name used to access the user's name
-		 * from the user info response
-		 * @return the {@link Builder}
+		 * @deprecated Use {@link #usernameExpression(String)} instead
 		 */
+		@Deprecated
```
rwinch: I think that you can remove the Deprecated as this simplifies setting the expression. Just update the Javadoc.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistration.java
```diff
@@ -552,17 +572,25 @@ public Builder userInfoAuthenticationMethod(AuthenticationMethod userInfoAuthent
...
 		public Builder userNameAttributeName(String userNameAttributeName) {
 			this.userNameAttributeName = userNameAttributeName;
 			return this;
 		}
 
+		/**
+		 * Sets the SpEL expression used to extract the username from user info response.
```
rwinch, line 589: Please provide a few examples for the value.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/user/DefaultOAuth2User.java
```diff
@@ -77,11 +82,47 @@ public DefaultOAuth2User(Collection<? extends GrantedAuthority> authorities, Map
...
+	}
+
+	/**
+	 * Creates a new {@code DefaultOAuth2User} with the username injected directly.
+	 * @param authorities the authorities granted to the user
+	 * @param attributes the attributes about the user
+	 * @param username the user's name
+	 * @return a new {@code DefaultOAuth2User}
+	 */
+	public static DefaultOAuth2User withUsername(Collection<? extends GrantedAuthority> authorities,
```
rwinch: I think mentioned a factory method previously, but I do not like this with multiple arguments. Please create a `public static Builder withUsername(String)` method instead.

rwinch (changes requested):
Thank you for the updates. This is taking shape. I've provided feedback inline.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
yybmion: Done.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/userinfo/DefaultOAuth2UserService.java
yybmion: Thanks for highlighting this important integration issue with gh-15012!

I've addressed this by following your suggestion to expose the username directly rather than the attribute name.

- Added `username` field to OAuth2UserAuthority with `getUsername()` method
- Builder pattern: `OAuth2UserAuthority.withUsername(evaluatedUsername)`
- Pass the evaluated username result (not the SpEL expression) to OAuth2UserAuthority

I think this way, OAuth2UserAuthority always receives the final evaluated username string, preserving compatibility with gh-15012 regardless of expression complexity.

yybmion:
Thanks for the feedback @rwinch ! I've addressed all the review comments and updated the code accordingly.

But the build failure appears to be related to Kotlin version. What would you recommend I do?

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/user/OAuth2UserAuthority.java
```diff
@@ -44,6 +45,8 @@ public class OAuth2UserAuthority implements GrantedAuthority {
 
 	private final String userNameAttributeName;
```
rwinch, line 46: This property should be deprecated as should all the methods/constructors that use it.

Please also address `OidcUserAuthority` which extends from this class. It will likely need it's own `Builder` that extends from this `Builder` and have an overridden `withUsername` static method.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/user/OAuth2UserAuthority.java
```diff
@@ -88,6 +91,39 @@ public OAuth2UserAuthority(String authority, Map<String, Object> attributes, Str
...
+	/**
+	 * Constructs a {@code OAuth2UserAuthority} using the provided parameters. This
+	 * constructor is used by Jackson for deserialization.
+	 * @param authority the authority granted to the user
+	 * @param attributes the attributes about the user
+	 * @param userNameAttributeName the attribute name used to access the user's name from
+	 * the attributes - preserved for backwards compatibility
+	 * @param username the username
+	 */
+	private OAuth2UserAuthority(String authority, Map<String, Object> attributes, String userNameAttributeName,
```
rwinch: since the Builder doesn't use `usernameAttributeName`, we shouldn't include this as a property here, so that this constructor does not deprecated properties.

rwinch: Consider the order of username, authority, attributes since username is now a required property.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/user/OAuth2UserAuthority.java
```diff
@@ -113,6 +149,18 @@ public String getUserNameAttributeName() {
 		return this.userNameAttributeName;
 	}
 
+	/**
+	 * Returns the username of the OAuth2 user.
+	 * <p>
+	 * This method provides direct access to the username without requiring knowledge of
+	 * the attribute structure or SpEL expressions used to extract it.
+	 * @return the username
+	 * @since 6.5
```
rwinch: The PR targets main, so this (and any other since) should be since 7.0

rwinch (changes requested):
This is getting very close!

Please address the comments and ensure to update the documentation (including the What's New Section)

yybmion:
Hi @rwinch.  I've addressed all the review feedback and updated the documentation as requested. 
Please let me know if there are any areas in the documentation that need further improvement or clarification.

docs/modules/ROOT/pages/reactive/oauth2/client/core.adoc
```diff
@@ -36,12 +36,12 @@ public final class ClientRegistration {
 			private String uri;	<14>
 			private AuthenticationMethod authenticationMethod;  <15>
 			private String userNameAttributeName;	<16>
-
+            private String usernameExpression;    <17>
```
rwinch: Can you please fix this to use tabs?

docs/modules/ROOT/pages/reactive/oauth2/login/core.adoc
```diff
@@ -144,6 +144,9 @@ The following table outlines the mapping of the Spring Boot OAuth Client propert
 
 |`spring.security.oauth2.client.provider._[providerId]_.user-name-attribute`
 |`providerDetails.userInfoEndpoint.userNameAttributeName`
+
+|`spring.security.oauth2.client.provider._[providerId]_.username-expression`
+|`providerDetails.userInfoEndpoint.usernameExpression`
```
rwinch: This won't work yet because Spring Boot has not added support. Please remove it for now. If you like, you can save it in another branch in its own commit and we can use it when Spring Boot adds the support.

docs/modules/ROOT/pages/servlet/oauth2/login/core.adoc
```diff
@@ -142,6 +142,9 @@ The following table outlines the mapping of the Spring Boot OAuth Client propert
 
 |`spring.security.oauth2.client.provider._[providerId]_.user-name-attribute`
 |`providerDetails.userInfoEndpoint.userNameAttributeName`
+
+|`spring.security.oauth2.client.provider._[providerId]_.username-expression`
+|`providerDetails.userInfoEndpoint.usernameExpression`
```
rwinch: This won't work yet because Spring Boot has not added support. Please remove it for now. If you like, you can save it in another branch in its own commit and we can use it when Spring Boot adds the support.

docs/modules/ROOT/pages/whats-new.adoc
```diff
@@ -8,3 +8,29 @@ Below are the highlights of the release, or you can view https://github.com/spri
...
 * Added javadoc:org.springframework.security.web.authentication.preauth.x509.SubjectX500PrincipalExtractor[]
 * Added OAuth2 Support for xref:features/integrations/rest/http-interface.adoc[HTTP Interface Integration]
+
+== OAuth 2.0
+
+=== Username Expression Support for Nested Attributes - https://github.com/spring-projects/spring-security/pull/16390[gh-16390]
+
+OAuth2 Client now supports SpEL expressions for extracting usernames from nested UserInfo responses, eliminating the need for custom `OAuth2UserService` implementations in many cases. This is particularly useful for APIs like Twitter API v2 that return nested user data:
+
+[source,yaml]
```
rwinch: The yaml won't work yet because Spring Boot has not added support. Please remove it for now. If you like, you can save it in another branch in its own commit and we can use it when Spring Boot adds the support.

rwinch (changes requested):
Thank you for the changes.

I've included some comments inline around Spring Boot not supporting the new property yet. We will need to create a ticket with them once we merge here. Once they add the support we can add the documentation back.

Please go through and find/remove any usage (within main) for any of the constructors/methods that you deprecated. For example, I see that `OidcUserRequestUtils` still references `UserInfoEndpoint.getUserNameAttributeName()` We should update it to use non-deprecated methods.

yybmion:
Thank you for the review @rwinch ! 
I'm planning to create a separate commit to address the deprecated constructor/method usage you mentioned. However, I have a few questions that came up during the modifications.

I'm wondering about your thoughts on :

1. Deprecating the nameAttributeKey field in DefaultOAuth2User constructors for consistency with the new usernameExpression approach
2. Adding builder pattern to DefaultOidcUser since it extends DefaultOAuth2User and currently calls the deprecated parent constructor via super()

Should I include these changes in this PR?

rwinch:
Thanks for reaching out and your patience as we put the final touches on this @yybmion 

> Deprecating the nameAttributeKey field in DefaultOAuth2User constructors for consistency with the new usernameExpression approach

Yes. I think that this is good. Generally, we want people to move to the builder instead.

> Adding builder pattern to DefaultOidcUser since it extends DefaultOAuth2User and currently calls the deprecated parent constructor via super().

Yes I think that is good too. You will likely need to ensure the non-deprecated constructor in DefaultOAuth2User is protected so that it can be used by DefaultOidcUser.

> Should I include these changes in this PR?

Yes please. It is important that we push people towards the new, more powerful, approach.

yybmion:
Thank you for the guidance, @rwinch !
I've updated the deprecated constructor/method usage **in main** based on your review feedback. After searching through the codebase, I found that most of the deprecated usage was in `OidcUserRequestUtils` using deprecated constructors/methods. The remaining usage was in test code and documentation, which I left unchanged as requested.

Additionally, since the username SpEL evaluation logic was being used in multiple places, I created a utility class `OAuth2UsernameExpressionUtils` to centralize this functionality and eliminate code duplication.

yybmion:
Hi @rwinch, I noticed that the "What's New" documentation for Spring Security 7 has recently been updated, which caused a conflict with the changes I previously added. I've resolved the conflict accordingly.

yybmion:
Hi @rwinch,
I noticed that the "What's New" documentation for Spring Security 7 has recently been updated, which caused a conflict with the changes I previously added. I've resolved the conflict accordingly.

rwinch:
Thanks again for all of your work on this PR!

I'd like to encapsulate the logic (keep private) in `OAuth2UsernameExpressionUtils` so that we can refactor without breaking anything later. One example of such a refactoring is that at some point the `ExpressionParser` might need to be injected to allow resolving Bean references in the expressions. Even if we allow injecting the `ExpressionParser`, I'd still like to keep this logic encapsulated in case of other refactorings.

To work toward that end I created and resolved gh-17626. In the case where we invoke the `OAuth2UserService` (or the reactive equivalent), we could use the result to obtain the username. The problem is that OidcUserService (and the reactive equivalent) leverage  `userNameAttributeName` even when the `OAuth2Service` is not invoked and it does not use the result from the delegate service even when it does invoke the user info endpoint.

To address those concerns I created gh-17627 gh-17628

I believe we could extract the logic for injecting the username rather than the attribute into a separate PR and merge that without being blocked since that behavior would be passive and not be impacted by the decisions of the issues above. Feel free to do this if you like. However, as of now we are blocked on merging resolving the username using an expression until those two issues are resolved.

cc @jgrandja

rwinch:
FYI I rebased and pushed the commits

yybmion:
Thanks for the rebase and I understood about the blocking issues.

PR #16859, opened 2025-04-01
Update kotlin.adoc to add required spread operator(*)

msamborski-orbis:
During implementation of SecurityFilterChain in kotlin, it was noticed that the documented examples for 
public final fun securityMatcher(vararg pattern: String): Unit was missing the spread operator to turn the arrayOf into a varargs


rwinch:
Thanks for the PR! This will be merged as soon as the build completes.

Issue #16656, opened 2025-02-26
HttpSessionRequestCache#getMatchingRequest passes decoded Request URL to UriComponentsBuilder


ClausMie:
Hi there!
Thanks for taking a look at this issue. Please let me know if you require more information. 
Let's discuss!

**Describe the bug**

`HttpSessionRequestCache#getMatchingRequest` relies on `org.springframework.web.util.UriComponentsBuilder#fromUriString(String)` to be able to handle decoded `%` characters.

**To Reproduce**

The `org.springframework.web.util.UriComponentsBuilder#fromUriString(String)` cannot handle a `%` followed by characters that result in false %-encoding.

```  
@Test
void unencodedPercent(){
  assertThatRuntimeException().isThrownBy(() -> UriComponentsBuilder.fromUriString("/30 % off"));
}
```

This leads to a `org.springframework.web.util.InvalidUrlException: Bad path` when such a request gets to https://github.com/spring-projects/spring-security/blob/ec3cc66b647d35365c2f165c263f83ba3d27f063/web/src/main/java/org/springframework/security/web/savedrequest/HttpSessionRequestCache.java#L107
An encoded request `https://example.com/myapp/discounts/30%20%25%20off` is returned decoded by https://github.com/spring-projects/spring-security/blob/ec3cc66b647d35365c2f165c263f83ba3d27f063/web/src/main/java/org/springframework/security/web/util/UrlUtils.java#L91-L94
The `UriComponentsBuilder` cannot handle this decoded `%`

**Expected behavior**
My request does not result in an `InvalidUrlException`.

mkleine:
In fact every request containing a properly encoded % sign followed by a non-hex character will cause an exception in line https://github.com/spring-projects/spring-security/blob/ec3cc66b647d35365c2f165c263f83ba3d27f063/web/src/main/java/org/springframework/security/web/savedrequest/HttpSessionRequestCache.java#L107 as the path of the URI to be checked is decoded as documented https://github.com/spring-projects/spring-security/blob/ec3cc66b647d35365c2f165c263f83ba3d27f063/web/src/main/java/org/springframework/security/web/util/UrlUtils.java#L85

[...]

PR #16914, opened 2025-04-09
Fix HttpSessionRequestCache#getMatchingRequest query string parsing

Kehrlann:
## Issue

URL parsing changed in framework 6.2 (security 6.4), and now fails when path contains a % sign. As such, the `HttpSessionRequestCache` fails on every request containing a % sign.

## Resolution

We do not need to parse a full URL, just to inspect the query string for the matching parameter. We don't use the current request path in `UriComponentsBuilder` to extract the query string.

Fixes gh-16656


jzheaux:
Thanks for the PR, @Kehrlann. This has been merged to `6.5.x`, `7.0.x`, and `main`.

Issue #16939, opened 2025-04-15
Add support one-time token value customization


franticticktick:
It is necessary to add the ability to customize the generation of the one-time token value. Now it is:

```
String token = UUID.randomUUID().toString();
```

evgeniycheban:
I think we could introduce a `OneTimeTokenProvider` with default implementation that uses `UUID` to generate one time tokens, I can prepare a draft PR.

evgeniycheban:
Yes, that would make sense to add it to `GenerateOneTimeTokenRequest`, I think `setTokenValueFactory` may accept a `Supplier<String>` or maybe it is worth introducing a dedicated interface for that.

evgeniycheban:
@franticticktick would you be able to submit a PR or I can take this one?

PR #16946, opened 2025-04-16
Add support one-time token value customization

franticticktick:
Closes gh-16939

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/ott/DefaultGenerateOneTimeTokenRequestResolver.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/ott/DefaultGenerateOneTimeTokenRequestResolver.java
@@ -17,6 +17,8 @@
 package org.springframework.security.web.authentication.ott;
 
 import java.time.Duration;
+import java.util.UUID;
+import java.util.function.Supplier;
 
 import jakarta.servlet.http.HttpServletRequest;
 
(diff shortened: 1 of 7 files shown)


web/src/main/java/org/springframework/security/web/authentication/ott/DefaultGenerateOneTimeTokenRequestResolver.java
```diff
@@ -37,13 +39,15 @@ public final class DefaultGenerateOneTimeTokenRequestResolver implements Generat
...
+	private Supplier<String> tokenValueFactory = () -> UUID.randomUUID().toString();
+
 	@Override
 	public GenerateOneTimeTokenRequest resolve(HttpServletRequest request) {
 		String username = request.getParameter("username");
 		if (!StringUtils.hasText(username)) {
 			return null;
 		}
-		return new GenerateOneTimeTokenRequest(username, this.expiresIn);
+		return new GenerateOneTimeTokenRequest(username, this.expiresIn, this.tokenValueFactory.get());
```
evgeniycheban, line 50: I think we should not generate token value here, since it breaks the semantic of `GenerateOneTimeTokenRequestResolver` which is:
"A strategy for resolving a `GenerateOneTimeTokenRequest` from the `HttpServletRequest`".

It might also be confusing for users who provide a custom `tokenValueFactory` and notice that their factory gets called before `OneTimeTokenService#generate` method call.

So I would consider passing `Supplier<String> tokenValueFactory` to `GenerateOneTimeTokenRequest` and call it in the `OneTimeTokenService#generate` method.

evgeniycheban (changes requested):
Hi, @franticticktick I've left one comment below with some thoughts regarding token value generation.

PR #16971, opened 2025-04-21
Fix IllegalArgumentException message for unknown Argon2 types

rntrp:
Parsing Argon2 params relies on `String[] parts = encodedHash.split("\\$")`. For a string beginning with a `$`, the first element of the resulting array will be empty. This is already addressed in the code, with the variable `int currentPart` being initialized with `1`. But the IllegalArgumentException message incorrectly refers to `part[0]`, which will be empty for a well-formed Argon2 string starting with `$`.

The fix is trivial in this case, just use the array index 1.

P.S. one could also check for the 0-th array element to be an empty string and throw an IllegalArgumentException if it's not. However, this requires additional domain expertise and as such goes beyond the scope.


jzheaux:
Hi, @rntrp, thanks for the fix. Will you please add a unit test that confirms the behavior? It should be a test that fails before your fix and passes after the fix is applied.

rntrp:
Hello @jzheaux, thanks for the advice. I think it should be sufficient here to extend the existing test case. AssertJ failure messages are concise enough to see whether the right exception wasn't thrown or the message didn't contain the string. Please review the new commit.

jzheaux:
Thanks, @rntrp ! This is now merged into `main`.

PR #16976, opened 2025-04-21
Example test showing the minimalist response to an empty Authorizatio…

danielshiplett:
…n header

relates to #16977 

We recently had an application request fail to provide any bearer token in the Authorization header.  They received a 401 as expected.  However, the `www-authenticate` response header only had the word `Bearer` in the value.  This made it difficult to debug as it implied that they were sending a bearer token when they weren't.

I suggest an improvement in Spring Security to add an error message such as `missing_token` or some other similar value to help identify the root problem.

This PR contains a test case that will pass when the additional error message is provided.

I am willing to provide a fix for this as well.  However, I don't know where in Spring Security to implement such a fix.  I know about `BearerTokenAuthenticationEntryPoint` where other bearer token related errors have their details added.  However, that doesn't feel right in this case because we aren't 100% guaranteed the intent of the request was to provide a bearer token (some other form of Authorization may be supported along with OAuth2 Resource Server).

### DIFF
--- a/config/src/test/java/org/springframework/security/config/web/server/OAuth2ResourceServerSpecTests.java
+++ b/config/src/test/java/org/springframework/security/config/web/server/OAuth2ResourceServerSpecTests.java
@@ -179,6 +179,19 @@ public class OAuth2ResourceServerSpecTests {
 	}
 
 	@Test
+	public void getWhenBearerMissingInvalidToken() {
+		this.spring.register(PublicKeyConfig.class).autowire();
+		// @formatter:off
+		this.client.get()
+				.headers((headers) -> headers
+						.set("Authorization", ""))
+				.exchange()
+				.expectStatus().isUnauthorized()
+				.expectHeader().value(HttpHeaders.WWW_AUTHENTICATE, startsWith("Bearer error=\"missing_token\""));
+		// @formatter:on
+	}
+
+	@Test
 	public void getWhenUnsignedThenReturnsInvalidToken() {
 		this.spring.register(PublicKeyConfig.class).autowire();
 		// @formatter:off


jzheaux:
Thanks for the idea, @danielshiplett, though this deviates from the RFC [which states](https://datatracker.ietf.org/doc/html/rfc6750#section-3.1) (emphasis *mine*):

> *If the request lacks any authentication information* (e.g., the client
   was unaware that authentication is necessary or attempted using an
   unsupported authentication method), *the resource server SHOULD NOT
   include an error code or other error information*.

As such, I'm going to close this PR. Note that you can publish your own entry point that does as you describe:

```java
class MyAuthenticationEntryPoint implements AuthenticationEntryPoint {
    private final BearerAuthenticationEntryPoint delegate = new BearerAuthenticationEntryPoint();

    @Override 
    public void commence(HttpServletRequest request, HttpServletResponse response, AuthenticationException ex) {
        if (ex instanceof InsufficientAuthenticationException ex) {
            this.delegate.commence(request, response, new OAuth2AuthenticationException("missing token"));
            return;
        }
        this.delegate.commence(request, response, ex);
    }
}
```

danielshiplett:
Hey Josh.  Great feedback.  Thanks for the callout on the RFC and the example.  I take it, like most of the Beans, if I create my own `MyAuthenticationEntryPoint` Bean, the standard one(s) will back off?

jzheaux:
That is not the case with `AuthenticationEntryPoint` since each authentication type often has a different entry point (for example, the entry point for form login is redirecting to a login page).

If you give `.oauth2ResourceServer` the entry point like so:

```java
http
    .oauth2ResourceServer((oauth2) -> oauth2.authenticationEntryPoint(entryPoint))
```

then it will get applied both to `BearerTokenAuthenticationFilter` (where the error is occurring for you) as well as in `ExceptionTranslationFilter` for post-authentication failures like insufficient scopes.

Issue #16980, opened 2025-04-22
SubjectDnX509PrincipalExtractor should update to getSubjectX500Principal


jzheaux:
Recent versions of Java state that `getSubjectDn` should no longer be used:

```java
/**
  * @deprecated Use {@link #getSubjectX500Principal} instead. This method
  * returns the {@code subject} as an implementation specific
  * {@code Principal} object, which should not be relied upon by portable
  * code.
  */
```

However, since `getSubjectDn` is abstract, applications may be relying on provider implementations. For example, Bouncycastle returns an `org.bouncycastle.jce.X509Principal` instance for `getSubjectDn` and a `javax.security.auth.x500.X500Principal` for `getSubjectX500Principal`.

For this reason, we should add a toggle for this value, something like:

```java
boolean extractPrincipalNameFromX500Principal = false;
```

This default would change to `true` in Spring Security 8.

franticticktick:
Hi @jzheaux , could you assign this ticket to me please?

jzheaux:
Closed in favor of https://github.com/spring-projects/spring-security/pull/16984

PR #16984, opened 2025-04-23
Add Support Extracting DN From X500Principal

franticticktick:
Closes gh-16980

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/preauth/x509/SubjectDnX509PrincipalExtractor.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/preauth/x509/SubjectDnX509PrincipalExtractor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2020 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -43,6 +43,7 @@ import org.springframework.util.Assert;
  * "EMAILADDRESS=jimi@hendrix.org, CN=..." giving a user name "jimi@hendrix.org"
  *
  * @author Luke Taylor
+ * @author Max Batischev
  */
 public class SubjectDnX509PrincipalExtractor implements X509PrincipalExtractor, MessageSourceAware {
 
@@ -52,14 +53,16 @@ public class SubjectDnX509PrincipalExtractor implements X509PrincipalExtractor, 
 
 	private Pattern subjectDnPattern;
 
+	private boolean extractPrincipalNameFromX500Principal = false;
+
 	public SubjectDnX509PrincipalExtractor() {
 		setSubjectDnRegex("CN=(.*?)(?:,|$)");
 	}
 
 	@Override
 	public Object extractPrincipal(X509Certificate clientCert) {
-		// String subjectDN = clientCert.getSubjectX500Principal().getName();
-		String subjectDN = clientCert.getSubjectDN().getName();
+		String subjectDN = this.extractPrincipalNameFromX500Principal ? clientCert.getSubjectX500Principal().getName()
+				: clientCert.getSubjectDN().getName();
 		this.logger.debug(LogMessage.format("Subject DN is '%s'", subjectDN));
 		Matcher matcher = this.subjectDnPattern.matcher(subjectDN);
 		if (!matcher.find()) {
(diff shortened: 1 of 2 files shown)


rwinch:
Thanks for the PR @franticticktick 

Upon review, I think that it makes more sense to deprecate and create a `SubjectX500PrincipalExtractor` class that extracts using `getSubjectX500Principal`.

What are your thoughts @jzheaux ?

franticticktick:
@rwinch @jzheaux there is one problem I would like to discuss. By default, the `getName()` method of the `X500Principal` class uses the [RFC2253](https://datatracker.ietf.org/doc/html/rfc2253) format. That is, in tests, the `dn` will be generated:
```
1.2.840.113549.1.9.1=#16126c756b65406d6f6e6b65796d616368696e65,CN=Luke Taylor,OU=Open Source Development Lab.,O=Monkey Machine Ltd,L=Glasgow,ST=Scotland,C=UK
```
We see an oid instead of an `emailaddress`, which is in line with the format. At the same time, we see an encoded string instead of a value. As the specification says:
```
If the AttributeValue is of a type which does not have a string
representation defined for it, then it is simply encoded as an
octothorpe character ('#' ASCII 35) followed by the hexadecimal
representation of each of the bytes of the BER encoding of the X.500
AttributeValue.  This form SHOULD be used if the AttributeType is of
the dotted-decimal form.
```
To get a text string instead of an encoded value, we need to explicitly specify the format, this is [RFC1779](https://datatracker.ietf.org/doc/html/rfc1779).
```
String subjectDN = clientCert.getSubjectX500Principal().getName("RFC1779");
```
In this case, the generated `dn` will look like this:
```
OID.1.2.840.113549.1.9.1=luke@monkeymachine, CN=Luke Taylor, OU=Open Source Development Lab., O=Monkey Machine Ltd, L=Glasgow, ST=Scotland, C=UK
```
In this format we can get email, but no matter what format we specify, we will still get oid in the string. Even if we leave the dn getting by pattern, it will be very non-obvious. From example:
```
extractor.setFormat("RFC1779");
extractor.setSubjectDnRegex("OID.1.2.840.113549.1.9.1=(.*?),");
```
This is a pretty hard to understand API.  At the same time, the `toString()` gives this result:
```
EMAILADDRESS=luke@monkeymachine, CN=Luke Taylor, OU=Open Source Development Lab., O=Monkey Machine Ltd, L=Glasgow, ST=Scotland, C=UK
```
And this will give the same result as the current implementation. If you have any ideas on how to better implement this feature, I'm ready to listen.

rwinch:
@franticticktick Thanks for reaching out.

I do not think that we can use the toString method because it is implementation dependent. I think that we should probably allow for the format and regex to be injected. We can have an additional boolean property for extracting the email address that in turn sets the format and regex to extract the email address. This will simplify the usage for users that do not want to dig into the details of how things work under the covers. Does that work?

franticticktick:
@rwinch @jzheaux Could you please review this PR? I hope I haven't missed anything.

rwinch (changes requested):
Thanks for the changes @franticticktick! I think this is getting close. I'd like to preserve the idea of having a regex for extracting the principal name and allowing the user to specify the format. The difference would be that the extractPrincipalNameFromEmail would set the correct format + the correct regex without needing to understand how it works.

I'd argue that we probably don't need any properties on the DSLs but instead just allow the user to set the properties on SubjectX500PrincipalExtractor and then the extractor on the DSL (or exposed as a Bean...however it works now)

rwinch:
@franticticktick I'm not sure if we need to expose additional properties on the configurer. Instead we could allow users to specify the `SubjectX500PrincipalExtractor` with the default being what we think is the most common scenario for users. Thoughts?

franticticktick:
@rwinch I am afraid that the user may not get the result from the regexp that he wants because of the wrong format. Most likely he will try to find this property in the configurer and will be confused when he does not find it. We can make the first version without the format in the configurer and look at the feedback of users. If there are many complaints about the lack of a format, we can always add it.

In addition, I will open a separate ticket for the feature of initializing `X509PrincipalExtractor` as a bean.

rwinch:
@franticticktick I think that is a valid concern. I should have said that we should consider deprecating the other properties in another ticket. What are your thoughts?

rwinch:
Thanks for the update. I've pushed a few changes and this will be merged as soon as the build passes. You can see the commits for details, but a summary of the main changes are:

- Encapsulate the extractPrincipalNameFromEmail property
- Configuration uses a principalExtractor reference rather than adding new properties
- Update the reference documentation

PR #16998, opened 2025-04-25
Roles and authorities are being overridden in GrantedAuthority and are not accessible within the AuthenticationManager.

surajbh123:
**Scenario 1**

```java
 UserDetails user = User.builder()
                .username("user")
                .password(passwordEncoder.encode("user"))
// first adding authorities and then role
                .authorities("read", "item:view") 
// authorities getting overridden by role
                .roles("USER")
                .accountExpired(false)
                .accountLocked(false)
                .credentialsExpired(false)
                .disabled(false)
                .build();
```
authorities = ["ROLE_USER"]

**Scenario 2**
```java
 UserDetails user = User.builder()
                .username("user")
                .password(passwordEncoder.encode("user"))
// first adding role and then authorities
                 .roles("USER")
// role getting overridden by authorities
                .authorities("read", "item:view") 
                .accountExpired(false)
                .accountLocked(false)
                .credentialsExpired(false)
                .disabled(false)
                .build();
```
authorities = ["read", "item:view"]

### DIFF
--- a/core/src/main/java/org/springframework/security/core/userdetails/User.java
+++ b/core/src/main/java/org/springframework/security/core/userdetails/User.java
@@ -441,7 +441,7 @@ public class User implements UserDetails, CredentialsContainer {
 		 */
 		public UserBuilder authorities(Collection<? extends GrantedAuthority> authorities) {
 			Assert.notNull(authorities, "authorities cannot be null");
-			this.authorities = new ArrayList<>(authorities);
+			this.authorities.addAll(authorities);
 			return this;
 		}
 


core/src/main/java/org/springframework/security/core/userdetails/User.java
```diff
@@ -441,7 +441,7 @@ public UserBuilder authorities(GrantedAuthority... authorities) {
 		 */
 		public UserBuilder authorities(Collection<? extends GrantedAuthority> authorities) {
 			Assert.notNull(authorities, "authorities cannot be null");
-			this.authorities = new ArrayList<>(authorities);
```
surajbh123, line 444: A NullPointerException is not possible in this case since the authorities list is explicitly initialized at line 332 using new ArrayList<>().
```java
		private List<GrantedAuthority> authorities = new ArrayList<>();
```

PR #17010, opened 2025-04-28
Remove duplicate lines from X.509 documentation

### LINKED ISSUE #16701: Repeated text in servlet X.509 authentication docs
**Describe the bug**
Two of the paragraphs in the servlet X.509 authentication documentation contain the same three sentences:

> **You can also use SSL with “mutual authentication”. The server then requests a valid certificate from the client as part of the SSL handshake. The server authenticates the client by checking that its certificate is signed by an acceptable authority.** If a valid certificate has been provided, it can be obtained through the servlet API in an application. The Spring Security X.509 module extracts the certificate by using a filter. It maps the certificate to an application user and loads that user’s set of granted authorities for use with the standard Spring Security infrastructure.
>
> **You can also use SSL with “mutual authentication”. The server then requests a valid certificate from the client as part of the SSL handshake. The server authenticates the client by checking that its certificate is signed by an acceptable authority.** For example, if you use Tomcat, you should read the [Tomcat SSL instructions](https://tomcat.apache.org/tomcat-10.1-doc/ssl-howto.html). You should get this working before trying it out with Spring Security.

**To Reproduce**
See https://docs.spring.io/spring-security/reference/servlet/authentication/x509.html

**Expected behavior**
Remove the duplicated sentences, perhaps merging the two paragraphs if it makes sense to do so.

**Sample**
N/A

ronodhirSoumik:
Fixes #16701

### DIFF
--- a/docs/modules/ROOT/pages/servlet/authentication/x509.adoc
+++ b/docs/modules/ROOT/pages/servlet/authentication/x509.adoc
@@ -8,13 +8,11 @@ The browser automatically checks that the certificate presented by a server has 
 You can also use SSL with "`mutual authentication`". The server then requests a valid certificate from the client as part of the SSL handshake.
 The server authenticates the client by checking that its certificate is signed by an acceptable authority.
 If a valid certificate has been provided, it can be obtained through the servlet API in an application.
-The Spring Security X.509 module extracts the certificate by using a filter.
-It maps the certificate to an application user and loads that user's set of granted authorities for use with the standard Spring Security infrastructure.
-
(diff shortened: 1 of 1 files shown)


jzheaux:
Good catch, @ronodhirSoumik! Can you please [sign-off your commit](https://github.com/apps/dco) so that we can merge it?

ronodhirSoumik:
sure.. I will do that

ronodhirSoumik:
> Good catch, @ronodhirSoumik! Can you please [sign-off your commit](https://github.com/apps/dco) so that we can merge it?

Done

jzheaux:
Thanks, @ronodhirSoumik! This is now merged into `6.3.x`, `6.4.x`, and `main`.

PR #17021, opened 2025-04-30
LDAP: allow multiple roles to be fetched from group role attribute

fer-marino:
For some weird reasons, only the first element of the list returned by the LDAP query is used to create authorities. This PR fixes it creating an authority for all the returned roles.


jzheaux:
@fer-marino, thanks for contributing. I don't think we should go further down the route of using low-level types, but instead would be better off using `DirContextOperations` directly. This allows the application to be more expressive without needing to know how to navigate Spring Security's special set of keys like `DN_KEY`.

I've added https://github.com/spring-projects/spring-security/issues/17028 to track this work. In the meantime, have you tried using `NestedLdapAuthoritiesPopulator` with a max search depth of 1? I believe that implementation adds all the roles by default.

fer-marino:
thank you for your feedback. I'll close this and open a new one with your recommendations

Issue #17004, opened 2025-04-27
JOSE Validators claim mandatoriness option


FerencKemeny:
**Expected Behavior**

I would like to build a control in ```JwtTimestampValidator```, ```JwtIssuerValidator``` and ```JwtAudienceValidator``` that mandatoriness of the specific JWT clause is specified. So that a ```required``` parameter can specify the same way like it is done in JwtIssuedAtValidator.

**Current Behavior**

The above mentioned validators are simply falling through with successful check when the given clause is missing form JWT. This may be a misleading behavior because validator is created for a reason. However it is understandable the claims are optional by the specification - in general.

**Context**

I would like to create the alternative that the implementor could control if the fields must be mandatory. Currently this could be achieved by adding multiple validators. It is more elegant to specify if the given validator requires the claim and make the validation fail if the claim is missing. So this way more strict and rigorous control could be built.

jzheaux:
Note that `JwtTypValidator` has a method `setAllowEmpty` that may be of interest. This could be added to other validators. For backward compatibility in other vaildators, `allowEmpty` would be `true` by default.

FerencKemeny:
> This could be added to other validators.

So do you mean you would like to see ```setAllowEmpty``` method implemented the same way as it is in ```JwtTypeValidator```, instead of contructor parameters? It is true, number of ```JwtTimestampValidator``` constructors increased a bit this way. ```setAllowEmpty``` method may make it simpler.

I think this is more like code cleanness but backward compatibility. I did not change the original contructor signatures and their function.

Let me change as you suggested.

jzheaux:
Closing in favor of https://github.com/spring-projects/spring-security/pull/17030

PR #17030, opened 2025-05-02
JwtTimestampsValidator can require exp and nbf claims

FerencKemeny:
I implemented ```required``` parameter in ```JwtTimestampValidator```, ```JwtIssuerValidator``` and ```JwtAudienceValidator```. I left the function of the original constructors untouched. In the original implementation successful validation was returned even if timestamps, issues or audience claims were missing. So this way previous API, implementations are not breaking. With my changes it is possible now to specify more strict specification, to tell if the given claim is mandatory to present and indiate it with failed validation if the claim is missing.

This is the feature indicated in #17004

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtTimestampValidator.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtTimestampValidator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2018 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -29,6 +29,7 @@ import org.springframework.security.oauth2.core.OAuth2ErrorCodes;
 import org.springframework.security.oauth2.core.OAuth2TokenValidator;
 import org.springframework.security.oauth2.core.OAuth2TokenValidatorResult;
 import org.springframework.util.Assert;
+import org.springframework.util.ObjectUtils;
 
 /**
  * An implementation of {@link OAuth2TokenValidator} for verifying claims in a Jwt-based
@@ -54,6 +55,10 @@ public final class JwtTimestampValidator implements OAuth2TokenValidator<Jwt> {
 
 	private final Duration clockSkew;
 
+	private boolean allowEmptyExpiryClaim = true;
+
+	private boolean allowEmptyNotBeforeClaim = true;
+
 	private Clock clock = Clock.systemUTC();
 
 	/**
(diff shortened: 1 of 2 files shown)


oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtAudienceValidator.java
```diff
@@ -32,19 +32,30 @@ public final class JwtAudienceValidator implements OAuth2TokenValidator<Jwt> {
 
 	private final JwtClaimValidator<Collection<String>> validator;
 
+	private boolean allowEmpty;
```
jzheaux, line 35: Having thought about this further, I think we should wait on adding this toggle to validators that require the claim by default. We can relax it when a clear need arises.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtIssuerValidator.java
```diff
@@ -32,20 +30,30 @@ public final class JwtIssuerValidator implements OAuth2TokenValidator<Jwt> {
 
 	private final JwtClaimValidator<Object> validator;
 
+	private boolean allowEmpty;
```
jzheaux, line 33: Having thought about this further, I think we should wait on adding this toggle to validators that require the claim by default. We can relax it when a clear need arises.

Other than the error message cleanup you added in `validate`, let's please not make further functional changes to this class in this PR.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtTimestampValidator.java
```diff
@@ -65,20 +67,33 @@ public JwtTimestampValidator() {
...
 		Assert.notNull(clockSkew, "clockSkew cannot be null");
+		this.allowEmpty = true;
 		this.clockSkew = clockSkew;
 	}
 
+	/**
+	 * Whether to allow the {@code exp} or {@code nbf} header to be empty. The default value is
+	 * {@code true}
+	 */
+	public void setAllowEmpty(boolean allowEmpty) {
```
jzheaux, line 78: I think this needs to be more flexible, given that there are two different claims. It seems reasonable to say that `exp` is required while `nbf` is not.

Will you please change this to be `setAllowEmptyExpiryClaim` and `setAllowEmptyNotBeforeClaim`? These would each have their own toggle and be `true` by default.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtTimestampValidator.java
```diff
@@ -65,20 +67,33 @@ public JwtTimestampValidator() {
 
 	public JwtTimestampValidator(Duration clockSkew) {
 		Assert.notNull(clockSkew, "clockSkew cannot be null");
+		this.allowEmpty = true;
 		this.clockSkew = clockSkew;
 	}
 
+	/**
+	 * Whether to allow the {@code exp} or {@code nbf} header to be empty. The default value is
+	 * {@code true}
```
jzheaux, line 76: On new public methods, please add `@since 7.0`

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtTimestampValidator.java
```diff
@@ -52,6 +52,8 @@ public final class JwtTimestampValidator implements OAuth2TokenValidator<Jwt> {
 
 	private static final Duration DEFAULT_MAX_CLOCK_SKEW = Duration.of(60, ChronoUnit.SECONDS);
 
+	private boolean allowEmpty;
```
jzheaux, line 55: Please set the default value here instead of in the constructor. Also, please declare it after the `final` fields.

jzheaux, line 55: Will you please ensure that this file's copyright changes to `2025`?

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/JwtTimestampValidatorTests.java
```diff
@@ -128,30 +128,77 @@ public void validateWhenConfiguredWithFixedClockThenValidatesUsingFixedTime() {
 	}
 
 	@Test
-	public void validateWhenNeitherExpiryNorNotBeforeIsSpecifiedThenReturnsSuccessfulResult() {
```
jzheaux, line 131: Will you please ensure that this file's copyright changes to `2025`?

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/JwtIssuerValidatorTests.java
```diff
@@ -86,16 +130,37 @@ public void validateWhenIssuerMatchesAndIsNotAUriThenReturnsSuccess() {
...
+		JwtIssuerValidator validator = new JwtIssuerValidator("issuer");
+		validator.setAllowEmpty(false);
+		// @formatter:off
+		assertThat(validator.validate(jwt))
+				.isEqualTo(OAuth2TokenValidatorResult.success());
+		// @formatter:on
+	}
+
+	@Test
+	public void validateWhenAllowEmptyDefaultAndJwtIsNullThenThrowsIllegalArgumentException() {
```
jzheaux, line 146: Thank you for these additional tests. Will you please place any tests or other changes that are unrelated to the feature in separate commits so that each commit is atomic?

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtIssuerValidator.java
```diff
@@ -32,20 +30,30 @@ public final class JwtIssuerValidator implements OAuth2TokenValidator<Jwt> {
...
+	 * {@code false}
+	 */
+	public void setAllowEmpty(boolean allowEmpty) {
+		this.allowEmpty = allowEmpty;
 	}
 
 	@Override
 	public OAuth2TokenValidatorResult validate(Jwt token) {
-		Assert.notNull(token, "token cannot be null");
+		Assert.notNull(token, "jwt cannot be null");
```
jzheaux, line 56: Thank you for this additional cleanup. Will you please place this is in a previous commit?

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtIssuerValidator.java
```diff
@@ -32,20 +30,30 @@ public final class JwtIssuerValidator implements OAuth2TokenValidator<Jwt> {
...
+	 * Whether to allow the {@code iss} claim to be empty. The default value is
+	 * {@code false}
+	 */
+	public void setAllowEmpty(boolean allowEmpty) {
+		this.allowEmpty = allowEmpty;
 	}
 
 	@Override
 	public OAuth2TokenValidatorResult validate(Jwt token) {
-		Assert.notNull(token, "token cannot be null");
```
jzheaux, line 48: I think the intent is to say the parameter name, not the datatype. Will you please leave this as-is?

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/JwtAudienceValidator.java
```diff
@@ -32,19 +32,30 @@ public final class JwtAudienceValidator implements OAuth2TokenValidator<Jwt> {
...
+	 * Whether to allow the {@code aud} claim to be empty. The default value is
+	 * {@code false}
+	 */
+	public void setAllowEmpty(boolean allowEmpty) {
+		this.allowEmpty = allowEmpty;
 	}
 
 	@Override
 	public OAuth2TokenValidatorResult validate(Jwt token) {
-		Assert.notNull(token, "token cannot be null");
```
jzheaux, line 47: I think the intent is to say the parameter name, not the datatype. Will you please leave this as-is?

jzheaux (changes requested):
Thanks, @FerencKemeny! I've leave my feedback inline.

jzheaux (approved):
Thanks, @FerencKemeny, for the PR! We'll merge this shortly after we've wrapped up the 6.5 release.

FerencKemeny:
Thank you, @jzheaux, for all of your help on the PR. Appreciated!

jzheaux (approved):

jzheaux:
Thanks, @FerencKemeny! This has been merged into `main`.

FerencKemeny:
I could not figure out why my commits appear only "Partially verified" and "Unverified" after you've merged. Anyhow I hope I can do this part better next time.

PR #17033, opened 2025-05-02
NimbusJwtEncoder should simplify constructing with javax.security Keys 

surajbh123:


surajbh123:
Working on test cases and code reformatting

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -102,6 +115,12 @@ public final class NimbusJwtEncoder implements JwtEncoder {
 	public NimbusJwtEncoder(JWKSource<SecurityContext> jwkSource) {
 		Assert.notNull(jwkSource, "jwkSource cannot be null");
 		this.jwkSource = jwkSource;
+		this.DEFAULT_JWS_HEADER = JwsHeader.with(SignatureAlgorithm.RS256).build();
+	}
+	public NimbusJwtEncoder(JWKSource<SecurityContext> jwkSource, JwsHeader jwsHeader) {
```
jzheaux, line 120: Please replace this constructor with a setDefaultJwsHeader method. This is because there is already a reasonable default for `defaultJwsHeader` and we prefer to manage non-required members in setter methods.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -369,4 +388,295 @@ private static URI convertAsURI(String header, URL url) {
...
+		/**
+		 * Configures the {@link Converter} used to select the JWK when multiple keys match
+		 * the header criteria. This is generally not needed for single-key setups but is
+		 * provided for consistency.
+		 *
+		 * @param jwkSelector the {@link Converter} to select a {@link JWK}
+		 * @return this builder instance for method chaining
+		 * @since // TODO: Update version
+		 */
+		public SecretKeyJwtEncoderBuilder jwkSelector(Converter<List<JWK>, JWK> jwkSelector) {
```
jzheaux, line 478: Given that the jwkSelector is not a required parameter, let's leave it out of the builder. Applications can set it once construction is complete.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -369,4 +388,295 @@ private static URI convertAsURI(String header, URL url) {
...
+
+		private final SecretKey secretKey;
+
+		private String keyId;
+
+		private JWSAlgorithm jwsAlgorithm = JWSAlgorithm.HS256;
+
+		private Converter<List<JWK>, JWK> jwkSelector;
+
+		private JwsHeader jwsHeader;
```
jzheaux, line 438: This member variable doesn't seem to be needed since it's only ever used in the `build()` method.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -369,4 +388,295 @@ private static URI convertAsURI(String header, URL url) {
...
+					() -> "The algorithm '" + this.jwsAlgorithm + "' is not compatible with a SecretKey. "
+							+ "Please use one of the HS256, HS384, or HS512 algorithms.");
+
+			OctetSequenceKey.Builder builder = new OctetSequenceKey.Builder(this.secretKey)
+					.keyUse(KeyUse.SIGNATURE)
+					.algorithm(this.jwsAlgorithm);
+
+			this.jwsHeader = JwsHeader.with(MacAlgorithm.from(this.jwsAlgorithm.getName())).build();
+
+			if (StringUtils.hasText(this.keyId)) {
```
jzheaux, line 504: I believe this null check is not needed since Nimbus allows for null. That is, Spring Security's `keyId` should follow the same semantics as Nimbus's `keyID`.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -369,4 +388,295 @@ private static URI convertAsURI(String header, URL url) {
...
+				builder.keyID(this.keyId);
+			}
+
+			OctetSequenceKey jwk = builder.build();
+			JWKSource<SecurityContext> jwkSource = new ImmutableJWKSet<>(new JWKSet(jwk));
+			NimbusJwtEncoder encoder = this.jwsHeader  != null
```
jzheaux, line 510: It doesn't seem like `jwsHeader` could ever be null. Are you able to remove this check?

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -369,4 +388,295 @@ private static URI convertAsURI(String header, URL url) {
...
+			return encoder;
+		}
+
+	}
+
+	/**
+	 * A builder for creating {@link NimbusJwtEncoder} instances configured with a
+	 * {@link KeyPair}.
+	 *
+	 * @since // TODO: Update version
```
jzheaux, line 527: This will likely be "7.0"

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -369,4 +388,295 @@ private static URI convertAsURI(String header, URL url) {
...
+
+		private String keyId;
+
+		private JWSAlgorithm jwsAlgorithm;
+
+		private Converter<List<JWK>, JWK> jwkSelector;
+
+		private JwsHeader jwsHeader;
+
+		private static boolean isEcAlgorithm(SignatureAlgorithm algorithm) {
```
jzheaux, line 541: This method appears to be unused

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -369,4 +388,295 @@ private static URI convertAsURI(String header, URL url) {
...
+					new NimbusJwtEncoder(jwkSource, this.jwsHeader);
+			if (this.jwkSelector != null) {
+				encoder.setJwkSelector(this.jwkSelector);
+			} else {
+				encoder.setJwkSelector(jwkSet -> jwkSet.stream().findFirst().orElse(null));
+			}
+			return encoder;
+		}
+
+		private JWK buildJwk() {
```
jzheaux, line 614: I believe if you rearrange the logic in the following way, there will be less nesting and fewer member variables:

1. Compute the `JwsHeader` in a separate method, say `JwsHeader buildDefaultJwsHeader(Jwk jwk)`
2. In `build` do something like:

```java
JWK jwk = buildJwk();
JwsHeader defaultJwsHeader = buildDefaultJwsHeader(jwk);
// ...
```

Doing that will allow this method to become:

```java
if (this.keyPair.getPrivate() instanceof RSAPrivateKey) {
    return buildRsaJwk();
}
if (this.keyPair.getPrivate() instanceof ECPrivateKey) {
    return buildEcJwk();
}
throw new ...
```

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -369,4 +388,295 @@ private static URI convertAsURI(String header, URL url) {
...
+						.keyId(ecKey.getKeyID())
+						.build();
+				return ecKey;
+			} else {
+				throw new IllegalStateException("Unsupported key pair type: " + this.keyPair.getPrivate().getClass().getName());
+			}
+		}
+
+		private RSAKey buildRsaJwk() {
+			if (this.jwsAlgorithm == null) {
```
jzheaux, line 633: Note that if the signature algorithm is verified in the `signatureAlgorithm` method, then this check and the `Assert#state` check are unneeded here.

jzheaux (changes requested):
Thanks, @surajbh123! This PR will make using `NimbusJwtEncoder` much simpler.

I've left my feedback inline. In addition to that, please ensure that your commit only has the signoff in the description and not in the title. Please also ensure that the description has the issue that this PR resolves:

```
Add NimbusJwtEncoder Builders

Closes gh-16267

Signed-off-by ...
```

jzheaux:
Just one more thing, @surajbh123, when you submit, it will save you time in the review process if you run the following before submitting:

```
./gradlew format && ./gradlew :spring-security-oauth2-jose:check
```

surajbh123:
Working on reviews comments 
Not resolved all yet

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
surajbh123: done , removed

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -79,7 +92,7 @@ public final class NimbusJwtEncoder implements JwtEncoder {
 
 	private static final String ENCODING_ERROR_MESSAGE_TEMPLATE = "An error occurred while attempting to encode the Jwt: %s";
 
-	private static final JwsHeader DEFAULT_JWS_HEADER = JwsHeader.with(SignatureAlgorithm.RS256).build();
+	private static JwsHeader DEFAULT_JWS_HEADER;
```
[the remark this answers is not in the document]
surajbh123: noted

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
surajbh123: for `Key Pair algorithm` We cannot pass null `keyId`  for other it can be `null`

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -369,4 +388,295 @@ private static URI convertAsURI(String header, URL url) {
...
+		 * Sets the JWS algorithm to use for signing. Must be compatible with the key type
+		 * (RSA or EC). If not set, a default algorithm will be chosen based on the key
+		 * type (e.g., RS256 for RSA, ES256 for EC).
+		 *
+		 * @param signatureAlgorithm the {@link SignatureAlgorithm} to use
+		 * @return this builder instance for method chaining
+		 */
+		public KeyPairJwtEncoderBuilder signatureAlgorithm(SignatureAlgorithm signatureAlgorithm) {
+			Assert.notNull(signatureAlgorithm, "signatureAlgorithm cannot be null");
+			this.jwsAlgorithm = JWSAlgorithm.parse(signatureAlgorithm.getName());
```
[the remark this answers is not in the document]
surajbh123: can,t do now as as now we have two separate builder for RSA and EC

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
surajbh123: done

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
surajbh123: done

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -369,4 +388,295 @@ private static URI convertAsURI(String header, URL url) {
...
+		Assert.isTrue(keyPair.getPrivate() instanceof java.security.interfaces.RSAKey || keyPair.getPrivate() instanceof java.security.interfaces.ECKey,
+				"keyPair must be an RSAKey or an ECKey");
+		return new KeyPairJwtEncoderBuilder(keyPair);
+	}
+
+	/**
+	 * A builder for creating {@link NimbusJwtEncoder} instances configured with a
+	 * {@link SecretKey}.
+	 *
+	 * @since // TODO: Update version
```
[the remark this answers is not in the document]
surajbh123: done

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -369,4 +388,295 @@ private static URI convertAsURI(String header, URL url) {
...
+	 * @return a {@link KeyPairJwtEncoderBuilder} for further configuration
+	 * @since // TODO: Update version
+	 */
+	public static KeyPairJwtEncoderBuilder withKeyPair(KeyPair keyPair) {
+		Assert.notNull(keyPair, "keyPair cannot be null");
+		Assert.notNull(keyPair.getPrivate(), "keyPair must contain a private key");
+		Assert.notNull(keyPair.getPublic(), "keyPair must contain a public key");
+		Assert.isTrue(keyPair.getPrivate() instanceof java.security.interfaces.RSAKey || keyPair.getPrivate() instanceof java.security.interfaces.ECKey,
+				"keyPair must be an RSAKey or an ECKey");
+		return new KeyPairJwtEncoderBuilder(keyPair);
```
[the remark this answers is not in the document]
surajbh123: Done , I have created two builders with parent is abstract static

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -119,14 +133,16 @@ public void setJwkSelector(Converter<List<JWK>, JWK> jwkSelector) {
 		this.jwkSelector = jwkSelector;
 	}
 
+	public void setJwsHeader(JwsHeader jwsHeader) {
+		this.jwsHeader = jwsHeader;
```
jzheaux, line 137: Please check for `null` here instead of in `encode`. You can use `Assert.notNull` for this.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -369,4 +385,249 @@ private static URI convertAsURI(String header, URL url) {
...
+	 * {@link KeyPair}. The key pair must contain either an {@link RSAKey} or an
+	 * {@link ECKey}.
+	 * @param keyPair the {@link KeyPair} to use for signing JWTs
+	 * @return a {@link KeyPairJwtEncoderBuilder} for further configuration
+	 * @since 7.0
+	 */
+	public static KeyPairJwtEncoderBuilder withKeyPair(KeyPair keyPair) {
+		Assert.isTrue(keyPair != null && keyPair.getPrivate() != null && keyPair.getPublic() != null,
+				"keyPair, its private key, and public key must not be null");
+		Assert.isTrue(
```
jzheaux, line 411: I think this is no longer necessary since you are performing these checks below now.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -369,4 +385,249 @@ private static URI convertAsURI(String header, URL url) {
...
+		}
+
+		/**
+		 * Builds the {@link NimbusJwtEncoder} instance.
+		 * @return the configured {@link NimbusJwtEncoder}
+		 * @throws IllegalStateException if the configured JWS algorithm is not compatible
+		 * with a {@link SecretKey}.
+		 */
+		public NimbusJwtEncoder build() {
+			this.jwsAlgorithm = (this.jwsAlgorithm != null) ? this.jwsAlgorithm : JWSAlgorithm.HS256;
```
jzheaux, line 477: Since you are giving this a default value and there isn't a way for the application to set it back to `null`, I believe you can remove this line.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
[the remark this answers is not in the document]
jzheaux: please push this down to the subclass or do not use the abstract class so that you can validate here. From a security perspective, it's valuable to validate as early as possible.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -369,4 +385,249 @@ private static URI convertAsURI(String header, URL url) {
...
+
+	}
+
+	/**
+	 * A builder for creating {@link NimbusJwtEncoder} instances configured with a
+	 * {@link KeyPair}.
+	 *
+	 * @since 7.0
+	 */
+	public abstract static class KeyPairJwtEncoderBuilder {
```
jzheaux, line 498: I think it might be best to not have an abstract class since that can be quickly complex for builders. As it stands, the amount of reuse is quite small. However, if this is kept, making it package-private is prefered so it's clear that it's not part of the public API for applications to make their own extensions.

jzheaux (changes requested):
Thanks for the quick turnaround, @surajbh123! This is coming together nicely. I've added some additional inline feedback.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
surajbh123: ok

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
surajbh123: In the following case, it may be null.

        SecretKey defaultSecretKey = createSecretKey("HmacSHA256", 256); // Key for default HS256
        NimbusJwtEncoder encoderWithSecretDefaults = NimbusJwtEncoder.withSecretKey(defaultSecretKey)
                // .macAlgorithm(MacAlgorithm.HS256) // Default is HS256
                // .keyId(null)                     // Default keyId is null
                // .jwkSelector(...)               // Default selector throws if multiple keys match
                .build();

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
surajbh123: done , u r right

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
surajbh123: it can never be null here
as we setting default value if jws header is null i.e. not sent by user

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
surajbh123: Hey @jzheaux 
Running ./gradlew :spring-security-oauth2-jose:check is raising a warning that says:
`
Class KeyPairJwtEncoderBuilder should be declared as final. |  
 `
So I think we should keep the class abstract to satisfy that rule.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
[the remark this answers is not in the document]
surajbh123: done

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoderTests.java
```diff
@@ -344,6 +353,187 @@ public void encodeWhenNoKeysThenJwkSelectorIsNotUsed() throws Exception {
...
+		assertThat(jwt.getHeaders().get("typ").toString()).isEqualTo("JWT");
+		assertThat(jwt.getHeaders().get("cty").toString()).isEqualTo("application/jwt");
+		assertThat(jwt.getHeaders().get("alg").toString()).isEqualTo("HS256");
+		assertThatNoException().isThrownBy(() -> jwt.getClaims());
+		assertJwt(jwt);
+	}
+
+	@Test
+	void keyPairBuilderWithRsaDefaultAlgorithm() throws Exception {
+		KeyPairGenerator keyPairGenerator = KeyPairGenerator.getInstance("RSA");
```
jzheaux, line 424: Instead of re-constructing the generator, please move this to a member variable. Also, I feel this will be simpler if you use `RSAKeyGenerator` from Nimbus instead:

```java
private final RSAKeyGenerator rsaGenerator = new RSAKeyGenerator(2048);
```

Then in this test you can do:

```java
RSAKey jwk = this.rsaGenerator.generate();
```

and then use that instance to construct the builder. The other unit tests would benefit from this same polish.

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoderTests.java
```diff
@@ -344,6 +353,187 @@ public void encodeWhenNoKeysThenJwkSelectorIsNotUsed() throws Exception {
...
+		assertThat(jwt).isNotNull();
+		assertThat(jwt.getHeaders().get("alg").toString()).isEqualTo("RS512");
+		assertThat(jwt.getSubject()).isEqualTo(claims.getSubject());
+		assertThatNoException().isThrownBy(() -> jwt.getClaims());
+		assertJwt(jwt);
+	}
+
+	@Test
+	void keyPairBuilderWithEcDefaultAlgorithm() throws Exception {
+		KeyPairGenerator keyPairGenerator = KeyPairGenerator.getInstance("EC");
```
jzheaux, line 461: Instead of re-constructing the generator, please move this to a member variable. Also, I feel this will be simpler if you use `ECKeyGenerator` from Nimbus instead:

```java
private final ECKeyGenerator ecGnerator = new ECKeyGenerator(Curve.P_256);
```

Then in this test you can do:

```java
ECKey jwk = this.ecGnerator.generate();
```

and then use that instance to construct the builder.

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoderTests.java
```diff
@@ -344,6 +353,187 @@ public void encodeWhenNoKeysThenJwkSelectorIsNotUsed() throws Exception {
...
+	}
+
+	@Test
+	void keyPairBuilderWithEcCustomAlgorithm() throws Exception {
+		KeyPairGenerator keyPairGenerator = KeyPairGenerator.getInstance("EC");
+		keyPairGenerator.initialize(256);
+		KeyPair keyPair = keyPairGenerator.generateKeyPair();
+		NimbusJwtEncoder encoder = NimbusJwtEncoder.withKeyPair(keyPair)
+			.keyId(UUID.randomUUID().toString())
+			.signatureAlgorithm(SignatureAlgorithm.ES256)
```
jzheaux, line 483: Since EC defaults to `ES256`, I think this should be a different algorithm, given the test name.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -369,4 +385,250 @@ private static URI convertAsURI(String header, URL url) {
...
 
+	/**
+	 * Creates a builder for constructing a {@link NimbusJwtEncoder} using the provided
+	 * {@link SecretKey}.
+	 * @param secretKey the {@link SecretKey} to use for signing JWTs
+	 * @return a {@link SecretKeyJwtEncoderBuilder} for further configuration
+	 * @since 7.0
+	 */
+	public static SecretKeyJwtEncoderBuilder withSecretKey(SecretKey secretKey) {
+		Assert.notNull(secretKey, "secretKey cannot be null");
```
jzheaux, line 396: Please move the assertion into the builder constructor so that the subclass can be self-sufficient. This is helpful in the event that we turn the subclass into a top-level class.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -369,4 +385,250 @@ private static URI convertAsURI(String header, URL url) {
...
+	/**
+	 * Creates a builder for constructing a {@link NimbusJwtEncoder} using the provided
+	 * {@link KeyPair}. The key pair must contain either an {@link RSAKey} or an
+	 * {@link ECKey}.
+	 * @param keyPair the {@link KeyPair} to use for signing JWTs
+	 * @return a {@link KeyPairJwtEncoderBuilder} for further configuration
+	 * @since 7.0
+	 */
+	public static KeyPairJwtEncoderBuilder withKeyPair(KeyPair keyPair) {
+		Assert.isTrue(keyPair != null && keyPair.getPrivate() != null && keyPair.getPublic() != null,
```
jzheaux, line 409: Please move any null checking to the subclass constructors so they are self-sufficient. This is helpful in the event that we turn the subclasses into top-level classes.

jzheaux (changes requested):
Hi, @surajbh123, thanks for the updates and for your feedback. 

Given some of the confusion I caused with some of my feedback, I wanted to take some more time with the PR before replying. You'll see that there are many more comments now with more specifics about what and why; hopefully that's clearer this time around.

To help with processing the comments, I also added a commit to my fork for you to review which applies my comments. I feel these changes make the builder classes easier to use and more maintainable by our team. You are welcome to review this commit to get a holistic view of the changes I'm proposing: 

https://github.com/jzheaux/spring-security/commit/7e1847dfc7a819a19ab3e6b9d18a94f0313c6cf6

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
surajbh123: ok

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
surajbh123: done

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -119,14 +133,16 @@ public void setJwkSelector(Converter<List<JWK>, JWK> jwkSelector) {
...
+
 	@Override
 	public Jwt encode(JwtEncoderParameters parameters) throws JwtEncodingException {
 		Assert.notNull(parameters, "parameters cannot be null");
 
 		JwsHeader headers = parameters.getJwsHeader();
-		if (headers == null) {
-			headers = DEFAULT_JWS_HEADER;
-		}
+		headers = (headers != null) ? headers : (this.jwsHeader != null) ? this.jwsHeader : DEFAULT_JWS_HEADER;
```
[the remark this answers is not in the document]
surajbh123: done

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -369,4 +385,250 @@ private static URI convertAsURI(String header, URL url) {
...
+		@Override
+		protected JWK buildJwk() {
+			if (super.jwsAlgorithm == null) {
+				super.jwsAlgorithm = JWSAlgorithm.ES256;
+			}
+
+			ECPublicKey publicKey = (ECPublicKey) super.keyPair.getPublic();
+			Curve curve = Curve.forECParameterSpec(publicKey.getParams());
+			if (curve == null) {
+				throw new JwtEncodingException("Unable to determine Curve for EC public key.");
```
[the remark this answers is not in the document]
surajbh123: done

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
surajbh123: done

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -369,4 +385,250 @@ private static URI convertAsURI(String header, URL url) {
...
+		}
+
+		/**
+		 * Sets the JWS algorithm to use for signing. Defaults to
+		 * {@link JWSAlgorithm#HS256}. Must be an HMAC-based algorithm (HS256, HS384, or
+		 * HS512).
+		 * @param macAlgorithm the {@link MacAlgorithm} to use
+		 * @return this builder instance for method chaining
+		 */
+		public SecretKeyJwtEncoderBuilder macAlgorithm(MacAlgorithm macAlgorithm) {
```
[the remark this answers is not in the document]
surajbh123: done

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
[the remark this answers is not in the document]
surajbh123: done

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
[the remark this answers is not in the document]
surajbh123: We cant do it as it will break older flows which are not using builders
so ` this.jwsHeader`   will be null

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoderTests.java
surajbh123: Supported algorithms: oct [HS256]
Supported algorithms: EC [ES256]
Supported algorithms: RSA [RS256, RS384, RS512, PS256, PS384, PS512]

@jzheaux   can we remove validation and let NimbusJwt do it

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoderTests.java
surajbh123: done

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoderTests.java
surajbh123: done

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -369,4 +385,250 @@ private static URI convertAsURI(String header, URL url) {
...
+
+		/**
+		 * Builds the {@link NimbusJwtEncoder} instance.
+		 * @return the configured {@link NimbusJwtEncoder}
+		 * @throws IllegalStateException if the key type is unsupported or the configured
+		 * JWS algorithm is not compatible with the key type.
+		 * @throws JwtEncodingException if the key is invalid (e.g., EC key with unknown
+		 * curve)
+		 */
+		public NimbusJwtEncoder build() {
```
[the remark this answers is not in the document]
surajbh123: done

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -369,4 +385,250 @@ private static URI convertAsURI(String header, URL url) {
...
+			this.jwsAlgorithm = (this.jwsAlgorithm != null) ? this.jwsAlgorithm : JWSAlgorithm.HS256;
+
+			OctetSequenceKey.Builder builder = new OctetSequenceKey.Builder(this.secretKey).keyUse(KeyUse.SIGNATURE)
+				.algorithm(this.jwsAlgorithm)
+				.keyID(this.keyId);
+
+			OctetSequenceKey jwk = builder.build();
+			JWKSource<SecurityContext> jwkSource = new ImmutableJWKSet<>(new JWKSet(jwk));
```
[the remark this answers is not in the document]
surajbh123: for default JWS header we can not move to private constructor
as it require algorithm and user can set or keep it default
so if we pass in constructer the default value will override user input

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtEncoder.java
```diff
@@ -119,14 +133,17 @@ public void setJwkSelector(Converter<List<JWK>, JWK> jwkSelector) {
 		this.jwkSelector = jwkSelector;
 	}
 
+	public void setDefaultJwsHeader(JwsHeader defaultJwsHeader) {
```
jzheaux: Please remove this method. The default header can be derived in the constructor.

jzheaux (approved):

jzheaux:
Thanks for your consistent effort on this PR, @surajbh123! It's now be merged into `main`.

Issue #17028, opened 2025-05-01
Deprecate SpringSecurityLdapTemplate


jzheaux:
`SpringSecurityLdapTemplate` exposes many operations that are available in modern-day Spring LDAP by way of `LdapClient` and `LdapQueryBuilder`. We should try and replace usage of `SpringSecurityLdapTemplate` with Spring LDAP's default `LdapClient` implementation to simplify maintenance and configuration.

- [ ] Change `PasswordComparisonAuthenticator` to use `LdapClient`
- [ ] Change `FilterBasedLdapUserSearch` to use `LdapClient`
- [ ] Change `ActiveDirectoryLdapAuthenticationProvider` to use `LdapClient`
- [ ] Add an `LdapAuthoritiesPopulator` implementation that uses `LdapClient`
- [ ] Deprecate `SpringSecurityLdapTemplate`

minkukjo:
@jzheaux 
I would like to work on this issue.
Please let me know if it's available for contribution.
Thanks!

jzheaux:
Thanks, @minkukjo! I've converted the first three to issues and we can start from there. If you would please comment on one of those and I'll assign it to you.

PR #17035, opened 2025-05-03
Update LDAP authority and role handling to use LdapClient

fer-marino:
refer https://github.com/spring-projects/spring-security/issues/17028
Still a work in progress.

Replaced SpringSecurityLdapTemplate with LdapClient for improved LDAP search and entry handling. Refactored related methods and tests to handle `NamingException`, use `LdapName` for DNs, and support updated attribute handling with `Attributes`. Updated authority mapping logic to streamline nested group resolution and enhance test cases.

Integrarion tests for the NestedLdapAuthoritiesPopulator are failing. I can't find an efficient way to get the DN of the result set to build LdapAuthorities, so the resulting set is messed up.

Another major difference with the LdapTemplate is that every search result contains at least the key `spring.security.ldap.dn` while the LdapClient does not. This breaks some other tests that maybe can be removed.

Another note worth mentioning is that integration tests still uses `ApacheDSContainer`, which is deprecated and should be replaced by `UnboundIdContainer`

Comments are very welcome! :)


jzheaux (changes requested):
Thanks for the contribution, @fer-marino. It think there may be some misunderstanding that is causing this extra work for you. We cannot change public or protected constructors or methods as this will break applications when they upgrade. For this reason, the ticket indicates the following (emphasis *mine*)

> *Add* an LdapAuthoritiesPopulator implementation that uses LdapClient

I hope this helps with some of your concerns about failing tests, etc., as well.

Issue #7056, opened 2019-06-28
NimbusJwtDecoder.withSecretKey() does not work with JWT that contains kid 


jzheaux:
Related to #7049

### Summary
When using NimbusJwtDecoder.withSecretKey() to decode a JWT token that contains a kid the decoding always fails as the provided SecretKey does not match.

### Actual Behavior

``` Signed JWT rejected: Another algorithm expected, or no matching key(s) found ``` is raised, because no match key is found (JWT token contains key, provided secret key not).

### Expected Behavior
As I specify the public key to used, the jwt token should be tried to be verified using this key. There is no need to find a matching key.
### Version

5.2.0.M2

### Sample

```
SecretKey verificationKey = // ... some key
String token = "someTokenWithKid";
NimbusJwtDecoder decoder = NimbusJwtDecoder.withSecretKey(secretKey).build();
decoder.decode(token);
```

PR #17046, opened 2025-05-06
Add possibility to customize JwkSource of NimbusJwtDecoder

marbon87:
We want to increase the cache-ttl of the NimbusJwtDecoder respectively the underlying JWKSource. Furthermore we want to use the refresh-ahead caching of JWKSource.

The com.nimbusds.jose.jwk.source.JWKSourceBuilder provides a lot of features to customize how jwks are cached / update etc. but currently there is no way to customize it.
The implemented possiiblity to customiz the used JWKSourceBuilder in NimbusJwtDecoder allows to use all features from JWKSourceBuilder.

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
@@ -262,6 +262,16 @@ public final class NimbusJwtDecoder implements JwtDecoder {
 	}
 
 	/**
+	 * Use the given {@code JWKSource} to create a JwkSourceJwtDecoderBuilder.
+	 * @param jwkSource the JWK Source to use
+	 * @return a {@link JwkSetUriJwtDecoderBuilder} for further configurations
+	 * @since 7.0
+	 */
+	public static JwkSourceJwtDecoderBuilder withJwkSource(JWKSource<SecurityContext> jwkSource) {
+		return new JwkSourceJwtDecoderBuilder(jwkSource);
+	}
+
+	/**
 	 * A builder for creating {@link NimbusJwtDecoder} instances based on a
 	 * <a target="_blank" href="https://tools.ietf.org/html/rfc7517#section-5">JWK Set</a>
 	 * uri.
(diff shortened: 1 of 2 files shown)


marbon87:
Hi @jzheaux, thanks for your feedback!

If i unterstand the current implementation correctly, adding a separate builder-creation method like `NimbusReactiveJwtDecoder#withJwkSource` would mean that the jwkSetUri must be specified explicitly / constructed manually. I prefer configuring the issuer-uri and let spring query the jwkSetUri by calling the well-known-config endpoint.

Furthermore i cannot use SpringJWKSource because it's private nor can i customize it because it's final.

Overall I have to write a lot more code as a user of spring-security, if i want to increase the timeouts.

jzheaux:
Possibly, though I wonder if it is as much as you are thinking (perhaps it's also more than I'm thinking). `SpringJWKSource` allows you to use `RestOperations` and Spring `Cache` instances. However, you are wanting to use Nimbus's caching. Because of that, I imagine you'll be satisfied with working with `JWKSourceBuilder` directly. 

As for the URI, I think that Nimbus has an API that's quite adept at this:

```java
AuthorizationServerMetadata metadata = AuthorizationServerMetadata.parse(new Issuer("https://example.org/issuer"));
String jwkSetUri = metadata.getJwkSetUri();
// ...
```

It seems to me that this would result in the following:

```java
@Bean
JwtDecoder jwtDecoder(String issuer) {
	AuthorizationServerMetadata metadata = AuthorizationServerMetadata.parse(new Issuer("https://example.org/issuer"));
	String jwkSetUri = metadata.getJwkSetUri();
    JWKSource<SecurityContext> source = JWKSourceBuilder.create(new URL(jwkSetUri))
        // your caching settings
        .build();
	return NimbusJwtDecoder.withJwkSource(source)
        // your decoding settings
        .build();
}
```

What complexities am I failing to consider?

marbon87:
Your suggestion should work but i am not sure what's missing to get it "completed". 
For example the SpringJWKSource seems to be thread-safe, because of the usage of a ReentrantLock.
Furthermore i thought that there must be a reason why spring does not use your approach as the default and instead implements a custom JWKSetSource. 

Regarding to spring cache, i did not find an simple solutation for a refresh or/and fault tolerant ahead cache.

jzheaux:
It uses a custom `JWKSetSource` to allow applications to use their `RestOperations` and `Cache` instances.

If you want to look further into Spring Cache, I think [Spring's JCache integration](https://docs.spring.io/spring-boot/reference/io/caching.html#io.caching.provider.jcache) would make sense since you can integrate in this way with Hazelcast and other caches that support refresh-ahead. That said, you know your requirements best and whether the ease of Nimbus's in-memory caching will suit your needs. 

> Your suggestion should work but i am not sure what's missing to get it "completed".

Am I understanding correctly that you are asking for how the PR would change? To add `withJwkSource`, you would need to add the appropriate static method to `NimbusJwtDecoder`; it's implementation would likely be similar to `withJwkSource` in `NimbusReactiveJwtDecoder`. Would you like me to push a stub implementation to your branch to get you started?

marbon87:
> Am I understanding correctly that you are asking for how the PR would change? To add `withJwkSource`, you would need to add the appropriate static method to `NimbusJwtDecoder`; it's implementation would likely be similar to `withJwkSource` in `NimbusReactiveJwtDecoder`. Would you like me to push a stub implementation to your branch to get you started?

Thanks for your offer, but i think i can manage that. Should i close this pr and create a new one referencing this?

marbon87:
Hi @jzheaux,

i updated the pr. Please have a look.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -261,6 +261,16 @@ public static SecretKeyJwtDecoderBuilder withSecretKey(SecretKey secretKey) {
 		return new SecretKeyJwtDecoderBuilder(secretKey);
 	}
 
+	/**
+	 * Use the given <a href="https://tools.ietf.org/html/rfc7517#section-5">JWK Set</a>
+	 * uri.
+	 * @param jwkSetUri the JWK Set uri to use
+	 * @return a {@link JwkSetUriJwtDecoderBuilder} for further configurations
+	 */
+	public static JwkSetUriJwtDecoderBuilder withJwkSource(JWKSource<SecurityContext> jwkSetUri) {
```
jzheaux, line 270: Please check the parameter name here. It's probably a left-over from the copy-paste.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -261,6 +261,16 @@ public static SecretKeyJwtDecoderBuilder withSecretKey(SecretKey secretKey) {
 		return new SecretKeyJwtDecoderBuilder(secretKey);
 	}
 
+	/**
+	 * Use the given <a href="https://tools.ietf.org/html/rfc7517#section-5">JWK Set</a>
+	 * uri.
+	 * @param jwkSetUri the JWK Set uri to use
+	 * @return a {@link JwkSetUriJwtDecoderBuilder} for further configurations
```
jzheaux, line 268: Please add `@since 7.0`

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -274,7 +284,7 @@ public static final class JwkSetUriJwtDecoderBuilder {
 		private static final JOSEObjectTypeVerifier<SecurityContext> NO_TYPE_VERIFIER = (header, context) -> {
 		};
 
-		private final Function<RestOperations, String> jwkSetUri;
+		private Function<RestOperations, String> jwkSetUri;
```
jzheaux, line 287: Since providing a `JWKSource` supercedes a `RestOperations` and a `Cache`, please create `JwkSourceJwtDecoderBuilder` instead of repurposing `JwkSetUriJwtDecoderBuilder`. In that way, people won't provide a `JWKSource` and then also try and provide a `RestOperations` or `Cache`.

jzheaux: Note that `JwkSourceJwtDecoderBuilder` will not need the `validateTypes` flag since that was added as part of migrating from 6.x to 7.0. Since this class will be added in 7.0, it doesn't need the same transition flag.

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
```diff
@@ -559,6 +560,22 @@ public void decodeWhenUsingSecretKeyWithKidThenStillUsesKey() throws Exception {
 		// @formatter:on
 	}
 
+	// gh-7056
+	@Test
+	public void decodeWhenUsingJwkSource() throws Exception {
+		JWKSource<SecurityContext> source = (a, b) -> {
```
jzheaux, line 566: Please use Mockito instead of an inline stub. That will allow you to use `verify` at the end of the test to confirm its usage

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
```diff
@@ -559,6 +560,22 @@ public void decodeWhenUsingSecretKeyWithKidThenStillUsesKey() throws Exception {
 		// @formatter:on
 	}
 
+	// gh-7056
+	@Test
+	public void decodeWhenUsingJwkSource() throws Exception {
```
jzheaux, line 565: To follow the naming convention in this file, will you please follow this format for the method name:

```java
{methodName}When{Condition}Then{Expectation}
```

For example:

```java
withJwkSourceWhenDefaultsThenUsesProvidedJwkSource
```

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
```diff
@@ -559,6 +560,22 @@ public void decodeWhenUsingSecretKeyWithKidThenStillUsesKey() throws Exception {
 		// @formatter:on
 	}
 
+	// gh-7056
```
jzheaux, line 563: This looks like it might be a stray line from a copy-paste. Since this test wasn't added as part of #7056, let's please leave it off.

jzheaux (changes requested):
Thanks for your patience, @marbon87! I realized once your PR was submitted that https://github.com/spring-projects/spring-security/issues/17181 was needed first. With that completed, I've left some inline feedback for you.

In addition to that feedback, will you please help me with your commit message as well? It looks like there isn't a blank line between your title and your sign-off message which can mess with how certain tools render the commit.

Finally, some of these changes I'd normally do for the contributor, however this is trickier when the PR is from your `main` branch, which I don't want to force push to. In the future, if you first cut a branch and then submit, it's a little easier for me to apply some of the needed changes.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
marbon87: Wouldn't it be possible to backport this PR to the 6.5er branch?

jzheaux: No, we cannot backport it to 6.x as we only add features in minor branches. Since there is no 6.6 planned, 7.0 is the next available release.

jzheaux:
Thanks, @marbon87, for this PR! It has been merged to `main`.

In the meantime on 6.x, please consider using this snippet:

```java
@Bean 
JwtDecoder jwtDecoder() {
  AuthorizationServerMetadata metadata = AuthorizationServerMetadata
          .resolve(new Issuer("https://example.org/issuer"));
  String jwkSetUri = metadata.getJwkSetUri();
  JWKSource<SecurityContext> source = JWKSourceBuilder.create(new URL(jwkSetUri))
      // your caching settings
      .build();
  ConfigurableJWTProcessor<SecurityContext> jwtProcessor = new DefaultJWTProcessor<>();
  JWSKeySelector<SecurityContext> selector = new JWSVerificationKeySelector<>(
    Set.of(JWSAlgorithm.RS256), jwkSource)
  jwtProcessor.setJWSKeySelector(selector);
  return new NimbusJwtDecoder(jwtProcessor);
}  
```

Issue #17041, opened 2025-05-06
SwitchUserGrantedAuthority cannot be deserialized in Webflux


JohnNiang:
**Describe the bug**

I used SwitchUserWebfilter to implement an impersonation function, but it didn't work very well, pleas see the log below:

I found a similar issue in <https://github.com/spring-projects/spring-security/issues/11775>, but the PR https://github.com/spring-projects/spring-security/pull/11758 only resolved in Servlet.

```java
Halo  | Caused by: com.fasterxml.jackson.databind.JsonMappingException: The class with org.springframework.security.web.authentication.switchuser.SwitchUserGrantedAuthority and name of org.springframework.security.web.authentication.switchuser.SwitchUserGrantedAuthority is not in the allowlist. If you believe this class is safe to deserialize, please provide an explicit mapping using Jackson annotations or by providing a Mixin. If the serialization is only done by a trusted source, you can also enable default typing. See https://github.com/spring-projects/spring-security/issues/4370 for details (through reference chain: org.springframework.security.core.context.SecurityContextImpl["authentication"])
Halo  | 	at com.fasterxml.jackson.databind.JsonMappingException.wrapWithPath(JsonMappingException.java:401) ~[jackson-databind-2.18.3.jar:2.18.3]
Halo  | 	at com.fasterxml.jackson.databind.JsonMappingException.wrapWithPath(JsonMappingException.java:360) ~[jackson-databind-2.18.3.jar:2.18.3]
Halo  | 	at com.fasterxml.jackson.databind.deser.BeanDeserializerBase.wrapAndThrow(BeanDeserializerBase.java:1964) ~[jackson-databind-2.18.3.jar:2.18.3]
Halo  | 	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserializeFromObject(BeanDeserializer.java:401) ~[jackson-databind-2.18.3.jar:2.18.3]
Halo  | 	at com.fasterxml.jackson.databind.deser.BeanDeserializer._deserializeOther(BeanDeserializer.java:220) ~[jackson-databind-2.18.3.jar:2.18.3]
Halo  | 	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:187) ~[jackson-databind-2.18.3.jar:2.18.3]
[...]

PR #17064, opened 2025-05-07
Fix the problem of not deserializing SwitchUserGrantedAuthority in Webflux

JohnNiang:
This PR moves the SwitchUserGrantedAuthorityMixIn annotations  into WebJackson2Module instead of WebServlet Jackson2Module to fix the deserialization problem of SwitchUserGrantedAuthority in Webflux.

Fixes https://github.com/spring-projects/spring-security/issues/17041

### DIFF
--- a/web/src/main/java/org/springframework/security/web/jackson2/WebJackson2Module.java
+++ b/web/src/main/java/org/springframework/security/web/jackson2/WebJackson2Module.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 the original author or authors.
+ * Copyright 2015-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


web/src/main/java/org/springframework/security/web/jackson2/WebServletJackson2Module.java
```diff
@@ -58,7 +57,6 @@ public void setupModule(SetupContext context) {
 		context.setMixInAnnotations(SavedCookie.class, SavedCookieMixin.class);
 		context.setMixInAnnotations(DefaultSavedRequest.class, DefaultSavedRequestMixin.class);
 		context.setMixInAnnotations(WebAuthenticationDetails.class, WebAuthenticationDetailsMixin.class);
-		context.setMixInAnnotations(SwitchUserGrantedAuthority.class, SwitchUserGrantedAuthorityMixIn.class);
```
jzheaux, line 61: While I see your point, we can't do this since it's a public class and applications may be using this separately from `WebJackson2Module`. As such, will you please leave it in place here?

web/src/main/java/org/springframework/security/web/jackson2/WebJackson2Module.java
```diff
@@ -21,14 +21,16 @@
 
 import org.springframework.security.jackson2.SecurityJackson2Modules;
 import org.springframework.security.web.authentication.preauth.PreAuthenticatedAuthenticationToken;
+import org.springframework.security.web.authentication.switchuser.SwitchUserGrantedAuthority;
```
jzheaux, line 24: Will you please update the copyright header so that it ends in `2025`?

jzheaux (changes requested):
Nice, @JohnNiang! Thanks for the quick turnaround.

In addition to my inline feedback will you please also:

1. Add unit tests to demonstrate that the feature works
2. Update the commit message to contains `Closes gh-17041` on a separate line from the title like so:

```
Add SwitchUserGrantedAuthority to WebFlux Jackson Module

Closes gh-17041
```

jzheaux (approved):

jzheaux:
Thanks, @JohnNiang, this is now merged into `main`!

Issue #17062, opened 2025-05-06
Consider removal of APPLICATION_JSON_UTF8 (and MediaTypes)


rwinch:
MediaTypes was put in as a temporary measure for getting updated to Spring Framework 7 so that we could more carefully consider the removal of APPLICATION_JSON_UTF8.

Spring Framework [removed APPLICATION_JSON_UTF8](https://github.com/spring-projects/spring-framework/commit/5044b70a4038f2ae29cee262f63fb0e940297bda#diff-ad51108bdef54c453f4b18b455169d4d5db0b469b1be2033cf521edbf92fd3b9L123-L136) because:

> Deprecated as of 5.2 in favor ofAPPLICATION_JSON since major browsers like Chrome <a href="https://bugs.chromium.org/p/chromium/issues/detail?id=438464"> now comply with the specification</a> and interpret correctly UTF-8 special characters without requiring a UTF-8 parameter.

We should review if APPLICATION_JSON_UTF8 can be removed from Spring Security. In order to do this, we need to review Spring Security implementations that use it and see if the corresponding specifications require APPLICATION_JSON_UTF8 or if APPLICATION_JSON is acceptable. In particular, this appears to impact a number of OAuth classes, so their corresponding specs must be reviewed.

rwinch:
Closing in favor of gh-17070

PR #17067, opened 2025-05-07
Remove MediaTypes

ngocnhan-tran1996:
Closes gh-17062


rwinch:
Thank you. I'm sorry that #17062 was not very clear. The point was not just to remove the `MediaTypes` class, but carefully consider removal of APPLICATION_JSON_UTF8 as well. I've updated the description of the ticket to be more clear

PR #17068, opened 2025-05-07
Check for null Principal#getName

piotrmacha:
# Problem

`BasicAuthenticationFilter` is comparing existing `Authentication` with username but it assumes that `existingAuth.getName()` returns non-null `String` and calls `.equals(username)` on it. If `existingAuth != null` and `existingAuth.getName()` returns null, it leads to NullPointerException.

This is an unlikely scenario but I have encountered it in an application with highly customized security config.

# Solution

Add additonal `existingAuth.getName() == null` check before calling `.equals()` on it.


jzheaux:
Thanks, @piotrmacha, for the contribution. Can you share more about the configurations where you've seen this happen? 

It feels very much like a usage error for `getName` to return `null`; I'd be more inclined to add `@NonNull` to `Authentication#getName` than for an application to silently continue.

ronodhirSoumik:
Just to clear my understanding here- why is the change needed?
the `getName()` of `Principle.java` (implemented by `Authentication.java`) should always nonNull. So the `existingAuth == null` condition only should suffice, right?

jzheaux:
I'm going to close this PR as it appears unnecessary. Please feel free to update with a use case, and we can take another look.

Issue #16863, opened 2025-04-01
JdbcUserDetailsManager.setEnableUpdatePassword


rwinch:
`JdbcUserDetailsManager` should implement `UserDetailsPasswordService` and have a property `enableUpdatePassword` that is `false` by default (to prevent migrations that might produce passwords that are too large for the current database schema). This would allow users to easily opt into migrating passwords with JDBC. The new method would look something like this:

```java
public UserDetails updatePassword(UserDetails user, String newPassword) {
    UserDetails updated = User.withUserDetails(user).password(newPassword).build();
    updateUser(updated);
    return updated;
}
```

franticticktick:
Hi @rwinch, could you assign this ticket to me please?

vasanth-79:
Hi @rwinch, Could you please assign this to me?

I've made the changes and raised PR #16881

rwinch:
@vasanth-79  Thank you for the PR. Can you please add a property `enableUpdatePassword` that is false by default and only runs this method if true? Please also add some tests.

jhl221123:
Hi @rwinch,

Following up on the closure of PR #16881 and your recent comment there, I would like to work on this issue #16863. I understand the need for the `enableUpdatePassword` property (defaulting to false) and additional tests as discussed.

Could you please assign this to me?

Thanks

injae-kim:
> fix PR: https://github.com/spring-projects/spring-security/pull/17071

@jhl221123 made awesome fix PR https://github.com/spring-projects/spring-security/pull/17071, PTAL when you have time 😄

PR #17071, opened 2025-05-07
Implement UserDetailsPasswordService in JdbcUserDetailsManager

jhl221123:
**Summary:**

Implements `UserDetailsPasswordService` in `JdbcUserDetailsManager` to allow password updates, controlled by a new flag.

**Changes:**

* Added `updatePassword(UserDetails user, String newPassword)` method.
* Introduced `enableUpdatePassword` boolean property (default: false) and setter `setEnableUpdatePassword(boolean)`.
* `updatePassword` method logic updated:
  * If `enableUpdatePassword` is true, it updates the user's password via updateUser.
  * If `enableUpdatePassword` is false, it returns the original `UserDetails` without performing any update.
* Added tests in `JdbcUserDetailsManagerTests` to verify both scenarios (`updatePasswordWhenDisabledReturnOriginalUser` and `updatePasswordWhenEnabledShouldUpdatePassword`).

Closes gh-16863

### DIFF
--- a/core/src/main/java/org/springframework/security/provisioning/JdbcUserDetailsManager.java
+++ b/core/src/main/java/org/springframework/security/provisioning/JdbcUserDetailsManager.java
@@ -64,6 +64,7 @@ import org.springframework.util.Assert;
  * using this implementation for managing your users.
  *
  * @author Luke Taylor
+ * @author Junhyeok Lee
  * @since 2.0
  */
 public class JdbcUserDetailsManager extends JdbcDaoImpl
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/springframework/security/provisioning/JdbcUserDetailsManager.java
```diff
@@ -65,7 +66,8 @@
  * @author Luke Taylor
```
injae-kim, line 66: can you add your name on here?

core/src/main/java/org/springframework/security/provisioning/JdbcUserDetailsManager.java
```diff
@@ -603,4 +621,14 @@ private void validateAuthorities(Collection<? extends GrantedAuthority> authorit
 		}
 	}
 
+	@Override
+	public UserDetails updatePassword(UserDetails user, String newPassword) {
```
injae-kim, line 625:
```suggestion
	@Override
	/**
	  * comments like "updatePassword only when {@link #enableUpdatePassword} is true?
	  */
	public UserDetails updatePassword(UserDetails user, String newPassword) {
```
nit. how about adding comment about `updatePassword ` flag?

rwinch:
Thank you for the PR @jhl221123! This is now queued to be merged into main as soon as the build passes

Issue #16058, opened 2024-11-11
Exceptions for Authorized Objects should propagate when returned from a Controller


jzheaux:
When using `@AuthorizeReturnObject` on a controller method like so:

```java
@AuthorizeReturnObject
@GetMapping(...)
public MyObject getObject() {
    return ...
}
```

Any unhandled authorization failures are not propagated to the `ExceptionTranslationFilter` and instead result in a 500.

Security currently supports using `@HandleAuthorizationDenied` to provide behavior like ignoring that field when the response is serialized. However, it would be helpful to be able to leverage the logic in `ExceptionTranslationFilter`, including calling any configured `AccessDeniedHandler`.

One way to do this may be to provide an implementation of `AbstractView` that is used by a default exception handler. It would ideally have a similar effect to:

```
@ExceptionHandler(HttpMessageNotWritableException.class)
View handleWrite(HttpMessageNotWritableException ex) {
    if (ex.getRootCause() instanceof AuthorizationDeniedException denied) {
       return new AbstractView() {
          @Override
          protected void renderMergedOutputModel(Map<String, Object> model,
                HttpServletRequest request, HttpServletResponse response)
                throws Exception {
             throw ex;
          }
       };
    }
    throw ex;
}
```

but without applications needing to register it.

evgeniycheban:
Hi, @jzheaux can you assign it to me?

PR #17074, opened 2025-05-08
Exceptions for Authorized Objects should propagate when returned from a Controller

evgeniycheban:
Closes gh-16058

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/method/configuration/AuthorizationProxyWebConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/method/configuration/AuthorizationProxyWebConfiguration.java
@@ -16,7 +16,11 @@
 
 package org.springframework.security.config.annotation.method.configuration;
 
+import java.util.List;
 import java.util.Map;
+
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
 
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.context.annotation.Bean;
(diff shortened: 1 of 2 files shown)


config/src/main/java/org/springframework/security/config/annotation/method/configuration/AuthorizationProxyWebConfiguration.java
```diff
@@ -16,27 +16,41 @@
...
 
 	@Bean
 	@Role(BeanDefinition.ROLE_INFRASTRUCTURE)
 	AuthorizationAdvisorProxyFactory.TargetVisitor webTargetVisitor() {
 		return new WebTargetVisitor();
 	}
 
+	@Override
+	public void extendHandlerExceptionResolvers(List<HandlerExceptionResolver> resolvers) {
+		resolvers.add(0, new HttpMessageNotWritableAccessDeniedExceptionResolver());
```
jzheaux, line 51: One thing that concerns me is that this might break applications that have an `@ExceptionHandler` for `AccessDeniedException`-caused `HttpMessageNotWritableException` already.

Further, it would be nice if there were a clear way for applications to override Security's default handling of this scenario.

jzheaux (changes requested):
Nice solution, @evgeniycheban! I've left a piece of feedback inline.

config/src/main/java/org/springframework/security/config/annotation/method/configuration/AuthorizationProxyWebConfiguration.java
evgeniycheban: I implemented a fix for this scenario by finding the `DefaultHandlerExceptionResolver`'s index and adding Security's `ExceptionResolver` before the default resolver.  This way user-defined `@ExceptionHandler` will always take precedence and try to resolve an exception first since the `ExceptionHandlerExceptionResolver` is added before the `extendHandlerExceptionResolvers` method call. I think we could also consider providing a configuration option to control wether Spring Security should handle this type of scenario or not.

evgeniycheban:
Hi @jzheaux I updated the PR, according to your feedback, thanks.

config/src/main/java/org/springframework/security/config/annotation/method/configuration/AuthorizationProxyWebConfiguration.java
jzheaux: I imagine that the way to override would be to add an exception handler or add their own exception resolver. Is there another way that you had in mind?

evgeniycheban: I think then it's possible now to override it by defining a custom '@ExceptionHandler', I have added a test covering that.

jzheaux (approved):

jzheaux:
Thanks, @evgeniycheban! This will merge to `main` once the build completes.

Issue #16012, opened 2024-10-30
Add JdbcRelyingPartyRegistrationRepository


sasirekha98:
I have an application where the SP metadata is generated as XML and put in the database. We pull the XML and generate the authentication request. When I want to move to Spring Security, I don't find the same option where I can get the relying party details from the data source or directly provide the XML file location.

Currently, as per my understanding, we have the option to load the XML file of the IDP through metadata locations through spring security like this 

RelyingPartyRegistrations
                .fromMetadataLocation(oktaMetadataUrl)
                .registrationId("okta")).build();

Similar to that, it would be nice to have an option where we can load the relying party details when we have them in the database or some location already. We just need to load that file instead of adding the details one by one to the RelyingpartyRegistration.

it will be easier if I have an option where I have to mention the metadata location of the sp as well like this 
RelyingPartyRegistrations
                .fromMetadataLocationofSP(location)
                .fromMetadataLocationofIDP(oktaMetadataUrl)
                .registrationId("okta")).build();

jzheaux:
Thanks for the suggestion, @sasirekha98. Are you interested in supplying a PR for this? You can look at the various `JdbcXXX` classes already in Spring Security for guidance.

wapkch:
@jzheaux May i work on this?

jzheaux:
Sorry that I missed you, @wapkch. Are you still interested in working on this? If so, please comment here and feel free also to proceed with a PR.

wapkch:
Yes, I would be happy to work on this. I’ll proceed with the PR shortly.

PR #17077, opened 2025-05-08
Add JdbcRelyingPartyRegistrationRepository

wapkch:
Closes gh-16012


jzheaux (changes requested):
Hi, @wapkch, thanks for this PR! Looking a #16012, I believe the ask is to be able to have separate places from which to draw information for the RP and AP metadata. As such, I think let's begin with implementing `AssertingPartyMetadataRepository` (so `JdbcAssertingPartyMetadataRepository`). You should be able to preserve most of what you've written, just staying focused on the asserting party. Are you able to make these changes?

wapkch:
> Hi, @wapkch, thanks for this PR! Looking at #16012, I believe the ask is to be able to have separate places from which to draw information for the RP and AP metadata. As such, I think let's begin with implementing `AssertingPartyMetadataRepository` (so `JdbcAssertingPartyMetadataRepository`). You should be able to preserve most of what you've written, just staying focused on the asserting party. Are you able to make these changes?

Thanks for the feedback! Yes, I can make those changes. Appreciate the guidance!

jzheaux:
This is looking good, @wapkch. I added some polishes in 86189f380d88c8670e113341ecb20bdd213e2c96. Also, I polished your original `save` method and placed that in a separate commit that marks you as co-author so that you get credit for that initial work. Let me know if that works for you. If so, I think that this is ready to merge; thanks for your help!

jzheaux (approved):

jzheaux:
Thanks again, @wapkch! This will merge into `main` once the build completes.

PR #17080, opened 2025-05-08
Fix DPoP jkt claim to be JWK SHA-256 thumbprint

### LINKED ISSUE #17079: DPoP JWK Thumbprint validation does not conform to RFC7638
**Describe the bug**
The DPoP Thumbprint in the `jkt` claim is calculated according to RFC7638, and the spring implementation does a certificate thumbprint, not a JWK Thumbprint

**To Reproduce**
1. Generate a DPoP Proof using: https://github.com/panva/openid-client
2. This will create the `jtk` claim using a JWK Thumbprint according to the RFC
3. Spring cannot validate it, and the DPoP Proof is rejected. :(

**Expected behavior**
The DPoP proof should be valid!

**Sample**

PR to follow shortly showing the difference and fix once validated, if we are unable to prove that it works, we'll create a sample repo.

dkowis:
This is the proper implementation for a [JWK Thumbprint](https://datatracker.ietf.org/doc/html/rfc7638#section-3). Spring Security was doing a Certificate Thumbprint, which is correct for `ath` claims to verify the certificate used in the JWK, but it's not correct for a DPoP verification `jkt` claim.

Resolves #17079

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/DPoPAuthenticationProvider.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/DPoPAuthenticationProvider.java
@@ -193,25 +193,22 @@ public final class DPoPAuthenticationProvider implements AuthenticationProvider 
 				return OAuth2TokenValidatorResult.failure(error);
 			}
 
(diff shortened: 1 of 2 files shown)


jgrandja:
@dkowis Thank you for catching this! This is now merged along with a minor polish commit.

Issue #17012, opened 2025-04-29
Base64StringKeyGenerator should not have a minimum length


OrangeDog:
**Expected Behavior**

As with all other `StringKeyGenerator`s and `BytesKeyGenerator`s, developers should be free to choose an output length as small as they wish.

**Current Behavior**

Throws `IllegalArgumentException` if `keyLength` is less than 32.

**Context**

I want to generate random keys that are reasonably easy for humans to transcribe.
Base64 provides a good balance of usable character set and entropy per character.

However, this forces a _minimum_ of 43 characters (`⌈ 32 * 4/3 ⌉`), which is unreasonable for human use.

OrangeDog:
It may also be a bug that `keyLength` is not the actual length of the key, but the pre-encoded length.

PR #17091, opened 2025-05-12
Remove 32-byte minimum keyLength restriction in `Base64StringKeyGenerator`

therepanic:
The main purpose of this change is to essentially remove the strict 32 byte key length limit to allow developers to generate shorter Base64 strings if needed. Now, when trying to pass `keyLength < DEFAULT_KEY_LENGTH (32)` into the constructor, an exception is thrown, which is not always justified for less critical token generation tasks.

I believe that we don't need such a strict condition and we can soften the condition and just check that the key length is not less than or equal to 0.

Fix: #17012


rwinch:
Thanks for the PR @therepanic! I've scheduled this for merge as soon as the build passes

Issue #17099, opened 2025-05-13
Remove GET request support from Saml2AuthenticationTokenConverter


jzheaux:
`Saml2AuthenticationTokenConverter` tests if the HTTP method is GET in order to correctly translate `SAMLResponse` parameter.

However, neither [the SAML spec](https://github.com/spring-projects/spring-security/issues/11161#issuecomment-1119832599) nor Spring Security support processing the `<saml2:Response>` in a GET request. As such, we should remove this to alleviate confusion.

To keep upgrade passive, the first step is to add a property that defaults to `true`:

```java
public void setShouldInflateResponse(boolean shouldInflate);
```

In Spring Security 8, this can be deprecated and switched to `false`, then in Spring Security 9 it can be removed.

- [ ] Add `setShouldInflateResponse`
- [ ] Deprecate `setShouldInflateResponse`

jzheaux:
Closing in favor of #17108

PR #17108, opened 2025-05-14
Remove GET request support from Saml2AuthenticationTokenConverter

ngocnhan-tran1996:
I think we can define `Boolean shouldInflate` and method `decode` will be added

```java
if (this.shouldInflate == null) {
	this.shouldInflate = HttpMethod.GET.matches(request.getMethod());
}
```

If not, this will break `Saml2LoginBeanDefinitionParserTests` and `Saml2AuthenticationTokenConverterTests`

```text
Saml2LoginBeanDefinitionParserTests > authenticateWhenCustomAuthnRequestRepositoryThenUses() FAILED
    org.mockito.exceptions.verification.WantedButNotInvoked at Saml2LoginBeanDefinitionParserTests.java:323

Saml2LoginBeanDefinitionParserTests > authenticateWhenAuthenticationResponseValidThenAuthenticate() FAILED
    java.lang.AssertionError at Saml2LoginBeanDefinitionParserTests.java:213

Saml2LoginBeanDefinitionParserTests > authenticateWhenAuthenticationResponseValidThenAuthenticationSuccessEventPublished() FAILED
    java.lang.AssertionError at Saml2LoginBeanDefinitionParserTests.java:245

Saml2LoginBeanDefinitionParserTests > authenticateWhenCustomSecurityContextHolderStrategyThenUses() FAILED
    java.lang.AssertionError at Saml2LoginBeanDefinitionParserTests.java:228

Saml2LoginBeanDefinitionParserTests > authenticateWhenCustomAuthenticationManagerThenUses() FAILED
    java.lang.AssertionError at Saml2LoginBeanDefinitionParserTests.java:280
```

Issue: gh-17099

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverter.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -42,6 +42,8 @@ public final class Saml2AuthenticationTokenConverter implements AuthenticationCo
 	private final RelyingPartyRegistrationResolver relyingPartyRegistrationResolver;
 
 	private Saml2AuthenticationRequestRepository<AbstractSaml2AuthenticationRequest> authenticationRequestRepository;
+
+	private boolean shouldConvertGetRequests = true;
 
 	/**
 	 * Constructs a {@link Saml2AuthenticationTokenConverter} given a strategy for
@@ -86,16 +88,27 @@ public final class Saml2AuthenticationTokenConverter implements AuthenticationCo
 		this.authenticationRequestRepository = authenticationRequestRepository;
 	}
 
+	/**
+	 * Use the given {@code shouldConvertGetRequests} to convert {@code GET} requests.
+	 * Default is {@code true}.
+	 * @param shouldConvertGetRequests the {@code shouldConvertGetRequests} to use
+	 * @since 7.0
+	 */
+	public void setShouldConvertGetRequests(boolean shouldConvertGetRequests) {
+		this.shouldConvertGetRequests = shouldConvertGetRequests;
+	}
+
 	private String decode(HttpServletRequest request) {
 		String encoded = request.getParameter(Saml2ParameterNames.SAML_RESPONSE);
 		if (encoded == null) {
 			return null;
 		}
+		boolean isGet = HttpMethod.GET.matches(request.getMethod());
+		if (!this.shouldConvertGetRequests && isGet) {
+			return null;
+		}
 		try {
-			return Saml2Utils.withEncoded(encoded)
-				.requireBase64(true)
-				.inflate(HttpMethod.GET.matches(request.getMethod()))
-				.decode();
+			return Saml2Utils.withEncoded(encoded).requireBase64(true).inflate(isGet).decode();
 		}
 		catch (Exception ex) {
 			throw new Saml2AuthenticationException(new Saml2Error(Saml2ErrorCodes.INVALID_RESPONSE, ex.getMessage()),
--- a/saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverterTests.java
+++ b/saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverterTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2021 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverter.java
```diff
@@ -43,6 +43,8 @@ public final class Saml2AuthenticationTokenConverter implements AuthenticationCo
 
 	private Saml2AuthenticationRequestRepository<AbstractSaml2AuthenticationRequest> authenticationRequestRepository;
 
+	private Boolean shouldInflate;
```
jzheaux, line 46: Please use `boolean` and give it an initial value of `true`. We typically reserve `null` in `Boolean` for "no opinion".

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverter.java
```diff
@@ -86,16 +88,26 @@ public void setAuthenticationRequestRepository(
 		this.authenticationRequestRepository = authenticationRequestRepository;
 	}
 
+	/**
+	 * Use the given {@code shouldInflate} to inflate request.
```
jzheaux, line 92: Please indicate the default value in this JavaDoc

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverter.java
```diff
@@ -86,16 +88,26 @@ public void setAuthenticationRequestRepository(
...
+	 * @since 7.0
+	 */
+	public void setShouldInflateResponse(boolean shouldInflate) {
+		this.shouldInflate = shouldInflate;
+	}
+
 	private String decode(HttpServletRequest request) {
+		// prevent to break passivity in Saml2LoginBeanDefinitionParserTests
+		if (this.shouldInflate == null) {
+			this.shouldInflate = HttpMethod.GET.matches(request.getMethod());
```
jzheaux, line 103: This is not thread-safe since multiple requests will use the same instance of this class. Please approach this in a way that does not involve updating a member variable.

saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverterTests.java
```diff
@@ -61,6 +61,7 @@ public class Saml2AuthenticationTokenConverterTests {
 	public void convertWhenSamlResponseThenToken() {
 		Saml2AuthenticationTokenConverter converter = new Saml2AuthenticationTokenConverter(
 				this.relyingPartyRegistrationResolver);
+		converter.setShouldInflateResponse(false);
```
jzheaux, line 64: Please don't alter any unit tests. In this way, we can have high confidence that the change is a passive one. 

That said, I believe adding a test would be a good idea. I think we should test when `shouldInflate` is `false` and the request is a GET.

jzheaux (changes requested):
Thanks, @ngocnhan-tran1996! I've left some feedback inline.

saml2/saml2-service-provider/src/test/java/org/springframework/security/saml2/provider/service/web/Saml2AuthenticationTokenConverterTests.java
ngocnhan-tran1996: >Please don't alter any unit tests.
This is unclear to me. I've made changes on my local machine, and some tests will fail. We should either set the value to false or update the expected results in the tests. Which option should we choose?

jzheaux: In order for the change to be passive, all existing unit tests should pass. If they fail, it's because the default behavior has changed. If you default `shouldInflateResponse` to `true`, I believe you'll be able to leave those tests as-is. If you run into trouble with that, ping me and I can help.

jzheaux: This is the code I would expect:

```java
boolean isGet = HttpMethod.GET.matches(request.getMethod());
if (!this.shouldInflateResponse && isGet) {
    return null;
}
try {
	return Saml2Utils.withEncoded(encoded).requireBase64(true).inflate(isGet).decode();
}
catch (Exception ex) {
	throw new Saml2AuthenticationException(new Saml2Error(Saml2ErrorCodes.INVALID_RESPONSE, ex.getMessage()), ex);
}
```

However, I can see how the name I suggested is causing confusion and will likely cause confusion with others. If we change the name to `convertGetRequests`, does that help? Then the code would instead be:

```java
boolean isGet = HttpMethod.GET.matches(request.getMethod());
if (!this.convertGetRequests && isGet) {
    return null;
}
try {
	return Saml2Utils.withEncoded(encoded).requireBase64(true).inflate(isGet).decode();
}
catch (Exception ex) {
	throw new Saml2AuthenticationException(new Saml2Error(Saml2ErrorCodes.INVALID_RESPONSE, ex.getMessage()), ex);
}
```

jzheaux:
Thanks, @ngocnhan-tran1996! This is now merged into `main`. Also in f73f253beb4bddbbabe1abe9bc58737a3ca82b6f I updated related authentication converter implementations. I also added migration docs.

PR #17126, opened 2025-05-16
ClearSiteDataHeaderWriter log is misleading 

joaquinjsb:
Hello,
this is just a simple log typo which could be misleading, whether the requestmatcher matches or not, it's logging that it isn't.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/header/writers/ClearSiteDataHeaderWriter.java
+++ b/web/src/main/java/org/springframework/security/web/header/writers/ClearSiteDataHeaderWriter.java
@@ -74,11 +74,12 @@ public final class ClearSiteDataHeaderWriter implements HeaderWriter {
 		if (this.requestMatcher.matches(request)) {
 			if (!response.containsHeader(CLEAR_SITE_DATA_HEADER)) {
 				response.setHeader(CLEAR_SITE_DATA_HEADER, this.headerValue);
-			}
+			}			
+		} else {
+			this.logger.debug(
+					LogMessage.format("Not injecting Clear-Site-Data header since it did not match the requestMatcher %s",
+							this.requestMatcher));
 		}
-		this.logger.debug(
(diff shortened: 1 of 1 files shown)


web/src/main/java/org/springframework/security/web/header/writers/ClearSiteDataHeaderWriter.java
```diff
@@ -74,11 +74,12 @@ public void writeHeaders(HttpServletRequest request, HttpServletResponse respons
 		if (this.requestMatcher.matches(request)) {
 			if (!response.containsHeader(CLEAR_SITE_DATA_HEADER)) {
 				response.setHeader(CLEAR_SITE_DATA_HEADER, this.headerValue);
-			}
+			}			
+		} else {
```
jzheaux, line 78: Instead of adding an `else` clause, will you please add a `return;` inside of the `if`? This will leave the nesting as-is.

web/src/main/java/org/springframework/security/web/header/writers/ClearSiteDataHeaderWriter.java
```diff
@@ -74,11 +74,12 @@ public void writeHeaders(HttpServletRequest request, HttpServletResponse respons
 		if (this.requestMatcher.matches(request)) {
 			if (!response.containsHeader(CLEAR_SITE_DATA_HEADER)) {
 				response.setHeader(CLEAR_SITE_DATA_HEADER, this.headerValue);
-			}
+			}
```
jzheaux, line 77: Will you please update the copyright header to end in `2025`?

jzheaux (commented):
Good catch, @joaquinjsb! I've left some feedback inline. Also, will you please squash your commits and change the message to something closer to:

```
Log Request Mismatch Only When Mismatches
```

joaquinjsb:
should be good to go now, waiting for tests

joaquinjsb:
tests are failing but doesn't seem to be related to this change?

Issue #17125, opened 2025-05-16
OidcBackChannelLogoutWebFilter make use of a servlet class


Dyskal:
**Describe the bug**
The OidcBackChannelLogoutWebFilter class make use of an undefined status code when the project doesn't have spring-webmvc, and only spring-webflux on its classpath.
https://github.com/spring-projects/spring-security/blob/07a50b460a75d122e808f26fd2aca53c3886c5b8/config/src/main/java/org/springframework/security/config/web/server/OidcBackChannelLogoutWebFilter.java#L21
https://github.com/spring-projects/spring-security/blob/07a50b460a75d122e808f26fd2aca53c3886c5b8/config/src/main/java/org/springframework/security/config/web/server/OidcBackChannelLogoutWebFilter.java#L111

**Expected behavior**
The class should make use of https://github.com/spring-projects/spring-framework/blob/bd83fb7021cab948dea1fee4a65e560a5bb4b4c0/spring-web/src/main/java/org/springframework/http/HttpStatus.java#L157 instead, with  `exchange.getResponse().setStatusCode(HttpStatus.BAD_REQUEST);` .

jzheaux:
Closed in favor of #17128

PR #17128, opened 2025-05-16
Use HttpStatus in back-channel logout filters

therepanic:
A couple of classes in the webflux module used the `jakarta.servlet` module class to specify statuses, which in a way makes us dependent on them and that is wrong. In my fix I used `org.springframework.http.HttpStatus`

Fixes: #17125


jzheaux:
Thanks, @therepanic! This is now merged into `6.3.x`, `6.4.x`, `6.5.x`, and `main`

Issue #12610, opened 2023-02-01
Multiple traces are generated for failing requests


MatthiasDrews:
**Describe the bug**
When a REST request fails, duplicate traces are generated. This behavior exists for authentication errors (e.g. unauthenticated -> 401), as well as for other errors dispatching to the error page (e.g. wrong URL -> 404).
It seems that requests which are dispatched to the error page by the framework, get a fresh observation and thus traceId, instead of taking any existing observation which was generated by the initial request handling.

The result in Grafana looks similar to this:
_example with 404_
![error-traces](https://user-images.githubusercontent.com/16904899/216035618-f8644d23-351a-475e-87a2-affb754b9a19.png)

**To Reproduce**
Fire a request against a secured RestController method, without authentication token.
**or**
Fire a request with a wrong URL which will result in a 404.

**Expected behavior**
A single traceId should be generated, dispatched requests should reuse the original traceId

nkonev:
Hi, @jzheaux, @marcusdacoregio 

Seems I managed to [reproduce](https://github.com/nkonev/trace-discrepancy-reproducer/tree/master) this exactly bug.

to reproduce invoke
```
./mvnw spring-boot:run # or launch AaaApplication.main via IDE

curl -i 'http://localhost:8060/api/aaa/login' -X POST -H 'Accept: application/json, text/plain, */*' -H 'Content-Type: application/x-www-form-urlencoded;charset=utf-8' -H 'X-XSRF-TOKEN: 724f5acd-3d1e-421b-a386-eb17dcacece8' -H 'Cookie: VIDEOCHAT_XSRF_TOKEN=724f5acd-3d1e-421b-a386-eb17dcacece8' --data-raw 'login=admin&password=admin'
```
and check logs (stdout).

Also I noticed that if I comment out `.failureHandler(authenticationFailureHandler)` [here](https://github.com/nkonev/trace-discrepancy-reproducer/blob/master/src/main/java/name/nkonev/aaa/security/SecurityConfig.java#L107) then the bug disappears.

PR #17134, opened 2025-05-18
Fix traceId discrepancy in case error in servlet web

nkonev:
Fixes https://github.com/spring-projects/spring-security/issues/12610

---
As I see from my research, the bug happens only in the real servlet environment, supposedly because of HttpServletResponse.sendError(), I didn't manage to reproduce it via mockMvc.
Tomcat creates a new HttpServletRequest in case error, which is related to "/error" path, which will be served by ErrorController.
Fortunately, request attributes are copied from the original request to the "/error"-related one, so we can put the ObservationView into the original request in order to extract this ObservationView from "/error"-related request and create a new parent observation with a given parent ObservationView. The last ObservationView propagates TraceContext to the new Observation so we have the same traceId.

---
Reproducer is here https://github.com/nkonev/trace-discrepancy-reproducer

```bash
curl -i 'http://localhost:8060/internal/profile/auth' -X GET -H 'Accept: text/plain, */*'
```

To check that the bug is fixed, you can run `publishMavenJavaPublicationToMavenLocal` in spring-security:web
```bash
./gradlew :spring-security-web:publishMavenJavaPublicationToMavenLocal --offline
```
or via IDE
image

then switch to the SNAPSHOT dependency https://github.com/nkonev/trace-discrepancy-reproducer/commit/131a271130726f8332db19d1e286f1debae18ff2

Before:
Screenshot From 2025-05-21 17-45-08
Screenshot From 2025-05-21 17-45-31

After:
Screenshot From 2025-05-21 17-54-26
Screenshot From 2025-05-21 17-54-45

---

UPD:
I also checked this PR with [jetty](https://github.com/nkonev/trace-discrepancy-reproducer/tree/jetty) and [undertow](https://github.com/nkonev/trace-discrepancy-reproducer/tree/undertow), it works fine.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/ObservationFilterChainDecorator.java
+++ b/web/src/main/java/org/springframework/security/web/ObservationFilterChainDecorator.java
@@ -46,13 +46,14 @@ import org.springframework.util.StringUtils;
  * wraps the chain in before and after observations
  *
  * @author Josh Cummings
+ * @author Nikita Konev
  * @since 6.0
  */
 public final class ObservationFilterChainDecorator implements FilterChainProxy.FilterChainDecorator {
 
 	private static final Log logger = LogFactory.getLog(FilterChainProxy.class);
 
-	private static final String ATTRIBUTE = ObservationFilterChainDecorator.class + ".observation";
+	static final String ATTRIBUTE = ObservationFilterChainDecorator.class + ".observation";
 
 	static final String UNSECURED_OBSERVATION_NAME = "spring.security.http.unsecured.requests";
 
@@ -250,6 +251,16 @@ public final class ObservationFilterChainDecorator implements FilterChainProxy.F
 		private AroundFilterObservation parent(HttpServletRequest request) {
 			FilterChainObservationContext beforeContext = FilterChainObservationContext.before();
 			FilterChainObservationContext afterContext = FilterChainObservationContext.after();
+
+			AroundFilterObservation existingParentObservation = (AroundFilterObservation) request
+				.getAttribute(ATTRIBUTE);
+			if (existingParentObservation != null) {
+				beforeContext
+					.setParentObservation(existingParentObservation.before().getContext().getParentObservation());
+				afterContext
+					.setParentObservation(existingParentObservation.after().getContext().getParentObservation());
+			}
+
 			Observation before = Observation.createNotStarted(this.convention, () -> beforeContext, this.registry);
 			Observation after = Observation.createNotStarted(this.convention, () -> afterContext, this.registry);
 			AroundFilterObservation parent = AroundFilterObservation.create(before, after);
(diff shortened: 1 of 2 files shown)


nkonev:
PTAL @jzheaux

nkonev:
@jzheaux 
When you have a chance, I’d appreciate your review or thoughts.

nkonev:
@marcusdacoregio @jzheaux 
please take a look when you have a time

web/src/main/java/org/springframework/security/web/ObservationFilterChainDecorator.java
```diff
@@ -250,9 +258,30 @@ private void wrapFilter(ServletRequest request, ServletResponse response, Filter
 		private AroundFilterObservation parent(HttpServletRequest request) {
 			FilterChainObservationContext beforeContext = FilterChainObservationContext.before();
 			FilterChainObservationContext afterContext = FilterChainObservationContext.after();
+
+			Object maybeBeforeObservationView = request.getAttribute(ATTRIBUTE_BEFORE_OBSERVATION_VIEW);
```
jzheaux: Given that the parent observations are available through the `AroundFilterObservation` instance, it seems like the `ATTRIBUTE` attribute should be sufficient.

For example, instead of two other attributes, would this give you what you need:

```java
AroundFilterObservation parent = (AroundFilterObservation) request.getAttribute(ATTRIBUTE);
if (parent != null) {
    beforeContext.setParentObservation(parent.before().getContext().getParentObservation());
    afterContext.setParentObservation(parent.after().getContext().getParentObservation());
}
```

Note that I tried this modification locally and the traceId is preserved in your project.

nkonev:
Thank you for review @jzheaux !

> That may mean dispatching to ERROR manually in the test.

Can you please explain it a bit more ? I still don't understand for what I need to write an assertion ? 

May be somewhere there is a similar test which I can copy and modify for this particular issue ?

Did you mean to use MockMvc ?

---
In the time when I wrote this PR I considered two options - to make the test in `itest/web/src/integration-test/java/org/springframework/security/integration` folder or in `web/src/test/java/org/springframework/security/web/server/ObservationWebFilterChainDecoratorTests.java` file. I didn't manage to decide which is better for this particular bug.

I considered to make the following test:
Create request 1
then somehow fail it
then manually copy attributes to the request 2 (/error-related) -- to simulate servlet container
then assert the tracing contexts has the similar traceId.

I thought that approach like this is too complex, too artificial and it seems I would tested the mock's behavior.

Do you mean to make the test like this ? Or maybe something different ?

nkonev:
Thank you for review @jzheaux !

UPD: I managed to make the test in `web/src/test/java/org/springframework/security/web/FilterChainProxyTests.java` looking similar to other tests in this file.

nkonev:
Hi, just noticed that this PR was moved to the next milestone.

Just want to ensure that all is OK and it won't be abandoned.

@jzheaux @rwinch 

P. S. sorry for bothering if my concerns are wrong

rwinch:
It's still planned. It's just that our releases are time bound and we didn't have time to review and merge so we pushed it back

nkonev:
@jzheaux Let's do another attempt to merge this PR =)

I'm ready to answer questions and to make necessary edits if need.

jzheaux (approved):

jzheaux:
Thanks for your patience, @nkonev. This is now merged into `6.4.x`, `6.5.x`, and `main`.

PR #17141, opened 2025-05-19
Advise overriding equals() and hashCode() in UserDetails implementations

Gurunathan16:
This commit adds a documentation note explaining the importance of overriding equals() and hashCode() in custom UserDetails implementations.

The default SessionRegistryImpl in Spring Security uses an in-memory ConcurrentMap<Object, Set<String>>, Map<String,SessionInformation> to associate principals with sessions. If a custom UserDetails class does not properly override equals() and hashCode(), user sessions may not be tracked or matched correctly.

I believe this helps developers avoid subtle session management issues when implementing custom authentication logic.

Reference: org.springframework.security.core.session.SessionRegistryImpl


jzheaux:
Thanks, @Gurunathan16! This is now merged into `6.3.x`, `6.4.x`, `6.5.x`, and `main`

Gurunathan16:
Thanks a lot for reviewing and merging this, @jzheaux ! 🙌  
This was my first open source contribution, and I really appreciate the guidance and support from the Spring team.  
Looking forward to contributing more in the future!

Issue #17155, opened 2025-05-21
WebSecurity.ignoring().anyRequest() no longer works


crmky:
**Describe the bug**

In Spring Security 6.3.x, we can use `WebSecurityCustomizer` to customize `WebSecurity` to ignore all security checks. This is not for production, but useful for local testing when some properties are set. A simplified version:

```java
@Configuration(proxyBeanMethods = false)
public class ApplicationConfiguration {

    @Bean
    WebSecurityCustomizer ignoreAllCustomizer() {
        return web -> web.ignoring().anyRequest();
    }
}

```

However because of https://github.com/spring-projects/spring-security/issues/15220 and https://github.com/spring-projects/spring-security/issues/15982, ignore all requests doesn't work anymore. Below exception will be thrown:

```
Caused by: org.springframework.security.web.UnreachableFilterChainException: A filter chain that matches any request [DefaultSecurityFilterChain matching [any request] and having filters []] has already been configured, which means that this filter chain [DefaultSecurityFilterChain defined as 'managementSecurityFilterChain' in [class path resource [org/springframework/boot/actuate/autoconfigure/security/servlet/ManagementWebSecurityAutoConfiguration.class]] matching [any request] and having filters [DisableEncodeUrl, WebAsyncManagerIntegration, SecurityContextHolder, HeaderWriter, Cors, Csrf, Logout, UsernamePasswordAuthentication, DefaultResources, DefaultLoginPageGenerating, DefaultLogoutPageGenerating, BasicAuthentication, RequestCacheAware, SecurityContextHolderAwareRequest, AnonymousAuthentication, ExceptionTranslation, Authorization]] will never get invoked. Please use `HttpSecurity#securityMatcher` to ensure that there is only one filter chain configured for 'any request' and that the 'any request' filter chain is published last.
```

The cause of this issue is `WebSecurityFilterChainValidator` now validate `WebSecurity` `ignoredRequests` together with `HttpSecurity` filter chain together. It will throw the above exception in that case.

**To Reproduce**

Create a Configuration class like the one mentioned above. 

**Expected behavior**

I think `WebSecurity.ignoring().anyRequest()` should still get supported. I understand this is not for the production, but it provides a way to skip all the customizations in `HttpSecurity`. 

**Sample**

A link to a GitHub repository with a [minimal, reproducible sample](https://stackoverflow.com/help/minimal-reproducible-example).

Reports that include a sample will take priority over reports that do not.
[...]

PR #17158, opened 2025-05-22
Do not validate ignoredRequests from WebSecurity.ignoring()

kse-music:
Closes gh-17155


jzheaux:
Hi, @kse-music, I'm not sure that we want to make an exception for the web-ignoring configuration. If such a configuration were released into production, there would be no security and no one would get notified. It appears to me that the error message in #17155 is correct; I think it's a matter of finding a better way to turn off security during development.

I'm going to close this PR for now while discussion continues in the ticket. We can always reopen it if this route is revisited.

Issue #17132, opened 2025-05-18
Ability to disable anonymous authentication in RSocketSecurity


Aaur1s:
**Expected Behavior**

RSocketSecurity dsl should have ability to disable anonymous auth interceptor

**Current Behavior**

RSocketSecurity has hardcode in private method that adds anonymous interceptor no matter what

**Context**

I'm doing method-level security and rely on @PreAuthorize("authenticated"), my global security config has just permitAll. So anonymous authentication ruining my rsocket experience, for http security i can just disable it. 
As workaround i constructed PayloadSocketAcceptorInterceptor entirely by hand, but this process is quite tedious, it would be nice to be able to disable it in RSocketSecurity dsl like in http.

therepanic:
If no one minds I can work on this, can I @jzheaux?

mehrdadbozorgmehr:
Can I work on this issue? Please assign it to me.

therepanic:
Hi, @mehrdadbozorgmehr! You may have noticed that I attached the PR earlier and am already working on it. You can follow the process at https://github.com/spring-projects/spring-security/pull/17159.

jzheaux:
Thanks for your interest, @mehrdadbozorgmehr. We'd love to have you contribute on another ticket!

I'll close this in favor of https://github.com/spring-projects/spring-security/pull/17159 to alleviate confusion.

PR #17159, opened 2025-05-22
Add option to disable anonymous authentication in `RSocketSecurity`

therepanic:
Adding anonymous interceptor hardcoded to `RSocketSecurity#payloadInterceptors` and is added anyway. I think you should add an option to disable adding anonymous interceptor.

I think the most logical way is to add a flag, as in our case I did. I also added a method `disableAnonymous` which sets the field to true and we no longer add the anonymous interceptor.

Fixes: #17132

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/rsocket/RSocketSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/rsocket/RSocketSecurity.java
@@ -171,8 +171,9 @@ public class RSocketSecurity {
 	 * Adds anonymous authentication
 	 * @param anonymous a customizer
 	 * @return this instance
+	 * @since 7.0
 	 */
-	public RSocketSecurity anonymousAuthentication(Customizer<AnonymousAuthenticationSpec> anonymous) {
+	public RSocketSecurity anonymous(Customizer<AnonymousAuthenticationSpec> anonymous) {
 		if (this.anonymousAuthSpec == null) {
 			this.anonymousAuthSpec = new AnonymousAuthenticationSpec(this);
 		}
--- a/config/src/main/java/org/springframework/security/config/annotation/rsocket/RSocketSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/rsocket/RSocketSecurity.java
@@ -109,6 +109,7 @@ import org.springframework.security.rsocket.util.matcher.RoutePayloadExchangeMat
  * @author Manuel Tejeda
  * @author Ebert Toribio
  * @author Ngoc Nhan
+ * @author Andrey Litvitski
  * @since 5.2
  */
 public class RSocketSecurity {
@@ -118,6 +119,8 @@ public class RSocketSecurity {
 	private BasicAuthenticationSpec basicAuthSpec;
 
 	private SimpleAuthenticationSpec simpleAuthSpec;
+
+	private AnonymousAuthenticationSpec anonymousAuthSpec = new AnonymousAuthenticationSpec(this);
 
 	private JwtSpec jwtSpec;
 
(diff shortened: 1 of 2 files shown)


therepanic:
Hmmm, is the build going down by my mistake or is it not about me?

config/src/main/java/org/springframework/security/config/annotation/rsocket/RSocketSecurity.java
```diff
@@ -179,6 +182,15 @@ public RSocketSecurity basicAuthentication(Customizer<BasicAuthenticationSpec> b
 		return this;
 	}
 
+	/**
+	 * Disables anonymous authentication.
+	 * @return the builder for additional customizations
+	 */
+	public RSocketSecurity disableAnonymous() {
```
jzheaux: While I agree that this is a straightforward way to expose the feature, I see some additional benefit in exposing it in a way that's similar to `HttpSecurity` and `ServerHttpSecurity.` Are you able to add this feature [in the following way](https://github.com/spring-projects/spring-security/blob/main/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java#L568) instead?

jzheaux (changes requested):
Thanks for the PR, @therepanic! I've left feedback inline.

Also, are you able to add a test that confirms that when anonymous is disabled then it isn't active in the filter chain?

therepanic:
Thanks for review, @jzheaux! I'd be happy to add the test. I just have one question about it.

I think we could make a test that checks the functionality you described by creating a new test class, let's call it `AnonymousAuthenticationITests` (similar to `SimpleAuthenticationITests`). And, similar to how `SimpleAuthenticationITests` works, we will check the functionality of disabling the addition of the anonymous auth interceptor. I think this is okay, what do you think? Or is there an easier way?

config/src/main/java/org/springframework/security/config/annotation/rsocket/RSocketSecurity.java
therepanic: I agree, that would indeed be better. I think we should add a class similar to `RSocketSecurity.BasicAuthenticationSpec`, but without fields, and the build() method will execute the logic of the current anonymous() method. And I think we can remove it as unnecessary.

jzheaux:
@therepanic, that sounds like a good approach. Also, you might consider a test like `AnonymousConfigurerTests#requestWhenAnonymousDisabledInLambdaThenRespondsWithForbidden` that authorizes only anonymous users. When `anonymous` is disabled, the requests in that test should fail.

`RSocketSecurity` doesn't expose the `anonymous()` access rule, but you can use `access(ReactiveAuthorizationManager)` to provide one:

```java
AuthenticationTrustResolver trustResolver = new AuthenticationTrustResolverImpl();
ReactiveAuthorizationManager<PayloadExchange> anonymous = (authentication, exchange) -> 
    authentication.map(trustResolver::isAnonymous).map(AuthorizationDecision::new);
// ...

.anyExchange().access(anonymous)
```

config/src/integration-test/java/org/springframework/security/config/annotation/rsocket/RSocketMessageHandlerITests.java
```diff
@@ -268,7 +268,8 @@ PayloadSocketAcceptorInterceptor rsocketInterceptor(RSocketSecurity rsocket) {
 					.route("secure.*").authenticated()
 					.anyExchange().permitAll()
 				)
-				.basicAuthentication(Customizer.withDefaults());
+				.basicAuthentication(Customizer.withDefaults())
+					.anonymousAuthentication(Customizer.withDefaults());
```
therepanic: I believe this test assumed AnonymousAuthInterceptor was enabled; without it, it fails with our new changes.

jzheaux: Anonymous should continue to be enabled by default. Disabling is what should be optional.

Please set `anonymousAuthSpec` to a default value and add a `disable` method to the `AnonymousAuthSpec` class so that someone can do:

```java
// ...
.basicAuthentication(Customizer.withDefaults())
// ... anonymous is already active
```

or

```java
// ...
.basicAuthentication(Customizer.withDefaults())
.anonymous((anonymous) -> anonymous.disable())
```


The `disable` method will set `anonymousAuthSpec` to `null`.

therepanic:
Thanks for review @jzheaux again.

Yes, you are right, I did make a mistake. In a separate commit, for your convenience, I added the creation of an anonymousAuthSpec instance by default, and also added a disable method to `AnonymousAuthenticationSpec`. In addition, I rolled back all changes in `RSocketMessageHandlerITests` (as unnecessary), and in our `AnonymousAuthenticationITests#requestWhenAnonymousDisabledThenRespondsWithForbidden` test, we manually disable anonymous authentication using the disable method.

jzheaux (approved):

jzheaux:
Thanks for your help with this, @therepanic! This will merge to `main` once the build completes.

Issue #17173, opened 2025-05-27
DPoP filter is ignored when another AuthenticationFilter is present


gbaso:
DPoP filter is implemented as a an `AuthenticationConverter` applied to an `AuthenticationFilter`. `AuthenticationFilter` extends `OncePerRequestFilter`, which means that only the first filter of that type actually do its filtering. An attribute then is added to the request that prevents additional filters of the same type to do anything other than proceeding to the next filter.

In my experience, `AuthenticationFilter` plus `AuthenticationConverter` is the easiest way to create a custom filter, therefore I believe it to be very commonly used. Depending on the configuration, either DPoP or the custom filter is completely disabled, based on the filter ordering.

Sample application with oauth2 resource server + custom API key: https://github.com/gbaso/oauth2-api-server

jgrandja:
@gbaso Thanks for reporting this. Indeed this is an issue if there is more than one `AuthenticationFilter` in the chain.

Overriding `OncePerRequestFilter.getAlreadyFilteredAttributeName()` would solve this issue.

Would you be interested in submitting a fix?

therepanic:
If no one would mind, I could prepare a fix.

gbaso:
@therepanic you can go ahead

jgrandja:
@gbaso This is now resolved via gh-17216. I'll mark this as a duplicate.

PR #17186, opened 2025-05-29
Fix to allow multiple AuthenticationFilter instances to process each request

therepanic:
This change will enable us to use multiple `AuthenticationFilter`.

Fix: #17173

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/AuthenticationFilter.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/AuthenticationFilter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -64,6 +64,7 @@ import org.springframework.web.filter.OncePerRequestFilter;
  * </ul>
  *
  * @author Sergey Bespalov
+ * @author Andrey Litvitski
  * @since 5.2.0
  */
 public class AuthenticationFilter extends OncePerRequestFilter {
@@ -220,6 +221,15 @@ public class AuthenticationFilter extends OncePerRequestFilter {
 			throw new ServletException("AuthenticationManager should not return null Authentication object.");
 		}
 		return authenticationResult;
+	}
+
+	@Override
+	protected String getAlreadyFilteredAttributeName() {
+		String name = getFilterName();
+		if (name == null) {
+			name = getClass().getName().concat("-" + System.identityHashCode(this));
+		}
+		return name + ALREADY_FILTERED_SUFFIX;
 	}
 
 }
--- a/web/src/test/java/org/springframework/security/web/authentication/AuthenticationFilterTests.java
+++ b/web/src/test/java/org/springframework/security/web/authentication/AuthenticationFilterTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,6 +17,7 @@
 package org.springframework.security.web.authentication;
 
 import jakarta.servlet.FilterChain;
+import jakarta.servlet.Servlet;
 import jakarta.servlet.ServletException;
 import jakarta.servlet.ServletRequest;
 import jakarta.servlet.ServletResponse;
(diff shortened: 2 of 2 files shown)


web/src/main/java/org/springframework/security/web/authentication/AuthenticationFilter.java
```diff
@@ -222,4 +223,9 @@ private Authentication attemptAuthentication(HttpServletRequest request, HttpSer
 		return authenticationResult;
 	}
 
+	@Override
+	protected String getAlreadyFilteredAttributeName() {
+		return getClass().getName() + ".FILTERED";
```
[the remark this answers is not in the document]
therepanic: I apologize for not accurately solving the problem. We do want each `AuthenticationFilter` instance to be honored in the chain. What I did I guess won't work if the filters are the same and added to the chain. Then I guess we need to use something like `identityHashCode` for `getAlreadyFilteredAttributeName()`? Thanks.

web/src/main/java/org/springframework/security/web/authentication/AuthenticationFilter.java
```diff
@@ -222,4 +223,9 @@ private Authentication attemptAuthentication(HttpServletRequest request, HttpSer
 		return authenticationResult;
 	}
 
+	@Override
+	protected String getAlreadyFilteredAttributeName() {
```
jgrandja, line 227: I just realized that this is a public API change and therefore cannot be applied in `6.5.1` release. We'll need to address this issue with `AuthenticationFilter` in a separate issue targeted for `7.0`.

For now, we need to address gh-17173 and I think the easiest solution is to define the following in `DPoPAuthenticationConfigurer`:

```
private static final class DPoPAuthenticationFilter extends AuthenticationFilter {

	private DPoPAuthenticationFilter(AuthenticationManager authenticationManager, AuthenticationConverter authenticationConverter) {
		super(authenticationManager, authenticationConverter);
	}

}
```

I think this should solve the bug.

jgrandja: @therepanic Please hold off on the proposed change in my previous comment. The root issue is in `AuthenticationFilter` and we still need to address it as a bug. We might still be able to get your change into `6.5.1` but let me think about this for a bit.

therepanic: Thanks for the advice. You want me to create a separate issue and link the PR to it?

jgrandja: @therepanic See [comment](https://github.com/spring-projects/spring-security/pull/17186#discussion_r2123320045)

therepanic:
I believe that it is #17173 that we have now solved. However I also understand the problem that I solved in the last commit as you said for merge is now unavailable? What should we do about it now?

jgrandja:
@therepanic Maybe you missed my last [comment](https://github.com/spring-projects/spring-security/pull/17186#discussion_r2123320045)?

We should fix this in `AuthenticationFilter` as it's a bug. Please go back to previous commit and override `getAlreadyFilteredAttributeName()` as follows:

```java
@Override
protected String getAlreadyFilteredAttributeName() {
	String name = getFilterName();
	if (name == null) {
		name = AuthenticationFilter.class.getSimpleName() + "." +
				getAuthenticationConverter().getClass().getSimpleName();
	}
	return name + ALREADY_FILTERED_SUFFIX;
}
```

We'll need to get this fix into all supported OSS branches so please rebase the fix off of `6.3.x`.

therepanic:
I apologize for the misunderstanding. I really didn't see that you asked in the next comment to leave it as is.

You posted how it should be overrided and I have a question. Will this work at all? In my last commit I added a test like this:
```java
@Test
public void filterWhenInFilterChainThenBothAreExecuted() throws Exception {
	MockHttpServletRequest request = new MockHttpServletRequest(“GET”, “/”);
	MockHttpServletResponse response = new MockHttpServletResponse();
	AuthenticationFilter filter1 = new AuthenticationFilter(this.authenticationManager,
			this.authenticationConverter);
	AuthenticationConverter converter2 = mock(AuthenticationConverter.class);
	AuthenticationFilter filter2 = new AuthenticationFilter(this.authenticationManager, converter2);
	FilterChain chain = new MockFilterChain(mock(Servlet.class), filter1, filter2);
	verify(this.authenticationConverter).convert(any());
	verify(converter2).convert(any());
}
```

Tell me if it is incorrect. However, if it is correct, then the test will fail with the solution you submitted.

jgrandja:
You are right the test will fail because of how Mockito names the mocks.

Let's change the implementation as follows:

```java
@Override
protected String getAlreadyFilteredAttributeName() {
	String name = getFilterName();
	if (name == null) {
		name = getClass().getName().concat("-" + System.identityHashCode(this));
	}
	return name + ALREADY_FILTERED_SUFFIX;
}
```

therepanic:
Thanks, my previous implementation in the last commit used `System.identityHashCode` as well. I reverted everything and now I think we are fine.

jgrandja:
Thanks for the updates @therepanic. This is now merged.

PR #17190, opened 2025-06-01
Fix users schema documentation

Lidoca:
docs: match the database schema with https://github.com/spring-projects/spring-security/blob/6.5.0/docs/modules/ROOT/pages/servlet/authentication/passwords/jdbc.adoc


rwinch:
Thank you for the pull request. This is not scheduled to be merged automatically when the build passes.

PR #17191, opened 2025-06-01
Support custom `OAuth2AuthenticatedPrincipal` in Jwt-based authentication flow

therepanic:
This PR implements a simpler approach, as suggested by @jzheaux, to support `OAuth2AuthenticatedPrincipal` injection into `JwtAuthenticationToken`

Resolves: #6237

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationConverter.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationConverter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2022 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,6 +21,7 @@ import java.util.Collection;
 import org.springframework.core.convert.converter.Converter;
 import org.springframework.security.authentication.AbstractAuthenticationToken;
 import org.springframework.security.core.GrantedAuthority;
+import org.springframework.security.oauth2.core.OAuth2AuthenticatedPrincipal;
 import org.springframework.security.oauth2.jwt.Jwt;
 import org.springframework.security.oauth2.jwt.JwtClaimNames;
 import org.springframework.util.Assert;
@@ -30,10 +31,12 @@ import org.springframework.util.Assert;
  * @author Josh Cummings
  * @author Evgeniy Cheban
  * @author Olivier Antoine
+ * @author Andrey Litvitski
  * @since 5.1
  */
 public class JwtAuthenticationConverter implements Converter<Jwt, AbstractAuthenticationToken> {
 
+	private Converter<Jwt, OAuth2AuthenticatedPrincipal> jwtPrincipalConverter;
 	private Converter<Jwt, Collection<GrantedAuthority>> jwtGrantedAuthoritiesConverter = new JwtGrantedAuthoritiesConverter();
 
 	private String principalClaimName = JwtClaimNames.SUB;
@@ -42,8 +45,26 @@ public class JwtAuthenticationConverter implements Converter<Jwt, AbstractAuthen
 	public final AbstractAuthenticationToken convert(Jwt jwt) {
 		Collection<GrantedAuthority> authorities = this.jwtGrantedAuthoritiesConverter.convert(jwt);
 
-		String principalClaimValue = jwt.getClaimAsString(this.principalClaimName);
-		return new JwtAuthenticationToken(jwt, authorities, principalClaimValue);
+		if (this.jwtPrincipalConverter == null) {
+			String principalClaimValue = jwt.getClaimAsString(this.principalClaimName);
+			return new JwtAuthenticationToken(jwt, authorities, principalClaimValue);
+		} else {
+			OAuth2AuthenticatedPrincipal principal = this.jwtPrincipalConverter.convert(jwt);
+			authorities.addAll(principal.getAuthorities());
+			return new JwtAuthenticationToken(jwt, principal, authorities);
+		}
+	}
+
+	/**
+	 * Sets the {@link Converter Converter&lt;Jwt, Collection&lt;OAuth2AuthenticatedPrincipal&gt;&gt;}
+	 * to use.
+	 * @param jwtPrincipalConverter The converter
+	 * @since 6.5.0
+	 */
+	public void setJwtPrincipalConverter(
+			Converter<Jwt, OAuth2AuthenticatedPrincipal> jwtPrincipalConverter) {
+		Assert.notNull(jwtPrincipalConverter, "jwtPrincipalConverter cannot be null");
+		this.jwtPrincipalConverter = jwtPrincipalConverter;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


therepanic:
I think the current solution can't be merged now since at the very least we don't have tests. I would like to hear feedback if I have understood the solution to the current problem correctly.

Also I guess we can't add a constructor to `(Jwt jwt, Object principal, Collection<GrantedAuthority> authorities)` as written in [#32](https://github.com/spring-projects/spring-security/issues/6237#issuecomment-2913474083) because we already have a constructor with 3 arguments `(Jwt jwt, Collection<? extends GrantedAuthority> authorities, String name)`. Because of this we have a build failing now.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationConverter.java
```diff
@@ -42,8 +45,26 @@ public class JwtAuthenticationConverter implements Converter<Jwt, AbstractAuthen
...
+			authorities.addAll(principal.getAuthorities());
+			return new JwtAuthenticationToken(jwt, principal, authorities);
+		}
+	}
+
+	/**
+	 * Sets the {@link Converter Converter&lt;Jwt, Collection&lt;OAuth2AuthenticatedPrincipal&gt;&gt;}
+	 * to use.
+	 * @param jwtPrincipalConverter The converter
+	 * @since 6.5.0
```
jzheaux: This should be `@since 7.0` as that's the current feature release (now that 6.5 is released, it will only take bug fixes).

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationConverter.java
```diff
@@ -42,8 +45,26 @@ public class JwtAuthenticationConverter implements Converter<Jwt, AbstractAuthen
 	public final AbstractAuthenticationToken convert(Jwt jwt) {
 		Collection<GrantedAuthority> authorities = this.jwtGrantedAuthoritiesConverter.convert(jwt);
 
-		String principalClaimValue = jwt.getClaimAsString(this.principalClaimName);
-		return new JwtAuthenticationToken(jwt, authorities, principalClaimValue);
+		if (this.jwtPrincipalConverter == null) {
```
jzheaux: With the help of a private inner class, I believe we can adapt `Jwt` into an `OAuth2AuthenticatedPrincipal` so that this if statement is not necessary. Consider adding a class like this:

```java
private static final class JwtAuthenticatedPrincipal extends Jwt implements OAuth2AuthenticatedPrincipal {

	private final String principalClaimName;

	JwtAuthenticatedPrincipal(Jwt jwt) {
		this(jwt, JwtClaimNames.SUB);
	}

	JwtAuthenticatedPrincipal(Jwt jwt, String principalClaimName) {
		super(jwt.getTokenValue(), jwt.getIssuedAt(), jwt.getExpiresAt(), jwt.getHeaders(), jwt.getClaims());
		this.principalClaimName = principalClaimName;
	}

	@Override
	public Map<String, Object> getAttributes() {
		return getClaims();
	}

	@Override
	public Collection<? extends GrantedAuthority> getAuthorities() {
		return List.of();
	}

	@Override
	public String getName() {
		return getClaimAsString(this.principalClaimName);
	}

}
```

Then, I believe `jwtPrincipalConverter` can default to `JwtAuthenticatedPrincipal::new` and `principalClaimName` can be removed by having `setPrincipalClaimName` use `JwtAuthenticatedPrincipal` as well.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtBearerTokenAuthenticationConverter.java
```diff
@@ -58,4 +60,16 @@ public AbstractAuthenticationToken convert(Jwt jwt) {
 		return new BearerTokenAuthentication(principal, accessToken, authorities);
```
jzheaux, line 61: This should use the principal provided by the principal conversion instead of constructing a new `DefaultOAuth2AuthenticatedPrincipal`.

It might be a little simpler at this point to change this class to have its own `Converter<Jwt, OAuth2AuthenticatedPrincipal>` and `Converter<Jwt, Collection<GrantedAuthority>>` references.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtBearerTokenAuthenticationConverter.java
```diff
@@ -58,4 +60,16 @@ public AbstractAuthenticationToken convert(Jwt jwt) {
...
 	}
 
+	/**
+	 * Sets the {@link Converter Converter&lt;Jwt, Collection&lt;OAuth2AuthenticatedPrincipal&gt;&gt;}
+	 * to use.
+	 * @param jwtPrincipalConverter The converter
+	 * @since 6.5.0
```
jzheaux: Will you please update this to `@since 7.0`?

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtBearerTokenAuthenticationConverter.java
```diff
@@ -58,4 +60,16 @@ public AbstractAuthenticationToken convert(Jwt jwt) {
 		return new BearerTokenAuthentication(principal, accessToken, authorities);
 	}
 
+	/**
+	 * Sets the {@link Converter Converter&lt;Jwt, Collection&lt;OAuth2AuthenticatedPrincipal&gt;&gt;}
+	 * to use.
```
jzheaux: Will you please state the default value as well? For example:

```java
* <p>By default, constructs a {@link DefaultOAuth2AuthenticatedPrincipal} based on the claims and authorities derived from the {@link Jwt}.
```

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtBearerTokenAuthenticationConverter.java
therepanic: Adding your own links to converters will indeed be easier, but I have a question. Do we want to create the same `jwtPrincipalConverter` in `JwtBearerTokenAuthenticationConverter` as in `JwtAuthenticationConverter`? That is, by default with the type `JwtAuthenticatedPrincipal::new`? If so, then in the version you suggested, this class is private, and therefore we will not be able to create a converter by default. Should we make `JwtAuthenticatedPrincipal` package-private? Or we can just create getters for converters in `JwtAuthenticationConverter` and use them?

jzheaux: I'd recommend that the principal converter in `JwtBearerTokenAuthenticationConverter` be one that creates a `DefaultOAuth2AuthenticatedPrincipal` in order to preserve existing behavior.

And I'm not sure, but I believe that may require calling the authorities converter twice, once to populate `DefaultOAuth2AuthenticatedPrincipal` and once to populate `BearerTokenAuthentication`; however, I don't think that's a big issue as the point is to allow the strategy to be overridden.

jzheaux:
Thanks for getting this PR started, @therepanic! I was able to add the needed tests and documentation. It will merge into `main` once the build completes.

Issue #17194, opened 2025-06-02
Allow customization of `WebClient` used by `SpringReactiveOpaqueTokenIntrospector`


pinguet62:
**Expected Behavior**

Allow custimization of [`WebClient` used by `SpringReactiveOpaqueTokenIntrospector`](https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringReactiveOpaqueTokenIntrospector.java#L69).

**Current Behavior**

[`SpringReactiveOpaqueTokenIntrospector` uses a basic `WebClient`](https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringReactiveOpaqueTokenIntrospector.java#L317): 
```java
WebClient.builder() /*setBasicAuth*/ .build()
```

**Context**

I have to add some behaviors to its `WebClient`: retry, metrics, ...

Previously I used [the `SpringReactiveOpaqueTokenIntrospector(String introspectionUri, WebClient webClient)` constructor](https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringReactiveOpaqueTokenIntrospector.java#L273).
But now it's deprecated...

I think it can be done easily by adding a `withWebClientBuilder(WebClient.Builder webClientBuilder)` to [its `Builder`](https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringReactiveOpaqueTokenIntrospector.java#L273).

Please let me know if it's a good idead!?

jzheaux:
Closing in favor of #17196

PR #17196, opened 2025-06-02
Make `WebClient.Builder` configurable in `SpringReactiveOpaqueTokenIntrospector`

therepanic:
This will help us to throw our own `WebClient.Builder` into `SpringReactiveOpaqueTokenIntrospector.Builder` with our own parameters that we specify.

Also for this class we have one test where a builder is used. I decided to just throw `WebClient.Builder` in there in order to use the method, I think it doesn't make much sense to test it, wdyt?

Resolves: #17194

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringReactiveOpaqueTokenIntrospector.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringReactiveOpaqueTokenIntrospector.java
@@ -268,6 +268,7 @@ public class SpringReactiveOpaqueTokenIntrospector implements ReactiveOpaqueToke
 	 * Used to build {@link SpringReactiveOpaqueTokenIntrospector}.
 	 *
 	 * @author Ngoc Nhan
+	 * @author Andrey Litvitski
 	 * @since 6.5
 	 */
 	public static final class Builder {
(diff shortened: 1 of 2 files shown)


jzheaux:
Thanks for the suggestion, @therepanic. I'm inclined to leave the class as-is since you can do:

```java
@Bean 
ReactiveOpaqueTokenIntrospector opaqueToken() {
  WebClient.Builder builder = WebClient.builder()
      .defaultHeaders((headers) -> headers.setBasicAuth(clientId, clientSecret))
      // ... your settings
  return new SpringReactiveOpaqueTokenIntrospector(introspectionUri, builder.build());
}
```

Adding `WebClient.Builder` to the builder doesn't seem to add much:

```java
@Bean 
ReactiveOpaqueTokenIntrospector opaqueToken() {
  WebClient.Builder builder = WebClient.builder()
      // ... your settings
  return SpringReactiveOpaqueTokenIntrospector.withIntrospectionUri(introspectionUri)
    .clientId(clientId)..clientSecret(clientSecret)
    .build();
}
```

Will it work for you to construct the `WebClient` and then pass that into the constructor?

therepanic:
Yes, I think that will work.

jzheaux:
Great, @therepanic. I'll close this for now, then. We can always revisit if something comes up.

pinguet62:
Hello 🖖 @jzheaux

> you can do
```java
new SpringReactiveOpaqueTokenIntrospector(introspectionUri, builder.build())
```

Yes, but
* in this case the header building (`setBasicAuth()`) is **delegated to caller**
    while in the actual solution it's responsability of `SpringReactiveOpaqueTokenIntrospector` & it's builder.
* this constructor is deprecated, isn't it?
    Probably for removable?

PR #17197, opened 2025-06-02
Fix inconsistent constructor declaration for `ReactiveAuthorizationManagerMethodSecurityConfiguration`

damable-nuvolex:
Closes gh-16325

A [previous solution](https://github.com/spring-projects/spring-security/pull/16326) was rejected where it was stated the `expressionHandler` parameter should remain optional.

This alternative allows an optional `expressionHandler` without producing the inconsistent constructor declaration log.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/method/configuration/ReactiveAuthorizationManagerMethodSecurityConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/method/configuration/ReactiveAuthorizationManagerMethodSecurityConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2024 the original author or authors.
+ * Copyright 2002-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


config/src/main/java/org/springframework/security/config/annotation/method/configuration/ReactiveAuthorizationManagerMethodSecurityConfiguration.java
```diff
@@ -83,15 +85,16 @@ final class ReactiveAuthorizationManagerMethodSecurityConfiguration
 
 	private final AuthorizationManagerAfterReactiveMethodInterceptor postAuthorizeMethodInterceptor;
 
-	@Autowired(required = false)
-	ReactiveAuthorizationManagerMethodSecurityConfiguration(MethodSecurityExpressionHandler expressionHandler,
+	@Autowired
+	ReactiveAuthorizationManagerMethodSecurityConfiguration(Optional<MethodSecurityExpressionHandler> expressionHandler,
```
jzheaux: Let's please use `ObjectProvider` as this is semantically aligned with the other dependencies. Then you can do:

```java
MethodSecurityExpressionHandler expressionHandler = expressionHandlers.getIfUnique();
if (expressionHandler != null) {
    // set the values
}
```

config/src/main/java/org/springframework/security/config/annotation/method/configuration/ReactiveAuthorizationManagerMethodSecurityConfiguration.java
```diff
@@ -83,15 +85,16 @@ final class ReactiveAuthorizationManagerMethodSecurityConfiguration
 
 	private final AuthorizationManagerAfterReactiveMethodInterceptor postAuthorizeMethodInterceptor;
 
-	@Autowired(required = false)
-	ReactiveAuthorizationManagerMethodSecurityConfiguration(MethodSecurityExpressionHandler expressionHandler,
+	@Autowired
```
jzheaux: This annotation isn't necessary. As the previous PR alluded to, the annotation was likely misplaced originally.

jzheaux (changes requested):
Thanks, @damable-nuvolex! I've left my feedback inline.

damable-nuvolex:
@jzheaux thanks for the feedback, included the changes in [8e3f490](https://github.com/spring-projects/spring-security/pull/17197/commits/8e3f490402bcf4dd712bf0ea72ebbce04b82890f)

Issue #16940, opened 2025-04-15
Add support dpop customization


franticticktick:
Currently `dpop` can't be configured. `OAuth2ResourceServerConfigurer` has `dPoPAuthenticationConfigurer` and it is already initialized with `DPoPAuthenticationConfigurer` and in `configure` method it is always applied to `http`. It would be nice to separate dpop configuration instead of applying it by default.
For example:
```
		@Bean
		SecurityFilterChain securityFilterChain(HttpSecurity http) throws Exception {
			// @formatter:off
			http
				.authorizeHttpRequests((authorize) ->
					authorize.anyRequest().authenticated()
				)
				.oauth2ResourceServer((oauth2ResourceServer) ->
					oauth2ResourceServer
						.jwt(Customizer.withDefaults())
						.dpop(Customizer.withDefaults())
				);
			// @formatter:on
			return http.build();
		}
```

jgrandja:
@franticticktick Can you provide some specific use cases on what you need to customize for a DPoP flow?

franticticktick:
@jgrandja Firstly, it would be nice if dpop was optional, now it is always enabled. Secondly, the same `Authorization` header is specified everywhere, it is standard, but it can be different, which is a normal situation. That is, at least I want to be able to customize `authenticationConverter` and `requestMatcher`. In addition, I may need additional audit of successful authentication and I need a custom `AuthenticationSuccessHandler`.

jgrandja:
Thank you @franticticktick. I've assigned it to you.

No rush as this is scheduled for `7.0.x`.

pheyken:
Hi @franticticktick & @jgrandja, 

while trying out spring-boot 3.5.0 (which includes spring-security 6.5.0) we noticed that our setup does not work anymore due to the auto init of DPoP.

We are using the oauth 2.0 resource server with an OpaqueTokenIntrospector and not JWT, therefore we do not have `spring-security-oauth2-jose ` as a dependency and the classes are not on the class path during runtime.

A ClassNotFoundException (org.springframework.security.oauth2.jwt.JwtException) is thrown during start-up while configuring DPoP.

In the end there are three possible solutions for us:
- we just include `spring-security-oauth2-jose` as a dependency. however, this would be temporary mitigation
- DPoP is only configured in case of JWT being enabled / `spring-security-oauth2-jose` on the class path 
- there is an option to disable DPoP

[...]

PR #17202, opened 2025-06-03
Add Support DPoP Customization

franticticktick:
Closes gh-16940

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/DPoPAuthenticationConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/DPoPAuthenticationConfigurer.java
@@ -16,32 +16,13 @@
 
 package org.springframework.security.config.annotation.web.configurers.oauth2.server.resource;
 
-import java.util.Collections;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-
-import jakarta.servlet.http.HttpServletRequest;
-import jakarta.servlet.http.HttpServletResponse;
-
-import org.springframework.http.HttpHeaders;
-import org.springframework.http.HttpStatus;
 import org.springframework.security.authentication.AuthenticationManager;
 import org.springframework.security.config.annotation.web.HttpSecurityBuilder;
 import org.springframework.security.config.annotation.web.configurers.AbstractHttpConfigurer;
-import org.springframework.security.core.Authentication;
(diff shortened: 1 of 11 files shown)


jgrandja:
Thanks for the PR @franticticktick. I will review this soon.

jgrandja:
Thanks for your patience @franticticktick and apologies that I could not get this into `7.0`. We had quite a workload for this round of majors.

I have scheduled this for `7.1.0-M1` and will review this soon. In the meantime, when you have a moment can you please rebase. Thank you.

renannprado:
@franticticktick @jgrandja I haven't gone through the details of the PR, but I just wanted to add the following:

It's not uncommon for ppl to have a reverse proxy in front of their applications, having a reverse proxy _could mean_ (like in my use case) that you have different _paths_ per application. For example, `https://api.example.com/svc1/*` routes requests to application `svc1` and `https://api.example.com/svc2/*`routes requests to application `svc2`. The problem here is that normally applications would sign the DPoP for `https://api.example.com/svc2/my-api` and not `https://api.example.com/my-api`. So when Spring Security's DPoP validation kicks in, it will look at the URL `https://api.example.com/svc2/my-api` and find no endpoint matching that URL, because the application doesn't know about `/svc2/my-api`. Now our current workaround is that our frontends replace the first path segment of the URL so that they send the request to `/svc2/my-api`, but sign the DPoP for `/my-api`, but IMHO frontends should be aware of the reverse proxy whatsoever.

TL;DR it should be possible to customize the DPoP validation to be reverse-proxy aware. I'm just trying to scratch my itch here, but I do believe this is not an uncommon use case. I'm happy to contribute somehow if needed.

Edit: 

Also, can we also get this into the 6.5.x?

ferrerogg:
> TL;DR it should be possible to customize the DPoP validation to be reverse-proxy aware. I'm just trying to scratch my itch here, but I do believe this is not an uncommon use case. I'm happy to contribute somehow if needed.

I've a similar setup where reverse proxy is Spring Cloud Gateway 5 MVC. You must configure your reverse proxy to add X-Forwarded-xxxx headers (in your case, X-Forwarded-Prefix should report /svc1 or /svc2 to your resource server). 
In my setup, I've configured my resource server with `server.forward-headers-strategy = framework` and it works as expected (see [Running Behind a Front-end Proxy Server](https://docs.spring.io/spring-boot/how-to/webserver.html#howto.webserver.use-behind-a-proxy-server))

renannprado:
@ferrerogg it worked. Thank you so much!

jgrandja:
@franticticktick I went ahead and applied a polish commit so we can get this into `7.1.0-RC1` (April 20). This is now merged.

Issue #17164, opened 2025-05-23
WebAuthn: credential registration fails with an unknown credential type error


ltanguy:
**Describe the bug**
When following the steps described in [the documentation](https://docs.spring.io/spring-security/reference/servlet/authentication/passkeys.html#passkeys-register-create), the credential registration fails with the following exception:
`java.lang.IllegalArgumentException: Cannot convert unknown credential type org.springframework.security.web.webauthn.api.PublicKeyCredentialType@4cfbb790 to webauthn4j
`

**Expected behavior**
The problem seems to originate from [this line](https://github.com/spring-projects/spring-security/blob/main/web/src/main/java/org/springframework/security/web/webauthn/management/Webauthn4JRelyingPartyOperations.java#L307).
The debugger indicates that the type of the request I generate is indeed "public-key", but as the two object instances are different this test fails.
The error message could be clearer by referencing the value of the PublicKeyCredentialType object

ngocnhan-tran1996:
Could you please provide a sample?

ltanguy-s4a:
As a side note, PublicKeyCredentialCreationOptions is not declared as Serializable (contrary to the other classes in this package). This makes this authentication method not compatible with a vanilla Spring Session installation that relies on the default Java object serializer.

re1709:
Finding the same issue regarding PublicKeyCredentialCreationOptions not being serializable as [@Itang](https://github.com/ltanguy)

We are also using Spring Session JDBC, but haven't got as far as seeing the parameter type check issue (halted by the serialization issue) but likely would see that next.

Is there any update on this and the PR being reviewed? WebAuthN is not currently usable with Spring JDBC Session.

PR #17223, opened 2025-06-10
WebAuthn: Fix credential parameters conversion to WebAuthn4j object model when used with Spring Session

ltanguy:
Replace comparison based on the != operator by an .equals() comparison based on the text value. The 2 objects have the same value in most setups, but not when used in conjunction with Spring Session for example, as the object in session is built for each request.

Fixes gh-17164


rwinch (changes requested):
Thanks for the pull request! Can you please:

- add a unit test for this change?
- rebase this off the oldest version of the code that contains this bug so that it is fixed for older versions? We will merge the bug forward

ltanguy:
Thanks for the feedback.
I rebased the branch on the oldest commit and added a unit test.
I also added a commit that makes all classes that are used by PublicKeyCredentialCreationOptions serializable, in order to be fully compatible with Spring Session that uses the Java object serializer by default.

re1709:
Any update on this please? Been holding off implementing passkeys as would require the same changes made here.

ltanguy:
@rwinch Is there something else expected on this pull request ?

Issue #16605, opened 2025-02-14
Allow setting Oauth 2.0 Resource Indicators in Oauth client calls


pat-mccusker:
**Expected Behavior**

Allow the setting of [Resource Indicators for Oauth 2.0](https://tools.ietf.org/html/rfc8707) on `ClientRegistration` to then be used for authentication requests.

**Current Behavior**

The current process for setting an RFC 8707 resource indicator might appear to be begin with either

1. Implementing a custom `Oauth2AuthorizedClientManager` and in the `authorize` method adding the indicator as a custom attribute entry to the `Oauth2AuthorizationContext`.  
2. Instantiating an existing `Oauth2AuthorizedClientManager` implementation and setting a `contextAttributesMapper` on that instance that would create the resource indicator entry.

However, it seems that this value would ultimately not be applied by a parametersConverter such as the `OAuth2ClientCredentialsGrantRequestEntityConverter`, so quite a bit more custom work appears to be necessary.

**Context**

My specific case involves a client credentials flow involving a `resource` indicator rather than core oauth 2 `scopes`, using `WebClient` and a `ServletOAuth2AuthorizedClientExchangeFilterFunction`.  I'm also leveraging spring-boot oauth client autoconfiguration properties, but as that project's property values are applied to spring-security `ClientRegistration`s starting here seems to make the most sense.

The only existing issues involving resource indicators that I'm aware of are this project's [#6972](https://github.com/spring-projects/spring-security/issues/6972) and the spring-authorization-server project's [support request](https://github.com/spring-projects/spring-authorization-server/issues/279).

Are there any other workarounds that I'm unaware of?

pat-mccusker:
Prior to spring-security 6.4, my workaround turned out to be roughly the following:

1. Create a custom `OAuth2ClientCredentialsGrantRequestEntityConverter` that could set the resource indicator after the default parameter values had been converted:
```@Component
public class ResourceIndicatorRequestEntityConverter extends OAuth2ClientCredentialsGrantRequestEntityConverter {
	private Map<String, String> clientResource = new HashMap<>();

	@Override
	protected MultiValueMap<String, String> createParameters(OAuth2ClientCredentialsGrantRequest request) {
		var defaultParameters = super.createParameters(request);

		String registrationId = request.getClientRegistration().getRegistrationId();
		var resource = clientResource.get(registrationId);
		if (hasText(resource)) {
			defaultParameters.add("resource", resource);
		}
		return defaultParameters;
	}
[...]

PR #17235, opened 2025-06-11
Add documentation for custom OAuth2 client_credentials request fields

pat-mccusker:
Closes gh-16605

While the process for adding these custom indicators is in retrospect pretty straightforward, it took me a bit of time to conceptually put it all together and so I thought it wouldn't hurt to add some documentation for it.

### DIFF
--- a/docs/modules/ROOT/pages/servlet/oauth2/client/client-authentication.adoc
+++ b/docs/modules/ROOT/pages/servlet/oauth2/client/client-authentication.adoc
@@ -78,6 +78,62 @@ spring:
             ...
 ----
 
+[[oauth2-client-authentication-client-credentials-custom-request-params]]
+=== Add custom OAuth 2.0 request parameters
+
+Spring Security does not support certain OAuth 2.0 request parameters
+(e.g. https://datatracker.ietf.org/doc/html/rfc8707#name-resource-parameter[Resource Indicators for OAuth 2.0] or others mentioned
+in https://datatracker.ietf.org/doc/html/rfc8693#section-2.1[RFC 8693]), however adding such fields in addition to the default token request conversion is straightforward.
+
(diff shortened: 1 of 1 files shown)


jgrandja:
Thanks for the PR @pat-mccusker. However, it doesn't make sense to add this documentation and code sample for Resource Indicators for OAuth 2.0 given that the feature is not even implemented.

Feel free to add the full code sample to gh-16605. I'll leave the issue open for now but will close this PR.

As an FYI, I don't see Resource Indicators for OAuth 2.0 being supported anytime soon as it doesn't appear there is much demand for it and we have a number of higher priority items on our list for the major release of Spring Security `7.0` and Spring Authorization Server `2.0`.

pat-mccusker:
@jgrandja When you say the feature is not implemented, are you referring to specifically on the resource server side?  I ask because its not being supported on the client side was my reasoning for adding the example.

I still think it's a valid use case as some may want to interact with some non-spring resource server that does act on the `resource` or `audience` request parameter and it might not be immediately obvious how this could be done with minimal additions to the existing client configuration.

jgrandja:
@pat-mccusker 

> When you say the feature is not implemented, are you referring to specifically on the resource server side?

I'm referring to both resource server AND client side, as all OAuth2 flows involve at least the 2 parties and also the AS in other flows.

Either way, it's not supported on client and resource server side in Spring Security. 

Furthermore, it's already documented on how to [customize the authorization request](https://docs.spring.io/spring-security/reference/servlet/oauth2/client/authorization-grants.html#oauth2-client-authorization-code-authorization-request-resolver) or [customize token request](https://docs.spring.io/spring-security/reference/servlet/oauth2/client/authorization-grants.html#oauth2-client-client-credentials-access-token-request), which generally shows how one would add a custom parameter (e.g. `resource` or `audience`), and therefore I feel the PR adds more specific documentation but at the same time is a duplicate of what we already have.

Issue #17188, opened 2025-05-30
Ensure ID Token is updated after refresh token (Reactive)


jgrandja:
We need to implement the Reactive counterpart of gh-16589.

evgeniycheban:
Hi @jgrandja, can you assign this one to me?

jgrandja:
Thank you @evgeniycheban. I've assigned it to you.

evgeniycheban:
Hi @jgrandja, I have opened a PR, I have some doubts about the correct implementation of this.

I have added a `RefreshTokenReactiveOAuth2AuthorizationSuccessHandler` that handles a `SecurityContext` refresh, however it depends on a `ServerSecurityContextRepository` which requires a `ServerWebExchange`, it will work for use within the context of a `ServerWebExchange`, but if we want to refresh a `SecurityContext` for those clients that are used outside of a `ServerWebExchange` context, we might need to think about having a different abstraction here, one thing that comes to mind is to bind an `Authentication` object to `ClientRequest` similar how it's proposed to be done in gh-16284, what do you think?

jgrandja:
Thanks for the PR @evgeniycheban. I will do my best to review this soon. The team has a few high priority items for the upcoming major releases of Spring Security `7.0` and Spring Authorization Server `2.0` so we need to focus on those items first. Thank you for your patience.

Issue #16665, opened 2025-02-26
Add support for automatic context-propagation with Micrometer


sjohnr:
`ThreadLocalAccessor` support to be added for:
 
* `SecurityContextHolder`
* `ReactiveSecurityContextHolder`
* `ServerWebExchange` (Context attribute set by [`ServerWebExchangeReactorContextWebFilter`](https://github.com/spring-projects/spring-security/blob/f2d78a01c3f2401dfdb780b4c9be13564e9048d7/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java#L4055-L4062))

Related gh-13733

sjohnr:
**Note** that this enables `DefaultReactiveOAuth2AuthorizedClientManager` (and `ServerOAuth2AuthorizedClientExchangeFilterFunction` for `WebClient`) to be used with [blocking execution](https://docs.spring.io/spring-framework/reference/web/webflux/config.html#webflux-config-blocking-execution) when [automatic context propagation](https://docs.spring.io/spring-boot/reference/actuator/observability.html#actuator.observability.context-propagation) is enabled and the [context-propagation](https://github.com/micrometer-metrics/context-propagation) library is on the classpath.

PR #17246, opened 2025-06-14
Ensure ID Token is updated after refresh token (Reactive)

evgeniycheban:
Closes gh-17188

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshTokenReactiveOAuth2AuthorizedClientProvider.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshTokenReactiveOAuth2AuthorizedClientProvider.java
@@ -21,7 +21,9 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
 
 import reactor.core.publisher.Mono;
@@ -40,6 +42,7 @@ import org.springframework.util.Assert;
  * {@link AuthorizationGrantType#REFRESH_TOKEN refresh_token} grant.
  *
  * @author Joe Grandja
+ * @author Evgeniy Cheban
  * @since 5.2
  * @see ReactiveOAuth2AuthorizedClientProvider
  * @see WebClientReactiveRefreshTokenTokenResponseClient
@@ -48,6 +51,8 @@ public final class RefreshTokenReactiveOAuth2AuthorizedClientProvider
 		implements ReactiveOAuth2AuthorizedClientProvider {
 
 	private ReactiveOAuth2AccessTokenResponseClient<OAuth2RefreshTokenGrantRequest> accessTokenResponseClient = new WebClientReactiveRefreshTokenTokenResponseClient();
+
+	private ReactiveOAuth2AuthorizationSuccessHandler authorizationSuccessHandler = new RefreshOidcUserReactiveOAuth2AuthorizationSuccessHandler();
 
 	private Duration clockSkew = Duration.ofSeconds(60);
 
@@ -96,12 +101,17 @@ public final class RefreshTokenReactiveOAuth2AuthorizedClientProvider
 			.flatMap(this.accessTokenResponseClient::getTokenResponse)
 			.onErrorMap(OAuth2AuthorizationException.class,
 					(e) -> new ClientAuthorizationException(e.getError(), clientRegistration.getRegistrationId(), e))
-			.map((tokenResponse) -> new OAuth2AuthorizedClient(clientRegistration, context.getPrincipal().getName(),
-					tokenResponse.getAccessToken(), tokenResponse.getRefreshToken()));
-	}
-
-	private boolean hasTokenExpired(OAuth2Token token) {
-		return this.clock.instant().isAfter(token.getExpiresAt().minus(this.clockSkew));
+			.flatMap((tokenResponse) -> {
+				OAuth2AuthorizedClient refreshedClient = new OAuth2AuthorizedClient(clientRegistration,
+						context.getPrincipal().getName(), tokenResponse.getAccessToken(),
+						tokenResponse.getRefreshToken());
+				Map<String, Object> attributes = new HashMap<>(context.getAttributes());
+				attributes.putAll(tokenResponse.getAdditionalParameters());
+				return this.authorizationSuccessHandler
+					.onAuthorizationSuccess(refreshedClient, context.getPrincipal(),
+							Collections.unmodifiableMap(attributes))
+					.thenReturn(refreshedClient);
+			});
 	}
 
 	/**
@@ -114,6 +124,19 @@ public final class RefreshTokenReactiveOAuth2AuthorizedClientProvider
 			ReactiveOAuth2AccessTokenResponseClient<OAuth2RefreshTokenGrantRequest> accessTokenResponseClient) {
 		Assert.notNull(accessTokenResponseClient, "accessTokenResponseClient cannot be null");
 		this.accessTokenResponseClient = accessTokenResponseClient;
+	}
+
+	/**
+	 * Sets a {@link ReactiveOAuth2AuthorizationSuccessHandler} to use for handling
+	 * successful refresh token response, defaults to
+	 * {@link RefreshOidcUserReactiveOAuth2AuthorizationSuccessHandler}.
+	 * @param authorizationSuccessHandler the
+	 * {@link ReactiveOAuth2AuthorizationSuccessHandler} to use
+	 * @since 7.1
+	 */
+	public void setAuthorizationSuccessHandler(ReactiveOAuth2AuthorizationSuccessHandler authorizationSuccessHandler) {
+		Assert.notNull(authorizationSuccessHandler, "authorizationSuccessHandler cannot be null");
+		this.authorizationSuccessHandler = authorizationSuccessHandler;
 	}
 
 	/**
@@ -141,6 +164,10 @@ public final class RefreshTokenReactiveOAuth2AuthorizedClientProvider
 	public void setClock(Clock clock) {
 		Assert.notNull(clock, "clock cannot be null");
 		this.clock = clock;
+	}
+
+	private boolean hasTokenExpired(OAuth2Token token) {
+		return this.clock.instant().isAfter(token.getExpiresAt().minus(this.clockSkew));
 	}
 
 }
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultReactiveOAuth2AuthorizedClientManager.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultReactiveOAuth2AuthorizedClientManager.java
@@ -318,14 +318,14 @@ public final class DefaultReactiveOAuth2AuthorizedClientManager implements React
 			return Mono.justOrEmpty(serverWebExchange)
 					.switchIfEmpty(currentServerWebExchangeMono)
 					.flatMap((exchange) -> {
-						Map<String, Object> contextAttributes = Collections.emptyMap();
+						Map<String, Object> contextAttributes = new HashMap<>();
+						contextAttributes.put(ServerWebExchange.class.getName(), serverWebExchange);
 						String scope = exchange.getRequest().getQueryParams().getFirst(OAuth2ParameterNames.SCOPE);
 						if (StringUtils.hasText(scope)) {
-							contextAttributes = new HashMap<>();
 							contextAttributes.put(OAuth2AuthorizationContext.REQUEST_SCOPE_ATTRIBUTE_NAME,
 									StringUtils.delimitedListToStringArray(scope, " "));
 						}
-						return Mono.just(contextAttributes);
+						return Mono.just(Collections.unmodifiableMap(contextAttributes));
 					})
 					.defaultIfEmpty(Collections.emptyMap());
 			// @formatter:on
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
@@ -51,6 +51,8 @@ import org.springframework.security.oauth2.core.OAuth2AuthorizationException;
 import org.springframework.security.oauth2.core.OAuth2Error;
 import org.springframework.security.oauth2.core.OAuth2ErrorCodes;
 import org.springframework.security.oauth2.core.endpoint.OAuth2ParameterNames;
+import org.springframework.security.web.server.context.ServerSecurityContextRepository;
+import org.springframework.security.web.server.context.WebSessionServerSecurityContextRepository;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 import org.springframework.web.reactive.function.client.ClientRequest;
@@ -96,6 +98,7 @@ import org.springframework.web.server.ServerWebExchange;
  * @author Rob Winch
  * @author Joe Grandja
  * @author Phil Clay
+ * @author Evgeniy Cheban
  * @since 5.1
  */
 public final class ServerOAuth2AuthorizedClientExchangeFilterFunction implements ExchangeFilterFunction {
@@ -138,6 +141,8 @@ public final class ServerOAuth2AuthorizedClientExchangeFilterFunction implements
 	private String defaultClientRegistrationId;
 
 	private ClientResponseHandler clientResponseHandler;
+
+	private ServerSecurityContextRepository serverSecurityContextRepository = new WebSessionServerSecurityContextRepository();
 
 	/**
 	 * Constructs a {@code ServerOAuth2AuthorizedClientExchangeFilterFunction} using the
@@ -330,8 +335,11 @@ public final class ServerOAuth2AuthorizedClientExchangeFilterFunction implements
 	}
 
 	private Mono<ClientResponse> exchangeAndHandleResponse(ClientRequest request, ExchangeFunction next) {
-		return next.exchange(request)
-			.transform((responseMono) -> this.clientResponseHandler.handleResponse(request, responseMono));
+		// Re-request an Authentication from serverSecurityContextRepository since it
+		// might have been changed during provider invocation.
+		return effectiveAuthentication(request).flatMap((authentication) -> next.exchange(request)
+			.transform((responseMono) -> this.clientResponseHandler.handleResponse(request, responseMono))
+			.contextWrite(ReactiveSecurityContextHolder.withAuthentication(authentication)));
 	}
 
 	private Mono<OAuth2AuthorizedClient> authorizedClient(ClientRequest request) {
@@ -359,6 +367,17 @@ public final class ServerOAuth2AuthorizedClientExchangeFilterFunction implements
 					t3.getT3().ifPresent((exchange) -> builder.attribute(ServerWebExchange.class.getName(), exchange));
 					return builder.build();
 				});
+		// @formatter:on
+	}
+
+	private Mono<Authentication> effectiveAuthentication(ClientRequest request) {
+		// @formatter:off
+		return effectiveServerWebExchange(request)
+			.filter(Optional::isPresent)
+			.map(Optional::get)
+			.flatMap(this.serverSecurityContextRepository::load)
+			.mapNotNull(SecurityContext::getAuthentication)
+			.switchIfEmpty(this.currentAuthenticationMono);
 		// @formatter:on
 	}
 
@@ -443,6 +462,19 @@ public final class ServerOAuth2AuthorizedClientExchangeFilterFunction implements
 	public void setAuthorizationFailureHandler(ReactiveOAuth2AuthorizationFailureHandler authorizationFailureHandler) {
 		Assert.notNull(authorizationFailureHandler, "authorizationFailureHandler cannot be null");
 		this.clientResponseHandler = new AuthorizationFailureForwarder(authorizationFailureHandler);
+	}
+
+	/**
+	 * Sets a {@link ServerSecurityContextRepository} to use for re-obtaining a
+	 * {@link SecurityContext} if it has been refreshed during provider invocation,
+	 * defaults to {@link WebSessionServerSecurityContextRepository}.
+	 * @param serverSecurityContextRepository the {@link ServerSecurityContextRepository}
+	 * to use
+	 * @since 7.1
+	 */
+	public void setServerSecurityContextRepository(ServerSecurityContextRepository serverSecurityContextRepository) {
+		Assert.notNull(serverSecurityContextRepository, "serverSecurityContextRepository cannot be null");
+		this.serverSecurityContextRepository = serverSecurityContextRepository;
 	}
 
 	@FunctionalInterface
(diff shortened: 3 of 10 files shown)


jgrandja:
Thanks for your patience @evgeniycheban and apologies that I could not get this into `7.0`. We had quite a workload for this round of majors.

I have scheduled this for `7.1.0-M1` and will review this soon. In the meantime, when you have a moment can you please rebase. Thank you.

evgeniycheban:
Hi @jgrandja, I have rebased my branch and also updated license headers to meet new pattern, let me know your feedback once you review it, thanks.

jgrandja:
@evgeniycheban I took a look at the PR and noticed that the implementation is different compared to the Servlet implementation. 

`RefreshTokenReactiveOAuth2AuthorizationSuccessHandler` combines the logic in `OidcAuthorizedClientRefreshedEventListener` and `OidcUserRefreshedEventListener` and does not make use of Spring Framework's `ApplicationEventPublisher` and `ApplicationListener`.

The preference is to keep the Servlet and Reactive implementations as close as possible. Is there a reason you did not make use of `ApplicationEventPublisher` and `ApplicationListener` ?

Just a heads up that I'm on PTO starting tomorrow and returning Jan 5 so I'll reply when I return and we can work through this.

evgeniycheban:
Hi @jgrandja, thank you for reviewing it. The reason I did not use the `ApplicationEventPublisher` is because in the current implementation there is no support for reactive application events, I can recall that there was a draft [PR](https://github.com/spring-projects/spring-security/pull/14361) raised by @marcusdacoregio but it didn't get merged.

jgrandja:
@evgeniycheban Got it. I'll do a detailed review next week and we can work towards merging this.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshTokenReactiveOAuth2AuthorizedClientProvider.java
```diff
@@ -49,6 +53,8 @@ public final class RefreshTokenReactiveOAuth2AuthorizedClientProvider
 
 	private ReactiveOAuth2AccessTokenResponseClient<OAuth2RefreshTokenGrantRequest> accessTokenResponseClient = new WebClientReactiveRefreshTokenTokenResponseClient();
 
+	private ReactiveOAuth2AuthorizationSuccessHandler refreshTokenSuccessHandler = new RefreshTokenReactiveOAuth2AuthorizationSuccessHandler();
```
jgrandja: `ReactiveOAuth2AuthorizationSuccessHandler` is not designed to be used with a `ReactiveOAuth2AuthorizedClientProvider` and instead is meant to be used with a `ReactiveOAuth2AuthorizedClientManager`.

Can you re-work this and set `RefreshTokenReactiveOAuth2AuthorizationSuccessHandler` in `DefaultReactiveOAuth2AuthorizedClientManager` ?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshTokenReactiveOAuth2AuthorizationSuccessHandler.java
```diff
@@ -0,0 +1,302 @@
...
+
+/**
+ * A {@link ReactiveOAuth2AuthorizationSuccessHandler} that refreshes an {@link OidcUser}
+ * in the {@link SecurityContext} if the refreshed {@link OidcIdToken} is valid according
+ * to <a href=
+ * "https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse">OpenID
+ * Connect Core 1.0 - Section 12.2 Successful Refresh Response</a>
+ *
+ * @author Evgeniy Cheban
+ * @since 7.0
```
jgrandja: Update to `7.1`

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshTokenReactiveOAuth2AuthorizationSuccessHandler.java
```diff
@@ -0,0 +1,302 @@
...
+	private ReactiveOAuth2UserService<OidcUserRequest, OidcUser> userService = new OidcReactiveOAuth2UserService();
+
+	private GrantedAuthoritiesMapper authoritiesMapper = (authorities) -> authorities;
+
+	private Duration clockSkew = Duration.ofSeconds(60);
+
+	@Override
+	public Mono<Void> onAuthorizationSuccess(OAuth2AuthorizedClient authorizedClient, Authentication principal,
+			Map<String, Object> attributes) {
+		if (!(principal instanceof OAuth2AuthenticationToken authenticationToken)
```
jgrandja, line 101: `OidcAuthorizedClientRefreshedEventListener` performs a couple of validation checks that are missing here. Specifically, checks `OAuth2AccessTokenResponse` to ensure `openid` scope is associated with the access token and an `id_token` is returned. These pre-checks should be in place as we only want to proceed if this is an OIDC refresh token flow.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
```diff
@@ -330,8 +335,11 @@ public Mono<ClientResponse> filter(ClientRequest request, ExchangeFunction next)
 	}
 
 	private Mono<ClientResponse> exchangeAndHandleResponse(ClientRequest request, ExchangeFunction next) {
-		return next.exchange(request)
-			.transform((responseMono) -> this.clientResponseHandler.handleResponse(request, responseMono));
+		// Re-request an Authentication from serverSecurityContextRepository since it
+		// might have been changed during provider invocation.
+		return effectiveAuthentication(request).flatMap((authentication) -> next.exchange(request)
```
jgrandja, line 340: Instead of adding this logic, can we not update the `SecurityContext` in the reactor context in `RefreshTokenReactiveOAuth2AuthorizationSuccessHandler` right after `this.serverSecurityContextRepository.save(exchange, securityContext)` to avoid this `serverSecurityContextRepository.load()`?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshTokenReactiveOAuth2AuthorizationSuccessHandler.java
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	/**
+	 * Sets a {@link ServerSecurityContextRepository} to use for refreshing a
+	 * {@link SecurityContext}, defaults to
+	 * {@link WebSessionServerSecurityContextRepository}.
+	 * @param serverSecurityContextRepository the {@link ServerSecurityContextRepository}
+	 * to use
+	 */
+	public void setServerSecurityContextRepository(ServerSecurityContextRepository serverSecurityContextRepository) {
```
jgrandja: Please move the `public` methods above the `private` methods.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
evgeniycheban: The tricky part here is that we can't propagate the refreshed `SecurityContext` to the upstream in a reactor context, it is propagated always one-way from upstream to downstream, so that's why I used saving the refreshed `SecurityContext` in `ServerSecurityContextRepository` to be able later to continue the reactive chain with the refreshed `SecurityContext` and propagate it using `contextWrite`.

I also considered using a custom `CoreSubscriber` similar to how it's done in `SecurityReactorContextConfiguration`, but I'm not sure if it would be the right approach here.
What do you think?

jgrandja: > it is propagated always one-way from upstream to downstream

I'm not sure I understand this?

> The tricky part here is that we can't propagate the refreshed `SecurityContext` to the upstream in a reactor context

Please see the following issue to see if it helps?

https://github.com/spring-projects/spring-security/issues/16665

evgeniycheban: I will have a look at `context-propagation` library, thank you.

evgeniycheban: I get the point you provided here @jgrandja, however, in order to propagate the refreshed `SecurityContext` to the next `ExchangeFilterFunction` in `ServerOAuth2AuthorizedClientExchangeFilterFunction`:
```java
return next.exchange(request)
	.transform((responseMono) -> this.clientResponseHandler.handleResponse(request, responseMono));
```
updating `SecurityContext` in the reactor context in `RefreshTokenReactiveOAuth2AuthorizationSuccessHandler` has no effect, so we need to write the new context to the reactive chain before calling `next.exchange(request)`:
```java
return effectiveAuthentication(request).flatMap((authentication) -> next.exchange(request)
	.transform((responseMono) -> this.clientResponseHandler.handleResponse(request, responseMono))
	.contextWrite(ReactiveSecurityContextHolder.withAuthentication(authentication)));
```
At this point we `load` the `SecurityContext` to get the most recent version to be passed to the downstream.

Another option could be moving the logic from `RefreshTokenReactiveOAuth2AuthorizationSuccessHandler` to `RefreshTokenReactiveOAuth2AuthorizedClientProvider` and enriching an `OAuth2AuthorizedClient` with the new `Authentication` object when the token is refreshed and then propagate it to the downstream to the next `ExchangeFilterFunction` in `ServerOAuth2AuthorizedClientExchangeFilterFunction`.

@jgrandja, I'm also available for a short call if you want to discuss this in more details, just let me know, thanks.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshTokenReactiveOAuth2AuthorizedClientProvider.java
evgeniycheban: Done

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshTokenReactiveOAuth2AuthorizationSuccessHandler.java
evgeniycheban: Done

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshTokenReactiveOAuth2AuthorizationSuccessHandler.java
evgeniycheban: Done.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshTokenReactiveOAuth2AuthorizationSuccessHandler.java
evgeniycheban: Done.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
evgeniycheban: @jgrandja I can as well commit the change that would demonstrate the failing test if we do `.contextWrite` in `RefreshTokenReactiveOAuth2AuthorizationSuccessHandler` instead of `ServerOAuth2AuthorizedClientExchangeFilterFunction`, please let me know.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshTokenReactiveOAuth2AuthorizedClientProvider.java
evgeniycheban: I wonder if it makes sense to have a separate setter for `refreshTokenSuccessHandler`, I think we could build a default success handler:
```java
private ReactiveOAuth2AuthorizationSuccessHandler getDefaultAuthorizationSuccessHandler(
		ServerOAuth2AuthorizedClientRepository authorizedClientRepository) {
	return (authorizedClient, principal, attributes) -> {
		Mono<Void> saveAuthorizedClient = authorizedClientRepository.saveAuthorizedClient(authorizedClient,
					principal, (ServerWebExchange) attributes.get(ServerWebExchange.class.getName()));
		return saveAuthorizedClient.then(Mono.defer(() -> this.refreshTokenSuccessHandler
				.onAuthorizationSuccess(authorizedClient, principal, attributes)));
	};
}
```
For now, I added a setter for `refreshTokenSuccessHandler`, but maybe it would be better to have a public setter only for `authorizationSuccessHandler` which would override the whole success handling logic instead of having setters for overriding some specific parts of it.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshTokenReactiveOAuth2AuthorizationSuccessHandler.java
```diff
@@ -0,0 +1,312 @@
...
+ * A {@link ReactiveOAuth2AuthorizationSuccessHandler} that refreshes an {@link OidcUser}
+ * in the {@link SecurityContext} if the refreshed {@link OidcIdToken} is valid according
+ * to <a href=
+ * "https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse">OpenID
+ * Connect Core 1.0 - Section 12.2 Successful Refresh Response</a>
+ *
+ * @author Evgeniy Cheban
+ * @since 7.1
+ */
+public final class RefreshTokenReactiveOAuth2AuthorizationSuccessHandler
```
jgrandja: Please rename to `RefreshOidcUserReactiveOAuth2AuthorizationSuccessHandler`

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshTokenReactiveOAuth2AuthorizationSuccessHandler.java
```diff
@@ -0,0 +1,312 @@
...
+	 * @param clockSkew the maximum acceptable clock skew to use
+	 */
+	public void setClockSkew(Duration clockSkew) {
+		Assert.notNull(clockSkew, "clockSkew cannot be null");
+		Assert.isTrue(clockSkew.getSeconds() >= 0, "clockSkew must be >= 0");
+		this.clockSkew = clockSkew;
+	}
+
+	@Override
+	public Mono<Void> onAuthorizationSuccess(OAuth2AuthorizedClient authorizedClient, Authentication principal,
```
jgrandja: Please move `onAuthorizationSuccess` above the setters as it should be right after the members

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshTokenReactiveOAuth2AuthorizationSuccessHandler.java
```diff
@@ -0,0 +1,312 @@
...
+
+	private Mono<ServerWebExchange> serverWebExchange(Map<String, Object> attributes) {
+		if (attributes.get(ServerWebExchange.class.getName()) instanceof ServerWebExchange exchange) {
+			return Mono.just(exchange);
+		}
+		return currentServerWebExchangeMono;
+	}
+
+	private Mono<OAuth2AccessTokenResponse> accessTokenResponse(Map<String, Object> attributes) {
+		if (attributes.get(OAuth2AccessTokenResponse.class.getName()) instanceof OAuth2AccessTokenResponse response) {
```
jgrandja: Please move these 2 checks so they are evaluated first, preserving the same order as in `OidcAuthorizedClientRefreshedEventListener`

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultReactiveOAuth2AuthorizedClientManager.java
```diff
@@ -115,6 +117,8 @@ public final class DefaultReactiveOAuth2AuthorizedClientManager implements React
 
 	private Function<OAuth2AuthorizeRequest, Mono<Map<String, Object>>> contextAttributesMapper = new DefaultContextAttributesMapper();
 
+	private ReactiveOAuth2AuthorizationSuccessHandler refreshTokenSuccessHandler = new RefreshTokenReactiveOAuth2AuthorizationSuccessHandler();
```
jgrandja: Please remove this as `authorizationSuccessHandler` already defaults to the combined (save/refresh) behaviour and applications can override using `setAuthorizationSuccessHandler()`.

jgrandja (changes requested):
Thank you for the updates @evgeniycheban. I think we'll be ready to merge after you address some minor feedback.

Also, please add an integration test in `ServerOAuth2AuthorizedClientExchangeFilterFunctionITests` to ensure `RefreshTokenReactiveOAuth2AuthorizationSuccessHandler` is fully tested.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServerOAuth2AuthorizedClientExchangeFilterFunction.java
jgrandja: @evgeniycheban I took a look at this in more detail and I think we are good with what you have there now.

evgeniycheban:
Hi @jgrandja, thank you for the review, after we moved `refreshTokenSuccessHandler` to `DefaultReactiveOAuth2AuthorizedClientManager` I noticed that we no longer can access `OAuth2AccessTokenResponse` which is needed to get `id_token` from `additionalParameters`, would it be an option to add a `String idToken` field to `OAuth2AuthorizedClient` or even `Map<String, Object> additionalParameters` to store an additional context for the client?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshTokenReactiveOAuth2AuthorizationSuccessHandler.java
evgeniycheban: Done

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshTokenReactiveOAuth2AuthorizationSuccessHandler.java
evgeniycheban: Done

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshTokenReactiveOAuth2AuthorizationSuccessHandler.java
evgeniycheban: Done

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/DefaultReactiveOAuth2AuthorizedClientManager.java
evgeniycheban: Done

jgrandja:
@evgeniycheban Instead of adding `String idToken` to `OAuth2AuthorizedClient` please add `private Map<String, Object> attributes` and in there you can map `OidcParameterNames.ID_TOKEN`.

Please ensure to update `OAuth2AuthorizedClient` in `OAuth2LoginAuthenticationFilter` and the reactive counterpart.

evgeniycheban:
Hi @jgrandja, I've pushed the latest changes, please have a final look :)
I've also replaced deprecated `org.springframework.lang.Nullable` annotation with jspecify in some files.

jgrandja:
@evgeniycheban The addition of `OAuth2AuthorizedClient.attributes` resulted in too many touch points in the code that doesn't feel right. In addition to the last updates, we still need to update `oauth2-client-schema.sql`, `oauth2-client-schema-postgres.sql`, `JdbcOAuth2AuthorizedClientService` and `R2dbcReactiveOAuth2AuthorizedClientService`, which is making me rethink things.

Looking at my previous comment:

> `ReactiveOAuth2AuthorizationSuccessHandler` is not designed to be used with a `ReactiveOAuth2AuthorizedClientProvider` and instead is meant to be used with a `ReactiveOAuth2AuthorizedClientManager`.
>
> Can you re-work this and set `RefreshTokenReactiveOAuth2AuthorizationSuccessHandler` in `DefaultReactiveOAuth2AuthorizedClientManager` ?

This is where things took a turn and after reviewing the API's, I now feel `ReactiveOAuth2AuthorizationSuccessHandler` can also be used by a `ReactiveOAuth2AuthorizedClientProvider` even though it was initially designed to be used with a `ReactiveOAuth2AuthorizedClientManager`.

Having said that, your initial motivation to associate `RefreshOidcUserReactiveOAuth2AuthorizationSuccessHandler` with `RefreshTokenReactiveOAuth2AuthorizedClientProvider` was correct and I should have stuck with that as it's a much cleaner implementation and keeps the changes very isolated. Apologies for this but can you please revert `OAuth2AuthorizedClient.attributes` and associated updates and add  `RefreshTokenReactiveOAuth2AuthorizedClientProvider.setAuthorizationSuccessHandler(ReactiveOAuth2AuthorizationSuccessHandler)` making `RefreshOidcUserReactiveOAuth2AuthorizationSuccessHandler` the default.

evgeniycheban:
> @evgeniycheban The addition of `OAuth2AuthorizedClient.attributes` resulted in too many touch points in the code that doesn't feel right. In addition to the last updates, we still need to update `oauth2-client-schema.sql`, `oauth2-client-schema-postgres.sql`, `JdbcOAuth2AuthorizedClientService` and `R2dbcReactiveOAuth2AuthorizedClientService`, which is making me rethink things.
> 
> Looking at my previous comment:
> 
> > `ReactiveOAuth2AuthorizationSuccessHandler` is not designed to be used with a `ReactiveOAuth2AuthorizedClientProvider` and instead is meant to be used with a `ReactiveOAuth2AuthorizedClientManager`.
> > Can you re-work this and set `RefreshTokenReactiveOAuth2AuthorizationSuccessHandler` in `DefaultReactiveOAuth2AuthorizedClientManager` ?
> 
> This is where things took a turn and after reviewing the API's, I now feel `ReactiveOAuth2AuthorizationSuccessHandler` can also be used by a `ReactiveOAuth2AuthorizedClientProvider` even though it was initially designed to be used with a `ReactiveOAuth2AuthorizedClientManager`.
> 
> Having said that, your initial motivation to associate `RefreshOidcUserReactiveOAuth2AuthorizationSuccessHandler` with `RefreshTokenReactiveOAuth2AuthorizedClientProvider` was correct and I should have stuck with that as it's a much cleaner implementation and keeps the changes very isolated. Apologies for this but can you please revert `OAuth2AuthorizedClient.attributes` and associated updates and add `RefreshTokenReactiveOAuth2AuthorizedClientProvider.setAuthorizationSuccessHandler(ReactiveOAuth2AuthorizationSuccessHandler)` making `RefreshOidcUserReactiveOAuth2AuthorizationSuccessHandler` the default.

Hi @jgrandja, no worries, I will do it and notify you once I finish. Initially, I wasn't sure about the design of this feature and chose the minimal changes path to see how it would work in the current architecture.

evgeniycheban:
Hi @jgrandja, I've updated the PR, please have a look when you have time, thanks.

jgrandja:
Thank you for all the updates and patience @evgeniycheban. Very solid work! 👍 

This is now merged.

PR #17247, opened 2025-06-14
Improve authoritiesClaimName validation in JwtGrantedAuthoritiesConverter

chanbinme:
## Summary
Use `StringUtils.hasText()` instead of null check in `getAuthoritiesClaimName()` to properly handle empty strings and whitespace-only strings.

## Problem
The current null check (`!= null`) incorrectly treats empty strings (`""`) and whitespace-only strings (`"   "`) as valid claim names. While `setAuthoritiesClaimName()` validates with `Assert.hasText()`, the field can be set through other means (reflection, constructors, etc.) that bypass this validation.

## Changes
- Replace `!= null` check with `StringUtils.hasText()` 
- Add comprehensive test coverage for blank claim names

## Testing
Added parameterized tests covering empty strings, whitespace strings, and null values using `ReflectionTestUtils` to simulate edge cases.

## Impact
- Fixes edge case bugs with blank claim names
- Maintains full backward compatibility
- Follows defensive programming principles
- All existing tests pass

This is a straightforward bug fix that improves robustness without breaking changes.

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverter.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverter.java
@@ -53,7 +53,7 @@ public final class JwtGrantedAuthoritiesConverter implements Converter<Jwt, Coll
 
 	private String authoritiesClaimDelimiter = DEFAULT_AUTHORITIES_CLAIM_DELIMITER;
 
-	private String authoritiesClaimName;
+	private Collection<String> authoritiesClaimNames = WELL_KNOWN_AUTHORITIES_CLAIM_NAMES;
 
 	/**
 	 * Extract {@link GrantedAuthority}s from the given {@link Jwt}.
@@ -102,14 +102,11 @@ public final class JwtGrantedAuthoritiesConverter implements Converter<Jwt, Coll
 	 */
 	public void setAuthoritiesClaimName(String authoritiesClaimName) {
 		Assert.hasText(authoritiesClaimName, "authoritiesClaimName cannot be empty");
-		this.authoritiesClaimName = authoritiesClaimName;
+		this.authoritiesClaimNames = Collections.singletonList(authoritiesClaimName);
 	}
 
 	private String getAuthoritiesClaimName(Jwt jwt) {
-		if (this.authoritiesClaimName != null) {
-			return this.authoritiesClaimName;
-		}
-		for (String claimName : WELL_KNOWN_AUTHORITIES_CLAIM_NAMES) {
+		for (String claimName : this.authoritiesClaimNames) {
 			if (jwt.hasClaim(claimName)) {
 				return claimName;
 			}


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverter.java
```diff
@@ -106,7 +106,7 @@ public void setAuthoritiesClaimName(String authoritiesClaimName) {
 	}
 
 	private String getAuthoritiesClaimName(Jwt jwt) {
-		if (this.authoritiesClaimName != null) {
+		if (StringUtils.hasText(this.authoritiesClaimName)) {
```
jzheaux: Instead of continuing to check for `null`, let's remove the null check by doing the following:

1. Change the `authoritiesClaimName` to `Collection<String> authoritiesClaimNames = WELL_KNOWN_AUTHORITIES_CLAIM_NAMES;`
2. Change `setAuthoritiesClaimName` to set `this.authoritiesClaimNames` to a single value
3. Replace the for loop to use `this.authoritiesClaimNames`
4. Remove the null check

In this way, `authoritiesClaimNames` can never contain an empty claim name.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverter.java
```diff
@@ -106,7 +106,7 @@ public void setAuthoritiesClaimName(String authoritiesClaimName) {
 	}
 
 	private String getAuthoritiesClaimName(Jwt jwt) {
-		if (this.authoritiesClaimName != null) {
```
jzheaux, line 109: Please also update the copyright year in this class to 2025.

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverterTests.java
```diff
@@ -270,4 +274,21 @@ public void convertWithCustomAuthoritiesSplitRegexWhenTokenHasScopeAttributeThen
 				new SimpleGrantedAuthority("SCOPE_message:write"));
 	}
 
+	@ParameterizedTest
+	@ValueSource(strings = { "", "   " })
+	@NullSource
+	public void convertWhenAuthoritiesClaimNameIsBlankThenUsesWellKnownClaims(String invalidClaimName)
```
jzheaux: With the refactor made, this test should now be unnecessary.

jzheaux (changes requested):
Thanks for the PR, @chanbinme! I've left my feedback inline.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverter.java
chanbinme: Hi @jzheaux,

I've implemented the refactoring as suggested, but I noticed an important behavioral difference that I wanted to discuss:

**Original behavior:**
- When `authoritiesClaimName` is explicitly set → return it directly without JWT validation
- When using default claim names → return the first claim name that exists in the JWT

**Initial refactoring attempt:**
- All claim names go through JWT validation, which changes the behavior for explicitly set claim names

**My solution:**
To preserve the original behavior, I added an `isExplicitlySet` flag that tracks whether the claim name was explicitly configured. This ensures:
- Explicitly set claim names are returned without JWT validation (original behavior)
- Default claim names are validated against the JWT (original behavior)
- Null checks are eliminated since `authoritiesClaimNames` is always initialized

Would you prefer this approach, or would you like to intentionally change the behavior to always validate claim names against the JWT? I wanted to confirm the intended behavior before finalizing the implementation.

The changes include:
1. ✅ Changed `authoritiesClaimName` to `Collection<String> authoritiesClaimNames`
2. ✅ Updated `setAuthoritiesClaimName` to use `Collections.singletonList()`
3. ✅ Modified the for loop to use `this.authoritiesClaimNames`
4. ✅ Removed null checks
5. ➕ Added `isExplicitlySet` flag to preserve original behavior

Let me know your thoughts on this approach!

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverter.java
chanbinme: I've addressed your feedback. Thank you!

oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverterTests.java
chanbinme: I've removed the unnecessary test code as suggested. Thanks for the feedback!

chanbinme:
Hi @jzheaux,

Thank you so much for your helpful feedback! 
I've incorporated your suggestions and pushed the changes. 
When you have a moment, could you please take another look? 
I've also left some comments on the inline feedback for further discussion.  

Thanks again for your time and support!

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverter.java
```diff
@@ -102,14 +104,15 @@ public void setAuthoritiesClaimDelimiter(String authoritiesClaimDelimiter) {
...
 		Assert.hasText(authoritiesClaimName, "authoritiesClaimName cannot be empty");
-		this.authoritiesClaimName = authoritiesClaimName;
+		this.authoritiesClaimNames = Collections.singletonList(authoritiesClaimName);
+		this.isExplicitlySet = true;
 	}
 
 	private String getAuthoritiesClaimName(Jwt jwt) {
-		if (this.authoritiesClaimName != null) {
-			return this.authoritiesClaimName;
+		if (this.isExplicitlySet) {
```
chanbinme: Hi @jzheaux, 

I've implemented the refactoring as suggested, but I noticed an important behavioral difference that I wanted to discuss:

Original behavior:
When `authoritiesClaimName` is explicitly set → return it directly without JWT validation
When using default claim names → return the first claim name that exists in the JWT

Initial refactoring attempt:
All claim names go through JWT validation, which changes the behavior for explicitly set claim names

My solution:
To preserve the original behavior, I added an `isExplicitlySet` flag that tracks whether the claim name was explicitly configured. This ensures:

- Explicitly set claim names are returned without JWT validation (original behavior)
- Default claim names are validated against the JWT (original behavior)
- Null checks are eliminated since `authoritiesClaimNames` is always initialized

Would you prefer this approach, or would you like to intentionally change the behavior to always validate claim names against the JWT? I wanted to confirm the intended behavior before finalizing the implementation.

The changes include:
✅ Changed `authoritiesClaimName` to `Collection<String> authoritiesClaimNames`
✅ Updated `setAuthoritiesClaimName` to use `Collections.singletonList()`
✅ Modified the for loop to use `this.authoritiesClaimNames`
✅ Removed null checks
➕ Added `isExplicitlySet` flag to preserve original behavior

Let me know your thoughts on this approach!

chanbinme:
> Thanks for the PR, @chanbinme! I've left my feedback inline.

Hi @jzheaux 👋 

No rush at all, just wanted to make sure this didn't get lost in notifications. 
Happy to address any additional feedback when you have time.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/JwtGrantedAuthoritiesConverter.java
jzheaux: Thanks for the updates, @chanbinme. I'm having trouble seeing when this would matter.

If the claim is not present, the outcome is still that no authorities are added. If the claim is present, the outcome is still that the claim value is evaluated.

Are you able to share a unit test that fails unless `this.isExplicitlySet` is part of the solution?

chanbinme: Thanks for the clarification, @jzheaux. I’ve removed the `isExplicitlySet` flag.

After reviewing the logic again, I confirmed that removing `isExplicitlySet` doesn’t affect the behavior.  
If the claim is not present in the JWT, no authorities are added — as you mentioned.  
So, the flag isn’t needed in the current implementation.

jzheaux (approved):

jzheaux:
Thanks for the updates, @chanbinme! This is now merged into `main`.

PR #17251, opened 2025-06-15
Add null check for authentication token in JwtAuthenticationProvider

chanbinme:
## Summary
Add `Assert.notNull` validation in `JwtAuthenticationProvider.authenticate()` to ensure the authentication token returned by `jwtAuthenticationConverter` is not null.

## Problem
While the `JwtAuthenticationConverter.convert()` method typically returns a valid `AbstractAuthenticationToken`, there's no explicit null check before accessing the token's properties. This could potentially lead to a `NullPointerException`  if:
- Custom converter implementations return null
- Unexpected edge cases occur during token conversion
- External factors affect the conversion process

## Changes
- Added `Assert.notNull(token, "token cannot be null")` validation
- Added comprehensive test to verify the null check behavior

## Testing
- New test passes and verifies proper exception handling
- No breaking changes to existing functionality


rwinch:
Thank you for the pull request @chanbinme! I've pushed a fix to the checkstyle and as soon as this build finishes, this will automatically be merged.

In the future, you can catch any errors using `./gradlew format && ./gradlew check` as outlined in [CONTRIBUTING](https://github.com/spring-projects/spring-security/blob/main/CONTRIBUTING.adoc#source-code-style)

chanbinme:
Thank you so much for fixing the checkstyle issue and for the helpful guidance, @rwinch ! I'll make sure to use `./gradlew format && ./gradlew check` before submitting future PRs. Really appreciate your time and support! 🙏

PR #17310, opened 2025-06-18
Apply missing diamond operators

ngocnhan-tran1996:
This PR includes
- Add diamond operator


jzheaux:
Thanks for the polish, @ngocnhan-tran1996! This will merge once the build completes.

Issue #17299, opened 2025-06-18
Remove deprecated elements using AuthorizationDecision


jzheaux:
They are replaced by methods and constructors that use `AuthorizationResult`.

PR #17322, opened 2025-06-20
Remove deprecated elements using AuthorizationDecision

ngocnhan-tran1996:
Closes gh-17299


jzheaux:
That's for helping out with more deprecations, @ngocnhan-tran1996. This will merge into `main` once the build completes.

Issue #17302, opened 2025-06-18
Remove Nimbus(Reactive)OpaqueTokenIntrospector


jzheaux:
It has been superceded by `Spring(Reactive)OpaqueTokenIntrospector`

jzheaux:
Closing in favor of https://github.com/spring-projects/spring-security/issues/17302

PR #17326, opened 2025-06-21
Remove Nimbus(Reactive)OpaqueTokenIntrospector

ngocnhan-tran1996:
Closes gh-17302


config/src/test/java/org/springframework/security/config/http/OAuth2ResourceServerBeanDefinitionParserTests.java
```diff
@@ -963,14 +965,29 @@ private void mockWebServer(String response) {
 			.setBody(response));
 	}
 
-	private void mockRestOperations(String response) {
+	private void mockJwksRestOperations(String response) {
```
jzheaux, line 968: Are you able to place this change in a separate commit? In this way, the deprecation commit only contains the changes needed for deprecation to happen.

jzheaux (changes requested):
Thanks, @ngocnhan-tran1996 for the PR! I've left my feedback inline.

config/src/test/java/org/springframework/security/config/http/OAuth2ResourceServerBeanDefinitionParserTests.java
ngocnhan-tran1996: Thanks for your feedback. I've updated it

jzheaux (approved):

jzheaux:
Thanks, @ngocnhan-tran1996! This will merge into `main` once the build passes.

Issue #13068, opened 2023-04-18
Remove deprecations


marcusdacoregio:
The following are deprecation tasks and we should make sure that we remove them in 7.0.

- [ ] https://github.com/spring-projects/spring-security/issues/13067

ASDasd341:
can i look into this?

marcusdacoregio:
Hi @ASDasd341, thanks for volunteering to take it.

I am afraid that this is not the right time to start working on this issue since the replacement for some deprecation might not still be in place and some things might still change.

In addition to that, there is no branch for 7.0 so the PR would stay open for too long.

sjohnr:
@jzheaux you may consider adding https://github.com/spring-projects/spring-security/issues/16909 to this issue as well.

PR #17330, opened 2025-06-22
Removed deprecated classes

ronodhirSoumik:
closes [#13068](https://github.com/spring-projects/spring-security/issues/13068)


jzheaux:
Thanks, @ronodhirSoumik. This looks great. When you squash, will you please ensure that each class that is removed is in its own commit? In this way, if one class removal causes a problem, we can easily revert it.

ronodhirSoumik:
> Thanks, @ronodhirSoumik. This looks great. When you squash, will you please ensure that each class that is removed is in its own commit? In this way, if one class removal causes a problem, we can easily revert it.

Got it. I will finalize this PR with four class-specific commits.

ronodhirSoumik:
@jzheaux requesting for final review

jzheaux:
Thanks, @ronodhirSoumik! I've updated the commit messages to contain the ticket they are addressing. This will merge into `main` once the build completes.

Issue #17180, opened 2025-05-28
PathPatternRequestMatcher should check method for equals and hashCode


bartvr:
The PathPatternRequestMatcher has been introduced since spring security 6.5.
However the equals() method only takes into account the pattern, while extra criteria exists like method, servlet path. 
Shouldn't the equals() and hashCode() be extended to use these extra fields as well ?

This equals() logic might give issues, a dummy example could be to use a websecurity customizer, and 
ignore 2 entries e.g. path pattern(/foobar/, HEAD) and path pattern (/foobar/, PUT). Such config can lead to errors
like UnreachableFilterChainException due to duplicates (both have same pattern and http method is ignored). 

With an adapted equals() method, these would be seen as different request matchers.
And hence no errors like UnreachableFilterchainException will be raised. Also see some
other implementations like the mvc request matcher that does take into account the method.

jzheaux:
Sure, @therepanic, a PR would be most welcome. Will you please base the PR off the `6.5.x` branch?

therepanic:
Thanks, @jzheaux, no problem

jzheaux:
Closed in favor of https://github.com/spring-projects/spring-security/pull/17337

PR #17337, opened 2025-06-23
Fix equals and hashCode in `PathPatternRequestMatcher` to include HTTP method

therepanic:
Fix: #17180


jzheaux:
Thanks, @therepanic! This was merged into `6.5.x` and `main`. It will go out in the next 6.5 point release and in the next 7.0 milestone.

PR #17381, opened 2025-06-27
Allow multiple ServerLogoutHandler instances in ServerHttpSecurity

blake-bauman:
Spring Security for Spring MVC allows for specifying multiple LogoutHandler implementations which get wrapped in a DelegatingLogoutHandler. Spring Security for WebFlux currently only allows a single ServerLogoutHandler implementation.

This PR puts both Spring MVC and Spring WebFlux on equal functionality when it comes to logout handlers.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
@@ -3033,7 +3033,8 @@ public class ServerHttpSecurity {
 
 		/**
 		 * Configures the logout handler. Default is
-		 * {@code SecurityContextServerLogoutHandler}
+		 * {@code SecurityContextServerLogoutHandler}. This clears any previous handlers
+		 * configured.
 		 * @param logoutHandler
 		 * @return the {@link LogoutSpec} to configure
 		 */
@@ -3046,6 +3047,18 @@ public class ServerHttpSecurity {
 		private LogoutSpec addLogoutHandler(ServerLogoutHandler logoutHandler) {
 			Assert.notNull(logoutHandler, "logoutHandler cannot be null");
 			this.logoutHandlers.add(logoutHandler);
+			return this;
+		}
+
+		/**
+		 * Allows managing the list of {@link ServerLogoutHandler} instances.
+		 * @param handlersConsumer {@link Consumer} for managing the list of handlers.
+		 * @return the {@link LogoutSpec} to configure
+		 * @since 7.0
+		 */
+		public LogoutSpec logoutHandler(Consumer<List<ServerLogoutHandler>> handlersConsumer) {
+			Assert.notNull(handlersConsumer, "consumer cannot be null");
+			handlersConsumer.accept(this.logoutHandlers);
 			return this;
 		}
 
--- a/config/src/test/java/org/springframework/security/config/web/server/LogoutSpecTests.java
+++ b/config/src/test/java/org/springframework/security/config/web/server/LogoutSpecTests.java
@@ -16,18 +16,27 @@
 
 package org.springframework.security.config.web.server;
 
+import org.jspecify.annotations.Nullable;
 import org.junit.jupiter.api.Test;
 import org.openqa.selenium.WebDriver;
+import reactor.core.publisher.Mono;
 
+import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
 import org.springframework.security.config.annotation.web.reactive.ServerHttpSecurityConfigurationBuilder;
+import org.springframework.security.core.context.SecurityContext;
 import org.springframework.security.htmlunit.server.WebTestClientHtmlUnitDriverBuilder;
 import org.springframework.security.test.web.reactive.server.WebTestClientBuilder;
 import org.springframework.security.web.server.SecurityWebFilterChain;
+import org.springframework.security.web.server.authentication.logout.ServerLogoutHandler;
+import org.springframework.security.web.server.context.ServerSecurityContextRepository;
 import org.springframework.security.web.server.context.WebSessionServerSecurityContextRepository;
 import org.springframework.security.web.server.util.matcher.ServerWebExchangeMatchers;
 import org.springframework.test.web.reactive.server.WebTestClient;
+import org.springframework.util.LinkedMultiValueMap;
+import org.springframework.util.MultiValueMap;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.server.ServerWebExchange;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.springframework.security.config.Customizer.withDefaults;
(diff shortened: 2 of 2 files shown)


config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -3039,7 +3040,12 @@ public LogoutSpec logoutHandler(ServerLogoutHandler logoutHandler) {
 			return addLogoutHandler(logoutHandler);
 		}
 
-		private LogoutSpec addLogoutHandler(ServerLogoutHandler logoutHandler) {
+		/**
+		 * Adds a logout handler in the last position.
+		 * @param logoutHandler
+		 * @return the {@link LogoutSpec} to configure
+		 */
+		public LogoutSpec addLogoutHandler(ServerLogoutHandler logoutHandler) {
```
jzheaux: Instead of exposing `addLogoutHandler`, would `logout(Consumer<List<ServerLogoutHandler>> consumer)` also service your needs? The reason this is nice it because it also allows you to remove values. Please see `OneTimeTokenLogoutSpec#authenticationSuccessHandler` for an example.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -3039,7 +3040,12 @@ public LogoutSpec logoutHandler(ServerLogoutHandler logoutHandler) {
 			return addLogoutHandler(logoutHandler);
 		}
 
-		private LogoutSpec addLogoutHandler(ServerLogoutHandler logoutHandler) {
+		/**
+		 * Adds a logout handler in the last position.
+		 * @param logoutHandler
+		 * @return the {@link LogoutSpec} to configure
```
jzheaux, line 3056: Will you please add `@since 7.0`

jzheaux (changes requested):
Thanks, @blake-bauman, for the PR! In addition to my inline feedback, will you please add some tests to confirm that the new functionality works?

jzheaux:
@blake-bauman, please also make sure to [sign your commit](https://github.com/apps/dco) and keep the commit title to about 50 characters, for example:

```
Support Multiple ServerLogoutHandlers

This commit adds support to ServerHttpSecurity for registering
multiple ServerLogoutHandlers. This is handy so that an application
does not need to re-supply any handlers already configured by
the DSL
```

blake-bauman:
I'm a little confused because I did sign it using the `-s`

Screenshot 2025-07-23 at 4 58 43 PM

blake-bauman:
@jzheaux I didn't see any existing tests for the Logout handler.  Could you point to where they might be so that I can add on?

blake-bauman:
> Instead of exposing addLogoutHandler, would logout(Consumer<List<ServerLogoutHandler>> consumer) also service your needs?

@jzheaux The reason I did it this way is because `LogoutConfigurer` for Servlet has `addLogoutHandler()` as public, and was attempting to match the spec for consistency.  However, if you'd prefer that I switch to using `Consumer<List<...>>`, I can do that instead.

jzheaux:
@blake-bauman, thanks for the updates. Here is the error message from the DCO bot:

```
Commit sha: [8fea725](https://github.com/spring-projects/spring-security/pull/17381/commits/8fea725b8c527233224b71bb5d01e3329f1feb41), Author: blake_bauman, Committer: blake_bauman; Expected "blake_bauman [blake_bauman@apple.com](mailto:blake_bauman@apple.com)", but got "Blake Bauman [blake_bauman@apple.com](mailto:blake_bauman@apple.com)".
```

It looks like it's complaining about "blake_bauman" being different from "Blake Bauman".

jzheaux:
> The reason I did it this way is because LogoutConfigurer for Servlet has addLogoutHandler() as public, and was attempting to match the spec for consistency.

I can see that, it makes sense. My guess is that the addLogoutHandler method is there bc it is an artifact from copy-pasting from LogoutHandlerConfigurer when LogoutSpec was first being written. That was before my time, though, so I'm not sure. 

Either way, yes, the consumer is prefererred. This allows complete control over teh default list without also needing to add something like `ObjectPostProcessor` into the DSL.

jzheaux:
> @jzheaux I didn't see any existing tests for the Logout handler. Could you point to where they might be so that I can add on?

I don't think there are any yet. You can add them in `LogoutSpecTests`.

blake-bauman:
@jzheaux Sorry for the delay.  Been on vacation.  Made the changes you requested.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -3088,7 +3101,7 @@ private ServerLogoutHandler createLogoutHandler() {
 				this.DEFAULT_LOGOUT_HANDLER.setSecurityContextRepository(securityContextRepository);
 			}
 			if (this.logoutHandlers.isEmpty()) {
-				return null;
+				return this.DEFAULT_LOGOUT_HANDLER;
```
jzheaux: Sorry I missed this earlier. Will you please place this polish in a separate commit so that the feature commit contains only what is needed?

Also I believe at that point `createLogoutHandler()` cannot return `null`; as such, the null check in `configure` is unnecessary; would you please update that as part of this cleanup?

jzheaux (changes requested):
Thanks, @blake-bauman! I've left some additional feedback inline.

Also, I think we are close enough to squash your commits. Will you please ensure that there are two commits:

1. for this feature
2. for the cleanup regarding the default logout handler null check (please see my inline comment)

Also, please double-check your DCO signature. The check is failing with the following error:

> Commit sha: [8fea725](https://github.com/spring-projects/spring-security/pull/17381/commits/8fea725b8c527233224b71bb5d01e3329f1feb41), Author: blake_bauman, Committer: blake_bauman; Expected "blake_bauman [blake_bauman@apple.com](mailto:blake_bauman@apple.com)", but got "Blake Bauman [blake_bauman@apple.com](mailto:blake_bauman@apple.com)".

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
blake-bauman: Looking at this again, I'm thinking it might be better to revert it.  It made sense in the original commit with addLogoutHandler(), but now with the `Consumer`, it may make more sense to leave it as `null` which will disable the logout handler in configure().  Is that alright?

jzheaux: Yes, I think that's fine.

jzheaux:
No problem, @blake-bauman. DCO is a new thing for a lot of people and can be confusing.

The issue was that there were three commits, but only one of them was signed in a way that matched the commit author. The merge commit was unsigned and the initial commit used "Blake Bauman" instead of "blake_bauman".

To fix this, I rebased with `main` to remove the merge commit and then I squashed your two commits together, preserving the signature of the latest commit.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
blake-bauman: @jzheaux I'm not sure what's happening.  I'm trying to revert this, and I'm able to commit the change.  but as soon as I squash the commits, the change goes away.  I'm still working on it.

blake-bauman:
Ok, I think I got it now and it's ready.  I reverted the change to the returning null and it should be only 1 commit now, with a proper signoff.

jzheaux:
Thanks again, @blake-bauman! This will merge into `main` once the build completes.

blake-bauman:
NP.  Thanks for merging this!

Issue #17447, opened 2025-07-02
Allow custom conversation for enum values in meta annotation expression templates


mheath:
Arrays of enums in security meta annotations have limited utility because there's no clean mechanism to convert the enum to a custom value.

If I have an enum that specifies my permissions, such as:

```java
enum Permission {
  READ("permission.read"),
  WRITE("permission.write");

  private final value;

  // constructor implementation

  @Override
  public String toString() {
    return value;
  }
}
```

I cannot have a working meta annotation with an array of enums such as the following:

```java
@Documented
@Retention(RetentionPolicy.RUNTIME)
@Target({ ElementType.TYPE, ElementType.METHOD })
@PreAuthorize("hasAnyAuthority({permissions})")
@interface HasAnyPermission {
 Permission[] permissions();
}
```

The `ExpressionTemplateSecurityAnnotationScanner.conversionService` includes `EnumToStringConverter` that will always use `Enum.name()` instead of my overridden `toString()` method.

Modifying the static `conversionService` initializer in `ExpressionTemplateSecurityAnnotationScanner` to:
```java
static {
	conversionService.addConverter(new ClassToStringConverter());
	conversionService.removeConvertible(Enum.class, String.class);
}
```
corrects the issue.

Since the default `Enum.toString()` returns the enum's name, removing the explicit Enum converter would be backwards compatible for most code. I'm sure there's a more clever solution that ensures backwards compatibility.

PR #17448, opened 2025-07-02
Add ExpressionTemplateValueProvider

mheath:
Implements a simple mechanism for providing custom expression template values from enums used in expression templates.

Closes gh-17447

### DIFF
--- a/core/src/main/java/org/springframework/security/core/annotation/ExpressionTemplateValueProvider.java
+++ b/core/src/main/java/org/springframework/security/core/annotation/ExpressionTemplateValueProvider.java
@@ -1,0 +1,35 @@
+package org.springframework.security.core.annotation;
+
+/**
+ * Provides a mechanism for providing custom values from enum types used in security
+ * meta-annotation expressions. For example:
+ *
+ * <pre>
+ * enum Permission implements ExpressionTemplateValueProvider {
+ *   READ,
+ *   WRITE;
+ *
+ *   &#64;Override
+ *   public String getExpressionTemplateValue() {
+ *     return switch (this) {
+ *       case READ -> "user.permission-read";
+ *       case WRITE -> "user.permission-write";
+ *     }
+ *   }
+ *
+ * }
+ * </pre>
+ *
+ * @since 6.5
+ * @author Mike Heath
+ */
+public interface ExpressionTemplateValueProvider {
+
+	/**
+	 * Returns the value to be used in an expression template.
+	 *
+	 * @return the value to be used in an expression template
+	 */
+	String getExpressionTemplateValue();
+
+}
--- a/core/src/main/java/org/springframework/security/core/annotation/ExpressionTemplateValueProvider.java
+++ b/core/src/main/java/org/springframework/security/core/annotation/ExpressionTemplateValueProvider.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2004-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.springframework.security.core.annotation;
 
 /**
@@ -20,14 +36,13 @@ package org.springframework.security.core.annotation;
  * }
  * </pre>
  *
- * @since 6.5
  * @author Mike Heath
+ * @since 7.0
  */
 public interface ExpressionTemplateValueProvider {
 
 	/**
 	 * Returns the value to be used in an expression template.
-	 *
 	 * @return the value to be used in an expression template
 	 */
 	String getExpressionTemplateValue();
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/springframework/security/core/annotation/ExpressionTemplateSecurityAnnotationScannerTests.java
```diff
@@ -54,6 +54,43 @@ void parseMultipleMetaSourceAnnotationParameterWithAliasFor() throws Exception {
 		assertThat(preAuthorize.value()).isEqualTo("check(#name)");
 	}
 
+	@Test
+	void parseMetaSourceAnnotationWithEnumImplementingExpressionTemplateValueProvider() throws Exception {
+		Method method = MessageService.class.getDeclaredMethod("process");
+		PreAuthorize preAuthorize = this.scanner.scan(method, method.getDeclaringClass());
+		assertThat(preAuthorize.value()).isEqualTo("hasAnyAuthority(user.READ,user.WRITE)");
```
jzheaux: The resulting expression may not be useful since SpEL may try and interpret `user` as a method on the root object. I think it should be:

```java
hasAnyAuthority('user.READ','user.WRITE')
```

which I believe you can achieve by placing single quotes in `Permission#getExpressionTemplateValue`

core/src/main/java/org/springframework/security/core/annotation/ExpressionTemplateValueProvider.java
```diff
@@ -0,0 +1,35 @@
...
+ *     return switch (this) {
+ *       case READ -> "user.permission-read";
+ *       case WRITE -> "user.permission-write";
+ *     }
+ *   }
+ *
+ * }
+ * </pre>
+ *
+ * @since 6.5
```
jzheaux: Will you update this to 7.0 now that that's the target release?

core/src/main/java/org/springframework/security/core/annotation/ExpressionTemplateValueProvider.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * <pre>
+ * enum Permission implements ExpressionTemplateValueProvider {
+ *   READ,
+ *   WRITE;
+ *
+ *   &#64;Override
+ *   public String getExpressionTemplateValue() {
+ *     return switch (this) {
+ *       case READ -> "user.permission-read";
```
jzheaux, line 31: While not a guarantee, I imagine that most folks will want to inject a value as a string, which will mean applying single quotes around it. Because it won't work to do `'{permissions}'` in an expression, it seems preferable to always surround the value in single quotes, thus allowing a template to do `{permissions}` or `{permission}` and still have the template compile.

If you agree with that, will you please place single quotes around each returned value?

jzheaux (changes requested):
Thanks, @mheath, for the PR! It's nice to hear from you again. I've left some feedback inline.

core/src/test/java/org/springframework/security/core/annotation/ExpressionTemplateSecurityAnnotationScannerTests.java
mheath: Good catch. Even though there's no SpEL evaluation, always best to produce valid SpEL expressions. :)

core/src/main/java/org/springframework/security/core/annotation/ExpressionTemplateValueProvider.java
mheath: I pushed up a commit with this change.

core/src/main/java/org/springframework/security/core/annotation/ExpressionTemplateValueProvider.java
mheath: Actually, our use case that drove me to make this PR involves producing SpEL type references to the enum so there are cases where the template value will not be a string.

mheath:
Hey Josh. Thanks for taking a look at my PR. It's been too long. We need to catch up sometime in more than just PR comments!

core/src/test/java/org/springframework/security/core/annotation/ExpressionTemplateSecurityAnnotationScannerTests.java
mheath: This is fixed now.

core/src/main/java/org/springframework/security/core/annotation/ExpressionTemplateValueProvider.java
jzheaux: Interesting, okay thank you.

jzheaux:
Thanks for the updates, @mheath. Are you able to sign your commit? 

(Note that I rebased, squashed, and added a polish commit before I noticed. Apologies, you may need to reset in order to push a change to your commit message.)

mheath:
Yep. My commit is signed.

jzheaux:
This is what the CI sees:

> Commit sha: [0b963d3](https://github.com/spring-projects/spring-security/pull/17448/commits/0b963d35d3b93c7e3415d081e552dc75accba303), Author: Mike Heath, Committer: Mike Heath; The sign-off is missing.

image

In addition to the title and the reference to the ticket number, the commit message should say something like:

```
Signed-off-by: ...
```

I'm not sure why the commit in the PR doesn't have that. I'm sure you are already familiar, but just for easy reference, here is where GitHub [reviews the DCO process](https://github.com/apps/dco).

mheath:
Oops. I forgot that --gpg-sign doesn't add the DCO line. It's there now.

jzheaux:
Excellent, @mheath. This will merge once the build completes. Thanks again!

Issue #15156, opened 2024-05-24
Should OidcIdToken implement equals?


MatthiasWinzeler:
**Describe the bug**
I wondered if [`OidcIdToken`](https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/oidc/OidcIdToken.java) should implement `equals`. While running some test, I realized that the `claims` on `OidcIdToken` are not compared when using `assertEquals`. 

Interestingly, [`OAuth2AccessToken`](https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2AccessToken.java) which follows the same pattern does implement it.

**Expected behavior**
`claims` (and potentially other fields) are respected when comparing via `equals`. This might be considered a breaking change.

jzheaux:
Hi, @MatthiasWinzeler. It may be valuable, though I'm not sure yet. Usually, we try and wait to add code until there is a clear reason. Is there something that you are trying to do in your production code that needs or is expecting `OidcIdToken#equals` to compare the claims?

If not, then I suggest we close this for now and reopen once the need arises.

> This might be considered a breaking change.

That's correct. Unless it is tied to problematic behavior, we'd defer this change to Spring Security 7.

MatthiasWinzeler:
@jzheaux For me, the benefit would mainly be in testing - so that `assertEquals` can be used. I was recently building an application based on [Spring Authorization Server](https://spring.io/projects/spring-authorization-server) and tried to compare its [OAuth2Authorization](https://github.com/spring-projects/spring-authorization-server/blob/main/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2Authorization.java#L55) (which holds `OidcIdToken` next to many other fields). `assertEquals` for two `OAuth2Authorization`s was failing (since the ID token claims were not the same) but it was very hard to track down the difference. Every other field (for example the `OAuthAccessToken` was properly implementing `equals`, so I thought it was just forgotten on `OidcIdToken`.

That's my reason - I think there might be others. I am far away from state of the art Java best practices these days, but isn't it a [code smell to not override equals when a subclass adds fields](https://rules.sonarsource.com/java/type/Code%20Smell/RSPEC-2160/)?

jzheaux:
[...]

PR #17485, opened 2025-07-04
Implement equals and hashCode in `OidcIdToken`

therepanic:
Closes: #15156

### DIFF
--- a/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/oidc/OidcIdToken.java
+++ b/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/oidc/OidcIdToken.java
@@ -73,6 +73,23 @@ public class OidcIdToken extends AbstractOAuth2Token implements IdTokenClaimAcce
 		return this.claims;
 	}
 
+	@Override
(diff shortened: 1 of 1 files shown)


oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/oidc/OidcIdToken.java
```diff
@@ -73,6 +73,23 @@ public Map<String, Object> getClaims() {
 		return this.claims;
 	}
 
+	@Override
+	public boolean equals(Object obj) {
+		if (this == obj) {
```
ngocnhan-tran1996, line 78: I think pattern matching can be applied here

```
if (obj instanceof OidcIdToken that) {
	return this.claims.equals(that.claims);
}

return false;
```

therepanic: I don't think there's any point in that. You can implement the method in different ways, but make it perform the same task. I copied and polished these methods from a neighboring class, so I don't think anything needs to be changed.

Issue #17321, opened 2025-06-20
Add lambda DSL method for featurePolicy


jzheaux:
Even though `Permission-Policy` is targeted to replace `Feature-Policy`, browsers still [largely support `Feature-Policy`](https://caniuse.com/feature-policy), not [`Permission-Policy`](https://caniuse.com/permissions-policy). Further , [`Permission-Policy` is still in draft](https://w3c.github.io/webappsec-permissions-policy/).

As such, we should add the lambda DSL for featurePolicy to `HeadersConfigurer` and to `ServerHttpSecurity`. We should also consider removing deprecation markers for `Feature-Policy` from the Servlet, Reactive, and Kotlin DSLs.

Related #9262

therepanic:
Hi, @jzheaux! You marked this as `ideal-for-contribution`, so I guess I can try to solve this issue?

jzheaux:
For sure, @therepanic, thanks for volunteering!

DeepDhamala:
@jzheaux  Totally agree — especially given the case you presented and the fact that Permission-Policy is still an experimental technology, the deprecation of Feature-Policy feels premature.

@therepanic Have you had a chance to start on this? If not, I’d be happy to work on adding a DSL for Feature-Policy.

therepanic:
Hi, @DeepDhamala. I already working on this, thanks

PR #17492, opened 2025-07-07
Add lambda DSL method for featurePolicy

therepanic:
Closes: gh-17321

### DIFF
--- a/web/src/main/java/org/springframework/security/web/header/writers/FeaturePolicyHeaderWriter.java
+++ b/web/src/main/java/org/springframework/security/web/header/writers/FeaturePolicyHeaderWriter.java
@@ -43,6 +43,12 @@ public final class FeaturePolicyHeaderWriter implements HeaderWriter {
 	private String policyDirectives;
 
 	/**
+	 * Create a new instance of {@link FeaturePolicyHeaderWriter}
+	 */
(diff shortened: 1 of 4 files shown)


web/src/main/java/org/springframework/security/web/header/writers/FeaturePolicyHeaderWriter.java
```diff
@@ -42,6 +42,12 @@ public final class FeaturePolicyHeaderWriter implements HeaderWriter {
 
 	private String policyDirectives;
 
+	/**
+	 * Create a new instance of {@link FeaturePolicyHeaderWriter}
+	 */
+	public FeaturePolicyHeaderWriter() {
+	}
+
```
therepanic, line 50: `PermissionsPolicyHeaderWriter` has an empty constructor, while `FeaturePolicyHeaderWriter` does not. To create a method with a customizer, I created an empty constructor for it. Is that okay?

DeepDhamala:
I think it's reasonable that you added the constructor. The default constructor was originally absent to ensure that FeaturePolicyHeaderWriter couldn't be instantiated without explicitly providing policyDirectives at the time of instantiation.
Now that we're requiring the Customizer to be passed via the DSL, it makes sense that policyDirectives must be set using .setPolicyDirectives(...).

Issue #13259, opened 2023-06-01
Customize when UserInfo is called


youlabi:
**Describe the bug**
org.springframework.security.oauth2.client.oidc.userinfo.OidcReactiveOAuth2UserService#getUserInfo calls OidcUserRequestUtils::shouldRetrieveUserInfo that uses the scopes in the OAuth2AccessToken to determine whether it should fetch user infos or not. 

In the non-reactive OidcUserService shouldRetrieveUserInfo was extended to return true if either, the access token has no scopes or the accessibleScopes is empty:
https://github.com/spring-projects/spring-security/commit/fde26e003a648a670be40c690a028970f0d1137b

This fix was not applied to the reactive version

**To Reproduce**
Set up OIDC server to return an Opaque Token, which automatically has no scopes. 

**Expected behavior**
Userinfo endpoint is called

**Sample**

No sample present

Reports that include a sample will take priority over reports that do not.
At times, we may require a sample, so it is good to try and include a sample up front.

youlabi:
See related issues:

- https://github.com/spring-projects/spring-security/issues/12144
- https://github.com/spring-projects/spring-security/issues/12205

sjohnr:
Thanks for the report, @youlabi! I think the `OidcReactiveOAuth2UserService` needs to be aligned with `OidcUserService` by applying both [this fix](https://github.com/spring-projects/spring-security/commit/fde26e003a648a670be40c690a028970f0d1137b) and the [addition of `accessibleScopes`](https://github.com/spring-projects/spring-security/commit/3f2108921e49fe3a751767b54615c3cd0d2c28bc).

Would you be interested in submitting a PR?

youlabi:
I can not promise anything :). Can you please forward me to some resources and things I need to know, on how to submit a PR ? 

Regarding the actual change. This seems to me like it is exactly the same logic duplicated in two places. Should this be kept this way? This seems highly inefficient, and the next fix will again have to be transferred and likely be forgotten. 

Another question I am unsure about: why does the reactive stack make use of a static class? Is this just legacy ?

sjohnr:
Hi @youlabi, thanks for the reply.

> Another question I am unsure about: why does the reactive stack make use of a static class? Is this just legacy ?

Sadly, I don't know the answer to that question. It's just however the developer chose to do it.

[...]

PR #17511, opened 2025-07-11
Address Incorrect scope map fix in Reactive service

asinghania71:
gh-12144 was resolved in commit fde26e003a648a670be40c690a028970f0d1137b
But only resolves issue for non reactive service, refactored and implemented the same fix in reactive service
https://github.com/spring-projects/spring-security/issues/12144


asinghania71:
@sjohnr would be great if you can help me get this issue addressed

jgrandja:
@asinghania71 The `5.6.x` branch is no longer supported. See [supported branches](https://spring.io/projects/spring-security#support).

Also, `OidcUserService.setAccessibleScopes()` was deprecated since `6.3.0` (gh-13259) and should not be added in `OidcReactiveOAuth2UserService`. Instead, use `setRetrieveUserInfo()` as a direct replacement, which was added in both.

I'll close this PR.

PR #17512, opened 2025-07-13
Correct `@NonNull` and `@Nullable` package name

ngocnhan-tran1996:


jzheaux:
Thanks for the correction here, @ngocnhan-tran1996. Note that the project as a whole is moving to JSpecify for nullability. However, this PR will help the tooling that Rob is using to move everything over from Spring Framework. It will merge once the build completes.

Issue #17359, opened 2025-06-25
Add getCreationTime() to SessionInformation for enhanced session lifecycle support


sweidner-rio:
**Summary**

I'd like to propose adding a `getCreationTime()` method to the `org.springframework.security.core.session.SessionInformation` interface to enhance session lifecycle visibility and control.
See also the discussion with @jgrandja in [auth_time claim doesn't show the time of the original authentication. spring-projects/spring-security#18282](https://github.com/spring-projects/spring-security/issues/18282)

**Motivation**
Currently, `SessionInformation` provides metadata such as the session ID, principal, last request time, and expiration status, but it lacks the creation timestamp of the session.

Adding `getCreationTime()` would provide key benefits for a variety of real-world use cases:

*Use Cases*

1. Auditing & Compliance
- Enable "Who logged in when?" tracking.
- Useful for audit trails in regulated environments.

2. Custom Session Expiration Rules based on creation time

3. Session Monitoring & Admin Tools
- Example session display: "User X has been logged in since 08:12, last activity at 08:55."

4. OpenID Connect Support
- Support for `auth_time` claim in ID Tokens.
- Easier implementation of the `max_age` query parameter in OIDC requests (see: [OpenID Connect Core 1.0 – §3.1.2.1](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) ).

therepanic:
I read the discussion and that's a really good feature. If no one objects, I could implement it

jgrandja:
@sweidner-rio Thanks for  logging the details. Instead of `getCreationTime()` let's introduce `getCreatedTime()`.

@therepanic Thanks for the offer. I assigned the issue to you.

sweidner-rio:
Hi @jgrandja , 
Sorry for asking, but I am a little bit confused. 
What is the status here?

jgrandja:
@sweidner-rio Apologies, but I haven't had a chance to get to this as of yet. I do plan on getting this in for the `7.1` release cycle.

jgrandja:
@sweidner-rio Closing this as a duplicate. Please see [comment](https://github.com/spring-projects/spring-security/issues/18282#issuecomment-4202018655).

PR #17513, opened 2025-07-13
Add `createdTime` field to `SessionInformation`

therepanic:
Closes: gh-17359


core/src/main/java/org/springframework/security/core/session/SessionRegistryImpl.java
```diff
@@ -133,7 +133,9 @@ public void registerNewSession(String sessionId, Object principal) {
 		if (this.logger.isDebugEnabled()) {
 			this.logger.debug(LogMessage.format("Registering session %s, for principal %s", sessionId, principal));
 		}
-		this.sessionIds.put(sessionId, new SessionInformation(principal, sessionId, new Date()));
+		Date currentDate = new Date();
+		this.sessionIds.put(sessionId, new SessionInformation(principal, sessionId, new Date(currentDate.getTime()),
+				new Date(currentDate.getTime())));
```
therepanic, line 138: I think that's okay, since we wouldn't want to have a link to the same Date?

core/src/main/java/org/springframework/security/core/session/SessionInformation.java
```diff
@@ -49,13 +50,17 @@ public class SessionInformation implements Serializable {
...
+	public SessionInformation(Object principal, String sessionId, Date lastRequest, Date createdTime) {
 		Assert.notNull(principal, "Principal required");
 		Assert.hasText(sessionId, "SessionId required");
 		Assert.notNull(lastRequest, "LastRequest required");
+		Assert.notNull(lastRequest, "CreatedTime required");
 		this.principal = principal;
 		this.sessionId = sessionId;
 		this.lastRequest = lastRequest;
+		this.createdTime = createdTime;
 	}
```
therepanic, line 64: I think that `SessionInformation` is created in `SessionRegistry`, so we don't need to maintain backward compatibility, right?

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/session/OidcSessionInformation.java
```diff
@@ -46,7 +46,11 @@ public class OidcSessionInformation extends SessionInformation {
...
 	 */
 	public OidcSessionInformation(String sessionId, Map<String, String> authorities, OidcUser user) {
-		super(user, sessionId, new Date());
+		this(sessionId, authorities, user, new Date());
+	}
+
+	private OidcSessionInformation(String sessionId, Map<String, String> authorities, OidcUser user, Date now) {
+		super(user, sessionId, new Date(now.getTime()), new Date(now.getTime()));
 		this.authorities = (authorities != null) ? new LinkedHashMap<>(authorities) : Collections.emptyMap();
 	}
```
therepanic, line 55: This is the only way to solve the problem I wrote about here https://github.com/spring-projects/spring-security/pull/17513#discussion_r2203466996

core/src/main/java/org/springframework/security/core/session/SessionInformation.java
```diff
@@ -49,13 +50,17 @@ public class SessionInformation implements Serializable {
...
 	private boolean expired = false;
 
-	public SessionInformation(Object principal, String sessionId, Date lastRequest) {
+	private final Date createdTime;
+
+	public SessionInformation(Object principal, String sessionId, Date lastRequest, Date createdTime) {
 		Assert.notNull(principal, "Principal required");
 		Assert.hasText(sessionId, "SessionId required");
 		Assert.notNull(lastRequest, "LastRequest required");
+		Assert.notNull(lastRequest, "CreatedTime required");
```
amriadeo: you're checking lastRequest again instead of createdTime

therepanic: Nice catch, @amriadeo!

jgrandja:
@therepanic Thanks for the PR but this enhancement is no longer needed. Please see [comment](https://github.com/spring-projects/spring-security/issues/18282#issuecomment-4202018655).

PR #17532, opened 2025-07-16
Update Angular documentation links in csrf.adoc

perceptron8:
Replaced `angular.io` links with their corresponding `angular.dev` URLs.
This change ensures that users referencing CSRF documentation are directed to the most current Angular resources.


jzheaux:
Thanks for the updated link, @perceptron8! This is now merged into `6.3.x`, `6.4.x`, `6.5.x`, and `main`.

PR #17540, opened 2025-07-17
Switch to JSpecify annotations

kse-music:
Closes gh-17534


rwinch:
Thanks for the PR! This was a good start for the effort needed, but we also needed to apply jspecify and enable it for all of the code vs just migrate the old annotations to the new. I've addressed this more completely in gh-17534 and thus I'm closing this issue

PR #17541, opened 2025-07-17
Switch to JSpecify annotations in spring-security-web

kse-music:
Closes gh-17535


rwinch:
Thanks for the PR! This was a good start for the effort needed, but we also needed to apply jspecify, enable it, and rethink some of the semantics of null within the APIs. I think this is probably a ticket that the team needs to work on due to the complexity of rethinking the APIs.

Issue #17543, opened 2025-07-17
UnboundIdContainer fails with TestContext


rwinch:
This appears to be related to

* https://github.com/spring-projects/spring-framework/commit/dbe89abd7b39217044b3e021b4fe740c7ef88a4b
*  https://github.com/spring-projects/spring-framework/issues/35171
* https://github.com/spring-projects/spring-framework/issues/35168

Adding `@DirtiesContext` works around the issue

PR #17546, opened 2025-07-17
Adapt `UnboundIdContainer` to `SmartLifecycle` to support `ApplicationContext` restart

therepanic:
`UnboundIdContainer` now implements `SmartLifecycle`, which allows Spring TestContext to correctly manage its lifecycle when using `pause()` and `resume()`. Thanks to this, `UnboundIdContainer` starts and stops correctly without the need for `@DirtiesContext`.

Closes: gh-17543

### DIFF
--- a/ldap/src/main/java/org/springframework/security/ldap/server/UnboundIdContainer.java
+++ b/ldap/src/main/java/org/springframework/security/ldap/server/UnboundIdContainer.java
@@ -32,15 +32,17 @@ import org.springframework.beans.factory.InitializingBean;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.Lifecycle;
+import org.springframework.context.SmartLifecycle;
 import org.springframework.core.io.Resource;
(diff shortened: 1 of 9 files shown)


ldap/src/main/java/org/springframework/security/ldap/server/UnboundIdContainer.java
```diff
@@ -142,4 +144,20 @@ public boolean isRunning() {
...
+	@Override
+	public void stop(Runnable callback) {
+		stop();
+		callback.run();
+	}
+
+	@Override
+	public int getPhase() {
+		return 0;
+	}
```
therepanic, line 161: Since we don't have anything that we can prioritize, I think neutral as 0 will work fine for us.

rwinch:
Thank you for the PR, but this will restart on the same port that was it was initially bound to. This means that we might get a port collision.

I'm going to close this as we are still sorting through the best approach with the Framework team. In the meantime, others can follow with https://github.com/spring-projects/spring-security/issues/17543

Issue #17584, opened 2025-07-22
Use final fields in equals and hashCode of PathPatternRequestMatcher


codingtim:
The equals and hashCode implementation of `PathPatternRequestMatcher` was altered in 
https://github.com/spring-projects/spring-security/pull/17337

The current implementation of `PathPatternRequestMatcher` has a non-final field `RequestMatcher method`. 
A setter is present for this field and is used in the Builder. 

It seems like a bad practice to use non-final fields inside the equals and hashCode implementation.

I suggest improving this: https://github.com/codingtim/spring-security/commit/2b5df1ceb22291160f246a0275ee48cd8f93282a

jzheaux:
Sounds great, @codingtim, and good catch. Are you able to submit this as a PR?

jzheaux:
Closing in favor of https://github.com/spring-projects/spring-security/pull/17621

PR #17621, opened 2025-07-24
Use final values in equals and hashCode

codingtim:
Closes [gh-17584](https://github.com/spring-projects/spring-security/issues/17584)


jzheaux:
Thanks for the PR, @codingtim! I'll merge and forward-port this once the build completes.

PR #17624, opened 2025-07-24
Update index.adoc

ml054:
Make sure users are aware that `User.withDefaultPasswordEncoder` is deprecated to avoid using it on production


jzheaux:
Thanks, @ml054! It would be nice to be able to annotate docs so that deprecated methods are appropriately rendered.

This PR will merge once the build completes.

PR #17656, opened 2025-08-02
Validate account status in OneTimeTokenAuthenticationProvider

### LINKED ISSUE #17655: OneTimeTokenAuthenticationProvider should validate UserDetails account status like DaoAuthenticationProvider
**Describe the bug**
OneTimeTokenAuthenticationProvider does not validate UserDetails account status checks (isEnabled, isAccountNonExpired, isAccountNonLocked). Users can authenticate with one-time tokens even when their accounts are disabled, expired, or locked. Currently, these checks must be handled manually in OneTimeTokenGenerationSuccessHandler, but we expect the AuthenticationProvider to handle these checks automatically like DaoAuthenticationProvider does.

**To Reproduce**
1. Create a UserDetails implementation that returns false for isEnabled(), isAccountNonExpired(), or isAccountNonLocked()
2. Attempt to authenticate using a valid one-time token for this user
3. Authentication succeeds despite the account status
4. Account status validation must be implemented separately in OneTimeTokenGenerationSuccessHandler

**Expected behavior**
Authentication should fail when UserDetails returns false for any of the following methods:
- isEnabled() - account is disabled
- isAccountNonExpired() - account has expired  
- isAccountNonLocked() - account is locked

therepanic:
The main problem is that `OneTimeTokenAuthenticationProvider` does not extend from `AbstractUserDetailsAuthenticationProvider`, which has a preauthentication check for user details. However, we do not need to extend from it because it does not fit the context of the class. In this regard, I decided to add my own checker to this commit, which performs a preauthentication check before authorizing the account, similar to how it is done in `AbstractUserDetailsAuthenticationProvider`. I also added a test to `OneTimeTokenAuthenticationProviderTests` that identifies this problem.

Closes: #17655

### DIFF
--- a/core/src/main/java/org/springframework/security/authentication/ott/OneTimeTokenAuthenticationProvider.java
+++ b/core/src/main/java/org/springframework/security/authentication/ott/OneTimeTokenAuthenticationProvider.java
@@ -19,21 +19,12 @@ package org.springframework.security.authentication.ott;
 import java.util.Collection;
 import java.util.HashSet;
 
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
-
-import org.springframework.context.MessageSource;
-import org.springframework.context.MessageSourceAware;
-import org.springframework.context.support.MessageSourceAccessor;
-import org.springframework.security.authentication.AccountExpiredException;
+import org.springframework.security.authentication.AccountStatusUserDetailsChecker;
 import org.springframework.security.authentication.AuthenticationProvider;
 import org.springframework.security.authentication.BadCredentialsException;
-import org.springframework.security.authentication.DisabledException;
-import org.springframework.security.authentication.LockedException;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
 import org.springframework.security.core.GrantedAuthority;
-import org.springframework.security.core.SpringSecurityMessageSource;
 import org.springframework.security.core.authority.FactorGrantedAuthority;
 import org.springframework.security.core.userdetails.UserDetails;
 import org.springframework.security.core.userdetails.UserDetailsChecker;
@@ -50,7 +41,7 @@ import org.springframework.util.Assert;
  * @author Andrey Litvitski
  * @since 6.4
  */
-public final class OneTimeTokenAuthenticationProvider implements AuthenticationProvider, MessageSourceAware {
+public final class OneTimeTokenAuthenticationProvider implements AuthenticationProvider {
 
 	private static final String AUTHORITY = FactorGrantedAuthority.OTT_AUTHORITY;
 
@@ -58,11 +49,8 @@ public final class OneTimeTokenAuthenticationProvider implements AuthenticationP
 
 	private final UserDetailsService userDetailsService;
 
-	private final Log logger = LogFactory.getLog(getClass());
-
-	private UserDetailsChecker authenticationChecks = new DefaultAuthenticationChecks();
-
-	private MessageSourceAccessor messages = SpringSecurityMessageSource.getAccessor();
+	private UserDetailsChecker userDetailsChecker = (user) -> {
+	};
 
 	public OneTimeTokenAuthenticationProvider(OneTimeTokenService oneTimeTokenService,
 			UserDetailsService userDetailsService) {
@@ -81,7 +69,7 @@ public final class OneTimeTokenAuthenticationProvider implements AuthenticationP
 		}
 		try {
 			UserDetails user = this.userDetailsService.loadUserByUsername(consumed.getUsername());
-			this.authenticationChecks.check(user);
+			this.userDetailsChecker.check(user);
 			Collection<GrantedAuthority> authorities = new HashSet<>(user.getAuthorities());
 			authorities.add(FactorGrantedAuthority.fromAuthority(AUTHORITY));
 			OneTimeTokenAuthentication authenticated = new OneTimeTokenAuthentication(user, authorities);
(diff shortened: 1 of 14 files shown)


core/src/main/java/org/springframework/security/authentication/ott/OneTimeTokenAuthenticationProvider.java
```diff
@@ -71,4 +89,39 @@ public boolean supports(Class<?> authentication) {
...
+
+	private class DefaultAuthenticationChecks implements UserDetailsChecker {
+
+		@Override
+		public void check(UserDetails user) {
+			if (!user.isAccountNonLocked()) {
+				OneTimeTokenAuthenticationProvider.this.logger
+					.debug("Failed to authenticate since user account is locked");
+				throw new LockedException(OneTimeTokenAuthenticationProvider.this.messages
+					.getMessage("AbstractUserDetailsAuthenticationProvider.locked", "User account is locked"));
```
therepanic: To be honest, I agree with you that this is a slippery slope. On the other hand, it will work.

Is there any other way to do this? Yes, we could hypothetically copy the message text personally for `OneTimeTokenAuthenticationProvider`, but I believe that this is not flexible at all.

Are there any suggestions?

core/src/main/java/org/springframework/security/authentication/ott/OneTimeTokenAuthenticationProvider.java
```diff
@@ -56,6 +74,7 @@ public Authentication authenticate(Authentication authentication) throws Authent
 		}
 		try {
```
ronodhirSoumik, line 70: the catch can be common something considering your added exceptions - like AuthenticationException
Then you can send specific exception message

therepanic: Thank you in advance for your comment.

Am I correct in assuming that you are suggesting something like this:
```java
try {...

} catch (AuthenticationException e) {
throw new BadCredentialsException(e.getMessage());
}
```
In general, I personally think that this is not a good idea, as we would like the user to know what type of exception is being thrown. Since in `check()` we throw exceptions that are primarily inherited from `AccountStatusException`. We should also understand that `BadCredentialsException` is not related to `AccountStatusException`, which I think could confuse the user. 

You can also see how this is implemented in `AbstractUserDetailsAuthenticationProvider`.

core/src/main/java/org/springframework/security/authentication/ott/OneTimeTokenAuthenticationProvider.java
therepanic: I think we should just throw messages directly.

In `AbstractUserDetailsAuthenticationProvider` for `BadCredentialsException`, we do 
```java
this.messages.getMessage(“AbstractUserDetailsAuthenticationProvider.badCredentials”, “Bad credentials”)
```
In turn, in `OneTimeTokenAuthenticationProvider`, we write this directly. Therefore, I think it would be best to simply display error messages without using `MessageSourceAccessor`.

tinystorm:
Hi, thanks for the work on this PR!

This feature would be very useful for us.
Is there any update on the status of this PR or plans for merging it?

Appreciate the effort!

core/src/main/java/org/springframework/security/authentication/ott/OneTimeTokenAuthenticationProvider.java
jzheaux: Spring Security has been slowly moving away from internationalizing its exception messages. So in this case, let's wait to use `MessageSource`.

core/src/main/java/org/springframework/security/authentication/ott/OneTimeTokenAuthenticationProvider.java
```diff
@@ -36,16 +42,21 @@
...
-	private static final String AUTHORITY = FactorGrantedAuthority.OTT_AUTHORITY;
+public final class OneTimeTokenAuthenticationProvider implements AuthenticationProvider, MessageSourceAware {
 
 	private final OneTimeTokenService oneTimeTokenService;
 
 	private final UserDetailsService userDetailsService;
 
+	private final Log logger = LogFactory.getLog(getClass());
+
+	private UserDetailsChecker authenticationChecks = new DefaultAuthenticationChecks();
```
jzheaux: We can't have the default something that creates a more strict validation and remain passive across releases. As such, let's please have the default be no checking.

To address that, we may want to extract `DefaultAuthenticationChecks` into a public class, having `AbstractUserDetailsAuthenticationProvider` refer to it. This will allow folks to call the authentication provider UserDetailsChecker setter in this class.

We can also add a ticket to update the default in 8.x for this class.

core/src/main/java/org/springframework/security/authentication/ott/OneTimeTokenAuthenticationProvider.java
```diff
@@ -79,4 +90,39 @@ public boolean supports(Class<?> authentication) {
 		return OneTimeTokenAuthenticationToken.class.isAssignableFrom(authentication);
 	}
 
+	@Override
+	public void setMessageSource(MessageSource messageSource) {
+		this.messages = new MessageSourceAccessor(messageSource);
+	}
+
+	public void setAuthenticationChecks(UserDetailsChecker authenticationChecks) {
```
jzheaux: Please add JavaDoc, including an explanation for what the default is. Also, I'd recommend calling this `setUserDetailsChecker`. This gives us room in case we later on have something that is a checker of the `Authentication`.

jzheaux: Let's please also check for null. Not everyone is checking with JSpecify, and so `Assert.notNull` is still valuable.

jzheaux (changes requested):
Thanks, @therepanic, for the PR! I've left some feedback inline.

jzheaux (approved):

jzheaux:
Thanks, @therepanic! This is now merged into `main` and will go out in the next milestone.

PR #17665, opened 2025-08-05
Warning when EnableTransactionManagement has lower precedence than EnableMethodSecurity

### LINKED ISSUE #17544: Consider warning when EnableTransactionManagement has lower precedence than EnableMethodSecurity
The 5.8 migration guide advises applications to set `@EnableTransationManagement(order = 0)` in order to ensure that it has higher priority than `@EnableMethodSecurity`.

While this is technically correct, in practice and application may have other AOP modules to consider in order to determine where transaction management should go, which Security doesn't know about.

Due to this, it may be helpful for security to warn if `@EnableTransactionManagement` has a lower precedence than `@EnableMethodSecurity`, since that's all that Security recommends. It would do this only when `prePostEnabled` is `true` and would take into account the `order` value for `@EnableTransactionManagement` and `offset` for `@EnableMethodSecurity`.

I feel it should be a warning instead of an error since `@PostAuthorize` is not as commonly used on methods with side-effects and may not be used at all in an application.

yybmion:
Adds validation to ensure `@EnableTransactionManagement` has higher precedence than `@EnableMethodSecurity`, allowing proper transaction rollback when `@PostAuthorize` fails.

- Warn when transaction management has same or lower precedence than method security
- Add comprehensive test coverage for various precedence scenarios

Fixes #17544

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/method/configuration/PrePostMethodSecurityConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/method/configuration/PrePostMethodSecurityConfiguration.java
@@ -16,13 +16,18 @@
 
 package org.springframework.security.config.annotation.method.configuration;
 
+import java.util.Map;
+
 import org.aopalliance.intercept.MethodInterceptor;
 import org.aopalliance.intercept.MethodInvocation;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 
 import org.springframework.aop.Pointcut;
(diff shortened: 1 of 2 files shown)


jzheaux:
Thanks for this PR, @yybmion. I'm having some internal conversations with the Framework team about this, so I'd like to hold off on this warning for the moment. I'll post my findings here.

jzheaux:
@yybmion, after some consultation, I think we should leave this to documentation. As such, I applied #17797 and this PR is no longer needed.

Thank you, I appreciate your efforts and sorry that we didn't move forward on this one.

Issue #15866, opened 2024-09-30
carrier thread be suspended by synchronized in RemoteJWKSet


me0106:
RemoteJWKSet use synchronized for synchronization. 
But this will suspend the carrier thread.
In some specific situations, it may cause the server to hang: 

`single carrier thread.(Single core CPU). And call itself to retrieve jwk`.

`RemoteJWKSet is deprecated`. Should we replace it with `JwkSourceBuilder`?

> SpringBoot: 3.3.1
> Java: Temurin-21.0.4+7

![image](https://github.com/user-attachments/assets/5f0dab95-56e8-4ab1-a704-8d5f23c480f3)
![image](https://github.com/user-attachments/assets/99c77866-c571-46ff-86bb-b0be78a346ec)

rwinch:
Thank you for the report. Would you be interested in submitting a pull request to replace RemoteJWKSet?

CrazyParanoid:
Hi @rwinch, in this issue we need to replace the old jose api with a new one. This is not a very simple issue, for example `JWKSetCache` is now deprecated, and it is needed for `JwkSourceBuilder`. I can think about how to solve this issue.

rwinch:
Closing as duplicate of gh-16251

frido37:
Removing the deprecated usages of RemoteJWKSet ([issue #16251](https://github.com/spring-projects/spring-security/issues/16251)) did not resolve the underlying issue: there is still a remote HTTP call performed within a critical section. While the synchronization mechanism has changed from synchronized to a ReentrantLock, the fundamental concern remains.

To address this, I’ve proposed a fix in [pull request #17669](https://github.com/spring-projects/spring-security/pull/17669).

PR #17669, opened 2025-08-07
Align NimbusJwtDecoder HTTP timeout defaults with Nimbus by setting to 500ms

frido37:
Summary
This change proposes a potential fix for [spring-security#15866](https://github.com/spring-projects/spring-security/issues/15866), which we encountered in a production environment.

Problem
We experienced a critical "stop-the-world" event lasting approximately 15 minutes due to the absence of connection and read timeouts in the default RestTemplate configuration used by NimbusJwtDecoder. Specifically, the underlying SimpleClientHttpRequestFactory does not define any timeouts, which can cause the application to hang indefinitely in certain network failure scenarios.

Current Workaround
To mitigate this in our own application, we use a JwkSetUriJwtDecoderBuilderCustomizer to configure the RestTemplate with appropriate timeouts. While this works, the current default behavior is problematic and may impact other users as well.

```
@Configuration
public class JwtDecoderConfig {

    @Bean
    public JwkSetUriJwtDecoderBuilderCustomizer jwkSetUriJwtDecoderBuilderCustomizer() {
        return builder -> {
            SimpleClientHttpRequestFactory requestFactory = new SimpleClientHttpRequestFactory();
            requestFactory.setConnectTimeout(RemoteJWKSet.DEFAULT_HTTP_CONNECT_TIMEOUT);
            requestFactory.setReadTimeout(RemoteJWKSet.DEFAULT_HTTP_READ_TIMEOUT);

            builder.restOperations(new RestTemplate(requestFactory));
        };
    }
}
```

Proposal
We suggest that reasonable default timeouts (both connect and read) should be applied to the RestTemplate used by default in NimbusJwtDecoder. This change will help prevent similar problems for others.

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
@@ -66,6 +66,7 @@ import org.springframework.http.HttpMethod;
 import org.springframework.http.MediaType;
 import org.springframework.http.RequestEntity;
 import org.springframework.http.ResponseEntity;
+import org.springframework.http.client.SimpleClientHttpRequestFactory;
 import org.springframework.security.oauth2.core.OAuth2Error;
 import org.springframework.security.oauth2.core.OAuth2TokenValidator;
 import org.springframework.security.oauth2.core.OAuth2TokenValidatorResult;
@@ -293,7 +294,7 @@ public final class NimbusJwtDecoder implements JwtDecoder {
 
 		private final Set<SignatureAlgorithm> signatureAlgorithms = new HashSet<>();
 
-		private RestOperations restOperations = new RestTemplate();
+		private RestOperations restOperations = new RestTemplateWithNimbusDefaultTimeouts();
 
 		private Cache cache = new NoOpCache("default");
 
@@ -541,6 +542,21 @@ public final class NimbusJwtDecoder implements JwtDecoder {
 
 			}
 
+		}
+
+	}
+
+	/**
+	 * A RestTemplate with timeouts configured to avoid blocking indefinitely when
+	 * fetching JWK Sets while holding the reentrantLock.
+	 */
+	private static final class RestTemplateWithNimbusDefaultTimeouts extends RestTemplate {
+
+		private RestTemplateWithNimbusDefaultTimeouts() {
+			SimpleClientHttpRequestFactory requestFactory = new SimpleClientHttpRequestFactory();
+			requestFactory.setConnectTimeout(JWKSourceBuilder.DEFAULT_HTTP_CONNECT_TIMEOUT);
+			requestFactory.setReadTimeout(JWKSourceBuilder.DEFAULT_HTTP_READ_TIMEOUT);
+			setRequestFactory(requestFactory);
 		}
 
 	}


oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -545,6 +547,21 @@ public void close() {
...
 
+	/**
+	 * A RestTemplate with timeouts configured to avoid blocking indefinitely when
+	 * fetching JWK Sets while holding the reentrantLock.
+	 */
+	private static final class  RestTemplateWithTimeouts extends RestTemplate {
+
+		private RestTemplateWithTimeouts () {
+			SimpleClientHttpRequestFactory requestFactory = new SimpleClientHttpRequestFactory();
+			requestFactory.setConnectTimeout(RemoteJWKSet.DEFAULT_HTTP_CONNECT_TIMEOUT);
```
jzheaux: Let's please set this value to `JWKSourceBuilder.DEFAULT_HTTP_CONNECT_TIMEOUT` since `RemoteJWKSet` is deprecated.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -545,6 +547,21 @@ public void close() {
...
+			requestFactory.setReadTimeout(RemoteJWKSet.DEFAULT_HTTP_READ_TIMEOUT);
```
jzheaux: Let's please set this value to `JWKSourceBuilder.DEFAULT_HTTP_READ_TIMEOUT` since `RemoteJWKSet` is deprecated.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -545,6 +547,21 @@ public void close() {
 
 	}
 
+	/**
+	 * A RestTemplate with timeouts configured to avoid blocking indefinitely when
+	 * fetching JWK Sets while holding the reentrantLock.
+	 */
+	private static final class  RestTemplateWithTimeouts extends RestTemplate {
```
jzheaux: To make the intent of this class clearer, will you please name it `RestTemplateWithNimbusDefaultTimeouts`.

Please also remove the extra whitespace between `class` and the class name.

jzheaux (changes requested):
Thanks, @frido37, for the PR! I've left some feedback inline.

jzheaux:
This is a note for myself:

As part of merging, upgrade instructions should be added to the 6.5 guide for how to update an application's configuration to use the new timeout values. In this way, they can know whether the new timeouts will be an issue before updating to 7.0.

frido37:
Thanks for the feedback. I Resolved it with the second commit.

jzheaux:
Thanks, @frido37. Will you please squash your commits and then we can merge it?

In the future, if you publish a PR on a branch instead of `main`, I can help a bit more with cleanups like that.

Please also change your commit to the following title and description:

```
Provider Default Timeouts For JWK Retrieval

Issue gh-14269
```

frido37:
I’ve squashed the commits and updated the commit message. Let me know if anything else is needed.

jzheaux (approved):

jzheaux:
Thanks, @frido37! This has been merged into `main`.

marbon87:
Hi @jzheaux and @frido37,
i stumpled upon the new default read timeout of only 500ms in flaky tests this week. From my point of view 500ms for read (not connection) timeout is too little, i would recommend 2000-3000ms. 
I could provide a pull request.

jzheaux:
[0f6f453](https://github.com/spring-projects/spring-security/commit/0f6f453ea0) raises the default read and connect timeout to 30 seconds. Hopefully, this addresses what you're describing here, @marbon87.

marbon87:
Thanks @jzheaux

PR #17672, opened 2025-08-08
Change @Bean method signature to return RsaKeyConversionServicePostProcessor instead of BeanFactoryPostProcessor

quaff:
It's friendly for Spring Boot's `@ConditionalOnMissingBean`, and:

>> When defining a Spring `@Bean` method, it is generally recommended to declare the most specific type possible as the method's return type. This means returning the concrete class of the bean, or the most specific interface that the bean implements and through which it will be referenced in the application.


config/src/main/java/org/springframework/security/config/annotation/web/reactive/WebFluxSecurityConfiguration.java
```diff
@@ -44,12 +43,13 @@
...
 /**
  * @author Rob Winch
+ * @author Yanming Zhou
  * @since 5.0
  */
 @Configuration(proxyBeanMethods = false)
 class WebFluxSecurityConfiguration {
 
-	public static final int WEB_FILTER_CHAIN_FILTER_ORDER = 0 - 100;
+	public static final int WEB_FILTER_CHAIN_FILTER_ORDER = -100;
```
jzheaux, line 52: Thanks for this polish, @quaff. Will you do me the favor of placing it in a separate commit since it is unrelated to updating the return type of the post processor?

quaff: done.

PR #17673, opened 2025-08-08
Add AuthorizationManagerFactory

sjohnr:
This PR is a work in progress. cc: @rwinch


core/src/main/java/org/springframework/security/access/expression/AbstractSecurityExpressionHandler.java
```diff
@@ -47,7 +49,9 @@ public abstract class AbstractSecurityExpressionHandler<T>
 
 	private @Nullable BeanResolver beanResolver;
 
-	private @Nullable RoleHierarchy roleHierarchy;
+	private final DefaultAuthorizationManagerFactory<T> defaultAuthorizationManagerFactory = new DefaultAuthorizationManagerFactory<>();
```
rwinch: remove the `defaultAuthorizationManagerFactory` in favor of a method named `getDefaultAuthorizationManagerFactory()`  that:

- is deprecated for removal in 8.0 in favor of `getAuthorizationManagerFactory()`. Any methods that need to access it to perform setters should be deprecated in favor of setting an already initialized (properties already set) `AuthorizationManagerFactory` directly vs accessing the properties on the current member.
- Throws an `IllegalStateException` if `authorizationManagerFactory` is not an instance of `DefaultAuthorizationManagerFactory`, but otherwise performs a cast. This ensures that the `AuthorizationManagerFactory` being used aligns with what is being configured or gotten.

core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java
```diff
@@ -38,19 +37,20 @@
...
+	private static final AuthorizationManagerFactory<?> DEFAULT_AUTHORIZATION_MANAGER_FACTORY = new DefaultAuthorizationManagerFactory<>();
 
-	private AuthenticationTrustResolver trustResolver = new AuthenticationTrustResolverImpl();
+	private final Supplier<Authentication> authentication;
 
-	private @Nullable RoleHierarchy roleHierarchy;
+	private final @Nullable T object;
 
-	private @Nullable Set<String> roles;
+	private @Nullable DefaultAuthorizationManagerFactory<T> defaultAuthorizationManagerFactory;
```
rwinch: Same feedback as the [comment](https://github.com/spring-projects/spring-security/pull/17673/files#diff-a2a6b9b39fbb2d4a8a25dd51661f0930c997fa4634b873d14dd7f1ac0511a292R52) on `AbstractSecurityExpressionHandler.defaultAuthorizationManagerFactory`.

core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java
```diff
@@ -38,19 +37,20 @@
...
  * @author Luke Taylor
  * @author Evgeniy Cheban
+ * @author Steve Riesenberg
  * @since 3.0
  */
-public abstract class SecurityExpressionRoot implements SecurityExpressionOperations {
+public abstract class SecurityExpressionRoot<T> implements SecurityExpressionOperations {
 
-	private final Supplier<Authentication> authentication;
+	private static final AuthorizationManagerFactory<?> DEFAULT_AUTHORIZATION_MANAGER_FACTORY = new DefaultAuthorizationManagerFactory<>();
```
rwinch: `DEFAULT_AUTHORIZATION_MANAGER_FACTORY` should also be removed in favor of just using `authorizationManagerFactory`. as described [below](https://github.com/spring-projects/spring-security/pull/17673/files#diff-781b390033e0d5931fb33e1612daa05296f0f3f1c7f46697d9ab77516b2e96e5R51).

core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java
```diff
@@ -38,19 +37,20 @@
  *
  * @author Luke Taylor
  * @author Evgeniy Cheban
+ * @author Steve Riesenberg
  * @since 3.0
  */
-public abstract class SecurityExpressionRoot implements SecurityExpressionOperations {
+public abstract class SecurityExpressionRoot<T> implements SecurityExpressionOperations {
```
rwinch:
```suggestion
public abstract class SecurityExpressionRoot<T extends @Nullable Object> implements SecurityExpressionOperations {
```

I think this is necessary for the generic to allow for a Nullable generic (e.g. `MyExpressionRoot <@Nullable SomeType>`)

core/src/main/java/org/springframework/security/authorization/DefaultAuthorizationManagerFactory.java
```diff
@@ -0,0 +1,145 @@
...
+import org.springframework.util.Assert;
+
+/**
+ * A factory for creating different kinds of {@link AuthorizationManager} instances.
+ *
+ * @param <T> the type of object that the authorization check is being done on
+ * @author Steve Riesenberg
+ * @since 7.0
+ */
+public final class DefaultAuthorizationManagerFactory<T> implements AuthorizationManagerFactory<T> {
```
rwinch:
```suggestion
public final class DefaultAuthorizationManagerFactory<T extends @Nullable Object> implements AuthorizationManagerFactory<T> {
```

core/src/main/java/org/springframework/security/access/expression/method/MethodSecurityExpressionRoot.java
```diff
@@ -28,22 +29,20 @@
  *
  * @author Luke Taylor
  * @author Evgeniy Cheban
+ * @author Steve Riesenberg
  * @since 3.0
  */
-class MethodSecurityExpressionRoot extends SecurityExpressionRoot implements MethodSecurityExpressionOperations {
+final class MethodSecurityExpressionRoot extends SecurityExpressionRoot<MethodInvocation>
```
rwinch:
```suggestion
class MethodSecurityExpressionRoot extends SecurityExpressionRoot<MethodInvocation>
```

Very minor, but I'd rather leave this as non-final in this PR to ensure that we are keeping things focused. I'm fine within it being final, but in a separate ticket so that it can be tracked and the change is isolated.

core/src/main/java/org/springframework/security/authorization/AuthorizationManagerFactory.java
```diff
@@ -0,0 +1,122 @@
...
+package org.springframework.security.authorization;
+
+/**
+ * A factory for creating different kinds of {@link AuthorizationManager} instances.
+ *
+ * @param <T> the type of object that the authorization check is being done on
+ * @author Steve Riesenberg
+ * @since 7.0
+ */
+public interface AuthorizationManagerFactory<T> {
```
rwinch:
```suggestion
public interface AuthorizationManagerFactory<T extends @Nullable Object> {
```

core/src/test/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandlerTests.java
```diff
@@ -182,7 +181,7 @@ public void createEvaluationContextSupplierAuthentication() {
 		verifyNoInteractions(mockAuthenticationSupplier);
 		assertThat(context.getRootObject()).extracting(TypedValue::getValue)
 			.asInstanceOf(InstanceOfAssertFactories.type(MethodSecurityExpressionRoot.class))
-			.extracting(SecurityExpressionRoot::getAuthentication)
+			.extracting(MethodSecurityExpressionRoot::getAuthentication)
```
rwinch: Is there a reason for this change in the his PR? I ask because, I try not to change any existing tests if at all possible in passive updates to ensure that nothing breaks.

core/src/test/java/org/springframework/security/access/expression/method/MethodSecurityExpressionRootTests.java
```diff
@@ -53,11 +55,13 @@ public class MethodSecurityExpressionRootTests {
 	@BeforeEach
 	public void createContext() {
 		this.user = mock(Authentication.class);
-		this.root = new MethodSecurityExpressionRoot(this.user);
+		this.root = new MethodSecurityExpressionRoot(() -> this.user, mock(MethodInvocation.class));
```
rwinch, line 57: Just a note that this is not public, so making this sort of minimal change to the tests tests is fine.

core/src/test/java/org/springframework/security/access/expression/method/MethodSecurityExpressionRootTests.java
```diff
@@ -53,11 +55,13 @@ public class MethodSecurityExpressionRootTests {
...
 		this.ctx = new StandardEvaluationContext();
 		this.ctx.setRootObject(this.root);
 		this.trustResolver = mock(AuthenticationTrustResolver.class);
-		this.root.setTrustResolver(this.trustResolver);
```
rwinch: While `setTrustResolver` is deprecated, I'd prefer to keep this change in the test as a distinct commit/ticket to ensure that we didn't break anything.

data/src/main/java/org/springframework/security/data/repository/query/SecurityEvaluationContextExtension.java
```diff
@@ -95,14 +97,12 @@ public class SecurityEvaluationContextExtension implements EvaluationContextExte
 
 	private Authentication authentication;
 
-	private AuthenticationTrustResolver trustResolver = new AuthenticationTrustResolverImpl();
+	private final DefaultAuthorizationManagerFactory<Object> defaultAuthorizationManagerFactory = new DefaultAuthorizationManagerFactory<>();
```
rwinch: Same feedback as the [comment](https://github.com/spring-projects/spring-security/pull/17673/files#diff-a2a6b9b39fbb2d4a8a25dd51661f0930c997fa4634b873d14dd7f1ac0511a292R52) on `AbstractSecurityExpressionHandler.defaultAuthorizationManagerFactory`.

core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java
rwinch:
```suggestion
public abstract class SecurityExpressionRoot<T extends @Nullable Object> implements SecurityExpressionOperations {
```

core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java
```diff
@@ -38,19 +37,20 @@
...
 
-	private final Supplier<Authentication> authentication;
+	private static final AuthorizationManagerFactory<?> DEFAULT_AUTHORIZATION_MANAGER_FACTORY = new DefaultAuthorizationManagerFactory<>();
 
-	private AuthenticationTrustResolver trustResolver = new AuthenticationTrustResolverImpl();
+	private final Supplier<Authentication> authentication;
 
-	private @Nullable RoleHierarchy roleHierarchy;
+	private final @Nullable T object;
```
rwinch:
```suggestion
	private final T object;
```

core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java
```diff
@@ -77,9 +77,11 @@ public abstract class SecurityExpressionRoot implements SecurityExpressionOperat
 	/**
 	 * Creates a new instance
 	 * @param authentication the {@link Authentication} to use. Cannot be null.
+	 * @deprecated use {@link #SecurityExpressionRoot(Supplier, Object)} instead
 	 */
+	@Deprecated(since = "7.0")
```
rwinch, line 80:
```suggestion
	@Deprecated(since = "7.0")
	@SuppressWarnings("NullAway")
```

core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java
```diff
@@ -88,44 +90,48 @@ public SecurityExpressionRoot(Authentication authentication) {
 	 * @param authentication the {@link Supplier} of the {@link Authentication} to use.
 	 * Cannot be null.
 	 * @since 5.8
+	 * @deprecated use {@link #SecurityExpressionRoot(Supplier, Object)} instead
 	 */
+	@Deprecated(since = "7.0")
```
rwinch, line 94:
```suggestion
	@Deprecated(since = "7.0")
	@SuppressWarnings("NullAway")
```

core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java
```diff
@@ -88,44 +90,48 @@ public SecurityExpressionRoot(Authentication authentication) {
...
+
+	/**
+	 * Creates a new instance that uses lazy initialization of the {@link Authentication}
+	 * object.
+	 * @param authentication the {@link Supplier} of the {@link Authentication} to use.
+	 * Cannot be null.
+	 * @param object the object being authorized
+	 * @since 7.0
+	 */
+	public SecurityExpressionRoot(Supplier<Authentication> authentication, @Nullable T object) {
```
rwinch:
```suggestion
	public SecurityExpressionRoot(Supplier<Authentication> authentication, T object) {
```

core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java
```diff
@@ -135,33 +141,38 @@ public final Authentication getAuthentication() {
...
 	}
 
 	@Override
 	public final boolean isFullyAuthenticated() {
-		Authentication authentication = getAuthentication();
-		return this.trustResolver.isFullyAuthenticated(authentication);
+		return isGranted(this.authorizationManagerFactory.fullyAuthenticated());
+	}
+
+	@SuppressWarnings({ "DataFlowIssue", "NullAway" })
```
rwinch:
```suggestion
```

The suppress is no longer needed once the member is fixed as shown https://github.com/spring-projects/spring-security/pull/17673/files#diff-781b390033e0d5931fb33e1612daa05296f0f3f1c7f46697d9ab77516b2e96e5R49

core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java
rwinch: We may get a NPE if someone uses one of the deprecated constructors and is using a non-null generic type, but I view this as a better tradeoff then changing the type of the member and the constructor argument to `@Nullable T` which means that it can always be `null`.

core/src/main/java/org/springframework/security/access/expression/method/MethodSecurityExpressionRoot.java
rwinch:
```suggestion
class MethodSecurityExpressionRoot extends SecurityExpressionRoot<MethodInvocation>
```
I prefer not to add the final in this PR. It is ok, but should be a distinct ticket so it can be properly tracked.

core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java
rwinch: This should only be nullable if the generic type was specified as nullable

core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java
rwinch: We may get a NPE if someone uses one of the deprecated constructors and is using a non-null generic type, but I view this as a better tradeoff then changing the type of the member and the constructor argument to `@Nullable T` which means that it can always be `null`.

core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java
rwinch: `object` can only be `null` if the generic type is `@Nullable`

messaging/src/main/java/org/springframework/security/messaging/access/expression/MessageSecurityExpressionRoot.java
```diff
@@ -29,7 +29,7 @@
  * @author Evgeniy Cheban
  * @since 4.0
  */
-public class MessageSecurityExpressionRoot extends SecurityExpressionRoot {
+public class MessageSecurityExpressionRoot extends SecurityExpressionRoot<Message<?>> {
```
rwinch:
```suggestion
public class MessageSecurityExpressionRoot<T extends @Nullable Object> extends SecurityExpressionRoot<Message<T>> {
```

This allows us to preserve type safety in `DefaultMessageSecurityExpressionHandler`

messaging/src/main/java/org/springframework/security/messaging/access/expression/MessageSecurityExpressionRoot.java
```diff
@@ -45,7 +45,7 @@ public MessageSecurityExpressionRoot(Authentication authentication, Message<?> m
 	 * @since 5.8
 	 */
 	public MessageSecurityExpressionRoot(Supplier<Authentication> authentication, Message<?> message) {
```
rwinch:
```suggestion
	public MessageSecurityExpressionRoot(Supplier<Authentication> authentication, Message<T> message) {
```

messaging/src/main/java/org/springframework/security/messaging/access/expression/MessageSecurityExpressionRoot.java
```diff
@@ -29,7 +29,7 @@
  * @author Evgeniy Cheban
  * @since 4.0
  */
-public class MessageSecurityExpressionRoot extends SecurityExpressionRoot {
+public class MessageSecurityExpressionRoot extends SecurityExpressionRoot<Message<?>> {
 
 	public final Message<?> message;
```
rwinch:
```suggestion
	public final Message<T> message;
```

messaging/src/main/java/org/springframework/security/messaging/access/expression/DefaultMessageSecurityExpressionHandler.java
```diff
@@ -60,14 +56,15 @@ private MessageSecurityExpressionRoot createSecurityExpressionRoot(Supplier<Auth
 			Message<T> invocation) {
 		MessageSecurityExpressionRoot root = new MessageSecurityExpressionRoot(authentication, invocation);
 		root.setPermissionEvaluator(getPermissionEvaluator());
-		root.setTrustResolver(this.trustResolver);
-		root.setRoleHierarchy(getRoleHierarchy());
+		// Since MessageSecurityExpressionRoot is of a different type, use the default
+		// factory instead of the one configured with this class.
+		root.setTrustResolver(getDefaultAuthorizationManagerFactory().getTrustResolver());
+		root.setRoleHierarchy(getDefaultAuthorizationManagerFactory().getRoleHierarchy());
```
rwinch:
```suggestion
		root.setAuthorizationManagerFactory(getAuthorizationManagerFactory());
```

If a custom `authorizationManagerFactory` is used, this does not use it. I've [fixed the generic types](https://github.com/spring-projects/spring-security/pull/17673/files#diff-f689a2b72218c39b6e9094d4ce8c7f47fd4bc841e6ccf269e1df65cbc6deb4c4R32) so that they align

web/src/main/java/org/springframework/security/web/access/expression/DefaultWebSecurityExpressionHandler.java
```diff
@@ -33,18 +32,16 @@
...
-		root.setTrustResolver(this.trustResolver);
-		root.setRoleHierarchy(getRoleHierarchy());
-		root.setDefaultRolePrefix(this.defaultRolePrefix);
+		// Since WebSecurityExpressionRoot is of a different type, use the default factory
+		// instead of the one configured with this class.
+		root.setTrustResolver(getDefaultAuthorizationManagerFactory().getTrustResolver());
+		root.setRoleHierarchy(getDefaultAuthorizationManagerFactory().getRoleHierarchy());
+		root.setDefaultRolePrefix(getDefaultAuthorizationManagerFactory().getRolePrefix());
 		return root;
 	}
```
rwinch:
```suggestion
		root.setAuthorizationManagerFactory(getAuthorizationManagerFactory());
		return root;
	}

	@Deprecated(forRemoval = true, since = "7.0")
	protected SecurityExpressionOperations createSecurityExpressionRoot(Authentication authentication,
			FilterInvocation fi) {
		RequestAuthorizationContext authorizationContext = new RequestAuthorizationContext(
				fi.getRequest());
		return createSecurityExpressionRoot(authentication, authorizationContext);
	}

```

web/src/main/java/org/springframework/security/web/access/expression/DefaultWebSecurityExpressionHandler.java
```diff
@@ -33,18 +32,16 @@
...
 		WebSecurityExpressionRoot root = new WebSecurityExpressionRoot(authentication, fi);
 		root.setPermissionEvaluator(getPermissionEvaluator());
-		root.setTrustResolver(this.trustResolver);
-		root.setRoleHierarchy(getRoleHierarchy());
-		root.setDefaultRolePrefix(this.defaultRolePrefix);
+		// Since WebSecurityExpressionRoot is of a different type, use the default factory
+		// instead of the one configured with this class.
+		root.setTrustResolver(getDefaultAuthorizationManagerFactory().getTrustResolver());
+		root.setRoleHierarchy(getDefaultAuthorizationManagerFactory().getRoleHierarchy());
+		root.setDefaultRolePrefix(getDefaultAuthorizationManagerFactory().getRolePrefix());
```
rwinch: This is a similar problem to the comment for `DefaultMessageSecurityExpressionHandler`.

To solve it, we should create a package scoped `FilterInvocationContextRoot` that has a `public final HttpServletRequest request` member and extends `SecurityExpressionRoot<FilterInvocation>`.  This method would then use `FilterInvocationContextRoot` instead of `WebSecurityExpressionRoot`.

Then, we should create a ticket to consider deprecating the authorization logic that depends on `FilterInvocation` (e.g. the SecurityExressionHandler<FilterInvocation> and `FilterInvocationContextRoot`)

sjohnr:
@rwinch Thanks for the review! Changing the generic type of `MessageSecurityExpressionRoot ` to `Message<T>` is a breaking change. Please confirm if that's something we want to do here and I will work on this and the other changes from your review.

rwinch:
@sjohnr 

> @rwinch Thanks for the review! Changing the generic type of `MessageSecurityExpressionRoot ` to `Message<T>` is a breaking change. Please confirm if that's something we want to do here and I will work on this and the other changes from your review.

I think that this can work if we update the constructor too. For example:

```java
public class MessageSecurityExpressionRoot<T> extends SecurityExpressionRoot {

	public final Message<T> message;

	public MessageSecurityExpressionRoot(Authentication authentication, Message<T> message) {
	  ...
	public MessageSecurityExpressionRoot(Supplier<Authentication> authentication, Message<T> message) {
	  ...
```

Then this will compile:

```java
Message<?> message = ...
MessageSecurityExpressionRoot root = new MessageSecurityExpressionRoot<>(authentication, message);
```

core/src/test/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandlerTests.java
sjohnr: I think this was from an earlier iteration. I'll revert.

sjohnr:
@rwinch I've working through the changes and it's looking really good. The only thing I'm noticing is that a few packages are missing nullability defaults.

* `org.springframework.security.web.access.expression` has a `package-info.java` but no `@NullMarked`. Is this intended or perhaps it should be addressed separately?
* `org.springframework.security.messaging.access.expression` is missing `package-info.java` along with all the other packages in `spring-security-messaging`.

As it stands now, classes in these packages show as having inconsistent nullability. These show up as IntelliJ inspections for me, similar to our conversation a few weeks ago. I think adding `@NullMarked` at the package level will address this, but if omitting that was intentional, let me know. I've added `@NullMarked` at the class level to show where these issues are but I will revert that after you review.

data/src/test/java/org/springframework/security/data/repository/query/SecurityEvaluationContextExtensionTests.java
```diff
@@ -102,9 +101,11 @@ public void setTrustResolverWhenNullThenIllegalArgumentException() {
 	public void setTrustResolverWhenNotNullThenVerifyRootObject() {
 		TestingAuthenticationToken explicit = new TestingAuthenticationToken("explicit", "password", "ROLE_EXPLICIT");
 		this.securityExtension = new SecurityEvaluationContextExtension(explicit);
-		AuthenticationTrustResolver trustResolver = new AuthenticationTrustResolverImpl();
+		AuthenticationTrustResolver trustResolver = mock(AuthenticationTrustResolver.class);
+		given(trustResolver.isAuthenticated(explicit)).willReturn(true);
 		this.securityExtension.setTrustResolver(trustResolver);
-		assertThat(getRoot()).extracting("trustResolver").isEqualTo(trustResolver);
+		assertThat(getRoot().isAuthenticated()).isTrue();
+		verify(trustResolver).isAuthenticated(explicit);
```
sjohnr, line 108: Note that changes in this test were required because it was reflectively checking for internal fields that no longer exist. I've changed the test to verifying behavior instead of the existence of fields.

core/src/main/java/org/springframework/security/access/expression/method/DefaultMethodSecurityExpressionHandler.java
```diff
@@ -286,16 +289,24 @@ public void setReturnObject(@Nullable Object returnObject, EvaluationContext ctx
...
+	 * @deprecated Use
+	 * {@link #setAuthorizationManagerFactory(AuthorizationManagerFactory)} instead
 	 */
-	public void setDefaultRolePrefix(String defaultRolePrefix) {
-		this.defaultRolePrefix = defaultRolePrefix;
+	@Deprecated(since = "7.0")
+	public void setDefaultRolePrefix(@Nullable String defaultRolePrefix) {
+		if (defaultRolePrefix == null) {
+			defaultRolePrefix = "";
+		}
```
sjohnr, line 312: This is another workaround for passivity, similar to stripping the role prefix. These expression handlers were allowing null or blank, but the new implementation only allows a blank prefix.

core/src/main/java/org/springframework/security/authorization/DefaultAuthorizationManagerFactory.java
```diff
@@ -0,0 +1,148 @@
...
+	private RoleHierarchy roleHierarchy = new NullRoleHierarchy();
+
+	private String rolePrefix = "ROLE_";
+
+	/**
+	 * Returns the {@link AuthenticationTrustResolver} used to check the user's
+	 * authentication.
+	 * @return the {@link AuthenticationTrustResolver}
+	 */
+	public AuthenticationTrustResolver getTrustResolver() {
```
rwinch: Rather than exposing a getter here, I'd prefer that `DefaultMethodSecurityExpressionHandler` keep a copy of the `TrustResolver` and return that. 

```java
public class DefaultMethodSecurityExpressionHandler .... {
    ...
    public void setTrustResolver(AuthenticationTrustResolver trustResolver) {
        this.trustResolver = trustResolver;
        getDefaultMethodSecurityExpressionHandler().setTrustResolver(trustResolver);
    }

    public AuthenticationTrustResolver getTrustResolver() {
        return this.trustResolver;
    }
}
```

The rational is twofold:

1) As we've seen in places like `DefaultMethodSecurityExpressionHandler`, exposing getters unnecessarily makes refactoring more difficult

2) The `trustResolver` property is deprecated and so we only need to keep the copy around until it is removed. When we remove it, I don't want to be stuck with a getter on `DefaultAuthorizationManagerFactory` that we only created for a deprecated property.

core/src/main/java/org/springframework/security/authorization/DefaultAuthorizationManagerFactory.java
```diff
@@ -0,0 +1,148 @@
...
+	public void setTrustResolver(AuthenticationTrustResolver trustResolver) {
+		Assert.notNull(trustResolver, "trustResolver cannot be null");
+		this.trustResolver = trustResolver;
+	}
+
+	/**
+	 * Returns the {@link RoleHierarchy} used to discover reachable authorities.
+	 * @return the {@link RoleHierarchy}
+	 */
+	public RoleHierarchy getRoleHierarchy() {
```
rwinch: Similar to trustResolver, I'd prefer to remove this getter and keep a copy of RoleHiearchy in the classes that need access to it.

core/src/main/java/org/springframework/security/authorization/DefaultAuthorizationManagerFactory.java
```diff
@@ -0,0 +1,148 @@
...
+	public void setRoleHierarchy(RoleHierarchy roleHierarchy) {
+		Assert.notNull(roleHierarchy, "roleHierarchy cannot be null");
+		this.roleHierarchy = roleHierarchy;
+	}
+
+	/**
+	 * Returns the prefix used to create an authority name from a role name.
+	 * @return the role prefix
+	 */
+	public String getRolePrefix() {
```
rwinch: Similar to [`trustResolver`](https://github.com/spring-projects/spring-security/pull/17673/files#diff-54ecc5093fe59d2c083f99edfc1afc91a9fc38047e7f80e44d5dbf436ababceeR48), I'd prefer to remove this getter and keep a copy of `rolePrefix` in the classes that need access to it.

data/src/test/java/org/springframework/security/data/repository/query/SecurityEvaluationContextExtensionTests.java
rwinch: That is a great change to make regardless!

etc/checkstyle/checkstyle-suppressions.xml
```diff
@@ -36,6 +36,7 @@
 	<suppress files="JoseHeader\.java" checks="SpringMethodVisibility"/>
 	<suppress files="DefaultLoginPageGeneratingFilterTests\.java" checks="SpringLeadingWhitespace"/>
 	<suppress files="AuthenticationException\.java" checks="MutableException"/>
+	<suppress files="FilterInvocationExpressionRoot\.java" checks="SpringMethodVisibility|RedundantModifier"/>
```
rwinch: I think that if you make the constructor for `FilterInvocationExpressionRoot` package-private, then you can remove `RedundantModifier`.

web/src/main/java/org/springframework/security/web/access/expression/FilterInvocationExpressionRoot.java
```diff
@@ -0,0 +1,62 @@
...
+import org.springframework.security.access.expression.SecurityExpressionRoot;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.web.FilterInvocation;
+import org.springframework.security.web.util.matcher.IpAddressMatcher;
+
+/**
+ * @author Steve Riesenberg
+ * @since 7.0
+ */
+final class FilterInvocationExpressionRoot extends SecurityExpressionRoot<FilterInvocation> {
```
rwinch, line 32: I think that this makes more sense to be called `HttpServletRequestExpressionRoot` rather than `FilterInvocationExpressionRoot` (I know I proposed using `FilterInvocationExpressionRoot`, but seeing it shows me that it doesn't make sense - sorry about that!). It would then accept just the `HttpServletRequest` instance.

web/src/main/java/org/springframework/security/web/access/expression/FilterInvocationExpressionRoot.java
```diff
@@ -0,0 +1,62 @@
...
+
+import org.springframework.security.access.expression.SecurityExpressionRoot;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.web.FilterInvocation;
+import org.springframework.security.web.util.matcher.IpAddressMatcher;
+
+/**
+ * @author Steve Riesenberg
+ * @since 7.0
+ */
```
rwinch, line 31: FYI As per this [comment](https://github.com/spring-projects/spring-security/pull/17673#discussion_r2277596690) I created gh-17781

rwinch:
We've only added nullability for crypto and core so feel free to leave as is. We'll get the rest of the module in a different ticket commit.

web/src/main/java/org/springframework/security/web/access/expression/FilterInvocationExpressionRoot.java
sjohnr: @rwinch Thanks for the feedback. If I make this change (renaming the package-private class and updating the constructor), I would also need to update the generic type. So it would become:

```
final class HttpServletRequestExpressionRoot extends SecurityExpressionRoot<HttpServletRequest> { ... }
```

The problem is that `DefaultWebSecurityExpressionHandler` already declares type `FilterInvocation`:

```
public class DefaultWebSecurityExpressionHandler extends AbstractSecurityExpressionHandler<FilterInvocation>
		implements SecurityExpressionHandler<FilterInvocation> { ... }
```

Unfortunately, this won't work, because we would be passing an `AuthorizationManagerFactory<FilterInvocation>` to a `SecurityExpressionRoot<HttpServletRequest>`, which doesn't compile.

So while I agree that it would make more sense to rename and change the type of this class, it doesn't seem worth it to need to work around the type mismatch. Since this is just a package-private class that I assume will be removed in the future, I think we should just leave it as `FilterInvocationExpressionRoot` like we originally discussed. Let me know if you have any feedback on this.

rwinch: Ok thanks for the reply. I agree with your assessment. Let's leave this as is.

config/src/main/java/org/springframework/security/config/annotation/method/configuration/PrePostMethodSecurityConfiguration.java
```diff
@@ -121,6 +123,16 @@ void setRoleHierarchy(RoleHierarchy roleHierarchy) {
 		this.expressionHandler.setRoleHierarchy(roleHierarchy);
 	}
 
+	@Autowired(required = false)
+	void setTrustResolver(AuthenticationTrustResolver trustResolver) {
+		this.expressionHandler.setTrustResolver(trustResolver);
+	}
```
sjohnr: I think I added this for consistency but technically it isn't required for this PR. Shall I remove it to keep the PR minimal?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/AuthorizeHttpRequestsConfigurer.java
```diff
@@ -81,20 +77,43 @@ public AuthorizeHttpRequestsConfigurer(ApplicationContext context) {
...
+	private AuthorizationManagerFactory<RequestAuthorizationContext> getAuthorizationManagerFactory(
+			ApplicationContext context) {
+		ResolvableType authorizationManagerFactoryType = ResolvableType
+			.forClassWithGenerics(AuthorizationManagerFactory.class, RequestAuthorizationContext.class);
+
+		// Handle fallback to generic type
+		if (context.getBeanNamesForType(authorizationManagerFactoryType).length == 0) {
+			authorizationManagerFactoryType = ResolvableType.forClassWithGenerics(AuthorizationManagerFactory.class,
+					Object.class);
+		}
```
sjohnr, line 97: FYI: I've added this check to handle the case where a generic `AuthorizationManagerFactory` is published as a bean. This is required (AFAIK) due to the manual use of `ResolvableType` here since we cannot use an `@Autowired` field in the configurer.

Here is a condensed version of what I put in the documentation example:

```java
@Bean
<T> AuthorizationManagerFactory<T> authorizationManagerFactory() {
	return new DefaultAuthorizationManagerFactory<>();
}
```

In compiled code and as a Spring bean, this looks identical to:

```java
@Bean
AuthorizationManagerFactory<Object> authorizationManagerFactory() {
	return new DefaultAuthorizationManagerFactory<>();
}
```

hence the use of `forClassWithGenerics` with `Object.class`.

sjohnr (commented):
@rwinch I've added documentation, code examples with `include-code`, and tests. I've also added two additional notes for your review below. Aside from the note below, I believe this PR is ready to merge.

I will also be creating a GH issue for the Kotlin `authorizeHttpRequests` DSL, which doesn't currently support using the new `AuthorizationManagerFactory`. I doubt I'll have time to work on that change though (if needed for 7.0).

config/src/main/java/org/springframework/security/config/annotation/method/configuration/PrePostMethodSecurityConfiguration.java
rwinch: Yes. Please remove and create a separate ticket if it isn't needed.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/AuthorizeHttpRequestsConfigurer.java
rwinch: Thanks. Do you have a realistic looking test that will fail if this is removed. I ask because this is always the best reminder of why some code exists.

rwinch:
@sjohnr Thanks for the PR. Overall, I think that we are ready to merge after you address my two replies to your comments.

> I will also be creating a GH issue for the Kotlin `authorizeHttpRequests` DSL, which doesn't currently support using the new `AuthorizationManagerFactory`. I doubt I'll have time to work on that change though (if needed for 7.0).

Thanks for the heads up. We will ensure that this gets taken care of.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/AuthorizeHttpRequestsConfigurer.java
sjohnr: Yes. The tests in the docs will fail if it is removed. I didn't have a test initially as I was focused in unit tests on just testing the direct integration with a parameterized type. If we need a test outside of docs, it would probably be relatively easy to add one.

rwinch (changes requested):
Can you please include a link in the whats-new.adoc as well?

Issue #16406, opened 2025-01-13
OAuth2ResourceServerConfigurer#authenticationManagerResolver should override #jwt


jgrandja:
Given the following configuration:

```java
@Bean
public SecurityFilterChain authorizationServerSecurityFilterChain(HttpSecurity http)
		throws Exception {

	OAuth2AuthorizationServerConfigurer authorizationServerConfigurer =
			OAuth2AuthorizationServerConfigurer.authorizationServer();

	http
		.securityMatcher(authorizationServerConfigurer.getEndpointsMatcher())
		.with(authorizationServerConfigurer, (authorizationServer) ->
			authorizationServer
				.oidc(Customizer.withDefaults())	// Enable OpenID Connect 1.0
		)
		.authorizeHttpRequests((authorize) ->
			authorize
				.anyRequest().authenticated()
		)
		.exceptionHandling((exceptions) -> exceptions
			.defaultAuthenticationEntryPointFor(
				new LoginUrlAuthenticationEntryPoint("/login"),
				new MediaTypeRequestMatcher(MediaType.TEXT_HTML)
			)
		)
		.oauth2ResourceServer(resourceServer ->
			resourceServer
				.authenticationManagerResolver((context) -> customAuthenticationManager)
		);

	return http.build();
}
```

The application context will fail to build with the error message:

> Caused by: java.lang.IllegalStateException: If an authenticationManagerResolver() is configured, then it takes precedence over any jwt() or opaqueToken() configuration.

The reason is because `OAuth2AuthorizationServerConfigurer` will [default to `resourceServer.jwt()`](https://github.com/spring-projects/spring-authorization-server/commit/2c7975485fca39338b1fa4457b84708b0296eca2#diff-e6f82bec8a38484c25e3df06dea75a70f3901b24d668195e77f5357710a08ed1R363) if the OIDC UserInfo endpoint or OIDC Client Registration endpoint is enabled. However, if an application configures a client to use opaque tokens for an OpenID Connect flow, then configuring the `authenticationManagerResolver()` should be possible if support for both JWT and Opaque access tokens is required. As of now, it's not possible since `resourceServer.jwt()` was previously configured as the default by `OAuth2AuthorizationServerConfigurer`.

A similar error condition occurs with the following configuration:

```java
@Bean
public SecurityFilterChain authorizationServerSecurityFilterChain(HttpSecurity http)
		throws Exception {

	OAuth2AuthorizationServerConfigurer authorizationServerConfigurer =
[...]

PR #17675, opened 2025-08-10
Allow authenticationManagerResolver to override jwt/opaqueToken configuration

academey:
When OAuth2ResourceServerConfigurer.authenticationManagerResolver() is configured after jwt() or opaqueToken(), it now properly overrides the previous configuration instead of throwing an exception. This follows Spring Security's convention where the last configured component takes precedence.

## Changes
- Modified `authenticationManagerResolver()` method to clear existing jwt/opaqueToken configurations when called
- Removed the validation check that prevented authenticationManagerResolver from being used with jwt/opaqueToken
- Added tests to verify the new behavior works correctly

## Testing
- Added `configureWhenUsingBothAuthenticationManagerResolverAndOpaqueThenAuthenticationManagerResolverTakesPrecedence()` test
- Added `configureWhenUsingBothAuthenticationManagerResolverAndJwtThenAuthenticationManagerResolverTakesPrecedence()` test

Closes gh-16406


academey:
Closing this PR to submit a better solution. The initial approach was too destructive (clearing existing configurations). A new PR will be submitted with a non-destructive solution that properly implements the 'last-configured wins' principle.

Issue #17764, opened 2025-08-19
Missing nullability information in AuthenticationTrustResolver


bclozel:
As of #17534, JSpecify annotations have been introduced in spring-security-core.

`SecurityContext.getAuthentication` returns a `@Nullable` value, but the `AuthenticationTrustResolver` counterpart is out of sync.

The `AuthenticationTrustResolver` interface is missing `@Nullable` annotations for `isAnonymous`, `isRememberMe` and `isFullyAuthenticated` as they all take nullable `Authentication` parameters per Javadoc.

rwinch:
Thanks for the report @bclozel! I closed this in favor of the PR gh-17774

PR #17774, opened 2025-08-19
Annotate AuthenticationTrustResolver methods with `@Nullable`

therepanic:
Since `AuthenticationTrustResolver` can handle null arguments (this is also stated in the implementation of this interface), we should mark these arguments as `@Nullable`.

Closes: gh-17764


rwinch:

Issue #16226, opened 2024-12-05
Use Include for All Samples in Documentation


rwinch:
We should use [include-code](https://github.com/spring-io/asciidoctor-extensions?tab=readme-ov-file#include-code) for all samples within the documentation. This will allow us to test the sample code to ensure that it works.

If you are wanting to contribute to this issue, please:

- Select a single adoc page that you will work on
- Check to see that it isn't already being worked on
- Comment on this ticket that you will be working on it
- Submit a pull request to the main branch with the changes using https://github.com/spring-projects/spring-security/commit/6eee256e12663a5465be6fceffe0b75178f23c37 as a guide to how to make the changes
  - However, please make the changes to the entire adoc page in a single commit / PR.
  - Before submitting the pull request, please diff the html page that is generated by `./gradlew antora` before and after your changes
  - Please also search the updated HTML output for `::` as this can find any macros that were not ran .
  - Consider adding tests if it makes sense

# PRs

- [x] gh-18856

christophsj:
Hi! I am a spring security user and I would love to contribute.

However, I do not seem to be able to create a sub issue. I am a first time contributor,  is this expected?

jkuhel:
Hi @rwinch 

I would like to help on this issue if it is still available to work on.

To get started, I did attempt to create a sub-issue to work on `docs/modules/ROOT/pages/servlet/test/method.adoc`. However, I was not able to because it looks like only accounts with at [least triage permissions](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/adding-sub-issues) for a given repository can create sub-issues. I'd be glad to create sub-issues if my account has permissions. Let me know what's best.

In the mean time, I followed your example and made changes to the above page to use include-code and added some tests. For the time being, I opened this [PR](https://github.com/spring-projects/spring-security/pull/17749) to the 6.5.x branch to start with. Please let me know if this helpful and on the right track.

If everything looks good here as a starting point, I can recreate the branch and PR once a sub-issue is created and link the commit/PR accordingly.

I can also take on the updates to additional pages and happy to collaborate/divide up the pages with others (like @christophsj).

rwinch:
@christophsj @jkuhel Thanks for the feedback. I've updated the instructions so that to claim a file you should just post it as a comment on this ticket. Then the PR should be linked to this ticket.

Also note that I added a recommendation to perform a diff on the before and after on the html page that is generated.

jkuhel:
[...]

PR #17775, opened 2025-08-19
Add Multi-factor Authentication Support

jzheaux:
Related to https://github.com/spring-projects/spring-security-samples/pull/351

Implement N authentication factors and they will be required in the order that they are declared:

```java
http
    .authorizeHttpRequests((authorize) -> authorize.anyRequest()
        .access(allOf(hasAuthority("FACTOR_PASSWORD"), hasAuthority("FACTOR_OTT")))
    )
    .formLogin(Customizer.withDefaults())
    .oneTimeTokenLogin(Customizer.withDefaults())
    // ...
```

This will ask for a username/password first and a one-time token second. Thereafter, the user will be considered sufficiently authenticated.

Note that you can also publish an `AuthorizationManagerFactory<Object>` bean that checks for `FACTOR_PASSWORD` and `FACTOR_OTT`; however, this has not been added to this PR.

You can also specify a custom action to perform when a given factor is missing:

```java
http
    .authorizeHttpRequests((authorize) -> authorize.anyRequest()
        .access(allOf(hasAuthority("FACTOR_PASSWORD"), hasAuthority("FACTOR_WEBAUTHN")))
    )
    .formLogin(Customizer.withDefaults())
    .webauthn((webauthn) -> ...)
    .exceptionHandling((exceptions) -> exceptions
        .defaultAuthenticationEntryPointFor(new LoginUrlAuthenticationEntryPoint("/webauthn"), "FACTOR_WEBAUTHN")
    )
    // ...
```

Note that authentication factors already integrate with `defaultAuthenticationEntryPointFor`  in this PR. The above is needed for WebAuthn since it doesn't expose a custom entry point page in its DSL.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
@@ -80,8 +80,7 @@ public final class ExceptionHandlingConfigurer<H extends HttpSecurityBuilder<H>>
 
 	private LinkedHashMap<RequestMatcher, AccessDeniedHandler> defaultDeniedHandlerMappings = new LinkedHashMap<>();
 
-	private final DelegatingMissingAuthorityAccessDeniedHandler.Builder missingAuthoritiesHandlerBuilder = DelegatingMissingAuthorityAccessDeniedHandler
-		.builder();
+	private DelegatingMissingAuthorityAccessDeniedHandler.@Nullable Builder missingAuthoritiesHandlerBuilder;
 
 	/**
 	 * Creates a new instance
@@ -142,8 +141,11 @@ public final class ExceptionHandlingConfigurer<H extends HttpSecurityBuilder<H>>
 	 * @return the {@link ExceptionHandlingConfigurer} for further customizations
 	 * @since 7.0
 	 */
-	public ExceptionHandlingConfigurer<H> defaultAuthenticationEntryPointFor(AuthenticationEntryPoint entryPoint,
+	public ExceptionHandlingConfigurer<H> defaultDeniedHandlerForMissingAuthority(AuthenticationEntryPoint entryPoint,
 			String authority) {
+		if (this.missingAuthoritiesHandlerBuilder == null) {
+			this.missingAuthoritiesHandlerBuilder = DelegatingMissingAuthorityAccessDeniedHandler.builder();
+		}
 		this.missingAuthoritiesHandlerBuilder.addEntryPointFor(entryPoint, authority);
 		return this;
 	}
@@ -158,8 +160,11 @@ public final class ExceptionHandlingConfigurer<H extends HttpSecurityBuilder<H>>
 	 * @return the {@link ExceptionHandlingConfigurer} for further customizations
 	 * @since 7.0
 	 */
-	public ExceptionHandlingConfigurer<H> defaultAuthenticationEntryPointFor(
+	public ExceptionHandlingConfigurer<H> defaultDeniedHandlerForMissingAuthority(
 			Consumer<DelegatingAuthenticationEntryPoint.Builder> entryPoint, String authority) {
+		if (this.missingAuthoritiesHandlerBuilder == null) {
+			this.missingAuthoritiesHandlerBuilder = DelegatingMissingAuthorityAccessDeniedHandler.builder();
+		}
 		this.missingAuthoritiesHandlerBuilder.addEntryPointFor(entryPoint, authority);
 		return this;
 	}
@@ -267,6 +272,9 @@ public final class ExceptionHandlingConfigurer<H extends HttpSecurityBuilder<H>>
 
 	private AccessDeniedHandler createDefaultDeniedHandler(H http) {
 		AccessDeniedHandler defaults = createDefaultAccessDeniedHandler(http);
+		if (this.missingAuthoritiesHandlerBuilder == null) {
+			return defaults;
+		}
 		DelegatingMissingAuthorityAccessDeniedHandler deniedHandler = this.missingAuthoritiesHandlerBuilder.build();
 		deniedHandler.setRequestCache(getRequestCache(http));
 		deniedHandler.setDefaultAccessDeniedHandler(defaults);
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
@@ -16,40 +16,26 @@
 
 package org.springframework.security.config.annotation.web.configurers;
 
-import java.io.IOException;
-import java.util.Collection;
 import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.function.Consumer;
 
-import jakarta.servlet.ServletException;
-import jakarta.servlet.http.HttpServletRequest;
-import jakarta.servlet.http.HttpServletResponse;
 import org.jspecify.annotations.Nullable;
 
-import org.springframework.security.access.AccessDeniedException;
-import org.springframework.security.authentication.InsufficientAuthenticationException;
-import org.springframework.security.authorization.AuthorityAuthorizationDecision;
-import org.springframework.security.authorization.AuthorizationDeniedException;
 import org.springframework.security.config.Customizer;
 import org.springframework.security.config.annotation.web.HttpSecurityBuilder;
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
-import org.springframework.security.core.AuthenticationException;
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.web.AuthenticationEntryPoint;
 import org.springframework.security.web.access.AccessDeniedHandler;
 import org.springframework.security.web.access.AccessDeniedHandlerImpl;
+import org.springframework.security.web.access.DelegatingMissingAuthorityAccessDeniedHandler;
 import org.springframework.security.web.access.ExceptionTranslationFilter;
 import org.springframework.security.web.access.RequestMatcherDelegatingAccessDeniedHandler;
 import org.springframework.security.web.authentication.DelegatingAuthenticationEntryPoint;
 import org.springframework.security.web.authentication.Http403ForbiddenEntryPoint;
 import org.springframework.security.web.savedrequest.HttpSessionRequestCache;
-import org.springframework.security.web.savedrequest.NullRequestCache;
 import org.springframework.security.web.savedrequest.RequestCache;
-import org.springframework.security.web.util.ThrowableAnalyzer;
-import org.springframework.security.web.util.matcher.AnyRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher;
-import org.springframework.util.Assert;
 
 /**
  * Adds exception handling for Spring Security related exceptions to an application. All
@@ -94,7 +80,8 @@ public final class ExceptionHandlingConfigurer<H extends HttpSecurityBuilder<H>>
 
 	private LinkedHashMap<RequestMatcher, AccessDeniedHandler> defaultDeniedHandlerMappings = new LinkedHashMap<>();
 
-	private Map<String, LinkedHashMap<RequestMatcher, AuthenticationEntryPoint>> entryPoints = new LinkedHashMap<>();
+	private final DelegatingMissingAuthorityAccessDeniedHandler.Builder missingAuthoritiesHandlerBuilder = DelegatingMissingAuthorityAccessDeniedHandler
+		.builder();
 
 	/**
 	 * Creates a new instance
@@ -147,6 +134,37 @@ public final class ExceptionHandlingConfigurer<H extends HttpSecurityBuilder<H>>
 	}
 
 	/**
+	 * Sets a default {@link AuthenticationEntryPoint} to be used which prefers being
+	 * invoked for the provided missing {@link GrantedAuthority}.
+	 * @param entryPoint the {@link AuthenticationEntryPoint} to use for the given
+	 * {@code authority}
+	 * @param authority the authority
+	 * @return the {@link ExceptionHandlingConfigurer} for further customizations
+	 * @since 7.0
+	 */
+	public ExceptionHandlingConfigurer<H> defaultAuthenticationEntryPointFor(AuthenticationEntryPoint entryPoint,
+			String authority) {
+		this.missingAuthoritiesHandlerBuilder.addEntryPointFor(entryPoint, authority);
+		return this;
+	}
+
+	/**
+	 * Sets a default {@link AuthenticationEntryPoint} to be used which prefers being
+	 * invoked for the provided missing {@link GrantedAuthority}.
+	 * @param entryPoint a consumer of a
+	 * {@link DelegatingAuthenticationEntryPoint.Builder} to use for the given
+	 * {@code authority}
+	 * @param authority the authority
+	 * @return the {@link ExceptionHandlingConfigurer} for further customizations
+	 * @since 7.0
+	 */
+	public ExceptionHandlingConfigurer<H> defaultAuthenticationEntryPointFor(
+			Consumer<DelegatingAuthenticationEntryPoint.Builder> entryPoint, String authority) {
+		this.missingAuthoritiesHandlerBuilder.addEntryPointFor(entryPoint, authority);
+		return this;
+	}
+
+	/**
 	 * Sets the {@link AuthenticationEntryPoint} to be used.
 	 *
 	 * <p>
@@ -186,26 +204,6 @@ public final class ExceptionHandlingConfigurer<H extends HttpSecurityBuilder<H>>
 			this.defaultEntryPoint = DelegatingAuthenticationEntryPoint.builder();
 		}
 		this.defaultEntryPoint.addEntryPointFor(entryPoint, preferredMatcher);
-		return this;
-	}
-
-	public ExceptionHandlingConfigurer<H> defaultAuthenticationEntryPointFor(AuthenticationEntryPoint entryPoint,
-			RequestMatcher preferredMatcher, String authority) {
-		this.defaultEntryPointMappings.put(preferredMatcher, entryPoint);
-		LinkedHashMap<RequestMatcher, AuthenticationEntryPoint> byMatcher = this.entryPoints.get(authority);
-		if (byMatcher == null) {
-			byMatcher = new LinkedHashMap<>();
-		}
-		byMatcher.put(preferredMatcher, entryPoint);
-		this.entryPoints.put(authority, byMatcher);
-		return this;
-	}
-
-	public ExceptionHandlingConfigurer<H> defaultAuthenticationEntryPointFor(AuthenticationEntryPoint entryPoint,
-			String authority) {
-		LinkedHashMap<RequestMatcher, AuthenticationEntryPoint> byMatcher = new LinkedHashMap<>();
-		byMatcher.put(AnyRequestMatcher.INSTANCE, entryPoint);
-		this.entryPoints.put(authority, byMatcher);
 		return this;
 	}
 
@@ -269,22 +267,10 @@ public final class ExceptionHandlingConfigurer<H extends HttpSecurityBuilder<H>>
 
 	private AccessDeniedHandler createDefaultDeniedHandler(H http) {
 		AccessDeniedHandler defaults = createDefaultAccessDeniedHandler(http);
-		if (this.entryPoints.isEmpty()) {
-			return defaults;
-		}
-		Map<String, AccessDeniedHandler> deniedHandlers = new LinkedHashMap<>();
-		for (Map.Entry<String, LinkedHashMap<RequestMatcher, AuthenticationEntryPoint>> entry : this.entryPoints
-			.entrySet()) {
-			AuthenticationEntryPoint entryPoint = entryPointFrom(entry.getValue());
-			AuthenticationEntryPointAccessDeniedHandlerAdapter deniedHandler = new AuthenticationEntryPointAccessDeniedHandlerAdapter(
-					entryPoint);
-			RequestCache requestCache = http.getSharedObject(RequestCache.class);
-			if (requestCache != null) {
-				deniedHandler.setRequestCache(requestCache);
-			}
-			deniedHandlers.put(entry.getKey(), deniedHandler);
-		}
-		return new AuthenticationFactorDelegatingAccessDeniedHandler(deniedHandlers, defaults);
+		DelegatingMissingAuthorityAccessDeniedHandler deniedHandler = this.missingAuthoritiesHandlerBuilder.build();
+		deniedHandler.setRequestCache(getRequestCache(http));
+		deniedHandler.setDefaultAccessDeniedHandler(defaults);
+		return deniedHandler;
 	}
 
 	private AccessDeniedHandler createDefaultAccessDeniedHandler(H http) {
@@ -299,29 +285,10 @@ public final class ExceptionHandlingConfigurer<H extends HttpSecurityBuilder<H>>
 	}
 
 	private AuthenticationEntryPoint createDefaultEntryPoint(H http) {
-		AuthenticationEntryPoint defaults = entryPointFrom(this.defaultEntryPointMappings);
-		if (this.entryPoints.isEmpty()) {
-			return defaults;
-		}
-		Map<String, AuthenticationEntryPoint> entryPoints = new LinkedHashMap<>();
-		for (Map.Entry<String, LinkedHashMap<RequestMatcher, AuthenticationEntryPoint>> entry : this.entryPoints
-			.entrySet()) {
-			entryPoints.put(entry.getKey(), entryPointFrom(entry.getValue()));
-		}
-		return new AuthenticationFactorDelegatingAuthenticationEntryPoint(entryPoints, defaults);
-	}
-
-	private AuthenticationEntryPoint entryPointFrom(
-			LinkedHashMap<RequestMatcher, AuthenticationEntryPoint> entryPoints) {
-		if (entryPoints.isEmpty()) {
+		if (this.defaultEntryPoint == null) {
 			return new Http403ForbiddenEntryPoint();
 		}
-		if (entryPoints.size() == 1) {
-			return entryPoints.values().iterator().next();
-		}
-		DelegatingAuthenticationEntryPoint entryPoint = new DelegatingAuthenticationEntryPoint(entryPoints);
-		entryPoint.setDefaultEntryPoint(entryPoints.values().iterator().next());
-		return entryPoint;
+		return this.defaultEntryPoint.build();
 	}
 
 	/**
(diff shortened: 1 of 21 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
```diff
@@ -263,4 +336,128 @@ private RequestCache getRequestCache(H http) {
 		return new HttpSessionRequestCache();
 	}
 
+	private static final class AuthenticationFactorDelegatingAuthenticationEntryPoint
```
rwinch: I think that we can delete `AuthenticationFactorDelegatingAuthenticationEntryPoint` because we don't need to control the log in page based upon the factors.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/X509Configurer.java
```diff
@@ -225,4 +235,28 @@ private <C> C getSharedOrBean(H http, Class<C> type) {
 		return context.getBeanProvider(type).getIfUnique();
 	}
 
+	private static final class AuthorityGrantingAuthenticationProvider implements AuthenticationProvider {
```
rwinch: Can this be removed in favor of `PreAuthenticatedAuthenticationProvider` adding the additional factor? If the factor needs to vary, perhaps allow injecting it into `PreAuthenticatedAuthenticationProvider`.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
```diff
@@ -165,6 +185,26 @@ public ExceptionHandlingConfigurer<H> defaultAuthenticationEntryPointFor(Authent
 		return this;
 	}
 
+	public ExceptionHandlingConfigurer<H> defaultAuthenticationEntryPointFor(AuthenticationEntryPoint entryPoint,
```
rwinch: Remove this method. It is only used once and the value of preferredMatcher is `AnyRequestMatcher`. If users need an `AuthenticationEntryPoint` that can delegate by `RequestMatcher`, then they can inject `DelegatingAuthenticationEntryPoint`. 

This should also allow changing the `entryPoints` member variable to be of type `Map<String, AuthenticationEntryPoint>`

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
```diff
@@ -75,6 +93,8 @@ public final class ExceptionHandlingConfigurer<H extends HttpSecurityBuilder<H>>
 
 	private LinkedHashMap<RequestMatcher, AccessDeniedHandler> defaultDeniedHandlerMappings = new LinkedHashMap<>();
 
+	private Map<String, LinkedHashMap<RequestMatcher, AuthenticationEntryPoint>> entryPoints = new LinkedHashMap<>();
```
rwinch: Try renaming this to something that helps describe what it does. Something like `missingAuthorityToEntryPoint` makes it clear what the key is.

NOTE: They type can be simplified to `Map<String, AuthenticationEntryPoint>` as described in [this comment](https://github.com/spring-projects/spring-security/pull/17775/files#diff-ca8dd458d8499b62e481ddd01feb3bbf5d735cb3decb6ee38ea649b076bbb3cfR188).

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ott/OneTimeTokenLoginConfigurer.java
```diff
@@ -391,4 +410,52 @@ public ApplicationContext getContext() {
 		return this.context;
 	}
 
+	private static final class PostAuthenticationEntryPoint implements AuthenticationEntryPoint {
```
rwinch: This introduces a CSRF attack that could lead to leaking the OTT through social engineering. Consider a user who is authenticated, but who has not performed MFA with OTT. They then go to evil.example.com which performs a request to a resource that requires OTT. This will automatically submit a token to the user which evil.example.com could say we sent you a OTT please paste it in this form on evil.example.com

What's more is there is no controls in place for how many OTT are created and submitted. I think that we want a confirmation page that looks much like the login page, but does not allow the username to be changed.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
```diff
@@ -165,6 +185,26 @@ public ExceptionHandlingConfigurer<H> defaultAuthenticationEntryPointFor(Authent
...
+		LinkedHashMap<RequestMatcher, AuthenticationEntryPoint> byMatcher = this.entryPoints.get(authority);
+		if (byMatcher == null) {
+			byMatcher = new LinkedHashMap<>();
+		}
+		byMatcher.put(preferredMatcher, entryPoint);
+		this.entryPoints.put(authority, byMatcher);
+		return this;
+	}
+
+	public ExceptionHandlingConfigurer<H> defaultAuthenticationEntryPointFor(AuthenticationEntryPoint entryPoint,
```
rwinch: The method name for this is not very clear:

- It isn't clear that the `AuthenticationEntryPoint` is used for `AccessDeniedException`
- It isn't clear that the authority is a missing authority. We can change the param but String is a common type and this might collide with other methods in the future.

I wonder if we can expose a builder object (e.g. `DelegatingMissingAuthorityAccessDeniedHandler.Builder`) that would help make this more clear.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
```diff
@@ -235,15 +295,28 @@ private AccessDeniedHandler createDefaultDeniedHandler(H http) {
...
+		}
+		Map<String, AuthenticationEntryPoint> entryPoints = new LinkedHashMap<>();
+		for (Map.Entry<String, LinkedHashMap<RequestMatcher, AuthenticationEntryPoint>> entry : this.entryPoints
+			.entrySet()) {
+			entryPoints.put(entry.getKey(), entryPointFrom(entry.getValue()));
+		}
+		return new AuthenticationFactorDelegatingAuthenticationEntryPoint(entryPoints, defaults);
+	}
+
+	private AuthenticationEntryPoint entryPointFrom(
```
rwinch: If we remove `AuthenticationFactorDelegatingAuthenticationEntryPoint` then we can remove this method

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
```diff
@@ -263,4 +336,128 @@ private RequestCache getRequestCache(H http) {
...
+			}
+			if (!(denied.getAuthorizationResult() instanceof AuthorityAuthorizationDecision authorization)) {
+				return List.of();
+			}
+			return authorization.getAuthorities();
+		}
+
+	}
+
+	private static final class AuthenticationEntryPointAccessDeniedHandlerAdapter implements AccessDeniedHandler {
```
rwinch: This can be removed in favor of `AuthenticationFactorDelegatingAccessDeniedHandler` accepting `AuthenticationFactorDelegatingAccessDeniedHandler(Map<String,AuthenticationEntryPoint> missingAuthorityToEntryPoint, AccessDeniedHandler defaultHandler)`. If an `AuthenticationEntryPoint` is used, then it will use a `RequestCache` before invoking `commence`.

By removing this we have a lot fewer wrapper objects and less references to `RequestCache` to keep track of.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
```diff
@@ -263,4 +336,128 @@ private RequestCache getRequestCache(H http) {
...
+		public void handle(HttpServletRequest request, HttpServletResponse response, AccessDeniedException denied)
+				throws IOException, ServletException {
+			AuthenticationException ex = new InsufficientAuthenticationException("access denied", denied);
+			this.requestCache.saveRequest(request, response);
+			this.entryPoint.commence(request, response, ex);
+		}
+
+	}
+
+	private static final class AuthenticationFactorDelegatingAccessDeniedHandler implements AccessDeniedHandler {
```
rwinch: Rename this to something that is not related to factors. For example, I think that something like  `DelegatingMissingAuthorityAccessDeniedHandler` would make sense.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
```diff
@@ -263,4 +336,128 @@ private RequestCache getRequestCache(H http) {
...
+
+		private final ThrowableAnalyzer throwableAnalyzer = new ThrowableAnalyzer();
```
rwinch: `ThrowableAnalyzer` is performance intensive. I think that this can be removed because the `ExceptionTranslationFilter` is already finding the `AccessDeniedException`. Are there cases where it isn't an `AuthorizationDeniedException`? If so, is that something that we can fix?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
```diff
@@ -263,4 +336,128 @@ private RequestCache getRequestCache(H http) {
...
+				AccessDeniedHandler deniedHandler = this.deniedHandlers.get(needed.getAuthority());
+				if (deniedHandler != null) {
+					return deniedHandler;
+				}
+			}
+			return this.defaults;
+		}
+
+		private Collection<GrantedAuthority> authorizationRequest(Exception ex) {
+			Throwable[] chain = this.throwableAnalyzer.determineCauseChain(ex);
```
rwinch: This method is costly. If we keep the `ThrowableAnalyzer` logic in this class, we should do an instanceof check on the `ex` argument first. Perhaps we should have a way for this to be built into `ThrowableAnalyzer` if it isn't there already.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
rwinch: This should be extracted into a class in spring-security-web. I believe it makes sense as a sibling of `AccessDeniedHandler`, but change for tangles due to authentication APIs in here (access historically has had `AuthenticationManager` in it though so I think this works).

core/src/main/java/org/springframework/security/authentication/dao/AbstractUserDetailsAuthenticationProvider.java
```diff
@@ -197,8 +200,12 @@ protected Authentication createSuccessAuthentication(Object principal, Authentic
 		// so subsequent attempts are successful even with encoded passwords.
 		// Also ensure we return the original getDetails(), so that future
 		// authentication events after cache expiry contain the details
-		UsernamePasswordAuthenticationToken result = UsernamePasswordAuthenticationToken.authenticated(principal,
-				authentication.getCredentials(), this.authoritiesMapper.mapAuthorities(user.getAuthorities()));
+		UsernamePasswordAuthenticationToken result = UsernamePasswordAuthenticationToken
+			.authenticated(principal, authentication.getCredentials(),
+					this.authoritiesMapper.mapAuthorities(user.getAuthorities()))
+			.toBuilder()
```
rwinch: Can we avoid creating  two`UsernamePasswordAuthentiationToken` instances?

ldap/src/main/java/org/springframework/security/ldap/authentication/AbstractLdapAuthenticationProvider.java
```diff
@@ -100,8 +103,11 @@ protected Authentication createSuccessfulAuthentication(UsernamePasswordAuthenti
 			UserDetails user) {
 		Object password = this.useAuthenticationRequestCredentials ? authentication.getCredentials()
 				: user.getPassword();
-		UsernamePasswordAuthenticationToken result = UsernamePasswordAuthenticationToken.authenticated(user, password,
-				this.authoritiesMapper.mapAuthorities(user.getAuthorities()));
+		UsernamePasswordAuthenticationToken result = UsernamePasswordAuthenticationToken
+			.authenticated(user, password, this.authoritiesMapper.mapAuthorities(user.getAuthorities()))
+			.toBuilder()
```
rwinch: Can we avoid creating two instances of `UsernamePasswordAuthenticationToken`?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
```diff
@@ -263,4 +336,128 @@ private RequestCache getRequestCache(H http) {
...
+			for (GrantedAuthority needed : authorities) {
+				AccessDeniedHandler deniedHandler = this.deniedHandlers.get(needed.getAuthority());
+				if (deniedHandler != null) {
+					return deniedHandler;
+				}
+			}
+			return this.defaults;
+		}
+
+		private Collection<GrantedAuthority> authorizationRequest(Exception ex) {
```
rwinch: rename to something like `missingAuthorities`

web/src/main/java/org/springframework/security/web/access/DelegatingMissingAuthorityAccessDeniedHandler.java
```diff
@@ -0,0 +1,179 @@
...
+
+			private final String[] authorities;
+
+			private RequestMatcher requestMatcher = AnyRequestMatcher.INSTANCE;
+
+			private EntryPoint(String... authorities) {
+				this.authorities = authorities;
+			}
+
+			public EntryPoint whenRequestMatches(RequestMatcher requestMatcher) {
```
rwinch: Having `when` makes this method name feel mismatched with `authorities` and `commence`.

If we stick with having `when`, then I'd expect usage to be something like:

```
handler.whenMissingAuthorities("FACTOR_AUTHORIZATION_CODE")
				.whenRequestMatches(getAuthenticationEntryPointMatcher(http))
				.willCommenceWith(loginEntryPoint)
```

web/src/main/java/org/springframework/security/web/access/DelegatingMissingAuthorityAccessDeniedHandler.java
```diff
@@ -0,0 +1,179 @@
...
+				this.requestMatcher = requestMatcher;
+				return this;
+			}
+
+			public Builder commence(AuthenticationEntryPoint entryPoint) {
```
rwinch: When I see this method I read that it is commencing the `AuthenticationEntryPoint` rather than going to use the `AuthenticationEntryPoint`.

web/src/main/java/org/springframework/security/web/access/DelegatingMissingAuthorityAccessDeniedHandler.java
```diff
@@ -0,0 +1,179 @@
...
+				String... authorities) {
+			for (String authority : authorities) {
+				Map<RequestMatcher, AuthenticationEntryPoint> requestMatcherByAuthority = this.entryPointByRequestMatcherByAuthority
+					.computeIfAbsent(authority, (k) -> new LinkedHashMap<>());
+				requestMatcherByAuthority.put(requestMatcher, entryPoint);
+			}
+			return this;
+		}
+
+		public EntryPoint authorities(String... authorities) {
```
rwinch: I think that method name should reflect what type of authorities they are (e.g. `missingAuthorities`)

rwinch: When will we need a vargs? Can it be simplified to a single missing authority?

web/src/main/java/org/springframework/security/web/access/DelegatingMissingAuthorityAccessDeniedHandler.java
```diff
@@ -0,0 +1,179 @@
...
+			return new EntryPoint(authorities);
+		}
+
+		public DelegatingMissingAuthorityAccessDeniedHandler build() {
+			Map<String, AuthenticationEntryPoint> entryPointByAuthority = new LinkedHashMap<>();
```
rwinch, line 202: Rather that duplicating the logic for creating a `DelegatingAuthenticationEntryPoint`, it might be better to use the new `DelegatingAuthenticationEntryPoint.Builder` API

web/src/main/java/org/springframework/security/web/access/DelegatingMissingAuthorityAccessDeniedHandler.java
```diff
@@ -0,0 +1,179 @@
...
+					DelegatingAuthenticationEntryPoint entryPoint = new DelegatingAuthenticationEntryPoint(
+							new LinkedHashMap<>(entryPointByRequestMatcher));
+					entryPoint.setDefaultEntryPoint(defaultEntryPoint);
+					entryPointByAuthority.put(authority, entryPoint);
+				}
+			}
+			return new DelegatingMissingAuthorityAccessDeniedHandler(entryPointByAuthority);
+		}
+
+		public final class EntryPoint {
```
rwinch: This name is confusing because it is not an `EntryPoint` from the perspective of `AuthenticationEntryPoint` which is a well established API.

web/src/main/java/org/springframework/security/web/access/DelegatingMissingAuthorityAccessDeniedHandler.java
```diff
@@ -0,0 +1,179 @@
...
+
+	public static final class Builder {
+
+		private final Map<String, Map<RequestMatcher, AuthenticationEntryPoint>> entryPointByRequestMatcherByAuthority = new LinkedHashMap<>();
+
+		private Builder() {
+
+		}
+
+		Builder entryPointFor(RequestMatcher requestMatcher, AuthenticationEntryPoint entryPoint,
```
rwinch: While not a public API, I wonder if this could take advantage of [the new DelegatingAuthenticationEntryPoint.Builder API](https://github.com/spring-projects/spring-security/issues/17915):


```java
public Builder entryPointForMissingAuthority(
                Consumer<DelegatingAuthenticationEntryPoint.Builder> entryPoint,
                String missingAuthority) {
...
}

public Builder entryPointForMissingAuthority(AuthenticationEntryPoint entryPoint, String missingAuthority) {
    return entryPointForMissingAuthority((builder) -> builder.defaultEntryPoint(entryPoint), missingAuthority);
}
```

web/src/main/java/org/springframework/security/web/authentication/preauth/PreAuthenticatedAuthenticationProvider.java
```diff
@@ -142,6 +152,14 @@ public void setUserDetailsChecker(UserDetailsChecker userDetailsChecker) {
 		this.userDetailsChecker = userDetailsChecker;
 	}
 
+	/**
+	 * Sets authorities that this provider should grant once authentication completes
+	 * @param grantedAuthoritySupplier the supplier that grants authorities
+	 */
+	public void setGrantedAuthoritySupplier(Supplier<Collection<GrantedAuthority>> grantedAuthoritySupplier) {
```
rwinch, line 159: Add `Assert.notNull`

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
```diff
@@ -127,6 +132,18 @@ public ExceptionHandlingConfigurer<H> defaultAccessDeniedHandlerFor(AccessDenied
 		return this;
 	}
 
+	public ExceptionHandlingConfigurer<H> defaultAuthenticationEntryPointFor(AuthenticationEntryPoint entryPoint,
```
rwinch: This name is still a bit confusing to me because it is configuring the access denied handler and does not clarify what the String type is. Perhaps something like `defaultDeniedHandlerForMissingAuthority`

config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
```diff
@@ -127,6 +132,18 @@ public ExceptionHandlingConfigurer<H> defaultAccessDeniedHandlerFor(AccessDenied
...
+			String authority) {
+		this.missingAuthoritiesHandlerBuilder.addEntryPointFor(entryPoint, authority);
+		return this;
+	}
+
+	public ExceptionHandlingConfigurer<H> defaultAuthenticationEntryPointFor(
```
rwinch: This name is a bit confusing to me because it is configuring the access denied handler and does not clarify what the String type is. Perhaps something like `defaultDeniedHandlerForMissingAuthority`.

NOTE We had talked about naming it based on the `Consumer<B>` type, but `B` is being passed into another builder which is building an `AuthorizationDeniedHandler`.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/X509Configurer.java
```diff
@@ -177,8 +179,16 @@ public X509Configurer<H> subjectPrincipalRegex(String subjectPrincipalRegex) {
...
 		authenticationProvider.setPreAuthenticatedUserDetailsService(getAuthenticationUserDetailsService(http));
+		authenticationProvider.setGrantedAuthoritySupplier(() -> AuthorityUtils.createAuthorityList("FACTOR_X509"));
 		http.authenticationProvider(authenticationProvider)
 			.setSharedObject(AuthenticationEntryPoint.class, new Http403ForbiddenEntryPoint());
+		ExceptionHandlingConfigurer<H> exceptions = http.getConfigurer(ExceptionHandlingConfigurer.class);
+		if (exceptions != null) {
+			AuthenticationEntryPoint forbidden = new Http403ForbiddenEntryPoint();
+			exceptions.defaultAuthenticationEntryPointFor(forbidden, AnyRequestMatcher.INSTANCE);
+			exceptions.defaultAuthenticationEntryPointFor(
+					(ep) -> ep.addEntryPointFor(forbidden, AnyRequestMatcher.INSTANCE), "FACTOR_X509");
```
rwinch:
```suggestion
			exceptions.defaultAuthenticationEntryPointFor(
					(ep) -> ep.defaultEntryPoint(forbidden), "FACTOR_X509");
```

This seems a bit more concise and does the same thing.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/X509Configurer.java
```diff
@@ -177,8 +179,16 @@ public X509Configurer<H> subjectPrincipalRegex(String subjectPrincipalRegex) {
 	public void init(H http) {
 		PreAuthenticatedAuthenticationProvider authenticationProvider = new PreAuthenticatedAuthenticationProvider();
 		authenticationProvider.setPreAuthenticatedUserDetailsService(getAuthenticationUserDetailsService(http));
+		authenticationProvider.setGrantedAuthoritySupplier(() -> AuthorityUtils.createAuthorityList("FACTOR_X509"));
 		http.authenticationProvider(authenticationProvider)
 			.setSharedObject(AuthenticationEntryPoint.class, new Http403ForbiddenEntryPoint());
+		ExceptionHandlingConfigurer<H> exceptions = http.getConfigurer(ExceptionHandlingConfigurer.class);
+		if (exceptions != null) {
+			AuthenticationEntryPoint forbidden = new Http403ForbiddenEntryPoint();
+			exceptions.defaultAuthenticationEntryPointFor(forbidden, AnyRequestMatcher.INSTANCE);
```
rwinch, line 188: I'm not sure about this change.

If you have X509 invoked first then the default `AuthenticationEntryPoint` is forbidden for everything. What happens if X509 and password are required and a user is authenticated with X509 but not password? I'd expect that the result would be prompt for a password but it appears that this will result in forbidden.

core/src/main/java/org/springframework/security/core/GrantedAuthority.java
```diff
@@ -31,6 +31,8 @@
  */
 public interface GrantedAuthority extends Serializable {
 
+	String MISSING_AUTHORITIES_ATTRIBUTE = GrantedAuthority.class + ".missingAuthorities";
```
rwinch: What is this intended to be used for? Thus far it seems that it is the `HttpServletRequest` attribute name. If it is only used for web (I think that is the only place we will use it), I believe this makes more sense in the web jar / package, perhaps in `WebAttributes`.

web/src/main/java/org/springframework/security/web/access/DelegatingMissingAuthorityAccessDeniedHandler.java
```diff
@@ -0,0 +1,217 @@
...
+		}
+		this.defaultAccessDeniedHandler.handle(request, response, denied);
+	}
+
+	/**
+	 * Use this {@link AccessDeniedHandler} for {@link AccessDeniedException}s that this
+	 * handler doesn't support. By default, this uses {@link AccessDeniedHandlerImpl}.
+	 * @param defaultAccessDeniedHandler the default {@link AccessDeniedHandler} to use
+	 */
+	public void setDefaultAccessDeniedHandler(AccessDeniedHandler defaultAccessDeniedHandler) {
```
rwinch, line 117: Add `Assert.notNull`

web/src/main/java/org/springframework/security/web/access/DelegatingMissingAuthorityAccessDeniedHandler.java
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	/**
+	 * Use this {@link RequestCache} to remember the current request.
+	 * <p>
+	 * Uses {@link NullRequestCache} by default
+	 * </p>
+	 * @param requestCache the {@link RequestCache} to use
+	 */
+	public void setRequestCache(RequestCache requestCache) {
```
rwinch, line 129: Add `Assert.notNull`

web/src/main/java/org/springframework/security/web/access/DelegatingMissingAuthorityAccessDeniedHandler.java
```diff
@@ -0,0 +1,217 @@
...
+
+	private final ThrowableAnalyzer throwableAnalyzer = new ThrowableAnalyzer();
+
+	private final Map<String, AuthenticationEntryPoint> entryPoints;
+
+	private RequestCache requestCache = new NullRequestCache();
+
+	private AccessDeniedHandler defaultAccessDeniedHandler = new AccessDeniedHandlerImpl();
+
+	private DelegatingMissingAuthorityAccessDeniedHandler(Map<String, AuthenticationEntryPoint> entryPoints) {
```
rwinch, line 88: Add `Assert.notEmpty`

web/src/main/java/org/springframework/security/web/access/DelegatingMissingAuthorityAccessDeniedHandler.java
```diff
@@ -0,0 +1,217 @@
...
+				String... authorities) {
+			for (String authority : authorities) {
+				entryPoint.accept(Builder.this.entryPointBuilder(authority));
+			}
+			return this;
+		}
+
+		public DelegatingMissingAuthorityAccessDeniedHandler build() {
+			Map<String, AuthenticationEntryPoint> entryPointByAuthority = new LinkedHashMap<>();
+			for (String authority : this.entryPointByRequestMatcherByAuthority.keySet()) {
```
rwinch: Consider iterating over `entrySet()` instead so that an additional lookup of the value is not required.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/X509Configurer.java
jzheaux: If the user is authenticated with X509, but not password, then this `AuthenticationEntryPoint` is not exercised. Instead the one configured in the missing-authorities handler is.

I think this speaks to your other comments about the method names being confusing.

jzheaux: Either way, I think [it might fix a bug](https://github.com/spring-projects/spring-security/blob/main/config/src/main/java/org/springframework/security/config/annotation/web/configurers/X509Configurer.java#L184), which I'd like to confirm with you. It doesn't have to do with the PR, and I will log it separately if it needs fixing.

web/src/main/java/org/springframework/security/web/access/DelegatingMissingAuthorityAccessDeniedHandler.java
jzheaux: Sorry I missed this. I'm glad to see we came to the same conclusion.

web/src/main/java/org/springframework/security/web/access/DelegatingMissingAuthorityAccessDeniedHandler.java
jzheaux: After writing tests using the API, I settled on a version with less nesting, similar to what `DelegatingAuthenticationEntryPoint.Builder` looks like.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/X509Configurer.java
jzheaux: As far as I can tell, this causes the `DelegatingAuthenticationEntryPoint` with throw an exception since `List<RequestMatcherEntry>` is empty. That's because `.addEntryPointFor` is never called for this builder elsewhere in the code.

web/src/main/java/org/springframework/security/web/access/DelegatingMissingAuthorityAccessDeniedHandler.java
jzheaux: I'm happy to wait and see how people use the feature; I'm imagining folks doing things like:

```java
.defaultDeniedHandlerForMissingAuthority(myEntryPoint, "my:authority", "my:otherauthority")
```

(I agree that it would be a little odd to wire two separate factors to the same entry point.)

It's simple to add varargs in the event folks need this. Or they can obviously just call the method twice themselves, so I'll remove it for now.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/X509Configurer.java
rwinch: Sorry I did describe my concern incorrectly.

Consider the following:

```java
@Bean
SecurityFilterChain securityFilterChain(HttpSecurity http) throws Exception {
	// @formatter:off
	http
		.authorizeHttpRequests((requests) -> requests
			.anyRequest().authenticated()
		)
		.x509(withDefaults())
		.formLogin(withDefaults());
	// @formatter:on
	return http.build();
}
```

If the user is not authenticated and makes a request that accepts `text/html` what would you expect the response to be? I'd expect to be redirected to the log in page since authenticating with username/password would fulfill the requirement. Without the line above, a redirect to the log in page does happen. However, with the change above a 403 is returned.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/WebAuthnConfigurer.java
```diff
@@ -150,6 +153,16 @@ public WebAuthnConfigurer<H> creationOptionsRepository(
 		return this;
 	}
 
+	@Override
+	public void init(H http) throws Exception {
+		ExceptionHandlingConfigurer<H> exceptions = http.getConfigurer(ExceptionHandlingConfigurer.class);
+		if (exceptions != null) {
+			AuthenticationEntryPoint entryPoint = new LoginUrlAuthenticationEntryPoint("/login");
+			exceptions.defaultDeniedHandlerForMissingAuthority(
+					(ep) -> ep.addEntryPointFor(entryPoint, AnyRequestMatcher.INSTANCE), "FACTOR_WEBAUTHN");
```
rwinch, line 162: Minor, but I think that this can be simplified

```suggestion
					(ep) -> ep.defaultEntryPoint(entryPoint), "FACTOR_WEBAUTHN");
```

docs/modules/ROOT/pages/servlet/authentication/adaptive.adoc
```diff
@@ -0,0 +1,528 @@
...
+Java::
++
+[source,java,role="primary"]
+----
+@Bean
+public SecurityFilterChain securityFilterChain(HttpSecurity http) throws Exception {
+    // @formatter:off
+    http
+        .authorizeHttpRequests((authorize) -> authorize
+            .anyRequest().access(allOf(hasAuthority("FACTOR_PASSWORD"), hasAuthority("FACTOR_OTT"))) <1>
```
rwinch, line 127:
```suggestion
            .anyRequest().hasAllAuthorities("FACTOR_PASSWORD", "FACTOR_OTT") <1>
```

This can now use https://github.com/spring-projects/spring-security/issues/17932

docs/modules/ROOT/pages/servlet/authentication/adaptive.adoc
```diff
@@ -0,0 +1,528 @@
...
+Then, if later on a scope that they did not grant is needed, consent can be re-requested for just that scope.
+4. *Opting-in to Stronger Authentication Mechanisms* - Users may not be ready yet to start using MFA, but the application wants to allow the subset of security-minded users to opt-in.
+5. *Requiring Additional Steps for Suspicious Logins* - The application may notice that the user's IP address has changed, that they are behind a VPN, or some other consideration that requires additional verification
+
+== Re-authentication
+
+The most common of these is re-authentication.
+Imagine an application configured in the following way:
+
+[tabs]
```
rwinch, line 19: Please use include-code the code is compiled. Ideally, there is a test for each sample as well. See https://github.com/spring-projects/spring-security/issues/16226

web/src/main/java/org/springframework/security/web/WebAttributes.java
```diff
@@ -52,6 +53,15 @@ public final class WebAttributes {
 	public static final String WEB_INVOCATION_PRIVILEGE_EVALUATOR_ATTRIBUTE = WebAttributes.class.getName()
 			+ ".WEB_INVOCATION_PRIVILEGE_EVALUATOR_ATTRIBUTE";
 
+	/**
+	 * Used to specify to the view layer what missing authorities caused an
+	 * {@link AuthorizationDeniedException}
```
rwinch, line 58: I'd prefer that this not talk about where it came from what what the value is (the type) and what it is used for. AuthorizationDeniedException is likely only one way to come across the required authorities.

rwinch (changes requested):
I think we are very close now. I've provided feedback inline. In addition, can you please log a ticket to provide static constants for common factors?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/WebAuthnConfigurer.java
jzheaux: I might be misunderstanding. Is this similar to the recommendation for `FACTOR_X509`? If so, please see https://github.com/spring-projects/spring-security/pull/17775#discussion_r2364698452 and if there is a way I can look at this differently.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/X509Configurer.java
jzheaux: Added https://github.com/spring-projects/spring-security/issues/17955 to track this

jzheaux:
Merged in 28aad8855ce8931bda20348ef80ce7a06befbba4

PR #17790, opened 2025-08-21
Authentication Builder

jzheaux:
Applications can use `AuthenticationBuilder` to apply existing authentications to new ones.

For example, if the current logged in user is represented by:

```java
Authentication firstFactor = ...
```

And they provide a second set of authenticated credentials, represented by:

```java
Authentication secondFactor = ...
```

Then the first factor can be applied to the second factor as follows:

```java
secondFactor = secondFactor.toBuilder().apply(firstFactor).build();
```

This draft PR adds a basic builder to each Authentication implementation that implements `AuthenticationBuilder`. In order to simplify upgrades, `toBuilder` by default returns a no-op implementation of `AuthenticationBuilder` that ultimately returns the same authentication unchanged.


core/src/main/java/org/springframework/security/authentication/AbstractAuthenticationToken.java
```diff
@@ -185,4 +187,36 @@ public String toString() {
...
+		}
+
+		@Override
+		public A build() {
+			return build(this.authorities);
+		}
+
+		@Override
+		public B apply(Authentication token) {
+			Assert.isTrue(token.isAuthenticated(), "cannot mutate an unauthenticated token");
```
rwinch: Why can't an unauthenticated user be mutated? I believe that the builders should be independent of the MFA feature. If we keep this assertion, it would be good to include the why in the error message.

core/src/main/java/org/springframework/security/authentication/AbstractAuthenticationToken.java
```diff
@@ -185,4 +187,36 @@ public String toString() {
...
+			Assert.notNull(token.getPrincipal(), "principal cannot be null");
```
rwinch: Why can't a principal be null? If we keep this assertion, it would be good to include the why in the error message.

core/src/main/java/org/springframework/security/authentication/AbstractAuthenticationToken.java
```diff
@@ -185,4 +187,36 @@ public String toString() {
 		return sb.toString();
 	}
 
+	protected abstract static class AbstractAuthenticationBuilder<A extends Authentication, B extends AbstractAuthenticationBuilder<A, B>>
```
rwinch: I'd expect that this builder's properties aligns with the properties available on `AbstractAuthentication`.

core/src/main/java/org/springframework/security/authentication/AbstractAuthenticationToken.java
```diff
@@ -185,4 +187,36 @@ public String toString() {
...
+			return (B) this;
+		}
+
+		@Override
+		public A build() {
+			return build(this.authorities);
+		}
+
+		@Override
+		public B apply(Authentication token) {
```
rwinch: Related to the comment about only having an `authorities` member, it is surprising that an `Authentication` is passed in here, but only the `authorities` are used from it. There are quite a few other properties that I think should be taken into account.

core/src/main/java/org/springframework/security/authentication/AbstractAuthenticationToken.java
```diff
@@ -185,4 +187,36 @@ public String toString() {
...
+			Assert.isTrue(token.isAuthenticated(), "cannot mutate an unauthenticated token");
+			Assert.notNull(token.getPrincipal(), "principal cannot be null");
+			this.authorities.addAll(token.getAuthorities());
+			return (B) this;
+		}
+
+		protected abstract A build(Collection<GrantedAuthority> authorities);
```
rwinch: I think that I'd prefer to make authorities protected and not pass anything into this method. This gives us flexibility of adding additional properties later.

core/src/main/java/org/springframework/security/authentication/ProviderManager.java
```diff
@@ -265,6 +271,20 @@ public Authentication authenticate(Authentication authentication) throws Authent
 		throw lastException;
 	}
 
+	private @Nullable Authentication applyPreviousAuthentication(@Nullable Authentication result) {
```
rwinch: I think that we should move this to the controller logic (e.g. the Filter).

The concern that I have with the way it is now is that the `ProviderManager` knows about the `SecurityContextHolderStrategy` to obtain the current `Authentication`, but it is only reading it and not writing it (which is a bit of a mismatch). We also don't want to move the logic into `AuthenticationManager` because we don't want to tie `AuthenticationManager` to the context.

core/src/main/java/org/springframework/security/core/Authentication.java
```diff
@@ -64,7 +70,7 @@ public interface Authentication extends Principal, Serializable {
 	 * instance.
 	 * </p>
 	 * @return the authorities granted to the principal, or an empty collection if the
-	 * token has not been authenticated. Never null.
+	 * token has not been authenticated. Never null.Saml2AssertAu
```
rwinch: This looks like it was accidentally added to the commit.

core/src/main/java/org/springframework/security/core/Authentication.java
```diff
@@ -136,4 +142,50 @@ public interface Authentication extends Principal, Serializable {
...
 	void setAuthenticated(boolean isAuthenticated) throws IllegalArgumentException;
 
+	/**
+	 * Return an {@link Builder} based on this instance
+	 * @return an {@link Builder} for building a new {@link Authentication} based on this
+	 * instance
+	 * @since 7.0
+	 */
+	default Builder<?, ?> toBuilder() {
+		return new NoopAuthenticationBuilder<>(this);
```
rwinch: It seems like returning a Noop implementation is a bit fragile. Instead, I think that we should be returning a builder that can perform the operations on the Builder interface.

core/src/main/java/org/springframework/security/core/Authentication.java
```diff
@@ -136,4 +142,50 @@ public interface Authentication extends Principal, Serializable {
...
+		/**
+		 * Apply this {@link Authentication} to the builder.
+		 * <p>
+		 * By default, this method adds the authorities from {@code authentication} to
+		 * this builder
+		 * @return the {@link Builder} for further configuration
+		 */
+		default B apply(Authentication authentication) {
+			Assert.isTrue(authentication.isAuthenticated(), "cannot apply an unauthenticated token");
+			return authorities((a) -> a.addAll(authentication.getAuthorities()));
```
rwinch: Why does the default method ignore other properties?

core/src/main/java/org/springframework/security/core/Authentication.java
```diff
@@ -136,4 +142,50 @@ public interface Authentication extends Principal, Serializable {
...
+
+		/**
+		 * Apply this {@link Authentication} to the builder.
+		 * <p>
+		 * By default, this method adds the authorities from {@code authentication} to
+		 * this builder
+		 * @return the {@link Builder} for further configuration
+		 */
+		default B apply(Authentication authentication) {
+			Assert.isTrue(authentication.isAuthenticated(), "cannot apply an unauthenticated token");
```
rwinch: Why doesn't this allow building when Authentiation.isAuthenticated() is false?

core/src/main/java/org/springframework/security/core/Authentication.java
```diff
@@ -136,4 +142,50 @@ public interface Authentication extends Principal, Serializable {
...
+	}
+
+	/**
+	 * A builder based on a given {@link Authentication} instance
+	 *
+	 * @param <A> the type of {@link Authentication}
+	 * @author Josh Cummings
+	 * @since 7.0
+	 */
+	interface Builder<A extends Authentication, B extends Builder<A, B>> {
```
rwinch: It seems like this should have methods for every property on `Authentication`. This would allow us to also deprecate the `setAuthenticated` method.

core/src/main/java/org/springframework/security/core/Authentication.java
```diff
@@ -136,4 +142,50 @@ public interface Authentication extends Principal, Serializable {
...
+
+		/**
+		 * Apply this {@link Authentication} to the builder.
+		 * <p>
+		 * By default, this method adds the authorities from {@code authentication} to
+		 * this builder
+		 * @return the {@link Builder} for further configuration
+		 */
+		default B apply(Authentication authentication) {
```
rwinch: This method seems to be used to initialize the Builder with `this` (the current `Authentication`). I think that it might be better to use the constructor instead.

rwinch: The docs say Apply `this`, if that is the case should it be `apply(A authentication)`

cas/src/main/java/org/springframework/security/cas/authentication/CasAuthenticationToken.java
```diff
@@ -162,4 +169,66 @@ public String toString() {
 		return (sb.toString());
 	}
 
+	/**
+	 * A builder preserving the concrete {@link Authentication} type
+	 *
+	 * @since 7.0
+	 */
+	public static final class Builder extends AbstractAuthenticationBuilder<@NonNull CasAuthenticationToken, Builder> {
```
rwinch: There is no need to explicitly declare `@NonNull` as this is the default

cas/src/main/java/org/springframework/security/cas/authentication/CasAuthenticationToken.java
```diff
@@ -162,4 +169,66 @@ public String toString() {
...
+			return this;
+		}
+
+		public Builder assertion(Assertion assertion) {
+			this.assertion = assertion;
+			return this;
+		}
+
+		@Override
+		protected @NonNull CasAuthenticationToken build(Collection<GrantedAuthority> authorities) {
```
rwinch: There is no need to explicitly declare `@NonNull` as this is the default

core/src/main/java/org/springframework/security/authentication/DelegatingReactiveAuthenticationManager.java
```diff
@@ -57,11 +58,24 @@ public DelegatingReactiveAuthenticationManager(List<ReactiveAuthenticationManage
...
+		return ReactiveSecurityContextHolder.getContext().flatMap((context) -> {
+			Mono<Authentication> result = doAuthenticate(authentication);
+			Authentication current = context.getAuthentication();
+			if (current == null) {
+				return result;
+			}
+			if (!current.isAuthenticated()) {
+				return result;
+			}
+			return doAuthenticate(current).map((r) -> r.toBuilder().apply(current).build());
```
rwinch: Just as in the ProviderManager, the combining of these should not be done in the `ReactiveAuthenticationManager`.

https://github.com/spring-projects/spring-security/pull/17790/files#diff-84d496ceee0087c25a0972331a1e127cc87ca0310c854c76b43e083304e83cb2R274

core/src/main/java/org/springframework/security/core/Authentication.java
```diff
@@ -54,6 +57,9 @@
  */
 public interface Authentication extends Principal, Serializable {
 
+	@Serial
+	long serialVersionUID = -3884394378624019849L;
```
rwinch: Can you help me to understand why this is specified?

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/provider/service/authentication/Saml2AssertionAuthentication.java
```diff
@@ -62,4 +65,56 @@ public String getRelyingPartyRegistrationId() {
...
+		return new Builder().apply(this);
+	}
+
+	/**
+	 * A builder preserving the concrete {@link Authentication} type
+	 *
+	 * @since 7.0
+	 */
+	public static final class Builder
+			extends AbstractAuthenticationBuilder<@NonNull Saml2AssertionAuthentication, @NonNull Builder> {
```
rwinch: There is no need to explicitly declare `@NonNull` as this is the default

core/src/main/java/org/springframework/security/core/Authentication.java
jzheaux: Deserialization of concrete implementations fails otherwise. With the inclusion of a `default` method, Java now treats it like an `abstract` class from a serialization perspective. From what I understand, any class, abstract or otherwise, should have a version id if it extends `Serializable`.

config/src/test/resources/serialized/6.4.x/org.springframework.security.authentication.event.InteractiveAuthenticationSuccessEvent.serialized
rwinch, line 1: We cannot update `*.serialized` files from previous generations. The reason it exists is to validate passivity, but we've made a breaking change.

core/src/main/java/org/springframework/security/core/Authentication.java
```diff
@@ -136,4 +137,43 @@ public interface Authentication extends Principal, Serializable {
...
+		return new SimpleAuthentication.Builder(this);
+	}
+
+	/**
+	 * A builder based on a given {@link Authentication} instance
+	 *
+	 * @author Josh Cummings
+	 * @since 7.0
+	 */
+	interface Builder<P, C, B extends Builder<P, C, B>> {
```
rwinch: Please add missing javadoc for the builder.

core/src/main/java/org/springframework/security/core/Authentication.java
```diff
@@ -136,4 +137,43 @@ public interface Authentication extends Principal, Serializable {
...
+		return new SimpleAuthentication.Builder(this);
+	}
+
+	/**
+	 * A builder based on a given {@link Authentication} instance
+	 *
+	 * @author Josh Cummings
+	 * @since 7.0
+	 */
+	interface Builder<B extends Builder<B>> {
```
rwinch, line 168: Please add missing javadoc

PR #17801, opened 2025-08-24
Cleanup

ngocnhan-tran1996:
This PR includes
- Use `String#isEmpty`
- Remove redundant check and exception
- Update javadoc
- Polish some methods

### DIFF
--- a/cas/src/main/java/org/springframework/security/cas/web/authentication/DefaultServiceAuthenticationDetails.java
+++ b/cas/src/main/java/org/springframework/security/cas/web/authentication/DefaultServiceAuthenticationDetails.java
@@ -70,14 +70,13 @@ final class DefaultServiceAuthenticationDetails extends WebAuthenticationDetails
 
 	@Override
 	public boolean equals(Object obj) {
-		if (this == obj) {
+		if (super.equals(obj)) {
 			return true;
 		}
-		if (!super.equals(obj) || !(obj instanceof DefaultServiceAuthenticationDetails)) {
-			return false;
+		if (obj instanceof DefaultServiceAuthenticationDetails that) {
+			return this.serviceUrl.equals(that.getServiceUrl());
 		}
-		ServiceAuthenticationDetails that = (ServiceAuthenticationDetails) obj;
-		return this.serviceUrl.equals(that.getServiceUrl());
+		return false;
 	}
 
 	@Override
--- a/cas/src/main/java/org/springframework/security/cas/web/authentication/DefaultServiceAuthenticationDetails.java
+++ b/cas/src/main/java/org/springframework/security/cas/web/authentication/DefaultServiceAuthenticationDetails.java
@@ -110,7 +110,7 @@ final class DefaultServiceAuthenticationDetails extends WebAuthenticationDetails
 			return null;
 		}
 		String result = artifactPattern.matcher(query).replaceFirst("");
-		if (result.length() == 0) {
+		if (result.isEmpty()) {
 			return null;
 		}
 		// strip off the trailing & only if the artifact was the first query param
--- a/cas/src/main/java/org/springframework/security/cas/web/authentication/DefaultServiceAuthenticationDetails.java
+++ b/cas/src/main/java/org/springframework/security/cas/web/authentication/DefaultServiceAuthenticationDetails.java
@@ -100,7 +100,11 @@ final class DefaultServiceAuthenticationDetails extends WebAuthenticationDetails
 	/**
 	 * If present, removes the artifactParameterName and the corresponding value from the
 	 * query String.
-	 * @param request
+	 * @param request the current {@link HttpServletRequest} to obtain the
+	 * {@link #getServiceUrl()} from.
+	 * @param artifactPattern the {@link Pattern} that will be used to clean up the query
+	 * string from containing the artifact name and value. This can be created using
+	 * {@link #createArtifactPattern(String)}.
 	 * @return the query String minus the artifactParameterName and the corresponding
 	 * value.
 	 */
@@ -121,8 +125,10 @@ final class DefaultServiceAuthenticationDetails extends WebAuthenticationDetails
 	 * Creates a {@link Pattern} that can be passed into the constructor. This allows the
 	 * {@link Pattern} to be reused for every instance of
 	 * {@link DefaultServiceAuthenticationDetails}.
-	 * @param artifactParameterName
-	 * @return
+	 * @param artifactParameterName the artifactParameterName that is removed from the
+	 * current URL. The result becomes the service url. Cannot be null and cannot be an
+	 * empty String.
+	 * @return a {@link Pattern}
 	 */
 	static Pattern createArtifactPattern(String artifactParameterName) {
 		Assert.hasLength(artifactParameterName, "artifactParameterName is expected to have a length");
--- a/cas/src/main/java/org/springframework/security/cas/web/authentication/DefaultServiceAuthenticationDetails.java
+++ b/cas/src/main/java/org/springframework/security/cas/web/authentication/DefaultServiceAuthenticationDetails.java
@@ -33,6 +33,7 @@ import org.springframework.util.Assert;
  * and using the current URL minus the artifact and the corresponding value.
  *
  * @author Rob Winch
+ * @author Ngoc Nhan
  */
 final class DefaultServiceAuthenticationDetails extends WebAuthenticationDetails
 		implements ServiceAuthenticationDetails {
--- a/core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java
+++ b/core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java
@@ -167,7 +167,8 @@ public abstract class SecurityExpressionRoot implements SecurityExpressionOperat
 	/**
 	 * Convenience method to access {@link Authentication#getPrincipal()} from
 	 * {@link #getAuthentication()}
-	 * @return
+	 * @return the <code>Principal</code> being authenticated or the authenticated
+	 * principal after authentication.
 	 */
 	public @Nullable Object getPrincipal() {
 		return getAuthentication().getPrincipal();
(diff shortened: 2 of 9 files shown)


cas/src/main/java/org/springframework/security/cas/web/authentication/DefaultServiceAuthenticationDetails.java
```diff
@@ -110,7 +114,7 @@ private String getQueryString(final HttpServletRequest request, final Pattern ar
 			return null;
 		}
 		String result = artifactPattern.matcher(query).replaceFirst("");
-		if (result.length() == 0) {
+		if (result.isEmpty()) {
 			return null;
 		}
```
ronodhirSoumik, line 120:
```
String result = (query == null) ? "" : artifactPattern.matcher(query).replaceFirst("");

if (result.isEmpty()) {
    return null;
}
```

cas/src/main/java/org/springframework/security/cas/web/authentication/DefaultServiceAuthenticationDetails.java
```diff
@@ -121,8 +125,10 @@ private String getQueryString(final HttpServletRequest request, final Pattern ar
 	 * Creates a {@link Pattern} that can be passed into the constructor. This allows the
 	 * {@link Pattern} to be reused for every instance of
 	 * {@link DefaultServiceAuthenticationDetails}.
-	 * @param artifactParameterName
-	 * @return
+	 * @param artifactParameterName the artifactParameterName that is removed from the
+	 * current URL. The result becomes the service url. Cannot be null and cannot be an
+	 * empty String.
```
ronodhirSoumik: Cannot be null and cannot be an empty String -> Cannot be null or empty String

core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java
```diff
@@ -228,15 +230,15 @@ public void setPermissionEvaluator(PermissionEvaluator permissionEvaluator) {
...
+	 * @param defaultRolePrefix the default prefix to add to roles.
+	 * @param role the role that should be required.
+	 * @return a {@code String} role
 	 */
 	private static String getRoleWithDefaultPrefix(@Nullable String defaultRolePrefix, String role) {
 		if (role == null) {
 			return role;
 		}
-		if (defaultRolePrefix == null || defaultRolePrefix.length() == 0) {
+		if (defaultRolePrefix == null || defaultRolePrefix.isEmpty()) {
```
ronodhirSoumik: As there are several uses of `str == null || str.isEmpty()` I think a common util function will be helpful

```
boolean isEmpty(String str) {
	return str == null || str.length == 0;
}
```

ngocnhan-tran1996: I think we can use `!org.springframework.util.StringUtils#hasLength`

cas/src/main/java/org/springframework/security/cas/web/authentication/DefaultServiceAuthenticationDetails.java
```diff
@@ -70,14 +71,13 @@ public String getServiceUrl() {
 
 	@Override
 	public boolean equals(Object obj) {
-		if (this == obj) {
```
rwinch: We want to preserve this because it a performance optimization for a frequently used piece of code.

cas/src/main/java/org/springframework/security/cas/web/authentication/DefaultServiceAuthenticationDetails.java
```diff
@@ -70,14 +71,13 @@ public String getServiceUrl() {
 
 	@Override
 	public boolean equals(Object obj) {
```
rwinch: Generally, I'm not in favor of these changes as the potential benefits are a matter of opinion

cas/src/main/java/org/springframework/security/cas/web/authentication/DefaultServiceAuthenticationDetails.java
```diff
@@ -70,14 +71,13 @@ public String getServiceUrl() {
...
-		if (this == obj) {
+		if (super.equals(obj)) {
```
jzheaux: Let's please stay with `this == obj` as this is a bit more obvious what kind of shortcut we are doing.

cas/src/main/java/org/springframework/security/cas/web/authentication/DefaultServiceAuthenticationDetails.java
```diff
@@ -70,14 +71,13 @@ public String getServiceUrl() {
...
 			return true;
 		}
-		if (!super.equals(obj) || !(obj instanceof DefaultServiceAuthenticationDetails)) {
-			return false;
+		if (obj instanceof DefaultServiceAuthenticationDetails that) {
```
jzheaux: Thanks for this change. In Spring Security, it's preferred to exit early for simple cases. Would you please flip the logic here to:

```java
if (!(obj instanceof DefaultServiceAuthenticationDetails that)) {
    return false;
}
// ... continue
```

cas/src/main/java/org/springframework/security/cas/web/authentication/DefaultServiceAuthenticationDetails.java
```diff
@@ -100,17 +100,18 @@ public String toString() {
...
+	 * {@link #getServiceUrl()} from.
+	 * @param artifactPattern the {@link Pattern} that will be used to clean up the query
+	 * string from containing the artifact name and value. This can be created using
+	 * {@link #createArtifactPattern(String)}.
 	 * @return the query String minus the artifactParameterName and the corresponding
 	 * value.
 	 */
 	private String getQueryString(final HttpServletRequest request, final Pattern artifactPattern) {
 		final String query = request.getQueryString();
-		if (query == null) {
```
jzheaux, line 110: Let's please keep checking for null here since that allows us to exit early without needing to instantiate a variable.

cas/src/main/java/org/springframework/security/cas/web/authentication/DefaultServiceAuthenticationDetails.java
jzheaux: Fair point, @ronodhirSoumik, though the Spring checkstyle requires a negation in ternaries. So, this is probably the result of having run `checkstyleTest` on the PR.

core/src/main/java/org/springframework/security/access/expression/SecurityExpressionRoot.java
```diff
@@ -228,15 +231,15 @@ public void setPermissionEvaluator(PermissionEvaluator permissionEvaluator) {
 	/**
 	 * Prefixes role with defaultRolePrefix if defaultRolePrefix is non-null and if role
 	 * does not already start with defaultRolePrefix.
-	 * @param defaultRolePrefix
-	 * @param role
-	 * @return
+	 * @param defaultRolePrefix the default prefix to add to roles.
```
jzheaux: Let's please give an example here since "role prefix" could be a bit esoteric. For example:

> the default prefix to add to roles, for example {@code ROLE_}

crypto/src/main/java/org/springframework/security/crypto/password/AbstractValidatingPasswordEncoder.java
```diff
@@ -18,6 +18,14 @@
 
 import org.jspecify.annotations.Nullable;
 
+import org.springframework.util.StringUtils;
+
+/**
+ * Implementation of PasswordEncoder.
```
jzheaux: I wouldn't quite say that this is an implementation as much as it is an abstract class that allows subclasses to except the password to be non-{@code null}.

Perhaps change this to:

```java
/**
  * An abstract {@link PasswordEncoder} that implementers can use for expecting the
  * password to be non-{@code null}. Each common password API method is accompanied
  * with an abstract method with a {@code NonNull} prefix. By implementing this, the concrete
  * class is specifying what to do with the password when it is non-{@code null}, allowing this
  * class to handle the {@code null} case.
  * 
  * @author Rob Winch
  * @since 7.0
  */
```

crypto/src/main/java/org/springframework/security/crypto/password/AbstractValidatingPasswordEncoder.java
```diff
@@ -32,21 +40,20 @@ public abstract class AbstractValidatingPasswordEncoder implements PasswordEncod
...
+		return false;
 	}
 
 	protected abstract boolean matchesNonNull(String rawPassword, String encodedPassword);
 
 	@Override
 	public final boolean upgradeEncoding(@Nullable String encodedPassword) {
-		if (encodedPassword == null || encodedPassword.length() == 0) {
-			return false;
+		if (StringUtils.hasLength(encodedPassword)) {
```
jzheaux: Please favor the simpler exits coming first:

```java
if (!StringUtils.hasLength(encodedPassword)) {
    return false;
}
return upgradeEncodingNonNull(encodedPassword);
```

This also has the added benefit of making the overall change smaller.

web/src/main/java/org/springframework/security/web/firewall/RequestWrapper.java
```diff
@@ -56,7 +57,7 @@ final class RequestWrapper extends FirewalledRequest {
 		super(request);
 		this.strippedServletPath = strip(request.getServletPath());
 		String pathInfo = strip(request.getPathInfo());
-		if (pathInfo != null && pathInfo.length() == 0) {
+		if (pathInfo != null && pathInfo.isEmpty()) {
```
jzheaux: Would `!StringUtils.hasLength(pathInfo))` work here as well?

jzheaux (changes requested):
Thanks for all the great cleanup, @ngocnhan-tran1996. I've left some feedback inline.

jzheaux:
Thanks, @ngocnhan-tran1996! This is now merged into `main`.

PR #17813, opened 2025-08-29
Remove redundant code in document

quaff:


jzheaux:
Nice catch, @quaff! This is now merged into `main`.

Issue #17706, opened 2025-08-14
Add password4j implementation of PasswordEncoder


rwinch:
We should add a (or multiple?) [password4j](https://github.com/Password4j/password4j) implementation of `PasswordEncoder`. I haven't looked at the API, so we might be able to get away with a single implementation (e.g. `Password4jPasswordEncoder`) that supports multiple algorithms. If not, then we may need multiple implementations (e.g. `Password4jBCryptPasswordEncoder`, `Password4jSCryptPasswordEncoder`, etc).

mehrdadbozorgmehr:
Can I work on these issues? @jzheaux  @rwinch

rwinch:
Thanks for volunteering @mehrdadbozorgmehr Yes please send a PR 😄

rwinch:
Closing this in favor of the pull request gh-17825

PR #17825, opened 2025-08-30
Add password4j implementation of PasswordEncoder 

mehrdadbozorgmehr:
Closes gh-17706

### DIFF
--- a/crypto/spring-security-crypto.gradle
+++ b/crypto/spring-security-crypto.gradle
@@ -8,7 +8,7 @@ dependencies {
 	management platform(project(":spring-security-dependencies"))
 	optional 'org.springframework:spring-core'
 	optional 'org.bouncycastle:bcpkix-jdk18on'
-	optional 'com.password4j:password4j'
+	optional libs.com.password4j.password4j
 
 	testImplementation "org.assertj:assertj-core"
 	testImplementation "org.junit.jupiter:junit-jupiter-api"
--- a/docs/modules/ROOT/pages/features/authentication/password-storage.adoc
+++ b/docs/modules/ROOT/pages/features/authentication/password-storage.adoc
@@ -483,37 +483,15 @@ This provides an alternative to Spring Security's built-in `Argon2PasswordEncode
 Argon2 is the winner of the https://en.wikipedia.org/wiki/Password_Hashing_Competition[Password Hashing Competition] and is recommended for new applications.
 This implementation leverages Password4j's Argon2 support which properly includes the salt in the output hash.
 
+Create an encoder with default settings:
+
 .Argon2Password4jPasswordEncoder
-[tabs]
-======
-Java::
-+
-[source,java,role="primary"]
-----
-// Create an encoder with default settings
-Argon2Password4jPasswordEncoder encoder = new Argon2Password4jPasswordEncoder();
-String result = encoder.encode("myPassword");
-assertTrue(encoder.matches("myPassword", result));
+include-code::./Argon2UsageTests[tag=default-params,indent=0]
 
-// Create an encoder with custom Argon2 function
-Argon2Function customArgon2 = Argon2Function.getInstance(65536, 3, 4, 32, Argon2.ID);
-Argon2Password4jPasswordEncoder customEncoder = new Argon2Password4jPasswordEncoder(customArgon2);
-----
+Create an encoder with custom Argon2 parameters:
 
-Kotlin::
-+
-[source,kotlin,role="secondary"]
-----
-// Create an encoder with default settings
-val encoder = Argon2Password4jPasswordEncoder()
-val result: String = encoder.encode("myPassword")
-assertTrue(encoder.matches("myPassword", result))
-
-// Create an encoder with custom Argon2 function
-val customArgon2 = Argon2Function.getInstance(65536, 3, 4, 32, Argon2.ID)
-val customEncoder = Argon2Password4jPasswordEncoder(customArgon2)
-----
-======
+.Argon2Password4jPasswordEncoder Custom
+include-code::./Argon2UsageTests[tag=custom-params,indent=0]
 
 [[password4j-bcrypt]]
 === BcryptPassword4jPasswordEncoder
@@ -524,37 +502,15 @@ This provides an alternative to Spring Security's built-in `BCryptPasswordEncode
 BCrypt is a well-established password hashing algorithm that includes built-in salt generation and is resistant to rainbow table attacks.
 This implementation leverages Password4j's BCrypt support which properly includes the salt in the output hash.
 
+Create an encoder with default settings:
+
 .BcryptPassword4jPasswordEncoder
-[tabs]
-======
-Java::
-+
-[source,java,role="primary"]
-----
-// Create an encoder with default settings
-BcryptPassword4jPasswordEncoder encoder = new BcryptPassword4jPasswordEncoder();
-String result = encoder.encode("myPassword");
-assertTrue(encoder.matches("myPassword", result));
+include-code::./BcryptUsageTests[tag=default-params,indent=0]
 
-// Create an encoder with custom round count
-BcryptFunction customBcrypt = BcryptFunction.getInstance(12);
-BcryptPassword4jPasswordEncoder customEncoder = new BcryptPassword4jPasswordEncoder(customBcrypt);
-----
+Create an encoder with custom bcrypt parameters:
 
-Kotlin::
-+
-[source,kotlin,role="secondary"]
-----
-// Create an encoder with default settings
-val encoder = BcryptPassword4jPasswordEncoder()
-val result: String = encoder.encode("myPassword")
-assertTrue(encoder.matches("myPassword", result))
-
-// Create an encoder with custom round count
-val customBcrypt = BcryptFunction.getInstance(12)
-val customEncoder = BcryptPassword4jPasswordEncoder(customBcrypt)
-----
-======
+.BcryptPassword4jPasswordEncoder Custom
+include-code::./BcryptUsageTests[tag=custom-params,indent=0]
 
 [[password4j-scrypt]]
 === ScryptPassword4jPasswordEncoder
@@ -565,37 +521,16 @@ This provides an alternative to Spring Security's built-in `SCryptPasswordEncode
 SCrypt is a memory-hard password hashing algorithm designed to be resistant to hardware brute-force attacks.
 This implementation leverages Password4j's SCrypt support which properly includes the salt in the output hash.
 
+
+Create an encoder with default settings:
+
 .ScryptPassword4jPasswordEncoder
-[tabs]
-======
-Java::
-+
-[source,java,role="primary"]
-----
-// Create an encoder with default settings
-ScryptPassword4jPasswordEncoder encoder = new ScryptPassword4jPasswordEncoder();
-String result = encoder.encode("myPassword");
-assertTrue(encoder.matches("myPassword", result));
+include-code::./ScryptUsageTests[tag=default-params,indent=0]
 
-// Create an encoder with custom SCrypt parameters
-ScryptFunction customScrypt = ScryptFunction.getInstance(32768, 8, 1, 32);
-ScryptPassword4jPasswordEncoder customEncoder = new ScryptPassword4jPasswordEncoder(customScrypt);
-----
+Create an encoder with custom scrypt parameters:
 
-Kotlin::
-+
-[source,kotlin,role="secondary"]
-----
-// Create an encoder with default settings
-val encoder = ScryptPassword4jPasswordEncoder()
-val result: String = encoder.encode("myPassword")
-assertTrue(encoder.matches("myPassword", result))
-
-// Create an encoder with custom SCrypt parameters
-val customScrypt = ScryptFunction.getInstance(32768, 8, 1, 32)
-val customEncoder = ScryptPassword4jPasswordEncoder(customScrypt)
-----
-======
+.ScryptPassword4jPasswordEncoder Custom
+include-code::./ScryptUsageTests[tag=custom-params,indent=0]
 
 [[password4j-pbkdf2]]
 === Pbkdf2Password4jPasswordEncoder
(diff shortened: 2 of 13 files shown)


mehrdadbozorgmehr:
Here is my PR. I’d really appreciate any feedback or suggestions for improvements.
Thanks for your time and guidance 🙏
@rwinch

crypto/src/main/java/org/springframework/security/crypto/password4j/Password4jPasswordEncoder.java
```diff
@@ -0,0 +1,249 @@
...
+ * Implementation of {@link org.springframework.security.crypto.password.PasswordEncoder} that uses the Password4j library.
+ * This encoder supports multiple password hashing algorithms including BCrypt, SCrypt, Argon2, and PBKDF2.
+ *
+ * <p>The encoder determines the algorithm used based on the algorithm type specified during construction.
+ * For verification, it can automatically detect the algorithm used in existing hashes.</p>
+ *
+ * <p>This implementation is thread-safe and can be shared across multiple threads.</p>
+ *
+ * @author Mehrdad Bozorgmehr
+ * @since 6.5
```
rwinch:
```suggestion
 * @since 7.0
```

Since this is a new feature it would get merged with 7.0

crypto/src/main/java/org/springframework/security/crypto/password4j/Password4jPasswordEncoder.java
```diff
@@ -0,0 +1,249 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.crypto.password4j;
+
+import com.password4j.*;
```
rwinch: We do not use wildcard imports. You should run `./gradlew check` to perform all checks (including checkstyle). You can run `./gradlew checkstyleMain checkstyleTest` to just do checkstyle tests to speed things up.

crypto/src/main/java/org/springframework/security/crypto/factory/PasswordEncoderFactories.java
```diff
@@ -87,6 +92,14 @@ public static PasswordEncoder createDelegatingPasswordEncoder() {
...
 		encoders.put("argon2", Argon2PasswordEncoder.defaultsForSpringSecurity_v5_2());
 		encoders.put("argon2@SpringSecurity_v5_8", Argon2PasswordEncoder.defaultsForSpringSecurity_v5_8());
+
+		// Password4j implementations
+		encoders.put("password4j-bcrypt", Password4jPasswordEncoder.bcrypt(10));
+		encoders.put("password4j-scrypt", Password4jPasswordEncoder.scrypt(16384, 8, 1, 32));
+		encoders.put("password4j-argon2", Password4jPasswordEncoder.argon2(65536, 3, 4, 32,
+			com.password4j.types.Argon2.ID));
+		encoders.put("password4j-pbkdf2", Password4jPasswordEncoder.pbkdf2(310000, 32));
+
```
rwinch:
```suggestion
```

I don't think we should be prefixing any of the algorithms with password4j because validating bcrypt work with any bcrypt implementation. Since we already have support for all of these algorithms, I'd prefer to just remove these for now.

After this ticket, we might consider adding a ticket `Classpath Selection for PasswordEncoder Implementation`. Before implementing it, we'd want to discuss how it would work and determine if this is something we want to do.

crypto/src/main/java/org/springframework/security/crypto/password4j/Password4jPasswordEncoder.java
```diff
@@ -0,0 +1,249 @@
...
+		this.hashingFunction = createDefaultHashingFunction(algorithm);
+	}
+
+	/**
+	 * Constructs a Password4j password encoder with a custom hashing function.
+	 *
+	 * @param hashingFunction the custom hashing function to use
+	 * @param algorithm       the password hashing algorithm type
+	 */
+	public Password4jPasswordEncoder(HashingFunction hashingFunction, Password4jAlgorithm algorithm) {
```
rwinch: This does not validate that the `HashingFunction` and the `Password4jAlgorithm` are aligned. For example, this should fail but doesn't

```java
new Password4jPasswordEncoder(
                AlgorithmFinder.getBcryptInstance(),
                Password4jPasswordEncoder.Password4jAlgorithm.ARGON2);
```

As [mentioned previously](https://github.com/spring-projects/spring-security/pull/17825/files#diff-9c9904ce161dfc76ad4c40b678206d91204142b10dc6b439aa7f4d45f3262ed2R50), I think we can remove `Password4jAlgorithm` all together. However, I wanted to point this out as another reason we do not want/need it.

Instead, please have a single constructor that allows injecting a non-null `HashingFunction`. Please be sure to document how to create instances in the javadoc. For example, defaults can be obtained from `AlgorithmFinder`. However, an example of a single algorithm that has been customized would be useful as well.

crypto/src/main/java/org/springframework/security/crypto/password4j/Password4jPasswordEncoder.java
```diff
@@ -0,0 +1,249 @@
...
+	 * Creates a Password4j password encoder with default settings for Spring Security v5.8+.
+	 * This uses BCrypt with 10 rounds.
+	 *
+	 * @return a new Password4j password encoder with recommended defaults
+	 * @since 6.5
+	 */
+	public static Password4jPasswordEncoder defaultsForSpringSecurity() {
+		return bcrypt(10);
+	}
+
```
rwinch:
```suggestion
```

I think that all of the static factory methods can be removed. Users should be encouraged to take advantage of password4j's `AlgorithmFinder` API to create a  `HashingFunction` instance. This reduces the maintenance for Spring Security and also gives users more flexibility.

crypto/src/main/java/org/springframework/security/crypto/password4j/Password4jPasswordEncoder.java
```diff
@@ -0,0 +1,249 @@
...
+	/**
+	 * Gets the hashing function used by this encoder.
+	 *
+	 * @return the hashing function
+	 */
+	public HashingFunction getHashingFunction() {
+		return this.hashingFunction;
+	}
```
rwinch:
```suggestion
```

We generally try to avoid exposing getters as this breaks encapsulation. For example, a user may inject a Jackson `ObjectMapper` and that might be used internally for a while. 

```java
public class SomeClass {
    // jackson object mapper
    private ObjectMapper objectMapper;

    public SomeClass(ObjectMapper objectMapper) {
        this.objectMapper = Objects.requireNonNull(objectMapper);
    }

    public SomePojo convert(String string) {
         // ... use objectMapper for conversion
    }

    // makes refactoring hard
    public ObjectMapper getObjectMapper() {
        return this.objectMapper;
    }
}
```


If we expose a getObjectMapper() method, then it makes breaks encapsulation and makes refactoring more difficult. For example, if we want more flexibility we might decide to support a `Converter`

```java
public class SomeClass {
    private Converter converter;


    public SomeClass(Converter converter) {
        this.converter = Objects.requireNonNull(converter);
    }

    public SomeClass(ObjectMapper objectMapper) {
        this(new MappingJackson2HttpMessageConverter(objectMapper));
    }

    public SomePojo convert(String string) {
         // ... can update to use converter because its internal
    }

    // makes refactoring hard
    public ObjectMapper getObjectMapper() {
        // Cannot be passive here because the converter replaced ObjectMapper!!
    }
}
```

crypto/src/main/java/org/springframework/security/crypto/password4j/Password4jPasswordEncoder.java
```diff
@@ -0,0 +1,249 @@
...
+		/**
+		 * PBKDF2 algorithm.
+		 */
+		PBKDF2,
+		/**
+		 * Compressed PBKDF2 algorithm.
+		 */
+		COMPRESSED_PBKDF2
+	}
+
```
rwinch:
```suggestion
```

Please remove `Password4jAlgorithm`.

Generally, we try to avoid implementation specific APIs on our interfaces. However, behind an interface we encourage injecting the implementation directly rather than abstracting it out. This gives the users as much flexibility as possible and reduces the complexity.

crypto/src/test/java/org/springframework/security/crypto/password4j/Password4jPasswordEncoderTests.java
```diff
@@ -0,0 +1,596 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatIllegalArgumentException;
+
+/**
+ * Tests for {@link Password4jPasswordEncoder}.
+ *
+ * @author Mehrdad Bozorgmehr
+ */
+class Password4jPasswordEncoderTests {
```
rwinch, line 30: I'd assume that the underlying password4j library works, so we don't need tests like timing attacks, strong randomness, etc. Instead, you can inject a mocked `HashingFunction` and validate the functionality that way.

There likely is some value in a quick sanity check for each `HashingFunction` on `AlgorithmFinder` to ensure that a valid password is true and then another one that ensures that an invalid password is false. However, I would not try to cover every edge case.

In a new `PasswordCompatabilityTests` class is probably also value in ensuring that the existing encoders work with Password4j as well. For example, encode with `BCryptPasswordEncoder` and then validate with the `Password4jPasswordEncoder` and then vice versa.

crypto/src/main/java/org/springframework/security/crypto/password4j/Password4jPasswordEncoder.java
```diff
@@ -0,0 +1,138 @@
...
+	@Override
+	protected boolean matchesNonNull(String rawPassword, String encodedPassword) {
+		try {
+			// Use the specific hashing function for verification
+			return Password.check(rawPassword, encodedPassword).with(this.hashingFunction);
+		}
+		catch (Exception ex) {
+			this.logger.warn("Password verification failed for encoded password: " + encodedPassword, ex);
```
rwinch: We should not log the encodedPassword since it could be a weak encoding that caused the failure and that could reveal the password.

crypto/src/main/java/org/springframework/security/crypto/password4j/Password4jPasswordEncoder.java
```diff
@@ -0,0 +1,138 @@
...
+			return hash.getResult();
+		}
+		catch (Exception ex) {
+			throw new IllegalStateException("Failed to encode password using Password4j", ex);
+		}
+	}
+
+	@Override
+	protected boolean matchesNonNull(String rawPassword, String encodedPassword) {
+		try {
```
rwinch: There is no need for a try/catch here

crypto/src/main/java/org/springframework/security/crypto/password4j/Password4jPasswordEncoder.java
```diff
@@ -0,0 +1,138 @@
...
+	 * @throws IllegalArgumentException if hashingFunction is null
+	 */
+	public Password4jPasswordEncoder(HashingFunction hashingFunction) {
+		Assert.notNull(hashingFunction, "hashingFunction cannot be null");
+		this.hashingFunction = hashingFunction;
+	}
+
+	@Override
+	protected String encodeNonNullPassword(String rawPassword) {
+		try {
```
rwinch: There is no need for a try/catch here

crypto/src/test/java/org/springframework/security/crypto/password4j/Password4jPasswordEncoderTests.java
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	@Test
+	void algorithmFinderScryptSanityCheck() {
+		Password4jPasswordEncoder encoder = new Password4jPasswordEncoder(AlgorithmFinder.getScryptInstance());
+
+		String encoded = encoder.encode(PASSWORD);
+		assertThat(encoder.matches(PASSWORD, encoded)).isTrue();
+		assertThat(encoder.matches(WRONG_PASSWORD, encoded)).isFalse();
+	}
```
rwinch: I noticed that not all `HashingFunction` implementations were being tested and ran the following test:

```
	@ParameterizedTest
	@MethodSource("hashingFunctions")
	void encodeAndMatch(HashingFunction hashFn) {
		Password4jPasswordEncoder encoder = new Password4jPasswordEncoder(hashFn);

		String encoded = encoder.encode(PASSWORD);
		assertThat(encoder.matches(PASSWORD, encoded)).isTrue();
		assertThat(encoder.matches(WRONG_PASSWORD, encoded)).isFalse();
	}

	static HashingFunction[] hashingFunctions() {
		return new HashingFunction[] { AlgorithmFinder.getArgon2Instance(), AlgorithmFinder.getBcryptInstance(),
				AlgorithmFinder.getScryptInstance(), AlgorithmFinder.getBalloonHashingInstance(),
				AlgorithmFinder.getPBKDF2Instance() };
	}
```

It appears that `getPBKDF2Instance()` and `getBalloonHashingInstance()` fail. At least part of the problem is that Password4j does not include the resulting salt in the output.

For this reason, we will need to back track a little on design (sorry about that). 

1) Make `Password4jPasswordEncoder` abstract and package private
2) Create public subclasses that only accept the `HashingFunction` implementations that work (e.g. exclude pbkdf2 and baloon hashing function but create the ability to have `new BcryptPassword4jEncoder(BcryptFunction)`)
3) Create a specific implementation (if you like we can do this in a separate PR) for the algorithms that are not working. For example, the PBKDF2 implementation will need to combine the salt into the result of calling encode. Then for the match it will extract that value out.

rwinch (changes requested):
While we can reuse it as a base class for most of the algorithms, it looks like a single `Password4jPasswordEncoder` will not be possible. I've provided some feedback inline. 

PS: Sorry for finding this late and reverting some of my previous feedback.

mehrdadbozorgmehr:
Thanks for pointing this out. I’ve updated the design by making Password4jPasswordEncoder abstract and introducing algorithm-specific subclasses for the working implementations. I’ve also implemented a PBKDF2-specific encoder and Balloon hashing in this PR with proper salt handling. 

@rwinch , please let me know if any additional adjustments are required.

mehrdadbozorgmehr:
> Thank you @mehrdadbozorgmehr!
> 
> I pushed a few minor changes to this branch with explanations for the changes. I've marked this to merge as soon as the build passes.

@rwinch Thanks so much for your review and guidance! I really appreciate it. 🙏

PR #17832, opened 2025-09-01
Add Jackson 3 support and deprecate Jackson 2 one

sdeleuze:
This PR adds support for Jackson 3 which has the following major differences with the Jackson 2 one:
 - `jackson` subpackage instead of `jackson2`
 - `Jackson` type prefix instead of `Jackson2`
 - `JsonMapper` instead of `ObjectMapper`
 - For configuration, `JsonMapper.Builder` instead of `ObjectMapper` since the latter is now immutable
 - `AllowlistTypeResolverBuilder` in new a public type in order to be used easily with the `JsonMapper.Builder` API

Jackson 3 changes compared to Jackson 2 are documented on https://github.com/FasterXML/jackson-future-ideas/discussions/72.

It also deprecates Jackson 2 support for removal.

### DIFF
--- a/cas/src/main/java/org/springframework/security/cas/jackson/CasJacksonModule.java
+++ b/cas/src/main/java/org/springframework/security/cas/jackson/CasJacksonModule.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2004-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.cas.jackson;
+
+import org.apereo.cas.client.authentication.AttributePrincipalImpl;
+import org.apereo.cas.client.validation.AssertionImpl;
+import tools.jackson.core.Version;
+import tools.jackson.databind.cfg.MapperBuilder;
+import tools.jackson.databind.module.SimpleModule;
+
+import org.springframework.security.cas.authentication.CasAuthenticationToken;
+import org.springframework.security.jackson.AllowlistTypeResolverBuilder;
+import org.springframework.security.jackson.SecurityJacksonModules;
+
+/**
+ * Jackson module for spring-security-cas. This module register
+ * {@link AssertionImplMixin}, {@link AttributePrincipalImplMixin} and
+ * {@link CasAuthenticationTokenMixin}. If no default typing enabled by default then it'll
+ * enable it because typing info is needed to properly serialize/deserialize objects. In
+ * order to use this module just add this module into your ObjectMapper configuration.
+ *
+ * <pre>
+ * JsonMapper mapper = JsonMapper.builder()
+ * 				.addModule(new CasJacksonModule())
+ * 				.build();
+ * </pre>
+ *
+ * <b>Note: use {@link SecurityJacksonModules#getModules(ClassLoader)} to get list of all
+ * security modules on the classpath.</b>
+ *
+ * @author Sebastien Deleuze
+ * @author Jitendra Singh
+ * @since 7.0
+ * @see SecurityJacksonModules
+ */
+public class CasJacksonModule extends SimpleModule {
+
+	public CasJacksonModule() {
+		super(CasJacksonModule.class.getName(), new Version(1, 0, 0, null, null, null));
+	}
+
+	@Override
+	public void setupModule(SetupContext context) {
+		((MapperBuilder<?, ?>) context.getOwner()).setDefaultTyping(new AllowlistTypeResolverBuilder());
+		context.setMixIn(AssertionImpl.class, AssertionImplMixin.class);
+		context.setMixIn(AttributePrincipalImpl.class, AttributePrincipalImplMixin.class);
+		context.setMixIn(CasAuthenticationToken.class, CasAuthenticationTokenMixin.class);
+	}
+
+}
--- a/core/src/main/java/org/springframework/security/jackson/AbstractUnmodifiableCollectionDeserializer.java
+++ b/core/src/main/java/org/springframework/security/jackson/AbstractUnmodifiableCollectionDeserializer.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2004-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.jackson;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Set;
+
+import tools.jackson.core.JacksonException;
+import tools.jackson.core.JsonParser;
+import tools.jackson.databind.DeserializationContext;
+import tools.jackson.databind.JsonNode;
+import tools.jackson.databind.ValueDeserializer;
+import tools.jackson.databind.node.ArrayNode;
+
+/**
+ * Abstract base class for deserializers that create unmodifiable collections from JSON
+ * data. Subclasses like {@link UnmodifiableListDeserializer} and
+ * {@link UnmodifiableSetDeserializer} should implement the method to define the specific
+ * collection type and handle the deserialization logic.
+ *
+ * @param <T> the type of the unmodifiable collection, such as {@link List} or
+ * {@link Set}.
+ * @author Sebastien Deleuze
+ * @author Hyunmin Choi
+ * @since 7.0
+ */
+abstract class AbstractUnmodifiableCollectionDeserializer<T> extends ValueDeserializer<T> {
+
+	@Override
+	public T deserialize(JsonParser jp, DeserializationContext ctxt) throws JacksonException {
+		JsonNode node = ctxt.readTree(jp);
+		Collection<Object> values = new ArrayList<>();
+		if (node instanceof ArrayNode arrayNode) {
+			for (JsonNode elementNode : arrayNode) {
+				values.add(ctxt.readTreeAsValue(elementNode, Object.class));
+			}
+		}
+		else if (node != null) {
+			values.add(ctxt.readTreeAsValue(node, Object.class));
+		}
+		return createUnmodifiableCollection(values);
+	}
+
+	/**
+	 * Creates an unmodifiable collection from the given JSON node.
+	 * @param values the values to add to the unmodifiable collection
+	 * @return an unmodifiable collection with the deserialized elements.
+	 * @throws JacksonException if an error occurs during deserialization.
+	 */
+	abstract T createUnmodifiableCollection(Collection<Object> values) throws JacksonException;
+
+}
--- a/core/src/main/java/org/springframework/security/jackson/AllowlistTypeResolverBuilder.java
+++ b/core/src/main/java/org/springframework/security/jackson/AllowlistTypeResolverBuilder.java
@@ -1,0 +1,161 @@
+/*
+ * Copyright 2004-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.jackson;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+import com.fasterxml.jackson.annotation.JacksonAnnotation;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import tools.jackson.databind.DatabindContext;
+import tools.jackson.databind.DefaultTyping;
+import tools.jackson.databind.DeserializationConfig;
+import tools.jackson.databind.JavaType;
+import tools.jackson.databind.jsontype.BasicPolymorphicTypeValidator;
+import tools.jackson.databind.jsontype.NamedType;
+import tools.jackson.databind.jsontype.PolymorphicTypeValidator;
+import tools.jackson.databind.jsontype.TypeIdResolver;
+import tools.jackson.databind.jsontype.impl.DefaultTypeResolverBuilder;
+
+import org.springframework.core.annotation.AnnotationUtils;
+
+/**
+ *
+ * An implementation of {@link DefaultTypeResolverBuilder} that inserts an
+ * {@code allow all} {@link PolymorphicTypeValidator} and overrides the
+ * {@code TypeIdResolver}
+ *
+ * @author Sebastien Deleuze
+ * @author Rob Winch
+ * @since 7.0
+ */
+public class AllowlistTypeResolverBuilder extends DefaultTypeResolverBuilder {
+
+	public AllowlistTypeResolverBuilder() {
+		this(DefaultTyping.NON_FINAL);
+	}
+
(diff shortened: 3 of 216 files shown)


cas/src/main/java/org/springframework/security/cas/jackson/CasJacksonModule.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+public class CasJacksonModule extends SimpleModule {
+
+	public CasJacksonModule() {
+		super(CasJacksonModule.class.getName(), new Version(1, 0, 0, null, null, null));
+	}
+
+	@Override
+	public void setupModule(SetupContext context) {
+		((MapperBuilder<?, ?>) context.getOwner()).setDefaultTyping(new AllowlistTypeResolverBuilder());
```
rwinch: This approach appears to create a new instance of `AllowlistTypeResolverBuilder` for every module. Is that necessary or do we only want a single instance?

core/src/main/java/org/springframework/security/jackson/AbstractUnmodifiableCollectionDeserializer.java
```diff
@@ -0,0 +1,68 @@
...
+ * {@link UnmodifiableSetDeserializer} should implement the method to define the specific
+ * collection type and handle the deserialization logic.
+ *
+ * @param <T> the type of the unmodifiable collection, such as {@link List} or
+ * {@link Set}.
+ * @author Sebastien Deleuze
+ * @author Hyunmin Choi
+ * @since 7.0
+ */
+abstract class AbstractUnmodifiableCollectionDeserializer<T> extends ValueDeserializer<T> {
```
rwinch: I wonder if there is better built in support for unmodifiable collections now?

core/src/main/java/org/springframework/security/jackson/AllowlistTypeResolverBuilder.java
```diff
@@ -0,0 +1,161 @@
...
+ *
+ * An implementation of {@link DefaultTypeResolverBuilder} that inserts an
+ * {@code allow all} {@link PolymorphicTypeValidator} and overrides the
+ * {@code TypeIdResolver}
+ *
+ * @author Sebastien Deleuze
+ * @author Rob Winch
+ * @since 7.0
+ */
+public class AllowlistTypeResolverBuilder extends DefaultTypeResolverBuilder {
```
rwinch: Is there a reason that this is now public vs being encapsulated as it is for the Jackson 2 support?

rwinch: I think that now is a good time for Spring Security to stop using global default typing. 

While the `AllowlistTypeResolverBuilder` worked well as a security patch in the Jackson 2 support, I think that we should remove it for Jackson 3 support and better align with best practices with Jackson. Instead, I think that we should use https://github.com/FasterXML/jackson-docs/wiki/JacksonPolymorphicDeserialization#12-per-class-annotations This would mean that users would need to take the risk of enabling global default typing or opt into default typing per type.

I'd like to ensure that we have a test that verifies that a custom type does not just work after applying Spring Security's module. Another test would verify that users could add their own type. The docs would be updated to explain that global default typing is off, why it is off, and how they can add their own types.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/jackson/Saml2JacksonModule.java
```diff
@@ -0,0 +1,72 @@
...
+@SuppressWarnings("serial")
+public class Saml2JacksonModule extends SimpleModule {
+
+	public Saml2JacksonModule() {
+		super(Saml2JacksonModule.class.getName(), new Version(1, 0, 0, null, null, null));
+	}
+
+	@Override
+	public void setupModule(SetupContext context) {
+		// TODO Is it expected that default typing in not configured in the Jackson 2
```
rwinch: This is not expected, but please see my [comment](https://github.com/spring-projects/spring-security/pull/17832/files#diff-969e063e4e2fc7da1261e91e59ca43be2709f82129df5dc516f9a40b17102b61R48) about no longer enabling global default typing.

saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/jackson2/Saml2Jackson2Module.java
```diff
@@ -50,6 +54,7 @@ public Saml2Jackson2Module() {
 
 	@Override
 	public void setupModule(SetupContext context) {
+		// TODO Is it expected that default typing in not configured here?
```
rwinch, line 57: This is not expected, but please see my [comment](https://github.com/spring-projects/spring-security/pull/17832/files#diff-969e063e4e2fc7da1261e91e59ca43be2709f82129df5dc516f9a40b17102b61R48) about no longer enabling global default typing.

rwinch (changes requested):
Thanks for the pull request. Generally, I think this is looking good. I've provided some feedback inline.

It also appears that there are still some classes that need migrated. Searching for jackson on main should give you a complete list of classes that are using jackson.

For example, [webauthn has jackson support](https://github.com/spring-projects/spring-security/tree/7.0.0-M2/webauthn/src/main/java/org/springframework/security/web/webauthn/jackson). I'm guessing this was missed because it is inconsistent with the rest of the jackson support in that it is in a jackson package rather than jackson2. It is also automatically registered because it does not use default types which would make it insecure.

There are also some tests that still use `ObjectMapper` or deprecated Spring Framework Jackson 2 classes (e.g. [JwtDecodersTests.java#L388](https://github.com/spring-projects/spring-security/blob/7.0.0-M2/oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/JwtDecodersTests.java#L388)). Unless the test is specifically for the Jackson 2 support, we should update these to use non-deprecated classes (e.g. `JsonMapper`).


There are various Spring framework jackson based classes that have been deprecated that we should migrate away from. For example, `MappingJackson2HttpMessageConverter` usage (e.g. [WebAuthnAuthenticationFilter.converter](https://github.com/spring-projects/spring-security/blob/7.0.0-M2/webauthn/src/main/java/org/springframework/security/web/webauthn/authentication/WebAuthnAuthenticationFilter.java#L75)) should be migrated to `JacksonJsonHttpMessageConverter` if jackson 3 is on the classpath. Likely there could be a static factory method used that Spring Security uses to obtain the correct default json converter instance.

core/src/main/java/org/springframework/security/jackson/AllowlistTypeResolverBuilder.java
sdeleuze: From your comment, looks like Spring Security Jackson 3 modules would stop configuring `AllowlistTypeResolverBuilder`. But would we keep a Jackson 3 `AllowlistTypeResolverBuilder` or not (I think not but please confirm)?

Do you expect Jackson 3 mixins provided by Spring Security to use per class annotations instead?

> This would mean that users would need to take the risk of enabling global default typing or opt into default typing per type.

I don't understand this sentence, could you please elaborate?

> The docs would be updated to explain that global default typing is off, why it is off, and how they can add their own types

A raw draft of what the doc should say would be useful for my own understanding and for this PR, I will likely let you refine post-PR.

core/src/main/java/org/springframework/security/jackson/AbstractUnmodifiableCollectionDeserializer.java
sdeleuze: Could you please detail the former issues you had for, or give me related pointer (issue, tests checking this)?

core/src/main/java/org/springframework/security/jackson/AllowlistTypeResolverBuilder.java
sdeleuze: With `ObjectMapper` / `JsonMapper` now immutable and the usage of the builder, the former static utility method looked very unatural to use from an API perspective and not very discoverable, hence my proposal to expose the type (but maybe not needed anymore, depending on your answer to [this comment](https://github.com/spring-projects/spring-security/pull/17832#discussion_r2330681178)).

cas/src/main/java/org/springframework/security/cas/jackson/CasJacksonModule.java
sdeleuze: Maybe not relevant anymore depending on you answer to [this comment](https://github.com/spring-projects/spring-security/pull/17832#discussion_r2330681178). We could potentially use a shared instance if you think that's important and if still relevant.

sdeleuze:
@rwinch I have pushed additional commits:
 - [Upgrade to Jackson 3.0.0-rc9 and 2.20.0](https://github.com/sdeleuze/spring-security/commit/51d8816ec599f34d46140d65752389824baba1ee)
 - [Change ObjectMapper to JsonMapper where relevant](https://github.com/sdeleuze/spring-security/commit/82312db391813de66381b04fff0e96a686e0dbb6)
 - [Add webauthn Jackson 3 support and deprecate Jackson 2 one](https://github.com/sdeleuze/spring-security/commit/bafa603488c9c5878e26ebe225e34c01b6c6f1f2)

I will do another pass with your feedback to my questions in the comments.

core/src/main/java/org/springframework/security/jackson/AllowlistTypeResolverBuilder.java
rwinch: > From your comment, looks like Spring Security Jackson 3 modules would stop configuring AllowlistTypeResolverBuilder. But would we keep a Jackson 3 AllowlistTypeResolverBuilder or not (I think not but please confirm)?

No. We should remove Jackson 3 `AllowlistTypeResolverBuilder` so that we don't need to maintain that logic once we remove Jackson 2 support.

> Do you expect Jackson 3 mixins provided by Spring Security to use per class annotations instead?

Yes.

> I don't understand this sentence, could you please elaborate?

Sorry that this wasn't clear.

Previously, Spring Security's Jackson support forced enabling global default typing which is dangerous due to deserialization gadgets that might be exposed. Spring Security's Jackson 3 support should only enable per class (with classes that are safe). If users want global default typing, they will need to do that explicitly outside of Spring Security's Jackson 3 support.

> A raw draft of what the doc should say would be useful for my own understanding and for this PR, I will likely let you refine post-PR.

Expanding this statement, but keeping to a summary the docs would:
1) Explain that Spring Security's Jackson 3 support does not enable [global default typing](https://github.com/FasterXML/jackson-docs/wiki/JacksonPolymorphicDeserialization#11-global-default-typing) because it is [insecure](https://github.com/FasterXML/jackson-docs/wiki/JacksonPolymorphicDeserialization#111-security-risks-using-global-default-typing).
2) If users want to add additional classes, they can use [per-class annotations](https://github.com/FasterXML/jackson-docs/wiki/JacksonPolymorphicDeserialization#12-per-class-annotations). Alternatively, if users understand the security risks they can enable global default typing themselves.

rwinch: Yes. I think that this should ultimately be removed.

core/src/main/java/org/springframework/security/jackson/AbstractUnmodifiableCollectionDeserializer.java
rwinch: I think that this is it here:

* https://github.com/FasterXML/jackson-databind/issues/1881
* https://github.com/FasterXML/jackson-databind/issues/1880

cas/src/main/java/org/springframework/security/cas/jackson/CasJacksonModule.java
rwinch: Yes. I think that this should ultimately be removed.

core/src/main/java/org/springframework/security/jackson/AllowlistTypeResolverBuilder.java
sdeleuze: So I made a try to remove global default typing, that's **very** involved. Almost all the tests are broken and this require much more refinements and thinking than adding annotation to individual classes as collection deserialization is broken in multiple ways, mixins require significant updates, etc.

Not saying you should not do it, but that is not directly related to the Jackson 2 to 3 migration, I don't have a real added value here, I lack of Spring Security knowledge to decide what are the right choices to do and this PR is already pretty huge.

I think the side effects should be carefully assessed by the Spring Security team, as it would be a major breaking change that would add its side effects to the Jackson 3 migration itself. It should probably have its own dedicated issue and I recommend handling that after merging this PR.

sdeleuze:
@rwinch I have removed the custom support for unmodifiable collections (seems to work fine with Jackson 3), reworked the commit for a cleaner git history, and rebased the commits on `main`.

jvalkeal:
@sdeleuze @rwinch I was wondering if message converters in `oauth2-core` will also be part of this PR.

https://github.com/spring-projects/spring-security/blob/f3761aff990cf043b5793a609623eda8e2756424/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/http/converter/HttpMessageConverters.java#L41-L47

That's then causing NPE's i.e in `OAuth2AccessTokenResponseHttpMessageConverter` as `jsonMessageConverter` will always be null. Same with other classes in that same package if I only have jackson3 in a classpath.

sdeleuze:
@jvalkeal Good catch, I will add a related commit.

core/src/main/java/org/springframework/security/jackson/AllowlistTypeResolverBuilder.java
sdeleuze: After more work, I have pushed a commit that removes the global default typing and replace it by a safer alternative using a `PolymorphicTypeValidator` (see [related Jackson blog post](https://cowtowncoder.medium.com/jackson-2-10-safe-default-typing-2d018f0ce2ba)) that is contributed by the various modules.

I still need to add an API for end-user customization of the authorized classes.

sdeleuze:
@rwinch The PR is ready for a new review.

It removes global default typing, allow customization of the authorized types via a [`PolymorphicTypeValidator `](https://cowtowncoder.medium.com/jackson-2-10-safe-default-typing-2d018f0ce2ba), including by the end user. The `HttpMessageConverters` issue reported by @jvalkeal has also been fixed.

The main missing part is the "Jackson 2 compatibility mode with Jackson 3". It will be implemented via a `SecurityJackson2CompatibilityModules` that will create a builder with `JsonMapper.builderWithJackson2Defaults()`, reuse Jackson 2 mixins and apply them to Jackson 3 and keep global default typing enabled.

sdeleuze:
PR updated to use Jackson 3.0.0 GA which has just been released.

The `DefaultLoginPageGeneratingFilterTests` broken test on the CI works fine locally, and was fine with previous build with Jackson 3.0.0-rc10 which is identical to 3.0.0, so I guess that's flaky test that should be fine in the next build.

sdeleuze:
Looks like I have missed `OAuth2AuthorizationServerJackson2Module` which has been added recently, so I will migrate that one as well.

rwinch:
Thanks for all the updates @sdeleuze I'm reviewing now.

> The `DefaultLoginPageGeneratingFilterTests` broken test on the CI works fine locally, and was fine with previous build with Jackson 3.0.0-rc10 which is identical to 3.0.0, so I guess that's flaky test that should be fine in the next build.

If you rebase, this should be fixed in https://github.com/spring-projects/spring-security/commit/83da86a3588e147a4933f31dfc4bcdfd6f71710b which will be resolved more comprehensively in https://github.com/spring-projects/spring-security/issues/18012

sdeleuze:
@rwinch Rebasing is blocked on `FactorGrantedAuthority` not providing any public constructor but only a builder. It is possible to configure Jackson to use it, but supporting both Jackson 2 and 3 will require a bit more work, so before doing that I would like to double check with you that's expected.

sdeleuze:
@rwinch I have pushed a new update of the PR with the compatibility mode and the Jackson 3 support for `oauth2-authorization-server`. The latter was very hard to achieve due to the combination of global default typing disabled, top level `Map<String, Object>` and the usage of nested  `Object` based collection. Said otherwise, `oauth2-authorization-server` design is strongly relying on global default typing.

As discussed in https://github.com/FasterXML/jackson-databind/issues/3583, with the default `MapSerializer`,  values don't have type information, so I had to provide custom `MapSerializer` and `MapDeserializer` which could be tricky to maintain and trigger regression.

More generally, I am quite happy about the Jackson 3 migration itself but still have doubts that disabling global default typing is the right path. The tests passes but Jackson does not allow to enable type information serialization transitively on values, and I am wondering if [safe-default-typing](https://cowtowncoder.medium.com/jackson-2-10-safe-default-typing-2d018f0ce2ba) via `activateDefaultTyping(PolymorphicTypeValidator subtypeValidator, DefaultTyping applicability, JsonTypeInfo.As includeAs)` instead of just doing `polymorphicTypeValidator(PolymorphicTypeValidator subtypeValidator)` would be something you would consider. We should probably have a related discussion.

sdeleuze:
After a discussion with @rwinch, we agreed on using [safe default typing](https://cowtowncoder.medium.com/jackson-2-10-safe-default-typing-2d018f0ce2ba) instead of disabling it totally. I will refine the PR accordingly.

sdeleuze:
@rwinch The PR is ready for review in its hopefully final incarnation modulo your potential refinements.

rwinch:
Thank you for all of your hard work on this @sdeleuze This, along with some changes, have been merged into main. You can see the commits on this PR for details around the changes that were applied. I've also created some tickets to resolve some remaining issues:

* gh-18078
* gh-18077
* gh-18073

paulvas:
spring-security-web org.springframework.security.web.authentication.HttpMessageConverterAuthenticationSuccessHandler is also using MappingJackson2HttpMessageConverter. @rwinch I can migrate it, if nobody else is working on it.

Issue #17806, opened 2025-08-25
Support @ClientRegistrationId at Class Level


rwinch:
So that users do not need to repeat the annotation for every method, we should support adding `@ClientRegistrationId` at the class level

cc @joshlong

bbudano:
Hi @rwinch , I'd like to work on this enhancement.

rwinch:
Thank you @bbudano! I've assigned the issue to you and will await your PR.

bbudano:
@rwinch I've submitted a PR. Looking forward to your feedback!

PR #17838, opened 2025-09-03
Support @ClientRegistrationId at Class Level

bbudano:
If `@ClientRegistrationId` is not found on a method, the declaring class will be checked instead.

Closes gh-17806

### DIFF
--- a/docs/modules/ROOT/pages/features/integrations/rest/http-interface.adoc
+++ b/docs/modules/ROOT/pages/features/integrations/rest/http-interface.adoc
@@ -51,6 +51,13 @@ include-code::./UserService[tag=getAuthenticatedUser]
 
 The xref:features/integrations/rest/http-interface.adoc#client-registration-id[`@ClientRegistrationId`] will be processed by xref:features/integrations/rest/http-interface.adoc#client-registration-id-processor[`ClientRegistrationIdProcessor`]
 
+[[type]]
+=== Type Level Declarations
+
+`@ClientRegistrationId` can also be added at the type level to avoid repeating the declaration on every method.
+
+include-code::./UserService[tag=type]
+
 [[client-registration-id-processor]]
 == `ClientRegistrationIdProcessor`
 
--- a/docs/modules/ROOT/pages/whats-new.adoc
+++ b/docs/modules/ROOT/pages/whats-new.adoc
@@ -49,7 +49,7 @@ http.csrf((csrf) -> csrf.spa());
 * Added OAuth2 Support for xref:features/integrations/rest/http-interface.adoc[HTTP Interface Integration]
 * Added support for custom `JwkSource` in `NimbusJwtDecoder`, allowing usage of Nimbus's `JwkSourceBuilder` API
 * Added builder for `NimbusJwtEncoder`, supports specifying an EC or RSA key pair or a secret key
-* Added support for `@ClientRegistrationId` at class level, eliminating the need for method level repetition
+* Added support for `@ClientRegistrationId` at the xref:features/integrations/rest/http-interface.adoc#type[type level], eliminating the need for method level repetition
 
 == SAML 2.0
 
--- a/docs/src/test/java/org/springframework/security/docs/features/integrations/rest/type/Hovercard.java
+++ b/docs/src/test/java/org/springframework/security/docs/features/integrations/rest/type/Hovercard.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2004-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.docs.features.integrations.rest.type;
+
+/**
+ * Used to ensure {@link UserService} compiles, but not show in the documentation.
+ *
+ * @author Rob Winch
+ */
+public record Hovercard() {
+}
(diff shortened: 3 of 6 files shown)


Inder782:
Sorry for the confusion! I accidentally referenced the wrong issue above. This PR is not related to that issue.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/client/ClientRegistrationIdProcessor.java
```diff
@@ -37,17 +38,20 @@ public final class ClientRegistrationIdProcessor implements HttpRequestValues.Pr
 
 	public static ClientRegistrationIdProcessor DEFAULT_INSTANCE = new ClientRegistrationIdProcessor();
 
+	private ClientRegistrationIdProcessor() {
```
rwinch: Please don't make changes that are unrelated to the PR

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/client/ClientRegistrationIdProcessorTests.java
```diff
@@ -48,32 +50,42 @@ void processWhenClientRegistrationIdPresentThenSet() {
...
 
 		String registrationId = ClientAttributes.resolveClientRegistrationId(builder.build().getAttributes());
-		assertThat(registrationId).isEqualTo(RestService.REGISTRATION_ID);
+		assertThat(registrationId).isEqualTo(REGISTRATION_ID);
 	}
 
 	@Test
 	void processWhenNoClientRegistrationIdPresentThenNull() {
 		HttpRequestValues.Builder builder = HttpRequestValues.builder();
-		Method hasClientRegistrationId = ReflectionUtils.findMethod(RestService.class, "noClientRegistrationId");
```
rwinch: While these changes are valuable, please stick to the changes related to this ticket. This helps us to track issues and ensure that, when applicable (e.g. a bug fix that is applicable to older supported branches), we can easily selectively apply changes to other branches.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/client/ClientRegistrationIdProcessorTests.java
```diff
@@ -86,9 +98,16 @@ interface RestService {
...
 
 	@Retention(RetentionPolicy.RUNTIME)
-	@ClientRegistrationId(RestService.REGISTRATION_ID)
+	@ClientRegistrationId(REGISTRATION_ID)
 	@interface MetaClientRegistrationId {
 
 	}
 
+	@ClientRegistrationId(REGISTRATION_ID)
+	interface AnnotatedRestService {
```
rwinch:
```suggestion
	interface TypeAnnotatedRestService {
```

This is more clear since all of the interfaces are annotated.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/client/ClientRegistrationIdProcessor.java
```diff
@@ -37,17 +38,20 @@ public final class ClientRegistrationIdProcessor implements HttpRequestValues.Pr
 
 	public static ClientRegistrationIdProcessor DEFAULT_INSTANCE = new ClientRegistrationIdProcessor();
```
rwinch, line 40:
```suggestion

	private SecurityAnnotationScanner<ClientRegistrationId> scanner = SecurityAnnotationScanners.requireUnique(ClientRegistrationId.class);
```

Use a `SecurityAnnotationScanner<ClientRegistrationId>` instead. This will automatically follow the same logic as we do for method security. It ensures that we do not get duplicate annotations which might provide conflicting information and thus use the wrong registration id.

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/client/ClientRegistrationIdProcessor.java
```diff
@@ -37,17 +38,20 @@ public final class ClientRegistrationIdProcessor implements HttpRequestValues.Pr
...
+	}
+
 	@Override
 	public void process(Method method, MethodParameter[] parameters, @Nullable Object[] arguments,
 			HttpRequestValues.Builder builder) {
-		ClientRegistrationId registeredId = AnnotationUtils.findAnnotation(method, ClientRegistrationId.class);
+		ClientRegistrationId registeredId = Optional
+			.ofNullable(AnnotationUtils.findAnnotation(method, ClientRegistrationId.class))
+			.orElseGet(() -> AnnotationUtils.findAnnotation(method.getDeclaringClass(), ClientRegistrationId.class));
+
```
rwinch:
```suggestion
		ClientRegistrationId registeredId = this.scanner.scan(method, method.getDeclaringClass());
```

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/client/ClientRegistrationIdProcessorTests.java
```diff
@@ -39,6 +39,8 @@
  */
 class ClientRegistrationIdProcessorTests {
```
rwinch, line 42: Add tests that verify failure in the event that there are duplicate annotations. This should be handled by the `SecurityAnnotationScanners.requireUnique`. An example would be `AggregateService.toTest()` should fail due to duplicate annotations at the same level.

```
@ClientRegistrationId("a")
interface AService {}

@ClientRegistrationId("b")
interface BService {}

interface AggregateService extends AService, BService {
	void toTest();
}
```

oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/client/ClientRegistrationIdProcessor.java
bbudano: This change occurred while formatting and wasn't intentional , I will revert it.

oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/web/client/ClientRegistrationIdProcessorTests.java
bbudano: I will revert these changes.

rwinch:
Thanks for the PR! I pushed an example of using `@ClientRegistrationId` at the type level. Once the build passes, this will automatically be merged into main

bbudano:
@rwinch Cool, thanks for the opportunity!

PR #17859, opened 2025-09-09
Refine SimpleGrantedAuthority

quaff:
1. Deprecate field `role` in favor of `authority` since role is specific authority.
2. Add Javadoc to state that role should be prefixed.

### DIFF
--- a/core/src/main/java/org/springframework/security/core/authority/SimpleGrantedAuthority.java
+++ b/core/src/main/java/org/springframework/security/core/authority/SimpleGrantedAuthority.java
@@ -27,21 +27,31 @@ import org.springframework.util.Assert;
  * {@link org.springframework.security.core.Authentication Authentication} object.
  *
  * @author Luke Taylor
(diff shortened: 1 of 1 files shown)


rwinch:
Can you elaborate on why you want to make this change? Given that role is internal, I don't see a lot of value making a breaking change.

quaff:
> Can you elaborate on why you want to make this change? Given that role is internal, I don't see a lot of value making a breaking change.

I'm confused by the name while reading the source, I force pushed a commit, it's not a breaking change now.
I'm OK to revert the code changes, but I think it's valuable to improve Javadoc to elaborate why the name must be `role` not `authority`, and if the authority is a role it should be prefixed.

rwinch:
We can change the constructor argument name & add documentation, but the private member variable name is not public and I'd prefer to keep that the same to avoid breaking our users unnecessarily.

I'm closing this PR, but feel free to create a new one as outlined above.

PR #17865, opened 2025-09-10
Get rid of deprecated JdbcDaoSupport

quaff:


rwinch:
Thank you for the pull request. However, I think that we want to deprecate those methods as well. We won't remove the methods until `JdbcDaoSupport` is removed, so I'd prefer to leave this until Framework removes the class. I'm going to close this pull request.

If you are looking for a pull request, Spring Security should have alternative ways to inject the `DataSource` or `JdbcOperations` that are not deprecated (most likely a constructor argument) and is missing some of these. `JdbcDaoImpl` does not have a way to inject either, so you can add a two new constructors for users to leverage without getting a deprecation warning. You would then deprecate the default constructor for removal. `JdbcUserDetailsManager` has a `DataSource` constructor, but you can add a constructor that accepts `JdbcOperations` and deprecate the default constructor.

quaff:
If breaking changes is acceptable, I think it's better to drop default constructor and setter injection methods:
```java
	private final JdbcOperations jdbcOperations;

	public JdbcDaoImpl(JdbcOperations jdbcOperations) {
		this.jdbcOperations = jdbcOperations;
	}

	public JdbcDaoImpl(DataSource dataSource) {
		this.jdbcOperations = new JdbcTemplate(dataSource);
	}
```

Issue #17781, opened 2025-08-20
Deprecate Authorization Logic that uses FilterInvocation


rwinch:
We should deprecate authorization logic that uses `FilterInvocation` in favor of using `RequestAuthorizationContext`. A few examples:

* `SecurityExpressionHandler<FilterInvocation>` in favor of `SecurityExpressionHandler<RequestAuthorizationContext>`
* `FilterInvocationExpressionRoot` (might change to `HttpServletRequestExpressionRoot`) in favor of `WebSecurityExpressionRoot`
* `DefaultWebSecurityExpressionHandler` in favor of `DefaultHttpSecurityExpressionHandler`

Related gh-17673

therepanic:
Hi, @rwinch. Can you assign me to this issue?

rwinch:
Thanks @therepanic! Yes please go ahead and do it. Please ensure to remove the deprecated usage within Spring Security in a passive way (e.g. adapt classes that use `SecurityExpressionHandler<FilterInvocation>` to use `SecurityExpressionHandler<RequestAuthorizationContext>`)

PR #17884, opened 2025-09-14
Deprecate Authorization Logic that uses `FilterInvocation`

therepanic:
Closes: gh-17781

### DIFF
--- a/access/src/main/java/org/springframework/security/web/access/expression/DefaultWebSecurityExpressionHandler.java
+++ b/access/src/main/java/org/springframework/security/web/access/expression/DefaultWebSecurityExpressionHandler.java
@@ -32,7 +32,9 @@ import org.springframework.security.web.FilterInvocation;
  * @author Eddú Meléndez
  * @author Steve Riesenberg
  * @since 3.0
+ * @deprecated in favor of {@link DefaultHttpSecurityExpressionHandler}
  */
+@Deprecated(since = "7.0.0")
 public class DefaultWebSecurityExpressionHandler extends AbstractSecurityExpressionHandler<FilterInvocation>
 		implements SecurityExpressionHandler<FilterInvocation> {
 
--- a/web/src/main/java/org/springframework/security/web/access/expression/FilterInvocationExpressionRoot.java
+++ b/web/src/main/java/org/springframework/security/web/access/expression/FilterInvocationExpressionRoot.java
@@ -28,7 +28,9 @@ import org.springframework.security.web.util.matcher.IpAddressMatcher;
 /**
  * @author Steve Riesenberg
  * @since 7.0
+ * @deprecated in favor of {@link WebSecurityExpressionRoot}
  */
+@Deprecated(since = "7.0.0")
 final class FilterInvocationExpressionRoot extends SecurityExpressionRoot<FilterInvocation> {
 
 	/**


therepanic:
However, I believe that we can also replace `SecurityExpressionHandler<FilterInvocation>` in favor of `SecurityExpressionHandler<RequestAuthorizationContext>` in the `AbstractAuthorizeTag` and `JspAuthorizeTag` classes.

However, there is a nuance: in `FilterInvocation`, if we pass request parameters, a `DummyRequest` is created that inherits from `HttpServletRequestWrapper`. We could introduce the same class and pass it to `RequestAuthorizationContext`. What do you think about this? Where should we place such a class? I think in our case it would be enough to copy it from `FilterInvocation` directly into `AbstractAuthorizeTag`.

jzheaux:
Hi, @therepanic, thanks for the PR. Can you clarify where it is that the dummy request is created, as far as JSP Tag usage is concerned? I imagine that an evaluating tag would have access to the original `HttpServletRequest` and not need a wrapper dummy request.

I agree that this is used in the default `WebInvocationPrivilegeEvaluator`; however, I believe that is the extent of it.

therepanic:
Thanks for your reply, @jzheaux. Yes, I was probably wrong.

I've thought about it now, and I think we can't replace `SecurityExpressionHandler<FilterInvocation>` in favor of `SecurityExpressionHandler<RequestAuthorizationContext>` in the `AbstractAuthorizeTag and JspAuthorizeTag`.

If we do that, it will be a breaking change. We can see a tests like this:
```java
    @Test
    @SuppressWarnings(“rawtypes”)
    public void expressionFromChildContext() throws IOException {
        SecurityContextHolder.getContext().setAuthentication(new TestingAuthenticationToken(“user”, ‘pass’, “USER”));
		DefaultWebSecurityExpressionHandler expected = new DefaultWebSecurityExpressionHandler();
...

```
In order for this to work, I believe we need to replace DefaultWebSecurityExpressionHandler in the test, but this would again be a breaking change. If we cannot replace them, do we need to deprecate `JspAuthorizeTag` and `AbstractAuthorizeTag`?

WDYT?

jzheaux:
@therepanic, good question. This issue is also addressed in [`WebSecurity`](https://github.com/spring-projects/spring-security/blob/main/config/src/main/java/org/springframework/security/config/annotation/web/builders/WebSecurity.java#L259). You can see the `SecurityExpressionHandler` adapter there.

Let's consider having it continue to look for `SecurityExpressionHandler<FilterInvocation>` and also falling back to `DefaultHttpSecurityExpressionHandler`, adapting it to `SecurityExpressionHandler<FilterInvocation>`. This will match the behavior in `WebSecurity`.

In a separate ticket, we can look at `WebSecurity`, `AbstractAuthorizeTag`, and other places that look for a `SecurityExpressionHandler<FilterInvocation>` and enhancing them to also look for a `SecurityExpressionHandler<RequestAuthorizationContext>`.

How do you feel about that plan?

therepanic:
Hi, @jzheaux. I think that's a great plan and will be better. Avoiding the breaking change while enabling the deprecation path is the right approach.

I suppose we need to create an adapter something like this? As I understand it, you mean searching for the `RequestAuthorizationContext` and adapting it in `FilterInvocation`? For example, in the `AbstractAuthorizeTag#getExpressionHandler` method.
```java
class RequestAuthorizationContextToFilterInvocationExpressionHandlerAdapter
implements SecurityExpressionHandler<FilterInvocation> {
```

Issue #17900, opened 2025-09-15
Support Automatically Checking for Required Authorities in Authorization Rules


jzheaux:
It would be nice if `DefaultAuthorizationManagerFactory` could apply authorization rules automatically, like for certain authorities that are always required.

```java
@Bean 
AuthorizationManagerFactory<Object> authorizationManagerFactory() {
    return DefaultAuthorizationManagerFactory.withAuthorities("FACTOR_PASSWORD", "FACTOR_X509");
}
```

These would then be applied to all authorization managers relating to authenticated users. That is, `permitAll`, `denyAll`, and `anonymous` are not affected.

Note, given https://github.com/spring-projects/spring-security/issues/17932, I've updated the suggested static factory method to avoid a collision.

therepanic:
Hi, @jzheaux. May I attempt to resolve this issue?

rwinch:
@therepanic Thank you for volunteering. The issue is yours

I think that this can be done now by having `DefaultAuthorizationManagerFactory` combine `AllAuthoritiesAuthorizationManager` and the existing `AuthorizationManager` using `AuthorizationManagers.allOf(new AuthorizationDecision(false), AuthorizationManager...)`

rwinch:
@therepanic Sorry I realized that it was ready to go as I wasn't thinking correctly. Please see update above

PR #17942, opened 2025-09-21
Support Automatically Checking for Required Authorities in Authorization Rules

therepanic:
In order to achieve the goal of the update, I wrap every method related to authorized users in `withRequiredAuthorities`, which looks at the `requiredAuthorities` we specified.

Closes: gh-17900


rwinch:
@therepanic Thanks for the pull request. I refactored the code to:

- Support more than requiring additional authorities by allowing the `AuthorizationManager` to be injected instead of just additional authorities. This will allow for more complex scenarios that might want to lookup in a database if the current user is configured to require MFA.
- A Builder to simplify configuring static additional authorities. This will also allow adding another method that supports adding rules where the factor might need to have happened within a certain time period (e.g. require password authentication within 10 minutes, otherwise reprompt the user)

I've marked this to be automatically merged once the build completes. Thanks again!

Issue #17963, opened 2025-09-25
Create SqlAllAuthoritiesAuthorizationManager


rwinch:
We should create an `AuthorizationManager` that can lookup authorities using a configured SQL statement given the username of the currently authenticated user. If any values are returned, then it returns those authorities otherwise it grants access to the user.

Pseudo code looks like:

```java

List<String> additionalAuthorities = findAdditionalAuthorities(authentication.get().getName()); // lookup from SQL statement
if (additionalAuthorities.isEmpty() {
    return new AccessDecision(true);
}
else {
    return AllAuthoritiesAuthorizationManager.hasAllAuthorities(additionalAuthorities).authorize(authentication, object);
}

```

NOTE: `findAdditionalAuthorities` should use a mapper object of sorts that can swap out. It can return some constants if the SQL returns true or it can lookup the additional authorities in SQL.

This can then be used for situations where it is a user preference to enable multi-factor authentication. We should also update the documentation to show how to use it. Something like this: 

```java
AuthorizationManager<Object> additionalAuthorization =  SqlAllAuthoritiesAuthorizationManager
    .whenTrue("select true from users where mfa_enabled = true and username = :username")
    .additionalAuthorities("FACTOR_PASSWORD", "FACTOR_OTT")
    .dataSource(dataSource)
    .build()

return new DefaultAuthorizationManagerFactory.build()
    .additionalAuthorization(additionalAuthorization)
    .build();
```

an additional option for users that gets the authorities from SQL:

```java
AuthorizationManager<Object> additionalAuthorization =  SqlAllAuthoritiesAuthorizationManager
    .selectAuthorities("select true from additional_authorities where username = :username")
    .dataSource(dataSource)
    .build()

return new DefaultAuthorizationManagerFactory.build()
    .additionalAuthorization(additionalAuthorization)
    .build();
```

rwinch:
**NOTE**: If you want this issue, know that it is fairly urgent (we want to get it in for RC1). Also please ensure to claim it (by commenting here) after validating that it has not been assigned or otherwise claimed (via a comment) by someone else before working on it.

therepanic:
Hi, @rwinch! Can I work on it? Also, could you please explain how long RC1 lasts and within what timeframe it should be completed?

rwinch:
@therepanic Thank you for volunteering. The issue is yours.

[...]

PR #17975, opened 2025-09-30
Create SqlAllAuthoritiesAuthorizationManager

therepanic:
Closes: gh-17963


therepanic:
I believe that the current implementation is early and we can still change it, but I have provided an example of how it might look.

By the way, I have a couple of questions.

In https://github.com/spring-projects/spring-security/issues/17963, what was meant by:


> NOTE: findAdditionalAuthorities should use a mapper object of sorts that can swap out. It can return some constants if the SQL returns true or it can look up the additional authorities in SQL.



In the current implementation, we use `whenTrueMode` in `findAdditionalAuthorities` as a signal for how to process the SQL provided to us. How would this look more correctly with a mapper object?

Also, how should we do the tests? Are there already existing tests in spring-security where we test the class that uses `DataSource`? Should we do an integration test?

Issue #17999, opened 2025-10-03
`GrantedAuthority.getAuthority()` should be marked `@Nullable`


rwinch:
The javadoc of `GrantedAuthority.getAuthority()` states that the return value an be `null`, but it is not marked as `@Nullable`. We should update it to be marked as `@Nullable` and update the build to account for this.

therepanic:
Hi, @rwinch. Can I work on this?

rwinch:
Thank you for volunteering @therepanic! The issue is yours

PR #18014, opened 2025-10-06
Mark `GrantedAuthority#getAuthority` as `@Nullable`

therepanic:
Since the `GrantedAuthority#getAuthority` contract itself implies that it can return null, we should mark it as `@Nullable`.

Closes: gh-17999


therepanic:
Currently, the build crashes in the `docs` module, and it's not entirely clear why. Can you explain please why the build crashes in the following class test?
```
ObtainingMoreAuthorizationTests. profileWhenMissingAuthorityConfigurationThenRedirectsToAuthorizationServer()
ObtainingMoreAuthorizationTests. profileWhenMissingAuthorityConfigurationThenRedirectsToAuthorizationServer()
```

**Note: Now it is fixed**

web/src/main/java/org/springframework/security/web/access/DelegatingMissingAuthorityAccessDeniedHandler.java
```diff
@@ -247,17 +247,17 @@ public DelegatingMissingAuthorityAccessDeniedHandler build() {
...
 	private static final class AuthorityRequiredFactorErrorEntry {
 
-		private final String authority;
+		@Nullable private final String authority;
 
 		private final @Nullable RequiredFactorError error;
 
-		private AuthorityRequiredFactorErrorEntry(String authority, @Nullable RequiredFactorError error) {
+		private AuthorityRequiredFactorErrorEntry(@Nullable String authority, @Nullable RequiredFactorError error) {
 			Assert.notNull(authority, "authority cannot be null");
```
ronodhirSoumik, line 260: this contradicts with the `@Nullable` use

therepanic: Good catch, @ronodhirSoumik! Thank you.

web/src/main/java/org/springframework/security/web/access/DelegatingMissingAuthorityAccessDeniedHandler.java
```diff
@@ -160,12 +160,12 @@ private List<AuthorityRequiredFactorErrorEntry> authorityErrors(AccessDeniedExce
...
 				.map((grantedAuthority) -> {
 					String authority = grantedAuthority.getAuthority();
-					if (authority.startsWith("FACTOR_")) {
+					if (authority != null && authority.startsWith("FACTOR_")) {
 						RequiredFactor required = RequiredFactor.withAuthority(authority).build();
 						return new AuthorityRequiredFactorErrorEntry(authority, RequiredFactorError.createMissing(required));
 					}
 					else {
-						return new AuthorityRequiredFactorErrorEntry(authority, null);
+						return new AuthorityRequiredFactorErrorEntry(null, null);
```
ronodhirSoumik: I think can keep `return new AuthorityRequiredFactorErrorEntry(authority, null);`

therepanic: Another good catch! After this fix, all the problems disappeared and the build started to run successfully!

core/src/main/java/org/springframework/security/core/GrantedAuthority.java
```diff
@@ -46,6 +48,6 @@ public interface GrantedAuthority extends Serializable {
 	 * granted authority cannot be expressed as a <code>String</code> with sufficient
 	 * precision).
 	 */
-	String getAuthority();
+	@Nullable String getAuthority();
```
ronodhirSoumik, line 51: Question :: though this implementation supports null case, but it is supposed to be avoided (as per the documentation -> **returning null should be avoided unless actually required.** ) So dont we losing the feature/control with the overall changes?
@therepanic @rwinch

jzheaux: @ronodhirSoumik, good question. The documentation also says that `null` "should be returned" in cases where it cannot be expressed with sufficient precision. In other words, returning null is recommended when needed.

web/src/main/java/org/springframework/security/web/access/DelegatingMissingAuthorityAccessDeniedHandler.java
```diff
@@ -160,7 +160,7 @@ private List<AuthorityRequiredFactorErrorEntry> authorityErrors(AccessDeniedExce
 			return authorityDecision.getAuthorities().stream()
 				.map((grantedAuthority) -> {
 					String authority = grantedAuthority.getAuthority();
-					if (authority.startsWith("FACTOR_")) {
+					if (authority != null && authority.startsWith("FACTOR_")) {
```
jzheaux: If the authority is `null`, then that means that `AuthorityRequiredFactorErrorEntry` has neither an authority nor a `RequiredFactorError`. In those cases, we don't have enough information to report an error, so I'd recommend instead filtering out null authorities from the result, e.g. `filter((ga) -> ga.getAuthority() != null)`

web/src/main/java/org/springframework/security/web/access/DelegatingMissingAuthorityAccessDeniedHandler.java
```diff
@@ -247,17 +247,16 @@ public DelegatingMissingAuthorityAccessDeniedHandler build() {
...
 	private static final class AuthorityRequiredFactorErrorEntry {
 
-		private final String authority;
+		@Nullable private final String authority;
 
 		private final @Nullable RequiredFactorError error;
 
-		private AuthorityRequiredFactorErrorEntry(String authority, @Nullable RequiredFactorError error) {
-			Assert.notNull(authority, "authority cannot be null");
+		private AuthorityRequiredFactorErrorEntry(@Nullable String authority, @Nullable RequiredFactorError error) {
```
jzheaux: If both are null, then this object doesn't provide any error information. Let's continue to require that the authority be non-null (please see earlier comment).

web/src/main/java/org/springframework/security/web/access/DelegatingMissingAuthorityAccessDeniedHandler.java
therepanic: Good point, then we won't have to mark `authority` as `@Nullable` in `AuthorityRequiredFactorErrorEntry`.

therepanic:
Thanks for the review, @jzheaux.

I agree with everything. It was a bad idea to mark authority as `@Nullable` when there was a simpler way to do it. I implemented this change as a separate commit to make it easier to read the changes.

I don't think the build crashes because of the changes in the PR.

jzheaux (approved):

jzheaux:
Thanks, @therepanic! This is now merged into `main`

Issue #17902, opened 2025-09-16
InitializeUserDetailsBeanManagerConfigurer should respect primary flag


quaff:
**Expected Behavior**
If one of multiple `UserDetailsService` beans is primary, then it should be configured.

**Current Behavior**
Warning log message instead.

https://github.com/spring-projects/spring-security/blob/c905ac359dde44cb35bca688444438c2b8fe6beb/config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/InitializeUserDetailsBeanManagerConfigurer.java#L86-L92

**Context**

I want to enhance existing `UserDetailsService` like this:

```java
    @Primary
    @Bean
    public UserDetailsService primaryUserDetailsService(UserDetailsService userDetailsService) {
        return (username) -> {
            // TODO load user from other source
            return userDetailsService.loadUserByUsername(username);
        };
    }
```

I'd like to contribute if it's accepted.

PR #18015, opened 2025-10-06
Honor @Primary for UserDetailsService and UserDetailsPasswordService in InitializeUserDetailsBeanManagerConfigurer

siva-sai-udaygiri:
### Motivation
In scenarios where multiple `UserDetailsService` or `UserDetailsPasswordService` beans
are defined, Spring may raise `NoUniqueBeanDefinitionException` or select the wrong
bean without explicit resolution.

### Changes
- Added `@Primary` to `UserDetailsService` and `UserDetailsPasswordService` beans in
  `InitializeUserDetailsBeanManagerConfigurer`
- Ensures deterministic bean selection

### Issue Reference
Fixes gh-17902

### Testing
- Added a test in `InitializeUserDetailsBeanManagerConfigurerTests` to validate
  resolution when multiple candidates are present
- Verified that all existing tests pass with `./gradlew check`

### Notes
- No breaking changes
- Consistent with Spring idioms for bean resolution

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/InitializeUserDetailsBeanManagerConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/InitializeUserDetailsBeanManagerConfigurer.java
@@ -21,6 +21,7 @@ import java.util.Arrays;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
+import org.springframework.beans.BeansException;
 import org.springframework.context.ApplicationContext;
 import org.springframework.core.Ordered;
 import org.springframework.core.annotation.Order;
@@ -33,9 +34,12 @@ import org.springframework.security.core.userdetails.UserDetailsService;
 import org.springframework.security.crypto.password.PasswordEncoder;
 
 /**
- * Lazily initializes the global authentication with a {@link UserDetailsService} if it is
- * not yet configured and there is only a single Bean of that type. Optionally, if a
- * {@link PasswordEncoder} is defined will wire this up too.
+ * Lazily initializes the global authentication with a {@link UserDetailsService}. If
+ * multiple beans of that type exist, the container's autowire rules are used to select a
+ * single candidate (e.g. {@code @Primary}). If no single candidate can be resolved, the
+ * configurer logs a warning and does not auto-wire. Optionally wires a
+ * {@link PasswordEncoder}, {@link UserDetailsPasswordService}, and
+ * {@link CompromisedPasswordChecker} when available.
  *
  * @author Rob Winch
  * @author Ngoc Nhan
@@ -48,9 +52,6 @@ class InitializeUserDetailsBeanManagerConfigurer extends GlobalAuthenticationCon
 
 	private final ApplicationContext context;
 
-	/**
-	 * @param context
-	 */
 	InitializeUserDetailsBeanManagerConfigurer(ApplicationContext context) {
 		this.context = context;
 	}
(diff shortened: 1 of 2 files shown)


config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/InitializeUserDetailsBeanManagerConfigurer.java
```diff
@@ -83,18 +85,27 @@ public void configure(AuthenticationManagerBuilder auth) {
...
 			}
-			UserDetailsService userDetailsService = InitializeUserDetailsBeanManagerConfigurer.this.context
-				.getBean(beanNames[0], UserDetailsService.class);
-			PasswordEncoder passwordEncoder = getBeanOrNull(PasswordEncoder.class);
-			UserDetailsPasswordService passwordManager = getBeanOrNull(UserDetailsPasswordService.class);
-			CompromisedPasswordChecker passwordChecker = getBeanOrNull(CompromisedPasswordChecker.class);
+
+			PasswordEncoder passwordEncoder = getBeanIfUnique(PasswordEncoder.class);
+			// Also resolve UDPS via container so @Primary is honored
+			UserDetailsPasswordService passwordManager = getAutowireCandidateOrNull(UserDetailsPasswordService.class);
```
jzheaux: I believe this changes the existing semantics by calling `getIfAvailable` instead of `getIfUnique`. The `ObjectProvider` JavaDoc implies that it accounts for `@Primary`:

```java
	/**
	 * Return an instance (possibly shared or independent) of the object
	 * managed by this factory.
	 * @return an instance of the bean, or {@code null} if not available or
	 * not unique (i.e. multiple candidates found **with none marked as primary**)
	 * @throws BeansException in case of creation errors
	 * @see #getObject()
	 */
```

Can you confirm with unit tests that changing to `getIfAvailable` is necessary to support `@Primary`?

jzheaux (changes requested):
Thanks, @siva-sai-udaygirl, for the PR! I've left some feedback inline.

config/src/main/java/org/springframework/security/config/annotation/authentication/configuration/InitializeUserDetailsBeanManagerConfigurer.java
siva-sai-udaygiri: Thanks, @jzheaux! I added unit tests to show why we need to rely on the container’s autowire-candidate resolution to honor @Primary.

• With getIfUnique(...) when there are two UserDetailsService beans and one is @Primary, it still returns null (multiple names), so the global AuthenticationManager isn’t configured.
• With getIfAvailable(...) it delegates to autowire-candidate selection (same as normal injection), so the @Primary bean is chosen, and the AuthenticationManager is configured.

New tests:
- whenMultipleUdsAndOneResolvableCandidate_thenPrimaryIsAutoWired
- whenMultipleUdsAndNoSingleCandidate_thenSkipAutoWiring

Location: spring-security-config/src/test/java/org/springframework/security/config/annotation/authentication/configuration/InitializeUserDetailsBeanManagerConfigurerTests.java

This keeps behavior the same for the single-bean case, and for multiple beans without a single resolvable candidate we still skip auto-wiring (and log). Please take another look and let me know if you’d prefer a different approach.

siva-sai-udaygiri:
Implemented getAutowireCandidateOrNull(...) to honor @Primary.
Added unit tests covering: (a) multiple UDS with one @Primary → auto-wired; (b) multiple UDS with no single candidate → skipped; (c) single UDS unchanged.
Rebased on latest main and all config checks pass locally.

PR #18026, opened 2025-10-10
Correct the org.springframework.security.config.annotation.web.LogoutDsl's property description

SimonVonXCVII:
1. Correct the org.springframework.security.config.annotation.web.LogoutDsl's property description


jzheaux:
Good catch, @SimonVonXCVII! This is now merged into `main`

PR #18035, opened 2025-10-11
Fix-typos

ngocnhan-tran1996:


rwinch:
Thanks for the changes. I pushed a change to deprecate the misspelling of `CacheControlServerHttpHeadersWriter.CACHE_CONTRTOL_VALUE` rather than delete it since it is a public variable.

PR #18050, opened 2025-10-14
Enhancements to Authentication#toBuilder

jzheaux:
This PR moves `Authentication#toBuilder` to a new sub-interface, `BuildableAuthentication`. This will improve passivity in the Security 7.0 release for applications that are using custom authentication implementations and relying on those concrete types.

Specifically, instead of doing:

```java
Authentication merged = result.toBuilder().authorities((a) -> a.addAll(current.getAuthorities())).build();
```

You can do:

```java
if (result instanceof BuildableAuthentication buildable) {
    Authentication merged = buildable.toBuilder().authorities((a) -> a.addAll(current.getAuthorities())).build();
}
```

Additionally, the PR adds `authentication` to `BuilderAuthentication.Builder` for easily application of common merging logic:

```java
if (result instanceof BuildableAuthentication buildable) {
    Authentication merged = buildable.toBuilder().authentication(current).build();
}
```

### DIFF
--- a/core/src/main/java/org/springframework/security/core/BuildableAuthentication.java
+++ b/core/src/main/java/org/springframework/security/core/BuildableAuthentication.java
@@ -16,42 +16,11 @@
 
 package org.springframework.security.core;
 
-import java.util.Collection;
-import java.util.Set;
-import java.util.function.Consumer;
-import java.util.stream.Collectors;
-
-import org.jspecify.annotations.Nullable;
-
-import org.springframework.security.authentication.AuthenticationManager;
-import org.springframework.security.core.context.SecurityContextHolder;
-
 /**
- * Represents the token for an authentication request or for an authenticated principal
- * once the request has been processed by the
- * {@link AuthenticationManager#authenticate(Authentication)} method.
- * <p>
- * Once the request has been authenticated, the <tt>Authentication</tt> will usually be
- * stored in a thread-local <tt>SecurityContext</tt> managed by the
- * {@link SecurityContextHolder} by the authentication mechanism which is being used. An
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/springframework/security/core/BuildableAuthentication.java
```diff
@@ -0,0 +1,186 @@
...
+	 */
+	Builder<?> toBuilder();
+
+	/**
+	 * A builder based on a given {@link BuildableAuthentication} instance
+	 *
+	 * @author Josh Cummings
+	 * @since 7.0
+	 */
+	interface Builder<B extends Builder<B>> {
```
rwinch: I think that it makes more sense to keep `Builder` in `Authentication` so that classes can have a static builder method independent of implementing `BuildableAuthentication`.

oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/authentication/BearerTokenAuthenticationFilter.java
```diff
@@ -182,9 +183,9 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
 			}
 			Authentication current = this.securityContextHolderStrategy.getContext().getAuthentication();
 			if (current != null && current.isAuthenticated()) {
-				authenticationResult = authenticationResult.toBuilder()
```
rwinch, line 185: Nice! Looks like you caught a spot that was missing the checks for duplicate authorities

PR #18062, opened 2025-10-17
Polish SimpleGrantedAuthority

quaff:
1. Add Javadoc to state that role is prefixed.
2. Rename constructor argument from `role` to `authority` for better readability.

See https://github.com/spring-projects/spring-security/pull/17859#issuecomment-3412420577


jzheaux:
Thanks, @quaff, for this polish. It's now merged into `main`.

Issue #18069, opened 2025-10-17
RegexRequestMatcher factory methods and their documentation are not consistent


edudar-chwy:
`RegexRequestMatcher.regexMatcher(HttpMethod method, String pattern)` says in the Javadoc that `method` "May be null to match all methods." which makes sense as `RegexRequestMatcher` constructor takes null, but then we have `Assert.notNull(method, "method cannot be null")` as the first line in the method, which is completely the opposite of what is expected.

jzheaux:
Thanks for the suggestion, @edudar-chwy, I'll close this in favor of #18076. Feel free to take a look at it and consider whether it meets your needs.

PR #18076, opened 2025-10-19
Fixed nullity discreapency between documenation and assert check in R…

ronodhirSoumik:
Related to #18069

### DIFF
--- a/web/src/main/java/org/springframework/security/web/util/matcher/RegexRequestMatcher.java
+++ b/web/src/main/java/org/springframework/security/web/util/matcher/RegexRequestMatcher.java
@@ -81,7 +81,6 @@ public final class RegexRequestMatcher implements RequestMatcher {
 	 * @since 5.8
 	 */
 	public static RegexRequestMatcher regexMatcher(HttpMethod method, String pattern) {
-		Assert.notNull(method, "method cannot be null");
 		Assert.hasText(pattern, "pattern cannot be empty");
 		return new RegexRequestMatcher(pattern, method.name());
 	}


web/src/main/java/org/springframework/security/web/util/matcher/RegexRequestMatcher.java
```diff
@@ -76,7 +76,7 @@ public static RegexRequestMatcher regexMatcher(HttpMethod method) {
 
 	/**
 	 * Creates a case-sensitive {@code Pattern} instance to match against the request.
-	 * @param method the HTTP method to match. May be null to match all methods.
+	 * @param method the HTTP method to match. May not be null to match all methods.
```
jzheaux: > May not be null to match all methods.

This sentence now sounds a little cryptic. Can we simply leave it out? Folks can call another static method if they only care about the URI.

jzheaux (changes requested):
Hi, @ronodhirSoumik! Thanks for this cleanup. I've left some feedback inline.

jzheaux:
Also, @ronodhirSoumik, will you please reference #18069 in your commit message. Something like this:

```
Correct JavaDoc about HttpMethod Parameter

Closes gh-18069
```

web/src/main/java/org/springframework/security/web/util/matcher/RegexRequestMatcher.java
edudar-chwy: While Javadoc can be improved, I don't see a good reason to restrict input value here artificiallly. Constructor already takes care of the `null` case; throughout the class `httpMethod` is assumed to be nullable. Just let it be `null`. A sibling `PathPatternRequestMatcher`, for example,  allows `method` as `null`, explicitly marking it as `@Nullable`. While it would be good to get rid of nullable all together, `HttpMethod` does not have an "any/all" option.

Ideally, a builder similar to one in `PathPatternRequestMatcher` would do a good job here. Meanwhile, I'm just using constructor, because I get the `method` as an input myself, so checking `null` and deciding between faactory methods is much more verbose then just calling a constructor.

ronodhirSoumik: I am undoing the change in javadoc and removing the` Assert.notNull(method, "method cannot be null");` - hope that do as a minimal change

ronodhirSoumik:
Hi @jzheaux, is this issue #18069  already fixed? Should I close this PR?

ronodhirSoumik:
Hi @jzheaux @edudar-chwy, can you please review whether this PR is all good?

ronodhirSoumik:
hi @jzheaux can you please review whether this PR is good to go now?

Issue #18018, opened 2025-10-08
Add possibility to customize refresh token and DPoP binding


makcpopTwo:
**Expected Behavior**

It should be possible to customize refresh token and DPoP binding, so engineers can use any binding scheme.

**Current Behavior**

Currently, it is not possible to customize this. Only one scheme is supported: [binding through access_token](https://github.com/spring-projects/spring-security/blob/469ed096459c6015a38d0849f5a3f53f9d1cf393/oauth2/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2RefreshTokenAuthenticationProvider.java#L167-L173)

**Context**

The [DPoP RFC](https://datatracker.ietf.org/doc/html/rfc9449#section-5-8) is not specify how to bind refresh token and DPoP [link](https://datatracker.ietf.org/doc/html/rfc9449#section-5-8) does not specify how to bind the refresh token to DPoP; instead, it leaves this decision to the authorization server.

Currently, I have an authorization server (not using the latest Spring Security version) with custom DPoP logic implemented. In my implementation, I store the DPoP thumbprint inside the refresh token, so I don't need to use the access token for the binding.

It would be very useful to be able to customize the binding (validation) process between DPoP and refresh tokens.

jgrandja:
@makcpopTwo 

> It would be very useful to be able to customize the binding (validation) process between DPoP and refresh tokens.

Can you provide an example configuration that would allow you to customize the validation?

makcpopTwo:
@jgrandja 
Sorry for the delayed response.

I’ve created a POC — this is not a final implementation, just a conceptual demonstration.
I hope I correctly understood how Spring Security handles this flow.

1) Add a new interface for extracting the DPoP JWK thumbprint from the authorization
```
public interface JwkThumbprintClaimRefreshTokenExtractor {

	String extract(OAuth2Authorization authorization);
}
```

2) Modify [OAuth2RefreshTokenAuthenticationProvider](https://github.com/spring-projects/spring-security/blob/fc795a81d43c08b0e75570eb7da9004e2d9bd311/oauth2/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2RefreshTokenAuthenticationProvider.java#L167)
Add a JwkThumbprintClaimRefreshTokenExtractor field with a default implementation, allowing the behavior to be overridden if needed.
The default implementation keeps the same logic as the current one.
```
[...]

PR #18098, opened 2025-10-22
Add request validator for refresh token grant

therepanic:
Recently, I needed to use custom logic for validation, but this was not possible without making changes to the current code. Here, I presented an interface strategy that can be replaced with a custom one via a setter if desired. This is described here: https://github.com/spring-projects/spring-security/issues/18266, and also here https://github.com/spring-projects/spring-security/issues/18018#issuecomment-3431835851.

### DIFF
--- a/oauth2/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2RefreshTokenAuthenticationContext.java
+++ b/oauth2/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2RefreshTokenAuthenticationContext.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2004-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


jgrandja:
Thanks for the PR @therepanic. I'll review this after we release `7.0`. In the meantime, when you have a moment, please add tests. Thank you.

makcpopTwo:
Copying my thoughts from: https://github.com/spring-projects/spring-security/issues/18018

My case: I have custom DPoP and refresh token mapping, so I need the ability to override the extraction of the thumbprint from the refresh token.

Currently, I would need to duplicate some common parts of the DPoP validation in my own code (retrieving the DPoP from the context/authentication, verifying DPoP and ClientAuthenticationMethod, parsing the JWT, computing the thumbprint, comparing thumbprints). All of these are quite standard procedures defined by the RFC, and I’d prefer not to duplicate this logic in my validator — I only need to specify how the thumbprint should be extracted from the authentication (i.e., the refresh token).

If I completely override the validator, I might miss some validation logic that could be added to it in the future.

If it would be helpful, I can reference the relevant RFC parts to clarify which parts of the code are standardized.

joaquinjsb:
I think that this code isn't yet enough (or should it be customized)?
what happens if I acquired an access token with DPoP? subsequent refresh tokens should  succeed only if it has a valid DPoP, and if no DPoP has been provided, it should fail?

Issue #18092, opened 2025-10-21
JwtTypeValidator is case sensitive in 7.0-m3


wuem:
**Describe the bug**
The bug itself is really simple. Version 6.5 introduces JwtTypeValidator that is meant to replace Nimbus type validators. While making migrations to incoming Spring boot 4.0 I noticed that this validator is case sensitive and previously used Nimbus JOSEObjectTypeVerifier is not. 

**Expected behavior**
 new JwtTypeValidator should be backward compatible with JOSEObjectTypeVerifier

Those two classes are really similar but Spring native implementation uses Strings as type values when JOSEObjectTypeVerifier uses JOSEObjectType with overridden equals that is not case sensitive.

namest504:
I've opened PR #18101 to fix this. I hope this helps!

jzheaux:
Thanks for the PR, @wuem! I'll close this in favor of https://github.com/spring-projects/spring-security/pull/18101

PR #18101, opened 2025-10-24
`typ` values should not be case-sensitive in `JwtTypeValidator`

namest504:
Fixed: #18092

The JwtTypeValidator was performing a case-sensitive check for the JWT typ header, which was a change from the behavior of the previous Nimbus JOSEObjectTypeVerifier.

This commit updates the JwtTypeValidator to be case-insensitive.
New test cases have been added to verify this change.


jzheaux:
It's important that we make sure we are following specs as well as providing a reasonable migration path. Checking [the JWS RFC](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.9), it states (emphasis **mine**):

> Per [RFC 2045](https://datatracker.ietf.org/doc/html/rfc2045) [[RFC2045](https://datatracker.ietf.org/doc/html/rfc2045)], all media type values, subtype values, and parameter names are **case insensitive**.

So I agree that this is a bug and can be fixed in time for the 7.0 release. Thanks for spotting it, @namest504!

jzheaux:
Thanks, @namest504! This is now merged into `6.5.x` (since that's where the bug was introduced) and `main`.

PR #18119, opened 2025-10-31
Update webauthn4j usage, use non-deprecated methods

Kehrlann:
The most important change is in the authentication ceremony, migrating from wa4j `Authenticator` interface to the `CredentialRecord` interface.

We cannot pass all values to `CredentialRecord`, as they are not stored at registration time. However, they are nullable, and were not stored previously. Passing null is functionally equivalent.

Tested with `WebAuthnWebDriverTests`


rwinch:

Issue #10821, opened 2022-02-10
Saml2LogoutConfigurer: Use RequestMatcher from LogoutConfigurer


chschu:
**Expected Behavior**

The SAML Single Logout (initiated by relying party) should use `LogoutConfigurer.getLogoutRequestMatcher(...)`, combined with `Saml2LogoutConfigurer.Saml2RequestMatcher` to detect logout requests.

It would also be great if the `logoutMatcher` could be customized, because `Saml2LogoutConfigurer.Saml2RequestMatcher` is not always desirable.

**Current Behavior**

Currently, only a single logout URL can be configured with `Saml2LogoutConfigurer`, while `LogoutConfigurer` allows full customization of the `RequestMatcher`. If that matcher is not a simple "must be a POST to a single URL", the `LogoutFilter` created by `Saml2LogoutConfigurer` cannot be configured appropriately.

**Context**

I have an application with different logout URLs that redirect the user to different logout success URLs. That requires a complex `RequestMatcher` in `LogoutConfigurer`, and it cannot currently be done in `Saml2LogoutConfigurer`.

Workaround: Configure an `ObjectPostProcessor` on `Saml2LogoutConfigurer` that creates the `LogoutFilter` from scratch.

jzheaux:
@chschu thanks for posting this.

> Workaround: Configure an ObjectPostProcessor on Saml2LogoutConfigurer that creates the LogoutFilter from scratch.

I wonder why this is. Can you not call `setLogoutRequestMatcher` on the `LogoutFilter` passed to your `ObjectPostProcessor`?

> The SAML Single Logout (initiated by relying party) should use LogoutConfigurer.getLogoutRequestMatcher(...), combined with Saml2LogoutConfigurer.Saml2RequestMatcher to detect logout requests.

This makes sense since it seems quite odd that an application would configure `logout` to have a custom URL and then leave `saml2Logout` using `/logout`. It does seem that `logout`'s custom URL should influence `saml2Logout`'s. 

Can you provide a PR along these lines that determines `Saml2LogoutConfigurer`'s request matcher like so:

* If the application provided a `logoutUrl`, use it
* Otherwise, if the application has a `LogoutConfigurer`, use its request matcher combined with `Saml2RequestMatcher`
* Otherwise, default to `POST /logout`

Since this would be a breaking change, I'm still thinking about whether it can go into a 5.x release or whether it needs to wait until 6.x. Either way, a PR is welcome.

[...]

PR #18122, opened 2025-11-01
Add `logoutRequestMatcher` to `Saml2LogoutConfigurer` for custom matching

therepanic:
This solves the problem that previously SAML2 logout could only be triggered from the default POST /logout. Now you can configure any endpoint/method for SAML2 logout using the DSL.

Closes: gh-10821

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LogoutConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LogoutConfigurer.java
@@ -105,6 +105,7 @@ import org.springframework.security.web.util.matcher.RequestMatcher;
  *
  * @author Josh Cummings
  * @author Ngoc Nhan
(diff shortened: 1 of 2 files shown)


therepanic:
Note: the final idea proposed by @jzheaux, described here, was adopted to resolve the issue: https://github.com/spring-projects/spring-security/issues/10821#issuecomment-2576375276

config/src/main/java/org/springframework/security/config/annotation/web/configurers/saml2/Saml2LogoutConfigurer.java
```diff
@@ -127,6 +128,8 @@ public final class Saml2LogoutConfigurer<H extends HttpSecurityBuilder<H>>
 
 	private LogoutResponseConfigurer logoutResponseConfigurer;
 
+	private RequestMatcher logoutRequestMatcher;
+
```
therepanic, line 132: I guess there is no better way than to carry the matcher in the class.

PR #18123, opened 2025-11-02
Fix docs typo X-Requested-By -> X-Requested-With

martinboulais:
## Simple documentation typo fix

The documentation states that setting the header `X-Requested-By` will remove the `WWW-Authenticate` header from the response. 

However, after testing this and reading the library code it looks like the header to set should be `X-Requested-With` (X-Requested-By is mentioned nowhere except in this documentation file), so I propose this simple PR to fix this.


rwinch:
Thanks for the pull request. I've rebased it and it will be merged as soon as the build passes 😄

Issue #18013, opened 2025-10-06
Add hasScope as a valid SpEL expression to PreAuthorize, etc.


kistlers:
**Expected Behavior**

In a similar fashion to the added `hasScope` for authorization rules (added with #13754), I would like to use `hasScope(scope)` in SpEL expressions in `@PreAuthorize`, i.e.:
```java
@PreAuthorize("hasScope('read'")
```

**Current Behavior**

This can already be achieved using `hasAuthority`:
```java
@PreAuthorize("hasAuthority('SCOPE_read'")
```

Evidently, not a big issue. But I'd like to add that `hasRole` also exists:
```java
private String defaultRolePrefix = "ROLE_";

@Override
public final boolean hasAnyRole(String... roles) {
    return hasAnyAuthorityName(this.defaultRolePrefix, roles);
}
```

PR #18151, opened 2025-11-08
Support hasScope in Method Security

ngocnhan-tran1996:
Closes: gh-18013

### DIFF
--- a/docs/modules/ROOT/pages/servlet/oauth2/resource-server/jwt.adoc
+++ b/docs/modules/ROOT/pages/servlet/oauth2/resource-server/jwt.adoc
@@ -949,6 +949,21 @@ fun getMessages(): List<Message> { }
 ----
 ======
 
+[[method-security-has-scope]]
+=== Using `hasScope` in Method Security
+
+Because method security expressions can evaluation `AuthorizationManager` instances, you can also use the `hasScope` API by publishing a `DefaultOAuth2AuthorizationManagerFactory` `@Bean`:
+
+include-code::./MethodSecurityHasScopeConfiguration[tag=declare-factory,indent=0]
+
+and then doing:
+
+include-code::./MessageService[tag=protected-method,indent=0]
+
+If you are using xref:servlet/authentication/mfa.adoc[Spring Security's MFA feature], then you can supply its `AuthorizationManagerFactory` instance to ensure that your authentication factors are automatically checked as well by including it in your `DefaultOAuth2AuthorizationManagerFactory` constructor as follows:
+
+include-code::./MethodSecurityHasScopeMfaConfiguration[tag=declare-factory,indent=0]
+
 [[oauth2resourceserver-jwt-authorization-extraction]]
 === Extracting Authorities Manually
 
--- a/docs/modules/ROOT/pages/servlet/oauth2/resource-server/opaque-token.adoc
+++ b/docs/modules/ROOT/pages/servlet/oauth2/resource-server/opaque-token.adoc
@@ -638,6 +638,21 @@ fun getMessages(): List<Message?> {}
 ----
 ======
 
+[[method-security-has-scope]]
+=== Using `hasScope` in Method Security
+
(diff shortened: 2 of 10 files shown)


jzheaux:
Hi, @ngocnhan-tran1996. We want to be careful about adding to the expression root, especially now that it has implications for `AuthorizationManagerFactory`. Alternatively, we could consider an interface `OAuth2AuthorizationManagerFactory` like this:

```java
public interface OAuth2AuthorizationManagerFactory<T> {
    default AuthorizationManager<T> hasScope(String scope) {
        return OAuth2AuthorizationManagers.hasScope(scope);
    }

    // ...
}
```

And a default implementation:

```java
@Bean 
OAuth2AuthorizationManagerFactory<Object> oauth2() {
    return new DefaultOAuth2AuthorizationManagerFactory();
}
```

That takes an `AuthorizationManagerFactory` as a parameter in support of MFA:

```java
@Bean 
OAuth2AuthorizationManagerFactory<Object> oauth2(AuthorizationManagerFactory<Object> mfa) {
    return new OAuth2AuthorizationManagerFactory(mfa);
}
```

And then do:

```java
@PreAuthorize("@oauth2.hasScope('message:read')")
```

I like this pattern since it allows for other modules to add their own expressions as well, without needing to change or extend `SecurityExpressionRoot`.

ngocnhan-tran1996:
@jzheaux 

I’ve made the requested changes. Let me know if anything else is needed.

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/authorization/DefaultOAuth2AuthorizationManagerFactory.java
```diff
@@ -0,0 +1,79 @@
...
+import org.springframework.security.authorization.AuthorizationManagerFactory;
+import org.springframework.security.authorization.DefaultAuthorizationManagerFactory;
+import org.springframework.util.Assert;
+
+/**
+ * A factory for creating different kinds of {@link AuthorizationManager} instances.
+ *
+ * @param <T> the type of object that the authorization check is being done on
+ * @author Ngoc Nhan
+ * @since 7.0
```
jzheaux: Will you please update this to `7.1` since 7 is already released?

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/authorization/OAuth2AuthorizationManagerFactory.java
```diff
@@ -0,0 +1,73 @@
...
+
+import org.springframework.security.authorization.AuthorizationManager;
+import org.springframework.security.core.Authentication;
+
+/**
+ * A factory for creating different kinds of {@link AuthorizationManager} instances.
+ *
+ * @param <T> the type of object that the authorization check is being done on
+ * @author Ngoc Nhan
+ * @since 7.0
```
jzheaux: Will you please update this to `7.1` since 7 is already released?

oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/authorization/DefaultOAuth2AuthorizationManagerFactory.java
```diff
@@ -0,0 +1,79 @@
...
+	public AuthorizationManager<T> hasAnyScope(String... scopes) {
+		Assert.notNull(scopes, "scopes can not be null");
+		String[] mappedScopes = new String[scopes.length];
+		for (int i = 0; i < scopes.length; i++) {
+			assertScope(scopes[i]);
+			mappedScopes[i] = this.scopePrefix + scopes[i];
+		}
+		return this.authorizationManagerFactory.hasAnyAuthority(mappedScopes);
+	}
+
```
jzheaux, line 82: Will you please add `hasAllScopes`? It can call `AuthorizationManagerFactory#hasAllAuthorities`

jzheaux (approved):

jzheaux:
Thanks for the PR, @ngocnhan-tran1996! This is now merged into `main`

Issue #18144, opened 2025-11-05
Add a Minimal Authorization Server Configuration


rwinch:
The reference currently has two samples. It would be nice if there was an intermediate example that shows a minimal configuration with `HttpSecurity` similar to:

```java
@Bean
SecurityFilterChain springSecurity(HttpSecurity http) {
    http
        .authorizeHttpRequests(requests -> requests
            .anyRequest().authenticated()
        )
        .authorizationServer(auth -> auth
            .oidc(Customizer.withDefaults())
        )
        .formLogin(Customizer.withDefaults());
    return http.build();
}
```

therepanic:
Hi, @rwinch. Can I work on this?

PR #18153, opened 2025-11-09
Add a minimal authorization server configuration

therepanic:
Add minimal authorization server config with `HttpSecurity` example to docs for a simpler setup before advanced customization.

Closes: gh-18144


jgrandja:
Thanks for the PR @therepanic. This is now merged along with a polish commit.

Issue #18158, opened 2025-11-12
WebAuthn login fails when validating `allowCredentials`


Kehrlann:
**Describe the bug**

When used as a single factor, WebAuthn login works.

Used as a second factor, WebAuthn login fails, because `allowCredentials` is not null and the sent `credentialId` does not match the stored `allowCredentials`. The following error is thrown:

```
org.springframework.security.authentication.BadCredentialsException: credentialId not listed in allowCredentials is used.
	at org.springframework.security.web.webauthn.authentication.WebAuthnAuthenticationProvider.authenticate(WebAuthnAuthenticationProvider.java:80) ~[spring-security-webauthn-7.0.0-RC3.jar:7.0.0-RC3]
	at org.springframework.security.authentication.ProviderManager.authenticate(ProviderManager.java:183) ~[spring-security-core-7.0.0-RC3.jar:7.0.0-RC3]
	at org.springframework.security.web.webauthn.authentication.WebAuthnAuthenticationFilter.attemptAuthentication(WebAuthnAuthenticationFilter.java:121) ~[spring-security-webauthn-7.0.0-RC3.jar:7.0.0-RC3]
	at org.springframework.security.web.authentication.AbstractAuthenticationProcessingFilter.doFilter(AbstractAuthenticationProcessingFilter.java:249) ~[spring-security-web-7.0.0-RC3.jar:7.0.0-RC3]
	at org.springframework.security.web.authentication.AbstractAuthenticationProcessingFilter.doFilter(AbstractAuthenticationProcessingFilter.java:239) ~[spring-security-web-7.0.0-RC3.jar:7.0.0-RC3]
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:380) ~[spring-security-web-7.0.0-RC3.jar:7.0.0-RC3]
[...]
Caused by: com.webauthn4j.verifier.exception.NotAllowedCredentialIdException: credentialId not listed in allowCredentials is used.
	at com.webauthn4j.verifier.internal.CredentialIdVerifier.verify(CredentialIdVerifier.java:17) ~[webauthn4j-core-0.29.7.RELEASE.jar:na]
	at com.webauthn4j.verifier.AuthenticationDataVerifier.verify(AuthenticationDataVerifier.java:92) ~[webauthn4j-core-0.29.7.RELEASE.jar:na]
[...]

PR #18163, opened 2025-11-12
Fix webauthn multifactor authentication

Kehrlann:
Fixes gh-18158 and improves webauthn webdriver tests.


rwinch:
Thanks again for your help identifying the issue, providing tests, and a fix for it. 

I pushed a few minor changes:

- The fix now passes the raw id vs the id (which is base64 encoded) into webauthn4j APIs rather than base64 encoding the allowed credentialIds
- I added a unit test to ensure that this is fixed in unit tests and not just the WebDriver tests

PR #18166, opened 2025-11-14
Add Test for ServletRequestPathUtils.parseAndCache(method=null)

ronodhirSoumik:


web/src/main/java/org/springframework/security/web/servlet/util/matcher/PathPatternRequestMatcher.java
```diff
@@ -336,7 +335,8 @@ private static final class HttpMethodRequestMatcher implements RequestMatcher {
 
 		@Override
 		public boolean matches(HttpServletRequest request) {
-			return this.method.name().equals(request.getMethod());
+			String requestMethod = request.getMethod();
+			return requestMethod != null && this.method.name().equals(requestMethod);
```
ngocnhan-tran1996: We don’t need to check for null because `this.method.name()` will return a `String` type and `String#equals(null)` will return false, so the test passes without any changes.

ronodhirSoumik: but this fixes the NPE case - ig that is our expectation of getting false for invalid

ngocnhan-tran1996: `this.method.name().equals(request.getMethod())` will return false if `request.getMethod() == null` and won’t throw an NPE unless request is null or method is null. The test passes without any changes.

Image

rwinch:
Thank you for your submission @ronodhirSoumik 

As pointed out, the test will pass without changes to the code. Additionally, the issue you are linking to is caused by ConcurrentHashMap not supporting a null key (which is of type ServletContext) and is unrelated to the method being null.

I think that the test would still provide value by ensuring that we do not get a NullPointerException later. If you would modify the PR to just have the test and update the PR subject, I can merge it.

web/src/main/java/org/springframework/security/web/servlet/util/matcher/PathPatternRequestMatcher.java
ronodhirSoumik: oh got it - thanks for the explanation

ronodhirSoumik:
@rwinch I have updated the PR with only test file (also unlinked the previous issue number) - plz review

rwinch:

PR #18169, opened 2025-11-15
fix: Allow OidcAuthorizedClientRefreshedEventListener refreshing if authentication subclasses OAuth2AuthenticationToken

mpalourdio:
hi,

This PR considers relaxing the check that is done is `OidcAuthorizedClientRefreshedEventListener`. If for some reason, you want/need to subclass `OAuth2AuthenticationToken` to add custom data to `Principal`, you lose the ability of `OidcAuthorizedClientRefreshedEventListener` to refresh expiration dates.

The doc states that _If the application customizes the authentication result, then a custom event handler should be provided._  **But**

- This class is final
- Not exposed as a @Bean
- Its full instance configuration is handled by `OAuth2LoginConfigurer` and is quite rough to replicate manually.

This PR relaxes this check. Hope this may be reconsidered.

Thanks

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcAuthorizedClientRefreshedEventListener.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcAuthorizedClientRefreshedEventListener.java
@@ -106,11 +106,10 @@ public final class OidcAuthorizedClientRefreshedEventListener
 
 		// The current authentication must be an OAuth2AuthenticationToken
 		Authentication authentication = this.securityContextHolderStrategy.getContext().getAuthentication();
-		if (!(authentication instanceof OAuth2AuthenticationToken authenticationToken)
-				|| authenticationToken.getClass() != OAuth2AuthenticationToken.class) {
+		if (!(authentication instanceof OAuth2AuthenticationToken authenticationToken)) {
 			// This event listener only handles the default authentication result. If the
-			// application customizes the authentication result, then a custom event
-			// handler should be provided.
+			// application customizes the authentication result by not subclassing
+			// OAuth2AuthenticationToken, then a custom event handler should be provided.
 			return;
 		}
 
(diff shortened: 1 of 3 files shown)


oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcAuthorizedClientRefreshedEventListener.java
```diff
@@ -106,11 +106,10 @@ public void onApplicationEvent(OAuth2AuthorizedClientRefreshedEvent event) {
 
 		// The current authentication must be an OAuth2AuthenticationToken
 		Authentication authentication = this.securityContextHolderStrategy.getContext().getAuthentication();
-		if (!(authentication instanceof OAuth2AuthenticationToken authenticationToken)
-				|| authenticationToken.getClass() != OAuth2AuthenticationToken.class) {
+		if (!(authentication instanceof OAuth2AuthenticationToken authenticationToken)) {
```
jgrandja, line 109: This change would not work because the "refreshed" `OAuth2AuthenticationToken` would overwrite the custom/subclass `OAuth2AuthenticationToken` here:

https://github.com/spring-projects/spring-security/blob/7503d8018dc1e59bf809b5349e5a90c92a489f9d/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcAuthorizedClientRefreshedEventListener.java#L136

As [documented](https://github.com/spring-projects/spring-security/blob/7503d8018dc1e59bf809b5349e5a90c92a489f9d/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcAuthorizedClientRefreshedEventListener.java#L111), this `ApplicationListener` only supports `OAuth2AuthenticationToken`. If you have a custom/subclass of `OAuth2AuthenticationToken` then you need to register your own `ApplicationListener`.

jgrandja:
@mpalourdio I'm going to close this PR based on this [comment](https://github.com/spring-projects/spring-security/pull/18169#discussion_r2585440509).

Also, in the future, please log an issue first so we can discuss and see if the enhancement request makes sense and if others are looking for the same. Thank you.

Issue #18157, opened 2025-11-11
NPE in FilterChainProxy.getFilters(String)


lrozenblyum:
**Describe the bug**
`org.springframework.security.web.FilterChainProxy.getFilters(String)` throws NPE in Spring Security 6.5.6.
It used to work in Spring Security 5.8.12.

```
java.lang.NullPointerException
	at java.base/java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1693)
	at org.springframework.security.config.annotation.web.AbstractRequestMatcherRegistry$DeferredRequestMatcher.lambda$new$0(AbstractRequestMatcherRegistry.java:420)
	at org.springframework.security.config.annotation.web.AbstractRequestMatcherRegistry$DeferredRequestMatcher.matches(AbstractRequestMatcherRegistry.java:430)
	at org.springframework.security.web.DefaultSecurityFilterChain.matches(DefaultSecurityFilterChain.java:89)
	at org.springframework.security.web.FilterChainProxy.getFilters(FilterChainProxy.java:248)
	at org.springframework.security.web.FilterChainProxy.getFilters(FilterChainProxy.java:261)
	at org.springframework.security.config.annotation.web.configuration.WebSecurityConfiguration$CompositeFilterChainProxy.getFilters(WebSecurityConfiguration.java:324)
	at org.springframework.security.config.annotation.web.configuration.WebMvcSecurityConfiguration$CompositeFilterChainProxy.getFilters(WebMvcSecurityConfiguration.java:245)

```

**To Reproduce**
Execute the method above with some argument like `/'

**Expected behavior**
List of filters available should be loaded instead of NPE.

**Sample**
```
@Controller
public class MainController {

	@Autowired
	private FilterChainProxy filterChainProxy;
	@GetMapping("/")
	public void mainPage() {
		List<Filter> filters = filterChainProxy.getFilters("/"); // NPE here
	}
```

Same issue is easily reproduced in a `@SpringBootTest`.

ngocnhan-tran1996:
I'm using Spring Boot 3.5.7 and Spring Security 6.5.6, it doesn't throw NPE. May I miss something?

<img width="1730" height="832" alt="Image" src="https://github.com/user-attachments/assets/e7a0d693-a5b6-4357-8799-a3dd1b4e1b09" />

[spring-security-gh-18157.zip](https://github.com/user-attachments/files/23590725/spring-security-gh-18157.zip)

[...]

PR #18194, opened 2025-11-18
Handle Request to Prevent Exceptions

ngocnhan-tran1996:
Related: gh-18157

1. With `PathPatternRequestMatcher`

    If we define a bean like the one below:
    
    ```java
    @Bean
    WebSecurityCustomizer configureWebSecurity() {
        return customizer -> customizer.ignoring()
            .requestMatchers(PathPatternRequestMatcher.withDefaults().matcher("/test/**"));
    }
    ```
  
    an exception will occur because `PathPatternRequestMatcher` invokes `ServletRequestPathUtils#parseAndCache(request)`:
    
    ```
    java.lang.UnsupportedOperationException: public abstract void jakarta.servlet.ServletRequest.setAttribute(java.lang.String, java.lang.Object) is not supported
    ```

2. `Request#getServletContext` can be null, so we need to perform a null check before invoking the `apply` method.

### DIFF
--- a/web/src/test/java/org/springframework/security/web/FilterInvocationTests.java
+++ b/web/src/test/java/org/springframework/security/web/FilterInvocationTests.java
@@ -168,7 +168,7 @@ public class FilterInvocationTests {
 		DummyRequest request = new DummyRequest();
 		request.setAttribute("name", "value");
 		request.setAttribute("removeName", "removeValue");
-		request.removeAttribute("remove");
+		request.removeAttribute("removeName");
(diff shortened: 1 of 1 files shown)


lrozenblyum:
Just wondering, will `FilterChainProxy.getFilters` return anything useful now or it will be always empty?

rwinch:
Thanks for the PR @ngocnhan-tran1996 I'm closing this because, [as pointed out above](https://github.com/spring-projects/spring-security/pull/18194#issuecomment-3562440559), it is incorrect to always return false when the ServletContext is null.

PR #18223, opened 2025-11-26
Fix MyCustomDsl to use csrf(Customizer) instead of removed csrf().disabled()

ghusta:
It should use lambda dsl to compile


rwinch:

Issue #17957, opened 2025-09-25
please don't declare checked exceptions in the configuration DSLs


joshlong:
so, i lvoe the new `Customizer` approach but they dont declare checked exceptions.  (Yes, I know Spring itself has `Throwing*` variants...) and so when i want to use things like the new authorizationServer() DSL method, it requires a try / catch block. 

my $0.02 is that its a config DSL. either fix the DSL so we can't configure ourselves into an invalid state OR report the error at startup time along with everything else. but the way it is, we have to deal with the error at design time AND wait till runtime to figure out what went wrong. only to then have to re-do the DSL, since there's no logical step we could take to compensate for the error. its simply an invalid config. 

```

    @Bean
    Customizer<HttpSecurity> securityCustomizer() {
        return httpSecurity -> {
            try {
                httpSecurity
                        .oauth2AuthorizationServer(x -> x.oidc(Customizer.withDefaults()))
                        .webAuthn(x -> x
                                .allowedOrigins("http://localhost:9090")
                                .rpId("localhost")
                                .rpName("bootiful")
                        )
                        .oneTimeTokenLogin(ott -> ott.tokenGenerationSuccessHandler(
                                new OneTimeTokenGenerationSuccessHandler() {
                                    @Override
                                    public void handle(HttpServletRequest request,
                                                       HttpServletResponse response,
                                                       OneTimeToken oneTimeToken) throws IOException, ServletException {

                                    }
                                }
                        ));
            } catch (Exception e) {
                throw new RuntimeException(e);
            }
        };
    }

```

jzheaux:
@joshlong you make a good point here that I think is worth discussing.

In the meantime, are you able to [use `ThrowingCustomizer`](https://docs.spring.io/spring-security/reference/7.0/servlet/configuration/java.html#httpsecurity-customizer-bean)? (This might be the `Throwing` variant you were referring to, just checking):


```java
@Bean
ThrowingCustomizer<HttpSecurity> securityCustomizer() {
    return httpSecurity -> {
        httpSecurity
            .oauth2AuthorizationServer(x -> x.oidc(Customizer.withDefaults()))
            .webAuthn(x -> x
                .allowedOrigins("http://localhost:9090")
                .rpId("localhost")
                .rpName("bootiful")
            )
            .oneTimeTokenLogin(ott -> ott.tokenGenerationSuccessHandler(
                new OneTimeTokenGenerationSuccessHandler() {
                    @Override
[...]

PR #18224, opened 2025-11-26
Remove throws from MyCustomDsl in docs

ghusta:
In `init` and `configure`, throws Exception has been removed in the super interface `SecurityConfigurer`, since Spring Security 7.0. This change is the consequence of https://github.com/spring-projects/spring-security/issues/17957


rwinch:

PR #18235, opened 2025-11-29
Create SHA-1 MessageDigest for every new check request in Compromised Password Checker

### LINKED ISSUE #18234: HaveIBeenPwnedRestApiPasswordChecker is not thread-safe
`HaveIBeenPwnedRestApiPasswordChecker` stores a single `MessageDigest` instance as a field and reuses it across all invocations of `check()`. Since `MessageDigest` is not thread-safe, concurrent calls can produce incorrect hash values.

**To Reproduce**

1. Register `HaveIBeenPwnedRestApiPasswordChecker` as a singleton bean
2. Call `check()` concurrently from multiple threads
3. Hash computation becomes unreliable due to shared mutable state in `MessageDigest`

**Expected behavior**

The checker should produce correct results under concurrent access. A new `MessageDigest` instance should be created per invocation instead of reusing a shared instance.

garvit-joshi:
closes #18234 .

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/password/HaveIBeenPwnedRestApiPasswordChecker.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/password/HaveIBeenPwnedRestApiPasswordChecker.java
@@ -52,18 +52,12 @@ public final class HaveIBeenPwnedRestApiPasswordChecker implements CompromisedPa
 
 	private final Log logger = LogFactory.getLog(getClass());
 
-	private final MessageDigest sha1Digest;
-
 	private RestClient restClient = RestClient.builder().baseUrl(API_URL).build();
-
-	public HaveIBeenPwnedRestApiPasswordChecker() {
-		this.sha1Digest = getSha1Digest();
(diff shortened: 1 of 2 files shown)


rwinch:
Thank you for the PR @garvit-joshi For bug fixes, the target branch should be [the earliest supported branch the bug is on](https://github.com/spring-projects/spring-security/blob/main/CONTRIBUTING.adoc#issue-lifecycle:~:text=The%20target%20branch%20is%20the%20earliest%20supported%20branch%20where%20the%20change%20will%20be%20applied.).

Would you mind pushing an update that targets the 6.5.x branch (which is the oldest branch that is supported)? I will then merge the PR forward to have it included in the other supported branches.

garvit-joshi:
Hi @rwinch,                                                                                                                                                                        
                                                                                                                                                                                     
I've rebased this PR onto the 6.5.x branch and updated the target accordingly.

rwinch:
Thanks for the PR! This is now merged into 6.5.x, 7.0.x, and main!

Issue #18221, opened 2025-11-25
Default clientSettings.isRequireProofKey=true does not make sense for non auth-code-flow clients


marbon87:
I have a spring boot 4 app with an oauth2 client registration, that is configured with refresh_token for the authorization-grant-type. The client is configured from properties.
Now i get the following warning on app start:

```
WARN  .s.o.c.r.ClientRegistration$Builder - clientSettings.isRequireProofKey=true is only valid with authorizationGrantType=AuthorizationGrantType{value='authorization_code'}. Got authorizationGrantType=AuthorizationGrantType{value='refresh_token'}. Resetting to clientSettings.isRequireProofKey=false
```

Clients with client_credentials-flow will result in the same warning. 

**To Reproduce**
Start an spring boot 4 app with the following properties:
```
spring.security.oauth2.client.registration.example.client-id=my-client
spring.security.oauth2.client.registration.example.authorization-grant-type=refresh_token
spring.security.oauth2.client.registration.example.scope=openid
spring.security.oauth2.client.registration.example.provider=google
```

**Expected behavior**
No warning is logged when a non auth-code-client-registrations is used with defaults.
From my point of view the warning or the default, that pkce is enabled, only makes sense for client-registrations with authorization code flow.

phactum-mnestler:
Same issue here, using `client_credentials` causes this warning to appear in Spring Security 7 with no obvious workaround.

In the meantime we're filtering the logs for log hygiene:
```yaml
logging:
  level:
    "[org.springframework.security.oauth2.client.registration.ClientRegistration$Builder]": ERROR
```

rohan-naik07:
@jgrandja I can take this up if we taking this forward

jgrandja:
@marbon87 Thanks for logging this issue. I believe the only change required here is to remove `logger.warn()`.

@rohan-naik07 Are you able to make this change?

PR #18263, opened 2025-12-04
Remove requireProofKey warning for non-auth-code flows

therepanic:
The warning is unnecessary since PKCE only applies to `authorization_code` flow and the code already corrects this silently.

Closes: gh-18221


jgrandja:
Thanks for the update @therepanic. This is now merged.

Issue #18257, opened 2025-12-03
UserDetailsManager should support password change with nulls


OrangeDog:
**Describe the bug**
Upgrading from 6.x to 7.0, the `org.springframework.security.provisioning` package has now been `@NullMarked`.

Both `getPassword` and `setPassword` in `MutableUser` have been marked as `@Nullable`, but `changePassword` in `UserDetailsManager` has not.

There are situations where you would want to change a password without having the old one (e.g. forgot password resets and admin changes), which naturally involved passing `null` as the `oldPassword`. However, any code that implemented this now generates analysis errors due to the implicit `@NonNull`.

Please add `@Nullable` to at least `oldPassword`, and possibly `newPassword` as well for consistency with `MutableUserDetails`.

PR #18271, opened 2025-12-05
Add `@Nullable` to changePassword parameters in `UserDetailsManager`

therepanic:
Also, an unused argument was removed from `JdbcUserDetailsManager#createNewAuthentication` that was not used and would break the build of the current change.

Closes: gh-18257

### DIFF
--- a/core/src/main/java/org/springframework/security/provisioning/InMemoryUserDetailsManager.java
+++ b/core/src/main/java/org/springframework/security/provisioning/InMemoryUserDetailsManager.java
@@ -51,6 +51,7 @@ import org.springframework.util.Assert;
  * system isn't required.
  *
  * @author Luke Taylor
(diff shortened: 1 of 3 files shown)


therepanic:
Since there's a test that checks protected `JdbcUserDetailsManager#createNewAuthentication` with a password (currently a pointless test), I decided to annotate it with `@Nullable` instead of deleting it.

On the other hand, it might be better to delete the test, since again, the argument isn't used and it's a protected method. WDYT?

rwinch:
@therepanic Thank you for the pull request this is now merged into main 😄 

> On the other hand, it might be better to delete the test, since again, the argument isn't used and it's a protected method. WDYT?

I think it is good to have the test since we don't want to test based upon implementation details (that the argument isn't used) and we want to test protected methods since users can pass in values to it

PR #18272, opened 2025-12-06
KerberosRestClient should rethrow original RestClientException

morokosi:
Currently, the doExecute method in spring−security−kerberos−client always wraps exceptions in a RestClientException. This unnecessarily complicates exception handling for HTTP status errors and other exceptions.

This change modifies the behavior to avoid wrapping an exception if it is already an instance of RestClientException.

This aligns the exception handling with other RestTemplate implementations, making it more consistent and easier to manage.


jzheaux:
Thanks for the PR, @morokosi! This will merge into `main` once the build completes.

Issue #17986, opened 2025-10-02
userDetailsPasswordService cannot be null error when using ReactiveUserDetailsService


matthew-js-porter:
**Describe the bug**
When using Spring Security 7.0.0-M3 and defining a `ReactiveUserDetailsService` the application
fails to start with error `java.lang.IllegalArgumentException: userDetailsPasswordService cannot be null`

**To Reproduce**
- start.spring.io and generate a new 4.0.0 project with webflux and security
- Add `SecurityWebFilterChain` and `ReactiveUserDetailsService` beans.
- start the application and observe the error

**Expected behavior**
The application should start and default a userDetailsPasswordService.

**Sample**
[Sample Repository](https://github.com/matthew-js-porter/spring-security-user-details-password-service-issue)

sephiroth-j:
Same here with [my reactive sample app for LTPA2](/sephiroth-j/spring-security-ltpa2-reactive-sample) and Spring Security v7.0.0.

**stacktace**
```
2025-12-07T17:36:57.688+01:00  WARN 6492 --- [  restartedMain] onfigReactiveWebServerApplicationContext : Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'org.springframework.security.config.annotation.web.reactive.WebFluxSecurityConfiguration': Unsatisfied dependency expressed through method 'setSecurityWebFilterChains' parameter 0: Error creating bean with name 'springSecurityFilterChain' defined in class path resource [hello/config/WebSecurityConfig.class]: Unsatisfied dependency expressed through method 'springSecurityFilterChain' parameter 0: Error creating bean with name 'org.springframework.security.config.annotation.web.reactive.HttpSecurityConfiguration.httpSecurity' defined in class path resource [org/springframework/security/config/annotation/web/reactive/ServerHttpSecurityConfiguration.class]: Failed to instantiate [org.springframework.security.config.web.server.ServerHttpSecurity]: Factory method 'httpSecurity' threw exception with message: userDetailsPasswordService cannot be null
2025-12-07T17:36:57.693+01:00  INFO 6492 --- [  restartedMain] .s.b.a.l.ConditionEvaluationReportLogger : 

Error starting ApplicationContext. To display the condition evaluation report re-run your application with 'debug' enabled.
2025-12-07T17:36:57.705+01:00 ERROR 6492 --- [  restartedMain] o.s.boot.SpringApplication               : Application run failed

[...]

PR #18276, opened 2025-12-08
ServerHttpSecurityConfiguration should not set userDetailsPasswordService to a null value

sephiroth-j:
... when using `ReactiveUserDetailsService` without `ReactiveUserDetailsPasswordService`.

The existing test uses a `MapReactiveUserDetailsService` that implements `ReactiveUserDetailsPasswordService`, which does not trigger the error.

fixes gh-17986

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/reactive/ServerHttpSecurityConfiguration.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/reactive/ServerHttpSecurityConfiguration.java
@@ -255,7 +255,9 @@ class ServerHttpSecurityConfiguration {
 			if (this.passwordEncoder != null) {
 				manager.setPasswordEncoder(this.passwordEncoder);
 			}
-			manager.setUserDetailsPasswordService(this.userDetailsPasswordService);
+			if (this.userDetailsPasswordService != null) {
+				manager.setUserDetailsPasswordService(this.userDetailsPasswordService);
+			}
 			manager.setCompromisedPasswordChecker(this.compromisedPasswordChecker);
 			return this.postProcessor.postProcess(manager);
 		}
--- a/config/src/test/java/org/springframework/security/config/annotation/web/reactive/ServerHttpSecurityConfigurationTests.java
+++ b/config/src/test/java/org/springframework/security/config/annotation/web/reactive/ServerHttpSecurityConfigurationTests.java
@@ -107,9 +107,19 @@ public class ServerHttpSecurityConfigurationTests {
 	}
 
 	@Test
-	public void loadConfigWhenReactiveUserDetailsServiceConfiguredThenServerHttpSecurityExists() {
+	public void loadConfigWhenReactiveUserAuthenticationServiceConfiguredThenServerHttpSecurityExists() {
 		this.spring
 			.register(ServerHttpSecurityConfiguration.class, ReactiveAuthenticationTestConfiguration.class,
+					WebFluxSecurityConfiguration.class)
+			.autowire();
+		ServerHttpSecurity serverHttpSecurity = this.spring.getContext().getBean(ServerHttpSecurity.class);
+		assertThat(serverHttpSecurity).isNotNull();
+	}
+
+	@Test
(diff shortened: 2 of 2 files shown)


config/src/test/java/org/springframework/security/config/annotation/web/reactive/ServerHttpSecurityConfigurationTests.java
```diff
@@ -581,4 +591,14 @@ static Customizer<ServerHttpSecurity> httpSecurityCustomizer0() {
 
 	}
 
+	@Configuration
```
ronodhirSoumik: `@Configuration(proxyBeanMethods = false)`
safer to add this -setting this avoids unwanted proxies

config/src/test/java/org/springframework/security/config/annotation/web/reactive/ServerHttpSecurityConfigurationTests.java
```diff
@@ -107,7 +107,7 @@ void setup(ApplicationContext context) {
 	}
 
 	@Test
-	public void loadConfigWhenReactiveUserDetailsServiceConfiguredThenServerHttpSecurityExists() {
+	public void loadConfigWhenReactiveUserDetailsAndPasswordServiceConfiguredThenServerHttpSecurityExists() {
```
ronodhirSoumik: `loadConfigWhenReactiveUserDetailsAndPasswordServiceConfiguredThenServerHttpSecurityExists` ->
`loadConfigWhenReactiveUserAuthenticationServiceConfiguredThenServerHttpSecurityExists`

config/src/main/java/org/springframework/security/config/annotation/web/reactive/ServerHttpSecurityConfiguration.java
```diff
@@ -255,7 +255,9 @@ private ReactiveAuthenticationManager authenticationManager() {
 			if (this.passwordEncoder != null) {
 				manager.setPasswordEncoder(this.passwordEncoder);
 			}
-			manager.setUserDetailsPasswordService(this.userDetailsPasswordService);
+			if (this.userDetailsPasswordService != null) {
```
jzheaux, line 258: Can you please set the initial value of `userDetailsPasswordService` to `ReactiveUserDetailsPasswordService.NOOP` instead? You can see https://github.com/spring-projects/spring-security/blob/main/core/src/main/java/org/springframework/security/authentication/AbstractUserDetailsReactiveAuthenticationManager.java#L64 for an example.

jzheaux (changes requested):
Thanks for the PR, @sephiroth-j! 

It appears this was introduced by 7c887d2da1b201f15815e180365f7cf7955e9fb5 in 7.0. As such, if you are able, can you please also rebase this PR to `7.0.x`? If that gets tricky, no worries, I can jump in and help.

config/src/main/java/org/springframework/security/config/annotation/web/reactive/ServerHttpSecurityConfiguration.java
sephiroth-j: Is this necessary if `AbstractUserDetailsReactiveAuthenticationManager` already uses `ReactiveUserDetailsPasswordService.NOOP` by default? If `NOOP` were also used here, there would be two places that would need to be changed if `AbstractUserDetailsReactiveAuthenticationManager` changed its default value for `ReactiveUserDetailsPasswordService`.

jzheaux: You are right to note the tradeoff. It's common to Spring Security configuration, favoring lightweight duplication over conditional statements.

That said, it appears that this configuration file uses a condition for password encoder, so using a condition for the password service maintains the pattern internal to this class. Let's keep it as you have it, and we can review this or other configuration files' prevailing pattern at another time.

sephiroth-j:
Hello @jzheaux, I noticed that you have already rebased the branch. Is there anything else that needs to be done/changed?

jzheaux:
Hi, @sephiroth-j, no nothing more is needed, thanks for checking. This is now merged into `7.0.x` and `main`.

Issue #18229, opened 2025-11-26
Multi tenancy and AuthenticationManagerResolver


alfonsoristorato:
**Describe the bug**
Up to 6.5.7, you were able to create (in Kotlin) a class that implements `AuthenticationManagerResolver<String>` and handle potential multi-tenancy as below (below is a simplified version, but you get the gist):
```kotlin
@Component
class AuthManagerResolver(
    @Value($$"${security.issuer1}") val issuer1: String,
    @Value($$"${security.issuer2}") val issuer2: String,
) : AuthenticationManagerResolver<String> {
    private val authenticationManagerCreator = { issuer: String ->
        when (issuer) {
            issuer1 ->
                AuthenticationManager { authentication: Authentication ->
                    val provider =
                        JwtAuthenticationProvider(JwtDecoders.fromIssuerLocation(issuer))
                    provider.authenticate(authentication)
                }
            issuer2 ->
                AuthenticationManager { authentication: Authentication ->
                    val provider =
                        JwtAuthenticationProvider(JwtDecoders.fromIssuerLocation(issuer))
                    provider.authenticate(authentication)
                    
                }
            else -> throw IllegalArgumentException("Unsupported issuer: $issuer")
        }
    }
    private val authenticationManagers: Map<String, AuthenticationManager> =
        mapOf(
            issuer1 to authenticationManagerCreator(issuer1),
            issuer2 to authenticationManagerCreator(issuer2),
        )

    override fun resolve(issuer: String): AuthenticationManager? = authenticationManagers[issuer]
}
```
with v7.0.0, there seems to be some nullability removed:
- in `resolve` -> this makes sense, mostly, you would not want to return a `null` `AuthenticationManager`, although I did not see an issue with that initially, as a null `authManager` would fail authenticating anyway
- in `AuthenticationManager` interface -> not sure I can find where the nullability was removed, but with 7.0.0, the code above requires `JwtAuthenticationProvider.authenticate` to not return nullable `authentication`, which at the moment it does.

Was it intentional to change the nullability in these classes? If so, do you suggest that in Kotlin we handle nullability by ourselves in these cases?

**To Reproduce**
- use Kotlin, Spring 4, Spring Security 7
- create a class as the example above

**Expected behavior**
For `resolve`, as above, I am quite happy to handle the non-presence of the issuer in the map.

For `.authenticate`, given the way it is structured, there should not be any need for it to return a nullable `Authentication`, given how it asserts that the token is not null, as per below:
```java
[...]

PR #18305, opened 2025-12-13
`JwtAuthenticationProvider` `authenticate` nullability fix

alfonsoristorato:
Addressing issue I created https://github.com/spring-projects/spring-security/issues/18229. Not sure if this was ever intended to be nullable, but as we throw when token is null it makes sense to me to add `@NotNull`

No tests added as this is already covered in https://github.com/spring-projects/spring-security/blob/f6cb0bd610ec5cc1739f26f438505ffbf2da5c12/oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/authentication/JwtAuthenticationProviderTests.java#L157


alfonsoristorato:
@rwinch I see you are very active on PRs, sorry to tag :), but it would be good to have some feedback on this or docs update if this nullability is indeed the way forward

alfonsoristorato:
closing this as the nullability got resolved with 7.2

PR #18336, opened 2025-12-19
Fix typo in "Preparing for 7.0" in reference to PathPatternRequestMatcher

ghusta:
In section 'Include the Servlet Path Prefix in Authorization Rules', `PathPatternRequestParser` should be replaced by `PathPatternRequestMatcher`.


ghusta:
I took inspiration from this page, where it seems correct:
- https://docs.spring.io/spring-security/reference/6.5/servlet/authorization/authorize-http-requests.html#match-by-mvc

jzheaux:
Good catch, @ghusta! This is now merged into `6.5.x`, `7.0.x`, and `main`

PR #18345, opened 2025-12-20
Clarify Session Management Persistence Documentation

sankranty:
Corrected that starting from Spring Security 6 security context is not automatically saved by default.

### DIFF
--- a/docs/modules/ROOT/pages/servlet/authentication/session-management.adoc
+++ b/docs/modules/ROOT/pages/servlet/authentication/session-management.adoc
@@ -3,7 +3,8 @@
 
 Once you have got an application that is xref:servlet/authentication/index.adoc[authenticating requests], it is important to consider how that resulting authentication will be persisted and restored on future requests.
 
-This is done automatically by default. If you have a custom filter or controller that is setting the security context, you will need to use a `SecurityContextRepository` to persist it across requests.
+This is done automatically by default.
+If you have a custom filter or controller that is setting the security context, you will need to use a `SecurityContextRepository` to persist it across requests.
 
 If you are upgrading from an older version, you may be interested in the `requireExplicitSave` setting that preserves Spring Security 5's default, though note that this is primarily for migration purposes.
 


docs/modules/ROOT/pages/servlet/authentication/session-management.adoc
```diff
@@ -3,7 +3,7 @@
 
 Once you have got an application that is xref:servlet/authentication/index.adoc[authenticating requests], it is important to consider how that resulting authentication will be persisted and restored on future requests.
 
-This is done automatically by default, so no additional code is necessary, though it is important to know what `requireExplicitSave` means in `HttpSecurity`.
+Starting from Spring Security 6 security context is not persisted automatically by default, thus it is important to know what `requireExplicitSave` means in `HttpSecurity`.
```
ronodhirSoumik: Starting from Spring Security 6,(add comma)

jzheaux: While this first sentence could possibly be clearer, it's not true that the security context is not persisted by default.

The important change from 5 to 6 was that the `SecurityContextPersistenceFilter` would not automatically attempt to save the security context for you. It is still true, though, that each Spring Security filter that interacts with the Security Context will persist it as needed (instead of all of them relying on one filter).

Perhaps the following:

```
This is done automatically by default. If you have a custom filter or controller that is setting the security context, you will need to use a `SecurityContextRepository` to persist it across requests.

If you are upgrading from an older version, you may be interested in the `requireExplicitSave` setting that preserves Spring Security 5's default, though note that this is primarily for migration purposes.
```

jzheaux: Thanks, @ronodhirSoumik, this is now not needed, given later revisions.

jzheaux:
@sankranty, I polished some of the wording to clarify how sessions are persisted in modern Spring Security versions. 

In order to merge your contribution, your DCO signature needs to be corrected. Please note how the email address in your commit header is different from the email address in your DCO signature. Can you take a look at that please and update that commit?

sankranty:
@jzheaux Thank you for adding detailed clarification. I updated sign-off, now it's correct.

sankranty:
Also I noticed that commit https://github.com/spring-projects/spring-security/commit/ea3b112bea2b16757a23a88f2121f78f63fd4aee was added to this branch, what I guess was by accident, so I removed it.

jzheaux:
Thanks, @sankranty! This is now merged into `6.5.x`, `7.0.x`, and `main`.

Issue #18355, opened 2025-12-23
Incorrect description of Passkeys webauthn


iqolome:
**Describe the bug**
The document describes dependency errors related to WebAuthn

**To Reproduce**

<img width="1545" height="606" alt="Image" src="https://github.com/user-attachments/assets/959fe2fa-f2c9-4ea1-bf08-96ee5093977e" />

**Expected behavior**

```xml

// Dependencies should be like this
<dependency>
    <groupId>org.springframework.security</groupId>
    <artifactId>spring-security-webauthn</artifactId>
</dependency>

```

**Sample**

A link to a GitHub repository with a [minimal, reproducible sample]

[docs/modules/ROOT/pages/servlet/authentication/passkeys.adoc](https://github.com/spring-projects/spring-security/blob/main/docs/modules/ROOT/pages/servlet/authentication/passkeys.adoc?plain=1)

```xml

<!-- 

// Remove this spring-boot-starter-security  Already included
<dependency>
    <groupId>org.springframework.security</groupId>
    <artifactId>spring-security-web</artifactId>
</dependency>

-->


// Add this   JdbcPublicKeyCredentialUserEntityRepository and JdbcUserCredentialRepository Included here
 <dependency>
    <groupId>org.springframework.security</groupId>
    <artifactId>spring-security-webauthn</artifactId>
</dependency>



```

rwinch:
Closed in favor of gh-18356

PR #18356, opened 2025-12-23
Fix spring-security-webauthn dependency in passkeys documentation

ngocnhan-tran1996:
Closes gh-18355


rwinch:
Thank you for the PR @ngocnhan-tran1996 

Since this was a bug I changed the base to b 7.0.x. When the build passes it will be merged into 7.0.x and main

PR #18378, opened 2026-01-04
Replace method call with 'Builder.configureMessageConverters()'

Been24:
Deprecated
since 7.0 in favor of configureMessageConverters(Consumer)


jgrandja:
Thanks for the update @Been24. This is now merged.

Issue #17729, opened 2025-08-14
allSerializableClassesShouldHaveSerialVersionOrSuppressWarnings should correctly check for SuppressWarnings


jzheaux:
Since `@SuppressWarnings` uses `RetentionPolicy.SOURCE`, we cannot use it as a criteria for determining whether something in the codebase is correctly serialized, leading to, for example, #17622.

It would be nice to improve this test. It's possible that ArchUnit may be an option to inspect the source code.

johnycho:
Hi @rwinch @jzheaux 

Following up on this issue, I’ve opened a PR to address the problem described here:
[Improve serialVersionUID check in tests #18474](https://github.com/spring-projects/spring-security/pull/18474)

Any feedback would be greatly appreciated. Thanks!

PR #18474, opened 2026-01-10
Improve serialVersionUID check in tests

johnycho:
Fixes gh-17729

## Description

The `allSerializableClassesShouldHaveSerialVersionOrSuppressWarnings` test logic had two issues that caused it to pass incorrectly (false positives):

### 1.  Logic Error
It permitted classes to pass if the annotation was `null` (`suppressWarnings == null`). 
Since `@SuppressWarnings` has `RetentionPolicy.SOURCE`, it is not visible at runtime via reflection, causing `getAnnotation` to always return `null`. Consequently, the test was skipping validation for all classes.
### 2.  Typo
It checked for `"Serial"` (case-sensitive) instead of `"serial"`.

## Fix

This PR updates the validation logic to:
*   Correct the value to `"serial"`.
*   Enforce a strict check by ensuring the class is ignored **only if** the annotation is explicitly present (non-null) and contains the correct value.

## Verification

To demonstrate the fix, I temporarily removed `serialVersionUID` from a class (e.g., `SimpleGrantedAuthority`) and ran the test.

| State | Screenshot | Result |
| :--- | :--- | :--- |
| **Before Fix** | Before fix screenshot | **PASSED** (False Positive) ❌<br/>The test incorrectly passed even though `serialVersionUID` was missing, because the verification logic was broken. |
| **After Fix** | After fix screenshot | **FAILED** (Correct Behavior) ✅<br/>The test now correctly detects the missing `serialVersionUID` and fails, enforcing the strict check as intended. |

## Note
This change fixes the test logic to be correct. However, due to the `RetentionPolicy.SOURCE` limitation of `@SuppressWarnings`, this strict check might cause the test to fail for classes that rely on `@SuppressWarnings("serial")` without a `serialVersionUID`. This reveals that the previous test was not actually verifying those classes.

### DIFF
--- a/cas/src/main/java/org/springframework/security/cas/jackson/CasJacksonModule.java
+++ b/cas/src/main/java/org/springframework/security/cas/jackson/CasJacksonModule.java
@@ -48,6 +48,7 @@ import org.springframework.security.jackson.SecurityJacksonModules;
  * @since 7.0
  * @see SecurityJacksonModules
  */
+@SuppressWarnings("serial")
 public class CasJacksonModule extends SecurityJacksonModule {
 
 	public CasJacksonModule() {
--- a/core/src/test/java/org/springframework/security/authentication/NonBuildableAuthenticationToken.java
+++ b/core/src/test/java/org/springframework/security/authentication/NonBuildableAuthenticationToken.java
@@ -16,6 +16,7 @@
 
 package org.springframework.security.authentication;
 
+@SuppressWarnings("serial")
 public class NonBuildableAuthenticationToken extends TestingAuthenticationToken {
 
 	public NonBuildableAuthenticationToken(String user, String password, String... authorities) {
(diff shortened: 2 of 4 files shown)


jzheaux:
@johnycho, good catch! This has been merged into `6.5.x` in 1a130fca3cc7d8c87a6d3746b3b4a341e2d16239, I also added a polish to look for `@SuppressWarnings` on the source files in acabacb9711c2e6896fe96cbe6db1e849f295994.

Together, these found a number of classes that needed updating. These were added in 08fca57d129aa7b305ad38500bae9aa001c19dbd for `6.5.x`, d4678c8e04a12c79685dd9f7e839b3e71f9f3e75 for `7.0.x`, and 5b8d81828a4f2715b54f2c7c7c16cfa48e64cd15 for `main`.

Issue #18472, opened 2026-01-10
Fix deprecated Gradle plugin usages for Gradle 9.0 upgrade


paulvas:
Running the build with Gradle 8.x emits deprecation warnings for features scheduled for removal in Gradle 9.0. These warnings primarily originate from the io-github-gradle-nexus-publish-plugin, org-jfrog-buildinfo-build-info-extractor-gradle plugins and internal convention plugins (SchemaZipPlugin, JavadocApiPlugin, IntegrationTestPlugin, DocsPlugin).

1. **io.github.gradle-nexus:publish-plugin Plugin** The io.github.gradle-nexus:publish-plugin (currently version 1.3.0) triggers a deprecation warning regarding org.gradle.api.plugins.Convention.
```The org.gradle.api.plugins.Convention type has been deprecated. This is scheduled to be removed in Gradle 9.0. Consult the upgrading guide for further information: https://docs.gradle.org/8.14/userguide/upgrading_version_8.html#deprecated_access_to_conventions
        at spring_security_access_6te1i8djhga35cb865d0zukyu.run(...\spring-security\access\spring-security-access.gradle:1)
        (Run with --stacktrace to get the full stack trace of this deprecation warning.)
```
**Fix:** Update plugin version in libs.versions.toml
```
io-github-gradle-nexus-publish-plugin = "io.github.gradle-nexus:publish-plugin:2.0.0"
```

2. **io.github.gradle-nexus:publish-plugin Plugin** The io.github.gradle-nexus:publish-plugin (currently version 4.34.2) triggers a deprecation warning regarding org.gradle.api.plugins.Convention.
```
The org.gradle.api.plugins.Convention type has been deprecated. This is scheduled to be removed in Gradle 9.0. Consult the upgrading guide for further information: https://docs.gradle.org/8.14/userguide/upgrading_version_8.html#deprecated_access_to_conventions
        at build_61nyef417be5wuvtdx1369msd.run(...\spring-security\build.gradle:23)
        (Run with --stacktrace to get the full stack trace of this deprecation warning.)
The org.gradle.util.ConfigureUtil type has been deprecated. This is scheduled to be removed in Gradle 9.0. Consult the upgrading guide for further information: https://docs.gradle.org/8.14/userguide/upgrading_version_8.html#org_gradle_util_reports_deprecations
        at build_61nyef417be5wuvtdx1369msd.run(...\spring-security\build.gradle:23)
[...]

PR #18477, opened 2026-01-11
Fix/gradle 9 deprecations

paulvas:
### Summary
This Pull Request addresses deprecation warnings in the build configuration to ensure forward compatibility with Gradle 9.

### Changes
- Updated Gradle build scripts to resolve deprecation warnings.
- Adjusted task configurations and dependency declarations where necessary.

### Related Issues
- Fixes gh-18472


rwinch:
Thank you for the PR @paulvas! As this is an enhancement, I can understand why you would have based this off of main. However, I think there is enough value in it and it is part of the build rather tha user facing, so I have changed it to be based on 7.0.x. Upon the build passing, I will merge into 7.0.x and then into main

PR #18478, opened 2026-01-11
Fix duplicated use-authorization-manager in docs

qkrrlgus114:
The <http> namespace appendix listed use-authorization-manager twice; the second entry corresponds to authorization-manager-ref. This PR fixes the attribute name.


jzheaux:
Nice catch, @qkrrlgus114! This will merge into `main` once the build completes.

PR #18479, opened 2026-01-11
Improve And/Or-RequestMatcher/ServerWebExchangeMatcher API

ziqin:
Currently, the List-receiving constructors of `AndRequestMatcher`, `OrRequestMatcher`, `AndServerWebExchangeMatcher`, and `OrServerWebExchangeMatcher` don't support covariance, which adds obstacles to users of these APIs.  For example, one cannot pass a `List<PathPatternRequestMatcher>` to `OrRequestMatcher(List<RequestMatcher>)`.

This PR resolves the aforementioned problem.  It should not break existing code.

### DIFF
--- a/web/src/main/java/org/springframework/security/web/server/util/matcher/AndServerWebExchangeMatcher.java
+++ b/web/src/main/java/org/springframework/security/web/server/util/matcher/AndServerWebExchangeMatcher.java
@@ -42,9 +42,9 @@ public class AndServerWebExchangeMatcher implements ServerWebExchangeMatcher {
 
 	private static final Log logger = LogFactory.getLog(AndServerWebExchangeMatcher.class);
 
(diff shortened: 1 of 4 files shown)


ziqin:
I discovered this problem when working on PR #18499, where I had to explicitly specify the return type of `builder::matcher` as the interface `RequestMatcher` so that I can get a `List<RequestMatcher>` instead of `List<PathPatternRequestMatcher>`.

https://github.com/spring-projects/spring-security/blob/b18c078681b9e17adbd6c47c33e99b433e0bad73/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java#L1031-L1039

If this PR is merged, we can simply use `Arrays.stream(pathPatterns).map(builder::matcher).toList()`.

jzheaux:
Thanks, @ziqin! This is now merged into `main`.

Issue #18330, opened 2025-12-17
Use DefaultParameterNameDiscoverer.getSharedInstance()


rwinch:
As of Spring Framework 7.0.3 it is recommended to use `DefaultParameterNameDiscoverer.getSharedInstance()` rather than re-instantiating a custom `DefaultParameterNameDiscoverer` in every internal place. Most importantly, `MethodParameter#getParameterName()` uses that default discoverer as well now, so there is no need to call `initParameterNameDiscovery` anymore unless you got a custom discoverer to apply.

See https://github.com/spring-projects/spring-framework/issues/36024

therepanic:
Hi, @rwinch. Can I work on this issue?

rwinch:
Thank you @therepanic! Yes the issue is yours

rwinch:
Closing in favor of gh-18481

PR #18481, opened 2026-01-11
Use DefaultParameterNameDiscoverer#getSharedInstance

therepanic:
Due to https://github.com/spring-projects/spring-framework/issues/36024, we should use the shared `DefaultParameterNameDiscoverer#getSharedInstance` instead of creating separate instances of the object.

Closes: gh-18330

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -12,7 +12,7 @@ org-jetbrains-kotlin = "2.2.21"
 org-jetbrains-kotlinx = "1.10.2"
 org-mockito = "5.17.0"
 org-opensaml5 = "5.1.6"
-org-springframework = "7.0.2"
+org-springframework = "7.0.3-SNAPSHOT"
 com-password4j = "1.8.4"
 
 [libraries]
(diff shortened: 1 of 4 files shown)


gradle/libs.versions.toml
```diff
@@ -12,7 +12,7 @@ org-jetbrains-kotlin = "2.2.21"
 org-jetbrains-kotlinx = "1.10.2"
 org-mockito = "5.17.0"
 org-opensaml5 = "5.1.6"
-org-springframework = "7.0.2"
+org-springframework = "7.0.3-SNAPSHOT"
```
therepanic, line 15: I'd like to apologize right away for submitting a relatively late PR. I'll try to avoid that again.

Because `DefaultParameterNameDiscoverer#getSharedInstance` is allowed in the unreleased Spring Framework 7.0.3, and we didn't have it enabled, I thought we should wait until the version was released. I'm still not sure about that.

rwinch: @therepanic No problem. I assumed this was why you were waiting. I'm fine with how you did this.

rwinch:
Thanks for the PR @therepanic! I've merged this into 7.0.x and main

Issue #10826, opened 2022-02-10
Consider having ContentSecurityPolicyHeaderWriter supply a script nonce


jzheaux:
In order to have inline scripts and use a strict CSP, it's important to [provide a `nonce`](https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html#nonces).

Spring Security's `ContentSecurityPolicyHeaderWriter` could generate a nonce if a `{nonce}` placeholder is in the policy directive configuration.

Then, it could be made available as a request attribute for use [in views](https://github.com/spring-projects/spring-security/issues/9529).

ziqin:
Instead of a `{nonce}` placeholder in the directives string, would it be a better idea to accept a `Function<String, String>` in `ContentSecurityPolicyConfig`?

```java
@Bean
SecurityFilterChain filterChain(HttpSecurity http) {
    http
        .headers(headers -> headers
            .contentSecurityPolicy(csp -> csp
                .nonceAttributeName("_csp_nonce")
                .policyDirectivesWithNonce(nonce -> "script-src 'self' 'nonce-" + nonce + "'")
            )
        );
    return http.build();
}
```

ziqin:
I tried to implement nonce support in two approaches. It turns out that the `{nonce}` placeholder approach can be implemented with less changes under the constraint of not breaking existing APIs.

ziqin:
Hi, @jzheaux!

I have implemented the `{nonce}` approach in PR #18499. Would you like to review it?

PR #18499, opened 2026-01-14
Support nonce-based Content Security Policy

ziqin:
This PR implements gh-10826.

## Introduction

When strict Content Security Policy is used, web browsers block inline `<script>` or `<style>` blocks in HTML to mitigate XSS attacks injecting malicious inline blocks. To allow intended inline blocks, web developers can generate a hard-to-guess nonce and specify it in both the CSP and allowed inline blocks.

Currently, Spring Security only supports specifying static content security policy directives. This PR introduces support for dynamically generating a secure random nonce for CSP.

## Implementation

A nonce is written to the CSP header in 2 steps:

1. `NonceGeneratingFilter` / `NonceGeneratingWebFilter` generates a nonce and set it as a request attribute named `_csp_nonce`;
2. `ContentSecurityPolicyHeaderWriter` / `ContentSecurityPolicyServerHttpHeadersWriter` reads the `_csp_nonce` attribute and write it to the CSP header, replacing the `{nonce}` placeholder in the configured `policyDirectives`.

Note:

- The whole process is separated in 2 steps because by default a header writer cannot set a request attribute visible to views rendering the nonce in HTML.
- `_csp_nonce` is chosen as the default attribute name because it has a similar format with the existing `_csrf` attribute. The attribute name is configurable.

## Configuration

This PR also adds configurers for setting up nonce-based CSP with Java/Kotlin lambda DSL, including the ability to specify a request matcher to determine whether a request requires CSP protection.

```java
@Bean
SecurityFilterChain filterChain(HttpSecurity http) {
    http
        .headers((headers) -> headers
            .contentSecurityPolicy((csp) -> csp
                .policyDirectives("default-src 'self'; script-src 'self' 'nonce-{nonce}'; style-src 'self' 'nonce-{nonce}'")
                .nonceAttributeName("_csp_nonce")
                .requireCspMatchers("/admin/**", "/user/**")
            )
        );
    return http.build();
}
```

The ability to enable CSP conditionally is useful especially when the CSP directives contain a nonce, because the HTTP header value becomes dynamic and may change the cacheability of static asserts protected by Spring Security.

The conditional enabling of CSP protection is implemented in `spring-security-config` by wrapping the `ContentSecurityPolicyHeaderWriter` / `ContentSecurityPolicyServerHttpHeadersWriter` with the existing `DelegatingRequestMatcherHeaderWriter` / `ServerWebExchangeDelegatingServerHttpHeadersWriter`. The `_csp_nonce` attribute is generated unconditionally if nonce-based CSP is configured.

## Breaking changes in corner case

In commit 381dc3884b22f8ec9d9d451994d48389db03f3d9 I changed the return type of 2 public configuration APIs:

- `ContentSecurityPolicySpec#reportOnly(boolean)`
- `ContentSecurityPolicySpec#policyDirectives(String)`

The return type is changed from `HeaderSpec` to `ContentSecurityPolicySpec` to allow method chaining.

I believe this API change was missed during the migration to lambda DSL and think it's small enough to be updated when releasing v7.1, but if 100% API stability is required, we could simply revert that commit, or introduce new APIs and deprecate the old ones.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
@@ -1009,13 +1009,13 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 		 * @return the {@link ContentSecurityPolicyConfig} for additional configuration
 		 * @throws IllegalArgumentException if {@code requestMatcher} is null
 		 * @throws IllegalStateException if a {@link RequestMatcher} is already configured
-		 * by a previous call of this method or {@link #requireCspMatchers(String...)}
+		 * by a previous call of this method or {@link #requestMatchers(String...)}
 		 * @since 7.1
-		 * @see #requireCspMatchers(String...)
+		 * @see #requestMatchers(String...)
 		 */
-		public ContentSecurityPolicyConfig requireCspMatcher(RequestMatcher requestMatcher) {
+		public ContentSecurityPolicyConfig requestMatcher(RequestMatcher requestMatcher) {
 			Assert.notNull(requestMatcher, "RequestMatcher cannot be null");
-			Assert.state(this.requestMatcher == null, "RequireCspMatcher(s) is already configured");
+			Assert.state(this.requestMatcher == null, "RequestMatcher(s) is already configured");
 			this.requestMatcher = requestMatcher;
 			return this;
 		}
@@ -1030,14 +1030,14 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 		 * @throws IllegalArgumentException if any path pattern if rejected by
 		 * {@link PathPatternRequestMatcher.Builder#matcher(String)}
 		 * @throws IllegalStateException if a {@link RequestMatcher} is already configured
-		 * by a previous call of this method or {@link #requireCspMatcher(RequestMatcher)}
+		 * by a previous call of this method or {@link #requestMatcher(RequestMatcher)}
 		 * @since 7.1
-		 * @see #requireCspMatcher(RequestMatcher)
+		 * @see #requestMatcher(RequestMatcher)
 		 */
-		public ContentSecurityPolicyConfig requireCspMatchers(String... pathPatterns) {
+		public ContentSecurityPolicyConfig requestMatchers(String... pathPatterns) {
 			PathPatternRequestMatcher.Builder builder = HeadersConfigurer.this.getRequestMatcherBuilder();
 			OrRequestMatcher matcher = new OrRequestMatcher(Arrays.stream(pathPatterns).map(builder::matcher).toList());
-			return this.requireCspMatcher(matcher);
+			return this.requestMatcher(matcher);
 		}
 
 		HeaderWriter getWriter() {
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
@@ -18,10 +18,12 @@ package org.springframework.security.config.annotation.web.configurers;
 
 import java.net.URI;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
 import jakarta.servlet.http.HttpServletRequest;
+import org.jspecify.annotations.Nullable;
 
 import org.springframework.security.config.Customizer;
 import org.springframework.security.config.annotation.configuration.ObjectPostProcessorConfiguration;
@@ -30,11 +32,13 @@ import org.springframework.security.config.annotation.web.builders.HttpSecurity;
 import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
 import org.springframework.security.web.header.HeaderWriter;
 import org.springframework.security.web.header.HeaderWriterFilter;
+import org.springframework.security.web.header.NonceGeneratingFilter;
 import org.springframework.security.web.header.writers.CacheControlHeadersWriter;
 import org.springframework.security.web.header.writers.ContentSecurityPolicyHeaderWriter;
 import org.springframework.security.web.header.writers.CrossOriginEmbedderPolicyHeaderWriter;
 import org.springframework.security.web.header.writers.CrossOriginOpenerPolicyHeaderWriter;
 import org.springframework.security.web.header.writers.CrossOriginResourcePolicyHeaderWriter;
+import org.springframework.security.web.header.writers.DelegatingRequestMatcherHeaderWriter;
 import org.springframework.security.web.header.writers.FeaturePolicyHeaderWriter;
 import org.springframework.security.web.header.writers.HpkpHeaderWriter;
 import org.springframework.security.web.header.writers.HstsHeaderWriter;
@@ -45,6 +49,8 @@ import org.springframework.security.web.header.writers.XContentTypeOptionsHeader
 import org.springframework.security.web.header.writers.XXssProtectionHeaderWriter;
 import org.springframework.security.web.header.writers.frameoptions.XFrameOptionsHeaderWriter;
 import org.springframework.security.web.header.writers.frameoptions.XFrameOptionsHeaderWriter.XFrameOptionsMode;
+import org.springframework.security.web.servlet.util.matcher.PathPatternRequestMatcher;
+import org.springframework.security.web.util.matcher.OrRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher;
 import org.springframework.util.Assert;
 
@@ -75,6 +81,7 @@ import org.springframework.util.Assert;
  * @author Vedran Pavic
  * @author Ankur Pathak
  * @author Daniel Garnier-Moiroux
+ * @author Ziqin Wang
  * @since 3.2
  */
 public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
@@ -273,6 +280,14 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 	public void configure(H http) {
 		HeaderWriterFilter headersFilter = createHeaderWriterFilter();
 		http.addFilter(headersFilter);
+		configureCspNonceGeneratingFilter(http);
+	}
+
+	private void configureCspNonceGeneratingFilter(H http) {
+		ContentSecurityPolicyHeaderWriter writer = this.contentSecurityPolicy.writer;
+		if (writer != null && writer.isNonceBased()) {
+			http.addFilterBefore(new NonceGeneratingFilter(writer.getNonceAttributeName()), HeaderWriterFilter.class);
+		}
 	}
 
 	/**
@@ -302,7 +317,7 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 		addIfNotNull(writers, this.hsts.writer);
 		addIfNotNull(writers, this.frameOptions.writer);
 		addIfNotNull(writers, this.hpkp.writer);
-		addIfNotNull(writers, this.contentSecurityPolicy.writer);
+		addIfNotNull(writers, this.contentSecurityPolicy.getWriter());
 		addIfNotNull(writers, this.referrerPolicy.writer);
 		addIfNotNull(writers, this.featurePolicy.writer);
 		addIfNotNull(writers, this.permissionsPolicy.writer);
@@ -937,11 +952,15 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 
 		private ContentSecurityPolicyHeaderWriter writer;
 
+		private @Nullable RequestMatcher requestMatcher;
+
 		private ContentSecurityPolicyConfig() {
 		}
 
 		/**
-		 * Sets the security policy directive(s) to be used in the response header.
+		 * Sets the security policy directive(s) to be used in the response header. The
+		 * {@code policyDirectives} may contain {@code {nonce}} as placeholders for a
+		 * generated secure random nonce, e.g., {@code script-src 'self' 'nonce-{nonce}'}.
 		 * @param policyDirectives the security policy directive(s)
 		 * @return the {@link ContentSecurityPolicyConfig} for additional configuration
 		 * @throws IllegalArgumentException if policyDirectives is null or empty
@@ -959,6 +978,66 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 		public ContentSecurityPolicyConfig reportOnly() {
 			this.writer.setReportOnly(true);
 			return this;
+		}
+
+		/**
+		 * Sets the name of the servlet request attribute for the generated nonce. Views
+		 * can read this attribute to render the nonce in HTML.
+		 * @param nonceAttributeName the name of the nonce attribute
+		 * @return the {@link ContentSecurityPolicyConfig} for additional configuration
+		 * @throws IllegalArgumentException if {@code nonceAttributeName} is {@code null}
+		 * or empty
+		 * @since 7.1
+		 */
+		public ContentSecurityPolicyConfig nonceAttributeName(String nonceAttributeName) {
+			this.writer.setNonceAttributeName(nonceAttributeName);
+			return this;
+		}
+
+		/**
+		 * Specifies the {@link RequestMatcher} to use for determining when CSP should be
+		 * applied. The default is to enable CSP in every response if
+		 * {@link HeadersConfigurer#contentSecurityPolicy(Customizer)} is configured.
+		 * @param requestMatcher the {@link RequestMatcher} to use
+		 * @return the {@link ContentSecurityPolicyConfig} for additional configuration
+		 * @throws IllegalArgumentException if {@code requestMatcher} is null
+		 * @throws IllegalStateException if a {@link RequestMatcher} is already configured
+		 * by a previous call of this method or {@link #requireCspMatchers(String...)}
+		 * @since 7.1
+		 * @see #requireCspMatchers(String...)
+		 */
+		public ContentSecurityPolicyConfig requireCspMatcher(RequestMatcher requestMatcher) {
+			Assert.notNull(requestMatcher, "RequestMatcher cannot be null");
+			Assert.state(this.requestMatcher == null, "RequireCspMatcher(s) is already configured");
+			this.requestMatcher = requestMatcher;
+			return this;
+		}
+
+		/**
+		 * Specifies the matching path patterns for determining when CSP should be
+		 * applied. The default is to enable CSP in every response if
+		 * {@link HeadersConfigurer#contentSecurityPolicy(Customizer)} is configured.
+		 * @param pathPatterns the path patterns to be matched with a
+		 * {@link PathPatternRequestMatcher}
+		 * @return the {@link ContentSecurityPolicyConfig} for additional configuration
+		 * @throws IllegalArgumentException if any path pattern if rejected by
+		 * {@link PathPatternRequestMatcher.Builder#matcher(String)}
+		 * @throws IllegalStateException if a {@link RequestMatcher} is already configured
+		 * by a previous call of this method or {@link #requireCspMatcher(RequestMatcher)}
+		 * @since 7.1
+		 * @see #requireCspMatcher(RequestMatcher)
+		 */
+		public ContentSecurityPolicyConfig requireCspMatchers(String... pathPatterns) {
+			PathPatternRequestMatcher.Builder builder = HeadersConfigurer.this.getRequestMatcherBuilder();
+			OrRequestMatcher matcher = new OrRequestMatcher(Arrays.stream(pathPatterns).map(builder::matcher).toList());
+			return this.requireCspMatcher(matcher);
+		}
+
+		HeaderWriter getWriter() {
+			if (this.requestMatcher != null) {
+				return new DelegatingRequestMatcherHeaderWriter(this.requestMatcher, this.writer);
+			}
+			return this.writer;
 		}
 
 	}
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
@@ -286,8 +286,11 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 	private void configureCspNonceGeneratingFilter(H http) {
 		ContentSecurityPolicyHeaderWriter writer = this.contentSecurityPolicy.writer;
 		if (writer != null && writer.isNonceBased()) {
-			http.addFilterBefore(new ContentSecurityPolicyNonceGeneratingFilter(writer.getNonceAttributeName()),
-					HeaderWriterFilter.class);
+			ContentSecurityPolicyNonceGeneratingFilter filter = new ContentSecurityPolicyNonceGeneratingFilter();
+			if (this.contentSecurityPolicy.nonceAttributeName != null) {
+				filter.setAttributeName(this.contentSecurityPolicy.nonceAttributeName);
+			}
+			http.addFilterBefore(filter, HeaderWriterFilter.class);
 		}
 	}
 
@@ -953,6 +956,8 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 
 		private ContentSecurityPolicyHeaderWriter writer;
 
+		private @Nullable String nonceAttributeName;
+
 		private @Nullable RequestMatcher requestMatcher;
 
 		private ContentSecurityPolicyConfig() {
@@ -991,7 +996,8 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 		 * @since 7.1
 		 */
 		public ContentSecurityPolicyConfig nonceAttributeName(String nonceAttributeName) {
-			this.writer.setNonceAttributeName(nonceAttributeName);
+			Assert.hasLength(nonceAttributeName, "NonceAttributeName must not be null or empty");
+			this.nonceAttributeName = nonceAttributeName;
 			return this;
 		}
 
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
@@ -30,9 +30,9 @@ import org.springframework.security.config.annotation.configuration.ObjectPostPr
 import org.springframework.security.config.annotation.web.HttpSecurityBuilder;
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
 import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
+import org.springframework.security.web.header.ContentSecurityPolicyNonceGeneratingFilter;
 import org.springframework.security.web.header.HeaderWriter;
 import org.springframework.security.web.header.HeaderWriterFilter;
-import org.springframework.security.web.header.NonceGeneratingFilter;
 import org.springframework.security.web.header.writers.CacheControlHeadersWriter;
 import org.springframework.security.web.header.writers.ContentSecurityPolicyHeaderWriter;
 import org.springframework.security.web.header.writers.CrossOriginEmbedderPolicyHeaderWriter;
@@ -286,7 +286,8 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 	private void configureCspNonceGeneratingFilter(H http) {
 		ContentSecurityPolicyHeaderWriter writer = this.contentSecurityPolicy.writer;
 		if (writer != null && writer.isNonceBased()) {
-			http.addFilterBefore(new NonceGeneratingFilter(writer.getNonceAttributeName()), HeaderWriterFilter.class);
+			http.addFilterBefore(new ContentSecurityPolicyNonceGeneratingFilter(writer.getNonceAttributeName()),
+					HeaderWriterFilter.class);
 		}
 	}
 
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
@@ -280,18 +280,7 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 	public void configure(H http) {
 		HeaderWriterFilter headersFilter = createHeaderWriterFilter();
 		http.addFilter(headersFilter);
-		configureCspNonceGeneratingFilter(http);
-	}
-
-	private void configureCspNonceGeneratingFilter(H http) {
-		ContentSecurityPolicyHeaderWriter writer = this.contentSecurityPolicy.writer;
-		if (writer != null && writer.isNonceBased()) {
-			ContentSecurityPolicyNonceGeneratingFilter filter = new ContentSecurityPolicyNonceGeneratingFilter();
-			if (this.contentSecurityPolicy.nonceAttributeName != null) {
-				filter.setAttributeName(this.contentSecurityPolicy.nonceAttributeName);
-			}
-			http.addFilterBefore(filter, HeaderWriterFilter.class);
-		}
+		http.addFilterBefore(this.contentSecurityPolicy.getNonceGeneratingFilter(), HeaderWriterFilter.class);
 	}
 
 	/**
@@ -1045,6 +1034,14 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 				return new DelegatingRequestMatcherHeaderWriter(this.requestMatcher, this.writer);
 			}
 			return this.writer;
+		}
+
+		ContentSecurityPolicyNonceGeneratingFilter getNonceGeneratingFilter() {
+			var filter = new ContentSecurityPolicyNonceGeneratingFilter();
+			if (this.nonceAttributeName != null) {
+				filter.setAttributeName(this.nonceAttributeName);
+			}
+			return filter;
 		}
 
 	}
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
@@ -985,7 +985,7 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 		 * @return the {@link ContentSecurityPolicyConfig} for additional configuration
 		 * @throws IllegalArgumentException if {@code nonceAttributeName} is {@code null}
 		 * or empty
-		 * @since 7.1
+		 * @since 7.2
 		 */
 		public ContentSecurityPolicyConfig nonceAttributeName(String nonceAttributeName) {
 			Assert.hasLength(nonceAttributeName, "NonceAttributeName must not be null or empty");
@@ -1002,7 +1002,7 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 		 * @throws IllegalArgumentException if {@code requestMatcher} is null
 		 * @throws IllegalStateException if a {@link RequestMatcher} is already configured
 		 * by a previous call of this method or {@link #requestMatchers(String...)}
-		 * @since 7.1
+		 * @since 7.2
 		 * @see #requestMatchers(String...)
 		 */
 		public ContentSecurityPolicyConfig requestMatcher(RequestMatcher requestMatcher) {
@@ -1024,7 +1024,7 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 		 * {@link PathPatternRequestMatcher.Builder#matcher(String)}
 		 * @throws IllegalStateException if a {@link RequestMatcher} is already configured
 		 * by a previous call of this method or {@link #requestMatcher(RequestMatcher)}
-		 * @since 7.1
+		 * @since 7.2
 		 * @see #requestMatcher(RequestMatcher)
 		 */
 		public ContentSecurityPolicyConfig requestMatchers(String... pathPatterns) {
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
@@ -38,7 +38,6 @@ import org.springframework.security.web.header.writers.ContentSecurityPolicyHead
 import org.springframework.security.web.header.writers.CrossOriginEmbedderPolicyHeaderWriter;
 import org.springframework.security.web.header.writers.CrossOriginOpenerPolicyHeaderWriter;
 import org.springframework.security.web.header.writers.CrossOriginResourcePolicyHeaderWriter;
-import org.springframework.security.web.header.writers.DelegatingRequestMatcherHeaderWriter;
 import org.springframework.security.web.header.writers.FeaturePolicyHeaderWriter;
 import org.springframework.security.web.header.writers.HpkpHeaderWriter;
 import org.springframework.security.web.header.writers.HstsHeaderWriter;
@@ -253,6 +252,7 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 	public HeadersConfigurer<H> contentSecurityPolicy(
 			Customizer<ContentSecurityPolicyConfig> contentSecurityCustomizer) {
 		this.contentSecurityPolicy.writer = new ContentSecurityPolicyHeaderWriter();
+		this.contentSecurityPolicy.nonceGeneratingFilter = new ContentSecurityPolicyNonceGeneratingFilter();
 		contentSecurityCustomizer.customize(this.contentSecurityPolicy);
 		return HeadersConfigurer.this;
 	}
@@ -280,7 +280,10 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 	public void configure(H http) {
 		HeaderWriterFilter headersFilter = createHeaderWriterFilter();
 		http.addFilter(headersFilter);
-		http.addFilterBefore(this.contentSecurityPolicy.getNonceGeneratingFilter(), HeaderWriterFilter.class);
+		// nonceGeneratingFilter is instantiated iff CSP is configured
+		if (this.contentSecurityPolicy.nonceGeneratingFilter != null) {
+			http.addFilterBefore(this.contentSecurityPolicy.nonceGeneratingFilter, HeaderWriterFilter.class);
+		}
 	}
 
 	/**
@@ -310,7 +313,7 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 		addIfNotNull(writers, this.hsts.writer);
 		addIfNotNull(writers, this.frameOptions.writer);
 		addIfNotNull(writers, this.hpkp.writer);
-		addIfNotNull(writers, this.contentSecurityPolicy.getWriter());
+		addIfNotNull(writers, this.contentSecurityPolicy.writer);
 		addIfNotNull(writers, this.referrerPolicy.writer);
 		addIfNotNull(writers, this.featurePolicy.writer);
 		addIfNotNull(writers, this.permissionsPolicy.writer);
@@ -945,7 +948,7 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 
 		private ContentSecurityPolicyHeaderWriter writer;
 
-		private @Nullable String nonceAttributeName;
+		private ContentSecurityPolicyNonceGeneratingFilter nonceGeneratingFilter;
 
 		private @Nullable RequestMatcher requestMatcher;
 
@@ -986,7 +989,7 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 		 */
 		public ContentSecurityPolicyConfig nonceAttributeName(String nonceAttributeName) {
 			Assert.hasLength(nonceAttributeName, "NonceAttributeName must not be null or empty");
-			this.nonceAttributeName = nonceAttributeName;
+			this.nonceGeneratingFilter.setAttributeName(nonceAttributeName);
 			return this;
 		}
 
@@ -1005,6 +1008,7 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 		public ContentSecurityPolicyConfig requestMatcher(RequestMatcher requestMatcher) {
 			Assert.notNull(requestMatcher, "RequestMatcher cannot be null");
 			Assert.state(this.requestMatcher == null, "RequestMatcher(s) is already configured");
+			this.writer.setRequestMatcher(requestMatcher);
 			this.requestMatcher = requestMatcher;
 			return this;
 		}
@@ -1027,21 +1031,6 @@ public class HeadersConfigurer<H extends HttpSecurityBuilder<H>>
 			PathPatternRequestMatcher.Builder builder = HeadersConfigurer.this.getRequestMatcherBuilder();
 			OrRequestMatcher matcher = new OrRequestMatcher(Arrays.stream(pathPatterns).map(builder::matcher).toList());
 			return this.requestMatcher(matcher);
-		}
-
-		HeaderWriter getWriter() {
-			if (this.requestMatcher != null) {
-				return new DelegatingRequestMatcherHeaderWriter(this.requestMatcher, this.writer);
-			}
-			return this.writer;
-		}
-
-		ContentSecurityPolicyNonceGeneratingFilter getNonceGeneratingFilter() {
-			var filter = new ContentSecurityPolicyNonceGeneratingFilter();
-			if (this.nonceAttributeName != null) {
-				filter.setAttributeName(this.nonceAttributeName);
-			}
-			return filter;
 		}
 
 	}
--- a/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
+++ b/config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
@@ -2845,21 +2845,21 @@ public class ServerHttpSecurity {
 			 * in the response. Otherwise, defaults to the {@code Content-Security-Policy}
 			 * header.
 			 * @param reportOnly whether to only report policy violations
-			 * @return the {@link HeaderSpec} to continue configuring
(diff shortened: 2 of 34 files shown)


ziqin:
It seems that #18840 may affect the two-step assumption made here.

If HTTP headers are written eagerly, `HeaderWriter#writeHeaders(HttpServletRequest, HttpServletResponse)` may be changed to run earlier, which means it may be possible to set a nonce attribute directly in `ContentSecurityPolicyHeaderWriter` without requiring an extra `NonceGeneratingFilter`.

web/src/main/java/org/springframework/security/web/server/header/NonceGeneratingWebFilter.java
```diff
@@ -0,0 +1,85 @@
...
+ * technology can render the nonce in generated HTML to allow intended inline
+ * {@code <script>} or {@code <style>} blocks.
+ *
+ * <p>
+ * This filter may be used to generate a nonce attribute for other purposes.
+ *
+ * @author Ziqin Wang
+ * @since 7.1
+ */
+public final class NonceGeneratingWebFilter implements WebFilter {
```
jzheaux: It might be a bit ambitious to create a general-purpose nonce-generating filter; I'd want to spend more time thinking of other places where we need nonces placed in the request before that.

Instead, could we please call this `ContentSecurityPolicyNonceGeneratingFilter`? This clarifies the scope of the filter and allows us to create a more general-purpose one down the road as more use cases appear.

web/src/main/java/org/springframework/security/web/server/header/NonceGeneratingWebFilter.java
```diff
@@ -0,0 +1,85 @@
...
+	 * @param attributeName the name of the exchange attribute to generate
+	 * @throws IllegalArgumentException if {@code attributeName} is null or empty string
+	 */
+	public NonceGeneratingWebFilter(String attributeName) {
+		this(attributeName, new Base64StringKeyGenerator(Base64.getEncoder().withoutPadding(), 16));
+	}
+
+	@Override
+	public Mono<Void> filter(ServerWebExchange exchange, WebFilterChain chain) {
+		return Mono.fromSupplier(this.nonceGenerator::generateKey).flatMap((nonce) -> {
```
jzheaux: Let's please place the `Mono` directly into the exchange as an attribute. This will allow so that only when the attribute is read that we have the expense of generating a nonce. You can see `ServerCsrfTokenRequestAttributeHandler` for an example.

web/src/test/java/org/springframework/security/web/header/writers/ContentSecurityPolicyHeaderWriterTests.java
```diff
@@ -126,4 +130,49 @@ public void writeContentSecurityPolicyReportOnlyHeaderWhenNotPresent() {
 		assertThat(this.response.getHeader(CONTENT_SECURITY_POLICY_REPORT_ONLY_HEADER)).isSameAs(value);
 	}
 
+	/** @since 7.1 */
```
jzheaux: No need to add `@since` markers to the tests. Spring Security only adds them to the public API.

web/src/main/java/org/springframework/security/web/server/header/NonceGeneratingWebFilter.java
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * <p>
+ * This filter may be used to generate a nonce attribute for other purposes.
+ *
+ * @author Ziqin Wang
+ * @since 7.1
+ */
+public final class NonceGeneratingWebFilter implements WebFilter {
+
+	private final String attributeName;
```
jzheaux: Let's consider giving this attribute a default name.

web/src/main/java/org/springframework/security/web/server/header/ContentSecurityPolicyServerHttpHeadersWriter.java
```diff
@@ -65,16 +114,42 @@ public void setPolicyDirectives(String policyDirectives) {
...
+	}
+
+	/**
+	 * Returns the name of the {@link ServerWebExchange#getAttribute(String) request
+	 * attribute} from which the nonce value is taken. Defaults to {@code _csp_nonce} if
+	 * unset.
+	 * @return the name of the nonce attribute.
+	 * @since 7.1
+	 */
+	public String getNonceAttributeName() {
```
jzheaux: For now, let's please leave the getters out of the definition This will help reduce the public API. It's easier to add things like this later on once their need is clearer.

web/src/main/java/org/springframework/security/web/server/header/ContentSecurityPolicyServerHttpHeadersWriter.java
```diff
@@ -16,46 +16,95 @@
...
+
+	private boolean isNonceBased;
 
 	@Override
 	public Mono<Void> writeHttpHeaders(ServerWebExchange exchange) {
-		return (this.delegate != null) ? this.delegate.writeHttpHeaders(exchange) : Mono.empty();
+		return Mono.justOrEmpty(this.policyDirectives).flatMap((csp) -> {
+			String headerName = resolveHeader(this.reportOnly);
+			HttpHeaders headers = exchange.getResponse().getHeaders();
+			if (!headers.containsHeader(headerName)) {
```
jzheaux: Please favor separated if statements like so:

```java
if (headers.containsHeader(headerName)) {
    return Mono.empty();
}

if (!this.isNonceBased) {
    headers.put(headerName, List.of(csp));
    return Mono.empty();
}

// ...
```

web/src/main/java/org/springframework/security/web/server/header/ContentSecurityPolicyServerHttpHeadersWriter.java
```diff
@@ -65,16 +114,42 @@ public void setPolicyDirectives(String policyDirectives) {
...
+	}
+
+	/**
+	 * Returns whether the content security policy is nonce-based. The CSP is considered
+	 * nonce-based if the configured {@code policyDirectives} string contains a
+	 * {@code {nonce}} placeholder.
+	 * @return whether the content security policy is nonce-based
+	 * @since 7.1
+	 */
+	public boolean isNonceBased() {
```
jzheaux: For now, let's please leave the getters out of the definition This will help reduce the public API. It's easier to add things like this later on once their need is clearer.

web/src/main/java/org/springframework/security/web/server/header/ContentSecurityPolicyServerHttpHeadersWriter.java
```diff
@@ -16,46 +16,95 @@
...
  * @since 5.1
+ * @see org.springframework.security.web.server.header.NonceGeneratingWebFilter
  */
 public final class ContentSecurityPolicyServerHttpHeadersWriter implements ServerHttpHeadersWriter {
 
 	public static final String CONTENT_SECURITY_POLICY = "Content-Security-Policy";
 
 	public static final String CONTENT_SECURITY_POLICY_REPORT_ONLY = "Content-Security-Policy-Report-Only";
 
+	public static final String DEFAULT_NONCE_ATTRIBUTE_NAME = "_csp_nonce";
```
jzheaux: I'd prefer that Spring Security's internals be independent from how applications want to handle their views. As such, let's please use an internal name for the attribute (like `ContentSecurityPolicyNonceGeneratingWebFilter.class.getName()`, for example). In the web filter, write the nonce to two request attributes; the internal one and the external one. The external one's attribute name is the one that is configurable by the application. Since it is configurable in just one location (per stack) at that point, we should be able to no longer publish a public static value.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
```diff
@@ -961,6 +980,66 @@ public ContentSecurityPolicyConfig reportOnly() {
...
+		 * {@link HeadersConfigurer#contentSecurityPolicy(Customizer)} is configured.
+		 * @param requestMatcher the {@link RequestMatcher} to use
+		 * @return the {@link ContentSecurityPolicyConfig} for additional configuration
+		 * @throws IllegalArgumentException if {@code requestMatcher} is null
+		 * @throws IllegalStateException if a {@link RequestMatcher} is already configured
+		 * by a previous call of this method or {@link #requireCspMatchers(String...)}
+		 * @since 7.1
+		 * @see #requireCspMatchers(String...)
+		 */
+		public ContentSecurityPolicyConfig requireCspMatcher(RequestMatcher requestMatcher) {
```
jzheaux: Let's please call this `requestMatcher`/`requestMatchers` to align with the rest of the DSL

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -2560,6 +2563,10 @@ protected void configure(ServerHttpSecurity http) {
 			ServerHttpHeadersWriter writer = new CompositeServerHttpHeadersWriter(this.writers);
 			HttpHeaderWriterWebFilter result = new HttpHeaderWriterWebFilter(writer);
 			http.addFilterAt(result, SecurityWebFiltersOrder.HTTP_HEADERS_WRITER);
+			if (this.contentSecurityPolicy.isNonceBased()) {
```
jzheaux: Please note my idea about placing a `Mono` (or `Supplier` in the imperative case) in the request, allowing this filter to always be in the filter chain.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -2845,21 +2852,81 @@ private ContentSecurityPolicySpec() {
 			 * in the response. Otherwise, defaults to the {@code Content-Security-Policy}
 			 * header.
 			 * @param reportOnly whether to only report policy violations
-			 * @return the {@link HeaderSpec} to continue configuring
+			 * @return the {@link ContentSecurityPolicySpec} to continue configuring
 			 */
-			public HeaderSpec reportOnly(boolean reportOnly) {
+			public ContentSecurityPolicySpec reportOnly(boolean reportOnly) {
```
jzheaux: As you noted, we cannot change these. Instead, let's deprecate them and reproduce inside of `ContentSecurityPolicySpec`.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -2845,21 +2852,81 @@ private ContentSecurityPolicySpec() {
...
+			 * @throws IllegalStateException if a {@link ServerWebExchangeMatcher} is
+			 * already configured by a previous call of this method or
+			 * {@link #requireCspMatchers(String...)}
+			 * @since 7.1
+			 * @see #requireCspMatchers(String...)
+			 */
+			public ContentSecurityPolicySpec requireCspMatcher(ServerWebExchangeMatcher matcher) {
+				Assert.notNull(matcher, "Matcher must not be null");
+				// Replace the CSP writer in the list with a matcher-decorated writer
+				int idx = HeaderSpec.this.writers.indexOf(HeaderSpec.this.contentSecurityPolicy);
```
jzheaux: Let's instead keep track of the request matcher(s) inside of `ContentSecurityPolicySpec` and apply them when the filter chain is being built.

web/src/main/java/org/springframework/security/web/server/header/ContentSecurityPolicyServerHttpHeadersWriter.java
```diff
@@ -16,46 +16,95 @@
...
+
+	public static final String NONCE_PLACEHOLDER = "{nonce}";
```
jzheaux: It's a little concerning that the definition of the placeholder is tied to the reactive/servlet stack. For that reason, let's please leave this as private for the moment.

config/src/main/kotlin/org/springframework/security/config/annotation/web/headers/ContentSecurityPolicyDsl.kt
```diff
@@ -18,21 +18,63 @@ package org.springframework.security.config.annotation.web.headers
...
+ * @author Ziqin Wang
  * @since 5.3
- * @property policyDirectives the security policy directive(s) to be used in the response header.
- * @property reportOnly includes the Content-Security-Policy-Report-Only header in the response.
  */
 @HeadersSecurityMarker
 class ContentSecurityPolicyDsl {
+    /**
+     * The security policy directive(s) to be used in the response header.
+     * The [policyDirectives] may contain `{code}` as placeholders for a generated secure
```
jzheaux: I think this should be "my contain `{nonce}` as placeholders"

web/src/main/java/org/springframework/security/web/server/header/ContentSecurityPolicyServerHttpHeadersWriter.java
```diff
@@ -16,46 +16,95 @@
...
 	public Mono<Void> writeHttpHeaders(ServerWebExchange exchange) {
-		return (this.delegate != null) ? this.delegate.writeHttpHeaders(exchange) : Mono.empty();
+		return Mono.justOrEmpty(this.policyDirectives).flatMap((csp) -> {
+			String headerName = resolveHeader(this.reportOnly);
+			HttpHeaders headers = exchange.getResponse().getHeaders();
+			if (!headers.containsHeader(headerName)) {
+				if (this.isNonceBased) {
+					String nonce = exchange.getAttribute(this.nonceAttributeName);
+					if (nonce == null) {
+						return Mono.error(new IllegalStateException("Nonce is unset"));
```
jzheaux: Could this error message be more informative? I'm thinking something like "Failed to replace {nonce} placeholders since no nonce found as a request attribute _csp_nonce"

web/src/main/java/org/springframework/security/web/server/header/NonceGeneratingWebFilter.java
```diff
@@ -0,0 +1,85 @@
...
+	/**
+	 * Creates a new instance.
+	 * <p>
+	 * For each exchange, the created filter will generate a secure random nonce value
+	 * with 128-bit entropy and encode it as a Base64 string without padding.
+	 * @param attributeName the name of the exchange attribute to generate
+	 * @throws IllegalArgumentException if {@code attributeName} is null or empty string
+	 */
+	public NonceGeneratingWebFilter(String attributeName) {
+		this(attributeName, new Base64StringKeyGenerator(Base64.getEncoder().withoutPadding(), 16));
```
jzheaux: Let's please use `getUrlEncoder()` since, ensuring that the nonce is URL-safe.

jzheaux (changes requested):
Thanks for contributing this feature, @ziqin! I've left some feedback inline. Mostly, I focused on the reactive stack; please consider applying my observations to both stacks.

web/src/main/java/org/springframework/security/web/server/header/NonceGeneratingWebFilter.java
ziqin: I intentionally chose to use the standard Base64 encoding here for better compatibility.

Content Security Policy Level 2 [only allows Base64 encoding](https://www.w3.org/TR/CSP2/#base64_value):

>		base64-value      = 1*( ALPHA / DIGIT / "+" / "/" )*2( "=" )
>		nonce-value       = base64-value
>		nonce-source      = "'nonce-" nonce-value "'"

The URL-safe variant [is allowed by Content Security Policy Level 3](https://www.w3.org/TR/CSP3/#grammardef-base64-value), which is still a draft currently:

>		nonce-source  = "'nonce-" base64-value "'"
>		base64-value  = 1*( ALPHA / DIGIT / "+" / "/" / "-" / "_" )*2( "=" )

web/src/main/java/org/springframework/security/web/server/header/NonceGeneratingWebFilter.java
ziqin: Resolved in 66c02422c58ba8ffa618ee087053a58e028e0f2b

web/src/test/java/org/springframework/security/web/header/writers/ContentSecurityPolicyHeaderWriterTests.java
ziqin: I have removed `@since` tags in tests. See 6e3e0946ca88a79f00328b84d9af8853e187af42.

web/src/main/java/org/springframework/security/web/server/header/NonceGeneratingWebFilter.java
ziqin: For a general-purpose NonceGeneratingFilter, a default attribute name may not be appropriate. But now that the filter has been renamed to focus on CSP, a default attribute name is quite reasonable.

I have removed `attributeName` from constructor parameters and added a mutator instead.

Updated in bf6f1d4f26ef9defbb2aa73dab902016e8e82968 together with other changes.

web/src/main/java/org/springframework/security/web/server/header/ContentSecurityPolicyServerHttpHeadersWriter.java
ziqin: Resolved in bf6f1d4f26ef9defbb2aa73dab902016e8e82968

config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
ziqin: Now they are renamed to `requestMatcher` (for servlet stack) or `exchangeMatcher` (for reactive stack) in 4c838b34253ae8ce56e616566f3ddb9e80283586

web/src/main/java/org/springframework/security/web/server/header/ContentSecurityPolicyServerHttpHeadersWriter.java
ziqin: Resolved in f666e444f8a28c05a6fd3c6ac2f11e27df521c3f

config/src/main/kotlin/org/springframework/security/config/annotation/web/headers/ContentSecurityPolicyDsl.kt
ziqin: Fixed in 855e946628e5c181781b0ac0361869a208c79f7d

web/src/main/java/org/springframework/security/web/server/header/NonceGeneratingWebFilter.java
ziqin: Lazy generation of nonce is beneficial for performance. Thank you for pointing out this aspect which I didn't take into account. I have updated the code accordingly in 56e388416952bb47745a8590edb2ef8c006c28cd with `SingletonSupplier` and `Mono#cache()` so that the nonce used by views is the same as the one appearing in the CSP header.

Nevertheless, I still have a concern on how to access the nonce in a template. If I remember correctly, in WebFlux applications a `Mono<String>` can be directly accessed in a Thymeleaf template like `<style th:nonce="${_csp_nonce}">`, but I doubt whether there is similar support for `Supplier<String>` in servlet applications.

Is there anything we can do to avoid having users to write different templates for different stacks?

web/src/main/java/org/springframework/security/web/server/header/ContentSecurityPolicyServerHttpHeadersWriter.java
ziqin: Updated in 56e388416952bb47745a8590edb2ef8c006c28cd

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
ziqin: To avoid the hack here, the initialization of `this.writers` in `HeaderSpec` constructor is moved to `HeaderSpec#configure`, which requires me to change the `disable()` method implementations of other `XxxSpec`s. See [my comment](https://github.com/spring-projects/spring-security/pull/18499#pullrequestreview-4058714803) for how theses changes may affect some weird applications.

This is updated in acf956ae4e95b5f8d45b6076f9cad306ff055a3e. Could you please check whether these changes are appropriate?

web/src/main/java/org/springframework/security/web/server/header/ContentSecurityPolicyServerHttpHeadersWriter.java
ziqin: Removed in bf6f1d4f26ef9defbb2aa73dab902016e8e82968 together with other changes.

This was accessed by `HeaderSpec#configure`.  Now that `ContentSecurityPolicyServerHttpHeadersWriter` reads from the `ContentSecurityPolicyNonceGeneratingWebFilter.class.getName()` attribute, there is no need to expose a `getNonceAttributeName()`.

web/src/main/java/org/springframework/security/web/server/header/ContentSecurityPolicyServerHttpHeadersWriter.java
ziqin: Resolved in acf956ae4e95b5f8d45b6076f9cad306ff055a3e and 939ae10c57e2dee5f608323cf1dc17b08845fb2a

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
ziqin: Resolved in acf956ae4e95b5f8d45b6076f9cad306ff055a3e

web/src/main/java/org/springframework/security/web/server/header/ContentSecurityPolicyServerHttpHeadersWriter.java
ziqin: Updated in bf6f1d4f26ef9defbb2aa73dab902016e8e82968 together with other changes.

Note: The attribute name in the error message is not `_csp_nonce` but the class name of the respective nonce-generating filter because header writers read from internal attributes.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
ziqin: I revoked 381dc3884b22f8ec9d9d451994d48389db03f3d9 in 1cec46598241c62d7cbd53e1ca34d3fc440a97e1.

In 11279f95275c1f07083c714a081edceb7523ae70, `ContentSecurityPolicySpec#reportOnly(boolean)` is deprecated in favor of `ContentSecurityPolicySpec#reportOnly()`. The no-parameter `reportOnly()` aligns with the style of `ContentSecurityPolicyConfig#reportOnly()`.

In 32a8249b3ce594d6db26b017a42cc3c46285ed05, `ContentSecurityPolicySpec#policyDirectives(String)` is deprecated in favor of `ContentSecurityPolicySpec#directives(String)`. This change, however, introduces **naming inconsistency** because we don't need to rename in neither Kotlin DSL nor servlet-stack configurer.

Would you like to share your opinion about it?

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -2728,7 +2738,7 @@ public HeaderSpec mode(XFrameOptionsServerHttpHeadersWriter.Mode mode) {
 			 * @return the {@link HeaderSpec} to continue configuring
 			 */
 			public HeaderSpec disable() {
-				HeaderSpec.this.writers.remove(HeaderSpec.this.frameOptions);
+				HeaderSpec.this.frameOptions = null;
 				return HeaderSpec.this;
 			}
```
ziqin, line 2742: If someone has written some weird code like below, the changes here could cause `NullPointerException` which the old code won't throw.

```java
@Bean
SecurityWebFilterChain springSecurity(ServerHttpSecurity http) {
	http
		.headers((headers) -> headers
			.frameOptions((frame) -> {
				frame.disable();
				frame.mode(XFrameOptionsServerHttpHeadersWriter.Mode.DENY);
			})
		);
	return http.build();
}
```

ziqin:
@jzheaux Thank you for your detailed and insightful review feedback!

I have resolved most problems you pointed out, but there are still some points that I am uncertain about (see my comments for details). Would you like to help me about these?

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -2557,9 +2558,18 @@ public HeaderSpec hsts(Customizer<HstsSpec> hstsCustomizer) {
...
+						this.featurePolicy, this.permissionsPolicy, this.contentSecurityPolicy.getWriter(),
+						this.referrerPolicy, this.crossOriginOpenerPolicy, this.crossOriginEmbedderPolicy,
+						this.crossOriginResourcePolicy)
+				.filter(Objects::nonNull);
+			ServerHttpHeadersWriter writer = new CompositeServerHttpHeadersWriter(
+					Stream.concat(builtInWriters, this.customHeadersWriters.stream()).toList());
 			HttpHeaderWriterWebFilter result = new HttpHeaderWriterWebFilter(writer);
 			http.addFilterAt(result, SecurityWebFiltersOrder.HTTP_HEADERS_WRITER);
+			http.addFilterBefore(this.contentSecurityPolicy.getNonceGeneratingFilter(),
+					SecurityWebFiltersOrder.HTTP_HEADERS_WRITER);
```
ziqin: Should we define a new `SecurityWebFiltersOrder` enum for `ContentSecurityPolicyNonceGeneratingWebFilter`? In `SecurityWebFiltersOrder`, is it a good idea to expose the presence of `ContentSecurityPolicyNonceGeneratingWebFilter` to users considering that it may not be necessary in the future if headers are written eagerly?

web/src/main/java/org/springframework/security/web/server/header/NonceGeneratingWebFilter.java
ziqin: A possible solution could be `${_csp.nonce}`, where the request attribute named `_csp` is an object with a `getNonce()` method performing the lazy evaluation, just like `SupplierCsrfToken#getToken()`.

ziqin: Another possible approach is to hack the `toString()` method of the `Supplier<String>` instance to perform the deferred generation.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
```diff
@@ -273,6 +280,7 @@ public HeadersConfigurer<H> defaultsDisabled() {
 	public void configure(H http) {
 		HeaderWriterFilter headersFilter = createHeaderWriterFilter();
 		http.addFilter(headersFilter);
+		http.addFilterBefore(this.contentSecurityPolicy.getNonceGeneratingFilter(), HeaderWriterFilter.class);
```
jzheaux: This seems odd that this filter is added even if CSP is turned off. Can this be added only when `contentSecurityPolicy` is configured?

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -2846,24 +2862,123 @@ private ContentSecurityPolicySpec() {
 			 * header.
 			 * @param reportOnly whether to only report policy violations
 			 * @return the {@link HeaderSpec} to continue configuring
+			 * @deprecated Use {@link #reportOnly()} instead
 			 */
+			@Deprecated(since = "7.1")
 			public HeaderSpec reportOnly(boolean reportOnly) {
-				HeaderSpec.this.contentSecurityPolicy.setReportOnly(reportOnly);
```
jzheaux, line 2851: While I realize its a funky API right now, let's address this in a separate PR so that this one is focused on adding the nonce functionality.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -2846,24 +2862,123 @@ private ContentSecurityPolicySpec() {
...
 			/**
 			 * Sets the security policy directive(s) to be used in the response header.
+			 * The {@code policyDirectives} may contain {@code {nonce}} as placeholders
+			 * for a generated secure random nonce, e.g., {@code script-src 'self'
+			 * 'nonce-{nonce}'}.
 			 * @param policyDirectives the security policy directive(s)
 			 * @return the {@link HeaderSpec} to continue configuring
+			 * @deprecated Use {@link #directives(String)} instead
 			 */
+			@Deprecated(since = "7.1")
```
jzheaux: While I realize its a funky API right now, let's address this in a separate PR so that this one is focused on adding the nonce functionality.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -2483,11 +2486,9 @@ public final class HeaderSpec {
 
 		private CrossOriginResourcePolicyServerHttpHeadersWriter crossOriginResourcePolicy = new CrossOriginResourcePolicyServerHttpHeadersWriter();
 
+		private List<ServerHttpHeadersWriter> customHeadersWriters = new ArrayList<>();
+
 		private HeaderSpec() {
-			this.writers = new ArrayList<>(Arrays.asList(this.cacheControl, this.contentTypeOptions, this.hsts,
```
jzheaux, line 2487: Instead of making this change, can we be a little more conservative and change `this.contentSecurityPolicy` for `this.contentSecurityPolicy.writer` instead?

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -2836,8 +2846,14 @@ public final class ContentSecurityPolicySpec {
 
 			private static final String DEFAULT_SRC_SELF_POLICY = "default-src 'self'";
 
+			private final ContentSecurityPolicyServerHttpHeadersWriter writer = new ContentSecurityPolicyServerHttpHeadersWriter();
+
+			private @Nullable String nonceAttributeName;
```
jzheaux: I lean towards giving this a default value as having a `nonce` is the stronger security position. It also prevents folks from having to specify an attribute name that Spring Security already has defined.

web/src/main/java/org/springframework/security/web/header/writers/ContentSecurityPolicyHeaderWriter.java
```diff
@@ -79,20 +108,26 @@
...
  * @author Joe Grandja
  * @author Ankur Pathak
+ * @author Ziqin Wang
  * @since 4.1
+ * @see org.springframework.security.web.header.ContentSecurityPolicyNonceGeneratingFilter
  */
 public final class ContentSecurityPolicyHeaderWriter implements HeaderWriter {
 
-	private static final String CONTENT_SECURITY_POLICY_HEADER = "Content-Security-Policy";
+	public static final String CONTENT_SECURITY_POLICY_HEADER = "Content-Security-Policy";
```
jzheaux: Unless this is needed in order to introduce `nonce` support, let's please not make these `public` in the same PR. We can take a look at this at another time.

web/src/main/java/org/springframework/security/web/header/writers/ContentSecurityPolicyHeaderWriter.java
```diff
@@ -116,22 +151,38 @@ public ContentSecurityPolicyHeaderWriter(String policyDirectives) {
...
 	 */
 	@Override
+	@SuppressWarnings("unchecked")
 	public void writeHeaders(HttpServletRequest request, HttpServletResponse response) {
 		String headerName = (!this.reportOnly) ? CONTENT_SECURITY_POLICY_HEADER
 				: CONTENT_SECURITY_POLICY_REPORT_ONLY_HEADER;
 		if (!response.containsHeader(headerName)) {
-			response.setHeader(headerName, this.policyDirectives);
+			String csp;
+			if (this.isNonceBased) {
```
jzheaux, line 165: I think we can be avoid the extra state if we do:

```java
if (this.policyDirectives.contains(NONCE_PLACEHOLDER)) {
    Supplier<String> deferredNonce = ...;
    Assert.state(deferredNonce != null, ...);
    csp = this.policyDirectives.replace(NONCE_PLACEHOLDER, deferredNonce.get());
} else { 
   // ...
}
```

web/src/main/java/org/springframework/security/web/header/writers/ContentSecurityPolicyHeaderWriter.java
```diff
@@ -79,20 +108,26 @@
...
 
 	private static final String DEFAULT_SRC_SELF_POLICY = "default-src 'self'";
 
+	private static final String NONCE_PLACEHOLDER = "{nonce}";
+
 	private String policyDirectives;
 
 	private boolean reportOnly;
 
+	private boolean isNonceBased;
```
jzheaux, line 131: I'd prefer to not have this extra state if we can help it. If we introduce other placeholders, a set of booleans can get tricky. Please see my other comment for how we could remove this.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
```diff
@@ -961,6 +975,75 @@ public ContentSecurityPolicyConfig reportOnly() {
...
+		/**
+		 * Specifies the {@link RequestMatcher} to use for determining when CSP should be
+		 * applied. The default is to enable CSP in every response if
+		 * {@link HeadersConfigurer#contentSecurityPolicy(Customizer)} is configured.
+		 * @param requestMatcher the {@link RequestMatcher} to use
+		 * @return the {@link ContentSecurityPolicyConfig} for additional configuration
+		 * @throws IllegalArgumentException if {@code requestMatcher} is null
+		 * @throws IllegalStateException if a {@link RequestMatcher} is already configured
+		 * by a previous call of this method or {@link #requestMatchers(String...)}
+		 * @since 7.1
```
jzheaux: Will you please update these to 7.2?

config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
```diff
@@ -961,6 +975,75 @@ public ContentSecurityPolicyConfig reportOnly() {
...
+		 * can read this attribute to render the nonce in HTML.
+		 * @param nonceAttributeName the name of the nonce attribute
+		 * @return the {@link ContentSecurityPolicyConfig} for additional configuration
+		 * @throws IllegalArgumentException if {@code nonceAttributeName} is {@code null}
+		 * or empty
+		 * @since 7.1
+		 */
+		public ContentSecurityPolicyConfig nonceAttributeName(String nonceAttributeName) {
+			Assert.hasLength(nonceAttributeName, "NonceAttributeName must not be null or empty");
+			this.nonceAttributeName = nonceAttributeName;
```
jzheaux: Because `ContentSecurityPolicyHeaderWriter` is reconstructed each time `headers.contentSecurityPolicy` is invoked, there is a lifecycle mismatch between `nonceAttributeName`/`requestMatcher` and `reportOnly`/`policyDirectives`.

Let's please see if we can set these values eagerly as well. We can use another ticket to see change the fact that a second call to `contentSecurityPolicy` clobbers the first.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
```diff
@@ -961,6 +975,75 @@ public ContentSecurityPolicyConfig reportOnly() {
...
+		 */
+		public ContentSecurityPolicyConfig requestMatchers(String... pathPatterns) {
+			PathPatternRequestMatcher.Builder builder = HeadersConfigurer.this.getRequestMatcherBuilder();
+			OrRequestMatcher matcher = new OrRequestMatcher(Arrays.stream(pathPatterns).map(builder::matcher).toList());
+			return this.requestMatcher(matcher);
+		}
+
+		HeaderWriter getWriter() {
+			if (this.requestMatcher != null) {
+				return new DelegatingRequestMatcherHeaderWriter(this.requestMatcher, this.writer);
```
jzheaux: I appreciate the code reuse. However, I'd recommend adding `requestMatcher` awareness to the content security policy filter, following the pattern of other similar filters.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
```diff
@@ -961,6 +975,75 @@ public ContentSecurityPolicyConfig reportOnly() {
...
+			}
+			return this.writer;
+		}
+
+		ContentSecurityPolicyNonceGeneratingFilter getNonceGeneratingFilter() {
+			var filter = new ContentSecurityPolicyNonceGeneratingFilter();
```
jzheaux: I think it would be okay to default to having this be a pre-constructed field in the config. If they set `nonceAttributeName` to `null`, then that filter can be set to null.

web/src/main/java/org/springframework/security/web/server/header/ContentSecurityPolicyServerHttpHeadersWriter.java
```diff
@@ -16,46 +16,99 @@
...
+
+			if (!this.isNonceBased) {
+				headers.put(headerName, List.of(csp));
+				return Mono.empty();
+			}
+
+			Mono<String> deferredNonce = exchange
+				.getAttribute(ContentSecurityPolicyNonceGeneratingWebFilter.class.getName());
+			if (deferredNonce == null) {
+				return Mono.error(new IllegalStateException(
```
jzheaux: I believe this is correct. Also, this means that `CompositeServerHttpHeadersWriter` will fail fast, and the ensuing headers will not be written.

If you are able, can you please add a separate commit for updating `CompositeServerHttpHeadersWriter` to *not* fail fast? I believe it would mean changing the `concatMap` call to `concatMapDelayError`.

jzheaux (changes requested):
Thanks, @ziqin, for this thorough PR! 

In addition to my inline notes, will you please take a look at <content-security-policy> XML support? We can either:

* Add the nonce-generating filter there too so that the XML support continues to work, or
* It will need an error message stating that nonces aren't supported in XML config

Either way, no need to add nonce-attribute-name, etc. This is to make sure that this change doesn't inadvertently break XML config.

Additionally, for brevity, I made some comments in servlet that naturally also apply to reactive. If you are unsure, please feel free to ask!

web/src/main/java/org/springframework/security/web/server/header/NonceGeneratingWebFilter.java
jzheaux: I agree that introducing a getter would be helpful, like `SupplierCsrfToken`.

ziqin: An interface named `ContentSecurityPolicyNonce` is introduced in aeeafb9518f7699725dcec73911ae688f59838c5. Now both Spring MVC and WebFlux applications can render the nonce with a template like `<style th:nonce="${_csp.nonce}">`.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
ziqin: I've reverted these changes about API deprecation in 0fbc6767bfd325f66f24bacedb20cdf7bfc3d261

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
ziqin: I've reverted these changes about API deprecation in 0fbc6767bfd325f66f24bacedb20cdf7bfc3d261

config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
ziqin: Updated in 9ebbf70cc5ecfff28d0f391ccfd3d15e1bffd609

web/src/main/java/org/springframework/security/web/header/writers/ContentSecurityPolicyHeaderWriter.java
ziqin: Resolved in 3dae3b0b627ae086dac1f9e0868e362c574c570f

config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
ziqin: The initial PR added the nonce-generating filter conditionally by checking whether a `{nonce}` placeholder exists in the policy directives. I had changed it to add the filter unconditionally because I misunderstood your previous comment of “allowing this filter to always be in the filter chain”.

Now, in 6c1fbf343c7844e84750a8dea786c00c8061e129, the configurer is updated to add the nonce-generating filter if and only if contentSecurityPolicy is configured, regardless of whether there's a `{nonce}` placeholder in the directives.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
ziqin: Now the request matcher is moved to `ContentSecurityPolicyHeaderWriter` and `ContentSecurityPolicyServerHttpHeadersWriter` in 6c1fbf343c7844e84750a8dea786c00c8061e129.

web/src/main/java/org/springframework/security/web/server/header/ContentSecurityPolicyServerHttpHeadersWriter.java
ziqin: `CompositeServerHttpHeadersWriter` is updated to not fail fast in 2be35ce3de61089812005ef08347e4f406cdb7c7.

I also updated `CompositeHeaderWriter` analogously for the servlet stack in cba8dd8dfea12f556cc6231f67ecc6bb71c51fb6, but I’m not very confident about the type of exception thrown there (I used `RuntimeException`).

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
ziqin: The return type is changed back to be untouched in this PR.  See 0fbc6767bfd325f66f24bacedb20cdf7bfc3d261.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
ziqin: I had adjusted the logic of header writer composition because I had wanted to wrap the `ContentSecurityPolicyServerHttpHeadersWriter` with a `ServerWebExchangeDelegatingServerHttpHeadersWriter` in `ContentSecurityPolicySpec#getWriter()`.  See also: https://github.com/spring-projects/spring-security/pull/18499#discussion_r3035372442

Now, instead of using a `ServerWebExchangeDelegatingServerHttpHeadersWriter`, the `ServerWebExchangeMatcher` is moved to `ContentSecurityPolicyServerHttpHeadersWriter` as per https://github.com/spring-projects/spring-security/pull/18499#discussion_r3917896892.  Therefore, there is no need of `ContentSecurityPolicySpec#getWriter()` anymore and we can directly put `ContentSecurityPolicyServerHttpHeadersWriter` back into `HeaderSpec` in 6c1fbf343c7844e84750a8dea786c00c8061e129, just like what the code before this PR did.   Consequently, the changes on header writer disabling and composition could be undone in 15a761e491e5eea77c5224162a0861cc9a01ac19.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
ziqin: Having a null `nonceAttributeName` here in `ContentSecurityPolicySpec` means using the default attribute name `_csp_nonce`. (Now the default name is `_csp`.)  When it's not null, it would be passed to `ContentSecurityPolicyNonceGeneratingWebFilter#setAttributeName(String)`, otherwise the filter will use the default attribute name.

In 6c1fbf343c7844e84750a8dea786c00c8061e129, this `nonceAttributeName` field is removed from `ContentSecurityPolicySpec` together with the removal of `ContentSecurityPolicySpec#getNonceGeneratingFilter()`. Now `ContentSecurityPolicySpec#nonceAttributeName(String)` directly configures the pre-constructed `HeaderSpec#nonceGeneratingFilter`.  I believe such changes simplify the code and reduce changes in this PR.

config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
ziqin: Now the `ContentSecurityPolicyServerHttpHeadersWriter` is moved back to `HeaderSpec` from `ContentSecurityPolicySpec` in 6c1fbf343c7844e84750a8dea786c00c8061e129.

`ContentSecurityPolicySpec#nonceAttributeName(String)` and `ContentSecurityPolicySpec#exchangeMatcher(ServerWebExchangeMatcher)` now directly configures on `HeaderSpec.this.nonceGeneratingFilter`, just like `reportOnly` and `policyDirectives`.

The remaining `ContentSecurityPolicySpec#exchangeMatcher` field is for detecting the conflict between `exchangeMatcher(ServerWebExchangeMatcher)` and `exchangeMatchers(String...)`.  The detection should work if conflicting matchers are configured within a single `HeaderSpec#contentSecurityPolicy(Customizer<ContentSecurityPolicySpec>)` lambda, but I'm unsure whether it works with 2 parallel modular configurations declared with 2 `Customizer<ServerHttpSecurity.HeaderSpec>` beans.

ziqin: In 6c1fbf343c7844e84750a8dea786c00c8061e129, `ContentSecurityPolicyConfig#requestMatcher(RequestMatcher)` now directly configures on `ContentSecurityPolicyConfig.this.writer`, just like `reportOnly` and `policyDirectives`.

Similarly, `ContentSecurityPolicyConfig#nonceAttributeName(String)` now directly configures on `ContentSecurityPolicyConfig.this.nonceGeneratingFilter`.

Thus, I believe the lifecycle of these properties should be the same now.

The problem that a second call to `contentSecurityPolicy` clobbers the first could be addressed with the following solution, which is applied by many other inner config classes in `HeadersConfigurer`:

```java
// 1. Add an enable() method to class ContentSecurityPolicyConfig
private ContentSecurityPolicyConfig enable() {
	if (this.writer == null) {
		this.writer = new ContentSecurityPolicyHeaderWriter();
	}
	if (this.nonceGeneratingFilter == null) {
		this.nonceGeneratingFilter = new ContentSecurityPolicyNonceGeneratingFilter();
	}
	return this;
}

// 2. Call contentSecurityPolicy.enable() in HeadersConfigurer#contentSecurityPolicy(Customizer)
// to preserve already constructed writer and nonceGeneratingFilter
public HeadersConfigurer<H> contentSecurityPolicy(
		Customizer<ContentSecurityPolicyConfig> contentSecurityCustomizer) {
	contentSecurityCustomizer.customize(this.contentSecurityPolicy.enable());
	return HeadersConfigurer.this;
}
```

I've added the above changes to a [new branch](https://github.com/ziqin/spring-security/tree/multi-csp-config-call).

config/src/main/java/org/springframework/security/config/annotation/web/configurers/HeadersConfigurer.java
ziqin: In 6c1fbf343c7844e84750a8dea786c00c8061e129, `ContentSecurityPolicyNonceGeneratingFilter` becomes a field in `ContentSecurityPolicyConfig`. Currently it is pre-constructed by `HeadersConfigurer#contentSecurityPolicy(Customizer)`, but is expected to be constructed by a separate `enable()` method.  See the discussion in https://github.com/spring-projects/spring-security/pull/18499#discussion_r3983166754.

I don't understand neither why a user might want to set `nonceAttributeName` to `null`, nor the semantic of a null `nonceAttributeName`.  From my point of view, if a nonce is generated and written to the `Content-Security-Policy` header but cannot be referenced by any inline blocks in server-rendered HTML, the nonce is useless and the user had better remove that `{nonce}` placeholder from the policy directives.  The current implementation of `ContentSecurityPolicyConfig#nonceAttributeName(String)` rejects null argument.

web/src/main/java/org/springframework/security/web/header/writers/ContentSecurityPolicyHeaderWriter.java
ziqin: This extra state was meant to serve as a minor performance enhancement so that the header writer doesn't need to check `policyDirectives.contains(NONCE_PLACEHOLDER)` repeatedly in `writeHeaders`. I haven't done any benchmark about it.

This field is removed in d0733be43f07b582e512313aee95a39b37a2b24e.

web/src/main/java/org/springframework/security/web/header/writers/ContentSecurityPolicyHeaderWriter.java
ziqin: Updated in d0733be43f07b582e512313aee95a39b37a2b24e

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
ziqin: Superseded by https://github.com/spring-projects/spring-security/pull/18499#discussion_r3917650355

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
ziqin: This change is abandoned by 15a761e491e5eea77c5224162a0861cc9a01ac19

ziqin:
@jzheaux

Thanks again for your patient review.

Following your guidance, I have resolved most problems. Please see my replies to inline comments for details and my remaining questions.

- By introducing a request matcher to `ContentSecurityPolicyHeaderWriter` instead of using `DelegatingRequestMatcherHeaderWriter`, I managed to change several places in `HeadersConfigurer` and `ServerHttpSecurity` back to their original state before this PR.

- The request/exchange attribute for the nonce object is renamed to `_csp` so that the nonce value could be referenced in a template as `${_csp.nonce}`.

- XML namespace configuration support is added in 55d55a9a137734b51075b895e5c8e22a6d5d5062 to allow `{nonce}` placeholders in CSP directives.  I used `CompositeFilter` there for ease of implementation, and consequently there will be structural differences between the XML-built filter chain and that built with Java lambda DSL.  I hope they provide the same functionalities.

PR #18506, opened 2026-01-16
Feature/reactive oauth2 token validator

iain-henderson:
OAuth2TokenValidators are currently synchronous only. If a validator has an asynchronous dependency it cannot be used.

This PR updates NimbusReactiveJwtDecoder to utilize ReactiveOAuth2TokenValidator and adds a ReactiveWrappingOAuth2TokenValidator to facilitate using OAuth2TokenValidators in a reactive context.


jgrandja:
@iain-henderson

You are right, when the `OAuth2TokenValidator` is called by `NimbusReactiveJwtDecoder` it will block the reactor event loop if the `OAuth2TokenValidator` performs a synchronous operation.

I'm actually very surprised that this is coming up just now given that `NimbusReactiveJwtDecoder` was introduced in `5.1`. I guess it hasn't been an issue thus far.

Regardless, it is an issue that could happen. However, instead of introducing a new API, this can be solved quite easily by wrapping the validation in `Mono.fromSupplier(supplier).subscribeOn(Schedulers.boundedElastic())`. This pattern is already being used here:

https://github.com/spring-projects/spring-security/blob/14d469cec1bc21d68fa4163a70be8e8f0b6afdca/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java#L501

This would be the preference over introducing a new API and associated implementations.

iain-henderson:
I think I have cleaned this up.
I was focused on enabling a reactive JwtValidator (our implementation uses a Mono to retrieve the signing key to verify the JWT), but I am glad that we can resolve a potential issue blocking on the wrong scheduler.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusReactiveJwtDecoder.java
```diff
@@ -167,7 +180,7 @@ private Mono<Jwt> decode(JWT parsedToken) {
 			// @formatter:off
 			return this.jwtProcessor.convert(parsedToken)
 					.map((set) -> createJwt(parsedToken, set))
-					.map(this::validateJwt)
```
jgrandja, line 170: @iain-henderson 

I believe **the only change required** here is to change `.map(this::validateJwt)` to:

`.flatMap((jwt) -> Mono.fromSupplier(() -> validateJwt(jwt)).subscribeOn(Schedulers.boundedElastic()))`

All other updates should be reverted.

iain-henderson: Then how would you use a reactive validator?

jgrandja: I don't understand? The framework does not have/support a reactive validator.

iain-henderson: This PR does support a reactive validator.

PR #18518, opened 2026-01-18
Clarify Resource Server startup expectations

Hann244:
Refine the documentation to clarify Resource Server startup expectations now that
Spring Boot 2.6+ defers OIDC discovery by default.

* Clarify that, by default, the authorization server does not need to be available at startup.
* Add a concrete code example for users who require eager validation and want startup to fail when the authorization server is unavailable.
* Remove conflicting or outdated statements implying that startup will always fail if the authorization server is down.

Fixes gh-16708


jzheaux:
Thanks for the PR, @Hann244! I polished the documentation slightly, specifically removing the nuance about Boot 2.6 since that version is no longer supported. I also applied similar changes to the equivalent documentation for Reactive Resource Server.

Finally, I rebased it to `6.5.x` so that the update appears in earlier versions of the documentation as well.

This will merge into `6.5.x` once the build completes and then I'll forward-port it to `7.0.x` and `main`.

Issue #18450, opened 2026-01-09
Remove javadoc warnings for spring-security-crypto


rwinch:
See gh-18443 for more details.

jwbeen:
Hi, I’ll take this one and work on fixing the Javadoc warnings in spring-security-crypto and applying the javadoc warnings error plugin.

PR #18519, opened 2026-01-18
Fix Javadoc warnings in spring-security-crypto

jwbeen:
### Summary
This PR removes Javadoc warnings in `spring-security-crypto` and ensures that future warnings fail the build.

### Changes
- Wrap bit-shift expressions in `{@code ...}` to prevent Javadoc from parsing  `<` as HTML.
- Apply the `javadoc-warnings-error` plugin to `spring-security-crypto` so that new Javadoc warnings cause the build to fail.

Closes gh-18450

### DIFF
--- a/crypto/spring-security-crypto.gradle
+++ b/crypto/spring-security-crypto.gradle
@@ -1,5 +1,6 @@
 plugins {
 	id 'security-nullability'
+	id 'javadoc-warnings-error'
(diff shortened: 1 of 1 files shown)


rwinch:
Thanks for the Pull Request! This will be merged into main when the build passes :smile:

PR #18536, opened 2026-01-21
Update servlet/architecture.adoc to use include-code

jkuhel:
This PR updates the servlet architecture (`servlet/architecture.adoc`) docs to use `include-code` block macro and included some additional unit tests. 

Also updated the `docs/antora.xml` configuration to add `include-xml` which is configured to  `docs/src/test/resources`. This allowed the `include-code` macro to include XML-based snippets (e.g. the creation of the `nullRequestCache` bean).  

The only sample that did not use `include-code` was the following due to the `FilterRegistrationBean` from the Spring Boot project not being available. 

E.g. 
```java
@Bean
public FilterRegistrationBean<TenantFilter> tenantFilterRegistration(TenantFilter filter) {
    FilterRegistrationBean<TenantFilter> registration = new FilterRegistrationBean<>(filter);
    registration.setEnabled(false);
    return registration;
}
```

References https://github.com/spring-projects/spring-security/issues/16226

### DIFF
--- a/docs/modules/ROOT/pages/servlet/architecture.adoc
+++ b/docs/modules/ROOT/pages/servlet/architecture.adoc
@@ -301,7 +301,7 @@ In this case, remove the call to `httpBasic` since you are constructing `BasicAu
 ====
 In the event that you are unable to reconfigure `HttpSecurity` to not add a certain filter, you can typically disable the Spring Security filter by calling its DSL's `disable` method like so:
 
-include-java::./CustomizingFilterTests[tag=disable,indent=0]
+include-code::./CustomizingFilterTests[tag=disable,indent=0]
 ====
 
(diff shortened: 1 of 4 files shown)


rwinch:
Thanks for the PR! There were a few changes that I applied:

1) I've updated it to use `include-code` since `include-java` is not valid. With that update, I also updated the kotlin code to have the disable tag See https://github.com/spring-projects/spring-security/pull/18536/commits/4c9ee66e8bd13ba50d3cc63d25949667566248f4
2) I rebased on main and updated the PR to merge into main. I know the directions said to use 6.5.x (you correctly followed the directions) but the directions were stale so this was a problem on our side (sorry). I've also updated the instructions so that doesn't happen for the next contribution

Issue #18543, opened 2026-01-21
AuthoritiesAuthorizationManager.authorize throws NPE if GrantedAuthority returns null


ffray:
**Describe the bug**
When using a complex GrantedAuthority in an Authentication and `AuthoritiesAuthorizationManager.authorize` is used, `AuthoritiesAuthorizationManager.isAuthorized` throws an NPE.

**To Reproduce**
Create a subclass of GrantedAuthority, whose getAuthority returns `null`.
Create an Authentication and add an instance of the custom class as authority.
Create an AuthoritiesAuthorizationManager and call `authorize(authentication, 'SOME_VALUE')`.

**Expected behavior**
As null values are valid according to the documentation and are required in such cases,
a null-check should be introduced.

**Affected version**
The class is located in `org.springframework.security:spring-security-core:7.0.2`

**Sample**

See this [Demo project](https://github.com/ffray/authority-manager-npe.git) showing the issue.
It uses Spring Boot 4.0.1 for Dependency Management.

Khyojae:
I can reproduce this issue with the demo project. I'd like to work on this!

Khyojae:
I have submitted a PR for this issue. You can check it here: #18544

rwinch:
Thanks for the report. I'm closing this in favor of the PR

PR #18544, opened 2026-01-21
 Fix GrantedAuthority.authority null in AuthoritiesAuthorizationManager

Khyojae:
Check if GrantedAuthority is null before accessing it in 
isAuthorized method to prevent NullPointerException.

Closes gh-18543

Signed-off-by: Khyojae <khjae201@gmail.com>

### DIFF
--- a/core/src/test/java/org/springframework/security/authorization/AuthoritiesAuthorizationManagerTests.java
+++ b/core/src/test/java/org/springframework/security/authorization/AuthoritiesAuthorizationManagerTests.java
@@ -14,12 +14,12 @@
  * limitations under the License.
  */
 
-
 package org.springframework.security.authorization;
 
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Set;
 import java.util.function.Supplier;
 
 import org.junit.jupiter.api.Test;
@@ -32,6 +32,7 @@ import org.springframework.security.core.Authentication;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatIllegalArgumentException;
+import static org.assertj.core.api.Assertions.assertThatNullPointerException;
 
 /**
  * Tests for {@link AuthoritiesAuthorizationManager}.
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/springframework/security/authorization/AuthoritiesAuthorizationManager.java
```diff
@@ -69,6 +69,9 @@ private boolean isGranted(Authentication authentication, Collection<String> auth
 
 	private boolean isAuthorized(Authentication authentication, Collection<String> authorities) {
 		for (GrantedAuthority grantedAuthority : getGrantedAuthorities(authentication)) {
+			if (grantedAuthority == null) {
```
ffray: It's not about the grantedAuthority itself, but the result of grantedAuthority.getAuthority() can be null.
I don't think that getGrantedAuthorities may contain a null itself, so the check should be:
```if (grantedAuthority.getAuthority() == null) { continue; } ```

Khyojae: Thank you for the sharp feedback. You're right—it wasn't the grantedAuthority object itself, but the null return value of getAuthority() that caused the NPE.

I've updated the logic to check grantedAuthority.getAuthority() == null as you suggested. Additionally, I've refactored the test case to use a lambda-based GrantedAuthority that returns null, ensuring the issue is accurately reproduced without unnecessary Mockito mocks. Everything is verified with a successful local build. Please take another look

rwinch:
Thanks for the PR! I've:

- changed this to point to 6.5.x since this is a bug and 6.5.x is the oldest OSS supported branch
- Updated the PR to fix checks (checkstyle and ensure that the test is valid since the test used a Collection implementation -- Set.of() reproduces the NPE, but Collections.singleton() does not -- that did not replicate the issue). Related [JDK-83331228](https://bugs.openjdk.org/browse/JDK-8331228?filter=-4&jql=text%20~%20%22ImmutableCollections%20contains%22%20order%20by%20created%20DESC)
- Submitted PRs for merging into 7.0.x and main

rwinch:
This is now merged into 6.5.x, 7.0.x, and main. Thanks again for the PR!

Issue #16869, opened 2025-04-02
ServerCsrfTokenRequestHandler should return reactive types


jzheaux:
If `ServerCsrfTokenRequestHandler#handle` were to return `Mono<Void>` instead of `void`, then implementations could write the token to the response without blocking.

franticticktick:
Hi @jzheaux please assign this ticket to me.

PR #18558, opened 2026-01-22
Add reactive handle method to ServerCsrfTokenRequestHandler

therepanic:
Returning Mono<Void> from `ServerCsrfTokenRequestHandler#handle` would allow non-blocking response writes.

Closes: gh-16869

### DIFF
--- a/web/src/main/java/org/springframework/security/web/server/csrf/ServerCsrfTokenRequestHandler.java
+++ b/web/src/main/java/org/springframework/security/web/server/csrf/ServerCsrfTokenRequestHandler.java
@@ -29,6 +29,7 @@ import org.springframework.web.server.ServerWebExchange;
  * made available to the application through exchange attributes.
  *
  * @author Steve Riesenberg
+ * @author Andrey Litvitski
  * @since 5.8
  * @see ServerCsrfTokenRequestAttributeHandler
  */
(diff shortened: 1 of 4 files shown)


rwinch:
Thanks for the PR! Unfortunately, the changes in this pull request are not passive and we cannot accept it as it is. For this to be accepted, we cannot change the method signature. Instead, you probably need to:

- Introduce a new default method on the interface that returns `Mono<Void>`. The default implementation would call the old method.
- Deprecate the old method for removal in Spring Security 8
- Update Spring Security to call the new method that returns `Mono<Void>`

Are you able to update the PR accordingly?

therepanic:
Sure, I will update the PR ASAP

web/src/main/java/org/springframework/security/web/server/csrf/ServerCsrfTokenRequestHandler.java
```diff
@@ -40,9 +41,23 @@ public interface ServerCsrfTokenRequestHandler extends ServerCsrfTokenRequestRes
...
+	 * @param exchange the {@code ServerWebExchange} with the request being handled
+	 * @param csrfToken the {@code Mono<CsrfToken>} created by the
+	 * {@link ServerCsrfTokenRepository}
+	 * @return a {@code Mono} that completes when handling is finished
+	 */
+	default Mono<Void> handleAsync(ServerWebExchange exchange, Mono<CsrfToken> csrfToken) {
+		handle(exchange, csrfToken);
+		return Mono.empty();
+	}
+
```
therepanic, line 60: Will this method name suit us?

therepanic:
Flaky test I guess?

Issue #17824, opened 2025-08-30
[spring-security/crypto/bcrypt] [performance] SecureRandom created every for single password encode


Mechite:
**Describe the bug**
---
`BCrypt#gensalt` without providing a `SecureRandom` instance, seems to generate one every single time.
https://github.com/spring-projects/spring-security/blob/be64c67af5b4ad732de1b731ece6ce4fe3286932/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCrypt.java#L721

...and `gensalt` seems to be called without this argument if the default constructor for `BCryptPasswordEncoder` is called. 
https://github.com/spring-projects/spring-security/blob/be64c67af5b4ad732de1b731ece6ce4fe3286932/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.java#L105-L116

---

**To Reproduce** - N/A
**Expected behavior** - A persistent instance of `SecureRandom` is created if the user did not provide one. If the `gensalt` method without a provided instance is no longer required, it should then be considered deprecated API.
**Sample** - N/A

yxinot:
Hi! @mechite @rwinch 
I'd like to work on this issue.

**Proposed Solution:**
1. Deprecate `BCrypt.gensalt(String, int)` method
2. Modify `BCryptPasswordEncoder` constructors to create `SecureRandom` once and reuse it
```java
// BCrypt.java - Line 721
@Deprecated(since = "6.4", forRemoval = false)
public static String gensalt(String prefix, int log_rounds) {
    return gensalt(prefix, log_rounds, new SecureRandom());
}

// BCryptPasswordEncoder.java
public BCryptPasswordEncoder() {
    this(-1, new SecureRandom());  // Reuse instance
}
```

This approach:
- Maintains backward compatibility
- Each encoder instance reuses its SecureRandom
- Expected 10-15% performance improvement

I've reproduced the issue locally. Will submit a PR with tests if this looks good!

PR #18560, opened 2026-01-22
Construct SecureRandom in BCryptPasswordEncoder

yxinot:
Deprecate `BCrypt.gensalt(String, int)` and modify `BCryptPasswordEncoder` constructors to create and reuse `SecureRandom` instances.

Closes gh-17824

### DIFF
--- a/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.java
@@ -17,6 +17,7 @@
 package org.springframework.security.crypto.bcrypt;
 
 import java.security.SecureRandom;
+import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -44,7 +45,7 @@ public class BCryptPasswordEncoder extends AbstractValidatingPasswordEncoder {
 
 	private final BCryptVersion version;
 
-	private final SecureRandom random;
+	private final Supplier<SecureRandom> random;
 
 	public BCryptPasswordEncoder() {
 		this(-1);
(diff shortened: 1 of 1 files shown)


crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCrypt.java
```diff
@@ -739,7 +744,12 @@ public static String gensalt(int log_rounds, SecureRandom random) throws Illegal
 	 * factor therefore increases as 2**log_rounds.
 	 * @return an encoded salt value
 	 * @exception IllegalArgumentException if log_rounds is invalid
+	 * @deprecated since 6.4 in favor of {@link #gensalt(int, SecureRandom)}.
+	 * Creating a new {@code SecureRandom} instance on every invocation incurs significant
+	 * performance overhead. Use {@link #gensalt(int, SecureRandom)} with a reusable
+	 * {@code SecureRandom} instance instead.
 	 */
+	@Deprecated(since = "6.4", forRemoval = false)
```
jzheaux: I'm not understanding why the PR deprecates some `gensalt` methods but not all of them; see lines 757 and 766.

Instead of deprecating, I wonder if a better approach would be to have the class use a `SingletonSupplier` to lazily initialize `SecureRandom` only once.

```java
private static final Supplier<SecureRandom> random = SingletonSupplier.of(SecureRandom::new);
```

crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCrypt.java
```diff
@@ -716,7 +716,12 @@ public static String gensalt(String prefix, int log_rounds, SecureRandom random)
 	 * factor therefore increases as 2**log_rounds.
 	 * @return an encoded salt value
 	 * @exception IllegalArgumentException if prefix or log_rounds is invalid
+	 * @deprecated since 6.4 in favor of {@link #gensalt(String, int, SecureRandom)}.
+	 * Creating a new {@code SecureRandom} instance on every invocation incurs significant
+	 * performance overhead. Use {@link #gensalt(String, int, SecureRandom)} with a reusable
+	 * {@code SecureRandom} instance instead.
 	 */
+	@Deprecated(since = "6.4", forRemoval = false)
```
jzheaux: These should be deprecated since `7.1`, though please see my other comment about an alternative to deprecation.

PR #18562, opened 2026-01-23
Remove compiler warnings for spring-security-oauth2-authorization-server

### LINKED ISSUE #18432: Remove compiler warnings for spring-security-oauth2-authorization-server
See gh-18405 for more details.

gisu1102:
## Summary
Closes #18432 

- AOT/runtime hints in the OAuth2 Authorization Server module
  - Before: Included ACCESS_DECLARED_FIELDS, which is no longer applicable in Spring Framework 7 and could trigger compiler warnings.
  - After: Removed ACCESS_DECLARED_FIELDS, keeping only the required INVOKE_DECLARED_* categories.

- Deprecated Jackson2 adapter usage
  - Before: Deprecated Jackson2 adapter types were used without explicit removal suppression.
  - After: Added @SuppressWarnings("removal") only to deprecated Jackson2 adapter classes to document intentional usage.

## Testing

- ./gradlew --no-build-cache clean :spring-security-oauth2-authorization-server:check

## Notes
Other module warnings are unchanged and outside this module’s scope.

### DIFF
--- a/oauth2/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/aot/hint/OAuth2AuthorizationServerBeanRegistrationAotProcessor.java
+++ b/oauth2/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/aot/hint/OAuth2AuthorizationServerBeanRegistrationAotProcessor.java
@@ -139,8 +139,7 @@ class OAuth2AuthorizationServerBeanRegistrationAotProcessor implements BeanRegis
 						TypeReference.of(OAuth2AuthorizationResponseType.class),
 						TypeReference.of(OAuth2TokenFormat.class)),
 						(builder) -> builder.withMembers(MemberCategory.ACCESS_DECLARED_FIELDS,
-								MemberCategory.INVOKE_DECLARED_CONSTRUCTORS,
-								MemberCategory.INVOKE_DECLARED_METHODS));
+								MemberCategory.INVOKE_DECLARED_CONSTRUCTORS, MemberCategory.INVOKE_DECLARED_METHODS));
 
 			// Jackson Modules
 			if (jackson2Present) {
@@ -283,8 +282,7 @@ class OAuth2AuthorizationServerBeanRegistrationAotProcessor implements BeanRegis
 								TypeReference.of(WebServletJackson2Module.class),
 								TypeReference.of(OAuth2AuthorizationServerJackson2Module.class)),
 						(builder) -> builder.withMembers(MemberCategory.ACCESS_DECLARED_FIELDS,
-								MemberCategory.INVOKE_DECLARED_CONSTRUCTORS,
-								MemberCategory.INVOKE_DECLARED_METHODS));
+								MemberCategory.INVOKE_DECLARED_CONSTRUCTORS, MemberCategory.INVOKE_DECLARED_METHODS));
 		}
 
 		private static Class<?> loadClass(String className) {


oauth2/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/aot/hint/OAuth2AuthorizationServerBeanRegistrationAotProcessor.java
```diff
@@ -138,18 +137,17 @@ private void registerHints(RuntimeHints hints) {
...
 						TypeReference.of(OAuth2TokenFormat.class)),
-						(builder) -> builder.withMembers(MemberCategory.DECLARED_FIELDS,
-								MemberCategory.INVOKE_DECLARED_CONSTRUCTORS, MemberCategory.INVOKE_DECLARED_METHODS));
+						(builder) -> builder.withMembers(MemberCategory.INVOKE_DECLARED_CONSTRUCTORS,
+								MemberCategory.INVOKE_DECLARED_METHODS));
 
 			// Jackson Modules
 			if (jackson2Present) {
 				hints.reflection()
 					.registerTypes(
```
rwinch: This needs to continue to register the jackson 2 modules for passivity. You can extract it into a separate method and add a suppression to it.

oauth2/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/aot/hint/OAuth2AuthorizationServerBeanRegistrationAotProcessor.java
```diff
@@ -116,12 +113,14 @@ public void applyTo(GenerationContext generationContext, BeanRegistrationCode be
 		private void registerHints(RuntimeHints hints) {
 			// Collections -> UnmodifiableSet, UnmodifiableList, UnmodifiableMap,
 			// UnmodifiableRandomAccessList, etc.
-			hints.reflection().registerType(Collections.class, MemberCategory.DECLARED_CLASSES);
+			hints.reflection()
+				.registerType(Collections.class, MemberCategory.INVOKE_DECLARED_CONSTRUCTORS,
+						MemberCategory.INVOKE_DECLARED_METHODS);
```
rwinch: This and the similar changes don't seem to be a one to one change from deprecation to replacement. For example, I'd expect this change to be as shown in this suggestion since the deprecation for DECLARED_CLASSES states it is deprecated with no replacement since registerType registers the class. 

```suggestion
			hints.reflection().registerType(Collections.class);
```

Can you help me understand why these changes do not appear to be a one to one replacement for the deprecations?

oauth2/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/aot/hint/OAuth2AuthorizationServerBeanRegistrationAotProcessor.java
```diff
@@ -116,12 +113,14 @@ public void applyTo(GenerationContext generationContext, BeanRegistrationCode be
...
 
 			// HashSet
 			hints.reflection()
-				.registerType(HashSet.class, MemberCategory.DECLARED_FIELDS,
-						MemberCategory.INVOKE_DECLARED_CONSTRUCTORS, MemberCategory.INVOKE_DECLARED_METHODS);
+				.registerType(HashSet.class, MemberCategory.INVOKE_DECLARED_CONSTRUCTORS,
```
rwinch: Can you help me understand why `MemberCategory.DECLARED_FIELDS` wasn't migrated to `MemberCategory.INVOKE_DECLARED_FIELDS` as documented within the Javadoc?

oauth2/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/aot/hint/OAuth2AuthorizationServerBeanRegistrationAotProcessor.java
gisu1102: Agreed. Jackson 2 module registration is still required for passivity when Jackson 2
is present on the classpath.

I will keep the Jackson 2 registrations, extract them into a dedicated helper method,
and scope `@SuppressWarnings("removal")` to that method only so the intent is explicit
while preserving the existing behavior.

Jackson 3 module registration will remain unchanged.
Thank you.

oauth2/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/aot/hint/OAuth2AuthorizationServerBeanRegistrationAotProcessor.java
gisu1102: Thanks for the clarification.

You're right — my change removed MemberCategory.DECLARED_CLASSES but didn’t make it explicit that the type itself is still being registered.
Since DECLARED_CLASSES is deprecated with no replacement (because registerType already registers the class), I’ll add an explicit
`hints.reflection().registerType(T.class)` alongside the existing `INVOKE_*` categories. This preserves the original “type registration”
intent while keeping the reflective invocation hints where needed.

oauth2/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/aot/hint/OAuth2AuthorizationServerBeanRegistrationAotProcessor.java
gisu1102: Thank you for pointing this out.

You’re right — MemberCategory.DECLARED_FIELDS should be migrated to MemberCategory.ACCESS_DECLARED_FIELDS as documented in the Javadoc.

In my earlier change, I removed DECLARED_FIELDS without explicitly replacing it with ACCESS_DECLARED_FIELDS, which made the migration look incomplete and not 1:1.

I’ve updated the code so that:
- All previous DECLARED_FIELDS usages are now migrated to
  ACCESS_DECLARED_FIELDS.
- INVOKE_DECLARED_CONSTRUCTORS / INVOKE_DECLARED_METHODS are kept only where
  reflective invocation was already required.

This preserves the original intent of field access while aligning with the documented replacement in Spring Framework 7.

gisu1102:
@rwinch 
Thanks for the review!

I’ve addressed all of the requested changes:
- Restored Jackson2 module registration for passivity
- Migrated DECLARED_CLASSES and DECLARED_FIELDS 1:1 according to the Spring Framework 7 Javadocs

Could you please take another look when you have a moment?

rwinch:
Thank you for the updates! This will be merged once the build passes

gisu1102:
> Thank you for the updates! This will be merged once the build passes

Thanks for the review and the helpful comments.
I’ll check the build and formatting issues and follow up with an update shortly.

gisu1102:
@rwinch Auto-merge seems disabled after my latest push—would you mind re-enabling it when you get a chance? Thanks!

rwinch:
Thanks for the PR @gisu1102 This is now merged into main

PR #18563, opened 2026-01-23
fix: replace hard-coded JWKS timeouts

knoobie:
Replace the hard-coded 500ms connect- and read-timeouts for the jwks request to use the recommended method that evaluates the system property to lower or increase the timeouts.

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
@@ -42,6 +42,7 @@ import com.nimbusds.jose.jwk.source.JWKSetCacheRefreshEvaluator;
 import com.nimbusds.jose.jwk.source.JWKSetSource;
 import com.nimbusds.jose.jwk.source.JWKSource;
 import com.nimbusds.jose.jwk.source.JWKSourceBuilder;
+import com.nimbusds.jose.jwk.source.RemoteJWKSet;
 import com.nimbusds.jose.proc.DefaultJOSEObjectTypeVerifier;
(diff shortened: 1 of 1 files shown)


oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -561,8 +562,8 @@ private static final class RestTemplateWithNimbusDefaultTimeouts extends RestTem
 
 		private RestTemplateWithNimbusDefaultTimeouts() {
 			SimpleClientHttpRequestFactory requestFactory = new SimpleClientHttpRequestFactory();
-			requestFactory.setConnectTimeout(JWKSourceBuilder.DEFAULT_HTTP_CONNECT_TIMEOUT);
-			requestFactory.setReadTimeout(JWKSourceBuilder.DEFAULT_HTTP_READ_TIMEOUT);
+			requestFactory.setConnectTimeout(RemoteJWKSet.resolveDefaultHTTPConnectTimeout());
```
jzheaux, line 565: I'd prefer to not go from a non-deprecated to a deprecated API. I've added https://bitbucket.org/connect2id/nimbus-jose-jwt/issues/614 to see what the recommended approach is.

jzheaux:
Thanks for this, @knoobie. [0f6f453](https://github.com/spring-projects/spring-security/commit/0f6f453ea0) resolves the underlying timeout concern instead by having `NimbusJwtDecoder` reuse `JwtDecoderProviderConfigurationUtils`'s existing `sun.net.client.defaultConnectTimeout`/`defaultReadTimeout`-aware logic, falling back to 30 seconds. This avoids referencing the deprecated `RemoteJWKSet` constants entirely.

Let's close this in favor of that.

Issue #16530, opened 2025-02-03
Clarify in JSP documentation where method attribute is required


jzheaux:
See #16529 for background.

We should update the reference to indicate that if the target `RequestMatcher` requires an HTTP method, then this method should be indicated in related  `<jsp:authorize>` tags.

Hann244:
Hi @jzheaux — Opened PR #18566 to address this. It adds a NOTE clarifying that <sec:authorize> should specify method when authorization rules use a method-specific RequestMatcher. Thanks!

Issue #16529, opened 2025-02-03
Clarify WebInvocationPrivilegeEvaluator JavaDoc


jzheaux:
When someone passes `null` for the `method` to `WebInvocationPrivilegeEvaluator`, they are saying that the HTTP method doesn't matter to them for matching purposes. However, it's non-trivial to make an authorization decision in any circumstance where the relevant request matcher does require a method.

Imagine the following arrangement:

```java
.requestMatchers(HttpMethod.GET, "/path").hasAuthority("USER")
.requestMatchers(HttpMethod.POST, "/path").permitAll()
```

When `WebInvocationPrivilegeEvaluator#isAllowed("/path", authentication)` is called, which matcher's authorization rules should it use?

This happens in the more generic case as well:

```java
.requestMatchers(HttpMethod.GET, "/path").hasAuthority("USER")
.anyRequest().denyAll()
```

Or, in other words: "if GET /path, then require USER authority; if POST | PUT | DELETE /path, then deny". Here, also, there's no way to know which the user intends, unless they also specify a method.

As such, I believe it's reasonable to require passing a method if you want method-specific request matchers to be considered, which is what the method currently does.

We should update the JavaDoc to be clearer about this, something like:

```
* Note that this will only match authorization rules that don't require a certain {@code HttpMethod}
```

Both `isAllowed` methods should contain this clarification.

PR #18566, opened 2026-01-24
Clarify need for method attribute in JSP authorize tag

Hann244:
Closes gh-16530

Adds a NOTE to the JSP taglibs documentation clarifying that when authorization is configured with a method-specific RequestMatcher, <sec:authorize> should specify the corresponding method so it is evaluated against the intended rule.

This follows the background discussion in gh-16529.


jzheaux:
Thanks, @Hann244! This clarification is now merged into `6.5.x`, `7.0.x`, and `main`.

Issue #18565, opened 2026-01-23
Authorization Server consent fails on client with no registered scopes


Kehrlann:
**Describe the bug**

If a registered client supporting the `authorization_code`  has **no scopes**, and **requires authorization consent**.
When a client makes an authorization request decision, they see a consent screen with no scopes displayed (see screenshot).
If the user clicks "submit consent", an oauth2 error redirect happens, because it is expected that there is at least one scope consented to ([source](https://github.com/spring-projects/spring-security/blob/1b3cf72fc95fc80f840fa1d1b9865365f315120a/oauth2/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationConsentAuthenticationProvider.java#L203-L220))

**Expected behavior**

I see three options:

- Disallow the "require consent + 0 scope" combination for clients
- Allow the combination but skip the consent screen
- Allow the combination, present the screen and allow the user to click "submit consent"

**Screenshot**

<img width="794" height="466" alt="Image" src="https://github.com/user-attachments/assets/661c5170-bffa-40a8-8dfb-35bb2129c0cf" />

ryujungkyun:
I'd like to work on this issue.

Kehrlann:
For info, my current workaround, to skip the screen if there are no registered scopes:

```java
/**
 * Customizer for {@link OAuth2AuthorizationCodeRequestAuthenticationProvider}, that
 * removes the need for consent when a client was registered with 0 scopes. Otherwise, the
 * user cannot get past the consent screen with no scopes present. It is lifted from
 * {@link OAuth2AuthorizationCodeRequestAuthenticationProvider}'s
 * {@code isAuthorizationConsentRequired}, with a special case for "no scopes registered".
 *
 * @see <a href=
 * "https://github.com/spring-projects/spring-security/issues/18565">spring-security/issues/18565</a>
 * @author Daniel Garnier-Moiroux
 */
class McpNoScopeClientConsentNotRequired implements Predicate<OAuth2AuthorizationCodeRequestAuthenticationContext> {

	@Override
	public boolean test(OAuth2AuthorizationCodeRequestAuthenticationContext authenticationContext) {
		if (!authenticationContext.getRegisteredClient().getClientSettings().isRequireAuthorizationConsent()) {
			return false;
		}
		if (authenticationContext.getAuthorizationRequest().getScopes().isEmpty()) {
			return false;
		}
[...]

PR #18574, opened 2026-01-25
Fix gh-18565: Support consent for clients with no registered scopes

ryujungkyun:
### Description
This PR addresses **gh-18565**, where the `OAuth2AuthorizationConsentAuthenticationProvider` incorrectly handles authorization consent for a `RegisteredClient` with no registered scopes.

When `requireAuthorizationConsent(true)` is set for a client with no scopes, the provider currently:
1. Throws an `ACCESS_DENIED` error because it assumes an empty authority set implies user denial.
2. Throws an `IllegalArgumentException` from `OAuth2AuthorizationConsent.Builder` because building with an empty authority set is not permitted.

### Changes
- Updated the logic to only throw `ACCESS_DENIED` if `requestedScopes` is not empty.
- Conditionally call `OAuth2AuthorizationConsent.Builder.build()` only when `authorities` are present.
- Added a cleanup step to remove stale consent records when the resulting authority set is empty.

### Related Issue
Ref: gh-18565

### How to Test
I have added a reproduction test case: `authenticateWhenNoScopesAndConsentRequiredThenReturnAuthorizationCode()` in `OAuth2AuthorizationConsentAuthenticationProviderTests`.


jgrandja:
@ryujungkyun I'm going to close this as I don't believe any changes are necessary. Please see [comment](https://github.com/spring-projects/spring-security/issues/18565#issuecomment-3846928365).

PR #18599, opened 2026-01-27
Support single-line PEM encoded RSA keys in `RsaKeyConverters`

earlgrey02:
`RsaKeyConverters` currently assumes that PEM-encoded RSA keys are provided in a multi-line format, where the PEM header, footer, and Base64-encoded content are separated across multiple lines.
However, in environment variable–based configurations (ex. `.env` files), PEM keys are commonly stored as single-line strings, with newline characters removed or escaped.

Examples:

- `-----BEGIN PUBLIC KEY-----MIIBIjANBgkq...IDAQAB-----END PUBLIC KEY-----`
- `-----BEGIN PUBLIC KEY-----\nMIIBIjANBgkq...IDAQAB\n-----END PUBLIC KEY-----`

While these representations are functionally equivalent to the traditional multi-line PEM format, they are currently not supported by `RsaKeyConverters` and result in conversion failures.
Therefore, this PR improves `RsaKeyConverters` to correctly handle single-line RSA keys while preserving the existing conversion structure.


jzheaux:
Thanks, @earlgrey02! This is now merged into `main`.

PR #18304, opened 2025-12-13
Document Client PKCE settings

bloomsei:
Updates documentation to reflect that PKCE is now enabled by default for `authorization_code` flows in both authorization server and client.

Changes include:
- Documenting the default PKCE behavior for authorization code flows
- Adding instructions for disabling PKCE when not supported

The documented changes were introduced by:
- gh-16391
- gh-17507

PR #18614, opened 2026-01-28
Document RegisteredClient.ClientSettings

bloomsei:
Added documentation to describe the possible client configuration options using `ClientSettings` when setting up an Oauth2 Authorization Server.

This change was discussed in: gh-18304

### DIFF
--- a/docs/modules/ROOT/pages/servlet/oauth2/authorization-server/core-model-components.adoc
+++ b/docs/modules/ROOT/pages/servlet/oauth2/authorization-server/core-model-components.adoc
@@ -92,6 +92,41 @@ public class RegisteredClient implements Serializable {
 <12> `clientSettings`: The custom settings for the client – for example, require https://datatracker.ietf.org/doc/html/rfc7636[PKCE], require authorization consent, and others.
 <13> `tokenSettings`: The custom settings for the OAuth2 tokens issued to the client – for example, access/refresh token time-to-live, reuse refresh tokens, and others.
 
+[[oauth2AuthorizationServer-client-settings]]
+== ClientSettings
+
+`ClientSettings` contains configuration for the `RegisteredClient`. The following example shows the available settings:
+
+[source,java]
+----
+public final class ClientSettings extends AbstractSettings {
+
+	...
+
+	public static Builder builder() {
+		return new Builder()
+			.requireProofKey(true)  <1>
+			.requireAuthorizationConsent(false)  <2>
+			.jwkSetUrl("https://client.example.com/jwks")  <3>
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/servlet/oauth2/authorization-server/core-model-components.adoc
```diff
@@ -92,6 +92,32 @@ public class RegisteredClient implements Serializable {
...
+
+[source,java]
+----
+ClientSettings.builder()
+	.requireProofKey()  <1>
+	.requireAuthorizationConsent()  <2>
+	.jwkSetUrl()  <3>
+	.tokenEndpointAuthenticationSigningAlgorithm()  <4>
+	.x509CertificateSubjectDN()  <5>
+	.build();
```
elaynebloom: I know this does not match the rest of this documentation page. To match it, I would have had to describe `ClientSettings` and not its builder. However, that would then describe methods like `isRequireProofKey`, which is not very helpful for the users. 
Personally, as I user I would more expect configuration options to be described as what I would use in my `application.yml/.properties` rather than configuration classes themselves, but that would be even further away from the rest of the configuration.
I am happy to adopt another way of documenting this, if you tell me what to prefer in this case.

jgrandja: @bloomsei Can you change the format to match the code sample in [Configuring Authorization Server Settings](https://docs.spring.io/spring-security/reference/servlet/oauth2/authorization-server/configuration-model.html#oauth2AuthorizationServer-configuring-authorization-server-settings)

elaynebloom: I followed the code sample that you specified and used the example values from your tests. The other option would be to use the variables like [Configuring Client Authentication](https://docs.spring.io/spring-security/reference/servlet/oauth2/authorization-server/configuration-model.html#oauth2AuthorizationServer-configuring-client-authentication), but I find this provides useful additional information. I hope this fits, otherwise happy to go into another direction.

Thank you for reviewing!

jgrandja: Apologies @bloomsei but I lead you down the wrong path when I asked you to follow the Authorization Server settings code sample. To be consistent with the rest of the page, I think describing the accessors makes the most sense, given that there are no typed members since it's backed by a `Map`.

I went ahead and made the change in a polish commit. This is now merged. Thanks for the updates!

Issue #18622, opened 2026-01-30
OAuth2 client sessionManagement ineffective with DefaultOidcUser


zhanming:
**Describe the bug**

I want to limit user sessions in OAuth2 client like this, but It did not work.

.sessionManagement(session -> session
    .maximumSessions(1)
    .maxSessionsPreventsLogin(false) // old session will expired.
    .sessionRegistry(sessionRegistry()
)

I debug into the SessionRegistryImpl.class.

public List<SessionInformation> getAllSessions(Object principal, boolean includeExpiredSessions)

this first line:

Set<String> sessionsUsedByPrincipal = this. principals.get(principal);

When use OIDC login from OP (Spring Authorization Server) and login successfully, when RP(OAuth2 client) want to compare user session, the  getAllSessions method always return an empty collection, because the principal is DefaultOidcUser, and extends DefaultOAuth2User, and DefaultOAuth2User's equals and hashCode method compares the Attributes, but each time after OIDC Login, the attributes have these fields: nonce, jti and sid and they are not the same value every time, and they should not the same, so the question is can we not compare attributes?

Can DefaultOidcUser have its own equals and hashCode method without compare attributes?
or can DefaultOAuth2User change to not compare attributes?
or other way?

Thanks.

**To Reproduce**

User Spring Boot 4.0.2, Spring Security 7.0.2

OAuth2 client want limit user session. 

**Expected behavior**

maximumSession work as expected.

**Sample**

OAuth2 client's SecurityConfig like this:

```
@Configuration
@EnableWebSecurity
public class OAuth2ClientSecurityConfig {

    // 1. Define the SessionRegistry bean
    @Bean
    public SessionRegistry sessionRegistry() {
        return new SessionRegistryImpl();
    }

    // 2. Define the HttpSessionEventPublisher bean
   @Bean
    public HttpSessionEventPublisher httpSessionEventPublisher() {
        return new HttpSessionEventPublisher();
    }

    // 3. Configure the SecurityFilterChain
    @Bean
    public SecurityFilterChain securityFilterChain(HttpSecurity http) throws Exception {
        http
            .authorizeHttpRequests(authorize -> authorize
                .anyRequest().authenticated()
            )
            .oauth2Login(oauth2 -> oauth2
                .defaultSuccessUrl("/")
            )
            .sessionManagement(session -> session
                .maximumSessions(1)
                .maxSessionsPreventsLogin(false) 
                .sessionRegistry(sessionRegistry())
            );

[...]

Issue #16685, opened 2025-03-01
Logging in with passkeys bypasses maximum session restriction


ziqin:
## Describe the bug

In a Spring-built website with maximum session restriction, a user logging in with a passkey can bypass the maximum session restriction.

The above phenomenon is observed in Spring Security 6.4.3.

## To Reproduce

1. Use spring initializr to generate a new Spring project with “Spring Web” and “Spring Security” dependencies.
2. Configure a `SecurityFilterChain` with `WebAuthnConfigurer` and `SessionManagementConfigurer`, setting `maximumSession(1)`. Then, run the project.
3. Visit http://localhost:8080/webauthn/register in a non-private browser window, sign in as `user` with username + password, and register a passkey.
4. Visit http://localhost:8080/ in an “InPrivate” mode window, sign in as `user` with the registered passkey.
5. Go back to the non-private window and refresh the authenticated page, we can observe that the user is not signed out by Spring Security.

## Expected behavior

After the `user` started the second session in Step 4, he/she should be signed out when going back to the first session to refresh an authenticated page in Step 5.

## Sample

The code of interest is listed below.

```java
@Controller
@SpringBootApplication
public class WebauthnSessionApplication {

    public static void main(String[] args) {
        SpringApplication.run(WebauthnSessionApplication.class, args);
    }

    @Bean
    public SecurityFilterChain securityFilterChain(HttpSecurity http) throws Exception {
        http
                .sessionManagement(session -> session
                        .maximumSessions(1)
                        .expiredUrl("/login?logout"))
                .formLogin(Customizer.withDefaults())
                .webAuthn(webAuthn -> webAuthn
                        .rpId("localhost")
                        .rpName("Demo")
                        .allowedOrigins("http://localhost:8080"))
                .authorizeHttpRequests(authz -> authz
                        .anyRequest().authenticated());
        return http.build();
    }

    @Bean
    public UserDetailsService userDetailsService() {
        return new InMemoryUserDetailsManager(User.withDefaultPasswordEncoder()
                .username("user")
                .password("password")
                .roles("USER")
                .build());
    }
}
```

A runnable project is uploaded as [webauthn-session.tar.gz](https://github.com/user-attachments/files/19039293/webauthn-session.tar.gz).

## BTW

[...]

PR #18624, opened 2026-01-30
Fix session concurrency for OAuth2/OIDC authentication

Gautam-aman:
### Background

Session concurrency control relies on principal equality when resolving
existing sessions. For OAuth2/OIDC logins, the principal is typically an
`Authentication` whose underlying principal (e.g. `DefaultOidcUser`)
uses attribute-based equality.

Since attributes such as `nonce`, `jti`, and `sid` differ per authentication,
session lookups fail and `maximumSessions` becomes ineffective.

### Approach

When the principal is an `Authentication`, this change keys session registry
entries by `Authentication#getName()`, which represents the stable user
identity in Spring Security.

Existing behavior is preserved for non-Authentication principals.

### Tests

Adds a regression test demonstrating correct session resolution when using
distinct `Authentication` instances with the same name.

### DIFF
--- a/core/src/main/java/org/springframework/security/core/session/SessionRegistryImpl.java
+++ b/core/src/main/java/org/springframework/security/core/session/SessionRegistryImpl.java
@@ -32,6 +32,7 @@ import org.jspecify.annotations.Nullable;
 
 import org.springframework.context.ApplicationListener;
 import org.springframework.core.log.LogMessage;
+import org.springframework.security.core.Authentication;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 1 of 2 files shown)


rwinch:
FYI I rebased based off origin/main

jgrandja:
Related gh-18622

ziqin:
~~This may also relate to #16685~~

jgrandja:
@Gautam-aman Thanks for the PR, however, gh-18622 requires a bug fix in `DefaultOidcUser`.

Closing this in favour of eefbb4da6497a34661e6baeea6a55e8e6f8a7702

Issue #18623, opened 2026-01-30
Make authenticationConverter customizable in SpringOpaqueTokenIntrospector.Builder


jzheaux:
See https://github.com/spring-projects/spring-boot/issues/47138

PR #18625, opened 2026-01-31
Add `postProcessor` to `SpringOpaqueTokenIntrospector` Builders

itsmevichu:
Fixes https://github.com/spring-projects/spring-security/issues/18623

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringOpaqueTokenIntrospector.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringOpaqueTokenIntrospector.java
@@ -362,13 +362,13 @@ public class SpringOpaqueTokenIntrospector implements OpaqueTokenIntrospector {
 		 * Adds a {@link Consumer} to customize the {@link SpringOpaqueTokenIntrospector}
 		 * after it is built. This allows for additional configuration that cannot be
 		 * expressed through the builder methods.
-		 * @param customizer the {@link Consumer} to customize the introspector
+		 * @param postProcessor the {@link Consumer} to customize the introspector
 		 * @return the {@link SpringOpaqueTokenIntrospector.Builder}
-		 * @since 7.x.x
+		 * @since 7.1.0
 		 */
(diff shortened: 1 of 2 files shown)


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/introspection/SpringOpaqueTokenIntrospector.java
```diff
@@ -355,6 +357,22 @@ public Builder clientSecret(String clientSecret) {
...
+		/**
+		 * Sets the {@link Converter} used for converting the
+		 * {@link OAuth2TokenIntrospectionClaimAccessor} to an
+		 * {@link OAuth2AuthenticatedPrincipal}.
+		 * @param authenticationConverter the {@link Converter} used for converting to an
+		 * {@link OAuth2AuthenticatedPrincipal}
+		 * @return the {@link SpringOpaqueTokenIntrospector.Builder}
+		 * @since 7.x.x
+		 */
+		public Builder authenticationConverter(
```
jzheaux: Instead of adding just an `authenticationConverter` method, will you please add a `customize` method that takes the `SpringOpaqueTokenIntrospector` itself? Like so:

```java
public Builder postProcessor(Consumer<SpringOpaqueTokenIntrospector> customizer) {
    // ...
}
```

It would have an initial value of `(introspector) -> {}` and the build method would call it after constructing the introspector instance.

In this way, any other setters like `setRequestEntityConverter` can be exposed as well.

itsmevichu: Got it! Thanks for the feedback.

jzheaux:
Hi, @itsmevichu! Thanks for the update! Will you please remove the `authenticationConverter` method since `postProcessor` can serve that purpose?

itsmevichu:
I have removed the `authenticationConverter` method; And FYI, the `@since` version needs to be updated before merging.

jzheaux (approved):

jzheaux:
Thanks again, @itsmevichu for the PR and for your timely updates. This will merge into `main` once the build completes.

itsmevichu:
Thanks! Glad I could contribute.

Issue #18413, opened 2026-01-09
Remove compiler warnings for spring-security-acl


rwinch:

dasog94:
I am going to resolve it

PR #18626, opened 2026-01-31
Fix compiler warnings in spring-security-acl

dasog94:
- Use asSubclass() in AclClassIdUtils to avoid a unchecked cast warning
- Replace raw Map type with Map<?, ?> unbounded wildcard to avoid raw type warnings
- Use ArgumentMatchers to avoid a unchecked cast warning
- Suppress an unavoidable unchecked warning in reflection-based test code

Closes gh-18413


rwinch:
Thanks for the PR @dasog94! This will be merged one the PR build succeeds

Issue #17291, opened 2025-06-18
Change `ActiveDirectoryLdapAuthenticationProvider` to use `LdapClient`


jzheaux:

dineshgupta630:
@jzheaux 
I would love the opportunity to contribute to this issue, if it's still available. Please do let me know.
Thank you so much!

jzheaux:
Thanks, @dineshgupta630, for volunteering! I've assigned the issue to you.

dineshgupta630:
Just a quick note to say I’m on it - planning to raise a PR in the next few days. Thanks!

therepanic:
I apologize for taking on the work despite the issue being assigned - for me it looks dead since no one has worked on it for more than 8 months.

PR #18627, opened 2026-02-01
Change `ActiveDirectoryLdapAuthenticationProvider` to use `LdapClient`

therepanic:
Replaces `SpringSecurityLdapTemplate` with `LdapClient` for user search operations.

Closes: gh-17291

### DIFF
--- a/ldap/src/main/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProvider.java
+++ b/ldap/src/main/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProvider.java
@@ -39,7 +39,12 @@ import org.springframework.core.log.LogMessage;
 import org.springframework.dao.IncorrectResultSizeDataAccessException;
 import org.springframework.ldap.CommunicationException;
 import org.springframework.ldap.core.DirContextOperations;
+import org.springframework.ldap.core.LdapClient;
 import org.springframework.ldap.core.support.DefaultDirObjectFactory;
+import org.springframework.ldap.core.support.SingleContextSource;
+import org.springframework.ldap.query.LdapQuery;
+import org.springframework.ldap.query.LdapQueryBuilder;
+import org.springframework.ldap.query.SearchScope;
 import org.springframework.ldap.support.LdapUtils;
 import org.springframework.security.authentication.AccountExpiredException;
 import org.springframework.security.authentication.BadCredentialsException;
@@ -50,7 +55,6 @@ import org.springframework.security.authentication.LockedException;
 import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.userdetails.UsernameNotFoundException;
-import org.springframework.security.ldap.SpringSecurityLdapTemplate;
 import org.springframework.security.ldap.authentication.AbstractLdapAuthenticationProvider;
 import org.springframework.security.ldap.userdetails.LdapAuthoritiesPopulator;
 import org.springframework.util.Assert;
@@ -96,6 +100,7 @@ import org.springframework.util.StringUtils;
  * @author Luke Taylor
  * @author Rob Winch
  * @author Roman Zabaluev
+ * @author Andrey Litvitski
  * @since 3.1
  */
 public final class ActiveDirectoryLdapAuthenticationProvider extends AbstractLdapAuthenticationProvider {
@@ -299,10 +304,22 @@ public final class ActiveDirectoryLdapAuthenticationProvider extends AbstractLda
 		searchControls.setSearchScope(SearchControls.SUBTREE_SCOPE);
 		String bindPrincipal = createBindPrincipal(username);
 		String searchRoot = (this.rootDn != null) ? this.rootDn : searchRootFromPrincipal(bindPrincipal);
-
+		SingleContextSource contextSource = new SingleContextSource(context);
+		LdapClient ldapClient = LdapClient.builder()
+			.contextSource(contextSource)
+			.defaultSearchControls(() -> searchControls)
+			.ignorePartialResultException(true)
+			.build();
 		try {
-			return SpringSecurityLdapTemplate.searchForSingleEntryInternal(context, searchControls, searchRoot,
-					this.searchFilter, new Object[] { bindPrincipal, username });
+			LdapQuery query = LdapQueryBuilder.query()
+				.base(searchRoot)
+				.searchScope(SearchScope.SUBTREE)
+				.filter(this.searchFilter, bindPrincipal, username);
+			DirContextOperations result = ldapClient.search().query(query).toEntry();
+			if (result == null) {
+				throw new IncorrectResultSizeDataAccessException(1, 0);
+			}
+			return result;
 		}
 		catch (CommunicationException ex) {
 			throw badLdapConnection(ex);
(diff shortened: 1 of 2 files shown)


ldap/src/main/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProvider.java
```diff
@@ -299,10 +303,23 @@ private DirContextOperations searchForUser(DirContext context, String username)
 		searchControls.setSearchScope(SearchControls.SUBTREE_SCOPE);
 		String bindPrincipal = createBindPrincipal(username);
 		String searchRoot = (this.rootDn != null) ? this.rootDn : searchRootFromPrincipal(bindPrincipal);
-
+		SingleContextSource contextSource = new SingleContextSource(context);
+		LdapClient ldapClient = LdapClient.builder()
+			.contextSource(contextSource)
+			.defaultSearchControls(() -> searchControls)
+			.ignorePartialResultException(true)
```
therepanic, line 311: `SpringSecurityLdapTemplate#searchForSingleEntryInternal` behavior also ignores partial result exceptions.

ldap/src/main/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProvider.java
```diff
@@ -316,6 +333,9 @@ private DirContextOperations searchForUser(DirContext context, String username)
 			UsernameNotFoundException userNameNotFoundException = UsernameNotFoundException.fromUsername(username, ex);
 			throw badCredentials(userNameNotFoundException);
 		}
+		catch (org.springframework.ldap.NamingException ex) {
+			throw badCredentials(ex);
+		}
```
therepanic, line 341: This is what you should do, otherwise the tests will fail.

ldap/src/test/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProviderTests.java
```diff
@@ -97,7 +98,7 @@ public void customSearchFilterIsUsedForSuccessfulAuthentication() throws Excepti
 		String customSearchFilter = "(&(objectClass=user)(sAMAccountName={0}))";
 		DirContextAdapter dca = new DirContextAdapter();
 		SearchResult sr = new SearchResult("CN=Joe Jannsen,CN=Users", dca, dca.getAttributes());
-		given(this.ctx.search(any(Name.class), eq(customSearchFilter), any(Object[].class), any(SearchControls.class)))
+		given(this.ctx.search(any(Name.class), any(String.class), any(SearchControls.class)))
```
therepanic: I'm not entirely sure whether this is a breaking change or not. Because when switching to LdapClient, we don't accept any(Object[].class) as the fourth argument, so we don't need to mock it.

therepanic:
I don't quite understand whether this is breaking Change or not, please take a look at https://github.com/spring-projects/spring-security/pull/18627#discussion_r2751346380

rwinch:
FYI I rebased based off origin/main

ldap/src/test/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProviderTests.java
jzheaux: I think this is okay, though I'd recommend that we still use `eq` to test the filter so that the expectations are still the same level of specificity:

```java
String customSearchFilter = "(&(objectClass=user)(sAMAccountName={0}))";
String domain = "mydomain.eu";
String encoded = MessageFormat.format(customSearchFilter, this.joe.getPrincipal() + "@" + domain);
...
given(this.ctx.search(any(Name.class), eq(encoded), any(SearchControls.class)))
		.willReturn(new MockNamingEnumeration(sr));
...
```

ldap/src/main/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProvider.java
jzheaux: I think the reason is because `SpringSecurityLdapTemplate` propagates the `javax.naming` exception while `LdapClient` wraps them. For passivity reasons, let's do something closer to:

```java
if (ex.getCause() instanceof NamingException original) {
    throw original;
}
throw badCredentials(ex);
```

In this way, the original naming exception is propagated like it was before.

ldap/src/main/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProvider.java
```diff
@@ -299,10 +303,23 @@ private DirContextOperations searchForUser(DirContext context, String username)
...
 		try {
-			return SpringSecurityLdapTemplate.searchForSingleEntryInternal(context, searchControls, searchRoot,
-					this.searchFilter, new Object[] { bindPrincipal, username });
+			DirContextOperations result = ldapClient.search()
+				.query(LdapQueryBuilder.query()
+					.base(searchRoot)
+					.searchScope(SearchScope.SUBTREE)
+					.filter(this.searchFilter, bindPrincipal, username))
+				.toEntry();
+			if (result == null) {
```
jzheaux, line 319: This seems like something we should consider as an enhancement to `LdapClient`; a way to indicate that no results is an error. Will you please [file an issue](https://github.com/spring-projects/spring-ldap/issues/new)?

ldap/src/main/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProvider.java
```diff
@@ -299,10 +303,23 @@ private DirContextOperations searchForUser(DirContext context, String username)
...
+		SingleContextSource contextSource = new SingleContextSource(context);
+		LdapClient ldapClient = LdapClient.builder()
+			.contextSource(contextSource)
+			.defaultSearchControls(() -> searchControls)
+			.ignorePartialResultException(true)
+			.build();
 		try {
-			return SpringSecurityLdapTemplate.searchForSingleEntryInternal(context, searchControls, searchRoot,
-					this.searchFilter, new Object[] { bindPrincipal, username });
+			DirContextOperations result = ldapClient.search()
```
jzheaux: This might be slightly more readable as:

```java
LdapQuery query = LdapQueryBuilder.query()
		.base(searchRoot)
		.searchScope(SearchScope.SUBTREE)
		.filter(this.searchFilter, bindPrincipal, username);
DirContextOperations result = ldapClient.search().query(query).toEntry();
```

ldap/src/test/java/org/springframework/security/ldap/authentication/ad/ActiveDirectoryLdapAuthenticationProviderTests.java
therepanic: I decided to follow your advice and added this to the tests where the filters themselves are the subject of testing.

therepanic:
Thanks for the great review. I've updated the tests and made them more readable, as you described here: https://github.com/spring-projects/spring-security/pull/18627#discussion_r2854205066.

Regarding the null result check in https://github.com/spring-projects/spring-security/pull/18627#discussion_r2854194145, I've created an issue: https://github.com/spring-projects/spring-ldap/issues/1404. Whether we add this commit to the spring-ldap repository before or after the update, I'll implement the update there.

jzheaux:
Thanks, @therepanic! This is now merged into `main`.

Issue #18498, opened 2026-01-14
Add InetAddress filtering for client side SSRF protection


rstoyanchev:
An application that makes HTTP requests to user-provided URL's needs to prevent malicious users from directing traffic to forbidden IP addresses (e.g. internal or external, RFC 1918 ranges, etc.) even if the requested hostname appears harmless to protect against SSRF/DNS rebinding issues.

It would be useful for an application to be able to configure restrictions on the target addresses an application is allowed to make requests to. This is best done at the at the level of resolved IP addresses.

HTTP client libraries include a DNS resolver mechanism that can be used to apply checks. What is needed is a common configuration and filtering mechanism to apply such checks to any client.

Spring Security already provides [IpAddressMatcher](https://github.com/spring-projects/spring-security/blob/main/web/src/main/java/org/springframework/security/web/util/matcher/IpAddressMatcher.java) for IP address matching. We can build on that to provide such a mechanism.

The actual code to plug this into specific HTTP client DNS resolvers could exist in Spring Boot, which already helps to configure client libraries.

Note that this has been a collaborative effort with @rwinch and the Google security team, and there are specific code changes in advanced preparation that will be pushed after 7.1 M1 is released next week.

rwinch:
Thanks for the report. I'm closing in favor of https://github.com/spring-projects/spring-security/pull/18634

PR #18634, opened 2026-02-02
Add InetAddressMatcher

rwinch:
Extracts logic for matching `InetAddress` from the `HttpServletRequest`. This allows for general reuse of the API. In particular:

* `IpInetAddressMatcher` now delegates to `InetAddressMatcher`
* `IpAddressServerWebExchangeMatcher` now delegates to `InetAddressMatcher`

Closes gh-18498

### DIFF
--- a/web/src/main/java/org/springframework/security/web/util/matcher/InetAddressMatcher.java
+++ b/web/src/main/java/org/springframework/security/web/util/matcher/InetAddressMatcher.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2004-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.web.util.matcher;
+
+import java.net.InetAddress;
+
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Matches an {@link InetAddress}.
+ *
+ * @author Rossen Stoyanchev
+ * @author Rob Winch
+ * @since 7.1
+ */
+@FunctionalInterface
+public interface InetAddressMatcher {
+
+	/**
+	 * Whether the given address matches.
+	 * @param address the {@link InetAddress} to check (may be {@code null})
+	 * @return {@code true} if the address matches, {@code false} otherwise
+	 */
+	boolean matches(@Nullable InetAddress address);
+
+	/**
+	 * Whether the given address string matches.
+	 * @param address the IP address string to check (may be {@code null})
+	 * @return {@code true} if the address matches, {@code false} otherwise
+	 */
+	default boolean matches(@Nullable String address) {
+		return (address != null) ? matches(InetAddressParser.parseAddress(address)) : false;
+	}
+
+}
--- a/web/src/main/java/org/springframework/security/web/util/matcher/InetAddressMatchers.java
+++ b/web/src/main/java/org/springframework/security/web/util/matcher/InetAddressMatchers.java
@@ -1,0 +1,379 @@
+/*
+ * Copyright 2004-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.web.util.matcher;
+
+import java.net.InetAddress;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.util.Assert;
+
+/**
+ * Factory for creating {@link InetAddressMatcher} instances with various matching
+ * strategies for IP addresses.
+ *
+ * @author Rob Winch
+ * @since 7.1
+ */
+public final class InetAddressMatchers {
+
+	private InetAddressMatchers() {
+	}
+
+	/**
+	 * Creates a new builder for configuring an {@link InetAddressMatcher}.
+	 * @return a new {@link Builder} instance
+	 */
+	public static Builder builder() {
+		return new Builder();
+	}
+
+	/**
+	 * Creates a new builder configured to match external (non-private) IP addresses.
+	 * @return a {@link Builder} configured to match external addresses
+	 */
+	public static Builder matchExternal() {
+		return builder().matchAll(ExternalInetAddressMatcher.getInstance());
+	}
+
+	/**
+	 * Creates a new builder configured to match internal (private) IP addresses.
+	 * <p>
+	 * Internal addresses include loopback addresses (127.0.0.0/8 for IPv4, ::1 for IPv6),
+	 * private IPv4 address ranges (10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16), and IPv6
+	 * Unique Local Addresses (fc00::/7).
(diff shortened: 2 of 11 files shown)


web/src/main/java/org/springframework/security/web/util/matcher/InetAddressMatcher.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+package org.springframework.security.web.util.matcher;
+
+import java.net.InetAddress;
+
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Matches an {@link InetAddress}.
```
rstoyanchev, line 24: After the name change from Filter to Matcher, I think it can be a bit unclear what "match" means. For example, with a deny list matcher, if an address matches, it is denied and the return value would is false. So it is matching and then a decision applied separately.

One option is to reconsider the name, and on that note I don't think "filter" contradicts the package name since the filtering here involves matching first.

Either way the Javadoc could use some elaboration and context around the purpose. For example, it could clarify that it is intended to make decisions to include or exclude an address, and that in that sense it's about matching to a policy more rather than simply matching. Specific implementations could be mentioned as specific examples for illustration.

The Javadoc could describe more concretely where this actually meant to be used, i.e. in DNS resolvers of HTTP client libraries to enforce a security policy, and bring in some of the broader context in the main issue regarding SSRF.

web/src/main/java/org/springframework/security/web/util/matcher/InetAddressMatcher.java
```diff
@@ -0,0 +1,51 @@
...
+	 * @return {@code true} if the address matches, {@code false} otherwise
+	 */
+	boolean matches(InetAddress address);
+
+	/**
+	 * Whether the given address string matches.
+	 * @param address the IP address string to check (may be {@code null})
+	 * @return {@code true} if the address matches, {@code false} otherwise or if
+	 * {@code null}
+	 * @since 7.1
```
rstoyanchev: Is the since tag here necessary? The whole type is new in 7.1.

web/src/main/java/org/springframework/security/web/util/matcher/InetAddressMatcher.java
```diff
@@ -0,0 +1,51 @@
...
+	 */
+	default boolean matches(@Nullable String address) {
+		return (address != null) ? matches(InetAddressParser.parseAddress(address)) : false;
```
rstoyanchev, line 46: Should this allow null input I wonder? Perhaps it won't harm to return false, but I'm not sure there is good reason to expect null within a DNS resolver, and what would one even do with a null address? 

The two methods are inconsistent in this regard, and I note the one accepting InetAddress will throw NPE if null is actually passed in.

I lean towards not allowing null on both and asserting to result in ISA rather than NPE. It would make the contract more clear, and there should be no good reason to pass null.

Another way to resolve this is to consider whether this method is even needed, and leave it out until there are concrete scenarios where it comes up.

web/src/main/java/org/springframework/security/web/util/matcher/IpInetAddressMatcher.java
```diff
@@ -0,0 +1,110 @@
...
+ * CIDR notation (e.g., 192.168.1.0/24).
+ * <p>
+ * Both IPv4 and IPv6 addresses are supported. The matcher can be configured with either a
+ * specific IP address or a subnet using CIDR notation.
+ *
+ * @author Rossen Stoyanchev
+ * @author Gábor Vaspöri
+ * @author Kian Jamali
+ * @author Rob Winch
+ * @since 7.1
```
rstoyanchev, line 43: As a lot of the functionality here came from the long existing `IpAddressMatcher` I would vote to retain and extend the original author tag list. Also, as I did not make any actual updates in InetAddressMatcher, it makes sense to leave my name out.

web/src/main/java/org/springframework/security/web/util/matcher/InetAddressMatcher.java
rwinch: This is good feedback. You are right that the terminology is inconsistent in the Builder API. I've pushed a [commit](https://github.com/spring-projects/spring-security/pull/18634/commits/7d614ac5e51ecdaea2a5c3c6ca0269209be08a28) that I hope helps address this. In summary I've 

* renamed `allowAddresses` to `includeAddresses`
* renamed `denyAddresses` to `excludeAddresses`
* renamed `allowList` to `matchAll`

web/src/main/java/org/springframework/security/web/util/matcher/InetAddressMatcher.java
rwinch: Good catch! I've [removed it](https://github.com/spring-projects/spring-security/pull/18634/commits/59984df9bb129243210b16a1becd3935c4cfee12) and will mark this as resolved.

web/src/main/java/org/springframework/security/web/util/matcher/IpInetAddressMatcher.java
rwinch: Good point. I've [updated](https://github.com/spring-projects/spring-security/pull/18634#:~:text=Verified-,46120d2,-This%20branch%20has) this to reflect the authors from the original IpAddressMatcher.

web/src/main/java/org/springframework/security/web/util/matcher/InetAddressMatcher.java
rwinch: I think that is a good point about consistency. Rather than prohibiting `null` on the `String` variant, I've [updated](https://github.com/spring-projects/spring-security/pull/18634/commits/d85df0c431c24b991780a9c110f0bb736034990e) `matches(InetAddress)` to allow `null` values. This provides consistency with how `IpAddressMatcher` behaves and it provides flexibility of passing in potentially `null` IP addresses into the API without needing to perform a `null` check first.

For example, [ServerHttpRequest.getRemoteAddress()]( https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/server/reactive/ServerHttpRequest.html#getRemoteAddress()) can return `null` and we don't want users to have to check if it might be `null` before passing it into the API. 

I know that this change is in large part to support SSRF, but the API is more broadly applicable to match on IP Addresses since it is used by Spring Security (in `IpAddressMatcher` which allows `null` IP Addresses) on the server side to determine if an IP Address matches then apply the associated authorization rule.

rwinch:
> Looks great, but I did leave a couple of comments and suggestions.

Thanks I've responded inline and posted updates where I think that it makes sense.

> I think the main contract is a good place to provide context and background from a high level perspective, something along the lines of the description of #18498 especially since this is all just a foundational, building block.

I'm going to avoid documenting anything other than what the API does. It is a general purpose IP address matching API that is used by Spring Security to map server side authorization logic. It can be used by users to do anything that involves matching an IP address (e.g. client side firewall rules -- perhaps to prevent SSRF, MFA logic, etc).

web/src/main/java/org/springframework/security/web/util/matcher/InetAddressMatcher.java
KJ202: I believe that some DNS resolvers end up returning a null address under certain conditions (when there is something wrong with the resolution). So I would also say that it's safer to keep handling this case (as it is right now).

web/src/main/java/org/springframework/security/web/util/matcher/InetAddressMatchers.java
```diff
@@ -0,0 +1,374 @@
...
+
+		private InternalInetAddressMatcher() {
+		}
+
+		@Override
+		public boolean matches(@Nullable InetAddress address) {
+			if (address == null) {
+				return false;
+			}
+			if (address.isLoopbackAddress()) {
```
KJ202: Should we do "(address.isLoopbackAddress() || address.isLinkLocalAddress())" instead?

This would include link-local addresses (such as 169.254.169.254) that are commonly targeted for SSRF attacks.

KJ202:
I'd suggest using isSiteLocalAddress() which might be more robust. The current manual check for the 172 range only looks for 172.16, missing IPs from 172.17 through 172.31. The built-in method handles the bitmask correctly and keeps the logic much simpler.

I'm also a bit concerned about the usage IPv4-Mapped IPv6 addresses as a bypass. Standard methods should handle this inherently.

web/src/main/java/org/springframework/security/web/util/matcher/InetAddressParser.java
```diff
@@ -0,0 +1,73 @@
...
+
+	private static boolean isIpAddress(String ipAddress) {
+		// @formatter:off
+		return IPV4.matcher(ipAddress).matches()
+			|| ipAddress.charAt(0) == '['
+			|| ipAddress.charAt(0) == ':'
+			|| Character.digit(ipAddress.charAt(0), 16) != -1
+			&& ipAddress.indexOf(':') > 0;
+		// @formatter:on
+	}
```
KJ202, line 71: Should we defensively check whether the IP address to avoid throwing an exception? (e.g `if (!StringUtils.hasText(ipAddress)) {
        return false;
    }`)

PR #18642, opened 2026-02-04
TokenBasedRememberMeServices documentation snippets should compile

ngocnhan-tran1996:
Related gh-18639


jzheaux:
@ngocnhan-tran1996 thanks for migrating and correcting these samples. This will merge into `7.0.x` once the build completes, and I'll forward port to `main` at that point.

PR #18721, opened 2026-02-12
Add tests for PathPatternRequestMatcher request path caching

coehgns:
## What changed
- Add tests asserting PathPatternRequestMatcher does not leave a parsed request path behind when it parses one
- Add test asserting an already-parsed request path remains available after matching
- Tests cover both cases: request path not pre-parsed vs pre-parsed
## Why
PathPatternRequestMatcher parses and caches the request path when needed and should clear it afterwards to avoid
unexpected side effects on downstream components that rely on ServletRequestPathUtils caching.
## Verification
- ./gradlew :spring-security-web:check


rwinch:
Thank you for your PR! This is now merged into main

Issue #18736, opened 2026-02-13
Cleanup legacy XML test resource names referencing mvc/ant matchers


meliezer:
Some XML test resources still reference `mvc`/`ant` in their filenames, even though the
corresponding `request-matcher` value has been updated to `path`.

This issue tracks a potential cleanup to evaluate whether these XML resources and the
related tests should be removed or updated, as this involves behavior-specific decisions
outside the scope of #18641.

hqqw2h-lgtm:
I'd like to work on this. Will evaluate and clean up the legacy XML test resource names.

PR #18737, opened 2026-02-13
Update request-matcher XML property to support PathPatternRequestMatcher

### LINKED ISSUE #18641: Spring-security.xsd schema validates request-matcher incorrectly
**Describe the bug**
As of Spring Security 7 the `AntPathRequestMatcher `and the `MvcRequestMatcher `have been consolidated into the `PathPatternRequestMatcher`.
Therefore, all occurrences of` request-matcher="ant"` or `request-matcher="mvc"` should be replaced by `request-matcher="path"`, at least if I understand the changes correctly.

However, the spring-security.xsd does not reflect the changes to the MatcherType correctly.
It still validates against the previous value set of **`mvc`, `ant`, `regex `and `ciRegex`**
It should instead validate agains the set **`path`, `regex `and `ciRegex`**

**To Reproduce**
Use the value "path" for the attribute "request-matcher" in some xml configuration.

**Expected behavior**
The value "path" should be accepted as valid value for the attribute request-matcher.

**Sample**

See: https://www.[springframework.org/schema/security/spring-security.xsd](https://www.springframework.org/schema/security/spring-security.xsd)
Or: https://www.[springframework.org/schema/security/spring-security-7.0.xsd](https://www.springframework.org/schema/security/spring-security-7.0.xsd)

Also described here: https://github.com/spring-projects/spring-security/issues/18211

meliezer:
This PR updates the request-matcher schema to reflect the current supported matcher values.

- Adds `path` and removes legacy `mvc`/`ant` from the schema
- Updates XML test resources that use `request-matcher="mvc"`/`"ant"` to use `path`

The `.xsd` file was generated from the updated `.rnc` using:
`./gradlew :spring-security-config:rncToXsd`

Fixes #18641.

Note: XML test resource filenames still reference `mvc`/`ant`. A follow-up cleanup issue #18736 has been opened to evaluate whether these resources and the related tests should be removed or updated, as this is outside the scope of this PR.

### DIFF
--- a/config/src/main/resources/org/springframework/security/config/spring-security-7.0.rnc
+++ b/config/src/main/resources/org/springframework/security/config/spring-security-7.0.rnc
@@ -12,8 +12,8 @@ base64 =
 	## Whether a string should be base64 encoded
 	attribute base64 {xsd:boolean}
 request-matcher =
-	## Defines the strategy use for matching incoming requests. Currently the options are 'mvc' (for Spring MVC matcher), 'ant' (for ant path patterns), 'regex' for regular expressions and 'ciRegex' for case-insensitive regular expressions.
-	attribute request-matcher {"mvc" | "ant" | "regex" | "ciRegex"}
+	## Defines the strategy use for matching incoming requests. Currently the options are 'path' (for PathPatternRequestMatcher), 'regex' for regular expressions and 'ciRegex' for case-insensitive regular expressions.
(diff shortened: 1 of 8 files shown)


jzheaux:
Thanks for the PR, @meliezer! I've rebased it onto `7.0.x` so that folks can have the support earlier.

I've also opened #18807 to investigate the possibility of backporting support for `request-matcher="path"` to `6.5.x` to further simplify migration.

Issue #18405, opened 2026-01-09
Remove compiler warnings


rwinch:
We should remove the compiler warnings in the build. 

## Guidance

If you would like to help, please select a sub-issue and comment on the issue, so that it can be assigned to you. There is no need to wait for a reply if you were the first person to request the issue, then it is yours and you can work on it and send a pull request.

### Finding compiler warnings

You can find the compiler warnings by running all of the compile tasks within the module. Ensure to run compilation for main java and kotlin and test java and kotlin. An easy way to do this is to run the check task for the specific module. For example, with the core module you can run:

```
./gradlew --no-build-cache clean :spring-security-core:check
```

Look through the output for any warnings specific to that module (it may cause other modules to build too)

### Deprecations 

If there are deprecation warnings, then you should use this as guidance.

- If there is a deprecation being used, we should only suppress it if the method itself is deprecated.
- Adding additional deprecations should not be done blindly though (perhaps we should adapt the code to work in a non-deprecated way).
- Tests need to continue testing deprecated methods and thus can have suppress added along with new tests with the replacement method (if it doesn't exist)

### Configure Error on Warning

After you have fixed the compilation warnings, configure the module to error on warning. The first project should create a new [Pre-compiled Script Plugin](https://docs.gradle.org/current/userguide/implementing_gradle_plugins_precompiled.html) at `buildSrc/src/main/groovy/javadoc-warnings-error.gradle` at `buildSrc/src/main/groovy/compile-warnings-error.gradle` that configures the compilers (for main and test, java and kotlin) to fail on warnings. Subsequent projects should just apply this plugin.

### Questions

If you have questions, feel free to reach out on your specific issue. We will update the instructions as we go.

PR #18738, opened 2026-02-15
Remove compiler warnings for spring-security-access

dasog94:
While working this issue, I found that this issue is not in #18405.
But warning logs still exist, so I proceed this issue.

What I did is below.
- added @Suppresswarnings
- refactor codes causing compile warnings
- add test for setAuthorizationManagerFactory that is a alternative to setTrustResolver and setDefaultRolePrefix

About the third, some tests using deprecated functions(setTrustResolver and setDefaultRolePrefix) have to be replaced with setAuthorizationManagerFactory, but there isn't setAuthorizationManagerFactory test.
Because of that, I added new tests even if it is located in core package.
I didn't still replace all the setTrustResolver and setDefaultRolePrefix because it might be used somewhere.

### DIFF
--- a/core/src/test/java/org/springframework/security/access/expression/AbstractSecurityExpressionHandlerTests.java
+++ b/core/src/test/java/org/springframework/security/access/expression/AbstractSecurityExpressionHandlerTests.java
@@ -18,12 +18,16 @@ package org.springframework.security.access.expression;
 
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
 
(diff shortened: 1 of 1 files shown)


rwinch:
Thanks for the PR. Unfortunately the build fails with the submitted changes.

It isn't listed because it is a deprecated module and I'm not too concerned with it as long as it isn't adding to the logs (we can probably just suppress the warnings).

In order for us to merge this the build needs fixed and we'd also need to add the compile-warnings-error plugin. Alternatively, we could suppress the warnings/output since this module is deprecated.

Can you please let me know if you intend to make those changes? If not, I will be closing this PR.

dasog94:
I will fix. Thank you for answer.

dasog94:
It seems it passed checks. It already has the compile-warnings-error plugin.
Is there anything else?

jzheaux:
Thanks, @dasog94! This is now merged into `main`.

Issue #18755, opened 2026-02-18
Specify `charset` in `WWW-Authenticate` for Basic Auth


jbb01:
**Expected Behavior**

RFC7617 (The 'Basic' HTTP Authentication Scheme) defines an optional parameter `charset` in the `WWW-Authenticate` header that indicates the character encoding scheme the server expects the user agent to use when generating "user-pass". Spring Security currently expects the "user-pass" to be UTF-8 encoded by default, but does not include the `charset` parameter in the `WWW-Authenticate` header. The `HttpBasicConfigurer` does not expose a simple way to change the default charset or include the parameter; you have to implement your own `AuthenticationEntryPoint`.

**Context**

Some user agents, most notably the JDK's own `HttpClient`, rely on this parameter to choose the correct encoding and fall back to a different default charset (ISO-8859-1 in case of the `HttpClient`) which causes problems with passwords containing special characters.

The JDK's `HttpClient` doesn't even allow configuration of the charset when using an `Authenticator` instead of manually setting the `Authorization` header, which makes it incompatible with a standard Spring Security Basic Auth configuration.

therepanic:
Thanks for reporting this, Jonah @jbb01.

It's really helpful that you provided an example of the problem this feature will solve. Thanks a lot. I've attached the PR #18760, you can check it out.

PR #18760, opened 2026-02-18
Specify charset in `WWW-Authenticate` for Basic Auth

therepanic:
In this commit, we add support for the charset from [RFC-7617](https://datatracker.ietf.org/doc/html/rfc7617), which definitely solves the problem when the client does not know what charset we are parsing with.

Closes: gh-18755

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationEntryPoint.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationEntryPoint.java
@@ -80,6 +80,7 @@ public class BasicAuthenticationEntryPoint implements AuthenticationEntryPoint, 
 	 * the charset attribute from the header. As per RFC 7617, only UTF-8 is permitted.
 	 * @param charset the charset to use ({@link StandardCharsets#UTF_8} is the only
 	 * accepted value), or {@code null} to remove the charset attribute
+	 * @since 7.1
 	 */
 	public void setCharset(@Nullable Charset charset) {
 		Assert.isTrue(charset == null || StandardCharsets.UTF_8.equals(charset),
--- a/docs/modules/ROOT/pages/whats-new.adoc
+++ b/docs/modules/ROOT/pages/whats-new.adoc
@@ -1,4 +1,7 @@
 [[new]]
 = What's New in Spring Security 7.1
 
+== Web
+
 * https://github.com/spring-projects/spring-security/pull/18634[gh-18634] - Added javadoc:org.springframework.security.web.util.matcher.InetAddressMatcher[]
+* https://github.com/spring-projects/spring-security/issues/18755[gh-18755] - Include `charset` in `WWW-Authenticate` header


web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilter.java
```diff
@@ -353,7 +354,11 @@ public void setCredentialsCharset(String credentialsCharset) {
 		Assert.hasText(credentialsCharset, "credentialsCharset cannot be null or empty");
 		this.credentialsCharset = credentialsCharset;
 		if (this.authenticationConverter instanceof BasicAuthenticationConverter basicAuthenticationConverter) {
-			basicAuthenticationConverter.setCredentialsCharset(Charset.forName(credentialsCharset));
+			Charset charset = Charset.forName(credentialsCharset);
+			basicAuthenticationConverter.setCredentialsCharset(charset);
+			if (this.authenticationEntryPoint instanceof BasicAuthenticationEntryPoint basicAuthenticationEntryPoint) {
+				basicAuthenticationEntryPoint.setCharset(charset);
+			}
 		}
```
therepanic, line 357: BTW, I think this is a cool solution for charset forwarding, just my 50 cents.

web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationEntryPoint.java
```diff
@@ -53,7 +58,8 @@ public void afterPropertiesSet() {
 	@Override
 	public void commence(HttpServletRequest request, HttpServletResponse response,
 			AuthenticationException authException) throws IOException {
-		response.setHeader("WWW-Authenticate", "Basic realm=\"" + this.realmName + "\"");
+		String header = "Basic realm=\"" + this.realmName + "\", charset=\"" + this.charset.name() + "\"";
+		response.setHeader("WWW-Authenticate", header);
```
therepanic: [RFC-7617](https://datatracker.ietf.org/doc/html/rfc7617) states that charset is recommended but not mandatory.

This essentially leads us to different conclusions. Ideally, we should make it nullable to comply with this RFC. However, on the other hand, I see no reason to go with this convention. In what I believe is an elegant solution I've presented, unfortunately, we can't pass a nullable charset because `BasicAuthenticationFilter#setCredentialsCharset` doesn't accept a nullable value, and I certainly don't think we should change the contract for this reason. I'd like to hear other opinions on this matter, in case anyone needs it.

web/src/test/java/org/springframework/security/web/authentication/www/BasicAuthenticationEntryPointTests.java
```diff
@@ -77,7 +78,18 @@ void commenceWhenResponseHasHeaderThenOverride() throws IOException {
 		ep.commence(request, response, new DisabledException("Disabled"));
 		List<String> headers = response.getHeaders("WWW-Authenticate");
 		assertThat(headers).hasSize(1);
-		assertThat(headers.get(0)).isEqualTo("Basic realm=\"hello\"");
+		assertThat(headers.get(0)).isEqualTo("Basic realm=\"hello\", charset=\"UTF-8\"");
```
therepanic, line 82: I believe this isn't a breaking change. Clients will ignore the header parameters, and this is just a convention of our test.

therepanic:
It turned out that some tests also required modifications and the addition of a charset. I wrote about this here: https://github.com/spring-projects/spring-security/pull/18760#discussion_r2823934023 and here: https://github.com/spring-projects/spring-security/pull/18760#discussion_r2823916764.

And over time, I'm starting to think that maybe the tests should still pass in this case? Then we'll have to create some kind of toggle switch, so we can enable charset additions. I'd like to know the best way to do this to fit the project.

I'm not sure what's best.

web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationFilter.java
jzheaux: Thanks for the creativity here. I think my hestitation is that this method is about setting the charset for the credentials. 

While I imagine it quite common that applications will be happy to have the charset for the credentials as well as the `WWW-Authenticate` header the same, I also imagine the surprise of learning that when I call `setCredentialsCharset`, it also sets charsets for non-credential-related attributes.

Instead, could we leave this out of the setter for now and rely on applications configuring the `BasicAuthenticationEntryPoint`? I think this is fine since I think the main reason applications will have to change it will be to set it to null to restore earlier behavior.

web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationEntryPoint.java
```diff
@@ -40,11 +42,14 @@
  * authorized, causing it to prompt the user to login again.
  *
  * @author Ben Alex
+ * @author Andrey Litvitski
  */
 public class BasicAuthenticationEntryPoint implements AuthenticationEntryPoint, InitializingBean {
 
 	private @Nullable String realmName;
 
+	private Charset charset = StandardCharsets.UTF_8;
```
jzheaux: Since it's optional, let's mark it and its setter as `@Nullable`, check for null in `commence`, and add a test that confirms that it can be set to `null`.

web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationEntryPoint.java
```diff
@@ -65,4 +71,12 @@ public void setRealmName(String realmName) {
 		this.realmName = realmName;
 	}
 
+	public void setCharset(Charset charset) {
+		this.charset = charset;
+	}
+
+	public Charset getCharset() {
```
jzheaux: Let's please leave the getter our for now

web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationEntryPoint.java
```diff
@@ -65,4 +71,12 @@ public void setRealmName(String realmName) {
 		this.realmName = realmName;
 	}
 
+	public void setCharset(Charset charset) {
```
jzheaux: Please add a JavaDoc that indicates what the default value is. It should also explain that an application can set the value to `null` to remove the `charset` header attribute.

therepanic:
Thanks for review. I followed your suggestions and pushed the change. I agree with your suggestions and followed them, pushing the change.

web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationEntryPoint.java
```diff
@@ -65,4 +74,14 @@ public void setRealmName(String realmName) {
 		this.realmName = realmName;
 	}
 
+	/**
+	 * Sets the charset to include in the {@code WWW-Authenticate} response header. By
+	 * default, it is set to {@link StandardCharsets#UTF_8}. Set to {@code null} to omit
+	 * the charset attribute from the header.
+	 * @param charset the charset to use, or {@code null} to remove the charset attribute
+	 */
+	public void setCharset(@Nullable Charset charset) {
```
jzheaux, line 85: Since RFC 7617 only allows UTF-8, let's assert that `charset` is either `null` or UTF-8. We can always relax that requirement in the future.

web/src/main/java/org/springframework/security/web/authentication/www/BasicAuthenticationEntryPoint.java
```diff
@@ -53,7 +58,11 @@ public void afterPropertiesSet() {
 	@Override
 	public void commence(HttpServletRequest request, HttpServletResponse response,
 			AuthenticationException authException) throws IOException {
-		response.setHeader("WWW-Authenticate", "Basic realm=\"" + this.realmName + "\"");
+		String header = "Basic realm=\"" + this.realmName;
```
jzheaux: This seems like a bug. If `this.charset` is `null`, then `header` will be missing a trailing quote.

web/src/test/java/org/springframework/security/web/authentication/www/BasicAuthenticationEntryPointTests.java
```diff
@@ -62,7 +64,7 @@ public void testNormalOperation() throws Exception {
 		ep.commence(request, response, new DisabledException("These are the jokes kid"));
 		assertThat(response.getStatus()).isEqualTo(401);
 		assertThat(response.getErrorMessage()).isEqualTo(HttpStatus.UNAUTHORIZED.getReasonPhrase());
-		assertThat(response.getHeader("WWW-Authenticate")).isEqualTo("Basic realm=\"hello\"");
+		assertThat(response.getHeader("WWW-Authenticate")).isEqualTo("Basic realm=\"hello\", charset=\"UTF-8\"");
```
jzheaux, line 67: Can you please add a test for `setCharset(null)`?

jzheaux:
Thanks for the updates, @therepanic! I've also polished the documentation a bit, including mentioning this in `whats-new.adoc` so folks can be aware of the change. This will merge once the build passes.

jzheaux (approved):

Issue #18698, opened 2026-02-07
Kerberos fails with Spring Security 7.0.2 despite being integrated


busoco-sjb:
**Describe the bug**
According to "What's new in Spring Security 7", "The Spring Security Kerberos Extension is now part of Spring Security"
However, Kerberos authentication fails with 

```
java.lang.ClassNotFoundException: org.springframework.security.crypto.codec.Base64
 at org.springframework.security.kerberos.web.authentication.SpnegoAuthenticationProcessingFilter.doFilterInternal(SpnegoAuthenticationProcessingFilter.java:147)

```

**To Reproduce**
See https://github.com/spring-projects/spring-security-kerberos/issues/338 and https://github.com/spring-projects/spring-security-kerberos/pull/340

**Expected behavior**
Spring Security Kerberos, if integrated with Spring Security 7, should not use classes which have been removed.

**Sample**

See links under "To Reproduce"

dodgex:
Hey @busoco-sjb I just had the Idea to create an issue here similar to yours trying to bring attention to the https://github.com/spring-projects/spring-security-kerberos repository. your note, that it is now part of Spring Security led me to this finding: https://github.com/spring-projects/spring-security/tree/main/kerberos

The whole project/code of `spring-security-kerberos` is now part of this repository. Also, the artifacts with fixed stuff have changed thier groupId from `org.springframework.security.kerberos` to `org.springframework.security`!

https://mvnrepository.com/artifact/org.springframework.security.kerberos/spring-security-kerberos-web has the old 2.x releases with the bad code using the removed `Base64` class.
https://mvnrepository.com/artifact/org.springframework.security/spring-security-kerberos-web has the new 7.x releases with the code from this repository, that should contain the fixes. 🥳 

So chaning jar coordinates from groupId `org.springframework.security.kerberos` to `org.springframework.security` and version to the matching `7.x` (or leaving it to the dependency management when using spring boot)

busoco-sjb:
Thank you @dodgex!

The change in the dependency coordinates doesn't seem to be documented anywhere except as a very vague hint in [What's New][https://docs.spring.io/spring-security/reference/whats-new.html] with "The [Spring Security Kerberos Extension](https://github.com/spring-projects/spring-security-kerberos) is now part of Spring Security."
Simply removing the Spring Security Kerberos extension dependencies fails the build, because Kerberos is *not* fully part of Spring Security, and still requires extra dependencies.

dodgex:
[...]

PR #18773, opened 2026-02-20
Document the change in dependency coordinates with Spring Security 7

busoco-sjb:
Include the updated dependency coordinates for Spring Security Kerberos, so that people upgrading to Spring Security 7 realise that they still have to bring in extra dependencies, but with a different group ID.

Closes gh-18698


rwinch:
Thanks for the PR! I've merged this into 7.0.x and main along with updates to the migration documentation https://github.com/rwinch/spring-security/commit/fec988c82dcc8614912734f5828f9d1bbba016cd

Issue #18781, opened 2026-02-23
Refactor: Improve logic rigor for upgradeEncoding in DaoAuthenticationProvider.createSuccessAuthentication


beijixing1745:
in spring security core 7.0.3 class  DaoAuthenticationProvider  method createSuccessAuthentication upgradeEncoding logic，default condition will be always true，some body will not want to updatePassword。

```
public DaoAuthenticationProvider(UserDetailsService userDetailsService) {
        this.userDetailsPasswordService = UserDetailsPasswordService.NOOP;
        Assert.notNull(userDetailsService, "userDetailsService cannot be null");
        this.userDetailsService = userDetailsService;
    }
......
protected Authentication createSuccessAuthentication(Object principal, Authentication authentication, UserDetails user) {
        Assert.notNull(authentication.getCredentials(), "Authentication.getCredentials() cannot be null");
        String presentedPassword = authentication.getCredentials().toString();
        boolean isPasswordCompromised = this.compromisedPasswordChecker != null && this.compromisedPasswordChecker.check(presentedPassword).isCompromised();
        if (isPasswordCompromised) {
            throw new CompromisedPasswordException("The provided password is compromised, please change your password");
        } else {
            String existingEncodedPassword = user.getPassword();
            boolean upgradeEncoding = existingEncodedPassword != null && this.userDetailsPasswordService != null && ((PasswordEncoder)this.passwordEncoder.get()).upgradeEncoding(existingEncodedPassword);
            if (upgradeEncoding) {
                String newPassword = ((PasswordEncoder)this.passwordEncoder.get()).encode(presentedPassword);
                user = this.userDetailsPasswordService.updatePassword(user, newPassword);
            }

            return super.createSuccessAuthentication(principal, authentication, user);
        }
    }
```

boolean upgradeEncoding = existingEncodedPassword != null && this.userDetailsPasswordService != null && ((PasswordEncoder)this.passwordEncoder.get()).upgradeEncoding(existingEncodedPassword);

default upgradeEncoding  will be always true?

Improvement:

```
boolean upgradeEncoding = existingEncodedPassword != null && !Objects.equals(this.userDetailsPasswordService,UserDetailsPasswordService.NOOP) && ((PasswordEncoder)this.passwordEncoder.get()).upgradeEncoding(existingEncodedPassword);
```

therepanic:
Thanks for the report, @beijixing1745. 

[...]

PR #18788, opened 2026-02-23
Restore upgradeEncoding condition in DaoAuthenticationProvider

therepanic:
After adding jspecify support in the module that contains the DaoAuthenticationProvider class, we actually changed the contract logic, which is a good thing, and this commit fixes it.

Closes: gh-18781

### DIFF
--- a/core/src/main/java/org/springframework/security/authentication/dao/DaoAuthenticationProvider.java
+++ b/core/src/main/java/org/springframework/security/authentication/dao/DaoAuthenticationProvider.java
@@ -16,6 +16,7 @@
 
 package org.springframework.security.authentication.dao;
 
+import java.util.Objects;
 import java.util.function.Supplier;
 
 import org.jspecify.annotations.Nullable;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/springframework/security/authentication/dao/DaoAuthenticationProvider.java
```diff
@@ -131,7 +133,8 @@ protected Authentication createSuccessAuthentication(Object principal, Authentic
 			throw new CompromisedPasswordException("The provided password is compromised, please change your password");
 		}
 		String existingEncodedPassword = user.getPassword();
-		boolean upgradeEncoding = existingEncodedPassword != null && this.userDetailsPasswordService != null
+		boolean upgradeEncoding = existingEncodedPassword != null
+				&& !Objects.equals(this.userDetailsPasswordService, UserDetailsPasswordService.NOOP)
```
therepanic, line 137: Since this problem, I believe, is directly related to semantics, I assume we don’t need tests for this?

jzheaux:
Thanks for the PR, @therepanic! I rebased your PR to `7.0.x` and then also forward-ported it to `main`.

Issue #18259, opened 2025-12-03
PermissionEvaluator targetDomainObject should be @Nullable


OrangeDog:
**Describe the bug**
Upgrading from 6.x to 7.0, the `org.springframework.security.access` package has now been `@NullMarked`.

The documentation for `PermissionEvaluator#hasPermission(..., Object targetDomainObject, ...)` says:

> **May be null** in which case implementations should return false, as the null condition can be checked explicitly in the expression.

However, it is not marked `@Nullable`, so cases where it is `null` will cause analysis errors.

rwinch:
Thank you for creating the issue @OrangeDog I've closed this in favor of the pull request which is now merged into main

OrangeDog:
That pull request was fixing this issue. That's why it says "Closes: gh-18259". It's not a duplicate or "in favor of"...

rwinch:
Reopening since the [pull request failed after being rebased](https://github.com/spring-projects/spring-security/pull/18260#issuecomment-3953031980).

rwinch:
Thanks to @therepanic this is now resolved in main via gh-18796

workmanw:
Is there any chance of this showing up in a 7.0.x patch release? It's currently a big upgrade blocker for our teams so I was just curious to get a line of sight on when this might be available in a GA release.

PR #18260, opened 2025-12-03
Mark `targetDomainObject` as `@Nullable` in `PermissionEvaluator`

therepanic:
Currently, there are other places where `PermissionEvaluator` is inherited and methods are overridden. However, these places aren't yet covered by jspecify. I expect that in the future, when these places are covered by jspecify, we'll ensure that `@Nullable` is used (the build simply won't pass).

Closes: gh-18259

PR #18796, opened 2026-02-24
Mark `targetDomainObject` as `@Nullable` in `PermissionEvaluator`

therepanic:
Second attempt. OG PR: https://github.com/spring-projects/spring-security/pull/18260

Closes: gh-18259

### DIFF
--- a/acl/src/main/java/org/springframework/security/acls/AclPermissionEvaluator.java
+++ b/acl/src/main/java/org/springframework/security/acls/AclPermissionEvaluator.java
@@ -23,6 +23,7 @@ import java.util.Locale;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.jspecify.annotations.Nullable;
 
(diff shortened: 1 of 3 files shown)


acl/src/main/java/org/springframework/security/acls/AclPermissionEvaluator.java
```diff
@@ -73,7 +74,7 @@ public AclPermissionEvaluator(AclService aclService) {
 	 * be overridden using a null check in the expression itself).
 	 */
 	@Override
-	public boolean hasPermission(Authentication authentication, Object domainObject, Object permission) {
+	public boolean hasPermission(Authentication authentication, @Nullable Object domainObject, Object permission) {
 		if (domainObject == null) {
 			return false;
 		}
```
therepanic, line 80: I believe the problem with the original PR was that for some reason I forgot to make the change in `AclPermissionEvaluator` as well, but the tests passed because while I was making the PR, the acl module did not have jspecify support added, however, after the PR it was added.

PR #18806, opened 2026-02-26
Fallback defaultTargetUrl if refererHeader is empty

### LINKED ISSUE #18805: AbstractAuthenticationTargetUrlRequestHandler.determineTargetUrl does not fallback to default url
spring-security-web-6.5.5.jar

**Describe the bug**
Javadoc states :
_If the useReferer property is set, the "Referer" HTTP header value will be used, if present.
As a fallback option, the defaultTargetUrl value will be used._
In current implementation if useReferer is true it always returns header value without any fallback as described in javadoc.

in version spring-security-web-5.8.19.jar we had:
`if (this.useReferer && !StringUtils.hasLength(targetUrl)) {
    targetUrl = request.getHeader("Referer");
    if (this.logger.isTraceEnabled()) {
        this.logger.trace(LogMessage.format("Using url %s from Referer header", targetUrl));
    }
}

if (!StringUtils.hasText(targetUrl)) {
    targetUrl = this.defaultTargetUrl;
    if (this.logger.isTraceEnabled()) {
        this.logger.trace(LogMessage.format("Using default url %s", targetUrl));
    }
}

return targetUrl;`

in spring-security-web-6.5.5.jar we have:
`if (this.useReferer) {
			trace("Using url %s from Referer header", request.getHeader("Referer"));
			return request.getHeader("Referer");
		}
return this.defaultTargetUrl`

In Spring Security 5.8.x useReferer=true falls back to defaultTargetUrl when Referer header is null or empty.
In 6.5.5 determineTargetUrl returns null if Referer header is missing, causing redirect failures.

**To Reproduce**
1. set useReferer=true
2. request.getHeader("Referer") should be set as null
3. set defaultTargetUrl="/sampleUrl"

**Expected behavior**
"/sampleUrl" is returned

**Current behavior**
null is returned

ngocnhan-tran1996:
Closes gh-18805

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/AbstractAuthenticationTargetUrlRequestHandler.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/AbstractAuthenticationTargetUrlRequestHandler.java
@@ -112,9 +112,11 @@ public abstract class AbstractAuthenticationTargetUrlRequestHandler {
 			trace("Using url %s from request parameter %s", targetUrlParameterValue, this.targetUrlParameter);
 			return targetUrlParameterValue;
 		}
-		if (this.useReferer) {
-			trace("Using url %s from Referer header", request.getHeader("Referer"));
-			return request.getHeader("Referer");
+
+		String refererHeader = request.getHeader("Referer");
+		if (this.useReferer && StringUtils.hasText(refererHeader)) {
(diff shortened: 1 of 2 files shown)


jzheaux:
Thanks for this fix, @ngocnhan-tran1996! This is now merged into `6.5.x`, `7.0.x`, and `main`.

Issue #12650, opened 2023-02-09
Programmatic way to use expression-based authorization manager for websockets


lrozenblyum:
**Expected Behavior**
When we configure `MessageMatcherDelegatingAuthorizationManager` it should be possible to provide an instance of a class that supports expression-based authorization to the method `org.springframework.security.messaging.access.intercept.MessageMatcherDelegatingAuthorizationManager.Builder.Constraint.access(AuthorizationManager<MessageAuthorizationContext<?>>)`

In non-websocket world this is supported by `org.springframework.security.web.access.expression.WebExpressionAuthorizationManager.WebExpressionAuthorizationManager(String)`

**Current Behavior**
WebSocket class that seems to be facilitating the same task (`org.springframework.security.config.websocket.WebSocketMessageBrokerSecurityBeanDefinitionParser.ExpressionBasedAuthorizationManager`) is private and looks to be used for XML config only

**Context**
Spring Security 5.8.1.
The issue is affecting us during migration from older Spring Security which supported expression-based access control via the method `org.springframework.security.config.annotation.web.messaging.MessageSecurityMetadataSourceRegistry.Constraint.access(String)`

i-filaliawb:
same problem with us.

can not find any way to use expression-based access control for destination.

reckart:
Has anybody figured out how to do expression-based authentication with WebSocket in Spring 6?

jzheaux:
Hi, @lrozenblyum, thanks for reaching out. The main reason this does not exist in the `MessageMatcherDelegatingAuthorizationManager` is that we'd like to encourage people to move away from SpEL in their web authorization rules and use a concrete implementation of `AuthorizationManager` instead. The primary benefit is that it is independently testable.

That said, I think there's value in adding a messaging equivalent to `WebExpressionAuthorizationManager` to simplify migration.

Can someone submit a PR?

In the meantime, you are welcome to use the following implementation:

```java
public final class MessageExpressionAuthorizationManager implements AuthorizationManager<MessageAuthorizationContext<?>> {

	private SecurityExpressionHandler<Message<?>> expressionHandler = new DefaultMessageSecurityExpressionHandler();

	private Expression expression;

	public MessageExpressionAuthorizationManager(String expressionString) {
		Assert.hasText(expressionString, "expressionString cannot be empty");
		this.expression = this.expressionHandler.getExpressionParser().parseExpression(expressionString);
	}

	@Override
[...]

PR #18813, opened 2026-02-27
Add MessageExpressionAuthorizationManager

wonderfulrosemari:
Closes gh-12650

This PR promotes expression-based message authorization to a public API by
introducing `MessageExpressionAuthorizationManager` in the messaging module.

Changes include:
- adding `MessageExpressionAuthorizationManager` and unit tests
- updating the WebSocket XML parser to use the new public manager instead of
  the private inner class
- updating WebSocket migration docs to reference the built-in manager

### DIFF
--- a/messaging/src/main/java/org/springframework/security/messaging/access/expression/MessageExpressionAuthorizationManager.java
+++ b/messaging/src/main/java/org/springframework/security/messaging/access/expression/MessageExpressionAuthorizationManager.java
@@ -1,0 +1,137 @@
+/*
+ * Copyright 2004-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.messaging.access.expression;
+
+import java.util.function.Supplier;
+
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.expression.EvaluationContext;
+import org.springframework.expression.Expression;
+import org.springframework.security.access.expression.ExpressionUtils;
+import org.springframework.security.access.expression.SecurityExpressionHandler;
+import org.springframework.security.authorization.AuthorizationManager;
+import org.springframework.security.authorization.AuthorizationResult;
+import org.springframework.security.authorization.ExpressionAuthorizationDecision;
+import org.springframework.security.core.Authentication;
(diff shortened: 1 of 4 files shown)


messaging/src/main/java/org/springframework/security/messaging/access/expression/MessageExpressionAuthorizationManager.java
```diff
@@ -0,0 +1,104 @@
...
+
+	/**
+	 * Sets the {@link SecurityExpressionHandler} to be used. The default is
+	 * {@link MessageAuthorizationContextSecurityExpressionHandler}.
+	 * @param expressionHandler the {@link SecurityExpressionHandler} to use
+	 * @deprecated Use
+	 * {@link #MessageExpressionAuthorizationManager(SecurityExpressionHandler, String)}
+	 */
+	@Deprecated
+	public void setExpressionHandler(SecurityExpressionHandler<MessageAuthorizationContext<?>> expressionHandler) {
```
jzheaux: Let's please leave this method out so that we don't add `@Deprecated` features.

messaging/src/main/java/org/springframework/security/messaging/access/expression/MessageExpressionAuthorizationManager.java
```diff
@@ -0,0 +1,104 @@
...
+		Assert.hasText(expressionString, "expressionString cannot be empty");
+		this.expression = this.expressionHandler.getExpressionParser().parseExpression(expressionString);
+	}
+
+	/**
+	 * Creates an instance with the given expression handler.
+	 * @param expressionHandler the expression handler to use
+	 * @param expressionString the raw expression string to parse
+	 */
+	public MessageExpressionAuthorizationManager(
```
jzheaux: Instead of a public constructor, please take a look at `WebExpressionAuthorizationManager#withSecurityExpressionHandler` and its accompanying builder. I think it would be nice to follow that practice to make it simpler to create many authorization managers while only specifying the expression handler once.

jzheaux (changes requested):
Hi, @wonderfulrosemari! Thanks for this PR. I've left some feedback inline.

messaging/src/main/java/org/springframework/security/messaging/access/expression/MessageExpressionAuthorizationManager.java
wonderfulrosemari: Thanks for the suggestion. I replaced the public constructor with a builder-style API (withSecurityExpressionHandler(...).expression(...)) and updated the XML parser accordingly in 2dd897815d.

jzheaux:
Thanks for the updates, @wonderfulrosemari. In preparation for merging, will you please squash your commits and change the title of your new commit to "Add MessageExpressionAuthorizationManager"?

jzheaux:
Thanks, @wonderfulrosemari! Sorry for a miscommunication, it is still helpful to have the `Closes gh-12650` in the commit message. So, like this:

```
Add MessageExpressionAuthorizationManager

Closes gh-12650

Signed-off-by: wonderfulrosemari <whwlsgur1419@naver.com>
```

This helps with future research when trying to track down why a specific change to the codebase was made. Will you please change your commit message to include the above?

PR #18819, opened 2026-02-28
Use attributes in Antora to replace the original links

ngocnhan-tran1996:


jzheaux:
Thanks, @ngocnhan-tran1996! I build your changes with `./gradlew antora` and used `linkchecker` locally to test the links.

This will merge once the build passes.

PR #18820, opened 2026-02-28
Remove compiler warnings in spring-security-web

### LINKED ISSUE #18441: Remove compiler warnings for spring-security-web
See gh-18405 for more details.

jkuhel:
This PR resolves the compiler warnings in the `spring-security-web` module.

Some of these warnings were straight forward changes:
- `ServerOneTimeTokenAuthenticationConverter`:  Replace deprecated call `OneTimeTokenAuthenticationToken.unauthenticated` with `OneTimeTokenAuthenticationToken(String token)` 
- Update `HttpMessageConverterAuthenticationSuccessHandler` to replace deprecated `MappingJackson2HttpMessageConverter` with `JacksonJsonHttpMessageConverter`
- Replace updated `OneTimeTokenAuthenticationConverter` to use non-deprecated `OneTimeTokenAuthenticationToken` constructor
- update tests classes `DelegatingAuthenticationEntryPointTests` / `OneTimeTokenAuthenticationFilterTests` to remove use of deprecated methods
- Enabled `compile-warnings-error` for `spring-security-web`

### `JdbcTokenRepositoryImpl` changes

@rwinch The more involved change was in `JdbcTokenRepositoryImpl` with the extension of deprecated `JdbcDaoSupport`. What would be the best way to handle this?

To start off for now, I opted to try an implementation that removes `extends JdbcDaoSupport`. However, this _could_ be a breaking change for anyone that has extended `JdbcTokenRepositoryImpl` directly via the no-arg constructor with `setDataSource()` pattern. 

If it would be best to not introduce this change or take a different approach, let me know.

Here is a summary of changes I made:
- Added a `DataSource` constructor parameter (replaces inherited `setDataSource()`) and used to create `JdbcTemplate` in the class
- Added own `logger` field (previously inherited from `JdbcDaoSupport`)
- preserved the `setCreateTableOnStartup(true)` by implementing `InitializingBean` to preserve the `afterPropertiesSet()` call to `initDao()` lifecycle
- Updated `RememberMeBeanDefinitionParser` to accommodate changes from property injection to constructor injection of the `DataSource`
- Updated `JdbcTokenRepositoryImplTests` accordingly

Closes #18441

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/rememberme/JdbcTokenRepositoryImpl.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/rememberme/JdbcTokenRepositoryImpl.java
@@ -35,6 +35,7 @@ import org.springframework.jdbc.core.support.JdbcDaoSupport;
  * @author Luke Taylor
  * @since 2.0
  */
+@SuppressWarnings("removal")
 public class JdbcTokenRepositoryImpl extends JdbcDaoSupport implements PersistentTokenRepository {
 
 	/** Default SQL for creating the database table to store the tokens */
--- a/web/spring-security-web.gradle
+++ b/web/spring-security-web.gradle
@@ -1,6 +1,7 @@
 plugins {
 	id 'io.spring.convention.spring-module'
 	id 'security-nullability'
+	id 'compile-warnings-error'
 	id 'javadoc-warnings-error'
 	id 'test-compile-target-jdk25'
 }
--- a/web/src/main/java/org/springframework/security/web/authentication/ott/OneTimeTokenAuthenticationConverter.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/ott/OneTimeTokenAuthenticationConverter.java
@@ -46,7 +46,7 @@ public class OneTimeTokenAuthenticationConverter implements AuthenticationConver
 			this.logger.debug("No token found in request");
 			return null;
 		}
-		return OneTimeTokenAuthenticationToken.unauthenticated(token);
+		return new OneTimeTokenAuthenticationToken(token);
 	}
 
 }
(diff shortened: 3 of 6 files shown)


web/src/main/java/org/springframework/security/web/authentication/rememberme/JdbcTokenRepositoryImpl.java
```diff
@@ -20,22 +20,29 @@
...
-import org.springframework.jdbc.core.support.JdbcDaoSupport;
+import org.springframework.util.Assert;
 
 /**
  * JDBC based persistent login token repository implementation.
  *
  * @author Luke Taylor
  * @since 2.0
  */
-public class JdbcTokenRepositoryImpl extends JdbcDaoSupport implements PersistentTokenRepository {
```
jzheaux, line 38: Thanks for your analysis here. It indeed would be a breaking change and so we'll need to leave the `extends` clause. 

I believe we'll need a new implementation so that folks have a class they can move to before `JdbcDaoSupport` is removed. That said, that's something that could benefit from a separate review process, so I've added https://github.com/spring-projects/spring-security/issues/18982 to address it. This class here can delegate to the new implementation.

For this PR, let's please suppress this particular warning.

jkuhel: Thank you @jzheaux! Sounds like a great call! I'll back these changes out to leave the `extends` clause. 

I follow up in https://github.com/spring-projects/spring-security/issues/18982 and work on removing JdbcDaoSupport in a separate PR.

jzheaux:
Thanks for the updates, @jkuhel! This will merge into `main` once the build completes

Issue #8843, opened 2020-07-17
JspAuthorizeTag cannot be used if spring-security configurations (beans) is not put in ROOT context (XML config).


bboyz269:
**Describe the bug**
I have a empty root context and an app (child) context contains all my beans. I have a separated xml file for spring security config, imported in child context.
Received the following error when using JSPAuthorizeTag `<sec:authorize>`
```
javax.servlet.ServletException: javax.servlet.jsp.JspException: java.io.IOException: No visible WebSecurityExpressionHandler instance could be found in the application context. There must be at least one in order to support expressions in JSP 'authorize' tags.
	org.apache.jasper.runtime.PageContextImpl.doHandlePageException(PageContextImpl.java:907)
	org.apache.jasper.runtime.PageContextImpl.handlePageException(PageContextImpl.java:840)
```

**To Reproduce**
As above
Create an app (Spring mvc) with empty root context and child context, contains spring-security configurations.
Error happens if you try to use `<sec:authorize>` tag in JSP.

**Expected behavior**
If it's not by design that spring security configs must be in root context. I expect to use jsp authorize tag normally.

**Sample**
I don't have a sample ready, but got a point to suspected piece of code that cause the problem (again, if it's not by design).
org.springframework.security.taglibs.authz.AbstractAuthorizeTag#getExpressionHandler()
  calling org.springframework.security.web.context.support.SecurityWebApplicationContextUtils#_findWebApplicationContext()

The logic was "Look for a bean of type SecurityExpressionHandler in root context. If there's no root context, look for the bean in any child context".
While I think the correct one should be "Look for a bean of type in preferred context first, then root context then, the rest"
Similar to how DelegatingFilterProxy of spring-web works ...

If I could have confirmation that it's not required spring security to be in root context, I could make a pull request with suggested fix.

rwinch:
> Similar to how DelegatingFilterProxy of spring-web works

As far as I can tell `DelegatingFilterProxy` and `JspAuthorizeTag` are consistent in how they look up the beans. We should be consistent to ensure that the same `ApplicationContext` that is used to create the `FilterChainProxy` is used by `JspAuthorizeTag`.

[...]

PR #18822, opened 2026-02-28
Prefer dispatcher context for authorize tag beans

wonderfulrosemari:
Closes gh-8843

When both root and child web application contexts are present, JSP authorize
tags should resolve security beans from the DispatcherServlet context used for
the current request.

Previously, `AbstractAuthorizeTag` always resolved beans from
`findRequiredWebApplicationContext(servletContext)`, which prefers the root
context. If security beans were defined only in the child context, this could
cause failures like missing `WebSecurityExpressionHandler`.

Changes include:

- resolve application context from the current request's DispatcherServlet
  context attribute when available
- fall back to `SecurityWebApplicationContextUtils.findRequiredWebApplicationContext`
  when no dispatcher context is present
- add regression coverage for root+child context setup to ensure
  `<sec:authorize>` expression evaluation succeeds

### DIFF
--- a/taglibs/src/main/java/org/springframework/security/taglibs/authz/AbstractAuthorizeTag.java
+++ b/taglibs/src/main/java/org/springframework/security/taglibs/authz/AbstractAuthorizeTag.java
@@ -62,8 +62,6 @@ import org.springframework.util.StringUtils;
  */
 public abstract class AbstractAuthorizeTag {
 
-	private static final String DISPATCHER_SERVLET_CONTEXT_ATTRIBUTE = "org.springframework.web.servlet.DispatcherServlet.CONTEXT";
-
 	@SuppressWarnings("NullAway.Init")
 	private @Nullable String access;
 
@@ -226,11 +224,15 @@ public abstract class AbstractAuthorizeTag {
 	}
 
 	private ApplicationContext getApplicationContext() {
-		Object dispatcherContext = getRequest().getAttribute(DISPATCHER_SERVLET_CONTEXT_ATTRIBUTE);
-		if (dispatcherContext instanceof ApplicationContext applicationContext) {
-			return applicationContext;
+		Object value = getRequest().getAttribute(WebAttributes.APPLICATION_CONTEXT_ATTRIBUTE);
+		if (value == null) {
+			return SecurityWebApplicationContextUtils.findRequiredWebApplicationContext(getServletContext());
 		}
-		return SecurityWebApplicationContextUtils.findRequiredWebApplicationContext(getServletContext());
+		if (value instanceof ApplicationContext context) {
+			return context;
+		}
+		throw new IllegalArgumentException("WebAttributes.APPLICATION_CONTEXT_ATTRIBUTE value must be of type "
+				+ "ApplicationContext, found type " + value.getClass());
 	}
 
 }
--- a/taglibs/src/test/java/org/springframework/security/taglibs/authz/AbstractAuthorizeTagTests.java
+++ b/taglibs/src/test/java/org/springframework/security/taglibs/authz/AbstractAuthorizeTagTests.java
@@ -42,6 +42,7 @@ import org.springframework.web.context.WebApplicationContext;
 import org.springframework.web.context.support.GenericWebApplicationContext;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatIllegalArgumentException;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.BDDMockito.given;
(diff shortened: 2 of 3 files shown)


taglibs/src/main/java/org/springframework/security/taglibs/authz/AbstractAuthorizeTag.java
```diff
@@ -62,6 +62,8 @@
  */
 public abstract class AbstractAuthorizeTag {
 
+	private static final String DISPATCHER_SERVLET_CONTEXT_ATTRIBUTE = "org.springframework.web.servlet.DispatcherServlet.CONTEXT";
```
jzheaux: Let's please make this public so it's clear that Spring Security expects applications to set this attribute themselves. `WebAttributes` is a good home for this. Please also take a look at the naming convention used in `WebAttributes` for both the name and the value.

jzheaux: Additionally, it doesn't seem to me that what is being retrieved here is the dispatcher servlet's context, per se. The point of [Rob's suggestion](https://github.com/spring-projects/spring-security/issues/8843#issuecomment-668066652) is to decouple the application context from its source. Thus, a better name might be `APPLICATION_CONTEXT_ATTRIBUTE`.

taglibs/src/main/java/org/springframework/security/taglibs/authz/AbstractAuthorizeTag.java
```diff
@@ -226,4 +225,12 @@ private WebInvocationPrivilegeEvaluator getPrivilegeEvaluator() throws IOExcepti
 		return (WebInvocationPrivilegeEvaluator) wipes.values().toArray()[0];
 	}
 
+	private ApplicationContext getApplicationContext() {
+		Object dispatcherContext = getRequest().getAttribute(DISPATCHER_SERVLET_CONTEXT_ATTRIBUTE);
```
jzheaux: At this point in the code, we aren't sure what it is, so perhaps `value` is a better variable name.

taglibs/src/main/java/org/springframework/security/taglibs/authz/AbstractAuthorizeTag.java
```diff
@@ -226,4 +225,12 @@ private WebInvocationPrivilegeEvaluator getPrivilegeEvaluator() throws IOExcepti
...
+		if (dispatcherContext instanceof ApplicationContext applicationContext) {
```
jzheaux: Instead of silently falling back, let's please error if the attribute is being misused:

```java
if (value == null) {
    return SecurityWebApplicationContextUtils.findRequiredWebApplicationContext(getServletContext());
}
if (value instanceof ApplicationContext context) {
    return context;
}
throw new IllegalArgumentException("WebAttributes.APPLICATION_CONTEXT_ATTRIBUTE value must be of type ApplicationContext, found type " + value.getClass());
```

The reason for this is to alert the application developer to the misconfiguration as early as possible.

taglibs/src/test/java/org/springframework/security/taglibs/authz/AbstractAuthorizeTagTests.java
```diff
@@ -134,6 +134,24 @@ public void expressionFromChildContext() throws IOException {
 		assertThat(this.tag.authorize()).isTrue();
 	}
 
+	@Test
+	@SuppressWarnings("rawtypes")
+	public void expressionFromDispatcherContextWhenRootContextPresent() throws IOException {
```
jzheaux: Please remove the dispatcher-specific language in this test.

taglibs/src/test/java/org/springframework/security/taglibs/authz/AbstractAuthorizeTagTests.java
```diff
@@ -134,6 +134,24 @@ public void expressionFromChildContext() throws IOException {
...
+		DefaultWebSecurityExpressionHandler expected = new DefaultWebSecurityExpressionHandler();
+		WebApplicationContext dispatcher = mock(WebApplicationContext.class);
+		given(dispatcher.getBeansOfType(SecurityExpressionHandler.class))
+			.willReturn(Collections.<String, SecurityExpressionHandler>singletonMap("wipe", expected));
+		given(dispatcher.getBeanNamesForType(SecurityContextHolderStrategy.class)).willReturn(new String[0]);
+		this.request.setAttribute("org.springframework.web.servlet.DispatcherServlet.CONTEXT", dispatcher);
+		this.tag.setAccess("permitAll");
+		assertThat(this.tag.authorize()).isTrue();
+	}
+
```
jzheaux, line 159: Please add a test to confirm an error if the attribute is not of type `ApplicationContext`.

taglibs/src/test/java/org/springframework/security/taglibs/authz/AbstractAuthorizeTagTests.java
```diff
@@ -134,6 +134,24 @@ public void expressionFromChildContext() throws IOException {
...
+		given(root.getBeanNamesForType(SecurityContextHolderStrategy.class)).willReturn(new String[0]);
+		this.servletContext.setAttribute(WebApplicationContext.ROOT_WEB_APPLICATION_CONTEXT_ATTRIBUTE, root);
+		DefaultWebSecurityExpressionHandler expected = new DefaultWebSecurityExpressionHandler();
+		WebApplicationContext dispatcher = mock(WebApplicationContext.class);
+		given(dispatcher.getBeansOfType(SecurityExpressionHandler.class))
+			.willReturn(Collections.<String, SecurityExpressionHandler>singletonMap("wipe", expected));
+		given(dispatcher.getBeanNamesForType(SecurityContextHolderStrategy.class)).willReturn(new String[0]);
+		this.request.setAttribute("org.springframework.web.servlet.DispatcherServlet.CONTEXT", dispatcher);
+		this.tag.setAccess("permitAll");
+		assertThat(this.tag.authorize()).isTrue();
```
jzheaux, line 149: Please verify here that the mocked context was the one that was used.

taglibs/src/test/java/org/springframework/security/taglibs/authz/AbstractAuthorizeTagTests.java
jzheaux: The name here seems to indicate that we would only expect looking up the "dispatcher context" when the "root context" is present. However, the logic you are adding is about using the WebAttribute-based Application Context when one is specified. Will you please adjust the name?

jzheaux (changes requested):
Thanks, @wonderfulrosemari, for the PR! I've left feedback inline.

PR #18824, opened 2026-02-28
Enable Enter key submission for One-Time Token login

### LINKED ISSUE #18817: Enter Should Submit Request OTT Token
The Request a One-Time Token page should allow pressing Enter to submit the form without the need to click.

Image

AnanthaKrishnanJ2001:
### Description
This PR fixes a bug in the default login page where pressing "Enter" in the One-Time Token (OTT) username field would not submit the form. By moving the submit button inside the `<form>` tags and removing the redundant `form` attribute, we restore standard browser submission behavior.

### Testing
- Updated `DefaultLoginPageGeneratingFilterTests` to match the corrected HTML structure.
- Verified that pressing Enter now triggers the submission.

Closes #18817

### DIFF
--- a/config/src/test/java/org/springframework/security/config/annotation/web/configurers/ott/OneTimeTokenLoginConfigurerTests.java
+++ b/config/src/test/java/org/springframework/security/config/annotation/web/configurers/ott/OneTimeTokenLoginConfigurerTests.java
@@ -174,10 +174,10 @@ public class OneTimeTokenLoginConfigurerTests {
 
 						        <p>
 						          <label for="ott-username" class="screenreader">Username</label>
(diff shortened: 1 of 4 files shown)


jzheaux:
Hi, @AnanthaKrishnanJ2001. Thanks for your submission.

I'm not clear on how removing the extra form attribute addresses the submission issue that @rwinch described. I think the real fix is ensuring that `autofocus` is correctly set on the form; on the `input` field if it is editable and on the `button` field if the `input` field is readonly.

Does removing this attribute somehow do this? If not, can you please update the PR to add `autofocus` as described?

jzheaux:
Thanks for getting this PR started, @AnanthaKrishnanJ2001! It will merge into `main` once the build completes.

Issue #18783, opened 2026-02-23
ContentSecurityPolicyConfig: Add Report-To Header


bahuma20:
**Expected Behavior**

Add configuration option for the `Report-to` HTTP-header to the `CspSecurityPolicyConfig`.

Example how this could look like:
```java
public SecurityFilterChain securityFilterChain(HttpSecurity http) {
  return http.headers(headers -> 
    headers.contentSecurityPolicy(csp ->
      csp
        .reportOnly()
        .reportingEndpoints("csp-report-endpoint=\"https://example.com/csp-reports\"")
        .policyDirectives("default-src 'self'; report-to csp-report-endpoint")
    )
  ).build();
}
```
**Current Behavior**
To get the same thing to work right now, the code looks like this:
```java
public SecurityFilterChain securityFilterChain(HttpSecurity http) {
  return http.headers(headers -> headers
    .contentSecurityPolicy(csp ->csp
      .reportOnly()
      .policyDirectives("default-src 'self'; report-to csp-report-endpoint")
    )
    .addHeaderWriter((request, response) -> response.setHeader("Reporting-Endpoints", "csp-report-endpoint=\"https://example.com/csp-reports\""))
  ).build();
}
```

This is unnecessary complex.

**Context**
The `Reporting-Endpoints` HTTP-header and `report-to` directive is the sucessor to the `report-uri` directive which is deprecated.

So everyone implementing a CSP with reporting should set the `Reporting-Endpoints` HTTP-header.

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP#violation_reporting for details.

_Alternative Suggestion_: Full-blown API that allows building the policy and headers declaratively, with support for configuration via properties. (So instead of just specifying a string with a policy, you would use a typed Java api which declares the directives to add, what values those directives have, and where to report to. There could also be support for nonces)

therepanic:
Thanks for reporting, Max @bahuma20. Since I don't see any obstacles to adding another header included in the CSP and consider it justified, I attached PR: gh-18833

PR #18833, opened 2026-03-02
Add `Reporting-Endpoints` header support to `ContentSecurityPolicyConfig`

therepanic:
`Reporting-Endpoints` is a full member of CSP (Content Security Policy), and if we ignore deprecations, it's the third member. The others have already been added and implemented in `ContentSecurityPolicyConfig` and `ContentSecurityPolicyHeaderWriter`. The goal of this commit is to provide a simpler API for interacting with and installing `Reporting-Endpoints`.

Closes: gh-18783

### DIFF
--- a/web/src/main/java/org/springframework/security/web/header/writers/ContentSecurityPolicyHeaderWriter.java
+++ b/web/src/main/java/org/springframework/security/web/header/writers/ContentSecurityPolicyHeaderWriter.java
@@ -18,6 +18,7 @@ package org.springframework.security.web.header.writers;
 
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
+import org.jspecify.annotations.Nullable;
 
 import org.springframework.security.web.header.HeaderWriter;
 import org.springframework.util.Assert;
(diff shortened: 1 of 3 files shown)


web/src/main/java/org/springframework/security/web/header/writers/ContentSecurityPolicyHeaderWriter.java
```diff
@@ -87,12 +96,16 @@ public final class ContentSecurityPolicyHeaderWriter implements HeaderWriter {
...
 
 	private String policyDirectives;
 
 	private boolean reportOnly;
 
+	@Nullable private String reportingEndpoints;
+
 	/**
 	 * Creates a new instance. Default value: default-src 'self'
 	 */
```
therepanic, line 111: I set this field as Nullable, because when writing it will be set and will be an empty field, I think this will then be a big surprise for users.

therepanic:
I think this is a cute change. As described in gh-18783, you can do things like:
```java
public SecurityFilterChain securityFilterChain(HttpSecurity http) {
  return http.headers(headers -> 
    headers.contentSecurityPolicy(csp ->
      csp
        .reportOnly()
        .reportingEndpoints("csp-report-endpoint=\"https://example.com/csp-reports\"")  //<----
        .policyDirectives("default-src 'self'; report-to csp-report-endpoint")
    )
  ).build();
}
```

PR #18746, opened 2026-02-16
Add RestClientOpaqueTokenIntrospector

therepanic:
Since similar classes have alternative versions using RestClient instead of RestTemplate, I think we should do the same with this class.

Closes: gh-18745

PR #18836, opened 2026-03-02
Update `RestTemplateBuilder` usage in `opaque-token.adoc`

therepanic:
We just now use a new form instead of the deprecate one.

Mentioned in: https://github.com/spring-projects/spring-security/pull/18746#discussion_r2873996649


jzheaux:
Since this changes to using `RestClient` on `main`, I rebased this to `7.0.x` and committed it to 57434fc5971c8537863430dc1a35ab802246499f, leaving the change made earlier on https://github.com/spring-projects/spring-security/pull/18746 as-is.

Issue #17782, opened 2025-08-20
Consider preventing null `defaultRolePrefix` in `SecurityExpressionRoot`


sjohnr:
As a follow-up to gh-17585, consider preventing a null `defaultRolePrefix` in `SecurityExpressionRoot` and subclasses of `AbstractSecurityExpressionHandler` for more consistent behavior. This would break passivity in Spring Security 7.

**Context:**

Currently, `SecurityExpressionRoot` and subclasses of `AbstractSecurityExpressionHandler` allow a null `defaultRolePrefix` to be configured. The javadoc states:

> If null or empty, then no default role prefix is used.

When switching to using an `AuthorizationManager` created by `AuthorizationManagerFactory` in `SecurityExpressionRoot`, we can no longer set a null role prefix. To work around this for passivity, we can set the role prefix to blank before setting it on an `AuthorizationManagerFactory`.  See updates to `SecurityExpressionRoot` in #17673 for context.

Instead, we can consider throwing an `IllegalArgumentException` when null is passed into any `setDefaultRolePrefix` method.

ngocnhan-tran1996:
Can I work on this?

jzheaux:
@sjohnr, thanks as always for your insight. I'd prefer to leave it as-is since it is deprecated and will eventually be removed. Folks either way should change to setting the `AuthorizationManagerFactory` instead.

The point in time at which we can change the contract is the same point in time that we can remove the method. Is there a reason you need the method's contract to be changed in advance of its removal? I unfortunately couldn't find a related discussion in the linked issues.

sjohnr:
Hi @jzheaux! This was opened after a discussion with @rwinch around this API while working out the details for the `AuthorizationManagerFactory`. I think it was just an idea he had (or maybe I had, I can't quite remember) so I captured it via an issue for investigation. I think since the window for the major is closed, that's another reason this doesn't really make sense now. Feel free to close it. 😄

jzheaux:
Thanks, @sjohnr. I'm going to decline since it breaks passivity and AFAICT would be removed anyway in the next major. We can always revisit and reopen if needed.

PR #18846, opened 2026-03-04
Throw IllegalArgumentException for null defaultRolePrefix in SecurityExpressionRoot

jayychoi:
Closes gh-17782


jzheaux:
I'm not convinced that we should do this. I've blocked the PR while discussion continues on https://github.com/spring-projects/spring-security/issues/17782

jzheaux:
Thanks, @jayychoi for your contribution. I'm declining since #17782 was declined.

Issue #18832, opened 2026-03-02
Please provide a way to use a RestClient instead of a RestTemplate for restOperations(..) of the NimbusJwtDecoder


tkrah:
At the moment the `NimbusJwtDecoder` does expect a `RestOperations` interface which is implemented by `RestTemplate`:

```
public JwkSetUriJwtDecoderBuilder restOperations(RestOperations restOperations) {
			Assert.notNull(restOperations, "restOperations cannot be null");
			this.restOperations = restOperations;
			return this;
}
```

As `RestTemplate` is deprecated (https://spring.io/blog/2025/09/30/the-state-of-http-clients-in-spring), please provide a method / way to use a `RestClient` instead.

jzheaux:
Closing in favor of https://github.com/spring-projects/spring-security/pull/18847

tkrah:
Can we reopen this one please, none of the closed and referenced / linked PRs got merged or had the suggested / planned RestClient changes?

Issue #18381, opened 2026-01-05
Simplify at+jwt validation


jzheaux:
`at+jwt` validation was added in 6.x, and it would be nice if this were the simplest validation method to add, given that it is [a standard](https://datatracker.ietf.org/doc/html/rfc9068) for JWT validation.

Some ways to achieve this are:

* Add `Nimbus(Reactive)JwtDecoder` builders that adhere to RFC 9068
* Add `Nimbus(Reactive)AtJwtDecoder` implmentations that adhere to RFC 9068

This may require deprecating `Nimbus(Reactive)JwtDecoder` or some of their builders.

The DSL should also be considered. Perhaps something like:

```java
oauth2ResourceServer((oauth2) -> oauth2
    .atJwt()
        .issuer(...)
        .audience(...)
)
```

Note that this ticket remains a work in progress and is not ideal for contribution at this time.

gbaso:
I would love to see this functionality, but I feel that at the moment the `at+jwt` validator (built from `AtJwtBuilder`) is unnecessarily stringent. Specifically, the builder requires you to specify a single `client_id`, and access tokens issued to other clients will be rejected. RFC 9068 [does require](https://datatracker.ietf.org/doc/html/rfc9068#section-2.2) that `client_id` is present, but if I'm interpreting it correctly it is not required that it has to be a specific value, or that a resource server can only accept a specific value.

In fact, for the `client_credentials` flow, it is perfectly normal to have multiple clients with distinct `client_id`s that obtain tokens from the same issuer, and to then call the same resource server. For example, different departments using the corporate issuer with different scopes. Such `client_id`s would change over time and not be known to the resource server beforehand.

On the other hand, one could implement their own validator without this issue, but `AtJwtBuilder` does a lot of heavy lifting and it would be disappointing to not be able to rely on it.

jzheaux:
Hi, @gbaso, thanks for the insight about `AtJwtBuilder`. Perhaps there should be a separate ticket that sets checking for `client_id`'s existence as the default, e.g., the constructor could add:

```java
this.validators.put("client_id", require("client_id"));
```

This could then be optionally replaced by calling the `clientId` builder method, but would not be required.

Would you be interested in submitting a PR to relax that requirement or opening an issue if you feel there is more to be discussed?

In the meantime, note that you can specify something more lenient yourself already by calling the `validators` method on the builder.

gbaso:
@jzheaux I have opened #18890 to address the proposed changes.

[...]

PR #18847, opened 2026-03-04
Add RestClient support to NimbusJwtDecoder

AnanthaKrishnanJ2001:
### Motivation
Following the deprecation of RestTemplate, this PR enables NimbusJwtDecoder to use the modern, fluent RestClient for JWK set retrieval.

### Modifications
- Introduced JwkSetRetriever strategy to abstract HTTP transport from the underlying client.
- Added restClient(RestClient) to JwkSetUriJwtDecoderBuilder.
- Refactored SpringJWKSource to utilize the injected retriever strategy.
- Maintained backward compatibility for RestOperations users.

### Result
Developers can now provide a RestClient bean to NimbusJwtDecoder, allowing for modern Spring 6+ integration and easier customization of the HTTP layer.

##Issue Number
Closes gh-18832

### DIFF
--- a/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
+++ b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
@@ -16,12 +16,17 @@
 
 package org.springframework.security.oauth2.jwt;
 
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.net.URI;
+import java.nio.charset.StandardCharsets;
 import java.security.interfaces.RSAPublicKey;
 import java.text.ParseException;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.Map;
@@ -63,8 +68,13 @@ import org.springframework.cache.support.NoOpCache;
 import org.springframework.core.convert.converter.Converter;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatusCode;
 import org.springframework.http.MediaType;
 import org.springframework.http.RequestEntity;
+import org.springframework.http.ResponseEntity;
+import org.springframework.http.client.ClientHttpRequest;
+import org.springframework.http.client.ClientHttpRequestFactory;
+import org.springframework.http.client.ClientHttpResponse;
 import org.springframework.http.client.SimpleClientHttpRequestFactory;
 import org.springframework.security.oauth2.core.OAuth2Error;
 import org.springframework.security.oauth2.core.OAuth2TokenValidator;
@@ -297,11 +307,17 @@ public final class NimbusJwtDecoder implements JwtDecoder {
 
 		private RestOperations restOperations = new RestTemplateWithNimbusDefaultTimeouts();
 
+		private static final MediaType APPLICATION_JWK_SET_JSON = new MediaType("application", "jwk-set+json");
+
 		private Cache cache = new NoOpCache("default");
 
 		private Consumer<ConfigurableJWTProcessor<SecurityContext>> jwtProcessorCustomizer;
 
-		private JwkSetRetriever jwkSetRetriever = this::defaultRestOperationsRetriever;
+		private RestClient restClient = RestClient.builder()
+			.requestFactory(new RestTemplateWithNimbusDefaultTimeouts().getRequestFactory())
+			.defaultHeaders(
+					(headers) -> headers.setAccept(Arrays.asList(MediaType.APPLICATION_JSON, APPLICATION_JWK_SET_JSON)))
+			.build();
 
 		private JwkSetUriJwtDecoderBuilder(String jwkSetUri) {
 			Assert.hasText(jwkSetUri, "jwkSetUri cannot be empty");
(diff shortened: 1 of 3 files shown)


oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -330,26 +332,26 @@ private JwkSetUriJwtDecoderBuilder(Function<RestOperations, String> jwkSetUri,
 		 *
 		 * <p>
 		 * When this is set to {@code false}, this: <code>
-		 *     NimbusJwtDecoder jwtDecoder = NimbusJwtDecoder.withIssuerLocation(issuer).build();
-		 *     jwtDecoder.setJwtValidator(JwtValidators.createDefaultWithIssuer(issuer);
+		 * NimbusJwtDecoder jwtDecoder = NimbusJwtDecoder.withIssuerLocation(issuer).build();
```
jzheaux, line 351: Will you please preserve the formatting in the JavaDoc?

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -400,9 +402,36 @@ public JwkSetUriJwtDecoderBuilder jwsAlgorithms(Consumer<Set<SignatureAlgorithm>
 		public JwkSetUriJwtDecoderBuilder restOperations(RestOperations restOperations) {
 			Assert.notNull(restOperations, "restOperations cannot be null");
 			this.restOperations = restOperations;
+			this.jwkSetRetriever = this::defaultRestOperationsRetriever;
```
jzheaux: Instead of having two variables that could now get out of sync, could we instead change the internal member variable to `this.restClient` and then this method could do:

```java
Assert.notNull(restOperations, "restOperations cannot be null");
this.restClient = RestClient.create(restOperations);
```

This will hopefully allow removing the `JwkSetRetriever` abstraction and allow `SpringJWKSource` to make the `RestClient` call directly.

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
```diff
@@ -879,6 +882,41 @@ public void decodeWhenSecretKeyValidateTypeFalseThenSkipsNimbusTypeValidation()
 		jwtDecoder.decode(jwt.serialize());
 	}
 
+	@Test
+	void buildWhenUsingRestClientThenFetchesJwkSet() throws Exception {
+		try (MockWebServer server = new MockWebServer()) {
+			String jwkSetJson = "{\"keys\":[{\"kty\":\"RSA\",\"kid\":\"1\",\"n\":\"vGByo16S49YPs0zS06pM\",\"e\":\"AQAB\"}]}";
```
jzheaux: Can you use `JWK_SET` in the class instead of creating your own?

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
```diff
@@ -879,6 +882,41 @@ public void decodeWhenSecretKeyValidateTypeFalseThenSkipsNimbusTypeValidation()
 		jwtDecoder.decode(jwt.serialize());
 	}
 
+	@Test
+	void buildWhenUsingRestClientThenFetchesJwkSet() throws Exception {
+		try (MockWebServer server = new MockWebServer()) {
```
jzheaux: Given that this test is providing its own `RestClient`, it seems like a mocked `ClientHttpRequestFactory` is more appropriate than starting up a web server.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoder.java
```diff
@@ -482,23 +511,29 @@ public NimbusJwtDecoder build() {
...
+		@FunctionalInterface
+		private interface JwkSetRetriever {
+
+			String fetch(String jwkSetUri);
+
+		}
+
 		private static final class SpringJWKSource<C extends SecurityContext> implements JWKSetSource<C> {
 
 			private static final MediaType APPLICATION_JWK_SET_JSON = new MediaType("application", "jwk-set+json");
```
jzheaux: Ideally, you'd use this property instead of recreating the `MediaType` [here](https://github.com/spring-projects/spring-security/pull/18847/changes#diff-1751b483420d664cb6c94db6f5fb1724b7a9bb5b2f8924a60ca03f86987fbdb7R430) and [here](https://github.com/spring-projects/spring-security/pull/18847/changes#diff-1751b483420d664cb6c94db6f5fb1724b7a9bb5b2f8924a60ca03f86987fbdb7R422). Please keep this in mind as you consider [my related comment](https://github.com/spring-projects/spring-security/pull/18847/changes#r2886648655)

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
```diff
@@ -879,6 +882,41 @@ public void decodeWhenSecretKeyValidateTypeFalseThenSkipsNimbusTypeValidation()
...
+
+			RestClient restClient = RestClient.builder().requestFactory(requestFactory).build();
+
+			NimbusJwtDecoder decoder = NimbusJwtDecoder.withJwkSetUri(server.url("/jwks").toString())
+				.restClient(restClient)
+				.build();
+
+			// 2. Use a structurally valid (3-part) JWT string to ensure the decoder tries
+			// to fetch keys
+			String minimalJwt = "eyJhbGciOiJSUzI1NiJ9.eyJzdWIiOiIxMjM0NTY3ODkwIn0.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c";
```
jzheaux: Can you use `SIGNED_JWT` instead of creating your own?

oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/NimbusJwtDecoderTests.java
```diff
@@ -879,6 +882,41 @@ public void decodeWhenSecretKeyValidateTypeFalseThenSkipsNimbusTypeValidation()
...
+
+			try {
+				decoder.decode(minimalJwt);
```
jzheaux: I think the test can likely be written in a way that the `decode` method passes. I'd prefer not to add the noise of ignoring an exception to the test.

jzheaux (changes requested):
Thanks for the PR, @AnanthaKrishnanJ2001! I've left my feedback inline.

AnanthaKrishnanJ2001:
Thanks for the feedback, @jzheaux! These suggestions make total sense. I'll work on refactoring the builder to use a single RestClient and clean up the tests.

AnanthaKrishnanJ2001:
Hi @jzheaux, I've updated the PR to address your feedback:

--Consolidated the retriever logic into a single restClient field, removing the JwkSetRetriever abstraction. Since RestClient.create(RestOperations) is not available in this Spring version, I bridged RestOperations into RestClient via a RestOperationsClientHttpRequestFactory adapter, so SpringJWKSource now makes RestClient calls directly.
--Moved APPLICATION_JWK_SET_JSON to the builder level and replaced all duplicated MediaType instances with the shared constant.
--Restored JavaDoc indentation in all three validateType methods.
--Updated the test to use the existing JWK_SET and SIGNED_JWT constants, and replaced MockWebServer with a mocked ClientHttpRequestFactory.
--Verified the build locally with ./gradlew :spring-security-oauth2-jose:check.

jzheaux:
Hi, @AnanthaKrishnanJ2001, looking at the latest, I appreciate the creativity and the updates; however, I think it's clear that `RestClient` and `RestOperations` are not well-suited to be in the same builder as the adaptation requires a lot of internal boilerplate as well as some fiddling with the transport between maps and strings and input streams.

I'm going to close this in favor of https://github.com/spring-projects/spring-security/issues/18381 for the moment and see whether or not that ticket presents an opportunity to add `RestClient` support for JWT validation.

PR #18850, opened 2026-03-05
Relax auth_time validation on ID token refresh

### LINKED ISSUE #18839: auth_time validation fails when SSO session is renewed
**Describe the bug**
The `auth_time` in the ID token is validated against the `auth_time` of the original access token in the session. But in some cases, this auth_time can be reset on the Identity Provider, resulting in an `[invalid_id_token] Invalid authenticated at time` on the client side whenever the token is refreshed.

**Background**
I ran into this bug when updating from Spring Boot 3.3.x to 3.5.x, so somewhere in the newer Spring Security version this validation has been added. Refreshing tokens has been working perfectly fine before, but after the update this error broke our code as we enforce authentication on other applications with `prompt=login`. I read the documentation of the OIDC specification and think that the current validation of the auth_time might be a little too strict. I am curious on what others would think of this situation.

**To Reproduce**
* Log in to a Spring Boot app via OIDC SSO
* Log in to another application on the same Identity Provider, but enforce entering credentials again with `prompt=login` 
* Wait at least 5 minutes for the first access token to expire
* Try to refresh the token from the (first) Spring Boot app.
* Actual behavior: An `invalid_id_token` error pops up as the `auth_time` of the refreshed ID token doesn't match the `auth_time` of the original one within the application session.

**Expected behavior**
The tokens are refreshed, without any arror.

The [OIDC specification](https://openid.net/specs/openid-connect-core-1_0.html#:~:text=Time%20when%20the%20End%2DUser%20authentication%20occurred) mentions: 
> auth_time: Time when the End-User authentication occurred.

And also the [requirements for a refreshed ID token](https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse:~:text=if%20the%20ID%20Token%20contains%20an%20auth_time%20Claim%2C%20its%20value%20MUST%20represent%20the%20time%20of%20the%20original%20authentication%20%2D%20not%20the%20time%20that%20the%20new%20ID%20token%20is%20issued%2C):
> if the ID Token contains an auth_time Claim, its value MUST represent the time of the original authentication - not the time that the new ID token is issued,

It might be debatable what 'time of original authentication' means here, but I would interpret that as: 'the time  when authentication (for this new ID token) originated at the Identity Provider'. And as the Identity Provider only keeps track of one authentication session, that latest `auth_time` would be used here. Which is a perfectly valid for the refreshed ID token.

Spring Security wants to [validate this auth_time after the ID token is being refreshed](https://github.com/spring-projects/spring-security/blob/25ccb1fd702f00f27e1c212d7e2cfbfb218c5b11/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcAuthorizedClientRefreshedEventListener.java#L298) against the auth_time in the original access token it received earlier. But I think it is not said in the specification that this should be done that way.

To fix this, Spring Security could either:
* remove the `validateAuthenticatedAt` validation as you cannot know the time of the original authentication at this point.
* or adapt it to only validate that it is earlier than the `issued_at` of the newly refreshed ID token. As that is the only thing you can validate for certain at this point.
* or make it configurable, so it can be disabled to allow auth_time changes.

answndud:
When an OP refreshes an ID token and updates auth_time (for example after SSO session renewal), strict equality with the previous ID token can reject an otherwise valid refreshed token.

This change relaxes that check in both servlet and reactive paths:
- keep issuer/sub/aud/nonce checks as-is
- validate auth_time is not after refreshed token iat (with configured clock skew)

Updated tests:
- OidcAuthorizedClientRefreshedEventListenerTests
- RefreshOidcUserReactiveOAuth2AuthorizationSuccessHandlerTests

Verification:
- ./gradlew -PtestToolchain=21 -PtestCompileTargetVersion=21 :spring-security-oauth2-client:test --tests "org.springframework.security.oauth2.client.oidc.authentication.OidcAuthorizedClientRefreshedEventListenerTests" --tests "org.springframework.security.oauth2.client.RefreshOidcUserReactiveOAuth2AuthorizationSuccessHandlerTests" -x :spring-security-javascript:nodeSetup -x :spring-security-javascript:npmInstall -x :spring-security-javascript:npm_run_assemble -x :spring-security-javascript:assemble

Fixes #18839

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshOidcUserReactiveOAuth2AuthorizationSuccessHandler.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshOidcUserReactiveOAuth2AuthorizationSuccessHandler.java
@@ -273,7 +273,7 @@ public final class RefreshOidcUserReactiveOAuth2AuthorizationSuccessHandler
 		if (idToken.getAuthenticatedAt() == null) {
 			return;
 		}
-		if (!idToken.getAuthenticatedAt().equals(existingOidcUser.getIdToken().getAuthenticatedAt())) {
+		if (idToken.getAuthenticatedAt().isAfter(idToken.getIssuedAt().plus(this.clockSkew))) {
 			OAuth2Error oauth2Error = new OAuth2Error(INVALID_ID_TOKEN_ERROR_CODE, "Invalid authenticated at time",
 					REFRESH_TOKEN_RESPONSE_ERROR_URI);
 			throw new OAuth2AuthenticationException(oauth2Error, oauth2Error.toString());
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcAuthorizedClientRefreshedEventListener.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/oidc/authentication/OidcAuthorizedClientRefreshedEventListener.java
@@ -300,7 +300,7 @@ public final class OidcAuthorizedClientRefreshedEventListener
 			return;
 		}
 
-		if (!idToken.getAuthenticatedAt().equals(existingOidcUser.getIdToken().getAuthenticatedAt())) {
+		if (idToken.getAuthenticatedAt().isAfter(idToken.getIssuedAt().plus(this.clockSkew))) {
 			OAuth2Error oauth2Error = new OAuth2Error(INVALID_ID_TOKEN_ERROR_CODE, "Invalid authenticated at time",
 					REFRESH_TOKEN_RESPONSE_ERROR_URI);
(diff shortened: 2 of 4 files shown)


answndud:
Hi team, friendly follow-up on this PR. If you'd prefer a different direction or any changes, I'm happy to update it.

jgrandja:
@answndud Closing this as a fix is not required. Please see this [comment](https://github.com/spring-projects/spring-security/issues/18839#issuecomment-4083299089).

In the future, please hold off on submitting a fix until the team has evaluated the issue and determined if a fix is required. Thanks.

answndud:
Sorry for jumping ahead with a fix before the team had finished evaluating the issue. I’ve been experimenting with an AI agent to automate parts of the contribution process, and I mistakenly allowed it to submit this PR prematurely without proper oversight. I’ll make sure to wait for the team's evaluation in the future before submitting any further fixes. Thanks for your understanding."

Issue #18204, opened 2025-11-20
CookieRequestCache SavedRequest has no parameters


StefanPopa99:
**Describe the bug**
RequestCache.getRequest returns the SavedRequest. However, CookieRequestCache.getRequest implementation returns a SavedRequest without parameters; Hence, calling getParameterValues will always return null. On the other hand, HttpSessionRequestCache implementation works as expected.

As a workaround, I rely on SavedRequest.getRedirectUrl and extract the parameters from there.

**To Reproduce**
For example, use CookieRequestCache and go through an oauth2 flow. On callback, the SavedRequest has no parameters.

**Expected behavior**
SavedRequest.getParameterValues should return the parameters if they are available in the url.

**Proposed solution**
class CookieRequestCache
```
@Override
public @Nullable SavedRequest getRequest(HttpServletRequest request, HttpServletResponse response) {
	Cookie savedRequestCookie = WebUtils.getCookie(request, COOKIE_NAME);
	if (savedRequestCookie == null) {
		return null;
	}
	String originalURI = decodeCookie(savedRequestCookie.getValue());
	if (originalURI == null) {
		return null;
	}
	UriComponents uriComponents = UriComponentsBuilder.fromUriString(originalURI).build();
	DefaultSavedRequest.Builder builder = new DefaultSavedRequest.Builder();
	int port = getPort(uriComponents);
	MultiValueMap<String, String> queryParams = uriComponents.getQueryParams();
	Map<String, String[]> parameterMap = queryParams.entrySet().stream().collect(
			Collectors.toMap(
					Map.Entry::getKey,
					entry -> entry.getValue().toArray(new String[0])));
	return builder.setScheme(uriComponents.getScheme())
		.setServerName(uriComponents.getHost())
		.setRequestURI(uriComponents.getPath())
		.setQueryString(uriComponents.getQuery())
		.setServerPort(port)
		.setMethod(request.getMethod())
		.setLocales(Collections.list(request.getLocales()))
		.setParameters(parameterMap)
		.build();
}
```

rwinch:
Closing in favor of the PR gh-18857

PR #18857, opened 2026-03-08
Fix CookieRequestCache parameters

itsmevichu:
Fixes https://github.com/spring-projects/spring-security/issues/18204


rwinch:
Thanks for the PR @itsmevichu! Since this is a bug, I've changed the base to be 6.5.x When this passes it will be merged into 6.5.x, 7.0.x and main

PR #18862, opened 2026-03-09
Spring Session management docs correction

sankranty:
Adjusted authentication description to reflect that starting from Spring Security 6 security context persistence behavior has changed.


sankranty:
It duplicates https://github.com/spring-projects/spring-security/pull/18345 (was opened by accident), so closing it.

Issue #18034, opened 2025-10-11
WebAuthnAuthentication Jackson config is missing and breaks oauth2 authorization server


mariusleu:
**Describe the bug**
There are missing Jackson mixins, serializers and deserializers for `WebAuthnAuthentication` making it impossible to use with the spring authorization-server `JdbcOAuth2AuthorizationService`.

**To Reproduce**
Configure the spring security app with OAuth2AuthorizationServerConfigurer  and webAuthn at the same time, while the authorization server uses a persistent authorization repository such as `JdbcOAuth2AuthorizationService`

**Expected behavior**
WebAuthnAuthentication must work with spring oauth2 authorization server.

**Sample**

This exception is thrown by the authorization service on the `/token` endpoint:

`
Caused by: java.lang.IllegalArgumentException: The class with org.springframework.security.web.webauthn.authentication.WebAuthnAuthentication and name of org.springframework.security.web.webauthn.authentication.WebAuthnAuthentication is not in the allowlist. If you believe this class is safe to deserialize, please provide an explicit mapping using Jackson annotations or by providing a Mixin. If the serialization is only done by a trusted source, you can also enable default typing. See https://github.com/spring-projects/spring-security/issues/4370 for details
`

rwinch:
Thanks to @making this is now fixed in 7.0.x and main

PR #18878, opened 2026-03-11
Add Jackson Mixin for WebAuthnAuthentication

making:
This PR adds Jackson Mixin for WebAuthnAuthentication.
Without this change, Passkeys support with Authorization Server + `JdbcOAuth2AuthorizationService` won't work.

Closes gh-18034


rwinch:
Thank. you @making! This is merged into 7.0.x and main now

PR #18883, opened 2026-03-12
Implement equals and hashCode in ImmutablePublicKeyCredentialUserEntity

aspan:


webauthn/src/main/java/org/springframework/security/web/webauthn/api/ImmutablePublicKeyCredentialUserEntity.java
```diff
@@ -135,6 +136,19 @@ public static PublicKeyCredentialUserEntityBuilder builder() {
 		return new PublicKeyCredentialUserEntityBuilder();
 	}
 
+	@Override
+	public boolean equals(Object obj) {
+		if (!(obj instanceof ImmutablePublicKeyCredentialUserEntity that)) {
+			return false;
+		}
+		return Objects.equals(getName(), that.getName()) && Objects.equals(getId(), that.getId());
```
jzheaux: Will you please use `this.name`/`that.name` and `this.id`/`that.id`? The reason is so that this class doesn't get surprised with overridden implementations, should this class ever become non-`final`.

webauthn/src/main/java/org/springframework/security/web/webauthn/api/ImmutablePublicKeyCredentialUserEntity.java
```diff
@@ -135,6 +136,19 @@ public static PublicKeyCredentialUserEntityBuilder builder() {
 		return new PublicKeyCredentialUserEntityBuilder();
 	}
 
+	@Override
+	public boolean equals(Object obj) {
+		if (!(obj instanceof ImmutablePublicKeyCredentialUserEntity that)) {
```
jzheaux, line 144: Will you please add a quick equality check:

```java
if (this == obj) {
    return true;
}
```

aspan:
Hi Thanks for the feedback. I've made the requested changes.

jzheaux:
Thanks, @aspan! This is now merged into `main`.

Issue #18871, opened 2026-03-10
`ExceptionHandlingConfigurer#createDefaultAccessDeniedHandler` should always fallback to `AccessDeniedHandlerImpl`


liias:
`org.springframework.security.config.annotation.web.configurers.ExceptionHandlingConfigurer#createDefaultAccessDeniedHandler` has this implementation:

```java
LinkedHashMap<RequestMatcher, AccessDeniedHandler> defaultDeniedHandlerMappings

private AccessDeniedHandler createDefaultAccessDeniedHandler(H http) {
    if (this.defaultDeniedHandlerMappings.isEmpty()) {
        return new AccessDeniedHandlerImpl();
    }
    if (this.defaultDeniedHandlerMappings.size() == 1) {
        return this.defaultDeniedHandlerMappings.values().iterator().next();
    }
    return new RequestMatcherDelegatingAccessDeniedHandler(this.defaultDeniedHandlerMappings,
				new AccessDeniedHandlerImpl());
}
```

I'd propose that special case for `this.defaultDeniedHandlerMappings.size() == 1` is not intuitive, because `defaultDeniedHandlerMappings` is always added for a **specific RequestMatcher**.

- If I don't add a handler for a request matcher, then `AccessDeniedHandlerImpl` is always used.
- If I add two handlers for request matchers, then `AccessDeniedHandlerImpl` is used when those matchers don't match
- **If I add exactly one handler for a request matcher, then that handler is also used when its request matcher does not match**

I assume the reason is backwards compatibility rather than design choice?
Sorry if it has been reported before.

answndud:
I would like to work on this.

My plan is to remove the single-mapping special case so that the fallback AccessDeniedHandlerImpl is used consistently whenever no configured RequestMatcher matches, and to add a regression test for the single-matcher case.

answndud:
Opened PR #18889 with the fix and regression test for the single-matcher fallback case.

Tests run locally:
- `export JAVA_HOME=/opt/homebrew/opt/openjdk/libexec/openjdk.jdk/Contents/Home; export PATH="$JAVA_HOME/bin:$PATH"; export NPM_CONFIG_WORKSPACES=false; ./gradlew :spring-security-config:test --tests org.springframework.security.config.annotation.web.configurers.ExceptionHandlingConfigurerAccessDeniedHandlerTests`

liias:
That's cool, but it would be a breaking change, so spring people need to decide how to make it not breaking, or if to do no changes at all.

PR #18889, opened 2026-03-13
Always fallback to AccessDeniedHandlerImpl for unmatched requests

answndud:
Fixes gh-18871

`ExceptionHandlingConfigurer#createDefaultAccessDeniedHandler` currently returns the configured handler directly when only one `defaultAccessDeniedHandlerFor` mapping is registered. In that case, unmatched requests do not fall back to `AccessDeniedHandlerImpl`.

This change always uses `RequestMatcherDelegatingAccessDeniedHandler` when default denied handler mappings are present so that unmatched requests consistently fall back to `AccessDeniedHandlerImpl`.

It also updates the regression test for the single-matcher case to verify that only matching requests use the custom handler and non-matching requests return `403 Forbidden`.

Testing:
- `export JAVA_HOME=/opt/homebrew/opt/openjdk/libexec/openjdk.jdk/Contents/Home; export PATH="$JAVA_HOME/bin:$PATH"; export NPM_CONFIG_WORKSPACES=false; ./gradlew :spring-security-config:test --tests org.springframework.security.config.annotation.web.configurers.ExceptionHandlingConfigurerAccessDeniedHandlerTests`

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/ExceptionHandlingConfigurer.java
@@ -285,9 +285,6 @@ public final class ExceptionHandlingConfigurer<H extends HttpSecurityBuilder<H>>
 		if (this.defaultDeniedHandlerMappings.isEmpty()) {
 			return new AccessDeniedHandlerImpl();
 		}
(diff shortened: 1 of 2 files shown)


answndud:
Closing this for now. I opened it too early before the team had decided whether this behavior should change, and I do not want to create extra noise while that decision is still open. If maintainers decide this direction is desirable later, I can revisit it.

PR #18890, opened 2026-03-13
Relax client_id validation in AtJwtBuilder

gbaso:
`AtJwtBuilder` currently requires configuring a specific `client_id` value, and tokens that containing a different `client_id` are rejected. As discussed in https://github.com/spring-projects/spring-security/issues/18381#issuecomment-3999085699, this is a stronger requirement than that specified in RFC 9068, which only requires that a `client_id` claim is present.

In scenarios like the `client_credentials` grant type, multiple clients may obtain tokens from the same issuer and call the same resource server. In these cases, the resource server may not even know all possible `client_id` values in advance.

This change relaxes the default behavior of `AtJwtBuilder` so that it validates only the presence of the `client_id` claim. A specific value can still be required by calling `AtJwtBuilder#clientId(...)`, just as before.

Relates to gh-18381


jzheaux:
Thanks, @gbaso, for the quick turnaround! This will merge once the build completes.

PR #18895, opened 2026-03-14
Make authenticationSuccessHandler Configurable in Resource Server DSL

iain-henderson:
The lack of a configurable ServerAuthenticationSuccessHandler in OAuth2ResourceServerSpec seemed like an oversight.

This PR corrects that.


config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
```diff
@@ -4186,6 +4188,12 @@ public OAuth2ResourceServerSpec authenticationFailureHandler(
 			return this;
 		}
 
+		public OAuth2ResourceServerSpec authenticationSuccessHandler(
```
jzheaux, line 4200: Will you please add a JavaDoc here, please? Please include a description of the method, what the default value is, and a `@since 7.1` indicating the version in which it was added.

jzheaux (changes requested):
Thanks, @iain-henderson! I've left some feedback inline. Will you also please add a unit test to `OAuth2ResourceServerSpecTests` to confirm the `authenticationSuccessHandler` is used when specified?

jzheaux:
In addition to the review comments, will you please make sure to [sign the DCO](https://github.com/spring-projects/spring-security/pull/18895/checks?check_run_id=67068690315)?

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
iain-henderson: While I'm at it should I add a JavaDoc for the failure handler too?

iain-henderson:
DCO Signoff fixed and unit tests added. Please review to see if changes cover what is needed.

config/src/main/java/org/springframework/security/config/web/server/ServerHttpSecurity.java
jzheaux: not at this time, thanks for asking. This is mostly because we are down to the wire on features, so that will be less to review

jzheaux (approved):

jzheaux:
@iain-henderson can you clarify for me what you are needing this for? Is the primary motivation to address what feels like an oversight?

jzheaux:
I've added some polish as well as the symmetric Kotlin support. However, to be clear, I'd like to wait until we have a clear use case to add this feature.

iain-henderson:
@jzheaux Just to clarify, my primary motivation is to address an oversight.

Our implementation supports multiple authentication methods (including OAUTH) and we were using successHandlers on all of these, but we had to re-implement our success handlers as a WebFilter to work around this oversight.

It actually might be more maintainable the way it is (it matches our non-reactive implementation), but I'm sure we aren't the only ones impacted by this.

jzheaux:
Thanks again, @iain-henderson! This will merge into `main` once the build completes.

PR #18928, opened 2026-03-18
Exclude build output directories from nohttp source set

ribafish:
## Summary
- The `checkstyleNohttp` task scans the entire project tree for non-HTTPS URLs
- Without excluding `**/build/**`, subproject build output directories generated during the first build become additional source inputs for subsequent builds
- This changes the cache key and causes cache misses when using the Gradle build cache
- Adds `**/build/**` to the `nohttp` source exclusion list alongside the existing `buildSrc/build/**` exclusion

## Test plan
- [x] Verify `checkstyleNohttp` still runs correctly (the excluded directories only contain build outputs, not source files)
- [x] Verify cache hit on consecutive builds with `--build-cache`

Tested that caching now works for all cases here: https://github.com/gradle/develocity-oss-projects/actions/runs/23246860010

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -80,7 +80,7 @@ develocity {
 }
 
 nohttp {
-	source.exclude "buildSrc/build/**", "**/build/**", "javascript/.gradle/**", "javascript/package-lock.json", "javascript/node_modules/**", "javascript/build/**", "javascript/dist/**"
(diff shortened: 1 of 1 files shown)


rwinch:
Thank you for the PR @ribafish I've marked this to be merged as soon as the build passes

Issue #18929, opened 2026-03-18
All equals methods with jspecify support must have a Nullable argument


therepanic:
All module classes that support jspecify (and will support it in the future) must have `@Nullable` in their equals method.

I believe this is necessary for several reasons. The main one is that the Object.equals contract itself can always accept null, and we're removing this. This leads to bugs like https://github.com/spring-projects/spring-security/issues/18927. For example, in this specific issue, the Kotlin compiler can't correctly resolve the equals signature.

Examples are readily available. In Spring projects, this has already been implemented at the very beginning of jspecify implementation. For example:

Spring Framework:
https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/util/comparator/BooleanComparator.java#L74
https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/http/HttpHeaders.java#L2066

Spring grpc
https://github.com/spring-projects/spring-grpc/blob/main/spring-grpc-test-spring-boot-autoconfigure/src/main/java/org/springframework/boot/grpc/test/autoconfigure/InProcessTransportContextCustomizerFactory.java#L73

We should fix this and take this into account in the future when adding jspecify support to other modules.

Relates: https://github.com/spring-projects/spring-security/issues/18927

Issue #18927, opened 2026-03-18
org.springframework.security.core.userdetails.User.equals should accept @Nullable parameter


npruehs:
Hey everyone,

when upgrading our Spring Boot web application from Spring Boot version 3.5.11 to 4.0.3, we ran into an issue with the Kotlin compiler 2.3.0. 

We have declared a Kotlin class extending `org.springframework.security.core.userdetails.User`, without overriding `equals`. Now, when compiling that class, the following error occurs when specifying the `jsr305=strict` Kotlin compiler flag, even though we did not override `equals` ourselves:

```
Kotlin: Inherited platform declarations clash: The following declarations have the same JVM signature (equals(Ljava/lang/Object;)Z):
    fun equals(obj: Any): Boolean defined in de.hhla.coast.api.security.MyUser
    fun equals(other: Any?): Boolean defined in de.hhla.coast.api.security.MyUser
```

Taking a closer look, we found that the `equals` method of the Spring Security `User` class is declared as follows:

`public boolean equals(Object obj)`

At other places, such as for the private `password` field, nullability is now explicitly indicated by specifying `@Nullable`. As the default declaration of `equals` in `Object` allows for `null` parameters, I suggest explicitly specifying it here as well, changing the declaration to

`public boolean equals(@Nullable Object obj)`

I'd expect this to fix the above compilation error.

Please let me know what you think!

Best regards,
Nick

therepanic:
Thanks a lot Nick @npruehs  for reporting this. I think this issue is directly related to the design itself (yes, Object.equals) and therefore to consistency, as other Spring projects have already added `@Nullable` to equals and bugs like this.

I think we definitely need to fix this not just in `.userdetails.User`, but in all classes that contain jspecify and an equals method. I've opened a new issue to track this and already attached a PR: https://github.com/spring-projects/spring-security/issues/18929

hqqw2h-lgtm:
I'd like to work on this. Will add @Nullable to the User.equals parameter.

npruehs:
This has been fixed as part of https://github.com/spring-projects/spring-security/pull/18930. Thanks everyone!

PR #18930, opened 2026-03-18
Fix equals nullability annotations for jspecify compliance

therepanic:
In this commit, we added `@Nullable` to equals methods of classes that support `jspecify` for consistency with other Spring projects and to avoid bugs that caused other Spring projects to do this natively.

Closes: gh-18929

### DIFF
--- a/access/src/main/java/org/springframework/security/access/method/DelegatingMethodSecurityMetadataSource.java
+++ b/access/src/main/java/org/springframework/security/access/method/DelegatingMethodSecurityMetadataSource.java
@@ -114,8 +114,10 @@ public final class DelegatingMethodSecurityMetadataSource extends AbstractMethod
 		}
 
 		@Override
-		public boolean equals(Object other) {
-			DefaultCacheKey otherKey = (DefaultCacheKey) other;
+		public boolean equals(@Nullable Object other) {
+			if (!(other instanceof DefaultCacheKey otherKey)) {
+				return false;
+			}
 			return (this.method.equals(otherKey.method)
 					&& ObjectUtils.nullSafeEquals(this.targetClass, otherKey.targetClass));
 		}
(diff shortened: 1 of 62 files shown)


kerberos/kerberos-core/src/main/java/org/springframework/security/kerberos/authentication/KerberosServiceRequestToken.java
```diff
@@ -99,10 +99,13 @@ public KerberosServiceRequestToken(byte[] token) {
 	 * equals() is based only on the Kerberos token
 	 */
 	@Override
-	public boolean equals(Object obj) {
+	public boolean equals(@Nullable Object obj) {
 		if (this == obj) {
 			return true;
 		}
+		if (obj == null) {
+			return false;
```
therepanic: In some places this is necessary, otherwise the project will not build.

access/src/main/java/org/springframework/security/access/method/DelegatingMethodSecurityMetadataSource.java
```diff
@@ -114,7 +114,10 @@ private static class DefaultCacheKey {
 		}
 
 		@Override
-		public boolean equals(Object other) {
+		public boolean equals(@Nullable Object other) {
+			if (other == null) {
```
jzheaux: This one feels odd since I would imagine it best caught by a type check. 

Can we remove this and also the unsafe type check by doing:

```java
if (!(other instanceof DefaultCacheKey otherKey)) {
    return false;
}
```

kerberos/kerberos-core/src/main/java/org/springframework/security/kerberos/authentication/KerberosServiceRequestToken.java
jzheaux: I believe this one could also be addressed by fixing the unsafe type check that follows it. I'd recommend it just after the `super.equals(obj)` clause.

therepanic:
Thanks for the great review, @jzheaux. I polished everything up and pushed it.

Yes, that closes #18927, I included it in the commit message.

jzheaux:
Nice, @therepanic! Thanks for the updates. This will merge into `main` once the build completes.

Issue #18925, opened 2026-03-17
Support Disabling MFA when the User has a Specified Authority


rwinch:
This can be useful if a particular factor or type of user should not have MFA enabled. For example, if a bearer token is used or a passkey is used, that might be enough to not require multiple factors

therepanic:
Hi, @rwinch. Can I work on this?

rwinch:
Even though I created it, this is not yet triaged. Please view waiting-for-triage as an indication that it should not yet be implemented.

PR #18939, opened 2026-03-19
Add unless support to AdditionalRequiredFactorsBuilder

therepanic:
In this commit, we are adding support for 'unless' to 'AdditionalRequiredFactorsBuilder'. This is useful if we want a specific user type or factor to disable MFA.

Closes: gh-18925

### DIFF
--- a/core/src/main/java/org/springframework/security/authorization/AuthorizationManagerFactories.java
+++ b/core/src/main/java/org/springframework/security/authorization/AuthorizationManagerFactories.java
@@ -29,6 +29,7 @@ import org.springframework.util.Assert;
  * Creates common {@link AuthorizationManagerFactory} instances.
  *
  * @author Rob Winch
+ * @author Andrey Litvitski
  * @since 7.0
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/springframework/security/authorization/AuthorizationManagerFactories.java
```diff
@@ -95,6 +114,20 @@ public AdditionalRequiredFactorsBuilder<T> withWhen(
...
 		}
 
+		/**
+		 * Customize the condition that determines if the required factors are skipped.
+		 * @param condition a function that takes the current condition and returns the
+		 * new condition
+		 * @return the {@link AdditionalRequiredFactorsBuilder} to further customize
+		 * @since 7.1
+		 */
+		public AdditionalRequiredFactorsBuilder<T> withUnless(
```
therepanic, line 124: For consistency, we'll add this method (since there's `withWhen`)

Issue #18970, opened 2026-03-23
OnCommittedResponseWrapper#setHeader does not properly handle null value


Jusjuc:
**Describe the bug**

The stacktrace looks like this ...

```

org.eclipse.jetty.server.ResourceService - Failed to serve resource: /index.html

java.lang.NumberFormatException: Cannot parse null string

                           at java.base/java.lang.Long.parseLong(Long.java:674)

                           at java.base/java.lang.Long.parseLong(Long.java:836)

                           at org.springframework.security.web.util.OnCommittedResponseWrapper.checkContentLengthHeader(OnCommittedResponseWrapper.java:91)

                           at org.springframework.security.web.util.OnCommittedResponseWrapper.setHeader(OnCommittedResponseWrapper.java:73)

                           at org.eclipse.jetty.ee10.servlet.ServletCoreResponse$HttpServletResponseHttpFields.remove(ServletCoreResponse.java:390)

                           at org.eclipse.jetty.ee10.servlet.ServletCoreResponse$HttpServletResponseHttpFields.remove(ServletCoreResponse.java:398)

                           at org.eclipse.jetty.server.ResourceService.putHeaders(ResourceService.java:741)

                           at org.eclipse.jetty.server.ResourceService.sendData(ResourceService.java:681)

```

 

What is happening is that the HttpServletResponse.setHeader("Content-Length", null) method is being called.

 

The OnCommittedResponseWrapper.setHeader(String, String) sees the Content-Length header and then blindly tries parse the value.

 

[spring-security/web/src/main/java/org/springframework/security/web/util/OnCommittedResponseWrapper.java](https://github.com/spring-projects/spring-security/blob/main/web/src/main/java/org/springframework/security/web/util/OnCommittedResponseWrapper.java#L73-L94)

 

```

                           public void setHeader(String name, String value) {

                                                      checkContentLengthHeader(name, value);

                                                      super.setHeader(name, value);

                           }

 

                           ...

 

                           private void checkContentLengthHeader(String name, String value) {

                                                      if ("Content-Length".equalsIgnoreCase(name)) {

                                                                                 setContentLength(Long.parseLong(value));

                                                      }

                           }

```

 

The servlet spec has always treated a null in the value parameter for setHeader as a remove of that header:

https://github.com/jakartaee/servlet/issues/159

 

Similar issue has already been reported and fixed for spring-framework:

https://github.com/spring-projects/spring-framework/issues/34460

 

**To Reproduce**

[...]

PR #18971, opened 2026-03-23
Handle setHeader null values in response wrappers

therepanic:
The main problem is that `HttpServletResponse`, according to its contract, accepts null values in `setHeader`. When an argument is null, we must remove the header. So, the main idea is to make these arguments `@Nullable` in our implementations and handle cases where they are null as described in the contract.

By the way, some methods, such as `addHeader` in the same interface, can also accept null by contract, but in that case, we don't have to do anything. I think we're handling this correctly now, and we can afford to make their arguments nonnullable, as is now.

https://github.com/jakartaee/servlet/blob/main/api/src/main/java/jakarta/servlet/http/HttpServletResponse.java#L306

Closes: gh-18970

### DIFF
--- a/web/src/main/java/org/springframework/security/web/firewall/FirewalledResponse.java
+++ b/web/src/main/java/org/springframework/security/web/firewall/FirewalledResponse.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import jakarta.servlet.http.Cookie;
 import jakarta.servlet.http.HttpServletResponse;
 import jakarta.servlet.http.HttpServletResponseWrapper;
+import org.jspecify.annotations.Nullable;
 
 import org.springframework.util.Assert;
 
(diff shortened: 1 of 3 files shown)


web/src/main/java/org/springframework/security/web/firewall/FirewalledResponse.java
```diff
@@ -49,7 +51,7 @@ public void sendRedirect(String location) throws IOException {
 	}
 
 	@Override
-	public void setHeader(String name, String value) {
+	public void setHeader(String name, @Nullable String value) {
 		validateCrlf(name, value);
 		super.setHeader(name, value);
 	}
```
therepanic, line 57: We just need to mark it as `@Nullable` because if `value` is null, `validateCrlf` will work as expected anyway.

rwinch:
Thanks for the contribution! We decided to take a slightly different approach — focusing the fix on `OnCommittedResponseWrapper` and handling null values comprehensively across all its header, writer, and output stream methods. Since this is a bug fix, we also wanted to start from `6.5.x` and forward-merge from there. The fix has been applied in #18989 (6.5.x), #18990 (7.0.x), and #18991 (main).

If you'd like, feel free to resubmit a new PR focusing on the `FirewalledResponse` changes.

Issue #18967, opened 2026-03-23
Use idiomatic Kotlin in custom filter documentation


jzheaux:
The [Adding a Custom Filter](https://docs.spring.io/spring-security/reference/servlet/architecture.html#_adding_a_custom_filter) section of the servlet architecture reference only demonstrates `addFilterAfter`; it would be nice if it demonstrated `addFilterBefore` as well. 

More specifically, though, the Kotlin snippets would be more helpful if they demonstrated idiomatic Kotlin using reified type parameters (e.g., `addFilterBefore<LogoutFilter>(TenantFilter())`). That is, for the `addFilterAfter` documentation, something like:

```kotlin
http {
    addFilterAfter<AnonymousAuthenticationFilter>(TenantFilter())
}
```

When updating these snippets, please move the snippet into `docs/src/test/java` and use `include::` syntax to include the snippet instead of including them inline.

Related to #17993

therepanic:
Hi, @jzheaux. Can I work on it?

jzheaux:
It's yours, @therepanic

PR #18976, opened 2026-03-24
Use idiomatic Kotlin in custom filter documentation

therepanic:
This will make Kotlin and all users more native and readable.

Closes: gh-18967

### DIFF
--- a/docs/src/test/java/org/springframework/security/docs/servlet/addingcustomfilter/CustomFilterBeforeTests.java
+++ b/docs/src/test/java/org/springframework/security/docs/servlet/addingcustomfilter/CustomFilterBeforeTests.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2004-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.docs.servlet.addingcustomfilter;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.access.AccessDeniedException;
+import org.springframework.security.core.userdetails.User;
+import org.springframework.security.core.userdetails.UserDetails;
+import org.springframework.security.core.userdetails.UserDetailsService;
+import org.springframework.security.provisioning.InMemoryUserDetailsManager;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+import org.springframework.test.context.web.WebAppConfiguration;
+import org.springframework.test.web.servlet.MockMvc;
+import org.springframework.test.web.servlet.setup.MockMvcBuilders;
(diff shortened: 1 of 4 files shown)


docs/modules/ROOT/pages/servlet/architecture.adoc
```diff
@@ -246,9 +246,15 @@ Instead of implementing `Filter`, you can extend from {spring-framework-api-url}
...
+Based on the rule of thumb, you can add it before xref:servlet/authentication/logout.adoc[`LogoutFilter`], like so:
 
-include-code::./SecurityConfig[tag=snippet,indent=0]
+include-code::./SecurityConfig[tag=snippet-before,indent=0]
+
+<1> Use `HttpSecurity#addFilterBefore` to add the `TenantFilter` before the `LogoutFilter`.
+
+Or after xref:servlet/authentication/anonymous.adoc[`AnonymousAuthenticationFilter`], the last authentication filter in the chain, like so:
+
+include-code::./SecurityConfig[tag=snippet-after,indent=0]
```
therepanic: I think it would be better to show `before` and then `after`, rather than `after` and then `before`.

therepanic:
Tests failing due to https://github.com/spring-projects/spring-security/commit/2a8976f2f083218b358545eaa900c306e54254f8

jzheaux:
Great, @therepanic! This will merge into `main` once the build passes.

docs/src/test/java/org/springframework/security/docs/servlet/addingcustomfilter/SecurityConfig.java
```diff
@@ -22,21 +22,32 @@
...
 @Configuration
 @ContextConfiguration(classes = { SecurityConfig.class })
 @EnableWebSecurity
 public class SecurityConfig {
 
-	// tag::snippet[]
+	// tag::snippet-before[]
 	@Bean
-	public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
+	public SecurityFilterChain filterChainBefore(HttpSecurity http) throws Exception {
```
jzheaux: To have both of these declared in the same class, at least one will need to have an `@Order` value and a `securityMatchers` declaration. Otherwise, they are both configured to work for all requests, which Spring Security does not allow.

Alternatively, you can place them in different configurations for different `Tests` files.

jzheaux (changes requested):
Thanks for the updates, @therepanic. I've left some feedback inline.

docs/src/test/java/org/springframework/security/docs/servlet/addingcustomfilter/CustomFilterBeforeTests.java
```diff
@@ -47,7 +47,7 @@
 @ContextConfiguration(classes = {
 		CustomFilterTests.UserDetailsConfig.class,
 		CustomFilterTests.ApiController.class,
-		SecurityConfig.class })
+		SecurityConfigBefore.class })
```
therepanic, line 50: Since this test class was using the old `SecurityConfig`, we now make it use `SecurityConfigBefore` to have the same behavior as before.

docs/src/test/kotlin/org/springframework/security/kt/docs/servlet/addingcustomfilter/CustomFilterAfterTests.kt
```diff
@@ -45,13 +45,13 @@ import org.springframework.web.bind.annotation.RestController
...
-		CustomFilterTests.ApiController::class,
-		SecurityConfig::class
+		CustomFilterAfterTests.UserDetailsConfig::class,
+		CustomFilterAfterTests.ApiController::class,
+		SecurityConfigAfter::class
 	]
 )
 @WebAppConfiguration
-class CustomFilterTests {
+class CustomFilterAfterTests {
```
therepanic, line 55: I suppose we can just copy the existing tests to check if they work with `after`?

jzheaux:
Great! This is now merged into `main`.

docs/src/test/kotlin/org/springframework/security/kt/docs/servlet/addingcustomfilter/CustomFilterAfterTests.kt
jzheaux: Yes, this in this simple case, `addBeforeFilter` and `addAfterFilter` achieve largely the same thing, so it's understandable if their tests look quite similar.

Issue #18960, opened 2026-03-20
Add `AllRequiredFactorsAuthorizationManager.anyOf`


rwinch:
In order to support something like webauthn OR (password and OTT) and get the correct `AuthorizationResult` when access is denied, we should create a built in way to perform a logical or on `AllRequiredFactorsAuthorizationManager` instances that is smart enough to combine the `FactorAuthorizationDecision` into a meaningful response. Something like this comes to mind

```java
        public AuthorizationResult authorize(Supplier<? extends @Nullable Authentication> authentication, T object) {
            List<RequiredFactorError> factorErrors = new ArrayList<>();
            for (AllRequiredFactorsAuthorizationManager<T> factor : this.factors) {
                FactorAuthorizationDecision result = factor.authorize(authentication, object);
                if (result.isGranted()) {
                    return result;
                }
                factorErrors.addAll(result.getFactorErrors());
            }
            return new FactorAuthorizationDecision(factorErrors);
        }
```

Since it is bound to `AllRequiredFactorsAuthorizationManager` would likely be available as `public <T> static AuthorizationManager<T> anyOf(AllRequiredFactorsAuthorizationManager<T>... managers)`

NOTE: `AuthorizationManagers.anyOf` is not sufficient because it does not know how to combine the results properly. Nor should it be updated to be aware of ``FactorAuthorizationDecision`.`

evgeniycheban:
Hi @rwinch, can I work on this one?

ziqin:
I like AnyFactorAuthorizationManager. It is expressive because theoretically any combinatorial boolean expression can be normalized as sum of products (though we don’t support negation here in AnyFactorAuthorizationManager).

To make it more robust, we may need to consider deduplicating factor errors.

Say we want to express “any 2 of Password, OTT, X509” using AnyFactorAuthorizationManager. The sum-of-products would be “(Password ∧ OTT) ∨ (Password ∧ X509) ∨ (OTT ∧ X509)”.
Suppose the user is authenticating with only its password, if each error is simply appended to the list, the final factorErrors list we get would be [OTT, X509, OTT,  X509], which may not be what we want.

rwinch:
@evgeniycheban Yes please work on this. Please see @ziqin comment which is valid

evgeniycheban:
Hi @rwinch, just created a PR, please have a look when you have time.

@ziqin, please join the review as well, thanks.

PR #18997, opened 2026-03-27
Add AllRequiredFactorsAuthorizationManager.anyOf

evgeniycheban:
Closes gh-18960

### DIFF
--- a/core/src/main/java/org/springframework/security/authorization/AllRequiredFactorsAuthorizationManager.java
+++ b/core/src/main/java/org/springframework/security/authorization/AllRequiredFactorsAuthorizationManager.java
@@ -20,11 +20,11 @@ import java.time.Clock;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/springframework/security/authorization/AllRequiredFactorsAuthorizationManager.java
```diff
@@ -49,6 +52,31 @@ public final class AllRequiredFactorsAuthorizationManager<T> implements Authoriz
...
 
+	/**
+	 * Creates an {@link AuthorizationManager} that grants access if at least one
+	 * {@link AllRequiredFactorsAuthorizationManager} granted, collects
+	 * {@link RequiredFactorError}s omitting duplicate errors of the same factor.
+	 * @param <T> the type of object that is being authorized
+	 * @param managers the {@link AllRequiredFactorsAuthorizationManager}s to use
+	 * @return the {@link AuthorizationManager} to use
+	 * @since 7.1
+	 */
```
ziqin, line 65: Would you consider adding a `@see` tag in the Javadoc of `AuthorizationManagers#anyOf(AuthorizationManager...)` to remind users of the existence of this method?

evgeniycheban:
> It is worth consideration whether we should deduplicate errors caused by `RequiredFactor`s sharing the same `authority` but having different `validDuration`s.

I also thought about that, and it's a really valid point, let's wait what @rwinch thinks about it.

rwinch:
Thanks for the contribution @evgeniycheban I've merged the PR along with [some polish](https://github.com/spring-projects/spring-security/pull/18997/changes/ff820a868e7baf5e690a081f2f4d7c5380f0aa69) into main. In response to uniqueness by the requiredFactory, I did change this to be based upon the a `Set` to take into account the other members on `RequiredFactor`.

Issue #16920, opened 2025-04-10
Support custom CAs in oauth login


gbaso:
**Expected Behavior**

The internal rest clients should trust custom CAs.

**Current Behavior**

The internal clients only support CAs in the jdk truststore.

**Context**

In corporate environments sometime the identity provider is part of the corporate network and has an HTTPS certificate that is signed by the corporate CA.

Applications with oauth/oidc login perform several requests to the IdP, for retrieving metadata or for exchanging tokens. This is done with internal clients, for example in [ClientRegistrations](https://github.com/spring-projects/spring-security/blob/546dba7dc2657e22466ef38098ea898668ad1dc4/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java#L199) or in [AbstractRestClientOAuth2AccessTokenResponseClient](https://github.com/spring-projects/spring-security/blob/546dba7dc2657e22466ef38098ea898668ad1dc4/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/AbstractRestClientOAuth2AccessTokenResponseClient.java#L137), that are not configurable. While the latter allows to replace the `RestClient`, at the cost of having to reconfigure the message converters and status handler, the former does not.

The only avenue then is to add your custom CAs to the jdk truststore, usually done with the [ca-certificates](https://github.com/paketo-buildpacks/ca-certificates) buildpack. Unfortunately, loading CAs at runtime [doesn't seem to work in native image](https://github.com/spring-cloud/spring-cloud-bindings/issues/112).

It would be useful to be able to specify ssl bundles when configuring registrations via properties, and for those bundles to be applied to the internal clients.

lbruun:
Similar issue encountered.

When the OAuth2 Authorization server is using a custom CA you really need to jump through an incredible amount of hoops.

I believe what is often missed is the fact that in a high-security environment we cannot just add CAs at the JVM level. In such an environment a CA is often only applicable for a specific outbound http connection, it must not apply to _any_ outbound http connection that the application makes.

Another issue encountered is the fact the an application may interact with _several_ OAuth2 Authorization Servers. And the same argument again: a given CA applies to a specific one. Not _any_ outbound connection to an OAuth2 Authorization Server.

## What we have

### Discovery or provider's config/metadata

[...]

Issue #18992, opened 2026-03-26
Relax OIDC discovery validation to support Keycloak Dynamic Back Channel URL


krezovic:
Not to be confused with Back Channel Logout, Dynamic Back Channel URL allows communication from OAuth2 Clients using internal-only URLs while all user facing communication goes through a predefined external URL.

https://www.keycloak.org/server/hostname#_utilizing_an_internal_url_for_communication_among_clients

Consider the Kubernetes Deployment where Keycloak is deployed and exposed via Kubernetes Service, reachable from Spring Boot backend deployed on the same cluster using Kubernetes Cluster IP resolved via Service Name

http://keycloak-http.keycloak.svc.cluster.local/realms/master/.well-known/openid-configuration

The Kubernetes Service is exposed via External Load Balancer, using either Ingress or Gateway API, and is available to external users via https://keycloak.example.com/realms/master.

The second value is an issuer and I can specify it in Spring Security properties. This however means that my backend communicate with Keycloak instance, which is in the same cluster, in a way that it leaves the cluster, and then comes back to it via load balancer.

Here is the full OIDC discovery metadata received when calling the internal URL above

```json
{
  "issuer": "https://keycloak.example.com/realms/master",
  "authorization_endpoint": "https://keycloak.example.com/realms/master/protocol/openid-connect/auth",
  "token_endpoint": "http://keycloak-http.keycloak.svc.cluster.local/realms/master/protocol/openid-connect/token",
  "introspection_endpoint": "http://keycloak-http.keycloak.svc.cluster.local/realms/master/protocol/openid-connect/token/introspect",
  "userinfo_endpoint": "http://keycloak-http.keycloak.svc.cluster.local/realms/master/protocol/openid-connect/userinfo",
  "end_session_endpoint": "https://keycloak.example.com/realms/master/protocol/openid-connect/logout",
  "frontchannel_logout_session_supported": true,
  "frontchannel_logout_supported": true,
  "jwks_uri": "http://keycloak-http.keycloak.svc.cluster.local/realms/master/protocol/openid-connect/certs",
  "check_session_iframe": "https://keycloak.example.com/realms/master/protocol/openid-connect/login-status-iframe.html",
  "grant_types_supported": [
    "authorization_code",
    "client_credentials",
    "implicit",
    "password",
    "refresh_token",
    "urn:ietf:params:oauth:grant-type:device_code",
    "urn:ietf:params:oauth:grant-type:token-exchange",
[...]

PR #19011, opened 2026-03-31
Add RestTemplate overloads to ClientRegistrations methods

### LINKED ISSUE #16833: Adjust ClientRegistrations to allow to inject own instance of RestTemplate, ideally to use RestClient instead of ResTemplate
**Expected Behavior**

`ClientRegistrations` should provide ability to inject own instance of `RestTemplate`, ideally if it would be `RestClient`, instead of `RestTemplate`.

**Current Behavior**

Right now  `org.springframework.security.oauth2.client.registration.ClientRegistrations` creates instance of `RestTemplate` and doesn't provide any ability to replace it with another instance.

**Context**

In some cases it is needed to use `RestTemplate` with disabled SSL check. Also it would be great to be able to use `RestClient` instead of `RestTemplate`, so it can be defined in declarative way with https://github.com/ch4mpy/spring-addons library. 

More context with alternative solutions can be found in comment https://github.com/ch4mpy/spring-addons/issues/254#issuecomment-2756709068.

vpelikh:
This PR introduces overloaded methods in `ClientRegistrations` that accept a custom `RestTemplate`:

- `fromOidcIssuerLocation(RestTemplate restTemplate, String issuer)`
- `fromIssuerLocation(RestTemplate restTemplate, String issuer)`

Fixes #16833 #16920

### Motivation
Currently, the discovery endpoints (OIDC and OAuth2) are called using a static `RestTemplate` with fixed connection/read timeouts. This limits customization for users who need to:

- Configure a proxy
- Set custom timeouts
- Add interceptors (e.g., for logging or authentication)
- Use a different HTTP client (e.g., with mutual TLS)

By providing overloads that accept a `RestTemplate`, users can now inject their own configured instance while reusing the existing discovery logic.

### Changes
- Added two new public static methods as described above.
- The existing methods (`fromOidcIssuerLocation(String)` and `fromIssuerLocation(String)`) delegate to the new overloads using the default static `RestTemplate`.
- Internal `oidc(...)`, `oidcRfc8414(...)`, `oauth(...)` and `getRfc8414Builder(...)` methods were updated to accept a `RestTemplate` parameter; the existing static supplier-based approach remains unchanged.
- Added comprehensive tests for the new overloads, verifying:
  - Success scenarios (OIDC, OAuth2, and fallback flows)
  - Null `RestTemplate` or empty issuer handling
  - Error cases with proper exception propagation

### Impact
- Backward compatible: existing callers are unaffected.
- Users can now leverage their own `RestTemplate` for discovery requests.

### Testing
- All existing tests pass.
- New tests cover the added methods, using `MockWebServer` to simulate various discovery endpoints and responses.

Please review and let me know if any additional changes are needed.

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/registration/ClientRegistrations.java
@@ -146,9 +146,14 @@ public final class ClientRegistrations {
 	 * @return a {@link ClientRegistration.Builder} that was initialized by the OpenID
 	 * Provider Configuration.
 	 */
-	public static ClientRegistration.Builder fromOidcIssuerLocation(String issuer) {
+	public static ClientRegistration.Builder fromOidcIssuerLocation(RestTemplate restTemplate, String issuer) {
+		Assert.notNull(restTemplate, "restTemplate cannot be null");
 		Assert.hasText(issuer, "issuer cannot be empty");
-		return getBuilder(issuer, oidc(issuer));
+		return getBuilder(issuer, oidc(restTemplate, issuer));
+	}
+
+	public static ClientRegistration.Builder fromOidcIssuerLocation(String issuer) {
+		return fromOidcIssuerLocation(rest, issuer);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


vpelikh:
Hey @jgrandja! Can you check this one?

jgrandja:
@vpelikh We do not want to expose `RestTemplate` in the API.

Please see this [comment](https://github.com/spring-projects/spring-security/issues/18992#issuecomment-4170221638) as we plan to address this in the `7.2` release cycle.

PR #19016, opened 2026-04-01
Complete nullability annotations in remaining equals methods

therepanic:
In this commit, we'll finally add nullability annotations to the equals method for consistency with other projects. We've already done this, but we recently added jspecify support to new modules.

### DIFF
--- a/webauthn/src/main/java/org/springframework/security/web/webauthn/api/ImmutablePublicKeyCredentialUserEntity.java
+++ b/webauthn/src/main/java/org/springframework/security/web/webauthn/api/ImmutablePublicKeyCredentialUserEntity.java
@@ -137,7 +137,7 @@ public final class ImmutablePublicKeyCredentialUserEntity implements PublicKeyCr
 	}
 
 	@Override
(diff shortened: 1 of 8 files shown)


therepanic:
This PR is a follow-up to: https://github.com/spring-projects/spring-security/pull/18930

BTW, the build crashes not because of the current changes https://github.com/spring-projects/spring-security/actions/runs/23822120186/job/69436701182/

webauthn/src/main/java/org/springframework/security/web/webauthn/api/ImmutablePublicKeyCredentialUserEntity.java
```diff
@@ -137,7 +137,7 @@ public static PublicKeyCredentialUserEntityBuilder builder() {
 	}
 
 	@Override
-	public boolean equals(Object obj) {
+	public boolean equals(@Nullable Object obj) {
 		if (this == obj) {
```
therepanic, line 141: This module had jspecify support before, I guess I accidentally missed it.

PR #19023, opened 2026-04-02
Deprecate Hex in favor of `java.util.HexFormat`

### LINKED ISSUE #16867: Depecate org.springframework.security.crypto.codec.Hex
Spring Security requires a minimum of Java 17, which has the class [`java.util.Hexformat`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/HexFormat.html).

Usage of the `Hex` class should be deprecated and replaced, as the `Base64` codec class was.

therepanic:
The main point is that java.util.HexFormat essentially does the same thing as our Hex. We had a similar candidate with Base and deprecated it in favor of java.util.Base64. Moreover, we've already removed it: https://github.com/therepanic/spring-security/commit/2f53a2edb34b690ba2f533732d62f7a58a55d577.

Closes: gh-16867

### DIFF
--- a/crypto/src/main/java/org/springframework/security/crypto/codec/Hex.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/codec/Hex.java
@@ -24,7 +24,9 @@ package org.springframework.security.crypto.codec;
  *
  * @author Luke Taylor
  * @since 3.0
+ * @deprecated Use java.util.HexFormat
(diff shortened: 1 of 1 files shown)


therepanic:
If we are satisfied with this, I can create an issue with a fix where we will replace all usage with the new Java 17 api.

Issue #18975, opened 2026-03-24
Make FactorGrantedAuthority Builder Constructor Public


re1709:
Allow use of the FactorGrantedAuthority.Builder constructor outside of the FactorGrantedAuthority class.

This would allow creating a FactorGrantedAuthority and setting an issuedAt value which isn't just the current date & time, which is currently the only possibility when creating a FactorGrantedAuthority.

For example, mfa last issued at (along with user agent and ip address) can be stored (for example in a database) so the user isn't prompted for X amount of time when next logging in.

It would make sense to then grant the FactoryAuthority with the issued at time stored in the database. So endpoints protected by a time duration for mfa would then trigger - e.g. user logs in the following day but isn't prompted for mfa at log in but should be for a secure page such as api token (which is configured in security to have mfa recently, such as 15 mins).

hqqw2h-lgtm:
I'd like to work on this. Will make the FactorGrantedAuthority Builder constructor public.

re1709:
Thanks @hqqw2h-lgtm for picking this up. Just seen the comments on the pull request...very sorry don't know how i didn't realise those methods exist. Will close this.

PR #19035, opened 2026-04-04
Make FactorGrantedAuthority Builder constructor public

hqqw2h-lgtm:
## Summary

Make the `FactorGrantedAuthority.Builder` constructor public so that external code can create a `Builder` directly and set a custom `issuedAt` value.

This enables scenarios like persisting MFA factor timestamps (along with user agent and IP address) so the user is not prompted for a configurable period when next logging in.

- Changed `Builder(String authority)` from `private` to `public`
- Added test to verify direct `Builder` construction

Closes gh-18975

### DIFF
--- a/core/src/main/java/org/springframework/security/core/authority/FactorGrantedAuthority.java
+++ b/core/src/main/java/org/springframework/security/core/authority/FactorGrantedAuthority.java
@@ -184,7 +184,7 @@ public final class FactorGrantedAuthority implements GrantedAuthority {
 
 		private @Nullable Instant issuedAt;
 
-		private Builder(String authority) {
+		public Builder(String authority) {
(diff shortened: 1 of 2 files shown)


jzheaux:
Hi, @hqqw2h-lgtm, thanks for the PR.

I'm not sure I understand the value of the current just yet. Why are the static factories, withFactor and withAuthority, insufficient in your case? Perhaps you could demonstrate in your unit testing what you can do now that couldn't be done before.

hqqw2h-lgtm:
Thanks for the feedback. You're right that the existing static factories (withAuthority/withFactor) already support setting issuedAt via the builder, so this constructor visibility change does not add sufficient value. Closing this PR.

Issue #14656, opened 2024-02-27
Spring Security SAML is not FIPS compliance as OpenSAML does not support BC FIPS


sumeetpri:
Hello Team,

I'm currently spring security 6.2 which internally uses OpenSAML 4.3 Java library to handle the SAML assertion received from the IDP. However, I've encountered an issue where OpenSAML relies on the bcprov-jdk18on library, which is not compliant with FIPS standards. I integrated bc-fips version 1.0.2.4. However, this change has led to numerous "class not found" errors, and the system is not functioning correctly. 

Spring Security SAML uses OpenSAML 4.3  which has hard dependency with non FIPS library which makes   Spring Security SAML as not useable for federal projects .

sumeetpri:
Hi @jzheaux , looking forward for your input as software like spring security SAML becoming non fips compliance is major blocker to be used in federal project

sumeetpri:
Please suggest if there is any alternative option available

rwinch:
@sumeetpri Can you put together a minimal sample that reproduces the errors that you are seeing?

alla-gofman:
The class org.opensaml.security.crypto.ec.curves.AbstractNamedCurve has  
import to org.bouncycastle.jce.ECNamedCurveTable
You force to import the bcprov-jdk18on library, but I need to work with FIPS compliant library bc-fips instead.

Refer to AbstractNamedCurve.buildParameterSpecmethod documentation:
```Build an instance of ECParameterSpec corresponding to this curve.
The default implementation here is that it first attempts to resolve the curve from Bouncy Castle's ECNamedCurveTable. If that is unsuccessful then it attempts a brute force approach by generating a key pair using a ECGenParameterSpec based on the curve's name from getName(), returning the parameter instance from the resulting ECPublicKey.
Returns:
the parameter spec instance, or null if can not be built
  Maven: org.opensaml:opensaml-security-api:4.3.1 (opensaml-security-api-4.3.1-sources.jar)
```
So you should also treat ClassNotFoundException in case of unsuccessful attempt to resolve the curve from Bouncy Castle's ECNamedCurveTable.
```
Caused by: java.lang.ClassNotFoundException: org.bouncycastle.jce.ECNamedCurveTable
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641) ~[?:?]
[...]

PR #19059, opened 2026-04-11
Announce external OpenSAML init in spring-security

strehle:
Problem: see details of https://github.com/spring-projects/spring-security/issues/14656

Solution. Allow to mark with "initializedAlready" (name can be discussed), that OpenSAML was initialized outside of spring-security

There are situation, e.g. omit an algorithm and change one because you use FIPS, then you need another way of opensaml initialization and spring should not try to init it again.

This PR is in order to solve https://github.com/cloudfoundry/uaa/issues/3811

### DIFF
--- a/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/core/OpenSamlInitializationService.java
+++ b/saml2/saml2-service-provider/src/main/java/org/springframework/security/saml2/core/OpenSamlInitializationService.java
@@ -88,7 +88,12 @@ public final class OpenSamlInitializationService {
 	 * @throws Saml2Exception if OpenSAML failed to initialize
 	 */
 	public static boolean initialize() {
-		return initialize((registry) -> {
+		return initialize(true, (registry) -> {
+		});
+	}
+
+	public static boolean initializedAlready() {
+		return initialize(false, (registry) -> {
 		});
 	}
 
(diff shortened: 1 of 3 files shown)


jzheaux:
Thanks for the contribution, @strehle. I will not be able to review this likely until after the 7.1 release in May.

In the meantime, can you confirm if https://github.com/spring-projects/spring-security/issues/14656#issuecomment-2442335475 works in your case?

strehle:
> Thanks for the contribution, @strehle. I will not be able to review this likely until after the 7.1 release in May.
> 
> In the meantime, can you confirm if [#14656 (comment)](https://github.com/spring-projects/spring-security/issues/14656#issuecomment-2442335475) works in your case?

Yes it works and I would also provide PRs for your downports, means 6.5.x because in CF we first had the challenge to decouple from the fix version of opensaml, e.g. 
https://github.com/cloudfoundry/uaa/releases/tag/v78.11.0

and then as preparation before boot 4.x we will do
https://github.com/cloudfoundry/uaa/pull/3840

Therefore I can provide a similar PR for your opensaml4 integration. 

---
because of the need for BC-FIPS we have forked parts of spring security but we will also remove that step by step.... before that PR, I have verified other approaches from other projects , e.g. https://github.com/elastic/elasticsearch/pull/98785 but that was also somehow dangerous, so we know that SAML with EC keys will have restrictions, but this combination is not that common in enterprises

strehle:
@jzheaux any timeline to get this merged ?
I ask because it could help other projects to get again a latest opensaml together with spring security .
UAA and the whole public CF community is running with this since some month now, so it is working - I would say , FYI, @fhanik ?

PR #19073, opened 2026-04-14
Fix `InetAddressMatchers#matchExternal` to exclude null and any local addresses

### LINKED ISSUE #19072: `InetAddressMatchers.matchExternal()` matches `null` and `0.0.0.0`
**Describe the bug**

`InetAddressMatchers.matchExternal().build()` returns a negated `matchInternal()` builder, however, I'm not sure that `null` or `0.0.0.0` should be classified as external addresses.

**To Reproduce**

```java
InetAddressMatcher matcher = InetAddressMatchers.matchExternal().build();
assertThat(matcher.matches(null)).isFalse();
assertThat(matcher.matches("0.0.0.0")).isFalse();
assertThat(matcher.matches("::")).isFalse();
``` 

**Expected behavior**
See above

**Sample**
See above

therepanic:
This commit fixes the incorrect behavior of
`InetAddressMatchers#matchExternal`, where `matchExternal` returns an incorrect answer when passing a null / local address argument, which is not reliable.

Closes: gh-19072

### DIFF
--- a/core/src/main/java/org/springframework/security/util/matcher/InetAddressMatchers.java
+++ b/core/src/main/java/org/springframework/security/util/matcher/InetAddressMatchers.java
@@ -31,6 +31,7 @@ import org.springframework.util.Assert;
  * strategies for IP addresses.
  *
  * @author Rob Winch
+ * @author Andrey Litvitski
  * @since 7.1
  */
 public final class InetAddressMatchers {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/springframework/security/util/matcher/InetAddressMatchers.java
```diff
@@ -256,6 +257,9 @@ public boolean matches(@Nullable InetAddress address) {
 			if (address == null) {
 				return false;
 			}
+			if (address.isAnyLocalAddress()) {
+				return true;
+			}
 			if (address.isLoopbackAddress() || address.isLinkLocalAddress() || address.isSiteLocalAddress()) {
```
therepanic, line 263: This would probably be more elegant than adding the same check but on `matchExternal`, since the logic of `matchExternal` is to return the response back from `matchInternal`.

PR #19074, opened 2026-04-14
Allow fluent use of `InetAddressMatchers#builder`

### LINKED ISSUE #19071: Allow `InetAddressMatchers.builder().include(...)` to be called in a fluent style
**Expected Behavior**

`InetAddressMatchers.builder().includeAddresses(...)` returns `this` for fluent style calls, however, calling `includeAddresses` again doesn't work in a useful way.

**Current Behavior**

`InetAddressMatchers.builder().includeAddresses(...)` accepts a list of include addresses and the javadoc states "Adds an include list matcher that permits only the specified addresses.". If you call `InetAddressMatchers.builder().includeAddresses(List.of("192.168.1.1", "192.168.1.2")).build()` you'll get a matcher that matcher "192.168.1.1" or 192.168.1.2".

If however, you call the method more than once, e.g. `InetAddressMatchers.builder().includeAddresses(List.of("192.168.1.1")).includeAddresses(List.of("192.168.1.2")).build()` you get a matcher that won't match anything. This is because internally there are two `IncludeListInetAddressMatcher` instances. One will only match 192.168.1.1 and the other will only match 192.168.1.2.

**Context**

This came up during the development of https://github.com/spring-projects/spring-boot/issues/49687. You could argue that the API is working as designed, so perhaps this issue can just be closed.

therepanic:
The main goal of this commit is to allow fluent use of `InetAddress Matchers#builder` to be called in a fluent manner.

Previously, calling `includeAddresses` twice would cause the logic to stop working, since our IP address can't be equal to two addresses in the list at the same time.

I think the current behavior is rather non-obvious, and we should definitely implement this change.

Closes: gh-19071

### DIFF
--- a/core/src/main/java/org/springframework/security/util/matcher/InetAddressMatchers.java
+++ b/core/src/main/java/org/springframework/security/util/matcher/InetAddressMatchers.java
@@ -18,7 +18,11 @@ package org.springframework.security.util.matcher;
 
 import java.net.InetAddress;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Set;
+import java.util.function.Function;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -31,6 +35,7 @@ import org.springframework.util.Assert;
  * strategies for IP addresses.
  *
  * @author Rob Winch
+ * @author Andrey Litvitski
  * @since 7.1
  */
 public final class InetAddressMatchers {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/springframework/security/util/matcher/InetAddressMatchers.java
```diff
@@ -86,42 +91,44 @@ public static InetAddressMatcher fromIpAddress(String ipAddress) {
...
 	 * @author Rossen Stoyanchev
 	 * @author Rob Winch
+	 * @author Andrey Litvitski
 	 */
 	public static final class Builder {
 
-		private final List<InetAddressMatcher> matchers = new ArrayList<>();
+		private final Set<String> includeAddresses = new LinkedHashSet<>();
+
+		private final Set<String> excludeAddresses = new LinkedHashSet<>();
```
therepanic, line 100: Would it be more obvious if we used `LinkedHashSet`? When debugging, the user will see the addresses added to the list in the order they were added.

PR #19089, opened 2026-04-16
Fix CVE-2026-22732: Track Content-Length for setHeader, setIntHeader,…

Chhida:
> … Backport

… addIntHeader

When applications specify HTTP response headers for servlet applications using Spring Security, there is the possibility that the HTTP Headers will not be written. This can open up applications to various attacks including exposing sensitive data via caching mechanisms.

Add Content-Length tracking to setHeader, setIntHeader, and addIntHeader methods in OnCommittedResponseWrapper to ensure HTTP headers are properly written.


jgrandja:
Closing this as the fix for `6.4.x` is available in [Enterprise Support Only](https://spring.io/security/cve-2026-22732).

PR #19102, opened 2026-04-18
Support pepper in Password4j encoders

### LINKED ISSUE #18299: Allow setting of shared secret (pepper) for password storage
**Expected Behavior**

When using the `Password4j` password encoders, I can optionally provide a shared secret/pepper that is then used as the pepper in the `Password4j` library (and therefore is not part of the stored hash value).

**Current Behavior**

Currently, it is not possible to use a shared secret pepper with Spring Security's default password encoders. To use a pepper, one must write a custom password encoder.

**Context**

We are building a system where we want a shared secret/pepper, as described in the OWASP Password Storage Cheat Sheet. Since logins are distributed across multiple devices, including offline devices, we would like to provide a shared secret/pepper to the application through another secure means so it can be used in the password storage process.

therepanic:
Currently, we can't forward pepper unless we encounter an anti-pattern and implement our own encoder.

To solve this problem, we need to provide pepper support out of the box, specifically in `Password4jPasswordEncoder` and all its subclasses, as well as `BalloonHashingPassword4jPasswordEncoder` and `Pbkdf2Password4jPasswordEncoder`, which aren't direct implementations of `Password4jPasswordEncoder` but also use `password4j` and require pepper support.

Closes: gh-18299

### DIFF
--- a/crypto/src/main/java/org/springframework/security/crypto/password4j/BcryptPassword4jPasswordEncoder.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/password4j/BcryptPassword4jPasswordEncoder.java
@@ -18,6 +18,7 @@ package org.springframework.security.crypto.password4j;
 
 import com.password4j.AlgorithmFinder;
 import com.password4j.BcryptFunction;
+import org.jspecify.annotations.Nullable;
 
 /**
  * Implementation of {@link org.springframework.security.crypto.password.PasswordEncoder}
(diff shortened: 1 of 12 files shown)


crypto/src/main/java/org/springframework/security/crypto/password4j/BcryptPassword4jPasswordEncoder.java
```diff
@@ -66,7 +68,20 @@ public BcryptPassword4jPasswordEncoder() {
 	 * @throws IllegalArgumentException if bcryptFunction is null
 	 */
 	public BcryptPassword4jPasswordEncoder(BcryptFunction bcryptFunction) {
-		super(bcryptFunction);
+		this(bcryptFunction, null);
```
therepanic, line 71: i guess this is ok?

PR #19132, opened 2026-04-28
Update jwt.adoc

zariok:
Fix: typo validateTypes -> validateType

https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/oauth2/jwt/NimbusReactiveJwtDecoder.JwkSetUriReactiveJwtDecoderBuilder.html#validateType(boolean)


zariok:
Worth noting it appears all builders have validateType(false) as the default and this could probably be removed entirely.

PR #19144, opened 2026-05-02
Add OneTimeUse support for SAML assertions

### LINKED ISSUE #19130: Add OneTimeUse support for SAML assertions
It would be nice if Spring Security SAML had out-of-the-box support for `<OneTimeUse>` elements. With the introduction of `OpenSaml5AuthenticationProvider.AssertionValidator`, this is much easier to do than before by way of adding something like:
                                                                                                                                                                             
```java
Builder replayCache(Cache cache)
```

to `OpenSaml5AuthenticationProvider.AssertionValidator.Builder`.
                                                                                                                                                                             
Internally, the builder could adapt the Spring `Cache` implementation to an OpenSAML `ReplayCache` and by virtue of this, construct an OpenSAML `OneTimeUseConditionValidator`.

Note that `ReplayCache` includes expiry semantics in its `check` method, but Spring Cache does not. As such, the adapter will need to account for the TTL in its cache lookup, for example by including the expiry time in the cached element. The key itself can be the assertion ID; we can be more strict about this since applications can quite easily construct a `OneTimeUserConditionValidator` if they want different behavior.

For passivity reasons, this is strictly an opt-in feature; for applications not using this feature, they remain responsible for validating `OneTimeUse` elements as they already are today.

Please include unit tests, JavaDoc, and reference documentation.

therepanic:
This change will allow us to add `<OneTimeUse>` elements out of the box. To do this, we can pass the cache through
`OpenSaml5AuthenticationProvider.AssertionValidator.Builder#replayCache` and add the condition itself directly when building the builder. I also created a package-private `SpringCacheReplayCache` that implements `ReplayCache` and is needed to work with the constructor of `OneTimeUseConditionValidator`.

Closes: gh-19130

### DIFF
--- a/saml2/saml2-service-provider/src/opensaml5Main/java/org/springframework/security/saml2/provider/service/authentication/OpenSaml5AuthenticationProvider.java
+++ b/saml2/saml2-service-provider/src/opensaml5Main/java/org/springframework/security/saml2/provider/service/authentication/OpenSaml5AuthenticationProvider.java
@@ -42,6 +42,7 @@ import org.opensaml.saml.saml2.assertion.SubjectConfirmationValidator;
 import org.opensaml.saml.saml2.assertion.impl.AudienceRestrictionConditionValidator;
 import org.opensaml.saml.saml2.assertion.impl.BearerSubjectConfirmationValidator;
 import org.opensaml.saml.saml2.assertion.impl.DelegationRestrictionConditionValidator;
+import org.opensaml.saml.saml2.assertion.impl.OneTimeUseConditionValidator;
 import org.opensaml.saml.saml2.assertion.impl.ProxyRestrictionConditionValidator;
 import org.opensaml.saml.saml2.core.Assertion;
 import org.opensaml.saml.saml2.core.Condition;
@@ -57,6 +58,7 @@ import org.opensaml.saml.saml2.encryption.Decrypter;
 import org.opensaml.xmlsec.signature.support.SignaturePrevalidator;
 import org.opensaml.xmlsec.signature.support.SignatureTrustEngine;
 
+import org.springframework.cache.Cache;
 import org.springframework.core.convert.converter.Converter;
 import org.springframework.security.authentication.AbstractAuthenticationToken;
 import org.springframework.security.authentication.AuthenticationProvider;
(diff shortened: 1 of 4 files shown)


saml2/saml2-service-provider/src/opensaml5Main/java/org/springframework/security/saml2/provider/service/authentication/OpenSaml5AuthenticationProvider.java
```diff
@@ -737,10 +740,11 @@ public static final class Builder {
 
 			private final Map<String, Object> validationParameters = new HashMap<>();
 
+			@Nullable private Cache replayCache;
+
 			private Builder() {
 				this.conditions.add(new AudienceRestrictionConditionValidator());
 				this.conditions.add(new DelegationRestrictionConditionValidator());
-				this.conditions.add(new ValidConditionValidator(OneTimeUse.DEFAULT_ELEMENT_NAME));
 				this.conditions.add(new ProxyRestrictionConditionValidator());
```
therepanic, line 748: Given how validation works, I don't think it would be a bad idea to move the logic for adding `OneTimeUse.DEFAULT_ELEMENT_NAME` into the `build` method itself

saml2/saml2-service-provider/src/opensaml5Main/java/org/springframework/security/saml2/provider/service/authentication/OpenSaml5AuthenticationProvider.java
```diff
@@ -819,11 +823,34 @@ public Builder subjectValidators(Consumer<List<SubjectConfirmationValidator>> su
...
 			/**
 			 * Build the {@link AssertionValidator}
 			 * @return the {@link AssertionValidator}
 			 */
 			public AssertionValidator build() {
+				if (this.replayCache != null) {
+					this.conditions
+						.add(new OneTimeUseConditionValidator(new SpringCacheReplayCache(this.replayCache), null));
+				}
+				else {
```
therepanic, line 851: I think we can afford to ignore `null` since we set the duration manually in `SpringCacheReplayCache#check`?

therepanic:
I left a couple of comments on topics that interest me, PTAL.

PR #19146, opened 2026-05-03
Catch `WebClientException` in reactive HaveIBeenPwned checker

### LINKED ISSUE #19142: Consider catching WebClientException in HaveIBeenPwndApiReactivePasswordChecker
`HaveIBeenPwnedRestApiPasswordChecker` (blocking) catches `RestClientException` on failure, which covers both HTTP error responses and lower-level I/O errors such as connection timeouts and DNS failures.

The reactive counterpart, `HaveIBeenPwnedRestApiReactivePasswordChecker`, uses `onErrorResume(WebClientResponseException.class, ...)`. `WebClientResponseException` only covers HTTP error responses (4xx/5xx). 

The semantic equivalent of `RestClientException` in the reactive stack is `WebClientException`. It seems more reasonable for the reactive one to use `onErrorResume(WebClientException.class, ...)` to align with the blocking implementation.

therepanic:
This change will ensure consistent behavior in the event of an I/O error with the non-reactive version of `HaveIBeenPwnedRestApiPasswordChecker`.

Closes: gh-19142

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/password/HaveIBeenPwnedRestApiReactivePasswordChecker.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/password/HaveIBeenPwnedRestApiReactivePasswordChecker.java
@@ -34,7 +34,7 @@ import org.springframework.security.crypto.codec.Hex;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 import org.springframework.web.reactive.function.client.WebClient;
-import org.springframework.web.reactive.function.client.WebClientResponseException;
+import org.springframework.web.reactive.function.client.WebClientException;
 
 /**
  * Checks if the provided password was leaked by relying on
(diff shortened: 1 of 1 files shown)


web/src/main/java/org/springframework/security/web/authentication/password/HaveIBeenPwnedRestApiReactivePasswordChecker.java
```diff
@@ -77,7 +78,7 @@ private Flux<String> getLeakedPasswordsForPrefix(String prefix) {
 			return Flux.empty();
 		})
 			.doOnError((ex) -> this.logger.error("Request for leaked passwords failed", ex))
-			.onErrorResume(WebClientResponseException.class, (ex) -> Flux.empty());
+			.onErrorResume(WebClientException.class, (ex) -> Flux.empty());
```
therepanic, line 81: There is no test in either the reactive or non-reactive version that checks this, so I think we should either do nothing, or add two tests - one for the reactive version and one for the non-reactive version.

Issue #18982, opened 2026-03-25
Migrate JDBC PersistentToken Support


jzheaux:
Since `JdbcDaoSupport` is deprecated, a migration strategy from `JdbcTokenRepositoryImpl` is needed.

Related to #18441

therepanic:
Hi, @jzheaux. Can I work on it?

jzheaux:
Given https://github.com/spring-projects/spring-security/pull/18820 and https://github.com/spring-projects/spring-security/pull/18820#discussion_r2991143747, let's see if @jkuhel wants to work on this ticket. Please decide between the two of you.

jkuhel:
Thank you @jzheaux! @therepanic if you don't mind, I would like to work on this issue.

@jzheaux Just to confirm I have the order and targeted release, would this be right?

**7.1.x**
1. #18987 
2. #18986
3. #18985

**8.0.x**
1. #18984

jzheaux:
@jkuhel, given that `RC1` goes out on Monday, which is historically the cutoff for feature development, I'm moving this to `7.2.x` to give everyone more time.

therepanic:
Hi, @jkuhel. You haven't been active on this issue for a while, and I figured you were too busy to work on it, so I decided to take the initiative and resolve almost the entire issue myself by submitting a PR: https://github.com/spring-projects/spring-security/pull/19163. I apologize if you’ve been actively working on this recently. In any case, we’ll still need to remove `JdbcTokenRepositoryImpl` (wait for the 8.0.0 branch) and document the migration to the new `JdbcPersistentTokenRepository` for version 8.0.0.

jkuhel:
> Hi, @jkuhel. You haven't been active on this issue for a while, and I figured you were too busy to work on it, so I decided to take the initiative and resolve almost the entire issue myself by submitting a PR: https://github.com/spring-projects/spring-security/pull/19163. I apologize if you’ve been actively working on this recently. In any case, we’ll still need to remove `JdbcTokenRepositoryImpl` (wait for the 8.0.0 branch) and document the migration to the new `JdbcPersistentTokenRepository` for version 8.0.0.

Hi @therepanic. Yes I actually have been working on it in between some other work that has taken higher priority. Perhaps it would be best to reach out prior to doing the work so as to not waste effort?
I appreciate you reaching out though.

PR #19163, opened 2026-05-09
Migrate `JdbcTokenRepositoryImpl` to `JdbcPersistentTokenRepository`

### LINKED ISSUE #18986: Deprecate `JdbcTokenRepositoryImpl`

### LINKED ISSUE #18987: Add `JdbcPersistentTokenRepository`

therepanic:
Since `JdbcDaoSupport` has been deprecated, we should deprecate our implementation of `JdbcTokenRepositoryImpl`, which directly inherits from `JdbcDaoSupport`. Instead of using `JdbcDaoSupport`, we are advised to inject `JdbcTemplate` or `JdbcClient` into the field, so we should create a new implementation of `JdbcPersistentTokenRepository`.

References: gh-18982
Closes: gh-18987
Closes: gh-18986

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/rememberme/JdbcPersistentTokenRepository.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/rememberme/JdbcPersistentTokenRepository.java
@@ -20,6 +20,10 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.Date;
 
+import javax.sql.DataSource;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.jspecify.annotations.Nullable;
 
 import org.springframework.core.log.LogMessage;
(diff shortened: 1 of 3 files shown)


web/src/main/java/org/springframework/security/web/authentication/rememberme/JdbcPersistentTokenRepository.java
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	public JdbcPersistentTokenRepository(DataSource dataSource) {
+		this.jdbcClient = JdbcClient.create(dataSource);
+	}
+
+	public JdbcPersistentTokenRepository(JdbcTemplate jdbcTemplate) {
+		this.jdbcClient = JdbcClient.create(jdbcTemplate);
+	}
+
```
therepanic, line 85: This is a great opportunity for us to start using the new `JdbcClient` API, which is fully compatible with the existing `JdbcClient`. Nice!

web/src/main/java/org/springframework/security/web/authentication/rememberme/JdbcTokenRepositoryImpl.java
```diff
@@ -34,7 +34,9 @@
  *
  * @author Luke Taylor
  * @since 2.0
+ * @deprecated Use {@link JdbcPersistentTokenRepository}
  */
+@Deprecated(since = "7.1.0")
 @SuppressWarnings("removal")
 public class JdbcTokenRepositoryImpl extends JdbcDaoSupport implements PersistentTokenRepository {
```
therepanic, line 41: I didn't mark it as "forRemoval" because otherwise the build would fail.

therepanic:
I don't think the 8.0.0 branch is available yet to remove `JdbcTokenRepositoryImpl`. In any case, aside from that, we'll need to use our new implementation by default in `RememberMeBeanDefinitionParser#parse` instead of the removed `JdbcTokenRepositoryImpl`.

PR #19175, opened 2026-05-13
Add EdDSA support to NimbusJwtEncoder

### LINKED ISSUE #17098: NimbusJwtEncoder does not support Edwards Curve signature (EdDSA) family algorithms.
**Description**
It is currently possible to create a `NimbusJwtEncoder` using an `ImmutableJWKSet` that contains an Ed25519 key (via `OctetKeyPair`), but it is not possible to use that encoder to generate a JWT signed with Ed25519 (EdDSA).

**Problem Details**

- `JwtEncoderParameters.from(header, claims)` does not allow custom algorithms in the header.
  - Only enums implementing `JwsAlgorithm` or `JwaAlgorithm` are allowed.
  - The default `SignatureAlgorithm` enum implements `JwsAlgorithm`, but does not include any EdDSA algorithms like `EdDSA` (`com.nimbusds.jose.JWSAlgorithm.EdDSA`).
- Even if a custom implementation of `JwsAlgorithm` is provided to represent EdDSA, it does not resolve the problem because of how `NimbusJwtEncoder` internally selects a signing key.
- During encoding, the `selectJwk` method is called. This uses a `JWKSelector` created with a matcher generated by the private `createJwkMatcher` method.
  - Unfortunately, `createJwkMatcher` does not support the EdDSA algorithm family, which results in the JWKSelector being constructed with a null matcher.
  - This causes `IllegalArgumentException` to be thrown at runtime due to the missing matcher.

When running following code:
```
val header = JwsHeader.with(FixedSignatureAlgorithm.Ed25519).build()
val claims = JwtClaimsSet.builder()
    .issuer(issuer)
    .issuedAt(Instant.now())
    .expiresAt(expiresAt)
    .claim(SESSION_ID_CLAIM_NAME, sessionId)
    .build()

jwtEncoder.encode(JwtEncoderParameters.from(header, claims)).tokenValue
```

```
public enum FixedSignatureAlgorithm implements JwsAlgorithm {
	RS256(JwsAlgorithms.RS256),
	RS384(JwsAlgorithms.RS384),
	RS512(JwsAlgorithms.RS512),
	ES256(JwsAlgorithms.ES256),
	ES384(JwsAlgorithms.ES384),
	ES512(JwsAlgorithms.ES512),
	PS256(JwsAlgorithms.PS256),
	PS384(JwsAlgorithms.PS384),
	PS512(JwsAlgorithms.PS512),
    EdDSA(JWSAlgorithm.EdDSA.getName()),
    Ed25519(JWSAlgorithm.Ed25519.getName());

	private final String name;

	FixedSignatureAlgorithm(String name) {
		this.name = name;
	}

	@Override
	public String getName() {
		return this.name;
	}

	public static FixedSignatureAlgorithm from(String name) {
		for (FixedSignatureAlgorithm value : values()) {
			if (value.getName().equals(name)) {
				return value;
			}
		}
		return null;
	}
}
```

then such error is thrown:
```
org.springframework.security.oauth2.jwt.JwtEncodingException: An error occurred while attempting to encode the Jwt: Failed to select a JWK signing key -> The JWK matcher must not be null
	at org.springframework.security.oauth2.jwt.NimbusJwtEncoder.selectJwk(NimbusJwtEncoder.java:123)
	at org.springframework.security.oauth2.jwt.NimbusJwtEncoder.encode(NimbusJwtEncoder.java:108)
Caused by: java.lang.IllegalArgumentException: The JWK matcher must not be null
	at com.nimbusds.jose.jwk.JWKSelector.<init>(JWKSelector.java:51)
	at org.springframework.security.oauth2.jwt.NimbusJwtEncoder.selectJwk(NimbusJwtEncoder.java:119)
	... 158 common frames omitted
```

**Suggested fix**
To support Ed25519/EdDSA, the following would need to be addressed:
1. Add EdDSA (JWSAlgorithm.EdDSA) support to the SignatureAlgorithm enum or allow flexible/custom algorithm headers in JwtEncoderParameters.
2. Update createJwkMatcher in NimbusJwtEncoder to recognize and match EdDSA algorithm family keys like OctetKeyPair with Curve.Ed25519.

**Used versions**
com.springframework.security:spring-security-oauth2-jose:6.3.3
com.nimbusds:nimbus-jose-jwt:9.40

therepanic:
NimbusJwtEncoder did not support EdDSA family algorithms. This commit adds support for signing JWTs with EdDSA keys.

Closes: gh-17098

**Note:** This is still a WIP and tests are still failing.


oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jose/jws/SignatureAlgorithm.java
```diff
@@ -79,7 +79,22 @@ public enum SignatureAlgorithm implements JwsAlgorithm {
...
+
+	/**
+	 * EdDSA signature algorithms (optional).
+	 */
+	EdDSA(JwsAlgorithms.EdDSA),
+
+	/**
+	 * EdDSA signature algorithms using Ed448 curve (optional).
+	 */
+	ED448(JwsAlgorithms.ED448),
```
therepanic: I think we should support this too.

therepanic:
An interesting observation: Nimbus uses `com.google.crypto.tink` for EdDSA. I wonder how much of a barrier this is for us. @jgrandja WDYT? Because this is the reason why, in fact, the tests are falling for now.

oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jose/jws/SignatureAlgorithm.java
gbaso: A small note: ED448 should be listed after ED25519, as the latter is based on the prime `2^255 - 19` and offers ~128 bits of security (similar to PS256), while the former is based on the larger prime `2^448 - 2^224 - 1`, and offers ~224 bits of security (between PS384 and PS512).

I think listing ED448 before ED25519 could give the wrong impression that it is cryptographically weaker.

therepanic: Thanks, that is a very good point. Exactly the same thing is done in `com.nimbusds.jose.JWSAlgorithm`.

ziqin:
Although Spring Security requires Java 17+, Nimbus still targets Java 7+ and is [still working on upgrading to Java 8](https://bitbucket.org/connect2id/nimbus-jose-jwt/issues/394/upgrade-to-java-8-tracks-multiple-tickets). I'm not sure whether it's the reason why Nimbus uses tink for EdDSA.

BTW: Java 15+ offers native support for EdDSA.

ziqin:
In the document, Spring Security should perhaps ask users to add tink dependency if they want to support EdDSA JWS.

I believe we could consider adding tink as a `testImplementation` or `testRuntimeOnly` dependency to Spring Security if that helps the test pass.

PR #19231, opened 2026-05-28
docs: document using spring security outside a servlet request

### LINKED ISSUE #11910: Document non Servlet code separately
The main reference documentation has a servlet section https://docs.spring.io/spring-security/reference/servlet/index.html, however if you are building code that does not need servlet access or cannot access servlet applications such as quartz, some unit tests, cli applications, event handlers (e.g. rabbitmq), etc, the documentation for this is hidden or mixed in in a way that is hard to find and understand. A new section that explicitly documents using spring security in these contexts would be especially helpful

cfcromn:
## Summary

Closes #11910

Adds `features/integrations/outside-servlet.adoc` as a small entry-point page for using Spring Security outside the servlet filter chain.

This PR also adds a corresponding navigation entry under Features > Integrations.

## Why

The reference docs include generic integrations that are not specific to Servlet or Reactive environments, but non-servlet use cases such as command-line runners, scheduled jobs, and message handlers can be hard to discover from a non-servlet starting point.

This keeps the initial change intentionally small and links to existing documentation for concurrency support and test support.

## Changes

- Adds `docs/modules/ROOT/pages/features/integrations/outside-servlet.adoc`
- Adds `Outside a Servlet Request` to `docs/modules/ROOT/nav.adoc`

## Test plan

- [x] `JAVA_HOME=$(/usr/libexec/java_home -v 25) ./gradlew :spring-security-docs:antora`
- [x] Confirmed `outside-servlet.html` is generated
- [x] Confirmed `Outside a Servlet Request` appears in the rendered navigation
- [x] Confirmed xref links resolve
- [x] Confirmed no AsciiDoc syntax errors

### DIFF
--- a/docs/modules/ROOT/nav.adoc
+++ b/docs/modules/ROOT/nav.adoc
@@ -26,6 +26,7 @@
 *** xref:features/integrations/cryptography.adoc[Cryptography]
 *** xref:features/integrations/data.adoc[Spring Data]
 *** xref:features/integrations/concurrency.adoc[Java's Concurrency APIs]
+*** xref:features/integrations/outside-servlet.adoc[Outside a Servlet Request]
 *** xref:features/integrations/jackson.adoc[Jackson]
 *** xref:features/integrations/localization.adoc[Localization]
 * xref:modules.adoc[Project Modules]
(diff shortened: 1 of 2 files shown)


cfcromn:
Hi, just following up on this PR.

This adds a small documentation entry point for using Spring Security outside the servlet filter chain, as requested in #11910. I kept the scope intentionally small by adding one new page and a navigation entry, linking to existing concurrency and test support documentation instead of duplicating content.

Could you let me know if this direction looks acceptable, or if the documentation should be organized differently?

PR #19233, opened 2026-05-28
AuthenticatorTransport#values should include SMART_CARD

### LINKED ISSUE #19232: AuthenticatorTransport#values should return all values
`AuthenticatorTransport#values()` omits `SMART_CARD`, even though it is a defined constant and is handled correctly in `valueOf(String)`.

Before:
```java
public static AuthenticatorTransport[] values() {
    return new AuthenticatorTransport[] { USB, NFC, BLE, HYBRID, INTERNAL };
}
```

After:
```java
public static AuthenticatorTransport[] values() {
    return new AuthenticatorTransport[] { USB, NFC, BLE, SMART_CARD, HYBRID, INTERNAL };
}
```

therepanic:
We do not specify the SMART_CARD constant in
`AuthenticatorTransport#values`, even though it is processed in `AuthenticatorTransport#valueOf`.

Closes: gh-19232

### DIFF
--- a/webauthn/src/main/java/org/springframework/security/web/webauthn/api/AuthenticatorTransport.java
+++ b/webauthn/src/main/java/org/springframework/security/web/webauthn/api/AuthenticatorTransport.java
@@ -26,6 +26,7 @@ import java.io.Serializable;
  * order to obtain an assertion for a specific credential.
  *
  * @author Rob Winch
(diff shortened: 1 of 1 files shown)


jzheaux:
Thanks, @therepanic! This will merge into `7.0.x` when the build completes.

PR #19243, opened 2026-05-30
Add allowIfSubType for URL and Instant in OAuth2ClientJacksonModule 

### LINKED ISSUE #19241: OAuth2ClientJacksonModule is missing allowIfSubType for URL and Instant
**Describe the bug**
setup a spring mvc app with oauth2Login and redis session. When trying to login with any OIDC / OAuth 2 provider, after redirecting back to the app, it throws an error.

org.springframework.data.redis.serializer.SerializationException: Could not read JSON: Could not resolve type id 'java.net.URL' as a subtype of `java.lang.Object`: Configured `PolymorphicTypeValidator` (of type `tools.jackson.databind.jsontype.BasicPolymorphicTypeValidator`) denied resolution
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); byte offset: #1982] (through reference chain: org.springframework.security.core.context.SecurityContextImpl["authentication"]->org.springframework.security.oauth2.client.authentication.OAuth2AuthenticationToken["principal"]->org.springframework.security.oauth2.core.oidc.user.DefaultOidcUser["authorities"]->java.util.HashSet[0]->org.springframework.security.oauth2.core.oidc.user.OidcUserAuthority["idToken"]->org.springframework.security.oauth2.core.oidc.OidcIdToken["claims"]->java.util.LinkedHashMap["iss"])
	at org.springframework.data.redis.serializer.JacksonJsonRedisSerializer.deserialize(JacksonJsonRedisSerializer.java:143)

org.springframework.data.redis.serializer.SerializationException: Could not read JSON: Could not resolve type id 'java.time.Instant' as a subtype of `java.lang.Object`: Configured `PolymorphicTypeValidator` (of type `tools.jackson.databind.jsontype.BasicPolymorphicTypeValidator`) denied resolution
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); byte offset: #2435] (through reference chain: org.springframework.security.core.context.SecurityContextImpl["authentication"]->org.springframework.security.oauth2.client.authentication.OAuth2AuthenticationToken["principal"]->org.springframework.security.oauth2.core.oidc.user.DefaultOidcUser["authorities"]->java.util.HashSet[0]->org.springframework.security.oauth2.core.oidc.user.OidcUserAuthority["idToken"]->org.springframework.security.oauth2.core.oidc.OidcIdToken["claims"]->java.util.LinkedHashMap["exp"])
	at org.springframework.data.redis.serializer.JacksonJsonRedisSerializer.deserialize(JacksonJsonRedisSerializer.java:143)

qwerty7878:
Fixes gh-19241

## Problem

Deserializing an `OAuth2AuthenticationToken` from a Redis session fails
with a `SerializationException`, because `OidcIdToken` stores `iss` as a
`java.net.URL` inside its claims map and the configured
`PolymorphicTypeValidator` denies it.

`java.net.URL` is present in the Jackson 2 allowlist
(`AllowlistTypeIdResolver.ALLOWLIST_CLASS_NAMES`) but missing from the
Jackson 3 modules.

## Fix

Added `allowIfSubType(URL.class)` to `CoreJacksonModule`.
`SecurityJacksonModules` passes a single shared builder to every
`SecurityJacksonModule`, so the entry applies to the OAuth2 modules as
well.

## Tests

- `deserializeWhenMapContainsUrlThenDeserializes` in
  `SecurityJacksonModulesTests` — a `Map` value of type `URL` round-trips
- `deserializeWhenClaimsContainUrlThenDeserializes` in
  `OAuth2AuthenticationTokenMixinTests` — regression test for the
  originally reported scenario

### DIFF
--- a/core/src/main/java/org/springframework/security/jackson/CoreJacksonModule.java
+++ b/core/src/main/java/org/springframework/security/jackson/CoreJacksonModule.java
@@ -16,6 +16,7 @@
 
 package org.springframework.security.jackson;
 
+import java.net.URL;
 import java.time.Duration;
 import java.time.Instant;
 
@@ -73,6 +74,7 @@ public class CoreJacksonModule extends SecurityJacksonModule {
 	public void configurePolymorphicTypeValidator(BasicPolymorphicTypeValidator.Builder builder) {
 		builder.allowIfSubType(Instant.class)
 			.allowIfSubType(Duration.class)
+			.allowIfSubType(URL.class)
 			.allowIfSubType(SimpleGrantedAuthority.class)
 			.allowIfSubType(FactorGrantedAuthority.class)
 			.allowIfSubType(UsernamePasswordAuthenticationToken.class)
--- a/core/src/test/java/org/springframework/security/jackson/SecurityJacksonModulesTests.java
+++ b/core/src/test/java/org/springframework/security/jackson/SecurityJacksonModulesTests.java
@@ -16,7 +16,11 @@
 
 package org.springframework.security.jackson;
 
+import java.net.URI;
+import java.net.URL;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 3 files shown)


filiphr:
I was hit by this problem as well. The Jackson 2 modules already handle this properly in 

https://github.com/spring-projects/spring-security/blob/5bd6b4a570765a1da963d6826a15b4e3fdd71463/core/src/main/java/org/springframework/security/jackson2/SecurityJackson2Modules.java#L245.

@qwerty7878 I think that the right fix would be to do this in the [`CoreJacksonModule`](https://github.com/spring-projects/spring-security/blob/5bd6b4a570765a1da963d6826a15b4e3fdd71463/core/src/main/java/org/springframework/security/jackson/CoreJacksonModule.java). The `java.time.Instant` is already there.

Maybe someone from the spring-security team can also look into this. Currently, there is a gap between the Jackson 2 and Jackson 3 support

qwerty7878:
Thanks @filiphr, that makes sense.

The failure happens on OidcIdToken#getClaims(), which is a Map<String, Object>, so a URL can appear in any map-valued payload rather than only in the OAuth2 modules. And since SecurityJacksonModules passes a single shared builder to every SecurityJacksonModule, adding it to CoreJacksonModule covers the OAuth2 path as well.

I've moved allowIfSubType(URL.class) there and dropped the Instant entry since it's already present. The OAuth2 test is kept as a regression test for the originally reported scenario, and it passes with the fix only in core.

On the Jackson 2 / Jackson 3 gap: I compared AllowlistTypeIdResolver.ALLOWLIST_CLASS_NAMES against CoreJacksonModule entry by entry. java.net.URL is the only one present in the Jackson 2 allowlist and missing from the Jackson 3 one, so this change should close it.

DarkAtra:
@jzheaux we're also hitting this issue. what is missing to get this PR merged?

Issue #19256, opened 2026-06-02
Improve OAuth2 Device Authorization consent handling for missing scopes


therepanic:
There are two related issues in the OAuth2 Device Authorization flow when no scopes are requested.

The first issue is that when a `device_authorization` request is made without a scope, everything works fine, but when both `authorities` and `scope` are empty, an error is thrown that does not comply with [Section 3.3 of RFC-6749](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3). The second issue is that the error itself is reported as a `client_id` error rather than a `scope` error. Therefore, we need to resolve these two issues.

- [ ] https://github.com/spring-projects/spring-security/pull/19257
- [ ] https://github.com/spring-projects/spring-security/pull/19258

therepanic:
Inspiration: https://github.com/spring-projects/spring-security/issues/19238

PR #19257, opened 2026-06-02
Fix device authorization flow without requested scopes

### LINKED ISSUE #19238: OAuth2 Device Code Flow requires to be used with a requested scope
I use the OAuth2 Authorization Server (V. 4.0.6) with an OAuth Client which supports the AuthorizationGrantType "urn:ietf:params:oauth:grant-type:device_code". When I request a device authorization via a POST request to "/oauth2/device_authorization" without any scope and later giving the consent on the "DefaultConsentPage" to the device an error is thrown.

I think the problem is located in 'OAuth2DeviceAuthorizationConsentAuthenticationProvider' line 182 following. In case the list of authorities is empty an error will be thrown. The message of that error is also some kind of misleading because the property "client_id" is mentioned missing. Should be "scope" instead. 
 
**To Reproduce**
- Start the Application from the example
- Execute a device authorization request without any scope
- Open Browser and browse to URL given in the property "verification_uri_complete" of the previous authorization request.
- Login (test/test)
- Click "Submit Consent"

**Expected behavior**
Because the scope parameter in the initial device authorization request is optional I expect that i can give the consent if there was no requested scope.

**Sample**
Example Project: https://github.com/cwiejack/spring-authorizationserver-device-code-bug
see exampleRequests.http

therepanic:
The main issue is inconsistent behavior. A `device_authorization` request without a `scope` is always accepted, but then, during the `consent` phase, if `authorities` is empty, we always get an `access_denied` error. However, would not it make sense to continue processing when both `scope` and `authorities` are empty?

The `scope` parameter itself is optional according to [RFC-8628](https://datatracker.ietf.org/doc/html/rfc8628#section-3.1). There is also a note attached to it that refers to [Section 3.3 in RFC-6749](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3). It states that if `scope` is empty, then we must either handle it with a default value or throw an error.

And here’s the interesting part. I propose two solutions to the problem; I’ve implemented the first one in the current commit, and in the future we can pivot to the other solution. The first is to directly handle the case at the `consent` stage when `scope` and `authorities` are empty and allow the event to proceed. The second is to throw an error on a `device_authorization` request if `scope` is empty.

I think the first solution is better because it’s also easier to maintain. With the first solution, the test breaks, but that’s because it handles all cases and is abstract, so I refined it for the case where `authorities` is empty but `scope` is not empty. The second solution breaks the specific test case `OAuth2DeviceAuthorizationRequestAuthenticationProviderTests#authenticateWhenNoScopesRequestedThenReturnDeviceCodeAndUserCode`.

Closes: gh-19238
Ref: gh-19256

### DIFF
--- a/oauth2/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2DeviceAuthorizationConsentAuthenticationProvider.java
+++ b/oauth2/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2DeviceAuthorizationConsentAuthenticationProvider.java
@@ -50,6 +50,7 @@ import org.springframework.util.Assert;
  * used in the OAuth 2.0 Device Authorization Grant.
  *
  * @author Steve Riesenberg
+ * @author Andrey Litvitski
  * @since 7.0
  * @see OAuth2DeviceAuthorizationConsentAuthenticationToken
  * @see OAuth2AuthorizationConsent
@@ -183,7 +184,7 @@ public final class OAuth2DeviceAuthorizationConsentAuthenticationProvider implem
 		OAuth2Authorization.Token<OAuth2UserCode> userCodeToken = authorization.getToken(OAuth2UserCode.class);
 		Assert.notNull(userCodeToken, "userCode cannot be null");
 
-		if (authorities.isEmpty()) {
+		if (authorities.isEmpty() && !requestedScopes.isEmpty()) {
 			// Authorization consent denied (or revoked)
 			if (currentAuthorizationConsent != null) {
 				this.authorizationConsentService.remove(currentAuthorizationConsent);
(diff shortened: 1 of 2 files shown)


therepanic:
I renamed the PR to avoid being tied to the commit title in case we decide to change direction.

oauth2/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2DeviceAuthorizationConsentAuthenticationProvider.java
```diff
@@ -203,11 +204,13 @@ public Authentication authenticate(Authentication authentication) throws Authent
...
 		}
 
-		OAuth2AuthorizationConsent authorizationConsent = authorizationConsentBuilder.build();
-		if (currentAuthorizationConsent == null || !authorizationConsent.equals(currentAuthorizationConsent)) {
-			this.authorizationConsentService.save(authorizationConsent);
-			if (this.logger.isTraceEnabled()) {
-				this.logger.trace("Saved authorization consent");
+		if (!authorities.isEmpty()) {
+			OAuth2AuthorizationConsent authorizationConsent = authorizationConsentBuilder.build();
+			if (currentAuthorizationConsent == null || !authorizationConsent.equals(currentAuthorizationConsent)) {
```
therepanic, line 209: If left unprocessed, the following `OAuth2AuthorizationConsent.Builder#build` assertiong will be executed

```java
public OAuth2AuthorizationConsent build() {
    Assert.notEmpty(this.authorities, "authorities cannot be empty");
    return new OAuth2AuthorizationConsent(this.registeredClientId, this.principalName, this.authorities);
}
```

PR #19262, opened 2026-06-03
Add SMART_CARD to AuthenticatorTransport values

### LINKED ISSUE #19232: AuthenticatorTransport#values should return all values

sathish256:
`AuthenticatorTransport#values()` omitted the `SMART_CARD` constant, even though it is a defined constant and is handled correctly by `valueOf(String)`. This adds `SMART_CARD` to the array returned by `values()`, ordered consistently with `valueOf(String)`.

A regression test (`AuthenticatorTransportTests`) is added, mirroring the existing `COSEAlgorithmIdentifierTests`, that reflectively verifies `values()` returns all declared constants, plus an explicit check for `SMART_CARD`.

Closes gh-19232

### DIFF
--- a/webauthn/src/main/java/org/springframework/security/web/webauthn/api/AuthenticatorTransport.java
+++ b/webauthn/src/main/java/org/springframework/security/web/webauthn/api/AuthenticatorTransport.java
@@ -118,7 +118,7 @@ public final class AuthenticatorTransport implements Serializable {
 	}
 
 	public static AuthenticatorTransport[] values() {
-		return new AuthenticatorTransport[] { USB, NFC, BLE, HYBRID, INTERNAL };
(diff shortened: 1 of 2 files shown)


sathish256:
Closing as a duplicate — I see #19233 already addresses this. Apologies for the noise, I missed the existing PR. Deferring to #19233.

PR #19307, opened 2026-06-09
Add missing nullability annotations in `AbstractConfiguredSecurityBuilder`

### LINKED ISSUE #19304: `AbstractConfiguredSecurityBuilder.getSharedObject` should be marked `@Nullable`
**Describe the bug**
`AbstractConfiguredSecurityBuilder` is in a `@NullMarked` package, is documented to possible return `null`, but its return type is not `@Nullable`.

This causes erroneous results from static analysis tools.

**Expected behavior**
Anything that can be null should be marked `@Nullable`, since the default was set to be `@NonNull` in Spring Security 7.

Previous instances of this bug: #18275, #18259

therepanic:
Since the package containing `AbstractConfiguredSecurityBuilder` is annotated with `@NullMarked`, we should annotate the methods that return null with `@Nullable`.

Closes: gh-19304

### DIFF
--- a/config/src/main/kotlin/org/springframework/security/config/annotation/web/HttpSecurityDsl.kt
+++ b/config/src/main/kotlin/org/springframework/security/config/annotation/web/HttpSecurityDsl.kt
@@ -73,6 +73,7 @@ operator fun HttpSecurity.invoke(httpConfiguration: HttpSecurityDsl.() -> Unit) 
  * in order to configure [HttpSecurity] using idiomatic Kotlin code.
  *
  * @author Eleftheria Stein
+ * @author Andrey Litvitski
  * @since 5.3
  * @param http the [HttpSecurity] which all configurations will be applied to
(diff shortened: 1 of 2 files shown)


config/src/main/kotlin/org/springframework/security/config/annotation/web/HttpSecurityDsl.kt
```diff
@@ -82,7 +83,7 @@ operator fun HttpSecurity.invoke(httpConfiguration: HttpSecurityDsl.() -> Unit)
 class HttpSecurityDsl(private val http: HttpSecurity, private val init: HttpSecurityDsl.() -> Unit) {
 
     var authenticationManager: AuthenticationManager? = null
-    val context: ApplicationContext = http.getSharedObject(ApplicationContext::class.java)
+    val context: ApplicationContext = http.getSharedObject(ApplicationContext::class.java)!!
```
therepanic, line 87: Without this, the build fails.

PR #19320, opened 2026-06-11
Add hook for adding authentication filters

### LINKED ISSUE #19319: Ability to use AbstractAuthenticationFilterConfigurer for custom filters
**Expected Behavior**

A new custom auth processor can be added by implementing `AbstractAuthenticationFilterConfigurer` and `AbstractAuthenticationProcessingFilter`.

**Current Behavior**

Trying to do this fails, as `AbstractAuthenticationFilterConfigurer` calls `http.addFilter()` at the end of its quite long `configure()` method, which fails on any non-builtin filter.

**Context**

Specifically I was adding a system to force a password change if credentials are expired.
After the `CredentialsExpiredException` is handled, there is another POST request with the new password, which should go through all the same "authentication success" steps when it completes. The way to do this appears to be to implement an `AbstractAuthenticationProcessingFilter`, but to ensure it's configured correctly you need to either copy-paste a lot of `AbstractAuthenticationFilterConfigurer`, or do this:

```java
@Override
public void configure(HttpSecurity http) {
    try {
        super.configure(http);
    } catch (IllegalArgumentException ex) {
        http.addFilterAfter(getAuthenticationFilter(), UsernamePasswordAuthenticationFilter.class);
    }
}
```

Some way to define the order of a custom filter (e.g. another `protected` method in `AbstractAuthenticationFilterConfigurer`) would fix this.

therepanic:
This change will allow us to directly add authentication filters without resorting to various hacks by directly overriding the protected `addAuthenticationFilter` method.

Closes: gh-19319

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/web/configurers/AbstractAuthenticationFilterConfigurer.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/web/configurers/AbstractAuthenticationFilterConfigurer.java
@@ -56,6 +56,7 @@ import org.springframework.web.accept.HeaderContentNegotiationStrategy;
  * @param <F> refers to the {@link AbstractAuthenticationProcessingFilter} that is being
  * built
  * @author Rob Winch
+ * @author Andrey Litvitski
  * @since 3.2
  * @see FormLoginConfigurer
  */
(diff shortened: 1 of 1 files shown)


config/src/main/java/org/springframework/security/config/annotation/web/configurers/AbstractAuthenticationFilterConfigurer.java
```diff
@@ -300,6 +301,15 @@ public void configure(B http) {
...
+		addAuthenticationFilter(http, filter);
+	}
+
+	/**
+	 * Adds the authentication filter
+	 * @param http the {@link HttpSecurityBuilder}
+	 * @param filter the authentication filter
+	 */
+	protected void addAuthenticationFilter(B http, F filter) {
 		http.addFilter(filter);
```
therepanic, line 313: Although this name makes perfect sense, we could consider another option if anyone has better ideas.

Issue #19324, opened 2026-06-12
ServletOAuth2AuthorizedClientExchangeFilterFunction doesn't obtain a token if no servlet request


petergphillips:
**Describe the bug**
After the upgrade to Spring Security 7.1.0 our web clients no longer attempt to retrieve an oauth2 token when used in batch jobs and when fired from event listeners.

If there is a servlet request (so triggered from an incoming request) then the web clients will retrieve an oauth2 token as expected.

**To Reproduce**
Create a web client with a `ServletOAuth2AuthorizedClientExchangeFilterFunction` that is created from a `AuthorizedClientServiceOAuth2AuthorizedClientManager`. This manager (according to the [docs](https://github.com/spring-projects/spring-security/blob/baad23caab88cb7912cb718ea4a422804801a96e/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/AuthorizedClientServiceOAuth2AuthorizedClientManager.java#L38-L40)) should be capable of operating in a scheduled / background thread. So:
```java
new ServletOAuth2AuthorizedClientExchangeFilterFunction(
		new AuthorizedClientServiceOAuth2AuthorizedClientManager(this.clientRegistrationRepository,
				new InMemoryOAuth2AuthorizedClientService(this.clientRegistrationRepository)
		)
);
```
Call the web client without any request / response attributes in thread local scope.

**Expected behavior**
A token is retrieved from the oauth2 provider for the web client request.

**Actual behavior**
No token is retrieved at all. Looking at the recent changes to the `ServletOAuth2AuthorizedClientExchangeFilterFunction` there was a [null check](https://github.com/spring-projects/spring-security/blob/2ada3f00fa8c5927b8dc750efc6966f23607137d/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java#L514-L516) that was added in https://github.com/spring-projects/spring-security/commit/baad23caab88cb7912cb718ea4a422804801a96e:
```java
		if (servletRequest == null || servletResponse == null) {
			return Mono.empty();
		}
```
Our service then reaches that code and then returns, so no token is then retrieved. The web client then fails as no authentication header is set in the call.

mike-thecodebuilders:
I'm facing the same problem after upgrading to the latest Spring (Boot) version. 

[...]

PR #19326, opened 2026-06-12
Tests to demonstrate gh-19324 when the servlet request is null

petergphillips:
Tests to demonstrate https://github.com/spring-projects/spring-security/issues/19324.

Both fail as the recent changes to the filter function cause the authentication to be skipped.


jgrandja:
Closing in favour of gh-19421

PR #19330, opened 2026-06-12
#19329 fix ipv4 vs ipv6 in ip inet address matcher

### LINKED ISSUE #19329: `ArrayIndexOutOfBoundsException` using `IpAddressMatcher` 7.1
**Describe the bug**
We noticed that since release 7.1.0 (cc6a005a to be precise) an `IpAddressMatcher` for an IPv6 address will crash with `ArrayIndexOutOfBoundsException` when trying to match an IPv4 address being equal to the first 4 bytes of the IPv6 address.

It is easy to fix this and return to the behavior before 7.1.0.

**To Reproduce**
very simple:
```java
var matcher= new IpAddressMatcher("::1/128");
var result = machter.matches("0.0.0.0");
// java.lang.ArrayIndexOutOfBoundsException: Index 4 out of bounds for length 4
```

**Expected behavior**
No exception. Result should be `false`.

**Sample**

See above

I will submit a PR with regression test.

winfriedgerlach:
Closes #19329

### DIFF
--- a/core/src/main/java/org/springframework/security/util/matcher/IpInetAddressMatcher.java
+++ b/core/src/main/java/org/springframework/security/util/matcher/IpInetAddressMatcher.java
@@ -17,7 +17,6 @@
 package org.springframework.security.util.matcher;
 
 import java.net.InetAddress;
-import java.net.UnknownHostException;
 import java.util.Objects;
 
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/springframework/security/util/matcher/IpInetAddressMatcher.java
```diff
@@ -70,25 +69,14 @@ final class IpInetAddressMatcher implements InetAddressMatcher {
 			.format("IP address %s is too short for bitmask of length %d", requiredAddress, this.nMaskBits));
 	}
 
-	private static InetAddress parse(String address) {
```
winfriedgerlach, line 73: this was unused residue from the earlier refactoring

core/src/main/java/org/springframework/security/util/matcher/IpInetAddressMatcher.java
```diff
@@ -70,25 +69,14 @@ final class IpInetAddressMatcher implements InetAddressMatcher {
...
-			throw new IllegalArgumentException(String.format("Failed to parse address '%s'", address), ex);
-		}
-	}
-
 	@Override
 	public boolean matches(@Nullable InetAddress toCheck) {
 		if (toCheck == null) {
 			return false;
 		}
+		if (!this.requiredAddress.getClass().equals(toCheck.getClass())) {
```
winfriedgerlach, line 77: this check was present before 7.1.0, I am just re-introducing it

PR #19331, opened 2026-06-13
Avoid reflective final field mutation in BasicLookupStrategy

### LINKED ISSUE #19127: Final field mutation reported on Java 26
When running on java 26 the following warning is logged

WARNING: Final field aces in class org.springframework.security.acls.domain.AclImpl has been mutated reflectively by class org.springframework.security.acls.jdbc.BasicLookupStrategy in unnamed module @1d8bd0de (file:~/.m2/repository/org/springframework/security/spring-security-acl/7.0.5/spring-security-acl-7.0.5.jar)
WARNING: Use --enable-final-field-mutation=ALL-UNNAMED to avoid a warning
WARNING: Mutating final fields will be blocked in a future release unless final field mutation is enabled

seonwooj0810:
Closes gh-19127

## Background

On Java 26 the JVM warns whenever ACLs are loaded through `BasicLookupStrategy`:

```
WARNING: Final field aces in class org.springframework.security.acls.domain.AclImpl has been mutated reflectively
by class org.springframework.security.acls.jdbc.BasicLookupStrategy in unnamed module ...
WARNING: Use --enable-final-field-mutation=ALL-UNNAMED to avoid a warning
WARNING: Mutating final fields will be blocked in a future release unless final field mutation is enabled
```

`BasicLookupStrategy.convert(...)` is responsible for re-parenting a loaded `AclImpl` onto a freshly constructed one: stub parent references are resolved and the back-reference on every `AccessControlEntryImpl` is rewritten to point at the new ACL (per SEC-951). The previous implementation did this by reflectively rewriting two `final` fields:

- `AclImpl.aces` was replaced through `Field.set(...)` (`setAces`).
- `AccessControlEntryImpl.acl` was replaced through `Field.set(...)` (`setAclOnAce`).

Both writes trigger the Java 26 warning today and will be rejected outright on the JVMs that flip the default. Reading the `aces` field reflectively is not affected — only mutation is.

## Change

Neither final field actually needs to be replaced:

- The new `AclImpl` is created with an empty, mutable `aces` list (initialised inline as `new ArrayList<>()`). Adding to that existing list in place achieves the same end state without touching the field reference. `addAces` (replacing `setAces`) reuses the existing read helper and calls `addAll` on the list.
- `AccessControlEntryImpl` is documented as immutable and exposes a public constructor that takes every field. Re-creating each ACE with the converted `Acl` lets us substitute the `StubAclParent`-bearing ACE for an equivalent one without mutating any final field. `setAclOnAce` and the cached `Field` for `AccessControlEntryImpl.acl` go away entirely.

The reflective *read* of `AclImpl.aces` is preserved — it is necessary to populate the result list and is unaffected by the upcoming JVM restriction. The internal `ProcessResultSet.mapRow` path is unchanged: it already uses the same read-then-`add` idiom to populate the input ACLs.

## Tests

Added `readAclsByIdWhenChildLoadedThenAcesPointToConvertedAcl` to `AbstractBasicLookupStrategyTests`, which asserts the SEC-951 invariant directly: after `readAclsById(...)`, every ACE returned by `acl.getEntries()` reports the converted `Acl` (not a `StubAclParent`) via `getAcl()`. Both `BasicLookupStrategyTests` and `BasicLookupStrategyWithAclClassTypeTests` extend the abstract class, so the new test runs against both class-id strategies.

`./gradlew :spring-security-acl:check` passes (test, checkFormat, checkstyle, license, jacoco).

## Verification done

1. **No in-flight PR**: `gh pr list --repo spring-projects/spring-security --search "AclImpl OR \"final field\" OR \"Java 26\""` returns nothing related, and `gh pr list --search "19127"` is empty.
2. **No claim in the issue thread**: gh-19127 has zero comments at the time of writing.
3. **Code-focused fix**: touches `acl/src/main/java/.../BasicLookupStrategy.java` and the existing abstract test class.
4. **Pattern still present on main**: confirmed `fieldAces.set(...)` and `fieldAcl.set(...)` still live in `BasicLookupStrategy` at HEAD (`5594fc018f`).
5. **Reproducer matches**: the warning text in gh-19127 names `BasicLookupStrategy` mutating `AclImpl.aces`, which is the call this change removes.

### DIFF
--- a/acl/src/main/java/org/springframework/security/acls/jdbc/BasicLookupStrategy.java
+++ b/acl/src/main/java/org/springframework/security/acls/jdbc/BasicLookupStrategy.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2004, 2005, 2006, 2017 Acegi Technology Pty Limited
+ * Copyright 2004, 2005, 2006, 2017, 2026 Acegi Technology Pty Limited
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -127,8 +127,6 @@ public class BasicLookupStrategy implements LookupStrategy {
 
 	private final Field fieldAces = FieldUtils.getField(AclImpl.class, "aces");
 
-	private final Field fieldAcl = FieldUtils.getField(AccessControlEntryImpl.class, "acl");
-
 	// SQL Customization fields
 	private String selectClause = DEFAULT_SELECT_CLAUSE;
 
@@ -171,7 +169,6 @@ public class BasicLookupStrategy implements LookupStrategy {
 		this.objectIdentityGenerator = new ObjectIdentityRetrievalStrategyImpl();
 		this.aclClassIdUtils = new AclClassIdUtils();
 		this.fieldAces.setAccessible(true);
-		this.fieldAcl.setAccessible(true);
 	}
 
 	private String computeRepeatingSql(String repeatingSql, int requiredRepetitions) {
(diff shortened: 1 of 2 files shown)


seonwooj0810:

On Java 26 every ACL load through `BasicLookupStrategy` emits the "Final field aces ... has been mutated reflectively" warning, and that mutation is slated to be blocked outright in a future release (gh-19127). The PR drops the two reflective final-field writes: instead of replacing the `aces` list reference on `AclImpl`, it populates the existing (mutable) list in place, and it no longer sets the `acl` back-reference on each `AccessControlEntryImpl` reflectively.

Is mutating the existing list the direction you want, or would you rather `AclImpl` grow explicit API for re-parenting? Still merges cleanly, checks green.

seonwooj0810:
Following up: on Java 26 this warning is emitted on every ACL load, and the JDK states that final-field mutation will be blocked by default in a future release, at which point `BasicLookupStrategy` would fail outright (gh-19127). Happy to rework it if you would rather take a different direction.

PR #19337, opened 2026-06-14
Find mixed-case usernames in InMemoryUserDetailsManager#changePassword

### LINKED ISSUE #19336: InMemoryUserDetailsManager.changePassword fails for non-lowercase usernames
**Describe the bug**

`InMemoryUserDetailsManager.changePassword(...)` cannot change the password of a user whose username contains uppercase letters.

The manager keys its internal user map on the lower-cased username in every other method (`createUser`, `updateUser`, `deleteUser`, `userExists`, `loadUserByUsername`, `updatePassword`), but `changePassword(...)` looks the current user up with the raw `Authentication.getName()`.

**To Reproduce**

```java
InMemoryUserDetailsManager manager = new InMemoryUserDetailsManager(
        User.withUsername("User").password("{noop}password").roles("USER").build());

// authenticate as "User" (e.g. place a UsernamePasswordAuthenticationToken("User", ...) in the SecurityContext)

manager.changePassword("password", "newPassword");
// throws IllegalStateException: Current user doesn't exist in database.
```

The user is stored under the key `"user"`, but `changePassword` calls `users.get("User")`, which returns `null`.

**Expected behavior**

The password is changed, consistent with every other method on the manager, which treat the username case-insensitively.

I will submit a PR with a regression test.

junhyeong9812:
## Overview

`InMemoryUserDetailsManager.changePassword(...)` fails for users whose username contains uppercase letters.

## Problem

The manager keys its `Map<String, MutableUserDetails> users` on the lower-cased username in `createUser`, `updateUser`, `deleteUser`, `userExists`, `loadUserByUsername`, and `updatePassword`. Only `changePassword(...)` looked the current user up with the raw name:

```java
String username = currentUser.getName();
...
MutableUserDetails user = this.users.get(username); // not lower-cased
Assert.state(user != null, "Current user doesn't exist in database.");
```

So a user created as e.g. `User.withUsername("User")...` (stored under the key `"user"`) cannot change its password: `users.get("User")` is `null` and `IllegalStateException("Current user doesn't exist in database.")` is thrown, even though authentication and every other operation on that user succeed.

## Fix

Lower-case the lookup key, matching the rest of the class:

```java
MutableUserDetails user = this.users.get(username.toLowerCase(Locale.ROOT));
```

A regression test creates the manager with an uppercase username, authenticates as that user, and asserts `changePassword(...)` updates the password (observed through `loadUserByUsername`).

Closes gh-19336

### DIFF
--- a/core/src/main/java/org/springframework/security/provisioning/InMemoryUserDetailsManager.java
+++ b/core/src/main/java/org/springframework/security/provisioning/InMemoryUserDetailsManager.java
@@ -150,7 +150,7 @@ public class InMemoryUserDetailsManager implements UserDetailsManager, UserDetai
 		else {
 			this.logger.debug("No authentication manager set. Password won't be re-checked.");
 		}
-		MutableUserDetails user = this.users.get(username);
+		MutableUserDetails user = this.users.get(username.toLowerCase(Locale.ROOT));
 		Assert.state(user != null, "Current user doesn't exist in database.");
 		user.setPassword(newPassword);
 	}
--- a/core/src/main/java/org/springframework/security/provisioning/InMemoryUserDetailsManager.java
+++ b/core/src/main/java/org/springframework/security/provisioning/InMemoryUserDetailsManager.java
@@ -149,7 +149,7 @@ public class InMemoryUserDetailsManager implements UserDetailsManager, UserDetai
 		else {
 			this.logger.debug("No authentication manager set. Password won't be re-checked.");
 		}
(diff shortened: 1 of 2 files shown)


jzheaux:
Thanks, @junhyeong9812! This is now merged into `7.0.x` and will go out in the next release.

junhyeong9812:
Thanks for merging this, @jzheaux — and thank you for maintaining such a great framework. Happy to contribute back.

PR #19369, opened 2026-06-23
Fix WebAuthn unknown transport deserialization

### LINKED ISSUE #19366: NullPointerException during WebAuthn registration when authenticator transport is unknown (e.g. "cable")
## Describe the bug

When invoking the WebAuthn registration endpoint (`/webauthn/register`), Spring Security throws a `NullPointerException` if the registration payload contains an authenticator transport value that is not recognized by Spring Security.

For example, when the transport list contains `"cable"` (Cloud-assisted Bluetooth Low Energy), the request processing fails with:

```text
java.lang.NullPointerException: Cannot invoke
"org.springframework.security.web.webauthn.api.AuthenticatorTransport.getValue()"
because "transport" is null
    at org.springframework.security.web.webauthn.management.Webauthn4JRelyingPartyOperations.convertTransportsToString(Webauthn4JRelyingPartyOperations.java:309)
```

**To Reproduce**
1. Configure a Spring Security application with WebAuthn enabled.
2. Call the WebAuthn registration endpoint (/webauthn/register).
3. Include an authenticator transport value that is not mapped to Spring Security's AuthenticatorTransport enum, for example:

```
{
...
"transports": ["cable"]
...
}
```
Observe that request processing fails with a NullPointerException.

**Expected behavior**
The framework should not throw a NullPointerException when an unknown transport is received.
Possible expected behaviors:

- Reject the request with a meaningful validation exception.
- Ignore unsupported transports.
- Preserve unknown transports as strings if supported by the implementation.

In any case, the framework should fail gracefully instead of throwing an NPE.

## Analysis

The issue appears to originate in:

```java
org.springframework.security.web.webauthn.jackson.AuthenticatorTransportDeserializer
```

When an unknown authenticator transport value is encountered (for example `"cable"`), the deserializer's `deserialize()` method returns `null`.

The resulting collection is later processed by:

```java
org.springframework.security.web.webauthn.management.Webauthn4JRelyingPartyOperations
```

Specifically, in:

```java
convertTransportsToString(...)
```

which assumes all transport values are non-null and invokes:

```java
transport.getValue()
```

As a result, the following exception is thrown:

```text
java.lang.NullPointerException: Cannot invoke
"org.springframework.security.web.webauthn.api.AuthenticatorTransport.getValue()"
because "transport" is null
```

From my analysis, the root cause is that unsupported transport values are silently converted to `null` during deserialization and are subsequently dereferenced without validation.

Spring Security should either:

- Reject unknown transport values with a meaningful exception during deserialization, or
- Filter/ignore unsupported values before processing, or
- Preserve unknown values in a forward-compatible manner.

Any of these approaches would prevent the `NullPointerException` and provide a more predictable failure mode.

xfocus3:
Fixes gh-19366

`AuthenticatorTransport` already preserves unknown transport names, but the Jackson deserializers returned `null` for transport strings outside the built-in constants. Registration payloads that include values such as `"cable"` could then fail later when the registration flow reads the transport value.

This updates both Jackson 3 and Jackson 2 deserializers to delegate to `AuthenticatorTransport.valueOf(...)`, preserving unknown transport values while still returning the existing constants for known values.

Tests:
- RED: `./gradlew :spring-security-webauthn:test --tests org.springframework.security.web.webauthn.jackson.JacksonTests.readAuthenticatorTransportWhenUnknownThenPreservesValue --tests org.springframework.security.web.webauthn.jackson.Jackson2Tests.readAuthenticatorTransportWhenUnknownThenPreservesValue --no-daemon`
- GREEN: `./gradlew :spring-security-webauthn:test --tests org.springframework.security.web.webauthn.jackson.JacksonTests.readAuthenticatorTransportWhenUnknownThenPreservesValue --tests org.springframework.security.web.webauthn.jackson.Jackson2Tests.readAuthenticatorTransportWhenUnknownThenPreservesValue --no-daemon`
- `./gradlew :spring-security-webauthn:test --tests org.springframework.security.web.webauthn.jackson.JacksonTests --tests org.springframework.security.web.webauthn.jackson.Jackson2Tests --no-daemon`
- `./gradlew :spring-security-webauthn:test --no-daemon`
- `git diff --check HEAD^ HEAD`

### DIFF
--- a/webauthn/src/main/java/org/springframework/security/web/webauthn/jackson/AuthenticatorTransportDeserializer.java
+++ b/webauthn/src/main/java/org/springframework/security/web/webauthn/jackson/AuthenticatorTransportDeserializer.java
@@ -41,12 +41,10 @@ class AuthenticatorTransportDeserializer extends StdDeserializer<AuthenticatorTr
 	public @Nullable AuthenticatorTransport deserialize(JsonParser parser, DeserializationContext ctxt)
 			throws JacksonException {
 		String transportValue = parser.readValueAs(String.class);
-		for (AuthenticatorTransport transport : AuthenticatorTransport.values()) {
-			if (transport.getValue().equals(transportValue)) {
-				return transport;
-			}
+		if (transportValue == null) {
+			return null;
 		}
-		return null;
+		return AuthenticatorTransport.valueOf(transportValue);
 	}
 
 }
--- a/webauthn/src/main/java/org/springframework/security/web/webauthn/jackson/AuthenticatorTransportJackson2Deserializer.java
+++ b/webauthn/src/main/java/org/springframework/security/web/webauthn/jackson/AuthenticatorTransportJackson2Deserializer.java
@@ -47,12 +47,10 @@ class AuthenticatorTransportJackson2Deserializer extends StdDeserializer<Authent
 	public @Nullable AuthenticatorTransport deserialize(JsonParser parser, DeserializationContext ctxt)
 			throws IOException, JacksonException {
 		String transportValue = parser.readValueAs(String.class);
-		for (AuthenticatorTransport transport : AuthenticatorTransport.values()) {
-			if (transport.getValue().equals(transportValue)) {
-				return transport;
-			}
+		if (transportValue == null) {
+			return null;
 		}
-		return null;
+		return AuthenticatorTransport.valueOf(transportValue);
 	}
 
 }
--- a/webauthn/src/test/java/org/springframework/security/web/webauthn/jackson/Jackson2Tests.java
+++ b/webauthn/src/test/java/org/springframework/security/web/webauthn/jackson/Jackson2Tests.java
@@ -63,6 +63,14 @@ class Jackson2Tests {
 	}
 
 	@Test
+	void readAuthenticatorTransportWhenUnknownThenPreservesValue() throws Exception {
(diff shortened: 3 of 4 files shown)


webauthn/src/main/java/org/springframework/security/web/webauthn/jackson/AuthenticatorTransportDeserializer.java
```diff
@@ -41,12 +41,10 @@ class AuthenticatorTransportDeserializer extends StdDeserializer<AuthenticatorTr
 	public @Nullable AuthenticatorTransport deserialize(JsonParser parser, DeserializationContext ctxt)
 			throws JacksonException {
 		String transportValue = parser.readValueAs(String.class);
-		for (AuthenticatorTransport transport : AuthenticatorTransport.values()) {
```
jzheaux: We don't want to make a change like this since that would allow a client to inject an uncontrolled value into the runtime. This is an issue in cases, for example, where an application is using `JdbcUserCredentialRepository` and this value is persisted to the database.

As far as I can tell, [the spec](https://www.w3.org/TR/webauthn-3/#enum-transport) allows transports to be [any string](https://webidl.spec.whatwg.org/#idl-DOMString) so I'm not sure that there is a way in this deserializer to validate an arbitrary value as safe to be added into the deserialized object.

PR #19372, opened 2026-06-23
Correct validation logic in CasAuthenticationToken

### LINKED ISSUE #19368: credentials validation in CasAuthenticationToken.Builder uses Assert.notNull on a boolean instead of Assert.isTrue
**Describe the bug**

In the `Builder`-based constructor of `CasAuthenticationToken` (added in 7.0), the credentials precondition uses `Assert.notNull` on a boolean expression instead of `Assert.isTrue`:

```java
protected CasAuthenticationToken(Builder<?> builder) {
    super(builder);
    Assert.isTrue(!"".equals(builder.principal), "principal cannot be null or empty");
    Assert.notNull(!"".equals(builder.credentials), "credentials cannot be null or empty");
    ...
}
```

`!"".equals(builder.credentials)` is a `boolean`, which is autoboxed to a non-null `Boolean` when passed to `Assert.notNull(Object, String)`. The assertion therefore always passes and the check is dead code, so building a token with empty (`""`) credentials is silently accepted. The legacy constructor and the adjacent `principal` check (one line above) both reject empty values.

**To Reproduce**

```java
CasAuthenticationToken token = new CasAuthenticationToken("key", "alice", "pass",
        AuthorityUtils.createAuthorityList("ROLE_USER"), userDetails, new AssertionImpl("test"));
token.toBuilder().credentials("").build(); // no exception
```

**Expected behavior**

Building with empty credentials should throw `IllegalArgumentException("credentials cannot be null or empty")`, consistent with the `principal` check on the preceding line and with the legacy constructor. The fix is replacing `Assert.notNull` with `Assert.isTrue` on that line.

ngocnhan-tran1996:
Closes gh-19368

### DIFF
--- a/cas/src/main/java/org/springframework/security/cas/authentication/CasAuthenticationToken.java
+++ b/cas/src/main/java/org/springframework/security/cas/authentication/CasAuthenticationToken.java
@@ -108,7 +108,7 @@ public class CasAuthenticationToken extends AbstractAuthenticationToken implemen
 	protected CasAuthenticationToken(Builder<?> builder) {
 		super(builder);
 		Assert.isTrue(!"".equals(builder.principal), "principal cannot be null or empty");
-		Assert.notNull(!"".equals(builder.credentials), "credentials cannot be null or empty");
+		Assert.isTrue(!"".equals(builder.credentials), "credentials cannot be null or empty");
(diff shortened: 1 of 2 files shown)


jzheaux:
Thanks, @ngocnhan-tran1996! This is now merged into `7.0.x`, `7.1.x`, and `main`.

PR #19373, opened 2026-06-23
fix: allow nullable uid in InetOrgPerson

### LINKED ISSUE #19370: InetOrgPerson uid should not be a mandatory field - may cause regression as of version 7.1.0
**Describe the bug**
1. CTOR `org.springframework.security.ldap.userdetails.InetOrgPerson.Essence.Essence(DirContextOperations)` with attribute "uid" null, throws `IllegalArgumentException` with "uid cannot be null".
2. `org.springframework.security.ldap.userdetails.InetOrgPerson.getUid()` while uid is null throws `NullPointerException` with "uid cannot be null".

The source code contradicts itself in these lines

```
	private @Nullable String uid;

	public String getUid() {
		return Objects.requireNonNull(this.uid, "uid cannot be null");
	}
```

Came with this commit:
https://github.com/spring-projects/spring-security/commit/c5632ccd838fcb2753a978918561081cff037510

**To Reproduce**
divert from https://github.com/spring-projects/spring-security/blob/ed7ae7969ed2452e6ab76c505bdc6afb9bb957ac/ldap/src/test/java/org/springframework/security/ldap/userdetails/InetOrgPersonTests.java#L120 with uid null

**Expected behavior**
"uid" should be nullable throughout (to my knowledge and according to LDAP Schema, it's not a mandatory field).
To my knowledge, this was also the behaviour before 7.1.0

ezhilnn:
Fixes #19370

## Problem
`uid` is marked `@Nullable` on the field but two places contradicted 
this:
- `getUid()` called `Objects.requireNonNull()`, throwing NPE if uid is null
- `Essence(DirContextOperations)` constructor called `Assert.notNull()`, 
  throwing IllegalArgumentException if uid is absent from LDAP context

According to the LDAP inetOrgPerson schema, `uid` is not a mandatory 
field. This behavior also regressed from pre-7.1.0 behavior (introduced 
in commit c5632cc).

## Changes
- `getUid()` now returns `@Nullable String` and returns the field directly
- Constructor no longer asserts uid is non-null; skips `setUid()` if absent
- `setUid()` now accepts `@Nullable String` and guards username mapping 
  with a null check

## Testing
- Added `testNullUidIsAllowed()` to verify null uid no longer throws

### DIFF
--- a/ldap/src/main/java/org/springframework/security/ldap/userdetails/InetOrgPerson.java
+++ b/ldap/src/main/java/org/springframework/security/ldap/userdetails/InetOrgPerson.java
@@ -77,8 +77,8 @@ public class InetOrgPerson extends Person {
 
 	private @Nullable String uid;
 
-	public String getUid() {
-		return Objects.requireNonNull(this.uid, "uid cannot be null");
+	public @Nullable String getUid() {
+		return this.uid;
 	}
 
 	public @Nullable String getMail() {
@@ -222,8 +222,9 @@ public class InetOrgPerson extends Person {
 			setStreet(ctx.getStringAttribute("street"));
 			setTitle(ctx.getStringAttribute("title"));
 			String uid = ctx.getStringAttribute("uid");
-			Assert.notNull(uid, "uid cannot be null");
-			setUid(uid);
+			if (uid != null) {
+				setUid(uid);
+			}
 		}
 
 		@Override
@@ -236,11 +237,10 @@ public class InetOrgPerson extends Person {
 			((InetOrgPerson) this.instance).mail = email;
 		}
 
-		public void setUid(String uid) {
+		public void setUid(@Nullable String uid) {
 			Assert.notNull(this.instance, "Essence can only be used to create a single instance");
 			((InetOrgPerson) this.instance).uid = uid;
-
-			if (this.username == null) {
+			if (this.username == null && uid != null) {
 				setUsername(uid);
 			}
 		}
--- a/ldap/src/test/java/org/springframework/security/ldap/userdetails/InetOrgPersonTests.java
+++ b/ldap/src/test/java/org/springframework/security/ldap/userdetails/InetOrgPersonTests.java
@@ -81,6 +81,14 @@ public class InetOrgPersonTests {
 		assertThat(p.getDisplayName()).isEqualTo("Ghengis McCann");
 		assertThat(p.getInitials()).isEqualTo("G");
 	}
+	@Test
+	public void testNullUidIsAllowed() {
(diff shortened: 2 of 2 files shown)


ezhilnn:
@spring-projects-issues  can someone approve this

ldap/src/main/java/org/springframework/security/ldap/userdetails/InetOrgPerson.java
romge, line 178: This can be removed like in the previous version when the `instance.getUsername()` is directly checked in `setUid()`

romge, line 251: Remove this, no override is necessary when no local username variable is used.

ldap/src/main/java/org/springframework/security/ldap/userdetails/InetOrgPerson.java
```diff
@@ -236,11 +237,10 @@ public void setMail(@Nullable String email) {
 			((InetOrgPerson) this.instance).mail = email;
 		}
 
-		public void setUid(String uid) {
+		public void setUid(@Nullable String uid) {
 			Assert.notNull(this.instance, "Essence can only be used to create a single instance");
 			((InetOrgPerson) this.instance).uid = uid;
-
-			if (this.username == null) {
+			if (this.username == null && uid != null) {
```
romge, line 243:
```suggestion
			if (this.instance.getUsername() == null) {
```
Directly check the instance.getUsername(). The check for uid != null is not necessary.

ldap/src/main/java/org/springframework/security/ldap/userdetails/InetOrgPerson.java
```diff
@@ -222,8 +222,9 @@ public Essence(DirContextOperations ctx) {
 			setStreet(ctx.getStringAttribute("street"));
 			setTitle(ctx.getStringAttribute("title"));
 			String uid = ctx.getStringAttribute("uid");
-			Assert.notNull(uid, "uid cannot be null");
-			setUid(uid);
+			if (uid != null) {
+				setUid(uid);
+			}
```
romge, line 227:
```suggestion
			setUid(ctx.getStringAttribute("uid"));
```
I would just restore the original code, also to be consistent with the other attributes:

romge (changes requested):
The commit c5632ccd838fcb2753a978918561081cff037510 broke the uid nullability and introduced an additional local username variable that seems to be unnecessary.
I am suggesting to restore these parts to the original code. I have added detailed comments for the changes.

PR #19379, opened 2026-06-25
Fix TokenType comparison logic

### LINKED ISSUE #19377: TokenType reference comparison is used in BearerTokenAuthentication
**Describe the bug**

During creating a BearerTokenAuthentication instance by passing an OAuth2AccessToken which was created by hand and TokenType param was created by its constructor instead of using pre-created static variables, got an error about `credentials must be a bearer token`. However the token type is bearer, but the class is another reference, so the comparison fails.

Used versions:

- Spring boot security 4.0.6
- Spring framework security 7.0.5

I also check the issue still present on the main branch.

**To Reproduce**

_most easiest way is to reproduce it from test (check sample)_

- create a `OAuth2AccessToken.TokenType` using  constructor with value `"Bearer"`
- create a valid `OAuth2AccessToken` with the previously created `OAuth2AccessToken.TokenType`
- create a valid `BearerTokenAuthentication` with the previously created `OAuth2AccessToken`

**Expected behavior**

the `BearerTokenAuthentication` successfully created

**Sample**
``` 
@Test
void bearerTokenAuthenticationBug() {
    OAuth2AccessToken.TokenType tokenType = new OAuth2AccessToken.TokenType("Bearer");
    assert tokenType.equals(OAuth2AccessToken.TokenType.BEARER);

    OAuth2AccessToken accessToken = new OAuth2AccessToken(tokenType, "token-value", Instant.now(), Instant.now().plusSeconds(3600));

    OAuth2AuthenticatedPrincipal principal = new DefaultOAuth2AuthenticatedPrincipal(Map.of("testAttributeKey", "testAttributeValue"), List.of(new SimpleGrantedAuthority("testAuthority")));

    IllegalArgumentException testAuthority = assertThrows(IllegalArgumentException.class, () -> {
        new BearerTokenAuthentication(principal, accessToken, List.of(new SimpleGrantedAuthority("testAuthority")));
    });

    assert testAuthority.getMessage().equals("credentials must be a bearer token");
}
```

ngocnhan-tran1996:
Closes gh-19377

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/BearerTokenAuthentication.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/authentication/BearerTokenAuthentication.java
@@ -121,7 +121,8 @@ public class BearerTokenAuthentication extends AbstractOAuth2TokenAuthentication
 		 */
 		@Override
 		public B token(OAuth2AccessToken token) {
-			Assert.isTrue(token.getTokenType() == OAuth2AccessToken.TokenType.BEARER, "token must be a bearer token");
+			Assert.isTrue(OAuth2AccessToken.TokenType.BEARER.equals(token.getTokenType()),
+					"token must be a bearer token");
 			super.credentials(token);
 			return super.token(token);
 		}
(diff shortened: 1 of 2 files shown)


DragonFSKY:
Thanks for working on this. One small gap: `BearerTokenAuthentication.Builder#token` still uses the same reference comparison against `OAuth2AccessToken.TokenType.BEARER`, so `authentication.toBuilder().token(new OAuth2AccessToken(new OAuth2AccessToken.TokenType("Bearer"), ...)).build()` would still fail even after the constructor path is fixed.

I think this PR should update that builder check to `OAuth2AccessToken.TokenType.BEARER.equals(token.getTokenType())` as well, with a small builder-path regression test.

DragonFSKY:
Thanks for updating it.

jzheaux:
Thanks, @ngocnhan-tran1996! This will merge into `7.0.x` when the build completes.

PR #19412, opened 2026-07-02
Allow custom AuthenticationConverter in WebAuthnAuthenticationFilter

### LINKED ISSUE #19411: Allow custom request context in WebAuthnAuthenticationFilter via AuthenticationConverter
### Expected Behavior

`WebAuthnAuthenticationFilter` should expose an `AuthenticationConverter` setter
so that applications can extract additional request context (e.g. custom JSON fields,
query parameters, or headers) and make them available during the authentication flow.

A default implementation preserves existing behavior. Custom implementations could
enrich the resulting `Authentication` token or store extra context as request
attributes, making them accessible to the `AuthenticationSuccessHandler`.

This is consistent with patterns already established in other Spring Security filters
(e.g. `UsernamePasswordAuthenticationFilter`, `BearerTokenAuthenticationFilter`).

### Current Behavior

`attemptAuthentication` hard-codes the deserialization target to
`PublicKeyCredential<AuthenticatorAssertionResponse>`:

```java
ResolvableType resolvableType = ResolvableType.forClassWithGenerics(
    PublicKeyCredential.class, AuthenticatorAssertionResponse.class);
publicKeyCredential = this.converter.read(resolvableType, httpRequest, null);
```

Any extra fields sent in the request body are silently dropped by Jackson.
The resulting `WebAuthnAuthenticationRequestToken` only carries a
`RelyingPartyAuthenticationRequest` (credential + options) with no room
for application-level context.

There is no hook to intercept or extend what is read from the request before
the token is built and passed to the `AuthenticationManager`.

### Context

In a stateless passkey flow (JWT-based, no server-side session), the frontend
needs to pass additional data alongside `POST /login/webauthn` — for example,
a device identifier or login-context token — so that the `AuthenticationSuccessHandler`
can include it in the issued JWT.

The only working workaround today is URL query parameters:

```
POST /login/webauthn?deviceId=abc123
```

These are readable from `HttpServletRequest` in the success handler, but query
parameters appear in server access logs, which is undesirable for identifiers
that should be treated as semi-sensitive.

A custom `AuthenticationConverter` setter on `WebAuthnAuthenticationFilter` would
allow structured data to be read from the body or headers and forwarded cleanly
through the authentication flow without workarounds.

therepanic:
This change allows us to delegate the `attemptAuthentication` logic to the `AuthenticationConverter` interface and enables the definition of custom instances. Additionally, we have added a default implementation, `WebAuthnAuthenticationConverter`, whose behavior is fully deterministic with respect to the previous implementation.

Closes: gh-19411

### DIFF
--- a/webauthn/src/main/java/org/springframework/security/web/webauthn/authentication/WebAuthnAuthenticationFilter.java
+++ b/webauthn/src/main/java/org/springframework/security/web/webauthn/authentication/WebAuthnAuthenticationFilter.java
@@ -38,20 +38,15 @@ import org.springframework.http.converter.HttpMessageNotReadableException;
 import org.springframework.http.converter.HttpMessageNotWritableException;
 import org.springframework.http.converter.SmartHttpMessageConverter;
 import org.springframework.http.converter.json.JacksonJsonHttpMessageConverter;
-import org.springframework.http.server.ServletServerHttpRequest;
-import org.springframework.security.authentication.BadCredentialsException;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
 import org.springframework.security.web.authentication.AbstractAuthenticationProcessingFilter;
+import org.springframework.security.web.authentication.AuthenticationConverter;
 import org.springframework.security.web.authentication.AuthenticationEntryPointFailureHandler;
 import org.springframework.security.web.authentication.HttpMessageConverterAuthenticationSuccessHandler;
 import org.springframework.security.web.authentication.HttpStatusEntryPoint;
 import org.springframework.security.web.context.HttpSessionSecurityContextRepository;
-import org.springframework.security.web.webauthn.api.AuthenticatorAssertionResponse;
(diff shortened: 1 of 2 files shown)


webauthn/src/main/java/org/springframework/security/web/webauthn/authentication/WebAuthnAuthenticationFilter.java
```diff
@@ -144,6 +131,21 @@ public void setConverter(GenericHttpMessageConverter<Object> converter) {
 	public void setConverter(SmartHttpMessageConverter<Object> converter) {
 		Assert.notNull(converter, "converter cannot be null");
 		this.converter = converter;
+		this.webAuthnAuthenticationConverter.setConverter(converter);
```
therepanic, line 134: Perhaps this behavior should be better documented?

therepanic:
Does anyone have any ideas on the best place in the documentation to put information about `WebAuthnAuthenticationConverter` - assuming such a place exists?

PR #19421, opened 2026-07-08
Fix to allow null servlet request/response in ServletOAuth2AuthorizedClientExchangeFilterFunction

### LINKED ISSUE #19324: ServletOAuth2AuthorizedClientExchangeFilterFunction doesn't obtain a token if no servlet request

petergphillips:
I've added tests to demonstrate the fix works too. Please see the github issue for full details.

Signed-off-by: Peter Phillips <5099053+petergphillips@users.noreply.github.com>

Closes gh-19324.

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/reactive/function/client/ServletOAuth2AuthorizedClientExchangeFilterFunction.java
@@ -511,9 +511,6 @@ public final class ServletOAuth2AuthorizedClientExchangeFilterFunction implement
 		}
 		HttpServletRequest servletRequest = getRequest(attrs);
 		HttpServletResponse servletResponse = getResponse(attrs);
-		if (servletRequest == null || servletResponse == null) {
-			return Mono.empty();
-		}
 		OAuth2AuthorizeRequest.Builder builder = OAuth2AuthorizeRequest.withClientRegistrationId(clientRegistrationId)
 			.principal(authentication);
 		builder.attributes((attributes) -> addToAttributes(attributes, servletRequest, servletResponse));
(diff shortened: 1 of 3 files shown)


jgrandja:
Thanks for the fix @petergphillips. This is now merged.

PR #19445, opened 2026-07-14
Fix `OAuth2PushedAuthorizationRequestUri` parsing

### LINKED ISSUE #19444: OAuth2PushedAuthorizationRequestUri generates URIs which cannot be parsed due to ___ delimiter in the state part
**Describe the bug**
spring-security-oauth2-authorization-server

A percentage of PAR (Pushed Authorization Request) requests fail randomly. 
[OAuth2PushedAuthorizationRequestUri](https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2PushedAuthorizationRequestUri.java) generates URIs in this format:  
```
urn:ietf:params:oauth:request_uri:<Base64URL_state>___<epoch_millis>  
Generator: Base64StringKeyGenerator(Base64.getUrlEncoder())  
- Base64URL alphabet: A-Z, a-z, 0-9, -, _ (64 characters)  
- Character '_' (underscore) IS in the output!  
- Delimiter '___' (triple underscore) also uses '_'  
=> CONFLICT!  
``` 
the parse method contains the bug:

```
urn:ietf:params:oauth:request_uri:xXMGJTZwzXIFL8i_DFu_EM8IeWC___frCWjpiF2q-xs=___1781670640281  
^^^ ^^^  
indexOf() matches triple underscore delimiter  
Result:  
- substring = "frCWjpiF2q-xs=___1781670640281"  
- Long.parseLong(...) → NumberFormatException!  
- catch(Exception) → throwError("invalid_request", "request_uri")  
- User receives error, authentication FAILS  
``` 

**To Reproduce**
Failure probability is 1/6,554 - we see thousands of occurances per day, the sample project demonstrates this with a unit test.

**Expected behavior**
All created URIs can be parsed.

Either the state generation is changed or the parse method accepts URIs with ___ in the state part.

A suggested fix is to change the parse method to replace `indexOf` with `lastIndexOf` when looking for the delimiter position. 

**Sample**

https://github.com/christopherm-backbase/spring-auth-server-par-request-uri-repro

therepanic:
`OAuth2PushedAuthorizationRequestUri` uses Base64URL encoding for the state, which can produce a value containing `___`, the same sequence used as the delimiter. As a result, `indexOf` may locate a delimiter within the state instead of the delimiter preceding the expiration timestamp.

Use `lastIndexOf` to reliably locate the delimiter before the timestamp without changing the existing request URI format.

Closes gh-19444

### DIFF
--- a/oauth2/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2PushedAuthorizationRequestUri.java
+++ b/oauth2/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2PushedAuthorizationRequestUri.java
@@ -27,6 +27,7 @@ import org.springframework.security.crypto.keygen.StringKeyGenerator;
  * Requests.
  *
  * @author Joe Grandja
+ * @author Andrey Litvitski
  * @since 7.0
  */
 final class OAuth2PushedAuthorizationRequestUri {
@@ -60,7 +61,7 @@ final class OAuth2PushedAuthorizationRequestUri {
 
 	static OAuth2PushedAuthorizationRequestUri parse(String requestUri) {
 		int stateStartIndex = REQUEST_URI_PREFIX.length();
-		int expiresAtStartIndex = requestUri.indexOf(REQUEST_URI_DELIMITER) + REQUEST_URI_DELIMITER.length();
+		int expiresAtStartIndex = requestUri.lastIndexOf(REQUEST_URI_DELIMITER) + REQUEST_URI_DELIMITER.length();
 		OAuth2PushedAuthorizationRequestUri pushedAuthorizationRequestUri = new OAuth2PushedAuthorizationRequestUri();
 		pushedAuthorizationRequestUri.requestUri = requestUri;
 		pushedAuthorizationRequestUri.state = requestUri.substring(stateStartIndex);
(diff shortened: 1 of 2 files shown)


oauth2/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2PushedAuthorizationRequestUri.java
```diff
@@ -57,7 +58,7 @@ static OAuth2PushedAuthorizationRequestUri create(Instant expiresAt) {
 
 	static OAuth2PushedAuthorizationRequestUri parse(String requestUri) {
 		int stateStartIndex = REQUEST_URI_PREFIX.length();
-		int expiresAtStartIndex = requestUri.indexOf(REQUEST_URI_DELIMITER) + REQUEST_URI_DELIMITER.length();
+		int expiresAtStartIndex = requestUri.lastIndexOf(REQUEST_URI_DELIMITER) + REQUEST_URI_DELIMITER.length();
 		String state = requestUri.substring(stateStartIndex);
```
therepanic: It does not break backward compatibility.

jzheaux:
Thanks, @therepanic, for the PR! This is now merged into `7.0.x` and `main` and will go out in the next release.

Issue #19436, opened 2026-07-11
Token exchange: the RFC 8693 type identifiers cannot be stated by the application


shin1488:
## Summary

[RFC 8693 section 2.1](https://www.rfc-editor.org/rfc/rfc8693#section-2.1) gives this grant three type parameters — `subject_token_type`, `actor_token_type` and `requested_token_type` — each specified as "an identifier, as described in Section 3". None of the three can be stated by the application. They are broken in three distinct ways.

**`subject_token_type`: the inference is undecidable.** It is derived from the Java type of the subject token by `TokenExchangeGrantRequest.tokenType()`. The same `Jwt` class arrives both when the token was issued by the authorization server being called — where `...:access_token` is correct — and in the cross-domain case, where `...:jwt` is. The correct identifier differs; the Java type does not. This is the framework's own documented path, and Ping ([gh-16486](https://github.com/spring-projects/spring-security/issues/16486)) and Keycloak reject what it sends.

**`subject_token_type`: one identifier has no type at all.** `tokenType()` has two outputs. Section 3 defines five identifiers and leaves the set open; the type hierarchy is closed, and nothing in it denotes a SAML 2.0 assertion — which the Identity Assertion JWT Authorization Grant accepts as a subject token today. Consequence 1 can at least be worked around by misrepresenting the Java type; here there is no type to misrepresent into.

**`requested_token_type`: it can be neither omitted nor changed.** It is set to `...:access_token` unconditionally. That value is right for most exchanges, so the constant is not a defect on its own — but section 2.1 makes it OPTIONAL and gives omission a meaning, and an application cannot omit it. Nor can it send `...:id-jag`, which the Identity Assertion JWT Authorization Grant requires.

No better mapping resolves any of the three: a mapping can only infer an identifier, and what is missing is the ability to state one — `requested_token_type` is not derived from a token at all.

Overriding the request parameters works — it is the answer given in gh-16486, and it is what I do. That is not the ask. The ask is that these identifiers be expressible rather than inferred.

[#19448](https://github.com/spring-projects/spring-security/issues/19448) covers a separable defect in the same method: an `OidcIdToken` subject token is labelled `...:access_token`. A mapping change resolves that one. It does not make any of the three above statable.

Details and sources below. I am happy to submit the PR.

## Expected Behavior

[...]

Issue #16486, opened 2025-01-27
Token exchange with a jwt  access_token send a subject token type 'jwt', my oidc provider only accept 'access_token'


GregoireW:
**Describe the bug**

I use ping as oidc provider. When I try to do a token exchange from springboot application, I got a "[invalid_request]" from ping.

When I drill down a little bit, I end up on the field value of the subject_token_type:

https://github.com/spring-projects/spring-security/blob/7030a62c76b0a0588ac5b99d605bbeabb8d4d783/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/TokenExchangeGrantRequest.java#L114 

basically, I got a jwt access token,  this class make the subject_token_type to "jwt" format  and ping won't accept this value and wait for the access token type ( urn:ietf:params:oauth:token-type:access_token ) 

The RFC do not specify you must use a "jwt" type in the case of a jwt access token, and it specify that "access_token" type is OK even if your access token is jwt.   ( https://datatracker.ietf.org/doc/html/rfc8693#name-token-type-identifiers ) so I guess the certification tool do not check that, and not supporting the 'jwt' type is fine at this level.

**To Reproduce**

Need to use a ping server to bootstrap the token exchange connection and use it.

**Expected behavior**

Is there a way to "force" the use of access_token type instead of jwt ?

**Sample**

-

sjohnr:
@GregoireW thanks for reaching out!
 
> Is there a way to "force" the use of access_token type instead of jwt ?

Generally, we prefer questions be asked on stack overflow. However, given that this question could require discussion that might require opening an issue, I'll respond directly here.

Yes, it is possible to customize the OAuth 2.0 Access Token Request for the `token-exchange` grant ([RFC 8693](https://datatracker.ietf.org/doc/html/rfc8693)). The [Token Exchange Grant](https://docs.spring.io/spring-security/reference/servlet/oauth2/client/authorization-grants.html#oauth2-client-token-exchange) is discussed in the reference docs, specifically [Customizing Request Parameters for Token Exchange](https://docs.spring.io/spring-security/reference/servlet/oauth2/client/authorization-grants.html#oauth2-client-token-exchange-access-token-request-parameters).

One way to do this would be to _override_ parameters with `setParametersConverter`, like so:

```java
@Bean
public OAuth2AccessTokenResponseClient<TokenExchangeGrantRequest> accessTokenResponseClient() {
[...]

PR #19450, opened 2026-07-15
Fix OIDC ID token exchange token type

### LINKED ISSUE #19448: `TokenExchangeGrantRequest` sends an `OidcIdToken` subject token as `...:access_token`
## Describe the bug

[`TokenExchangeGrantRequest.tokenType()`](https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/TokenExchangeGrantRequest.java) derives `subject_token_type` (and `actor_token_type`) from the Java type of the token:

```java
private static String tokenType(OAuth2Token token) {
    return (token instanceof Jwt) ? JWT_TOKEN_TYPE_VALUE : ACCESS_TOKEN_TYPE_VALUE;
}
```

[`OidcIdToken`](https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/oidc/OidcIdToken.java) and [`Jwt`](https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/Jwt.java) are siblings, not parent and child — both extend [`AbstractOAuth2Token`](https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/AbstractOAuth2Token.java):

```java
// oauth2-core
public class OidcIdToken extends AbstractOAuth2Token implements IdTokenClaimAccessor { ... }

// oauth2-jose  (depends on oauth2-core)
public class Jwt extends AbstractOAuth2Token implements JwtClaimAccessor { ... }
```

An `OidcIdToken` is therefore an `OAuth2Token` but not a `Jwt`, so it takes the `ACCESS_TOKEN_TYPE_VALUE` branch. An ID token is labelled an access token, and there is no scenario in which that is the correct identifier.

[RFC 8693 section 3](https://www.rfc-editor.org/rfc/rfc8693#section-3) defines `urn:ietf:params:oauth:token-type:access_token` as indicating "an OAuth 2.0 access token issued by the given authorization server", and defines `urn:ietf:params:oauth:token-type:id_token` separately for an ID token.

This is not an oversight that a subclass would fix. `OidcIdToken` lives in `oauth2-core` and `Jwt` lives in `oauth2-jose`, which depends on `oauth2-core`; the dependency direction means `OidcIdToken` cannot extend `Jwt`. So `instanceof Jwt` has never been — and can never be — true for an ID token, even though an ID token is by definition a JWT.

The behaviour has been present since the grant was introduced in 6.3. The logic moved from [`TokenExchangeGrantRequestEntityConverter`](https://github.com/spring-projects/spring-security/blob/6.3.x/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/TokenExchangeGrantRequestEntityConverter.java) to `TokenExchangeGrantRequest.defaultParameters()` in 6.4; the method itself is unchanged. Confirmed on `main`.

## To Reproduce

Public API only — no external project or custom resolver needed:

```java
ClientRegistration registration = ClientRegistration.withRegistrationId("exchange")
        .clientId("client")
        .authorizationGrantType(AuthorizationGrantType.TOKEN_EXCHANGE)
        .tokenUri("https://example.com/oauth2/token")
        .build();

OidcIdToken idToken = OidcIdToken.withTokenValue("id-token").claim("sub", "user").build();

MultiValueMap<String, String> parameters =
        new DefaultOAuth2TokenRequestParametersConverter<TokenExchangeGrantRequest>()
                .convert(new TokenExchangeGrantRequest(registration, idToken, null));

parameters.getFirst("subject_token_type");
// -> "urn:ietf:params:oauth:token-type:access_token"
```

## Expected behavior

`subject_token_type` should be `urn:ietf:params:oauth:token-type:id_token`.

## How this is reached in practice

An `OidcIdToken` reaches the grant request through a custom subject token resolver — for example one returning `oidcUser.getIdToken()`. No `Authentication` shipped with Spring Security carries an `OidcIdToken` as its principal, so the default resolver never produces one.

That is the "Federated Identity" and "Backend-to-Backend with User Context" use cases raised in [gh-19048](https://github.com/spring-projects/spring-security/issues/19048#issuecomment-4199641042) and acknowledged there as ["solid use cases for `id_token` exchange"](https://github.com/spring-projects/spring-security/issues/19048#issuecomment-4244352313) — in both, a Spring application is the client of the exchange and the subject token is an ID token. [gh-19076](https://github.com/spring-projects/spring-security/pull/19076) is implementing the server half now: an `OidcIdTokenSubjectTokenResolver` reading the IdP's JWKS URL from `ClientSettings.idTokenJwkSetUrl()`, whose first test scenario exchanges a valid ID token and expects an access token with the correct `sub`.

## Relationship to #19436

[#19436](https://github.com/spring-projects/spring-security/issues/19436) asks that the RFC 8693 type identifiers be stated rather than inferred. That ask stands whatever happens here: no better mapping resolves any of its three defects — a mapping can only infer an identifier, and what is missing is the ability to state one.

This one is separable: gh-19436 stands without it, and a mapping change resolves this while leaving gh-19436's three defects untouched. And whatever resolver strategy eventually lands, `tokenType()` remains the default — an application that does not state an identifier still gets one from this method, and for an `OidcIdToken` it will still be wrong.

Mapping `OidcIdToken` to `...:id_token` is not a behaviour change for anyone who is correct today, because labelling an ID token as an access token is never right.

I am happy to submit the PR.

xfocus3:
Fixes gh-19448.

This maps `OidcIdToken` to `urn:ietf:params:oauth:token-type:id_token` when building token exchange request parameters. The default converter continues to use the existing fallback for other token classes.

Regression coverage checks OIDC ID tokens used as both `subject_token` and `actor_token`.

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/TokenExchangeGrantRequest.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/endpoint/TokenExchangeGrantRequest.java
@@ -22,6 +22,7 @@ import org.springframework.security.oauth2.client.registration.ClientRegistratio
 import org.springframework.security.oauth2.core.AuthorizationGrantType;
 import org.springframework.security.oauth2.core.OAuth2Token;
 import org.springframework.security.oauth2.core.endpoint.OAuth2ParameterNames;
+import org.springframework.security.oauth2.core.oidc.OidcIdToken;
 import org.springframework.security.oauth2.jwt.Jwt;
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
@@ -48,6 +49,8 @@ import org.springframework.util.StringUtils;
 public class TokenExchangeGrantRequest extends AbstractOAuth2AuthorizationGrantRequest {
 
 	private static final String ACCESS_TOKEN_TYPE_VALUE = "urn:ietf:params:oauth:token-type:access_token";
+
+	private static final String ID_TOKEN_TYPE_VALUE = "urn:ietf:params:oauth:token-type:id_token";
 
 	private static final String JWT_TOKEN_TYPE_VALUE = "urn:ietf:params:oauth:token-type:jwt";
 
@@ -113,6 +116,9 @@ public class TokenExchangeGrantRequest extends AbstractOAuth2AuthorizationGrantR
 	}
 
 	private static String tokenType(OAuth2Token token) {
+		if (token instanceof OidcIdToken) {
+			return ID_TOKEN_TYPE_VALUE;
+		}
 		return (token instanceof Jwt) ? JWT_TOKEN_TYPE_VALUE : ACCESS_TOKEN_TYPE_VALUE;
 	}
 
--- a/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/endpoint/DefaultOAuth2TokenRequestParametersConverterTests.java
+++ b/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/endpoint/DefaultOAuth2TokenRequestParametersConverterTests.java
@@ -220,4 +220,22 @@ public class DefaultOAuth2TokenRequestParametersConverterTests {
 		assertThat(parameters.get(OAuth2ParameterNames.SUBJECT_TOKEN_TYPE)).containsExactly(ID_TOKEN_TYPE_VALUE);
 	}
 
+	@Test
+	public void convertWhenGrantRequestIsTokenExchangeAndActorTokenIsOidcIdTokenThenActorTokenTypeIsIdToken() {
+		ClientRegistration clientRegistration = this.clientRegistration
+			.authorizationGrantType(AuthorizationGrantType.TOKEN_EXCHANGE)
+			.build();
+		OAuth2Token subjectToken = TestOAuth2AccessTokens.scopes("read", "write");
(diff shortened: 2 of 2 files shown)


shin1488:
The mapping is correct. The description is not.

"That correctly maps `Jwt` to `urn:ietf:params:oauth:token-type:jwt`" is the part gh-19436 (open, linked from gh-19448) disputes. RFC 8693 section 3 defines `...:jwt` for sending a JWT to a *different* authorization server, and `...:access_token` for a token issued by the authorization server being called — both arrive as a `Jwt`, so the Java type cannot tell them apart. The reference documentation's own token exchange example is the second case, where this mapping sends `...:jwt`, and Ping (gh-16486) and Keycloak reject it.

`defaultParameters()` also calls `tokenType()` for `actor_token_type`, and `convertWhenGrantRequestIsTokenExchangeThenParametersProvided` in this file already asserts it for a `Jwt` actor token. The new test passes `null` as the actor token, so the `OidcIdToken` actor case — which this change also alters — is not covered.

gh-19436 has the rest: widening the mapping closes one identifier and leaves the application no more able to state one than before. gh-19076 is adding a resolver strategy for this same identifier on the authorization server side — this is the opposite shape on the client side. This PR does not close gh-19436.

Issue #17847, opened 2025-09-04
Move Access API to Separate Module


jzheaux:
The Access API has been deprecated for a few years now, though some applications continue to need it. We can move these into a separate legacy jar so that they remain accessible while also cleaning up the codebase and still encouraging migration.

PR #19460, opened 2026-07-20
Fail fast when spring-security-access is missing

### LINKED ISSUE #19441: NoClassDefFoundError: MethodSecurityMetadataSourceAdvisor missing from spring-security-core 7.1.0 but still referenced by spring-security-config
### Describe the bug

`spring-security-config:7.1.0` still references 
`org.springframework.security.access.intercept.aopalliance.MethodSecurityMetadataSourceAdvisor` 
inside `MethodSecurityMetadataSourceAdvisorRegistrar`, but this class is no longer 
present in `spring-security-core:7.1.0`. This causes a `NoClassDefFoundError` at 
application startup.

### Evidence
jar tf spring-security-core-7.1.0.jar | findstr MethodSecurityMetadataSourceAdvisor
_(no output - class not present)_
jar tf spring-security-config-7.1.0.jar | findstr MethodSecurityMetadataSourceAdvisorRegistrar
_org/springframework/security/config/annotation/method/configuration/MethodSecurityMetadataSourceAdvisorRegistrar.class_

Decompiled bytecode of `MethodSecurityMetadataSourceAdvisorRegistrar.registerBeanDefinitions` 
still contains:
_ldc #7  // class org/springframework/security/access/intercept/aopalliance/MethodSecurityMetadataSourceAdvisor
invokestatic #9  // BeanDefinitionBuilder.rootBeanDefinition(Class)_

### To Reproduce

1. Spring Boot 4.1.0 project with `spring-boot-starter-security` and 
   `spring-boot-starter-oauth2-resource-server` (which pulls in 
   `spring-security-config:7.1.0` transitively via `spring-boot-security:4.1.0`)
2. Start the application

### Expected behavior

Application context starts normally.

### Actual behavior
```
java.lang.NoClassDefFoundError: org/springframework/security/access/intercept/aopalliance/MethodSecurityMetadataSourceAdvisor
at org.springframework.security.config.annotation.method.configuration.MethodSecurityMetadataSourceAdvisorRegistrar.registerBeanDefinitions(MethodSecurityMetadataSourceAdvisorRegistrar.java:47)
at org.springframework.context.annotation.ImportBeanDefinitionRegistrar.registerBeanDefinitions(ImportBeanDefinitionRegistrar.java:86)
...
Caused by: java.lang.ClassNotFoundException: org.springframework.security.access.intercept.aopalliance.MethodSecurityMetadataSourceAdvisor
```
### Workaround

Pinning `spring-security.version` to `7.0.6` resolves the issue.

### Environment

- Spring Boot: 4.1.0
- Spring Security: 7.1.0
- Spring Framework: 7.0.8
- Java: 17
- Build tool: Maven

### Note

We do not use `@EnableGlobalMethodSecurity` directly in our code — only 
`@EnableMethodSecurity`. It's unclear what triggers the import of 
`MethodSecurityMetadataSourceAdvisorRegistrar` in this scenario; it may be 
triggered indirectly by autoconfiguration.

jyx-07:
## Summary

Fixes #19441.

`spring-security-config`'s deprecated legacy method-security support throws a raw `NoClassDefFoundError` at startup instead of a clear error when `spring-security-access` isn't on the classpath.

## Root cause

#17847 moved `MethodSecurityMetadataSourceAdvisor` and `MethodSecurityInterceptor` from `spring-security-core` (a mandatory dependency of `spring-security-config`) into the new `spring-security-access` module, which `spring-security-config` only depends on **optionally**. Several deprecated method-security paths still construct these types unconditionally:

- `GlobalMethodSecuritySelector` → `MethodSecurityMetadataSourceAdvisorRegistrar` (proxy mode) and `GlobalMethodSecurityConfiguration` (imported for **both** proxy and aspectj mode)
- `ReactiveMethodSecuritySelector` → `ReactiveMethodSecurityConfiguration` (`@EnableReactiveMethodSecurity(useAuthorizationManager = false)`)

`@EnableMethodSecurity` and `@EnableReactiveMethodSecurity`'s default mode — the non-deprecated replacements — never reference these classes and are unaffected.

## Fix

Added a `ClassUtils.isPresent(...)` guard (the same pattern already used for other optional dependencies in these selectors) so the legacy paths fail fast with a clear `IllegalStateException` naming the missing dependency, instead of a `NoClassDefFoundError` deep in Spring's configuration-processing machinery.

Chose this over making `spring-security-access` mandatory again, since that would silently reintroduce the footprint #17847 removed for the majority of apps using `@EnableMethodSecurity`. `@EnableGlobalMethodSecurity` stays deprecated; this adds no new investment in it beyond a clear diagnostic.

## Testing

Added `Gh19441Tests`, using the existing `@ClassPathExclusions` harness to simulate `spring-security-access` being absent:

- `@EnableMethodSecurity` and `@EnableReactiveMethodSecurity` default mode start cleanly (locks in that the modern paths are unaffected).
- `@EnableGlobalMethodSecurity` in proxy mode, aspectj mode, and `@EnableReactiveMethodSecurity(useAuthorizationManager = false)` all now fail with the clear message instead of `NoClassDefFoundError`/`ClassNotFoundException`.

Full `spring-security-config` suite passes (3640 tests; the only 3 failures are pre-existing, locale-related, and unrelated to this change). `./gradlew :spring-security-config:check` passes.

### DIFF
--- a/config/src/main/java/org/springframework/security/config/annotation/method/configuration/GlobalMethodSecuritySelector.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/method/configuration/GlobalMethodSecuritySelector.java
@@ -39,6 +39,8 @@ import org.springframework.util.ClassUtils;
 @Deprecated
 final class GlobalMethodSecuritySelector implements ImportSelector {
 
+	private static final String METHOD_SECURITY_METADATA_SOURCE_ADVISOR = "org.springframework.security.access.intercept.aopalliance.MethodSecurityMetadataSourceAdvisor";
+
 	@Override
 	public String[] selectImports(AnnotationMetadata importingClassMetadata) {
 		Class<EnableGlobalMethodSecurity> annoType = EnableGlobalMethodSecurity.class;
@@ -57,6 +59,16 @@ final class GlobalMethodSecuritySelector implements ImportSelector {
 		String autoProxyClassName = isProxy ? AutoProxyRegistrar.class.getName()
 				: GlobalMethodSecurityAspectJAutoProxyRegistrar.class.getName();
 		boolean jsr250Enabled = attributes.getBoolean("jsr250Enabled");
+		if (isProxy || !skipMethodSecurityConfiguration) {
+			// The proxy-mode advisor registrar and GlobalMethodSecurityConfiguration
+			// (imported for both proxy and aspectj modes) need types that only exist in
+			// the optional spring-security-access module.
+			Assert.state(
+					ClassUtils.isPresent(METHOD_SECURITY_METADATA_SOURCE_ADVISOR, ClassUtils.getDefaultClassLoader()),
+					() -> "@EnableGlobalMethodSecurity requires the spring-security-access dependency on the "
+							+ "classpath. Please add spring-security-access, or migrate to @EnableMethodSecurity "
+							+ "which does not require it.");
+		}
 		List<String> classNames = new ArrayList<>(4);
 		if (isProxy) {
 			classNames.add(MethodSecurityMetadataSourceAdvisorRegistrar.class.getName());
--- a/config/src/main/java/org/springframework/security/config/annotation/method/configuration/GlobalMethodSecuritySelector.java
+++ b/config/src/main/java/org/springframework/security/config/annotation/method/configuration/GlobalMethodSecuritySelector.java
@@ -39,7 +39,8 @@ import org.springframework.util.ClassUtils;
 @Deprecated
 final class GlobalMethodSecuritySelector implements ImportSelector {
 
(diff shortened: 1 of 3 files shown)


jzheaux:
Thanks for the contribution, @jyx-07! 

Note I've rebased this onto 7.0.x since I've confirmed the bug there as well.

Regarding the solution, the `IllegalStateException` is an improvement, thank you. I've adjusted the PR to place the checks closer to the problematic code itself. The reason is that the boolean logic originally proposed was missing a check for `jsr250Enabled`, something else that also requires spring-security-access — which is a hint that we should reuse the existing boolean logic instead of replicating it. I've also removed the comment, since moving the behavior into a separate method makes the code self-documenting. I also moved the `ClassUtils` check in order to play nicely with Framework's AOT support. I also added a test for `jsr250Enabled` to make sure we have that covered.

jzheaux:
Thanks, @jyx-07, for the PR! This is now merged into `7.0.x` and `main` and will go out in the next release.

PR #19468, opened 2026-07-24
Introduce Scoped Value-based SecurityContextHolderStrategy

### LINKED ISSUE #19259: ScopedValue-based implementation of SecurityContextHolderStrategy
Spring Framework provides an extensive support for virtual threads. Virtual thread guidelines, however, [recommend]( https://docs.oracle.com/en/java/javase/21/core/virtual-threads.html#GUID-68216B85-7B43-423E-91BA-11489B1ACA61) to replace the `ThreadLocal` with `ScopedValue` while Spring Security is one of the systems which uses `ThreadLocal`s extensively - its default implementation of `SecurityContextHolderStrategy` is `ThreadLocal`-based. Although it hasn't been officially announced, as of JDK 25 `ScopedValue` is no longer annotated with `PreviewFeature`.

Does it make sense to implement `ScopedValue`-based `SecurityContextHolderStrategy` - even though the `SecurityContext` or `Authentication` are not necessarily _expensive_ reusable objects as the Oracle guidelines state?

This issue has been addressed more than a year ago in a Stack Overflow thread [Spring Security Virtual Threads and ThreadLocal](https://stackoverflow.com/questions/78166517/spring-security-virtual-threads-and-threadlocal) and in [my answer](https://stackoverflow.com/a/78224869/2366397) to it. The [small POC project](https://github.com/eager-zh/spring-security-scoped-value), brought in the answer, offers some solution to this issue; this solution, non-systematic and ad-hoc, may, nevertheless, serve as a starting point of the discussion. The problems and ramifications of the solution are also discussed in a readme file of the project.

eager-zh:
Closes gh-19259.

The implementation of Scoped Value-based `SecurityContextHolderStrategy`, henceforth Strategy, is rather straightforward: 

-  a `ScopedValue` instance holds a reference to `SecurityContext` similar to that of `ThreadLocal`;

-  `SecurityContextHolder` is changed to be able to be configured with new `SecurityContextHolderStrategy`;

- a newly introduced `ScopedSecurityContextHolderFilter` orchestrates the usage of new Strategy, invoking upstack/downstream Security Filters in a Filter Chain with the `ScopedValue` instance bound to `SecurityContext` for the current thread;

- `SecurityContextConfigurer` is changed to replace `SecurityContextHolderFilter` with `ScopedSecurityContextHolderFilter` _if_ current, globally configured Strategy is Scoped Value-based one;

- it was unnecessary, IMHO, to alter XML-based configuration as custom Strategy can be set for a particular Filter Chain by `security-context-holder-strategy-ref` attribute of `http` XML element.

	
What is not that straightforward is the configuration of Java compiler to be able to access Java 25 features, `ScopedValue` class in this case. For that purpose, Java compiler option `--release` was replaced with `--source` and `--target` options in `java-toolchain.gradle`; it also turned out to be necessary to set JVM target to 17 for Kotlin test compilation in `test-compile-target-jdk25.gradle`. As a result of these changes, the project now builds, uses `ScopedValue`, and still can run under JVM 17.
	
It is impossible to replace `Subject.doAs` calls with its Java 18 replacement `Subject.callAs` due to the requirement of compatibility with Java 17, and the correspondent deprecation warnings are suppressed.

### DIFF
--- a/buildSrc/src/main/groovy/java-toolchain.gradle
+++ b/buildSrc/src/main/groovy/java-toolchain.gradle
@@ -17,7 +17,8 @@ java {
 tasks.withType(JavaCompile).configureEach {
 	options.encoding = "UTF-8"
 	options.compilerArgs.add("-parameters")
-	options.release = 17
+    sourceCompatibility = '17'
+    targetCompatibility = '17'
 }
 
 pluginManager.withPlugin("org.jetbrains.kotlin.jvm") {
--- a/buildSrc/src/main/groovy/test-compile-target-jdk25.gradle
+++ b/buildSrc/src/main/groovy/test-compile-target-jdk25.gradle
@@ -25,6 +25,6 @@ tasks.withType(JavaCompile).configureEach { task ->
 
 tasks.withType(KotlinCompile).configureEach { task ->
 	if (task.name == 'compileTestKotlin' || task.name == 'compileIntegrationTestKotlin') {
-		task.kotlinOptions.jvmTarget = '25'
+		task.kotlinOptions.jvmTarget = '17'
 	}
 }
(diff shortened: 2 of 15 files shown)


zhujianxintian:
This PR has some incompatibility issues. For details, please refer to https://github.com/spring-projects/spring-security/issues/19259#issuecomment-5563951840

Issue #19461, opened 2026-07-20
Improve reference documentation on Forwarded vs X-Forwarded headers


rstoyanchev:
Most proxies currently support the alternative `"X-Forwarded-*"` headers rather than the standard "`Forwarded`" header from RFC 7239 while the Spring Framework and servers like Reactor Netty and Jetty support both. That means a proxy at the edge needs to be configured correctly to drop untrusted values.

The [current coverage of forwarded headers](https://docs.spring.io/spring-security/reference/features/exploits/http.html#http-proxy-server) in the reference docs conflates RFC 7239 and X-Forwarded headers a bit, but should instead guide towards differentiating between both types of headers, and ensuring that proxies are configured correctly to with awareness of both types of forwarded headers.

jzheaux:
Thanks for flagging this, @rstoyanchev — #19477 fixes it. Setting the milestone there.

PR #19477, opened 2026-07-29
Differentiate Forwarded and X-Forwarded headers in proxy docs

skdas20:
See gh-19461.

The proxy server section currently says:

> To fix this, you can use RFC 7239 to specify that a load balancer is being used.
> To make the application aware of this, you need to configure your application server to be aware of the X-Forwarded headers.

which points at RFC 7239 and then, in the next sentence, refers to the `X-Forwarded-*` headers as though they were the same mechanism. They are not: RFC 7239 defines a single standard `Forwarded` header, while `X-Forwarded-Host` / `X-Forwarded-Proto` / `X-Forwarded-For` are the older non-standard set.

This change:

- describes the two kinds of headers separately, so a reader can tell which applies to their deployment;
- notes that most proxies still send `X-Forwarded-*` while Spring Framework and servers such as Reactor Netty and Jetty understand both, so neither can be assumed to be the only one in use;
- adds a note that both kinds are client-supplied unless a proxy overwrites them, and states that the edge proxy has to drop or overwrite untrusted values for **both** kinds — dropping only one set still leaves the application spoofable through the other;
- mentions that `ForwardedHeaderFilter` / `ForwardedHeaderTransformer` handle both kinds and can be configured to remove the headers instead of applying them.

The wording follows the framing in the issue. Only `features/exploits/http.adoc` needed changing — the servlet and reactive pages `xref` this section rather than repeating it.

Happy to adjust the emphasis or trim it if you had a different structure in mind; this is meant as a starting point rather than a finished opinion on the guidance.

### DIFF
--- a/docs/modules/ROOT/pages/features/exploits/http.adoc
+++ b/docs/modules/ROOT/pages/features/exploits/http.adoc
@@ -23,10 +23,28 @@ When using a proxy server, it is important to ensure that you have configured yo
 For example, many applications have a load balancer that responds to request for `\https://example.com/` by forwarding the request to an application server at `\https://192.168.0.107`
 Without proper configuration, the application server can not know that the load balancer exists and treats the request as though `\https://192.168.0.107:8080` was requested by the client.
 
-To fix this, you can use https://tools.ietf.org/html/rfc7239[RFC 7239] to specify that a load balancer is being used.
-To make the application aware of this, you need to configure your application server to be aware of the X-Forwarded headers.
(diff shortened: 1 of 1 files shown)


jzheaux:
Rebased this onto `7.0.x` since the same conflated text exists there too. No content changes.

jzheaux:
Thanks, @skdas20, for the PR! This is now merged into `7.0.x` and `main` and will go out in the next release.

PR #19493, opened 2026-08-01
Add reactive protected resource metadata discovery

### LINKED ISSUE #19446: Align Reactive BearerTokenServerAuthenticationEntryPoint with Servlet RFC 9728 WWW-Authenticate behavior
## Expected Behavior

The Reactive OAuth2 Resource Server should produce the same default `WWW-Authenticate` header as the Servlet implementation.

Since Spring Security 7 / Spring Boot 4.1, the Servlet implementation includes RFC 9728 Bearer challenge metadata in the `WWW-Authenticate` header by default. The Reactive implementation should provide the same behavior for consistency across both stacks.

## Current Behavior

The Servlet stack returns a `WWW-Authenticate` header containing the additional Bearer challenge metadata defined by RFC 9728.

For example:

```http
WWW-Authenticate: Bearer resource_metadata="https://example.com/.well-known/oauth-protected-resource"
```

However, the Reactive stack still returns only:

```http
WWW-Authenticate: Bearer
```

As a result, equivalent applications built on the Servlet and Reactive stacks produce different default authentication challenge responses.

## Context

This appears to be an inconsistency between the Servlet and Reactive implementations.

Relevant references:

- RFC 9728 (OAuth 2.0 Protected Resource Metadata), Section 5.1 – WWW-Authenticate Response:
  https://www.rfc-editor.org/rfc/rfc9728

- Servlet implementation (`BearerTokenAuthenticationEntryPoint`):
  https://docs.spring.io/spring-security/reference/api/java/org/springframework/security/oauth2/server/resource/web/BearerTokenAuthenticationEntryPoint.html

- Reactive implementation (`BearerTokenServerAuthenticationEntryPoint`):
  https://docs.spring.io/spring-security/reference/api/java/org/springframework/security/oauth2/server/resource/web/server/BearerTokenServerAuthenticationEntryPoint.html

Would it make sense to align the Reactive implementation with the Servlet implementation so that both produce the same default `WWW-Authenticate` header?

therepanic:
We support [RFC 9728](https://www.rfc-editor.org/rfc/rfc9728) Protected Resource Metadata discovery for servlet-based applications, but have not done so for reactive resource servers. We should make this behavior consistent across both stacks.

Closes: gh-19446

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/server/OAuth2ProtectedResourceMetadataWebFilter.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/server/OAuth2ProtectedResourceMetadataWebFilter.java
@@ -1,0 +1,138 @@
+/*
+ * Copyright 2004-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.oauth2.server.resource.web.server;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.function.Consumer;
+
+import org.jspecify.annotations.Nullable;
+import reactor.core.publisher.Mono;
+
+import org.springframework.core.ResolvableType;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.MediaType;
+import org.springframework.http.codec.HttpMessageWriter;
+import org.springframework.http.codec.ServerCodecConfigurer;
+import org.springframework.security.oauth2.server.resource.OAuth2ProtectedResourceMetadata;
+import org.springframework.security.web.server.util.matcher.ServerWebExchangeMatcher;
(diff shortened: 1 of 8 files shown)


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/server/OAuth2ProtectedResourceMetadataWebFilter.java
```diff
@@ -0,0 +1,138 @@
...
+	@SuppressWarnings("unchecked")
+	private static @Nullable HttpMessageWriter<Object> getJsonMessageWriter() {
+		return ServerCodecConfigurer.create()
+			.getWriters()
+			.stream()
+			.filter((writer) -> writer.canWrite(STRING_OBJECT_MAP, MediaType.APPLICATION_JSON))
+			.map((writer) -> (HttpMessageWriter<Object>) writer)
+			.findFirst()
+			.orElse(null);
+	}
```
therepanic, line 136: Perhaps there is a workaround on the Spring side like this for servlets to replace [OAuth2ProtectedResourceMetadataFilter#HttpMessageConverters](https://github.com/spring-projects/spring-security/blob/61fae604d2ea7e0bf947313686b4de77fa1b906e/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/OAuth2ProtectedResourceMetadataFilter.java#L144)

therepanic:
`OAuth2ProtectedResourceMetadataWebFilterTests` contains tests similar to the [servet alternative](https://github.com/spring-projects/spring-security/blob/61fae604d2ea7e0bf947313686b4de77fa1b906e/oauth2/oauth2-resource-server/src/test/java/org/springframework/security/oauth2/server/resource/web/OAuth2ProtectedResourceMetadataFilterTests.java). `OAuth2ProtectedResourceMetadataTests` [too](https://github.com/spring-projects/spring-security/blob/main/config/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/resource/OAuth2ProtectedResourceMetadataTests.java).

PR #19494, opened 2026-08-02
Exclude null and wildcard from external matching

### LINKED ISSUE #19072: `InetAddressMatchers.matchExternal()` matches `null` and `0.0.0.0`

ryandiginomad:
`ExternalInetAddressMatcher` is defined as `!InternalInetAddressMatcher`, and the internal matcher returns `false` for a `null` address and for the wildcard addresses `0.0.0.0` and `::` — they are neither loopback, link-local, site-local, nor ULA, so they fall through to the final `return false`. Negating that classified all three as external, which is what gh-19072 reports.

This returns `false` for those cases before delegating, so they now match neither `matchExternal()` nor `matchInternal()`. That asymmetry looks deliberate to me: none of the three identifies a host a request could originate from, so neither classification is meaningful — but if you would rather the wildcard addresses count as internal, that is a one-line change instead.

`InetAddress.isAnyLocalAddress()` covers both wildcard forms. I checked the IPv4-mapped spelling too, since it is the easy one to miss:

```
0.0.0.0          -> Inet4Address   anyLocal=true   loopback=false  siteLocal=false
::               -> Inet6Address   anyLocal=true   loopback=false  siteLocal=false
::ffff:0.0.0.0   -> Inet4Address   anyLocal=true   loopback=false  siteLocal=false
::1              -> Inet6Address   anyLocal=false  loopback=true   siteLocal=false
127.0.0.1        -> Inet4Address   anyLocal=false  loopback=false  siteLocal=false
8.8.8.8          -> Inet4Address   anyLocal=false  loopback=false  siteLocal=false
```

The JDK folds `::ffff:0.0.0.0` into an `Inet4Address`, so it is covered by the same check, and loopback/public addresses are untouched.

Closes gh-19072

### Testing

Three tests added to `ExternalInetAddressMatcherTests`, following the surrounding naming convention and `@ValueSource` style: one for `matches((InetAddress) null)` and a parameterized one for `0.0.0.0` and `::`. I confirmed all three fail on `main` before the change and pass after it.

`./gradlew :spring-security-core:check` is green — 1507 tests across 185 classes, 0 failures, checkstyle clean. I ran it with `-PtestToolchain=21` because I do not have JDK 25 locally; `options.release` is 17 regardless.

`matchExternal` and `ExternalInetAddressMatcher` have no other callers in the repository, so the change is contained to this API.

### 🤖 AI assistance

Developed with Claude Code. I reviewed every line of this diff, and the verification above reflects test runs I actually performed and observed.

### DIFF
--- a/core/src/main/java/org/springframework/security/util/matcher/InetAddressMatchers.java
+++ b/core/src/main/java/org/springframework/security/util/matcher/InetAddressMatchers.java
@@ -48,6 +48,9 @@ public final class InetAddressMatchers {
 
 	/**
 	 * Creates a new builder configured to match external (non-private) IP addresses.
+	 * <p>
+	 * A {@code null} address and the wildcard addresses ({@code 0.0.0.0} and {@code ::})
+	 * are not treated as external.
 	 * @return a {@link Builder} configured to match external addresses
 	 */
 	public static Builder matchExternal() {
@@ -314,6 +317,10 @@ public final class InetAddressMatchers {
 	 * External addresses are any addresses that are not internal (private) addresses.
 	 * This matcher delegates to {@link InternalInetAddressMatcher} and negates the
 	 * result.
+	 * <p>
+	 * A {@code null} address and the wildcard addresses ({@code 0.0.0.0} and {@code ::})
+	 * are not external. They do not identify a host that a request could originate from,
+	 * so negating the internal check is not meaningful for them.
 	 *
 	 * @author Gábor Vaspöri
 	 * @author Kian Jamali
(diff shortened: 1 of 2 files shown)


ryandiginomad:
Closing this. After 15a31d7dd7 and 057581584a, `matchExternal()` on main already returns false for null and for the wildcard addresses, so there's nothing left here to merge.

Something I noticed while checking: `ExternalInetAddressMatcherTests` has no wildcard case. `0.0.0.0` and `::` appear only in the internal tests, asserting true, so the behavior reported in gh-19072 is covered only transitively right now.

PR #19512, opened 2026-08-07
Fix Bearer token parsing for multiple spaces

### LINKED ISSUE #19500: DefaultBearerTokenResolver rejects RFC-legal multi-space `Authorization` header
**Component:** spring-security-oauth2-resource-server · **Affects:** 6.5.x, 7.1.0, current `main`

## Problem

`DefaultBearerTokenResolver` matches the header against
`^Bearer (?<token>[a-zA-Z0-9-._~+/]+=*)$` — a single literal space. RFC 6750 §2.1
(`credentials = "Bearer" 1*SP b64token`) and RFC 9110 §11.6.2 (`1*SP`) allow **one or more**
spaces. A conformant `Authorization: Bearer␣␣<token>` is therefore rejected with
`invalid_token` / "Bearer token is malformed".

## Reproduce
Send an otherwise valid request with two spaces after `Bearer`:
```
Authorization: Bearer  eyJhbGciOi...
```
→ HTTP 401, `WWW-Authenticate: Bearer error="invalid_token", error_description="Bearer token is malformed"`.

The token never reaches the decoder; it fails at header resolution.

## Expected
Per `1*SP`, the token is resolved and validated normally (→ 200 for a valid token).

## Notes
- Masked by some servlet containers: Undertow (2.3.24) collapses the double space to one so
  Spring accepts it; Jetty preserves it verbatim so Spring rejects it. Switching containers
  surfaces the latent rejection with no client change.
- Regex identical in 6.5.11 and 7.1.0.

## Suggested fix
Match `1*SP`:
```java
Pattern.compile("^Bearer +(?<token>[a-zA-Z0-9-._~+/]+=*)$", Pattern.CASE_INSENSITIVE);
```

## Failing test
```java
@Test
public void resolveWhenHeaderHasMultipleSpacesBeforeTokenThenTokenIsResolved() {
	// RFC 6750 section 2.1: credentials = "Bearer" 1*SP b64token (one or more spaces)
	MockHttpServletRequest request = new MockHttpServletRequest();
	request.addHeader("Authorization", "Bearer  " + TEST_TOKEN);
	assertThat(this.resolver.resolve(request)).isEqualTo(TEST_TOKEN);
}
```

therepanic:
[RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750#section-2.1) allows one or more spaces. This change updates Bearer token parsing accordingly.

Closes: gh-19500

### DIFF
--- a/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/server/authentication/ServerBearerTokenAuthenticationConverter.java
+++ b/oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/server/authentication/ServerBearerTokenAuthenticationConverter.java
@@ -44,6 +44,7 @@ import org.springframework.web.server.ServerWebExchange;
  * Token</a>s from the {@link ServerWebExchange}.
  *
  * @author Rob Winch
+ * @author Andrey Litvitski
  * @since 5.1
  * @see <a href="https://tools.ietf.org/html/rfc6750#section-2" target="_blank">RFC 6750
  * Section 2: Authenticated Requests</a>
@@ -52,7 +53,7 @@ public class ServerBearerTokenAuthenticationConverter implements ServerAuthentic
 
 	private static final String ACCESS_TOKEN_PARAMETER_NAME = "access_token";
 
-	private static final Pattern authorizationPattern = Pattern.compile("^Bearer (?<token>[a-zA-Z0-9-._~+/]+=*)$",
+	private static final Pattern authorizationPattern = Pattern.compile("^Bearer +(?<token>[a-zA-Z0-9-._~+/]+=*)$",
 			Pattern.CASE_INSENSITIVE);
 
 	private boolean allowFormEncodedBodyParameter = false;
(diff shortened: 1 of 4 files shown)


oauth2/oauth2-resource-server/src/main/java/org/springframework/security/oauth2/server/resource/web/server/authentication/ServerBearerTokenAuthenticationConverter.java
```diff
@@ -52,7 +53,7 @@ public class ServerBearerTokenAuthenticationConverter implements ServerAuthentic
 
 	private static final String ACCESS_TOKEN_PARAMETER_NAME = "access_token";
 
-	private static final Pattern authorizationPattern = Pattern.compile("^Bearer (?<token>[a-zA-Z0-9-._~+/]+=*)$",
+	private static final Pattern authorizationPattern = Pattern.compile("^Bearer +(?<token>[a-zA-Z0-9-._~+/]+=*)$",
```
therepanic, line 56: It might be clearer to use \x20+, since the RFC allows only spaces, but the current change is more minimal and seems like a better fit.

PR #19521, opened 2026-08-11
Avoid final-field mutation in security tests

### LINKED ISSUE #19520: WebTestUtils mutates a final SecurityContextHolderFilter field
## Describe the bug

`WebTestUtils.setSecurityContextRepository` replaces `SecurityContextHolderFilter.securityContextRepository` through `ReflectionTestUtils.setField`. That field is final. On current JDKs this emits a restricted final-field mutation warning, and the JDK warning states that this mutation will be blocked in a future release.

```text
WARNING: Final field securityContextRepository in class
org.springframework.security.web.context.SecurityContextHolderFilter has been mutated reflectively by
org.springframework.util.ReflectionUtils
WARNING: Mutating final fields will be blocked in a future release unless final field mutation is enabled
```

## To reproduce

1. Build a Spring Boot MVC application on JDK 26 with `spring-security-test`.
2. Configure MockMvc with `SecurityMockMvcConfigurers.springSecurity()`.
3. Execute a request that uses a test security context.

The warning is emitted from `WebTestUtils.setSecurityContextRepository`. The existing `WebTestUtilsTests.setSecurityContextRepositoryWhenSecurityContextHolderFilter` also exercises this reflective write.

## Expected behavior

Spring Security test support should replace the repository through a supported API rather than mutating a final field reflectively.

## Proposed solution

Add a null-checked `setSecurityContextRepository` method to `SecurityContextHolderFilter`, make the backing field non-final, and have `WebTestUtils` call the setter. Keep reflective reads unchanged to avoid broadening the production API further. Add/adjust tests for the setter and utility path.

gregjotau:
Closes gh-19520

`WebTestUtils` currently replaces `SecurityContextHolderFilter.securityContextRepository` reflectively even though it is final. Current JDKs warn about this operation and state it will be blocked in a future release.

This adds a null-checked repository setter and uses it from test support. Reflective reads remain unchanged, so the new setter is the only additional production API.

Verification (JDK 25):

```
./gradlew :spring-security-web:test \
  --tests org.springframework.security.web.context.SecurityContextHolderFilterTests \
  :spring-security-test:test \
  --tests org.springframework.security.test.web.support.WebTestUtilsTests
./gradlew format
```

### DIFF
--- a/test/src/main/java/org/springframework/security/test/web/support/WebTestUtils.java
+++ b/test/src/main/java/org/springframework/security/test/web/support/WebTestUtils.java
@@ -98,7 +98,7 @@ public abstract class WebTestUtils {
 		}
 		SecurityContextHolderFilter holderFilter = findFilter(request, SecurityContextHolderFilter.class);
 		if (holderFilter != null) {
-			ReflectionTestUtils.setField(holderFilter, "securityContextRepository", securityContextRepository);
+			holderFilter.setSecurityContextRepository(securityContextRepository);
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


gregjotau:
The Spring team can take over this. I dont' need ownership, I just want stuff done and efficient ;)

PR #19532, opened 2026-08-12
Document DaoAuthenticationProvider timing attack mitigation

### LINKED ISSUE #19082: Document Timing Attack Mitigation Support
`DaoAuthenticationProvider` protects against timing attacks by encoding an internally-generated password in lieu of the user's password when the user is not found, allowing both found and not-found paths to take approximately the same time.

It would be nice to have this noted in https://docs.spring.io/spring-security/reference/servlet/authentication/passwords/dao-authentication-provider.html and explained in https://docs.spring.io/spring-security/reference/servlet/exploits/index.html.

Some notes about it to consider:

* `DaoAuhenticationProvider` uses whatever password encoder that the website uses for users, giving a srong guarantee that the not-found and found paths will take the same order-of-magnitude time to execute
* There are limits to when this same timing is plausible, though. For example, a database that has more than one kind of password hash in it, Spring Security cannot anticipate how long to take in the not-found scenario since there are multiple found scenarios that take orders of magnitude different amounts of time. The best defense here is to upgrade your users passwords to a modern password algorithm.

AzazelSensei:
Closes gh-19082

Documents that `DaoAuthenticationProvider` still runs password validation through the configured `PasswordEncoder` when a username is not found (dummy encoded password + `matches`), so the not-found path stays roughly in the same time range as a found path using that encoder.

Also documents the limitation when a credential store mixes hash algorithms: Spring Security cannot pick one not-found duration that matches every found path, and upgrading stored passwords to a modern algorithm is the practical mitigation.

### Changes
- Short note on the `DaoAuthenticationProvider` reference page
- New Servlet exploits page on timing attacks, linked from the exploits index and nav

### Verification
- Cross-checked against `DaoAuthenticationProvider#retrieveUser`, `prepareTimingAttackProtection`, and `mitigateAgainstTimingAttack`
- Docs describe only the behavior present in that implementation

### DIFF
--- a/docs/modules/ROOT/nav.adoc
+++ b/docs/modules/ROOT/nav.adoc
@@ -110,6 +110,7 @@
 *** xref:servlet/exploits/headers.adoc[]
 *** xref:servlet/exploits/http.adoc[]
 *** xref:servlet/exploits/firewall.adoc[]
+*** xref:servlet/exploits/timing.adoc[Timing Attacks]
 ** xref:servlet/integrations/index.adoc[Integrations]
 *** xref:servlet/integrations/concurrency.adoc[Concurrency]
 *** xref:servlet/integrations/localization.adoc[Localization]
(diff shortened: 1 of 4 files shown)


AzazelSensei:
Superseded by clean branch: https://github.com/spring-projects/spring-security/pull/19533

PR #19533, opened 2026-08-12
Document DaoAuthenticationProvider timing-attack mitigation

### LINKED ISSUE #19082: Document Timing Attack Mitigation Support

AzazelSensei:
Document that `DaoAuthenticationProvider` mitigates username-enumeration timing attacks on the user-not-found path by performing equivalent password-encoding work, and note the limitation when mixed `PasswordEncoder` / hash algorithms are in use.

Changes are aligned with the current `retrieveUser` / `mitigateAgainstTimingAttack` implementation.

Closes #19082

### DIFF
--- a/docs/modules/ROOT/nav.adoc
+++ b/docs/modules/ROOT/nav.adoc
@@ -110,6 +110,7 @@
 *** xref:servlet/exploits/headers.adoc[]
 *** xref:servlet/exploits/http.adoc[]
 *** xref:servlet/exploits/firewall.adoc[]
+*** xref:servlet/exploits/timing.adoc[Timing Attacks]
 ** xref:servlet/integrations/index.adoc[Integrations]
 *** xref:servlet/integrations/concurrency.adoc[Concurrency]
 *** xref:servlet/integrations/localization.adoc[Localization]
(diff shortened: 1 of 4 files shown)


AzazelSensei:
Closing. Doesn't look like anyone's actually looking at this repo. Shame to waste the work.

PR #19585, opened 2026-08-22
Ensure WebSession ID is not changed after token refresh (Reactive)

### LINKED ISSUE #19424: Reactive RefreshOidcUserReactiveOAuth2AuthorizationSuccessHandler rotates the WebSession id on token refresh, stranding concurrent requests in a WebFlux BFF
### Describe the bug

In a reactive (WebFlux) OAuth2 client / BFF, `RefreshOidcUserReactiveOAuth2AuthorizationSuccessHandler`
(the default `authorizationSuccessHandler` of `RefreshTokenReactiveOAuth2AuthorizedClientProvider`)
re-saves the `SecurityContext` on every refresh-token response that carries a new `id_token`. It does
so through a **private, internally-constructed `WebSessionServerSecurityContextRepository`**, whose
`save(...)` ends with `WebSession#changeSessionId()`.

This has two consequences:

1. **The session id is rotated on every token refresh.** In a BFF the access token is refreshed
   transparently and frequently (every few minutes). If any other requests for that session are
   in flight when the refresh completes, they are still carrying the previous session id, which the
   rotation has just removed from the session store. Those requests then find no session, are treated
   as unauthenticated, and fail (401). If the front-end reacts to 401 by re-authenticating, the next
   refresh strands the next batch — a self-sustaining re-login loop. This is very easy to hit with a
   SPA that polls several endpoints in parallel.

2. **The repository is not configurable.** The handler instantiates
   `new WebSessionServerSecurityContextRepository()` internally and only exposes
   `setServerSecurityContextRepository(...)` on the handler itself — which is buried inside the
   refresh-token provider's success handler. It does **not** honor the `securityContextRepository`
   configured on the `ServerHttpSecurity` / oauth2Login chain, so there is no supported way to
   influence this behavior through the normal DSL.

Session-fixation rotation on refresh is also arguably unnecessary: a refresh does not change the
authenticated identity (no privilege transition), so rotating the id provides no fixation benefit
while breaking concurrent requests.

### To Reproduce

- Reactive OAuth2 login (BFF) with an OIDC provider that returns a new `id_token` on the refresh-token
  grant (e.g. Keycloak).
- Sessions stored via the default `WebSessionServerSecurityContextRepository`.
- A front-end that issues several requests in parallel while the access token is due for a background
  refresh.

Observed: when a refresh completes during a parallel request burst, the session id changes; the
concurrent in-flight requests lose their session and return 401; the browser re-authenticates; repeat.

### Expected behavior

A transparent token refresh should not rotate the session id (or, at minimum, should honor the
`securityContextRepository` configured on the chain so applications can opt out). Concurrent requests
for the same session should not be invalidated by a background refresh.

### Configuration / versions

- Spring Security 7.1.0 (the reactive handler is not present in 6.5.x; introduced with the 7.x line).
- Spring Boot 4.x, WebFlux, Reactor Netty.
- OIDC provider: Keycloak (rotates / re-issues id_token on refresh).

### Additional context

The feature that added refresh-time `Authentication`/`OidcUser` updates is #15509; this report is
about the reactive implementation's side effect on the WebSession id under concurrency. The servlet
variant surfaced a related follow-up in #18168.

Suggested directions:
- Have the reactive handler default to / respect the chain-configured `ServerSecurityContextRepository`.
- Persist the refreshed context without calling `changeSessionId()` on refresh (rotation only on the
  anonymous → authenticated transition).

### Workaround (for others hitting this)

Supply the handler with a `WebSessionServerSecurityContextRepository` subclass whose `save(...)` omits
`changeSessionId()`, wired via the refresh-token provider:

```java
var handler = new RefreshOidcUserReactiveOAuth2AuthorizationSuccessHandler();
handler.setServerSecurityContextRepository(nonRotatingRepository);

var provider = ReactiveOAuth2AuthorizedClientProviderBuilder.builder()
    .authorizationCode()
    .refreshToken(c -> c.authorizationSuccessHandler(handler))
    .build();
```

evgeniycheban:
Closes gh-19424

### DIFF
--- a/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshOidcUserReactiveOAuth2AuthorizationSuccessHandler.java
+++ b/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/RefreshOidcUserReactiveOAuth2AuthorizationSuccessHandler.java
@@ -24,9 +24,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.jspecify.annotations.Nullable;
 import reactor.core.publisher.Mono;
 
+import org.springframework.core.log.LogMessage;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.authority.mapping.GrantedAuthoritiesMapper;
@@ -79,7 +82,7 @@ public final class RefreshOidcUserReactiveOAuth2AuthorizationSuccessHandler
 			.map((c) -> c.get(ServerWebExchange.class));
 	// @formatter:on
 
-	private ServerSecurityContextRepository serverSecurityContextRepository = new WebSessionServerSecurityContextRepository();
+	private ServerSecurityContextRepository serverSecurityContextRepository = new NonRotatingWebSessionServerSecurityContextRepository();
 
 	private ReactiveJwtDecoderFactory<ClientRegistration> jwtDecoderFactory = new ReactiveOidcIdTokenDecoderFactory();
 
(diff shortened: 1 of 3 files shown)


jgrandja:
Thanks for the fix @evgeniycheban. This is now merged.

PR #19599, opened 2026-08-25
Test cache full rejection and replay detection

anthonydahanne:
**Test-only; no production code changes.**

### Adds a regression test to DPoPProofJwtDecoderFactoryTests for the replay cache under flooding.

The existing tests cover a plain single-proof replay, but not the case where the cache is filled past its size limit before the replay. 
The new test floods with proofs signed by other keys; split across two so neither hits the per-key request limit.
Then confirms a fresh proof from the original key is still accepted while a replayed one is rejected, so only replay detection (not the per-key throttle) can account for the rejection.

### DIFF
--- a/oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/DPoPProofJwtDecoderFactoryTests.java
+++ b/oauth2/oauth2-jose/src/test/java/org/springframework/security/oauth2/jwt/DPoPProofJwtDecoderFactoryTests.java
@@ -16,11 +16,16 @@
 
 package org.springframework.security.oauth2.jwt;
 
+import java.security.KeyPair;
+import java.security.KeyPairGenerator;
+import java.security.interfaces.RSAPrivateKey;
+import java.security.interfaces.RSAPublicKey;
 import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.function.Function;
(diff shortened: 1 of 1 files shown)


jgrandja:
@anthonydahanne Please do not report CVE's in the public. See https://spring.io/security-policy

jgrandja:
Closing and removing all contents

anthonydahanne:
Hello @jgrandja 

> @anthonydahanne Please do not report CVE's in the public. See https://spring.io/security-policy

I did not report any CVE.
My contribution was not production code, just an additional test to make sure the existing fix https://github.com/spring-projects/spring-security/commit/3b528e0c4c724f8635880ed19846e25133a20058  - which is public and created by you - was tested thoroughly, since the test attached with it did not cover a use case.

The community contribution that you just closed was enhancing test coverage of public code, without mentioning any CVE.

I do not understand your decision, happy to discuss it further though.

anthonydahanne:
@jgrandja thank you for re considering; I really appreciate it 🙏 .

I have re-submitted title / description / content - hopefully it's a valuable addition to the `spring-security` tests.

jgrandja:
Thank you @anthonydahanne and apologies for the confusion.

I will get to this soon. Thanks.

jgrandja:
@anthonydahanne I took a look at the test and I'm not understanding exactly what the difference is compared to `decodeWhenJtiAlreadyUsedThenThrowBadJwtException()`? The existing test demonstrates replay detection.

The test in this PR also demonstrates replay detection and it additionally floods the cache but I don't see the value of it? It floods the cache with 1200 entries - 600 entries per key bucket so within the per key limit of 1000. The default max entries allowed is 100,000 and also within the limit given 1200 entries plus 1 entry of the original proof.

Can you provide some clarification on the main goal of the test? Maybe I'm missing something?

anthonydahanne:
Hello @jgrandja !
Thanks for having a look!

The distinction I was aiming for is that `decodeWhenJtiAlreadyUsedThenThrowBadJwtException()` checks an immediate duplicate while the `jti` is still present.

The former implementation had `JtiCache.MAX_SIZE = 1000` and used `removeEldestEntry`, so that existing test still passed even though more than 1,000 unrelated proofs could evict the original jti and allow its replay. 
That is why I placed 1,200 proofs between the original use and replay, split across two keys to stay below the new per-key limit.

That said, you are right about the test as currently written: it injects the new `InMemoryCache` with its default maximum size of 100,000, so those 1,200 entries never exercise the current cache-full behavior. 
In this setup, the original jti simply remains cached, making the final assertion effectively equivalent to the existing replay test.

I suggest to you that either:
* we drop this test and PR, since as you noticed it's not adding value in this form, I'm sorry about that, I kept in my mind the maxSize set to 1000...
* I rework it to make it up-to-date and relevant: as in I use a deliberately small configured `maxSize`, fill it with unexpired entries, and verify that overflow is rejected rather than evicting an existing jti; verifying explicitly that the original entry remains to confirm that version genuinely distinguishes retention from a plain "cache full" rejection.

jgrandja:
@anthonydahanne Can you please rework the test and configure `maxSize` and `maxRequestsPerKey` to a smaller size and test the cache full rejection and replay detection. Thanks!

anthonydahanne:
@jgrandja I updated the test with smaller sizes and replay detection and rejections
let me know if that's valuable!
Thanks!

anthonydahanne:
👋

jgrandja:
Thanks for the update @anthonydahanne. This is now merged.

PR #19641, opened 2026-09-01
Use HexFormat instead of custom Hex codec

therepanic:
Replace internal `org.springframework.security.crypto.codec.Hex` utility class with standard Java 17 `java.util.HexFormat`. We already did something similar in this module earlier: 2f53a2edb34b690ba2f533732d62f7a58a55d577


core/src/main/java/org/springframework/security/core/token/Sha512DigestUtils.java
```diff
@@ -82,7 +81,7 @@ public static String shaHex(byte[] data) {
 	 * @return SHA digest as a hex string
 	 */
 	public static String shaHex(String data) {
-		return new String(Hex.encode(sha(data)));
+		return HexFormat.of().formatHex(sha(data));
```
therepanic, line 84: Previously, the string was created without regard for the string pool, but now the string pool is being used. Perhaps it should be wrapped in `new String()`?

therepanic:
It also renders https://github.com/spring-projects/spring-security/pull/19023 unnecessary.

core/src/main/java/org/springframework/security/core/token/Sha512DigestUtils.java
jzheaux: Thanks for checking. I believe the code is fine as-is. The `new String()` from before and `Hex.encode` returning `char[]` was not about preventing the value from entering the String pool. In that case, to add it, someone would need to call `.intern()`.

And since `HexFormat` does not do that, the interning semantics of this method do not change with your edit. 👍

jzheaux (changes requested):
Thanks, @therepanic! It's great to be able to change to use official libraries.

In addition to my inline feedback, will you please also still deprecate `Hex`? We need to do this since it is a `public` class. Will you add a `@Deprecated` annotation as well as a `@deprecated` javadoc note that indicates folks should use `HexFormat`?

This can go in an earlier commit so you can reference the other ticket you created.

therepanic:
Do we really need to deprecate Hex, given that Hex itself is marked `For internal use only`?

PR #19672, opened 2026-09-07
Handle null UserDetails in TokenBasedRememberMeServices

### LINKED ISSUE #19535: TokenBasedRememberMeServices::onLoginSuccess can throw a NPE when UserDetailsService::loadUserByUsername returns null.

GangEunzzang:
`TokenBasedRememberMeServices#onLoginSuccess` falls back to `UserDetailsService#loadUserByUsername` when the `Authentication` carries no password, and then reads `getPassword()` on the result without checking it. A `UserDetailsService` that returns `null` for a user (the reporter's case: only some users support remember-me) therefore throws a `NullPointerException` on login.

The method already returns without setting a cookie when the password cannot be obtained, so a `null` `UserDetails` now takes the same path, with its own debug message.

Adds `loginSuccessWhenUserDetailsServiceReturnsNullThenNoCookieIsSet`, built on the existing `udsWillReturnNull()` setup; it throws on `main` and passes here.

Closes gh-19535

### DIFF
--- a/web/src/main/java/org/springframework/security/web/authentication/rememberme/TokenBasedRememberMeServices.java
+++ b/web/src/main/java/org/springframework/security/web/authentication/rememberme/TokenBasedRememberMeServices.java
@@ -225,6 +225,10 @@ public class TokenBasedRememberMeServices extends AbstractRememberMeServices {
 		}
 		if (!StringUtils.hasLength(password)) {
 			UserDetails user = getUserDetailsService().loadUserByUsername(username);
+			if (user == null) {
+				this.logger.debug("Unable to obtain user details for user: " + username);
+				return;
+			}
 			password = user.getPassword();
 			if (!StringUtils.hasLength(password)) {
 				this.logger.debug("Unable to obtain password for user: " + username);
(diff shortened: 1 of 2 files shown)


GangEunzzang:
Closing — #19552 already has this exact change, with the reporter's feedback folded in, and I missed it before opening this one. Sorry for the noise.

PR #19714, opened 2026-09-14
Add success and failure handlers to code grant filter

### LINKED ISSUE #11069: Enable forward request (like ForwardSuccessHandler) in OAuth2AuthorizationCodeGrantFilter
**Expected Behavior**

The developer can override the redirect strategy in configuration to for OAuth2AuthorizationCodeGrantFilter to issue request forward into the existing web context (i.e., a relative path) at the end of authentication.

**Current Behavior**

Current behavior enables only redirects.  The redirect strategy in OAuth2AuthorizationCodeGrantFilter cannot be overridden by configuration.

The active request is the authorization server's callback after a successful grant.  Within the context of that request, this filter will send a request to the token URI to obtain an access token.  After completion, the filter responds with a redirect to the authorization server.  The target of the redirect is the original incoming request that triggered the authentication.

**Context**
 
How has this issue affected you? 
I have a client whose custom authorization server does not respond to this last redirect.  My application doesn't get the callback to the business level request.

What are you trying to accomplish?

My client is a SmartOnFhir implementation.  I'm implementing a 'launch' use case.   To initiate a launch, the SmartOnFhir implementation sends an opaque token as a query parameter to an endpoint in my application.   I'm using the oauth2client configuration.  In my controller, I call OAuth2AuthorizedClientProvider.authenticate to start the authentication flow.   SmartOnFhir will send confidential data within the access token after successful authentication.  My business logic has to obtain and decode the access token.   This is why I require a callup to my controller.  If the redirect worked, things would work.

What other alternatives have you considered?
I'm using the oauth2client configuration.  Maybe I should try the oauth2login configuration.  Using oauth2login, I believe I can easily use the ForwardAuthenticationSuccessHandler without a problem.  I'm not wild about this approach: I'm not that interested in user data in my application, and I have to figure out how to disable the login page (if necessary) because there is really only one authorization provider by the requirement of SmartOnFhir.

Are you aware of any workarounds?
No, only to copy the OAuth2AuthorizationCodeGrantFilter into my project, make a change within it to support forward, and add it to the security filter change before the existing original filter.

prasanna164-code:
Adds `AuthenticationSuccessHandler` and `AuthenticationFailureHandler` support to `OAuth2AuthorizationCodeGrantFilter`, following the structure of `OAuth2AuthorizationCodeGrantWebFilter`, so that the outcome of the Authorization Code grant can be customized, for example to forward into the application instead of redirecting.

**Filter**

- `setAuthenticationSuccessHandler`: invoked with the `OAuth2AuthorizationCodeAuthenticationToken` after the `OAuth2AuthorizedClient` has been saved. The default preserves the existing behaviour: redirect to the saved request if present, otherwise to the `redirect_uri` of the `OAuth2AuthorizationRequest`. `setRequestCache` keeps its contract and feeds the default handler.
- `setAuthenticationFailureHandler`: invoked with an `OAuth2AuthenticationException` wrapping the `OAuth2AuthorizationException` from the `AuthenticationManager`. The default preserves the existing redirect to the `redirect_uri` with `error`, `error_description` and `error_uri` appended.
- `RedirectStrategy` and `RequestCache` remain private to the default handlers; no public API is removed or deprecated.

**DSL and docs**

- `OAuth2ClientConfigurer.AuthorizationCodeGrantConfigurer` and the Kotlin `AuthorizationCodeGrantDsl` expose `authenticationSuccessHandler` and `authenticationFailureHandler`; the XML namespace gains `authentication-success-handler-ref` and `authentication-failure-handler-ref` on `authorization-code-grant`, so all three configuration styles offer the same options for this element.
- New "Handling the Authorization Response" section in the servlet OAuth2 client "Authorization Grant Support" page, with Java, Kotlin and XML examples.

**Tests**

- `OAuth2AuthorizationCodeGrantFilterTests`: null checks for both setters, custom success handler receives the token and suppresses the redirect while the authorized client is still saved, `ForwardAuthenticationSuccessHandler` forwards, custom failure handler receives the wrapped exception with the original `OAuth2Error`.
- `OAuth2ClientConfigurerTests`, `AuthorizationCodeGrantDslTests` and `OAuth2ClientBeanDefinitionParserTests`: both handlers wired through the Java DSL, the Kotlin DSL and XML.
- Existing redirect and request-cache tests are unchanged.

Closes gh-11069


prasanna164-code:
A few notes on the choices here, in case they come up in review:

- **Default success handler.** The plan in the issue suggested initialising the field to `SavedRequestAwareAuthenticationSuccessHandler`. I used a private default instead, because that handler falls back to its `defaultTargetUrl` (`/`) when there is no saved request, while this filter currently redirects to the `redirect_uri` of the `OAuth2AuthorizationRequest`. The private default keeps today's behaviour, and `setRequestCache` still feeds it as in the reactive filter.
- **Default failure handler.** `OAuth2AuthorizationCodeGrantWebFilter` rethrows by default. Here the default keeps the current servlet behaviour of redirecting to the `redirect_uri` with `error`, `error_description` and `error_uri`, so existing applications are unaffected.
- **Ordering.** The reactive filter invokes the success handler before saving the `OAuth2AuthorizedClient`; this filter saves first and then invokes the handler, which keeps the existing order and lets a forwarded request read the authorized client.

Happy to change any of these if you would rather match the reactive filter exactly.

PR #19765, opened 2026-09-18
Empty or null scope claim in dynamic client registration request should be lenient

Kehrlann:
**tl;dr:** allow `null` or `""` scope parameters in DCR requests

## Context

In MCP, some clients send dynamic client registration (DCR) requests with an empty scope claim.

For example, the [MCP inspector](https://github.com/modelcontextprotocol/inspector) sends the following registration request when no scope has been configured:

```json
{
  "redirect_uris": [
    "http://127.0.0.1:6274/oauth/callback"
  ],
  "token_endpoint_auth_method": "none",
  "grant_types": [
    "authorization_code"
  ],
  "response_types": [
    "code"
  ],
  "client_name": "MCP Inspector",
  "client_uri": "https://github.com/modelcontextprotocol/inspector",
  "scope": "",
  "application_type": "native"
}
```

This seems contentious, according to spec: [DCR 2.](https://datatracker.ietf.org/doc/html/rfc7591#section-2) is vague about it, while [OAuth2 3.3](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3) has a strict ABNF requiring at least one character.

Still, I think we should be lenient in what we accept.

## Bug description

Since org.springframework.security.oauth2.server.authorization.AbstractOAuth2ClientRegistration.AbstractBuilder#validate disallows empty scopes, this gets rejected with the following error:

```
org.springframework.security.oauth2.core.OAuth2AuthenticationException: An error occurred reading the OAuth 2.0 Client Registration: scope cannot be empty
```

A request with a `"scope": null` yields the same result ; however, a request omitting the `scope` param entirely succeeds.


jgrandja:
Thanks for the update @Kehrlann. This is now merged.

